PR #10180, opened 2019-10-15
Fix missing Assembler defines

slontis:
AES_ASM define was missing from libimplementations.a which disabled AESNI
aarch64 changes were made by xkqian.

##### Checklist

### DIFF
--- a/crypto/aes/build.info
+++ b/crypto/aes/build.info
@@ -61,9 +61,15 @@ ENDIF
 
 $COMMON=aes_misc.c aes_ecb.c $AESASM
 SOURCE[../../libcrypto]=$COMMON aes_cfb.c aes_ofb.c aes_ige.c aes_wrap.c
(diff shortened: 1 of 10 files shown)


levitte:
You will want to do this in:

`crypto/bn/build.info`
`crypto/modes/build.info`
`crypto/sha/build.info`
`crypto/whrlpool/build.info`

levitte:
Also, a comment above those lines would be a good thing, 'cause you can be sure that, in half a year, someone's going to ask.

Basically, what happens is that we have spread the implementations into several libraries, so we need to attach these defines to all affected libraries and modules.

slontis:
@levitte I have added extra entries..

slontis:
This might make the travis tests run a bit faster :)

slontis:
ta. Merged to master.

PR #10184, opened 2019-10-15
Add EVP_PKEY_CTX_new_provided()

levitte:
This works as much as possible EVP_PKEY_CTX_new_id(), except it takes
data that's relevant for providers, algorithm name and property query
string instead of NID and engine.

Additionally, if EVP_PKEY_CTX_new() or EVP_PKEY_CTX_new_id() was
called, the algorithm name in the EVP_PKEY context will be set to the
short name of the given NID (explicit or the one of the given
EVP_PKEY), thereby giving an easier transition from legacy methods to
provided methods.

The intent is that operations will use this information to fetch
provider methods implicitly as needed.

### DIFF
--- a/doc/man3/EVP_PKEY_CTX_new.pod
+++ b/doc/man3/EVP_PKEY_CTX_new.pod
@@ -10,6 +10,8 @@ EVP_PKEY_CTX_new, EVP_PKEY_CTX_new_id, EVP_PKEY_CTX_dup, EVP_PKEY_CTX_free - pub
 
  EVP_PKEY_CTX *EVP_PKEY_CTX_new(EVP_PKEY *pkey, ENGINE *e);
  EVP_PKEY_CTX *EVP_PKEY_CTX_new_id(int id, ENGINE *e);
+ EVP_PKEY_CTX *EVP_PKEY_CTX_new_provided(const char *name,
+                                         const char *propquery);
  EVP_PKEY_CTX *EVP_PKEY_CTX_dup(const EVP_PKEY_CTX *ctx);
  void EVP_PKEY_CTX_free(EVP_PKEY_CTX *ctx);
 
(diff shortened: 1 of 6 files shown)


levitte:
No tests are added here, that would be pointless.  That will come when other functions are adapted to this.  I already have creation of domain params and keys from numbers as well as domain param and key generation in the pipeline.

t8m:
Documentation?
Also what about the name and propquery memory management - this way these either have to be const strings or application needs to hold them until the EVP_PKEY_CTX is released and free them only after that - this is not very convenient.

levitte:
Is a `EVP_PKEY_CTX` held for very long or is it a per operation thing?  There's the exact same issue with `EVP_PKEY_new()`, the passed EVP_PKEY must be held for the life time of the operation.

So I have to wonder if it's really *that* inconvenient...

levitte:
I'm working on documentation now.

levitte:
I just noticed that the given pkey is uprefed, so maybe the passed strings are inconvenient unless copied.  I'm not *entirely* convinced yet, but am open to see examples.

t8m:
I do not insist on the copying, but the handling of these strings should be at least documented.

levitte:
If it turns out that we do need to copy the strings later on, we can deal with it then.

levitte:
CIs seem to be happy

doc/man3/EVP_PKEY_CTX_new.pod
```diff
@@ -2,31 +2,44 @@
...
-EVP_PKEY_CTX_dup() duplicates the context B<ctx>.
+The EVP_PKEY_CTX_new_provided() function allocates a public key
+algorithm context using the algorithm specified by I<name> and the
+property query I<propquery>.  The strings aren't duplicated, so they
+must remain unchanged for the lifetime of the returned B<EVP_PKEY_CTX>
+or of any of its duplicates.
 
-EVP_PKEY_CTX_free() frees up the context B<ctx>.
-If B<ctx> is NULL, nothing is done.
+EVP_PKEY_CTX_new_id() and The EVP_PKEY_CTX_new_provided() are normally
```
t8m: The .... and the ....

doc/man3/EVP_PKEY_CTX_new.pod
```diff
@@ -2,31 +2,44 @@
...
+used when no B<EVP_PKEY> structure is associated with the operations,
+for example during parameter generation of key generation for some
```
t8m: Shouldn't the `of` be `or` instead?

t8m (commented):
I suppose the actual use of these attributes will be implemented later.

doc/man3/EVP_PKEY_CTX_new.pod
levitte: Ah, yes

levitte:
> I suppose the actual use of these attributes will be implemented later

That's what I was hinting at in https://github.com/openssl/openssl/pull/10184#issuecomment-542162979

doc/man3/EVP_PKEY_CTX_new.pod
levitte: Done

t8m (approved):

levitte:
Merged.

a07c17ef57da20b7c6d075b303a6506f625dcd4e Add EVP_PKEY_CTX_new_provided()
028687c08b20ecc463afaf2d692542b88ebfe8bc doc/man3/EVP_PKEY_CTX_new.pod: change markup according to conventions

PR #10185, opened 2019-10-15
WIP: Bitflag based TLS Signature Algorithm Processing 

raja-ashok:
Introduced bit flags for the supported signature algorithm to optimise its processing. By which we can eliminate few nested iterations on key flows.

Code changes are in draft phase. Requesting to review and provide your opinions.

##### Checklist


kaduk:
I agree that there seems to be room for efficiency gains in this space, where we currently end up doing lots of nested list traversals over data that is fixed for the lifetime of the connection.  Making the SIGALGS structure so we can track indices and get constant-time lookup is probably a fine way to go, but I have mixed feelings about whether a bitmask is the right thing to do or not.  For one, I'm not sure when we'll need to know the set of algorithms without their order, and it also of course introduces a cap on the number of sigalgs that the current implementation can handle.

I think we'd also benefit from some statistics (multiple trials, stddev, etc.) to show the actual runtime impact of these changes, and I'd expect the commit messages to get fleshed out before anything actually gets merged (but I understand that this is still WIP so don't expect you to have done any of that before seeking initial feedback).

I'm curious if anyone else on @openssl/committers has thoughts about the use of a bitmask to track which sigalgs are present in a given list thereof.

beldmit:
> For one, I'm not sure when we'll need to know the set of algorithms without their order, and it also of course introduces a cap on the number of sigalgs that the current implementation can handle.

Do I correctly remember that there is a 1-byte codespace? If yes, it means that we have 8 bytes enough until the next version of the TLS protocol appears.

kaduk:
> > For one, I'm not sure when we'll need to know the set of algorithms without their order, and it also of course introduces a cap on the number of sigalgs that the current implementation can handle.
> 
> Do I correctly remember that there is a 1-byte codespace? If yes, it means that we have 8 bytes enough until the next version of the TLS protocol appears.

Around the TLS 1.3 timeframe we switched from separate (8-bit each) registries for signature algorithm and hash algorithm, so we now have a combined 16-bit codepoint space for SignatureScheme.  This takes into account things like Ed25519 that don't really have a separate "hash" step.

raja-ashok:
>  For one, I'm not sure when we'll need to know the set of algorithms without their order,

Currently the introduced new date type `SIGALGS` maintains signalg list in 3 form
- Ordered 2 octect value
- Ordered lookup array's index value
- Unordered bit flags

And passing variable of this type to all places, so based on the need we can access any one. 

> and it also of course introduces a cap on the number of sigalgs that the current implementation can handle.

That is true. I hope in near future also this list wont grow more than 64. In that case 64 bit variable will be sufficient for us.  


> I think we'd also benefit from some statistics (multiple trials, stddev, etc.) to show the actual runtime impact of these changes,

Sure. I will measure the statistics in few days and update over here.

Thanks for your valuable feedback.

raja-ashok:
With this optimization there is no change on the amount of connections made by `s_time`. But profiling report says CPU spend on `tls1_process_sigalgs`, `tls12_shared_sigalgs`, `tls_choose_sigalg` has been reduced with this changes.

t8m:

t8m:

PR #10169, opened 2019-10-14
KEYMGMT: export/import domain parameters

levitte:
Including DH and DSA functionality for import and export of domain parameters and keys

This is the foundation needed to manipulate domain parameters and keys with providers beyond what we currently do.  What we've done so far is allow libcrypto generated keys to be transported to providers to be able to use them for signature and key exchange operations.  However, with things like key generation, not to mention *any* domain parameter handling, we're soon going to find ourselves in a situation where the libcrypto generated keys are not relevant at all for some operations (such as domain parameter and key generation).

Tests will happen as soon as the diverse upcoming PKEY operations start using this functionality.  (next up is key generation)

PR #10187, opened 2019-10-15
Add EVP functionality to create domain params and keys by user data

levitte:
This is the EVP operation that corresponds to creating direct RSA, DH
and DSA keys and set their numbers, to then assign them to an EVP_PKEY,
but done entirely using an algorithm agnostic EVP interface.

- [x] Documentation
- [x] Test

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,13 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) Added functionality to create an EVP_PKEY from user data.  This
+     is effectively the same as creating a RSA, DH or DSA object and
+     then assign them to an EVP_PKEY, but directly using algorithm
+     agnostic EVP functions.  A benefit is that this should be future
+     proof for public key algorithms to come.
+     [Richard Levitte]
+
   *) Added the X509_LOOKUP_METHOD called X509_LOOKUP_store, to allow
      access to certificate and CRL stores via URIs and OSSL_STORE
      loaders.
--- a/CHANGES
+++ b/CHANGES
@@ -11,7 +11,7 @@
 
   *) Added functionality to create an EVP_PKEY from user data.  This
      is effectively the same as creating a RSA, DH or DSA object and
-     then assign them to an EVP_PKEY, but directly using algorithm
+     then assigning them to an EVP_PKEY, but directly using algorithm
      agnostic EVP functions.  A benefit is that this should be future
      proof for public key algorithms to come.
      [Richard Levitte]
--- a/crypto/evp/keymgmt_lib.c
+++ b/crypto/evp/keymgmt_lib.c
@@ -254,6 +254,24 @@ void evp_keymgmt_clear_pkey_cache(EVP_PKEY *pk)
     }
 }
 
+void *evp_keymgmt_fromdata(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
+                           const OSSL_PARAM params[], int domainparams)
+{
+    void *provctx = ossl_provider_ctx(EVP_KEYMGMT_provider(keymgmt));
+    void *provdata = domainparams
+        ? keymgmt->importdomparams(provctx, params)
+        : keymgmt->importkey(provctx, params);
+
+    evp_keymgmt_clear_pkey_cache(target);
+    if (provdata != NULL) {
+        EVP_KEYMGMT_up_ref(keymgmt);
+        target->pkeys[0].keymgmt = keymgmt;
+        target->pkeys[0].provdata = provdata;
+        target->pkeys[0].domainparams = domainparams;
+    }
+
+    return provdata;
+}
 
 /* internal functions */
 /* TODO(3.0) decide if these should be public or internal */
--- a/doc/man3/EVP_PKEY_fromdata.pod
+++ b/doc/man3/EVP_PKEY_fromdata.pod
@@ -1,0 +1,69 @@
+=pod
+
+=head1 NAME
+
+EVP_PKEY_param_fromdata_init, EVP_PKEY_key_fromdata_init, EVP_PKEY_fromdata,
+EVP_PKEY_param_fromdata_settable, EVP_PKEY_key_fromdata_settable
+- functions to create domain parameters and keys from user data
+
+=head1 SYNOPSIS
+
+ #include <openssl/evp.h>
+
+ int EVP_PKEY_param_fromdata_init(EVP_PKEY_CTX *ctx);
+ int EVP_PKEY_key_fromdata_init(EVP_PKEY_CTX *ctx);
+ int EVP_PKEY_fromdata(EVP_PKEY_CTX *ctx, EVP_PKEY **ppkey, OSSL_PARAM params[]);
+ const OSSL_PARAM *EVP_PKEY_param_fromdata_settable(EVP_PKEY_CTX *ctx);
+ const OSSL_PARAM *EVP_PKEY_key_fromdata_settable(EVP_PKEY_CTX *ctx);
+
+=head1 DESCRIPTION
+
+EVP_PKEY_param_fromdata_init() initializes a public key algorithm context
+for creating domain parameters from user data.
+
+EVP_PKEY_key_fromdata_init() initializes a public key algorithm context for
+creating a key from user data.
+
+EVP_PKEY_fromdata() creates domain parameters or a key, given data from
+I<params> and a context that's been initialized with
+EVP_PKEY_param_fromdata_init() or EVP_PKEY_key_fromdata_init().  The result is
+written to I<*ppkey>.
+
+EVP_PKEY_param_fromdata_settable() and EVP_PKEY_key_fromdata_settable()
+get a constant B<OSSL_PARAM> array that describes the settable parameters
+that can be used with EVP_PKEY_fromdata().
+See L<OSSL_PARAM(3)> for the use of B<OSSL_PARAM> as parameter descriptor.
+
+=head1 NOTES
+
+These functions only work with key management methods coming from a
+provider.
+
+=for comment We may choose to make this available for legacy methods too... 
+
+=head1 RETURN VALUES
+
+EVP_PKEY_key_fromdata_init(), EVP_PKEY_param_fromdata_init() and
+EVP_PKEY_fromdata() return 1 for success and 0 or a negative value for
+failure.  In particular a return value of -2 indicates the operation is
+not supported by the public key algorithm.
+
+=head1 SEE ALSO
+
+L<EVP_PKEY_CTX_new(3)>, L<provider(7)>
+
+=head1 HISTORY
+
+These functions were added in OpenSSL 3.0.
+
+=head1 COPYRIGHT
+
+Copyright 2019 The OpenSSL Project Authors. All Rights Reserved.
+
+Licensed under the Apache License 2.0 (the "License").  You may not use
+this file except in compliance with the License.  You can obtain a copy
+in the file LICENSE in the source distribution or at
+L<https://www.openssl.org/source/license.html>.
+
+=cut
+
--- a/test/evp_fromdata_test.c
+++ b/test/evp_fromdata_test.c
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2019 The OpenSSL Project Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License 2.0 (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
+ */
+
+#include <openssl/evp.h>
+#include <openssl/provider.h>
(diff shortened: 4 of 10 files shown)


levitte:
@slontis, I think you're interested in this

levitte:
Note that the added functionality deviates slightly from a previous pattern.  The param and key *generation* functionality has these functions:

``` C
int EVP_PKEY_paramgen_init(EVP_PKEY_CTX *ctx);
int EVP_PKEY_paramgen(EVP_PKEY_CTX *ctx, EVP_PKEY **ppkey);
int EVP_PKEY_keygen_init(EVP_PKEY_CTX *ctx);
int EVP_PKEY_keygen(EVP_PKEY_CTX *ctx, EVP_PKEY **ppkey);
```

The making functions are these:

``` C
int EVP_PKEY_parammake_init(EVP_PKEY_CTX *ctx);
int EVP_PKEY_keymake_init(EVP_PKEY_CTX *ctx);
int EVP_PKEY_make(EVP_PKEY_CTX *ctx, EVP_PKEY **ppkey, OSSL_PARAM param[]);
```

Two functions to init, but just one to actually make.  I find it quite unnecessary to have two making functions when they can differentiate by the operation number that's already assigned in the EVP_PKEY context.

levitte:
This will remain WIP until #10169 and #10184 are in

CHANGES
```diff
@@ -9,6 +9,14 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) Added functionality to create an EVP_PKEY context based on data
```
slontis: This seems like a confusing description here. The API docs are clearer than this.

levitte: Uhmmmmm...  this description has nothing to do with this PR, it belongs with #10184...  did I mix the CHANGES changes up? ...  yup, this seems to be a rebasse goof-up

crypto/evp/keymgmt_lib.c
```diff
@@ -200,6 +200,28 @@ void evp_keymgmt_clear_pkey_cache(EVP_PKEY *pk)
...
+    evp_keymgmt_clear_pkey_cache(target);
+    if (provdata != NULL) {
+        EVP_KEYMGMT_up_ref(keymgmt);
+        target->pkeys[0].keymgmt = keymgmt;
+#if 1
+        target->pkeys[0].provkey = provdata;
+#else
+        target->pkeys[0].provdata = provdata;
+        target->pkeys[0].domainparams = domainparams;
+#endif
```
levitte: As soon as #10169 is in, lines 215-220 will be reduced to this:

``` C
        target->pkeys[0].provdata = provdata;
        target->pkeys[0].domainparams = domainparams;
```

levitte:
Now that #10169 and #10184 have been merged, I rebased this.  It's still in draft, as a test program should be written for this.

levitte:
Not WIP any more

slontis:
I am not completely sold on the name 'make'.
EVP_PKEY_fromparams
EVP_PKEY_fromraw

which could then match something like..
EVP_PKEY_toparams

levitte:
I have no problems changing to those names.  But...  `EVP_PKEY_fromraw`?  Does that imply that we need that "raw key" functionality in providers?  I thought that was a hack needed to the MAC-as-PKEY stuff, which is entirely libcrypto side...  or is this somethiing else?

levitte:
Although...  thinking a bit more about it, those names may add confusion.  There are two thing that we need to build, domain parameters (classic *p* and *g* and possibly *q* for DH / DSA), and the key itself...  so what would we then call the init functions?  `EVP_fromparams_params_init`, `EVP_fromparams_key_init`?

My "I have no problems..." was a bit premature, it seems

slontis:
The fromraw was just an alternate suggestion for the name.. I think the fromparams is better.
I was hoping I could use these for the self tests (which sit in the fips provider).

slontis:
We have given params 2 different meanings now, which makes using params very confusing.
Coming up with names is hard :)
These functions load a PKEY with custom user data. So something that indicates we are setting userdata into a key would be a good name.

EVP_PKEY_param_setuserdata_init
EVP_PKEY_key_setuserdata_init
EVP_PKEY_setuserdata

levitte:
Hmmm, I think we're getting somewhere...  Mind you, I *like* having "from" in there, so merging all ideas together, what would you say about this?

EVP_PKEY_param_fromdata_init
EVP_PKEY_key_fromdata_init
EVP_PKEY_fromdata

(could be "fromuserdata" too, but I'm unsure if "user" is really needed in there)

slontis:
Yeah sounds good to me - I like 'from' because you can then do 'to' for the other direction if required.

levitte:
I put this back as WIP 'cause I realised it's not quite ready yet.

So here's an auxiliary function that I forgot to implement; the function that returns a `const OSSL_PARAM` array describing what can be used with (presumed) `EVP_PKEY_fromdata`.  Is `EVP_PKEY_fromdata_settable` good enough?  It's not *exactly* wrong, but not perfect (enough) either...

slontis:
> In my mind, it's not _exactly_ wrong, but not perfect (enough) either...

I guess that matches the other ones.. 
Does this return every type of key data (for RSA, DSA, DH, EC).

levitte:
It should return the data corresponding to the algorithm, so should take a EVP_PKEY_CTX as argument.

slontis:
Which comes from EVP_PKEY_CTX_new_provided - which has the type.. Ok.

levitte:
(or even the old ones, 'cause they do set the algorithm name too, just not the props)

levitte:
Ok, I'll got with that.  It's also up for reconsideration as long as we haven't released 3.0, so I'm not terribly worried about getting it absolutely right in its final form now, as long ass it's good enough for us to be happy.

levitte:
I'm wondering about the `settable` names again.  Is there a reason for them to be tied by name to some specific operation?  Can't they just be this `EVP_PKEY_param_settable()` and `EVP_PKEY_key_settable`?

levitte:
Ping.  Question needs answering

beldmit:
> The making functions are these:
> 
> ```c
> int EVP_PKEY_parammake_init(EVP_PKEY_CTX *ctx);
> int EVP_PKEY_keymake_init(EVP_PKEY_CTX *ctx);
> int EVP_PKEY_make(EVP_PKEY_CTX *ctx, EVP_PKEY **ppkey, OSSL_PARAM param[]);
> ```
> 
> Two functions to init, but just one to actually make. I find it quite unnecessary to have two making functions when they can differentiate by the operation number that's already assigned in the EVP_PKEY context.

Is it reasonable for the algorithms with parameters treated as special objects?

levitte:
> Is it reasonable for the algorithms with parameters treated as special objects?

Not really. A key is a key is a key. That it carries domain parameters should not make a difference.

beldmit:
> > Is it reasonable for the algorithms with parameters treated as special objects?
> 
> Not really. A key is a key is a key. That it carries domain parameters should not make a difference.

It may take time to generate parameters for a, e.g., DSA key, and one may want possible to manipulate the parameters generation and the key generation separately, and/or serialize the params. If it is out of the scope of the API you suggest, it's OK.

levitte:
> It may take time to generate parameters for a, e.g., DSA key, and one may want possible to manipulate the parameters generation and the key generation separately, and/or serialize the params. If it is out of the scope of the API you suggest, it's OK.

Not sure I understand...  You *do* generate the parameters separately, and they end up in a `EVP_PKEY`, and then you *do* generate the key itself separately, and it ends up in another `EVP_PKEY`.

levitte:
However, please note that *this PR isn't about domparam or key generation*.  There will be another PR to do that.

levitte:
The question I'm still waiting for an answer to is this:

> I'm wondering about the `settable` names again.  Is there a reason for them to be tied by name to some specific operation?  Can't they just be this `EVP_PKEY_param_settable()` and `EVP_PKEY_key_settable`?

slontis:
> I'm wondering about the `settable` names again

If this operation is only related to the  fromdata  methods then the name is fine..

levitte:
OK, I will stay with that, but will add a TODO comment to re-evaluate the name later.

levitte:
This is now ready for final evaluation

CHANGES
```diff
@@ -9,6 +9,14 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) Added functionality to create an EVP_PKEY from user data.  This
+     is effectively the same as creating an RSA, DH or DSA object and
+     assign them numbers and then assign them to an EVP_PKEY, but is
```
slontis: Although I know what you mean - it is not very clear. Maybe somnething like...

This is equivalent to creating a EVP_PKEY using RSA, DH, or DSA methods.

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -0,0 +1,69 @@
+=pod
+
+=head1 NAME
+
+EVP_PKEY_param_fromdata_init, EVP_PKEY_key_fromdata_init, EVP_PKEY_fromdata,
+EVP_PKEY_param_fromdata_settable, EVP_PKEY_key_fromdata_settable
+- functions to create domain parameters and keys from numbers
```
slontis: replace 'numbers' with user data.

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -0,0 +1,69 @@
...
+EVP_PKEY_key_fromdata_init() initializes a public key algorithm context for
+creating a key from user data.
+
+EVP_PKEY_fromdata() creates domain parameters or a key, given data from
+I<params> and a context that's been initialized with
+EVP_PKEY_param_fromdata_init() or EVP_PKEY_key_fromdata_init().  The result is
+written to I<*ppkey>.
+
+EVP_PKEY_param_fromdata_settable() and EVP_PKEY_key_fromdata_settable()
+get a constant B<OSSL_PARAM> array that decribes the settable parameters
```
slontis: describes

test/evp_fromdata_test.c
```diff
@@ -0,0 +1,87 @@
...
+#include <openssl/provider.h>
+#include <openssl/params.h>
+#include <openssl/core_names.h>
+#include "internal/nelem.h"
+#include "crypto/evp.h"          /* For the internal API */
+#include "testutil.h"
+
+static int test_fromdata_rsa(void)
+{
+    /* Array indexes */
```
slontis: Maybe just move these defines up - The comment looks ugly indented.

CHANGES
levitte: How about I simply remove "and assign them numbers"?

doc/man3/EVP_PKEY_fromdata.pod
levitte: Okie

doc/man3/EVP_PKEY_fromdata.pod
levitte: Okie

test/evp_fromdata_test.c
levitte: Okie

levitte:
Okie, @slontis, I hope I've addressed all your comments

levitte:
I rebased, and squashed while I was at it.

@slontis, or anyone really...  final verdict?

CHANGES
```diff
@@ -9,6 +9,13 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) Added functionality to create an EVP_PKEY from user data.  This
+     is effectively the same as creating a RSA, DH or DSA object and
+     then assign them to an EVP_PKEY, but directly using algorithm
```
slontis: NIT: assigning

test/evp_fromdata_test.c
```diff
@@ -0,0 +1,87 @@
...
+
+    if (!TEST_ptr(ctx = EVP_PKEY_CTX_new_provided("RSA", NULL)))
+        goto err;
+
+    if (!TEST_true(EVP_PKEY_key_fromdata_init(ctx))
+        || !TEST_true(EVP_PKEY_fromdata(ctx, &pk, fromdata_params)))
+        goto err;
+
+    /*
+     * TODO(3.0) We can't do much more at this point without using internals,
```
slontis: I am not sure what you mean here..
Surely the EVP_PKEY_sign operation will work at some point?

levitte: Yes, exactly.  So when that's in place, this test could get some additional code that does something with the key that was created, to verify that the key is functional.

slontis (approved):
Provided the tests pass, this looks good to me. Good work..

levitte:
Merged.

46e2dd05ef1456e3e8fc3d12bd839bae01576c19 Add EVP functionality to create domain params and keys by user data
ab14d2af5386897eba8307c9f3220a6d775c0898 Add a test for EVP_PKEY_keymake() and EVP_PKEY_make()

Issue #10166, opened 2019-10-13
Configure: accept --api=3 and --api=3.x options


mspncp:
Currently the configure command only accepts `--api=3.0.0`, although the third digit is only a patch level which does not permit api changes. It should be `--api=3.0` instead. (This is the bug report).

In fact, it would be convenient, if missing trailing version numbers would be automatically set to 0, i.e., if one could specify `--api=3` and this would be expanded to `--api=3.0` (this is the feature request).

In fact, do we actuall support deprecations at minor revision boundaries? If not, then we should allow only `--api=n` for n>=3.

```
msp@msppc:~/openssl/openssl$ perl Configure linux-x86_64 --api=3.0.0
Configuring OpenSSL version 3.0.0-dev for target linux-x86_64
Using os-specific seed configuration
Creating configdata.pm
Running configdata.pm
Creating Makefile
...
```

```
msp@msppc:~/openssl/openssl$ perl Configure linux-x86_64 --api=3.0

Failure!  build file wasn't produced.
Please read INSTALL and associated NOTES files.  You may also have to look over
your available compiler tool chain or change your configuration.

***** Unsupported api compatibility level: 3.0
```

```
msp@msppc:~/openssl/openssl$ perl Configure linux-x86_64 --api=3

Failure!  build file wasn't produced.
Please read INSTALL and associated NOTES files.  You may also have to look over
your available compiler tool chain or change your configuration.

***** Unsupported api compatibility level: 3
```

cc @levitte

levitte:
We do support deprecation at minor boundaries. *Removal* is restricted to major boundaries, however.

piraces:
Hi, I will like to contribute and I think this is a good starting point for it.
Can I work on this bug and feature request?

mspncp:
> Can I work on this bug and feature request?

Sure, every new contributor is welcome. I think @levitte has so many other things to do at the moment that he won't mind if you pick this issue. But of course if you need some advice or help during your work, don't hesitate to ask him, he's our Perl Wizard ;-).

mattcaswell:
This was fixed by overridden by #10364. Closing.

Issue #5042, opened 2018-01-09
Please follow semantic versioning


probonopd:
OpenSSL is a central piece of infrastructure on virtually every Linux system. As such, applications linking to it should have the assurance that new OpenSSL versions don't break applications compiled against older OpenSSL versions.

According to https://semver.org/,

-  **MAJOR** version when you make incompatible API changes,
-  **MINOR** version when you add functionality in a _backwards-compatible_ manner, and
-  **PATCH** version when you make backwards-compatible bug fixes.

So OpenSSL should have added "functionality in a backwards-compatible manner" going from 1.0 to 1.1. However, applications compiled against OpenSSL 1.0 seem to break when OpenSSL is updated to 1.1.

There are two ways to address this:
1. Ensure that all applications compiled against OpenSSL 1.0 continue to run when OpenSSL is updated to 1.1; or
2. Rename OpenSSL 1.1 to 2.0 in order to make clear that you made incompatible API changes; advise users to install 1.x in addition to 2.x if they want to continue to run applications compiled against 1.x

If application developers (such as Firefox and Chrome) cannot rely on backward compatibility in OpenSSL, they are effectively forced to ship privately bundled versions, leading to duplication and locking them out from system-wide security patches.
 
References:

- https://github.com/probonopd/AppImages/commit/6c7473d8cdaaa2572248dcc53d7f617a577ade6b
- http://stackoverflow.com/questions/32644157/forcing-a-binary-to-use-a-specific-newer-version-of-a-shared-library-so
- https://abi-laboratory.pro/tracker/timeline/openssl/

probonopd:
Even though OpenSSL may not have been using semantic versioning so far, it is _now_ the time to start embracing it. The project is just now going from 1.0 to 1.1 - so this is the last possible chance to fix this before 1.1 find its way into LTS distributions. At which point the mess will not be easily reversible anymore.

levitte:
Actually, our versioning scheme *is* semantic as defined by http://semver.org, semantically speaking.  The only actual difference is our version number *syntax*.

[...]

PR #10189, opened 2019-10-15
Configure: accept --api=3 and --api=3.x options

piraces:
Resolves: https://github.com/openssl/openssl/issues/10166

Restrict `--api` option to allow detail until minor versions (as there is only deprecation at minor boundaries, not patches). 
For example: `--api=3.0`.

Add support for expanding api version if missing trailing minor version number. 
For example: `--api=3`.

##### Checklist

### DIFF
--- a/Configure
+++ b/Configure
@@ -45,7 +45,7 @@ my $usage="Usage: Configure [no-<cipher> ...] [enable-<cipher> ...] [-Dxxx] [-lx
 #
 # --cross-compile-prefix Add specified prefix to binutils components.
 #
-# --api         One of 0.9.8, 1.0.0, 1.0.1, 1.0.2, 1.1.0, 1.1.1, or 3.0.0 / 3.
+# --api         One of 0.9, 1.0, 1.1, or 3.0.0 / 3.
 #               Do not compile support for interfaces deprecated as of the
 #               specified OpenSSL version.
 #
@@ -182,15 +182,12 @@ our $BSDthreads="-pthread -D_THREAD_SAFE -D_REENTRANT";
 #
 # API compatibility name to version number mapping.
 #
-my $maxapi = "3.0.0";           # API for "no-deprecated" builds
+my $maxapi = "3.0";           # API for "no-deprecated" builds
 my $apitable = {
-    "3.0.0" => 3,
-    "1.1.1" => 2,
-    "1.1.0" => 2,
-    "1.0.2" => 1,
-    "1.0.1" => 1,
-    "1.0.0" => 1,
-    "0.9.8" => 0,
+    "3.0" => 3,
+    "1.1" => 2,
+    "1.0" => 1,
+    "0.9" => 0,
 };
 
 our %table = ();
@@ -839,7 +836,14 @@ while (@argvcopy)
                         }
                 elsif (/^--api=(.*)$/)
                         {
-                        $config{api}=$1;
+                        my $api = $1;
+                        if (index($api, '.') == -1)
+                                {
+                                $api = $api . ".0";
+                                }
+                        $config{api}=$api;
+                        die "Unknown --api choice $1\n"
+                            unless $api =~ /^([0-9]*(.[0-9]+)?)$/;
                         }
                 elsif (/^--libdir=(.*)$/)
                         {
--- a/INSTALL
+++ b/INSTALL
@@ -141,11 +141,11 @@
  --openssldir depend in what configuration is used and what Windows
  implementation OpenSSL is built on.  More notes on this in NOTES.WIN):
 
-  --api=x.y.z
+  --api=x(.y)?
                    Don't build with support for deprecated APIs below the
-                   specified version number. For example "--api=1.1.0" will
+                   specified version number. For example "--api=1.1" will
                    remove support for all APIS that were deprecated in OpenSSL
-                   version 1.1.0 or below. This is a rather specialized option
+                   version 1.1 or below. This is a rather specialized option
                    for developers. If you just intend to remove all deprecated
                    APIs entirely (up to the current version), it is easier
                    to add the 'no-deprecated' option instead (see below).


piraces:
**_Note:_** I have already submitted CLA, I'm awaiting for response.

mspncp:
Sorry, I should have told you in advance, but I didn't think of it: The versioning in OpenSSL is a little bit more complicated as you might think. OpenSSL changed it's versioning scheme recently when switching to version 3.0:

While version numbers in software libraries are often three decimals separated by dots, interpreted as <major>.<minor>.<patch>, OpenSSL had a slightly shifted interpretation of the version numbers: it was effectively 1.<minor>.<minor><patch>, where the <patch> is represented by a letter, not a digit (like in 1.1.0d).

There was a lot of discussion (e.g. #5042), after which the OpenSSL team decided to follow more closely (but not exactly) the so called semantic versioning scheme (https://semver.org). You can read about the versioning change in [this blog post](https://www.openssl.org/blog/blog/2018/11/28/version).

However, the new scheme applies only to version numbers >= 3.0, the old number need to remain as they are. This means, you can't shorten 1.1.0 to 1.1. Please leave all version numbers 0.x.y and 1.x.y as they are.

I hope that my explanation was helpful. If you have further questions, don't hesitate to ask.

levitte:
I have to disagree with you there, @mspncp. Even though a bit clumsily specified, the --api option expresses what *major* release boundary the API should be presented at (we don't *really* do that, but the boundary is still the same), so you can say 1.0.0, 1.0.1 or 1.0.2 all you want, but the boundary is still the same (1.0.0), so 1.0 is correct in those terms. The same goes for 1.1. 0.9.8 is the ugly duckling in this scenario...

mspncp:
> I have to disagree with you there, @mspncp.

Oh, it was not my intention to disagree with you. If I did, it was due to misunderstanding on my side. The least thing I would want to do is to start yet another discussion about versioning semantics. ;-)

Part of my misunderstanding (as I noticed yesterday when reading Matt's [blog post](https://www.openssl.org/blog/blog/2018/11/28/version)) comes from the fact that although _de-facto_ we acted as if the meaning of the version number was  `1.MAJOR.MINOR[PATCH]`  (where PATCH is a letter), the official interpretation was never changed, so it is still `MAJOR.MINOR.FIX[PATCH]`. So from your saying that deprecation can occur at minor revision numbers, I inferred incorrectly that you were talking about three digit version numbers 1.x.y without patch letter.

In the end I will follow your opinion, you have the deepest knowledge of the licensing changes. However, I still see an inconsistency: You say that api level changes only at major versions. What sense does it make then to specify `--api=3.1` as configure argument? Shouldn't the consequence be that two digit arguments are allowed only for old versions, but not for version 3 and beyound? Looking at the naming scheme for the defines, this would be more consistent IMHO.

[include/openssl/macros.h:89:](https://github.com/openssl/openssl/blob/3d5a7578e09a984c6475b1c008f5c76f850328cb/include/openssl/macros.h#L89) #  define OPENSSL_API_4 1
[include/openssl/macros.h:94:](https://github.com/openssl/openssl/blob/3d5a7578e09a984c6475b1c008f5c76f850328cb/include/openssl/macros.h#L94) #  define OPENSSL_API_3 0
[include/openssl/macros.h:97:](https://github.com/openssl/openssl/blob/3d5a7578e09a984c6475b1c008f5c76f850328cb/include/openssl/macros.h#L97) #  define OPENSSL_API_3 1
[include/openssl/macros.h:102:](https://github.com/openssl/openssl/blob/3d5a7578e09a984c6475b1c008f5c76f850328cb/include/openssl/macros.h#L102) #  define OPENSSL_API_1_1_0 0
[include/openssl/macros.h:105:](https://github.com/openssl/openssl/blob/3d5a7578e09a984c6475b1c008f5c76f850328cb/include/openssl/macros.h#L105) #  define OPENSSL_API_1_1_0 1
[include/openssl/macros.h:110:](https://github.com/openssl/openssl/blob/3d5a7578e09a984c6475b1c008f5c76f850328cb/include/openssl/macros.h#L110) #  define OPENSSL_API_1_0_0 0
[include/openssl/macros.h:113:](https://github.com/openssl/openssl/blob/3d5a7578e09a984c6475b1c008f5c76f850328cb/include/openssl/macros.h#L113) #  define OPENSSL_API_1_0_0 1
[include/openssl/macros.h:118:](https://github.com/openssl/openssl/blob/3d5a7578e09a984c6475b1c008f5c76f850328cb/include/openssl/macros.h#L118) #  define OPENSSL_API_0_9_8 0
[include/openssl/macros.h:121:](https://github.com/openssl/openssl/blob/3d5a7578e09a984c6475b1c008f5c76f850328cb/include/openssl/macros.h#L121) #  define OPENSSL_API_0_9_8 1

mattcaswell:
Close/reopen to kick CLA bot.

richsalz:
Should we remove the define's and ifdef's for releases that are out of support?

piraces:
Thank you for your help. If I understand well, we would need one digit for the `--api` option (as @levitte said).

The question is what to do for older versions `n<3`

Also the support for expanding the api version wouldn't make sense for newer api versions (`n>=3`).

What should I do?

levitte:
I'm having some second thoughts about this, because the `$apitable` lists minor releases that have actually existed.  With a more "flexible" parsing of the version, someone could write `--api=1.0.42` and get away with it, and while it *technically* wouldn't matter, those that are a bit more - how to say this diplomatically - "puristically inclined" will shiver in horror.

However, considering the new version scheme from 3.0 on, shortening the indicator to two digits would be correct.  And additionally, accept a single number that most be 3 or above and cannot be over the current major version number (found in `$config{major}`) is reasonable.  Again, this can be done with parsing magic, or by simply adding the `$apitable` entry `3 => 3'.

I guess it's a question of how precise and how deterministic we want this to be...  and also how we think around maintenance.

Questions, questions, sorry that I don't have many straight answers...

levitte:
> Only listen to what @levitte tells you;-)

I have *no* need of hubris when others take so very good care of it for me :wink:

mspncp:
> However, considering the new version scheme from 3.0 on, shortening the indicator to two digits would be correct.
> ...
> Questions, questions, sorry that I don't have many straight answers...


But the answer to one question (posted in https://github.com/openssl/openssl/pull/10189#issuecomment-542590311) would interest me: What is the use case for a two digit `--api=x.y` flag, if api changes like removals can only occur at major version boundaries? And if two digits make sense, shouldn't our defines then be

```
    OPENSSL_API_3_0, OPENSSL_API_3_1, ..., OPENSSL_API_4_0, OPENSSL_API_4_1, ...
```

instead of
	
```
    OPENSSL_API_3, OPENSSL_API_3_1, ..., OPENSSL_API_4, OPENSSL_API_4_1, ...
```

levitte:
> In the end I will follow your opinion, you have the deepest knowledge of the licensing changes. However, I still see an inconsistency: You say that api level changes only at major versions. What sense does it make then to specify `--api=3.1` as configure argument? Shouldn't the consequence be that two digit arguments are allowed only for old versions, but not for version 3 and beyond? Looking at the naming scheme for the defines, this would be more consistent IMHO.

The confusion is historical.  Here are the major and minor *releases* that we have made:

* 0.9.8 - I'm not exactly sure how to qualify that release, I think it mostly serves as the oldest point where we know we had a fairly stable API.  I *think* that it can be regarded as some kind of 1.0.0-RC.
* 1.0.0 - A *major* release.  We simply decided that we'll go for it...  with much apprehension, I might add, it was quite comfortable to remain in the free-for-all pre-1.0.0 closet :wink:
* 1.0.1 - A *minor* release.
* 1.0.2 - A *minor* release.
* 1.1.0 - A *major* release.
* 1.1.1 - A *minor* release.
* 3.0 - A future *major* release.
* 3.x - A future *minor* release.

If your whole world centers around semantic versioning, that versioning scheme looks weird, and there are different opinions within the central team what exactly each digit represented, and that's why I'm expressing it in terms of **releases**, to separate it from the semantics of *versions*.  Our release scheme until now has simply not been aligned with a semantic versioning scheme, and that's that.  We're re-aligning releases and version numbers starting with 3.0.

That is why the set of checking macros look so weird.

As for the permitted values for `--api`, it simply about letting humans enter something that makes sense to them.  Why should anyone have to remember that only so and so many digits should be entered, when it's easy for us to have a table of existing and foreseeable versions?

levitte:
There's been related rework in #10364, so this PR is not quite relevant any more.

I want to thank you, @piraces, because this work you did sparked and was a huge inspiration for the development that happened, so even if *this* work doesn't get merged in its current form, it was still valuable work.

PR #10194, opened 2019-10-16
Fix memory leak with no-ec config

raja-ashok:
There is a memory leak when `no-ec` is configured, which is fixed. Requesting to review.

### DIFF
--- a/ssl/ssl_lib.c
+++ b/ssl/ssl_lib.c
@@ -1195,9 +1195,9 @@ void SSL_free(SSL *s)
 #ifndef OPENSSL_NO_EC
     OPENSSL_free(s->ext.ecpointformats);
     OPENSSL_free(s->ext.peer_ecpointformats);
+#endif                          /* OPENSSL_NO_EC */
     OPENSSL_free(s->ext.supportedgroups);
     OPENSSL_free(s->ext.peer_supportedgroups);
-#endif                          /* OPENSSL_NO_EC */
     sk_X509_EXTENSION_pop_free(s->ext.ocsp.exts, X509_EXTENSION_free);
 #ifndef OPENSSL_NO_OCSP
     sk_OCSP_RESPID_pop_free(s->ext.ocsp.ids, OCSP_RESPID_free);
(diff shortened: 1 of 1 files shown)


mattcaswell:
> Although I think the supportedgroups should be included into the NO_EC macro for a more nit solution.
>
> Edit: put supportedgroups in NO_EC marco in SSL_CTX and any other places.

`supportedgroups` can now contain groups that are not EC related, therefore they should not be inside NO_EC guards.

InfoHunter:
> supportedgroups can now contain groups that are not EC related, therefore they should not be inside NO_EC guards.

Hmmm, for what kind of usage? I think the value transmitted from peer should be used to indicate the groups used with EC only...

mattcaswell:
> Hmmm, for what kind of usage? I think the value transmitted from peer should be used to indicate the groups used with EC only...

That used to be the case in TLSv1.2 (when the extension was called `supportedcurves`). In TLSv1.3 they renamed it to `supportedgroups` and added the ability to specify DH groups.

InfoHunter:
Ahh, thanks for the explanation!

slontis:
Is this required for 111?

InfoHunter:
> Is this required for 111?

Yes, this affects 1.1.1 as well.

raja-ashok:
TLSv1.3 DH group exchange feature is present only on `master`. So this fix is not applicable to `1.1.1`.

mattcaswell:
> Yes, this affects 1.1.1 as well.

No, master only

slontis:
Thanks.. Merged to master.

InfoHunter:
> No, master only

Sure, I checked it again. This does only affect master

kaduk:
> > Hmmm, for what kind of usage? I think the value transmitted from peer should be used to indicate the groups used with EC only...

> That used to be the case in TLSv1.2 (when the extension was called supportedcurves). In TLSv1.3 they renamed it to supportedgroups and added the ability to specify DH groups.

Not that it is terribly important, but RFC 7919 also specified named (FFDH) groups for pre-TLS-1.3

PR #10195, opened 2019-10-16
Add X509_V_FLAG_NO_CHECK_SUBJECT and change default to be safe

dwmw2:
If no hostname/IP/email is added for verification, OpenSSL will currently
allow verification to succeed for *any* validly signed certificate,
regardless of its subject.

This behaviour isn't just a suboptimal default; in any real application
it is egregiously and insecurely wrong. If any application really wants
it, they should have to jump through hoops to *ask* for it.

Add X509_V_FLAG_NO_CHECK_SUBJECT to permit that behaviour, and change
the default to fail verification if no identity is set.

This would have prevented CVE-2019-14553
(https://bugzilla.tianocore.org/show_bug.cgi?id=960) and others. I got
this wrong in OpenConnect myself in the early days too, when I was
young and naïve and didn't *expect* that the crypto libraries were
all out to get me :)

This straw man commits breaks some tests; if we agree on the principle
then it should be simple enough to go through and fix them to either
set X509_V_FLAG_NO_CHECK_SUBJECT or provide the expected subject.

##### Checklist

### DIFF
--- a/crypto/x509/x509_vfy.c
+++ b/crypto/x509/x509_vfy.c
@@ -729,6 +729,7 @@ static int check_id(X509_STORE_CTX *ctx)
 {
     X509_VERIFY_PARAM *vpm = ctx->param;
     X509 *x = ctx->cert;
+
     if (vpm->hosts && check_hosts(x, vpm) <= 0) {
         if (!check_id_error(ctx, X509_V_ERR_HOSTNAME_MISMATCH))
             return 0;
@@ -741,7 +742,17 @@ static int check_id(X509_STORE_CTX *ctx)
         if (!check_id_error(ctx, X509_V_ERR_IP_ADDRESS_MISMATCH))
             return 0;
     }
-    return 1;
+
+    /*
+     * When verifying SSL server certificates, require that an
+     * identity be set for validating the certificate subject unless
+     * the X509_V_FLAG_ALLOW_NO_SUBJECT_CHECK flags is set instead.
+     */
+    if (vpm->purpose != X509_PURPOSE_SSL_SERVER
+        || (vpm->flags & X509_V_FLAG_ALLOW_NO_SUBJECT_CHECK))
+        return 1;
+
+    return (vpm->hosts || vpm->email || vpm->ip);
 }
 
 static int check_trust(X509_STORE_CTX *ctx, int num_untrusted)
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,13 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) Validation of SSL server certificates will no longer succeed by
+     default when no identity has been set with SSL_set1_host() (or
+     similar methods). Applications which manually verify the subject
+     of the server certificate can restore the original behaviour by
+     setting X509_V_FLAG_ALLOW_NO_SUBJECT_CHECK.
+     [David Woodhouse]
+
   *) Added functionality to create an EVP_PKEY context based on data
      for methods from providers.  This takes an algorithm name and a
      property query string and simply stores them, with the intent
--- a/apps/include/opt.h
+++ b/apps/include/opt.h
@@ -29,7 +29,7 @@
         OPT_V_SUITEB_128_ONLY, OPT_V_SUITEB_128, OPT_V_SUITEB_192, \
         OPT_V_PARTIAL_CHAIN, OPT_V_NO_ALT_CHAINS, OPT_V_NO_CHECK_TIME, \
         OPT_V_VERIFY_AUTH_LEVEL, OPT_V_ALLOW_PROXY_CERTS, \
-        OPT_V__LAST
+        OPT_V_ALLOW_NO_SUBJECT_CHECK, OPT_V__LAST
 
 # define OPT_V_OPTIONS \
         { "policy", OPT_V_POLICY, 's', "adds policy to the acceptable policy set"}, \
@@ -79,7 +79,8 @@
             "accept chains anchored by intermediate trust-store CAs"}, \
         { "no_alt_chains", OPT_V_NO_ALT_CHAINS, '-', "(deprecated)" }, \
         { "no_check_time", OPT_V_NO_CHECK_TIME, '-', "ignore certificate validity time" }, \
-        { "allow_proxy_certs", OPT_V_ALLOW_PROXY_CERTS, '-', "allow the use of proxy certificates" }
+        { "allow_proxy_certs", OPT_V_ALLOW_PROXY_CERTS, '-', "allow the use of proxy certificates" }, \
+        { "allow_no_subject_check", OPT_V_ALLOW_NO_SUBJECT_CHECK, '-', "do not require target identity to be provided" }
 
 # define OPT_V_CASES \
         OPT_V__FIRST: case OPT_V__LAST: break; \
@@ -112,7 +113,8 @@
         case OPT_V_PARTIAL_CHAIN: \
         case OPT_V_NO_ALT_CHAINS: \
         case OPT_V_NO_CHECK_TIME: \
-        case OPT_V_ALLOW_PROXY_CERTS
+        case OPT_V_ALLOW_PROXY_CERTS: \
+        case OPT_V_ALLOW_NO_SUBJECT_CHECK
 
 /*
  * Common "extended validation" options.
--- a/apps/lib/opt.c
+++ b/apps/lib/opt.c
@@ -608,6 +608,9 @@ int opt_verify(int opt, X509_VERIFY_PARAM *vpm)
     case OPT_V_ALLOW_PROXY_CERTS:
         X509_VERIFY_PARAM_set_flags(vpm, X509_V_FLAG_ALLOW_PROXY_CERTS);
         break;
+    case OPT_V_ALLOW_NO_SUBJECT_CHECK:
+        X509_VERIFY_PARAM_set_flags(vpm, X509_V_FLAG_ALLOW_NO_SUBJECT_CHECK);
+        break;
     }
     return 1;
 
--- a/doc/man3/X509_VERIFY_PARAM_set_flags.pod
+++ b/doc/man3/X509_VERIFY_PARAM_set_flags.pod
@@ -307,6 +307,17 @@ The B<X509_V_FLAG_NO_CHECK_TIME> flag suppresses checking the validity period
 of certificates and CRLs against the current time. If X509_VERIFY_PARAM_set_time()
 is used to specify a verification time, the check is not suppressed.
 
+The B<X509_V_FLAG_ALLOW_NO_SUBJECT_CHECK> flag allows verification of a
+certificate for the B<X509_PURPOSE_SSL_SERVER> purpose to succeed even if
+the application has not provided a DNS hostname, IP or email address to
+be checked against the Subject of the certificate. Since OpenSSL 3.0.0,
+validation of SSL server certificates will fail unless either a matching
+identity has been provided or this flag is set.
+This flag has no effect if an identity has been set; it only affects the
+default behaviour when no identity has been provided for checking the
+subject of the peer certificate.
+
+
 =head1 INHERITANCE FLAGS
 
 These flags specify how parameters are "inherited" from one structure to
--- a/include/openssl/x509_vfy.h
+++ b/include/openssl/x509_vfy.h
@@ -246,6 +246,8 @@ void X509_STORE_CTX_set_depth(X509_STORE_CTX *ctx, int depth);
 # define X509_V_FLAG_NO_ALT_CHAINS               0x100000
 /* Do not check certificate/CRL validity against current time */
 # define X509_V_FLAG_NO_CHECK_TIME               0x200000
+/* Do not require an identity for certificate subject matching */
+# define X509_V_FLAG_ALLOW_NO_SUBJECT_CHECK      0x400000
 
 # define X509_VP_FLAG_DEFAULT                    0x1
 # define X509_VP_FLAG_OVERWRITE                  0x2
--- a/test/recipes/25-test_verify.t
+++ b/test/recipes/25-test_verify.t
@@ -18,7 +18,7 @@ setup("test_verify");
 
 sub verify {
     my ($cert, $purpose, $trusted, $untrusted, @opts) = @_;
-    my @args = qw(openssl verify -auth_level 1 -purpose);
+    my @args = qw(openssl verify -auth_level 1 -allow_no_subject_check -purpose);
     my @path = qw(test certs);
     push(@args, "$purpose", @opts);
     for (@$trusted) { push(@args, "-trusted", srctop_file(@path, "$_.pem")) }
--- a/test/recipes/60-test_x509_store.t
+++ b/test/recipes/60-test_x509_store.t
@@ -27,7 +27,7 @@ plan skip_all => "test_rehash is not available on this platform"
 
 sub verify {
     my ($cert, $purpose, $trustedpath, $untrusted, @opts) = @_;
-    my @args = qw(openssl verify -auth_level 1 -purpose);
+    my @args = qw(openssl verify -auth_level 1 -allow_no_subject_check -purpose);
     my @path = qw(test certs);
     push(@args, "$purpose", @opts);
     push(@args, "-CApath", $trustedpath);
--- a/test/ssl_test.c
+++ b/test/ssl_test.c
@@ -482,6 +482,14 @@ static int test_handshake(int idx)
         && !SSL_CTX_config(resume_client_ctx, "resume-client"))
         goto err;
 
+    {
+        X509_VERIFY_PARAM *param = SSL_CTX_get0_param(client_ctx);
+        X509_VERIFY_PARAM_set_flags(param, X509_V_FLAG_ALLOW_NO_SUBJECT_CHECK);
+    }
+    if (resume_client_ctx != NULL) {
+        X509_VERIFY_PARAM *param = SSL_CTX_get0_param(resume_client_ctx);
+        X509_VERIFY_PARAM_set_flags(param, X509_V_FLAG_ALLOW_NO_SUBJECT_CHECK);
+    }
     result = do_handshake(server_ctx, server2_ctx, client_ctx,
                           resume_server_ctx, resume_client_ctx, test_ctx);
 
--- a/test/ssltest_old.c
+++ b/test/ssltest_old.c
@@ -1530,6 +1530,11 @@ int main(int argc, char *argv[])
     }
 
     {
+        X509_VERIFY_PARAM *param = SSL_CTX_get0_param(c_ctx);
+        X509_VERIFY_PARAM_set_flags(param, X509_V_FLAG_ALLOW_NO_SUBJECT_CHECK);
+    }
+
+    {
         int session_id_context = 0;
         if (!SSL_CTX_set_session_id_context(s_ctx, (void *)&session_id_context,
                                             sizeof(session_id_context)) ||


mattcaswell:
This seems sensible to me.

t8m:
Well, in case of client certificate authentication verification there is nothing to check against and the subject is supposed to be eventually checked by the application itself. And this is not that marginal use-case. Certainly there are apps where the subject verification of a certificate inside openssl does not make sense as it would be completely application specific. This change would be API break for them.
I am not saying we cannot do this API break in 3.0 however we should probably start a page that would list all the known API breaks in 3.0 so the application writers know about them in advance.

dwmw2:
Switched logic around to have `X509_V_FLAG_ALLOW_EMPTY_IDENT`, and to set that for the entries in `default_table[]` other than `"default"` and `"ssl_server"`.

Tested by hacking `demos/bio/sconnect.c` to "forget" to call `SSL_set1_host()`, and observing that it fails.

mattcaswell:
Travis reports various test failures which look relevant.

I wonder whether the logic in this is the wrong way around. I worry that this could be a major breaking change beyond libssl with some unintended consequences. A possible alternative route is to *switch on* subject checking by default in a libssl client - otherwise you don't get it.

dwmw2:
I just inverted and tweaked the logic to make it saner for client vs. other use cases. Will fix tests up once we've agreed on what the flag looks like.

I thought this way looks better. 

The capacity for 'major breaking change' is mostly that someone who does subject checks for themselves forgets to set the `X509_V_FLAG_ALLOW_EMPTY_IDENT` flag. Surely that would show up *really* early in testing as they upgrade to the new version of OpenSSL?

t8m:
ALLOW_EMPTY_IDENT is really misleading name for the flag - it is not about allowing certificates which have empty subject. ALLOW_NO_SUBJECT_CHECK or ALLOW_NO_IDENT_CHECK or something like that would be more appropriate.

However I am with Matt that this should be handled differently by making the ip/hostname check on by default in the libssl client code.

dwmw2:
Hm, let's make sure I understand correctly. As I see it, the IP/hostname check *is* on by default for the libssl client code. It's not done separately; it's done as part of the call to `X509_verify_cert()`.  If the application actually remembers to set a hostname, it all works correctly.

This PR is about changing the behaviour if the application author *forgets* to specify the hostname. Which they do. In that case, for SSL client code, we should fail safe. We shouldn't **default** to accepting *any* signed certificate regardless of its subject.

So, if we really wanted to pull that *right* out into the libssl code then I suppose we could. It could be something like this (more illustrative than something that would actually compile):
```
--- a/ssl/ssl_cert.c
+++ b/ssl/ssl_cert.c
@@ -423,9 +431,14 @@ int ssl_verify_cert_chain(SSL *s, STACK_OF(X509) *sk)
 
     if (s->ctx->app_verify_callback != NULL)
         i = s->ctx->app_verify_callback(ctx, s->ctx->app_verify_arg);
-    else
-        i = X509_verify_cert(ctx);
+    else {
+        if (!s->server && !param->hosts && !param->ip && !param->email /* ! */) {
+            SSLerr(SSL_F_SSL_VERIFY_CERT_CHAIN, ERR_R_NO_IDENTITY);
+            goto end;
+        }
 
+        i = X509_verify_cert(ctx);
+    }
     s->verify_result = X509_STORE_CTX_get_error(ctx);
     sk_X509_pop_free(s->verified_chain, X509_free);
     s->verified_chain = NULL;
```
Of course, if they do provide their own `app_verify_callback` which eventually calls `X509_verify_cert()`, they still don't get the benefit of default fail-safe behaviour.

I was definitely leaning towards having the actual implementation *in* `X509_verify_cert()`, controlled by one of the already numerous `X509_V_FLAG_…` flags. Once we've finished bikeshedding the name and semantics of the flag, it becomes a one-liner in the `check_id()` function, as in my original patches. And OK, let's take Matt's words literally, that the libssl client code should "switch on" the check... by setting (or clearing, whatever) that flag? What might that look like?

Again illustrative; we probably wouldn't put it right here, but I really wanted that previous line to show in the context of the "patch":
```
--- a/ssl/ssl_cert.c
+++ b/ssl/ssl_cert.c
@@ -413,6 +413,9 @@ int ssl_verify_cert_chain(SSL *s, STACK_OF(X509) *sk)
      */
 
     X509_STORE_CTX_set_default(ctx, s->server ? "ssl_client" : "ssl_server");
+    if (!s->server)
+        X509_VERIFY_PARAM_set_flags(param, X509_V_FLAG_REQUIRE_IDENT);
+

```

Having already requested the defaults for an SSL client, we then manually add the extra flag that SSL clients should use? Isn't that what `X509_STORE_CTX_set_default()` is *for*?

Ultimately, while I feel quite strongly that the default *should* be changed to protect users from themselves, I don't really care much how we do it. But I'm not sure I quite understand what I'm being asked here, so please help me and then I'll set about implementing it however you prefer.

mattcaswell:
I was thinking more along the lines of having a `X509_V_FLAG_DISALLOW_EMPTY_SUBJECT` and then changing the default for "ssl_server" (only) to include it.

dwmw2:
Ack. May rename to `REQUIRE_SUBJECT_MATCH` to accommodate @t8m's concerns above. Will post new code in a while.

t8m:
I think in this form the API break should be acceptable for 3.0.

dwmw2:
To be clear to those who haven't dug into the commits,  *"this form"* at the time of the above comment looked like this:

```
--- a/doc/man3/X509_VERIFY_PARAM_set_flags.pod
+++ b/doc/man3/X509_VERIFY_PARAM_set_flags.pod
@@ -307,6 +307,19 @@ The B<X509_V_FLAG_NO_CHECK_TIME> flag suppresses checking the validity period
 of certificates and CRLs against the current time. If X509_VERIFY_PARAM_set_time()
 is used to specify a verification time, the check is not suppressed.
 
+The B<X509_V_FLAG_REQUIRE_SUBJECT_MATCH> flag enforces that verification
+will not succeed unless a DNS hostname, IP or email address has been set
+which matches the subject of the certificate. Without this flag, if no
+identity is set then validation will succeed for any validly signed
+certificate regardless of its subject.
+This flag was introduced in OpenSSL 3.0.0, and is included in the
+"ssl_server" default verification parameters returned by the
+X509_VERIFY_PARAM_lookup() function, used for SSL client connections.
+Applications which perform subject checks manually, and wish to preserve
+the historical behaviour of accepting all signed certificates, must
+explicitly clear this flag.


```

mattcaswell:
Travis failures are relevant. I suspect we have *lots* of places in the tests where we don't set this up (because its a test). My guess is this will also break third party test suites....and probably lots of apps that deserve to be broken by this because they need to be fixed. There should be a CHANGES entry describing the change, and what people need to do to fix errors encountered because of this.

dwmw2:
Working on test failures now.

dwmw2:
Hm, the problem with this "polarity" for the flag — calling it 'REQUIRE' rather than 'NO_REQUIRE' — is that the *absence* of the flag isn't inherited.

There's nothing I can set in the SSL's `s->param` which is then inherited into the param of the `X509_STORE_CTX` when `ssl_verify_cert_chain()` creates it, and *clears* the flag therein.

This worked when the meaning of the flag was inverted, before @mattcaswell's comment above made me change it. I think changing it back to `ALLOW_NO_SUBJECT_CHECK` and setting it in all the profiles except `"default"` and `"ssl_server"` makes most sense. Matt?

Or maybe we don't set it in *any* default params, and actually we trigger based on the purpose?

```
--- a/crypto/x509/x509_vfy.c
+++ b/crypto/x509/x509_vfy.c
@@ -729,6 +729,7 @@ static int check_id(X509_STORE_CTX *ctx)
 {
     X509_VERIFY_PARAM *vpm = ctx->param;
     X509 *x = ctx->cert;
+
     if (vpm->hosts && check_hosts(x, vpm) <= 0) {
         if (!check_id_error(ctx, X509_V_ERR_HOSTNAME_MISMATCH))
             return 0;
@@ -741,7 +742,13 @@ static int check_id(X509_STORE_CTX *ctx)
         if (!check_id_error(ctx, X509_V_ERR_IP_ADDRESS_MISMATCH))
             return 0;
     }
-    return 1;
+
+    if (vpm->purpose == X509_PURPOSE_SSL_SERVER &&
+       !(vpm->flags & X509_V_FLAG_ALLOW_NO_SUBJECT_CHECK))
+        return 1;
+
+    /* If nothing matched, that is a failure unless checking was bypassed */
+    return (vpm->hosts || vpm->email || vpm->ip);
 }
 
 static int check_trust(X509_STORE_CTX *ctx, int num_untrusted)

```

dwmw2:
Here's that version, with tests passing locally.

crypto/x509/x509_vfy.c
```diff
@@ -741,7 +742,13 @@ static int check_id(X509_STORE_CTX *ctx)
         if (!check_id_error(ctx, X509_V_ERR_IP_ADDRESS_MISMATCH))
             return 0;
     }
-    return 1;
+
+    if (vpm->purpose != X509_PURPOSE_SSL_SERVER ||
+	(vpm->flags & X509_V_FLAG_ALLOW_NO_SUBJECT_CHECK))
```
mattcaswell: Indentation appears off here

dwmw2:
+`CHANGES` note:
```
	  *) Validation of SSL server certificates will no longer succeed by
     default when no identity has been set with SSL_set1_host() (or
     similar methods). Applications which manually verify the subject
     of the server certificate can restore the original behaviour by
     setting X509_V_FLAG_ALLOW_NO_SUBJECT_CHECK.
     [David Woodhouse]
```

mattcaswell (changes requested):
LGTM. I have just one minor nit below.

crypto/x509/x509_vfy.c
richsalz: put the ``||`` on the second line and indent that line an extra four.

crypto/x509/x509_vfy.c
```diff
@@ -741,7 +742,17 @@ static int check_id(X509_STORE_CTX *ctx)
...
             return 0;
     }
-    return 1;
+
+    /*
+     * When verifying SSL server certificates, require that an
+     * identity be set for validating the certificate subject unless
+     * the X509_V_FLAG_ALLOW_NO_SUBJECT_CHECK flags is set instead.
+     */
+    if (vpm->purpose != X509_PURPOSE_SSL_SERVER
```
vdukhovni: Instead of `vpm->purpose != X509_PURPOSE_SSL_SERVER` it would have to be `vpm->purpose == X509_PURPOSE_SSL_CLIENT`.  Otherwise, this breaks `X509_PURPOSE_SMIME_SIGN` and possibly others.

vdukhovni (commented):
This PR breaks application-level subject checks, or explicit lack thereof. It therefore requires code changes in the application SSL client code. It is my understanding that we are also aiming to provide (whenever possible) source-level compatibility in 3.0 (internal restructuring + FIPS), with avoidable breaking API changes in later releases.
Has anyone performed any application audits to see which use OpenSSL's built-in subject checks, and which do their own? (FWIW, Postfix performs its own).
So I think this change would have to wait till 4.0.
[ And of course this PR cannot be backported in any 1.1.1 or 1.0.2 patch updates. ]

crypto/x509/x509_vfy.c
dwmw2: When we are a client, the purpose of the certificate we are checking is "server".

I learned this the fun way, of course...

vdukhovni:
Another datapoint, I don't see any calls to `SSL_set1_host` or `X509_VERIFY_PARAM_set1_host` or similar in curl https://github.com/curl/curl/blob/master/lib/vtls/openssl.c

dwmw2:
Yes, this breaks applications like that. Deliberately so. A fast and obvious break, easily fixed by setting a single flag.

Those application authors clueful enough to implement subject checking — correctly — can probably cope.

The change protects naïve users from shooting themselves in the foot by not checking at all.

vdukhovni:
> Yes, this breaks applications like that. Deliberately so. A fast and obvious break, easily fixed by setting a single flag.
> 
> The change protects naïve users from shooting themselves in the foot by not checking at all.

All true, but it was my understanding that we're only doing unavoidable API changes in 3.0, and expecting most applications to not need source changes. Perhaps I've not got the nuance quite right...

vdukhovni:
Github advanced search shows 184k hits (in C code) for `SSL_CTX_new`, but only 28k hits for `X509_VERIFY_PARAM_set1_host` and 10k hits for `SSL_set1_host`.

dwmw2:
What ratio are we prepared to accept of application authors immediately noticing that they have to set a single flag when they update, to recurrences of CVE-2019-14553? :)

kroeckx:
What about assuming that they check the hostname/ip/email when they set up the callback?

vdukhovni:
> What about assuming that they check the hostname/ip/email when they set up the callback?

The verify callback is not necessarily the right place to do the hostname checks. It is called for more than just the leaf certificate, and most often used for logging/debugging the chain, or to accept some errors. It is not a very convenient place to put the application hostname checks, it is easier to do that at the completion of the handshake. So applications that have a verification callback don't necessarily check the hostname there...

The key question here is whether the need to add the flag is an acceptable barrier to adoption for 3.0, for what I suspect would be a significant fraction of applications. Obviously I am positioned to do that for Postfix if the change is merged here, but I am concerned about introducing a barrier that applies broadly, rather than in a small number of cases. We can discuss this next week at the f2f, unless I am already a lone voice on this, and my memory of how we're positioning 3.0 compatibility is out of sync of where we are now...

dwmw2:
The app callback set by `SSL_CTX_set_cert_verify_callback()` is a one-off, not per-cert like the one set by `SSL_set_verify()`.

But I can raise my hand as an application author who forgot to do it there too, and would have benefited from the proposed failsafe in `X509_verify_cert()`.

And if I'm supposed to check the name constraints all the way down from the root CA I'm probably still getting it wrong.

vdukhovni:
> The app callback set by `SSL_CTX_set_cert_verify_callback()` is a one-off, not per-cert like the one set by `SSL_set_verify()`.

True, is also rarely used (though I am one of its "rare" users). One way to use it safely, is to tweak the chain to be checked, and then call into the usual X509_verify_cert(), that way you don't end reinventing the rather complex machinery there...

> But I can raise my hand as an application author who forgot to do it there too, and would have benefited from the proposed failsafe in `X509_verify_cert()`.

Once you're using the manifestly advanced features of the one shot app verify callback, you really are obligated to carefully study the code you're bypassing.  I read the OpenSSL source multiple times before braving to use that feature.

> And if I'm supposed to check the name constraints all the way down from the root CA I'm probably still getting it wrong.

FWIW, the name constraints are built-in, whether you validate the names or not.  Anyway I understand and sympathise with your motivation, my objection is only about whether 3.0 is a valid opportunity to make this change. It was my impression that 3.0 was supposed to just require re-compilation for "most" applications, without code changes.

dwmw2:
If 3.0.0 being 100% compatible is indeed the intent *(and if we were really going to achieve that in practice)*, then I'll concede there's no desperate rush for this default to be changed.

After all, I should probably have thrown my toys out of the pram a decade ago when I first made this mistake for myself, and got it fixed then — which would have avoided CVE-2019-14553 in the first place. Another release isn't going to kill anyone. I'll let you decide that at the f2f as you suggest.

vdukhovni:
My impression, which may be in error, was that the aim is compatibility for *most* applications. So not 100% compatible, but rather an expectation of compatibility in the typical case, with code changes required only for more unusual applications. Now if that's not the goal, or we decide to make an exception for this PR (accept the security-by-default/compatibility trade-off) then this PR could land in 3.0.

paulidale:
An alternative to consider: insist that `SSL_set1_host()` is called either with a hostname or NULL for no checking.  If `SSL_set1_host()` isn't called, return an error _must call SSL_set1_host_.

t-j-h:
This PR should also fix the issue that BIO_set_conn_hostname also flows things down into SSL_set1_host.

t-j-h:
Note hold label added on behalf of @vdukhovni

kaduk:
Did this end up getting discussed at the f2f?

paulidale:
Yes it was discussed in great depth but to no good result.  We couldn't think of a way to address this that doesn't break a significant portion of our users but it is a security fix and is thus very desirable.

We do need a capability to allow fixing this kind of problem using configuration as opposed to code changes and this will likely be forthcoming.

mattcaswell:
IIRC the action is with @vdukhovni to propose a route ahead.

petrovr:
Modification in crypto/x509/x509_vfy.c is generic and definitely impacts third party software.
Note that function X509_verify_cert could be used out SSL/TLS context.
New(changed) functionality is acceptable only if is explicitly requested.

dwmw2:
> New(changed) functionality is acceptable only if is explicitly requested

Hm? Did you mean to quality this statement with "in OpenSSL 3.0.0"?

Imposing the functionality by default and making it opt-out was kind of the point.

petrovr:
Hi,

David Woodhouse wrote:
>> New(changed) functionality is acceptable only if is explicitly requested
> 
> Hm? Did you mean to quality this statement with "in OpenSSL 3.0.0"?

Version does not make sense. Context is the key. Modification is in 
crypto/x509/x509_vfy and so is common for all. Even for those who do not 
use network ssl connection.

> 
> Imposing the functionality by default and making it opt-out was kind of the point.
> 

No problem to be default but in context of network use.

After reading again modification may I ask you to confirm that verify 
command is not impacted?

Roumen

vdukhovni:
Questions of whether this affects X.509 below the SSL layer are rather moot at the moment, the PR is unlikely to move forward as-is. We'll need to try to find a more conservative approach in which the opt-in/opt-out choice is configurable (per application), though even that is not sufficient, because applications may consume multiple SSL-enabled libraries, with differing requirements.

kroeckx:
So I think there are 3 types of applications related to this PR:
- They use the new API that makes OpenSSL do the check
- They use the old API, and do the check themself
- They don't do the check

The problem is that we want to find the 3rd, so they can get
fixed. But the only solutions we seems to find are:
- Break both the 2nd and 3rd type
- Create a warning, for instance deprecating some interface, which
  will most likely affect all SSL and/or X509 users, and some will
  turn that warning into an error.

Maybe we should somehow encourage people to switch to the new API
without actually breaking something. The only option I currently
see for that is to file bugs against all the open source software
packages that don't use the new API.

We could also try to identify all open source packages that fall
in the 3rd type, which results in filing less bugs, but then we'd
consider that security bugs, which probably actually takes more
time to properly report to the various projects.

Both suggestions are only about open source software, because
that's what we can check, and I can't think of a good way to get
non-open source software fixed.

paulidale:
A link time warning if the old API isn't accessed would be ideal but I don't think it's possible to do that.

richsalz:
Related work happening at the IETF: https://tools.ietf.org/html/draft-ietf-uta-use-san-00  An update to RFC 6125. I suggest that when someone picks this up, they look at accommodating the draft (which will be a standards-track RFC at some point).

mattcaswell:
OMC: Referring this to OTC

t8m:
**Summary for the OTC decision:**

The current behavior of not verifying the identity of the server (i.e. whether the IP or hostname of the server matches the data in the certificate) by default (that is unless the application calls one of the functions that set the IP or hostname of the server to be validated) can lead to a serious vulnerability of TLS client applications that do not call that function by mistake or inadvertent omission of the call.

This proposal wants to change the default behavior to indicate verification error in such situation unless the `X509_V_FLAG_ALLOW_NO_SUBJECT_CHECK` is set in the verification parameters.

This is effectively an API break, however it would prevent security issues in applications. On the other hand this change as proposed would not break just TLS client applications but for example also standalone certificate verifications (i.e. as in the `openssl verify` command or an application that uses the standalone certificate verification as part of its custom SSL certificate verification callback which performs the hostname check by itself.

**Which release this would be acceptable for, if any?**

petrovr:
Hello,

Tomáš Mráz wrote:
> **Summary for the OTC decision:**
> 
> The current behavior of not verifying the identity of the server (i.e. whether the IP or hostname of the server matches the data in the certificate) by default (that is unless the application calls one of the functions that set the IP or hostname of the server to be validated) can lead to a serious vulnerability of TLS client applications that do not call that function by mistake or inadvertent omission of the call.
> 
> This proposal wants to change the default behavior to indicate verification error in such situation unless the `X509_V_FLAG_ALLOW_NO_SUBJECT_CHECK` is set in the verification parameters.
> 
> This is effectively an API break, however it would prevent security issues in applications. On the other hand this change as proposed would not break just TLS client applications but for example also standalone certificate verifications (i.e. as in the `openssl verify` command or an application that uses the standalone certificate verification as part of its custom SSL certificate verification callback which performs the hostname check by itself.
> 
> **Which release this would be acceptable for, if any?**

All above mean changes in common functionality. The problem is that such checks are protocol specific.
It seems to me only for TLS code should activate such checks.
With other words there is no version that should do TLS specific checks by default in common method(functionality).

It is not about only for hostname. If I remember well SSL and/or TLS recommend some other checks that are not acceptable in common functionality.

Regards,
Roumen Petrov

mattcaswell:
> All above mean changes in common functionality. The problem is that such checks are protocol specific.
It seems to me only for TLS code should activate such checks.
With other words there is no version that should do TLS specific checks by default in common method(functionality).

My understanding of the change in this PR is that it only applies if the purpose is explicitly set to `X509_PURPOSE_SSL_SERVER`, otherwise it does nothing. So this should have no impact outside of a TLS context.

t8m:
> So this should have no impact outside of a TLS context.

It would have an impact for things like `openssl verify` command if you set the ssl server purpose. But yeah that is a pretty special context.

mattcaswell:
> It would have an impact for things like openssl verify command if you set the ssl server purpose. But yeah that is a pretty special context.

Well...yes...but if you are explicitly setting the purpose to ssl server then presumably this kind of check seems reasonable.

petrovr:
Tomáš Mráz wrote:
>> So this should have no impact outside of a TLS context.
> 
> It would have an impact for things like `openssl verify` command if you set the ssl server purpose. But yeah that is a pretty special context.
> 

"ssl server purpose" does not mean that verification is used TLS/SSL protocol.

Logic is reverse TLS/SSL protocol should perform verification according RFC A,B,C. SSH should perform verification RFC D,E. What about other non protocol related uses.


Regards,
Roumen Petrov

mattcaswell:
> "ssl server purpose" does not mean that verification is used TLS/SSL protocol.

That statement makes no sense to me. Setting "ssl server purpose" means the purpose of this verification is for verifying an SSL/TLS server certificate. Rules applicable for an SSL/TLS server should be applied.

paulidale:
OTC: consider this for 4.0

Similar item discussed in #15216

vdukhovni:
> OTC: consider this for 4.0
> 
> Similar item discussed in #15216

This is a major problem if 3.0 becomes an LTS release.  It requires the applications that want the new feature to conditionally compile to a different API, and would break existing applications.  I rather strongly suggest this needs to be fixed in 3.0, sooner rather than later, before too many applications start relying on this (probably none so far, or only a few early adopters still able to change their implementations).

vdukhovni:
[ Sorry, commented on wrong topic, comment deleted ].

PR #10207, opened 2019-10-17
Replace BUF_ string functions with OPENSSL_ ones

richsalz:
Deprecate the BUF_ string macros


levitte:
Merged

3d48457478bd61030c370e4090c1462fc4453d81 Replace BUF_ string function calls with OPENSSL_ ones

mattcaswell:
Please can we not push deprecation changes through quite so quickly? Any deprecation does have implications for users and we should have careful consideration about each instance. As it happens I do agree with this case, but I would have liked the opportunity to consider it before it was pushed. This PR was raised, reviewed and merged before I had a chance to even look at it!

PR #10208, opened 2019-10-17
Cleanup man7/ssl

richsalz:
There were two paragraphs of useful information about SSL_dup, so
copy that to the right manpage.

It's funny, that page said "there are 240 functions"  There are now 580.  Just delete information that is woefully out of date and will never get updated.

I'm sure this will get rejected.  That would be wrong, but I'm okay with it. :)


doc/man7/ssl.pod
```diff
@@ -107,679 +107,6 @@ it's already included by ssl.h>.
 
 =back
 
-=head1 API FUNCTIONS
-
-Currently the OpenSSL B<ssl> library exports 214 API functions.
-They are documented in the following:
-
-=head2 Dealing with Protocol Methods
```
levitte: While I agree with the removal of inaccurate information, or for that matter a "complete" run-down of the API in a man7 page, I do not agree with  removing the guidance completely.  Of course, there's that SEE ALSO list, but quite frankly, with that Big Wall of Links, my personal reaction would be to thrown my hands in the air and admit defeat...

So I propose some kind of guidance still, maybe roughly like this:

``` pod
=head1 TOPICS

=head2 Dealing with Protocol Methods

There are a few TLS protocols to choose from, do see L<SSL_...(3)> and L<SSL_...(3)> as a starting point.

=head2 Dealing with Ciphers

.... yada yada a couple of links to starting points yada yada...

=head2 Dealing with Protocol Contexts

... et ...

=head2 Dealing with Sessions

... cetera ...

=head2 Dealing with Connections

...

```

How does that sound?

richsalz: > So I propose some kind of guidance still, maybe roughly like this:

I think that sounds great, if someone else wants to do it.  I just wanted to get the outdated stuff gone.

levitte: I think I'll wait for a PR that doesn't slash and leave the user helpless, but replaces it with something sensible.  Thank you.

richsalz: We differ on whether the current page is helpful or not. :)  I'll wait a bit and see if anyone else on the project likes this before closing it.

levitte: If this page isn't useful at all, why not simply removing it? So I guess that my question, do you think that a guide in this form may be useful at all or not? If you think it is, why not make it so?

richsalz:
I removed man7/ssl.pod and ssl(7) links.  Still not heavily invested, if you disagree and want to close this. :)

richsalz:
Rebased to remove the conflict and pushed.

levitte:
I'm undecided re removal, so er, if someone else can make the decision, that would be a relief.

doc/man7/ssl.pod
```diff
@@ -1,883 +0,0 @@
...
-I<Usually you don't have to include it explicitly because
-it's already included by ssl.h>.
-
-=item B<tls1.h>
-
-This is the sub header file dealing with the TLSv1 protocol only.
-I<Usually you don't have to include it explicitly because
-it's already included by ssl.h>.
-
-=back
```
mattcaswell: Aside from the opening sentence of the "DESCRIPTION" section which needs updating, I think everything above this point in the file still seems to hold some value - so should be kept (and probably extended or rewritten to be made better at some point (not this PR)). Everything below this point is fairly useless and should be removed.

richsalz: I think that keeping just the top part is a reasonable compromise.  Since the function declarations are removed, do you have any thoughts on removing the SEE ALSO links that I did?

richsalz:
additional commit pushed to restore a minimal man7/pod.ssl  I also rewrite some of the text there.

doc/man7/ssl.pod
mattcaswell: I think pointing at the overview page from other SSL related pages seems like a reasonable thing to do, so I would keep the SEE ALSO links.

richsalz: I want to push back on that a bit.  There are roughly 135 SSL manpages but only 100 of them link to ssl(7).  I think I should either add the missing links or keep them as currently removed.
```
; ls SSL*|wc
    135     135    3511
; grep 'ssl(7)' SSL* | wc
    105     186    5495
```

(Yes I know that's counting files, not those with multiple API's in them, which is why I said "roughly")

If we just look for '(7)' among all manpages the ratio is still about one-third:
```
; g grep '(7)' | wc    
    305     694   15384
; ls | wc
    496     496   10820
```

So add ssl(7) to all SSL pages, keep it with the current partial, or remove ssl(7) altogether?

mattcaswell: I don't particularly feel the need to add `ssl(7)` SEE ALSO links to the remaining SSL man pages.  But I certainly wouldn't object if you did it.

richsalz: What the L it was only 30 files, so I added it. :)

(Get the joke?)

richsalz:
Feedback from @mattcaswell and @levitte addressed.

t8m:
Merged to master as 
16f8a61
98ca37e 
ccd9e70

PR #10196, opened 2019-10-16
Unify BN_rshift design

cpereida:
This PR aims at refactoring the `BN_rshift` by making it a wrapper around `bn_rshift_fixed_top`, in order to match the current design of `BN_lshift`, as suggested in the discussion at https://github.com/openssl/openssl/pull/10122#discussion_r332474277 .

As described in the code, by refactoring this function, `BN_rshift` provides a constant-time behavior for sufficiently[!] zero-padded inputs under the following assumptions: `|n < BN_BITS2|` or `|n / BN_BITS2|` being non-secret.

Notice that `BN_rshift` returns a canonical representation of the BIGNUM, if a `fixed_top` representation is required, the caller should call `bn_rshift_fixed_top` instead.

## Benchmarks
To validate the impact of this change, we ran `openssl speed` across different cryptosystems following different code paths. These are the results:

### Current `BN_rshift` (a397aca43598ef20c84e69f6d6e5d95652aa0325)
```
~/openssl$ ./util/shlib_wrap.sh ./apps/openssl speed dsa2048 rsa2048 ecdsap224 ecdsap256 ecdsap384 ecdsab283
Doing 2048 bits private rsa's for 10s: 16617 2048 bits private RSA's in 10.00s
Doing 2048 bits public rsa's for 10s: 562429 2048 bits public RSA's in 10.00s
Doing 2048 bits sign dsa's for 10s: 37590 2048 bits DSA signs in 9.99s
Doing 2048 bits verify dsa's for 10s: 44473 2048 bits DSA verify in 10.00s
Doing 224 bits sign ecdsa's for 10s: 174958 224 bits ECDSA signs in 9.94s
Doing 224 bits verify ecdsa's for 10s: 86694 224 bits ECDSA verify in 10.00s
Doing 256 bits sign ecdsa's for 10s: 336600 256 bits ECDSA signs in 9.89s
Doing 256 bits verify ecdsa's for 10s: 146099 256 bits ECDSA verify in 10.00s
Doing 384 bits sign ecdsa's for 10s: 9701 384 bits ECDSA signs in 9.99s
Doing 384 bits verify ecdsa's for 10s: 13086 384 bits ECDSA verify in 10.00s
Doing 283 bits sign ecdsa's for 10s: 17822 283 bits ECDSA signs in 9.98s
Doing 283 bits verify ecdsa's for 10s: 8963 283 bits ECDSA verify in 9.98s
version: 3.0.0-dev
[...]

PR #10122, opened 2019-10-08
Constant-time GCD implementation and testing

cpereida:
Hey OpenSSL team,

Following up our recent [work](https://arxiv.org/abs/1909.01785) highlighting issues with the current  implementation of `BN_gcd` in OpenSSL, we have decided to contribute with a constant-time version of `BN_gcd`.
This PR replaces the current `BN_gcd` function by a constant-time implementation based on the work by [Bernstein and Yang](https://eprint.iacr.org/2019/266). Also, we add testing infrastructure for the `BN_gcd` function which was missing in the code base.
 
It is worth mentioning that currently the GCD function has limited use to only some RSA code paths and the primality testing function, thus the possible impact of replacing it for the constant-time version is really low. We include some benchmarks using the OpenSSL speed command for RSA, which might be relevant.

#### Current GCD speed results
 ```
 ~/openssl$ taskset -c 3 ./apps/openssl speed rsa
 Doing 512 bits private rsa's for 10s: 242294 512 bits private RSA's in 10.00s
 Doing 512 bits public rsa's for 10s: 4155401 512 bits public RSA's in 10.00s
 Doing 1024 bits private rsa's for 10s: 114724 1024 bits private RSA's in 10.00s
 Doing 1024 bits public rsa's for 10s: 1748093 1024 bits public RSA's in 9.99s
 Doing 2048 bits private rsa's for 10s: 16856 2048 bits private RSA's in 10.00s
 Doing 2048 bits public rsa's for 10s: 571001 2048 bits public RSA's in 10.00s
 Doing 3072 bits private rsa's for 10s: 5575 3072 bits private RSA's in 10.00s
 Doing 3072 bits public rsa's for 10s: 274365 3072 bits public RSA's in 10.00s
 Doing 4096 bits private rsa's for 10s: 2442 4096 bits private RSA's in 10.01s
 Doing 4096 bits public rsa's for 10s: 160917 4096 bits public RSA's in 10.00s
 Doing 7680 bits private rsa's for 10s: 265 7680 bits private RSA's in 10.04s
 Doing 7680 bits public rsa's for 10s: 47779 7680 bits public RSA's in 10.00s
 Doing 15360 bits private rsa's for 10s: 54 15360 bits private RSA's in 10.15s
 Doing 15360 bits public rsa's for 10s: 12421 15360 bits public RSA's in 10.00s
 version: 3.0.0-dev
 built on: built on: Tue Oct  8 08:30:34 2019 UTC
[...]

PR #10209, opened 2019-10-17
[crypto/bn] fix a few small timing leaks in BN_lshift1 and BN_rshift1

bbbrumley:
Lots of conditional stuff going on here, so straightline some of it and make the memory access more regular.

If any performance impact, I would bet money on very slight positive.

Relates to #10122 and #10196.

Tagging @romen @mattcaswell.


mattcaswell:

PR #10214, opened 2019-10-18
Fix some wording and markup

richsalz:


doc/man3/OpenSSL_version.pod
```diff
@@ -102,46 +102,47 @@ The value of B<OPENSSL_FULL_VERSION_STR>
...
 The value has the same syntax as the environment variable.
-For x86 the string looks like "CPUINFO: OPENSSL_ia32cap=0x123:0x456".
-Or "CPUINFO: N/A" if not available, e.g. no-asm build.
+For x86 the string looks like C<CPUINFO: OPENSSL_ia32cap=0x123:0x456>.
+Or C<CPUINFO: N/A> if not available, such as when configured with B<no-asm>.
 
 =back
 
-For an unknown B<t>, the text "not available" is returned.
+For an unknown B<t>, the text C<not available> is returned.
```
t8m: `I<t>`

doc/man3/OpenSSL_version.pod
```diff
@@ -102,46 +102,47 @@ The value of B<OPENSSL_FULL_VERSION_STR>
...
 
 OPENSSL_info() also returns different strings depending on B<t>:
```
t8m: And here as well.

doc/man3/OpenSSL_version.pod
richsalz: are you sure about that?  it's a parameter

richsalz:
made a few other fixes (missed some quotes -> C markup) fixup comment pushed.

doc/man3/OpenSSL_version.pod
t8m: Yes, parameters should be in italics. https://linux.die.net/man/7/man-pages

richsalz: Okay, thanks for the education. There were actually four places to fix.  Fixup commit pushed.

doc/man3/OpenSSL_version.pod
```diff
@@ -52,22 +52,22 @@ Deprecated:
...
-The macro B<OPENSSL_VERSION_BUILD_METADATA> is extra metadata, reserved
-for other parties (examples: C<"+fips">, C<"+vendor.1">).
+The macro B<OPENSSL_VERSION_BUILD_METADATA> is extra information, reserved
+for other parties, such as C<"+fips">, or C<"+vendor.1">.
 The OpenSSL project will not touch this macro.
 The value must be a string.
 
 B<OPENSSL_VERSION_STR> is a convenience macro to get the short version
-number string, "MAJOR.MINOR.PATCH".
+number string, C<MAJOR.MINOR.PATCH>.
```
levitte: I'm ok with the markup, but the quotes should stay because it really is a string.

richsalz: @levitte and @t8m please come to some agreement and tell me what to do.

My preference is that since it is output, notably *literal* output, it should be in ```C<>``` markup.  But the two of you decide.

levitte: What part of "I'm ok with the markup" got lost here?  What I'm proposing is `C<"MAJOR.MINOR.PATCH">`...  although `C<"I<MAJOR>.I<MINOR>.I<PATCH>">` would probably be more accurate.

Why do I propose this change?  Because `OPENSSL_VERSION_STR` *is* literally defined to a string, i.e.:

``` C
#define OPENSSL_VERSION_STR "3.0.0"
```

not:

``` C
#define OPENSSL_VERSION_STR 3.0.0
```

Do you see the difference?

richsalz: That helps, but still not explicit enough for me, sorry.  Do you want *all* of them changed, or just that one?  For example, at line 104:
```
The compiler flags set for the compilation process in the form
"compiler: ..."  if available or "compiler: information not available"
 otherwise.
```

Add ```C<>``` around the quotes or, as I currently have it, use ```C<>``` instead of the quotes?

levitte: Line 104 isn't about a macro definition, it's the returned value from `OpenSSL_version`, which *is* a string, but there are no literal quote in that string, and I could frankly go either way with those, depending on if you want express that it's a string (quotes included) or just display the contents of that string (quotes excluded).

richsalz: I changed the one instance at line 70.  I also changed a couple "version number" to "version identifier"  Fixup commit pushed.

doc/man3/OpenSSL_version.pod
```diff
@@ -52,25 +52,25 @@ Deprecated:
 =head2 Macros
 
 The three macros B<OPENSSL_VERSION_MAJOR>, B<OPENSSL_VERSION_MINOR> and
-B<OPENSSL_VERSION_PATCH> represent the three parts of a 3 numbered version
-number, MAJOR.MINOR.PATCH.
+B<OPENSSL_VERSION_PATCH> represent the three parts of a version
+identifier, B<MAJOR.MINOR.PATCH>.
```
levitte: In the interest of consistency, this should be `B<I<MAJOR>.I<MINOR>.I<PATCH>>`

levitte:
Fix that last and I'll approve

doc/man3/OpenSSL_version.pod
richsalz: I wondered if that was going to get called out :)  Fixup commit pushed.

t8m:
Still approved by me.

richsalz:
And likewise, I appreciate your patience with me when I closed a bunch of PR's in a huff.

t8m:
Merged to master as 3c77a41b30

PR #10215, opened 2019-10-18
Fix an s_server arbitrary file read issue on Windows

mattcaswell:
Running s_server in WWW mode on Windows can allow a client to read files
outside the s_server directory by including backslashes in the name, e.g.

GET /..\myfile.txt HTTP/1.0

There exists a check for this for Unix paths but it is not sufficient
for Windows.

Since s_server is a test tool no CVE is assigned.

Thanks to Jobert Abma for reporting this.


mattcaswell:
Pushed. Thanks.

PR #10216, opened 2019-10-18
Document "get/set-app-data" macros.

richsalz:
Documenting the macros removes 14 undocumented items.
Merge the two SSL exdata/appdata manpages into one.
Rename BIO exdata page to be consistent.

This should backport trivially.


richsalz:
find-doc-nits complains, but weirdly:
```
doc/man3/SSL_CTX_set_ex_data.pod:1: SSL_CTX_set_app_data is not public
doc/man3/SSL_CTX_set_ex_data.pod:1: SSL_get_app_data is not public
doc/man3/SSL_CTX_set_ex_data.pod:1: SSL_SESSION_get_app_data is not public
doc/man3/SSL_CTX_set_ex_data.pod:1: SSL_CTX_get_app_data is not public
doc/man3/SSL_CTX_set_ex_data.pod:1: SSL_SESSION_set_app_data is not public
doc/man3/SSL_CTX_set_ex_data.pod:1: SSL_set_app_data is not public
```

NOT that they're out of order :) but why didn't it complain about the other appdata additions?

richsalz:
It didn't complain because those functions were listed EXPLICITLY, and not in a "for openssl generic" manpage.  I think find-doc-nits scans the synopsis, not the NAME section.  Oh well, at least I know why.  New commit pushed.

crypto/rand/drbg_lib.c
```diff
@@ -503,7 +503,7 @@ void RAND_DRBG_free(RAND_DRBG *drbg)
         drbg->meth->uninstantiate(drbg);
     rand_pool_free(drbg->adin_pool);
     CRYPTO_THREAD_lock_free(drbg->lock);
-    CRYPTO_free_ex_data(CRYPTO_EX_INDEX_DRBG, drbg, &drbg->ex_data);
+    CRYPTO_free_ex_data(CRYPTO_EX_INDEX_RAND_DRBG, drbg, &drbg->ex_data);
```
levitte: How did this even build?????

levitte: Oh, now I see Tha change further down. Never mind

levitte:
There is a check of SYNOPSIS to see if what's in there matches what's given in the NAME section.

levitte:
Merged.

d318389eff0fb80d4a88c7af1cc314f185ba94e1 Document "get/set-app-data" macros.

PR #10217, opened 2019-10-18
Allow EVP_PKEY_get0_RSA for RSA-PSS keys

tniessen:
RSA-PSS keys use the same internal structure as RSA keys but do not allow accessing it through `EVP_PKEY_get0_RSA`. This commit changes that behavior.

I am not sure how to test this (or whether a test is even necessary), the function is never called directly from existing tests.

##### Checklist

### DIFF
--- a/crypto/evp/p_lib.c
+++ b/crypto/evp/p_lib.c
@@ -493,7 +493,7 @@ int EVP_PKEY_set1_RSA(EVP_PKEY *pkey, RSA *key)
 
 RSA *EVP_PKEY_get0_RSA(const EVP_PKEY *pkey)
 {
-    if (pkey->type != EVP_PKEY_RSA) {
+    if (pkey->type != EVP_PKEY_RSA && pkey->type != EVP_PKEY_RSA_PSS) {
         EVPerr(EVP_F_EVP_PKEY_GET0_RSA, EVP_R_EXPECTING_AN_RSA_KEY);
         return NULL;
     }


crypto/evp/p_lib.c
```diff
@@ -493,7 +493,7 @@ int EVP_PKEY_set1_RSA(EVP_PKEY *pkey, RSA *key)
 
 RSA *EVP_PKEY_get0_RSA(const EVP_PKEY *pkey)
 {
-    if (pkey->type != EVP_PKEY_RSA) {
+    if (pkey->type != EVP_PKEY_RSA && pkey->type != EVP_PKEY_RSA_PSS) {
```
bernd-edlinger: how did you create this object?
with EVP_PKEY_assign_RSA?

tniessen: I can't say for sure, but most likely using one of `PEM_read_bio_PrivateKey`, `d2i_PKCS8PrivateKey_bio`, `EVP_PKCS82PKEY`, `d2i_PUBKEY`, or `X509_get_pubkey` depending on the input format and context.

bernd-edlinger: You know, you are supposed to use EVP_PKEY_sign with the private key,
with the way you want to do it it will be an OAEP signature?

bernd-edlinger: I meant: EVP_DigestSign

tniessen: I am not sure what you mean, could you elaborate?  Why do you think this change is related to signatures?

bernd-edlinger: You are not supposed to use the RSA key directly,  creating a PSS signature
requires you to know the correct PSS-parameters, and would simply be ugly
to do with EVP_PKEY_get0_RSA, RSA_padding_add_PKCS1_PSS_mgf1 and
RSA_private_encrypt.  If you still want to do it you can of course
go with `(RSA*)EVP_PKEY_get0(evp)`.
The correct way to do it is using EVP_MD_CTX_new, EVP_DigestSignInit, EVP_DigestSignUpdate, and EVP_DigestSignFinal.
If you do it this way it will also work for ECC keys without any changes.

tniessen: Again, I am not trying to use the returned `RSA*` to sign data, I need to find out how large the modulus is and what the public exponent of the key is. (Copied from my last comment.)

bernd-edlinger: For number of bits of the public exponend use EVP_PKEY_size,
I don't understand what you could do with the public exponent,
but if you want to export a public key from the private key, you can do that with
i2d_PublicKey in principle.

tniessen: I am trying to implement a crypto API that conforms to the WebCrypto standard, which requires users to be able to retrieve the public exponent.

tniessen: >  If you still want to do it you can of course go with `(RSA*)EVP_PKEY_get0(evp)`.

Thank you for the hint, this seems to be a good workaround for now!

mattcaswell: Regardless of the above discussion, it seems a strange thing that `EVP_PKEY_get0_RSA` can't return an RSA-PSS key, so this change seems reasonable to me. Probably even for backport to 1.1.1

tniessen:
@mattcaswell Thank you!

On a related note: The last time I checked, OpenSSL did not provide APIs to access RSA-PSS parameters (such as the minimum salt length). Would it be appropriate to add such APIs, and if so, should those act on an `RSA*` or on an `EVP_PKEY*`?

mattcaswell:
Salt length can be accessed via `EVP_PKEY_CTX_get_rsa_pss_saltlen()`. See:

https://www.openssl.org/docs/man1.1.1/man3/EVP_PKEY_CTX_ctrl.html

mspncp:
ping: ready to merge

tniessen:
@mattcaswell Thank you! I am specifically looking to retrieve the values from RSA-PSS keys using OpenSSL:

```
      RSASSA-PSS-params  ::=  SEQUENCE  {
         hashAlgorithm     [0] HashAlgorithm DEFAULT
                                  sha1Identifier,
         maskGenAlgorithm  [1] MaskGenAlgorithm DEFAULT
                                  mgf1SHA1Identifier,
         saltLength        [2] INTEGER DEFAULT 20,
         trailerField      [3] INTEGER DEFAULT 1  }
```

Is that possible from the `EVP_PKEY*` without having to create a new `EVP_PKEY_CTX` first?

(Sorry to derail the conversation to a slightly off-topic area.)

mattcaswell:
Pushed. Thanks.

> Is that possible from the EVP_PKEY* without having to create a new EVP_PKEY_CTX first?

The pss params are stored in the rsa->pss structure. Unfortunately I don't think we expose this directly, so the only way I can think of getting this data is to create the EVP_PKEY_CTX first.

PR #10221, opened 2019-10-19
Configure: also break up the OPTIONS line

bbolli:
c00d9311 ("Configure: break long lines in build files", 2019-05-23)
introduced wrapping of long Makefile lines, but it forgot to wrap the
configuration options, which are generally also longer than one screen
line.

Signed-off-by: Beat Bolli <dev@drbeat.li>

### DIFF
--- a/Configurations/unix-Makefile.tmpl
+++ b/Configurations/unix-Makefile.tmpl
@@ -57,7 +57,9 @@
      '';
 -}
 PLATFORM={- $config{target} -}
-OPTIONS={- $config{options} -}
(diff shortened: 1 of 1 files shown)


levitte:
This one worries me, 'cause if there's anything quoted amongst the options, that string will be corrupted.

The options may cover more than one terminal line, but not too many. It's not like the *huge* list of files you get for some targets...

bbolli:
All right, that's a usage I didn't think of. I just looked at the
generated Makefile and the OPTIONS line really jarred.

Thanks anyway for the quick review!

PR #10231, opened 2019-10-21
Type casting (remove warning)

Kochise:
Trivial modification, one liner, type casting
CLA:trivial

### DIFF
--- a/crypto/threads_win.c
+++ b/crypto/threads_win.c
@@ -155,7 +155,7 @@ int CRYPTO_THREAD_compare_id(CRYPTO_THREAD_ID a, CRYPTO_THREAD_ID b)
 
 int CRYPTO_atomic_add(int *val, int amount, int *ret, CRYPTO_RWLOCK *lock)
 {
-    *ret = InterlockedExchangeAdd(val, amount) + amount;
+    *ret = (int)InterlockedExchangeAdd((long volatile*)val, (long)amount) + amount;
     return 1;
 }
 


Kochise:
Do not see what else I should provide.

t8m:
There must be CLA:trivial on a single separate line in the commit message. However I am not sure this change is right, the pointer should be cast to (void *).

Kochise:
Ok, but I guess getting this information only once you've committed and pushed makes things difficult to change. Should have been clearer in the "CLA help" with a working example of what is expected.

About the cast, I used the function declaration, which is "extern long __cdecl InterlockedExchangeAdd(long volatile*, long)" and does not imply any "void *". The compiler was happy enough with my cast. Otherwise, as an old timer's advice, avoid this kind of cascading macro redirection that only makes things rather confusing to follow. It rapidly becomes a mess...

t8m:
You can do git commit --amend and then git push --force. So no problem with adjusting patch messages (esp. for the last commit, but there are also ways to do it for older ones). git commit --amend also allows modifying the patch contents if you git add before it.
As for the pointer cast - perhaps the MSVC compiler does not care but in general it is not a good idea to cast between pointer types due to aliasing rules.

Kochise:
Commit : did it online so I guess I've lost track of the commit uuid already, hence no opportunity to do an amend.

Casting : used gcc. Not my fault if the source code mix 'long' with 'int' all over the place. And should I mention the silly macro redirection ? Oops, did that already in a previous comment.

Kochise:
Until further help is provided to me, I'll not be able to solve this issue by myself. I've opened a second similar pull request that have the same problem without any clue given about how to do the cla stuff that is requested,

t8m:
I can only repeat what I wrote above: There must be text `CLA: trivial` on a single separate line in the commit message.

Kochise:
Look in https://github.com/openssl/openssl/pull/11337 I just did that and it doesn't work.

kaduk:
There's a space: `CLA: trivial`.  Yours in #11337 has no space (`CLA:trivial`)

Kochise:
t8m commented on 5 Mar with "CLA:trivial" (no space), and I followed that because there is no other mention of the correct syntax elsewhere. Either get your documentation fixed, or set your system a bit more permissive.

t8m:
I'm sorry for misleading you with the missing space.

Kochise:
No problem, I'm not angry with you, I've a grief against this stubborn bureaucratic process and unfriendly integration system.

Kochise:
Finally integrated in #11337

PR #10232, opened 2019-10-21
Update control logic for BN_gcd

### LINKED ISSUE #10224: regression test failures with --api=3.0.0
Build without deprecated API is fine
```
...
/usr/bin/perl ./Configure --prefix=XXXX/master_inplace shared no-ssl3 zlib-dynamic enable-unit-test --api=3.0.0 linux-x86_64
...
CFLAGS          = -Wall -O3 -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -g
...
```

```
No compiler warning with above flags.

Unfortunately now some tests start to fail:
...
Test Summary Report
-------------------
03-test_internal_bn.t            (Wstat: 256 Tests: 1 Failed: 1)
  Failed test:  1
  Non-zero exit status: 1
03-test_internal_rsa_sp800_56b.t (Wstat: 256 Tests: 1 Failed: 1)
  Failed test:  1
  Non-zero exit status: 1
10-test_bn.t                     (Wstat: 512 Tests: 7 Failed: 2)
  Failed tests:  6-7
  Non-zero exit status: 2
15-test_genrsa.t                 (Wstat: 512 Tests: 5 Failed: 2)
  Failed tests:  3, 5
  Non-zero exit status: 2
15-test_mp_rsa.t                 (Wstat: 512 Tests: 31 Failed: 2)
  Failed tests:  8, 23
  Non-zero exit status: 2
15-test_rsa.t                    (Wstat: 256 Tests: 6 Failed: 1)
  Failed test:  3
  Non-zero exit status: 1
20-test_enc_more.t               (Wstat: 256 Tests: 2 Failed: 1)
  Failed test:  1
  Non-zero exit status: 1
30-test_evp_extra.t              (Wstat: 256 Tests: 1 Failed: 1)
  Failed test:  1
  Non-zero exit status: 1
Files=180, Tests=1576, 200 wallclock secs (11.03 usr  0.85 sys + 158.77 cusr 23.30 csys = 193.95 CPU)
```

Dunno if above is related to failure in e_nss engine - segmentation fail when is created DSA certificate for test purpose (only OpenSSL commands).

cpereida:
PR https://github.com/openssl/openssl/pull/10122 introduced changes to the BN_gcd function and the control logic inside it accessed `g->d[0]` irrespective of `g->top`.

When BN_add is called, in case the result is zero, `BN_zero` is called.
The latter behaves differently depending on the API compatibility level flag: normally `g->d[0]` is cleared but in `no-deprecated` builds only `g->top` is set to zero.

This commit uses bitwise logic to ensure that `g` is treated as zero if `g->top` is zero, irrespective of `g->d[0]`.

### Side notes

- `BN_add` requires refactoring to be constant-time.
- After refactoring these changes can probably go away since we can expect `g->d` to be properly zeroed.
- The second commit in this PR should be dropped, it is only meant to run `make test` in `no-deprecated` builds.
- `20-test_enc_more` seems to be currently failing but does not seem to depend on #10122.

@romen co-authored this.

Partially fixes #10224

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ matrix:
         - os: linux
           dist: trusty
           compiler: clang
-          env: CONFIG_OPTS="--strict-warnings -D__NO_STRING_INLINES no-deprecated"
+          env: CONFIG_OPTS="--strict-warnings -D__NO_STRING_INLINES no-deprecated" BUILDONLY="yes"
         - os: linux
           dist: bionic
           compiler: clang
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ matrix:
         - os: linux
           dist: trusty
           compiler: clang
-          env: CONFIG_OPTS="--strict-warnings -D__NO_STRING_INLINES no-deprecated" BUILDONLY="yes"
+          env: CONFIG_OPTS="--strict-warnings -D__NO_STRING_INLINES no-deprecated"
         - os: linux
           dist: bionic
           compiler: clang
--- a/crypto/bn/bn_gcd.c
+++ b/crypto/bn/bn_gcd.c
@@ -593,7 +593,9 @@ int BN_gcd(BIGNUM *r, const BIGNUM *in_a, const BIGNUM *in_b, BN_CTX *ctx)
 
     for (i = 0; i < m; i++) {
         /* conditionally flip signs if delta is positive and g is odd */
-        cond = (-delta >> (8 * sizeof(delta) - 1)) & g->d[0] & 1;
+        cond = (-delta >> (8 * sizeof(delta) - 1)) & g->d[0] & 1
+            /* make sure g->top > 0 (i.e. if top == 0 then g == 0 always) */
+            & (~((g->top - 1) >> (sizeof(g->top) * 8 - 1)));
         delta = (-cond & -delta) | ((cond - 1) & delta);
         r->neg ^= cond;
         /* swap */
@@ -603,7 +605,10 @@ int BN_gcd(BIGNUM *r, const BIGNUM *in_a, const BIGNUM *in_b, BN_CTX *ctx)
         delta++;
         if (!BN_add(temp, g, r))
             goto err;
-        BN_consttime_swap(g->d[0] & 1, g, temp, top);
+        BN_consttime_swap(g->d[0] & 1 /* g is odd */
+                /* make sure g->top > 0 (i.e. if top == 0 then g == 0 always) */
+                & (~((g->top - 1) >> (sizeof(g->top) * 8 - 1))),
+                g, temp, top);
         if (!BN_rshift1(g, g))
             goto err;
     }


cpereida:
Ping @bbbrumley, @mattcaswell, @levitte

romen:
@mattcaswell @levitte given that it seems we are starting to touch things that tend to break with `no-deprecated` at runtime rather than at compilation time, should we think about actually applying the second commit so that Travis can pick up these things already during the review?

t8m:
The question is why do we have such deep low-level behavioral difference based on enabling/disabling no-deprecated. I do not think this is right.

levitte:
I agree that the Travis change should stay, exactly for the reason @romen mentions.

@t8m has a valid concern, the reasons might be historical, I dunno.  BN code is one of those areas I haven't dived in too much lately...

levitte:
There's a Travis failure in https://travis-ci.org/openssl/openssl/jobs/600706103, I'm unsure if that has anything to do with this PR...

romen:
@t8m @levitte 

I will write-up later a longer explanation of the problem and why there even is a difference in `no-deprecated` builds.

I will also describe some alternative choices and why we decided to go this route rather than changing things elsewhere, but it will have to wait until tonight.

romen:
> There's a Travis failure in https://travis-ci.org/openssl/openssl/jobs/600706103, I'm unsure if that has anything to do with this PR...

@levitte yes, that's the one we mentioned in the description: we noticed it after fixing the changes from #10122 and we know for sure that the cause is coming in one of the commits since 3d48457478bd61030c370e4090c1462fc4453d81 as the error was not there in our development branch before rebasing on latest `master`

(one more reason to actually do runtime tests for `no-deprecated` builds.)

romen:
Here is a more exhaustive write-up of the problem associated with `no-deprecated` builds and the original changes from #10122 .

# The problem _(a.k.a. Funny Monday rabbit hole)_

The following snippet include the main loop inside the new implementation of `BN_gcd()`:
https://github.com/openssl/openssl/blob/f3c4adfc7eb13e9eff514039b4c60b457bdba433/crypto/bn/bn_gcd.c#L594-L609

In particular https://github.com/openssl/openssl/blob/f3c4adfc7eb13e9eff514039b4c60b457bdba433/crypto/bn/bn_gcd.c#L596
and
https://github.com/openssl/openssl/blob/f3c4adfc7eb13e9eff514039b4c60b457bdba433/crypto/bn/bn_gcd.c#L606
perform bitwise logic to determine if `g` (a `BIGNUM*`) is odd: we use a bitwise `g->d[0] & 1` and avoid calling `BN_is_odd()` so that the compiler does not try to optimize in ways that break the constant-time guarantees we are trying to achieve.

Through the various iterations and conditional constant-time swaps, the state of `g` is mostly determined by the result of
https://github.com/openssl/openssl/blob/f3c4adfc7eb13e9eff514039b4c60b457bdba433/crypto/bn/bn_gcd.c#L604

Here is the current implementation for `BN_add()` (which is most definitely likely to leak information through branching, timing and instruction cache attacks):
https://github.com/openssl/openssl/blob/f3c4adfc7eb13e9eff514039b4c60b457bdba433/crypto/bn/bn_add.c#L14-L42

Focus on the case in which `a + b == 0`, line 34 calls `BN_zero()` and here is where the behaviour in `no-deprecated` builds change:

https://github.com/openssl/openssl/blob/f3c4adfc7eb13e9eff514039b4c60b457bdba433/include/openssl/bn.h#L201-L205

In _standard_ builds `BN_zero()` behaves like this:

https://github.com/openssl/openssl/blob/f3c4adfc7eb13e9eff514039b4c60b457bdba433/crypto/bn/bn_lib.c#L403-L414

In `no-deprecated` builds like this:

https://github.com/openssl/openssl/blob/f3c4adfc7eb13e9eff514039b4c60b457bdba433/crypto/bn/bn_lib.c#L893-L898

In the first case `a->d[0]` is set to `0`, while in the latter only `a->top` is set to 0, and `a->d[0]` is left unchanged (which is totally legit, as `BN_is_zero(a)` has historically been equivalent to `a->top == 0`.

# Potential fixes

## ❌ Solution 1: "patch" `BN_zero_ex()`

The first solution I evaluated was to _fix_ `BN_zero_ex()` so that `a->d[0]` is also cleared.
It seems like a harmless minimal change, but it has indeed the potential of masking away places where we are violating our own existing internal contract from the definition of `BN_is_zero()` and the `BIGNUM` struct itself:

https://github.com/openssl/openssl/blob/f3c4adfc7eb13e9eff514039b4c60b457bdba433/crypto/bn/bn_local.h#L218-L226

We shouldn't be accessing `bn->d[0]` if `bn->top == 0`

## ❌ Solution 1.bis: "patch" `BN_add()`

Similarly to the previous solution, inside `BN_add()`, after the `BN_zero()` call, we could add a

```c
r->d[0] = 0;
```

but this seems conceptually wrong on all levels: `BN_zero()` knows hot to set `BIGNUM`s to 0, it shouldn't be messing with it.

## ✓ Solution 2: "fix" the control logic in the new `BN_gcd()`

One could say that the problem that we overlooked during the review of #10122 is that in the new `BN_gcd()` we (semi)-unconditionally access `g->d[0]` to check if `g` is odd, irrespective of `g->top`, while our internal contract says that if `g->top == 0` then we shouldn't access `g->d[0]`.

This was masked by the default implementation of `BN_zero()` which was conveniently also zeroing `g->d[0]`.

To solve this so that the code is formally correct, and retain the constant-time properties, we need to use clever bitwise masking based on `g->top` to discard any result based on `g->d[0]` when the latter is in an undefined state (i.e., `g->top == 0`).

We envisioned two potential ways of fixing this

### Solution 2.a: (current) fix every access to `g->d[0]`

This is the current changeset, we make sure that every access to `g->d[0]` is formally correct by also mixing in the state of `g->top`:

```diff
diff --git a/crypto/bn/bn_gcd.c b/crypto/bn/bn_gcd.c
index fbefe4ab6a..bed9fca4d9 100644
--- a/crypto/bn/bn_gcd.c
+++ b/crypto/bn/bn_gcd.c
@@ -593,7 +593,9 @@ int BN_gcd(BIGNUM *r, const BIGNUM *in_a, const BIGNUM *in_b, BN_CTX *ctx)

     for (i = 0; i < m; i++) {
         /* conditionally flip signs if delta is positive and g is odd */
-        cond = (-delta >> (8 * sizeof(delta) - 1)) & g->d[0] & 1;
+        cond = (-delta >> (8 * sizeof(delta) - 1)) & g->d[0] & 1
+            /* make sure g->top > 0 (i.e. if top == 0 then g == 0 always) */
+            & (~((g->top - 1) >> (sizeof(g->top) * 8 - 1)));
         delta = (-cond & -delta) | ((cond - 1) & delta);
         r->neg ^= cond;
         /* swap */
@@ -603,7 +605,10 @@ int BN_gcd(BIGNUM *r, const BIGNUM *in_a, const BIGNUM *in_b, BN_CTX *ctx)
         delta++;
         if (!BN_add(temp, g, r))
             goto err;
-        BN_consttime_swap(g->d[0] & 1, g, temp, top);
+        BN_consttime_swap(g->d[0] & 1 /* g is odd */
+                /* make sure g->top > 0 (i.e. if top == 0 then g == 0 always) */
+                & (~((g->top - 1) >> (sizeof(g->top) * 8 - 1))),
+                g, temp, top);
         if (!BN_rshift1(g, g))
             goto err;
     }
```

### Solution 2.b: apply the bitwise masking after `BN_add()`

Alternatively we could do the bitwise masking only once, after `BN_add()`:

```diff
diff --git a/crypto/bn/bn_gcd.c b/crypto/bn/bn_gcd.c
index fbefe4ab6a..e9be4dba92 100644
--- a/crypto/bn/bn_gcd.c
+++ b/crypto/bn/bn_gcd.c
@@ -603,6 +603,10 @@ int BN_gcd(BIGNUM *r, const BIGNUM *in_a, const BIGNUM *in_b, BN_CTX *ctx)
         delta++;
         if (!BN_add(temp, g, r))
             goto err;
+
+        /* Ensure `temp->d[0] == 0` if `temp == 0` (i.e., `temp->top == 0`) */
+        temp->d[0] &= ~((temp->top - 1) >> (sizeof(temp->top) * 8 - 1));
+
         BN_consttime_swap(g->d[0] & 1, g, temp, top);
         if (!BN_rshift1(g, g))
             goto err;

```

This looks more self-contained, potentially more readable, and nicer performance-wise as we do this bitwise masking only once per iteration rather than twice (as in *Sol 2.a*).

The drawback is that I feel it's less maintainable:
- on the very first iteration of this loop, the first execution of line 596 https://github.com/openssl/openssl/blob/f3c4adfc7eb13e9eff514039b4c60b457bdba433/crypto/bn/bn_gcd.c#L596 is guaranteed to be correct only because we check the corner cases where either one of the inputs is zero at the beginning of the function: https://github.com/openssl/openssl/blob/f3c4adfc7eb13e9eff514039b4c60b457bdba433/crypto/bn/bn_gcd.c#L537-L549 . If we were to later change our minds on that comment and stop treating these two cases as special cases, then line 596 would fall again into undefined behavior;
- logically we are basically ad-hoc patching the behaviour of `BN_zero()` called from `BN_add()`, to ensure our accesses to `g->d[0]` are valid even if `g->top` is zero. This is conceptually not different from *Sol 1.b* that we disliked, and it only works as long as `BN_add()` is the only place that could result in zeroing one of the `BIGNUM`s that define the internal state of the algorithm. If we later were to modify this function for some reason changing the order of steps or adding other potential callers of `BN_zero()` the very same bug could creep up on us again.


# Conclusions (TL;DR)

So here are the reasons why we are proposing this solution over the described alternatives.

In the end it could be summarized as:
- we believe #10122 introduced undefined behaviour by accessing `g->d[0]` irrespective of `g->top`, thus breaking our internal contract defined by the `BIGNUM` struct definition;
- the proper way to fix it, even if not the most convenient, is to patch every access to `g->d[0]` to be always defined, factoring in the state of `g->top`


# P.S.

- Ping to @levitte and @t8m as they raised this absolutely valid concern:
   > The question is why do we have such deep low-level behavioral difference based on enabling/disabling no-deprecated. I do not think this is right.

- Ping to @cpereida in case he wants to add something to this write-up that I might have forgotten from our discussions today while debugging and designing a fix for this

romen:
@levitte I have a sidenote for you:

Looking at https://github.com/openssl/openssl/blob/f3c4adfc7eb13e9eff514039b4c60b457bdba433/include/openssl/bn.h#L201-L205  to me it's counter-intuitive that `no-deprecated` (i.e., `--api=3.0.0`) selects the first definition rather than the second.

Is it just too late for me to be looking at this or indeed there is a separate problem to be fixed there?

mattcaswell:
Nice write up @romen!

kroeckx:
Is there a way to fix BN_is_odd() so it provides the properies that
you need? Maybe an internal function?

romen:
@kroeckx 
> Is there a way to fix BN_is_odd() so it provides the properies that you need? Maybe an internal function?

One option is to have it as a macro, and we actually considered if it was worth to define a `local_bn_is_odd()` macro local to this file to improve readability.

The problem with using it as a function is that different versions of different compilers targeting different platforms can start trying to do all kinds of optimizations when they see a function call.
In many cases they will translate into assembly with a jump to the function memory location and a conditional branch on returning.
An `inline` function declaration might be helpful, but again, not all compilers support `inline` as it is not technically `C89/C90`, and even then there is no guarantee that an `inline` function will not translate into a conditional branch.

We want to avoid branches (and especially branches silently inserted by compilers in the assembly code we usually we don't look at) because inevitably they will leave different footprints in caches/TLBs/etc. due to different instruction-related memory access patterns, and potentially other side-channels like timing differences, branch predictor states, port contention footprint, etc.

Luckily there is only so much that compilers try to optimize about bitwise logic, so that's the most reliable way (that we know of) of avoiding branching across a plethora of compilers, compiler versions and platforms.

As a side note, of course we could dream of a day where we will stop fighting the compilers and actually be able to reliably (as in a widely adopted standard) annotate functions or code blocks so that we can instead collaborate with compilers and signal the programmer intention of locally disabling some/most optimization steps for critical sections.
Unfortunately even if we were able to release a new C standard right now establishing such annotations, it would still take eons before all the platforms OpenSSL supports would be guaranteed to reliably support them.

romen:
I would like the reviewers to give feedback also on the following nit:

`sizeof(g->top)` VS `sizeof(int)`

Which one would you prefer? I was undecided.

levitte:
The BN_zero definition depending on OPENSSL_API_0_9_8 shows a pretty big confusion between an intent with the `--api` configuration option (which would intuitively mean "please give me API from this and that version") and `no-deprecated`. They should really not be intermixed. @t-j-h pointed this out in a comment somewhere a while ago, and it's been in the back burner of my mind since then...  So yeah, it would be logical to think that *unless* someone has explicitly asked for `--api=0.9.8`, the "new and improved" BN_zero should be the one we use.

levitte:
> `sizeof(g->top)` VS `sizeof(int)`

À very simple question to ask oneself for this case (or any case like this) is, what if `g->top` changes type at some point?

levitte:
I fully intend to look over the logic of the macros surrounding `--api` and `no-deprecated`.

t8m:
Does it even make sense to support something like --api=0.9.8 when there are so many things removed from the public OpenSSL headers that were there in 0.9.8? To me going back further than 1.1.0 does not make much sense.

romen:
> > `sizeof(g->top)` VS `sizeof(int)`
> 
> A very simple question to ask oneself for this case (or any case like this) is, what if `g->top` changes type at some point?

Right, and that's why I defaulted to `sizeof(g->top)` but I heard contrasting opinions on the matter before (similar to my habit of writing `NULL == p` in conditions to avoid inserting typos where I forget an `=` sign.

romen:
-Regarding the first commit of this PR, have all other concerns be addressed?
-Regarding the second commit, should we edit the commit message to make it permanent as part of this PR or should that be addressed in its own PR?

romen:
Also notice that the first commit is meant to be backported to `1.1.1` as well, as a bugfix.

If the second commit was to be merged as the result of this PR, should it also be backported to `1.1.1`?

levitte:
I think both commits should be cherry-picked to 1.1.1.

levitte:
> Does it even make sense to support something like --api=0.9.8 when there are so many things removed from the public OpenSSL headers that were there in 0.9.8? To me going back further than 1.1.0 does not make much sense.

Agreed.

romen:
@kroeckx resuming from https://github.com/openssl/openssl/pull/10232#issuecomment-544739949 this is an interesting read about our daily fight with compilers: https://dsprenkels.com/cmov-conversion.html

kroeckx:
Which is why we have value_barrier() now in constant_time.h

romen:
> I think both commits should be cherry-picked to 1.1.1.

@levitte we pushed two more commits: the first reverts the previous temporary commit, the second reapplies the same change, but with a proper commit message.

- 374f790
- 9f6322a

richsalz:
Re: sizeof.  IMO, *always* use the instance (p->foo) not the type (int) for future safety.  In 1.1.0 we did a pass over all malloc code to ensure that this was followed.  It was never made part of the coding style, perhaps it should be.

bbbrumley:
> Which is why we have value_barrier() now in constant_time.h

I'm ashamed to admit I did not know this functionality existed. TIL. Looks like I have some homework to do. Thanks @kroeckx !

romen:
@t8m @kroeckx you raised a few concerns: would you like to keep the discussion open or should I proceed merging?

kroeckx:
Feel free to merge it.

t8m:
You can merge. My concern is for separate Issue/PR.

romen:
Merged on `master` with:
- 8aca4bfe82 Update control logic for BN_gcd
- c89799605b Enable runtime testing of no-deprecated builds in Travis

Merged on `1.1.1` with:
- 85728d08ae Update control logic for BN_gcd
- 432d6953d6 Enable runtime testing of no-deprecated builds in Travis


Thanks all!

Issue #9095, opened 2019-06-06
Newly added undocumented symbols


mattcaswell:
PR #9094 revealed that since 1.1.1 was released we have introduced 91 new public symbols that are undocumented. The complete list is:

````
# Found 4826 in util/libcrypto.num
crypto:CRYPTO_siv128_aad
crypto:CRYPTO_siv128_cleanup
crypto:CRYPTO_siv128_copy_ctx
crypto:CRYPTO_siv128_decrypt
crypto:CRYPTO_siv128_encrypt
crypto:CRYPTO_siv128_finish
crypto:CRYPTO_siv128_get_tag
crypto:CRYPTO_siv128_init
crypto:CRYPTO_siv128_new
crypto:CRYPTO_siv128_set_tag
crypto:CRYPTO_siv128_speed
crypto:ERR_load_CMP_strings
crypto:ERR_load_CRMF_strings
crypto:ERR_load_ESS_strings
crypto:ESS_SIGNING_CERT_V2_new_init
crypto:ESS_SIGNING_CERT_new_init
crypto:EVP_CIPHER_fetch
crypto:EVP_CIPHER_upref
crypto:EVP_MAC_do_all
crypto:EVP_MAC_do_all_sorted
crypto:EVP_aes_128_siv
crypto:EVP_aes_192_siv
crypto:EVP_aes_256_siv
crypto:EVP_hex2ctrl
crypto:EVP_str2ctrl
crypto:OPENSSL_LH_flush
crypto:OSSL_PARAM_construct_double
crypto:OSSL_PARAM_construct_int
crypto:OSSL_PARAM_construct_int32
crypto:OSSL_PARAM_construct_int64
crypto:OSSL_PARAM_construct_long
crypto:OSSL_PARAM_construct_size_t
crypto:OSSL_PARAM_construct_uint
crypto:OSSL_PARAM_construct_uint32
crypto:OSSL_PARAM_construct_uint64
crypto:OSSL_PARAM_construct_ulong
crypto:OSSL_PARAM_get_double
crypto:OSSL_PARAM_get_int
crypto:OSSL_PARAM_get_int32
crypto:OSSL_PARAM_get_int64
crypto:OSSL_PARAM_get_long
crypto:OSSL_PARAM_get_size_t
crypto:OSSL_PARAM_get_uint
crypto:OSSL_PARAM_get_uint32
crypto:OSSL_PARAM_get_uint64
crypto:OSSL_PARAM_get_ulong
[...]

PR #10236, opened 2019-10-22
add compensational definition for OPENSSL_MSTR in macros.h

DDvO:
The recent commit f3866324f0e9bd33feb48aefbe18d8f465d8f2a2 broke `include/opensll/cmp_util.h` when the trace API is enabled.
This PR adds in that header file a compensational declaration.

### DIFF
--- a/doc/man3/OPENSSL_FILE.pod
+++ b/doc/man3/OPENSSL_FILE.pod
@@ -41,7 +41,8 @@ L<crypto(7)>
 
 =head1 HISTORY
 
-B<OPENSSL_FUNC> was added in OpenSSL 3.0.
+B<OPENSSL_FUNC>, B<OPENSSL_MSTR>, and B<OPENSSL_MSTR_HELPER>
+were added in OpenSSL 3.0.
 
 =head1 COPYRIGHT
 
--- a/doc/man3/OPENSSL_FILE.pod
+++ b/doc/man3/OPENSSL_FILE.pod
@@ -1,0 +1,55 @@
+=pod
+
+=head1 NAME
+
+OPENSSL_FILE, OPENSSL_LINE, OPENSSL_FUNC,
+OPENSSL_MSTR, OPENSSL_MSTR_HELPER
+- generic C programming utility macros
+
+=head1 SYNOPSIS
+
+ #include <openssl/macros.h>
+
+ #define OPENSSL_FILE /* typically: __FILE__ */
+ #define OPENSSL_LINE /* typically: __LINE__ */
+ #define OPENSSL_FUNC /* typically: __func__ */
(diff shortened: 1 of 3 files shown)


DDvO:
all right, done

mattcaswell:
These two macros were two of the "newly added but undocumented" macros that got into the 3.0 codebase before we set doc-nits to start complaining about undocumented stuff. See #9095. f386632 co-incidentally fixed that by removing the macros - so this PR introduces it again. doc-nits doesn't complain because they weren't also removed form missingmacros.txt. But we should probably document these macros.

DDvO:
No problem to add some documentation for these two macros, but where to do this?

I would have put it wherever similar macros are documented, but also (most of) the other macros in `macros.h`, including  `OPENSSL_API_LEVEL`, `OPENSSL_FILE` and `OPENSSL_FUNC` are not (yet) documented either.

levitte:
Do these have to be public?

DDvO:
Currently the two macros are used only in the public header `include/openssl/cmp_util.h`, namely in the implementation of the `OSSL_CMP_LOG_START` macro. So, for technical reasons, the macro is already public, but we could `#undef` it after use.

On the other hand, the CCP string composition trick is a generally useful macro (and can be found online as C programming hint under various names). So we could also export it for the benefit of others.

Which way do you prefer?
If we go for exporting it, in which file should I document it?

levitte:
Nah, if we use it publically, let's expose if publically.

DDvO:
Fine, and where to document these macros?

DDvO:
Since I got no response on where to document these two macros,
I've introduced for this purpose `doc/man3/OPENSSL_FILE.pod`,
where I also documented `OPENSSL_FILE`, `OPENSSL_LINE`, and `OPENSSL_FUNC`.

levitte:
Uhmmmm, this PR is suddenly a minor gazillion commits.  What happened?

DDvO:
Hmm, it looks like rebase went somewhat wrong. Fixed.

DDvO:
Recent Travis CI  hang resulted in false positive. This time it worked.

DDvO:
Can someone please give a second approval on this and merge it soon?
Our CMP contribution chunk 6 depends on this.

mattcaswell:
> Can someone please give a second approval on this and merge it soon?

Merge happens (at the earliest) 24 hours after "approval: done" is reached.

DDvO:
Thanks for the swift (re-)action!

mspncp:
ping: ready to merge

mattcaswell:
I pushed this. Note: I merged the last 2 commits into 1.

PR #10238, opened 2019-10-22
Fix memory deallocation code in `kssl_build_principal_2`

liumuqing:
##### Checklist

##### 
The error handling code in `kssl_build_principal_2` wrongly use `free(new_p[1].data)`, while `new_p` is not a pointer to array, when allocation fail.(when `calloc` return `NULL`);
I think it is intended to free `p_data[0].data` and `p_data[1].data`.
This patch just fix it.

### DIFF
--- a/ssl/kssl.c
+++ b/ssl/kssl.c
@@ -2234,10 +2234,10 @@ krb5_error_code kssl_build_principal_2(
     return 0;
 
  err:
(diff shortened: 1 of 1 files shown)


mattcaswell:
OpenSSL 1.0.2 is in security-fix only mode and goes out of support completely at the end of this year, so this won't be fixed.

PR #10247, opened 2019-10-23
WIP: Convenience function for -XXXform flags

richsalz:
I am not sure this is worth doing because it put all the "form" (e.g., -certform, -keyform) items in together, as opposed to being next to the thing (e.g., -cert -key).  But it does show to @levitte that I'm learning Perl whether I want to or not :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,35 @@
 /include/openssl/opensslconf.h
 /include/openssl/opensslv.h
 
+# Auto generated doc files
+doc/man1/openssl-ca.pod
+doc/man1/openssl-cms.pod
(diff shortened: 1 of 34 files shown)


richsalz:
Based on #10159 so only the first commit is part of this PR.

levitte:
I see what that's doing, and while I agree with the charm of that function, I start to worry about what that does for those who might want to write more docs but not necessarily want to dive too much into perl.  Somewhere, there needs to be a balance...

richsalz:
I totally agree with your comments, https://github.com/openssl/openssl/pull/10247#issuecomment-545746871  This was a fun hack and learning exercise :)  Closing.

PR #10242, opened 2019-10-23
20-test_enc_more.t: Replace deprecated -ciphers option with -list

t8m:
Fixes #10224 

##### Checklist

- [x] tests are added or updated

PR #10248, opened 2019-10-24
TESTS: use 'openssl enc -list' instead of 'openssl enc -ciphers'

levitte:
'-ciphers' is deprecated, using it makes some builds fail.


romen:
Should be addressed with #10242

levitte:
Thanks, yeah, I just discovered that PR myself...

PR #10100, opened 2019-10-05
Bad links 2

richsalz:
This brings the number of broken links in our manpages down from 87 to 46.  Each "unit" of change is a separate commit.

These errors were found via ```./util/find-doc-nits -l```

Many of these commits could be backported.

PR #10104, opened 2019-10-05
Fix the almost-final set of broken L<> links.

richsalz:
See #10103 for the methods used.

PR #10256, opened 2019-10-24
Ignore some links to missing docs

richsalz:
This is the "great compromise" that made me rethink #10100.  Have a file that lists the missing manpages (those items that are an L<...> target but don't exist) and don't complain about them.

To avoid bit-rot in that file, it's an error if a file listed as missing is found to exist.


util/find-doc-nits
```diff
@@ -872,6 +872,22 @@ if ( $opt_c ) {
...
+        or die "Can't open missingdocs.txt, $!";
+    while ( <$MISS> ) {
+        chop;
+        next unless m@man[1357]/@;
+        my $file = $_;
+        err("util/missingdocs.txt:1: File doc/$file exists")
+            if -f "doc/$file";
+        if ( $file =~ m@man([1357])/(.*).pod@ ) {
+            my $link = "$2($1)";
+            $name_map{$link} = $file;
```
levitte: There's a problem with this; we often have several functions in one pod, so name-to-pod mapping isn't accurate.

For accuracy, it would probably be better to have a db of all documented names including the section number (should be easy to generate from the collected names coupled with the section number of the currently processed), and a db of existing links (same format), a record any name in the latter db that's missing in the former db.

richsalz: Yes, it's not perfect.  It gets things much further along however.  The issue you mention, however, can be worked-around by adding multiple entries man3/foo and man4/bar if both are in the to-be-written foo.pod page.

In my view, doing a fully accurate DB of all that information is overkill. It's also not something I think I will have the time to do.  You can consider this PR to be the first pass of that, and a later PR could turn the missing file into a api/file list.

richsalz: Another thought that occurred to me is that database will not capture the "man7" missing references (which don't have function names)

levitte: That last remark doesn't really matter, because the links *must* match a name in a NAME section. It just so happens that man3 pages have function names there.

richsalz: Er, okay, maybe.  I still think this is a very useful step toward what you think of as the ultimate way to go.

richsalz:
fixup commit pushed to resolve https://github.com/openssl/openssl/pull/10100#discussion_r339113582

util/find-doc-nits
richsalz: It would be cool if the www Makefile used this to create "this page not yet written" pages. :)

richsalz:
rebased, resolved conflicts, pushed.

richsalz:
... and just to you know, I really wanted to call the file "missing.links" :)

util/find-doc-nits
levitte: We could have a generic 404 page that does just that...

richsalz:
It would be great if this PR, #10104, and #10100 could get reviewed and merged soon.
Together they address a whole class of errors (links to not-written manpages).

util/find-doc-nits
richsalz: @levitte Could you think about this again, and see if you can live with the approach in this PR, at least for now?

richsalz: Reposting: @levitte, could you think about this again and see if you can live with the approach in this PR, at least for now?

richsalz:
This had merge conflicts so I rebased and fixed the conflict in find-doc-nits

util/find-doc-nits
levitte: No.  However, I solved the problem for you, see #10621

levitte:
This should be pending #10621

richsalz:
I fixed the conflict, but not sure I did it right since now there are a bunch of "reference to non-existing" :(  Will look at this probably after the holidays

util/find-doc-nits
```diff
@@ -708,7 +708,7 @@ sub collectnames {
 
     my @foreign_names =
         map { map { s/\s+//g; $_ } split(/,/, $_) }
-        $podinfo{contents} =~ /=for\s+openssl\s+foreign\s+manuals:\s*(.*)\n\n/;
+        $podinfo{contents} =~ /=for openssl foreign manual (.*)\n/;
```
levitte: Why?

util/missingdocs.txt
```diff
@@ -0,0 +1,45 @@
+# Manpages references, but not yet written.pod
+man3/ASN1_item_sign.pod
```
levitte: What's this?

util/find-doc-nits
richsalz: simpler syntax, more consisten with the other meta-comments.

richsalz: (see 'for openssl ifdef' for example)

util/missingdocs.txt
richsalz: It was related to https://github.com/openssl/openssl/pull/10256#discussion_r338804384 which is now gone.  I need to look at this PR all over again :(  Hasty rebase I guess.

util/find-doc-nits
levitte: ... ok.

I would push for the other way around, for several reasons:

1. yours is unforgiving.  One space mistake somewhere (one too few or one too many) and it won't catch.  Silently.
2. the construct allows for several manuals, hence plural

Personally, I prefer the colon to be there, but oh well.

(why I haven't said anything before?  A lapse, I suppose...)

richsalz: The "for ifdef" allows several words as well.  And if the syntax is wrong, then the line won't be parsed and it will show up as doc-link errors.  Once I fix this PR, that will be analogous to a doc-nits error so I'm not worried about getting the syntax wrong.  A followon PR can always change 'for manuals' and 'for ifdef' to whatever you want, of course :)

levitte: Not sure I understand what you're saying. Should we be more unforgiving to those who write the pod files (more unforgiving than the real pod parsers)?

richsalz: Right now the concern is theoretical, and will be caught at build time.

richsalz:
I fixed most of the damage I left happen when rebasing to capture #10621 and pushed a commit. There are now 19 errors that I need to patch up.

util/find-doc-nits
```diff
@@ -872,6 +872,18 @@ if ( $opt_l || $opt_u || $opt_v ) {
...
 if ( $opt_l ) {
+    # Load in manpages that are known to not exist
+    open FH, "util/missingdocs.txt"
+        or die "Can't open util/missingdocs.txt, $!";
+    while ( <FH> ) {
+        chop;
+        next if /^#/;
+        my $pod = $_;
+        err("$pod in missingdocs.txt but it exists") if -f "doc/$pod";
+        $pod =~ s@man(.)/(.*)\.pod@$2($1)@;
```
levitte: Why not have in *name*(*n*) for in the file? After all, that's what we're checking, isn't it? That, say, the manual for `whatever(3)` doesn't exist? Why do you want us to go through hoops to rewrite that as `man3/whatever.pod` before putting it in `util/missingdocs.txt`, when we could just put in `whatever(3)` directly?

richsalz: Line 802 and 803 do different things; one checks for the file existing, the substitution turns it into an ```L<>``` reference.  Now, it *is* possible to move all of the entires in missingdocs.txt into manpage-format links and merge it into missingcrypto (which should be named missingapi I guess).  I can do that if you want, but we lose the check in line 802. If I do that, then some mode of find-doc-nits should warn if something is in missing but is documented.

richsalz: https://github.com/openssl/openssl/issues/10681

levitte: The notion that there is a 1:1 mapping  between manual references and pod files is a *delusion*. It simply doesn't work that way, since every pod file can have multiple names (and thereby be known as multiple manual references). And since the links are checked a few lines down, you do take the lines from this missingdocs.txt into account anyway, so line 882 (not 802) misses the point.

richsalz: Well, these are things that do not exist right now, so it isn't necessarily delusional.  There's several things that have a link to OSSL_STORE.  What API is that?

levitte: OSSL_STORE is the store API, i.e. crypto/store (dude, where's your grep-fu?).  These are the missing manuals:

```
OSSL_STORE_do_all_loaders(3)
OSSL_STORE_vctrl(3)
```

I would intuitively add `OSSL_STORE_do_all_loaders` to `doc/man3/OSSL_STORE_LOADER.pod` and `OSSL_STORE_vctrl` to `doc/man3/OSSL_STORE_open.pod`

QED

richsalz: What should existing ```L<OSSL_STORE(3)>``` resolve to?

levitte: It should *probably* be changed to `L<ossl_store(7)>`
I'm not in a position to confirm right now

richsalz:
I removed the missingdocs file and merged things into missingcrypto.
I removed references to functions that do not exist.
I fixed some typo's.
I added the "missing link" check to doc-nits (adds a couple of seconds to the build).
OSSL_STORE references should be fixed, but this is ready for a real review.

util/find-doc-nits
levitte: I just checked.  Confirmed.

richsalz: fixed.

richsalz:
Rebased to fix a conflict in util/missingcrypto.txt

richsalz:
Travis failure is a timeout.

richsalz:
ready for merge.

t8m:
Merged to master as fadb57e5 and 177b60b48

PR #10257, opened 2019-10-24
crypto/evp/evp_fetch.c: Make it more prominent that these functions are EVP

levitte:


mattcaswell (approved):
I actually think this makes things considerably clearer.

levitte:
> I actually think this makes things considerably clearer.

I was thinking that the file name `evp_fetch.c` would make the intent clear enough, but sure, no trouble.

levitte:
Merged.

181ea366f67f46cab093d6a7bbb1b2f35125b9f2 crypto/evp/evp_fetch.c: Make it more prominent that these functions are EVP

PR #10264, opened 2019-10-25
Fix find/rm command in Unix clean recipe

tanzislam:
The `./pyca-cryptography/.travis/downstream.d` subdirectory that causes the `rm` command to fail (albeit harmlessly, but with a warning from `make` nonetheless).

>rm -f \`find . -name '\*.d' \\! -name '.\*' -print\`
>rm: cannot remove './pyca-cryptography/.travis/downstream.d': Is a directory
>make: [Makefile:1910: clean] Error 1 (ignored)

Exclude directories from being matched by the `find` commands.

CLA: trivial


mattcaswell:
Pushed to master and 1.1.1. There was a trivial conflict when cherry-picking to 1.1.1 which I resolved.

Thanks for the contribution!

PR #10266, opened 2019-10-25
github: switch issue templates to new labels

mspncp:
(To be sure about the syntax, I created a [test template](https://github.com/mspncp/playground/pull/1/files) using the wizard.)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,6 +1,6 @@
 ---
 name: Bug report
-labels: bug
+labels: 'issue: bug report'
 about: Report a defect in the software
 
 ---
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,6 +1,6 @@
 ---
 name: Feature request
-labels: feature
+labels: 'issue: feature request'
 about: Propose a feature you would like to see added in the software
 
 ---


mspncp:
Merged to cfa71c5dc80c0ad783a6c4f18e10a0819eff3edf and removed the [feature] and the [bug] label.

mattcaswell:
I notice that you removed the "approval: done" label when you added "ready to merge"...is that they way we should do things - or should we keep the "approval: done" after "ready to merge" is additionally added? To me it makes more sense to keep it.

mspncp:
I wasn’t sure myself whether I should remove it or not.  If you prefer to keep it, I‘m ok with it.

mattcaswell:
To me the "approval: done" status is not something that is lost simply because we have additionally gained the "ready to merge" status.

mspncp:
Ok, then I'll restore the label on the recent tickets.

_(FWIW: bulk adding/removal of labels is possible using multiselection on the issue search page.)_

mspncp:
> To me the "approval: done" status is not something that is lost simply because we have additionally gained the "ready to merge" status.

P.S: The motivation behind removing the 'approval: done' label was to consider 'approval: *' and 'ready for merge' as mutual exclusive ticket states. But viewing them as independent properties makes equally sense.

Anyway, I think the @openssl/omc needs to state a clear policy for handling the 'approval: *' and 'ready to merge' labels, in particular in the case when the 'approval: ready' state needs to be withdrawn (e.g. because there was a force push after 'ready to merge'). Partial support by a GitHub bot might be helpful, but not sufficient for all use cases.

PR #10274, opened 2019-10-27
Fix undeclared identifier OPENSSL_ia32cap_P for e_rc4_hmac_md5.c.

kiyolee:
When compiling with VS2019, the following error occurs in e_rc4_hmac_md5.c:
1>e_rc4_hmac_md5.c
1>C:\openssl\crypto\evp\e_rc4_hmac_md5.c(90,31): error C2065: 'OPENSSL_ia32cap_P': undeclared identifier
1>C:\openssl\crypto\evp\e_rc4_hmac_md5.c(90,33): error C2109: subscript requires array or pointer type

### DIFF
--- a/crypto/evp/e_rc4_hmac_md5.c
+++ b/crypto/evp/e_rc4_hmac_md5.c
@@ -7,6 +7,7 @@
  * https://www.openssl.org/source/license.html
  */
 
+#include <internal/cryptlib.h>
 #include <openssl/opensslconf.h>
 
 #include <stdio.h>


mattcaswell:
Please either send us a CLA - details here:

Or, if you consider your contribution trivial from a copyright perspective, then update your commit to include the text "CLA: trivial" on a line on its own in the body of the commit description.

For the record I consider this a trivial contribution.

kiyolee:
CLA:trivial

mspncp:
> @mattcaswell : is the cla required label to be removed manually in this case ?

We usuallly reopen the ticket to kick the cla-bot.

mspncp:
The bot still complains because only one of the three commits has the 'CLA: trivial' label. But that can trivially be resolved when squashing the pr. So well, let's just remove the label manually...

p-steuer:
I noticed the bot still complaining and assumed it was because of the missing space ie "CLA:trivial" instead of "CLA: trivial". Could be resolved anyway.

mspncp:
You’re right, it’s in the subject line.

kiyolee:
It is interesting that I did the same "git commit --ammend" to another PR and added a single line "CLA:trivial" (no space) to the comment and that one worked nicely. (Although it is a bit puzzling for me why that ended up 2 extra commits.) Anyway, the bot automatically fixed (so it seemed) the cla-check for that other PR but not for this one.
Was "git commit --ammend" not the right thing to do?

mattcaswell:
You seem to have added a merge commit as result of pulling master. We don't actually allow those - always make sure you use `git pull --rebase` when pulling a fresh master. That is where the confusion has come from I think.

p-steuer:
Looks like its only needed for master.

p-steuer:
merged, thanks.

master: https://github.com/openssl/openssl/commit/f702256d5514e54138279fd4637353e328262817

PR #10275, opened 2019-10-27
Fix simple typo in cipher_chacha20_poly1305_hw.c

kiyolee:
Compile error occurs at:
1>cipher_chacha20_poly1305_hw.c
1>C:\openssl\providers\implementations\ciphers\cipher_chacha20_poly1305_hw.c(145,22): error C2039: 'enc': is not a member of 'PROV_CHACHA20_POLY1305_CTX'
1>C:\openssl\providers\implementations\ciphers\cipher_chacha20_poly1305.h(18): message : see declaration of 'PROV_CHACHA20_POLY1305_CTX'

### DIFF
--- a/providers/implementations/ciphers/cipher_chacha20_poly1305_hw.c
+++ b/providers/implementations/ciphers/cipher_chacha20_poly1305_hw.c
@@ -142,7 +142,7 @@ static int chacha20_poly1305_tls_cipher(PROV_CIPHER_CTX *bctx,
         ctx->len.text = plen;
 
         if (plen) {
-            if (ctx->base.enc)
+            if (bctx->enc)
                 ctr = xor128_encrypt_n_pad(out, in, ctr, plen);
             else
                 ctr = xor128_decrypt_n_pad(out, in, ctr, plen);
--- a/providers/implementations/ciphers/cipher_chacha20_poly1305_hw.c
+++ b/providers/implementations/ciphers/cipher_chacha20_poly1305_hw.c
@@ -142,7 +142,7 @@ static int chacha20_poly1305_tls_cipher(PROV_CIPHER_CTX *bctx,
         ctx->len.text = plen;
 
         if (plen) {
-            if (ctx->enc)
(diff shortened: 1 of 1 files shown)


mattcaswell:
Please either send us a CLA - details here:

Or, if you consider your contribution trivial from a copyright perspective, then update your commit to include the text "CLA: trivial" on a line on its own in the body of the commit description.

For the record I consider this a trivial contribution.

providers/implementations/ciphers/cipher_chacha20_poly1305_hw.c
```diff
@@ -142,7 +142,7 @@ static int chacha20_poly1305_tls_cipher(PROV_CIPHER_CTX *bctx,
         ctx->len.text = plen;
 
         if (plen) {
-            if (ctx->enc)
+            if (bctx->enc)
```
levitte: The other option is `ctx->base.enc`, which is the exact same thing, but is more in style with the rest of the function.

I leave it to you to decide.

kiyolee:
CLA:trivial

providers/implementations/ciphers/cipher_chacha20_poly1305_hw.c
kiyolee: ctx->base.enc was my first attempt. Then I found bctx->... are used in multiple places and thought that seems cleaner.

levitte: Okie

mattcaswell:
@levitte - for the record can you confirm you consider this trivial?

Whoever merges will need to remove the merge commit, and squash all other commits.

levitte:
> @levitte - for the record can you confirm you consider this trivial?

Yes

> Whoever merges will need to remove the merge commit, and squash all other commits.

A rebase upon a fresh master makes any merge commit go away.

mattcaswell:
Pushed. Thanks for the contribution!

PR #10276, opened 2019-10-28
Use large enough buffer for signature in dgst.c

### LINKED ISSUE #9732: Unable to use custom digital signature algorithm with openssl dgst command
<!--
Thank you for your bug report.

NOTE:

    If you're asking about how to use OpenSSL, this isn't the right 
    forum.  Please see our User Support resources:
    https://github.com/openssl/openssl/blob/master/.github/SUPPORT.md

Please remember to tell us in what OpenSSL version you found the issue.

For build issues:

    If this is a build issue, please include the configuration output
    as well as a log of all errors.  Don't forget to include the exact
    commands you typed.

    With OpenSSL before 1.1.1, the configuration output comes from the
    configuration command.  With OpenSSL 1.1.1 and on, it's the output
    of `perl configdata.pm --dump`

For other issues:

    If it isn't a build issue, example code or commands to reproduce
    the issue is highly appreciated.
    Also, please remember to tell us if you worked with your own
    OpenSSL build or if it is system provided.

Please remember to put ``` lines before and after any commands plus
output and code, like this:

    ```
    $ echo output output output
    output output output
    ```

    ```
    #include <stdio.h>
    
    int main() {
        int foo = 1;
        printf("%d\n", foo);
    }
    ```
-->

For the following command line:

```
echo 'Hello world' > sign.txt
openssl dgst -engine <engine_name> -sha256 -sign CA.key -out sign.txt.sha256 sign.txt
```

I got the following output:

```
Error Signing Data
139681221683008:error:0608C09B:digital envelope routines:EVP_PKEY_sign:buffer too small:crypto/evp/pmeth_fn.c:65
```
Used signature schemes signature size is 49216 bytes (and EVP_PKEY_size returns this size).

After debugging session I discovered, that siglen is hardcoded at: `apps/dgst.c:453`

```c
if (key != NULL) {                                                                                                                                                                                                                                
          EVP_MD_CTX *ctx;                                                                                                                                                                                                                              
          BIO_get_md_ctx(bp, &ctx);                                                                                                                                                                                                                     
          len = BUFSIZE;  // <----- HARDCODE                                                                                                                                                                                                                               
          if (!EVP_DigestSignFinal(ctx, buf, &len)) {                                                                                                                                                                                                   
              BIO_printf(bio_err, "Error Signing Data\n");                                                                                                                                                                                              
              ERR_print_errors(bio_err);                                                                                                                                                                                                                
              return 1;                                                                                                                                                                                                                                 
          }                                                                                                                                                                                                                                             
      }
```

Is that an issue of my engines implementation? What should exactly return `EVP_PKEY_size` signature size or maximum digest size?

nblka:
Fixes #9732

### DIFF
--- a/apps/dgst.c
+++ b/apps/dgst.c
@@ -535,8 +535,9 @@ int do_fp(BIO *out, unsigned char *buf, BIO *bp, int sep, int binout,
     }
     if (key != NULL) {
         EVP_MD_CTX *ctx;
+        int pkey_len;
         BIO_get_md_ctx(bp, &ctx);
-        int pkey_len = EVP_PKEY_size(key);
+        pkey_len = EVP_PKEY_size(key);
         if (pkey_len > BUFSIZE) {
             len = pkey_len;
             sigbuf = app_malloc(len, "Signature buffer");
--- a/apps/dgst.c
+++ b/apps/dgst.c
@@ -502,7 +502,7 @@ int do_fp(BIO *out, unsigned char *buf, BIO *bp, int sep, int binout,
           const char *sig_name, const char *md_name,
           const char *file)
 {
-    size_t len;
+    size_t len = BUFSIZE;
     int i, backslash = 0, ret = 1;
     unsigned char *sigbuf = NULL;
 
@@ -542,8 +542,6 @@ int do_fp(BIO *out, unsigned char *buf, BIO *bp, int sep, int binout,
             len = pkey_len;
             sigbuf = app_malloc(len, "Signature buffer");
             buf = sigbuf;
-        } else {
-            len = BUFSIZE;
         }
         if (!EVP_DigestSignFinal(ctx, buf, &len)) {
             BIO_printf(bio_err, "Error Signing Data\n");
--- a/apps/dgst.c
+++ b/apps/dgst.c
@@ -503,14 +503,15 @@ int do_fp(BIO *out, unsigned char *buf, BIO *bp, int sep, int binout,
           const char *file)
 {
     size_t len;
-    int i, backslash = 0;
+    int i, backslash = 0, ret = 1;
+    unsigned char *sigbuf = NULL;
 
     while (BIO_pending(bp) || !BIO_eof(bp)) {
         i = BIO_read(bp, (char *)buf, BUFSIZE);
         if (i < 0) {
             BIO_printf(bio_err, "Read Error in %s\n", file);
             ERR_print_errors(bio_err);
-            return 1;
+            goto end;
         }
         if (i == 0)
             break;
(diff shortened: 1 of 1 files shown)


apps/dgst.c
```diff
@@ -523,28 +524,35 @@ int do_fp(BIO *out, unsigned char *buf, BIO *bp, int sep, int binout,
...
+        ret = 0;
+        goto end;
     }
     if (key != NULL) {
         EVP_MD_CTX *ctx;
+        int pkey_len;
         BIO_get_md_ctx(bp, &ctx);
-        len = BUFSIZE;
+        pkey_len = EVP_PKEY_size(key);
+        if (pkey_len > BUFSIZE) {
```
mattcaswell: Do we actually have any signature schemes that exceed this limit? Or have you hit this with an engine supplied signature scheme?

apps/dgst.c
```diff
@@ -579,5 +587,11 @@ int do_fp(BIO *out, unsigned char *buf, BIO *bp, int sep, int binout,
         }
         BIO_printf(out, "\n");
     }
-    return 0;
+
+    ret = 0;
+ end:
+    if (NULL != sigbuf) {
```
mattcaswell: Normally, we tend to do this the other way around, i.e. `sigbuf != NULL`

apps/dgst.c
```diff
@@ -579,5 +587,11 @@ int do_fp(BIO *out, unsigned char *buf, BIO *bp, int sep, int binout,
...
+        OPENSSL_clear_free(sigbuf, len);
+    }
```
mattcaswell: Nit: Don't use {} for a single line if.

mattcaswell:
Close/reopen to kick cla bot

apps/dgst.c
mkwork: Such practice could avoid misspelling:
e.g:

```c
int a = 42;

if (a = 3) 
   abort(); // you would always abort

if (3 == a) // you would got compilation error here
   abort();
```

p-steuer: on the other hand its more difficult to read (at least for lt gt) and afaik most modern compilers would emit a warning telling you to put additional braces around the assignment to indicate that is really what you want to do.

apps/dgst.c
nblka: This PR is to support our custom engine based signature scheme with large signature sizes. It is originated from #9732. As you commented this issue, PRs are welcome, so I prepared this PR. Hope you'll find it useful. I tried to change existing behavior as little as possible. The new buffer will not be allocated if the existing buffer passed from `dgst_main` is enough to store.

apps/dgst.c
nblka: Thank you, I've updated this line

apps/dgst.c
nblka: Got it, fixed

apps/dgst.c
mattcaswell: No. This is fine as is. I added the labels for the target branches.

p-steuer:
The commit message reads:
``` 
Allocate and use new signature buffer with desired length if existing…

… buffer is not large enough

Fixes #9732
```
Since this description is very general, would you ( @nblka ) mind if it was changed to something along the lines of
```
apps/dgst.c: allocate a new signature buffer

... if the fixed-size buffer is too small. 

Fixes #9732
```
when squashed ?

nblka:
@p-steuer no, feel free to change it

p-steuer:
merged, thanks.

master: https://github.com/openssl/openssl/commit/7c2d95d47ccb3797f0da6bd4446747c6eee07b87
1.1.1: https://github.com/openssl/openssl/commit/43a8f91f00e35b20e33c393f5b79215c277c508e

PR #8321, opened 2019-02-25
Support SM2 certificate verification

InfoHunter:
PR #6837 stalled for a long time so I open this PR to add the support of SM2 certificate verification. The main different between this PR and #6837 is I added the ability of specifying the SM2 ID rather than hardcoding it into "1234567812345678".

Several changes made to X509 code to achieve this, including an extra field carrying the ID value was added to X509 structure and the apps/verify command was also changed to let user set the ID (although the user is limited to only specify an ASCII string).

The usage of the `verify` app is something as this:

`openssl verify -sm2-id 1234567812345678 -CAfile sm2-ca.crt sm2.crt`

This is WIP status, so doc and test case are not updated yet. Please review this.

##### Checklist

- [ ] documentation is added or updated
- [ ] tests are added or updated

PR #10278, opened 2019-10-28
Suppress an error when doing SM2 sign/verify ops

InfoHunter:
This was fixed in #8321 right after the 1.1.1 was released but never
back ported to 1.1.1. Now fix it.

Issue reported from lua-openssl project.


InfoHunter:
Hope I have added to right labels according to what was discussed on last week's f2f meeting ;-)

InfoHunter:
Pushed to 1.1.1 as 325c9ac198, thanks.

PR #10292, opened 2019-10-29
evp_pkey_ctx_free_old_ops(): Make sure to assign NULL to freed pointers

levitte:
Otherwise, should this function be called more than once on the same
EVP_PKEY_CTX, we get double free issues.


t8m (approved):

levitte:
Merged.

fb1ecf85c9f732e5827771ff243d7a70e06ce112 evp_pkey_ctx_free_old_ops(): Make sure to assign NULL to freed pointers

PR #10293, opened 2019-10-30
KDF: use string names not macros to align with other algorithms.

paulidale:
Only the KDF and PRF algorithms used the macros for their names, all other
algorithms used a string name directly.  This brings the KDFs and PRFs into
line with the rest.

This change is cosmetic only.

### DIFF
--- a/providers/defltprov.c
+++ b/providers/defltprov.c
@@ -326,21 +326,21 @@ static const OSSL_ALGORITHM deflt_macs[] = {
 };
 
 static const OSSL_ALGORITHM deflt_kdfs[] = {
-    { OSSL_KDF_NAME_HKDF, "default=yes", kdf_hkdf_functions },
-    { OSSL_KDF_NAME_SSKDF, "default=yes", kdf_sskdf_functions },
-    { OSSL_KDF_NAME_PBKDF2, "default=yes", kdf_pbkdf2_functions },
-    { OSSL_KDF_NAME_SSHKDF, "default=yes", kdf_sshkdf_functions },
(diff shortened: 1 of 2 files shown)


levitte:
The reason those macros exist at all is because we use those names in libcrypto, either to fetch an algo implementation, or to check that a method does in fact implement that algo (using `EVP_(WHATEVER) _is_a()`), i.e to avoid the "magic strings" that some seem to dislike so much, and it seemed wise to use the same macro in our OSSL_ALGORITHM tables when available.

This is food for thought, I remain neutral as to how we should do this in the end.
(and yeah, the thought occurred to me that we could have macros for *all* the names. I don't see how that's necessary for single use, though)

paulidale:
i would be fine going the other way and using the macros for each implementation.  The problem would be with the aliases which would remain text.

levitte:
That's not much of a problem, technically speaking...

t8m:
The only advantage of macros in this case that I can see is the higher probability of eventual typo being found already during the compilation and not only later during the build-time tests. Otherwise the long macro names seem to me to be rather a code-bloat than anything else when compared to the simple canonical algorithm names in string literals.

paulidale:
@t8m I tend to agree, basic quoted strings would be nicer in many way.  The only advantage of macros is their compile time checking.

levitte:
This PR is approved, and while we have commented, none of the comments say "do not do this", so it looks like it will be merged in due time.

paulidale:
Merged to master, thanks for the input.

PR #10298, opened 2019-10-30
Remove potential usage of CRT object across DLL boundaries

longagofaraway:
If we compile shared libcrypto and openssl tool with static runtime
on windows, we'll be using libcrypto's CRT object in another instance
of CRT (the one in the tool). This commit fixes it.

CLA: trivial

### DIFF
--- a/apps/lib/apps.c
+++ b/apps/lib/apps.c
@@ -1382,7 +1382,6 @@ CA_DB *load_index(const char *dbfile, DB_ATTR *db_attr)
     CONF *dbattr_conf = NULL;
     char buf[BSIZE];
 #ifndef OPENSSL_NO_POSIX_IO
-    FILE *dbfp;
     struct stat dbst;
 #endif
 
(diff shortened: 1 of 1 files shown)


levitte:
I think you haven't understood what `OPENSSL_NO_POSIX_IO` means. `stat()` is, to the best of our judgment, a POSIX IO function.

longagofaraway:
Sure it is. The problem is that the handler, that fstat accepts, is a C runtime object on windows.

levitte:
Ah, gotcha, and since we're using DLL code to get the handle, it misses the target, because it will give a DLL-specific handle instead of the application specific one.  This is usually handled with UpLink / AppLink, but I see that it's not usable here.

We should really have a BIO specific stat() function that just does the right thing...

levitte:
... that is to say that yes, you did indeed find a bug, but there are better solutions.

levitte:
(I've assigned myself this bug, because it aligns with some thoughts I have on revamping / completing the BIO API...  it's seriously lacking in some areas)

t8m:

PR #10300, opened 2019-10-30
Don't assume the type we read was the type we expected

mattcaswell:
`i2v_GENERAL_NAME` and `GENERAL_NAME_print` were assuming that the type of of a `GENERAL_NAME` (`OTHERNAME`) that we read in was the type we expected it to be. If its something else then this can cause unexpected behaviour. In the added fuzz test case an OOB read was occurring.

This issue was recently added by commit 4baee2d.

Credit to OSSFuzz for finding this issue.

In fixing this problem I also ran into a mem leak on a failure path in `i2v_GENERAL_NAMES` which I have also fixed.

The first commit (mem leak fix) is for both master and 1.1.1. The second commit is for master only.


mattcaswell:
@beldmit

beldmit:
In the original report, we had a NULL-pointer dereference at 

```BIO_printf(out, "othername:UPN:%s", gen->d.otherName->value->value.utf8string->data);```

`(gdb) p gen->d.otherName->value->value.utf8string
$1 = (ASN1_UTF8STRING *) 0x0`

What happens when we have a valid type and NULL here? Or we can't get such a situation?

**Comment by beldmit** [comment]
No crash now on Kurt's examples.

File: crypto/x509/v3_alt.c [hunk 1]
```diff
@@ -161,21 +186,36 @@ STACK_OF(CONF_VALUE) *i2v_GENERAL_NAME(X509V3_EXT_METHOD *method,
...
+        switch (nid) {
         case NID_id_on_SmtpUTF8Mailbox:
-            BIO_printf(out, "othername:SmtpUTF8Mailbox:%s", gen->d.otherName->value->value.utf8string->data);
+            BIO_printf(out, "othername:SmtpUTF8Mailbox:%s",
+                       gen->d.otherName->value->value.utf8string->data);
             break;
         case NID_XmppAddr:
-            BIO_printf(out, "othername:XmppAddr:%s", gen->d.otherName->value->value.utf8string->data);
+            BIO_printf(out, "othername:XmppAddr:%s",
+                       gen->d.otherName->value->value.utf8string->data);
```
vdukhovni: This isn't right, srvName is an "IA5String".

mattcaswell:
Pushed to master, and the first commit additionally to 1.1.1.

PR #10152, opened 2019-10-11
Implement asymmetric cipher support

mattcaswell:
This PR implements asymmetric cipher support for providers, and moves that aspect of RSA into the default provider.

This is WIP because there are a number of outstanding issues:

- Test failure in CMS for unknown reasons
- There's a weird memory issue I need to investigate
- It currently doesn't merge with master because of @levitte moving everything around under my feet :-)
- Needs some documentation updates

PR #10303, opened 2019-10-30
EVP: Make the SIGNATURE implementation leaner

levitte:
Because the algorithm to use is decided already when creating an
EVP_PKEY_CTX regardless of how it was created, it turns out that it's
unnecessary to provide the SIGNATURE method explicitly, and rather
always have it be fetched implicitly.

This means fewer changes for applications that want to use new
signature algorithms / implementations.

### DIFF
--- a/crypto/evp/m_sigver.c
+++ b/crypto/evp/m_sigver.c
@@ -44,24 +44,38 @@ static int do_sigver_init(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
         ctx->provctx = NULL;
     }
 
-    if (ctx->pctx == NULL) {
+    if (ctx->pctx == NULL)
         ctx->pctx = EVP_PKEY_CTX_new(pkey, e);
-        if (ctx->pctx == NULL)
-            return 0;
-    } else if (pkey != NULL) {
-        if (!EVP_PKEY_up_ref(pkey))
-            return 0;
-        EVP_PKEY_free(ctx->pctx->pkey);
-        ctx->pctx->pkey = pkey;
-    }
+    if (ctx->pctx == NULL)
+        return 0;
+
     locpctx = ctx->pctx;
     evp_pkey_ctx_free_old_ops(locpctx);
-    if (locpctx->pkey == NULL)
+
+    if (locpctx->algorithm == NULL)
         goto legacy;
 
-    if (e != NULL || locpctx->engine != NULL)
-        goto legacy;
+    if (mdname == NULL) {
+        if (type != NULL) {
+            mdname = EVP_MD_name(type);
+        } else if (pkey != NULL) {
+            /*
+             * TODO(v3.0) work out a better way for EVP_PKEYs with no legacy
+             * component.
+             */
+            if (pkey->pkey.ptr != NULL) {
+                int def_nid;
+                if (EVP_PKEY_get_default_digest_nid(pkey, &def_nid) > 0)
+                    mdname = OBJ_nid2sn(def_nid);
+            }
+        }
+    }
 
+    /*
+     * Because we cleared out old ops, we shouldn't need to worry about
+     * checking if signature is already there.  Keymgmt is a different
+     * matter, as it isn't tied to a specific EVP_PKEY op.
+     */
     signature =
         EVP_SIGNATURE_fetch(NULL, locpctx->algorithm, locpctx->propquery);
     if (signature != NULL && locpctx->keymgmt == NULL) {
--- a/crypto/evp/m_sigver.c
+++ b/crypto/evp/m_sigver.c
@@ -31,7 +31,6 @@ static int do_sigver_init(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
 {
     EVP_PKEY_CTX *locpctx = NULL;
     EVP_SIGNATURE *signature = NULL;
-    EVP_KEYMGMT *keymgmt = NULL;
     void *provkey = NULL;
     int ret;
 
@@ -68,25 +67,24 @@ static int do_sigver_init(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
     if (signature != NULL && locpctx->keymgmt == NULL) {
         int name_id = EVP_SIGNATURE_number(signature);
 
-        keymgmt =
+        locpctx->keymgmt =
             evp_keymgmt_fetch_by_number(NULL, name_id, locpctx->propquery);
     }
 
-    if (keymgmt == NULL
+    if (locpctx->keymgmt == NULL
         || signature == NULL
-        || (EVP_KEYMGMT_provider(keymgmt)
+        || (EVP_KEYMGMT_provider(locpctx->keymgmt)
             != EVP_SIGNATURE_provider(signature))) {
         /*
          * We don't have the full support we need with provided methods,
-         * let's go see if legacy does
+         * let's go see if legacy does.  Also, we don't need to free
+         * ctx->keymgmt here, as it's not necessarily tied to this
+         * operation.  It will be freed by EVP_PKEY_CTX_free().
          */
-        EVP_KEYMGMT_free(keymgmt);
         EVP_SIGNATURE_free(signature);
         goto legacy;
     }
 
-    if (keymgmt != NULL)
-        locpctx->keymgmt = keymgmt;
     locpctx->op.sig.signature = signature;
 
     locpctx->operation = ver ? EVP_PKEY_OP_VERIFYCTX
--- a/crypto/evp/m_sigver.c
+++ b/crypto/evp/m_sigver.c
@@ -76,13 +76,14 @@ static int do_sigver_init(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
      * checking if signature is already there.  Keymgmt is a different
      * matter, as it isn't tied to a specific EVP_PKEY op.
      */
-    signature =
-        EVP_SIGNATURE_fetch(NULL, locpctx->algorithm, locpctx->propquery);
+    signature = EVP_SIGNATURE_fetch(locpctx->libctx, locpctx->algorithm,
+                                    locpctx->propquery);
     if (signature != NULL && locpctx->keymgmt == NULL) {
         int name_id = EVP_SIGNATURE_number(signature);
 
         locpctx->keymgmt =
-            evp_keymgmt_fetch_by_number(NULL, name_id, locpctx->propquery);
+            evp_keymgmt_fetch_by_number(locpctx->libctx, name_id,
+                                        locpctx->propquery);
     }
 
     if (locpctx->keymgmt == NULL
--- a/crypto/evp/m_sigver.c
+++ b/crypto/evp/m_sigver.c
@@ -129,11 +129,8 @@ static int do_sigver_init(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
          * man pages), i.e. the ref count is not updated so the EVP_MD should
          * not be used beyound the lifetime of the EVP_MD_CTX.
          */
-        OSSL_PROVIDER *sigprov = EVP_SIGNATURE_provider(signature);
-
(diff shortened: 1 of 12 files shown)


levitte:
@mattcaswell, I suspect you're keenly interested.  This shows how to do what I suggested in #10152

crypto/evp/m_sigver.c
```diff
@@ -61,41 +63,43 @@ static int do_sigver_init(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
...
+    signature =
+        EVP_SIGNATURE_fetch(NULL, locpctx->algorithm, locpctx->propquery);
+    if (signature != NULL && locpctx->keymgmt == NULL) {
+        int name_id = EVP_SIGNATURE_number(signature);
+
+        keymgmt =
+            evp_keymgmt_fetch_by_number(NULL, name_id, locpctx->propquery);
+    }
+
+    if (keymgmt == NULL
```
mattcaswell: Can it be the case that `locpctx->keymgmt != NULL` but `keymgmt == NULL`? In which case do we need to check `locpctx->keymgmt` here?

crypto/evp/pmeth_fn.c
```diff
@@ -360,41 +348,41 @@ static int evp_pkey_signature_init(EVP_PKEY_CTX *ctx, EVP_SIGNATURE *signature,
...
+     * matter, as it isn't tied to a specific EVP_PKEY op.
+     */
+    signature = EVP_SIGNATURE_fetch(NULL, ctx->algorithm, ctx->propquery);
+    if (signature != NULL && ctx->keymgmt == NULL) {
+        int name_id = EVP_SIGNATURE_number(signature);
+
+        keymgmt = evp_keymgmt_fetch_by_number(NULL, name_id, ctx->propquery);
+    }
 
+    if (keymgmt == NULL
```
mattcaswell: Again here do we need to check `ctx->keymgmt`?

mattcaswell (changes requested):
I think there will need to be some documentation changes where you have changed the parameters for public API functions.

mattcaswell:
Travis errors look relevant

crypto/evp/m_sigver.c
levitte: Hmmm... You're right, something is off. Not sure that your suggestion is right, though. Gimme a moment to think it over.

levitte:
One thing i found disturbing, btw, is that some of the code in `m_sigver.c` is *identical* to the code in `pmeth_fn.c`...  is there no way to refactor that?

levitte:
Documentation, thanks for the reminder!

levitte:
I found that I had to do more rework in `m_sigver.c`, it's quite a mess.  I mess I can understand, give the conditions when this was first developed, but that takes a little bit to untangle now that other parts of the EVP_PKEY and EVP_PKEY_CTX API is a bit saner / leaner.

levitte:
This should be complete enough by now....  or so I hope

levitte:
Ping

mattcaswell:
Travis failures look relevant:

````
    # [12:18:44]  1484 file=crypto/threads_pthread.c, line=33, thread=00:37:0B:4D:76:7F:00:00, number=56, address=0x18f9c10
    # [12:18:44]  1343 file=crypto/provider_core.c, line=209, thread=00:37:0B:4D:76:7F:00:00, number=8, address=0x18e6ab0
    # [12:18:44]  1342 file=crypto/provider_core.c, line=204, thread=00:37:0B:4D:76:7F:00:00, number=128, address=0x18e69c0
    # [12:18:44]  1483 file=crypto/evp/digest.c, line=709, thread=00:37:0B:4D:76:7F:00:00, number=216, address=0x18f9ad0
    # 408 bytes leaked in 4 chunks
/home/travis/build/openssl/openssl/util/shlib_wrap.sh /home/travis/build/openssl/openssl/test/evp_extra_test => 1
````

File: crypto/evp/m_sigver.c [hunk 3]
```diff
@@ -43,59 +44,74 @@ static int do_sigver_init(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
...
+    if (ctx->pctx == NULL)
         ctx->pctx = EVP_PKEY_CTX_new(pkey, e);
-        if (ctx->pctx == NULL)
-            return 0;
-    } else if (pkey != NULL) {
-        if (!EVP_PKEY_up_ref(pkey))
-            return 0;
-        EVP_PKEY_free(ctx->pctx->pkey);
-        ctx->pctx->pkey = pkey;
-    }
```
mattcaswell: I'm confused by the removal of this `else if` block. Why is this no longer needed?

mattcaswell (changes requested):

romen:
Given that this is modifying `m_sigver.c` would it make sense to rebase it on top of #10312 and fix the issue?

Here is how it is related: https://github.com/openssl/openssl/issues/10224#issuecomment-547949149

crypto/evp/m_sigver.c
levitte: It seriously changes the semantics.  Here's what the 1.1.1 code does:

https://github.com/openssl/openssl/blob/689c07b8d502301d82e09c1db04daf3c13b1d35d/crypto/evp/m_sigver.c#L28-L45

So `pkey` is *only* assigned to the PKEY context if it's created here, otherwise, it's *only* used to set `type` if that one is NULL, and only under specific conditions.

Spuriously assigning an *already existing* PKEY context requires more work (there's otherwise a great risk of a mismatch between the new key and the rest of the PKEY context data), to the point that it's better to simply free the context and re-create it, and I think that would be a sore surprise for the caller.

levitte: Mind you, if you look below, I'm using `type` or `pkey` to set `mdname` if the latter is NULL without looking at more conditions...  that broadens the conditions to set the MD type internally, but I think that's within safe limits.

crypto/evp/m_sigver.c
```diff
@@ -43,59 +44,74 @@ static int do_sigver_init(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
...
+             */
+            if (pkey->pkey.ptr != NULL) {
+                int def_nid;
+                if (EVP_PKEY_get_default_digest_nid(pkey, &def_nid) > 0)
+                    mdname = OBJ_nid2sn(def_nid);
+            }
         }
-        signature
-            = EVP_SIGNATURE_fetch(NULL, OBJ_nid2sn(locpctx->pkey->type), NULL);
+    }
```
romen: Do we need to handle the cases in which after this block mdname is still null?

E.g. mdname, type and pkey are NULL or no default digest nid, etc.

levitte: Ah, good question...  If I understand correctly, we've left that for the provided init functions to decide.  @mattcaswell, please confirm.

Originally, I think a NULL `type` was allowed to accommodate contraptions like MACs implemented as PKEY signature methods, which can obviously take any size input.  However, that doesn't exclude the possibility that some public key algos might be able to sign any size data without an intermediary hash function, now or in the future, so it may still be wise to let the provided init functions decide.  After all, the EVP API is supposed to be a fairly thin layer, let's try and keep it that way.

levitte: However, I just noticed that `mdname` is processed a second time in lines 120+.  I need to have a closer look at that.

crypto/evp/m_sigver.c
mattcaswell: > It seriously changes the semantics. 

Looking at the history of this I have no idea why I added that. So, ok.

crypto/evp/m_sigver.c
mattcaswell: This function is called as a result of the application calling `EVP_DigestSignInit` or `EVP_DigestSignInit_ex`. In the former case we pass through the EVP_MD from the application in the `type` parameter. Otherwise we pass through the `mdname`. So, normally, one or the other is set. The only case where they are both NULL would be if the application explicitly passed a NULL - in which case we should just pass that through to the provider to interpret.

mattcaswell: > However, I just noticed that mdname is processed a second time in lines 120+. I need to have a closer look at that.

That second processing doesn't quite look right...

levitte: > That second processing doesn't quite look right...

It isn't right at all...  it's a remnant that I missed.  I'll fix.

crypto/evp/m_sigver.c
```diff
@@ -43,59 +44,74 @@ static int do_sigver_init(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
...
-            ERR_raise(ERR_LIB_EVP, EVP_R_INITIALIZATION_ERROR);
-            goto err;
-        }
+    /*
+     * Because we cleared out old ops, we shouldn't need to worry about
+     * checking if signature is already there.  Keymgmt is a different
+     * matter, as it isn't tied to a specific EVP_PKEY op.
+     */
+    signature =
+        EVP_SIGNATURE_fetch(NULL, locpctx->algorithm, locpctx->propquery);
```
mattcaswell: Since this is now the only way we can fetch signatures, for a digest sign operation, the initial NULL looks wrong. We should be able to specify a `libctx` somehow.

levitte: Hmmmmm....  so, a quick idea top of my head, should `EVP_PKEY_CTX_new_provided()` take a library context pointer along with the algorithm name and property query string?  In all other case, we'll have to pilfer from whatever we can...

mattcaswell: That was my immediate thought, yes.

levitte: Okie...  pondering if that should really be part of this PR, or an addition to #10308.  The latter seems a bit fitting, no?

mattcaswell: Makes sense

levitte: Cool.  I'll make the appropriate changes here and mark this pending on #10308

levitte:
This will not build right before #10308 is merged

util/libcrypto.num
```diff
@@ -4742,10 +4742,10 @@ EVP_SIGNATURE_free                      4858	3_0_0	EXIST::FUNCTION:
 EVP_SIGNATURE_up_ref                    4859	3_0_0	EXIST::FUNCTION:
 EVP_SIGNATURE_provider                  4860	3_0_0	EXIST::FUNCTION:
 EVP_SIGNATURE_fetch                     4861	3_0_0	EXIST::FUNCTION:
-EVP_PKEY_sign_init_ex                   4862	3_0_0	EXIST::FUNCTION:
+EVP_PKEY_sign_init_ex                   4862	3_0_0	NOEXIST::FUNCTION:
```
kroeckx: Is there a reason to mark it as NOEXIST rather than just delete it?

levitte: It's what the updating scripts do.  It's what they have always done.  I don't *really* know why, possibly for reasons of uncertainty, in case the symbol would come back with the same meaning.

I'm putting that down to "tradition" :wink:

kroeckx: It's not like there ever was a release with them

levitte: That's quite true.  But that's for another PR.  I have this action item on me to change the script so it only assigns numbers at release time (I'm going make the number marker a simple question mark).  At that point, it becomes much easier to automatically remove symbols that disappear and haven't yet been released.  So if you don't mind, save this discussion for then, yeah?

kroeckx: sure

levitte:
#10308 is now in, and this is now rebased

romen:
The `doc-nits` issue seems relevant

levitte:
Yup, I forgot to fix the docs.  Done now.

mattcaswell (approved):

levitte:
Merged.

0e52100400e647aeb5b8ac1b92c74d53918d39a0 EVP: Make the SIGNATURE implementation leaner
c9c4a356b778d042fee3d86f7178012ec0077205 Minimal adaptation of tests back to how it was before

PR #10305, opened 2019-10-30
EVP: Make the KEYEXCH implementation leaner

levitte:
Because the algorithm to use is decided already when creating an
EVP_PKEY_CTX regardless of how it was created, it turns out that it's
unnecessary to provide the KEYEXCH method explicitly, and rather
always have it be fetched implicitly.

This means fewer changes for applications that want to use new key
exchange algorithms / implementations.

### DIFF
--- a/doc/man3/EVP_PKEY_derive.pod
+++ b/doc/man3/EVP_PKEY_derive.pod
@@ -2,43 +2,34 @@
 
 =head1 NAME
 
-EVP_PKEY_derive_init, EVP_PKEY_derive_init_ex, EVP_PKEY_derive_set_peer,
-EVP_PKEY_derive - derive public key algorithm shared secret
+EVP_PKEY_derive_init, EVP_PKEY_derive_set_peer, EVP_PKEY_derive
+- derive public key algorithm shared secret
 
 =head1 SYNOPSIS
 
  #include <openssl/evp.h>
 
- int EVP_PKEY_derive_init_ex(EVP_PKEY_CTX *ctx, EVP_KEYEXCH *exchange);
  int EVP_PKEY_derive_init(EVP_PKEY_CTX *ctx);
  int EVP_PKEY_derive_set_peer(EVP_PKEY_CTX *ctx, EVP_PKEY *peer);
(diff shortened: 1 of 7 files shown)


levitte:
This turned out to be much much easier than #10303...  possibly because I had already put in the work in #10303 and this practically became just a copy-n-paste job

mattcaswell:
I'll hold back reviewing this one for a bit until #10303 is a little further along. Most of the same comments from there apply here too.

levitte:
Yup

levitte:
I hope this is complete enough now

levitte:
Ping

levitte:
This will not build right before #10308 is merged

levitte:
#10308 is now in, and this is now rebased

doc/man3/EVP_PKEY_derive.pod
```diff
@@ -2,43 +2,34 @@
...
 The EVP_PKEY_derive_set_peer() function sets the peer key: this will normally
 be a public key.
 
-The EVP_PKEY_derive() derives a shared secret using B<ctx>.
-If B<key> is B<NULL> then the maximum size of the output buffer is written to
-the B<keylen> parameter. If B<key> is not B<NULL> then before the call the
-B<keylen> parameter should contain the length of the B<key> buffer, if the call
-is successful the shared secret is written to B<key> and the amount of data
-written to B<keylen>.
+The EVP_PKEY_derive() derives a shared secret using I<ctx>.
```
mattcaswell: Either `The EVP_PKEY_derive() function derives...` or just `EVP_PKEY_derive() derives...` - but not as it is at the moment (an extra "The").

mattcaswell (approved):
Approved assuming the documentation nit is fixed.

doc/man3/EVP_PKEY_derive.pod
levitte: You're right. I will fix that

levitte:
Merged.

c0e0984f125ee30820520829c4c112b1e166f871 EVP: Make the KEYEXCH implementation leaner

PR #10308, opened 2019-10-31
Make EVP_PKEY_CTX initialization more precise

levitte:
There is a vagueness around how the provider data (algorithm name and
property query string) is initialized in the presence of an engine.
This change modifies this slightly so that the algorithm name for use
with providers is never set if the initilization was given an engine.

This makes it easier for other functions to simply check ctx->algorithm
to see if the context is meant to be used for strictly legacy stuff or
not.

Additionally, we change `EVP_PKEY_CTX_new_provided()` to take a library
context alongside the algorithm name and property query string.

### DIFF
--- a/crypto/evp/pmeth_lib.c
+++ b/crypto/evp/pmeth_lib.c
@@ -111,7 +111,8 @@ const EVP_PKEY_METHOD *EVP_PKEY_meth_find(int type)
     return (**ret)();
 }
 
-static EVP_PKEY_CTX *int_ctx_new(EVP_PKEY *pkey, ENGINE *e,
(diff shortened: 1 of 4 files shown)


mattcaswell (dismissed):

levitte:
Please review again

mattcaswell (approved):
LGTM

levitte:
CIs are happy too

levitte:
Merged.

60653e5b25242555446f8acf0abd5ab9ff83010c Make EVP_PKEY_CTX initialization more precise
3ee348b0dc5cd904fc2c022e6543f478c3d78732 Change EVP_PKEY_CTX_new_provided() to take a library context too.

PR #10311, opened 2019-10-31
md4/md5: macros should not include the line following them

p-steuer:


richsalz:
Is there a story behind how you found out this problem? :)

p-steuer:
A colleague of mine regularly ports OpenSSL to z/VSE. Their compiler imposes a ton of restrictions to the source code (low length limits on lines/names/files, no underscores ..). So he wrote a java tool to process the whole code base in a way their compiler can digest it. And it also happens to remove all empty lines ..  : -)

mspncp:
>Good cleanup. This has gone unnoticed for a looooong time, since the following line was simply empty
 
Note: Some people (including me) do this on purpose. It makes it easier to add lines, because you never have to worry about adding the backslash to the previous line (which is also reduces the diffs). It only works of course if you _always_ leave an empty line at the end. (And don‘t use tools that remove them ;-)

p-steuer:
merged.

master: https://github.com/openssl/openssl/commit/351ba5bd27645d5b5a2bc643b2709bd30bcdf09c
1.1.1: https://github.com/openssl/openssl/commit/ef0be09e045a934e2bb07337218fc336f7f722d7

PR #10316, opened 2019-10-31
Use a function to generate do-not-edit comment

richsalz:
The final part of my refactoring command flags.
Getting the spacing right on the "do not edit" warning was a pain, so use a function to do it.
Also, because this is the end of the series :) I added a CHANGES entry.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -150,6 +150,10 @@
      pages for further details.
      [Matt Caswell]
 
+  *) Most common options (such as -rand/-writerand, TLS version control, etc)
+     were refactored and point to newly-enhanced descriptions in openssl.pod
+     [Rich Salz]
+
   *) Over two thousand fixes were made to the documentation, including:
      adding missing command flags, better style conformance, documentation
      of internals, etc.
--- a/doc/man1/openssl-ca.pod.in
+++ b/doc/man1/openssl-ca.pod.in
@@ -1,9 +1,5 @@
 =pod
-
-=begin comment
-{- join("\n", @autowarntext) -}
-
-=end comment
+{- OpenSSL::safe::output_do_not_edit_headers(); -}
 
 =head1 NAME
 
--- a/doc/man1/openssl-cms.pod.in
+++ b/doc/man1/openssl-cms.pod.in
@@ -1,9 +1,5 @@
 =pod
-
-=begin comment
-{- join("\n", @autowarntext) -}
-
-=end comment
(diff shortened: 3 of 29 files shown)


levitte:
I disagree with the addition of the do_not_edit function. It makes it unclear what's going when just looking in the .pod.in file.
Also, considering that the only difference between what the function outputs and what you had there explicitly is an empty line, I have a hard time understanding the note about spacing being difficult... 

The CHANGES entry should go in, though could that be an independent PR?

richsalz:
Well, as the person who did all the pod.in editing, I got tripped up occasionally. :)
The point is that the pod.in is simple, and the output pod is correct.
I'll leave this up for a bit to see if anyone else has opinions.

richsalz:
I decided I feel a bit more strongly that this is the *right* thing to do. :)  Right now, the .pod.in files look like this:
```
=pod

=begin comment
{- join("\n", @autowarntext) -}

=end comment

=head1 NAME
```

With this PR, they look like this:
```
=pod
{- &OpenSSL::safe::do_not_edit(); -}

=head1 NAME
```

I think it's pretty obvious that the second is better.  But I'm biased, and I *did* do all the templating stuff for the docs. :)

richsalz:
Re-re-thinking, I don't feel so strongly. :)  I can see how the "do not edit" function name might be off-putting but don't think that is very important.

paulidale:
How about renaming the function to _output_do_not_edit_header_ or similar?
I think that bringing repeated sequences together like this is a good thing.

richsalz:
Rebased, pushed new version to change the name.

levitte:
Abusive use of `&` detected.  Bad form, please don't.

richsalz:
Abusive?  Maybe "legacy"  Whatever.  Revised commit pushed.

levitte:
> Revised commit pushed.

Maybe do it for real? :wink:

richsalz:
>Maybe do it for real?

Reality is over-rated :)  This time for real.

richsalz:
Travis failure is a timeout, unrelated.

richsalz:
This is ready for merge BTW.

levitte:
Merged.

625c781dc76df145cfd30eeff7b6de7dfcc8afc7 Use a function to generate do-not-edit comment

richsalz:
Thanks, at least now that typo-ridden branch name can go :)

PR #10317, opened 2019-10-31
s390x: fix build errors

p-steuer:
ecp_s390x_nistp.c and ecx_meth.c need to include s390x_arch.h.


p-steuer:
mMy i merge this one even though the 24h grace period has not yet expired so our nightly runs deliver meaningful results instead of reporting build breakage ?

levitte:
(it's interesting that we have started to strictly apply the 24h grace period, even though it hasn't gone through the approval process yet...  at least, we show a good example)

The 24h grace period has exceptions for minor and breakage fixes, so yes, go ahead.

p-steuer:
merged.

master: https://github.com/openssl/openssl/commit/c47a56d615a7baf974de85e7f609cacbf5297dd5

PR #10318, opened 2019-11-01
test_keymgmt_internal: fix endianness problems

p-steuer:

### DIFF
--- a/test/keymgmt_internal_test.c
+++ b/test/keymgmt_internal_test.c
@@ -163,8 +163,17 @@ static int test_pass_rsa(FIXTURE *fixture)
      * from the key.
      */
 
-    for (i = 0; i < OSSL_NELEM(expected); i++)
-        ret += !! TEST_int_eq(expected[i], keydata[i]);
+    for (i = 0; i < OSSL_NELEM(expected); i++) {
+        const union {
+            int buff;
+            char little;
+        } endian = {1};
+        size_t len = params[i].return_size;
+        size_t off = endian.little ? 0 : sizeof(expected[0]) - len;
+
+        ret += !! TEST_mem_eq((unsigned char *)&expected[i] + off, len,
+                              &keydata[i], len);
+    }
 
     ret = (ret == OSSL_NELEM(expected));
 
--- a/test/keymgmt_internal_test.c
+++ b/test/keymgmt_internal_test.c
@@ -17,6 +17,7 @@
 #include <openssl/core_names.h>
 #include "internal/nelem.h"
 #include "crypto/evp.h"          /* For the internal API */
+#include "ossl_test_endian.h"
 #include "testutil.h"
 
 typedef struct {
@@ -69,7 +70,8 @@ static int test_pass_rsa(FIXTURE *fixture)
 #define QINV    9
 #define C3      10               /* Extra coefficient */
 
-    size_t i;
+    DECLARE_IS_ENDIAN;
+    size_t i, len, off;
     int ret = 0;
     RSA *rsa = NULL;
     BIGNUM *bn1 = NULL, *bn2 = NULL, *bn3 = NULL;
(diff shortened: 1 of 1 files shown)


test/keymgmt_internal_test.c
```diff
@@ -163,8 +163,17 @@ static int test_pass_rsa(FIXTURE *fixture)
...
+        const union {
+            int buff;
+            char little;
+        } endian = {1};
+        size_t len = params[i].return_size;
+        size_t off = endian.little ? 0 : sizeof(expected[0]) - len;
+
+        ret += !! TEST_mem_eq((unsigned char *)&expected[i] + off, len,
+                              &keydata[i], len);
+    }
```
levitte: Wait, hold on..... We're comparing native unsigned long types, so this *should* be endianness agnostic... Why is this necessary?

This does worry me, because it means that something has changed numerically along the way, and that's simply not ok.

p-steuer: > it means that something has changed numerically along the way

No, also in big endian case, the values are placed to lowest address ("left aligned") in OSSL_PARAM_ulong/keydata buffers. So comparing these ulongs as uints will result in the 4 highest address bytes (rightmost) being compared and comparing them as ulongs will obv. also yield a wrong result. So only option is to resort to memcmp here.

Of course this raises the questions if numerical values should always be right aligned in big-endian case (That is, in terms of OSSL_PARAM struct, if "data_size" > "return_size", the returned bytes had to be right aligned in the buffer pointed to by "data"):
```
diff --git a/crypto/params.c b/crypto/params.c
index b2ceb13..fd1f84d 100644
--- a/crypto/params.c
+++ b/crypto/params.c
@@ -641,7 +641,7 @@ int OSSL_PARAM_set_BN(OSSL_PARAM *p, const BIGNUM *val)
     if (p->data == NULL)
         return 1;
     return p->data_size >= bytes
-        && BN_bn2nativepad(val, p->data, bytes) >= 0;
+        && BN_bn2nativepad(val, p->data, p->data_size) >= 0;
 }
```
The above change makes the keymanagement test pass on BE, but a bunch of other tests will fail. So i assumed it works as expected/designed (always left align the ossl param data) and the test case has to be changed ..

p-steuer: tl;dr: The question, if code or testcase has to be changed, comes down a design decision: should big-endian numerical values be left of right aligned in OSSL_PARAM?

p-steuer: BTW is s390 the only BE platform that we regularly test on ?

levitte: Ah, it actually looks like `OSSL_PARAM_set_BN()` has a bug.  `p->data` should point at whatever the caller deems a proper buffer, and should be aligned for native numbers fit into that buffer.  The diff you display there should become a PR in place of this one (i.e. fix the actual problem instead of working around it).

levitte: For now, yeah, it seems the s390 is the BE platform that gets regularly tested...  simply because a few of us have that sort of access.  We did test on PPC for a while, but I'm frankly unsure if we have regular access to such a system right now.

p-steuer: Okay i had a closer look at it: right alligning BE numerical values would require to completely track to offset into the buffer, besides its size and returned size. That seems not like its supposed to work. If you look at it from an applications view, you get a buffer of return_size length in host byte order such that there is no right or left alignment. However, on BE application could not simply cast the buffer to some data type which is not exactly returned_with. The same is formerly true for LE for data types wider than returned_with since the api does not claim to 0-pad the whole data_size buffer.
Thats why i still think code is alright and testcase should be fixed.

Pushed a small update.

levitte: So, a small misunderstanding...  The caller *provides* the buffer and its length, and whoever fills them in should do so according to what they get from the caller.  `return_size` is only useful to the caller if they haven't given a buffer at all (because they want to find out the minimum size needed) or provided a too small buffer.

So like I said, it seems like you have found a case where the OSSL_PARAM functions aren't doing things quite right.

p-steuer: I also experimented with your solution (among others) today but i always had to change at least one test case. Imho what the test cases require to code to do is only achievable with keeping track of offset.

p-steuer:
closed in favor of https://github.com/openssl/openssl/pull/10326.

PR #10320, opened 2019-11-01
crypto/sm2/sm2_pmeth.c: add sm2 default userid

moehuster:
##### Checklist

### DIFF
--- a/crypto/sm2/sm2_pmeth.c
+++ b/crypto/sm2/sm2_pmeth.c
@@ -32,12 +32,17 @@ typedef struct {
 static int pkey_sm2_init(EVP_PKEY_CTX *ctx)
 {
     SM2_PKEY_CTX *smctx;
+    const char *id = SM2_DEFAULT_USERID;
+    size_t id_len = strlen(id);
 
     if ((smctx = OPENSSL_zalloc(sizeof(*smctx))) == NULL) {
         SM2err(SM2_F_PKEY_SM2_INIT, ERR_R_MALLOC_FAILURE);
         return 0;
     }
 
+    smctx->id = (uint8_t*)OPENSSL_strndup(id, id_len);
+    smctx->id_len = id_len;
+    smctx->id_set = 1;
     ctx->data = smctx;
     return 1;
 }
@@ -62,6 +67,9 @@ static int pkey_sm2_copy(EVP_PKEY_CTX *dst, EVP_PKEY_CTX *src)
         return 0;
     sctx = src->data;
     dctx = dst->data;
+    /* SM2_PKEY_CTX has a default id */
+    OPENSSL_free(dctx->id);
+    dctx->id = NULL;
     if (sctx->gen_group != NULL) {
         dctx->gen_group = EC_GROUP_dup(sctx->gen_group);
         if (dctx->gen_group == NULL) {


crypto/sm2/sm2_pmeth.c
```diff
@@ -32,12 +32,24 @@ typedef struct {
...
+    uint8_t *id = SM2_DEFAULT_USERID;
+    size_t id_len = strlen(id);
+    uint8_t *tmp_id;
 
     if ((smctx = OPENSSL_zalloc(sizeof(*smctx))) == NULL) {
         SM2err(SM2_F_PKEY_SM2_INIT, ERR_R_MALLOC_FAILURE);
         return 0;
     }
 
+    tmp_id = OPENSSL_malloc(id_len);
```
richsalz: replace lines 44-49 with OPENSSL_strdup ?

moehuster: OPENSSL_strdup need more than once strlen invoketion

richsalz: Optimizing a strlen compared with code clarity is the wrong thing to do.

moehuster: got it!

levitte:
This should definitely be cherry-picked up to master as well, so I've labelled this for both master and 1.1.1.  *To anyone who merges:* it should be absolutely fine to start with applying this on 1.1.1 and cherry-pick it up to master, but the reverse should be harmless as well.

levitte:
It looks like this triggered a bug somewhere, see https://travis-ci.org/openssl/openssl/jobs/605862230

More precisely, there's a memory leak.  And just FYI, it's *not* this PR's fault as far as I can see, so I think we've a case of sloppy cleaning somewhere else.

```
    # [05:46:59]  3748 file=crypto/sm2/sm2_pmeth.c, line=43, thread=47963119216704, number=17, address=0x13fe600
    # [05:46:59]  4537 file=crypto/sm2/sm2_pmeth.c, line=43, thread=47963119216704, number=17, address=0x13fe640
    # [05:46:59]  3555 file=crypto/sm2/sm2_pmeth.c, line=43, thread=47963119216704, number=17, address=0x1401100
    # 51 bytes leaked in 3 chunks
/home/travis/build/openssl/openssl/util/shlib_wrap.sh /home/travis/build/openssl/openssl/test/evp_extra_test => 1
not ok 1 - running evp_extra_test
```

mattcaswell:
@InfoHunter - please take a look at this.

moehuster:
> @InfoHunter - please take a look at this.

sorry for commit nothing, just ignore it

InfoHunter:
I am not sure yet if we need to set the default ID to SM2 - discussions welcome ;-)

moehuster:
We use the default ID in most cases. User defined ids only appear in TEST scripts.

InfoHunter:
> We use the default ID in most cases. User defined ids only appear in TEST scripts.

I was not saying the default ID should not be used. I was just explaining why OpenSSL doesn't set the ID by default. And there is indeed a reason about this:

For what OpenSSL has discussed before, we decided at that time it's the application's responsibility to set the default SM2 ID instead of OpenSSL's. So I don't know it's worth to change this consideration or not...

moehuster:
> It looks like this triggered a bug somewhere, see https://travis-ci.org/openssl/openssl/jobs/605862230
> 
> More precisely, there's a memory leak. And just FYI, it's _not_ this PR's fault as far as I can see, so I think we've a case of sloppy cleaning somewhere else.
> 
> ```
>     # [05:46:59]  3748 file=crypto/sm2/sm2_pmeth.c, line=43, thread=47963119216704, number=17, address=0x13fe600
>     # [05:46:59]  4537 file=crypto/sm2/sm2_pmeth.c, line=43, thread=47963119216704, number=17, address=0x13fe640
>     # [05:46:59]  3555 file=crypto/sm2/sm2_pmeth.c, line=43, thread=47963119216704, number=17, address=0x1401100
>     # 51 bytes leaked in 3 chunks
> /home/travis/build/openssl/openssl/util/shlib_wrap.sh /home/travis/build/openssl/openssl/test/evp_extra_test => 1
> not ok 1 - running evp_extra_test
> ```

t8m:
This would need to be submitted against the master branch to be considered. This is not a bug fix to be done on a released branch.

Issue #9262, opened 2019-06-27
s390x aes regression


p-steuer:
Some commit from(*) df05f2ce6d496232f3c86acb299a128d0eb3ef42 to 64adf9aac765f0872c33d225c57e5c128f5d7c69 introduces regressions with aes on s390. Hard to say which exactly, because s390 build is broken in between.

edit: (*) i mean df05f2ce6d496232f3c86acb299a128d0eb3ef42 is the last commit thats working.

Havent been able to figure it out for myself just yet.

OPENSSL_s390xcap=z900 TESTS=test_bio_enc make test -> FAIL
OPENSSL_s390xcap=z990 TESTS=test_bio_enc make test -> FAIL
OPENSSL_s390xcap=z9 TESTS=test_bio_enc make test -> PASS
[...]

Hints: seems aes-cbc s390 assembly software code path broke: test output points to aes-cbc. with z9 KM instruction is used for aes-128 blocks instead of sw (which could explain the test case passes from here).

OPENSSL_s390xcap=z900 TESTS=test_evp_extra make test -> FAIL
OPENSSL_s390xcap=z990 TESTS=test_evp_extra make test -> FAIL
OPENSSL_s390xcap=z9 TESTS=test_evp_extra make test -> FAIL
OPENSSL_s390xcap=z10 TESTS=test_evp_extra make test -> PASS
[...]

Hints: with z10 KM instruction is also used for aes-192 and -256 instead of sw (which could again explain the test case passes for newer models).

For reference, some sample outputs with VERBOSE=1 ..

bio_enc :
```
make depend && make _tests
make[1]: Entering directory '/root/msa9'
make[1]: Leaving directory '/root/msa9'
make[1]: Entering directory '/root/msa9'
( cd test; \
  mkdir -p test-runs; \
  SRCTOP=../. \
  BLDTOP=../. \
  RESULT_D=test-runs \
  PERL="/usr/bin/perl" \
  EXE_EXT= \
  OPENSSL_ENGINES=`cd .././engines 2>/dev/null && pwd` \
  OPENSSL_DEBUG_MEMORY=on \
    /usr/bin/perl .././test/run_tests.pl test_bio_enc )
../test/recipes/90-test_bio_enc.t ..
1..1
    # Subtest: ../../test/bio_enc_test
    1..6
        # Subtest: test_bio_enc_aes_128_cbc
        1..2
        # ERROR: (memory) 'inp == out' failed @ test/bio_enc_test.c:142
[...]

Issue #9431, opened 2019-07-22
SHA3_squeeze causes segmentation fault on some platforms


tniessen:
While trying to add support for variable output lengths for shake128 and shake256 to Node.js, I noticed a number of segmentation faults on some of our CI platforms such as s390x. We are using OpenSSL 1.1.c.

One test case calls `EVP_DigestFinalXOF` with a requested output size of zero bytes. I expected OpenSSL not to write to the supplied buffer in this case, and that is true on some platforms. On other platforms such as s390x, `SHA3_squeeze` causes a segmentation fault, maybe trying to access the buffer which is `NULL`.

It is possible that other inputs are affected as well, even though I was not able to produce any segmentation faults for inputs other than `0`.

Arguably, calling `EVP_DigestFinalXOF` with an output size of `0` bytes makes little sense, but it still should not cause a segmentation fault in my opinion. The C implementation of `SHA3_squeeze` seems to handle this case nicely.

mattcaswell:
Ping @p-steuer.

Is it only s390x, or are there are platforms impacted?

tniessen:
The following platforms appear to be impacted:

- s390x
- Raspberry Pi 2, Raspberry Pi 3
- ppc64
- armv7
- ppcle

There might be other affected platforms that we are not testing against.

p-steuer:
Im working on a fix (that is, match sha3_squeeze asm implementations to the reference implementation), but may need help to verify the fix for architectures i have no access.

p-steuer:
should be firxed by a890ef8 (master) resp. 6087d4a (1.1.1).

please verify.

tniessen:
I can confirm that this fixes the problem for output lengths of zero bytes, I have never been able to produce segfaults using other sizes.

PR #10323, opened 2019-11-01
Dont pass zero length input to asm modules for ciphers

p-steuer:
The asm modules may assume an input length > 0.

Fixes: #9262

Remember the same applied to keccak https://github.com/openssl/openssl/issues/9431, so in general we shouldnt pass zero length input to an asm module that does not explicitely claim to handle that case. This was handeled correctly before the re-plumbing, but seems to have been lost here and there.

### DIFF
--- a/providers/common/ciphers/cipher_common.c
+++ b/providers/common/ciphers/cipher_common.c
@@ -180,6 +180,8 @@ int cipher_generic_block_update(void *vctx, unsigned char *out, size_t *outl,
             ERR_raise(ERR_LIB_PROV, PROV_R_OUTPUT_BUFFER_TOO_SMALL);
             return 0;
         }
+    }
+    if (nextblocks > 0) {
         if (!ctx->hw->cipher(ctx, out, in, nextblocks)) {
             ERR_raise(ERR_LIB_PROV, PROV_R_CIPHER_OPERATION_FAILED);
             return 0;


p-steuer:
@openssl/omc anyone ?

t8m:
Should there be a test for this case?

p-steuer:
How would a test check at runtime which code path is taken in non-debug build i.e., w/o using debugger breakpoint ?

t8m:
I mean is there any asm cipher implementation which if it is passed zero length input misbehaves and produces some incorrect output? Then it could be tested that this check in cipher_generic_block_update avoids it.

p-steuer:
> I mean is there any asm cipher implementation which if it is passed zero length input misbehaves and produces some incorrect output?

Yes, see https://github.com/openssl/openssl/issues/9262 .

> Then it could be tested that this check in cipher_generic_block_update avoids it.

Yes, this check avoids it. Therefore commit message has "Fixes: #9262".

t8m:
OK, but unfortunately #9262 is not something which could be tested in our CI. So never mind my request.

p-steuer:
I mean the openssl test suite finds it, there is no new test needed or something. Our (openssl) travis/appveyor CI doesnt find it, only because s390 is no travis target. Our (ibm) internal CI finds it.

p-steuer:
With PR https://github.com/openssl/openssl/pull/10446 our CI could find it, if i ran it with OPENSSL_s390xcap=z900 env variable exported, but running the test suite with all possible settings would result in inacceptable long travis runs.

p-steuer:
merged.

master: https://github.com/openssl/openssl/commit/51356a066e936354c817097cf873f4974bbea291
fixed: https://github.com/openssl/openssl/issues/9262

PR #10324, opened 2019-11-01
 Ensure EVP_get_digestbyname() and EVP_get_cipherbyname() know all aliases

mattcaswell:
Now that we have an EVP namemap containing all aliases that providers know about for any given algorithm, it is possible that an application attempts to look up a digest or a cipher via `EVP_get_digestbyname()` or `EVP_get_cipherbyname()` with an algorithm name that is unknown to the legacy method database. Therefore we extend those functions to additionally check the aliases in the namemap when searching for a method in the event that our initial lookup attempt fails.

### DIFF
--- a/crypto/evp/pmeth_lib.c
+++ b/crypto/evp/pmeth_lib.c
@@ -593,7 +593,11 @@ int EVP_PKEY_CTX_get_signature_md(EVP_PKEY_CTX *ctx, const EVP_MD **md)
     if (!EVP_PKEY_CTX_get_params(ctx, sig_md_params))
         return 0;
 
-    tmp = EVP_get_digestbyname(name);
+    /*
+     * TODO(3.0): We need to pass the library context here - but at the time
+     * of writing we don't seem to store that in the EVP_PKEY_CTX.
+     */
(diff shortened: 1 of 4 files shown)


crypto/evp/pmeth_lib.c
```diff
@@ -563,7 +563,11 @@ int EVP_PKEY_CTX_get_signature_md(EVP_PKEY_CTX *ctx, const EVP_MD **md)
     if (!EVP_PKEY_CTX_get_params(ctx, sig_md_params))
         return 0;
 
-    tmp = EVP_get_digestbyname(name);
+    /*
+     * TODO(3.0): We need to pass the library context here - but at the time
+     * of writing we don't seem to store that in the EVP_PKEY_CTX.
+     */
+    tmp = evp_get_digestbyname_ex(NULL, name);
```
levitte: Considering #10308, you could put in `ctx->libctx` and mark this `[Pending on #10308]`

mattcaswell: Rebased to pull in #10308, and fixed this.

mattcaswell:
Pushed. Thanks.

PR #10326, opened 2019-11-01
Fix OSSL_PARAM_set_BN() to fill the given buffer correctly.

levitte:
OSSL_PARAM_set_BN() filled the buffer from the left with as many bytes
as that the BIGNUM takes, regardless of buffer size or native
endianness.  This was due to BN_bn2nativepad() being given the size of
the BIGNUM rather than the size of the buffer (which meant it never
had to pad anything).

The fix is to given BN_bn2nativepad() the size of the buffer instead.
This aligns well with the corresponding _set_ functions for native
integer types work.

Alternative to #10318

### DIFF
--- a/crypto/params.c
+++ b/crypto/params.c
@@ -640,8 +640,11 @@ int OSSL_PARAM_set_BN(OSSL_PARAM *p, const BIGNUM *val)
     p->return_size = bytes;
     if (p->data == NULL)
         return 1;
-    return p->data_size >= bytes
-        && BN_bn2nativepad(val, p->data, bytes) >= 0;
+    if (p->data_size >= bytes) {
+        p->return_size = p->data_size;
+        return BN_bn2nativepad(val, p->data, p->data_size) >= 0;
+    }
+    return 0;
 }
 
 OSSL_PARAM OSSL_PARAM_construct_BN(const char *key, unsigned char *buf,
--- a/doc/man3/OSSL_PARAM.pod
+++ b/doc/man3/OSSL_PARAM.pod
@@ -100,10 +100,12 @@ accepted, otherwise it specifies the maximum size allowed.
 =item I<return_size>
 
 When an array of B<OSSL_PARAM> is used to request data, the
-I<responder> must set this field to indicate the actual size of the
-parameter data.
-In case the I<data_size> is too small for the data, the I<responder>
-must still set this field to indicate the minimum data size required.
+I<responder> must set this field to indicate size of the parameter
+data, including padding as the case may be.
+In case the I<data_size> is an unsuitable size for the data, the
+I<responder> must still set this field to indicate the minimum data
+size required.
+(further notes on this in L</NOTES> below).
 
 When the B<OSSL_PARAM> is used as a parameter descriptor,
 I<return_size> should be ignored.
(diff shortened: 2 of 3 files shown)


levitte:
I have just verified the `test_internal_keymgmt` test with this on s390x.  The failure that's reported in #10318 is now gone.

p-steuer:
Like this, the whole buffer will allways be zero padded, no matter how big and return_size will always return the buffer size instead of the BN's size. If thats the desired/documented behavior im okay with it ( I havent read the corresponding man pages yet).

However, in that case params_appi test case has to be changed, since it expects returned_size == BN's size:
```  
    # ERROR: (size_t) 'cp->return_size == sizeof(bn_val)' failed @ test/params_api_test.c:551
    # [100] compared to [16] 
```

p-steuer (changes requested):
see https://github.com/openssl/openssl/pull/10326#issuecomment-548939268

levitte:
> Like this, the whole buffer will always be zero padded, no matter how big

Yes.  Basically, the caller rules, the only thing someone filling in the data can do is adapt, or fail (if you look at the other integer setter functions, you will see that they fail if the buffer has a size they cannot handle, and you can specifically see that for `set_int32` and `set_uint32`, they can fill a 64-bit buffer.

> and return_size will always return the buffer size instead of the BN's size.

Again, look at the `set_int` and `set_uint` functions, they do the same thing, i.e. "this is how many bytes we filled in".

Now, BIGNUMs don't have a fixed size, so they can be used liberally to fill any size buffer.  It's up to the caller to set the limit.

Do note that then the `data` field is NULL, *then* `return_size` is set to the minimum required space to fill in the data available.  That is on purpose to allow the caller to know how much space it needs to allocate.

> If thats the desired/documented behavior im okay with it (havent read the corresponding man pages yet).

levitte:
> However, in that case params_api test case has to be changed instead, since it expects returned_size == BN's size:

Now, *that* test is incorrect, but this is also a test that was written together with function this PR corrects, so the test is bug compatible :laughing:.  I will fix that as well.

p-steuer:
Thanks for the explaination, makes sense. My knowledge of the OSSL_PARAM interface was too limited to judge which part of the puzzle didnt fit here.

levitte:
Gotcha...

levitte:
Test correction added.  Curiously, no-one noticed that the size check for BN didn't match the size check when setting all other integers.  That should have been a red flag...

levitte:
And now, *all* tests pass on s390x

p-steuer (approved):
> And now, all tests pass on s390x

At least if you dont export OPENSSL_s390xcap=z900, z990 or z9 ;-)   (https://github.com/openssl/openssl/pull/10323).

levitte:
I added some clarifications in OSSL_PARAM(3), please re-review

p-steuer (approved):
Doc also looks correct.

p-steuer:
merged, thanks.

master: https://github.com/openssl/openssl/commit/0f73e719c6ca6c2e955e6c08a3ab171642dc2dc0 https://github.com/openssl/openssl/commit/2c99372dbdab23726327777f36e2cbc0a1af0873 https://github.com/openssl/openssl/commit/909ef4de3187b752710b7ae69b6df5df51251600

PR #10329, opened 2019-11-01
Fix EC_POINT_bn2point() for BN_zero()

### LINKED ISSUE #10258: `EC_POINT_hex2point` incorrectly decodes the point at infinity
`EC_POINT_hex2point` incorrectly decodes the point at infinity, which is encoded as the hexstring `00`. This is because `hex2point` calls `oct2point` after converting `00` into a `BIGNUM` and calling `BN_bn2bin` on this bignum, which outputs 0 bytes whereas the decoder is expecting exactly `1` `0` byte in this case.

romen:
# Fix `EC_POINT_bn2point()` for `BN_zero()`

`EC_POINT_bn2point()` rejected `BIGNUM`s with a zero value.

This behavior indirectly caused failures when converting a point
at infinity through `EC_POINT_point2hex()` and then back to a point with
`EC_POINT_hex2point()`.

With this change such `BIGNUM`s are treated like any other and exported to
an octet buffer filled with zero.
It is then `EC_POINT_oct2point()` (either the default implementation or
the custom one in `group->meth->oct2point`) to determine if such encoding
maps to a valid point (generally the point at infinity is encoded as
`0x00`).

Fixes #10258

# Add self-test for `EC_POINT_hex2point()`

Adds tests for each curve to ensure that encodings obtained through
`EC_POINT_hex2point()` can be fed to `EC_POINT_point2hex()` yielding a point
identical to the one from which the encoding is generated.

##### Checklist
- [x] documentation added

### DIFF
--- a/crypto/ec/ec_print.c
+++ b/crypto/ec/ec_print.c
@@ -39,13 +39,13 @@ EC_POINT *EC_POINT_bn2point(const EC_GROUP *group,
     EC_POINT *ret;
 
     if ((buf_len = BN_num_bytes(bn)) == 0)
-        return NULL;
+        buf_len = 1;
     if ((buf = OPENSSL_malloc(buf_len)) == NULL) {
         ECerr(EC_F_EC_POINT_BN2POINT, ERR_R_MALLOC_FAILURE);
         return NULL;
     }
 
-    if (!BN_bn2bin(bn, buf)) {
+    if (!BN_bn2binpad(bn, buf, buf_len)) {
         OPENSSL_free(buf);
         return NULL;
     }
--- a/doc/man3/EC_POINT_new.pod
+++ b/doc/man3/EC_POINT_new.pod
@@ -171,6 +171,26 @@ The functions EC_POINT_point2oct(), EC_POINT_oct2point(), EC_POINT_point2bn(),
 EC_POINT_bn2point(), EC_POINT_point2hex() and EC_POINT_hex2point() convert from
 and to EC_POINTs for the formats: octet, BIGNUM and hexadecimal respectively.
 
+The function EC_POINT_point2oct() encodes the given curve point B<p> as an
+octet string into the buffer B<buf> of size B<len>, using the specified
+conversion form B<form>.
+The encoding conforms with Sec. 2.3.3 of the SECG SEC 1 ("Elliptic Curve
+Cryptography") standard.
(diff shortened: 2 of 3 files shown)


slontis:
Does this need a CHANGES entry? (I dont know the answer)

romen:
> Does this need a CHANGES entry? (I dont know the answer)

I am not sure either, but gravitating toward a "probably not" answer.

Maybe the OMC reviewer can give feedback on it. (ping @mattcaswell )

I am currently looking into #10366 which is somehow related, but I am still undecided if that should have an impact on this PR or not.

mattcaswell:
> Does this need a CHANGES entry? (I dont know the answer)

I'd say probably not. However perhaps the man page should be updated with something about what happens with points at infinity?

Other than that this PR looks good.

romen:
> > Does this need a CHANGES entry? (I dont know the answer)
> 
> I'd say probably not. However perhaps the man page should be updated with something about what happens with points at infinity?
> 
> Other than that this PR looks good.

Ultimately it's each `EC_METHOD` responsibility to decide what is an acceptable representation of any point, including point at infinity, through the `point2oct` and `oct2point` pointers.
I believe all of the currently included implementations translate `Pinf -> \0\0 -> Pinf`, but can we really document something specific?

mattcaswell:
> Ultimately it's each EC_METHOD responsibility to decide what is an acceptable representation of any point, including point at infinity, through the point2oct and oct2point pointers.

Surely, if an application calls `EC_POINT_bn2point()` then it is going to expect to always get the same result for a given point, regardless of which EC_METHOD happened to have been chosen? Similarly it should not matter which GFp curve you use they should all behave consistently with each other. And all GF2m curves should also be consistent with each other.

romen:
@mattcaswell do you think the added lines clarify things enough?

romen:
Merged to `master` with:
- d47c108756 Fix EC_POINT_bn2point() for BN_zero()
- 35ed029b5a Add self-test for EC_POINT_hex2point
- 3cc26f2eba Extend docs for EC_POINT conversion functions

Merged to `1.1.1` with:
- 6f6adf1d7b Fix EC_POINT_bn2point() for BN_zero()
- f59967cb72 Add self-test for EC_POINT_hex2point _(trivial changes on cherry-picking)_
- 3c25ac2d2e Extend docs for EC_POINT conversion functions

Thanks!

PR #10330, opened 2019-11-01
s390x assembly pack: enable clang build

p-steuer:
clang imposes some restrictions on the assembler code that
gcc does not.

### DIFF
--- a/crypto/bn/asm/s390x-mont.pl
+++ b/crypto/bn/asm/s390x-mont.pl
@@ -147,7 +147,7 @@ $code.=<<___;
 	lghi	$NHI,0
 	alcgr	$NHI,$nhi
 
-	la	$j,8(%r0)	# j=1
+	la	$j,8		# j=1
 	lr	$count,$num
 
 .align	16
@@ -199,7 +199,7 @@ $code.=<<___;
 	lghi	$NHI,0
 	alcgr	$NHI,$nhi
 
-	la	$j,8(%r0)	# j=1
+	la	$j,8		# j=1
 	lr	$count,$num
 
 .align	16
@@ -243,7 +243,7 @@ $code.=<<___;
 	la	$ap,$stdframe($sp)
 	ahi	$num,1		# restore $num, incidentally clears "borrow"
 
-	la	$j,0(%r0)
+	la	$j,0
 	lr	$count,$num
 .Lsub:	lg	$alo,0($j,$ap)
 	lg	$nlo,0($j,$np)
@@ -257,7 +257,7 @@ $code.=<<___;
 	lghi	$NHI,-1
 	xgr	$NHI,$AHI
 
-	la	$j,0(%r0)
+	la	$j,0
 	lgr	$count,$num
 .Lcopy:	lg	$ahi,$stdframe($j,$sp)	# conditional copy
 	lg	$alo,0($j,$rp)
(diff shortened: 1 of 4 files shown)


levitte:
Not sure I understand how the changes become the same thing.  Is `%r0` implied anyway?

paulidale:
Would it be better to have a conditionally defined macro to handle the two different cases?
I don't know much about 390 ASM but it seems odd removing and index register.

p-steuer:
```
A  zero  in  any  of  the  B1,  B2,  X2,  B3,  or  B4 fields  indi-
cates  the  absence  of  the  corresponding  address
component.  For  the  absent  component,  a  zero  is
used  in  forming  the  intermediate  sum,  regardless  of
the contents of general register 0. A displacement of
zero has no special significance. 
```
[ Operand-Address Generation p.5-11 IBM z/Architecture POP SA22-7832-12  https://publibfp.dhe.ibm.com/epubs/pdf/a227832c.pdf ]

Register general purpose register (R0-R15) operands are encoded in the instructions as 4 bit fields (0000-1111) ie, R0 is encoded as 0000. All general purpose registers are data registers, but only R1-R15 are address registers. That is, specifying R0 in an address generation as base resp. index means putting zeros in that instruction field to ignore base resp. index register component in the calculation. While gcc simply does that (replace R0 by 0000) when expanding the memnonic, clang will fail with "invalid address register". clang enforces, that if you dont want to use base or index register in address generation, to not specify those components.

levitte:
So...  does that mean that you can't load an address as a constant from a base register, i.e. something like this?

```
         la      %r3,45(0,%r2)
```

That seems odd...  Or could it be that if any of the base or index register is given, clang wants both, i.e. while clang won't tolerate this:

```
        la      %r3,8(%r0)
```

...  it does tolerate this?

```
        la      %r3,8(%r0,%r0)
```

With your explanation, I'm fine with your change, or could go the fully explicit way.  I *am* curious to know if clang tolerates the fully explicit way, would you be up for satisfying my curiousity?

p-steuer:
No, clang simply doesnt want R0 to be specified as an address register (and rightly so):
```
crypto/bn/s390x-mont.S:100:13: error: %r0 used in an address
 la %r7,8(0,%r0)
            ^
crypto/bn/s390x-mont.S:100:11: error: %r0 used in an address
 la %r7,8(%r0)
          ^
crypto/bn/s390x-mont.S:100:11: error: %r0 used in an address
 la %r7,8(%r0,%r0)
          ^
```
Plus there seems to be no way to leave some kind of place-holder for the instruction's to be ignored fields:
```
crypto/bn/s390x-mont.S:100:13: error: register expected
 la %r7,8(0,0)
            ^
crypto/bn/s390x-mont.S:100:11: error: unknown token in expression
 la %r7,8(,)
          ^
crypto/bn/s390x-mont.S:100:11: error: unknown token in expression
 la %r7,8()
          ^
```

p-steuer:
Think it makes sense to backport this to 1.1.1.

levitte:
Ok then.  Thanks for all the clarifications!

p-steuer:
merged.

master: https://github.com/openssl/openssl/commit/6f93f06135cbbd36c3fe98d63717e8303a5d559b
1.1.1: https://github.com/openssl/openssl/commit/72f4d2f8eb0a95bbac7e247ff571bbef4158018c

PR #10331, opened 2019-11-01
Allow specifying the tag after AAD in CCM mode (2)

p-steuer:
In addition to 67c81ec3 which introduced this behavior in CCM mode
docs but only implemented it for AES-CCM.


slontis:
Hi Patrick - there are a few issues with this..
1) This looks like a feature - so maybe should not be going back ported to 1.1.1?
2) e_aria.c is going to be deprecated in master and replaced by the provider code so this seems like the wrong file to do this change in. (As the ccm code is shared in the provider it most likely already works with your test change ).

p-steuer:
Hi Shane,
i classified it as a 1.1.1 bug because said commit changed the doc, saying setting tag early is possible for CCM mode.
You are right, provider code already handles it correctly and the additional test passes.

(1) Regarding 1.1.1, i would only put it there if its seen as a bug.

(2) Regarding master, i would only put the test case change there, because e_aria.c is deprecated and the change would not even be tested by default because provider code is used instead.

Opinions ?

paulidale:
For master and 1.1.1.

p-steuer:
merged.

master: https://github.com/openssl/openssl/commit/f7382fbbd846dd3bdea6b8c03b6af22faf0ab94f
1.1.1: https://github.com/openssl/openssl/commit/a13dddea6b71743116e86e39d446b9fb43e1799e

PR #10333, opened 2019-11-02
BIO_s_connect: add an error state and use it [1.1.1]

levitte:
If no connection could be made, addr_iter will eventually end up being
NULL, and if the user didn't check the returned error value, the
BIO_CONN_S_CONNECT code will be performed again and will crash.

So instead, we add a state BIO_CONN_S_CONNECT_ERROR that we enter into
when we run out of addresses to try.  That state will just simply say
"error" back, until the user does something better with the BIO, such
as free it or reset it.

-----

Backport of #7630


mattcaswell (approved):

levitte:
Merged.

f6483fc2db36fbfd622b6aff19d8ebfaa274c996 BIO_s_connect: add an error state and use it

PR #10334, opened 2019-11-02
test/build.info: add missing inclusion for ssl_ctx_test

levitte:

PR #10335, opened 2019-11-02
Fix --strict-warnings build

p-steuer:
The %zd format corresponds to ssize_t which is used for
function to either return a valid size or a negative value
to indicate an error. Since size_t is in [-1,SSIZE_MAX] it
is not a portable way to represent a pointer diff. For
the %td format which corresponds to ptrdiff_t is C11,
we chose to cast to long instead as it is already done
in other places.

### DIFF
--- a/crypto/x509/pcy_tree.c
+++ b/crypto/x509/pcy_tree.c
@@ -49,8 +49,8 @@ static void tree_print(BIO *channel,
         curr++;
 
     BIO_printf(channel, "Level print after %s\n", str);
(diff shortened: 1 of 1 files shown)


p-steuer:
Travis/AppVeyor red crosses look unrelated to this change.

levitte:
Yes, correct. Please have a look at #10334 for the fix

p-steuer:
merged.

master: https://github.com/openssl/openssl/commit/e774adb593e7bf9057775d85ecc7c24e9bacee12

Issue #9538, opened 2019-08-06
test_ssl_new fails with asan


p-steuer:
char * (alignment 1) ist casted to union sctp_notification * (alignment 4) :
```
crypto/bio/bss_dgram.c:1077:35: runtime error: member access within misaligned address 0x62900016d203 for type 'union sctp_notification', which requires 4 byte alignment
```
char * pointer is parameter of BIO_METHOD's bread function pointer and sctp_notification is from sctp.h. Ideas how to fix this ? 

```
./config --strict-warnings enable-asan enable-ubsan enable-zlib-dynamic enable-unit-test enable-md2 enable-rc5 enable-buildtest-c++ enable-weak-ssl-ciphers enable-ssl3 enable-ssl3-method enable-sctp

Command line (with current working directory = .):

    /usr/bin/perl ./Configure linux64-s390x --strict-warnings enable-asan enable-ubsan enable-zlib-dynamic enable-unit-test enable-md2 enable-rc5 enable-buildtest-c++ enable-weak-ssl-ciphers 
enable-ssl3 enable-ssl3-method enable-sctp

perl configdata.pm --dump
Perl information:

    /usr/bin/perl
    5.28.1 for s390x-linux-thread-multi

Enabled features:

    afalgeng
    aria
    asan
    asm
    async
    autoalginit
    autoerrinit
    autoload-config
    bf
    blake2
    buildtest-c++
    camellia
    capieng
    cast
    chacha
    cmac
    cmp
    cms
    comp
    ct
    deprecated
    des
    dgram
    dh
    dsa
    dtls
    dynamic-engine
    ec
    ec2m
    ecdh
    ecdsa
    engine
    err
    filenames
    fips
    gost
   idea
    legacy
    makedepend
    md2
    md4
    mdc2
    module
    multiblock
    nextprotoneg
    pinshared
    ocb
    ocsp
    padlockeng
    pic
    poly1305
    posix-io
    psk
    rc2
    rc4
    rc5
    rdrand
    rfc3779
    rmd160
    scrypt
    sctp
    seed
    shared
    siphash
    siv
    sm2
    sm3
    sm4
    sock
    srp
    srtp
    sse2
    ssl
    static-engine
    stdio
    tests
    threads
    tls
    ts
    ubsan
    ui-console
    unit-test
    whirlpool
    weak-ssl-ciphers
    zlib
   zlib-dynamic
    ssl3
    ssl3-method
    tls1
    tls1-method
    tls1_1
[...]

PR #10336, opened 2019-11-02
bss_dgram.c: fix unaligned access

p-steuer:
char (alignment 1) casted to union sctp_notification (alignment > 1).

A suggestion for a fix for #9538. Please review.

### DIFF
--- a/crypto/bio/bss_dgram.c
+++ b/crypto/bio/bss_dgram.c
@@ -1125,7 +1125,7 @@ static int dgram_sctp_read(BIO *b, char *out, int outl)
                     data->handle_notifications(b, data->notification_context,
                                                (void *)out);
 
-                OPENSSL_cleanse(&snp, sizeof(snp));
+                memset(&snp, 0, sizeof(snp));
                 memset(out, 0, outl);
             } else {
                 ret += n;
--- a/crypto/bio/bss_dgram.c
+++ b/crypto/bio/bss_dgram.c
@@ -1009,7 +1009,6 @@ static int dgram_sctp_read(BIO *b, char *out, int outl)
     int ret = 0, n = 0, i, optval;
     socklen_t optlen;
     bio_dgram_sctp_data *data = (bio_dgram_sctp_data *) b->ptr;
-    union sctp_notification *snp;
     struct msghdr msg;
     struct iovec iov;
     struct cmsghdr *cmsg;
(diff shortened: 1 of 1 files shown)


crypto/bio/bss_dgram.c
```diff
@@ -1116,17 +1117,19 @@ static int dgram_sctp_read(BIO *b, char *out, int outl)
...
+                if (snp.sn_header.sn_type == SCTP_AUTHENTICATION_EVENT)
+                    dgram_sctp_handle_auth_free_key_event(b, &snp);
 #  endif
 
                 if (data->handle_notifications != NULL)
                     data->handle_notifications(b, data->notification_context,
                                                (void *)out);
 
+                OPENSSL_cleanse(&snp, sizeof(snp));
                 memset(out, 0, outl);
```
mattcaswell: It seems strange to "cleanse" `snp` but to memset `out`. They're either both sensitive (in which case "cleanse" is appropriate" for both), or they're not, in which case memset is sufficient for both (or in fact not necessary at all...)

p-steuer: In case they are sensitive, snp has to be cleansed because its a local stack variable. In case of out it depends on the context in which the function is called. If the buffer it points to is always application provided, compiler would not be allowed to optimize the memset.

In case they are not sensitive, you would not even have to memset snp.

So in case of snp its either cleanse or nothing. Not knowing the code here well enough here, i went with the conservative choice. Maybe someone who is more into sctp can comment.

mattcaswell: I find it difficult to believe that this structure holds sensitive data. @tuexen might have an opinion.

tuexen: Not sure what sensitive data means... It is metadata from the SCTP stack. Something like a path state has changed (so a remote IP address, a state), key management information (a key identifier), or similar information.

p-steuer: Sensitive means e.g., key material. Id consder the information you mentioned non-sensitive. Replaced the cleanse by a normal memset.

tuexen: There is no key material involved.

p-steuer:
merged.

master: https://github.com/openssl/openssl/commit/287e1a7eac1330f0e0e4fc443f9f762835246e9e

PR #6553, opened 2018-06-21
Refuse min proto version > max proto version (cont.)

levitte:
(originally by @tiran)

SSL_CTX_set_min_proto_version() and SSL_CTX_set_max_proto_version() as
well as their SSL equivalents now validate that the mininum protocol
version is lower or requal the maximum protocol version.

Add test cases for min/max protocol API

Closes: https://github.com/openssl/openssl/issues/5127
Signed-off-by: Christian Heimes <christian@python.org>

Also see https://github.com/openssl/openssl/pull/5126

Issue #5127, opened 2018-01-21
set_min_proto_version() does not validate that min <= max


tiran:
The setter combination ``SSL_CTX_set_min_proto_version()`` and ``SSL_CTX_set_max_proto_version()`` as well as  ``SSL_set_min_proto_version()`` and ``SSL_set_max_proto_version()`` accept invalid combinations of min and max protocol. For example it is possible to set a minimum version of TLS 1.2 but a maximum version of TLS 1.0. The helper function ``ssl_check_allowed_versions`` should prevent ``min_version > max_version``.

kaduk:
Discussion on #5128 indicates that this is not an issue, or rather, that valid combinations of setting min and max proto versions may validly progress through a state with the configured min > the configured max, so we cannot safely assert that that relation holds when setting either the min or max version.
So, closing (but leaving the pull request open for the added tests).

PR #10337, opened 2019-11-02
Add ssl_ctx_test to test suite.

richsalz:
Did anyone try to build https://github.com/openssl/openssl/pull/6553? :)
And then run it? :(

This PR makes the test program compile.  But it fails.  What to do?
Ping the guilty parties: @paulidale, @levitte, @tiran.

### DIFF
--- a/test/recipes/71-test_ssl_ctx.t
+++ b/test/recipes/71-test_ssl_ctx.t
@@ -1,0 +1,16 @@
+#! /usr/bin/env perl
+# Copyright 2015-2018 The OpenSSL Project Authors. All Rights Reserved.
+#
+# Licensed under the Apache License 2.0 (the "License").  You may not use
+# this file except in compliance with the License.  You can obtain a copy
+# in the file LICENSE in the source distribution or at
+# https://www.openssl.org/source/license.html
+
+use strict;
+use warnings;
+use OpenSSL::Test;
+
+setup("ssl_ctx_test");
+
+plan tests => 1;
+ok(run(test(["ssl_ctx_test"])));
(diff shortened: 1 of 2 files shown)


levitte:
Yup, I have run it, and think that's valuable to figure out what we do right or wrong. TDD (Test Driven Development) is a thing that we sometimes do too little of.

So basically, either the ssl code isn't quite right, or the test isn't. Fixing one or the other based on the results shown in this PR should spark another PR or two.

richsalz:
TDD is for branches, not master, IMO. :)

levitte:
Feature branches, yeah.  But that's what I'm saying, the presence of this failing test, *even though "just" in a PR*, may be reason for someone wanting to fix the problem.  Quite obviously, the fixing branch will have to be merged before this test is, to not disturb the nice green colour of master builds.

richsalz:
The "quite obviously" part wasn't clear from your earlier comments.

But WTH, master's been broken for weeks anyway.

kroeckx:
The error is most likely: https://github.com/openssl/openssl/pull/6553#pullrequestreview-132296477

kroeckx:
That test made sense for #5127, which tried to change the behaviour, but as pointed out, is not what we wanted.

levitte:
So that means we need to make some small adjustment to the test program. Would you be willing to, @kroeckx?

kroeckx:
On Sun, Nov 03, 2019 at 02:55:47AM -0800, Richard Levitte wrote:
> So that means we need to make some small adjustment to the test program. Would you be willing to, @kroeckx? 

It's just changing a 0 into a 1 on that line.

t8m:
Not just that. Also the expected max value needs to be changed.
The line should be:
`{TLS1_2_VERSION, TLS1_1_VERSION, 1, 1, TLS1_2_VERSION, TLS1_1_VERSION}`

t8m:
@richsalz could you please fix the ssl_ctx_test.c accordingly?

richsalz:
rebased and updated commit pushed.  locally-run test passes.

richsalz:
not sure what you're asking; feel free to edit the commit message yourself.

t8m:
Merged to master as 310553591c

PR #10338, opened 2019-11-02
AES_ige_encrypt: ensure memcpy buffers dont overlap

p-steuer:
AES_ige_encrypt must not process inputs less than 16 bytes.
Those inputs would lead to overlapping buffers with memcpy
and sanitizers may error.

### DIFF
--- a/crypto/aes/aes_ige.c
+++ b/crypto/aes/aes_ige.c
@@ -46,7 +46,7 @@ void AES_ige_encrypt(const unsigned char *in, unsigned char *out,
     size_t n;
     size_t len = length;
 
-    if (length == 0)
+    if (length / AES_BLOCK_SIZE == 0)
         return;
 
     OPENSSL_assert(in && out && key && ivec);
--- a/crypto/aes/aes_ige.c
+++ b/crypto/aes/aes_ige.c
@@ -46,7 +46,7 @@ void AES_ige_encrypt(const unsigned char *in, unsigned char *out,
     size_t n;
     size_t len = length;
 
-    if (length / AES_BLOCK_SIZE == 0)
+    if (length == 0)
         return;
 
     OPENSSL_assert(in && out && key && ivec);
@@ -54,6 +54,8 @@ void AES_ige_encrypt(const unsigned char *in, unsigned char *out,
     OPENSSL_assert((length % AES_BLOCK_SIZE) == 0);
 
     len = length / AES_BLOCK_SIZE;
+    if (!ossl_assert(len > 0))
+        return;
 
     if (AES_ENCRYPT == enc) {
         if (in != out &&


crypto/aes/aes_ige.c
```diff
@@ -46,7 +46,7 @@ void AES_ige_encrypt(const unsigned char *in, unsigned char *out,
     size_t n;
     size_t len = length;
 
-    if (length == 0)
+    if (length / AES_BLOCK_SIZE == 0)
```
bernd-edlinger: we have this `OPENSSL_assert((length % AES_BLOCK_SIZE) == 0)` a few lines later,
since this is a void function there is no way of returning a result, so
you could instead change the assert below to `if (!OPENSSL_assert(...)) return;`

levitte: In that case, `ossl_assert` is preferred

p-steuer: AES_ige_encrypt is an API so i didnt want to abort() applications which do this wrong and just do nothing instead (doing nothing is the current behavior, just that it leads to overlapping memcpy).

levitte: `ossl_assert` only aborts when building in debug mode, otherwise it returns the result of the assertion condition.

bernd-edlinger: > AES_ige_encrypt is an API so i didnt want to abort() applications which do this wrong and just do nothing instead (doing nothing is the current behavior, just that it leads to overlapping memcpy).

Are you sure?  OPENSSL_assert calls OPENSSL_die, unconditionally, so
the current behaviour is abort, or am I missing something?

p-steuer: > OPENSSL_assert calls OPENSSL_die, unconditionally
Yes that was exactly my point: the function did not do OPENSSL_assert((length / AES_BLOCK_SIZE) == 0) and i didnt want to add such an assertion because of the unconditional abort.

p-steuer: i no changed it to ossl_assert, so only DEBUG builds will die, indicating a missuse of the function, while production builds will continue to do nothing.

crypto/aes/aes_ige.c
```diff
@@ -54,6 +54,8 @@ void AES_ige_encrypt(const unsigned char *in, unsigned char *out,
     OPENSSL_assert((length % AES_BLOCK_SIZE) == 0);
```
bernd-edlinger: But this assert is going to abort if `length > 0 && length < AES_BLOCK_SIZE`

p-steuer: Your right. this assertion in combination with the length == 0 check should be enough, that len (number of AES blocks processed) > 0...

This is the build error i want to address:
```
crypto/aes/aes_ige.c: In function 'AES_ige_encrypt':
crypto/aes/aes_ige.c:135:13: error: 'memcpy' accessing 16 bytes at offsets 0 and 0 overlaps 16 bytes at offset 0 [-Werror=restrict]
  135 |             memcpy(ivec, ivp->data, AES_BLOCK_SIZE);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
crypto/aes/aes_ige.c:82:13: error: 'memcpy' accessing 16 bytes at offsets 0 and 0 overlaps 16 bytes at offset 0 [-Werror=restrict]
   82 |             memcpy(ivec, ivp->data, AES_BLOCK_SIZE);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Well, the code does some pointer aliasing but i can only see this ever happening if the while(len) loop above those lines is skipped (i.e. len==0).
Also adding the proposed changes makes the error disappear, indicating that compiler assumes len==0 can actually happen.

Something is wrong here. Another hint is that i only see this warning in 32 bit build.

bernd-edlinger: Looks more like a compiler bug to me.
Is it gcc?

p-steuer:
```
# gcc --version
gcc (GCC) 9.1.1 20190503 (Red Hat 9.1.1-1)
```
I use the following configuration:
```
/usr/bin/perl ./Configure linux32-s390x --strict-warnings --strict-warnings enable-unit-test enable-md2 enable-rc5  enable-weak-ssl-ciphers enable-ssl3 enable-ssl3-method enable-sctp enable-asan enable-ubsan
```

bernd-edlinger: Hmm, using `gcc-Version 10.0.0 20191026 (experimental) (GCC) `
I don't get that far:
```
gcc  -I. -Iinclude -Iproviders/common/include -Iproviders/implementations/include -Icrypto/include  -DAES_ASM -DBSAES_ASM -DECP_NISTZ256_ASM -DGHASH_ASM -DKECCAK1600_ASM -DMD5_ASM -DOPENSSL_BN_ASM_GF2m -DOPENSSL_BN_ASM_MONT -DOPENSSL_BN_ASM_MONT5 -DOPENSSL_CPUID_OBJ -DOPENSSL_IA32_SSE2 -DPOLY1305_ASM -DSHA1_ASM -DSHA256_ASM -DSHA512_ASM -DVPAES_ASM -DWHIRLPOOL_ASM -DX25519_ASM -fPIC -pthread -m64 -Wa,--noexecstack -Wall -O3 -DDEBUG_UNUSED -DPEDANTIC -pedantic -Wno-long-long -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -Wswitch -Wsign-compare -Wshadow -Wformat -Wtype-limits -Wundef -Werror -Wmissing-prototypes -Wstrict-prototypes -DOPENSSL_USE_NODELETE -DL_ENDIAN -DOPENSSL_PIC -DOPENSSLDIR="\"/usr/local/ssl\"" -DENGINESDIR="\"/usr/local/lib/engines-3\"" -DMODULESDIR="\"/usr/local/lib/ossl-modules\"" -DNDEBUG  -MMD -MF crypto/libcrypto-lib-o_dir.d.tmp -MT crypto/libcrypto-lib-o_dir.o -c -o crypto/libcrypto-lib-o_dir.o crypto/o_dir.c
In file included from /usr/include/string.h:635,
                 from crypto/LPdir_unix.c:42,
                 from crypto/o_dir.c:28:
crypto/LPdir_unix.c: In function 'OPENSSL_DIR_read':
crypto/LPdir_unix.c:134:5: error: '__builtin_strncpy' offset [275, 4095] from the object at 'direntry' is out of the bounds of referenced subobject 'd_name' with type 'char[256]' at offset 19 []8;;https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Warray-bounds-Werror=array-bounds]8;;]
  134 |     strncpy((*ctx)->entry_name, direntry->d_name,
      |     ^~~~~~~
In file included from /usr/include/dirent.h:61,
                 from crypto/LPdir_unix.c:44,
                 from crypto/o_dir.c:28:
/usr/include/x86_64-linux-gnu/bits/dirent.h:33:10: note: subobject 'd_name' declared here
   33 |     char d_name[256];  /* We must not include limits.h! */
      |          ^~~~~~
make[1]: *** [crypto/libcrypto-lib-o_dir.o] Error 1
```

bernd-edlinger: Looks valid, I need to fix that first...

kroeckx:
We should probably change in (in master) to return something.

kroeckx:
But since we also deprecated it, changing the return value is probably not needed

crypto/aes/aes_ige.c
bernd-edlinger: Yeah, using arm-linux-gnueabihf-gcc, gcc-Version 10.0.0 20190907 I get now:
```
arm-linux-gnueabihf-gcc  -Icrypto -Iinclude -Iproviders/common/include  -fPIC -pthread -Wa,--noexecstack -Wall -O3 -DDEBUG_UNUSED -DPEDANTIC -pedantic -Wno-long-long -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -Wswitch -Wsign-compare -Wshadow -Wformat -Wtype-limits -Wundef -Werror -Wmissing-prototypes -Wstrict-prototypes -DOPENSSL_USE_NODELETE -DOPENSSL_PIC -DOPENSSLDIR="\"/usr/local/ssl\"" -DENGINESDIR="\"/usr/local/lib/engines-3\"" -DMODULESDIR="\"/usr/local/lib/ossl-modules\"" -DNDEBUG  -MMD -MF providers/common/ciphers/libcommon-lib-block.d.tmp -MT providers/common/ciphers/libcommon-lib-block.o -c -o providers/common/ciphers/libcommon-lib-block.o providers/common/ciphers/block.c
arm-linux-gnueabihf-gcc  -Icrypto -Iinclude -Iproviders/common/include  -fPIC -pthread -Wa,--noexecstack -Wall -O3 -DDEBUG_UNUSED -DPEDANTIC -pedantic -Wno-long-long -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -Wswitch -Wsign-compare -Wshadow -Wformat -Wtype-limits -Wundef -Werror -Wmissing-prototypes -Wstrict-prototypes -DOPENSSL_USE_NODELETE -DOPENSSL_PIC -DOPENSSLDIR="\"/usr/local/ssl\"" -DENGINESDIR="\"/usr/local/lib/engines-3\"" -DMODULESDIR="\"/usr/local/lib/ossl-modules\"" -DNDEBUG  -MMD -MF providers/common/ciphers/libcommon-lib-cipher_ccm.d.tmp -MT providers/common/ciphers/libcommon-lib-cipher_ccm.o -c -o providers/common/ciphers/libcommon-lib-cipher_ccm.o providers/common/ciphers/cipher_ccm.c
In function 'ccm_tls_cipher',
    inlined from 'ccm_cipher_internal' at providers/common/ciphers/cipher_ccm.c:359:16,
    inlined from 'ccm_stream_final' at providers/common/ciphers/cipher_ccm.c:265:9:
providers/common/ciphers/cipher_ccm.c:317:5: error: argument 2 null where non-null expected [-Werror=nonnull]
  317 |     memcpy(ctx->iv + EVP_CCM_TLS_FIXED_IV_LEN, in, EVP_CCM_TLS_EXPLICIT_IV_LEN);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from include/internal/cryptlib.h:14,
                 from providers/common/include/prov/ciphercommon.h:14,
                 from providers/common/ciphers/cipher_ccm.c:12:
providers/common/ciphers/cipher_ccm.c: In function 'ccm_stream_final':
/home/ed/gnu/arm-linux-gnueabihf-linux64/arm-linux-gnueabihf/sys-include/string.h:44:14: note: in a call to function 'memcpy' declared here
   44 | extern void *memcpy (void *__restrict __dest,
      |              ^~~~~~
make[1]: *** [providers/common/ciphers/libcommon-lib-cipher_ccm.o] Error 1
make[1]: Leaving directory `/home/ed/OPC/openssl'
```

PR #10339, opened 2019-11-02
s390x: fix x448, add x448 test vector, ctime for x25519 and x448

p-steuer:
The s390x x448 implementation does not correctly reduce non-canonical values, ie u-coordinates greater than or equal to p = 2^448 - 2^224 - 1. The first commit fixes this.

The bug went unnoticed because there is no test vector for that case. x25519 has such a test vector obtained from wycheproof but wycheproof does not have a corresponding x448 test vector. So the second commit adds a self-generated test vector for that case.

While we are at it, the third commit makes the processing of those non-canonical values constant time for x25519 and x448.

### DIFF
--- a/crypto/ec/ecx_meth.c
+++ b/crypto/ec/ecx_meth.c
@@ -855,6 +855,25 @@ static const EVP_PKEY_METHOD ed448_pkey_meth = {
 #ifdef S390X_EC_ASM
 # include "s390x_arch.h"
 
+/*
+ * constant-time conditional swap of the len-byte buffers a and b,
+ * depending on cond in {0,1}.
+ */
+static inline void s390x_cswap(unsigned char *a, unsigned char *b, size_t len,
+                               unsigned int cond)
+{
+    size_t i;
+    unsigned char tmp;
+    unsigned char mask = 0 - (unsigned char)cond;
+
+    for (i = 0; i < len; i++) {
+        tmp = a[i] ^ b[i];
+        tmp &= mask;
+        a[i] ^= tmp;
+        b[i] ^= tmp;
+    }
+}
+
 static void s390x_x25519_mod_p(unsigned char u[32])
 {
     unsigned char u_red[32];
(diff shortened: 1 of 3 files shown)


levitte:
Nice description. It should end up in the commit messages, at least for the first and second commit

p-steuer:
> It should end up in the commit messages, at least for the first and second commit

Done.

I dont agree with you title change: Its actually x448 which has a bug (not x25519) and the x448 test vector is added to avoid regressions.
Only the constant-time changes also affect x25519.

levitte:
> I dont agree with you title change

Totally fine. I misread it as "... x448 and x448..." instead of "... x448 and *add* x448...", so took it for a typo

p-steuer:
travis red corss not relevant

crypto/ec/ecx_meth.c
```diff
@@ -855,6 +855,25 @@ static const EVP_PKEY_METHOD ed448_pkey_meth = {
...
+    unsigned char tmp;
+    unsigned char mask = 0 - (unsigned char)cond;
+
+    for (i = 0; i < len; i++) {
+        tmp = a[i] ^ b[i];
+        tmp &= mask;
+        a[i] ^= tmp;
+        b[i] ^= tmp;
+    }
+}
```
mattcaswell: It seems like this function belongs in `constant_time.h`. We already have uint32_t and uint64_t versions. This just adds another variant.

crypto/ec/ecx_meth.c
```diff
@@ -966,7 +981,7 @@ static int s390x_x448_mul(unsigned char u_dst[56],
     memcpy(param.x448.d_src, d_src, 56);
 
     s390x_flip_endian64(param.x448.u_src, param.x448.u_src);
-    s390x_x448_mod_p(param.x448.u_src);
+    s390x_x448_mod_p(param.x448.u_src + 8);
```
mattcaswell: I'm struggling to understand what's going on in this change and the other related change in your first commit. Can you explain in a bit more detail what is happening here?

p-steuer: The structure of the parameter block the pcc instruction works on is reflected by the param union. The 56 byte values x448 operates on are stored right-aligned in its 64 byte fields. The x448_mod_p reduction function works on 56 bytes, therefore it must start on an 8 bytes offset into the buffer. The bug was that without the offset, the leading 8 bytes were always zero, so reduction never kicked in.

crypto/ec/ecx_meth.c
p-steuer: okay, i will move it there.

crypto/ec/ecx_meth.c
p-steuer: The second bug in the reduction (which was obv. shadowed by the first bug) was that i should have checked one bit higher for overflow at the end of the reduction (charck the least significant carry bit instead of most significant bit in most significant byte of u_red).

crypto/ec/ecx_meth.c
p-steuer: done

p-steuer:
merged.

master: https://github.com/openssl/openssl/commit/58738b1cad8777230226009a6bb4efa3def22216 https://github.com/openssl/openssl/commit/6376c229c44a355248db17e9f0bb2e4567a16d0d https://github.com/openssl/openssl/commit/677c4a012a7e72b5f2dd239639034f01fad850bf

PR #10343, opened 2019-11-03
Fix a -Warray-bounds gcc warning in OPENSSL_DIR_read

bernd-edlinger:
Actually since direntry->d_name is guaranteed to be zero-terminated,
it does not matter that entry_name is larger.

### DIFF
--- a/crypto/LPdir_unix.c
+++ b/crypto/LPdir_unix.c
@@ -131,9 +131,8 @@ const char *LP_find_file(LP_DIR_CTX **ctx, const char *directory)
         return 0;
     }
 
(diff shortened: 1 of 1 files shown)


kroeckx:
I assume the warning is actually wrong, but it's a good chance anyway.

bernd-edlinger:
Yes, the warning does not know that the src is guaranteed to be zero-terminated,
There is even an attribute IIRC, to tell gcc that the string is zero-terminated.
But we fixed similar places in a similar way...

bernd-edlinger:
Merged to master and 1.1.1, thanks!

PR #10348, opened 2019-11-04
Rework ordinals

levitte:
The ordinals file are tedious to handle when there are merge conflicts, with all the renumbering that always has to happen.

During the last OpenSSL face2face, this was an item of discussion, and we concluded that during development, these fixed positions aren't important, as we don't guarantee in any way that a symbol location in a transfer vector stays the same.  However, as soon as features are essentially frozen, i.e. in beta or final releases, these numbers are important, at least on operating systems where the location in a transfer vector is still important for shared libraries and the like.

So we introduce a couple of markers for symbols that haven't been assigned a number yet:

    ?   signifying it gets the previous symbol's number plus one
    ?+  signifying it gets the same number as the previous symbol

That means that for example, one of the last symbols in `util/libssl.num`, which has this line:

    OSSL_default_ciphersuites               509    3_0_0   EXIST::FUNCTION:

is changed to:

    OSSL_default_ciphersuites               ?      3_0_0   EXIST::FUNCTION:

At beta or final release, whichever comes first (for 3.0, that's going to be the first beta), such symbols will be assigned a fixed number again, which happens to be suitable at the time.


paulidale (dismissed):
Was the idea of sorting the ordinals alphabetically also suggested to reduce the number of merge conflicts further?

Adding new items to the end of the .num file will still result in conflicts.  They'll be easier to fix at least.

levitte:
I don't recall that, and chances are that it won't reduce conflicts that much.
Maybe in another iteration, but that's much harder to deal with.

levitte:
Test added

mattcaswell:
> I don't recall that, and chances are that it won't reduce conflicts that much.
Maybe in another iteration, but that's much harder to deal with.

Yes, that was discussed. My guess is that will reduce conflicts significantly.

levitte:
How about we start with this?  Meanwhile, I'll have another think on the alphabetical sort (and if I understand correctly, only of the symbols that are yet unassigned, right?), and save that for another PR.

paulidale:
Why just the new symbols?  Sorting is cheap and the more that is sorted, the fewer the collisions.

paulidale (approved):

mattcaswell (approved):
I'm fine with this as-is. I don't think it will reduce any conflicts though until we also add sorting....but I'm ok with leaving that until another PR.

mattcaswell:
> Why just the new symbols? Sorting is cheap and the more that is sorted, the fewer the collisions.

I tend to agree with @paulidale here.

levitte:
Ohhhhhhh....  While preserving numbers, right?  A little like we do with error strings, yeah?

Yeahok, I see how that would work, with just one exception; aliases, where there are needed, will be quite hard to deal with.  The `?+` marker assumes attachment to the previous line (which is presumably marked with `?`), and that assumption doesn't hold after a shuffle.
(aliases are necessary for the rare case where we have symbols that differ from others only by casing, see `include/openssl/symhacks.h` for the gory details)

But, *other PR*

mattcaswell:
> Ohhhhhhh.... While preserving numbers, right?

Right!

richsalz:
I am all in favor of sorting.  I suggest ```=otherName``` for "aliases" so that they are not dependant on the other.

levitte:
> I am all in favor of sorting. I suggest `=otherName` for "aliases" so that they are not dependant on the other.

That...  was a workable idea!

levitte:
Merged.

5d61758ee70aebb479d6f7a43483af405c0031a2 util/*.num: deassign ordinal numbers from new symbols
3da95f3c5159b8404917e3ff349566488c28f798 OpenSSL::Ordinals: Handle symbols with unassigned ordinal numbers
81ddd952ea1c4f80df45435fb927fd0c6314a008 OpenSSL::Ordinals: add a renumber() function, to assign unassigned symbols
a4aab7871939d3661557fd843e0d411bbaa78bc6 OpenSSL::Ordinals: when validating, collect statistics on unassigned syms
e4f2d539f631060654faf3b45f54e222637109ac util/mkdef.pl: writer_VMS(): handle symbols with no assigned number
8635730333af792fae75394bb593afe4054da238 util/mknum.pl: Call OpenSSL::Ordinals::renumber() for real releases
b6fc6620cff7e5971757ad1c50edd022fdaffc04 util/mknum.pl: output stats on unassigned symbols
6af1b11848f000c900877f1289a42948d415f21c test/recipes/02-test_ordinals.t: Take '?' and '?+' into account

kaduk:
I confess I'm still a little confused at when ```?+``` would be used; is there a simple-to-digest example?

levitte:
It's for symbols that are aliases for other symbols. Nowadays, it's all derived from include/openssl/symhacks.h. If you take any of the symbols there and do a case insensitive grep, you might seen why those hacks are there at all, at least if you bear in mind that on some systems, symbol names are normally viewed *without* regard for case.

In util/libcrypto.num, this translates into symbols with the same ordinal number (whenever they get one assigned)

levitte:
We have become much better lately at not having symbols that only differ by case, so I hope that some day, symhacks.h will simply go away.

DDvO:
@Akretsch and myself are very glad about this pragmatic simplification!
In particular when a PR that adds items to `libcrypto.num` needs to be rebased several times over a longer period, handling spurious merge conflicts was a PITA. This burden is gone now :)

mattcaswell:
> In particular when a PR that adds items to libcrypto.num needs to be rebased several times over a longer period, handling spurious merge conflicts was a PITA. This burden is gone now :)

Unfortunately its not enough by itself to avoid constant rebases :-(
It *is* a good step towards that, but we additionally need the .num files to be sorted alphabetically instead of numerically to finish the job. That's a bit trickier though.

PR #10349, opened 2019-11-04
Remove build dependency of kernel netinet/sctp.h in bss_dgram.c

NaveenShivanna86:
After adding 'enable-sctp' compile option, OpenSSL (DTLS) can work with
SCTP as transport.

OpenSSL bss_dgram.c file includes the kernel /netinet/sctp.h.

If we have our own custom SCTP implementation (which also implements  custom BIO
METHODS, do not use the default methods), then we still need netinet/sctp.h header file to compile OpenSSL. Our build environment do not have the sctp.h and we are not supposed to install lksctp-tools.

So tailoring the bss_gram.c source file such that netinet/sctp.h is not required to include whenever there is custom SCTP (not kernel SCTP)

Discussion thread: https://www.mail-archive.com/openssl-users@openssl.org/msg87036.html

### DIFF
--- a/crypto/bio/bss_dgram.c
+++ b/crypto/bio/bss_dgram.c
@@ -14,7 +14,9 @@
 #ifndef OPENSSL_NO_DGRAM
 
 # ifndef OPENSSL_NO_SCTP
+# ifndef OPENSSL_BUILD_WO_LKSCTP
 #  include <netinet/sctp.h>
+# endif
 #  include <fcntl.h>
 #  define OPENSSL_SCTP_DATA_CHUNK_TYPE            0x00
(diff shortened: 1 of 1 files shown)


NaveenShivanna86:
Requesting to review.

mattcaswell:
I wonder whether a slight redefinition of what "no-sctp" actually means might be a better answer. The primary goal of that option is to avoid having to add additional external dependencies. We could have it not build the BIO parts, but have the libssl parts always included unconditionally (because they have no dependence on "netinet/sctp.h"). Functions like `BIO_dgram_is_sctp()` and `BIO_dgram_sctp_wait_for_dry` would still have to always be built. That way if an application happens to have its own SCTP capable BIO it can just use it with a standard build of OpenSSL

PR #10363, opened 2019-11-05
add util/check-format.pl for checking adherence to C coding style

DDvO:
When contributing PRs to OpenSSL it typically happens that the code does not (fully) adhere to the coding style guidelines at https://www.openssl.org/policies/codingstyle.html and to further implicit conventions that are not (yet) part of the rules stated there.

In particular when we contributed larger chunks of code, reviewers typically found many formatting issues that are usually trivial to resolve but tend to cause extra overhead and distract from more important issues.

There is already a tool that is supposed to automatically (re-)format code: `util/openssl-format-source`, yet this appears rather outdated and not really practical to use.
So I spent some effort on a new Perl tool that aims at checking adherence at least to most of the OpenSSL formatting rules we are aware of.

This new tool is in preliminary state: it is incomplete and yields some false positives.
Still it has already proved pretty useful for detecting most typical formatting glitches.


richsalz:
Cool!  I think part of this PR should be to remove the old script.

mattcaswell:
Nice!

levitte:
I agree with @richsalz

DDvO:
Pleased to hear!
I invite you to try it out and potentially give more detailed feedback.

util/openssl-format-check.pl
```diff
@@ -0,0 +1,293 @@
...
+    $in_multiline_macro = 0;
+    $in_multiline_comment = 0;
+}
+
+reset_file_state();
+while(<>) {
+    $line++;
+    my $orig_ = $_;
+
+    if(m/[\x09]/) {
```
richsalz: A block comment like "# Look for illegal characters" before clumps of lines would be helpful.  (I prefer that to inline right-margin comments but YMMV)

Especially starting around 132 etc.

levitte:
I tried it on `crypto/store/*.[ch]`, and got a few interesting nits:

-----

```
crypto/store/loader_file.c:225:indent=16!={3,20}:                 || PKCS12_verify_mac(p12, NULL, 0)) {
```

Here's the line with a bit of context:

https://github.com/openssl/openssl/blob/6af1b11848f000c900877f1289a42948d415f21c/crypto/store/loader_file.c#L224-L226

So it seems that this scripts want to make the extra indent "mandatory", while this isn't policy (and is still subject for debate).  We *allow* that extra indentation, so this script should currently accept both styles, i.e. both this:

``` C
            if (PKCS12_verify_mac(p12, "", 0)
                || PKCS12_verify_mac(p12, NULL, 0)) {
                pass = "";
```

and this:

``` C
            if (PKCS12_verify_mac(p12, "", 0)
                    || PKCS12_verify_mac(p12, NULL, 0)) {
                pass = "";
```

-----

```
crypto/store/loader_file.c:281:indent=5!=1:      p12_end:
```

We have this kind of hanging indent of labels all over the place, i.e. a hang back of 3 columns, not 4.  This came to be with our `util/openssl-format-source` with the auxilliary file `util/indent.pro`, where you find this:

https://github.com/openssl/openssl/blob/6af1b11848f000c900877f1289a42948d415f21c/util/indent.pro#L15

The indent manual explains it like this:

> **-il**_n_, **--indent-label**_n_
> Set offset for labels to column _n_.
> See  **INDENTATION**.

Why this was a good idea at the time, I cannot remember.

-----

```
crypto/store/loader_file.c:1109:indent=8!=4:         : PEM_read_bio(bp, pem_name, pem_header, data, len);
```

So this is a complete mystery to me.  This is the line with a bit of context:

https://github.com/openssl/openssl/blob/6af1b11848f000c900877f1289a42948d415f21c/crypto/store/loader_file.c#L1106-L1110

-----

```
crypto/store/loader_file.c:1419:indent=8!={4,4}:         "file",
crypto/store/loader_file.c:1420:indent=8!={4,4}:         NULL,
crypto/store/loader_file.c:1421:indent=8!={4,4}:         file_open,
crypto/store/loader_file.c:1422:indent=8!={4,4}:         file_ctrl,
crypto/store/loader_file.c:1423:indent=8!={4,4}:         file_expect,
crypto/store/loader_file.c:1424:indent=8!={4,4}:         file_find,
crypto/store/loader_file.c:1425:indent=8!={4,4}:         file_load,
crypto/store/loader_file.c:1426:indent=8!={4,4}:         file_eof,
crypto/store/loader_file.c:1427:indent=8!={4,4}:         file_error,
crypto/store/loader_file.c:1428:indent=8!={4,4}:         file_close
```

This seems to be some kind of fault in how indentation is calculated.  Here are those lines plus a little extra:

https://github.com/openssl/openssl/blob/6af1b11848f000c900877f1289a42948d415f21c/crypto/store/loader_file.c#L1417-L1430

So er, the actual complaint would really be that the braces are on a line of their own...  right?

util/openssl-format-check.pl
DDvO: I've meanwhile added block comments and improved some comments and variable names.

DDvO:
> I tried it on `crypto/store/*.[ch]`, and got a few interesting nits:

Thanks for trying it out and your detailed feedback!
Meanwhile I've handled all issues your reported: fix false positives and allow for both possibilities where the specification is unclear.

> So it seems that this scripts want to make the extra indent "mandatory", while this isn't policy (and is still subject for debate). We _allow_ that extra indentation, so this script should currently accept both styles

Done.

> We have this kind of hanging indent of labels all over the place, i.e. a hang back of 3 columns, not 4. This came to be with our `util/openssl-format-source` with the auxilliary file `util/indent.pro`, 

I've relaxed the check for label indents to allow them both at column 1 and (current_indent-3).

> ```
> crypto/store/loader_file.c:1109:indent=8!=4:         : PEM_read_bio(bp, pem_name, pem_header, data, len);
> ```
> So this is a complete mystery to me.

This was a false positive and went away with my today's improvements.
I know that handling of conditional expressions ( ? : ) is still to be improved.

> This seems to be some kind of fault in how indentation is calculated. Here are those lines plus a little extra:
> 
> https://github.com/openssl/openssl/blob/6af1b11848f000c900877f1289a42948d415f21c/crypto/store/loader_file.c#L1417-L1430
> 
> So er, the actual complaint would really be that the braces are on a line of their own... right?

This was due to incomplete detection of complex multi-line assignments involving braces; I've strongly improved this (also for the rare case of constant value decls within `enum`).

richsalz:
This work is amazing.

beldmit:
Great job!

DDvO:
Thanks!
I was quite a piece of work to get it that far.

DDvO:
I've just fixed two small further issues, preventing false positives also on `crypto/{cmp,crmf}/*.[ch]`

DDvO:
I've improved the handling of comments and string literals 
and added TODOs on multi-line string literals.

mspncp:
Note: looking at the util directory, many of the scripts are named using imperative style ("find doc nits", "merge err lines", "format source", "find unused errs", etc.) and not a (typical german?) nominal style  as in "openssl format check". Also most of those scripts don't have a .pl extension. 

How about just calling it `openssl-check-format`?

mspncp:
Or `openssl-check-coding-style`?

DDvO:
I like `openssl-check-format`, 
while "check coding style" would be too broad, at least for the current state of the tool.

DDvO:
Should I remove the `.pl` suffix?

DDvO:
I there sufficient agreement to remove the old script (`openssl-format-source`)?

DDvO:
I've meanwhile resolved the TODOs on handling multi-line string literals.

mspncp:
> Should I remove the .pl suffix?

It‘s not a requirement, only a suggestion based on recent habits. Unless the script works and is intended to be used on Windows too, I see no reason to keep it.

richsalz:
I think the "openssl-" prefix wastes needless space.  After all, this is in the openssl source distribution. :)  I like "check-formatting" or something like that.  Yes, remove the .pl extension. As for removing the old file, an OMC member said yes; https://github.com/openssl/openssl/pull/10363#issuecomment-549823550

DDvO:
`check-format` is even shorter than  `check-formatting` :)
Since the script certainly works on Windows as well (as long as Perl is installed), I suggest keeping the `.pl` suffix.

So my proposal is to go for `check-format.pl`.

richsalz:
One last attempt on the .pl extension: Most development is on unix-like platforms; those windows developers can put "perl" in front of the command-line. :)

mspncp:
> So my proposal is to go for `check-format.pl`.

O.k. It does have the advantage that the script can be invoked on Windows without prefixing it with `perl`.

DDvO:
I'm also quite in favor of not puting an extra burden on those poor guys (sometimes including myself) developing in Windows ;)

For Unix use, we have the first line referring to 'Perl', and on Windows the name extension usually does the trick.
Moreover, there are various Perl scripts in `util/` that do have the `.pl` ending as well.

richsalz:
:)

mspncp:
cool. Adding a unit test for a utility script seems to be unprecedented.

DDvO:
Yeah, given the non-trivial nature of the thing,
I felt that tests assuring that there are only known false negatives is important. I'll add test cases for more interesting indentation cases in complex expressions etc. soon.
On the other hand, false positives are rather obvious when running the tool on actual source files, so I do not plan to add test cases for these.

BTW, meanwhile I've further reduced both false positives and false negatives; some more of this is to come.

util/check-format-test.c
```diff
@@ -9,29 +9,66 @@
  * https://www.openssl.org/source/license.html
  */
 
-/* this demonstrates/tests cases where check-format.pl should complain */
+/* This demonstrates/tests cases where check-format.pl should complain */
+/* In total, 55 formattiing issues should be reported for this file */
```
mspncp: I‘m not sure whether it’s a good idea to hardcode the expected  number of  reported issues in a comment. People will forget to update it. Better make it a variable (expected_issues) and add a test condition for it. 

typo: formatting

DDvO: Indeed stating the number of expected issues cannot be the real solution.
I'll add a tag a per issue in the test input and these tags will be counted automatically.

richsalz:
Long thread :)  4bac25e1115b8c613f9fff12b835aca47e2bdef7 just merged code with trailing whitespace; perhaps check for that, too?

mspncp:
> Long thread :) [4bac25e](https://github.com/openssl/openssl/commit/4bac25e1115b8c613f9fff12b835aca47e2bdef7) just merged code with trailing whitespace; perhaps check for that, too?

@richsalz  grepping for trailing whitespace I found a lot of page breaks (`^L`), most of them in perl code, but a few also in c files. Do they still serve a useful purpose, or can those be removed, too?

mspncp:
> ...grepping for trailing whitespace...

`git grep -E '^[[:space:]]+$'`

levitte:
^L does have a useful purpose for anyone that wants to print the source.

mspncp:
> ^L does have a useful purpose for anyone that wants to print the source.

One reason more to remove it: to save the trees by preventing people from printing out our source code ;-)

levitte:
Pdf, epub, ... (reading pages instead of a gazillion lines is a list art) 😉

richsalz:
"git diff" highlights trailing spaces; I don't know if it highlights the formfreed character. I was only suggesting space, *not* whitespace.   So ```[ ]$`` :)

Horrors, it shows find-doc-nits.  The shame, the shame.

DDvO:
> Long thread :) [4bac25e](https://github.com/openssl/openssl/commit/4bac25e1115b8c613f9fff12b835aca47e2bdef7) just merged code with trailing whitespace; perhaps check for that, too?

The new tool does already check for trailing whitespace, any further non-printable stuff including ^L, and trailing empty/all-whitespace lines.
Yet it is meant for checking C source (including header) files, so for Perl code it makes little sense.

DDvO:
I've just pushed major improvements of the tool.
It now does proper checking of indentation in complex multi-line expressions/statements/declarations.
Moreover, I've added checks for various space nits such as superfluous or missing spaces.
I also reduced the number of false positives, while many more true positives are found.

I've tested this with `crypto/store/*.[ch]`, 
`crypto/{crmf,cmp}/*.[ch]`, and `include/openssl/{crmf.cmp}.h`.

DDvO:
Since certain types of findings may be tolerable and
the number of findings may be overwhelming at first,
I've added a couple of options that can be used to disable complaints on certain mistakes:
* `-l | --sloppy-len` increases accepted max line length from 80 to 84
* `-s | --sloppy-space` disables reporting of whitespace nits
* `-e | --sloppy-expr` adds some grace when checking multi-line expr/stmt/decl indentation

DDvO:
Nevertheless the tool can be (ab-)used to find low-level issues also in Perl scripts, e.g.;
`util/check-format.pl util/*.pl | grep -v -E 'indent|too many|unclosed|://|\*/|/\*|:\{|\}:' `

levitte:
Where do we stand with this.  @DDvO, do you feel it's ready for final review, or are you still actively working on it?

util/check-format-test.c
DDvO: I've added the tags in `check-format-test.c` in a way such that the script can use them for self-tests.

DDvO:
> Where do we stand with this? @DDvO, do you feel it's ready for final review, or are you still actively working on it?

Good that you ask. 
I've meanwhile done numerous further (small) improvements and started cleaning up the code.
Would be good to get further feedback due to experimental use, which I could then take into account.
After that it a review certainly will make sense, but I'd welcome feedback on the code already now.

levitte:
Cool. I'll try to remember to play with it in days to come.

util/check-format-test.c
```diff
@@ -20,9 +20,9 @@
 
 /* For each of the following set of lines the tool should complain */
 /*@ tab character: 	 */
-/*@ cr character:  */
+/*@ carriage return character:  */
```
levitte: I wonder what happens with this on macos, which classically has CR as line termination character...

levitte: Not that it matters much... Perhaps it's enough with a warning somewhere that this won't be properly detected on all platforms

DDvO: CR characters are orthogonal to line termination of macros. Multi-line macro definitions use a backslash `\` at/near line end, and these are already properly handled by the tool.

On Windows systems, line ends have a CR just before the LF character, while on Unix-like systems only LF is used. The tool so far complained about all CR characters, which was not suitable for line ends on Windows. I've changed this such that only intra-line CRs are reported, which - just like TAB and FF characters - should be avoided in source code (though from the C perspective these are just other forms of whitespace) regardless of the platform.

levitte:
I did a run with `apps/*.[ch]` and found the following false positives:

*I would like to point out that this run also found a huge number of actual issues, so there's a lot of good*

-----

```
apps/ca.c:29:'#if' nesting=1!=2:# elif !defined(OPENSSL_SYS_VXWORKS) && !defined(OPENSSL_SYS_WINDOWS)
```

https://github.com/openssl/openssl/blob/a7bdd0433c0360fccb4dd9f5c43c6fabc4e8c659/apps/ca.c#L26-L32

It looks to me like `elif` isn't quite correctly processed

-----

```
apps/ca.c:47:dbl SPC:#define BASE_SECTION            "ca"
apps/ca.c:49:dbl SPC:#define ENV_DEFAULT_CA          "default_ca"
apps/ca.c:77:dbl SPC:#define ENV_DATABASE            "database"
```

I do notice this comment in the script, but am not sure I understand how it applies here:

``` perl
# * Use of multiple consecutive spaces is regarded a coding style nit except
#   when done in order to align certain columns over multiple lines, e.g.:
#   # define AB  1
#   # define CDE 22
#   # define F   3333
#   This pattern is recognized - and consequently double space not reported -
#   for a given line if in the line before or after (as far as these exist)
#   for each occurrence of "  \S" (where \S means non-space) in the given line
#   there is " \S" in the other line in the respective column position.
#   This may lead to both false negatives (in case of coincidental " \S")
#   and false negatives (in case of more complex multi-column alignment).
```

-----

```
apps/ca.c:539:no SPC'*'/:    /*****************************************************************/
apps/ca.c:539:'/*'no SPC:    /*****************************************************************/
```

Errrrrrrr....  where did this come from?  There's *nothing* saying you *must* have a space efter `/*` or before `*/`  (as a matter of fact, you will find `/*-` in a number of places, which is an indicator for `indent` to leave that comment alone)

However, considering our demands on block comments, I can see where this comes from, but the error message doesn't quite say that, i.e. is quite confusing.  And I can recognise the difficulty here, since this is a style of multi-line comment formed with multiple single-line comments.

-----

```
apps/ca.c:753:dbl SPC:    /* Read extensions config file                                   */
```

I think this should be relaxed for comments.  We have plenty of places where there are double spaces, for example in the license boilerplate at the top of C source file.

I have a better example, though:

```
apps/s_client.c:3505:dbl SPC in multi-line comment:     *              extendedResp          ExtendedResponse,
apps/s_client.c:3507:dbl SPC in multi-line comment:     *         controls       [0] Controls OPTIONAL }
apps/s_client.c:3516:dbl SPC in multi-line comment:     *              success                      (0),
apps/s_client.c:3518:dbl SPC in multi-line comment:     *              other                        (80),
```

https://github.com/openssl/openssl/blob/a7bdd0433c0360fccb4dd9f5c43c6fabc4e8c659/apps/s_client.c#L3498-L3523

There's no reason why this shouldn't be permitted.
(although, it's curious that we didn't make that block comment start with `/*-`)

-----

```
apps/ca.c:2054:normal stmt indent=16!=12:                buf++;
apps/ca.c:2055:normal stmt indent=16!=12:                if (*buf)
apps/ca.c:2056:extra  stmt indent=20!=16:                    type = buf;
apps/ca.c:2057:normal stmt indent=16!=12:                break;
apps/ca.c:2058:normal stmt indent=12!=8:            }
```

https://github.com/openssl/openssl/blob/a7bdd0433c0360fccb4dd9f5c43c6fabc4e8c659/apps/ca.c#L2052-L2058

```
apps/enc.c:324:normal stmt indent=12!=8:            if (enc)
apps/enc.c:325:extra  stmt indent=16!=12:                outformat = FORMAT_BASE64;
apps/enc.c:326:normal stmt indent=12!=8:            else
apps/enc.c:327:extra  stmt indent=16!=12:                informat = FORMAT_BASE64;
apps/enc.c:328:normal stmt indent=8!=4:        }
```

https://github.com/openssl/openssl/blob/a7bdd0433c0360fccb4dd9f5c43c6fabc4e8c659/apps/enc.c#L323-L328

It looks like the parser doesn't quite handle the mixture of single line and multi-line bodies.

-----

```
apps/dhparam.c:330:hanging '(' indent=24!=19:                        "    BIGNUM *p, *g;\n"
apps/dhparam.c:331:hanging '(' indent=24!=19:                        "\n"
apps/dhparam.c:332:hanging '(' indent=24!=19:                        "    if (dh == NULL)\n"
apps/dhparam.c:333:hanging '(' indent=24!=19:                        "        return NULL;\n");
apps/dhparam.c:339:hanging '(' indent=24!=19:                        "            || !DH_set0_pqg(dh, p, NULL, g)) {\n"
apps/dhparam.c:340:hanging '(' indent=24!=19:                        "        DH_free(dh);\n"
apps/dhparam.c:341:hanging '(' indent=24!=19:                        "        BN_free(p);\n"
apps/dhparam.c:342:hanging '(' indent=24!=19:                        "        BN_free(g);\n"
apps/dhparam.c:343:hanging '(' indent=24!=19:                        "        return NULL;\n"
apps/dhparam.c:344:hanging '(' indent=24!=19:                        "    }\n");
apps/dhparam.c:347:hanging '(' indent=24!=23:                        "    if (!DH_set_length(dh, %ld)) {\n"
apps/dhparam.c:348:hanging '(' indent=24!=23:                        "        DH_free(dh);\n"
apps/dhparam.c:349:hanging '(' indent=24!=23:                        "        return NULL;\n"
apps/dhparam.c:350:hanging '(' indent=24!=23:                        "    }\n", DH_get_length(dh));
```

*This* one is hard, that should be recognised.  Basically, partial strings have been aligned to make the intended output more readable:

https://github.com/openssl/openssl/blob/a7bdd0433c0360fccb4dd9f5c43c6fabc4e8c659/apps/dhparam.c#L329-L351

I don't know exactly how you should do this.  Is there a way to detect strings in the source along with their position and make an exception of they are aligned with each other?

We have exactly the same issue here:

https://github.com/openssl/openssl/blob/a7bdd0433c0360fccb4dd9f5c43c6fabc4e8c659/apps/dsaparam.c#L206-L223

... and here:

https://github.com/openssl/openssl/blob/a7bdd0433c0360fccb4dd9f5c43c6fabc4e8c659/apps/ecparam.c#L365-L406

-----

```
apps/passwd.c:37:'{' @ EOL:static const char ascii_dollar[] = { 0x24, 0x00 };
```

https://github.com/openssl/openssl/blob/a7bdd0433c0360fccb4dd9f5c43c6fabc4e8c659/apps/passwd.c#L37

Errr....  what?

...  this is also a call for less cryptic messages.  *You* know what that means, I'm sure, but for anyone else, that was incomprehensible.  Is that trying to say that a `{` must be the last character on the line?

Either way, this line is absolutely permissible as far as I know.

-----

```
apps/rehash.c:93:extra  stmt indent=0!=8:};
apps/rehash.c:93:unexpected '}' @ expr:};
apps/rehash.c:95:normal stmt indent=0!=4:enum Hash {
apps/rehash.c:96:normal stmt indent=4!=8:    HASH_OLD, HASH_NEW, HASH_BOTH
apps/rehash.c:97:normal stmt indent=0!=4:};
```

https://github.com/openssl/openssl/blob/a7bdd0433c0360fccb4dd9f5c43c6fabc4e8c659/apps/rehash.c#L90-L97

It seems the parser got lost here.  Is it the lack of comma (perfectly valid!) at the end of the enumerations?

There was no point looking at the rest of the messages for this file....

-----

```
apps/req.c:35:dbl SPC:#define SECTION         "req"
apps/req.c:40:dbl SPC:#define DISTINGUISHED_NAME      "distinguished_name"
```

These are a bit interesting.  The `SECTION` definition is separated from the rest by an empty line, but its value is aligned with the definitions following.  I can see why you'd separate the section name from the rest like that.  As for `DISTINGUISHED_NAME`, that macro name is so long it pushes the value to the right, making it an exception of alignment.  Should all values be pushed to the right because of that?  Possibly, but if you look at really long names like we have in `include/openssl/core_names.h`, it can get ridiculous, and having the majority of the values *too* far away from the macro name might make it less readable (the further apart they are, the harder it is to see that they belong together, and it's even harder on people with astigmatism).

I'm not sure how to solve this in a sensible manner...  it's quite possible that going the other way and relaxing the double space rules on macro definition is the way to preserve our sanity (i.e. instead of bending over backwards to try and add a pile of alignment rules, just don't check)

-----

```
apps/req.c:499:no leading '*' in multi-line comment:                BIO_printf(bio_err,"problems opening %s for extra oid's\n",p);
apps/req.c:499:multi-line comment indent=16!=17:                BIO_printf(bio_err,"problems opening %s for extra oid's\n",p);
```

This is a comment that starts with  `/*-`, i.e. a marker we usse for `indent` to leave that comment alone.  I think you should do the same

-----

```
apps/speed.c:1575:normal stmt indent=4!=8 or 12 for '||':    || !defined(OPENSSL_NO_EC)
```

I can't even begin to understand this one.


https://github.com/openssl/openssl/blob/a7bdd0433c0360fccb4dd9f5c43c6fabc4e8c659/apps/speed.c#L1574-L1575

-----

```
apps/speed.c:1579:hanging '{' indent=36!=35 or 8 for lines after '{':                                    ECDSA_SECONDS, ECDH_SECONDS,
apps/speed.c:1580:hanging '{' indent=36!=35 or 8 for lines after '{':                                    EdDSA_SECONDS, SM2_SECONDS };
```

Interesting!  So here it seems that values are permitted after a `{` (contrary to the case with the array, see the case in `apps/passwd.c` above).  However, if I understand this correctly, you don't allow a space between the `{` and the first value, is that it?  It *is* permissible...  however, our habit of avoiding that space in apps option arrays is becaue `indent` would otherwise reformat them in an ugly manner.

https://github.com/openssl/openssl/blob/a7bdd0433c0360fccb4dd9f5c43c6fabc4e8c659/apps/speed.c#L1578-L1580

That being said, this could actually be reformatted more nicely, like this:

``` C
    openssl_speed_sec_t seconds = {
        SECONDS, RSA_SECONDS, DSA_SECONDS,
        ECDSA_SECONDS, ECDH_SECONDS,
        EdDSA_SECONDS, SM2_SECONDS
    };
```

-----

```
apps/x509.c:788:no SPC'/*':                print_name(out, "/*\n"
apps/x509.c:788:'/*'no SPC:                print_name(out, "/*\n"
apps/x509.c:788:text after '/*' in multi-line comment:                print_name(out, "/*\n"
apps/x509.c:789:no leading '*' in multi-line comment:                                " * Subject: ", X509_get_subject_name(x), get_nameopt());
apps/x509.c:789:len=90>80:                                " * Subject: ", X509_get_subject_name(x), get_nameopt());
apps/x509.c:789:multi-line comment indent=32!=34:                                " * Subject: ", X509_get_subject_name(x), get_nameopt());
apps/x509.c:790:no leading '*' in multi-line comment:                print_name(out, " * Issuer:  ", X509_get_issuer_name(x), get_nameopt());
apps/x509.c:790:len=89>80:                print_name(out, " * Issuer:  ", X509_get_issuer_name(x), get_nameopt());
apps/x509.c:790:multi-line comment indent=16!=34:                print_name(out, " * Issuer:  ", X509_get_issuer_name(x), get_nameopt());
apps/x509.c:791:no leading '*' in multi-line comment:                BIO_puts(out, " */\n");
apps/x509.c:791:text before '*/' in multi-line comment:                BIO_puts(out, " */\n");
apps/x509.c:791:multi-line comment indent=16!=34:                BIO_puts(out, " */\n");
```

https://github.com/openssl/openssl/blob/a7bdd0433c0360fccb4dd9f5c43c6fabc4e8c659/apps/x509.c#L788-L791

Maybe trying to check comments in strings is over-doing it a bit? :wink:

levitte:
Regarding this comment in the script:

``` perl
# * There is the special OpenSSL rule not to unnecessarily use braces around
#   a single statement :
#   {
#       single statement;
#   }
#   except within if .. else constructs where some branch contains more than one
#   statement. The exception is not recognized - and thus false positives are
#   reported - when such a branch occurs after the current position.
#   Moreover, false negatives occur if the braces are more than two lines apart.
#
```

There are a *lot* of such false positives.  This, along with the issue with mixed one-line and multi-line bodies suggest to me that the parser might need to deal with *bodies* of `if` / `else if` / `else`, `for`, `while`, `switch` statements differently, i.e. still parse them, but then retain information for higher level analysis.  A sequence of `if` etc could be reduced to `if (cond) single-stmt; else if (cond) multi-stmt; else single-stmt;`, where a mixture of `single-stmt` and `multi-stmt` would be detected and warned upon.  Likewise, I think we frown upon things like this:

``` C
    for (init; cond; step)
        if (cond) {
            stmt1;
            stmt2;
        }
```

the argument being that if the body of an `if`, `for` or `while` is complex enough, braces should be added.  A reduction of the above would be `for (init; cond; step) if (cond) multi-stmt;`, and one might say that a body that contains a `multi-stmt` can't be reduced to a `single-stmt`, making it impossible to reduce that to a `for (init; cond; step) single-stmt;`

Makes sense?

DDvO:
@levitte, yes, the known false positives on mixed one-line and multi-line `if` and `else` bodies (while checking the rule to leave out braces around single statements) could be avoided by retaining more information across lines for later higher-level analysis.

Unfortunately the situation is in general more complex than the solution sketched above could handle due to potential nesting of `if` ... `else`. So we'd need a kind of AST (abstract syntax tree) for the block/statement-level structure of each function body. Implementing this is certainly possible, but beyond the effort I can spend on the tool - it already has taken me much more time than expected. 

I've just implemented an itermediate 'solution' that prefers false negatives over false positives: 
By default the single-statement rule is checked only for `do`, `while`, and `for` bodies. 
The previous, more aggressive, behavior is still available via the new `-1` (or `--1-stmt`) option.

DDvO:
Thanks a lot @levitte also for your detailed feedback after trying out (a rather recent version of) the tool on `apps/*.[ch]`.

> _I would like to point out that this run also found a huge number of actual issues, so there's a lot of good_

Pleased to hear :-)


> It looks to me like `elif` isn't quite correctly processed

Yeah, this was because I confused the keyword with `elsif` as it is called in Perl :-|
So the fix for this was easy.


> ```
> apps/ca.c:47:dbl SPC:#define BASE_SECTION            "ca"
> apps/ca.c:49:dbl SPC:#define ENV_DEFAULT_CA          "default_ca"
> apps/ca.c:77:dbl SPC:#define ENV_DATABASE            "database"
> ```
> 
> I do notice this comment in the script, but am not sure I understand how it applies here:
> 
> ```perl
> # * Use of multiple consecutive spaces is regarded a coding style nit except
> #   when done in order to align certain columns over multiple lines, e.g.:
> #   # define AB  1
> #   # define CDE 22
> #   # define F   3333
> #   This pattern is recognized - and consequently double space not reported -
> #   for a given line if in the line before or after (as far as these exist)
> #   for each occurrence of "  \S" (where \S means non-space) in the given line
> #   there is " \S" in the other line in the respective column position.
> #   This may lead to both false negatives (in case of coincidental " \S")
> #   and false negatives (in case of more complex multi-column alignment).
> ```

The twist here is that so far the matching with the lines before and after did not work as intuitively expected for the above situation with  `ca.c` simply because there are empty lines in between.
I've generalized the handling by skipping over lines that are essentially empty (i.e., contain just whitespace plus possibly a `\` typically used for multi-line macro definitions).


> ```
> apps/ca.c:539:no SPC'*'/:    /*****************************************************************/
> apps/ca.c:539:'/*'no SPC:    /*****************************************************************/
> ```
> 
> Errrrrrrr.... where did this come from? There's _nothing_ saying you _must_ have a space after `/*` or before `*/`

I've relaxed the matching such that also `*` is allowed as in the example above.
Other characters after `/*` and before `*/` would look bad at least to me.

> (as a matter of fact, you will find `/*-` in a number of places, which is an indicator for `indent` to leave that comment alone)

Yes, but this has already been handled.


> However, considering our demands on block comments, I can see where this comes from, but the error message doesn't quite say that, i.e. is quite confusing. And I can recognise the difficulty here, since this is a style of multi-line comment formed with multiple single-line comments.
> 
> ```
> apps/ca.c:753:dbl SPC:    /* Read extensions config file                                   */
> ```

I think in this example repeated spaces in the comment should better not be used.
 
> I think this should be relaxed for comments. We have plenty of places where there are double spaces, for example in the license boilerplate at the top of C source file.

The license texts contain double SPC only after periods/full stop.
The tool already implements an exception that allows `.` to be followed by two (but not more) spaces.

> I have a better example, though:
> 
> ```
> apps/s_client.c:3505:dbl SPC in multi-line comment:     *              extendedResp          ExtendedResponse,
> apps/s_client.c:3507:dbl SPC in multi-line comment:     *         controls       [0] Controls OPTIONAL }
> apps/s_client.c:3516:dbl SPC in multi-line comment:     *              success                      (0),
> apps/s_client.c:3518:dbl SPC in multi-line comment:     *              other                        (80),
> ```
> 
> https://github.com/openssl/openssl/blob/a7bdd0433c0360fccb4dd9f5c43c6fabc4e8c659/apps/s_client.c#L3498-L3523
> 
> There's no reason why this shouldn't be permitted.
> (although, it's curious that we didn't make that block comment start with `/*-`)

Indeed `/*-` should have been used here, and if you do the tool won't complain.
So the tool correctly (though indirectly) brought up an issue here.

> ```
> apps/ca.c:2054:normal stmt indent=16!=12:                buf++;
> apps/ca.c:2055:normal stmt indent=16!=12:                if (*buf)
> apps/ca.c:2056:extra  stmt indent=20!=16:                    type = buf;
> apps/ca.c:2057:normal stmt indent=16!=12:                break;
> apps/ca.c:2058:normal stmt indent=12!=8:            }
> ```
> 
> https://github.com/openssl/openssl/blob/a7bdd0433c0360fccb4dd9f5c43c6fabc4e8c659/apps/ca.c#L2052-L2058

This, as well as a couple of further cases of indenting false positives you reported, was due to a deep bug in handling nested hanging indents at block/statement level. I've meanwhile fixed this (by adding a stack of indents that is governed by `{` and `}`).


> 
> ```
> apps/dhparam.c:330:hanging '(' indent=24!=19:                        "    BIGNUM *p, *g;\n"
> apps/dhparam.c:331:hanging '(' indent=24!=19:                        "\n"
> apps/dhparam.c:332:hanging '(' indent=24!=19:                        "    if (dh == NULL)\n"
> apps/dhparam.c:333:hanging '(' indent=24!=19:                        "        return NULL;\n");
> apps/dhparam.c:339:hanging '(' indent=24!=19:                        "            || !DH_set0_pqg(dh, p, NULL, g)) {\n"
> apps/dhparam.c:340:hanging '(' indent=24!=19:                        "        DH_free(dh);\n"
> apps/dhparam.c:341:hanging '(' indent=24!=19:                        "        BN_free(p);\n"
> apps/dhparam.c:342:hanging '(' indent=24!=19:                        "        BN_free(g);\n"
> apps/dhparam.c:343:hanging '(' indent=24!=19:                        "        return NULL;\n"
> apps/dhparam.c:344:hanging '(' indent=24!=19:                        "    }\n");
> apps/dhparam.c:347:hanging '(' indent=24!=23:                        "    if (!DH_set_length(dh, %ld)) {\n"
> apps/dhparam.c:348:hanging '(' indent=24!=23:                        "        DH_free(dh);\n"
> apps/dhparam.c:349:hanging '(' indent=24!=23:                        "        return NULL;\n"
> apps/dhparam.c:350:hanging '(' indent=24!=23:                        "    }\n", DH_get_length(dh));
> ```
> 
> _This_ one is hard, that should be recognised. Basically, partial strings have been aligned to make the intended output more readable:
> 
> https://github.com/openssl/openssl/blob/a7bdd0433c0360fccb4dd9f5c43c6fabc4e8c659/apps/dhparam.c#L329-L351
> 
> I don't know exactly how you should do this. Is there a way to detect strings in the source along with their position and make an exception of they are aligned with each other?

Here the parser is right reporting wrong indentation of the strings.
Regardless of being strings or anything else, within multi-line expressions the indentation is just after last still open `(`, `{`, or `[` (plus optional special indenting of `:` after `?` conditionals).

The real solution in the above example is simply to to add a newline before the first string literal, such that all arguments of `BIO_printf` are aligned to the right of the open `(`:
```
        BIO_printf(out, 
                   "    DH *dh = DH_new();\n"
                   "    BIGNUM *p, *g;\n"
                   "\n"
                   "    if (dh == NULL)\n"
                   "        return NULL;\n");
```


> ```
> apps/passwd.c:37:'{' @ EOL:static const char ascii_dollar[] = { 0x24, 0x00 };
> ```
> 
> https://github.com/openssl/openssl/blob/a7bdd0433c0360fccb4dd9f5c43c6fabc4e8c659/apps/passwd.c#L37
> 
> Errr.... what?
> 
> ... this is also a call for less cryptic messages. 

Originally I tried to make the error texts as short as possible (since the tool output including the respective source line tends to become rather wide anyway), but I've meanwhile given up on this and tried to make the text much less cryptic - although we're doing applied cryptography, don't we ;-)

> _You_ know what that means, I'm sure, but for anyone else, that was incomprehensible. Is that trying to say that a `{` must be the last character on the line?
> 
> Either way, this line is absolutely permissible as far as I know.

Yes, right, the line is fine, and here the tool was misinterpreting the above line as the header of a function definition, and moreover the complaint was expressed in a misleading way. I've fixed both the analysis bug and the text reporting the situation that was meant here: `'{' not at line start`.
 

> ```
> apps/req.c:35:dbl SPC:#define SECTION         "req"
> apps/req.c:40:dbl SPC:#define DISTINGUISHED_NAME      "distinguished_name"
> ```
> 
> These are a bit interesting. The `SECTION` definition is separated from the rest by an empty line, but its value is aligned with the definitions following. I can see why you'd separate the section name from the rest like that. 

Yes, the first of the two lines is not any more reported after the improvement I mentioned above (ignoring the empty line in between).

> As for `DISTINGUISHED_NAME`, that macro name is so long it pushes the value to the right, making it an exception of alignment. Should all values be pushed to the right because of that? Possibly, but if you look at really long names like we have in `include/openssl/core_names.h`, it can get ridiculous, and having the majority of the values _too_ far away from the macro name might make it less readable (the further apart they are, the harder it is to see that they belong together, and it's even harder on people with astigmatism).

The problem with long names is really notorious. It can become impossible to correctly format within line length things like
```
int fuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuunction(LOOOOOOOOOOOOOOOOONG_TYPE x)
```
The tool already implements an exception for over-long string literals,
and the `--sloppy-len` and `--sloppy-hang` options enable some (limited) relaxations.
 
> I'm not sure how to solve this in a sensible manner... it's quite possible that going the other way and relaxing the double space rules on macro definition is the way to preserve our sanity (i.e. instead of bending over backwards to try and add a pile of alignment rules, just don't check)

At least in the above example from the beginning of `req.c` the solution is easy: push the other string literals to the right.

 
> ```
> apps/req.c:499:no leading '*' in multi-line comment:                BIO_printf(bio_err,"problems opening %s for extra oid's\n",p);
> apps/req.c:499:multi-line comment indent=16!=17:                BIO_printf(bio_err,"problems opening %s for extra oid's\n",p);
> ```
> 
> This is a comment that starts with `/*-`, i.e. a marker we use for `indent` to leave that comment alone. I think you should do the same

While the `-` indicates that the comment should not be re-formatted (which the tool already interprets as not checking SPC within the comment text) I interpret the coding guidelines in a way that the leading `*` (and consequently its proper indentation) in multi-line comments is mandatory. 
 

> ```
> apps/speed.c:1575:normal stmt indent=4!=8 or 12 for '||':    || !defined(OPENSSL_NO_EC)
> ```
> 
> I can't even begin to understand this one.

This was a false positive due to the deep indentation checking bug I mentioned above.
If the above indentation error report/complaint was right, its interpretation would be:

The actual indentation of the code line in question is 4 (), while it should be 8 as for normal statements, while here due to `||` also +4 = 12 would be acceptable.

I've just re-phrased such reports as:
```
apps/speed.c:1575:indent = 4 != 8 for normal stmt or 12 for '||':    || !defined(OPENSSL_NO_EC)
```


> ```
> apps/speed.c:1579:hanging '{' indent=36!=35 or 8 for lines after '{':                                    ECDSA_SECONDS, ECDH_SECONDS,
> apps/speed.c:1580:hanging '{' indent=36!=35 or 8 for lines after '{':                                    EdDSA_SECONDS, SM2_SECONDS };
> ```
> 
> Interesting! So here it seems that values are permitted after a `{` (contrary to the case with the array, see the case in `apps/passwd.c` above). However, if I understand this correctly, you don't allow a space between the `{` and the first value, is that it?

You are right - I thought that the respective section must be formatted as
```
    openssl_speed_sec_t seconds = {SECONDS, RSA_SECONDS, DSA_SECONDS,
                                   ECDSA_SECONDS, ECDH_SECONDS,
                                   EdDSA_SECONDS, SM2_SECONDS };
```
but...

> It _is_ permissible... however, our habit of avoiding that space in apps option arrays is because `indent` would otherwise reformat them in an ugly manner.

... indeed my interpretation was too strict. I've verified this with what the Emacs C mode does.
The tool now allows for extra space in hanging expression indentation after opening `(`, `[`, and `{`.

> That being said, this could actually be reformatted more nicely, like this:
> 
> ```c
>     openssl_speed_sec_t seconds = {
>         SECONDS, RSA_SECONDS, DSA_SECONDS,
>         ECDSA_SECONDS, ECDH_SECONDS,
>         EdDSA_SECONDS, SM2_SECONDS
>     };
> ```

Yes, and the tool already allowed for that.


> ```
> apps/x509.c:788:no SPC'/*':                print_name(out, "/*\n"
> apps/x509.c:788:'/*'no SPC:                print_name(out, "/*\n"
> apps/x509.c:788:text after '/*' in multi-line comment:                print_name(out, "/*\n"
> apps/x509.c:789:no leading '*' in multi-line comment:                                " * Subject: ", X509_get_subject_name(x), get_nameopt());
> apps/x509.c:789:len=90>80:                                " * Subject: ", X509_get_subject_name(x), get_nameopt());
> apps/x509.c:789:multi-line comment indent=32!=34:                                " * Subject: ", X509_get_subject_name(x), get_nameopt());
> apps/x509.c:790:no leading '*' in multi-line comment:                print_name(out, " * Issuer:  ", X509_get_issuer_name(x), get_nameopt());
> apps/x509.c:790:len=89>80:                print_name(out, " * Issuer:  ", X509_get_issuer_name(x), get_nameopt());
> apps/x509.c:790:multi-line comment indent=16!=34:                print_name(out, " * Issuer:  ", X509_get_issuer_name(x), get_nameopt());
> apps/x509.c:791:no leading '*' in multi-line comment:                BIO_puts(out, " */\n");
> apps/x509.c:791:text before '*/' in multi-line comment:                BIO_puts(out, " */\n");
> apps/x509.c:791:multi-line comment indent=16!=34:                BIO_puts(out, " */\n");
> ```
> 
> https://github.com/openssl/openssl/blob/a7bdd0433c0360fccb4dd9f5c43c6fabc4e8c659/apps/x509.c#L788-L791
> 
> Maybe trying to check comments in strings is over-doing it a bit? wink

Absolutely. I had already placed TODOs in the tool code that comment delimiters must not be matched within string literals, and luckily it turned out this was pretty easy to achieve (since it already internally blinds the contents of string and char literals, which I simply moved before the analysis of comments).

DDvO:
Finally the tool code contains all features and bug fixes I aimed to include, 
so I consider it ready for review.

levitte:
Small itty gritty ones:

```
apps/asn1pars.c:149:no SPC after ';':                for (tmp = 0;; tmp++) {
```

I think a series of semicolons with no spaces in between should be admissible.

richsalz:
> I think a series of semicolons with no spaces in between should be admissible.

FWIW I disagree.  The loop reads "better" with a space in the condition-part.  And it's more likely to be a source of errors (macro expansion etc) of having two ; on a line.

levitte:
> And it's more likely to be a source of errors (macro expansion etc) of having two ; on a line.

As far as I know, this checker doesn't expand macros, so that's kinda lost...

DDvO:
@levitte is right that the checker tool does not expand macros, and this is on purpose: it is not meant to perform semantic checks but to check the _appearance_ of source files, which is _before_ macro expansion.

For the check that a `;` should be followed by whitespace I've already implemented the exception that the idiomatic `for(;;)` is acceptable. For mixed cases like `for (tmp = 0;; tmp++)` as given above I tend to agree with @richsalz that one should add a space after the first `;` just for better readability.

levitte:
> For mixed cases like `for (tmp = 0;; tmp++)` as given above I tend to agree with @richsalz that one should add a space after the first `;` just for better readability.

Ok, sure, I can work with that.

DDvO:
Over the last couple of weeks I've been doing some further improvements on the tool,
mainly to strengthen checks on comment indentation and missing/extra whitespace and
to reduce false positives in particular when lines contain more than a single statements.

@levitte, all false positives on `crypto/store/*`  are meanwhile eliminated.

DDvO:
How to proceed with this PR?
Should it receive two formal reviews, like normal OpenSSL library code?

DDvO:
I've just addded a further special check, namely for missing brace before or after 'else',
due to a reviewer's finding in our CMP contribution: https://github.com/openssl/openssl/pull/10620#discussion_r366428585.
The tool now finds 50 such cases in `apps/*.c`.

levitte:
> The tool now finds 50 such cases in `apps/*.c`.

Do those cases have preprocessor stuff around them?  That's often the case...

DDvO:
@levitte, no, these findings are on code outside preprocessor directives / macro definitions.
You can see yourself using 
```
util/check-format.pl apps/*.c | grep "same line" | grep "else"
```
which yields
```
apps/crl.c:287:no '{' on same line after '} else':                } else
apps/dgst.c:353:no '}' on same line before 'else ... {':    else {
apps/dhparam.c:211:no '{' on same line after '} else':        } else
apps/dhparam.c:254:no '{' on same line after '} else':        } else
apps/enc.c:655:no '{' on same line after '} else':    } else
apps/ocsp.c:809:no '}' on same line before 'else ... {':            else {
apps/ocsp.c:1377:no '}' on same line before 'else ... {':        else if (isxdigit(_UC(p[1])) && isxdigit(_UC(p[2]))) {
apps/ocsp.c:1383:'else' not on same line as preceding '}':        else
apps/pkcs12.c:740:no '{' on same line after '} else':        } else if (options & CLCERTS)
apps/pkcs8.c:278:no '}' on same line before 'else ... {':            else {
apps/req.c:986:no '{' on same line after '} else':        } else
apps/s_client.c:1682:no '{' on same line after '} else':    } else
apps/s_client.c:1752:no '{' on same line after '} else':        } else if (bio_c_out == NULL)
apps/s_client.c:2168:no '{' on same line after '} else':    } else
apps/s_client.c:2936:no '{' on same line after '} else':                } else
apps/s_client.c:3132:no '{' on same line after '} else':            } else
apps/s_client.c:3394:no '{' on same line after '} else':        } else
apps/smime.c:579:no '}' on same line before 'else ... {':        else {
apps/speed.c:2225:no '}' on same line before 'else ... {':        else {
apps/speed.c:2242:no '}' on same line before 'else ... {':        else {
apps/speed.c:2259:no '}' on same line before 'else ... {':        else {
apps/speed.c:2274:no '}' on same line before 'else ... {':        else {
apps/speed.c:2288:no '}' on same line before 'else ... {':        else {
apps/speed.c:2302:no '}' on same line before 'else ... {':        else {
apps/speed.c:2314:no '}' on same line before 'else ... {':        else {
apps/speed.c:2325:no '}' on same line before 'else ... {':        else {
apps/speed.c:2338:no '}' on same line before 'else ... {':        else {
apps/speed.c:2349:no '}' on same line before 'else ... {':        else {
apps/speed.c:3677:no '}' on same line before 'else ... {':        else {
apps/speed.c:4034:no '}' on same line before 'else ... {':            else if (strncmp(buf, "+F3:", 4) == 0) {
apps/speed.c:4050:no '}' on same line before 'else ... {':            else if (strncmp(buf, "+F4:", 4) == 0) {
apps/speed.c:4088:no '}' on same line before 'else ... {':            else if (strncmp(buf, "+F7:", 4) == 0) {
apps/speed.c:4105:no '}' on same line before 'else ... {':            else if (strncmp(buf, "+H:", 3) == 0) {
apps/speed.c:4107:no '{' on same line after '} else':            } else
apps/s_server.c:433:no '}' on same line before 'else ... {':        else if (buf[i] == '\n') {
apps/s_server.c:2167:no '{' on same line after '} else':    } else
apps/s_server.c:2364:no '{' on same line after '} else':        } else
apps/s_server.c:2373:no '{' on same line after '} else':    } else
apps/s_server.c:3355:no '}' on same line before 'else ... {':                        else {
apps/s_server.c:3374:no '{' on same line after '} else':        } else
apps/s_time.c:417:no '}' on same line before 'else ... {':    else {
apps/ts.c:365:no '}' on same line before 'else ... {':            else {
apps/ts.c:369:no '{' on same line after '} else':        } else
apps/ts.c:867:no '}' on same line before 'else ... {':    else {
apps/x509.c:725:no '}' on same line before 'else ... {':            else if (subject_hash_old == i) {
apps/x509.c:729:no '}' on same line before 'else ... {':            else if (issuer_hash == i) {
apps/x509.c:733:no '}' on same line before 'else ... {':            else if (issuer_hash_old == i) {
apps/x509.c:737:no '}' on same line before 'else ... {':            else if (pprint == i) {
apps/x509.c:760:no '{' on same line after '} else':                } else
apps/x509.c:767:no '{' on same line after '} else':                } else
apps/x509.c:837:no '}' on same line before 'else ... {':            else if (sign_flag == i && x509req == 0) {
```

DDvO:
@levitte, should I add a report for one-letter names other than `i` and `j` to the tool?

DDvO:
BTW, this PR is still waiting for a formal review.

levitte:
I only looked at this one to understand what you were on to in https://github.com/openssl/openssl/pull/10363#issuecomment-574349219

```
apps/crl.c:287:no '{' on same line after '} else':                } else
```

levitte:
> @levitte, should I add a report for one-letter names other than `i` and `j` to the tool?

No, I think that's a bit over the top.  There's more over the place, such as the local pointers `p` and `q`, and there are places where you see `a` and `b` (typical for cmp functions), or `k` for a third index, etc.  we will drive ourselves crazy...  *some* things can be left to the humans to judge.

DDvO:
We could automatically check for names that definitely should not be used, as mentioned by @bernd-edlinger: https://github.com/openssl/openssl/issues/10725#issuecomment-578910153, namely 
`l` (lower case el) and `I` (upper case i) and `O` (upper case o).

levitte:
That, I think is a better idea.  Advice against specifics instead of disallowing all but just a few.  Yes.

DDvO:
Done. The tool now reports 5 one-letter variables '`l`' in `apps/*.c`

kaduk:
> There's more over the place, such as the local pointers `p` and `q`, and there are places where you see `a` and `b` (typical for cmp functions), or `k` for a third index, etc. 

Don't forget `s` for SSL objects!

DDvO:
> Don't forget `s` for SSL objects!

There is also `x` for X509 objects, sigh.

levitte:
See "etc" :wink:

DDvO:
I don't like those one-char variable names at all: 
searching for all occurrences of such a variable becomes a nightmare of false positives.

richsalz:
So maybe don't worry about one-character variables for now.  The phrase "diminishing returns" comes to mind.

levitte:
> I don't like those one-char variable names at all:

Noted. `foo`, `bar`, `baz` it is, then :wink:

kaduk:
> > I don't like those one-char variable names at all:
> 
> Noted. `foo`, `bar`, `baz` it is, then wink

Or `qux`, `quux`, `quuux`.  "Now which was `quux` and which one was `quuux` again?"

DDvO:
Sorry to disappoint your creativity :wink:
 We already have on the list in https://github.com/openssl/openssl/issues/10725:
> Use brief but descriptive names for variables, types, functions, etc.

(Yet surely this is something a tool cannot check.)

DDvO:
Pushed (using `ghmerge --nobuild --squash 10363 levitte @DDvO`).

Thanks for all feedback by @levitte and many others.

PR #10364, opened 2019-11-05
Change the logic and behaviour surrounding '--api' and 'no-deprecated'

levitte:
At some point in time, there was a 'no-deprecated' configuration
option, which had the effect of hiding all declarations of deprecated
stuff, i.e. make the public API look like they were all removed.

At some point in time, there was a '--api' configuration option, which
had the effect of having the public API look like it did in the version
given as value, on a best effort basis.  In practice, this was used to
get different implementations of BN_zero(), depending on the desired
API compatibility level.

At some later point in time, '--api' was changed to mean the same as
'no-deprecated', but only for the deprecations up to and including the
desired API compatibility level.  BN_zero() has been set to the
pre-1.0.0 implementation ever since, unless 'no-deprecation' has been
given.

This change turns these options back to their original meaning, but
with the slight twist that when combined, i.e. both '--api' and
'no-deprecated' is given, the declarations that are marked deprecated
up to an including the desired API compatibility level are hidden,
simulating that they have been removed.

If no desired API compatibility level has been given, then
configuration sets the current OpenSSL version by default.

Furthermore, the macro OPENSSL_API_LEVEL is now used exclusively to
check what API compatibility level is desired.  For checking in code
if `no-deprecated` has been configured for the desired API
compatibility level, macros for each supported level is generated,
such as OPENSSL_NO_DEPRECATED_1_1_1, corresponding to the use of
DEPRECATEDIN_ macros, such as DEPRECATEDIN_1_1_1().

Just like before, to set an API compatibility level when building an
application, define OPENSSL_API_COMPAT with an appropriate value.  If
it's desirable to hide deprecated functions up to and including that
level, additionally define OPENSSL_NO_DEPRECATED (the value is
ignored).

##### Checklist


richsalz:
Is this appropriate for 3.0?  Seems like a breaking config change.

levitte:
Kinda sorta, but these are both pretty specialised config options. INSTALL says so, even:

>  --api=x.y.z
>                   Don't build with support for deprecated APIs below the
>                   specified version number. For example "--api=1.1.0" will
>                   remove support for all APIS that were deprecated in OpenSSL
>                   version 1.1.0 or below. --->> This is a rather specialized option
>                   for developers.<<--- If you just intend to remove all deprecated
>                   APIs entirely (up to the current version), it is easier
>                   to add the 'no-deprecated' option instead (see below).

apps/speed.c
```diff
@@ -169,11 +169,6 @@ static int DES_ede3_cbc_encrypt_loop(void *args);
 static int AES_cbc_128_encrypt_loop(void *args);
 static int AES_cbc_192_encrypt_loop(void *args);
 static int AES_cbc_256_encrypt_loop(void *args);
-#if !OPENSSL_API_3
```
t-j-h: I don't get why these changes are included in this PR - they are actually unrelated to the adjustment of the option handling IMHO.

levitte: It starts with this change in `include/openssl/aes.h`:

``` diff
@@ -73,17 +73,18 @@ void AES_cfb8_encrypt(const unsigned char *in, unsigned char *out,
 void AES_ofb128_encrypt(const unsigned char *in, unsigned char *out,
                         size_t length, const AES_KEY *key,
                         unsigned char *ivec, int *num);
-# if !OPENSSL_API_3
 /* NB: the IV is _two_ blocks long */
-void AES_ige_encrypt(const unsigned char *in, unsigned char *out,
-                     size_t length, const AES_KEY *key,
-                     unsigned char *ivec, const int enc);
+DEPRECATEDIN_3_0(void AES_ige_encrypt(const unsigned char *in,
+                                      unsigned char *out,
+                                      size_t length, const AES_KEY *key,
+                                      unsigned char *ivec, const int enc))
 /* NB: the IV is _four_ blocks long */
-void AES_bi_ige_encrypt(const unsigned char *in, unsigned char *out,
-                        size_t length, const AES_KEY *key,
-                        const AES_KEY *key2, const unsigned char *ivec,
-                        const int enc);
-# endif
+DEPRECATEDIN_3_0(void AES_bi_ige_encrypt(const unsigned char *in,
+                                         unsigned char *out,
+                                         size_t length, const AES_KEY *key,
+                                         const AES_KEY *key2,
+                                         const unsigned char *ivec,
+                                         const int enc))
 
 int AES_wrap_key(AES_KEY *key, const unsigned char *iv,
                  unsigned char *out,
```

This turn a "simulated deprecation" to an actual one (i.e. we not have a proper `deprecated` attribute attached to those functions with gcc and clang).

The consequence, though, is that any `--strict-warning` will make the build fail because the warning for using deprecated functions is turned into an error.  That's why I need to make that removal.  This is, btw, our M.O. so far, to remove all uses of deprecated functions, but leave the implementation around in case someone still wants to use them.

The alternative is to keep this as a "simulated deprecation" in this PR and handle the "real" deprecation of AES_ige functions in a separate PR.  I can certainly do that if that helps reviewing this PR.

levitte: So...  was that explanation enough justification, or would you prefer I make a separate PR for this part?

t-j-h: I'd be happier with them in a separate PR as they basically don't match the PR description (which was nicely worded and easy to follow). They just popped out as a "surprise" ...

levitte: Okie.  Will do.

levitte: Done.  I hope I didn't miss anything.

levitte:
Found a logic error.  `--api=1.0.2 no-deprecated` errors all over the place.

t-j-h (approved):

t-j-h:
Thanks - looks clean now. 
Approved along with #10367

levitte:
Logic error fixed.  @t-j-h, you were a little too fast there, please review the last fixup commit

t-j-h (approved):

levitte:
Merged.

a6a4d0acd23b3fc85041d9096b67bcf18ccb635c Change the logic and behaviour surrounding '--api' and 'no-deprecated'
936c2b9e933eacae80d3489e5f7316589f3e9a07 Update source files for deprecation at 3.0
00db8c60aaecd2ff9b0a5d9fd10c7771a1e1a73e Update source files for pre-3.0 deprecation
9fd9951e4b49051c8f69157b478070951c43c5c4 include/openssl/opensslconf.h.in: remove spurious HEADER_FILE_H definition
285a0db396ff3a7801c199b6d812eae298178249 Update some inclusions of <openssl/macros.h>
a3844061df7d0e588d018370efd1d309f856e6f3 Update the check surround the BN_zero() implementation
686923b19824546f5c76a012e41e1fc7141ae712 include/openssl/asn1.h: Remove version dependent inclusion of <openssl/bn.h>
54a0d4ceb28d53f5b00a27fc5ca8ff8f0ddf9036 Update INSTALL to document the changed '--api' and 'no-deprecated'

PR #10367, opened 2019-11-06
[Pending on #10538] Deprecate AES_ige functions for real

levitte:
AES_ige_encrypt() and AES_bi_ige_encrypt() were deprecated in a
non-standard manner, now we do it for real.

### DIFF
--- a/include/openssl/aes.h
+++ b/include/openssl/aes.h
@@ -73,17 +73,18 @@ void AES_cfb8_encrypt(const unsigned char *in, unsigned char *out,
 void AES_ofb128_encrypt(const unsigned char *in, unsigned char *out,
                         size_t length, const AES_KEY *key,
                         unsigned char *ivec, int *num);
-# ifndef OPENSSL_NO_DEPRECATED_3_0
 /* NB: the IV is _two_ blocks long */
-void AES_ige_encrypt(const unsigned char *in, unsigned char *out,
-                     size_t length, const AES_KEY *key,
-                     unsigned char *ivec, const int enc);
+DEPRECATEDIN_3_0(void AES_ige_encrypt(const unsigned char *in,
+                                      unsigned char *out, size_t length,
+                                      const AES_KEY *key, unsigned char *ivec,
+                                      const int enc))
 /* NB: the IV is _four_ blocks long */
-void AES_bi_ige_encrypt(const unsigned char *in, unsigned char *out,
-                        size_t length, const AES_KEY *key,
-                        const AES_KEY *key2, const unsigned char *ivec,
-                        const int enc);
-# endif
+DEPRECATEDIN_3_0(void AES_bi_ige_encrypt(const unsigned char *in,
+                                         unsigned char *out, size_t length,
+                                         const AES_KEY *key,
+                                         const AES_KEY *key2,
+                                         const unsigned char *ivec,
+                                         const int enc))
 
 int AES_wrap_key(AES_KEY *key, const unsigned char *iv,
                  unsigned char *out,


levitte:
Note: if #10364 is merged first, this PR will need a slight adjustment

levitte:
Travis failure not relevant

slontis:
Excellent - I did wonder about this when moving ciphers into providers..

levitte:
As expected, this needed some adjustment after rebase on top of a merged #10364.  It was substantial enough that I prefer a re-review, at least one.

t-j-h:
This doesn't make sense to me. 
Marked for deprecation doesn't mean remove the code IMHO.
Removal happens at actual removal - i.e. 4.0 time frame.

levitte:
Interestingly enough, you approved this change yesterday.

However, this doesn't remove the central itself.  `crypto/aes/aes_ige.c` remains intact and the `AES_ige` can be called by anyone who wants (unless they have removed them prematurely with `no-deprecated`).

*However*, using them produces a warning, and since *we* make `-Werror` builds, *our application* cannot use them because the builds break.  In other worlds, our own application lives on the bleeding edge, and this is the consequence.

The implication that I have hereby removed the function itself is a red herring.

levitte:
Another option could be to have a more lax attitude when building the application, i.e. let warnings just be warnings there...  or to turn off deprecation warnings.  We have the possibility to set flags specifically for programs, so that wouldn't be very hard to do.

richsalz:
If you want to keep around a test suite, for example, that seems the only place to leave warnings as warnings.  Do we have any for AES-ige?

t-j-h:
Taking out test code for functions which remain in the library seems to be the *wrong* approach in all contexts.

mattcaswell:
> Taking out test code for functions which remain in the library seems to be the wrong approach in all contexts.

That is actually a good point. Relaxing "-Werror" for that particular file seems like a possible approach (is there an option to suppress deprecation warnings that we can add on just that file?)

t-j-h:
A -werror build should not fail for a depreciation warning for *our* usage. That is the cause of the problem as your have made deprecation mean must-not-be-used and that isn't reality.

levitte:
Sure.  @mattcaswell, we don't have the fine grained mechanism to select flags per application, yet.  It could be done with some use of `build.info` attributes, but that seems a bit far fetched.  The easy option we have is to turn off the warning entirely for all programs.  The question is, do we want to hide the warning entirely, or just make it so it doesn't convert to an error?

levitte:
Do not, @t-j-h, that we have removed all calls on deprecated functions in all our applications.  It usually goes unnoticed because we're basically replacing old calls with new ones, which flies under the radar.  In this case, we're deprecating with no specific replacement (because security issue, the message on the wall is "don't use this, use some other AES variant"), so this is a first that actually gets noticed.

There are already a number of other `DEPRECATEDIN_3_0` functions, which we don't call from anywhere any more.

t-j-h:
If we remove test code then it is a mistake in all contexts to do so - whether we call it outside the test code or not - basically if a test exists and the function remains then the test should always remain.

It certainly shows up a lot more clearly there than elsewhere - but the same principles apply.

Deprecated functions are still supported except for special builds that ask for them to not be. So their corresponding tests must remain.

I also wouldn't remove it from speed either - as it is still supported. That is also a mistake - but a different one - where an app feature is removed when it should not be - that too should remain.

t-j-h:
And we do need to write this down so we have a clear policy on it - I think we seem to be in agreement conceptually even if in the past we may have dropped some stuff we shouldn't have.

richsalz:
AES_ige is not secure; see https://news.ycombinator.com/item?id=6915741 for example.
I think, in this case, deprecating it and not including the test makes sense.

levitte:
Ok, so it sounds like trying to remove the deprecation warning, or at least not making it an error is the way forward.  I have nothing against doing so.

t-j-h:
@richsalz it is in the library so it is supported and should be tested. Users can 
use it. Deprecation is not removal.

levitte:
I've dismissed all approvals and set this back to WIP.

richsalz:
Okay, @t-j-h you convinced me.

levitte:
I've a question...  with a definition of `--strict-warnings` where programs get an additional flag making sure the deprecation warning doesn't become an error, do we still want to see the warning (`-Wno-error=deprecated-declarations`), or do we want to turn a blind eye entirely (`-Wno-deprecated-declarations`)?

Personally, I prefer to be informed, but I know others who'd rather have it as silent as possible as long as it's not an error.

mattcaswell:
Hmm, good question. I think I tend towards wanting to see the warning.

levitte:
This depends on #10538 to not generate errors just because we're still using and testing it.

levitte:
I have removed all the removals from this PR, only the real deprecation remains.

mattcaswell:
Alternative approach shown in #10540.

levitte:
#10558 showed the path forward.  Closing this

PR #10368, opened 2019-11-06
Reinstate the KDF error macros

### LINKED ISSUE #10340: kdferr.h got removed from installed headers
Comparing the list of installed header files from 1.1.1 to current master, I notice at least kdferr.h got removed.

We also have the following new public files, not sure if they're all supposed to be public:
cmperr.h
cmp.h
cmp_util.h
core.h
core_names.h
core_numbers.h
crmferr.h
crmf.h
esserr.h
ess.h
fips_names.h
macros.h
params.h
provider.h
trace.h
types.h

levitte:
include/openssl/kdferr.h was removed because it got empty.  However,
that will break the build of any application that includes it
directly, so we add it back as a placeholder.

Fixes #10340


t-j-h (dismissed):

kroeckx:
I expect all the old defines to be in there. That is, the KDF_F_* defined to 0, the KDF_R_* to some distinct value >= 100. There also was an ERR_load_KDF_strings().

kroeckx (changes requested):
All old defines/functions should be in it.

levitte:
Ah, so what you're asking for is a revert of fe6ec26b204a056aee2a24b79df09a45b2308603, basically

kroeckx (approved):

kroeckx:
# Number of lines in @nm_lines before massaging: 4935
# Number of lines in @def_lines before massaging: 4795
# Number of lines in @nm_lines after massaging: 4790
# Number of lines in @def_lines after massaging: 4791
# The following symbols are missing in libcrypto.so:
#   ERR_load_KDF_strings

include/openssl/kdferr.h
```diff
@@ -0,0 +1,110 @@
...
+# define HEADER_KDFERR_H
+
+# include <openssl/opensslconf.h>
+# include <openssl/symhacks.h>
+
+
+# ifdef  __cplusplus
+extern "C"
+# endif
+int ERR_load_KDF_strings(void);
```
kroeckx: I guess you want to mark this as deprecated

kroeckx (changes requested):

kroeckx:
and run make update

levitte:
Ah, yeah, forgot that.

include/openssl/kdferr.h
levitte: Good idea

levitte:
I reverted a lot more than might be practically needed...  I could see reason to empty `crypto/kdf/kdf_err.c` of almost all content, just leaving `ERR_load_KDF_strings()` as a function that does nothing.

Still, it was an interesting exercise, to invent some kind of "conservation state" for this...  we may need to do the same with others, eventually.

kroeckx:
I don't see a reason to keep the strings in kdf_err.c

kroeckx:
```
crypto: function ERR_load_KDF_strings undocumented
# 1 in util/libcrypto.num are not documented
```

levitte:
Eh gods. Could we perhaps make a documentation exception on deprecated functions?
Meanwhile, I'll add it to the appropriate missing file.

levitte:
All done

crypto/err/err_all.c
```diff
@@ -38,6 +38,7 @@
 #include <openssl/cmperr.h>
 #include <openssl/cterr.h>
 #include <openssl/asyncerr.h>
+#include <openssl/kdferr.h>
```
kroeckx: Why this include?

include/openssl/kdferr.h
```diff
@@ -0,0 +1,116 @@
...
+#  define KDF_F_X942KDF_HASH_KDM                           0
+#  define KDF_F_X942KDF_NEW                                0
+#  define KDF_F_X942KDF_SIZE                               0
+#  define KDF_F_X963KDF_DERIVE                             0
+# endif
+
+/*
+ * KDF reason codes.
+ */
+# define KDF_R_BAD_ENCODING                               122
```
kroeckx: Mark as deprecated too?

crypto/err/err_all.c
levitte: Because I actually did a revert and didn't notice this :wink:

include/openssl/kdferr.h
levitte: Good idea

crypto/err/err_all.c
levitte: Fixed

include/openssl/kdferr.h
levitte: Done

kroeckx:
I think it looks good, but I'm not going to review (the perl part).

t8m (approved):
Looks good, including the perl.

levitte:
Merged.

7c6a0d909abda7c854b44831ef43ebdda49ed42e Reinstate the KDF error macros
dc5ce51983713dfbb8f3d0c8b198dee32b563ac6 util/perl/OpenSSL/ParseC.pm: Fix 'extern "C" DEPRECATEDIN_x_y(...)'
905b097fd586ed0279aee05b5961224212014ec9 Deprecate ERR_load_KDF_strings()
4d301427a96010468da2bb67bf1025fa8d886ab9 Make sure KDF reason codes are conserved in their current state

Issue #10313, opened 2019-10-31
The numbits argument is not documented in openssl genrsa -help


MaharishiCanada:
Using OpenSSL 1.1.1c  28 May 2019 in Ubuntu 19.10, I noticed that  the output of `openssl genrsa -help` does not document the numbits argument, which is documented  here https://www.openssl.org/docs/man1.0.2/man1/genrsa.html . This is perhaps normal.

richsalz:
You are pointing to the 1.0.2 version.  But all three versions do document it; it's right after the -engine documentation. :)

mattcaswell:
No, the reporter seems to be correct. I see no mention of it in the "-help" output:

````
$ openssl genrsa -help
Usage: genrsa [options]
Valid options are:
 -help               Display this summary
 -3                  Use 3 for the E value
 -F4                 Use F4 (0x10001) for the E value
 -f4                 Use F4 (0x10001) for the E value
 -out outfile        Output the key to specified file
 -rand val           Load the file(s) into the random number generator
 -writerand outfile  Write random data to the specified file
 -passout val        Output file pass phrase source
 -*                  Encrypt the output with any supported cipher
 -engine val         Use engine, possibly a hardware device
 -primes +int        Specify number of primes
 -verbose            Verbose output
````

**Comment by MaharishiCanada** [comment]
I was not clear apparently.  I was pointing to the output  of `openssl genrsa -help` in OpenSSL 1.1.1c 28 May 2019 in Ubuntu 19.10. What was confusing is that I also pointed to the online documentation, but it was only to say that IT IS documented elsewhere.  Again, maybe it is normal, something well known that nobody cares about and I am just making  a full of myself by showing that I just realized that now...

**Comment by richsalz** [comment]
Thanks for the clarification.  You are not making a fool of yourself!  I made one of myself by not reading carefully.

In general, the -help option only documents the options, but it's not consistent.  Using the #9953 stuff, we could add a "Parameters" section to those commands that take parameters and document them.

**Comment by mattcaswell** [comment]
> In general, the -help option only documents the options, but it's not consistent.

I really think it should list all arguments - not just the option flags.

**Comment by richsalz** [comment]
> I really think it should list all arguments - not just the option flags.

I'd be willing to do that. I had already committed to documenting all the missing flags.  But I need all of my other doc PR's to get reviewed first.

**Comment by richsalz** [comment]
I have the infrastructure to do this, here's sample output:
```
; ./apps/openssl genrsa --help
[...]

PR #10371, opened 2019-11-06
Help cmd params

richsalz:
This adds documentation for all commands that take parameters.

It makes the help strings consistent.

It is based on #9953 so only look at the top commit.  Or, better yet, go review and approve that one first.

### DIFF
--- a/apps/cms.c
+++ b/apps/cms.c
@@ -88,9 +88,7 @@ typedef enum OPTION_choice {
 } OPTION_CHOICE;
 
 const OPTIONS cms_options[] = {
-    {OPT_HELP_STR, 1, '-', "Usage: %s [options] cert.pem...\n"},
-    {OPT_HELP_STR, 1, '-',
-        "  cert.pem... recipient certs for encryption\n"},
+    {OPT_HELP_STR, 1, '-', "Usage: %s [options] [cert...]\n"},
 
     OPT_SECTION("General"),
     {"help", OPT_HELP, '-', "Display this summary"},
@@ -219,6 +217,9 @@ const OPTIONS cms_options[] = {
 
     OPT_R_OPTIONS,
     OPT_V_OPTIONS,
+
+    OPT_PARAMETERS(),
+    {"cert", 0, 0, "Recipient certs, used when encrypting"},
     {NULL}
 };
 
--- a/apps/cms.c
+++ b/apps/cms.c
@@ -219,7 +219,7 @@ const OPTIONS cms_options[] = {
     OPT_V_OPTIONS,
 
     OPT_PARAMETERS(),
-    {"cert", 0, 0, "Recipient certs, used when encrypting"},
+    {"cert", 0, 0, "Recipient certs (optional; used only when encrypting)"},
     {NULL}
 };
 
--- a/apps/speed.c
+++ b/apps/speed.c
@@ -315,7 +315,7 @@ typedef enum OPTION_choice {
 } OPTION_CHOICE;
 
 const OPTIONS speed_options[] = {
-    {OPT_HELP_STR, 1, '-', "Usage: %s [options] ciphers...\n"},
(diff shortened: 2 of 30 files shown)


richsalz:
Now that #9953 was merged, I rebased this.  Also fixed the mac and kdf commands to properly use ```OPT_MORE_STRING```

richsalz:
ping @mattcaswell who commented over at #10313

apps/cms.c
```diff
@@ -88,9 +88,7 @@ typedef enum OPTION_choice {
 } OPTION_CHOICE;
 
 const OPTIONS cms_options[] = {
-    {OPT_HELP_STR, 1, '-', "Usage: %s [options] cert.pem...\n"},
-    {OPT_HELP_STR, 1, '-',
-        "  cert.pem... recipient certs for encryption\n"},
+    {OPT_HELP_STR, 1, '-', "Usage: %s [options] [cert...]\n"},
```
t8m: In general to distinguish between literal and replaceable parameter names (as italics cannot be used in usage outputs) the commands use capitalized words for the replaceable texts. But that would be probably a separate PR to do and maybe even not worth it.

apps/cms.c
```diff
@@ -219,6 +217,9 @@ const OPTIONS cms_options[] = {
 
     OPT_R_OPTIONS,
     OPT_V_OPTIONS,
+
+    OPT_PARAMETERS(),
+    {"cert", 0, 0, "Recipient certs, used when encrypting"},
```
t8m: perhaps `Recipient certs (optional, required and used only when encrypting)` ?

apps/verify.c
```diff
@@ -72,6 +72,9 @@ const OPTIONS verify_options[] = {
     {"sm2-hex-id", OPT_SM2HEXID, 's',
      "Specify a hex ID string to verify an SM2 certificate"},
 #endif
+
+    OPT_PARAMETERS(),
+    {"cert", 0, 0, "Certificate(s) to verify (optional)"},
```
t8m: Maybe `(optional; stdin used otherwise)` ?

apps/cms.c
richsalz: I'd prefer someone do that as a separate PR if it seems worthwhile.

apps/cms.c
richsalz: I did "Recipient certs (optional; required only when encrypting)"

richsalz: That wording isn't great either, since you can use -recip.  Sigh. So I made it "used only"

apps/verify.c
richsalz: done.

richsalz:
fixup commit pushed.

apps/cms.c
t8m: That's definitely OK.

apps/speed.c
```diff
@@ -356,6 +356,9 @@ const OPTIONS speed_options[] = {
      "Use specified offset to mis-align buffers"},
 
     OPT_R_OPTIONS,
+
+    OPT_PARAMETERS(),
+    {"algorithm", 0, 0, "Algorithm(s) to test"},
```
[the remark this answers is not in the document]
t8m: @richsalz What about this one?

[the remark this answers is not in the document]
richsalz: The behavior if not specified is wonky. It depends on what flags you use. I tried to verify the behavior and got this:
```
; ./apps/openssl speed
Doing md2 for 3s on 16 size blocks: EVP error!
        exit 1
```

But I'll add optional.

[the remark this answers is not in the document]
t8m: That would be a bug in speed.c. It works fine in 1.1.1.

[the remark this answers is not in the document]
richsalz: the bug is more fundamental, but I opened #10435 to catch it.  resolving this suyb-thread.

t8m:
Merged to master as 92de469fbd6

PR #10372, opened 2019-11-06
Support ciphersuites using a SHA2 384 digest in FreeBSD KTLS.

bsdjhb:

### DIFF
--- a/ssl/t1_enc.c
+++ b/ssl/t1_enc.c
@@ -410,6 +410,9 @@ int tls1_change_cipher_state(SSL *s, int which)
         case SSL_SHA256:
             crypto_info.auth_algorithm = CRYPTO_SHA2_256_HMAC;
             break;
+        case SSL_SHA384:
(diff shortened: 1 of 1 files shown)


kroeckx:
Do all FreeBSD versions that support KTLS also support SHA384? Does the code detect that the kernel doesn't support it, and not do KTLS in that case?

bsdjhb:
The support varies by the KTLS backends, but the socket option ktls_start() calls will fail and fall back to non-KTLS if it can't find a suitable backend.  It turns out that the initial commit for KTLS for FreeBSD did include SHA 384 support (I thought I hadn't yet included it in the "generic" layer), but the KTLS code in FreeBSD is also careful to fail any cipher suites or versions it doesn't recognize to permit adding support for more suites or versions in the future.

In the case of the constants (CRYPTO_*) used in the code, those are from FreeBSD's OCF crypto layer and have been around much longer than KTLS in FreeBSD, so this should safely compile on any FreeBSD system with KTLS support.

PR #10374, opened 2019-11-07
Add FIPS self test (KATS)

slontis:
Added an API to optionally set a self test callback.
The callback has the following 2 purposes
(1) Output information about the KAT tests.
(2) Allow the ability to corrupt one of the KAT's
The fipsinstall program uses the API.

Some KATS are not included in this PR since the required functionality does not yet exist in the provider.

##### Checklist


slontis:
Documentation is still to be done.
Feedback on the callback mechanism and the way it is passed from the core to the fips provider would be welcome.

apps/fipsinstall.c
```diff
@@ -418,3 +443,53 @@ int fipsinstall_main(int argc, char **argv)
...
 }
+
+static int self_test_events(OSSL_PARAM *params)
+{
+    OSSL_PARAM *p = NULL;
+    const char *phase = NULL, *type= NULL, *desc = NULL;
+    int ret = 0;
+
+    p = OSSL_PARAM_locate(params, OSSL_PROV_PARAM_SELF_TEST_PHASE);
+    if (p == NULL || p->data_type != OSSL_PARAM_UTF8_STRING)
```
paulidale: It might not be relevant here but calling `OSSL_PARAM_get_utf8_string` might be a better option.  It mallocs the string storage space if the passed pointer is null.  This avoids string life times and size concerns and does some of the checking for you.

It doesn't mean deallocation at the end though.  The _PTR types are designed for long lived things, the base string types are not.

crypto/provider_core.c
```diff
@@ -873,8 +876,8 @@ static const OSSL_DISPATCH core_dispatch_[] = {
     { OSSL_FUNC_BIO_NEW_MEMBUF, (void (*)(void))BIO_new_mem_buf },
     { OSSL_FUNC_BIO_READ_EX, (void (*)(void))BIO_read_ex },
     { OSSL_FUNC_BIO_FREE, (void (*)(void))BIO_free },
+    { OSSL_FUNC_SELF_TEST_CB, (void (*)(void))OSSL_SELF_TEST_get_callback },
```
paulidale: This seems good enough until we figure out something better.

apps/fipsinstall.c
slontis: The strings are all static so it should be ok how it is..

levitte:
In #10289, I have a callback mechanism that is quite similar to yours, except it uses a common structure instead a direct function pointer, to get some kind of closure mechanism, and also to make it possible to "translate" to whatever kind of function on libcrypto side.
I wonder, should we keep the fips self test cb separate, or should we eventually merge them?

apps/fipsinstall.c
```diff
@@ -285,6 +296,15 @@ int fipsinstall_main(int argc, char **argv)
         case OPT_OUT:
             out_fname = opt_arg();
             break;
+        case OPT_SELF_TEST_NO_LOG:
+            self_test_log = 0;
+            break;
+        case OPT_SELF_TEST_CORRUPT_DESC:
```
richsalz: can you have both corruptions at the same time?

richsalz:
@levitte, what parts of #10289 show the callback mechanism you mentioned?  That's a big PR with many changes.  It'd be nice to look at the callback stuff in isolation if reasonable.

apps/fipsinstall.c
slontis: The callback gets passed 3 fields currently
(1) phase (e,g: begin, corrupt, pass, fail)
(2) desc
(3) type

The user callback decides what to do with these fields.

desc could possibly be used by more that 1 self test, so it could corrupt multiple tests.
if both type and desc are set then it checks for both being true (i.e its an AND operation).

richsalz: Perhaps one way to integrate both callbacks is to pass a void* that is a structure containing the custom info?  I dunno, wait for Levitte to reply :)

slontis:
> In #10289, I have a callback mechanism that is quite similar to yours, except it uses a common structure instead a direct function pointer, to get some kind of closure mechanism, and also to make it possible to "translate" to whatever kind of function on libcrypto side.
> I wonder, should we keep the fips self test cb separate, or should we eventually merge them?

I am thinking that it needs a seperate callback. I want to be able to test for a null self test callback and then do nothing in this case - having an all emcompassing callback seems like the wrong approach to me (unless I am not understanding what you are saying).
Also we will need another callback for the CAVS data testing. This would also be an optional thing that you do not normally want enabled (As it has to do nasty stuff that would generally be frowned upon in normal use cases - such as returning internals of signature data). I am undecided if this CAVS testing one needs to be conditionally enabled via a #ifdef that is just used by the labs.

levitte:
@richsalz, have a look at the commits tab, there's a separate commit that lays out the callback mechanism

richsalz:
Thanks, I took a look.  This is for a single callback function if I read the code right.  Not sure how useful that is.

slontis:
@levitte - what do I need to do to move this forward?

mspncp:
Just an idea which came to my mind today (cc @openssl/omc):

Wouldn't it be useful to have the KATs as an (optional) feature for the default provider as well? (With a shared implementation.)

Enabling the self test for the default provider could be a Configure option (default: enabled) combined with a runtime configuration option (default: disabled).

mspncp:
I am thinking of European companies or governments which might not be so keen about FIPS, but would appreciate power on self tests.

mspncp:
> power on self tests.

Wrong word. The tests in this pr are only executed at installation time right? But in either  case it would be helpful to make the test suites available also for the default provider (possible loosening some restrictions).

slontis:
> But in either case it would be helpful to make the test suites available also for the default provider
This sounds outside the scope of what we are trying to achieve for 3.0.
Note that 'test suites' is not really what you end up with the FIPS self tests. Its just a small subset used as a sanity check. I dont really have time to 'feature creep' this to something else at the moment :)

slontis:
@paulidale this has been rebased so your approval is no longer valid.

slontis:
@levitte are you happy with the callback used here?

slontis:
ping - rebased again..

crypto/self_test_core.c
```diff
@@ -0,0 +1,20 @@
...
+ * Licensed under the Apache License 2.0 (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
+ */
+
+#include <openssl/self_test.h>
+
+static OSSL_SELF_TEST_event_fn *ossl_self_test_cb = NULL;
+void OSSL_SELF_TEST_set_callback(OSSL_SELF_TEST_event_fn *cb)
```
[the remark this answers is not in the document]
levitte: So, er, multiplexing?

doc/man3/OSSL_SELF_TEST_set_callback.pod
```diff
@@ -0,0 +1,44 @@
...
+ void OSSL_SELF_TEST_set_callback(OSSL_SELF_TEST_event_fn *cb);
+ OSSL_SELF_TEST_event_fn *OSSL_SELF_TEST_get_callback(void);
+
+=head1 DESCRIPTION
+
+OSSL_SELF_TEST_set_callback() sets the function B<cb> that is called during
+fips self testing. The callback function has an array of B<OSSL_PARAM> that
+contains information about the self test event that relate to the phase, type and
+description. During the 'corrupt' phase the callback may return 0 to force the
+self test to fail.
```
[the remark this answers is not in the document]
levitte: I haven't looked at the self test much at all, so have an "external" view when reading this, and I don't understand at all. What are "phases", what does "corrupt" mean, how is that related to the parameters, and if I would want to write a FIPS provider, what do I need to do to be a well behaved provider? Is this documented anywhere? Will this be documented? Please link.

doc/man3/OSSL_SELF_TEST_set_callback.pod
```diff
@@ -0,0 +1,44 @@
...
+
+=head1 RETURN VALUES
+
+OSSL_SELF_TEST_get_callback() returns the callback that has been set,
```
levitte: Incomplete

include/openssl/core_names.h
```diff
@@ -40,6 +40,22 @@ extern "C" {
...
+ * A string for a self test callback's type.
+ * Type: OSSL_PARAM_UTF8_STRING
+ */
+#define OSSL_PROV_PARAM_SELF_TEST_TYPE   "st-type"
+/*
+ * A string for a self test callback's description.
+ * Type: OSSL_PARAM_UTF8_STRING
+ */
+#define OSSL_PROV_PARAM_SELF_TEST_DESC   "st-desc"
+
```
levitte: I like these comments, maybe we should do this consistently.

include/openssl/self_test.h
```diff
@@ -0,0 +1,61 @@
...
+# pragma once
+
+# include <openssl/params.h>
+
+/* The test event phases */
+# define SELF_TEST_PHASE_NONE     "None"
+# define SELF_TEST_PHASE_START    "Start"
+# define SELF_TEST_PHASE_CORRUPT  "Corrupt"
+# define SELF_TEST_PHASE_PASS     "Pass"
+# define SELF_TEST_PHASE_FAIL     "Fail"
```
levitte: Public header, please let the names start with `OSSL_`

include/openssl/self_test.h
```diff
@@ -0,0 +1,61 @@
...
+# define SELF_TEST_DESC_SIGN_RSA       "RSA"
+# define SELF_TEST_DESC_SIGN_ECDSA     "ECDSA"
+# define SELF_TEST_DESC_DRBG_CTR       "CTR"
+# define SELF_TEST_DESC_DRBG_HASH      "HASH"
+# define SELF_TEST_DESC_DRBG_HMAC      "HMAC"
+# define SELF_TEST_DESC_KA_ECDH        "ECDH"
+# define SELF_TEST_DESC_KA_ECDSA       "ECDSA"
+# define SELF_TEST_DESC_KDF_HKDF       "HKDF"
+
+typedef int (OSSL_SELF_TEST_event_fn)(OSSL_PARAM *params);
```
levitte: This is almost exactly the same as OSSL_CALLBACK...

apps/fipsinstall.c
```diff
@@ -49,6 +57,9 @@ const OPTIONS fipsinstall_options[] = {
     {"mac_name", OPT_MAC_NAME, 's', "MAC name"},
     {"macopt", OPT_MACOPT, 's', "MAC algorithm parameters in n:v form. "
                                 "See 'PARAMETER NAMES' in the EVP_MAC_ docs"},
+    {"noout", OPT_SELF_TEST_NO_LOG, '-', "Disable logging of self test events"},
+    {"corrupt_desc", OPT_SELF_TEST_CORRUPT_DESC, 's', "Corrupt a self test by description"},
+    {"corrupt_type", OPT_SELF_TEST_CORRUPT_TYPE, 's', "Corrupt a self test by type"},
```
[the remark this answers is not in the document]
richsalz: yeah, the defines are local to this file, remove the "self_test" part of the name.

crypto/self_test_core.c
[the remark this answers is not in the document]
slontis: The original intention was that this mechanism was supposed to be an application callback that the user provides to receive information back when the self tests are running at startup for the FIPS module (I dont quite get why an OPENSSL_CTX is needed for this? - it runs from the provider so it knows its context). 
Do you also want to run self tests for the default provider (This would require quite a bit more work to figure out what needs to be passed to the default provider - i.e: I really dont want to do integrity checks on the default provider because it is all loaded via configuration entries).

include/openssl/core_names.h
slontis: I just simplified them to match the others.

include/openssl/self_test.h
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright 2019 The OpenSSL Project Authors. All Rights Reserved.
+ *
+ * Licensed under the OpenSSL license (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
+ */
+
+#ifndef OPENSSL_SELF_TEST_H
```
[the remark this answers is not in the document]
slontis: I would prefer to keep it seperate.

crypto/self_test_core.c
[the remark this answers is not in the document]
slontis: Hmm looking st this a bit more - if I use OSSL_CALLBACK I can kill 2 birds... since it already has an args param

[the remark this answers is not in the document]
levitte: Note that the semantics of that void argument is that it's whatever the caller chose to pass along with the callback itself.

include/openssl/core_numbers.h
```diff
@@ -135,6 +136,9 @@ OSSL_CORE_MAKE_FUNC(int, BIO_free, (BIO *bio))
 OSSL_CORE_MAKE_FUNC(int, BIO_vprintf, (BIO *bio, const char *format,
                                        va_list args))
 
+#define OSSL_FUNC_SELF_TEST_CB                28
+OSSL_CORE_MAKE_FUNC(OSSL_SELF_TEST_event_fn *, SELF_TEST_CB, (void))
```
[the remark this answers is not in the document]
slontis: Removed and use OSSL_CALLBACK instead..

crypto/self_test_core.c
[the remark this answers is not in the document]
mattcaswell: What I'm thinking is that you could have more than one FIPS provider loaded at the same time into different OPENSSL_CTXs. You might want to set a callback on one of those OPENSSL_CTXs but not the other.

doc/man3/OSSL_SELF_TEST_set_callback.pod
```diff
@@ -0,0 +1,257 @@
...
+fips providers are present.
+
+The OSSL_PARAM names used are:
+
+=over 4
+
+=item "st-phase" (B<OSSL_PROV_PARAM_SELF_TEST_PHASE>) <UTF8 string>
+
+Each self test calls the callback 3 times with the following string values
+for the phase.
```
levitte: ...

So...  when you read a section 3 manual, you expect a reference description of the functions, and that's about it?  I find it a little unexpected to find a description of the FIPS provider in this page.  I would rather see that in a different file, and I think `doc/man7` is the appropriate place for it (nothing stops you from linking to it from the SEE ALSO in this file).  That would also allow us to add more documentation on that provider and its behaviour.

I'm unsure what name it would have, though...  `doc/man7/provider-*.pod` is kinda taken to describe the libcrypto\<-\>provider interface.  `doc/man7/FIPS-provider.pod`?  `doc/man7/OSSL_PROVIDER-FIPS.pod` (same pattern that we've used for some implementations of EVP APIs)?

slontis: Are you saying you want all the OSSL_CALLBACK related stuff in another file?

levitte: Uhm, not sure how to answer that question, so let me try like this... 

`OSSL_SELF_TEST_set_callback` looks like a general function that could be useful with any provider, not just our FIPS provider. Technically speaking, that is possible, and would be feasible at least with any other FIPS provider. Therefore, the general description of those functions belong here.

The description of the `OSSL_PARAM` values, however, are specific to our FIPS provider. There's nothing that I can see that obligates any other FIPS provider to pass those values. They belong in another file that describes our FIPS provider, and the programming example as well.

levitte: While you're documenting, a description of the function added in `core_numbers.h` should be added in `doc/man7/provider-base.pod`

include/openssl/configuration.h
```diff
@@ -0,0 +1,126 @@
+/*
+ * WARNING: do not edit!
+ * Generated by Makefile from include/openssl/configuration.h.in
```
levitte: This file shouldn't be verwioned

doc/man3/OSSL_SELF_TEST_set_callback.pod
levitte: I realised that there's another possible view on this; is your intention that these functions aren't as general as they appear, but are rather very specific and only useful with our FIPS provider and simply shouldn't be used with any other provider?

That *is* a possibility... But in that case, that should be made *very* clear. For the moment, nothing in this manual or in the implementation suggest such a limitation.

slontis:
Updated docs and rebased..

apps/fipsinstall.c
slontis: Changed to use OSSL_CALLBACK

slontis: Callback is the same as other callback mechanisms using OSSL_CALLBACK

doc/man3/OSSL_SELF_TEST_set_callback.pod
slontis: docs updated

slontis:
ping

slontis:
rebased again..

doc/man3/OSSL_SELF_TEST_set_callback.pod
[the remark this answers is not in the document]
slontis: docs added

crypto/self_test_core.c
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Licensed under the Apache License 2.0 (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
+ */
+
+#include <openssl/self_test.h>
+
+static OSSL_CALLBACK *ossl_self_test_cb = NULL;
```
mattcaswell: OPENSSL_CTX? Will we want to be able to set a callback in one OPENSSL_CTX but not another? This sets one callback globally. We've been trying to move away from global data.

doc/man7/OSSL_PROVIDER-FIPS.pod
```diff
@@ -0,0 +1,260 @@
...
+
+=item "DRBG" (B<OSSL_SELF_TEST_TYPE_DRBG>)
+
+Known answer test for a Deterministic Random Bit Generator.
+
+=item "Pairwise_Consistency_Test" (B<OSSL_SELF_TEST_TYPE_PCT>)
+
+Conditional test that is run during the generation of key pairs.
+
+The "Module_Integrity" self_test is always run at startup.
```
mattcaswell: Shouldn't these two paragraphs be *after* the `=back` that follows them?

crypto/self_test_core.c
slontis: That stuff hurts my brain..
You mean use OPENSSL_CTX_METHOD and openssl_ctx_get_data()

levitte: So, err, that takes us back to callback dispatching on a per OPENSSL_CTX basis, doesn't it?  I.e. more or less my original OSSL_CALLBACK proposal (#10390) :wink:

slontis: Not sure #10390 is really relevant here.
@mattcaswell How do I move this forward? This PR has been kicking around for a few months now :)

mattcaswell: > You mean use OPENSSL_CTX_METHOD and openssl_ctx_get_data()?

Yes, exactly that.

slontis:
@paulidale Your approval no longer counts :) I dont need you to reapprove..

crypto/self_test_core.c
slontis: Yeah I tried doing that, but this just is not going to work per libctx for self tests..

The fipinstall application does not know what its libctx is as it loads via configuration - and it needs to set the callback before doing that. i.e it just calls CONF_modules_load()

mattcaswell: So in that case it just passes NULL through for the libctx in order to get the default. The function that implements setting the callback will then resolve that to the "real" libctx via `openssl_ctx_get_concrete`

slontis: There is no default libctx associated with the fips provider...

mattcaswell: The FIPS install applications does not run inside the FIPS provider. It links against libcrypto which does have a default context.

slontis: Fair enough.. But the callback gets called from the fips provider for the install.
How would it know which libctx to use in order to find the callback in the exdata (because it wont be the fips libctx). Yeah we could use NULL here too - but is that what we want?
A single callback associated with the default libctx? Storing the libctx in the ctx_method struct isnt going to help for this case..

slontis: I think a single callback that gets passed the libctx is probably a better way to handle this (which it already does). This callback is really meant for diagnostic testing.

mattcaswell: > How would it know which libctx to use in order to find the callback in the exdata

The set and get callback functions are modified to take a libctx as a parameter:

```` C
void OSSL_SELF_TEST_set_callback(OPENSSL_CTX *libctx, OSSL_CALLBACK *cb);
OSSL_CALLBACK *OSSL_SELF_TEST_get_callback(OPENSSL_CTX *libctx);
````

The application calls OSSL_SELF_TEST_set_callback() and passes the libctx that should be used (or NULL for the default libctx).

The implementation of OSSL_SELF_TEST_set_callback() stores the provided callback function pointer in the libctx that it has been passed.

The implementation of OSSL_SELF_TEST_get_callback() looks up the callback to be used in the libctx and returns it.

fipsprov.c calls OSSL_get_self_test_cb() which just calls OSSL_SELF_TEST_get_callback() under the covers. It passes as an argument the result of calling c_get_libctx, i.e. something like this:

    OSSL_self_test_cb_fn *fn = OSSL_get_self_test_cb(in);

    selftest_params.event_cb = (fn != NULL) ? fn(c_get_libctx(provider)) : NULL;

Comment (slontis) [reply crypto/self_test_core.c]: I was doing this locally except for the fn(c_get_libctx(provider))..
So the 'provider' pointer in this case is the fips provider (which just happens to have a NULL libctx during the self test - which sounds like it works by accident).

Comment (slontis) [reply crypto/self_test_core.c]: I have updated the PR to include the OPENSSL_CTX

File: doc/man3/OSSL_SELF_TEST_set_callback.pod [hunk 17]
```diff
@@ -0,0 +1,46 @@
...
+
+ void OSSL_SELF_TEST_set_callback(OPENSSL_CTX *ctx, OSSL_CALLBACK *cb);
+ OSSL_CALLBACK *OSSL_SELF_TEST_get_callback(void);
+
+=head1 DESCRIPTION
+
+OSSL_SELF_TEST_set_callback() sets the function B<cb> that is called during
+self testing that is associated with a given library context B<ctx>.
+The callback function receives information about a running self test and may
+return a result.
```
mattcaswell: A reference to the page which describes the callback should be given here.

doc/man3/OSSL_SELF_TEST_set_callback.pod
```diff
@@ -0,0 +1,46 @@
...
+
+ #include <openssl/self_test.h>
+
+ void OSSL_SELF_TEST_set_callback(OPENSSL_CTX *ctx, OSSL_CALLBACK *cb);
+ OSSL_CALLBACK *OSSL_SELF_TEST_get_callback(void);
+
+=head1 DESCRIPTION
+
+OSSL_SELF_TEST_set_callback() sets the function B<cb> that is called during
+self testing that is associated with a given library context B<ctx>.
```
mattcaswell: Link to the OPENSSL_CTX page?

doc/man3/OSSL_SELF_TEST_set_callback.pod
```diff
@@ -0,0 +1,46 @@
...
+The callback function receives information about a running self test and may
+return a result.
+
+=head1 RETURN VALUES
+
+OSSL_SELF_TEST_get_callback() returns the callback that has been set via
```
mattcaswell: The get_callback() function is not actually mentioned in the DESCRIPTION section above.

apps/fipsinstall.c
```diff
@@ -420,3 +445,53 @@ int fipsinstall_main(int argc, char **argv)
     free_config_and_unload(conf);
     return ret;
 }
+
+static int self_test_events(const OSSL_PARAM params[], void *arg)
+{
+    const OSSL_PARAM *p = NULL;
+    const char *phase = NULL, *type= NULL, *desc = NULL;
```
FdaSilvaYY: nit in `type=`

doc/man1/openssl-fipsinstall.pod
```diff
@@ -123,6 +140,13 @@ Verify that the configuration file F<fips.conf> contains the correct info:
           -section_name fips_install -mac_name HMAC -macopt digest:SHA256 \
           -macopt hexkey:000102030405060708090A0B0C0D0E0F10111213 -verify
 
+Corrupt any self tests which have the description 'SHA1':
+
+ openssl fipsinstall -module ./fips.so -out fips.conf -provider_name fips \
+         -section_name fipsinstall -mac_name HMAC -macopt digest:SHA256 \
+         -macopt hexkey:000102030405060708090A0B0C0D0E0F10111213 \
+         -corrupt_desc', 'SHA1'
```
FdaSilvaYY: copy paste error in `corrupt_desc` parameter value, IMHO ?

doc/man7/OSSL_PROVIDER-FIPS.pod
```diff
@@ -0,0 +1,260 @@
...
+
+  #include <openssl/self_test.h>
+
+  static OSSL_CALLBACK self_test_cb;
+
+  static int self_test_cb(const OSSL_PARAM params[], void *arg)
+  {
+    int ret = 0;
+    const OSSL_PARAM *p = NULL;
+    const char *phase = NULL, *type= NULL, *desc = NULL;
```
FdaSilvaYY: same nit : spaces around =

providers/fips/self_test_data.inc
```diff
@@ -0,0 +1,191 @@
...
+        ITM(aes_256_gcm_key),
+        ITM(aes_256_gcm_iv),
+        ITM(aes_256_gcm_aad),
+        ITM(aes_256_gcm_tag)
+    }
+};
+
+/*- KDF TEST DATA */
+
+static ST_KAT_NVP hkdf_ctrl[] =
```
FdaSilvaYY: this declaration can be const, I think.

providers/fips/self_test_data.inc
```diff
@@ -0,0 +1,191 @@
...
+{
+    { "digest", "SHA256" },
+    { "key", "secret" },
+    { "salt", "salt" },
+    { "info", "label" },
+    { NULL, NULL }
+};
+static unsigned char hkdf_expected[] = {
```
FdaSilvaYY: const ?

providers/fips/self_test_data.inc
```diff
@@ -0,0 +1,191 @@
...
+    0x2a, 0xc4, 0x36, 0x9f, 0x52, 0x59, 0x96, 0xf8, 0xde, 0x13
+};
+
+static ST_KAT_KDF st_kat_kdf_tests[] =
```
FdaSilvaYY: const ?

providers/fips/self_test_kats.c
```diff
@@ -0,0 +1,245 @@
...
+
+    ret = 1;
+end:
+    EVP_CIPHER_free(cipher);
+    EVP_CIPHER_CTX_free(ctx);
+    SELF_TEST_EVENT_onend(event, ret);
+    return ret;
+}
+
+static int self_test_kdf(ST_KAT_KDF *t, OSSL_ST_EVENT *event, OPENSSL_CTX *libctx)
```
FdaSilvaYY: self_test_kdf(**const** ST_KAT_KDF *t, ...

doc/man3/OSSL_SELF_TEST_set_callback.pod
slontis: reworded

doc/man3/OSSL_SELF_TEST_set_callback.pod
slontis: Added a link to the see also.. Seems redundant to have to add a link elsewhere in a pod with half a page of info in it.

doc/man3/OSSL_SELF_TEST_set_callback.pod
slontis: reword so the decription doesnt mention set_callback either..

doc/man1/openssl-fipsinstall.pod
slontis: No it is correct. The SHA256 refers to the HMAC integrity check.. SHA1 refers to the self test that we are going to corrupt.

crypto/self_test_core.c
```diff
@@ -0,0 +1,55 @@
...
+                                &self_test_set_callback_method);
+}
+
+void OSSL_SELF_TEST_set_callback(OPENSSL_CTX *libctx, OSSL_CALLBACK *cb)
+{
+    SELF_TEST_CB *stcb = get_self_test_callback(libctx);
+
+    if (stcb != NULL)
+        stcb->cb = cb;
+}
```
levitte: I would like to see a cbarg being passed as well. If the application wants an OPENSSL_CTX, then it will have to say so, but if it wants something else (which might contain an OPENSSL_CTX pointer for all I know), then it will do that.

The point is that the cbarg is supposed to be something passed from the code that sets up the callback.

slontis: done

providers/fips/self_test.h
```diff
@@ -25,8 +26,33 @@ typedef struct self_test_post_params_st {
...
 } SELF_TEST_POST_PARAMS;
 
+typedef struct st_event_st
+{
+    /* local state variables */
+    const char *phase;
+    const char *type;
+    const char *desc;
+    OSSL_CALLBACK *cb;
+    OPENSSL_CTX *libctx;
```
levitte: As far as I can tell, libctx isn't needed any more here.

providers/fips/self_test.h
```diff
@@ -25,8 +26,33 @@ typedef struct self_test_post_params_st {
...
+    OSSL_PARAM params[4];
+    void *cb_arg;
+
+} OSSL_ST_EVENT;
+
 int SELF_TEST_post(SELF_TEST_POST_PARAMS *st, int on_demand_test);
+int SELF_TEST_kats(OSSL_ST_EVENT *event, OPENSSL_CTX *libctx);
+
+void SELF_TEST_EVENT_init(OSSL_ST_EVENT *ev, OSSL_CALLBACK *cb, void *cbarg,
+                          OPENSSL_CTX *libctx);
```
levitte: So there is no need to pass it...

providers/fips/self_test_event.c
```diff
@@ -0,0 +1,95 @@
...
+    ev->params[n++] = OSSL_PARAM_construct_end();
+}
+
+void SELF_TEST_EVENT_init(OSSL_ST_EVENT *ev, OSSL_CALLBACK *cb, void *cbarg,
+                          OPENSSL_CTX *libctx)
+{
+    if (ev == NULL)
+        return;
+
+    ev->libctx = libctx;
```
levitte: ... nor to save it

providers/fips/self_test.h
slontis: Fixed

slontis:
Thanks for reviewing. Merged to master.

PR #10376, opened 2019-11-07
Parse large GOST ClientKeyExchange messages

beldmit:
Large GOST ClientKeyExchange messages are sent by VipNet CSP, one of
Russian certified products implementing GOST TLS, when a server
certificate contains 512-bit keys.

This behaviour was present in 1.0.2 branch and needs to be restored.

##### Checklist

### DIFF
--- a/ssl/statem/statem_srvr.c
+++ b/ssl/statem/statem_srvr.c
@@ -3353,6 +3353,13 @@ static int tls_process_cke_gost(SSL *s, PACKET *pkt)
                    SSL_R_DECRYPTION_FAILED);
          goto err;
     }
+     
+    if (PACKET_remaining(pkt) != 0) {
+         SSLfatal(s, SSL_AD_INTERNAL_ERROR, SSL_F_TLS_PROCESS_CKE_GOST,
+                   SSL_R_DECRYPTION_FAILED);
+         goto err;
+    }
+
     inlen = pKX->kxBlob->value.sequence->length;
     start = pKX->kxBlob->value.sequence->data;
 
--- a/ssl/statem/statem_srvr.c
+++ b/ssl/statem/statem_srvr.c
@@ -3344,20 +3344,20 @@ static int tls_process_cke_gost(SSL *s, PACKET *pkt)
        || pKX->kxBlob == NULL
        || ASN1_TYPE_get(pKX->kxBlob) != V_ASN1_SEQUENCE) {
          SSLfatal(s, SSL_AD_DECODE_ERROR, SSL_F_TLS_PROCESS_CKE_GOST,
-                   SSL_R_DECRYPTION_FAILED);
+                  SSL_R_DECRYPTION_FAILED);
          goto err;
     }
 
     if (!PACKET_forward(pkt, ptr - PACKET_data(pkt))) {
-         SSLfatal(s, SSL_AD_INTERNAL_ERROR, SSL_F_TLS_PROCESS_CKE_GOST,
-                   SSL_R_DECRYPTION_FAILED);
-         goto err;
+        SSLfatal(s, SSL_AD_INTERNAL_ERROR, SSL_F_TLS_PROCESS_CKE_GOST,
+                 SSL_R_DECRYPTION_FAILED);
+        goto err;
     }
      
     if (PACKET_remaining(pkt) != 0) {
-         SSLfatal(s, SSL_AD_INTERNAL_ERROR, SSL_F_TLS_PROCESS_CKE_GOST,
-                   SSL_R_DECRYPTION_FAILED);
-         goto err;
+        SSLfatal(s, SSL_AD_INTERNAL_ERROR, SSL_F_TLS_PROCESS_CKE_GOST,
+                 SSL_R_DECRYPTION_FAILED);
+        goto err;
     }
 
     inlen = pKX->kxBlob->value.sequence->length;
--- a/ssl/statem/statem_srvr.c
+++ b/ssl/statem/statem_srvr.c
@@ -3353,7 +3353,7 @@ static int tls_process_cke_gost(SSL *s, PACKET *pkt)
                  SSL_R_DECRYPTION_FAILED);
         goto err;
     }
-     
+
     if (PACKET_remaining(pkt) != 0) {
         SSLfatal(s, SSL_AD_INTERNAL_ERROR, SSL_F_TLS_PROCESS_CKE_GOST,
                  SSL_R_DECRYPTION_FAILED);
--- a/ssl/statem/statem_srvr.c
+++ b/ssl/statem/statem_srvr.c
@@ -25,8 +25,23 @@
 #include <openssl/md5.h>
 #include <openssl/trace.h>
 #include <openssl/core_names.h>
+#include <openssl/asn1t.h>
 
 #define TICKET_NONCE_SIZE       8
+
+typedef struct {
+  ASN1_TYPE *kxBlob;
+  ASN1_TYPE *opaqueBlob;
+} GOST_KX_MESSAGE;
+
+DECLARE_ASN1_FUNCTIONS(GOST_KX_MESSAGE)
+
+ASN1_SEQUENCE(GOST_KX_MESSAGE) = {
+  ASN1_SIMPLE(GOST_KX_MESSAGE,  kxBlob, ASN1_ANY),
+  ASN1_OPT(GOST_KX_MESSAGE, opaqueBlob, ASN1_ANY),
+} ASN1_SEQUENCE_END(GOST_KX_MESSAGE)
+
+IMPLEMENT_ASN1_FUNCTIONS(GOST_KX_MESSAGE)
 
 static int tls_construct_encrypted_extensions(SSL *s, WPACKET *pkt);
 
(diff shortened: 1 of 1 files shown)


ssl/statem/statem_srvr.c
```diff
@@ -3374,28 +3374,26 @@ static int tls_process_cke_gost(SSL *s, PACKET *pkt)
...
+    if (asn1len >= 0x81) {
         /*
-         * Long form length. Should only be one byte of length. Anything else
-         * isn't supported.
-         * We did a successful peek before so this shouldn't fail
+         * Long form length.
          */
-        if (!PACKET_forward(pkt, 1)) {
+        size_t length = asn1len & 0x7F;
+        if (!PACKET_forward(pkt, length)) {
```
mattcaswell: This doesn't look right. Previously we supported two formats:

1) Long form with length with a single byte of length. In this case `pkt` looks like this:

`0x81, len (1 byte), len bytes of data`


We then moved forward by 1 byte so that pkt then looks like:

`len (1 byte), len bytes of data`

2) Short form length

In that case pkt looked like

`len (1 byte), len bytes of data`

In both cases `pkt` ended up with 1 byte of length data followed by the actual data, and therefore we could use the `PACKET_as_length_prefixed_1` call to read it. That expects exactly 1 byte of length data, followed by the actual data (which should extend to the end of the packet).

With these new changes lets say we have a packet with long form length where `asn1len == 0x82`. In this case `pkt` will look like this:

`0x82, hi-byte-of-len, lo-byte-of-len, len bytes of data`

In this example `length` will have the value 2 to indicate there are 2 bytes of length data. So after the forward we will have:

`lo-byte-of-len, len bytes of data`

This means that calling `PACKET_as_length_prefixed_1` will fail because we will only read part of the data. Since that call expects us to consume *all* the data in the packet it fails. Changing it to PACKET_get_length_prefixed_1 doesn't really solve the problem. It just stops the call from failing. It still doesn't read all the data, and doesn't consume all the PACKET - it just throws it away instead.

beldmit: Yes, it's the desired behavior. Here is the KX blob struct:
```
  0 509: SEQUENCE {
  4 239:   SEQUENCE {
  7  40:     SEQUENCE {
  9  32:       OCTET STRING
       :         47 D8 50 C0 57 9D 30 1A 51 50 F8 7D 0B AF C5 EC
       :         C3 DE 46 85 A4 95 A2 6E 01 5A A3 1A 85 6D 0F F4
 43   4:       OCTET STRING 82 8B 09 71
       :       }
 49 194:     [0] {
 52   9:       OBJECT IDENTIFIER '1 2 643 7 1 2 5 1 1'
 63 170:       [0] {
 66  33:         SEQUENCE {
 68   8:           OBJECT IDENTIFIER '1 2 643 7 1 1 1 2'
 78  21:           SEQUENCE {
 80   9:             OBJECT IDENTIFIER '1 2 643 7 1 2 1 2 1'
 91   8:             OBJECT IDENTIFIER '1 2 643 7 1 1 2 3'
       :             }
       :           }
101 132:         BIT STRING, encapsulates {
105 128:           OCTET STRING
       :             8B 69 FC EC 67 D2 46 C6 A5 1F 25 D2 37 19 3E 27
       :             B5 F9 7E E5 37 E7 9B 65 76 11 AE DD EB E3 21 21
       :             B8 69 A0 EC 0B F4 A9 B6 5F 5F 26 14 07 5D 1F DA
       :             36 43 F0 76 B4 52 56 64 DF FF C4 6C 23 0E 7B 1A
       :             D8 84 20 59 69 34 4B 2D C2 E9 0C 2A CA 68 E7 EA
       :             0D 2D 88 93 9E 93 2C 0F F2 E4 2C F1 B5 F2 78 76
       :             C0 83 DF 9B D4 37 55 4C C4 1C 3A 82 0D B8 CC 02
       :             1A B1 51 C3 DF A1 4A 0D 52 36 7E BD 6A 86 D9 C4
       :           }
       :         }
236   8:       OCTET STRING 8F AA 92 02 D8 F2 C0 D3
       :       }
       :     }
246 263:   SEQUENCE {
250 259:     SEQUENCE {
254  65:       SEQUENCE {
256  40:         SEQUENCE {
258  32:           OCTET STRING
       :             47 D8 50 C0 57 9D 30 1A 51 50 F8 7D 0B AF C5 EC
       :             C3 DE 46 85 A4 95 A2 6E 01 5A A3 1A 85 6D 0F F4
292   4:           OCTET STRING 82 8B 09 71
       :           }
298  21:         [0] {
300   9:           OBJECT IDENTIFIER '1 2 643 7 1 2 5 1 1'
311   8:           OCTET STRING 8F AA 92 02 D8 F2 C0 D3
       :           }
       :         }
321 189:       OCTET STRING, encapsulates {
324 186:         OCTET STRING
       :           ':Licinfo_itcs:dcgGN+4vmMNL/Yc1pRtKSWmvUv6xu5CZKT'
       :           'JTMYvPktB0ui0KN3BWkuR/nkUhwCiVifkJhEIg3wpHt1rEI+'
       :           'JQNkTEr1oQN+vxZmDQdOPJNC0CABwWQg2lcJloPayYJUM3Z9'
       :           'pWDVrc18o0+Wg4sOzZarEYprOJWePs7skFGH5/1AA='
       :         }
       :       }
       :     }
       :   }
```
So we take the 1st element of the sequence containing the wrapped key, and ignore the rest as it does not contain any information necessary for us.

mattcaswell: What?? And that *works*? So we end up using what ever random length is in the lo-byte (which presumably could be anything from 0 to 255) and that's the amount of data we feed to the following EVP_PKEY_decrypt()? Surely that will just fail unless you happen to pass it the correct length?

beldmit: It works, but I will rewrite it using a proper ASN1 parser.
Turned to WIP.

mattcaswell: One option is to add a new packet function...something like `PACKET_as_der_length_prefixed()` or similar.

beldmit: I always disliked this piece of code, both for a client and a server, but never had a reason to rewrite it.

https://tools.ietf.org/html/draft-chudov-cryptopro-cptls-04#section-3.6 describes a basic form of the message. The OPTIONAL element was never supported in practice, but formally leaves a chance for implementors to add extra information. So I'll try to extract the necessary part using i2d code, and the GOST engine invokes i2d function for the extracted blob, as it does now.

beldmit:
@mattcaswell, I have no idea how to write it more defensively...

beldmit:
I'm afraid it will require a separate PR for the 1.1.1 branch because there are some differences in the code.

beldmit:
@mattcaswell Could you please take a look?

beldmit:
ping @mattcaswell

beldmit:
Ping @mattcaswell

ssl/statem/statem_srvr.c
```diff
@@ -3366,42 +3381,26 @@ static int tls_process_cke_gost(SSL *s, PACKET *pkt)
...
-        SSLfatal(s, SSL_AD_DECODE_ERROR, SSL_F_TLS_PROCESS_CKE_GOST,
-                 SSL_R_DECRYPTION_FAILED);
-        goto err;
-    } /* else short form length */
 
-    if (!PACKET_as_length_prefixed_1(pkt, &encdata)) {
-        SSLfatal(s, SSL_AD_DECODE_ERROR, SSL_F_TLS_PROCESS_CKE_GOST,
-                 SSL_R_DECRYPTION_FAILED);
-        goto err;
+    if (!PACKET_forward(pkt, ptr - PACKET_data(pkt))) {
```
mattcaswell: Should we not verify that `PACKET_remaining` is 0 after this forward? Otherwise it means we have trailing data in the PACKET.

beldmit: Sure. Will fix.

ssl/statem/statem_srvr.c
```diff
@@ -3366,42 +3381,33 @@ static int tls_process_cke_gost(SSL *s, PACKET *pkt)
...
+         SSLfatal(s, SSL_AD_INTERNAL_ERROR, SSL_F_TLS_PROCESS_CKE_GOST,
+                   SSL_R_DECRYPTION_FAILED);
```
mattcaswell: Nit: Indentation seems off here

ssl/statem/statem_srvr.c
```diff
@@ -3320,42 +3335,33 @@ static int tls_process_cke_gost(SSL *s, PACKET *pkt)
...
+         goto err;
     }
-    inlen = PACKET_remaining(&encdata);
-    start = PACKET_data(&encdata);
+     
+    if (PACKET_remaining(pkt) != 0) {
+         SSLfatal(s, SSL_AD_INTERNAL_ERROR, SSL_F_TLS_PROCESS_CKE_GOST,
+                   SSL_R_DECRYPTION_FAILED);
```
mattcaswell: And here

ssl/statem/statem_srvr.c
beldmit: Fixed. Ok now?

ssl/statem/statem_srvr.c
beldmit: Fixed. Ok now?

beldmit:
Committed to master, but the commit does not clearly cherry-picks to 1.1.1 because of ```#include``` changes.

@mattcaswell, should I create a separate PR?

beldmit:
Just for the record: the commit id is 4e3ee452d091615e52a43e6e7c6db7d09e260353

beldmit:
Merged to master, separate PR for 1.1.1

Thanks!

PR #10377, opened 2019-11-07
Workaround for Windows-based GOST implementations

beldmit:
Many Windows-based GOST TLS implementations are unable to extend the
list of supported SignatureAlgorithms because of lack of the necessary
callback in Windows. So for TLS 1.2 it makes sense to imply the support
of GOST algorithms in case when the GOST ciphersuites are present.

##### Checklist


beldmit:
Merged. Thanks!

PR #10378, opened 2019-11-07
Windows-based GOST implementations compatibility

beldmit:
Many Windows-based GOST TLS implementations are unable to extend the
list of supported SignatureAlgorithms because of lack of the necessary
callback in Windows. So for TLS 1.2 it makes sense to imply the support
of GOST algorithms in case when the GOST ciphersuites are present.

This is a backport of #10377 to 1.1.1 branch

##### Checklist

### DIFF
--- a/ssl/t1_lib.c
+++ b/ssl/t1_lib.c
@@ -2774,6 +2774,26 @@ int tls_choose_sigalg(SSL *s, int fatalerrs)
 #endif
                         break;
                 }
+#ifndef OPENSSL_NO_GOST
+                /*
+                 * Some Windows-based implementations do not send GOST algorithms indication
+                 * in supported_algorithms extension, so when we have GOST-based ciphersuite,
+                 * we have to assume GOST support.
+                 */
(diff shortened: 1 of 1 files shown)


beldmit:
@paulidale, many thanks!
What about #10377?

beldmit:
@mspncp, could you please take a look? I think this PR along with #10377 should be labeled as "Ready to merge" because 24 hours have passed. Does the label "ready to merge" apply automatically?

Thanks!

paulidale:
The tag isn't automatic yet.  It also isn't mandatory yet.
I've added it.

paulidale:
I think an automatic email after the 24 hours is a good idea too.

beldmit:
> The tag isn't automatic yet. It also isn't mandatory yet.
> I've added it.

My fault. I thought it is.

mspncp:
> My fault. I thought it is.

For the moment, you can set it yourself until it is done automatically (any progress @levitte?).

But the [approval: *] tags are meant to be used like a state, so there shouldn't be multiple of them set at the same time. That's why I removed the [approval: done] tag.

beldmit:
Merged. Thanks!

PR #10380, opened 2019-11-07

raniervf:
##### Checklist

### DIFF
--- a/apps/s_server.c
+++ b/apps/s_server.c
@@ -2594,7 +2594,8 @@ static int sv_body(int s, int stype, int prot, unsigned char *context)
                     continue;
                 }
                 if (buf[0] == 'P') {
-                    BIO_write(SSL_get_wbio(con), "Lets print some clear text\n", sizeof("Lets print some clear text\n") -1);
+                    static const char str[] = "Lets print some clear text\n";
+                    BIO_write(SSL_get_wbio(con), "Lets print some clear text\n", sizeof(str) -1);
                 }
                 if (buf[0] == 'S') {
                     print_stats(bio_s_out, SSL_get_SSL_CTX(con));
--- a/apps/s_server.c
+++ b/apps/s_server.c
@@ -2524,7 +2524,7 @@ static int sv_body(int s, int stype, int prot, unsigned char *context)
                 for (j = i - 1; j >= 0; j--) {
                     buf[j + lf_num] = buf[j];
                     if (buf[j] == '\n') {
-                        lf_num--;
+                        lf_num--; 
                         i++;
                         buf[j + lf_num] = '\r';
                     }
@@ -2595,7 +2595,7 @@ static int sv_body(int s, int stype, int prot, unsigned char *context)
                 }
                 if (buf[0] == 'P') {
                     static const char str[] = "Lets print some clear text\n";
-                    BIO_write(SSL_get_wbio(con), "Lets print some clear text\n", sizeof(str) -1);
+                    BIO_write(SSL_get_wbio(con), str, sizeof(str) -1);
                 }
                 if (buf[0] == 'S') {
                     print_stats(bio_s_out, SSL_get_SSL_CTX(con));
--- a/apps/s_server.c
+++ b/apps/s_server.c
@@ -3545,6 +3545,7 @@ static int generate_session_id(SSL *ssl, unsigned char *id,
 {
     unsigned int count = 0;
     unsigned int session_id_prefix_len = strlen(session_id_prefix);
+  
     do {
         if (RAND_bytes(id, *id_len) <= 0)
             return 0;
--- a/apps/s_server.c
+++ b/apps/s_server.c
@@ -2594,8 +2594,7 @@ static int sv_body(int s, int stype, int prot, unsigned char *context)
                     continue;
                 }
                 if (buf[0] == 'P') {
-                    static const char *str = "Lets print some clear text\n";
-                    BIO_write(SSL_get_wbio(con), str, strlen(str));
+                    BIO_write(SSL_get_wbio(con), "Lets print some clear text\n", sizeof("Lets print some clear text\n") -1);
                 }
                 if (buf[0] == 'S') {
                     print_stats(bio_s_out, SSL_get_SSL_CTX(con));
@@ -3544,6 +3543,7 @@ static int generate_session_id(SSL *ssl, unsigned char *id,
                                unsigned int *id_len)
 {
     unsigned int count = 0;
+    unsigned int session_id_prefix_len = strlen(session_id_prefix);
     do {
         if (RAND_bytes(id, *id_len) <= 0)
             return 0;
@@ -3555,8 +3555,8 @@ static int generate_session_id(SSL *ssl, unsigned char *id,
          * conflicts.
          */
         memcpy(id, session_id_prefix,
-               (strlen(session_id_prefix) < *id_len) ?
-               strlen(session_id_prefix) : *id_len);
+               (session_id_prefix_len < *id_len) ?
+                session_id_prefix_len : *id_len);
     }
     while (SSL_has_matching_session_id(ssl, id, *id_len) &&
            (++count < MAX_SESSION_ID_ATTEMPTS));
--- a/apps/s_server.c
+++ b/apps/s_server.c
@@ -2524,7 +2524,7 @@ static int sv_body(int s, int stype, int prot, unsigned char *context)
                 for (j = i - 1; j >= 0; j--) {
                     buf[j + lf_num] = buf[j];
                     if (buf[j] == '\n') {
-                        lf_num--; 
+                        lf_num--;
                         i++;
                         buf[j + lf_num] = '\r';
                     }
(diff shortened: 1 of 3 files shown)


apps/s_server.c
```diff
@@ -2594,8 +2594,7 @@ static int sv_body(int s, int stype, int prot, unsigned char *context)
                     continue;
                 }
                 if (buf[0] == 'P') {
-                    static const char *str = "Lets print some clear text\n";
-                    BIO_write(SSL_get_wbio(con), str, strlen(str));
+                    BIO_write(SSL_get_wbio(con), "Lets print some clear text\n", sizeof("Lets print some clear text\n") -1);
```
richsalz: Make the text string a "static const char text[] = " and then use sizeof on that.

raniervf:
The declaration static const char text[]  is complete unnecessary.

mspncp:
@raniervf the other pr's are closed now. Please remember to go edit the files in your [patch-2](https://github.com/raniervf/openssl/tree/patch-2) branch, if you wan't to make further changes and commit them to this pull request.

raniervf:
Ok, thank you.

apps/s_server.c
mspncp: > The declaration static const char text[] is complete unnecessary.

No @raniervf, it is not unnecessary. It avoids duplicating the message as you did. @richsalz's suggestion was actually to change the `const char *` into a `const char []`, which gives a much nicer result:

```patch
index 5fa7c2fb42..08e73202ea 100644
--- a/apps/s_server.c
+++ b/apps/s_server.c
@@ -2594,8 +2594,8 @@ static int sv_body(int s, int stype, int prot, unsigned char *context)
                     continue;
                 }
                 if (buf[0] == 'P') {
-                    static const char *str = "Lets print some clear text\n";
-                    BIO_write(SSL_get_wbio(con), str, strlen(str));
+                    static const char str[] = "Lets print some clear text\n";
+                    BIO_write(SSL_get_wbio(con), str, sizeof(str)-1);
                 }
                 if (buf[0] == 'S') {
                     print_stats(bio_s_out, SSL_get_SSL_CTX(con));
```

If the difference between the `const char *` and the `const char []` is not obvious to you, the following toy example might help to understand it ;-)


**testo.c**
```C
#include <stdio.h>

static const char* text1   = "Lets print some clear text\n";
static const char  text2[] = "Lets print some clear text\n";

void main(int argc, char* argv[])
{
    printf("sizeof(text1) = %lu\n", sizeof(text1));
    printf("sizeof(text2) = %lu\n", sizeof(text2));
}
```

```bash
$ make testo

$ ./testo 
sizeof(text1) = 8
sizeof(text2) = 28
```

ssl/ssl_rsa.c
```diff
@@ -934,7 +937,9 @@ int SSL_CTX_use_serverinfo_file(SSL_CTX *ctx, const char *file)
         SSLerr(SSL_F_SSL_CTX_USE_SERVERINFO_FILE, ERR_R_SYS_LIB);
         goto end;
     }
-
+  
+    namePrefix1_len = strlen(namePrefix1);
+    namePrefix2_len = strlen(namePrefix2);
```
InfoHunter: `sizeof` can be used instead of `strlen` in this case.

ssl/ssl_rsa.c
```diff
@@ -951,19 +956,20 @@ int SSL_CTX_use_serverinfo_file(SSL_CTX *ctx, const char *file)
                 break;
         }
         /* Check that PEM name starts with "BEGIN SERVERINFO FOR " */
-        if (strlen(name) < strlen(namePrefix1)) {
+        name_len = strlen(name);
+        if (name_len < namePrefix1_len) {
```
InfoHunter: I just moved my previous comment here:

> Or you can replace strlen() into sizeof here to avoid introducing more local variables.

apps/s_server.c
InfoHunter: It seems he meant the `static` is not necessary

mspncp: In my opinion the `static` is necessary, because it it is static const then the compiler does not have to copy the characters onto the stack on every call. (He might be smart enough to do it anyway, but it's better to be explicit.)

apps/s_server.c
```diff
@@ -2594,7 +2594,8 @@ static int sv_body(int s, int stype, int prot, unsigned char *context)
                     continue;
                 }
                 if (buf[0] == 'P') {
-                    BIO_write(SSL_get_wbio(con), "Lets print some clear text\n", sizeof("Lets print some clear text\n") -1);
+                    static const char str[] = "Lets print some clear text\n";
+                    BIO_write(SSL_get_wbio(con), "Lets print some clear text\n", sizeof(str) -1);
```
mspncp: No need to duplicate the text:

```C
                    BIO_write(SSL_get_wbio(con), str, sizeof(str) -1);
```

mspncp:
@raniervf thank you, the reviewing workflow is much simpler now, since all changes go into a single branch.

Note that we will very likely squash the entire pull request into a single commit at the end, with a nice commit message.

ssl/ssl_rsa.c
```diff
@@ -914,10 +914,8 @@ int SSL_CTX_use_serverinfo_file(SSL_CTX *ctx, const char *file)
     long extension_length = 0;
     char *name = NULL;
     char *header = NULL;
-    char namePrefix1[] = "SERVERINFO FOR ";
-    char namePrefix2[] = "SERVERINFOV2 FOR ";
-    unsigned int namePrefix1_len;
-    unsigned int namePrefix2_len;
+    static char namePrefix1[] = "SERVERINFO FOR ";
+    static char namePrefix2[] = "SERVERINFOV2 FOR ";
```
mspncp: `static const char`. The _const_ allows the compiler to place the entire string into readonly memory and optimize away the variable entirely.

ssl/ssl_rsa.c
```diff
@@ -957,19 +953,19 @@ int SSL_CTX_use_serverinfo_file(SSL_CTX *ctx, const char *file)
...
             version = SSL_SERVERINFOV1;
         } else {
-            if (name_len < namePrefix2_len) {
+            if (name_len < sizeof(namePrefix2) - 1) {
                 SSLerr(SSL_F_SSL_CTX_USE_SERVERINFO_FILE,
                        SSL_R_PEM_NAME_TOO_SHORT);
                 goto end;
             }
-            if (strncmp(name, namePrefix2, namePrefix2_len) != 0) {
+            if (strncmp(name, namePrefix2, sieof(namePrefix2) - 1) != 0) {
```
mspncp: typo: sieof -> sizeof

raniervf:
Sorry, by typos errors. I believe which now agrees.

ssl/ssl_rsa.c
```diff
@@ -934,7 +935,7 @@ int SSL_CTX_use_serverinfo_file(SSL_CTX *ctx, const char *file)
         SSLerr(SSL_F_SSL_CTX_USE_SERVERINFO_FILE, ERR_R_SYS_LIB);
         goto end;
     }
-
+
```
t8m: trailing whitespace added

richsalz:
@raniervf you got feedback from two OpenSSL project members that they prefer "static char XXX[]" for the declarations (and me, but I don't vote).  You should change it.

mspncp:
> @raniervf you got feedback from two OpenSSL project members that they prefer "static char XXX[]" for the declarations (and me, but I don't vote). You should change it.

He already changed it, see [here](https://github.com/openssl/openssl/pull/10380/files#diff-65856f482216d2b6581d549dd52afa08R2597-R2598) and [here](https://github.com/openssl/openssl/pull/10380/files#diff-3dca9542ce522fa8763812df514d769bR917-R919).

richsalz:
"never mind"

apps/s_server.c
```diff
@@ -3544,6 +3544,7 @@ static int generate_session_id(SSL *ssl, unsigned char *id,
                                unsigned int *id_len)
 {
     unsigned int count = 0;
+    unsigned int session_id_prefix_len = strlen(session_id_prefix);
```
InfoHunter: A coding style nit: there should be a blank line after this line. (between variables declaration area and other code)

raniervf: Done. Thank you by review work.

ssl/ssl_rsa.c
```diff
@@ -934,7 +935,7 @@ int SSL_CTX_use_serverinfo_file(SSL_CTX *ctx, const char *file)
...
```
paulidale: Spurious whitespace?

raniervf:
"Spurius white space?"
Sorry @paulidale, I closed the conversation unintentionally, but I think, that fixed.
Here: https://github.com/openssl/openssl/pull/10380/commits/bfc27c3390e96fadc8f65487357b217981b0d06c

paulidale:
It looks like there is still a space on the empty line I tagged.  I'll fix this during merge.

paulidale:

PR #10381, opened 2019-11-07

raniervf:
##### Checklist

### DIFF
--- a/crypto/pem/pem_lib.c
+++ b/crypto/pem/pem_lib.c
@@ -818,6 +818,7 @@ static int get_header_and_data(BIO *bp, BIO **header, BIO **data, char *name,
         return 0;
     }
 
+    namelen = strlen(name);
     for (line = 0; ; line++) {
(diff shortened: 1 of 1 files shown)


crypto/pem/pem_lib.c
```diff
@@ -849,7 +850,6 @@ static int get_header_and_data(BIO *bp, BIO **header, BIO **data, char *name,
         /* Check for end of stream (which means there is no header). */
         if (strncmp(linebuf, endstr, ENDLEN) == 0) {
             p = linebuf + ENDLEN;
-            namelen = strlen(name);
```
InfoHunter: I think there is no need to move this outside since this `if` branch will finally reach to a `break`, which can break out of the loop.

t8m:
As @InfoHunter already commented this change does not bring any improvement so I am closing the PR.
There could be a code cleanup done by moving the declaration of the namelen inside the if block where the variable is actually used however that would be a different PR.

PR #10382, opened 2019-11-07

raniervf:
##### Checklist

### DIFF
--- a/ssl/ssl_rsa.c
+++ b/ssl/ssl_rsa.c
@@ -916,6 +916,9 @@ int SSL_CTX_use_serverinfo_file(SSL_CTX *ctx, const char *file)
     char *header = NULL;
     char namePrefix1[] = "SERVERINFO FOR ";
     char namePrefix2[] = "SERVERINFOV2 FOR ";
+    unsigned int namePrefix1_len;
+    unsigned int namePrefix2_len;
+    unsigned int name_len;
     int ret = 0;
     BIO *bin = NULL;
     size_t num_extensions = 0, contextoff = 0;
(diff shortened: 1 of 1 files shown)


ssl/ssl_rsa.c
```diff
@@ -916,6 +916,9 @@ int SSL_CTX_use_serverinfo_file(SSL_CTX *ctx, const char *file)
     char *header = NULL;
     char namePrefix1[] = "SERVERINFO FOR ";
     char namePrefix2[] = "SERVERINFOV2 FOR ";
+    unsigned int namePrefix1_len;
```
InfoHunter: you can just use `sizeof(namePrefix1) - 1` here and avoid more `strlen`s

ssl/ssl_rsa.c
```diff
@@ -951,19 +956,20 @@ int SSL_CTX_use_serverinfo_file(SSL_CTX *ctx, const char *file)
                 break;
         }
         /* Check that PEM name starts with "BEGIN SERVERINFO FOR " */
-        if (strlen(name) < strlen(namePrefix1)) {
+        name_len = strlen(name);
+        if (name_len < namePrefix1_len) {
             SSLerr(SSL_F_SSL_CTX_USE_SERVERINFO_FILE, SSL_R_PEM_NAME_TOO_SHORT);
             goto end;
         }
-        if (strncmp(name, namePrefix1, strlen(namePrefix1)) == 0) {
```
InfoHunter: Or you can replace `strlen()` into `sizeof` here to avoid introducing more local variables.

raniervf:
The declaration char namePrefix1[] = "SERVERINFO FOR ";, , leave doubts, if original author,
will introduce namePrefix mutable in future.

mspncp:
Joined with #10380.

mspncp:
> The declaration char namePrefix1[] = "SERVERINFO FOR ";, , leave doubts, if original author,
> will introduce namePrefix mutable in future.

That's why I would make it a `static const char namePrefix1[]`

ssl/ssl_rsa.c
richsalz: the variables should be "static"

PR #10385, opened 2019-11-08
Add EVP_PKEY_meth setter and getter for digestsign and digestverify

anhu:
Added EVP_PKEY_meth_set_digestsignverify() and EVP_PKEY_meth_get_digestsignverify()

Ran `make tests` and `make doc-nits`. 
Everything passed.


InfoHunter:
Branch 1.1.1 accepts only bug-fixes at current stage. This should go into master branch.

anhu:
Please note that I have two email addresses associated with this account.  anth_hu@homtail.com is the one that is on file for the CLA. Therefore, you can waive the cla-check failure.

anhu:
> Branch 1.1.1 accepts only bug-fixes at current stage. This should go into master branch.

oops...ok.  Will submit to master.

richsalz:
BUT ... if making things opaque means that code broke, then adding these kinds of things is treated as a bug.

anhu:
> BUT ... if making things opaque means that code broke, then adding these kinds of things is treated as a bug.

So in 1.0.2 there was no `digestsign` and `digestverify` and as such, nothing could be broken. So... I think this is some kind of weird edge case.  I'll keep it closed for now.

anhu:
By the way, I'm really impressed with the responsiveness here.  Thank you very much.

mattcaswell:
Looks like a missing accessor to me so could be eligible for 1.1.1.

mattcaswell:
> Please note that I have two email addresses associated with this account. anth_hu@homtail.com is the one that is on file for the CLA.

Please could you send an updated CLA listing *all* the email addresses you are likely to use for submissions? This helps us to keep our paper trail complete.

mspncp:
See #10388

PR #10386, opened 2019-11-08
Fix EVP_PKEY_asn1_find[_str]() to always look in engines

anhu:
Find an implementation of an ASN1 algorithm. Also search through engines.
If 'pe' is not NULL set *pe to a functional reference of the engine
implementing 'type' or NULL if no engine implements it.

Ran `make tests` and `make doc-nits`. 
Everything looked fine.


anhu:
Please note that I have two email addresses associated with this account.  anth_hu@homtail.com is the one that is on file for the CLA. Therefore, you can waive the cla-check failure.

InfoHunter:
Same as previous PR. New feature on master branch only :-)

InfoHunter:
Ohh, I am sorry, is this a bug fix or an enhancement?

InfoHunter:
Not sure if this could break the behavior of this API or not, since the engine-searching conditions are changed.

anhu:
This is an actual bugfix, IMHO.  I will leave this open and also submit to `master`.

anhu:
oh noes.  I didn't realize forced pushes would affect open PRs. Closing and resubmitting to master.

PR #10387, opened 2019-11-08
Fix EVP_PKEY_asn1_find() and EVP_PKEY_asn1_find_str() to always look …

anhu:
…in engines

Ran `make tests` and make `doc-nits` and everything was fine.

### DIFF
--- a/crypto/asn1/ameth_lib.c
+++ b/crypto/asn1/ameth_lib.c
@@ -79,6 +79,9 @@ static const EVP_PKEY_ASN1_METHOD *pkey_asn1_find(int type)
 const EVP_PKEY_ASN1_METHOD *EVP_PKEY_asn1_find(ENGINE **pe, int type)
 {
     const EVP_PKEY_ASN1_METHOD *t;
+#ifndef OPENSSL_NO_ENGINE
+    ENGINE *e;
+#endif
 
     for (;;) {
         t = pkey_asn1_find(type);
@@ -88,7 +91,6 @@ const EVP_PKEY_ASN1_METHOD *EVP_PKEY_asn1_find(ENGINE **pe, int type)
     }
 
 #ifndef OPENSSL_NO_ENGINE
-    ENGINE *e;
     /* type will contain the final unaliased type */
     e = ENGINE_get_pkey_asn1_meth_engine(type);
     if (e) {
@@ -107,12 +109,14 @@ const EVP_PKEY_ASN1_METHOD *EVP_PKEY_asn1_find_str(ENGINE **pe,
 {
     int i;
     const EVP_PKEY_ASN1_METHOD *ameth = NULL;
+#ifndef OPENSSL_NO_ENGINE
+    ENGINE *e;
+#endif
 
     if (len == -1)
         len = strlen(str);
 
 #ifndef OPENSSL_NO_ENGINE
-        ENGINE *e;
         ameth = ENGINE_pkey_asn1_find_str(&e, str, len);
         if (ameth) {
             /*
--- a/crypto/asn1/ameth_lib.c
+++ b/crypto/asn1/ameth_lib.c
@@ -71,8 +71,8 @@ static const EVP_PKEY_ASN1_METHOD *pkey_asn1_find(int type)
 }
 
 /*
- * Find an implementation of an ASN1 algorithm. If 'pe' is not NULL also
- * search through engines and set *pe to a functional reference to the engine
+ * Find an implementation of an ASN1 algorithm. Also search through engines.
+ * If 'pe' is not NULL set *pe to a functional reference of the engine
  * implementing 'type' or NULL if no engine implements it.
  */
 
@@ -86,18 +86,19 @@ const EVP_PKEY_ASN1_METHOD *EVP_PKEY_asn1_find(ENGINE **pe, int type)
             break;
         type = t->pkey_base_id;
     }
-    if (pe) {
+
 #ifndef OPENSSL_NO_ENGINE
-        ENGINE *e;
-        /* type will contain the final unaliased type */
-        e = ENGINE_get_pkey_asn1_meth_engine(type);
-        if (e) {
+    ENGINE *e;
+    /* type will contain the final unaliased type */
+    e = ENGINE_get_pkey_asn1_meth_engine(type);
+    if (e) {
+        if (pe)
             *pe = e;
-            return ENGINE_get_pkey_asn1_meth(e, type);
-        }
-#endif
-        *pe = NULL;
+        return ENGINE_get_pkey_asn1_meth(e, type);
     }
+#endif
+    if (pe)
+        *pe = NULL;
     return t;
 }
 
@@ -109,7 +110,7 @@ const EVP_PKEY_ASN1_METHOD *EVP_PKEY_asn1_find_str(ENGINE **pe,
 
     if (len == -1)
         len = strlen(str);
-    if (pe) {
+
 #ifndef OPENSSL_NO_ENGINE
         ENGINE *e;
         ameth = ENGINE_pkey_asn1_find_str(&e, str, len);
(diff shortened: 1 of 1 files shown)


anhu:
please note that I am anth_hu@hotmail.com and so you clear the CLA hold.

richsalz:
you can change the author on the commit (git commit --am --au .....) or perhaps send email to the CLA email address so that they can update the database.

anhu:
ugh...didn't notice that email address in the Author field. thank you.

anhu:
please note commit how has an author that is in your CLA list. please clear the flag.

levitte:
This is a semantic change that I'm not too sure about. What's the use case that got you to make this change?

anhu:
I'm building a version of openssl where algorithms are not implemented in openssl but in an engine so its critical that all calls to `EVP_PKEY_asn1_find()` looks for implementations in engines.  I know what you're thinking, just make sure all calls to `EVP_PKEY_asn1_find()` send in a tmp_eng ptr.  

But I think this change has much more value to the wider community because a developer might try to disable the openssl implementation of an algorithm and only use their own engine implementation.

richsalz:
@anhu -- the flag is done automatically, such as when you force-push :)

anhu:
> @anhu -- the flag is done automatically, such as when you force-push :)

Ah....thank you for letting me know.

anhu:
> This is a semantic change that I'm not so sure about. What's the use case that got you to make this change?

Also, I can't think of a use case where checking the engines would be a bad idea.

anhu:
All the following cases trigger not checking for engines: 
```
ssl/ssl_ciph.c:326:    ameth = EVP_PKEY_asn1_find_str(NULL, pkey_name, -1);
crypto/x509/x509_set.c:211:        const EVP_PKEY_ASN1_METHOD *ameth = EVP_PKEY_asn1_find(NULL, pknid);
crypto/x509/t_x509.c:318:            ameth = EVP_PKEY_asn1_find(NULL, pkey_nid);
crypto/pem/pem_lib.c:144:            ameth = EVP_PKEY_asn1_find_str(NULL, nm, slen);
crypto/store/loader_file.c:424:                && (ameth = EVP_PKEY_asn1_find_str(NULL, pem_name,
```
Why not future proof it?

levitte:
> Why not future proof it?

You may not have followed recent development...  ENGINE is considered legacy with upcoming v3.0, and a new concept called "provider" replaces it.  See the [recent blog post](https://www.openssl.org/blog/blog/2019/11/07/3.0-update/)...

anhu:
Oh, yes.  But from what I understand, the engine API will still remain for backwards compatibility, correct? As well, 1.1.1 is LTS so this change could potentially help the community until  2023-09-11.

mattcaswell:
> Oh, yes. But from what I understand, the engine API will still remain for backwards compatibility, correct?

It will remain for now but be deprecated. In some future version it will be removed. We will strongly encourage everyone to move to providers.

anhu:
Hi, 

Might I propose an alternative? All the following calls could have engine pointers that would do nothing.  Would that be a more preferable solution?  While I understand that engines are going out, I still they they will be around on the order of years...no? 

```
ssl/ssl_ciph.c:326:    ameth = EVP_PKEY_asn1_find_str(NULL, pkey_name, -1);
crypto/x509/x509_set.c:211:        const EVP_PKEY_ASN1_METHOD *ameth = EVP_PKEY_asn1_find(NULL, pknid);
crypto/x509/t_x509.c:318:            ameth = EVP_PKEY_asn1_find(NULL, pkey_nid);
crypto/pem/pem_lib.c:144:            ameth = EVP_PKEY_asn1_find_str(NULL, nm, slen);
crypto/store/loader_file.c:424:                && (ameth = EVP_PKEY_asn1_find_str(NULL, pem_name,
```

anhu:
Anyone have any objections? 

Upon closer inspection, this one:

```
ssl/ssl_ciph.c:326:    ameth = EVP_PKEY_asn1_find_str(NULL, pkey_name, -1);
```
...does not need modification because there is another version of that function that is implemented if engines are enabled. (coincidentally, that other version could be removed if this change were to be accepted. :) )

anhu:
> It will remain for now but be deprecated. In some future version it will be removed. 
I would assume not until 1.1.1 is EOL.  Is that a bad assumption?

anhu:
Hi, 
It seems that there isn't very much enthusiasm for this PR.  Should I close?

crypto/asn1/ameth_lib.c
```diff
@@ -71,33 +71,36 @@ static const EVP_PKEY_ASN1_METHOD *pkey_asn1_find(int type)
...
-    if (pe) {
+
 #ifndef OPENSSL_NO_ENGINE
-        ENGINE *e;
-        /* type will contain the final unaliased type */
-        e = ENGINE_get_pkey_asn1_meth_engine(type);
-        if (e) {
+    /* type will contain the final unaliased type */
+    e = ENGINE_get_pkey_asn1_meth_engine(type);
+    if (e) {
```
mspncp: Please note that our coding style mandates `if (e != NULL)` for NULL pointer checks.

_(I know it was this way already, but please fix it while you're at it)_

crypto/asn1/ameth_lib.c
```diff
@@ -71,33 +71,36 @@ static const EVP_PKEY_ASN1_METHOD *pkey_asn1_find(int type)
...
+        if (pe)
```
mspncp: same here

crypto/asn1/ameth_lib.c
```diff
@@ -71,33 +71,36 @@ static const EVP_PKEY_ASN1_METHOD *pkey_asn1_find(int type)
...
             *pe = e;
-            return ENGINE_get_pkey_asn1_meth(e, type);
-        }
+        return ENGINE_get_pkey_asn1_meth(e, type);
+    }
 #endif
+    if (pe)
```
mspncp: and here

crypto/asn1/ameth_lib.c
```diff
@@ -120,12 +125,17 @@ const EVP_PKEY_ASN1_METHOD *EVP_PKEY_asn1_find_str(ENGINE **pe,
             if (!ENGINE_init(e))
                 ameth = NULL;
             ENGINE_free(e);
-            *pe = e;
+            if (pe) {
```
mspncp: here

crypto/asn1/ameth_lib.c
```diff
@@ -120,12 +125,17 @@ const EVP_PKEY_ASN1_METHOD *EVP_PKEY_asn1_find_str(ENGINE **pe,
...
+                *pe = NULL;
+                if (ameth)
```
mspncp: here

crypto/asn1/ameth_lib.c
```diff
@@ -120,12 +125,17 @@ const EVP_PKEY_ASN1_METHOD *EVP_PKEY_asn1_find_str(ENGINE **pe,
...
+                    *pe = e;
+            }
             return ameth;
         }
 #endif
+    if (pe)
```
mspncp: here

crypto/asn1/ameth_lib.c
anhu: OH!  For sure!  `!= NULL` has always been my preference, but I put pre-existing style conformance as the top priority.  Will happily make these changes soon.

levitte:
This PR will simply not do, because it will leak ENGINE references.  It's all in this line:

https://github.com/openssl/openssl/blob/3478a2102e8bfc63f20e22826ab39b407ff43f62/crypto/asn1/ameth_lib.c#L93

`ENGINE_get_pkey_asn1_meth_engine()` returns not just an ENGINE pointer, but a functional reference, which means that this ENGINE instance has been initialized for this use.  At some point, someone must call `ENGINE_finish()` on this pointer, or that ENGINE's functional reference count will not be decremented.

This is the whole reason why it won't look at engines unless `pe` is non-NULL.

So, the better fix is actually to make sure all calls are with a non-NULL `pe`.

anhu:
Sorry for leaving this dangling for so long!  Will close and eventually implement "all calls are with a non-NULL pe"

PR #10389, opened 2019-11-08
Rename providers/common/provlib.c to nid_to_name.c

levitte:
It contains only one function, which should only get added to non-FIPS
providers.

### DIFF
--- a/providers/common/build.info
+++ b/providers/common/build.info
@@ -1,6 +1,6 @@
 SUBDIRS=digests ciphers
 
-SOURCE[../libcommon.a]=provider_err.c provlib.c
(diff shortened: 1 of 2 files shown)


levitte:
A name like `provlib.c` kind of encourages you to add more functions in that file, and if those functions are used by the FIPS module, the whole object file gets sucked in, and there will be a clash with the FIPS module's implementation of `ossl_prov_util_nid_to_name`

paulidale (approved):

levitte:
Merged.

a8f6d2642d1b011961022a5d9aa9e1156097b85c Rename providers/common/provlib.c to nid_to_name.c

Issue #10360, opened 2019-11-04
FIPS provider TLS tests


paulidale:
The FIPS provider needs to be tested against TLS in most or all of the CI builds.

This requires proper interconnection of libssl and providers via properties and library contexts.

paulidale:
We do TLS tests using the FIPS provider.  Closing this.

PR #10394, opened 2019-11-09
SERIALIZER: New API for serialization of objects through providers

levitte:
Serialization is needed to be able to take a provider object (such as
the provider side key data) and output it in PEM form, DER form, text
form (for display), and possibly other future forms (XML? JSON? JWK?)

The idea is that a serializer should be able to handle objects it has
intimate knowledge of, as well as object data in OSSL_PARAM form.  The
latter will allow libcrypto to serialize some object with a different
provider than the one holding the data, if exporting of that data is
allowed and there is a serializer that can handle it.

We will provide serializers for the types of objects we know about,
which should be useful together with any other provider that provides
implementations of the same type of object.

Serializers are selected by method name and a couple of additional
properties:

- format
                used to tell what format the output should be in.
                Possibilities could include "format=text",
                "format=pem", "format=der", "format=pem-pkcs1"
                (traditional), "format=der-pkcs1" (traditional)
- type
                used to tell exactly what type of data should be
                output, for example "type=public" (the public part of
                a key), "type=private" (the private part of a key),
                "type=domainparams" (domain parameters).

##### Checklist


levitte:
## Background

when implementing key and domparam generation (#10289), I realised that we couldn't do much with the generated keys, apart from signing, verifying, that sort of thing.  But no way whatsoever to save them.

Some kind of solution was needed, and after a conversation with @mattcaswell, we came to the conclusion that the smoothest would be to provide serializers, and to allow the serializers to handle both provider side objects directly (they have to come from the same provider as the keymgmt that manages the key itself) as well as key data in OSSL_PARAM array form (which allows a serializer to handle anything it recognises coming from any other provider, as long as that provider allows export of said data).

## Why not an EVP API?

Serializers may be extended some day to handle any type of data, such as X.509, which isn't an EVP layer object.  Therefore, serializers need to be a higher level class.

levitte:
If anyone wonders, the mechanism this offers would allow us to extract the data from a key made in the FIPS provider module, and serialize it using serializers in the default provider.  This is by design.

levitte:
This is no longer WIP, so can take reviews

doc/man7/provider-serializer.pod
```diff
@@ -0,0 +1,180 @@
...
+ /*
+  * None of these are actual functions, but are displayed like this for
+  * the function signatures for functions that are offered as function
+  * pointers in OSSL_DISPATCH arrays.
+  */
+
+ /* Functions to serialize object data */
+ int OP_serializer_serialize_data(void *provctx, const OSSL_PARAM *data,
+                                  BIO *out);
+ int OP_serializer_serialize_object(void *provobj, BIO *out);
```
mattcaswell: I notice this is write serialization only. Are you planning read serialization in some later PR?

doc/man7/provider-serializer.pod
```diff
@@ -0,0 +1,180 @@
...
+
+The SERIALIZER is a generic method to serialize any set of object data
+in L<OSSL_PARAM(3)> array form, or any provider side object into
+serialized form, and write it to the given BIO.  If the caller wants
+to get the serialized stream to memory, it should provide a
+L<BIO_s_membuf(3)>.
+
+=begin comment
+
+Future development will also include deserializing functions.
```
mattcaswell: Ah!!

doc/man7/provider-serializer.pod
mattcaswell: It occurs that we might want serialization to something other than a BIO, such that the serialization is entirely internal to the provider so that (for example) private keys never leave the provider boundary. In such a case we might want some interface along the lines of "serialize this key and give me back some identifier that I can use to read it back later". Another PR though. We might need to think about the API a little bit. In some cases I imagine the "id" is provider determined (e.g. sequence number from a provider managed database), but in others the id might be client determined (e.g. key name is in some provider managed filesystem).

doc/man7/provider-serializer.pod
```diff
@@ -0,0 +1,180 @@
...
+ OP_serializer_serialize_object    OSSL_FUNC_SERIALIZER_SERIALIZE_OBJECT
+
+=head2 Names and properties
+
+The name of an implementation should match the type of object it
+handles.  For example, an implementation that serializes an RSA key
+should be named accordingly.
+
+To be able to specify exactly what serialization format and what type
+out data a serializer implementation is expected to handle, two
```
mattcaswell: Typo: type *of* data

doc/man7/provider-serializer.pod
```diff
@@ -0,0 +1,180 @@
...
+This property is used to specify what kind of output format the
+implementation produces.  Currently known formats are:
+
+=over 4
+
+=item text
+
+An implementation with that format property value outputs human
+readable text, making that implementation suitable for C<-text> output
+in diverse L<openssl(1)> commands.
```
mattcaswell: How does indentation work in this case?

doc/man7/provider-serializer.pod
```diff
@@ -0,0 +1,180 @@
...
+=item der
+
+An implementation with that format property value outputs DER
+formatted data.
+
+=back
+
+=item type
+
+With object that have multiple purposes, this can be used to specify
```
mattcaswell: Typo: objects

doc/man7/provider-serializer.pod
```diff
@@ -0,0 +1,180 @@
...
+know anything about.
+
+=head2 Serializing functions
+
+=for comment There will be a "Deserializing functions" title as well
+
+OP_serializer_serialize_data() should take an array of B<OSSL_PARAM>,
+I<data>, and if it contains the data necessary for the object type
+that the implementation handles, it should output the object in
+serialized form the the B<BIO>.
```
mattcaswell: Typo: the the

doc/man7/provider-serializer.pod
```diff
@@ -0,0 +1,180 @@
...
+All these "functions" have a corresponding function type definition
+named B<OSSL_{name}_fn>, and a helper function to retrieve the
+function pointer from a B<OSSL_DISPATCH> element named
+B<OSSL_get_{name}>.
+For example, the "function" OP_keymgmt_importdomparams() has these:
+
+ typedef int
+     (OSSL_OP_serializer_serialize_data_fn)(void *provctx,
+                                            const OSSL_PARAM params[],
+                                            BIO *out);
```
mattcaswell: I do wonder why we even need this one. The same effect can be achieved by using the keymgmt "import" capability - and then use the "serialize_object" function. It seems a little strange to have two ways of achieving this.

providers/common/bio_prov.c
```diff
@@ -0,0 +1,96 @@
...
+    for (; fns->function_id != 0; fns++) {
+        switch (fns->function_id) {
+        case OSSL_FUNC_BIO_NEW_FILE:
+            if (c_bio_new_file == NULL)
+                c_bio_new_file = OSSL_get_BIO_new_file(fns);
+            break;
+        case OSSL_FUNC_BIO_NEW_MEMBUF:
+            if (c_bio_new_membuf == NULL)
+                c_bio_new_membuf = OSSL_get_BIO_new_membuf(fns);
+            break;
```
mattcaswell: I am slightly perplexed by the above two functions. Why does a provider need to be able to *create* a BIO? Doesn't it just use BIOs that are passed to it?

providers/implementations/serializers/rsa_print.c
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    bytes = BN_num_bytes(n);
+    words = bn_get_words(n);
+    neg = BN_is_negative(n) ? "-" : "";
+
+    if (BN_is_zero(n))
+        return ossl_prov_bio_printf(out, "%s%s0\n", label, post_label_spc);
+
+#if 0
```
mattcaswell: Why?

providers/implementations/serializers/rsa_print.c
```diff
@@ -0,0 +1,273 @@
...
+    if (BN_num_bytes(n) <= BN_BYTES)
+        return ossl_prov_bio_printf(out,
+                                    "%s%s%s" BN_FMTu " (%s0x" BN_FMTx ")\n",
+                                    label, post_label_spc, neg, words[0],
+                                    neg, words[0]);
+#endif
+
+    if (neg[0] == '-')
+        neg = " (Negative)";
```
mattcaswell: RSA values are never negative. Or do you plan for this to become a more generic function? In which case it doesn't belong in this file.

mattcaswell (changes requested):

crypto/core_namemap.c
```diff
@@ -251,10 +251,74 @@ int ossl_namemap_add_n(OSSL_NAMEMAP *namemap, int number,
...
+    return ossl_namemap_add_name_n(namemap, number, name, strlen(name));
 }
+
+int ossl_namemap_add_names(OSSL_NAMEMAP *namemap, int number,
+                           const char *names, const char separator)
+{
+    const char *p, *q;
+    size_t l;
+
+    /* Check that we have a namemap and that there is at least one name */
```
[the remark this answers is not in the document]
levitte: It's a remainder from...  I think it's from the time I was going with an array of strings.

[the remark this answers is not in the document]
levitte: I'll remove that part of the comment

crypto/core_namemap.c
```diff
@@ -251,10 +251,74 @@ int ossl_namemap_add_n(OSSL_NAMEMAP *namemap, int number,
...
+    for (p = names; *p != '\0'; p = (q == NULL ? p + l : q + 1)) {
+        int this_number;
+
+        if ((q = strchr(p, separator)) == NULL)
+            l = strlen(p);       /* offset to \0 */
+        else
+            l = q - p;           /* offset to the next separator */
+
+        this_number = ossl_namemap_add_name_n(namemap, number, p, l);
+        if (number == 0)
```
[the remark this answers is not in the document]
levitte: Done

crypto/core_namemap.c
```diff
@@ -251,10 +251,74 @@ int ossl_namemap_add_n(OSSL_NAMEMAP *namemap, int number,
...
+        else
+            l = q - p;           /* offset to the next separator */
+
+        this_number = ossl_namemap_name2num_n(namemap, p, l);
+
+        if (*p == '\0' || *p == separator) {
+            ERR_raise(ERR_LIB_CRYPTO, CRYPTO_R_BAD_ALGORITHM_NAME);
+            return 0;
+        }
+        if (number == 0)
```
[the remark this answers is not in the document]
levitte: Done

crypto/serializer/serializer_lib.c
```diff
@@ -0,0 +1,154 @@
...
+{
+    struct serializer_write_data_st *d = arg;
+    void *provctx =
+        ossl_provider_ctx(OSSL_SERIALIZER_provider(d->serializer));
+
+    return d->serializer->serialize_data(provctx, params, d->out);
+}
+
+/*
+ * OSSL_SERIALIZER_EVP_PKEY_to_bio() returns -2 if it can't find a suitable
```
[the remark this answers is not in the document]
levitte: I'm used to -2 meaning that the call is unsupported, that's a value that has been injected in my brain for a long time with things OpenSSL.  If you grep for it in existing documentation, you'll see that it's fairly pervasive, at least there (I noticed just now that the documentation for `EVP_PKEY_print_{public,private,params}` says they return that value, but they don't in actual fact do so).

So a name.  Do you have a suggestion?  We do have `EVP_CTRL_RET_UNSUPPORTED`, but these functions aren't ctrls, so it seems like an inappropriate name.

The reason for -2 is, if I remember history right, that there are functions where 0 is a valid success value (for example returned length, and zero length is valid), so -1 was made to mean error, and when an indicator that a call isn't supported (i.e. the caller should try a fallback), the natural choice came to -2.

include/openssl/pem.h
```diff
@@ -143,6 +144,60 @@ int PEM_write_bio_##name(BIO *bp, type *x, const EVP_CIPHER *enc, \
...
+        const char *pq = OSSL_SERIALIZER_##asn1##_TO_PEM_PQ;            \
+        int wd = OSSL_SERIALIZER_##asn1##_WD;                    \
+        int ret = OSSL_SERIALIZER_##type##_to_bio(bp, x, pq, wd);       \
+        if (ret != -2)                                                  \
+            return ret;                                                 \
+        return PEM_ASN1_write_bio((i2d_of_void *)i2d_##asn1,str,bp,     \
+                                  (void *)x, NULL,NULL,0,NULL,NULL);    \
+    }
+
+# if 0                           /* Yet to be implemented */
```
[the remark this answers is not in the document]
levitte: Let me have a think on that.

crypto/serializer/serializer_lib.c
```diff
@@ -0,0 +1,154 @@
...
+ * serializer.  This allows any caller to detect this and to use fallback
+ * methods, as opposed to when 0 is returned to signify that there was a
+ * fatal error.
+ */
+int OSSL_SERIALIZER_EVP_PKEY_to_bio(BIO *out, const EVP_PKEY *pkey,
+                                    const char *propquery, int want_domparams)
+{
+    EVP_KEYMGMT *keymgmt = pkey->pkeys[0].keymgmt;
```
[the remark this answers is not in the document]
levitte: No...  the serializers can be completely independent.  More on that below.

doc/man3/OSSL_SERIALIZER_EVP_PKEY_to_bio.pod
```diff
@@ -0,0 +1,63 @@
...
+
+OSSL_SERIALIZER_EVP_PKEY_to_fp() does the same thing as
+OSSL_SERIALIZER_EVP_PKEY_to_bio(), except the output is written to the
+B<FILE> I<fp>.
+
+=head1 NOTES
+
+These function is usually not used directly, it's primary purpose is
+to support PEM and DER serialers implemented via macros such as
+IMPLEMENT_PEM_write().
```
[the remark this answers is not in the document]
levitte: Good point.  I'll rephrase that.

doc/man7/provider-serializer.pod
levitte: I think you were thinking of deserialization (what you've called "read serialization"), right?

This has indeed occurred to me regarding how `OSSL_STORE` is supposed to works, and only for loaders that find data in serialized form.

As for a key identity, did you notice the `loadkey` function in the KEYMGMT downcalls?  It's already on my mind and will play a fairly huge part in the provider portion of `OSSL_STORE`

doc/man7/provider-serializer.pod
levitte: Yup

doc/man7/provider-serializer.pod
levitte: This, I have another plan for.  Basically, I'm thinking that the serializer shouldn't need to know, that it should be dealt with a BIO filter, which can be pushed on, even just temporarly, by libcrypto before calling the serializer.

doc/man7/provider-serializer.pod
levitte: Fixed

doc/man7/provider-serializer.pod
levitte: Fixed

doc/man7/provider-serializer.pod
levitte: The serializers are supposed to be able to be independent, i.e. not necessarily need to be coupled with a KEYMGMT.  Otherwise, it essentially means that at least one provider that supplies KEYMGMT operations for a certain key type *must* also supply corresponding serializers.

providers/common/bio_prov.c
levitte: I have *no* idea, I simply added support for the BIO upcalls that are already there.  The only upcall *I* have added is the `BIO_vprintf` one.

providers/implementations/serializers/rsa_print.c
levitte: Ah!  Good catch!  The test in `test/evp_pkey_provided_test.c` makes a very small (32-bit) RSA key, and I wanted to check that the print of larger numbers worked correctly, so I disabled the small number printout...  if was meant to be temporary, of course, but I obviously forgot.  So thank you for notifying

levitte: Fixed

providers/implementations/serializers/rsa_print.c
levitte: Granted, I basically copied what is done downcall from `ASN1_bn_print`.  Are you *absolutely* sure we will never have a negative large number?  Not even in parameters such as PSS, or any other future set of params we don't know about?

levitte:
I just realised that I do have DER serializers, but haven't hooked them in anywhere.  So er, not as done as I thought...

levitte:
(don't let that stop anyone from looking at what's already here, though)

CHANGES
```diff
@@ -9,6 +9,25 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) Introduced a new method type and API, OSSL_SERIALIZER, to
+     represent generic serializers.  An implementation is expected to
+     be able to serialize an object associate with a given name (such
```
slontis: associated

CHANGES
```diff
@@ -9,6 +9,25 @@
...
+     as an algorithm name for an asymmetric key) into forms given by
+     implementation properties.
+
+     Serializers are primarily used from inside libcrypto, through
+     calls to functions like EVP_PKEY_print_private(),
+     PEM_write_bio_PrivateKey() and similar.
+
+     Serializers are specified in such a way that they can be made to
+     directly handle provider side portion of an object, if this
```
slontis: portions

CHANGES
```diff
@@ -9,6 +9,25 @@
...
+     provider side part comes from the same provider as the serializer
+     itself, but can also be made to handle objects in parametrized
```
slontis: parameterized

CHANGES
```diff
@@ -9,6 +9,25 @@
...
+     form (in form of an OSSL_PARAM array of data).  This allows a
```
slontis: in the form

crypto/evp/p_lib.c
```diff
@@ -666,6 +666,13 @@ static int unsup_alg(BIO *out, const EVP_PKEY *pkey, int indent,
 int EVP_PKEY_print_public(BIO *out, const EVP_PKEY *pkey,
                           int indent, ASN1_PCTX *pctx)
 {
+    const char *pq = OSSL_SERIALIZER_PUBKEY_TO_TEXT_PQ;
```
slontis: What is the meaning of _PQ and _WD?

crypto/serializer/serializer_lib.c
```diff
@@ -0,0 +1,154 @@
...
+ * Support for OSSL_SERIALIZER_TYPE_to_bio:
+ * finding a suitable serializer
+ */
+
+struct selected_serializer_st {
+    OPENSSL_CTX *libctx;
+    const OSSL_PROVIDER *desired_provider;
+    const char *propquery;
+    /*
+     * Serializers offer two functions, one that handles object data in
```
slontis: in the

crypto/evp/p_lib.c
slontis: Arrg.. propertyquery and wantdomainparams - maybe use longer names?

crypto/serializer/serializer_meth.c
```diff
@@ -0,0 +1,358 @@
...
+    struct serializer_do_all_data_st data;
+
+    data.user_fn = (void (*)(void *, void *))fn;
+    data.user_arg = arg;
+    ossl_algorithm_do_all(libctx, OSSL_OP_SERIALIZER, NULL,
+                          serializer_do_one, &data);
+}
+
+void OSSL_SERIALIZER_names_do_all(const OSSL_SERIALIZER *serializer,
+                                  void (*fn)(const char *name, void *data),
```
slontis: maybe just call one of these data fields something else (like dat) OR typedef the callback and just supply that..

doc/internal/man3/ossl_namemap_new.pod
```diff
@@ -55,25 +58,30 @@ names already associated with that number.
...
-
 ossl_namemap_doall_names() walks through all names associated with
 I<number> in the given I<namemap> and calls the function I<fn> for
 each of them.
 I<fn> is also passed the I<data> argument, which allows any caller to
 pass extra data for that function to use.
 
+ossl_namemap_add_names() divides up a set of names given in I<names>,
+separated by I<separator>, and adds each to the I<namemap>, all with
+the same number.  If some of them already exist in the I<namemap>,
```
slontis: I probably would have used identifier instead of number throughout the doc.

doc/man3/OSSL_SERIALIZER.pod
```diff
@@ -0,0 +1,120 @@
...
+                                      void (*fn)(OSSL_SERIALIZER *serializer,
+                                                 void *arg),
+                                      void *arg);
+ void OSSL_SERIALIZER_names_do_all(const OSSL_SERIALIZER *serializer,
+                                   void (*fn)(const char *name, void *data),
+                                   void *data);
+
+=head1 DESCRIPTION
+
+=for comment Future developmen should also talk about deserialization
```
slontis: development

doc/man3/OSSL_SERIALIZER_EVP_PKEY_to_bio.pod
```diff
@@ -0,0 +1,63 @@
...
+                                     const char *propquery,
+                                     int want_domparams);
+ int OSSL_SERIALIZER_EVP_PKEY_to_fp(FILE *fp, const EVP_PKEY *pkey,
+                                    const char *propquery,
+                                    int want_domparams);
+
+=head1 SYNOPSIS
+
+OSSL_SERIALIZER_EVP_PKEY_to_bio() finds the L<OSSL_SERIALIZER(3)> that
+matches the assymetric key I<pkey> and the desired properties
```
slontis: asymmetric

crypto/serializer/serializer_lib.c
```diff
@@ -0,0 +1,154 @@
...
+ * serializer.  This allows any caller to detect this and to use fallback
+ * methods, as opposed to when 0 is returned to signify that there was a
+ * fatal error.
+ */
+int OSSL_SERIALIZER_EVP_PKEY_to_bio(BIO *out, const EVP_PKEY *pkey,
+                                    const char *propquery, int want_domparams)
+{
+    EVP_KEYMGMT *keymgmt = pkey->pkeys[0].keymgmt;
+
+    if (keymgmt != NULL && pkey->pkeys[0].domainparams == want_domparams) {
```
slontis: Hmm the want_domparams is an interesting check.. What if the key had domain params but we dont care that want_domparams = 0. Currently it returns -2?

CHANGES
richsalz: or perhaps "as"

doc/man7/provider-serializer.pod
```diff
@@ -0,0 +1,180 @@
...
+know anything about.
+
+=head2 Serializing functions
+
+=for comment There will be a "Deserializing functions" title as well
+
+OP_serializer_serialize_data() should take an array of B<OSSL_PARAM>,
+I<data>, and if it contains the data necessary for the object type
+that the implementation handles, it should output the object in
+serialized form the B<BIO>.
```
slontis: form for (of?) the

richsalz: doesn't it serialize *to* the BIO ?

richsalz: And isn't the operation really "written to the BIO" ?

providers/common/bio_prov.c
slontis: Needed for FIPS integrity checks

crypto/evp/p_lib.c
```diff
@@ -684,8 +698,16 @@ int EVP_PKEY_print_private(BIO *out, const EVP_PKEY *pkey,
...
 {
+    const char *pq = OSSL_SERIALIZER_Parameters_TO_TEXT_PQ;
+    int wd = OSSL_SERIALIZER_Parameters_WD;
+    int ret = OSSL_SERIALIZER_EVP_PKEY_to_bio(out, pkey, pq, wd);
+
+    if (ret != -2)
+        return ret;
+
     if (pkey->ameth && pkey->ameth->param_print)
         return pkey->ameth->param_print(out, pkey, indent, pctx);
```
slontis: Is this line correct?

crypto/evp/p_lib.c
levitte: Right, because they aren't long enough already. I was hitting the 80 char limit in ugly ways in other places.

crypto/evp/p_lib.c
levitte: Er... Why wouldn't it be?

doc/internal/man3/ossl_namemap_new.pod
levitte: This was chosen so we wouldn't confuse them with (constant) NIDs

crypto/serializer/serializer_meth.c
levitte: Not sure I understand why "dat" is better than "data". Typedeffing could be something, do you have a suitable name?

crypto/evp/p_lib.c
slontis: I was thinking that the new code above it effectively does the same thing - guess I am wrong :)

levitte: These lines are conserved for legacy code.  The new code *only* deals with keys in providers.

crypto/serializer/serializer_lib.c
levitte: The way we (well, I) have structured it conceptually in providers make domain parameters a separate thing from keys.  Sure, our *implementation* uses the same structure for both, but still...

So yeah, if this particular `EVP_PKEY` holds domain parameters, and someone wants to get a key out of it, signalling that it's unsupported is the right thing to do...  I think.  Or should that be an error, i.e. return 0?

(the way the EVP API works, there actually is a separation between domain parameters and keys, even though subtle...  the separation visavi providers makes it more concrete)

crypto/evp/p_lib.c
levitte: ...  so I guess I should add a TODO comment...

levitte:
I've updated this branch to depend on #10412 and #10414

crypto/serializer/serializer_meth.c
richsalz: Well if you want to go all patterns, then "vistor_fn" as the function typedef?  Or like the rest of openssl, serializer_cb maybe?

crypto/serializer/serializer_lib.c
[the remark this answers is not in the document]
levitte: I'm preparing another PR where -2 gets a macro `OSSL_RET_UNSUPPORTED`.
It turns out, btw, that `EVP_CTRL_RET_UNSUPPORTED` was incorrect...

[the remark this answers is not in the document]
levitte: See #10431

include/openssl/pem.h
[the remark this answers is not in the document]
levitte: I'll add password callback support

doc/man3/OSSL_SERIALIZER_EVP_PKEY_to_bio.pod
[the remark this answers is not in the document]
levitte: Done

providers/implementations/serializers/rsa_print.c
levitte: From a quick read `crypto/rsa/rsa_ameth.c`, it seems that the PSS saltlen can have negative values...  seeing that this comes in form of a `ASN1_INTEGER`, which translates into an BIGNUM, I think it may be wise to keep the negative number support...

levitte:
I've tinkered most of the day with this, to try and have encrypted serializations, and came to realize that I need to redesign the libcrypto\<-\>provider interface and make it more like the others, i.e. functionauty to create context, set diverse params, to the actual operation, free context.

richsalz:
The idea of changing the internal interface now scares me.  There are work-arounds, serialize to a BIO and then encrypt it.  Please don't rip up things now.

levitte:
> The idea of changing the internal interface now scares me.

This is *new* interface.  The issue I came to is that hooking in the PEM cb writers became insanely clumsy, to the point that the PEM stuff got seriously messed up.  You wanna see scary, then you should have seen that...  but I'm not showing.

> There are work-arounds, serialize to a BIO and then encrypt it. Please don't rip up things now.

Huuuuhhhhhh?????  No, you want to encrypt it *before* serializing.  Think PKCS#8.  Think "traditional" encrypted PEM.

richsalz:
Yes, I know it's a new interface.  Does it have to be fixed *now*  Look at what does have to happen for this release and assess the priority of ripping up what' been done so far.
I look at the list of open issues for FIPS and to me the answer is very clearly no.  But perhaps you can come to a different answer.

levitte:
We definitely want it now.

Before starting on this, I was looking at the prospect of putting a lot of ugly legacy hackery in to be able to get out der and pem from provider backed keys. Just thinking about it was heart wrenching, and it was highly probable that we'd get stuck, *and* give no way whatsoever for other types of keys (GOST keys, for example) to get their keys serialized. The result would most likely be a pile of shit that we'd want to replace as fast as we can. 

Plus, you get to say in the keygen PR that it's very very cool...

richsalz:
I am responding to this comment, https://github.com/openssl/openssl/pull/10394#issuecomment-554116061, and ONLY to that comment.

Overall, serializing is very cool and yes, we need it do do things like export key material.  Do we need to export *encrypted* key material right now?  More than we need, say #10360 ?  Clearly not.

levitte:
Er... So there is this command, `openssl genpkey -algorithm rsa -aes-256-cbc`. Without encryption support in the serializer, that will simply stop working, or be seriously crippled at best.

levitte:
Also, pitting different efforts against each other is not cool.

richsalz:
I am just saying that other things are more important right now, and I showed one example of many. Come back to it later. If the worst thing happens, it's a post-3.0 fix.

I am not pitting anyone's efforts against anyone else's.  And, I don't know why you say that.  (There are plenty of reasons why I am not cool, but that's not one of them :)  I am just trying to encourage the appropriate priotization so that the release doesn't slip again.

levitte:
There, reworked, and I think it was for the better.

crypto/serializer/serializer_lib.c
levitte: I've reworked this to look at the property query string and figure it out from there.

@paulidale, what I've done is kind of a hack, and I wonder if there's a better way to check if some specific property was queried using your property parser.

levitte:
I recommend looking at this PR a commit at a time.  I've tried to structure the commit sequence in a nice and comprehensible way.

crypto/serializer/serializer_lib.c
paulidale: I think something will need to be added to the property code to extract values.

crypto/serializer/serializer_pkey.c
```diff
@@ -0,0 +1,361 @@
...
+
+    if ((s = strstr(propquery, "type")) == NULL)
+        goto parse_end;
+    if (!(s == propquery || s[-1] == ',' || s[-1] == ' '))
+        goto parse_end;
+    for (s += 4; s[0] == ' '; s++)
+        continue;
+    if (s[0] != '=')
+        goto parse_end;
+    for (s += 4; s[0] == ' '; s++)
```
paulidale: +4 ??

This needs to be wrapped into the parser, there are too many missing cases.  For example:
```
type="domainparams"
type='domainparams'
? type = domainparams
```
are all legal (the latter is a bit nonsensical in context).

Is making a `type=domainparams` query against the fetched serialiser's properties a possibility?

levitte: Argh, copy-n-paste from line 313...

The kind of check you're talking about has an interesting property...  it means that I'll have to take `pkey->pkeys[0].domainparams` on faith, but then use its value to limit what serializers are considered.  Yeah, sure, that can be done.

Thanks, I might even figure out on my own  how to do that (although, giving me a pointer at the function to call might be helpful).

paulidale: Still pass the original query through which should do the value limiting.  Just make a second query solely about the type to figure out what.

It means exposing the property parse/query internals a bit and keeping the property definitions associated with algorithms in an accessible manner.

levitte: Ok, I've reworked this as you suggested.  It's not very ugly at all!

crypto/serializer/serializer_lib.c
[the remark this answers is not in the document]
levitte: At this point, a -2 return here is superfluous, as there are other ways to see if the operation is unsupported, so I'm thinking I'll convert this to a genuine error, i.e. if the caller has come this far and hasn't checked earlier if the operation is supported (by checking if the context has an associated serializer), then an error is well deserved :wink:

include/openssl/pem.h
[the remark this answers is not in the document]
levitte: Pass phrase callback support worked in, which resolves this.

CHANGES

CHANGES
levitte: I actually think singular is fitting.  It's the provider side portion of *an object*, and from libcrypto's point of view, each object only has one portion.

levitte: I've added "the", though, making it "directly handle the provider side portion of an object, ...", maybe that makes it clearer?

CHANGES
levitte: It looks like both variants are acceptable: https://en.wiktionary.org/wiki/parametrise#English

CHANGES
levitte: I like @richsalz's suggestion, so changing to that

crypto/serializer/serializer_lib.c
levitte: Fixed

crypto/serializer/serializer_meth.c
levitte: This sounds like a subkect for a separate PR

doc/man3/OSSL_SERIALIZER.pod
levitte: Fixed

doc/man3/OSSL_SERIALIZER_EVP_PKEY_to_bio.pod
levitte: I've rewritten much of the docs, and this seems to be gone now

doc/man7/provider-serializer.pod
levitte: Adding *to* seems like the right move.  Done

levitte:
Interesting...  in travis, `test/evp_pkey_provided_test` segfaults.  In my local environment, it works fine.  Oh boy, that's going to be interesting...

levitte:
Something we might want to consider is having the serializers in a separate provider.  We do want them to work together with the FIPS provider, but without *being part* of the FIPS module, and yet not be forced to load the default module...

mspncp:
The amount of fixups in this pull request is really mind-blowing. How about doing an autosquash _without moving_ the merge-base?

```
$ git rebase -i --autosquash $(git merge-base HEAD master)
```

Such a change could be force-pushed without requiring a restart of the review process, because the tree remains (provable) unchanged. And it would simplify reviewing the parts.

mspncp:
P.S: I just learned from the manual pages that git supports a simpler and more elegant way to rebase without moving the merge-base:


```
$ git rebase -i --autosquash master --onto master...
```


**GIT-REBASE(1)**

```
       --onto <newbase>
           Starting point at which to create the new commits. If the --onto option is not specified, the starting point is <upstream>. May be any valid commit, and not
           just an existing branch name.

           As a special case, you may use "A...B" as a shortcut for the merge base of A and B if there is exactly one merge base. You can leave out at most one of A and B,
           in which case it defaults to HEAD.
```

levitte:
I haven't moved the merge-base for a while.  I *have* moved around the fixup commits so they are directly following the commit they fix, to catch any conflict that might get worse further on.

However, I'm done re-arranging things and expect all coming fixups to just be appended.

levitte:
On the other hand, since this PR has been reworked entirely over the weekend, perhaps I should just squash all current fixups, get us restarted on the cleaner slate?

mspncp:
> On the other hand, since this PR has been reworked entirely over the weekend, perhaps I should just squash all current fixups, get us restarted on the cleaner slate?

Makes sense to me. The latest I would do it when it goes out of WIP.

crypto/serializer/serializer_pkey.c
paulidale: Looks neat.

levitte:
> The latest I would do it when it goes out of WIP.

Fair enough.  Right now, I'm trying to beat some sense into the CIs (or well, some parts of the code, really :wink:)

levitte:
Squashed to a clean slate, as discussed with @mspncp above.  I believe I've addressed all concerns, and fixed all the CI failures, so taking this out of WIP.

Let reviewing begin!

levitte:
Something to be noted is that the implemented serializers have room for improvement, which depends on other parts.

RSA-PSS keys are currently not implemented in the RSA KEYMGMT implementation, so I haven't been able to verify that the serialization works properly for them, and quite frankly, I'm not sure how to deal with that sort of keys, 'cause in the legacy code, it's implemented as kinda sorta a separate type of RSA and kinda sorta a normal RSA with added domain parameters, and I haven't been able to decide what to decide for the provided implementation.  *[help me figure it out, please]*

Likewise for X9.42 DH (`EVP_PKEY_DHX`), we don't currently support that type of DH key in the DH KEYMGMT implementation as far as I know.

levitte:
Regarding the room for improvement I mentioned, I expect that to happen in follow-up PRs

mattcaswell:
> . We do want them to work together with the FIPS provider, but without being part of the FIPS module, and yet not be forced to load the default module...

This makes sense to me. They could still be tagged with the "default" property even if they are in another provider.

levitte:
> > . We do want them to work together with the FIPS provider, but without being part of the FIPS module, and yet not be forced to load the default module...
> 
> This makes sense to me. They could still be tagged with the "default" property even if they are in another provider.

I would rather make that change in another PR. This one is loaded enough as it is.

mattcaswell:
> I would rather make that change in another PR. This one is loaded enough as it is.

No objection to that.

levitte:
Ping

providers/implementations/serializers/serializer_dh_priv.c
```diff
@@ -0,0 +1,240 @@
...
+
+static int dh_priv_set_ctx_params(void *vctx, const OSSL_PARAM params[])
+{
+    struct dh_priv_ctx_st *ctx = vctx;
+    const OSSL_PARAM *p;
+
+    if ((p = OSSL_PARAM_locate_const(params, OSSL_SERIALIZER_PARAM_CIPHER))
+        != NULL) {
+        const OSSL_PARAM *propsp =
+            OSSL_PARAM_locate_const(params, OSSL_SERIALIZER_PARAM_PROPERTIES);
```
mattcaswell: Does this imply that the cipher and the properties *must* be passed in the same set_ctx_params call? i.e. you can't set one and then the other. Is that desirable? Should it be documented?

providers/implementations/serializers/serializer_dsa_priv.c
```diff
@@ -0,0 +1,234 @@
...
+static int dsa_priv_set_ctx_params(void *vctx, const OSSL_PARAM params[])
+{
+    struct dsa_priv_ctx_st *ctx = vctx;
+    const OSSL_PARAM *p;
+
+    if ((p = OSSL_PARAM_locate_const(params, OSSL_SERIALIZER_PARAM_CIPHER))
+        != NULL) {
+        const OSSL_PARAM *propsp =
+            OSSL_PARAM_locate_const(params, OSSL_SERIALIZER_PARAM_PROPERTIES);
+        const char *props = NULL;
```
mattcaswell: Same comment as before applies here

util/missingcrypto.txt
```diff
@@ -455,6 +455,7 @@ ERR_load_PKCS12_strings
 ERR_load_PKCS7_strings
 ERR_load_RAND_strings
 ERR_load_RSA_strings
+ERR_load_OSSL_SERIALIZER_strings
```
mattcaswell: Why?

In fact why does this need to be public?

mattcaswell (changes requested):

providers/common/bio_prov.c
```diff
@@ -0,0 +1,96 @@
...
+            if (c_bio_vprintf == NULL)
+                c_bio_vprintf = OSSL_get_BIO_vprintf(fns);
+            break;
+        }
+    }
+
+    return 1;
+}
+
+BIO *ossl_prov_bio_new_file(const char *filename, const char *mode)
```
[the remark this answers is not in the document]
levitte: You already asked once before and got the answer here: https://github.com/openssl/openssl/pull/10394#discussion_r344973132

providers/implementations/serializers/serializer_common.c
```diff
@@ -0,0 +1,340 @@
...
+
+    return xpk;
+}
+
+OSSL_OP_keymgmt_importkey_fn *ossl_prov_get_importkey(const OSSL_DISPATCH *fns)
+{
+    OSSL_OP_keymgmt_importkey_fn *importkey = NULL;
+
+    /* Pilfer the keymgmt dispatch table */
+    if (importkey == NULL) {
```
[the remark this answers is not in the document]
levitte: Because I didn't want to make the actual function non-static.

util/missingcrypto.txt
levitte: It's generated by `util/mkerr.pl`, so while it's a worthy question, I'd say it's out of scope for this PR.

crypto/serializer/serializer_meth.c
```diff
@@ -0,0 +1,510 @@
...
+
+    if (store == NULL && (store = get_serializer_store(libctx)) == NULL)
+        return 0;
+
+    return ossl_method_store_add(store, prov, id, propdef, method,
+                                 (int (*)(void *))OSSL_SERIALIZER_up_ref,
+                                 (void (*)(void *))OSSL_SERIALIZER_free);
+}
+
+/* Create a populate a serializer method */
```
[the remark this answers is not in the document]
levitte: Thanks

providers/implementations/serializers/serializer_common.c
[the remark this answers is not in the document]
mattcaswell: Probably my comment was woefully lacking in explaining what I was actually commenting on. Why check a variable is NULL, when you set it to NULL on the line before? And why have the variable anyway when it is never used?

crypto/serializer/serializer_meth.c
```diff
@@ -0,0 +1,510 @@
...
+     */
+    if (ser->serialize_data == NULL
+        && ser->serialize_object == NULL) {
+        OSSL_SERIALIZER_free(ser);
+        ERR_raise(ERR_LIB_OSSL_SERIALIZER, ERR_R_INVALID_PROVIDER_FUNCTIONS);
+        return NULL;
+    }
+    ser->prov = prov;
+    if (prov != NULL)
+        ossl_provider_up_ref(prov);
```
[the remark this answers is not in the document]
levitte: Good point...  and I just looked, and we need to review *all* other `_from_dispatch` functions, 'cause none of them check this.

doc/man3/OSSL_SERIALIZER_to_bio.pod
```diff
@@ -0,0 +1,94 @@
...
+ int OSSL_SERIALIZER_CTX_set_do_output(OSSL_SERIALIZER_CTX *ctx,
+                                       int (*do_output)(OSSL_SERIALIZER_CTX *ctx,
+                                                        BIO *out));
+
+Feature availability macros:
+
+=over 4
+
+=item OSSL_SERIALIZER_to_fp() is only available with B<OPENSSL_NO_STDIO>
+is undefined.
```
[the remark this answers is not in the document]
levitte: Yup, thanks.

doc/man3/OSSL_SERIALIZER_to_bio.pod
```diff
@@ -0,0 +1,94 @@
...
+OSSL_SERIALIZER_CTX_set0_object() sets the object to be serialized in
+the context.  This object must be intimately known by the processing
+function gitven with OSSL_SERIALIZER_CTX_set_do_output() (see it's
+description below).  No limit is imposed on the passed I<object>, so
+NULL is a valid value.
+
+OSSL_SERIALIZER_CTX_get0_object() gets back the object that was set in
+I<ctx> with OSSL_SERIALIZER_CTX_set0_object().
+
+OSSL_SERIALIZER_CTX_set_do_output() adds the pointer to a serialization
```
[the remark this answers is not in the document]
levitte: I hadn't thought of that, but if some application wants to make a serializer of their own for some type, sure why not?

doc/man7/provider-serializer.pod
```diff
@@ -0,0 +1,261 @@
...
+F<libcrypto> and the provider in B<OSSL_DISPATCH> arrays via
+B<OSSL_ALGORITHM> arrays that are returned by the provider's
+provider_query_operation() function
+(see L<provider-base(7)/Provider Functions>).
+
+All these "functions" have a corresponding function type definition
+named B<OSSL_{name}_fn>, and a helper function to retrieve the
+function pointer from a B<OSSL_DISPATCH> element named
+B<OSSL_get_{name}>.
+For example, the "function" OP_keymgmt_importdomparams() has these:
```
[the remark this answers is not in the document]
levitte: Thanks, fixing.

doc/man7/provider-serializer.pod
```diff
@@ -0,0 +1,261 @@
...
+
+=over 4
+
+=item "cipher" (B<OSSL_SERIALIZER_PARAM_CIPHER>) <UTF8 string>
+
+=item "properties" (B<OSSL_SERIALIZER_PARAM_PROPERTIES>) <UTF8 string>
+
+The name of the encryption cipher along with the desired properties to
+fetch it with.  The cipher will be used when produced an encrypted
+serialization.
```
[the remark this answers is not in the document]
levitte: Typo fixed.  The rest, I'm saving for another moment.

doc/man3/OSSL_SERIALIZER_to_bio.pod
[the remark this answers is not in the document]
mattcaswell: Well...if that was not your original intention...how were you expecting this to be used? Does it belong on the public API?

If some application does that it reminds me of the legacy _meth_ APIs that we have that we want to move away from, so I'm not sure we should be adding new APIs to do something very similar.

include/openssl/core_names.h
```diff
@@ -202,6 +202,17 @@ extern "C" {
 #define OSSL_ASYM_CIPHER_PARAM_OAEP_LABEL           "oaep-label"
 #define OSSL_ASYM_CIPHER_PARAM_OAEP_LABEL_LEN       "oaep-label-len"
 
+/*
+ * Serializer parameters
+ */
+/* The passphrase may be passed as a utf8 string or an octet string */
+#define OSSL_SERIALIZER_PARAM_CIPHER    OSSL_ALG_PARAM_CIPHER
+#define OSSL_SERIALIZER_PARAM_PROPERTIES OSSL_ALG_PARAM_PROPERTIES
+#define OSSL_SERIALIZER_PARAM_PASS      "passphrase"
```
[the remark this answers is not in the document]
levitte: Well, I aligned with the majority.  The middle one's macro name is a tad too long, though.
Meanwhile, the ASYM_CIPHER ones have appeared, with values a bit more to the right, so I've re-aligned accordingly.

crypto/serializer/serializer_lib.c
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    return ctx->object;
+}
+
+int OSSL_SERIALIZER_CTX_set_do_output(OSSL_SERIALIZER_CTX *ctx,
+                                      int (*do_output)(OSSL_SERIALIZER_CTX *ctx,
+                                                       BIO *out))
+{
+    if (ctx == NULL) {
```
richsalz: I dislike these NULL checks.  Is that policy now?  We don't do it in most places.

include/openssl/pem.h
```diff
@@ -187,6 +188,120 @@ extern "C" {
         IMPLEMENT_PEM_read(name, type, str, asn1) \
         IMPLEMENT_PEM_write_cb(name, type, str, asn1)
 
+/* Alternatives using provided serializers */
```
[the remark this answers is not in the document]
levitte: That can be said about *all* IMPLEMENT macros.  They could absolutely be moved out of the way, but in that case they should *all* be moved out of the way.  Another PR.

doc/man3/OSSL_SERIALIZER_to_bio.pod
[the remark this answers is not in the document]
richsalz: should this use the exdata facility to be consistent and generic?

providers/implementations/serializers/serializer_common.c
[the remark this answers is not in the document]
levitte: (if you look at the symbols visible in libcrypto.a, you'll see what I'm talking about.  we are *not* very clean)

include/openssl/pem.h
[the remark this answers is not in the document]
mattcaswell: I (strongly) disagree. Just because we did something wrong before, it doesn't mean we should do new stuff wrongly. There is nothing preventing us from putting this stuff in a private header, whilst leaving existing IMPLEMENT macros where they are.

providers/implementations/serializers/serializer_dh_priv.c
levitte: This what we do with MACs...  I had forgotten, however, about the routines in `providers/common/provider_util.c`, which were created exactly for this case.  I'll get to it.

crypto/serializer/serializer_lib.c
levitte: [sigh]

We don't never had an actual policy on the matter as far as I know, and depending on who you ask, you get very different answers.

providers/implementations/serializers/serializer_common.c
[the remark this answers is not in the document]
levitte: :astonished:

I'm being an utter and complete doofus, that's why.

This function was more elaborate in an earlier incantation, and I obviously didn't clean it up well enough...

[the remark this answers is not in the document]
levitte: Fixed

doc/man3/OSSL_SERIALIZER_CTX_new_by_EVP_PKEY.pod
```diff
@@ -0,0 +1,128 @@
...
+still creates a B<OSSL_SERIALIZER_CTX>, but with no associated
+serializer (L<OSSL_SERIALIZER_CTX_get_serializer(3)> returns NULL).
+This helps the caller distinguish between  an error when creating
+the B<OSSL_SERIALIZER_CTX>, and the lack the serializer support and
+act accordingly.
+
+OSSL_SERIALIZER_CTX_set_cipher_name() tells the implementation what
+cipher should be used to encrypt serialized keys.  If no cipher is
+specified, the implementation may still choose to serialize the key,
+but with unencrypted result.
```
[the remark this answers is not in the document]
levitte: I need to have a think on this, and am currently pressed for time, so I'll get back

doc/man3/OSSL_SERIALIZER_to_bio.pod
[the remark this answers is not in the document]
levitte: If you guys look at the implementation of `OSSL_SERIALIZER_CTX_new_by_EVP_PKEY`, you'll see how I *intend* these to be used.

My thought was that anyone could add support to serialize some other type, something like this for example:

``` C
static int serialize_x509_to_bio(OSSL_SERIALIZER_CTX *ctx, BIO *out)
{
    const X509 *x509 = OSSL_SERIALIZER_CTX_get0_object(ctx);

    if (ctx->ser == NULL)
        return 0;

    /*
     * ... do whatever's needed to call ctx->ser->serialize_object or
     * ctx->ser->serialize_data with required arguments
     */
}

OSSL_SERIALIZER_CTX *OSSL_SERIALIZER_CTX_new_by_X509(const X509 *x509,
                                                     const char *propquery)
{
    OSSL_SERIALIZER *ser = OSSL_SERIALIZER_fetch(NULL, "X509", propquery);

    ctx = OSSL_SERIALIZER_CTX_new(ser); /* refcnt(ser)++ */
    OSSL_SERIALIZER_free(ser);          /* refcnt(ser)-- */

    if (ctx != NULL) {
        OSSL_SERIALIZER_CTX_set0_object(ctx, x509);
        OSSL_SERIALIZER_CTX_set_do_output(ctx, serializer_x509_to_bio);
    }

    return ctx;
}
```

I'm thinking that if *someone else* wants to write support for serializing types that we haven't written support for, why not?

`OSSL_SERIALIZER_CTX_set_do_output()` is *nothing* like method building.  It sets a function pointer in a context, i.e. is a fly-by thing rather than a method building thing.  It's close to the diverse function setter you find in `include/openssl/x509_vfy.h`, like `X509_STORE_set_verify`

crypto/serializer/serializer_lib.c
richsalz: I believe we don't check most of the time and it would be good for consistency to follow that practice.  Perhaps @mspncp can come up with a "git grep" pattern to answer for sure :)

levitte: I can make them `ossl_assert`s if that suits everyone better. *That* would be consistent if anything! 'cause if you think about it, not checking NULL is a hard assertion, and we have generally said no to hard assertions, so one might ask why we should have that specifically for pointers?

richsalz: I dont want to beat a dead horse, as they say, but OpenSSL does not defend against many types programming errors, such as passing random data into a bignum or dong memcmp with a NULL argument.

Turning programming errors, that would normally be crashes, into run-time errors is controversial at best.  Runtime errors as we have seen  are often/usually ignored.

doc/man3/OSSL_SERIALIZER_to_bio.pod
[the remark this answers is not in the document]
levitte: I've had a think on these functions, and have reconsidered.  They will become internal.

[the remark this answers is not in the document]
levitte: ... or actually, disappear.  I was going to make them local only, and consider the context structure is available locally anyway, it became ridiculous to go via functions.

crypto/serializer/serializer_lib.c
levitte: Considering how we go back and forth on this *as a group*, it is indeed controversial.

crypto/serializer/serializer_meth.c
```diff
@@ -0,0 +1,510 @@
...
+            break;
+        case OSSL_FUNC_SERIALIZER_SERIALIZE_OBJECT:
+            if (ser->serialize_object == NULL)
+                ser->serialize_object =
+                    OSSL_get_OP_serializer_serialize_object(fns);
+            break;
+        }
+    }
+    /*
+     * Try to check that the method is sensible.
```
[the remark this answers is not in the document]
levitte: Done

doc/man3/OSSL_SERIALIZER_CTX_new_by_EVP_PKEY.pod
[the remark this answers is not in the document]
levitte: The current consequence is sloppy...  `set_ctx_params` will return an error, but will do nothing to prevent serialization (which will become unencrypted)...  so OK, I should probably add an indicator that the caller tried to set a cipher, so serialization will also fail due to failed settings.

doc/man7/provider-serializer.pod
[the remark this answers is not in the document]
levitte: We have the exact same problem with MACs, where exactly this went uncommented.  However, parameters are always optional, the only requirement for this is that properties come in the same OSSL_PARAM array as the cipher itself.

So we need to think another round how we think about properties passed down to providers, and how we should document that, more generally.  That takes this question out of scope of this PR.

providers/implementations/serializers/serializer_dh_priv.c
levitte: I have now documented it a bit more clearly (I copied the documentation from MACs and fiddled a bit)

include/openssl/pem.h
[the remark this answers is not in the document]
mattcaswell: Seems this is still outstanding

providers/implementations/serializers/serializer_dsa.c
```diff
@@ -0,0 +1,157 @@
...
+#include "prov/bio.h"             /* ossl_prov_bio_printf() */
+#include "prov/implementations.h" /* rsa_keymgmt_functions */
+#include "prov/providercommonerr.h" /* PROV_R_BN_ERROR */
+#include "serializer_local.h"
+
+OSSL_OP_keymgmt_importkey_fn *ossl_prov_get_dsa_importkey(void)
+{
+    static OSSL_OP_keymgmt_importkey_fn *dsa_importkey = NULL;
+
+    if (dsa_importkey == NULL)
```
mattcaswell: This variable seem unnecessary

providers/implementations/serializers/serializer_rsa.c
```diff
@@ -0,0 +1,108 @@
...
+#include "prov/implementations.h" /* rsa_keymgmt_functions */
+#include "serializer_local.h"
+
+DEFINE_SPECIAL_STACK_OF_CONST(BIGNUM_const, BIGNUM)
+
+OSSL_OP_keymgmt_importkey_fn *ossl_prov_get_rsa_importkey(void)
+{
+    static OSSL_OP_keymgmt_importkey_fn *rsa_importkey = NULL;
+
+    if (rsa_importkey == NULL)
```
mattcaswell: Variable seems unnecessary.

util/missingcrypto.txt
mattcaswell: We really should *not* be adding stuff to missingcrypto.txt except in exceptional circumstances. Those exceptional circumstances being we're going to come along real soon and take it out again. Either we should document this, or we should agree a way forward to make sure this isn't public by the time 3.0 goes live.

mattcaswell (changes requested):

mattcaswell:
Travis failures appear relevant.

util/missingcrypto.txt
levitte: Please create an issue for it, as this is a generic problem.

include/openssl/pem.h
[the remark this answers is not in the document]
levitte: I disagree with spreading things like this over several headers.  It makes this sort of stuff harder to follow.  However, I'll do that for this PR, with the intent to follow up with another PR that moves *all* IMPLEMENT macros.

crypto/serializer/serializer_pkey.c
```diff
@@ -0,0 +1,375 @@
...
+
+    prompt_idx = UI_add_input_string(ui, prompt,
+                                     UI_INPUT_FLAG_DEFAULT_PWD,
+                                     pass, 0, pass_size - 1) - 1;
+    if (prompt_idx < 0) {
+        ERR_raise(ERR_LIB_OSSL_SERIALIZER, ERR_R_UI_LIB);
+        goto end;
+    }
+
+
```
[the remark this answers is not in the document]
levitte: True

doc/man3/OSSL_SERIALIZER_CTX.pod
```diff
@@ -0,0 +1,94 @@
...
+
+OSSL_SERIALIZER_CTX_get_serializer() gets the serializer method
+currently associated with the context I<ctx>.
+
+OSSL_SERIALIZER_settable_ctx_params() returns an L<OSSL_PARAM(3)>
+array of parameter descriptors.
+
+OSSL_SERIALIZER_CTX_set_params() attempts to set parameters specified
+with an L<OSSL_PARAM(3)> array I<params>.  Parameters that the
+implementation doesn't recognised should be ignored.
```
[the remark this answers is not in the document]
levitte: Fixed

doc/man3/OSSL_SERIALIZER_CTX_new_by_EVP_PKEY.pod
```diff
@@ -0,0 +1,128 @@
...
+search for a serializer implementation that matches the algorithm of
+the B<EVP_PKEY> and the property query given with I<propquery>.  It
+will prefer to find a serializer from the same provider as the key
+data of the B<EVP_PKEY> itself, but failing that, it will choose the
+first serializer that supplies a generic serializing function.
+
+If no suitable serializer was found, OSSL_SERIALIZER_CTX_new_by_EVP_PKEY()
+still creates a B<OSSL_SERIALIZER_CTX>, but with no associated
+serializer (L<OSSL_SERIALIZER_CTX_get_serializer(3)> returns NULL).
+This helps the caller distinguish between  an error when creating
```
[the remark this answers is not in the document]
levitte: Fixed

providers/implementations/serializers/serializer_dsa.c
levitte: Fixed

providers/implementations/serializers/serializer_rsa.c
levitte: Fixed

doc/man3/OSSL_SERIALIZER_CTX_new_by_EVP_PKEY.pod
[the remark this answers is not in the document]
levitte: I won't change this here, as we have so far *assumed* that whatever cipher / digest name we pass down to providers, they are going to fetch using libcrypto functions.  To modify this, we need a higher level discussion, which makes it out of scope for this PR.

doc/man7/provider-serializer.pod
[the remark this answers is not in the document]
levitte: SSee https://github.com/openssl/openssl/pull/10394#discussion_r350172558

levitte:
> Travis failures appear relevant.

I'll have a look later

levitte:
Travis failure fixed, I hope

doc/man3/OSSL_SERIALIZER_CTX_new_by_EVP_PKEY.pod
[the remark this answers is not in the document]
mattcaswell: > I won't change this here, as we have so far assumed that whatever cipher / digest name we pass down to providers, they are going to fetch using libcrypto functions.

Have we??? I've never made that assumption!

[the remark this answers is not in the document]
levitte: You have obviously not looked at the MAC stuff.  We pass down properties and engine names along with the cipher / digest name.  That's just as much an assumption as I'm making here.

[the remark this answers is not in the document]
levitte: [rewriting]

Gah! I think I have mixed up two discussions here...

So that indicator I talked about, it's going to happen...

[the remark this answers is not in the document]
levitte: I've made our serializers a little smarter.  It's possible to pass a NULL, which will clear away any cipher setting, and either way, the provider side context now registers a *intent* to encrypt, and refuses to do anything if the cipher isn't present even though the intent was to encrypt.  That should take care of an invlid cipher.

levitte:
Apart from a timeout, CIs seem to be happy

crypto/serializer/serializer_lib.c
levitte: I've taken up this question on [openssl-project](https://mta.openssl.org/pipermail/openssl-project/2019-November/001646.html)

Meanwhile, I don't think that question should be a show-stopper for this PR.  This check can be adjusted later on.

util/missingcrypto.txt
mattcaswell: I created a new issue (#10527)...

...and assigned you to it ;-)

crypto/serializer/serializer_pkey.c
```diff
@@ -0,0 +1,374 @@
...
+#include "internal/property.h"
+#include "crypto/evp.h"
+#include "serializer_local.h"
+
+int OSSL_SERIALIZER_CTX_set_cipher_name(OSSL_SERIALIZER_CTX *ctx,
+                                        const char *cipher_name)
+{
+    OSSL_PARAM params[] = { OSSL_PARAM_END, OSSL_PARAM_END };
+
+    params[0] = OSSL_PARAM_construct_utf8_string(OSSL_SERIALIZER_PARAM_CIPHER,
```
mattcaswell: We don't seem to have support here for setting the properties (which must be set at the same time as the cipher name).

doc/man7/provider-serializer.pod
```diff
@@ -0,0 +1,266 @@
...
+=item "cipher" (B<OSSL_SERIALIZER_PARAM_CIPHER>) <UTF8 string>
+
+The name of the encryption cipher to be used when generating encrypted
+serialization.  This is used when serializing private keys, as well as
+other objects that need protection.
+
+=item "properties" (B<OSSL_SERIALIZER_PARAM_PROPERTIES>) <UTF8 string>
+
+The properties to be queried when trying to fetch the algorithm given
+with the "cipher" parameter.
```
mattcaswell: I think we should be explicit about allowing the possibility that a provider may not support fetching. E.g. see:

https://github.com/openssl/openssl/blame/master/doc/man3/EVP_DigestVerifyInit.pod#L28-L37

mattcaswell (changes requested):

crypto/pem/pem_local.h
```diff
@@ -0,0 +1,128 @@
...
+        IMPLEMENT_PEM_provided_write_body_pass();                       \
+        IMPLEMENT_PEM_provided_write_body_main(type, outtype);          \
+        IMPLEMENT_PEM_provided_write_body_fallback_cb(str, asn1,        \
+                                                      writename);       \
+    }
+
+# ifdef OPENSSL_NO_STDIO
+
+#  define IMPLEMENT_PEM_provided_write_fp(name, type, str, asn1) /**/
+#  define IMPLEMENT_PEM_provided_write_cb_fp(name, type, str, asn1) /**/
```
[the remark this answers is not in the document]
levitte: Copied from the corresponding macros in include/ooenssl/pem.h, lines 69+

crypto/serializer/serializer_pkey.c
levitte: Good point, will fix

crypto/pem/pem_local.h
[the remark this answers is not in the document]
levitte: Removed

crypto/serializer/serializer_pkey.c
levitte: Done

doc/man7/provider-serializer.pod
levitte: Done.  I hope the added text is satisfactory.

crypto/serializer/serializer_pkey.c
```diff
@@ -0,0 +1,381 @@
...
+#include "internal/property.h"
+#include "crypto/evp.h"
+#include "serializer_local.h"
+
+int OSSL_SERIALIZER_CTX_set_cipher(OSSL_SERIALIZER_CTX *ctx,
+                                   const char *cipher_name,
+                                   const char *propquery)
+{
+    OSSL_PARAM params[] = {
+        OSSL_PARAM_END, OSSL_PARAM_END , OSSL_PARAM_END, OSSL_PARAM_END
```
mattcaswell: There seems to be one two many `OSSL_PARAM_END` here

doc/man3/OSSL_SERIALIZER_CTX_new_by_EVP_PKEY.pod
```diff
@@ -0,0 +1,134 @@
+=pod
+
+=head1 NAME
+
+OSSL_SERIALIZER_CTX_new_by_EVP_PKEY,
+OSSL_SERIALIZER_CTX_set_cipher_name,
```
mattcaswell: This needs updating

mattcaswell (changes requested):

crypto/serializer/serializer_pkey.c
levitte: Uhmmm, no?

doc/man3/OSSL_SERIALIZER_CTX_new_by_EVP_PKEY.pod
levitte: Fixed

crypto/serializer/serializer_pkey.c
levitte: Oh silly me.  I cannot count today, it seems...

levitte: Fixed

levitte:
I don't remember exactly where, but I believe indentation was mentioned somewhere among all the comments.  And sure, it makes sense to talk about that, since some of the printouts happen as part of `openssl x509 -text`, which does specify indent.  The code here *ignores indentation completely*, and I did promise a plan to deal with that.  #10531 is it, and can be treated independently from this.

mattcaswell (approved):

levitte:
Merged.

3d83c7353630450a4ce4fffc9c75693565e094c2 CORE: ossl_namemap_add_names(): new function to add multiple names
36fa4d8a0df9dc168047fadd0365966c7116b31d CORE: pass the full algorithm definition to the method constructor
0d003c52d3dcf4b076bb01a6767cdd5ace2d79f6 SERIALIZER: New API for serialization of objects through providers
742496f1309d04f4921ca64e4b6315a45a47b4af SERIALIZER: add functions for serialization to file
1793d270f3c6f6b859e40ef1fa7bea1fd7e447c8 CORE: expose the property parsers and checker to the rest of the libraries
866234ac35e665f20c646059b1d92c5e9eb0c7ab SERIALIZER: add support for serializing EVP_PKEYs
f864a9396a690958d26c87827ba6f26e7b010caf SERIALIZER: add hooks in PEM_write_bio_ and PEM_write_fp_ routines
54c1711f87cdc50e2266e2f0a063ccace08173d2 SERIALIZER: add hooks in EVP_PKEY_print_ routines
63665fff84a4c79cd2acece4409036699f2e44a7 PROV BIO: add a BIO_vprintf() upcall, and a provider BIO library
cb58d81e68c72ab0128e0a5fc6faa007f8632acd PROV SERIALIZER: add common functionality to serialize keys
677add38003b37cab06ebc59260c8d28745bb948 PROV SERIALIZER: add support for writing RSA keys
045e51cbf49a0773f7d045180d9d8b8b73d70e90 PROV SERIALIZER: add support for writing DH keys and parameters
264b789bc240e6d0f642b274b6ab3cf8470e4aef PROV SERIALIZER: add support for writing DSA keys and parameters
6ae5543c5f8476f4cae9c00209ef062746bbfd3c TEST: add tests of text and PEM printout of a provider made key
c48e2d106b8740f817b099310fb084375b743196 Add NEWS and CHANGES entries about OSSL_SERIALIZER
31d3a7590274b48b194ed070ae531238764647f9 util/find-doc-nits: limit the prototype check
14ee781eef0e55563432f377d8911529823bee58 util/find-doc-nits: ignore macros ending in _fnsig

PR #10407, opened 2019-11-11
WIP: Remove methods from const EVP_CIPHER structures

slontis:
Tests that use engines are current broken as they still run through the deprecated path.

##### Checklist


crypto/evp/evp_enc.c
```diff
@@ -276,6 +276,7 @@ int EVP_CipherInit_ex(EVP_CIPHER_CTX *ctx, const EVP_CIPHER *cipher,
         case NID_rc5_ecb:
         case NID_rc5_cfb64:
         case NID_rc5_ofb64:
+        case NID_rc2_ecb:
```
levitte: At some point, we should stop maintaining this absurd list, and instead simply try to fetch, and fall back to legacy when that fails.

slontis: At some point there should be no legacy (so also no list ) :)

include/crypto/evp.h
```diff
@@ -278,6 +278,51 @@ struct evp_cipher_st {
...
+IMPLEMENT_EVP_CIPHER_CONST(alg, nid##_##keybits##_cfb1, keybits, 1, 16, cfb1 ,CFB, flags)  \
+IMPLEMENT_EVP_CIPHER_CONST(alg, nid##_##keybits##_cfb8, keybits, 1, 16, cfb8, CFB, flags)
+
+#define IMPLEMENT_EVP_CIPHER_CONST_modes2(alg, nid, keybits, blocklen, ivlen, cbits, flags)   \
+IMPLEMENT_EVP_CIPHER_CONST2(alg, nid##_ecb, keybits, blocklen, 0, ecb, ECB, flags)            \
+IMPLEMENT_EVP_CIPHER_CONST2(alg, nid##_cbc, keybits, blocklen, ivlen, cbc, CBC, flags)        \
+IMPLEMENT_EVP_CIPHER_CONST2(alg, nid##_ofb##cbits, keybits, 1, ivlen, ofb, OFB, flags)        \
+IMPLEMENT_EVP_CIPHER_CONST2(alg, nid##_cfb##cbits, keybits, 1, ivlen, cfb##cbits, CFB, flags)
+
+
```
levitte: Do we have *any* use of these macros outside of `crypto/evp/`?  If not, they could as well be placed in `crypto/evp/evp_local.h`

crypto/evp/evp_enc.c
levitte: Yes, absolutely...  but I'm thinking that *even now*, we could throw away that list.  It was a help before we got everything down, but at this point, I think the new `EVP_CIPHER` and `EVP_MD` have matured enough to make that change.

include/crypto/evp.h
slontis: updated

levitte:
The CIs are having a fit...  which simply indicates that there are still corners in the code that depend on the presence of the legacy stuff.

This (from Travis) could be a starting point:

```
/home/travis/build/openssl/openssl/_build/util/shlib_wrap.sh /home/travis/build/openssl/openssl/_build/test/dtls_mtu_test => 139
not ok 1 - running dtls_mtu_test
#   Failed test 'running dtls_mtu_test'
#   at ../../_srcdist/test/recipes/80-test_dtls_mtu.t line 21.
80-test_dtls_mtu.t ................. Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/1 subtests 
```

(139 is a SIGSEGV if I'm not mistaken)

slontis:
Most of the 70- ones are related to engines ossltest being used. So this is a good test of engines :)
Some of the others are related to the 2 missing ciphers (one PR just merged + one more existing one to go)

slontis:
Closing as this is not going to work for engines.

PR #10408, opened 2019-11-11
Properties: make query cache reference count aware.

### LINKED ISSUE #10379: Thread-safety issue in evp_generic_fetch
There appears to be a thread-safety issue in `evp_generic_fetch` where one thread could be fetching a cached `EVP_CIPHER` object that another thread is about to free. The reason is that the fetching from the cache and the bumping of the refcount don't happen atomically.

**Steps to reproduce (Linux)**:

1. Write the following code to a file called `crash.c`:

```c
#include <pthread.h>
#include <openssl/evp.h>

void* run(void* p)
{
    for (;;)
    {
        EVP_CIPHER_CTX* ctx = EVP_CIPHER_CTX_new();
        EVP_EncryptInit(ctx, EVP_aes_256_cbc(), NULL, NULL);
        EVP_CIPHER_CTX_free(ctx);
    }
}

int main()
{
    pthread_t t1, t2;
    pthread_create(&t1, NULL, &run, NULL);
    pthread_create(&t2, NULL, &run, NULL);
    pthread_join(t1, NULL);
    pthread_join(t2, NULL);
}
```

2. Build OpenSSL as follows from master:

```
$ ./Configure threads enable-crypto-mdebug --debug -fsanitize=address debug-linux-x86_64
$ make -j$(nproc)
```

3. Build the test application:

```
$ gcc -o crash -fsanitize=address -Iinclude /tmp/crash.c -L. -lcrypto -lpthread
```

4. Run the application and observe the crash:

```
$ LD_LIBRARY_PATH=. ./crash
=================================================================
==47155==AddressSanitizer: while reporting a bug found another one. Ignoring.
==47155==ERROR: AddressSanitizer: heap-use-after-free on address 0x611000010370 at pc 0x7f0d17d7eb03 bp 0x7f0d138fda50 sp 0x7f0d138fda40
READ of size 8 at 0x611000010370 thread T2
    #0 0x7f0d17d7eb02 in EVP_CIPHER_up_ref crypto/evp/evp_enc.c:1505
    #1 0x7f0d17d7e9a4 in evp_cipher_up_ref crypto/evp/evp_enc.c:1482
    #2 0x7f0d17d8013c in inner_evp_generic_fetch crypto/evp/evp_fetch.c:354
    #3 0x7f0d17d8026b in evp_generic_fetch crypto/evp/evp_fetch.c:370
    #4 0x7f0d17d7ea0b in EVP_CIPHER_fetch crypto/evp/evp_enc.c:1493
    #5 0x7f0d17d77281 in EVP_CipherInit_ex crypto/evp/evp_enc.c:320
    #6 0x7f0d17d769f2 in EVP_CipherInit crypto/evp/evp_enc.c:77
    #7 0x7f0d17d784ba in EVP_EncryptInit crypto/evp/evp_enc.c:540
    #8 0x55d425f96b7c in run (/tmp/crash+0xb7c)
    #9 0x7f0d16fa16da in start_thread (/lib/x86_64-linux-gnu/libpthread.so.0+0x76da)
    #10 0x7f0d176e688e in __clone (/lib/x86_64-linux-gnu/libc.so.6+0x12188e)

0x611000010370 is located 112 bytes inside of 232-byte region [0x611000010300,0x6110000103e8)
freed by thread T1 here:
    #0 0x7f0d184f17b8 in __interceptor_free (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xde7b8)
    #1 0x7f0d17dc69fb in CRYPTO_free crypto/mem.c:306
    #2 0x7f0d17d7ed37 in EVP_CIPHER_free crypto/evp/evp_enc.c:1521
    #3 0x7f0d17d76709 in EVP_CIPHER_CTX_reset crypto/evp/evp_enc.c:38
    #4 0x7f0d17d7697f in EVP_CIPHER_CTX_free crypto/evp/evp_enc.c:68
    #5 0x55d425f96b88 in run (/tmp/crash+0xb88)
    #6 0x7f0d16fa16da in start_thread (/lib/x86_64-linux-gnu/libpthread.so.0+0x76da)

previously allocated by thread T2 here:
    #0 0x7f0d184f1b50 in __interceptor_malloc (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xdeb50)
    #1 0x7f0d17dc66ac in CRYPTO_malloc crypto/mem.c:215
    #2 0x7f0d17dc6713 in CRYPTO_zalloc crypto/mem.c:230
    #3 0x7f0d17d7dc9a in evp_cipher_new crypto/evp/evp_enc.c:1314
    #4 0x7f0d17d7df5f in evp_cipher_from_dispatch crypto/evp/evp_enc.c:1360
    #5 0x7f0d17d7f997 in construct_evp_method crypto/evp/evp_fetch.c:257
    #6 0x7f0d17dbf74b in ossl_method_construct_this crypto/core_fetch.c:34
    #7 0x7f0d17dbf377 in algorithm_do_this crypto/core_algorithm.c:50
    #8 0x7f0d17dd9aa0 in provider_forall_loaded crypto/provider_core.c:581
    #9 0x7f0d17dd9cd8 in ossl_provider_forall_loaded crypto/provider_core.c:650
    #10 0x7f0d17dbf5b9 in ossl_algorithm_do_all crypto/core_algorithm.c:72
    #11 0x7f0d17dbfdd0 in ossl_method_construct crypto/core_fetch.c:89
    #12 0x7f0d17d7ffff in inner_evp_generic_fetch crypto/evp/evp_fetch.c:339
    #13 0x7f0d17d8026b in evp_generic_fetch crypto/evp/evp_fetch.c:370
    #14 0x7f0d17d7ea0b in EVP_CIPHER_fetch crypto/evp/evp_enc.c:1493
    #15 0x7f0d17d77281 in EVP_CipherInit_ex crypto/evp/evp_enc.c:320
    #16 0x7f0d17d769f2 in EVP_CipherInit crypto/evp/evp_enc.c:77
    #17 0x7f0d17d784ba in EVP_EncryptInit crypto/evp/evp_enc.c:540
    #18 0x55d425f96b7c in run (/tmp/crash+0xb7c)
    #19 0x7f0d16fa16da in start_thread (/lib/x86_64-linux-gnu/libpthread.so.0+0x76da)

Thread T2 created by T0 here:
    #0 0x7f0d1844ad2f in __interceptor_pthread_create (/usr/lib/x86_64-linux-gnu/libasan.so.4+0x37d2f)
    #1 0x55d425f96c68 in main (/tmp/crash+0xc68)
    #2 0x7f0d175e6b96 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21b96)

Thread T1 created by T0 here:
    #0 0x7f0d1844ad2f in __interceptor_pthread_create (/usr/lib/x86_64-linux-gnu/libasan.so.4+0x37d2f)
    #1 0x55d425f96c4b in main (/tmp/crash+0xc4b)
    #2 0x7f0d175e6b96 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21b96)

SUMMARY: AddressSanitizer: heap-use-after-free crypto/evp/evp_enc.c:1505 in EVP_CIPHER_up_ref
Shadow bytes around the buggy address:
  0x0c227fffa010: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c227fffa020: fd fd fd fd fd fd fd fd fd fd fd fd fd fa fa fa
  0x0c227fffa030: fa fa fa fa fa fa fa fa fd fd fd fd fd fd fd fd
  0x0c227fffa040: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c227fffa050: fd fd fd fd fd fa fa fa fa fa fa fa fa fa fa fa
=>0x0c227fffa060: fd fd fd fd fd fd fd fd fd fd fd fd fd fd[fd]fd
  0x0c227fffa070: fd fd fd fd fd fd fd fd fd fd fd fd fd fa fa fa
  0x0c227fffa080: fa fa fa fa fa fa fa fa fd fd fd fd fd fd fd fd
  0x0c227fffa090: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c227fffa0a0: fd fd fd fd fd fa fa fa fa fa fa fa fa fa fa fa
  0x0c227fffa0b0: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==47155==ABORTING
```

This was reproduced on this Linux machine:

```
$ uname -a
Linux 4.19.67-microsoft-standard #1 SMP Sun Aug 18 13:37:54 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
$ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.3 LTS"
```

**Root cause**:

The following interleaving of operations would cause this:

1. Thread A calls `EVP_EncryptInit` to fetch a brand new `EVP_CIPHER`. Because the cipher is not in the cache, a new one is created and store via `ossl_method_store_cache_set`.
2. Thread B calls `EVP_EncryptInit` to fetch the same `EVP_CIPHER`. Because it's already in the cache, [this call to `ossl_method_store_cache_get`](https://github.com/openssl/openssl/blob/master/crypto/evp/evp_fetch.c#L317) returns it.
3. Before thread B can [increase the reference count](https://github.com/openssl/openssl/blob/master/crypto/evp/evp_fetch.c#L354) on this fetched cipher, thread A calls `EVP_CIPHER_CTX_free`, and because it finds the cipher to have a reference count of 1, it decrements it and frees it.
4. Thread B now tries to increase the reference count on a freed cipher.

paulidale:
The property query cache was not reference count aware and this could cause
problems if the property store removes an algorithm while it is being returned
from an asynchronous query.  This change makes the cache reference count aware
and avoids disappearing algorithms.

Fixes #10379

### DIFF
--- a/crypto/evp/evp_fetch.c
+++ b/crypto/evp/evp_fetch.c
@@ -186,13 +186,9 @@ static void *get_evp_method_from_store(OPENSSL_CTX *libctx, void *store,
         && (store = get_evp_method_store(libctx)) == NULL)
         return NULL;
 
-    (void)ossl_method_store_fetch(store, meth_id, methdata->propquery,
-                                  &method);
-
-    if (method != NULL
-        && !methdata->refcnt_up_method(method)) {
-        method = NULL;
-    }
+    if (!ossl_method_store_fetch(store, meth_id, methdata->propquery,
+                                 &method))
+        return NULL;
     return method;
 }
 
@@ -328,7 +324,6 @@ inner_evp_generic_fetch(OPENSSL_CTX *libctx, int operation_id,
         mcmdata.names = name;
         mcmdata.propquery = properties;
         mcmdata.method_from_dispatch = new_method;
-        mcmdata.destruct_method = free_method;
         mcmdata.refcnt_up_method = up_ref_method;
         mcmdata.destruct_method = free_method;
         if ((method = ossl_method_construct(libctx, operation_id,
@@ -343,10 +338,9 @@ inner_evp_generic_fetch(OPENSSL_CTX *libctx, int operation_id,
             if (name_id == 0)
                 name_id = ossl_namemap_name2num(namemap, name);
             meth_id = evp_method_id(operation_id, name_id);
-            ossl_method_store_cache_set(store, meth_id, properties, method);
+            ossl_method_store_cache_set(store, meth_id, properties, method,
+                                        up_ref_method, free_method);
         }
-    } else {
-        up_ref_method(method);
     }
 
     return method;
--- a/crypto/property/property.c
+++ b/crypto/property/property.c
@@ -22,20 +22,25 @@
 #include "property_local.h"
 
 /* The number of elements in the query cache before we initiate a flush */
-#define IMPL_CACHE_FLUSH_THRESHOLD  500
+#define IMPL_CACHE_FLUSH_THRESHOLD  50
+
+typedef struct {
+    void *method;
+    int (*up_ref)(void *);
+    void (*free)(void *);
+} METHOD;
 
 typedef struct {
     const OSSL_PROVIDER *provider;
     OSSL_PROPERTY_LIST *properties;
-    void *method;
-    void (*method_destruct)(void *);
+    METHOD method;
 } IMPLEMENTATION;
 
 DEFINE_STACK_OF(IMPLEMENTATION)
 
 typedef struct {
     const char *query;
-    void *method;
+    METHOD method;
     char body[1];
 } QUERY;
 
@@ -67,6 +72,16 @@ DEFINE_SPARSE_ARRAY_OF(ALGORITHM);
 static void ossl_method_cache_flush(OSSL_METHOD_STORE *store, int nid);
 static void ossl_method_cache_flush_all(OSSL_METHOD_STORE *c);
 
+static int ossl_method_up_ref(METHOD *method)
+{
+    return (*method->up_ref)(method->method);
+}
+
+static void ossl_method_free(METHOD *method)
+{
+    (*method->free)(method->method);
+}
+
 int ossl_property_read_lock(OSSL_METHOD_STORE *p)
 {
     return p != NULL ? CRYPTO_THREAD_read_lock(p->lock) : 0;
@@ -95,15 +110,17 @@ static int query_cmp(const QUERY *a, const QUERY *b)
 static void impl_free(IMPLEMENTATION *impl)
 {
     if (impl != NULL) {
-        if (impl->method_destruct)
-            impl->method_destruct(impl->method);
+        ossl_method_free(&impl->method);
         OPENSSL_free(impl);
     }
 }
 
 static void impl_cache_free(QUERY *elem)
 {
-    OPENSSL_free(elem);
+    if (elem != NULL) {
+        ossl_method_free(&elem->method);
+        OPENSSL_free(elem);
+    }
 }
 
 static void alg_cleanup(ossl_uintmax_t idx, ALGORITHM *a)
@@ -132,7 +149,7 @@ OSSL_METHOD_STORE *ossl_method_store_new(OPENSSL_CTX *ctx)
             return NULL;
         }
         if ((res->lock = CRYPTO_THREAD_lock_new()) == NULL) {
-            OPENSSL_free(res->algs);
+            ossl_sa_ALGORITHM_free(res->algs);
             OPENSSL_free(res);
             return NULL;
         }
@@ -180,13 +197,14 @@ int ossl_method_store_add(OSSL_METHOD_STORE *store, const OSSL_PROVIDER *prov,
     impl = OPENSSL_malloc(sizeof(*impl));
     if (impl == NULL)
         return 0;
-    if (method_up_ref != NULL && !method_up_ref(method)) {
+    impl->method.method = method;
+    impl->method.up_ref = method_up_ref;
+    impl->method.free = method_destruct;
+    if (!ossl_method_up_ref(&impl->method)) {
         OPENSSL_free(impl);
         return 0;
     }
     impl->provider = prov;
-    impl->method = method;
-    impl->method_destruct = method_destruct;
 
     /*
      * Insert into the hash table if required.
@@ -262,10 +280,10 @@ int ossl_method_store_remove(OSSL_METHOD_STORE *store, int nid,
     for (i = 0; i < sk_IMPLEMENTATION_num(alg->impls); i++) {
         IMPLEMENTATION *impl = sk_IMPLEMENTATION_value(alg->impls, i);
 
-        if (impl->method == method) {
+        if (impl->method.method == method) {
+            impl_free(impl);
             sk_IMPLEMENTATION_delete(alg->impls, i);
             ossl_property_unlock(store);
-            impl_free(impl);
             return 1;
         }
     }
@@ -279,6 +297,7 @@ int ossl_method_store_fetch(OSSL_METHOD_STORE *store, int nid,
     ALGORITHM *alg;
     IMPLEMENTATION *impl;
     OSSL_PROPERTY_LIST *pq = NULL, *p2;
+    METHOD *best_method = NULL;
     int ret = 0;
     int j, best = -1, score, optional;
 
@@ -302,7 +321,7 @@ int ossl_method_store_fetch(OSSL_METHOD_STORE *store, int nid,
 
     if (prop_query == NULL) {
         if ((impl = sk_IMPLEMENTATION_value(alg->impls, 0)) != NULL) {
-            *method = impl->method;
+            best_method = &impl->method;
             ret = 1;
         }
         goto fin;
@@ -322,14 +341,18 @@ int ossl_method_store_fetch(OSSL_METHOD_STORE *store, int nid,
         impl = sk_IMPLEMENTATION_value(alg->impls, j);
         score = ossl_property_match_count(pq, impl->properties);
         if (score > best) {
-            *method = impl->method;
+            best_method = &impl->method;
+            best = score;
             ret = 1;
             if (!optional)
                 goto fin;
-            best = score;
         }
     }
 fin:
+    if (ret && ossl_method_up_ref(best_method))
+        *method = best_method->method;
+    else
+        ret = 0;
     ossl_property_unlock(store);
     ossl_property_free(pq);
     return ret;
@@ -414,7 +437,7 @@ static void impl_cache_flush_cache(QUERY *c, IMPL_CACHE_FLUSH *state)
     state->seed = n;
 
     if ((n & 1) != 0)
-        OPENSSL_free(lh_QUERY_delete(state->cache, c));
+        impl_cache_free(lh_QUERY_delete(state->cache, c));
     else
         state->nelem++;
 }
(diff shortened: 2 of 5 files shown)


paulidale:
Travis is relevant -- some things aren't getting cleaned up properly.

crypto/property/property.c
```diff
@@ -67,6 +72,17 @@ DEFINE_SPARSE_ARRAY_OF(ALGORITHM);
 static void ossl_method_cache_flush(OSSL_METHOD_STORE *store, int nid);
 static void ossl_method_cache_flush_all(OSSL_METHOD_STORE *c);
 
+static int ossl_method_up_ref(METHOD *method)
+{
+    return method->up_ref != NULL ? (*method->up_ref)(method->method) : 1;
```
levitte: Small note: you do know that the `(*` and `)` wrapping is quite unnecessary, right?  I so not used to see them that I go "Huh?...  oh" every time I see them.  I'm not going to be hard-headed about it, though, just sharing the thought

paulidale: Yeah, I do.  I find including them makes the fact that it is a pointer to a function far more obvious.  It's not an issue here since there is the `->` operator but in many other cases it isn't so obvious.

Likewise, I tend to use `&` before a function name to take its address.

jblazquez:
Thank you! I can no longer reproduce the issue with this change.

doc/internal/man3/OSSL_METHOD_STORE.pod
```diff
@@ -33,7 +33,9 @@ ossl_method_store_cache_get, ossl_method_store_cache_set
  int ossl_method_store_cache_get(OSSL_METHOD_STORE *store, int nid,
                                  const char *prop_query, void **method);
  int ossl_method_store_cache_set(OSSL_METHOD_STORE *store, int nid,
-                                 const char *prop_query, void *method);
+                                 const char *prop_query, void *method
```
jblazquez: Missing comma at the end of this line

doc/internal/man3/OSSL_METHOD_STORE.pod
```diff
@@ -95,6 +97,10 @@ The result, if any, is returned in I<method>.
 ossl_method_store_cache_set() sets a cache entry identified by I<nid> with the
 property query I<prop_query> in the I<store>.
 Future calls to ossl_method_store_cache_get() will return the specified I<method>.
+If the I<method_up_ref> function is given, it's called to increment the
+reference count of the method.
+If the I<method_destruct> function is given, it's called when this function
```
jblazquez: Should this say "when this _method_ is being freed from the cache"?

doc/internal/man3/OSSL_METHOD_STORE.pod
paulidale: Okay.

doc/internal/man3/OSSL_METHOD_STORE.pod
paulidale: I'll have a look.

paulidale:
> I can no longer reproduce the issue with this change.

Unfortunately, there are some references being kept around which shouldn't be.

crypto/property/property.c
```diff
@@ -67,6 +72,17 @@ DEFINE_SPARSE_ARRAY_OF(ALGORITHM);
...
+}
+
+static void ossl_method_free(METHOD *method)
+{
+    if (method->free != NULL)
+        (*method->free)(method->method);
+}
```
levitte: Re both up_ref and free, I think you shouldn't check for NULL, 'cause internal.  If your code hasn't set things up correctly, it deserves a _KABOOM_.  That may also help you find intricate stoopid bugs quicker.

paulidale: I was thinking of making these call mandatory through out -- i.e. the pointers can never be NULL.

Taking out the null checks isn't helping.

paulidale: I'm guessing but I've a feeling the store needs to clean the cache better -- without ref counting in the cache, it won't matter but with counting it does.

Something for when I'm feeling more alert.

crypto/evp/evp_fetch.c
```diff
@@ -343,10 +344,9 @@ inner_evp_generic_fetch(OPENSSL_CTX *libctx, int operation_id,
             if (name_id == 0)
                 name_id = ossl_namemap_name2num(namemap, name);
             meth_id = evp_method_id(operation_id, name_id);
-            ossl_method_store_cache_set(store, meth_id, properties, method);
+            ossl_method_store_cache_set(store, meth_id, properties, method,
+                                        up_ref_method, free_method);
         }
-    } else {
-        up_ref_method(method);
```
levitte: I was surprised by this, but seeing the rest of the change, I get it, the recount responsibility has moved entirely to the store & cache code.

doc/internal/man3/OSSL_METHOD_STORE.pod
```diff
@@ -95,6 +97,10 @@ The result, if any, is returned in I<method>.
 ossl_method_store_cache_set() sets a cache entry identified by I<nid> with the
 property query I<prop_query> in the I<store>.
 Future calls to ossl_method_store_cache_get() will return the specified I<method>.
+If the I<method_up_ref> function is given, it's called to increment the
+reference count of the method.
+If the I<method_destruct> function is given, it's called when this
+implementation is being freed from the I<cache>.
```
levitte: This needs to be rephrase, since they are mandatory

crypto/evp/evp_fetch.c
paulidale: This leads to double free problems.
It does solve the current leaked memory problems.

There is something fundamental I'm missing & my brain isn't working well at present.

doc/internal/man3/OSSL_METHOD_STORE.pod
paulidale: Agreed, will change.

crypto/evp/evp_fetch.c
paulidale: oops, wrong place for this comment :(

doc/internal/man3/OSSL_METHOD_STORE.pod
paulidale: Done.

crypto/evp/evp_fetch.c
paulidale: Figured it out, missed a ref counter update.

But yes, the store and cache are responsible for the reference counting now.  A return from either will have the count increased.

paulidale:
Rebased and squashed.  Passing CI now.

paulidale:
Ping @levitte?

paulidale:
Merged, thanks.

PR #10312, opened 2019-10-31
Add more tests for `apps/req`

romen:
https://github.com/openssl/openssl/issues/10224#issuecomment-546593113
highlighted that existing testing infrastructure is not covering common
usage patterns of the `req` app.

This commit explicitly adds request generations thorugh the CLI using
RSA, DSA and ECDSA (P-256) keys.

##### Checklist

- [x] tests are added or updated

Issue #6988, opened 2018-08-16
Support Ed25519 sign/verify operations from OpenSSL CLI


parasssh:
The issue #5873 only fixed the error reporting for Ed25519 sign operation with "dgst" utility. 

But as it stands, as of today, there is no way to sign/verify a message/file using Ed25519 keys and signature algorithm using the OpenSSL CLI. 

I suggest adding a option -one-shot to the pkeyutl

CodeGuro:
After coming across this issue, I did some research and found the threads #5873, #5880, #6284, and followed them up to this one. I came to say that the same issue affects Ed448 sign operation as well.

la-urre:
Hey ! Any update on this ?

mattcaswell:
> I suggest adding a option -one-shot to the pkeyutl

This sounds reasonable but this is very low priority at the moment. Everyone is concentrating on the changes required for 3.0. Adding the "help wanted" label. PRs welcome.

romen:
`master`, and the upcoming 3.0 alpha1 release, ships with limited (there are size limits due to buffering) support for signing files from the command line with `pkeyutl`.

A related discussion is happening on the `openssl-users` mailing list in this thread: https://mta.openssl.org/pipermail/openssl-users/2020-April/012248.html


An extract from it that is relevant to this PR:

```sh
λ /tmp/test25519/ ### Ensure OpenSSL dev build is in use for this shell
λ /tmp/test25519/ which openssl ; openssl version
/opt/openssl-master/bin/openssl
OpenSSL 3.0.0-dev xx XXX xxxx (Library: OpenSSL 3.0.0-dev xx XXX xxxx)
λ /tmp/test25519/ ### Generate Ed25519 private key
λ /tmp/test25519/ openssl genpkey -algorithm Ed25519 -out priv.pem
λ /tmp/test25519/ ### Extract pub key from private key
λ /tmp/test25519/ openssl pkey -in priv.pem -pubout -out pub.pem
λ /tmp/test25519/ ###
λ /tmp/test25519/ ### Up to this point all the commands were compatible
λ /tmp/test25519/ ### with OpenSSL 1.1.1 releases, the next one is the
λ /tmp/test25519/ ### one that requires OpenSSL 3.0.0-dev as `pkeyutl`
λ /tmp/test25519/ ### now has support for `-rawin` which is required
λ /tmp/test25519/ ### for signing/veryfing files with Ed25519 keys. 
λ /tmp/test25519/ ###
[...]

PR #10410, opened 2019-11-11
More tests for CLI usage of asymmetric keys

romen:
As followup to #10312 this PR started with adding `req` tests for Ed25519 keys similar to the ones in the parent PR.

I then went ahead and added "Conversion tests" similar to what is currently done for ECDSA keys and other asymmetric keys.

Finally, noticing that there was limited testing for generating and verifying signatures using `dgst` or `pkeyutl` I added more tests to give full coverage for this use case with RSA/DSA/ECDSA/EdDSA keys.

Currently this is limited to Ed25519 keys, as I used RFC 8410 Sec. 10 to retrieve a "test vector" key and could not find something similar in the standards for Ed448.

Note this is currently based on top of #10312 and not on latest `master`, so only the last 2 commits are actually relevant when reviewing.

@levitte (and any other Perl savvy reviewer): I ask forgiveness in advance for the way the perl code is written. Please don't spare any feedback on better ways to achieve what I wanted, so that I can improve my (nonexistent) perl-fu.

##### Checklist


romen:
Notice https://github.com/openssl/openssl/blob/1ee53d38822258c0c9c41fd1415bf4de935acf1b/test/recipes/20-test_dgst.t#L87-L98

that this last test is always skipped, because of #6988 .

The idea of having it there is so that when we (or a generous contributor) add support for EdDSA in `dgst` line 91 can be simply removed to assist TDD (and include testing for the new feature right away).

romen:
Suggestions are welcome for a "normative" source of test vectors for encoding of Ed448 keys.

romen:
Rebased on latest `master` after #10312 got merged.

Ready for approval.

mattcaswell:
> Currently this is limited to Ed25519 keys, as I used RFC 8410 Sec. 10 to retrieve a "test vector" key and could not find something similar in the standards for Ed448.

While its a nice to have to use a key from the RFC (i.e. to prove that we can read and use such keys), I don't think that's the only type of key we should use. There is a great value from a regression test perspective in using a self-generated key if we can't get one from somewhere else. I'm not aware of a source for "test vector" Ed448 keys, therefore I would just generate one and create tests using that.

Aside from the above this PR looks good and I have no comments.

romen:
> > Currently this is limited to Ed25519 keys, as I used RFC 8410 Sec. 10 to retrieve a "test vector" key and could not find something similar in the standards for Ed448.
> 
> While its a nice to have to use a key from the RFC (i.e. to prove that we can read and use such keys), I don't think that's the only type of key we should use. There is a great value from a regression test perspective in using a self-generated key if we can't get one from somewhere else. I'm not aware of a source for "test vector" Ed448 keys, therefore I would just generate one and create tests using that.
> 
> Aside from the above this PR looks good and I have no comments.

I am actually planning to use this, generated using openssl from the values in RFC 8032:
https://github.com/openssl/openssl/blob/3cc26f2eba8a8c16ac559e68c05c094d7ea6bd8b/test/recipes/30-test_evp_data/evppkey.txt#L17609-L17615

and associated 
https://github.com/openssl/openssl/blob/3cc26f2eba8a8c16ac559e68c05c094d7ea6bd8b/test/recipes/30-test_evp_data/evppkey.txt#L17659-L17663

mattcaswell:
Works for me.

romen:
Pushed commits for Ed448 testing!

romen:
@mattcaswell could you review this?
@levitte do you have comments on Perl?

levitte:
The perl looks good. I found it very readable.

levitte:
Having read what @mattcaswell has written so far, I'm confident he'd approve as well.

levitte:
Merged.

master:
81722fdf2e01cfa71c46abbcc19e65aa003e083f More testing for CLI usage of Ed25519 and Ed448 keys
ef1e59ed833e8ed1d5f4de5b0c734da8561890e3 More testing for sign/verify through `dgst` and `pkeyutl`

I couldn't cherry-pick cleanly to 1.1.1, please make a backport PR

levitte:
Ping @romen

romen:
Will do!

PR #8446, opened 2019-03-09
WIP: New RSA decryption API

kroeckx:
This is the start of a new API that should make it easier to avoid an attack.

##### Checklist

- [ ] documentation is added or updated
- [ ] tests are added or updated

PR #10411, opened 2019-11-11
Move constant time RSA padding checks out of libssl

mattcaswell:
In TLSv1.2 a pre-master secret value is passed from the client to the
server encrypted using RSA PKCS1 type 2 padding in a ClientKeyExchange
message. As well as the normal formatting rules for RSA PKCA1 type 2
padding TLS imposes some additional rules about what constitutes a well
formed key. Specifically it must be exactly the right length and
encode the TLS version originally requested by the client (as opposed to
the actual negotiated version) in its first two bytes.

All of these checks need to be done in constant time and, if they fail,
then the TLS implementation is supposed to continue anyway with a random
key (and therefore the connection will fail later on). This avoids
padding oracle type attacks.

This PR implements this within the RSA padding code so that we keep
all the constant time padding logic in one place and moves it out of libssl.

Marked as WIP because this code depends on #10152. Only the last 3 commits of this PR are relevant.


kroeckx:
You might want to read #8446

mattcaswell:
> You might want to read #8446

Thanks that's interesting reading. This PR doesn't attempt to solve the problem of the public API so in a sense this PR is tackling a slightly smaller problem, i.e. it is motivated by the need to remove the constant time padding code out of libssl and into the providers - and in particular the FIPS provider. For FIPS we're not allowed to have this crypto code outside of the module boundary.

However there is clearly a big overlap with what you started in #8446. It seems our thinking is largely aligned judging by your last comment:

> So I think the suggestion is to have something for use in TLS only. Either a new API, or maybe just a padding mode in the current API. I'm currently leaning towards just creating a padding mode and not check the TLS version.

This PR implements a new padding mode. Although I'm not clear what you are saying in the last bit about "not checking the TLS version". If you are referring to the client version check - I think that does need to be done in the constant time code.

kroeckx:
What I would like to see is that the new padding check function
only accepts properly padded input. That is, the caller should
always use BN_bn2binpad(). The padding check function should also
know the expected output size, which is known for TLS. If you
combine those 2 properties, the code can be simplified a lot.

levitte:
A rebase would be nice...

mattcaswell:
> What I would like to see is that the new padding check function
only accepts properly padded input. That is, the caller should
always use BN_bn2binpad(). The padding check function should also
know the expected output size, which is known for TLS. If you
combine those 2 properties, the code can be simplified a lot.

I separated out the TLS padding check into its own function rather than mixing it with the existing one. This new function expects properly padded input and knows the output size as you suggested.

I've also rebased this PR and taken it out of WIP. This is now ready for review.

crypto/err/openssl.txt
```diff
@@ -2490,34 +2490,6 @@ EVP_R_WRAP_MODE_NOT_ALLOWED:170:wrap mode not allowed
 EVP_R_WRONG_FINAL_BLOCK_LENGTH:109:wrong final block length
 EVP_R_XTS_DATA_UNIT_IS_TOO_LARGE:191:xts data unit is too large
 EVP_R_XTS_DUPLICATED_KEYS:192:xts duplicated keys
-KDF_R_BAD_ENCODING:122:bad encoding
```
t8m: Should these kind-of unrelated changes in openssl.txt be in a separate commit? They still can be part of this PR.

crypto/rsa/rsa_pk1.c
```diff
@@ -253,3 +256,120 @@ int RSA_padding_check_PKCS1_type_2(unsigned char *to, int tlen,
 
     return constant_time_select_int(good, mlen, -1);
 }
+
+int rsa_padding_check_PKCS1_type_2_TLS(unsigned char *to, size_t tlen,
+                                       const unsigned char *from, size_t flen,
+                                       int client_version, int alt_version)
```
kroeckx: Please document the function.

crypto/rsa/rsa_pk1.c
```diff
@@ -253,3 +256,120 @@ int RSA_padding_check_PKCS1_type_2(unsigned char *to, int tlen,
...
+    unsigned char rand_premaster_secret[SSL_MAX_MASTER_KEY_LENGTH];
+    unsigned int version_good;
+    int j;
+
+    /*
+     * If these checks fail then either the message in publicly invalid, or
+     * we've been called incorrectly. We can fail immediately.
+     */
+    if (flen < RSA_PKCS1_PADDING_SIZE + SSL_MAX_MASTER_KEY_LENGTH
+            || tlen < SSL_MAX_MASTER_KEY_LENGTH) {
```
kroeckx: I think we wan tlen == SSL_MAX_MASTER_KEY_LENGTH, so fail when tlen != SSL_MAX_MASTER_KEY_LENGTH

crypto/rsa/rsa_pk1.c
```diff
@@ -253,3 +256,120 @@ int RSA_padding_check_PKCS1_type_2(unsigned char *to, int tlen,
...
+
+    /* scan over padding data */
+    found_zero_byte = 0;
+    for (i = 2; i < flen; i++) {
+        unsigned int equals0 = constant_time_is_zero(from[i]);
+
+        zero_index = constant_time_select_int(~found_zero_byte & equals0,
+                                              i, zero_index);
+        found_zero_byte |= equals0;
+    }
```
kroeckx: This can be replaced by:
```
    for (i = 2; i < flen-tlen-1; i++) {
	good &= ~constant_time_is_zero_8(from[i]);
    }
    good &= constant_time_is_zero_8(from[flen-tlen-1]);
```

crypto/rsa/rsa_pk1.c
```diff
@@ -253,3 +256,120 @@ int RSA_padding_check_PKCS1_type_2(unsigned char *to, int tlen,
...
+
+    /*
+     * Skip the zero byte. This is incorrect if we never found a zero-byte
+     * but in this case we also do not copy the message out.
+     */
+    msg_index = zero_index + 1;
+    mlen = flen - msg_index;
+
+    /* We're doing a TLS padding check so we expect an exact size */
+    good &= constant_time_eq(mlen, SSL_MAX_MASTER_KEY_LENGTH);
```
kroeckx: With the above changes, this can all be removed.

crypto/err/openssl.txt
mattcaswell: This actually appears to be caused by issue #10522 fixed by #10549.

mattcaswell: I've removed these extraneous changes which were an artefact of "make update" out of this PR.

crypto/rsa/rsa_pk1.c
mattcaswell: I added a comment at the front describing it.

crypto/rsa/rsa_pk1.c
mattcaswell: I did have that in one local version of the code and then changed my mind. Does it actually matter if the to buffer is *bigger* than actually needed? Not sure why we would fail in that case. It also occurred to me that if you *have* to pass SSL_MAX_MASTER_KEY_LENGTH as tlen - do we actually need tlen at all? Anyway I've left this as is for now.

crypto/rsa/rsa_pk1.c
mattcaswell: Nice! Done.

crypto/rsa/rsa_pk1.c
mattcaswell: Removed.

mattcaswell:
I've updated as per the feedback above. I also had to rebase due to conflicts with master.

crypto/rsa/rsa_pk1.c
```diff
@@ -253,3 +256,122 @@ int RSA_padding_check_PKCS1_type_2(unsigned char *to, int tlen,
 
     return constant_time_select_int(good, mlen, -1);
 }
+
+/*
+ * rsa_padding_check_PKCS1_type_2_TLS() checks and removes the PKCS1 type 2
+ * padding from a decrypted RSA message in a TLS signature. The result is stored
+ * in the buffer pointed to by |to| which should be |tlen| bytes long. |tlen|
+ * must be at least SSL_MAX_MASTER_KEY_LENGTH. The original decrypted message
+ * should be stored in |from| which must be |flen| bytes in length. The TLS
```
kroeckx: Please add something that it should be padded up to the RSA_size(), and so flen == RSA_size().

crypto/rsa/rsa_pk1.c
kroeckx: The simplification of the code that still used tlen relied on it. Note that you don't use tlen anymore in the code, you've replaced it with SSL_MAX_MASTER_KEY_LENGTH. So with the current code it doesn't matter, it can be bigger.

doc/man3/EVP_PKEY_CTX_ctrl.pod
```diff
@@ -352,6 +354,25 @@ B<label>. The return value is the label length. The padding mode
...
 by the library and should not be freed by the caller.
 
+B<RSA_PKCS1_WITH_TLS_PADDING> is used when decrypting an RSA encrypted TLS
+pre-master secret in a TLS ClientKeyExchange message. It is the same as
+RSA_PKCS1_PADDING except that it additionally verifies that the result is the
+correct length and the first two bytes are the protocol version initially
+requested by the client. If the padding checks fail then decryption will still
+appear to succeed but a random TLS premaster secret will be returned instead
+unless the encrypted content is publicly invalid (in which case the decryption
+will fail as normal). This padding mode accepts two parameters which can be set
```
kroeckx: This is hard to parse

doc/man7/provider-asymcipher.pod
```diff
@@ -181,6 +181,13 @@ algorithms:
 
 =item "pad-mode" (B<OSSL_ASYM_CIPHER_PARAM_PAD_MODE>) <integer>
 
+The type of padding to be used. The interpretation of this value will depend
+on the algorithm in use. The default provider understands these RSA padding
+modes: 1 (RSA_PKCS1_PADDING), 2 (RSA_SSLV23_PADDING), 3 (RSA_NO_PADDING),
+4 (RSA_PKCS1_OAEP_PADDING), 5 (RSA_X931_PADDING), 6 (RSA_PKCS1_PSS_PADDING) and
+7 (RSA_PKCS1_WITH_TLS_PADDING). See L<EVP_PKEY_CTX_set_rsa_padding(3)> for
```
kroeckx: Do we want to document the numbers? I think it should just list  the defines.

providers/implementations/asymciphers/rsa_enc.c
```diff
@@ -142,7 +146,13 @@ static int rsa_decrypt(void *vprsactx, unsigned char *out, size_t *outlen,
...
 
-    if (prsactx->pad_mode == RSA_PKCS1_OAEP_PADDING) {
+    if (outsize < len) {
+        ERR_raise(ERR_LIB_PROV, PROV_R_BAD_LENGTH);
+        return 0;
+    }
+
+    if (prsactx->pad_mode == RSA_PKCS1_OAEP_PADDING
+            || prsactx->pad_mode == RSA_PKCS1_WITH_TLS_PADDING) {
         int rsasize = RSA_size(prsactx->rsa);
```
kroeckx: rsasize == len

providers/implementations/asymciphers/rsa_enc.c
```diff
@@ -252,6 +279,14 @@ static int rsa_get_ctx_params(void *vprsactx, OSSL_PARAM *params)
     if (p != NULL && !OSSL_PARAM_set_size_t(p, prsactx->oaep_labellen))
         return 0;
 
+    p = OSSL_PARAM_locate(params, OSSL_ASYM_CIPHER_PARAM_TLS_CLIENT_VERSION);
+    if (p != NULL && !OSSL_PARAM_set_uint(p, prsactx->client_version))
+        return 0;
```
kroeckx: Should that be `p == NULL || !OSSL_PARAM_set_uint`

ssl/statem/statem_srvr.c
```diff
@@ -3000,131 +3000,82 @@ static int tls_process_cke_rsa(SSL *s, PACKET *pkt)
      * (SSL_MAX_MASTER_KEY_LENGTH). Reject overly short RSA keys because
      * their ciphertext cannot accommodate a premaster secret anyway.
      */
-    if (RSA_size(rsa) < SSL_MAX_MASTER_KEY_LENGTH) {
+    if (EVP_PKEY_size(rsa) < SSL_MAX_MASTER_KEY_LENGTH) {
```
kroeckx: That should probably say RSA_PKCS1_PADDING_SIZE + SSL_MAX_MASTER_KEY_LENGTH

ssl/statem/statem_srvr.c
```diff
@@ -3000,131 +3000,82 @@ static int tls_process_cke_rsa(SSL *s, PACKET *pkt)
...
+            || EVP_PKEY_CTX_set_rsa_padding(ctx, RSA_PKCS1_WITH_TLS_PADDING) <= 0) {
         SSLfatal(s, SSL_AD_DECRYPT_ERROR, SSL_F_TLS_PROCESS_CKE_RSA,
-                 ERR_R_INTERNAL_ERROR);
+                 SSL_R_DECRYPTION_FAILED);
         goto err;
     }
 
-    /* Check the padding. See RFC 3447, section 7.2.2. */
+    *p++ = OSSL_PARAM_construct_uint(OSSL_ASYM_CIPHER_PARAM_TLS_CLIENT_VERSION,
+                                     (unsigned int *)&s->client_version);
```
kroeckx: I dislike the *p++. Do we have something in our style guide about it?

kroeckx:
I think this looks good, but I currently don't understand the code about the parameters, so I won't completely review this.

crypto/rsa/rsa_pk1.c
mattcaswell: Done

doc/man3/EVP_PKEY_CTX_ctrl.pod
mattcaswell: I reworded it a little and tried to break it up into two sentences. Hopefully its better.

doc/man7/provider-asymcipher.pod
mattcaswell: This provides part of the public API to providers. Providers don't need to link against libcrypto at all if they don't want to and may not include our headers (or even be written in C - as long as they conform to the C calling convention). For that reason we've tended to document the numbers in the man7 provider documentation (at least we have for params, and if we do it there it makes sense to me to do it here too).

providers/implementations/asymciphers/rsa_enc.c
mattcaswell: I removed rsasize and just used len instead.

providers/implementations/asymciphers/rsa_enc.c
mattcaswell: No. The `OSSL_ASYM_CIPHER_PARAM_TLS_CLIENT_VERSION` parameter value may not have been requested. If it wasn't requested we skip this. We only fail if it was requested and we fail to set its value.

ssl/statem/statem_srvr.c
mattcaswell: Changed.

ssl/statem/statem_srvr.c
mattcaswell: I'm not sure its in our style guide anywhere. However this idiom is *everywhere* we use parameters. This PR is probably not the place to adopt a new style.

mattcaswell:
> I think this looks good, but I currently don't understand the code about the parameters, so I won't completely review this.

Thanks. I've pushed some fixup commits to address your latest feedback.

Since @kroeckx won't be providing the sign off for this, can someone else please review?

crypto/rsa/rsa_pk1.c
```diff
@@ -253,3 +256,123 @@ int RSA_padding_check_PKCS1_type_2(unsigned char *to, int tlen,
...
+ * must be at least SSL_MAX_MASTER_KEY_LENGTH. The original decrypted message
+ * should be stored in |from| which must be |flen| bytes in length and padded
+ * such that |flen == RSA_size()|. The TLS protocol version that the client
+ * originally requested should be passed in |client_version|. Some buggy clients
+ * can exist which use the negotiated version instead of the originally
+ * requested protocol version. If it is necessary to work around this bug then
+ * the negotiated protocol version can be passed in |alt_version|, otherwise 0
+ * should be passed.
+ *
+ * If the passed message is publicly invalid or some other error than can be
```
t8m: s/than/that/

ssl/statem/statem_srvr.c
```diff
@@ -3000,131 +3000,83 @@ static int tls_process_cke_rsa(SSL *s, PACKET *pkt)
...
-                                    sizeof(rand_premaster_secret), 0)) {
+    if (!ssl_generate_master_secret(s, rsa_decrypt,
+                                    SSL_MAX_MASTER_KEY_LENGTH, 0)) {
         /* SSLfatal() already called */
         goto err;
     }
 
     ret = 1;
  err:
     OPENSSL_free(rsa_decrypt);
```
t8m: Should this be OPENSSL_clear_free()?

crypto/rsa/rsa_pk1.c
mattcaswell: Fixed.

ssl/statem/statem_srvr.c
mattcaswell: No. `ssl_generate_master_secret` also cleanses the pre master secret (on success or failure). However, your question prompted me to review this and there is another spot just above here where we do need to cleanse it. I also then discovered some other problems with the length handling in this code. Fixed now.

mattcaswell:
Please take a look at the latest fixups!

mattcaswell:

mattcaswell:
Pushed. Thanks.

crypto/rsa/rsa_pk1.c
```diff
@@ -253,3 +256,123 @@ int RSA_padding_check_PKCS1_type_2(unsigned char *to, int tlen,
...
+        ERR_raise(ERR_LIB_RSA, RSA_R_PKCS_DECODING_ERROR);
+        return -1;
+    }
+
+    /*
+     * Generate a random premaster secret to use in the event that we fail
+     * to decrypt.
+     */
+    if (RAND_priv_bytes(rand_premaster_secret,
+                      sizeof(rand_premaster_secret)) <= 0) {
```
bernd-edlinger: indentation wrong

PR #10412, opened 2019-11-11
CORE: Add a generic callback function type

levitte:
This offers a very generic way to define a callback as well as calling
it.


levitte:
This is an alternative to #10390

levitte:
So, I've made an alternative for #10391 that works with this PR (#10414).  As for #10394 and #10289, I think I'll wait with the work on those until I have a yay or nay here (I sure hope I get a yay).

levitte:
Ping.

mattcaswell (approved):

levitte:
Merged.

10e7216e502916f8e7917ac4f21411fffd1fb882 CORE: Add a generic callback function type

PR #10414, opened 2019-11-11
CORE & PROV: make export of key data leaner through callback, take 2

levitte:
Exporting data from a provider owned domainparams or key is quite an
ordeal, with having to figure out what parameter keys an
implementation supports, call the export function a first time to find
out how large each parameter buffer must be, allocate the necessary
space for it, and call the export function again.

So how about letting the export function build up the key data params
and call back with that?  This change implements exactly such a
mechanism.

-----

This is an alternative to #10391

### DIFF
--- a/crypto/evp/keymgmt_lib.c
+++ b/crypto/evp/keymgmt_lib.c
@@ -219,6 +219,10 @@ const OSSL_PARAM *evp_keymgmt_importdomparam_types(const EVP_KEYMGMT *keymgmt)
     return keymgmt->importdomparam_types();
 }
 
+/*
+ * TODO(v3.0) investigate if we need this function.  'openssl provider' may
+ * be a caller...
+ */
 const OSSL_PARAM *evp_keymgmt_exportdomparam_types(const EVP_KEYMGMT *keymgmt)
 {
     return keymgmt->exportdomparam_types();
@@ -265,6 +269,10 @@ const OSSL_PARAM *evp_keymgmt_importkey_types(const EVP_KEYMGMT *keymgmt)
     return keymgmt->importkey_types();
 }
 
+/*
+ * TODO(v3.0) investigate if we need this function.  'openssl provider' may
+ * be a caller...
+ */
 const OSSL_PARAM *evp_keymgmt_exportkey_types(const EVP_KEYMGMT *keymgmt)
 {
     return keymgmt->exportkey_types();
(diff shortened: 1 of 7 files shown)


mattcaswell:
Travis failures are relevant

levitte:
Darnit

levitte:
Oh wait, that's because #10412 isn't merged yet, and I haven't included it in this PR.
Comments on the code and the idea would still be appreciated.

crypto/evp/keymgmt_lib.c
```diff
@@ -147,69 +87,39 @@ void *evp_keymgmt_export_to_provider(EVP_PKEY *pk, EVP_KEYMGMT *keymgmt,
...
-                int (*exportfn)(void *provctx, OSSL_PARAM params[]) = NULL;
-
-                if (pk->pkeys[j].domainparams != want_domainparams)
-                    continue;
-
-                exportfn = want_domainparams
-                    ? pk->pkeys[j].keymgmt->exportdomparams
-                    : pk->pkeys[j].keymgmt->exportkey;
-
-                paramdefs = pk->pkeys[j].keymgmt->exportkey_types();
```
mattcaswell: With this deletion we never seem to call the `exportkey_types()` keymgmt function anywhere. Similarly we never seem to use the other `[export|import][key|domparams]_types()` functions. Should they be deleted?

mattcaswell (changes requested):
This looks a lot neater. I just have one comment.

crypto/evp/keymgmt_lib.c
levitte: Can we leave that undecided for now? I'm thinking of provider exploration with `openssl provider`, but you may be right.

It will remain on my mind, so post-pone to a cleanup PR to give me time to think? Feel free to hold me to it :wink:

mattcaswell: Could we add a `TODO` in the code as a placeholder so that we do remember to come back and consider it at some point?

levitte: Yes

mattcaswell (approved):
Approved subject to a clean CI run once #10412 is merged.

levitte:
#10412 now being merged, I've rebased this

levitte:
CIs are happy, apart from the Appveyor things that aren't relevant here

levitte:
Merged.

1640d48c5b4ee0a3ff5a2a5015ee17ac163d9cd4 CORE & PROV: make export of key data leaner through callback

PR #10419, opened 2019-11-12
Fix the issue that aes-xxx-siv algorithms failed in OpenSSL speed test

### LINKED ISSUE #10416: OpenSSL failed in speed evp test for all of the aes-xxx-siv algorithms
I found the issue in daily performance test on latest version of OpenSSL.  Run OpenSSL speed test, and it report errors on all of the aes-xxx-siv test cases.  Suppose the new add code cause this issue.
One reproduce example as below: 
Run command : openssl speed -evp aes-128-siv
Then you will see the fail log as below:
"
Doing aes-128-siv for 3s on 16 size blocks:
EVP_CipherInit_ex failure
"

xkqian:
Speed test, aes-siv related cases fail on both x86 and arm.
The return value of siv_init() causes this problem.
Verify it locally, the result is pass.

Fixes #10416

Change-Id: If1a18599f3d0f56f22a1ce4f8f114b8db0f68cca

##### Checklist

### DIFF
--- a/providers/implementations/ciphers/cipher_aes_siv.c
+++ b/providers/implementations/ciphers/cipher_aes_siv.c
@@ -49,9 +49,6 @@ static int siv_init(void *vctx, const unsigned char *key, size_t keylen,
 
     ctx->enc = enc;
 
-    if (iv != NULL)
-        return 0;
-
     if (key != NULL) {
         if (keylen != ctx->keylen) {
             ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_KEY_LENGTH);


xkqian:
Hi all:
I have signed CCLA for cooperation.
For ICLA, I have sent email to legal@opensslfoundation.org on 11, Oct 2019 with my sign agreement and the needed material, but didn't receive the response.
Can someone help me check what's problem with my ICLA? My email is xiaokang.qian@arm.com. thanks a lot

providers/implementations/ciphers/cipher_aes_siv.c
```diff
@@ -50,7 +50,7 @@ static int siv_init(void *vctx, const unsigned char *key, size_t keylen,
     ctx->enc = enc;
 
     if (iv != NULL)
-        return 0;
+        return 1;
```
levitte: So if the IV is set, the key gets entirely ignored, and there's nothing indicating that.
The sane change would perhaps be to ignore the IV for real, i.e to remove the check entirely, but I'm not at all sure that's a viable solution either.

The thing I do notice, though, is that we don't raise an error here, just return 0 with no further explanation, so the best thing may be to actually raise one.

xkqian: But if simple return 0, the call of EVP_CipherInit_ex() in speed_main will think it has some problems, and report error.   I Debug and find that when it want to set key in EVP_CipherInit_ex(), it will set IV to NULL. if not, leave it there, seems do nothing, but return 0 will trigger one error.  I test it with evp_test and it didn't affect the result.

xkqian: Hi Levitte:
Can someone help me about the ICLA? I have sent out the email about 1 month, still have no reply, not sure what's the problem of my email.

mattcaswell: Do you know the date you sent it? I had a look through my spam folder but couldn't find it. Perhaps resend?

slontis: The check was just there to pick up someone trying to set an iv which it does not use.
If we are going to ignore the iv being set then do what @levitte said - 
i.e. remove the if (iv != NULL) return X;

xkqian: > Do you know the date you sent it? I had a look through my spam folder but couldn't find it. Perhaps resend?

I sent it on on 11, Oct 2019(Beijing TIme zone), anyway, I will resend it to  legal@opensslfoundation.org.

xkqian: > The check was just there to pick up someone trying to set an iv which it does not use.
> If we are going to ignore the iv being set then do what @levitte said -
> i.e. remove the if (iv != NULL) return X;

Then, should we simple remove the "if (iv != NULL) return X;" ?  Seems the common call flow like this: EVP_CipherInit_ex() with (key = NULL) to set iv, EVP_CipherInit_ex() with (iv = NULL)to set key, then the following other calls. If we remove the check, somebody need take care that the first time call EVP_CipherInit_ex(),   input key be NULL to avoid unneccessary processing of key.   If everybody happy about this solution, I will remove the check and commit another patch

xkqian: > > Do you know the date you sent it? I had a look through my spam folder but couldn't find it. Perhaps resend?
> 
> I sent it on on 11, Oct 2019(Beijing TIme zone), anyway, I will resend it to [legal@opensslfoundation.org](mailto:legal@opensslfoundation.org).
Hi mattcaswell:
I Have resent out the ICLA request, please help it, thanks.

xkqian:
I have updated the code as the comment from you, please help take a look.

xkqian:
@mattcaswell  Did you receive my  ICLA request?

mattcaswell:
Close/reopen to kick CLA bot.

mattcaswell:
> Can someone help me about the ICLA? I have sent out the email about 1 month, still have no reply, not sure what's the problem of my email.

Thanks for sending your CLA. I also found your original request. We actually replied and added you to the CLA database on the day you sent it to us, so I'm not sure why you didn't get our reply. Anyway, it seems there was a problem with the entry which I've now fixed, and the CLA hold has been removed.

providers/implementations/ciphers/cipher_aes_siv.c
levitte: Regarding patterns, the *intended* pattern is that if the caller gives an iv (non-null), then we use the IV, and if the caller gives a key, then we use that key. They can both be given in the same `EVP_CipherInit_ex` call.

So this makes me wonder about SIV. Why did someone give it an IV? Is it an acceptable behaviour from the backend to simply ignore it? Wouldn't the caller rather know that it was wrong to give an IV to a SIV cipher? Otherwise, will the caller think that the IV had an actual impact? In that case, what will the caller do with the IV? Will they do the classic thing and somehow pass it on to the recipient of the encrypted message? What's the recipient going to do with it? 

All of this can be summarised with "know the cipher you use, or there will be hell to pay" 

It's possible that we should look back at other ciphers that don't take an IV, such as, dare I say it, any ECB mode cipher...

xkqian: Actually, it depend on the design of speed test.  It always firstly call EVP_CipherInit_ex to set iv(iv!=NULL, key=NULL) then call EVP_CipherInit_ex() set key(iv=NULL, key!=NULL) whether the mode is ecb,siv, cbc, etc.  AS this:
`for (testnum = 0; testnum < size_num; testnum++) {
                print_message(names[D_EVP], save_count, lengths[testnum],
                              seconds.sym);

                for (k = 0; k < loopargs_len; k++) {
                    loopargs[k].ctx = EVP_CIPHER_CTX_new();
                    if (loopargs[k].ctx == NULL) {
                        BIO_printf(bio_err, "\nEVP_CIPHER_CTX_new failure\n");
                        exit(1);
                    }
                    if (!EVP_CipherInit_ex(loopargs[k].ctx, evp_cipher, NULL,
                                           NULL, **iv**, decrypt ? 0 : 1)) {
                        BIO_printf(bio_err, "\nEVP_CipherInit_ex failure\n");
                        ERR_print_errors(bio_err);
                        exit(1);
                    }

                    EVP_CIPHER_CTX_set_padding(loopargs[k].ctx, 0);

                    keylen = EVP_CIPHER_CTX_key_length(loopargs[k].ctx);
                    loopargs[k].key = app_malloc(keylen, "evp_cipher key");
                    EVP_CIPHER_CTX_rand_key(loopargs[k].ctx, loopargs[k].key);
                    if (!EVP_CipherInit_ex(loopargs[k].ctx, NULL, NULL,
                                           **loopargs[k].key**, NULL, -1)) {
                        BIO_printf(bio_err, "\nEVP_CipherInit_ex failure\n");
                        ERR_print_errors(bio_err);
                        exit(1);
                    }`

So, for ECB, which doesn't need iv, it also call EVP_CipherInit_ex() to set iv, but the backend in cipher_common.c:cipher_generic_init_internal() ignore the iv settting and return 1 directly. Code as below:
`static int cipher_generic_init_internal(PROV_CIPHER_CTX *ctx,
                                        const unsigned char *key, size_t keylen,
                                        const unsigned char *iv, size_t ivlen,
                                        int enc)
{
    ctx->enc = enc ? 1 : 0;

    **if (iv != NULL && ctx->mode != EVP_CIPH_ECB_MODE**) {
        if (!cipher_generic_initiv(ctx, iv, ivlen))
            return 0;
    }
    if (key != NULL) {
        if ((ctx->flags & EVP_CIPH_VARIABLE_LENGTH) == 0) {
            if (keylen != ctx->keylen) {
                ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_KEYLEN);
                return 0;
            }
        } else {
            ctx->keylen = keylen;
        }
        return ctx->hw->init(ctx, key, ctx->keylen);
    }
    return 1;
}`
 For siv, it also call EVP_CipherInit_ex() to set iv, our backend exist in cipher_aes_siv.c:siv_einit(), per my understanding, it should return 1 as what ECB does.

Seems this bakend design is compatible with the unified callflows and even some errors, which means: you can follow the set iv and set key flows, backend can respond properly. This is my investigation and my guess, please help take a look.

levitte: And me, I went back to remind myself how 1.1.1 EVP_CipherInit_ex deals with the given IV, and indeed, unless the mode *expects* and IV, the iv parameter seems to be completely ignored.

So after all my philosophical question, it does seems like removing this check is indeed the appropriate thing to do.
Thanks for indulging me.

xkqian:
> > Can someone help me about the ICLA? I have sent out the email about 1 month, still have no reply, not sure what's the problem of my email.
> 
> Thanks for sending your CLA. I also found your original request. We actually replied and added you to the CLA database on the day you sent it to us, so I'm not sure why you didn't get our reply. Anyway, it seems there was a problem with the entry which I've now fixed, and the CLA hold has been removed.

I have gone through my email and spam again, strange and interesting, I still can't find the reply.  maybe something wrong with our email agent or somewhere. Anyway, so, have I had the ICLA now without any other further actions to do? Or shall I follow up other steps to achieve it?

mattcaswell:
>  Anyway, so, have I had the ICLA now without any other further actions to do? Or shall I follow up other steps to achieve it?

No further steps are required.

xkqian:
> > Anyway, so, have I had the ICLA now without any other further actions to do? Or shall I follow up other steps to achieve it?
> 
> No further steps are required.

Got it, thanks.

xkqian:
Hi:
Seems the check fail "continuous-integration/appveyor/pr — AppVeyor build failed" is a common issue(CA has some problems, and other PRs still have such issue), and is not triggered by my patch. if has no other concern, can you help merge my patch? thanks.

xkqian:
What extra steps should I do before this patch merged? Should I make another push to make the check "continuous-integration/appveyor/pr" passed ? or just wait for be merged?

mattcaswell:
> What extra steps should I do before this patch merged? Should I make another push to make the check "continuous-integration/appveyor/pr" passed ? or just wait for be merged?

It just waiting for merge.

Ping @levitte, @slontis.

slontis:
Thanks for the fix. Merged to master.

PR #10420, opened 2019-11-12
Fix L<EVP_MD_CTX_set_pkey_ctx> links

moehuster:
##### Checklist


mattcaswell:
> Can you confirrm your approval for both master and 1.1.1?

That was supposed to be directed to @levitte...

levitte:
I agree with both branches

t8m:
Merged to OpenSSL_1_1_1-stable as ca7c982821.
Merged to master as 3c86a2b5c8 after fixing trivial conflict.

PR #10424, opened 2019-11-12
Fix no-cmac and no-camellia

mattcaswell:
Guard two tests that depend on CMAC and Camellia so that we don't fail if those algorithms are not available.


mattcaswell:
Pushed. Thanks.

PR #10425, opened 2019-11-12
Fix no-blake2

mattcaswell:
If blake2 has been disabled then the legacy_blake2.c has no content,
which was causing a compilation failure. Therefore we add the macro
NON_EMPTY_TRANSLATION_UNIT.

### DIFF
--- a/crypto/evp/build.info
+++ b/crypto/evp/build.info
@@ -17,7 +17,7 @@ SOURCE[../../libcrypto]=$COMMON\
         e_aes_cbc_hmac_sha1.c e_aes_cbc_hmac_sha256.c e_rc4_hmac_md5.c \
         e_chacha20_poly1305.c \
         pkey_mac.c exchange.c \
-        legacy_sha.c legacy_md5_sha1.c legacy_blake2.c
+        legacy_sha.c legacy_md5_sha1.c
 
 IF[{- !$disabled{md2} -}]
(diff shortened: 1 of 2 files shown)


slontis:
I would prefer these to get fixed in build.info
See https://github.com/openssl/openssl/pull/10407

levitte:
I agree with @slontis

mattcaswell:
> I would prefer these to get fixed in build.info

Sure. Done.

richsalz:
Consistency is important.  Does it make sense to change most similar ifdef's to do it in the build.info file?

richsalz:
An issue with doing it in the build.info file is that this makes it harder for other projects to just pick up and re-use OpenSSL sources in their project.  I prefer doings things in the standalone approach everywhere, but please be consistent.

levitte:
"standalone" is sooo 1.0.x 😉

I prefer the build.info way. The `NON_EMPTY_TRANSLATION_UNIT` macro is an ugly kludge... Always been.

richsalz:
You can say it's a kludge, but you can also say that about many things done for portability. :)

At any rate, the main point is please do things CONSISTENTLY.

richsalz:
I created #10434 to address the consistency issue.

mattcaswell:
Pushed. Thanks!

PR #10426, opened 2019-11-12
A few fixes in the EVP_KDF-KB and EVP_KDF-KRB5KDF manpages

t8m:
The context and label is info and salt respectively - fix
the example in KB manpage for that.

There are some typos and bug in EVP_KDF_derive call in the
KRB5KDF example.

##### Checklist


t8m:
Appveyor failure is unrelated.

t8m:
Merged to master as a90311fead

PR #8882, opened 2019-05-06
Add a .pragma directive for configuration files

levitte:
Currently added pragma:

.pragma dollarid:on

This allows dollar signs to be a keyword character unless it's
followed by a opening brace or parenthesis.

Fixes #8207

##### Checklist

- [ ] documentation is added or updated
- [ ] tests are added or updated

PR #10428, opened 2019-11-12
Fix an uninitialised read in conf_def.c

mattcaswell:
PR #8882 added a new field to the CONF structure. Unfortunately this
structure was created using OPENSSL_malloc() instead of OPENSSL_zalloc() and
therefore the new field was not created in an initialised state. Therefore
when we came to read it for the first time we got an uninitialised read.

### DIFF
--- a/crypto/conf/conf_def.c
+++ b/crypto/conf/conf_def.c
@@ -107,7 +107,7 @@ static CONF *def_create(CONF_METHOD *meth)
 {
     CONF *ret;
 
-    ret = OPENSSL_zalloc(sizeof(*ret));
+    ret = OPENSSL_malloc(sizeof(*ret));
     if (ret != NULL)
         if (meth->init(ret) == 0) {
             OPENSSL_free(ret);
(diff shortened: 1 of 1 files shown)


crypto/conf/conf_def.c
```diff
@@ -123,7 +123,6 @@ static int def_init_default(CONF *conf)
 
     conf->meth = &default_method;
     conf->meth_data = (void *)CONF_type_default;
-    conf->data = NULL;
```
levitte: Why?

crypto/conf/conf_def.c
```diff
@@ -136,7 +135,6 @@ static int def_init_WIN32(CONF *conf)
 
     conf->meth = &WIN32_method;
     conf->meth_data = (void *)CONF_type_win32;
-    conf->data = NULL;
```
levitte: Why?

crypto/conf/conf_def.c
mattcaswell: Because now we're using zalloc they're already initialised to NULL.

levitte: So this is only called directly after def_new()?

mattcaswell: > So this is only called directly after def_new()?

Ah! You're right! It might be called through other routes. Fixup pushed.

levitte:
I've labeled this "urgent" because this fixes one of the Appveyor build failures we're currently seeing.

mattcaswell:
Pushed. Thanks!

PR #10430, opened 2019-11-12
Add confdump to .gitignore

mattcaswell:
We added a new executable to the test directory which didn't match the
existing gitignore rules, so we add it explicitly.


mattcaswell:
Pushed! Thanks.

PR #10431, opened 2019-11-13
Add OSSL_RET_UNSUPPORTED

levitte:
It's been complained numerous times that OpenSSL uses "magic values" with not much of an explanation, except manual pages mentioning it a little here and there.

-2 is such a value, which is used fairly consistently throughout the code, to signify that a function call with its combination of parameters is unsupported or unrecognised.  This allows a caller to use fallbacks, or otherwise treat this return value in special ways.

There are some API functions that return -2 for other reasons, and those are left alone.  Typically, this applies to comparison functions, where -1 is a valid return value, and -2 is therefore used to signify that an error of some kind has occurred.  This *could* be interpreted as unsupported, but it's unclear whether that is useful or not.

An oddity was discovered as part of this work; some BIO functions return -2 when the BIO itself hasn't been initialised.  That strikes me as odd, isn't that a "normal" error?

It also turns out that our previous `EVP_CTRL_RET_UNSUPPORTED` is incorrect, as there are backend ctrls that do return -2 to signify that the command number is unsupported.

### DIFF
--- a/crypto/bio/bio_lib.c
+++ b/crypto/bio/bio_lib.c
@@ -254,7 +254,7 @@ static int bio_read_intern(BIO *b, void *data, size_t dlen, size_t *readbytes)
 
     if ((b == NULL) || (b->method == NULL) || (b->method->bread == NULL)) {
         BIOerr(BIO_F_BIO_READ_INTERN, BIO_R_UNSUPPORTED_METHOD);
-        return -2;
+        return OSSL_RET_UNSUPPORTED;
     }
 
     if ((b->callback != NULL || b->callback_ex != NULL) &&
@@ -327,7 +327,7 @@ static int bio_write_intern(BIO *b, const void *data, size_t dlen,
 
     if ((b->method == NULL) || (b->method->bwrite == NULL)) {
         BIOerr(BIO_F_BIO_WRITE_INTERN, BIO_R_UNSUPPORTED_METHOD);
-        return -2;
+        return OSSL_RET_UNSUPPORTED;
     }
 
     if ((b->callback != NULL || b->callback_ex != NULL) &&
@@ -391,7 +391,7 @@ int BIO_puts(BIO *b, const char *buf)
 
     if ((b == NULL) || (b->method == NULL) || (b->method->bputs == NULL)) {
         BIOerr(BIO_F_BIO_PUTS, BIO_R_UNSUPPORTED_METHOD);
-        return -2;
+        return OSSL_RET_UNSUPPORTED;
     }
 
     if (b->callback != NULL || b->callback_ex != NULL) {
@@ -436,7 +436,7 @@ int BIO_gets(BIO *b, char *buf, int size)
 
     if ((b == NULL) || (b->method == NULL) || (b->method->bgets == NULL)) {
         BIOerr(BIO_F_BIO_GETS, BIO_R_UNSUPPORTED_METHOD);
-        return -2;
+        return OSSL_RET_UNSUPPORTED;
     }
 
     if (size < 0) {
@@ -516,7 +516,7 @@ long BIO_ctrl(BIO *b, int cmd, long larg, void *parg)
 
     if ((b->method == NULL) || (b->method->ctrl == NULL)) {
         BIOerr(BIO_F_BIO_CTRL, BIO_R_UNSUPPORTED_METHOD);
-        return -2;
+        return OSSL_RET_UNSUPPORTED;
     }
 
     if (b->callback != NULL || b->callback_ex != NULL) {
@@ -544,7 +544,7 @@ long BIO_callback_ctrl(BIO *b, int cmd, BIO_info_cb *fp)
     if ((b->method == NULL) || (b->method->callback_ctrl == NULL)
             || (cmd != BIO_CTRL_SET_CALLBACK)) {
         BIOerr(BIO_F_BIO_CALLBACK_CTRL, BIO_R_UNSUPPORTED_METHOD);
-        return -2;
+        return OSSL_RET_UNSUPPORTED;
     }
 
     if (b->callback != NULL || b->callback_ex != NULL) {
--- a/crypto/bio/bio_lib.c
+++ b/crypto/bio/bio_lib.c
@@ -264,7 +264,7 @@ static int bio_read_intern(BIO *b, void *data, size_t dlen, size_t *readbytes)
 
     if (!b->init) {
         BIOerr(BIO_F_BIO_READ_INTERN, BIO_R_UNINITIALIZED);
-        return -2;
+        return -2;               /* shouldn't this be -1? */
     }
 
     ret = b->method->bread(b, data, dlen, readbytes);
@@ -337,7 +337,7 @@ static int bio_write_intern(BIO *b, const void *data, size_t dlen,
 
     if (!b->init) {
         BIOerr(BIO_F_BIO_WRITE_INTERN, BIO_R_UNINITIALIZED);
-        return -2;
+        return -2;               /* shouldn't this be -1? */
     }
 
     ret = b->method->bwrite(b, data, dlen, written);
@@ -402,7 +402,7 @@ int BIO_puts(BIO *b, const char *buf)
 
     if (!b->init) {
         BIOerr(BIO_F_BIO_PUTS, BIO_R_UNINITIALIZED);
-        return -2;
+        return -2;               /* shouldn't this be -1? */
     }
 
     ret = b->method->bputs(b, buf);
@@ -452,7 +452,7 @@ int BIO_gets(BIO *b, char *buf, int size)
 
     if (!b->init) {
         BIOerr(BIO_F_BIO_GETS, BIO_R_UNINITIALIZED);
-        return -2;
+        return -2;               /* shouldn't this be -1? */
     }
 
     ret = b->method->bgets(b, buf, size);
--- a/crypto/cms/cms_env.c
+++ b/crypto/cms/cms_env.c
@@ -67,7 +67,7 @@ int cms_env_asn1_ctrl(CMS_RecipientInfo *ri, int cmd)
     if (pkey->ameth == NULL || pkey->ameth->pkey_ctrl == NULL)
         return 1;
     i = pkey->ameth->pkey_ctrl(pkey, ASN1_PKEY_CTRL_CMS_ENVELOPE, cmd, ri);
-    if (i == -2) {
+    if (i == OSSL_RET_UNSUPPORTED) {
         CMSerr(CMS_F_CMS_ENV_ASN1_CTRL,
                CMS_R_NOT_SUPPORTED_FOR_THIS_KEY_TYPE);
         return 0;
--- a/crypto/cms/cms_env.c
+++ b/crypto/cms/cms_env.c
@@ -267,6 +267,7 @@ int CMS_RecipientInfo_ktri_get0_signer_id(CMS_RecipientInfo *ri,
     return cms_SignerIdentifier_get0_signer_id(ktri->rid, keyid, issuer, sno);
 }
 
+/* -1 is a valid return value, so we use -2 for error here */
 int CMS_RecipientInfo_ktri_cert_cmp(CMS_RecipientInfo *ri, X509 *cert)
 {
     if (ri->type != CMS_RECIPINFO_TRANS) {
@@ -440,6 +441,7 @@ static int cms_RecipientInfo_ktri_decrypt(CMS_ContentInfo *cms,
 
 /* Key Encrypted Key (KEK) RecipientInfo routines */
 
+/* -1 is a valid return value, so we use -2 for error here */
 int CMS_RecipientInfo_kekri_id_cmp(CMS_RecipientInfo *ri,
                                    const unsigned char *id, size_t idlen)
 {
--- a/crypto/evp/evp_lib.c
+++ b/crypto/evp/evp_lib.c
@@ -54,7 +54,7 @@ int EVP_CIPHER_param_to_asn1(EVP_CIPHER_CTX *c, ASN1_TYPE *type)
         case EVP_CIPH_CCM_MODE:
         case EVP_CIPH_XTS_MODE:
         case EVP_CIPH_OCB_MODE:
-            ret = -2;
+            ret = OSSL_RET_UNSUPPORTED;
             break;
 
         default:
@@ -90,11 +90,11 @@ int EVP_CIPHER_param_to_asn1(EVP_CIPHER_CTX *c, ASN1_TYPE *type)
             OPENSSL_free(der);
         }
     } else {
-        ret = -2;
+        ret = OSSL_RET_UNSUPPORTED;
     }
 
  err:
-    if (ret == -2)
+    if (ret == OSSL_RET_UNSUPPORTED)
         EVPerr(EVP_F_EVP_CIPHER_PARAM_TO_ASN1, ASN1_R_UNSUPPORTED_CIPHER);
     else if (ret <= 0)
         EVPerr(EVP_F_EVP_CIPHER_PARAM_TO_ASN1, EVP_R_CIPHER_PARAMETER_ERROR);
@@ -136,7 +136,7 @@ int EVP_CIPHER_asn1_to_param(EVP_CIPHER_CTX *c, ASN1_TYPE *type)
         case EVP_CIPH_CCM_MODE:
         case EVP_CIPH_XTS_MODE:
         case EVP_CIPH_OCB_MODE:
-            ret = -2;
+            ret = OSSL_RET_UNSUPPORTED;
             break;
 
         default:
@@ -157,10 +157,10 @@ int EVP_CIPHER_asn1_to_param(EVP_CIPHER_CTX *c, ASN1_TYPE *type)
             OPENSSL_free(der);
         }
     } else {
-        ret = -2;
+        ret = OSSL_RET_UNSUPPORTED;
     }
 
-    if (ret == -2)
+    if (ret == OSSL_RET_UNSUPPORTED)
         EVPerr(EVP_F_EVP_CIPHER_ASN1_TO_PARAM, EVP_R_UNSUPPORTED_CIPHER);
     else if (ret <= 0)
         EVPerr(EVP_F_EVP_CIPHER_ASN1_TO_PARAM, EVP_R_CIPHER_PARAMETER_ERROR);
--- a/crypto/evp/evp_lib.c
+++ b/crypto/evp/evp_lib.c
@@ -403,7 +403,7 @@ int EVP_CIPHER_CTX_iv_length(const EVP_CIPHER_CTX *ctx)
 
     params[0] = OSSL_PARAM_construct_size_t(OSSL_CIPHER_PARAM_IVLEN, &v);
     rv = evp_do_ciph_ctx_getparams(ctx->cipher, ctx->provctx, params);
-    if (rv == EVP_CTRL_RET_UNSUPPORTED)
+    if (rv == OSSL_RET_UNSUPPORTED)
         goto legacy;
     return rv != 0 ? (int)v : -1;
     /* TODO (3.0) Remove legacy support */
@@ -486,7 +486,7 @@ int EVP_CIPHER_CTX_num(const EVP_CIPHER_CTX *ctx)
     params[0] = OSSL_PARAM_construct_uint(OSSL_CIPHER_PARAM_NUM, &v);
     ok = evp_do_ciph_ctx_getparams(ctx->cipher, ctx->provctx, params);
 
-    return ok != 0 ? (int)v : EVP_CTRL_RET_UNSUPPORTED;
+    return ok != 0 ? (int)v : -1;
 }
 
 int EVP_CIPHER_CTX_set_num(EVP_CIPHER_CTX *ctx, int num)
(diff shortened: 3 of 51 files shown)


mspncp:
Referring to this discussion and also to https://github.com/openssl/openssl/pull/10394#discussion_r344717668:

> I'm used to -2 meaning that the call is unsupported, that's a value that has been injected in my brain for a long time with things OpenSSL. If you grep for it in existing documentation, you'll see that it's fairly pervasive, at least there (I noticed just now that the documentation for `EVP_PKEY_print_{public,private,params}` says they return that value, but they don't in actual fact do so).
> 
> So a name. Do you have a suggestion? We do have `EVP_CTRL_RET_UNSUPPORTED`, but these functions aren't ctrls, so it seems like an inappropriate name.
> 
> The reason for -2 is, if I remember history right, that there are functions where 0 is a valid success value (for example returned length, and zero length is valid), so -1 was made to mean error, and when an indicator that a call isn't supported (i.e. the caller should try a fallback), the natural choice came to -2.

It makes sense to reuse the same value for the 'unsupported' case wherever possible. And it is a good idea to have a searchable identifier for that error return. However, I would go even further and add an entire general purpose set of reusable return codes instead of just a single one. Borrowing an [idea from COM](https://docs.microsoft.com/en-us/windows/win32/seccrypto/common-hresult-values) (about names, not about values) I would use the `OSSL_S_` prefix for successful return values and the `OSSL_E_` prefix for error returns.


Here some examples:


For functions which return an integer value with boolean semantics, but also error codes if necessary:

```C
#define OSSL_S_TRUE          1
#define OSSL_S_FALSE         0

#define OSSL_E_FAILED       -1        /* generic failure */
#define OSSL_E_UNSUPPORTED  -2        /* feature not supported */
...                                   /* other error codes? */
```

(Callers can check for `ret == OSSL_TRUE`, `ret == OSSL_FALSE` , ...  or `ret < 0`.)


For functions functions returning errno-like status codes:

```
#define OSSL_S_OK            0
```

(Callers can check for `ret == OSSL_S_OK` , `ret != OSSL_S_OK` or explicit errno codes.)

crypto/cms/cms_env.c
```diff
@@ -267,6 +267,7 @@ int CMS_RecipientInfo_ktri_get0_signer_id(CMS_RecipientInfo *ri,
     return cms_SignerIdentifier_get0_signer_id(ktri->rid, keyid, issuer, sno);
 }
 
+/* -1 is a valid return value, so we use -2 for error here */
```
mspncp: Why not just return `OSSL_E_UNSUPPORTED`? Is this name really unapt in this context? If it is, you can figure out a different name ([look here](https://docs.microsoft.com/en-us/windows/win32/com/com-error-codes-1) for inspirations). Returning a different negative value, because the return value is not properly documented anyway ;-)

```
=head1 RETURN VALUES

...

CMS_RecipientInfo_ktri_cert_cmp() and CMS_RecipientInfo_kekri_cmp() return 0
for a successful comparison and non zero otherwise.
```

crypto/cms/cms_env.c
```diff
@@ -440,6 +441,7 @@ static int cms_RecipientInfo_ktri_decrypt(CMS_ContentInfo *cms,
 
 /* Key Encrypted Key (KEK) RecipientInfo routines */
 
+/* -1 is a valid return value, so we use -2 for error here */
```
mspncp: Same here, etc...

crypto/cms/cms_env.c
levitte: The comparison result is clearly modelled after the likes of strcmp, even though the documentation is crap. If we change the return values, we risk breaking applications unnecessarily.

levitte:
I would rather avoid `_S_`, `_E_` and similar. Too close to the naming of reason codes, which these returns are not. These returns are more of a rough classification.

crypto/evp/evp_lib.c
```diff
@@ -281,7 +281,7 @@ int EVP_CIPHER_block_size(const EVP_CIPHER *cipher)
     params[0] = OSSL_PARAM_construct_size_t(OSSL_CIPHER_PARAM_BLOCK_SIZE, &v);
     ok = evp_do_ciph_getparams(cipher, params);
 
-    return ok != 0 ? (int)v : EVP_CTRL_RET_UNSUPPORTED;
+    return ok != 0 ? (int)v : -1;
```
slontis: I guess these ones are ok.. In 1_1_1 I guess it always returned something non negative.

slontis (approved):

slontis (changes requested):
Needs a rebase...

crypto/evp/evp_lib.c
levitte: Yeah, I take the common pattern to follow from 1.1.1's `EVP_CIPHER_CTX_iv_length()`

levitte:
I've been busy elsewhere, but intend to come back to complete this...  I got a tentative approval, and I have a suggestion from @mspncp to make this more feature complete, so I'm kinda stumped as to how far to take this.  Someone help me decide?  Making this feature complete *could* be a matter of another PR.

t8m:
I would not try to invent additional error-signalling mechanism here to be consumed by calling applications. This particular error return value is useful also internally and so it makes sense to have it but trying to add more just in case they would be handy in future looks to me like a slippery slope.

mspncp:
> I would not try to invent additional error-signalling mechanism 

I was not suggesting to add new return _values_, or even create a new new mechanism. 
My intention was to make existing practice more explicit by replacing magic constants with literate names.

There are a lot of functions in the code with `int` as return type, where the semantics of the return value is a boolean value. Wouldn't it be better to return `OSSL_TRUE` and `OSSL_FALSE` instead of `1` and `0`?.  Also, in some cases  '-1' returned as a third value to indicate an error IIRC. Those values could be replaced by something like `OSSL_FAILED`. 

(If @levitte dislikes the `_S_` and `_E_` infixes, no problem. Let's just drop them from https://github.com/openssl/openssl/pull/10431#issuecomment-553343970.)

mspncp:
>  Making this feature complete as @mspncp suggests could be a matter of another PR.

At least we should _decide_ now whether we want to add a more complete set of symbolic constants or not and do it immediately in this pr. Changing existing magic constants in the code will be a longer process which can happen continually over a period of time and in different pr's. If we postpone the decision, it just means that we will have to restart the discussion or that it will never happen.

mattcaswell:
> There are a lot of functions in the code with int as return type, where the semantics of the return value is a boolean value. Wouldn't it be better to return OSSL_TRUE and OSSL_FALSE instead of 1 and 0?.

I think this is actually a really good idea and would make things much more explicit. Probably not one for this PR though.

mspncp:
My updated suggestion for this pull request would be to simply add five defines instead of just one:

```C
#define OSSL_TRUE          1
#define OSSL_FALSE         0

#define OSSL_FAILED       -1   /* generic failure */
#define OSSL_UNSUPPORTED  -2   /* feature not supported */

#define OSSL_SUCCESS       0   /* for functions with errno-like return codes */
```



Callers of 'boolean' functions can continue to test
```C
    if (function()) {
    }

    /* respectively */

    if (!function()) {
    }
```
and nobody will be forced to write
```C
    if (function() == OSSL_TRUE) {
    }

    /* respectively */

    if (function() == OSSL_FALSE) {
    }
```

_Note: these if conditions obviously don't work for boolean functions which also return an error value. Callers need to check something like_

```C
    switch(function()) {
    case OSSL_TRUE:  ...
    case OSSL_FALSE: ...
    default:
    }
```
_But this has nothing to do with the introduction of symbolic names._

t8m:
Then I have a question - will we start rewriting documentation to mention these macros instead of numbers? And isn't that an API break in a theoretical sense - because in theory it means we can in future change the values to other numbers so application writers should start using these macros everywhere to be protected against this theoretical API break.

I mean if we used these constants from start I would have nothing against it but I do not really get what would be the improvement if we introduce them now.
Also where they are supposed to be used (only in new code, or everywhere) and will they be introduced into 1.1.1 branch because otherwise the cherry-picks will be impossible?

I am not a fan of this as it really looks to me rather like a busywork.

Improvement related to the return values would be to have somehow a clear indication of whether a function can return negative value for error or not, but I do not know how to achieve this - apart from making all the functions that cannot return negative value to return some unsigned value type. However that would be an API break as well I think.

BTW - do we really have any API call that uses the libc semantics of 0 return meaning success? That would be a major inconsistency but I do not know of any such function in OpenSSL API.

mattcaswell:
> Then I have a question - will we start rewriting documentation to mention these macros instead of numbers? And isn't that an API break in a theoretical sense - because in theory it means we can in future change the values to other numbers so application writers should start using these macros everywhere to be protected against this theoretical API break.

I see no reason why we shouldn't mention those macros in documentation, and I absolutely do not agree that it would be an API break to do so. If we ever changed their values then *that* would be a major API break. I don't expect us to do that though.

> I mean if we used these constants from start I would have nothing against it but I do not really get what would be the improvement if we introduce them now.

I think there are *lots* of functions where the return value semantics is not clear. Is a function *supposed* to be returning boolean values or are other values acceptable? Does a function most return 1 for success and 0 for failure but occasionally return -1 in some situations? I think using macros would help document the *intent* much better.

> Also where they are supposed to be used (only in new code, or everywhere) and will they be introduced into 1.1.1 branch because otherwise the cherry-picks will be impossible?

I don't see the need to introduce them  in 1.1.1. I also don't agree with an argument that says we shouldn't make improvements to the code because that makes cherry-picking harder. If we need to make improvements then we should do so.

> BTW - do we really have any API call that uses the libc semantics of 0 return meaning success? That would be a major inconsistency but I do not know of any such function in OpenSSL API.

Yes. Some examples:

SSL_CTX_set_tlsext_use_srtp(), SSL_set_tlsext_use_srtp(), SSL_CTX_set_alpn_protos(), SSL_set_alpn_protos(), UI_UTIL_read_pw_string(), UI_UTIL_read_pw(), UI_set_result(), UI_dup_user_data(), UI_process(), UI_get_result_length(), UI_method_set_prompt_constructor(), BIO_seek(). Also the BIO_ctrls BIO_CTRL_DGRAM_SCTP_ADD_AUTH_KEY, BIO_CTRL_DGRAM_SCTP_NEXT_AUTH_KEY, BIO_CTRL_DGRAM_SCTP_AUTH_CCS_RCVD.

BIO_reset() returns 0 for success for file BIOs only!

levitte:
I think that historically, most functions had the semantics that 0 or above was success while a negative number was an error. This follows the POSIX IO semantics (which I suspect is the reason for most BIO function returns).

These days, we have an exasperating mish-mash of those semantics and boolean semantics... And quite frankly, for quite a lot of functions, the older semantics make quite a lot of sense (and is more consistent)... for example, the length returning functions, where à negative number works as an error indicator.

t8m:
> I think that historically, most functions had the semantics that 0 or above was success while a negative number was an error. This follows the POSIX IO semantics (which I suspect is the reason for most BIO function returns).

Really? Then it must be much older history than I remember. Because if I remember correctly for example the EVP API was always on the -1/0/1 return values with 0 not meaning success but things like "expectable error" - i.e. signature verification error because of data corruption or similar things. Matt's examples do not look like something from old OpenSSL history.

> I think using macros would help document the intent much better.

It is not clear to me how this would help except if we started consistently using TRUE and FALSE macros for return values only in functions where negative return is not expected and kept using numbers (or completely different macros) elsewhere.

mattcaswell:
> It is not clear to me how this would help except if we started consistently using TRUE and FALSE macros for return values only in functions where negative return is not expected and kept using numbers (or completely different macros) elsewhere.

That's exactly what I would expect. If a function returns TRUE or FALSE as return values I would expect it to never return anything else. Functions that return -1/0/1 might (for example) return ERROR, FAIL and SUCCESS.

tmshort:
The OpenSSL API has a lot of quirks, and I appreciate the attempts to clean these things up. For those APIs that have inconsistent returns (e.g. 0 = success, rather than 1 = success), would it make sense to introduce (in another PR) new APIs that follow the standard? I could see  many other API type (e.g. setting/getting callbacks) that could be made to follow a single standard.

Of course, by saying "standard", someone is going to have to write up such a standard and get it approved...

levitte:
I don't think it's a wise idea to introduce a whole new API *now*. We have enough going on as it is.
(I really don't look forward to the standardisation "discussion", BTW)

mspncp:
As a ompromise you could just rename your define to OSSL_UNSUPPORTED und leave the other defines for a separate pr.

richsalz:
I'll just mention that "return -1 or the size" is hard to make work when we want to convert size-related things to the unsigned "size_t"

mspncp:
> -2 is such a value, which is used fairly consistently throughout the code, to signify that a function call with its combination of parameters is unsupported or unrecognised. This allows a caller to use fallbacks, or otherwise treat this return value in special ways.

FWIW: I just happened to stumble over a case where `-1` is returned instead of `-2` to indicate that a feature is not supported.

**RAND(3)**

```
RETURN VALUES
       RAND_bytes() and RAND_priv_bytes() return 1 on success, -1 if not supported by the current RAND method, or 0 on other failure. The error code can be obtained by
       ERR_get_error(3).
```

levitte:
Yup, we're not 100% consistent. That's why I said "*fairly* consistently".

What seems to have happened historically is that for each subsystem, someone looked at the return values that were possible at the time, and simply took the next one down to mean "unsupported". In most cases, -1 was already taken as the generic error indicator (making -2 the next choice), but there were subsystems where 0 was the generic error indicator (making -1 the next choice).
Something like that...

levitte:
I've mulled this over for some time, and I think that while it's useful to have a macro for -2, the other error indicating returns are too diverse to get anywhere near consistency.  We're basically stuck with the fact that depending on the function, `== 0`, or `<= 0`, or `< 0` is an error.  It sucks.

crypto/bio/bio_lib.c
```diff
@@ -337,7 +337,7 @@ static int bio_write_intern(BIO *b, const void *data, size_t dlen,
 
     if (!b->init) {
         BIOerr(BIO_F_BIO_WRITE_INTERN, BIO_R_UNINITIALIZED);
-        return -2;
+        return -2;               /* shouldn't this be -1? */
```
mspncp: Do you intend to still do anything about this line in this pr? If you just want to leave the comment, maybe you should prefix it with a TODO?   `/* TODO: shouldn't this be -1? */`

crypto/bio/bio_lib.c
```diff
@@ -402,7 +402,7 @@ int BIO_puts(BIO *b, const char *buf)
 
     if (!b->init) {
         BIOerr(BIO_F_BIO_PUTS, BIO_R_UNINITIALIZED);
-        return -2;
+        return -2;               /* shouldn't this be -1? */
```
mspncp: Same here.

crypto/bio/bio_lib.c
```diff
@@ -452,7 +452,7 @@ int BIO_gets(BIO *b, char *buf, int size)
 
     if (!b->init) {
         BIOerr(BIO_F_BIO_GETS, BIO_R_UNINITIALIZED);
-        return -2;
+        return -2;               /* shouldn't this be -1? */
```
mspncp: and here

crypto/evp/evp_lib.c
```diff
@@ -403,7 +403,7 @@ int EVP_CIPHER_CTX_iv_length(const EVP_CIPHER_CTX *ctx)
 
     params[0] = OSSL_PARAM_construct_size_t(OSSL_CIPHER_PARAM_IVLEN, &v);
     rv = evp_do_ciph_ctx_getparams(ctx->cipher, ctx->provctx, params);
-    if (rv == EVP_CTRL_RET_UNSUPPORTED)
+    if (rv == OSSL_RET_UNSUPPORTED)
```
mspncp: Here you are changing `EVP_CTRL_RET_UNSUPPORTED=-1` to `OSSL_RET_UNSUPPORTED=-2`? 

_(several locations)_

crypto/evp/evp_utils.c
```diff
@@ -18,7 +18,7 @@
 #include "evp_local.h"
 
 /*
- * EVP_CTRL_RET_UNSUPPORTED = -1 is the returned value from any ctrl function
+ * OSSL_RET_UNSUPPORTED = -2 is the returned value from any ctrl function
```
mspncp: Regardless of whether `EVP_CTRL_RET_UNSUPPORTED` had the wrong value or not, don't you break applications by returning a different value now?

crypto/trace.c
```diff
@@ -96,7 +96,7 @@ static long trace_ctrl(BIO *channel, int cmd, long argl, void *argp)
     default:
         break;
     }
-    return -2;                   /* Unsupported */
+    return OSSL_RET_UNSUPPORTED; /* Unsupported */
```
mspncp: This comment can be removed now. (It's not present elsewhere either.)

mspncp (dismissed):
The raw text replacements `-2 -> OSSL_RET_UNSUPPORTED` look good, although it really made me dizzy to look at so many identical hunks.

What makes me worry are the locations, where you change return values from `-1` to `-2`. Are you sure you won't break any applications by doing that?

levitte:
> What makes me worry are the locations, where you change return values from -1 to -2. Are you sure you won't break any applications by doing that?

Those are locations where -2 is returned in 1.1.1 (you just have to dive down into the backends to discover it), so I'm actually *restoring* the previous functionality.

crypto/bio/bio_lib.c
levitte: It's meant as an open question, which I would like to have an answer on here.

In *my* opinion, this should be changed to -1, and in the common case, callers check if the return is `<= 0` or `< 0`, so this should be a safe change.

levitte:
Evidence from 1.1.1, EVP_DigestInit_ex *expects* -2 from `EVP_PKEY_CTX_ctrl()`:

https://github.com/openssl/openssl/blob/d4f094a04a3a89319f5bf9277e79bd8d77ba10fa/crypto/evp/digest.c#L141-L144

levitte:
Evidence from 1.1.1, `e_aes.c`'s `ctrl()`:

https://github.com/openssl/openssl/blob/d4f094a04a3a89319f5bf9277e79bd8d77ba10fa/crypto/evp/m_md5_sha1.c#L57-L58

levitte:
Evidence from `p_lib.c`:

https://github.com/openssl/openssl/blob/d4f094a04a3a89319f5bf9277e79bd8d77ba10fa/crypto/evp/p_lib.c#L661-L666

levitte:
...  I think that should be enough...

richsalz:
In my view, changing any of the return functions breaks the ABI promise.

mspncp:
Ok, it seems like the root problem is that two different return values were mapped to the same \#define in commit e870791a4d6aea3a0275396bd01da629cb6f4ac8. While you are restoring the original value here (and in other locations):

https://github.com/openssl/openssl/commit/e870791a4d6aea3a0275396bd01da629cb6f4ac8#diff-63ab5e427422da7c0b8a22fe56fad0dbL929-R929

you are nevertheless changing it there (and in other locations):

https://github.com/openssl/openssl/commit/e870791a4d6aea3a0275396bd01da629cb6f4ac8#diff-63ab5e427422da7c0b8a22fe56fad0dbL1042-R1042

I don't have enough background knowledge to decide whether this is a breaking change or not. Your arguments sound plausible and I am tempted to believe you, but my believe does not justify an approval. So I'll have to leave the approval to others.

levitte:
:-(

So we now discover that *cipher* controls return -1 for diverse unsupported cases, while we have *digests* ***possibly*** using -2, and a quick run through our current EVP_PKEY_METHOD and EVP_PKEY_ASN1_METHOD ctrl functions seem to use -2 (it seems like the EVP_PKEY stuff is the most consistent).
Meanwhile, I find that BIO ctrls use 0 for unsupported cases...  no biggie, it's not like there are fallback cases there.

Either way, suffice to say that thing are less consistent, numerically speaking, than I initially thought.
That has me conclude that we will have to live with these "magic numbers", and stay aware of what API uses what values.

I give this PR up.

mspncp:
> I give up this PR.

Sorry about that.  One question remains however: what to do about the return values which were changed by @slontis in e870791a4d6aea3a0275396bd01da629cb6f4ac8 (intentionally or not)?

levitte:
> Sorry about that. One question remains however: what to do about the return values which were changed by @slontis in e870791 (intentionally or not)?

Changed *compared to what*?  Like I said above, it seems that all the *cipher* backends in 1.1.1 return -1 to mark something unsupported, and since e870791a4d6aea3a0275396bd01da629cb6f4ac8 is all about *ciphers*, that change seems correct (it's a correction of something that we got wrong in master).
One might *possibly* argue that the change in `crypto/evp/evp_utils.c` is dubious, but considering how very few digest backends are returning a negative value at all, one might as well argue that *any negative value* from their ctrl functions should be taken as "unsupported".

EVP_PKEY functions are a whole different story and shouldn't be mixed up with *cipher* and *digest* functionality (well, except that digest functionality is mixed into the lot, but still), and very consistently use -2 for unsupported (there are some backends that use -1 as well, but that seems to be with other types of errors for the cases I've looked at).

richsalz:
Rats, too bad.  Nice try tho.

PR #10432, opened 2019-11-13
Delete recently documented functions from missingcrypto

t8m:
Subject says it all


levitte:
Maybe someone should teach find-doc-nits to say something when a symbol in the "missing" files actually is documented...

levitte:
Ci failures not relevant as far as I can see

t8m:
Merged as f3d0c92b42 to master.

PR #9162, opened 2019-06-14
Fix a race condition in supported groups and ciphers handling

mattcaswell:
In TLSv1.3 the supported groups can be negotiated each time a handshake
occurs, regardless of whether we are resuming or not. We should not store
the supported groups information in the session because session objects
can be shared between multiple threads and we can end up with race
conditions. For most users this won't be seen because, by default, we
use stateless tickets in TLSv1.3 which don't get shared. However if you
use SSL_OP_NO_TICKET (to get stateful tickets in TLSv1.3) then this can
happen.

The answer is to move the supported the supported group information into
the SSL object instead.

PR #10441, opened 2019-11-13
Fix a race condition in SNI handling

### LINKED ISSUE #10210: Mutability of TLS 1.3 SSL_SESSIONs and thread-safety
This is mostly a follow up from my comment at https://github.com/openssl/openssl/pull/6563#issuecomment-541906622 to attempt to determine whether there's an error in my understanding/expectations or the code.  (Both cases would require code changes, just different ones, so this isn't a pull request.)

During the development of our TLS 1.3 support, the only session resumption mechanism that we supported was the stateless ticket design that's described repeatedly in RFC 8446.  Accordingly, our libssl behavior involved a new SSL_SESSION being created for each incoming TLS 1.3 connection, regardless of whether or not it was a resumption handshake or a full handshake.  This crept into our design assumptions repeatedly, e.g. in e3743355e840dadff0b09582942b7a90db607894 and 1c4aa31d79821dee9be98e915159d52cc30d8403.  (The fact that these are the only instances easily detectable via commit messages and are both authored by me does not escape my notice.)  When we changed our resumption behavior in #6563 to allow the use of the stateful session cache, the invariant that TLS 1.3 connections get new SSL_SESSIONS was removed, and so then TLS 1.3 processing is subject to the same thread-safety concerns that all other connections are: the SSL_SESSION object is immutable (other than refcounts) after it's in the cache.  However, we don't currently adhere to the new invariant; the first example that comes to mind is in final_server_name(), where we store the received hostname in the session for non-resumptions and for TLS 1.3 connections.  Until somewhat recently we also did this for supported_groups and peer_ciphers, which was fixed in #9176 by moving those fields out of the SSL_SESSION and into the SSL.

Given the above, is the right conclusion to draw that we want to stick with the new invariant of SSL_SESSION immutability and remove any lingering vestiges of the previous invariant that TLS 1.3 connections always get a new SSL_SESSION?

kaduk:
As was done for ciphers, supported groups, and EC point formats in
https://github.com/openssl/openssl/pull/9162, only write the negotiated
SNI hostname value to the session object when not resuming, even for
TLS 1.3 resumptions.  Otherwise, when using a stateful session cache
(as is done by default when 0-RTT data is enabled), we can have multiple
SSLs active using the same in-memory session object, which leads to
double-frees and similar race conditions in the SNI handler prior
to this commit.

Fortunately, since draft-ietf-tls-tls13-22, there is no requirement
that the SNI hostname be preserved across TLS 1.3 resumption, and thus
not a need to continually update the session object with the "current"
value (to be used when producing session tickets, so that the subsequent
resumption can be checked against the current value).  So we can just
relax the logic and only write to the session object for initial handshakes.
This still leaves us in a somewhat inconsistent state, since if the SNI value
does change across handshakes, the session object will continue to record
the initial handshake's value, even if that bears no relation to the
current handshake.  The current SSL_get_servername() implementation
prefers the value from the session if s->hit, but a more complete fix
for that and related issues is underway in
https://github.com/openssl/openssl/pull/10018; there is no need to wait
for the complete fix for SNI name handling in order to close the
race condition and avoid runtime crashes.

Fixes #10210

### DIFF
--- a/ssl/statem/extensions.c
+++ b/ssl/statem/extensions.c
@@ -949,7 +949,7 @@ static int final_server_name(SSL *s, unsigned int context, int sent)
      */
     if (s->server) {
         /* TODO(OpenSSL1.2) revisit !sent case */
-        if (sent && ret == SSL_TLSEXT_ERR_OK && (!s->hit || SSL_IS_TLS13(s))) {
+        if (sent && ret == SSL_TLSEXT_ERR_OK && !s->hit) {
             /* Only store the hostname in the session if we accepted it. */
             OPENSSL_free(s->session->ext.hostname);
             s->session->ext.hostname = OPENSSL_strdup(s->ext.hostname);


kaduk:
Unsurprisingly, if you cherry-pick the test from #10018 on top of this, sslapitest's test_servername fails, since SSL_get_servername() returns NULL.

kaduk:
If I'm reading the logs properly, the CI issues are spurious.

kaduk:
Pushed to master and 1.1.1 (and semi-accidentally pushed a rebase to this PR's branch); closing.

PR #10443, opened 2019-11-14
Fix Use after free when copying cipher ctx

### LINKED ISSUE #10438: Use after free after copying cipher ctx
The function `cipher_hw_aes_initkey` (and possibly there are other similar occurrences), initialises a ctx like this:
````
static int cipher_hw_aes_initkey(PROV_CIPHER_CTX *dat,
                                 const unsigned char *key, size_t keylen)
{
    int ret;
    PROV_AES_CTX *adat = (PROV_AES_CTX *)dat;
    AES_KEY *ks = &adat->ks.ks;

    dat->ks = ks;
````
So `dat->ks` is actually a pointer to somewhere else in the structure.

However, when we "dup" a ctx we do it like this:

````
static void *aes_dupctx(void *ctx)
{
    PROV_AES_CTX *in = (PROV_AES_CTX *)ctx;
    PROV_AES_CTX *ret = OPENSSL_malloc(sizeof(*ret));

    if (ret == NULL) {
        ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
        return NULL;
    }
    *ret = *in;

    return ret;
}
````

So, now the value of `ret->ks` is pointing to somewhere inside the `in` structure. If `in` subsequently gets freed and you use the `ret` ctx you get a use after free.

### PR DESCRIPTION
**Pull request by slontis**
Fixes #10438 
issue found by clusterfuzz/ossfuzz

The dest was getting a copy of the src structure which contained a pointer that should point to an offset inside itself - because of the copy it was pointing to the original structure.

The setup for a ctx is mainly done by the initkey method in the PROV_CIPHER_HW structure. Because of this it makes sense that the structure should also contain a copyctx method that is use to resolve any pointers that need to be setup.

A dup_ctx has been added to the cipher_enc tests in evp_test. It does a dup after setup and then frees the original ctx. This detects any floating pointers in the duplicated context that were pointing back to the freed ctx.

##### Checklist

### DIFF
--- a/providers/common/include/prov/ciphercommon.h
+++ b/providers/common/include/prov/ciphercommon.h
@@ -68,6 +68,7 @@ struct prov_cipher_ctx_st {
 struct prov_cipher_hw_st {
     int (*init)(PROV_CIPHER_CTX *dat, const uint8_t *key, size_t keylen);
     PROV_CIPHER_HW_FN *cipher;
+    void (*copyctx)(PROV_CIPHER_CTX *dst, const PROV_CIPHER_CTX *src);
 };
 
 OSSL_OP_cipher_encrypt_init_fn cipher_generic_einit;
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

File: providers/common/include/prov/ciphercommon.h [hunk 1]
```diff
@@ -233,6 +234,16 @@ static int cipher_hw_##NAME##_##MODE##_cipher(PROV_CIPHER_CTX *ctx,            \
     return 1;                                                                  \
 }
 
+#define IMPLEMENT_CIPHER_HW_COPYCTX(name, CTX_TYPE)                            \
+static void name (PROV_CIPHER_CTX *dst, const PROV_CIPHER_CTX *src)            \
```
mattcaswell: Space after `name` here

mattcaswell:
Hmm. Appveyor test results look strange - but they don't *look* related.

providers/common/include/prov/ciphercommon.h
slontis: NIT fixed.. I did that so I didnt have to figure out if ## was required (it isnt required :))

slontis:
Merged to master. Thanks..

PR #9562, opened 2019-08-09
 Verify that an EVP_CIPHER_CTX_set_key_length() call has worked

mattcaswell:
Not all ciphers will implement the OSSL_CIPHER_PARAM_KEYLEN parameter for setting, so if we attempt to set it we should check that it worked afterwards.

PR #10449, opened 2019-11-14
EVP_CIPHER_CTX_set_keylen should not succeed if a bad keylen is passed

mattcaswell:
EVP_CIPHER_CTX_set_keylen() was succeeding even though a bad key length
is passed to it. This is because the set_ctx_params() were all accepting
this parameter and blindly changing the keylen even though the cipher did
not accept a variable key length. Even removing this didn't entirely
resolve the issue because set_ctx_params() functions succeed even if
passed a parameter they do not recognise.

This should fix various issues found by OSSfuzz/Cryptofuzz.

Note that this is an alternative to, and more complete version of #9562 which this replaces.


mattcaswell:
@guidovranken

slontis:
I think this now means that all ciphers with the EVP_CIPH_VARIABLE_LENGTH need a set_ctx function.. i.e blowfish, rc4, rc5, rc4_hmac_md5

crypto/evp/evp_enc.c
```diff
@@ -1030,13 +1030,27 @@ int EVP_CIPHER_CTX_set_key_length(EVP_CIPHER_CTX *c, int keylen)
     OSSL_PARAM params[2] = { OSSL_PARAM_END, OSSL_PARAM_END };
```
levitte: Move this line to be inside the non-legacy block.

levitte:
So if I understand @slontis correctly, this PR needs completing?

levitte:
Travis failure relevant

crypto/evp/evp_enc.c
mattcaswell: Done

mattcaswell:
> I think this now means that all ciphers with the EVP_CIPH_VARIABLE_LENGTH need a set_ctx function.. i.e blowfish, rc4, rc5, rc4_hmac_md5.

Ah - right of course. `rc4_hmac_md5` is actually already correct. I missed these because there are no tests which cover non-default key lengths for these ciphers. Or at least, there is for rc5 but it is not enabled by default.

I've made the code a bit more generic and made the necessary changes for these other ciphers. I've also added some self-generated test vectors with non-default key lengths.

Please take another look.

test/recipes/30-test_evp_data/evpciph.txt
```diff
@@ -2462,3 +2462,62 @@ AAD = 8008315ebf2e6fe020e8f5eb
 Tag = 3615b7f90a651de15da20fb6
 Plaintext = f57af5fd4ae19562976ec57a5a7ad55a5af5c5e5c5fdf5c55ad57a4a7272d57262e9729566ed66e97ac54a4a5a7ad5e15ae5fdd5fd5ac5d56ae56ad5c572d54ae54ac55a956afd6aed5a4ac562957a9516991691d572fd14e97ae962ed7a9f4a955af572e162f57a956666e17ae1f54a95f566d54a66e16e4afd6a9f7ae1c5c55ae5d56afde916c5e94a6ec56695e14afde1148416e94ad57ac5146ed59d1cc5
 Ciphertext = ff78128ee18ee3cb9fb0d20726a017ff67fbd09d3a4c38aa32f6d306d3fdda378e459b83ed005507449d6cd981a4c1e3ff4193870c276ef09b6317a01a2283206ae4b4be0d0b235422c8abb00122410656b75e1ffc7fb49c0d0c5d6169aa7623610579968037aee8e83fc26264ea866590fd620aa3c0a5f323d953aa7f8defb0d0d60ab5a9de44dbaf8eae74ea3ab5f30594154f405fd630aa4c4d5603efdfa1
+
```
slontis: These might be better in a seperate file with a entry on 30-test_evp.t

my @bffiles = qw( evpciph_bf.txt );
push @defltfiles, @bffiles unless disabled("bf");

mattcaswell:
> Apart from seperating the blowfish test vectors into a seperate file., this looks good..

Done in new fixup commit. Please take a look.

mattcaswell:
Pushed. Thanks.

PR #10452, opened 2019-11-15
PEM: constify PEM_write_ routines

levitte:
There's no reason why the object to be written, or the key string
given by the caller should be non-const.

This makes the `IMPLEMENT_PEM_..._const` and `DECLARE_PEM_..._const`
macros superfluous, so we keep them around but mark them deprecated.

In all places where `IMPLEMENT_PEM_..._const` and `DECLARE_PEM_..._const`
are used, they are replaced with the corresponding macros without
`_const`.


levitte:
This will help #10394, as the PEM hooks can become much cleaner

mattcaswell (approved):

levitte:
Merged.

de0799b0fc845869d775520382b4e7f9995732e5 PEM: constify PEM_write_ routines

PR #10454, opened 2019-11-15
s390x assembly pack: fix bn_mul_comba4

p-steuer:
looks like it is never called though: Only place where its not #if 0 -ed is in bn_mul_recursive which is inside #ifdef BN_RECURSION which is never defined as far as i can tell.

### DIFF
--- a/crypto/bn/asm/s390x.S
+++ b/crypto/bn/asm/s390x.S
@@ -511,7 +511,7 @@ bn_mul_comba4:
 	lghi	zero,0
 
 	mul_add_c(0,0,c1,c2,c3);
-	stg	c1,0*8(%r3)
+	stg	c1,0*8(%r2)
 	lghi	c1,0
 
 	mul_add_c(0,1,c2,c3,c1);


paulidale:
If I'm mistaken, please let me know and I'll try to dig a lot deeper....

p-steuer:
At that point c1 holds the least significant word of the product, which should be written to the leftmost part of the output buffer (pointed to by r2). Instead it was written to the leftmost part of the first factor (pointed to by r3) leading to wrong results of the multiplication. I cant tell how this happened, but it i guess its most likely a typo.

p-steuer:
Id also fix it for 1.1.1, please indicate if you approve.

paulidale:
Thanks for the explanation.

p-steuer:
merged.

master: https://github.com/openssl/openssl/commit/97a986f78289fef71bf8778dc4763458e983750c
1.1.1: https://github.com/openssl/openssl/commit/62c20887be277df27b287451615c90515784a02b

PR #5580, opened 2018-03-10
DRBG: Use the EVP layer to do AES encryption

kroeckx:
This is WIP because there are still style issues (line too long) in it, and I want to properly check errors. But it passes the regression tests.

To generate 1000000000 bytes of random data with "openssl rand", I need this amount of time:
1.1.0: 20.233s
HEAD: 12.944s
This PR: 3.024s

Issue #5559, opened 2018-03-08
Lack of performances compared to 1.1.0


EmericBr:
Hi All,

Benchmark of my soft results in lack of performances upgrading to latest master commit.

Using openssl-v1.1.0 my soft handle 24k TPS.

Upgrading to latest master i observe only 14k TPS

A loss of 40% of performances.

I've started a bisect, what i see is that loss is progressiv.

I've currently identified one commit: 32bda2b2e4900308cb025020d8c8692e1d3c2ba9

It is responsible of 19k -> 15k TPS.

R,
Emeric

mattcaswell:
Ouch that's a big loss in one commit. @kroeckx, @mspncp perhaps we should rethink that change?

@EmericBr what happens if you disable TLSv1.3?

mspncp:
It seems like we are not ready for a AES-256-CTR DRBG yet. Since a security level of 128bit should be enough for most use cases, it's probably best to revert the commit for the moment.

BTW:  IIRC, the AES-256-CTR DRBG is implemented entirely in software. (The code was taken mostly unmodified from the FIPS 2.0 object module). There was an attempt by @p-steuer to use platform specific aes support (#4530). This would probably help improving performance in the future. The pr however is not ready yet, since it was requested that hardware support should be based on the EVP API.

richsalz:
+1 to revering the commit.

kroeckx:
I planned to write an other PR that switches it to EVP.

20928ff635772182fe8696d618a19f9101a5370f might be an other cause
of slowdowns by using the additional data.

EmericBr:
@mattcaswell , i've forced to tlsv1.2 using no-tls13. So my benchmark use tlsv1.2.

I continue to bissect to find if there is an over commit responsible of the lack of performances. Reverting the identified one it remains a loss of 20% of performances compared to 1.1.0.

mattcaswell:
>  it remains a loss of 20% of performances compared to 1.1.0.

Probably switching the DRBG to use EVP as suggested by @kroeckx will help.

EmericBr:
@mattcaswell I'm lauching a test reverting that one to. I keep you informed

richsalz:
Also look at OPENSSL_cleanse

richsalz:
[...]

PR #4530, opened 2017-10-13
aes ctr_drbg: use all platform specific aes support

p-steuer:
Let aes-ctr drbg's generate operation use all available platform specific aes support.

I tested this on little and big endian (x86 and s390x). The third patch is quite big since is has all the cavs test data. I can remove it, just wanted to put it here temporary, in case somebody wants to do some testing on more platforms.

Some pre- and post-patch perf data from my thinkpad:

    type             16 bytes     64 bytes    256 bytes   1024 bytes   8192 bytes  16384 bytes
    rand             32840.74k    77676.84k   117831.68k   132720.64k   140705.79k   140771.76k
    rand             85196.83k   337101.99k  1086201.86k  2433860.27k  3830319.79k  4126512.47k

- [x] tests are added or updated

PR #10456, opened 2019-11-15
WIP: AESNI based DRBG CTR

raja-ashok:
Current DRBG CTR rand performs each block by block, changing that to use AESNI based multi block CTR operation gives better performance on rand generation.This inturn helps in improving some of the ECDSA curves (P224, P256) sign operation which invokes random nonce generation.

Improvement in performance measured using `speed` app are listed below

Operations | Base Perf | Perf with Changes
---------- | --------- | -----------------
 16 bytes rand gen | 2 MB/s | 17 MB/s
 64 bytes rand gen | 9 MB/s | 71 MB/s
 256 bytes rand gen | 28 MB/s | 271 MB/s
 1024 bytes rand gen | 52 MB/s | 888 MB/s
 P224 sign | 2346 op/s | 2441 op/s
 P224 verify | 2858 op/s | 2942 op/s
 P256 sign | 27487 op/s | 32717 op/s
 P256 verify | 10422 op/s | 10492 op/s
 P384 sign | 927 op/s | 938 op/s
 P384 verify | 1242 op/s | 1219 op/s
 P521 sign | 394 op/s | 394 op/s
 P521 verify | 553 op/s | 551 op/s

This performance is measured on a ubuntu 18.04 x86_64 PC.

### DIFF
--- a/crypto/rand/drbg_ctr.c
+++ b/crypto/rand/drbg_ctr.c
@@ -12,6 +12,8 @@
 #include <openssl/crypto.h>
 #include <openssl/err.h>
 #include <openssl/rand.h>
+#include "crypto/modes.h"
+#include "crypto/ciphermode_platform.h"
 #include "internal/thread_once.h"
 #include "rand_local.h"
 
@@ -30,6 +32,32 @@ static void inc_128(RAND_DRBG_CTR *ctr)
         *p = c;
         if (c != 0) {
             /* If we didn't wrap around, we're done. */
+            break;
+        }
+    }
+}
+
+static size_t inc_8_by_value(unsigned char *p, size_t value)
+{
+    size_t result;
+
+    result = *p + value;
+    *p = result % 256;
+    return result / 256;
+}
+
+static void inc_128_by_value(RAND_DRBG_CTR *ctr, size_t value)
+{
+    int i;
+    unsigned char *p = &ctr->V[15];
+    unsigned int carry = 0, next_carry;
+
+    for (i = 0; i < 16; i++, p--) {
+        next_carry = inc_8_by_value(p, value & 0xFF);
+        if (carry != 0)
+            next_carry += inc_8_by_value(p, carry);
+        /* Break when both carry and value becomes zero */
+        if ((carry = next_carry) == 0 && (value = value >> 8) == 0) {
             break;
         }
     }
@@ -72,9 +100,12 @@ __owur static int ctr_BCC_block(RAND_DRBG_CTR *ctr, unsigned char *out,
     for (i = 0; i < 16; i++)
         out[i] ^= in[i];
 
-    if (!EVP_CipherUpdate(ctr->ctx_df, out, &outlen, out, AES_BLOCK_SIZE)
-        || outlen != AES_BLOCK_SIZE)
-        return 0;
+    if (ctr->block128 != NULL)
+        ctr->block128(out, out, &ctr->ks_df);
+    else
+        if (!EVP_CipherUpdate(ctr->ctx_df, out, &outlen, out, AES_BLOCK_SIZE)
+            || outlen != AES_BLOCK_SIZE)
+            return 0;
     return 1;
 }
 
@@ -218,6 +249,53 @@ __owur static int ctr_df(RAND_DRBG_CTR *ctr,
     return 1;
 }
 
+__owur static int ctr_aesni_df(RAND_DRBG_CTR *ctr,
+                         const unsigned char *in1, size_t in1len,
+                         const unsigned char *in2, size_t in2len,
+                         const unsigned char *in3, size_t in3len)
+{
+    static unsigned char c80 = 0x80;
+    size_t inlen;
+    unsigned char *p = ctr->bltmp;
+
+    if (!ctr_BCC_init(ctr))
+        return 0;
+    if (in1 == NULL)
+        in1len = 0;
+    if (in2 == NULL)
+        in2len = 0;
+    if (in3 == NULL)
+        in3len = 0;
+    inlen = in1len + in2len + in3len;
+    /* Initialise L||N in temporary block */
+    *p++ = (inlen >> 24) & 0xff;
+    *p++ = (inlen >> 16) & 0xff;
+    *p++ = (inlen >> 8) & 0xff;
+    *p++ = inlen & 0xff;
+
+    /* NB keylen is at most 32 bytes */
+    *p++ = 0;
+    *p++ = 0;
+    *p++ = 0;
+    *p = (unsigned char)((ctr->keylen + 16) & 0xff);
+    ctr->bltmp_pos = 8;
+    if (!ctr_BCC_update(ctr, in1, in1len)
+        || !ctr_BCC_update(ctr, in2, in2len)
+        || !ctr_BCC_update(ctr, in3, in3len)
+        || !ctr_BCC_update(ctr, &c80, 1)
+        || !ctr_BCC_final(ctr))
+        return 0;
+    /* Set up key K */
+    if (aesni_set_encrypt_key(ctr->KX, ctr->keylen * 8, &ctr->ks) < 0)
+        return 0;
+    /* X follows key K */
+    ctr->block128(ctr->KX + ctr->keylen, ctr->KX, &ctr->ks);
+    ctr->block128(ctr->KX, ctr->KX + 16, &ctr->ks);
+    if (ctr->keylen != 16)
+        ctr->block128(ctr->KX + 16, ctr->KX + 32, &ctr->ks);
+    return 1;
+}
+
 /*
  * NB the no-df Update in SP800-90A specifies a constant input length
  * of seedlen, however other uses of this algorithm pad the input with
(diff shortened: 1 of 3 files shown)


mattcaswell:
Why doesn't the existing CTR implementation use AESNI anyway? It should be using the fastest available implementation for the platform.

Adding a separate AESNI DRBG variant seems like completely the wrong approach to me.

p-steuer:
I tried to do the same optimization (use platform specific aes-ctr support instead of platform specific plain aes) for all platforms long time ago: https://github.com/openssl/openssl/pull/4530

It was rejected back then and the PR stalled since then, for the reasons, check my PR.

mattcaswell:
It seems to me that the DRBG code should use the EVP interface and not call low level implementation functions. If EVP isn't performing as well as it could we should look at why that is and see what we can do to improve it.

p-steuer:
> If EVP isn't performing as well as it could we should look at why that is and see what we can do to improve it.

Its not an EVP vs low level thing.

Its using AES-CTR vs (simple/blockwise) AES. There are platforms which have dedicated support for the former.

If the opinion has changed in the meantime, I would suggest doing it for all platforms, as in my original proposal.

raja-ashok:
> Why doesn't the existing CTR implementation use AESNI anyway? It should be using the fastest available implementation for the platform.

Existing DRBG CTR creates cipher context of `ECB` type and random generator calls `aesni_ecb_encrypt` for every block, instead of `aesni_ctr32_encrypt_blocks`.



> Adding a separate AESNI DRBG variant seems like completely the wrong approach to me.

Correct, Currently I added as separate DRBG method mainly to compare the differences. Need to make it as one.

mattcaswell:
> Existing DRBG CTR creates cipher context of ECB type and random generator calls aesni_ecb_encrypt for every block, instead of aesni_ctr32_encrypt_blocks.

So this seems to be the cause of the problem. I see we are doing this:

https://github.com/openssl/openssl/blob/93863030867ad0331bf7f199f8a266ad99ee4cf2/crypto/rand/drbg_ctr.c#L375-L391

If those `EVP_CIPHER_fetch` calls were changed to fetch `AES-*-*CTR` instead, and the drbg_ctr implementation modified to use those rather than trying to manage the ctr operation itself then it should automatically gain the performance benefits of an optimised implementation (where one exists for the platform).

I wonder why ECB is being used here? Is there some limitation of the EVP interface that prevents us using the `AES-*-CTR` EVP implementations here for this use case?

raja-ashok:
> If those EVP_CIPHER_fetch calls were changed to fetch AES-*-CTR

Along with this, can we change block wise call in `drbg_ctr_generate()` to multi block call ?

But we are currently passing CTR counter as input block in every call, that we will not be able to do easily. I feel mainly for this we would have used ECB instead of CTR cipher.

p-steuer:
> I wonder why ECB is being used here? Is there some limitation of the EVP interface that prevents us using the AES-*-CTR EVP implementations here for this use case?

No, you simpley need plain AES operations to implement the DRBG. Even with the CTR optimization you will still need some to do a single block AES encryption at times.

mattcaswell:
> Along with this, can we change block wise call in drbg_ctr_generate() to multi block call ?

Well, we can't change the arguments to that function very easily (because it implements a public function in RAND_DRBG_METHOD). But we can change its implementation anyway we see fit. If it makes sense to do multi-blocks in one go then I see no reason not to. Are there any downsides?

> But we are currently passing CTR counter as input block in every call, that we will not be able to do easily. I feel mainly for this we would have used ECB instead of CTR cipher.

Maybe I'm just not understanding what you're saying, or I'm not familiar enough with the drbg implementation. Surely if we swap to use a CTR cipher we don't need to pass the ctr with every call? That's all managed internally to the implementation of the cipher.

> No, you simply need plain AES operations to implement the DRBG.

Why is this?

> Even with the CTR optimization you will still need to do a single block AES encryption at times.

Again why?

p-steuer:
Well, look at all DRBG operations except generate e.g., update. I mean one *could* always implement a single AES encryption using CTR by using input as IV and encrypt a 0 buffer instead ... but that would just be weird.

kroeckx:
> I wonder why ECB is being used here?

It's because the counter mode used by the DRBG is not provided by
the EVP interface. It is not CTR32. Using ECB and implementing the
correct counter in the DRBG was the easiest thing to do.

mattcaswell:
> It's because the counter mode used by the DRBG is not provided by
the EVP interface. It is not CTR32. Using ECB and implementing the
correct counter in the DRBG was the easiest thing to do.

Ah ok. That makes sense. Why are we using a non-standard counter? Is that because of the NIST standard? Would it make it in anyway less secure to use the standard counter? If not, I could see us implementing a non-nist-drbg-ctr as well as the standard drbg-ctr....but still using EVP to fetch the ciphers (without using the low level interface). I could also see us teaching the low level implementations about the alternative counter variant...either as a completely new mode, or as some kind of ctrl/param of the existing one.

kroeckx:
Yes, it's required by the NIST standard.

I assume that if we use a different counter, the amount of data we
can generate with it becomes lower. We already use a lower limit
than allowed by NIST before we reseed. So my guess is that it's safe
to use a different counter.

mspncp:
> > It's because the counter mode used by the DRBG is not provided by
> > the EVP interface. It is not CTR32. Using ECB and implementing the
> > correct counter in the DRBG was the easiest thing to do.
> 
> Ah ok. That makes sense. Why are we using a non-standard counter? Is that because of the NIST standard? Would it make it in anyway less secure to use the standard counter? If not, I could see us implementing a non-nist-drbg-ctr as well as the standard drbg-ctr....but still using EVP to fetch the ciphers (without using the low level interface). I could also see us teaching the low level implementations about the alternative counter variant...either as a completely new mode, or as some kind of ctrl/param of the existing one.

Some more historical background: The original FIPS 2.0 DRBG implemented the AES-CTR mode strictly following the NIST SP800 90A reference implementation and using the low level AES functions (see [fips_drbg_ctr.c](https://github.com/openssl/openssl/blob/OpenSSL-fips-2_0-stable/fips/rand/fips_drbg_ctr.c)). 

When it was reported in

* Issue #5559 - Lack of performances compared to 1.1.0

that the CTR-DRBG performed badly, the implementation was changed blockwise from AES to EVP by @kroeckx, leaving the counting scheme untouched.

* PR #5580 DRBG: Use the EVP layer to do AES encryption  (merged to dbdcc04f27db70ac71748eb595ce23c9733afbe7)

It would have been nicer (and faster) to use @p-steuer's approach

* #4530 - aes ctr_drbg: use all platform specific aes support

but it turned out that the counting in the DRBG was done slightly differently from the standard AES-CTR way. The incompatible implementations of the counter mode turned a problem too difficult to solve for the late 1.1.1 beta (commit dbdcc04f27db70ac71748eb595ce23c9733afbe7 went into OpenSSL_1_1_1-pre4).

I actually believe that initially the NIST SP800 90A reference implementation was not mandatory, it was meant to be an example. So in theory, it could be possible to change the DRBG counting scheme and have it FIPS approved.
However, in practice everybody followed the reference implementation literally and many of the the CAVS and KAT test vectors probably rely on it and would break if we change the counting. And that would make FIPS evaluation of the DRBG much more complicated if not impractical.

p-steuer:
Good summary.

> I actually believe that initially the NIST SP800 90A reference implementation was not mandatory, it was meant to be an example. So in theory, it could be possible to change the DRBG counting scheme and have it FIPS approved.

Yes.

> All DRBG mechanisms and algorithms are described in this document in pseudocode, which is intended to explain functionality. The pseudocode is not intended to constrain real-world implementations.

[Chapter 9, NIST SP 800-90A Rev. 1]

 
> However, in practice everybody followed the reference implementation literally and many of the the CAVS and KAT test vectors probably rely on it and would break if we change the counting.

No. All NIST KAT vectors were originally part of my PR. Actually that was the only part of my PR that got merged.

mspncp:
Well, if you think it can be done, go ahead with #4530 :-)

mspncp:
> My proposal would have been behavorial equivalent to an outside observer, just some counter in the internal state would have been kept at +1 or -1 (dont remember) compared to the reference implmentation.

In hindsight it might be that I did not fully understand this part of your pull request at that time. I thought you were effectively changing the counting scheme. If the output is exactly the same (i.e. the CAVS tests succed) and only the internal representation slightly different, I don't see why it could'nt be FIPS approved. In fact it should be if it increases performance significantly. Even a backport to 1.1.1 as performance enhancement would be an option.

p-steuer:
One could even easily work around the counter thing by doing a decrement after the aes-ctr operation (which is unusual, but would keep internal state in sync with pseudo code).

The major blocker at that time was that it didnt use EVP though.

Im looking into it again right now.

mattcaswell:
> The major blocker at that time was that it didnt use EVP though.

I still consider EVP to be a requirement. But we can add ctrls/params/special modes to make it do what we want.

mspncp:
@raja-ashok thank you for your contribution nevertheless, even though we decided to take a different path.

PR #10461, opened 2019-11-18
Cache constants for fetched EVP_cipher

slontis:
##### Checklist

### DIFF
--- a/crypto/evp/evp_lib.c
+++ b/crypto/evp/evp_lib.c
@@ -272,16 +272,38 @@ int EVP_CIPHER_type(const EVP_CIPHER *ctx)
     }
 }
 
-int EVP_CIPHER_block_size(const EVP_CIPHER *cipher)
+int evp_cipher_cache_constants(EVP_CIPHER *cipher)
 {
     int ok;
-    size_t v = cipher->block_size;
-    OSSL_PARAM params[2] = { OSSL_PARAM_END, OSSL_PARAM_END };
+    size_t ivlen = 0;
+    size_t blksz = 0;
+    size_t keylen = 0;
+    unsigned int mode = 0;
+    unsigned long flags = 0;
+    OSSL_PARAM params[6];
 
-    params[0] = OSSL_PARAM_construct_size_t(OSSL_CIPHER_PARAM_BLOCK_SIZE, &v);
+    params[0] = OSSL_PARAM_construct_size_t(OSSL_CIPHER_PARAM_BLOCK_SIZE, &blksz);
+    params[1] = OSSL_PARAM_construct_size_t(OSSL_CIPHER_PARAM_IVLEN, &ivlen);
+    params[2] = OSSL_PARAM_construct_size_t(OSSL_CIPHER_PARAM_KEYLEN, &keylen);
+    params[3] = OSSL_PARAM_construct_uint(OSSL_CIPHER_PARAM_MODE, &mode);
+    params[4] = OSSL_PARAM_construct_ulong(OSSL_CIPHER_PARAM_FLAGS, &flags);
+    params[5] = OSSL_PARAM_construct_end();
     ok = evp_do_ciph_getparams(cipher, params);
+    if (ok) {
+        /* Provided implementations may have a custom cipher_cipher */
+        if (cipher->prov != NULL && cipher->ccipher != NULL)
+            flags |= EVP_CIPH_FLAG_CUSTOM_CIPHER;
+        cipher->block_size = blksz;
+        cipher->iv_len = ivlen;
+        cipher->key_len = keylen;
+        cipher->flags = flags | mode;
+    }
+    return ok;
+}
 
-    return ok != 0 ? (int)v : EVP_CTRL_RET_UNSUPPORTED;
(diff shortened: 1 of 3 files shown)


crypto/evp/evp_lib.c
```diff
@@ -272,16 +272,38 @@ int EVP_CIPHER_type(const EVP_CIPHER *ctx)
...
+
+    params[0] = OSSL_PARAM_construct_size_t(OSSL_CIPHER_PARAM_BLOCK_SIZE, &blksz);
+    params[1] = OSSL_PARAM_construct_size_t(OSSL_CIPHER_PARAM_IVLEN, &ivlen);
+    params[2] = OSSL_PARAM_construct_size_t(OSSL_CIPHER_PARAM_KEYLEN, &keylen);
+    params[3] = OSSL_PARAM_construct_uint(OSSL_CIPHER_PARAM_MODE, &mode);
+    params[4] = OSSL_PARAM_construct_ulong(OSSL_CIPHER_PARAM_FLAGS, &flags);
+    params[5] = OSSL_PARAM_construct_end();
     ok = evp_do_ciph_getparams(cipher, params);
+    if (ok) {
+        /* Provided implementations may have a custom cipher_cipher */
```
richsalz: provider?

richsalz:
much gain for few lines of code; nice!

beldmit:
Is it safe enough to cache flags? I'm not familiar with the provider code so are the cached values cached per implementation?

slontis:
The values that are cached here are equivalent to values that were in the old const EVP_CIPHER structs.. i.e- they are constant default values that should not change. The flags should be treated in the same way as the key_length (i.e this is the constant default key length).

beldmit:
I'm not sure about flags because some of them can be implementation-defined (e.g. `EVP_CIPH_FLAG_FIPS`).

slontis:
The values have been stored into the EVP_cipher during the fetch operation as they need to exist before either the new_ctx method or the ctx_int have been called.

e.g- it needs to handle the sequence
ciph = EVP_CIPHER_fetch(.....)
len = EVP_CIPHER_key_length(ciph)

slontis:
> I'm not sure about flags because some of them can be implementation-defined (e.g. `EVP_CIPH_FLAG_FIPS`).

Well that particular one is a dead flag left over from the old fips module.

crypto/evp/evp_lib.c
levitte: Either works.  The implementations are provided by a provider.

crypto/evp/evp_lib.c
```diff
@@ -272,16 +272,38 @@ int EVP_CIPHER_type(const EVP_CIPHER *ctx)
...
+        if (cipher->prov != NULL && cipher->ccipher != NULL)
+            flags |= EVP_CIPH_FLAG_CUSTOM_CIPHER;
```
levitte: Why is this necessary.  `EVP_CIPH_FLAG_CUSTOM_CIPHER` is only looked at in legacy code, and the place where we do use `ccipher`, we check for its existance, not this flag:

https://github.com/openssl/openssl/blob/bdbf2df2e685ae653f3c683ce2f734eb0c0888e0/crypto/evp/evp_lib.c#L312-L316

slontis: Happy to take it out,
It was added in
https://github.com/openssl/openssl/commit/f7397f0d58ce7ddf4c5366cd1846f16b341fbe43
The documentation for EVP_cipher that was changed in this commit seems to not be correct now.. It said it looks for the flag (it doesnt seem to though??)

levitte: "historical semantics"...  oh, right, that.

No, *we* don't check because we don't need to.  However, callers may want to know what semantics to expect, and the only recourse *they* have is that flag.

[runs in circles, screaming in horror]

levitte: So yeah, let's leave this alone.

paulidale:
@t8m, I don't think it changes the semantics quite like that: a cipher that doesn't know/support one of the cached parameters won't return it from the get query and it will default to zero.  True, the _unsupported_ return code won't happen anymore.

It would be an easy enough change to keep the _unsupported_ return code, if that is preferable.

I like the idea of all ciphers supporting all of these constants.  I do wonder what the IV length for an ECB mode cipher should be.  Unsupported seems more reasonable than zero.

levitte:
Why do you want to distinguish an unsupported IV length from zero? Is there a practical reason?

paulidale:
It's the first example of a change that sprung to mind -- agreed, it isn't a good one :sleeping:

This does represent a (small) change of behaviour which is the issue.

levitte:
Actually, this gets us back to pre-3.0 behavior, no?

paulidale:
@levitte, did you intend to approve this?  It is still missing an OMC approval....

levitte:
> @levitte, did you intend to approve this? It is still missing an OMC approval....

I was too quick on the labeling...  but since you approved, all's good.

p-steuer:
merged.

master: https://github.com/openssl/openssl/commit/3c957bcd54d097167e53660fa100aa1ba85d63b5

PR #10464, opened 2019-11-18
Remove handling of outdated macro's

richsalz:
code cleanup.  could probably be backported.

### DIFF
--- a/include/openssl/asn1.h
+++ b/include/openssl/asn1.h
@@ -276,7 +276,8 @@ typedef struct ASN1_VALUE_st ASN1_VALUE;
 # define TYPEDEF_I2D_OF(type) typedef int i2d_of_##type(const type *,unsigned char **)
 # define TYPEDEF_D2I2D_OF(type) TYPEDEF_D2I_OF(type); TYPEDEF_I2D_OF(type)
 
-TYPEDEF_D2I2D_OF(void);
+typedef void *d2i_of_void(void **, const unsigned char **, long);
+typedef int i2d_of_void(const void *, unsigned char **);
 
 /*-
  * The following macros and typedefs allow an ASN1_ITEM
--- a/util/mkerr.pl
+++ b/util/mkerr.pl
@@ -321,7 +321,7 @@ while ( ( my $hdr, my $lib ) = each %libinc ) {
         s/[\n\s]*$//g;
 
         # Skip over recognized non-function declarations
-        next if /typedef\W/ or /DECLARE_STACK_OF/ or /TYPEDEF_.*_OF/;
+        next if /typedef\W/;
 
         # Remove STACK_OF(foo)
         s/STACK_OF\(\w+\)/void/;
--- a/util/perl/OpenSSL/ParseC.pm
+++ b/util/perl/OpenSSL/ParseC.pm
@@ -372,34 +372,9 @@ EOF
     { regexp   => qr/DEFINE_STACK_OF_CONST<<<\((.*)\)>>>/,
       massager => sub { return ("SKM_DEFINE_STACK_OF($1,const $1,$1)"); },
     },
-    { regexp   => qr/PREDECLARE_STACK_OF<<<\((.*)\)>>>/,
-      massager => sub { return ("STACK_OF($1);"); }
-    },
-    { regexp   => qr/DECLARE_STACK_OF<<<\((.*)\)>>>/,
-      massager => sub { return ("STACK_OF($1);"); }
-    },
-    { regexp   => qr/DECLARE_SPECIAL_STACK_OF<<<\((.*?),\s*(.*?)\)>>>/,
-      massager => sub { return ("STACK_OF($1);"); }
-     },
 
     #####
     # ASN1 stuff
-
-    { regexp   => qr/TYPEDEF_D2I_OF<<<\((.*)\)>>>/,
-      massager => sub {
-          return ("typedef $1 *d2i_of_$1($1 **,const unsigned char **,long)");
-      },
-    },
-    { regexp   => qr/TYPEDEF_I2D_OF<<<\((.*)\)>>>/,
-      massager => sub {
-          return ("typedef $1 *i2d_of_$1($1 *,unsigned char **)");
-      },
-    },
-    { regexp   => qr/TYPEDEF_D2I2D_OF<<<\((.*)\)>>>/,
-      massager => sub {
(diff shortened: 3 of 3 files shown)


levitte:
Public macros...
Are you sure the DECLARE_STACK_OF and friends are gone?

richsalz:
Yes I am sure. I don't know when they were removed from the code, but they only were left behind in the tools.

richsalz:
Travis failure is unrelated:
```
providers/implementations/ciphers/cipher_aes_hw.c:136:203: error: expected '}' before 'cipher_hw_aes_copyctx'
```

richsalz:
rebased and pushed to clear CI error.

levitte:
Regarding this PR, I see no problem with removing support for the DECLARE macros that do not exists any more.  Removing *existing* public macros is a different story.

It's unfortunate that those two cases are bundled in one and the same PR, 'cause I would approve the former case immediately, but be more cautious regarding the former case.  So I can't approve this as it currently stands.  If someone else choose to approve it in its current form, I will however not stand in the way.

richsalz:
You are objecting to these macros in asn1.h, is that correct?
```
-# define TYPEDEF_D2I_OF(type) typedef type *d2i_of_##type(type **,const unsigned char **,long)
-# define TYPEDEF_I2D_OF(type) typedef int i2d_of_##type(const type *,unsigned char **)
-# define TYPEDEF_D2I2D_OF(type) TYPEDEF_D2I_OF(type); TYPEDEF_I2D_OF(type)
```

levitte:
I'm objecting to that removal, yes

richsalz:
I restored the define's but not the use of them, I left the expansion.

include/openssl/asn1.h
```diff
@@ -276,7 +276,8 @@ typedef struct ASN1_VALUE_st ASN1_VALUE;
 # define TYPEDEF_I2D_OF(type) typedef int i2d_of_##type(const type *,unsigned char **)
 # define TYPEDEF_D2I2D_OF(type) TYPEDEF_D2I_OF(type); TYPEDEF_I2D_OF(type)
 
-TYPEDEF_D2I2D_OF(void);
+typedef void *d2i_of_void(void **, const unsigned char **, long);
+typedef int i2d_of_void(const void *, unsigned char **);
```
levitte: Since you restored `TYPEDEF_D2I2D_OF`, the remaining change in this file is rather pointless, wouldn't you say?

util/mkerr.pl
```diff
@@ -321,7 +321,7 @@ sub help
         s/[\n\s]*$//g;
 
         # Skip over recognized non-function declarations
-        next if /typedef\W/ or /DECLARE_STACK_OF/ or /TYPEDEF_.*_OF/;
+        next if /typedef\W/;
```
levitte: You will need to partially restore the removed condition

util/perl/OpenSSL/ParseC.pm
```diff
@@ -372,34 +372,9 @@ EOF
...
-    { regexp   => qr/TYPEDEF_I2D_OF<<<\((.*)\)>>>/,
-      massager => sub {
-          return ("typedef $1 *i2d_of_$1($1 *,unsigned char **)");
-      },
-    },
-    { regexp   => qr/TYPEDEF_D2I2D_OF<<<\((.*)\)>>>/,
-      massager => sub {
-          return ("TYPEDEF_D2I_OF($1); TYPEDEF_I2D_OF($1)");
-      },
-    },
```
levitte: Ah, were you thinking that if you don't restore the "pointless change" in `include/openssl/asn1.h`, you won't have to restore this removal?

include/openssl/asn1.h
richsalz: I agree it's not necessary, but it's far from pointless.

The macros have one public use and that's not necessary.  This makes the transition of eventually removing them easier.  And also, the code is much more clear now IMO.

util/mkerr.pl
richsalz: Why?  It is not used in openssl code, and mkerr is only "suppported" for use in openssl code.

util/perl/OpenSSL/ParseC.pm
richsalz: yes.

include/openssl/asn1.h
levitte: Okie, I can accept that.

util/perl/OpenSSL/ParseC.pm
levitte: Understood

levitte: ... and acceptable.

richsalz:
ping @t8m if he has the cycles :)

richsalz:
Travis failure is not relevant.

mspncp:
> Public macros...
> Are you sure the DECLARE_STACK_OF and friends are gone?

FWIW: `DECLARE_STACK_OF` was renamed to `DEFINE_STACK_OF` in 2015 (commit 858857157290dd35145b14044ae96be9cd8eb0df).

util/perl/OpenSSL/ParseC.pm
mspncp: Late comment: I don't think the removal of `TYPEDEF_I2D_OF` is pointless. It was used only once, and it was obfuscating two simple declarations. So +1 for removing it.

richsalz:
I took your suggestion and expanded on it, squashed, pushed.

richsalz:
Reviewed and approved ten days ago; this is ready to merge.

mspncp:
Yup. Still waiting for the magic reminder bot...

mspncp:
Merged to trunk in 436c21a0fd317cb929fb00eefb1ac84e90676097. Thanks for your patience :-)

richsalz:
Thanks.  Curious about the bot, tho.  The "approval: done" label was added over a week ago.  Any thoughts on why the delay?

mspncp:
Because there is no reminder bot 🤖 yet. That‘s @levitte‘s homework from the f2f and still tbd. So we need to check the 24h grace period manually. Which is not perfect as you can see.

richsalz:
Aha, so the "still waiting for the magic reminder bot" meant: waiting for it to be written, not waiting for it to change the flag.  Got it.  Thanks.  And therefore my periodic trolls through my pull requests are still needed.

t8m:
I'll need to enhance my github mail filters as well. There is too much e-mail from it that I receive and I lose sometimes even the messages directly mentioning me.

Issue #9581, opened 2019-08-13
OpenSSL app isn't compatible with dynamic engines


UndefBehavior:
Environment: windows 10 x64, OpenSSL 1.1.0i/1.1.1b

There is no engine cleanup at the end of openssl app execution so dynamic engines will be freed in `exit()` function. as a result, dynamic engine will be freed by OS first and then libcrypto calls ENGINE_finish/ENGINE_destroy implementations from freed engine. As a result, there is a segfault in case of malloc from libc (not OPENSSL_malloc) in ENGINE_create and free (not OPENSSL_free) in ENGINE_destroy.

Reproducing example:
```sh
./openssl.exe speed -evp 123
```
This example loads dynamic engine via config file and then tries to launch speed test for invalid cipher (there is the same behaviour for valid cipher). At the end of the execution I have the following stacktraces (OpenSSL 1.1.0i):
```
>   my_engine.dll!DllMain(HINSTANCE__ * instance, unsigned long reason, void * reserved) Line 21 C++
    my_engine.dll!dllmain_dispatch(HINSTANCE__ * const instance, const unsigned long reason, void * const reserved) Line 200 C++
    my_engine.dll!_DllMainCRTStartup(HINSTANCE__ * const instance, const unsigned long reason, void * const reserved) Line 254   C++
    ntdll.dll!00007ffbfa654143()    Unknown
    ntdll.dll!00007ffbfa660af5()    Unknown
    ntdll.dll!00007ffbfa6609b8()    Unknown
    kernel32.dll!00007ffbf811b19a() Unknown
    openssl.exe!exit_or_terminate_process(const unsigned int return_code) Line 144  C++
    openssl.exe!common_exit(const int return_code, const _crt_exit_cleanup_mode cleanup_mode, const _crt_exit_return_mode return_mode) Line 282 C++
    openssl.exe!exit(int return_code) Line 294  C++
    openssl.exe!main(int argc, char * * argv) Line 267  C
    openssl.exe!invoke_main() Line 75   C++
    openssl.exe!__scrt_common_main_seh() Line 264   C++
    openssl.exe!__scrt_common_main() Line 309   C++
    openssl.exe!mainCRTStartup() Line 17    C++
    kernel32.dll!00007ffbf8113034() Unknown
    ntdll.dll!00007ffbfa691551()    Unknown
```
and AFTER then:
```
    my_engine.dll!engine_finish(engine_st * engine) Line 596 C
[...]

PR #10465, opened 2019-11-18
Finalize dynamic engines before OpenSSL tool execution's end

UndefBehavior:
OpenSSL calls cleanup functions in atexit() fuction by default for
libraries which were freed by OS to this moment. This lead to double free
in some legal cases such as calling free() for memory which was allocated
in initialization routine with malloc().

Fixes: #9581
CLA: trivial

##### Checklist

### DIFF
--- a/apps/openssl.c
+++ b/apps/openssl.c
@@ -55,6 +55,7 @@ static int apps_startup(void)
 
     /* Set non-default library initialisation settings */
     if (!OPENSSL_init_ssl(OPENSSL_INIT_ENGINE_ALL_BUILTIN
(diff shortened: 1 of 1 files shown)


mattcaswell:
This is not an acceptable change. If there is a problem with ordering in the ENGINE deinit code then we need to look into the reasons for that. That answer is almost certainly not to avoid using auto-deinit.

levitte:
I agree with @mattcaswell. This fix is a band-aid, and we should dive into the problem and figure out what's actually going on.

t8m:
Closing per comments above.

PR #10474, opened 2019-11-20
EVP p_lib: Add NULL check to EVP_PKEY_missing_parameters.

### LINKED ISSUE #10404: Potential NULL dereference
Most of the functions that receive EVP_PKEY have a null check on pkey argument, like:

https://github.com/openssl/openssl/blob/1903a9b77af1aa289ce858de90790f3e2749fea9/crypto/evp/p_lib.c#L41
https://github.com/openssl/openssl/blob/1903a9b77af1aa289ce858de90790f3e2749fea9/crypto/evp/p_lib.c#L50

While a null check is missed in `EVP_PKEY_missing_parameters`:
https://github.com/openssl/openssl/blob/1903a9b77af1aa289ce858de90790f3e2749fea9/crypto/evp/p_lib.c#L106-L108

We noticed it is possible the input of this function might be null. For example `X509_get0_pubkey ` might return null:
https://github.com/openssl/openssl/blob/5388f9862d9aaf3c7cf7a70c1e36e7e983c26cfc/apps/ca.c#L1917-L1918

We believe there should be a null check on `pkey` in `EVP_PKEY_missing_parameters` function, otherwise, there will be a NULL dereference.

Thanks!

paulidale:
Check for NULL and return error if so.
This can possibly be called from `apps/ca.c` with a NULL argument.

Fixes #10404

### DIFF
--- a/crypto/evp/p_lib.c
+++ b/crypto/evp/p_lib.c
@@ -105,7 +105,7 @@ int EVP_PKEY_copy_parameters(EVP_PKEY *to, const EVP_PKEY *from)
 
 int EVP_PKEY_missing_parameters(const EVP_PKEY *pkey)
 {
-    if (pkey->ameth && pkey->ameth->param_missing)
+    if (pkey != NULL && pkey->ameth && pkey->ameth->param_missing)
         return pkey->ameth->param_missing(pkey);
     return 0;
 }


paulidale:

PR #10475, opened 2019-11-20
ECDSA: don't clear free memory after verify.

### LINKED ISSUE #10406: OPENSSL_clear_free instead of OPENSSL_free
It seems `OPENSSL_clear_free` needs to be called instead of `OPENSSL_free` here:
https://github.com/openssl/openssl/blob/1903a9b77af1aa289ce858de90790f3e2749fea9/crypto/sm2/sm2_sign.c#L475

As it has done here:
https://github.com/openssl/openssl/blob/1903a9b77af1aa289ce858de90790f3e2749fea9/crypto/ec/ecdsa_ossl.c#L347

paulidale:
Verifications are public, there is no need to clear the used storage before
freeing it.

Fixes #10406


romen:
Nice! This was in my to do list!

paulidale:

PR #10478, opened 2019-11-20
Update NOTES.ANDROID for newer NDK versions + small fixes.

### LINKED ISSUE #8941: No NDK aarch64-linux-android-gcc on $PATH at (eval 6) line 124.
I am working with 1.1.1b and trying to compile for android-arm64.

I am currently using ndk r15c.

I run the ./setenv-android.sh script with my fields set properly
_ANDROID_NDK="android-ndk-r15c"
_ANDROID_EABI="arm-linux-androideabi-4.9"
_ANDROID_ARCH=arch-arm
_ANDROID_API="android-18"

and I don't touch anything else in that file. The script runs without issue.

Next, I set my ANDROID_NDK_HOME to the location of the r15c ndk.

Then I set my PATH according to the NOTES.ANDROID instructions provided in the openSSL tar, just changing from 4.8 to 4.9

PATH=$ANDROID_NDK_HOME/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin:$PATH

Then, I attempt to do ./Configure android-arm64 and get the following error:

Configuring OpenSSL version 1.1.1b (0x1010102fL) for android-arm64
Using os-specific seed configuration
no NDK aarch64-linux-android-gcc on $PATH at (eval 6) line 124.

Can anyone tell me how to fix this?

Kagetsuki:
Fixes #8941

*This has only been tested by myself - it would be good if someone can check the accuracy of new information/instructions before merging.*

The newer NDK's don't include GCC and have different structures and paths, especially if you are using the new side-by-side NDK. I've retained the information for older NDK's as it's still relevant for people who are not updating their toolchains (EG maintaining somewhat legacy apps).

##### Checklist

### DIFF
--- a/NOTES.ANDROID
+++ b/NOTES.ANDROID
@@ -36,7 +36,7 @@
  to compile for Android 10 arm64 with a side-by-side NDK r20.0.5594570
 
 	export ANDROID_NDK_HOME=/home/whoever/Android/android-sdk/ndk/20.0.5594570
-	PATH=$ANDROID_NDK_HOME/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin:$PATH
+	PATH=$ANDROID_NDK_HOME/toolchains/llvm/prebuilt/linux-x86_64/bin:$ANDROID_NDK_HOME/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin:$PATH
 	./Configure android-arm64 -D__ANDROID_API__=29
 	make
  
--- a/NOTES.ANDROID
+++ b/NOTES.ANDROID
@@ -15,22 +15,33 @@
  Configuration
  -------------
 
- Android is naturally cross-compiled target and you can't use ./config.
+ Android is a naturally cross-compiled target and you can't use ./config.
  You have to use ./Configure and name your target explicitly; there are
  android-arm, android-arm64, android-mips, android-mip64, android-x86
- and android-x86_64. Do not pass --cross-compile-prefix (as you might
- be tempted), as it will be "calculated" automatically based on chosen
- platform. Though you still need to know the prefix to extend your PATH,
- in order to invoke $(CROSS_COMPILE)gcc and company. (Configure will fail
- and give you a hint if you get it wrong.) Apart from PATH adjustment
- you need to set ANDROID_NDK_HOME environment to point at NDK directory
- as /some/where/android-ndk-<ver>. Both variables are significant at both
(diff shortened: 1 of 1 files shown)


levitte:
Isn't this update supposed to include information for NDK 20+? It seems to be missing...

Kagetsuki:
> Isn't this update supposed to include information for NDK 20+? It seems to be missing...

There are several references, but you made me realize I forgot to include part of the path needed for the clang toolchain. I've updated the document.

levitte:
Cool. Do you need to say anything special to get clang? I usually have to say CC=clang on my linux builds...

Kagetsuki:
> Cool. Do you need to say anything special to get clang? I usually have to say CC=clang on my linux builds...

No extra tags needed! clang detection was added as a fallback for when GCC is not found at some point. The part of the path that I had forgotten to add that your comment reminded me of was actually the path to the Android NDK llvm/clang tools, which for some reason are in a different path than all the other tools... Otherwise there really wasn't much of a big change for r20+ other than a few caveats (like paths being slightly different for side-by-side NDK, which is now the "recommended" way to obtain it).

I'm currently trying to get Gradle to build out all platforms through the side-by-side NDK and their new external build tools. I suppose if I get that working I should add that to the document as well and put up another PR - but it's not exactly vital information just "convenient" to some.

levitte:
Thanks for the explanation

mattcaswell:
Travis failure is not relevant to this change but is interesting (@paulidale):

````
    # ERROR: (int) 'errors < max - tail' failed @ test/property_test.c:379
    # [9990] compared to [9990]
    not ok 8 - test_query_cache_stochastic
````

paulidale:
Another that dropped off the radar.  Merged to master and 1.1.1.
Travis failure is not related.

petrovr:
mattcaswell wrote:
> mattcaswell approved this pull request.
I wonder how you check that documentation is correct?

levitte:
We can't always double check things ourselves, so we have to trust our community.
If you find an error, feel free to submit a correcting PR

petrovr:
> We can't always double check things ourselves, so we have to trust our community.
> If you find an error, feel free to submit a correcting PR

Vote of openssl team member is important. Vote of other does not make sense if I remember rules.
As result (even for changes from team members) team member must be more precise - double check is required in all cases.

I don't have enough spare time to propose changes.
In addition I have enough experience to use work-around broken build system. But in this case build system is fine and so not precise documentation is out of my scope.

PR #9039, opened 2019-05-29
Prepare the RAND code for moving into the FIPS module

mattcaswell:
This is mostly about changing the way the RAND code initialises itself so that it is based on OPENSSL_CTX instead.

(Needed in order to move the RAND code into the FIPS provider - see #9035)

PR #10483, opened 2019-11-20
rand_lib.c: fix null pointer dereferences after RAND_get_rand_method() failure

### LINKED ISSUE #10480: Null-pointer-dereference bug at rand_lib.c
Hello, I have found an out-of-memory bug with my fuzzing tool **FIFUZZ**. The bug information is following.

**Describe the bug**
The **malloc()** in the call stack shown below may fail:

    #0 Call malloc() in CRYPTO_malloc(), at mem.c: 222
    #1 Call CRYPTO_malloc() in CRYPTO_zalloc(), at mem.c: 230
    #2 Call CRYPTO_zalloc() in CRYPTO_THREAD_lock_new(), at threads_pthread.c: 33
    #3 Call CRYPTO_THREAD_lock_new() in do_rand_init(), at rand_lib.c: 250
    #4 Call do_rand_init() in do_rand_init_ossl_(), at rand_lib.c: 242
    #5 Call do_rand_init_ossl_() in pthread_once()
    #6 Call pthread_once() in CRYPTO_THREAD_run_once(), at threads_pthread.c: 122
    #7 Call CRYPTO_THREAD_run_once() in RAND_get_rand_method(), at rand_lib.c: 777
    #8 Call RAND_get_rand_method() in rand_bytes_ex(), at rand_lib.c: 877
    #9 Call rand_bytes_ex() in RAND_bytes(), at rand_lib.c: 896
    #10 Call RAND_bytes() in enc_main(), at enc.c: 446
    #11 Call enc_main() in do_cmd(), at openssl.c:481
    #12 Call do_cmd() in main(), at openssl.c: 294

If the **malloc()** in this call stack fails, **RAND_get_rand_method()** in **rand_bytes_ex()** will return NULL at rand_lib.c: 877, so **meth** will become NULL. And then the program will check whether **meth** is equal to **RAND_OpenSSL()**. Since **FIPS_MODE** is not defined, so **RAND_OpenSSL()** returns **&rand_meth**, and it is not equal to **meth** (it is NULL). And then **meth->bytes** causes a **null-pointer-dereference** at rand_lib.c: 880.

The Stack Traceback when the bug occurs:

    (gdb) backtrace
    #0  0x00007ffff753023b in rand_bytes_ex (ctx=0x0, buf=0x7fd810 "\300\332\377\377\377\177", num=      8) at crypto/rand/rand_lib.c:880
    #1  0x00007ffff7530498 in RAND_bytes (buf=0x7fd810 "\300\332\377\377\377\177", num=8) at crypto/rand/rand_lib.c      :896
    #2  0x04579e0 in enc_main (argc=12, argv=         0x7fe2a0) at apps/enc.c:446
    #3  0x047cee2 in do_cmd (prog=0x7b7700, argc=12, argv=         0x7fe2a0) at apps/openssl.c:481
    #4  0x047bd1f in main (argc=12, argv=         0x7fe2a0) at apps/openssl.c:294

**To Reproduce**
Detailed steps to reproduce the behavior:
1. Type openssl enc -e -des3 -a -salt -in /tmp/test.c -out test.cipher -k testtesttest -pbkdf2
1. Make the _malloc()_ shown at **Describe the bug** fail.
2. The program will crash.

**Expected behavior**
The program would not crash when _malloc()_ fails.

**Environment (please complete the following information):**
 - Version: 3.0.0-dev
 - OS: Ubuntu 18.04
 - Compiler: Clang-6.0
 - Application: openssl 3.0.0-dev

**Additional context**
This bug is found by my fuzzing tool and the bug is triggered at runtime.

mspncp:
rand_bytes_ex() calls RAND_get_rand_method(), which can return a NULL
method pointer in the case of a malloc failure.

Reported-by: Zu-Ming Jiang (detected by FIFUZZ)

Fixes #10480


JZuming:
It seems that this patch can fix the bug because my fuzzer doesn't report it after applying the patch.

mspncp:
The 1.1.1 stable branch needs to be fixed, too, but the commits don't cherry-pick cleanly. I'll create a separate pr....

crypto/rand/rand_lib.c
```diff
@@ -854,6 +854,9 @@ int rand_priv_bytes_ex(OPENSSL_CTX *ctx, unsigned char *buf, int num)
     int ret;
     const RAND_METHOD *meth = RAND_get_rand_method();
 
+    if (meth == NULL)
+        return 0;
+
```
mspncp: @paulidale I see that a NULL pointer is returned in FIPS mode on purpose. Does this mean that this code path here is never intended to be reached in FIPS mode, or am I missing something?

https://github.com/openssl/openssl/blob/04bb55daeafc6da36f4445c26cc95543e9c04e51/crypto/rand/rand_lib.c#L770-L774

mspncp:
I located some more potential null pointer dereferences, so I squashed them and adjusted the commit message.

crypto/rand/rand_lib.c
mspncp: Or maybe @mattcaswell is the right person to ask: What was the reason why you added the NULL return in FIPS mode? The commit message below does not give an indication about it.

commit a2f27fd750b9ae62a571a9212c7154889100bdb0
Author: Matt Caswell <matt@openssl.org>
Date:   Fri May 24 16:36:44 2019 +0100

    Move the rand_nonce_lock code into drbg_lib.c
    
    It was previously rand_lib but it makes more sense in drbg_lib.c since
    all the functions that use this lock are only ever called from drbg_lib.c
    
    We add some FIPS_MODE defines in preparation for later moving this code
    into the FIPS module.
    
    (Merged from https://github.com/openssl/openssl/pull/9039)

crypto/rand/rand_lib.c
```diff
@@ -831,15 +834,15 @@ void RAND_seed(const void *buf, int num)
 {
     const RAND_METHOD *meth = RAND_get_rand_method();
 
-    if (meth->seed != NULL)
+    if (meth != NULL && meth->seed != NULL)
```
mspncp: I'm feeling a bit uncomfortable about this location here, because there is no way to indicate an error to caller. OTOH: if the CSPRNG is already initialized, the  failure will not be disastrous, and if it is not, the generate call will fail later on.

crypto/rand/rand_lib.c
```diff
@@ -831,15 +834,15 @@ void RAND_seed(const void *buf, int num)
...
         meth->seed(buf, num);
 }
 
 void RAND_add(const void *buf, int num, double randomness)
 {
     const RAND_METHOD *meth = RAND_get_rand_method();
 
-    if (meth->add != NULL)
+    if (meth != NULL && meth->add != NULL)
```
mspncp: Same here.

crypto/rand/rand_lib.c
levitte: How much of an API break is it to return an int (that current application code would simply ignore)?  If considered harmless, we *could* change the return value for this function (and also for the other functions that are currently void)

crypto/rand/rand_lib.c
mspncp: Note: It seems to me that the FIPS changes are not final yet. At least, they are not consistent:

While for `RAND_get_rand_method()` only the implementation is removed in FIPS mode,

https://github.com/openssl/openssl/blob/04bb55daeafc6da36f4445c26cc95543e9c04e51/crypto/rand/rand_lib.c#L770-L774

the `RAND_set_rand_engine()` method is removed entirely,


https://github.com/openssl/openssl/blob/04bb55daeafc6da36f4445c26cc95543e9c04e51/crypto/rand/rand_lib.c#L804-L807

even though there seems to be no FIPS-replacement for this public API function (yet).

paulidale: The FIPS RAND stuff is in flux.  Moving the DRBG/seeding into a provider is proving troublesome.  The code needs to be split both along the FIPS/non-FIPS axis but also along an algorithm/context one.  I'm also trying to merge the DRBG and seeding into a single algorithm class.  Maybe I'm being too ambitious :grin:

crypto/rand/rand_lib.c
paulidale: I didn't think `RAND_get_rand_method` could fail after the RNG was initialised.  Isn't it's failure path when creating the method on first call?

The API break would be when taking the address of the function.  How likely is it for `&RAND_seed` (or just `RAND_seed` since the _&_ is optional) to appear in code?

I think having `RAND_seed` return a failure indicator would be an improvement however.

crypto/rand/rand_lib.c
levitte: > Maybe I'm being too ambitious :grin:

I know that feeling...

mspncp: > The FIPS RAND stuff is in flux. 

Ok, I'll just ignore the current fluctuations and temporary inconsistencies and only add null pointer checks where necessary.

richsalz: maybe going back to just one supported algorithm (aes-ctr) would help simplify?

paulidale: The actual DRBGs are looking like being the easy bit.

crypto/rand/rand_lib.c
mspncp: After thinking about it for a while, I don't believe it is worth breaking the `RAND_{add,seed}()` API in order to return an error, for the following reasons:

People currently don't check the return value of `RAND_{add,seed}()`  anyway, as @levitte already noted. And actually it does not matter, because if the (re-)seeding fails then the DRBG will remain in the error state, so `RAND_status()` and `RAND_bytes()` will fail anyway.

The `RAND_{add,seed}()` APIs are obsolescent, because the DRBG reseeds itself anyway. We shouldn't change an API which we don't want to be used anymore.

levitte: Perhaps some deprecation is in order

mspncp:
Found one more missing NULL pointer check. I think this pull request is ready for review, subjects like API changes go beyound the scope of this pr.

mspncp:
The [branch: 1.1.1] label just indicates my intention to backport. The commits don't cherry-pick cleanly anyway.

crypto/rand/rand_lib.c
paulidale: I figure we'll need to do a big pass over everything at some point and mass deprecate old APIs.

mspncp:
Backport is in #10490.

paulidale:
CIs aren't happy.
It doesn't look related but there are a number of failures that are similar.

mspncp:
> I'll kick them anew

I noticed it. They suddenly restarted while I was looking at them. ;-) 

(After seeing two jobs hanging in the fipsinstall tests)

mspncp:
Hmmm... fipsinstall test is still hanging. When I try to run it on my machine, it does not hang but it fails.

I have no idea how the hang could be related to my pr. Any ideas @paulidale?


```
~/src/openssl$ make test TESTS=test_fipsinstall V=1
make depend && make _tests
make[1]: Entering directory '/home/msp/src/openssl'
make[1]: Leaving directory '/home/msp/src/openssl'
make[1]: Entering directory '/home/msp/src/openssl'
( cd test; \
  mkdir -p test-runs; \
  SRCTOP=../. \
  BLDTOP=../. \
  RESULT_D=test-runs \
  PERL="/usr/bin/perl" \
  EXE_EXT= \
  OPENSSL_ENGINES=`cd .././engines 2>/dev/null && pwd` \
  OPENSSL_MODULES=`cd .././providers 2>/dev/null && pwd` \
  OPENSSL_DEBUG_MEMORY=on \
    /usr/bin/perl .././test/run_tests.pl test_fipsinstall )
03-test_fipsinstall.t .. 
1..6
fipsinstall: Use -help for summary.
INSTALL FAILED
/home/msp/src/openssl/util/shlib_wrap.sh /home/msp/src/openssl/apps/openssl fipsinstall -out fips.conf -module -provider_name fips -macopt 'digest:SHA256' -macopt 'hexkey:00' -section_name fips_install => 1
ok 1 - fipinstall fail
Can't open dummy.tmp for reading, No such file or directory
40:51:71:52:40:7F:00:00:error:system library:BIO_new_file:No such file or directory:crypto/bio/bss_file.c:69:calling fopen(dummy.tmp, r)
40:51:71:52:40:7F:00:00:error:BIO routines:BIO_new_file:no such file:crypto/bio/bss_file.c:77:
VERIFY FAILED
/home/msp/src/openssl/util/shlib_wrap.sh /home/msp/src/openssl/apps/openssl fipsinstall -in dummy.tmp -module /home/msp/src/openssl/providers/fips.so -provider_name fips -mac_name HMAC -macopt 'digest:SHA256' -macopt 'hexkey:00' -section_name fips_install -verify => 1
ok 2 - fipinstall verify fail
Failed to load FIPS module
INSTALL FAILED
40:41:AE:AF:40:7F:00:00:error:common libcrypto routines:provider_activate:init fail:crypto/provider_core.c:453:
40:41:AE:AF:40:7F:00:00:error:configuration file routines:module_run:module initialization error:crypto/conf/conf_mod.c:181:module=providers, value=provider_section, retcode=-1      
40:41:AE:AF:40:7F:00:00:error:DSO support routines:DSO_get_filename:passed a null parameter:crypto/dso/dso_lib.c:234:
40:41:AE:AF:40:7F:00:00:error:system library:BIO_new_file:Bad address:crypto/bio/bss_file.c:69:calling fopen(<NULL>, rb)
40:41:AE:AF:40:7F:00:00:error:BIO routines:BIO_new_file:system lib:crypto/bio/bss_file.c:79:
40:41:AE:AF:40:7F:00:00:error:common libcrypto routines:provider_activate:init fail:crypto/provider_core.c:453:
/home/msp/src/openssl/util/shlib_wrap.sh /home/msp/src/openssl/apps/openssl fipsinstall -out fips.conf -module /home/msp/src/openssl/providers/fips.so -provider_name fips -mac_name HMAC -macopt 'digest:SHA256' -macopt 'hexkey:00' -section_name fips_install => 139
not ok 3 - fipinstall
#   Failed test 'fipinstall'
#   at ../test/recipes/03-test_fipsinstall.t line 48.
Can't open fips.conf for reading, No such file or directory
40:C1:E6:78:9F:7F:00:00:error:system library:BIO_new_file:No such file or directory:crypto/bio/bss_file.c:69:calling fopen(fips.conf, r)
40:C1:E6:78:9F:7F:00:00:error:BIO routines:BIO_new_file:no such file:crypto/bio/bss_file.c:77:
VERIFY FAILED
/home/msp/src/openssl/util/shlib_wrap.sh /home/msp/src/openssl/apps/openssl fipsinstall -in fips.conf -module /home/msp/src/openssl/providers/fips.so -provider_name fips -mac_name HMAC -macopt 'digest:SHA256' -macopt 'hexkey:00' -section_name fips_install -verify => 1
not ok 4 - fipinstall verify
#   Failed test 'fipinstall verify'
#   at ../test/recipes/03-test_fipsinstall.t line 55.
Can't open fips.conf for reading, No such file or directory
40:11:B9:34:A8:7F:00:00:error:system library:BIO_new_file:No such file or directory:crypto/bio/bss_file.c:69:calling fopen(fips.conf, r)
40:11:B9:34:A8:7F:00:00:error:BIO routines:BIO_new_file:no such file:crypto/bio/bss_file.c:77:
VERIFY FAILED
/home/msp/src/openssl/util/shlib_wrap.sh /home/msp/src/openssl/apps/openssl fipsinstall -in fips.conf -module /home/msp/src/openssl/providers/fips.so -provider_name fips -mac_name HMAC -macopt 'digest:SHA256' -macopt 'hexkey:01' -section_name fips_install -verify => 1
ok 5 - fipinstall verify fail bad key
Can't open fips.conf for reading, No such file or directory
40:31:B9:A9:73:7F:00:00:error:system library:BIO_new_file:No such file or directory:crypto/bio/bss_file.c:69:calling fopen(fips.conf, r)
40:31:B9:A9:73:7F:00:00:error:BIO routines:BIO_new_file:no such file:crypto/bio/bss_file.c:77:
VERIFY FAILED
/home/msp/src/openssl/util/shlib_wrap.sh /home/msp/src/openssl/apps/openssl fipsinstall -in fips.conf -module /home/msp/src/openssl/providers/fips.so -provider_name fips -mac_name HMAC -macopt 'digest:SHA512' -macopt 'hexkey:00' -section_name fips_install -verify => 1
ok 6 - fipinstall verify fail incorrect digest
# Looks like you failed 2 tests of 6.
Dubious, test returned 2 (wstat 512, 0x200)
Failed 2/6 subtests 

Test Summary Report
-------------------
03-test_fipsinstall.t (Wstat: 512 Tests: 6 Failed: 2)
  Failed tests:  3-4
  Non-zero exit status: 2
Files=1, Tests=6,  0 wallclock secs ( 0.01 usr  0.00 sys +  0.15 cusr  0.03 csys =  0.19 CPU)
Result: FAIL
make[1]: *** [Makefile:1891: _tests] Error 1
make[1]: Leaving directory '/home/msp/src/openssl'
make: *** [Makefile:1888: tests] Error 2
```

mspncp:
So the fipsinstall test requires a`fips.conf` file? Acccording to FIPS_CONFIG(5), this file gets created at installation time only? How am going to be able to run the test suite _before_ installing then?

```
Can't open fips.conf for reading, No such file or directory
```

paulidale:
That sounds like a concern.  @slontis??

mspncp:
@paulidale do you consider the CI failure unrelated? Or should I hold this pr back until the problem is better understood?

paulidale:
I'd hold off for now.  I don't think the problem is related but the second build failed the same way and other PRs are building okay.  @slontis will be online in a couple of hours and he knows more about the FIPS tests and I do.

slontis:
> So the fipsinstall test requires a`fips.conf` file? According to FIPS_CONFIG(5), this file gets created at installation time only. How am I going to be able to run the test suite _before_ installing then?
> 
> ```
> Can't open fips.conf for reading, No such file or directory
> ```

The fipsinstall app creates the fips.conf file.
If it fails it means that the fips provider failed to load.. (the self tests shouldnt be using rand yet - but something must be accessing it and causing an error).
So somehow this new check is causing the failure..

rand_bytes_ex() is a new function that takes a libctx so I assume this is the one that is supposed to work within the fips module. I will track down the actual path that causes the failure..

mspncp:
> I will track down the actual path that causes the failure..

I'm almost sure the path to failure passes through rand_lib.c:776. That's where I would set a breakpoint.  ;-)

https://github.com/openssl/openssl/blob/1d981c156113cb92bd067bb30b7c40877acfa340/crypto/rand/rand_lib.c#L775-L777

crypto/rand/rand_lib.c
```diff
@@ -876,6 +882,9 @@ int rand_bytes_ex(OPENSSL_CTX *ctx, unsigned char *buf, int num)
     int ret;
     const RAND_METHOD *meth = RAND_get_rand_method();
 
+    if (meth == NULL)
```
paulidale: @slontis found the code path...

This check here is wrong as is, it needs to be removed.  Then add a `meth != NULL &&` to the test against `RAND_OpenSSL()` below.

If the FIPS case, there is a second copy of this code running in the provider and both `RAND_get_rand_method()` and `RAND_OpenSSL` legitimately return NULL.  The provider relies on both `rand_bytes` functions falling through to the drbg case below.  By returning error instead it fails.

I'm assuming that's the only problem introduced.  😸

mspncp:
I followed your advice @paulidale, see d75c801ba0bc592d6c6a7aa63892364dd177ef97, but it seems that the fipsinstall tests still fail.

```
~/src/openssl$ make test TESTS=test_fipsinstall V=1

...

Failed to load FIPS module
INSTALL FAILED
40:91:CF:CA:34:7F:00:00:error:common libcrypto routines:provider_activate:init fail:crypto/provider_core.c:453:
40:91:CF:CA:34:7F:00:00:error:configuration file routines:module_run:module initialization error:crypto/conf/conf_mod.c:181:module=providers, value=provider_section, retcode=-1      
40:91:CF:CA:34:7F:00:00:error:DSO support routines:DSO_get_filename:passed a null parameter:crypto/dso/dso_lib.c:234:
40:91:CF:CA:34:7F:00:00:error:system library:BIO_new_file:Bad address:crypto/bio/bss_file.c:69:calling fopen(<NULL>, rb)
40:91:CF:CA:34:7F:00:00:error:BIO routines:BIO_new_file:system lib:crypto/bio/bss_file.c:79:
40:91:CF:CA:34:7F:00:00:error:common libcrypto routines:provider_activate:init fail:crypto/provider_core.c:453:
/home/msp/src/openssl/util/shlib_wrap.sh /home/msp/src/openssl/apps/openssl fipsinstall -out fips.conf -module /home/msp/src/openssl/providers/fips.so -provider_name fips -mac_name HMAC -macopt 'digest:SHA256' -macopt 'hexkey:00' -section_name fips_install => 139
not ok 3 - fipinstall
```

The first failing test, number 3, returns an interesting error code, 139. I looked it up, error 139 indicates a [segmentation fault](https://www.student.cs.uwaterloo.ca/~cs136/seashell/docs/seashell-error-codes.html#%28part._.Error_code_139_-_.Segmentation_.Fault%29). It seems that independently of the error caused by my pull request, there seems to be another error, a segmentation fault happening during cleanup, which is caused by a null pointer deference.


I checked the crash it in the debugger, using

```
OPENSSL_MODULES=./providers /home/msp/src/openssl/util/shlib_wrap.sh gdb --args /home/msp/src/openssl/apps/openssl fipsinstall -out fips.conf -module /home/msp/src/openssl/providers/fips.so -provider_name fips -mac_name HMAC -macopt 'di
```

Here is the call stack of the segmentation fault:

```
Program received signal SIGSEGV, Segmentation fault.
__GI___pthread_rwlock_rdlock (rwlock=0x0) at pthread_rwlock_rdlock.c:27
27	pthread_rwlock_rdlock.c: No such file or directory.
(gdb) bt
#0  __GI___pthread_rwlock_rdlock (rwlock=0x0) at pthread_rwlock_rdlock.c:27
#1  0x00007ffff786ab1f in CRYPTO_THREAD_read_lock (lock=0x0) at crypto/threads_pthread.c:69
#2  0x00007ffff7864137 in openssl_ctx_get_data (ctx=0x55555568ce60, index=8, meth=0x7ffff78f2bc0 <thread_event_ossl_ctx_method>) at crypto/context.c:217
#3  0x00007ffff7866ab2 in ossl_ctx_thread_stop (arg=0x55555568ce60) at crypto/initthread.c:286
#4  0x00007ffff7d10b3b in init_thread_stop (arg=0x0, hands=0x5555556459c0) at crypto/initthread.c:312
#5  0x00007ffff7d10a63 in OPENSSL_thread_stop () at crypto/initthread.c:227
#6  0x00007ffff7d0fedc in OPENSSL_cleanup () at crypto/init.c:355
#7  0x00007ffff793584c in __run_exit_handlers (status=1, listp=0x7ffff7ac3718 <__exit_funcs>, run_list_atexit=run_list_atexit@entry=true, run_dtors=run_dtors@entry=true) at exit.c:108
#8  0x00007ffff793598a in __GI_exit (status=<optimized out>) at exit.c:139
#9  0x00005555555ad3a4 in main (argc=15, argv=0x7fffffffd5f0) at apps/openssl.c:379
```

mspncp:
> I followed your advice @paulidale, see [d75c801](https://github.com/openssl/openssl/commit/d75c801ba0bc592d6c6a7aa63892364dd177ef97), but it seems that the fipsinstall tests still fail.

I mad a stupid mistake, fixed in b47523bfbc938b5f491ab3acd208df71047642a6. Now the test hopefully passes. The segmentation fault during cleanup needs to be fixed nevertheless. I'll take a look at it.

mspncp:
> a segmentation fault happening during cleanup, which is caused by a null pointer deference.

Further analysis shows that in frame \# 2 (openssl_ctx_get_data() line 217)

https://github.com/openssl/openssl/blob/d75c801ba0bc592d6c6a7aa63892364dd177ef97/crypto/context.c#L213-L219

the `*ctx` structure seems to be initialized but `ctx->lock` is a null pointer. Is this expected behaviour? In other words, is it enought to add a null check before calling `CRYPTO_THREAD_read_lock()`, or do I need to investigate why `ctx->lock` is unset?

```
$1 = (OPENSSL_CTX *) 0x55555568ce60
(gdb) p	*ctx
$2 = {lock = 0x0, data = {ctx = 0x555555640010,	sk = 0x0}, global = {ex_data_lock = 0x0, ex_data = {{
        meth = 0x0} <repeats 17 times>}}, dyn_indexes = {-1, -1, -1, 1,	-1, -1,	-1, -1,	3, 2}, index_locks = {
    0x555555662560, 0x555555661570, 0x555555661680, 0x5555556619c0, 0x5555556c6d00, 0x555555655720, 0x55555568ca20,
    0x55555568ca60, 0x55555568caa0, 0x55555568c360}, oncelock = 0x555555662a90, run_once_done = {0, 0, 0}, run_once_ret = {
    0, 0, 0}, onfreelist = 0x0}
```

mspncp:
My last commit fixed the fipsinstall test, but the [property test on AppVeyor](https://ci.appveyor.com/project/openssl/openssl/builds/29059506/job/op8y21jea98ciddh) is still failing:

```
    ...

    ok 6 - test_register_deregister
    ok 7 - test_property
    # ERROR: (int) 'errors < max - tail' failed @ ..\test\property_test.c:379
    # [9991] compared to [9990]
    not ok 8 - test_query_cache_stochastic
C:\projects\openssl\_build\test\property_test.exe => 1
not ok 1 - running property_test
#   Failed test 'running property_test'
#   at C:\projects\openssl\util\perl/OpenSSL/Test/Simple.pm line 77.
03-test_property.t ................. 
```

paulidale:
I've noticed that that test has been failing intermittently :(  I've not managed to get to have a good look and have no idea why it should be intermittent -- no threading, no stochastic testing.

I'd not seen this on AppVeyor before but on Travis rerunning the tests makes it go away.

paulidale:
Fixed the test problem with #10505.

richsalz:
perhaps try an address-sanitizer and/or memory-sanitizer build

mspncp:
I rebased the pr without further changes to pickup #10505.

levitte:
For some of these, particularly those that just return a "false" value (0 or NULL), an error should be added too:

``` C
        ERR_raise(ERR_LIB_XXX, ERR_R_PASSED_NULL_PARAMETER);
```

mspncp:
> For some of these, particularly those that just return a "false" value (0 or NULL), an error should be added too:
> 
> ```c
>         ERR_raise(ERR_LIB_XXX, ERR_R_PASSED_NULL_PARAMETER);
> ```

@levitte I'm not sure whether this is a misunderstanding or if not, why you picked `PASSED_NULL_PARAMETER` as error code. There is no NULL (out) parameter passed into the function. In most cases it is a about a NULL return value received from `RAND_get_rand_method()` and others. In theses cases, it is not entirely clear what caused the failure, so at best it should be the callee that places an error code on the stack.

The only reasonable error code, `RAND_R_FUNC_NOT_IMPLEMENTED` when a method callback is missing, code is returned only in a single location, although there are more possible locations where it could be returned.

https://github.com/openssl/openssl/blob/24d20043526e7f8dce2060a99f935fd95348be9b/crypto/rand/rand_lib.c#L880-L883

mspncp:
@levitte I added some error codes, see cc036f065194314e9844cd67eedab6d0a192279a. When I ran `make update` afterwards, I lost a lot of KDF error codes even though I did a reconfigure first, see d3dc8897527d37cbc1840856bd888ff5310c7a78.

From commit 4d301427a96010468da2bb67bf1025fa8d886ab9 I infer that the KDF error codes are treated specially, but that commit is in my timeline. What am I missing?

include/openssl/randerr.h
```diff
@@ -10,12 +10,6 @@
 
 #ifndef OPENSSL_RANDERR_H
 # define OPENSSL_RANDERR_H
-# pragma once
-
-# include <openssl/macros.h>
-# ifndef OPENSSL_NO_DEPRECATED_3_0
-#  define HEADER_RANDERR_H
-# endif
```
mspncp: This missing compatibility header is a mystery to me, too.

mspncp:
> @levitte I added some error codes, see [cc036f0](https://github.com/openssl/openssl/commit/cc036f065194314e9844cd67eedab6d0a192279a). When I ran `make update` afterwards, I lost a lot of KDF error codes even though I did a reconfigure first, see [d3dc889](https://github.com/openssl/openssl/commit/d3dc8897527d37cbc1840856bd888ff5310c7a78).

This problem seems to be independent of this pull request. I get the same effect, if I add an arbitrary error message somewhere on the master branch and then run `make update`. For example, I applied this patch to master:

```diff
diff --git a/crypto/rand/rand_lib.c b/crypto/rand/rand_lib.c
index c6b4ad3ae4..b747aa7cec 100644
--- a/crypto/rand/rand_lib.c
+++ b/crypto/rand/rand_lib.c
@@ -861,6 +861,8 @@ int rand_priv_bytes_ex(OPENSSL_CTX *ctx, unsigned char *buf, int num)
     if (drbg == NULL)
         return 0;
 
+    RANDerr(RAND_F_RAND_PRIV_BYTES_EX, RAND_R_FUNC_NOT_IMPLEMENTED);
+
     ret = RAND_DRBG_bytes(drbg, buf, num);
     return ret;
 }
```

mspncp:
I removed all unwanted changes made by `make update` and replaced the last commit by 5630bb87fabdf25b37126859438633c600b88b01.

mspncp:
For the record: issue #10522 has been created to address https://github.com/openssl/openssl/pull/10483#issuecomment-558395257.

levitte:
Please merge this, @paulidale

paulidale:
Merged to master.  Thanks.

crypto/rand/rand_lib.c
t8m: I do not think the RAND_add/RAND_seed should be deprecated. They should be clearly declared as unnecessary in documentation but there might be special environments and applications where such calls can be handy. Just looking at rand_unix.c what about the case of OPENSSL_SYS_VXWORKS or OPENSSL_SYS_UEFI?

paulidale: I'm sceptical that applications will pass decent randomness to `RAND_{add,seed}`...  I've seen far too many poor attempts that were about as good as doing nothing.

With a HW RNG, it's possible but if the OS doesn't supply any/enough, where will a user find more?

t8m: Having the call available means that if you have one application for a special device where openssl is compilable but does not have implemented any method to poll for entropy in the current sources, but there is for example HW RNG source, you can still workaround this just by the call from application. Otherwise it would require patching the openssl code. What we would get from the deprecation when we will have to have some call like this internally for the RAND_poll anyway.

PR #10489, opened 2019-11-20
Enable KTLS on connections using BIO_connect

bsdjhb:
The first commit introduces a potential behavior change that it would be good to check with on the Linux side in that ktls_start() might now be called on listen sockets and since BIO_socket can be used to create listen sockets (for example by bss_acpt.c).  Presumably Linux just ignores the socket option in ktls_start on such a socket, but it would be good to confirm that this is ok.

The second commit is in a response to a core dump on FreeBSD when 'openssl s_time' switches to reusing sessions since free() of a stack pointer triggers a crash inside of jemalloc (the libc malloc in FreeBSD).

### DIFF
--- a/crypto/bio/b_sock2.c
+++ b/crypto/bio/b_sock2.c
@@ -12,6 +12,7 @@
 #include <errno.h>
 
 #include "bio_local.h"
+#include "internal/ktls.h"
 
 #include <openssl/err.h>
 
@@ -51,6 +52,17 @@ int BIO_socket(int domain, int socktype, int protocol, int options)
         BIOerr(BIO_F_BIO_SOCKET, BIO_R_UNABLE_TO_CREATE_SOCKET);
         return INVALID_SOCKET;
     }
+# ifndef OPENSSL_NO_KTLS
+    {
+        /*
+         * The new socket is created successfully regardless of ktls_enable.
+         * ktls_enable doesn't change any functionality of the socket, except
+         * changing the setsockopt to enable the processing of ktls_start.
+         * Thus, it is not a problem to call it for non-TLS sockets.
+         */
+        ktls_enable(sock);
+    }
+# endif
 
     return sock;
 }
--- a/crypto/bio/bss_conn.c
+++ b/crypto/bio/bss_conn.c
@@ -11,6 +11,7 @@
 #include <errno.h>
 
 #include "bio_local.h"
+#include "internal/ktls.h"
 
 #ifndef OPENSSL_NO_SOCK
 
@@ -20,6 +21,9 @@ typedef struct bio_connect_st {
     char *param_hostname;
     char *param_service;
     int connect_mode;
+# ifndef OPENSSL_NO_KTLS
+    unsigned char record_type;
+# endif
 
     BIO_ADDRINFO *addr_first;
     const BIO_ADDRINFO *addr_iter;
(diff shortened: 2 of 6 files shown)


crypto/bio/b_sock2.c
```diff
@@ -51,6 +52,17 @@ int BIO_socket(int domain, int socktype, int protocol, int options)
...
     }
+# ifndef OPENSSL_NO_KTLS
+    {
+        /*
+         * The new socket is created successfully regardless of ktls_enable.
+         * ktls_enable doesn't change any functionality of the socket, except
+         * changing the setsockopt to enable the processing of ktls_start.
+         * Thus, it is not a problem to call it for non-TLS sockets.
+         */
+        ktls_enable(sock);
```
BorisPis: This would be problematic in Linux as the TCP_ULP socket option will fail for sockets that are not in the ESTABLISHED state. I realize that this might be too strict, and we should try to fix it in the Kernel.

bsdjhb: To be clear, does Linux panic, or does it return an error in this case?  Nothing is checking the error from ktls_enable except in the KTLS tests, and those are only used on non-listen sockets.  If the setsockopt() call fails with an error but otherwise doesn't disturb the socket, then I don't think this will hurt anything?

bsdjhb:
The commit message for the first change (and my initial comment above) both need to say "ktls_enable()" rather than "ktls_start()"

crypto/bio/b_sock2.c
BorisPis: It returns an error without panicking, so it just won't work for BIO_TYPE_CONNECT as it is today. So this change should be fine.

bsdjhb:
Ping

mattcaswell:
@openssl/committers needs second review

crypto/bio/bss_conn.c
```diff
@@ -333,7 +342,16 @@ static int conn_write(BIO *b, const char *in, int inl)
...
+# ifndef OPENSSL_NO_KTLS
+    if (BIO_should_ktls_ctrl_msg_flag(b)) {
+        ret = ktls_send_ctrl_message(b->num, data->record_type, in, inl);
+        if (ret >= 0) {
+            ret = inl;
+            BIO_clear_ktls_ctrl_msg_flag(b);
+        }
+    } else
+# endif
+      ret = writesocket(b->num, in, inl);
```
t8m: wrong indentation here - should be 2 spaces more.

richsalz: And now you've got conflicting style rules because if {} means else{}.  Perhaps best to duplicate the line of code and use an #else for the non-ktls case.

t8m: I think it would be reasonable exception to ignore this rule. However the indentation is wrong anyway.

bsdjhb: Oops, thanks.  Fixed.

t8m:
@mattcaswell  Does your approval still hold?

crypto/bio/b_sock2.c
rmchelsio: This change also enables use of TLS with TOE on Chelsio adapters in Linux.

t8m:
Merged to master, thank you.

PR #10497, opened 2019-11-21
chacha/asm/chacha-armv8.pl: preserve FP registers d8 and d9 correctly

### LINKED ISSUE #10470: Regression in ChaCha20 on 64 bit ARM
The commit 86a7ac5e76fd3df9a5c3422153ae5a3e4f99b3a8 broke the 64 bit ARM implementation of ChaCha20.

Running the evp_test on evpciph_chacha.txt segfaults (due to memory corruption) on the first test data that is larger than 512 bytes (it has 640 plaintext bytes to encrypt).

ardbiesheuvel:
Depending on the size of the input, we may take different paths through
the accelerated arm64 ChaCha20 routines, each of which use a different
subset of the FP registers, some of which need to be preserved and
restored, as required by the AArch64 calling convention (AAPCS64)

In some cases, (e.g., when the input size is 640 bytes), we call the 512
byte NEON path followed directly by the scalar path, and in this case,
we preserve and restore d8 and d9, only to clobber them again
immediately before handing over to the scalar path which does not touch
the FP registers at all, and hence does not restore them either.

Fix this by moving the restoration of d8 and d9 to a later stage in the
512 byte routine, either before calling the scalar path, or when exiting
the function.

Fixes #10470
CLA: trivial

### DIFF
--- a/crypto/chacha/asm/chacha-armv8.pl
+++ b/crypto/chacha/asm/chacha-armv8.pl
@@ -1232,8 +1232,7 @@ $code.=<<___;
 	adds	$len,$len,#512
 	ushr	$ONE,$ONE,#1			// 4 -> 2
 
-	ldp	d8,d9,[sp,#128+0]		// meet ABI requirements
-	ldp	d10,d11,[sp,#128+16]
+	ldp	d10,d11,[sp,#128+16]		// meet ABI requirements
 	ldp	d12,d13,[sp,#128+32]
 	ldp	d14,d15,[sp,#128+48]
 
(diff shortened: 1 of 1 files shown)


ardbiesheuvel:
I should be covered by my company's CLA (Arm Ltd)

ardbiesheuvel:
Ah never mind, we have the CCLA but I don't have the ICLA

paulidale:
Confirming that the ICLA is missing and the CCLA is good.

For a trivial commit, the `CLA: trivial` needs to be in the commit message.

mattcaswell:
> Ah never mind, we have the CCLA but I don't have the ICLA

It would be nice if we could have your ICLA and then we can accept this and any subsequent contributions without having to worry about CLAs at all.

levitte:
@dot-asm has an alternative suggestion in #10470, and I see it being actively discussed.  I would like to see the outcome of that discussion before this gets merged.

ardbiesheuvel:
> @dot-asm has an alternative suggestion in #10470, and I see it being actively discussed. I would like to see the outcome of that discussion before this gets merged.

I'll gladly defer to @dot-asm to decide how this should be fixed.

levitte:
Everyone involved seems to agree with this PR at this point, see https://github.com/openssl/openssl/issues/10470#issuecomment-557490376

t8m:
Merged to master as 07a470729c

PR #10501, opened 2019-11-22
Remove deadlock that was caused by calling pthread_rwlock_wrlock() 

MatusKysel:
I have try to build current master with musl instead of glibc and I have found that fuzz test will get stucked. Issue seems to be deadlock caused by calling pthread_rwlock_wrlock() on same thread. I guess glibc is not following POSIX so strictly, because by description this may cause this behaviour "The calling thread may deadlock if at the time the call is made it holds the read-write lock (whether a read or write lock)."

CLA: trivial

Signed-off-by: Matus Kysel <mkysel@tachyum.com>


levitte:
> Agreeing that this is trivial.

Yup

paulidale:

PR #10502, opened 2019-11-22
Add X509_cmp_timeframe()

DDvO:
This PR adds the function `X509_cmp_timeframe()` along with documentation and thorough tests.
It also fixes a parameter renaming glitch in the current documentation of `X509_cmp_time()`.

The function can be used for instance as follows:
```
    X509_VERIFY_PARAM *vpm = ... /* may include flags and reference time used below */
    X509 *cert = ...
    int time_cmp = OSSL_CMP_cmp_timeframe(vpm,
                                          X509_get0_notBefore(cert),
                                          X509_get0_notAfter(cert));
    if (time_cmp > 0)
        error("certificate expired");
    if (time_cmp < 0)
        error("certificate not yet valid");
```

The motivation for adding this function to `x509_vfy.c` is that it is currently defined in a preview of the upcoming CMP contribution chunk 7 while it could be of broader use, as discussed with @mattcaswell here: https://github.com/mpeylo/cmpossl/pull/199#discussion_r349115202

### DIFF
--- a/crypto/x509/x509_vfy.c
+++ b/crypto/x509/x509_vfy.c
@@ -1851,6 +1851,31 @@ int X509_cmp_time(const ASN1_TIME *ctm, time_t *cmp_time)
     return ret;
 }
 
+/*
+ * Return 0 if time should not be checked or reference time is in range,
+ * or else 1 if it is past the end, or -1 if it is before the start
+ */
+int X509_cmp_timeframe(const X509_VERIFY_PARAM *vpm,
+                       const ASN1_TIME *start, const ASN1_TIME *end)
+{
+    time_t ref_time;
+    time_t *time = NULL;
+    unsigned long flags = vpm == NULL ? 0 : X509_VERIFY_PARAM_get_flags(vpm);
+
+    if ((flags & X509_V_FLAG_USE_CHECK_TIME) != 0) {
+        ref_time = X509_VERIFY_PARAM_get_time(vpm);
+        time = &ref_time;
+    } else if ((flags & X509_V_FLAG_NO_CHECK_TIME) != 0) {
+        return 0; /* this means ok */
+    } /* else reference time is the current time */
+
+    if (end != NULL && X509_cmp_time(end, time) < 0)
+        return 1;
+    if (start != NULL && X509_cmp_time(start, time) > 0)
+        return -1;
+    return 0;
+}
+
 ASN1_TIME *X509_gmtime_adj(ASN1_TIME *s, long adj)
 {
     return X509_time_adj(s, adj, NULL);
--- a/doc/man3/X509_cmp_time.pod
+++ b/doc/man3/X509_cmp_time.pod
@@ -19,7 +19,7 @@ X509_time_adj, X509_time_adj_ex
 =head1 DESCRIPTION
 
 X509_cmp_time() compares the ASN1_TIME in B<asn1_time> with the time
-in <cmp_time>.
+in <in_tm>.
 
 X509_cmp_current_time() compares the ASN1_TIME in
 B<asn1_time> with the current time, expressed as time_t.
@@ -53,7 +53,7 @@ X509_cmp_current_time() return 0 on error.
 =head1 RETURN VALUES
 
 X509_cmp_time() and X509_cmp_current_time() return -1 if B<asn1_time>
-is earlier than, or equal to, B<cmp_time> (resp. current time), and 1
+is earlier than, or equal to, B<in_tm> (resp. current time), and 1
 otherwise. These methods return 0 on error.
 
 X509_cmp_timeframe() returns 0 if B<vpm> is not NULL and the verification
--- a/doc/man3/X509_cmp_time.pod
+++ b/doc/man3/X509_cmp_time.pod
@@ -2,13 +2,16 @@
 
 =head1 NAME
 
-X509_cmp_time, X509_cmp_current_time, X509_time_adj, X509_time_adj_ex
+X509_cmp_time, X509_cmp_current_time, X509_cmp_timeframe,
+X509_time_adj, X509_time_adj_ex
 - X509 time functions
 
 =head1 SYNOPSIS
 
  int X509_cmp_time(const ASN1_TIME *asn1_time, time_t *in_tm);
  int X509_cmp_current_time(const ASN1_TIME *asn1_time);
+ int X509_cmp_timeframe(const X509_VERIFY_PARAM *vpm,
+                        const ASN1_TIME *start, const ASN1_TIME *end);
  ASN1_TIME *X509_time_adj(ASN1_TIME *asn1_time, long offset_sec, time_t *in_tm);
  ASN1_TIME *X509_time_adj_ex(ASN1_TIME *asn1_time, int offset_day, long
                              offset_sec, time_t *in_tm);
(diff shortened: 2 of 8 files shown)


crypto/x509/x509_vfy.c
```diff
@@ -1851,6 +1851,31 @@ int X509_cmp_time(const ASN1_TIME *ctm, time_t *cmp_time)
     return ret;
 }
 
+/*
+ * Return 0 if time should not be checked or reference time is in range,
+ * or else 1 if it is past the end, or -1 if it is before the start
+ */
+int X509_cmp_timeframe(const X509_VERIFY_PARAM *vpm,
+                       const ASN1_TIME *start, const ASN1_TIME *end)
```
mattcaswell: I notice that the function operates on an `X509_VERIFY_PARAM`, and that makes me wonder whether this would be better called `X509_VERIFY_PARAM_cmp_timeframe`. Normally the initial part of the name "X509" implies the object that we operate on. If we rename it, then it is also probably better placed in the x509_vpm.c file. OTOH, I notice that other X509_* time functions don't conform to this, and also that vpm can be NULL, which is somewhat unusual.

So, I'd like other opinions on this.,

doc/man3/X509_cmp_time.pod
```diff
@@ -2,24 +2,31 @@
...
+X509_cmp_time, X509_cmp_current_time, X509_cmp_timeframe,
+X509_time_adj, X509_time_adj_ex
 - X509 time functions
 
 =head1 SYNOPSIS
 
  int X509_cmp_time(const ASN1_TIME *asn1_time, time_t *in_tm);
  int X509_cmp_current_time(const ASN1_TIME *asn1_time);
+ int X509_cmp_timeframe(const X509_VERIFY_PARAM *vpm,
+                        const ASN1_TIME *start, const ASN1_TIME *end);
```
mattcaswell: Can we add a HISTORY section to this file noting that the function was added in OpenSSL 3.0?

test/x509_time_test.c
```diff
@@ -297,6 +297,56 @@ static int test_x509_cmp_time_current(void)
...
+static int test_X509_cmp_timeframe_vpm(X509_VERIFY_PARAM *vpm,
+                                       ASN1_TIME *asn1_before,
+                                       ASN1_TIME *asn1_mid,
+                                       ASN1_TIME *asn1_after)
+{
+    int always_0 = vpm != NULL
+        && (X509_VERIFY_PARAM_get_flags(vpm) & X509_V_FLAG_USE_CHECK_TIME) == 0
+        && (X509_VERIFY_PARAM_get_flags(vpm) & X509_V_FLAG_NO_CHECK_TIME) != 0;
+
+    return asn1_before != NULL && asn1_mid != NULL && asn1_after != NULL
```
mattcaswell: Probably should use TEST_ptr for these

crypto/x509/x509_vfy.c
DDvO: Another option would be to change the order of the parameters, putting `vpm` last.
Yet since the semantics of the function is governed by the `vpm` parameter I appeared more intuitive to have it first, but others may see this differently.

doc/man3/X509_cmp_time.pod
DDvO: Sure, done.

test/x509_time_test.c
DDvO: I don't think adding `TEST_ptr` here would be appropriate because these pointers/structures are not under test. The `NULL` comparisons are needed just as guards.

DDvO:
The CI tests pointed out that a `const` glitch, which I solved essentially by constifying `X509_VERIFY_PARAM_get_flags` (which was still missing anyway).

include/openssl/x509_vfy.h
```diff
@@ -558,7 +558,7 @@ int X509_VERIFY_PARAM_set_flags(X509_VERIFY_PARAM *param,
                                 unsigned long flags);
 int X509_VERIFY_PARAM_clear_flags(X509_VERIFY_PARAM *param,
                                   unsigned long flags);
-unsigned long X509_VERIFY_PARAM_get_flags(X509_VERIFY_PARAM *param);
+unsigned long X509_VERIFY_PARAM_get_flags(const X509_VERIFY_PARAM *param);
```
mattcaswell: Probably this part of the change should be backported. Maybe in a different PR.

richsalz:
FWIW, when I saw the function name, I thought it would take an X509 cert, not a VPM.

include/openssl/x509_vfy.h
DDvO: I've tentatively started constifying many functions in `x509_vfy.h`, indeed better in a separate PR...

DDvO: Done, see #10504

mattcaswell:
Ping - needs second review

doc/man3/X509_cmp_time.pod
```diff
@@ -43,12 +53,24 @@ X509_cmp_current_time() return 0 on error.
...
+is earlier than, or equal to, B<in_tm> (resp. current time), and 1
 otherwise. These methods return 0 on error.
 
+X509_cmp_timeframe() returns 0 if B<vpm> is not NULL and the verification
+parameters do not contain B<X509_V_FLAG_USE_CHECK_TIME>
+but do contain B<X509_V_FLAG_NO_CHECK_TIME>. Otherwise it returns
+1 if the end time is not NULL and the reference time (which has determined as
+stated above) is past the end time, -1 if the start time is not NULL and the
+reference time is before, else 0 to indicate that the reference time is in range
+(implying that the end time is not before the start time if both are present).
```
paulidale: What about changing _implying_ to _on the assumption that_.  Implying kind of means that what follows is true and it needn't be.

DDvO: I really meant 'implying' in the propositional/logical sense: 
If the result is 0 then ref time <= end and and start < ref time, which does imply start < end. So the correct range is a consequence here.
Conversely, if (as usual) start < end and the ref time is in between, then the result is 0. So the correct range can also be seen as an assumption.

DDvO:
> FWIW, when I saw the function name, I thought it would take an X509 cert, not a VPM.

Concerning the function name, right that the `X509_` prefix should indicate that the function takes a cert argument (as its first parameter). Yet this also applies to all the the other time-related functions defined in `x509_vfy.c`, and the new function just follows the naming model of its peers.

One could consider consistently renaming all these functions, including the new one, to something that does not contain `X509_` and move the functions out of `crypto/x509/` but this would be the topic of a separate PR.

crypto/x509/x509_vfy.c
levitte: Heh, the name discussion goes on...

This function doesn't *operate on* a `X509_VERIFY_PARAM` (i.e. it's not a constructor, destructor, getter or setter), the first parameter is simply *input* just as much as the others.  This function is simply part of the `X509` API, and the way I see it, the current name is consistent with the rest.  `X509_VERIFY_PARAM_cmp_timeframe` would convey a whole different message, and actually be confusing about what it does.

levitte: I could *possibly* see `X509_VERIFY_cmp_timeframe()`, but since we don't have a specific `X509_VERIFY` namespace, that simply doesn't fit.

DDvO: I agree that here the 'X509_VERIFY_PARAM *` parameter, though it us the first parameter, should not influence the name of the function.

DDvO: If we really wanna point out what the new function and its already existing friends operate on then all their names should start with `ASN1_TIME_`. Yet also this would be the topic of a different PR.

DDvO:
Can this PR be flagged for merging now?

levitte:
> Can this PR be flagged for merging now?

The conclusive approval was done 16 hours ago.  So, with regards to 24h grace period that we have regimented ourselves to, it can in 8 hours.

DDvO:
In the meantime I had to rebase this (fixing a minor merge conflict in `libcrypto.num`).

Could someone please merge this PR.

mattcaswell:
I squashed the various commits and merged this. Closing.

Issue #3742, opened 2017-06-21
more const


richsalz:
We could make more things const -- in particular API parameters.
There's been discussion on whether not adding const to an API breaks ABI, but for now let's just record the issue as something we want to address.

levitte:
I agree with you (except I'd say that *only* API parameters can be safely consitified).
Some care must be taken, though...  there are some types that can't be constified as things currently stand, even as input parameters, because they cache things in the structure on the fly.  `X509` is one of them.

mattcaswell:
Moving this to post-1.1.1 milestone.

DDvO:
https://github.com/openssl/openssl/pull/7646/commits/168824237fd33e634b9e3f021b8404ea758e5871 provides some more const, mostly on  `*_dup()` and `*i2d_*()` and related functions.

DDvO:
PR #8029 provides some more const, mostly on  `*_dup()` and `*i2d_*()` and related functions.

t8m:
This is too broad issue. Closing. If there is anything that can be easily constified without an API break, feel free to open a concrete issue or even better a PR.

PR #10504, opened 2019-11-22
constify many functions in x509_vfy.h etc.

DDvO:
Due to a little `const` incident in #10502 I've tentatively constified most of `x509_vfy.h` and friends.
As far as folks are fine with these changes I'll update the respective doc files accordingly.

I had to introduce type casts from `const X509_NAME*` to non-const `X509_NAME` at five places where `name` parameter pointers are stored in a local structure, but these should be harmless because the structure contents are used only for lookup and are not modified.


levitte:
Why not make the name attribute in those local structs const?

DDvO:
> Why not make the name attribute in those local structs const?

Can this be done for situations like this in `x509_lu.c`?
```
    X509 x509_s;
...

        x509_s.cert_info.subject = (X509_NAME *)name;
```

DDvO:
BTW, I find the programming pattern in `x509_object_idx_cnt()`, as well as in the places where I had to apply the de-constification type cast, pretty awkward: 
whole - pretty fat - `X509` and `X509_CRL` structures are allocated on the stack as auto variables and get very sparsely initialized, just to serve as reference objects for cert/CRL comparison. 
Directly providing one or two reference values (e.g., subject name or issuer name and serial) to a generalized comparison function would be sufficient.

DDvO:
Looking for a solution that avoids the type casts to non-const,
I first experimented with using a direct initializer expression like 
```
    X509_CRL crl_s = { NULL, { 0, }, name, };
```
but this did not work for two reasons: the compiler cannot handle non-static initialization (with the `name` parameter), and even if it could, it (naturally) reports the same `const` conflict.

What would work is to copy the name:
```
    x509_s.cert_info.subject =X509_NAME_dup(name);
```
and later free it, but this would be inefficient.

DDvO:
Then I realized that it would be much better to replace the `X509_NAME *` types of the `subject` and `issuer` fields of `X509`, `X509_REQ`, `X509_CRL`, and `PKCS7_ISSUER_AND_SERIAL` structures by `const X509_NAME *` (and consequently also for the `directoryName` field of `GENERAL_NAME` etc.). 
This allows for much more constification of function parameters and is also pretty natural because it resembles `const char *`.

Doing so I had to introduce just one type cast to non-const in
```
    X509_NAME_free((X509_NAME *)*xn);
```
of `X509_NAME_set()`, which I had to fully constify as
```
int X509_NAME_set(const X509_NAME **xn, const X509_NAME *name)
```
I added the following comment there for justification:
```
/*
 * The use of 'const' for, e.g., the subject and issuer fields of X509 and
 * X509_CRL structures etc. pointed to by *xn, as well as the cast to
 * non-const 'X509_NAME *' in the below call of X509_NAME_free(), should be okay
 * because due to the modification of the pointer value of the field
 * the original pointer value has been invalidated anyway, such that any aliases
 * are not usable any more, neither for reading nor for writing name contents.
 */
```
I also had to improve a couple of fishy code sections, e.g., in `apps/`.

Due to the new constifications I was able to remove again the five casts to non-const that I had introduced yesterday. I also found some new opportunities for little code simplifications, which I did.
I also found and eliminated some casts to non-const that were no longer needed for instance due to my earlier constification of ASN.1 `*_dup()` functions.

include/crypto/x509.h
```diff
@@ -55,7 +55,7 @@ struct x509_sig_info_st {
 struct X509_req_info_st {
     ASN1_ENCODING enc;          /* cached encoding of signed part */
     ASN1_INTEGER *version;      /* version, defaults to v1(0) so can be NULL */
-    X509_NAME *subject;         /* certificate request DN */
+    const X509_NAME *subject;   /* certificate request DN */
```
kroeckx: I don't think we want to make a change like that. This can potentially break other applications.

kroeckx: This seems to be a local include file, that's fine then.

include/openssl/cms.h
```diff
@@ -313,16 +313,17 @@ int CMS_RecipientInfo_kari_get0_orig_id(CMS_RecipientInfo *ri,
                                         X509_ALGOR **pubalg,
                                         ASN1_BIT_STRING **pubkey,
                                         ASN1_OCTET_STRING **keyid,
-                                        X509_NAME **issuer,
-                                        ASN1_INTEGER **sno);
+                                        const X509_NAME **issuer,
+                                        const ASN1_INTEGER **sno);
```
kroeckx: This is an incompatible change. Don't change it for a **

include/openssl/crmf.h
```diff
@@ -121,8 +121,10 @@ int OSSL_CRMF_MSGS_verify_popo(const OSSL_CRMF_MSGS *reqs,
                                int rid, int acceptRAVerified);
 OSSL_CRMF_CERTTEMPLATE *OSSL_CRMF_MSG_get0_tmpl(const OSSL_CRMF_MSG *crm);
 ASN1_INTEGER *OSSL_CRMF_CERTTEMPLATE_get0_serialNumber(OSSL_CRMF_CERTTEMPLATE *t);
-X509_NAME *OSSL_CRMF_CERTTEMPLATE_get0_issuer(OSSL_CRMF_CERTTEMPLATE *tmpl);
-X509_NAME *OSSL_CRMF_CERTID_get0_issuer(const OSSL_CRMF_CERTID *cid);
+const X509_NAME
+    *OSSL_CRMF_CERTTEMPLATE_get0_issuer(const OSSL_CRMF_CERTTEMPLATE *tmpl);
+const X509_NAME
+    *OSSL_CRMF_CERTID_get0_issuer(const OSSL_CRMF_CERTID *cid);
```
kroeckx: Don't change it for return types

kroeckx: But it's a new file, so this should be fine.

include/openssl/pkcs7.h
```diff
@@ -36,7 +36,7 @@ Key_Encryption_ID       rsaEncryption
 */
 
 typedef struct pkcs7_issuer_and_serial_st {
-    X509_NAME *issuer;
+    const X509_NAME *issuer;
```
kroeckx: Don't change it for public structs.

include/openssl/store.h
```diff
@@ -191,7 +191,8 @@ void OSSL_STORE_SEARCH_free(OSSL_STORE_SEARCH *search);
 
 /* Search term accessors */
 int OSSL_STORE_SEARCH_get_type(const OSSL_STORE_SEARCH *criterion);
-X509_NAME *OSSL_STORE_SEARCH_get0_name(const OSSL_STORE_SEARCH *criterion);
+const X509_NAME *OSSL_STORE_SEARCH_get0_name(const OSSL_STORE_SEARCH
+                                             *criterion);
```
kroeckx: Don't change the return type of existing public functions.

include/openssl/x509.h
```diff
@@ -547,7 +548,7 @@ DECLARE_ASN1_FUNCTIONS(X509_NAME_ENTRY)
 
 DECLARE_ASN1_FUNCTIONS(X509_NAME)
 
-int X509_NAME_set(X509_NAME **xn, const X509_NAME *name);
+int X509_NAME_set(const X509_NAME **xn, const X509_NAME *name);
```
kroeckx: Note for **

include/openssl/x509.h
```diff
@@ -644,10 +645,10 @@ int X509_set_version(X509 *x, long version);
...
 ASN1_INTEGER *X509_get_serialNumber(X509 *x);
 const ASN1_INTEGER *X509_get0_serialNumber(const X509 *x);
-int X509_set_issuer_name(X509 *x, X509_NAME *name);
-X509_NAME *X509_get_issuer_name(const X509 *a);
-int X509_set_subject_name(X509 *x, X509_NAME *name);
-X509_NAME *X509_get_subject_name(const X509 *a);
+int X509_set_issuer_name(X509 *x, const X509_NAME *name);
+const X509_NAME *X509_get_issuer_name(const X509 *a);
+int X509_set_subject_name(X509 *x, const X509_NAME *name);
+const X509_NAME *X509_get_subject_name(const X509 *a);
```
kroeckx: Not for the return type

include/openssl/x509.h
```diff
@@ -683,8 +684,8 @@ int X509_certificate_type(const X509 *x, const EVP_PKEY *pubkey);
 
 long X509_REQ_get_version(const X509_REQ *req);
 int X509_REQ_set_version(X509_REQ *x, long version);
-X509_NAME *X509_REQ_get_subject_name(const X509_REQ *req);
-int X509_REQ_set_subject_name(X509_REQ *req, X509_NAME *name);
+const X509_NAME *X509_REQ_get_subject_name(const X509_REQ *req);
```
kroeckx: Not for the return type

include/openssl/x509_vfy.h
```diff
@@ -80,9 +80,9 @@ typedef int (*X509_STORE_CTX_cert_crl_fn)(X509_STORE_CTX *ctx,
                                           X509_CRL *crl, X509 *x);
 typedef int (*X509_STORE_CTX_check_policy_fn)(X509_STORE_CTX *ctx);
 typedef STACK_OF(X509) *(*X509_STORE_CTX_lookup_certs_fn)(X509_STORE_CTX *ctx,
-                                                          X509_NAME *nm);
-typedef STACK_OF(X509_CRL) *(*X509_STORE_CTX_lookup_crls_fn)(X509_STORE_CTX *ctx,
-                                                             X509_NAME *nm);
+                                                          const X509_NAME *nm);
+typedef STACK_OF(X509_CRL)
+    *(*X509_STORE_CTX_lookup_crls_fn)(X509_STORE_CTX *ctx, const X509_NAME *nm);
```
kroeckx: I'm not sure about this one.

kroeckx:
For backwards compatibility, you need to be careful with everything in the public headers. Some const changes will require the applications to also change it to const, which is not something we want to force them to do (at this time.)

include/openssl/cms.h
DDvO: Thanks @kroeckx for having a closer look at this commit and providing feedback.

Why should adding `const ` to a function parameter type be an incompatible change?

include/openssl/pkcs7.h
DDvO: I thought there is some freedom to adapt types where useful for v3.0, or am I wrong on this?

include/openssl/store.h
DDvO: Well, I know that this can cause compatibility issues, but maybe this is acceptable for v3.0?

include/openssl/x509.h
DDvO: Also here (adding `const ` to the types of function parameters) I see no compatibility issue.

include/openssl/x509.h
DDvO: See my above question regarding v3.0

BTW, IMHO these (and other similar) getters should be renamed from `_get_`to `_get0_`.
Can we do this (adding compatibility aliases)?

include/openssl/x509.h
DDvO: See my respective question above

include/openssl/x509_vfy.h
DDvO: Also here, adding `const` (to param types) should be harmless.

DDvO:
@kroeckx, I understand the point regarding backward compatibility.
Yet is there then a chance at all to correct such issues where `const` would make much sense?

If I revert my improvements regarding `const` getter responses (and `const` `X509_NAME *` fields in structures) then most of the last commit would need to be reverted.

DDvO:
I actually felt the incompatibilities of constifying `X509_NAME *` getter results.
These materialize in particular in `ssl/` and `apps/`.
Yet having a look at these it turns out that the changes needed are bearable, and actually contribute to code cleanliness.

kroeckx:
OpenSSL 3.0 is supposed to remain as compatible as possible, we
expect most applications to just need to rebuild and not make any
changes. Anything that requires them to make changes needs to wait
until at least 4.0.

I think there are 3 types that I complained about:
- Changing the return type to const
- Changing a "TYPE **" parameter to a "const TYPE **"
- Chaning a public member of a struct to const.

They all require the users to also change their variable to const.

DDvO:
I see - thanks for the clarification!

include/openssl/store.h
slontis: 3.0 is trying to limit the amount of breaking changes..

include/openssl/cms.h
DDvO: Meanwhile I understand that adding `const` to a `**` parameter (as opposed to adding it to a `*` parameter) of a function can cause trouble (because the function could use it to return a pointer to `const` and then the caller could use it as pointer to mutable) and is therefore considered incompatible.

kaduk: It's not even the *change* that is incompatible; modern compilers will complain about having ```const foo **x``` in a function prototype (as opposed to ```foo const * const *x```, recalling that const binds to the left)

DDvO: Thanks @kaduk for this hint!
Using double `const` turned out to be the right solution (AFAICS) for situations where the function (for whatever reasons) takes a double pointer argument but does not modify the pointer nor the contents it points to.

include/openssl/x509.h
DDvO: I've reverted this change.

include/openssl/x509.h
DDvO: I've reverted these changes on the return type as well, adding a comment here that the functions should better have `get0_` in their names (rather than `get_`).

include/openssl/x509_vfy.h
DDvO: Here we don't face the double pointer issue.

include/openssl/x509.h
DDvO: Reverted as well for backward compatibility.

DDvO:
@kroeckx, finally I've found the time to execute all requested changes.
Hope everything is fine/compatible now.

kroeckx:
I've just did a quick look at this, and currently don't have any objections. Didn't have time to do a full review.

DDvO:
Thanks nevertheless for your quick scan & response!

DDvO:
This PR contributes to the long-standing issue #3742 ("more const").

DDvO:
> I've just did a quick look at this, and currently don't have any objections. Didn't have time to do a full review.

@kroeckx, when do you think you'll be able to do a full review?

DDvO:
I've just picked up this stalled PR, rebasing and simplifying it.

@kroeckx, can you resume reviewing this soon, or who else?

kroeckx:
I have no idea when I'll find the time to review this

DDvO:
> I have no idea when I'll find the time to review this

I see. Anyone willing to fill in?

Actually after my recent simplifications the PR consists basically just of a large number of straightforward additions of `const` (plus some re-formatting where needed). 
Where API compatibility (with pre-v.3.0 versions) is an issue it adds `const` only for function arguments, where it should be harmless.

I've also added a couple of comments why certain constifications cannot be done (at least as long as compatibility is an issue).

Since meanwhile I'm confident that the set of changes done in this PR has stabilized I've updated the documentation accordingly.

t8m:
Please rebase

crypto/cmp/cmp_local.h
```diff
@@ -93,9 +93,9 @@ struct ossl_cmp_ctx_st {
     /* certificate template */
     EVP_PKEY *newPkey; /* explicit new private/public key for cert enrollment */
     int newPkey_priv; /* flag indicating if newPkey contains private key */
-    X509_NAME *issuer; /* issuer name to used in cert template */
+    const X509_NAME *issuer; /* issuer name to used in cert template */
```
t8m: I am not sure about this change if it forces you to cast away the const in the free calls above.

crypto/cmp/cmp_local.h
```diff
@@ -93,9 +93,9 @@ struct ossl_cmp_ctx_st {
...
     int days; /* Number of days new certificates are asked to be valid for */
-    X509_NAME *subjectName; /* subject name to be used in the cert template */
+    const X509_NAME *subjectName; /* subject name to be used in cert template */
```
t8m: And here.

crypto/cmp/cmp_ctx.c
```diff
@@ -590,7 +590,7 @@ int OSSL_CMP_CTX_set1_##FIELD(OSSL_CMP_CTX *ctx, const TYPE *val) \
     \
     if (val != NULL && (val_dup = TYPE##_dup(val)) == NULL) \
         return 0; \
-    TYPE##_free(ctx->FIELD); \
+    TYPE##_free((TYPE *)ctx->FIELD);                  \
```
t8m: This is really suspicious. The FIELD should not be const then.

crypto/cmp/cmp_local.h
DDvO: I believe this is no real problem because the `..._free()` falls discard the contents pointed to by the given argument.
Yet to avoid further delays of this PR due to potentially onging discussion I've removed the `const` of the `X509_NAME *` fields (and consequently the cast to non-const on deallocation).

crypto/cmp/cmp_local.h
DDvO: See above

crypto/cmp/cmp_ctx.c
DDvO: See above

DDvO:
Thanks @t8m for taking over the review of this (simplified) PR!
I've rebased it and handled all open issues.

t8m:
The Travis failure is just a timeout.

t8m:
Please squash the commits and add some descriptive commit message when merging.

DDvO:
Merged, squashing and adding cumulative commit message.
Thanks @kroeckx and @t8m!

PR #10505, opened 2019-11-22
Property: fix intermittent test failure.

paulidale:
The reduction in the cache flush threshold in #10408 caused the stochastic test
to fail with noticeable probability.  Revert that part of the change.


paulidale:
AppVeyor failure isn't relevant.

paulidale:
Merged to master.  Thanks for the review.

PR #10511, opened 2019-11-25
Set argument only after successful dup on CMP APIs

raja-ashok:
Changes done to set the argument passed only after successful dup.


paulidale:
This seems to have dropped off the radar.  Merged to master.

PR #10516, opened 2019-11-25
test/cipher_overhead_test.c: build unconditionally

levitte:
Build it against static libraries always, since that's the only way it
can work as intended.


paulidale (approved):
LGTM.

levitte:
Merged

b7f491cbd65116761da0565895069431b6cb8ef3 test/cipher_overhead_test.c: build unconditionally

PR #10518, opened 2019-11-26
Optimize AES-ECB mode in OpenSSL for both aarch64 and aarch32

xkqian:
Aes-ecb mode can be optimized by inverleaving cipher operation on
several blocks and loop unrolling. Interleaving needs one ideal
unrolling factor, here we adopt the same factor with aes-cbc,
which is described as below:
    If blocks number > 5, select 5 blocks as one iteration,every
    loop, decrease the blocks number by 5.
    If 3 < left blocks < 5 select 3 blocks as one iteration, every
    loop, decrease the block number by 3.
    If left blocks < 3, treat them as tail blocks.
Detailed implementation will have a little adjustment for squeezing
code space.
With this way, for small size such as 16 bytes, the performance is
similar as before, but for big size such as 16k bytes, the performance
improves a lot, even reaches to 100%, for some arches such as A57,
the improvement  even exceeds 100%. The following table will list the
encryption performance data on aarch64, take a72 and a57 as examples.
Performance value takes the unit of cycles per byte, takes the format
as comparision of values. List them as below:
A72(freq: 2.4G):
Benchmark name                  Before optimization        After optimization      Improve
evp-aes-128-ecb@16           17.26538237                   16.82663866               2.61%
evp-aes-128-ecb@64           5.50528499                     5.222637557               5.41%
evp-aes-128-ecb@256         2.632700213                   1.908442892               37.95%
evp-aes-128-ecb@1024       1.876102047                   1.078018868               74.03%
evp-aes-128-ecb@8192       1.6550392                       0.853982929               93.80%
evp-aes-128-ecb@16384      1.636871283                  0.847623957               93.11%
evp-aes-192-ecb@16            17.73104961                  17.09692468               3.71%
evp-aes-192-ecb@64            5.78984398                    5.418545192               6.85%
evp-aes-192-ecb@256          2.872005308                  2.081815274               37.96%
evp-aes-192-ecb@1024        2.083226672                  1.25095642                 66.53%
evp-aes-192-ecb@8192        1.831992057                  0.995916251               83.95%
evp-aes-192-ecb@16384      1.821590009                  0.993820525               83.29%
evp-aes-256-ecb@16            18.0606306                    17.96963317               0.51%
evp-aes-256-ecb@64            6.19651997                    5.762465812               7.53%
evp-aes-256-ecb@256          3.176991394                  2.24642538                 41.42%
evp-aes-256-ecb@1024        2.385991919                  1.396018192               70.91%
evp-aes-256-ecb@8192        2.147862636                  1.142222597               88.04%
evp-aes-256-ecb@16384      2.131361787                  1.135944617               87.63%

A57(freq: 2.0G):
Benchmark name                    Before optimization       After optimization      Improve
evp-aes-128-ecb@16             18.61045121                  18.36456218              1.34%
evp-aes-128-ecb@64             6.438628994                  5.467959461              17.75%
evp-aes-128-ecb@256           2.957452881                  1.97238604                 49.94%
evp-aes-128-ecb@1024         2.117096219                  1.099665054               92.52%
evp-aes-128-ecb@8192         1.868385973                  0.837440804               123.11%
evp-aes-128-ecb@16384       1.853078526                  0.822420027               125.32%
evp-aes-192-ecb@16             19.07021756                  18.50018552               3.08%
evp-aes-192-ecb@64             6.672351486                  5.696088921               17.14%
evp-aes-192-ecb@256           3.260427769                  2.131449916               52.97%
evp-aes-192-ecb@1024         2.410522832                  1.250529718               92.76%
evp-aes-192-ecb@8192         2.17921605                    0.973225504               123.92%
evp-aes-192-ecb@16384       2.162250997                  0.95919871                 125.42%
evp-aes-256-ecb@16            19.3008384                    19.12743654                0.91%
evp-aes-256-ecb@64             6.992950658                 5.92149541                  18.09%
evp-aes-256-ecb@256           3.576361743                 2.287619504                 56.34%
evp-aes-256-ecb@1024         2.726671027                 1.381267599                 97.40%
evp-aes-256-ecb@8192         2.493583657                 1.110959913                 124.45%
evp-aes-256-ecb@16384       2.473916816                 1.099967073                 124.91%

### DIFF
--- a/crypto/aes/asm/aesv8-armx.pl
+++ b/crypto/aes/asm/aesv8-armx.pl
@@ -384,6 +384,836 @@ ___
 &gen_block("en");
 &gen_block("de");
 }}}
+
+# Performance in cycles per byte.
+# Processed with AES-ECB different key size.
+# It shows the value before and after optimization as below:
+# (before/after):
+#
+#		AES-128-ECB		AES-192-ECB		AES-256-ECB
+# Cortex-A57	1.85/0.82		2.16/0.96		2.47/1.10
+# Cortex-A72	1.64/0.85		1.82/0.99		2.13/1.14
+
+# Optimization is implemented by loop unrolling and interleaving.
+# Commonly, we choose the unrolling factor as 5, if the input
+# data size smaller than 5 blocks, but not smaller than 3 blocks,
+# choose 3 as the unrolling factor.
+# If the input data size dsize >= 5*16 bytes, then take 5 blocks
+# as one iteration, every loop the left size lsize -= 5*16.
+# If 5*16 > lsize >= 3*16 bytes, take 3 blocks as one iteration,
+# every loop lsize -=3*16.
+# If lsize < 3*16 bytes, treat them as the tail, interleave the
+# two blocks AES instructions.
+# There is one special case, if the original input data size dsize
+# = 16 bytes, we will treat it seperately to improve the
+# performance: one independent code block without LR, FP load and
+# store, just looks like what the original ECB implementation does.
+
+{{{
+my ($inp,$out,$len,$key)=map("x$_",(0..3));
+my ($enc,$rounds,$cnt,$key_,$step)=("w4","w5","w6","x7","x8");
+my ($dat0,$dat1,$in0,$in1,$tmp0,$tmp1,$tmp2,$rndlast)=map("q$_",(0..7));
+
+my ($dat,$tmp,$rndzero_n_last)=($dat0,$tmp0,$tmp1);
+
+### q7	last round key
+### q10-q15	q7 Last 7 round keys
+### q8-q9	preloaded round keys except last 7 keys for big size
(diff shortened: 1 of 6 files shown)


kroeckx:
I have to wonder why we need to optimize ECB mode, what is the use case?

xkqian:
> I have to wonder why we need to optimize ECB mode, what is the use case?

We are monitoring the performance of the dedicated algorithms in openssl, and find the big gap between aarch and x86, seems the x86 has optimization there but aarch have no.  So there is the optimization for aarch64 and aarch32.   For the use case, I am not sure who will use it, but ECB is the backbone of AES. It's  useful in cases  where an Initialization vector is not provided. It is also useful when we want to do a random access of encrypted data.

levitte:
Merged.

2ff16afc17af7ee8dbff1fb34b71a51ecd55811e Optimize AES-ECB mode in OpenSSL for both aarch64 and aarch32

Issue #10522, opened 2019-11-26
'make update' drops KDF symbols after adding new error code


mspncp:
After adding a new error code, 'make update' drops a lot of KDF symbols and also one legacy include guard.

This issue was noticed in https://github.com/openssl/openssl/pull/10483#issuecomment-558395257.

### Apply the following patch to master

```patch
From 879b0fc2b2467d24050334e73c247e21424f5ab0 Mon Sep 17 00:00:00 2001
From: "Dr. Matthias St. Pierre" <Matthias.St.Pierre@ncp-e.com>
Date: Tue, 26 Nov 2019 13:01:43 +0100
Subject: [PATCH 1/2] rand_lib.c: Add some error message

---
 crypto/rand/rand_lib.c | 2 ++
 1 file changed, 2 insertions(+)

diff --git a/crypto/rand/rand_lib.c b/crypto/rand/rand_lib.c
index c6b4ad3ae4..8a657b499f 100644
--- a/crypto/rand/rand_lib.c
+++ b/crypto/rand/rand_lib.c
@@ -841,6 +841,8 @@ void RAND_add(const void *buf, int num, double randomness)
 
     if (meth->add != NULL)
         meth->add(buf, num, randomness);
+
+    RANDerr(RAND_F_RAND_ADD, RAND_R_SOME_ERROR_MESSAGE);
 }
 
 /*
-- 
2.21.0
```

### Reconfigure and run 'make update'

```
msp@msppc:~/openssl/openssl$ ./myconfigure
+ perl Configure debug-linux-x86_64 enable-crypto-mdebug --strict-warnings --prefix=/opt/openssl-dev
Configuring OpenSSL version 3.0.0-dev for target debug-linux-x86_64
Using os-specific seed configuration
Creating configdata.pm
Running configdata.pm
Creating Makefile

**********************************************************************
***                                                                ***
***   OpenSSL has been successfully configured                     ***
***                                                                ***
***   If you encounter a problem while building, please open an    ***
***   issue on GitHub <https://github.com/openssl/openssl/issues>  ***
***   and include the output from the following command:           ***
***                                                                ***
***       perl configdata.pm --dump                                ***
***                                                                ***
***   (If you are new to OpenSSL, you might want to consult the    ***
***   'Troubleshooting' section in the INSTALL file first)         ***
***                                                                ***
**********************************************************************
msp@msppc:~/openssl/openssl$ git diff
msp@msppc:~/openssl/openssl$ make update
[...]

PR #10526, opened 2019-11-26
Add some missing NULL pointer return checks

### LINKED ISSUE #10525: Missing null check after OPENSSL_zalloc
Here're places that null pointer check is missing after `OPENSSL_zalloc`. 

1. The pointer `exchange` is used directly in line 23 without being null checked.

https://github.com/openssl/openssl/blob/3c957bcd54d097167e53660fa100aa1ba85d63b5/crypto/evp/exchange.c#L21-L23

2. The pointer `signature` is also used directly in line 23. Similar to the first one

https://github.com/openssl/openssl/blob/3c957bcd54d097167e53660fa100aa1ba85d63b5/crypto/evp/pmeth_fn.c#L21-L23

3. In the function `OPENSSL_buf2hexstr`, the result of `OPENSSL_zalloc` is directly returned without null check. The dependents of `OPENSSL_buf2hexstr` are not checking null-pointers too.

https://github.com/openssl/openssl/blob/3c957bcd54d097167e53660fa100aa1ba85d63b5/crypto/o_str.c#L242

mattcaswell:
Check the return value after OPENSSL_zalloc() for some EVP types.

We  should also check the return value from OPENSSL_buf2hextr() which may also be NULL.

Fixes #10525.

### DIFF
--- a/apps/openssl.c
+++ b/apps/openssl.c
@@ -112,6 +112,8 @@ static size_t internal_trace_cb(const char *buf, size_t cnt,
 
         tid = CRYPTO_THREAD_get_current_id();
         hex = OPENSSL_buf2hexstr((const unsigned char *)&tid, sizeof(tid));
+        if (hex == NULL)
+            return 0;
         BIO_snprintf(buffer, sizeof(buffer), "TRACE[%s]:%s: ",
                      hex, OSSL_trace_get_category_name(category));
         OPENSSL_free(hex);
--- a/apps/openssl.c
+++ b/apps/openssl.c
@@ -112,10 +112,9 @@ static size_t internal_trace_cb(const char *buf, size_t cnt,
 
         tid = CRYPTO_THREAD_get_current_id();
         hex = OPENSSL_buf2hexstr((const unsigned char *)&tid, sizeof(tid));
-        if (hex == NULL)
-            return 0;
         BIO_snprintf(buffer, sizeof(buffer), "TRACE[%s]:%s: ",
-                     hex, OSSL_trace_get_category_name(category));
+                     hex == NULL ? "<null>" : hex,
+                     OSSL_trace_get_category_name(category));
         OPENSSL_free(hex);
         BIO_ctrl(trace_data->bio, PREFIX_CTRL_SET_PREFIX,
                  strlen(buffer), buffer);
--- a/crypto/err/err_prn.c
+++ b/crypto/err/err_prn.c
@@ -36,8 +36,8 @@ void ERR_print_errors_cb(int (*cb) (const char *str, size_t len, void *u),
             data = "";
         hex = OPENSSL_buf2hexstr((const unsigned char *)&tid, sizeof(tid));
         BIO_snprintf(buf, sizeof(buf), "%s:error:%s:%s:%s:%s:%d:%s\n",
-                     hex == NULL ? "" : hex, lib, func, reason, file, line,
-                     data);
+                     hex == NULL ? "<null>" : hex, lib, func, reason, file,
+                     line, data);
         OPENSSL_free(hex);
         if (cb(buf, strlen(buf), u) <= 0)
             break;              /* abort outputting the error report */
--- a/crypto/err/err_prn.c
+++ b/crypto/err/err_prn.c
@@ -36,7 +36,8 @@ void ERR_print_errors_cb(int (*cb) (const char *str, size_t len, void *u),
             data = "";
         hex = OPENSSL_buf2hexstr((const unsigned char *)&tid, sizeof(tid));
         BIO_snprintf(buf, sizeof(buf), "%s:error:%s:%s:%s:%s:%d:%s\n",
-                     hex, lib, func, reason, file, line, data);
+                     hex == NULL ? "" : hex, lib, func, reason, file, line,
+                     data);
         OPENSSL_free(hex);
         if (cb(buf, strlen(buf), u) <= 0)
             break;              /* abort outputting the error report */
--- a/crypto/evp/exchange.c
+++ b/crypto/evp/exchange.c
@@ -20,11 +20,14 @@ static EVP_KEYEXCH *evp_keyexch_new(OSSL_PROVIDER *prov)
 {
     EVP_KEYEXCH *exchange = OPENSSL_zalloc(sizeof(EVP_KEYEXCH));
 
-    if (exchange == NULL)
+    if (exchange == NULL) {
+        ERR_raise(ERR_LIB_EVP, ERR_R_MALLOC_FAILURE);
         return NULL;
+    }
 
     exchange->lock = CRYPTO_THREAD_lock_new();
     if (exchange->lock == NULL) {
+        ERR_raise(ERR_LIB_EVP, ERR_R_MALLOC_FAILURE);
         OPENSSL_free(exchange);
         return NULL;
     }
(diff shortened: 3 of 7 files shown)


richsalz:
Shouldn't the commits have a "thanks to" credit?

crypto/evp/exchange.c
```diff
@@ -20,6 +20,9 @@ static EVP_KEYEXCH *evp_keyexch_new(OSSL_PROVIDER *prov)
 {
     EVP_KEYEXCH *exchange = OPENSSL_zalloc(sizeof(EVP_KEYEXCH));
 
+    if (exchange == NULL)
+        return NULL;
```
mspncp: In [line 43/46](https://github.com/openssl/openssl/pull/10526/commits/4c0c53b009d6a3e7a44b43d8238698faf8ad674d#diff-7f75adab80125f8e46d5e500b71fff74R46), you raise an `ERR_R_MALLOC_FAILURE` when heap allocation fails. Shouldn't this be done consistently for all malloc failures? (several locations)

mspncp: _Note: if you add new error codes, take a look at issue #10522 before running `make update`._

mspncp:
> Shouldn't the commits have a "thanks to" credit?

I agree with @richsalz.

To give credits in a consistent way, I would suggest to use unified tags like `Reported-by:` or `Suggested-by:`, which are used for example in the  [Linux Kernel Sources](https://git.wiki.kernel.org/index.php/CommitMessageConventions).

I did this for example in commit 88d90400c07bd46840b53ecedce370ebe19d9905. 

Your opinion, @openssl/committers?

beldmit:
I think it is a very good practice but it requires a bit more attention that will be missed from time to time.

mspncp:
Well, I wouldn’t consider it a strict requirement, but a courtesy to the reporter for taking the time to analyze and report the issue.

crypto/evp/pmeth_fn.c
```diff
@@ -20,6 +20,9 @@ static EVP_SIGNATURE *evp_signature_new(OSSL_PROVIDER *prov)
 {
     EVP_SIGNATURE *signature = OPENSSL_zalloc(sizeof(EVP_SIGNATURE));
 
+    if (signature == NULL)
+        return NULL;
+
```
levitte: You could cascade this with the lock creation check and have them both set `ERR_R_MALLOC_FAILURE`

crypto/evp/pmeth_fn.c
```diff
@@ -760,6 +763,9 @@ static EVP_ASYM_CIPHER *evp_asym_cipher_new(OSSL_PROVIDER *prov)
 {
     EVP_ASYM_CIPHER *cipher = OPENSSL_zalloc(sizeof(EVP_ASYM_CIPHER));
 
+    if (cipher == NULL)
+        return NULL;
+
```
levitte: Same here

levitte:
We have been very sloppy in adding errors to the queue, and we really need to stop that.  This is as good a starting point as any.

mattcaswell:
Ok, I pushed some fixups to add more errors to the error queue, and added "Reported-by" fields to the commit messages.

mspncp:
> Ok, I pushed some fixups to add more errors to the error queue, and added "Reported-by" fields to the commit messages.

Thanks. I would however prefer to use the real name if it is visible in [the public profile](https://github.com/Liby99).

    Reported-by: Ziyang Li

because that's much more personal. If desired, the GitHub handle could be added optionally:

    Reported-by: Ziyang Li  (@Liby99)

mattcaswell:
> Thanks. I would however prefer to use the real name if it is visible in the public profile.

Ok. Done.

apps/openssl.c
```diff
@@ -112,6 +112,8 @@ static size_t internal_trace_cb(const char *buf, size_t cnt,
 
         tid = CRYPTO_THREAD_get_current_id();
         hex = OPENSSL_buf2hexstr((const unsigned char *)&tid, sizeof(tid));
+        if (hex == NULL)
+            return 0;
         BIO_snprintf(buffer, sizeof(buffer), "TRACE[%s]:%s: ",
                      hex, OSSL_trace_get_category_name(category));
         OPENSSL_free(hex);
```
mspncp: The only purpose for the hex value is to distinguish thread IDs in the trace output. So stopping trace output silently is confusing and hard to find. It would be better to send the output anyway, giving a visual indication of the failure:

```C
        BIO_snprintf(buffer, sizeof(buffer), "TRACE[%s]:%s: ",
                     hex != NULL ? hex : "<null ptr?!>", OSSL_trace_get_category_name(category));
```

crypto/err/err_prn.c
```diff
@@ -36,7 +36,8 @@ void ERR_print_errors_cb(int (*cb) (const char *str, size_t len, void *u),
             data = "";
         hex = OPENSSL_buf2hexstr((const unsigned char *)&tid, sizeof(tid));
         BIO_snprintf(buf, sizeof(buf), "%s:error:%s:%s:%s:%s:%d:%s\n",
-                     hex, lib, func, reason, file, line, data);
+                     hex == NULL ? "" : hex, lib, func, reason, file, line,
```
mspncp: Same here.

_(If you prefer to check `hex == NULL` instead of `hex != NULL` as I did above, that's ok for me too._)

crypto/mem_dbg.c
```diff
@@ -374,8 +374,8 @@ static void print_leak(const MEM *m, MEM_LEAK *l)
 
     hex = OPENSSL_buf2hexstr((const unsigned char *)&m->threadid,
                              sizeof(m->threadid));
-    n = BIO_snprintf(bufp, len, "thread=%s, number=%d, address=%p\n", hex,
-                     m->num, m->addr);
+    n = BIO_snprintf(bufp, len, "thread=%s, number=%d, address=%p\n",
+                     hex == NULL ? "" : hex, m->num, m->addr);
```
mspncp: Same here.

crypto/x509/v3_akey.c
```diff
@@ -42,13 +42,18 @@ static STACK_OF(CONF_VALUE) *i2v_AUTHORITY_KEYID(X509V3_EXT_METHOD *method,
     char *tmp;
     if (akeyid->keyid) {
         tmp = OPENSSL_buf2hexstr(akeyid->keyid->data, akeyid->keyid->length);
-        X509V3_add_value((akeyid->issuer || akeyid->serial) ? "keyid" : NULL, tmp, &extlist);
+        if (tmp == NULL)
+            return NULL;
```
mspncp: Add an error code?

crypto/x509/v3_akey.c
```diff
@@ -42,13 +42,18 @@ static STACK_OF(CONF_VALUE) *i2v_AUTHORITY_KEYID(X509V3_EXT_METHOD *method,
...
+        X509V3_add_value((akeyid->issuer || akeyid->serial) ? "keyid" : NULL,
+                         tmp, &extlist);
         OPENSSL_free(tmp);
     }
     if (akeyid->issuer)
         extlist = i2v_GENERAL_NAMES(NULL, akeyid->issuer, extlist);
     if (akeyid->serial) {
         tmp = OPENSSL_buf2hexstr(akeyid->serial->data, akeyid->serial->length);
+        if (tmp == NULL)
+            return NULL;
```
mspncp: Here too?

mattcaswell:
New fixup commit pushed addressing latest feedback.

mattcaswell:
Pushed. Thanks.

Issue #8675, opened 2019-04-04
CHACHA20_POLY1305 different results for chunked/non-chunked updating


guidovranken:
The following SHOULD produce the same output whether you compile with ```-DCHUNKED``` or not, but the last byte differs.

I deliberately pass NULL to ```EVP_DecryptUpdate``` if the chunk size is 0.
If you pass a non-NULL pointer, the results are the same.

```c
#include <openssl/evp.h>

#define CF_CHECK_EQ(expr, res) if ( (expr) != (res) ) { goto end; }
#define CF_CHECK_NE(expr, res) if ( (expr) == (res) ) { goto end; }

int main(void)
{
    const EVP_CIPHER* cipher = NULL;
    EVP_CIPHER_CTX* ctx = NULL;
    const unsigned char key[32] = {0x00, 0x00, 0x00, 0x00, 0x00, 0xf9, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x00, 0x00, 0xdc, 0x4d, 0xad, 0x6b, 0x06, 0x93, 0x4f, 0x29, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
    unsigned char iv[8] = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
    unsigned char ciphertext[100] = {
     0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
     0xd5, 0x00, 0x00, 0x00, 0x00, 0x05, 0x00, 0x00, 0x00, 0xd5, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
     0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
     0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
     0x00, 0x9c, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
[...]

PR #9057, opened 2019-05-31
 EVP_*Update: ensure that input NULL with length 0 isn't passed

mattcaswell:
Even with custom ciphers, the combination in == NULL && inl == 0 should not be passed down to the backend cipher function.  The reason is that these are the values passed by EVP_*Final, and some of the backend cipher functions do check for these to see if a "final" call is made.

An exception is made for CCM mode which has special handling for the case where inl == 0: this may mean the total plaintext or ciphertext length is 0.

This is based on an original commit by @levitte which was later reverted in 2c236894. It was reverted because it has the unintended consequence of breaking CCM. This new commit is tweaked to avoid the CCM failure. I've also added a new test to catch the issue that was originally described in #8675. Commit ecb0f148a9 also added additional CCM test vectors which would have caught the problem in the original commit.

Fixes #8675

PR #10530, opened 2019-11-27
 Make sure we handle input NULL with length

mattcaswell:
Even with custom ciphers, the combination in == NULL && inl == 0 should not be passed down to the backend cipher function. The reason is that these are the values passed by EVP_*Final, and some of the backend cipher functions do check for these to see if a "final" call is made.

An exception is made for CCM mode which has special handling for the case where inl == 0: this may mean the total plaintext or ciphertext length is 0.

This is a replacement for #9057 for the master branch only. It has the same test case but we leave it up to the provider implementations to make whatever checks they see fit. This keeps special case CCM code out of the EVP layer. #9057 still applies for 1.1.1 where it seems the best route ahead.

### DIFF
--- a/providers/common/ciphers/cipher_common.c
+++ b/providers/common/ciphers/cipher_common.c
@@ -291,6 +291,11 @@ int cipher_generic_stream_update(void *vctx, unsigned char *out, size_t *outl,
 {
     PROV_CIPHER_CTX *ctx = (PROV_CIPHER_CTX *)vctx;
 
+    if (inl == 0) {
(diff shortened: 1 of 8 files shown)


kroeckx:
This should probably mention it fixes #8675.

mattcaswell:
> This should probably mention it fixes #8675.

Oh yes - thanks for pointing that out. It does say that in the commit, but I forgot to write it in the PR description.

mattcaswell:
Pushed. Thanks.

PR #9759, opened 2019-09-03
[WIP, Pending on #10538]  Deprecate all EVP_MD_meth_ and EVP_CIPHER_meth_ functions in 3.0.0

levitte:
As a consequence, all engine functionality is deprecated as well,
since the `EVP_{TYPE}_meth_` functions form the basis for them.

PR #10538, opened 2019-11-29
Configure: No error on use of deprecated functions

levitte:
So far, we have let usage of deprecated functions become errors with
--strict-warnings.  This turns out to be too hard as we're looking at
deprecating functionality that we are still going to have to support
until they are removed, which means that usage of deprecated functions
must still be able to compile.

### DIFF
--- a/Configure
+++ b/Configure
@@ -145,7 +145,7 @@ my @gcc_devteam_warn = qw(
     -Werror
     -Wmissing-prototypes
     -Wstrict-prototypes
-    -Wno-error=deprecated-declarations
+    -Wno-deprecated-declarations
 );
 
 # These are used in addition to $gcc_devteam_warn when the compiler is clang.
--- a/Configure
+++ b/Configure
@@ -145,6 +145,7 @@ my @gcc_devteam_warn = qw(
     -Werror
     -Wmissing-prototypes
     -Wstrict-prototypes
+    -Wno-error=deprecated-declarations
 );
 
 # These are used in addition to $gcc_devteam_warn when the compiler is clang.


levitte:
Other PRs, such as #9759, will have to depend on this.

mattcaswell:
Is there an easy way for someone to add this back in? Sometimes this *is* the behaviour we want.

Which leads me to another thought...wouldn't it be nice if we could declare something as deprecated in two different ways, i.e. deprecated for application use only, vs deprecated for application *and* internal use.

levitte:
> Is there an easy way for someone to add this back in? Sometimes this is the behaviour we want.

Yes, looks like this:

```
$ ./config --strict-warnings -Werror=deprecated-declarations
```

The way `--strict-warnings` works is that it inserts the appropriate set of warning options in the user provided flags, so whatever comes after is added to the set of use provided flags.

> Which leads me to another thought...wouldn't it be nice if we could declare something as deprecated in two different ways, i.e. deprecated for application use only, vs deprecated for application and internal use.

How much do you want to over engineer configuration?  I do have some work which adds the possibility to give selective choice like you express, but that does add quite a bit of complexity, and I wonder if it's worth it...

mattcaswell:
> How much do you want to over engineer configuration?

This isn't about *configuration* but about *declaration*. I.e. where we declare something as deprecated we declare it to be either "ok for internal use", or "not ok for internal use"

levitte:
That's another way we can over engineer things...  I seriously think that's making things overly complex.  Do you *really* want another set of macros or whatever with almost but not quite the same meaning as the current ones?

With this change, we still get the warnings all over the place, it's not like we're not going to be informed by the compilers that we're using deprecated stuff :wink:

mattcaswell:
> With this change, we still get the warnings all over the place, it's not like we're not going to be informed by the compilers that we're using deprecated stuff

But actually that is really a bad thing. The whole point of doing this is that we want to be able to use deprecated stuff internally (for good reasons). That means that very quickly we're going to see lots of warnings in our builds. As soon as that happens we're immediately going to filter out and ignore the warnings that flash past as we do a build and just assume they're all ok. But we won't see the ones that *do* matter.

I actually think its a really good thing that we have warning free builds at the moment and I am loathe to lose that.

levitte:
Ok, so that contradicts what you said in #10367, and I quote:

> Hmm, good question. I think I tend towards wanting to see the warning.

Ref: https://github.com/openssl/openssl/pull/10367#issuecomment-552363734

I have *no* issues whatsoever turning of that particular warning entirely as part of `--strict-warnings`.  Please make up your mind.

mattcaswell:
> Ok, that contradicts what you said in #10367, 

Hey - I never claimed to be consistent! I changed my mind. Or rather, to be more accurate, my thinking has moved on, and what I am proposing now is a middle ground, i.e. "no warnings at all even for stuff which should really be an error" isn't desirable, and neither is "warnings for lots of stuff, some of which might be an error and most of which isn't". I prefer an option which gives us errors when we use stuff internally when we shouldn't, and doesn't complain when we use stuff that we've decided is ok.

levitte:
I'm not quite sure where that leaves this PR.

Do note, btw, that we still have, and test, the configuration option `no-deprecated`, which simulates a removal of everything that's deprecated (up to and including the version given with `--api`).  That might be the answer to your desire for errors when we're using things we shouldn't, right?

t8m:
We could add a macro using _Pragma() to switch off the warning selectively at least with compilers that support it.

t8m:
But yeah, probably the no-deprecated build should be sufficient to test for this kind of errors.

levitte:
So, on where that leaves this PR, it sounds like I should change `-Wno-error=deprecated-declarations` to `-Wno-deprecated-declarations`.  Do note that in builds with no configuration options, we'll still get those warnings.  I don't think that's much of an issue, though (and make -s silences them anyway)

mattcaswell:
Won't "no-deprecated" actually break if we're using things internally that are deprecated but we decided are ok for internal use?

t-j-h:
I think @levitte has a convincing view point here - no-deprecated builds to the rescue and we keep simplicity. And I also think we shouldn't be using things we deprecate - we should take the same approach as our users are being forced to take.

levitte:
>we should take the same approach as our users are being forced to take.

A *warning* doesn't force the user to do anything, at least for the moment.

> And I also think we shouldn't be using things we deprecate

Yeeeeeah...  so imagine deprecating the ENGINE API.  The whole lot.  Consider how invasive that becomes if we follow that train of thought, and what an impossible situation we're putting ourselves in, not to mention our users.  Suddenly, deprecation actually becomes removal, which defeats the purpose of using deprecation as a warning of future removal.

t8m:
The point is we should be able to internally use the deprecated stuff to implement other deprecated stuff. Not otherwise.

levitte:
> The point is we should be able to internally use the deprecated stuff to implement other deprecated stuff. Not otherwise.

Exactly.  The way things currently are, `--strict-warnings` prevents us from using deprecated stuff *at all*, so it's practically the same as configuring with `no-deprecated`.

mattcaswell:
> we should take the same approach as our users are being forced to take.

I don't see who made this comment that @levitte was replying to. However, this is the approach that we have taken to date. Unfortunately I don't think this is going to be viable any more. I expect us to deprecate lots of low-level APIs. For example I think we might deprecate:

````
int RSA_public_encrypt(int flen, const unsigned char *from,
int RSA_private_encrypt(int flen, const unsigned char *from,
int RSA_public_decrypt(int flen, const unsigned char *from,
int RSA_private_decrypt(int flen, const unsigned char *from,
````

We should be pushing users to use the EVP interface instead of the low level one, so deprecating would make sense. *But* we need to still use those functions internally in order to *implement* the EVP version. We *could* make the public functions simple wrappers around some other internal function, and use the internal function where we need to....but that's fairly horrible.

levitte:
@mattcaswell [said](https://github.com/openssl/openssl/pull/10538#issuecomment-559729062)
> I don't see who made this comment that @levitte was replying to. 

That was @t-j-h, here: https://github.com/openssl/openssl/pull/10538#issuecomment-559725111
(thanks for the hint, I'll try to remember to add backlinks, like I do here)

> but that's *fairly* horrible.

I'd call that understating it :wink:

levitte:
I've changed this PR so `--strict-warnings` now includes `-Wno-deprecated-declarations`.  "Normal" builds will still warn, and `no-deprecated` will still simulate removal of deprecated stuff.

From all that I've heard so far, that seems like the best compromise without over complicating stuff further.

mattcaswell:
> I've changed this PR so --strict-warnings now includes -Wno-deprecated-declarations. "Normal" builds will still warn, and no-deprecated will still simulate removal of deprecated stuff.

I fail to see how this helps at all. Won't no-deprecated just break entirely as soon as we use something internally that has been deprecated for external users?

levitte:
Matt [said](https://github.com/openssl/openssl/pull/10538#issuecomment-559734673)
> I fail to see how this helps at all. Won't no-deprecated just break entirely as soon as we use something internally that has been deprecated for external users?

What I'm trying to say is that `no-deprecated` is not changed by this PR, and that we can use it to mark our uses of deprecated stuff with guards checking the macros `OPENSSL_NO_DEPRECATED_x_y[_z]`
So basically, consider that part a side note.

t8m:
But Matt has a good point that you will still need RSA_private_encrypt() and other similar stuff internally in the provider implementations of RSA operations even with no-deprecated.

mattcaswell:
> But Matt has a good point that you will still need RSA_private_encrypt() and other similar stuff internally in the provider implementations of RSA operations even with no-deprecated.

Exactly! no-deprecated will just break in this scenario and there's no way around it. We will still need to use this deprecated stuff even in a no-deprecated build!

t8m:
Could it be solved by just re-declaring these in an internal header without the deprecation?

mattcaswell:
> Could it be solved by just re-declaring these in an internal header without the deprecation?

That's a possible solution. It would have to mean that for internal builds only DEPRECATEDIN would have to work like `no-deprecated` so that we never pick up the declarations from the public header files in that case, i.e. we only ever get them from the internal header file. It does also mean declaring things twice.

levitte:
Ah, good point.  But that's kind of a different situation, as we're now talking about functionality that we want to *retain internally*, just stop *exposing publically* at some point.  So it actually makes sense to re-declare them as internal functions and provide deprecated public wrappers.

levitte:
I'd like to see focus back on this PR, though.  It should be simple enough, really.

mattcaswell:
> So it actually makes sense to re-declare them as internal functions and provide deprecated public wrappers.

That was the scenario I described as "fairly horrible" and you said was understating it!

mattcaswell:
> I'd like to see focus back on this PR, though. It should be simple enough, really.

If we solve the problem I described above, I fail to see why we need this PR at all.

levitte:
Matt [said](https://github.com/openssl/openssl/pull/10538#issuecomment-559742599):
> If we solve the problem I described above, I fail to see why we need this PR at all.

Again, consider deprecation of the ENGINE API, the whole lot.  This is something that we *don't* want to preserve internally, but still need to support until it's actually removed.  This is where the understatement of "fairly horrible" comes in, as making the split into internal functions with public wrappers is *unnecessary*.

Also, consider #10367, where we deprecate the `AES_ige_` functions, but retain all the tests until actual removal.

This PR solves those problems.

mattcaswell:
> Again, consider deprecation of the ENGINE API, the whole lot. This is something that we don't want to preserve internally, but still need to support until it's actually removed. This is where the understatement of "fairly horrible" comes in, as making the split into internal functions with public wrappers is unnecessary.

Isn't this *exactly* a case where an API is deprecated externally, but we still need to use it internally? If someone passes an ENGINE reference to `EVP_EncryptInit_ex` then, no matter that it is deprecated, we still need to do something with it.

> Also, consider #10367, where we deprecate the AES_ige_ functions, but retain all the tests until actual removal.

Same problem...we're using something internally that is deprecated externally.

levitte:
Matt [said](https://github.com/openssl/openssl/pull/10538#issuecomment-559748068):
> Isn't this *exactly* a case where an API is deprecated externally, but we still need to use it internally? If someone passes an ENGINE reference to `EVP_EncryptInit_ex` then, no matter that it is deprecated, we still need to do something with it.

Nope, it's not exactly the same.  The ENGINE code will be removed at some point, i.e. not exist internally at all.  I totally plan on deprecating `EVP_EncryptInit_ex` alongside that, for exactly the reason that it takes an ENGINE pointer.

The RSA encrypt and decrypt *code* will still be retained...  actually, I imagine that as soon as the public interface is removed, that code will move to `providers/`, and possibly be refactored.  The same goes for all other low level interfaces (ciphers, digests, what have you) we choose to deprecate.

So I see a difference between *code that we intend remove* and *code that we intend to retain internally*, long term.

levitte:
Matt [also said](https://github.com/openssl/openssl/pull/10538#issuecomment-559748068):

> > Also, consider #10367, where we deprecate the AES_ige_ functions, but retain all the tests until actual removal.
> 
> Same problem...we're using something internally that is deprecated externally.

I think it's *wrong* to use internal libcrypto functions in the `openssl` app.  No.

mattcaswell:
> Nope, it's not exactly the same. The ENGINE code will be removed at some point, i.e. not exist internally at all. I totally plan on deprecating EVP_EncryptInit_ex alongside that, for exactly the reason that it takes an ENGINE pointer.

But the same solution can be used for both the RSA_public_encrypt problem and the ENGINE problem, e.g. @t8m's solution would work perfectly well for both situations. Given that we need the solution anyway for RSA_public_encrypt, why wouldn't we do it the same way for ENGINE?

> I think it's wrong to use internal libcrypto functions in the openssl app. No.

Errr... we're not talking about the openssl app? We're talking about *tests*?

levitte:
Matt [said](https://github.com/openssl/openssl/pull/10538#issuecomment-559752645):
> Errr... we're not talking about the openssl app? We're talking about *tests*?

#10367 deprecates `AES_ige_` functions, which are used in `openssl speed`.  That PR did remove all those uses, but that got a pushback.  So for the moment, the build of that PR fails because we use it there.

That's what started all this.

levitte:
> But the same solution can be used for both the RSA_public_encrypt problem and the ENGINE problem, e.g. @t8m's solution would work perfectly well for both situations. Given that we need the solution anyway for RSA_public_encrypt, why wouldn't we do it the same way for ENGINE?

You know what?  I think I'll leave it to you to do that job, just so you understand the size and ramifications of that task.

levitte:
> But the same solution can be used for both the RSA_public_encrypt problem and the ENGINE problem, e.g. @t8m's solution would work perfectly well for both situations. Given that we need the solution anyway for RSA_public_encrypt, why wouldn't we do it the same way for ENGINE?

For code that we plan to remove, I think that's a waste of time, and makes things harder later on.

levitte:
@t8m [said](https://github.com/openssl/openssl/pull/10538#issuecomment-559737565):
> Could it be solved by just re-declaring these in an internal header without the deprecation?

How do you see that happening without duplication and the maintenance burden that incurs?  And how would that interact with `no-deprecated`?

levitte:
So here's something that's a mystery to me;  The way `--strict-warnings` currently works, it practically becomes the same as `no-deprecated`, because it will produce an error whenever we happen to use something deprecated internally.  Why is that valuable?  Why is it such a bad thing to make `--strict-warnings` a bit more lenient regarding deprecated stuff, to the point of inventing all kinds of fairly complicated / burdensome schemes to deal with them?

I do not understand this.

mattcaswell:
> it practically becomes the same as no-deprecated,

No "no-deprecated" option does not even compile the deprecated code. Applications are *unable* to use deprecated functions because they do not exist in the library.

--strict-warnings still compiles the deprecated code into the library. Applications can use those functions if they want to. Its just that *we* can't use them.

levitte:
> --strict-warnings still compiles the deprecated code into the library. Applications can use those functions if they want to. Its just that *we* can't use them.

And why is that valuable to us, especially since `no-deprecated` *also* puts us in a situation where *we* can't use them?

levitte:
*Another* option could be to have `--strict-warnings` also mean `no-deprecated`.  Considering that's what it means to *us* currently anyway...

levitte:
I will remind you that `--strict-warnings` was primarily invented for *our* (i.e. OpenSSL developers) use.  Others are free to use it, of course, but...

mattcaswell:
Alternative approach shown in #10540

mattcaswell:
> And why is that valuable to us, especially since no-deprecated also puts us in a situation where we can't use them?

I think it is desirable for us to not use deprecated APIs for our own use *except* where there are scenarios where there is no alternative, e.g.

1) We are testing the deprecated the interface
2) The apps are using a deprecated interface which has no replacement and we don't want to actually remove the feature (yet)
3) We have deprecated a lower level interface in order to move applications to use the higher level one - but we still need to implement the high level interface in terms of the low level one
4) Where applications have called deprecated interfaces anyway we may need to call other deprecated interfaces in order to implement them

t8m:
But still successful no-deprecated build would ensure that we are not using deprecated APIs in our own except for the 1-4 scenarios you mention. All these scenarios would be switched off when compiling with no-deprecated.

t8m:
> @t8m [said](https://github.com/openssl/openssl/pull/10538#issuecomment-559737565):
> 
> > Could it be solved by just re-declaring these in an internal header without the deprecation?
> 
> How do you see that happening without duplication and the maintenance burden that incurs? And how would that interact with `no-deprecated`?

I actually think using internal functions plus public deprecated wrapper functions is cleaner approach. However it still brings duplication and maintenance burden.

paulidale:
I'm leaning towards renaming deprecated functions which we need internally and providing wrappers with the original names.  I expect a pile of user queries/complaints about why function `foo()` is in the code but their program that uses it doesn't compile anymore.

For things we're going to completely remove, there is no need to rename.  These will need to be handled differently.

I'm not keen on the `OPENSSL_INTERNAL_USE` macro infecting the public headers.

levitte:
> For things we're going to completely remove, there is no need to rename. These will need to be handled differently.

Yup, the problem from #10367 remains. And, considering all the parts of the openssl commands that use *public functions*, with some of them being legacy (`openssl rsautil`, for example), we will have the same problem there as long as those exist in their current form.

paulidale:
I think the answer is that the command line apps shouldn't use deprecated APIs.  I'm aware of at least some of the sticky problems this could cause (e.g. converting ctrl calls to param ones is nasty).  I guess we could delay any update but then we're back to the warnings.

levitte:
@paulidale, please read the discussion in #10367, have a think and come back.  I did remove all uses of `AES_ige` functions, *exactly* for the reason you state, and it got some pushback.

Also, as soon as we deprecate the ENGINE API, the impact will be *huge*, the `openssl engine` command will have to go away, yada yada yada.

richsalz:
I think you're heading towards having a separate flags/warnings for the apps (and also the tests).  I am sure that this can be done, but do you want to do it in this release?

The build.info files have already evolved into a very obscure domain-specific language.  I fear that https://softwareengineering.stackexchange.com/questions/150254/what-does-jamie-zawinskis-law-mean can't be far away :)

Saying "ENGINE is deprecated, move to providers" might be all that's worth doing at this point.  Even add it to the engine command.

levitte:
It's possible to move to having app-specific flags, and that will have absolutely nothing to do with build.info flags and everything to do with make variables

paulidale:
@levitte, would we be okay with a single CI build specifically for legacy/deprecated?  The rest keep the strict warnings etc.  That would provide test coverage for the old stuff -- I can't think of any device specific stuff we're planning on deprecating (I'm probably wrong).

levitte:
Not sure I understand the implications of your question. Wanna elaborate a bit?

paulidale:
I'm wonder if we can accept some/most of the CI builds not testing deprecated functions.
Would one build doing so be sufficient?

It means conditional code in the test suite and lots of updating of tests and apps.

richsalz:
Perhaps simpler, just do build-only and don't run the tests when no-deprecated is on internally.

levitte:
Guys, this is a *build* problem.

mattcaswell:
> All these scenarios would be switched off when compiling with no-deprecated.

This is just not possible. The RSA examples I gave *must* still be available for internal use even in a no-deprecated build. The only difference is they should not be exported to applications.

t8m:
I'd like to summarize what we got so far:
We cannot use -Wdeprecated-declarations in strict warnings because we have to be able to use the deprecated APIs for:
1. testing the deprecated the interface
2. the apps that are using a deprecated interface which has no replacement and we don't want to actually remove the feature (yet) (engines f.e.)
3. to implement the high level interface in terms of the low level one which we want to deprecate for public use
4. to implement some deprecated interface by calling other deprecated interfaces

On the other hand we would like to be able to check that we are not calling deprecated interfaces in any other cases.

IMO all this can be accomplished by having no-deprecated build that would of course have to disable also all the deprecated apps commands (or options) that use the deprecated APIs. We could add the -Werror=deprecated-declarations to the no-deprecated build to ensure that there is nothing deprecated left.

3. would have to be solved by having the deprecated public API call implemented as a wrapper to the internal low level call and internally we would have to use the non-public API call.

IMO this is the least invasive approach that would not require spending any serious time on fixing/re-implementing the deprecated functionality.

levitte:
Closing this in favor of #10558, which basically does the same thing, but in source.

levitte:
Oh, didn't see the latest discussion...

levitte:
> We could add the -Werror=deprecated-declarations to the no-deprecated build to ensure that there is nothing deprecated left.

`no-deprecated` *hides* all declarations of deprecated stuff, so that warning does nothing in that case.

t8m:
> > We could add the -Werror=deprecated-declarations to the no-deprecated build to ensure that there is nothing deprecated left.
> 
> `no-deprecated` _hides_ all declarations of deprecated stuff, so that warning does nothing in that case.
I understand that, I meant it as a safe-guard.

PR #10539, opened 2019-11-29
Update bio_ssl.c

suishixingkong:
fix a bug

##### Checklist

### DIFF
--- a/ssl/bio_ssl.c
+++ b/ssl/bio_ssl.c
@@ -284,6 +284,7 @@ static long ssl_ctrl(BIO *b, int cmd, long num, void *ptr)
             ssl_free(b);
             if (!ssl_new(b))
                 return 0;
+            bs = BIO_get_data(b);
         }
         BIO_set_shutdown(b, num);
         ssl = (SSL *)ptr;


t8m:
Also could you please amend the commit message to be more descriptive? I.e. what bug is fixed and how.

kaduk:
@suishixingkong will you be able to mark this commit as trivial from a copyright perspective, or submit a CLA?  If not we will have to convert this to an issue and produce an independent reimplementation.

ssl/bio_ssl.c
```diff
@@ -284,6 +284,7 @@ static long ssl_ctrl(BIO *b, int cmd, long num, void *ptr)
             ssl_free(b);
             if (!ssl_new(b))
                 return 0;
+            bs = BIO_get_data(b);
```
mspncp, line 287: @mattcaswell it seems like this bug was introduced way back in 2016 when the BIO was made opaque in commit a146ae55ba479a5c7aa2a6afba1b2b93102a152c and line 291 below [was changed as follows](https://github.com/openssl/openssl/commit/a146ae55ba479a5c7aa2a6afba1b2b93102a152c#diff-aa46282ee9d01cf3c7f3d1f9a4c3bfafR340):

```diff
@@ -330,17 +335,17 @@ static long ssl_ctrl(BIO *b, int cmd, long num, void *ptr)
...
-        ((BIO_SSL *)b->ptr)->ssl = ssl;
+        bs->ssl = ssl;
```

It's hard to believe that this bug went unnoticed for such a long time (or is this ctrl used only rarely?). This error should be fixed for master and 1.1.1 before the next release IMHO.

t8m: I suppose it is usually (practically always given there were no bug reports before) called only once on a SSL BIO so this code does not come into play.

t8m:
Given the non-responsibility of the reporter and triviality of the patch, should we just create a separate PR and just mention the reporter in the commit message so we give him credits for finding the issue? Or do you think we can merge and add the CLA: trivial ourselves without the author's explicit consent?

mattcaswell:
> Given the non-responsibility of the reporter and triviality of the patch, should we just create a separate PR and just mention the reporter in the commit message so we give him credits for finding the issue? Or do you think we can merge and add the CLA: trivial ourselves without the author's explicit consent?

I think someone could recreate it and given credit for finding it.

mspncp:
Since I already wrote a commit message, I might as well just create a pr.

mspncp:
See #11746.

mspncp:
Closing this pull request in favour of #11746, since the pr owner did not respond.

PR #10540, opened 2019-11-29
Deprecate the AES_ige_*() functions

mattcaswell:
These functions were already partially deprecated. Now we do it fully.

This is an alternative approach to that shown in #10538 and #10367.

### DIFF
--- a/include/openssl/macros.h
+++ b/include/openssl/macros.h
@@ -203,6 +203,16 @@
 # endif
 
 /*
+ * Some deprecated functions are still needed for use internally within OpenSSL.
+ * Applications should *not* define the OPENSSL_INTERNAL_USE macro.
+ */
+# ifdef OPENSSL_INTERNAL_USE
+#  define DEPRECATEDIN_3_0_EXTERN(f)       f;
+# else
+#  define DEPRECATEDIN_3_0_EXTERN(f)       DEPRECATEDIN_3_0(f)
+# endif
+
+/*
  * Make our own variants of __FILE__ and __LINE__, depending on configuration
  */
 
--- a/apps/speed.c
+++ b/apps/speed.c
@@ -17,6 +17,9 @@
 #define EdDSA_SECONDS   10
 #define SM2_SECONDS     10
 
+/* We need to use some deprecated APIs */
+#define OPENSSL_INTERNAL_USE
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
--- a/include/openssl/aes.h
+++ b/include/openssl/aes.h
@@ -73,17 +73,16 @@ void AES_cfb8_encrypt(const unsigned char *in, unsigned char *out,
 void AES_ofb128_encrypt(const unsigned char *in, unsigned char *out,
                         size_t length, const AES_KEY *key,
                         unsigned char *ivec, int *num);
-# ifndef OPENSSL_NO_DEPRECATED_3_0
+
 /* NB: the IV is _two_ blocks long */
-void AES_ige_encrypt(const unsigned char *in, unsigned char *out,
-                     size_t length, const AES_KEY *key,
-                     unsigned char *ivec, const int enc);
+DEPRECATEDIN_3_0_EXTERN(void AES_ige_encrypt(const unsigned char *in,
+    unsigned char *out, size_t length, const AES_KEY *key, unsigned char *ivec,
+    const int enc))
 /* NB: the IV is _four_ blocks long */
-void AES_bi_ige_encrypt(const unsigned char *in, unsigned char *out,
-                        size_t length, const AES_KEY *key,
-                        const AES_KEY *key2, const unsigned char *ivec,
-                        const int enc);
-# endif
+DEPRECATEDIN_3_0_EXTERN(void AES_bi_ige_encrypt(
+    const unsigned char *in, unsigned char *out, size_t length,
+    const AES_KEY *key, const AES_KEY *key2, const unsigned char *ivec,
+    const int enc))
 
 int AES_wrap_key(AES_KEY *key, const unsigned char *iv,
                  unsigned char *out,
--- a/test/igetest.c
+++ b/test/igetest.c
@@ -7,6 +7,9 @@
  * https://www.openssl.org/source/license.html
  */
 
+/* The AES_ige_* functions are deprecated, so we suppress warnings about them */
+#define OPENSSL_INTERNAL_USE
+
 #include <openssl/crypto.h>
 #include <openssl/aes.h>
 #include <openssl/rand.h>
--- a/util/perl/OpenSSL/ParseC.pm
+++ b/util/perl/OpenSSL/ParseC.pm
@@ -264,7 +264,7 @@ my @opensslchandlers = (
     # We trick the parser by pretending that the declaration is wrapped in a
     # check if the DEPRECATEDIN macro is defined or not.  Callers of parse()
     # will have to decide what to do with it.
-    { regexp   => qr/(DEPRECATEDIN_\d+_\d+(?:_\d+)?)<<<\((.*)\)>>>/,
+    { regexp   => qr/(DEPRECATEDIN_\d+_\d+(?:_\d+)?)(?:_EXTERN)?<<<\((.*)\)>>>/,
       massager => sub { return (<<"EOF");
 #ifndef $1
 $2;


levitte:
Try a `make update`. You will probably have to do a small bit of tuning in util/perl/OpenSSL/ParseC.pm

levitte:
I'm worried a bit about corner cases... but am currently short on time...

levitte:
Also... Visavi the libraries, the openssl app *is* an external application. This is intentional, to give us a hint what we're pushing on our users. Making it using something "internal" defeats that purpose.

mattcaswell:
> Also... Visavi the libraries, the openssl app is an external application. This is intentional, to give us a hint what we're pushing on our users. Making it using something "internal" defeats that purpose.

Right - but ultimately either we want the app to use deprecated interfaces or we don't. The APIs themselves *are* external. The only thing "internal" about it is the ability to suppress deprecation warnings. I see no problem with that.

mattcaswell:
> Try a make update. You will probably have to do a small bit of tuning in util/perl/OpenSSL/ParseC.pm

Fixed. Thanks for the hint.

levitte:
Ok, having ruminated, I'm wondering about the intended use of `DEPRECATEDIN_3_0_EXTERN`

See, I'm leaning back on what you said in https://github.com/openssl/openssl/pull/10538#issuecomment-559729062, where you mentioned these functions as examples of things that we want to deprecate (and eventually remove) the public interface for, but retain "internally", while there are other functions that are intended to be removed, code and all.

Example of functions to retain "internally" (from the comment I linked to):

``` C
int RSA_public_encrypt(int flen, const unsigned char *from,
int RSA_private_encrypt(int flen, const unsigned char *from,
int RSA_public_decrypt(int flen, const unsigned char *from,
int RSA_private_decrypt(int flen, const unsigned char *from,
```

Examples of functions that are *not* intended to be retained "internally":

``` C
void AES_ige_encrypt(const unsigned char *in,
                     unsigned char *out, size_t length,
                     const AES_KEY *key, unsigned char *ivec,
                     const int enc);
void AES_bi_ige_encrypt(const unsigned char *in,
                     unsigned char *out, size_t length,
                     const AES_KEY *key,
                     const AES_KEY *key2,
                     const unsigned char *ivec,
                     const int enc);
```

Now, the way I understand `DEPRECATEDIN_3_0_EXTERN`, it's that it's supposed to be used for code that we plan to retain "internally" after the public interface has been removed.  So looking at the examples above, I would totally understand if the mentioned RSA functions were deprecated using `DEPRECATEDIN_3_0_EXTERN`, but then it doesn't make sense at all to use it with the `AES_ige` functions, unless you intend to use that code after the public interface has been removed (in that case, why???).

Part of the reason I think we don't understand each other is because we're interpretting "internal" differently.  *My* understanding of that term in this context is that it denominates symbols that are present within libcrypto or built from libcrypto source (the FIPS module), but not outside of libcrypto (i.e. is not exported by libcrypto).

include/openssl/macros.h
```diff
@@ -202,6 +202,16 @@
 #  define DEPRECATEDIN_0_9_8(f)      f;
 # endif
 
+/*
+ * Some deprecated functions are still needed for use internally within OpenSSL.
+ * Applications should *not* define the OPENSSL_INTERNAL_USE macro.
+ */
```
levitte: And yet, we're using it in our app...  That usage kinda flies in the face of this comment...

levitte:
Regarding our `openssl` app, I think it should be treated as if it was a project separate from the libraries.  The source should be made in such a way that it could actually move to its own repo.  I know that it isn't quite like that for the moment, but I think that would be a worthy goal, and that a move to make it *less* so is a move in the wrong direction.

Why?  Because that's what we ask of our users, and we should have something that forces us to eat our own dog food, so to say.  Also, users have historically looked there to figure out how to do certain things, especially things that are badly documented, if at all.

mattcaswell:
> Part of the reason I think we don't understand each other is because we're interpretting "internal" differently. My understanding of that term in this context is that it denominates symbols that are present within libcrypto or built from libcrypto source (the FIPS module), but not outside of libcrypto (i.e. is not exported by libcrypto).

Err, no. These deprecated symbols must continue to be exported by libcrypto and applications may continue to use them. This is just a warning that we intend to remove them in some later version. The only thing that makes them special for "internal" use is that we do not expect to see the deprecated warning in that case. This is because we are deliberately using the symbol in spite of the deprecation and we know that it is "ok" to do so. 

The openssl app is exactly the same as any other app in this regards. With this change it is still only using the same symbols that any other application can use - and I fully expect that other applications will continue to use deprecated symbols for some while. Each of those applications must decide how they will handle the deprecated warnings that will arise as a result of an upgrade to 3.0. They could do so by ignoring them, or they could do so by suppressing deprecated warnings when compiling files that use those deprecated interfaces. This is exactly the same thing that we are doing here - suppressing warnings for those particular files that use them. The only difference is that we, as maintainers of both the library and the app, have an extra mechanism at our disposal to suppress those warnings. Actually there is nothing to stop other applications using the same mechanism as we do in this PR to suppress those warnings - other than a strongly worded comment discouraging them to do so. I could tone down the discouragement in the comment, or remove it altogether, if it makes you feel happier about it - but personally I see no particular need to do so.

mattcaswell:
> Regarding our openssl app, I think it should be treated as if it was a project separate from the libraries. The source should be made in such a way that it could actually move to its own repo. I know that it isn't quite like that for the moment, but I think that would be a worthy goal, and that a move to make it less so is a move in the wrong direction.

There is nothing in this PR that prevents the openssl app from being moved to a separate repo if we chose to do so.

levitte:
@mattcaswell, wanna say *again* what `DEPRECATED_3_0_EXTERN` is meant to be used for?

levitte:
Let me ask this another way...  so far as I understand, `DEPRECATED_3_0_EXTERN` is meant for functions that we want to retain the use of internally (i.e. inside libcrypto / providers), even after its symbol isn't exported any more (i.e. it's removed from public interface).

So I wonder, where do we use or want to ue `AES_ige_encrypt` _internally_?

mattcaswell:
> is meant for functions that we want to retain the use of internally (i.e. inside libcrypto / providers), even after its symbol isn't exported any more (i.e. it's removed from public interface).

This is a primary motivation. But I see no reason to restrict its use to only that scenario. I have a broader definition of the term internal than you do (in this context). I am using it in the sense of internal to the project. So, in that sense, our use of `AES_ige_encrypt` *is* internal, by definition. Even when only used in the app or tests.

I do agree that our app should not use any APIs that are not available to any other application. But this doesn't do that. When we mark something as deprecated we are not preventing an application from using that API. We're just giving a hint to the compiler that the application maintainers might want to consider changing their code. Those application maintainers will then need to look at those deprecation warnings and make a policy decision as to whether to 1) ignore them, or 2) suppress the warnings or 3) change their code.

We are the maintainers of the openssl and test apps and therefore we know that (2) is the right answer in this case.

We *could* change this PR to make this a generic switch that any application maintainer could use. E.g. forget about the distinction between `DEPRECATED_3_0` and `DEPRECATED_3_0_EXTERN` - we just have `DEPRECATED_3_0` with a define you can set up front to suppress the deprecation - and make it available to all application maintainers. If that makes everyone happy with this PR then I could accept that change - although I don't personally see the need to make it.

levitte:
That's basically what this boils down to, that *everything* that we mark deprecated will simply end up having its external interface removed at some point, but may be retained internally.

That's quite a shift of how we have done things so far.

Regarding the openssl app, you really have to make up your mind.  If it's assumed to be internal, it can't be treated as an external project (since when do we invite external application to use our internal stuff?  things we don't export?).  If it is to be treated as an external project, we cannot allow it to use internal stuff.  That's black and white, I do not see a gray area between those views.

mattcaswell:
> Regarding the openssl app, you really have to make up your mind. If it's assumed to be internal, it can't be treated as an external project (since when do we invite external application to use our internal stuff? things we don't export?). If it is to be treated as an external project, we cannot allow it to use internal stuff. That's black and white, I do not see a gray area between those views.

What in this PR means that the openssl app is using stuff we don't export??? The openssl app is only using public, supported functions before this PR and after it. If we chose to separate it out into a different repo - it would still be fine. Nothing in this PR changes that. The *only* thing "internal" that the app uses as a result of this PR is the ability to suppress deprecation warnings.

Since that seems to be controversial (for reasons I fail to understand) I've created a version in #10558 that doesn't have the distinction between internal/external.

levitte:
> What in this PR means that the openssl app is using stuff we don't export???

It's still the same confusion about what we mean with "internal"...  I still have in my brain that "internal" means something not exported by the libraries, so every time you talk about the application using internal stuff, I hear un-exported stuff.

`AES_ige_encrypt`, in *my* mind, is not internal before we stop exporting it.  Until then, it *is* exported.

mattcaswell:
Closing this in favour of the latest iteration in #10558 which seems to have more support.

PR #10541, opened 2019-11-29
Avoid invoking memcpy on NULL pointer in ASN1_TYPE_get_octetstring

pauldreik:
Invoking memcpy on NULL is undefined behaviour, even if the size is zero.

https://en.cppreference.com/w/c/string/byte/memcpy

CLA: trivial

### DIFF
--- a/crypto/asn1/evp_asn1.c
+++ b/crypto/asn1/evp_asn1.c
@@ -27,7 +27,10 @@ int ASN1_TYPE_set_octetstring(ASN1_TYPE *a, unsigned char *data, int len)
     return 1;
 }
 
-/* int max_len:  for returned value    */
+/* int max_len:  for returned value
+ * if passing NULL in data, nothing is copied but the necessary length
+ * for it is returned.
+ */
 int ASN1_TYPE_get_octetstring(const ASN1_TYPE *a, unsigned char *data, int max_len)
 {
     int ret, num;
@@ -43,7 +46,8 @@ int ASN1_TYPE_get_octetstring(const ASN1_TYPE *a, unsigned char *data, int max_l
         num = ret;
     else
         num = max_len;
-    memcpy(data, p, num);
+    if (num > 0 && data != NULL)
+        memcpy(data, p, num);
     return ret;
 }
 


InfoHunter:
It seems the C90 standard says either NULL `dst` or `src` will be an undefined behaviour. The patch you proposed seems not to address that?

pauldreik:
> It seems the C90 standard says either NULL `dst` or `src` will be an undefined behaviour. The patch you proposed seems not to address that?

I wanted to keep the change minimal, to avoid UB invoked by the following user code (error handling omitted for clarity): 
```c
int ret = ASN1_TYPE_get_octetstring(a, NULL, 0);
// create a buffer of the appropriate size
unsigned char* buf = makeBuffer(ret);
ASN1_TYPE_get_octetstring(a, buf, ret);
```

I am positive to asserting the pointers not being NULL in the non-zero size case, but I am not sure what the openssl policy is on assertions (except that it is being worked on: https://github.com/openssl/web/issues/49).  Is it ok to just add an assert(data) and assert(p)?

paulidale:
Assertions are not allowed.  We do have an `ossl_assert` which can be used.  For debug builds it asserts for non-debug it just tests the condition.

In this case I'd just NULL check the pointers and return a failure.

A case could be made that when data is NULL the function could return the size without copying anything.

pauldreik:
I remade the commit such that it won't copy unless size is positive and the supplied buffer is non-null. This solves my problem and enables the user to query for the necessary buffer length, suggested by @paulidale above. The test passes locally.

paulidale (approved):

t8m (approved):

t8m:
I think this is acceptable with CLA: trivial although borderline. @paulidale do you agree?

pauldreik:
I agree this is maybe not trivial anymore.

I have no problem signing a CLA, just have to read it first :-)

pauldreik:
I just signed the ICLA and followed the instructions on 
https://www.openssl.org/policies/cla.html to send it via email. Let's 
see what happens.

mattcaswell:
CLA has been received and recorded.

pauldreik:
I rebased on master and removed the trivial CLA note in the commit message.

t8m (approved):

t8m:
Merged to master. Thank you for your contribution.

PR #10550, opened 2019-12-01
Difference between EVP_CipherInit and EVP_CipherInit_ex

### LINKED ISSUE #10455: EVP_CIPHER_CTX_FLAG_WRAP_ALLOW flag support
I'm implementing the new Russian GOST CMS specification. It uses the key wrap algorithm described here: https://tools.ietf.org/html/draft-smyshlyaev-tls12-gost-suites-06#section-8.2

I've implemented the algorithm as a cipher with the `EVP_CIPH_WRAP_MODE` flag.

The `EVP_CipherInit_ex` function does not reset the `EVP_CIPHER_CTX` object and the `EVP_CIPHER_CTX_FLAG_WRAP_ALLOW` stays untouched, so the behavior seems a bit controversial (and undocumented, at least for the 1.1.1 branch). The `EVP_CipherInit_ex` function also especially preserves the `EVP_CIPHER_CTX_FLAG_WRAP_ALLOW` flag. The 
`EVP_CipherInit` function cleans up the passed context. 

I see the following variants:

1. Document the difference in behavior.
2. Update the `EVP_CipherInit` function to preserve the `EVP_CIPHER_CTX_FLAG_WRAP_ALLOW` flag on context reset.

I'm ready to implement any of these variants.

beldmit:
Fixes #10455

##### Checklist

### DIFF
--- a/doc/man3/EVP_EncryptInit.pod
+++ b/doc/man3/EVP_EncryptInit.pod
@@ -668,13 +668,13 @@ EVP_CipherInit_ex() and EVP_CipherFinal_ex() because they can reuse an
 existing context without allocating and freeing it up on each call.
 
 There are some differences between functions EVP_CipherInit() and
-EVP_CipherInit_ex(), significant in some circumstances. First, EVP_CipherInit()
-fills the passed context object with zeroes. It means that the flags set for
-the CTX are removed, and it is especially important for the
+EVP_CipherInit_ex(), significant in some circumstances. EVP_CipherInit() fills
+the passed context object with zeroes. It means that the flags set for the CTX
+are removed, and it is especially important for the
 B<EVP_CIPHER_CTX_FLAG_WRAP_ALLOW> flag treated specially in
-EVP_CipherInit_ex(). Second, EVP_CipherInit() does not allow step-by-step
-initialization of the ctx when the B<key> and B<iv> are passed in separate
-calls.
+EVP_CipherInit_ex(). As a consequence, EVP_CipherInit() does not allow
+step-by-step initialization of the ctx when the B<key> and B<iv> are passed in
+separate calls.
 
 EVP_get_cipherbynid(), and EVP_get_cipherbyobj() are implemented as macros.
 
--- a/doc/man3/EVP_EncryptInit.pod
+++ b/doc/man3/EVP_EncryptInit.pod
@@ -669,7 +669,7 @@ existing context without allocating and freeing it up on each call.
 
 There are some differences between functions EVP_CipherInit() and
 EVP_CipherInit_ex(), significant in some circumstances. EVP_CipherInit() fills
-the passed context object with zeroes. It means that the flags set for the CTX
+the passed context object with zeros. It means that the flags set for the CTX
 are removed, and it is especially important for the
 B<EVP_CIPHER_CTX_FLAG_WRAP_ALLOW> flag treated specially in
 EVP_CipherInit_ex(). As a consequence, EVP_CipherInit() does not allow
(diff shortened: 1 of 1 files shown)


beldmit:
When the text is cleared out, it's worth backporting it to the 1.1.1 branch.

doc/man3/EVP_EncryptInit.pod
```diff
@@ -667,6 +667,15 @@ EVP_EncryptFinal_ex(), EVP_DecryptInit_ex(), EVP_DecryptFinal_ex(),
 EVP_CipherInit_ex() and EVP_CipherFinal_ex() because they can reuse an
 existing context without allocating and freeing it up on each call.
 
+There are some differences between functions EVP_CipherInit() and
+EVP_CipherInit_ex(), significant in some circumstances. First, EVP_CipherInit()
```
levitte: Drop "First,"

doc/man3/EVP_EncryptInit.pod
```diff
@@ -667,6 +667,15 @@ EVP_EncryptFinal_ex(), EVP_DecryptInit_ex(), EVP_DecryptFinal_ex(),
...
+fills the passed context object with zeroes. It means that the flags set for
+the CTX are removed, and it is especially important for the
+B<EVP_CIPHER_CTX_FLAG_WRAP_ALLOW> flag treated specially in
+EVP_CipherInit_ex(). Second, EVP_CipherInit() does not allow step-by-step
```
levitte: Replace "Second," with "As a consequence,"

doc/man3/EVP_EncryptInit.pod
beldmit: Done.

doc/man3/EVP_EncryptInit.pod
beldmit: Done

beldmit:
zeroes => zeros

doc/man3/EVP_EncryptInit.pod
```diff
@@ -667,6 +667,15 @@ EVP_EncryptFinal_ex(), EVP_DecryptInit_ex(), EVP_DecryptFinal_ex(),
 EVP_CipherInit_ex() and EVP_CipherFinal_ex() because they can reuse an
 existing context without allocating and freeing it up on each call.
 
+There are some differences between functions EVP_CipherInit() and
+EVP_CipherInit_ex(), significant in some circumstances. EVP_CipherInit() fills
+the passed context object with zeros. It means that the flags set for the CTX
+are removed, and it is especially important for the
+B<EVP_CIPHER_CTX_FLAG_WRAP_ALLOW> flag treated specially in
+EVP_CipherInit_ex(). As a consequence, EVP_CipherInit() does not allow
```
t8m: These two are a little bit unrelated. Or rather I would switch the order of these two sentences: ...fills the passed context object with zeros. As a consequence, EVP_CipherInit() does not allow....calls. It also means that flags set for the CTX prior to the call are removed and that is especially important...

doc/man3/EVP_EncryptInit.pod
```diff
@@ -667,6 +667,15 @@ EVP_EncryptFinal_ex(), EVP_DecryptInit_ex(), EVP_DecryptFinal_ex(),
...
+step-by-step initialization of the ctx when the B<key> and B<iv> are passed in
```
t8m: Also `I<key> and I<iv>`

doc/man3/EVP_EncryptInit.pod
beldmit: Fixed. Thanks!

doc/man3/EVP_EncryptInit.pod
beldmit: Fixed. Thanks!

beldmit:
Merged. Thanks!

PR #10556, opened 2019-12-02
Added parameter -win-comp to CMS command line utility

davrux:
Added parameter -win-comp to CMS command line utility, to support
Windows 10 compatible KARI keyEncryptionAlgorithm ASN.1.

Windows 10 expects wrapping algorithm parameter to be NULL. On absent
parameter, Windows 10 is not able to decrypt CMS.

Output OpenSSL CMS without -win-comp. Not decryptable on Windows 10:

keyEncryptionAlgorithm:
          algorithm: dhSinglePass-stdDH-sha1kdf-scheme (1.3.133.16.840.63.0.2)
          parameter: SEQUENCE:
    0:d=0  hl=2 l=  11 cons: SEQUENCE
    2:d=1  hl=2 l=   9 prim:  OBJECT            :id-aes128-wrap

Output OpenSSL CMS with win-comp. Decryptable on Windows 10:

keyEncryptionAlgorithm:
          algorithm: dhSinglePass-stdDH-sha1kdf-scheme (1.3.133.16.840.63.0.2)
          parameter: SEQUENCE:
    0:d=0  hl=2 l=  13 cons: SEQUENCE
    2:d=1  hl=2 l=   9 prim:  OBJECT            :id-aes128-wrap
   13:d=1  hl=2 l=   0 prim:  NULL

CLA: trivial

##### Checklist
- [ x] documentation is added or updated
- [ x] tests are added or updated


slontis:
This does not look trivial to me - it looks like a workaround [also known as a hack :)]
 If we are going to handle this kind of thing I think it may need a bit more thought about if this is the correct place for this. Ideally it would be handled at a lower level - not just in the command line utility..

beldmit:
I agree that the variants of the parameters should be processed somewhere in the low-level cipher code but anyway it makes sense to be compatible with some buggy wide-used implementations and have a way to manage it at the application level.

slontis:
>but anyway it makes sense to be compatible with some buggy wide-used implementations and have a way to manage it at the application level.

Propagating hacks that use EVP_CIPHER_meth_dup() [which we want to deprecate at some point] is probably not the way forward :)

beldmit:
Sure. The other way to nowhere is the usage of an environment variable.

slontis:
I was think the app would set a override flag to force this behavior in the cms output.. The flag would then need to be handled down in the cms code..

t8m:
This needs to be PR against master as this is a feature PR.

t8m:
Closing. Please open a new PR against master if still interested in the functionality being added and have CLA signed.

PR #9138, opened 2019-06-12
Remove asn1 module dependency from RSASSA-PKCS1-v1_5.

davidmakepeace:
Replace use of the asn1 module (X509_SIG, X509_ALGOR, ASN1_TYPE, ASN1_OCTET_STRING, i2d_X509_SIG(), etc.) as well as OID lookups using OBJ_nid2obj() with pre-generated DigestInfo encodings for MD2, MD5, MDC-2, SHA-1, SHA-2 and SHA-3; the encoding is selected based on the NID. This is
similar to the approach used by the FOM. 

It is assumed that NID constants such as NID_sha256 will be available in all providers.

An alternative to pre-generating DigestInfo encodings for digests known to be used with RSASSA-PKCS1-v1_5 would be to use a simple ASN.1 encoder based on #9067 or #9111, but that assumes that providers will have the ability to map from NIDs to objs (OBJ_nid2obj()).

##### Checklist

- [ ] documentation is added or updated
- [ ] tests are added or updated

PR #19, opened 2013-09-12
Update custom TLS extension and supplemental data 'generate' callbacks t...

scottdeboy:
...o support sending an alert.

If multiple TLS extensions are expected but not received, the TLS extension and supplemental data 'generate' callbacks are the only chance for the receive-side to trigger a specific TLS alert during the handshake.

Removed logic which no-op'd TLS extension generate callbacks (as the generate callbacks need to always be called in order to trigger alerts), and updated the serverinfo-specific custom TLS extension callbacks to track which custom TLS extensions were received by the client, where no-ops for 'generate' callbacks are appropriate.

PR #10557, opened 2019-12-02
PROV: add RSA signature implementation

levitte:
This is almost an exact copy of the DSA implementation, except we need
to pass the MD type to RSA_sign() and RSA_verify() (the corresponding
DSA functions don't care).


mattcaswell:
I get large numbers of test failures when I try this locally. Still investigating, however I'm surprised not to see any handling of `EVP_PKEY_CTX_set_rsa_pss_saltlen()` in this PR (and possibly other similar ctrls) - so that could be a cause??

mattcaswell:
...in fact no handling of pss at all AFAICT?

levitte:
Yup, PSS support remains to be done.  But I'm surprised that actually fails, because that should go to the legacy routines, since RSASSA-PSS is supposedly treated as a separate algorithm!

levitte:
There seems to be more going on.  I recognise the 'test_req` failures from my keygen work (I've a local fix there, and wonder if the same applies here)

levitte:
> ... since RSASSA-PSS is supposedly treated as a separate algorithm!

... 'xcept no, that's not at all true.  `test_rsapss.t` uses "normal" RSA keys and adds PSS parameters for signing.  Oh boy...

levitte:
I put this in WIP, it needs further work

richsalz:
> This is almost an exact copy of the DSA implementation, except we need
> to pass the MD type to RSA_sign() and RSA_verify() (the corresponding
> DSA functions don't care).

Can you pass the extra parameter to the DSA functions and have them ignore it?  Or is there current API's that get in the way of doing that?  (I'm hoping that the PARAM layer can just ignore things, e.g.)

levitte:
It wasn't as easy as I thought

levitte:
Almost there.  As usual, the CMS test is the last stumbling block...

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,884 @@
...
+        ret = RSA_private_encrypt(tbslen, tbs, sig, prsactx->rsa,
+                                  prsactx->pad_mode);
+    }
+
+    if (ret <= 0) {
+        ERR_raise(ERR_LIB_PROV, ERR_LIB_RSA);
+        return 0;
+    }
+
+    *siglen = ret;
```
mattcaswell: Do we need to cleanse tbuf?

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,884 @@
...
+            if (ret <= 0) {
+                ERR_raise(ERR_LIB_PROV, ERR_LIB_RSA);
+                return 0;
+            }
+            ret = sltmp;
+        } else {
+            ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_PADDING_MODE);
+            return 0;
+        }
+    } else {
```
mattcaswell: Should we be verifying the pad mode in this event? i.e. that its not one of the ones handled above

levitte:
Now, even CMS passes through on my machine.  It turned out that I had forgotten to take care of PSS_SALTLEN params

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,922 @@
...
+        if (prsactx->pad_mode == RSA_PKCS1_PADDING) {
+            if (!RSA_verify(EVP_MD_type(prsactx->md), tbs, tbslen,
+                            sig, siglen, prsactx->rsa)) {
+                ERR_raise(ERR_LIB_PROV, ERR_LIB_RSA);
+                return 0;
+            }
+            return 1;
+        }
+
+        if (tbslen != prsactx->mdsize) {
```
mattcaswell: Shouldn't this check happen before we call RSA_verify?

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,922 @@
...
+            return 0;
+        rslen = RSA_public_decrypt(siglen, sig, prsactx->tbuf, prsactx->rsa,
+                                   prsactx->pad_mode);
+        if (rslen == 0) {
+            ERR_raise(ERR_LIB_PROV, ERR_LIB_RSA);
+            return 0;
+        }
+    }
+
+    if ((rslen != tbslen) || memcmp(tbs, prsactx->tbuf, rslen))
```
mattcaswell: Is CRYPTO_memcmp necessary here or not?

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,922 @@
...
+    PROV_RSA_CTX *prsactx = (PROV_RSA_CTX *)vprsactx;
+
+    if (prsactx == NULL)
+        return;
+
+    RSA_free(prsactx->rsa);
+    EVP_MD_CTX_free(prsactx->mdctx);
+    EVP_MD_free(prsactx->md);
+    EVP_MD_free(prsactx->mgf1_md);
+    OPENSSL_free(prsactx->tbuf);
```
mattcaswell: clear_free?

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,922 @@
...
+        str = mdname;
+        if (!OSSL_PARAM_get_utf8_string(p, &str, sizeof(mdname)))
+            return 0;
+
+        str = mdprops;
+        if (propsp != NULL
+            && !OSSL_PARAM_get_utf8_string(propsp, &str, sizeof(mdprops)))
+            return 0;
+
+        if (prsactx->pad_mode != RSA_PKCS1_PSS_PADDING) {
```
mattcaswell: Is it allowable to set the digests before setting the pad mode? Or do we require the pad mode to be set first?

test/recipes/30-test_evp_data/evppkey.txt
```diff
@@ -378,7 +378,7 @@ Reason = operation not supported for this keytype
 
 Sign = RSA-2048
 Ctrl = rsa_mgf1_md:sha1
-Result = PKEY_CTRL_INVALID
+Result = PKEY_CTRL_ERROR
```
mattcaswell: Why do we get this change in behaviour?

util/libcrypto.num
```diff
@@ -4909,3 +4909,5 @@ i2d_X509_PUBKEY_fp                      ?	3_0_0	EXIST::FUNCTION:STDIO
 d2i_X509_PUBKEY_bio                     ?	3_0_0	EXIST::FUNCTION:
 i2d_X509_PUBKEY_bio                     ?	3_0_0	EXIST::FUNCTION:
 RSA_get0_pss_params                     ?	3_0_0	EXIST::FUNCTION:RSA
+EVP_PKEY_CTX_set_rsa_pss_saltlen        ?	3_0_0	EXIST::FUNCTION:RSA
+EVP_PKEY_CTX_get_rsa_pss_saltlen        ?	3_0_0	EXIST::FUNCTION:RSA
```
mattcaswell: The documentation for these functions needs updating - they are described as macros.

mattcaswell:
Is this out of WIP now?

crypto/rsa/rsa_lib.c
```diff
@@ -1127,3 +1135,68 @@ int EVP_PKEY_CTX_get0_rsa_oaep_label(EVP_PKEY_CTX *ctx, unsigned char **label)
...
+            && ctx->pmeth->pkey_id != EVP_PKEY_RSA_PSS)
+        return -1;
+
+    /* TODO(3.0): Remove this eventually when no more legacy */
+    if (!EVP_PKEY_CTX_IS_SIGNATURE_OP(ctx)
+        || ctx->op.sig.sigprovctx == NULL)
+        return EVP_PKEY_CTX_ctrl(ctx, -1, -1, EVP_PKEY_CTRL_RSA_PSS_SALTLEN,
+                                 saltlen, NULL);
+
+    /* We put in both, 'cause the pad mode applies for both */
```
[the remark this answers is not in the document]
levitte: To be cleaned away

providers/defltprov.c
```diff
@@ -366,6 +366,7 @@ static const OSSL_ALGORITHM deflt_signature[] = {
 #ifndef OPENSSL_NO_DSA
     { "DSA:dsaEncryption", "default=yes", dsa_signature_functions },
 #endif
+    { "RSA:rsaEncryption", "default=yes", rsa_signature_functions },
```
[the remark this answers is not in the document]
levitte: I'm confused, I thought that was to be treated as a separate algorithm...  (that's what the EVP_PKEY_METHOD does, even though it ends up calling very much the same functions otherwise).

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,884 @@
...
+    if (!setup_md(prsactx))
+        return 0;
+
+    if (prsactx->md != NULL) {
+        if (tbslen != prsactx->mdsize) {
+            ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_DIGEST_LENGTH);
+            return 0;
+        }
+
+        if (EVP_MD_type(prsactx->md) == NID_mdc2) {
```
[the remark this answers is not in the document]
levitte: I blame the legacy code:

https://github.com/openssl/openssl/blob/d3a27c5ee45a29edd9c0d60ad5929f67996f89fd/crypto/rsa/rsa_pmeth.c#L139

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,884 @@
...
+    char mgf1_mdname[80];        /* Should be big enough */
+    char mgf1_mdprops[80];       /* Should be big enough */
+    EVP_MD *mgf1_md;
+    /* PSS salt length */
+    int saltlen;
+    /* Minimum salt length or -1 if no PSS parameter restriction */
+    int min_saltlen;
+
+    /* Temp buffer */
+    unsigned char *tbuf;
```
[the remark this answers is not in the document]
levitte: I blame rsa_pmeth.c...  However, it is used several times throughout the life of a PROV_RSA_CTX, so this avoids having to allocate and free it for every time it's being used.

providers/implementations/signature/rsa.c
levitte: Possibly.  In that case, we should in rssa_pmeth.c as well.

providers/implementations/signature/rsa.c
levitte: Why?  Doesn't `RSA_public_decrypt()` do that?

providers/implementations/signature/rsa.c
levitte: No.  `RSA_verify()` has its own verification, and has more complex checks of this length.

providers/defltprov.c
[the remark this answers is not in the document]
mattcaswell: Hmmm. Well yes. And no. As you discovered you can do pss signatures (or just pkcs1 signatures) with an an RSA key with the "rsaEncryption" OID. But you can't do asym encryption with a key with the pss oid. But as far as this code is concerned you can sign with either. With the asym cipher code if you attempt to do an encryption operation with a PSS pad mode it will fail.

providers/implementations/signature/rsa.c
levitte: In that case, rsa_pmeth.c should be fixed too:

https://github.com/openssl/openssl/blob/d3a27c5ee45a29edd9c0d60ad5929f67996f89fd/crypto/rsa/rsa_pmeth.c#L287

providers/implementations/signature/rsa.c
mattcaswell: Ah yes! You're right. It does.

providers/implementations/signature/rsa.c
levitte: Your guess is as good as mine:

https://github.com/openssl/openssl/blob/d3a27c5ee45a29edd9c0d60ad5929f67996f89fd/crypto/rsa/rsa_pmeth.c#L517-L523

providers/implementations/signature/rsa.c
mattcaswell: In which case is this more of an assertion?

providers/implementations/signature/rsa.c
mattcaswell: Ok...looks like we already have this restriction so, no problem.

test/recipes/30-test_evp_data/evppkey.txt
levitte: The commit message explains it (d92bb0cdda3a40f4dc445090f04b8fed5749230b):

## test/recipes/30-test_evp_data/evppkey.txt

Tests that go through provider cannot recognise PKEY_CTRL_INVALID from
PKEY_CTRL_ERROR any more, because provided implementations' param
setting functions return 0 or 1.

util/libcrypto.num
levitte: Thanks, will do

levitte:
> Is this out of WIP now?

Yes

crypto/evp/m_sigver.c
```diff
@@ -102,6 +102,9 @@ static int do_sigver_init(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
 
     /* No more legacy from here down to legacy: */
 
+    if (pctx)
```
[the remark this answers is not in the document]
levitte: Fixed

crypto/rsa/rsa_lib.c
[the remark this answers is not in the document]
levitte: Done

providers/implementations/signature/rsa.c
levitte: Note that the manual does say this for getting:

> The EVP_PKEY_CTX_get_rsa_pss_saltlen() macro gets the RSA PSS salt length
for B<ctx>. The padding mode must have been set to B<RSA_PKCS1_PSS_PADDING>.

providers/implementations/signature/rsa.c
t8m: It's public key operation, so it shouldn't be necessary?

util/libcrypto.num
levitte: Done

providers/implementations/signature/rsa.c
mattcaswell: Good point. So, never mind.

providers/defltprov.c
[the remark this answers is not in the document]
levitte: The thing is that we *must* keep them separate, or there will be hell to pay later, when we attempt to wrap RSA-PSS keys in PKCS#8.  Making that an alias for RSA will lead to all kinds of strange results.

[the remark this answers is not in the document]
levitte: Wouldn't aliasing them also mean that the ASYM_CIPHER RSA implementation suddenly be called for RSA-PSS keys, when it really shouldn't?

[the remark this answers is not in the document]
levitte: I have in my mind that I would do a sweep over all our provided RSA implementations and update for PSS where needed.  Do note that the KEYMGMT implementation has poor PSS support for the moment.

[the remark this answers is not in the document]
mattcaswell: > The thing is that we must keep them separate, or there will be hell to pay later, when we attempt to wrap RSA-PSS keys in PKCS#8. Making that an alias for RSA will lead to all kinds of strange results.

So do we just need a duplicate line here with the PSS names, but still pointing at the same signature functions (and similarly for the KEYMGMT line)?

providers/implementations/signature/rsa.c
levitte: Again, I blame rsa_pmeth.c:

https://github.com/openssl/openssl/blob/d3a27c5ee45a29edd9c0d60ad5929f67996f89fd/crypto/rsa/rsa_pmeth.c#L251-L277

Do note, though, that in that piece of code, `RSA_verify_PKCS1_PSS_mgf1()` is called with `tbs`, but `tbslen` isn't passed (gah!  Why the hell do we have such inconsistent APIs?????), so I suppose that's what this check protects.

levitte: So maybe I should move the check to make it more obvious what it protects

providers/defltprov.c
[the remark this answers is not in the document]
mattcaswell: > I have in my mind that I would do a sweep over all our provided RSA implementations and update for PSS where needed. Do note that the KEYMGMT implementation has poor PSS support for the moment.

Do you see that as a separate PR? I could accept that as an answer.

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,922 @@
...
+        }
+        prsactx->pad_mode = pad_mode;
+    }
+
+    p = OSSL_PARAM_locate_const(params, OSSL_SIGNATURE_PARAM_PSS_SALTLEN);
+    if (p != NULL) {
+        int saltlen;
+
+        if (prsactx->pad_mode != RSA_PKCS1_PSS_PADDING) {
+            ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_PSS_SALTLEN);
```
t8m: Should the reason value be different here?

providers/implementations/signature/rsa.c
levitte: Done

providers/defltprov.c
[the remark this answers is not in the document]
levitte: I see that as a separate PR, yes.

> So do we just need a duplicate line here with the PSS names, but still pointing at the same signature functions (and similarly for the KEYMGMT line)?

Not quite.  The newctx or init functions needs to be different.  And that's my main reason to want to make that a separate effort.

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,922 @@
...
+            return 0;
+        }
+        if (!OSSL_PARAM_get_int(p, &saltlen))
+            return 0;
+
+        if (saltlen < RSA_PSS_SALTLEN_MAX) {
```
t8m: Maybe a comment that this is the lowest possible negative value?

providers/implementations/signature/rsa.c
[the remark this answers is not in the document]
levitte: We do test it...

https://github.com/openssl/openssl/blob/d3a27c5ee45a29edd9c0d60ad5929f67996f89fd/test/recipes/30-test_evp_data/evppkey.txt#L287-L341

providers/implementations/signature/rsa.c
levitte: Done

providers/implementations/signature/rsa.c
levitte: Done

levitte:
I think I've addressed all reviewing comments so far.

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,922 @@
...
+
+    RSA_free(prsactx->rsa);
+    prsactx->rsa = vrsa;
+    if (RSA_get0_pss_params(prsactx->rsa) != NULL)
+        prsactx->pad_mode = RSA_PKCS1_PSS_PADDING;
+    else
+        prsactx->pad_mode = RSA_PKCS1_PADDING;
+    /* Maximum for sign, auto for verify */
+    prsactx->saltlen = RSA_PSS_SALTLEN_AUTO;
+    prsactx->min_saltlen = -1;
```
t8m: Where is the min_saltlen set to other value? Should it be initialized here from the RSA key pss params if it is set there? Basically do what pkey_pss_init() does if the key is PSS key.

levitte: As already commented (https://github.com/openssl/openssl/pull/10557#discussion_r353141265), I'm going to complete PSS support in a different PR.

providers/implementations/signature/rsa.c
levitte: Why?  This code handle the OSSL_SIGNATURE_PARAM_PSS_SALTLEN param, and *any* saltlen setting is invalid unless in PSS padding mode

providers/implementations/signature/rsa.c
levitte: Done

providers/implementations/signature/rsa.c
levitte: If you want, I could make that an `ERR_raise_data()` and have additional text explaining this further.

doc/man3/EVP_PKEY_CTX_ctrl.pod
```diff
@@ -272,16 +272,16 @@ buffer is expected to be the algorithm identifier byte.
...
+the digest length, B<RSA_PSS_SALTLEN_MAX> sets the salt length to the maximum
 permissible value. When verifying B<RSA_PSS_SALTLEN_AUTO> causes the salt length
 to be automatically determined based on the B<PSS> block structure. If this
-macro is not called maximum salt length is used when signing and auto detection
-when verifying is used by default.
+function is not called maximum salt length is used when signing and auto
+detection when verifying is used by default.
 
-The EVP_PKEY_CTX_get_rsa_pss_saltlen() macro gets the RSA PSS salt length
+The EVP_PKEY_CTX_get_rsa_pss_saltlen() function gets the RSA PSS salt length
```
mattcaswell: This HISTORY section of this file notes a number of functions that used to be macros. It should be updated with this change as well.

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,943 @@
...
+            ret--;
+            if (prsactx->tbuf[ret] != RSA_X931_hash_id(EVP_MD_type(prsactx->md))) {
+                ERR_raise(ERR_LIB_PROV, PROV_R_ALGORITHM_MISMATCH);
+                return 0;
+            }
+            if (ret != EVP_MD_size(prsactx->md)) {
+                ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_DIGEST_LENGTH);
+                return 0;
+            }
+            if (rout && (routsize == 0 || routsize >= (size_t)ret))
```
mattcaswell: This condition doesn't make sense to me. `rout` should never be NULL here (we already checked it above). Also if `routsize == 0` why do we attempt a memcpy into `rout`?

providers/implementations/signature/rsa.c
t8m: I'd see something like PROV_R_NOT_SUPPORTED or PROV_R_INVALID_PARAMETER to be more appropriate because the PROV_R_INVALID_PSS_SALTLEN would mean to me that the value is incorrect not that even the appearance of the saltlen param is incorrect.

doc/man3/EVP_PKEY_CTX_ctrl.pod
levitte: Fixed

providers/implementations/signature/rsa.c
levitte: I opted for already existing PROV_R_NOT_SUPPORTED, and added additional text.

providers/implementations/signature/rsa.c
levitte: I've rewritten this a bit to make better sense.

levitte:
Anything else?  The CIs seem happy

levitte:
... or not.  Damn github delayed update...

doc/man3/EVP_PKEY_CTX_ctrl.pod
```diff
@@ -272,16 +272,16 @@ buffer is expected to be the algorithm identifier byte.
 
 The EVP_PKEY_CTX_get_rsa_padding() function gets the RSA padding mode for B<ctx>.
 
-The EVP_PKEY_CTX_set_rsa_pss_saltlen() macro sets the RSA PSS salt length to
-B<len>. As its name implies it is only supported for PSS padding. Three special
-values are supported: B<RSA_PSS_SALTLEN_DIGEST> sets the salt length to the
-digest length, B<RSA_PSS_SALTLEN_MAX> sets the salt length to the maximum
+The EVP_PKEY_CTX_set_rsa_pss_saltlen() function sets the RSA PSS salt length to
+B<saltlen>. As its name implies it is only supported for PSS padding. Three
+special values are supported: B<RSA_PSS_SALTLEN_DIGEST> sets the salt length to
```
slontis: maybe itemize these values..

slontis: B<RSA_PSS_SALTLEN_AUTO> causes the salt length to be automatically determined based on the B<PSS> block structure when verifying.

doc/man3/EVP_PKEY_CTX_ctrl.pod
```diff
@@ -272,16 +272,16 @@ buffer is expected to be the algorithm identifier byte.
...
+the digest length, B<RSA_PSS_SALTLEN_MAX> sets the salt length to the maximum
 permissible value. When verifying B<RSA_PSS_SALTLEN_AUTO> causes the salt length
 to be automatically determined based on the B<PSS> block structure. If this
-macro is not called maximum salt length is used when signing and auto detection
-when verifying is used by default.
+function is not called maximum salt length is used when signing and auto
```
slontis: called then the

doc/man3/EVP_PKEY_CTX_ctrl.pod
```diff
@@ -272,16 +272,16 @@ buffer is expected to be the algorithm identifier byte.
...
-The EVP_PKEY_CTX_set_rsa_pss_saltlen() macro sets the RSA PSS salt length to
-B<len>. As its name implies it is only supported for PSS padding. Three special
-values are supported: B<RSA_PSS_SALTLEN_DIGEST> sets the salt length to the
-digest length, B<RSA_PSS_SALTLEN_MAX> sets the salt length to the maximum
+The EVP_PKEY_CTX_set_rsa_pss_saltlen() function sets the RSA PSS salt length to
+B<saltlen>. As its name implies it is only supported for PSS padding. Three
+special values are supported: B<RSA_PSS_SALTLEN_DIGEST> sets the salt length to
+the digest length, B<RSA_PSS_SALTLEN_MAX> sets the salt length to the maximum
 permissible value. When verifying B<RSA_PSS_SALTLEN_AUTO> causes the salt length
 to be automatically determined based on the B<PSS> block structure. If this
```
slontis: Should the If be on a newline?

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,962 @@
...
+
+        if (EVP_MD_type(prsactx->md) == NID_mdc2) {
+            unsigned int sltmp;
+
+            if (prsactx->pad_mode != RSA_PKCS1_PADDING) {
+                ERR_raise_data(ERR_LIB_PROV, PROV_R_INVALID_PADDING_MODE,
+                               "only PKCS#1 padding supported with MDC2");
+                return 0;
+            }
+            ret = RSA_sign_ASN1_OCTET_STRING(0, tbs, tbslen, sig, &sltmp,
```
slontis: NOOOOOOOOOOOOOOOO!!!!!!
All ASN1 needs to be outside the provider boundary.

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,962 @@
...
+                ERR_raise(ERR_LIB_PROV, PROV_R_KEY_SIZE_TOO_SMALL);
+                return 0;
+            }
+            if (!setup_tbuf(prsactx)) {
+                ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+                return 0;
+            }
+            memcpy(prsactx->tbuf, tbs, tbslen);
+            prsactx->tbuf[tbslen] = RSA_X931_hash_id(EVP_MD_type(prsactx->md));
+            ret = RSA_private_encrypt(tbslen + 1, prsactx->tbuf,
```
slontis: RSA_private_encrypt() currently does
{
    return rsa->meth->rsa_priv_enc
}
I assume this is another PR?

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,962 @@
...
+                                      sig, prsactx->rsa, RSA_X931_PADDING);
+            clean_tbuf(prsactx);
+        } else if (prsactx->pad_mode == RSA_PKCS1_PADDING) {
+            unsigned int sltmp;
+
+            ret = RSA_sign(EVP_MD_type(prsactx->md), tbs, tbslen,
```
slontis: RSA_sign() is full of stuff you dont want to do directly..
i.e:- encode_pkcs1()

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,962 @@
...
+            if (!setup_tbuf(prsactx))
+                return 0;
+            if (!RSA_padding_add_PKCS1_PSS_mgf1(prsactx->rsa,
+                                                prsactx->tbuf, tbs,
+                                                prsactx->md, prsactx->mgf1_md,
+                                                prsactx->saltlen)) {
+                ERR_raise(ERR_LIB_PROV, ERR_LIB_RSA);
+                return 0;
+            }
+            ret = RSA_private_encrypt(RSA_size(prsactx->rsa), prsactx->tbuf,
```
slontis: same issue here...
RSA_private_encrypt() calls rsa->meth??

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,962 @@
...
+            *routlen = ret;
+            if (routsize < (size_t)ret) {
+                ERR_raise(ERR_LIB_PROV, PROV_R_OUTPUT_BUFFER_TOO_SMALL);
+                return 0;
+            }
+            memcpy(rout, prsactx->tbuf, ret);
+        } else if (prsactx->pad_mode == RSA_PKCS1_PADDING) {
+            size_t sltmp;
+
+            ret = int_rsa_verify(EVP_MD_type(prsactx->md),
```
slontis: This calls RSA_public_decrypt() which has the ->meth accessors

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,962 @@
...
+
+    if (padding == RSA_X931_PADDING) {
+        if (RSA_X931_hash_id(mdnid) == -1) {
+            ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_X931_DIGEST);
+            return 0;
+        }
+    } else {
+        switch(mdnid) {
+        /* List of all supported RSA digests */
+        case NID_sha1:
```
slontis: Does this list need #ifdef FIPS_MODE ??

providers/implementations/signature/rsa.c
slontis: @davidmakepeace & @mattcaswell did some work in a PR related to doing special ASN1 code for the provider??

slontis: PR #9138

levitte: #9138 isn't merged yet, are you saying this PR should be pending on that?

providers/implementations/signature/rsa.c
levitte: Correct. This PR does *not* rewrite the RSA subsystem

providers/implementations/signature/rsa.c
levitte: Why?

slontis: Because there is a list of approved algorithms in fips mode. The same ones listed in fipsprov.c..

providers/implementations/signature/rsa.c
slontis: yes..

providers/implementations/signature/rsa.c
levitte: Also, I don't see the issue with going through rsa->meth per se, as long as we know that we have control over what low-level method we use.

slontis: As long as the ->meth are not settable in FIPS_MODE its ok..
It seems strange to have a dispatch function that calls a method table..

providers/implementations/signature/rsa.c
levitte: Or we'll, it doesn't really have to wait. You might note that only the default provider is affected here, I haven't touched the FIPS module, and was actually imagining making that a separate PR, since as you noted, that requires a rewrite of the RSA library proper. So from *this* implementation, the calls should stay the same.

providers/implementations/signature/rsa.c
levitte: The ->meth should be settable from within FIPS, but only to *our* RSA methods. 

You haven't noticed yet that we have layers upon layers of method structures? However, one might argue that those at such a low level should be thrown away, but that needs to be investigated. Meanwhile, as long as we have control, we should be fine.

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,962 @@
...
+}
+
+static int rsa_sign(void *vprsactx, unsigned char *sig, size_t *siglen,
+                    size_t sigsize, const unsigned char *tbs, size_t tbslen)
+{
+    PROV_RSA_CTX *prsactx = (PROV_RSA_CTX *)vprsactx;
+    int ret;
+
+    if (sig == NULL) {
+        *siglen = RSA_size(prsactx->rsa);
```
levitte: This might not be quite right...

slontis: rsa_sign.pod thinks it is correct??

providers/implementations/signature/rsa.c
levitte: Not sure how else to produce a signature. What output do you expect? The way it's currently designed, the produced signature is a DER encoded X509_SIG. The only way I can see that change is to make signatures a first class provider side object, and use serializers to encode them.

providers/implementations/signature/rsa.c
levitte: Ah, good point

providers/implementations/signature/rsa.c
levitte: (comment removed, it was utter bollocks... I should not comment before morning coffee...)

levitte: I'm not sure how all that is supposed to be done, er, indirectly. I'm also not quite sure what kind of output is expected, it seems a bit different depending on the algo

providers/implementations/signature/rsa.c
levitte: Yeah, sorry, got mixed up. RSA encodes a X509_SIG, *then* public-encrypts the result, while DSA does it the other way around...

providers/implementations/signature/rsa.c
slontis: PR #9138 can be done later.. All it does is a hardwired version of the ASN1 code..

doc/man3/EVP_PKEY_CTX_ctrl.pod
levitte: Done

doc/man3/EVP_PKEY_CTX_ctrl.pod
levitte: Done

doc/man3/EVP_PKEY_CTX_ctrl.pod
levitte: I moved it to appear before the enumeration

providers/implementations/signature/rsa.c
levitte: Right.

levitte: BTW, a solution for this part is to have it appear in the legacy provider only...  considering MDC2 is only provided there anyway, that should be doable.

slontis: Err in that case #ifndef FIPS_MODE will also do :)

providers/implementations/signature/rsa.c
levitte: Note that since there's no engine support in the FIPS module, and no other way to specify a different default RSA method, rsa->meth will point at rsa_pkcs1_ossl_meth (the method structure found in `crypto/rsa/rsa_ossl.c`

providers/implementations/signature/rsa.c
levitte: It shouldn't matter here, though...  since the digests available in the FIPS module is more limited than these constants show, all this does is check for constants that will never appear.

I'm more nervous about getting digests that don't have an associated NID...

levitte: I've changed things a bit, to have a local translation to NIDs.  Like I said, I see no harm in having a check for things that don't exist in the FIPS module, they will simply not be recognised here.  Otherwise, this function will have to move outside of this file and into its own, so we don't have to place this whole implementation in libfips and libnonfips.

providers/implementations/signature/rsa.c
levitte: I suppose that since that will be fixed by #9138, we can stop worrying here.

providers/implementations/signature/rsa.c
levitte: No

levitte:
CIs are finally happy again

levitte:
A rebase was needed.  I took the opportunity to squash

doc/man3/EVP_PKEY_CTX_ctrl.pod
```diff
@@ -274,17 +274,32 @@ buffer is expected to be the algorithm identifier byte.
...
-B<len>. As its name implies it is only supported for PSS padding. Three special
-values are supported: B<RSA_PSS_SALTLEN_DIGEST> sets the salt length to the
-digest length, B<RSA_PSS_SALTLEN_MAX> sets the salt length to the maximum
-permissible value. When verifying B<RSA_PSS_SALTLEN_AUTO> causes the salt length
-to be automatically determined based on the B<PSS> block structure. If this
-macro is not called maximum salt length is used when signing and auto detection
-when verifying is used by default.
+The EVP_PKEY_CTX_set_rsa_pss_saltlen() function sets the RSA PSS salt
+length to B<saltlen>. As its name implies it is only supported for PSS
+padding. If this function is not then the called maximum salt length
```
slontis: is not ??

mspncp: Today I learned "[gobbledygook](https://en.wiktionary.org/wiki/gobbledygook)". That's an interesting word and it sounds really funny, I'll have to remember it. ;-)

mspncp: You native speakers, you always use so many words which we foreign language speakers need to look up in the dictionary. This reminds me of a nice little online article in the [Financial Times](https://www.ft.com/content/bc8f9c5c-12a3-11ea-a225-db2f231cfeae?segmentId=09cf3415-e461-2c4a-a8cc-80acc4846679) which I happened to find this week via Twitter.  ;-)

levitte: I love getting to learn new words (didn't happen this time, though), so I'm partial for getting confused :wink:

mspncp: Mee too, of course. Just kiddin'.

levitte: Either way, I agree the sentence was awful.  Fixed.

richsalz: I'm flabbergasted that you managed to work so long on OpenSSL and didn't know of gobbledygook.  But I'm chuffed you know now :)

levitte: Er, who're you talking to @richsalz?  If it's to me, you didn't read "(didn't happen this time, though)" properly :wink:

levitte:
Travis issue was a timeout, not relevant here

doc/man3/EVP_PKEY_CTX_ctrl.pod
richsalz: @mspncp  Figure it couldn't hurt to add some new words :)

slontis:
Looks like this at least needs a rebase..
Did I see you mention handling PSS differently in another PR?

slontis:
travis error looks like a certificate has just reached its not before date :)

levitte:
> travis error looks a bit sinister..

... and I can't reproduce it, yet.  Trying with no-asm, maybe that makes a difference.

levitte:
Still can't reproduce it :-/

levitte:
Kicking the CIs

levitte:
If Travis is still acting up, I need some help trying to figure it out.  I have tried everything I can think of, but cannot reproduce the failure (see https://travis-ci.org/openssl/openssl/jobs/633824206)

levitte:
Oh cool, Travis stopped acting up!  I really hope that was a glitch or that it hadn't picked up on the latest tweaks.

doc/man3/EVP_PKEY_CTX_ctrl.pod
```diff
@@ -272,45 +272,60 @@ padding for RSA the algorithm identifier byte is added or checked and removed
...
-B<len>. As its name implies it is only supported for PSS padding. Three special
-values are supported: B<RSA_PSS_SALTLEN_DIGEST> sets the salt length to the
-digest length, B<RSA_PSS_SALTLEN_MAX> sets the salt length to the maximum
-permissible value. When verifying B<RSA_PSS_SALTLEN_AUTO> causes the salt length
-to be automatically determined based on the B<PSS> block structure. If this
-macro is not called maximum salt length is used when signing and auto detection
-when verifying is used by default.
+The EVP_PKEY_CTX_set_rsa_pss_saltlen() function sets the RSA PSS salt
+length to I<saltlen>. As its name implies it is only supported for PSS
+padding. If this function is not called then the called maximum salt
```
mattcaswell: Too many `called` here?

doc/man3/EVP_PKEY_CTX_ctrl.pod
```diff
@@ -272,45 +272,60 @@ padding for RSA the algorithm identifier byte is added or checked and removed
...
+sets the salt length to the digest length.
+
+=item B<RSA_PSS_SALTLEN_MAX>
+
+sets the salt length to the maximum permissible value.
+
+=item B<RSA_PSS_SALTLEN_AUTO>
+
+causes the salt length to be automatically determined based on the
+B<PSS> block structure when verifying.
```
mattcaswell: So, is it an error to use then when signing? If so might be worth saying that.

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,1006 @@
...
+    if (!setup_md(prsactx))
+        return 0;
+
+    if (prsactx->md != NULL) {
+        if (tbslen != prsactx->mdsize) {
+            ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_DIGEST_LENGTH);
+            return 0;
+        }
+
+#ifdef LEGACY_MODE
```
mattcaswell: Huh? Why would we be compiling this code for the legacy provider?? Surely it only ever sits in the FIPS or default providers...even if in the default provider it happens to be using a digest implemented in the legacy provider?

I would have expected this condition to be:

`#if !defined(FIPS_MODE)`

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,1006 @@
...
+     * Because the RSA library deals with NIDs, we need to translate.
+     * We do so using EVP_MD_is_a(), and therefore need a name to NID
+     * map.
+     */
+    struct name_to_nid_st {
+        const char *name;
+        int nid;
+    };
+    static const struct name_to_nid_st name_to_nid[] = {
+        { OSSL_DIGEST_NAME_SHA1,        NID_sha1      },
```
mattcaswell: Are these NID defines available in the FIPS provider? Just thinking ahead to when we pull this code in there.

levitte: *All* named NIDs are available in `include/openssl/obj_mac.h`, even for the FIPS module.

crypto/evp/pmeth_lib.c
```diff
@@ -873,7 +877,19 @@ static int legacy_ctrl_str_to_param(EVP_PKEY_CTX *ctx, const char *name,
         return ret;
     }
 
-
+    if (strcmp(name, "rsa_pss_saltlen") == 0) {
+        int saltlen;
+
+        if (!strcmp(value, "digest"))
```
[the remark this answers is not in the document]
levitte: Done

crypto/rsa/rsa_lib.c
```diff
@@ -1127,3 +1135,67 @@ int EVP_PKEY_CTX_get0_rsa_oaep_label(EVP_PKEY_CTX *ctx, unsigned char **label)
...
+    OSSL_PARAM pad_params[2], *p = pad_params;
+
+    if (ctx == NULL) {
+        ERR_raise(ERR_LIB_EVP, EVP_R_COMMAND_NOT_SUPPORTED);
+        /* Uses the same return values as EVP_PKEY_CTX_ctrl */
+        return -2;
+    }
+
+    /* If key type not RSA or RSA-PSS return error */
+    if (ctx->pmeth != NULL
```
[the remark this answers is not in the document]
levitte: Not quite.  If anything, we should check when `ctx->pmeth` *is* NULL as well, but using `EVP_KEYMGMT_is_a()` with suitable parameters...

That sounds like something for another PR.  I can imagine that a function `EVP_PKEY_is_a()` to check the key type would be generally useful.  I'm currently working through existing `EVP_PKEY_` functions that lack support for provided keys, I could make this part of that work.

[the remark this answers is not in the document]
levitte: In other words, I think that as part of *this* PR, this check is enough.  We do the same type of check all over the place, see `EVP_PKEY_CTX_set0_rsa_oaep_label()` (written by, I believe, you :wink:)

crypto/rsa/rsa_lib.c
```diff
@@ -1127,3 +1135,67 @@ int EVP_PKEY_CTX_get0_rsa_oaep_label(EVP_PKEY_CTX *ctx, unsigned char **label)
...
+    OSSL_PARAM pad_params[2], *p = pad_params;
+
+    if (ctx == NULL || saltlen == NULL) {
+        ERR_raise(ERR_LIB_EVP, EVP_R_COMMAND_NOT_SUPPORTED);
+        /* Uses the same return values as EVP_PKEY_CTX_ctrl */
+        return -2;
+    }
+
+    /* If key type not RSA or RSA-PSS return error */
+    if (ctx->pmeth != NULL
```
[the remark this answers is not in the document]
levitte: As above

doc/man3/EVP_PKEY_CTX_ctrl.pod
levitte: Yup.  Fixed

doc/man3/EVP_PKEY_CTX_ctrl.pod
levitte: No, it's not an error.  I've added a little more text explaining what happens when signing.

providers/implementations/signature/rsa.c
levitte: Well...  no.  MDC2 is a legacy algorithm, found in the legacy provider, and we *do* support the `no-legacy` configuration.

However, the code in here is implementation agnostic, it only depends on the presence of the macro `OSSL_DIGEST_NAME_MDC2`, so that guard should be completely unnecessary.  It may be that I had a stricter dependence on the MDC2 code at some point, and forgot to remove the guard when it became more agnostic.

Removing the guard now.

mattcaswell: The reason for the guard was https://github.com/openssl/openssl/pull/10557#pullrequestreview-326547695

This piece of code should not appear in the FIPS provider. It has nothing to do with the legacy provider. The fact that the MDC2 digest is found in the legacy provider is irrelevant (as you say the code here is implementation agnostic). It still seems to me that the correct guard is as I showed above (which I note that @slontis also suggests in the comment above!)

Mind you  - you can argue that we can defer this conversation until we actually pull this code into the FIPS provider.

providers/implementations/signature/rsa.c
```diff
@@ -190,7 +190,6 @@ static int rsa_sign(void *vprsactx, unsigned char *sig, size_t *siglen,
             return 0;
         }
 
-#ifdef LEGACY_MODE
```
t8m: The end label is also guarded by the `#ifdef LEGACY_MODE`

providers/implementations/signature/rsa.c
levitte: No, I'll add the FIPS_MODE guard.  Thanks for the reminder

levitte:
I'm about to test this merged with #10797, to see how it fares under *actual* use.  It's possible that this PR must go back to WIP as a result.

Please hold your approvals for now, but please still comment if you find odd things.

levitte:
I took this back to WIP, because it's missing a few crucial things, such as producing AlgorithmIdentifiers.  I have code to do that, but we need to get through a few more fixes for #10797 to be able to get this through existing tests (such as producing CSRs)

levitte:
This PR has matured quite a bit and should be reviewable at this point.  I'm keeping it WIP for a bit to allow investigation of a merge with #10797, but I don't expect any dramatic changes.

crypto/rsa/rsa_aid.c
```diff
@@ -0,0 +1,94 @@
...
+
+#define MD_CASE(name)                                                   \
+    case NID_##name:                                                    \
+        *len = sizeof(algorithmidentifier_##name##_der);                \
+        return algorithmidentifier_##name##_der
+
+const unsigned char *rsa_algorithmidentifier_encoding(int md_nid, size_t *len)
+{
+    switch (md_nid) {
+        MD_CASE(md2);
```
slontis: #ifndef FIPS_MODE around some of these?

crypto/rsa/rsa_lib.c
```diff
@@ -972,20 +980,20 @@ int EVP_PKEY_CTX_set_rsa_mgf1_md_name(EVP_PKEY_CTX *ctx, const char *mdname,
             && ctx->pmeth->pkey_id != EVP_PKEY_RSA_PSS)
         return -1;
 
-    *p++ = OSSL_PARAM_construct_utf8_string(OSSL_ASYM_CIPHER_PARAM_MGF1_DIGEST,
+    *p++ = OSSL_PARAM_construct_utf8_string(OSSL_PKEY_PARAM_MGF1_DIGEST,
```
slontis: Why does the legacy check on line 970 look different to the previous checks (that also check for NULL)

crypto/rsa/rsa_lib.c
```diff
@@ -1011,7 +1019,7 @@ int EVP_PKEY_CTX_get_rsa_mgf1_md_name(EVP_PKEY_CTX *ctx, char *name,
...
```
slontis: same question for line 1009

doc/man3/EVP_PKEY_CTX_ctrl.pod
```diff
@@ -270,45 +270,61 @@ padding for RSA the algorithm identifier byte is added or checked and removed
...
+
+=back
+
+The EVP_PKEY_CTX_get_rsa_pss_saltlen() function gets the RSA PSS salt length
+for I<ctx>. The padding mode must already have been set to
+B<RSA_PKCS1_PSS_PADDING>.
 
 The EVP_PKEY_CTX_set_rsa_keygen_bits() macro sets the RSA key length for
-RSA key generation to B<bits>. If not specified 1024 bits is used.
+RSA key generation to I<bits>. If not specified 1024 bits is used.
```
slontis: I dont think 1024 is correct anymore..

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,1104 @@
...
+        { NID_md5,       OSSL_DIGEST_NAME_MD5       },
+        { NID_md5_sha1,  OSSL_DIGEST_NAME_MD5_SHA1  },
+        { NID_md2,       OSSL_DIGEST_NAME_MD2       },
+        { NID_md4,       OSSL_DIGEST_NAME_MD4       },
+        { NID_mdc2,      OSSL_DIGEST_NAME_MDC2      },
+        { NID_ripemd160, OSSL_DIGEST_NAME_RIPEMD160 },
+        { NID_sha3_224,  OSSL_DIGEST_NAME_SHA3_224  },
+        { NID_sha3_256,  OSSL_DIGEST_NAME_SHA3_256  },
+        { NID_sha3_384,  OSSL_DIGEST_NAME_SHA3_384  },
+        { NID_sha3_512,  OSSL_DIGEST_NAME_SHA3_512  },
```
slontis: maybe remove the trailing ,

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,1104 @@
...
+            if ((size_t)RSA_size(prsactx->rsa) < tbslen + 1) {
+                ERR_raise(ERR_LIB_PROV, PROV_R_KEY_SIZE_TOO_SMALL);
+                return 0;
+            }
+            if (!setup_tbuf(prsactx)) {
+                ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+                return 0;
+            }
+            memcpy(prsactx->tbuf, tbs, tbslen);
+            prsactx->tbuf[tbslen] = RSA_X931_hash_id(prsactx->mdnid);
```
slontis: Does this function still return -1

providers/implementations/signature/rsa.c
levitte: Remind me, why? A trailing comma is perfectly valid as far as I know, and having it there means not having to remember later, when another line is added.

slontis:
needs a rebase

crypto/rsa/rsa_sign.c
```diff
@@ -97,6 +107,16 @@ static const unsigned char digestinfo_mdc2_der[] = {
       ASN1_OCTET_STRING, MDC2_DIGEST_LENGTH
 };
 # endif
+# ifndef OPENSSL_NO_RIPEMD160
+/* RIPEMD160 (1 3 36 3 3 1 2) */
+static const unsigned char digestinfo_ripemd160_der[] = {
+    ASN1_SEQUENCE, 0x0c + MDC2_DIGEST_LENGTH,
```
slontis: oops

crypto/rsa/rsa_aid.c
levitte: Done

crypto/rsa/rsa_lib.c
levitte: I...  don't know, and not sure I see clearly what you're talking about...

doc/man3/EVP_PKEY_CTX_ctrl.pod
levitte: Okie, fixed

providers/implementations/signature/rsa.c
levitte: Not here.  The MD nid is checked in `rsa_setup_md`, so the nid can't be faulty here.
(but, I found a bug in `rsa_setup_md`  thanks to this remark)

crypto/rsa/rsa_sign.c
levitte: Oops indeed.  Fixed

levitte:
Rebased

crypto/rsa/rsa_lib.c
```diff
@@ -774,6 +774,12 @@ int rsa_get0_all_params(RSA *r, STACK_OF(BIGNUM_const) *primes,
     if (r == NULL)
         return 0;
 
+    pnum = RSA_get_multi_prime_extra_count(r);
```
slontis: looks like this line needs to be removed..

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,1106 @@
...
+#include "internal/cryptlib.h"
+#include "internal/nelem.h"
+#include "internal/sizes.h"
+#include "crypto/rsa.h"
+#include "prov/providercommonerr.h"
+#include "prov/implementations.h"
+#include "prov/provider_ctx.h"
+
+static OSSL_OP_signature_newctx_fn rsa_newctx;
+static OSSL_OP_signature_sign_init_fn rsa_signature_init;
```
slontis: rsa_signature_init is declared many times here - looks like a block copy bug.

crypto/rsa/rsa_lib.c
levitte: Fixed

crypto/rsa/rsa_sign.c
```diff
@@ -97,6 +107,16 @@ static const unsigned char digestinfo_mdc2_der[] = {
       ASN1_OCTET_STRING, MDC2_DIGEST_LENGTH
 };
 # endif
+# ifndef OPENSSL_NO_RIPEMD160
```
slontis: legacyprov.c uses OPENSSL_NO_RMD160
Replace any references to OPENSSL_NO_RIPEMD160

levitte: Fixed

slontis:
Needs a rebase.. Also is this still WIP?

doc/man3/EVP_PKEY_CTX_ctrl.pod
```diff
@@ -206,29 +206,29 @@ These functions must only be called after the EVP_PKEY_CTX has been initialised
...
-B<p1> and B<p2>.
+I<ctx>. The key type used must match I<keytype> if it is not -1. The parameter
+I<optype> is a mask indicating which operations the control can be applied to.
+The control command is indicated in I<cmd> and any additional arguments in
+I<p1> and I<p2>.
 
-For B<cmd> = B<EVP_PKEY_CTRL_SET_MAC_KEY>, B<p1> is the length of the MAC key,
-and B<p2> is MAC key. This is used by Poly1305, SipHash, HMAC and CMAC.
+For I<cmd> = B<EVP_PKEY_CTRL_SET_MAC_KEY>, I<p1> is the length of the MAC key,
+and I<p2> is MAC key. This is used by Poly1305, SipHash, HMAC and CMAC.
```
slontis: is the

doc/man3/EVP_PKEY_CTX_ctrl.pod
```diff
@@ -374,31 +390,31 @@ negotiated protocol version. Otherwise it should be left unset.
 =head2 DSA parameters
 
 The EVP_PKEY_CTX_set_dsa_paramgen_bits() macro sets the number of bits used
-for DSA parameter generation to B<nbits>. If not specified, 1024 is used.
+for DSA parameter generation to I<nbits>. If not specified, 1024 is used.
```
slontis: 2048 instead of 1024

doc/man3/EVP_PKEY_CTX_ctrl.pod
```diff
@@ -374,31 +390,31 @@ negotiated protocol version. Otherwise it should be left unset.
...
 
 The EVP_PKEY_CTX_set_dsa_paramgen_q_bits() macro sets the number of bits in the
-subprime parameter B<q> for DSA parameter generation to B<qbits>. If not
+subprime parameter I<q> for DSA parameter generation to I<qbits>. If not
 specified, 160 is used. If a digest function is specified below, this parameter
```
slontis: 224 instead of 160

doc/man3/EVP_PKEY_CTX_ctrl.pod
levitte: Fixed

doc/man3/EVP_PKEY_CTX_ctrl.pod
levitte: fixed

doc/man3/EVP_PKEY_CTX_ctrl.pod
levitte: Fixed.

levitte:
It's out of WIP, nits are fixed, and rebase on a fresh master done

slontis:
Maybe remove [Pending on #11025] as well..
If 11025 find some other issues then deal with them in another PR.
It looks like this is self contained..

levitte:
> Maybe remove [Pending on #11025] as well..

Good point

slontis:
travis has a failure with the no options enabled for some reason..

levitte:
> travis has a failure with the no options enabled for some reason..

Are we looking at the same build?  When I look, it's the small footprint that fails.
I'll have a look at that.

slontis:
> Are we looking at the same build? When I look, it's the small footprint that fails.
> I'll have a look at that.

I mean build #19 which is configured with no_dh no_XX no_

levitte:
Yeah, that's a small footprint one

levitte:
There was a missing check.  Travis will hopefully be happier now.

levitte:
All CIs are happy

levitte:
@t8m, does your approval still stand?  Things have happened since, so perhaps a re-approval would be safer?

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,1109 @@
...
+
+static void clean_tbuf(PROV_RSA_CTX *ctx)
+{
+    if (ctx->tbuf != NULL)
+        OPENSSL_cleanse(ctx->tbuf, RSA_size(ctx->rsa));
+}
+
+static void free_tbuf(PROV_RSA_CTX *ctx)
+{
+    if (ctx->tbuf != NULL)
```
t8m: Just a nit not blocking approval: the if is unnecessary.

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,1109 @@
...
+                                                prsactx->saltlen);
+                if (ret <= 0) {
+                    ERR_raise(ERR_LIB_PROV, ERR_LIB_RSA);
+                    return 0;
+                }
+                return 1;
+            }
+        default:
+            ERR_raise_data(ERR_LIB_PROV, PROV_R_INVALID_PADDING_MODE,
+                           "Only X.931 or PKCS#1 v1.5 padding allowed");
```
t8m: PSS is also allowed here.

providers/implementations/signature/rsa.c
```diff
@@ -0,0 +1,1109 @@
...
+            return 0;
+        if (propsp != NULL
+            && !OSSL_PARAM_get_utf8_string(propsp, &pmdprops, sizeof(mdprops)))
+            return 0;
+
+        /* TODO(3.0) PSS check needs more work */
+        if (rsa_pss_restricted(prsactx)) {
+            /* TODO(3.0) figure out what to do for prsactx->md == NULL */
+            if (prsactx->md == NULL || EVP_MD_is_a(prsactx->md, mdname))
+                return 1;
```
t8m: Is return 1 here correct? No other parameters will be set then. But I don't see min_saltlen being set to anything else than -1 anywhere, so probably this is part of the TODO.

t8m:
And there is merge conflict now.

crypto/rsa/rsa_sign.c
```diff
@@ -124,18 +144,24 @@ ENCODE_DIGESTINFO_SHA(sha3_512, 0x0a, SHA512_DIGEST_LENGTH)
...
+# ifndef OPENSSL_NO_MD4
+    MD_CASE(md4)
+# endif
 # ifndef OPENSSL_NO_MD5
     MD_CASE(md5)
+# endif
+# ifndef OPENSSL_NO_RMD160
+    MD_CASE(ripemd160)
 # endif
     MD_CASE(sha1)
```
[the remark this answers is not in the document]
levitte: That's a good question.  According to [NIST's algorithm registration](https://csrc.nist.gov/projects/computer-security-objects-register/algorithm-registration), sha1WithRSA is perfectly acceptable, so I'll change that here.

providers/implementations/signature/rsa.c
levitte: This is part of the TODO, yes.  I'm planning a RSA-PSS sweep when all the basics have fallen into place.  Except a "RSA-PSS package" that touches all places at once.

providers/implementations/signature/rsa.c
levitte: Fixed

levitte:
> And there is merge conflict now.

Damnit!

levitte:
At least, it wasn't a conflict I had caused myself :wink:

t8m:
Now there is a travis failure and there seem to be some deprecated declarations warnings as well in the build logs.

levitte:
Ah, right!  Totally forgot that part

levitte:
Travis failure is just the usual timeout...

levitte:
I'll merge later tonight, or tomorrow morning

levitte:
Done

t8m:
This regressed MD5-SHA1 signatures.

PR #10558, opened 2019-12-02
 Deprecate the AES_ige_*() functions (take 3)

mattcaswell:
These functions were already partially deprecated. Now we do it fully.

This is an alternative approach to that shown in #10540, and also to the one in #10538 and #10367.

### DIFF
--- a/include/openssl/aes.h
+++ b/include/openssl/aes.h
@@ -73,17 +73,16 @@ void AES_cfb8_encrypt(const unsigned char *in, unsigned char *out,
 void AES_ofb128_encrypt(const unsigned char *in, unsigned char *out,
                         size_t length, const AES_KEY *key,
                         unsigned char *ivec, int *num);
-# ifndef OPENSSL_NO_DEPRECATED_3_0
+
 /* NB: the IV is _two_ blocks long */
-void AES_ige_encrypt(const unsigned char *in, unsigned char *out,
-                     size_t length, const AES_KEY *key,
-                     unsigned char *ivec, const int enc);
+DEPRECATEDIN_3_0(void AES_ige_encrypt(const unsigned char *in,
+    unsigned char *out, size_t length, const AES_KEY *key, unsigned char *ivec,
+    const int enc))
 /* NB: the IV is _four_ blocks long */
-void AES_bi_ige_encrypt(const unsigned char *in, unsigned char *out,
-                        size_t length, const AES_KEY *key,
-                        const AES_KEY *key2, const unsigned char *ivec,
-                        const int enc);
-# endif
+DEPRECATEDIN_3_0(void AES_bi_ige_encrypt(
+    const unsigned char *in, unsigned char *out, size_t length,
+    const AES_KEY *key, const AES_KEY *key2, const unsigned char *ivec,
+    const int enc))
 
 int AES_wrap_key(AES_KEY *key, const unsigned char *iv,
                  unsigned char *out,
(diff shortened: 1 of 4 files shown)


levitte:
This makes sense to me.

include/openssl/aes.h
```diff
@@ -73,17 +73,16 @@ void AES_cfb8_encrypt(const unsigned char *in, unsigned char *out,
...
 /* NB: the IV is _four_ blocks long */
-void AES_bi_ige_encrypt(const unsigned char *in, unsigned char *out,
-                        size_t length, const AES_KEY *key,
-                        const AES_KEY *key2, const unsigned char *ivec,
-                        const int enc);
-# endif
+DEPRECATEDIN_3_0(void AES_bi_ige_encrypt(
+    const unsigned char *in, unsigned char *out, size_t length,
+    const AES_KEY *key, const AES_KEY *key2, const unsigned char *ivec,
+    const int enc))
```
levitte: Is this alignment according to code style?  I'm sure @DDvO's checker script is gonna say a few things about these lines.

(and yes, I'm perfectly aware how very ugly the alternative is.  Still...)

t8m:
This has a disadvantage that you cannot find out where openssl itself uses the deprecated APIs with the simple addition of -Wdeprecated-declarations flag to the config.

mattcaswell:
> This has a disadvantage that you cannot find out where openssl itself uses the deprecated APIs with the simple addition of -Wdeprecated-declarations flag to the config.

True, although I don't anticipate that we would need to do that very often. Where we do I imagine it is only us (i.e. the committer team) that would need to, and the deprecations can be re-enabled with a very simple edit to macros.h

levitte:
no-deprecated should still be useful in that regard

include/openssl/aes.h
DDvO: > Is this alignment according to code style? I'm sure @DDvO's checker script is gonna say a few things about these lines.
> 
> (and yes, I'm perfectly aware how very ugly the alternative is. Still...)

Indeed, `check-format.pl` complains on that decl and the one before like this:
```
aes.h:79:indent = 4 != 38 for hanging '(':    unsigned char *out, size_t length, const AES_KEY *key, unsigned char *ivec,
aes.h:80:indent = 4 != 38 for hanging '(':    const int enc))
aes.h:83:indent = 4 != 41 for hanging '(':    const unsigned char *in, unsigned char *out, size_t length,
aes.h:84:indent = 4 != 41 for hanging '(':    const AES_KEY *key, const AES_KEY *key2, const unsigned char *ivec,
aes.h:85:indent = 4 != 41 for hanging '(':    const int enc))
```

A viable alternative formatting could be:
```
/* NB: the IV is _two_ blocks long */
DEPRECATEDIN_3_0(void
                 AES_ige_encrypt(const unsigned char *in, unsigned char *out,
                                 size_t length, const AES_KEY *key,
                                 unsigned char *ivec, const int enc))
/* NB: the IV is _four_ blocks long */
DEPRECATEDIN_3_0(void
                 AES_bi_ige_encrypt(const unsigned char *in, unsigned char *out,
                                    size_t length, const AES_KEY *key,
                                    const AES_KEY *key2,
                                    const unsigned char *ivec, const int enc))
```

richsalz: Does this work?
```
DEPRECATEDIN_3_0(
    void
    AES_bi_ige_encrypt(const unsigned char *in, unsigned char *out, size_t length,
                       const AES_KEY *key, const AES_KEY *key2,
                       const unsigned char *ivec, const int enc)
)
```
And ```const int``` is pointless BTW.

DDvO: The tool does not like that, neither does the Emacs C mode.

I've already added the possibility for such a relaxation in the tool the allows for alternative indentation like this within expressions, but not (yet?) at statement/declaration level.

mattcaswell:
I updated the formatting in line with the suggestion from @DDvO.

mattcaswell:
Pushed. Thanks.

PR #10561, opened 2019-12-03
Fix exit issue in travisci

zorrorffm:
Ungraceful 'exit' probably causes unexpeced error on background activity.
So replace 'exit' with recommended 'travis_terminate'. Also see
https://travis-ci.community/t/exit-0-cannot-exit-successfully-on-arm/5731/4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -130,7 +130,7 @@ matrix:
 before_script:
     - env
     - if [ "$TRAVIS_PULL_REQUEST" != "false" -a -n "$EXTENDED_TEST" ]; then
-          (git log -1 $TRAVIS_COMMIT_RANGE | grep '\[extended tests\]' > /dev/null) || exit 0;
+          (git log -1 $TRAVIS_COMMIT_RANGE | grep '\[extended tests\]' > /dev/null) || travis_terminate 0;
       fi
     - if [ -n "$DESTDIR" ]; then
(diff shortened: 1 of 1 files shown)


mspncp:
Just for curiosity: what does the Change-Id in the commit message refer to?

zorrorffm:
> 
> 
> Just for curiosity: what does the Change-Id in the commit message refer to?

Change-Id identifies commits that belong to the same review in our internal Gerrit . We can use it to track whether this commit is merged or not.  See https://gerrit-review.googlesource.com/Documentation/user-changeid.html

paulidale:
Travis has a [pull request](https://github.com/travis-ci/travis-build/pull/1816/files) fixing this which leads to the question should we also fix an upstream problem?

zorrorffm:
> 
> 
> Travis has a [pull request](https://github.com/travis-ci/travis-build/pull/1816/files) fixing this which leads to the question should we also fix an upstream problem?

 From API perspective, IMHO, 'travis_terminate' is still the preferred  way safely exit shell even upstream has a workaround avoiding the error caused by misusing 'exit'.

levitte:
I agree with @zorrorffm, let's use the documented stuff.

slontis:
Thanks zorrorffm.
As it was marked urgent I approved and merged to both branches.

PR #10563, opened 2019-12-03
Add a Setter for X509_REQ_get0_signature.

### LINKED ISSUE #10560: Setter equivalent of X509_REQ_get0_signature for OpenSSL > 1.1.0 appears missing
I've got a chunk of code that carefully builds up a clean certificate from another request.

```
   X509_REQ * req == read input
....
   creq = X509_REQ_new();
   ...
```
and at some point I need to populate the signature alg. Which I used to do with:

`   creq->sig_alg = X509_ALGOR_dup(req->sig_alg);
`
I understand that this const struct X509_algor_st * has become private around 1.1.x.

Now I can get the value with help of the getter; and duplicate it fine:
```

   const X509_ALGOR *psigalg;

   X509_REQ_get0_signature(req, NULL /* no need for signature */,&psigalg);

   X509_ALGOR *pcsigalg = X509_ALGOR_dup((X509_ALGOR *) psigalg);
```

but what is the right way to fill copy this into the X509_REQ struct ? Is there a setter I've missed? Or what is the modern 1.1.1 way of `not poking' in the DER struct during an d2i/i2d dance

Or should one approach this very differently ?

PS: actual code @ line 900 in https://source.redwax.eu/projects/RS/repos/mod_csr/browse/mod_csr.c#900

dirkx:
This should resolve #10560 

### DIFF
--- a/crypto/x509/x509_req.c
+++ b/crypto/x509/x509_req.c
@@ -286,6 +286,18 @@ void X509_REQ_get0_signature(const X509_REQ *req, const ASN1_BIT_STRING **psig,
         *palg = &req->sig_alg;
 }
 
+void X509_REQ_set0_signature(X509_REQ *req, ASN1_BIT_STRING *psig)
+{
+    if (req->signature)
+           ASN1_BIT_STRING_free(req->signature);
+    req->signature = psig;
+}
+
+int X509_REQ_set1_signature_algo(X509_REQ *req, X509_ALGOR *palg)
+{
+    return X509_ALGOR_copy(&req->sig_alg, palg);
+}
+
 int X509_REQ_get_signature_nid(const X509_REQ *req)
 {
     return OBJ_obj2nid(req->sig_alg.algorithm);
--- a/doc/man3/X509_ALGOR_dup.pod
+++ b/doc/man3/X509_ALGOR_dup.pod
@@ -2,7 +2,7 @@
 
 =head1 NAME
 
-X509_ALGOR_dup, X509_ALGOR_set0, X509_ALGOR_get0, X509_ALGOR_set_md, X509_ALGOR_cmp - AlgorithmIdentifier functions
+X509_ALGOR_dup, X509_ALGOR_set0, X509_ALGOR_get0, X509_ALGOR_set_md, X509_ALGOR_cmp, X509_ALGOR_copy - AlgorithmIdentifier functions
 
 =head1 SYNOPSIS
 
@@ -14,6 +14,7 @@ X509_ALGOR_dup, X509_ALGOR_set0, X509_ALGOR_get0, X509_ALGOR_set_md, X509_ALGOR_
                       const void **ppval, const X509_ALGOR *alg);
  void X509_ALGOR_set_md(X509_ALGOR *alg, const EVP_MD *md);
  int X509_ALGOR_cmp(const X509_ALGOR *a, const X509_ALGOR *b);
+ int X509_ALGOR_copy(X509_ALGOR *dest, const X509_ALGOR *src);
 
 =head1 DESCRIPTION
 
@@ -36,17 +37,24 @@ values for the message digest B<md>.
 X509_ALGOR_cmp() compares B<a> and B<b> and returns 0 if they have identical
 encodings and nonzero otherwise.
 
+X509_ALGOR_copy() copies the source values into the dest structs; making
+a duplicate of each (and free any thing pointed to from within *dest).
+
 =head1 RETURN VALUES
 
 X509_ALGOR_dup() returns a valid B<X509_ALGOR> structure or NULL if an error
 occurred.
 
-X509_ALGOR_set0() returns 1 on success or 0 on error.
+X509_ALGOR_set0() and X509_ALGOR_copy() return 1 on success or 0 on error.
 
 X509_ALGOR_get0() and X509_ALGOR_set_md() return no values.
 
 X509_ALGOR_cmp() returns 0 if the two parameters have identical encodings and
 nonzero otherwise.
+
+=head1 HISTORY
+
+The X509_ALGOR_copy() was added in 1.1.1e.
 
 =head1 COPYRIGHT
 
--- a/doc/man3/X509_get0_signature.pod
+++ b/doc/man3/X509_get0_signature.pod
@@ -2,10 +2,10 @@
 
 =head1 NAME
 
-X509_get0_signature, X509_get_signature_nid, X509_get0_tbs_sigalg,
-X509_REQ_get0_signature, X509_REQ_get_signature_nid, X509_CRL_get0_signature,
-X509_CRL_get_signature_nid, X509_get_signature_info, X509_SIG_INFO_get,
-X509_SIG_INFO_set - signature information
+X509_get0_signature, X509_REQ_set0_signature, X509_REQ_set1_signature_algo,
+X509_get_signature_nid, X509_get0_tbs_sigalg, X509_REQ_get0_signature, 
+X509_REQ_get_signature_nid, X509_CRL_get0_signature, X509_CRL_get_signature_nid, 
+X509_get_signature_info, X509_SIG_INFO_get, X509_SIG_INFO_set - signature information
 
 =head1 SYNOPSIS
 
@@ -14,6 +14,8 @@ X509_SIG_INFO_set - signature information
  void X509_get0_signature(const ASN1_BIT_STRING **psig,
                           const X509_ALGOR **palg,
                           const X509 *x);
+ void X509_REQ_set0_signature(X509_REQ *req, ASN1_BIT_STRING *psig);
+ int X509_REQ_set1_signature_algo(X509_REQ *req, X509_ALGOR *palg);
  int X509_get_signature_nid(const X509 *x);
  const X509_ALGOR *X509_get0_tbs_sigalg(const X509 *x);
 
@@ -40,6 +42,9 @@ X509_SIG_INFO_set - signature information
 X509_get0_signature() sets B<*psig> to the signature of B<x> and B<*palg>
 to the signature algorithm of B<x>. The values returned are internal
 pointers which B<MUST NOT> be freed up after the call.
+
+X509_set0_signature() and X509_REQ_set1_signature_algo() are the
+equivalent setters for the two values of X509_get0_signature().
 
 X509_get0_tbs_sigalg() returns the signature algorithm in the signed
 portion of B<x>.
@@ -88,6 +93,10 @@ X509_get_signature_info() returns 1 if the signature information
 returned is valid or 0 if the information is not available (e.g.
 unknown algorithms or malformed parameters).
 
+X509_REQ_set1_signature_algo() returns 0 on success; or 1 on an
+error (e.g. null ALGO pointer). X509_REQ_set0_signature does
+not return an error value.
+
 =head1 SEE ALSO
 
 L<d2i_X509(3)>,
@@ -117,6 +126,9 @@ The
 X509_REQ_get0_signature(), X509_REQ_get_signature_nid(),
 X509_CRL_get0_signature() and X509_CRL_get_signature_nid() were
 added in OpenSSL 1.1.0.
+
+The X509_REQ_set0_signature() and X509_REQ_set1_signature_algo()
+were added in OpenSSL 1.1.1e.
 
 =head1 COPYRIGHT
 
--- a/util/libcrypto.num
+++ b/util/libcrypto.num
@@ -4912,3 +4912,6 @@ RSA_get0_pss_params                     ?	3_0_0	EXIST::FUNCTION:RSA
 X509_cmp_timeframe                      ?	3_0_0	EXIST::FUNCTION:
 OSSL_CMP_MSG_get0_header                ?	3_0_0	EXIST::FUNCTION:CMP
 BIO_f_prefix                            ?	3_0_0	EXIST::FUNCTION:
+X509_ALGOR_copy                         ?	3_0_0	EXIST::FUNCTION:
+X509_REQ_set0_signature                 ?	3_0_0	EXIST::FUNCTION:
+X509_REQ_set1_signature_algo            ?	3_0_0	EXIST::FUNCTION:
--- a/crypto/asn1/x_algor.c
+++ b/crypto/asn1/x_algor.c
@@ -92,3 +92,31 @@ int X509_ALGOR_cmp(const X509_ALGOR *a, const X509_ALGOR *b)
         return 0;
     return ASN1_TYPE_cmp(a->parameter, b->parameter);
 }
+
+int X509_ALGOR_copy(X509_ALGOR *dest, const X509_ALGOR *src)
+{
+    if (src == NULL || dest == NULL)
+	return 0;
+
+    if (dest->algorithm)
+         ASN1_OBJECT_free(dest->algorithm);
+    dest->algorithm = NULL;
+
+    if (dest->parameter)
+        ASN1_TYPE_free(dest->parameter);
+    dest->parameter = NULL;
+
+    if (src->algorithm)
+        if ((dest->algorithm = OBJ_dup(src->algorithm)) == NULL)
+	    return 0;
+
+    if (src->parameter)
+        /* Assuming this is also correct for a BOOL.
+         * set does copy as a side effect.
+         */
+        if (ASN1_TYPE_set1(dest->parameter, 
+              src->parameter->type, src->parameter->value.ptr) == 0)
+	    return 0;
+
+    return 1;
+}
--- a/include/openssl/x509.h
+++ b/include/openssl/x509.h
@@ -489,6 +489,7 @@ void X509_ALGOR_get0(const ASN1_OBJECT **paobj, int *pptype,
                      const void **ppval, const X509_ALGOR *algor);
 void X509_ALGOR_set_md(X509_ALGOR *alg, const EVP_MD *md);
 int X509_ALGOR_cmp(const X509_ALGOR *a, const X509_ALGOR *b);
+int X509_ALGOR_copy(X509_ALGOR *dest, const X509_ALGOR *src);
 
 DECLARE_ASN1_DUP_FUNCTION(X509_NAME)
 DECLARE_ASN1_DUP_FUNCTION(X509_NAME_ENTRY)
@@ -693,6 +694,8 @@ X509_NAME *X509_REQ_get_subject_name(const X509_REQ *req);
 int X509_REQ_set_subject_name(X509_REQ *req, X509_NAME *name);
 void X509_REQ_get0_signature(const X509_REQ *req, const ASN1_BIT_STRING **psig,
                              const X509_ALGOR **palg);
+void X509_REQ_set0_signature(X509_REQ *req, ASN1_BIT_STRING *psig);
+int X509_REQ_set1_signature_algo(X509_REQ *req, X509_ALGOR *palg);
 int X509_REQ_get_signature_nid(const X509_REQ *req);
 int i2d_re_X509_REQ_tbs(X509_REQ *req, unsigned char **pp);
 int X509_REQ_set_pubkey(X509_REQ *x, EVP_PKEY *pkey);


crypto/x509/x509_req.c
```diff
@@ -286,6 +286,15 @@ void X509_REQ_get0_signature(const X509_REQ *req, const ASN1_BIT_STRING **psig,
         *palg = &req->sig_alg;
 }
 
+void X509_REQ_set0_signature(X509_REQ *req, ASN1_BIT_STRING *psigOrNull,
+                             X509_ALGOR *palgOrNull)
```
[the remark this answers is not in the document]
mspncp: Correction: I just noticed that it does not make sense to call the third parameter `sig_alg` because it is dereferenced. 

```
        req->sig_alg = *sig_alg;
```
But then at least I would call it `palg` (as in the getter) or `psig_alg`.

https://github.com/openssl/openssl/blob/429f4997de63209da2a90cdcf617587dab338182/crypto/x509/x509_req.c#L280-L296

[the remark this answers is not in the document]
mspncp: _Side note: The getter shows nicely how useless hungarian notation is, in particular if it is not used correctly: it should be `**ppalg`, not `**palg` ;-)_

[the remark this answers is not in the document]
dirkx: Thanks - should be fixed in the updated PR.

crypto/x509/x509_req.c
```diff
@@ -278,21 +278,21 @@ X509_NAME *X509_REQ_get_subject_name(const X509_REQ *req)
 }
 
 void X509_REQ_get0_signature(const X509_REQ *req, const ASN1_BIT_STRING **psig,
-                             const X509_ALGOR **palg)
+                             const X509_ALGOR **ppalg)
```
mspncp: Sorry, I shouldn't have mentioned the Hungarian notation. I was actually just pointing out in a side note that it was used inconsistently and therefore useless. I didn't expect expect that my remark  would inspire you to change `palg` to `ppalg`. Your change made things a little bit more inconsistent than they was before, because now the function prototype has a `**psig` with a single 'p' and a `**ppalg` with a double 'p'.

Also the inconsistency between the prototypes of the three similar functions below has increased. (The worst inconsistency being the different ordering of the arguments in `X509_get0_signature()`, sigh.).

https://github.com/openssl/openssl/blob/6eac9d95cf75964c8c80ed7bd15f2651a6b3b888/crypto/x509/x509_req.c#L280-L281

https://github.com/openssl/openssl/blob/6eac9d95cf75964c8c80ed7bd15f2651a6b3b888/crypto/x509/x509cset.c#L121-L122

https://github.com/openssl/openssl/blob/6eac9d95cf75964c8c80ed7bd15f2651a6b3b888/crypto/x509/x_x509.c#L243-L244


So I would ask you to revert the `ppalg` back to a `palg`, sorry for the inconvenience. The rest is fine, I'll approve as soon as the `ppalg` is reverted.

Sorry for the inconvenience I caused :-/

dirkx:
No worries !

crypto/x509/x509_req.c
dirkx: No worries. Reverted & pushed.

mspncp:
Note to committers: this pull request cherry-picks cleanly, except for the usual conflict in libcrypto.num, which can be trivially resolved:

```
diff --git a/util/libcrypto.num b/util/libcrypto.num
index 1dde397bed..4d11ad9f46 100644
--- a/util/libcrypto.num
+++ b/util/libcrypto.num
@@ -4586,3 +4586,4 @@ EVP_PKEY_meth_set_digestsign            4539      1_1_1e  EXIST::FUNCTION:
 EVP_PKEY_meth_set_digestverify          4540   1_1_1e  EXIST::FUNCTION:
 EVP_PKEY_meth_get_digestverify          4541   1_1_1e  EXIST::FUNCTION:
 EVP_PKEY_meth_get_digestsign            4542   1_1_1e  EXIST::FUNCTION:
+X509_REQ_set0_signature                 4543   1_1_1e  EXIST::FUNCTION:
```

dirkx:
No - is totally fine as is.

If you want a justification; then this setter is used/was needed in https://redwax.eu / the RedWax project -- a set of apache modules that make up a miniature CA, scep-agent, etc.

And it is specifically to not need this code https://source.redwax.eu/projects/RS/repos/mod_csr/commits/de009d533c0b2927be9efcb6456a1e4cb23b84d8#openssl_setter_compat.h any longer.

mspncp:
Thanks for the background information. However, no extra justification is needed. Missing accessors for structures which were made opaque in 1.1.0 are considered a bug and qualify for backporting to 1.1.1.

crypto/x509/x509_req.c
```diff
@@ -286,6 +286,15 @@ void X509_REQ_get0_signature(const X509_REQ *req, const ASN1_BIT_STRING **psig,
         *palg = &req->sig_alg;
 }
 
+void X509_REQ_set0_signature(X509_REQ *req, ASN1_BIT_STRING *psig,
+                             X509_ALGOR *palg)
+{
+    if (psig != NULL)
+        req->signature = psig;
```
mattcaswell: Shouldn't this free any pre-existing value first?

crypto/x509/x509_req.c
```diff
@@ -286,6 +286,15 @@ void X509_REQ_get0_signature(const X509_REQ *req, const ASN1_BIT_STRING **psig,
...
+    if (palg != NULL)
+        req->sig_alg = *palg;
```
mattcaswell: But an X509_ALGOR structure contains pointers which will just get duplicated by this copy operation. Doesn't this lead to double-frees and/or use-after-free if the original X509_ALGOR structure gets freed?

crypto/x509/x509_req.c
```diff
@@ -286,6 +286,15 @@ void X509_REQ_get0_signature(const X509_REQ *req, const ASN1_BIT_STRING **psig,
         *palg = &req->sig_alg;
 }
 
+void X509_REQ_set0_signature(X509_REQ *req, ASN1_BIT_STRING *psig,
```
mattcaswell: I think the memory management semantics of this function is very confusing. It took me several minutes of staring at this PR before I realised this - so I think users will also be confused. A "set0" function in OpenSSL is usually used to mean that the memory management of the passed arguments is transferred to the object. However this is only the case for *one* of these values. So one argument has its ownership transferred to the object, while the ownership remains with the caller for the other one.

Perhaps this function would be better split into two separate setters, i.e. a "set0" and a "set1"

doc/man3/X509_get0_signature.pod
```diff
@@ -41,6 +43,14 @@ X509_get0_signature() sets B<*psig> to the signature of B<x> and B<*palg>
 to the signature algorithm of B<x>. The values returned are internal
 pointers which B<MUST NOT> be freed up after the call.
 
+X509_set0_signature() is the equivalent setter. If NULL is passed - that
+value is ignored; and the value in B<*req> is not overwritten. The algorithm
+value B<*palg> is copied into the struct of the B<*req>; but the 
+signature value B<*palg> is simply a pointer from B<*req>. So calling 
+this function transfers the memory management of that value; and therefore 
+the  B<*palg> tha has been passed in should not be freed directly after
```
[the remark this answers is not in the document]
dirkx-saltrino-2019: done

doc/man3/X509_get0_signature.pod
```diff
@@ -41,6 +43,14 @@ X509_get0_signature() sets B<*psig> to the signature of B<x> and B<*palg>
 to the signature algorithm of B<x>. The values returned are internal
 pointers which B<MUST NOT> be freed up after the call.
 
+X509_set0_signature() is the equivalent setter. If NULL is passed - that
```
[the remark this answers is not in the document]
dirkx-saltrino-2019: done.

crypto/x509/x509_req.c
dirkx-saltrino-2019: fixed.

crypto/x509/x509_req.c
dirkx-saltrino-2019: I took earlier advise and tried to keep close to the get0 function. Happy to split it in two functions - that also makes it perhaps easier to explain the copy of the struct v.s.  pointer.

Does that also mean the get0 needs to be changed in a get0/1-- that is sort of an API change ?

crypto/x509/x509_req.c
dirkx-saltrino-2019: I think we're copying a struct here ?

crypto/x509/x509_req.c
mattcaswell: > I took earlier advise and tried to keep close to the get0 function. Happy to split it in two functions - that also makes it perhaps easier to explain the copy of the struct v.s. pointer.

I do think it should be split.

> Does that also mean the get0 needs to be changed in a get0/1-- that is sort of an API change

Well, no the get0 is actually correct. In both cases you get a pointer to the internal structure back. There is no "copying" going on. So the memory semantics is the same for both fields - the ownership remains with the object (which is the intended meaning for get0).

mspncp: First of all: thanks @mattcaswell for spotting the memory issues. I was too much distracted by other things to notice them.


IMHO splitting the setter into two is not really optimal because signature and signature algorithm belong together. An alternative to splitting would be to make the last parameter a `const X509_ALGOR *palg`, to emphasize that the struct is passed "by reference" only for efficiency reasons, but effectively only the value is copied.

mspncp:
BTW: `git diff master...` shows a lot of trailing whitespace.

mspncp:
Another thing came to my mind: currently there are three similar getters

https://github.com/openssl/openssl/blob/73ce68d4e1b525c3ce3f5769ab955e860632540b/include/openssl/x509.h#L576
https://github.com/openssl/openssl/blob/73ce68d4e1b525c3ce3f5769ab955e860632540b/include/openssl/x509.h#L692
https://github.com/openssl/openssl/blob/73ce68d4e1b525c3ce3f5769ab955e860632540b/include/openssl/x509.h#L746

and you are adding the setter counterpart for one of those three.

https://github.com/openssl/openssl/blob/73ce68d4e1b525c3ce3f5769ab955e860632540b/include/openssl/x509.h#L694

Wouldn't it make sense to add setters for the others, too?

crypto/x509/x509_req.c
mattcaswell: Yes, but that struct looks like this:

````
struct X509_algor_st {
    ASN1_OBJECT *algorithm;
    ASN1_TYPE *parameter;
} /* X509_ALGOR */ ;
````
i.e. it contains references to other objects. When you free the original X509_ALGOR structure using X509_ALGOR_free() it will also free those other objects. However since all you did was copy the references not the objects themselves - you'll get a use-after-free or a double-free.

File: crypto/x509/x509_req.c (hunk 5 above)
Comment (dirkx) [reply crypto/x509/x509_req.c]: Ok  so what is the consensus of this community  ? Happy to adapt.

**1.**  one get0 for both -- as they are a pair; or a get0 / 1 for each ?

**2.** And should the set0 be:

   **2.A:** much like all the other set0's which will do a free and the update the pointer; but not DUP (transfer mem mngt)

```
void NAMING_AUTHORITY_set0_authorityId(NAMING_AUTHORITY *n, ASN1_OBJECT* id)
{
    ASN1_OBJECT_free(n->namingAuthorityId);
    n->namingAuthorityId = id;
}
```

   **2.B:** or much like X509_NAME_set() and most other 'non 0/1'  plain XX_set()s which mostly do a DUP ?

```
int X509_NAME_set(X509_NAME **xn, const X509_NAME *name)
{
    X509_NAME *name_copy;

    if (*xn == name)
        return *xn != NULL;
    if ((name_copy = X509_NAME_dup(name)) == NULL)
        return 0;
    X509_NAME_free(*xn);
    *xn = name_copy;
    return 1;
}
```

**3)** Happy to add the other two methods - now or in a separate pull. Just let me know what is required/process wise desired.

With kind regards,

Dw.

File: crypto/x509/x509_req.c [hunk 8]
```diff
@@ -286,6 +286,22 @@ void X509_REQ_get0_signature(const X509_REQ *req, const ASN1_BIT_STRING **psig,
...
+    req->signature = psig;
+}
+
+void X509_REQ_set1_signature(X509_REQ *req, X509_ALGOR *palg)
+{
+    if (req->sig_alg.algorithm)
+        ASN1_OBJECT_free(req->sig_alg.algorithm);
+    if (req->sig_alg.parameter)
+        ASN1_TYPE_free(req->sig_alg.parameter);
+    req->sig_alg = *palg;
```
mattcaswell: This is still not quite right. It does not duplicate the internal pointers correctly. *Possibly* something like this might do it:

````
tmp = X509_ALGOR_dup(palg);
if (tmp == NULL)
    return 0;
req->sig_alg = *tmp;
OPENSSL_free(tmp);
````

But even this code makes all sorts of assumptions about the internal working of X509_ALGOR that it probably shouldn't. A nicer answer would be to add an internal `x509_algor_copy` function in x_algor.c that knows how to copy X509_ALGOR structures properly, e.g.

````
int x509_algor_copy(X509_ALGOR *dest, X509_ALGOR *src);
````

And then this setter simply becomes:
````
    return x509_algor_copy(&req->sig_alg, palg);
````

Comment (dirkx) [reply crypto/x509/x509_req.c]: Yes -fully agreed. So it seems we have two approaches in the codebase;

First -- much like all the other set0's which will do a free and the update the pointer; but not DUP (transfer mem mngt). E.g:

```
void NAMING_AUTHORITY_set0_authorityId(NAMING_AUTHORITY *n, ASN1_OBJECT* id)
{
    ASN1_OBJECT_free(n->namingAuthorityId);
    n->namingAuthorityId = id;
}
```

i.e. the recent setters.

Or - like what you allude to - and much like X509_NAME_set() and most other 'non 0/1'  plain XX_set()s which mostly do a DUP  -- the older setters. E.g:

```
int X509_NAME_set(X509_NAME **xn, const X509_NAME *name)
{
    X509_NAME *name_copy;

    if (*xn == name)
        return *xn != NULL;
    if ((name_copy = X509_NAME_dup(name)) == NULL)
        return 0;
    X509_NAME_free(*xn);
    *xn = name_copy;
    return 1;
}
```
So I can stay close the the other set0's and just prevent a leak (and note in the docs that the memory should not be freed - like the other set0's do).

Or do it more like the FOO_set_bar()'s which have a decent dup.

Comment (dirkx) [reply crypto/x509/x509_req.c]: And obviously use the x509_algor_copy() as that is shorted.

Comment (dirkx) [reply crypto/x509/x509_req.c]: And change the set0 to return an int (like set() - but not like the the void set0/1's).

Comment (mattcaswell) [reply crypto/x509/x509_req.c]: I think you're going to need to do the "set1" style approach for the X509_ALGOR. The thing is *this* setter is a little different to lots of the other setters in the codebase. The problem is for most other setters all we have to do is set a pointer. Easy. So in the `NAMING_AUTHORITY` example you give the members are all pointers:

````
struct NamingAuthority_st {
    ASN1_OBJECT* namingAuthorityId;
    ASN1_IA5STRING* namingAuthorityUrl;
    ASN1_STRING* namingAuthorityText;          /* i.e. DIRECTORYSTRING */
};
````
So, in a set 0, you can simply set the pointer. Or in a set1 you can dup it and then set it.

The sig_alg member of the X509_REQ structure is different though:

````
struct X509_req_st {
    X509_REQ_INFO req_info;     /* signed certificate request data */
    X509_ALGOR sig_alg;         /* signature algorithm */
    ASN1_BIT_STRING *signature; /* signature */
    CRYPTO_REF_COUNT references;
    CRYPTO_RWLOCK *lock;
# ifndef OPENSSL_NO_SM2
    ASN1_OCTET_STRING *sm2_id;
# endif
};
````
In this case the `sig_alg` member is *not* a pointer. It is an embeded structure. So you cannot simply do the set0 approach and set a pointer. You could individually set all the members and then free the passed in pointer - but I think this would be quite unexpected for the caller. Hence I think the set1 approach is the way to go. In this case you have dup all the *members* of the passed in X509_ALGOR, but not dup the whole containing X509_ALGOR.

This is why I suggest creating the function x509_algor_copy() to help implement this.

**Comment by dirkx** [comment]
@mattcaswell - so tried to make those changes - including a copy. This passes various tests; including valgrind on our sample code. But am not entirely sure that this is what you meant.

File: crypto/x509/x509_req.c (hunk 8 above)
Comment (dirkx) [reply crypto/x509/x509_req.c]: Ok - understood and think I got it implemented right.

**Comment by dirkx** [comment]
Is there a way to restart a CI test that failed on timeout ?

**Comment by mspncp** [comment]
> Is there a way to restart a CI test that failed on timeout ?

I just restarted the single travis job which was failing.

**Comment by mspncp** [comment]
That [job](https://travis-ci.org/openssl/openssl/jobs/625390692?utm_medium=notification&utm_source=github_status) seems to run in slooow mootion. It took 700 seconds just to execute the `ar` commands and it is still working on the tests...

```
$ if ! $make2; then echo -e '\052\052 FAILED -- MAKE'; travis_terminate 1; fi;
ar: creating apps/libapps.a
ar: creating libcrypto.a
ar: creating libssl.a
ar: creating providers/libcommon.a
ar: creating providers/libfips.a
ar: creating providers/libimplementations.a
ar: creating providers/liblegacy.a
ar: creating providers/libnonfips.a
...
```

**Comment by mspncp** [comment]
> > Is there a way to restart a CI test that failed on timeout ?
> 
> I just restarted the single travis job which was failing.

Ah, correction: The sevenhundred seconds are probably for the entire build. I forgot that most of the commands are not echoed anymore but written to a file which is only shown on error.

**Comment by mspncp** [comment]
Too slow again...


```
80-test_dtlsv1listen.t ............. ok
80-test_ocsp.t ..................... ok
80-test_pkcs12.t ................... ok
```

File: crypto/x509/x509_req.c [hunk 9]
```diff
@@ -286,6 +286,18 @@ void X509_REQ_get0_signature(const X509_REQ *req, const ASN1_BIT_STRING **psig,
...
 }
 
+void X509_REQ_set0_signature(X509_REQ *req, ASN1_BIT_STRING *psig)
+{
+    if (req->signature)
+           ASN1_BIT_STRING_free(req->signature);
+    req->signature = psig;
+}
+
+int X509_REQ_set1_signature(X509_REQ *req, X509_ALGOR *palg)
```
t8m: The @mattcaswell 's comment is still relevant. The name of the function is wrong. It should be X509_REQ_set1_sigalg or X509_REQ_set1_signature_algo or something similar.

dirkx: Missed that. Done & tested.

dirkx:
@mattcaswell Anything left for me to do here ? Any way I can re-trigger the CI build and hope that it does not time out this time without doing a fluff commit?

doc/man3/X509_ALGOR_dup.pod
```diff
@@ -36,12 +37,15 @@ values for the message digest B<md>.
...
+X509_ALGOR_copy() copies the source values into the dest structs; making
+a duplicate of each (and free any thing pointed to from within *dest).
+
 =head1 RETURN VALUES
 
 X509_ALGOR_dup() returns a valid B<X509_ALGOR> structure or NULL if an error
 occurred.
 
-X509_ALGOR_set0() returns 1 on success or 0 on error.
+X509_ALGOR_set0() and _X509_ALGOR_copy() return 1 on success or 0 on error.
```
t8m: Typo - superfluous underscore before the X509_ALGOR_copy. Also there is missing a HISTORY note that it was added in 1.1.1e.

doc/man3/X509_get0_signature.pod
```diff
@@ -41,6 +43,10 @@ X509_get0_signature() sets B<*psig> to the signature of B<x> and B<*palg>
 to the signature algorithm of B<x>. The values returned are internal
 pointers which B<MUST NOT> be freed up after the call.
 
+X509_set0_signature() and X509_REQ_set1_signature_algo() are the equivalent setter.
+value is ignored; and the value in B<*req> is not overwritten.  Calling
```
t8m: Some text seems to be missing here?

doc/man3/X509_get0_signature.pod
```diff
@@ -118,6 +128,8 @@ X509_REQ_get0_signature(), X509_REQ_get_signature_nid(),
 X509_CRL_get0_signature() and X509_CRL_get_signature_nid() were
 added in OpenSSL 1.1.0.
 
+The X509_REQ_set0_signature() was added in OpenSSL 1.1.1e.
```
t8m: ... and X509_REQ_set1_signature_algo().

util/libcrypto.num
```diff
@@ -4912,3 +4912,6 @@ RSA_get0_pss_params                     ?	3_0_0	EXIST::FUNCTION:RSA
 X509_cmp_timeframe                      ?	3_0_0	EXIST::FUNCTION:
 OSSL_CMP_MSG_get0_header                ?	3_0_0	EXIST::FUNCTION:CMP
 BIO_f_prefix                            ?	3_0_0	EXIST::FUNCTION:
+X509_ALGOR_copy                         ?	3_0_0	EXIST::FUNCTION:
```
t8m: We will need a separate PR for the 1.1.1 branch because the 3_0_0 needs to be replaced with 1_1_1e.

dirkx: So how is that done ?  I currently use the make command to (re)generate this file ? Should I check out that branch & regenerate ?

t8m: Yes. You will get a conflict in this file when cherry-picking the patches on the 1.1.1 branch anyway.

t8m:
As there are merge commits could you please just apply your changes in a single commit to fresh check-out and force-push?

dirkx:
@t8m - got them all merged into one single commit at the head of master in a branch (https://github.com/dirkx/openssl/tree/setter-getter-x509-req-algo) & regression tested -- but am struggling to rebase this pull request -- I keep getting problems with conflicts in libcrypto.num.
 
Or is it wiser to close this PR and file above (https://github.com/dirkx/openssl/tree/setter-getter-x509-req-algo) as a new PR ?

mspncp:
> ... am struggling to rebase this pull request -- I keep getting problems with conflicts in libcrypto.num.

Instead of rebasing your branch, you can just reset it to the the tip of the new one (i.e. setter-getter-x509-req-algo) and force-push the result.

Here's the magic spell how to do it. (Don't forget to check first that you have no uncommitted changes!):

```
git checkout master
git reset --hard setter-getter-x509-req-algo
git push origin master --force
```

(assuming that 'origin' points to your public GitHub fork of openssl)

> Or is it wiser to close this PR ... ?

It's a possibility, but it is not wiser ;-)

mspncp:
_For the future, I recommand to always create a separate topic branch for a pull request, instead of (mis-)using the master branch for it. That helps keeping things simpler._

dirkx:
@mspncp agreed - my bad -- this started as small tiny temporary side tweak that turned out to be much more needed than expected.

dirkx:
@t8m, @mattcaswell I take it I should file a totally separate pull request for  1.1.1  ?

mspncp:
@dirkx your pull request has no conflicts with 1.1.1 except for the usual ones in `libcrypto.num`. Those conflicts can be trivially resolved by discarding the conflicting file, rerunning make update (and discarding unrelated date changes in other files.) IMHO one of the committer could do this when merging your pull request. Do you agree @t8m, @mattcaswell?

```
git cherry-pick ...
... conflict ...
git checkout --ours -- util/libcrypto.num
make update
git add -- util/libcrypto.num
git checkout crypto  # workaround: discard  unrelated date changes by make update
git rebase --continue
```

t8m:
I have no problem with this unless this semi-automated conflict resolution in libcrypto.num somehow goes against the review/commit rules.

dirkx:
Anything I can do here ? Happy to re-submit the PR without libcrypto.num so it won't conflict with either branch - or split it into two PR's; one for each ?

mspncp:
> @belmit : Anything I can do here ? Happy to re-submit the PR without libcrypto.num so it won't conflict with either branch - or split it into two PR's; one for each ?

The best strategy for conflicts in libcrypto.num is normally to just discard the incoming changes and rerun make update:

```
git rebase master
# <git stops on conflict>
git checkout --ours util/libcrypto.num
make update
git add -u
git rebase --continue
```

mspncp:
> git checkout --ours util/libcrypto.num

(just in case you wonder: '--ours' == master, '--theirs' == the incoming changes from your branch being rebased)

t8m:
@beldmit ping, can you please re-approve? I can resolve the libcrypto.num conflict when merging as that is a trivial conflict.

t8m:
Merged to master as c72e593

t8m:
Merged to 1.1.1 as 4738d9e

PR #10564, opened 2019-12-03
Move providers/common/{ciphers,digests}/* to providers/implementations

levitte:
The idea to have all these things in providers/common was viable as
long as the implementations was spread around their main providers.
This is, however, no longer the case, so we move the common blocks
closer to the source that use them.


paulidale (dismissed):

levitte:
> CI failures might be relevant.

Very relevant.  Fixed

paulidale (approved):

slontis:
LGTM

levitte:
Merged.

68a51d59a29796803af3b6e7f0142feca2622c9e Move providers/common/{ciphers,digests}/* to providers/implementations

levitte:
I've discovered that there are "common" things that aren't so "common".  Follow-up PR coming up

Issue #3531, opened 2017-05-23
Internal treatment of character encoding in passwords


levitte:
This is a point of discussion.

Currently, passwords are treated internally as strings encoded in the current locale (Unix and VMS) or code page (Windows).  That is what is received from the `UI` API and older style password reading callbacks alike.  Applications are assumed to treat them as such and to do conversions to other encodings if required, regardless of if we're talking front end or back end (such as engines, or in #2011, STORE loaders).

What needs discussing is if this is how we should do it in the future, and also what future we're talking about (some changes are most likely impossible to implement before the next major version - 1.2.0 for the moment - because of compatibility demands).  For example, should we make sure that password strings are always converted to UTF8?

The question was raised in #2011.

petrovr:
Richard Levitte wrote:
> This is a point of discussion.
>
> Currently, passwords are treated internally as strings encoded in the current locale (Unix and VMS) or code page (Windows).  That is what is received from the `UI` API and older style password reading callbacks alike.  Applications are assumed to treat them as such and to do conversions to other encodings if required, regardless of if we're talking front end or back end (such as engines, or in #2011, STORE loaders).
Until now I use UI to get password as raw string, i.e. input is passed 
as to third party as is.
> What needs discussing is if this is how we should do it in the future, and also what future we're talking about (some changes are most likely impossible to implement before the next major version - 1.2.0 for the moment - because of compatibility demands).
Yes current behavior should be keep by default.

> For example, should we make sure that password strings are always converted to UTF8?
I wonder how UI will know encoding of input.
What about is input is from file in encoding A, application(executable) 
is not locale aware and  app is run in environment where locale is with 
different encoding.
Assuming "current locale" is not correct.

> The question was raised in #2011.
Dunno what could be relation of store2 to password. UI password 
management should be transparent.

If is so important I would like to propose conversion to be managed by 
two values : input (source) encoding and encoding of result (target). It 
should work for all UI event external implementation.

Roumen

levitte:
> Dunno what could be relation of store2 to password. UI password management should be transparent.

I agree.  However, the issue was brought up there.

> I wonder how UI will know encoding of input.
> What about if input is from file in encoding A, application(executable) is not locale aware and  app is run in environment where locale is with different encoding.
> Assuming "current locale" is not correct.

*Excellent* question!

[...]

PR #10570, opened 2019-12-04
Introduce additional -subj override, utf8 and multivalue RDN support

dirkx:
Replaces what was still left from PR # 1729 from 2016 (as https://github.com/openssl/openssl/commit/529586085e38487d45974817d4f3ff40f30e19f6 introduced most of this code for a slightly different usecase already).

Note to the reviewer - I am not clear why that patch also changes the Issuer around line 609. But have not changed that.

- [X] documentation is added or updated
- [X] tests are added and updated

### DIFF
--- a/apps/x509.c
+++ b/apps/x509.c
@@ -626,7 +626,7 @@ int x509_main(int argc, char **argv)
 
         if (req == NULL) {
             if (fsubj == NULL) {
-                BIO_printf(bio_err, "We need either a request/certificate to extract a subject fomr; or one set with -subj\n");
+                BIO_printf(bio_err, "We need either a request/certificate to extract a subject from; or one set with -subj\n");
                 goto end;
             }
             n = fsubj;
@@ -641,7 +641,7 @@ int x509_main(int argc, char **argv)
 
         if (n == fsubj) {
             X509_NAME_free(fsubj);
-	    fsubj = NULL;
+            fsubj = NULL;
         }
         if (!set_cert_times(x, NULL, NULL, days))
             goto end;
--- a/apps/x509.c
+++ b/apps/x509.c
@@ -633,10 +633,12 @@ int x509_main(int argc, char **argv)
             // Supress Overwritten warning below.
             fsubj = NULL;
         } else {
-           n = X509_REQ_get_subject_name(req);
+            n = X509_REQ_get_subject_name(req);
         }
 
         if (!X509_set_issuer_name(x, n))
+            goto end;
+        if (!X509_set_subject_name(x, n))
             goto end;
         if (!set_cert_times(x, NULL, NULL, days))
             goto end;
--- a/apps/x509.c
+++ b/apps/x509.c
@@ -630,8 +630,6 @@ int x509_main(int argc, char **argv)
                 goto end;
             }
             n = fsubj;
-            // Supress Overwritten warning below.
-            fsubj = NULL;
         } else {
             n = X509_REQ_get_subject_name(req);
         }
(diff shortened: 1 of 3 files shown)


dirkx:
Sorry - this will take some time - as the tests pass on my end.

kroeckx:
This seems related to #3531

apps/x509.c
```diff
@@ -612,22 +624,45 @@ int x509_main(int argc, char **argv)
             goto end;
         }
 
-        n = req == NULL ? fsubj : X509_REQ_get_subject_name(req);
-        if (!X509_set_issuer_name(x, n) || !X509_set_subject_name(x, n))
+        if (req == NULL) {
+            if (fsubj == NULL) {
+                BIO_printf(bio_err, "We need either a request/certificate to extract a subject fomr; or one set with -subj\n");
```
FdaSilvaYY: Typo : fomr ?

apps/x509.c
```diff
@@ -612,22 +624,45 @@ int x509_main(int argc, char **argv)
...
+        }
+
+        if (!X509_set_issuer_name(x, n))
             goto end;
+        if (!X509_set_subject_name(x, n))
+            goto end;
+
+        if (n == fsubj) {
+            X509_NAME_free(fsubj);
+	    fsubj = NULL;
```
FdaSilvaYY: nit : indent is wrong

dirkx:
I'd like to pause this pull request - as I am not confident that the CI tests by AppVeyor and my own 'make test' with valgrind & electric fence are good enough. Will be adding some more test cases first.

Let me know if it is easier process wise if I close this request & re-open. Or if I should leave it here 'as is' and continue once I've gotten the right test coverage.

richsalz:
You can close if you want.  Or edit the subject line of the PR and put "WIP:" in front, to indicate this is a work in progress.

dirkx:
Removed the WIP - added the other testcases to confirm that the alternative (only a public key; no certificate) works correctly.

apps/x509.c
dirkx: Fixed in this/last commit.

dirkx:
Fair to assume no-one is waiting for me for further changes ?

kroeckx:
I think we should always look at the config file as UTF-8. A command line switch should not have any effect on that. The question is just if we should enforce that it's valid UTF-8 and if we should do normalization. But I don't see how this PR has any effect on it, nor do I see any tests for it.

For command line, I think we need a general solution. I prefer one that just takes the locale settings into account and does the conversion as needed.

dirkx:
That was how this patch originally was. 

But commit 5295860 introduced in [line 166](https://github.com/openssl/openssl/commit/529586085e38487d45974817d4f3ff40f30e19f6#diff-f48b02166abaa1faa6c10358046a8fb2R166) it set to ASC. We occasionally need to force it to UTF8. 

And I would avoid normalisation within openssl - we have situations (germany v.s. swiss springs to mind) where some entities want their normalisation subtly different from each other. So best I guess to just accept the UTF8 as is.

Happy to remove that utf8 flag from this pull request though - as needing control over UTF8 for strings that do not need it is pretty rare anyway.

dirkx:
It is the override and multi-value RDN that we'd really like to see upstream.

dirkx:
@kroeckx what would you prefer met to ?

- Resubmit with the utf8 flag removed; and reverting to MBSTRING_ASC (as it was prior) ?
- Resubmit with the latter set to MBSTRING_UTF8 and perhaps a flag to force ASCII (e.g. if you have a nordic or austrian issue were you want to control the normalisation to match local customs) ?

With respect to:

kroeckx:
I suggest that you do the minimum amount of changes requires for
what you want to do, and leave all characterset related things to
some other PR.

It would also be good to hear from somebody else.

dirkx:
Ok - will do (and found some odd effects when writing a test case for it anyway).

dirkx:
Vacating this pull request - as someone else has introduced the UTF8 change that was to be removed from this pull request.

PR #10577, opened 2019-12-05
Make ECDSA_size() use consistent asn1 encoder.

slontis:
ECDSA signature lengths are calculated using i2d_ECDSA_SIG().
i2d_ECDSA_SIG() was changed in a previous PR to use a custom ASN1 encoder (using WPACKET)
so that the normal ASN1 encoder does not need to be pulled into the provider boundary.
For consistency ECDSA_size() has been changed to also use i2d_ECDSA_SIG() - this can now
be used directly inside the FIPS provider.

##### Checklist

### DIFF
--- a/crypto/ec/ec_asn1.c
+++ b/crypto/ec/ec_asn1.c
@@ -1342,32 +1342,27 @@ int ECDSA_SIG_set0(ECDSA_SIG *sig, BIGNUM *r, BIGNUM *s)
     return 1;
 }
 
-#ifndef FIPS_MODE
-int ECDSA_size(const EC_KEY *r)
+int ECDSA_size(const EC_KEY *ec)
 {
-    int ret, i;
-    ASN1_INTEGER bs;
-    unsigned char buf[4];
+    int ret;
+    ECDSA_SIG sig;
     const EC_GROUP *group;
+    const BIGNUM *bn;
 
-    if (r == NULL)
+    if (ec == NULL)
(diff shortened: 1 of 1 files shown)


crypto/ec/ec_asn1.c
```diff
@@ -1342,32 +1342,27 @@ int ECDSA_SIG_set0(ECDSA_SIG *sig, BIGNUM *r, BIGNUM *s)
...
+    if (ec == NULL)
         return 0;
-    group = EC_KEY_get0_group(r);
+    group = EC_KEY_get0_group(ec);
     if (group == NULL)
         return 0;
 
-    i = EC_GROUP_order_bits(group);
-    if (i == 0)
+    bn = EC_GROUP_get0_order(group);
```
slontis: I was concerned there might be a off by 1 problem here if the order does not have the top bit set..
There are some curves that have 1FFFFF at the top but those one should be ok..

slontis: r & s are modulus of the order.. so should be  less than this upper value

kaduk:
(There's a typo "ECDA_size()" in the commit message)

slontis:
Updated the commit to remove the typo..

slontis:
ping

slontis:
rebased with no new changes to the code.

slontis:
Thanks romen. Merged to master.

PR #10580, opened 2019-12-05
Deprecate the low level AES functions 

mattcaswell:
Use of the low level AES functions has been informally discouraged for a
long time. We now formally deprecate them.

Applications should instead use the EVP APIs, e.g. EVP_EncryptInit_ex,
EVP_EncryptUpdate, EVP_EncryptFinal_ex, and the equivalently named decrypt
functions.

### DIFF
--- a/crypto/cms/cms_env.c
+++ b/crypto/cms/cms_env.c
@@ -671,6 +671,7 @@ static int cms_RecipientInfo_kekri_encrypt(const CMS_ContentInfo *cms,
         goto err;
     }
 
+    /* 8 byte prefix for AES wrap ciphers */
     wkey = OPENSSL_malloc(ec->keylen + 8);
     if (wkey == NULL) {
         CMSerr(CMS_F_CMS_RECIPIENTINFO_KEKRI_ENCRYPT, ERR_R_MALLOC_FAILURE);
@@ -683,6 +684,7 @@ static int cms_RecipientInfo_kekri_encrypt(const CMS_ContentInfo *cms,
         goto err;
     }
 
+    EVP_CIPHER_CTX_set_flags(ctx, EVP_CIPHER_CTX_FLAG_WRAP_ALLOW);
     if (!EVP_EncryptInit_ex(ctx, cipher, NULL, kekri->key, NULL)
             || !EVP_EncryptUpdate(ctx, wkey, &wkeylen, ec->key, ec->keylen)
             || !EVP_EncryptFinal_ex(ctx, wkey + wkeylen, &outlen)) {
@@ -690,6 +692,10 @@ static int cms_RecipientInfo_kekri_encrypt(const CMS_ContentInfo *cms,
         goto err;
     }
     wkeylen += outlen;
+    if (!ossl_assert((size_t)wkeylen == ec->keylen + 8)) {
+        CMSerr(CMS_F_CMS_RECIPIENTINFO_KEKRI_ENCRYPT, CMS_R_WRAP_ERROR);
+        goto err;
+    }
 
     ASN1_STRING_set0(kekri->encryptedKey, wkey, wkeylen);
 
--- a/crypto/cms/cms_env.c
+++ b/crypto/cms/cms_env.c
@@ -13,7 +13,7 @@
 #include <openssl/x509v3.h>
 #include <openssl/err.h>
 #include <openssl/cms.h>
-#include <openssl/aes.h>
+#include <openssl/evp.h>
 #include "cms_local.h"
 #include "crypto/asn1.h"
 #include "crypto/evp.h"
@@ -625,6 +625,23 @@ int CMS_RecipientInfo_set0_key(CMS_RecipientInfo *ri,
     return 1;
 }
 
+static const EVP_CIPHER *cms_get_key_wrap_cipher(size_t keylen)
+{
+    switch(keylen) {
+    case 16:
+        return EVP_aes_128_wrap();
+
+    case 24:
+        return EVP_aes_192_wrap();
+
+    case 32:
+        return EVP_aes_256_wrap();
+    }
+
+    return NULL;
+}
+
+
 /* Encrypt content key in KEK recipient info */
 
 static int cms_RecipientInfo_kekri_encrypt(const CMS_ContentInfo *cms,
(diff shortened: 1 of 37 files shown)


levitte:
`make update` needed

mattcaswell:
> make update needed

Unfortunately much more than that is needed. This PR just doesn't work at all with "no-deprecated". I'm working on it.

mattcaswell:
Temporarily I moved this into WIP because its currently broken. I've pushed a bunch of fixes for "no-deprecated" but there's still more to fix.

levitte:
The last commit added OPENSSL_SUPRESS_DEPRECATION a little all over the place.  Wouldn't the following be much simpler and easier to follow?

```
diff --git a/include/openssl/macros.h b/include/openssl/macros.h
index a38387f131..484d76c1e1 100644
--- a/include/openssl/macros.h
+++ b/include/openssl/macros.h
@@ -25,6 +25,9 @@
 
 /*
  * Generic deprecation macro
+ *
+ * If OPENSSL_SUPPRESS_DEPRECATED is defined, then DECLARE_DEPRECATED
+ * becomes a no-op
  */
 # ifndef DECLARE_DEPRECATED
 #  define DECLARE_DEPRECATED(f)   f;
@@ -43,6 +46,13 @@
 #  endif
 # endif
 
+/*
+ * If OPENSSL_SUPPRESS_DEPRECATED is defined, then turn off 'no-deprecated'
+ */
+#ifdef OPENSSL_SUPPRESS_DEPRECATED
+# undef OPENSSL_NO_DEPRECATED
+#endif
+
 /*
  * Applications should use -DOPENSSL_API_COMPAT=<version> to suppress the
  * declarations of functions deprecated in or before <version>.  If this is
```

mattcaswell:
You mean instead of the other changes in macros.h? Yes that could work.

levitte:
> You mean instead of the other changes in macros.h? Yes that could work.

And elsewhere.  Basically, you shouldn't need to check OPENSSL_SUPPRESS_DEPRECATE all over the place

mattcaswell:
Updated so that it should now work in both normal builds and no-deprecated builds. I've left this as WIP for now because it now incorporates the commits from #10606 and 10608. Once those are merged I will rebase this and take it out of WIP.

levitte:
I wouldn't mind if #10608 got reviewed. :wink:

mattcaswell:
Now that  #10606 and #10608 have been merged I've rebased this and taken it out of WIP.

Please review.

slontis:
travis isnt looking very healthly,
since deprecated warnings results in errors

crypto/cms/cms_env.c
```diff
@@ -625,56 +625,80 @@ int CMS_RecipientInfo_set0_key(CMS_RecipientInfo *ri,
...
-        CMSerr(CMS_F_CMS_RECIPIENTINFO_KEKRI_ENCRYPT,
-               CMS_R_ERROR_SETTING_KEY);
+    cipher = cms_get_key_wrap_cipher(kekri->keylen);
+    if (cipher == NULL) {
+        CMSerr(CMS_F_CMS_RECIPIENTINFO_KEKRI_ENCRYPT, CMS_R_INVALID_KEY_LENGTH);
         goto err;
     }
 
     wkey = OPENSSL_malloc(ec->keylen + 8);
-
```
slontis: The + 8 on the line above in the existing code looks a bit dangerous to me :)
Hopefully it equals wkeylen below.
A check might have been nice.

include/internal/deprecated.h
```diff
@@ -0,0 +1,29 @@
...
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
+ */
+
+/*
+ * This header file should be included by internal code that needs to use APIs
+ * that have been deprecated for public use, but where those symbols will still
+ * be available internally. For example the EVP code need to use low level APIs
+ * that are otherwise deprecated.
```
levitte: Not to mention our provider code

levitte:
> travis isnt looking very healthly,
> since deprecated warnings results in errors

`crypto/aes/aes_cbc.c` needs to include `"internsl/deprecated.h`

crypto/cms/cms_env.c
mattcaswell: I added a comment about this, and later `ossl_assert` check

include/internal/deprecated.h
mattcaswell: I added that to this sentence

mattcaswell:
> crypto/aes/aes_cbc.c needs to include "internsl/deprecated.h"

Yes, and a couple of other spots. Now fixed.

Please take a look.

mattcaswell:

Ping?

include/crypto/ciphermode_platform.h
```diff
@@ -12,7 +12,8 @@
 
 # include "openssl/aes.h"
 
-# ifdef VPAES_ASM
+# ifndef OPENSSL_NO_DEPRECATED
```
slontis: Not quite sure I understand why the NO_DEPRECTAED is required in this internal file?
(This file is used also inside provider implementations).

mattcaswell: This file is included by e_camellia.c which does not include the "internal/deprecated.h" file, because it doesn't use any externally deprecated APIs. However this header file contains references to AES things that may not be defined in a no-deprecated build - and so the compiler complains loudly. Adding this guard means e_camellia.c compiles correctly. Another fix would be to instead include "internal/deprecated.h" in e_camellia.c - but I'd rather not do that.

levitte: cipher_platform.h should really be divided into algo specific headers.

mattcaswell: Maybe. But I'd see that as outside of the scope of this PR.

levitte: #10662

mattcaswell: Now that #10662 has gone in, this file no longer exists and these changes aren't necessary.

mattcaswell:
I've rebased this to pick up the changes from #10662. Please take another look.

mattcaswell:
Ping?

include/internal/deprecated.h
```diff
@@ -0,0 +1,29 @@
...
+/*
+ * This header file should be included by internal code that needs to use APIs
+ * that have been deprecated for public use, but where those symbols will still
+ * be available internally. For example the EVP and provider code needs to use
+ * low level APIs that are otherwise deprecated.
+ *
+ * This header *must* be the first OpenSSL header included by a source file.
+ */
+
+#ifndef OSSL_INTERNAL_DEPRECTATED_H
```
t8m: Typo: deprectated

mattcaswell: Fixed.

mattcaswell:
Fixup commit pushed addressing latest feedback comment.

mattcaswell:
Pushed. Thanks.

engines/e_padlock.c
```diff
@@ -7,6 +7,12 @@
  * https://www.openssl.org/source/license.html
  */
 
+/*
+ * This file uses the low level AES functions (which are deprecated for
+ * non-internal use) in order to implement the padlock engine AES ciphers.
+ */
+#define OPENSSL_SUPPRESS_DEPRECATED
```
paulidale: @mattcaswell does this work for an engine?
Won't the low level calls be hidden and non-resolvable?

mattcaswell: There are two scenarios:

1) Compilation when "no-deprecated" has been defined

2) Compilation when "no-deprecated" has not been defined

In the first case this PR changed Configure so that padlock never gets compiled. In the second case the low level calls are resolvable and we are only concerned about the deprecation warnings that the compiler will emit. This line is to suppress those.

paulidale: Thanks for the explanation, I missed the Configure change somehow.

Is this how we want to deal with engines that use deprecated APIs?  e.g. the e_dasync engine uses SHA-1 and I modified it to not over in #10791.

mattcaswell: In general no. I think where possible we should convert usage to use EVP instead. I took the decision not to do that with padlock because the usage was so extensive.

PR #10586, opened 2019-12-07
Make Windows build more robust for the ONECORE variant

haohui:
I am trying to build OpenSSL on Windows and have discovered two issues that prevent a successful build of the ONECORE variant:

1. The makefile fails to execute the binaries (e.g., `CC` / `CPP`) if there are spaces in the path.
2. The makefile misses the checks of whether the files exist when calling `copy.pl`.

This PR addresses the two issues above and improve the robustness of the build system.

### DIFF
--- a/Configurations/windows-makefile.tmpl
+++ b/Configurations/windows-makefile.tmpl
@@ -194,23 +194,22 @@ CPPFLAGS={- our $cppflags1 = join(" ",
                                   (map { " -I ".$_} @{$config{CPPINCLUDES}}),
                                   @{$config{CPPFLAGS}}) -}
 CFLAGS={- join(' ', @{$config{CFLAGS}}) -}
-LD={- $config{LD} -}
+LD="{- $config{LD} -}"
 LDFLAGS={- join(' ', @{$config{LDFLAGS}}) -}
 EX_LIBS={- join(' ', @{$config{LDLIBS}}) -}
 
 PERL={- $config{PERL} -}
 
-AR={- $config{AR} -}
+AR="{- $config{AR} -}"
 ARFLAGS= {- join(' ', @{$config{ARFLAGS}}) -}
 
-MT={- $config{MT} -}
+MT="{- $config{MT} -}"
 MTFLAGS= {- join(' ', @{$config{MTFLAGS}}) -}
 
-AS={- $config{AS} -}
+AS="{- $config{AS} -}"
 ASFLAGS={- join(' ', @{$config{ASFLAGS}}) -}
 
 RC="{- $config{RC} -}"
-RC={- $config{RC} -}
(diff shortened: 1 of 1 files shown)


Configurations/windows-makefile.tmpl
```diff
@@ -209,6 +209,7 @@ MTFLAGS= {- join(' ', @{$config{MTFLAGS}}) -}
 AS={- $config{AS} -}
 ASFLAGS={- join(' ', @{$config{ASFLAGS}}) -}
 
+RC="{- $config{RC} -}"
 RC={- $config{RC} -}
```
levitte: Shouldn't you remove the old RC?

haohui: Made a mistake during merging. It has been addressed in the newer version.

haohui:
Just wondering, any plan to merge this PR?

levitte:
We need another review, not just mine.  @openssl/committers?

(I hadn't updated the labels, done now)

levitte:
So FYI, @haohui, we apply a 24h grace period after the `approval: done` label has been set.  In other words, unless some reviewer has a different opinion, this gets merged tomorrow, same time as now.

Configurations/windows-makefile.tmpl
```diff
@@ -499,12 +500,16 @@ install_runtime_libs: build_libs
 install_programs: install_runtime_libs build_programs
 	@if "$(INSTALLTOP)"=="" ( $(ECHO) "INSTALLTOP should not be empty" & exit 1 )
 	@$(ECHO) "*** Installing runtime programs"
-	@"$(PERL)" "$(SRCDIR)\util\mkdir-p.pl" "$(INSTALLTOP)\bin"
-	@"$(PERL)" "$(SRCDIR)\util\copy.pl" $(INSTALL_PROGRAMS) \
+	@if not "$(INSTALL_PROGRAMS)"=="" \
```
mspncp: Just one more question, for my understanding: does the ONECORE build generate only libraries and no binaries? Or asking differently: why and when is  `$(INSTALL_PROGRAMS)` empty?

haohui: It’s orthogonal. No binaries will be generated when you specify no-apps or any other options imply that

levitte: With some configuration settings, "apps" is disabled, which results in an empty `$(INSTALL_PROGRAMS)`, so this actually clears a possible issue that has nothing per se to do with ONECORE....  except if you build for ARM, for which we mandate `no-stdio`, which in turn disables "apps"

mspncp: Thanks.

levitte:
Merged.

b5756085078f43c0a6c9a22ba6e87bc073515780 Make Windows build more robust

PR #10588, opened 2019-12-08
add extern "c" into sm2.h, sm3.h and sm4.h to make it can be called by c++ successfully

cyjseagull:


paulidale:
@mspncp is correct, these header files are internal.  If you are including them, you shouldn't.  Use the EVP APIs instead.

cyjseagull:
> @cyjseagull these are internal headers and not meant to be included by external applications or libraries. AFAIK, none of the internal headers has C++ guards.

In some cases, we need calls sm2_do_verify or sm2_do_sign directly to support Chinese-Crypto-Suites, I think it's better to expose these interfaces to high-level applications.

paulidale:
The header files in include/crypto are not intended for third part use, they are internal to OpenSSL and will change.  They will never be updated so they can be #incuded by applications.

If there is some functionality missing, it will need to be added elsewhere.
Can you be more specific about why Chinese crypto needs this so we've got an idea what is missing?

@mattcaswell, give the "suites" mention it's possibly TLS related.

cyjseagull:
> The header files in include/crypto are not intended for third part use, they are internal to OpenSSL and will change. They will never be updated so they can be #incuded by applications.
> 
> If there is some functionality missing, it will need to be added elsewhere.
> Can you be more specific about why Chinese crypto needs this so we've got an idea what is missing?
> 
> @mattcaswell, give the "suites" mention it's possibly TLS related.

You are right, it maybe need to be placed in elsewhere.
The other cases are that we need sign, verify message using sm2 and calculate hash use sm3, or encrypt/decrypt message use sm4.

cyjseagull:
> The header files in include/crypto are not intended for third part use, they are internal to OpenSSL and will change. They will never be updated so they can be #incuded by applications.
> 
> If there is some functionality missing, it will need to be added elsewhere.
> Can you be more specific about why Chinese crypto needs this so we've got an idea what is missing?
> 
> @mattcaswell, give the "suites" mention it's possibly TLS related.

By the way, is the performance of openssl-1.1 higher than openssl-1.0.2 ?

beldmit:
> The other cases are that we need sign, verify message using sm2 and calculate hash using sm3, or encrypt/decrypt message using sm4. It will convenient for us if you expose it to the app direclty.

EVP interface should fit these purposes. If not, please let us know what's the exact problem.

mattcaswell:
> The other cases are that we need sign, verify message using sm2 and calculate hash using sm3, or encrypt/decrypt message using sm4. It will convenient for us if you expose it to the app direclty.

All of these things should be possible already using the EVP APIs.

paulidale:
Raw performance should be as good or better than 1.0.2, however there are additional overheads for setting up and tearing down algorithms.  This means the actual performance will vary depending on the specific application.

We'll need a lot more specific details (ideally, example code) to understand why the existing EVP APIs are not capable of doing what you require.

Issue #5050, opened 2018-01-09
Revisit and modernize ciphersuite specification mechanism ("cipher string")


kaduk:
There is pent-up demand for a more flexible solution than the current cipherstring setup, most notably for equal-preference grouping (#541) and a more general solution than ```SSL_OP_PRIORITIZE_CHACHA``` (#4436).

Rather than focusing on a specific mechanism ("bracketed equal-preference" or "prioritize chacha"), let's try to step back and think about what is actually needed (including whether ```@STRENGTH``` still makes sense, since that is not easily compatible with the bracketed equal-preference scheme) and how we can fill that need.

However, this is likely to be a big revamp and might be backwards incompatible, so let's mark this as for 1.2.0/post-1.1.1 to start.

blblack:
`SSL_OP_PRIORITIZE_CHACHA` and various `ForMobile` sorts of thinking are short-term and myopic takes on all related things at best.  Why are we now pushing off to design discussion headed for 1.2.0 when the core issues here were brought up 2 years ago in #541 and neglected?

As this ticket rightly notes, it's not about a specific implementation.  It's about expressing the logical capabilities that are desirable from the server administrator's point of view.  Those capabilities happen to map well to the equal-preference grouping proposal, which is probably why that solution was chosen elsewhere as well.  It's ok to come up with a better design that still captures all the desirable properties, but I think it's a shame to reject it without proposing a superior alternative in all this time.

What's logically desirable to the server administrator is the ability to express a policy that server preference comes first, but that within certain sub-lists, the client's preferences should be honored.  In other words, the ability to say things like: "I'm enforcing a server policy preference that `[AB]` is preferable to `[CD]` is preferable to `E`, but I'd rather leave it up to client policy choices which specific cipher to use within either bracketed set".  This is already desirable in TLSv1.2, most-notably due to the ChaPoly preference issues.  It will also be desirable in TLSv1.3 for all the same reasons plus unknown future ones.

I feel like a broken record here, but I'll re-iterate again:

1) TLSv1.3 won't always have just the 3 cipher choices it has today.
2) TLSv1.3's ciphers won't always have the property that they're all widely considered very loosely equal in pragmatic strength by many.
3) Some administrators don't view any of the current 3 choices to be equivalent, either, and some of that thinking depends on your organization's policy stances about different forms of risk.
[...]

Issue #541, opened 2016-01-11
Support bracketed equal-preference groups in SSL[_CTX]_set_cipher_list


rcombs:
BoringSSL implemented this in https://boringssl.googlesource.com/boringssl/+/858a88daf27975f67d9f63e18f95645be2886bfb
This allows a server consumer to specify multiple cipher suites at the same level of preference, so the server can enable `SSL_OP_CIPHER_SERVER_PREFERENCE`, but also take the client's preferences into account when selecting one of several ciphers that are all equally preferred.
The upshot, as described in that BoringSSL commit message, is that the server can prefer AES-GCM and ChaCha20 equally. Android devices without hardware AES support put ChaCha20 first in their suite preference list, so with this feature, the server could select AES-GCM for clients with hardware support and ChaCha20 for ones without it, which optimizes performance for both cases.

shigeki:
I also would like to have this feature. Firefox will prioritize ChaCha20/Poly1305 ciphers over AES-GCM for ARM builds like Chrome in https://bugzilla.mozilla.org/show_bug.cgi?id=1126830 . 
With the equal-preference group, we can make the most use of client performance for both AES-NI and ARM devices, not giving up using server preference of ciphers.

richsalz:
I wanted it too, but we ran out of time. :(

zh99998:
+1 for this request

blblack:
@richsalz: I haven't tested the commit above thoroughly (or added new testsuite tests, either) yet, so there's a little more yet to go here.  Is there any chance this can still make 1.1.0 or is it far too late for something like this?  It's kind of a big pragmatic issue with using the new chacha20-poly1305 stuff in the real world (where only the client knows whether chapoly or aes-gcm is the better perf choice among AEADs).

richsalz:
For a variety of reasons, it is not going to be in 1.1.

blblack:
Planning a better overall mechanism maybe?  The one in BoringSSL is fairly limited in that it doesn't allowing nesting of ordered sets within un-ordered preferences, etc.  It **would** be nice to do something more-general in terms of expressiveness, but probably a lot more invasive as well.

richsalz:
I think the whole CIPHER syntax needs to be reviewed, especially with TLS 1.3 requirements.

MichalStaruch:
[...]

PR #10590, opened 2019-12-08
Improvement of the ciphersuite selection

OttoHollmann:
I had read some issues (#5050, #541, ...) about the current ciphersuite specification and I decided to implement some improvements:
[https://github.com/OttoHollmann/openssl/tree/Improvement-of-the-ciphersuite-selection](https://github.com/OttoHollmann/openssl/tree/Improvement-of-the-ciphersuite-selection)

- Operator `|` for specifying version of each ciphersuite/alias
	There are these versions: `SSLv3`, `TLSv1`, `TLSv1.2`, `TLSv1.3`, `ALL` and multiple versions can be used at one time: `CHACHA20|TLSv1.2|TLSv1.3`

- Parameter for setting default version (`-version_mask`). If no version operator `|` or version alias (e.g. `TLSv1.2`) or specific ciphersuite is set, the default version mask will be applied. If no `version_mask` is provided, it will be set to all versions up to TLSv1.2 (for backwards compatibility).

  e.g. `TLSv1.3:AES128|ALL:ECDH+AES|TLSv1.2|TLSv1:DH` with no `version_mask` will enable:
   - `TLSv1.3` - all TLSv1.3 ciphers (because this alias represent version)
   - `AES128|ALL` - all ciphersuites using `AES128` from all versions
   - `ECDH+AES|TLSv1.2|TLSv1` - `ECDH+AES` from versions TLSv1.2 or TLS1
   - `DH` - `DH` from versions except TLSv1.3 (because of default value of version mask)

- Allowing to set TLSv1.3 ciphersuites using cipher list interface.
If at least one valid TLSv1.3 ciphersuite is in the cipher list, or parameter `version_mask` is used, all ciphers entered using ciphersuites interface will be erased (ciphersuites will be set to empty string).

- Add operator `^` for moving cipher at the top of the list (opposite of `+`)

- Operator `*` for prioritizing cipher depending on client's list. If server ciphers are prioritized and a cipher has this flag and it is at the top of the client list, it will be selected.

  e.g. (suppose only TLSv1.3 ciphersuites):
   - server: `TLSv1.3:+TLS_CHACHA20_POLY1305_SHA256:*CHACHA20|ALL`
      - `TLS_AES_256_GCM_SHA384`
      - `TLS_AES_128_GCM_SHA256`
      - `TLS_AES_128_CCM_8_SHA256`
      - `TLS_AES_128_CCM_SHA256`
      - `*TLS_CHACHA20_POLY1305_SHA256`
   - client: `DEFAULT|TLSv1.3:^TLS_CHACHA20_POLY1305_SHA256`
      - `TLS_CHACHA20_POLY1305_SHA256`
      - `TLS_AES_256_GCM_SHA384`
      - `TLS_AES_128_GCM_SHA256`

  Will result into selecting `TLS_CHACHA20_POLY1305_SHA256` because it's at the top of the client's list and has priority flag in the server list.
  This feature was modified to check first **common** ciphersuite in client list instead of first ciphersuites.

- Equal-preference groups - group is defined by these brackets `[` `]` and can contain ciphersuites of cipher alias (even with specified version)
	- operators `+^` can not be used inside of group
	- operator `*` used inside of group will affect at most all ciphers in current group
	- operators `+^*` used outside of group will not affect ciphersuites inside group
	- operators `@SECLEVEL` and `@STRENGTH` can not be used with groups
  e.g. command `openssl ciphers -v -version_mask ALL '[TLSv1.3:-CHACHA20]:AES256-CCM:CHACHA20:*CHACHA20'` returns this (encryption column omitted):
```
┍   TLS_AES_256_GCM_SHA384         TLSv1.3 Kx=any      Au=any
│   TLS_AES_128_GCM_SHA256         TLSv1.3 Kx=any      Au=any
│   TLS_AES_128_CCM_8_SHA256       TLSv1.3 Kx=any      Au=any
┕   TLS_AES_128_CCM_SHA256         TLSv1.3 Kx=any      Au=any
    AES256-CCM                     TLSv1.2 Kx=RSA      Au=RSA
  * TLS_CHACHA20_POLY1305_SHA256   TLSv1.3 Kx=any      Au=any
  * ECDHE-ECDSA-CHACHA20-POLY1305  TLSv1.2 Kx=ECDH     Au=ECDSA
  * ECDHE-RSA-CHACHA20-POLY1305    TLSv1.2 Kx=ECDH     Au=RSA
  * DHE-RSA-CHACHA20-POLY1305      TLSv1.2 Kx=DH       Au=RSA
  * RSA-PSK-CHACHA20-POLY1305      TLSv1.2 Kx=RSAPSK   Au=RSA
  * DHE-PSK-CHACHA20-POLY1305      TLSv1.2 Kx=DHEPSK   Au=PSK
  * ECDHE-PSK-CHACHA20-POLY1305    TLSv1.2 Kx=ECDHEPSK Au=PSK
  * PSK-CHACHA20-POLY1305          TLSv1.2 Kx=PSK      Au=PSK
  ```

- ~Documentation and tests are missing but will be added soon.~

Any feedback will be greatly appreciated.

### CODE REVIEW

File: ssl/ssl_ciph.c [hunk 1]
```diff
@@ -1622,25 +1900,90 @@ STACK_OF(SSL_CIPHER) *ssl_create_cipher_list(const SSL_METHOD *ssl_method,
...
+    SSL_CIPHER_FLAGS *cipher_list_flags = SSL_get_ciphers_flags(s);
+    uint8_t *flags;
+    char *group = " ";
+    char prefer = ' ';
+    static const char *format = "%s %c ";
+
+    if (buf == NULL) {
+        len = 8;
+        if ((buf = OPENSSL_malloc(len)) == NULL) {
+            SSLerr(SSL_F_SSL_CIPHER_DESCRIPTION, ERR_R_MALLOC_FAILURE);
```
levitte: This function code is detected as a mismatch. However, we do not use function codes any more (upcoming in 3.0), so the simplest thing you can do is have a `0` instead.

levitte:
I've only done a very quick first read, will return for more in the week to come

ssl/ssl_conf.c
```diff
@@ -94,6 +94,10 @@ struct ssl_conf_ctx_st {
     int *min_version;
     /* Pointer to SSL or SSL_CTX max_version field or NULL if none */
     int *max_version;
+    /* Default version mask for cipher_list if none provided, will be set to
+     * all versions up to TLSv1.2
```
kaduk: I didn't read the patch enough to know exactly what this means, but it sounds like it's either making the default only TLS 1.3 or disallowing TLS 1.3 by default, both of which sound weird.

ssl/ssl_local.h
```diff
@@ -737,9 +744,25 @@ typedef struct ssl_ctx_ext_secure_st {
...
+
+struct ssl_cipher_list_flags_st {
+    STACK_OF(SSL_CIPHER) *cipher_list;
+    uint8_t *flags;
+};
+
 struct ssl_ctx_st {
     const SSL_METHOD *method;
-    STACK_OF(SSL_CIPHER) *cipher_list;
+    SSL_CIPHER_FLAGS *cipher_list_flags;
```
kaduk: I think it's kind of a shame if we have to make the new thing have a long name like this.  Changing the name in some way is probably a good idea, to make breakage more obvious, but perhaps something like ```ciphf_list``` could work?

ssl/ssl_ciph.c
OttoHollmann: Fixed, replaced with `0`

ssl/ssl_ciph.c
```diff
@@ -950,25 +1011,102 @@ static int ssl_cipher_strength_sort(CIPHER_ORDER **head_p,
...
-                                  tail_p);
+            ssl_cipher_apply_rule(0, 0, 0, 0, 0, 0, 0, 0, CIPHER_ORD, i,
+                                  0, head_p, tail_p);
 
     OPENSSL_free(number_uses);
     return 1;
 }
 
+/*
+ * Process version string and save it to the bitvise version mask.
```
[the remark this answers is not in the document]
OttoHollmann: Fixed

ssl/ssl_local.h
OttoHollmann: Ok, it's good idea. Done

ssl/ssl_conf.c
OttoHollmann: Yes, it sounds weird, but let me explain it, it's for backwards compatibility. Suppose following configuration:
`MaxProtocol = TLSv1.3`
`CipherString = AES`

This will automatically enable all 3 defaults ciphersuites from TLSv1.3 (like in current version).

When using these improved cipher list with version mask set to `ALL`, the same configuration in TLSv1.3 will enable only ciphersuites using AES. That means, 4 ciphersuites using AES will be enabled and ChaCha20 will be disabled. Because if at least one valid TLSv1.3 ciphersuite is in the `cipherlist`, or parameter `version_mask` is used, parameter `ciphersuites` will be set to empty string (all ciphers from them will be  erased)

Yesterday i pushed documentation, so you can read more information about that. These improvements should'n replace `ciphersuites` interface (or should?), but provide alternative way how to specify TLSv1.3 ciphersuites. 

So if we agree on, that we don't need backwards compatibility, i can make default all versions or e.g. only  TLS 1.2 and TLS 1.3.
Or will be better some option for seting default values at compile time?

ssl/ssl_ciph.c
```diff
@@ -1521,12 +1778,12 @@ STACK_OF(SSL_CIPHER) *ssl_create_cipher_list(const SSL_METHOD *ssl_method,
...
      */
-    ssl_cipher_apply_rule(0, 0, 0, 0, 0, TLS1_2_VERSION, 0, CIPHER_BUMP, -1,
-                          &head, &tail);
+    ssl_cipher_apply_rule(0, 0, 0, 0, 0, TLS1_2_VERSION, 0, 0, CIPHER_BUMP, -1,
+                          0, &head, &tail);
 
     /*
      * Irrespective of strength, enforce the following order:
-     * (EC)DHE + AEAD > (EC)DHE > rest of AEAD > rest.
+     * TLSv1.3 > (EC)DHE + AEAD > (EC)DHE > rest of AEAD > rest.
```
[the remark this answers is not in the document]
OttoHollmann: Ok, removed. I know that, but I just want to point out, that these at the top of the list.

doc/man3/SSL_CTX_set_cipher_list.pod
```diff
@@ -100,12 +116,16 @@ cipher list defines are deprecated as of 3.0.
...
-SSL_CTX_set_cipher_list() and SSL_set_cipher_list() return 1 if any cipher
-could be selected and 0 on complete failure.
+SSL_CTX_set_cipher_list(), SSL_set_cipher_list(),
+SSL_CTX_set_cipher_list_and_mask() and SSL_set_cipher_list_and_mask()
+return 1 if any cipher could be selected and 0 on complete failure.
 
 SSL_CTX_set_ciphersuites() and SSL_set_ciphersuites() return 1 if the requested
 ciphersuite list was configured, and 0 otherwise.
 
+OPENSSL_version_list() returns number of characters that were suscessfully
```
FdaSilvaYY: typo : suscessfully

doc/man3/SSL_CTX_set_cipher_list.pod
```diff
@@ -100,12 +116,16 @@ cipher list defines are deprecated as of 3.0.
...
+procesed and 0 when no versin name was recognized.
```
FdaSilvaYY: typo : versin

OttoHollmann: Fixed

doc/man3/SSL_CTX_set_cipher_list.pod
OttoHollmann: Fixed

ssl/ssl_lib.c
```diff
@@ -575,6 +575,63 @@ static void clear_ciphers(SSL *s)
     ssl_clear_hash_ctx(&s->write_hash);
 }
 
+SSL_CIPHER_FLAGS *ssl_ciphf_list_new(STACK_OF(SSL_CIPHER) *cipher_list)
+{
+    SSL_CIPHER_FLAGS *ret;
+    size_t cnt;
+
+    ret = OPENSSL_malloc(sizeof(SSL_CIPHER_FLAGS));
```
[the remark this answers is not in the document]
OttoHollmann: Fixed even second occurrence.

ssl/ssl_conf.c
```diff
@@ -253,11 +257,19 @@ static int cmd_ECDHParameters(SSL_CONF_CTX *cctx, const char *value)
 static int cmd_CipherString(SSL_CONF_CTX *cctx, const char *value)
 {
     int rv = 1;
-
-    if (cctx->ctx)
-        rv = SSL_CTX_set_cipher_list(cctx->ctx, value);
-    if (cctx->ssl)
-        rv = SSL_set_cipher_list(cctx->ssl, value);
+    SSL_CONF_CTX_set_flags(cctx, SSL_CONF_FLAG_CIPHERSTRING);
```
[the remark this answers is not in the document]
OttoHollmann: Fixed

ssl/s3_lib.c
```diff
@@ -3258,17 +3251,30 @@ long ssl3_default_timeout(void)
...
+
 int ssl3_num_ciphers(void)
 {
     return SSL3_NUM_CIPHERS;
 }
 
 const SSL_CIPHER *ssl3_get_cipher(unsigned int u)
 {
-    if (u < SSL3_NUM_CIPHERS)
+    if (u < SSL3_NUM_CIPHERS) {
```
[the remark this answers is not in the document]
OttoHollmann: Fixed, I had more lines inside these brackets, but then I decided to simplify it and forget to remove the brackets.

OttoHollmann:
Can we move forward please? Which changes are needed to accept this PR? I am looking forward to your response!

mattcaswell:
I have to admit to being quite sceptical about this PR. Do we really need to make ciphersuite selection *more* complicated? It's already quite complicated and I'm not sure adding more operators and ways to do things particularly helps most users.

I'm not at all convinced that version based selection is a good idea. People are already very confused by what the versions associated with ciphersuites actually mean and try to control the negotiated versions via ciphersuites.

I'm also not convinced by the proposal to specify TLSv1.3 ciphersuites via the old style cipher list. We had many discussions about how to do TLSv1.3 ciphersuite configuration during the 1.1.1 development phase and many different proposals were made (and more than one implemented in the code and then backed out again). Having decided on an approach and pushing it out, I'm unconvinced that changing tack now is a good plan.

richsalz:
Merging the version into the list of ciphers makes this very very complicated. Understanding what happens to the version flags (or OP settings) and their interaction with the cipher list is tricky and hard to understand.

This shouldn't be accepted.

OttoHollmann:
Yes, I would agree, that the ciphersuite selection process is quite complicated - this was the main reason which prompted the idea behind this PR.

I didn't know about previous discussions and attempts. But I noticed, that ciphersuites were  entered using old cipher list interface in past. Because it can break some existing configuration, ciphersuites interface was created.

The main goal of this PR was not introducing additional way of doing things but to introduce an unified way. I know a few people who would appreciate this an unified interface.

And just a side note: I don't know any other crypto library which has two interfaces for specifying ciphers. This results into an issue with third party applications which have to implement a second interface otherwise you are unable to modify ciphersuites (e.g. nginx). I believe this makes users much more confused.

We don't have to accept this PR, but I'll be very happy, if we at least come up with a unified interface before closing this PR.

paulidale:
Does anyone object to this being closed?

PR #10594, opened 2019-12-09
Add support for otherName:NAIRealm in output

Janfred:
The experimental [RFC 7585](https://tools.ietf.org/html/rfc7585#section-2.2) adds a new otherName. This commit adds support for displaying this information.

I'm not completely sure if this is a trivial change.


richsalz:
It's possible to make a case that this is a trivial (for copyright purposes) because you added one line to objects.txt, and six lines to v3_alt.c.  And both of those are obvious, and the only way to do what you did.

Janfred:
@beldmit Unfortunately I'm not quite familiar with openssl. Is there a way to verify the output of the GENERAL_NAME_print function?

beldmit:
If you have a valid certificate containing otherName:NAIRealm, ```openssl x509 -text -in mycert.pem``` does the trick.

Janfred:
@beldmit Currently I'm using `openssl req -noout -text -in ~/tmp/cert.csr`. It seems that `i2v_GENERAL_NAME` does the output there.
I'm currently struggling with creating an actual certificate with this attribute, I'm currently using only the CSR for testing.

Janfred:
@beldemit @paulidale I have no problem with signing the CLA, so I'll just do this.

paulidale:
@Janfred, thanks.

levitte:
This is ready to be merged

beldmit:
Will do today but a bit later.

beldmit:
Squashed and merged. Many thanks!

PR #10601, opened 2019-12-10
test/namemap_internal_test.c: use "cookie" instead of "foo"...

### LINKED ISSUE #10401: internal_namemap.t fails
From https://travis-ci.org/openssl/openssl/jobs/609686570?utm_medium=notification&utm_source=github_status
```
03-test_internal_namemap.t ......... 1/1 
1..1
    # RAND SEED 1573316976
    # Subtest: /home/travis/build/openssl/openssl/test/namemap_internal_test
    1..4
    ok 1 - test_digestbyname
    ok 2 - test_cipherbyname
    ok 3 - test_namemap_independent
    # ERROR: (int) 'false1 == 0' failed @ test/namemap_internal_test.c:39
    # [1] compared to [0]
    not ok 4 - test_namemap_stored
/home/travis/build/openssl/openssl/util/shlib_wrap.sh /home/travis/build/openssl/openssl/test/namemap_internal_test => 1
not ok 1 - running namemap_internal_test
#   Failed test 'running namemap_internal_test'
#   at /home/travis/build/openssl/openssl/test/../util/perl/OpenSSL/Test/Simple.pm line 77.
03-test_internal_namemap.t ......... Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/1 subtests 
```

levitte:
... in test_namemap()

Because tests may sometimes run in random order (subject of the
environment variable OPENSSL_TEST_RAND_ORDER being defined), and we're
dealing with the global namemap, each test must use names that are
globally unique for that test.  Unfortunately, we used "foo" in two of
them, which might lead to surprising results.

Fixes #10401

### DIFF
--- a/test/namemap_internal_test.c
+++ b/test/namemap_internal_test.c
@@ -26,7 +26,7 @@ static int test_namemap(OSSL_NAMEMAP *nm)
     int check2 = ossl_namemap_name2num(nm, NAME2);
     int check3 = ossl_namemap_name2num(nm, ALIAS1);
     int check4 = ossl_namemap_name2num(nm, ALIAS1_UC);
-    int false1 = ossl_namemap_name2num(nm, "foo");
+    int false1 = ossl_namemap_name2num(nm, "cookie");
 
     return TEST_int_ne(num1, 0)
         && TEST_int_ne(num2, 0)


richsalz:
So this just kicks the can down the road, right?  Perhaps a test-function that does the equivalent of ```$(basename $(mktemp))``` to generate a random name when needed?

levitte:
Yeah, an individual OPENSSL_CTX might be the answer.

t8m (approved):
I think the fix is ok as is although using something more random would perhaps make the "kick of the can" being much farther.

levitte:
Merged.

97ba39547d4c79b60131475a1512cc7d7e8952b2 test/namemap_internal_test.c: use "cookie" instead of "foo"...

PR #10602, opened 2019-12-11
Cleanup legacy digest methods.

slontis:
Macros have been added to generate the simple legacy methods.
Engines and EVP_MD_METH_get methods still require access to the old legacy methods,
so they needed to be added back in.
They may only be removed after engines are deprecated and removed.
Removed some unnecessary #includes and #ifndef guards (which are done in build.info instead).

##### Checklist


romen:
Should the legacy methods be deprecated as well in favor of EVP?
(Maybe as a separate PR if it requires some discussion and votes)

paulidale:
Yes, they should be.  There will have to be a big deprecation pass before 3.0 is released.

mattcaswell:
> There will have to be a big deprecation pass before 3.0 is released.

I already started this with some of the low level APIs (see #10580). Unfortunately I've run into some problems with that PR.

slontis:
Merged to master.. Thanks.

PR #10606, opened 2019-12-11
PROV: Move AES_CCM specialisation away from common cipher header

levitte:
The AES_CCM specialisation was defined in the common cipher header
providers/implementations/include/prov/ciphercommon_ccm.h, when it
should in fact be in a local providers/implementations/ciphers/
header.


levitte:
This should make @mattcaswell's life slightly easier in #10580

levitte:
I've added the same fix for GCM

slontis (approved):
Fair enough..

levitte:
Merged.

e85f3a14f8be2e1ee2bb362b64e24ac6d0cb8b2f PROV: Move AES_CCM specialisation away from common cipher header
dc7aa029007da3849a04291a4c258587228daac8 PROV: Move AES_GCM specialisation away from common cipher header

PR #10607, opened 2019-12-11
Check return value after loading config file

### LINKED ISSUE #10288: No checking return value of app_load_config() and app_load_config_bio() in apps/req.c
Found this with 1.1.1d.
apps/req.c doesn't check return value of app_load_config() and app_load_config_bio().
It should `goto end;` if req_conf or addext_conf is NULL.
If we provide invalid value for -addext, it shows "req: Error on line 1 of config input", but continues processing like below.

```
$ openssl version
OpenSSL 1.1.1d  10 Sep 2019
$
$ openssl req -new  \
> -addext "subjectAltName := test.com" \
> -newkey rsa:2048 -nodes -keyout key.pem -out req.pem
req: Error on line 1 of config inputGenerating a RSA private key
..........+++++
......................+++++
writing new private key to 'key.pem'
-----
You are about to be asked to enter information that will be incorporated
into your certificate request.
What you are about to enter is what is called a Distinguished Name or a DN.
There are quite a few fields but you can leave some blank
For some fields there will be a default value,
If you enter '.', the field will be left blank.
-----
Country Name (2 letter code) [AU]:
```

kinichiro:
Fixes #10288

### DIFF
--- a/apps/req.c
+++ b/apps/req.c
@@ -476,12 +476,14 @@ int req_main(int argc, char **argv)
 
     if (verbose)
         BIO_printf(bio_err, "Using configuration from %s\n", template);
-    req_conf = app_load_config(template);
+    if ((req_conf = app_load_config(template)) == NULL)
+        goto end;
     if (addext_bio) {
         if (verbose)
             BIO_printf(bio_err,
                        "Using additional configuration from command line\n");
-        addext_conf = app_load_config_bio(addext_bio, NULL);
+        if ((addext_conf = app_load_config_bio(addext_bio, NULL)) == NULL)
+            goto end;
     }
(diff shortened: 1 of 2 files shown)


InfoHunter:
The original issue #10288 mentioned this was found in 1.1.1d, so I think this also should be back ported to 1.1.1 branch.

paulidale:
I'm okay with 1.1.1 too.

paulidale:
Merged to master and 1.1.1, thanks.

paulidale:
This change is trivial.

InfoHunter:
Yes, I agree this change is trivial.

PR #10608, opened 2019-12-11
Add better support for using deprecated symbols internally

levitte:
OPENSSL_SUPPRESS_DEPRECATED only does half the job, in telling the
deprecation macros not to add the warning attribute.  However, with
'no-deprecated', the symbols are still removed entirely, while we
might still want to use them internally.

The solution is to permit <openssl/opensslconf.h> macros to be
modified internally, such as undefining OPENSSL_NO_DEPRECATED in this
case.

However, with the way <openssl/opensslconf.h> includes
<openssl/macros.h>, that's easier said than done.  That's solved by
generating <openssl/configuration.h> instead, and add a new
<openssl/opensslconf.h> that includes <openssl/configuration.h> as
well as <openssl/macros.h>, thus allowing to replace an inclusion of
<openssl/opensslconf.h> with this:

    #undef OPENSSL_NO_DEPRECATED
    #define OPENSSL_SUPPRESS_DEPRECATED

    #include <openssl/macros.h>

Or simply add the following prior to any other openssl inclusion:

    #undef OPENSSL_NO_DEPRECATED
    #define OPENSSL_SUPPRESS_DEPRECATED

Note that undefining OPENSSL_NO_DEPRECATED must never be done by
applications, since the symbols must still be exported by the
library.  Internal test programs are excempt of this rule, though.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -10,7 +10,7 @@
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
   *) Removed include/openssl/opensslconf.h.in and replaced it with
-     include/openssl/opensslconf.h.in, which differs in not including
+     include/openssl/configuration.h.in, which differs in not including
      <openssl/macros.h>.  A short header include/openssl/opensslconf.h
      was added to include both.
 
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,26 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) Removed include/openssl/opensslconf.h.in and replaced it with
+     include/openssl/opensslconf.h.in, which differs in not including
+     <openssl/macros.h>.  A short header include/openssl/opensslconf.h
+     was added to include both.
+
+     This allows internal hacks where one might need to modify the set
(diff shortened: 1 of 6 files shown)


levitte:
This should make @mattcaswell's life slightly easier in #10580

t-j-h (approved):

CHANGES
```diff
@@ -9,6 +9,26 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) Removed include/openssl/opensslconf.h.in and replaced it with
+     include/openssl/opensslconf.h.in, which differs in not including
```
t8m: s/opensslconf.h.in/configuration.h.in/

t8m (changes requested):

CHANGES

levitte:
Does @t-j-h's approval hold through the tiny change to CHANGES?  If not, someone needs to approve/re-approve

t8m (approved):
I can provide the approval as well.

paulidale (approved):

levitte:
Merged.

46994f71631922565924e3ca6303950c36337b33 Add better support for using deprecated symbols internally

PR #10616, opened 2019-12-12
fix ecp_nistz256_avx2_mul_g bugs in ecp_nistz256.c

d5c5ceb0:
`make test` faild when enable ECP_NISTZ256_AVX2.
get such recipes:

15-test_ec.t                     (Wstat: 256 Tests: 5 Failed: 1)
  Failed test:  2
  Non-zero exit status: 1

the reason is that there is no check whether the points are at
infinity.

Signed-off-by: ZhangWei <d5c5ceb0@gmail.com>

##### Checklist

### DIFF
--- a/crypto/ec/ecp_nistz256.c
+++ b/crypto/ec/ecp_nistz256.c
@@ -942,6 +942,29 @@ static void booth_recode_w7(unsigned char *sign,
     *digit = d;
 }
 
+int avx2_is_at_infinity(BN_ULONG *Ax4, int idx)
+{
+    BN_ULONG infty = 0;
+    int i;
+    for (i=0; i<4*9*2; i++) {
+		BN_ULONG mask = 0 - (i%4 != idx);
+		mask = ~mask;
+		infty |= Ax4[i]&mask;
+    }
+    return is_zero(infty);
+}
+
+void avx2_set0(BN_ULONG *Ax4, BN_ULONG infty, int idx)
+{
+	int i = 0;
+
+	infty = ~(-infty);
+	for (i=0; i<4*9; i++) {
+		BN_ULONG temp = infty&Ax4[i];
+		Ax4[i] = (i%4 == idx)? temp: Ax4[i];
+	}
+}
+
 /*
  * ecp_nistz256_avx2_mul_g performs multiplication by G, using only the
  * precomputed table. It does 4 affine point additions in parallel,
@@ -1003,6 +1026,11 @@ static void ecp_nistz256_avx2_mul_g(P256_POINT *r,
     ecp_nistz256_avx2_to_mont(&aX4[4 * 9], &aX4[4 * 9]);
     ecp_nistz256_avx2_set1(&aX4[4 * 9 * 2]);
 
+    avx2_set0(&aX4[4 * 9 * 2], avx2_is_at_infinity(aX4, 0), 0);
+    avx2_set0(&aX4[4 * 9 * 2], avx2_is_at_infinity(aX4, 1), 1);
+    avx2_set0(&aX4[4 * 9 * 2], avx2_is_at_infinity(aX4, 2), 2);
+    avx2_set0(&aX4[4 * 9 * 2], avx2_is_at_infinity(aX4, 3), 3);
+
     wvalue = *((u16 *) & p_str[(idx - 1) / 8]);
     wvalue = (wvalue >> ((idx - 1) % 8)) & mask;
     idx += window_size;
@@ -1037,6 +1065,12 @@ static void ecp_nistz256_avx2_mul_g(P256_POINT *r,
     ecp_nistz256_avx2_to_mont(&bX4[4 * 9], &bX4[4 * 9]);
     /* Optimized when both inputs are affine */
     ecp_nistz256_avx2_point_add_affines_x4(aX4, aX4, bX4);
+
+    //check whether aX4 is at infinity
+    avx2_set0(&aX4[4 * 9 * 2], avx2_is_at_infinity(aX4, 0), 0);
+    avx2_set0(&aX4[4 * 9 * 2], avx2_is_at_infinity(aX4, 1), 1);
+    avx2_set0(&aX4[4 * 9 * 2], avx2_is_at_infinity(aX4, 2), 2);
+    avx2_set0(&aX4[4 * 9 * 2], avx2_is_at_infinity(aX4, 3), 3);
 
     for (i = 2; i < 9; i++) {
         wvalue = *((u16 *) & p_str[(idx - 1) / 8]);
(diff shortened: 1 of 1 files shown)


slontis:
@romen : I dont know if this matters or not - but the normal else case goes to great lengths to remove any ifs (constant time), and checks for 32 bit also i.e: 

                infty = (p.p.X[0] | p.p.X[1] | p.p.X[2] | p.p.X[3] |
                         p.p.Y[0] | p.p.Y[1] | p.p.Y[2] | p.p.Y[3]);
                if (P256_LIMBS == 8)
                    infty |= (p.p.X[4] | p.p.X[5] | p.p.X[6] | p.p.X[7] |
                              p.p.Y[4] | p.p.Y[5] | p.p.Y[6] | p.p.Y[7]);

                infty = 0 - is_zero(infty);
                infty = ~infty;

                p.p.Z[0] = ONE[0] & infty;
                p.p.Z[1] = ONE[1] & infty;
                p.p.Z[2] = ONE[2] & infty;
                p.p.Z[3] = ONE[3] & infty;
                if (P256_LIMBS == 8) {
                    p.p.Z[4] = ONE[4] & infty;
                    p.p.Z[5] = ONE[5] & infty;
                    p.p.Z[6] = ONE[6] & infty;
                    p.p.Z[7] = ONE[7] & infty;
                }

romen:
Yeah, I need to think more about introducing all those new if branches, to understand what kind of information an attacker would potentially recover.

But I am a bit surprised there is a functional problem in that piece of code that went undetected for so long. 

@d5c5ceb0 can you explain more in detail how to reproduce the failure? How are you building?
Also, please notice that to include your contributions into the project we need you to file a CLA form as per the instructions at https://www.openssl.org/policies/cla.html

romen:
@bbbrumley you know the history of this code way better than me and can most likely spot if there is some problem sooner than I can. Do you have some cycles left to spare to give a look at this?

d5c5ceb0:
> @romen : I dont know if this matters or not - but the normal else case goes to great lengths to remove any ifs (constant time), and checks for 32 bit also i.e:
> 
> ```
>             infty = (p.p.X[0] | p.p.X[1] | p.p.X[2] | p.p.X[3] |
>                      p.p.Y[0] | p.p.Y[1] | p.p.Y[2] | p.p.Y[3]);
>             if (P256_LIMBS == 8)
>                 infty |= (p.p.X[4] | p.p.X[5] | p.p.X[6] | p.p.X[7] |
>                           p.p.Y[4] | p.p.Y[5] | p.p.Y[6] | p.p.Y[7]);
> 
>             infty = 0 - is_zero(infty);
>             infty = ~infty;
> 
>             p.p.Z[0] = ONE[0] & infty;
>             p.p.Z[1] = ONE[1] & infty;
>             p.p.Z[2] = ONE[2] & infty;
>             p.p.Z[3] = ONE[3] & infty;
>             if (P256_LIMBS == 8) {
>                 p.p.Z[4] = ONE[4] & infty;
>                 p.p.Z[5] = ONE[5] & infty;
>                 p.p.Z[6] = ONE[6] & infty;
>                 p.p.Z[7] = ONE[7] & infty;
>             }
> ```

These code is in normal branch. and the problem is in avx2 branch.

d5c5ceb0:
> Yeah, I need to think more about introducing all those new if branches, to understand what kind of information an attacker would potentially recover.
> 
> But I am a bit surprised there is a functional problem in that piece of code that went undetected for so long.
> 
> @d5c5ceb0 can you explain more in detail how to reproduce the failure? How are you building?
> Also, please notice that to include your contributions into the project we need you to file a CLA form as per the instructions at https://www.openssl.org/policies/cla.html

OK, I will describe the failure  In detail.

slontis:
> These code is in normal branch. and the problem is in avx2 branch.

Yes - that is correct, I show that code as a reference to illustrate how the code may need to work...
So the questions are:
(1) The normal branch path has code in it to try to avoid 'if' logic (so that it runs in constant time).
Do the changes you propose need to be changed to account for constant time? (i.e: will this new code be affected by timing attacks?)
(2) In the normal code there is a test for 32bit test (if P256_LIMBS == 8). Does your code need to handle the 32 bit case?

bbbrumley:
> But I am a bit surprised there is a functional problem in that piece of code that went undetected for so long.

It is not surprising to me, since I cannot even make it build with `ECP_NISTZ256_AVX2`, let alone get to `make test` ;)

If someone wants to give me instructions to reproduce the bug, I'm happy to take a look.

OTOH, a straight up "unit test fail" is a good indicator exactly zero people are using this code. So a better fix in that case is to just remove it.

d5c5ceb0:
> > These code is in normal branch. and the problem is in avx2 branch.
> 
> Yes - that is correct, I show that code as a reference to illustrate how the code may need to work...
> So the questions are:
> (1) The normal branch path has code in it to try to avoid 'if' logic (so that it runs in constant time).
> Do the changes you propose need to be changed to account for constant time? (i.e: will this new code be affected by timing attacks?)
> (2) In the normal code there is a test for 32bit test (if P256_LIMBS == 8). Does your code need to handle the 32 bit case?

You are right， I did not  consider the side channel attacks. I will fix it according your suggest。

d5c5ceb0:
The steps to reproduce the failure:
1. use the master branch code
2. modify "crypto/ec/build.info" file with
```
  $ECASM_x86_64=ecp_nistz256.c ecp_nistz256-x86_64.s ecp_nistz256-avx2.s x25519-x86_64.s
  $ECDEF_x86_64=ECP_NISTZ256_ASM X25519_ASM ECP_NISTZ256_AVX2
```
3. `./config && make`
4. `make test`

romen:
> OTOH, a straight up "unit test fail" is a good indicator exactly zero people are using this code. So a better fix in that case is to just remove it.

@bbbrumley raises a very good point. It seems like this AVX2 optimized code has been part of the code base for a long time without ever being tested (for functional correctness or for leaks).

While we welcome the fix proposed by @d5c5ceb0, we should consider the alternative of just dropping code that (excluding changes to build.info files) cannot be reached and has very little test coverage.

@mattcaswell what are your thoughts on the matter?

bernd-edlinger:
This comment in ecp_nistz256.c says why this assembler module is always excluded from build: https://github.com/openssl/openssl/blob/e7b834b6bb0a32be694ebc8e614247c9af735c0f/crypto/ec/ecp_nistz256.c#L902-L908
So with more constant time code in place, the performance improvement will be even less
significant.
And moreover the assembler module ecp_nistz256-avx2.pl hasn't any CFI frame infos,
while all other x86_64 assembler modules do have them now.
Therefore, my preference would be to remove that code instead.

romen:
Thanks @d5c5ceb0 for raising awareness on this issue.

I am closing this as we merged the alternative #12019 that simply removes this code for the reasons mentioned in this discussion and on the removal PR.

Thanks everyone for the discussion on this issue!

PR #10620, opened 2019-12-12
chunk 7 of CMP contribution to OpenSSL

Akretsch:
Certificate Management Protocol (CMP, RFC 4210) extension to OpenSSL
Also includes CRMF (RFC 4211) and HTTP transfer (RFC 6712)

CMP and CRMF API is added to libcrypto, and the "cmp" app to the openssl CLI.
Adds extensive man pages and tests. Integration into build scripts.

7th chunk: CMP protection verification
in `crypto/cmp/cmp_vfy.c` and related files. 

This PR includes also some improvements in code formatting (triggered by https://github.com/openssl/openssl/pull/10363) , logging and internal documentation.

##### Checklist


Akretsch:
@mattcaswell, I created this PR starting from the last version of https://github.com/mpeylo/cmpossl/pull/199. If successfully build I would like to ask you for a review.

Akretsch:
The time exceeding travis job is probably not related to us.

Akretsch:
moved some file updates between the commits, the final result is the same.

DDvO:
I've just pushed a new version that completes the cleanup of commits that @Akretsch did 8 hours before.
@mattcaswell, when doing the merge after this PR has been approved, I suggest not squashing these four commits into one:
```
commit 3c550e0fbcd69810de5cf85d71f29c278859ec90 (HEAD -> cmpossl_incremental7,  origin/cmpossl_fixup6_incremental7)
Author: Dr. David von Oheimb <David.von.Oheimb@siemens.com>
Date:   Fri Dec 13 19:45:53 2019 +0100

    chunk 7 of CMP contribution to OpenSSL: CMP message validation and related tests

commit 6d87fc4a1a23014a96364ea6352cf016fbb6967b
Author: Dr. David von Oheimb <David.von.Oheimb@siemens.com>
Date:   Fri Dec 13 20:07:08 2019 +0100

    fix various formatting nits in CMP contribution chunks 1-6 found by the new util/check-format.pl
    correct wording in doc, comments, and parameter names: self-signed -> self-issued where appropriate

commit 241a9a67b007124aa79816b25ab3a11049f074c2
Author: Dr. David von Oheimb <David.von.Oheimb@siemens.com>
Date:   Fri Dec 13 19:50:20 2019 +0100

    improve CMP logging after changes according to comments on CMP chunk7 preview

commit c5e413ee7c2bc69b44c2e3ca3e1972aaf97cebb9
Author: Dr. David von Oheimb <David.von.Oheimb@siemens.com>
Date:   Fri Dec 13 18:54:15 2019 +0100

    add internal doc files actually belonging to CMP contribution chunk 6

```

DDvO:
There is - like after the push before - one strange Travis CI failure, which I cannot reproduce locally:

```
80-test_ocsp.t ..................... ok
80-test_pkcs12.t ................... ok

```

DDvO:
For what ever reason, Travis keeps aborting its very last job on this PR at
```
80-test_ocsp.t ..................... ok
80-test_pkcs12.t ................... ok
The job exceeded the maximum time limit for jobs, and has been terminated.´
```

Does anyone understand why? This does not seem CMP related.

mattcaswell:
It seems to be a frequent and very annoying problem at the moment. Don't know why - but it's definitely not CMP related.

h-vetinari:
It's been plaguing the CI [since](https://travis-ci.org/openssl/openssl/builds/623799966?utm_source=github_status&utm_medium=notification) https://github.com/openssl/openssl/commit/77fedcdc8cf5446354a1e7ff9123d8ada5b79b06, with one of the jobs running into a hard cutoff of 50min by Travis. See #10639 for a fix by @slontis.

DDvO:
Thanks @mattcaswell and @h-vetinari for your (comforting to me) comments!

mattcaswell:
Travis failures seem relevant. Also, already needs a rebase!

Akretsch:
@mattcaswell I fixed the Travis failure and did a rebase.

crypto/cmp/cmp_msg.c
```diff
@@ -334,10 +334,10 @@ OSSL_CMP_MSG *ossl_cmp_certReq_new(OSSL_CMP_CTX *ctx, int type, int err_code)
...
         }
         if ((crm = crm_new(ctx, type, OSSL_CMP_CERTREQID, rkey)) == NULL
-                || !OSSL_CRMF_MSG_create_popo(crm, privkey, ctx->digest,
-                                              ctx->popoMethod)
-                /* value.ir is same for cr and kur */
-                || !sk_OSSL_CRMF_MSG_push(msg->body->value.ir, crm))
+            || !OSSL_CRMF_MSG_create_popo(crm, privkey, ctx->digest,
+                                          ctx->popoMethod)
+            /* value.ir is same for cr and kur */
+            || !sk_OSSL_CRMF_MSG_push(msg->body->value.ir, crm))
```
bernd-edlinger: why this change?

DDvO: As @Akretsch wrote above when introducing this PR, it includes also some improvements in code formatting (triggered by #10363), logging and internal documentation.
There change you are referring to here is part of the commit with message:
```
    fix various formatting nits in CMP contribution chunks 1-6 found by the new util/check-format.pl
    correct wording in doc, comments, and parameter names: self-signed -> self-issued where appropriate
```
and apparently was meant to improve code formatting and silence a finding by the new `check-format.pl` tool.

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,860 @@
...
+ * X509_STORE_set_verify_cb(), such that it gets called by OpenSSL's
+ * verify_cert() function at the end of a cert verification as an opportunity
+ * to gather and queue information regarding a (failing) cert verification,
+ * and to possibly change the result of the verification (not done here).
+ * The CLI also calls it on error while cert status checking using OCSP stapling
+ * via a callback function set with SSL_CTX_set_tlsext_status_cb().
+ *
+ * Returns 0 if and only if the cert verification is considered failed.
+ */
+int OSSL_CMP_print_cert_verify_cb(int ok, X509_STORE_CTX *ctx)
```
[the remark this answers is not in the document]
DDvO: Good point that there is nothing CMP-specific in what this function does. 
Just its implementation has so far been based on some features available only for CMP,
but also those are actually more general, so I moved them as well to where they fit much better:

I've moved OSSL_CMP_print_cert_verify_cb() to crypto/x509/t_x509.c as `X509_STORE_CTX_print_verify_cb()`,
    moving ossl_cmp_add_error_txt() and add_mem_bio_to_error_line() to crypto/err/err_prn.c
    as `ERR_add_error_txt()` and `ERR_add_error_mem_bio()`, respectively,
    moving print_cert() etc. to crypto/x509/t_x509.c as `X509_print_ex_brief()` etc.,
    and moving ossl_cmp_X509_STORE_get1_certs() to crypto/x509/x509_lu.c as `X509_STORE_get1_certs()`.
I've updated the documentation and meta files accordingly.

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,860 @@
...
+        X509 *cert = sk_X509_value(certs, i);
+        if (!ossl_assert(cert != NULL))
+            return 0;
+        if (!cert_acceptable(ctx, "cert from", desc, cert,
+                             already_checked1, already_checked2, msg))
+            continue;
+        n_acceptable_certs++;
+        if (mode_3gpp ? check_msg_valid_cert_3gpp(ctx, cert, msg)
+                      : check_msg_valid_cert(ctx, ctx->trusted, cert, msg)) {
+            /* store successfull sender cert for further msgs in transaction */
```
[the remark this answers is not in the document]
DDvO: fixed

doc/internal/man3/ossl_cmp_certReq_new.pod
```diff
@@ -0,0 +1,193 @@
...
+ #define OSSL_CMP_PKIBODY_PKICONF  19
+ #define OSSL_CMP_PKIBODY_NESTED   20
+ #define OSSL_CMP_PKIBODY_GENM     21
+ #define OSSL_CMP_PKIBODY_GENP     22
+ #define OSSL_CMP_PKIBODY_ERROR    23
+ #define OSSL_CMP_PKIBODY_CERTCONF 24
+ #define OSSL_CMP_PKIBODY_POLLREQ  25
+ #define OSSL_CMP_PKIBODY_POLLREP  26
+
+CMP_rr_new() creates a Revocation Request message from the
```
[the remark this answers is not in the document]
DDvO: Good spotting; fixed.

DDvO:
> @mattcaswell I fixed the Travis failure and did a rebase.

@Akretsch, when adding fixes to this PR please always use `--fixup` to make clear which commit is improved and to make sure that auto-squashing will work fine. I've done this by amending your recent commit (of Dec 19).

crypto/cmp/cmp_msg.c
bernd-edlinger: Revert this.  Since this is a logical expression it is okay to indent the lines beginning with ||

Please see the e-mail discussion on the project mailing list:
https://www.mail-archive.com/openssl-project@openssl.org/msg00164.html

Maybe update your script to at least accept both variants, while new code
should follow the style guide more strictly, I see no need to change everything at once.

crypto/cmp/cmp_local.h
```diff
@@ -699,25 +744,47 @@ int ossl_cmp_asn1_get_int(const ASN1_INTEGER *a);
...
                                     const ASN1_OCTET_STRING *src);
 int ossl_cmp_asn1_octet_string_set1_bytes(ASN1_OCTET_STRING **tgt,
                                           const unsigned char *bytes, int len);
 STACK_OF(X509) *ossl_cmp_build_cert_chain(STACK_OF(X509) *certs, X509 *cert);
 
 /* from cmp_ctx.c */
+int ossl_cmp_print_log(OSSL_CMP_severity level, const OSSL_CMP_CTX *ctx,
+                       const char *func, const char *file, int line,
+                       const char *level_str, const char *format, ...);
+# define OSSL_CMP_log(level, ctx, msg) \
```
bernd-edlinger: is this an internal, thus lower case ossl_cmp_ thing?

crypto/cmp/cmp_msg.c
```diff
@@ -225,14 +225,14 @@ static OSSL_CRMF_MSG *crm_new(OSSL_CMP_CTX *ctx, int bodytype,
...
-                                            subject, ctx->issuer,
-                                            NULL/* serial */))
+        /*
+         * fill certTemplate, corresponding to CertificationRequestInfo
+         * of PKCS#10. The rkey param cannot be NULL so far -
+         * it could be NULL if centralized key creation was supported
+         */
+        || !OSSL_CRMF_CERTTEMPLATE_fill(OSSL_CRMF_MSG_get0_tmpl(crm), rkey,
+                                        subject, ctx->issuer,
+                                        NULL /* serial */))
```
bernd-edlinger: well, can we keep the indentation here as well?

crypto/cmp/cmp_msg.c
DDvO: The mail archive link you provided works only partially for me, but I was already aware of the convention that `&&` and `||` may (but is not required to) have an extra indent. 
Although (for whatever reason) this new convention is still not part of the official coding style https://www.openssl.org/policies/codingstyle.html the tool https://github.com/openssl/openssl/pull/10363 already supports it.

Yet there is a specific problem here: 
between the two lines starting with `||` there is a one-line comment, where it is not clear if the exceptional indentation of the line before/after should be carried over to the comment line. The tool (so far) does not support this potential extra indentation for the comment, and I do not plan to add such a special feature unless the coding guidelines explicitly demand it. 

When @Akretsch prepared the above mentioned commit (which I amended later, that's why you see my name) he apparently took the simple way out and removed the extra indentation for all three lines.
Although his change is fully in line with https://www.openssl.org/policies/codingstyle.html 
I've just reverted this change as you requested.

bernd-edlinger: Regarding the one-line comment, if it is referring to the expression below,
please indent it too.

crypto/cmp/cmp_local.h
DDvO: Good point - done.

crypto/cmp/cmp_msg.c
DDvO: Yep, done. 
Basically the same comment applies as I wrote above, except that here a _multi-line_ comment followed by a line starting with `||` is affected.

bernd-edlinger: Yes. Thanks!

DDvO: BTW, I cannot mark these findings as (at least from my perspective) resolved 
because this PR was opened by @Akretsch, who currently is on vacation.

crypto/cmp/cmp_msg.c
DDvO: We'll try to keep this in mind.

bernd-edlinger: Ok, thanks.  Regarding the coding style document, it looks like
it offers way too much freedom, at least when I do a review I need
to follow a specific style at least for the new code added.  However
I would not use mix the PR with new code and style-fixes on existing
code.  Since it is already some 67 files right now.

DDvO: Sorry for the large number of files this time in the PR.
The actual chunk 7 (CMP verify) would just a few files, but over the last two months various improvements on the earlier chunks have come up, partly related to the pre-review on https://github.com/mpeylo/cmpossl/pull/199.
So we thought it would make sense to do these as separate commits but as a prefix to the commit actually adding the new chunk 7 material, trying to keep up with the timeline https://github.com/mpeylo/cmpossl/wiki#planning-and-current-status-of-the-contribution-chunks where the status is already behind schedule again.

DDvO: Regarding the outdated coding style document I've just filed an issue: #10725.

util/missingcrypto.txt
```diff
@@ -1327,6 +1327,7 @@ X509_policy_tree_level_count(3)
 X509_print(3)
 X509_print_ex(3)
 X509_print_ex_fp(3)
+X509_print_ex_brief(3)
```
bernd-edlinger: Isnt it kind of forbidden to add new functions here?

mattcaswell: Yes! Good catch!

crypto/err/err_prn.c
```diff
@@ -44,6 +45,105 @@ void ERR_print_errors_cb(int (*cb) (const char *str, size_t len, void *u),
...
+}
+
+void ERR_add_error_mem_bio(const char *separator, BIO *bio)
+{
+    if (bio != NULL) {
+        char *str;
+        long len = BIO_get_mem_data(bio, &str);
+
+        if (len > 0) {
+            str[len - 1] = '\0'; /* replace last '\n', terminating str */
```
bernd-edlinger: No. I don't like that at all.

bernd-edlinger: I remember, that I once suggested that you use a BIO where a transcript of the
processing including detailed information about successful operation or possible
errors is returned.  Since the error queue is overloaded by this huge amount of
text processing.

crypto/cmp/cmp_msg.c
DDvO: This specific issue has been resolved /by returning to the original formatting).

util/missingcrypto.txt
DDvO: I was already expecting a comment here. When introducing this auxiliary function (in order to fulfill a structural change request my @mattcaswell) I was considering trying to add documentation for it, but I decided against it because this function is strongly related to the functions around it, all of which are not (yet) documented. So I see it as consequent not (yet) to document it either.

Since we likely cannot wait for the other functions to finally get documented (ideally by their author; others like me would not very be suitable fill the gap), I see two options now - which one do you prefer?
* I try to produce some basic documentation for the new function `X509_print_ex_brief()`
* we un-export the new function - in this case: in which header file could we place its declaration such that it is available to the CMP implementation?

crypto/err/err_prn.c
DDvO: Right, the way of chopping off the trailing separator was unclean. I'd say it was bearable as long as the function was private, but at least when exporting it as `ERR_add_error_mem_bio()` this should have been improved. Done now, moving the feature to `ERR_add_error_txt()` where it fits better.
Doing to I also fixed a bug in `ERR_add_error_txt()` regarding leading separators.

util/missingcrypto.txt
mattcaswell: > we un-export the new function - in this case: in which header file could we place its declaration such that it is available to the CMP implementation?

By this I assume you mean it needs to be available to the cmp command line app? In which case it kind of needs to be in a public header.

Therefore I prefer option (1) (produce some basic documentation)

DDvO: The CLI will use it only indirectly - via the certificate verification print callback, which is has become part of libcrypto.
Nevertheless I can implement (1).

mattcaswell: > The CLI will use it only indirectly - via the certificate verification print callback, which is has become part of libcrypto.

Ah. In that case. I might change my mind. Do we actually need to export this? I like to keep the external API to a minimum.

DDvO: Fine. Then again my question:
in which header file could we place its declaration such that the internal function is usable for all modules of libcrypto?

mattcaswell: It could go in include/crypto/x509.h (Note that it should probably be in lowercase if it is an internal function)

DDvO: Done internalizing ` {X,x}509_print_ex_brief()` .

Akretsch:
@mattcaswell (Back from vacation) I set all the review comments handled by @DDvO to resolved. Do you see further open points in this PR?

crypto/err/err_prn.c
```diff
@@ -44,6 +45,117 @@ void ERR_print_errors_cb(int (*cb) (const char *str, size_t len, void *u),
...
+static void put_error(int lib, const char *func, int reason,
+                      const char *file, int line)
+{
+    ERR_new();
+    ERR_set_debug(file, line, func);
+    ERR_set_error(lib, reason, NULL /* no data here, so fmt is NULL */);
+}
+
+#define TYPICAL_MAX_OUTPUT_BEFORE_DATA 100
+#define MAX_DATA_LEN (ERR_PRINT_BUF_SIZE-TYPICAL_MAX_OUTPUT_BEFORE_DATA)
```
mattcaswell: Nit: spaces around "-"

crypto/err/err_prn.c
```diff
@@ -44,6 +45,117 @@ void ERR_print_errors_cb(int (*cb) (const char *str, size_t len, void *u),
...
+            }
+        } else {
+            while (*next != '\0' && (size_t)(next - txt) <= available_len) {
+                curr = next;
+                next = strstr(curr, separator);
+                if (next != NULL) {
+                    next += strlen(separator);
+                    trailing_separator = *next == '\0';
+                }
+                else {
```
mattcaswell: Nit: else on same line as }

DDvO:
> Aside from a couple of very minor nits I spotted, I'm ok with this.

Great - thanks a lot for finalizing your review!

crypto/err/err_prn.c
DDvO: Oops, if we had run the tool from #10363 also on this source file we would have avoided this.

We'll fix this tomorrow, as well as two indentation nits that the tool found in addition.

crypto/err/err_prn.c
DDvO: The tool would not have found this,
but I'll add a check for this case and for `} else` as well.

DDvO: After adding the mentioned check the tool found a further such occurrence on line 102.

Akretsch:
@bernd-edlinger Also the new comments by Matt have been addressed and resolved.

DDvO:
@bernd-edlinger, 
as your last reviewing activity on this PR was more than two weeks back (on Jan 3rd),
could you please give us an indication when you think you'll be able to continue with this?

bernd-edlinger:
I am extremely busy with work right now, I guess that will be earliest on the coming weekend.

DDvO:
I see, thanks for letting us know!

crypto/cmp/cmp_local.h
```diff
@@ -48,16 +48,21 @@ struct ossl_cmp_ctx_st {
     void *http_cb_arg; /* allows to store optional argument to cb */
 
     /* server authentication */
-    int unprotectedErrors; /* accept neg. response with no/invalid protection */
-                           /* to cope with broken server */
+    int unprotectedErrors; /*-
+                            * accept neg. response with no/invalid protection
+                            * to cope with broken server
+                            */
```
bernd-edlinger: Well, actually that looks clumsy to have a multiline comment hanging over like this.
Could you pleas move those comments before the item, or shorten them a bit?

crypto/cmp/cmp_local.h
```diff
@@ -48,16 +48,21 @@ struct ossl_cmp_ctx_st {
...
     X509_STORE *trusted; /* trust store maybe w CRLs and cert verify callback */
     STACK_OF(X509) *untrusted_certs; /* untrusted (intermediate) certs */
     int ignore_keyusage; /* ignore key usage entry when validating certs */
-    int permitTAInExtraCertsForIR; /* allow use of root certs in extracerts */
-             /* when validating message protection; used for 3GPP-style E.7 */
+    int permitTAInExtraCertsForIR; /*-
+                                    * allow use of root certs in extracerts
+                                    * when validating message protection;
+                                    * used for 3GPP-style E.7
+                                    */
```
bernd-edlinger: likewise here.

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,742 @@
...
+ */
+static int verify_signature(const OSSL_CMP_CTX *cmp_ctx,
+                            const OSSL_CMP_MSG *msg, X509 *cert)
+{
+    EVP_MD_CTX *ctx = NULL;
+    CMP_PROTECTEDPART prot_part;
+    int digest_nid, pk_nid;
+    EVP_MD *digest = NULL;
+    EVP_PKEY *pubkey = NULL;
+    int l;
```
bernd-edlinger: please no l, it is visually ambiguos

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,742 @@
...
+        CMPerr(0, CMP_R_FAILED_EXTRACTING_PUBKEY);
+        goto sig_err;
+    }
+
+    /* create the DER representation of protected part */
+    prot_part.header = msg->header;
+    prot_part.body = msg->body;
+
+    l = i2d_CMP_PROTECTEDPART(&prot_part, &prot_part_der);
+    if (l < 0 || prot_part_der == NULL)
```
bernd-edlinger: here I see if (1 < 0) ;-)

crypto/err/err_prn.c
```diff
@@ -44,6 +45,115 @@ void ERR_print_errors_cb(int (*cb) (const char *str, size_t len, void *u),
...
+}
+
+void ERR_add_error_mem_bio(const char *separator, BIO *bio)
+{
+    if (bio != NULL) {
+        char *str;
+        long len = BIO_get_mem_data(bio, &str);
+
+        if (len > 0)
+            ERR_add_error_txt(separator, str);
```
bernd-edlinger: Sorry, I don't think the pointer here is always null terminated.

crypto/err/err_prn.c
```diff
@@ -44,6 +45,115 @@ void ERR_print_errors_cb(int (*cb) (const char *str, size_t len, void *u),
...
+                OPENSSL_free(tmp);
+            } else {
+                ERR_add_error_data(2, leading_separator, txt);
+            }
+            txt = next; /* finished */
+        }
+    } while (*txt != '\0');
+}
+
+void ERR_add_error_mem_bio(const char *separator, BIO *bio)
```
bernd-edlinger: Actually I am not very convinced that printing text of arbitrary size into a memory bio
and pushing that afterwards to the error queue is a good idea at all.
The error queue will always be limited in size, therefore no guarantee that all the text survives.
I would let the user pass in a BIO where a transcript of the whole process is collected,
and let the user pass it in as a parameter, so it can be printed to the terminal, or discarded
if all works fine, if that is a memory BIO or a file BIO could be the choice of the user.

crypto/err/err_prn.c
Akretsch: Maybe I did not get the point. But its done in a similar way in  t_x509.c:476.

crypto/cmp/cmp_vfy.c
Akretsch: Renamed, thanks.

crypto/cmp/cmp_vfy.c
Akretsch: Renamed, thanks.

crypto/err/err_prn.c
bernd-edlinger: I think if you find another place where BIO_get_mem_data is
used in a context where the returned buffer MUST be zero-terminated
then that is also a bug.

DDvO: @Akretsch, interesting idea to make sure that the input BIO contents are NUL terminated simly by doing
```
BIO_write(bio, "", 1);
```
but please do this only if the BIO is not (yet) NUL-terminated
and add to the documentation of the function that the BIO contents do not need to be NUL-terminated, and if not, this termination is enforced.

crypto/cmp/cmp_local.h
DDvO: @Akretsch, please just move the multi-line comment without merging it with the comment before.

Note that your current solution:
```
/* server authentication
 *
 * accept neg. response with no/invalid protection
 * to cope with broken server
 */
```
also does not adhere to the format of multi-line comments according to the coding guidelines.

crypto/cmp/cmp_local.h
DDvO: @Akretsch, the change you did here is fine.

crypto/err/err_prn.c
DDvO: Well, you are right that also the solution implemented here has its limits (due to the limit of the number of error queue entries, which BTW could be relaxed to some extent at least).

Yet for the intended use case the given implementation is (in all likelihood) sufficient, namely:
In case a certificate verification failed, the new cert verify callback function `X509_STORE_CTX_print_verify_cb()` prints relevant portions not only of the failed cert itself, but - where it is relevant - also the list of untrusted certs and trusted certs contained in the `X509_STORE_CTX`. 
It turned out that occasionally this pretty helpful information is truncated by the 4096 bytes error data limit per error queue entry. With the help of `ERR_add_error_txt()` the cert info output can overflow, if needed, into a second error queue entry, and more than 2*4096 characters are not (reasonably) needed for this use case.

Letting the caller/user optionally pass in a BIO for receiving extended diagnostics would certainly be a more flexible solution, but at the expense of further complication of the cert verify API, 
and I fear that the typical programmer won't spend the effort to provide and handle such a BIO, such that in effect the end user will likely not have the benefit of the extended diagnostic info provided by `X509_STORE_CTX_print_verify_cb()`.

crypto/err/err_prn.c
Akretsch: Fiexed.

crypto/cmp/cmp_local.h
```diff
@@ -48,16 +48,22 @@ struct ossl_cmp_ctx_st {
     void *http_cb_arg; /* allows to store optional argument to cb */
 
     /* server authentication */
-    int unprotectedErrors; /* accept neg. response with no/invalid protection */
-                           /* to cope with broken server */
+    /*
+     * unprotectedErrors may be set accept negative response with no protection
+     * of invalid protection in order to cope with broken server
```
bernd-edlinger: I don’t really understand this sentence

crypto/cmp/cmp_local.h
```diff
@@ -536,67 +542,107 @@ typedef struct ossl_cmp_pkibody_st {
...
+         * rann       [17] RevAnnContent,            --Revocation Ann.
+         */
         OSSL_CMP_REVANNCONTENT *rann; /* 17 */
-        /* crlann     [18] CRLAnnContent,            --CRL Announcement */
-        /* CRLAnnContent ::= SEQUENCE OF CertificateList */
+        /*-
+         * crlann     [18] CRLAnnContent,            --CRL Announcement
+         * CRLAnnContent ::= SEQUENCE OF CertificateList
+         */
         OSSL_CMP_CRLANNCONTENT *crlann;
```
bernd-edlinger: NIT: all but this one and the two values at the end are numbered in the comment.

crypto/cmp/cmp_util.c
```diff
@@ -106,129 +107,69 @@ const char *ossl_cmp_log_parse_metadata(const char *buf,
...
-static void put_error(int lib, const char *func, int reason,
-                      const char *file, int line)
+#define UNKNOWN_FUNC "(unknown function)" /* the default for OPENSSL_FUNC */
+static const char *improve_func_name(const char *func)
 {
-    ERR_new();
-    ERR_set_debug(file, line, func);
-    ERR_set_error(lib, reason, NULL /* no data here, so fmt is NULL */);
+    return func == NULL ? UNKNOWN_FUNC
+                        : strcmp(func, UNKNOWN_FUNC) == 0 ? "" : func;
```
bernd-edlinger: I wonder if you could clarify this with a comment somehow what this function is doing.
And are you sure this is the most simple way to achieve what you want to achieve here?

crypto/cmp/cmp_local.h
DDvO: Oh, I made two critical typos when re-working the sentence. Ok now?

bernd-edlinger: Yes. At least from the grammar POV ;-)

crypto/cmp/cmp_local.h
DDvO: Good spotting - fixed

crypto/cmp/cmp_util.c
DDvO: I've added a little explaining comment.
The problem is that for C90 `__FUNCTION__`  is not officially defined, and `macros.h` then inserts the bulky but useless and slightly misleading text "(unknown function)". The function name would be known, but with pedantic compiler options it is not accessible. In such case I find an empty string more appropriate.

DDvO: I had experimented with various places and ways of replacing that "(unknown function)" string. 
Apart from modifying `include/openssl/macros.h`, (which would not be the topic of this PR) I had not found a simpler way of getting rid of the useless text output.

crypto/err/err_prn.c
Akretsch: Thanks to @DDvO, I would like to resolve this comment. @bernd-edlinger is it OK for your?

crypto/cmp/cmp_util.c
```diff
@@ -106,129 +107,70 @@ const char *ossl_cmp_log_parse_metadata(const char *buf,
...
+        level == OSSL_CMP_LOG_CRIT ? "CRIT" :
+        level == OSSL_CMP_LOG_ERR ? "error" :
+        level == OSSL_CMP_LOG_WARNING ? "warning" :
+        level == OSSL_CMP_LOG_NOTICE ? "NOTE" :
+        level == OSSL_CMP_LOG_INFO ? "info" :
+        level == OSSL_CMP_LOG_DEBUG ? "DEBUG" : "(unknown level)";
+#ifndef NDEBUG
+    const char *comp = improve_func_name(component);
+
+    if (component == NULL || comp[0] == '\0')
```
bernd-edlinger: every time I see this I have to scratch my head.
each of the two places where this improve function is called,
the return value seems to be not right and needs to be fixed up.
How about just telling that function what you want from it, and let it return
the right result ?

crypto/err/err_prn.c
```diff
@@ -44,6 +45,121 @@ void ERR_print_errors_cb(int (*cb) (const char *str, size_t len, void *u),
...
+{
+    if (bio != NULL) {
+        char *str;
+        long len = BIO_get_mem_data(bio, &str);
+
+        if (len > 0) {
+            if (str[len - 1] != '\0') {
+                BIO_write(bio, "", 1);
```
bernd-edlinger: you fail to check the return value of BIO_write.  It can fail,
to allocate that null byte, and in that case the sting is again not null-terminated.

Akretsch:
rebased and review comment fixed

crypto/err/err_prn.c
Akretsch: Thanks, fixed.

crypto/cmp/cmp_util.c
DDvO: Good point. That aux function and its uses had evolved over the time and become inconsequential. I've fixed this.

Akretsch:
@bernd-edlinger We fixed all of your review comments. Is there something more we could do?

DDvO:
I've just rebased this CMP contribution chunk 7.
@bernd-edlinger, is there anything more to do from your perspective?
We should get this PR done soon.

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,742 @@
...
+    if ((csc = X509_STORE_CTX_new()) == NULL
+            || !X509_STORE_CTX_init(csc, trusted_store,
+                                    cert, ctx->untrusted_certs))
+        goto err;
+
+    valid = X509_verify_cert(csc) > 0;
+
+    /* make sure suitable error is queued even if callback did not do */
+    if (!valid && ERR_GET_REASON(ERR_peek_last_error())
+        != CMP_R_POTENTIALLY_INVALID_CERTIFICATE)
```
bernd-edlinger: indentation

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,742 @@
...
+    /* make sure that a matching name is there */
+    if (actual_name == NULL) {
+        ossl_cmp_log1(WARN, ctx, "missing %s", actual_desc);
+        return 0;
+    }
+    if (X509_NAME_cmp(actual_name, expect_name) == 0)
+        return 1;
+
+    actual = X509_NAME_oneline(actual_name, NULL, 0);
+    expect = X509_NAME_oneline(expect_name, NULL, 0);
```
bernd-edlinger: below you check the result for non-zero but not here.

doc/internal/man3/ossl_cmp_msg_create.pod
```diff
@@ -0,0 +1,80 @@
...
+L<OSSL_CMP_CTX_new(3)>, L<OSSL_CMP_exec_IR_ses(3)>,
+L<OSSL_CMP_MSG_http_perform(3)>
+
+=head1 HISTORY
+
+The OpenSSL CMP support was added in OpenSSL 3.0.
+
+=head1 COPYRIGHT
+
+Copyright 2007-2019 The OpenSSL Project Authors. All Rights Reserved.
```
bernd-edlinger: Copyright 2020 ?

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,742 @@
+/*
+ * Copyright 2007-2019 The OpenSSL Project Authors. All Rights Reserved.
```
bernd-edlinger: Copyright 2020 ?

crypto/cmp/cmp_vfy.c
DDvO: I don't see what should be wrong about these lines, and 
also the format checking tool https://github.com/openssl/openssl/pull/10363 is fine with the file,
nevertheless I've beautified the layout using a new aux variable.

crypto/cmp/cmp_vfy.c
DDvO: I've added such checks.

doc/internal/man3/ossl_cmp_msg_create.pod
DDvO: Updated - in the very sense of the word ;-)

crypto/cmp/cmp_vfy.c
DDvO: Updated dates in this file, too.

DDvO:
Thanks @bernd-edlinger for your further comments!
I've already handled them.

crypto/cmp/cmp_vfy.c
bernd-edlinger: I meant something like
```C
if (!valid 
    && ERR_GET_REASON(...)
          != CMP_R_...)
```
or
```C
if (!valid && ERR_GET_REASON(...)
                   != CMP_R_...)
```

bernd-edlinger:
Please have a look at the CI builds

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,743 @@
...
+    int valid = 0;
+
+    /* generate expected protection for the message */
+    if ((protection = ossl_cmp_calc_protection(msg, secret, NULL)) == NULL)
+        return 0;               /* failed to generate protection string! */
+
+    valid = (msg->protection != NULL && msg->protection->length >= 0)
+        && (msg->protection->type == protection->type)
+        && (msg->protection->length == protection->length)
+        && (CRYPTO_memcmp(msg->protection->data, protection->data,
```
bernd-edlinger: I think here are lots of unnecessary parentheses, and I would indent the expression
to line up with the first line

richsalz:
Copyright is automatically patched up when a release is made, I wouldn't worry about copyright dates.  YMMV.

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,743 @@
...
+                || ASN1_OCTET_STRING_cmp(ctx->transactionID,
+                                         msg->header->transactionID) != 0)) {
+        CMPerr(0, CMP_R_TRANSACTIONID_UNMATCHED);
+        return -1;
+    }
+
+    /* compare received nonce with the one we sent */
+    if (ctx->senderNonce != NULL
+            && (msg->header->recipNonce == NULL
+                    || ASN1_OCTET_STRING_cmp(ctx->senderNonce,
```
bernd-edlinger: The indentation is different here and above, you can choose which one you prefer.

doc/internal/man3/ossl_cmp_msg_create.pod
```diff
@@ -0,0 +1,80 @@
...
+  int ossl_cmp_msg_gen_ITAV_push0(OSSL_CMP_MSG *msg, OSSL_CMP_ITAV *itav);
+  int ossl_cmp_msg_gen_ITAVs_push1(OSSL_CMP_MSG *msg,
+                                    STACK_OF(OSSL_CMP_ITAV) *itavs);
+
+=head1 DESCRIPTION
+
+ossl_cmp_bodytype_to_string() returns the name of the given body type as string,
+or "illegal body type" on error.
+
+ossl_cmp_msg_get_bodytype() returns the body type of the given PKIMessage.
```
bernd-edlinger: this does not mention negative return values, but the code seems to have checks for < 0

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,743 @@
...
+        /* 5.1.3.1.  Shared Secret Information */
+    case NID_id_PasswordBasedMAC:
+        if (verify_PBMAC(msg, ctx->secretValue)) {
+            /*
+             * RFC 4210, 5.3.2: 'Note that if the PKI Message Protection is
+             * "shared secret information", then any certificate transported in
+             * the caPubs field may be directly trusted as a root CA
+             * certificate by the initiator.'
+             */
+            switch (ossl_cmp_msg_get_bodytype(msg)) {
```
bernd-edlinger: here is no check for negative rv, and no default label

crypto/cmp/cmp_vfy.c
DDvO: Good spotting; fixed.

crypto/cmp/cmp_vfy.c
DDvO: I've removed the extra parens.
The hanging indentation on subsequent lines of a multi-line variable assignment is automatically performed by Emacs (and also required by the format checker tool):

doc/internal/man3/ossl_cmp_msg_create.pod
DDvO: I've extended the description accordingly.

crypto/cmp/cmp_vfy.c
DDvO: I've added a `default:` case (which just does `break;`)

DDvO:
> Please have a look at the CI builds

Ah, there was a mistake while manually resolving a merge conflict 
in the usual culprit (`util/libcrypto.num`). 
The respective Travis failures should be gone now.

crypto/cmp/cmp_vfy.c
bernd-edlinger: Not really nice this way.

DDvO: At first I was puzzled that GitHub still shows the lines with the parentheses, but now I see the yellow marker "Outdated".

The hanging indentation on assignments and in other cases appears pretty common and I'm used to it.
In this special case with the `&&` at the front of each subsequent line we do have the option to add extra indentation of four spaces, which happens to align things very nicely here :smile: 
Done.

bernd-edlinger: Thanks.

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,745 @@
...
+                if (ctx->trusted != NULL) {
+                    STACK_OF(X509) *certs = msg->body->value.ip->caPubs;
+                    /* value.ip is same for cp, kup, and ccp */
+
+                    if (!ossl_cmp_X509_STORE_add1_certs(ctx->trusted, certs, 0))
+                        /* adds both self-issued and not self-issued certs */
+                        break;
+                }
+            default:
+                break;
```
bernd-edlinger: This is in the inner switch, and the break will continue with "return 1;"
but this is in fact an error?

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,745 @@
...
+            case OSSL_CMP_PKIBODY_CCP:
+                if (ctx->trusted != NULL) {
+                    STACK_OF(X509) *certs = msg->body->value.ip->caPubs;
+                    /* value.ip is same for cp, kup, and ccp */
+
+                    if (!ossl_cmp_X509_STORE_add1_certs(ctx->trusted, certs, 0))
+                        /* adds both self-issued and not self-issued certs */
+                        break;
+                }
+            default:
```
bernd-edlinger: I think this will be a fallthrough warning/error here?

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,745 @@
...
+            case OSSL_CMP_PKIBODY_CP:
+            case OSSL_CMP_PKIBODY_KUP:
+            case OSSL_CMP_PKIBODY_CCP:
+                if (ctx->trusted != NULL) {
+                    STACK_OF(X509) *certs = msg->body->value.ip->caPubs;
+                    /* value.ip is same for cp, kup, and ccp */
+
+                    if (!ossl_cmp_X509_STORE_add1_certs(ctx->trusted, certs, 0))
+                        /* adds both self-issued and not self-issued certs */
+                        break;
```
bernd-edlinger: So same issue as the break below, you should probably just " return 0; " here

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,745 @@
...
+     * RFC 4210 section 5.1.1 states: the recipNonce is copied from
+     * the senderNonce of the previous message in the transaction.
+     * --> Store for setting in next message
+     */
+    if (!ossl_cmp_ctx_set1_recipNonce(ctx, msg->header->senderNonce))
+        return -1;
+
+    /* if not yet present, learn transactionID */
+    if (ctx->transactionID == NULL
+        && !OSSL_CMP_CTX_set1_transactionID(ctx, msg->header->transactionID))
```
bernd-edlinger: I would suggest to indent this line.

crypto/cmp/cmp_vfy.c
DDvO: No, the semantics is fine this way: 
The CMP message field `caPubs` is present only for `IP`, `CP`, `KUP`, and `CCP` message types. Else there is nothing to do.

bernd-edlinger: I mean what happens if ossl_cmp_msg_get_bodytype returns < 0 ?
that is an error condition, right?

crypto/cmp/cmp_vfy.c
DDvO: Well, if we added the optional extra indent at this point, the line would get more than 80 chars long and would have to be split, making it less readable. Therefore I prefer not to do that here.

crypto/cmp/cmp_vfy.c
DDvO: Indeed there is an implicit fall-through here, which is semantically fine, but I've added an explicit `break` for clarity, which has the same effect in this case.

crypto/cmp/cmp_vfy.c
DDvO: Yes, here (in case of malloc failure) a `return 0` was missing - thanks!

crypto/cmp/cmp_vfy.c
bernd-edlinger: I wonder why the CIs did not catch this...

DDvO:
@bernd-edlinger, CI succeeded and all comments resolved.

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,746 @@
...
+        if (!cert_acceptable(ctx, "cert from", desc, cert,
+                             already_checked1, already_checked2, msg))
+            continue;
+        n_acceptable_certs++;
+        if (mode_3gpp ? check_msg_valid_cert_3gpp(ctx, cert, msg)
+                      : check_msg_valid_cert(ctx, ctx->trusted, cert, msg)) {
+            /* store successful sender cert for further msgs in transaction */
+            if (X509_up_ref(cert)
+                    && !ossl_cmp_ctx_set0_validatedSrvCert(ctx, cert))
+                X509_free(cert);
```
bernd-edlinger: shouldn't this be an error condition?

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,746 @@
...
+             * "shared secret information", then any certificate transported in
+             * the caPubs field may be directly trusted as a root CA
+             * certificate by the initiator.'
+             */
+            switch (ossl_cmp_msg_get_bodytype(msg)) {
+            case OSSL_CMP_PKIBODY_IP:
+            case OSSL_CMP_PKIBODY_CP:
+            case OSSL_CMP_PKIBODY_KUP:
+            case OSSL_CMP_PKIBODY_CCP:
+                if (ctx->trusted != NULL) {
```
bernd-edlinger: and when ctx->trusted==NULL, is that also ok to return 1 then?

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,746 @@
...
+                    STACK_OF(X509) *certs = msg->body->value.ip->caPubs;
+                    /* value.ip is same for cp, kup, and ccp */
+
+                    if (!ossl_cmp_X509_STORE_add1_certs(ctx->trusted, certs, 0))
+                        /* adds both self-issued and not self-issued certs */
+                        return 0;
+                    break;
```
bernd-edlinger: to be really clear, I would move this break immediately before the defaut:

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,746 @@
...
+                }
+            default:
```
bernd-edlinger: Please add an explicit case -1: return 0;

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,746 @@
...
+    }
+    return rcvd_type;
+}
+
+int ossl_cmp_verify_popo(const OSSL_CMP_MSG *msg, int accept_RAVerified)
+{
+    if (!ossl_assert(msg != NULL && msg->body != NULL))
+        return 0;
+    switch (msg->body->type) {
+    case OSSL_CMP_PKIBODY_P10CR: {
```
bernd-edlinger: the { should be on a line of its own. or declare req above, and dont need compound statement here.

crypto/cmp/cmp_vfy.c
DDvO: Storing the validated peer cert is actually just a performance optimization, so if this fails we do not strictly need to indicate an error.
Yet since this indicates out-of-memory or a potential internal error I've added `return 0` for such cases.

crypto/cmp/cmp_vfy.c
DDvO: Good point, done.

crypto/cmp/cmp_vfy.c
DDvO: Yes, better be very conservative here, done.

crypto/cmp/cmp_vfy.c
DDvO: So I've moved the `{` to the next line.

Your comment indicates that the format checker #10363 should better complain if there is anything (except comments) after the `:` of a `case` or `default`. 
I'll add this, and also to the draft extension of coding style rules in #10725.

crypto/cmp/cmp_vfy.c
DDvO: Yes, this is fine: we can ignore any `caPubs` contents if there is no need for them, which is indicated by an unset trust store.

DDvO:
@bernd-edlinger, again all (new) comments addressed.

crypto/cmp/cmp_vfy.c
```diff
@@ -0,0 +1,754 @@
...
+    len = i2d_CMP_PROTECTEDPART(&prot_part, &prot_part_der);
+    if (len < 0 || prot_part_der == NULL)
+        goto end;
+    prot_part_der_len = (size_t) len;
+
+    /* verify signature of protected part */
+    if (!OBJ_find_sigid_algs(OBJ_obj2nid(msg->header->protectionAlg->algorithm),
+                             &digest_nid, &pk_nid)
+            || digest_nid == NID_undef || pk_nid == NID_undef
+            || (digest = (EVP_MD *)EVP_get_digestbynid(digest_nid)) == NULL) {
```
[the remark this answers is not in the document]
DDvO: Good question - likely whoever originally coded this did not realize that one can simply add `const` to the variable decl: `const EVP_MD *digest = NULL;` to make the cast superfluous.
Fixed.

DDvO:
Thanks a lot @bernd-edlinger for very swiftly finishing your review 
and the quick re-approval by @mattcaswell!

So looks like I can squash and merge this chunk after the usual one day grace period?

mattcaswell:
> So looks like I can squash and merge this chunk after the usual one day grace period?

Looks like it - assuming no last minute problems from the CIs.

DDvO:
AppVeyor is always pretty slow, and this time it took even some 12 hours to complete (after presumably being busy grinding on other PRs), but finally it confirmed that the builds and tests pass.

DDvO:
I'll squash the commits already now in preparation of the later merge,
which among others should raise the chances that the merge 
will be rewarded by that lovely purple-colored 'merged' icon :purple_heart:

DDvO:
Done squashing (and already adding the reviewer & pr info).
Do I need an explicit 'ready to merge' label before (potentially rebasing again) and pushing?

bernd-edlinger:
We do usually the final squash and rebase without pushing that to the PR again,
but directly to the openssl git server, once the grace period is over, the labels are just
optional, the final decision is human.
Howver now, I have problems to fetch the revision f431d686479a36131bdbbab23f9399c15d066637
from this pr which @mattcaswell originally approved,  so I am unable to double-check that the
squashing / re-basing did not introduce incorrectness.

DDvO:
> We do usually the final squash and rebase without pushing that to the PR again,
> but directly to the openssl git server, once the grace period is over, 

Yes, but @mattcaswell recently asked me regarding PR #10667 to squash it before it was merged, and this had the nice effect that GitHub has been able to recognize the merge automatically, which is indicated by the purple 'Merged' label. I appreciate this, and others as well, and @beldmit even asked to 'incorporate the brilliant changes he made to the workflow' - see #11038

> the labels are just optional, the final decision is human.

Good to know, thanks @bernd-edlinger.

> However now, I have problems to fetch the revision [f431d68](https://github.com/openssl/openssl/commit/f431d686479a36131bdbbab23f9399c15d066637)
> from this pr which @mattcaswell originally approved,

Yes, this indeed is a drawback. One could still follow up the commits via the link you gave and via its parent link etc, but this would be pretty cumbersome.
A workaround would be to save the original/unsquashed PR branch in a backup branch.

> so I am unable to double-check that the squashing / re-basing did not introduce incorrectness.

Before force-pushing I had checked this using 
```
git diff origin/cmpossl_incremental7
```

DDvO:
@Akretsch, if you still have the previous state of `cmpossl_incremental7` locally, please push it as `cmpossl_incremental7-orig` such that we have it as backup/reference just in case.

Akretsch:
cmpossl_incremental7-orig pushed

bernd-edlinger:
yes, that would be nice,
please do `git reset --hard f431d686479a36131bdbbab23f9399c15d066637` before pushing that
since that was the originally approved version.  It should be available in your repo,
but I cannot pull if from github, although I know for sure, that revision is still in their database.

DDvO:
@bernd-edlinger, good idea to reset to the commit in question first.
This did not work our for @Akretsch (since he had fetched last on Feb 3rd), but for me.
I've force-pushed f431d686 to `cmpossl_incremental7-orig` now.

crypto/err/err_prn.c
```diff
@@ -44,6 +45,123 @@ void ERR_print_errors_cb(int (*cb) (const char *str, size_t len, void *u),
...
+    int line;
+    const char *func = NULL;
+    const char *data = NULL;
+    int flags;
+    unsigned long err = ERR_peek_last_error();
+
+    if (separator == NULL)
+        separator = "";
+    if (err == 0)
+        put_error(ERR_LIB_CMP, NULL, 0, "", 0);
```
levitte: CMP specific errors...

crypto/err/err_prn.c
```diff
@@ -44,6 +45,123 @@ void ERR_print_errors_cb(int (*cb) (const char *str, size_t len, void *u),
...
+        if (curr != NULL) {
+            /* split error msg at curr since error data would get too long */
+            if (curr != txt) {
+                tmp = OPENSSL_strndup(txt, curr - txt);
+                if (tmp == NULL)
+                    return;
+                ERR_add_error_data(2, separator, tmp);
+                OPENSSL_free(tmp);
+            }
+            put_error(ERR_LIB_CMP, func, err, file, line);
```
levitte: CMP specific errors

yangyangtiantianlonglong:
@DDvO  Have the CMP feature and CRMF feature  been completed? Are the test cases completed?

FdaSilvaYY:
> 
> 
> @DDvO Have the CMP feature and CRMF feature been completed? Are the test cases completed? When are these features planned for openssl 3.0.0?

@yangyangtiantianlonglong Next chunk is [here](https://github.com/mpeylo/cmpossl/pull/201). A  you may find their roadmap in this repo.

crypto/err/err_prn.c
DDvO: @levitte, the function itself can be more generally useful and belongs in the context of `crypto/err/`, that's why we moved it here. We just forgot to update this lib reference (and the one below) from `ERR_LIB_CMP` to `ERR_LIB_SYS`. We plan to fix this nit along with the next CMP chunk. 
@Akretsch, please include the commit I've just pushed to our `cmp-dev` branch.

crypto/err/err_prn.c
DDvO: see above

DDvO:
@yangyangtiantianlonglong, the state of the CMP contribution is shown and kept up-to-date at https://github.com/mpeylo/cmpossl/wiki#planning-and-current-status-of-the-contribution-chunks.

crypto/err/err_prn.c
DDvO: It turns out that the replacement here better is `ERR_LIB_NONE`.

crypto/err/err_prn.c
DDvO: @levitte, it turned out that the replacement here better is `ERR_GET_LIB(err)`.
@mattcaswell, for simplicity, these small fixes will become part of the PR for CMP chunk 8.

mattcaswell: Ok!

DDvO:
> Where is this callback? I get:
> `test/cmp_vfy_test.c(72): error C2065: 'OSSL_CMP_print_cert_verify_cb': undeclared identifier`

Looks like you had an inconsistent source tree or build.
The callback function is meanwhile called `X509_STORE_CTX_print_verify_cb`.

PR #10626, opened 2019-12-13
Support for KTLS on FreeBSD for TLS 1.3.

bsdjhb:
For TLS 1.3, only the final key used for application data is sent to
the kernel.  The kernel is responsible for framing the record
including placing the record type after the record payload and
appending any padding.

To avoid pulling in more dependencies for the tls13secretstest,
explicitly disable KTLS when building that test.

Note that only support for offloading TX is supported as FreeBSD's
KTLS only supports offloading TX.

### DIFF
--- a/ssl/tls13_enc.c
+++ b/ssl/tls13_enc.c
@@ -516,10 +516,12 @@ int tls13_change_cipher_state(SSL *s, int which)
     const EVP_MD *md = NULL;
     const EVP_CIPHER *cipher = NULL;
 #if !defined(OPENSSL_NO_KTLS) && defined(OPENSSL_KTLS_TLS13)
-# ifndef __FreeBSD__
+# ifdef __FreeBSD__
+    struct tls_enable crypto_info;
+# else
     struct tls_crypto_info_all crypto_info;
-    BIO *bio;
 # endif
+    BIO *bio;
 #endif
 
     if (which & SSL3_CC_READ) {
@@ -784,7 +786,6 @@ int tls13_change_cipher_state(SSL *s, int which)
         s->statem.enc_write_state = ENC_WRITE_STATE_VALID;
 #ifndef OPENSSL_NO_KTLS
 # if defined(OPENSSL_KTLS_TLS13)
-#  ifndef __FreeBSD__
     if (!(which & SSL3_CC_WRITE) || !(which & SSL3_CC_APPLICATION)
         || ((which & SSL3_CC_WRITE) && (s->mode & SSL_MODE_NO_KTLS_TX)))
         goto skip_ktls;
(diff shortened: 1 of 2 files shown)


junjunk:
Hi! There is also a problem with ktls_read_record() used in sock_read(), because it adds TLSv12 header hardcoded. I think that this should be rewritten somehow and after that it will be possible to add support for linux kernel >= 5.3.x

bsdjhb:
This patch only supports TX on FreeBSD in tls13_enc.c.  It does not try to offload RX on either FreeBSD or Linux.  When it grows RX support in the future then the OS-specific parts would need to add suitable checks.  For example, tls1_enc.c's version of this code uses different version checks on Linux (1.2 with AES-GCM128 only) vs FreeBSD (1.0-1.2 with either AES-CBC or AES-GCM).  The current code is also under `#ifdef FreeBSD` so that it won't try to do TX offload of TLS 1.3 on Linux either.  When someone adds the support to Linux, they can then update the KTLS bits in this file similar to what is in tls1_enc.c to add the Linux-specific checks.  However, I don't expect that it will need further changes beyond the OS-specific cipher, etc. checks and using the OS-specific data structure to pass to `BIO_set_ktls`

junjunk:
My bad, you are right in case that you use only TX path in this patch. But it is not so clear because you use checks of RX path too:
```
    if (which & SSL3_CC_WRITE)
        bio = s->wbio;
    else
        bio = s->rbio;
....

/* ktls doesn't support renegotiation */

    if ((BIO_get_ktls_send(s->wbio) && (which & SSL3_CC_WRITE)) ||
        (BIO_get_ktls_recv(s->rbio) && (which & SSL3_CC_READ))) {
        SSLfatal(s, SSL_AD_NO_RENEGOTIATION, SSL_F_TLS1_CHANGE_CIPHER_STATE,
                 ERR_R_INTERNAL_ERROR);
        goto err;
    }

```
Also, I think `SSLfatal` calls should be with `SSL_F_TLS13_CHANGE_CIPHER_STATE` rather then `SSL_F_TLS1_CHANGE_CIPHER_STATE`

Right now I'm trying to implement the same function for Linux and will try to use your's ABI for `derive_secret_key_and_iv`, but I didn't understand why are you not changing the call to `derive_secret_key_and_iv` in `tls13_update_key`?

bsdjhb:
I kept the RX checks mostly because I plan on adding RX support once I add RX support to FreeBSD's KTLS.  You are right about the `SSLfatal` calls, I will fix that.

Hmm, I did change the call in `tls13_update_key` to store the key in a local buffer on the stack to match the different signature, so I'm not sure what your question is there?

junjunk:
> I kept the RX checks mostly because I plan on adding RX support once I add RX support to FreeBSD's KTLS.
Ok, maybe it's better to comment this case, because without this comment the code seems to be unclear
> Hmm, I did change the call in `tls13_update_key` to store the key in a local buffer on the stack to match the different signature, so I'm not sure what your question is there?
I'm sorry, I suppose I have missed that change while looking on the other parts of code.

ssl/tls13_enc.c
```diff
@@ -516,10 +516,12 @@ int tls13_change_cipher_state(SSL *s, int which)
     const EVP_MD *md = NULL;
     const EVP_CIPHER *cipher = NULL;
 #if !defined(OPENSSL_NO_KTLS) && defined(OPENSSL_KTLS_TLS13)
-# ifndef __FreeBSD__
+# ifdef __FreeBSD__
+    struct tls_enable crypto_info;
+# else
     struct tls_crypto_info_all crypto_info;
```
mattcaswell, line 522: We seem to have done this in several places already - and here is another instance. Perhaps it would be better to introduce our own typedef for this - the definition of which can have the `__FreeBSD__` guards around it - and then from then on we just use that typedef without having to special case FreeBSD everywhere.

ssl/tls13_enc.c
```diff
@@ -797,9 +798,27 @@ int tls13_change_cipher_state(SSL *s, int which)
...
+        crypto_info.iv_len = EVP_CIPHER_CTX_iv_length(ciph_ctx);
+        break;
+    default:
+        goto skip_ktls;
+    }
+    crypto_info.cipher_key = key;
+    crypto_info.cipher_key_len = EVP_CIPHER_key_length(cipher);
+    crypto_info.iv = iv;
+    crypto_info.tls_vmajor = (s->version >> 8) & 0x000000ff;
+    crypto_info.tls_vminor = (s->version & 0x000000ff);
```
mattcaswell, line 816: This block seems to do a similar thing to the `ktls_check_supported_cipher` and `ktls_configure_crypto` functions. It seems like it would be better to modify those functions to do this, rather than insert this block here.

bsdjhb: So I've actually posted a new PR #12111 which does do that, but with some caveats as FreeBSD and Linux need some slightly different arguments.  I also raised the question about a typedef in that review as well.  The new PR also includes #11679 since the diffs for both of these ended up much smaller after doing the refactoring so that I think they are ok to review together.  Apologies if that wasn't the best way to do that?

bsdjhb:
Replaced by #12111

PR #10632, opened 2019-12-16
Cleanup _CONF, LHASH, STACK API's

richsalz:
This is a WIP because the no-deprecated build breaks.  I'm doing something wrong in lhash.h I guess.

This deprecates statistics which removes an interlocked increment from *every* HASH lookup, and at least one from *every* hash insert.


levitte:
In any of our .c files where the deprecated stuff is still used, define `OPENSSL_SUPPRESS_DEPRECATED`. That should clear the errors you see.

paulidale:
For lhash, replacement with a more modern algorithm might be prudent.  The lhash algorithm is quite clever with its low cost incremental resizing (both up and down) but it isn't cache friendly and we're hamstrung by the APIs.  For the majority of cases, we don't remove items from our hash tables which defeats one of the big advantages of lhash.  [Hopscotch hashing](https://en.wikipedia.org/wiki/Hopscotch_hashing) would be worth considering as an alternative.

Stack is trying to be a universal data type (effectively C++'s vector, deque, list and stack) and again we're stuck with the existing APIs.  I'd prefer to see multiple types, each tuned for its specific purpose.  I've also a feeling that the _unshift_ call isn't ever used (by us).  These data structures are well known and fairly straightforward to implement.  The hard bit will be going through the OpenSSL code and intelligently replacing each instance.

richsalz:
> In any of our .c files where the deprecated stuff is still used, define `OPENSSL_SUPPRESS_DEPRECATED`. That should clear the errors you see.

That means putting it in lhash.h since the define generates static inline.  That means it's almost everywhere:
```
; g grep lhash.h include/openssl/
include/openssl/conf.h:# include <openssl/lhash.h>
include/openssl/conf_api.h:# include <openssl/lhash.h>
include/openssl/err.h:# include <openssl/lhash.h>
include/openssl/ssl.h:# include <openssl/lhash.h>
include/openssl/txt_db.h:# include <openssl/lhash.h>
include/openssl/x509_vfy.h:# include <openssl/lhash.h>
; 
```
which isn't a good solution.

I'm going to try to not-depecreate the deprecated LH functions and see what happens. That's also not good, but is clearly less bad.

richsalz:
As for https://github.com/openssl/openssl/pull/10632#issuecomment-566393560, I suggest you open an issue to capture this.

richsalz:
I pushed a fixup commit that makes the not no-deprecated build work.  Ready for review.

t8m:
The Travis failure is relevant. You need to mark the deprecated symbols as such in libcrypto.num.

richsalz:
Could you have been looking at an old Travis output?  The only two failures I see are timeouts.

richsalz:
Oh, I see the bug. Updated commit pushed.

richsalz:
Travis failures are a timeout and an apt-get failure and are unrelated.

richsalz:
reseolved missingmacro.txt conflict and pushed.

richsalz:
No longer WIP, ready for review.

richsalz:
I made the ```_CONF_xxx``` API deprecated.  (Nobody's looked at this in weeks so I figure it's safe to make changes. :)  Since it's used internally, I had to base this on #10666 now.

richsalz:
Travis failure is a timeout

richsalz:
Travis failure is a timeout.

richsalz:
Rebased on master since #10753 was merged.

richsalz:
@levitte, it appears from looking at the appveyor build log, https://ci.appveyor.com/project/openssl/openssl/builds/29981668/job/5gs1qv84vw3e35hu, that OPENSSL_BUILDING_OPENSSL is not passed down.  (Look at the end of the log)  Is #10753 correct for windows?

richsalz:
hey @paulidale since you like deprecating things, how about this, which is useful :)

include/openssl/lhash.h
```diff
@@ -86,9 +86,17 @@ void OPENSSL_LH_doall(OPENSSL_LHASH *lh, OPENSSL_LH_DOALL_FUNC func);
...
+ * which is include'd in several other files, so that's not good.
+ * Instead, we just declare these as normal, and the |DEFINE_LASH_OF|
+ * macro includes the type-safe wrapper functions or not, depending on
+ * deprecation.
+ */
 unsigned long OPENSSL_LH_get_down_load(const OPENSSL_LHASH *lh);
 void OPENSSL_LH_set_down_load(OPENSSL_LHASH *lh, unsigned long down_load);
-
 # ifndef OPENSSL_NO_STDIO
 void OPENSSL_LH_stats(const OPENSSL_LHASH *lh, FILE *fp);
```
paulidale: These should be wrapped by the `DEPRECATEDIN_3_0` macro to give users a warning that they are going.
This would then change `util/libcrypto.num`.

Or have I missed something subtle about using them subsequently in the type-safe wrappers?

crypto/lhash/lhash.c
```diff
@@ -197,6 +204,7 @@ static void doall_util_fn(OPENSSL_LHASH *lh, int use_arg,
      * reverse the order so we search from 'top to bottom' We were having
      * memory leaks otherwise
      */
+    down_load = lh->down_load;
```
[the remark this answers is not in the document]
richsalz: aargh!  fixed.

crypto/lhash/lhash.c
```diff
@@ -208,6 +216,7 @@ static void doall_util_fn(OPENSSL_LHASH *lh, int use_arg,
             a = n;
         }
     }
+    lh->down_load = down_load;
```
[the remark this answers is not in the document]
richsalz: good idea.  I put the if test in a macro.

include/openssl/lhash.h
richsalz: Nobody should be using those functions directly, and the type-safe macro marks them correctly (see ```LHASH_OF_TYPE_DEPRECATED```).  Is that sufficient?

richsalz:
fixup commit pushed.

richsalz:
The appveyor failure seems to be something screwy with how this is based on main.  I do not understand it.  Any clues from anyone?

apps/progs.c
```diff
@@ -24,15 +24,6 @@ FUNCTION functions[] = {
...
     {FT_general, "dgst", dgst_main, dgst_options, NULL},
-#if !defined(OPENSSL_NO_DH) && !defined(OPENSSL_NO_DEPRECATED_3_0)
-    {FT_general, "dhparam", dhparam_main, dhparam_options, "pkeyparam"},
-#endif
-#if !defined(OPENSSL_NO_DSA) && !defined(OPENSSL_NO_DEPRECATED_3_0)
-    {FT_general, "dsa", dsa_main, dsa_options, "pkey"},
-#endif
-#if !defined(OPENSSL_NO_DSA) && !defined(OPENSSL_NO_DEPRECATED_3_0)
-    {FT_general, "dsaparam", dsaparam_main, dsaparam_options, "pkeyparam"},
-#endif
```
levitte: Why?

apps/progs.c
```diff
@@ -45,9 +36,6 @@ FUNCTION functions[] = {
 #endif
     {FT_general, "errstr", errstr_main, errstr_options, NULL},
     {FT_general, "fipsinstall", fipsinstall_main, fipsinstall_options, NULL},
-#if !defined(OPENSSL_NO_DSA) && !defined(OPENSSL_NO_DEPRECATED_3_0)
-    {FT_general, "gendsa", gendsa_main, gendsa_options, "genpkey"},
-#endif
```
levitte: Why?

apps/progs.h
```diff
@@ -19,16 +19,12 @@ extern int cms_main(int argc, char *argv[]);
 extern int crl_main(int argc, char *argv[]);
 extern int crl2pkcs7_main(int argc, char *argv[]);
 extern int dgst_main(int argc, char *argv[]);
-extern int dhparam_main(int argc, char *argv[]);
-extern int dsa_main(int argc, char *argv[]);
-extern int dsaparam_main(int argc, char *argv[]);
```
levitte: Why?

apps/progs.h
```diff
@@ -19,16 +19,12 @@ extern int cms_main(int argc, char *argv[]);
...
 extern int ec_main(int argc, char *argv[]);
 extern int ecparam_main(int argc, char *argv[]);
 extern int enc_main(int argc, char *argv[]);
 extern int engine_main(int argc, char *argv[]);
 extern int errstr_main(int argc, char *argv[]);
 extern int fipsinstall_main(int argc, char *argv[]);
-extern int gendsa_main(int argc, char *argv[]);
```
levitte: Why?

apps/progs.h
```diff
@@ -73,16 +69,12 @@ extern const OPTIONS cms_options[];
 extern const OPTIONS crl_options[];
 extern const OPTIONS crl2pkcs7_options[];
 extern const OPTIONS dgst_options[];
-extern const OPTIONS dhparam_options[];
-extern const OPTIONS dsa_options[];
-extern const OPTIONS dsaparam_options[];
```
levitte: Why?

apps/progs.h
```diff
@@ -73,16 +69,12 @@ extern const OPTIONS cms_options[];
...
 extern const OPTIONS ec_options[];
 extern const OPTIONS ecparam_options[];
 extern const OPTIONS enc_options[];
 extern const OPTIONS engine_options[];
 extern const OPTIONS errstr_options[];
 extern const OPTIONS fipsinstall_options[];
-extern const OPTIONS gendsa_options[];
```
levitte: Why?

include/openssl/conf_api.h
```diff
@@ -7,40 +7,4 @@
...
-long _CONF_get_number(const CONF *conf, const char *section,
-                      const char *name);
-
-int _CONF_new_data(CONF *conf);
-void _CONF_free_data(CONF *conf);
-
-#ifdef  __cplusplus
-}
-#endif
-#endif
```
levitte: Something to be noted is that you have not *deprecated* the `_CONF` functions, you have *removed* them from public view.  Considering how so many seem to be quite confused regarding deprecation vs removal, I think you need to be more precise about what this is doing, in commit message in and CHANGES.

util/missingcrypto.txt
```diff
@@ -383,6 +383,8 @@ COMP_expand_block(3)
 COMP_get_name(3)
 COMP_get_type(3)
 COMP_zlib(3)
+_CONF_get_section(3)
+_CONF_get_string(3)
```
levitte: You could document them internally, in `doc/internal/man3`

apps/progs.c
richsalz: my wrapper script did no-deprecated; fixed (I hope).

include/openssl/conf_api.h
richsalz: rewrote the commit message, thanks.

util/missingcrypto.txt
richsalz: Is that "could" or "should"?  They are really a low-level internal implementation detail that should not be documented; the real solution is to cleanup and rewrite NCONF in a future release.

CHANGES
```diff
@@ -230,7 +230,17 @@
      [Andy Polyakov]
 
   *) Most memory-debug features have been deprecated, and the functionality
-     replaced with no-ops.
+     replaced with no-ops. {CRYPTO,OPENSSL}_mem_debug_{push,pop} are now
+     no-ops and have been deprecated.
+     [Rich Salz]
+
+  *) The undocumented _CONF_xxx API has been deprecated. The undocumented
```
levitte: "have been removed from public headers"

crypto/conf/conf_api.c
```diff
@@ -75,144 +34,23 @@ char *_CONF_get_string(const CONF *conf, const char *section,
...
-    sk_CONF_VALUE_free(sk);
-    if (v != NULL)
-        OPENSSL_free(v->section);
-    OPENSSL_free(v);
-    return NULL;
+    vv.section = "default";
+    vv.name = (char *)name;
+    v = lh_CONF_VALUE_retrieve(conf->data, &vv);
+    return v == NULL ? NULL : v->value;
 }
```
levitte: So little remains in this file that I think it can be moved to `conf_def.c` as well, even thought they aren't static...  just to keep the `_CONF` functions together.

crypto/conf/conf_lib.c
```diff
@@ -239,6 +239,18 @@ int NCONF_load_bio(CONF *conf, BIO *bp, long *eline)
     return conf->meth->load_bio(conf, bp, eline);
 }
 
+static STACK_OF(CONF_VALUE) *_CONF_get_section_values(const CONF *conf,
+                                                      const char *section)
```
levitte: This is called in exactly one place, and is very little code.  I see no reason to keep it as a separate function.

You might actually want to consider the same thing for the other `_CONF` functions, i.e. integrate their code in the spots where they are used, unless they are multiple spots.

levitte:
I'm surprised at the lack of changes in `util/libcrypto.num`.  `make update`, perhaps?  That should also clear the new entries in `util/missingcrypto.txt`...

CHANGES
richsalz: fixup commit pushed.

crypto/conf/conf_lib.c
richsalz: I wanted to make as few changes as possible, figuring it would ease the review (hah).  I inlined it.  I don't think there are others, but I will check.  Additional commit pushed.

richsalz:
The two functions ```_CONF_get_section``` and ```_CONF_get_string``` are used in multiple places. The code could be rewritten to not do that, but not for this PR.  All of NCONF should be rewritten and not layered on top of sand on top of mud (CONF and _CONF) as it currently is.

There are few libcrypto changes because the ```_CONF_``` functions were not public, and still aren't.  The two above show up as "public" because they are in a header file that is no longer skipped/special-cased.

levitte:
> The two above [ `_CONF_get_section` and `_CONF_get_string` ] show up as "public" because they are in a header file that is no longer skipped/special-cased.

Huh?????  I'm ssorry, I fail to understand why `util/mknum.pl` or `util/find-doc-nits` are looking at `crypto/conf/conf_local.h`...

crypto/conf/conf_api.c
richsalz: is that a request or just a side-note?

richsalz:
I do not understand your question at https://github.com/openssl/openssl/pull/10632#issuecomment-587070486  The functions *were not* in libcrypto.num and they still aren't.

crypto/conf/conf_api.c
levitte: It's a request

levitte:
So why did they end up in util/missingcrypto.txt?

richsalz:
> So why did they end up in util/missingcrypto.txt?

Now I understand.  And there seems no need for it.  Removed.  Part of the same commit that moved conf_api into conf_def.

richsalz:
Pushed an updated commit to fix "no-deprecated" build.

richsalz:
Rebased after #10669 went in. This deprecates statistics which removes an interlocked increment from every HASH lookup, and at least one from every hash insert.  It also does general code cleanup in this area.

crypto/lhash/lh_stats.c
```diff
@@ -10,19 +10,16 @@
...
 
 #include <openssl/bio.h>
 #include <openssl/lhash.h>
 #include "lhash_local.h"
 
-# ifndef OPENSSL_NO_STDIO
+#ifndef OPENSSL_NO_STDIO
 void OPENSSL_LH_stats(const OPENSSL_LHASH *lh, FILE *fp)
 {
+# ifndef OPENSSL_NO_DEPRECATED_3_0
```
t8m, line 22: It looks wrong to drop just the code of the functions and not the complete functions. If some functionality is deprecated, the related APIs should be dropped completely with no-deprecated build.

include/openssl/lhash.h
t8m: I do not see the DEPRECATEDIN_3_0 in the type safe macros either.

util/missingcrypto.txt
richsalz: I think the proper solution is to rewrite and replace NCONF and remove all the lower-level routines.  I will resolve this comment thread and opened #11662.

crypto/lhash/lh_stats.c
richsalz: There was a comment in lhash.h:
```
 * We would like to deprecate these functions, but doing so gets errors in
 * builds when allowing deprecated, and we'd have to annotate this file
 * which is include'd in several other files, so that's not good.
 * Instead, we just declare these as normal, and the |DEFINE_LASH_OF|
 * macro includes the type-safe wrapper functions or not, depending on
 * deprecation.
```

But I just tried ```config``` and ```config no-deprecated``` and both builds worked so I guess along the way I fixed something.

Fixup commit pushed.

richsalz:
This has been waiting for an OTC respond for over seven weeks.  It needs to go in before the beta release, as it makes API changes.

richsalz:

PR #10635, opened 2019-12-17
Fix unwind info for some trivial functions

bernd-edlinger:
While stack unwinding works with gdb here, the
function _Unwind_Backtrace gives up when something outside
.cfi_startproc/.cfi_endproc is found in the call stack, like
OPENSSL_cleanse, OPENSSL_atomic_add, OPENSSL_rdtsc, CRYPTO_memcmp
and other trivial functions which don't save anything in the stack.

I played around with this
```
diff --git a/apps/speed.c b/apps/speed.c
index e4b104e..6b0cef0 100644
--- a/apps/speed.c
+++ b/apps/speed.c
@@ -8,6 +8,7 @@
  */
 
+#include <unwind.h>
 #undef SECONDS
 #define SECONDS                 3
 #define RSA_SECONDS             10
@@ -225,8 +226,44 @@ static const int aead_lengths_list[] = {
 
 #ifdef SIGALRM
 
+struct ctxt
+{
+  void **res;
+  int cnt;
+  int max;
+};
+
+_Unwind_Reason_Code bt (struct _Unwind_Context *c, void *x)
+{
+  struct ctxt *ct = x;
+  if (ct->cnt < ct->max)
+    ct->res[ct->cnt++] = (void*)_Unwind_GetIP(c);
+  return _URC_NO_REASON;
+}
+
 static void alarmed(int sig)
 {
+    void *res[20];
+    struct ctxt ct;
+    int i;
+    ct.res = res;
+    ct.cnt = 0;
+    ct.max = 20;
+    write(1,"alarmed\n", 8);
+    _Unwind_Backtrace(bt, &ct);
+    for (i = 0; i < ct.cnt; i++)
+    {
+       unsigned long p = (unsigned long) res[i];
+       char q[18];
+       int j = sizeof(q);
+       q[--j] = 10;
+       do {
+          q[--j] = (p & 15) < 10 ? '0' + (p & 15) : 'A' + (p & 15) - 10;
+          p >>= 4;
+       } while(p && j > 0);
+       write(1, q+j, sizeof(q)-j);
+    }
+    write(1,"finally\n", 8);
     signal(SIGALRM, alarmed);
     run = 0;
 }
```

and noticed the stack trace stops in OPENSSL_cleanse, when that is
interrupted by a signal. (as it may happen)

### DIFF
--- a/crypto/x86_64cpuid.pl
+++ b/crypto/x86_64cpuid.pl
@@ -41,6 +41,7 @@ print<<___;
 .type	OPENSSL_atomic_add,\@abi-omnipotent
 .align	16
 OPENSSL_atomic_add:
+.cfi_startproc
 	movl	($arg1),%eax
 .Lspin:	leaq	($arg2,%rax),%r8
 	.byte	0xf0		# lock
@@ -49,16 +50,19 @@ OPENSSL_atomic_add:
 	movl	%r8d,%eax
 	.byte	0x48,0x98	# cltq/cdqe
 	ret
+.cfi_endproc
 .size	OPENSSL_atomic_add,.-OPENSSL_atomic_add
 
 .globl	OPENSSL_rdtsc
 .type	OPENSSL_rdtsc,\@abi-omnipotent
 .align	16
 OPENSSL_rdtsc:
+.cfi_startproc
 	rdtsc
 	shl	\$32,%rdx
 	or	%rdx,%rax
 	ret
+.cfi_endproc
 .size	OPENSSL_rdtsc,.-OPENSSL_rdtsc
 
 .globl	OPENSSL_ia32_cpuid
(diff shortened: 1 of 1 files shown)


richsalz:
What's the overhead of the CFI stuff?

bernd-edlinger:
It is information mapped in the address space.  It is smaller than the debug info.
when you don't need it you may either use -fno-asynchronous-unwind-tables
(I don't know if that removes CFI info also from assembler files, that explicitly include it).
or just remove certain segments using a custom linker script.

kroeckx:
To clarify what @bernd-edlinger said, it does not have an effect on the generated code. The information is stored the .eh_frame and that is an allocated section. I guess in theory we could move it to the .debug_frame section. You might want to read https://stackoverflow.com/questions/26300819/why-gcc-compiled-c-program-needs-eh-frame-section

kroeckx:
As far as I understand, moving it to .debug_frame would involve using `.cfi_sections .debug_frame`. I assume that's what `-fno-asynchronous-unwind-tables` does. But if that's something we want to do, that should go to some other PR.

richsalz:
Thanks for the explanations.

bernd-edlinger:
Merged to master as 8913378 and 1.1.1 as 7540f7b
Thanks!

PR #10636, opened 2019-12-17
[WIP] EVP: Make it possible to have keymgmt name that differs from the algo

levitte:
In some cases, the key has a different type name than the algorithms
that use them.  For example, both ECDH and ECDSA uses EC keys.

To accomodate this, we give the algorithms the possibility to give
back the type name they expect for their keys.

Since the resulting keymgmt MUST be from the same provider, it's up to
the provider authors to make sure the names match.

### DIFF
--- a/crypto/evp/exchange.c
+++ b/crypto/evp/exchange.c
@@ -44,7 +44,7 @@ static void *evp_keyexch_from_dispatch(int name_id,
                                        OSSL_PROVIDER *prov)
 {
     EVP_KEYEXCH *exchange = NULL;
-    int fncnt = 0, paramfncnt = 0;
+    int fncnt = 0, paramfncnt = 0, ctxparamfncnt = 0;
 
     if ((exchange = evp_keyexch_new(prov)) == NULL) {
         ERR_raise(ERR_LIB_EVP, ERR_R_MALLOC_FAILURE);
@@ -93,29 +93,33 @@ static void *evp_keyexch_from_dispatch(int name_id,
             if (exchange->get_params != NULL)
                 break;
             exchange->get_params = OSSL_get_OP_keyexch_get_params(fns);
+            paramfncnt++;
             break;
         case OSSL_FUNC_KEYEXCH_GETTABLE_PARAMS:
             if (exchange->gettable_params != NULL)
                 break;
             exchange->gettable_params
                 = OSSL_get_OP_keyexch_gettable_params(fns);
+            paramfncnt++;
             break;
         case OSSL_FUNC_KEYEXCH_SET_CTX_PARAMS:
             if (exchange->set_ctx_params != NULL)
                 break;
             exchange->set_ctx_params = OSSL_get_OP_keyexch_set_ctx_params(fns);
-            paramfncnt++;
+            ctxparamfncnt++;
             break;
         case OSSL_FUNC_KEYEXCH_SETTABLE_CTX_PARAMS:
             if (exchange->settable_ctx_params != NULL)
                 break;
             exchange->settable_ctx_params
                 = OSSL_get_OP_keyexch_settable_ctx_params(fns);
-            paramfncnt++;
+            ctxparamfncnt++;
             break;
         }
     }
-    if (fncnt != 4 || (paramfncnt != 0 && paramfncnt != 2)) {
+    if (fncnt != 4
+        || (paramfncnt != 0 && paramfncnt != 2)
+        || (ctxparamfncnt != 0 && ctxparamfncnt != 2)) {
         /*
          * In order to be a consistent set of functions we must have at least
          * a complete set of "exchange" functions: init, derive, newctx,
--- a/crypto/evp/exchange.c
+++ b/crypto/evp/exchange.c
@@ -88,6 +88,17 @@ static void *evp_keyexch_from_dispatch(int name_id,
                 break;
             exchange->dupctx = OSSL_get_OP_keyexch_dupctx(fns);
             break;
+        case OSSL_FUNC_KEYEXCH_GET_PARAMS:
+            if (exchange->get_params != NULL)
+                break;
+            exchange->get_params = OSSL_get_OP_keyexch_get_params(fns);
+            break;
+        case OSSL_FUNC_KEYEXCH_GETTABLE_PARAMS:
+            if (exchange->gettable_params != NULL)
+                break;
+            exchange->gettable_params
+                = OSSL_get_OP_keyexch_gettable_params(fns);
+            break;
         case OSSL_FUNC_KEYEXCH_SET_CTX_PARAMS:
             if (exchange->set_ctx_params != NULL)
                 break;
--- a/crypto/evp/exchange.c
+++ b/crypto/evp/exchange.c
@@ -10,6 +10,7 @@
 #include <openssl/crypto.h>
 #include <openssl/evp.h>
 #include <openssl/err.h>
+#include <openssl/core_names.h>
 #include "internal/refcount.h"
 #include "crypto/evp.h"
 #include "internal/provider.h"
@@ -170,6 +171,25 @@ EVP_KEYEXCH *EVP_KEYEXCH_fetch(OPENSSL_CTX *ctx, const char *algorithm,
                              (void (*)(void *))EVP_KEYEXCH_free);
 }
 
+static EVP_KEYMGMT *evp_keyexch_get_keymgmt(EVP_KEYEXCH *exchange,
+                                            int fallback_id,
+                                            OPENSSL_CTX *libctx,
+                                            const char *propquery)
+{
+    char *keytype_name = NULL;
+    OSSL_PARAM params[] = { OSSL_PARAM_END, OSSL_PARAM_END };
+
+    params[0] =
+        OSSL_PARAM_construct_utf8_ptr(OSSL_ALG_PARAM_KEYTYPE,
+                                      &keytype_name, 0);
+    if (exchange->get_params != NULL && !exchange->get_params(params))
+        return NULL;
+    if (keytype_name == NULL) {
+        return evp_keymgmt_fetch_by_number(libctx, fallback_id, propquery);
+    }
+    return EVP_KEYMGMT_fetch(libctx, keytype_name, propquery);
+}
+
 int EVP_PKEY_derive_init(EVP_PKEY_CTX *ctx)
 {
     int ret;
@@ -193,12 +213,10 @@ int EVP_PKEY_derive_init(EVP_PKEY_CTX *ctx)
      * matter, as it isn't tied to a specific EVP_PKEY op.
      */
     exchange = EVP_KEYEXCH_fetch(ctx->libctx, ctx->algorithm, ctx->propquery);
-    if (exchange != NULL && ctx->keymgmt == NULL) {
-        int name_id = EVP_KEYEXCH_number(exchange);
-
+    if (exchange != NULL && ctx->keymgmt == NULL)
         ctx->keymgmt =
-            evp_keymgmt_fetch_by_number(ctx->libctx, name_id, ctx->propquery);
-    }
+            evp_keyexch_get_keymgmt(exchange, EVP_KEYEXCH_number(exchange),
+                                    ctx->libctx, ctx->propquery);
 
     if (ctx->keymgmt == NULL
         || exchange == NULL
--- a/crypto/evp/pmeth_fn.c
+++ b/crypto/evp/pmeth_fn.c
@@ -46,7 +46,8 @@ static void *evp_signature_from_dispatch(int name_id,
     EVP_SIGNATURE *signature = NULL;
     int ctxfncnt = 0, signfncnt = 0, verifyfncnt = 0, verifyrecfncnt = 0;
     int digsignfncnt = 0, digverifyfncnt = 0;
-    int gparamfncnt = 0, sparamfncnt = 0, gmdparamfncnt = 0, smdparamfncnt = 0;
+    int gparamfncnt = 0, gctxparamfncnt = 0, sctxparamfncnt = 0;
+    int gmdparamfncnt = 0, smdparamfncnt = 0;
 
     if ((signature = evp_signature_new(prov)) == NULL) {
         ERR_raise(ERR_LIB_EVP, ERR_R_MALLOC_FAILURE);
@@ -158,40 +159,42 @@ static void *evp_signature_from_dispatch(int name_id,
             if (signature->get_params != NULL)
                 break;
             signature->get_params = OSSL_get_OP_signature_get_params(fns);
+            gparamfncnt++;
             break;
         case OSSL_FUNC_SIGNATURE_GETTABLE_PARAMS:
             if (signature->gettable_params != NULL)
                 break;
             signature->gettable_params
                 = OSSL_get_OP_signature_gettable_params(fns);
+            gparamfncnt++;
             break;
         case OSSL_FUNC_SIGNATURE_GET_CTX_PARAMS:
             if (signature->get_ctx_params != NULL)
                 break;
             signature->get_ctx_params
                 = OSSL_get_OP_signature_get_ctx_params(fns);
-            gparamfncnt++;
+            gctxparamfncnt++;
             break;
         case OSSL_FUNC_SIGNATURE_GETTABLE_CTX_PARAMS:
             if (signature->gettable_ctx_params != NULL)
                 break;
             signature->gettable_ctx_params
                 = OSSL_get_OP_signature_gettable_ctx_params(fns);
-            gparamfncnt++;
+            gctxparamfncnt++;
             break;
         case OSSL_FUNC_SIGNATURE_SET_CTX_PARAMS:
             if (signature->set_ctx_params != NULL)
                 break;
             signature->set_ctx_params
                 = OSSL_get_OP_signature_set_ctx_params(fns);
-            sparamfncnt++;
+            sctxparamfncnt++;
             break;
         case OSSL_FUNC_SIGNATURE_SETTABLE_CTX_PARAMS:
             if (signature->settable_ctx_params != NULL)
                 break;
             signature->settable_ctx_params
                 = OSSL_get_OP_signature_settable_ctx_params(fns);
-            sparamfncnt++;
+            sctxparamfncnt++;
             break;
         case OSSL_FUNC_SIGNATURE_GET_CTX_MD_PARAMS:
             if (signature->get_ctx_md_params != NULL)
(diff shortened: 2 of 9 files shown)


crypto/evp/pmeth_fn.c
```diff
@@ -153,6 +154,17 @@ static void *evp_signature_from_dispatch(int name_id,
...
+            if (signature->get_params != NULL)
+                break;
+            signature->get_params = OSSL_get_OP_signature_get_params(fns);
+            break;
+        case OSSL_FUNC_SIGNATURE_GETTABLE_PARAMS:
+            if (signature->gettable_params != NULL)
+                break;
+            signature->gettable_params
+                = OSSL_get_OP_signature_gettable_params(fns);
+            break;
```
mattcaswell: We have a check below that if "get_ctx_params" is present then so must "gettable_ctx_params" and vice versa. It seems we should have something similar here. Similarly for the other dispatch functions.

doc/man7/provider-asymcipher.pod
```diff
@@ -165,9 +169,13 @@ See L<OSSL_PARAM(3)> for further details on the parameters structure used by
 the OP_asym_cipher_get_ctx_params() and OP_asym_cipher_set_ctx_params()
 functions.
 
+OP_asym_cipher_get_params() gets details of the algorithm implementation
+and stores them in I<params>.
```
mattcaswell: Shouldn't we be documenting the known common parameters here? Similarly for the other provider-*.pod files.

mattcaswell (changes requested):

crypto/evp/pmeth_fn.c
levitte: Fixed

doc/man7/provider-asymcipher.pod
levitte: Done

levitte:
Ping

doc/man7/provider-keyexch.pod
```diff
@@ -29,6 +29,8 @@ provider-keyexch - The keyexch library E<lt>-E<gt> provider functions
                        size_t outlen);
 
  /* Key Exchange parameters */
+ int OP_keyexch_get_params(const OSSL_PARAM params[]);
```
romen: this should drop `const`, right?

include/openssl/core_numbers.h
```diff
@@ -413,6 +415,8 @@ OSSL_CORE_MAKE_FUNC(int, OP_keyexch_derive, (void *ctx,  unsigned char *secret,
 OSSL_CORE_MAKE_FUNC(int, OP_keyexch_set_peer, (void *ctx, void *provkey))
 OSSL_CORE_MAKE_FUNC(void, OP_keyexch_freectx, (void *ctx))
 OSSL_CORE_MAKE_FUNC(void *, OP_keyexch_dupctx, (void *ctx))
+OSSL_CORE_MAKE_FUNC(int, OP_keyexch_get_params, (const OSSL_PARAM params[]))
```
romen: It should be an array of writeable OSSL_PARAM objects, to return the value to the caller.

include/openssl/core_numbers.h
```diff
@@ -481,6 +487,8 @@ OSSL_CORE_MAKE_FUNC(int, OP_signature_digest_verify_final,
                     (void *ctx, const unsigned char *sig, size_t siglen))
 OSSL_CORE_MAKE_FUNC(void, OP_signature_freectx, (void *ctx))
 OSSL_CORE_MAKE_FUNC(void *, OP_signature_dupctx, (void *ctx))
+OSSL_CORE_MAKE_FUNC(int, OP_signature_get_params, (const OSSL_PARAM params[]))
```
romen: Same here, I notice that in the documentation for signatures the `const` is correctly omitted.

include/openssl/core_numbers.h
```diff
@@ -528,6 +538,8 @@ OSSL_CORE_MAKE_FUNC(int, OP_asym_cipher_decrypt, (void *ctx, unsigned char *out,
                                                   size_t inlen))
 OSSL_CORE_MAKE_FUNC(void, OP_asym_cipher_freectx, (void *ctx))
 OSSL_CORE_MAKE_FUNC(void *, OP_asym_cipher_dupctx, (void *ctx))
+OSSL_CORE_MAKE_FUNC(int, OP_asym_cipher_get_params, (const OSSL_PARAM params[]))
```
romen: Same here, I notice that in the documentation for aym_ciphers the `const` is correctly omitted.

romen (changes requested):
If I am understanding correctly how to implement the getter, the argument cannot be `const OSSL_PARAM params[]`, as I need to be able to write something like:

```C
static
int ecdh_get_params(OSSL_PARAM params[])
{
    OSSL_PARAM *p;

    p = OSSL_PARAM_locate(params, OSSL_ALG_PARAM_KEYTYPE);
    if (p != NULL && !OSSL_PARAM_set_utf8_ptr(p, "EC"))
        return 0;

    return 1;
}
```

I also noticed that that regarding that `const`, documentation across algorithm types is not consistent (the .pod for keyexch has it, signature and asym_cipher omit it).

romen:
Sidenote:

maybe this is intentional, but looking at the debugger after implementing

```C
    { OSSL_FUNC_KEYEXCH_GETTABLE_PARAMS, (void (*)(void))ecdh_gettable_params },
    { OSSL_FUNC_KEYEXCH_GET_PARAMS, (void (*)(void))ecdh_get_params },
```

I notice that `ecdh_get_params` is called directly (correctly asking about `OSSL_ALG_PARAM_KEYTYPE`) without ever calling `ecdh_gettable_params` to check what params are gettable.

Is this behaviour by design or are we missing a check?

doc/man7/provider-keyexch.pod
levitte: Yes. This is a copy-n-paste mistake, I copied the set_ctx_param function and forgot to unconstify.

levitte:
> Is this behaviour by design or are we missing a check?

By design. If there is no name, the variable in libcrypto will remain NULL.

doc/man7/provider-keyexch.pod
levitte: Fixed

include/openssl/core_numbers.h
levitte: Fixed

include/openssl/core_numbers.h
levitte: Fixed

include/openssl/core_numbers.h
levitte: Fixed

romen (dismissed):
LGTM !

I believe this should go in already, but also that as a separate PR we will somehow need to let the EVP layer generate the inverse relationship.

When a PEM key is encountered and parsed it will be recognized by OID as an `id-ecPublicKey` object (yes, that is the `sn` for both private and public EC keys).

A provider (or several) register an entry like

```C
    { "EC:id-ecPublicKey", "default=yes", ec_keymgmt_functions },
```

This lets EVP know that providers ProvA and ProvB can handle such objects (and in the future also generate such objects), but still somehow EVP needs to know what kind of operations each provider allows on such key objects (or if it cannot enumerate all the supported operations, because maybe it's superfluous, it needs to know what to do if the user asks to generate a signature or a keyexchange using such a key object).

Bonus points if we manage to also handle that for a single key type there can be alternative implementations of, say, pkey_signature (e.g. ECDSA, deterministic ECDSA, another ElGamal scheme, EdDSA-like, etc.)

levitte:
> This lets EVP know that providers ProvA and ProvB can handle such objects (and in the future also generate such objects), but still somehow EVP needs to know what kind of operations each provider allows on such key objects (or if it cannot enumerate all the supported operations, because maybe it's superfluous, it needs to know what to do if the user asks to generate a signature or a keyexchange using such a key object).

Oooohhhhh...  hmm, yeah I see what you mean.  Dang, I actually think you approved this too early.

romen:
You can decide to put it back to WIP if you prefer and we will resume the reviews once it satisfies you again!

crypto/evp/exchange.c
```diff
@@ -159,6 +175,25 @@ EVP_KEYEXCH *EVP_KEYEXCH_fetch(OPENSSL_CTX *ctx, const char *algorithm,
...
+{
+    char *keytype_name = NULL;
+    OSSL_PARAM params[] = { OSSL_PARAM_END, OSSL_PARAM_END };
+
+    params[0] =
+        OSSL_PARAM_construct_utf8_ptr(OSSL_ALG_PARAM_KEYTYPE,
+                                      &keytype_name, 0);
+    if (exchange->get_params != NULL && !exchange->get_params(params))
+        return NULL;
+    if (keytype_name == NULL) {
```
slontis: lose the {

crypto/evp/pmeth_fn.c
```diff
@@ -317,6 +333,25 @@ void EVP_SIGNATURE_names_do_all(const EVP_SIGNATURE *signature,
...
+{
+    char *keytype_name = NULL;
+    OSSL_PARAM params[] = { OSSL_PARAM_END, OSSL_PARAM_END };
+
+    params[0] =
+        OSSL_PARAM_construct_utf8_ptr(OSSL_ALG_PARAM_KEYTYPE,
+                                      &keytype_name, 0);
+    if (signature->get_params != NULL && !signature->get_params(params))
+        return NULL;
+    if (keytype_name == NULL) {
```
slontis: same here

crypto/evp/pmeth_fn.c
```diff
@@ -569,6 +603,25 @@ int EVP_PKEY_verify_recover(EVP_PKEY_CTX *ctx,
...
+{
+    char *keytype_name = NULL;
+    OSSL_PARAM params[] = { OSSL_PARAM_END, OSSL_PARAM_END };
+
+    params[0] =
+        OSSL_PARAM_construct_utf8_ptr(OSSL_ALG_PARAM_KEYTYPE,
+                                      &keytype_name, 0);
+    if (cipher->get_params != NULL && !cipher->get_params(params))
+        return NULL;
+    if (keytype_name == NULL) {
```
slontis: and here :)

slontis (commented):
Some form of test might be nice... I guess romen will test it for you :)

levitte:
> Some form of test might be nice... I guess romen will test it for you :)

I'm totally trusting that this will get exercised by new code, yeah.

levitte:
So, the grander issue here, and this is what @romen identified, is that the "algorithm name" for keymgmt has served two purposes:

1.  Identifying the name of the algorithm that can make use of a particular key.
2.  Identifying the type of the key.

Operations, such as signature, key exchange, need the first identification, to make sure they have the right key to do their job.  This becomes crucial with `EVP_PKEY_CTX` based functions, since they often (not always, but often) work from a given key to find the appropriate algorithm to do signatures, key exchange, whatnot.

Key exporters (the export_to functions that we add to the diverse `_ameth.c` files), serializers (and upcoming, deserializers), and other things I cannot remember for the moment, are much more interested in the second identification, i.e. the key type.

We sniffed at this problem when discussing RSA vs RSA-PSS, but came out lucky, because RSA keys with the associated PSS parameters attached are identified with a separate OID at a PKCS#8 level, so we're getting away with it because the algorithm and key type happen to match, so we didn't need to bother, and as @mattcaswell put it, the "algorithm name" for keymgmt expressed what a specific key could be used for rather than what it is.  So far so good.

Enter EC...  and suddenly we have a distinguishable key *type* that doesn't have the same name as the algorithms that can use it.  In other words, the same problem re-surfaces again, but much more tangible and without the possibility to "get away with it" (**).

In summary, it looks like we need to separate what a key *is*, and what *it's used for*, while being able to tell how the two sides of that coin are related.

-----

(*) This is a *hard problem* which was partly resolved by plain C types, and partly by making certain assumptions.  Essentially, the key *type* was always a C type, and since the support of *all* operations was in one and the same method structure, we could quite liberally use whatever underlying key type we wanted without batting an eye (not entirely...  fitting the EC stuff into it was not an easy task and was paved with some pretty heavy refactoring at one point).

(**) This is not quite true, but the solution is full of assumptions.  In the use case of EC, for example, the assumption in the legacy `EVP_PKEY_METHOD` for `EVP_PKEY_EC` (which is the NID for "id-ecPublicKey") is that the signing function would do ECDSA.  This is quite visible when you have a look at [`crypto/ec/ecdsa_sign.c`](https://github.com/openssl/openssl/blob/master/crypto/ec/ecdsa_sign.c).  This will fall apart as soon as another generic signature algorithm that uses EC keys comes along.

levitte:
A problem is to figure out a way that doesn't put too much burden on the provider author.

The `EVP_PKEY_CTX` is created from a key, or in some cases, from a *key type* (`EVP_PKEY_CTX_new_id` is expressly said to be useful for domain parameter and key generation, which suggests that the NID used is meant to identity the key type rather than what it can be used for).  With that in mind, it seems reasonable to think that what we really need to do is the inverse of what this PR does, i.e. we need to find an appropriate operation algorithm for the key type associated with the EVP_PKEY_CTX rather than finding the appropriate key type, coming from operation that was derived...  from a key type.

I can see two ways of doing this:

1.  An exhaustive search for algorithms that respond with the correct name when asked with OSSL_ALG_PARAM_KEYTYPE...  but only as a fallback in case there is no algorithm with the same name as the key type.
2.  Add a query function to keymgmt, just like the provider-wide operation query, but specific to keys.  Again, only needed in case the key type name doesn't match the operation algorithm.

The first alternative is fairly easy to do, but has a computational impact that I'm not sure I want to deal with.  The second alternative would be easy enough *as long as* there's only one algorithm for each type of operation.

Unfortunately, time is running out on me today, so I won't get to this before possibly tomorrow.

Ideas and thoughts welcome.

mattcaswell:
> This will fall apart as soon as another generic signature algorithm that uses EC keys comes along.

You mean like SM2!

romen:
I am thinking about the second option: if I am understanding correctly it would mean that the keymgmt struct will have to signal by answering to a GET_KEYEXCHANGE, GET_SIGNATURE, GET_ASYM_ENC if it supports any of the above operations and what is the name of the associated algorithm (or maybe directly the pointer to the functions structure, as that is internal to the provider anyway and maybe we can save the second lookup).

This would limit to one keyexchange/signature/asym_enc for key type for provider. 

I think this might work, as long as we manage to support that we will have loaded at the same time more than one providers that can deserialize a given key. 

----

## Example: SM2 and EC

The keys are serialized using the same global OID "id-ecPublicKey", and we can distinguish them only when parsing inside the ASN1 structure, as luckily the default curve for SM2 has its own unique OID (even if technically if I remember correctly is exactly the same curve as NIST P-256).

So right now, the EC deserializer flags the key as an SM2 key if it detects that it is serialized as a named elliptic curve and curve matches the default SM2 curve. 

First problem here is that in theory one could run SM2 as a signature scheme on any other curve, but we can hope that the Chinese standardization body will keep the trend of publishing a dedicated OID for aliased curves, so we can pretend this is not a problem 

The real problem is that if we have the default provider supporting deserializing EC keys and a separate Chinese (or other national) crypto provider also capable of deserializing EC keys we shouldn't pick only one but have both try and hopefully only one will be successful (e.g. the default provider could not recognize the SM2 default curve OID while parsing and fail, while the Chinese crypto provider should fail if it does not find the OID for the default SM2 curve or future SM2 specific curve aliases).

If more than one providers manage to deserialize the key, then the conflict could be solved by establishing priorities in the conf file maybe? 

That would also allow to avoid running always all the potential deserializers (e.g. we try the Chinese provider only if default failed or viceversa). 


----

If we manage to have this flexibility then I don't see the limit of only one operation per operation type per key type per provider as a limit for provider authors, as basically the problem can then be solved by instantiating 2 different providers rather than a monolithic one.

romen:
Other example meaningful for external provider authors.

I could be writing a provider that supports an hardware accelerator that only supports NIST P-256 keys and only ECDH/ECDSA on it.

The provider should be able to deserialize EC keys (hopefully core can provide help here to avoid duplicating all these EC ASN1 structure parsers!) and retrieve what is the OID of the named curve. 
It should return a valid key object only for P-256 keys, and fail otherwise.

This failure shouldn't prevent applications from using e.g. P-521 (or any other curve, or SM2) using the software implementation from the default provider or another provider.

Right now in 1.1.1 this is already quite painful for ENGINEs that need to become providers of the whole EC_KEY_method, save a reference to the default method before registering, intercept any EC operation and fallback on the previous default method in case they don't recognize a supported curve NID.
I really hope we can do better with providers!

levitte:
A thing to remember here is that keymgmt and whatever operation implementation that's being used must be able to cooperate internally. Libcrypto must ensure this by making sure they both come from the same provider. So you're not escaping the monolithic nature...

romen:
> A thing to remember here is that keymgmt and whatever operation implementation that's being used must be able to cooperate internally. Libcrypto must ensure this by making sure they both come from the same provider. So you're not escaping the monolithic nature...

Well, authors could use `#ifdefs` to avoid duplicating source code while still exploiting their build system to build 2 different provider binaries, one that does ECDSA over supported EC keys and the other doing e.g. EdDSA over the same keys.

Keymgmt and signature functions structures will be self contained in each provider to ensure consistency, you will have 2 separate providers, but the code will still be unified. 

Not at all dissimilar from what we do with the default and the FIPS providers.

levitte:
Yup, that works

levitte:
> You mean like SM2!

I was thinking í future proofing terms, but yeah, that's a fitting example

levitte:
> I am thinking about the second option: if I am understanding correctly it would mean that the keymgmt struct will have to signal by answering to a GET_KEYEXCHANGE, GET_SIGNATURE, GET_ASYM_ENC if it supports any of the above operations and what is the name of the associated algorithm (or maybe directly the pointer to the functions structure, as that is internal to the provider anyway and maybe we can save the second lookup).
> 
> This would limit to one keyexchange/signature/asym_enc for key type for provider.
> 
> I think this might work, as long as we manage to support that we will have loaded at the same time more than one providers that can deserialize a given key.

The more I think about it, the more I get to the same conclusion, that we simply cannot support more than one signature scheme, key exchange scheme, yada yada yada, per key type...  at least on a libcrypto\<-\>provider interface level (nothing stops a provider from having variants of those schemes based on parameters that come along with the keys, if there are such variations (RSA-OAEP comes to mind)).

For *simplicity*, I think we need to deal with operation algorithm *names* and let libcrypto fetch them as it sees fit.  Anything else will add complexity to the fetching mechanism, and I think is complex enough as it is.  So, a keymgmt query function, much like the provider-wide one, that takes an operation identity, but simply returns a constant string?

We could have it in the back of our minds to allow for multiple schemes of each operation type.  I think the answer to that stares us in the face, as we have already solved that problem for algorithm names in `OSSL_ALGORITHM`: multiple names separate by colons.

levitte:
See #10647 for an implementation of the other way around.  I believe that's the way to do it and still keep our sanity.

levitte:
(if accepted, I will close this PR in favor of that one)

levitte:
Closing thi in favor of #10647

PR #10637, opened 2019-12-17
Enforce the order of CMS creating ctrl calls

beldmit:
On encryption
1. Allocate the necessary EVP_PKEY_CTX
2. Set passed parameters, if any
3. Set information about the cipher used, if necessary
4. Call the ASN1 envelope ctrl

##### Checklist

### DIFF
--- a/apps/cms.c
+++ b/apps/cms.c
@@ -22,6 +22,7 @@
 # include <openssl/x509_vfy.h>
 # include <openssl/x509v3.h>
 # include <openssl/cms.h>
+# include <openssl/core_names.h>
 
 static int save_certs(char *signerfile, STACK_OF(X509) *signers);
 static int cms_cb(int ok, X509_STORE_CTX *ctx);
@@ -877,23 +878,36 @@ int cms_main(int argc, char **argv)
         for (i = 0; i < sk_X509_num(encerts); i++) {
             CMS_RecipientInfo *ri;
             cms_key_param *kparam;
-            int tflags = flags;
+            EVP_PKEY_CTX *pctx;
             X509 *x = sk_X509_value(encerts, i);
+            OSSL_PARAM ctx_params[2];
+
             for (kparam = key_first; kparam; kparam = kparam->next) {
                 if (kparam->idx == i) {
-                    tflags |= CMS_KEY_PARAM;
                     break;
                 }
             }
-            ri = CMS_add1_recipient_cert(cms, x, tflags);
+            ri = CMS_add1_recipient_cert(cms, x, flags | CMS_KEY_PARAM);
             if (ri == NULL)
                 goto end;
+
+            pctx = CMS_RecipientInfo_get0_pkey_ctx(ri);
             if (kparam != NULL) {
-                EVP_PKEY_CTX *pctx;
-                pctx = CMS_RecipientInfo_get0_pkey_ctx(ri);
                 if (!cms_set_pkey_param(pctx, kparam->param))
                     goto end;
             }
+            /* As we do not get parameter here, the cast is safe */
+            ctx_params[0] = OSSL_PARAM_construct_utf8_string(
+                OSSL_ALG_PARAM_CIPHER, (char *)EVP_CIPHER_name(cipher), 
+                strlen(EVP_CIPHER_name(cipher)));
+            ctx_params[1] = OSSL_PARAM_construct_end();
+
+            if (OSSL_PARAM_locate_const(EVP_PKEY_CTX_settable_params(pctx),
+                OSSL_ALG_PARAM_CIPHER) != NULL) {
+                if (EVP_PKEY_CTX_set_params(pctx, ctx_params) <=0)
+                    goto end;
+            }
+
             if (CMS_RecipientInfo_type(ri) == CMS_RECIPINFO_AGREE
                 && wrap_cipher) {
                 EVP_CIPHER_CTX *wctx;


beldmit:
@slontis, @levitte - could you take a look, please?

beldmit:
It's currently a semi-hack solution. To finalize it, we need to get rid of `CMS_KEY_PARAM` processing on encryption.

beldmit:
Travis failures are relevant.

beldmit:
Travis builds has succeeded all but 2 hanged ones.

crypto/evp/pmeth_lib.c
```diff
@@ -747,7 +747,7 @@ static int legacy_ctrl_to_param(EVP_PKEY_CTX *ctx, int keytype, int optype,
                   EVP_R_OPERATION_NOT_SUPPORTED_FOR_THIS_KEYTYPE);
         return -2;
     }
-    return 0;
+    return -2;
```
slontis: hmm.. It looks like the only way this function runs is when new provider code is running..
if ((EVP_PKEY_CTX_IS_DERIVE_OP(ctx) && ctx->op.kex.exchprovctx != NULL)
            || (EVP_PKEY_CTX_IS_SIGNATURE_OP(ctx)
                && ctx->op.sig.sigprovctx != NULL)
            || (EVP_PKEY_CTX_IS_ASYM_CIPHER_OP(ctx)
                && ctx->op.ciph.ciphprovctx != NULL))

So does it need to be a legacy ctrl?
Or could be just use EVP_PKEY_CTX_set_params() instead?

apps/cms.c
```diff
@@ -877,23 +877,31 @@ int cms_main(int argc, char **argv)
         for (i = 0; i < sk_X509_num(encerts); i++) {
             CMS_RecipientInfo *ri;
             cms_key_param *kparam;
-            int tflags = flags;
+            int tflags = flags | CMS_KEY_PARAM; /* This flag enforces allocating the EVP_PKEY_CTX for the recipient here */
```
slontis: This top bit is not too different as it was doing this most of the time anyway.
Looks like the KeyAgree path should also work with this change as it doesnt use the flag.

apps/cms.c
```diff
@@ -877,23 +877,31 @@ int cms_main(int argc, char **argv)
...
+
+            pctx = CMS_RecipientInfo_get0_pkey_ctx(ri);
             if (kparam != NULL) {
-                EVP_PKEY_CTX *pctx;
-                pctx = CMS_RecipientInfo_get0_pkey_ctx(ri);
                 if (!cms_set_pkey_param(pctx, kparam->param))
                     goto end;
             }
+
+            res = EVP_PKEY_CTX_ctrl(pctx, -1, -1,
```
slontis: See below comment - could we just use params here instead?

beldmit: Yes, EVP_PKEY_CTX_set_params seems relevant here too.

crypto/evp/pmeth_lib.c
beldmit: Yes, EVP_PKEY_CTX_set_params seems relevant here.

apps/cms.c
beldmit: Here is the place where we process this bit:

https://github.com/openssl/openssl/blob/f64f26220442db3c6913188e6014e5bc5bc34653/crypto/cms/cms_env.c#L164-L171

If it is not set, we just call the default envelope ASN1 control. If it is, we first initialize the ctx, then set params and call the envelope ASN1 control that can refer to set parameters.

apps/cms.c
beldmit: So what's the name of the parameter is relevant here?

beldmit:
I've replaced EVP_PKEY_CTX_ctrl to EVP_PKEY_CTX_set_params but had to change the returning code of EVP_PKEY_CTX_set_params to -2 in case we don't have any appropriate settings. I'm not sure it's a good decision.

Any guidelines will be appropriated. @slontis, could you please look again?

apps/cms.c
beldmit: Fixed

crypto/evp/pmeth_lib.c
beldmit: Rolled back

beldmit:
It's also worth mentioning that EVP_PKEY_CTX_set_params is incompatible with ENGINE interfaces, and I'm not sure that all the engines will be converted to providers very fast...

crypto/evp/pmeth_lib.c
```diff
@@ -588,7 +588,7 @@ int EVP_PKEY_CTX_set_params(EVP_PKEY_CTX *ctx, OSSL_PARAM *params)
             && ctx->op.ciph.cipher->set_ctx_params != NULL)
         return ctx->op.ciph.cipher->set_ctx_params(ctx->op.ciph.ciphprovctx,
                                                      params);
-    return 0;
```
slontis: I dont think this is a good idea.. see above

apps/cms.c
```diff
@@ -897,8 +900,10 @@ int cms_main(int argc, char **argv)
                     goto end;
             }
 
-            res = EVP_PKEY_CTX_ctrl(pctx, -1, -1,
-                        EVP_PKEY_CTRL_CIPHER, EVP_CIPHER_nid(cipher), NULL);
+            ctx_params[0] = OSSL_PARAM_construct_utf8_string("cipher",
+                cipher_name, strlen(cipher_name));
+            ctx_params[1] = OSSL_PARAM_construct_end();
+            res = EVP_PKEY_CTX_set_params(pctx, ctx_params);
             if (res <= 0 && res != -2)
```
slontis: You can query if its settable via..
EVP_PKEY_CTX_settable_params

apps/cms.c
```diff
@@ -897,8 +900,10 @@ int cms_main(int argc, char **argv)
                     goto end;
             }
 
-            res = EVP_PKEY_CTX_ctrl(pctx, -1, -1,
-                        EVP_PKEY_CTRL_CIPHER, EVP_CIPHER_nid(cipher), NULL);
+            ctx_params[0] = OSSL_PARAM_construct_utf8_string("cipher",
```
slontis: This string would need to be defined properly as a #define and documented in a pod.

crypto/evp/pmeth_lib.c
```diff
@@ -747,7 +747,7 @@ static int legacy_ctrl_to_param(EVP_PKEY_CTX *ctx, int keytype, int optype,
                   EVP_R_OPERATION_NOT_SUPPORTED_FOR_THIS_KEYTYPE);
         return -2;
     }
-    return -2;
```
slontis: probably the same here..

apps/cms.c
beldmit: Will do.

apps/cms.c
beldmit: Fix applied.

apps/cms.c
beldmit: Hopefully, it's enough or should it still be documented?

crypto/evp/pmeth_lib.c
beldmit: Removed

beldmit:
I think we need smth like EVP_PKEY_CTX_param_is_settable(EVP_PKEY_CTX *, const char *).

beldmit:
Also not sure whether I should deal with OPENSSL_strdup and OPENSSL_free on error or just remove the const modifier from the result of EVP_CIPHER_name explicitly.

apps/cms.c
```diff
@@ -908,13 +912,15 @@ int cms_main(int argc, char **argv)
 
             settable = EVP_PKEY_CTX_settable_params(pctx);
```
slontis: An example is in evp_enc.c.. 
       /* Check the cipher actually understands this parameter */
        if (OSSL_PARAM_locate_const(EVP_CIPHER_settable_ctx_params(c->cipher),
                                    OSSL_CIPHER_PARAM_KEYLEN) == NULL)
            return 0;

slontis: There is also a OSSL_PARAM_locate

slontis:
> Also not sure whether I should deal with OPENSSL_strdup and OPENSSL_free on error or just remove the const modifier from the result of EVP_CIPHER_name explicitly.

Depends on how it gets used.. It looks like it is not stored - it is just used to resolve a cipher and then thrown away - so the cast should be ok.. Maybe a comment saying this and a cast is ok.

apps/cms.c
beldmit: Applied. Many thanks for the clue!

beldmit:
CI is happy now.

beldmit:
@slontis, could you please look at it now?

apps/cms.c
```diff
@@ -877,23 +878,38 @@ int cms_main(int argc, char **argv)
...
+            /* As we do not get parameter here, the cast is safe */
+            ctx_params[0] = OSSL_PARAM_construct_utf8_string(
+                OSSL_ALG_PARAM_CIPHER, (char *)EVP_CIPHER_name(cipher), 
+                strlen(EVP_CIPHER_name(cipher)));
+            ctx_params[1] = OSSL_PARAM_construct_end();
+
+            if (OSSL_PARAM_locate_const(EVP_PKEY_CTX_settable_params(pctx),
+                OSSL_ALG_PARAM_CIPHER) != NULL) {
+                res = EVP_PKEY_CTX_set_params(pctx, ctx_params);
+                if (res <= 0 && res != -2)
```
slontis: The -2 check should no longer be needed..

beldmit:
Travis is happy.

beldmit:
@slontis, could you please re-review and approve if there are no problems with this patch?

beldmit:
ping @openssl/otc

beldmit:
See #10916 as full implementation of Russian GOST CMS.

t8m:
Ping @beldmit  this is ready-to-merge

beldmit:
This PR contradicts to the #10904 suggesting more comprehensive approach. So I'd prefer to leave unmerged for some time.

levitte:
Ok, what's the state of this PR now?

beldmit:
It is mostly superseded by #10904 so if you could look at it, it will be great.

beldmit:
As #10904 is merged, this PR may be closed.

PR #10639, opened 2019-12-18
Fix travis timeout by excluding arm64 gcc -fsanitize=address build

slontis:
Travis constantly timing out after 50 minutes has finally annoyed me enough to fix it..

##### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -150,6 +150,12 @@ matrix:
           compiler: gcc
         - arch: arm64
           os: osx
+        # arm64 times out when using -fsanitize=address
+        # The amd64 build should be sufficient to test this.
+        # Note the env line must exactly match the env setting above 
+        - arch: arm64
+          env: CONFIG_OPTS="no-asm -Werror --debug no-afalgeng no-shared enable-rc5 enable-md2 -fsanitize=address" LSAN_OPTIONS="report_objects=1"
+
 
 before_script:
     - env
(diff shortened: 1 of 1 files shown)


richsalz:
Build fix, so not subject to 24 hours?

.travis.yml
```diff
@@ -150,6 +150,12 @@ matrix:
           compiler: gcc
         - arch: arm64
           os: osx
+        # arm64 times out when using -fsanitize=address
+        # The amd64 build should be sufficient to test this.
+        # Note the env line must exactly match the env setting above
```
mspncp: It took me a while to understand the meaning of line 155. How about making the following clarification?

```
# Note the env line must exactly match the env line from the build
# matrix above which contains the `-fsanitize=address` option.
```

levitte: Hmm, this gives me the heeby jeebies, someone is bound to make a mistake some day (say, update the config higher up without modifying it accordingly here). I guess it can be OK now, but should be refactored later on. Please add a TODO comment to that effect.

slontis: Updated the comment : Also added another note above the actual build matrix entry so that it might be read :)

slontis:
ping

mspncp:
@levitte wouldn’t it be better to mark it ‚ready to merge‘ right away if it has the ‚urgent‘ label?

levitte:
Good point.

slontis:
Thanks for reviewing quickly. Merged to master.

PR #10640, opened 2019-12-18
Add fips self test DEP for solaris and hpux

slontis:
##### Checklist

### DIFF
--- a/providers/fips/selftest.c
+++ b/providers/fips/selftest.c
@@ -50,6 +50,10 @@ DEFINE_RUN_ONCE_STATIC(do_fips_self_test_init)
     return self_test_lock != NULL;
 }
 
+#define DEP_DELCARE()                                                          \
+void init(void);                                                               \
+void cleanup(void);
+
 /*
  * This is the Default Entry Point (DEP) code. Every platform must have a DEP.
  * See FIPS 140-2 IG 9.10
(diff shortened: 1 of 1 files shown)


mattcaswell:

levitte:
> Travis red cross is not relevant.

That build job consistently fails with a timeout. We need to take it off, or deal with it. 
Not for this PR, though

providers/fips/selftest.c
```diff
@@ -85,14 +85,44 @@ BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpvReserved)
...
+void dep_init(void)
+{
+    FIPS_state = FIPS_STATE_SELFTEST;
+}
+
+#pragma fini "dep_cleanup"
+void dep_cleanup(void)
+{
+    CRYPTO_THREAD_lock_free(self_test_lock);
+}
```
mattcaswell: Could we use init/cleanup names which are consistent with the GNUC version?

Also I wonder if we could somehow combine all these function implementations together somehow. They're all basically the same function implementations with variations around how we declare them or mark them as init/fini functions.

slontis: Ok I have attempted to do the second thing..
you cant put #pragmas inside macros so it does become a bit messy...

providers/fips/selftest.c
```diff
@@ -50,6 +50,10 @@ DEFINE_RUN_ONCE_STATIC(do_fips_self_test_init)
     return self_test_lock != NULL;
 }
 
+#define DEP_DELCARE()                                                          \
```
levitte: Who's care? 😉

slontis: Not me :) Least I spelt it consistently wrong..
Updated

slontis:
Merged to master.

PR #10642, opened 2019-12-18
Add some missing cfi frame info in camellia-x86_64.pl

bernd-edlinger:
A few functions missed cfi instrumentation in this assembler module


bernd-edlinger:
Merged to master and 1.1.1
Thanks!

PR #10643, opened 2019-12-18
Fix no-des build

bernd-edlinger:
##### Checklist


bernd-edlinger:
Merged to master as 74a5808
Thanks!

PR #10644, opened 2019-12-18
Fix no-dh and no-dsa

mattcaswell:
Fix various miscellaneous missing guards for no-dh and no-dsa.

The first commit (i.e. the changes to evp_extra_test.c) will also be applied to the 1.1.1 branch. The other changes are only relevant in master.


mattcaswell:
Pushed. Thanks.

PR #10631, opened 2019-12-15
Move EC & ECDH to providers

romen:
This PR should address #10351 and #10352,  and pave the way to complete #10250 & #10356.

## TODO

- [x] ~~core ⇔ provider glue~~ ~~(depends on #10647)~~
- [x] ~~`EC_KEY` dirty counting~~
- [x] ~~basic ECDH exchange~~
  - [x] use padded BN export to params for secret scalars
  - [x] dirty_cnt flag for cofactor ECDH
- [x] complete ECDH
  - [x] cofactor
  - [x] KDF (core side)
  - [x] KDF (provider side)
- [x] tests?
- [x] update documentation (~~all~~ most of the new params)
- [x] rebase & git history massaging

## Features intentionally deferred to future PRs/issues:
- [ ] provider EC key generation (either as part of #10289, or separately)
- [ ] provider EC key serialization/deserialization 
- [ ] tweaks for inclusion in FIPS provider (left for #10250)
- [ ] SM2 support (#10357 & #10358)
- [ ] explicit parameters curves (discussion needed)
  - [ ] support in provider-side EC keymgmt
  - [ ] support in core-side `eckey_asn1_meth->export_to()`
- [ ] ECDSA support (#10356)
- [ ] more tests (can we systematically test for proper dirty counting?)
- [ ] documentation of parameters for import/export and get/set_params for EC (but also for DH, DSA, RSA, ECX) (see https://github.com/openssl/openssl/pull/10631#discussion_r379477175)

## Notes about SM2

Although the bulk of the work for moving SM2 into the default provider is left to a future PR, some notes are relevant here, as they shape some design decisions for this PR.

 - On the provider side, for **keymgmt**, SM2 and EC will likely be 2 separate (but similar) `OSSL_DISPATCH` objects associated to 2 `OSSL_ALGORITHM`s with different names (see https://github.com/openssl/openssl/pull/10631#discussion_r363436692 and previous comments).
 - On core side, `eckey_asn1_meth->export_to()` will likely end up having to handle both EC and SM2 keys, so on the provider-side keymgmt the import/export functions of SM2 and EC should be kept in sync.
- Key generation and signature setup of SM2 slightly differs from EC (they have different ranges for the secret scalar, see https://github.com/openssl/openssl/issues/8435#issuecomment-481540010), so `keymgmt->keygen` and the signature `OSSL_ALGORITHM`s will have to reflect this difference
[...]

PR #10647, opened 2019-12-18
CORE & EVP: Adapt KEYEXCH, SIGNATURE and ASYM_CIPHER to handle key types better

levitte:
The adaptation is to handle the case when key types and operations
that use these keys have different names.  For example, EC keys can be
used for ECDSA and ECDH.

This also involves specifying OP_query_operation_name() for KEYMGMT

This will allow keymgmt implementation for key types that need it to
specify the names of the diverse operation algorithms it can be used
with.  Currently, only one name per key type and operation is allowed.

### DIFF
--- a/doc/man3/EVP_PKEY_CTX_new.pod
+++ b/doc/man3/EVP_PKEY_CTX_new.pod
@@ -21,14 +21,14 @@ EVP_PKEY_CTX_dup, EVP_PKEY_CTX_free
 =head1 DESCRIPTION
 
 The EVP_PKEY_CTX_new() function allocates public key algorithm context using
-the algorithm specified in I<pkey> and ENGINE I<e>.
+the I<pkey> key type and ENGINE I<e>.
 
 The EVP_PKEY_CTX_new_id() function allocates public key algorithm context
-using the algorithm specified by I<id> and ENGINE I<e>.
+using the key type specified by I<id> and ENGINE I<e>.
 
 The EVP_PKEY_CTX_new_provided() function allocates a public key algorithm
 context using the library context I<libctx> (see L<OPENSSL_CTX(3)>), the
-algorithm specified by I<name> and the property query I<propquery>.  None
+key type specified by I<name> and the property query I<propquery>.  None
 of the arguments are duplicated, so they  must remain unchanged for the
 lifetime of the returned B<EVP_PKEY_CTX> or of any of its duplicates.
 
@@ -44,10 +44,24 @@ If I<ctx> is NULL, nothing is done.
 
 =head1 NOTES
 
+=over 4
+
+=item 1.
+
 The B<EVP_PKEY_CTX> structure is an opaque public key algorithm context used
 by the OpenSSL high level public key API. Contexts B<MUST NOT> be shared between
 threads: that is it is not permissible to use the same context simultaneously
 in two threads.
+
+=item 2.
+
+We mention "key type" in this manual, which is the same
+as "algorithm" in most cases, allowing either term to be used
+interchangeably.  There are algorithms when they I<key type> and the
+I<algorithm> of the operations that use the keys are not the same,
+such as EC keys being used for ECDSA and ECDH operations.
+
+=back
 
 =head1 RETURN VALUES
 
--- a/doc/man3/EVP_PKEY_CTX_new.pod
+++ b/doc/man3/EVP_PKEY_CTX_new.pod
@@ -57,7 +57,7 @@ in two threads.
 
 We mention "key type" in this manual, which is the same
 as "algorithm" in most cases, allowing either term to be used
-interchangeably.  There are algorithms when the I<key type> and the
+interchangeably.  There are algorithms where the I<key type> and the
 I<algorithm> of the operations that use the keys are not the same,
 such as EC keys being used for ECDSA and ECDH operations.
 
--- a/doc/man3/EVP_PKEY_CTX_new.pod
+++ b/doc/man3/EVP_PKEY_CTX_new.pod
@@ -57,7 +57,7 @@ in two threads.
 
 We mention "key type" in this manual, which is the same
 as "algorithm" in most cases, allowing either term to be used
-interchangeably.  There are algorithms when they I<key type> and the
+interchangeably.  There are algorithms when the I<key type> and the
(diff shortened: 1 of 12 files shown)


levitte:
This is an alternative to #10636, which implements kind of the same thing, but the other way around.  Instead of asking the operation implementations what key types they support, we ask the KEYMGMT for the key what operations it supports and by what name.

This should fit better with the way OpenSSL works with EVP_PKEY_CTX, where the key type governs what is being done.

romen (changes requested):
Just a nit for now, but I will have to wait until later today for testing.

romen:
@levitte do you already have ideas of what could happen when we will have a `"SM2:ec-PublicKey", sm2_keymgmt_functions` entry?

The SM2 key type is serialized exactly like any other EC key type and differs only by the specific curve OID used inside the ASN1 structure, so both the EC and the SM2 keymgmt should intercept deserialization requests. 

Do you think it would be possible to establish some priority order that works within a provider and across providers? 

Ideally such a mechanism could allow to try to deserialize in the following scenario:

1. External EC provider for P-256 hw module
2. External EC provider for nation-specific EC (e.g. GOST) 
3. Default provider: try SM2
4. Default provider: try EC

This order should be configurable by the user/system administrator and not coded inside the providers or libcrypto

levitte:
I see two ways to handle EC vs SM2:

1. Treat them as distinct key types. Deserializers will have to reject data that doesn't fit the criteria for the type, even if they can decode the incoming payload. Are SM2 keys identified with a distinct OID when wrapped in PKCS#8? In that case, this seems like the way to go, even if the internal structure used in our providers when operating on it is the same as for ECDSA et al.
Incidently, this is what we already do in the legacy implementation of RSA-PSS. 
2. Treat them as the same, i.e see then as EC keys with some distinct parameters attached. This would be the choice if the PKCS#8 algorithm identifier is id-ecPublicKey for SM2 keys. Internally in the provider, we will still get a parameterised key structure, but the difference is that the operation algorithm will have to be able to support all variants.

The second choice makes me think that we might want to support recursive providers, i.e implementations that do their own fetching. Oh boy...

levitte:
In the end, we need to look at PKCS#8 when looking at what types we face...

romen:
Unfortunately, SM2 keys use the same OID as EC keys.
But the approach in number 2 would mean that any provider capable of handling id-ecPubliKey keys must support every EC (or derived) algorithm. 

For example if a user wants to load a provider to enable her shiny hw crypto accelerator (that only supports P-256) we don't want (unless configured otherwise) to lose the ability to handle other curves through the default provider or SM2, etc.

levitte:
Oh, one thing that I'm forgetting here is that we currently don't have any provided deserializers. They will normally be used via the OSSL_STORE "file:" backend, and my plan is to do exactly what I think you request, to loop through all available deserializers until one successfully returns an unpacked object.

romen:
Same problem would arise if we carry on the plans to have national-based crypto in separate providers.

Let's say the brainpool curves are not in the default provider anymore but in a German specific provider, or the binary curves are moved to the legacy provider.

We really don't won't to repeat the mistake of having ENGINE/providers have to hijack the whole EC module for that!

levitte:
It dawned on me that we're not as dependent on pkcs#8 as I was thinking, so we can *choose* to make SM2 keys a distinct key type for keymgmt, and teach our deserializers to check for specific parameters to decide if the payload is acceptable or not.
Thank you, your input also made it clearer how to treat RSA-PSS keys.

levitte:
> We really don't won't to repeat the mistake of having ENGINE/providers have to hijack the whole EC module for that!

Agreed.  So I think that between the two choices in https://github.com/openssl/openssl/pull/10647#issuecomment-567035099, the first choice provides the possibility for distinct and less monolithic providers.

levitte:
Rebased for better Travis results

romen:
Minus the nit mentioned in the previous review, the testing done so far in https://github.com/openssl/openssl/pull/10631#issuecomment-570932253 suggests everything seems to work fine in this PR.

doc/man3/EVP_PKEY_CTX_new.pod
```diff
@@ -44,11 +44,25 @@ If I<ctx> is NULL, nothing is done.
...
 The B<EVP_PKEY_CTX> structure is an opaque public key algorithm context used
 by the OpenSSL high level public key API. Contexts B<MUST NOT> be shared between
 threads: that is it is not permissible to use the same context simultaneously
 in two threads.
 
+=item 2.
+
+We mention "key type" in this manual, which is the same
+as "algorithm" in most cases, allowing either term to be used
+interchangeably.  There are algorithms when they I<key type> and the
```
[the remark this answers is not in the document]
levitte: Fixed

levitte:
Fixed, and rebased

doc/man3/EVP_PKEY_CTX_new.pod
```diff
@@ -44,11 +44,25 @@ If I<ctx> is NULL, nothing is done.
...
 The B<EVP_PKEY_CTX> structure is an opaque public key algorithm context used
 by the OpenSSL high level public key API. Contexts B<MUST NOT> be shared between
 threads: that is it is not permissible to use the same context simultaneously
 in two threads.
 
+=item 2.
+
+We mention "key type" in this manual, which is the same
+as "algorithm" in most cases, allowing either term to be used
+interchangeably.  There are algorithms when the I<key type> and the
```
slontis: where = when

slontis (commented):
Apart from a small NIT it LGTM

doc/man3/EVP_PKEY_CTX_new.pod
levitte: Fixed

slontis (approved):

levitte:
Merged.

e62a45b60e3166d995eb83840f437debf575c328 CORE & EVP: Specify OP_query_operation_name() for KEYMGMT
f23bc0b770efd229e27519786d9f28da56752d0a EVP: Adapt KEYEXCH, SIGNATURE and ASYM_CIPHER to handle key types better
2293032121d33299be4f21a599aab84413467f8d PROV: Adjust the KEYMGMT name specs to include all names

PR #10650, opened 2019-12-18
Add some missing cfi frame info in aes-x86_64.pl

bernd-edlinger:
##### Checklist


bernd-edlinger:
There is a typo in the commit message
 "Add some missing cfi frame info in aes-x64_64.pl"

will change that to aes-x86_64.pl before committing

bernd-edlinger:
Merged to master, only.
I think have removed that assembler module from 1.1.1 a while ago.

PR #10651, opened 2019-12-18
Add some missing cfi frame info in x86_64-mont5.pl

bernd-edlinger:
##### Checklist

### DIFF
--- a/crypto/bn/asm/x86_64-mont5.pl
+++ b/crypto/bn/asm/x86_64-mont5.pl
@@ -582,6 +582,7 @@ $code.=<<___;
 .type	mul4x_internal,\@abi-omnipotent
 .align	32
 mul4x_internal:
+.cfi_startproc
 	shl	\$5,$num		# $num was in bytes
 	movd	`($win64?56:8)`(%rax),%xmm5	# load 7th argument, index
 	lea	.Linc(%rip),%rax
@@ -1076,6 +1077,7 @@ $code.=<<___
 ___
 }
 $code.=<<___;
+.cfi_endproc
 .size	mul4x_internal,.-mul4x_internal
 ___
 }}}
@@ -1241,6 +1243,7 @@ $code.=<<___;
 .align	32
 bn_sqr8x_internal:
 __bn_sqr8x_internal:
+.cfi_startproc
 	##############################################################
 	# Squaring part:
 	#
@@ -2032,6 +2035,7 @@ __bn_sqr8x_reduction:
 	cmp	%rdx,$tptr		# end of t[]?
 	jb	.L8x_reduction_loop
 	ret
+.cfi_endproc
 .size	bn_sqr8x_internal,.-bn_sqr8x_internal
 ___
 }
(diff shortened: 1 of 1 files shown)


crypto/bn/asm/x86_64-mont5.pl
```diff
@@ -3742,6 +3759,7 @@
 	lea	(%r10),%rsp
 	ret
 .LSEH_end_bn_gather5:
+.cfi_endproc
```
kroeckx: The .LSEH_end_bn_gather5 seems to be unused. I find it rather strange after the ret.

bernd-edlinger: I did not check, just assumed that is used for windows seh?

crypto/bn/asm/x86_64-mont5.pl
```diff
@@ -2752,6 +2761,7 @@
 	mov	8*2(%rbp),%r14
 	mov	8*3(%rbp),%r15
 	jmp	.Lsqrx4x_sub_entry	# common post-condition
+.cfi_endproc
```
kroeckx: This look wrong at first sight, I don't see a ret.

bernd-edlinger: consider it somehing like a tail-call, so
the function containing .Lsqrx4x_sub_entry does the ret

crypto/bn/asm/x86_64-mont5.pl
bernd-edlinger: it seems to be used in windows at line 3915
```
$code.=<<___;
        .rva    .LSEH_begin_bn_gather5
        .rva    .LSEH_end_bn_gather5
        .rva    .LSEH_info_bn_gather5
```

crypto/bn/asm/x86_64-mont5.pl
kroeckx: I think this is already covered by the one at the end of __bn_postx4x_internal

kroeckx: I guess I don't know enough about how this is supposed to work

bernd-edlinger: Right, as soon as the control flow enters __bn_postx4x_internal, the
cfi info is correct, just between mulx4x_internal and here, the
unwinder would need to guess.

bernd-edlinger:
I can basically grep for "^.size" and look at the previous line,
if it is not .cfi_endproc that is suspicious, I just leave Windows SEH handler functions alone.

crypto/bn/asm/x86_64-mont5.pl
bernd-edlinger: no idea if the windows seh handling is sane or not...

crypto/bn/asm/x86_64-mont5.pl
bernd-edlinger: when gdb walks a function without any cfi instruction, it can try to
load the return address from sp[0] and catch any faults in accessing
those frames, the stack walking code in https://github.com/openssl/openssl/pull/10635#issue-353973686

I do basically just use this instrumented version of the speed command
and look at the stack traces from `opensssl speed` where they do not end
at PC=0

kroeckx:
I wonder if it makes sense to put the the common code in it's own section.

bernd-edlinger:
Merged to master and 1.1.1
Thanks!

PR #10653, opened 2019-12-18
Add some missing cfi frame info in aesni-x86_64.pl

bernd-edlinger:
##### Checklist

### DIFF
--- a/crypto/aes/asm/aesni-x86_64.pl
+++ b/crypto/aes/asm/aesni-x86_64.pl
@@ -986,6 +986,7 @@ $code.=<<___;
 .type	aesni_ccm64_encrypt_blocks,\@function,6
 .align	16
 aesni_ccm64_encrypt_blocks:
+.cfi_startproc
 ___
 $code.=<<___ if ($win64);
 	lea	-0x58(%rsp),%rsp
@@ -1068,6 +1069,7 @@ $code.=<<___ if ($win64);
 ___
 $code.=<<___;
 	ret
+.cfi_endproc
 .size	aesni_ccm64_encrypt_blocks,.-aesni_ccm64_encrypt_blocks
 ___
 ######################################################################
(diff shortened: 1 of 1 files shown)


crypto/aes/asm/aesni-x86_64.pl
```diff
@@ -4710,6 +4725,7 @@ sub aesni_generate8 {
 	shufps	\$0b10101010,%xmm1,%xmm1	# critical path
 	xorps	%xmm1,%xmm2
 	ret
+.cfi_endproc
```
kroeckx: Lkey_expansion_256a and key_expansion_256b are something that's called, so maybe they should should have thrown own .cfi_startproc and .cfi_endproc

kroeckx: Oh, for added fun, key_expansion_256a_cold also has it's own one.

crypto/aes/asm/aesni-x86_64.pl
```diff
@@ -986,6 +986,7 @@ sub aesni_generate8 {
 .type	aesni_ccm64_encrypt_blocks,\@function,6
 .align	16
 aesni_ccm64_encrypt_blocks:
+.cfi_startproc
 ___
 $code.=<<___ if ($win64);
 	lea	-0x58(%rsp),%rsp
```
kroeckx: Should we do something with the CFA here?

bernd-edlinger: no, I can't do anything for win64

bernd-edlinger: win64 unwind info is totally broken, I wonder if it would
be better to remove that altogether, @davidben ?

kroeckx: I missed the win64 part, but saw the win64 part in one of the other PRs.

bernd-edlinger: as far as I can tell the .cfi directives are ignored by windows.
The unwinding cannot not work correctly when this block is interrupted.
It appears to be correcting the stack and restoring xmm6-9 only after
.Lccm64_enc_body:
but when interrupted directly after the lea only the stack needs to be corrected,
but the registers are not yet saved.
I wonder how that stuff was tested.

bernd-edlinger:
Merge to master and 1.1.1
Thanks!

PR #10663, opened 2019-12-19
Updates NEWS for new release

mattcaswell:
This is for the 1.0.2 branch only and, since this is a NEWS update for the new release the 24hr rule does not apply.


mattcaswell:
Pushed. Thanks.

Issue #10271, opened 2019-10-26
OSCP HTTP client and other apps could benefit from improvements in cmp_http.c


DDvO:
When @mpeylo submitted his CMP implementation to OpenSSL a couple of years back he got the feedback that it should not use libCURL but OpenSSL's own HTTP client, which so far is used mainly for fetching OCSP responses (but also CRLs and certificates) .
So we changed CMPforOpenSSL to make use of the implementation in `crypto/ocsp/ocsp_ht.c`.

Unfortunately this HTTP client is tailored for use with OCSP and is very primitive - in particular, handling of delayed responses and errors is very poor and there is no support for more advanced features such redirection and use of proxies.

One of the next chunks of our current CMP contribution contains the HTTP client to be used for CMP, which will reside in `crypto/cmp/cmp_http.c`. The current state of its development can be found in https://github.com/mpeylo/cmpossl/blob/cmp-dev/crypto/cmp/cmp_http.c
This implementation is based on the mentioned HTTP code in `ocsp_ht.c`, while it contains improvements for those of its shortcomings critical for CMP: it
* can send and receive any types of ASN.1-formatted requests and responses
* does not include a potentially busy loop when waiting for responses 
but makes use of timeout mechanism integrated with socket-based BIO
* returns more useful diagnostics in various error situations
* supports the use of HTTP proxies and of TLS (including TLS over proxies)

I've already shaped the code in `cmp_http.c` such that several portions of it could more or less directly generalize and improve the existing HTTP client code in `ocsp_ht.c`, `apps/ocsp.c`, `apps/s_client.c`, and `apps/lib/apps.c`.

What is your view on these improvements - should I push them upstream (as a couple of independent PRs) for general use or should we keep them in `cmp_http.c`?

richsalz:
It would be a shame to have two pieces of code that do the same thing.

levitte:
I agree with @richsalz. Please make a generic http client that can be used by both CMP and OCSP. I'm not sure where it should be located in the end, but `crypto/http/` is at least a start.

mattcaswell:
Agree with @levitte and @richsalz.

DDvO:
Fine. I've meanwhile started implementing this.
Though I'm on vacation this week I estimate to be able to present a first version on Friday.

DDvO:
Phew, implementing this generalized HTTP client was more work than expected.

[...]

PR #9982, opened 2019-09-23
Add a minimal build target for Travis and Appveyor

bernd-edlinger:

PR #4992, opened 2017-12-28
Introducing an application-level library for the CLI and OpenSSL-based applications

DDvO:
So far, the OpenSSL code has essentially a three-level structure with a stack of two libraries and a command-line application at the top:

`apps/openssl`
`libssl`
`libcrypto`

The `apps/` directory contains various generally useful code like handling crypto-related files and messages, general TLS client/server and CA functionality, support for higher-level protocols like S/MIME, CRL fetching, OCSP, and certainly more to come.
While this code serves as a model for using the crypto and ssl libraries and can be used in a limited way by invoking the `openssl` application binary (CLI), it cannot be re-used directly. Other applications that need similar functionality need to copy or re-implement portions of that code and then maintain it.
On the other hand, the two libraries contain some code that is actually too high-level for them, for instance the minimal HTTP client as part of the crypto library (`crypto/ocsp/ocsp_ht.c`). It would be very helpful to introduce a further level in the hierarchy containing a more application-oriented library:

`apps/openssl` and external OpenSSL-based applications
`libsecapp` <-- new (with tentative name)
`libssl`
`libcrypto`

Then all more high-level and application support functionality can go there. This would make much of the generally useful code that so far resides in the `apps/` folder directly accessible to other applications at the programming level, i.e., in the form of a library/API, with all the re-usability advantages that this brings. It would also relieve `libcrypto `from more high-level topics like HTTP.
This library would also form an ideal condensation point for further higher-level uses of crypto and TLS that may in the future get integrated with OpenSSL. 

This pull request contains a fully functional proof-of-concept implementation for the above proposal. It introduces a library with the preliminary name 'secapp'.

For now, the library consists of essentially the code that so far as been in `apps/apps.c` and `apps/app_rand.c`, moved to the `secapp/` directory. (I had to move and partly rename some functions between these two source files and `apps/opt.c` because the cut between these sources was not entirely clean. Of course, when this PR is accepted, some further cleanup would make sense.) I moved most of `apps/apps.h` to `include/openssl/apps.h`, while the strictly internal declarations in it went in a new file `apps/apps_locl.h` and the ones related to the command-line option functionality went in a new file `apps/opt.h`.

The high-level `build.info` mechanism proved very useful for adding the new library, stating its dependencies etc. Yet it turned out that its implementation in `Configure` and `util/mkdef.pl` is not complete - I added respective TODOs there.

I made sure that all this (including tests) builds and runs both in a Linux-like environment (Cygwin) and native Windows (Visual C).

##### Checklist

[...]

PR #10667, opened 2019-12-20
Generalize and improve the HTTP client

DDvO:
This PR solves/implements #10271 as discussed there by
generalizing the HTTP client so far implemented mostly in `crypto/ocsp/ocsp_ht.c`.

The new HTTP client has become an independent libcrpyto module in `crypto/http/` and
- can handle any types of requests and responses (both ASN.1-encoded and plain)
- does not include potentially busy loops when waiting for responses but
- makes use of a new timeout mechanism integrated with socket-based BIO
- supports the use of HTTP proxies and TLS, including HTTPS over proxies, via a callback
- supports HTTP redirection via codes 301 and 302 for GET requests
- returns more useful diagnostics in various error situations

I had to adapt - and was able to simplify and improve - hitherto uses of HTTP in 
`crypto/ocsp/`, `crypto/x509/x_all.c`, `apps/lib/apps.c`, and `apps/{ocsp,s_client,s_server}.c`.
The new features will be used heavily by chunk 10 of the CMP contribution.
The various documentation, header, and meta files are updated accordingly.


DDvO:
I've aimed at not changing the existing OCSP-centric HTTP API, which was a challenge in itself.
I think/hope I've managed to keep the API of those functions documented so far,
yet I took the liberty to adapt the types of those functions that have been undocumented.
See the respective comments towards the end of `include/openssl/http.h` for details.

crypto/err/openssl.txt
```diff
@@ -2524,6 +2526,28 @@ EVP_R_WRAP_MODE_NOT_ALLOWED:170:wrap mode not allowed
...
+HTTP_R_ERROR_RECEIVING:103:error receiving
+HTTP_R_ERROR_SENDING:102:error sending
+HTTP_R_INCONSISTENT_CONTENT_LENGTH:120:inconsistent content length
+HTTP_R_MAX_RESP_LEN_EXCEEDED:117:max resp len exceeded
+HTTP_R_MISSING_ASN1_ENCODING:110:missing asn1 encoding
+HTTP_R_MISSING_CONTENT_TYPE:121:missing content type
+HTTP_R_MISSING_REDIRECT_LOCATION:111:missing redirect location
+HTTP_R_REDIRECTION_FROM_HTTPS_TO_HTTP:112:redirection from https to http
+HTTP_R_REDIRECTION_NOT_ENABLED:116:redirection not enabled
+HTTP_R_RESPONE_LINE_TOO_LONG:113:respone line too long
```
FdaSilvaYY: typo in error code and message: `respone`

kroeckx:
> - supports the use of HTTP proxies and TLS, including HTTPS over proxies

Does that mean libcrypto now depends on libssl? Does that work on all platforms?

doc/man3/OSSL_HTTP_transfer.pod
```diff
@@ -0,0 +1,195 @@
...
+On success it returns the response as a pointer to B<ASN1_VALUE>.
+
+OSSL_HTTP_transfer() exchanges an HTTP request and response with
+the given B<server> and optional B<port> and B<path>, which defaults to "/".
+If B<use_ssl> is nonzero a TLS connection is requested and the B<bio_update_fn>
+parameter, described below, must be provided.
+If B<req_mem> is NULL it uses the HTTP GET method, else it uses HTTP POST to
+send a request with the contents of the memory BIO and optional B<content_type>.
+The optional list B<headers> may contain additional custom HTTP header lines.
+If B<req_mem> is NULL (i.e., the HTTP method is GET) and B<redirection_url>
```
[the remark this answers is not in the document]
FdaSilvaYY: Forget my comment, I checked its use; you are right.

doc/man3/OCSP_sendreq_new.pod
```diff
@@ -10,22 +10,22 @@ OCSP_REQ_CTX_set1_req, OCSP_sendreq_bio - OCSP responder query functions
...
 
  int OCSP_REQ_CTX_add1_header(OCSP_REQ_CTX *rctx,
                               const char *name, const char *value);
 
- int OCSP_REQ_CTX_set1_req(OCSP_REQ_CTX *rctx, OCSP_REQUEST *req);
+ int OCSP_REQ_CTX_set1_req(OCSP_REQ_CTX *rctx, const OCSP_REQUEST *req);
 
- OCSP_RESPONSE *OCSP_sendreq_bio(BIO *io, const char *path, OCSP_REQUEST *req,
-                                 int maxline);
+ OCSP_RESPONSE *OCSP_sendreq_bio(BIO *io, const char *path, OCSP_REQUEST *req);
```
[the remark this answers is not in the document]
DDvO: Indeed this change is a fix of an error in the pre-existing documentation.
Is your suggestion to backport this fix also to other branches?

doc/man3/OSSL_CMP_CTX_new.pod
```diff
@@ -351,18 +349,17 @@ OSSL_CMP_CTX_set_http_cb_arg() or NULL if unset.
...
-                                       OSSL_CMP_MSG **res);
+ typedef OSSL_CMP_MSG *(*OSSL_cmp_transfer_cb_t) (OSSL_CMP_CTX *ctx,
+                                                  const OSSL_CMP_MSG *req);
+
 Returns 1 on success, 0 on error.
 
 Default is NULL, which implies the use of L<OSSL_CMP_MSG_http_perform(3)>.
-The callback should send the CMP request it obtains via the B<req> parameter
-and on success place the response in the B<*res> output parameter.
+The callback should send the CMP request messaage it obtains via the B<req>
```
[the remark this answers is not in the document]
DDvO: fixed

DDvO:
> - supports the use of HTTP proxies and TLS, including HTTPS over proxies
> Does that mean libcrypto now depends on libssl? 

No (and if it did, this would mean a cyclic dependency).
I solved this challenge via the callback `OSSL_HTTP_bio_cb_t bio_update_fn`.
See the new doc (https://github.com/mpeylo/cmpossl/blob/http_client_generalized/doc/man3/OSSL_HTTP_transfer.pod) for details.
An instance of the callback is defined, e.g., in `apps/lib/apps.c`, where libssl is available.

> Does that work on all platforms?

I believe so (but haven't tried this out).
The adapted/extended code does not make use of new platform-specific facilities.

DDvO:
> Quick review, impressive work, IMHO. +1

Thanks!

This PR is the result of a really large amount of work that stretched over 2+ years.

DDvO:
BTW, this work dates back to an issue I brought up in late Aug 2017: 
https://mta.openssl.org/pipermail/openssl-dev/2017-August/009629.html

DDvO:
> Does that mean libcrypto now depends on libssl?

This touches a long-standing deficiency in the structure of the OpenSSL libraries:
The crypto lib includes many modules that do not really belong there but 
should be part of a more application-level library that should be above libssl. 

See also https://github.com/openssl/openssl/pull/4992 and my original report and solution suggestion dating back to May 2016:
https://mta.openssl.org/pipermail/openssl-dev/2017-September/009712.html

doc/man3/OCSP_sendreq_new.pod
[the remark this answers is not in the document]
FdaSilvaYY: Yes, if this doc-error is also present on current active branchs like 1.1.0

DDvO:
Thanks for the initial comments on this generalized HTTP implementation given more than two weeks ago, which I answered.

Ping - can someone please do a formal review?

DDvO:
@levitte, in early November you had given some initial remarks on a preview of  this PR and showed interest in having a closer look at this code: https://github.com/openssl/openssl/issues/10271#issuecomment-549348671
When do you think you can do a review?

doc/man3/OCSP_sendreq_new.pod
```diff
@@ -35,14 +35,14 @@ response header maximum line length of B<maxline>. If B<maxline> is zero a
 default value of 4k is used. The OCSP request B<req> may be set to B<NULL>
 and provided later if required.
 
-OCSP_sendreq_nbio() performs non-blocking I/O on the OCSP request context
-B<rctx>. When the operation is complete it returns the response in B<*presp>.
+OCSP_sendreq_nbio() performs I/O on the OCSP request context B<rctx>.
```
mattcaswell: Why removal of "non-blocking"?

doc/man3/OCSP_sendreq_new.pod
```diff
@@ -92,14 +90,8 @@ a Host header for B<ocsp.com> you would call:
...
-If OCSP_sendreq_nbio() indicates an operation should be retried the
-corresponding BIO can be examined to determine which operation (read or
-write) should be retried and appropriate action taken (for example a select()
-call on the underlying socket).
-
-OCSP_sendreq_bio() does not support retries and so cannot handle non-blocking
-I/O efficiently. It is retained for compatibility and its use in new
-applications is not recommended.
+OCSP_sendreq_bio() does not support timeout nor setting extra headers.
+It is retained for compatibility.
```
mattcaswell: If you are intending to encourage some other function instead you should state what the preferred function is.

doc/man3/OSSL_HTTP_transfer.pod
```diff
@@ -0,0 +1,195 @@
...
+         BIO *sbio = BIO_new_ssl(ctx, 1);
+         hbio = sbio != NULL ? BIO_push(sbio, hbio) : NULL;
+     } else if (!connect && !detail) { /* disconnecting after error */
+         /* optionally add diagnostics here */
+     }
+     return hbio;
+ }
+
+A callback function that also supports TLS connection with an HTTPS proxy
+is implemented under the name app_http_tls_cb() in apps/lib/apps.c.
```
mattcaswell: I don't think references to the apps code belong in the man pages.

include/openssl/http.h
```diff
@@ -0,0 +1,95 @@
...
+void OCSP_set_max_response_length(OCSP_REQ_CTX *rctx, unsigned long len);
+/* The parameters of this fn, which never was documented, have been changed: */
+int OCSP_REQ_CTX_http(OCSP_REQ_CTX *rctx,
+                      const char *server, const char *port, const char *path);
+int OCSP_REQ_CTX_add1_header(OCSP_REQ_CTX *rctx,
+                             const char *name, const char *value);
+/* The parameters of this fn, which never was documented, have been extended: */
+int OCSP_REQ_CTX_i2d(OCSP_REQ_CTX *rctx, const char *content_type,
+                     const ASN1_ITEM *it, ASN1_VALUE *req);
+int OCSP_REQ_CTX_nbio(OCSP_REQ_CTX *rctx);
```
mattcaswell: Moving these function declarations seems to be a breaking change

util/missingcrypto.txt
```diff
@@ -445,6 +445,7 @@ ERR_load_ENGINE_strings(3)
 ERR_load_ERR_strings(3)
 ERR_load_ESS_strings(3)
 ERR_load_EVP_strings(3)
+ERR_load_HTTP_strings(3)
```
mattcaswell: Hmmmm.....I'm not sure what we should do about these functions. Do new ones actually need to be public?

crypto/bio/b_sock.c
```diff
@@ -369,4 +376,111 @@ int BIO_sock_info(int sock,
...
+        BIOerr(0, rv == 0 ? BIO_R_TRANSFER_TIMEOUT : BIO_R_TRANSFER_ERROR);
+    return rv;
+}
+
+/*
+ * Connect via the given BIO using BIO_do_connect() until success/timeout/error.
+ * Parameter timeout == 0 means infinite, < 0 leads to immediate timeout error.
+ * Returns -1 on error, 0 on timeout, and 1 on success.
+ */
+int BIO_connect_retry(BIO *bio, long timeout)
```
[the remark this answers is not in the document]
DDvO: Thanks a lot @mattcaswell  for your review comments!
I've already started fixing the various issues.

Concerning the use of `long` for timeout values, the idea was to support time intervals as long as (reasonably) possible.
I've just realized that `int` should be sufficient here, since even if it is just 32 bits wide and we use only positive values they support 2^31 seconds, which is more than 68 years :-)
So I'll change the timeout value type to `int`.

apps/lib/apps.c
```diff
@@ -1981,6 +1933,133 @@ void store_setup_crl_download(X509_STORE *st)
...
+    case SSL_R_WRONG_VERSION_NUMBER:
+        return "The server does not support (a suitable version of) TLS";
+    case SSL_R_UNKNOWN_PROTOCOL:
+        return "The server does not support HTTPS";
+    case SSL_R_CERTIFICATE_VERIFY_FAILED:
+        return "Cannot authenticate server via its TLS certificate, likely due to mismatch with our trusted TLS certs or missing revocation status";
+    case SSL_AD_REASON_OFFSET + TLS1_AD_UNKNOWN_CA:
+        return "Server did not accept our TLS certificate, likely due to mismatch with server's trust anchor or missing revocation status";
+    case SSL_AD_REASON_OFFSET + SSL3_AD_HANDSHAKE_FAILURE:
+        return "Server requires our TLS certificate but did not receive one";
```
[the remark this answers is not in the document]
DDvO: Good to know. I had found out on the situations described in this aux function mostly by experiments some 2+ years ago.
I've generalized the description.

apps/lib/apps.c
```diff
@@ -1981,6 +1933,133 @@ void store_setup_crl_download(X509_STORE *st)
...
+        BIO_set_ssl(sbio, ssl, BIO_CLOSE);
+
+        hbio = BIO_push(sbio, hbio);
+    } else if (!connect && !detail) { /* disconnecting after error */
+        const char *hint = tls_error_hint();
+        if (hint != NULL)
+            ERR_add_error_data(1, hint);
+        /*
+         * as workaround for libssl double free, do not pop the sbio,
+         * but rely on BIO_free_all() done by HTTP_transfer() in http_client.c
```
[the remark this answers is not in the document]
DDvO: I've improved the comment to
```
         * If we pop sbio and BIO_free() it this may lead to libssl double free.
         * Rely on BIO_free_all() done by OSSL_HTTP_transfer() in http_client.c
```
This is due to my experience from about 1+ years back that I get crashes due to double free when I pop the BIO and free it myself. This smells like a libssl bug to me.

crypto/bio/b_sock.c
```diff
@@ -369,4 +376,111 @@ int BIO_sock_info(int sock,
     return 1;
 }
 
-#endif
+/* TODO use this function, simplify further other uses of select() in apps/ */
```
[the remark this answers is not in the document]
DDvO: The TODO was a bit outdated - I've improved it to
```
/* TODO simplify by BIO_socket_wait() further other uses of select() in apps/ */
```
The idea is that there are further uses of `select()` in a couple of OpenSSL apps that could be made simpler by using this function (but that's not the focus of this PR).

crypto/bio/b_sock.c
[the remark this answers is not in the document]
DDvO: Done meanwhile.

crypto/bio/b_sock.c
```diff
@@ -369,4 +376,111 @@ int BIO_sock_info(int sock,
...
+        return 0;
+    }
+
+    if (!blocking)
+        BIO_set_nbio(bio, 1);
+
+ retry: /* it does not help here to set SSL_MODE_AUTO_RETRY */
+    rv = BIO_do_connect(bio); /* This indirectly calls ERR_clear_error(); */
+
+    if (rv <= 0 && (errno == ETIMEDOUT
```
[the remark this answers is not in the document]
DDvO: Yep, changed to `get_last_sys_error()`

crypto/bio/b_sock.c
```diff
@@ -369,4 +376,111 @@ int BIO_sock_info(int sock,
...
+                        || ERR_GET_REASON(ERR_peek_error()) == ETIMEDOUT)) {
```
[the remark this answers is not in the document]
DDvO: Oops, this must be due to some code refactoring accident. 
Since I could not find any other case where a timeout could have been flagged via the error queue I've removed this part of the condition.

crypto/http/http_client.c
```diff
@@ -0,0 +1,1199 @@
...
+
+#define HTTP_DEFAULT_MAX_LINE_LENGTH (4 * 1024)
+#define HTTP_DEFAULT_MAX_RESP_LEN (100 * 1024)
+
+/* HTTP states */
+
+#define OHS_NOREAD          0x1000 /* If set no reading should be performed */
+#define OHS_ERROR           (0 | OHS_NOREAD) /* Error condition */
+#define OHS_FIRSTLINE       1 /* First line being read */
+#define OHS_REDIRECT        10 /* Looking for redirection location */
```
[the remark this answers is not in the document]
DDvO: I've accordingly changed the `10` to `0xa`.

crypto/http/http_client.c
```diff
@@ -0,0 +1,1199 @@
...
+#define OHS_CONTENT         4 /* HTTP content octets being read */
+#define OHS_WRITE_INIT     (5 | OHS_NOREAD) /* 1st call: ready to start I/O */
+#define OHS_WRITE          (6 | OHS_NOREAD) /* Request being sent */
+#define OHS_FLUSH          (7 | OHS_NOREAD) /* Request being flushed */
+#define OHS_DONE           (8 | OHS_NOREAD) /* Completed */
+#define OHS_HTTP_HEADER    (9 | OHS_NOREAD) /* Headers set, w/o final \r\n */
+
+OSSL_HTTP_REQ_CTX *OSSL_HTTP_REQ_CTX_new(BIO *bio, int method_GET, int maxline,
+                                         unsigned long max_resp_len,
+                                         long timeout,
```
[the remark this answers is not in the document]
DDvO: Yep, I've consistently replaced all `long timeout`  by `int timeout`
(while the `unsigned long max_resp_len` cannot be changed because it's legacy).

crypto/http/http_client.c
```diff
@@ -0,0 +1,1199 @@
...
+
+    if ((req_len = BIO_get_mem_data(req_mem, &req)) <= 0)
+        return 0;
+    rctx->state = OHS_WRITE_INIT;
+
+    return BIO_printf(rctx->mem, "Content-Length: %ld\r\n\r\n", req_len) > 0
+        && BIO_write(rctx->mem, req, req_len) == (int)req_len;
+}
+
+BIO *HTTP_ASN1_item2BIO(const ASN1_ITEM *it, ASN1_VALUE *val)
```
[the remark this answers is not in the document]
DDvO: Changed to `HTTP_asn1_item2bio()`.

crypto/http/http_client.c
```diff
@@ -0,0 +1,1199 @@
...
+
+static int OSSL_HTTP_REQ_CTX_add1_headers(OSSL_HTTP_REQ_CTX *rctx,
+                                          const STACK_OF(CONF_VALUE) *headers,
+                                          const char *host)
+{
+    int i;
+    int add_host = 1;
+
+    for (i = 0; i < sk_CONF_VALUE_num(headers); i++) {
+        CONF_VALUE *hdr = sk_CONF_VALUE_value(headers, i);
```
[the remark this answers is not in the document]
DDvO: Fixed

crypto/http/http_client.c
```diff
@@ -0,0 +1,1199 @@
...
+        && (req_mem == NULL
+            || OSSL_HTTP_REQ_CTX_content(rctx, content_type, req_mem)))
+        return rctx;
+
+    OSSL_HTTP_REQ_CTX_free(rctx);
+    return NULL;
+}
+
+/*
+ * Parse first HTTP response line. This should like this: "HTTP/1.0 200 OK".
```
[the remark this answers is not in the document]
DDvO: Fixed

crypto/http/http_client.c
```diff
@@ -0,0 +1,1199 @@
...
+    for (reason = code; *reason != '\0' && !ossl_isspace(*reason); reason++)
+        continue;
+
+    if (*reason == '\0') {
+        HTTPerr(0, HTTP_R_SERVER_RESPONSE_PARSE_ERROR);
+        return 0;
+    }
+
+    /* Set end of response code and start of message */
+    *reason++ = 0;
```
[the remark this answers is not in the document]
DDvO: Yes, fixed

crypto/http/http_client.c
```diff
@@ -0,0 +1,1199 @@
...
+        val = rctx->max_resp_len;
+    }
+    if (rctx->resp_len != 0 && rctx->resp_len != len) {
+        HTTPerr(0, HTTP_R_INCONSISTENT_CONTENT_LENGTH);
+        tag = ",before=";
+        val = rctx->resp_len;
+    }
+    if (tag != NULL) {
+        char len_str[32];
+        char str[32];
```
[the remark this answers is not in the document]
DDvO: Fixed

crypto/http/http_client.c
```diff
@@ -0,0 +1,1199 @@
...
+        }
+        if (*p) /* not end of headers */
+            goto next_line;
+
+        if (rctx->expected_ct != NULL) {
+            HTTPerr(0, HTTP_R_MISSING_CONTENT_TYPE);
+            ERR_add_error_data(2, "expected=", rctx->expected_ct);
+            return 0;
+        }
+        if (rctx->state == OHS_REDIRECT) {
```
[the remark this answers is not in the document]
DDvO: This check cannot be done before all the header has been parsed.
I've added an explaining comment:
```
            /* http status code indicated redirect but there was no Location */
```

crypto/http/http_client.c
```diff
@@ -0,0 +1,1199 @@
...
+            if (redirection_url == NULL)
+                HTTPerr(0, HTTP_R_REDIRECTION_NOT_ENABLED);
+            else
+                /* may be NULL if out of memory: */
+                (*redirection_url) = OPENSSL_strdup(rctx->redirection_url);
+        } else {
+            char buf[200];
+            unsigned long err = ERR_peek_error();
+            int lib = ERR_GET_LIB(err);
+            int reason = ERR_GET_REASON(err);
```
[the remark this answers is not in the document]
DDvO: Fixed

crypto/http/http_client.c
```diff
@@ -0,0 +1,1199 @@
...
+    if (out == NULL)
+        return 0;
+
+    i = EVP_EncodeBlock((unsigned char *)out, buf, len);
+    if (!ossl_assert((size_t)i <= outl)) {
+        OPENSSL_free(out);
+        return NULL;
+    }
+    if (i < 0)
+        *out = '\0';
```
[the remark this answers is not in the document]
DDvO: Me too.
I had taken over this function (as far as I recall) from `ocsp.c`.
Actually `EVP_EncodeBlock()` never returns `< 0`.
I've moved the check into the assertion above.

crypto/http/http_client.c
```diff
@@ -0,0 +1,1199 @@
...
+ * This is typically called by an app, so bio_err and prog are used unless NULL
+ * to print additional diagnostic information in a user-oriented way.
+ */
+int OSSL_HTTP_proxy_connect(BIO *bio, const char *server, const char *port,
+                            const char *proxyuser, const char *proxypass,
+                            long timeout, BIO *bio_err, const char *prog)
+{
+# undef BUF_SIZE
+# define BUF_SIZE (8 * 1024)
+    char *mbuf = OPENSSL_malloc(BUF_SIZE);
```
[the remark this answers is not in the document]
DDvO: I don't know what the reasoning was since I basically just moved this function.
Yet in my view allocating 8k on the stack is a bit much. 
I know that in other situations (at most) 4k are allocated.

doc/man3/OCSP_sendreq_new.pod
DDvO: I've removed "non-blocking" in the documentation in order to rectify an inconsistency: 
already quite some time ago the `_nbio` and `_bio` functions where unified, and whoever did this forgot to update the documentation accordingly. 
The `_nbio` prevailed likely for backward compatibility.

doc/man3/OSSL_HTTP_transfer.pod
DDvO: I thought it would have been helpful, nevertheless removed it
(and added in somewhat compensation a comment before the function definition).

include/openssl/http.h
DDvO: I see, so I've moved these declarations back to `ocsp.h`.

util/missingcrypto.txt
DDvO: Errrr..... this issue is not the point of this PR.
I know that _thy shall not add new functions to missingcrypto.txt_
but I did this here fully in line with existing practice for the other string loading functions.
I don't know if these need to be public (but I do know that `ERR_load_HTTP_strings()`) needs to be called on lib init, which I made sure),
and if they don't need to be public IMHO this should be done in some independent cleanup PR.

mattcaswell: Fair enough...my comment was more of a side note.

doc/man3/OCSP_sendreq_new.pod
DDvO: I just extended that existing sentence so that statement should have been given already earlier.
T rectify this I've now added
```
Better use B<OCSP_sendreq_nbio()> instead.
```

DDvO:
@mattcaswell, very pleased to receive your comments!
I've addressed all of them.

util/missingcrypto.txt
DDvO: Well, the cleanest solution would be to simply add in `doc/man3/ERR_load_crypto_strings.pod` one or two sentences generically for all these string loading functions.

On the other hand, apparently none of them need to be exported. 
For curiosity I just did a little experiment: remove all those specific `ERR_load_*_strings()`  functions from `util/libcrypto.um` and call `make test`. 
Nothing bad happened. Even `make doc-nits` is happy (also with the then redundant entries in `util/missingcrypto.txt`).

crypto/http/http_client.c
```diff
@@ -0,0 +1,1199 @@
...
+        goto end;
+
+    resp = OSSL_HTTP_REQ_CTX_transfer(rctx);
+    if (resp == NULL) {
+        if (rctx->redirection_url != NULL) {
+            if (redirection_url == NULL)
+                HTTPerr(0, HTTP_R_REDIRECTION_NOT_ENABLED);
+            else
+                /* may be NULL if out of memory: */
+                (*redirection_url) = OPENSSL_strdup(rctx->redirection_url);
```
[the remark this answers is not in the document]
DDvO: Yep, removed.

crypto/http/http_client.c
```diff
@@ -0,0 +1,1199 @@
...
+        (void)ERR_pop_to_mark();
+    else
+        (void)ERR_clear_last_mark();
+
+    return resp;
+}
+
+static int redirection_ok(int n_redir, const char *old_url, const char *new_url)
+{
+    static const char *const https = "https:";
```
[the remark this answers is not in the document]
DDvO: Yes, done.

crypto/http/http_client.c
```diff
@@ -0,0 +1,1199 @@
...
+        char *proxyauth, *proxyauthenc;
+
+        if (proxypass != NULL)
+            len += strlen(proxypass);
+        proxyauth = OPENSSL_malloc(len + 2);
+        if (proxyauth == NULL)
+            goto end;
+        BIO_snprintf(proxyauth, len + 2, "%s:%s", proxyuser,
+                     (proxypass != NULL) ? proxypass : "");
+        proxyauthenc = base64encode(proxyauth, strlen(proxyauth));
```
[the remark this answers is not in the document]
DDvO: Thanks @FdaSilvaYY for your comments.
Good point to re-use the length returned by `BIO_snprintf`.
I've further improved the length handling and added a sanity check.

DDvO:
I've meanwhile rebased this PR and Travis has recovered from an intermittent timeout issue that occurred yesterday, so it's is up and running again.

DDvO:
@mattcaswell, when can you check if the changes you requested are acceptable for you,
and do you plan to provide further comments for this PR?

doc/man3/OSSL_HTTP_transfer.pod
```diff
@@ -0,0 +1,193 @@
...
+                                 const char *path, int use_ssl,
+                                 const char *proxy, const char *proxy_port,
+                                 OSSL_HTTP_bio_cb_t bio_update_fn, void *arg,
+                                 const STACK_OF(CONF_VALUE) *headers,
+                                 const char *content_type,
+                                 ASN1_VALUE *req, const ASN1_ITEM *req_it,
+                                 int maxline, unsigned long max_resp_len,
+                                 int timeout, const char *expected_ct,
+                                 const ASN1_ITEM *rsp_it);
+ BIO *OSSL_HTTP_transfer(const char *server, const char *port, const char *path,
```
mattcaswell: I'm wondering whether the API functions above should allow you to control the underlying transport BIO that gets used rather than it being created indirectly via the server/port or url. E.g. by optionally passing the BIO in, or having a callback which allows you to create he BIO yourself.

crypto/http/http_client.c
```diff
@@ -0,0 +1,1204 @@
...
+
+    /* Read past all following headers */
+    do {
+        /*
+         * TODO: This does not necessarily catch the case when the full
+         * HTTP response came in in more than a single TCP message.
+         */
+        read_len = BIO_gets(fbio, mbuf, BUF_SIZE);
+    }
+    while (read_len > 2);
```
mattcaswell: Nit: while should be on the same line as }

doc/man3/X509_load_http.pod
```diff
@@ -0,0 +1,50 @@
...
+
+=head1 SYNOPSIS
+
+ #include <openssl/x509.h>
+
+ X509 *X509_load_http(const char *url, int timeout);
+ X509_CRL *X509_CRL_load_http(const char *url, int timeout);
+
+ #define X509_http_nbio(url)     X509_load_http(url, 0)
+ #define X509_CRL_http_nbio(url) X509_CRL_load_http(url, 0)
```
mattcaswell: As above - should there be an option to provide the underlying BIO?

crypto/http/http_client.c
DDvO: Fixed, as well as two similar code formatting nits in `http_lib.c`.
I've also added checks for situations like this in the format checking tool #10363.

doc/man3/OSSL_HTTP_transfer.pod
DDvO: Good idea! Done.

Adding a further (optional) parameter `BIO *cbio` for a custom connection BIO was easy to do.
Just note that the `server` and `port` (and `proxy`) parameters are still used/needed, e.g., to fill in the HTTP header.

doc/man3/X509_load_http.pod
DDvO: Yes.
Analogously to above, the`url` parameter is still used/needed, e.g., to fill in the HTTP header.

DDvO:
> A couple of questions and a minor nit. Otherwise this looks good.

@mattcaswell, pleased to hear!
I've fulfilled also the new change request(s).

mattcaswell:
> @mattcaswell, pleased to hear!
I've fulfilled also the new change request(s).

Ok - will try and take a look tomorrow.

mattcaswell:
Oh,...and in the meantime, it looks like a rebase is already needed!

DDvO:
> Oh,...and in the meantime, it looks like a rebase is already needed!

Indeed - thanks for pointing this out.
Lately new functions have been added to crypto/ quite often, and this time it was even a new lib module.

crypto/bio/b_sock.c
```diff
@@ -369,4 +375,110 @@ int BIO_sock_info(int sock,
...
+        BIOerr(0, rv == 0 ? BIO_R_TRANSFER_TIMEOUT : BIO_R_TRANSFER_ERROR);
+    return rv;
+}
+
+/*
+ * Connect via the given BIO using BIO_do_connect() until success/timeout/error.
+ * Parameter timeout == 0 means infinite, < 0 leads to immediate timeout error.
+ * Returns -1 on error, 0 on timeout, and 1 on success.
+ */
+int BIO_connect_retry(BIO *bio, int timeout)
```
mattcaswell: So now the BIO might have been passed in from the application...I wonder how this function copes with something like a mem bio?

doc/man3/OSSL_HTTP_transfer.pod
```diff
@@ -0,0 +1,198 @@
...
+OSSL_HTTP_get() uses HTTP GET to obtain data (of any type) from the given B<url>,
+using B<cbio> as connection BIO if not NULL, and returns it as a memory BIO.
+
+OSSL_HTTP_get_asn1() uses HTTP GET to obtain an ASN.1-encoded value
+(e.g., an X.509 certificate) with the expected structure specified by B<it>
+(e.g., I<ASN1_ITEM_rptr(X509)>) from the given B<url>,
+using B<cbio> as connection BIO if not NULL,
+and returns it on success as a pointer to I<ASN1_VALUE>.
+
+OSSL_HTTP_post_asn1() uses the HTTP POST method to send a request B<req>
```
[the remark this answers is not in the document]
DDvO: Oops, fixed

crypto/bio/b_sock.c
DDvO: Good thought!

I've meanwhile tried with a mem BIO, and it turns out that the function fails (as expected) but without an entry in the error queue. It turns out that a
```
        BIOerr(0, BIO_R_UNSUPPORTED_METHOD);
```
was missing in the default case of `mem_ctrl()` in `crypto/bio/bss_mem.c`.

I've simply added this line (as a separate commit in this PR) rather than issuing an independent PR.
Moreover, since it may happen that the BIO call fails also for other reasons (and maybe for a different BIO type) without adding an error queue entry I've added a catch-up to `BIO_connect_retry()`.

Doing this I found that also the distinction within `BIO_connect_retry()` between timeout and any other type of error could be improved, which I did.

DDvO: My goodness, adding the error queue entry lead to several SSL test failures.
There must be something pretty fishy in the SSL state machine.
I'll transform this to an independent issue.
So for the mem BIO not giving a suitable error on unsupported `BIO_C_DO_STATE_MACHINE` the workaround I just introduced in `BIO_connect_retry()` applies.

DDvO:
@mattcaswell, rebased and new change requests fulfilled.

crypto/bio/b_sock.c
mattcaswell: So, I'm a little unclear, is the result of this issue that we require BIOs to be capable of handling `BIO_C_DO_STATE_MACHINE` for us to be able to use the HTTP client with them?

DDvO: An HTTP client needs a BIO that can be used for connecting over a network,
and AFAICS this boils down to supporting at least `BIO_C_DO_STATE_MACHINE`.

If this is not the case the HTTP functions will yield `BIO_R_CONNECT_ERROR` (or any more specific error message like `BIO_R_UNSUPPORTED_METHOD` if the respective BIO implementation gives such error detail, which currently is not the case for mem BIOs due to #10955).

DDvO: Shall I make explicit in the documentation that if a connection BIO is provided to the HTTP client it naturally needs to support a networked connection?

mattcaswell: I'm wondering  if there is anything we can do to just ignore the error if the `BIO_connnect_retry()` call fails? Failing that I guess we'll have to document the restriction

DDvO: I do not understand what you mean here. 
We do not and cannot ignore any error returned by `BIO_connect_retry()` - or why/when do you think this would make sense?
We usually cannot ignore errors that the called function `BIO_do_connect()` (and thus errors by the underlying BIO) may return, except where a retry is due for some reason. 

The issue mentioned above regarding non-supported `BIO_C_DO_STATE_MACHINE` only affects the exact error detail provided: ideally the underlying BIO provides a to-the-point error entry such as `BIO_R_UNSUPPORTED_METHOD`, but if it returns `0` (failure) without pushing such an entry the enhancement I added to `BIO_connect_retry()` pushes the more general entry `BIO_R_CONNECT_ERROR` as a workaround. In any case it propagates the negative outcome of  `BIO_do_connect()`.

mattcaswell: Well, fundamentally I see no reason why someone should not be able to use a mem bio (or any other "non network" BIO) to do HTTP. For example, often people will use a mem bio as an intermediate step before pushing on to some underlying network connection that is unknown to OpenSSL. I'm just trying to see if there is a way to not force restrictions on what type of BIO can be used. I don't think we do that anywhere else and it detracts significantly from the power of a BIO.

DDvO: Ah, now I understand!
Can the same (mem or other non-network) BIO be used at first for writing (HTTP request) data to and then for reading (HTTP response) data from?

mattcaswell: Hmmm. Well actually you often have a different BIO for reading and writing. The sslapitest does this with clients and servers communicating entirely via mem bios. There is one mem BIO for the "write" side on the client - and that is used as the "read" side on the server. Similarly there is another mem BIO on the "read" side for the client and that is used as the "write" side on the server.

mattcaswell: You can also use a "bio pair" which probably enables you to do this with a single BIO.

DDvO: This makes more sense to me.
To support fully flexible BIO-based HTTP, I can add optional `wbio` and `rbio` parameters.
Shall I also keep the recently added `cbio` parameter, which may be used to provide a custom connection BIO which is then also used for reading and writing data (such that  `wbio` and `rbio` are ignored in this case)?

DDvO: My last response overlapped with your most recent one.

DDvO: Yes, a BIO pair sounds even better.

DDvO: Hmm, I've meanwhile had a look at the documentation of BIO pairs: https://www.openssl.org/docs/manmaster/man3/BIO_new_bio_pair.html which I find hard to understand.
I've also searched for uses of BIO pairs within OpenSSL and found just one, namely in an apparently outdated test: `test/ssltest_old.c`.
So I think it's better to keep away from that and instead follow the model you mentioned before, namely to allow to give two BIOs: one for writing requests and one to read responses from.

DDvO: I've replaced the optional `cbio` parameter by the more flexible optional `wbio` and `rbio` parameters, which was rather straightforward, and updated the docs accordingly.

This was the perfect moment to add also an HTTP client unit test since now mem BIOs can be used to pseudo-connect to an internal mock HTTP server. The tests cover nearly all client code, including GET redirections. This way I discovered and fixed an issue with server-relative redirection locations.

DDvO:
@mattcaswell, all change requests handled, PR rebased, and even clang made happy again :-)

apps/s_server.c
```diff
@@ -1387,10 +1387,9 @@ int s_server_main(int argc, char *argv[])
         case OPT_STATUS_URL:
 #ifndef OPENSSL_NO_OCSP
             s_tlsextstatus = 1;
-            if (!OCSP_parse_url(opt_arg(),
-                                &tlscstatp.host,
-                                &tlscstatp.port,
-                                &tlscstatp.path, &tlscstatp.use_ssl)) {
+            if (!!OSSL_HTTP_parse_url(opt_arg(),
```
mattcaswell: I don't think you meant `!!` here

crypto/http/http_client.c
```diff
@@ -0,0 +1,1238 @@
...
+ * We need to obtain the numeric code and (optional) informational message.
+ */
+
+static int parse_http_line1(char *line)
+{
+    int retcode;
+    char *code, *reason, *end;
+
+    /* Skip to first whitespace (past protocol info) */
+    for (code = line; *code && !ossl_isspace(*code); code++)
```
mattcaswell: `*code != '\0'`

crypto/http/http_client.c
```diff
@@ -0,0 +1,1238 @@
...
+                    return 0;
+            }
+        }
+
+        /* Look for blank line: end of headers */
+        for (p = rctx->iobuf; *p; p++) {
+            if ((*p != '\r') && (*p != '\n'))
+                break;
+        }
+        if (*p) /* not end of headers */
```
mattcaswell: `*p != '\0'`

crypto/http/http_client.c
```diff
@@ -0,0 +1,1238 @@
...
+            goto next_io;
+
+        /* Check it is an ASN1 SEQUENCE */
+        if (*p++ != (V_ASN1_SEQUENCE | V_ASN1_CONSTRUCTED)) {
+            HTTPerr(0, HTTP_R_MISSING_ASN1_ENCODING);
+            return 0;
+        }
+
+        /* Check out length field */
+        if (*p & 0x80) {
```
mattcaswell: `*p & 0x080 != 0`

crypto/http/http_client.c
```diff
@@ -0,0 +1,1238 @@
...
+            /*
+             * If MSB set on initial length octet we can now always read 6
+             * octets: make sure we have them.
+             */
+            if (n < 6)
+                goto next_io;
+            n = *p & 0x7F;
+            /* Not NDEF or excessive length */
+            if (!n || (n > 4)) {
```
mattcaswell: `n == 0 || n > 4`

crypto/http/http_client.c
```diff
@@ -0,0 +1,1238 @@
...
+        *redirection_url = NULL; /* do this beforehand to prevent dbl free */
+
+    if (use_ssl && bio_update_fn == NULL) {
+        HTTPerr(0, HTTP_R_TLS_NOT_ENABLED);
+        return NULL;
+    }
+    if (rbio != NULL && (wbio == NULL || bio_update_fn != NULL)) {
+        HTTPerr(0, ERR_R_PASSED_INVALID_ARGUMENT);
+        return NULL;
+    }
```
mattcaswell: What about the case where `wbio != NULL` and `rbio == NULL`?

doc/man3/OSSL_HTTP_transfer.pod
```diff
@@ -0,0 +1,210 @@
...
+                                    int connect, int detail);
+ BIO *OSSL_HTTP_get(const char *url, const char *proxy, const char *proxy_port,
+                    BIO *wbio, BIO *rbio,
+                    OSSL_HTTP_bio_cb_t bio_update_fn, void *arg,
+                    const STACK_OF(CONF_VALUE) *headers,
+                    int maxline, unsigned long max_resp_len, int timeout,
+                    const char *expected_content_type, int expect_asn1);
+ ASN1_VALUE *OSSL_HTTP_get_asn1(const char *url,
+                                const char *proxy, const char *proxy_port,
+                                BIO *wbio, BIO *rbio,
```
mattcaswell: I don't know how important it is, but I observe that with `SSL_set_bio` we pass the bios in the order rbio, wbio. But here we do it the other way around. Should we consistently do it the same?

test/http_test.c
```diff
@@ -0,0 +1,175 @@
...
+
+void cleanup_tests(void)
+{
+    X509_free(x509);
+}
+
+int setup_tests(void)
+{
+    x509_it = ASN1_ITEM_rptr(X509);
+    if (!TEST_ptr((x509 = load_pem_cert(test_get_argument(0)))))
```
mattcaswell: Please add this prior to the `test_get_argument(0)` call (see 8d242823ed):

````
    if (!test_skip_common_options()) {
        TEST_error("Error parsing test options\n");
        return 0;
    }
````

File: crypto/http/http_client.c (hunk 40 above)
Comment (DDvO) [reply crypto/http/http_client.c]: As (newly ) documented, this selects the case that `wbio` is used for connecting, writing, and reading. 
I might rename the `wbio` parameter to `bio` (or `cwrbio`) - what you think?

File: doc/man3/OSSL_HTTP_transfer.pod (hunk 41 above)
Comment (DDvO) [reply doc/man3/OSSL_HTTP_transfer.pod]: While I see the consistency point, here I suggest to keep the current order because 
* the client uses the first mentioned BIO (whose name may change, see my last comment) first: the client uses it for connecting and/or for writing requests before optionally using `rbio` for reading responses
* if the first BIO is NULL the second must be NULL, too (but not the other way round), so the first BIO is used in more use cases.

File: apps/s_server.c (hunk 35 above)
Comment (DDvO) [reply apps/s_server.c]: Oops, of course - this was a pretty bad copy&paste mistake. Fixed.

File: crypto/http/http_client.c (hunk 37 above)
Comment (DDvO) [reply crypto/http/http_client.c]: This comparison coding style nit, as well as the further ones you mentioned plus a couple of similar further ones I've just found myself, had been inherited from `crypto/ocsp/ocsp_ht.c`.
I've fixed all of them.

File: crypto/http/http_client.c (hunk 38 above)
Comment (DDvO) [reply crypto/http/http_client.c]: fixed

File: crypto/http/http_client.c (hunk 39 above)
Comment (DDvO) [reply crypto/http/http_client.c]: fixed

File: crypto/http/http_client.c (hunk 36 above)
Comment (DDvO) [reply crypto/http/http_client.c]: fixed

File: crypto/http/http_client.c (hunk 40 above)
Comment (DDvO) [reply crypto/http/http_client.c]: I've changed the `wbio` parameter name to `bio` because it may be used more generally.

File: doc/man3/OSSL_HTTP_transfer.pod (hunk 41 above)
Comment (DDvO) [reply doc/man3/OSSL_HTTP_transfer.pod]: Due to renaming `wbio` to  `bio`, which indicates that it may be of more broad use, in my view the consistency point has become even weaker. 
So in my view the current order of parameters is adequate.

File: test/http_test.c (hunk 42 above)
Comment (DDvO) [reply test/http_test.c]: Done

**Comment by DDvO** [comment]
> Did another review pass. Mostly minor issues. 

Thanks @mattcaswell. I've handled all of them.

> Please could you also clean up the commit history to be what you actually intend to push?

Sure. I've squashed all commits into just two ones:
* the extension of the BIO socket API with timeout-related functions
* the new HTTP client, making use of the extended socket features

**Comment by DDvO** [comment]
@mattcaswell, thanks a lot for your swift review of this non-trivial PR!
Thanks also @iamamoose for adding the 'ready to merge' tag. 
So it looks like I can try doing my first merge in my new committer's career :)

In preparation I realized that I had to rebase this PR and fix the usual trivial merge conflict on `util/libcrypto.num` because in the meantime some new functions have been exported on master.
Is there a re-confirmation of the approval needed after such straightforward rebasing?

I've meanwhile installed the tools from openssl-git@git.openssl.org:tools.git .
BTW, in addition to the instructions at `OpenSSL-Query/README.md` I had to use 
```
sudo apt install libplack-perl libdancer2-perl
```
Then I tried `addrev --list` and found that there is no entry (yet) for me. 
Is it needed for pushing my own PRs, and if so, how to add my data there?

Then I tried `addrev --prnum=10667 -2 matt @DDvO` but got the following `gitaddrev` error:
```
Rewrite bebd8c78255aee55d2c7f6128fed711217efa636 (1/2) (0 seconds passed, remaining 0 predicted)    Reviewers without CLA: @DDvO
msg filter failed: gitaddrev --prnum=10667 --reviewer=matt --reviewer=@DDvO --myemail=David.von.Oheimb@siemens.com
```
So the tool apparently complains that CLA is missing somehow, although at least as a contributer both my ICLA and a Siemens CCLA are already on file. How to solve this?

Once this is solved and the `addrev` has succeeded, can I then simply issue the following command?
```
ghmerge --nomerge 10667 DDvO
```

**Comment by DDvO** [comment]
Ah, I should better have used `addrev --prnum=10667 -2 matt DDvO master..` but then get
```
Rewrite c0d47492a77e67ce49cf3c538bdaf82ca90367cb (1/292) (0 seconds passed, remaining 0 predicted)    Unknown reviewers: David.von.Oheimb@siemens.com
msg filter failed: gitaddrev --prnum=10667 --reviewer=matt --myemail=David.von.Oheimb@siemens.com
```

mattcaswell:
> In preparation I realized that I had to rebase this PR and fix the usual trivial merge conflict on util/libcrypto.num because in the meantime some new functions have been exported on master.
Is there a re-confirmation of the approval needed after such straightforward rebasing?

No. For a trivial rebase like that there is no need to seek reconfirmation.

> Then I tried addrev --prnum=10667 -2 matt @DDvO but got the following gitaddrev error:

This seems to be because our database had your personal email address as your "reviewer" identity - but apparently your personal email address isn't listed on your Individual CLA (only your work email address is listed there). I've changed it to use your work email address for now - so try it again.

> Once this is solved and the addrev has succeeded, can I then simply issue the following command?

Actually I never use ghmerge so I can't really advise on its proper use. I think most people do it "manually", i.e. cherry-pick the commits onto a fresh master and then just `git push` (assuming origin is set to the OpenSSL git repo).

DDvO:
> For a trivial rebase like that there is no need to seek reconfirmation.

Good to hear.

> This seems to be because our database had your personal email address as your "reviewer" identity - but apparently your personal email address isn't listed on your Individual CLA (only your work email address is listed there). I've changed it to use your work email address for now - so try it again.

Yes, now `addrev --prnum=10667 -2 matt @DDvO` works, thanks :)

> Actually I never use ghmerge so I can't really advise on its proper use. I think most people do it "manually", i.e. cherry-pick the commits onto a fresh master and then just `git push` (assuming origin is set to the OpenSSL git repo).

I see - so I'll just try the official `ghmerge` tool, and if I have trouble with it I'll move on to the manual approach.

After the rebase Travis failed apparently just because it stalled on one build.
So to be on the safe side I've just force-pushed the branch with the reviewer and PR annotations to force a CI rebuild. I'll wait until the CI has succeeded again and try doing the push tomorrow...

DDvO:
After tweaking the `ghmerge` tool a little, I've successfully pushed the PR.
Thanks a lot @mattcaswell !

mattcaswell:
> After tweaking the ghmerge tool a little, I've successfully pushed the PR.

Great!! Congratulations on your first commit!

paulidale:
Let us know the tweaks to the `ghmerge` script.  I've got a couple of things I'd like to add eventually.

DDvO:
> Let us know the tweaks to the `ghmerge` script. I've got a couple of things I'd like to add eventually.

I was about to share the diff here, but apparently cannot add an attachment to this thread.
Can I do a PR to openssl-git@git.openssl.org:tools.git instead?
And how is the review process for adaptations to the tools repo?

DDvO:
I'd also have a few additions / corrections to two README files there.

mattcaswell:
> Can I do a PR to openssl-git@git.openssl.org:tools.git instead?

Yes, that's the preferred way to do things.

> And how is the review process for adaptations to the tools repo?

At the moment its an OMC owned repo, so reviews/pushes can only be done by OMC members.

mattcaswell:
> At the moment its an OMC owned repo, so reviews/pushes can only be done by OMC members.

BTW, I see no reason why tools like ghmerge and addrev shouldn't be owned by OTC rather than OMC...but at the moment they're not.

paulidale:
Anyone can make a pull request against the tools repository.
The review process requires two OMC approvers currently.

paulidale:
> BTW, I see no reason why tools like ghmerge and addrev shouldn't be owned by OTC rather than OMC...but at the moment they're not.

Agreed.

DDvO:
I've meanwhile added two respective PRs:
* https://github.com/openssl/tools/pull/58
* https://github.com/openssl/tools/pull/59

mattcaswell:
Oops. Looks like this broke Travis:

https://travis-ci.org/openssl/openssl/jobs/648882003

This is actually the new travis job introduced by #9982. Since that job went into master slightly after this PR went in, but this PR was committed after the last travis run for #9982 was performed this problem didn't show up in either PRs.

DDvO:
I'm trying to reproduce and fix this...

DDvO:
I found this happens due to configuration options `no-cmp` and `no-ocsp`.
The former was trivial to fix, now I'm looking for a good solution for the latter...

DDvO:
Luckily, also the fix for `no-ocsp` was not difficult.

DDvO:
Fixup PR #11058 has been merged.

PR #10673, opened 2019-12-21
improve 'typedef' patterns of find-doc-nits

DDvO:
`make doc-nits` meanwhile sometimes yields hard-to-understand/misleading and actually wrong error messages such as `* missing from NAME section` for a couple of doc files.
It turns out that this happens when a function pointer `typedef` has a space character before the opening parenthesis of the argument list, as in
```
 typedef OSSL_CMP_MSG *(*OSSL_cmp_transfer_cb_t) (OSSL_CMP_CTX *ctx,
                                                  const OSSL_CMP_MSG *req);
```
because the pattern in `find-doc-nits` that should recognize such typedefs is too restrictive.
I've relaxed it accordingly by inserting `\s*`
and on this occasion removed two superfluous occurrences of `.*` at pattern end.

Now `find-doc-nits` recognizes a couple of further typedefs in doc files, which has the effect that some omissions on those declared types get newly reported. I've fixed them in this little PR as well.


richsalz:
That's a good catch but I thnk it should be an error with a "space before argument list" message.

DDvO:
Thanks for your immediate feedback.

Is it the job of `find-doc-nits` to detect and report also code formatting issues?

BTW, I've just realized that the new format checker https://github.com/openssl/openssl/pull/10363 does not yet detect such extra whitespace (in cases where there is a `)` after the function name).

richsalz:
> Is it the job of find-doc-nits to detect and report also code formatting issues?

It's the job of the script to find nits and inconsistencies in the docs.  The one you found seems like something to fix, IMO.  Others on the team (who vote!) may disagree.

DDvO:
I meanwhile realize that `find-doc-nits` already does some checks of this sort.
So I've made it report also `function typedef has space before arg list`
and fixed the new findings in the respective doc files.

DDvO:
ping second review

DDvO:
@committers, we still need a 2nd review here. Should not take much time
as it's just a few lines of changed Perl script, some whitespace changes in C code, 
and a couple of new lines in doc and meta files.
@richsalz, since you already gave comments, would it make sense if you do this?

levitte:
Side note: why were some types named `OSSL_cmp_*` rather than `OSSL_CMP_*`?  It's a bit of a nit, but it does look curious...

DDvO:
Thanks @levitte 

Yes, the 'upper-case-ness' of some type names are a bit inconsistent also in my view meanwhile.

As far as I recall this was because once we were asked to add `OSSL_` prefixes to all exported symbols and lacking complete/detailed instructions how to do this we did not adapt the existing name prefix. 
This BTW forms part of the missing detail in https://www.openssl.org/policies/codingstyle.html
on which I recently filed issue #10725.

richsalz:
@ddvo, I'm not a member of the project any more, just a contributor like you :)

DDvO:
Pleased to see that that has been marked 'ready to merge' a couple of days ago.
Can someone please do this now?

mattcaswell:
Pushed. Thanks!

PR #9624, opened 2019-08-18
Fix sha512_block_data_order_avx2 backtrace info

bernd-edlinger:
Add a frame pointer that is always reachable at rsp-8, the red zone.
Well, red zone it is not guaranteed to be preserved on windows but
for purpose of restoring the return frame we have the original copy.
When no free register is available I use a pushq / lea sequence.
When a free scratch register is available I use two mov insns.
I don't know if two mov are more expensive or not.

PR #10674, opened 2019-12-21
Fix aesni_cbc_sha256_enc_avx2 backtrace info

bernd-edlinger:
This fixes a crash in the unwinder when the function aesni_cbc_sha256_enc_avx2
is interrupted.

It is reproducibly with `openssl speed -evp aes-128-cbc-hmac-sha256` and this patch
```
diff --git a/apps/speed.c b/apps/speed.c
index fa306c9..19f660b 100644
--- a/apps/speed.c
+++ b/apps/speed.c
@@ -8,6 +8,7 @@
  */
 
+#include <unwind.h>
 #undef SECONDS
 #define SECONDS                 3
 #define RSA_SECONDS             10
@@ -163,8 +164,44 @@ static const int aead_lengths_list[] = {
 
 #ifdef SIGALRM
 
+struct ctxt
+{
+  void **res;
+  int cnt;
+  int max;
+};
+
+_Unwind_Reason_Code bt (struct _Unwind_Context *c, void *x)
+{
+  struct ctxt *ct = x;
+  if (ct->cnt < ct->max)
+    ct->res[ct->cnt++] = (void*)_Unwind_GetIP(c);
+  return _URC_NO_REASON;
+}
+
 static void alarmed(int sig)
 {
+    void *res[20];
+    struct ctxt ct;
+    int i;
+    ct.res = res;
+    ct.cnt = 0;
+    ct.max = 20;
+    write(1,"alarmed\n", 8);
+    _Unwind_Backtrace(bt, &ct);
+    for (i = 0; i < ct.cnt; i++)
+    {
+       unsigned long p = (unsigned long) res[i];
+       char q[18];
+       int j = sizeof(q);
+       q[--j] = 10;
+       do {
+          q[--j] = (p & 15) < 10 ? '0' + (p & 15) : 'A' + (p & 15) - 10;
+          p >>= 4;
+       } while(p && j > 0);
+       write(1, q+j, sizeof(q)-j);
+    }
+    write(1,"finally\n", 8);
     signal(SIGALRM, alarmed);
     run = 0;
 }
```

The change in aesni_cbc_sha256_enc_shaext is trivial, but
the other one is not.  But it is almost 1:1 identical to #9624

### DIFF
--- a/crypto/aes/asm/aesni-sha256-x86_64.pl
+++ b/crypto/aes/asm/aesni-sha256-x86_64.pl
@@ -1088,7 +1088,23 @@ $code.=<<___;
 	vmovdqa	$t0,0x00(%rsp)
 	xor	$a1,$a1
 	vmovdqa	$t1,0x20(%rsp)
+___
+$code.=<<___ if (!$win64);
+# temporarily use %rsi as frame pointer
+        mov     $_rsp,%rsi
+.cfi_def_cfa    %rsi,8
+___
+$code.=<<___;
 	lea	-$PUSH8(%rsp),%rsp
+___
+$code.=<<___ if (!$win64);
+# the frame info is at $_rsp, but the stack is moving...
+# so a second frame pointer is saved at -8(%rsp)
+# that is in the red zone
+        mov     %rsi,-8(%rsp)
+.cfi_cfa_expression     %rsp-8,deref,+8
+___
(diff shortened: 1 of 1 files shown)


bernd-edlinger:
Merged to master and 1.1.1 branch.  Thanks!

PR #10678, opened 2019-12-22
Add some missing cfi frame info in poly1305-x86_64.pl

bernd-edlinger:
##### Checklist

### DIFF
--- a/crypto/poly1305/asm/poly1305-x86_64.pl
+++ b/crypto/poly1305/asm/poly1305-x86_64.pl
@@ -170,6 +170,7 @@ $code.=<<___;
 .type	poly1305_init,\@function,3
 .align	32
 poly1305_init:
+.cfi_startproc
 	xor	%rax,%rax
 	mov	%rax,0($ctx)		# initialize hash value
 	mov	%rax,8($ctx)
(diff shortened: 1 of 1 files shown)


crypto/poly1305/asm/poly1305-x86_64.pl
```diff
@@ -344,15 +348,18 @@ sub poly1305_iteration {
...
+.cfi_endproc
 .size	__poly1305_block,.-__poly1305_block
 
 .type	__poly1305_init_avx,\@abi-omnipotent
 .align	32
 __poly1305_init_avx:
+.cfi_startproc
 	mov	$r0,$h0
 	mov	$r1,$h1
 	xor	$h2,$h2
```
kroeckx: Does this function need to CFA?

kroeckx: Doesn't look like it.

bernd-edlinger: It is called from a CFA instrumented function, so the runtime stack walking
will not be able to continue the stack walk.

kroeckx: I think that since you start a new CFI, the CFA defaults to the stack pointer again.

bernd-edlinger: Right, and gdb's bt command works just fine,
but runtime-stack walks don't, and tools using that
for performance/coverage metrics don't work right,
when that .cfi is missing, the runtime-stack walk
assumes it can't unwind beyond this frame.

bernd-edlinger:
Merged to master and 1.1.1 branch.  Thanks!

PR #10680, opened 2019-12-22
Fix a race condition in the speed command

bernd-edlinger:
The timer alarm sets run = 0, while the benchmark
does run = 1 in the initialization code.  That is
a race condition, if the timer goes off too early
the benchmark runs forever.

##### Checklist

### DIFF
--- a/apps/speed.c
+++ b/apps/speed.c
@@ -1273,8 +1273,6 @@ static int run_benchmark(int async_jobs,
     OSSL_ASYNC_FD job_fd = 0;
     size_t num_job_fds = 0;
 
-    run = 1;
-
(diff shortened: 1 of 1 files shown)


bernd-edlinger:
Discovered this, while testing the last patch.
reproduce:
`gdb --args openssl speed -evp rc4-hmac-md5`
and set a breakpoint on RC4_set_key.
Wait 3 seconds, there, continue, see speed run forever.
Funny thing: thread sanitizer does not detect this race condition.

bernd-edlinger:
Pushed to master and 1.1.1 branch.  Thanks!

PR #10685, opened 2019-12-23
Fix aes_platform.h missing issue

zorrorffm:
This came from commit "cc731bc3f". 'aes_platform.h' is supposed to replace
'ciphermode_platform.h'. Without aes_platform.h included, AES_GCM_ASM is
undefined so that hardware accelerated AES implementation won't be used.
Affects x86 and arm64.

##### Checklist

### DIFF
--- a/providers/implementations/ciphers/ciphercommon_gcm_hw.c
+++ b/providers/implementations/ciphers/ciphercommon_gcm_hw.c
@@ -7,9 +7,14 @@
  * https://www.openssl.org/source/license.html
  */
 
+/*
+ * This file uses the low level AES functions (which are deprecated for
+ * non-internal use) in order to implement provider AES ciphers.
+ */
+#include "internal/deprecated.h"
 #include "prov/ciphercommon.h"
 #include "prov/ciphercommon_gcm.h"
-
+#include "crypto/aes_platform.h"
 
 int gcm_setiv(PROV_GCM_CTX *ctx, const unsigned char *iv, size_t ivlen)
 {


t8m:
Shouldn't it be rather #included in ciphercommon_gcm_hw.c?

zorrorffm:
> 
> 
> Shouldn't it be rather #included in ciphercommon_gcm_hw.c?

Thanks, I correct it now.

t8m:
Hmm, the Travis failure is relevant. You will need something like:

`/*
 * This file uses the low level AES functions (which are deprecated for
 * non-internal use) in order to implement the AES ciphers.
 */

`
added to the aes_platform.h

t8m:
Or maybe rather put it into ciphercommon_gcm_hw.c

zorrorffm:
> 
> 
> Or maybe rather put it into ciphercommon_gcm_hw.c

Done.

t8m:
However thinking about it more it would be probably even more appropriate to just rename the `gcm_cipher_update` function to `aes_gcm_cipher_update` and move it to `cipher_aes_gcm_hw.c`. Let's see what other reviewers think.

levitte:
We should still have a generic `gcm_cipher_update`, shouldn't we?

t8m:
We could but currently it is not used by anything else than AES AFAIK. The way it is coded currently it is not generic anyway.

t8m:
Actually the `aria_cipher_update` from `cipher_aria_gcm_hw.c` could be the generic implementation.

zorrorffm:
Close this PR because #10783 solves this issue and is more reasonable.

PR #9294, opened 2019-07-02
Use leak sanitizer instead of internal mdebug to check for memory leaks

levitte:
The leak sanitizer gives better reports (complete stack traces) and
works as a wrapper around the application instead of relying on
cooperative enabling and disabling calls (which are too easy to get
unbalanced).

Related to #8322

Issue #10648, opened 2019-12-18
windows10 - `nmake install` fails with fatal error u1077: "pod2html": return-code "0xff"


ML0401:
I am trying to install openssl on windows 10, therefor i have already installed and downloaded:

- Active State Perl 5.26.3 (MSWin32_x64)
- visual studio 17 and 19 (only one required)
- openssl-master folder

and executed the following steps via admin and command prompt:
> cd \my path to Perl folder\
> perl \my path to Configure file from openssl-master\ no-asm VC-WIN32
> nmake (no errors)
> nmake test (test pass, no errors)
> nmake install gets me the following error message:

The command "href" is wrong or could not be found 
NMAKE: fatal error u1077: "pod2html": return-code "0xff"

I have already installed openssl the same way on windows7 without any complications. Am I missing something or mixing up the different version? VC-WIN64A and VC-WIN64I are not working as well. Thanks in advance for any kind of help.

mattcaswell:
> NMAKE: fatal error u1077: "opd2html": return-code "0xff"

Is this a typo? Does this actually say "pod2html", or is this the real error?

It seems that it is failing for some reason during installation of the documentation. If you don't care about that a workaround might be to only install the software itself, i.e.

````
nmake install_sw
````

**Comment by ML0401** [comment]
Yes, you are right. I changed it to “pod2html“ and will check if it is working. Thank you!

**Comment by ML0401** [comment]
> > NMAKE: fatal error u1077: "opd2html": return-code "0xff"
> 
> Is this a typo? Does this actually say "pod2html", or is this the real error?
> 
> It seems that it is failing for some reason during installation of the documentation. If you don't care about that a workaround might be to only install the software itself, i.e.
> 
> ```
> nmake install_sw
> ```

**Comment by ML0401** [comment]
I changed it. You were right it is “pod2html”. I will try your solution. I need OpenSSL to create certificates, so I hope this will work.

**Comment by mspncp** [comment]
I can confirm this bug. I just encountered it on my notebook which is also running Windows 10.

```
...
*** Installing runtime programs
Copying: apps//openssl.exe to /dev/openssl/bin/openssl.exe
Copying: apps//openssl.pdb to /dev/openssl/bin/openssl.pdb
Copying: tools//c_rehash.pl to /dev/openssl/bin/c_rehash.pl
Copying: ./apps/openssl.cnf to /dev/openssl/ssl/openssl.cnf.dist
[...]

PR #10572, opened 2019-12-04
Deprecate most of debug-memory

richsalz:
The leak-checking (and backtrace option, on some platforms) provided
by crypto-mdebug and crypto-mdebug-backtrace have been mostly neutered;
only the "make malloc fail" capability remains.  OpenSSL recommends using
clang leak-detection instead.

The OPENSSL_DEBUG_MEMORY environment variable is no longer used.
CRYPTO_mem_ctrl(), CRYPTO_set_mem_debug(), CRYPTO_mem_leaks(),
CRYPTO_mem_leaks_fp() and CRYPTO_mem_leaks_cb() return a failure code.
CRYPTO_mem_debug_{malloc,realloc,free}() have been removed.  All of the
above are now deprecated.

PR #10327, opened 2019-11-01
Appveyor: update to Visual Studio 2017.

FdaSilvaYY:
Default image was currently "Visual Studio 2015"

See https://www.appveyor.com/docs/windows-images-software/
and https://github.com/Microsoft/vssetup.powershell

##### Checklist

- [ ] documentation is added or updated
- [ ] tests are added or updated

PR #10689, opened 2019-12-23
Fix Travis

mattcaswell:
This PR intends to make Travis go green for master builds. Once this works a subsequent PR will address 1.1.1 builds.

This PR does 3 things:

- Temporarily reverts the changes done in #9294 which switched from using the internal mdebug to leak sanitizer in Travis. Unfortunately it looks like this change needs more work because it is causing Travis builds to fail.
- Disables test_conf in cross compiled builds. In Travis we build mingw targets on Linux and then run the tests via wine. Unfortunately this causes test_conf to fail because of different line break conventions between platforms.
- Disables the pyca-cryptography and krb5 external tests in Travis

For the external tests both the pyca-cryptography and krb5 external tests having been failing for a long time. In the krb5 case there are test failures that I cannot reproduce locally. In the pyca-cryptography case it seems that upstream will need to make some changes in order to adapt to 3.0 (specifically the move of some symbols from macros to functions has caused problems).

Marked as WIP until Travis actually goes green.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -113,9 +113,10 @@ matrix:
         - os: linux
           compiler: clang
           env:  EXTENDED_TEST="yes" CONFIG_OPTS="no-asm enable-ubsan enable-rc5 enable-md2 enable-ssl3 enable-ssl3-method enable-nextprotoneg no-shared enable-buildtest-c++ -fno-sanitize=alignment -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -D__NO_STRING_INLINES -Wno-unused-command-line-argument" CXX="clang++"
-        - os: linux
-          compiler: clang
-          env: EXTENDED_TEST="yes" CONFIG_OPTS="no-asm enable-asan enable-rc5 enable-md2 no-shared -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -D__NO_STRING_INLINES -Wno-unused-command-line-argument"
+#Temporarily removed because memleak test fails for unknown reasons in this build. Other builds with memleak testing ("-fsanitize=address") are passing.
+#        - os: linux
+#          compiler: clang
+#          env: EXTENDED_TEST="yes" CONFIG_OPTS="no-asm enable-asan enable-rc5 enable-md2 no-shared -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -D__NO_STRING_INLINES -Wno-unused-command-line-argument"
         - os: linux
           compiler: gcc
           env: EXTENDED_TEST="yes" CONFIG_OPTS="--debug no-asm enable-ubsan enable-rc5 enable-md2 enable-buildtest-c++ -DPEDANTIC" OPENSSL_TEST_RAND_ORDER=0
--- a/.travis.yml
+++ b/.travis.yml
@@ -105,8 +105,8 @@ matrix:
                   packages:
                       - golang-1.10
           compiler: gcc
-          # External tests pyca-cryptography and krb5 temporarily disabled due to long term travis failures
-          env:  EXTENDED_TEST="yes" CONFIG_OPTS="--debug enable-ssl3 enable-ssl3-method enable-weak-ssl-ciphers enable-external-tests enable-buildtest-c++" BORINGSSL_TESTS="yes" CXX="g++" TESTS=test_external_boringssl
+          # External test pyca-cryptography temporarily disabled due to long term travis failures
+          env:  EXTENDED_TEST="yes" CONFIG_OPTS="--debug enable-ssl3 enable-ssl3-method enable-weak-ssl-ciphers enable-external-tests enable-buildtest-c++" BORINGSSL_TESTS="yes" CXX="g++" TESTS="test_external_boringssl test_external_krb5"
         - os: linux
           compiler: clang
           env: EXTENDED_TEST="yes" CONFIG_OPTS="enable-msan disable-afalgeng -D__NO_STRING_INLINES -Wno-unused-command-line-argument"
--- a/.travis.yml
+++ b/.travis.yml
@@ -105,7 +105,8 @@ matrix:
                   packages:
                       - golang-1.10
           compiler: gcc
-          env:  EXTENDED_TEST="yes" CONFIG_OPTS="--debug enable-ssl3 enable-ssl3-method enable-weak-ssl-ciphers enable-external-tests enable-buildtest-c++" BORINGSSL_TESTS="yes" CXX="g++" TESTS=95
+          # External tests pyca-cryptography and krb5 temporarily disabled due to long term travis failures
+          env:  EXTENDED_TEST="yes" CONFIG_OPTS="--debug enable-ssl3 enable-ssl3-method enable-weak-ssl-ciphers enable-external-tests enable-buildtest-c++" BORINGSSL_TESTS="yes" CXX="g++" TESTS=test_external_boringssl
         - os: linux
           compiler: clang
           env: EXTENDED_TEST="yes" CONFIG_OPTS="enable-msan disable-afalgeng -D__NO_STRING_INLINES -Wno-unused-command-line-argument"
--- a/test/recipes/04-test_conf.t
+++ b/test/recipes/04-test_conf.t
@@ -10,6 +10,7 @@ use strict;
 use warnings;
 
 use OpenSSL::Test qw(:DEFAULT data_file);
+use OpenSSL::Test::Utils;
 use File::Compare qw(compare_text);
 
 setup('test_conf');
@@ -18,6 +19,9 @@ my %input_result = (
     'dollarid_on.conf'  => 'dollarid_on.txt',
     'dollarid_off.conf' => 'dollarid_off.txt',
 );
+
+plan skip_all => 'This is unsupported for cross compiled configurations'
+    if config('CROSS_COMPILE');
 
 plan tests => 2 * scalar(keys %input_result);
 


mattcaswell:
Building this with "[extended tests]"

mattcaswell:
All the builds that were red are now green. Unfortunately new builds have turned red. It seems the reversion of #9294 had unexpected consequences. That change was further built on by #10572 which is quite a large change. #9294 was quite small and self contained so reverting seemed like a viable option. But #10572 complicates the picture.

For now I've removed the #9294 revert commits from this PR. I'll have a think to see if there is some other workaround we can temporarily employ.

richsalz:
It's unfortunate that the builds weren't consistently passing before as I am sure that obscured the errors.

Where can we see a detailed example of the travis failures that this is trying to address?

mattcaswell:
This is a typical example of a master build:

https://travis-ci.org/openssl/openssl/builds/628227806

You'll see there are four failed build jobs. One is failed job is due to the pyca/krb5 failures, one is due to the memleak test failing, and two are due to test_conf failing.

As of now in this PR (with one build still running) all the travis tests are geen, except for one due to the memleak test failure.

richsalz:
So the simplest work-around, for now, is to just remove the memleak test, isn't it?

mattcaswell:
> So the simplest work-around, for now, is to just remove the memleak test, isn't it?

That might be the only answer - although I am worried. Does this shows that *all* memleaks will go undetected at the moment - if so that is *bad*.

mattcaswell:
> That might be the only answer - although I am worried. Does this shows that all memleaks will go undetected at the moment - if so that is bad.

Just pushed a commit that disables that one build.

It is strange because there are other builds that pass "-fsanitize=address" explicitly seem to complete successfully. I have been able to reproduce a local failure in memleak test but *only* if I use gcc, whereas the build that fails is using clang.

richsalz:
>  Does this shows that all memleaks will go undetected at the moment 

No, because the ASAN builds will catch leaks.  The memleak test only shows that ASAN works even if there are no memleaks.

mattcaswell:
> No, because the ASAN builds will catch leaks. The memleak test only shows that ASAN works even if there are no memleaks.

Right - but that was my concern. This memleak test failure showed that ASAN *wasn't* working, i.e. it failed to detect a leak that it should have done.

mattcaswell:
Travis is now green. However it looks like we might be able to keep the krb5 external tests without disabling them based on the test results from #10690, so we might want to wait for that one to be merged first.

richsalz:
I think this is too heavy-handed a change (but it is much better than reverting the world) and would prefer to just disable the one test.  That makes it easier for folks to fix and test the fix. I am bemused about this, given how long the builds have been broken for external tests and the project tolerated that. :)

mspncp:
Since @mattcaswell is on vacation, I rebased this over @kaduk's 3e73f558af600ea068bb2132988c31ddb444e13e without further changes. Let's see whether the CIs turn green :-)

mspncp:
Of course testing only makes sense if the external krb5 tests are re-enabled, see 263abfe1a31f3a234925fef24ed8a8924bca0998  ;-)

mspncp:
Yay! Tests are green, for the first time in a long time. Let's take it out of WIP :-)

image

mspncp:
Since I modified Matt's pr, I would like to see a reconfirmation from @mattcaswell or another committer before marking this pr ready.

.travis.yml
```diff
@@ -105,16 +105,18 @@ matrix:
                   packages:
                       - golang-1.10
           compiler: gcc
-          env:  EXTENDED_TEST="yes" CONFIG_OPTS="--debug enable-ssl3 enable-ssl3-method enable-weak-ssl-ciphers enable-external-tests enable-buildtest-c++" BORINGSSL_TESTS="yes" CXX="g++" TESTS=95
+          # External tests pyca-cryptography and krb5 temporarily disabled due to long term travis failures
```
romen: should this comment be updated given that krb5 tests are not disabled?

romen:
@mspncp can you push an empty commit (to be removed upon merge) to run the `[extended tests]`: looking at the latest Travis log it seems like all tests with `EXTENDED_TESTS=YES` were skipped (duration seems to always be below 2 minutes)

mspncp:
I‘ll push a fixup later this evening.

richsalz:
Congratulations to everyone involved!

Now, how can we prevent this from happening again?  A practice to say "once the build breaks, only PR's that address it are allowed" is something I'd like to see, for example.  (And note I said address, not fix)

.travis.yml
mspncp: Amended my last fixup to adjust the comment and add `[extended tests]`, see e643b8089588097504dcb7ddf5382affb22504bd.

mspncp:
Travis is still green, but AppVeyor fails during installation. Interestingly, this failure did not occur when `[extended tests]` was not specified.

Since the goal of this pull request was to fix travis, the AppVeyor failure should not stop it from being merged IMO. I'll create an issue for the AppVeyor failure later.

```
*** Installing runtime programs
Copying: apps//openssl.exe to ../_install/Program Files/OpenSSL/bin/openssl.exe
Copying: apps//openssl.pdb to ../_install/Program Files/OpenSSL/bin/openssl.pdb
Copying: tools//c_rehash.pl to ../_install/Program Files/OpenSSL/bin/c_rehash.pl
created directory `../_install/Program Files/Common Files'
created directory `../_install/Program Files/Common Files/SSL'
created directory `../_install/Program Files/Common Files/SSL/certs'
created directory `../_install/Program Files/Common Files/SSL/private'
created directory `../_install/Program Files/Common Files/SSL/misc'
Copying: ../apps/openssl.cnf to ../_install/Program Files/Common Files/SSL/openssl.cnf.dist
Copying: ../apps/openssl.cnf to ../_install/Program Files/Common Files/SSL/openssl.cnf
Copying: apps//CA.pl to ../_install/Program Files/Common Files/SSL/misc/CA.pl
Copying: apps//tsget.pl to ../_install/Program Files/Common Files/SSL/misc/tsget.pl
Copying: ../apps/ct_log_list.cnf to ../_install/Program Files/Common Files/SSL/ct_log_list.cnf.dist
Copying: ../apps/ct_log_list.cnf to ../_install/Program Files/Common Files/SSL/ct_log_list.cnf
NMAKE : warning U4006: special macro undefined : '$<"'
	pod2html "--podroot=../doc" --htmldir=..  --podpath=man1:man3:man5:man7 "--infile=" "--title=openssl-asn1parse"  | C:\Perl\bin\perl.exe -pe "s|href=\"http://man\.he\.net/(mand/[^\"]+)(?:\.html)?\"|href=\"../.html|g;"  > doc\html\man1\openssl-asn1parse.html
'href' is not recognized as an internal or external command,
operable program or batch file.
NMAKE : fatal error U1077: 'pod2html' : return code '0xff'
Stop.
Command exited with code 2
```

mspncp:
When merging 7095234b97ddb27ea2d9e8988ecef58a50c3520e, I'll remove the references to the krb5 test from its commit message, changing it to the following:

```
Temporarily disable some Travis external tests

The pyca-cryptography external test has been failing for a long time.
It looks like upstream needs to make some changes to adapt to 3.0.

[extended tests]
```

FdaSilvaYY:
> 
> 
> Travis is still green, but AppVeyor fails during installation. Interestingly, this failure did not occur when `[extended tests]` was not specified.
> 
> Since the goal of this pull request was to fix travis, the AppVeyor failure should not stop it from being merged IMO. I'll create an issue for the AppVeyor failure later.
> 
> ```
> *** Installing runtime programs
...
> Copying: ../apps/ct_log_list.cnf to ../_install/Program Files/Common Files/SSL/ct_log_list.cnf
> NMAKE : warning U4006: special macro undefined : '$<"'
> 	pod2html "--podroot=../doc" --htmldir=..  --podpath=man1:man3:man5:man7 "--infile=" "--title=openssl-asn1parse"  | C:\Perl\bin\perl.exe -pe "s|href=\"http://man\.he\.net/(mand/[^\"]+)(?:\.html)?\"|href=\"../.html|g;"  > doc\html\man1\openssl-asn1parse.html
> 'href' is not recognized as an internal or external command,
> operable program or batch file.
> NMAKE : fatal error U1077: 'pod2html' : return code '0xff'
> Stop.
> Command exited with code 2
> ```

I get the same error a week ago, in my private Travis builds, while working on #10327

mattcaswell:
Apparently I can't approve my own PRs. But anyway, I'm fine with this updated PR. If someone else wants to push - please feel free to go ahead.

richsalz:
Opened issue #10715 to discuss memleak.

mspncp:
Merged to master, thank you everybody!

5f95fbf399 Temporarily disable the explicit enable-asan build
20e6631c73 Temporarily disable some Travis external tests
f798c27958 Don't run test_conf in cross compiled builds

mspncp:
> Travis is still green, but AppVeyor fails during installation. Interestingly, this failure did not occur when `[extended tests]` was not specified.

The AppVeyor failure of the extended tests is caused by issue 10648, see https://github.com/openssl/openssl/issues/10648#issuecomment-569351985

mspncp:
See also the [nmake documentation for $<](https://docs.microsoft.com/en-us/cpp/build/reference/filename-macros?redirectedfrom=MSDN&view=vs-2019):


    $< | Dependent file with a later timestamp than the current target.
         Valid only in commands in inference rules.

mspncp:
> [20e6631](https://github.com/openssl/openssl/commit/20e6631c73cfc643f57c3219a9c7acf4678e9b84) Temporarily disable some Travis external tests

This commit is being backported to 1.1.1. in #10722.

PR #10690, opened 2019-12-23
Update the krb5 submodule

kaduk:
Bring us up to date with upstream's 1.17.1 release.  Among other
things, it includes commit c2497d46b4bad473e164943d67b58cd1ae261c3a
which fixes several issues that affect running the test suite under
Travis CI.  Hopefully those will work transitively for us as well.

[extended tests]


kaduk:
I will plan to close/reopen several times to get multiple test runs (unless someone wants to manually kick off builds instead?).

mspncp:
> unless someone wants to manually kick off builds instead?

FWIW: As a committer you should be able to restart Travis jobs for openssl/openssl. All you need to do is to log into Travis using your GitHub account. (I’ve done it previously.)

mattcaswell:
Looks good. Based on this I should be able to keep krb5 enabled in #10689.

kaduk:
Okay, I ran job 19 a total of six times (i.e., five times "by hand" plus the original build) and in all cases, only the pyca-cryptography test failed (and the krb5 test succeeded).  That's enough for me to "take this out of WIP"; how about for you @mattcaswell ?

mspncp:
Ready to merge.

kaduk:
Pushed to master.  I guess we might need this on 1.1.1 as well, though; should have thought of that earlier

mspncp:
> I guess we might need this on 1.1.1 as well, though; should have thought of that earlier

Agreed. Since this commit cherry-picks cleanly, I think only a reapproval of Matt is necessary, plus another grace period.

mspncp:
OTOH, if @mattcaswell thinks that no extra grace period is necessary, I don't mind.

mspncp:
Since @mattcaswell is on vacation can anyone else of @openssl/committers confirm the cherry-pick to 1.1.1?

kaduk:
I'll re-run the 24-hour timer to keep matt's approval, unless someone wants this in 1.1.1 sooner.

mspncp:
> I'll re-run the 24-hour timer to keep matt's approval, unless someone wants this in 1.1.1 sooner.

I don't think an additional approval should reset the grace period. It would be a different case if another reviewer from the team would make change requests. In that case it is reasonable to mandate that the change request needs to be addressed or resolved first and then the grace period restarts. Looks like some more clarification is necessary, @openssl/otc?

mspncp:
(IOW, I think the grace period started with @levitte's approval and was not reset by @mattcaswell's approval and both approvals count.)

mspncp:
Cherry-picked to 1.1.1 as a6194c315091feb22e7cb6bf0bf339cabae733f3, thanks!

kaduk:
> Cherry-picked to 1.1.1 as a6194c3, thanks!

Thanks @mspncp !

frozencemetery:
Thanks for fixing this over the holidays!

I lack the capacity to follow openssl as closely as I'd like, but you all should feel free to reach out to me, or other krb5 folks, when things like this happen.

PR #10691, opened 2019-12-24
Return correct code on operation not supported

### LINKED ISSUE #8766: ECDSA_do_verify and ECDSA_verify incorrect return values.
In https://github.com/openssl/openssl/blob/master/crypto/ec/ecdsa_vrf.c there are two functions, both feature the same...

``` C
/*-
 * returns
 *      1: correct signature
 *      0: incorrect signature
 *     -1: error
 */
int ECDSA_do_verify(const unsigned char *dgst, int dgst_len,
                    const ECDSA_SIG *sig, EC_KEY *eckey)
{
    if (eckey->meth->verify_sig != NULL)
        return eckey->meth->verify_sig(dgst, dgst_len, sig, eckey);
    ECerr(EC_F_ECDSA_DO_VERIFY, EC_R_OPERATION_NOT_SUPPORTED);
    return 0;
}
```

In both cases if the verify_sig is not available then a no supported error is created but a return of incorrect signature provided. The signature has not been checked here, the facility to check is missing, surely the correct return in this case is 'error' and the user will be able to retrieve not supported and wonder to their hearts content.

a--hoang:
Fixes #8766

### DIFF
--- a/crypto/ec/ecdsa_vrf.c
+++ b/crypto/ec/ecdsa_vrf.c
@@ -23,7 +23,7 @@ int ECDSA_do_verify(const unsigned char *dgst, int dgst_len,
     if (eckey->meth->verify_sig != NULL)
         return eckey->meth->verify_sig(dgst, dgst_len, sig, eckey);
     ECerr(EC_F_ECDSA_DO_VERIFY, EC_R_OPERATION_NOT_SUPPORTED);
-    return 0;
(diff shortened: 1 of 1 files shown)


a--hoang:
mailed CLA under different email, lmk if it is a blocker

a--hoang:
will open a new PR with correct email address

PR #10693, opened 2019-12-24
Return correct code on operation not supported

### LINKED ISSUE #8766: ECDSA_do_verify and ECDSA_verify incorrect return values.

a--hoang:
Fixes #8766 by returning -1 when an eckey does not have a verify_sig function as the function signature specifies -1.

### DIFF
--- a/crypto/ec/ecdsa_vrf.c
+++ b/crypto/ec/ecdsa_vrf.c
@@ -23,7 +23,7 @@ int ECDSA_do_verify(const unsigned char *dgst, int dgst_len,
     if (eckey->meth->verify_sig != NULL)
         return eckey->meth->verify_sig(dgst, dgst_len, sig, eckey);
     ECerr(EC_F_ECDSA_DO_VERIFY, EC_R_OPERATION_NOT_SUPPORTED);
-    return 0;
+    return -1;
 }
 
 /*-
@@ -39,5 +39,5 @@ int ECDSA_verify(int type, const unsigned char *dgst, int dgst_len,
         return eckey->meth->verify(type, dgst, dgst_len, sigbuf, sig_len,
                                    eckey);
     ECerr(EC_F_ECDSA_VERIFY, EC_R_OPERATION_NOT_SUPPORTED);
-    return 0;
+    return -1;
 }


romen:
Actually there is one thing I'd like to improve before merge: the commit message.

The idea is to have the git history as self contained as possible, so that a person working on a local clone does not need to constantly check github to grasp the intent or motivation of commits. 

It would be nice if you could force push a new commit with the same changes, but an improved message.

For example, in the tile you could already narrow the scope of what is exactly being fixed (the current title is almost ok, but someone reading through the titles of commits would need to check the diff to understand that this affects the low level ECDSA code, and the verify path in particular).

In the body you could include a more detailed explanation of the original bug, how behavior did not follow documentation and how this commit fixes it. 

Finally, it would be best to also include at the end of the body a line with just  "Fixes #8766"

romen:
Ping @mspncp in case he has more suggestions about the commit message

mspncp:
@romen I think you already described it very nicely, so there is nothing for me to add at the moment.

a--hoang:
updated commit message as per suggestion from @romen

romen:
> updated commit message as per suggestion from @romen

Looks very good, thanks!

The only thing I would still add is

```
Fixes #8766
```

at the very end.

Other than that my approval stands.

romen:
@mspncp is your approval OTC worth? Is this now completely approved and just waiting for the grace period?

mspncp:
Well, It wasn’t when I gave it, but I guess it has been upgraded ;-)

romen:
We might need to wait until https://github.com/openssl/web/pull/146 is merged though? (or an OMC member gives his approval before that).

romen:
In any case, @a--hoang, even if this is not completely merged in yet, let me congratulate on your first approved contribution to the project!

I personally particularly appreciate these kind of contributions where implementation is checked against the official documentation: the code changes might seem very small at times, but their value goes well beyond the number of lines of code!

Thanks again for your work, hopefully this will be just the first of many other quality contributions to the project!

mspncp:
> We might need to wait until [openssl/web#146](https://github.com/openssl/web/pull/146) is merged though? (or an OMC member gives his approval before that).

Good point. Since the [approval: omc review pending] has already been renamed, I added a hold label to indicate that this pr either requires an OMC approval or needs to wait until [openssl/web#146](https://github.com/openssl/web/pull/146) has been voted on by the OMC.

mspncp:
_(@a--hoang don't worry, the hold has nothing to do with your pull request. It's just a formality, because we are currently changing our policies.)_

romen:
Merged to `master` with:
- 26583f6aa8 Fix incorrect return code on ECDSA key verification

Merged to `1.1.1-stable` with:
- 0fcc6e70bc Fix incorrect return code on ECDSA key verification

Thanks for your contribution @a--hoang and congrats again on your first successfully merged pull request!

PR #10699, opened 2019-12-24
FIPS 140-2 4.9.2 Continuous random number generator test

ManishPatidar1:
FIPS140-2 4.9.2 section:
Continuous  random  number  generator  test is added 

##### Checklist

### DIFF
--- a/crypto/rand/drbg_ctr.c
+++ b/crypto/rand/drbg_ctr.c
@@ -334,10 +334,14 @@ __owur static int drbg_ctr_generate(RAND_DRBG *drbg,
                 || outl != AES_BLOCK_SIZE)
                 return 0;
             memcpy(out, ctr->K, outlen);
+            if (!drbg_cprng_test(drbg, ctr->K, AES_BLOCK_SIZE))
+                return 0;
             break;
         }
         if (!EVP_CipherUpdate(ctr->ctx, out, &outl, ctr->V, AES_BLOCK_SIZE)
             || outl != AES_BLOCK_SIZE)
+            return 0;
+        if (!drbg_cprng_test(drbg, out, AES_BLOCK_SIZE))
             return 0;
         out += 16;
         outlen -= 16;
--- a/crypto/rand/drbg_hash.c
+++ b/crypto/rand/drbg_hash.c
@@ -190,9 +190,13 @@ static int hash_gen(RAND_DRBG *drbg, unsigned char *out, size_t outlen)
             if (!EVP_DigestFinal(hash->ctx, hash->vtmp, NULL))
                 return 0;
             memcpy(out, hash->vtmp, outlen);
+            if (!drbg_cprng_test(drbg, hash->vtmp, hash->blocklen))
+                return 0;
             return 1;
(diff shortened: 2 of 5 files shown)


slontis:
@paulidale - I thought you had already added a check somewhere..

ManishPatidar1:
crypto/rand/rand_crng_test.c will test only entropy input to DRNG  but FIPS140-2 4.9.2  is for random output from DRBG .  Please refer the below section of FIPS140-2 .  Moreover same check was present in OpenSSL FIPS 2.0 , 

Continuous random number generator test. If a cryptographic module employs Approved or nonApproved RNGs in an Approved mode of operation, the module shall perform the following continuous random number generator test on each RNG that tests for failure to a constant value.
1. If each call to a RNG produces blocks of n bits (where n > 15), the first n-bit block generated
after power-up, initialization, or reset shall not be used, but shall be saved for comparison with
the next n-bit block to be generated. Each subsequent generation of an n-bit block shall be
compared with the previously generated block. The test shall fail if any two compared n-bit
blocks are equal.
2. If each call to a RNG produces fewer than 16 bits, the first n bits generated after power-up,
initialization, or reset (for some n > 15) shall not be used, but shall be saved for comparison with
the next n generated bits. Each subsequent generation of n bits shall be compared with the
previously generated n bits. The test fails if any two compared n-bit sequences are equal.

paulidale:
A DRBG is **not** a RNG.

ManishPatidar1:
does it mean this check is not required  ?

paulidale:
Correct.

ManishPatidar1:
But same check was present in OpenSSL FIPS 2.0 . you can refer to fips_drbg_cprng_test function .

ManishPatidar1:
RNG is generic term for all random number generator.  DRBG is one of the approve RNG.
so I think , this check is required .

https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.140-2.pdf
**4.7.1 Random Number Generators (RNGs)**
A cryptographic module may employ random number generators (RNGs). If a cryptographic module employs Approved or non-Approved RNGs in an Approved mode of operation, the data output from the RNG shall pass the continuous random number generator test as specified in Section 4.9.2. Approved RNGs shall be subject to the cryptographic algorithm test in Section 4.9.1. Approved RNGs are listed in Annex C to this standard.

Annex C
https://csrc.nist.gov/csrc/media/publications/fips/140/2/final/documents/fips1402annexc.pdf

paulidale:
Have a look at [FIPS 140-2 IG 9.8](https://csrc.nist.gov/csrc/media/projects/cryptographic-module-validation-program/documents/fips140-2/fips1402ig.pdf): _Continuous Random Number Generator Tests_.  Several pages in (bottom of 155) it says:

> The SP-800-90A-compliant DRBGs are not required to perform the test as described in AS.09.42 and AS.09.43.

The myriad of FIPS standards cannot be read in isolation.

ManishPatidar1:
ok , got it .
Thanks for clarification.

PR #10719, opened 2019-12-28
nmake: fix install_html_docs target

### LINKED ISSUE #10648: windows10 - `nmake install` fails with fatal error u1077: "pod2html": return-code "0xff"

### LINKED ISSUE #10749: Broken AppVeyor builds in master after commit b0940b33a60
The commit b0940b33a60cbfa broke the AppVeyor builds in the master branch CI:
See for example https://ci.appveyor.com/project/openssl/openssl/builds/29200540

mspncp:
The $< target[1] does not work for regular rules and is expanded to an empty string after issuing the warning

    NMAKE : warning U4006: special macro undefined : '$<"'

This leads to the following error message by pod2html:

    'href' is not recognized as an internal or external command,
    operable program or batch file.

[1] https://docs.microsoft.com/en-us/cpp/build/reference/filename-macros

_Edit: an updated description of the error can be found [further down](https://github.com/openssl/openssl/pull/10719#issuecomment-569456069)._

Fixes #10648
Fixes #10749

[extended tests]

### DIFF
--- a/Configurations/windows-makefile.tmpl
+++ b/Configurations/windows-makefile.tmpl
@@ -588,8 +588,8 @@ reconfigure reconf:
 	  return <<"EOF";
 $args{src}: $args{generator}->[0]
 	pod2html "--podroot=\$(SRCDIR)/doc" --htmldir=.. \\
-		 --podpath=man1:man3:man5:man7 "--infile=\$<" "--title=$title" \\
-	| \$(PERL) -pe "s|href=\\"http://man\\.he\\.net/(man\d/[^\\"]+)(?:\\.html)?\\"|href=\\"../\$1.html|g;" \\
+		 --podpath=man1:man3:man5:man7 "--infile=\$?" "--title=$title" \\
+	| \$(PERL) -pe ^"s^|href=\\^"http://man\\.he\\.net/^(man\\d/[^^\\^"]+^)^(?:\.html^)?^"^|href=\\^"../\$\$1.html^|g;^" \\
 	> \$\@
 EOF
       } elsif (platform->isdef($args{src})) {


mspncp:
Hmmm... or should I better use `$**` instead of `$?`?

```
$** | All dependents of the current target.
 -- | --
 $? | All dependents with a later timestamp than the current target.
```
https://docs.microsoft.com/en-us/cpp/build/reference/filename-macros

mspncp:
Hmmm, this [did not work as expected](https://ci.appveyor.com/project/openssl/openssl/builds/29788602/job/o57aq9fmqdfmqbee):

```
*** Installing runtime programs
Copying: apps//openssl.exe to ../_install/Program Files/OpenSSL/bin/openssl.exe
Copying: apps//openssl.pdb to ../_install/Program Files/OpenSSL/bin/openssl.pdb
Copying: tools//c_rehash.pl to ../_install/Program Files/OpenSSL/bin/c_rehash.pl
created directory `../_install/Program Files/Common Files'
created directory `../_install/Program Files/Common Files/SSL'
created directory `../_install/Program Files/Common Files/SSL/certs'
created directory `../_install/Program Files/Common Files/SSL/private'
created directory `../_install/Program Files/Common Files/SSL/misc'
Copying: ../apps/openssl.cnf to ../_install/Program Files/Common Files/SSL/openssl.cnf.dist
Copying: ../apps/openssl.cnf to ../_install/Program Files/Common Files/SSL/openssl.cnf
Copying: apps//CA.pl to ../_install/Program Files/Common Files/SSL/misc/CA.pl
Copying: apps//tsget.pl to ../_install/Program Files/Common Files/SSL/misc/tsget.pl
Copying: ../apps/ct_log_list.cnf to ../_install/Program Files/Common Files/SSL/ct_log_list.cnf.dist
Copying: ../apps/ct_log_list.cnf to ../_install/Program Files/Common Files/SSL/ct_log_list.cnf
	pod2html "--podroot=../doc" --htmldir=..  --podpath=man1:man3:man5:man7 "--infile=..\doc\man1\openssl-asn1parse.pod" "--title=openssl-asn1parse"  | C:\Perl\bin\perl.exe -pe "s|href=\"http://man\.he\.net/(mand/[^\"]+)(?:\.html)?\"|href=\"../.html|g;"  > doc\html\man1\openssl-asn1parse.html
'href' is not recognized as an internal or external command,
operable program or batch file.
NMAKE : fatal error U1077: 'pod2html' : return code '0xff'
Stop.

```

mspncp:
Looks like the perl regex for the substitution is not quoted properly and the ‚href‘ is treated as a command by cmd.exe.

mspncp:
I force-pushed a new fix. Here's its commit message:

The nmake rule contains actually two errors:

1. The $< target[1] does not work for regular rules and is
expanded to an empty string after issuing the warning

    NMAKE : warning U4006: special macro undefined : '$<"'

Solution: replace $< by $?

2. The substitution regex is not quoted correctly, which leads
to the following error message by cmd.exe:

    'href' is not recognized as an internal or external command,
    operable program or batch file.

Solution: Quoting arguments for cmd.exe is really a nightmare,
but with the help of the excellent description [2] I was able to
properly quote the regex. Things were complicated by the fact that
a lot of levels of unquoting needed to be considered:

 * perl (windows-makefile.tmpl -> makefile)
 * make (reading the makefile)
 * cmd.exe (executed by make)
 * perl (scanning command line using CommandLineToArgvW())

The fix works, but the regex has become unmaintainable. It would actually
be better to wrap the entire command (including the regex) into a little
perl script which can be called by make directly.

[1] https://docs.microsoft.com/en-us/cpp/build/reference/filename-macros
[2] https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/

Fixes #10648
Fixes the installation test on Windows

[extended tests]

mspncp:
> The fix works, but the regex has become unmaintainable. It would actually
> be better to wrap the entire command (including the regex) into a little
> perl script which can be called by make directly.

I'll leave that part for the @levitte ;-)

mspncp:
I checked explicitly for the file openssl-cipher.html that the substitution produces the expected output, and tested the installed html file in the browser to verify that the link works correctly.

Before replacement:

```html
<p>Verbose output: For each cipher suite, list details as provided by
    <a href="http://man.he.net/man3/SSL_CIPHER_description">SSL_CIPHER_description(3)</a>.</p>
```

After replacement:

```html
<p>Verbose output: For each cipher suite, list details as provided by
    <a href="../man3/SSL_CIPHER_description.html">SSL_CIPHER_description(3)</a>.</p>
```

Configurations/windows-makefile.tmpl
```diff
@@ -588,8 +588,8 @@ reconfigure reconf:
 	  return <<"EOF";
 $args{src}: $args{generator}->[0]
 	pod2html "--podroot=\$(SRCDIR)/doc" --htmldir=.. \\
-		 --podpath=man1:man3:man5:man7 "--infile=\$<" "--title=$title" \\
-	| \$(PERL) -pe "s|href=\\"http://man\\.he\\.net/(man\d/[^\\"]+)(?:\\.html)?\\"|href=\\"../\$1.html|g;" \\
+		 --podpath=man1:man3:man5:man7 "--infile=\$?" "--title=$title" \\
+	| \$(PERL) -pe ^"s^|href=\\^"http://man\\.he\\.net/^(man\\d/[^^\\^"]+^)^(?:\.html^)?^"^|href=\\^"../\$\$1.html^|g;^" \\
```
mspncp: Here an explanation of the tricky parts
* The `^` characters deactivate all special cmd.exe characters, in particular `^"` deactives the inside-string-mode of cmd.exe entirely.
* The `\\^"` is passed as `\"` to perl.exe, where it gets converted to `"` by CommandLineToArgvW().
* `../\$\$1.html`  is seen by make as `../$$1.html` and the double $$ prevents it from expanding `../$1.html` to `../.html`.
* `[^^\\^"]` is passed as  `[^\"]`  from cmd.exe to perl and ends up as `[^"]` when parsed by CommandLineToArgvW().

mspncp: It [feels like a victory](https://en.wikipedia.org/wiki/Pyrrhic_victory) over cmd.exe!

mspncp:
Oh c'mon coveralls, you're a spoiler!

image

mspncp:
I dropped the [urgent] label because nobody else seems to consider this urgent. ;-)

mspncp:
Added a Fixes line to the commit message without [any other changes](https://github.com/openssl/openssl/compare/5e19b5dc00cd92d689c8d53161430ac10f2ea4ed..ff8cea2b5dc24c69b46a5c9fd538c98212badf61)

mspncp:
AppVeyor and Travis are both happy (for the first time in a long time).

Configurations/windows-makefile.tmpl
romen: I have reviewed the changes and they seem OK to me, but I have not formally approved because I have close to 0 experience on Windows-specific development and I wonder if the special `^` escape of quotes and `^` itself is compatible across all the supported windows platforms.

mspncp:
> > The fix works, but the regex has become unmaintainable. It would actually
> > be better to wrap the entire command (including the regex) into a little
> > perl script which can be called by make directly.
> 
> I'll leave that part for @levitte ;-)

@levitte unless you are planning to immediately supersede this pull request with an alternative fix, it would be great if you could approve this provisional fix to make AppVeyor happy.

Configurations/windows-makefile.tmpl
levitte: `^` has been part of the cmd.exe syntax for a while, so I consider it fairly safe.

mspncp:
Merged to master as a9e4e3c39eb361ddfb438edb27c754947009eed6, thanks!

PR #10720, opened 2019-12-28
appveyor.yml: speed up the extended test

mspncp:
OpenSSL is installed to a temporary folder when the extended tests
are enabled. We omit the documentation to save some time.

[extended tests]

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -61,6 +61,6 @@ test_script:
     - ps: >-
         if ($env:EXTENDED_TESTS) {
             mkdir ..\_install
-            cmd /c "nmake install DESTDIR=..\_install 2>&1"
+            cmd /c "nmake install_sw install_ssldirs DESTDIR=..\_install 2>&1"
         }
     - cd ..


mspncp:
_Side note: this pull request also "fixes" the current failure of the extended tests on AppVeyor._

romen:
I am conflicted about this: while it makes sense to speed up, at the same time I am wary of the fact that we would not have caught the current failure if this change was in!

mspncp:
I agree that we should have an installation test on windows and linux in the future. But it shouldn’t just be an undocumented side-effect of running the extended tests.

mspncp:
I am beginning to wonder whether this actually _is intended_ to be an installation test? Or is it just a prerequisite for other extended tests? The CI configuration files (AppVeyor, Travis) are heavily underdocumented IMHO.

romen:
> I am beginning to wonder whether this actually _is intended_ to be an installation test? Or is it just a prerequisite for other extended tests?

My interpretation so far has been that AppVeyor is testing both unit tests and the build system (install included) across different versions of win specific frameworks.

I do agree on the underdocumented comment, unfortunately I believe part of the problem is the CI/CD ecosystem relaying on an ever evolving plethora of DSL built on YAML or similar languages not always capable of separating the content of these configuration files from the inner workings of the specific tool.

mspncp:
I tracked the convoluted history of this code fragment

https://github.com/openssl/openssl/blob/5f95fbf39987848fc50fc8acf42eb8f4a7ec8666/appveyor.yml#L62-L65

all the way back to commit 14c4201958152392941615626da5d2800c98e943:

https://github.com/openssl/openssl/blob/14c4201958152392941615626da5d2800c98e943/appveyor.yml#L45-L50

This shows that it is indeed an installation test and @levitte's original comment clearly indicated that. Unfortunately, this helpful comment got lost on the way. It would have spared me a lot wasted time :-/
It  would be nice, if some helpful comment could be restored.

In view of this my pull request does not make sense anymore. So I'm closing it.

PR #10724, opened 2019-12-29
Declare lhash in headers, not define them, when building OpenSSL

richsalz:
This is just like #10669 except it removes the definitions of inline functions for the LHASH stuff.

I am not thrilled that I had to add pragma's in two places but there were already six files that had the same pragma.


richsalz:
Rebased on master now that #10753 has been merged.

crypto/lhash/lhash.c
```diff
@@ -35,6 +35,10 @@
  * https://pdfs.semanticscholar.org/ff4d/1c5deca6269cc316bfd952172284dbf610ee.pdf
  */
 
+#if !defined(OPENSSL_BUILDING_OPENSSL)
+#error wtf?
```
mspncp: Please don‘t use x-rated language.

richsalz: It is intended to be a temporary commit to help diagnose why the build's are broken. But point taken, I updated it.

richsalz:
I am moving this to WIP because it is not stictly required for the 3.0/FIPS release and I don't want other people spending time on it now.

richsalz:
Undid the WIP.  Pushed another commit that adds documentation.

richsalz:
```
; cat /home/rsalz/test-declare-builds
#! /bin/bash
set -e

cat <<EOF >a.c
#include <unistd.h>
#include <string.h>
#include <openssl/x509.h>

int
main(int argc, char **argv)
{
  EVP_add_cipher(NULL);
  write(1, argv[1], strlen(argv[1]));
  write(1, "\n", 1);
  return 0;
}
EOF

doit() {
./config $* >/dev/null
make -s clean >/dev/null
make -s >/dev/null
gcc -Iinclude a.c libcrypto.a -lpthread -ldl
./a.out "config $* --- GCC ok"
clang -Iinclude a.c libcrypto.a -lpthread -ldl
./a.out "config $* --- CLANG ok."
}
echo DEFAULT CONFIG
doit

echo NO-DEPRECATED CONFIG
doit no-deprecated

echo NO-DEPRECATED API=3.0.0 CONFIG
doit no-deprecated --api=3.0.0

rm a.c a.out
; /home/rsalz/test-declare-builds
DEFAULT CONFIG
ar: creating apps/libapps.a
ar: creating libcrypto.a
ar: creating libssl.a
ar: creating providers/libcommon.a
ar: creating providers/libfips.a
ar: creating providers/libimplementations.a
ar: creating providers/liblegacy.a
ar: creating providers/libnonfips.a
config  --- GCC ok
config  --- CLANG ok.
NO-DEPRECATED CONFIG
ar: creating apps/libapps.a
ar: creating libcrypto.a
ar: creating libssl.a
ar: creating providers/libcommon.a
ar: creating providers/libfips.a
ar: creating providers/libimplementations.a
ar: creating providers/liblegacy.a
ar: creating providers/libnonfips.a
config no-deprecated --- GCC ok
config no-deprecated --- CLANG ok.
NO-DEPRECATED API=3.0.0 CONFIG
ar: creating apps/libapps.a
ar: creating libcrypto.a
ar: creating libssl.a
ar: creating providers/libcommon.a
ar: creating providers/libfips.a
ar: creating providers/libimplementations.a
ar: creating providers/liblegacy.a
ar: creating providers/libnonfips.a
config no-deprecated --api=3.0.0 --- GCC ok
config no-deprecated --api=3.0.0 --- CLANG ok.
```

apps/req.c
```diff
@@ -32,6 +32,14 @@
 # include <openssl/dsa.h>
 #endif
 
+#ifdef __GNUC__
+# pragma GCC diagnostic ignored "-Wunused-function"
+#endif
+#ifdef __clang__
+# pragma clang diagnostic ignored "-Wunused-function"
+#endif
```
levitte: Why is this needed here?

crypto/txt_db/txt_db.c
```diff
@@ -14,6 +14,14 @@
 #include <openssl/buffer.h>
 #include <openssl/txt_db.h>
 
+#ifdef __GNUC__
+# pragma GCC diagnostic ignored "-Wunused-function"
+#endif
+#ifdef __clang__
+# pragma clang diagnostic ignored "-Wunused-function"
+#endif
```
levitte: Why is this needed?

test/confdump.c
```diff
@@ -14,8 +14,18 @@
 #include <openssl/safestack.h>
 #include <openssl/err.h>
 
+#ifdef __GNUC__
+# pragma GCC diagnostic ignored "-Wunused-function"
+#endif
+#ifdef __clang__
+# pragma clang diagnostic ignored "-Wunused-function"
+#endif
```
levitte: Why is this needed?

crypto/txt_db/txt_db.c
richsalz: because without it, clang generates an error. see https://github.com/openssl/openssl/blob/master/test/sparse_array_test.c#L21 for example.

test/confdump.c
richsalz: same as above.

apps/req.c
richsalz: Strangely, removing this does not cause errors for this one file.  fixup commit pushed.

richsalz:
Rebased after #10669 went in.  This is the same as that, but for LHASH not STACK.

richsalz:
This has now been waiting for OTC action for over seven weeks.  The BETA release, which is scheduled for 2-3 weeks from now, is an API freeze.  Does this change the API?

CHANGES.md
```diff
@@ -423,6 +423,11 @@ OpenSSL 3.0
   
    *Rich Salz*
 
+ * LHASH macros have better documentation, and OpenSSL source only
+   defines LHASH tables where the tables are actually used.
```
mattcaswell, line 427: IIUC this actually removes the implementation of the inline functions from the public headers in a no-deprecated build. I think there should be some discussion of that in CHANGES, and what applications that rely on this should do instead. Should inline functions implemented in the public functions be declared as deprecated as well? Otherwise there will be no indication to users of these functions that they are using something that is likely to disappear in the future.

crypto/txt_db/txt_db.c
mattcaswell, line 22: Really not enthusiastic about this...but struggling to think of an alternative.

include/openssl/err.h
```diff
@@ -247,7 +247,11 @@ typedef struct ERR_string_data_st {
     const char *string;
 } ERR_STRING_DATA;
 
+#if defined(OPENSSL_BUILDING_OPENSSL) || defined(OPENSSL_NO_DEPRECATED_3_0)
```
mattcaswell, line 250: I wonder if there should be some capability for applications to  decide to declare things in this way without having to build OpenSSL with `no-deprecated`.

providers/fipsinstall.cnf
```diff
@@ -0,0 +1,6 @@
+[fips_sect]
```
mattcaswell, line 1: This seems to be here in error?

CHANGES.md
richsalz: This follows the same wording that I used in #10669, which was merged seven months ago. The manpage in this PR explains things.  I'm happy to make a change here (and for the STACK entry as well), is it enough to point to the manpage?

I didn't think of deprecating the inline functions, and would rather not do that here as this has waited long enough.

include/openssl/err.h
richsalz: We haven't had finer-grain control over deprecation before, so I'd be reluctant to add it to this PR.  Anything we do, should also be done for the STACK changes.

providers/fipsinstall.cnf
richsalz: Yes, the file was renamed in #11177 which happened in April.  Lucky that this was the only rebase issue.

richsalz:
Rebased, squashed, and pushed (there was a conflict in include/openssl/conf.h)

include/openssl/err.h
mattcaswell: > We haven't had finer-grain control over deprecation before, so I'd be reluctant to add it to this PR. Anything we do, should also be done for the STACK changes.

The difference is that with other things we have deprecated application authors can simply choose to change their code to not do things the deprecated way any more, even if they have a normal build.

With this change we are effectively deprecating something, but unless you have a no-deprecated build you can't do it the "right" way. An application that uses the inline functions and adds the `DEFINE_LHASH_OF` macro will build successfully when using a no-deprecated build - but will fail to compile in a normal build and vice versa.

CHANGES.md
mattcaswell: Actually, I think the big problem here is that application authors can't do the right thing so its not clear what the CHANGES entry should direct them to do (see my other comment).

include/openssl/err.h
richsalz: I believe you are mistaken about what will compile/not-compile. That's the reasoning behind the ```OPENSSL_BUILDING_OPENSSL``` macro.  Do you have code/instructions I could try?  And doesn't this same issue up for the STACK changes also?

CHANGES.md
richsalz: I still think you are wrong.  Can you sketch out an example that shows your point? Applications that want to work in both worlds can still use ```DEFINE_LHASH_OF``` in a header file. Recall that these are *inline* functions, so there's nothing in openssl libs that matters.

include/openssl/err.h
mattcaswell: Consider this code:
````
#include <openssl/lhash.h>

int main(void)
{
    LHASH_OF(OPENSSL_STRING) *hash = lh_OPENSSL_STRING_new(NULL, NULL);

    if (hash == NULL)
        return 1;

    return 0;
}
````
This will build just fine when using a normal build of OpenSSL. However if I use a no-deprecated build it fails to compile. Fair enough - I have to do it the new "right" way like this:

````
#include <openssl/lhash.h>

DEFINE_LHASH_OF(OPENSSL_STRING);

int main(void)
{
    LHASH_OF(OPENSSL_STRING) *hash = lh_OPENSSL_STRING_new(NULL, NULL);

    if (hash == NULL)
        return 1;

    return 0;
}
````
This code will compile correctly with a no-deprecated build.

However if I try this second version with a normal build this code *also* fails to compile. So, as an application author, I cannot make changes to my code now to prepare for the later deprecation changes.

Comment (richsalz) [reply include/openssl/err.h:250]: I get errors when I build against master, too.  That's because OpenSSL declares the hash of cstring.
If, however, I change ```OEPNSSL_STRING``` in your code to add
```
typedef struct opaque OPAQUE;
```
And then change all strings to OPAQUE, the code compiles against master and this PR.

Comment (mattcaswell) [reply include/openssl/err.h:250]: > And then change all strings to OPAQUE

How does that help if my application needs to work with OPENSSL_STRINGs?

For example how can I write code that works in both types of build that uses the return value of `SSL_CTX_sessions`:

LHASH_OF(SSL_SESSION) *SSL_CTX_sessions(SSL_CTX *ctx);

Comment (richsalz) [reply include/openssl/err.h:250]: I added the line above to your sample, and it worked on master and this branch.

Comment (mattcaswell) [reply include/openssl/err.h:250]: By the "line above" you means this?

````
#if defined(OPENSSL_BUILDING_OPENSSL) || defined(OPENSSL_NO_DEPRECATED_3_0)
````

If so then that isn't a solution. We should not expect applications to have to do this ifdef hackery just to get their application to work in a no-deprecated vs normal build of OpenSSL. If it works in a "no-deprecated" build, then it should work unchanged in normal build without any ifdef.

Comment (richsalz) [reply include/openssl/err.h:250]: No, I meant the line you said "this won't work."  Here's the whole file.  As I said, it  compiles on both master and this PR.

```
#include <openssl/lhash.h>

typedef struct _opaque OPAQUE;
DEFINE_LHASH_OF(OPAQUE);
LHASH_OF(SSL_SESSION) *SSL_CTX_sessions(SSL_CTX *ctx);

int main(void)
{
    LHASH_OF(OPAQUE) *hash = lh_OPAQUE_new(NULL, NULL);

    if (hash == NULL)
        return 1;
    return 0;
}
```

Comment (mattcaswell) [reply include/openssl/err.h:250]: This is missing my point entirely. I'm not concerned about applications defining their own types. I'm concerned about them using *our* types. I'll try and give another example to illustrate my point. I'll use a stack because I believe the same issue already exists in master, that you are now introducing for hashes here - and there are more instances of stack usage that I can use as an example.

````
#include <openssl/x509_vfy.h>

int dostuff(X509_STORE_CTX *ctx)
{
    int i;
    STACK_OF(X509) *xstk = X509_STORE_CTX_get0_chain(ctx);

    for (i = 0; i < sk_X509_num(xstk); i++) {
        X509 *x = sk_X509_value(xstk, i);
        /* Do something with x */
    }
}
````
In this example an application wants to use our APIs to iterate through the contents of a stack. The above code compiles just fine on a normal build of current master. However it fails to compile in a no-deprecated build. I have to make it look like this instead:

````
#include <openssl/x509_vfy.h>

DEFINE_STACK_OF(X509)

int dostuff(X509_STORE_CTX *ctx)
{
    int i;
    STACK_OF(X509) *xstk = X509_STORE_CTX_get0_chain(ctx);

    for (i = 0; i < sk_X509_num(xstk); i++) {
        X509 *x = sk_X509_value(xstk, i);
        /* Do something with x */
    }
}
````
This code compiles fine in a no-deprecated build of master - but fails to compile on a normal build of master. As an application author it doesn't seem possible to get this code to work in both builds.

Comment (richsalz) [reply include/openssl/err.h:250]: > fails to compile in a no-deprecated build.

I took your first example and tried it on master and this branch:
```; g co master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
; ./config >/dev/null
; gcc -Iinclude -c a.c
; g co declare-lhash
Switched to branch 'declare-lhash'
; ./config >/dev/null  
; gcc -Iinclude -c a.c
; g co master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
; ./config no-deprecated >/dev/null  
; gcc -Iinclude -c a.c              
; g co declare-lhash
Switched to branch 'declare-lhash'
; ./config no-deprecated >/dev/null
; gcc -Iinclude -c a.c
;
```

Can you tell me what I'm doing wrong?  It compiles fine on both master and this branch with and without no-deprecated configurations.

mattcaswell: This is what I get against a no-deprecated build:

````
gcc -g -O0 -I/home/matt/dev/openssl-write/include -L/home/matt/dev/openssl-write -o stacktest stacktest.c -lcrypto
stacktest.c: In function ‘dostuff’:
stacktest.c:8:21: warning: implicit declaration of function ‘sk_X509_num’ [-Wimplicit-function-declaration]
    8 |     for (i = 0; i < sk_X509_num(xstk); i++) {
      |                     ^~~~~~~~~~~
stacktest.c:9:19: warning: implicit declaration of function ‘sk_X509_value’ [-Wimplicit-function-declaration]
    9 |         X509 *x = sk_X509_value(xstk, i);
      |                   ^~~~~~~~~~~~~
stacktest.c:9:19: warning: initialization of ‘X509 *’ {aka ‘struct x509_st *’} from ‘int’ makes pointer from integer without a cast [-Wint-conversion]
/usr/bin/ld: /tmp/cclgGumP.o: in function `dostuff':
/home/matt/dev/tmp/stacktest.c:9: undefined reference to `sk_X509_value'
/usr/bin/ld: /home/matt/dev/tmp/stacktest.c:8: undefined reference to `sk_X509_num'
collect2: error: ld returned 1 exit status
make: *** [Makefile:68: stacktest] Error 1

````
So, strictly speaking I get warnings during the compilation and an error during link because there is no implementation of the sk_X509_value and sk_X509_num functions. You only do the compilation stage...not sure why you don't see the warnings.

Comment (richsalz) [reply include/openssl/err.h:250]: I should have noticed before.  This PR is about *LHASH* not *STACK8.  So perhaps that earlier PR (#10669) has problems which need to be fixed.  Can you edit your program to try the same thing with a hash?  Should I copy your comment above to a new issue?

Comment (t8m) [reply include/openssl/err.h:250]: We could make the DEFINE_STACK_OF to be no-op when it is outside the openssl headers (i.e. OPENSSL_BUILDING_OPENSSL is not defined) in the regular (no no-deprecated) build. This way both no-deprecated and regular builds will work with the same application code. However it is an API breaking change that I did not realize when approving the stack change so we should have a vote about it.

Comment (mattcaswell) [reply include/openssl/err.h:250]: > I should have noticed before. This PR is about LHASH not STACK. So perhaps that earlier PR (#10669) has problems which need to be fixed. 

Right. As I said previously:

> I'll try and give another example to illustrate my point. I'll use a stack because I believe the same issue already exists in master, that you are now introducing for hashes here - and there are more instances of stack usage that I can use as an example.

So, it is my opinion that this problem exists in master right now for stacks, and you are about to introduce the same problem for hashes here as part of this PR.

> Can you edit your program to try the same thing with a hash?

I'll see if I can come up with another example

> Should I copy your comment above to a new issue?

Please do. I don't think this PR should go in until we've figured out the answer to that because I think it suffers from the same problem.

Comment (mattcaswell) [reply include/openssl/err.h:250]: > We could make the DEFINE_STACK_OF to be no-op when it is outside the openssl headers (i.e. OPENSSL_BUILDING_OPENSSL is not defined) in the regular (no no-deprecated) build. This way both no-deprecated and regular builds will work with the same application code. However it is an API breaking change that I did not realize when approving the stack change so we should have a vote about it.

The problem seems to mainly stem from the use of inline functions. I wonder whether we could just make them normal functions for external users, and keep them as inline for our own internal use.

File: include/openssl/conf.h [hunk 8]
```diff
@@ -34,8 +34,13 @@ typedef struct {
 } CONF_VALUE;
 
 DEFINE_OR_DECLARE_STACK_OF(CONF_VALUE)
-
+DEFINE_OR_DECLARE_STACK_OF(CONF_MODULE)
+DEFINE_OR_DECLARE_STACK_OF(CONF_VALUE)
```
mattcaswell, line 38: You have a duplicate entry for `DEFINE_OR_DECLARE_STACK_OF(CONF_VALUE)` here. This also produced compilation errors in my application.

include/openssl/conf.h
```diff
@@ -34,8 +34,13 @@ typedef struct {
 } CONF_VALUE;
 
 DEFINE_OR_DECLARE_STACK_OF(CONF_VALUE)
-
+DEFINE_OR_DECLARE_STACK_OF(CONF_MODULE)
```
mattcaswell, line 37: The `typedef struct conf_module_st CONF_MODULE;` around line 64 below needs to occur before this line. I was getting compile errors in my application including the conf.h file without this.

crypto/conf/conf_api.c
```diff
@@ -17,6 +15,7 @@
 #include <openssl/conf_api.h>
 
 DEFINE_STACK_OF(CONF_VALUE)
+DEFINE_LHASH_OF(CONF_VALUE);
```
mattcaswell, line 18: There is an inconsistency here between LHASH and STACK. In the former you don't need to add a ";" to the end of the `DEFINE_STACK_OF()` line. However with `DEFINE_LHASH_OF()` you must do so. They should probably work the same way.

include/openssl/err.h
mattcaswell: Here's another example which uses LHASH. I had to make a couple of tweaks to this PR to get my application to compile at all (see my latest review comments). Here is some code that builds just fine with a normal OpenSSL build (and those tweaks applied).

````
#include <openssl/conf.h>


static int dostuff(CONF_VALUE *val)
{
    int i;
    long eline;
    LHASH_OF(CONF_VALUE) *mycnf;
    
    mycnf = CONF_load(NULL, "blah.cnf", &eline);

    if (lh_CONF_VALUE_retrieve(mycnf, val) == NULL)
        return 0;

    return 1;
}
````

In a no-deprecated build I get these errors:

````
gcc -g -O0 -I/home/matt/dev/openssl-write/include -L/home/matt/dev/openssl-write -o hashtest hashtest.c -lcrypto
hashtest.c: In function ‘dostuff’:
hashtest.c:11:9: warning: implicit declaration of function ‘lh_CONF_VALUE_retrieve’ [-Wimplicit-function-declaration]
   11 |     if (lh_CONF_VALUE_retrieve(mycnf, val) == NULL)
      |         ^~~~~~~~~~~~~~~~~~~~~~
hashtest.c:11:44: warning: comparison between pointer and integer
   11 |     if (lh_CONF_VALUE_retrieve(mycnf, val) == NULL)
      |                                            ^~
/usr/bin/ld: /tmp/cc63nEkD.o: in function `dostuff':
/home/matt/dev/tmp/hashtest.c:11: undefined reference to `lh_CONF_VALUE_retrieve'
collect2: error: ld returned 1 exit status
make: *** [Makefile:74: hashtest] Error 1
````
Making the code look like this instead means it will compile/link in a no-deprecated build:

````
#include <openssl/conf.h>

DEFINE_LHASH_OF(CONF_VALUE);

static int dostuff(CONF_VALUE *val)
{
    long eline;
    LHASH_OF(CONF_VALUE) *mycnf;
    
    mycnf = CONF_load(NULL, "blah.cnf", &eline);

    if (lh_CONF_VALUE_retrieve(mycnf, val) == NULL)
        return 0;

    return 1;
}
````

But this fails with different compile errors against a normal build, e.g.:

````
gcc -g -O0 -I/home/matt/dev/openssl-write/include -L/home/matt/dev/openssl-write -o hashtest hashtest.c -lcrypto
In file included from /home/matt/dev/openssl-write/include/openssl/conf.h:20,
                 from hashtest.c:1:
/home/matt/dev/openssl-write/include/openssl/lhash.h:126:32: error: redefinition of ‘struct lhash_st_CONF_VALUE’
  126 | # define LHASH_OF(type) struct lhash_st_##type
      |                                ^~~~~~~~~
/home/matt/dev/openssl-write/include/openssl/lhash.h:129:5: note: in expansion of macro ‘LHASH_OF’
  129 |     LHASH_OF(type) { union lh_##type##_dummy { void* d1; unsigned long d2; int d3; } dummy; }; \
      |     ^~~~~~~~
hashtest.c:3:1: note: in expansion of macro ‘DEFINE_LHASH_OF’
    3 | DEFINE_LHASH_OF(CONF_VALUE);
      | ^~~~~~~~~~~~~~~
/home/matt/dev/openssl-write/include/openssl/lhash.h:126:32: note: originally defined here
  126 | # define LHASH_OF(type) struct lhash_st_##type
      |                                ^~~~~~~~~
/home/matt/dev/openssl-write/include/openssl/lhash.h:129:5: note: in expansion of macro ‘LHASH_OF’
  129 |     LHASH_OF(type) { union lh_##type##_dummy { void* d1; unsigned long d2; int d3; } dummy; }; \
      |     ^~~~~~~~
/home/matt/dev/openssl-write/include/openssl/conf.h:47:1: note: in expansion of macro ‘DEFINE_LHASH_OF’
   47 | DEFINE_LHASH_OF(CONF_VALUE);
      | ^~~~~~~~~~~~~~~
````
There are more errors than this - but just included this for brevity.

richsalz: https://github.com/openssl/openssl/issues/12193 to cover the general issue.

include/openssl/conf.h
richsalz: rebase error; fixed.

crypto/conf/conf_api.c
richsalz: okay DEFINE_LHASH_OF does not need the trailing semicolon now.

include/openssl/conf.h
richsalz: moved the macro after the typedef, thanks.

richsalz:
It is a pity that this PR languished for months and that nobody found the issues before.  Oh well.

If someone comes up with a solution for this and the stack, let me know and I'll do it.  But I'm out of time and if nobody is going to do anything, you should just revert #10669 and drop this.

I think the cleanup is worth it, so chalk this up directly to lack of attention/interest from the project.  Or, this just don't agree. :)

richsalz:

PR #10732, opened 2019-12-31
testutil_init.c: fix compilation error with enable-trace

### LINKED ISSUE #10731: test/testutil/testutil_init.c does not compile with enable-trace
When configured with `enable_trace` I get 
```
gcc  -Iinclude -Iapps/include -I.  -fPIC -pthread -m64 -Wa,--noexecstack -Wall -O0 -g -DDEBUG_UNUSED -DPEDANTIC -pedantic -Wno-long-long -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -Wswitch -Wsign-compare -Wshadow -Wformat -Wtype-limits -Wundef -Werror -Wmissing-prototypes -Wstrict-prototypes -std=gnu90 -DOPENSSL_USE_NODELETE -DL_ENDIAN -DOPENSSL_PIC -DOPENSSLDIR="\"/usr/local/ssl\"" -DENGINESDIR="\"/usr/local/lib/engines-3\"" -DMODULESDIR="\"/usr/local/lib/ossl-modules\""   -MMD -MF test/testutil/libtestutil-lib-testutil_init.d.tmp -MT test/testutil/libtestutil-lib-testutil_init.o -c -o test/testutil/libtestutil-lib-testutil_init.o test/testutil/testutil_init.c
test/testutil/testutil_init.c: In function 'internal_trace_cb':
test/testutil/testutil_init.c:39:35: error: 'PREFIX_CTRL_SET_PREFIX' undeclared (first use in this function); did you mean 'BIO_CTRL_SET_PREFIX'?
         BIO_ctrl(trace_data->bio, PREFIX_CTRL_SET_PREFIX,
                                   ^~~~~~~~~~~~~~~~~~~~~~
                                   BIO_CTRL_SET_PREFIX
test/testutil/testutil_init.c:39:35: note: each undeclared identifier is reported only once for each function it appears in
test/testutil/testutil_init.c: In function 'setup_trace_category':
test/testutil/testutil_init.c:79:32: error: implicit declaration of function 'apps_bf_prefix'; did you mean 'BIO_f_prefix'? [-Werror=implicit-function-declaration]
     channel = BIO_push(BIO_new(apps_bf_prefix()),
                                ^~~~~~~~~~~~~~
                                BIO_f_prefix
test/testutil/testutil_init.c:79:32: error: passing argument 1 of 'BIO_new' makes pointer from integer without a cast [-Werror=int-conversion]
     channel = BIO_push(BIO_new(apps_bf_prefix()),
                                ^~~~~~~~~~~~~~~~
In file included from include/openssl/trace.h:15,
                 from test/testutil/testutil_init.c:12:
include/openssl/bio.h:587:6: note: expected 'const BIO_METHOD *' {aka 'const struct bio_method_st *'} but argument is of type 'int'
 BIO *BIO_new(const BIO_METHOD *type);
      ^~~~~~~
make[1]: *** [Makefile:20369: test/testutil/libtestutil-lib-testutil_init.o] Error 1
```

mspncp:
In commit e79ae962fbed the tests were adapted to use the
new BIO_f_prefix() API which was introduced in 319cee9e2fc6.
This location was missed, because it is compiled only when
tracing is enabled.

Fixes #10731

### DIFF
--- a/test/testutil/testutil_init.c
+++ b/test/testutil/testutil_init.c
@@ -36,8 +36,7 @@ static size_t internal_trace_cb(const char *buf, size_t cnt,
         BIO_snprintf(buffer, sizeof(buffer), "TRACE[%s]:%s: ",
                      hex, OSSL_trace_get_category_name(category));
         OPENSSL_free(hex);
-        BIO_ctrl(trace_data->bio, PREFIX_CTRL_SET_PREFIX,
-                 strlen(buffer), buffer);
+        BIO_set_prefix(trace_data->bio, buffer);
         break;
     case OSSL_TRACE_CTRL_WRITE:
         ret = BIO_write(trace_data->bio, buf, cnt);
@@ -45,7 +44,7 @@ static size_t internal_trace_cb(const char *buf, size_t cnt,
     case OSSL_TRACE_CTRL_END:
         trace_data->ingroup = 0;
 
-        BIO_ctrl(trace_data->bio, PREFIX_CTRL_SET_PREFIX, 0, NULL);
+        BIO_set_prefix(trace_data->bio, NULL);
         break;
     }
 
@@ -76,7 +75,7 @@ static void setup_trace_category(int category)
     if (OSSL_trace_enabled(category))
         return;
 
-    channel = BIO_push(BIO_new(apps_bf_prefix()),
+    channel = BIO_push(BIO_new(BIO_f_prefix()),
                        BIO_new_fp(stderr, BIO_NOCLOSE | BIO_FP_TEXT));
     trace_data = OPENSSL_zalloc(sizeof(*trace_data));
 


test/testutil/testutil_init.c
```diff
@@ -36,16 +36,15 @@ static size_t internal_trace_cb(const char *buf, size_t cnt,
         BIO_snprintf(buffer, sizeof(buffer), "TRACE[%s]:%s: ",
                      hex, OSSL_trace_get_category_name(category));
         OPENSSL_free(hex);
-        BIO_ctrl(trace_data->bio, PREFIX_CTRL_SET_PREFIX,
-                 strlen(buffer), buffer);
+        BIO_set_prefix(trace_data->bio, buffer);
```
slontis: This call does a strdup internally which may fail and return 0 (not likely).. Does the error code need to be checked?

test/testutil/testutil_init.c
```diff
@@ -36,16 +36,15 @@ static size_t internal_trace_cb(const char *buf, size_t cnt,
...
         break;
     case OSSL_TRACE_CTRL_WRITE:
         ret = BIO_write(trace_data->bio, buf, cnt);
         break;
     case OSSL_TRACE_CTRL_END:
         trace_data->ingroup = 0;
 
-        BIO_ctrl(trace_data->bio, PREFIX_CTRL_SET_PREFIX, 0, NULL);
+        BIO_set_prefix(trace_data->bio, NULL);
```
slontis: same here

test/testutil/testutil_init.c
mspncp: Hmmm... this is low level tracing code. Failing is not an option imho and placing some error message between the trace messages might be confusing. Anyway,  If the strdup() fails, all that happens is that the prefix will be missing. So I'd tend to keep the code simple and don't do an error check here.

test/testutil/testutil_init.c
mspncp: No strdup() should happen here: The prefix is set to NULL, i.e., only a previously set prefix is cleared.

slontis:
Thanks.. merged to master.

PR #10738, opened 2020-01-02
Run make update

mattcaswell:
The New Year has caused various files to appear out of date to "make
update". This causes Travis to fail. Therefore we update those file.


mattcaswell:
Pushed. Thanks.

PR #10739, opened 2020-01-02
Run make update (1.1.1)

mattcaswell:
The New Year has caused various files to appear out of date to "make
update". This causes Travis to fail. Therefore we update those files.

This is the 1.1.1 version of #10738


mattcaswell:
Pushed. Thanks.

PR #10740, opened 2020-01-02
Deprecate the Low Level Blowfish APIs

mattcaswell:
Applications should instead use the higher level EVP APIs, e.g.
EVP_Encrypt*() and EVP_Decrypt*().

This is marked as WIP because it builds on the commits from #10580. Once that is merged I will rebase this and take it out of WIP.

Please only look at the last commit.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,15 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) All of the low level Blowfish functions have been deprecated including:
+     BF_set_key, BF_encrypt, BF_decrypt, BF_ecb_encrypt, BF_cbc_encrypt,
+     BF_cfb64_encrypt, BF_ofb64_encrypt, and BF_options.
+     Use of these low level functions has been informally discouraged for a long
+     time. Instead applications should use the high level EVP APIs, e.g.
+     EVP_EncryptInit_ex, EVP_EncryptUpdate, EVP_EncryptFinal_ex, and the
+     equivalently named decrypt functions.
+     [Matt Caswell]
+
   *) Removed include/openssl/opensslconf.h.in and replaced it with
      include/openssl/configuration.h.in, which differs in not including
      <openssl/macros.h>.  A short header include/openssl/opensslconf.h
--- a/apps/speed.c
+++ b/apps/speed.c
@@ -384,7 +384,7 @@ static const OPT_PAIR doit_choices[] = {
     {"seed-cbc", D_CBC_SEED},
     {"seed", D_CBC_SEED},
 #endif
-#ifndef OPENSSL_NO_BF
+#if !defined(OPENSSL_NO_BF) && !defined(OPENSSL_NO_DEPRECATED_3_0)
     {"bf-cbc", D_CBC_BF},
     {"blowfish", D_CBC_BF},
     {"bf", D_CBC_BF},
@@ -1461,7 +1461,7 @@ int speed_main(int argc, char **argv)
 #ifndef OPENSSL_NO_SEED
     SEED_KEY_SCHEDULE seed_ks;
 #endif
-#ifndef OPENSSL_NO_BF
+#if !defined(OPENSSL_NO_BF) && !defined(OPENSSL_NO_DEPRECATED_3_0)
     BF_KEY bf_ks;
 #endif
 #ifndef OPENSSL_NO_CAST
@@ -1986,7 +1986,7 @@ int speed_main(int argc, char **argv)
             goto end;
         }
 #endif
-#ifndef OPENSSL_NO_BF
+#if !defined(OPENSSL_NO_BF) && !defined(OPENSSL_NO_DEPRECATED_3_0)
     if (doit[D_CBC_BF])
         BF_set_key(&bf_ks, 16, key16);
 #endif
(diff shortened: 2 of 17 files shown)


mattcaswell:
Rebased and taken out of WIP. Please take a look!

richsalz:
I think the ```OPENSSL_BUILDING_OPENSSL``` concept is more clear than "clever" header file like internal/deprecated.h

mattcaswell:
> I think the OPENSSL_BUILDING_OPENSSL concept is more clear than "clever" header file like internal/deprecated.h

It took numerous attempts, much discussion and many PRs to figure out how to deprecate stuff like this for the AES_ige_*() and the low level AES functions: #10367, #10538, #10540, #10558, #10580, #10608. It's a deceptively complex topic. I don't think reopening the debate on how to do it at this point is going to be helpful. The solution we've arrived it, IMO, is quite straight forward and is already implemented. This PR and my other related ones is just rolling out that agreed approach for other APIs.

levitte:
Travis is unhappy with `no-deprecated`...  it doesn't seem to have to do specifically with *this* PR, but it rather looks like something is missing around AES?

levitte:
A `OPENSSL_BUILDING_OPENSSL` type of solution isn't feasible because it lacks granularity.  There are some functions that we want to abolish even internal use of, while there are others (such as the low level implementations, as they are the basis for our providers) that we want to retain internal use of even if they are publicly deprecated (and eventually removed from public header files).  Hence, we need to be able to select this on a per translation unit basis, not on a per library basis, and including a specific headers is a good enough marker of such selection.

richsalz:
I am confused by your comment.  You use ```OPENSSL_BUILDING_OPENSSL``` at the same header-file granularity.  The OBO is an indicator that it is used internally.  Seems like it providers more information.  Unless I misunderstand the comment.

levitte:
`OPENSSL_BUILDING_OPENSSL` is defined for *all* of libcrypto and libssl.  That's not very granular when you want to selectively continue to use certain deprecated functions internally (such as the blowfish functions), while banishing others completely, even internally (such as, say, the ENGINE API).

levitte:
> Travis is unhappy with `no-deprecated`...  it doesn't seem to have to do specifically with *this* PR, but it rather looks like something is missing around AES?

Fixed in #10766

richsalz:
You put ifdef's using OBO around the functions you still need internally. Am I missing something, or you?

The OBO define tells you when you are building OpenSSL itself, and you can leverage that to get better "deprecated" declarations in public header files.  See the #10724 and #10669 esp the introductory comment on that one.

levitte:
> Am I missing something, or you?

You are.  The thing is that with `--strict-warnings`, an implementation without a declaration will cause an error.  This means that when you want to still compile the functions that are publicly deprecated, you have to undefine `OPENSSL_NO_DEPRECATED` and define `OPENSSL_SUPPRESS_DEPRECATED` *in that translation unit*.  The whole mechanism to do so is this:

``` C

#undef OPENSSL_NO_DEPRECATED
#define OPENSSL_SUPPRESS_DEPRECATED

#include <openssl/macros.h>
```

... and that's *exactly* what `internal/deprecated.h` does.

I have *no* idea how OBO helps, 'cause in the translation units that make up the libraries, *we already know that we're building the libraries*

richsalz:
Your must recent explanation makes me think that you are missing something.  Sorry if I am not clear.  I know I wrote this somewhere in some PR, but I can't find the comment.

The following cases are handled by #10767 as an example for the other (stack and lhash and lhash-cruft) PR's.

1. We want to deprecate and we don't use internally; use the existing DEPRECATED macros.
1. We want to deprecate, but we use it internally; use the OBO test and the DEPRECATED macro

richsalz:
Perhaps the point you are missing is that in a public OpenSSL header file, if it is being *used to build openssl source* and we want to use the deprecated functions internally (such as in test code for the low-level crypto API's), then we want to global define OBO so that it can be used in *any file* that is part of *any translation unit* without touching a whole bunch of files.  As I detailed in #10767 that method, which is what was done for hte AES deprecation, is error-prone.

levitte:
> We want to deprecate, but we use it internally; use the OBO test and the DEPRECATED macro

I assume that to do this, we would have to do the OBO test somewhere in the public headers.  That will expose publicly exactly what functions we still use internally and which ones will be truly removed.  In previous discussions, we (well, we who discussed it) concluded that we had no interest in such exposure.

levitte:
The error-prone isn't very hard.  If we miss something somewhere, the CIs will fail.  It's that simple.

levitte:
@mattcaswell, #10766 has been merged, please rebase

richsalz:
>I assume

Well, yeah.  And you don't have to assume, you could look at the PR's I've already mentioned multiple times :)

A build failure shouldn't be a casual thing.

Shrug.  I guess this is a dead issue.  Too bad.

mattcaswell:
Rebased

levitte:
> And you don't have to assume, you could look at the PR's I've already mentioned multiple times :)

#10767 doesn't show me anything, all you've done in there is to add the OBO macro, but you haven't demonstrated how they would be used for the sort of selection implemented here.  The other two are very different topics...

mattcaswell:
Pushed. Thanks.

PR #10747, opened 2020-01-03
Don't store an HMAC key for longer than we need

### LINKED ISSUE #10743: HMAC_Init_ex failed to clean up the key buffer, leaving the HMAC key in plaintext
When HMAC_Init_ex() returned, the secret key for HMAC was kept in the `key[HMAC_MAX_MD_CBLOCK_SIZE]` inside HMAC_CTX (aka 'hmac_ctx_st') in plaintext, which is a security hole: hackers can obtain the secret key for HMAC by dumping the memory. In fact, this key buffer in HMAC_CTX was not used any more after the secretKey was supplied to HMAC_Init_ex(). 

In OpenSSL_1.0.2, we cleaned up the key buffer ourselves:
````
  HMAC_CTX_init(&m_ctx);
  HMAC_Init_ex( &m_ctx,  (const void*)(secretKey), keyLength,  HashAlgo(),   NULL);

  // erase key buffer inside hmac ctx
  memset(m_ctx.key, 0, m_ctx.key_length);
  m_ctx.key_length = 0;
````
However, because OpenSSL_1.1.1 makes the HMAC_CTX opaque, we cannot access the key buffer in HMAC_CTX like below,

````
  m_ctx = HMAC_CTX_new();
  HMAC_Init_ex(m_ctx,  (const void*)(secretKey),  keyLength, HashAlgo(), NULL);

  // erase key buffer inside hmac ctx
  memset(m_ctx->key, 0, m_ctx->key_length);
  m_ctx->key_length = 0;
````
we got  compilation `error: member access into incomplete type 'HMAC_CTX' (aka 'hmac_ctx_st')
  memset(m_ctx->key, 0, m_ctx->key_length);`

Our experiments show that to complete 100million HMAC digests using the same key,  init HMAC_CTX with secret key only once will be much faster.

Therefore, the key buffer in 'HMAC_CTX' (aka 'hmac_ctx_st') should be cleared up before HMAC_Init_ex() returns to keep good performance and avoid leaking the secret key.

### PR DESCRIPTION
**Pull request by mattcaswell**
The HMAC_CTX structure stores the original key in case the ctx is reused
without changing the key.

However, HMAC_Init_ex() checks its parameters such that the only code path
where the stored key is ever used is in the case where HMAC_Init_ex is
called with a NULL key and an explicit md is provided which is the same as
the md that was provided previously. But in that case we can actually reuse
the pre-digested key that we calculated last time, so we can refactor the
code not to use the stored key at all.

With that refactor done it is no longer necessary to store the key in the
ctx at all. This means that long running ctx's will not keep the key in
memory for any longer than required. Note though that the digested key
*is* still kept in memory for the duration of the life of the ctx.

Fixes #10743

Note: It is debatable whether this should come under the "bug" rules or not, and therefore be eligible for backport to 1.1.1. Since it provides no user visible features, and is a security hardening measure I have flagged it for 1.1.1.

### DIFF
--- a/crypto/hmac/hmac.c
+++ b/crypto/hmac/hmac.c
@@ -18,16 +18,17 @@
 int HMAC_Init_ex(HMAC_CTX *ctx, const void *key, int len,
                  const EVP_MD *md, ENGINE *impl)
 {
-    int rv = 0;
-    int i, j, reset = 0;
+    int rv = 0, reset = 0;
+    int i, j;
     unsigned char pad[HMAC_MAX_MD_CBLOCK_SIZE];
+    unsigned int keytmp_length;
+    unsigned char keytmp[HMAC_MAX_MD_CBLOCK_SIZE];
 
     /* If we are changing MD then we must have a key */
     if (md != NULL && md != ctx->md && (key == NULL || len < 0))
         return 0;
 
     if (md != NULL) {
-        reset = 1;
         ctx->md = md;
     } else if (ctx->md) {
         md = ctx->md;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by lantse-cap1** [comment]
A very nice fix. Thank you!

File: crypto/hmac/hmac.c [hunk 1]
```diff
@@ -18,16 +18,17 @@
 int HMAC_Init_ex(HMAC_CTX *ctx, const void *key, int len,
                  const EVP_MD *md, ENGINE *impl)
 {
-    int rv = 0;
-    int i, j, reset = 0;
+    int rv = 0, reset = 0;
+    int i, j;
```
paulidale: Kind of odd moving this up a line.

mspncp: It looked more odd as it was before. Now there are two zero-initialized local variables in line 21 and two uninitialized counter variables in line 22.

paulidale: I'm not suggesting it gets changed back, I agree it's better this way.  I just thought it odd to make a non-essential change.

mattcaswell: At one point during developing this change I removed the `reset` variable altogether. I later realised I still needed it so put it back again. Inadvertently it went to a slightly different place. But since everyone seems ok with it, and even to prefer it this way, I'll leave it.

mattcaswell:
Pushed to master. The 1.1.1 version apparently needs a slight adjustment. I'll raise a new PR for that.

PR #10753, opened 2020-01-04
Configuration: synchronise the variables on the build file templates

levitte:
For some reason, we didn't use some of the possible target attributes
in the Unix Makefile template, and there was a similar but much
smaller lack of use in the Windows makefile template as well.

### DIFF
--- a/Configurations/00-base-templates.conf
+++ b/Configurations/00-base-templates.conf
@@ -10,7 +10,7 @@ my %targets=(
 	includes	=> [],
 	lib_cflags	=> "",
 	lib_cppflags	=> "",
-	lib_defines	=> [],
+	lib_defines	=> [ 'OPENSSL_BUILDING_OPENSSL' ],
 	thread_scheme	=> "(unknown)", # Assume we don't know
 	thread_defines	=> [],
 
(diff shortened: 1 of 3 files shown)


mattcaswell (approved):

levitte:
Merged.

d368d9d2e4bce444b3c56db55aab5f8450814209 Configuration: synchronise the variables on the build file templates

richsalz:
I don't have permission to re-open this.  Ping @levitte, @openssl/comitters, whoever.  This does not work.  See https://api.travis-ci.org/v3/job/635494784/log.txt and look for ```OPENSSL_BUILDING_OPENSSL``` it shows up once.  It should show up in every compilation.

levitte:
Er, you do remember what `-s` in `make - s` does, right ?

richsalz:
yeah, maybe that's not the issue.  never mind.

richsalz:
@levitte, Look at https://ci.appveyor.com/project/openssl/openssl/builds/30049497/job/e17342lihomp389e#L1161  That doesn't appear to be "make -s" and I don't see the define present.

levitte:
Ah, figured it out, the definition was a bit misplaced.  However, now that I'm looking at it again, that definition should be attached to the public libraries we build, and is actually better placed in the dop build.info.

PR coming up!

levitte:
See #10813

Issue #10049, opened 2019-09-29
refactor random-related stuff out of e_os.h


richsalz:
I think only the crypto/rand files need the random stuff, so you can move 45 lines out of e_os.h to a separate file, or perhaps crypto/rand/rand_local.h ?

mspncp:
Good idea! See #10050.

PR #10761, opened 2020-01-06
Move random-related defines to "crypto/rand.h" [1.1.1]

### LINKED ISSUE #10759: build fail in relase branch due to missing DEVRANDOM related defines
Build of 1.1.1 stable branch now fail in apps/version.c. This is regression in last 3-4 months.

Releases follows an quite long process - alpha , beta, release candidates. In those phases external developers test programs. After first official release users start to test ...
So regression after 4 releases of 1.1.1 in absolutely unexpected.

I will lest as exercise build configuration that shows failure (linux).

mspncp:
This fixes commit 7b18d1a53f932391bbc599a4717d6f98a597849c, which moved the
DEVRANDOM and DEVRANDOM_EGD defines into rand_unix.c. That change introduced
the regression that the compiler complains about missing declarations in
crypto/info.c when OpenSSL is configured using `--with-rand-seed=devrandom`
(resp. `--with-rand-seed=egd`)

Fixes #10759

### DIFF
--- a/apps/version.c
+++ b/apps/version.c
@@ -31,6 +31,8 @@
 # include <openssl/blowfish.h>
 #endif
 
+#include "crypto/rand.h"
+
 typedef enum OPTION_choice {
     OPT_ERR = -1, OPT_EOF = 0, OPT_HELP,
     OPT_B, OPT_D, OPT_E, OPT_F, OPT_O, OPT_P, OPT_V, OPT_A, OPT_R
--- a/crypto/rand/rand_unix.c
+++ b/crypto/rand/rand_unix.c
@@ -20,55 +20,6 @@
 #include <stdio.h>
 #include "internal/dso.h"
 
-/*
- * Defines related to seed sources
- */
-#ifndef DEVRANDOM
-/*
- * set this to a comma-separated list of 'random' device files to try out. By
- * default, we will try to read at least one of these files
- */
-# define DEVRANDOM "/dev/urandom", "/dev/random", "/dev/hwrng", "/dev/srandom"
-# if defined(__linux) && !defined(__ANDROID__)
-#  ifndef DEVRANDOM_WAIT
-#   define DEVRANDOM_WAIT   "/dev/random"
-#  endif
-/*
- * Linux kernels 4.8 and later changes how their random device works and there
- * is no reliable way to tell that /dev/urandom has been seeded -- getentropy(2)
- * should be used instead.
- */
-#  ifndef DEVRANDOM_SAFE_KERNEL
-#   define DEVRANDOM_SAFE_KERNEL        4, 8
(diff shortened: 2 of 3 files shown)


mspncp:
> The patch looks OK. Although another option would be to just revert the commit that introduced the regression.

I did in fact consider this option for 1.1.1, but that would have had the disadvantage that the code, which originally came from `e_os.h`  would have ended up in different files in master and 1.1.1. (code cleanup of `e_os.h` was suggested by @richsalz in #10049).

apps/version.c
```diff
@@ -31,6 +31,8 @@
 # include <openssl/blowfish.h>
 #endif
 
+#include "crypto/rand.h"
```
levitte: I really really dislike this.

mspncp: What exactly do you dislike? I suppose it's the fact that a crypto-only header is used by the app? Would you prefer if I just reverted 7b18d1a53f932391bbc599a4717d6f98a597849c?

levitte: > I suppose it's the fact that a crypto-only header is used by the app?

Yes.  Let's not confuse those boundaries.

> Would you prefer if I just reverted 7b18d1a?

Yes, I think that's the more viable solution.  It's still ugly, but at least doesn't *add on* more ugliness.

(another solution might be to add a `OSSL_DEVRANDOM` in opensslconf.h, just a simple definition without any ifdef sallad, and use that instead)

mspncp: > > Would you prefer if I just reverted [7b18d1a](https://github.com/openssl/openssl/commit/7b18d1a53f932391bbc599a4717d6f98a597849c)?
> 
> Yes, I think that's the more viable solution. It's still ugly, but at least doesn't _add on_ more ugliness.

The reverting commit was already prepared, just pushed it. See #10764.

mspncp: > (another solution might be to add a `OSSL_DEVRANDOM` in opensslconf.h, just a simple definition without any ifdef sallad, and use that instead)

Nope. That means adding a new feature to a public header.

mspncp:
Ok, Tim approved #10764 in record time. I feel consent about reverting instead of fixing on 1.1.1.

apps/version.c
levitte: > Nope. That means adding a new feature to a public header.

...  there are times when we get overly strict with that...  but fine.

mspncp: > ... there are times when we get overly strict with that... but fine.

This regression is the example that we should be careful with adding unnecessary 'little' changes to stable branches. It wasn't even in public headers, but still broke some code.

> without any ifdef sallad

I agree with you that the entropy source ifdef salad is a technical debt and really needs to be cleaned up (in master).

PR #10762, opened 2020-01-06
Move random-related defines to "crypto/rand.h"

### LINKED ISSUE #10759: build fail in relase branch due to missing DEVRANDOM related defines

mspncp:
commit 7ec5c3109d5cea3d2b264ec1db61fd42ac97ec13
Author: Dr. Matthias St. Pierre <Matthias.St.Pierre@ncp-e.com>
Date:   Mon Jan 6 01:50:43 2020 +0100

    Move random-related defines to "crypto/rand.h"
    
    This fixes commit 01036e2afbe116d608be048ed15930fc885ab2a8, which moved the
    DEVRANDOM and DEVRANDOM_EGD defines into rand_unix.c. That change introduced
    the regression that the compiler complains about missing declarations in
    crypto/info.c when OpenSSL is configured using `--with-rand-seed=devrandom`
    (resp. `--with-rand-seed=egd`)
    
    Fixes #10759

commit ee502fd524d52602a61b67425f5e7f1de6f98a11 (HEAD -> pr-move-random-defines-again, msp/pr-move-random-defines-again)
Author: Dr. Matthias St. Pierre <Matthias.St.Pierre@ncp-e.com>
Date:   Mon Jan 6 02:38:14 2020 +0100

    Modify the add_seeds_stringlist() macro to fix a preprocessor error
    
    When OpenSSL is configured using `--with-rand-seed=devrandom`, the preprocessor
    reports the following error
    
        crypto/info.c:104:66: error:
                macro "add_seeds_stringlist" passed 3 arguments, but takes just 2
                add_seeds_stringlist("random-device", { DEVRANDOM, NULL });
    
    The reason why the preprocessor complains about three arguments being passed
    is that according to [1], balanced braces in macro arguments don't prevent the
    comma from acting as an argument separator:
    
        3.3 Macro Arguments
        ...
        Parentheses within each argument must balance;
        a comma within such parentheses does not end the argument.
        However, there is no requirement for square brackets or braces to balance,
        and they do not prevent a comma from separating arguments.
    
    Also introduced an iteration pointer `p`, because `dev` is not an lvalue:
    
        crypto/info.c:78:41: error:
                lvalue required as increment operand
                for (; *dev != NULL; dev++) {
    
    [1] https://gcc.gnu.org/onlinedocs/cpp/Macro-Arguments.html

### DIFF
--- a/crypto/info.c
+++ b/crypto/info.c
@@ -76,7 +76,7 @@ DEFINE_RUN_ONCE_STATIC(init_info_strings)
                 const char **p;                                         \
                 int first = 1;                                          \
                                                                         \
-                for (p=dev; *p != NULL; p++) {                          \
+                for (p = dev; *p != NULL; p++) {                        \
                     if (!first)                                         \
                         OPENSSL_strlcat(seeds, " ", sizeof(seeds));     \
                     first = 0;                                          \
--- a/crypto/info.c
+++ b/crypto/info.c
@@ -72,14 +72,15 @@ DEFINE_RUN_ONCE_STATIC(init_info_strings)
         do {                                                            \
             add_seeds_string(label "(");                                \
             {                                                           \
-                const char *dev[] = strlist;                            \
+                const char *dev[] =  { strlist, NULL };                 \
+                const char **p;                                         \
                 int first = 1;                                          \
                                                                         \
-                for (; *dev != NULL; dev++) {                           \
+                for (p=dev; *p != NULL; p++) {                          \
                     if (!first)                                         \
                         OPENSSL_strlcat(seeds, " ", sizeof(seeds));     \
                     first = 0;                                          \
-                    OPENSSL_strlcat(seeds, *dev, sizeof(seeds));        \
+                    OPENSSL_strlcat(seeds, *p, sizeof(seeds));          \
                 }                                                       \
             }                                                           \
             OPENSSL_strlcat(seeds, ")", sizeof(seeds));                 \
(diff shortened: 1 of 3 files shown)


crypto/info.c
```diff
@@ -101,10 +102,10 @@ DEFINE_RUN_ONCE_STATIC(init_info_strings)
         add_seeds_string("getrandom-syscall");
 #endif
 #ifdef OPENSSL_RAND_SEED_DEVRANDOM
-        add_seeds_stringlist("random-device", { DEVRANDOM, NULL });
+        add_seeds_stringlist("random-device", DEVRANDOM);
```
mspncp: @levitte I guess that `add_seeds_stringlist()` went into 096978f09908 without ever being expanded, let alone compiled?

levitte: [ahem] looks that way...

crypto/info.c
```diff
@@ -71,14 +72,15 @@ DEFINE_RUN_ONCE_STATIC(init_info_strings)
         do {                                                            \
             add_seeds_string(label "(");                                \
             {                                                           \
-                const char *dev[] = strlist;                            \
+                const char *dev[] =  { strlist, NULL };                 \
+                const char **p;                                         \
                 int first = 1;                                          \
                                                                         \
-                for (; *dev != NULL; dev++) {                           \
+                for (p=dev; *p != NULL; p++) {                          \
```
levitte: Spaces. `p = dev`

mspncp: Fixed by 424be4a8adc09701310406fc2396e8fe246fe7b2.

mspncp:
Side note: the fix was compiled and tested with `--with-rand-seed=devrandom`. Here's the output:

```
$ util/shlib_wrap.sh  apps/openssl info -seeds
random-device(/dev/urandom /dev/random /dev/hwrng /dev/srandom)
```

mspncp:
Merged to master, thanks!

9484b67dfb Modify the add_seeds_stringlist() macro to fix a preprocessor error
2e912f63a4 Move random-related defines to "crypto/rand.h"

PR #10766, opened 2020-01-06
Add missing inclusion of "internal/deprecated.h"

levitte:
A few provider implementations need this to build correctly with a
'no-deprecated' configuration.

PR #10666, opened 2019-12-20
Add -DOPENSSL_BUILDING_OPENSSL to our builds

richsalz:
Our public header files are also used internally. This can cause issues when we want things "slightly different" when building internally: for example, we want to build with no-deprecated *for some things* but we want to allow users to have a choice.

See https://github.com/openssl/openssl/issues/10657#issuecomment-567836928 and the links it points to, etc.

This PR adds a define OPENSSL_BUILDING_OPENSSL which is only set in our internal Makefiles.
This means we can do things like this in files that use lhash.h but don't want the static inline functions declared when building openssl.
```
#if defined(OPENSSL_BUILDING_OPENSSL) || !defined(OPENSSL_NO_DEPRECATED_3)
... old code ...
DEFINE_STACK_OF(xxxx)
#else
... new code ...
STACK_OF(xxxx);
#endif 
```

This can be used in other places when we want to deprecate internally but not externally, such as perhaps the recent AES low-level functions #10587

PR #10767, opened 2020-01-06
WIP: Better deprecation

richsalz:
This uses #10666 to show another way to do the deprecation that was done in #10587 

Note that the change is much smaller (only aes.h changes, not 28 other files as well) and much less error-prone (see #10766) because of that.

It is a WIP because ParseC doesn't understand the if construct used. :(

I understand this issue has been decided.  I hope this re-opens it.


levitte:
This is a huge pile of commits from diverse sources...  That makes it very hard to see what you want to demonstrate

richsalz:
Cleaned up most of the "huge pile of commits"  Will a WIP.

levitte:
An issue with this idea is that it forces us to expose publicly exactly what functions will only have their public declaration removed (i.e. they are going to be kept internally), and what functions are going to be removed entirely.  There's no reason why this should be exposed.

PR #10771, opened 2020-01-07
EVP: If a key can't be exported to provider, fallback to legacy

levitte:
Currently, the operations that do try to export a legacy key to
providers will fail if the export failed.  It makes more sense to
simply use the legacy method instead, as a fallback for things not
being implemented (yet) in a provider.


mattcaswell (approved):
LGTM

levitte:
Merged.

e0d8523e801b7a1fcdda698f9c28dd7a0617cd02 EVP: If a key can't be exported to provider, fallback to legacy

Issue #10542, opened 2019-11-29
Incorrect Key Size for Some EC Curves


git-blame:
This is on version 1.1.0l but may also affect newer versions. This public key for curve sect163r1 should be 163 bits.

```
openssl ec -in keys/alice2.pub -pubin -text
read EC key
Public-Key: (162 bit)
pub:
    04:02:e9:7f:a0:12:ab:f4:1e:4c:b0:f1:d6:18:5f:
    aa:d3:b8:f4:24:2f:96:03:18:f5:27:37:dc:21:74:
    7d:f2:7f:ab:f4:51:d6:53:92:c2:e6:90:cc
ASN1 OID: sect163r1
writing EC key
-----BEGIN PUBLIC KEY-----
MEAwEAYHKoZIzj0CAQYFK4EEAAIDLAAEAul/oBKr9B5MsPHWGF+q07j0JC+WAxj1
JzfcIXR98n+r9FHWU5LC5pDM
-----END PUBLIC KEY-----
```
A number of functions use `EC_GROUP_order_bits` (for example  `EVP_PKEY_bits`, `do_EC_KEY_print`) to calculate key/curve size.

The typical underlying function is `ec_group_simple_order_bits` which counts the number of bits in the BIGNUM group order.

However, as seen in curves like sect163r1, the order value can be 1-bit smaller than the actual curve size.

From SEC spec:
```
Finally the order n of G and the cofactor are:
n=03 FFFFFFFF FFFFFFFF FFFF48AA B689C29C A710279B
h=02
```
It may be better to use `EC_GROUP_get_degree`. From the openssl documentation:
> EC_GROUP_get_degree gets the degree of the field. For Fp fields this will be the number of bits in p. For F2^m fields this will be the value m.

Testing it with the same key:
```
(gdb) p EVP_PKEY_bits( key )
$22 = 162
(gdb) p EC_GROUP_order_bits( g )
$23 = 162
(gdb) p EC_GROUP_get_degree( g )
$24 = 163
```

mattcaswell:
@romen?

romen:
Thanks for reporting this @git-blame 

> This is on version 1.1.0l but may also affect newer versions. This public key for curve sect163r1 should be 163 bits.

This is not necessarily true: the 163 in the name refers to the underlying filed size not to the order of the curve.

[...]

PR #10778, opened 2020-01-08
EVP: make EVP_PKEY_{bits,security_bits,size} work with provider only keys

levitte:
These functions relied entirely on the presence of 'pkey->pmeth',
which is NULL on provider only keys.  This adds an interface to get
domparam and key data from a provider, given corresponding provider
data (the actual domparam or key).

The retrieved data is cached in the EVP_PKEY structure (lending the
idea from provided EVP_CIPHER).

##### Checklist


levitte:
Note that #10289 is now pending on this as well

romen (changes requested):
Seems fine at a first look, I will test more later on.

But I notice a doc update is missing in this PR and maybe it could also be the right PR to add simple tests for the involved EVP_PKEY_* functions.

Bonus points if you want to revise the doc for EVP_PKEY_bits vs security_bits because last time I checked I had trouble understanding the exact definition of the returned values!

levitte:
I have tweaked a test for this already. Docs to be added today

crypto/evp/keymgmt_lib.c
```diff
@@ -161,6 +157,48 @@ void evp_keymgmt_clear_pkey_cache(EVP_PKEY *pk)
                 keymgmt->freekey(provdata);
             EVP_KEYMGMT_free(keymgmt);
         }
+
+        pk->bits = 0;
```
slontis: why do you only clear bits here and not the other 2 cached fields?

providers/implementations/keymgmt/dh_kmgmt.c
```diff
@@ -196,5 +214,6 @@ const OSSL_DISPATCH dh_keymgmt_functions[] = {
     { OSSL_FUNC_KEYMGMT_IMPORTKEY, (void (*)(void))dh_importkey },
     { OSSL_FUNC_KEYMGMT_EXPORTKEY, (void (*)(void))dh_exportkey },
     { OSSL_FUNC_KEYMGMT_FREEKEY, (void (*)(void))DH_free },
+    { OSSL_FUNC_KEYMGMT_GET_KEY_PARAMS,  (void (*) (void))dh_get_key_params },
```
slontis: Is this supposed to have a get_domain_params entry?

levitte:
Documentation added.  @romen, are you happy with the result?

crypto/evp/keymgmt_lib.c
levitte: 'cause sloppy.  Fixed.  Thanks for catching it.

providers/implementations/keymgmt/dh_kmgmt.c
levitte: Yes.  The same goes for the DSA implementation, of course

levitte: Fixed

levitte:
Finally got all the docs bits right.  Final review, please?

providers/implementations/keymgmt/rsa_kmgmt.c
```diff
@@ -244,11 +245,29 @@ static const OSSL_PARAM *rsa_importkey_types(void)
...
+    return 1;
+}
+
 const OSSL_DISPATCH rsa_keymgmt_functions[] = {
     { OSSL_FUNC_KEYMGMT_IMPORTKEY, (void (*)(void))rsa_importkey },
     { OSSL_FUNC_KEYMGMT_IMPORTKEY_TYPES, (void (*)(void))rsa_importkey_types },
     { OSSL_FUNC_KEYMGMT_EXPORTKEY, (void (*)(void))rsa_exportkey },
     { OSSL_FUNC_KEYMGMT_EXPORTKEY_TYPES, (void (*)(void))rsa_exportkey_types },
     { OSSL_FUNC_KEYMGMT_FREEKEY, (void (*)(void))RSA_free },
+    { OSSL_FUNC_KEYMGMT_GET_KEY_PARAMS,  (void (*) (void))rsa_get_key_params },
```
slontis: OSSL_FUNC_KEYMGMT_GET_DOMPARAM_PARAMS

doc/man3/EVP_PKEY_size.pod
```diff
@@ -0,0 +1,39 @@
...
+I<pkey>, bits of security is defined in NIST SP800-57.
+
+=head1 RETURN VALUES
+
+EVP_PKEY_size(), EVP_PKEY_bits() and EVP_PKEY_security_bits() return a
+positive number, or 0 if this size isn't available.
+
+=head1 COPYRIGHT
+
+Copyright 2000-2018 The OpenSSL Project Authors. All Rights Reserved.
```
slontis: Maybe update copyright?

doc/man3/EVP_PKEY_size.pod
```diff
@@ -0,0 +1,39 @@
...
+
+ int EVP_PKEY_size(const EVP_PKEY *pkey);
+ int EVP_PKEY_bits(const EVP_PKEY *pkey);
+ int EVP_PKEY_security_bits(const EVP_PKEY *pkey);
+
+=head1 DESCRIPTION
+
+EVP_PKEY_size() returns the size of the key in I<pkey> in bytes.
+
+EVP_PKEY_bits() returns the size of the key in I<pkey> in bits.
```
romen: I would change the wording here to something similar to:

```
EVP_PKEY_bits() returns the cryptographic length in bits
of the cryptosystem to which the key in I<pkey> belongs.
Notice that the definition of cryptographic length is specific
to the key cryptosystem.
```
The difference is that a reader could be misled to think that this is some kind of measure of how big a buffer to contain some encoding of this key should be (see the recent #10542): it is not, and actually for some cryptosystems (e.g. DSA, ECDSA, DH, ECDH) this function does not give an information about the specific length of what constitutes the public key or the private key encapsulated by `pkey` as a particular instance, but rather a property of the domain parameters that define the cryptosystem `pkey` belongs to.
Using an analogy, if `DSA-2048` was a class and the argument `pkey` in a call to `EVP_PKEY_bits(pkey)` an instance of that class, the return value would be `2048`, and it would be a class attribute, not a instance attribute.

This view can actually be extended to RSA as well because even though there aren't really "domain parameters" associated with a key, and the _cryptographic length_ comes from the length of the instance attribute `n`, by construction we generally pick first an arbitrary bit length from a conventional list `{1024, 2048, 3072, 4096, ...}` and then we look for two primes that gives us a modulus of the desired length, so, maybe a bit forcefully, the key length could be seen as a "domain parameter" that we pick to define all the security properties of the cryptosystem to which our key belongs.

Anyway, this is the difference I try to highlight when I talk about _cryptographic length_: it should serve as a warning to the reader that it might differ from the conventional definition of the _concrete_ bit length of something stored or transmitted by a digital system, and that the nature of the return value of `EVP_PKEY_bits()` is somewhat more abstract.

## Some nasty details

I believe it is important to state that the definition of what constitutes the _cryptographic length_ of a key is part of the specific definition of the cryptosystem, as it is an arguably arbitrary choice of the designers.

### DSA keys (applies almost identically to DH as well)

- OpenSSL for DSA keys returns `BN_num_bits(pkey->pkey.dsa->p)`
- The DSA design defines the _length_ of the key as _L_, the bit length of the public parameter `p`; another relevant bit length is _N_ pertaining to the public parameter `q`.
- The value `x` that constitutes the private key is chosen randomly from `{1 ... (q - 1)}`, so its bit length is between 1 and _N_
- The value `y = g^x mod(p)` that constitutes the actual public key has a bit length between 1 and _L_
- In the cryptosystem as a whole we do some math modulo `q` and some math modulo `p`; the most computationally intense operations are done in the multiplicative group defined by `p`, so _L_ can be considered a metric for the computation level required for operations belonging to this cryptosystem; the definition of _L_ as the _key length_ also allowed direct comparison with the _key length_ of RSA (as the alternative and more well established asymmetric cryptosystem)

Given a chosen key length _L_, the _N_ length is basically picked through a table.
_L_ as a length is more closely related to the number of bits to transfer on the wire to distribute a public key (ignoring the fact that most likely you will need to transfer every time also the associated parameters, thus more than doubling the actual bits of information transferred), and only indirectly to the bits of security, which instead is more closely related to _N_ (intuitively, limiting us to a brute-force attack, there are at most `2^N` different private keys to explore to break the DSA cryptosystem with a given set of parameters `(p, q)`).

The choice of _L_ over _N_ as a measure of the _length_ of the key is thus a somewhat arbitrary design decision.
And it is also evident that this _cryptographic length_ comes from the parameters associated with the argument of the `EVP_PKEY_bits()` call, rather then being directly related to the size/length required to encode the particular key passed to the function.

### RSA keys

- OpenSSL for RSA keys returns `BN_num_bits(pkey->pkey.rsa->n)`
- The RSA design defines that the _length_ of a key is the bit length of the modulus `n` associated with the key pair, the secret key is the triplet `(p, q, d)` with `p * q = n`, and the public key is the pair `(e, n)`
- Complexity here is asymmetric, you optimize either for fast encryptions or decryptions, but roughly one could use this definition of length as a measure of complexity, area, or memory requirements for the cryptosystem operations
- The brute-force attack here does not lead far, as by construction `bitlen(p)` and `bitlen(q)` differ by a few bits, so blind brute-force requires more or less `bitlen(n)/2` steps; the best sieving attacks have well studied complexity though, so one can apply a formula to the bit length of `n` and directly derive the bits of strenght of a key (but generally one just uses a table because anyway we pick well determined lengths for `n` as the preliminary step of key generation).

It is again evident though, that the bit length of `n` is not the size of the private key (as `(p, q, d)` is already on average around `1.5 * n` and then usually we actually store more stuff to do math tricks for faster computations) but is close enough to the public key (`e` is usually small, so `(n, e)` is close in length.
Anyway it is definitely not the size in bits of any encoding of the specific key passed as an argument, but again the _length_ of the abstract class to which that key belongs.

### EC keys

- OpenSSL returns `EC_GROUP_order_bits(EC_KEY_get0_group(pkey->pkey.ec))`
- According to FIPS-186-3, for ECDSA, the order of the base point of the elliptic curve is `n`; and the bit length of `n` is considered to be the key size.
- The complications start when we note that in the cryptosystem we do some operations modulo `n`, the order of the EC group generated by the base point, and some operations modulo `q`, where `q` is the size of the field upon which the elliptic curve group is built (this means that the coordinates of any point on the curve are elements of the field and their maximum bit length is `bitlen(q)`). To complicate matters, usually the names assigned to standardized curves include a number, which corresponds to `q` rather than `n`. The bit sizes of `q` and `n` are bound by the Hasse theorem so that at most they differ by 1, but this can still lead to confusion as a curve named `sect163r1` has _key length_ 162 and this can appear as an off-by-one mistake.
- Another potential source of confusion is that in ECDH/ECDSA, `n` is directly related to the size of the **private key** rather than the public key (the opposite of what happens for DSA/DH/RSA where the `key length` is more closely related to the public key):
  + the secret key `d` is picked randomly in the range `{1, ... , n-1}`, so at most it has bit length `n`
  + the public key `Q = d * G` is a point on the elliptic curve with affine coordinates `(x, y)`, each in the range `{1, ... , q-1}`, thus at most twice bit length `q` (well point encoding actually wastes some bits with padding and optionally saves half the bits if using point compression)

### Curve25519 and Curve448

- Curve25519 and Curve448 are defined in an entirely different manner, and while their functioning is based on elliptic curve arithmetic
- While they have mathematical domain parameters that can be mapped to the traditional EC keys, they are defined as individual cryptosystems so that X25519 is one cryptosystem with exactly one set of parameters, Ed25519 is another distinct cryptosystem with its own set of domain parameters (equivalent to those of X25519, but not exactly the same), and the same goes for X448 and Ed448. This means that every Ed25519 key pair in existence has the same well defined set of parameters, there aren't "flavours" (vs RSA/DSA that can have 1024, 2048, etc. flavours, or DH keys that must be generated under the same agreed domain parameters, or EC keys over this or that curve parameters).
- As a consequence of the last point, the key lenght and the bits of security are defined directly by the _operational_ description of these cryptosystems as fixed constants (even though they can be derived from the mathematical description/proof of the systems) so we get:

https://github.com/openssl/openssl/blob/a0e594d6b5265efd412d248bd12e9d5305370286/crypto/ec/ecx_meth.c#L20-L21

https://github.com/openssl/openssl/blob/a0e594d6b5265efd412d248bd12e9d5305370286/crypto/ec/ecx_meth.c#L25-L27

or fixed constant definitions in IETF standards: https://tools.ietf.org/html/rfc8032#section-8.5

### Post-Quantum Crypto

We don't support any yet, and the only thing I dare say is that again each class of cryptosystems will have its own definition of what is considered the _cryptographic length_ of a key.

## Conclusion

Documenting `EVP_PKEY_bits()` is tricky!

romen (changes requested):

doc/man3/EVP_PKEY_size.pod
```diff
@@ -0,0 +1,39 @@
...
+
+ #include <openssl/evp.h>
+
+ int EVP_PKEY_size(const EVP_PKEY *pkey);
+ int EVP_PKEY_bits(const EVP_PKEY *pkey);
+ int EVP_PKEY_security_bits(const EVP_PKEY *pkey);
+
+=head1 DESCRIPTION
+
+EVP_PKEY_size() returns the size of the key in I<pkey> in bytes.
```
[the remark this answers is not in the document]
levitte: Errrrrrr...  The name `EVP_PKEY_size()` suggests that it's the size of a key.  Now, it's a well known asymmetric property that the signature will never be bigger than the size of the key, but quite frankly, the previous documentation was plain *odd*, considering that getting the size of the key is not reserved for signing only.

However, if you do read the documentation for `EVP_SignFinal()`, it does explain the signature / `EVP_PKEY_size()` relationship well enough, so that documentation isn't lost either way.

So what you seem to see as an incorrect rewrite, I consider a but fix.  Who wins?  :wink:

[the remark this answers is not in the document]
slontis: I think what you have is fine.

doc/man3/EVP_PKEY_size.pod
levitte: Will reword

doc/man3/EVP_PKEY_size.pod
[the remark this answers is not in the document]
levitte: Maybe this could be reworded in a more suitable manner?

doc/man3/EVP_PKEY_size.pod
levitte: Done

doc/man3/EVP_PKEY_size.pod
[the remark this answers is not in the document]
levitte: I reworded it so that the documentation you suggested for `EVP_PKEY_bits()` also applies to `EVP_PKEY_size()`.  Although...  Meh, I just looked at DSA  :confused: 

Too late at night for this, I'll get back tomorrow.

doc/man3/EVP_PKEY_size.pod
```diff
@@ -0,0 +1,40 @@
...
+
+ #include <openssl/evp.h>
+
+ int EVP_PKEY_size(const EVP_PKEY *pkey);
+ int EVP_PKEY_bits(const EVP_PKEY *pkey);
+ int EVP_PKEY_security_bits(const EVP_PKEY *pkey);
+
+=head1 DESCRIPTION
+
+EVP_PKEY_size() and EVP_PKEY_bits() returns the cryptographic length
```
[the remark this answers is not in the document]
romen: I believe that the saner thing to do is to deprecate EVP_PKEY_size as it is horribly named and the cause of this mess.

If, as the current internal implementations suggest, the return value of this function is to be used to preallocate buffers for the result of some specific operation on the key, it is best asked to the operation in question to avoid ambiguity on what length to return when a key supports more than just signatures for example.

Currently anyway we can already call the low level sign or derive operations with a NULL buffer pointer to retrieve the required buffer length to safely store the result of the operation.

romen (changes requested):

doc/man3/EVP_PKEY_size.pod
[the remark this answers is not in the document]
romen: I can see the need for "give me an estimate of a signature size" or "give me an estimate of a shared secret derived by key exchange" from a given key without having to setup all the necessary context.

But again it is a signature over key or kexchange over key property, not a key property, so I would favor deprecation of EVP_PKEY_size and introduction of EVP_PKEY_{signature, exchange, encrypt, ... }_output_size.

We can then document that for legacy compatibility EVP_PKEY_size returns the first non-zero result of the operation specific methods (in a given order: signature, exchange, ...) or 0 if the key does not support any operation (useful key!)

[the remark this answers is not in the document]
levitte: So, er, just woke up and reworded the description again.  Hold on.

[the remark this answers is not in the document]
romen: (for people reading the comments through email, I apologize but I had to edit a few times the comments above, they are all minor edits but for the last comment, where I also added more details and reworded more significantly)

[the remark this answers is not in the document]
levitte: The name is, as you said, not the best choice.  However, if we deprecate it, we need to invent a new name, and for the life of me, I don't have enough imagination for the moment.

I suggest that we post-pone any renaming for another PR.  It might be that we get the manual to a point where we're both happy for now :smile:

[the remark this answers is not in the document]
romen: > The name is, as you said, not the best choice. However, if we deprecate it, we need to invent a new name, and for the life of me, I don't have enough imagination for the moment.

What about [this proposal from the edited comment above](https://github.com/openssl/openssl/pull/10778#discussion_r364626127)?



> But again it is a signature-over-key or kexchange-over-key property, not a key property, so I would favor deprecation of `EVP_PKEY_size` and introduction of `EVP_PKEY_{signature, kexchange, encrypt, ... }_output_size` (with the caveat that these abstract sizes will likely be just informative estimates and deferring to the actual `EVP_PKEY_<operation>` function documentation for the actual requirements on buffer sizes)
> 
> We can then document that for legacy compatibility `EVP_PKEY_size` returns the first non-zero result of the operation specific methods (in a given order: signature, exchange, ...) or 0 if the key does not support any operation (useful key!)



> I suggest that we post-pone any renaming for another PR. It might be that we get the manual to a point where we're both happy for now

It's ok to discuss this as a separate dedicated PR, but I would prefer this PR to be on hold until this issue is solved, as I believe conceptually this should either split into properties belonging to the `OSSL_DISPATCH` of each <operation> or if for some reason I am not seeing right now we want this to stay into the `keymgmt` it should be a

~~~C
OSSL_CORE_MAKE_FUNC(const size_t, OP_keymgmt_query_operation_size, (int operation_id))
~~~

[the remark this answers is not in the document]
levitte: Er...  this PR already has the mechanism for getting the number for `EVP_PKEY_size()`...  that was the primary reason I submitted it!  Documentation was added later.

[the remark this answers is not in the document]
levitte: Regarding new names, and apparently completely new ways to get the size for signatures, public key encrypted secret keys and whatnot, we still need to maintain `EVP_PKEY_size()`, and since that only gets an `EVP_PKEY` as input, the only recourse we have is to go through the KEYMGMT interface.

The suggested new names are tied to contexts for the diverse operations, and would of course mean going through each operation interface, and considering how that will in all likelyhood explode into quite an effort, I'd like to *strongly* suggest another PR where that path can be explored.

providers/implementations/keymgmt/dsa_kmgmt.c
```diff
@@ -190,16 +192,50 @@ static int dsa_exportkey(void *key, OSSL_CALLBACK *param_cb, void *cbarg)
...
+
+    if ((p = OSSL_PARAM_locate(params, OSSL_PKEY_PARAM_BITS)) != NULL
+        && !OSSL_PARAM_set_int(p, DSA_bits(dsa)))
+        return 0;
+    if ((p = OSSL_PARAM_locate(params, OSSL_PKEY_PARAM_SECURITY_BITS)) != NULL
+        && !OSSL_PARAM_set_int(p, DSA_security_bits(dsa)))
+        return 0;
+    if ((p = OSSL_PARAM_locate(params, OSSL_PKEY_PARAM_SIZE)) != NULL
+        && !OSSL_PARAM_set_int(p, DSA_size(dsa)))
+        return 0;
```
romen: I disagree with this bit: `OSSL_PKEY_PARAM_SIZE` is not a key parameter at all, it is a parameter of an operation.

For the `EC` operations for example it is completely left to the whims of the provider developer what to return in this case: he could return the estimated output either of a signature operation or of a kexchange operation.

We should not push the handling of the `EVP_PKEY_size()` legacy ambiguity down the provider API, rather it should be core in charge of figuring out how to solve the ambiguity of the `EVP_PKEY_size()` definition.

I would expect a function in core that:
1. goes through the trouble of going through the operations supported by the `keymgmt` of a `pkey` using `OP_keymgmt_query_operation_name` (or the fallback of looking for signature, kexchange, etc. operations with the same name)
2. once the list of supported operations is available, go through it in the order we deem necessary to satisfy the legacy of `EVP_PKEY_size()` ambiguity

romen (changes requested):

doc/man3/EVP_PKEY_size.pod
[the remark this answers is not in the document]
romen: See https://github.com/openssl/openssl/pull/10778#discussion_r364660278

I don't believe this PR could be considered complete if there is no clear and unambiguous indication of what should be the return for a query for `OSSL_PKEY_PARAM_SIZE`.

A `keymgmt` can be associated with more than one operation (e.g. both signature and kexchange like for EC) and right now we are not defining what to do in that case!

levitte:
Let's move the discussion away from specific diff spots, shall we?

[for those not quite following, the discussion is about `EVP_PKEY_size()` not being quite suitable for what it does, how to implement it for provided keys, and whether it should be divided into operation specific size functions]

@romen said:
> For the `EC` operations for example it is completely left to the whims of the provider developer what to return in this case: he could return the estimated output either of a signature operation or of a kexchange operation.

With legacy keys, this is implemented in the `EVP_PKEY_ASN1_METHOD`, which is as close to the keymgmt interface as you can get.

Something to be noted is that `EVP_PKEY_size` returns a number large enough for *any* operation, i.e. in the case you mention, it would return MAX(sigsize, kexsize).

I'm quite doubtful that having `EVP_PKEY_size()` make an exhaustive search for the *exact* size is useful when the *max* size is good enough for its intended use.  Something to think of is the extra churn that generates.

levitte:
I've wrapped my brain around the idea of operation specific size functions and could give that a go in a separate PR.

romen:
@levitte

What if we reduce the size of this PR by excluding the _bits_ (pun intended) about `EVP_PKEY_size()` and its counterpart into the provider API?

That way we can proceed with the rest of this PR that seems pretty much ready, and debate to our hearts content about what to do for the naming and the ambiguity of `EVP_PKEY_size()`.

Does it sound a plausible approach?

(also I am out of cycles for reviews for today, so it is unlikely I can look at big changes before tomorrow)

romen:
> Something to be noted is that `EVP_PKEY_size` returns a number large enough for _any_ operation, i.e. in the case you mention, it would return MAX(sigsize, kexsize).

@levitte I think that this could become a good interim definition of what `EVP_PKEY_size()` and `OSSL_PKEY_PARAM_SIZE` should return. At least it removes the ambiguity for provider implementers and it also marginally improves the situation for developers linking against `libcrypto` that at least now would have a description of what `EVP_PKEY_size()` returns that matches more closely with the current and legacy behavior.

levitte:
@romen said:
> What if we reduce the size of this PR by excluding the bits (pun intended) about `EVP_PKEY_size()` and its counterpart into the provider API?

The problem is that this is used everywhere to allocate a suitable buffer.  Without this change, anything that uses provider-only keys (such as those created with `EVP_PKEY_fromdata()`, or those generated in #10289) will not be able to do any signing, because `EVP_PKEY_size()` returns zero for such keys.

This is actually a very serious stumbling block (of many) that has kinda halted #10289 for quite some time.

levitte:
@romen, there's one thing I don't quite understand, and it's why a KEYMGMT shouldn't be able to return a known max buffer size for all operations that will use the particular key.  I mean this *technically* rather than on principle.

However, I realise that you might not be entirely aware of one little fact: the KEYMGMT used with any operation *must* belong to the same provider as the operation that uses the keys.  This means that as far as provider authors' "whims" are concerned, it should not be a concern, because that provider will have to be self-consistent to work properly, i.e. what the KEYMGMT part returns *must* be correct for all the operations that the same provider implements.

levitte:
I've added some more changes that eliminate quite a lot of our internal use of `EVP_PKEY_size()` (including a number that went against our documentation!).

I haven't touched libssl at all.  The code in `ssl/statem/` scares me.  @mattcaswell, would you be willing to have a look?  Most specifically, these calls go against documented recommendations (*):

https://github.com/openssl/openssl/blob/1e8879691ff86bda75b6e2e66c0f528684df5d47/ssl/statem/statem_lib.c#L274
https://github.com/openssl/openssl/blob/1e8879691ff86bda75b6e2e66c0f528684df5d47/ssl/statem/statem_srvr.c#L2795

(*) `doc/man3/EVP_DigestSignInit.pod` explicitly says:

> ```
> The use of EVP_PKEY_size() with these functions is discouraged because some
> signature operations may have a signature length which depends on the
> parameters set. As a result EVP_PKEY_size() would have to return a value
> which indicates the maximum possible signature for any set of parameters.
> ```

romen:
With the definition of "is the maximum size of any supported operation on
this key" I don't see a technical problem.

I was against it when the definition seemed to be "it must return the first
meaningful size to conform with the legacy behavior of EVP_PKEY_size: i.e.
return signature size if supported, alternatively encryption size,
alternatively key exchange size, etc."

The first definition is self contained, the second is pushing hacks to
support the ambiguity of EVP_PKEY_size from core to providers!


That being said, I still feel that it is not particularly elegant to
hardcode in  keymgmt.c a number that depends on operations defined in a
different source files (say signature.c and exchange.c).
It means that as a provider implementer I have to **remember** to recheck
the maximum size of **every** possible operation even if I am modifying
only say exchange.c to introduce a new kdf ctrl flag.

It is bad enough that I have to consider every possible combination of
parameters for a single operation, having to recheck it also for the other
files might be problematic and error prone (in addition to being easily
forgotten!).

Anyway, while my first objection felt like something worth holding this PR,
elegance seems secondary and I would be fine with approving this PR even
against my second objection (if we opened an issue to keep track of the
fact that we wish to improve it as soon as possible)

romen:
@levitte Re: https://github.com/openssl/openssl/pull/10778#issuecomment-572670731

With the definition of "is the maximum size of any supported operation on
this key" I don't see a technical problem.

I was against it when the definition seemed to be "it must return the first
meaningful size to conform with the legacy behavior of EVP_PKEY_size: i.e.
return signature size if supported, alternatively encryption size,
alternatively key exchange size, etc."

The first definition is self contained, the second is pushing hacks to
support the ambiguity of EVP_PKEY_size from core to providers!


That being said, I still feel that it is not particularly elegant to
hardcode in  keymgmt.c a number that depends on operations defined in a
different source files (say signature.c and exchange.c).
It means that as a provider implementer I have to **remember** to recheck
the maximum size of **every** possible operation even if I am modifying
only say exchange.c to introduce a new kdf ctrl flag.

It is bad enough that I have to consider every possible combination of
parameters for a single operation, having to recheck it also for the other
files might be problematic and error prone (in addition to being easily
forgotten!).

Anyway, while my first objection felt like something worth holding this PR,
elegance seems secondary and I would be fine with approving this PR even
against my second objection (if we opened an issue to keep track of the
fact that we wish to improve it as soon as possible)

levitte:
> I was against it when the definition seemed to be "it must return the first
meaningful size to conform with the legacy behavior of EVP_PKEY_size: i.e.
return signature size if supported, alternatively encryption size,
alternatively key exchange size, etc."

That's not the legacy behaviour either.  The number returned for one legacy key is constant for that key and is large enough to accommodate a signature done using that key, as well as a secret key encrypted with key, as well as a derived key (DH).  It returns a *maximum* meaningful size for all.  That explains why it returns `ECDSA_size()` for our EC keys, that's simply because it's a larger number than `ECDH_size()` returns, i.e. whoever authored that code already knew the result of `MAX(ECDSA_size(), ECDH_size())`.

Actually, this also strengthens my believe that the description of `EVP_PKEY_size()` that makes you think that it's only about signatures is incorrect.  This has never matched the actual behaviour, or how we use it in our own code.

levitte:
> I've added some more changes that eliminate quite a lot of our internal use of `EVP_PKEY_size()` (including a number that went against our documentation!).

That was a fun experiment, but I'm going to move it to another branch / PR.  For the intents and purposes of *this* PR, that was a bit over the top.

kaduk:
> I've added some more changes that eliminate quite a lot of our internal use of `EVP_PKEY_size()` (including a number that went against our documentation!).
> 
> I haven't touched libssl at all. The code in `ssl/statem/` scares me. @mattcaswell, would you be willing to have a look? Most specifically, these calls go against documented recommendations (*):
> 
> https://github.com/openssl/openssl/blob/1e8879691ff86bda75b6e2e66c0f528684df5d47/ssl/statem/statem_lib.c#L274

This one looks like it's just avoiding having to make different allocation calls for the SSLv3 and other codepaths.

> 
> https://github.com/openssl/openssl/blob/1e8879691ff86bda75b6e2e66c0f528684df5d47/ssl/statem/statem_srvr.c#L2795

This one is a little awkward since it has to reserve space in the WPACKET and get a pointer thereto to sign directly into, but I think there's not any fundamental issue preventing us from moving that reservation later (when we can know the exact length).

levitte:
@kaduk, the `EVP_PKEY_size()` elimination got moved to a PR of its own, #10798.  @mattcaswell fixed all the ssl/statem/ ones there.

levitte:
So @romen, how do you feel about this PR now?  I've tried hard to described it in a way that is compatible with previous documentation as well as with our actual use.

kaduk:
> @kaduk, the `EVP_PKEY_size()` elimination got moved to a PR of its own, #10798. @mattcaswell fixed all the ssl/statem/ ones there.

Whoops, that's what I get for not being fully caught up.  Thanks for the pointer.

levitte:
> Whoops, that's what I get for not being fully caught up.

No problem

doc/man3/EVP_SignInit.pod
```diff
@@ -17,49 +15,36 @@ functions
...
-EVP_SignFinal() signs the data in B<ctx> using the private key B<pkey> and
-places the signature in B<sig>. B<sig> must be at least EVP_PKEY_size(pkey)
-bytes in size. B<s> is an OUT parameter, and not used as an IN parameter.
+EVP_SignFinal() signs the data in I<ctx> using the private key I<pkey> and
+places the signature in I<sig>. I<sig> must be at least C<EVP_PKEY_size(pkey)>
+bytes in size. I<s> is an OUT parameter, and not used as an IN parameter.
 The number of bytes of data written (i.e. the length of the signature)
-will be written to the integer at B<s>, at most EVP_PKEY_size(pkey) bytes
+will be written to the integer at I<s>, at most C<EVP_PKEY_size(pkey)> bytes
 will be written.
```
romen: I would take this occasion to add a note as I have seen quite a few users missing that even for the same `pkey`, identically initialized `ctx` and even the same amount of processed data, the length of the generated signatures can vary, so they really MUST use the length written to the integer at `I<s>`, or they will **occasionally** experience bugs (most likely random-looking SegFaults) if they use the output of `EVP_PKEY_size()` as the length of the just generated signature.

Most of the time I end up seeing code that looks like this:

```C
int s;
int size = EVP_PKEY_size(pkey);
unsigned char *sig = malloc(size);

EVP_SignFinal(ctx, sig, &s, pkey);

fwrite(sig, size, 1, outfile);
```

and the occurrence rate is so high that I believe maybe it is worth being even more explicit than we already are in pointing out that the return value of `EVP_PKEY_size()` is **NOT** the length of the signature.

---

My suggestion for the note would be:

```pod
Note that signatures generated using the same private I<pkey>
and identically configured I<ctx> may vary in length, independently
of the amount of processed data.  When handling the signature placed
in I<sig>, it is crucial to never ignore the associated length, written to
the integer at I<s>, to avoid bugs. 
```

include/crypto/evp.h
```diff
@@ -565,6 +565,11 @@ struct evp_pkey_st {
      * a copy of that key's dirty count.
      */
     size_t dirty_cnt_copy;
+
+    /* Cache of domain parameter / key information */
+    int bits;
+    int security_bits;
+    int size;
```
romen: This is a sidenote, not a request for changing anything, but to discuss if you believe there could be value in an alternative.

@levitte what do you think about including these in a `struct`?

```C
struct {
    int bits;
    int security_bits;
    int size;
} cache;
```

I believe it could be a good idea to embed the fact that they are under the control of  `evp_keymgmt_cache_pkey()` and help future contributors be aware of it.

romen:
I am fine with the current revision of `EVP_PKEY_size.pod`: now that is more or less stabilized we could include `pod` changes for the provider documentation about the new params and related functions.

crypto/evp/keymgmt_lib.c
```diff
@@ -161,6 +157,49 @@ void evp_keymgmt_clear_pkey_cache(EVP_PKEY *pk)
...
+            int bits = 0;
+            int security_bits = 0;
+            int size = 0;
+            OSSL_PARAM params[4];
+
+            params[0] = OSSL_PARAM_construct_int(OSSL_PKEY_PARAM_BITS, &bits);
+            params[1] = OSSL_PARAM_construct_int(OSSL_PKEY_PARAM_SECURITY_BITS,
+                                                 &security_bits);
+            params[2] = OSSL_PARAM_construct_int(OSSL_PKEY_PARAM_SIZE, &size);
+            params[3] = OSSL_PARAM_construct_end();
```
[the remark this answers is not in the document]
levitte: C90 doesn't allow the construct that you suggest (and we do have the corresponding macros too) for automatic (stack allocated) variables. That possibility came with C99, if memory serves...

doc/man3/EVP_SignInit.pod
levitte: Good point

include/crypto/evp.h
levitte: Sure

doc/man3/EVP_PKEY_size.pod
levitte: @romen took so much space that I totally glossed over this little comment. Thanks for the ping.
Yeah, copyright years will be changed.

providers/implementations/keymgmt/rsa_kmgmt.c
levitte: You're thinking of PSS, aren't you? Otherwise, rsa doesn't come with domain parameters as far as I know.

Either way, I'm finally about to put together PSS support across the diverse implementations where it's missing.

doc/man3/EVP_PKEY_size.pod
romen: I am the worst! Sorry @slontis 😂

crypto/evp/keymgmt_lib.c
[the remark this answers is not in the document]
romen: Ok, then just forget this comment, thanks for the feedback!

doc/man3/EVP_PKEY_size.pod
levitte: Fixed

doc/man3/EVP_PKEY_size.pod
```diff
@@ -0,0 +1,73 @@
...
+
+=head1 NOTES
+
+Most functions that have an output buffer and are mentioned with
+EVP_PKEY_size() have a functionality where you can pass NULL for the
+buffer and still pass a pointer to an integer and get the exact size
+that this function call delivers in the context that it's called in.
+This allows those functions to be called twice, once to find out the
+exact buffer size, then allocate the buffer in between, and call that
+function again actually output the data.  For those functions, is
```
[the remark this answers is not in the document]
levitte: Fixed

doc/man3/EVP_SignInit.pod
levitte: I reworded it a bit, but believe I got the spirit of your text.

include/crypto/evp.h
levitte: Done

doc/man3/EVP_SignInit.pod
romen: I prefer how you wrote it! Thanks!

doc/man3/EVP_PKEY_size.pod
richsalz: Futzing with copyright is also done when a release is made so I wouldn't lose sleep over it.

levitte: Except this is a *new* file, so it should only say 2020. Our scripts don't futz with the starting year

levitte: Either way, it's fixed already, so...

romen:
We should also include a CHANGES entry about the fixed/updated documentation

levitte:
> We should also include a CHANGES entry about the fixed/updated documentation

Done

levitte:
... and finally, provider documentation added.

doc/man7/provider-keymgmt.pod
```diff
@@ -171,6 +195,40 @@ returns NULL, the caller is free to assume that there's an algorithm
...
+
+=item "size" (B<OSSL_PKEY_PARAM_SIZE>) <integer>
+
+The value should be the maximum size that a caller should allocate to
+safely store a signature (called I<sig> in L<provider-signature(7)>),
+the result of asymmmetric encryption / decryption (I<out> in
+L<provider-asym_cipher(7)>, a derived secret (I<secret> in
+L<provider-keyexch(7)> and similar data.
+
+Because an EVP_KEYMGMT method is always tightly to another method
```
romen: Typo: missing word. Maybe bound, associated or connected?

CHANGES
```diff
@@ -9,6 +9,12 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) Enhanced the documentation of EVP_PKEY_size(), EVP_PKEY_bits()
+     and EVP_PKEY_security_bits().  Especially EVP_PKEY_size() needed
+     a new formulation to include all the things it can be used for,
+     as well as words of caution.
+     [Richard Levitte]
+
```
romen: Really nice text for the entry! Thanks!

doc/man7/provider-keymgmt.pod
```diff
@@ -171,6 +195,40 @@ returns NULL, the caller is free to assume that there's an algorithm
...
+(signature, asymmetric cipher, key exchange, ...) and must be of the
+same provider, this number only need to be synchronised with the
```
romen: Typo: needs

romen (changes requested):
@levitte Thanks for bearing with me so far!!

I believe these could be the last nits!

doc/man7/provider-keymgmt.pod
levitte: Fixed

doc/man7/provider-keymgmt.pod
levitte: Fixed

levitte:
All fixups squashed, and rebased.  I hope the CIs will be happy with it.

slontis:
> All fixups squashed, and rebased. I hope the CIs will be happy with it.

Nup :)

levitte:
Fixed

mspncp:
> Let's move the discussion away from specific diff spots, shall we?
> 
> [for those not quite following, the discussion is about `EVP_PKEY_size()` not being quite suitable for what it does, how to implement it for provided keys, and whether it should be divided into operation specific size functions]

Indeed, I'm a bit lost and unsure where the discussion about `EVP_PKEY_size()` ended up. GitHub's web interface makes it very hard to reconstruct the discussion thread behind all those hidden resolved comments.

@romen have all your concerns about `EVP_PKEY_size()` now been resolved? If you still think that the function is misnamed and has problems, it would be good if you could take the time to open a separate issue which summarizes your concerns and the current status-quo.

romen:
Yes all my concerns about EVP_PKEY_size() have been addressed so far, we are just trying to please the CIs

levitte:
Current travis failure is not relevant, it's just a worker that's slow (it seems to be one and the same, I've seen that job fail with early timeouts for a while now)

include/openssl/core_names.h
```diff
@@ -154,6 +154,11 @@ extern "C" {
 #define OSSL_KDF_NAME_KRB5KDF       "KRB5KDF"
 
 /* PKEY parameters */
+/* Common PKEY parameters */
+#define OSSL_PKEY_PARAM_BITS                "bits" /* integer */
+#define OSSL_PKEY_PARAM_SIZE                "size" /* integer */
```
romen: I would leave this as a thought, that could be discussed in a separate PR.

In the interface with the provider we are not forced to use the same name of the EVP layer interface, especially in cases where we agree the original naming is quite unfortunate and misleading. 

What about renaming this as OSSL_PKEY_PARAM_OUT_MAX_SIZE or something similar to overcome the poor naming of the EVP API?

providers/implementations/keymgmt/dh_kmgmt.c
```diff
@@ -185,16 +187,54 @@ static int dh_exportkey(void *key, OSSL_CALLBACK *param_cb, void *cbarg)
     return ret;
 }
 
+/*
+ * Same function for domain parameters and for keys.
+ * "dpk" = "domain parameters & keys"
+ */
+static int dh_get_dpk_params(void *key, OSSL_PARAM params[])
```
romen: Nit: should we add `ossl_inline`? (and do the same for the other keymgmt)

romen (approved):
I have a few very minor nits/side notes.

My approval stands even in case the resolution of these minor things is won't fix, as I don't really have a strong opinion on them and are just very minor things that popped out during my last review!

P. S. I agree that the Travis failure seems unrelated.

include/openssl/core_names.h
levitte: I would say that `OSSL_PKEY_PARAM_MAX_SIZE` is good enough, with "max-size" as the string

doc/man7/provider-keymgmt.pod
```diff
@@ -171,6 +195,42 @@ returns NULL, the caller is free to assume that there's an algorithm
...
+which the key belongs, in bits.  The definition of cryptographic
+length is specific to the key cryptosystem.
+
+=item "size" (B<OSSL_PKEY_PARAM_SIZE>) <integer>
+
+The value should be the maximum size that a caller should allocate to
+safely store a signature (called I<sig> in L<provider-signature(7)>),
+the result of asymmmetric encryption / decryption (I<out> in
+L<provider-asym_cipher(7)>, a derived secret (I<secret> in
+L<provider-keyexch(7)> and similar data.
```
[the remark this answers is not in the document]
levitte: Done

providers/implementations/keymgmt/dh_kmgmt.c
levitte: Sure

levitte:
The extra commits that came just now solve the nits @romen had for me.  From where I stand, the approval still stands.

romen:
Yes, I confirm that my approval was implicitly granted!

levitte:
I actually found it rather explicit :wink:

levitte:
Merged.

6508e858836020622efff5dd4be3fa4530d1584b EVP: make EVP_PKEY_{bits,security_bits,size} work with provider only keys
9e5aaf78868ca2f4d65d3bb1f9246659a6522241 PROV: Adapt the RSA, DSA and DH KEYMGMT implementations
81a624f2c215eb9def0917184b026b2e2db2dd54 TEST: Adapt test/evp_pkey_provided_test.c to check the key size
6942a0d6feb8d3dcbbc6a1ec6be9de7ab2df1530 DOC: New file for EVP_PKEY_size(), EVP_PKEY_bits() and EVP_PKEY_security_bits()
03d65ca2095777cf6314ad813eb7de5779c9b93d DOC: Make EVP_SignInit.pod conform with man-pages(7)
f17268d0d05add0240dcfdbb91d522aa7b10c790 Add CHANGES entry regarding the documentation of EVP_PKEY_size() et al
806253f31f663594c70e10862f0743530a75e094 DSA: Move DSA_security_bits() and DSA_bits()

PR #10783, opened 2020-01-08
PROV: Fix mixup between general and specialized GCM implementations

levitte:
providers/implementations/ciphers/ciphercommon_gcm_hw.c had an AES
specific GCM update function, while
providers/implementations/ciphers/cipher_aria_gcm_hw.c had the more
general implementation.

This moves them around to have the more general implementation in the
common source, and place the AES specialiation where it belongs.

### DIFF
--- a/providers/implementations/ciphers/cipher_aes_gcm_hw.c
+++ b/providers/implementations/ciphers/cipher_aes_gcm_hw.c
@@ -17,7 +17,7 @@
 
 #include "cipher_aes_gcm.h"
 
-static int generic_aes_gcm_initkey(PROV_GCM_CTX *ctx, const unsigned char *key,
+static int aes_gcm_initkey(PROV_GCM_CTX *ctx, const unsigned char *key,
                                    size_t keylen)
 {
     PROV_AES_GCM_CTX *actx = (PROV_AES_GCM_CTX *)ctx;
@@ -125,7 +125,7 @@ static int generic_aes_gcm_cipher_update(PROV_GCM_CTX *ctx, const unsigned char 
 }
 
 static const PROV_GCM_HW aes_gcm = {
-    generic_aes_gcm_initkey,
+    aes_gcm_initkey,
     gcm_setiv,
     gcm_aad_update,
     generic_aes_gmc_cipher_update,
--- a/providers/implementations/ciphers/cipher_aes_gcm_hw.c
+++ b/providers/implementations/ciphers/cipher_aes_gcm_hw.c
@@ -59,8 +59,8 @@ static int generic_aes_gcm_initkey(PROV_GCM_CTX *ctx, const unsigned char *key,
     return 1;
 }
 
-static int generic_aes_cipher_update(PROV_GCM_CTX *ctx, const unsigned char *in,
-                                     size_t len, unsigned char *out)
+static int generic_aes_gcm_cipher_update(PROV_GCM_CTX *ctx, const unsigned char *in,
+                                         size_t len, unsigned char *out)
 {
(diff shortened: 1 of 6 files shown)


providers/implementations/ciphers/cipher_aes_gcm_hw.c
```diff
@@ -59,11 +59,76 @@ static int generic_aes_gcm_initkey(PROV_GCM_CTX *ctx, const unsigned char *key,
     return 1;
 }
 
+static int generic_aes_cipher_update(PROV_GCM_CTX *ctx, const unsigned char *in,
```
t8m: Should it be called generic_aes_gcm_cipher_update? And is the generic_ prefix necessary?

slontis: aes_gcm_cipher_update seems like a reasonable name to me..

levitte: That seems inconsistent, considering `generic_aes_gcm_initkey`.  Should that one be renamed as well?

slontis: It is only used by this file - so maybe generic was a bad name.
All the hw platforms have there own name e.g- s390x_aes_gcm_initkey
This is the default name if there is no hw or if the hw is not capable..

t8m:
This replaces #10685

levitte:
> This replaces #10685

Er...  does it?

t8m:
Yes, AFAIK it does because the code depending on AES_GCM_ASM being defined is moved to a file that already indirectly through cipher_aes_gcm.h includes the aes_platform.h.

levitte:
Aha!  I didn't realise that...

t8m (dismissed):
Ping @levitte will you rename the generic_aes functions to drop the generic prefix? If not I give approval anyway although I'd prefer them to be renamed.

levitte:
I've renamed `generic_aes_gcm_initkey` to `aes_gcm_initkey`, since it's only used locally in `providers/implementations/ciphers/cipher_aes_gcm_hw.c`.  `generic_aes_gmc_cipher_update` is a different story, as it *is* generic for most of the different hardware implementations, so I left that alone.

If someone wants to clean away `generic` from all the names, it should be done everywhere, not just in one spot.  That subject for another PR (and someone other than me :wink:)

providers/implementations/ciphers/cipher_aes_gcm_hw.c
```diff
@@ -59,11 +59,76 @@ static int generic_aes_gcm_initkey(PROV_GCM_CTX *ctx, const unsigned char *key,
...
+    return 1;
+}
+
 static const PROV_GCM_HW aes_gcm = {
-    generic_aes_gcm_initkey,
+    aes_gcm_initkey,
     gcm_setiv,
     gcm_aad_update,
-    gcm_cipher_update,
+    generic_aes_gmc_cipher_update,
```
t8m: Typo gmc.

t8m (changes requested):

providers/implementations/ciphers/cipher_aes_gcm_hw.c
levitte: Fixed

levitte:
Ping?

slontis (approved):

t8m (approved):

levitte:
Merged.

7c66ad65f959fa05ad7b95b0c84384d6e63e56f1 PROV: Fix mixup between general and specialized GCM implementations

PR #10784, opened 2020-01-08
Check that the default signature type is allowed

kroeckx:
TLS < 1.2 has fixed signature algorithms: MD5+SHA1 for RSA and SHA1 for the
others. TLS 1.2 sends a list of supported ciphers, but allows not sending
it in which case SHA1 is used. TLS 1.3 makes sending the list mandatory.

When we didn't receive a list from the client, we always used the
defaults without checking that they are allowed by the configuration.

##### Checklist


kroeckx:
The tests fails because apps/server.pem is using sha1, so running the test at security level 2 fails. Replacing apps/server.pem with a sha256 based certificate will make the dtls test fail instead.

kroeckx:
The failure can be seen in #10786, which does replace the certificate.

test/recipes/70-test_sslsigalgs.t
```diff
@@ -132,35 +132,56 @@ SKIP: {
...
 
-    #Test 10: Sending no sig algs extension in TLSv1.2 should succeed
+    #Test 10: Sending no sig algs extension in TLSv1.2 should succeed at
+    #         security level 1
     $proxy->clear();
     $testtype = NO_SIG_ALGS_EXT;
-    $proxy->clientflags("-no_tls1_3");
-    $proxy->ciphers("ECDHE-RSA-AES128-SHA");
+    $proxy->clientflags("-no_tls1_3 -cipher DEFAULT\@SECLEVEL=1");
+    $proxy->ciphers("ECDHE-RSA-AES128-SHA\@SECLEVEL=1");
```
mattcaswell: Isn't this the default SECLEVEL anyway - so why does it need to be explicitly stated?

richsalz: Do you want the tests to pass if someone builds with a different seclevel as the default?

kroeckx: #10786 will need to change all the values to be 1 lower. It's also explicit at which level we expect to succeed and which to fail, we might at some point want to change the default security level.

kroeckx:
merged

PR #10785, opened 2020-01-08
Check the that ed25519 and ed448 are allowed by the security level

kroeckx:
Signature algorithms not using an MD weren't checked that they're
allowed by the security level.

##### Checklist


kroeckx:
Still need to update the tests

ssl/t1_lib.c
```diff
@@ -1062,6 +1062,30 @@ int tls_check_sigalg_curve(const SSL *s, int curve)
...
+        return 0;
+    if (md != NULL)
+    {
+        /* Security bits: half digest bits */
+        secbits = EVP_MD_size(md) * 4;
+    } else {
+        if (lu->sigalg == TLSEXT_SIGALG_ed25519)
+            secbits = 128;
+        else if (lu->sigalg == TLSEXT_SIGALG_ed448)
+            secbits = 224;
```
mattcaswell: Is there a reference for these values?

kaduk: https://tools.ietf.org/html/rfc8032#section-8.5

kroeckx: Yes, see ecx_meth.c line 21 and 27. I guess I could move the define to some public header.

mattcaswell: Thanks. Might be worth including the reference @kaduk gave in a comment here (optionally also in ecx_meth.c).

romen: I agree adding the reference on both files is a good addition!

kroeckx:
I've updated the tests and fixed a conflict.

mspncp:
Drive by comment: Typo in commit message: „Check ~the~ that“

mattcaswell:
LGTM, except that travis red cross seems relevant.

kroeckx:
For some reason I commited a test change to the .in file

PR #10786, opened 2020-01-08
Stop accepting certificates signed using SHA1 at security level 1

kroeckx:
Replaces apps/server.pem with certs/test/servercert.pem

##### Checklist


kroeckx:
It's failing because test/certs/embeddedSCTs1.pem is signed using sha1. I'm not sure how that file is generated.

crypto/x509/x509_set.c
```diff
@@ -222,6 +222,18 @@ static void x509_sig_info_init(X509_SIG_INFO *siginf, const X509_ALGOR *alg,
         return;
     /* Security bits: half number of bits in digest */
     siginf->secbits = EVP_MD_size(md) * 4;
+    /* SHA1 and MD5 are known broken. Reduce security bits so that they're no
```
richsalz: nit "/*" on a line by itself.  I wouldn't say SHA1 and MD5 are broken to the same level, but I don't get into that argument :)

kroeckx:
@mattcaswell: Can you look into why the dtls test fails? The change to apps/server.pem is causing it.

crypto/x509/x509_set.c
```diff
@@ -222,6 +222,18 @@ static void x509_sig_info_init(X509_SIG_INFO *siginf, const X509_ALGOR *alg,
...
+     * longer accepted at security level 1. The real values don't really matter
+     * as long as they're lower than 80, which is our security level 1.
+     * https://eprint.iacr.org/2020/014 puts a chosen-prefix attack for SHA1 at
+     * 2^63.4
+     * https://documents.epfl.ch/users/l/le/lenstra/public/papers/lat.pdf
+     * puts puts a chosen-prefix attack for MD5 at 2^39.
```
FdaSilvaYY: nit : `puts ` is repeated ?

crypto/x509/x509_set.c
mspncp: are known _to be_ broken

mattcaswell:
> Can you look into why the dtls test fails?

I will investigate later today and report back

mattcaswell:
>> Can you look into why the dtls test fails?
>
> I will investigate later today and report back

Ok, I've taken a look. The explanation is a little complicated but the fix is straight forward.

TL;DR the fix is this:

```` diff
diff --git a/test/dtlstest.c b/test/dtlstest.c
index 2484862860..022b4c0149 100644
--- a/test/dtlstest.c
+++ b/test/dtlstest.c
@@ -122,7 +122,7 @@ static int test_dtls_unprocessed(int testidx)
 #define CLI_TO_SRV_EPOCH_0_RECS 3
 #define CLI_TO_SRV_EPOCH_1_RECS 1
 #if !defined(OPENSSL_NO_EC) || !defined(OPENSSL_NO_DH)
-# define SRV_TO_CLI_EPOCH_0_RECS 12
+# define SRV_TO_CLI_EPOCH_0_RECS 10
 #else
 /*
  * In this case we have no ServerKeyExchange message, because we don't have
````

The test that was failing was testing connections between a dtls client and server. In particular it was checking that if we drop a record that the handshake recovers and still completes successfully. The test iterates a number of times. The first time through it drops the first record. The second time it drops the second one, and so on. In order to do this it has a hard-coded value for the expected number of records it should see in a handshake. That's ok because we completely control both sides of the handshake and know what records we expect to see. Small changes in message size would be tolerated because that is unlikely to have an impact on the number of records. Larger changes in message size however could increase or decrease the number of records and hence cause the test to fail.

This particular test uses a mem bio which doesn't have all the CTRLs that the dgram BIO has. When we are using a dgram BIO we query that BIO to determine the MTU size. The smaller the MTU the more fragmented handshakes become. Since the mem BIO doesn't report an MTU we use a rather small default value and get quite a lot of records in our handshake. This has the tendency to increase the likelihood of the number of records changing in the test if the message size changes.

It so happens that the new server certificate is *smaller* than the old one. AFAICT this is probably because the DNs for the Subject and Issuer are significantly shorter than previously. The result is that the number of records used to transmit the Certificate message is one less than it was before. This actually has a knock on impact for subsequent messages and how we fragment them resulting in one less ServerKeyExchange record too (the actual size of the ServerKeyExchange message hasn't changed, but where in that message it gets fragmented has). In total the number of records used in the handshake has decreased by 2 with the new server.pem file.

**Comment by kroeckx** [comment]
So I tried to to create a new embeddedSCTs1.pem, but I currently can't. There doesn't seem to be a key for the embeddedSCTs1_issuer.pem file, so I can't use that file to sign a new certificate. So we might need to replace that file too. If we need to replace the issuer, I suggest we replace it with ca-cert.pem/ca-cert.key file.

But that still leaves how to generate an SCT. I can add the extension using the x509 -extfile parameter, but I guess coping it from some other cert is going to cause an error validating it. We do not seem to have API to generate an SCT, just one to validate one.

The CT tests are as far as I know the only tests that currently fail.

@RJPercival: Can you help with this?

**Comment by RJPercival** [comment]
I believe this is the issuer key: https://github.com/google/certificate-transparency/blob/master/test/testdata/ca-key.pem

However, the simplest solution is probably to just grab any certificate from a CT log that has embedded SCTs (and doesn't use SHA1), along with the issuer cert.

**Comment by RJPercival** [comment]
Oh actually, you need the private key for the cert as well, don't you? I'll see if I can find any tools that help with generating an SCT - we have code for marshalling an SCT structure into TLS at https://github.com/google/certificate-transparency-go/blob/master/types.go, which is probably the most complicated part.

**Comment by RJPercival** [comment]
I've thrown together a tool that'll give you an SCT in TLS format, which is *almost* suitable for inserting into a certificate (it needs encoding as a SignedCertificateTimestampList, which I think just involves prefixing it with a few bytes): https://github.com/google/certificate-transparency-go/pull/640

**Comment by kroeckx** [comment]
Using the key you pointed to, I was able to create a new certificate, but I still have 2 errors:
```
            # ERROR: (int) 'result->result == test_ctx->expected_result' failed @ test/ssl_test.c:33
            # [2] compared to [0]
            # INFO:  @ test/ssl_test.c:34
            # ExpectedResult mismatch: expected Success, got ClientFail.
            # 00:B2:15:0F:92:7F:00:00:error:SSL routines::sslv3 alert handshake failure:ssl/record/rec_layer_s3.c:1596:SSL alert number 40
            not ok 4 - iteration 4
            ok 5 - iteration 5
            # ERROR: (int) 'result->result == test_ctx->expected_result' failed @ test/ssl_test.c:33
            # [4] compared to [0]
            # INFO:  @ test/ssl_test.c:34
            # ExpectedResult mismatch: expected Success, got FirstHandshakeFailed.
            # 00:B2:15:0F:92:7F:00:00:error:SSL routines::sslv3 alert handshake failure:ssl/record/rec_layer_s3.c:1596:SSL alert number 40
```

Not sure yet why it's failing.

**Comment by kroeckx** [comment]
I'm guessing there is something wrong with the SCT somehow, it should be a copy from the old file.

**Comment by RJPercival** [comment]
The SCT contains a signature over the certificate, so you can't hash algorithm without invalidating it. You'll need to use the linked tool to create a new SCT.

**Comment by kroeckx** [comment]
I think that the ct_test.c is then just going to fail. I think I'm just going to run that test at SECLEVEL=0 and open an issue for it instead.

**Comment by kroeckx** [comment]
So trying to set @SECLEVEL=0 in the config file for the server fails, because the config files are executed in the order as they are in the config, and the lines seem to be sorted alphabetically. But if you want to use a SHA1 based certificate, you need to set the cipherstring to change the security level before you try to load the certificate.

@levitte: Can you change generate_ssl_tests.pl so that it uses the same order as in the the .conf.in files?

**Comment by RJPercival** [comment]
@kroeckx: Why do you "think that the ct_test.c is then just going to fail"? If you use the linked tool to create an SCT, rather than copy-and-pasting the old SCT, then it should pass.

File: test/certs/mkcert.sh [hunk 3]
```diff
@@ -204,12 +204,14 @@ geneealt() {
 genee() {
     local OPTIND=1
     local purpose=serverAuth
+    local extention
```
RJPercival: s/extention/extension/

kroeckx:
I assume that it checks certain values from the file, but I might
also just not understand what the test is doing exactly.

Anyway, I need help, I don't have much time. At the current rate
it's going to take months before any of those related PRs are
getting merged.

kroeckx:
So I tried to create a new certificate, but I guess I'm still doing something wrong. Maybe the first certificate I make should be a precertificate?

RJPercival:
That's correct - you'll need to give a pre-cert to the tool, which will then give you an SCT suitable for embedding in the final cert.

kroeckx:
I just need to figure out the right syntax to generate such an extention.

kroeckx:
The error I'm now getting is:
```
    # ERROR: (int) 'valid_sct_count == fixture->expected_valid_sct_count' failed @ test/ct_test.c:211
    # [0] compared to [1]
    # INFO:  @ test/ct_test.c:215
    # 1 SCTs failed, 0 SCTs unverified
    not ok 4 - test_verify_one_sct
```

test/certs/mkcert.sh
```diff
@@ -288,4 +288,56 @@ gennocn() {
...
+    csr=$(req "$key" "CN = $cn") || return 1
+    echo "$csr" |
+	cert "$cert" "$exts" -CA "${ca}.pem" -CAkey "${cakey}.pem" \
+	    -set_serial 2 -days "${DAYS}" "$@"
+    cat ${cert}.pem ${ca}.pem > ${cert}-chain.pem
+    go run github.com/google/certificate-transparency-go/ctutil/sctgen \
+       --log_private_key ${logkey}.pem \
+       --timestamp="2020-01-01T00:00:00Z" \
+       --cert_chain ${cert}-chain.pem \
+       --tls_out ${cert}.tlssct
```
vdukhovni: This introduces a requirement on a live Internet connection, and running untrusted code from github.com into mkcert.sh. (And of course having "go" installed). While we tend to not run it during builds, and instead just cache the artefacts, I am not fond of this change. Perhaps just make it in a private branch, and commit just the output? Otherwise, we likely need actual OpenSSL code  that can generate this sort of object...

vdukhovni:
The actual change to the code again looks fine. Sadly, once again the bulk of the work is in fixing the tests, and this time in the details of CT generation and validation... :-(  Good luck with that!

test/certs/mkcert.sh
kroeckx: As far as I understand, you don't actually need to have a live internet connection, you always need to have it checked out locally.

I do wish that we didn't need this external code to generate such a certificate, and it's probably not that hard to write something that does the same, but we currently don't have it.

Anyway, the reason to put it in that file is so that we at least have a documented way of creating such files.

richsalz: I agree with Victor.  Adding "go" as a dependency to build openssl seems bad.  Keep the lines commented output and just checkin the output.

vdukhovni: @richsalz Just to be clear, mkcert.sh is not normally run during builds. It is used to manufacture certificates for the test suite, which are then checked into git. So it is not actually needed for the vast majority of users...

levitte: Er, we don't run this script when building, do we? This is a script to generate certs, which are checked in, isn't it? So kinda like "make update", but run even more seldom...

richsalz: thanks for the correction.  Is the project team okay with adding a go dependency?

vdukhovni: I am not fond of the `go` dependency, and would prefer to see it "go" away. But for that we'd need our own CT generation code (which probably should be added at some point). In the mean time this is only needed to generate test certs with SCTs. The rest of the cert generation bypasses this code.
So the real question is whether the `go` call should committed in tree, or just used privately, until we have our own code to do this.
My instinct was to not add the `go` call, but it is a very minor issue. This could e.g. just be commented out, as Kurt said, it is there primarily so that someone else (or he later) could retrace his steps.

kroeckx:
It seems that all that was missing was that I needed to change the time at which we check the SCT.

kroeckx:
Does someone want to review this?

CHANGES
```diff
@@ -9,6 +9,17 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) X509 certificates signed using SHA1 are no longer allowed at security
+     level 1, the highest security level it's accepted at is 0.
```
vdukhovni: Just say "at security level 1 and above", and delete the rest of the first sentence.

CHANGES
```diff
@@ -9,6 +9,17 @@
...
+     When used in TLS/SSL the default security level is 1. It can be set either
```
vdukhovni: s/When used in/In/.

CHANGES
```diff
@@ -9,6 +9,17 @@
...
+     using the cipher string with @SECLEVEL, or calling
+     SSL_CTX_set_security_level(). A call to SSL_CTX_use_certificate() will
+     fail if the security level is not lowered first.
```
vdukhovni: If the leaf certificate is signed with SHA-1.

CHANGES
```diff
@@ -9,6 +9,17 @@
...
+     Outside TLS/SSL, the default security level is -1. It can be set using
```
vdukhovni: Add (effectively 0). (The only difference relates to how verify parameter inheritance works, -1 means not set, while 0 is explicitly 0, so when inheritance only picks up new settings 0 stays unchanged, but -1 picks up new values from the default object.  Most users don't need to know or care, so you could just say "0").

NEWS
```diff
@@ -7,6 +7,8 @@
 
   Major changes between OpenSSL 1.1.1 and OpenSSL 3.0.0 [under development]
 
+      o X509 certificates signed using SHA1 are no longer allowed at security
+        level 1
```
vdukhovni: 1 or higher.

vdukhovni: Perhaps again mention that the default security level is 1 for TLS, which means that certificates signed with SHA-1 are by default no longer trusted to authenticate TLS servers or clients.

NEWS
```diff
@@ -8,7 +8,9 @@
   Major changes between OpenSSL 1.1.1 and OpenSSL 3.0.0 [under development]
 
       o X509 certificates signed using SHA1 are no longer allowed at security
-        level 1
+        level 1 or higher. The default security level for TLS is 1, so
+        certificates signed using SHA1 are by default no longer trusted to
+        authenticate servers.
```
vdukhovni: or clients using client certs (i.e. not just servers)

vdukhovni:
But it looks like you now need to rebase to pick up the changes to the "CHANGES" file.

mattcaswell:
Sorry - late to the party. I see this got backported to 1.1.1. Presumably that is going to break stuff. Are we ok with that? Does that need some kind of OMC decision?

vdukhovni:
I had only intended to approve it for master. Did not notice it was also intended for 1.1.1.  That's a mistake I think.  Has it been merged into 1.1.1???

vdukhovni:
We may need to revert this, unless there's a consensus to make the change also in 1.1.1.

mattcaswell:
> I had only intended to approve it for master. Did not notice it was also intended for 1.1.1. That's a mistake I think. Has it been merged into 1.1.1???

Yes, in 68436f0a8964e

vdukhovni:
Presumably a revert will need a new PR and approvals subject to consensus to do that...

mattcaswell:
> Presumably a revert will need a new PR and approvals subject to consensus to do that...

Yes

If we don't revert then IMO such a significant breaking change to 1.1.1 should at least require an OMC vote.

t-j-h:
I don't think this is something that should go into 1.1.1 without a  specific vote to do so.

I also note that @kroeckx did very clearly label his intention here from the beginning - viewing this as a security fix and treating it as such to apply in all currently supported releases.

So we may end up there - but I don't think the simple reviewer approval is the right level for this sort of impactful change to be approved - but no one (myself included) placed a hold to trigger a vote.

vdukhovni:
So, proceed as you see best.  As I said, I did not mean to approve 1.1.1, was looking just at the code, not the labels. Sorry about that.  Someone should open a revert PR and a vote.

t-j-h:
@mattcaswell I suggest just kicking off an OMC vote on the topic - the PR may or may not be needed - and items like this are voted on quickly.

mattcaswell:
Proposed vote text:

"We should backport the change to stop accepting certificates signed using SHA1 at security level 1 to 1.1.1 as per PR #10786"

I've deliberately framed this as a vote to do the backport rather than one to do the revert. I think a decision like this to backport such a breaking change should be a positive decision to make the change, rather than a failure to pass a vote to revert.

vdukhovni:
Vote proposal sounds good.

t-j-h:
LGTM. And I agree with the sentiment of wanting a positive vote.

mspncp:
Please see also the related issue #11028.

mattcaswell:
The vote on this was held:

"topic: We should backport the change to stop accepting certificates signed
       using SHA1 at security level 1 to 1.1.1 as per PR #10786"

The vote failed to pass:

+1: 2
 0: 1
-1: 4

kroeckx:
Is there a reason for this to be open?

mattcaswell:
> Is there a reason for this to be open?

No - it can be closed now.

PR #10787, opened 2020-01-08
Reduce the security bits for MD5 and SHA1 based signatures in TLS

kroeckx:
This has as effect that SHA1 and MD5+SHA1 are no longer supported at
security level 1. This has as effect that TLS < 1.2 is no longer
supported at the default security level of 1, and that you need to set
the security level to 0 to use TLS < 1.2.

The first commit is the commit from #10785.

This will probably not get applied to the 1.1.1 branch, only to master.

This currently has a lot of failing test, because all tests using TLS < 1.2 or DTLS < 1.2 now need to run at security level 0. I need to look at those tests again.

##### Checklist

### DIFF
--- a/ssl/t1_lib.c
+++ b/ssl/t1_lib.c
@@ -1119,8 +1119,26 @@ static int sigalg_security_bits(SSL_CTX *ctx, const SIGALG_LOOKUP *lu)
         return 0;
     if (md != NULL)
     {
+        int md_type = EVP_MD_type(md);
+
         /* Security bits: half digest bits */
         secbits = EVP_MD_size(md) * 4;
+        /*
+         * SHA1 and MD5 are known to be broken. Reduce security bits so that
+         * they're no longer accepted at security level 1. The real values don't
+         * really matter as long as they're lower than 80, which is our
+         * security level 1.
+         * https://eprint.iacr.org/2020/014 puts a chosen-prefix attack for
+         * SHA1 at 2^63.4 and MD5+SHA1 at 2^67.2
+         * https://documents.epfl.ch/users/l/le/lenstra/public/papers/lat.pdf
+         * puts a chosen-prefix attack for MD5 at 2^39.
+	 */
+        if (md_type == NID_sha1)
+            secbits = 64;
+        else if (md_type == NID_md5_sha1)
+            secbits = 67;
+        else if (md_type == NID_md5)
+            secbits = 39;
     } else {
         /* Values from https://tools.ietf.org/html/rfc8032#section-8.5 */
         if (lu->sigalg == TLSEXT_SIGALG_ed25519)
--- a/test/recipes/70-test_renegotiation.t
+++ b/test/recipes/70-test_renegotiation.t
@@ -56,7 +56,8 @@ SKIP: {
     #        handshake
     $proxy->clear();
     $proxy->filter(undef);
-    $proxy->clientflags("-no_tls1_3");
+    $proxy->ciphers("DEFAULT:\@SECLEVEL=0");
+    $proxy->clientflags("-no_tls1_3 -cipher AES128-SHA:\@SECLEVEL=0");
     $proxy->serverflags("-no_tls1_3 -no_tls1_2");
     $proxy->reneg(1);
     $proxy->start();
--- a/test/recipes/70-test_sslsigalgs.t
+++ b/test/recipes/70-test_sslsigalgs.t
@@ -138,32 +138,32 @@ SKIP: {
 
     $proxy->filter(\&sigalgs_filter);
 
-    #Test 10: Sending no sig algs extension in TLSv1.2 should succeed at
-    #         security level 1
+    #Test 10: Sending no sig algs extension in TLSv1.2 will make it use
+    #         SHA1, which is only supported at security level 0.
     $proxy->clear();
     $testtype = NO_SIG_ALGS_EXT;
-    $proxy->clientflags("-no_tls1_3 -cipher DEFAULT:\@SECLEVEL=1");
-    $proxy->ciphers("ECDHE-RSA-AES128-SHA:\@SECLEVEL=1");
+    $proxy->clientflags("-no_tls1_3 -cipher DEFAULT:\@SECLEVEL=0");
+    $proxy->ciphers("ECDHE-RSA-AES128-SHA:\@SECLEVEL=0");
     $proxy->start();
-    ok(TLSProxy::Message->success, "No TLSv1.2 sigalgs seclevel 1");
+    ok(TLSProxy::Message->success, "No TLSv1.2 sigalgs seclevel 0");
 
     #Test 11: Sending no sig algs extension in TLSv1.2 should fail at security
-    #         level 2 since it will try to use SHA1. Testing client at level 1,
-    #         server level 2.
-    $proxy->clear();
-    $testtype = NO_SIG_ALGS_EXT;
-    $proxy->clientflags("-tls1_2 -cipher DEFAULT:\@SECLEVEL=1");
-    $proxy->ciphers("DEFAULT:\@SECLEVEL=2");
-    $proxy->start();
-    ok(TLSProxy::Message->fail, "No TLSv1.2 sigalgs server seclevel 2");
-
-    #Test 12: Sending no sig algs extension in TLSv1.2 should fail at security
-    #         level 2 since it will try to use SHA1. Testing client at level 2,
+    #         level 1 since it will try to use SHA1. Testing client at level 0,
     #         server level 1.
     $proxy->clear();
     $testtype = NO_SIG_ALGS_EXT;
-    $proxy->clientflags("-tls1_2 -cipher DEFAULT:\@SECLEVEL=2");
+    $proxy->clientflags("-tls1_2 -cipher DEFAULT:\@SECLEVEL=0");
     $proxy->ciphers("DEFAULT:\@SECLEVEL=1");
+    $proxy->start();
+    ok(TLSProxy::Message->fail, "No TLSv1.2 sigalgs server seclevel 1");
+
+    #Test 12: Sending no sig algs extension in TLSv1.2 should fail at security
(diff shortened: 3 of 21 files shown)


richsalz:
This should have a really obvious user-visible change notice (NEWS or CHANGES or ...)

ssl/t1_lib.c
```diff
@@ -1062,6 +1062,48 @@ int tls_check_sigalg_curve(const SSL *s, int curve)
...
+        secbits = EVP_MD_size(md) * 4;
+	/*
+	 * SHA1 and MD5 are known broken. Reduce security bits so that they're
+	 * no longer accepted at security level 1. The real values don't
+	 * really matter as long as they're lower than 80, which is our
+	 * security level 1.
+	 * https://eprint.iacr.org/2020/014 puts a chosen-prefix attack for
+	 * SHA1 at 2^63.4 and MD5+SHA1 at 2^67.2
+	 * https://documents.epfl.ch/users/l/le/lenstra/public/papers/lat.pdf
+	 * puts puts a chosen-prefix attack for MD5 at 2^39.
```
FdaSilvaYY: nit : comment block indentation is is of by 2.
typo : repeated `puts`

kroeckx: Somehow I ended up with tabs, while vim should be set up to never use tabs.

CHANGES
```diff
@@ -9,6 +9,11 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) The security strength of SHA1 and MD5 based signatures has been reduced.
+     This results in SSL 3, TLS 1.0, TLS 1.1 and DTLS 1.0 no longer working at
+     the default security level of 1 and instead requires security level 0.
```
mattcaswell: A reference to how people can change the security levels would be useful.

CHANGES
```diff
@@ -9,6 +9,11 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) The security strength of SHA1 and MD5 based signatures has been reduced.
```
mattcaswell: We're talking X509 and TLS signatures specifically right?

CHANGES
```diff
@@ -9,6 +9,11 @@
...
+     This results in SSL 3, TLS 1.0, TLS 1.1 and DTLS 1.0 no longer working at
```
mattcaswell: It also results in X509 certs not verifying outside of TLS if they are signed using SHA1?

t8m: And if so, is there actually a workaround to reenable if you want to verify a SHA1 signed X.509 outside of TLS? Because for TLS you do it via the cipher string.

vdukhovni: IIRC, when verifying X.509 certificate chains outside of TLS, the default security level is 0. To get some other security level and "apps" user would have to pass an explicit "-auth_level" option, while an application would have to invoke an appropriate X509_VERIFY_PARAM call to set a non-zero security level.

This means we're changing the meaning of level 1 for applications that choose to apply level 1, but outside TLS the default level is IIRC still 0.

vdukhovni: Just checked, as the default security level is "-1" (not set), and with that certificate checks don't check the algorithm security levels.  See:
https://github.com/openssl/openssl/blob/master/crypto/x509/x509_vpm.c#L468-L530

kroeckx: No, the X509 related changes are in #10786, this PR is only about the signature in SSL/TLS. I've updated the commit and CHANGES entry to be more clear about it.

vdukhovni: So this PR looks reasonable, and seems to do what it claims to do.  What remains I guess is to carefully adjust the tests until they pass, making sure that all the adjustments are what we expect, and don't just sweep any problems under the rug...

kroeckx:
I've updated the tests.

kroeckx:
Can somebody review this?

test/recipes/70-test_renegotiation.t
```diff
@@ -56,7 +56,8 @@ SKIP: {
     #        handshake
     $proxy->clear();
     $proxy->filter(undef);
-    $proxy->clientflags("-no_tls1_3");
+    $proxy->ciphers("DEFAULT:\@SECLEVEL=0");
+    $proxy->clientflags("-no_tls1_3 -cipher AES128-SHA:\@SECLEVEL=0");
```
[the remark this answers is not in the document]
kroeckx: ciphers() only affects the server, not the client. There is no call to do it for the client as far as I can see.

kroeckx:
So does anybody want to review this?

test/recipes/70-test_sslsigalgs.t
```diff
@@ -136,32 +136,32 @@ SKIP: {
...
+    $proxy->clientflags("-tls1_2 -cipher DEFAULT:\@SECLEVEL=0");
+    $proxy->ciphers("DEFAULT:\@SECLEVEL=1");
     $proxy->start();
-    ok(TLSProxy::Message->fail, "No TLSv1.2 sigalgs server seclevel 2");
+    ok(TLSProxy::Message->fail, "No TLSv1.2 sigalgs server seclevel 1");
 
     #Test 12: Sending no sig algs extension in TLSv1.2 should fail at security
-    #         level 2 since it will try to use SHA1. Testing client at level 2,
-    #         server level 1.
+    #         level 2 since it will try to use SHA1. Testing client at level 1,
```
vdukhovni, line 159:
```suggestion
    #         level 1 since it will try to use SHA1. Testing client at level 1,
```

vdukhovni:
Please rebase so that CI tests will run...

test/recipes/70-test_sslsigalgs.t
kroeckx: Fixed

kroeckx:
rebased

kroeckx:
I don't remember any discussion about making this configurable. If you think this should be configurable, how do you suggest we do that?

vdukhovni:
> I don't remember any discussion about making this configurable. If you think this should be configurable, how do you suggest we do that?

Well, it would have to be in the configuration file, from there a new option in the SSL_CTX and/or SSL handle, and downward into X509_VERIFY_PARAM, because this is needed both in libssl to accept the TLS sigalg and in X.509 to accept certificate signatures.

Just implementing @AUTHLEVEL=0 (which has been requested in the past) is not quite what's needed, because on the one hand it permits too much (also weak public keys, ...), and on the other not enough (no effect on TLS sigalgs).

vdukhovni:
Any idea what the travis failure is about?

kroeckx:
> Any idea what the travis failure is about?

It's not related to this PR, I've seen in broken in other PRs too.

kroeckx:
Note that this PR does not have any effect on X509, just the TLS sigalgs. If you think we need more configuration support for this, I suggest we do that in an other PR.

mattcaswell:
> I forget, were we going to make this configurable in master, or unconditional?

I see no reason to make this configurable. We already have a configuration option to control this "@SECLEVEL". Having a configuration option to control how the configuration option behaves seems a bit nonsensical to me.

mattcaswell:
This seems like a significant breaking change. Should there be an OMC vote for it?

Darkspirit:
If the outcome is no, would you leave it insecure by default? ;) It would be really nice if you could vote about uplifting/backporting it into 1.1.1 by the end of the year - against original intent.

[Debian 10 (Stable)](https://wiki.debian.org/ContinuousIntegration/TriagingTips/openssl-1.1.1) and [Ubuntu **18**.04](https://packages.ubuntu.com/bionic/libssl1.1) ([debian/patches/Set-systemwide-default-settings-for-libssl-users.patch](http://archive.ubuntu.com/ubuntu/pool/main/o/openssl/openssl_1.1.1-1ubuntu2.1~18.04.5.debian.tar.xz)) already enforce SECLEVEL=2 by default. That covers mail servers as well.
GnuTLS wants to do this change as well: https://gitlab.com/gnutls/gnutls/-/issues/910

TLS 1.0 and 1.1 have been re-enabled in Chrome Stable and Firefox Stable (so that desperate users can access every carelessly configured government website in this exceptional situation), but are left disabled in Beta and Dev/Nightly. These websites are getting fixed.

The last known DTLS 1.0 user was Jitsi, but they have fixed it and web browsers will probably disable it in the summer. Originally it was supposed to be disabled in [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1615445) and [Chrome](https://bugs.chromium.org/p/webrtc/issues/detail?id=10261) last month. It was [left disabled](https://bugzilla.mozilla.org/show_bug.cgi?id=1623511#c10) in Firefox Nightly.

Chrome and Firefox want to disable the RSA_PKCS1_SHA1 signature algorithm in TLS 1.2 as soon as possible. Every announcement to nudge these few remaining large websites to update or replace their load balancers is highly appreciated.
https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/yaJcs4p9LNI
https://bugs.chromium.org/p/chromium/issues/detail?id=658905
https://github.com/ssllabs/ssllabs-scan/issues/465#issuecomment-554669595

mspncp:
> This seems like a significant breaking change. Should there be an OMC vote for it?

This question was asked by @mattcaswell  [a month ago](https://github.com/openssl/openssl/pull/10787#issuecomment-619825578). IMHO it deserves at least a discussion, so I placed an omc hold label on his behalf as a reminder. Matt, if you changed your mind, please clear it again.

mattcaswell:
I'm lifting the OMC hold on this PR. The OMC have voted on this:

> topic: The TLS security bit values for MD5, MD5_SHA1 and SHA1 should be set to
       39, 67 and 65 respectively in OpenSSL 3.0. Consequently TLS < 1.2 will
       be disallowed in the default security level

The result was:

+1: 4
 0: 1
-1: 2

So the vote passed and this PR can go ahead.

kroeckx:
So it's now waiting for someone to review it

PR #10789, opened 2020-01-08
Deprecate the low level RIPEMD160 functions.

paulidale:
Use of the low level RIPEMD160 functions has been informally discouraged for a long time. We now formally deprecate them.

### DIFF
--- a/include/openssl/ripemd.h
+++ b/include/openssl/ripemd.h
@@ -18,18 +18,18 @@
 
 # include <openssl/opensslconf.h>
 
-#ifndef OPENSSL_NO_RMD160
-# include <openssl/e_os2.h>
-# include <stddef.h>
-# ifdef  __cplusplus
+# ifndef OPENSSL_NO_RMD160
+#  include <openssl/e_os2.h>
+#  include <stddef.h>
+#  ifdef  __cplusplus
 extern "C" {
-# endif
+#  endif
 
-# define RIPEMD160_LONG unsigned int
+#  define RIPEMD160_LONG unsigned int
 
-# define RIPEMD160_CBLOCK        64
-# define RIPEMD160_LBLOCK        (RIPEMD160_CBLOCK/4)
-# define RIPEMD160_DIGEST_LENGTH 20
+#  define RIPEMD160_CBLOCK        64
+#  define RIPEMD160_LBLOCK        (RIPEMD160_CBLOCK/4)
+#  define RIPEMD160_DIGEST_LENGTH 20
 
 typedef struct RIPEMD160state_st {
     RIPEMD160_LONG A, B, C, D, E;
@@ -44,10 +44,8 @@ int RIPEMD160_Final(unsigned char *md, RIPEMD160_CTX *c);
 unsigned char *RIPEMD160(const unsigned char *d, size_t n, unsigned char *md);
 void RIPEMD160_Transform(RIPEMD160_CTX *c, const unsigned char *b);
 
-# ifdef  __cplusplus
+#  ifdef  __cplusplus
 }
+#  endif
 # endif
-# endif
-
-
 #endif
--- a/include/openssl/ripemd.h
+++ b/include/openssl/ripemd.h
@@ -21,15 +21,18 @@
 # ifndef OPENSSL_NO_RMD160
 #  include <openssl/e_os2.h>
 #  include <stddef.h>
+
+#  define RIPEMD160_DIGEST_LENGTH 20
+
 #  ifdef  __cplusplus
 extern "C" {
 #  endif
+#  if !defined(OPENSSL_NO_DEPRECATED_3_0)
 
-#  define RIPEMD160_LONG unsigned int
+#   define RIPEMD160_LONG unsigned int
 
-#  define RIPEMD160_CBLOCK        64
-#  define RIPEMD160_LBLOCK        (RIPEMD160_CBLOCK/4)
-#  define RIPEMD160_DIGEST_LENGTH 20
+#   define RIPEMD160_CBLOCK        64
+#   define RIPEMD160_LBLOCK        (RIPEMD160_CBLOCK/4)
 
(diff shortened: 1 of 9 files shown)


include/openssl/ripemd.h
```diff
@@ -18,36 +18,39 @@
...
 
-# define RIPEMD160_LONG unsigned int
+#   define RIPEMD160_LONG unsigned int
 
-# define RIPEMD160_CBLOCK        64
-# define RIPEMD160_LBLOCK        (RIPEMD160_CBLOCK/4)
-# define RIPEMD160_DIGEST_LENGTH 20
+#   define RIPEMD160_CBLOCK        64
+#   define RIPEMD160_LBLOCK        (RIPEMD160_CBLOCK/4)
+#   define RIPEMD160_DIGEST_LENGTH 20
```
mattcaswell: Arguably this define could still be useful even if not using the low level API. Should we retain it?

paulidale: Done.

levitte: I'm not sure I understand how it's useful if all other RIPEMD160 stuff is gone...

mattcaswell: RIPEMD160 is not going, only the low level API to it. The digest length doesn't change whether you're accessing it via the low level API or EVP. If via EVP that's something you still might want to know. For example here's a use of SHA512_DIGEST_LENGTH in a function that only uses the EVP interface:

https://github.com/openssl/openssl/blob/master/test/evp_extra_test.c#L1217

Now I'm not suggesting we should go adding these for all digests that we might add in the future. But I also see no reason to actually deprecate them.

levitte: I should have added "... from public view", 'cause that's basically what this is all about, yeah?

As for the use in `evp_extra_test.c`, that's quite arbitrary, isn't it?  Yeah, it looks nice with a macro, but would could just as well have been this and the test would run just as fine:

``` diff
diff --git a/test/evp_extra_test.c b/test/evp_extra_test.c
index 387e29496f..ed5b77c0b0 100644
--- a/test/evp_extra_test.c
+++ b/test/evp_extra_test.c
@@ -1201,6 +1201,10 @@ done:
 #endif /* OPENSSL_NO_EC */
 
 #ifndef OPENSSL_NO_DSA
+
+/* The exact number doesn't matter, as long as it's not zero */
+#define TEST_MDSIZE 100
+
 /* Test getting and setting parameters on an EVP_PKEY_CTX */
 static int test_EVP_PKEY_CTX_get_set_params(void)
 {
@@ -1214,7 +1218,7 @@ static int test_EVP_PKEY_CTX_get_set_params(void)
     EVP_PKEY *pkey = NULL;
     int ret = 0;
     const EVP_MD *md;
-    size_t mdsize = SHA512_DIGEST_LENGTH;
+    size_t mdsize = TEST_MDSIZE;
     char ssl3ms[48];
 
     /*
@@ -1289,7 +1293,7 @@ static int test_EVP_PKEY_CTX_get_set_params(void)
 
     mdsize = 0;
     if (!TEST_true(EVP_PKEY_CTX_get_params(ctx, ourparams))
-            || !TEST_size_t_eq(mdsize, SHA512_DIGEST_LENGTH))
+            || !TEST_size_t_eq(mdsize, TEST_MDSIZE))
         goto err;
 
     /*
```

levitte: (I just tested that change, it does work fine)

mattcaswell: Yes. It *could* have been written that way. But it wasn't. And I suspect there are many applications out there that use these macros. Even if they are using the EVP interface. I see no problem with those applications continuing to use those macros. No need to cause unnecessary changes to those applications with spurious deprecation warnings when its not needed.

levitte: Do note that they won't get a warning for the macros, just for the functions, i.e. when they attempt to use the low-level interface.  Quite frankly, though, it may actually be a bad thing to do what you suggest, i.e. use them with EVP functionality.  With EVP, if you want to know the size of a hash function, *you ask* rather than assume.  `EVP_MD_size()` has been around for long enough for people to have learned.

mattcaswell: I disagree. The length of SHA256 is fixed, and is not going to change. I see absolutely no problem with using a macro for that length. Yes the way forward for new digests is just to use EVP_MD_size(). But I don't see a need to make existing applications change and it represents a negligible maintenance burden for us to retain these.

levitte: I guess we will continue to disagree, and time will tell.  I'm dropping this one for now, I guess @paulidale will have heard our wisdom and will make a final choice.
(I'm learning to choose my battles :wink:)

paulidale:
Is a bulk CHANGES entry for all these deprecations appropriate?

mattcaswell:
> Is a bulk CHANGES entry for all these deprecations appropriate?

I'd be ok with that if you promise to create one (perhaps create a WIP one now which we update for each PR you do)? It should list all the functions that have been deprecated IMO.

include/openssl/ripemd.h
paulidale: I'm okay either way.

I'd prefer to deprecate them to cleanse things but the overhead for keeping them is effectively zero.  There are some where this is better not done because something relies on them: the tests, an engine, libssl and even some of libcrypto needed access to some of the digest sizes but none of the functions.  If some are retained, keeping them all is at least consistent.

paulidale:
CHANGES entry in #10802

include/openssl/ripemd.h
t8m: I would go with the consistency. So if we do not deprecate all the size macros, we should keep all of them.

kaduk: +1 for consistency; my personal preference is to keep the ones that are already around and not deprecate them

paulidale:
Merged to master.  Thanks all.

PR #10790, opened 2020-01-09
Deprecate the low level MD2, MD4 and MDC2 functions.

paulidale:
Use of the low level MD2, MD4 and MDC2 functions have been informally discouraged for a long time. We now formally deprecate them.

There is a change in the documentation which impacts MD5 as well but the corresponding code changes haven't been made yet.

### DIFF
--- a/crypto/md2/build.info
+++ b/crypto/md2/build.info
@@ -1,3 +1,10 @@
 LIBS=../../libcrypto
-SOURCE[../../libcrypto]=\
-        md2_dgst.c md2_one.c
+
+SOURCE[../../libcrypto]=md2_dgst.c md2_one.c
+
+# When all deprecated symbols are removed, libcrypto doesn't export the
+# MD2 functions, so we must include them directly in liblegacy.a
+IF[{- $disabled{"deprecated"}
+      && (defined $config{"api"} && $config{"api"} >= 30000) -}]
+  SOURCE[../../providers/liblegacy.a]=md2_dgst.c md2_one.c
+ENDIF
--- a/crypto/md4/build.info
+++ b/crypto/md4/build.info
@@ -1,3 +1,10 @@
 LIBS=../../libcrypto
-SOURCE[../../libcrypto]=\
-        md4_dgst.c md4_one.c
+
+SOURCE[../../libcrypto]=md4_dgst.c md4_one.c
+
+# When all deprecated symbols are removed, libcrypto doesn't export the
+# MD4 functions, so we must include them directly in liblegacy.a
+IF[{- $disabled{"deprecated"}
+      && (defined $config{"api"} && $config{"api"} >= 30000) -}]
+  SOURCE[../../providers/liblegacy.a]=md4_dgst.c md4_one.c
+ENDIF
--- a/crypto/mdc2/build.info
+++ b/crypto/mdc2/build.info
@@ -1,3 +1,10 @@
 LIBS=../../libcrypto
-SOURCE[../../libcrypto]=\
-        mdc2dgst.c mdc2_one.c
+
+SOURCE[../../libcrypto]=mdc2dgst.c mdc2_one.c
+
+# When all deprecated symbols are removed, libcrypto doesn't export the
+# MDC2 functions, so we must include them directly in liblegacy.a
+IF[{- $disabled{"deprecated"}
+      && (defined $config{"api"} && $config{"api"} >= 30000) -}]
+  SOURCE[../../providers/liblegacy.a]=mdc2dgst.c mdc2_one.c
+ENDIF
--- a/include/openssl/md2.h
+++ b/include/openssl/md2.h
@@ -19,15 +19,15 @@
 # include <openssl/opensslconf.h>
 
 # ifndef OPENSSL_NO_MD2
-# include <stddef.h>
-# ifdef  __cplusplus
+#  include <stddef.h>
+#  ifdef  __cplusplus
 extern "C" {
-# endif
+#  endif
 
 typedef unsigned char MD2_INT;
 
-# define MD2_DIGEST_LENGTH       16
-# define MD2_BLOCK               16
+#  define MD2_DIGEST_LENGTH       16
+#  define MD2_BLOCK               16
 
 typedef struct MD2state_st {
     unsigned int num;
@@ -42,9 +42,9 @@ int MD2_Update(MD2_CTX *c, const unsigned char *data, size_t len);
 int MD2_Final(unsigned char *md, MD2_CTX *c);
 unsigned char *MD2(const unsigned char *d, size_t n, unsigned char *md);
 
-# ifdef  __cplusplus
+#  ifdef  __cplusplus
 }
-# endif
+#  endif
 # endif
 
 #endif
--- a/include/openssl/md2.h
+++ b/include/openssl/md2.h
@@ -24,10 +24,13 @@
 extern "C" {
 #  endif
 
+#  define MD2_DIGEST_LENGTH       16
+
+#  if !defined(OPENSSL_NO_DEPRECATED_3_0)
+
 typedef unsigned char MD2_INT;
 
-#  define MD2_DIGEST_LENGTH       16
-#  define MD2_BLOCK               16
+#   define MD2_BLOCK               16
 
 typedef struct MD2state_st {
     unsigned int num;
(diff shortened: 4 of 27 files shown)


include/openssl/md2.h
```diff
@@ -19,32 +19,37 @@
...
 
 # ifndef OPENSSL_NO_MD2
-# include <stddef.h>
-# ifdef  __cplusplus
+#  if !defined(OPENSSL_NO_DEPRECATED_3_0)
+#   include <stddef.h>
+#   ifdef  __cplusplus
 extern "C" {
-# endif
+#   endif
```
levitte: I would put the deprecation guard after this. That will make things slightly less ugly

paulidale: This slightly changes the semantics by including `stddef.h`.  Should be harmless I know.

paulidale: Which is what this PR did and shouldn't have.  I'll fix.

include/openssl/md2.h
```diff
@@ -19,32 +19,37 @@
...
+
+DEPRECATEDIN_3_0(const char *MD2_options(void))
+DEPRECATEDIN_3_0(int MD2_Init(MD2_CTX *c))
+DEPRECATEDIN_3_0(int MD2_Update(MD2_CTX *c, const unsigned char *data,
+                                size_t len))
+DEPRECATEDIN_3_0(int MD2_Final(unsigned char *md, MD2_CTX *c))
+DEPRECATEDIN_3_0(unsigned char *MD2(const unsigned char *d, size_t n,
+                 unsigned char *md))
+
+#  if !defined(OPENSSL_NO_DEPRECATED_3_0)
```
t8m: This looks weird, the initial `extern "C" `is not inside `!defined(OPENSSL_NO_DEPRECATED_3_0)`

include/openssl/md4.h
```diff
@@ -19,39 +19,44 @@
...
-void MD4_Transform(MD4_CTX *c, const unsigned char *b);
+DEPRECATEDIN_3_0(int MD4_Init(MD4_CTX *c))
+DEPRECATEDIN_3_0(int MD4_Update(MD4_CTX *c, const void *data, size_t len))
+DEPRECATEDIN_3_0(int MD4_Final(unsigned char *md, MD4_CTX *c))
+DEPRECATEDIN_3_0(unsigned char *MD4(const unsigned char *d, size_t n,
+                 unsigned char *md))
+DEPRECATEDIN_3_0(void MD4_Transform(MD4_CTX *c, const unsigned char *b))
 
-# ifdef  __cplusplus
+#  if !defined(OPENSSL_NO_DEPRECATED_3_0)
```
t8m: Same here.

include/openssl/mdc2.h
```diff
@@ -18,31 +18,39 @@
...
-unsigned char *MDC2(const unsigned char *d, size_t n, unsigned char *md);
+DEPRECATEDIN_3_0(int MDC2_Init(MDC2_CTX *c))
+DEPRECATEDIN_3_0(int MDC2_Update(MDC2_CTX *c, const unsigned char *data,
+                 size_t len))
+DEPRECATEDIN_3_0(int MDC2_Final(unsigned char *md, MDC2_CTX *c))
+DEPRECATEDIN_3_0(unsigned char *MDC2(const unsigned char *d, size_t n,
+                 unsigned char *md))
 
-# ifdef  __cplusplus
+#  if !defined(OPENSSL_NO_DEPRECATED_3_0)
```
t8m: And here.

include/openssl/md2.h
paulidale: Missed this, the `extern "C"` was originally inside the `!defined(OPENSSL_NO_DEPRECATED_3_0)` but I moved it later after @levitte asked.

paulidale: Fixed now.

crypto/md2/build.info
```diff
@@ -1,3 +1,10 @@
 LIBS=../../libcrypto
-SOURCE[../../libcrypto]=\
-        md2_dgst.c md2_one.c
+
+SOURCE[../../libcrypto]=md2_dgst.c md2_one.c
+
+# When all deprecated symbols are removed, libcrypto doesn't export the
+# WHIRLPOOL functions, so we must include them directly in liblegacy.a
```
mattcaswell: copy and paste error here. s/WHIRLPOOL/md2/

crypto/md4/build.info
```diff
@@ -1,3 +1,10 @@
 LIBS=../../libcrypto
-SOURCE[../../libcrypto]=\
-        md4_dgst.c md4_one.c
+
+SOURCE[../../libcrypto]=md4_dgst.c md4_one.c
+
+# When all deprecated symbols are removed, libcrypto doesn't export the
+# WHIRLPOOL functions, so we must include them directly in liblegacy.a
```
mattcaswell: and here

crypto/mdc2/build.info
```diff
@@ -1,3 +1,10 @@
 LIBS=../../libcrypto
-SOURCE[../../libcrypto]=\
-        mdc2dgst.c mdc2_one.c
+
+SOURCE[../../libcrypto]=mdc2dgst.c mdc2_one.c
+
+# When all deprecated symbols are removed, libcrypto doesn't export the
+# WHIRLPOOL functions, so we must include them directly in liblegacy.a
```
mattcaswell: and here

crypto/rsa/rsa_sign.c
```diff
@@ -7,6 +7,12 @@
  * https://www.openssl.org/source/license.html
  */
 
+/*
+ * Some low level APIs are deprecated for public use, but still ok for
+ * internal use.
+ */
+#include "internal/deprecated.h"
```
mattcaswell: AFAICT this just seems to be for the various DIGEST_LENGTH macros. Perhaps we don't deprecate them?

include/openssl/md2.h
```diff
@@ -19,32 +19,35 @@
...
 extern "C" {
-# endif
+#  endif
+#  if !defined(OPENSSL_NO_DEPRECATED_3_0)
 
 typedef unsigned char MD2_INT;
 
-# define MD2_DIGEST_LENGTH       16
-# define MD2_BLOCK               16
+#   define MD2_DIGEST_LENGTH       16
```
mattcaswell: As noted above, perhaps don't deprecate this?

include/openssl/md4.h
```diff
@@ -19,39 +19,42 @@
...
  */
-# define MD4_LONG unsigned int
+#   define MD4_LONG unsigned int
 
-# define MD4_CBLOCK      64
-# define MD4_LBLOCK      (MD4_CBLOCK/4)
-# define MD4_DIGEST_LENGTH 16
+#   define MD4_CBLOCK      64
+#   define MD4_LBLOCK      (MD4_CBLOCK/4)
+#   define MD4_DIGEST_LENGTH 16
```
mattcaswell: and here

include/openssl/mdc2.h
```diff
@@ -18,31 +18,37 @@
 
 # include <openssl/opensslconf.h>
 
-#ifndef OPENSSL_NO_MDC2
-# include <stdlib.h>
-# include <openssl/des.h>
-# ifdef  __cplusplus
+# ifndef OPENSSL_NO_MDC2
+
+#  define MDC2_DIGEST_LENGTH      16
```
mattcaswell: But here you already decided not to deprecate!

paulidale: I had a reason but don't remember why at the moment and can't figure it out.

crypto/rsa/rsa_sign.c
paulidale: I'm not sure which way we should go.  Deprecating them seems like the right thing to do to reduce the APIs but they are never going to change.

The lengths can be determined via EVP, not that it helps here.

mattcaswell: Maintaining them seems ok to me and even quite useful. There's no reason to break code that just uses these defines.

paulidale: Okay, I'll update all of these in the morning.

paulidale: Updated here, will do the other PRs soon.

paulidale:
CHANGES entry in #10802

paulidale:
Merged to master.  Thanks for the feedback.

PR #10791, opened 2020-01-09
Deprecate the low level SHA* functions.

paulidale:
Use of the low level SHA-1 and SHA-2 functions has been informally discouraged for a long time. We now formally deprecate them.

### DIFF
--- a/crypto/ct/ct_log.c
+++ b/crypto/ct/ct_log.c
@@ -76,14 +76,14 @@ static int ct_v1_log_id_from_pkey(EVP_PKEY *pkey,
     int ret = 0;
     unsigned char *pkey_der = NULL;
     int pkey_der_len = i2d_PUBKEY(pkey, &pkey_der);
+    unsigned int len;
 
     if (pkey_der_len <= 0) {
         CTerr(CT_F_CT_V1_LOG_ID_FROM_PKEY, CT_R_LOG_KEY_INVALID);
         goto err;
     }
 
-    SHA256(pkey_der, pkey_der_len, log_id);
-    ret = 1;
+    ret = EVP_Digest(pkey_der, pkey_der_len, log_id, &len, EVP_sha256(), NULL);
 err:
     OPENSSL_free(pkey_der);
     return ret;
--- a/crypto/ec/curve25519.c
+++ b/crypto/ec/curve25519.c
@@ -9,6 +9,7 @@
 
 #include <string.h>
 #include "ec_local.h"
+#include <openssl/evp.h>
 #include <openssl/sha.h>
 
 #if defined(X25519_ASM) && (defined(__x86_64) || defined(__x86_64__) || \
@@ -5436,39 +5437,50 @@ int ED25519_sign(uint8_t *out_sig, const uint8_t *message, size_t message_len,
     uint8_t nonce[SHA512_DIGEST_LENGTH];
     ge_p3 R;
     uint8_t hram[SHA512_DIGEST_LENGTH];
-    SHA512_CTX hash_ctx;
+    EVP_MD *sha512 = EVP_MD_fetch(NULL, SN_sha512, NULL);
+    EVP_MD_CTX *hash_ctx = EVP_MD_CTX_new();
+    unsigned int sz;
+    int res = 0;
 
-    SHA512_Init(&hash_ctx);
-    SHA512_Update(&hash_ctx, private_key, 32);
-    SHA512_Final(az, &hash_ctx);
+    if (sha512 == NULL || hash_ctx == NULL)
+        goto err;
+
+    if (!EVP_DigestInit_ex(hash_ctx, sha512, NULL)
+        || !EVP_DigestUpdate(hash_ctx, private_key, 32)
+        || !EVP_DigestFinal_ex(hash_ctx, az, &sz))
+        goto err;
 
     az[0] &= 248;
     az[31] &= 63;
     az[31] |= 64;
 
-    SHA512_Init(&hash_ctx);
-    SHA512_Update(&hash_ctx, az + 32, 32);
-    SHA512_Update(&hash_ctx, message, message_len);
-    SHA512_Final(nonce, &hash_ctx);
+    if (!EVP_DigestInit_ex(hash_ctx, sha512, NULL)
+        || !EVP_DigestUpdate(hash_ctx, az + 32, 32)
+        || !EVP_DigestUpdate(hash_ctx, message, message_len)
+        || !EVP_DigestFinal_ex(hash_ctx, nonce, &sz))
+        goto err;
 
     x25519_sc_reduce(nonce);
     ge_scalarmult_base(&R, nonce);
     ge_p3_tobytes(out_sig, &R);
 
-    SHA512_Init(&hash_ctx);
-    SHA512_Update(&hash_ctx, out_sig, 32);
-    SHA512_Update(&hash_ctx, public_key, 32);
-    SHA512_Update(&hash_ctx, message, message_len);
-    SHA512_Final(hram, &hash_ctx);
+    if (!EVP_DigestInit_ex(hash_ctx, sha512, NULL)
+        || !EVP_DigestUpdate(hash_ctx, out_sig, 32)
+        || !EVP_DigestUpdate(hash_ctx, public_key, 32)
+        || !EVP_DigestUpdate(hash_ctx, message, message_len)
+        || !EVP_DigestFinal_ex(hash_ctx, hram, &sz))
+        goto err;
 
     x25519_sc_reduce(hram);
     sc_muladd(out_sig + 32, hram, az, nonce);
 
-    OPENSSL_cleanse(&hash_ctx, sizeof(hash_ctx));
+    res = 1;
+err:
     OPENSSL_cleanse(nonce, sizeof(nonce));
     OPENSSL_cleanse(az, sizeof(az));
-
-    return 1;
+    EVP_MD_free(sha512);
+    EVP_MD_CTX_free(hash_ctx);
+    return res;
 }
 
 static const char allzeroes[15];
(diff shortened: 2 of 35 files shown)


paulidale:
Builds with deprecation will fail because `ssl/s3_cbc.c` requires direct access to these functions and it hasn't been moved to a provider.

crypto/ct/ct_log.c
```diff
@@ -7,6 +7,12 @@
  * https://www.openssl.org/source/license.html
  */
 
+/*
+ * SHA256 low level APIs are deprecated for public use, but still ok for
+ * internal use.
+ */
+#include "internal/deprecated.h"
```
mattcaswell: This file seems to be an example where we *should* convert to using EVP.

crypto/ec/curve25519.c
```diff
@@ -7,6 +7,12 @@
  * https://www.openssl.org/source/license.html
  */
 
+/*
+ * SHA512 low level APIs are deprecated for public use, but still ok for
+ * internal use.
```
mattcaswell: Seems we should convert this to EVP

crypto/ec/ecx_meth.c
```diff
@@ -7,6 +7,12 @@
  * https://www.openssl.org/source/license.html
  */
 
+/*
+ * SHA-512 low level APIs are deprecated for public use, but still ok for
+ * internal use.
+ */
+#include "internal/deprecated.h"
```
mattcaswell: Convert to EVP?

crypto/evp/e_des.c
```diff
@@ -7,6 +7,12 @@
  * https://www.openssl.org/source/license.html
  */
 
+/*
+ * SHA-1 low level APIs are deprecated for public use, but still ok for
+ * internal use.
+ */
+#include "internal/deprecated.h"
```
mattcaswell: What is the SHA usage in this file that requires this? I couldn't find it.

crypto/evp/e_des3.c
```diff
@@ -7,6 +7,12 @@
  * https://www.openssl.org/source/license.html
  */
 
+/*
+ * SHA-1 low level APIs are deprecated for public use, but still ok for
+ * internal use.
+ */
```
mattcaswell: Convert to EVP?

engines/e_dasync.c
```diff
@@ -7,6 +7,13 @@
  * https://www.openssl.org/source/license.html
  */
 
+/*
+ * SHA-1 low level APIs are deprecated for public use, but still ok for
+ * internal use.  Note, that due to symbols not being exported, only the
+ * #defines can be accessed, in this case SHA_CBLOCK.
```
mattcaswell: Perhaps just define SHA_CBLOCK locally in this file rather than including deprecated.h?

engines/e_ossltest.c
```diff
@@ -13,6 +13,13 @@
  * used for any purpose except testing
  */
 
+/*
+ * SHA low level APIs are deprecated for public use, but still ok for
+ * internal use.  Note, that due to symbols not being exported, only the
+ * #defines can be accessed.
+ */
+#include "internal/deprecated.h"
```
mattcaswell: Perhaps just define SHA_CBLOCK locally?

paulidale:
CHANGES entry in #10802

crypto/evp/e_des.c
paulidale: Me either :(

crypto/evp/e_des3.c
levitte: Wait, what? This is an EVP backend...

paulidale: Yeah, there is some really crazy stuff in the code if you dig much beyond the superficial.

I was intending to convert this and the others to EVP next week.  Unless a good reason is given for not doing so.  Using EVP will add the cost of a level of function pointer indirection but could gain HW acceleration benefits.

Unless I'm missing something -- which seems to be quite common recently :(

engines/e_ossltest.c
levitte: I strongly disagree with having that defined "locally" in several places

levitte: That is, unless there's justification, such as values can vary depending on location. I assume that's not the case here.

paulidale: I'm with @levitte here, best keep the definition in one place.  For the test engine, I don't see a problem with accessing the symbols directly (it needs more than just `SHA_CBLOCK` BTW).  The async engine is example code unless I'm mistaken and it should do the right thing.

I could call `EVP_MD_meth_get_input_blocksize(EVP_sha())` or similar to get the value and set that into the created EVP_MD.  This seems like an all around better approach.

crypto/evp/e_des.c
paulidale: I'll take it out and see which CI breaks....

vdukhovni:
> Use of the low level SHA-1 and SHA-2 functions has been informally discouraged for a long time. We now formally deprecate them.

Just deprecate, right?  Not immediately remove...
I rather expect that various external projects that just need a specific hash function, and not much else from OpenSSL or EVP, may be using the underlying functions directly, or at least using EVP_sha1() rather than EVP_get_digestbyname("sha1").

Has anyone checked which projects might be affected? A common source of problems is foreign function bindings in other languages, e.g. crypto, or just hash function modules Perl, Python, Ruby, ... Where the re-exported function is part of the module API.

I suspect that we may not be able to remove these, and perhaps should not, but rather may end up sequestering them in the legacy provider (and/or suitable legacy crypto library where the interface is not EVP and so "providers" don't come into play).  If that's what deprecation means, that's OK. If the intent is to outright remove these functions in a later release, that may prove more difficult.

This is not to say I am opposed to doing this if the downstream impact is not too high, rather this is a note of concern, about perhaps burning too many bridges with the past (chequered as it may be).

levitte:
EVP_sha1() works as a sort of template, causing an implicit fetch of the corresponding provided implementation. This is by design.

Regarding removal of deprecated stuff, it means that the declarations will be removed from public headers, and will no longer be exported by shared libraries. As for the code itself, it will remain internally as long as we need it. If only our providers use it, I suspect it will physically move to that source tree. 

Does that answer your concerns, @vdukhovni ?

vdukhovni:
> Does that answer your concerns, @vdukhovni ?

It clarifies the intent, but the concern is that I don't know the extent to which external projects are then affected by this change. Perhaps old software can continue to use old versions of libcrypto, but the appetite for reworking stable legacy code is generally limited.

If we're providing a general purpose suite of algorithms, even historical ones, it is not clear that the legacy interfaces should be removed. Note again, I'm not saying "no", but raising some concern that perhaps the cost to others may exceed the benefit of doing this, unless there's some specific freedom this gives us to make improvements we could not make otherwise...

engines/e_ossltest.c
paulidale: All good plans seem to fail.  The engines are also directly accessing sizeof(SHA_CTX) and similar.

levitte: So?  the whole ENGINE API will become deprecated as well...  (I gotta finish up that branch!)

paulidale: This is better left for the ENGINE deprecation -- I can't see us dropping the either the test or async engine code soon.

I updated the comment to include the `sizeof(ctx)` bit.

crypto/evp/e_des3.c
paulidale: Removed this.

engines/e_dasync.c
paulidale: Added note that `sizeof(ctx)` is also required.

crypto/ec/ecx_meth.c
paulidale: Done.

crypto/ec/curve25519.c
paulidale: Done.

paulidale:
Out of WIP.

crypto/evp/e_des3.c
mattcaswell: > Wait, what? This is an EVP backend...

Its an EVP backend for DES not SHA1. It seems reasonable to still use EVP for the SHA-1 bits.

mattcaswell:
> It clarifies the intent, but the concern is that I don't know the extent to which external projects are then affected by this change. Perhaps old software can continue to use old versions of libcrypto, but the appetite for reworking stable legacy code is generally limited.

I'm surprised you bring this up. It has long been our intent to deprecate these low level APIs. At this stage though it is just deprecation. We might get a feel for  how many projects are impacted when 3.0 gets released and we see how many complaints we get about deprecation warnings. We are not tied to any particular timetable for the eventual removal of code - but our stability policy would apply, so this isn't happening any time soon.

crypto/ct/ct_log.c
paulidale: Converted to EVP.

paulidale:
Added MD5 here to reduce later merge conflicts.

paulidale:
Rebased and conflicts resolved (all bar in comments about internal use).

paulidale:
Ping reviewer.

test/build.info
```diff
@@ -601,6 +601,10 @@ IF[{- !$disabled{tests} -}]
     INCLUDE[curve448_internal_test]=.. ../include ../apps/include ../crypto/ec/curve448
     DEPEND[curve448_internal_test]=../libcrypto.a libtestutil.a
 
+    SOURCE[rc4test]=rc4test.c
```
mattcaswell: Same comment re internal tests as per the IDEA PR

crypto/evp/e_des3.c
levitte: > Its an EVP backend for DES not SHA1

Goo point.  Okie

paulidale:
> is it the intent to push all commits in this PR, including the "undo" one? Or are we waiting to push this until libssl is fixed so we don't need to have the "undo" commit at all? If the former we need to make sure we have a task somewhere to remind us to come back and fix this properly.

I'm okay either way.  Conflicts will be easier if this is pushed including the "undo" commit but they oughtn't be too terrible.  If this is pushed earlier, I'll create an issue to fix things.

mattcaswell:
> I'm okay either way. Conflicts will be easier if this is pushed including the "undo" commit but they oughtn't be too terrible. If this is pushed earlier, I'll create an issue to fix things.

I suggest we push and create an issue. Otherwise conflicts will be a pain.

test/build.info
paulidale: Done.

paulidale:
Redo changes are captured in #10877.
If (when) the relevant commit ID changes, I'll update the issue ticket.

paulidale:
Merged to master and undo issue updated.
Thanks for the reviewing and suggestions.

PR #10793, opened 2020-01-09
Configure: use $list_separator_re only for defines and includes

### LINKED ISSUE #10792: Can't use blibpath (for AIX) in LDFLAGS
On AIX, in order to use the "blibpath" linker flag (runtime library search path), I use the environment variable LDFLAGS like this:

    LDFLAGS="-Wl,blibpath:<foo>"

OpenSSL's Configure script interprets the colon character as a list separator (according to `$list_separator_re`) and eventually converts it to a space, causing the linker to fail ("ld: 0706-015 The -b libpath option needs a parameter.") 

The generated Makefile includes a different, "wrong" LDFLAGS:

```
# cd openssl-1.1.1d
# LDFLAGS="-Wl,blibpath:/usr/lib" ./Configure aix64-gcc
# grep -R blibpath .
./Makefile:LDFLAGS= -Wl,blibpath /usr/lib
./configdata.pm:  LDFLAGS => [ "-Wl,blibpath", "/usr/lib" ],
./configdata.pm:      "LDFLAGS" => "-Wl,blibpath:/usr/lib",
```

I could not find a way to change the list separator character or "escape" the colon character without editing the code of the configure script.

levitte:
This regexp was used a bit too uncontrolled, which had it split flag
values where it should not have.

Fixes #10792

### DIFF
--- a/Configure
+++ b/Configure
@@ -590,7 +590,7 @@ while ((my $first, my $second) = (shift @list, shift @list)) {
 
 &usage if ($#ARGV < 0);
 
-# For the "make variables" CINCLUDES and CDEFINES, we support lists with
+# For the "make variables" CPPINCLUDES and CPPDEFINES, we support lists with
 # platform specific list separators.  Users from those platforms should
 # recognise those separators from how you set up the PATH to find executables.
 # The default is the Unix like separator, :, but as an exception, we also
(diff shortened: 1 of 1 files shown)


Configure
```diff
@@ -590,7 +590,7 @@ while ((my $first, my $second) = (shift @list, shift @list)) {
 
 &usage if ($#ARGV < 0);
 
-# For the "make variables" CINCLUDES and CDEFINES, we support lists with
+# For the "make variables" CPPINCLUDES and CPPDEFINES, we support lists with
```
mattcaswell: Why do we do this only for `CPPINCLUDES` and not `CINCLUDES`?

levitte: CINCLUDES is a typo

wiggin15:
I confirmed that this fix resolves the issue reported in #10792 . Thank you!

paulidale (approved):

levitte:
Merged.

master:
bbe486cf6154df3d3aaedbae6c5b82d4ed31a5f8 Configure: use $list_separator_re only for defines and includes

1.1.1:
5f6343dea1233deb6d762e24e0e0d3d795b9513d Configure: use $list_separator_re only for defines and includes

PR #10796, opened 2020-01-09
Fix a seg fault in EVP_DigestSignUpdate()

mattcaswell:
If we're using an explicitly fetched digest in an `EVP_DigestUpdate`operation, then we should still go the legacy route if `EVP_MD_CTX_FLAG_NO_INIT` has been set because we are being used in the context of a legacy signature algorithm and `EVP_DigestInit` has not been called.


mattcaswell:

Ping?

mattcaswell:
Pushed. Thanks.

PR #10797, opened 2020-01-09
EXPERIMENTAL: remove all the legacy bits from a EVP_PKEY

levitte:
This adds `EVP_PKEY_deassign()`, which removes all the legacy bits from an `EVP_PKEY`, and adapts a few tests to use this function prior to using the keys.

This is currently *expected* to fail in the CIs, until we have adapted all the `EVP_PKEY` functions to deal with provider only keys.

### DIFF
--- a/crypto/evp/p_lib.c
+++ b/crypto/evp/p_lib.c
@@ -535,6 +535,37 @@ int EVP_PKEY_assign(EVP_PKEY *pkey, int type, void *key)
     return (key != NULL);
 }
 
+int EVP_PKEY_deassign(EVP_PKEY *pkey)
+{
+    if (pkey == NULL)
+        return 1;
+
+    /* If it was downgraded, we refuse to make it provider only again */
+    if (pkey->downgraded)
+        return 1;
+
+    /* Make sure it's exported to a provider */
+    {
+        void *keydata = NULL;
+        EVP_KEYMGMT *keymgmt = NULL;
+        size_t i = 0;
+
+        keydata = evp_pkey_export_to_provider(pkey, NULL, &keymgmt, NULL);
+        if (keydata == NULL || keymgmt == NULL)
+            return 0;
+
+        /* Here, we hack ferociously */
+        i = evp_keymgmt_util_find_operation_cache_index(pkey, keymgmt);
+        pkey->keymgmt = keymgmt;
+        pkey->keydata = keydata;
+        pkey->operation_cache[i].keymgmt = NULL;
+        pkey->operation_cache[i].keydata = NULL;
+    }
+
+    evp_pkey_free_legacy(pkey);
+    return 1;
+}
+
 void *EVP_PKEY_get0(const EVP_PKEY *pkey)
 {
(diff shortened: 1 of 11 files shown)


levitte:
Among functions where we have almost no support for provider only keys is `crypto/evp/p_lib.c`, most of it relies solely on `EVP_PKEY_METHOD` and `EVP_PKEY_ASN1_METHOD` structures.

levitte:
I just added `EVP_PKEY_deassign()` in the most evil of places: in the `load_key()` and `load_pubkey()` functions that all apps use.  Tests are expected to go *KABOOM*.

Incidently, I looked into some of them that took me by surprise.  `test_pem` is failing, and that caused me to look more closely at outputs...  and some of them come out differently when printed via the provider serializers.  This happens to be for invalid keys, but that still indicates that something's off, either with assumptions done when exporting from legacy to provider, or when serializing to text.

levitte:
Fixups in (will be updated)

#10557, #10778 #10803, #10805, #10806, #10807, #10808

levitte:
#11126 was a bit of a surprise...  so it turns out that there's code in a few subsystems that are bypassing other central subsystems like EVP and ASN1 and call their backends directly, so weren't at all caught by #10797.  CRMF seems to be one of them, I suspect CMS and a couple of others are similar.  That means that we can expect to find a few "surprises" like that one...

levitte:
All the PRs that were identified here are now merged, so a rebase was in order.

I *still* expects to see failures, but not quite as many.  I also have another experiment in the making that specifically targets the sub-systems `PKCS7` and `CM*`

levitte:
Closing this PR, as it seems to have served its purpose.  Some of the failures are still potentially there, but we do not need to do explicit deassignment of the legacy part of an EVP_PKEY to trigger them any more, as the key generation code is now serving us with such EVP_PKEYs and triggering code that hasn't been adapted for such keys yet.

levitte:
Resurrecting this branch, to test #11422.  Throw everything I have at it, basically...

crypto/evp/p_lib.c
```diff
@@ -535,6 +535,38 @@ int EVP_PKEY_assign(EVP_PKEY *pkey, int type, void *key)
...
+    if (pkey == NULL)
+        return 1;
+
+    /* If it was downgraded, we refuse to make it provider only again */
+    if (pkey->downgraded)
+        return 1;
+
+    /* Make sure it's exported to a provider */
+    {
+        void *keydata = NULL;
```
paulidale, line 549: Does this need to be in its own block?  The function isn't large and the variables can be declared at the top just as easily.

paulidale (commented):
This looks good -- I'll need to run through it again with a less befuddled mind though.

crypto/evp/p_lib.c
levitte: Do I have to?  Also, experimental, I really don't expect this code to end up in the main source, so really, [shrug]

levitte:
Closing, it doesn't serve its purpose any more (or if it should, it needs to be rewritten entirely)

Issue #10237, opened 2019-10-22
CI does not cover configuration option 'enable-trace'


DDvO:
The recent commit f386632 broke `include/openssl/cmp_util.h` when the trace API is enabled.
I've just provided a respective fix: #10236.

This incident indicates that the build checks by the Continuous Integration tools do not cover the case that the trace API is enabled in the configuration, since otherwise the above issue would have been caught right away.

mattcaswell:
We don't include all possible options in travis. There are simply too many of them and travis jobs would take far too long. Instead we have a cron job (the run-checker.sh script) that checks all of these options:

https://github.com/openssl/tools/blob/master/run-checker/run-checker.sh

Interestingly that script does have `enable-ssl-trace` as one of the options it checks - so its not clear why it didn't complain about it.

levitte:
@mattcaswell, `enable-trace` and `enable-ssl-trace` is not the same thing

mattcaswell:
Ah...right....I mixed them up. And now I see that run-checker *is* in fact failing on enable-trace.

DDvO:
> We don't include all possible options in travis. There are simply too many of them and travis jobs would take far too long.

I understand.
Yet one could try combining certain config options that are (linkely) independent into one CI run,
leading to a better overage of options without extra computational overhead.

DDvO:
Currently there is a further build failure that went unnoticed for a while 
just because `enable_trace` is not part of the CI builds: #10731.

mspncp:
> We don't include all possible options in travis. There are simply too many of them and travis jobs would take far too long. Instead we have a cron job (the run-checker.sh script) that checks all of these options:
> 
> https://github.com/openssl/tools/blob/master/run-checker/run-checker.sh

@mattcaswell Is this an internal job only? Or are its results posted somewhere? If not, is it possible to post failures to, e.g., the [openssl-commits](https://mta.openssl.org/pipermail/openssl-commits) list?

mspncp:
Oh, they _are_ posted, see for example

*  [FAILED build of OpenSSL branch master with options -d --strict-warnings enable-trace](https://mta.openssl.org/pipermail/openssl-commits/2019-December/026748.html).

@DDvO you just need to subscribe to the openssl-commits list and create a filter for 
'OpenSSL run-checker'.

mspncp:
Interestingly, the offending commits for issue #10731 are already listed in run-checker's report:

```
Commit log since last time:

[...]

PR #10799, opened 2020-01-10
Add a travis build that compiles with enable-trace

slontis:
This option seems to get broken quite often - so a travis build should
hopefully prevent these creeping into new PRs.

##### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,6 +79,10 @@ matrix:
           compiler: clang
           env: CONFIG_OPTS="--strict-warnings -D__NO_STRING_INLINES no-deprecated" BUILDONLY="yes"
         - os: linux
+          arch: arm64
+          compiler: gcc
+          env: CONFIG_OPTS="--strict-warnings enable-trace" BUILDONLY="yes"
+        - os: linux
           addons:
               apt:
                   packages:


mattcaswell:
I'm somewhat wary about adding a whole new build just for this one option. Travis is slow enough already and it could become the "thin end of the wedge". I did wonder about adding a build that disables as much as possible, to cover lots of options in one go.

levitte:
Our run-checker script has a run with `enable-trace`, and yup, it currently fails.

https://mta.openssl.org/pipermail/openssl-commits/2020-January/026906.html

mspncp:
I think you are repeating the discussion from #10237 here. IMHO the run-checker script is sufficient, the only problem was the we weren't watching the  [openssl-commit] list carefully enough ([which we should improve](https://github.com/openssl/openssl/issues/10237#issuecomment-569917958)).

Since @DDvO has a special interest in the tracing api, I [gave him the advice](https://github.com/openssl/openssl/issues/10237#issuecomment-569917035) to subscribe to [openssl-commit] and create a filter for 'OpenSSL run-checker'. (See also [his reply](https://github.com/openssl/openssl/issues/10237#issuecomment-569919500)).

mspncp:
> Our run-checker script has a run with `enable-trace`, and yup, it currently fails.
> 
> https://mta.openssl.org/pipermail/openssl-commits/2020-January/026906.html

P.S: this should be fixed now by 7671420d9ebd18242f571866c6adca6f8bcb2f90.

slontis:
While the CMP PR's may have broken it a few times, it was also broken multiple times before this work also.

mspncp:
> While the CMP PR's may have broken it a few times

Oh, I didn't mean to blame the CMP guys. (The last break wasn't their fault at all.) I just meant to say that they are the main users of the tracing api, so they are the first ones to suffer from the breakage.

slontis:
I am not blaming anyone either - my only point is it is easy to break things when the code is not being tested in our main CI loop..
It doesnt need to be a seperate build - it could be an option of one of the other existing builds also.

slontis:
Not to worrry - I would have preferred it to fall over before merging..

PR #10802, opened 2020-01-10
Digest function deprecation CHANGES.

paulidale:
Add a changes entry to cover the deprecation of the low level digest functions.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,25 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) All of the low level MD2, MD4, MD5, MDC2, RIPEMD160, SHA1, SHA224, SHA256,
+     SHA384, SHA512 and Whirlpool digest functions have been deprecated.
+     These include:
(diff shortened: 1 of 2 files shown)


paulidale:
The function list will be reformatted once complete.

paulidale:
Move out of WIP, although not all the digests have been deprecated yet.

paulidale:
Once #10791 is merged, this CHANGES entry will be complete and accurate.

levitte:
Shouldn't there be a NEWS entry as well?  Duplicating the first CHANGES sentence should be enough.

paulidale:
Would it be better to say that the "Cipher, digest, MAC and public key low level functions are deprecated" in NEWS?

I'm happy to make each algorithm type individual...

levitte:
If you look at the NEWS items for 1.1.1, you can see how we've done it before.  Following that pattern, it seems like a repeat of that first sentence is more fitting.

paulidale:
NEWS updated.

paulidale:
Merged, thanks.

PR #10804, opened 2020-01-10
CORE: renumber OSSL_FUNC_KEYMGMT macros

levitte:
An amount of upcoming work does this to make space for new functions
in different groups.

### DIFF
--- a/include/openssl/core_numbers.h
+++ b/include/openssl/core_numbers.h
@@ -345,9 +345,9 @@ OSSL_CORE_MAKE_FUNC(int, OP_kdf_set_ctx_params,
  */
 
 /* Key domain parameter creation and destruction */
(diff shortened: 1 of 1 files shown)


kaduk:
It looks like we're adding an extra space before the number on every changed line?

levitte:
Yes, to align them

kaduk (approved):
They were aligned before, too :)
(I assume this means that some to-be-added macro name is going to be longer and force the indentation out.)

I a little bit wonder if we want to jump to multiples of 100 rather than 10, but I forget if we use this as an array index anywhere.

levitte:
No, we don't. They are more like unique tags. See the OSSL_DISPATCH structure

levitte:
Merged.

c8a557357789a8a23fe7b87ec352bd5ea2c1251b CORE: renumber OSSL_FUNC_KEYMGMT macros

PR #10806, opened 2020-01-10
Fix EVP_Digest{Sign,Verify}Final() and EVP_Digest{Sign,Verify}() for provider only keys

levitte:
For provider only keys where the initialization didn't catch, we may
end up crashing because the legacy code path didn't check that it had
support carefully enough.  This only happens if the caller didn't
check if initialization worked or not.

For the one-shot case, it's very simply handling the case where the
key has no legacy implementation an fall back to the standard
init+update+final mechanism.

### DIFF
--- a/crypto/evp/m_sigver.c
+++ b/crypto/evp/m_sigver.c
@@ -302,13 +302,18 @@ int EVP_DigestSignFinal(EVP_MD_CTX *ctx, unsigned char *sigret,
                                                      sigret, siglen, SIZE_MAX);
 
  legacy:
+    if (pctx == NULL || pctx->pmeth == NULL) {
+        ERR_raise(ERR_LIB_EVP, EVP_R_INITIALIZATION_ERROR);
+        return 0;
+    }
+
     if (pctx->pmeth->flags & EVP_PKEY_FLAG_SIGCTX_CUSTOM) {
         if (!sigret)
             return pctx->pmeth->signctx(pctx, sigret, siglen, ctx);
         if (ctx->flags & EVP_MD_CTX_FLAG_FINALISE)
             r = pctx->pmeth->signctx(pctx, sigret, siglen, ctx);
         else {
-            EVP_PKEY_CTX *dctx = EVP_PKEY_CTX_dup(ctx->pctx);
+            EVP_PKEY_CTX *dctx = EVP_PKEY_CTX_dup(pctx);
             if (!dctx)
                 return 0;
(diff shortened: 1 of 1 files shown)


levitte:
This fixes more test crashes for #10797

crypto/evp/m_sigver.c
```diff
@@ -389,13 +394,18 @@ int EVP_DigestVerifyFinal(EVP_MD_CTX *ctx, const unsigned char *sig,
                                                        sig, siglen);
 
  legacy:
-    if (ctx->pctx->pmeth->verifyctx)
+    if (pctx == NULL || pctx->pmeth == NULL) {
+        ERR_raise(ERR_LIB_EVP, EVP_R_INITIALIZATION_ERROR);
+        return 0;
+    }
+
+    if (pctx->pmeth->verifyctx)
```
FdaSilvaYY: nit: test against NULL ?

crypto/evp/m_sigver.c
```diff
@@ -302,13 +302,18 @@ int EVP_DigestSignFinal(EVP_MD_CTX *ctx, unsigned char *sigret,
...
+    }
+
     if (pctx->pmeth->flags & EVP_PKEY_FLAG_SIGCTX_CUSTOM) {
         if (!sigret)
             return pctx->pmeth->signctx(pctx, sigret, siglen, ctx);
         if (ctx->flags & EVP_MD_CTX_FLAG_FINALISE)
             r = pctx->pmeth->signctx(pctx, sigret, siglen, ctx);
         else {
-            EVP_PKEY_CTX *dctx = EVP_PKEY_CTX_dup(ctx->pctx);
+            EVP_PKEY_CTX *dctx = EVP_PKEY_CTX_dup(pctx);
```
FdaSilvaYY: nit: blank line after declaration.

crypto/evp/m_sigver.c
levitte: Fixed

crypto/evp/m_sigver.c
levitte: Done.

levitte:
Travis failure not relevant

levitte:
Final review, anyone?

mattcaswell (approved):

levitte:
Merged.

8f020c3da571fd56c4a1ce16daad553193549a67 Fix EVP_Digest{Sign,Verify}Final() and EVP_Digest{Sign,Verify}() for provider only keys

PR #10807, opened 2020-01-10
EVP: Adapt EVP_PKEY checking, comparing and copying for provider keys

levitte:
This affects the following function, which can now deal with provider
side domain parameters and keys:

- EVP_PKEY_copy_parameters()
- EVP_PKEY_missing_parameters()
- EVP_PKEY_cmp_parameters()
- EVP_PKEY_cmp()

### DIFF
--- a/crypto/evp/keymgmt_lib.c
+++ b/crypto/evp/keymgmt_lib.c
@@ -119,6 +119,7 @@ void *evp_keymgmt_export_to_provider(EVP_PKEY *pk, EVP_KEYMGMT *keymgmt,
 {
     void *provdata = NULL;
     size_t i;
+    int domainparams = -1;
 
     /*
      * If there is an underlying legacy key and it has changed, invalidate
@@ -143,18 +144,24 @@ void *evp_keymgmt_export_to_provider(EVP_PKEY *pk, EVP_KEYMGMT *keymgmt,
     for (i = 0;
          i < OSSL_NELEM(pk->pkeys) && pk->pkeys[i].keymgmt != NULL;
          i++) {
+        if (domainparams == -1)
+            domainparams = pk->pkeys[i].domainparams;
         if (keymgmt == pk->pkeys[i].keymgmt
-            && want_domainparams == pk->pkeys[i].domainparams)
+            && (want_domainparams == -1
+                || want_domainparams == domainparams))
             return pk->pkeys[i].provdata;
     }
 
-    provdata = evp_keymgmt_export_to_provdata(pk, keymgmt, want_domainparams);
+    if (domainparams == -1)
+        domainparams = want_domainparams;
+
+    provdata = evp_keymgmt_export_to_provdata(pk, keymgmt, domainparams);
 
     /* Synchronize the dirty count, but only if we exported successfully */
     if (pk->pkey.ptr != NULL && provdata != NULL)
         pk->dirty_cnt_copy = pk->ameth->dirty_cnt(pk);
 
-    evp_keymgmt_cache_pkey(pk, i, keymgmt, provdata, want_domainparams);
+    evp_keymgmt_cache_pkey(pk, i, keymgmt, provdata, domainparams);
 
     return provdata;
 }
--- a/crypto/evp/keymgmt_lib.c
+++ b/crypto/evp/keymgmt_lib.c
@@ -103,6 +103,10 @@ static void *evp_keymgmt_export_to_provdata(const EVP_PKEY *pk,
          * TODO(3.0) Right now, we assume we have ample space.  We will
          * have to think about a cache aging scheme, though, if |i| indexes
          * outside the array.
+         *
+         * We assign the ossl_assert() result to i because it has an attribute
+         * saying the result can't go unchecked.
+         *
          */
         i = ossl_assert(i < OSSL_NELEM(pk->pkeys));
     }
@@ -253,65 +257,40 @@ int evp_keymgmt_is(const EVP_PKEY *pkey, int domainparams)
  *  0   not same key
  * -1   not same key type
  * -2   unsupported operation
+ *
+ * The code in p_lib.c must have ensured already that both keys have up to
+ * date keymgmt data before they call this function.
  */
 int evp_keymgmt_cmp(const EVP_PKEY *pkey1, const EVP_PKEY *pkey2,
                     int domainparams)
 {
     EVP_KEYMGMT *keymgmt1 = pkey1->pkeys[0].keymgmt;
     EVP_KEYMGMT *keymgmt2 = pkey2->pkeys[0].keymgmt;
-    void *provdata1 = NULL;
-    void *provdata2 = NULL;
-    void *exported_provdata2 = NULL;
-    int rv = 0;                  /* Assume the worst */
+    void *provdata1 = pkey1->pkeys[0].provdata;
+    void *provdata2 = pkey2->pkeys[0].provdata;
 
-    if (keymgmt1 == NULL && keymgmt2 == NULL)
-        return -2;   /* No support without at least one key manager */
+    /*
+     * If one of them isn't initialized yet, they don't have the same key type,
+     * the same way that a legacy key may have the NID type EVP_PKEY_NONE and
+     * the legacy key types do therefore not match.
+     */
+    if (keymgmt1 == NULL || keymgmt2 == NULL)
+        return -1;
 
-    if (keymgmt1 == NULL) {
-        EVP_KEYMGMT *swaptmp = keymgmt1;
-        keymgmt1 = keymgmt2;
-        keymgmt2 = swaptmp;
-    }
-
-    if (keymgmt1 != NULL && keymgmt2 != NULL
-        && EVP_KEYMGMT_number(keymgmt1) != EVP_KEYMGMT_number(keymgmt2))
+    if (EVP_KEYMGMT_number(keymgmt1) != EVP_KEYMGMT_number(keymgmt2))
         return -1;               /* key type mismatch */
-
-    provdata1 = pkey1->pkeys[0].provdata;
-    provdata2 = pkey2->pkeys[0].provdata;
 
     if (pkey1->pkeys[0].domainparams != domainparams
         || pkey2->pkeys[0].domainparams != domainparams)
         return -2;      /* No support for comparing stuff not asked */
 
-    /*
-     * If the two keys belong with different providers or the second key
-     * isn't provided, try to export the second provdata to the first
-     * provider and use that for comparing.
-     */
-    if (keymgmt1 != keymgmt2)
-        provdata2 = exported_provdata2 =
-            evp_keymgmt_export_to_provdata(pkey2, keymgmt1, domainparams);
-
-    /* TODO(3.0) Investigate if this should be -2, i.e unsupported */
-    if (provdata2 == NULL)
-        return -1;               /* Key type mismatch */
-
-    /*
-     * From here on, we will only return 0 or 1, because we know that
-     * the comparing functions are present in the EVP_KEYMGMT structure.
-     */
-    if (provdata1 != NULL && provdata2 != NULL)
-        rv = domainparams
+    /* The provided functions return 0 or 1 */
+    if (provdata1 != NULL && provdata2 == NULL)
+        return domainparams
             ? evp_keymgmt_cmpdomparams(keymgmt1, provdata1, provdata2)
             : evp_keymgmt_cmpkey(keymgmt1, provdata1, provdata2);
 
-    if (domainparams)
-        evp_keymgmt_freedomparams(keymgmt1, exported_provdata2);
-    else
-        evp_keymgmt_freekey(keymgmt1, exported_provdata2);
-
-    return rv;
+    return 0;
 }
 
 int evp_keymgmt_copy(EVP_PKEY *to, const EVP_PKEY *from, int domainparams)
@@ -327,7 +306,16 @@ int evp_keymgmt_copy(EVP_PKEY *to, const EVP_PKEY *from, int domainparams)
         EVP_KEYMGMT_up_ref(keymgmt_to); /* ref++ */
     evp_keymgmt_clear_pkey_cache(to);
 
-    if (keymgmt_to == NULL || keymgmt_to == keymgmt_from) {
+    if (keymgmt_to != NULL && keymgmt_to != keymgmt_from) {
+        /*
+         * We do not take any dirty counter into account, because we're
+         * exporting |from| to |to| unconditionally.
+         * Therefore, we use evp_keymgmt_export_to_provdata() rather than
+         * evp_keymgmt_export_to_provider()
+         */
+        provdata_to =
+            evp_keymgmt_export_to_provdata(from, keymgmt_to, domainparams);
+    } else {
         if (keymgmt_to != NULL)
             EVP_KEYMGMT_free(keymgmt_to); /* ref-- */
         keymgmt_to = keymgmt_from;
@@ -335,9 +323,6 @@ int evp_keymgmt_copy(EVP_PKEY *to, const EVP_PKEY *from, int domainparams)
         provdata_to = domainparams
             ? evp_keymgmt_dupdomparams(keymgmt_from, from->pkeys[0].provdata, 1)
             : evp_keymgmt_dupkey(keymgmt_from, from->pkeys[0].provdata, 1);
-    } else {
-        provdata_to =
-            evp_keymgmt_export_to_provdata(from, keymgmt_to, domainparams);
     }
 
     if (provdata_to == NULL) {
--- a/crypto/evp/keymgmt_lib.c
+++ b/crypto/evp/keymgmt_lib.c
@@ -278,6 +278,24 @@ evp_keymgmt_gettable_domparam_params(const EVP_KEYMGMT *keymgmt)
     return keymgmt->gettable_domparam_params();
 }
 
+int evp_keymgmt_isdomparams(const EVP_KEYMGMT *keymgmt,
+                            const void *domparams)
+{
+    return keymgmt->isdomparams(domparams);
+}
+
+int evp_keymgmt_cmpdomparams(const EVP_KEYMGMT *keymgmt,
+                             const void *domparams1, const void *domparams2)
+{
+    return keymgmt->cmpdomparams(domparams1, domparams2);
+}
+
+void *evp_keymgmt_dupdomparams(const EVP_KEYMGMT *keymgmt,
+                               void *domparams, int do_copy)
+{
+    return keymgmt->dupdomparams(domparams, do_copy);
+}
+
 
 void *evp_keymgmt_importkey(const EVP_KEYMGMT *keymgmt,
                             const OSSL_PARAM params[])
@@ -341,4 +359,20 @@ const OSSL_PARAM *evp_keymgmt_gettable_key_params(const EVP_KEYMGMT *keymgmt)
     if (keymgmt->gettable_key_params == NULL)
         return NULL;
     return keymgmt->gettable_key_params();
+}
+
+int evp_keymgmt_iskey(const EVP_KEYMGMT *keymgmt, const void *key)
+{
+    return keymgmt->iskey(key);
+}
+
+int evp_keymgmt_cmpkey(const EVP_KEYMGMT *keymgmt,
+                       const void *key1, const void *key2)
+{
+    return keymgmt->cmpkey(key1, key2);
+}
+
+void *evp_keymgmt_dupkey(const EVP_KEYMGMT *keymgmt, void *key, int do_copy)
+{
+    return keymgmt->dupkey(key, do_copy);
 }
--- a/crypto/evp/keymgmt_lib.c
+++ b/crypto/evp/keymgmt_lib.c
@@ -32,11 +32,89 @@ static int try_import(const OSSL_PARAM params[], void *arg)
     return data->provdata != NULL;
 }
 
+/*
+ * This only creates a new provider side key, unconditionally.  The returned
+ * pointer must be handled by the caller.
+ */
+static void *evp_keymgmt_export_to_provdata(const EVP_PKEY *pk,
+                                            EVP_KEYMGMT *keymgmt,
+                                            int want_domainparams)
+{
+    void *provdata = NULL;
+    size_t i;
+
+    if (pk->pkey.ptr != NULL) {
+        /* There is a legacy key, try to export that one to the provider */
+
+        /* If the legacy key doesn't have an export function, give up */
+        if (pk->ameth->export_to == NULL)
+            return NULL;
+
+        /* Otherwise, simply use it. */
+        provdata = pk->ameth->export_to(pk, keymgmt, want_domainparams);
+    } else {
+        /*
+         * Here, there is no legacy key, so we look at the already cached
+         * provider keys, and import from the first that supports it
+         * (i.e. use its export function), and export the imported data to
+         * the new provider.
+         */
+
+        /* Setup for the export callback */
+        struct import_data_st import_data;
+
+        import_data.importfn =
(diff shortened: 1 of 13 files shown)


levitte:
There was a small fixup needed, but other than that, at least my personal tests go through.  Final review?

crypto/evp/keymgmt_lib.c
```diff
@@ -118,19 +92,62 @@ void *evp_keymgmt_export_to_provider(EVP_PKEY *pk, EVP_KEYMGMT *keymgmt,
...
                     break;
             }
         }
+
+        /*
+         * TODO(3.0) Right now, we assume we have ample space.  We will
+         * have to think about a cache aging scheme, though, if |i| indexes
+         * outside the array.
+         */
+        i = ossl_assert(i < OSSL_NELEM(pk->pkeys));
```
mattcaswell: What is the purpose of the assignment to `i` here?

crypto/evp/keymgmt_lib.c
```diff
@@ -183,6 +200,124 @@ void *evp_keymgmt_fromdata(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+    void *exported_provdata2 = NULL;
+    int rv = 0;                  /* Assume the worst */
+
+    if (keymgmt1 == NULL && keymgmt2 == NULL)
+        return -2;   /* No support without at least one key manager */
+
+    if (keymgmt1 == NULL) {
+        EVP_KEYMGMT *swaptmp = keymgmt1;
+        keymgmt1 = keymgmt2;
+        keymgmt2 = swaptmp;
```
mattcaswell: Why this elaborate way of assigning NULL to `keymgmt2`?

crypto/evp/keymgmt_lib.c
```diff
@@ -183,6 +200,124 @@ void *evp_keymgmt_fromdata(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+
+    /* TODO(3.0) Investigate if this should be -2, i.e unsupported */
+    if (provdata2 == NULL)
+        return -1;               /* Key type mismatch */
+
+    /*
+     * From here on, we will only return 0 or 1, because we know that
+     * the comparing functions are present in the EVP_KEYMGMT structure.
+     */
+    if (provdata1 != NULL && provdata2 != NULL)
```
mattcaswell: `provdata2` can never be NULL here

crypto/evp/keymgmt_lib.c
```diff
@@ -183,6 +200,124 @@ void *evp_keymgmt_fromdata(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+        if (keymgmt_to != NULL)
+            EVP_KEYMGMT_free(keymgmt_to); /* ref-- */
+        keymgmt_to = keymgmt_from;
+        EVP_KEYMGMT_up_ref(keymgmt_to); /* ref++ */
+        provdata_to = domainparams
+            ? evp_keymgmt_dupdomparams(keymgmt_from, from->pkeys[0].provdata, 1)
+            : evp_keymgmt_dupkey(keymgmt_from, from->pkeys[0].provdata, 1);
+    } else {
+        provdata_to =
+            evp_keymgmt_export_to_provdata(from, keymgmt_to, domainparams);
```
mattcaswell: Same comment as above. Why avoid the dirt count checking?

crypto/evp/keymgmt_lib.c
```diff
@@ -183,6 +200,124 @@ void *evp_keymgmt_fromdata(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+        return -2;      /* No support for comparing stuff not asked */
+
+    /*
+     * If the two keys belong with different providers or the second key
+     * isn't provided, try to export the second provdata to the first
+     * provider and use that for comparing.
+     */
+    if (keymgmt1 != keymgmt2)
+        provdata2 = exported_provdata2 =
+            evp_keymgmt_export_to_provdata(pkey2, keymgmt1, domainparams);
```
mattcaswell: IIUC then by calling evp_keymgmt_export_to_provdata directly you avoid all the dirty count checking. But why? Don't we still need to check the dirty counts? What if the underlying legacy key has been amended?

mattcaswell:
Travis failures also look relevant.

crypto/evp/keymgmt_lib.c
levitte: The purpose here is to ensure that both keys have cached provdata with the same provider (i.e the same keymgmt), regardless of dirty count. 

However, you're right that I didn't check the dirty count of the first key...

crypto/evp/keymgmt_lib.c
levitte: Must be old cruft.  Removing.

crypto/evp/keymgmt_lib.c
levitte: Good question.  Simplifying.

crypto/evp/keymgmt_lib.c
levitte: Hmmmm, now that you get me looking again, I've been over thinking this, haven't I?  A call to `evp_keymgmt_export_to_provider` would suffice just as well...  BUT, I do need to make that call on `pkey1` as well, just to make sure it's up to date too...

crypto/evp/keymgmt_lib.c
levitte: Good point

crypto/evp/keymgmt_lib.c
levitte: Because in this case, we're not updating `from`, we're extracting provdata for `keymgmt_to`, based on `from`.

crypto/evp/keymgmt_lib.c
levitte: Ah, there was a reason after all...  `pkey1` and `pkey2` are constant.
However, we know for a fact that they can only be allocated dynamically, so no matter how much constness we pretend, they aren't really constant, so that might justify bending the rules here.  I'll go with that.
(the flip side of C not having mutables...)

levitte:
After all the excellent remarks from @mattcaswell, I've had the time to rethink this PR, and have refactored it quite a bit.

Also, this is now pending #10837.

crypto/evp/keymgmt_lib.c
levitte: This has been completely remade

crypto/evp/p_lib.c
```diff
@@ -80,66 +81,141 @@ int EVP_PKEY_save_parameters(EVP_PKEY *pkey, int mode)
...
+         * It's fine if the destination hasn't cached anything yet, that
+         * simply makes it a blank page to be filled in with the
+         * evp_keymgmt_copy() call below.
+         * That call will also detect |from->pkeys[0].keymgmt == NULL|,
+         * so we don't bother making an error here.
+         */
+        if (to->pkeys[0].keymgmt != NULL
+            && from->pkeys[0].keymgmt != NULL
+            && (EVP_KEYMGMT_number(to->pkeys[0].keymgmt)
+                != EVP_KEYMGMT_number(from->pkeys[0].keymgmt))) {
```
mattcaswell: So if the keymgmt instances are from different libctxs won't this comparison cause a problem (because the numbers might be different for the same keymgmt in different libctxs)?

crypto/evp/p_lib.c
```diff
@@ -80,66 +81,141 @@ int EVP_PKEY_save_parameters(EVP_PKEY *pkey, int mode)
...
         return 0;
     }
 
-    if (from->ameth && from->ameth->param_copy)
+    /*
+     * If |from| is provided, we ONLY copy the provided bits, and therefore
+     * clear the legacy bits in |to|.
+     */
+    if (is_provided) {
+        evp_pkey_free_legacy(to);
```
mattcaswell: So a copy parameters operation frees any pre-existing key data? Is that the correct semantics?

crypto/evp/p_lib.c
```diff
@@ -80,66 +81,141 @@ int EVP_PKEY_save_parameters(EVP_PKEY *pkey, int mode)
...
+        return -2;
+
+    /*
+     * Ensure on the fly that they both have a keymgmt if at least one of
+     * them does.
+     * To do so, we must break constness to be able to cache keymgmt data,
+     * but since we know that EVP_PKEYs are dynamically allocated, it's ok.
+     */
+    if (!evp_pkey_make_provided((EVP_PKEY *)a, NULL, NULL, NULL, domainparams)
+        || !evp_pkey_make_provided((EVP_PKEY *)b, NULL, NULL, NULL,
```
mattcaswell: Again, I'm concerned about the use of the default libctx here. What if the default libctx doesn't have a provider for this type of key loaded?

doc/man7/provider-keymgmt.pod
```diff
@@ -143,6 +159,22 @@ OP_keymgmt_gettable_domparam_params() should return a constant array
 of descriptor B<OSSL_PARAM>, for parameters that
 OP_keymgmt_get_domparam_params() can handle.
 
+OP_keymgmt_isdomparams() should check if the structure I<domparams> is
+or contains domain parameters, and return 1 if it is, otherwise 0.
```
mattcaswell: "is or contains"...I'm not sure I understand this. Please explain?

providers/implementations/keymgmt/rsa_kmgmt.c.orig
```diff
@@ -0,0 +1,273 @@
+/*
```
mattcaswell: What is this file?

providers/implementations/keymgmt/rsa_kmgmt.c.rej
```diff
@@ -0,0 +1,31 @@
+--- providers/implementations/keymgmt/rsa_kmgmt.c
++++ providers/implementations/keymgmt/rsa_kmgmt.c
```
mattcaswell: What is this file?

providers/implementations/keymgmt/rsa_kmgmt.c.orig
levitte: Oops, stray

providers/implementations/keymgmt/rsa_kmgmt.c.rej
levitte: Oops, stray

crypto/evp/keymgmt_lib.c
levitte: Nope, but if you read the comment a couple of lines up, I think the reason will become clear :wink:

crypto/evp/keymgmt_lib.c
```diff
@@ -217,6 +245,105 @@ void *evp_keymgmt_fromdata(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+
+    if (EVP_KEYMGMT_number(keymgmt1) != EVP_KEYMGMT_number(keymgmt2))
+        return -1;               /* key type mismatch */
+
+    if (pkey1->pkeys[0].domainparams != domainparams
+        || pkey2->pkeys[0].domainparams != domainparams)
+        return -2;      /* No support for comparing stuff not asked */
+
+    /* The provided functions return 0 or 1 */
+    if (provdata1 != NULL && provdata2 == NULL)
```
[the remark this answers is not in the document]
levitte: Good point.

crypto/evp/keymgmt_lib.c
```diff
@@ -217,6 +245,105 @@ void *evp_keymgmt_fromdata(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+{
+    EVP_KEYMGMT *keymgmt_to = to->pkeys[0].keymgmt;
+    EVP_KEYMGMT *keymgmt_from = from->pkeys[0].keymgmt;
+    void *provdata_to = NULL;
+
+    if (keymgmt_from == NULL || from->pkeys[0].domainparams != domainparams)
+        return 0;
+
+    if (keymgmt_to != NULL)
+        EVP_KEYMGMT_up_ref(keymgmt_to); /* ref++ */
```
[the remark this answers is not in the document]
levitte: I'll have another look.  Thanks for the heads up

crypto/evp/keymgmt_meth.c
```diff
@@ -170,7 +202,23 @@ static void *keymgmt_from_dispatch(int name_id,
...
+                 && keymgmt->cmpdomparams != NULL
+                 && keymgmt->dupdomparams != NULL))
+        /* Fundamentals, some things simply *must* be present */
+        || keymgmt->freekey == NULL
+        || (keymgmt->importkey == NULL
+            && keymgmt->genkey == NULL
+            && keymgmt->loadkey == NULL)
+        || keymgmt->iskey == NULL
+        || keymgmt->cmpkey == NULL
+        || keymgmt->dupkey == NULL) {
```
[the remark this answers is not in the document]
levitte: Yes.

crypto/evp/p_lib.c
```diff
@@ -80,66 +81,141 @@ int EVP_PKEY_save_parameters(EVP_PKEY *pkey, int mode)
 
 int EVP_PKEY_copy_parameters(EVP_PKEY *to, const EVP_PKEY *from)
 {
-    if (to->type == EVP_PKEY_NONE) {
-        if (EVP_PKEY_set_type(to, from->type) == 0)
+    int is_provided = 0;
+
+    if (from->ameth == NULL
+        && evp_pkey_make_provided((EVP_PKEY *)from, NULL, NULL, NULL, -1)) {
```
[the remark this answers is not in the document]
levitte: Not "whatever".  To a keymgmt that corresponds to the type of the `from` key, and only if it's not already exported to the keymgmt of that key.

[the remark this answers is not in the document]
levitte: What libctx would you pass to `evp_pkey_make_provided`?

crypto/evp/p_lib.c
levitte: Ouch...  Okie, will give this another thought

crypto/evp/p_lib.c
levitte: > frees any pre-existing key data

Just the legacy bits, and only if `from` doesn't have any legacy bits...  but I need to have another look at the `evp_pkey_make_provided` call...

crypto/evp/keymgmt_lib.c
```diff
@@ -75,11 +52,6 @@ void *evp_keymgmt_export_to_provider(EVP_PKEY *pk, EVP_KEYMGMT *keymgmt,
 
         /* Otherwise, simply use it. */
```
slontis: This comment no longer makes sense..

crypto/evp/keymgmt_lib.c
slontis: Wouldnt if (!ossl_assert())
                   return 0;
be better?

crypto/evp/keymgmt_lib.c
```diff
@@ -119,21 +93,75 @@ void *evp_keymgmt_export_to_provider(EVP_PKEY *pk, EVP_KEYMGMT *keymgmt,
...
-     * outside the array.
+     * See if we have exported to this provider already.
+     * If we have, return immediately.
      */
-    j = ossl_assert(i < OSSL_NELEM(pk->pkeys));
+    for (i = 0;
+         i < OSSL_NELEM(pk->pkeys) && pk->pkeys[i].keymgmt != NULL;
+         i++) {
+        if (domainparams == -1)
+            domainparams = pk->pkeys[i].domainparams;
```
slontis: Do all entries in pk->pkeys[i] have the same domainparams setting?

crypto/evp/keymgmt_lib.c
```diff
@@ -217,6 +245,105 @@ void *evp_keymgmt_fromdata(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+    return pkey->pkeys[0].domainparams
+        ? evp_keymgmt_isdomparams(keymgmt, pkey->pkeys[0].provdata)
+        : evp_keymgmt_iskey(keymgmt, pkey->pkeys[0].provdata);
+}
+
+/*
+ * This function adheres to EVP_PKEY_cmp() / EVP_PKEY_cmp_parameters()
+ * return values, i.e.:
+ *
+ *  1   same key
```
slontis: I really dislike this being called _cmp()
cmp() to me means return 0 if equal.

levitte: The public name will not change, obviously, but feel free to suggest a better name for this function.

crypto/evp/keymgmt_lib.c
levitte: When I designed this cache, I did intend for it to be able to hold a mixture, but that turned out difficult / flawed. That's why you see these indicators on a per cache item basis, and you're right, it can be taken out from there. I think that's a topic for a separate PR, though...

crypto/evp/p_lib.c
```diff
@@ -80,66 +81,141 @@ int EVP_PKEY_save_parameters(EVP_PKEY *pkey, int mode)
...
+
+    /* If unsupported with keymgmt, we try legacy */
+    if (rv != -2)
+        return rv;
+
+    /* Legacy code following.  TODO: remove it */
     if (a->type != b->type)
         return -1;
-    if (a->ameth && a->ameth->param_cmp)
+    if (a->ameth->param_cmp != NULL)
```
slontis: The old code checked for ameth == NULL (Maybe put that back in just in case)

doc/man7/provider-keymgmt.pod
```diff
@@ -185,6 +217,21 @@ OP_keymgmt_gettable_key_params() should return a constant array of
 descriptor B<OSSL_PARAM>, for parameters that
 OP_keymgmt_get_key_params() can handle.
 
+OP_keymgmt_iskey() should check if the structure I<key> is or contains
```
slontis: maybe remove is or

doc/man7/provider-keymgmt.pod
```diff
@@ -185,6 +217,21 @@ OP_keymgmt_gettable_key_params() should return a constant array of
...
+a key pair, and return 1 if it is, otherwise 0.
```
slontis: if it does

providers/implementations/keymgmt/dh_kmgmt.c
```diff
@@ -158,6 +164,44 @@ static int dh_exportdomparams(void *domparams, OSSL_CALLBACK *param_cb,
...
+     */
+    return DH_get0_p(domparams) != NULL && DH_get0_g(domparams) != NULL;
+}
+
+static int dh_cmpdomparams(const void *domparams1, const void *domparams2)
+{
+    const BIGNUM *q1, *q2;
+
+    if (BN_cmp(DH_get0_p(domparams1), DH_get0_p(domparams2)) != 0
+        && BN_cmp(DH_get0_g(domparams1), DH_get0_g(domparams2)) != 0)
```
slontis: Should this be ||

providers/implementations/keymgmt/dh_kmgmt.c
```diff
@@ -158,6 +164,44 @@ static int dh_exportdomparams(void *domparams, OSSL_CALLBACK *param_cb,
...
+        return 0;
+    /* Support DHX, compare Q if available */
+    q1 = DH_get0_q(domparams1);
+    q2 = DH_get0_q(domparams2);
+    if (q1 == NULL && q2 == NULL)
```
slontis: Not sure why you need these null checks : BN_cmp returns 0 if both NULL or if they are equal otherwise it returns -1 or 1 dependent on a < b.. So the cmp()==0 should be sufficient.

providers/implementations/keymgmt/dh_kmgmt.c
```diff
@@ -222,6 +266,33 @@ static int dh_get_key_params(void *key, OSSL_PARAM params[])
     return dh_get_dpk_params(key, params);
 }
 
+static int dh_iskey(const void *key)
+{
+    return DH_get0_pub_key(key) != NULL;
```
slontis: If we can set keys up via setparams would it be possible to only have the private key bit?

providers/implementations/keymgmt/dh_kmgmt.c
```diff
@@ -222,6 +266,33 @@ static int dh_get_key_params(void *key, OSSL_PARAM params[])
...
+}
+
+static int dh_cmpkey(const void *key1, const void *key2)
+{
+    if (BN_cmp(DH_get0_pub_key(key1), DH_get0_pub_key(key2)) != 0)
```
slontis: hmm this is an interesting check... Are we only interested in the public key comparison here?

providers/implementations/keymgmt/dsa_kmgmt.c
```diff
@@ -164,6 +170,55 @@ static int dsa_exportdomparams(void *domparams,
...
+     */
+    return DSA_get0_p(domparams) != NULL
+        && DSA_get0_q(domparams) != NULL
+        && DSA_get0_g(domparams) != NULL;
+}
+
+static int dsa_cmpdomparams(const void *domparams1, const void *domparams2)
+{
+    if (BN_cmp(DSA_get0_p(domparams1), DSA_get0_p(domparams2)) != 0
+        && BN_cmp(DSA_get0_q(domparams1), DSA_get0_q(domparams2)) != 0
```
slontis: ||'s

slontis: OR...
return BN_cmp() == 0
      && BN_cmp() == 0...

crypto/evp/keymgmt_lib.c
slontis: equals() ?

providers/implementations/keymgmt/dh_kmgmt.c
levitte: I hadn't looked at `BN_cmp`'s behavior. Thanks for the heads up

providers/implementations/keymgmt/dh_kmgmt.c
slontis: OR 
return == 0
   || == 0

providers/implementations/keymgmt/dh_kmgmt.c
levitte: The general assumption with openssl is that a valid key pair *always* has the public half. A lot will fall apart if that doesn't hold true.

providers/implementations/keymgmt/dh_kmgmt.c
levitte: I basically copied the code from dh_ameth.c. Interestingly enough, the legacy function pointer is called `pub_cmp`, so it's a bit more explicit. I was thinking that it would be up to the provider to decide exactly how detailed the comparison should be. In this implementation, I chose to do like we did before. For some implementations where the private half is hidden (typically HSMs), this would be the only possible comparison. 

With the assumption that the private half actually corresponds to the public part, this comparison is enough. I suppose that key validation will ensure this validity?

crypto/evp/p_lib.c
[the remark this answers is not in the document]
mattcaswell: I'm not sure, but this is problematic. If you're trying to copy parameters from one key to another for a key type that does not exist in the default provider (but does in some other libctx) then its not going to work.

levitte:
I've realized that this needs reworking on a deeper level than I thought.  Most of all, functions like `EVP_PKEY_copy_parameters` can't be implemented as a duplicating function, and it needs KEYMGMT support to import domain parameters into *an already existing key*.  The consequence is that the KEYMGMT libcrypto \<-\> provider interface needs to be somewhat redesigned.

Thank you all for the feedback here, it will absolutely work as valuable input for the redesign.

PR coming up!

levitte:
PR took a little longer than I expected.  However, I've raised an issue, see #10979

levitte:
#11006 is the new KEYMGMT design.  I'll work that one into this one.

levitte:
Killing this in favor of #11025

PR #10808, opened 2020-01-10
EVP: Adapt EVP_PKEY Seal and Open for provider keys

levitte:
This affects the following function, which can now deal with provider
side keys:

- EVP_SealInit()
- EVP_OpenInit()

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,7 +9,10 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
-  *) Deprecated EVP_PKEY_decrypt_old() and EVP_PKEY_encrypt_old()
+  *) Deprecated EVP_PKEY_decrypt_old(), please use EVP_PKEY_decrypt_init()
+     and EVP_PKEY_decrypt() instead.
+     Deprecated EVP_PKEY_encrypt_old(), please use EVP_PKEY_encrypt_init()
+     and EVP_PKEY_encrypt() instead.
      [Richard Levitte]
 
   *) Enhanced the documentation of EVP_PKEY_size(), EVP_PKEY_bits()
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,9 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) Deprecated EVP_PKEY_decrypt_old() and EVP_PKEY_encrypt_old()
+     [Richard Levitte]
+
   *) Enhanced the documentation of EVP_PKEY_size(), EVP_PKEY_bits()
      and EVP_PKEY_security_bits().  Especially EVP_PKEY_size() needed
      a new formulation to include all the things it can be used for,
--- a/crypto/evp/p_open.c
+++ b/crypto/evp/p_open.c
@@ -23,41 +23,44 @@ int EVP_OpenInit(EVP_CIPHER_CTX *ctx, const EVP_CIPHER *type,
                  EVP_PKEY *priv)
 {
     unsigned char *key = NULL;
-    int i, size = 0, ret = 0;
+    size_t keylen = 0;
+    int ret = 0;
+    EVP_PKEY_CTX *pctx = NULL;
 
     if (type) {
         EVP_CIPHER_CTX_reset(ctx);
         if (!EVP_DecryptInit_ex(ctx, type, NULL, NULL, NULL))
-            return 0;
+            goto err;
     }
 
     if (priv == NULL)
         return 1;
 
-    if (EVP_PKEY_id(priv) != EVP_PKEY_RSA) {
-        EVPerr(EVP_F_EVP_OPENINIT, EVP_R_PUBLIC_KEY_NOT_RSA);
-        goto err;
-    }
-
-    size = EVP_PKEY_size(priv);
-    key = OPENSSL_malloc(size);
-    if (key == NULL) {
-        /* ERROR */
+    if ((pctx = EVP_PKEY_CTX_new(priv, NULL)) == NULL) {
         EVPerr(EVP_F_EVP_OPENINIT, ERR_R_MALLOC_FAILURE);
         goto err;
     }
 
-    i = EVP_PKEY_decrypt_old(key, ek, ekl, priv);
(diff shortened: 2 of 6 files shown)


levitte:
Ping

paulidale (dismissed):
LGTM

crypto/evp/p_seal.c
```diff
@@ -41,13 +42,20 @@ int EVP_SealInit(EVP_CIPHER_CTX *ctx, const EVP_CIPHER *type,
         goto err;
 
     for (i = 0; i < npubk; i++) {
-        ekl[i] =
-            EVP_PKEY_encrypt_old(ek[i], key, EVP_CIPHER_CTX_key_length(ctx),
-                                 pubk[i]);
-        if (ekl[i] <= 0) {
-            rv = -1;
+        size_t keylen = ekl[i];
```
mattcaswell: Previously we wrote to ekl unconditionally (without reading it first). This seems to be reading potentially unitialised memory? We also seem to have lost the call to `EVP_CIPHER_CTX_key_length()`

include/openssl/evp.h
```diff
@@ -1089,12 +1089,13 @@ void EVP_MAC_names_do_all(const EVP_MAC *mac,
...
-                         const unsigned char *enc_key, int enc_key_len,
-                         EVP_PKEY *private_key);
-int EVP_PKEY_encrypt_old(unsigned char *enc_key,
-                         const unsigned char *key, int key_len,
-                         EVP_PKEY *pub_key);
+DEPRECATEDIN_3_0(int EVP_PKEY_decrypt_old(unsigned char *dec_key,
+                                          const unsigned char *enc_key,
+                                          int enc_key_len,
+                                          EVP_PKEY *private_key))
+DEPRECATEDIN_3_0(int EVP_PKEY_encrypt_old(unsigned char *enc_key,
```
mattcaswell: These deprecations need a CHANGES entry

mattcaswell (changes requested):

crypto/evp/p_seal.c
levitte: Right.  Fixed.

include/openssl/evp.h
levitte: Okie

levitte:
Ping

mattcaswell (changes requested):

crypto/evp/p_open.c
```diff
@@ -23,41 +23,44 @@ int EVP_OpenInit(EVP_CIPHER_CTX *ctx, const EVP_CIPHER *type,
...
         EVP_CIPHER_CTX_reset(ctx);
         if (!EVP_DecryptInit_ex(ctx, type, NULL, NULL, NULL))
-            return 0;
+            goto err;
     }
 
     if (priv == NULL)
         return 1;
 
-    if (EVP_PKEY_id(priv) != EVP_PKEY_RSA) {
```
slontis: The docs for EVP_SealInit() indicate the following..
The public key must be RSA because it is the only OpenSSL public key algorithm that supports key transport.
Maybe this check for the type is still needed?

levitte: Ah, I think that documentation needs to be changed. Why limit it to RSA when someone might come up with something new that supports key transport, and writes a provider for it at any given moment in time?

crypto/evp/p_open.c
```diff
@@ -23,41 +23,44 @@ int EVP_OpenInit(EVP_CIPHER_CTX *ctx, const EVP_CIPHER *type,
...
+        ERR_raise(ERR_LIB_EVP, ERR_R_MALLOC_FAILURE);
         goto err;
     }
 
-    size = EVP_PKEY_size(priv);
-    key = OPENSSL_malloc(size);
-    if (key == NULL) {
-        /* ERROR */
-        EVPerr(EVP_F_EVP_OPENINIT, ERR_R_MALLOC_FAILURE);
+    if (!EVP_PKEY_decrypt_init(pctx)
```
slontis: On Failure EVP_PKEY_decrypt_init() and EVP_PKEY_decrypt() can return < 0
So replace all !X with X <= 0

crypto/evp/p_seal.c
```diff
@@ -41,13 +42,19 @@ int EVP_SealInit(EVP_CIPHER_CTX *ctx, const EVP_CIPHER *type,
...
-            rv = -1;
+        size_t keylen = EVP_CIPHER_CTX_key_length(ctx);
+        EVP_PKEY_CTX *pctx = NULL;
+
+        if ((pctx = EVP_PKEY_CTX_new(pubk[i], NULL)) == NULL) {
+            ERR_raise(ERR_LIB_EVP, ERR_R_MALLOC_FAILURE);
             goto err;
         }
+
+        if (!EVP_PKEY_encrypt_init(pctx)
```
slontis: Same here with with return result <= 0 for the encrypt_init() and encrypt()

crypto/evp/p_open.c
slontis: (The EVP_DecryptInit_ex() variants are ok how they are)

crypto/evp/p_open.c
slontis: Like ECC key transport?

levitte: Possibly? Doesn't really matter, we should just avoid the of fairly arbitrary limit, especially when libcrypto is becoming a framework.

CHANGES
```diff
@@ -9,6 +9,9 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) Deprecated EVP_PKEY_decrypt_old() and EVP_PKEY_encrypt_old()
+     [Richard Levitte]
```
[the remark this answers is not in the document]
levitte: Sure

crypto/evp/p_open.c
levitte: Thanks, I think I've changed them all

levitte:
Ping

slontis (approved):

levitte:
Merged.

9420b403b72ecd74f55804f494346c926aa609c9 EVP: Adapt EVP_PKEY Seal and Open for provider keys

PR #10813, opened 2020-01-11
Move the definition of OPENSSL_BUILDING_OPENSSL

levitte:
It was placed in a spot that works as fallback, and would therefore
not at all for config targets that define library macros of their own.

Furthermore, this should be attached specifically to our public
libraries, and that's done more precisely in our build.info files.

We ddo not attach it to anything else, so that our programs and engine
become and therefore are closer to what our users are exposed to...
in the spirit of eating our own dog food.

### DIFF
--- a/Configurations/00-base-templates.conf
+++ b/Configurations/00-base-templates.conf
@@ -10,7 +10,7 @@ my %targets=(
 	includes	=> [],
 	lib_cflags	=> "",
 	lib_cppflags	=> "",
-	lib_defines	=> [ 'OPENSSL_BUILDING_OPENSSL' ],
+	lib_defines	=> [],
 	thread_scheme	=> "(unknown)", # Assume we don't know
 	thread_defines	=> [],
 
--- a/build.info
+++ b/build.info
@@ -7,6 +7,11 @@ INCLUDE[libcrypto]=. include
 INCLUDE[libssl]=. include
 DEPEND[libssl]=libcrypto
 
+# Marker to show that we're building the OpenSSL libraries.  This is usefull
+# for source that has multiple usage (specifically, header files).
+DEFINE[libcrypto]=OPENSSL_BUILDING_OPENSSL
+DEFINE[libssl]=OPENSSL_BUILDING_OPENSSL
+
 # Empty DEPEND "indices" means the dependencies are expected to be built
 # unconditionally before anything else.
 DEPEND[]=include/openssl/configuration.h include/openssl/opensslv.h \


levitte:
@richsalz, please try this out.

richsalz:
The apps also needs this.

richsalz:
And the test programs.

levitte:
Why?

richsalz:
>Why?

I am kind of gobsmacked to see this being asked, now.  I started with #10666 which has a rationale in it.  Please go read that.

I did two PR's #10724 and #10669 that assume that *all* of OpenSSL is built the  same way, as described in #10666.  Your first attempt to replace it, #10753, also added the define for building all of OpenSSL.  It didn't work, in spite of your snarky comment otherwise:  yes, I know what "make -s" does.

And now this PR changes the purpose of the define with some rationale about being "closer to customers."  Baloney.  We are building OpenSSL here.  Again, #10666 explains things.

richsalz:
BTW, this PR does nothing useful and if left as-is should be closed.  Defining it for ALL files compiled wuld change my opinion.

levitte:
I wholeheartedly disagree with you. There *should* be something that differentiates between building the libraries and building the stuff that uses the libraries. And making sure we eat our own dog food is not a bad thing.

levitte:
> And now this PR changes the purpose of the define

Actually, not really...  The `lib_defines` affects *libraries only*, so the intention I asserted in #10753 hasn't really changed that much.

mspncp:
It appears to me that this discussion is starting to get too emotional. Maybe it would be a good idea to take a deep breath and ask the other @openssl/committers for independent opinions?

levitte:
@richsalz, apologies for the snarky comment, it was obviously presumptuous.

That being said, the lack of `OPENSSL_BUILDING_OPENSSL` should not cause failure when including a public header file.

richsalz:
The two PR's *assumed* the define was set when compiling OpenSSL source. If the project goes through with this, as-is, then I will close my two PR's that deprecated stuff.  That's okay, the project can do what it wants, but I am disappointed.

As for the utility of this PR as it currently is, there is nothing that uses it.  Do you have any plans in mind, @levitte ?

levitte:
My entire purpose was to support what you needed, but also in a way that doesn't compromise our public header files without us noticing. If you don't want to carry on, then I agree that this PR is useless.

levitte:
Do note, BTW, that I do see value in forcing us to define stack types where we actually use them, rather than in public header files, so it's not that I'm trying to sabotage your effort (that's not my way). I hope that saying this can serve as encouragement.

(I might think that your effort isn't rad enough, but that's a different matter. It's still a step in the right direction, it just needs to be done carefully)

richsalz:
Why do you say our header files would be compromised?

Why do you say my effort isn't "rad" enough?  If anything, it's more radical than what this PR allows because it forces test, apps, engine to use deprecated things.

levitte:
Have you read my comment in #10669? The failure I see indicates, and I specified why, that the public headers as changed in that PR can't be used without `OPENSSL_BUILDING_OPENSSL` being defined. The consequence would then be that *any external program must define* `OPENSSL_BUILDING_OPENSSL` prior to including that header file. That's a compromise of those header files.

richsalz:
No, that's not the intent, which is why it's a WIP and why I said the test program works for me as-is; see https://github.com/openssl/openssl/pull/10669#issuecomment-572165728

levitte:
I have given you the necessary tools in #10669. With that, I don't see the sense in sticking with this PR, so closing it.

PR #10815, opened 2020-01-12
Change returned -2 to 0 in EVP_Digest{Sign,Verify}Init()

levitte:
The returned -2 was to mark when these operations are unsupported.
However, that breaks away from the previous API and expectations, and
there's not enough justification for that not being zero.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,15 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) Corrected the documentation of the return values from the EVP_DigestSign*
+     set of functions.  The documentation mentioned negative values for some
+     errors, but this was never the case, so the mention of negative values
+     was removed.
+
+     Code that followed the documentation and thereby check with something
+     like 'EVP_DigestSignInit(...) <= 0' will continue to work undisturbed.
+     [Richard Levitte]
+
   *) All of the low level Blowfish functions have been deprecated including:
      BF_set_key, BF_encrypt, BF_decrypt, BF_ecb_encrypt, BF_cbc_encrypt,
      BF_cfb64_encrypt, BF_ofb64_encrypt, and BF_options.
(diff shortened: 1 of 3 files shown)


levitte:
@romen, thanks for your input in #10805...  here's the other way to deal with this issue :smile:

romen (approved):
LGTM !

What could we do to put this on the path for revision for a release after 3.0?

I share the wish of uniformity across the board for the family of `EVP_<someop>_init()` and it seems reasonable to dedicate a return value (`-2` is somewhat our default for this in many places) for the "op not supported with this obscure key you gave me".

romen:
I believe that excluding the change to `crypto/evp/m_sigver.c` this PR should be backported to 1.1.1 as well.

Do we need a separate PR for that in this case?

levitte:
> Do we need a separate PR for that in this case?

I don't think so.  You've essentially approved a selective backport, and it should be clean.

t8m (approved):

kaduk:
> LGTM !
> 
> What could we do to put this on the path for revision for a release after 3.0?
> 
> I share the wish of uniformity across the board for the family of `EVP_<someop>_init()` and it seems reasonable to dedicate a return value (`-2` is somewhat our default for this in many places) for the "op not supported with this obscure key you gave me".

I'd really like to be able to do this as well, but having reflected about the broader topic a lot over the past several months, I'm coming around to think that the only reliable way to effect this kind of change is to define completely new APIs with the "proper semantics", and go through the full (5-year) deprecation cycle for the old APIs.  The risk of silently breaking applications seems to outweigh the marginal consistency gains, otherwise.

levitte:
Merged.

76123661a1db136b9ef368dc296a628818e7a4cc Change returned -2 to 0 in EVP_Digest{Sign,Verify}Init()

The backport to 1.1.1 has a different commit message, so I'm submitting a separate PR for it.

levitte:
See #10847 for the backport

PR #10794, opened 2020-01-09
Avoid leak in error path of asn1_parse2

kinichiro:

PR #8888, opened 2019-05-07
Coverity fixes.

paulidale:
Fix Coverity issues: 1444950, 1444951, 1444952, 1444953, 1444954, 1444955, 1444956, 1444957, 1444958, 1444959, 1444960, 1444961 and 1444963

- [x] tests are added or updated

PR #10816, opened 2020-01-12
Avoid leak in error path of PKCS5_PBE_keyivgen

kinichiro:
CLA: trivial

### DIFF
--- a/crypto/evp/p5_crpt.c
+++ b/crypto/evp/p5_crpt.c
@@ -51,11 +51,13 @@ int PKCS5_PBE_keyivgen(EVP_CIPHER_CTX *cctx, const char *pass, int passlen,
     ivl = EVP_CIPHER_iv_length(cipher);
     if (ivl < 0 || ivl > 16) {
         EVPerr(EVP_F_PKCS5_PBE_KEYIVGEN, EVP_R_INVALID_IV_LENGTH);
+        PBEPARAM_free(pbe);
         return 0;
     }
     kl = EVP_CIPHER_key_length(cipher);
     if (kl < 0 || kl > (int)sizeof(md_tmp)) {
         EVPerr(EVP_F_PKCS5_PBE_KEYIVGEN, EVP_R_INVALID_KEY_LENGTH);
+        PBEPARAM_free(pbe);
         return 0;
     }
 
@@ -106,6 +108,7 @@ int PKCS5_PBE_keyivgen(EVP_CIPHER_CTX *cctx, const char *pass, int passlen,
     OPENSSL_cleanse(iv, EVP_MAX_IV_LENGTH);
     rv = 1;
  err:
+    PBEPARAM_free(pbe);
     EVP_MD_CTX_free(ctx);
     return rv;
 }
--- a/crypto/evp/p5_crpt.c
+++ b/crypto/evp/p5_crpt.c
@@ -86,6 +86,7 @@ int PKCS5_PBE_keyivgen(EVP_CIPHER_CTX *cctx, const char *pass, int passlen,
     if (!EVP_DigestUpdate(ctx, salt, saltlen))
         goto err;
     PBEPARAM_free(pbe);
+    pbe = NULL;
     if (!EVP_DigestFinal_ex(ctx, md_tmp, NULL))
         goto err;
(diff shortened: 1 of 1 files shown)


mspncp:
Thank you @kinichiro for your contribution. Please note my reply https://github.com/openssl/openssl/pull/10794#issuecomment-573399603.

t8m:
I think this is acceptable under the CLA:trivial rules.

beldmit:
I agree with @t8m about triviality.

crypto/evp/p5_crpt.c
```diff
@@ -106,6 +108,7 @@ int PKCS5_PBE_keyivgen(EVP_CIPHER_CTX *cctx, const char *pass, int passlen,
     OPENSSL_cleanse(iv, EVP_MAX_IV_LENGTH);
     rv = 1;
  err:
+    PBEPARAM_free(pbe);
```
mspncp: In the successful case, the `pbe` pointer needs to be cleared after being freed in [line 86/88](https://github.com/openssl/openssl/pull/10816/files#diff-bd251e31a7360ecf59298a4344896afbR88), otherwise you'll get a double free error.

mspncp: Actually, it would be more consistent and less error prone, if you would remove the free in line 86/88 and change the `return 0` in line 91/93 to a `goto error` instead.

kinichiro: Sure, `pbe` should be cleared after PBEPARAM_free in success path. I added another commit.
If just changing `goto error;` to `return 0;`, I thought that `ctx` would leak in error path.

mspncp: What I was trying to say is that IMHO it is error prone to have a mixture of `return 0` (before the first successful allocation) and  `goto error` (after the first successful allocation) error treatment inside the same function. Because this inconsistency facilitates memory leaks when adding more allocations at the beginning of the function. I'd rather 

- initialize all pointers to allocated data with NULL at the beginning of the function,
- always `goto error` on failure,
- and free all allocated data after the `error:` label.

But unfortunately, this consistent pattern is not common in the OpenSSL codebase.

Your fix is fine as it is now, no need to change anything anymore.

mspncp:
For the record: I agree it's a trivial change.

t8m:
Merged to master as adc9086.
Cherry-pick to 1.1.1 requires dropping two hunks or picking the a05bf83c7964 first.

mspncp:
Cherry-picks as @t8m suggested apply cleanly:
commit 6527714c25a4266504e323395fd9ffd1bccb2041 (HEAD -> OpenSSL_1_1_1-stable, msp/OpenSSL_1_1_1-stable)
Author: kinichiro <kinichiro.inoguchi@gmail.com>
Date:   Sun Jan 12 17:35:39 2020 +0900

    Avoid leak in error path of PKCS5_PBE_keyivgen
    
    CLA: trivial
    
    Reviewed-by: Matt Caswell <matt@openssl.org>
    Reviewed-by: Matthias St. Pierre <Matthias.St.Pierre@ncp-e.com>
    (Merged from https://github.com/openssl/openssl/pull/10816)
    
    (cherry picked from commit adc9086beb21a91ca59aaf0c619b38b82c223f9b)

commit 3948408f2a6370556d833dc4947ffaaca82ebb78
Author: Pauli <paul.dale@oracle.com>
Date:   Tue May 7 10:42:58 2019 +1000

    Coverity CID 1444960: Error handling issues
    
    Reviewed-by: Matt Caswell <matt@openssl.org>
    (Merged from https://github.com/openssl/openssl/pull/8888)
    
    (cherry picked from commit a05bf83c7964bb3928b323fe356b9f70f105036d)


Do I get your approval @mattcaswell @paulidale for pushing both?

mattcaswell:
Approved for 1.1.1

paulidale:
Yes, approved for 1.1.1.

mspncp:
Pushed the cherry-picked commits from https://github.com/openssl/openssl/pull/10816#issuecomment-582985581 without further changes. Thank you.

6527714c25 Avoid leak in error path of PKCS5_PBE_keyivgen
3948408f2a Coverity CID 1444960: Error handling issues

PR #10819, opened 2020-01-13
 Deprecate the low level IDEA functions.

paulidale:
Use of the low level IDEA functions has been informally discouraged for a
long time. We now formally deprecate them.

Applications should instead use the EVP APIs, e.g. EVP_EncryptInit_ex,
EVP_EncryptUpdate, EVP_EncryptFinal_ex, and the equivalently named decrypt
functions.


include/openssl/idea.h
```diff
@@ -19,52 +19,60 @@
...
 
-# define IDEA_ENCRYPT    1
-# define IDEA_DECRYPT    0
+#  define IDEA_ENCRYPT    1
+#  define IDEA_DECRYPT    0
 
-# define IDEA_BLOCK      8
-# define IDEA_KEY_LENGTH 16
+#  define IDEA_BLOCK      8
+#  define IDEA_KEY_LENGTH 16
```
mattcaswell: Where did we get to with deprecating/not deprecating defines like this?

mattcaswell:
> We'll need a CHANGES entry for cipher functions too

And now I see that you added one!

paulidale:
Yes, CHANGES entries for ciphers are going into #10820.
I figure that one bulk item will be easier to read.

include/openssl/idea.h
paulidale: Digest sizes have been kept public.  AES exposes `AES_BLOCK_SIZE` so I guess the rest of the ciphers should retain that and the key length.

paulidale: I've exposed the block size and key length regardless of the deprecation status.  The encrypt and decrypt macros don't seem use to always expose.

paulidale:
Rebased and conflict resolved.

paulidale:
Rebased and conflict removed.

paulidale:
Ping reviewer.

test/build.info
```diff
@@ -553,6 +549,10 @@ IF[{- !$disabled{tests} -}]
     INCLUDE[tls13encryptiontest]=.. ../include ../apps/include
     DEPEND[tls13encryptiontest]=../libcrypto ../libssl.a libtestutil.a
 
+    SOURCE[ideatest]=ideatest.c
+    INCLUDE[ideatest]=../include ../apps/include
+    DEPEND[ideatest]=../libcrypto.a libtestutil.a
```
mattcaswell: If we're going to treat this as an "internal" test - then shouldn't this be moved from the `PROGRAMS{noinst}` on line 30, to the one on line 503 for consistency with all the other internal tests?

paulidale: Good point, this will impact the other cipher deprecation PRs.

It either has to be made internal or removed -- I'm fine with either so long as we're consistent about it.  If it is removed, there won't be a direct low level test (but the APIs are deprecated).

paulidale: Done.

paulidale:
Required a rebase to fix merge conflicts in comments.
Thanks for the reviewing.

PR #10821, opened 2020-01-13
md2: test doesn't require direct low level access

paulidale:
Low level access isn't required beyond the block size which remains available.


levitte:
Do we test md2 with evp_test? In that case, I see no use for this test program and would simply throw it away.

romen:
Should we though have in place something that is testing that the _legacy_ low level API is still working? (I see this is not the case anymore for this PR, but I am just wondering about the risk of losing coverage on the ancient API)

paulidale:
We don't seem to have any other MD2 tests.

This test is unchanged by the deprecation work -- i.e. the low level APIs were not tested previously.

levitte:
My suggestion would be to copy the test vectors to the appropriate evp_test input file, and refractor this one into testing the low-level API. Otherwise, this program is just redundant cruft.

paulidale:
I'm more inclined to copy the test cases to EVP and just remove this one.

levitte:
Sure. After all, that should still test the low level stuff, just via EVP

paulidale:
The low level calls will be as well tested as they were :)

I've pushed the changes.

levitte:
Approval is, of course, as long as CIs agree

paulidale:
CI failure was relevant.  Forgot to delete `05-test_md2.t`.  Scarily, most CIs passed with this omission.

paulidale:
@levitte this will need a new approval -- the only change is the file deletion.

levitte:
> Scarily, most CIs passed with this omission.

MD2 is disabled by default, isn't it? So of course, only the jobs that enable it explicitly will get failures...

paulidale:
Merged to master with one end of line space removed.

PR #10828, opened 2020-01-13
Add AES consttime code for no-asm configurations

bernd-edlinger:
This adds optional constant time support for AES
when building openssl for no-asm.

Enable with: ./config no-asm -DOPENSSL_AES_CONST_TIME

##### Checklist

This is about 12 times slower than table driven no-asm AES implementation
but constant time.


richsalz:
HOLY COW!  This is a *major* improvement.  Nice work.  Should have a NEWS or CHANGES entry (I can never figure out which is which).  Should be considered for backport as a security fix.

mattcaswell:
Nice!

crypto/aes/aes_core.c
```diff
@@ -50,7 +50,578 @@
...
+
+static void ShiftRows(u64 *state)
+{
+    unsigned char s[4];
+    unsigned char *s0;
+    int r;
+
+    s0 = (unsigned char *)state;
+    for (r = 0; r < 4; r++)
+    {
```
bernd-edlinger: I'll fix the formatting here.

bernd-edlinger:
Should I add something to the docs, and where?

beldmit:
> Enable with: ./config no-asm -DOPENSSL_AES_CONST_TIME

Shouldn't we enable it by default despite slow down?

crypto/aes/aes_local.h
```diff
@@ -24,6 +24,7 @@
 #  define PUTU32(ct, st) { (ct)[0] = (u8)((st) >> 24); (ct)[1] = (u8)((st) >> 16); (ct)[2] = (u8)((st) >>  8); (ct)[3] = (u8)(st); }
 # endif
 
+typedef unsigned long long u64;
```
paulidale: Better to use the `uint64_t` type which OpenSSL ensures is defined properly in `e_os2.h`.  This could conflict with the many _uLL_ constants.

crypto/aes/aes_core.c
```diff
@@ -50,7 +50,577 @@
...
+typedef union
+{
+    unsigned char b[8];
+    u32 w[2];
+    u64 d;
+} uni;
+
+static void XtimeWord(u32 *w)
+{
+    u32 a, b;
```
paulidale: NIT: blank line after declaration in many places.

crypto/aes/aes_core.c
```diff
@@ -50,7 +50,577 @@
...
+    b &= 0x1B1B1B1B1B1B1B1BuLL;
+    b ^= a << 1;
+    *w = b;
+}
+
+static void SubWord(u32 *w)
+{
+    u32 x, y, a1, a2, a3, a4, a5, a6;
+    x = *w;
+    y = ((x & 0xFEFEFEFEu) >> 1) | ((x & 0x01010101u) << 7);
```
paulidale: Would it be worthwhile making this into a macro with a descriptive name.
I find code like this hard to read -- my eyes head straight for the big constants.
```
#define bytewise_rotate_right_one_bit(n) ((n & 0xFEFEFEFEu) >> 1) | ((n & 0x01010101u) << 7)
```

bernd-edlinger: better shut your eyes here.

crypto/aes/aes_local.h
bernd-edlinger: In the next line there is a u32.
Do you see a problem with the LL constants?

paulidale: I did wonder about them.  However, we've got them elsewhere in the code (e.g. curve448) so I'm guessing that we accept them now.  This would also imply that _long long_ is okay.

davidben:
Exciting! What's the S-box implementation strategy? It's probably worth writing down in a comment. Also I'm curious. :-)

Did you consider bitslicing? That can help smooth over the performance hit on parallel modes like AES-GCM. (I recently finished a constant-time fallback AES myself. I used bitslicing with [this circuit](https://eprint.iacr.org/2009/191.pdf). It seems faster than this PR even on single-block operations, but it's also just as likely that I messed up the comparison, so take that with a grain of salt.)

crypto/aes/aes_local.h
```diff
@@ -24,6 +24,7 @@
...
 # ifdef AES_LONG
```
bernd-edlinger: as a side note, do we still use AES_LONG?

bernd-edlinger:
I use an isomorphism between GF(2^8) and GF(2^2^2^2)
and use for decryption the InvCipher not EqInvCipher from [FIPS-197](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.197.pdf), so I do not need
two different key expansions.
I do not use a normal base as [Canright](https://calhoun.nps.edu/bitstream/handle/10945/791/NPS-MA-04-001.pdf?sequence=1&isAllowed=y) suggests but instead I use a polynomial basis,
which is not much worse since I can do 8 GF(2^8), in parallel which involves
16 GF(2^4) and 32 GF(2^2) multiplies in parallel.

crypto/aes/aes_local.h
kaduk: I agree with Pauli that ```uint64_t``` is preferred over ```unsigned long long``` directly.  Constants are a bit awkward (and at least in C90 the compiler is supposed to pick the type automatically to fit the value), so portable code with strict compiler warnings may not be able to avoid casts.

bernd-edlinger:
> Did you consider bitslicing? That can help smooth over the performance hit on parallel modes like AES-GCM. (I recently finished a constant-time fallback AES myself. I used bitslicing with this circuit. It seems faster than this PR even on single-block operations, but it's also just as likely that I messed up the comparison, so take that with a grain of salt.)

I agree that this makes pretty much sense.
With less than 64 bytes my algorithm still appears to be faster,
while as soon as more than 64 bytes are encrypted the bitslice
performs indeed better.

```
$ apps/openssl speed -evp aes-128-ctr
Doing aes-128-ctr for 3s on 16 size blocks: 851415 aes-128-ctr's in 3.00s
Doing aes-128-ctr for 3s on 64 size blocks: 427719 aes-128-ctr's in 3.00s
Doing aes-128-ctr for 3s on 256 size blocks: 185652 aes-128-ctr's in 3.00s
Doing aes-128-ctr for 3s on 1024 size blocks: 56390 aes-128-ctr's in 2.99s
Doing aes-128-ctr for 3s on 8192 size blocks: 7570 aes-128-ctr's in 3.00s
Doing aes-128-ctr for 3s on 16384 size blocks: 3819 aes-128-ctr's in 3.00s
version: 3.0.0-dev
built on: built on: Sat Jan 18 12:10:07 2020 UTC
options:bn(64,64) rc4(int) des(int) aes(partial) idea(int) blowfish(ptr) 
compiler: gcc -pthread -m64 -Wall -O3 -g -DOPENSSL_USE_NODELETE -DL_ENDIAN -DOPENSSL_BUILDING_OPENSSL -DNDEBUG -DOPENSSL_AES_CONST_TIME
CPUINFO: N/A
aes-128-ctr       4540.88k     9124.67k    15842.30k    19312.16k    20671.15k    20856.83k
```

bernd-edlinger:
Updated, rebased & changed the default to constant time.
Because in the world that I see coming, safety will have a
much higher value than speed.

richsalz:
Strongly agree: secure by default.  If you need speed, use assembler or use the existing assembler as a guideline to write code for your custom CPU.

bernd-edlinger:
There is something with some CIs:
```
In file included from crypto/aes/aes_core.c:51:0:
include/crypto/aes_platform.h:67:25: fatal error: ppc_arch.h: No such file or directory
compilation terminated.
```
and this
```
In file included from crypto/aes/aes_core.c:51:0:
include/crypto/aes_platform.h:83:25: fatal error: arm_arch.h: No such file or directory
compilation terminated.
```
Anybody know what this means????

bernd-edlinger:
Investigating now.
At least linux-armv4 is broken #11453

bernd-edlinger:
```
In file included from crypto/aes/aes_core.c:51:0:
include/crypto/aes_platform.h:83:25: fatal error: arm_arch.h: No such file or directory
```
The problem persists.  I cannot reproduce at home,
since I do only have a cross-toolchain for armv7 but not for aarch64.
Building a full featured cross-toolchain would take me at least two
days, since I have to start at zero, I need first the system root files,
to get them, I would need a working system,
or instead build linux with default config, get the linux headers from there,
build glibc for that target, using those linux headers, use the headers,
and the libs from there.  To build a glibc you need a bare metal
cross C compiler first, then build a full-featured C,C++ or other languages
as well,  Each step can fail and need to be re-done with different config
options.  I wonder if I just dont include that header file, and copy the
prototype I need from the header file to here.

bernd-edlinger:
This design seems to be insane.

bernd-edlinger:
Now a different unrelated issue:
```
 if [ -n "$DESTDIR" ]; then mkdir "$top/$DESTDIR"; if ! $make install DESTDIR="$top/$DESTDIR" >~/install.log 2>&1 ; then echo -e '\052\052 FAILED -- MAKE INSTALL'; cat ~/install.log; travis_terminate 1; fi; fi

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```
But I know how long make install can take, if it takes longer than 10.0 min
no output is produced, and the build stops.  I can think about a solution.

davidben:
If you all are interested, we have our own constant-time AES and GHASH. The AES implementation is bitsliced, with 32-bit, 64-bit, and 128-bit (SSE2 intrinsics) variants. The GHASH implementation is based on Thomas Pornin's carried multiplication trick, with u32/u64, u64/u128, and SSE2 intrinsics variations.
https://boringssl.googlesource.com/boringssl/+/72cff81939dca644188558e08852bf04e77c727e/crypto/fipsmodule/aes/aes_nohw.c
https://boringssl.googlesource.com/boringssl/+/72cff81939dca644188558e08852bf04e77c727e/crypto/fipsmodule/modes/gcm_nohw.c
(Feel free to treat those two files as contributions under the CLA.)

I haven't tried integrating the GHASH implementation, but a quick and dirty AES integration suggests it's still faster than this one. (I just #include the file without headers and did the bare minimum of tweaking until it compiled.)

```
# This PR on my machine, built for x86_64:
aes-128-ctr      17327.89k    34527.49k    62841.51k    79059.97k    85540.86k    86239.91k
aes-128-cbc      17618.22k    17796.71k    17807.27k    17810.43k    17820.33k    17820.33k

# aes_nohw.c on my machine, built for x86_64:
aes-128-ctr      31501.38k   120127.98k   142609.58k   149623.13k   152272.90k   152562.35k
aes-128-cbc      31853.24k    32730.24k    33164.71k    33293.99k    33344.17k    33346.90k
```

This is without enabling the SSE2 intrinsics version, so it should be comparing a 64-bit bitsliced AES to this PR. It's without enabling the bitsliced AES CBC decryption (`aes_nohw_cbc_encrypt`), so the CBC numbers should be representative of a serial operation that doesn't take advantage of 64-bit bitslicing's 4-block parallelism. (But bitslicing will still act on the 16 bytes in a block together.)

bernd-edlinger:
Thanks for sharing your numbers here.

It looks like your implementation was successful in using int128, right,
all numbers are roughly double from my PR here.

I tried to use int128 but when I did I had the problem that it did not double
in the case of the cbc, I thing that has to do with all the 128bit constants,
and that I cannot use constants directly, I had a macro that does this
```C
#  define T __uint128_t
#  define L(x) ((T)(x) << 64 | (T)(x))
```

I did not check how stupid the code generation could possibly be here.
but if it involves a shift and an or that would of course explain why it
fails.

Since I had no luck with that I did not even try to use int128 for the bit-sliced
ctr code.

Could you please compile your test with u64 and give me the number that
you get then?

davidben:
The GHASH code has a u128 path, but I only measured the AES implementation. There is a 128-bit path to the AES implementation using SSE2 intrinsics, but the numbers were measured without them. (It also would only have affected CTR numbers rather than CBC numbers. I used bitslicing, so only parallel modes benefit from wider words. Even u16 words are sufficient parallelism for bitsliced single-block operations.)

bernd-edlinger:
This AES CBC algorithm would benefit from u128 since that is just one AES block.
I am not sure, if the problem is not only a locality of the constant values, or
due to awkward code generation.

davidben:
Hrm. I guess this is probably down to the different implementation strategies. Mine is just straight-up bitsliced with a word for each of the 8 bit positions in a byte. 128-bit words then acts on 128 bytes, which is 128/16 = 8 blocks. Whereas I guess you have both a bitsliced version plus some other strategy.

Although, do compilers typically promote bit operations on u128 over to SSE2/NEON registers? If not, a u128 would just get spread over two registers on a 64-bit machine anyway, right? I usually only reach for u128 when I want a full 64-bit multiplication.

romen:
@bernd-edlinger could you please apply something similar to https://github.com/openssl/openssl/pull/11411/commits/8d49615a77db9cb3ee6e6b83c523fe352b5e55ad also in this branch for the CI?

romen:
@bernd-edlinger , regarding https://github.com/openssl/openssl/pull/10828#issuecomment-608856418 , do you plan to further improve your implementation based on the feedback of the discussion with @davidben  ?

Do we need to choose between your implementation in this PR and the one that @davidben is willing to contribute?
From https://github.com/openssl/openssl/pull/10828#issuecomment-608856418 it looks like at least on x86-64 BoringSSL version is more performant. Do you have more feedback regarding other platforms?

What are your plans about this PR?

Notice I marked my approval for #11411 (the 1.1.1 backport) conditional to this also being approved.
If we are going to apply changes it would be better to do so in a single PR and open the backport PR (if needed) once this gets approved.

bernd-edlinger:
> could you please apply something similar to 8d49615 also in this branch for the CI?

No, that is not necessary, as I wrote here: eec8f938ef49abdae75da9cf5f43bd6a9ce6a190

It will be enabled by default in 3.0.  This PR uses OPENSSL_NO_AES_CONST_TIME.

I knew in advance that 3.0 will not have any non-constant AES code.

You can call it premonition :smiley:

bernd-edlinger:
> regarding #10828 (comment) , do you plan to further improve your implementation based on the feedback of the discussion with @davidben ?

I am in contact with him, and if we find a way to further improve the speed of the constant
time code, that will certainly be another PR.
But so far I have not received any further input from him, so that can take a while.

bernd-edlinger:
> Do we need to choose between your implementation in this PR and the one that @davidben is willing to contribute?

I think he just compared speed on boringssl to speed of openssl,
I do not know which was the CPU frequency when he measured that
once, when you build openssl the CPU core heats up and the frequency
throttles down, that can easily make 100% error on measuring,
I am a bit surprised that the CTR numbers are so much different, because
I used his advice and added that code for CTR only the CBC code is entirely
different, more like a VHDL compiler would solve the problem.

davidben:
> But so far I have not received any further input from him, so that can take a while.

Sorry, I'd meant to look over the math in this CL, but I've been pretty hosed with work. :-(

> I think he just compared speed on boringssl to speed of openssl,

Nah, this was OpenSSL. BoringSSL's `bssl speed` is a fresh program and enough other things are different that, yeah, it was clear that only an integration in OpenSSL would make any sense. I dropped in `aes_nohw.c` from BoringSSL, applied this very hacky diff.

```
diff --git a/crypto/aes/aes_core.c b/crypto/aes/aes_core.c
index 695a63fad9..75be394fd6 100644
--- a/crypto/aes/aes_core.c
+++ b/crypto/aes/aes_core.c
@@ -51,6 +51,10 @@
 #include "aes_local.h"
 
 #if !defined(OPENSSL_NO_AES_CONST_TIME) && !defined(AES_ASM)
+#if 1
+#define OPENSSL_64_BIT
+#include "aes_nohw.c"
+#else
 typedef union {
     unsigned char b[8];
     u32 w[2];
@@ -1030,6 +1034,7 @@ void AES_ctr32_encrypt(const unsigned char *in, unsigned char *out,
         }
     }
 }
+#endif
 # endif
 #elif !defined(AES_ASM)
 /*-
```

Then I did the bare minimum of tweaking of aes_nohw.c to get it compiling. (Tweaking #includes, renaming a couple functions to match and such, etc.) The SSE2 intrinsics flow is gated on an `OPENSSL_SSE2` symbol that OpenSSL does *not* define, so I compared this PR against the plain `uint64_t` path, *not* the SSE2 intrinsics.

> I do not know which was the CPU frequency when he measured that
> once, when you build openssl the CPU core heats up and the frequency
> throttles down, that can easily make 100% error on measuring,

I built statically-linked `openssl` binaries with the config line listed in the PR, made one build, copied the file to `openssl.from-pr`, repeated it with the other version, copied it to `openssl.mine`. Then I benchmarked the two against each other after all the building as done. (I forget if I did a prewarm run first, but I couldn't figure out how to benchmark both CTR and CBC in one run, so one set of numbers probably effectively prewarmed for the other.)

> I am a bit surprised that the CTR numbers are so much different, because
> I used his advice and added that code for CTR only the CBC code is entirely
> different, more like a VHDL compiler would solve the problem.

The difference in the CTR numbers does seem relatively consistent with the difference in the CBC (which would be serial), so I'm guessing it's from raw AES performance somehow, but yeah I don't know where it comes from.

bernd-edlinger:
Yeah, thanks, I am confident that over time improvement will happen.

romen:
It is unclear to me how we should proceed about this PR, @bernd-edlinger what is  your recommendation? Should we keep it as it is or should we prefer @davidben contribution?

bernd-edlinger:
Both, I would like to keep this and integrate david ben's contribution as a follow up.

romen:
I might be confused, isn't the alternative a replacement for this?

bernd-edlinger:
no it is only about CTR (and CBC in reverse mode) but not in CBC in forward mode.
That cant be solved in bitslice.

davidben:
We use that implementation for all of them. You can bitslice serial mode too (there are 16 bytes in a block to run in parallel). It doesn't perform as well as a parallel operation, but worked well enough for us. (The CBC numbers I listed are operating serially.)

davidben:
These are the single-operation functions: https://boringssl.googlesource.com/boringssl/+/72cff81939dca644188558e08852bf04e77c727e/crypto/fipsmodule/aes/aes_nohw.c#1144

The implementation encrypts 2, 4, or 8 (depending on word size) blocks at once, so the single-block operation just discards the excess. (Hence why parallel is faster, but not strictly necessary.)

t8m:
We should not keep the master branch being without the AES consttime code if we already have it merged into 1.1.1.

@bernd-edlinger Bernd, would you like to reopen this PR and just merge the code as is (of course with the AES ctr code removed)? Or do you prefer if someone else creates a new PR with cherry-pick of the 1.1.1 AES consttime code to master?

bernd-edlinger:
As you wish, but I must say that this takes a bit more effor than re-opening

bernd-edlinger:
oh, interesting, I would have expected a merge conflict....

bernd-edlinger:
Note that the CTR32 implementaion is also still in this PR,
wich I reverted in 1.1.1 due to the unclear patent situation.

bernd-edlinger:
So really someone more capable of understanding legal issues than
me, should look if the patent is compatible with the OpenSSL license.
Please.

bernd-edlinger:
Alternative:  I remove the CTR32 constant time code here as well?
Which one should I do?

mattcaswell:
> Alternative: I remove the CTR32 constant time code here as well?

I suggest removing here as well so that 1.1.1 and master are in sync.

romen:
I think you should be able to apply the commit from #11626 onto this branch.

At least that way we bring master and 1.1.1 back in sync about features.

bernd-edlinger:
Okay, done.

t8m:
I think this should have a CHANGES.md entry.

bernd-edlinger:
It already has a CHANGES entry here:
https://github.com/openssl/openssl/blob/cf94e8430f3cd7c17f62b74443d16347b4b97ac8/CHANGES#L26-L33

In the 1.1.1 branch, all those change entries are missing from the master version.

crypto/aes/aes_local.h
paulidale: It seems that agreement has been reached about this using the `uint64_t` type instead of `unsigned long long` (which BTW omits the final `int`).

In the past 64 bit integer constants have been wrapped up in macros that are given two 32 bit arguments.

paulidale (approved):

paulidale:
A rerun of the CIs is advisable.

Implementing this as a separate algorithm that can be fetched selectively might also be beneficial -- this can wait though.

crypto/aes/aes_core.c
```diff
@@ -50,7 +50,637 @@
 #include <openssl/aes.h>
 #include "aes_local.h"
 
-#ifndef AES_ASM
+#if !defined(OPENSSL_NO_AES_CONST_TIME) && !defined(AES_ASM)
```
slontis, line 53: There is a CHANGES.md entry in 1.1.1g..

Is this still valid????

   Enable with: ./config no-asm -DOPENSSL_AES_CONST_TIME
   Disable with: ./config no-asm -DOPENSSL_NO_AES_CONST_TIME
   At this time this feature is by default disabled.
   It will be enabled by default in 3.0.

bernd-edlinger: Yes, if this PR gets merged, it will be as it has been written there.

richsalz:
> Implementing this as a separate algorithm that can be fetched selectively might also be beneficial 

I strongly disagree.  Add a "non-constant-time" query property if desired.  Constant time *must* be true by default for every cipher exposed on the Internet.

paulidale:
Adding a non-constant-time property is little different to having a constant-time one.  _Neither_ will be selected by default.

A strike against it is that the negative is a detraction.  `non-constant-time=no` is harder to understand than `constant-time=yes`.

richsalz:
My point is that algorithms used on the Internet should be constant time by default.  If it is awkward for users to say that don't want constant-time, then that is fine with me. IT is the safest thing.

bernd-edlinger:
It is an inconsistency that Openssl 3.0 still have the BSAES algorithm, which is not completely constant time,
and even prefer BSAES over VPAES when both are available. (while 1.1.1 has at least no x86-BSAES).

richsalz:
> It is an inconsistency that Openssl 3.0 still have the BSAES algorithm, which is not completely constant time,
> and even prefer BSAES over VPAES when both are available. (while 1.1.1 has at least no x86-BSAES).

Not inconsitency, "security flaw" :)

bernd-edlinger:
Merged to master as 0051746

PR #10829, opened 2020-01-13
Hjl/cet/master

hjl-tools:
This set of patches enables CET on Linux/i686 and Linux/x86-64. Tested on CET processor.
Hopefully to fix Windows build.

### DIFF
--- a/crypto/async/arch/async_posix.h
+++ b/crypto/async/arch/async_posix.h
@@ -25,17 +25,27 @@
 #  define ASYNC_POSIX
 #  define ASYNC_ARCH
 
+#  ifdef __CET__
+#    define USE_SWAPCONTEXT
+#  endif
 #  include <ucontext.h>
-#  include <setjmp.h>
+#    ifndef USE_SWAPCONTEXT
+#      include <setjmp.h>
+#    endif
 
 typedef struct async_fibre_st {
     ucontext_t fibre;
+#    ifndef USE_SWAPCONTEXT
     jmp_buf env;
     int env_init;
+#    endif
 } async_fibre;
 
 static ossl_inline int async_fibre_swapcontext(async_fibre *o, async_fibre *n, int r)
 {
+#    ifdef USE_SWAPCONTEXT
+    swapcontext(&o->fibre, &n->fibre);
+#    else
     o->env_init = 1;
 
     if (!r || !_setjmp(o->env)) {
@@ -44,6 +54,7 @@ static ossl_inline int async_fibre_swapcontext(async_fibre *o, async_fibre *n, i
         else
             setcontext(&n->fibre);
     }
+#    endif
 
     return 1;
 }
--- a/crypto/md5/asm/md5-x86_64.pl
+++ b/crypto/md5/asm/md5-x86_64.pl
@@ -131,6 +131,9 @@ $0 =~ m/(.*[\/\\])[^\/\\]+$/; my $dir=$1; my $xlate;
 ( $xlate="${dir}../../perlasm/x86_64-xlate.pl" and -f $xlate) or
 die "can't locate x86_64-xlate.pl";
 
+push(@INC,"${dir}","${dir}../../perlasm");
+require "x86_64asm.pl";
+
 open OUT,"| \"$^X\" \"$xlate\" $flavour \"$output\""
     or die "can't call $xlate: $!";
 *STDOUT=*OUT;
@@ -392,4 +395,5 @@ ___
 
 print $code;
 
+&file_end($flavour);
 close STDOUT;
--- a/crypto/modes/asm/ghash-x86_64.pl
+++ b/crypto/modes/asm/ghash-x86_64.pl
@@ -102,6 +102,9 @@ $0 =~ m/(.*[\/\\])[^\/\\]+$/; $dir=$1;
 ( $xlate="${dir}../../perlasm/x86_64-xlate.pl" and -f $xlate) or
 die "can't locate x86_64-xlate.pl";
 
+push(@INC,"${dir}","${dir}../../perlasm");
+require "x86_64asm.pl";
+
 if (`$ENV{CC} -Wa,-v -c -o /dev/null -x assembler /dev/null 2>&1`
 		=~ /GNU assembler version ([2-9]\.[0-9]+)/) {
 	$avx = ($1>=2.20) + ($1>=2.22);
@@ -1817,4 +1820,5 @@ $code =~ s/\`([^\`]*)\`/eval($1)/gem;
 
 print $code;
 
+&file_end($flavour);
 close STDOUT;
--- a/crypto/modes/asm/ghash-x86_64.pl
+++ b/crypto/modes/asm/ghash-x86_64.pl
@@ -105,6 +105,8 @@ die "can't locate x86_64-xlate.pl";
 push(@INC,"${dir}","${dir}../../perlasm");
 require "x86_64asm.pl";
 
+$endbr=&endbr64();
+
 if (`$ENV{CC} -Wa,-v -c -o /dev/null -x assembler /dev/null 2>&1`
 		=~ /GNU assembler version ([2-9]\.[0-9]+)/) {
 	$avx = ($1>=2.20) + ($1>=2.22);
@@ -244,6 +246,7 @@ $code=<<___;
 .align	16
 gcm_gmult_4bit:
 .cfi_startproc
+	$endbr
 	push	%rbx
 .cfi_push	%rbx
 	push	%rbp		# %rbp and others are pushed exclusively in
@@ -291,6 +294,7 @@ $code.=<<___;
 .align	16
 gcm_ghash_4bit:
 .cfi_startproc
+	$endbr
 	push	%rbx
 .cfi_push	%rbx
 	push	%rbp
@@ -617,6 +621,7 @@ $code.=<<___;
 .align	16
 gcm_gmult_clmul:
 .cfi_startproc
+	$endbr
 .L_gmult_clmul:
 	movdqu		($Xip),$Xi
 	movdqa		.Lbswap_mask(%rip),$T3
@@ -668,6 +673,7 @@ $code.=<<___;
 .align	32
 gcm_ghash_clmul:
 .cfi_startproc
+	$endbr
 .L_ghash_clmul:
 ___
 $code.=<<___ if ($win64);
@@ -1171,6 +1177,7 @@ $code.=<<___;
 .align	32
 gcm_gmult_avx:
 .cfi_startproc
+	$endbr
 	jmp	.L_gmult_clmul
 .cfi_endproc
 .size	gcm_gmult_avx,.-gcm_gmult_avx
@@ -1182,6 +1189,7 @@ $code.=<<___;
 .align	32
 gcm_ghash_avx:
 .cfi_startproc
+	$endbr
 ___
 if ($avx) {
 my ($Xip,$Htbl,$inp,$len)=@_4args;
--- a/crypto/perlasm/cbc.pl
+++ b/crypto/perlasm/cbc.pl
@@ -6,6 +6,17 @@
 # in the file LICENSE in the source distribution or at
 # https://www.openssl.org/source/license.html
 
+# Collect all compiler flags.
+$cflags="";
+for($i=1;$i<$#ARGV;$i++) {
+    if ($ARGV[$i] eq "386") {
+	last;
+    }
+    $cflags.=" $ARGV[$i]";
+}
+
+# Check if Intel CET is enabled at compile time.
+$endbr32=(`echo "#ifdef __CET__\n#error CET is enabled\n#endif" | $ENV{CC} $cflags -c -o /dev/null -x c - 2>&1` =~ /CET is enabled/)?"endbr32":"";
 
 # void des_ncbc_encrypt(input, output, length, schedule, ivec, enc)
 # des_cblock (*input);
@@ -165,21 +176,28 @@ sub cbc
 	&jmp_ptr($count);
 
 &set_label("ej7");
+	&endbranch()		if ($endbr32);
 	&movb(&HB("edx"),	&BP(6,$in,"",0));
 	&shl("edx",8);
 &set_label("ej6");
+	&endbranch()		if ($endbr32);
 	&movb(&HB("edx"),	&BP(5,$in,"",0));
 &set_label("ej5");
+	&endbranch()		if ($endbr32);
 	&movb(&LB("edx"),	&BP(4,$in,"",0));
 &set_label("ej4");
+	&endbranch()		if ($endbr32);
 	&mov("ecx",		&DWP(0,$in,"",0));
 	&jmp(&label("ejend"));
 &set_label("ej3");
+	&endbranch()		if ($endbr32);
 	&movb(&HB("ecx"),	&BP(2,$in,"",0));
 	&shl("ecx",8);
 &set_label("ej2");
+	&endbranch()		if ($endbr32);
 	&movb(&HB("ecx"),	&BP(1,$in,"",0));
 &set_label("ej1");
+	&endbranch()		if ($endbr32);
 	&movb(&LB("ecx"),	&BP(0,$in,"",0));
 &set_label("ejend");
 
--- a/crypto/perlasm/x86_64asm.pl
+++ b/crypto/perlasm/x86_64asm.pl
@@ -1,0 +1,61 @@
+#! /usr/bin/env perl
+# Copyright 2020 The OpenSSL Project Authors. All Rights Reserved.
+#
+# Licensed under the OpenSSL license (the "License").  You may not use
+# this file except in compliance with the License.  You can obtain a copy
+# in the file LICENSE in the source distribution or at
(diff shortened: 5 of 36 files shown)


hjl-tools:
I replaced

```
$endbr32=(`echo __CET__ | $ENV{CC} -E -` !~ /__CET__/)?"endbr32":"";
```

with

```
$endbr32=(`echo "#ifdef __CET__\n#error CET is enabled\n#endif" | $ENV{CC} -c -o /dev/null -x c - 2>&1` =~ /CET is enabled/)?"endbr32":"";
```

crypto/modes/asm/ghash-x86_64.pl
```diff
@@ -613,6 +620,7 @@ sub reduction_alg9 {	# 17/11 times faster than Intel version
 .type	gcm_gmult_clmul,\@abi-omnipotent
 .align	16
 gcm_gmult_clmul:
+	$endbr
```
bernd-edlinger: If this adds an opcode here, then it will be before the .cfi_startproc,
and therefore make the stack unwinder fail

crypto/perlasm/cbc.pl
```diff
@@ -6,6 +6,8 @@
 # in the file LICENSE in the source distribution or at
 # https://www.openssl.org/source/license.html
 
+# Check if Intel CET is enabled at compile time.
+$endbr32=(`echo "#ifdef __CET__\n#error CET is enabled\n#endif" | $ENV{CC} -c -o /dev/null -x c - 2>&1` =~ /CET is enabled/)?"endbr32":"";
```
bernd-edlinger: I don't see how this is going to work if -fcf-protection or -fcf-protection=none is passed
to ./config.  this can only find out if the host compiler is configured to use that by default.

crypto/modes/asm/ghash-x86_64.pl
hjl-tools: Thanks.  I pushed a fix.

bernd-edlinger: What I also do not understand, why you only add that here, there are many more
functions here, look for ".cfi_startproc", what is the rationale for this?

crypto/perlasm/cbc.pl
hjl-tools: I couldn't find a way to access CFLAGS from perl script.  To enable CET, I did

```
export CC="gcc -fcf-protection -Wl,-z,cet-report=error" 
```
first.

crypto/modes/asm/ghash-x86_64.pl
hjl-tools: Only functions which are called indirectly need to start with `endbr64`.   Since most of
functions aren't called indirectly, they don't need `endbr64`.

crypto/perlasm/cbc.pl
bernd-edlinger: The invocation of the generate-asm script should pass all options,
for instance here:
```
crypto/aes/aesni-sha1-x86_64.s: crypto/aes/asm/aesni-sha1-x86_64.pl
        CC="$(CC)" $(PERL) crypto/aes/asm/aesni-sha1-x86_64.pl "$(PERLASM_SCHEME)" -I. -Iinclude -Iproviders/common/include -Iproviders/implementations/include -Icrypto/include $(LIB_CFLAGS) $(LIB_CPPFLAGS) -DAES_ASM -DBSAES_ASM -DCMLL_ASM -DECP_NISTZ256_ASM -DGHASH_ASM -DKECCAK1600_ASM -DMD5_ASM -DOPENSSL_BN_ASM_GF2m -DOPENSSL_BN_ASM_MONT -DOPENSSL_BN_ASM_MONT5 -DOPENSSL_CPUID_OBJ -DOPENSSL_IA32_SSE2 -DPOLY1305_ASM -DSHA1_ASM -DSHA256_ASM -DSHA512_ASM -DVPAES_ASM -DWHIRLPOOL_ASM -DX25519_ASM $(PROCESSSOR) $@
```
you should use them for such decisions (but note that PROCESSOR can be 386 or empty,
and PERLASM_SCHEME= elf or something else.

hjl-tools: I updated my patch with

```
# Collect all compiler flags.
my $cflags="";
for($i=0;$i<=$#ARGV;$i++) {
    if ($ARGV[$i] eq "386") {
	last;
    }
    $cflags.=" $ARGV[$i]";
}

# Check if Intel CET is enabled at compile time.
my $cet_enabled=(`echo "#ifdef __CET__\n#error CET is enabled\n#endif" | $ENV{CC} $cflags -c -o /dev/null -x c - 2>&1` =~ /CET is enabled/)?1:0;
```

crypto/md5/asm/md5-x86_64.pl
```diff
@@ -392,4 +395,5 @@ sub round4_step
 
 print $code;
 
+&file_end($flavour);
```
bernd-edlinger: can't you do just write ".file_end" here and do the replacement in x86_64-xlate.pl ?

hjl-tools: Currently everything is done in x86_64asm.pl.  Doing it in x86_64-xlate.pl needs to change
every invocation of x86_64-xlate.pl to pass CET info to it.   I don't see any advantage.

hjl-tools:
Something is wrong with the Travis CI build itself:

https://travis-ci.org/openssl/openssl/jobs/636696435?utm_medium=notification&utm_source=github_status

```
All tests successful.
Files=190, Tests=1749, 210 wallclock secs ( 5.14 usr  0.97 sys + 132.41 cusr 51.40 csys = 189.92 CPU)
Result: PASS
The command "if [ -z "$BUILDONLY" ]; then if [ -n "$CROSS_COMPILE" ]; then sudo dpkg --add-architecture i386; sudo apt-get update; sudo apt-get -yq --no-install-suggests --no-install-recommends --force-yes install wine; export EXE_SHELL="wine" WINEPREFIX=`pwd`; fi; if [ -e krb5/src ]; then sudo apt-get -yq install bison dejagnu gettext keyutils ldap-utils libldap2-dev libkeyutils-dev python-cjson python-paste python-pyrad slapd tcl-dev tcsh; fi; if ! HARNESS_VERBOSE_FAILURE=yes BORING_RUNNER_DIR=$top/boringssl/ssl/test/runner make test; then echo -e '\052\052 FAILED -- MAKE TEST'; travis_terminate 1; fi; else if ! $make build_tests >~/build.log 2>&1; then echo -e '\052\052 FAILED -- MAKE BUILD_TESTS'; cat ~/build.log travis_terminate 1; fi; fi" exited with 0.
$ if [ -n "$DESTDIR" ]; then mkdir "$top/$DESTDIR"; if ! $make install DESTDIR="$top/$DESTDIR" >~/install.log 2>&1 ; then echo -e '\052\052 FAILED -- MAKE INSTALL'; cat ~/install.log; travis_terminate 1; fi; fi
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

The 64-bit build failed on a commit which only touched one 32-bit file.  It indicates the build
machine is broken.

bernd-edlinger:
I wonder if this still needed then?
https://github.com/openssl/openssl/blob/7c66ad65f959fa05ad7b95b0c84384d6e63e56f1/crypto/perlasm/x86_64-xlate.pl#L1130-L1132

hjl-tools:
> I wonder if this still needed then?
> https://github.com/openssl/openssl/blob/7c66ad65f959fa05ad7b95b0c84384d6e63e56f1/crypto/perlasm/x86_64-xlate.pl#L1130-L1132

No, it isn't needed.

hjl-tools:
I pushed a new patch to remove $endbranch from x86_64-xlate.pl.

crypto/rc4/asm/rc4-x86_64.pl
```diff
@@ -140,7 +145,9 @@
 .globl	RC4
 .type	RC4,\@function,4
 .align	16
-RC4:	or	$len,$len
+RC4:
```
bernd-edlinger: while you are already here, could you please move the .cfi_startproc from line 154 to here?
Thanks!

crypto/rc4/asm/rc4-x86_64.pl
```diff
@@ -529,6 +537,7 @@ sub RC4_loop {
 .type	RC4_options,\@abi-omnipotent
 .align	16
 RC4_options:
+	$endbr
```
bernd-edlinger: hmm.  This function misses both .cfi_startproc and .cfi_endproc,
as well as the .size decoration, could you please add those here as well?

bernd-edlinger:
by the way, how did you find where the branch targets are missing and where not?

crypto/async/arch/async_posix.h
```diff
@@ -25,17 +25,27 @@
...
+#    ifndef USE_SWAPCONTEXT
     jmp_buf env;
     int env_init;
+#    endif
 } async_fibre;
 
 static ossl_inline int async_fibre_swapcontext(async_fibre *o, async_fibre *n, int r)
 {
+#    ifdef USE_SWAPCONTEXT
+    swapcontext(&o->fibre, &n->fibre);
```
bernd-edlinger: can you explain why this is done?
maybe even add a comment ?

crypto/md5/asm/md5-x86_64.pl
levitte: Except this mixes together two styles of generating assembler. I'm not so sure it's a good thing, even if it's an easy hack now...

levitte:
There are things in here that I find disturbing:

1)  The check of `__CET__` so deeply embedded that it approaches dark magic.  Do you know that you can use C preprocessor directives in the perlasm scripts?  The files are preprocessed before perls gets to see them.  All it takes for that to happen is to use the suffix `.S` instead of `.s` in the `build.info` file, i.e. for example:

    ``` diff
    diff --git a/crypto/aes/build.info b/crypto/aes/build.info
    index dc00df0cda..67e9a85135 100644
    --- a/crypto/aes/build.info
    +++ b/crypto/aes/build.info
    @@ -8,7 +8,7 @@ IF[{- !$disabled{asm} -}]
       $AESDEF_x86_sse2=VPAES_ASM
     
       $AESASM_x86_64=\
    -        aes-x86_64.s vpaes-x86_64.s bsaes-x86_64.s aesni-x86_64.s \
    +        aes-x86_64.S vpaes-x86_64.s bsaes-x86_64.s aesni-x86_64.s \
             aesni-sha1-x86_64.s aesni-sha256-x86_64.s aesni-mb-x86_64.s
       $AESDEF_x86_64=AES_ASM VPAES_ASM BSAES_ASM
     
    @@ -78,7 +78,7 @@ DEPEND[vpaes-586.s]=../perlasm/x86asm.pl
     GENERATE[aesni-x86.s]=asm/aesni-x86.pl
     DEPEND[aesni-586.s]=../perlasm/x86asm.pl
     
    -GENERATE[aes-x86_64.s]=asm/aes-x86_64.pl
    +GENERATE[aes-x86_64.S]=asm/aes-x86_64.pl
     GENERATE[vpaes-x86_64.s]=asm/vpaes-x86_64.pl
     GENERATE[bsaes-x86_64.s]=asm/bsaes-x86_64.pl
     GENERATE[aesni-x86_64.s]=asm/aesni-x86_64.pl
    ```

2)  It's possible to add custom directive and instructions in the translators.  For example, `crypto/perlasm/arm-xlate.pl` supports a `.arch` directive, and it would be perfectly possible to recognise a similar directive in `crypto/perlasm/x86_64-xlate.pl` and store it away as an indicator of some sort.  It would be perfectly possible to have some code at the end of this translator that does something like this:

    ``` perl
    print "\n$arch_file_end\n" if ($arch_file_end);
    ```

    and have the `.arch` directive fill in `$arch_file_end` appropriately, depend on argument value.  This means that, together with C preprocessing, you could add something like this at the beginning of the assembler code in, say, `crypto/aes/asm/aes-x86_64.pl`:

    ``` C
    #ifdef __CET__
    .arch cet // affects the end of the assembler output
    #endif
    ```

    Look for `package directive` for the section where directives are defined.

3)  Why not use the `endbranch` mnemonic?  `x86_64-xlate.pl` will try to evaluate it as a perl variable, i.e. `$endbranch`, and the opcodes emit from that variable are already the ENDBR64 opcodes.  Just use it, like this:

    ```
    #ifdef __CET__
    	endbranch
    #endif
    ```

levitte:
Regarding `endbranch`, it would absolutely be possible to have it emit those opcodes only when the `cet` "architecture" has been detected.  That way, you wouldn't need to guard the use with `#ifdef __CET__`

levitte:
Regarding a guard around `endbranch`, it may not even be necessary.  From what I'm reading, ENDBR64 will act as a NOP in x86_64 that doesn't have this feature, so...

crypto/rc4/asm/rc4-x86_64.pl
hjl-tools: I will submit a separate patch.

crypto/rc4/asm/rc4-x86_64.pl
hjl-tools: I will submit a separate patch.

hjl-tools:
> by the way, how did you find where the branch targets are missing and where not?

By running openssl tests and  visual inspection.

crypto/async/arch/async_posix.h
hjl-tools: I will add some comments.

hjl-tools:
> Regarding a guard around `endbranch`, it may not even be necessary. From what I'm reading, ENDBR64 will act as a NOP in x86_64 that doesn't have this feature, so...

Yes, ENDBR64 is NOP if CET isn't available.  Can I emit it unconditionally?

bernd-edlinger:
Is it measurable in performance?

hjl-tools:
> Is it measurable in performance?

No.

kroeckx:
> Yes, ENDBR64 is NOP if CET isn't available.  Can I emit it unconditionally?

I don't see a problem with that.

levitte:
> Yes, ENDBR64 is NOP if CET isn't available. Can I emit it unconditionally?

Yup. So that means you can use `endbranch` and be done. You do have `&endbranch` already for x86, BTW. No need to define `&endbr32`

levitte:
I ended up being inspired and submitted #10875.  With that, all that's needed is to start using the pseudo-mnemonic `endbranch` in a x86_64 perlasm script, and the CET property will be emitted automatically.

levitte:
#10875 has been merged.

@hjl-tools, are you still around?

crypto/perlasm/x86_64asm.pl
```diff
@@ -0,0 +1,61 @@
...
+
+# Check if Intel CET is enabled at compile time.
+my $cet_enabled=(`echo "#ifdef __CET__\n#error CET is enabled\n#endif" | $ENV{CC} $cflags -c -o /dev/null -x c - 2>&1` =~ /CET is enabled/)?1:0;
+
+sub ::file_end
+{
+    my($flavour)=@_;
+
+    if ($cet_enabled and $flavour =~ /elf/) {
+	my $p2align=(`echo "#ifdef __ILP32__\n#error ILP32 is enabled\n#endif" | $ENV{CC} $cflags -c -o /dev/null -x c - 2>&1` =~ /ILP32 is enabled/)?2:3;
```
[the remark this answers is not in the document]
hjl-tools: X32 must use 4 byte alignment and x86-64 must use 8 byte alignment.

hjl-tools:
> #10875 has been merged.
> 
> @hjl-tools, are you still around?

Yes.   FYI, CET capable compiler supports "#include <cet.h>" in assembly code (.S) to generate
CET property automatically.

levitte:
It would be preferable not to have to depend too much on things that may or may not be there. We try to have our perlasm scripts as environment agnostic as possible, which is why you see all those .byte emits for instructions that some assemblers might not support yet, among others...

crypto/perlasm/x86_64asm.pl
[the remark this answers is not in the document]
bernd-edlinger: Does anybody know if openssl runs on a x32 target at all?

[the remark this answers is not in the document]
kroeckx: Debian has a x32 port, and it successfully builds OpenSSL and passed the test suite.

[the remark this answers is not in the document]
kroeckx: We also have a "linux-x32" target, that has `asm_arch => 'x86_64'`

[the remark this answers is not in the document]
hjl-tools: X32 needs some a patch on master branch:

https://github.com/hjl-tools/openssl/commit/b5b2f8c57a529399903feccaed729f60cb42e1f9

t8m:
The CET support was already merged from other PRs, or is there anything missing?

hjl-tools:
> The CET support was already merged from other PRs, or is there anything missing?

No.

Issue #6384, opened 2018-05-31
Rename OpenSSL to OpenTLS To Comply with RFC7568


jandrusk:
It's 2018 and 'SSL' is now a deprecated term given RFC7568 along with the POODLE and BEAST vulnerabilities that explicitly require the disabling of the SSL protocols. I move to have OpenSSL re-branded to OpenTLS to comply with RFC7568 and to eliminate confusion.

richsalz:
The OpenSSL "brand" is very recognized.  We're not going to do this.

richsalz:
Another team member pointed out that I might be too harsh.  Perhaps better is "we have no plans to do this at this time."

jandrusk:
The name doesn't reflect reality.

richsalz:
But it does have mindshare and reflect history. :)

mspncp:
> I move to have OpenSSL re-branded to OpenTLS to comply with RFC7568 and to eliminate confusion.

FWIW: Today I stumbled over a very interesting historical fact about the rebranding of OpenSSL: Initially, when it emerged from SSLeay, it was called OpenTLS and only later on renamed to OpenSSL:

```
commit f1c236f849d9799a5de0ad1f6c64b33291f60c84
Author: Ralf S. Engelschall <rse@openssl.org>
Date:   Wed Dec 23 07:38:54 1998 +0000

    Switch to OpenSSL name
```
(see f1c236f849d9) 

```
commit c537fb08cd93c3c564478e6a238e502a7d17e216
Author: Ralf S. Engelschall <rse@openssl.org>
Date:   Tue Dec 22 16:01:06 1998 +0000

    OpenTLS ready
```
(see c537fb08cd93)

```
commit 651d0aff98d28e2db146afa1790e9e22f3ef22db
Author: Ralf S. Engelschall <rse@openssl.org>
Date:   Tue Dec 22 15:04:48 1998 +0000

    Various cleanups and fixed by Marc and Ralf to start the OpenTLS project
```
(see 651d0aff98d2)

mspncp:
Ok, looking a bit closer at the dates: it was called OpenTLS for less than a day ;-)

iamamoose:
Yes, the original planned name was going to be OpenTLS and it was changed to OpenSSL at the very last minute.  It was fortunate we'd registered both domain names!  I'll be writing a blog about this :)

InfoHunter:
> I'll be writing a blog about this :)

[...]

PR #10830, opened 2020-01-13
OpenSSL -> OpenTLS

the-xentropy:
Rename project to openTLS.


slontis:
This was discussed some time ago
https://github.com/openssl/openssl/issues/6384.

mattcaswell:
This has been discussed from time to time by the OMC. Whilst there are differing views on this within the OMC I don't think there is sufficient support there to make such a change. I doubt this will happen anytime soon.

paulidale:
My suspicion is that if this change is ever made, it will need to be done in a more gradual manner.

mattcaswell:
There are no plans to make this change, so closing this PR.

PR #10831, opened 2020-01-13
mdc2: use evp_test instead of a separate test application.

paulidale:
One of the MDC2 test applications can be done using evp_test.
This makes it so.


paulidale:
Note that MDC2 has a second test application that uses the low level APIs directly.

paulidale:
Rebased, ping.

paulidale:
Ping reviewer.

paulidale:
Merged to master.  Thanks.

PR #10820, opened 2020-01-13
changes: combined CHANGES entry for deprecated low level cipher functions.

paulidale:

PR #10833, opened 2020-01-13
Deprecate the low level SEED functions

paulidale:
Use of the low level SEED functions has been informally discouraged for a long time. We now formally deprecate them.


paulidale:
Changes entry in #10820

paulidale:
Merged to master, thanks.

PR #10834, opened 2020-01-14
Deprecate low level calls to RC2, RC4 and RC5.

paulidale:
Use of the low level RC2, RC4 and RC5 functions has been informally discouraged for a long time.  We now formally deprecate them.


paulidale:
CHANGES changes are in #10820.

paulidale:
Merged to master, thanks for the review.

PR #10836, opened 2020-01-14
mac: deprecate low level calls to CMAC and HMAC.

paulidale:
Use of the low level CMAC and HMAC functions has been informally discouraged for a long time. We now formally deprecate them.


paulidale:
CMAC is complete.  HMAC is more problematic.

include/openssl/hmac.h
```diff
@@ -24,34 +24,34 @@
...
-/*__owur*/ int HMAC_Update(HMAC_CTX *ctx, const unsigned char *data,
-                           size_t len);
-/*__owur*/ int HMAC_Final(HMAC_CTX *ctx, unsigned char *md,
-                          unsigned int *len);
-unsigned char *HMAC(const EVP_MD *evp_md, const void *key, int key_len,
+/*__owur*/ DEPRECATEDIN_3_0(int HMAC_Init_ex(HMAC_CTX *ctx, const void *key, int len,
+                            const EVP_MD *md, ENGINE *impl))
+/*__owur*/ DEPRECATEDIN_3_0(int HMAC_Update(HMAC_CTX *ctx, const unsigned char *data,
+                           size_t len))
+/*__owur*/ DEPRECATEDIN_3_0(int HMAC_Final(HMAC_CTX *ctx, unsigned char *md,
```
levitte: Drop the owur comments

paulidale: done.

test/hmactest.c
```diff
@@ -7,6 +7,12 @@
  * https://www.openssl.org/source/license.html
  */
 
+/*
+ * HMAC low level APIs are deprecated for public use, but still ok for internal
+ * use.
+ */
+#include "internal/deprecated.h"
+
```
levitte: Is there any point in keeping this program?  Replace it with stanzas for evp_test?

paulidale: It is testing the low level API.  There exist `evp_test` tests already.  I've not checked the overlap.

I'm fine to remove it.

levitte: How does this differ from #10831?

levitte: Going through evp_test still ends up testing the low level API, doesn't it?

paulidale: #10831 doesn't exercise the low level APIs, it's EVP all the way.

The functional is tested either way but the APIs not necessarily (at least in the future).

paulidale:
Bringing this out of WiP for review.

The Travis failure isn't relevant.

apps/lib/s_cb.c
```diff
@@ -770,13 +776,42 @@ int generate_cookie_callback(SSL *ssl, unsigned char *cookie,
     BIO_ADDR_rawaddress(peer, buffer + sizeof(port), NULL);
 
     /* Calculate HMAC of buffer using the secret */
-    HMAC(EVP_sha1(), cookie_secret, COOKIE_SECRET_LENGTH,
-         buffer, length, cookie, cookie_len);
-
+    hmac = EVP_MAC_fetch(NULL, SN_hmac, NULL);
```
mattcaswell: Should we really be using the `SN_*` defines for this sort of thing? They are related to the old legacy object database. Wouldn't it just be better to use "HMAC" explicitly, or provide new defines?

apps/lib/s_cb.c
```diff
@@ -770,13 +776,42 @@ int generate_cookie_callback(SSL *ssl, unsigned char *cookie,
...
+    if (hmac == NULL) {
+            BIO_printf(bio_err, "HMAC not found\n");
+            goto end;
+    }
+    ctx = EVP_MAC_CTX_new(hmac);
+    if (ctx == NULL) {
+            BIO_printf(bio_err, "HMAC context allocation failed\n");
+            goto end;
+    }
+    *p++ = OSSL_PARAM_construct_utf8_string(OSSL_MAC_PARAM_DIGEST, SN_sha1, 0);
```
mattcaswell: Similarly here

apps/lib/s_cb.c
```diff
@@ -770,13 +776,42 @@ int generate_cookie_callback(SSL *ssl, unsigned char *cookie,
...
+    }
+    if (!EVP_MAC_init(ctx)) {
+            BIO_printf(bio_err, "HMAC context initialisation failed\n");
+            goto end;
+    }
+    if (!EVP_MAC_update(ctx, buffer, length)) {
+            BIO_printf(bio_err, "HMAC context update failed\n");
+            goto end;
+    }
+    if (!EVP_MAC_final(ctx, cookie, NULL, (size_t)cookie_len)) {
```
mattcaswell: Side note: This is quite a lot of code required to replace the simple one-liner that was there before. I wonder if we should have some kind of simple "one shot" helper functions for doing this kind of thing?

CHANGES
```diff
@@ -15,6 +15,25 @@
...
+     [Paul Dale]
+
+  *) All of the low level CMAC functions have been deprecated including:
+     CMAC_CTX_new, CMAC_CTX_cleanup, CMAC_CTX_free, CMAC_CTX_get0_cipher_ctx,
+     CMAC_CTX_copy, CMAC_Init, CMAC_Update, CMAC_Final and CMAC_resume.
+     Use of these low level functions has been informally discouraged for a long
+     time.  Instead applications should use L<EVP_MAC_CTX_new(3)>,
+     L<EVP_MAC_CTX_free(3)>, L<EVP_MAC_init(3)>, L<EVP_MAC_update(3)>
+     and L<EVP_MAC_final(3)>.
+     [Paul Dale]
```
mattcaswell: The deprecation of `SSL_CTX_set_tlsext_ticket_key_cb` needs a mention here

doc/man3/SSL_CTX_set_tlsext_ticket_key_cb.pod
```diff
@@ -188,6 +224,10 @@ L<SSL_CTX_sess_number(3)>,
 L<SSL_CTX_sess_set_get_cb(3)>,
 L<SSL_CTX_set_session_id_context(3)>,
 
+=head1 HISTORY
+
+The SSL_CTX_set_tlsext_ticket_key_cb() function was deprecated in OpenSSL 3.0.
```
mattcaswell: We should mention that `SSL_CTX_set_tlsext_ticket_key_evp_cb` was added in OpenSSL 3.0.

ssl/statem/statem_srvr.c
```diff
@@ -3888,9 +3901,9 @@ static int construct_stateless_ticket(SSL *s, WPACKET *pkt, uint32_t age_add,
         if (RAND_bytes(iv, iv_len) <= 0
                 || !EVP_EncryptInit_ex(ctx, cipher, NULL,
                                        tctx->ext.secure->tick_aes_key, iv)
-                || !HMAC_Init_ex(hctx, tctx->ext.secure->tick_hmac_key,
-                                 sizeof(tctx->ext.secure->tick_hmac_key),
-                                 EVP_sha256(), NULL)) {
+                || !ssl_hmac_init(hctx, tctx->ext.secure->tick_hmac_key,
+                                  sizeof(tctx->ext.secure->tick_hmac_key),
+                                  SN_sha256)) {
```
mattcaswell: Another SN_* usage

ssl/t1_lib.c
```diff
@@ -1471,9 +1492,9 @@ SSL_TICKET_STATUS tls_decrypt_ticket(SSL *s, const unsigned char *etick,
             ret = SSL_TICKET_NO_DECRYPT;
             goto end;
         }
-        if (HMAC_Init_ex(hctx, tctx->ext.secure->tick_hmac_key,
-                         sizeof(tctx->ext.secure->tick_hmac_key),
-                         EVP_sha256(), NULL) <= 0
+        if (ssl_hmac_init(hctx, tctx->ext.secure->tick_hmac_key,
+                          sizeof(tctx->ext.secure->tick_hmac_key),
+                          SN_sha256) <= 0
```
mattcaswell: And here

ssl/t1_lib.c
```diff
@@ -2971,3 +2992,119 @@ uint8_t SSL_SESSION_get_max_fragment_length(const SSL_SESSION *session)
...
+#ifndef OPENSSL_NO_DEPRECATED_3_0
+    if (ctx->ext.ticket_key_evp_cb == NULL
+            && ctx->ext.ticket_key_cb != NULL) {
+        ret->old_ctx = HMAC_CTX_new();
+        if (ret->old_ctx == NULL)
+            goto err;
+        return ret;
+    }
+#endif
+    mac = EVP_MAC_fetch(NULL, SN_hmac, NULL);
```
mattcaswell: and here

test/sslapitest.c
```diff
@@ -6245,6 +6255,183 @@ static int test_ticket_callbacks(int tst)
...
+{
+    const unsigned char tick_aes_key[16] = "0123456789abcdef";
+    unsigned char tick_hmac_key[16] = "0123456789abcdef";
+    OSSL_PARAM params[3];
+
+    tick_key_cb_called = 1;
+    memset(iv, 0, AES_BLOCK_SIZE);
+    memset(key_name, 0, 16);
+    params[0] = OSSL_PARAM_construct_utf8_string(OSSL_MAC_PARAM_DIGEST,
+                                                 SN_sha256, 0);
```
mattcaswell: and here

test/sslapitest.c
```diff
@@ -6245,6 +6255,183 @@ static int test_ticket_callbacks(int tst)
...
+ * Test 4: TLSv1.2, no ticket key callback, ticket, no renewal
+ * Test 5: TLSv1.3, no ticket key callback, ticket, no renewal
+ * Test 6: TLSv1.2, no ticket key callback, ticket, renewal
+ * Test 7: TLSv1.3, no ticket key callback, ticket, renewal
+ * Test 8: TLSv1.2, ticket key callback, ticket, no renewal
+ * Test 9: TLSv1.3, ticket key callback, ticket, no renewal
+ * Test 10: TLSv1.2, ticket key callback, ticket, renewal
+ * Test 11: TLSv1.3, ticket key callback, ticket, renewal
+ */
+static int test_ticket_callbacks(int tst)
```
mattcaswell: Surely `test_ticket_callbacks` and `test_old_ticket_callbacks` can be combined in some way. Virtually of the code is identical. Tests 0-7 *are* identical aren't they (no ticket key callback)? Probably we should have just one function with tests 12-15 added which use the new callback instead.

apps/lib/s_cb.c
paulidale: Good point.  I've used the `SN_` form inside libcrypto and continued doing it here.  Since this isn't libcrypto, it makes sense to just use the string form directly.

mattcaswell: I wouldn't use it in libcrypto either

apps/lib/s_cb.c
paulidale: I think so, although I don't think it belongs in this PR which is about deprecation.

The change from `HMAC(...)` to fetch, ctx_new, set_param, init, update, final, ctx_free and free isn't ideal.  Getting an `EVP_MAC(...)` call correct might result in some discussion.

test/sslapitest.c
paulidale: I thought about this but got worried about the #ifdefs.  I'll revisit the idea.

apps/lib/s_cb.c
mattcaswell: Yes - agreed. This was a side note.

paulidale: I've raised an issue #10939 so we don't lose this.

apps/lib/s_cb.c
paulidale: All the `SN_` instanced are changed.

CHANGES
paulidale: I added this as a separate entry in CHANGES so it doesn't get lost in the noise of the HMAC deprecation note.

test/sslapitest.c
paulidale: This ended up pretty clean.

paulidale:
I think that's all the comments addressed.

test/bad_dtls_test.c
```diff
@@ -300,18 +304,26 @@ static int send_record(BIO *rbio, unsigned char type, uint64_t seqnr,
     memcpy(enc, msg, len);
 
     /* Append HMAC to data */
-    ctx = HMAC_CTX_new();
-    HMAC_Init_ex(ctx, mac_key, 20, EVP_sha1(), NULL);
-    HMAC_Update(ctx, epoch, 2);
-    HMAC_Update(ctx, seq, 6);
-    HMAC_Update(ctx, &type, 1);
-    HMAC_Update(ctx, ver, 2); /* Version */
+    hmac = EVP_MAC_fetch(NULL, SN_hmac, NULL);
```
mattcaswell: An "SN_hmac" remains here

test/bad_dtls_test.c
```diff
@@ -300,18 +304,26 @@ static int send_record(BIO *rbio, unsigned char type, uint64_t seqnr,
...
+    ctx = EVP_MAC_CTX_new(hmac);
+    EVP_MAC_free(hmac);
+    params[0] = OSSL_PARAM_construct_utf8_string(OSSL_MAC_PARAM_DIGEST,
+                                                 SN_sha1, 0);
```
mattcaswell: And an "SN_sha1" here

test/bad_dtls_test.c
paulidale: Bummer.  Fixed now.

paulidale:
Ping?

include/openssl/tls1.h
```diff
@@ -327,9 +327,14 @@ __owur int SSL_check_chain(SSL *s, X509 *x, EVP_PKEY *pk, STACK_OF(X509) *chain)
 # define SSL_CTX_get_tlsext_status_type(ssl) \
         SSL_CTX_ctrl(ssl,SSL_CTRL_GET_TLSEXT_STATUS_REQ_TYPE,0,NULL)
 
-# define SSL_CTX_set_tlsext_ticket_key_cb(ssl, cb) \
+# ifndef OPENSSL_NO_DEPRECATED_3_0
+#  define SSL_CTX_set_tlsext_ticket_key_cb(ssl, cb) \
         SSL_CTX_callback_ctrl(ssl,SSL_CTRL_SET_TLSEXT_TICKET_KEY_CB,\
                 (void (*)(void))cb)
```
levitte: Is `SSL_CTX_callback_ctrl` used for anything else?

levitte:
> Ping?

This does look good enough to me, but I would like to see @mattcaswell pitch in one last time as well before approval.

include/openssl/tls1.h
paulidale: It's used a few times in `tls1.h` and `ssl_lib.c`.

paulidale:
Merged to master.  Thanks for the positive input.

PR #10826, opened 2020-01-13
Add DH key exchange to fips provider

slontis:
##### Checklist

- [ ] documentation is added or updated
- [X] tests are added or updated

PR #10837, opened 2020-01-14
CRYPTO: Remove support for ex_data fields when building the FIPS module

### LINKED ISSUE #10835: Let's not have ex_data within the FIPS provider module
Recent changes in `crypto/dsa/` showed some key creating functions that take an `OPENSSL_CTX` as argument...  for the sole reason of supporting `ex_data` in the DSA structure.  Interestingly enough, the functions to get or set ex_data are not included in the compilation of the FIPS module, making that DSA field useless.

I would like to suggest that `ex_data` fields in diverse structures (such as low level key-pairs) are simply unsupported.  They exist purely for the sake of the application, which will never reach that deeply into the entrails of any provider, making `ex_data` support of that sort pretty moot.

**Note** There are still `CRYPTO_EX_DATA` items that need to remain supported, as the whole `OPENSSL_CTX` construct relies on that.  That's not what I'm talking about here.

levitte:
These fields are purely application data, and applications don't reach
into the bowels of the FIPS module, so these fields are never used
there.

Fixes #10835

### DIFF
--- a/crypto/dsa/dsa_lib.c
+++ b/crypto/dsa/dsa_lib.c
@@ -19,11 +19,6 @@
 
 #ifndef FIPS_MODE
 
-DSA *DSA_new(void)
-{
-    return DSA_new_method(NULL);
-}
-
 int DSA_set_ex_data(DSA *d, int idx, void *arg)
 {
     return CRYPTO_set_ex_data(&d->ex_data, idx, arg);
@@ -215,8 +210,10 @@ static DSA *dsa_new_method(OPENSSL_CTX *libctx, ENGINE *engine)
 
     ret->flags = ret->meth->flags & ~DSA_FLAG_NON_FIPS_ALLOW;
 
+#ifndef FIPS_MODE
     if (!crypto_new_ex_data_ex(libctx, CRYPTO_EX_INDEX_DSA, ret, &ret->ex_data))
         goto err;
+#endif
 
     if ((ret->meth->init != NULL) && !ret->meth->init(ret)) {
         DSAerr(DSA_F_DSA_NEW_METHOD, ERR_R_INIT_FAIL);
@@ -235,9 +232,9 @@ DSA *DSA_new_method(ENGINE *engine)
     return dsa_new_method(NULL, engine);
 }
 
-DSA *dsa_new(OPENSSL_CTX *libctx)
+DSA *DSA_new(void)
 {
-    return dsa_new_method(libctx, NULL);
+    return DSA_new_method(NULL);
 }
 
 void DSA_free(DSA *r)
@@ -259,7 +256,9 @@ void DSA_free(DSA *r)
     ENGINE_finish(r->engine);
 #endif
 
+#ifndef FIPS_MODE
     CRYPTO_free_ex_data(CRYPTO_EX_INDEX_DSA, r, &r->ex_data);
+#endif
 
     CRYPTO_THREAD_lock_free(r->lock);
 
--- a/crypto/rand/rand_local.h
+++ b/crypto/rand/rand_local.h
@@ -308,8 +308,10 @@ struct rand_drbg_st {
     size_t seedlen;
     DRBG_STATUS state;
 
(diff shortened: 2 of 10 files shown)


crypto/rand/rand_local.h
```diff
@@ -308,8 +308,10 @@ struct rand_drbg_st {
     size_t seedlen;
     DRBG_STATUS state;
 
+#ifndef FIPS_MODE
     /* Application data, mainly used in the KATs. */
```
paulidale: Doesn't this comment cause alarm?

paulidale:
I'm wondering why?

Why not leave them in the FIPS provider albeit unused?

crypto/rand/rand_local.h
levitte: I was wondering about that, but having looked around, I only saw that used in test programs (i.e applications), so. 

They can always be re-enabled selectively later on, if actually needed.

levitte:
It created havoc when implementing key copying in the dsa keymgmt, because suddenly I was *required* to pass a OPENSSL_CTX for no real reason.  This would have meant some pretty heavy refactoring of the dsa keymgmt. 

Incidently, I noticed that EC has already disabled ex_data, so it seems someone had already been thinking the same thing before.

slontis (changes requested):
I think this needs a discussion..
Not sure what the side effects of this will be...
libctx is being passed around for a few things in the fips module currently.
ex_data is one of them. BN_CTX and RAND also need it.
@mattcaswell ?

crypto/dsa/dsa_lib.c
```diff
@@ -235,9 +232,9 @@ DSA *DSA_new_method(ENGINE *engine)
     return dsa_new_method(NULL, engine);
 }
 
-DSA *dsa_new(OPENSSL_CTX *libctx)
```
slontis: https://github.com/openssl/openssl/pull/10826 is also doing this.

levitte:
>  libctx is being passed around for a few things in the fips module currently.

Yes?  But does that mean that the *keys* need to have an ex_data field?

levitte:
> BN_CTX and RAND also need it.

The only place I can see that being affected, *specifically for keymgmt*, is with key generation, and if you look in #10289, the key generation functions receive a `provctx`, which fixes that.

levitte:
So, uhmmm, ping ?

mattcaswell:
> I think this needs a discussion..
Not sure what the side effects of this will be...
libctx is being passed around for a few things in the fips module currently.
ex_data is one of them. BN_CTX and RAND also need it.

I see no problem with this.

mattcaswell (approved):

richsalz:
To this non-member, it looks like you're kinda rushing things.  I hope folks half-a-world away feel their concerns are being addressed.

mattcaswell:
> To this non-member, it looks like you're kinda rushing things. I hope folks half-a-world away feel their concerns are being addressed.

Not sure what concerns you mean. This all seems fairly innocuous to me.

richsalz:
Well, @slontis has the 24hour period to raise issues.  But seems like two folks in the same region wrote code, pinged, answered a query, approved all while he slept :)  Shrug.

mattcaswell:
> Well, @slontis has the 24hour period to raise issues. But seems like two folks in the same region wrote code, pinged, answered a query, approved all while he slept :) Shrug.

That's the whole point of the 24hr period...

levitte:
FYI, there was video call this morning, involving @mattcaswell, @paulidale, @slontis, @romen and myself, where this PR was discussed among other things.

slontis (approved):

levitte:
Merged.

a3327784d9526fd69649089ea5c78a55906e9b2c CRYPTO: Remove support for ex_data fields when building the FIPS module
9ec7b6ad12529d2ab05b0b18fdabe1b12123f9d5 PROV: Adapt the DSA keymgmt implementation to no ex_fields

PR #10758, opened 2020-01-04
EVP: Fix method to determine if a PKEY is legacy or not

levitte:
For the implementation of EVP_PKEY_CTX_new(), we determined if an
EVP_PKEY wass legacy or not by looking at 'pkey->pkey.ptr'.  It turns
out that this code could get an unassigned EVP_PKEY, with that pointer
being NULL, and the determination proven incorrect.

The check now looks at 'pkey->ameth' instead.

Fixes #10704

PR #10839, opened 2020-01-14
Restoring correct check for legacy PKEY

beldmit:
The fix inroduced in #10758 was rolled back by accident.
Restoring it.

##### Checklist

### DIFF
--- a/crypto/evp/pmeth_lib.c
+++ b/crypto/evp/pmeth_lib.c
@@ -150,7 +150,7 @@ static EVP_PKEY_CTX *int_ctx_new(OPENSSL_CTX *libctx,
      * If the key doesn't contain anything legacy, then it must be provided,
      * so we extract the necessary information and use that.
      */
-    if (pkey != NULL && pkey->pkey.ptr == NULL) {
+    if (pkey != NULL && pkey->ameth == NULL) {
         /* If we have an engine, something went wrong somewhere... */
         if (!ossl_assert(e == NULL))
             return NULL;


beldmit:
Found here:
https://github.com/gost-engine/engine/issues/201

@levitte, could you please approve?

beldmit:
BTW, could we add at least basic gost-engine tests to travis to avoid such incidents?

levitte:
> BTW, could we add at least basic gost-engine tests to Travis to avoid such incidents?

I thought there was a PR for this already...  did I just dream that up?

Anyway, I see no reason why not, that would be another one of those external tests, with gost as a submodule.

levitte:
Oh, hold on...  This sort of mistake will solve itself as soon as we have EVP_PKEYs without associated legacy keys, i.e. as soon as we get key generation in place (or make #10797 an actual change...  I'm pondering having some kind of test option or configuration option for this, as it's quite useful to keep us on our toes :wink:)

beldmit:
Sorry, I don't get your point.

If engines should work with 3.0 (maybe with reasonable changes), the GOST tests make sense to check whether the changes in main tree are breaking or not. I do not know the provider code so I'm not sure that key generation in place fixes the problem we can see in GOST tests.

levitte:
The point is that "recognizing legacy keys" also means "recognizing non-legacy keys", so either way we look at it, it's should cause failure. It's reasonable to think that the current lack of support for EVP_PKEYs without the legacy bits was hiding this mistake just as much as the lack of a gost test.

mouse07410:
Can I ask when/if this PR would be merged?

beldmit:
> Can I ask when/if this PR would be merged?

Tomorrow morning, when 24-hours quarantine expires.

mouse07410:
@levitte you have a point. 

IMHO, one lesson the Dev Team should take from this that a PR should be constrained to one issue, making it easier to spot outliers such as this. E.g., it seems unclear (to me, at least) why a DSA-reworking PR should mess with `int_ctx_new()` - it should've raised some alarm...

mattcaswell:
> IMHO, one lesson the Dev Team should take from this that a PR should be constrained to one issue, making it easier to spot outliers such as this. E.g., it seems unclear (to me, at least) why a DSA-reworking PR should mess with int_ctx_new() - it should've raised some alarm...

The PR in question was about moving the DSA code into the FIPS provider. Part of the scope of that was moving the code in crypto/evp/pmeth_lib.c into the FIPS provider too. You need that, otherwise you cannot make DSA calls via EVP from inside the FIPS module itself (required for self-testing for example).

slontis:
Apologies for the merge error.

beldmit:
Merged. Thanks!

PR #10840, opened 2020-01-14
param_bld: add a padded BN call.

paulidale:
To aviod leaking size information when passing private value using the
OSSL_PARAM builder, a padded BN call is required.

### DIFF
--- a/crypto/param_build.c
+++ b/crypto/param_build.c
@@ -138,21 +138,30 @@ int ossl_param_bld_push_double(OSSL_PARAM_BLD *bld, const char *key,
 int ossl_param_bld_push_BN(OSSL_PARAM_BLD *bld, const char *key,
                            const BIGNUM *bn)
 {
-    int sz = -1, secure = 0;
+    return ossl_param_bld_push_BN_pad(bld, key, bn,
+                                      bn == NULL ? 0 : BN_num_bytes(bn));
+}
+
+int ossl_param_bld_push_BN_pad(OSSL_PARAM_BLD *bld, const char *key,
+                               const BIGNUM *bn, size_t sz)
+{
+    int n, secure = 0;
     OSSL_PARAM_BLD_DEF *pd;
 
     if (bn != NULL) {
-        sz = BN_num_bytes(bn);
-        if (sz < 0) {
-            CRYPTOerr(CRYPTO_F_OSSL_PARAM_BLD_PUSH_BN,
-                      CRYPTO_R_ZERO_LENGTH_NUMBER);
+        n = BN_num_bytes(bn);
+        if (n < 0) {
+            CRYPTOerr(0, CRYPTO_R_ZERO_LENGTH_NUMBER);
+            return 0;
+        }
+        if (sz < (size_t)n) {
+            CRYPTOerr(0, CRYPTO_R_TOO_SMALL_BUFFER);
             return 0;
         }
         if (BN_get_flags(bn, BN_FLG_SECURE) == BN_FLG_SECURE)
             secure = 1;
     }
-    pd = param_push(bld, key, sz, sz >= 0 ? sz : 0,
-                    OSSL_PARAM_UNSIGNED_INTEGER, secure);
+    pd = param_push(bld, key, sz, sz, OSSL_PARAM_UNSIGNED_INTEGER, secure);
     if (pd == NULL)
         return 0;
     pd->bn = bn;
--- a/doc/internal/man3/ossl_param_bld_init.pod
+++ b/doc/internal/man3/ossl_param_bld_init.pod
@@ -8,9 +8,9 @@ ossl_param_bld_push_long, ossl_param_bld_push_ulong,
 ossl_param_bld_push_int32, ossl_param_bld_push_uint32,
 ossl_param_bld_push_int64, ossl_param_bld_push_uint64,
 ossl_param_bld_push_size_t, ossl_param_bld_push_double,
-ossl_param_bld_push_BN, ossl_param_bld_push_utf8_string,
-ossl_param_bld_push_utf8_ptr, ossl_param_bld_push_octet_string,
-ossl_param_bld_push_octet_ptr
+ossl_param_bld_push_BN, ossl_param_bld_push_BN_pad,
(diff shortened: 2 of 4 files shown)


doc/internal/man3/ossl_param_bld_init.pod
```diff
@@ -90,6 +92,14 @@ will also be securely allocated.
 The I<bn> argument is stored by reference and the underlying BIGNUM object
 must exist until after ossl_param_bld_to_param() has been called.
 
+ossl_param_bld_push_BN_pad() is a function that will create an OSSL_PARAM object
+that holds the specified BIGNUM I<bn>.
+The object will be padded to occupy exactly I<sz> bytes.
```
levitte: What happens when *sz* has a smaller value than the size of the bignum?  What happens when *sz* is zero?

crypto/param_build.c
```diff
@@ -159,6 +159,33 @@ int ossl_param_bld_push_BN(OSSL_PARAM_BLD *bld, const char *key,
...
+    int n, secure = 0;
+    OSSL_PARAM_BLD_DEF *pd;
+
+    if (bn != NULL) {
+        n = BN_num_bytes(bn);
+        if (n < 0) {
+            CRYPTOerr(CRYPTO_F_OSSL_PARAM_BLD_PUSH_BN,
+                      CRYPTO_R_ZERO_LENGTH_NUMBER);
+            return 0;
+        } else if (sz < (size_t)n) {
```
richsalz: Why do you have an "else after return" here, but not one at line 179?  Please remove this or (yuk) add it there.

crypto/param_build.c
```diff
@@ -159,6 +159,33 @@ int ossl_param_bld_push_BN(OSSL_PARAM_BLD *bld, const char *key,
     return 1;
 }
 
+int ossl_param_bld_push_BN_pad(OSSL_PARAM_BLD *bld, const char *key,
+                               const BIGNUM *bn, size_t sz)
+{
+    int n, secure = 0;
+    OSSL_PARAM_BLD_DEF *pd;
+
+    if (bn != NULL) {
```
richsalz: What are the semantics of pushing a NULL bn?  Isn't that an error?

levitte: It's perfectly fine. For example, dh keys may have a NULL Q, and even though silly to pass that, it makes it easier to call this function without surrounding it with extre checks (that you might forget to do).

It may also be that when getting parameters, BNs can get allocated automatically as needed on the owning side.

richsalz: ok, tnx.

slontis:
make update required..

crypto/param_build.c
```diff
@@ -159,6 +159,33 @@ int ossl_param_bld_push_BN(OSSL_PARAM_BLD *bld, const char *key,
...
+        n = BN_num_bytes(bn);
+        if (n < 0) {
```
slontis: Is n = 0 ok?

test/param_build_test.c
```diff
@@ -204,7 +205,7 @@ static int template_static_params_test(int n)
     if (!TEST_true(ossl_param_bld_push_uint(&bld, "i", 6))
         || !TEST_ptr(bn = (n & 1) == 0 ? BN_new() : BN_secure_new())
         || !TEST_true(BN_set_word(bn, 1337))
-        || !TEST_true(ossl_param_bld_push_BN(&bld, "bn", bn))
+        || !TEST_true(ossl_param_bld_push_BN_pad(&bld, "bn", bn, bn_bytes))
```
slontis: maybe add negative tests for sz = 0, n = 0,  sz < n

doc/internal/man3/ossl_param_bld_init.pod
paulidale: Added a note indicating an error will occur when insufficient space is specified.

crypto/param_build.c
paulidale: Yes.  Zero can have zero size I think.

richsalz: A padded BN that can only handle zero seems not useful.  I would make zero-size an error.

paulidale:
Feedback addressed.

crypto/param_build.c
paulidale: Too late in the evening :)

crypto/param_build.c
paulidale: I'd expect `ossl_param_bld_push_BN_pad(&bld, "bn", bn, BN_num_bytes(bn))` to always work, this suggestion would prevent that.

richsalz: I would not expect to make a call like that.  I would expect to compare a couple of BIGNUM's and use the same size padding for all BN params.  Why have a "padded" type if you're going to leak the size of zero?

paulidale:
Ping reviewer.

richsalz:
I wish there were a resolution to https://github.com/openssl/openssl/pull/10840#discussion_r366619100  :(

paulidale:
I think we disagree.
We'll see what any reviewers say -- it's an `=` sign change to the code.

levitte:
> I wish there were a resolution to https://github.com/openssl/openssl/pull/10840#discussion_r366619100  :(

There's nothing to resolve. That usage example is just that, but does not affect this PR per se. 
(and I agree that it wasn't an example of passing a BIGNUM with padding, it's an example of passing a BIGNUM without)

levitte:
I expected to see this rewrite of `ossl_param_bld_push_BN(OSSL_PARAM_BLD`:

``` C
int ossl_param_bld_push_BN(OSSL_PARAM_BLD *bld, const char *key,
                           const BIGNUM *bn)
{
    return ossl_param_bld_push_BN_pad(bld, key, bn, BN_num_bytes(bn));
}
```

paulidale:
That rewrite isn't quite right.  A NULL bn argument is permitted by that function.

levitte:
Er, what? Looking at the code, they seem to treat a NULL bn exactly the same way...

levitte:
Sorry, now I see. I've update my proposed rewrite

paulidale:
I've made the change.

levitte:
Please check the CIs before merging

paulidale:
Merged to master.  Thanks for the input.

PR #10845, opened 2020-01-15
Modify EVP_CIPHER_is_a() and EVP_MD_is_a() to handle legacy methods too

levitte:
These functions would only handle provided methods, but there are
cases where the caller just passes along a received method without
knowing the underlying method tech, so might pass along a legacy
method.  We therefore need to have them handle this case as well so
they don't cause any unnecessary surprises.

### DIFF
--- a/crypto/evp/evp_fetch.c
+++ b/crypto/evp/evp_fetch.c
@@ -444,11 +444,18 @@ const char *evp_first_name(OSSL_PROVIDER *prov, int name_id)
     return ossl_namemap_num2name(namemap, name_id, 0);
 }
 
-int evp_is_a(OSSL_PROVIDER *prov, int number, const char *name)
+int evp_is_a(OSSL_PROVIDER *prov, int number,
+             const char *legacy_name, const char *name)
 {
+    /*
+     * For a |prov| that is NULL, the library context will be NULL, which
+     * is fine, because that's the only place legacy stuff is relevant.
+     */
     OPENSSL_CTX *libctx = ossl_provider_library_context(prov);
     OSSL_NAMEMAP *namemap = ossl_namemap_stored(libctx);
 
+    if (prov == NULL)
+        number = ossl_namemap_name2num(namemap, legacy_name);
     return ossl_namemap_name2num(namemap, name) == number;
 }
 
--- a/crypto/evp/evp_fetch.c
+++ b/crypto/evp/evp_fetch.c
@@ -448,8 +448,7 @@ int evp_is_a(OSSL_PROVIDER *prov, int number,
              const char *legacy_name, const char *name)
 {
     /*
-     * For a |prov| that is NULL, the library context will be NULL, which
-     * is fine, because that's the only place legacy stuff is relevant.
+     * For a |prov| that is NULL, the library context will be NULL
      */
     OPENSSL_CTX *libctx = ossl_provider_library_context(prov);
     OSSL_NAMEMAP *namemap = ossl_namemap_stored(libctx);
--- a/test/namemap_internal_test.c
+++ b/test/namemap_internal_test.c
@@ -109,7 +109,7 @@ static int test_cipherbyname(void)
 }
 
 /*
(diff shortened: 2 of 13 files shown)


test/namemap_internal_test.c
```diff
@@ -108,12 +108,53 @@ static int test_cipherbyname(void)
...
+ */
+static int test_cipher_is_a(void)
+{
+    EVP_CIPHER *fetched = EVP_CIPHER_fetch(NULL, "AES-256-CCM", NULL);
+    int rv = 1;
+
+    if (!TEST_ptr_ne(fetched, NULL))
+        return 0;
+    if (!TEST_true(EVP_CIPHER_is_a(fetched, "id-aes256-CCM")))
+        rv = 0;
```
slontis: may as well add a negative test here also.

test/namemap_internal_test.c
```diff
@@ -108,12 +108,53 @@ static int test_cipherbyname(void)
     return 1;
 }
 
+/*
+ * Test that EVP_CIPHER_is_a() responds appropriately, even for cipherss that
```
slontis: make it less snakey by removing a s

slontis (commented):
I am not quite sure I understand why the legacy name cant just be put into the provider tables?

levitte:
> I am not quite sure I understand why the legacy name cant just be put into the provider tables?

Ah, it seems I was a bit unclear.

The main point of this exercise is that legacy EVP_CIPHERs and EVP_MDs have an unset `name_id`, so it can't be used to check that it name that's tested has the same number.

Or did you mean that we would insert the legacy *NID* into the namemap? Yup, that was my first thought too, but that turned out to be more difficult, I ran into corner cases where names that weren't pre-populated didn't get the legacy NID, even though they were new aliases for pre-populated names, so evp_is_a ended up giving the wrong answer... and the code to do all that became quite ugly, so...

test/namemap_internal_test.c
levitte: Done

test/namemap_internal_test.c
levitte: Done

crypto/evp/evp_fetch.c
```diff
@@ -444,11 +444,18 @@ const char *evp_first_name(OSSL_PROVIDER *prov, int name_id)
     return ossl_namemap_num2name(namemap, name_id, 0);
 }
 
-int evp_is_a(OSSL_PROVIDER *prov, int number, const char *name)
+int evp_is_a(OSSL_PROVIDER *prov, int number,
+             const char *legacy_name, const char *name)
 {
+    /*
+     * For a |prov| that is NULL, the library context will be NULL, which
+     * is fine, because that's the only place legacy stuff is relevant.
```
mattcaswell: I don't think this statement is true. I fully expect "legacy" ciphers/digests to be used in conjunction with non-default OPENSSL_CTXs. AFAIK there is nothing in the documentation suggesting you shouldn't be allowed to do that.

mattcaswell (changes requested):

crypto/evp/evp_fetch.c
levitte: Er...  so how do you associate legacy cipher/digests with a non-default OPENSSL_CTX?

levitte: I'm at a loss what to think or how to comment.  Ideas?

levitte: For everyone's information, @mattcaswell and I have a private chat where we concluded that this is not ideal, but ideal is hard to come by, so we're resorting to accepting this situation for legacy ciphers and just making sure it's documented.

Done

levitte:
Ping

mattcaswell (approved):

levitte:
Merged.

e4a1d0230016d090ba78bc7092384315f85b0e72 Modify EVP_CIPHER_is_a() and EVP_MD_is_a() to handle legacy methods too

PR #10846, opened 2020-01-15
Move the stored namemap pre-population to namemap construction

levitte:
Prepopulation of the stored namemap from the legacy method object
database happened on first EVP fetch.  However, there are moments when
that prepopulation needs to happen even though no fetching has been
performed yet.  We therefore move pre-population to happen when the
namemap is constructed.

### DIFF
--- a/crypto/core_namemap.c
+++ b/crypto/core_namemap.c
@@ -89,38 +89,6 @@ static const OPENSSL_CTX_METHOD stored_namemap_method = {
  * =============
  */
 
-OSSL_NAMEMAP *ossl_namemap_stored(OPENSSL_CTX *libctx)
-{
-    return openssl_ctx_get_data(libctx, OPENSSL_CTX_NAMEMAP_INDEX,
-                                &stored_namemap_method);
-}
-
-OSSL_NAMEMAP *ossl_namemap_new(void)
-{
-    OSSL_NAMEMAP *namemap;
-
-    if ((namemap = OPENSSL_zalloc(sizeof(*namemap))) != NULL
-        && (namemap->lock = CRYPTO_THREAD_lock_new()) != NULL
-        && (namemap->namenum =
-            lh_NAMENUM_ENTRY_new(namenum_hash, namenum_cmp)) != NULL)
-        return namemap;
-
-    ossl_namemap_free(namemap);
-    return NULL;
-}
-
-void ossl_namemap_free(OSSL_NAMEMAP *namemap)
-{
-    if (namemap == NULL || namemap->stored)
-        return;
-
-    lh_NAMENUM_ENTRY_doall(namemap->namenum, namenum_free);
-    lh_NAMENUM_ENTRY_free(namemap->namenum);
-
-    CRYPTO_THREAD_lock_free(namemap->lock);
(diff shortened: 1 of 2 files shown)


crypto/core_namemap.c
```diff
@@ -335,3 +303,107 @@ int ossl_namemap_add_names(OSSL_NAMEMAP *namemap, int number,
 
     return number;
 }
+
+/*-
```
paulidale: `/*-` to open a multi-line comment?  I thought it was just `/*`

crypto/core_namemap.c
```diff
@@ -335,3 +303,107 @@ int ossl_namemap_add_names(OSSL_NAMEMAP *namemap, int number,
...
+
+/* Creates an initial namemap with names found in the legacy method db */
+static void get_legacy_evp_names(const char *main_name, const char *alias,
+                                 void *arg)
+{
+    int main_id = ossl_namemap_add_name(arg, 0, main_name);
+
+    /*
+     * We could check that the returned value is the same as main_id,
+     * but since this is a void function, there's no sane way to report
```
paulidale: This is an internal function, it need not be `void`.  The calling functions are but isn't it early enough to change their behaviour?  [ or does it cascade up the call chain ]

crypto/core_namemap.c
```diff
@@ -335,3 +303,107 @@ int ossl_namemap_add_names(OSSL_NAMEMAP *namemap, int number,
...
+{
+    OSSL_NAMEMAP *namemap =
+        openssl_ctx_get_data(libctx, OPENSSL_CTX_NAMEMAP_INDEX,
+                             &stored_namemap_method);
+
+#ifndef FIPS_MODE
+    if (namemap != NULL && ossl_namemap_empty(namemap)) {
+        /* Before pilfering, we make sure the legacy database is populated */
+        OPENSSL_init_crypto(OPENSSL_INIT_ADD_ALL_CIPHERS
+                            |OPENSSL_INIT_ADD_ALL_DIGESTS, NULL);
```
paulidale: Space after `|`.

paulidale (commented):
Superficially it looks good.  I need another pass over when I'm less tired.

crypto/core_namemap.c
levitte: The dash is an accepted character to tell the 'indent' program not to reformat the comment.

paulidale: Cool.

crypto/core_namemap.c
levitte: This is downstream from a callback for OBJ_NAME_do_all(), so the returned value is ignored either way...

paulidale: I suspected as much.  Thanks for explaining.

mattcaswell (approved):
LGTM assuming @pauliidale's nit is fixed

crypto/core_namemap.c
levitte: Fixed

levitte:
Merged.

437ad983c6ca254cd1d4367fb1d68338c82e8ee3 Move the stored namemap pre-population to namemap construction

PR #10848, opened 2020-01-15
Update SSL_CTX_sess_set_new_cb(3) docs for refcounts

kaduk:
The existing documentation for the new-session callback was unclear
about the requirements on the callback with respect to reference-handling
of the session object being created.  Be more explicit about the
(non-)requirements on the callback code for "success" (1) and "ignore"
(0) return values.


kaduk:
Yes, for 1.1.1 as well; I forgot to set the labels.

kaduk:
Pushed to master and 1.1.1; closing

PR #10849, opened 2020-01-15
Build file templates: Use explicit files instead of $< or $? for pods

### LINKED ISSUE #10817: freebsd build failed on current openssl master branch
on gost-engine we use freebsd within cirrus-ci

https://github.com/gost-engine/engine/runs/385094611

we build current openssl master branch.
recently it started to fail with

```
/usr/local/bin/perl "-I." -Mconfigdata "util/dofile.pl"  "-oMakefile" util/shlib_wrap.sh.in > "util/shlib_wrap.sh"
chmod a+x util/shlib_wrap.sh
pod2man --name=CA.PL --section=1 --center=OpenSSL  --release=3.0.0-dev   > doc/man/man1/CA.pl.1
pod2man: unable to format standard input
*** Error code 1

Stop.
make: stopped in /tmp/cirrus-ci-build/openssl
```

looks like error on openssl side.
also, what if enable at least some basic cirrus-ci/freebsd for openssl itself ?

levitte:
When generating html or manpages from POD files, we used $< or $? to
get the file name to process.  It turns out, though, that some make
implementations only define $< with implicit rules, so its expansion
remains empty in explicit rules.  $? is a fine replacement, but only
as long as we have one dependency, so it may cause problems in the
future.

The final solution seems to be to use explicit POD file names
instead.  That leaves no doubts.

Fixes #10817

### DIFF
--- a/Configurations/unix-Makefile.tmpl
+++ b/Configurations/unix-Makefile.tmpl
@@ -1188,7 +1188,7 @@ reconfigure reconf:
 	  return <<"EOF";
 $args{src}: $pod
 	pod2html "--podroot=\$(SRCDIR)/doc" --htmldir=.. \\
-		 --podpath=man1:man3:man5:man7 "--infile=$pod" "--title=$title" \\
+		 --podpath=man1:man3:man5:man7 --infile=$pod "--title=$title" \\
 	| \$(PERL) -pe 's|href="http://man\\.he\\.net/(man\\d/[^"]+)(?:\\.html)?"|href="../\$1.html|g;' \\
 	> \$\@
 EOF
@@ -1199,7 +1199,7 @@ EOF
 	  return <<"EOF";
 $args{src}: $pod
 	pod2man --name=$name --section=$section --center=OpenSSL \\
-		--release=\$(VERSION) "$pod" \\
+		--release=\$(VERSION) $pod \\
 	> \$\@
 EOF
       } elsif (platform->isdef($args{src})) {
--- a/Configurations/unix-Makefile.tmpl
+++ b/Configurations/unix-Makefile.tmpl
@@ -1184,20 +1184,22 @@ reconfigure reconf:
 
       if ($args{src} =~ /\.html$/) {
 	  my $title = basename($args{src}, ".html");
+	  my $pod = $args{generator}->[0];
 	  return <<"EOF";
-$args{src}: $args{generator}->[0]
+$args{src}: $pod
 	pod2html "--podroot=\$(SRCDIR)/doc" --htmldir=.. \\
-		 --podpath=man1:man3:man5:man7 "--infile=\$<" "--title=$title" \\
+		 --podpath=man1:man3:man5:man7 "--infile=$pod" "--title=$title" \\
 	| \$(PERL) -pe 's|href="http://man\\.he\\.net/(man\\d/[^"]+)(?:\\.html)?"|href="../\$1.html|g;' \\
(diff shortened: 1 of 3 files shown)


mattcaswell (dismissed):
LGTM

richsalz:
Probably a separate PR, but do you think it's a good idea to put the pod2html and the post-proc patchup script into one single new perl program?

levitte:
That's actually a good idea. I know that it's possible to make hooks in a perl program for that sort of thing... But yeah, separate PR.

kaduk (dismissed):
Code looks good and verified to fix the freebsd build.
(We are perhaps a bit inconsistent about when we quote and when we don't quote ```$pod```, though.)

chipitsine:
I tested , looks different

```
$ ./config shared -d --prefix=${PREFIX} --openssldir=${PREFIX} && make all install_sw
Operating system: amd64-whatever-freebsd
You need Perl 5.
$ 
$ 
$ 
$ 
$ sudo sh
# pkg info | grep perl
perl5-5.30.1                   Practical Extraction and Report Language
perl5.26-5.26.3                Practical Extraction and Report Language

```

Configurations/unix-Makefile.tmpl
```diff
@@ -1184,20 +1184,22 @@ reconfigure reconf:
...
       } elsif ($args{src} =~ /\.(\d)$/) {
 	  my $section = $1;
 	  my $name = uc basename($args{src}, ".$section");
+	  my $pod = $args{generator}->[0];
 	  return <<"EOF";
-$args{src}: $args{generator}->[0]
+$args{src}: $pod
 	pod2man --name=$name --section=$section --center=OpenSSL \\
-		--release=\$(VERSION) \$< \\
+		--release=\$(VERSION) "$pod" \\
```
richsalz: why are quotes needed here, and not on line 1200?

kaduk: They aren't -- if ```$pod``` has spaces the dependency will be broken, and this is a relative path under our control anyway.
I guess I should have used more words in my previous comment.

kaduk:
@chipitsine what about ```type perl``` or ```echo $PATH``` -- the two data points are not commensurate, as one talks about what is (not) available at runtime and the other what is installed somewhere on the system

Configurations/unix-Makefile.tmpl
levitte: I added the quotes without even thinking about it.

However, "under our control" isn't *entirely* true.  In an out-of-source build, the relative path to the source has components that we don't control.

levitte:
`perl -v` gives a better answer on the perl that you *actually* use than any package informatoin

Configurations/unix-Makefile.tmpl
richsalz: then you need the quotes in the dependency line, too.

richsalz: So either add quotes or remove them, right?  That's *my* point.

levitte: I find that overly picky (the quotes are *really* harmless), but fine, changed.

levitte:
Re-approval needed, please

kaduk (approved):
reconfirm

levitte:
Merged.

98706c5a8cb2b0a1649add98125289db7da04861 Build file templates: Use explicit files instead of $< or $? for pods

PR #10850, opened 2020-01-15
EVP: Add evp_pkey_make_provided() and refactor around it

levitte:
The code to ensure that an EVP_PKEY is exported to providers is
repeated all over the place, enough that copying it again has the
usual future hazards with code copying.

Instead, we refactor that code into one function,
evp_pkey_make_provided(), and make sure to use that everywhere.
It relies on the creation of EVP_PKEY_CTX to figure out facts about
the input key, should it need to.

### DIFF
--- a/crypto/evp/exchange.c
+++ b/crypto/evp/exchange.c
@@ -164,6 +164,8 @@ int EVP_PKEY_derive_init(EVP_PKEY_CTX *ctx)
     int ret;
     void *provkey = NULL;
     EVP_KEYEXCH *exchange = NULL;
+    EVP_KEYMGMT *tmp_keymgmt = NULL;
+    const char *supported_exch = NULL;
 
     if (ctx == NULL) {
         EVPerr(0, EVP_R_OPERATION_NOT_SUPPORTED_FOR_THIS_KEYTYPE);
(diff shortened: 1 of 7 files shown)


levitte:
I have upcoming PRs that will rely on this function.  I *really* didn't want to copy that same big old block of code again.

levitte:
Oh, and #10807 is being changed to use this function as well

crypto/evp/exchange.c
```diff
@@ -176,33 +178,33 @@ int EVP_PKEY_derive_init(EVP_PKEY_CTX *ctx)
...
-            EVP_KEYMGMT_fetch(ctx->libctx, ctx->keytype, ctx->propquery);
-    if (ctx->keymgmt != NULL) {
-        const char *supported_exch = NULL;
+    /* Ensure that the key is provided.  If not, go legacy */
+    tmp_keymgmt = ctx->keymgmt;
+    provkey = evp_pkey_make_provided(ctx->pkey, ctx->libctx,
+                                     &tmp_keymgmt, ctx->propquery, 0);
+    if (provkey == NULL)
+        goto legacy;
+    EVP_KEYMGMT_up_ref(tmp_keymgmt);
```
mattcaswell: Check for failure here? Same comment for other similar occurrences in this PR.

doc/internal/man3/evp_pkey_make_provided.pod
```diff
@@ -0,0 +1,58 @@
...
+actual key.
+
+=head1 RETURN VALUES
+
+evp_pkey_make_provided() returns the provider key data that was exported if
+I<pk> was successfully provided.  Otherwise, NULL is returned.
+
+=head1 NOTES
+
+Some functions calling evp_pkey_make_provided() may have received the a const
```
mattcaswell: Typo: the a

mattcaswell (changes requested):

crypto/evp/exchange.c
levitte: Sure.  Done

doc/internal/man3/evp_pkey_make_provided.pod
levitte: Fixed

levitte:
Ping

mattcaswell (approved):

levitte:
Merged.

f6aa57741254723b0c32f0dfe1ed8ad886b43c80 EVP: Add evp_pkey_make_provided() and refactor around it

PR #10851, opened 2020-01-15
Adapt X509_PUBKEY_set() for use with provided implementations

levitte:
We do this by letting a serializer serialize the provider side key to
a DER blob formatted according to the SubjectPublicKeyInfo structure
(see RFC 5280), and deserialize it in libcrypto using the usual d2i
function.

### DIFF
--- a/crypto/x509/x_pubkey.c
+++ b/crypto/x509/x_pubkey.c
@@ -67,45 +67,41 @@ int X509_PUBKEY_set(X509_PUBKEY **x, EVP_PKEY *pkey)
     if (x == NULL)
         return 0;
 
-    if ((pk = X509_PUBKEY_new()) == NULL)
-        goto error;
+    if (pkey == NULL)
+        goto unsupported;
 
-    if (pkey != NULL) {
-        if (pkey->ameth != NULL) {
-            if (pkey->ameth->pub_encode != NULL) {
-                if (!pkey->ameth->pub_encode(pk, pkey)) {
-                    X509err(X509_F_X509_PUBKEY_SET,
-                            X509_R_PUBLIC_KEY_ENCODE_ERROR);
-                    goto error;
-                }
-            } else {
-                X509err(X509_F_X509_PUBKEY_SET, X509_R_METHOD_NOT_SUPPORTED);
+    if (pkey->ameth != NULL) {
+        if ((pk = X509_PUBKEY_new()) == NULL)
+            goto error;
+        if (pkey->ameth->pub_encode != NULL) {
+            if (!pkey->ameth->pub_encode(pk, pkey)) {
+                X509err(X509_F_X509_PUBKEY_SET,
+                        X509_R_PUBLIC_KEY_ENCODE_ERROR);
                 goto error;
             }
-        } else if (pkey->pkeys[0].keymgmt != NULL) {
-            BIO *bmem = BIO_new(BIO_s_mem());
-            const char *serprop = "format=der,type=public";
-            OSSL_SERIALIZER_CTX *sctx =
-                OSSL_SERIALIZER_CTX_new_by_EVP_PKEY(pkey, serprop);
-
-            if (OSSL_SERIALIZER_to_bio(sctx, bmem)) {
-                const unsigned char *der = NULL;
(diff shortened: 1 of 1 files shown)


crypto/x509/x_pubkey.c
```diff
@@ -69,27 +70,51 @@ int X509_PUBKEY_set(X509_PUBKEY **x, EVP_PKEY *pkey)
     if ((pk = X509_PUBKEY_new()) == NULL)
         goto error;
 
-    if (pkey != NULL && pkey->ameth) {
-        if (pkey->ameth->pub_encode) {
-            if (!pkey->ameth->pub_encode(pk, pkey)) {
-                X509err(X509_F_X509_PUBKEY_SET,
-                        X509_R_PUBLIC_KEY_ENCODE_ERROR);
+    if (pkey != NULL) {
```
mattcaswell: This might be better as

````
if (pkey == NULL)
    goto unsupported;
````
Then all following lines can be moved to the left by one indent.

File: crypto/x509/x_pubkey.c [hunk 2]
```diff
@@ -69,27 +70,51 @@ int X509_PUBKEY_set(X509_PUBKEY **x, EVP_PKEY *pkey)
...
+            BIO *bmem = BIO_new(BIO_s_mem());
+            const char *serprop = "format=der,type=public";
+            OSSL_SERIALIZER_CTX *sctx =
+                OSSL_SERIALIZER_CTX_new_by_EVP_PKEY(pkey, serprop);
+
+            if (OSSL_SERIALIZER_to_bio(sctx, bmem)) {
+                const unsigned char *der = NULL;
+                long derlen = BIO_get_mem_data(bmem, (char **)&der);
+
+                pk = d2i_X509_PUBKEY(NULL, &der, derlen);
```
mattcaswell: What does this do if it encounters an OID it doesn't recognise? E.g. because a provider has implemented an algorithm we don't know about?

mattcaswell (changes requested):

crypto/x509/x_pubkey.c
levitte: As far as I know, it doesn't care.  An OID is an OID is an OID.  Sure, the decode *tries* to find a corresponding NID, but doesn't fail when none is to be found, the only difference being that the internal ASN1_OBJECT will have a NULL `sn` and `ln`.  The function that deal with this is:

https://github.com/openssl/openssl/blob/7b131de2bb2be8582c6dfc50bdd3cbc49e1a08be/crypto/asn1/a_object.c#L263-L332

crypto/x509/x_pubkey.c
levitte: Okie

levitte: That was a good exercise, it had me discover a probably memory leak...

crypto/x509/x_pubkey.c
```diff
@@ -80,16 +83,34 @@ int X509_PUBKEY_set(X509_PUBKEY **x, EVP_PKEY *pkey)
...
+        BIO_free(bmem);
     }
 
+    if (pk == NULL)
+        goto unsupported;
+
     X509_PUBKEY_free(*x);
     *x = pk;
     pk->pkey = pkey;
     return EVP_PKEY_up_ref(pkey);
```
mattcaswell: This is pre-existing code but looking again it seems to me that the up-ref should happen before we amend `*x` so that if it fails we can free pk and leave the function with a failure code without any parameters being changed. As it stands if you get a failure from this function you don't know whether it is because the up-ref failed and therefore *x has been updated and needs to be freed, or whether some other error occurred.

levitte: That's reasonable.  Will fix

levitte: Done

levitte:
Ping

levitte:
Ping

mattcaswell (approved):

levitte:
Merged.

e56ba0e1229d3eacb443fa4935a4da7701dfb1f3 Adapt X509_PUBKEY_set() for use with provided implementations

PR #10855, opened 2020-01-15
For all assembler scripts where it matters, recognise clang > 9.x

### LINKED ISSUE #10853: libcrypto: incorrect compiler version check in perl script crypto/bn/asm/x86_64-mont5.pl
I am using OpenSSL 1.1.1  11 Sep 2018

In file https://github.com/openssl/openssl/blob/ed086e2/crypto/bn/asm/x86_64-mont5.pl#L65

The regexpr used to match the compiler version string expects a single digit before the first dot.
This is problematic because clang master (as well as the next clang release) is `clang version 10.0.0`.
So there are two digits before the first dot.

On targets that support ADX, a more optimal call to __bn_sqrx8x_reduction is generated (instead of a call to __bn_sqr8x_reduction which doesn't use MULX/ADCX/ADOX).
The calls to the hand written assembly variant that uses ADX instructions (i.e.  __bn_sqrx8x_reduction) is only emitted if variable $addx is set. Otherwise, the code falls back to the less optimal __bn_sqr8x_reduction.

regexpr:

    /((?:^clang|LLVM) version|.*based on LLVM) ([3-9])\.([0-9]+)/)

works fine with clang 9 because there is a single digit before the dot.
It fails on clang master (clang version 10.0.0 (https://github.com/llvm/llvm-project.git ce8795eb6c054328173876fe3fb126fd0b0b8aba))
It will start failing on the clang 10 release as soon as the branch is created.

Verified with clang master when building for znver1 (-march=znver1) which supports ADX.

This issue was originally filed against clang here: https://bugs.llvm.org/show_bug.cgi?id=44409
as a consequence of a 24% perf regression in OpenSSL reported by Phoronix here: https://www.phoronix.com/scan.php?page=article&item=gcc-clang-3960x&num=6

I was able to fix the issue and fully recover the lost performance by replacing the problematic regexpr with this one:

    /((?:^clang|LLVM) version|.*based on LLVM) ([0-9]+)\.([0-9]+)/)

Thanks
-Andrea

levitte:
Fixes #10853

### DIFF
--- a/crypto/aes/asm/aesni-mb-x86_64.pl
+++ b/crypto/aes/asm/aesni-mb-x86_64.pl
@@ -71,7 +71,7 @@ if (!$avx && $win64 && ($flavour =~ /masm/ || $ENV{ASM} =~ /ml64/) &&
 	$avx = ($1>=10) + ($1>=11);
 }
 
-if (!$avx && `$ENV{CC} -v 2>&1` =~ /((?:^clang|LLVM) version|.*based on LLVM) ([3-9]\.[0-9]+)/) {
+if (!$avx && `$ENV{CC} -v 2>&1` =~ /((?:^clang|LLVM) version|.*based on LLVM) ([0-9]+\.[0-9]+)/) {
 	$avx = ($2>=3.0) + ($2>3.0);
 }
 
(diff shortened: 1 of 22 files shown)


levitte:
Because the reporter was using 1.1.1, I'll make a separate PR for that branch, to make testing easier

paulidale (approved):

levitte:
Merged.

9bb3e5fd87905e3e9f5f7edcc2e22d98360510ab For all assembler scripts where it matters, recognise clang > 9.x

PR #10856, opened 2020-01-15
Use a wrapper for pod2html

richsalz:
A replacement for #10849 that merges the "sed" call into a single wrapper script that invokes pod2html and edits the output appropriately.


Configurations/descrip.mms.tmpl
```diff
@@ -781,13 +781,10 @@ reconfigure reconf :
...
+	  my $pod = $args{generator}->[0];
 	  return <<"EOF";
-$args{src}: $args{generator}->[0]
-	pipe pod2html "--podroot=\$(SRCDIR)/doc" --htmldir=.. -
-		       --podpath=man1:man3:man5:man7 "--infile=\$<" -
-		      "--title=$title" -
-	| \$(PERL) -pe "s|href=""http://man\\.he\\.net/(man\d/[^""]+)(?:\\.html)?""|href=""../\$1.html|g;" -
-	> \$\@
+$args{src}: $pod
+	\$(PERL) \$(SRCDIR)/util/mkpod2html.pl -i $pod -o \$\@ -t "$title" -r "\$(SRCDIR)/doc"
```
levitte: That doesn't work on VMS because of command line syntax differences.  It's fixed by insert this just below the definition of `$pod`:

``` perl
	my $mkpod2html = sourcefile('util', 'mkpod2html.pl');
```

... and then modifying this command like this:

``` make
	\$(PERL) $mkpod2html -i $pod -o \$\@ -t "$title" -r "\$(SRCDIR)/doc"
```

richsalz: thanks, fixup commit pushed.

levitte: Any second now? :wink:

levitte:
From Travis:

```
make: *** No rule to make target '"../_srcdist/doc/man1/CA.pl.pod"', needed by 'doc/html/man1/CA.pl.html'.  Stop.
```

Configurations/unix-Makefile.tmpl
```diff
@@ -1184,12 +1184,10 @@ reconfigure reconf:
...
 	  my $title = basename($args{src}, ".html");
+          my $pod = $args{generator}->[0];
 	  return <<"EOF";
-$args{src}: $args{generator}->[0]
-	pod2html "--podroot=\$(SRCDIR)/doc" --htmldir=.. \\
-		 --podpath=man1:man3:man5:man7 "--infile=\$<" "--title=$title" \\
-	| \$(PERL) -pe 's|href="http://man\\.he\\.net/(man\\d/[^"]+)(?:\\.html)?"|href="../\$1.html|g;' \\
-	> \$\@
+$args{src}: "$pod"
+	\$(PERL) \$(SRCDIR)/util/mkpod2html.pl -i "$pod" -o \$\@ -t "$title" -r "\$(SRCDIR)/doc"
```
levitte: So, er, better not to have quotes around `$pod` here in this .tmpl, as `make` uses them literally, i.e. the Travis error I showed indicates that it's looking for `"../_srcdist/doc/man1/CA.pl.pod"` literally, quotes included as part of the file name.

Configurations/windows-makefile.tmpl
```diff
@@ -586,12 +586,10 @@ reconfigure reconf:
...
 	  my $title = basename($args{src}, ".html");
+	  my $pod = $args{generator}->[0]
 	  return <<"EOF";
-$args{src}: $args{generator}->[0]
-	pod2html "--podroot=\$(SRCDIR)/doc" --htmldir=.. \\
-		 --podpath=man1:man3:man5:man7 "--infile=\$?" "--title=$title" \\
-	| \$(PERL) -pe ^"s^|href=\\^"http://man\\.he\\.net/^(man\\d/[^^\\^"]+^)^(?:\.html^)?^"^|href=\\^"../\$\$1.html^|g;^" \\
-	> \$\@
+$args{src}: "$pod"
+	\$(PERL) \$(SRCDIR)/util/mkpod2html.pl -i "$pod" -o \$\@ -t "$title" -r "\$(SRCDIR)/doc"
```
levitte: Note that it's different on Windows.  With `nmake`, quotes *do* have a meaning and are not taken literally in the same as on Unix.

Configurations/unix-Makefile.tmpl
richsalz: Okay, fixup commit removed them.  I think it's a likely to be a problem for us when someone uses a path with spaces in it.  I would rather we just say that's not supported on Unix but oh well.

Configurations/windows-makefile.tmpl
richsalz: rodger.

richsalz:
> From Travis:
> 
> ```
> make: *** No rule to make target '"../_srcdist/doc/man1/CA.pl.pod"', needed by 'doc/html/man1/CA.pl.html'.  Stop.
> ```

I can't think why that's happening, but it's late here.  I didn't change anything but the rules to build the files.  Thoughts, @levitte?

levitte:
> I can't think why that's happening, but it's late here. I didn't change anything but the rules to build the files. Thoughts, @levitte?

It was the quotes thing, which you have resolved

levitte:
You need to do a bit of rebasing...

richsalz:
Rebased and pushed.

util/mkpod2html.pl
```diff
@@ -0,0 +1,52 @@
...
+my $contents = '';
+{
+    local $/ = undef;
+    $contents = <F>;
+}
+close F;
+unlink $opt_o;
+
+$contents =~
+    s|href="http://man\.he\.net/(man\d/[^"]+)(?:\.html)?"|href="../$1.html|g;
```
mspncp: The substitution regex looks a lot better now after moving it into a perl file. But while reviewing the pr I noticed that this improvement actually just reverts 'complications' introduced by commit b0940b33a6:

The regex _used to be_ in a perl file,  `process_docs.pl`

https://github.com/openssl/openssl/blob/22db21099dd89f4b936137410a3193aebbc189f7/util/process_docs.pl#L122

until b0940b33a6 moved the code into the makefile templates. (*Note:* `process_docs.pl` is still there, but doesn't seem to be used anymore.)

This move came at the price of tripling the regex and of heavy regex quoting, because the regex is evaluated multiple times (template -> makefile -> shell/cmd.exe -> perl, see my comment in commit message a9e4e3c39e).

https://github.com/openssl/openssl/blob/c8a557357789a8a23fe7b87ec352bd5ea2c1251b/Configurations/descrip.mms.tmpl#L790
https://github.com/openssl/openssl/blob/c8a557357789a8a23fe7b87ec352bd5ea2c1251b/Configurations/unix-Makefile.tmpl#L1192
https://github.com/openssl/openssl/blob/c8a557357789a8a23fe7b87ec352bd5ea2c1251b/Configurations/windows-makefile.tmpl#L594


Now the code goes back into a perl file, albeit a different one, `mkpod2html.pl`. 

https://github.com/openssl/openssl/blob/22db21099dd89f4b936137410a3193aebbc189f7/util/mkpod2html.pl#L48

Why was `process_docs.pl` abandoned in the first place? I don't really understand the benefit of the changes in b0940b33a6. My personal impression is that the logic got more complicated but the speed didn't improve (I might be wrong, I didn't do any measurements). Also it confused me when reviewing #10132 that some `*.pod.in` files seem to be generated at build time while others are generated at installation time.

Maybe I am just missing the grand picture @levitte?

richsalz: @mspncp: what pod.in files are processed at install time?  If so, that's a bug -- all manpages should be generated as part of the ```build_generated``` target.

levitte: `process-docs.pl` was kind of a magic wand that did a little too much, and not always well.  It seemed better to use the actual commands for each case appropriately.  `mkpod2html.pl` is more of an almost drop in replacement for the normal pod2html.  Why I didn't remove it when I stopped its usage, I dunno.

mspncp: > @mspncp: what pod.in files are processed at install time? If so, that's a bug -- all manpages should be generated as part of the `build_generated` target.


I ran this command

```bash
$ find doc \( -name  '*.pod' -o -name '*.pod.in' \) | sort
```

three times on the checkout  31dc2ac6f63423be359da6985e57e72dd4d762a1 to obtain a sorted list of the files:

- 01-before       (after make clean)
- 02-after-make
- 03-after-make-install

Here are the changes:

#### After make

```diff
--- 01-before	2020-01-18 08:45:02.189862195 +0100
+++ 02-after-make	2020-01-18 08:48:05.961821622 +0100
@@ -24,57 +24,84 @@
 doc/internal/man7/build.info.pod
 doc/man1/CA.pl.pod
 doc/man1/openssl-asn1parse.pod
+doc/man1/openssl-ca.pod
 doc/man1/openssl-ca.pod.in
 doc/man1/openssl-ciphers.pod
 doc/man1/openssl-cmds.pod
+doc/man1/openssl-cms.pod
 doc/man1/openssl-cms.pod.in
+doc/man1/openssl-crl.pod
 doc/man1/openssl-crl.pod.in
 doc/man1/openssl-crl2pkcs7.pod
+doc/man1/openssl-dgst.pod
 doc/man1/openssl-dgst.pod.in
+doc/man1/openssl-dhparam.pod
 doc/man1/openssl-dhparam.pod.in
 doc/man1/openssl-dsa.pod.in
+doc/man1/openssl-dsaparam.pod
 doc/man1/openssl-dsaparam.pod.in
 doc/man1/openssl-ec.pod.in
+doc/man1/openssl-ecparam.pod
 doc/man1/openssl-ecparam.pod.in
+doc/man1/openssl-enc.pod
 doc/man1/openssl-enc.pod.in
 doc/man1/openssl-engine.pod
 doc/man1/openssl-errstr.pod
 doc/man1/openssl-fipsinstall.pod
+doc/man1/openssl-gendsa.pod
 doc/man1/openssl-gendsa.pod.in
 doc/man1/openssl-genpkey.pod.in
+doc/man1/openssl-genrsa.pod
 doc/man1/openssl-genrsa.pod.in
 doc/man1/openssl-info.pod
 doc/man1/openssl-kdf.pod
 doc/man1/openssl-list.pod
 doc/man1/openssl-mac.pod
 doc/man1/openssl-nseq.pod
+doc/man1/openssl-ocsp.pod
 doc/man1/openssl-ocsp.pod.in
+doc/man1/openssl-passwd.pod
 doc/man1/openssl-passwd.pod.in
+doc/man1/openssl-pkcs12.pod
 doc/man1/openssl-pkcs12.pod.in
 doc/man1/openssl-pkcs7.pod.in
+doc/man1/openssl-pkcs8.pod
 doc/man1/openssl-pkcs8.pod.in
 doc/man1/openssl-pkey.pod.in
 doc/man1/openssl-pkeyparam.pod.in
+doc/man1/openssl-pkeyutl.pod
 doc/man1/openssl-pkeyutl.pod.in
 doc/man1/openssl-prime.pod
 doc/man1/openssl-provider.pod
+doc/man1/openssl-rand.pod
 doc/man1/openssl-rand.pod.in
 doc/man1/openssl-rehash.pod
+doc/man1/openssl-req.pod
 doc/man1/openssl-req.pod.in
 doc/man1/openssl-rsa.pod.in
+doc/man1/openssl-rsautl.pod
 doc/man1/openssl-rsautl.pod.in
+doc/man1/openssl-s_client.pod
 doc/man1/openssl-s_client.pod.in
+doc/man1/openssl-s_server.pod
 doc/man1/openssl-s_server.pod.in
+doc/man1/openssl-s_time.pod
 doc/man1/openssl-s_time.pod.in
 doc/man1/openssl-sess_id.pod
+doc/man1/openssl-smime.pod
 doc/man1/openssl-smime.pod.in
+doc/man1/openssl-speed.pod
 doc/man1/openssl-speed.pod.in
 doc/man1/openssl-spkac.pod.in
+doc/man1/openssl-srp.pod
 doc/man1/openssl-srp.pod.in
 doc/man1/openssl-storeutl.pod.in
+doc/man1/openssl-ts.pod
 doc/man1/openssl-ts.pod.in
+doc/man1/openssl-verify.pod
 doc/man1/openssl-verify.pod.in
 doc/man1/openssl-version.pod
+doc/man1/openssl-x509.pod
 doc/man1/openssl-x509.pod.in
 doc/man1/openssl.pod
 doc/man1/tsget.pod
@@ -621,6 +648,7 @@
 doc/man7/evp.pod
 doc/man7/openssl-core.h.pod
 doc/man7/openssl-env.pod
+doc/man7/openssl_user_macros.pod
 doc/man7/openssl_user_macros.pod.in
 doc/man7/ossl_store-file.pod
 doc/man7/ossl_store.pod
```

#### After make install

```diff
--- 02-after-make	2020-01-18 08:48:05.961821622 +0100
+++ 03-after-make-install	2020-01-18 08:49:52.286955242 +0100
@@ -37,9 +37,11 @@
 doc/man1/openssl-dgst.pod.in
 doc/man1/openssl-dhparam.pod
 doc/man1/openssl-dhparam.pod.in
+doc/man1/openssl-dsa.pod
 doc/man1/openssl-dsa.pod.in
 doc/man1/openssl-dsaparam.pod
 doc/man1/openssl-dsaparam.pod.in
+doc/man1/openssl-ec.pod
 doc/man1/openssl-ec.pod.in
 doc/man1/openssl-ecparam.pod
 doc/man1/openssl-ecparam.pod.in
@@ -50,6 +52,7 @@
 doc/man1/openssl-fipsinstall.pod
 doc/man1/openssl-gendsa.pod
 doc/man1/openssl-gendsa.pod.in
+doc/man1/openssl-genpkey.pod
 doc/man1/openssl-genpkey.pod.in
 doc/man1/openssl-genrsa.pod
 doc/man1/openssl-genrsa.pod.in
@@ -64,10 +67,13 @@
 doc/man1/openssl-passwd.pod.in
 doc/man1/openssl-pkcs12.pod
 doc/man1/openssl-pkcs12.pod.in
+doc/man1/openssl-pkcs7.pod
 doc/man1/openssl-pkcs7.pod.in
 doc/man1/openssl-pkcs8.pod
 doc/man1/openssl-pkcs8.pod.in
+doc/man1/openssl-pkey.pod
 doc/man1/openssl-pkey.pod.in
+doc/man1/openssl-pkeyparam.pod
 doc/man1/openssl-pkeyparam.pod.in
 doc/man1/openssl-pkeyutl.pod
 doc/man1/openssl-pkeyutl.pod.in
@@ -78,6 +84,7 @@
 doc/man1/openssl-rehash.pod
 doc/man1/openssl-req.pod
 doc/man1/openssl-req.pod.in
+doc/man1/openssl-rsa.pod
 doc/man1/openssl-rsa.pod.in
 doc/man1/openssl-rsautl.pod
 doc/man1/openssl-rsautl.pod.in
@@ -92,9 +99,11 @@
 doc/man1/openssl-smime.pod.in
 doc/man1/openssl-speed.pod
 doc/man1/openssl-speed.pod.in
+doc/man1/openssl-spkac.pod
 doc/man1/openssl-spkac.pod.in
 doc/man1/openssl-srp.pod
 doc/man1/openssl-srp.pod.in
+doc/man1/openssl-storeutl.pod
 doc/man1/openssl-storeutl.pod.in
 doc/man1/openssl-ts.pod
 doc/man1/openssl-ts.pod.in
```

richsalz:
Thanks for all that detail, @mspncp.  I guess.  All podpages should be built as part of the "make_generated" recipe.  Got a patch to fix that?

richsalz:
Oh, and I believe this is fixed by #10873, that filelist looks familiar :)

mspncp:
> Oh, and I believe this is fixed by #10873, that filelist looks familiar :)

Yes, it seems like you fixed it. I reran the check from https://github.com/openssl/openssl/pull/10856#discussion_r368213802 on #10873 and there all pod files are generated at build time.

bernd-edlinger:
Do you have any insight why pod2html is polluting the root directory with
pod2htmd.tmp and pod2htmi.tmp (on windows) (see #10895)

richsalz:
see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=616582

richsalz:
works now on windows.  ready for review.

richsalz:
ping @levitte, @mspncp

richsalz:
Rebased (why not).  Ping for review, been a month.

mspncp:
This improvement certainly deserves to go in on the fast line. I'll give it a piggy-back approval as soon as @levitte has taken a deeper look.

t8m:
Merged to master as 1ec45faba5.

PR #10858, opened 2020-01-16
Deprecate the low level DES functions.

paulidale:
Use of the low level DES functions has been informally discouraged for a
long time. We now formally deprecate them.  Applications should instead use the EVP APIs, e.g. `EVP_EncryptInit_ex`, `EVP_EncryptUpdate`, `EVP_EncryptFinal_ex`, and the equivalently named decrypt functions.

### DIFF
--- a/apps/passwd.c
+++ b/apps/passwd.c
@@ -7,6 +7,9 @@
  * https://www.openssl.org/source/license.html
  */
 
+/* We need to use some deprecated APIs */
+#define OPENSSL_SUPPRESS_DEPRECATED
+
 #include <string.h>
 
 #include "apps.h"
@@ -16,7 +19,7 @@
 #include <openssl/err.h>
 #include <openssl/evp.h>
 #include <openssl/rand.h>
-#ifndef OPENSSL_NO_DES
+#if !defined(OPENSSL_NO_DES) && !defined(OPENSSL_NO_DEPRECATED_3_0)
 # include <openssl/des.h>
 #endif
 #include <openssl/md5.h>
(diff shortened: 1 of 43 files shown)


paulidale:
The outstanding major item is how to implement the `DES_crypt` and `DES_fcrypt` calls.

paulidale:
The `DES_crypt` and `DES_fcrypt` functions are deprecated.  This in turn means that the `-crypt` option to the `passwd` application needs to be deprecated.

Since crypt was the default algorithm, this has been changed to the BSD MD5 algorithm.

richsalz:
I think silently changing the default algorithm, which will depend on no-depracated configiuration, is a bad idea.  If not supporting deprecated, then make the default be an error.

apps/passwd.c
```diff
@@ -841,9 +845,10 @@ static int do_passwd(int passed_salt, char **salt_p, char **salt_malloc_p,
     assert(strlen(passwd) <= pw_maxlen);
 
     /* now compute password hash */
-#ifndef OPENSSL_NO_DES
+#if !defined(OPENSSL_NO_DES) && !defined(OPENSSL_NO_DEPRECATED_3_0)
     if (mode == passwd_crypt)
         hash = DES_crypt(passwd, *salt_p);
+    */
```
levitte: Spurious end if comment detected

apps/version.c
```diff
@@ -126,9 +123,6 @@ int version_main(int argc, char **argv)
     if (options) {
         printf("options: ");
         printf(" %s", BN_options());
-#ifndef OPENSSL_NO_DES
-        printf(" %s", DES_options());
-#endif
```
levitte: It's only deprecated, not removed

paulidale:
CI failures are relevant.  A function can't be deprecated but used.

apps/version.c
paulidale: All the other deprecations removed the option printing....

paulidale:
I've modified the crypt functions so they are callable as KDFs and not buried inside the passwd app.  The SHA based ones aren't working, they have some uglies to do with the salt and number of rounds which might be better refactored.

The alternative approach would be to bring the DES crypt code into the passwd app and leave the rest alone.

paulidale:
A third option would be to take the character set conversions and unpacking out of the KDFs and put them into the passwd app.  I'm not sure why each algorithm did this itself...

paulidale:
Moving this out of WIP.

paulidale:
Note: the default alg for passwd has not been changed.

paulidale:
Review?

paulidale:
Merged to master, thanks for the feedback.

PR #10862, opened 2020-01-16
Don't register drbg_delete_thread_state twice

mattcaswell:
drbg_delete_thread_state cleans up after both the public and the private
DRBG. It can be registered automtically by getting either of those DRBGs,
but it should not be registered twice.


mattcaswell:
Pushed. Thanks.

PR #10863, opened 2020-01-16
Fix init_thread_stop

mattcaswell:
init_thread_stop maintains a linked lists of handlers that it should
call when a thread finishes. The linked list handling wasn't quite right
resulting in corrupted data.

### DIFF
--- a/crypto/initthread.c
+++ b/crypto/initthread.c
@@ -297,7 +297,7 @@ void ossl_ctx_thread_stop(void *arg)
 
 static void init_thread_stop(void *arg, THREAD_EVENT_HANDLER **hands)
 {
-    THREAD_EVENT_HANDLER *curr, *prev = NULL;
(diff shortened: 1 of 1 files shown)


mattcaswell:

crypto/initthread.c
```diff
@@ -306,15 +306,20 @@ static void init_thread_stop(void *arg, THREAD_EVENT_HANDLER **hands)
     curr = *hands;
     while (curr != NULL) {
         if (arg != NULL && curr->arg != arg) {
+            prev = curr;
             curr = curr->next;
             continue;
         }
         curr->handfn(curr->arg);
-        prev = curr;
+        if (prev == NULL)
```
slontis: ?? With the way this loop is written I dont see how prev can be NULL.
prev is only ever set to curr and curr is always checked for NULL.

mattcaswell:
Pushed. Thanks.

PR #10864, opened 2020-01-16
Convert rand_bytes_ex and rand_priv_bytes_ex to public functions

mattcaswell:
These were initially added as internal functions only. However they will
also need to be used by libssl as well. Therefore it make sense to move
them into the public API.

##### Checklist


levitte:
Actually, you do have the choice to keep it internal and still export the symbols.  All you have to do it is to stick it in an internal header (preferably in `include/internal/` and if necessary, modify this perl nugget:

https://github.com/openssl/openssl/blob/437ad983c6ca254cd1d4367fb1d68338c82e8ee3/Configurations/unix-Makefile.tmpl#L1022-L1048

mattcaswell:
Yes, I did consider that option. But ultimately I think this is useful so should be done anyway.

levitte:
> Yes, I did consider that option. But ultimately I think this is useful so should be done anyway.

Sure, I'm fine with that, just wanted to make sure you knew you have the choice

doc/man3/RAND_bytes.pod
```diff
@@ -29,6 +33,15 @@ instance so that a compromise of the "public" PRNG instance will not
 affect the secrecy of these private values, as described in L<RAND(7)>
 and L<RAND_DRBG(7)>.
 
+RAND_bytes_ex() and RAND_priv_bytes_ex() are the same as RAND_bytes() and
+RAND_priv_bytes() except that they both take an additional I<ctx> parameter.
+The DRBG used for the operation is the public or private DRBG associated with
+the specified I<ctx>. The parameter can be NULL, in which case
```
slontis: Does fips need to be mentioned here -  as a NULL context isnt so great for that case :)

mattcaswell: As far as end user applications are concerned it is perfectly ok to use a NULL context with the FIPS module. A NULL context gets translated into a concrete instance by libcrypto - so by the time it gets to the FIPS module one is always present. Our own internal code will have to be careful to always supply one if we're making these calls from inside the FIPS module. But these man pages are definitely targeting end users - not us.

slontis: I assume you dont mean via openssl_ctx_get_concrete(NULL)

mattcaswell: Yes - I do (don't forget this happens in libcrypto not the FIPS module - so that's perfectly ok).

kaduk:
Could the commit message summary be updated to use the proper names of the functions, replete with majuscule letters?

PR #10865, opened 2020-01-16
Implement the NULL cipher in the default provider

mattcaswell:
Libssl uses the null cipher in certain situations. It should be
converted to a provided cipher.

### DIFF
--- a/crypto/evp/evp_enc.c
+++ b/crypto/evp/evp_enc.c
@@ -142,6 +142,7 @@ int EVP_CipherInit_ex(EVP_CIPHER_CTX *ctx, const EVP_CIPHER *cipher,
 
     if (tmpcipher->prov == NULL) {
         switch(tmpcipher->nid) {
+        case NID_undef:
         case NID_aes_256_ecb:
         case NID_aes_192_ecb:
         case NID_aes_128_ecb:
@@ -326,7 +327,10 @@ int EVP_CipherInit_ex(EVP_CIPHER_CTX *ctx, const EVP_CIPHER *cipher,
         return 0;
 #else
         EVP_CIPHER *provciph =
-            EVP_CIPHER_fetch(NULL, OBJ_nid2sn(cipher->nid), "");
+            EVP_CIPHER_fetch(NULL,
+                             cipher->nid == NID_undef ? "NULL"
+                                                      : OBJ_nid2sn(cipher->nid),
+                             "");
 
         if (provciph == NULL) {
             EVPerr(EVP_F_EVP_CIPHERINIT_EX, EVP_R_INITIALIZATION_ERROR);
--- a/providers/implementations/ciphers/cipher_null.c
+++ b/providers/implementations/ciphers/cipher_null.c
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2020 The OpenSSL Project Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License 2.0 (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
+ */
+
+#include <string.h>
+#include <openssl/crypto.h>
+#include <openssl/core_numbers.h>
+#include "prov/implementations.h"
+#include "prov/ciphercommon.h"
+#include "prov/providercommonerr.h"
+
+static OSSL_OP_cipher_newctx_fn null_newctx;
+static void *null_newctx(void *provctx)
+{
+    return OPENSSL_zalloc(1);
+}
+
+static OSSL_OP_cipher_freectx_fn null_freectx;
+static void null_freectx(void *vctx)
+{
+    OPENSSL_free(vctx);
+}
+
+static OSSL_OP_cipher_encrypt_init_fn null_init;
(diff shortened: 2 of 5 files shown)


providers/implementations/ciphers/cipher_null.c
```diff
@@ -0,0 +1,108 @@
...
+static void *null_newctx(void *provctx)
+{
+    return OPENSSL_zalloc(1);
+}
+
+static OSSL_OP_cipher_freectx_fn null_freectx;
+static void null_freectx(void *vctx)
+{
+    OPENSSL_free(vctx);
+}
```
mspncp: Since the`null_newctx()` and `null_freectx()` calls are paired, and since the data pointed to by`vctx` is not used by the NULL cipher anyway, I don't see any need to allocate the data from the heap. All that is required is to return a valid pointer:


```C
static OSSL_OP_cipher_newctx_fn null_newctx;
static void *null_newctx(void *provctx)
{
    static int dummy = 0;
    return &dummy;
}

static OSSL_OP_cipher_freectx_fn null_freectx;
static void null_freectx(void *vctx)
{
    (void)vctx; /* unused */
}
```

mattcaswell: Good idea. I've done that.

mattcaswell:
Fixup for the above feedback item pushed. That fixup also includes a fix for the travis failure that was occurring.

providers/implementations/ciphers/cipher_null.c
mspncp: Fixup LGTM.

levitte:
Darn...  I was hoping not having to see this.  But if we use it internally (er...  why?), then so be it.

crypto/evp/evp_enc.c
```diff
@@ -142,6 +142,7 @@ int EVP_CipherInit_ex(EVP_CIPHER_CTX *ctx, const EVP_CIPHER *cipher,
 
     if (tmpcipher->prov == NULL) {
         switch(tmpcipher->nid) {
+        case NID_undef:
```
levitte: Why do we still have this switch?  Wouldn't it be more sensible (and less error prone, because *you* might remember to fill this in, but others might not) to try to fetch, and only fall back to legacy if that fails?

might be subject for another PR...

mattcaswell: Another PR I'd say

mattcaswell:
> Darn... I was hoping not having to see this. But if we use it internally (er... why?), then so be it.

Code is here:

https://github.com/openssl/openssl/blob/9372ddf1a294d61dcbf507680e4e3d5b094ef71d/ssl/ssl_ciph.c#L519-L528

Basically if we negotiate a ciphersuite with "NULL" encryption (of which there are a number of ciphersuites that provide that), then we just use the NULL EVP cipher. That way libssl doesn't have to care - it just treats the "encryption" of those ciphersuites in the same way as any other ciphersuite. It would be a big pain to refactor things to remove that.

richsalz:
Seems wrong to provide the NULL cipher, but not other legacy ciphers by default.

crypto/evp/evp_enc.c
levitte: Okie

levitte:
I think that a point here is that we've kept all the ciphers we do use in libssl in the default provider.  In that sense, it makes sense to have the NULL cipher (which I would label "bastard" rather than "legacy", btw) in the default provider.

However, this makes me wish the SSL machinery had more of a pipe like implementation, so that if the "NULL" encryption is asked for, we'd simply not have the encryption bit in the pipe, rather than adding a NULL bit...

mattcaswell:
> However, this makes me wish the SSL machinery had more of a pipe like implementation, so that if the "NULL" encryption is asked for, we'd simply not have the encryption bit in the pipe, rather than adding a NULL bit...

I'm eagerly awaiting the PR :-)

levitte:
> I'm eagerly awaiting the PR :-)

It's already submitted into the "ideas" repository...  now, I only have to wait for it to take root in your brain :wink:

richsalz:
There is something to be said for the current structure which means piecewise verification of the code is easier.  It'd be a shame to lose that.

levitte:
> There is something to be said for the current structure which means piecewise verification of the code is easier. It'd be a shame to lose that.

I didn't suggest to lose that.  Units should, of course, be tested individually.  But there's something to be said about integration tests as well, and how the different parts of a pipe work together would probably fall into that category.
That being said, assuming that's what you're talking about...  otherwise, feel free to ignore me

richsalz:
The current flow is "lookup EVP, stick it into the processing."  I think the "skip EVP" would require more code complexity, several if-checks, etc.  Harder to verify.  Maybe not.  Hard to guess about code that's not written.

levitte:
Okie. That's a side discussion anyway...

mattcaswell:
> Do you need the NULL digest also (Let me know if you want it added)?

It's currently used by the ca app when EdDSA is in use (which does not have a separate digest). The EdDSA digest-sign implementations support the `EVP_PKEY_CTRL_MD` ctrl and will accept either NULL or the null-digest as a parameter. We need to think about how that might translate to providers. I suggest deferring that until EdDSA is made provider aware.

mattcaswell:
Pushed. Thanks.

PR #10866, opened 2020-01-16
Introduce SSL_CTX_new_ex()

mattcaswell:
Add the ability to specify an OPENSSL_CTX (which may be NULL for the
default context) and a property query string for use during algorithm
fetch operations.

For example, in this way one SSL_CTX could be used the default provider,
and another one could be used with the FIPS provider.

At this stage we don't use these values. That will come later. For the same reason there is no test (it doesn't do anything yet, so nothing to test).

See #10854 for the big picture of where this is heading.

### DIFF
--- a/doc/man3/SSL_CTX_new.pod
+++ b/doc/man3/SSL_CTX_new.pod
@@ -3,7 +3,7 @@
 =head1 NAME
 
 TLSv1_2_method, TLSv1_2_server_method, TLSv1_2_client_method,
-SSL_CTX_new, SSL_CTX_new_from_ctx, SSL_CTX_up_ref, SSLv3_method,
+SSL_CTX_new, SSL_CTX_new_with_libctx, SSL_CTX_up_ref, SSLv3_method,
 SSLv3_server_method, SSLv3_client_method, TLSv1_method, TLSv1_server_method,
 TLSv1_client_method, TLSv1_1_method, TLSv1_1_server_method,
 TLSv1_1_client_method, TLS_method, TLS_server_method, TLS_client_method,
@@ -18,8 +18,8 @@ functions
 
  #include <openssl/ssl.h>
 
- SSL_CTX *SSL_CTX_new_from_ctx(OPENSSL_CTX *libctx, const char *propq,
-                               const SSL_METHOD *method);
+ SSL_CTX *SSL_CTX_new_with_libctx(OPENSSL_CTX *libctx, const char *propq,
+                                  const SSL_METHOD *method);
  SSL_CTX *SSL_CTX_new(const SSL_METHOD *method);
  int SSL_CTX_up_ref(SSL_CTX *ctx);
 
@@ -73,7 +73,7 @@ functions
 
 =head1 DESCRIPTION
 
-SSL_CTX_new_from_ctx() creates a new B<SSL_CTX> object as a framework to
+SSL_CTX_new_with_libctx() creates a new B<SSL_CTX> object as a framework to
 establish TLS/SSL or DTLS enabled connections using the library context
 I<libctx> (see L<OPENSSL_CTX(3)>). Any cryptographic algorithms that are used
 by any B<SSL> objects created from this B<SSL_CTX> will be fetched from the
@@ -81,7 +81,7 @@ I<libctx> using the property query string I<propq> (see
 L<provider(7)/Fetching algorithms>. Either or both the I<libctx> or I<propq>
 parameters may be NULL.
 
-SSL_CTX_new() does the same as SSL_CTX_new_from_ctx() except that the default
+SSL_CTX_new() does the same as SSL_CTX_new_with_libctx() except that the default
 library context is used and no property query string is specified.
 
 An B<SSL_CTX> object is reference counted. Creating an B<SSL_CTX> object for the
@@ -224,7 +224,7 @@ and TLS_client_method() functions were added in OpenSSL 1.1.0.
 
 All version-specific methods were deprecated in OpenSSL 1.1.0.
 
-SSL_CTX_new_from_ctx() was added in OpenSSL 3.0.
+SSL_CTX_new_with_libctx() was added in OpenSSL 3.0.
 
 =head1 COPYRIGHT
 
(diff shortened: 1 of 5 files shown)


levitte:
Side note: Can't we try and get away from the `_ex` ending?  What happens the next time we need a new variant, are we going to call it `SSL_CTX_new_ex_ex`?  We might want to spend some time at some point reviewing new names...

mattcaswell:
> Other than being annoyed with the _ex ending, this looks good.

I had a total imagination failure when trying to come up with a name :-(

kaduk:
I feel like we've had some previous discussions that involved possibly wanting a ```SSL_CTX_new_ex()``` API; has anyone tried to search for those discussions so that we can get all the goodies into ```_ex``` and postpone the need for ```_ex_ex`?

levitte:
> I had a total imagination failure when trying to come up with a name :-(

`SSL_CTS_new_from_ctx`
`SSL_CTX_new_provided`

Spur of the moment names

doc/man3/SSL_CTX_new.pod
```diff
@@ -70,20 +72,29 @@ functions
...
-establish TLS/SSL or DTLS enabled connections. An B<SSL_CTX> object is
-reference counted. Creating an B<SSL_CTX> object for the first time increments
-the reference count. Freeing it (using SSL_CTX_free) decrements it. When the
-reference count drops to zero, any memory or resources allocated to the
-B<SSL_CTX> object are freed. SSL_CTX_up_ref() increments the reference count for
-an existing B<SSL_CTX> structure.
+SSL_CTX_new_ex() creates a new B<SSL_CTX> object as a framework to
+establish TLS/SSL or DTLS enabled connections using the library context
+I<libctx> (see L<OPENSSL_CTX(3)>). Any cryptographic algorithms that are used
+by any SSL objects will be fetched from the I<libctx> using the property query
```
kaduk: "any SSL objects created from this SSL_CTX", surely.

doc/man3/SSL_CTX_new.pod
```diff
@@ -70,20 +72,29 @@ functions
...
+string I<propq> (see L<provider(7)/Fetching algorithms>. Either the I<libctx>
+or I<propq> parameters may be NULL. Note that the I<propq> value is not
```
kaduk: This is a phrasing that some readers might expect to imply an exclusive or.  Maybe something like "NULL values are permitted for I<libctx> and I<propq> when the corresponding default behavior is acceptable" would be better, though there aren't exactly any great solutions in this space.

doc/man3/SSL_CTX_new.pod
```diff
@@ -70,20 +72,29 @@ functions
...
+duplicated and is assumed to exist for the lifetime of the SSL_CTX object.
```
kaduk: I don't particularly remember us using language like this for memory-lifetime issues, but I'm also not confident that I would remember if we already had something like it.

doc/man3/SSL_CTX_new.pod
```diff
@@ -70,20 +72,29 @@ functions
...
+
+SSL_CTX_new() does the same as SSL_CTX_new_ex() except that the default library
+context is used and no property query string is specified.
+
+An B<SSL_CTX> object is reference counted. Creating an B<SSL_CTX> object for the
+first time increments the reference count. Freeing it (using SSL_CTX_free)
+decrements it. When the reference count drops to zero, any memory or resources
```
kaduk: nit: using "it" twice in the same sentence for different antecedents is a bit unfortunate

t8m:
One of the SSL_CTX_new_ex ideas was about enforcing the server certificate hostname check on client side with the new call.

mattcaswell:
> One of the SSL_CTX_new_ex ideas was about enforcing the server certificate hostname check on client side with the new call.

So....if you call SSL_CTX_new_ex then any connection would fail if you haven't set a hostname? That could be done.

doc/man3/SSL_CTX_new.pod
mattcaswell: Fixed

doc/man3/SSL_CTX_new.pod
mattcaswell: I went with "Either or both"

doc/man3/SSL_CTX_new.pod
mattcaswell: I actually reconsidered this behaviour and removed the sentence entirely. I now duplicate the provided string. Normally I expect these strings will just be literals, but I can't exclude the possibility that in some circumstances they may be dynamically constructed - so probably "duping" the string is a better idea.

doc/man3/SSL_CTX_new.pod
mattcaswell: Reworded

mattcaswell:
> I had a total imagination failure when trying to come up with a name :-(

I've changed the name to `SSL_CTX_new_from_ctx`

> One of the SSL_CTX_new_ex ideas was about enforcing the server certificate hostname check on client side with the new call.

I had a think about this and I do believe its a good idea. We could check both the server certificate hostname and SNI hostname potentially. However I also think its probably best dealt with in a different follow-on PR. This PR has the objective of adding the libctx/propq values to SSL_CTX so as to enable #10854 to progress. I think the hostname checks thing has the potential to be a bit tricky to get right and I don't want to hold #10854 up while we do that.


@levitte - please can your reconfirm your approval?

doc/man3/SSL_CTX_new.pod
```diff
@@ -3,20 +3,23 @@
...
+DTLSv1_client_method, DTLSv1_2_method, DTLSv1_2_server_method,
+DTLSv1_2_client_method
 - create a new SSL_CTX object as framework for TLS/SSL or DTLS enabled
 functions
 
 =head1 SYNOPSIS
 
  #include <openssl/ssl.h>
 
+ SSL_CTX *SSL_CTX_new_from_ctx(OPENSSL_CTX *libctx, const char *propq,
```
slontis: SSL_CTX_new_from_libctx() ?

from_ctx is not great when there are many CTX related names.

t8m: Or SSL_CTX_new_with_libctx()? Using with instead of from because otherwise it would appear like the OPENSSL_CTX is a factory for SSL_CTX.

mattcaswell:
Ok - name changed again as requested. This time to `SSL_CTX_new_with_libctx()`

mattcaswell:
@levitte please reconfirm (again!)

mattcaswell:
Pushed. Thanks.

PR #10875, opened 2020-01-17
Add GNU properties note for Intel CET in x86_64-xlate.pl

levitte:
This appears to be emitted with gcc and clang with -fcf-protection
selected, so we should do the same.

We're trying to be smart, and only emit this when the 'endbranch'
pseudo-mnemonic has been used at least once.

This is inspired by and owes to work done by @hjl-tools (github)

### DIFF
--- a/crypto/perlasm/x86_64-xlate.pl
+++ b/crypto/perlasm/x86_64-xlate.pl
@@ -101,6 +101,24 @@ elsif (!$gas)
     $decor="\$L\$";
 }
 
+my $cet_property = <<'_____';
+	.section ".note.gnu.property", "a"
+	.align 8
+	.long 1f - 0f
+	.long 4f - 1f
+	.long 5
+0:
+	.asciz "GNU"
+1:
+	.align 8
+	.long 0xc0000002
+	.long 3f - 2f
+2:
+	.long 3
+3:
+	.p2align 3
+4:
+_____
 my $current_segment;
 my $current_function;
 my %globals;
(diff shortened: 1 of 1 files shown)


levitte:
@hjl-tools' work that I refer to is in #10829

crypto/perlasm/x86_64-xlate.pl
```diff
@@ -101,6 +101,24 @@
     $decor="\$L\$";
 }
 
+my $cet_property = <<'_____';
+	.section ".note.gnu.property", "a"
```
bernd-edlinger: does this also work for windows and non-elf targets?

bernd-edlinger: Ah, ok, I didn't see the "if ($gas " below

bernd-edlinger (approved):

crypto/perlasm/x86_64-xlate.pl
levitte: If there shows up something similar for non-GNU platforms, we'll have to deal with it then.  I tried to find out if there is something similar for other platforms, but either there isn't (yet), or my google-fu is too weak.

levitte:
Merged.

b2b43d1b698235ee674e588c224db38d9b78d1ea Add GNU properties note for Intel CET in x86_64-xlate.pl

hjl-tools:
This commit is incomplete:

https://github.com/openssl/openssl/issues/10896

crypto/perlasm/x86_64-xlate.pl
kaduk: Does the LLVM integrated assembler count as a "GNU platform" for this purpose?

hjl-tools: Yes, it is on Linux.

PR #10876, opened 2020-01-17
BIO: Make BIO_f_buffer's reader less desperate

### LINKED ISSUE #10859: BIO_f_buffer() wants to read more from a blocking socket than is available
```
cat > trivial-server.c <<EOF
#include <assert.h>
#include <openssl/bio.h>
#include <stdio.h>

int main()
{
    BIO *accept_bio = BIO_new_accept("8080");
    assert(accept_bio != NULL);
    int err = BIO_do_accept(accept_bio);
    assert(err == 1);

    err = BIO_do_accept(accept_bio);
    assert(err == 1);
    BIO *bio = BIO_pop(accept_bio);
    assert(bio != NULL);

#if DEMONSTRATE_FAILURE
    BIO *bbio = BIO_new(BIO_f_buffer());
    BIO_push(bbio, bio);
    bio = bbio;
    assert(bio != NULL);
#endif
    char buffer[1024];
    int nbytes = BIO_read(bio, buffer, 1024);
    printf("%.*s\n", nbytes, buffer);
    BIO_write(bio, "world", 5);
    BIO_flush(bio);
}
EOF
cat >trivial-client.c <<EOF
#include <assert.h>
#include <openssl/bio.h>
#include <stdio.h>

int main()
{
    BIO *bio = BIO_new_connect("localhost:8080");
    assert(bio != NULL);
    int err = BIO_do_connect(bio);
    assert(err == 1);
    BIO_write(bio, "hello", 5);
    BIO_flush(bio);
    char buffer[1024];
    int nbytes = BIO_read(bio, buffer, 1024);
    printf("%.*s\n", nbytes, buffer);
}
EOF
gcc -lcrypto trivial-server.c -o trivial-server -DDEMONSTRATE_FAILURE
gcc -lcrypto trivial-client.c -o trivial-client
./trivial-server &
./trivial-client
```

When compiled with `-DDEMONSTRATE_FAILURE`, the server chains a `BIO_f_buffer` in front of the BIO that reads from the socket. This means that the server's `BIO_read(bio, buffer, 1024)` attempts to read 1024 bytes — succeeds at reading 5 bytes — _attempts to read 1024 more bytes_ — and hangs forever, because the client is done sending data and is waiting for the server to process the data it's already got.

This means that it is _still_ unsafe to use a `BIO_f_buffer` downstream of a `BIO_new_ssl`, even after the bugfix in #322. Bug #322 was about how the SSL BIO broke the contract of the buffer BIO while writing; this new bug is about how the buffer BIO breaks the contract of the SSL BIO while reading.

I think a buffer BIO should never attempt to read any additional data into its buffer unless either (A) its buffer was actually empty on entry into `buffer_read`, or (B) it has gotten a positive value from `BIO_pending(BIO_next(self))`.

levitte:
BIO_f_buffer's BIO_read implementation desperately tries to fill the
caller's buffer, which may cause some surprise, as this effectively
blocks communication if less has been received from the next BIO.
This change makes the reader less desperate.  It will only attempt to
read from the next BIO at most once per call, and take whatever it
gets, into the internal buffer if necessary.

Fixes #10859

### DIFF
--- a/crypto/bio/bf_buff.c
+++ b/crypto/bio/bf_buff.c
@@ -88,7 +88,7 @@ static int buffer_free(BIO *a)
 
 static int buffer_read(BIO *b, char *out, int outl)
 {
-    int i, num = 0;
+    int i;
     BIO_F_BUFFER_CTX *ctx;
 
     if (out == NULL)
@@ -97,7 +97,6 @@ static int buffer_read(BIO *b, char *out, int outl)
 
     if ((ctx == NULL) || (b->next_bio == NULL))
         return 0;
-    num = 0;
     BIO_clear_retry_flags(b);
 
  start:
@@ -109,11 +108,7 @@ static int buffer_read(BIO *b, char *out, int outl)
         memcpy(out, &(ctx->ibuf[ctx->ibuf_off]), i);
         ctx->ibuf_off += i;
         ctx->ibuf_len -= i;
-        num += i;
-        if (outl == i)
-            return num;
-        outl -= i;
-        out += i;
+        return i;
     }
 
     /*
@@ -123,21 +118,10 @@ static int buffer_read(BIO *b, char *out, int outl)
      * address space
      */
     if (outl > ctx->ibuf_size) {
-        for (;;) {
-            i = BIO_read(b->next_bio, out, outl);
-            if (i <= 0) {
-                BIO_copy_next_retry(b);
-                if (i < 0)
-                    return ((num > 0) ? num : i);
-                if (i == 0)
-                    return num;
-            }
-            num += i;
-            if (outl == i)
-                return num;
-            out += i;
-            outl -= i;
-        }
+        i = BIO_read(b->next_bio, out, outl);
+        if (i <= 0)
+            BIO_copy_next_retry(b);
+        return i;
     }
     /* else */
 
@@ -145,10 +129,7 @@ static int buffer_read(BIO *b, char *out, int outl)
     i = BIO_read(b->next_bio, ctx->ibuf, ctx->ibuf_size);
     if (i <= 0) {
         BIO_copy_next_retry(b);
-        if (i < 0)
-            return ((num > 0) ? num : i);
-        if (i == 0)
-            return num;
+        return i;
     }
     ctx->ibuf_off = 0;
     ctx->ibuf_len = i;


levitte:
Note that I'm not proposing to backport it to the 1.1.1 branch, because it is a behaviour change, and it is unclear what the actual expectations are, i.e. should it or should it not try to desperately fill all `outl` bytes.

paulidale (approved):
LGTM.

The semantic change could have unexpected impacts.  E.g. asking how many lines are buffered could return 0 when there is plenty of data that could be read but hasn't.  Time will tell.

t8m:
Hopefully this is acceptable change for 3.0. Another option would be to add a flag

levitte:
> Another option would be to add a flag

And the default would be? I can see that question being endlessly discussed :wink:

Quuxplusone:
LGTM! (modulo my question in #10859 about whether OpenSSL has a test suite and if so whether this behavior should be tested)

t-j-h:
Ouch. That behaviour change makes the buffer not actually buffer in effect.

Quuxplusone:
@t-j-h: I recommend reading the discussion in #10859. The point of a buffer is to sometimes turn multiple reads into a single read, and to sometimes turn multiple writes into a single write. This PR preserves that "buffering" behavior (or at least if it doesn't, that's a bug).  But the bug in the existing `buffer_read`, which this PR is intended to fix, is that the existing `buffer_read` sometimes turns a _single_ read into _multiple_ (blocking) reads, and that breaks many network protocols, including HTTP and TLS.

paulidale:
I've added an OTC hold for this while the technical discussion continues.

t-j-h:
This BIO filter has always read additional data of a 4k block from the next BIO in the chain.
If that BIO is blocking, this BIO will wait. If it is non-blocking, this BIO returns the non-blocking result status to the caller.

There has never been semantics defined that says that this BIO will not try to read-ahead or will limit itself to a single read despite what the caller specifies is a major semantic change. If the caller asks for 1024 bytes to be returned then this filter will keep reading until such time as it has 1024 bytes to return or has an error. That is one of its purposes.

With this change if (for example) a TLS server is returning things with a write that writes out small records "immediately" which is incredibly common, then you will only get the first TLS record returned. i.e. a server doing a write of a single byte or using a pile of small writes and leaving it to the lower layers to buffer which in many contexts they don't, you will generate a very different result with this change.

Nothing in the code or the documentation supports anyone assuming the semantics are meant to be only-a-single-read-call-to-the-next-BIO-is-ever-used.

@Quuxplusone I've read #10859 (and had done so before commenting on this PR) - but you don't go changing the semantics in a general purpose buffering context because you have a problem at another layer. 

Your test code is making assumptions about BIO_f_buffer that are invalid. It shouldn't work.
It is documented that it has a buffer - and that you cannot set the buffer size below the compiled in value. Originally it was a 1024 byte buffer, and in 2001 it got changed to a 4096 byte buffer.

levitte:
Actually, the semantics regarding the internal buffer hasn't changed, it remains exactly the same. What has changed is the insistence on filling out the callers buffer (i.e absolutely filling out all `outl` bytes, even if the next bio gave back less than that)

t-j-h:
And that is a *semantic* change.  That is what I was describing with the small TLS records - you will return early and that is a behaviour change. You are no longer buffering until you get what the caller requested.

levitte:
To do that, having an internal 4k buffer is totally unnecessary, all that's needed is a loop around a read of the next BIO until `outl` bytes have been read.

petrovr:
Please propose changes in manual page - describe read behaviour.

petrovr:
From proposed patch is not what is changed. Looking into final (web-view) I'm not sure that proposed modification is correct.

Old code looks good. If program try to read N bytes from buffer fill with less then N bytes method should perform reads on maximum until read method return zero or error.

At moment is not clear to me goal of BIO_copy_next_retry(b) - ( left for home work ) .

levitte:
Apparently, #10859 disagrees with you, @petrovr.

The way BIO_f_buffer() works, it has a dual purpose:

1)  insist on filling what the caller asks for (there's zero need for an internal buffer to do that). 
2)  try to read a large chunk at a time to an internal buffer and pick off from that to fill what the caller asks for.

Those two things aren't conceptually the same, even though they have been combined into one.

levitte:
As for multiple small TLS records, they will still be read in one swoop, *if they are available in the call that fills the internal buffer*, i.e. this:

https://github.com/openssl/openssl/blob/bddbfae1cd667f0c1458deff98cbc03171e90d5a/crypto/bio/bf_buff.c#L145

That call isn't gone, as you can see for yourself.

t-j-h:
And now I finally understand the disconnect in this discussion. The record boundaries are preserved in the reading and writing. If the server writes a 1 byte block then another 1 byte block and you do a read as a client for 2 bytes you will get a 1 byte block back first - you don't get back 2 bytes. This is a fundamental property of how things work - the record boundaries themselves are respected in the layered handling. If you want to coalesce things you have to do that in **application code** or by pushing in a **buffering filter**.

They will **not** be read in one swoop.

richsalz:
Thanks for the explanation, @t-j-h .  I am not sure this abstraction (still?) makes sense with TLS which is all about a bytestream.

t-j-h:
TLS is a two protocols - explicitly noted in the specification (in every version of it)  - at the lowest level a record protocol, and a handshake protocol. It has always been a record protocol. 
Note that - **records**.  TLS is not a byte stream protocol. TLS sits on top of a _reliable, in-order data stream_ but is itself a record protocol. 

Quoting the specification - 

> The record protocol divides traffic up into a series of records, each of which is independently protected using the traffic keys.

richsalz:
I don't believe that implementation detail is how application developers think of TLS, but we just disagree.

levitte:
I submitted an alternative PR that adds details to the BIO_f_buffer documentation, in #10890. Comments appreciated.
(I should add a warning about chaining more than on BIO_f_buffer, as that will force a fill of the internal buffer)

levitte:
Something that's a bit problematic with `BIO_f_buffer` is that if you chain two of them, you'll get a surprising block, the reason being that the 4K read from the "upper" one will cause the next one to block until the 4K have been successfully read (or communication shutdown has been reached).

kaduk:
> TLS is a two protocols - explicitly noted in the specification (in every version of it) - at the lowest level a record protocol, and a handshake protocol. It has always been a record protocol.
> Note that - **records**. TLS is not a byte stream protocol. TLS sits on top of a _reliable, in-order data stream_ but is itself a record protocol.
> 
> Quoting the specification -
> 
> > The record protocol divides traffic up into a series of records, each of which is independently protected using the traffic keys.

I think it's more relevant to look at:

>    Application Data messages contain data that is opaque to TLS.
   Application Data messages are always protected.  Zero-length
   fragments of Application Data MAY be sent, as they are potentially
   useful as a traffic analysis countermeasure.  Application Data
   fragments MAY be split across multiple records or coalesced into a
   single record.

The message boundaries for messages provided as input to a TLS implementation on one endpoint are not, in general, reflected in the message boundaries of the output of the TLS implementation of the peer.  As a corollary, for application data, TLS record/fragment boundaries have no semantic meaning.

t-j-h:
Much more relevant @kaduk is you cannot state that returning records that match what the sender sent is any violation of the specification and happens to be the behaviour OpenSSL has always had - whatever is written (without buffering) is what is read. There is no coalescing of data in the SSL layer itself. What one end does as its writing pattern is what is visible in the reading pattern. 

Sure you can chose to implement things a different way - but this is how OpenSSL does things. And it is why @levitte suggested change has a problem in that it alters the behaviour because it was based on the (invalid) assumption that the read would coalesce which it doesn't. 

The TLS spec says nothing about what APIs that implement it should do. They are free to expose behaviours which reflect the underlying record protocol or to not do so. You pick which approach you want to take. And then applications will end up depending on that behaviour. Changing such a behaviour is not something to do lightly.

levitte:
> And it is why @levitte suggested change has a problem in that it alters the behaviour because it was based on the (invalid) assumption that the read would coalesce which it doesn't.

Did I make that assumption?  You may have misunderstood what I wrote.

levitte:
Regardless of the discussions here, I'm closing this in favour of #10890, since that was acceptable to @Quuxplusone

kaduk:
@t-j-h while I may end up agreeing with you about what openssl the project should do right now, solely as a pragmatic matter, I continue to believe that applications depending on read behavior tied to the underlying records are fundamentally broken.  In other words, such applications are relying not on TLS as the transport, but on OpenSSL, and attempting to slot in a different TLS implementation for one endpoint would risk devastating results.  The point of having specifically a TLS implementation is that it is fully interoperable with *other* TLS implementations; the TLS specification gives a sending endpoint the flexibility to do whatever fragmentation and coalescing it wants, and so a TLS receiver MUST be prepared to accept such behavior.  While the TLS specification does not constrain the specific APIs exposed by implementation, if the implementation chooses to provide API semantics that cause applications to break if the peer changes its sending behavior *in a way explicitly allowed by the TLS standard*, those API semantics are bad.
Whether openssl, having historically offered API semantics that are or could be interpreted by means of ambiguity in the documentation to fall into this "bad" case, is stuck continuing to offer them in order to preserve ABI compatibility, is a separate question from whether they are bad API semantics.
Based on my extensive work on the TLS specifications and the conversations with other experts involved in that work, I am convinced that the only sensible API model for TLS (1-RTT) application data is a reliable, in-order byte stream with no internal boundaries -- the same requirement that TLS has for its own underlying transport.

t-j-h:
@kaduk so your position is that APIs that respect the record boundary defined are fundamentally flawed and are _bad_. If I'm following your view correctly then I'm simply going to have to disagree - as you can make a byte stream view out of the record view, but you cannot re-make the record view out of a bytestream view - so layering wise if you pick the approach you are suggesting you lose information. 

BTW you would be classifying OpenSSL, LibreSSL, BoringSSL, Java and all Win32 SSPI API usage as _bad_ under your model and view as they **all** respect the record boundary and return accordingly.
And that's a pretty substantial portion of the majority of APIs. Sure they are a lot more APIs out there but this covers the majority. 

These are choices that weren't made lightly - nor are they unreasonable choices to make for an API - to actually return based on the underlying information. 

The specification does state the message boundaries are not preserved (logically) **because** an implementation is permitted to coalesce multiple client messages of the same content type into a single record - but it also does not **require** you to do so, nor does it recommend that APIs don't do this. Nor do the major implementations actually coalesce in this manner.  Effectively the specification is stating that you are permitted to coalesce but you are not required to do so and if you elect to coalesce you will not be preserving the applications record boundary. That does not mean that an API or an implementation should elect to read across record boundaries or should enforce coalescing on the client side - no such requirement or recommendation exists within the specification.

levitte:
Side note: `BIO_f_buffer()` can be used with other BIOs than the SSL ones, so coalescing or not depends on the situation.

PR #9496, opened 2019-07-31
Replace FUNCerr with ERR_raise_data

richsalz:
Now that #9452 has been merged, let's replace all ```FUNCerr``` calls with ``ERR_raise_data```

No compatibility is needed since ```ERRfunc()``` was only in master.

PR #10881, opened 2020-01-17
Make the RSA ASYM_CIPHER implementation available inside the FIPS module

mattcaswell:
RSA ASYM_CIPHER was already available within the default provider. We now make it also available from inside the FIPS module.


crypto/rsa/rsa_local.h
```diff
@@ -29,6 +29,8 @@ DECLARE_ASN1_ITEM(RSA_PRIME_INFO)
 DEFINE_STACK_OF(RSA_PRIME_INFO)
 
 struct rsa_st {
+    OPENSSL_CTX *libctx;
```
levitte: Here be dragons. I'm really unsure about this, and would rather have libctx passed to the functions that need it...

providers/implementations/keymgmt/rsa_kmgmt.c
```diff
@@ -157,8 +158,9 @@ static int key_to_params(RSA *rsa, OSSL_PARAM_BLD *tmpl)
 static void *rsa_importkey(void *provctx, const OSSL_PARAM params[])
 {
     RSA *rsa;
+    OPENSSL_CTX *libctx = PROV_LIBRARY_CONTEXT_OF(provctx);
 
-    if ((rsa = RSA_new()) == NULL
+    if ((rsa = rsa_new_with_ctx(libctx)) == NULL
```
levitte: Here be dragons. I would much rather have a keymgmt structure that contains an OPENSSL_CTX and an RSA than embedding the former in the latter...

levitte:
My comments are gut feeling, I need a bit of time to think it through

crypto/rsa/rsa_local.h
mattcaswell: Not sure why. We've done the same thing in lots of other places, e.g. EVP_PKEY_CTX, BN_CTX, CRYPTO_EX_DATA, EC_GROUP (I stopped looking after that - the list goes on). I don't see why this one would be any different.

slontis: Its also in the dh and dsa structs.. Dragons be everywhere. Its ok.

crypto/rsa/rsa_oaep.c
```diff
@@ -272,13 +285,19 @@ int RSA_padding_check_PKCS1_OAEP_mgf1(unsigned char *to, int tlen,
         to[i] = constant_time_select_8(mask, db[i + mdlen + 1], to[i]);
     }
 
+#ifndef FIPS_MODE
     /*
      * To avoid chosen ciphertext attacks, the error message should not
      * reveal which kind of decoding error happened.
+     *
+     * This trick doesn't work in the FIPS provider because libcrypto manages
+     * the error stack. Instead we opt not to put an error on the stack at all
```
slontis: There is nothing stopping this being a provider function..

slontis:
Need to change some errors to use 0 as the flags to fix the travis failures.

crypto/rsa/rsa_local.h
levitte: @slontis, that is incorrect, the DH and DSA do *not* contain a `OPENSSL_CTX` pointer.

levitte: We do have `OPENSSL_CTX` pointers in some context structures, and that's ok, those are fairly short-lived structures that are meant to use for a specific operation, and it's quite all right that they are tied to a specific library context.

When it comes to key structures (which I view as quite different from context structures), I'm not so sure.  I'll write a discussion point not tied to the diff.

levitte:
Regarding `OPENSSL_CTX *` in key structure, that raises questions.  Are low level keys suddenly tied to library context?  Why?  Does that extend to an EVP_PKEY?

I would rather see that the library functions that work with keys (not *on* keys) and that need a library context simply take a pointer as an argument.  After all, for all operation functions that work with asymmetric keys and need a context to perform their work, the context creating function gets the current library context already vi the input provider context.

Note that most KEYMGMT functions have no need whatsoever for a library context.  The exception I see is key generation, but I've built that up to work with a context of its own, similar to the way other operations usually work.

levitte:
Someone mentioned an `OPENSSL_CTX` pointer in `EC_GROUP`.  I forget if I paid attention at the time.  I imagine that I would have had the same questions then.

crypto/rsa/rsa_lib.c
```diff
@@ -50,6 +53,17 @@ int RSA_set_method(RSA *rsa, const RSA_METHOD *meth)
...
+    return rsa_new_intern(engine, NULL);
+}
+#endif
+
+RSA *rsa_new_with_ctx(OPENSSL_CTX *libctx)
+{
+    return rsa_new_intern(NULL, libctx);
+}
+
+static RSA *rsa_new_intern(ENGINE *engine, OPENSSL_CTX *libctx)
```
levitte: I think we all need to get into the habit of routinely changing calls like `FOOerr(FOO_F_WHATEVER, whatever)` to `FOOerr(0, whatever)` or `ERR_raise(ERR_LIB_FOO, whatever)`

richsalz: See #9496 which has scripts and instructions to do the conversions automatically, including a final "don't look back fix it all" just before the first (alpha?) release.

crypto/rsa/rsa_local.h
slontis: I was thinking of PROV_DSA_CTX & PROV_DH_CTX (in PR #10826)

levitte: I hope it's clear by now that I have no problem whatsoever with an `OPENSSL_CTX` pointer in those context structure.

mattcaswell: It is not straight forward to "pass the libctx to functions that need it". For example `rsa_ossl_public_decrypt` requires the libctx, but is an implementation of `rsa_pub_dec` from RSA_METHOD. Changing the signature of these functions is a breaking change. There are possible workarounds involving adding new functions to RSA_METHOD - but that's really nasty and messy and we should avoid doing this if at all possible. Adding the libctx to the RSA structure is simple and straight forward and we've done exactly that in many of the structures that I've previously mentioned. I see no reason why we can't do the same here (in fact I can't think of any downside at all - so I'm not sure where your objection is coming from).

levitte: "many other structures" mostly being context structures, which I view as a different kind of beast.
As for `rsa_ossl_public_decrypt`, it's perfectly possible to make a variant of it that takes an extra `OPENSSL_CTX` parameter, and have `rsa_ossl_public_decrypt` call that with NULL.  This is also something we do all over the place.

levitte: Believe you me, I do understand the luring easier way, and there are many place where we do that, but I feel that for these structures, which are payload and long lived rather than context and fairly short lived, we're walking onto dangerous grounds.  Also, like I try to discuss separately from the diff, there are moments where we suddenly have two libctx to deal with, one being in the provider operation context (what we often call provctx) and one being in the key, and I dread the moment we lose control and they are suddenly different.

levitte:
[Continued discussion on having  `OPENSSL_CTX` pointer in a key structure]

Something that's been on my mind for some time is that if we have an EVP_PKEY that was created with the DSA keymgmt loaded via one library context, and then use it via another library context (say, DSA signature), we end up fetching a keymgmt via that second library context, export it from the first keymgmt, import it with the second keymgmt (effectively making a copy on the fly), and then use that copy.  This, *even thought it's exactly the same DSA structure with exactly the same contents*.

It is reasonable to think that a key should be directly shareable between library contexts, as long as it's really the same keymgmt, i.e. the same code handling the same structure.  We could optimize the check of which keymgmt is used and decide to just use the provider side key as is instead of doing an unnecessary export and import of the key data.

However, if we tie a library context to the provider side keys (which is what happens when we add an `OPENSSL_CTX` pointer to a key structure), we make that optimization impossible.

mattcaswell:
> Something that's been on my mind for some time is that if we have an EVP_PKEY that was created with the DSA keymgmt loaded via one library context, and then use it via another library context (say, DSA signature), we end up fetching a keymgmt via that second library context, export it from the first keymgmt, import it with the second keymgmt (effectively making a copy on the fly), and then use that copy. This, even thought it's exactly the same DSA structure with exactly the same contents.

I think we should be strongly discouraging mixing objects from different libctxs together like this. The OPENSSL_CTX was always envisaged as a "scope". Mixing objects from one scope in another doesn't sound like a great idea to me.

> It is reasonable to think that a key should be directly shareable between library contexts, as long as it's really the same keymgmt, i.e. the same code handling the same structure. We could optimize the check of which keymgmt is used and decide to just use the provider side key as is instead of doing an unnecessary export and import of the key data.

We *could* do this. I'm not sure we should feel we have to if we discourage the mixing of objects anyway.

> However, if we tie a library context to the provider side keys (which is what happens when we add an OPENSSL_CTX pointer to a key structure), we make that optimization impossible.

Note that this PR does not do that for objects on the libcrypto side of things. It adds a libctx, but provides no additional public APIs for setting or getting that libctx - so it is effectively unused there. The internal interface for this is only ever called by providers (default or FIPS). It should never be the case that objects from different libctxs are mixed on the provider side.

levitte:
> The OPENSSL_CTX was always envisaged as a "scope".

Essentially, with that, you're saying that a low level key (RSA, DSA, DH) should only be used within the library context it's been loaded for.  That to be used within another library context, it should somehow be loaded again for that library context, even though it comes from the exact same source (file, say), contains the exact same payload, etc?

I utterly disagree with that idea.  If for nothing else, it adds unnecessary churn, at the very least possible export+import every time you use an EVP_PKEY in the "wrong" library context.

richsalz:
> . That to be used within another library context, it should somehow be loaded again for that library context

As we have no idea on what providers are doing, nor are any real constraints imposed upon them, this is the only safe thing to do.  Doing otherwise would require applications to have detailed knowledge of providers *at run time.*

If we arae concerned about the "inefficiency" of re-loading a key, then (a) don't be; (b) address it in a future release by having a "create key from key" operation.

levitte:
> As we have no idea on what providers are doing, nor are any real constraints imposed upon them, this is the only safe thing to do. Doing otherwise would require applications to have detailed knowledge of providers at run time.

Huh?  Tying them to a library context *is* a constraint.

There's another concern that came to mind this evening: what about providers that can't, for reasons of their own, export key material?  That will make some keymgmt functions (comparing keys, which I implement in #10807) quite difficult, and randomly so from an application point of view.  Typical such providers?  HSMs come to mind...

t-j-h:
A libcontext is basically a place to have a common set of things which implementations can refer to in order to gain access to stuff without having to make a pile of stuff essentially global. That it its purpose. It also allows for different choices of what those things can be inside a single application - i.e. we aren't hardwiring a fixed set of choices for an application in a library context. It is inherently an interface for unrelated items to be able to relate to each other without having a tight coupling between the items.

There should be no concept that things in one library context have any limitation on usage within another library context. The library context is how a given object finds other stuff it might need without hard bindings. It isn't about a strict containment separation between all things that relate or reference a library context.

Placing any limitation such that sharing isn't permitted doesn't make a lot of sense to me - and would be implying a meaning onto the library context that certainly was never in my conceptual model. 

Separately, I think everything should have reference to a library context and that the library context should be how we interact with the "environment" or gain reference to anything other than "ourselves" as that is its point/purpose. Effectively we should be able to move to a concept where there is nothing global at all ...

crypto/rsa/rsa_local.h
levitte: I'm still sceptical re the necessity for this, but figured that it's not as bad as I first believed.  So ok, I guess we'll see down the line what this gives...

crypto/rsa/rsa_lib.c
levitte: Yeah, noted.  However, nothing stops us from doing the small scale change when we stumble upon it, 'cause otherwise, we often end up with Travis failures because the `_F_` macro used no longer matches the function name.  This is one of those.

levitte:
I've decided to let go of my libctx concerns, it turns out that the tie doesn't have to be as tight as I made it out.  We'll see later if we need to rethink this.

levitte:
Please fix the Travis failure.  It's a simple modification of some error reporting in `rsa_new_intern`

mattcaswell:
Fixup pushed to address the travis failure. Please take another look.

slontis:
The RSA code has changed a bit since you started this PR :)

The following changes are also required.. 

crypto/rsa/rsa_sp800_56b_check.c              | 4 ++--
 crypto/rsa/rsa_sp800_56b_gen.c                | 2 +-
 providers/implementations/keymgmt/build.info  | 6 +++++-
 providers/implementations/keymgmt/rsa_kmgmt.c | 5 +++++
 4 files changed, 13 insertions(+), 4 deletions(-)

diff --git a/crypto/rsa/rsa_sp800_56b_check.c b/crypto/rsa/rsa_sp800_56b_check.c
index 93d3103..e272840 100644
--- a/crypto/rsa/rsa_sp800_56b_check.c
+++ b/crypto/rsa/rsa_sp800_56b_check.c
@@ -269,7 +269,7 @@ int rsa_sp800_56b_check_public(const RSA *rsa)
         return 0;
     }
 
-    ctx = BN_CTX_new();
+    ctx = BN_CTX_new_ex(rsa->libctx);
     gcd = BN_new();
     if (ctx == NULL || gcd == NULL)
         goto err;
@@ -358,7 +358,7 @@ int rsa_sp800_56b_check_keypair(const RSA *rsa, const BIGNUM *efixed,
         return 0;
     }
 
-    ctx = BN_CTX_new();
+    ctx = BN_CTX_new_ex(rsa->libctx);
     if (ctx == NULL)
         return 0;
 
diff --git a/crypto/rsa/rsa_sp800_56b_gen.c b/crypto/rsa/rsa_sp800_56b_gen.c
index 5474aaa..1f8d01d 100644
--- a/crypto/rsa/rsa_sp800_56b_gen.c
+++ b/crypto/rsa/rsa_sp800_56b_gen.c
@@ -297,7 +297,7 @@ int rsa_sp800_56b_generate_key(RSA *rsa, int nbits, const BIGNUM *efixed,
     if (!rsa_sp800_56b_validate_strength(nbits, -1))
         return 0;
 
-    ctx = BN_CTX_new();
+    ctx = BN_CTX_new_ex(rsa->libctx);
     if (ctx == NULL)
         return 0;
 
diff --git a/providers/implementations/keymgmt/build.info b/providers/implementations/keymgmt/build.info
index f4a9d1a..a96be30 100644
--- a/providers/implementations/keymgmt/build.info
+++ b/providers/implementations/keymgmt/build.info
@@ -11,4 +11,8 @@ ENDIF
 IF[{- !$disabled{dsa} -}]
   SOURCE[$DSA_GOAL]=dsa_kmgmt.c
 ENDIF
-SOURCE[$RSA_GOAL]=rsa_kmgmt.c
+
+# TODO(3.0) Remove once PSS is supported in FIPS_MODE
+#SOURCE[$RSA_GOAL]=rsa_kmgmt.c
+SOURCE[../../libfips.a]=rsa_kmgmt.c
+SOURCE[../../libnonfips.a]=rsa_kmgmt.c
diff --git a/providers/implementations/keymgmt/rsa_kmgmt.c b/providers/implementations/keymgmt/rsa_kmgmt.c
index a707b13..187d718 100644
--- a/providers/implementations/keymgmt/rsa_kmgmt.c
+++ b/providers/implementations/keymgmt/rsa_kmgmt.c
@@ -286,7 +286,12 @@ static int rsa_get_key_params(void *key, OSSL_PARAM params[])
     }
 #endif
     if ((p = OSSL_PARAM_locate(params, OSSL_PKEY_PARAM_DEFAULT_DIGEST)) != NULL
+/* TODO(3.0): Support PSS in FIPS_MODE */
+#ifndef FIPS_MODE
         && RSA_get0_pss_params(rsa) == NULL)
+#else
+        )
+#endif /* FIPS_MODE */
         if (!OSSL_PARAM_set_utf8_string(p, RSA_DEFAULT_MD))
             return 0;
 
-- 
2.7.4

crypto/rsa/rsa_oaep.c
slontis: Does this need to be addressed?

levitte:
Re clearerror, I think that last time this was discussed, we found that it was better to use the set_mark / clear_last_mark / pop_to_mark functions in library code.

slontis:
reconfirmed

levitte:
> Fixup pushed to address the travis failure.

Travis still fails in exactly the same way...  or at least the first job does, I frankly didn't look further

mattcaswell:
> Re clearerror, I think that last time this was discussed, we found that it was better to use the set_mark / clear_last_mark / pop_to_mark functions in library code.

This is a different issue because we're talking about the *constant time* clearing of an error - which we can't currently do in the FIPS provider. My work around for this is to just not put an error in the stack in the first place.

slontis:
> UPDATE: it fails in a new way

My changes listed above should get it to work..

Your workaround for the clearerror sounds good enough for now..

mattcaswell:
I rebased this and add the various fixes suggested by @slontis. Lets see what travis makes of it now.

Please take another look.

iamamoose:

slontis:
needs a rebase

mattcaswell:
Rebased. @slontis - please reconfirm

include/openssl/rsaerr.h
```diff
@@ -155,6 +155,7 @@ int ERR_load_RSA_strings(void);
 # define RSA_R_MP_EXPONENT_NOT_CONGRUENT_TO_D             169
 # define RSA_R_MP_R_NOT_PRIME                             170
 # define RSA_R_NO_PUBLIC_EXPONENT                         140
+# define RSA_R_NULL_ARGUMENT                              180
```
levitte: There is already an `ERR_R_PASSED_NULL_PARAMETER`, why do we need to have the same but specific for RSA?

Hint: `ERR_R_` reason codes can be used globally

slontis:
reapproved.. Up to you whether you fit the nit above..

include/openssl/rsaerr.h
mattcaswell: Fixed

mattcaswell:
> reapproved.. Up to you whether you fit the nit above..

I did fix the error issue noted above and have pushed a fixup. From your comment it seems your approval still counts, so I will set the "approval done" label.

mattcaswell:
Pushed. Thanks!

PR #10882, opened 2020-01-17
Detect EOF while reading in libssl (1.1.1)

### LINKED ISSUE #10880: Data loss with TLS 1.3
Hi,

## OpenSSL version:
```
OpenSSL 1.1.1d  10 Sep 2019
built on: Wed Nov 13 16:09:29 2019 UTC
platform: linux-x86_64
options:  bn(64,64) rc4(16x,int) des(int) idea(int) blowfish(ptr)
compiler: gcc -fPIC -pthread -m64 -Wa,--noexecstack -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -Wa,--noexecstack -D_FORTIFY_SOURCE=2 -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now -DOPENSSL_USE_NODELETE -DL_ENDIAN -DOPENSSL_PIC -DOPENSSL_CPUID_OBJ -DOPENSSL_IA32_SSE2 -DOPENSSL_BN_ASM_MONT -DOPENSSL_BN_ASM_MONT5 -DOPENSSL_BN_ASM_GF2m -DSHA1_ASM -DSHA256_ASM -DSHA512_ASM -DKECCAK1600_ASM -DRC4_ASM -DMD5_ASM -DVPAES_ASM -DGHASH_ASM -DECP_NISTZ256_ASM -DX25519_ASM -DPOLY1305_ASM -DNDEBUG -D_FORTIFY_SOURCE=2
OPENSSLDIR: "/etc/ssl"
ENGINESDIR: "/usr/lib/engines-1.1"
Seeding source: os-specific
```

## OS Name, Version, Hardware platform:
Arch Linux, Ubuntu Bionic x86_64

## Compiler Details (name, version):
gcc version 9.2.0

## Problem Description:

When a client sends data to a TLS 1.3 server and then closes the connection, the server won't read everything that was reported as sent on the client side. The amount of data read is non-deterministic.

Reproduction steps:

0. Ignore SIGPIPE
1. TLS client connects to server
2. TLS client sends data (in small chunks, using multiple `SSL_write()` calls)
3. TLS client closes the connection right after returning from all successful (blocking) `SSL_write()` calls
    - Delaying `close()` on the client side with `nanosleep()` or executing a complete TCP shutdown (`shutdown(SHUT_RDWR)`) decreases the possibility of data loss. I guess it's just because this issue has something to do with handling `ECONNRESET` and "new session ticket" messages.

A dummy example can be found here that reproduces the issue (localhost, loopback interface):
https://github.com/MrAnno/openssl-tls13-test/commit/3b101e3441bf69c3e47e73fcea2116718e818119

The data loss usually occurs if the server returns from a blocking `SSL_read()` with `SSL_ERROR_SYSCALL`:

```
$ ./tls13test
SSL_read returned SSL_ERROR_SYSCALL: Connection reset by peer
read: 3672 (should be 4080)
```

The reason I think it's a bug (even after reading #6904) is because `SSL_read()` sometimes returns with `SSL_ERROR_SYSCALL`, but the error queue and errno are empty:

```
$ ./tls13test
SSL_read returned SSL_ERROR_SYSCALL: Success
read: 3876 (should be 4080)
```

mattcaswell:
If we hit an EOF while reading in libssl then we will report an error
back to the application (SSL_ERROR_SYSCALL) but errno will be 0. We add
an error to the stack (which means we instead return SSL_ERROR_ssl) and
therefore give a hint as to what went wrong.

WIP because I'd like to check if this actually solves the problem. Also this currently targets 1.1.1 to ease testing by the reporter.

Darn...forgot to add "Fixes #10880"


levitte:
> Darn...forgot to add "Fixes #10880"

Nothing stops you amending the commit :wink:

ssl/record/rec_layer_s3.c
```diff
@@ -296,6 +296,12 @@ int ssl3_read_n(SSL *s, size_t n, size_t max, int extend, int clearold,
             ret = BIO_read(s->rbio, pkt + len + left, max - left);
             if (ret >= 0)
                 bioread = ret;
+            if (ret <= 0
+                    && !BIO_should_retry(s->rbio)
+                    && BIO_eof(s->rbio)) {
```
levitte: Why isn't the "should retry" test enough?

mattcaswell: Because if we haven't hit eof  then we want to be returning SSL_ERROR_SYSCALL and have the application check errno to discover the problem. However if we hit eof then errno is not set do we want to add our own error to the stack (in order to return SSL_ERROR_SSL)

mattcaswell:
So the master version of this has gone in (#10907). 

I have updated this backport to 1.1.1 with the final version that went in. I am in two minds about whether this is a good idea to backport or not. It is strictly speaking a bug - but I worry that a change from the old behaviour in a stable branch might surprise a lot of people.

Ping @openssl/otc. Thoughts?

kaduk:
I'm not on the OTC, but I do agree with the sentiment that this is "just" a bug fix, which would tend to put it on the queue for 1.1.1.

t8m:
I think the change of behavior is acceptable for 1.1.1. There will be error return anyway, just a different one but with similar handling requirements on the calling application. So no, I do not think this bug fix would be problematic on 1.1.1.

mattcaswell:
Ok. Based on @t8m's approval I've taken this out of WIP.

Just as a safety check I'd like a second approval on this PR if possible.

Ping @openssl/committers.

mattcaswell:
Thanks@beldmit. It seems there is sufficient consensus that this should be backported, so I added the "approval: done" label.

mattcaswell:
Pushed. Thanks.

jay:
Ref: curl/curl#5138

This may be an issue for us in curl, we allow server truncation by ignoring SYSCALL without an error. Server truncation is very unlikely but continues to be supported for legacy reasons, except in debug builds. Frankly I'd rather we didn't support it at all but we had some reports in the recent past. Regardless it was our intention to eventually treat truncation as an error with later versions of OpenSSL (which I guess would be 3). At this point I think I will suggest we wait for more reports, just to get an idea if this is an issue for curl users that use 1.1.1. 

/cc @bagder

vdukhovni:
> This may be an issue for us in curl, we allow server truncation by ignoring SYSCALL without an error. Server truncation is very unlikely but continues to be supported for legacy reasons, except in debug builds. Frankly I'd rather we didn't support it at all but we had some reports in the recent past. Regardless it was our intention to eventually treat truncation as an error with later versions of OpenSSL (which I guess would be 3). At this point I think I will suggest we wait for more reports, just to get an idea if this is an issue for curl users that use 1.1.1.

FWIW, in the internal OpenSSL builds I maintain for my employer I chickened out and reverted the associated commits.  In part because I had an unrelated internal issue to address, and did not want to risk having to roll that back if this one bit us.

And my sentiment is roughly in line with yours: would be nice to not have applications unwittingly oblivious of truncation attacks, but risky to spring the mitigation on them in a shared-library update, when by the far the more likely scenario is that the data source is doing sloppy session termination.

This should not affect HTTP though, since HTTP is self-framing with Content-Length, or chunked encoding, but could affect other protocols.  Or am I missing some corner case in HTTP?

jay:
> This should not affect HTTP though, since HTTP is self-framing with Content-Length, or chunked encoding, but could affect other protocols. Or am I missing some corner case in HTTP?

Yes legacy servers (or middleboxes maybe), truncation without an HTTP or TLS termination point. We had some reports and reverted to less sensitivity in curl/curl@78cef06.

kroeckx:
> This should not affect HTTP though, since HTTP is self-framing with Content-Length, or chunked encoding, but could affect other protocols.  Or am I missing some corner case in HTTP?

As far as I know, there is a corner case in HTTP 1.1, but not
anymore in 2, but I don't remember the details.

bagder:
That's correct, it *shouldn't* affect HTTP but it certainly and positively unconditionally will. The world of HTTP(S) is wild and crazy and every imaginable violation of the protocol exists out there. Those violations that we (as a collective) have previously silently accepted will of course especially prevail.

I spent [a significant effort](https://daniel.haxx.se/blog/2015/02/12/tightening-firefoxs-http-framing-again/) years ago while employed at Mozilla to try to make Firefox's HTTP code stricter when it comes to this HTTP framing, but eventually had to give up most of the tries simply because there are so many deployed server-side things not adhering and users are not very forgiving when one tool says there's an error with the server while the others don't...

HTTP/2 has much less wiggle room in this department.

mattcaswell:
Please track issues with this in #11378 rather than here. The current proposal there is to revert this in 1.1.1 but keep it in 3.0. Hearing opinions in that issue would be helpful.

kaduk:
> This should not affect HTTP though, since HTTP is self-framing with Content-Length, or chunked encoding, but could affect other protocols. Or am I missing some corner case in HTTP?

@vdukhovni there are also plain old bugs, like the one in Go's net/http mentioned at https://discord.statuspage.io/incidents/62gt9cgjwdgf

(Yes, I saw Matt's request to track issues elsewhere; this is not realted to issue tracking directly.)

PR #10883, opened 2020-01-17
Do not silently truncate files on perlasm errors

davidben:
If one of the perlasm xlate drivers crashes, OpenSSL's build will currently swallow the error and silently truncate the output to however far the driver got. This will hopefully fail to build, but better to check such things.

Handle this by checking for errors when closing `STDOUT` (which is a pipe to the xlate driver).

### DIFF
--- a/crypto/aes/asm/aes-586.pl
+++ b/crypto/aes/asm/aes-586.pl
@@ -2995,4 +2995,4 @@ sub deckey()
 
 &asm_finish();
 
-close STDOUT;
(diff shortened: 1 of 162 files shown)


t8m:
Merged to master as 32be631ca1

t8m:
Should it be also done on 1.1.1? However there are conflicts so a separate PR would be needed.

davidben:
I can put together a separate PR for 1.1.1. It's just a `sed` line. :-)

PR #9677, opened 2019-08-23
Remove x86/x86_64 BSAES and AES_ASM support

bernd-edlinger:
This leaves VPAES and AESNI support.
The VPAES performance is comparable but BSAES is not
completely constant time. There are table lookups
using secret key data in AES_set_encrypt/decrypt_key
and in ctr mode short data uses the non-constant
time AES_encrypt function instead of bit-slicing.
Furthermore the AES_ASM is by far outperformed
by recent GCC versions.
Since BSAES calls back to AES_ASM for short
data blocks the performance on those is also
worse than the pure software implementaion.

Fixes: #9640

##### Checklist

- [ ] documentation is added or updated
- [ ] tests are added or updated

PR #10884, opened 2020-01-17
Add vpaes_ctr32_encrypt_blocks for x86_64

davidben:
This adds an analog to the (currently unused) "2x" optimization from vpaes-armv8.pl and a ctr128_f implementation.

This PR probably should *not* be merged as-is. I haven't integrated it into the C portions or even tested that it works in OpenSSL. Rather, this PR is so @bernd-edlinger can look into integrating it into #9677 if it helps remove bsaes-x86_64.pl which, while constant-time, is currently integrated with other bits that are not constant-time and is a little awkward at small inputs.

In BoringSSL, we found that, while bsaes-x86_64.pl is still faster, the gap became small enough that we were comfortable removing bsaes-x86_64.pl and reducing the number of AES implementations we carried around. Perhaps this'll help you all too.

See also https://github.com/openssl/openssl/pull/9677#issuecomment-575374166 and https://boringssl-review.googlesource.com/c/boringssl/+/35364. (Note that, by then, we had already batched the non-constant-time fallback for small inputs in `bsaes_ctr32_encrypt_blocks`.)

##### Checklist

### DIFF
--- a/crypto/aes/asm/vpaes-x86_64.pl
+++ b/crypto/aes/asm/vpaes-x86_64.pl
@@ -178,6 +178,181 @@ _vpaes_encrypt_core:
 .size	_vpaes_encrypt_core,.-_vpaes_encrypt_core
 
 ##
+##  _aes_encrypt_core_2x
+##
+##  AES-encrypt %xmm0 and %xmm6 in parallel.
+##
+##  Inputs:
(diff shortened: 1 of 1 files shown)


bernd-edlinger:
Yeah, thanks!  I integrated this into #9677 and it seems to work.

PR #10191, opened 2019-10-16
Refactor SSL/TLS connection options

richsalz:
This makes the ordering follow the same as the man1 documentation.

Still to come, another PR that does editing.

This depends on #10159

PR #10885, opened 2020-01-18
WIP: Re-order options in s_client

richsalz:
There are over 175 options in this command. One hundred and seventy bleedin' five.  And 45 of them aren't documented.  This PR is intended to add that documentation.

The first commit re-orders the options to put them in the same order the ```-help``` output, and puts them into sub-heads.  Subsequent commits will add the missing documentation.

This PR is based on #10873, but it also probably needs to wait for #10132, and #10191 as well.


richsalz:
I changed the title and took this out of WIP.  I think it makes the most sense to do this one PR, that simply "fixes" the ordering.  And then a followon PR will add the missing options. I still think it will be best if the PR's above are reviewed and merged first.

richsalz:
Rebased because #10873 was rebased onto master.  So this is only waiting on #10191.  Still worth looking at the "blanks lines in synopsis" :)

richsalz:
Back to WIP.  Just leaving this open so that folks can look at s_client.pod and see what they think about re-ordering and blank lines in sections.

richsalz:
I am going to close this.  See #11033.  I took a poll and got three replies, not enough to make conclusions.

FdaSilvaYY:
Ping @openssl There a lot of documentation fixup here. but this PR is very large. 
@richsalz : nice job :+1:  yours  PR deserves to be split, to ease its review.  the s_client documentation fix-up deserves its own PR ;)

richsalz:
All the other doc fixes have happened.  I closed this because there did not seem to be a great deal of interest and it's lots of low-level tedious work.  I would be willing to do this for all "long" commands (where there were more than "n" options, like 50?) if it got support, particularly from the OTC.  But see also #11033 for another attempt.

PR #10887, opened 2020-01-18
add FFDH to speed command

### LINKED ISSUE #9475: speed should measure performance of non elliptic curve Diffie Hellman
Title says it all really. I should be able to compare 2048/3072/4096/etc. bit DHE against P-256/X25519/etc. ECDHE using `openssl speed`, like I can compare speed of RSA sign/verify against ECDSA and EdDSA.

tomato42:
the openssl speed command could not benchmark FFDH speed, but it could
benchmark ECDH, making comparisons between the two hard

this commit adds this feature

fixes #9475

### DIFF
--- a/apps/speed.c
+++ b/apps/speed.c
@@ -16,6 +16,7 @@
 #define ECDH_SECONDS    10
 #define EdDSA_SECONDS   10
 #define SM2_SECONDS     10
+#define FFDH_SECONDS    10
 
 /* We need to use some deprecated APIs */
 #define OPENSSL_SUPPRESS_DEPRECATED
@@ -98,6 +99,9 @@
 # include <openssl/rsa.h>
 # include "./testrsa.h"
 #endif
+#ifndef OPENSSL_NO_DH
+# include <openssl/dh.h>
+#endif
 #include <openssl/x509.h>
 #if !defined(OPENSSL_NO_DSA) && !defined(OPENSSL_NO_DEPRECATED_3_0)
 # include <openssl/dsa.h>
@@ -125,6 +129,7 @@
 #define MAX_MISALIGNMENT 63
 #define MAX_ECDH_SIZE   256
 #define MISALIGN        64
+#define MAX_FFDH_SIZE 1024
 
 typedef struct openssl_speed_sec_st {
     int sym;
@@ -134,6 +139,7 @@ typedef struct openssl_speed_sec_st {
     int ecdh;
     int eddsa;
     int sm2;
+    int ffdh;
 } openssl_speed_sec_t;
 
 static volatile int run = 0;
@@ -435,6 +441,22 @@ static const OPT_PAIR rsa_choices[RSA_NUM] = {
 static double rsa_results[RSA_NUM][2];  /* 2 ops: sign then verify */
 #endif /* OPENSSL_NO_RSA */
 
+#ifndef OPENSSL_NO_DH
+enum ff_params_t {
+    R_FFDH_2048, R_FFDH_3072, R_FFDH_4096, R_FFDH_6144, R_FFDH_8192, FFDH_NUM
+};
+
+static const OPT_PAIR ffdh_choices[FFDH_NUM] = {
+    {"ffdh2048", R_FFDH_2048},
+    {"ffdh3072", R_FFDH_3072},
+    {"ffdh4096", R_FFDH_4096},
+    {"ffdh6144", R_FFDH_6144},
+    {"ffdh8192", R_FFDH_8192},
+};
+
+static double ffdh_results[FFDH_NUM][1];  /* 1 op: derivation */
+#endif /* OPENSSL_NO_DH */
+
 #ifndef OPENSSL_NO_EC
 enum ec_curves_t {
     R_EC_P160, R_EC_P192, R_EC_P224, R_EC_P256, R_EC_P384, R_EC_P521,
@@ -561,6 +583,11 @@ typedef struct loopargs_st {
     unsigned char *secret_a;
     unsigned char *secret_b;
     size_t outlen[EC_NUM];
+#endif
+#ifndef OPENSSL_NO_DH
+    EVP_PKEY_CTX *ffdh_ctx[FFDH_NUM];
+    unsigned char *secret_ff_a;
+    unsigned char *secret_ff_b;
 #endif
     EVP_CIPHER_CTX *ctx;
 #ifndef OPENSSL_NO_DEPRECATED_3_0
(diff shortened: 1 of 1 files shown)


tomato42:
hmm, looks like my CLA is only on my work email, not my private one – I did sign the individual CLA, so I don't think it's necessary to send a new one, is it?

mattcaswell:
Close/reopen to kick CLA bot

mattcaswell:
Try again...

mattcaswell:
> @mattcaswell hmm, looks like my CLA is only on my work email, not my private one – I did sign the individual CLA, so I don't think it's necessary to send a new one, is it?

No, everything is already in place. Your private email address was in our database but with a spelling error which I have now corrected.

t8m:
Hi Hubert, can you please rebase the PR? I am sorry for neglecting it, I will review it after the rebase.

tomato42:
@t8m updated

apps/speed.c
```diff
@@ -1066,6 +1094,24 @@ static int RSA_verify_loop(void *args)
...
+static int FFDH_derive_key_loop(void *args)
+{
+        loopargs_t *tempargs = *(loopargs_t **) args;
+        DH *privkey = tempargs->ffdh_ctx[testnum];
+        BIGNUM *pubkey = tempargs->ffdh_pubkey[testnum];
+        unsigned char *derived_secret = tempargs->secret_ff_a;
+        int count;
+
+        for (count = 0; COND(ffdh_c[testnum][0]); count++)
+            DH_compute_key(derived_secret, pubkey, privkey);
```
mattcaswell: Should we be adding *new* uses of deprecated APIs? speed is kind-of special, but the idea is that all the other deprecated uses will be removed from speed as and when the deprecated APIs are also removed.

apps/speed.c
```diff
@@ -2122,6 +2205,20 @@ int speed_main(int argc, char **argv)
     }
 #  endif
 
+#  ifndef OPENSSL_NO_DH
+    ffdh_c[R_FFDH_2048][0] = count / 1000;
+    for (i = R_FFDH_3072; i <= R_FFDH_8192; i++) {
+        ffdh_c[i][0] = ffdh_c[i - 1][0] / 2;
+        if (ffdh_doit[i] <= 1 && ffdh_c[i][0] == 0)
+            ffdh_doit[i] = 0;
+        else {
```
mattcaswell: The "then" block above should also use {} because the else does.

apps/speed.c
```diff
@@ -2122,6 +2205,20 @@ int speed_main(int argc, char **argv)
...
+            if (ffdh_c[i][0] == 0) {
+                ffdh_c[i][0] = 1;
+            }
```
mattcaswell: No need to use {} here because its a single line "if"

apps/speed.c
```diff
@@ -2953,6 +3050,105 @@ int speed_main(int argc, char **argv)
...
+            if (!key_B) {
+                BIO_printf(bio_err, "Error while initialising FFDH context.\n");
+                ERR_print_errors(bio_err);
+                rsa_count = 1;
+                ffdh_checks = 0;
+                break;
+            }
+
+            if (1 != DH_generate_key(key_A) ||
+                1 != DH_generate_key(key_B)) {
```
mattcaswell: More usage of deprecated APIs here - so same question as above.

mattcaswell:

apps/speed.c
tomato42: I was looking for way how to do it with the EVP API, but didn't found definitions for the FFDHE parameters from the RFC.

tomato42: @mattcaswell did I miss that documentation or it's not there?

mattcaswell: You can still use `DH_new_by_nid()` (that is *not* deprecated). You can then use `EVP_PKEY_assign_DH` to create an EVP_PKEY from it. Then you can use EVP_PKEY_derive* to derive a shared key.

tomato42:
will continue work on this late next week

apps/speed.c
tomato42: fixed

apps/speed.c
tomato42: fixed

apps/speed.c
tomato42: fixed

apps/speed.c
tomato42: I wasn't able to make `EVP_PKEY_assign_DH` work with current master, I've used `EVP_PKEY_CTX_new_id(EVP_PKEY_DH, NULL)` in the end

tomato42:
@mattcaswell sorry for the delay, COVID kept me distracted. I've updated the PR to not use the deprecated function and rebased on top of master.

tomato42:
(side note: the reported times are buggy because of #11938 )

tomato42:
rebased on top of master to fix #11903

apps/speed.c
mattcaswell: > I wasn't able to make EVP_PKEY_assign_DH work with current master, I've used EVP_PKEY_CTX_new_id(EVP_PKEY_DH, NULL) in the end

Hmm. Ok...although I don't understand why `EVP_PKEY_assign_DH` did not work...it *should* do

tomato42: @mattcaswell IIRC without calling `EVP_PKEY_CTX_set_dh_nid` the `EVP_PKEY_keygen` complained about missing parameters in context and when both `EVP_PKEY_assign_DH` and `EVP_PKEY_CTX_set_dh_nid` were used the `EVP_PKEY_keygen` failed without logging errors (`ERR_print_errors` didn't print anything)

apps/speed.c
```diff
@@ -2958,6 +3053,188 @@ int speed_main(int argc, char **argv)
...
+        }
+        if (ffdh_checks != 0) {
+            pkey_print_message("", "ffdh", ffdh_c[testnum][0],
+                               ffdh_params[testnum].bits, seconds.ffdh);
+            Time_F(START);
+            count =
+                run_benchmark(async_jobs, FFDH_derive_key_loop, loopargs);
+            d = Time_F(STOP);
+            BIO_printf(bio_err,
+                       mr ? "+R7:%ld:%d:%.2f\n" :
```
t8m, line 3224: You should move the whole block testing the FFDHE to be the last among the algos tested and change the R7 to R12. So the output in -multi mode is not confusing.

apps/speed.c
```diff
@@ -3606,6 +3883,27 @@ int speed_main(int argc, char **argv)
...
+    testnum = 1;
+    for (k = 0; k < FFDH_NUM; k++) {
+        if (!ffdh_doit[k])
+            continue;
+        if (testnum && !mr) {
+            printf("%23sop     op/s\n", " ");
+            testnum = 0;
+        }
+        if (mr)
+            printf("+F5:%u:%u:%f:%f\n",
```
t8m, line 3896: Again you should move the whole added hunk to be the last and change the F5 to be F8. And you need to add appropriate else if block with F8 in do_multi().

apps/speed.c
tomato42: done

apps/speed.c
tomato42: and done, also verified that -multi works as expected

tomato42:
@t8m `-multi` issues fixed and rebased on top of master

tomato42:
this Travis failure is weird:
```
apps/speed.c:1100:12: error: 'FFDH_derive_key_loop' defined but not used [-Werror=unused-function]
 static int FFDH_derive_key_loop(void *args)
            ^~~~~~~~~~~~~~~~~~~~
apps/speed.c:1098:13: error: 'ffdh_c' defined but not used [-Werror=unused-variable]
 static long ffdh_c[FFDH_NUM][1];
             ^~~~~~
Makefile:21518: recipe for target 'apps/openssl-bin-speed.o' failed
make[1]: *** [apps/openssl-bin-speed.o] Error 1
Makefile:2942: recipe for target 'build_sw' failed
make: *** [build_sw] Error 2
** FAILED -- MAKE
```
the `FFDH_derive_key_loop` is used on line 3774 and `ffdh_c` is used 4 lines above that (and in the `FFDH_derive_key_loop` function) so it looks like a false positive to me

mattcaswell:
The usage is (incorrectly) inside an `OPENSSL_NO_EC` guard, but the definition is not. So this fails in a no-ec build.

tomato42:
> The usage is (incorrectly) inside an `OPENSSL_NO_EC` guard, but the definition is not. So this fails in a no-ec build.

ah, yes, fixed

t8m:
Merged to master as 60d3b5b9ff. Thank you for the PR and the patience when waiting for the reviews.

PR #10890, opened 2020-01-19
DOC: document in more detail what a BIO_read_ex() via BIO_f_buffer() does

### LINKED ISSUE #10859: BIO_f_buffer() wants to read more from a blocking socket than is available

levitte:
The BIO_f_buffer() documentation tells in enough detail how it affects
BIO_gets(), but not how it affects BIO_read_ex().  This change
remedies that.

Fixes #10859

-----

This is an alternative to #10876

### DIFF
--- a/doc/man3/BIO_f_buffer.pod
+++ b/doc/man3/BIO_f_buffer.pod
@@ -49,10 +49,20 @@ is expanded.
 
 These functions, other than BIO_f_buffer(), are implemented as macros.
 
-Buffering BIOs implement BIO_gets() by using BIO_read_ex() operations on the
-next BIO in the chain. By prepending a buffering BIO to a chain it is therefore
-possible to provide BIO_gets() functionality if the following BIOs do not
-support it (for example SSL BIOs).
+Buffering BIOs implement BIO_read_ex() and BIO_gets() by using
+BIO_read_ex() operations on the next BIO in the chain and storing the
+result in an internal buffer, from which bytes are given back to the
+caller as appropriate for the call; a BIO_gets() is guaranteed to give
+the caller a whole line, and BIO_read_ex() is guaranteed to give the
+caller the number of bytes it asks for, unless there's an error or end
+of communication is reached in the next BIO.  By prepending a
+buffering BIO to a chain it is therefore possible to provide
+BIO_gets() or exact size BIO_read_ex() functionality if the following
+BIOs do not support it.
+
+Do not add more than one BIO_f_buffer() to a BIO chain.  The result of
+doing so will force a full read of the size of the internal buffer of
+the top BIO_f_buffer(), which is 4 KiB at a minimum.
 
 Data is only written to the next BIO in the chain when the write buffer fills
 or when BIO_flush() is called. It is therefore important to call BIO_flush()


Quuxplusone:
IMHO the parenthetical "(for example SSL BIOs)" should be dropped, given the current difficulty of using a buffer BIO with SSL or HTTP protocols. I mean, technically it is impossible only to use a buffer BIO _below_ an SSL BIO; it is technically possible to use a buffer BIO _above_ an SSL BIO as described in the parenthetical. But if you do that, you still can't do the HTTP protocol on top of that buffer BIO.

I think that ***if*** the current runtime behavior is left alone, then it should be documented additionally that

- `BIO_f_buffer` should be used only with memory and file BIOs, never with socket-related BIOs [that is, only with inputs that "know their size" and where reads never block]

- Especially, a buffer BIO should never be chained directly to another buffer BIO [this is the `HI–ME–LO` test case in #10859]

However, with that extra documentation, this would be an acceptable fix to me. It would indicate clearly-enough-for-me that "`BIO_f_buffer` is not to be used haphazardly and certainly never as a performance tweak; it is a very specialized tool with arcane semantics and an intended use-case that is _not_ related to buffering of network input."

The undocumented intended-use-case is that you can `BIO_gets` on a buffer BIO. That calls `buffer_gets`, which does _not_ suffer from the multiple-reads issue: it stops issuing reads to the lower-level BIO as soon as any read contains `'\n'`. The 7 buffer BIOs in `apps/s_client.c`, the 1 in `apps/s_server.c`, and the 1 in `apps/ocsp.c` are created for use with `BIO_gets`, and never use `BIO_read` as far as I can tell. However, if `CHARSET_EBCDIC` is set, then `s_server.c` defines a `BIO_f_ebcdic` whose read function is `ebcdic_read`, which very carefully issues reads of size `1` in a loop because it knows it might be atop a buffer BIO, and if it ever asks for 2 bytes, it risks hanging the server. (However, `ebcdic_read` is never actually called, AFAICT.)

`crypto/asn1/asn_mime.c` uses `BIO_read` on a buffer BIO inside the function `SMIME_crlf_copy`, which will hang if the `BIO *in` passed to `CMS_encrypt()` refers to a network socket. However, `CMS_encrypt()` is intended for use only with local file BIOs.

levitte:
@Quuxplusone, `BIO_f_buffer` could very well be used on top of a socket BIO, for example in a record oriented stream where you know that every record starts with a fixed-size length. I think that the suitability of using `BIO_f_buffer` should be left to the reader to decide.

Quuxplusone:
> @Quuxplusone, `BIO_f_buffer` could very well be used on top of a socket BIO, for example in a record oriented stream where you know that every record starts with a fixed-size length.

Right, a single level `BIO_f_buffer — socket` would happen to work. But `BIO_f_buffer` isn't idempotent; if you stacked `BIO_f_buffer — BIO_f_buffer — socket`, then your code would hang. (You commented as much [here in #10876](https://github.com/openssl/openssl/pull/10876#issuecomment-576138113).)  In short, buffer BIOs currently don't compose. They can't be used safely with any kind of BIO where "reading" has a side effect on the world. I'd like to see the actual bug fixed (#10876), plus test cases to ensure it doesn't regress back; but if the bugfix can't be done, then I'd like to see the documentation be _very very clear_ that buffer BIOs can't be composed with arbitrary other BIOs.

levitte:
I added a bit of extra text, advising not to add more than one `BIO_f_buffer()` to the chain, and why.  I hope that is satisfactory enough.

Quuxplusone:
@levitte: The added paragraph looks good (although IMVHO "4096 bytes" would be more human-readable than "4 KiB").

I still object to bits of the preceding paragraph:

> BIO_read_ex() is guaranteed to give the caller the number of bytes it asks for, unless there's an error in the next BIO

That's not exactly the case, right? The caller might not get the number of bytes it asked for, if the buffer BIO hangs forever. I would accept "BIO_read_ex() is guaranteed to block until it can return the exact number of bytes the caller asked for." However, I think nobody in this thread has yet provided any use-case where this behavior is _useful_.

>  By prepending a buffering BIO to a chain it is therefore possible to provide BIO_gets() or exact size BIO_read_ex() functionality if the following BIOs do not support it (for example SSL BIOs).

Again the close proximity of "exact size BIO_read_ex()" and "SSL BIOs" worries me, because we've seen that they don't actually work together like that. Without the patch in #10876, buffer BIOs really provide three orthogonal features:

- They buffer reads and writes, for performance.
- They add `BIO_gets()` functionality on top of the following BIO, with correct semantics (it reads until newline or error, and no further).
- They disable the usual semantics of `BIO_read()` and replace it with an exact-sized `BIO_read()/BIO_read_ex()` that blocks forever if the following BIO does not have that much data available right now.

The latter two bullet points are so orthogonal that I don't like to see them conflated in a single paragraph. (And the third bullet point is incompatible with SSL BIOs. _And_ I consider it a bug.)

_With_ the patch in #10876, buffer BIOs would provide only the first two bullet points.

levitte:
> that blocks forever

This is simply not true.  It blocks until one of the following conditions occurs:

1. the other end shuts down communications (EOF)
2. there's an error
3. the other end sends enough data

I'll duly note that you didn't test case 1 in #10859

levitte:
> (for example SSL BIOs)

I've removed that.

levitte:
Ping!

levitte:
Ping!

PR #10891, opened 2020-01-19
PROV: Fix bignum printout in text serializers

levitte:
The common routine ossl_prov_print_labeled_bignum() didn't print the
BIGNUM quite the way it should.  It treated the limbs in a big endian
fashion, when they are really organised in a little endian fashion.

Furthermore, we make it inherit the behaviour from the print of legacy
keys, where a number starting with the high bit set gets an extra zero
printed first.

### DIFF
--- a/providers/implementations/serializers/serializer_common.c
+++ b/providers/implementations/serializers/serializer_common.c
@@ -138,29 +138,29 @@ OSSL_OP_keymgmt_importkey_fn *ossl_prov_get_importkey(const OSSL_DISPATCH *fns)
 # endif
 
 int ossl_prov_print_labeled_bignum(BIO *out, const char *label,
-                                   const BIGNUM *n)
+                                   const BIGNUM *bn)
 {
     const char *neg;
     const char *post_label_spc = " ";
     int bytes;
     BN_ULONG *words;
-    int i, off;
+    int n, i;
 
-    if (n == NULL)
+    if (bn == NULL)
         return 0;
     if (label == NULL) {
         label = "";
         post_label_spc = "";
     }
 
-    bytes = BN_num_bytes(n);
-    words = bn_get_words(n);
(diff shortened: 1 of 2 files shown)


levitte:
This is yet another that got uncovered by #10797

providers/implementations/serializers/serializer_common.c
```diff
@@ -172,44 +172,53 @@ int ossl_prov_print_labeled_bignum(BIO *out, const char *label,
...
-        for (j = 0; i + j < bytes && j < 16; j += BN_BYTES - step) {
-            int k;
-            BN_ULONG l = words[(i + j + off) / BN_BYTES];
-
-            step = (i + j + off) % BN_BYTES;
-
-            for (k = step; k < BN_BYTES; k++) {
-                int o = 8 * (BN_BYTES - k - 1);
-                int b = ((l & (0xffLU << o)) >> o) & 0xff;
+    /* Keep track of how many bytes we have printed out so far */
```
slontis: Could this just use BN_bn2hex instead and then print it out formatted..? (It does an alloc but it would be much simpler)

mattcaswell:
Ping @levitte - there's an outstanding comment here from @slontis.

providers/implementations/serializers/serializer_common.c
levitte: BN_bn2hex creates one long line.  We *still* need to keep track of how many bytes are printed out so far, when to insert the colon and all that.  The difference will be very little, at the cost of an unnecessary malloc().  I'm not sure there's much to gain with that...

mattcaswell (approved):

levitte:
Merged.

a88aef32c902b79261c53a16546f4adbf6b3a617 PROV: Fix bignum printout in text serializers

PR #10892, opened 2020-01-19
Fix TLS not using aes_cbc_hmac_sha ciphers

bernd-edlinger:
Mea culpa...

This broke with
87bea6550ae0dda7c40937cff2e86cc2b0b09491

where AES_ASM was no longer defined, since AES_encrypt is no longer done in assembler,
but what this function really needs is AESNI_ASM.


mattcaswell:
I assume the same change is not required on master?

bernd-edlinger:
Correct.

bernd-edlinger:
Merged to 1.1.1 as a3998ea8166e557c2dcf18325c8eb5c45109fa50. Thanks!

PR #10893, opened 2020-01-19
Remove remaining references to crypto/include

bernd-edlinger:
Configure creates an empty crypto/include which
gets not cleaned up with make distclean.

This is for 1.1.1 only at the time, master has more to clean up.


bernd-edlinger:
Merged to 1.1.1 as f50f2725c03710e183d855760c32d8df6cab1f94.  Thanks!

PR #10895, opened 2020-01-19
Fix make distclean

bernd-edlinger:
This makes distclean work correctly in master.
The windows issue is complicated by two effects,
which may be due to the specific windows machine.
First it has one of the commands to delete all generated targets
exceed the command line limit of windows.
And this machine has a cygwin somewhere in the path
almost at the end but its /usr/bin/rmdir nevetheless
overrides the builtin rmdir which makes the rmdir fail,
because it does not recognize "/S" and "/Q".

### DIFF
--- a/Configurations/unix-Makefile.tmpl
+++ b/Configurations/unix-Makefile.tmpl
@@ -524,6 +524,7 @@ clean: libclean
 
 distclean: clean
 	$(RM) configdata.pm
+	$(RM) -r doc/man doc/html
 	$(RM) Makefile
 
 # We check if any depfile is newer than Makefile and decide to
--- a/Configurations/unix-Makefile.tmpl
+++ b/Configurations/unix-Makefile.tmpl
@@ -520,6 +520,7 @@ clean: libclean
 	$(RM) -r test/test-runs
 	$(RM) openssl.pc libcrypto.pc libssl.pc
 	-$(RM) `find . -type l \! -name '.*' -print`
+	$(RM) pod2htmd.tmp
 	$(RM) $(TARFILE)
 
 distclean: clean
--- a/Configurations/windows-makefile.tmpl
+++ b/Configurations/windows-makefile.tmpl
@@ -416,23 +416,32 @@ libclean:
 	-del /Q /F $(LIBS) libcrypto.* libssl.* ossl_static.pdb
 
 clean: libclean
-	-rmdir /Q /S $(HTMLDOCS1_BLDDIRS)
-	-rmdir /Q /S $(HTMLDOCS3_BLDDIRS)
-	-rmdir /Q /S $(HTMLDOCS5_BLDDIRS)
-	-rmdir /Q /S $(HTMLDOCS7_BLDDIRS)
 	{- join("\n\t", map { "-del /Q /F $_" } @PROGRAMS) -}
-	-del /Q /F $(MODULES)
-	-del /Q /F $(SCRIPTS)
-	-del /Q /F $(GENERATED_MANDATORY)
-	-del /Q /F $(GENERATED)
+	-"$(PERL)" -e "while (<>) { chomp; @list = split(' ', $$_); foreach $$file (@list) { unlink $$file; } }" < <<
+$(MODULES)
+$(SCRIPTS)
+$(GENERATED_MANDATORY)
+$(GENERATED)
+pod2htmd.tmp pod2htmi.tmp
+<<
 	-del /Q /S /F *.d *.obj *.pdb *.ilk *.manifest
 	-del /Q /S /F engines\*.lib engines\*.exp
+	-del /Q /S /F providers\*.lib providers\*.exp
 	-del /Q /S /F apps\*.lib apps\*.rc apps\*.res apps\*.exp
-	-del /Q /S /F test\*.exp
-	-rmdir /Q /S test\test-runs
+	-del /Q /S /F test\*.lib test\*.exp .\*.res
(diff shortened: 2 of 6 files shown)


bernd-edlinger:
I would suggest to cherry-pick the second commit "Remove pod2htmd.tmp in make clean"
to 1.1.1.

I have not observed the overlong line in windows make clean, for 1.1.1,
but instead of pod2htmd.tmp there are two of them: pod2htmd.tmp *and* pod2htmi.tmp
and the rmdir does not work if any cygwin is around.

bernd-edlinger:
Richard, you probably remember the issue with the "echo *" on windows, which
wihich war supposed to just print "*" in this case, but that causes a surprise
when cygwin's echo command is also in the path. then "*" is expanded as if it
was a linux echo.  You wrote an echo.pl to fix that issue.
Now the rmdir is the same issue, I see no way how to make nmake use the builtin.
The other issue I had with windows is a command length in del $(GENERATED)
that happened only on master, 1.1.1 did not exceed the line length fortunately.
I saw no other solution than to use perl for that and use a temp file to pass the list.

Configurations/unix-Makefile.tmpl
```diff
@@ -520,10 +520,12 @@ clean: libclean
 	$(RM) -r test/test-runs
 	$(RM) openssl.pc libcrypto.pc libssl.pc
 	-$(RM) `find . -type l \! -name '.*' -print`
+	$(RM) pod2htmd.tmp
```
vdukhovni: By the way, thought this is not an issue in your PR, I am rather uncomfortable with the line above:
```
$(RM) `find . -type l \! -name '.*' -print`
```
It is blatantly unsafe when any names in the output have whitespace.  If we can't rely on a portable 'find ... -delete' option, then we have to do the removal via a Perl script, or something similarly robust to unexpected file names.

bernd-edlinger:
I did test that on a windows 7 machine.
I use remote access to the our internal company network.
I can still do that, but I consider this not necessary.

Configurations/unix-Makefile.tmpl
bernd-edlinger: Damnit this perl method would be far more safe then

bernd-edlinger:
> so I don't feel I can give the final thumbs up.

Maybe you just trust me then?

vdukhovni:
> > so I don't feel I can give the final thumbs up.
> 
> Maybe you just trust me then?

I do, but if Richard is free to take a look, he's much more familiar with the build-system than I.  I am not sure I be doing approvals after only cursory inspection.  Though perhaps with the build-system I may not need to be quite so cautious. Ping me again in a couple of days if nobody else steps forward...

mspncp:
> > so I don't feel I can give the final thumbs up.
> 
> Maybe you just trust me then?

You know it quite well, reviewing is not a matter of trust or distrust. It‘s simply better if someone else checks the changes independently. As the proverb says: „Vertrauen ist gut, Kontrolle ist besser.“

Maybe I find some time to test it...

bernd-edlinger:
Just for reference, when I post a GCC patch,
I write in the e-mail
```
Bootstrapped and Reg-tested on x86_64-pc-linux-gnu
Is it OK for thunk?
```
bootstrap takes hours, and the test suite a day or more.
every reviewer trusts me, that I say that honestly,
the review is about how the code looks like nothing more nothing less.
If they start to not trust a person that is listed in "Write after Approval"
they have to duplicate all my effort, and moreover there is
no CI, it would not be usable since the test suite that runs for days.
Just automated test runs after a patch is assigned.

Configurations/unix-Makefile.tmpl
levitte: Do we still have random symlinks?  I personally would *much* prefer to remove known files than that kind of broad removal, so if it was up to me, I'd remove that line, and then find out what files rename and remove them each explicitly.

levitte: In versions before 1.1.0, we did symlink headers quite a bit.  That line was much more justifiable then than now.

Configurations/windows-makefile.tmpl
```diff
@@ -416,23 +416,32 @@ libclean:
...
-	-del /Q /F $(SCRIPTS)
-	-del /Q /F $(GENERATED_MANDATORY)
-	-del /Q /F $(GENERATED)
+	-"$(PERL)" -e "while (<>) { chomp; @list = split(' ', $$_); foreach $$file (@list) { unlink $$file; } }" < <<
+$(MODULES)
+$(SCRIPTS)
+$(GENERATED_MANDATORY)
+$(GENERATED)
+pod2htmd.tmp pod2htmi.tmp
+<<
```
levitte: I obviously prefer the way it's done in #11171 :wink:

bernd-edlinger: > I obviously prefer the way it's done in #11171 wink

Feel free to merge some ideas from here with your PR, or the other way round,
you have write-access to this PR as well

bernd-edlinger: PS: I am not sure if meanwhile more files leak than before, this PR
is also about making distclean clean the build tree when it is in-tree

levitte: I only commented on these lines, not the whole PR

bernd-edlinger:
Removing windows stuff for now.
I don't know if the nmake distclean was fixed by @levitte
but I dont care about windows any more.
In the moment I see the following files are leaking again,
```
$ git clean -ndx
Would remove crypto/include/
Would remove providers/fipsinstall.cnf
Would remove test-runs/
```
So since I started this PR new bugs were introduced,
nobody reviewed this, for a long time, if I spend time
to fix this PR again, it will not be reviewed again,
probably I better give up right now before I waste more time.

PR #10897, opened 2020-01-20
Add support for DH 'modp' group parameters (RFC 3526)

slontis:
##### Checklist


richsalz:
What's the rationale/need for this, as opposed to ECDH with NISTP256?

slontis:
Just conforming to the standards that we are fips validating (which includes DH, not just ECDH)
See NIST SP 800-56A Rev. 3 Appendix D.
Basically for DH if you dont use named groups you are in big trouble because the value of 'q' is unknown. (i.e- q is not always saved into asn1 as well as not being transferred during TLS).
The value of q is required for key validation purposes when using DH.

richsalz:
I had forgotten we were doing it; it is listed in Appendix of the FIPS arch doc.

crypto/bn/bn_dh.c
```diff
@@ -11,474 +11,593 @@
...
-    0xA308B0FE64F5FBD3ULL, 0x99B1A47D1EB3750BULL, 0xB447997640129DA2ULL,
-    0x8CF83642A709A097ULL
+/*-
+ * "2048-bit MODP Group" from RFC3526, Section 3.
+ *
+ * The prime is: 2^2048 - 2^1984 - 1 + 2^64 * { [2^1918 pi] + 124476 }
+ *
+ * RFC3526 specifies a generator of 2.
+ */
+static const BN_ULONG modp_2048_p[] = {
```
mattcaswell: Don't these values already exist in bn_const.c?

slontis: Sigh... Will change bn_const so it uses the 'actual' consts in bn_dh.c.
There should not have been a need for a dynamic version of the BIGNUM.. 
(But will be for backwards compatability)

slontis:
Updated

crypto/dh/dh_group_params.c
```diff
@@ -16,18 +18,34 @@
...
+    if (q == NULL || !BN_rshift1(q, q)) {
+        BN_free(q);
+        DH_free(dh);
+        return NULL;
+    }
     dh->p = (BIGNUM *)p;
+    dh->q = (BIGNUM *)q;
     dh->g = (BIGNUM *)&_bignum_const_2;
-    dh->length = nbits;
+    dh->length = nbits; /* Private key length = 2 * s */
```
mattcaswell: What does "s" refer to here?

slontis: It is listed in the description in DH_new_by_nid()

mattcaswell: Ah - right. No where does it refer to this as "s" though. Perhaps change this to "2 * max_target_security_strength" to be consistent with the description in DH_new_by_nid()?

slontis: 's' would have come out of the docs that I was referring to - will change the name..

slontis:
nit fixups and rebased

test/evp_extra_test.c
```diff
@@ -1465,9 +1465,8 @@ static int test_EVP_PKEY_set1_DH(void)
             || !TEST_int_eq(EVP_PKEY_id(pkey1), EVP_PKEY_DHX))
         goto err;
 
-
     if(!TEST_true(EVP_PKEY_set1_DH(pkey2, pkcs3dh))
-            || !TEST_int_eq(EVP_PKEY_id(pkey2), EVP_PKEY_DH))
+            || !TEST_int_eq(EVP_PKEY_id(pkey2), EVP_PKEY_DHX))
```
mattcaswell: Errr,....say what??? I missed this before. Why is this changing? Surely pkcs3dh should have type EVP_PKEY_DH??

slontis: DHX is the newer format that supports saving of q. If we load a named group then it really does have a known 'q' value. The code determines if it uses dhx by checking if it has a q value.
If you really want the old behaviour then a hacky dont use dhx flag would need to be added when we load by nid (or store the nid). Which way do you want?
Is this what we really want?

paulidale:
Rebase required.

slontis:
rebased

test/evp_extra_test.c
mattcaswell: The whole point of this test is check that if you assign a X9.42 DH object to an EVP_PKEY then you get an EVP_PKEY_DHX, whereas if you assign a PKCS.3 DH object (no q) then you get an EVP_PKEY_DH. After this PR it no longer tests that.

mattcaswell: I'm also a little concerned about the various explicit uses of EVP_PKEY_DH in libssl (search for `EVP_PKEY_DH`)

mattcaswell: I think until this PR none of the DH "named groups" that we used within a libssl context ever returned EVP_PKEY_DHX. On the up-side I think all of the DH "named group" code in libssl is only in master (not 1.1.1).

slontis: > whereas if you assign a PKCS.3 DH object (no q) then you get an EVP_PKEY_DH. After this PR it no longer tests that.

Part of the issue is that TLS doesnt send 'q' so that is probably why it just uses the EVP_PKEY_DH type.
It is debatable whether this named group doesnt have a 'q' - because it does have a known q - we just werent setting it before.. So it looks like you want it to store some extra state and actually still use the old type for these groups (I will still store the q) but it wont save it as dhx..

I will think about this some more - maybe q is just a value required in some operations (such as printing and validation and we dont need to store it at all.. just calculate when required (or store into another variable or flag it)..

slontis:
@mattcaswell - I have cached the nid for the named group, and then check this when deciding to choose DH or DHX.

test/evp_extra_test.c
mattcaswell: IMO, if we have a q then we should use it.

I think there are two issues here:

1) The intention of the test was to check what happens when we assign DH objects without a q to an EVP_PKEY. It should end up with the EVP_PKEY_DH type - it no longer tests that

2) I have no problem with these named groups being EVP_PKEY_DHX type. We just need to review the libssl code to make sure it is still correct - because that code was written with the expectation that it would be EVP_PKEY_DH. I *think* all of the instances of this in libssl are related to TLSv1.3. The whole thing about q not being sent in TLS is irrelevant, because that only applies to TLSv1.2 and below. From TLSv1.3 we use named DH groups.

crypto/evp/p_lib.c
```diff
@@ -539,8 +540,16 @@ EC_KEY *EVP_PKEY_get1_EC_KEY(EVP_PKEY *pkey)
...
-    int ret = EVP_PKEY_assign(pkey, type, key);
+    int type = EVP_PKEY_DH;
+    int ret;
+
+    /*
+     * Use DHX for any DH key that has a q parameter,
+     * except for named groups (for backwards compatibility reasons).
+     */
+    if (DH_get0_q(key) != NULL && dh_get0_nid(key) == NID_undef)
+        type = EVP_PKEY_DHX;
```
mattcaswell: I'm not sure this is the right answer - and wasn't the point of my comments. It seems to me right that the type should be DHX if a "q" is present. The code in libssl I'm concerned about is only in master so we need not be worried about backwards compat there. The test I was worried about, was only an issue because it changed the intent of the test, i.e. the point was that if you create an EVP_PKEY from a DH without a q then you you should get EVP_PKEY_DH. Fudging things to force that to happen in libcrypto seems like the wrong answer. Instead the test should be changed to use a different DH group that doesn't have a q (perhaps a non-named group if needs be).

test/evp_extra_test.c
kaduk: > I *think* all of the instances of [named groups with EVP_PKEY_DH type] in libssl are related to TLSv1.3

I think if you support those curves you have to support them in 1.2 as well as 1.3, so it's not necessarily limited to TLS 1.3

crypto/evp/p_lib.c
slontis: I have changed the test input so it has no q and this produces a dhx_type. Got rid of the extra check for a named group when determining the type, so all named groups use the DHX type.

slontis: @mattcaswell can you review please/

test/evp_extra_test.c
mattcaswell: > I think if you support those curves you have to support them in 1.2 as well as 1.3, so it's not necessarily limited to TLS 1.3

Err...really? These groups are for DHE key exchange only. AFAIK TLSv1.2 lacks the ability to select a group by name in the ServerKeyExchange message for DHE - you can only provide the parameters explicitly.

kaduk: RFC 7919 specifies negotiated FFDHE named groups for TLS 1.0, 1.1, and 1.2

kaduk: Ah, but I see your point that the server still sends explicit parameters in ServerKeyExchange in that case, and the client's on the hook to validate them against the named group(s)

mattcaswell: We do not support RFC7919 (actually I wasn't aware of it until you pointed me at it). Looking at that RFC it still uses explicit parameters, i.e. it introduces named ffdhe groups in the "supported groups" extension,, but then the parameters themselves are still passed explicitly in the ServerKeyExchange. AFAIK there is nothing in the TLSv1.3 RFC that  requires us to start supporting RFC7919 in TLSv1.2. If we see the ffdhe groups in a TLSv1.2 connection we just ignore them.

kaduk: > AFAIK there is nothing in the TLSv1.3 RFC that requires us to start supporting RFC7919 in TLSv1.2.

Indeed, it is not listed in the "Updates Affecting TLS 1.2" section of RFC 8446

> If we see the ffdhe groups in a TLSv1.2 connection we just ignore them.

Do we want to make an issue to track changing that (to not ignore them)?

mattcaswell: #10971

slontis:
Thanks. Merged to master.

PR #10901, opened 2020-01-20
Russian Elbrus processors support

beldmit:
##### Checklist

### DIFF
--- a/config
+++ b/config
@@ -551,6 +551,7 @@ case "$GUESSOS" in
   pentium-*-vxworks*) OUT="vxworks-pentium" ;;
   simlinux-*-vxworks*) OUT="vxworks-simlinux" ;;
   mips-*-vxworks*) OUT="vxworks-mips";;
+  e2k-*-linux*) OUT="linux-generic64 -DL_ENDIAN" ;;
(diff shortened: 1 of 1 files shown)


config
```diff
@@ -551,6 +551,7 @@ case "$GUESSOS" in
   pentium-*-vxworks*) OUT="vxworks-pentium" ;;
   simlinux-*-vxworks*) OUT="vxworks-simlinux" ;;
   mips-*-vxworks*) OUT="vxworks-mips";;
+  e2k*) OUT="linux-generic64 -DL_ENDIAN" ;;
```
levitte: Will only Linux be implemented for those processors, ever?  If you can predict the future that well, this is fine, but if not, the case should be `e2k-*-linux*`

beldmit: I hardly can imagine that any not Linux-based OS will appear in Russia more or less soon.

beldmit: Well, I was wrong. @levitte, many thanks!

beldmit:
@paulidale, could you please reapprove?

beldmit:
Applied. Thanks!

PR #10860, opened 2020-01-16
Modify DSA and DH keys to use a shared FFC_PARAMS struct

slontis:
This is required in order to share code for FIPS related parameter generation and validation routinues.
Note the 'counter' field is now stored as a integer (as that is the form required for generation/validation functions).

##### Checklist

- [ ] documentation is added or updated
- [ ] tests are added or updated

PR #10909, opened 2020-01-21
Add FFC param/key generation

slontis:
##### Checklist

### DIFF
--- a/crypto/dh/dh_gen.c
+++ b/crypto/dh/dh_gen.c
@@ -15,19 +15,84 @@
 #include <stdio.h>
 #include "internal/cryptlib.h"
 #include <openssl/bn.h>
+#include <openssl/dh.h>
 #include "dh_local.h"
 
+#ifndef FIPS_MODE
 static int dh_builtin_genparams(DH *ret, int prime_len, int generator,
                                 BN_GENCB *cb);
+#endif /* FIPS_MODE */
+
+int DH_generate_ffc_parameters(OPENSSL_CTX *libctx, DH *dh, int bits,
+                               int qbits, int gindex, BN_GENCB *cb)
+{
+    int ret, res;
+
+    if (qbits <= 0) {
+        const EVP_MD *evpmd = bits >= 2048 ? EVP_sha256() : EVP_sha1();
+
+        qbits = EVP_MD_size(evpmd) * 8;
+    }
+    dh->params.gindex = gindex;
+    ret = ffc_params_FIPS186_4_generate(libctx, &dh->params, FFC_PARAM_TYPE_DH,
+                                        bits, qbits, NULL, &res, cb);
+    if (ret > 0)
+        dh->dirty_cnt++;
+    return ret;
+}
 
 int DH_generate_parameters_ex(DH *ret, int prime_len, int generator,
                               BN_GENCB *cb)
 {
+#ifdef FIPS_MODE
+    /*
+     * Just chose an approved safe prime group.
+     * The alternative to this is to generate FIPS186-4 domain parameters i.e.
+     * return DH_generate_ffc_parameters(ret, prime_len, -1, -1, cb);
+     * As the FIPS186-4 generated params are for backwards compatability,
+     * the safe prime group should be used as the default.
+     */
+    DH *dh = NULL;
+    int ok = 0, nid;
+
+    if (generator != 2)
+        return 0;
+
+    switch (prime_len) {
+    case 2048:
+        nid = NID_ffdhe2048;
+        break;
+    case 3072:
+        nid = NID_ffdhe3072;
+        break;
+    case 4096:
+        nid = NID_ffdhe4096;
+        break;
+    case 6144:
+        nid = NID_ffdhe6144;
+        break;
+    case 8192:
+        nid = NID_ffdhe8192;
+        break;
+    /* unsupported prime_len */
+    default:
+        return 0;
+    }
+    dh = DH_new_by_nid(nid);
+    if (dh != NULL && ffc_params_copy(&ret->params, &dh->params)) {
+        ok = 1;
+        ret->dirty_cnt++;
+    }
+    DH_free(dh);
+    return ok;
+#else
     if (ret->meth->generate_params)
         return ret->meth->generate_params(ret, prime_len, generator, cb);
     return dh_builtin_genparams(ret, prime_len, generator, cb);
+#endif /* FIPS_MODE */
 }
 
+#ifndef FIPS_MODE
 /*-
  * We generate DH parameters as follows
  * find a prime p which is prime_len bits long,
@@ -133,3 +198,4 @@ static int dh_builtin_genparams(DH *ret, int prime_len, int generator,
     BN_CTX_free(ctx);
     return ok;
 }
+#endif /* FIPS_MODE */
--- a/crypto/dh/dh_gen.c
+++ b/crypto/dh/dh_gen.c
@@ -15,7 +15,7 @@
 #include <stdio.h>
 #include "internal/cryptlib.h"
 #include <openssl/bn.h>
-#include <openssl/dh.h>
+#include "crypto/dh.h"
 #include "dh_local.h"
 
 #ifndef FIPS_MODE
@@ -23,7 +23,11 @@ static int dh_builtin_genparams(DH *ret, int prime_len, int generator,
                                 BN_GENCB *cb);
 #endif /* FIPS_MODE */
 
-int DH_generate_ffc_parameters(OPENSSL_CTX *libctx, DH *dh, int bits,
+/*
+ * TODO(3.0): keygen should be able to use this method to do a FIPS186-4 style
+ * paramgen.
+ */
+int dh_generate_ffc_parameters(OPENSSL_CTX *libctx, DH *dh, int bits,
                                int qbits, int gindex, BN_GENCB *cb)
 {
     int ret, res;
@@ -46,9 +50,9 @@ int DH_generate_parameters_ex(DH *ret, int prime_len, int generator,
 {
 #ifdef FIPS_MODE
     /*
-     * Just chose an approved safe prime group.
+     * Just choose an approved safe prime group.
      * The alternative to this is to generate FIPS186-4 domain parameters i.e.
-     * return DH_generate_ffc_parameters(ret, prime_len, -1, -1, cb);
+     * return dh_generate_ffc_parameters(ret, prime_len, -1, -1, cb);
      * As the FIPS186-4 generated params are for backwards compatability,
      * the safe prime group should be used as the default.
      */
--- a/crypto/dh/dh_key.c
+++ b/crypto/dh/dh_key.c
@@ -245,8 +245,8 @@ static int dh_generate_key(OPENSSL_CTX *libctx, DH *dh)
             if (dh->params.q == NULL) {
                 /* secret exponent length */
                 l = dh->length ? dh->length : BN_num_bits(dh->params.p) - 1;
-                if (!BN_priv_rand(priv_key, l, BN_RAND_TOP_ONE,
-                                  BN_RAND_BOTTOM_ANY))
+                if (!BN_priv_rand_ex(priv_key, l, BN_RAND_TOP_ONE,
+                                     BN_RAND_BOTTOM_ANY, ctx))
                     goto err;
                 /*
                  * We handle just one known case where g is a quadratic non-residue:
--- a/crypto/dh/dh_key.c
+++ b/crypto/dh/dh_key.c
@@ -1,5 +1,5 @@
 /*
- * Copyright 1995-2019 The OpenSSL Project Authors. All Rights Reserved.
+ * Copyright 1995-2020 The OpenSSL Project Authors. All Rights Reserved.
  *
  * Licensed under the Apache License 2.0 (the "License").  You may not use
  * this file except in compliance with the License.  You can obtain a copy
@@ -13,10 +13,7 @@
 #include "crypto/bn.h"
 #include "crypto/dh.h"
 
-#ifndef FIPS_MODE
 static int generate_key(DH *dh);
-#endif /* FIPS_MODE */
-
 static int dh_bn_mod_exp(const DH *dh, BIGNUM *r,
                          const BIGNUM *a, const BIGNUM *p,
                          const BIGNUM *m, BN_CTX *ctx, BN_MONT_CTX *m_ctx);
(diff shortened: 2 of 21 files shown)


slontis:
This PR is dependent on #10860 & #10897 
(which have been cherry-picked temporarily into this PR as 3 commits after the first one)

slontis:
ping

crypto/dh/dh_gen.c
```diff
@@ -15,19 +15,84 @@
...
 #include <openssl/bn.h>
+#include <openssl/dh.h>
 #include "dh_local.h"
 
+#ifndef FIPS_MODE
 static int dh_builtin_genparams(DH *ret, int prime_len, int generator,
                                 BN_GENCB *cb);
+#endif /* FIPS_MODE */
+
+int DH_generate_ffc_parameters(OPENSSL_CTX *libctx, DH *dh, int bits,
```
mattcaswell: Why is this public? We shouldn't be adding new public functions like this, whilst at the same time deprecating the rest of the low level functions.

crypto/dh/dh_gen.c
```diff
@@ -15,19 +15,84 @@
...
+        dh->dirty_cnt++;
+    return ret;
+}
 
 int DH_generate_parameters_ex(DH *ret, int prime_len, int generator,
                               BN_GENCB *cb)
 {
+#ifdef FIPS_MODE
+    /*
+     * Just chose an approved safe prime group.
```
mattcaswell: typo: choose

crypto/dh/dh_key.c
```diff
@@ -227,25 +223,52 @@ static int generate_key(DH *dh)
...
+                goto err;
+        } else {
+#ifdef FIPS_MODE
+            if (dh->params.q == NULL)
+                goto err;
+#else
+            if (dh->params.q == NULL) {
+                /* secret exponent length */
+                l = dh->length ? dh->length : BN_num_bits(dh->params.p) - 1;
+                if (!BN_priv_rand(priv_key, l, BN_RAND_TOP_ONE,
```
mattcaswell: BN_priv_rand_ex() in order to use the right libctx?

crypto/dh/dh_pmeth.c
```diff
@@ -310,22 +300,26 @@ static DSA *dsa_dh_generate(DH_PKEY_CTX *dctx, BN_GENCB *pcb)
...
-                                  NULL, 0, NULL, NULL, NULL, pcb);
-    else if (dctx->use_dsa == 2)
-        rv = dsa_builtin_paramgen2(ret, prime_len, subprime_len, md,
-                                   NULL, 0, -1, NULL, NULL, NULL, pcb);
+        rv = ffc_params_FIPS186_2_generate(libctx, &ret->params, FFC_PARAM_TYPE_DH,
+                                           prime_len, subprime_len, md, &res,
+                                           pcb);
+    else
+# endif
+    if (dctx->use_dsa)
```
mattcaswell: `dctx->use_dsa >= 1`

crypto/dh/dh_pmeth.c
```diff
@@ -372,22 +366,17 @@ static int pkey_dh_paramgen(EVP_PKEY_CTX *ctx, EVP_PKEY *pkey)
             return 0;
         evp_pkey_set_cb_translate(pcb, ctx);
     }
-#ifndef OPENSSL_NO_DSA
+# ifdef FIPS_MODE
+    dctx->use_dsa = 2;
```
mattcaswell: The use_dsa value of "2" does not seem to be documented:

> The EVP_PKEY_CTX_set_dh_paramgen_type() macro sets the key type for DH
> parameter generation. Use 0 for PKCS#3 DH and 1 for X9.42 DH.
> The default is 0.

Maybe it might be nice to use macros rather than "magic" numbers?

Side note: `use_dsa` seems like a terrible name

crypto/ffc/ffc_params_generate.c
```diff
@@ -0,0 +1,995 @@
...
+         * A.2.4 Step (9) & (10) & (11)
+         * W = Hash(seed || "ggen" || index || counter)
+         * g = W^e % p
+         */
+        static const unsigned char ggen[4] = { 0x67, 0x67, 0x65, 0x6e };
+
+        md[0] = (unsigned char)(gindex & 0xff);
+        md[1] = (unsigned char)((counter >> 8) & 0xff);
+        md[2] = (unsigned char)(counter & 0xff);
+        if (!(EVP_DigestInit_ex(mctx, evpmd, NULL)
```
mattcaswell: This is contrary to way we do this in most other places, i.e. `!... || !... || !...`

crypto/ffc/ffc_params_generate.c
```diff
@@ -0,0 +1,995 @@
...
+             */
+            if (!(EVP_Digest(buf, buf_len, md, NULL, evpmd, NULL)
+                    && (BN_bin2bn(md, mdsize, tmp) != NULL)
+                    /*
+                     * A.1.1.2 Step (11.2)
+                     * A.1.1.3 Step (13.2)
+                     * W += V(j) * 2^(outlen * j)
+                     */
+                    && BN_lshift(tmp, tmp, (mdsize << 3) * j)
+                    && BN_add(W, W, tmp)))
```
mattcaswell: Same here...and other similar occurrences below

crypto/ffc/ffc_params_generate.c
```diff
@@ -0,0 +1,995 @@
...
+        /* Step (11.9) : offset = offset + n + 1 is done auto-magically */
+    }
+    /* No prime P found */
+    ret = 0;
+    *res |= FFC_CHECK_P_NOT_PRIME;
+    goto err;
+found:
+    *counter = i;
+    ret = 1;
+    goto err;
```
mattcaswell: This `goto err` seems superfluous.

crypto/ffc/ffc_params_generate.c
```diff
@@ -0,0 +1,995 @@
...
+    return ret;
+}
+
+
+/*
+ * FIPS 186-4 FFC parameter generation (as defined in Appendix A).
+ * The same code is used for validation (when validate_flags != 0)
+ *
+ * The primes p & q are generated/validated using:
+ *   A.1.1.2 Generation or probable primes p & q using approved hash.
```
mattcaswell: Typo: of?

crypto/ffc/ffc_params_generate.c
```diff
@@ -0,0 +1,995 @@
...
+/* This should no longer be used in FIPS mode */
+int ffc_params_FIPS186_2_generate(OPENSSL_CTX *libctx, FFC_PARAMS *params,
+                                  int type, size_t L, size_t N,
+                                  const EVP_MD *evpmd, int *res, BN_GENCB *cb)
+{
+    return ffc_param_FIPS186_2_gen_verify(libctx, params, type, L, N, evpmd,
+                                          0, res, cb);
+}
+
+/* TODO - Add this in another PR -  just add a stub for now */
```
mattcaswell: TODO(3.0)?

util/libcrypto.num
```diff
@@ -4917,3 +4917,6 @@ PKCS8_pkey_add1_attr                    ?	3_0_0	EXIST::FUNCTION:
 PKCS8_pkey_add1_attr_by_OBJ             ?	3_0_0	EXIST::FUNCTION:
 EVP_PKEY_private_check                  ?	3_0_0	EXIST::FUNCTION:
 EVP_PKEY_pairwise_check                 ?	3_0_0	EXIST::FUNCTION:
+DH_generate_ffc_parameters              ?	3_0_0	EXIST::FUNCTION:DH
+DSA_generate_ffc_parameters             ?	3_0_0	EXIST::FUNCTION:DSA
+DSA_generate_parameters_ctx             ?	3_0_0	EXIST::FUNCTION:DSA
```
mattcaswell: It seems like a bad idea to make these public.

levitte:
I battled with the ffc functions, and found the lack of (internal) documentation... irritating.

Please add some. I don't care if it happens in this or another PR.

crypto/dh/dh_gen.c
slontis: Made it private - once keygen is done we need to be able to select this method (so added a TODO(3.0) to say this.
We need to figure out what we are going to do with the dsaparam and dhparam apps that currently only call low level DH and DSA functions.

crypto/dh/dh_pmeth.c
slontis: The name is mapping to what it does in the app...
i.e- dsa has extra fields in ANS1 that dh doesnt have - so it makes sense to me.

slontis: Changed name to paramgen_type and added defines..

crypto/ffc/ffc_params_generate.c
slontis: Changed (Dont agree but changed :))

crypto/ffc/ffc_params_generate.c
slontis: moved the code up.

slontis:
> I battled with the ffc functions, and found the lack of (internal) documentation... irritating

Not sure what you think is going to help you understand these functions..
This is straight out of the FIPS186-4 document. Hence the links to steps that are everywhere in the code.
The most confusing bit is that the generator functions are also used for verification.. (which is in another PR).

util/libcrypto.num
slontis: Removed...

crypto/ffc/ffc_key_generate.c
```diff
@@ -0,0 +1,61 @@
...
+            break;
+    } while (1);
+
+    ret = 1;
+err:
+    BN_free(two_powN);
+    return ret;
+#else
+    do {
+        if (!BN_priv_rand_range(priv, params->q))
```
mattcaswell: Shouldn't we use `BN_priv_rand_range_ex` even here?

crypto/dsa/dsa_gen.c
```diff
@@ -1,618 +1,88 @@
...
-                                    seed_in, seed_len, NULL, counter_ret,
-                                    h_ret, cb);
+    if (qbits <= 0) {
+        const EVP_MD *evpmd = pbits >= 2048 ? EVP_sha256() : EVP_sha1();
+
+        qbits = EVP_MD_size(evpmd) * 8;
     }
+    dsa->params.gindex = gindex;
+#ifndef FIPS_MODE
+    if (type == 1)
```
mattcaswell: Did we have macros for these types?

crypto/dsa/dsa_key.c
```diff
@@ -11,42 +11,53 @@
...
         return dsa->meth->dsa_keygen(dsa);
-    return dsa_builtin_keygen(dsa);
+    return dsa_builtin_keygen(NULL, dsa);
 }
 
-static int dsa_builtin_keygen(DSA *dsa)
+int dsa_generate_key_ctx(OPENSSL_CTX *libctx, DSA *dsa)
+{
+#ifndef FIPS_MODE
+    if (dsa->meth->dsa_keygen)
```
mattcaswell: `dsa_keygen != NULL`?

crypto/ffc/ffc_params_generate.c
```diff
@@ -0,0 +1,992 @@
...
+        && qsize != SHA224_DIGEST_LENGTH
+        && qsize != SHA256_DIGEST_LENGTH) {
+        /* invalid q size */
+        *res = FFC_CHECK_INVALID_Q_VALUE;
+        return FFC_PARAMS_RET_STATUS_FAILED;
+    }
+
+    if (evpmd == NULL) {
+        if (qsize == SHA_DIGEST_LENGTH)
+            evpmd = EVP_sha1();
```
mattcaswell: Does this even compile in FIPS_MODE? I don't think using implicitly fetched digests like this will work in the FIPS provider. (Side note: Perhaps we should prevent their use by wrapping the declarations in `#ifndef FIPS_MODE`)

crypto/dsa/dsa_gen.c
slontis: for dh yes .. added some for DH into crypto/dsa.h

crypto/dsa/dsa_key.c
slontis: changed in another place also

crypto/ffc/ffc_params_generate.c
slontis: It does compile.. I would prefer to leave it in so will change it to do a fetch. (We shouldnt break existing code just because it is FIPS if we can help it).

slontis:
@mattcaswell all changes have been addressed...

crypto/ffc/ffc_params_generate.c
mattcaswell: It's actually quite concerning that this built without an error. I just checked, and legacy_sha.c which provides the implementation for EVP_sha1() does not get linked into fips.so....so how is the link completing successfully???

mattcaswell: I investigated this further, and this turns out to be a real problem in our linking. You *should* have got a linker error from this. See #11020.

slontis: That may imply that there might be a few more issues :)

slontis:
merged to master.

junaruga:
Hi,
I faced the following error in the `generate_key` in the `crypto/dh/dh_key.c` in FIPS module on the pull-request https://github.com/ruby/openssl/pull/937.

https://github.com/openssl/openssl/pull/10909/files#diff-78ed4660fbeb3ac4a81589478f963d94a1452d534c6d7189d562512852a1efcfR242-R243

`crypto/dh/dh_key.c` - `generate_key`
```
#ifdef FIPS_MODE
            if (dh->params.q == NULL)
                goto err;
#else
```

> Not sure what you think is going to help you understand these functions.. This is straight out of the FIPS186-4 document. Hence the links to steps that are everywhere in the code. The most confusing bit is that the generator functions are also used for verification.. (which is in another PR).

Could you tell me where the above FIPS specific check is written in the FIPS 186-4 document? Thanks!

https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf

mattcaswell:
That particular failure is related to DH key generation (i.e. FFC key generation). Appendix B of 186-4 provides two methods for generating keys both of which require the params p, q and g as input:

> The subset of the domain parameters that are used for this process. p, q
> and g shall either be provided as integers during input, or shall be
> converted to integers prior to use.

If you are missing one of the required inputs (q) then you cannot generate the key.

slontis:
As this question relates to TLS, DH with randomly generated FFC style parameters will just not work with FIPS.
TLS does not send the q, so when the peer recieves the params there is no q and FIPS will error.
So for this case only named primes groups will work as params.. (e.g. FFDHE) i.e. When the peer recieves  the params it can tell that the constants are a named safe prime group, so it knows what the q is.

junaruga:
Guys, all right, thank you for your explanation!

PR #10916, opened 2020-01-21
Legacy digests can have custom control values

### LINKED ISSUE #10915: Minor EVP_MD_CTX_ctrl regression in master
In pre-3.0.0 times it was possible to use custom ctrl command values at least for digests. Now EVP_MD_CTX_ctrl checks against known values (https://github.com/openssl/openssl/blob/7b131de2bb2be8582c6dfc50bdd3cbc49e1a08be/crypto/evp/digest.c#L671-L687) and then returns an error.

It's not critical regression but it's still a regression. If it's the desired behavior, I can leave with it.

beldmit:
Fixes #10915.

##### Checklist

### DIFF
--- a/crypto/evp/digest.c
+++ b/crypto/evp/digest.c
@@ -663,7 +663,7 @@ int EVP_MD_CTX_ctrl(EVP_MD_CTX *ctx, int cmd, int p1, void *p2)
     }
 
     if (ctx->digest->prov == NULL
-        && (ctx->pctx == NULL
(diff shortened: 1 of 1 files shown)


beldmit:
Many thanks to @mattcaswell for a quick fix!

beldmit:
Many thanks!

mattcaswell:
Looks like this could be the cause of at least some of the recent travis failures. Reverting commit in #10989

PR #10924, opened 2020-01-22
"DTLS over SCTP"  support for sending ApplicationData on multiple sct…

NaveenShivanna86:
`DTLS over SCTP` support for sending ApplicationData on multiple sctp streams.

Description:
As per RFC 6083, section 4.4, all DTLS messages of the ChangeCipherSpec, Alert, or Handshake
MUST be transported on stream 0 and for ApplicationData, there is a provision to use multiple
streams. When custom SCTP implementation (not lksctp) is used then custom BIO METHODS are
registered to OpenSSL. Before calling the write method, we can save the record_type and based on whether it is DTLS control messages or ApplicationData, stream_id can be selected to send out the message finally.

RFC Snippet:
4.4.  Stream Usage

   All DTLS messages of the ChangeCipherSpec, Alert, or Handshake
   protocol MUST be transported on stream 0 with unlimited reliability
   and with the ordered delivery feature.

   DTLS messages of the ApplicationData protocol SHOULD use `multiple
   streams` other than stream 0; they MAY use stream 0 for everything if
   they do not care about minimizing head of line blocking.

### DIFF
--- a/crypto/bio/bss_dgram.c
+++ b/crypto/bio/bss_dgram.c
@@ -133,6 +133,7 @@ typedef struct bio_dgram_sctp_data_st {
     int ccs_sent;
     int save_shutdown;
     int peer_auth_tested;
+    int record_type;
 } bio_dgram_sctp_data;
 # endif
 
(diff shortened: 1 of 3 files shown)


mattcaswell:
Is this actually needed? Can't you just infer this by looking at the record type in the record header (i.e. "peeking" at the first byte of the record).

NaveenShivanna86:
> Is this actually needed? Can't you just infer this by looking at the record type in the record header (i.e. "peeking" at the first byte of the record).
--
OK. This is good solution. 
In my case, the BIO METHODS are with different adaptation layer module, so cannot suggest other module to peek the message header.
But definitely now I feel OpenSSL code need not be changed.
I will close this PR. Thanks.

PR #10928, opened 2020-01-22
Add internal maxsize macros

levitte:
We've started to see "magic" numbers being used for certain sizes,
such as algorithm names and property query strings.

This change takes care of the few items where buffers for algorithm
names and property query strings are used.

### DIFF
--- a/providers/implementations/signature/dsa.c
+++ b/providers/implementations/signature/dsa.c
@@ -50,7 +50,6 @@ typedef struct {
     OPENSSL_CTX *libctx;
     DSA *dsa;
     size_t mdsize;
-    /* Should be big enough */
     char mdname[OSSL_MAX_NAME_SIZE];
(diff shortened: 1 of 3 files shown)


providers/implementations/signature/dsa.c
```diff
@@ -50,7 +51,7 @@ typedef struct {
     DSA *dsa;
     size_t mdsize;
     /* Should be big enough */
```
mattcaswell: Probably this comment (and similarly above) is no longer required.

levitte: Fixed

mattcaswell (approved):

t8m (approved):

levitte:
Merged.

c24937d5e71a990763f227be229ad41aab44728f Add internal maxsize macros

PR #10930, opened 2020-01-22
Also check for errors in x86_64-xlate.pl.

davidben:
In https://github.com/openssl/openssl/pull/10883, I'd meant to exclude
the perlasm drivers since they aren't opening pipes and do not
particularly need it, but I only noticed x86_64-xlate.pl, so
arm-xlate.pl and ppc-xlate.pl got the change.

That seems to have been fine, so be consistent and also apply the change
to x86_64-xlate.pl. Checking for errors is generally a good idea.

### DIFF
--- a/crypto/aes/asm/aes-586.pl
+++ b/crypto/aes/asm/aes-586.pl
@@ -2995,4 +2995,4 @@ sub deckey()
 
 &asm_finish();
 
-close STDOUT or die "error closing STDOUT";
+close STDOUT or die "error closing STDOUT: $!";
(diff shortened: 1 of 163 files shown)


levitte:
The string should be "error closing STDOUT: $!"

levitte:
Travis failure is unrelated

davidben:
> The string should be "error closing STDOUT: $!"

Done. Applied this to all the files since the others already merged stuff in.

paulidale:
Merged to master.  Thanks.

PR #10931, opened 2020-01-22
Do not silently truncate files on perlasm errors

davidben:
If one of the perlasm xlate drivers crashes, OpenSSL's build will
currently swallow the error and silently truncate the output to however
far the driver got. This will hopefully fail to build, but better to
check such things.

Handle this by checking for errors when closing STDOUT (which is a pipe
to the xlate driver).

This is the OpenSSL 1.1.1 version of
https://github.com/openssl/openssl/pull/10883 and
https://github.com/openssl/openssl/pull/10930.

### DIFF
--- a/crypto/aes/asm/aes-armv4.pl
+++ b/crypto/aes/asm/aes-armv4.pl
@@ -1242,4 +1242,4 @@ while(<SELF>) {
 close SELF;
 
 print $code;
-close STDOUT;	# enforce flush
+close STDOUT or die "error closing STDOUT: $!";	# enforce flush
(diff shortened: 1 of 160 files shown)


levitte:
The string should be "error closing STDOUT: $!"

davidben:
> The string should be "error closing STDOUT: $!"

Done.

mspncp:
> > The string should be "error closing STDOUT: $!"
> 
> Done.

Would you mind adding it for master, too?

davidben:
> Would you mind adding it for master, too?

Already did in https://github.com/openssl/openssl/pull/10930.

paulidale:
Merged to master, thanks.

PR #10934, opened 2020-01-23
Fix coverity issues CID 1457745...1457752, 1457853, 1457854

slontis:
CID 1457854 - keymgmt_lib.c : OVERRUN
CID 1457853 - self_test_kats.c : UNINT
CID 1457752 - fipsprov.c RESOURCE_LEAK
CID 1457751 - apps/pkcs12.c CHECKED_RETURN
CID 1457750 - dsa_ossl.c RESOURCE_LEAK (marked as false positive since tmp != NULL)
CID 1457749 - apps/nseq.c : CHECKED_RETURN
CID 1457748 - cipher_aes_cbc_hmac_sha.c : SIZEOF_MISMATCH
CID 1457747 - cipher_aes_cbc_hmac_sha.c : SIZEOF_MISMATCH
CID 1457746 - same as 1457752
CID 1457745 - apps/ocsp  : CHECKED_RETURN

##### Checklist

### DIFF
--- a/providers/fips/self_test_kats.c
+++ b/providers/fips/self_test_kats.c
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 The OpenSSL Project Authors. All Rights Reserved.
+ * Copyright 2019-2020 The OpenSSL Project Authors. All Rights Reserved.
  *
  * Licensed under the OpenSSL license (the "License").  You may not use
  * this file except in compliance with the License.  You can obtain a copy
@@ -10,6 +10,7 @@
 #include <string.h>
 #include <openssl/evp.h>
 #include <openssl/kdf.h>
+#include "internal/cryptlib.h"
 #include "internal/nelem.h"
 #include "self_test.h"
 #include "self_test_data.inc"
@@ -140,14 +141,19 @@ static int self_test_kdf(const ST_KAT_KDF *t, OSSL_ST_EVENT *event,
                          OPENSSL_CTX *libctx)
 {
     int ret = 0;
-    int i;
+    int i, numparams;
     unsigned char out[64];
     EVP_KDF *kdf = NULL;
     EVP_KDF_CTX *ctx = NULL;
     OSSL_PARAM params[16];
     const OSSL_PARAM *settables = NULL;
 
+    numparams = OSSL_NELEM(params);
     SELF_TEST_EVENT_onbegin(event, OSSL_SELF_TEST_TYPE_KAT_KDF, t->desc);
+
+    /* Zeroize the params array to avoid mem leaks on error */
+    for (i = 0; i < numparams; ++i)
+        params[i] = OSSL_PARAM_construct_end();
 
     kdf = EVP_KDF_fetch(libctx, t->algorithm, "");
     ctx = EVP_KDF_CTX_new(kdf);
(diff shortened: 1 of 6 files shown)


slontis:
rebased

slontis:
Note the related fipsprov.c code changed in a previous merge so the rebase removed the fix..

slontis:
ping

crypto/dsa/dsa_ossl.c
```diff
@@ -95,7 +95,7 @@ DSA_SIG *dsa_do_sign_int(OPENSSL_CTX *libctx, const unsigned char *dgst,
     blind = BN_CTX_get(ctx);
     blindm = BN_CTX_get(ctx);
     tmp = BN_CTX_get(ctx);
-    if (tmp == NULL)
+    if (m == NULL || tmp == NULL)
```
mattcaswell: This is not necessary. You, by design, only need to check the last call the `BN_CTX_get()`

providers/fips/self_test_kats.c
```diff
@@ -149,6 +149,9 @@ static int self_test_kdf(const ST_KAT_KDF *t, OSSL_ST_EVENT *event,
 
     SELF_TEST_EVENT_onbegin(event, OSSL_SELF_TEST_TYPE_KAT_KDF, t->desc);
 
+    /* Terminate the array just in case we exit early */
+    params[0] = OSSL_PARAM_construct_end();
```
mattcaswell: This doesn't seem to quite work. If we exit early half way through the loop below, params will still not be terminated.

crypto/dsa/dsa_ossl.c
slontis: I am aware of this. Coverity however I dont think knows this, so it may be thinking that tmp is 'possibly' NULL. I can mark it as a false positive if you prefer - but thought it wasnt a big problem adding the extra check..

providers/fips/self_test_kats.c
slontis: Changed to clear the params (also added assert that we dont write to/past the end of the params array)

crypto/dsa/dsa_ossl.c
richsalz: Mark it as a false positive.  Adding the extra check will mean that future people reading the code will not understand the semantics properly.  We already have enough problems with that, as this diff shows :)

richsalz: Since we now use travis_terminate, this is less important that it used to be, but no harm in keeping it as-is.

mattcaswell: I agree. This should be marked as a false positive.

slontis: Done - rebased to update the commit + updated coverity database comment

providers/fips/self_test_kats.c
```diff
@@ -140,29 +141,35 @@ static int self_test_kdf(const ST_KAT_KDF *t, OSSL_ST_EVENT *event,
...
     EVP_KDF_CTX *ctx = NULL;
     OSSL_PARAM params[16];
     const OSSL_PARAM *settables = NULL;
 
+    numparams = OSSL_NELEM(params);
     SELF_TEST_EVENT_onbegin(event, OSSL_SELF_TEST_TYPE_KAT_KDF, t->desc);
 
+    /* Zeroize the params array to avoid mem leaks on error */
+    for (i = 0; i < numparams; ++i)
+        params[i] = OSSL_PARAM_construct_end();
```
mattcaswell: A memset would probably do this just as well...but ok.

slontis:
Thanks. Merged to master.

PR #10942, opened 2020-01-24
Refactor SM2

levitte:
Up until now, applications had to set the type of SM2 `EVP_PKEY`s explicitly for SM2 processing to take place, with a call like this:

``` C
    if (EVP_PKEY_id(pkey) == EVP_PKEY_EC
        && (eckey = EVP_PKEY_get0_EC_KEY(pkey)) != NULL
        && (group = EC_KEY_get0_group(eckey)) != NULL
        && EC_GROUP_get_curve_name(group) == NID_sm2)
        EVP_PKEY_set_alias_type(EVP_PKEY_SM2);
```

The goal of this refactoring is to eliminate the need for such extra application side processing, and instead having that happen automatically and an `EC_KEY` is assigned to an `EVP_PKEY`.

Furthermore, the functions `X509_REQ_verify()` and `X509_verify()` were doing some heavy lifting with SM2 keys, via a function that was almost a straight copy of `ASN1_item_verify()`.  This is replaced with a new function `ASN1_item_verify_ctx()` that takes an `EVP_MD_CTX` pointer instead of a `EVP_PKEY` one, and through which all sorts of parameters can be passed, such as the SM2 identity.
Doing this also prepares for a provider side key adaptation.


levitte:
I'm preparing yet another in the series of fixes for #10797 issues, which depends on this cleanup of `X509_verify()` and `X509_REQ_verify()`.  With this, it will be very difficult to go forward with that.

crypto/sm2/sm2_pmeth.c
```diff
@@ -168,21 +158,11 @@ static int pkey_sm2_ctrl(EVP_PKEY_CTX *ctx, int type, int p1, void *p2)
 
     switch (type) {
     case EVP_PKEY_CTRL_EC_PARAMGEN_CURVE_NID:
-        group = EC_GROUP_new_by_curve_name(p1);
-        if (group == NULL) {
+        if (p1 != NID_sm2
+            || (group = EC_GROUP_new_by_curve_name(p1)) == NULL) {
             SM2err(SM2_F_PKEY_SM2_CTRL, SM2_R_INVALID_CURVE);
             return 0;
```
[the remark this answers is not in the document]
romen: To this end, it probably make sense to add tests that showcase
- [x] how to create an EVP:EC key over the curve defined in the SM2 standard and ensure it can perform EC keygen and compute ECDH and ECDSA sign/verify [I saw you already updated the tests to include this]
- [ ] and a separate tests that showcases how to create an EVP:SM2 key over another arbitrary curve and verify that it does SM2 keygen and SM2 kexchange, sign/verify.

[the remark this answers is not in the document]
levitte: > I am not sure about this: doesn't this effectively prevent to execute SM2 keygen and signatures on a different curve than the standard one?

The only thing this prevents is setting an arbitrary curve on an SM2 EVP_PKEY, but since you can switch back and forth between `EVP_PKEY_EC` and `EVP_PKEY_SM2` with `EVP_PKEY_set_alias_type()`, it's perfectly possible to either create a normal EC key with whatever curve you want or switch an existing one to `EVP_PKEY_EC` and do what you need, then switch it to be `EVP_PKEY_SM2` and get SM2 computations.

This is, of course, complete hackery, non-standard stuff, and it *should* have to be done explicitly.

As for SM2 computation over another arbitrary curve, I'm perfectly willing to do that if someone (you?) provides me with a suitable test vector, i.e. a key, input, sm2 id, and signature output.  It should be someone else than me producing that, to avoid being self serving.
There's no need to create a new key, we already have plenty of keys to choose from in evppkey.txt.

crypto/ec/ec_ameth.c
```diff
@@ -507,9 +507,9 @@ static int ec_pkey_ctrl(EVP_PKEY *pkey, int op, long arg1, void *arg2)
         if (EVP_PKEY_id(pkey) == EVP_PKEY_SM2) {
             /* For SM2, the only valid digest-alg is SM3 */
             *(int *)arg2 = NID_sm3;
-        } else {
-            *(int *)arg2 = NID_sha256;
+            return 2;            /* Make it mandatory */
         }
+        *(int *)arg2 = NID_sha256;
```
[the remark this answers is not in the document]
levitte: Just this part, right?

apps/req.c
```diff
@@ -1814,23 +1778,19 @@ int do_X509_CRL_sign(X509_CRL *x, EVP_PKEY *pkey, const EVP_MD *md,
...
-         * only in SM2 case we need to free the pctx explicitly
-         * if do_sign_init() fails, no need to double free pctx
+         * With SM2, do_sign_init() attached an EVP_PKEY_CTX to the
+         * EVP_MD_CTX, and we have to free it explicitly.
          */
-        if (ec_pkey_is_sm2(pkey)) {
-            pctx = EVP_MD_CTX_pkey_ctx(mctx);
+        if (EVP_PKEY_id(pkey) == EVP_PKEY_SM2) {
+            EVP_PKEY_CTX *pctx = EVP_MD_CTX_pkey_ctx(mctx);
+
```
beldmit: I'm not fond of this piece of code repeating 3 times but have no idea how to deal with it.

crypto/ec/ec_ameth.c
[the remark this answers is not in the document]
beldmit: Hmmm... There was some construction to establish link between a signature and hash algorithm somewhere, and it was used for GOST algorithms.

apps/req.c
levitte: Good point, I'll refactor that.

levitte: Done

beldmit: It's significantly better but couldn't it be done in the sm2-related code, not in general code?

crypto/sm2/sm2_pmeth.c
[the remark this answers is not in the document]
levitte: Dead silence...

Ok, I worked `tesst/ecdsatest.c` a bit to run all curves with "normal" EC computations as well as with SM2 computations.  That should cover everything you asked for.

apps/req.c
```diff
@@ -1748,90 +1722,64 @@ static int do_sign_init(EVP_MD_CTX *ctx, EVP_PKEY *pkey,
...
-         */
-        if (ec_pkey_is_sm2(pkey)) {
-            pctx = EVP_MD_CTX_pkey_ctx(mctx);
-            EVP_PKEY_CTX_free(pctx);
-        }
-#endif
-    }
+    if (do_sign_init(mctx, pkey, md, sigopts) > 0
+        && X509_sign_ctx(x, mctx) > 0
+        && do_sign_cleanup(mctx, pkey) > 0)
```
mattcaswell: Previously we cleaned up regardless of the return code from `X509_sign_ctx`. Now we only clean up if it was successful. That doesn't seem quite right.

apps/req.c
```diff
@@ -1748,90 +1722,64 @@ static int do_sign_init(EVP_MD_CTX *ctx, EVP_PKEY *pkey,
...
-         */
-        if (ec_pkey_is_sm2(pkey)) {
-            pctx = EVP_MD_CTX_pkey_ctx(mctx);
-            EVP_PKEY_CTX_free(pctx);
-        }
-#endif
-    }
+    if (do_sign_init(mctx, pkey, md, sigopts) > 0
+        && X509_REQ_sign_ctx(x, mctx) > 0
+        && do_sign_cleanup(mctx, pkey) > 0)
```
mattcaswell: And here

apps/req.c
```diff
@@ -1748,90 +1722,64 @@ static int do_sign_init(EVP_MD_CTX *ctx, EVP_PKEY *pkey,
...
-         */
-        if (ec_pkey_is_sm2(pkey)) {
-            pctx = EVP_MD_CTX_pkey_ctx(mctx);
-            EVP_PKEY_CTX_free(pctx);
-        }
-#endif
-    }
+    if (do_sign_init(mctx, pkey, md, sigopts) > 0
+        && X509_CRL_sign_ctx(x, mctx) > 0
+        && do_sign_cleanup(mctx, pkey) > 0)
```
mattcaswell: And here

include/openssl/x509.h
```diff
@@ -637,6 +637,9 @@ int ASN1_item_digest(const ASN1_ITEM *it, const EVP_MD *type, void *data,
 
 int ASN1_item_verify(const ASN1_ITEM *it, X509_ALGOR *algor1,
                      ASN1_BIT_STRING *signature, void *data, EVP_PKEY *pkey);
+int ASN1_item_verify_ctx(const ASN1_ITEM *it, X509_ALGOR *algor1,
```
mattcaswell: There seems to be no documentation for this new function. Should it be public or internal?

crypto/evp/p_lib.c
```diff
@@ -401,8 +401,19 @@ ENGINE *EVP_PKEY_get0_engine(const EVP_PKEY *pkey)
 # endif
 int EVP_PKEY_assign(EVP_PKEY *pkey, int type, void *key)
 {
+    int alias = type;
+
+    if (EVP_PKEY_type(type) == EVP_PKEY_EC) {
+        const EC_GROUP *group = EC_KEY_get0_group(key);
+
+        if (group != NULL && EC_GROUP_get_curve_name(group) == NID_sm2)
+            alias = EVP_PKEY_SM2;
```
mattcaswell: I think this change of behaviour deserves a CHANGES entry

crypto/sm2/sm2_pmeth.c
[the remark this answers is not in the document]
mattcaswell: This definitely requires a CHANGES entry! Strictly speaking this is a breaking change - should we allow it? Does this require an OMC decision since it is against policy.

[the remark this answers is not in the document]
mattcaswell: ...oh but wait....on looking at this a little more deeply it seems that although you could set these parameters to arbitrary curves there was no way to use them because the paramgen and keygen functions were previously NULL. AFAICT the previous `gen_group` field was write only and never used! Am I right or did I miss something. If I'm right maybe this isn't so bad.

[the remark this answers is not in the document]
romen: I can't really sit in front of a computer to run tests and figure all the implications for now sorry!

Hopefully some other reviewer can have a look at it? Apart from the things I mentioned and Richard evaluated I did not have any blocking comment looking at the code on github!

crypto/sm2/sm2_pmeth.c
```diff
@@ -317,10 +329,10 @@ static const EVP_PKEY_METHOD sm2_pkey_meth = {
     pkey_sm2_cleanup,
 
     0,
-    0,
+    pkey_sm2_paramgen,
 
     0,
-    0,
+    pkey_sm2_keygen,
```
mattcaswell: So, if I understand this correct previously we could not gen params and keys directly for SM2. You had to gen them for EC and then convert the resulting key to SM2. With these changes we will be able to generate SM2 keys directly (but only on the SM2 curve). The old way should still work, and would allow SM2 over arbitrary curves. Right?

crypto/x509/x_all.c
```diff
@@ -20,175 +20,76 @@
...
-    /* Convert signature OID into digest and public key OIDs */
-    if (!OBJ_find_sigid_algs(OBJ_obj2nid(a->sig_alg.algorithm),
-                             &mdnid, &pknid)) {
-        X509err(X509_F_X509_VERIFY, ASN1_R_UNKNOWN_SIGNATURE_ALGORITHM);
-        return 0;
-    }
-
-    if (pknid == NID_sm2)
-        return x509_verify_sm2(a, r, mdnid, pknid);
+    ctx = make_id_ctx(r, a->sm2_id);
```
mattcaswell: Wouldn't it be better if `make_id_ctx` created a ctx regardless of whether sm2_id is NULL or not. That way we can just call `ASN1_item_verify_ctx` unconditionally below...can't we?

crypto/x509/x_all.c
```diff
@@ -20,175 +20,76 @@
...
-        X509err(X509_F_X509_REQ_VERIFY, ASN1_R_UNKNOWN_SIGNATURE_ALGORITHM);
-        return 0;
-    }
+    int rv = 0;
+    EVP_MD_CTX *ctx = NULL;
 
-    if (pknid == NID_sm2)
-        return x509_req_verify_sm2(a, r, mdnid, pknid);
+#ifndef OPENSSL_NO_SM2
+    ctx = make_id_ctx(r, a->sm2_id);
```
mattcaswell: Same comment as above

doc/man3/EVP_PKEY_set1_RSA.pod
```diff
@@ -103,6 +102,14 @@ EVP_PKEY_assign_RSA(), EVP_PKEY_assign_DSA(), EVP_PKEY_assign_DH(),
 EVP_PKEY_assign_EC_KEY(), EVP_PKEY_assign_POLY1305()
 and EVP_PKEY_assign_SIPHASH() are implemented as macros.
 
+EVP_PKEY_assign_EC_KEY() looks at the curve name id to determine if
+the passed B<EC_KEY> is an L<SM2(7)> key, and will set the B<EVP_PKEY>
+type to B<EVP_PKEY_SM2> in that case, instead of B<EVP_PKEY_EC>.
+This B<EVP_PKEY> type can be switched to B<EVP_PKEY_EC> with a call to
+EVP_PKEY_set_alias_type() if it's desirable to do a normal EC
+computations with the SM2 curve instead of the special SM2
+computations.
```
mattcaswell: And vice versa...i.e. you can take a key for a non-SM2 curve and convert it to SM2.

mattcaswell (changes requested):

mattcaswell:
I'd really like it if @InfoHunter commented on this PR

include/openssl/x509.h
levitte: There is no documentation for the old either...

crypto/evp/p_lib.c
levitte: Thanks for the reminder

crypto/x509/x_all.c
levitte: Hmm, I'll look into it. I'm a bit sceptical, but...

doc/man3/EVP_PKEY_set1_RSA.pod
levitte: Yes. You can see it happen in full action in ecdsatest. Basically, you can call `EVP_PKEY_set_alias_type` at any type with the desired key type, `EVP_PKEY_EC` or `EVP_PKEY_SM2`, and it will work, the next operation will select a pmeth according to that type

crypto/sm2/sm2_pmeth.c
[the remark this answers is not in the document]
levitte: Previously, you always generated an EC key and then changed the type to `EVP_PKEY_SM2` with `EVP_PKEY_set_alias_type`, so it was effectively impossible to set the `gen_group` in the SM2 `EVP_PKEY_METHOD`, or use it.

With this PR, generating SM2 keys can still be done as before, *but* now you can also generate SM2 keys following the same pattern as other key types, i.e.:

``` C
    EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_id(EVP_PKEY_SM2, NULL);
    EVP_PKEY *pkey = NULL;

    EVP_PKEY_keygen_init(ctx);
    EVP_PKEY_keygen(ctx, &pkey);
```

Doing it this way *forces* the SM2 curve, which I would say it should.  If you want to generate an EC key with the SM2 curve, you just do as before:

``` C
    EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_id(EVP_PKEY_EC, NULL);
    EVP_PKEY *pkey = NULL;

    EVP_PKEY_keygen_init(ctx);
    EVP_PKEY_CTX_set_ec_paramgen_curve_nid(ctx, NID_sm2);
    EVP_PKEY_keygen(ctx, &pkey);
```

So in essence, I haven't changed anything already present, I have just added the possibility to generate SM2 keys directly.

However, looking at the code we're commenting in, I'm wondering why I create a group that I do nothing with.  I'm amazed that hasn't caused a leak yet...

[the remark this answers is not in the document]
levitte: Obviously, I disagree regarding this being a breaking change.  I tried to avoid that quite carefully.

apps/req.c
levitte: Oh, good point!  Will fix

apps/req.c
levitte: Yeah

apps/req.c
levitte: Yeah

crypto/sm2/sm2_pmeth.c
levitte: Correct.  `test/ecdsatest.c` does a lot of tests of exactly this.  As a matter of fact, you can also generate a SM2 key and then switch its type to `EVP_PKEY_EC`

crypto/x509/x_all.c
levitte: Yeah, ok

doc/man3/EVP_PKEY_set1_RSA.pod
levitte: Heh, I discover that I'm repeating myself...  ah well :wink:

apps/req.c
levitte: Done

apps/req.c
levitte: Done

apps/req.c
levitte: Done

crypto/evp/p_lib.c
levitte: Done

crypto/sm2/sm2_pmeth.c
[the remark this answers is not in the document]
levitte: I changed to not create the group, and added a pretty thorough comment.

crypto/x509/x_all.c
levitte: Done

crypto/x509/x_all.c
levitte: Done

doc/man3/EVP_PKEY_set1_RSA.pod
levitte: Done

levitte:
Rebased, and I hope I've addressed all  concerns.  Please have a new look

beldmit:
Hmmm... Are travis failures relevant?

levitte:
> Hmmm... Are travis failures relevant?

Meh, very much so...
(was the question retorical?)

beldmit:
No, it was not, I wasn't sure if it is a problem of the PR or master.

levitte:
I've fixed all the things I could catch on my laptop, let's hope the CIs get happier

beldmit (approved):
Nice work!

include/openssl/x509.h
mattcaswell: I don't see this as a good enough reason to not document the new function. In most cases Travis would have actually complained about this missing documentation - but due to the issue fixed in #10980 it didn't. We should never again add new public symbols that aren't documented.

mattcaswell:
Looks good apart from the documentation issue - which I do think needs resolving.

include/openssl/x509.h
levitte: I would like to document *all* those functions, not just one.  Will you be satisfied with a separate PR?

mattcaswell: Well...in principle "yes"....but....

https://github.com/openssl/openssl/pull/8442/files#r373407266

levitte: Gotcha.  Still, it's a bit too much for this PR.  Make your mind up :wink:

mattcaswell: Ok...separate PR....but I'm going to check this one too ;-)

mattcaswell (approved):

include/openssl/x509.h
levitte: I'll make them a priority after lunch, how's that sound?

mattcaswell: WFM

levitte:
Merged

f4e4382cae1fb85fec6e9aa26f65fc729a40a039 EVP_PKEY_assign_EC_KEY(): detect SM2 curve and set EVP_PKEY type accordingly
0943d5dd6179e2e824ec9a38b7ea6caa17a84ebd Add SM2 specific parameter and key generation
bbaddbc0682a4cd3271ee41f58550f7c26e8194e X509: Refactor X509_verify() and X509_REQ_verify() for better streamlining
ef077ba0d2fa28fd1b481f80335bda723d3d1f20 Make SM3 a mandatory hash function for SM2.
3995de2c0356d8e3b34e8fb2ddf22b20210e498d Adapt tests for SM2 changes.
bac1030ae4f93f22b3a81e3a2d9d3c5db363d96f Adapt some 'openssl' commands for SM2 changes.
7f293d9f3b5cee4f4b15624fff15a45e0517334f CHANGES: Add note about the refactoring of SM2 EVP_PKEYs

crypto/sm2/sm2_pmeth.c
```diff
@@ -163,26 +153,21 @@ static int pkey_sm2_decrypt(EVP_PKEY_CTX *ctx,
...
-        group = EC_GROUP_new_by_curve_name(p1);
-        if (group == NULL) {
+        /*
+         * This control could be removed, which would signal it being
+         * unsupported.  However, that means that when the caller uses
+         * the correct curve, it may interpret the unsupported signal
+         * as an error, so it's better to accept the control, check the
+         * value and return a corresponding value.
+         */
+        if (p1 != NID_sm2) {
```
InfoHunter: I am sorry for the very late response - Just got back from the Chinese lunar new year vacation.

I would like to make a note here explaining that why we just use `NID_sm2` and ignore other curves. Though the current `NID_sm2` is the only recommended curve but SM2 algorithms (signature/encryption) can actually take other curves as well - if I recall correctly there exist test vectors for SM2 using other curves in some SM2 specification documents.

For OpenSSL I don't think we need to provide the ability to set other curves other than NID_sm2 because the NID_sm2 curve is the only one being used in practice. And when we talk about SM2 in different standard organization (ISO/IETF/...), there is also one curve NID_sm2.

But I still think it's better to leave a comment here (and the `paramgen` function below as well) that OpenSSL implements only the NID_sm2 curve and provides no other capability of setting other customized curves so that OpenSSL's implementation can only be valid with the SM2 test vectors that use NID_sm2.

@levitte ,what do you think?

doc/man3/EVP_PKEY_set1_RSA.pod
```diff
@@ -103,6 +102,16 @@ EVP_PKEY_assign_RSA(), EVP_PKEY_assign_DSA(), EVP_PKEY_assign_DH(),
...
 and EVP_PKEY_assign_SIPHASH() are implemented as macros.
 
+EVP_PKEY_assign_EC_KEY() looks at the curve name id to determine if
+the passed B<EC_KEY> is an L<SM2(7)> key, and will set the B<EVP_PKEY>
+type to B<EVP_PKEY_SM2> in that case, instead of B<EVP_PKEY_EC>.
+
+It's possible to switch back and forth between the types B<EVP_PKEY_EC>
+and B<EVP_PKEY_SM2> with a call to EVP_PKEY_set_alias_type() on keys
+assigned with this macro if it's desirable to do a normal EC
+computations with the SM2 curve instead of the special SM2
```
InfoHunter: I find this paragraph is misleading that it sounds we provide a way to encourage users to use the the SM2 curve to do a normal EC computation, which I think it should be discouraged to do so. Or should we make it mandatory to use SM2 curve with SM2 special computation only?

crypto/sm2/sm2_pmeth.c
levitte: I didn't limit it like that. What I did was to change the *default* behaviour, but you can still change the `EVP_PKEY` type with `EVP_PKEY_set_alias_type`

PR #10943, opened 2020-01-24
(minor) session thread-safety fixes

kaduk:
Plus some other random fixes encountered along the way.

WIP since I still want to write a test case that exercises the last commit (renegotiation changing cipher on resumed connection), but feedback on the other points is welcome.

This is all that came out from a deliberate audit of code conditional on ```s->hit``` and/or ```SSL_IS_TLS13()``` looking for places that might access ```s->session``` in an unsafe manner, with particular attention to instances of ```!s->hit || SSL_IS_TLS13(.)``` that led to #10441, #9162, etc.
There is an instance of that in ```ssl_update_cache()``` that is worrisome at first sight but is currently okay due to properties that currently hold invariant for all callers but we lack enforcement of: for TLS 1.3 on the server side, we only call ```ssl_update_cache()``` when generating NewSessionTicket messages, and ```tls_construct_new_session_ticket()``` will always generate a new session object in cases that could cause thread-safety issues:

```
        /*
         * If we already sent one NewSessionTicket, or we resumed then
         * s->session may already be in a cache and so we must not modify it.
         * Instead we need to take a copy of it and modify that.
         */
        if (s->sent_tickets != 0 || s->hit) {
            SSL_SESSION *new_sess = ssl_session_dup(s->session, 0);
```

This means that, in ```ssl_update_cache()``` as called from ```tls_construct_new_session_ticket()```, we will never be "updating" the cache in the sense of having a preexisting entry to be modified/replaced, but always adding a new one.  Perhaps it would be worth making a new function or an input parameter to ```ssl_update_cache()``` for the caller to indicate that this is expected to be the case, to allow for some runtime checking on this property that we want to remain invariant.


mspncp:
> Plus some other random fixes encountered along the way.

_(Funny: it's already the second time that it happens to me that I read "random fixes" somewhere here on GitHub and an internal alarm bell rings, until I realize that it does not say "random **generator** fixes" :-) )_

kaduk:
@mspncp sorry!  They are better characterized as "miscellaneous" than "random", now that you mention it :)

mspncp:
> They are better characterized as "miscellaneous" than "random", now that you mention it :)

Now that you mention it: your fixes should better **not** be random! No time to wait for [infinite monkeys](https://en.wikipedia.org/wiki/Infinite_monkey_theorem) ;-)

kaduk:
Finally got back to this one and added a test case, so taking out of WIP. (Rebased, too, for good measure.)

It turns out that the questionable logic in ChangeCipherSpec handling never actually triggers (as revealed by the test I wrote), so I was able to simplify the logic to just only change away from a NULL  cipher in the session, and to abort the handshake if we try to make any other changes to the cipher in the session.

kaduk:
> Finally got back to this one and added a test case

It helps if I actually push, whoops.  "This time out of WIP for sure."

kaduk:
Setting the 1.1.1 label since I think these fixes will also apply there

t8m:
The Travis failure looks relevant?

kaduk:
Ah, "one last-minute fix" strikes again.  Thanks for calling it out; things are pretty hectic over here...

kaduk:
Travis seems happier with the `make update` fixup commit.
(Appveyor was happy without it, but is backed up at the moment.)

test/sslapitest.c
```diff
@@ -651,6 +651,121 @@ static int test_no_ems(void)
 }
 #endif
 
+/*
+ * Very focused test to exercise a single case in the server-side state
+ * machine, when the ChangeCipherState message needs to actually change
+ * from one cipher to a different cipher (i.e., not changing from null
+ * encryption to reall encryption).
```
t8m: Typo: reall
But feel free to fix it when merging as this is certainly a trivial change.

kaduk: Whoops, I seem to have missed this typo fix :(

kaduk:
Cherry-pick had to cope with the new `SSL_F_OSSL_STATEM_SERVER_PRE_WORK` and the collapsing of `s->s3` but it's clean enough that I'm happy to stick with this PR number.
Pushed to master and 1.1.1, so closing.

mattcaswell:
This seems to have broken the 1.1.1 Travis build!! Not good...we're releasing tomorrow!!

test/sslapitest.c
```diff
@@ -651,6 +651,121 @@ static int test_no_ems(void)
...
+    serverssl = clientssl = NULL;
+
+    /*
+     * Now create a fresh connection and try to renegotiate a different
+     * cipher on it.
+     */
+    if (!TEST_true(create_ssl_ctx_pair(TLS_server_method(),
+                                       TLS_client_method(),
+                                       TLS1_VERSION, TLS1_2_VERSION,
+                                       &sctx, &cctx, cert, privkey))
```
mattcaswell: This seems to leak sctx and cctx from the earlier `create_ssl_ctx_pair` call. Do we need new CTXs here?

kaduk: We shouldn't need new ctxes, no.

kaduk:
Testing a fix now (which will also include the typo in the comment that I failed to pick up while merging)

mattcaswell:
Great - thanks

PR #10944, opened 2020-01-25
test/recipes/30-test_evp.t: Fix multiple definition of @bffiles

levitte:
Curiously enough, perl only warned about the shadowing.  However, the
following 'plan' statement got disturbed somehow, as one could notice
the test counter say "11/?" instead of "11/25".


beldmit (approved):

levitte:
I'm pondering if a label like "janitorial" would be a good thing, which would exempt a PR from the 24h grace while not being quite as *urgent* as the "urgent" label implies, but where the 24h grace period is just silly.  Such as this PR.

levitte:
Merged.

4ff5b41af15ce731f03bbb43250c8ef1ed88b513 test/recipes/30-test_evp.t: Fix multiple definition of @bffiles

Issue #10814, opened 2020-01-12
(dgst+hmac) MAC parameter error "key:123456901234567890123456789012"


vt-alt:
While testing [gost-engine](https://github.com/gost-engine/engine/pull/187#issuecomment-573364326) I've stumbled upon this issue with openssl master branch:

```
openssl (master)$ apps/openssl dgst -sha256 -mac hmac -macopt key:123456901234567890123456789012 </dev/null
MAC parameter error "key:123456901234567890123456789012"

openssl (master)$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
nothing to commit, working tree clean

openssl (master)$ git describe
OpenSSL_1_1_1-2428-ge0e68f9e34
openssl (master)$
```

levitte:
Yup, I can see why this happens, an think I know how to deal with it.  I'll be back.

slontis:
Multiple things are wrong..
(1) It looks like if mac_name is "hmac" then it needs to tack the digest md onto the macopts
i.e apps/openssl dgst -sha256 -mac hmac -macopt md:sha256 -macopt key:1234569012345678901234567890123 will get you a little further..

(2) It then tries to do EVP_PKEY_keygen(mac_ctx, ).. which falls over.. (Not sure what that is trying to do).

There is a openssl mac application that will work for HMAC that only uses -macopt..

levitte:
No, @slontis. The issue is that the ctrl_str that's being called is not coded correctly, as it will not pass command names that libcrypto doesn't recognise.

levitte:
I found out why this happens...  It's the case of raw keys that's not properly intercepted in `pkey_mac_str_ctrl()`.  We really shouldn't need to intercept that, but since it has local implications, we must.

levitte:
It turns out that my analysis was incorrect...  or well, partly incorrect.  #10947 solves the general problem of passing arbitrary string controls to provider implementations, but the error persists, so I did a bit of debugging, and it turns out that the interaction between the EVP_PKEY_METHOD \<-\> EVP_MAC bridge `crypto/evp/pkey_mac.c` handles the HMAC case differently than in 1.1.1.

[...]

PR #10947, opened 2020-01-25
Decentralize legacy_ctrl_str_to_param()

levitte:
This function did a bit too much in terms of central control, actually
more so than the legacy counterpart, where all the string processing
is done in the diverse *_pmeth.c.  Furthermore, there was no room
whatsoever for control keys that libcrypto isn't centrally aware of.

This function is changed to simply translating keys and values to
OSSL_PARAM form and then sent on their merry way to the provider
implementations through EVP_PKEY_CTX_set_params().  It translates
selected well known legacy names to their core name counterpart, and
that's as far as centralized control should extend.

There were consequences, of course.  For RSA, The padding mode could
be given in form of words, and that got moved to the RSA ASYM_CIPHER
implementation.  For DSA, it turned out that the SIGNATURE implementation
didn't like to get a "digest" param without the "digest-size" param...


levitte:
It was always the intention that string controls would be directly, simply and fairly uniformly translated to OSSL_PARAM.  Let's remember that libcrypto is becoming a framework, and can therefore not assume too much knowledge of what the provided implementations want.

slontis:
travis failures need addressing

providers/implementations/asymciphers/rsa_enc.c
```diff
@@ -37,6 +37,16 @@ static OSSL_OP_asym_cipher_gettable_ctx_params_fn rsa_gettable_ctx_params;
 static OSSL_OP_asym_cipher_set_ctx_params_fn rsa_set_ctx_params;
 static OSSL_OP_asym_cipher_settable_ctx_params_fn rsa_settable_ctx_params;
 
+static OSSL_ITEM padding_item[] = {
+    { RSA_PKCS1_PADDING,        "pkcs1"  },
+    { RSA_SSLV23_PADDING,       "sslv23" },
+    { RSA_NO_PADDING,           "none"   },
+    { RSA_PKCS1_OAEP_PADDING,   "oaep"   }, /* Correct spelling first */
+    { RSA_PKCS1_OAEP_PADDING,   "oeap"   },
```
slontis: dooh!

providers/implementations/asymciphers/rsa_enc.c
```diff
@@ -304,7 +341,7 @@ static int rsa_get_ctx_params(void *vprsactx, OSSL_PARAM *params)
 
 static const OSSL_PARAM known_gettable_ctx_params[] = {
     OSSL_PARAM_utf8_string(OSSL_ASYM_CIPHER_PARAM_OAEP_DIGEST, NULL, 0),
-    OSSL_PARAM_int(OSSL_ASYM_CIPHER_PARAM_PAD_MODE, NULL),
+    OSSL_PARAM_utf8_string(OSSL_ASYM_CIPHER_PARAM_PAD_MODE, NULL, 0),
```
slontis: This is a bit messed up when it uses multiple types.

providers/implementations/asymciphers/rsa_enc.c
```diff
@@ -353,8 +390,27 @@ static int rsa_set_ctx_params(void *vprsactx, const OSSL_PARAM params[])
...
+        case OSSL_PARAM_INTEGER: /* Support for legacy pad mode number */
+            if (!OSSL_PARAM_get_int(p, &pad_mode))
+                return 0;
+            break;
+        case OSSL_PARAM_UTF8_STRING:
+            {
+                int i;
+
+                for (i = 0; padding_item[i].id != 0; i++) {
+                    if (strcmp(p->data, padding_item[i].ptr) == 0) {
```
slontis: p->data == NULL would be bad here..

providers/implementations/signature/dsa.c
```diff
@@ -50,11 +53,50 @@ typedef struct {
     OPENSSL_CTX *libctx;
     DSA *dsa;
     size_t mdsize;
+    unsigned int flag_allow_md : 1;
```
slontis: Maybe put the reason for this flag as a comment (like you do when you actually use it)

providers/implementations/asymciphers/rsa_enc.c
levitte: Well, I'm thinking that one is legacy support and doesn't need announcement, so...

providers/implementations/asymciphers/rsa_enc.c
levitte: Good point

providers/implementations/signature/dsa.c
levitte: Will do

providers/implementations/asymciphers/rsa_enc.c
levitte: Not sure if that was just an exclamation or if you're trying to say something...  anyway, this is copied from existing legacy code, so it's obviously there for a reason...

https://github.com/openssl/openssl/blob/b0031e5dc2c8c99a6c04bc7625aa00d3d20a59a5/crypto/rsa/rsa_pmeth.c#L597-L619

providers/implementations/asymciphers/rsa_enc.c
levitte: Fixed

providers/implementations/signature/dsa.c
levitte: Done

providers/implementations/signature/dsa.c
```diff
@@ -63,9 +122,31 @@ static void *dsa_newctx(void *provctx)
...
+    if (mdname != NULL) {
+        EVP_MD *md = EVP_MD_fetch(ctx->libctx, mdname, mdprops);
+        int md_nid = dsa_get_md_nid(md);
+
+        EVP_MD_free(ctx->md);
+        ctx->md = NULL;
+        ctx->mdname[0] = '\0';
+
+        if (md_nid == NID_undef)
+            return 0;
```
slontis: looks like md would get leaked here?

levitte: Fixed

slontis (approved):
LGTM if the tests pass.

crypto/evp/pmeth_lib.c
```diff
@@ -816,76 +816,37 @@ int EVP_PKEY_CTX_ctrl_uint64(EVP_PKEY_CTX *ctx, int keytype, int optype,
...
-        if (lab == NULL)
+        if (!OSSL_PARAM_allocate_from_text(&params[0], settable, name, value,
+                                           strlen(value)))
             return 0;
-        ret = EVP_PKEY_CTX_set0_rsa_oaep_label(ctx, lab, lablen);
-        if (ret <= 0)
-            OPENSSL_free(lab);
-        return ret;
+        if (EVP_PKEY_CTX_set_params(ctx, params))
+            rv = 1;
```
romen: As it is this is not flexible enough:
- for some params we must make sure that the provider supports them as settable as ignoring a param in that case would be breaking existing legacy behavior and open security risks: e.g., `name == "ecdh_cofactor_mode"` if the provider does not support this parameter as settable and does not react on this request this is not signaled to the application with an error (i.e., breaking existing behavior) and the ECDH computation will be performed without the cofactor-tweak,  exposing to serious small subgroups attacks! (i.e. security risk)
- for some params we used to raise specific errors in specific cases: e.g., `name == ecdh_cofactor_md` we pass the name of the digest algorithm and we used to raise `EC_R_INVALID_DIGEST` if `EVP_get_digestbyname(value) == NULL`. This might be a minor concern but going directly as a param for the provider we add a useless round-trip (as we would expect the provider `ctx_set_param` implementations to try and go back into core to fetch the matching digest).

The problem of ignoring by default unknown parameters on a `set_params` (or `get_params`) and breaking existing behavior actually applies to both my examples (and I suspect also to the cases that this PR is already supporting.

If we tried to explicitly tweak an operation by setting a non-default digest, some special padding mode or a specific mode of operation, etc., we can't return success to the application (or `libcrypto`/`libssl`) if this has not been received correctly by the provider: this is required to not break existing behavior, but also to avoid undetected unsafe behavior!

romen (changes requested):
Actually I have a concern!

crypto/evp/pmeth_lib.c
romen: One more point that I forgot to mention before: some values might be received as strings by this function but might need conversion as the OSSL_PARAM type is not string (or is a string but not the value passed in as a string in the legacy API).

One example is once more `ecdh_cofactor_mode`. In this function the value is a string, but the OSSL_PARAM is an int that must be either -1, 0 or 1.

levitte: Before making assumptions on strings vs numbers, you might want to read up on `OSSL_PARAM_allocate_from_text()`

https://github.com/openssl/openssl/blob/master/doc/man3/OSSL_PARAM_construct_from_text.pod

levitte: Also, I believe that this same function answers your concern about finding out if a parameter is actually settable.

levitte:
I had to fix a couple of things, so a re-approval is needed, regardless of if @romen's concern is already answered or not.

crypto/evp/pmeth_lib.c
romen: I apologize for missing that function!

I can't do a proper review at the moment, the important thing for me is that the concerns above are indeed addressed.

levitte: > I apologize for missing that function!

No problem.  There's so many new things that we're all prone to miss a few...

providers/implementations/signature/dsa.c
```diff
@@ -331,13 +370,13 @@ static int dsa_get_ctx_params(void *vpdsactx, OSSL_PARAM *params)
         return 0;
 
     p = OSSL_PARAM_locate(params, OSSL_SIGNATURE_PARAM_DIGEST_SIZE);
```
mattcaswell: If we can now support a digest without a digest size - do we even need digest size at all? (I think this is new in master)

providers/implementations/asymciphers/rsa_enc.c
```diff
@@ -353,8 +389,32 @@ static int rsa_set_ctx_params(void *vprsactx, const OSSL_PARAM params[])
...
     if (p != NULL) {
-        if (!OSSL_PARAM_get_int(p, &pad_mode))
+        int pad_mode = 0;
+
+        switch (p->data_type) {
+        case OSSL_PARAM_INTEGER: /* Support for legacy pad mode number */
+            if (!OSSL_PARAM_get_int(p, &pad_mode))
+                return 0;
+            break;
+        case OSSL_PARAM_UTF8_STRING:
```
mattcaswell: I'm not keen on this "dual type" solution at all. This makes it difficult for third party provider authors to provide a drop-in replacement. They may have to support the "legacy" way of doing things too - otherwise it may not work for all applications.

mattcaswell (changes requested):

crypto/evp/digest.c
```diff
@@ -662,10 +662,7 @@ int EVP_MD_CTX_ctrl(EVP_MD_CTX *ctx, int cmd, int p1, void *p2)
         return 0;
     }
 
-    if (ctx->digest->prov == NULL
-        || (ctx->pctx == NULL
-            || (ctx->pctx->operation != EVP_PKEY_OP_VERIFYCTX
-                && ctx->pctx->operation != EVP_PKEY_OP_SIGNCTX)))
```
slontis: @mattcaswell - can you confirm this change is good?

mattcaswell: I think it should be ok

providers/implementations/signature/dsa.c
levitte: I have wondered too, but figured that there are hash functions that can produce varying sizes according to user desire, so why not? They are not accepted today, but could in the future.

I would have no remorse removing it if you prefer.

providers/implementations/asymciphers/rsa_enc.c
levitte: Yeah well, I do not know how otherwise. Are you saying that we should translate numbers to strings in libcrypto, in whatever'se the name of the specialised setter and getter functions? It is kinda odd, considering that this is really entirely in the backend (in this regard, the ameth and pmeth files are equal to providers)

levitte: The thing is that, even with legacy, it depends on what backend is used. Could be an engine...

crypto/evp/digest.c
levitte: Leaving this code in means that some things stop working, for example the operation codes `EVP_PKEY_OP_SIGN` and `EVP_PKEY_OP_VERIFY` (they are perfectly legitimate in the legacy cases)

levitte:
Ping?

beldmit:
@vt-alt, could you please remind the test case?

vt-alt:
@beldmit `openssl dgst -sha256 -mac hmac -macopt key:123456901234567890123456789012 </dev/null` ?

beldmit:
Sure! Many thanks!

levitte:
Hmmm, that test fails...  but does that have anything to do with *this* PR?
I'm all for correcting that error, but would rather do that in a separate PR.

slontis:
On my local build I get
MAC parameter error "key:123456901234567890123456789012"..
So not directly caused by this PR.

providers/implementations/signature/dsa.c
slontis: @mattcaswell ?

beldmit:
> Hmmm, that test fails... but does that have anything to do with _this_ PR?
> I'm all for correcting that error, but would rather do that in a separate PR.

My misunderstanding, sorry.

levitte:
> So not directly caused by this PR.

Thanks.  I'll investigate

vt-alt:
I thought this is related to issue #10814 where @levitte said "I found out why this happens... It's the case of raw keys that's not properly intercepted in pkey_mac_str_ctrl(). We really shouldn't need to intercept that, but since it has local implications, we must.".

levitte:
I'll comment on it there

levitte:
I've commented in #10814.  That problem turns out not to be relevant here.

So, final review?

providers/implementations/signature/dsa.c
mattcaswell: I'd say lets remove it

providers/implementations/asymciphers/rsa_enc.c
mattcaswell: > Are you saying that we should translate numbers to strings in libcrypto, in whatever'se the name of the specialised setter and getter functions? 

I think I'd prefer that since its legacy only and will eventually go away. It keeps things simpler for provider authors.

(BTW, should we deprecate EVP_PKEY_CTX_ctrl, EVP_MD_CTX_ctrl, etc?...not for htis PR)

providers/implementations/signature/dsa.c
levitte: Okie

levitte: Done

levitte:
I removed the need to pass the digest size down to the implementation.  I felt the need to rearrange the commit sequence, so ended up making a squashing rebase.  This PR isn't very big, so I hope that won't make too much of a problem...

slontis:
Approval reconfirmed

slontis:
Rebase needed

levitte:
That was a trivial conflict, easily resolved.  This PR isn't changed by it, so I'll take that reconfirmation and merge.

levitte:
Just to be safe, I'll wait a bit for the CIs, though

levitte:
Merged

972fa31895b38cbe91a87a04875f7dadee387dea Decentralize legacy_ctrl_str_to_param()
00bc1ad99a69f851ccdea3656445ae2daaf70717 Don't pass a digest-size to signature implementations
31a796d1cc884c727a226d3dfba6a1ad7a50b865 PROV: Implement padding mode words in the RSA ASYM_CIPHER implementation
8bee651213c4755761ed5dd32d416d782cc5fc3c PROV: Fix the DSA SIGNATURE implementation for better digests handling
e3b1ccad694aabfffbde68c56fb8d44c011f98b1 EVP_MD_CTX_ctrl(): Remove unnecessary control

romen:
I know my issue comes late, but I couldn't find cycles to spare until now.

I see now another problem with this: by going directly to the provider interface with `EVP_PKEY_CTX_set_params()` from inside `legacy_ctrl_str_to_param()` we are bypassing the dedicated `EVP_PKEY_CTX_set_foo()`.
I get this is exactly the intent of this PR, but the problem I see with this now is that we are pushing beyond the provider boundary all the tweaks we might need to retain backward compatibility and I think this is wrong.

Yes, the code inside libcrypto looks nicer and saner this way, but this way we lose the ability to return specific return values for different error conditions or we force every external provider to implement whatever workarounds we devise inside the providers we distribute to retain the legacy behavior.

Examples from #10631 `"ecdh_kdf_md"` and `"ecdh_cofactor_mode"`:

when `EVP_PKEY_CTX_set_ecdh_{kdf_md,cofactor_mode}` is called, for backward compatibility it should return:
- `-2` if this `EVP_PKEY_CTX` is not for a `DERIVE` operation
   * does not concern the provider implementation
   * `libcrypto` should return `-2` to the caller if this condition is detected
- `-1` if this `EVP_PKEY_CTX` does not contain an `EVP_PKEY_EC` key
   * does not concern the provider implementation
   * `libcrypto` should return `-1` to the caller if this condition is detected
- `-2` if this `EVP_PKEY_CTX` does not support setting the `kdf_md` or the `cofactor_mode`
   * the latter is particularly relevant because we historically allowed for underdefined curves that do not specify the cofactor, as it is considered optional in the serialization.
     For such curves, the cofactor used to be set to `0` and with b783beeadf6b80bc431e6f3230b5d5585c87ef87 we are now attempting to compute it, but there might be curves for which the computation fails to deliver the cofactor and so ecdh_cofactor_mode cannot be turned on!
   * `libcrypto` should return `-2` if the provider does not signal these as settable parameters
   * the provider should return a failure if it understands these parameters but something goes wrong when trying to set them, e.g., unknown and uncomputable cofactor, but it is arguably `libcrypto` responsibility to know that this failure should be mapped to a `-2` return value for legacy compatibility

Unfortunately we have very limited unit testing about being consistent w.r.t. legacy about specific return values in case of errors or about checking that the same errors are added to the stack, so there is a chance that we are breaking backward compatibility with this PR and not detecting it also for the RSA and DH that are currently included here.

romen:
@levitte could you have a look at this?
At the moment I am rebasing #10631 and I am tempted to keep the structure of pre-10947 for the EC stuff in `legacy_ctr_str_to_param()` so that it just parses the input value and calls the dedicated legacy `EVP_PKEY_CTX_set_foo()` that knows what are the different return values to return in different conditions!

levitte:
> I see now another problem with this: by going directly to the provider interface with `EVP_PKEY_CTX_set_params()` from inside `legacy_ctrl_str_to_param()` we are bypassing the dedicated `EVP_PKEY_CTX_set_foo()`.

Er, yes?

So let's see if I can clarify something that seems to be a point of confusion, and I'm going to do see by using the term "backend", as a broad term for "implementations under the hood".  In pre-3.0 terms, any EVP_CIPHER, EVP_MD, EVP_PKEY_METHOD, EVP_PKEY_ASN1_METHOD implements a backend, whether it happens to be built in (our ameth and pmeth files) or coming from en engine, *or coming from a provider* in the case of EVP_CIPHER and EVP_MD. The EVP API per se doesn't really know, and *shouldn't* really know what the backends are up to, and this is very visible when you look at how the pre-3.0 `EVP_PKEY_CTX_ctrl_str()` is implemented (this is from 1.1.1):

https://github.com/openssl/openssl/blob/ff6d0a6313a9082965f798a3102539b8d83a6a39/crypto/evp/pmeth_lib.c#L407-L418

So what had happened previous to this PR was that code that belonged in the backends was moved to central EVP functions when it should have moved from the old backend (whatever ameth or pmeth file) to the new backends (providers).  That was wrong, on principle.

> we force every external provider to implement whatever workarounds we devise inside the providers we distribute to retain the legacy behavior.

Do we?  Would you claim that any ENGINE implementor that wanted to implement their version of, say, RSA, was forced to implement the exact behaviour the our backends (crypto/rsa/rsa_ameth.c or crypto/rsa/rsa_pmeth.c)?  To what extent?  'cause believe you me, that wasn't necessarily the case...

The very simple truth is central EVP code such as `EVP_PKEY_CTX_ctrl_str()` *cannot possibly know what a backend will or will not accept*, the design is to simply pass on what it gets, in both directions (except for, weirdly enough, "digest", in the pre-3.0 code...  that one confuses me a bit).

All this being said, you're quite right that there are things that aren't the same between the old (ameth and pmeth code) and the new (providers) backends, and that can be summed up in three parts:

1.  The control names aren't necessarily the same as the parameter names (we could have made them the same, but quite frankly, the old control names were unnecessarily wordy, like why the hell do we need names like "rsa_keygen_bits" when it should already be known that we're dealing with RSA, and that we're doing key generation?  "bits" can and should be enough!), so `legacy_ctrl_str_to_param` translates the names it currently know, and I fully expect to see that list getting longer.
2.  Provider code isn't expected to have to parse the parameter value, such as all the `atoi()` calls we see.  The only place where the provider code still needs to fend for itself is with control names such av "rsa_pss_saltlen", where *our implementation* expects certain key words or a number, but other than that, `OSSL_PARAM_allocate_from_text()` does everything it can to parse the value into something that the provider can handle, according to what it declares.
3.  Return values.  Now...  here's something that I did gloss over a bit in this PR (and so did everyone that reviewed), and I agree that `OSSL_PARAM_allocate_from_text()` should be able to indicate to the caller that a parameter wasn't recognised, letting the caller act appropriately on that, such as having `legacy_ctrl_str_to_params()` return -2.

Now...  to your analysis of return codes:

> - `-2` if this `EVP_PKEY_CTX` is not for a `DERIVE` operation
>    * does not concern the provider implementation
>    * `libcrypto` should return `-2` to the caller if this condition is detected

I don't understand why `EVP_PKEY_CTX_ctrl_str()` or `legacy_ctrl_str_to_params()` should check what operation is currently handled, apart from what `EVP_PKEY_CTX_ctrl_str()` already does to differentiate calls that should go to providers from calls that  should go to our legacy backends.  *The method pointed at by `ctx->pmeth` (in the legacy case) or others (in the provider case, `EVP_PKEY_CTX_set_params()` takes care of ensuring that the correct one is used) should be trusted*.  This was true in pre-3.0 `EVP_PKEY_CTX_ctrLstr()`, it should contrinue to be true going forward.

> - `-1` if this `EVP_PKEY_CTX` does not contain an `EVP_PKEY_EC` key
>    * does not concern the provider implementation
>    * `libcrypto` should return `-1` to the caller if this condition is detected

Er....  If the backend code to handle an `EVP_PKEY_EC` key (or corresponding keytype in the provider case) is being called by `EVP_PKEY_CTX_ctrl_str()` with a key that's not an EC key, then something else is deeply wrong, and this isn't within `EVP_PKEY_CTX_ctrl_str()`'s control.

> - `-2` if this `EVP_PKEY_CTX` does not support setting the `kdf_md` or the `cofactor_mode`

Now, here you do have a point, and that gets back to what I said in 3. above, that `legacy_ctrl_str_to_params()` needs to be able to return -2 when it can detect that a specific parameter name isn't supported by the provider code.  I agree with this concern.  Fortunately, `OSSL_PARAM_allocate_from_text()` is a new function, we can change what it returns for this sort of case.

>     * for the former case we are also supposed to raise `EC_R_INVALID_DIGEST` if the string value cannot be resolved to a valid `EVP_MD *`

If you look at other code that we have moved from diverse built in methods to our providers, we haven't concerned ourselves with preserving exact error codes.  They are all over the place regardless.

>    * the latter is particularly relevant because we historically allowed for underdefined curves that do not specify the cofactor, as it is considered optional in the serialization.

This is again about -2, right?  I hope I have answered that clearly enough.

>      For such curves, the cofactor used to be set to `0` and with b783beeadf6b80bc431e6f3230b5d5585c87ef87 we are now attempting to compute it, but there might be curves for which the computation fails to deliver the cofactor and so ecdh_cofactor_mode cannot be turned on!

Er...  how is this a concern specifically for control values (legacy) or parameters (provider) being passed to the backend via `EVP_PKEY_CTX_ctrl_str()`?  This sounds much more like a key creation thing.

romen:
> > I see now another problem with this: by going directly to the provider interface with `EVP_PKEY_CTX_set_params()` from inside `legacy_ctrl_str_to_param()` we are bypassing the dedicated `EVP_PKEY_CTX_set_foo()`.
> 
> Er, yes?

It's not that I don't like the idea of doing it like this, it's just that I am afraid we are breaking backward compatibility: before `EVP_PKEY_CTX_ctrl_str()` was basically a string based front end for some of the `EVP_PKEY_CTX_ctrl()` options, now we are skipping altogether that layer, and there is a risk of pushing _legacy_ hacks onto the providers if they want to preserve backward compatibility.


> 
> The EVP API per se doesn't really know, and _shouldn't_ really know what the backends are up to, and this is very visible when you look at how the pre-3.0 `EVP_PKEY_CTX_ctrl_str()` is implemented (this is from 1.1.1):
> 
> https://github.com/openssl/openssl/blob/ff6d0a6313a9082965f798a3102539b8d83a6a39/crypto/evp/pmeth_lib.c#L407-L418
> 

This is not the whole picture though, as the code I linked before (also from 1.1.1) shows
> https://github.com/openssl/openssl/blob/894da2fb7ed5d314ee5c2fc9fd2d9b8b74111596/crypto/ec/ec_pmeth.c#L354-L392
> 

`EVP_PKEY_CTX_ctrl_str()` calls into the backed, which does parsing and minimal checks before calling `EVP_PKEY_CTX_ctrl()`, which does a lot of checks before eventually calling back to `pmeth->ctrl` implementation for that backend.
In turn the `pmeth->ctrl` implementation also had to adhere to the conventions set by the EVP API about behaviour of `EVP_PKEY_CTX_ctrl()` so that it was indeed possible to use this kind of _controls_ more or less uniformly across different backends.

> > we force every external provider to implement whatever workarounds we devise inside the providers we distribute to retain the legacy behavior.
> 
> Do we? Would you claim that any ENGINE implementor that wanted to implement their version of, say, RSA, was forced to implement the exact behaviour the our backends (crypto/rsa/rsa_ameth.c or crypto/rsa/rsa_pmeth.c)? To what extent? 'cause believe you me, that wasn't necessarily the case...

ENGINEs imposed their own restraints on external implementers, and returning the same return values and error codes for at least a subset of these _controls_ was indeed one of the restraints, but you are definitely more knowledgeable than me about ENGINEs so I probably am missing your point.

Here what I see as a problem is that if to behave like legacy a specific `EVP_PKEY_CTX_ctrl_str()` needs to do something shady like returning `-1` for some errors, `0` for others and `-2` for some more, by calling directly into `EVP_PKEY_CTX_set_params()` instead of routing through the dedicated `EVP_PKEY_CTX_set_foo()` we are pushing the task of being compatible with legacy onto the provider-side handling of `OSSL_PARAMS`, which I don't think is what we want!

> The very simple truth is central EVP code such as `EVP_PKEY_CTX_ctrl_str()` _cannot possibly know what a backend will or will not accept_, the design is to simply pass on what it gets, in both directions (except for, weirdly enough, "digest", in the pre-3.0 code... that one confuses me a bit).
> 

This is fine, as long the complexity of being backward compatible with legacy `libcrypto` behavior is not pushed onto the provider implementation.
Any "we used to do this in this case, so we have to add checks and try to reproduce as closely as possible what we did before" logic pertains inside the `libcrypto` boundary and not in the `provider`.
It can't be on `provider` side, otherwise we are pushing these workarounds onto every external provider implementation, and it becomes a hard requirement if the external implementer wants to avoid breaking legacy applications when the user loads their provider.

It is fine to resolve that being backward compatible with return values or error codes is just nit picking and deliberate that we will break our promise in this niche aspect, but it should still be discussed (and probably OMC voted).

> All this being said, you're quite right that there are things that aren't the same between the old (ameth and pmeth code) and the new (providers) backends, and that can be summed up in three parts:
> 
> 1. The control names aren't necessarily the same as the parameter names [...]

Agreed!

> 2. Provider code isn't expected to have to parse the parameter value, such as all the `atoi()` calls we see. [...]

Agreed!

> 3. Return values. [...]

Agreed!

> Now... to your analysis of return codes:
> 
> > * ~~`-2`~~ `-1` if this `EVP_PKEY_CTX` is not for a `DERIVE` operation
> >   
> >   * does not concern the provider implementation
> >   * `libcrypto` should return ~~`-2`~~ `-1` to the caller if this condition is detected
> 
> I don't understand why `EVP_PKEY_CTX_ctrl_str()` or `legacy_ctrl_str_to_params()` should check what operation is currently handled, apart from what `EVP_PKEY_CTX_ctrl_str()` already does to differentiate calls that should go to providers from calls that should go to our legacy backends. _The method pointed at by `ctx->pmeth` (in the legacy case) or others (in the provider case, `EVP_PKEY_CTX_set_params()` takes care of ensuring that the correct one is used) should be trusted_. This was true in pre-3.0 `EVP_PKEY_CTX_ctrLstr()`, it should contrinue to be true going forward.

Let's dive into the handling of "ecdh_cofactor_mode" from the point of view of a 1.1.1 application calling `EVP_PKEY_CTX_ctrl_str()` and evaluate which layer checks what and what values are returned and error codes raised for the cases where something fails.

https://github.com/openssl/openssl/blob/894da2fb7ed5d314ee5c2fc9fd2d9b8b74111596/crypto/ec/ec_pmeth.c#L385-L389

https://github.com/openssl/openssl/blob/894da2fb7ed5d314ee5c2fc9fd2d9b8b74111596/include/openssl/ec.h#L1387-L1390

https://github.com/openssl/openssl/blob/894da2fb7ed5d314ee5c2fc9fd2d9b8b74111596/crypto/evp/pmeth_lib.c#L366-L399

(shows 3 different conditions under which `EVP_PKEY_CTX_ctrl_str()` could return `-1`, each with specific error codes, and other 2 possible cases for returning `-2`)

https://github.com/openssl/openssl/blob/894da2fb7ed5d314ee5c2fc9fd2d9b8b74111596/crypto/ec/ec_pmeth.c#L247-L278

(return `-2` if `p1` out of range, or if the `group` is missing)

So https://github.com/openssl/openssl/blob/894da2fb7ed5d314ee5c2fc9fd2d9b8b74111596/crypto/evp/pmeth_lib.c#L387-L390

should answer to the "why `legacy_ctrl_str_to_param()` or  `EVP_PKEY_CTX_ctrl_str()` should check what operation is currently handled": because pre-3.0 was doing so and signaling with specific error codes and return value.
 
> 
> > * `-1` if this `EVP_PKEY_CTX` does not contain an `EVP_PKEY_EC` key
> >   
> >   * does not concern the provider implementation
> >   * `libcrypto` should return `-1` to the caller if this condition is detected
> 
> Er.... If the backend code to handle an `EVP_PKEY_EC` key (or corresponding keytype in the provider case) is being called by `EVP_PKEY_CTX_ctrl_str()` with a key that's not an EC key, then something else is deeply wrong, and this isn't within `EVP_PKEY_CTX_ctrl_str()`'s control.

Agreed that if this happens there are probably bigger problems: before 3.0 we used to handle these cases in `EVP_PKEY_CTX_ctrl_str()` and to signal with specific return values and error codes.
Now we don't anymore, and it is a potentially breaking change, which at least requires discussion, probably a vote, and an entry in the document that we will probably write for alpha1 to invite external developers to test their applications, engines, providers against 3.0 and pay attention to some details that we deliberately modified in the transition to 3.0.


> 
> > * `-2` if this `EVP_PKEY_CTX` does not support setting the `kdf_md` or the `cofactor_mode`
> 
> Now, here you do have a point, and that gets back to what I said in 3. above, that `legacy_ctrl_str_to_params()` needs to be able to return -2 when it can detect that a specific parameter name isn't supported by the provider code. I agree with this concern. Fortunately, `OSSL_PARAM_allocate_from_text()` is a new function, we can change what it returns for this sort of case.
> 

I agree this should be manageable without too much fuss!

> > * for the former case we are also supposed to raise `EC_R_INVALID_DIGEST` if the string value cannot be resolved to a valid `EVP_MD *`
> 
> If you look at other code that we have moved from diverse built in methods to our providers, we haven't concerned ourselves with preserving exact error codes. They are all over the place regardless.
> 

I noticed this, as well as the fact that in the DH provider implementation we have TODOs to raise errors.
What I wasn't aware of is the fact that we deliberately plan to break compatibility about error codes (we don't guarantee we check for all the corner cases we were checking before, we might change the reason _name_, same reason might have a different numeric value).
Was there a discussion about this that I missed?


> > * the latter is particularly relevant because we historically allowed for underdefined curves that do not specify the cofactor, as it is considered optional in the serialization.
> 
> This is again about -2, right? I hope I have answered that clearly enough.
> 

Yes, the `-2` on not supported by provider is clear enough.

> > For such curves, the cofactor used to be [...]
> 
> Er... how is this a concern specifically for control values (legacy) or parameters (provider) being passed to the backend via `EVP_PKEY_CTX_ctrl_str()`? This sounds much more like a key creation thing.

This was just meant to be an example of a case where it was and is reasonable that applications checked for that `-2` and to the error stack to handle the case of "we wished to use cofactor ecdh, it was not supported, but maybe it is one of those special alien curves that need special care to turn cofactor mode on".

levitte:
> So https://github.com/openssl/openssl/blob/894da2fb7ed5d314ee5c2fc9fd2d9b8b74111596/crypto/evp/pmeth_lib.c#L387-L390
> 
> should answer to the "why `legacy_ctrl_str_to_param()` or  `EVP_PKEY_CTX_ctrl_str()` should check what operation is currently handled": because pre-3.0 was doing so and signaling with specific error codes and return value.

It does not.  There's a huge structural difference that you've ignored.  In legacy code `EVP_PKEY_METHOD` is one big method that handles all sorts of operations using asymmetric keys, while we have split them up in provider code, so we have `EVP_SIGNATURE`, `EVP_ASYM_CIPHER`, `EVP_KEYEXCH` et al (will there be others?  I've lost track), so the check already happens at a higher level before provider code is even reached.  Therefore, *the provider code does not need to make that kind of check*.

What I would expect, the day that legacy code goes away, is that this:

https://github.com/openssl/openssl/blob/87d3bb8e861e54e2549e848a1adee6d2df9ed603/crypto/evp/pmeth_lib.c#L860-L867

gets replace with this:

``` C
     ERR_raise(ERR_LIB_EVP, EVP_R_INVALID_OPERATION); 
     return -1;
```

(or quite frankly, I hope that everything called `_ctrl` is gone by then, because it's considered legacy)

but since we *currently* have a mixture of providers and legacy backends, we simply let legacy take care of whatever there's no provider code.

-----

> (return -2 if `p1` out of range, or if the group is missing)

I actually think that returning -2 in such cases is wrong.  0 should be returned, as it is an error.

The manual for `EVP_PKEY_CTX_ctrl()` sheds a bit of clarity when -2 can be expected (italics mine, for clarification):

>        The function EVP_PKEY_CTX_ctrl() sends _a control operation_ to the
>        context ctx. The key type used must match keytype if it is not -1.
> ...
> **RETURN VALUES**
>        EVP_PKEY_CTX_ctrl() and its macros return a positive value for success
>        and 0 or a negative value for failure. In particular a return value of
>        -2 indicates _the operation is not supported_ by the public key
>        algorithm.

"the operation", here, refers to the *cmd*, i.e. things like `EVP_PKEY_CTRL_EC_ECDH_COFACTOR`.  Why someone thought it was a good idea to use the same return value because the *value* sent in with that *cmd* is beyond me.

levitte:
As for the diverse `EVP_PKEY_CTX_set_whatever()` that you see in the legacy backends, I assume you understand that they aren't expected to be reproduced in provider code.

romen:
Regarding the first half of https://github.com/openssl/openssl/pull/10947#issuecomment-583834618

I am not ignoring the fact that the architectural reorganization makes the check likely superfluous, I am just concerned that by skipping it (because it is superfluous and hopefully triggered before calling any _ctrl_ function) we might be altering the return value and error codes raised in these corner cases.

I am likely to be overthinking this, and probably you are completely right in assuming that these checks even if present would never trigger the special return value and raised error, but without extensive testing we have no way of guaranteeing this is the case!

---

> I actually think that returning -2 in such cases is wrong. 0 should be returned, as it is an error.

I tend to agree with you, and can't wait for the moment we will drop the _ctrl_ functions, but even if it is an error, it is an ossified behavior from at least 1.0.2, that should require a discussion before being changed!

levitte:
@romen, you should check #11046, I hope that it answers some of your concerns

levitte:
> I tend to agree with you, and can't wait for the moment we will drop the ctrl functions, but even if it is an error, it is an ossified behavior from at least 1.0.2, that should require a discussion before being changed!

If we *really* want to keep (wrongly) returning -2 for incorrect parameter values, then I can only see one answer, that the providers will have to return -2 in those cases.  Trying to do some magic in the EVP code will only create a total mess, of which we already saw the effect before this PR.

Basically, *the central EVP code **should not assume knowledge that belong in the backend***, and starting to mess with all kinds of values in the central code is diametrically opposite to that principle.  On this, I will stand firm.

romen:
> If we *really* want to keep (wrongly) returning -2 for incorrect parameter values, then I can only see one answer, that the providers will have to return -2 in those cases.

If we (or OMC) doesn't vote to approve breaking legacy compatibility we should figure out where all that ugliness would need to go.

I would argue that, with libcrypto becoming essentially a framework, any temporary workaround to maintain our promise of backward compatibility must be handled inside the framework code, no matter how ugly, and can't really be forced on external providers!

Maybe some other reviewer may chime in: does this really constitute breaking behavior? Do we need an OMC vote to get rid of the ugliness? 

I will look at #11046 as soon as possible!

PR #10144, opened 2019-10-10
WIP: Infrastructure for fetchable RAND.

paulidale:
This adds the infrastructure for a RAND algorithm type.  These will include the DRBGs and seeding sources.

- [ ] documentation is added or updated
- [ ] tests are added or updated

PR #10950, opened 2020-01-26
RAND_DRBG: add a callback data pointer for obtaining entropy and nonce.

mspncp:
This is a specialized feature which is used for example to implement
deterministic signatures according to RFC6979, which requires passing
a specified entropy and nonce for instantiating the DRBG.

See #9223 

(This is a private api intended to be used only in libcrypto)

### DIFF
--- a/crypto/rand/drbg_lib.c
+++ b/crypto/rand/drbg_lib.c
@@ -305,6 +305,34 @@ void rand_drbg_cleanup_nonce(RAND_DRBG *drbg,
 }
 
 /*
+ * Set the |drbg|'s callback data pointer for the entropy and nonce callbacks
+ *
+ * The ownership of the context data remains with the caller,
+ * i.e., it is the caller's responsibility to keep it available as long
+ * as it is need by the callbacks and free it after use.
+ *
+ * Setting the callback data is allowed only if the drbg has not been
+ * initialized yet. Otherwise, the operation will fail.
+ *
+ * Returns 1 on success, 0 on failure.
+ */
+int RAND_DRBG_set_callback_data(RAND_DRBG *drbg, void *data)
+{
+    if (drbg->state != DRBG_UNINITIALISED
+        || drbg->parent != NULL)
+        return 0;
+
+    drbg->callback_data = data;
+    return 1;
+}
+
+/* Retrieve the callback data pointer */
+void *RAND_DRBG_get_callback_data(RAND_DRBG *drbg)
+{
+    return drbg->callback_data;
+}
+
+/*
  * Set/initialize |drbg| to be of type |type|, with optional |flags|.
  *
  * If |type| and |flags| are zero, use the defaults
--- a/doc/man3/RAND_DRBG_set_callbacks.pod
+++ b/doc/man3/RAND_DRBG_set_callbacks.pod
@@ -62,7 +62,7 @@ RAND_DRBG_set_callback_data() can be used to store a pointer to some context
 specific data, which can subsequently be retrieved by the entropy and nonce
 callbacks using RAND_DRBG_get_callback_data().
 The ownership of the context data remains with the caller, i.e., it is the
-caller's responsibility to keep it available as long as it is need by the
+caller's responsibility to keep it available as long as it is needed by the
 callbacks and free it after use.
 For more information about the the callback data see the NOTES section.
 
@@ -140,7 +140,7 @@ In this case the DRBG will automatically request an extra amount of entropy
 utilize for the nonce, following the recommendations of [NIST SP 800-90A Rev. 1],
 section 8.6.7.
 
-The callback data a rather specialized feature, because in general the
+The callback data is a rather specialized feature, because in general the
 random sources don't (and in fact, they must not) depend on any state provided
 by the DRBG.
 There are however exceptional cases where this feature is useful, most notably
--- a/doc/man3/RAND_DRBG_set_callbacks.pod
+++ b/doc/man3/RAND_DRBG_set_callbacks.pod
@@ -3,6 +3,8 @@
 =head1 NAME
 
 RAND_DRBG_set_callbacks,
+RAND_DRBG_set_callback_data,
+RAND_DRBG_get_callback_data,
 RAND_DRBG_get_entropy_fn,
 RAND_DRBG_cleanup_entropy_fn,
 RAND_DRBG_get_nonce_fn,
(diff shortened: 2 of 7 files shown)


kaduk:
There are some fault-tolerance considerations applicable to purely deterministic signatures as discussed in the thread at https://mailarchive.ietf.org/arch/msg/cfrg/9yqTJuwxeSU3OCjbyyJqAJ-xTy0/; is there a specific reason that exactly RFC 6979 needs to be implemented as opposed to something that includes additional randomness as well as the normal deterministic procedure?  (I note that it is not a product of the IETF or IRTF CFRG, and is instead an RFC on the "Independent Submission" stream that is wholly independent of the others.)
I only skimmed #9223 so my apologies if I missed previous discussion there.

mspncp:
> I only skimmed #9223 so my apologies if I missed previous discussion there.

That would not surprise me if you had missed something: The discussion thread of pull request #9223 has become incredibly large, with a lot of collapsed sub-discussions, so it is hard to reconstruct all the details.

Regarding your questions: I think it is a good thing to have this discussion, but since this pull request her is actually only paving the way for #9223, it not quite the right location to do it. I think it would be a good thing if you could repost your question as a separate issue referring to #9223.

crypto/rand/drbg_lib.c
```diff
@@ -304,6 +304,41 @@ void rand_drbg_cleanup_nonce(RAND_DRBG *drbg,
...
+
+    drbg->callback_data = data;
+    return 1;
+}
+
+/* Retrieve the pointer to the callback data */
+void *rand_drbg_get_callback_data(RAND_DRBG *drbg)
+{
+    return drbg->callback_data;
+}
```
mspncp: I am pondering whether I should make those two functions public: They are a useful supplement and enhancement of the callback mechanism. It could not only be used for deterministic signatures, but also for the known answer tests, which currently utilize the ex_data mechanism. The new API would simplify the implementation, because it does not require a RUN_ONCE initialization for the ex_data index.

```C
    int RAND_DRBG_set_callbacks(RAND_DRBG *drbg,
                                RAND_DRBG_get_entropy_fn get_entropy,
                                RAND_DRBG_cleanup_entropy_fn cleanup_entropy,
                                RAND_DRBG_get_nonce_fn get_nonce,
                                RAND_DRBG_cleanup_nonce_fn cleanup_nonce);

    int RAND_DRBG_set_callback_data(RAND_DRBG *drbg, void *data);
    void *RAND_DRBG_get_callback_data(RAND_DRBG *drbg);
```
@paulidale @slontis what do you think about my suggestion?

paulidale: I was looking at a different approach to this problem as part of #10144: the introduction of a test DRBG that output fixed data.  A DRBG could use this as its parent.  The context was for the FIPS known answer tests.

Unfortunately, #10144 seems like it is being too ambitious on other fronts.  It tried to split the DRBGs into legacy APIs, an EVP_RAND layer and implementations in providers.  It also tried to allow either an entropy source or a DRBG to be the parent of a DRBG.

As for making these public, I don't have a strong feeling either way.  We will need fetchable RANDs for FIPS, so not making that more difficult is a consideration but I don't see any (more) hurt happening because of this.

I'm not sure if it is relevant, but `OSSL_CALLBACK` includes an extra argument which could be used for this purpose.  They are a bit lower level though.

slontis: The ex_data has been ripped out of FIPS_MODE so a mechanism using OSSL_CALLBACK would be a good idea.

mspncp:
Ok, I‘ll make it public then. Putting it into WIP until documentation has been added.

mspncp:
Still WIP. The KATs are still to be revised to use the new API.

doc/man3/RAND_DRBG_set_callbacks.pod
```diff
@@ -53,7 +58,16 @@ the nonce when reseeding the given B<drbg>.
 The callback functions are implemented and provided by the caller.
 Their parameter lists need to match the function prototypes above.
 
-Setting the callbacks is allowed only if the DRBG has not been initialized yet.
+RAND_DRBG_set_callback_data() can be used to store a pointer to some context
+specific data, which can subsequently be retrieved by the entropy and nonce
+callbacks using RAND_DRBG_get_callback_data().
+The ownership of the context data remains with the caller, i.e., it is the
+caller's responsibility to keep it available as long as it is need by the
```
slontis: need(ed)

doc/man3/RAND_DRBG_set_callbacks.pod
```diff
@@ -121,6 +140,14 @@ In this case the DRBG will automatically request an extra amount of entropy
 utilize for the nonce, following the recommendations of [NIST SP 800-90A Rev. 1],
 section 8.6.7.
 
+The callback data a rather specialized feature, because in general the
```
slontis: is a

mspncp:
I revised the drbgtest in commit 28dad0d3da02d47e209a93276992f9520d9f5863. @slontis, @paulidale, please take a closer look, in particular at the test_rand_drbg_reseed test.

    tests/drbgtest: use new RAND_DRBG callback_data API instead of ex_data
    
    It took me a little while to realize why the test_rand_drbg_reseed test
    kept crashing after replacing the RAND_DRBG_{gs}et_ex_data() calls by
    RAND_DRBG_{gs}et_callback_data().
    
    The reason was that the ex_data API prohibits modifying the callbacks
    or callback data of chained DRBGs and returned an error which was
    ignored by the `test_rand_drbg_reseed` test, for good reasons.
    
    The `test_rand_drbg_reseed` test is special in this respect, because
    it needs to install callbacks for all DRBGs, in order to intercept
    and count the reseeding events.
    
    Since the drbgtest module has access to the internal structures of
    the DRBG anyway, the problem could be solved by accessing the members
    directly. I added a warning comment in hook_drbg().

I ran the test several times with `OPENSSL_TEST_RAND_ORDER=yes`, to ensure that the tests also succeed in random order.

mspncp:
Rebased to resolve the conflict in libcrypto.num.

slontis:
Apart from a few NITS in the doc and a rebase it looks good to me.

mspncp:
> Apart from a few NITS in the doc and a rebase it looks good to me.

Thanks for the reminder, fixup in 2ff290a71c598587ff3bfdb117eed61c52823ad9.

slontis:
Probably change title to not WIP if you are happy with it.. just needs a rebase of the generated file now..

mspncp:
Rebased without further changes to fix the libcrypto.num conflict.

mspncp:
Merged, thanks for the review.

30a9d5d1a7 RAND_DRBG: add a callback data for entropy and nonce callbacks
09066cf2a1 tests/drbgtest: use new RAND_DRBG callback_data API instead of ex_data

PR #10952, opened 2020-01-27
Add pairwise consistency self tests to asym keygenerators

slontis:
##### Checklist


slontis:
This is dependent on keygen being added to providers for different key types.

crypto/rsa/rsa_gen.c
```diff
@@ -66,9 +67,19 @@ static int rsa_builtin_keygen(RSA *rsa, int bits, int primes, BIGNUM *e_value,
...
+
     if (primes != 2)
         return 0;
-    return rsa_sp800_56b_generate_key(rsa, bits, e_value, cb);
+    ret = rsa_sp800_56b_generate_key(rsa, bits, e_value, cb);
+    if (ret) {
+        /* TODO(3.0) - needs libctx */
+        OSSL_SELF_TEST_get_callback(NULL, &cb, &cbarg);
+        ret = SELF_TEST_keygen_pairwise_test_rsa(rsa, cb, cbarg);
+    }
```
levitte: This function already takes a BN_GENCB, why not use it?

crypto/ec/ec_key.c
```diff
@@ -280,7 +281,17 @@ int ec_key_simple_generate_key(EC_KEY *eckey)
...
+    {
+        OSSL_CALLBACK *cb = NULL;
+        void *cbarg = NULL;
+
+        /* TODO(3.0) - needs libctx */
+        OSSL_SELF_TEST_get_callback(NULL, &cb, &cbarg);
+        if (!SELF_TEST_keygen_pairwise_test_ecdsa(eckey, cb, cbarg))
+            goto err;
+    }
+#endif /* FIPS_MODE */
```
levitte: Isn't there a function that takes a BN_GENCB somewhere?

slontis: 1) Its not even vaguely related to BN.
2) Its a self test - so it should be handled in a similiar manner to all the other self test events thru the same callback mechanism.

crypto/rsa/rsa_gen.c
slontis: see below

crypto/ec/ec_key.c
levitte: To be blunt, it looks like hacks thrown in a little here and there rather than something done by design.  That's what I'm reacting against.  It's a little like having a global variable that gets used everywhere, and a few years ahead, you start wondering where the hell it's for, getting surprised by it being defined "over there", etc.

slontis: re you talking about this code, or OPENSSL_CTX ? :)

slontis: The requirement is that I want it to use the self test callback. Maybe if I shifted the code into the gen you would be happier..

levitte: I'm talking about the SELF_TEST calls.  They reside in a module outside of libcrypto, while this is primarly libcrypto code.

True, it would make more sense to have these in the gen code proper, but most of all, it would make even more sense to pass down the callback and its extra argument to these functions than having them grab for it because the code happens to be part of the FIPS module.

Part of my concern is also that this kind of hack effectively excludes anyone else from using this kind of callback.  I have no doubt that other provider authors may want to include this kind of self test, especially if they create providers around something FIPS-able.

slontis: This means the keygen function is going to need to pass in the callback into the keymanagement before it calls into the provider to do the actual keygen operation.. (I cant quite do that yet - but it will just be a moving the code about exercise). I dont think there is anything wrong with using the self test callback - I just need to move the code.

mattcaswell: > Part of my concern is also that this kind of hack effectively excludes anyone else from using this kind of callback. I have no doubt that other provider authors may want to include this kind of self test, especially if they create providers around something FIPS-able.

I don't understand your concern at all. Why is this excluding anyone?

levitte: > Why is this excluding anyone?

...
...  I may have over-inflated the issue.

So, er, never mind.

providers/fips/self_test_keygen_pairwise.c
```diff
@@ -0,0 +1,138 @@
+/*
+ * Copyright 2020 The OpenSSL Project Authors. All Rights Reserved.
+ *
+ * Licensed under the OpenSSL license (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
+ */
+
+#include <string.h> /* memcmp */
```
levitte: Because you're using publicaly deprecated low level functions, you need to add this above all other includes:

``` C
/*
 * RSA, ECDSA and DSA low level APIs are deprecated for public use, but still ok for
 * internal use.
 */
```

slontis:
Took this out of WIP.
After taking into account what @levitte has said..
I have move the self test methods into the self_test_core instead of being in the fips module and made them public.
The pairwise tests have also been moved into libcrypto so they can be used by any provider.
A pairwise test flag that should be able to be set by the keygen when it is done.
This flag will be enabled by default for FIPS_MODE.

slontis:
Put back to WIP: some other PR's need merging before this is rebased.

doc/man3/OSSL_SELF_TEST_new.pod
```diff
@@ -0,0 +1,113 @@
...
+=head1 DESCRIPTION
+
+During startup a user may want to run some self tests with known answers on
+algorithms to test that the system is operating correctly.
+Each self test that runs can use these methods to indicate
+the running status of a self test as it runs.
+
+OSSL_SELF_TEST_new() allocates an opaque B<OSSL_SELF_TEST> object that has a
+callback and callback argument associated with it. The callback is triggered by
+the 3 phases of the self test that can be setup.
```
levitte: Is the application expected to use this? A provider?
It might be worth a moment writing a few lines on expected usage.

slontis: updated the doc.

doc/man3/OSSL_SELF_TEST_new.pod
```diff
@@ -0,0 +1,177 @@
...
+    OSSL_CALLBACK *cb;
+    void *cbarg;
+    int ok = 0;
+    unsigned char out[EVP_MAX_MD_SIZE];
+    unsigned int out_len = 0;
+    EVP_MD_CTX *ctx = EVP_MD_CTX_new();
+    EVP_MD *md = EVP_MD_fetch(libctx, t->algorithm, NULL);
+
+    /*
+     * Retrieve the callback - will be NULL is not set by the application via
```
romen: Nit: is ->if

doc/man3/OSSL_SELF_TEST_new.pod
```diff
@@ -0,0 +1,177 @@
...
+                             const char *desc);
+ void OSSL_SELF_TEST_oncorrupt_byte(OSSL_SELF_TEST *st, unsigned char *bytes);
+ void OSSL_SELF_TEST_onend(OSSL_SELF_TEST *st, int ret);
+
+=head1 DESCRIPTION
+
+These methods are intended for use by provider implementors, to display
+diagnostic information during self testing.
+
+There are 3 types of self tests that can be set up:
```
levitte: The strings that are described don't correspond to anything in code.  Why are they presented as strings?  How does one know what type of self test is active in a certain piece of code?

doc/man3/OSSL_SELF_TEST_new.pod
```diff
@@ -0,0 +1,177 @@
...
+Used during an operation on an algorithm e.g. when performing a pairwise test
+during key generation of asymmetric keys.
+
+=back
+
+OSSL_SELF_TEST_new() allocates an opaque B<OSSL_SELF_TEST> object that has a
+callback and callback argument associated with it. The callback is triggered by
+the 3 phases of the self test that can be setup.
+
+The following OSSL_PARAMS are passed to the callback during each phase:
```
levitte: The description is specific to our FIPS provider, why was it moved from `doc/man7/OSSL_PROVIDER-FIPS.pod`?

doc/man3/OSSL_SELF_TEST_new.pod
```diff
@@ -0,0 +1,177 @@
...
+=item "Fail" (B<OSSL_SELF_TEST_PHASE_FAIL>) <UTF8 string>
+
+The value "Fail" is sent by OSSL_SELF_TEST_onend() if the B<ret> value is 0,
+otherwise it sends the value "Pass".
+
+=back
+
+=item "st-type" (B<OSSL_PROV_PARAM_SELF_TEST_TYPE>) <UTF8 string>
+
+The value is setup by the B<type> passed to OSSL_SELF_TEST_onbegin().
```
levitte: Use `I` for variables, i.e. `I<type>` in this case.  Bold is not conformant to man-page standards.

doc/man3/OSSL_SELF_TEST_new.pod
```diff
@@ -0,0 +1,177 @@
...
+This allows the callback to identify the type of test being run,
+
+=item "st-desc" (B<OSSL_PROV_PARAM_SELF_TEST_DESC>) <UTF8 string>
+
+The value is setup by the B<type> passed to OSSL_SELF_TEST_onbegin().
```
levitte: `I<type>`

doc/man3/OSSL_SELF_TEST_new.pod
```diff
@@ -0,0 +1,177 @@
...
+This allows the callback to identify the sub category of the test being run,
+
+=back
+
+OSSL_SELF_TEST_onbegin() is inserted at the start on the self test code. It can
+be used for diagnostic purposes.
+
+OSSL_SELF_TEST_oncorrupt_byte() is inserted just after the known answer is
+calculated, but before the self test compares the result. The first byte in the
+passed in array of B<bytes> will be corrupted if the callback returns 0,
```
levitte: `I<bytes>`

doc/man3/OSSL_SELF_TEST_new.pod
```diff
@@ -0,0 +1,177 @@
...
+otherwise it leaves the array unaltered. It can be used for failure testing.
+The B<type> and B<desc> can be used to identify an individual self test to
```
levitte: `I<type>` and `I<desc>`

doc/man3/OSSL_SELF_TEST_new.pod
slontis: Because the pairwise consistency tests in keygen use these.
The keygen self tests can be activated in non fips mode now, so they are not just fips.
In a future PR, I will probably move most self test code out of the fips module so that it can be reused by other providers. If they use it, it will use the given events. If they dont like it they can write their own.

doc/man3/OSSL_SELF_TEST_new.pod
slontis: Removed the descriptions. I was trying to be helpful - guess I failed.

doc/man3/OSSL_SELF_TEST_new.pod
levitte: Correction: Because *our implementation of* the pairwise consistency tests in keygen use these.

In other words, this is information about *our implementation* that is now inserted into the general reference documentation.  Other implementations could use whatever strings they want.

levitte: Since this is information about *out implementation*, I would say it doesn't belong here.

slontis: The 3 SELF_TEST_onXXX are visible here.. SO it does belong here.

doc/man3/OSSL_SELF_TEST_new.pod
slontis: Is B<OSSL_PROV_PARAM_SELF_TEST_DESC> alright? or should everything be using I<

doc/man3/OSSL_SELF_TEST_new.pod
levitte: Argh...

Another implementation can use whatever strings they want, right?  They have no obligation whatsoever to use the strings you mention here in *their* calls to SELF_TEST_onXXX, do they?

I don't quite understand why it's so hard to understand that the description of an API and the description of *implementation specific strings* to that API are two different things?

levitte: ("implementation" here actually means "provider implementation")

slontis: Will discuss with you tonight - you dont understand what I am saying either :)

slontis: I have reworded it so hopefully we are on the same (man)page now. If not - tell me off tonight..

slontis:
rebased to fix generated file.

doc/man3/OSSL_SELF_TEST_new.pod
richsalz: numbers under 10 get written out as words, hence "three"

doc/man3/OSSL_SELF_TEST_new.pod
```diff
@@ -0,0 +1,169 @@
...
+If OSSL_SELF_TEST_onbegin(), OSSL_SELF_TEST_oncorrupt_byte() or
+OSSL_SELF_TEST_onend() is called the following additional B<OSSL_PARAM> are
+passed to the callback.
+
+=over 4
+
+=item "st-type" (B<OSSL_PROV_PARAM_SELF_TEST_TYPE>) <UTF8 string>
+
+The value is setup by the I<type> passed to OSSL_SELF_TEST_onbegin().
+This allows the callback to identify the type of test being run.
```
levitte: You might want to add a note that an `OSSL_SELF_TEST_onend()` call will set the type to "None"

doc/man3/OSSL_SELF_TEST_new.pod
```diff
@@ -0,0 +1,169 @@
...
+
+=item "st-desc" (B<OSSL_PROV_PARAM_SELF_TEST_DESC>) <UTF8 string>
+
+The value is setup by the I<type> passed to OSSL_SELF_TEST_onbegin().
+This allows the callback to identify the sub category of the test being run.
```
levitte: Likewise here

doc/man3/OSSL_SELF_TEST_new.pod
slontis: Removed the docs containing the number 3 :)

doc/man3/OSSL_SELF_TEST_new.pod
slontis: I just put a comment in the onend() section which should suffice.

slontis:
Had to rebase..

doc/man3/OSSL_SELF_TEST_new.pod
```diff
@@ -0,0 +1,172 @@
...
+
+=item "st-phase" (B<OSSL_PROV_PARAM_SELF_TEST_PHASE>) <UTF8 string>
+
+The value of the string is "Pass" if I<ret> is non zero, otherwise it has the
+value "Fail".
+
+=back
+
+After the callback I<cb> has been called the values that were set by
+OSSL_SELF_TEST_onbegin for I<type> and I<desc> are set to the value "None".
```
levitte: Nit: `OSSL_SELF_TEST_onbegin()`

slontis: Updated

levitte:
Argh, forgot labels...  @slontis, would you mind rebasing?  If it's clean and CIs don't have a fit, my approval stands.

slontis:
>  If it's clean and CIs don't have a fit, my approval stands.

Not quite clean, another PR has been merged since then related to self tests that needed modifying.. See d4c9f41 above.

slontis:
ping

slontis:
@levitte can you re-review this please

slontis:
rebased and collapsed the older commits that were not affected by the initial rebase.

crypto/dsa/dsa_key.c
```diff
@@ -82,14 +84,74 @@ static int dsa_builtin_keygen(DSA *dsa)
...
         BN_free(priv_key);
     BN_CTX_free(ctx);
+
     return ok;
 }
+
+static int dsa_builtin_keygen(DSA *dsa)
+{
+    return dsa_keygen(dsa, 0);
+}
```
levitte: I didn't notice this before.  So you're renaming an *static* function, then creating a wrapper that's also a *static* function...

Why not just change the existing static function and change all calls from `dsa_builtin_keygen(dsa)` to `dsa_builtin_keygen(dsa, 0)`?

levitte: I do understand the corresponding `ec_key_simple_generate_key` change, since the calls are so much more widely spread.  But when it's limited to the same file?  Less so...

crypto/rsa/rsa_gen.c
```diff
@@ -398,11 +402,89 @@ static int rsa_builtin_keygen(RSA *rsa, int bits, int primes, BIGNUM *e_value,
...
+        }
+    }
+    return ok;
+}
+
+static int rsa_builtin_keygen(RSA *rsa, int bits, int primes, BIGNUM *e_value,
+                              BN_GENCB *cb)
+{
+    return rsa_keygen(NULL, rsa, bits, primes, e_value, cb, 0);
+}
```
levitte: Same kind of reasoning here

crypto/dsa/dsa_key.c
slontis: Its also just one call :) Changed

crypto/rsa/rsa_gen.c
slontis: Done

slontis:
Thanks.. Merged to master.

PR #10953, opened 2020-01-27
config: ensure the perl Configure run is the last statement

### LINKED ISSUE #10951: 1.1.1d ./config does not return an error during configuration error
I'm working from the OpenSSL 1.1.1d tarball. I'm catching a configuration error when running `./config`:

```
***** Unsupported options: no-comp --prefix=/home/jwalton/tmp/build-test --libdir=/home/jwalton/tmp/build-test/lib
```

The problem for this report is, `config` does not return an error so my script continues as if there is no error:

```
./config  no-comp --prefix=/home/jwalton/tmp/build-test --libdir=/home/jwalton/tmp/build-test/lib
Operating system: x86_64-whatever-linux2
...
***** Unsupported options: no-comp --prefix=/home/jwalton/tmp/build-test --libdir=/home/jwalton/tmp/build-test/lib
**********************
Building package
**********************
```

I should never get to the message *"Building Package"* because:

```
    KERNEL_BITS="$INSTX_BITNESS" \
./config "${CONFIG_FLAGS[*]}" \
    PKG_CONFIG_PATH=\""${BUILD_PKGCONFIG[*]}"\" \
    CPPFLAGS=\""${BUILD_CPPFLAGS[*]}" -DPEDANTIC\" \
    CFLAGS=\""${BUILD_CFLAGS[*]}"\" \
    CXXFLAGS=\""${BUILD_CXXFLAGS[*]}"\" \
    LDFLAGS=\""${BUILD_LDFLAGS[*]}"\"

if [[ "$?" -ne 0 ]]; then
    echo "Failed to configure OpenSSL"
    exit 1
fi

echo "**********************"
echo "Building package"
echo "**********************"

MAKE_FLAGS=("-j" "$INSTX_JOBS")
if ! "$MAKE" "${MAKE_FLAGS[@]}"
then
    echo "Failed to build OpenSSL"
    exit 1
fi
````

It appears `config` is returning success even during an error.

### PR DESCRIPTION
**Pull request by levitte**
Running any statement after Configure means we lose its exit code

Fixes #10951

### DIFF
--- a/config
+++ b/config
@@ -898,14 +898,12 @@ fi
 
 OUT="$OUT"
 
-$PERL $THERE/Configure LIST | grep "$OUT" > /dev/null
-
 if [ "$OUT" = "darwin64-x86_64-cc" ]; then
     echo "WARNING! If you wish to build 32-bit libraries, then you have to"
     echo "         invoke 'KERNEL_BITS=32 $THERE/config $options'."
 fi
 
-if [ $? = "0" ]; then
+if $PERL $THERE/Configure LIST | grep "$OUT" > /dev/null; then
   if [ "$VERBOSE" = "true" ]; then
     echo /usr/bin/env \
 	 __CNF_CPPDEFINES="'$__CNF_CPPDEFINES'" \
--- a/config
+++ b/config
@@ -899,6 +899,12 @@ fi
 OUT="$OUT"
 
 $PERL $THERE/Configure LIST | grep "$OUT" > /dev/null
+
+if [ "$OUT" = "darwin64-x86_64-cc" ]; then
+    echo "WARNING! If you wish to build 32-bit libraries, then you have to"
+    echo "         invoke 'KERNEL_BITS=32 $THERE/config $options'."
+fi
+
 if [ $? = "0" ]; then
   if [ "$VERBOSE" = "true" ]; then
     echo /usr/bin/env \
@@ -929,8 +935,5 @@ else
   exit 1
 fi
 
-if [ "$OUT" = "darwin64-x86_64-cc" ]; then
-    echo "WARNING! If you wish to build 32-bit libraries, then you have to"
-    echo "         invoke 'KERNEL_BITS=32 $THERE/config $options'."
-fi
+# Do not add anothing from here on, so we don't lose the Configure exit code
 )

### CODE REVIEW

**Review by mattcaswell — DISMISSED** [review]

File: config [hunk 1]
```diff
@@ -899,6 +899,12 @@ fi
 OUT="$OUT"
 
 $PERL $THERE/Configure LIST | grep "$OUT" > /dev/null
```
richsalz: What's the exit code of "foo | bar" ?  Is it foo or bar?  Answer is it depends on the shell.  Perhaps rewrite this into two parts and explicitly save the command status that you care about.  Also, the if/echo statements on the following line will clobber the exit status anyway; try this:
```
MAC ~$ cat test 
false
echo $?
if [ "$1" = "y" ] ; then echo ok; fi
echo $?

MAC ~$ sh test y
1
ok
0

MAC ~$ sh test
1
0

MAC ~$ 
```

levitte: Ah ! Good catch

kaduk (approved):

config
```diff
@@ -898,14 +898,12 @@ fi
...
 
-$PERL $THERE/Configure LIST | grep "$OUT" > /dev/null
-
 if [ "$OUT" = "darwin64-x86_64-cc" ]; then
     echo "WARNING! If you wish to build 32-bit libraries, then you have to"
     echo "         invoke 'KERNEL_BITS=32 $THERE/config $options'."
 fi
 
-if [ $? = "0" ]; then
+if $PERL $THERE/Configure LIST | grep "$OUT" > /dev/null; then
```
mspncp: You reversed the order of the Configure command and the warning, now the warning comes at the beginning instead of the end. This will make it more likely to be overlooked IMHO. How about just saving the error instead?

```bash
$PERL $THERE/Configure LIST | grep "$OUT" > /dev/null
result=$?

if [ "$OUT" = "darwin64-x86_64-cc" ]; then
    echo "WARNING! If you wish to build 32-bit libraries, then you have to"
    echo "         invoke 'KERNEL_BITS=32 $THERE/config $options'."
fi

if [ $result = "0" ]; then
...
fi

mspncp: Oh, forget my last comment, I overlooked the `> /dev/null`.

levitte:
Merged

master:
4bf3e989fef9268507ba02744e7f71ee5637681c config: ensure the perl Configure run is the last statement

1.1.1:
d2e8cbfb45e8dbd24f7bb394b0141130b4c0eb98 config: ensure the perl Configure run is the last statement

PR #10954, opened 2020-01-27
[WIP] Allow keys without available public key to be used for SSL

dwmw2:
This is the continuation of #1551. This is actually against 1.1.1 not 3.0.0 as discussed, but it's basically the same and testing on 1.1.1 means I don't have to rebuild the whole world including the PKCS#11 engine.

It's a special case for EC for now; we can later extend the generic EVP_PKEY API to have a generic "here's a public key I checked matches this private key; you might want to remember it" method.

But first we need to deal with the fact that X509_check_private_key(), where we want to do this thing, now takes a 'const' private key, so it's not clear if we should be doing *anything* of that form at all.

From my notes in #1551;

```
crypto/x509/x509_cmp.c: In function 'check_key_signature':
crypto/x509/x509_cmp.c:280:42: warning: passing argument 1 of 'EVP_PKEY_CTX_new' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
     EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new(priv, NULL);
                                          ^~~~
In file included from include/openssl/x509.h:18:0,
                 from crypto/x509/x509_cmp.c:14:
include/openssl/evp.h:1338:15: note: expected 'EVP_PKEY * {aka struct evp_pkey_st *}' but argument is of type 'const EVP_PKEY * {aka const struct evp_pkey_st *}'
 EVP_PKEY_CTX *EVP_PKEY_CTX_new(EVP_PKEY *pkey, ENGINE *e);
               ^~~~~~~~~~~~~~~~
crypto/x509/x509_cmp.c:304:28: warning: passing argument 1 of 'EVP_PKEY_CTX_new' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
     ctx = EVP_PKEY_CTX_new(pub, NULL);
                            ^~~
In file included from include/openssl/x509.h:18:0,
                 from crypto/x509/x509_cmp.c:14:
include/openssl/evp.h:1338:15: note: expected 'EVP_PKEY * {aka struct evp_pkey_st *}' but argument is of type 'const EVP_PKEY * {aka const struct evp_pkey_st *}'
 EVP_PKEY_CTX *EVP_PKEY_CTX_new(EVP_PKEY *pkey, ENGINE *e);
               ^~~~~~~~~~~~~~~~

```

The correct way to fix that is not obvious. Is the first argument to EVP_PKEY_CTX_new() really not const? Casting to non-const seems wrong, and that's just for the first part. When we later come to copy the public key data, it gets even wronger...

### DIFF
--- a/crypto/x509/x509_cmp.c
+++ b/crypto/x509/x509_cmp.c
@@ -275,6 +275,51 @@ EVP_PKEY *X509_get_pubkey(X509 *x)
     return X509_PUBKEY_get(x->cert_info.key);
 }
 
+static int check_key_signature(const EVP_PKEY *pub, const EVP_PKEY *priv)
+{
+    EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new(priv, NULL);
+    unsigned char digest[SHA_DIGEST_LENGTH];
+    unsigned char *sig = NULL;
+    size_t siglen;
+    int ret = 0;
+
+    if (ctx == NULL)
+        return 0;
+
+    /* We just sign dummy data. It could be random, but there's no point */
+    memset(digest, 0x5a, sizeof(digest));
+
+    if (!EVP_PKEY_sign_init(ctx))
+        goto out;
+    if (!EVP_PKEY_sign(ctx, NULL, &siglen, digest, sizeof(digest)))
+        goto out;
+    sig = OPENSSL_malloc(siglen);
+    if (!sig)
+        goto out;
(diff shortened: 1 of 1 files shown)


levitte:
I believe that the sane thing is to consider making it a separate public functions, something like `X509_check_signature`, to be called separately from, or maybe even instead of `X509_check_private_key`.

While you're at it, I suggest also adding `X509_REQ_check_private_key`, for API consistency.

PR #10957, opened 2020-01-27
WIP: Run cmd-option nits

richsalz:
This requires my other doc PR's to be merged, else it will fail.  So this is WIP testing failures :)
Based on #10873


richsalz:
Replaced by #10972; closing.

PR #10959, opened 2020-01-27
Remove unused HAS_LFN_SUPPORT

richsalz:
silly little nit.  could be backported.


paulidale:
Merged to master.  I don't think this qualifies for a 1.1.1 merge since it isn't a bug.

Thanks for the submission.

PR #10960, opened 2020-01-28
EC: deprecate the low level ECDH, ECDSA and EC_KEY_METHOD functions.

paulidale:
Deprecate the low level EC functions.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -22,6 +22,30 @@
      However, code that does the latter will still work as before.
      [Richard Levitte]
 
+  *) Deprecated low level ECDH and ECDSA functions.  These include:
+
+     ECDH_compute_key, ECDSA_do_sign, ECDSA_do_sign_ex, ECDSA_do_verify,
+     ECDSA_sign_setup, ECDSA_sign, ECDSA_sign_ex, ECDSA_verify and
+     ECDSA_size.
+
+     Use of these low level functions has been informally discouraged for a long
+     time.  Instead applications should use the EVP_PKEY_derive(3),
+     EVP_DigestSign(3) and EVP_DigestVerify(3) functions.
+     [Paul Dale]
+
+  *) Deprecated the EC_KEY_METHOD functions.  These include:
+
+     EC_KEY_METHOD_new, EC_KEY_METHOD_free, EC_KEY_METHOD_set_init,
+     EC_KEY_METHOD_set_keygen, EC_KEY_METHOD_set_compute_key,
+     EC_KEY_METHOD_set_sign, EC_KEY_METHOD_set_verify,
+     EC_KEY_METHOD_get_init, EC_KEY_METHOD_get_keygen,
+     EC_KEY_METHOD_get_compute_key, EC_KEY_METHOD_get_sign and
+     EC_KEY_METHOD_get_verify.
+
+     Instead applications and extension writers should use the OSSL_PROVIDER
+     APIs.
+     [Paul Dale]
+
   *) Deprecated EVP_PKEY_decrypt_old(), please use EVP_PKEY_decrypt_init()
      and EVP_PKEY_decrypt() instead.
      Deprecated EVP_PKEY_encrypt_old(), please use EVP_PKEY_encrypt_init()
@@ -60,21 +84,21 @@
   *) All of the low level MD2, MD4, MD5, MDC2, RIPEMD160, SHA1, SHA224, SHA256,
      SHA384, SHA512 and Whirlpool digest functions have been deprecated.
      These include:
-         MD2, MD2_options, MD2_Init, MD2_Update, MD2_Final, MD4, MD4_Init,
-         MD4_Update, MD4_Final, MD4_Transform, MD5, MD5_Init, MD5_Update,
-         MD5_Final, MD5_Transform, MDC2, MDC2_Init, MDC2_Update, MDC2_Final,
-         RIPEMD160, RIPEMD160_Init, RIPEMD160_Update, RIPEMD160_Final,
-         RIPEMD160_Transform, SHA1_Init, SHA1_Update, SHA1_Final,
-         SHA1_Transform, SHA224_Init, SHA224_Update, SHA224_Final,
-         SHA224_Transform, SHA256_Init, SHA256_Update, SHA256_Final,
-         SHA256_Transform, SHA384, SHA384_Init, SHA384_Update, SHA384_Final,
-         SHA512, SHA512_Init, SHA512_Update, SHA512_Final, SHA512_Transform,
-         WHIRLPOOL, WHIRLPOOL_Init, WHIRLPOOL_Update, WHIRLPOOL_BitUpdate
-         and WHIRLPOOL_Final.
 
-     Use of these low level functions has been informally discouraged for a long
-     time.  Instead applications should instead use the EVP_DigestInit_ex,
-     EVP_DigestUpdate(3) and EVP_DigestFinal_ex(3) functions.
+     MD2, MD2_options, MD2_Init, MD2_Update, MD2_Final, MD4, MD4_Init,
+     MD4_Update, MD4_Final, MD4_Transform, MD5, MD5_Init, MD5_Update,
+     MD5_Final, MD5_Transform, MDC2, MDC2_Init, MDC2_Update, MDC2_Final,
+     RIPEMD160, RIPEMD160_Init, RIPEMD160_Update, RIPEMD160_Final,
+     RIPEMD160_Transform, SHA1_Init, SHA1_Update, SHA1_Final, SHA1_Transform,
+     SHA224_Init, SHA224_Update, SHA224_Final, SHA224_Transform, SHA256_Init,
+     SHA256_Update, SHA256_Final, SHA256_Transform, SHA384, SHA384_Init,
+     SHA384_Update, SHA384_Final, SHA512, SHA512_Init, SHA512_Update,
+     SHA512_Final, SHA512_Transform, WHIRLPOOL, WHIRLPOOL_Init,
+     WHIRLPOOL_Update, WHIRLPOOL_BitUpdate and WHIRLPOOL_Final.
+
+     Use of these low level functions has been informally discouraged
+     for a long time.  Applications should use the EVP_DigestInit_ex(3),
+     EVP_DigestUpdate(3) and EVP_DigestFinal_ex(3) functions instead.
      [Paul Dale]
 
   *) Corrected the documentation of the return values from the EVP_DigestSign*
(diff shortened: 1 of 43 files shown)


paulidale:
Unsure if I've been too aggressive about this or not aggressive enough.

CHANGES
```diff
@@ -9,6 +9,26 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) Deprecated low level ECDH and ECDSA functions. These include:
+         ECDH_compute_key, ECDSA_SIG_new, d2i_ECDSA_SIG, i2d_ECDSA_SIG,
+         ECDSA_SIG_get0, ECDSA_SIG_get0_r, ECDSA_SIG_get0_s, ECDSA_SIG_set0,
+         ECDSA_do_sign, ECDSA_do_sign_ex, ECDSA_do_verify, ECDSA_sign_setup,
+         ECDSA_sign, ECDSA_sign_ex, ECDSA_verify and ECDSA_size.
```
mattcaswell: I kind of think that perhaps the `ECDSA_SIG_*` functions should stay? They are probably still useful even after the move to providers if for some reason you want do dive into the r and s values. I had a question just the other day where someone had the raw r and s values and wanted to construct a signature that was readable by OpenSSL with them.

CHANGES
```diff
@@ -9,6 +9,26 @@
...
+     EVP_DigestSign(3) and EVP_DigestVerify(3) functions.
+     [Paul Dale]
+
+  *) Deprecated the EC_KEY_METHOD functions. These include:
+         EC_KEY_METHOD_new, EC_KEY_METHOD_free, EC_KEY_METHOD_set_init,
+         EC_KEY_METHOD_set_keygen, EC_KEY_METHOD_set_compute_key,
+         EC_KEY_METHOD_set_sign, EC_KEY_METHOD_set_verify,
+         EC_KEY_METHOD_get_init.  EC_KEY_METHOD_get_keygen.
+         EC_KEY_METHOD_get_compute_key, EC_KEY_METHOD_get_sign and
+         EC_KEY_METHOD_get_verify.
```
mattcaswell: Needs a statement saying what people should do instead.

apps/speed.c
```diff
@@ -1100,6 +1100,7 @@ static int DSA_verify_loop(void *args)
 #endif
 
 #ifndef OPENSSL_NO_EC
+#ifndef OPENSSL_NO_DEPRECATED_3_0
```
mattcaswell: Needs a extra space indent (and on the endif)

apps/speed.c
```diff
@@ -3004,6 +3006,7 @@ int speed_main(int argc, char **argv)
 #endif                          /* OPENSSL_NO_DSA */
 
 #ifndef OPENSSL_NO_EC
+#ifndef OPENSSL_NO_DEPRECATED_3_0
```
mattcaswell: And here

apps/speed.c
```diff
@@ -3392,6 +3395,14 @@ int speed_main(int argc, char **argv)
                 EVP_PKEY_CTX_free(sm2_vfy_pctx);
                 break;
             }
+            if (!EVP_PKEY_sign_init(sm2_pctx)
+                || !EVP_PKEY_sign(sm2_pctx, NULL, &loopargs[i].sigsize,
+                                  NULL, 0)) {
+                EVP_PKEY_CTX_free(pctx);
+                EVP_PKEY_CTX_free(sm2_vfy_pctx);
+                break;
+            }
```
mattcaswell: Why are you adding this?

crypto/ec/ecdh_kdf.c
```diff
@@ -7,6 +7,12 @@
  * https://www.openssl.org/source/license.html
  */
 
+/*
+ * EC low level APIs are deprecated for public use, but still ok for
```
mattcaswell: Should this say ECDH not EC?

crypto/sm2/sm2_pmeth.c
```diff
@@ -7,6 +7,12 @@
  * https://www.openssl.org/source/license.html
  */
 
+/*
+ * EC low level APIs are deprecated for public use, but still ok for
```
mattcaswell: ECDSA?

CHANGES
t8m: I am definitely +1 to this as this is an interface to access the data and not to invoke any crypto operations.

crypto/ec/ecdh_kdf.c
paulidale: Yes.  Fixed.

CHANGES
paulidale: I've added such a statement, not sure what the best replacement is.

levitte: The replacement is to write a provider, possibly one built in to the application.

CHANGES
```diff
@@ -9,6 +9,28 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) Deprecated low level ECDH and ECDSA functions. These include:
+         ECDH_compute_key, ECDSA_do_sign, ECDSA_do_sign_ex, ECDSA_do_verify,
+         ECDSA_sign_setup, ECDSA_sign, ECDSA_sign_ex, ECDSA_verify and
+         ECDSA_size.
```
levitte: I suggest looking at previous examples for enumerations like this. I would intuitively add en empty line before the enumeration, and no extra indentation. I believe that's the style we have used previously.

CHANGES
```diff
@@ -9,6 +9,28 @@
...
+
+  *) Deprecated the EC_KEY_METHOD functions. These include:
+         EC_KEY_METHOD_new, EC_KEY_METHOD_free, EC_KEY_METHOD_set_init,
+         EC_KEY_METHOD_set_keygen, EC_KEY_METHOD_set_compute_key,
+         EC_KEY_METHOD_set_sign, EC_KEY_METHOD_set_verify,
+         EC_KEY_METHOD_get_init.  EC_KEY_METHOD_get_keygen.
+         EC_KEY_METHOD_get_compute_key, EC_KEY_METHOD_get_sign and
+         EC_KEY_METHOD_get_verify.
+
+     Instead applications and extension writers should use the OOSL_PROVIDER
```
levitte: Spelling

CHANGES
```diff
@@ -36,7 +58,7 @@
          WHIRLPOOL, WHIRLPOOL_Init, WHIRLPOOL_Update, WHIRLPOOL_BitUpdate
          and WHIRLPOOL_Final.
      Use of these low level functions has been informally discouraged for a long
-     time.  Instead applications should instead use the EVP_DigestInit_ex,
+     time.  Instead applications should use the EVP_DigestInit_ex(3),
      EVP_DigestUpdate(3) and EVP_DigestFinal_ex(3) functions.
```
levitte: How about "Applications should use ... functions instead"?

apps/speed.c
levitte: I like this, 'cause it gets the *actual* signature size. `ECDSA_size` only gives a guesstimate, and relies on the assumption that a signature will never be bigger than the ECDSA one.

levitte: If people end up doing this a lot, we might want to considering adding a public function that computes the signature size for a given key och key context. Maybe there already is one? I forget...

doc/man3/ECDSA_SIG_new.pod
```diff
@@ -69,6 +70,10 @@ after this function has been called.
 See L<i2d_ECDSA_SIG(3)> and L<d2i_ECDSA_SIG(3)> for information about encoding
 and decoding ECDSA signatures to/from DER.
 
+All of the functions described below are deprecated. Applications should
+use the higher level B<EVP> interface such as L<EVP_DigestSignInit(3)>
+or L<EVP_DigestVerifyInit(3)> instead.
+
 ECDSA_size() returns the maximum length of a DER encoded ECDSA signature
 created with the private EC key B<eckey>.
```
levitte: You might want to document recommended methods to get the actual signature size

CHANGES
paulidale: It's been variable previously.  I'll do something :)

doc/man3/EC_KEY_new.pod
```diff
@@ -19,6 +19,10 @@ EC_KEY objects
 
  #include <openssl/ec.h>
 
+Deprecated since OpenSSL 3.0, can be hidden entirely by defining
+B<OPENSSL_API_COMPAT> with a suitable version value, see
+L<openssl_user_macros(7)>:
+
```
levitte: Didn't we at some point decide not to depreciate key *creation* functions yet?
Low level key generation functions and functions that use these keys are up for deprecation, no doubt about that.

Although, with the EVP level fromdata functions, things may have shifted a bit? We should confirm among ourselves that this is the right move, though...

apps/speed.c
paulidale: An EVP_PKEY API for this is probably sensible.

CHANGES
paulidale: I reformatted this and the few after to the blank line before and after without indentation style.

doc/man3/ECDSA_SIG_new.pod
paulidale: Added a sentence suggesting `EVP_PKEY_sign(3)` with a NULL signature parameter.

doc/man3/EC_KEY_new.pod
paulidale: We did decide to not deprecate thees.  They are still in `ec.h` but over zealous copy paste resulted in the pod file being modified.  I've reverted.

apps/speed.c
mattcaswell: `EVP_PKEY_size()`?

paulidale: I feel a bit embarrassed at having missing this call.
I've changed now.

paulidale:
Rebased to remove merge conflict.

include/openssl/ec.h
```diff
@@ -1200,8 +1204,12 @@ void ECDSA_SIG_free(ECDSA_SIG *sig);
  *  \param  pp   pointer to a unsigned char pointer for the output or NULL
  *  \return the length of the DER encoded ECDSA_SIG object or a negative value
  *          on error
+ *
+ * Originally these were macro exapanded from:
+ *  DECLARE_ASN1_ENCODE_FUNCTIONS_only(ECDSA_SIG, ECDSA_SIG)
```
mattcaswell: Why do we make this change?

paulidale: Unless I introduce a new macro, these functions can't be deprecated otherwise.
Not deprecating them is an option -- do we want this?

mattcaswell: Well you're not deprecating them anymore anyway!

richsalz: Convention has been to use ```|foo|``` to indicate parameters, and just run the comment text as a set of prose.  Note to use javadoc conventions.  (For example, remove the : and newline if it doesn't wrap from 1208/1209.)

paulidale: Reminder to self: don't try to think late on a Friday evening after an hectic week.

paulidale: Reverted this to its original form.

Reworking the comments to follow the parameter convention is for another PR.

paulidale:
Rebased to remove the conflicts.

paulidale:
Merged to master.  Thanks for the reviews.

Issue #10962, opened 2020-01-28
Importing Arbitrary Key Data


mattcaswell:
We should have a mechanism for applications to import arbitrary key data into a provider. This is separate to key generation (where the keys are created inside the provider), or de-serialization (where the keys are loaded directly by the provider). This is the scenario where the application knows the low level key details and wants to create that key directly in the provider. We can do this with our built-in types (RSA, DSA, DH etc), via `evp_keymgmt_importkey`, but there is no application API directly to that  function.

levitte:
Er, yes there is.

https://github.com/openssl/openssl/blob/master/doc/man3/EVP_PKEY_fromdata.pod

mattcaswell:
Hmmmm...this seems to use `evp_keymgmt_fromdata` (which explains why I couldn't find it). What is the difference between that and `evp_keymgmt_importkey`? Why do we need both?

levitte:
All you have to do is look

https://github.com/openssl/openssl/blob/0cb3f4f985aa52688ddc90075f480ba612d34fe3/crypto/evp/keymgmt_lib.c#L206-L218

levitte:
It's true, though, that we I could have had `evp_keymgmt_importdomparams(keymgmt, params)` instead of `keymgmt->importdomparams(provctx, params)`, that would have avoided getting the provctx in this particular function.

mattcaswell:
Thanks for clearing my confusion. My issue was unfounded. Closing.

levitte:
> My issue was unfounded.

Not entirely, I can see how the way the function was written can be confounding.  I'll submit a small refactor for clarity.

PR #10963, opened 2020-01-28
EVP: Small refactor of keymgmt library code

levitte:
Some functions went directly on keymgmt function pointers instead of
going through the internal KEYMGMT API, which makes for a confusing
read.

Related to #10962

### DIFF
--- a/crypto/evp/keymgmt_lib.c
+++ b/crypto/evp/keymgmt_lib.c
@@ -18,7 +18,7 @@
 
 struct import_data_st {
     void *provctx;
-    void *(*importfn)(void *provctx, const OSSL_PARAM params[]);
+    void *(*importfn)(const EVP_KEYMGMT *keymgmt, const OSSL_PARAM params[]);
 
     /* Result */
     void *provdata;
@@ -92,7 +92,9 @@ void *evp_keymgmt_export_to_provider(EVP_PKEY *pk, EVP_KEYMGMT *keymgmt,
         struct import_data_st import_data;
 
         import_data.importfn =
-            want_domainparams ? keymgmt->importdomparams : keymgmt->importkey;
+            want_domainparams
+            ? evp_keymgmt_importdomparams
+            : evp_keymgmt_importkey;
         import_data.provdata = NULL;
 
         /*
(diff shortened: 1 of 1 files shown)


mattcaswell (approved):

crypto/evp/keymgmt_lib.c
```diff
@@ -92,7 +92,9 @@ void *evp_keymgmt_export_to_provider(EVP_PKEY *pk, EVP_KEYMGMT *keymgmt,
         struct import_data_st import_data;
 
         import_data.importfn =
-            want_domainparams ? keymgmt->importdomparams : keymgmt->importkey;
+            want_domainparams
```
slontis: I think once it becomes this messy it should become a if.
Feel free to ignore if you disagree

crypto/evp/keymgmt_lib.c
```diff
@@ -102,10 +104,11 @@ void *evp_keymgmt_export_to_provider(EVP_PKEY *pk, EVP_KEYMGMT *keymgmt,
             return NULL;
 
         for (j = 0; j < i && pk->pkeys[j].keymgmt != NULL; j++) {
-            int (*exportfn)(void *provctx, OSSL_CALLBACK *cb, void *cbarg) =
+            int (*exportfn)(const EVP_KEYMGMT *keymgmt, void *provdata,
```
slontis: A typedef is probably better.. especially if the function signature is going to change. (Again feel free to disagree)

levitte: I would if it was going to get used in more places, but as far as I'm concerned, this is the only spot where using a function pointer for this particular signature, so...

(don't you wish that C had a `typeof` that you could give a function name, and that would be the signature of that function?  Alas...)

crypto/evp/keymgmt_lib.c
levitte: I disagree.

levitte:
Merged

658608c471a6e1f9b6d7f88c060a7adb77d7d334 EVP: Small refactor of keymgmt library code

Issue #10353, opened 2019-11-04
Move X25519 implementation to provider.


paulidale:
Move the X25519 code to `providers/implementation/exchanges`.  The existing `dh_exch.c` file should give a general idea as to the bulk of the changes required.

This will probably need #10351 done first.

romen:
See #10964

PR #10964, opened 2020-01-28
Move X25519/X448 to the default provider

mattcaswell:
This doesn't yet implement the required serializers. That will come in a subsequent PR.


crypto/ec/ecx_meth.c
```diff
@@ -1061,18 +1098,18 @@ static int s390x_pkey_ecx_keygen25519(EVP_PKEY_CTX *ctx, EVP_PKEY *pkey)
         0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
         0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00
     };
-    ECX_KEY *key;
+    ECX_KEY *key = NULL;
```
paulidale: Setting this to NULL seems unnecessary given it is unconditionally assigned three lines later.

crypto/ec/ecx_meth.c
```diff
@@ -1106,18 +1141,18 @@ static int s390x_pkey_ecx_keygen448(EVP_PKEY_CTX *ctx, EVP_PKEY *pkey)
...
```
paulidale: Same here.

crypto/ec/ecx_meth.c
```diff
@@ -1154,19 +1187,19 @@ static int s390x_pkey_ecd_keygen25519(EVP_PKEY_CTX *ctx, EVP_PKEY *pkey)
         0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66,
     };
     unsigned char x_dst[32], buff[SHA512_DIGEST_LENGTH];
-    ECX_KEY *key;
+    ECX_KEY *key = NULL;
```
paulidale: Same here.

crypto/ec/ecx_meth.c
```diff
@@ -1214,19 +1245,19 @@ static int s390x_pkey_ecd_keygen448(EVP_PKEY_CTX *ctx, EVP_PKEY *pkey)
         0x24, 0xbc, 0xb6, 0x6e, 0x71, 0x46, 0x3f, 0x69, 0x00
     };
     unsigned char x_dst[57], buff[114];
-    ECX_KEY *key;
+    ECX_KEY *key = NULL;
```
paulidale: And here.

include/crypto/ecx.h
```diff
@@ -0,0 +1,66 @@
...
+
+/* Internal EC functions for other submodules: not for application use */
+
+#ifndef OSSL_CRYPTO_ECX_H
+# define OSSL_CRYPTO_ECX_H
+# include <openssl/opensslconf.h>
+
+# ifndef OPENSSL_NO_EC
+
+#include <openssl/e_os2.h>
```
paulidale: Preprocessor indentation is awry.

providers/implementations/exchange/ecx_exch.c
```diff
@@ -0,0 +1,200 @@
...
+        ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_KEY_LENGTH);
+        return 0;
+    }
+
+    if (secret == NULL) {
+        *secretlen = ecxctx->keylen;
+        return 1;
+    }
+    if (outlen < ecxctx->keylen) {
+        ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_KEY_LENGTH);
```
paulidale: Would it make sense to return the desired size in this case too?  Perhaps do it unconditionally before the `secret == NULL` check?

providers/implementations/keymgmt/ecx_kmgmt.c
```diff
@@ -0,0 +1,169 @@
...
+{
+    if (key == NULL)
+        return 0;
+
+    if (!ossl_param_bld_push_octet_string(tmpl, OSSL_PKEY_PARAM_ECX_PUB_KEY,
+                                          key->pubkey, keylen))
+        return 0;
+
+    if (key->privkey != NULL
+        && !ossl_param_bld_push_octet_string(tmpl, OSSL_PKEY_PARAM_ECX_PRIV_KEY,
```
paulidale: I wonder if the param builder should be able to accept private octet and utf8 strings (storing them in secure memory)?

paulidale:
Travis is also relevant.

crypto/ec/ecx_key.c
```diff
@@ -0,0 +1,68 @@
...
+    if (key == NULL)
+        return;
+
+    CRYPTO_DOWN_REF(&key->references, &i, key->lock);
+    REF_PRINT_COUNT("ECX_KEY", r);
+    if (i > 0)
+        return;
+    REF_ASSERT_ISNT(i < 0);
+
+    OPENSSL_secure_free(key->privkey);
```
slontis: old code was doing secure_clear_free()

crypto/ec/ecx_meth.c
```diff
@@ -410,6 +397,52 @@ static int ecx_get_pub_key(const EVP_PKEY *pkey, unsigned char *pub,
...
+    OSSL_PARAM *params;
+    void *provdata = NULL;
+
+    /* There are no domain parameters for ECX keys */
+    if (want_domainparams)
+        return NULL;
+
+    ossl_param_bld_init(&tmpl);
+
+    /* A key must at least have a public part */
```
slontis: Is this always true if it generates the public part from the private part?

crypto/ec/ecx_meth.c
```diff
@@ -1154,19 +1187,19 @@ static int s390x_pkey_ecd_keygen25519(EVP_PKEY_CTX *ctx, EVP_PKEY *pkey)
         0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66,
     };
     unsigned char x_dst[32], buff[SHA512_DIGEST_LENGTH];
-    ECX_KEY *key;
+    ECX_KEY *key = NULL;
     unsigned char *privkey = NULL, *pubkey;
```
slontis: Unless I am mistaken these 3 functions below are now identical - except for the id, and the ECErr function names.
So it could be refactored to be a single function,

include/openssl/core_names.h
```diff
@@ -185,6 +185,10 @@ extern "C" {
 #define OSSL_PKEY_PARAM_DSA_PUB_KEY  "pub"
 #define OSSL_PKEY_PARAM_DSA_PRIV_KEY "priv"
 
+/* ECX Keys */
+#define OSSL_PKEY_PARAM_ECX_PUB_KEY  "pub"
```
slontis: Maybe there should be a common define that is shared for these..

providers/defltprov.c
```diff
@@ -371,6 +371,8 @@ static const OSSL_ALGORITHM deflt_kdfs[] = {
 static const OSSL_ALGORITHM deflt_keyexch[] = {
 #ifndef OPENSSL_NO_DH
     { "DH:dhKeyAgreement", "default=yes", dh_keyexch_functions },
+    { "X25519", "default=yes", x25519_keyexch_functions },
```
slontis: Is there a OPENSSL_NO_ECX (or should it use NO_EC?)

providers/defltprov.c
```diff
@@ -395,6 +397,8 @@ static const OSSL_ALGORITHM deflt_keymgmt[] = {
     { "DSA:dsaEncryption", "default=yes", dsa_keymgmt_functions },
 #endif
     { "RSA:rsaEncryption", "default=yes", rsa_keymgmt_functions },
+    { "X25519", "default=yes", x25519_keymgmt_functions },
```
slontis: Does this need a OPENSSL_NO_EC?

providers/implementations/exchange/ecx_exch.c
```diff
@@ -0,0 +1,200 @@
...
+static OSSL_OP_keyexch_newctx_fn x448_newctx;
+static OSSL_OP_keyexch_init_fn ecx_init;
+static OSSL_OP_keyexch_set_peer_fn ecx_set_peer;
+static OSSL_OP_keyexch_derive_fn ecx_derive;
+static OSSL_OP_keyexch_freectx_fn ecx_freectx;
+static OSSL_OP_keyexch_dupctx_fn ecx_dupctx;
+
+/*
+ * What's passed as an actual key is defined by the KEYMGMT interface.
+ * We happen to know that our KEYMGMT simply passes DH structures, so
```
slontis: Replace DH..

providers/implementations/keymgmt/ecx_kmgmt.c
```diff
@@ -0,0 +1,169 @@
...
+        return NULL;
+
+    key = ecx_key_new(keylen);
+    if (key == NULL)
+        return NULL;
+
+    pubkey = key->pubkey;
+    if ((param_priv_key != NULL
+             && !OSSL_PARAM_get_octet_string(param_priv_key,
+                                            (void **)&key->privkey, keylen,
```
slontis: Not your fault, but I shudder everytime I see this (void **)& in OSSL_PARAM_get_octet_string

providers/implementations/keymgmt/ecx_kmgmt.c
```diff
@@ -0,0 +1,169 @@
...
+                                             &privkeylen))
+            || !OSSL_PARAM_get_octet_string(param_pub_key,
```
slontis: Hmm this might be more readable if there are 2 if's here..
One that just handles the  private bit and does the privlen check
And one that does the public bit and public length check.

romen:
Just cross-mentioning the related issues: #10353 and #10354

crypto/ec/ecx_meth.c
mattcaswell: Fixed - and the other similar occurrences.

include/crypto/ecx.h
mattcaswell: Fixed.

providers/implementations/exchange/ecx_exch.c
mattcaswell: I think we should avoid side effects in error conditions. I don't think its particularly useful here either.

providers/implementations/keymgmt/ecx_kmgmt.c
mattcaswell: Well the param builder only stores a reference anyway so as long as the original in in secure memory then it should be fine.

crypto/ec/ecx_key.c
mattcaswell: Fixed

crypto/ec/ecx_meth.c
mattcaswell: Yes. Internally if we have the private key then we have the public key too. The only time that isn't true is when the key is first created or the ECK_KEY is first constructed. But that is part of key gen or key mgmt. (Not yet moved that stuff to providers - another PR)

crypto/ec/ecx_meth.c
mattcaswell: Not quite although very similar. There are differences in digests, field multiplication function, key lengths. It could possibly still be refactored - but not in this PR.

include/openssl/core_names.h
mattcaswell: Done

providers/defltprov.c
mattcaswell: Yes - done.

providers/defltprov.c
mattcaswell: Should be OPENSSL_NO_EC. Done

providers/implementations/exchange/ecx_exch.c
mattcaswell: Fixed.

providers/implementations/keymgmt/ecx_kmgmt.c
mattcaswell: Yes - I agree.

providers/implementations/keymgmt/ecx_kmgmt.c
mattcaswell: Done

mattcaswell:
Fixup commits pushed addressing the above comments and the travis failure.

Please take another look.

mattcaswell:
Oh...and also I realised I missed off the s390 support. Now added.

@p-steuer

crypto/s390xcpuid.S
```diff
@@ -0,0 +1,428 @@
+#include "s390x_arch.h"
+
+.text
+
```
slontis: Where has this file come from?

p-steuer: Its generated by s390xcpuid.pl, @mattcaswell  probably added it accidentially.

mattcaswell: Yes I did. I'll remove it.

mattcaswell:
Fixed the s390xcpuid.S issue. The travis red cross was due to a bad auto-merge with master during the travis build. I've rebased to resolve that issue.

Please take another look.

p-steuer:
With moving the s390 code to provider, the corresponding run-time feature detection was lost. My tests of the S390X_EC_ASM build pass on z15 but it should SIGILL on any older machine, since ECC instruction set extension is not available there.

Specifically im talking about the following (non provider) code from crypto/ec/ecx_meth.c:
```
const EVP_PKEY_METHOD *ecx25519_pkey_method(void)
{
#ifdef S390X_EC_ASM
    if (OPENSSL_s390xcap_P.pcc[1] & S390X_CAPBIT(S390X_SCALAR_MULTIPLY_X25519))
        return &ecx25519_s390x_pkey_meth;
#endif
    return &ecx25519_pkey_meth;
}

const EVP_PKEY_METHOD *ecx448_pkey_method(void)
{
#ifdef S390X_EC_ASM
    if (OPENSSL_s390xcap_P.pcc[1] & S390X_CAPBIT(S390X_SCALAR_MULTIPLY_X448))
        return &ecx448_s390x_pkey_meth;
#endif
    return &ecx448_pkey_meth;
}
```

This could be easily fixed by checking the above conditions inside provider code's S390X_EC_ASM ifdefs (before every operation).

However, since this check is only required one time at init, i would prefer a solution similar to the non-provider code (to also get rid of a lot of ifdefs and to facilitate integration of future platform specific hw support). E.g. replace member pointing to an OPENSSL_DISPATCH object in OPENSSL_ALGORITHM structure by a member pointing to a function returning such objects. That way, you could define both a platform-spefific and a generic OPENSSL_DISPATCH array, and said function could chose which one to return, based on the hw capabilities detected on init (analogous to what i did to the standard_methods array here: 19bd1fa1ef17ff0ac071402cb6b002f1c29e37b9).

mattcaswell:
> This could be easily fixed by checking the above conditions inside provider code's S390X_EC_ASM ifdefs (before every operation).

Fixed. I have gone with this simple solution for now. I did look at an option for a one time capability check. We actually have a mechanism for that for ciphers. See:

https://github.com/openssl/openssl/blob/7fa8bcfe4342df41919f5564b315f9c85d0a02d6/providers/defltprov.c#L190-L197

It would not take much to extend that to key exchange. However in this case I think the benefit is very minimal at the cost of quite a lot of extra plumbing. I'm not actually sure you would gain anything very much.

providers/implementations/exchange/ecx_exch.c
```diff
@@ -0,0 +1,223 @@
...
+    }
+    if (outlen < ecxctx->keylen) {
+        ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_KEY_LENGTH);
+        return 0;
+    }
+
+    if (ecxctx->keylen == X25519_KEYLEN) {
+#ifdef S390X_EC_ASM
+        if (OPENSSL_s390xcap_P.pcc[1]
+                & S390X_CAPBIT(S390X_SCALAR_MULTIPLY_X448)) {
```
p-steuer: S390X_SCALAR_MULTIPLY_X448 -> S390X_SCALAR_MULTIPLY_X25519

mattcaswell: Oops. Fixed.

slontis:
needs a rebase

slontis:
@mattcaswell

mattcaswell:
> needs a rebase

Sigh. Needs a bit more work due to the all the keymgmt changes in master.

mattcaswell:
I rebased this and reworked it to take account of the refactored keymgmt in master.

Please take another look and reconfirm?

crypto/ec/ecx_meth.c
```diff
@@ -416,6 +403,48 @@ static int ecx_get_pub_key(const EVP_PKEY *pkey, unsigned char *pub,
...
+
+    params = ossl_param_bld_to_param(&tmpl);
+
+    /* We export, the provider imports */
+    rv = evp_keymgmt_import(to_keymgmt, to_keydata, OSSL_KEYMGMT_SELECT_ALL,
+                            params);
+
+ err:
+    ossl_param_bld_free(params);
+    return rv;
```
p-steuer: rv may be uninitialized here

providers/implementations/keymgmt/ecx_kmgmt.c
```diff
@@ -0,0 +1,231 @@
...
+     * If a private key is present then a public key must also be present.
+     * Alternatively we've just got a public key.
+     */
+    if (param_pub_key == NULL
+            || (param_priv_key == NULL
+                && (selection & OSSL_KEYMGMT_SELECT_PRIVATE_KEY) != 0))
+        return 0;
+
+    if (param_priv_key != NULL
+             && !OSSL_PARAM_get_octet_string(param_priv_key,
```
p-steuer: param_priv_key maybe uninitialized here. Should probably be initialized to NULL.

crypto/ec/ecx_meth.c
mattcaswell: Fixed

providers/implementations/keymgmt/ecx_kmgmt.c
mattcaswell: Fixed.

mattcaswell:
Fixups pushed addressing both of those issues.

mattcaswell:
Pushed. Thanks.

PR #10973, opened 2020-01-29
scrypt/pmeth: free the MD references correctly.

paulidale:
The code was calling EVP_MD_meth_free which is incorrect.  It should call
EVP_MD_free.  It happened to work but by luck rather than design.


paulidale:
Merged to master (albeit a few minutes early).
Thanks for the positive review.

PR #6975, opened 2018-08-16
Updated test command line parsing to support common commands

slontis:
Allows running a single test via the command line option -test X (where X is a id)
Other commands supported include:
-help to show usage
-list to display the id:name of tests
-iter X to specify a single iteration to run

It uses apps/opt.c to specify the options for both the common commands (e.g: -help) as well as the test specific options. The options are all set by the test apps (similar to the openssl apps)
If the test setup() returns 0 now the usage is displayed. (options can be specified for a test via a test_get_options() method). The inbuilt commands must be ignored by the test app as they are handled internally.

There is a unit test for app ui, so I have seperated out the UI specific code to avoid conflicts with the test system.

##### Checklist

- [ ] documentation is added or updated
- [x] tests are added or updated

PR #10975, opened 2020-01-30
Fix common test framework options

mattcaswell:
PR #6975 added the ability to our test framework to have common options to
all tests. For example providing the option "-test 5" to one of our test
programs will just run test number 5. This can be useful when debugging
tests.

Unforuntately this does not work well for a number of tests. In particular
those test that call test_get_argument() without first skipping over these
common test options will not got the expected value. Some tests did this
correctly but a large number did not.

A helper function is introduced, test_skip_common_options(), to make this
easier for those tests which do not have their own specialised test option
handling, but yet still need to call test_get_argument(). This function
call is then added to all those tests that need it.

##### Checklist

### DIFF
--- a/test/asynciotest.c
+++ b/test/asynciotest.c
@@ -397,6 +397,11 @@ OPT_TEST_DECLARE_USAGE("certname privkey\n")
 
 int setup_tests(void)
 {
+    if (!test_skip_common_options()) {
+        TEST_error("Error parsing test options\n");
+        return 0;
+    }
+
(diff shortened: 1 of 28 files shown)


mspncp:
Two typos in commit message:

In particular
those test**s** that call test_get_argument() without first skipping over these
common test options will not **get** the expected value.

mattcaswell:
Commit message updated to fix the typos.

mattcaswell:
Pushed. Thanks.

PR #10976, opened 2020-01-30
Remove unused KDF_R_xxx error codes

richsalz:
These were created during 3.0 development and aren't used.


mattcaswell:
> These were created during 3.0 development and aren't used.

Not all of these were created during 3.0 development. Some of these exist in the 1.1.1 tree.

richsalz:
You're right, I must have been looking at the wrong branch before.  Closing this.

PR #10977, opened 2020-01-31
DSA: deprecate the low level functions.

paulidale:


paulidale:
Still WIP: `evp_extra_test` needs to be updated to not use the low level calls.

slontis:
I foresee some merge problems for me.

CHANGES
```diff
@@ -9,6 +9,32 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) All of the low level DSA functions have been deprecated including:
+
+     DSA_do_sign, DSA_do_verify, DSA_OpenSSL, DSA_set_default_method,
+     DSA_get_default_method, DSA_set_method, DSA_get_method, DSA_new,
+     DSA_new_method, DSA_free, DSA_up_ref, DSA_size, DSA_bits,
```
mattcaswell: We still need to be able to create DSA objects/free them/up-ref them - so I think this goes a step too far. Applications should be able to get/set the pqg params (and I note you don't deprecate those) and then assign the DSA objects to an EVP_PKEY.

include/openssl/dsa.h
```diff
@@ -54,29 +56,36 @@ extern "C" {
...
+#   define d2i_DSAparams_fp(fp, x) \
+        (DSA *)ASN1_d2i_fp((char *(*)())DSA_new, \
+                           (char *(*)())d2i_DSAparams, (fp), \
+                           (unsigned char **)(x))
+#   define i2d_DSAparams_fp(fp, x) \
+        ASN1_i2d_fp(i2d_DSAparams,(fp), (unsigned char *)(x))
+#   define d2i_DSAparams_bio(bp, x) \
+        ASN1_d2i_bio_of(DSA, DSA_new, d2i_DSAparams, bp, x)
+#   define i2d_DSAparams_bio(bp, x) \
+        ASN1_i2d_bio_of(DSA, i2d_DSAparams, bp, x)
```
mattcaswell: Not sure why we deprecate these?

include/openssl/dsa.h
```diff
@@ -85,39 +94,61 @@ DECLARE_ASN1_ENCODE_FUNCTIONS_only(DSA_SIG, DSA_SIG)
...
+DEPRECATEDIN_3_0(void DSA_free(DSA *r))
 /* "up" the DSA object's reference count */
-int DSA_up_ref(DSA *r);
-int DSA_size(const DSA *);
-int DSA_bits(const DSA *d);
-int DSA_security_bits(const DSA *d);
+DEPRECATEDIN_3_0(int DSA_up_ref(DSA *r))
+DEPRECATEDIN_3_0(int DSA_size(const DSA *))
+DEPRECATEDIN_3_0(int DSA_bits(const DSA *d))
+DEPRECATEDIN_3_0(int DSA_security_bits(const DSA *d))
```
mattcaswell: I'm unsure about DSA_size/DSA_bits/DSA_security_bits

include/openssl/dsa.h
```diff
@@ -85,39 +94,61 @@ DECLARE_ASN1_ENCODE_FUNCTIONS_only(DSA_SIG, DSA_SIG)
...
+ */
+DEPRECATEDIN_3_0(DSA *d2i_DSAPublicKey(DSA **a, const unsigned char **in,
+                                       long len))
+DEPRECATEDIN_3_0(int i2d_DSAPublicKey(const DSA *a, unsigned char **out))
+DEPRECATEDIN_3_0(DSA *d2i_DSAPrivateKey(DSA **a, const unsigned char **in,
+                                        long len))
+DEPRECATEDIN_3_0(int i2d_DSAPrivateKey(const DSA *a, unsigned char **out))
+DEPRECATEDIN_3_0(DSA *d2i_DSAparams(DSA **a, const unsigned char **in,
+                                    long len))
+DEPRECATEDIN_3_0(int i2d_DSAparams(const DSA *a, unsigned char **out))
```
mattcaswell: Again - not sure why we deprecate these?

CHANGES
paulidale: Do we want to be able to do this from outside libcrypto?  I thought the idea was to make the DSA structures internal.

mattcaswell: Well, I currently see no reason to prevent people from creating a DSA object if all they're going to do with it is assign it an EVP_PKEY. Its calling the low level *crypto* routines that we want to prevent.

paulidale: I think I've undeprecated all of the relevant calls.
I also left the p, q, g, public and private key getters and setters.

include/openssl/dsa.h
```diff
@@ -130,53 +143,56 @@ DEPRECATEDIN_0_9_8(DSA *DSA_generate_parameters(int bits,
...
+DEPRECATEDIN_3_0(int DSA_generate_key(DSA *a))
 
 int DSAparams_print(BIO *bp, const DSA *x);
 int DSA_print(BIO *bp, const DSA *x, int off);
-# ifndef OPENSSL_NO_STDIO
+#  ifndef OPENSSL_NO_STDIO
 int DSAparams_print_fp(FILE *fp, const DSA *x);
 int DSA_print_fp(FILE *bp, const DSA *x, int off);
-# endif
+#  endif
```
levitte: Is there a reason you don't deprecate the printing functions?

include/openssl/dsa.h
```diff
@@ -85,35 +92,41 @@ DECLARE_ASN1_ENCODE_FUNCTIONS_only(DSA_SIG, DSA_SIG)
...
+                                int siglen, DSA *dsa))
+#  ifndef OPENSSL_NO_DEPRECATED_3_0
+#   define DSA_get_ex_new_index(l, p, newf, dupf, freef) \
+        CRYPTO_get_ex_new_index(CRYPTO_EX_INDEX_DSA, l, p, newf, dupf, freef)
+#  endif
+DEPRECATEDIN_3_0(int DSA_set_ex_data(DSA *d, int idx, void *arg))
+DEPRECATEDIN_3_0(void *DSA_get_ex_data(DSA *d, int idx))
 
 DECLARE_ASN1_ENCODE_FUNCTIONS_only(DSA, DSAPublicKey)
 DECLARE_ASN1_ENCODE_FUNCTIONS_only(DSA, DSAPrivateKey)
```
levitte: Is there a reason not to deprecate the ASN.1 encoding functions?  (just surround them with the `OPENSSL_NO_DEPRECATED_3_0` guard, no need to complicate things further for now)

include/openssl/dsa.h
```diff
@@ -85,35 +92,41 @@ DECLARE_ASN1_ENCODE_FUNCTIONS_only(DSA_SIG, DSA_SIG)
 void DSA_SIG_get0(const DSA_SIG *sig, const BIGNUM **pr, const BIGNUM **ps);
 int DSA_SIG_set0(DSA_SIG *sig, BIGNUM *r, BIGNUM *s);
```
levitte: Hmmm, should the `DSA_SIG` functions remain non-deprecated?  I'd say it's at least up for debate...

include/openssl/dsa.h
t8m: Can you do high level API calls to achieve the same? If so, deprecate, if not, don't deprecate.

levitte: Hmm.... Have do have the serializers, but not the deserializers yet. So OK, good point.
On the other hand, we need the deserializers anyway (upcoming), so this *could* be deprecated preliminarily, but if not, I think a TO-DO comment should be added.

paulidale: I originally did deprecate them but @mattcaswell asked for them not to be...

include/openssl/dsa.h
paulidale: It seemed like they were a good fit with the serialisers.

include/openssl/dsa.h
paulidale: I'm fine to go either way here.

paulidale:
> Poor @paulidale...

This is the least of my problems...

paulidale:
Open questions are:

1. DSA_SIG related function -- deprecate or not?
1. ASN.1 functions -- deprecate or not?
1. printing functions -- deprecate or not?

include/openssl/dsa.h
levitte: Why?

https://github.com/openssl/openssl/blob/7f293d9f3b5cee4f4b15624fff15a45e0517334f/include/openssl/evp.h#L1102-L1104

levitte:
The general viewpoint that at least @mattcaswell and I share nowadays is that the key creation functions should not be deprecated, but the rest goes.  In other words, it should remain possible to `DSA_new()`, then a bit of `DSA_set0_pqg()` and `DSA_set0_key()`, and finish it off with `EVP_PKEY_assign_DSA()`, and anything else should become a matter of using functions that accept a `EVP_PKEY`.

That's at least how I understood things.  But it dawns on me now that @mattcaswell talked about deprecating the functions that do calculations using the keys (encryption, decryption, ...), so it seems like he has a more conservative viewpoint than me, after all.

But on direct questions:

1.  DSA_SIG functions need to stay for the moment, for the simple reason that we don't have a replacement.  We should, however, have a think of making a signature not only a provided method, but also a first class object on the provider level, allowing us to pick out its pieces with OSSL_PARAM arrays just like any other provider object.
2.  ASN.1 encoding/decoding functions should essentially become increasingly deprecated, as we add serializers and deserializers.  We already have the serializers in place, but not the deserializers, and I think it's fine to wait until they are implemented.
3.  I would deprecate the printing functions without any hesitation, especially since they are just wrappers around the EVP_PKEY printing functions.

mattcaswell:
IMO:

- DSA_SIG functions should stay.
- ASN1 encoding/decoding should stay
- I'm ok with deprecating the printing functions
- size/bits/security_bits seems like a bit of a grey area to me. Is it more like getting parameters "p/q/g" which we allow, or is it more like a "crypto" operation. On reflection, I suppose there may actually be some "calculation" involved to return these values - so I'm ok to deprecate.

paulidale:
Thanks for the prompt feedback.  Okay, DSA_SIG and ASN.1 are staying.  The printing and bits/size functions are deprecated.  Clarity is achieved :)

Travis is relevant...

paulidale:
Travis is fixed.

Time to open the review floodgates ...

apps/dhparam.c
```diff
@@ -7,8 +7,11 @@
  * https://www.openssl.org/source/license.html
  */
 
+/* We need to use some deprecated APIs */
+#define OPENSSL_SUPPRESS_DEPRECATED
+
 #include <openssl/opensslconf.h>
-#ifdef OPENSSL_NO_DH
+#if defined(OPENSSL_NO_DSA) || defined(OPENSSL_NO_DEPRECATED_3_0)
```
levitte: Errrrrr...  DSA because?

paulidale: Copy/paste mistake :(
Fixed now.

paulidale:
Rebased to avoid conflicts.
Added deprecation warnings when command line apps are started.

apps/progs.pl
```diff
@@ -95,19 +95,31 @@
...
-        dsaparam => "dsa",
         gendh    => "dh",
-        dhparam  => "dh",
         ecparam  => "ec",
         pkcs12   => "des",
     );
+    my %cmd_deprecated = (
+        dsaparam => [ "3_0", "pkeyparam", "dsa" ],
+        dhparam  => [ "3_0", "pkeyparam", "dh" ],
+    );
```
levitte: I would actually also deprecate "dsa" and "dh" in favor of "pkey", and also "gendsa" in favor of "genpkey"

apps/progs.pl
```diff
@@ -121,6 +133,12 @@
         blake2b512 => "blake2",
         blake2s256 => "blake2",
     );
+    my %md_deprecated = (
+        md2 =>      [ "3_0", "md2" ],
+        md4  =>     [ "3_0", "md4" ],
+        mdc2 =>     [ "3_0", "mdc2" ],
+        rmd160 =>   [ "3_0", "rmd160" ],
+    );
```
levitte: Er, did you actually mean to have that in this PR?

apps/progs.pl
```diff
@@ -143,8 +170,53 @@
...
+        "rc5-ofb" =>     [ "3_0", "rc5" ],
+        "des-ecb" =>     [ "3_0", "des" ],
+        "des-ede" =>     [ "3_0", "des" ],
+        "des-cbc" =>     [ "3_0", "des" ],
+        "des-ede-cbc" => [ "3_0", "des" ],
+        "des-cfb" =>     [ "3_0", "des" ],
+        "des-ede-cfb" => [ "3_0", "des" ],
+        "des-ofb" =>     [ "3_0", "des" ],
+        "des-ede-ofb" => [ "3_0", "des" ],
+        "desx" =>        [ "3_0", "des" ],
```
levitte: Did you mean to have that in this PR?

apps/progs.pl
paulidale: Yes, I intend to push this as a separate commit.
I am willing to make it a separate PR if required, but it's easier here.

apps/progs.pl
paulidale: "dsa", "dh" and "gendsa" all work as is.  Only "gendsa" required changes to work with the deprecations.

I've no objection to deprecating them too, just noting that they weren't a problem converting.

apps/progs.pl
paulidale: _desx_ is going isn't it?????
I thought it was.

paulidale:
> In a `no-deprecated` configuration, we shouldn't even build the dsa (and dh, since you're already on a path to include that in this deprecation) related apps source.

This cascades to test failures.  I tried :)

paulidale:
The partial _dh_ deprecation is here only because _dh_ relies on _dsa_ for parameter generation.  Fully deprecating _dh_ is a lot more effort which I've got underway.

levitte:
> This cascades to test failures. I tried :)

Yup, there are a lot of tests that need a similar check.  #10797 had a lot of failures related to exactly this, but from a different viewpoint.

apps/progs.pl
paulidale: I'm willing to add the warnings here but the deprecation of the command line apps and the consequences should be dealt with in a separate PR and scheduled via the tracking spreadsheet.  I'm expecting this will turn into a can of worms pretty quickly.

paulidale: I've noted the app deprectation as a TODO comment and added a task to the sheet.

paulidale:
Unless we do want app deprecation as part of this, I think I've addressed everything.

apps/progs.pl
levitte: Just wanted to check your intention. I tend to view PRs as topics, and this one has "DSA" in its title, so deprecating a whole bunch of ciphers looked "off-topic"... but then, we all differ, hence my question.

paulidale: The `progs.pl` changes kind of became expedient here.

levitte:
> Unless we do want app deprecation as part of this

???
Considering the progs.pl changes, it looks like you got started on it...

paulidale:
I did what was necessary for this PR and yes, the Perl part is done.  I'm in no doubt that the majority of the work will be fixing problems that follow on.

levitte:
I guess I can be ok with that.  I had some plans on doing something similar, but take it the whole way.  This will at least get us started, then...

paulidale:
Assuming my Perl is adequate 😆

CHANGES
```diff
@@ -9,6 +9,44 @@
...
+     This means that applications don't have to look at the curve NID and
+     'EVP_PKEY_set_alias_type(pkey, EVP_PKEY_SM2)' to get SM2 computations.
+     However, they still can, that EVP_PKEY_set_alias_type() call acts as
+     a no-op when the EVP_PKEY is already of the given type.
+
+     Parameter and key generation is also reworked to make it possible
+     to generate EVP_PKEY_SM2 parameters and keys without having to go
+     through EVP_PKEY_EC generation and then change the EVP_PKEY type.
+     However, code that does the latter will still work as before.
+     [Richard Levitte]
```
levitte: you can remove this, since it's already there just below you added entries :wink:

apps/gendsa.c
```diff
@@ -126,18 +133,40 @@ int gendsa_main(int argc, char **argv)
...
+    if (!EVP_PKEY_keygen_init(ctx) || !EVP_PKEY_keygen(ctx, &pkey)) {
+        BIO_printf(bio_err, "Error generating key\n");
+        goto end;
+    }
+
+    dsa = EVP_PKEY_get1_DSA(pkey);
+    if (dsa == NULL) {
+        BIO_printf(bio_err, "Error extracting key\n");
+        goto end;
+    }
```
levitte: Ewwww...  I get wanting to just hack it until it builds, but what you end up doing is creating something that approaches `genpkey` but in a half baked way.  I strongly feel this is the wrong path.

There's been two ideas floating around how to solve this:

1.  When `no-deprecated` (simulating the future removal), simply disable the deprecated sub-command.  Period.  Refer to `genpkey` in this case, possibly as a sub-command shim, but most of all in documentation.
2.  Parse the arguments and rework them for `genpkey`, then call `genpkey_main()` with them.

CHANGES
paulidale: Sigh.

Done.

apps/gendsa.c
paulidale: Doesn't this come back to the properly deprecating commands question?

Without this change, the tests don't pass.

Catch 22.

paulidale: Also, this is being deprecated in 3.0 and the changes here can be reverted once the command line deprecation is done.

levitte: Some of the tests should simply be skipped in this case.  You have the perfect example on how to do so here:

https://github.com/openssl/openssl/blob/8b6ffd40401bd3b78538cb8d496db0c6926185b0/test/recipes/05-test_bf.t#L18-L20

and here:

https://github.com/openssl/openssl/blob/8b6ffd40401bd3b78538cb8d496db0c6926185b0/test/recipes/05-test_cast.t#L19-L21

This isn't fitting for all of the test recipes, there are some where you need to add a `SKIP:` section around the offending commands.

I can do those if you want and push a commit...

paulidale: I'm about to stop coding for the evening.
I can do it in the morning.

paulidale: This is a bit more complicated, the first tests that fail are some of the old SSL tests and only some of the ones being iterated over (inside a single test) do so.  I don't think we've got a good way of expressing this.

Later in the old SSL tests, other things fail because files from the earlier test haven't been properly created.

It's a big mess, but I'll see what I can do.

levitte:
Side note: the discussions on deprecating in build.info and perl scripts make me think that we probably need to move the mechanism for generating the diverse `OPENSSL_NO_DEPRECATED_x_y` macros should move to Configure, so it generates corresponding `$disabled{'deprecated_x.y'}`, which would allow us to do things like (in build.info):

```
  SOURCE[whatever]=something.c
ENDIF
```

and (in test recipes):

``` perl
plan skip_all => "Blah-di-blah"
    if disabled('deprecated-3.0');
```

This would corresponds exactly to our `OPENSSL_NO_DEPRECATED_3_0` guards in C.

paulidale:
Yes!

Perhaps after the first alpha though.

paulidale:
I do wonder why these things are being discussed so deeply.  This PR is about getting rid of the low level DSA calls.  It does that.  It does a little more by necessity.  As far as I am concerned, the little more isn't in scope.

I'm willing to put it in scope but this will delay the _α1_ date.  Concentrating our efforts towards the _α1_ date is essential at the moment.

mattcaswell:
IMO we should:

1) Not bother converting command line apps that we want to deprecate to use EVP. If we're going to deprecate them anyway just leave them as they are.
2) Fix tests that use those deprecated command line apps in a "no-deprecated" build so that:
    - If the test is there to test the app itself then we just skip those tests. This should be very simple to do.
    - If the test is there to test some other functionality (e.g. just using gendsa to generate a dsa key in order to do some other DSA releated test using EVP) then we should convert those tests to use the preferred alternative (my understanding from @levitte is that there are very few of these)
3) We should seek to do this for all apps that are to be deprecated due to low-level APIs being deprecated, and do it within the α1 timescale. If that is not possible due to time constraints then we should do as many of the deprecations as we can within that timescale and push others to α2.

paulidale:
The problems I encountered were category 2......
Hence my reluctance to fix the tests as part of this.

levitte:
The affected tests are two: test_dsa and test_ssl_old.  I have a commit coming up that deals with that.

levitte:
@paulidale, I've added test fixing commits here: https://github.com/levitte/openssl/tree/paulidale-dep-dsa

Feel free to grab them (I was planning on pushing directly on this branch, but...)

levitte:
How I figured out what to look for?  Easy, just insert `exit(1)` early in sub-command sources you want to disabled, then run the tests.

paulidale:
The CIs will tell, but this approach is looking quite promising....

slontis:
I merged FFC keygen code last night - you may want to rebase if you havent already :)

paulidale:
Already rebased :)

paulidale:
> IF[{- !$disabled{'deprecated-3.0'} -}]
>   SOURCE[whatever]=something.c
> ENDIF

This doesn't work for the command line apps.  Specifically, _apps/progs.c_ and _apps/progs.h_ start changing on a `make update`.

paulidale:
Finally, there and ready for a review.

richsalz:
if you turn on the deprecation does 'make doc-nits' give errors?

paulidale:
Not any more :)

paulidale:
Travis isn't relevant.

paulidale:
Oops, `make doc-nits` does produce errors.  I read that as `make update` for some reason.

paulidale:
`make doc-nits` is spitting out the same diagnostics for both normal and _no-deprecated_ builds.

richsalz:
doc-nits only finds errors in manpages; cmd-nits finds documentation/flags errors and often means a 'for openssl ifdef' needs updating.  Hope this helps.  Otherwise post the errors you see.

paulidale:
This is the output:

```
doc/internal/man3/cms_add1_signing_cert.pod:1: reference to non-existing openssl-cms(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-req(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-req(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-req(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-req(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-ca(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-pkcs12(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-ca(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-ca(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-x509(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-ca(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-ca(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-ca(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-verify(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-req(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-req(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-x509(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-ca(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-req(1)
doc/man1/CA.pl.pod:1: reference to non-existing openssl-pkcs12(1)
doc/man1/openssl-ciphers.pod:1: reference to non-existing openssl-s_client(1)
doc/man1/openssl-ciphers.pod:1: reference to non-existing openssl-s_server(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-ca(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-cms(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-crl(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-dgst(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-dhparam(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-dsa(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-dsaparam(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-ec(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-ecparam(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-enc(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-gendsa(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-genpkey(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-genrsa(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-ocsp(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-passwd(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-pkcs12(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-pkcs7(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-pkcs8(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-pkey(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-pkeyparam(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-pkeyutl(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-rand(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-req(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-rsa(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-rsautl(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-s_client(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-s_server(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-s_time(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-smime(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-speed(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-spkac(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-srp(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-storeutl(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-ts(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-verify(1)
doc/man1/openssl-cmds.pod:1: reference to non-existing openssl-x509(1)
doc/man1/openssl-crl2pkcs7.pod:1: reference to non-existing openssl-pkcs7(1)
doc/man1/openssl-kdf.pod:1: reference to non-existing openssl-pkeyutl(1)
doc/man1/openssl-list.pod:1: reference to non-existing openssl-dgst(1)
doc/man1/openssl-list.pod:1: reference to non-existing openssl-speed(1)
doc/man1/openssl-list.pod:1: reference to non-existing openssl-dgst(1)
doc/man1/openssl-list.pod:1: reference to non-existing openssl-speed(1)
doc/man1/openssl-rehash.pod:1: reference to non-existing openssl-crl(1)
doc/man1/openssl-rehash.pod:1: reference to non-existing openssl-x509(1)
doc/man1/openssl-sess_id.pod:1: reference to non-existing openssl-s_server(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-x509(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-genpkey(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-pkeyparam(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-genpkey(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-pkeyparam(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-genpkey(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-pkey(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-genpkey(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-pkeyutl(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-enc(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-s_server(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-s_time(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-x509(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-x509(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-ca(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-cms(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-crl(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-dgst(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-dhparam(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-dsa(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-dsaparam(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-ec(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-ecparam(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-enc(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-gendsa(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-genpkey(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-genrsa(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-ocsp(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-passwd(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-pkcs12(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-pkcs7(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-pkcs8(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-pkey(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-pkeyparam(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-pkeyutl(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-rand(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-req(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-rsa(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-rsautl(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-s_client(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-s_server(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-s_time(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-smime(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-speed(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-spkac(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-srp(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-storeutl(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-ts(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-verify(1)
doc/man1/openssl.pod:1: reference to non-existing openssl-x509(1)
doc/man1/tsget.pod:1: reference to non-existing openssl-ts(1)
doc/man1/tsget.pod:1: reference to non-existing openssl-ts(1)
doc/man3/BF_encrypt.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/BN_generate_prime.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/CONF_modules_free.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/DES_random_key.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/DH_generate_parameters.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/DSA_generate_parameters.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/DSA_meth_new.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/DSA_size.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/ECDSA_SIG_new.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/EC_GROUP_new.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/ENGINE_add.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/ERR_load_crypto_strings.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/ERR_remove_state.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/ERR_remove_state.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/EVP_DigestInit.pod:1: reference to non-existing openssl-dgst(1)
doc/man3/EVP_DigestSignInit.pod:1: reference to non-existing openssl-dgst(1)
doc/man3/EVP_DigestVerifyInit.pod:1: reference to non-existing openssl-dgst(1)
doc/man3/EVP_SignInit.pod:1: reference to non-existing openssl-dgst(1)
doc/man3/EVP_VerifyInit.pod:1: reference to non-existing openssl-dgst(1)
doc/man3/HMAC.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/HMAC.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/MD5.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/MD5.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/MD5.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/MDC2_Init.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/OBJ_nid2obj.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/OPENSSL_config.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/OpenSSL_add_all_algorithms.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/RAND_add.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/RAND_bytes.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/RAND_cleanup.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/RC4_set_key.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/RIPEMD160_Init.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/RSA_generate_key.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/RSA_print.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/RSA_print.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/RSA_print.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/SHA256_Init.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/SRP_VBASE_new.pod:1: reference to non-existing openssl-srp(1)
doc/man3/SRP_VBASE_new.pod:1: reference to non-existing openssl-srp(1)
doc/man3/SRP_create_verifier.pod:1: reference to non-existing openssl-srp(1)
doc/man3/SRP_user_pwd_new.pod:1: reference to non-existing openssl-srp(1)
doc/man3/SSL_COMP_add_compression_method.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/SSL_CTX_load_verify_locations.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/SSL_CTX_set_options.pod:1: reference to non-existing openssl-dhparam(1)
doc/man3/SSL_CTX_set_srp_password.pod:1: reference to non-existing openssl-srp(1)
doc/man3/SSL_CTX_set_srp_password.pod:1: reference to non-existing openssl-srp(1)
doc/man3/SSL_CTX_set_tlsext_ticket_key_cb.pod:1: reference to non-existing openssl_user_macros(7)
doc/man3/SSL_CTX_set_tmp_dh_callback.pod:1: reference to non-existing openssl-dhparam(1)
doc/man3/SSL_CTX_set_tmp_dh_callback.pod:1: reference to non-existing openssl-dhparam(1)
doc/man3/SSL_CTX_set_tmp_dh_callback.pod:1: reference to non-existing openssl-dhparam(1)
doc/man3/SSL_CTX_set_tmp_ecdh.pod:1: reference to non-existing openssl-ecparam(1)
doc/man3/SSL_get_verify_result.pod:1: reference to non-existing openssl-verify(1)
doc/man3/SSL_get_verify_result.pod:1: reference to non-existing openssl-verify(1)
doc/man3/SSL_set_verify_result.pod:1: reference to non-existing openssl-verify(1)
doc/man3/SSL_set_verify_result.pod:1: reference to non-existing openssl-verify(1)
doc/man3/X509_LOOKUP_hash_dir.pod:1: reference to non-existing openssl-x509(1)
doc/man3/X509_LOOKUP_hash_dir.pod:1: reference to non-existing openssl-crl(1)
doc/man3/X509_STORE_CTX_get_error.pod:1: reference to non-existing openssl-s_client(1)
doc/man3/X509_VERIFY_PARAM_set_flags.pod:1: reference to non-existing openssl-x509(1)
doc/man3/X509_VERIFY_PARAM_set_flags.pod:1: reference to non-existing openssl-x509(1)
doc/man3/X509_check_issued.pod:1: reference to non-existing openssl-verify(1)
doc/man3/X509_verify_cert.pod:1: reference to non-existing openssl-verify(1)
doc/man5/config.pod:1: reference to non-existing openssl-x509(1)
doc/man5/config.pod:1: reference to non-existing openssl-req(1)
doc/man5/config.pod:1: reference to non-existing openssl-ca(1)
doc/man5/x509v3_config.pod:1: reference to non-existing openssl-req(1)
doc/man5/x509v3_config.pod:1: reference to non-existing openssl-ca(1)
doc/man5/x509v3_config.pod:1: reference to non-existing openssl-x509(1)
doc/man7/Ed25519.pod:1: reference to non-existing openssl-speed(1)
doc/man7/SM2.pod:1: reference to non-existing openssl-speed(1)
doc/man7/proxy-certificates.pod:1: reference to non-existing openssl-x509(1)
doc/man7/proxy-certificates.pod:1: reference to non-existing openssl-req(1)
doc/man7/proxy-certificates.pod:1: reference to non-existing openssl-x509(1)
```

And cmd-nits outputs:
```
doc/man1/openssl-asn1parse.pod does not exist
doc/man1/openssl-ciphers.pod does not exist
doc/man1/openssl-crl2pkcs7.pod does not exist
doc/man1/openssl-engine.pod does not exist
doc/man1/openssl-errstr.pod does not exist
doc/man1/openssl-fipsinstall.pod does not exist
doc/man1/openssl-info.pod does not exist
doc/man1/openssl-kdf.pod does not exist
doc/man1/openssl-list.pod does not exist
doc/man1/openssl-mac.pod does not exist
doc/man1/openssl-nseq.pod does not exist
doc/man1/openssl-prime.pod does not exist
doc/man1/openssl-provider.pod does not exist
doc/man1/openssl-rehash.pod does not exist
doc/man1/openssl-sess_id.pod does not exist
doc/man1/openssl-version.pod does not exist
```

There is a bug with cmd-nits:

```
$ make cmd-nits
(cd ../openssl; /usr/bin/perl util/find-doc-nits -c )
./util/opensslwrap.sh: 28: exec: ./util/../apps/openssl: not found
./util/opensslwrap.sh: 28: exec: ./util/../apps/openssl: not found
```

It doesn't work when building out of the source directory.  I'm doing that here because I'm maintaining a normal and a no-deprecated build in parallel.

levitte:
Yeah, find-doc-nits has changed recently, so it's no longer a source only script.  That makes it depend on information from configdata.pm.

I'm tinkering on a fix...

paulidale:
Fix after alpha 1 ????
Time is short enough at the moment.

petrovr:
Quite interesting work. I wonder when was disabled use of key method function?
This is the only way to implement management of "external" keys.
Provider interface is not suitable for this - after review I can not see any useful.

I would like to request EC to be restored.
And before to make OpenSSL as unusable library switch engines to provider interface - just to shows how we to do with our code.

richsalz:
I am surprised that this needs a fix, since the Makefile has a dependency:
```
doc-nits: build_generated
        (cd $(SRCDIR); $(PERL) util/find-doc-nits -n -l -e )
```

richsalz:
find-doc-nits is source-only unless you do the -c flag which is only invoked by the cmd-nits target.

richsalz:
> Yeah, find-doc-nits has changed recently, so it's no longer a source only script. That makes it depend on information from configdata.pm.
> 
> I'm tinkering on a fix...

Perhaps a very simple thing is to add a flag passed from the makefile that says where the pod pages are found.

levitte:
@richsalz, so seem to forget that a good portion of the POD files are generated from .pod.in files.  Among others, *all* the man1 POD files end up in the build tree for exactly that reason, so when you look at the pile of POD files, there are two locations to look at.

Another fact is that we do have a back reference to the source tree from the build tree, as the value `$config{sourcedir}` in `configdata.pm`.  In the kind of mixture of source tree and build tree that we have, the safest is to simply use `configdata.pm` exactly like it's meant to be, as an information database for everything that has to do with the build, which includes where each source file is and where each generated file is.

levitte:
See #11045

levitte:
> Fix after alpha 1 ????

Er, I used a couple of hours on a Saturday, when I really didn't plan on doing any OpenSSL stuff, so...

paulidale:
> Quite interesting work. I wonder when was disabled use of key method function?

Nothing is being disabled.  They are being deprecated which means they are still usable by default.


> This is the only way to implement management of "external" keys.
> Provider interface is not suitable for this - after review I can not see any useful.

There is an `EVP_KEYMGMT` API underway which might help eventually.


> I would like to request EC to be restored.

Which particular functions are required?

paulidale:
Ping for review?

petrovr:
Pauli wrote:
>> Quite interesting work. I wonder when was disabled use of key method function?
> 
> Nothing is being disabled.  They are being deprecated which means they are still usable by default.

It seems to me you miss configuration that lead to effective removal.

> 
> 
>> This is the only way to implement management of "external" keys.
>> Provider interface is not suitable for this - after review I can not see any useful.
> 
> There is an `EVP_KEYMGMT` API underway which might help eventually.

Let avoid use bureaucratic answer.


> 
>> I would like to request EC to be restored.
> 
> Which particular functions are required?

All that could be used in loadable modules.


The new provider interface is not so stable and there will be period of 
time (at least 2 major releases) before to be considered stable.

If we could this mean 5.0. If you would like to deprecate something from 
loadable modules this should be in 5.0.

Except DSA_sign, DSA_verify  and 1 or 2 other functions other 
modification are not acceptable before 5.0.

About DSA_sign and DSA_verify - I agree that those could be deprecated. 
More over they could be removed. The reason for this is that first FIPS 
validation does not allow use of those functions.
So since OpenSSL 0.9.8 (released 2005 ? ) programs must  not use them.


I'm puzzled, OpenSSL team does not follow own roadmap.

Regards,
Roumen Petrov

levitte:
> I'm puzzled, OpenSSL team does not follow own roadmap.

How do you mean?

apps/build.info
```diff
@@ -31,6 +31,10 @@ IF[{- !$disabled{apps} -}]
     SOURCE[openssl]=openssl.rc
   ENDIF
 
+  IF[{- !$disabled{deprecated} || $config{api} < 30000 -}]
```
levitte: If you rebase, you should be able to change this to

```
```

CHANGES
```diff
@@ -56,7 +82,6 @@
 
      Instead applications and extension writers should use the OSSL_PROVIDER
      APIs.
-     [Paul Dale]
```
levitte: Spurious removal detected!

apps/dsa.c
```diff
@@ -8,7 +8,7 @@
  */
 
 #include <openssl/opensslconf.h>
-#ifdef OPENSSL_NO_DSA
+#if defined(OPENSSL_NO_DSA) || defined(OPENSSL_NO_DEPRECATED_3_0)
```
levitte: This file should get an added `#define OPENSSL_SUPPRESS_DEPRECATED`, just like dhparam.c, and the rest of the changes undone.

paulidale: The only substantive change is changing `DSA_print` to `EVP_PKEY_print_private` which it wraps....

Still, easy to undo.

apps/build.info
paulidale: Fixed.

apps/dsa.c
paulidale: Undone.

paulidale:
Review comment addressed.

paulidale:
> All that could be used in loadable modules. The new provider interface is not so stable and there will be period of time (at least 2 major releases) before to be considered stable. If we could this mean 5.0. If you would like to deprecate something from loadable modules this should be in 5.0.

I'm still not sure I understand the concern.

Our policies say a deprecated function can only be removed after five years and only in a major release.  If we deprecate these functions in 3.0, which is due at the end of 2020, they cannot be removed before the end of 2025.  They will all still be present and usable until at least then.

Delaying the deprecation means we have to support and maintain them even longer.

mattcaswell:
>> Nothing is being disabled.  They are being deprecated which means they are still usable by default.
> It seems to me you miss configuration that lead to effective removal.

I don't understand what you mean here. There is no intent to remove anything.

>> This is the only way to implement management of "external" keys.
>> Provider interface is not suitable for this - after review I can not see any useful.
>
> There is an `EVP_KEYMGMT` API underway which might help eventually.
>
> Let avoid use bureaucratic answer.

I fail to see why that is a "bureaucratic" answer. It is the correct one. `EVP_KEYMGMT` should enable you to manage "external" keys. It's still under development so its not all quite there yet (in particular key gen/param gen isn't done). Also not done is load of external keys via STORE. Nonetheless though this is still the intended interface for external keys, so if it can't do what you need it to do (or if you think it won't be able to when its done) then we need to fix it

apps/progs.c
```diff
@@ -13,388 +13,388 @@
...
-    {FT_general, "verify", verify_main, verify_options},
-    {FT_general, "version", version_main, version_options},
-    {FT_general, "x509", x509_main, x509_options},
-#ifndef OPENSSL_NO_MD2
-    {FT_md, "md2", dgst_main},
+    {FT_general, "verify", verify_main, verify_options, NULL},
+    {FT_general, "version", version_main, version_options, NULL},
+    {FT_general, "x509", x509_main, x509_options, NULL},
+#if !defined(OPENSSL_NO_MD2) && !defined(OPENSSL_NO_DEPRECATED_3_0)
+    {FT_md, "md2", dgst_main, NULL, DEPRECATED_NO_ALTERNATIVE},
```
mattcaswell: Why are we deprecating these digests? Our strategy so far has been to deprecate APIs. In this PR we introduce the concept of also deprecating apps that *use* those APIs. But this seems to be going much further by deprecating *algorithms*. So far we haven't deprecated any algorithms - only moved them to the legacy provider.

apps/progs.pl
```diff
@@ -94,22 +94,42 @@
...
+            if ($dep[2]) {
+                print "!defined(OPENSSL_NO_" . uc($dep[2]) . ") && ";
+            }
+            print "!defined(OPENSSL_NO_DEPRECATED_" . $dep[0] . ")";
+            my $dalt = "\"" . $dep[1] . "\"";
+            $str =~ s/NULL/$dalt/;
+            print "\n${str}#endif\n";
         } elsif (grep { $cmd eq $_ } @disablables) {
             print "#ifndef OPENSSL_NO_" . uc($cmd) . "\n${str}#endif\n";
+        } elsif (my $deprecated = $cmd_deprecated{$cmd}) {
```
mattcaswell: Why is the expression the same as the one above!? It seems this will never be true if the one above is true?

apps/progs.pl
```diff
@@ -143,8 +178,53 @@
...
+        "rc5-ofb" =>     [ "3_0", "rc5" ],
+        "des-ecb" =>     [ "3_0", "des" ],
+        "des-ede" =>     [ "3_0", "des" ],
+        "des-cbc" =>     [ "3_0", "des" ],
+        "des-ede-cbc" => [ "3_0", "des" ],
+        "des-cfb" =>     [ "3_0", "des" ],
+        "des-ede-cfb" => [ "3_0", "des" ],
+        "des-ofb" =>     [ "3_0", "des" ],
+        "des-ede-ofb" => [ "3_0", "des" ],
+        "desx" =>        [ "3_0", "des" ],
```
mattcaswell: None of these things should be deprecated. We are not deprecating algorithms.

apps/progs.c
levitte: Hmmm, the algorithms as command get the EVP_MD or EVP_CIPHER via EVP_MD_by_digestname() and EVP_CIPHER_by_ciphername(), no? Doesn't deprecation come into play there?

mattcaswell: Does it? We're not deprecating those functions are we?

levitte: I admit that my recollection is vague, gimme a moment or two to check things out, yeah?

petrovr:
Pauli wrote:
>> All that could be used in loadable modules. The new provider interface is not so stable and there will be period of time (at least 2 major releases) before to be considered stable. If we could this mean 5.0. If you would like to deprecate something from loadable modules this should be in 5.0.
> 
> I'm still not sure I understand the concern.
> 
> Our policies say a deprecated function can only be removed after five years and only in a major release.  If we deprecate these functions in 3.0, which is due at the end of 2020, they cannot be removed before the end of 2025.  They will all still be present and usable until at least then.
> 
> Delaying the deprecation means we have to support and maintain them even longer.
> 
> 

Common   --api=3.0.0  no-deprecated effectively removes today !

Roumen

petrovr:
Richard Levitte wrote:
>> I'm puzzled, OpenSSL team does not follow own roadmap.
> 
> How do you mean?
> 

s/roadmap/Strategic Architecture/

apps/progs.pl
levitte: [Reopening this conversation after seeing comments from Matt on the subject.]

It turns out that these deprecations are premature.

apps/progs.pl
```diff
@@ -143,8 +178,53 @@
...
     );
```
levitte: It turn out that these deprecations are premature too.

levitte:
> Common   --api=3.0.0  no-deprecated effectively removes today !

I will say this again: those are not default configuration options, *and they are not recommended for general use and cannot be considered "normal" use*.  The intention with `no-deprecated` is to check if your application can build with all deprecated symbols removed, for those who really want to be that much on the bleeding edge.  If that's not your application, *then do not use `no-deprecated`*, it's as simple as that.

mattcaswell:
> Common --api=3.0.0 no-deprecated effectively removes today !

no-deprecated removes all deprecated symbols. If we attempted to ensure that a no-deprecated build was backwards compatible then we could never deprecate anything! That was never the intention. Don't use no-deprecated if you care about this.

apps/progs.pl
paulidale: My error.  Gone now.

apps/progs.pl
paulidale: Okay, should I also remove the Perl changes associated with these or leave them for future use?

paulidale:
Comments addressed.

apps/progs.pl
levitte: Hmmm, just so we don't undo the framework you've worked out, I would keep the empty hash tables. Possibly with a comment.

paulidale: Okay, I've done this.

mattcaswell: I  actually think this should all be removed. This is now dead code. If it gets resurrected any time soon then I'm sure you'll keep a copy locally. Otherwise it will just stick around in the code not doing anything.

paulidale: It wasn't difficult to write even for a Perl-denier, so I'm easy either way.

paulidale: I've pushed a fixup with this removed.  I can unroll it if that's the consensus.

petrovr:
Richard Levitte wrote:
>> Common   --api=3.0.0  no-deprecated effectively removes today !
> 
> I will say this again: those are not default configuration options, *and they are not recommended for general use and cannot be considered "normal" use*.  The intention with `no-deprecated` is to check if your application can build with all deprecated symbols removed, for those who really want to be that much on the bleeding edge.  If that's not your application, *then do not use `no-deprecated`*, it's as simple as that.
> 

And the error is in design - some symbols are marked depreciated without 
alternative.
And this is my point - to mark as deprecated in 5.0 when is expected to 
exist valuable alternative.

Regards
Roumen Petrov

levitte:
> And this is my point - to mark as deprecated in 5.0 when is expected to exist valuable alternative.

Huh?  Of course, it depends on what you mean with "valuable", but the EVP API with provider backing *is* the alternative.

mattcaswell:
> And the error is in design - some symbols are marked depreciated without
alternative.

If you believe that then I'd like to understand what symbols you think they are. AFAIK all symbols that we are deprecating have an alternative.

paulidale:
Merged to master.  Thanks for the reviews and feedback.

PR #10980, opened 2020-01-31
Don't ignore ASN1 when checking for undocumented symbols

mattcaswell:
When we run "make doc-nits" (which happens during travis runs) it will
complain if we add any new symbols that aren't documented. However it
was suppressing anything starting with ASN1. There's no reason why we
should allow ASN1 symbols to go undocumented any more than any others.
Therefore we remove that exception.

##### Checklist


levitte:
H'oh boy...

Did you find a reasoning re skipping those functions in the commit log?

mattcaswell:
> Did you find a reasoning re skipping those functions in the commit log?

Seems to have always been there. No specific reasoning mentioned in the logs that I could find.

richsalz:
We never considered the ASN1 parsing to be really part of the public API.  Nobody on the team was really thrillled with the way that stuff works, and there was also a bubbling undercurrent of replacing it.  When developing find-doc-nits, I just codified that sentiment :)  Perhaps a more tractable solution is to just ignore the common new/free/it names?  And then trim down the missing list significantly.

mattcaswell:
Ping?

paulidale:
Knowing which of these isn't documented seems like a good idea.

mattcaswell:
Darn. PR #10942 went in and added a new and undocumented ASN1 function with the promise from @levitte that he would add the docs real soon. Until that happens we need to add that function to missingcrypto.txt in this PR to avoid a "make doc-nits" failure.

@paulidale (or anyone) please can you reconfirm?

mattcaswell:

Ping?

mattcaswell:
Pushed. Thanks.

PR #10981, opened 2020-01-31
Don't complain about documented symbols with find-doc-nits -e -o

mattcaswell:
find-doc-nits can give a list of symbols that were added since 1.1.1 and
are undocumented (using -o). To do this it uses the missingcrypto111.txt
and missingssl111.txt files which give a snapshot of the undocumented
symbols at the time of the 1.1.1 release. Currently it complains about
symbols that are in those files that have subsequently been documented.
This isn't particularly helpful so we suppress that feature when "-o"
is being used.

##### Checklist

### DIFF
--- a/util/find-doc-nits
+++ b/util/find-doc-nits
@@ -607,7 +607,7 @@ sub loadmissing($)
 
     for (@missing) {
         err("$missingfile:", "$_ is documented in $name_map{$_}")
-            if exists $name_map{$_} && defined $name_map{$_};
+            if !$opt_o && exists $name_map{$_} && defined $name_map{$_};
     }
 
     return @missing;


richsalz:
Is it time to just get rid of -o etc options? Or simplify them?  Just make it self-contained and not look at the 1.1.1 list?

mattcaswell:
> Is it time to just get rid of -o etc options? Or simplify them? Just make it self-contained and not look at the 1.1.1 list?

Err - no? It's useful at least until we get to the point where there are no additional symbols in missingcrypto/ssl/macros that aren't in the 1.1.1 versions. By the time of the 3.0 release we should have resolved all of those (see #9095) - and only *then* can we remove "-o".

mattcaswell:
Pushed. Thanks.

PR #10982, opened 2020-01-31
x86_64: Add endbranch at function entries for Intel CET

hjl-tools:
To support Intel CET, all indirect branch targets must start with
endbranch.  Here is a patch to add endbranch to function entries
in x86_64 assembly codes which are indirect branch targets as
discovered by running openssl testsuite on Intel CET machine and
visual inspection.

Verified with

$ CC="gcc -Wl,-z,cet-report=error" ./Configure shared linux-x86_64 -fcf-protection
$ make
$ make test

and

$ CC="gcc -mx32 -Wl,-z,cet-report=error" ./Configure shared linux-x32 -fcf-protection
$ make
$ make test # <<< passed with https://github.com/openssl/openssl/pull/10988

### DIFF
--- a/crypto/aes/asm/aes-x86_64.pl
+++ b/crypto/aes/asm/aes-x86_64.pl
@@ -606,6 +606,7 @@ $code.=<<___;
 asm_AES_encrypt:
 AES_encrypt:
 .cfi_startproc
+	endbranch
 	mov	%rsp,%rax
 .cfi_def_cfa_register	%rax
 	push	%rbx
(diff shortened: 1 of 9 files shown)


levitte:
> Could you please re-approve after the force push?

If nothing changed in *this PR*, then my earlier approval stands.

levitte:
Merged

98ad3fe82bd3e7e7f929dd1fa4ef3915426002c0 x86_64: Add endbranch at function entries for Intel CET

PR #10983, opened 2020-01-31
Use swapcontext for Intel CET

hjl-tools:
When Intel CET is enabled, makecontext will create a different shadow
stack for each context.  async_fibre_swapcontext cannot use _longjmp.
It must call swapcontext to swap shadow stack as well as normal stack.

### DIFF
--- a/crypto/async/arch/async_posix.h
+++ b/crypto/async/arch/async_posix.h
@@ -25,17 +25,33 @@
 #  define ASYNC_POSIX
 #  define ASYNC_ARCH
 
+#  ifdef __CET__
+/*
+ * When Intel CET is enabled, makecontext will create a different
+ * shadow stack for each context.  async_fibre_swapcontext cannot
+ * use _longjmp.  It must call swapcontext to swap shadow stack as
+ * well as normal stack.
+ */
+#   define USE_SWAPCONTEXT
+#  endif
 #  include <ucontext.h>
-#  include <setjmp.h>
+#  ifndef USE_SWAPCONTEXT
+#   include <setjmp.h>
+#  endif
 
 typedef struct async_fibre_st {
     ucontext_t fibre;
+#  ifndef USE_SWAPCONTEXT
     jmp_buf env;
     int env_init;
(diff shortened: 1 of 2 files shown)


crypto/async/arch/async_posix.h
```diff
@@ -25,17 +25,33 @@
...
 #  define ASYNC_ARCH
 
+#  ifdef __CET__
+/*
+ * When Intel CET is enabled, makecontext will create a different
+ * shadow stack for each context.  async_fibre_swapcontext cannot
+ * use _longjmp.  It must call swapcontext to swap shadow stack as
+ * well as normal stack.
+ */
+#    define USE_SWAPCONTEXT
```
mattcaswell: Nit: Too many indents here. We indent by one space for each additional pre-processor block - so there should be 3 spaces here.

crypto/async/arch/async_posix.h
```diff
@@ -25,17 +25,33 @@
...
 
 typedef struct async_fibre_st {
     ucontext_t fibre;
+#    ifndef USE_SWAPCONTEXT
     jmp_buf env;
     int env_init;
+#    endif
 } async_fibre;
 
 static ossl_inline int async_fibre_swapcontext(async_fibre *o, async_fibre *n, int r)
```
mattcaswell: Side note: As far as I can tell the `r` parameter is unused. Commit 7070e5ca2 originally changed the code from using `swapcontext` to `longjmp` on the basis of performance . The code used to just call `swapcontext` like you are changing it back to here - except it did something different in the case of `r == 0`. However, as far as i can tell *all* places where we call `async_fibre_swapcontext` set r to 1 - so presumbably this could get refactored away at some point.

crypto/async/arch/async_posix.h
hjl-tools: Fixed.

mattcaswell:
Needs second review.

mattcaswell:
Pushed. Thanks.

PR #10984, opened 2020-01-31
x86: Add endbranch to indirect branch targets for Intel CET

hjl-tools:
To support Intel CET, all indirect branch targets must start with
endbranch.  Here is a patch to add endbranch to all function entries
in x86 assembly codes which are indirect branch targets as discovered
by running openssl testsuite on Intel CET machine and visual inspection.

Since x86 cbc.pl uses indirect branch with a jump table, we also need
to add endbranch to all jump targets.


paulidale:
Merged to master.  Thanks for the submission.

Issue #10896, opened 2020-01-19
Intel CET support is incomplete


hjl-tools:
commit b2b43d1b698235ee674e588c224db38d9b78d1ea
Author: Richard Levitte <levitte@openssl.org>
Date:   Fri Jan 17 08:29:28 2020 +0100

    Add GNU properties note for Intel CET in x86_64-xlate.pl

has 2 issues.   Intel CET has 2 parts, shadow stack (SHSTK) and indirect branch tracking
(IBT).  Openssl is SHSTK compatible.  But ENDBR is needed at indirect branch targets for IBT.

1,  We need to mark SHSTK and IBT compatibility in GNU properties note for every object file,
not just those with ENDBR.   You can verify it on Fedora 31 with

$ CC="gcc -Wl,-z,cet-report=error" ./Configure shared linux-x86_64 -fcf-protection
$ make
$ make test

If you linker doesn't support

  -z cet-report=[none|warning|error] (default: none)

you need to install GNU binutils 2.33.  I got

/usr/local/bin/ld: crypto/aes/libcrypto-shlib-aes-x86_64.o: error: missing IBT and SHSTK properties
/usr/local/bin/ld: crypto/aes/libcrypto-shlib-aesni-mb-x86_64.o: error: missing IBT and SHSTK properties
/usr/local/bin/ld: crypto/aes/libcrypto-shlib-aesni-sha1-x86_64.o: error: missing IBT and SHSTK properties
/usr/local/bin/ld: crypto/aes/libcrypto-shlib-aesni-sha256-x86_64.o: error: missing IBT and SHSTK properties
/usr/local/bin/ld: crypto/aes/libcrypto-shlib-aesni-x86_64.o: error: missing IBT and SHSTK properties
/usr/local/bin/ld: crypto/aes/libcrypto-shlib-bsaes-x86_64.o: error: missing IBT and SHSTK properties
/usr/local/bin/ld: crypto/aes/libcrypto-shlib-vpaes-x86_64.o: error: missing IBT and SHSTK properties
/usr/local/bin/ld: crypto/bn/libcrypto-shlib-rsaz-avx2.o: error: missing IBT and SHSTK properties
/usr/local/bin/ld: crypto/bn/libcrypto-shlib-rsaz-x86_64.o: error: missing IBT and SHSTK properties
/usr/local/bin/ld: crypto/bn/libcrypto-shlib-x86_64-gf2m.o: error: missing IBT and SHSTK properties
[...]

PR #10985, opened 2020-01-31
x86_64: Always generate .note.gnu.property section for ELF outputs

hjl-tools:
This reverts commit b2b43d1b698235ee674e588c224db38d9b78d1ea.

This patch is incorrect since .note.gnu.property section is ELF only.
When endbranch is used, .note.gnu.property section is generated for
all targets.  On MacOS, we got

https://travis-ci.org/openssl/openssl/jobs/644331421?utm_medium=notification&utm
_source=github_status

crypto/aes/aes-x86_64.s:2641:11: error: mach-o section specifier requires a segm
ent whose length is between 1 and 16 characters
.section ".note.gnu.property", "a"

### DIFF
--- a/crypto/perlasm/x86_64-xlate.pl
+++ b/crypto/perlasm/x86_64-xlate.pl
@@ -101,24 +101,33 @@ elsif (!$gas)
     $decor="\$L\$";
 }
 
-my $cet_property = <<'_____';
+my $cet_property;
+if ($flavour =~ /elf/) {
+	# Always generate .note.gnu.property section for ELF outputs to
+	# mark Intel CET support since all input files must be marked
+	# with Intel CET support in order for linker to mark output with
+	# Intel CET support.
+	my $p2align=3; $p2align=2 if ($flavour eq "elf32");
+	$cet_property = <<_____;
 	.section ".note.gnu.property", "a"
-	.align 8
+	.p2align $p2align
 	.long 1f - 0f
 	.long 4f - 1f
 	.long 5
 0:
 	.asciz "GNU"
 1:
-	.align 8
+	.p2align $p2align
 	.long 0xc0000002
 	.long 3f - 2f
 2:
 	.long 3
 3:
-	.p2align 3
+	.p2align $p2align
 4:
 _____
+}
+
 my $current_segment;
 my $current_function;
 my %globals;
@@ -1145,9 +1154,7 @@ my $vprotq = sub {
 # Intel Control-flow Enforcement Technology extension. All functions and
 # indirect branch targets will have to start with this instruction...
 
-my $used_cet = 0;
 my $endbranch = sub {
-    $used_cet = 1;
     (0xf3,0x0f,0x1e,0xfa);
 };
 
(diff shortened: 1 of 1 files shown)


hjl-tools:
@levitte Adding Richard Levitte

hjl-tools:
We should always generate .note.gnu.property section in x86_64 assembly
codes for ELF outputs to mark Intel CET support since all input files
must be marked with Intel CET support in order for linker to mark output
with Intel CET support.  Also .note.gnu.property section in x32 should
be aligned to 4 bytes, not 8 bytes and .p2align should be used
consistently.

Verified with

$ CC="gcc -Wl,-z,cet-report=error" ./Configure shared linux-x86_64 -fcf-protecti
on
$ make
$ make test

and

$ CC="gcc -mx32 -Wl,-z,cet-report=error" ./Configure shared linux-x32 -fcf-prote
ction
$ make
$ make test # <<< 90-test_sslapi.t failed because 8-byte pointer size.

Fix #10896

levitte:
I wonder, what happens if you build OpenSSL *without* `-fcf-protection`?  I assume that with `-Wl,-z,cet-report=error`, it will go through silently, but are there subtleties we need to be aware of with a mixture of object files that have the added GNU properties and object files that don't?

hjl-tools:
> I wonder, what happens if you build OpenSSL _without_ `-fcf-protection`? I assume that without `-Wl,-z,cet-report=error`, it will go through silently, but are there subtleties we need to be aware of with a mixture of object files that have the added GNU properties and object files that don't?

```
$ CC="gcc -Wl,-z,cet-report=error" ./Configure shared linux-x86_64
...
$ make
...
gcc -Wl,-z,cet-report=error -fPIC -pthread -m64 -Wa,--noexecstack -Wall -O3 -Wl,-znodelete -shared -Wl,-Bsymbolic   -Wl,-soname=libcrypto.so.3 \
        -o libcrypto.so.3 -Wl,--version-script=libcrypto.ld \
        crypto/aes/libcrypto-shlib-aes-x86_64.o \
        crypto/aes/libcrypto-shlib-aes_cfb.o \
        crypto/aes/libcrypto-shlib-aes_ecb.o \
        crypto/aes/libcrypto-shlib-aes_ige.o \
```        providers/implementations/asymciphers/libcrypto-shlib-rsa_enc.o \
        providers/libcrypto-shlib-defltprov.o \
        providers/libimplementations.a providers/libcommon.a providers/libnonfips.a  -ldl -pthread
/usr/local/bin/ld: crypto/aes/libcrypto-shlib-aes-x86_64.o: error: missing IBT and SHSTK properties
/usr/local/bin/ld: crypto/aes/libcrypto-shlib-aes_cfb.o: error: missing IBT and SHSTK properties
/usr/local/bin/ld: crypto/aes/libcrypto-shlib-aes_ecb.o: error: missing IBT and SHSTK properties
/usr/local/bin/ld: crypto/aes/libcrypto-shlib-aes_ige.o: error: missing IBT and SHSTK properties
...
```

Mixing object files with and without GNU property is OK.   Older linkers treat them like
input files with just another NOTE section.  Newer linkers will drop GNU property in
output.

t8m:
Pushed to master as 51994e5

PR #10989, opened 2020-01-31
Revert "Legacy digests can have custom control values"

mattcaswell:
This reverts commit 1f457256ce6a1b2fd7e3f62eee8faa74cd5c835e.

This is causing Travis failures.

[extended tests]

### DIFF
--- a/crypto/evp/digest.c
+++ b/crypto/evp/digest.c
@@ -663,7 +663,7 @@ int EVP_MD_CTX_ctrl(EVP_MD_CTX *ctx, int cmd, int p1, void *p2)
     }
 
     if (ctx->digest->prov == NULL
-        || (ctx->pctx == NULL
+        && (ctx->pctx == NULL
             || (ctx->pctx->operation != EVP_PKEY_OP_VERIFYCTX
                 && ctx->pctx->operation != EVP_PKEY_OP_SIGNCTX)))
         goto legacy;


slontis:
Hmm I thought that fix was put in to fix other issues.. (So something else will be broken - I think this was needed by @beldmit for GOST)??

mattcaswell:
> Hmm I thought that fix was put in to fix other issues.. (So something else will be broken - I think this was needed by @beldmit for GOST)??

Yes, it was. But it breaks lots of stuff, so I think we need to back it out and have a rethink.

This fixes most of the Travis test failures we've been seeing (ssl_test_new, ssl_test_old, boringssl). It does not fix the krb5 test failure which appears to be an unrelated issue.

slontis:
From what was sent around the problem seemed to crop up when the modp changes went in, so I thought I had done something wrong.. Not sure I understand what that means..

mattcaswell:
No, the krb5 issue predates modp. The modp changes did introduce another problem though. See #10990.

beldmit:
> Hmm I thought that fix was put in to fix other issues.. (So something else will be broken - I think this was needed by @beldmit for GOST)??

I can rewrite my engine to avoid this pitfall, but it's a matter of policy.

mattcaswell:
> I can rewrite my engine to avoid this pitfall, but it's a matter of policy.

We should investigate whether there is another solution to the problem that does not require changes to the engine. But for now I think we revert in order to fix the tests.

beldmit:
Sure, not a problem at all. BTW, should I reopen the original report, create new or do anything else?

mattcaswell:
I'd suggest creating a new one when we merge this PR.

levitte:
Why isn't it enough to check `ctx->digest->prov`? It seems that we have somewhat over-complex checks...

mattcaswell:
Pushed. Thanks!

PR #10990, opened 2020-01-31
Fix builds with no-dh

mattcaswell:
The various functions in bn_const.c return primes that are
specified for use in DH. However they were not being excluded from
a no-dh build - and was therefore causing the build to fail.

### DIFF
--- a/crypto/bn/build.info
+++ b/crypto/bn/build.info
@@ -101,12 +101,18 @@ IF[{- !$disabled{asm} -}]
   ENDIF
 ENDIF
 
+IF[{- !$disabled{dh} -}]
+  $BNDH=bn_const.c
+ELSE
+  $BNDH=
+ENDIF
+
 $COMMON=bn_add.c bn_div.c bn_exp.c bn_lib.c bn_ctx.c bn_mul.c \
         bn_mod.c bn_conv.c bn_rand.c bn_shift.c bn_word.c bn_blind.c \
         bn_kron.c bn_sqrt.c bn_gcd.c bn_prime.c bn_sqr.c \
         bn_recp.c bn_mont.c bn_mpi.c bn_exp2.c bn_gf2m.c bn_nist.c \
-        bn_const.c bn_x931p.c bn_intern.c bn_dh.c \
-        bn_rsa_fips186_4.c $BNASM
+        bn_x931p.c bn_intern.c bn_dh.c \
+        bn_rsa_fips186_4.c $BNDH $BNASM
 SOURCE[../../libcrypto]=$COMMON bn_print.c bn_err.c bn_depr.c bn_srp.c
 SOURCE[../../providers/libfips.a]=$COMMON
 # Implementations are now spread across several libraries, so the defines
--- a/include/openssl/bn.h
+++ b/include/openssl/bn.h
@@ -527,6 +527,7 @@ int BN_generate_dsa_nonce(BIGNUM *out, const BIGNUM *range,
                           const BIGNUM *priv, const unsigned char *message,
                           size_t message_len, BN_CTX *ctx);
 
+# ifndef OPENSSL_NO_DH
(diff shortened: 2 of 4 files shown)


slontis:
Hmm is there a build that does no-dsa and no-dh together..
It looks like hextobn also needs to be #ifdefed if they are both not defined

Hmmm well at least this one is a simple breakage..
As this is happening quite a lot - so dont you think it would be worth running a travis build with a whole lot of -no options so this stops happening (not as part of extended tests).
We should also enable tracing on one of the builds (I tried to before but it was shot down due to speed of build reasons). The counter argument is that it saves us (you) time.

crypto/bn/build.info
```diff
@@ -101,12 +101,18 @@ IF[{- !$disabled{asm} -}]
   ENDIF
 ENDIF
 
+IF[{- !$disabled{dh} -}]
```
slontis: I assume this bit wasnt actually causing a problem, and you are just stripping constants out becuase they are not needed if DH is not used.
I wonder if you could use them for DSA (they are just p, q, g values).

mattcaswell: It was causing a problem. The functions in bn_const.c were being compiled unconditionally, but the constants that they referred to (which are in another file) were compiled out if no-dh was specified. Therefore we were ending up with linker errors. I chose to not compile the functions in the event of no-dh. An alternative solution would be to keep the constants compiled in, even in a no-dh build. However, I checked the RFCs and they state that they are intended for use in Diffie Hellman.

mattcaswell:
>It looks like hextobn also needs to be #ifdefed if they are both not defined in fipsprov.c

Fixed

mattcaswell:
Ping.

paulidale:
Can we get a list together of all the hex converters and make an attempt to unify them?

mattcaswell:
Probably this should have been labelled as "urgent" as we have done with the other run-checker break fixing PRs. I have done so now, but won't immediately merge in case of objections. I will merge later.

mattcaswell:
(It would be good to get confirmation from either @levitte or @t-j-h that you agree with the urgent label)

t-j-h:
I don't see a need to mark any of the current PRs that are marked with "urgent" as urgent as none of them are urgent in the sense of the definition of the word. I don't see a need to bypass the normal 24-hour wait period. 
If you overuse an exception process then it looses the exception feel.

Also the label changes don't go out in email notification - so it would be best to do a comment like you have noting you added that label in case others want to object - although the concept of urgent means bypass the wait so that you are indicating waiting is okay in this context it really doesn't make sense to actually use. At least that is the logic chain as I read through all these "urgent" items.

mattcaswell:
> I don't see a need to mark any of the current PRs that are marked with "urgent" as urgent as none of them are urgent in the sense of the definition of the word. I don't see a need to bypass the normal 24-hour wait period.

These are build breaks that break our run-checker builds - which we should be seeking to keep clean (in much the same was as with Travis).

> Also the label changes don't go out in email notification - so it would be best to do a comment like you have noting you added that label in case others want to object - although the concept of urgent means bypass the wait so that you are indicating waiting is okay in this context it really doesn't make sense to actually use. At least that is the logic chain as I read through all these "urgent" items.

The difference is "short wait" vs "long wait". In the case of an urgent label I would expect at least one other person to agree to it, and no one to object to it. Ideally the urgent label should be applied at the time the PR is created so that the reviewer can implicitly agree to it during the review. In this case that didn't occur, so I added it later and asked for  someone else to take a look and agree to it.

For now I will hold off on applying any of the PRs labelled as urgent since you have objected to it - although that will likely delay the time it will take to get run-checker clean again. It seems OTC (or is it OMC?) should issue more detailed guidance on how to interpret the build/test break 24-hour exception since there appears to be a difference of opinion here.

mattcaswell:
Pushed. Thanks.

PR #10991, opened 2020-02-01
Fix no-ec

mattcaswell:
The cmp_protect_test cert chain tests use some EC certs which breaks in
a no-ec build. The fix is to just skip those tests if no-ec has been
configured.


levitte:
What branches?

mattcaswell:
master only

mattcaswell:
Pushed. Thanks.

PR #10913, opened 2020-01-21
OpenSSL::Test: bring back the relative paths

levitte:
Because there was a bug in File::Spec::Unix' abs2rel when it was given
relative paths as both PATH and BASE arguments, the directories we
deal with were made to be all absolute.  Unfortunately, this meant
getting paths in our verbose test output which are difficult to use
anywhere else (such as a separate test build made for comparison), due
to the constant need to edit all the paths all the time.

We're therefore getting back the relative paths, by doing an extra
abs2rel() in __srctop_file, __srctop_dir, __bldtop_file and
__bldtop_dir, with a 'Cwd::getcwd' call as BASE argument.

Fixes #10628

PR #7031, opened 2018-08-22
Test.pm workaround mitigating Perl abs2rel bug

mpeylo:
If SRCTOP != BLDTOP, and SRCTOP is given in relative form, e.g.
"./config ../openssl", then a bug in Perl's abs2rel may trigger that directory-
rewriting in __cwd results in wrong entries in %directories under certain
circumstances, e.g. when a test executes run(app(["openssl"]) after indir.

There should not be any need to go to a higher directory from BLDDIR or SRCDIR,
so it should be OK to use them in their absolute form, also resolving all
possible symlinks, right from the start.

Following the File::Spec::Functions bug description (reported to perl.org):

When abs2rel gets a path argument with ..s that are crossing over the ..s
trailing the base argument, the result is wrong.

Example
PATH: /home/goal/test/..
BASE: /home/goal/test/../../base
Good result: ../goal
Bad  result: ../..

Bug verified with File::Spec versions
- 3.6301
- 3.74 (latest)

##### Checklist

- [ ] documentation is added or updated
- [ ] tests are added or updated

PR #10992, opened 2020-02-01
Fix for travis

mattcaswell:
This includes the commit from #10989 and another commit to revert the change made in #10913 to see if that can resolve the travis issues. (Running with extended tests)

### DIFF
--- a/test/recipes/95-test_external_krb5.t
+++ b/test/recipes/95-test_external_krb5.t
@@ -10,6 +10,7 @@
 use OpenSSL::Test;
 use OpenSSL::Test::Utils;
 use OpenSSL::Test qw/:DEFAULT data_file srctop_file bldtop_dir/;
+use Cwd qw(abs_path);
 
 setup("test_external_krb5");
 
@@ -20,6 +21,7 @@ plan skip_all => "krb5 not available"
 
 plan tests => 1;
 
-$ENV{OPENSSL_CONF} = srctop_file("test", "default-and-legacy.cnf");
+$ENV{OPENSSL_MODULES} = abs_path($ENV{OPENSSL_MODULES});
+$ENV{OPENSSL_CONF} = abs_path(srctop_file("test", "default-and-legacy.cnf"));
 
 ok(run(cmd([data_file("krb5.sh")])), "running krb5 tests");


mspncp:
How did #10913 break the extended tests? 

Pull request #10913 tried to fix an annoying side effect of #7031, which I still consider unnecessary unless somebody actually gives evidence that the Perl ab2rel issue indeed affected the OpenSSL tests. So if pull request #10913 needs to be reverted, could you please check whether travis remains green after reverting pull request #7031 as well?

mspncp:
> How did #10913 break the extended tests?

Do you think there is something wrong with the relative pathes in the krb5 tests?

https://travis-ci.org/openssl/openssl/jobs/644583874#L2712

mspncp:
If you think this is the case, it might help to set the `$debug` variable in Test.pm to 1 

https://github.com/openssl/openssl/blob/1e628153ff826e04f64d029a4b6e61eb681658e3/util/perl/OpenSSL/Test.pm#L106

in order to get more detailed output:

https://github.com/openssl/openssl/blob/1e628153ff826e04f64d029a4b6e61eb681658e3/util/perl/OpenSSL/Test.pm#L1154-L1167

mattcaswell:
> Do you think there is something wrong with the relative pathes in the krb5 tests?

I really have no idea. My fix was based on the observation that the  krb5 test started failing when that commit went in. My trial fix seems to confirm that theory - the build was green with the two reverted commits in this PR and extended tests.

I've changed the debug flag to on as suggested and we'll see what happens.

mattcaswell:
Of course...now I realise that turning that debug flag on in this PR isn't going to tell us much...we need to do it in a PR that doesn't have that suspicious commit reverted!!

mspncp:
> Of course...now I realise that turning that debug flag on in this PR isn't going to tell us much...we need to do it in a PR that doesn't have that suspicious commit reverted!!

Can you reproduce it locally by running the krb5 tests explicitly? That would make it easier to debug the issue. I would love to help with the relpath issue, but this weekend I am very busy with private things.

BTW: Thank you for your efforts to fix the travis builds!

mattcaswell:
I will try, but it is unlikely to happen before Monday.

mspncp:
> I will try, but it is unlikely to happen before Monday.

Sure, it's weekend ;-) I'll see what I can do in the meantime. 


BTW: I did some tests for the relpath pr recently. Maybe https://github.com/openssl/openssl/pull/10913#pullrequestreview-348334299 can provide some helpful information.

mspncp:
Strange: It just occured to me that @levitte's commit was cherry-picked to 1.1.1 in c8943eb04ae41506ec8a1869103e2e1dec6bfcd8, but the krb5 test [succeeds on 1.1.1](https://travis-ci.org/openssl/openssl/jobs/643941490).

levitte:
It seems real strange that the krb5 tests would be sensitive to relative paths. Unless they get blatantly wrong, it sounds like a red herring

mspncp:
> It seems real strange that the krb5 tests would be sensitive to relative paths. Unless they get blatantly wrong, it sounds like a red herring

I'm afraid in this case you are wrong...


I ran the krb5 tests on three different commits (bottom up)

3. 184e21dd61 Revert "Adding Test.pm with workaround for Perl abs2rel bug"
2. a22b9085ee Revert "OpenSSL::Test: bring back the relative paths"
1. 6ca2dd76b7 Add some prints (otherwise, identical with current master)

Commits 1 and 3 use both relative pathes and the middle commit 2 uses absolute pathes, like in the original fix of @mpeylo. In all three cases I tested using the command

```
make V=1 TESTS=test_external_krb5 test |& tee krb5-test.log
```

Both 'relative' tests 1 and 3 fail, their output is only slighty different:

```
msp@debian:~/src/openssl$ git --no-pager diff 6ca2dd76b7..184e21dd61 -- krb5-test.log
diff --git a/krb5-test.log b/krb5-test.log
index a651c8a607..64527a7319 100644
--- a/krb5-test.log
+++ b/krb5-test.log
@@ -21,23 +21,23 @@ BLDTOP: ../. = /home/msp/src/openssl
 run_tests: srctop =../.
 run_tests: bldtop =../.
 95-test_external_krb5.t .. 
-SRCTOP =/home/msp/src/openssl
-BLDTOP =/home/msp/src/openssl
+SRCTOP =../.
+BLDTOP =../.
 DEBUG: __cwd(), directories and files:
-  $directories{BLDTEST} = ".."
-  $directories{SRCTEST} = ".."
-  $directories{SRCDATA} = "../recipes/95-test_external_krb5_data"
+  $directories{BLDTEST} = "../../test"
+  $directories{SRCTEST} = "../../test"
+  $directories{SRCDATA} = "../../test/recipes/95-test_external_krb5_data"
   $directories{RESULTS} = "."
   $directories{BLDAPPS} = "../../apps"
   $directories{SRCAPPS} = "../../apps"
-  $directories{SRCTOP}  = "/home/msp/src/openssl"
-  $directories{BLDTOP}  = "/home/msp/src/openssl"
+  $directories{SRCTOP}  = "../.."
+  $directories{BLDTOP}  = "../.."
 
   current directory is "."
   the way back is ".."
 1..1
-DEBUG[__decorate_cmd]: $cmdstr = "../../util/shlib_wrap.sh ../recipes/95-test_external_krb5_data/krb5.sh"
-DEBUG[__decorate_cmd]: $display_cmd = "../../util/shlib_wrap.sh ../recipes/95-test_external_krb5_data/krb5.sh"
+DEBUG[__decorate_cmd]: $cmdstr = "../../util/shlib_wrap.sh ../../test/recipes/95-test_external_krb5_data/krb5.sh"
+DEBUG[__decorate_cmd]: $display_cmd = "../../util/shlib_wrap.sh ../../test/recipes/95-test_external_krb5_data/krb5.sh"
 + set -o xtrace
 + pwd
 + pwd
@@ -1808,7 +1808,7 @@ if test -d ../../../include/gssapi; then :; else (set -x; mkdir ../../../include
 + mkdir ../../../include/gssapi
 echo timestamp > ../../../include/gssapi/timestamp
 Creating gssapi.h
-+ mv -f gss83834 gssapi.h
++ mv -f gss101658 gssapi.h
 cp gssapi.h ../../../include/gssapi/gssapi.h
 cp ./gssapi_generic.h ../../../include/gssapi/gssapi_generic.h
 cp ./gssapi_alloc.h ../../../include/gssapi/gssapi_alloc.h
@@ -3995,7 +3995,7 @@ make[3]: *** [Makefile:992: check-recurse] Error 1
 make[3]: Leaving directory '/home/msp/src/openssl/krb5/src/lib'
 make[2]: *** [Makefile:1533: check-recurse] Error 1
 make[2]: Leaving directory '/home/msp/src/openssl/krb5/src'
-../../util/shlib_wrap.sh ../recipes/95-test_external_krb5_data/krb5.sh => 2
+../../util/shlib_wrap.sh ../../test/recipes/95-test_external_krb5_data/krb5.sh => 2
 not ok 1 - running krb5 tests
 #   Failed test 'running krb5 tests'
 #   at ../test/recipes/95-test_external_krb5.t line 25.
@@ -4008,7 +4008,7 @@ Test Summary Report
 95-test_external_krb5.t (Wstat: 256 Tests: 1 Failed: 1)
   Failed test:  1
   Non-zero exit status: 1
-Files=1, Tests=1, 94 wallclock secs ( 0.40 usr  0.09 sys + 68.23 cusr 26.05 csys = 94.77 CPU)
+Files=1, Tests=1, 108 wallclock secs ( 0.46 usr  0.03 sys + 77.36 cusr 30.01 csys = 107.86 CPU)
 Result: FAIL
 make[1]: *** [Makefile:2869: _tests] Error 1
 make[1]: Leaving directory '/home/msp/src/openssl'
msp@debian:~/src/openssl$ 
```

More interesting is the diff between test 1 (relative) and 2 (absolute):

https://github.com/mspncp/openssl/commit/a22b9085ee33ad918428f81a9fa244614cf1f07e?diff=split#diff-0dd1b26591e0accec843946712ebf1beL3989-L3990

You need to open the large diff manually and scroll down to line 3989, where you find the failure:

```
Aborted (core dumped)	Seeding random number generator: . . . OK
make[5]: *** [Makefile:685: check-unix] Error 134
```

Interestingly, the diff shows no significant differences in the build process of both tests before the crash occurs. So the only explanation seems to be: It's indeed a build failure inside krb5, caused by the abs2rel bug. And yes, krb5's build system. So it seems like @mpeylo's fix is finally justified.

mspncp:
So in view of my results you have my approval for reverting "OpenSSL::Test: bring back the relative paths", provided you amend the commit message with a short explanation for the reason.

levitte:
Hold on.
It shouldn't be too hard to simply ensure absolute paths are given to the krb5 build...

levitte:
BTW, I just tried a build of master (fresh pull) with `enable-external-tests`, and...

``` console
g++  -Itest/ossl_shim -Itest/ossl_shim/include -Iinclude -I../master/test/ossl_shim -I../master/test/ossl_shim/include -I../master/include  -std=c++11 -pthread -m64 -Wall -O0 -g   -MMD -MF test/ossl_shim/ossl_shim-bin-ossl_shim_cc.d.tmp -MT test/ossl_shim/ossl_shim-bin-ossl_shim_cc.o -c -o test/ossl_shim/ossl_shim-bin-ossl_shim_cc.o ../master/test/ossl_shim/ossl_shim.cc
../master/test/ossl_shim/ossl_shim.cc: In function 'int bssl::TicketKeyCallback(SSL*, uint8_t*, uint8_t*, EVP_CIPHER_CTX*, HMAC_CTX*, int)':
../master/test/ossl_shim/ossl_shim.cc:394:8: error: 'HMAC_Init_ex' was not declared in this scope
  394 |   if (!HMAC_Init_ex(hmac_ctx, kZeros, sizeof(kZeros), EVP_sha256(), NULL) ||
      |        ^~~~~~~~~~~~
../master/test/ossl_shim/ossl_shim.cc: In function 'bssl::UniquePtr<ssl_ctx_st> bssl::SetupCtx(const TestConfig*)':
../master/test/ossl_shim/ossl_shim.cc:628:5: error: 'SSL_CTX_set_tlsext_ticket_key_cb' was not declared in this scope; did you mean 'SSL_CTX_set_tlsext_ticket_keys'?
  628 |     SSL_CTX_set_tlsext_ticket_key_cb(ssl_ctx.get(), TicketKeyCallback);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |     SSL_CTX_set_tlsext_ticket_keys
make[1]: *** [Makefile:24396: test/ossl_shim/ossl_shim-bin-ossl_shim_cc.o] Error 1
make[1]: Leaving directory '/home/levitte/gitwrk/openssl.net/official/_build'
make: *** [Makefile:2825: build_sw] Fel 2
```

No one else?

levitte:
> More interesting is the diff between test 1 (relative) and 2 (absolute):

Which was this command:

```
LD_LIBRARY_PATH=`echo -L../../../lib | sed -e "s/-L//g" -e "s/ /:/g"` KRB5_CONFIG=../../../config-files/krb5.conf LC_ALL=C  ./t_encrypt
Aborted (core dumped)
make[5]: *** [Makefile:685: check-unix] Error 134
```

I did a little debugging of that program, and ended up [here](https://github.com/krb5/krb5/blob/a3a5c9b31cd9b421941eff9cd11dbc3876f4ff90/src/lib/crypto/openssl/hash_provider/hash_evp.c#L64-L68):

``` C
static krb5_error_code
hash_md4(const krb5_crypto_iov *data, size_t num_data, krb5_data *output)
{
    return hash_evp(EVP_md4(), data, num_data, output);
}
```

When seeing `EVP_md4()`, claxon went off in my head, because it's in the legacy provider, which hasn't been loaded here.

This has *nothing* to do with the paths, though, as `OPENSSL_MODULES` is always set to an absolute path in our `Makefile`...

Also, I did this mod, which should clear all doubts regarding the relativeness of our test paths:

``` diff
diff --git a/test/recipes/95-test_external_krb5_data/krb5.sh b/test/recipes/95-test_external_krb5_data/krb5.sh
index 6b267242f7..15d142ad0a 100755
--- a/test/recipes/95-test_external_krb5_data/krb5.sh
+++ b/test/recipes/95-test_external_krb5_data/krb5.sh
@@ -8,8 +8,8 @@
 
 # krb5's test suite clears LD_LIBRARY_PATH
-LDFLAGS="-L`pwd`/$BLDTOP -Wl,-rpath,`pwd`/$BLDTOP"
-CFLAGS="-I`pwd`/$BLDTOP/include -I`pwd`/$SRCTOP/include"
+LDFLAGS="-g -O0 -L`cd $BLDTOP; pwd` -Wl,-rpath,`cd $BLDTOP; pwd`"
+CFLAGS="-g -O0 -I`cd $BLDTOP/include; pwd` -I`cd $SRCTOP/include; pwd`"
 
 cd $SRCTOP/krb5/src
 autoreconf
```

mspncp:
> This has _nothing_ to do with the paths, though, as `OPENSSL_MODULES` is always set to an absolute path in our `Makefile`...

That's good news, Richard! I'd be more than happy if we wouldn't have to revert to absolute pathes.

> Also, I did this mod, which should clear all doubts regarding the relativeness of our test paths:

So is this already the fix, or are still figuring it out?

levitte:
Ah, found it.  OPENSSL_CONF is set with a relative path, and that's where we do get in trouble.

So, proposed fix:

``` diff
diff --git a/test/recipes/95-test_external_krb5.t b/test/recipes/95-test_external_krb5.t
index 0f2be87902..889626a1be 100644
--- a/test/recipes/95-test_external_krb5.t
+++ b/test/recipes/95-test_external_krb5.t
@@ -10,6 +10,7 @@
 use OpenSSL::Test;
 use OpenSSL::Test::Utils;
 use OpenSSL::Test qw/:DEFAULT data_file srctop_file bldtop_dir/;
+use Cwd qw(abs_path);
 
 setup("test_external_krb5");
 
@@ -20,6 +21,7 @@ plan skip_all => "krb5 not available"
 
 plan tests => 1;
 
-$ENV{OPENSSL_CONF} = srctop_file("test", "default-and-legacy.cnf");
+$ENV{OPENSSL_MODULES} = abs_path($ENV{OPENSSL_MODULES});
+$ENV{OPENSSL_CONF} = abs_path(srctop_file("test", "default-and-legacy.cnf"));
 
 ok(run(cmd([data_file("krb5.sh")])), "running krb5 tests");
```

levitte:
(I'm currently testing that fix, FYI)

levitte:
Yup, looks promising, my KRB5 test run it's getting *much* further now.

mattcaswell:
> Yup, looks promising, my KRB5 test run it's getting much further now.

@levitte - are you going to create a PR for this fix?

levitte:
I suggest you incorporate my fix in this PR

mattcaswell:
> I suggest you incorporate my fix in this PR

Ok - will do (although the first commit in this PR is already merged, and the second commit will be replaced by your fix - so this PR will only be your fix!)

mattcaswell:
I've pushed the suggested fix into this PR, and taken it out of WIP (author is set to @levitte since I had very little to do with it other than create the PR).

I also rebased to pick up my earlier travis fix that has already been merged, and set extended tests on. So lets see what travis makes of it.

mspncp:
> Ah, found it. OPENSSL_CONF is set with a relative path, and that's where we do get in trouble.

@levitte is my understanding correct that the crash occurs because `default-and-legacy.cnf` was not found and therefore the legacy algorithms (here: MD4) were not loaded?

If that's the case, there is still an NPD bug lurking in the krb5 test suite (t_encrypt.c), which could reoccur at any time, right? Shouldn't we raise an issue to inform the krb5 guys?

mattcaswell:
> If that's the case, there is still an NPD bug lurking in the krb5 test suite (t_encrypt.c), which could reoccur at any time, right? Shouldn't we raise an issue to inform the krb5 guys?

It's possible that its been fixed in more recent versions. We've not updated our krb5 submodule in quite a while.

mattcaswell:
> I think this deserves an [urgent] label, because the issue has been discussed long enough and Travis is waiting dearly for green light.

Agreed

mspncp:
Both CIs are green. Let's get it merged :-)

mattcaswell:
Already on it...

mattcaswell:
Pushed!

mspncp:
@mattcaswell one question in retrospect regarding https://github.com/openssl/openssl/pull/10992#issuecomment-581013007: 

> > Do you think there is something wrong with the relative pathes in the krb5 tests?
> 
> I really have no idea. My fix was based on the observation that the krb5 test started failing when that commit went in. My trial fix seems to confirm that theory - the build was green with the two reverted commits in this PR and extended tests.

I wonder how  you were able to quickly correlate the krb5 test failures with Richards relpath fix. Is there a more efficient way to do it than clicking and scrolling through painfully slowly loading travis logs? I suspect you utilized the [openssl-commits] mailing list, because it contains git push notifications together with ci reports?  (The reason why I suspect it is because you recently told me that you would rather not have the push notifications removed from [openssl-commits] and that you use them frequently.)

levitte:
> @levitte is my understanding correct that the crash occurred because default-and-legacy.cnf was not found and therefore the legacy algorithms (here: MD4) were not loaded?

Worse, we hadn't set `OPENSSL_CONF` at all, so the krb5 test only had the default provider to play with.

Now, the krb5 tests move around in its diverse directories, so `OPENSSL_CONF` *must* be an absolute path when running them, or we would get exactly the situation you mention.

mspncp:
> > If that's the case, there is still an NPD bug lurking in the krb5 test suite (t_encrypt.c), which could reoccur at any time, right? Shouldn't we raise an issue to inform the krb5 guys?
> 
> It's possible that its been fixed in more recent versions. We've not updated our krb5 submodule in quite a while.

I don't think so. krb5 was updated only recently to the latest official release by @kaduk #10690 during our last travis-fixing-session.

I would volunteer to further analyze the krb5 crash if @levitte could provide me with a hint (i.e., the commands) how to correctly configure  krb5 for debugging (ideally with CFLAGS='-g -O0'). I would have [debugged t_encrypt](https://github.com/openssl/openssl/pull/10992#issuecomment-581112523) myself before Richard did it, if I had succeded to get the debug infos working.

kaduk:
> > > If that's the case, there is still an NPD bug lurking in the krb5 test suite (t_encrypt.c), which could reoccur at any time, right? Shouldn't we raise an issue to inform the krb5 guys?
> > 
> > 
> > It's possible that its been fixed in more recent versions. We've not updated our krb5 submodule in quite a while.
> 
> I don't think so. krb5 was updated only recently to the latest official release by @kaduk in #10690 during our last travis-fixing-session.
> 
> I would volunteer to further analyze the krb5 crash if @levitte could provide me with a hint (i.e., the commands) how to correctly configure krb5 for debugging (ideally with CFLAGS='-g -O0'). I would have [debugged t_encrypt](https://github.com/openssl/openssl/pull/10992#issuecomment-581112523) myself before Richard did it, if I had succeded to get the debug infos working.

The(?) bug is not too hard to find going from `t_encrypt.c` and `EVP_md4()`:

https://github.com/krb5/krb5/blob/master/src/lib/crypto/openssl/hash_provider/hash_evp.c#L36-L45 receives a NULL `EVP_MD *` and then tries to use it without checks; it gets that NULL from https://github.com/krb5/krb5/blob/master/src/lib/crypto/openssl/hash_provider/hash_evp.c#L67

levitte:
@mspncp, the krb5 test recipe has a companion script. Have a look, I think it will be obvious where you add the debugging flags... 

https://github.com/openssl/openssl/blob/master/test/recipes/95-test_external_krb5_data/krb5.sh

levitte:
Actually, hash_evp *does* check. The "use" is codified like this:

``` C
    ok = ok && EVP_DigestUpdate(...);
```

If `ok` is already false, how many times do you think `EVP_DigestUpdate` is executed?
(this is a very smart construct, BTW) 

The issue there is that it ends up returning false and set ENOMEM. The false return is caught higher up in the call chain and generates a printout and a call to `abort()` (which dumps core, which enables you to do some forensic analysis)

mattcaswell:
> I wonder how you were able to quickly correlate the krb5 test failures with Richards relpath fix. Is there a more efficient way to do it than clicking and scrolling through painfully slowly loading travis logs?

No...that's exactly what I did. I knew that where the builds were green krb5 was obviously passing...and somewhere between the first red build and the HEAD the krb5 test had started failing. I then did a "manual" binary search to locate the failing build. A little time consuming but it didn't take too long.

kaduk:
> Actually, hash_evp _does_ check. The "use" is codified like this:

Yes, but the code is in general written on an assumption that `EVP_md4()` cannot fail, something that we even document: https://www.openssl.org/docs/man1.1.1/man3/EVP_md4.html 
Perhaps we should update documentation for 3.0 to note that things can fail in the absence of the proper provider

mattcaswell:
> Yes, but the code is in general written on an assumption that EVP_md4() cannot fail, 

Well, strictly speaking `EVP_md4()` still cannot fail. But if you attempt to *use* it to initialise an encrypt/decrypt operation then the initialisation will fail.

mspncp:
No need to open a krb5 issue anymore. We were already heared, see https://github.com/krb5/krb5/pull/1034.

mspncp:
> The(?) bug is not too hard to find going from `t_encrypt.c` and `EVP_md4()`:
> 
> https://github.com/krb5/krb5/blob/master/src/lib/crypto/openssl/hash_provider/hash_evp.c#L36-L45 receives a NULL `EVP_MD *` and then tries to use it without checks; it gets that NULL from https://github.com/krb5/krb5/blob/master/src/lib/crypto/openssl/hash_provider/hash_evp.c#L67

IIRC we plan to disable the legacy provider by default. I think we need to do document this problem clearly, otherwise we might break the promise that people just need to recompile their applications for OpenSSL 3.0 and everything will work.

mattcaswell:
> https://github.com/krb5/krb5/blob/master/src/lib/crypto/openssl/hash_provider/hash_evp.c#L36-L45 receives a NULL EVP_MD * 

This is surely not correct. Even with the legacy provider not loaded you will still not get a NULL EVP_MD object:

https://github.com/openssl/openssl/blob/450d12c825cc9016e5e8990423fa7ffdb843a1f0/crypto/evp/legacy_md4.c#L22-L33

PR #10996, opened 2020-02-02
Missing "obj_mac" header file in "dh_lib"

davxy:
Usage of `NID_undef` symbol without including its definition was causing
a build fail

### DIFF
--- a/crypto/dh/dh_lib.c
+++ b/crypto/dh/dh_lib.c
@@ -7,14 +7,15 @@
  * https://www.openssl.org/source/license.html
  */
 
-#include <stdio.h>
+#include "dh_local.h"
+#include "crypto/dh.h"
 #include "internal/cryptlib.h"
 #include "internal/refcount.h"
 #include <openssl/bn.h>
-#include "dh_local.h"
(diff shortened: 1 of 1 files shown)


levitte:
Good that you found it.  However, your change reorders the inclusion in a way that we don't usually do these days, which is exactly the inverse of the reordering you did.

davxy:
Ok, I reorder the includes.

In general, what is the include order strategy adopted by the project?

Is this correct?
1. C system files (e.g. <stdio.h>).
2. project's public header files (e.g. <openssl/bn.h>). 
3. project's private header files (e.g. "internal/refcount.h")
4. local private header files, such as headers in the same folder of sources (e.g. "dh_local.h")

One other question.

In these cases is preferred another independent commit or do you prefer an "amend commit" followed by a "force push"?

levitte:
That is the correct order.

As far as commits go, we're not very strict. Personally, I often make fixup commits (`git commit --fixup=xxxxx`)

levitte:
Done

mattcaswell:
Pushed.

PR #10997, opened 2020-02-02
WIP: Add EXDATA to X509_LOOKUP

richsalz:
This adds EXDATA to X509_LOOKUP.  It's a WIP because I'm looking for comments about doing this.

OpenSSL should be consistent, and that therefore there are other places that need to be converted as well.  If the team likes this PR then I will also address the following:
```
; g grep et_app_data | grep -v define  
asn1.h:void ASN1_SCTX_set_app_data(ASN1_SCTX *p, void *data);
asn1.h:void *ASN1_SCTX_get_app_data(ASN1_SCTX *p);
evp.h:*int EVP_MD_meth_set_app_datasize(EVP_MD *md, int datasize);
evp.h:*int EVP_MD_meth_get_app_datasize(const EVP_MD *md);
evp.h:void *EVP_CIPHER_CTX_get_app_data(const EVP_CIPHER_CTX *ctx);
evp.h:void EVP_CIPHER_CTX_set_app_data(EVP_CIPHER_CTX *ctx, void *data);
evp.h:void EVP_PKEY_CTX_set_app_data(EVP_PKEY_CTX *ctx, void *data);
evp.h:void *EVP_PKEY_CTX_get_app_data(EVP_PKEY_CTX *ctx);
```
The two marked with an asterisk are kinda problematic because they're not "appdata" like all the others, but rather context size.


levitte:
For context structures, I see zero need for EX_DATA, as they are typically are short term, fairly local instances.  Furthermore, I would say that `EVP_PKEY_CTX_set_app_data()` and `EVP_PKEY_CTX_get_app_data()` are misnamed, as their only purpose is to pass application data to a param/key generation callback; they should have been called `EVP_PKEY_CTX_set_cb_data()` and `EVP_PKEY_CTX_get_cb_data()`...  it's possible that at the time, the intended purpose was a bit more, how shall I say...  grandiose?

levitte:
As for X509_LOOKUP...  well, maybe, but I'm frankly a bit sceptical.  However, I'm not the best source of knowledge around those bits, so.  @petrovr tends to have something to say around these things.

richsalz:
does anyone else have a view on making EXDATA universally available, more or less?

kaduk:
There's a consistency argument for having `EX_DATA` universally available for "long-lived" objects.  I always confuse myself on the `X509_LOOKUP` lifecycle and have to re-read the code, though, so I won't comment on that specifically.

richsalz:
While waiting for @openssl/otc to decide, I rebased. :(

richsalz:
Does @openssl/omc or @openssl/otc care about this?

richsalz:
How long is reasonable to wait for feedback?  I'm closing this and will turn it into an issue.

Issue #10995, opened 2020-02-02
Why are int and size_t are used so inconsistently


davxy:
For example, the function:

` int CRYPTO_secure_malloc_init(size_t size, int minsize);`

What is the rational behind the choice of `size_t` for `size` and `int` for `minsize`?
I mean, why not declare both as `size_t`?

Studying the codebase I found other cases here and there like this one.

Is there any particular reason for this type of choice? 

The only rational explanation I can think of is that in the past the `int` parameter was used with a negative value to trigger some particular behavior... but looks strange anyway.

Can we change the interface types where appropriate and not dangerous? Or are we  condemned to this badly designed interfaces forever because is too risky at this point?

Thanks

levitte:
In this case, we can't even blame history, as this function appeared in 1.1.0.  Otherwise, it does turn out that history is to blame, and we're quite conservative in changing the argument types when we fear that applications rely on old signatures.

Our current main focus is on the internals for the moment, so don't expect much prompt change in this regard.  It is in the back of our minds, though, that we do need to clean up our history at some point.

mattcaswell:
This has been a problem for a long time. At one point I went through and converted lots of libssl to use size_t internally as much as possible - but I was unable to change usage in the actual API because of backwards compat rules. We should be using size_t for all new APIs where we're talking about the size of something. I don't know how this particular function slipped through. That just looks like a mistake to me.

richsalz:
Notwithstanding the general issue, which Matt and Richard explained, in this particular case there is no need for minsize to be size_t as it should generally be a small number (32 or 16 say), and having int provides a tiny sliver of safety.

levitte:
Er, I fail to see the safety with `int` over `size_t`. The latter is a well known type in C90 and up.

richsalz:
negative numbers.

mattcaswell:
> negative numbers.

That makes no sense to me. How does that provide you with any safety? If anything its the other way around. `minsize` is a size and should not be negative. The fact that it can be due to its type means there is less safety that had it been a size_t.

richsalz:
If a negative number is passed in, the implementation can check.  If a someone does ```(size_t)val``` where they get the val wrong, it will be a huge number.  Like I said, a tiny sliver of safety.

mattcaswell:
[...]

PR #11003, opened 2020-02-03
Make minimum size for secure memory a size_t.

paulidale:
The minimum size argument to CRYPTO_secure_malloc_init() was an int but ought
to be a size_t since it is a size.

From an API perspective, this is a change.  However, the minimum size is
verified as being a positive power of two and it will typically be a small
constant, so impact should be low to none.

### DIFF
--- a/crypto/mem_sec.c
+++ b/crypto/mem_sec.c
@@ -57,7 +57,7 @@ static CRYPTO_RWLOCK *sec_malloc_lock = NULL;
 /*
  * These are the functions that must be implemented by a secure heap (sh).
  */
-static int sh_init(size_t size, int minsize);
+static int sh_init(size_t size, size_t minsize);
 static void *sh_malloc(size_t size);
 static void sh_free(void *ptr);
 static void sh_done(void);
@@ -65,7 +65,7 @@ static size_t sh_actual_size(char *ptr);
 static int sh_allocated(const char *ptr);
 #endif
 
-int CRYPTO_secure_malloc_init(size_t size, int minsize)
+int CRYPTO_secure_malloc_init(size_t size, size_t minsize)
 {
 #ifdef OPENSSL_SECURE_MEMORY
     int ret = 0;
@@ -373,7 +373,7 @@ static void sh_remove_from_list(char *ptr)
 }
 
 
-static int sh_init(size_t size, int minsize)
+static int sh_init(size_t size, size_t minsize)
 {
     int ret;
     size_t i;
@@ -385,11 +385,10 @@ static int sh_init(size_t size, int minsize)
     /* make sure size and minsize are powers of 2 */
     OPENSSL_assert(size > 0);
     OPENSSL_assert((size & (size - 1)) == 0);
-    OPENSSL_assert(minsize > 0);
     OPENSSL_assert((minsize & (minsize - 1)) == 0);
     if (size <= 0 || (size & (size - 1)) != 0)
         goto err;
-    if (minsize <= 0 || (minsize & (minsize - 1)) != 0)
+    if (minsize == 0 || (minsize & (minsize - 1)) != 0)
         goto err;
 
     while (minsize < (int)sizeof(SH_LIST))
--- a/doc/man3/OPENSSL_secure_malloc.pod
+++ b/doc/man3/OPENSSL_secure_malloc.pod
@@ -14,7 +14,7 @@ CRYPTO_secure_used - secure heap storage
 
  #include <openssl/crypto.h>
 
- int CRYPTO_secure_malloc_init(size_t size, int minsize);
+ int CRYPTO_secure_malloc_init(size_t size, size_t minsize);
 
  int CRYPTO_secure_malloc_initialized();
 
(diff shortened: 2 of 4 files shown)


mattcaswell:
I wonder if there are any compilers which might complain about this if applications make this call? Does this deserve a CHANGE entry...or is it more like the "const" changes we make which don't usually have one.

paulidale:
`test/secmem.c` didn't actually have a **size_t** cast -- one in a comment, one inside `#if 0` because it tests an infinite loop condition.  Nonetheless, I've removed the cast from the second argument of both but not from the first where a wider than **int** value is required.

I've updated the documentation.

@mattcaswell, I've added a note in the _history_ section of the documentation.  CHANGES seems excessive but I'm happy to add an entry.

paulidale:
Is a vote required for this?

Technically, it is a breaking API change.  Realistically, it's very safe.  Unless, someone is taking the address of the function and assigning it to a variable.

DDvO:
> I wonder if there are any compilers which might complain about this if applications make this call? Does this deserve a CHANGE entry...or is it more like the "const" changes we make which don't usually have one.

One potential compile-time problem could be if a caller passes in a clearly negative value, thought is is very unlikely since there was already the dynamic check for negative input, which should have lead to a run-time error.
I've just tried out with `gcc` on code configured with `--strict-warnings`. Interestingly, it does not complain if a `-1` is passed as actual argument and one changes the parameter type from  `int`  to  `size_t`.

richsalz:
I think this is a bad thing to do, for a couple of reasons
- It could break something
- It works fine now
- It's a lot of changes (doc, code, casts) for no real gain other than nerd esthetics.
- It doesn't fully address the issue #10995 where this was used as an example.
- It establishes a bad precedent of "we think it safe let's fix it" which is counter to one of 3.0's stated goals: "most applications should work just be recompiling."  The slippery slope is very very slippery. :)

mattcaswell:
> a couple of reasons

It amuses me that you list 5 reasons:
https://xkcd.com/1070/

levitte:
> It's a lot of changes (doc, code, casts) for no real gain other than nerd esthetics.

I've heard that consistency is a good thing :wink:

richsalz:
>I've heard that consistency is a good thing

Okay, I'll take the bait since you're trying to use my words against me:  how does this PR make things more consistent?

richsalz:
> It amuses me that you list 5 reasons:

I was sloppy and didn't go back and fix the first line. :)  The bullets just kept a-comin'.

I had an English class once where we had to sort sentence in order:
```
A couple of the clarinetists...
Most of the clarinetists...
Many of the clarinetists...
A few of the clarinetists...
```
Wish I could find that page; it went on for 50 or so lines!

DDvO:
The event that a caller provides a negative value for the `minsize` parameter is very unlikely because it make no sense and typically the value will be a simple constant, so would be obvious.
And even if so, what could really go wrong here if that is subsequently taken as a huge unsigned value?
AFAICS:
* a huge about of heap memory is reserved successfully, which is again very unlikely in itself
* the caller obtains an error result, as before with the hitherto version of the function

levitte:
> how does this PR make things more consistent?

Easy.  Consistent use of `size_t` for passing sizes.

richsalz:
So "fix" SSL_read and SSL_write?  Those are far more glaring inconsistencies.  And fixing this one, which was done deliberately and correctly... shrug.  I've said why I think this is a bad idea and given explanations.  Oh well.

levitte:
Actually, I see no reason why we shouldn't fix `SSL_read` and `SSL_write` in a similar way...
However, that shouldn't be a factor in approving or not approving this PR.

DDvO:
As there are concerns on doing this and related changes right now, how about simply deferring them to after 3.0? 
There is nothing urgent in here.

paulidale:
Major releases are the time for changes like this -- API compatibility isn't guaranteed at these points.  A deferral until after 3.0 will be a long time after...

Changing the _int_ to a _size_t_ is definitely break ABI compatibility.

davxy:
Well... I'm not a committer so my opinion is not so influent. But I think that if you wait after 3.0 for stuff like this, the change will never happen.

A major release like 3.0 is not only an opportunity to massively change the internals, but is also the best opportunity to improve the API wherever is required.

Is not only a matter of nerd esthetics but is a matter of type safety and coerence.

As I understood, for someone (maybe the original author) these are stupid changes, but for the community means a lot to have a consistent api.

DDvO:
> Major releases are the time for changes like this -- API compatibility isn't guaranteed at these points. A deferral until after 3.0 will be a long time after...

Now I'm confused. @kroeckx explained to me a couple of months ago that API changes for 3.0 should be kept to the minimum: https://github.com/openssl/openssl/pull/10504#issuecomment-558249028
> OpenSSL 3.0 is supposed to remain as compatible as possible, we expect most applications to just need to rebuild and not make any changes. Anything that requires them to make changes needs to wait until at least 4.0.

paulidale:
Major releases are when we _can_ make changes.  For 3.0, the decision was made that we are aiming for a high level of source compatibility and this _will_ limit changes.

I can't see this particular change breaking anyone.

paulidale:
Merged to master.  Thanks for the discussion.

PR #9389, opened 2019-07-15
WIP: OSSL_STORE for providers

levitte:
This is a *very* early draft, just to show a line of thought to implement this.

PR #11006, opened 2020-02-03
Redesign the KEYMGMT libcrypto <-> provider interface

### LINKED ISSUE #10979: EVP_KEYMGMT needs a redesign
The EVP_KEYMGMT libcrypto \<-\> provider interface currently makes a few assumptions:

1.  provider side domain parameters and key data isn't mutable.  In other words, as soon as a key has been created in any (loaded, imported data, ...), it's set in stone.
2.  provider side domain parameters can be strictly separated from the key data.

This does work for the most part, but there are places where that's a bit too rigid for the functionality that the `EVP_PKEY` API delivers.  In this API, the `EVP_PKEY` is commonly assumed to hold any of these combinations (other combinations are probably possible as well, just not very useful with the current functions):

-   domain parameters only
-   public key only
-   public key + private key
-   domain parameters + public key
-   domain parameters + public key + private key

Most of the code using `EVP_PKEY` objects assumes that it holds either just domain parameters or domain parameters + keypair.  Our provided implementations reflect that assumptions, so the strict separation mentioned higher has already slipped.

Then, you stumble on this:

https://github.com/openssl/openssl/blob/c98eab85b8a6c5d49353a229ac1d2f4cd0b84682/apps/ca.c#L1922-L1925

For all intents and purposes, this takes the domain parameters from one key and *alters* another key with them.  This is not possible to support in the current KEYMGMT design.

### Enabling mutability

The simple way to do this is to introduce two new functions, a `new` and a `free` that simply constructs and destructs the provider side key data structure.  Other constructors may be added with time (the `load` function comes to mind).

By consequence, the `import` functions will no longer be constructed, but rather act on existing key data that gets passed in as an argument.

*[TOPIC: mutability]*

### Merging domain parameters and key data

This is the simplest part on the provider side, we can simply stop pretending that they are separate, and rather see them as attributes of the same object, which reflects the way `EVP_PKEY` works.

However, there is still code that passes domain parameters specifically, and code that wants just the key data specifically, so we must still be able to support that, even though the provider side object combines all elements.  This could be done in two ways:

1.  Have `import` and `export` functions for each set of data, i.e. `importdomparams`, `exportdomparams`, `importkey`, `exportkey`
2.  Have one `import` and one `export` that takes an argument indicating what the caller wants, which could be domain parameters, the key itself, or both.

*[TOPIC: importing and exporting]*

### Library code, while I'm at it

libcrypto has a number of small functions that are a direct mapping of the libcrypto \<-\> provider interface.  They all have names starting with `evp_keymgmt_`.

libcrypto also has a number of utility functions that take care of the more complex interactions between `EVP_PKEY` and the `EVP_KEYMGMT` libcrypto \<-\> provider interface.  They all have names starting with `evp_keymgmt_`.

Quite obviously, there is going to be a mixup between those two sets of functions, so a rename is prudent, such as prefixing the latter set with `evp_keymgmt_util_`.

*[TOPIC: library utility functions]*

levitte:
The KEYMGMT libcrypto <-> provider interface currently makes a few
assumptions:

1.  provider side domain parameters and key data isn't mutable. In
    other words, as soon as a key has been created in any (loaded,
    imported data, ...), it's set in stone.
2.  provider side domain parameters can be strictly separated from the
    key data.

This does work for the most part, but there are places where that's a
bit too rigid for the functionality that the EVP_PKEY API delivers.
Key data needs to be mutable to allow the flexibility that functions
like EVP_PKEY_copy_parameters promise, as well as to provide the
combinations of data that an EVP_PKEY is generally assumed to be able
to hold:

- domain parameters only
- public key only
- public key + private key
- domain parameters + public key
- domain parameters + public key + private key

To remedy all this, we:

1.  let go of the distinction between domain parameters and key
    material proper in the libcrypto <-> provider interface (but only
    to a fashion, we still leave the possibility to choose a selection
    when importing and exporting).
2.  Rework the libcrypto <-> provider interface so provider side key
    objects are created and destructed with a separate function, and
    get their data filled and extracted in through import and export.

(future work will see other key object constructors and other
functions to fill them with data)

Fixes #10979


levitte:
To be noted is that I removed the `OP_keymgmt_load` and `OP_keymgmt_gen` functions.  They will re-appear in the PRs that implement them, such as #10289 and #9389.

crypto/evp/keymgmt_meth.c
```diff
@@ -47,154 +47,99 @@ static void *keymgmt_from_dispatch(int name_id,
...
+            if (keymgmt->validatepairwise == NULL)
+                keymgmt->validatepairwise =
+                    OSSL_get_OP_keymgmt_validate_pairwise(fns);
             break;
         }
     }
     /*
      * Try to check that the method is sensible.
+     * At least one constructor and the destructor are MANDATORY
+     * The functions has_public_key an has_private_key are MANDATORY
```
slontis: and

doc/man7/provider-keymgmt.pod
```diff
@@ -85,131 +71,109 @@ function pointer from a B<OSSL_DISPATCH> element named
...
-OP_keymgmt_get_key_params() should extract information data associated
-with the given I<key>, see L</Information Parameters>.
+OP_keymgmt_validate_domparams() should check if I<keydata> contains
+valid domain parameters.
 
-OP_keymgmt_gettable_key_params() should return a constant array of
-descriptor B<OSSL_PARAM>, for parameters that
-OP_keymgmt_get_key_params() can handle.
+OP_keymgmt_validate_public() should check if I<keydata> contains a
+valud public key.
```
slontis: valid

doc/man7/provider-keymgmt.pod
```diff
@@ -85,131 +71,109 @@ function pointer from a B<OSSL_DISPATCH> element named
...
 
-OP_keymgmt_validate_public() should return 1 if the public component of the
-key is valid, or 0 if invalid.
-OP_keymgmt_validate_private() should return 1 if the private component of the
-key is valid, or 0 if invalid.
-OP_keymgmt_validate_pairwise() should return 1 if the the pairwise consistency
-of the key is valid, or 0 if invalid.
+OP_keymgmt_validate_public() should check if I<keydata> contains a
+valud private key.
```
slontis: valid

doc/man7/provider-keymgmt.pod
```diff
@@ -219,6 +183,30 @@ returns NULL, the caller is free to assume that there's an algorithm
 from the same provider, of the same name as the one used to fetch the
 keymgmt and try to use that.
 
+=head2 Import and Export Selections
+
+The I<selection> given to OP_keymgmt_import() and OP_keymgmt_export()
+indicates what values of the I<keydata> is of interest.  It can be
```
slontis: are of

doc/man7/provider-keymgmt.pod
```diff
@@ -219,6 +183,30 @@ returns NULL, the caller is free to assume that there's an algorithm
...
+=item B<OSSL_KEYMGMT_WANT_DOMPARAMS>
+
+to indicate that only the domain parameters are interesting.
+
+=item B<OSSL_KEYMGMT_WANT_KEY>
+
+to indicate that only the key data (the data that constitutes the
+private and the public keys) are interesting.
+
+=item B<OSSL_KEYMGMT_WANT_DOMPARAMS>
```
slontis: ??

include/openssl/core_numbers.h
```diff
@@ -333,109 +333,87 @@ OSSL_CORE_MAKE_FUNC(int, OP_kdf_set_ctx_params,
 /*-
  * Key management
  *
- * Key domain parameter references can be created in several manners:
- * - by importing the domain parameter material via an OSSL_PARAM array.
- * - by generating key domain parameters, given input via an OSSL_PARAM
- *   array.
+ * The Key Management takes care of provider side key objects, and includes
+ * all current functionality to create them, destroy them, setting parameters
```
slontis: set

providers/implementations/keymgmt/dh_kmgmt.c
```diff
@@ -128,70 +131,83 @@ static int key_to_params(DH *dh, OSSL_PARAM_BLD *tmpl)
...
 
+    if (dh == NULL)
+        return 0;
+
     ossl_param_bld_init(&tmpl);
-    if (dh == NULL
-        || !key_to_params(dh, &tmpl)
-        || (params = ossl_param_bld_to_param(&tmpl)) == NULL)
+
+    if (selection != OSSL_KEYMGMT_WANT_KEY
```
slontis: OSSL_KEYMGMT_WANT_BOTH ??

providers/implementations/keymgmt/rsa_kmgmt.c
```diff
@@ -162,30 +168,85 @@ static int key_to_params(RSA *rsa, OSSL_PARAM_BLD *tmpl)
...
+static int rsa_import(void *keydata, int selection, const OSSL_PARAM params[])
+{
+    RSA *rsa = keydata;
+
+    if (rsa == NULL)
+        return 0;
+
+    /* TODO(3.0) PSS and OAEP should bring on parameters */
+    /* If the caller wants domain parameters, we do nothing for the moment */
+    if (selection == OSSL_KEYMGMT_WANT_DOMPARAMS)
```
slontis: Just a note that : PSS and OAEP are NOT and SHOULD NOT EVER be treated as domain parameters. It would be very confusing.

providers/implementations/keymgmt/rsa_kmgmt.c
```diff
@@ -162,30 +168,85 @@ static int key_to_params(RSA *rsa, OSSL_PARAM_BLD *tmpl)
...
+        return 1;
+
+    /*
+     * TODO(3.0) The |selection| check is *currently* not necessary, but when
```
slontis: See above. There are no domain parameters. These are key properties.

providers/implementations/keymgmt/rsa_kmgmt.c
```diff
@@ -162,30 +168,85 @@ static int key_to_params(RSA *rsa, OSSL_PARAM_BLD *tmpl)
...
+{
+    RSA *rsa = keydata;
     OSSL_PARAM_BLD tmpl;
     OSSL_PARAM *params = NULL;
     int ret;
 
+    if (rsa == NULL)
+        return 0;
+
+    /* TODO(3.0) PSS and OAEP should bring on parameters */
```
slontis: Just to repeat... NOOOOOOOOOOOO!!!

doc/man7/provider-keymgmt.pod
levitte: !!

levitte: (I can do this all day, so maybe you should tell me what you wonder about...)

slontis: Block copy bug - should be OSSL_KEYMGMT_WANT_BOTH

providers/implementations/keymgmt/rsa_kmgmt.c
levitte: What should they be treated as, then?  Maybe it's just about wordings, should I call them "attributes" (mimicking the PKCS#8 terminology)?  I would have no remorse whatsoever making that kind of name change if that makes more sense.

Do note that `crypto/rsa/rsa_ameth.c` talks about them in terms of parameters...  is it "domain" specifically that's disturbing?

levitte: Or, seeing your other comment, "properties" work too.  Can domain parameters be called properties too, or is that like swearing in church?

slontis: Yep. RSA has no domain parameters.
Maybe making it bit flags and treating them differently would be better.
i.e have domain params, and some other params..

doc/man7/provider-keymgmt.pod
levitte: Ah!  Ok, will fix

providers/implementations/keymgmt/rsa_kmgmt.c
slontis: Arent those things you are talking about just padding related?

levitte:
I'm a bit amazed that "domainparams" didn't catch fire earlier, but hey, it's never too late...

So, I kinda like "properties" or "attributes"...  PKCS#8 uses the latter, so that could be quite fitting.

Anyone else with better ideas, or just helping me choose?

crypto/evp/keymgmt_meth.c
levitte: Fixed

doc/man7/provider-keymgmt.pod
levitte: Off-by-one keyboard error...

Fixed.

slontis:
> Anyone else with better ideas, or just helping me choose?

Dont care  too much , we can change that name if needed. I still want domainparams to exist. And flags would also be better just in case this expands to do other things in the future.

doc/man7/provider-keymgmt.pod
levitte: Yup

doc/man7/provider-keymgmt.pod
levitte: Fixed

doc/man7/provider-keymgmt.pod
levitte: Fixed

include/openssl/core_numbers.h
levitte: Fixed

providers/implementations/keymgmt/dh_kmgmt.c
levitte: So you'd prefer the more explicit `if (selection == OSSL_KEYMGMT_WANT_ATTRIBUTES || selection == OSSL_KEYMGMT_WANT_BOTH)`?

slontis: I would prefer flags so you dont need to do that.

levitte:
> I still want domainparams to exist

If we end up having esssentially the same thing, but call it "domainparams" in some cases and "attributes" in another, we haven't done things right.  I want *one* common name for these things, anddd then leave it to each implementation to treat that as they wish.  It's still perfectly acceptable to document that in the case of DH and DSA, "whatever-name-we-choose" holds the domain parameters.

levitte:
I think I'll let PKCS#8 lead the way, and rename DOMPARAMS to ATTRIBUTES

providers/implementations/keymgmt/rsa_kmgmt.c
levitte: > Arent those things you are talking about just padding related?

Not in the case of PSS

slontis:
From a conceptual point of view.. There are
domain params
pub key
private key.

These 'other' parameters sound like a hack.
They also sound more like they are associated with a key?

So I guess you are trying to lump domain parameters into the key_properties?
Should they be lumped together - I dont know - but it sounds wrong.

slontis:
If RSA returns true to hasdomparameters then I will not be pleased.

slontis:
We could also end up with a key that 'can' have domain parameters, that returns true because it has some other key properties but no domain parameters.

levitte:
Let me backtrack a little.  My talk of PKCS#8 was la-la-land-talk.

What I was actually leaning on was this:

```
   AlgorithmIdentifier  ::=  SEQUENCE  {
        algorithm               OBJECT IDENTIFIER,
        parameters              ANY DEFINED BY algorithm OPTIONAL  }
```

Now, looking up the latest [PKCS#1 RFC](https://tools.ietf.org/html/rfc8017):

> **[A.1.](https://tools.ietf.org/html/rfc8017#appendix-A.1) RSA Key Representation**
> ...
>    The object identifier rsaEncryption identifies RSA public and private
>    keys as defined in Appendices A.1.1 and A.1.2.  *The parameters field
>    has associated with this OID in a value of type AlgorithmIdentifier
>    SHALL have a value of type NULL.*

> **[A.2.1.](https://tools.ietf.org/html/rfc8017#appendix-A.2.1)  RSAES-OAEP**
> ...
>    *The parameters field associated with this OID in a value of type
>    AlgorithmIdentifier SHALL have a value of type RSAES-OAEP-params*:
> 
>    ```
>    RSAES-OAEP-params ::= SEQUENCE {
>        hashAlgorithm      [0] HashAlgorithm     DEFAULT sha1,
>        maskGenAlgorithm   [1] MaskGenAlgorithm  DEFAULT mgf1SHA1,
>        pSourceAlgorithm   [2] PSourceAlgorithm  DEFAULT pSpecifiedEmpty
>    }
>    ```

> [A.2.3.](https://tools.ietf.org/html/rfc8017#appendix-A.2.3)  RSASSA-PSS
> ...
>    *The parameters field associated with this OID in a value of type
>    AlgorithmIdentifier SHALL have a value of type RSASSA-PSS-params:*
> 
>    ```
>    RSASSA-PSS-params ::= SEQUENCE {
>        hashAlgorithm      [0] HashAlgorithm      DEFAULT sha1,
>        maskGenAlgorithm   [1] MaskGenAlgorithm   DEFAULT mgf1SHA1,
>        saltLength         [2] INTEGER            DEFAULT 20,
>        trailerField       [3] TrailerField       DEFAULT trailerFieldBC
>    }
>    ```

Looking at [RFC 3279 - Algorithms and Identifiers for the Internet X.509 Public Key Infrastructure Certificate and Certificate Revocation List (CRL) Profile](https://tools.ietf.org/html/rfc3279), I find:

> **[2.3.2](https://tools.ietf.org/html/rfc3279#section-2.3.2)  DSA Signature Keys**
> ...
>    If the DSA domain parameters are present in the subjectPublicKeyInfo
>    AlgorithmIdentifier, the parameters are included using the following
>    ASN.1 structure:
> 
>    ```
>       Dss-Parms  ::=  SEQUENCE  {
>           p             INTEGER,
>           q             INTEGER,
>           g             INTEGER  }
>    ```

> **[2.3.3](https://tools.ietf.org/html/rfc3279#section-2.3.3)  Diffie-Hellman Key Exchange Keys**
> ...
>    The dhpublicnumber OID is intended to be used in the algorithm field
>    of a value of type AlgorithmIdentifier.  The parameters field of that
>    type, which has the algorithm-specific syntax ANY DEFINED BY
>    algorithm, have the ASN.1 type DomainParameters for this algorithm.
> 
>    ```
>       DomainParameters ::= SEQUENCE {
>             p       INTEGER, -- odd prime, p=jq +1
>             g       INTEGER, -- generator, g
>             q       INTEGER, -- factor of p-1
>             j       INTEGER OPTIONAL, -- subgroup factor
>             validationParms  ValidationParms OPTIONAL }
> 
>       ValidationParms ::= SEQUENCE {
>             seed             BIT STRING,
>             pgenCounter      INTEGER }
>    ```


See what I'm getting at?  I *believe* the "parameters" that are mentioned all over the place comes directly from the `AlgorithmIdentifier` definition, and can hold whatever parameters that are associated with a key, be it the domain parameters for DSA and DH, OAEP or PSS parameters for RSA, etc etc etc.

So I guess this ends up being about removing "domain" everywhere I've mistakenly written "domain parameters".  Fine, no problem at all, and that approaches already existing code anyway, so...

t-j-h:
Those two groups of things are completely different concepts. 
They aren't something to mix together conceptually IMHO.

slontis:
fair enough - but you cant get rid of it 'everywhere'..
We still want to know if something has domain params.. (as opposed to these 'other' parameters).
And key validation is on the domain params..

slontis:
> I think I'll let PKCS#8 lead the way, and rename DOMPARAMS to ATTRIBUTES

If you had them as flags then it shouldnt be a big issue to do both at once without having to collapse them into one... i.e you could request  _ATTRIBUTES | _DOMPARAMS

levitte:
Having spoken in a vid call and thought about it a little bit more and sought some last minute feedback from @mattcaswell, I figured that the *selector* bit approach does make sense after all.  It would allow us to codify the needs of the EVP layer in an extensible manner, and the broader view is simply a combination of bits.

This could also allow us to reduce the set of `OP_keymgmt_has_something()` functions to just `OP_keymgmt_has()` that takes a *selector* alongside the *keydata*, and probably also reduce the diverse `OP_keymgmt_validate_something()` into just one `OP_keymgmt_validate()`.

A hard decision has to be made whether a combination of selector bits should mean that *all* of those subsets of *keydata* is considered, or just *any* (i.e. do we look on them with AND in mind or with OR in mind).  Intuitively, it seems that *all* inclusive approach makes better sense.  If everyone agrees with that approach (I already know that @mattcaswell does), we're good.

paulidale:
_all_.

levitte:
Actually, *all* or *any* could be left to the implementation to decide.  For example, the legacy implementations often only compare the public key when asked to compare a key, and the EVP layer doesn't really care, it only cares about the key pair (oh, and parameters if there are any, but that's less relevant here) for EVP_PKEY comparisons.  This does make sense, 'cause for valid keys, a matching public key implies a matching private key.

paulidale:
I think it is more subtle.  In each of the following cases the example assumes a public key is present but not the private.  The selection passed asks for both public and private

1. For _has_ requiring that all be present makes perfect sense.  In the example, _has_ would return false.
1. For _validate_ it could insist that everything selected is valid or it could validate only those that are selected and are present.  The example is broken here -- I don't know which way is better.
1. A _get_ would seem to do best to return everything that matches any selector bit.  In the example, _get_ would return the public key.

doc/man7/provider-keymgmt.pod
```diff
@@ -85,131 +71,109 @@ function pointer from a B<OSSL_DISPATCH> element named
...
-see L</Information Parameters>.
+OP_keymgmt_gettable_params() should return a constant array of
+descriptor B<OSSL_PARAM>, for parameters that OP_keymgmt_get_params()
+can handle.
 
-OP_keymgmt_gettable_domparam_params() should return a constant array
-of descriptor B<OSSL_PARAM>, for parameters that
-OP_keymgmt_get_domparam_params() can handle.
+If OP_keymgmt_gettable_params() is present, OP_keymgmt_get_params()
+must also be present.
```
mattcaswell: and vice versa right?

doc/man7/provider-keymgmt.pod
mattcaswell: Still needs fixing

include/openssl/core_numbers.h
```diff
@@ -333,109 +333,87 @@ OSSL_CORE_MAKE_FUNC(int, OP_kdf_set_ctx_params,
...
-/* Key information */
-# define OSSL_FUNC_KEYMGMT_GET_KEY_PARAMS           27
-# define OSSL_FUNC_KEYMGMT_GETTABLE_KEY_PARAMS      28
-OSSL_CORE_MAKE_FUNC(int, OP_keymgmt_get_key_params,
-                    (void *key, OSSL_PARAM params[]))
-OSSL_CORE_MAKE_FUNC(const OSSL_PARAM *, OP_keymgmt_gettable_key_params, (void))
+/* Import and export selections */
+# define OSSL_KEYMGMT_WANT_DOMPARAMS                   1
+# define OSSL_KEYMGMT_WANT_KEY                         2
+# define OSSL_KEYMGMT_WANT_BOTH                        3
```
mattcaswell: Possibly this is better expressed as `(OSSL_KEYMGMT_WANT_DOMPARAMS | OSSL_KEYMGMT_WANT_KEY)`

providers/implementations/keymgmt/dh_kmgmt.c
mattcaswell: `if ((selection & OSSL_KEYMGMT_WANT_PARAMS) != 0)`

providers/implementations/serializers/serializer_dh.c
```diff
@@ -51,7 +61,7 @@ int ossl_prov_print_dh(BIO *out, DH *dh, enum dh_print_type type)
     }
 
     p = DH_get0_p(dh);
-    g = DH_get0_p(dh);
+    g = DH_get0_g(dh);
```
mattcaswell: Oops!

mattcaswell (changes requested):

crypto/dsa/dsa_ameth.c
```diff
@@ -511,48 +511,48 @@ static size_t dsa_pkey_dirty_cnt(const EVP_PKEY *pkey)
...
+            || !ossl_param_bld_push_BN(&tmpl, OSSL_PKEY_PARAM_FFC_Q, q)
+            || !ossl_param_bld_push_BN(&tmpl, OSSL_PKEY_PARAM_FFC_G, g))
+            return 0;
+    }
+    if (selection != OSSL_KEYMGMT_WANT_DOMPARAMS) {
         /* A key must at least have a public part. */
         if (!ossl_param_bld_push_BN(&tmpl, OSSL_PKEY_PARAM_DSA_PUB_KEY,
                                     pub_key))
-            return NULL;
+            return 9;
```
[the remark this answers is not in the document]
levitte: Yes.

crypto/evp/keymgmt_meth.c
```diff
@@ -47,154 +47,99 @@ static void *keymgmt_from_dispatch(int name_id,
...
-        || (keymgmt->exportkey_types != NULL
-            && keymgmt->exportkey == NULL)
-        || (keymgmt->gettable_key_params != NULL
-            && keymgmt->get_key_params == NULL)) {
+    if (keymgmt->free == NULL
+        || keymgmt->new == NULL
+        || keymgmt->has_public_key == NULL
+        || keymgmt->has_private_key == NULL
+        || (keymgmt->gettable_params != NULL
+            && keymgmt->get_params == NULL)
```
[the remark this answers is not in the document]
levitte: This hasn't actually been made a rule yet...  I'm not 100% convinced yet

include/openssl/core_numbers.h
levitte: Coming up!

doc/man7/provider-keymgmt.pod
levitte: Done

providers/implementations/keymgmt/dh_kmgmt.c
levitte: Yup, that's what I ended up implementing.

providers/implementations/keymgmt/rsa_kmgmt.c
levitte: Changes have been done to allow the distinction while still maintaining the flexibility that the current EVP_PKEY functionality requires

crypto/evp/keymgmt_lib.c
```diff
@@ -74,12 +75,13 @@ void *evp_keymgmt_export_to_provider(EVP_PKEY *pk, EVP_KEYMGMT *keymgmt,
             return NULL;
 
         /* Otherwise, simply use it. */
-        provdata = pk->ameth->export_to(pk, keymgmt, want_domainparams);
+        if (!pk->ameth->export_to(pk, keydata, keymgmt, selection)) {
+            evp_keymgmt_freedata(keymgmt, keydata);
+            return NULL;
+        }
 
         /* Synchronize the dirty count, but only if we exported successfully */
```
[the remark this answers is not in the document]
levitte: Sure.  It's not *exacly* irrelevant, but the condition is more visible in the code.

crypto/evp/keymgmt_lib.c
```diff
@@ -74,12 +75,13 @@ void *evp_keymgmt_export_to_provider(EVP_PKEY *pk, EVP_KEYMGMT *keymgmt,
             return NULL;
```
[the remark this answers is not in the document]
levitte: Thanks.  Fixed

crypto/evp/keymgmt_meth.c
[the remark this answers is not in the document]
levitte: Not any more :wink:

include/openssl/core_numbers.h
levitte: Done

providers/implementations/serializers/serializer_dh.c
levitte: It wasn't the only embarrassing bug...

crypto/serializer/serializer_local.h
```diff
@@ -32,6 +32,9 @@ struct ossl_serializer_ctx_st {
     OSSL_SERIALIZER *ser;
     void *serctx;
 
+    /* One of OSSL_KEYMGMT_WANT_PARAMETERS and OSSL_KEYMGMT_WANT_KEY */
```
h-vetinari: Artefact of translating *_WANT_* to *_SELECT_*

crypto/evp/keymgmt_meth.c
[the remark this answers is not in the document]
levitte: I looked over all other method constructors, and apart from not being perfectly consistent (some constructors do not check this at all, but all others do check that either both are present or none of them are), it appears that this interface is the only one only checking one way.

So ok, I'll bow down and implement one of the schemes instead of this half-and-half scheme.

doc/man7/provider-keymgmt.pod
```diff
@@ -83,133 +64,143 @@ All these "functions" have a corresponding function type definition
...
-OP_keymgmt_exportdomparams() should extract values from the passed
-provider side domain parameter structure I<domparams> into the passed
-B<OSSL_PARAM> I<params>.
-Only the values specified in I<params> should be extracted.
+=over 4
 
-OP_keymgmt_importdomparam_types() should return a constant array of
-descriptor B<OSSL_PARAM>, for parameters that OP_keymgmt_importdomparams()
-can handle.
+=item B<OSSL_KEYMGMT_B_PRIVATE_KEY>
```
h-vetinari: Why `_B_*`? Wouldn't it be nicer to have uniform (if somewhat longer) names: 
```
OSSL_KEYMGMT_SELECT_PRIVATE_KEY
OSSL_KEYMGMT_SELECT_PUBLIC_KEY
OSSL_KEYMGMT_SELECT_KEY
OSSL_KEYMGMT_SELECT_DOMAIN_PARAMETERS
OSSL_KEYMGMT_SELECT_OTHER_PARAMETERS
OSSL_KEYMGMT_SELECT_PARAMETERS
OSSL_KEYMGMT_SELECT_ALL
```

rather than 
```
OSSL_KEYMGMT_B_PRIVATE_KEY
OSSL_KEYMGMT_B_PUBLIC_KEY
OSSL_KEYMGMT_SELECT_KEY
OSSL_KEYMGMT_B_DOMAIN_PARAMETERS
OSSL_KEYMGMT_B_OTHER_PARAMETERS
OSSL_KEYMGMT_SELECT_PARAMETERS
OSSL_KEYMGMT_SELECT_ALL
```

crypto/serializer/serializer_local.h
levitte: Thanks!  Will fix

doc/man7/provider-keymgmt.pod
levitte: I wanted a way to denote that some of them are distinct bits, while the other are collections of bits.  For example, `OSSL_KEYMGMT_SELECT_KEY` = `OSSL_KEYMGMT_B_PRIVATE_KEY | OSSL_KEYMGMT_B_PUBLIC_KEY`

richsalz: We have used ```_FLAG_``` where that is important.  I would do ```OSSL_FLAG_PRIVATE_KEY``` I think the "keymgmt" context is obvious :)

h-vetinari: I got that, but I still find the "`_B_`" confusing. Also, in some way, `{domain|other} parameterS` are already collections, no? (I get that they might still get returned in a single object, but the collection distinction is thin, IMO).

levitte: None of the bits are meant to denote one single item.  With RSA for example, the public key is *n* + *e* and the private key is *d* + *e* at a minimum, right?  So every bit represents a distinct category of stuff, and the `_SELECT_` values denote a set of categories.  Does saying it like this help?

Re `_B_` vs `_FLAG_`, I have no problems changing to the latter if that is more pleasing.  However, I will *not* drop `_KEYMGMT_`, that's just *asking* for future clashes.

levitte: Side note: regarding "domain" and "other" parameters, it comes from the dispute on how to view key parameters, where some are considered domain parameters (the FFC parameters, for example), while others aren't (PSS and OAEP parameters have been mentioned as such), so this distinction is really catering for those who do want to make a distinction...  as for "other", I simply couldn't figure out a better name.

richsalz: I strongly doubt it's asking for future clashes, but not my decision.

slontis: I would just leave all the names consistent.. They are all flags regardless of whether they are a collection of flags or not. Even doing this would be better..

OSSL_KEYMGMT_FLAG_DOMAIN_PARAMETERS
OSSL_KEYMGMT_FLAG_OTHER_PARAMETERS
OSSL_KEYMGMT_FLAG_PUBLIC_KEY
OSSL_KEYMGMT_FLAG_PRIVATE_KEY

OSSL_KEYMGMT_FLAGS_KEY
OSSL_KEYMGMT_FLAGS_PARAMETERS
OSSL_KEYMGMT_FLAGS_ALL

levitte: Actually, I'm not sure I like where this is going. The *selection* is fairly special purpose, while "FLAG" doesn't quite reflect that.

OSSL_KEYMGMT_SEL_DOMAIN_PARAMETERS
OSSL_KEYMGMT_SEL_OTHER_PARAMETERS
OSSL_KEYMGMT_SEL_PUBLIC_KEY
OSSL_KEYMGMT_SEL_PRIVATE_KEY

OSSL_KEYMGMT_SEL_ALL_KEYS
OSSL_KEYMGMT_SEL_ALL_PARAMETERS
OSSL_KEYMGMT_SEL_ALL

Something like that? SEL or SELECT?

slontis: +1 SELECT

slontis: I actually like what @h-vetinari suggested above. ALL_KEYS could be _KEYPAIR..
Either way I dont want you getting stuck on names, when there are lots of other things to do.

levitte: Hahaha, so after whining and moaning for a bit, I'll end up doing almost exactly what @h-vetinari suggests... I feel a bit embarrassed having whined 😉

levitte: Okie, done!

levitte: (or almost...)

crypto/evp/keymgmt_lib.c
```diff
@@ -91,42 +92,33 @@ void *evp_keymgmt_export_to_provider(EVP_PKEY *pk, EVP_KEYMGMT *keymgmt,
...
+            if (evp_keymgmt_export(exp_keymgmt, exp_keydata,
+                                   OSSL_KEYMGMT_SELECT_ALL,
+                                   &try_import, &import_data))
+                break;
+
+            /* If there was an error, bail out */
+            evp_keymgmt_freedata(keymgmt, keydata);
+            return NULL;
         }
     }
```
slontis: Is it supposed to return a void * here?

crypto/evp/keymgmt_meth.c
```diff
@@ -280,3 +207,81 @@ void EVP_KEYMGMT_names_do_all(const EVP_KEYMGMT *keymgmt,
...
+int evp_keymgmt_has(const EVP_KEYMGMT *keymgmt, void *keydata, int selection)
+{
+    /* This is mandatory, no need to check for its presence */
+    return keymgmt->has(keydata, selection);
+}
+
+int evp_keymgmt_validate(const EVP_KEYMGMT *keymgmt, void *keydata,
+                         int selection)
+{
+    return keymgmt->validate(keydata, selection);
```
slontis: this doesnt look mandatory so needs the NULL check?

crypto/evp/pmeth_gn.c
```diff
@@ -247,14 +251,16 @@ const OSSL_PARAM *EVP_PKEY_param_fromdata_settable(EVP_PKEY_CTX *ctx)
...
     return NULL;
 }
 
 const OSSL_PARAM *EVP_PKEY_key_fromdata_settable(EVP_PKEY_CTX *ctx)
 {
     /* We call fromdata_init to get ctx->keymgmt populated */
     if (fromdata_init(ctx, EVP_PKEY_OP_UNDEFINED))
-        return evp_keymgmt_importdomparam_types(ctx->keymgmt);
+        return evp_keymgmt_import_types(ctx->keymgmt,
+                                        OSSL_KEYMGMT_SELECT_KEYPAIR);
```
slontis: should this be SELECT_ALL ?

doc/man3/EVP_PKEY_check.pod
```diff
@@ -18,7 +18,7 @@ EVP_PKEY_private_check, EVP_PKEY_pairwise_check
 
 =head1 DESCRIPTION
 
-EVP_PKEY_param_check() validates the domain parameters component of the key
+EVP_PKEY_param_check() validates the parameters component of the key
```
slontis: I would probably leave the domain in this. That should be what is getting validated - not the other parameters

doc/man3/d2i_PrivateKey.pod
```diff
@@ -36,15 +36,15 @@ use any key specific format or PKCS#8 unencrypted PrivateKeyInfo format. The
 B<type> parameter should be a public key algorithm constant such as
 B<EVP_PKEY_RSA>. An error occurs if the decoded key does not match B<type>.
 d2i_PublicKey() does the same for public keys.
-d2i_KeyParams() does the same for domain parameter keys.
+d2i_KeyParams() does the same for key parameter.
```
slontis: parameters

include/crypto/evp.h
```diff
@@ -548,25 +548,23 @@ struct evp_pkey_st {
...
     /* == Provider attributes == */
     /*
-     * To support transparent export/import between providers that
-     * support the methods for it, and still not having to do the
-     * export/import every time a key or domain params are used, we
-     * maintain a cache of imported key / domain params, indexed by
-     * provider address.  pkeys[0] is *always* the "original" data.
+     * To support transparent export/import between providers that support
+     * the methods for it, and still not having to do the export/import
+     * every time a key object is, we maintain a cache of imported key
```
slontis: is changed

providers/implementations/keymgmt/dh_kmgmt.c
```diff
@@ -128,70 +135,134 @@ static int key_to_params(DH *dh, OSSL_PARAM_BLD *tmpl)
...
+    OSSL_PARAM_BN(OSSL_PKEY_PARAM_DH_PRIV_KEY, NULL, 0)
+static const OSSL_PARAM dh_all_types[] = {
+    DH_IMEXPORTABLE_PARAMETERS,
+    DH_IMEXPORTABLE_PUBLIC_KEY,
+    DH_IMEXPORTABLE_PRIVATE_KEY,
+    OSSL_PARAM_END
+};
+static const OSSL_PARAM dh_parameter_types[] = {
+    DH_IMEXPORTABLE_PARAMETERS,
+    DH_IMEXPORTABLE_PUBLIC_KEY,
```
slontis: Err - block copy bug.

providers/implementations/keymgmt/dh_kmgmt.c
```diff
@@ -128,70 +135,134 @@ static int key_to_params(DH *dh, OSSL_PARAM_BLD *tmpl)
...
+    DH_IMEXPORTABLE_PRIVATE_KEY,
+    OSSL_PARAM_END
+};
+static const OSSL_PARAM dh_key_types[] = {
+    DH_IMEXPORTABLE_PARAMETERS,
```
slontis: block copy bug - tables should be different

providers/implementations/keymgmt/dh_kmgmt.c
```diff
@@ -128,70 +135,134 @@ static int key_to_params(DH *dh, OSSL_PARAM_BLD *tmpl)
...
+    OSSL_PARAM_END
+};
+static const OSSL_PARAM *dh_types[] = {
+    NULL,                        /* Index 0 = none of them */
+    dh_parameter_types,          /* Index 1 = parameter types */
+    dh_key_types,                /* Index 2 = key types */
+    dh_all_types                 /* Index 3 = 1 + 2 */
+};
+
+static const OSSL_PARAM *dh_imexport_types(int selection)
```
slontis: I have no idea what the im notation means..

providers/implementations/keymgmt/dh_kmgmt.c
```diff
@@ -128,70 +135,134 @@ static int key_to_params(DH *dh, OSSL_PARAM_BLD *tmpl)
...
+    if ((selection & OSSL_KEYMGMT_SELECT_KEYPAIR) != 0)
+        type_select += 2;
+    return dh_types[type_select];
+}
+
+static const OSSL_PARAM *dh_import_types(int selection)
+{
+    return dh_imexport_types(selection);
+}
+
```
slontis: NIT: extra line

providers/implementations/keymgmt/dsa_kmgmt.c
```diff
@@ -136,70 +142,135 @@ static int key_to_params(DSA *dsa, OSSL_PARAM_BLD *tmpl)
...
+    OSSL_PARAM_BN(OSSL_PKEY_PARAM_DSA_PRIV_KEY, NULL, 0)
+static const OSSL_PARAM dsa_all_types[] = {
+    DSA_IMEXPORTABLE_PARAMETERS,
+    DSA_IMEXPORTABLE_PUBLIC_KEY,
+    DSA_IMEXPORTABLE_PRIVATE_KEY,
+    OSSL_PARAM_END
+};
+static const OSSL_PARAM dsa_parameter_types[] = {
+    DSA_IMEXPORTABLE_PARAMETERS,
+    DSA_IMEXPORTABLE_PUBLIC_KEY,
```
slontis: table is incorrect

providers/implementations/keymgmt/dsa_kmgmt.c
```diff
@@ -136,70 +142,135 @@ static int key_to_params(DSA *dsa, OSSL_PARAM_BLD *tmpl)
...
+    DSA_IMEXPORTABLE_PRIVATE_KEY,
+    OSSL_PARAM_END
+};
+static const OSSL_PARAM dsa_key_types[] = {
+    DSA_IMEXPORTABLE_PARAMETERS,
+    DSA_IMEXPORTABLE_PUBLIC_KEY,
```
slontis: table is incorrect

providers/implementations/keymgmt/dsa_kmgmt.c
```diff
@@ -136,70 +142,135 @@ static int key_to_params(DSA *dsa, OSSL_PARAM_BLD *tmpl)
...
+    if ((selection & OSSL_KEYMGMT_SELECT_KEYPAIR) != 0)
+        type_select += 2;
+    return dsa_types[type_select];
+}
+
+static const OSSL_PARAM *dsa_import_types(int selection)
+{
+    return dsa_imexport_types(selection);
+}
+
```
slontis: extra blankline

providers/implementations/keymgmt/rsa_kmgmt.c
```diff
@@ -20,14 +20,20 @@
...
+static OSSL_OP_keymgmt_has_fn rsa_has;
+static OSSL_OP_keymgmt_validate_fn rsa_validate;
+static OSSL_OP_keymgmt_import_fn rsa_import;
+static OSSL_OP_keymgmt_import_types_fn rsa_import_types;
+static OSSL_OP_keymgmt_export_fn rsa_export;
+static OSSL_OP_keymgmt_export_types_fn rsa_export_types;
 
 #define RSA_DEFAULT_MD "SHA256"
+#define RSA_POSSIBLE_SELECTIONS                 \
+    (OSSL_KEYMGMT_SELECT_KEYPAIR | OSSL_KEYMGMT_SELECT_ALL_PARAMETERS)
```
slontis: DOMAIN PARAMETERS is not a possibility fo rthis one - so specify OTHER instead of ALL_PARAMETERS

providers/implementations/serializers/serializer_dsa_pub.c
```diff
@@ -106,15 +116,20 @@ static int dsa_pub_pem(void *ctx, void *dsa, BIO *out,
 static int dsa_pub_print_data(void *ctx, const OSSL_PARAM params[], BIO *out,
```
slontis: If these functions remain almost identical then we may want to collapse them at some time in the future.

slontis (commented):
@levitte - apart from a few small changes this is looking good.
It certainly looks a bit nicer now that some of the if (domparams) blah1 else blah2 has been removed.

providers/implementations/keymgmt/dh_kmgmt.c
levitte: Import+export. Imexport

providers/implementations/serializers/serializer_dsa_pub.c
levitte: Yeah, I was thinking exactly that. It's for another refactoring, another PR

crypto/evp/keymgmt_meth.c
levitte: Ah, we get back to this that I was pondering a few days ago and then promptly forgot.

Thank you.

So, the question is, what should it return when the implementation hasn't supplied a validation function?  Between TRUE and FALSE, I haven't been able to decide...

Another option is to return another value that indicates that the validation function isn't present (unsupported).

What are your thoughts on that?

crypto/evp/pmeth_gn.c
levitte: This is `EVP_PKEY_key_fromdata_settable`, note **`key`** as opposed to **`param`** in `EVP_PKEY_param_fromdata_settable`.

If we collapse the `fromdata` functions to not distinguish between params and the key itself, then I can definitely see ussing SELECT_ALL.  That's for another PR, though...

doc/man3/EVP_PKEY_check.pod
levitte: I'm thinking that it's up to the provider to decide exactly what gets validated.  For *our* stuff, it absolutely makes sense to limit that to domain parameters, but other stuff we haven't heard about may have other parameters that aren't necessarily *domain* they want to include in the validation.
(that is, if everyone even agree on what constitutes *domain* parameters...  recent discussions suggest that is not always the case.  Me, I'm actually rather confused on the matter)

crypto/evp/keymgmt_meth.c
slontis: I dont know if there is a good answer.. If it doesnt have a method then maybe return 1. Not sure we want to head down the return -2 path quite yet..

i.e: we assume its valid..

doc/man3/EVP_PKEY_check.pod
slontis: The whole point of validation is that is comes from specs (particularly the SP800-56A/B keyagreement spec's - which is all about assurance that keys are valid. i.e. validation of domain params/ pub/priv and keypair checks. I dont see there being these other types of checks.. If there are then we add them when required. The other_parameter's does not fit into this scheme at all.

doc/man3/d2i_PrivateKey.pod
levitte: Fixed

include/crypto/evp.h
levitte: Fixed

providers/implementations/keymgmt/dh_kmgmt.c
levitte: Oops.  Fixed

providers/implementations/keymgmt/dh_kmgmt.c
levitte: Fixed

providers/implementations/keymgmt/dh_kmgmt.c
levitte: Removed

providers/implementations/keymgmt/dsa_kmgmt.c
levitte: Damn I was tired when I put them together...

Fixed

providers/implementations/keymgmt/dsa_kmgmt.c
levitte: Fixed

providers/implementations/keymgmt/dsa_kmgmt.c
levitte: Removed

providers/implementations/keymgmt/rsa_kmgmt.c
levitte: I'm frankly still unsure about the distinction, it feels a bit synthetic to make a point...  but ok, since we have it, I fix it.

doc/man3/EVP_PKEY_check.pod
levitte: This falls under the category of "know your provider", methinks, and also "know your algorithm"...

... which also makes me think that we're being sloppy regarding documentation.  Just like we have `EVP_MAC-something.pod` and `EVP_KDF-whatever.pod`, we should have a few `EVP_KEYMGMT-name.pod` and similar, where we describe all sorts of things that are specific to our implementation of the different key algorithms / types.

(and yeah, there should be `EVP_SIGNATURE-algo.pod` too, and...)

providers/implementations/keymgmt/dh_kmgmt.c
```diff
@@ -208,6 +208,8 @@ static int dh_export(void *keydata, int selection, OSSL_CALLBACK *param_cb,
     return ok;
 }
 
+/* IMEXPORT = IM- + EXPORT */
```
slontis: = IMPORT + EXPORT

crypto/evp/keymgmt_meth.c
levitte: Okie, I'll go with that

levitte: Fixed (and the other ones that didn't check as well)

providers/implementations/keymgmt/dh_kmgmt.c
levitte: Hmm?  That hyphenated form is unknown in English?  We do that all the time in Swedish...

levitte: Changed

levitte:
I believe I've finally addressed all concerns so far.  Anything else?

levitte:
Rebased on top of a fresh master

slontis (approved):

mattcaswell (approved):
Approved whether or not you make a change as a result of my question or not.

include/openssl/core_numbers.h
```diff
@@ -333,109 +333,93 @@ OSSL_CORE_MAKE_FUNC(int, OP_kdf_set_ctx_params,
...
-OSSL_CORE_MAKE_FUNC(const OSSL_PARAM *, OP_keymgmt_gettable_key_params, (void))
-
-/* Discovery of supported operations */
-# define OSSL_FUNC_KEYMGMT_QUERY_OPERATION_NAME     40
-OSSL_CORE_MAKE_FUNC(const char *,OP_keymgmt_query_operation_name,
+/* Key data subset selection - individual bits */
+# define OSSL_KEYMGMT_SELECT_PRIVATE_KEY            0x01
+# define OSSL_KEYMGMT_SELECT_PUBLIC_KEY             0x02
+# define OSSL_KEYMGMT_SELECT_DOMAIN_PARAMETERS      0x04
+# define OSSL_KEYMGMT_SELECT_OTHER_PARAMETERS       0x80
```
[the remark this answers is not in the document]
levitte: Yes.  I deliberately made that the "last bit" because it's kind of a "the rest" indicator.  I want it to be natural to insert new parameter categories *before* this one.

levitte:
There was a slight documentation fix in `doc/man7/provider-keymgmt.pod`.  Please confirm that previous approvals hold, or re-approve if you prefer (the latter resets the 24h wait)

mattcaswell:
Reconfirm my approval.

kaduk:
> Actually, _all_ or _any_ could be left to the implementation to decide. For example, the legacy implementations often only compare the public key when asked to compare a key, and the EVP layer doesn't really care, it only cares about the key pair (oh, and parameters if there are any, but that's less relevant here) for EVP_PKEY comparisons. This does make sense, 'cause for valid keys, a matching public key implies a matching private key.

I'm not sure whether I caught all/any (pun intended) followup discussion on this one.  I'm also not sure I understood the distinction properly -- if the "all" sense was in use, that would mean that if I wanted to get information about multiple things I'd end up making multiple calls, since each call would match "all" the flags I specify, and if I specify flags for "two different things" at once, nothing will match on that call?

slontis:
reconfirmed also..

levitte:
> I'm also not sure I understood the distinction properly -- if the "all" sense was in use, that would mean that if I wanted to get information about multiple things I'd end up making multiple calls, since each call would match "all" the flags I specify, and if I specify flags for "two different things" at once, nothing will match on that call?

Yeah, and I think there are cases when this is counter productive.  For example, if you ask for an export with the selection OSSL_KEYMGMT_SELECT_KEYPAIR and the provider side key currently only has the public key, should you get whatever's there, or should you only get anything if both halves are present?  I would expect to get whatever's there within the selection I asked for.

slontis:
>  I would expect to get whatever's there within the selection I asked for.

If you really need to make sure a component exists - you can use the has method...

include/openssl/core_numbers.h
```diff
@@ -333,109 +333,93 @@ OSSL_CORE_MAKE_FUNC(int, OP_kdf_set_ctx_params,
...
+
+/* Key data subset selection - combinations */
+# define OSSL_KEYMGMT_SELECT_ALL_PARAMETERS     \
+    ( OSSL_KEYMGMT_SELECT_DOMAIN_PARAMETERS     \
+      | OSSL_KEYMGMT_SELECT_OTHER_PARAMETERS)
+# define OSSL_KEYMGMT_SELECT_KEYPAIR            \
+    ( OSSL_KEYMGMT_SELECT_PRIVATE_KEY | OSSL_KEYMGMT_SELECT_PUBLIC_KEY )
+# define OSSL_KEYMGMT_SELECT_ALL                \
```
paulidale: Why not (-1) here?  Negative one captures all and covers an future additions.

providers/implementations/keymgmt/dh_kmgmt.c
```diff
@@ -128,70 +135,132 @@ static int key_to_params(DH *dh, OSSL_PARAM_BLD *tmpl)
...
-    return ret;
+    return ok;
 }
 
-/*
- * Same function for domain parameters and for keys.
- * "dpk" = "domain parameters & keys"
- */
-static ossl_inline int dh_get_dpk_params(void *key, OSSL_PARAM params[])
+/* IMEXPORT = IMPORT + EXPORT */
```
paulidale: Ugh.  It read as a badly spelt _unexport_.
I can live without changing this.

test/keymgmt_internal_test.c
```diff
@@ -145,7 +145,7 @@ static int test_pass_rsa(FIXTURE *fixture)
     BIGNUM *bn1 = NULL, *bn2 = NULL, *bn3 = NULL;
     EVP_PKEY *pk = NULL;
     EVP_KEYMGMT *km1 = NULL, *km2 = NULL;
-    void *provdata = NULL;
+    void *provkey = NULL;
     /*
```
paulidale: Blank line before.

paulidale (changes requested):
Looks like an improvement.  A few nits and some (existing) unchecked return values.  The latter need to be fixed.

crypto/dh/dh_ameth.c
```diff
@@ -482,51 +482,46 @@ static size_t dh_pkey_dirty_cnt(const EVP_PKEY *pkey)
...
+    /* A key must at least have a public part. */
+    if (!ossl_param_bld_push_BN(&tmpl, OSSL_PKEY_PARAM_DH_PUB_KEY, pub_key))
+        return 0;
+    if (priv_key != NULL) {
+        if (!ossl_param_bld_push_BN(&tmpl, OSSL_PKEY_PARAM_DH_PRIV_KEY,
+                                    priv_key))
+            return 0;
     }
 
     params = ossl_param_bld_to_param(&tmpl);
```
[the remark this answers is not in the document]
levitte: Fixed

crypto/dsa/dsa_ameth.c
```diff
@@ -511,48 +511,43 @@ static size_t dsa_pkey_dirty_cnt(const EVP_PKEY *pkey)
...
+    if (!ossl_param_bld_push_BN(&tmpl, OSSL_PKEY_PARAM_DSA_PUB_KEY,
+                                pub_key))
+        return 0;
+    if (priv_key != NULL) {
+        if (!ossl_param_bld_push_BN(&tmpl, OSSL_PKEY_PARAM_DSA_PRIV_KEY,
+                                    priv_key))
+            return 0;
     }
 
     params = ossl_param_bld_to_param(&tmpl);
```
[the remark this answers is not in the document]
levitte: Fixed

include/openssl/core_numbers.h
levitte: I preferred to be explicit.

test/keymgmt_internal_test.c
levitte: No.  This comment is about the next variable being defined.

levitte:
@paulidale's concerns addressed.

paulidale (approved):
All good now.

levitte:
Merged!!!!!

68552cdef7631191e77315e0faeb42c6893cafe3 Reorganize the internal evp_keymgmt functions
b305452f69fc97c586f2f9310014e332ae1d5cd5 Redesign the KEYMGMT libcrypto <-> provider interface - the basics
8dd5c603b60d8dd2907fea6c9cdefff20b787845 Adapt existing KEYMGMT implementations to the redesigned interface
72ec9648fee5612a9cf83ed2df1c6c0e240fa7de Adapt test/keymgmt_internal_test.c to the redesigned interface
32b0645cf9f1e5c11ed1cb0fc1ff1dcd47541a93 Adapt existing SERIALIZER implementations to the redesigned interface
273a67e345567e4dc95bd962605a3e34532f2649 KEYMGMT: Require both get_params and gettable_params, or none

levitte:
I'm very happy this went so well, this was a big stone lifting for me.  Thank you all for the feedback!

slontis:
good work richard..

paulidale:
Agreed.  This is a significant clean up.

levitte:
> good work richard..

Thanks, but I actually view it as good work on all of us.

romen:
@levitte

If `OP_keymgmt_import()` is called with `selection = OSSL_KEYMGMT_SELECT_PRIVATE_KEY` according to the documentation and intuition, shouldn't the provider only import the private key from the passed params? Same goes for any other atomic selector.

I believe that in general inside the provider implementation the use of the aggregated selectors should be close to zero (the only one that seems useful at the moment is the KEYPAIR selector for `validate()` as we document that in that specific case we should validate that private and public keys of the keypair match.

slontis:
Currently the private key implies the public key must also exist currently. I dont think this is right either, but there are other places that are assuming this also at the moment. This may bite us in the future (say with p11) and then be fairly hard to fix properly.

romen:
Then again, can we import a private key without importing the public key AND the domain parameters?
It would violate the convention that an `EVP_PKEY` object MUST either contain domain parameters only (for the key types that have domain parameters) or at least the public key part (which in key types that have domain parameters is defined only in the context of associated domain parameters, that MUST thus then be present).
This means that it is not possible to complete the import if we abide the selector when the selector is set to `OSSL_KEYMGMT_SELECT_PRIVATE_KEY` as we shouldn't (and most of the time cannot anyway) create keydata containing only a private key as we need also the public part which cannot be defined without the associated domain parameters..
And for export the situation is similar, we cannot really export only the private key into what is presumably going to be an `EVP_PKEY` object, without exporting also the public part and the associated domain parameters.

I feel that for import/export the selectors are creating more confusion than they should, and the current documentation does not clarify enough what the provider implementer should do about the selection bits for import and export.

levitte:
If there are bits that seem vague, it can certainly be changed for better clarity.  Generally speaking, I agree that there should be precision.  PR welcome (if I don't get to it first)!

That being said, the interpretation of *selection* is very much up to the provider.  The EVP code can only express a desire, then it's up to the provider code to respond to that desire in what way it deems appropriate.  Basically, the EVP code has *no* control of what stuff the provider side key actually contains.  However, a provider must also be smart, in that it needs to ensure that the provider side keys are actually usable for operations it's supposed to be usable with (signature, key exchange, whatever...)

doc/man7/provider-keymgmt.pod
```diff
@@ -219,13 +214,43 @@ returns NULL, the caller is free to assume that there's an algorithm
...
+I<param_cb> with that array as well as the given I<cbarg>.
+
+OP_keymgmt_import_types() should return a constant array of descriptor
+B<OSSL_PARAM> for data indicated by I<selection>, for parameters that
+OP_keymgmt_import() can handle.
+
+OP_keymgmt_export_types() should return a constant array of descriptor
+B<OSSL_PARAM> for data indicated by I<selection>, that the
+OP_keymgmt_export() callback can expect to receive.
+
```
romen: So looking at current `dh_kmgmt.c`, the implementation does not match the documentation.

Theoretically `libcrypto` can ask the provider to import only the private key in `keydata`, but that is not possible and no check is done to prevent it.

If the rule of thumb is that the `OSSL_PARAM` array resulting from `OP_keymgmt_export()` with `selection = FOO` must be at least capable of importing a valid `keydata` object when the same `OSSL_PARAM` array is passed to `OP_keymgmt_import()` with the same value for `selection`, this should be stated clearly.

We should also state what should the provider do if `libcrypto` sets `selection` to some combination that makes it impossible to do what stated above.
- should `OP_keymgmt_export()` ignore the restrictions set in `selection` and export whatever it wants?
- should `OP_keymgmt_export()` alternatively return an error as soon as an unsupported selection is requested?
- should `OP_keymgmt_import()` return an error as soon as an unsupported selection is requested?

The previous separation between `OP_keymgmt_importkey()`+`OP_keymgmt_exportkey()` and `OP_keymgmt_importdomparams()`+`OP_keymgmt_exportdomparams()` was IMHO less ambiguous and easier for provider implementers.

At this point I am not sure anymore when import and export can be triggered, but I see the potential risk for a key being generated in the provider, exported to `libcrypto` only partially because of `selection`, some `dirty_cnt` change is triggered in `libcrypto` which then triggers an import in the provider again resulting in loss of information or error.

romen:
Also, in contrast with DH and DSA, for EC we have separate objects to represent a domain parameters object (`EC_GROUP`) and a key object (`EC_KEY`), and while it would be possible to pass around an `EC_KEY` object not fully initialized, we shouldn't really do that.

The previous separation between importkey and importdomparams (and the equivalent exports) supported cleanly this model, but unifying them makes things quite ugly on the provider side!
Also this is another place where we can see the clear distinction between domain parameters and key parameters.

- `ECDH_COFACTOR_MODE` is a key parameter and can be set only on an `EC_KEY` object
- `EC_GROUP` is a standalone representation of the domain parameters under which an `EC_KEY` can be instantiated

levitte:
So the weird part is that, for EC keys, the assumptions done at the EVP_PKEY level were bad as well?  After all, exactly the same thing applies there, that the underlying backend key that an EVP_PKEY points at may contain all sorts of data combinations.  These are the combinations I have detected so far:

- parameters only
- public key only
- public key + private key
- parameters + public key
- parameters + public key + private key

So yeah, at the EVP level, *we already assume that the underlying key data is in all sorts of shapes.

But I hear you, the possible interpretations of *selection* needs a little bit more write-up.

romen:
> So the weird part is that, for EC keys, the assumptions done at the EVP_PKEY level were bad as well? After all, exactly the same thing applies there, that the underlying backend key that an EVP_PKEY points at may contain all sorts of data combinations. These are the combinations I have detected so far:
> 
> * parameters only

(DH, DSA, EC support this, but in EC parameters only is represented as an `EC_GROUP` (from which an `EC_KEY` can be instantiated), while DH and DSA have a single type for a proper key and only the domain parameters to instantiate a proper key)

> * public key only
> * parameters + public key

(only RSA, X/Ed25519 and X/Ed448 can have public key only, as there are no "domain parameters" or the "domain parameters" are embedded in the key type; but technically for the very same reason, public key only and public key + domain parameters for these key types is the same thing, so I don't really think public key only really exists)

> * public key + private key
> * parameters + public key + private key

(as above, arguably public key does not exist without domain parameters, although domain parameters could be an empty set, so these 2 items are conceptually one and the same)

I don't know how to factor the "other parameters" in this discussion, as they are still a bit alien to me.
In EC an "other parameter" would be the "cofactor_mode" that, in terms of legacy behavior, can be only set on an `EC_KEY` and not on an `EC_GROUP` so it requires different handling than "domain parameters".
From the previous conversations I understand that the "other parameters" for the fancy RSA keys (OEAP, PSS) are instead more akin to "domain parameters"?

> 
> So yeah, at the EVP level, *we already assume that the underlying key data is in all sorts of shapes.
> 
> But I hear you, the possible interpretations of _selection_ needs a little bit more write-up.

doc/man7/provider-keymgmt.pod
levitte: Uhmmm...  have you read what's written in [provider-keymgmt(7) / Key Object Import and Export Functions](https://github.com/openssl/openssl/blob/master/doc/man7/provider-keymgmt.pod#key-object-import-and-export-functions).  Doesn't "indicated by _selection_" tell you that _selection_ is not supposed to be ignored?

Of course, the provider code might ignore _selection_, but that seems rather dumb, doesn't it?  But most of all, how are we going to stop that?

levitte:
> arguably public key does not exist without domain parameters

I think you should study this code:

https://github.com/openssl/openssl/blob/87d3bb8e861e54e2549e848a1adee6d2df9ed603/apps/ca.c#L1922-L1925

levitte:
> From the previous conversations I understand that the "other parameters" for the fancy RSA keys (OEAP, PSS) are instead more akin to "domain parameters"?

Please don't ask me to explain the difference between domain parameters and "other" parameters, I find this part utterly confusing, and considering what has been said in diverse discussion, no one agrees on an absolutely clear definition.  All I know is that at least one person tells me that the FFC parameters associated with DH and DSA keys *are* domain parameters, and that the "fancy" RSA parameters - PSS and OAEP - *are not* domain parameters.
Me, I kinda go "ok", shrug a bit, and move on.  From a coding point of view, the distinction seems quite useless.

levitte:
As for EC specific stuff, maybe we should discuss that where it belongs, i.e. in #10631.  I'll get there, just not now.

PR #8797, opened 2019-04-19
WIP: master QUIC support

tmshort:
Port the BoringSSL QUIC-related APIs over to OpenSSL.

##### Checklist

- [X] documentation is added or updated
- [X] tests are added or updated

PR #11011, opened 2020-02-04
WIP: config as a perl script

richsalz:
Writing config as a Perl script.  Hopefully shorter and easier to
understand.

Worth doing?


levitte:
I would write it as a module, so it can be called from inside Configure.

richsalz:
Do you support this effort?  Do you think it worthwhile?  Happy to do it as a module.

levitte:
If it becomes a module, I support the effort.
I can handle the call from Configure if you wish (I have some idea regarding the precise spot to do the call, and that's not necessarily the first thing to do).

BTW, we will still need a `config`, which basically just calls `Configure`, something like this:

``` shell
#! /bin/sh

HERE=$(dirname $0)
exec $HERE/Configure "$@"
```

richsalz:
First checkpoint; works on my Mac platform.

config
```diff
@@ -1,937 +1,854 @@
...
+}
+
+sub get_sco_type() {
+    my $REL = shift;
+
+    if ( -f "/etc/kconfig" ) {
+        return "${MACHINE}-whatever-isc4" if $REL eq '4.0' || $REL eq '4.1';
+    } else {
+        return "whatever-whatever-sco3" if $REL eq '3.2v4.2';
+        return "whatever-whatever-sco5" if $REL =~ m@3.2v5.0.*@;
```
levitte: Don't forget to escape explicit periods.

config
```diff
@@ -1,937 +1,854 @@
...
+        }
+    }
+
+    # Now pattern-match
+    my $sys = "${SYSTEM}:${RELEASE}:${VERSION}:${MACHINE}";
+
+    # Simple cases
+    foreach my $tuple ( @$simple_sys_list ) {
+        my $pat = @$tuple[0];
+        next if $sys !~ /$pat/;
```
levitte: Add the ^ and $ anchors

config
```diff
@@ -1,937 +1,854 @@
...
+        my $result = @$tuple[1];
+        $result =~ s/\$\{MACHINE\}/${MACHINE}/;
+        return $result;
+    }
+
+    # Complex cases.
+    foreach my $tuple ( @$complex_sys_list ) {
+        my $pat = @$tuple[0];
+        next if $sys !~ /$pat/;
```
levitte: ^ and $ here as well

config
```diff
@@ -1,937 +1,854 @@
...
+##
+##  SKIP ALGOLRITHMS THAT DON'T EXIST. XXX NEEDED?
+my @cryptodir = (
+    'aes', 'aria', 'bf', 'camellia', 'cast', 'des', 'dh', 'dsa', 'ec', 'hmac',
+    'idea', 'md2', 'md5', 'mdc2', 'rc2', 'rc4', 'rc5', 'ripemd', 'rsa',
+    'seed', 'sha', 'sm2', 'sm3', 'sm4'
+);
+foreach my $d ( @cryptodir ) {
+    $options .= " no-$d" if ! -d "$THERE/crypto/$d";
+}
```
levitte: Oh wow... This is really old stuff, hardly relevant any more. Or does anyone still do that, remove whole crypto subsystems from the source?

config
```diff
@@ -1,937 +1,854 @@
...
+    }
+    foreach my $dir ( split(/:/, $ENV{PATH}) ) {
+        my $d = $dir eq '' ? '.' : $dir;
+        return "$d/perl${EXE}"
+            if -x "$d/perl${EXE}" && okrun("$d/perl${EXE} -e 'exit($]<5.0);");
+    }
+    die "You need Perl 5." if $PERL eq '';
+}
+
+$PERL = &find_perl_5();
```
levitte: This is quite unnecessary, since we have the variable `$^X`. See https://perldoc.perl.org/perlvar.html

config
paulidale: I very much suspect the answer is _yes_.

levitte: That check should be entirely in Configure, then...

mattcaswell:
Drive by comment: It would be nice if this supported Windows VC builds. Then we can remove the inconsistency that for almost all platforms you just run config and it works....except Windows.

config
richsalz: yes, fixed three places.

config
richsalz: added the anchors.

config
richsalz: added the anchors.

config
richsalz: I'll leave it now, and add a comment saying remove when this is moved into configure.

config
richsalz: are you saying remove those lines and do 
```
$PERL = $^X;
```
?

levitte: Remove them, their only purpose is to find a PERL to run Configure in, but since you're going to convert this to a module that Configure can `use`, those lines are pointless.

Either way, it's really silly to go search for perl when you're running a perl script, since the information is already there...

richsalz: done.

richsalz:
> Drive by comment: It would be nice if this supported Windows VC builds. 

It now does.  If running on windows, it requires one of the VC-xxx parameters (case ignored but turned into uppercase) to be provided.

config
```diff
@@ -1,937 +1,876 @@
...
+    [ 'MINGW64.*:.*x86_64',   '${MACHINE}-whatever-mingw64' ],
+    [ 'MINGW',                '${MACHINE}-whatever-mingw' ],
+    [ 'CYGWIN',               '${MACHINE}-pc-cygwin' ],
+    [ 'vxworks',              '${MACHINE}-whatever-vxworks' ],
+    [ 'Darwin:.*Power',       'ppc-apple-darwin' ],
+    [ 'Darwin:.*x86_64',      'x86_64-apple-darwin' ],
+    [ 'Darwin:',              'i686-apple-darwin' ],
+];
+
+# More complex cases that require riun-time code.
```
beldmit: Nit: runtime

richsalz: :)  Could also be "ruin" time.

beldmit: Had this suspicion too :)

richsalz:
Okay, this took a couple of days (the time I had alloted), and I learned some perl, and my last doc fixes have gone in so it's time for me to move on to more serious things (like those that my employer needs from this project).  In other words, I have taken this as far as I want to.  If someone wants to pick it up, particularly @levitte, please feel free.  Note that taking this could be risky -- certainly more risky than #8797 :) -- because it is unlikely to get extensive testing across all the platforms that are currently supported.

To whoever follows, good luck!

levitte:
Looking at this...

First of all, not bad at all, this is a good start. 

However, I would like to discuss the desired endgame. As it is right now, we have two scripts (really, three, I'll get back on that), `config` and `Configure`, where the first makes an educated guess.

*My* desire is to reduce it to just one, and in most cases, this is more or less what we promote anyway, so what I would like to do in the end is to merge `config` and `Configure` into just `config`. Does @openssl have a different opinion?

-----

I mentioned a third script. Not all operating systems can run perl scripts the way you do it on Unix, some require that you do type `perl script` on the command line, or write a wrapper script in their standard command line language. `command.com` is exactly that for VMS. I wonder what the corresponding thing to do would be for Windows cmd.

mspncp:
> I mentioned a third script. Not all operating systems can run perl scripts the way you do it on Unix, some require that you do type `perl script` on the command line, or write a wrapper script in their standard command line language. `command.com` is exactly that for VMS. I wonder what the corresponding thing to do would be for Windows cmd.

My Active Perl installed a handler for the .pl extension. You can just run it without prefixing it with the interpreter:

```
C:\Users\msp>type hello_world.pl
print("hello world");

C:\Users\msp>hello_world.pl
hello world
```

For completeness, here's the version:

```
C:\Users\msp\tmp>perl --version

This is perl 5, version 26, subversion 1 (v5.26.1) built for MSWin32-x64-multi-thread
(with 1 registered patch, see perl -V for more detail)

Copyright 1987-2017, Larry Wall

Binary build 2601 [404865] provided by ActiveState http://www.ActiveState.com
Built Dec 11 2017 12:23:25

Perl may be copied only under the terms of either the Artistic License or the
GNU General Public License, which may be found in the Perl 5 source kit.

Complete documentation for Perl, including FAQ lists, should be found on
this system using "man perl" or "perldoc perl".  If you have access to the
Internet, point your browser at http://www.perl.org/, the Perl Home Page.
```

I would guess that Strawberry Perl does it, too.

_(Note: Python also does it, and even more: it registers a little program py.exe, which reads the shebang line and dispatches the script to the correct python version, if you have both python2 and python3 installed.)_

1dzinek:
Temat: Re: [openssl/openssl] WIP: config as a perl script (#11011)Data: 2020-02-08 8:58Nadawca: "Matthias St. Pierre" &lt;notifications@github.com&gt;Adresat: "openssl/openssl" &lt;openssl@noreply.github.com&gt;; DW: "Subscribed" &lt;subscribed@noreply.github.com&gt;; I mentioned a third script. Not all operating systems can run perl scripts the way you do it on Unix, some require that you do type perl script on the command line, or write a wrapper script in their standard command line language. command.com is exactly that for VMS. I wonder what the corresponding thing to do would be for Windows cmd.My Active Perl installed a handler for the .pl extension. You can just run it without prefixing it with the interpreter:C:\Users\msp&gt;type hello_world.pl
print("hello world");

C:\Users\msp&gt;hello_world.pl
hello world
For completeness, here's the version:C:\Users\msp\tmp&gt;perl --version

This is perl 5, version 26, subversion 1 (v5.26.1) built for MSWin32-x64-multi-thread
(with 1 registered patch, see perl -V for more detail)

Copyright 1987-2017, Larry Wall

Binary build 2601 [404865] provided by ActiveState http://www.ActiveState.com
Built Dec 11 2017 12:23:25

Perl may be copied only under the terms of either the Artistic License or the
GNU General Public License, which may be found in the Perl 5 source kit.

Complete documentation for Perl, including FAQ lists, should be found on
this system using "man perl" or "perldoc perl".  If you have access to the
Internet, point your browser at http://www.perl.org/, the Perl Home Page.
I would guess that Strawberry Perl does it, too.(Note: Python also does it, and even more: it registers a little program py.exe, which reads the shebang line and dispatches the script to the correct python version, if you have both python2 and python3 installed.)—You are receiving this because you are subscribed to this thread.Reply to this email directly, view it on GitHub, or unsubscribe.

mspncp:
The only drop of bitterness: the .pl extension is not in the list of executable extensions.
```
C:\Users\msp\tmp>hello_world
Der Befehl "hello_world" ist entweder falsch geschrieben oder
konnte nicht gefunden werden.
```
(But there is a way to add it by modifying a registry entry.)

mspncp:
For example, I can omit the .py extension:

```
C:\Users\msp\tmp>type hello_python.py
print("hello python")

C:\Users\msp\tmp>hello_python.py
hello python

C:\Users\msp\tmp>hello_python
hello python
```
IIRC correctly, this feature was not preinstalled, I hacked my registry long time ago.

mspncp:
It's even simpler: you just need to edit the PATHEXT environment variable in the system settings:

```
C:\Users\msp>set PATHEXT
PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.PY;.PYW;.PL
```

levitte:
Oh, PATHEXT! Thank you, I had no clue about that one!

levitte:
FYI, I've pulled this PR and am doing further work.

richsalz:
I hope by "pulled" you mean taking it over :)  I'll delete my local branch.  Ideally this gets closed after you've pulled down anything useful.  I don't want this on my list of things to nag the project about :)

levitte:
When I've done what I intend to, I'll submit a PR.  That's when I plan on closing this PR

levitte:
#11230 is the PR that continues this work, and includes your commit.

PR #11014, opened 2020-02-04
Fix no-tls1_3

mattcaswell:
The hostname_cb in sslapitest.c was originally only defined if TLSv1.3
was enabled. A recently added test now uses this unconditionally, so we
move the function implementation earlier in the file, and always compile
it in.


mattcaswell:
Pushed. Thanks.

mattcaswell:
Forgot to push this to 1.1.1. Now done.

PR #11015, opened 2020-02-04
Fix no-sm2

mattcaswell:
Fix ecdsatest to not run the SM2 test if SM2 has been disabled.


levitte:
I marked it "urgent", since I imagine that a run-checker build fails.

mattcaswell:
Pushed. Thanks.

PR #11019, opened 2020-02-05
Params: change UTF8 construct calls to avoid explicit strlen(3) calls.

paulidale:
It is better, safer and smaller to let the library routine handle the `strlen(3)` call.

Added a note to the documentation suggesting this.


paulidale:
Merged to master.  Thanks @t-j-h

PR #11024, opened 2020-02-06
Deprecate low level Diffie Hellman functions.

paulidale:


paulidale:
This is extremely early.  Accepting the suite of deprecated functions is the first step.

1. The command line commit is a clone of the one in #10977 and will be removed once that is merged.
1. Documentation isn't updated.
1. The conversions in `ssl3_lib.c` have problems.
1. Tests are failing.

paulidale:
Bringing this out of WIP.  There are two outstanding functions: `DH_new_by_nid` and `DH_check_params` which should be deprecated but are used by TLS with no immediately obvious replacement.

Travis failures do not appear to be relevant.

levitte:
Travis failures not relevant here

levitte:
The corresponding thing for `DH_check_params` would be the key validation stuff, right?

levitte:
`DH_new_by_nid()` is a constructor, just like `DH_new()`, even though very special case.  I suggest not to deprecate it now.

ssl/s3_lib.c
```diff
@@ -4796,7 +4796,7 @@ EVP_PKEY *ssl_generate_pkey_group(SSL *s, uint16_t id)
 /*
  * Generate parameters from a group ID
  */
-EVP_PKEY *ssl_generate_param_group(uint16_t id)
+EVP_PKEY *ssl_generate_param_group(SSL *s, uint16_t id)
```
levitte: Where are you using this?

paulidale: It isn't being used yet, but it will be necessary when `EVP_PKEY_CTX_new_id` is changed over.  See the TODO comment about 20 lines later.

I started on the conversion but soon realised that it isn't possible to complete yet.  While undoing the relevant changes, I figured this was okay to stay since it will be required.

paulidale:
Yes, `DH_check_params` would be some kind of key validation.  I tried creating an EVP_PKEY_CTX and calling some of the check functions but none worked.

Happy to leave `DH_new_by_nid` non-deprecated.  I guess NIDs will work for a while yet.  There will need to be some refactoring to avoid NIDs internally.

levitte:
> There will need to be some refactoring to avoid NIDs internally.

Yes, agreed.  I guess that we can work on that when EC stuff has transferred over to providers

ssl/s3_lib.c
levitte: Okie.  Thanks

levitte:
> Yes, `DH_check_params` would be some kind of key validation.  I tried creating an EVP_PKEY_CTX and calling some of the check functions but none worked.

That's because there currently is no validation code added in the KEYMGMT implementations for DSA and DH.

levitte:
Travis and Appveyor are a bit unhappy...

test/build.info
```diff
@@ -493,10 +485,9 @@ IF[{- !$disabled{tests} -}]
                      tls13encryptiontest wpackettest ctype_internal_test \
                      rdrand_sanitytest property_test ideatest \
                      rsa_sp800_56b_test bn_internal_test ecdsatest \
-                     rc2test rc4test rc5test hmactest ffc_internal_test \
```
levitte: It's an internal test, we should keep it around and make it able to use the deprecated DH functions "internally"

test/build.info
```diff
@@ -34,8 +34,8 @@ IF[{- !$disabled{tests} -}]
           sanitytest rsa_complex exdatatest bntest \
           ectest ecstresstest gmdifftest pbelutest \
           destest mdc2test \
-          dhtest enginetest \
-          ssltest_old exptest rsa_test \
+          enginetest \
+          exptest rsa_test \
```
levitte: You need to add them back *somewhere*

test/build.info
paulidale: Bad rebase merge conflict fix :(

paulidale:
> That's because there currently is no validation code added in the KEYMGMT implementations for DSA and DH.

As I remembered after getting stuck into it.

test/build.info
paulidale: Putting these back, fixes the CI tests.

paulidale:
Travis isn't relevant, ping for review.

paulidale:
Rebased and conflicts merged -- no code changes just infrastructure.
I'll wait for the CIs before merging.

paulidale:
Merged to master.  Thanks for the feedback and assistance.

PR #11025, opened 2020-02-06
[Pending #11074] EVP: Adapt EVP_PKEY checking, comparing and copying for provider keys, take 2

levitte:
This is an adaptation of the following functions for provided keys:

-   EVP_PKEY_copy_parameters()
-   EVP_PKEY_missing_parameters()
-   EVP_PKEY_cmp_parameters()
-   EVP_PKEY_cmp()

This replaces #10807 and is built on top of #11006

### DIFF
--- a/crypto/evp/keymgmt_lib.c
+++ b/crypto/evp/keymgmt_lib.c
@@ -269,7 +269,8 @@ static void find_keymgmt(EVP_KEYMGMT **keymgmt1, EVP_KEYMGMT **keymgmt2,
 {
     EVP_KEYMGMT *tmp_keymgmt1 = NULL, *tmp_keymgmt2 = NULL;
     void *tmp_keydata1 = NULL, *tmp_keydata2 = NULL;
-    size_t i1, i2, end = OSSL_NELEM(pk1->pkeys);
+    size_t i1, i2;
+    size_t end1 = OSSL_NELEM(pk1->pkeys), end1 = OSSL_NELEM(pk2->pkeys);
 
     /*
      * Find cache elements that share the same keymgmt, which must also have
@@ -277,13 +278,13 @@ static void find_keymgmt(EVP_KEYMGMT **keymgmt1, EVP_KEYMGMT **keymgmt2,
      * since the cache is fairly small, we assume that it's still cheaper
      * than a export + import.
      */
-    for (i1 = 0; i1 < end && pk1->pkeys[i1].keymgmt != NULL; i1++) {
+    for (i1 = 0; i1 < end1 && pk1->pkeys[i1].keymgmt != NULL; i1++) {
         if (finder(pk1->pkeys[i1].keymgmt)) {
             tmp_keymgmt1 = pk1->pkeys[i1].keymgmt;
             tmp_keydata1 = pk1->pkeys[i1].keydata;
         }
 
-        for (i2 = 0; i2 < end && pk2->pkeys[i2].keymgmt != NULL; i2++) {
+        for (i2 = 0; i2 < end2 && pk2->pkeys[i2].keymgmt != NULL; i2++) {
             if (finder(pk2->pkeys[i2].keymgmt)) {
                 tmp_keymgmt2 = pk2->pkeys[i2].keymgmt;
                 tmp_keydata2 = pk2->pkeys[i2].keydata;
@@ -404,15 +405,15 @@ int evp_keymgmt_util_match(EVP_PKEY *pk1, EVP_PKEY *pk2, int selection)
             ERR_raise(ERR_LIB_EVP, EVP_R_DIFFERENT_KEY_TYPES);
             return -1;           /* Not the same type */
         }
-    }
 
-    /*
-     * If the two keymgmt aren't the same, try to prepare any of the two
-     * EVP_PKEYs so they end up matching.
-     */
-    if (impmatch_keymgmt1 != impmatch_keymgmt2)
+        /*
+         * We found that they do have matching key types, so let's try to
+         * prepare any of the two EVP_PKEYs so they end up with matching
+         * keymgmt implementations.
+         */
         prepare_binary_op(&impmatch_keymgmt1, &impmatch_keymgmt2,
                           &keydata1, &keydata2, pk1, pk2);
+    }
 
     /* If we still don't have matching keymgmt implementations, we give up */
     if (impmatch_keymgmt1 != impmatch_keymgmt2)
--- a/crypto/evp/keymgmt_lib.c
+++ b/crypto/evp/keymgmt_lib.c
@@ -135,7 +135,11 @@ void *evp_keymgmt_util_export_to_provider(EVP_PKEY *pk, EVP_KEYMGMT *keymgmt)
     return keydata;
 }
 
-void evp_keymgmt_util_clear_pkey_cache(EVP_PKEY *pk)
+static void int_clear_pkey_cache(EVP_PKEY *pk,
+                                 void (*freefn)(EVP_KEYMGMT *keymgmt,
+                                                void *keydata,
+                                                void *freefnarg),
+                                 void *freefnarg)
 {
     size_t i;
 
@@ -148,14 +152,24 @@ void evp_keymgmt_util_clear_pkey_cache(EVP_PKEY *pk)
 
             pk->pkeys[i].keymgmt = NULL;
             pk->pkeys[i].keydata = NULL;
-            evp_keymgmt_freedata(keymgmt, keydata);
-            EVP_KEYMGMT_free(keymgmt);
+            freefn(keymgmt, keydata, freefnarg);
         }
 
         pk->cache.size = 0;
         pk->cache.bits = 0;
         pk->cache.security_bits = 0;
     }
+}
+
+static void free_keydata(EVP_KEYMGMT *keymgmt, void *keydata, void *unused)
+{
+    evp_keymgmt_freedata(keymgmt, keydata);
+    EVP_KEYMGMT_free(keymgmt);
+}
+
+void evp_keymgmt_util_clear_pkey_cache(EVP_PKEY *pk)
+{
+    int_clear_pkey_cache(pk, free_keydata, NULL);
 }
 
 void evp_keymgmt_util_cache_pkey(EVP_PKEY *pk, size_t index,
@@ -405,4 +419,66 @@ int evp_keymgmt_util_match(EVP_PKEY *pk1, EVP_PKEY *pk2, int selection)
         return -2;
 
     return evp_keymgmt_match(impmatch_keymgmt1, keydata1, keydata2, selection);
+}
+
+static int implements_copy_and_export(EVP_KEYMGMT *keymgmt)
+{
+    return keymgmt->copy != NULL
+        && (keymgmt->import == NULL || keymgmt->export != NULL);
+}
+
+static void free_keydata_for_copy(EVP_KEYMGMT *keymgmt, void *keydata,
+                                  void *orig_keydata)
+{
+    if (keydata != orig_keydata) {
+        evp_keymgmt_freedata(keymgmt, keydata);
+        EVP_KEYMGMT_free(keymgmt);
+    }
+}
+
+int evp_keymgmt_util_copy(EVP_PKEY *to, EVP_PKEY *from, int selection)
+{
+    EVP_KEYMGMT *impcopy_keymgmt_to = NULL, *impcopy_keymgmt_from = NULL;
+    void *keydata_to = NULL, *keydata_from = NULL;
+
+    /*
+     * Find suitable keymgmt.  We insist that the one we find must not
+     * only implement |copy|, but must also implement |export| if it
+     * implements |import|.  This should ensure that anything we happen to
+     * import into can be exported as well, since we're going to replace
+     * the |to| pkey cache entirely.
+     */
+    find_keymgmt(&impcopy_keymgmt_to, &impcopy_keymgmt_from,
+                 &keydata_to, &keydata_from,
+                 to, from, &implements_copy_and_export);
+
+    /*
+     * If we didn't find any fitting keymgmt for pk2 (even for importing), or
+     * the types don't match, we give up.
+     */
+    if (impcopy_keymgmt_to == NULL
+        || !match_type(impcopy_keymgmt_to, impcopy_keymgmt_from)) {
+        ERR_raise(ERR_LIB_EVP, EVP_R_DIFFERENT_KEY_TYPES);
+        return 0;
+    }
+
+    /*
+     * If the two keymgmt aren't the same, try to prepare any of the two
+     * EVP_PKEYs so they end up matching.
+     */
+    if (impcopy_keymgmt_to != impcopy_keymgmt_from)
+        prepare_binary_op(&impcopy_keymgmt_to, &impcopy_keymgmt_from,
+                          &keydata_to, &keydata_from, to, from);
+
+    /* If we still don't have matching keymgmt implementations, we give up */
+    if (impcopy_keymgmt_to != impcopy_keymgmt_from)
+        return 0;
+
+    if (!evp_keymgmt_copy(impcopy_keymgmt_to, keydata_to, keydata_from,
+                          selection))
+        return 0;
+
+    int_clear_pkey_cache(to, free_keydata_for_copy, keydata_to);
+    evp_keymgmt_util_cache_pkey(to, 0, impcopy_keymgmt_to, keydata_to);
+    return 1;
 }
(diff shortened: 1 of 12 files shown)


levitte:
The commits from #11006 are here as well, but will disappear as soon as that one is merged and this gets rebased.  This PR will remain a draft until that happens.

Feel free to roam around in the last 6 commits and comment on those here.

slontis:
I dont think I am going to review this until it is rebased.
Looking through > 50 files (36 commits) takes quite a while :)

levitte:
It is possible to look at one commit at a time... But yeah, I get it, the combo is massive

crypto/evp/p_lib.c
```diff
@@ -108,8 +108,13 @@ int EVP_PKEY_copy_parameters(EVP_PKEY *to, const EVP_PKEY *from)
 
 int EVP_PKEY_missing_parameters(const EVP_PKEY *pkey)
 {
-    if (pkey != NULL && pkey->ameth && pkey->ameth->param_missing)
-        return pkey->ameth->param_missing(pkey);
+    if (pkey != NULL) {
+        if (pkey->pkeys[0].keymgmt != NULL)
+            return !evp_keymgmt_util_has((EVP_PKEY *)pkey,
+                                         OSSL_KEYMGMT_SELECT_ALL_PARAMETERS);
+        else if (pkey->ameth->param_missing != NULL)
```
slontis: you still need the ameth== NULL check probably..

levitte:
Rebase done, now that #11006 is in

crypto/evp/p_lib.c

levitte:
Please make initial reviews.  I'm still keeping it WIP for a little bit more while investigating what happens in a merge with #10797, but I suspect I'm pretty much done...

levitte:
(at the very least, I like this take much more than #10807)

crypto/evp/keymgmt_lib.c
```diff
@@ -211,3 +225,259 @@ void *evp_keymgmt_util_fromdata(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+
+int evp_keymgmt_util_copy(EVP_PKEY *to, EVP_PKEY *from, int selection)
+{
+    EVP_KEYMGMT *impcopy_keymgmt_to = NULL, *impcopy_keymgmt_from = NULL;
+    void *keydata_to = NULL, *keydata_from = NULL;
+
+    /*
+     * Find suitable keymgmt.  We insist that the one we find must not
+     * only implement |copy|, but must implement also |export| and if it
+     * implements |import|.  This should ensure that anything we happen to
```
mattcaswell: I read this "We insist..." sentence several times and still don't understand it.

crypto/evp/p_lib.c
```diff
@@ -100,43 +107,75 @@ int EVP_PKEY_copy_parameters(EVP_PKEY *to, const EVP_PKEY *from)
...
     }
 
-    if (from->ameth && from->ameth->param_copy)
+    /* Only use the routine for provided keys if both keys are provided */
+    if (to->pkeys[0].keymgmt != NULL && from->pkeys[0].keymgmt != NULL) {
+        int rv = evp_keymgmt_util_copy(to, (EVP_PKEY *)from,
+                                       OSSL_KEYMGMT_SELECT_ALL_PARAMETERS);
+
+        /*
+         * Because the provided keydata was update beyond what legacy key
```
mattcaswell: Typo: updated

crypto/evp/p_lib.c
```diff
@@ -100,43 +107,75 @@ int EVP_PKEY_copy_parameters(EVP_PKEY *to, const EVP_PKEY *from)
...
+        return evp_keymgmt_util_match((EVP_PKEY *)a, (EVP_PKEY *)b,
+                                      OSSL_KEYMGMT_SELECT_ALL_PARAMETERS);
+
+    /* All legacy keys */
     if (a->type != b->type)
         return -1;
-    if (a->ameth && a->ameth->param_cmp)
+    if (a->ameth != NULL && a->ameth->param_cmp != NULL)
         return a->ameth->param_cmp(a, b);
     return -2;
```
mattcaswell: So are we unable to compare parameters between a legacy key and a provided key?

crypto/evp/p_lib.c
```diff
@@ -878,6 +916,14 @@ static void evp_pkey_free_it(EVP_PKEY *x)
 #endif
 }
 
+static void evp_pkey_free_it(EVP_PKEY *x)
+{
+    /* internal function; x is never NULL */
```
mattcaswell: Sounds like a dangerous assumption

crypto/evp/keymgmt_lib.c
```diff
@@ -211,3 +225,259 @@ void *evp_keymgmt_util_fromdata(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+     * TODO(3.0) investigate whether the lack of any |has| function should
+     * mean that the key has the components indicated by |selection| (i.e.
+     * they are assumed to be the empty set) or not.
+     * TODO(3.0) Investigate if we should return a value that indicates that
+     * this function is unsupported, leaving it to the caller to decide what
+     * to do.
+     *
+     * We currently assume that if there is no |has| function, the components
+     * indicated by |selection| are present, because that's how
+     * EVP_PKEY_missing_parameters() behaves for legacy EVP_PKEYs.
```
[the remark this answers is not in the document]
levitte: I originally had that idea for #11006, actually. so was basically thinking that we could have these possible return values for all of EVP_KEYMGMT:

``` C
# define EVP_KEYMGMT_FAILURE        0
# define EVP_KEYMGMT_FALSE          0
# define EVP_KEYMGMT_SUCCESS        1
# define EVP_KEYMGMT_TRUE           1
# define EVP_KEYMGMT_UNSUPPORTED   -1
```

That would leave it to the caller to decide how to handle unknown / unsupported cases.

I'm not sure if such a change belongs in *this* PR, though...

crypto/evp/keymgmt_lib.c
```diff
@@ -211,3 +225,259 @@ void *evp_keymgmt_util_fromdata(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+int evp_keymgmt_util_match(EVP_PKEY *pk1, EVP_PKEY *pk2, int selection)
+{
+    EVP_KEYMGMT *impmatch_keymgmt1 = NULL, *impmatch_keymgmt2 = NULL;
+    void *keydata1 = NULL, *keydata2 = NULL;
+
+    find_keymgmt(&impmatch_keymgmt1, &impmatch_keymgmt2, &keydata1, &keydata2,
+                 pk1, pk2, &implements_match);
+
+    /* If we found no suitable keymgmt for either key, support is missing */
+    if (impmatch_keymgmt1 == NULL || impmatch_keymgmt1 == NULL)
```
levitte: I think I spotted a bug :wink:

paulidale: Redundant, yes.  Bug probably not 😁

crypto/evp/keymgmt_lib.c
```diff
@@ -211,3 +225,259 @@ void *evp_keymgmt_util_fromdata(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+     * Find cache elements that share the same keymgmt, which must also have
+     * function determined by |finder|.  This search is obviously O(n^2), but
+     * since the cache is fairly small, we assume that it's still cheaper
+     * than a export + import.
+     */
+    for (i1 = 0; i1 < end && pk1->pkeys[i1].keymgmt != NULL; i1++) {
+        if (finder(pk1->pkeys[i1].keymgmt)) {
+            tmp_keymgmt1 = pk1->pkeys[i1].keymgmt;
+            tmp_keydata1 = pk1->pkeys[i1].keydata;
+        }
```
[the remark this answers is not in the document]
levitte: Yes.  The idea is that it may be sufficient if either of the EVP_PKEYs have a keymgmt for which `finder` returns true.
The next step will be `prepare_binary_op`, which does a cross-wise export and should end up finding the most suitable keymgmt to use.

crypto/evp/keymgmt_lib.c
levitte: I can see why...  that comment has gone through so many edits that it ended up not being English any more.  Will fix

crypto/evp/p_lib.c
levitte: Damn...

Originally, I went with the provided routines as soon as any of the keys had `ameth == NULL`...  the trouble came with a fresh new key, just created with `EVP_PKEY_new()`, being passed in, and ending up in provided space, which simply did handle this well.

So yeahok, I need to look more closely at that.  Obviously, this PR will stay WIP a little longer than I expected....

crypto/evp/p_lib.c
levitte: That has always been the assumption (scroll up a little and you'll see that this comment is nothing new)
This is an *internal **static*** function...  if the caller doesn't get it right, the fault is none other than our own, and I expect there to be a [KABOOM] in our tests if we mes up.

crypto/evp/keymgmt_lib.c
levitte: Considering the comment just above that line, I call bug, so there :wink:

levitte: ... and fixed

crypto/evp/keymgmt_lib.c
levitte: Slightly rephrased

crypto/evp/p_lib.c

crypto/evp/p_lib.c
levitte: Okie, I reworked this a little bit.  It was not as hard as I feared it would be...

levitte:
This is now pendding on #11074

levitte:
This is not WIP any more, but still pending on #11074.  Early reviews are welcome

crypto/evp/keymgmt_lib.c
```diff
@@ -211,3 +225,260 @@ void *evp_keymgmt_util_fromdata(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+
+        for (i2 = 0; i2 < end && pk2->pkeys[i2].keymgmt != NULL; i2++) {
```
slontis: I think you need a end2 = OSSL_NELEM(pk2->pkeys)..

levitte: "need" might be overstating it, or do you expect `end1` and `end2` to differ much?

I do agree that it would be more elegant, though, so will fix

crypto/evp/keymgmt_lib.c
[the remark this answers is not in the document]
mattcaswell: Probably I'm misunderstanding something here. If `finder` doesn't return true here, then when we get to the inner loop one of the following is true:
1) tmp_keymgmt1 and tmpkeydata1 are both NULL - which means the "bigbreak" condition will never be hit in that case
or
2) tmp_keymgmt1 and tmpkeydata1 have the same values as the last time we went around the outer loop. So if we didn't find a match in the inner loop last time - why go around it again?

crypto/evp/p_lib.c
```diff
@@ -84,14 +85,27 @@ int EVP_PKEY_save_parameters(EVP_PKEY *pkey, int mode)
...
-    } else if (to->type != from->type) {
-        EVPerr(EVP_F_EVP_PKEY_COPY_PARAMETERS, EVP_R_DIFFERENT_KEY_TYPES);
-        goto err;
+    /*
+     * Only check that type match this early when both keys are legacy.
+     * If either of them is purely provided, we let evp_keymgmt_util_copy()
+     * do this check, after having exported either of them that isn't
+     * provided.
+     */
+    if (!(evp_pkey_is_pure_provided(to) && evp_pkey_is_pure_provided(from))) {
```
slontis: This logic doesnt match the comment does it? Dont you want !X && !Y
You are currently testing is either key a legacy key?

crypto/evp/p_lib.c
```diff
@@ -106,43 +120,101 @@ int EVP_PKEY_copy_parameters(EVP_PKEY *to, const EVP_PKEY *from)
         return 0;
     }
 
-    if (from->ameth && from->ameth->param_copy)
+    /*
+     * Only use the routine for provided keys if either key is purely provided.
```
slontis: This comment also doesn't match the logic below (which is both keys)

crypto/evp/keymgmt_lib.c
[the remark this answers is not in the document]
slontis: @levitte ?

crypto/evp/keymgmt_lib.c
slontis: I dont know will be in each one currently - so I would not like to guess that they are the same size :)

crypto/evp/keymgmt_lib.c
```diff
@@ -211,3 +225,260 @@ void *evp_keymgmt_util_fromdata(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+            return -1;           /* Not the same type */
+        }
+    }
+
+    /*
+     * If the two keymgmt aren't the same, try to prepare any of the two
+     * EVP_PKEYs so they end up matching.
+     */
+    if (impmatch_keymgmt1 != impmatch_keymgmt2)
+        prepare_binary_op(&impmatch_keymgmt1, &impmatch_keymgmt2,
```
slontis: this can be moved up to line 407 (and get rid of the if test)

crypto/evp/p_lib.c
levitte: You're right.  I'll have another look in the morning...

crypto/evp/keymgmt_lib.c
```diff
@@ -211,3 +225,260 @@ void *evp_keymgmt_util_fromdata(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+    if (impmatch_keymgmt1 != impmatch_keymgmt2)
+        return -2;
+
+    return evp_keymgmt_match(impmatch_keymgmt1, keydata1, keydata2, selection);
+}
+
+static int implements_copy_and_export(EVP_KEYMGMT *keymgmt)
+{
+    return keymgmt->copy != NULL
+        && (keymgmt->import == NULL || keymgmt->export != NULL);
```
slontis: Are you sure you mean  keymgmt->import == NULL ?
Below it says 'only implement |copy|, but must also implement |export| if it
     * implements |import|'

crypto/evp/p_lib.c
slontis: Cheers, I have just about finished reviewing.. I dont want you being held up. Is there any other PR's you need to be reviewed?

crypto/evp/keymgmt_lib.c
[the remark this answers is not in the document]
levitte: I believe this question needs to wait for #11074, as conditions are changing with that one.  Basically, when that one lands, any `EVP_PKEY` with `pkeys[0].keymgmt == NULL` is to be considered unassigned as far as this file goes (legacy export is taken care of in `evp_pkey_make_provided()`).

crypto/evp/keymgmt_lib.c
levitte: Yeahok.  I've added the change, I'm sure the result will amuse you :wink:

crypto/evp/keymgmt_lib.c
levitte: Good point, thanks.  Done

crypto/evp/keymgmt_lib.c
levitte: > but must also implement |export| if it implements |import|

That translates into: "it's ok if |import| isn't implemented, then we don't care about checking for |export|.  However, if |import| *is* implemented, then so must |export|"

In logical terms:
1.  IF |import| == NULL THEN True
2.  OTHERWISE, IF |export| != NULL THEN True
3.  OTHERWISE False

Makes sense?

crypto/evp/p_lib.c
levitte: I need #11074 in place before tackling this one

crypto/evp/p_lib.c
levitte: I need #11074 in place before tackling this one

crypto/evp/keymgmt_lib.c
richsalz: I don't understand.  A provider can do import-only or export-only or both.  Is that concept not supported by this change?

levitte: Hmmm, good question.  I'll have a look again.

levitte:
This one is closed in favor of #11158

PR #11030, opened 2020-02-06
Fix no-dh

mattcaswell:
Don't use DH specific macros that might need to be used in a no-dh build.


iamamoose:

mattcaswell:
Pushed. Thanks.

PR #11031, opened 2020-02-06
Fix no-engine

mattcaswell:
We don't need to check if an engine has a cipher/digest in a no-engine
build.


iamamoose:

mattcaswell:
Pushed. Thanks.

PR #11034, opened 2020-02-06
Fix no-des

mattcaswell:
Don't attempt to self-test DES in the FIPS provider if we have been built
without FIPS support.


iamamoose:

mattcaswell:
Pushed. Thanks.

PR #11041, opened 2020-02-07
test/recipes/80-test_ssl_old.t: use 'openssl genpkey'

levitte:
We used 'openssl gendsa', but that one has aged in favor of 'openssl genpkey'


richsalz:
I thought Emelia's intent was that this test suite goes away once "test_ssl" was done.

levitte:
Yeah well, we haven't done that yet. Meanwhile, this script risks failure any time we don't build `openssl dsa`, which could become a reality within not too long

paulidale:
This change is already part of #10977

levitte:
> This change is already part of #10977

... and that's been merged.  Cool, then I'm killing this.

PR #11044, opened 2020-02-08
x86: Always generate .note.gnu.property section for ELF outputs

hjl-tools:
We should always generate .note.gnu.property section in x86 assembly
codes for ELF outputs to mark Intel CET support since all input files
must be marked with Intel CET support in order for linker to mark output
with Intel CET support.

Verified with

$ CC="gcc -Wl,-z,cet-report=error" ./Configure shared linux-x86 -fcf-protection
$ make
$ make test


iamamoose:

paulidale:

PR #11045, opened 2020-02-08
Make util/find-doc-nits runnable from the build tree

levitte:
Because we generate an increasing number of POD files, some of them
end up in the build tree.  This makes it difficult for find-doc-nits
to work as desired when the build tree is separate from the source
tree.

The best supported way to make it work in such an environment is to
run it from the build tree and let it use the build information from
configdata.pm to find all the POD files.  To make this smooth enough,
we add a function 'files' that returns an array of file names
corresponding to criteria from the caller.

### DIFF
--- a/util/find-doc-nits
+++ b/util/find-doc-nits
@@ -11,20 +11,25 @@ require 5.10.0;
 use warnings;
 use strict;
 
+use Carp qw(:DEFAULT cluck);
 use Pod::Checker;
 use File::Find;
 use File::Basename;
 use File::Spec::Functions;
 use Getopt::Std;
-use lib catdir(dirname($0), "perl");
+use FindBin;
+use lib "$FindBin::Bin/perl";
+
 use OpenSSL::Util::Pod;
+
+use lib '.';
+use configdata;
 
 # Set to 1 for debug output
 my $debug = 0;
 
 # Where to find openssl command
-my $BLDTOP  = $ENV{BLDTOP} || ".";
-my $openssl = "$BLDTOP/util/opensslwrap.sh";
+my $openssl = "./util/opensslwrap.sh";
 
 # Options.
 our($opt_d);
@@ -77,6 +82,7 @@ my $OUT;
 my %public;
 my $status = 0;
 
+my @sections = ( 'man1', 'man3', 'man5', 'man7' );
 my %mandatory_sections = (
     '*' => [ 'NAME', 'DESCRIPTION', 'COPYRIGHT' ],
     1   => [ 'SYNOPSIS', 'OPTIONS' ],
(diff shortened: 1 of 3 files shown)


levitte:
@richsalz, you might be interested in this

richsalz:
Yeah, looks reasonable.  It's a pity that putting .pod.in split up manpages into build/source.  I hadn't realized that.  oh well.

levitte:
Ping!

levitte:
(the only reason we haven't had a Travis failure with find-doc-nits yet is that the documentation checking job builds in the source tree)

richsalz:
I have a suggestion which adds, like, 10 lines rather than a couple hundred (and silly modules like carp:;cluck).

Make sure that the makefile gets flags for *both* the source tree and build tree.  Then, when doing the glob() for input files, do both trees and give preference for the source tree over the build tree.

richsalz:
In other words, I think this is too complicated and too generalized and not needed.  I can do a PR to show my method as POC if desired.

levitte:
Whatever you do, you will have to do the same mod I did to Configurations/Unix-Makefile.tmpl, i.e. not run find-doc-nits from the source tree.

levitte:
The benefit with the method of this PR is that the information of where things are is on *one* place.
Should we rearrange the POD directory structure again, we will only have to make changes in the corresponding `build.info`, this script will get the new info automatically without changing one line.  With the "simpler" way you propose, we must remember to update in two places.  I see that as a maintenance issue.

(see, I've been around for a bit, and we had pretty big maintenance issues of that sort, and often forgot to update some spot...  I'm trying pretty hard to avoid that)

mspncp:
typo side note: it's "runnable", not "runable".

levitte:
"ruinable" :wink:

Thanks!

richsalz:
> (see, I've been around for a bit

I will not comment use your age against you :)

richsalz:
> Whatever you do, you will have to do the same mod I did to Configurations/Unix-Makefile.tmpl, i.e. not run find-doc-nits from the source tree.

Why?  The script needs to know  where the build tree is, but why not run the source tree version?

levitte:
Let me rephrase that: do not `cd $(SRCDIR)` before running find-doc-nits, it's pointless.

The main reason is that both `$(SRCDIR)` and `$(BLDDIR)` are relative paths from the build tree (in essence, `$(BLDDIR)` is always `.`).
(yes, of course you still *can* if you want, with a bit of `pwd`, but that is quickly becoming silly)

richsalz:
Right, not cd to srcdir makes sense.  Seems to me the trivial thing to do is glob for pod.in files in srcdir and pull out the pod file from blddir.  Then glob all pod files in srcdir.

levitte:
If you're going for simplicity, why glob the .pod.in files when you can just glob .pod files in the build tree?

richsalz:
To catch mistakes, every .pod.in from the source tree should end up in the build tree.

levitte:
No, that's just incredibly silly.

levitte:
BTW, it would be nice if someone could review *this PR*.

levitte:
(commit message fixed, @mspncp)

levitte:
Travis failures not relevant here

richsalz:
> No, that's just incredibly silly.

Why on earth is it incredibly silly?  We had that problem before, until one of my PR's fixed it..

levitte:
> Why on earth is it incredibly silly? We had that problem before, until one of my PR's fixed it..

Copy over a heap of POD files just because we can't be bothered to look in two trees (or use available information from `configdat.pm`)?  Come on!

richsalz:
> Copy over a heap of POD files

I didn't say that.  Let me restate: if we find "foo.pod.in" in the source tree, then we should find "foo.pod" in the build tree.

levitte:
D'oh...  I can't read

util/find-doc-nits
```diff
@@ -85,6 +91,164 @@ my %mandatory_sections = (
...
+#
+# Example:
+#
+#       files(TAGS => 'manual');
+#       files(TAGS => [ 'manual', 'man1' ]);
+#
+# This function returns an array of files corresponding to a set of tags
+# given with the options "TAGS".  The value of this option can be a single
+# word, or an array of several words, which work as inclusive or exclusive
+# selectors.  Inclusive selectors are used to add one more files to the
```
paulidale: Do you mean "one more file" here?

levitte: Yes. Will fix

levitte: Or wait, no. Should be "one more set of files"
Will fix either way

PR #11053, opened 2020-02-10
Cleanse newly introduced but unused functions.

paulidale:
The `ossl_param_bld_to_param_ex()` function is a recent introduction that hasn't been in a release and is only called by the unit tests.

The `OSSL_PARAM_construct_from_text()` function is also a recent introduction that hasn't been in a release and isn't called from anywhere.

Given the lack of use, keeping these is akin to retaining dead code and needless complexity.


paulidale:
Merged to master.  Thanks.

PR #11054, opened 2020-02-11
Add assembly config targets for UEFI build

cjzurch:
UEFI builds are currently limited to the basic algorithms. This adds assembly builds in the UEFI environment for x86 and x86_64 build targets.

### DIFF
--- a/Configurations/10-main.conf
+++ b/Configurations/10-main.conf
@@ -1483,6 +1483,16 @@ my %targets = (
         lib_cppflags     => "-DL_ENDIAN",
         sys_id           => "UEFI",
     },
+    "UEFI-x86" => {
+        inherit_from     => [ "UEFI" ],
+        asm_arch         => 'x86',
+        perlasm_scheme   => "win32n",
+    },
+    "UEFI-x86_64" => {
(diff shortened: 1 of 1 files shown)


Configurations/10-main.conf
```diff
@@ -1483,6 +1483,16 @@ my %targets = (
...
     },
+    "UEFI-x86" => {
+        inherit_from     => [ "UEFI" ],
+        asm_arch         => 'x86',
+        perlasm_scheme   => "win32n",
+    },
+    "UEFI-x86_64" => {
+        inherit_from     => [ "UEFI" ],
+        asm_arch         => 'x86_64',
+        perlasm_scheme   => "nasm",
```
levitte: Why not "auto", like VC-WIN64A?

cjzurch: EDK2 prescribes nasm for assembly files and this lets me trigger the ($flavour =~ /nasm/) check in the perl generator files.

levitte: Okie

levitte:
Merged

1b72105076bb2e73f3c8461f9c0ca5ecefe007c8 Add assembly config targets for UEFI build

PR #11056, opened 2020-02-11
Adapt i2d_PrivateKey for provider only keys

levitte:
It uses EVP_PKEY serializers to get the desired results.

One might think that ddoing this might make things a bit dicy for
existing serializers, as they should obviously use i2d functions.
However, since our serializers use much more primitive functions
such as i2d_ASN1_INTEGER(), or keytype specific ones such as
i2d_RSAPrivateKey(), there is no clash.

### DIFF
--- a/crypto/asn1/i2d_pr.c
+++ b/crypto/asn1/i2d_pr.c
@@ -10,6 +10,8 @@
 #include <stdio.h>
 #include "internal/cryptlib.h"
 #include <openssl/evp.h>
+#include <openssl/serializer.h>
+#include <openssl/buffer.h>
 #include <openssl/x509.h>
 #include "crypto/asn1.h"
 #include "crypto/evp.h"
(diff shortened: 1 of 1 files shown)


levitte:
> Oh...wait....Travis red cross looks relevant. Otherwise looks ok.

Yup, it's waiting for #11055

levitte:
The CIs should be happier now

levitte:
And with yet another rebase, the ECX issue should be no more...

crypto/asn1/i2d_pr.c
```diff
@@ -28,6 +30,38 @@ int i2d_PrivateKey(const EVP_PKEY *a, unsigned char **pp)
...
+                    *pp += ret;
+                }
+            }
+        }
+        BIO_free(out);
+        OSSL_SERIALIZER_CTX_free(ctx);
+        if (ret >= 0)
+            return ret;
+    }
     ASN1err(ASN1_F_I2D_PRIVATEKEY, ASN1_R_UNSUPPORTED_PUBLIC_KEY_TYPE);
```
slontis: Not sure this REASON  code is very good.

levitte: Good point.  The actual problem is that the new serialization is done unconditionally.  I've changed that.

PR #11058, opened 2020-02-11
Fix builds with new HTTP client configured with no-ocsp and no-cmp

DDvO:
This is a fixup of #10667 for build issues newly detected by Travis due to #9982. 
* fix build for new HTTP client in case `OPENSSL_NO_CMP`
* fix build with new HTTP client in case `OPENSSL_NO_OCSP`
* on this occasion, fix formatting nits w.r.t. #if indentations in `ocsp.h`

### DIFF
--- a/crypto/http/http_local.h
+++ b/crypto/http/http_local.h
@@ -16,18 +16,18 @@
 /* name aliases for legacy names with name prefix "OCSP_" */
 typedef OCSP_REQ_CTX OSSL_HTTP_REQ_CTX;
 /* functions meanwhile only used internally */
-# define OSSL_HTTP_REQ_CTX_new          OCSP_REQ_CTX_new
-# define OSSL_HTTP_REQ_CTX_free         OCSP_REQ_CTX_free
-# define OSSL_HTTP_REQ_CTX_header       OCSP_REQ_CTX_http
-# define OSSL_HTTP_REQ_CTX_add1_header  OCSP_REQ_CTX_add1_header
-# define OSSL_HTTP_REQ_CTX_i2d          OCSP_REQ_CTX_i2d
-# define OSSL_HTTP_REQ_CTX_nbio         OCSP_REQ_CTX_nbio
-# ifndef OPENSSL_NO_SOCK
-#  define OSSL_HTTP_REQ_CTX_sendreq_d2i OCSP_REQ_CTX_nbio_d2i
-# endif
+#  define OSSL_HTTP_REQ_CTX_new         OCSP_REQ_CTX_new
+#  define OSSL_HTTP_REQ_CTX_free        OCSP_REQ_CTX_free
+#  define OSSL_HTTP_REQ_CTX_header      OCSP_REQ_CTX_http
+#  define OSSL_HTTP_REQ_CTX_add1_header OCSP_REQ_CTX_add1_header
+#  define OSSL_HTTP_REQ_CTX_i2d         OCSP_REQ_CTX_i2d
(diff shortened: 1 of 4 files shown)


DDvO:
Travis is happy again :smile:

DDvO:
BTW, should we add `OPENSSL_NO_HTTP`?

crypto/http/http_local.h
```diff
@@ -16,18 +16,18 @@
...
-# endif
+#  define OSSL_HTTP_REQ_CTX_new         OCSP_REQ_CTX_new
+#  define OSSL_HTTP_REQ_CTX_free        OCSP_REQ_CTX_free
+#  define OSSL_HTTP_REQ_CTX_header      OCSP_REQ_CTX_http
+#  define OSSL_HTTP_REQ_CTX_add1_header OCSP_REQ_CTX_add1_header
+#  define OSSL_HTTP_REQ_CTX_i2d         OCSP_REQ_CTX_i2d
+#  define OSSL_HTTP_REQ_CTX_nbio        OCSP_REQ_CTX_nbio
+#  ifndef OPENSSL_NO_SOCK
+#   define OSSL_HTTP_REQ_CTX_sendreq_d2i OCSP_REQ_CTX_nbio_d2i
+#  endif
```
mattcaswell: These extra indents don't seem to be needed??

mattcaswell:
> BTW, should we add OPENSSL_NO_HTTP?

I'd say no. We have far too many "no" options at the moment IMO. I don't see a need for another one.

crypto/http/http_local.h
DDvO: Right, this was a mistake: At first I thought that these #defines had to be put under `#ifndef OPENSSL_NO_OCSP` but when I removed that again I forgot to un-indent them. Fixed.

DDvO:
@mattcaswell, hope you can approve the fixes in this PR regardless of the unrelated issue with `ECX_KEY`?
Then I'd push this as quickly as possible and however fixes the latter issue can take advantage of it.

mattcaswell:
I believe this is "urgent" because it is causing all PRs to show red crosses for Travis. @DDvO do you agree? I've applied the "urgent" label.

DDvO:
> I believe this is "urgent" because it is causing all PRs to show red crosses for Travis. 
> @DDvO do you agree? I've applied the "urgent" label.

Yes, thanks, I agree. 
Does the 'ugent' label mean that I can do the push right away?

mattcaswell:
> Does the 'ugent' label mean that I can do the push right away?

In theory yes. I'd like to wait a short while to see if anyone objects, and maybe push in 2 or 3 hours if no one does.

DDvO:
> Please insert an empty line between commit title and body.

@mspncp, as you mentioned to me elsewhere GitHub unfortunately does not always keep newlines in commit messages - thanks for that hint!

In this case the first line was not really meant as a title but just the first of two items.
Neverthess I've added an empty line
and adapted the following line such that it reads as subordinate to the first one.

DDvO:
Pushed. Thanks @mattcaswell and @mspncp!

DDvO:
I wonder why for this PR GitHub apparently did not realize that it already has been merged.
Maybe because I squashed the commits during merge?
Of course, I can also close the PR manually.

mattcaswell:
> I wonder why for this PR GitHub apparently did not realize that it already has been merged.

Because the commit hashes in this PR do not match those that eventually got pushed to master (because of rebases and addrev changes).

PR #11063, opened 2020-02-12
RSA: deprecate the low level RSA functions.

paulidale:


levitte:
I believe the Travis failures aren't relevant here

apps/genrsa.c
```diff
@@ -7,8 +7,11 @@
  * https://www.openssl.org/source/license.html
  */
 
+/* We need to use the deprecated RSA low level calls */
+#define OPENSSL_SUPPRESS_DEPRECATED
+
 #include <openssl/opensslconf.h>
-#ifdef OPENSSL_NO_RSA
+#if defined(OPENSSL_NO_RSA) || defined(OPENSSL_NO_DEPRECATED_3_0)
```
levitte: Is this even necessary?

apps/rsa.c
```diff
@@ -7,8 +7,11 @@
  * https://www.openssl.org/source/license.html
  */
 
+/* We need to use the deprecated RSA low level calls */
+#define OPENSSL_SUPPRESS_DEPRECATED
+
 #include <openssl/opensslconf.h>
-#ifdef OPENSSL_NO_RSA
+#if defined(OPENSSL_NO_RSA) || defined(OPENSSL_NO_DEPRECATED_3_0)
```
levitte: Is this necessary?

apps/rsautl.c
```diff
@@ -7,8 +7,11 @@
  * https://www.openssl.org/source/license.html
  */
 
+/* We need to use the deprecated RSA low level calls */
+#define OPENSSL_SUPPRESS_DEPRECATED
+
 #include <openssl/opensslconf.h>
-#ifdef OPENSSL_NO_RSA
+#if defined(OPENSSL_NO_RSA) || defined(OPENSSL_NO_DEPRECATED_3_0)
```
levitte: Is this necessary?

apps/speed.c
```diff
@@ -146,7 +146,8 @@ static void print_message(const char *s, long num, int length, int tm);
 static void pkey_print_message(const char *str, const char *str2,
                                long num, unsigned int bits, int sec);
 static void print_result(int alg, int run_no, int count, double time_used);
-#ifndef NO_FORK
+#if !defined (NO_FORK) && !defined(OPENSSL_NO_RSA) \
+    && !defined(OPENSSL_NO_DEPRECATED_3_0)
 static int do_multi(int multi, int size_num);
 #endif
```
levitte: Er...  No.  Disabling `do_multi` is premature, as it still handles things that aren't deprecated.  Furthermore, making that a condition of `OPENSSL_NO_RSA` is wrong.

apps/speed.c
```diff
@@ -265,7 +266,8 @@ const OPTIONS speed_options[] = {
     {"mb", OPT_MB, '-',
      "Enable (tls1>=1) multi-block mode on EVP-named cipher"},
     {"mr", OPT_MR, '-', "Produce machine readable output"},
-#ifndef NO_FORK
+#if !defined (NO_FORK) && !defined(OPENSSL_NO_RSA) \
+    && !defined(OPENSSL_NO_DEPRECATED_3_0)
     {"multi", OPT_MULTI, 'p', "Run benchmarks in parallel"},
 #endif
```
levitte: No

apps/speed.c
```diff
@@ -1453,7 +1455,8 @@ int speed_main(int argc, char **argv)
     unsigned int i, k, loopargs_len = 0, async_jobs = 0;
     int keylen;
     int buflen;
-#ifndef NO_FORK
+#if !defined (NO_FORK) && !defined(OPENSSL_NO_RSA) \
+    && !defined(OPENSSL_NO_DEPRECATED_3_0)
     int multi = 0;
 #endif
```
levitte: No

apps/speed.c
```diff
@@ -1897,7 +1903,8 @@ int speed_main(int argc, char **argv)
...
     }
 
-#ifndef NO_FORK
+#if !defined (NO_FORK) && !defined(OPENSSL_NO_RSA) \
+    && !defined(OPENSSL_NO_DEPRECATED_3_0)
```
levitte: No

apps/speed.c
```diff
@@ -3797,7 +3805,8 @@ static void print_result(int alg, int run_no, int count, double time_used)
     results[alg][run_no] = ((double)count) / time_used * lengths[run_no];
 }
 
-#ifndef NO_FORK
+#if !defined (NO_FORK) && !defined(OPENSSL_NO_RSA) \
+    && !defined(OPENSSL_NO_DEPRECATED_3_0)
```
levitte: No

engines/build.info
```diff
@@ -1,4 +1,4 @@
-IF[{- !$disabled{"engine"} -}]
+IF[{- !$disabled{"engine"} && !$disabled{'deprecated-3.0'} -}]
   $PADLOCKASM=
```
levitte: How is the padlock engine related to RSA to such a degree that you disable it entirely?

levitte:
I'm utterly surprised that no test recipes needed to be changed!

apps/genrsa.c
paulidale: I get build errors otherwise.  I'm not sure why, the file shouldn't be being built but seems to be.

levitte: Okie, I'll have a look at that

engines/build.info
paulidale: It is disabling all engines, not just the padlock engine.  With the various EVP_*_meth_ functions deprecated, engines aren't much use.  I'm only hitting this here because I didn't deprecate the meth calls for digests and ciphers.

apps/speed.c
paulidale: With the RSA stuff conditioned out, it failed to compile.
I'll go through tomorrow and do something more discerning.

paulidale:
> I'm utterly surprised that no test recipes needed to be changed!

Me too :)

I had to statically link a couple of test programs but that was all.

apps/genrsa.c
levitte: Perhaps if you remove them from the big `SOURCE[openssl]` a few lines up? :wink:

levitte: I just noticed that you have forgotten that for DSA too (I failed to notice when reviewing, sorry)

levitte: Sorry, I meant the big `$OPENSSLSRC` assignment

engines/build.info
levitte: The `_meth_` functions used in e_padlock.c haven't been deprecated yet as far as I know, as padlock is cipher only.  The same goes for the other engines.  This is much too broad a stroke.

apps/genrsa.c
paulidale: Oops :flushed:

That would do it.

engines/build.info
paulidale: There is an intention to deprecate these for 3.0 isn't there?  That was my understanding.

levitte: I would prefer if we left that to the future PR that does that work

engines/e_dasync.c
```diff
@@ -15,6 +15,7 @@
  */
 #include "internal/deprecated.h"
 
+#include <openssl/opensslconf.h>
```
levitte: Change that to include `openssl/macros.h`

levitte: Or wait, hold on...  there's that inclusion of "internal/deprecated.h"...  this file shouldn't need all the guarding, so I wonder what goes wrong with it.  @mattcaswell might be able to answer that, maybe?

levitte: Ah, never mind!  libcrypto.so doesn't export those symbols, so `e_dasync.c` can try to declare itself "internal" all it wants, that still won't work.

``` diff
diff --git a/engines/e_dasync.c b/engines/e_dasync.c
index d3536cf745..66207a27b5 100644
--- a/engines/e_dasync.c
+++ b/engines/e_dasync.c
@@ -7,15 +7,6 @@
  */
 
-/*
- * SHA-1 low level APIs are deprecated for public use, but still ok for
- * internal use.  Note, that due to symbols not being exported, only the
- * #defines and strucures can be accessed, in this case SHA_CBLOCK and
- * sizeof(SHA_CTX).
- */
-#include "internal/deprecated.h"
-
-#include <openssl/opensslconf.h>
 #if defined(_WIN32)
 # include <windows.h>
 #endif
@@ -23,6 +14,7 @@
 #include <stdio.h>
 #include <string.h>
 
+#include <openssl/macros.h>
 #include <openssl/engine.h>
 #include <openssl/sha.h>
 #include <openssl/aes.h>
```

levitte: Additionally:

``` diff
diff --git a/engines/e_dasync.c b/engines/e_dasync.c
index d3536cf745..785b55d937 100644
--- a/engines/e_dasync.c
+++ b/engines/e_dasync.c
@@ -103,6 +95,8 @@ static int dasync_digest_nids(const int **nids)
 
 /* RSA */
 
+
 static int dasync_pub_enc(int flen, const unsigned char *from,
 static int dasync_pub_dec(int flen, const unsigned char *from,
@@ -119,6 +113,8 @@ static int dasync_rsa_finish(RSA *rsa);
 
 static RSA_METHOD *dasync_rsa_method = NULL;
 
+#endif
+
 /* AES */
 
 static int dasync_aes128_cbc_ctrl(EVP_CIPHER_CTX *ctx, int type, int arg,
```

paulidale: The comment details the two things that are needed from the `sha.h` header, neither requires exported symbols: one is a macro, the other the size of a structure.

engines/build.info
paulidale: Sigh...

I'll put together a PR to deprecate all of the `_meth_` functions.

engines/e_dasync.c
levitte: "internal/deprecated.h" does much more than that.  It turns off everything about deprecation (so you don't get the `OPENSSL_NO_DEPRECATED_3_0` macro, for example) and assumes that it has access to all symbols that are kept internally (such as the deprecated RSA symbols).  Unfortunately, since it links with libcrypto.so, you end up with lots of linking errors.

engines/build.info
richsalz: Yes, "disabling low-level RSA disables all engines" is not good.

engines/e_dasync.c
paulidale: Because of the `RSA_meth_` functions.

apps/speed.c
paulidale: Fixed this, the RSA results weren't guarded properly.  The rest of the `do_multi` bits are okay.

engines/build.info
paulidale: I'm only disabling the _dasync_ engine now.

apps/genrsa.c
paulidale: Fix(es) pushed.

engines/e_dasync.c
paulidale: Rather than fixing something that's going to be deprecated, I stopped building this engine.

engines/build.info
richsalz: I still think this is not acceptable.   If you can't disable the low-level RSA functions without *leaving engines intact* than don't disable in this release.

(Of course, I question the overall worth of this and how important it is given the FIPS goal but you knew I was going to raise that. :)

paulidale:
The tests aren't passing, that's why there were no changes.

engines/build.info
paulidale: A decision was recently made to disable command line applets instead of converting them to avoid the low-level calls.  Why should engines be more worthy of such conversion?

The low-level functions are only deprecated, engines can still be built using them.  If the build includes the deprecated functions, the engine will be built.  It is disabled only in a `no-deprecated` build.

levitte: Engines are essentially intact...  except for the future prediction called `no-deprecated`.

paulidale: Apart from the earlier changes that needed to be made for the cipher & digest low level deprecation....

paulidale: One of the deprecated apps didn't require any changes to work -- it wasn't using any deprecated called.  It still went.

paulidale: It's also a demonstration engine that's been disabled.

richsalz: The difference is that with no-deprecated there are alternatives.  There are no available alternatives to engines.

This is really wrong-headed.  You are spending times on things that some in the project want to do, but that aren't necessary to a release that has already slipped once and is even now still seen as "aggressive."  And yet, you won't spend time on things people want, like the QUIC PR, because it's "not strictly required for FIPS."  Well, heck, neither is this; it's admittedly a nice to have.

So instead, the project is doing things that *almost nobody will turn on* and not doing things *everyone will enable.*

I really hope this doesn't turn OpenSSL into becoming a niche TLS offering.  Seriously.

paulidale:
Converted the demonstration engine to use PKEY not RSA.

engines/build.info
t8m: No, the non-deprecated thing for engines are (will be) providers. I have to repeat - if we intend to remove the engines support at some point it is better to deprecate it ASAP to signalize the intent clearly.
The no-deprecated build should not be normally used. Clearly the no-deprecated is not to be used in general purpose builds shared among various applications such as the system openssl library in an OS distribution.

paulidale:
Out of WIP.

paulidale:
Ping for review.
Travis isn't relevant.

engines/e_dasync.c
```diff
@@ -803,3 +787,125 @@ static int dasync_aes128_cbc_hmac_sha1_cleanup(EVP_CIPHER_CTX *ctx)
      */
     return dasync_cipher_cleanup_helper(ctx, EVP_aes_128_cbc_hmac_sha1());
 }
+
+
+/*
+ * RSA implementation
+ */
```
levitte: Really, *really* nice refactoring!  I know I was pushing hard when you tried to disable all engines here, so this turnaround gets the corresponding praise.  Well done!

paulidale: It was when I switched to only this engine being disabled that I had difficulties.  It's a demo, nobody in their right mind uses it, why??? :)

Still, it is done now.

levitte: If nothing else, we gain experience. I don't think that should be underestimated

paulidale:
Merged to master with one instance of end of line white space removal.

PR #11065, opened 2020-02-12
Fix no-ec build

mattcaswell:
Don't attempt to build ecx related source files in a "no-ec" build.


mattcaswell:
I believe this is urgent because it is causing all PRs to show a red cross in travis.

Urgent label applied.

mattcaswell:
Pushed. Thanks.

PR #11067, opened 2020-02-12
Apps: don't build deprecated DH and DSA apps.

paulidale:
This also means that there doesn't need to be any conditional checks in the .c
files to avoid deprecated builds.


paulidale:
Travis failures aren't relevant.

paulidale:
Merged to master.  Thanks.

PR #11069, opened 2020-02-12
Fix build with no-ec / no-tls1_3

bernd-edlinger:
##### Checklist


mattcaswell:
The `hostname_cb` fix is the same one  as in #11014. In fact that appears to have been approved for 1.1.1, but wasn't merged there - so I can do that immediately.

bernd-edlinger:
Ah, okay, closing then.

PR #11074, opened 2020-02-12
Refactor evp_pkey_make_provided() to do legacy to provider export

levitte:
Previously, evp-keymgmt_util_export_to_provider() took care of all
kinds of exports of EVP_PKEYs to provider side keys, be it from its
legacy key or from another provider side key.  This works most of the
times, but there may be cases where the caller wants to be a bit more
in control of what sort of export happens when.

Also, when it's time to remove all legacy stuff, that job will be much
easier if we have a better separation between legacy support and
support of provided stuff, as far as we can take it.

This changes moves the support of legacy key to provider side key
export from evp-keymgmt_util_export_to_provider() to
evp_pkey_make_provided(), and makes sure the latter is called from all
EVP_PKEY functions that handle legacy stuff.

##### Checklist

### DIFF
--- a/crypto/evp/exchange.c
+++ b/crypto/evp/exchange.c
@@ -309,8 +309,12 @@ int EVP_PKEY_derive_set_peer(EVP_PKEY_CTX *ctx, EVP_PKEY *peer)
         return -2;
     }
 
-    provkey = evp_keymgmt_util_export_to_provider(peer, ctx->keymgmt);
-    /* If export failed, legacy may be able to pick it up */
+    provkey = evp_pkey_make_provided(peer, ctx->libctx, &ctx->keymgmt,
+                                     ctx->propquery);
+    /*
+     * If making the key provided wasn't possible, legacy may be able to pick
+     * it up
+     */
     if (provkey == NULL)
         goto legacy;
     return ctx->op.kex.exchange->set_peer(ctx->op.kex.exchprovctx, provkey);
@@ -319,6 +323,10 @@ int EVP_PKEY_derive_set_peer(EVP_PKEY_CTX *ctx, EVP_PKEY *peer)
 #ifdef FIPS_MODE
     return ret;
 #else
+    /*
+     * TODO(3.0) investigate the case where the operation is deemed legacy,
+     * but the given peer key is provider only.
+     */
     if (ctx->pmeth == NULL
         || !(ctx->pmeth->derive != NULL
              || ctx->pmeth->encrypt != NULL
--- a/crypto/evp/keymgmt_lib.c
+++ b/crypto/evp/keymgmt_lib.c
@@ -16,6 +16,18 @@
 #include "internal/provider.h"
 #include "evp_local.h"
 
+/*
+ * match_type() checks if two EVP_KEYMGMT are matching key types.  This
+ * function assumes that the caller has made all the necessary NULL checks.
+ */
+static int match_type(const EVP_KEYMGMT *keymgmt1, const EVP_KEYMGMT *keymgmt2)
+{
+    const OSSL_PROVIDER *prov2 = EVP_KEYMGMT_provider(keymgmt2);
+    const char *name2 = evp_first_name(prov2, EVP_KEYMGMT_number(keymgmt2));
+
+    return EVP_KEYMGMT_is_a(keymgmt1, name2);
+}
+
 struct import_data_st {
     EVP_KEYMGMT *keymgmt;
     void *keydata;
@@ -34,92 +46,72 @@ static int try_import(const OSSL_PARAM params[], void *arg)
 void *evp_keymgmt_util_export_to_provider(EVP_PKEY *pk, EVP_KEYMGMT *keymgmt)
 {
     void *keydata = NULL;
+    struct import_data_st import_data;
     size_t i, j;
 
-    /*
-     * If there is an underlying legacy key and it has changed, invalidate
-     * the cache of provider keys.
-     */
-    if (pk->pkey.ptr != NULL) {
-        /*
-         * If there is no dirty counter, this key can't be used with
-         * providers.
-         */
-        if (pk->ameth->dirty_cnt == NULL)
-            return NULL;
+    /* Export to where? */
+    if (keymgmt == NULL)
+        return NULL;
 
-        if (pk->ameth->dirty_cnt(pk) != pk->dirty_cnt_copy)
-            evp_keymgmt_util_clear_pkey_cache(pk);
-    }
+    /* If we have an unassigned key, give up */
+    if (pk->pkeys[0].keymgmt == NULL)
+        return NULL;
 
     /*
      * See if we have exported to this provider already.
      * If we have, return immediately.
      */
-    for (i = 0;
-         i < OSSL_NELEM(pk->pkeys) && pk->pkeys[i].keymgmt != NULL;
-         i++) {
-        if (keymgmt == pk->pkeys[i].keymgmt)
-            return pk->pkeys[i].keydata;
-    }
+    i = evp_keymgmt_util_find_pkey_cache_index(pk, keymgmt);
 
+    /* If we're already exported to the given keymgmt, no more to do */
+    if (keymgmt == pk->pkeys[i].keymgmt)
+        return pk->pkeys[i].keydata;
+
+    /*
+     * Make sure that the type of the keymgmt to export to matches the type
+     * of already cached keymgmt
+     */
+    if (!ossl_assert(match_type(pk->pkeys[0].keymgmt, keymgmt)))
+        return NULL;
+
+    /* Create space to import data into */
     if ((keydata = evp_keymgmt_newdata(keymgmt)) == NULL)
         return NULL;
 
-    if (pk->pkey.ptr != NULL) {
-        /* There is a legacy key, try to export that one to the provider */
+    /*
+     * We look at the already cached provider keys, and import from the
+     * first that supports it (i.e. use its export function), and export
+     * the imported data to the new provider.
+     */
+
+    /* Setup for the export callback */
+    import_data.keydata = keydata;
+    import_data.keymgmt = keymgmt;
+    import_data.selection = OSSL_KEYMGMT_SELECT_ALL;
+
+    for (j = 0; j < i && pk->pkeys[j].keymgmt != NULL; j++) {
+        EVP_KEYMGMT *exp_keymgmt = pk->pkeys[j].keymgmt;
+        void *exp_keydata = pk->pkeys[j].keydata;
 
         /*
-         * If the legacy key doesn't have an export function or the export
-         * function fails, give up
+         * TODO(3.0) consider an evp_keymgmt_export() return value that
+         * indicates that the method is unsupported.
          */
-        if (pk->ameth->export_to == NULL
-            || !pk->ameth->export_to(pk, keydata, keymgmt)) {
-            evp_keymgmt_freedata(keymgmt, keydata);
-            return NULL;
-        }
+        if (exp_keymgmt->export == NULL)
+            continue;
 
-        /* Synchronize the dirty count */
-        pk->dirty_cnt_copy = pk->ameth->dirty_cnt(pk);
-    } else {
         /*
-         * Here, there is no legacy key, so we look at the already cached
-         * provider keys, and import from the first that supports it
-         * (i.e. use its export function), and export the imported data to
-         * the new provider.
+         * The export function calls the callback (try_import), which does
+         * the import for us.  If successful, we're done.
          */
+        if (evp_keymgmt_export(exp_keymgmt, exp_keydata,
+                               OSSL_KEYMGMT_SELECT_ALL,
+                               &try_import, &import_data))
+            break;
 
-        /* Setup for the export callback */
-        struct import_data_st import_data;
-
-        import_data.keydata = keydata;
-        import_data.keymgmt = keymgmt;
-        import_data.selection = OSSL_KEYMGMT_SELECT_ALL;
-
-        for (j = 0; j < i && pk->pkeys[j].keymgmt != NULL; j++) {
-            EVP_KEYMGMT *exp_keymgmt = pk->pkeys[i].keymgmt;
-            void *exp_keydata = pk->pkeys[i].keydata;
-
-            /*
-             * TODO(3.0) consider an evp_keymgmt_export() return value that
-             * indicates that the method is unsupported.
-             */
-            if (exp_keymgmt->export == NULL)
-                continue;
-
-            /*
-             * The export function calls the callback (try_import), which
-             * does the import for us.  If successful, we're done.
-             */
-            if (evp_keymgmt_export(exp_keymgmt, exp_keydata,
-                                   OSSL_KEYMGMT_SELECT_ALL,
-                                   &try_import, &import_data))
-                break;
-
-            /* If there was an error, bail out */
-            evp_keymgmt_freedata(keymgmt, keydata);
-            return NULL;
-        }
+        /* If there was an error, bail out */
+        evp_keymgmt_freedata(keymgmt, keydata);
+        return NULL;
     }
 
     /*
@@ -137,12 +129,10 @@ void *evp_keymgmt_util_export_to_provider(EVP_PKEY *pk, EVP_KEYMGMT *keymgmt)
 
 void evp_keymgmt_util_clear_pkey_cache(EVP_PKEY *pk)
 {
-    size_t i;
+    size_t i, end = OSSL_NELEM(pk->pkeys);
 
     if (pk != NULL) {
-        for (i = 0;
-             i < OSSL_NELEM(pk->pkeys) && pk->pkeys[i].keymgmt != NULL;
-             i++) {
+        for (i = 0; i < end && pk->pkeys[i].keymgmt != NULL; i++) {
             EVP_KEYMGMT *keymgmt = pk->pkeys[i].keymgmt;
             void *keydata = pk->pkeys[i].keydata;
 
@@ -156,6 +146,19 @@ void evp_keymgmt_util_clear_pkey_cache(EVP_PKEY *pk)
         pk->cache.bits = 0;
         pk->cache.security_bits = 0;
     }
+}
+
+size_t evp_keymgmt_util_find_pkey_cache_index(EVP_PKEY *pk,
+                                              EVP_KEYMGMT *keymgmt)
+{
+    size_t i, end = OSSL_NELEM(pk->pkeys);
+
+    for (i = 0; i < end && pk->pkeys[i].keymgmt != NULL; i++) {
+        if (keymgmt == pk->pkeys[i].keymgmt)
+            break;
+    }
+
+    return i;
 }
 
 void evp_keymgmt_util_cache_pkey(EVP_PKEY *pk, size_t index,
--- a/crypto/evp/p_lib.c
+++ b/crypto/evp/p_lib.c
@@ -932,6 +932,55 @@ void *evp_pkey_make_provided(EVP_PKEY *pk, OPENSSL_CTX *libctx,
         *keymgmt = NULL;
     }
 
+#ifndef FIPS_MODE
+    /*
+     * If there is an underlying legacy key and it has changed, invalidate
+     * the cache of provider keys.
+     */
+    if (pk->pkey.ptr != NULL) {
+        EVP_KEYMGMT *legacy_keymgmt = NULL;
+
+        /*
+         * If there is no dirty counter, this key can't be used with
+         * providers.
+         */
+        if (pk->ameth->dirty_cnt == NULL)
+            goto end;
+
+        /*
+         * If no keymgmt was given by the caller, we set it to the first
+         * that's cached, to become the keymgmt to re-export to if needed,
+         * or to have a token keymgmt to return on success.  Further checks
+         * are done further down.
+         *
+         * We need to carefully save the pointer somewhere other than in
+         * tmp_keymgmt, so the EVP_KEYMGMT_up_ref() below doesn't mistakenly
+         * increment the reference counter of a keymgmt given by the caller.
+         */
+        if (tmp_keymgmt == NULL)
+            legacy_keymgmt = pk->pkeys[0].keymgmt;
+
+        /*
+         * If the dirty counter changed since last time, we make sure to
+         * hold on to the keymgmt we just got (if we got one), then clear
+         * the cache.
+         */
+        if (pk->ameth->dirty_cnt(pk) != pk->dirty_cnt_copy) {
+            if (legacy_keymgmt != NULL && !EVP_KEYMGMT_up_ref(legacy_keymgmt))
+                goto end;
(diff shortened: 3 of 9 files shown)


crypto/evp/keymgmt_lib.c
```diff
@@ -34,92 +47,66 @@ static int try_import(const OSSL_PARAM params[], void *arg)
...
-    for (i = 0;
-         i < OSSL_NELEM(pk->pkeys) && pk->pkeys[i].keymgmt != NULL;
-         i++) {
-        if (keymgmt == pk->pkeys[i].keymgmt)
-            return pk->pkeys[i].keydata;
-    }
+    i = evp_keymgmt_util_find_pkey_cache_index(pk, keymgmt);
+
+    /* If we have an unassigned key, give up */
+    if (keymgmt == NULL && i < 1)
```
slontis: why is i == 0 not ok?

levitte: Because there are no provider side key in that case.  This checks basically expresses "we looked for the NULL keymgmt item and found it at index 0", that means there's no key, i.e. an unassigned EVP_PKEY (freshly created with `EVP_PKEY_new()`

levitte:
The only Travis failure was an early timeout

crypto/evp/keymgmt_lib.c
levitte: Are you satisfied with that answer?

crypto/evp/p_lib.c
```diff
@@ -941,10 +990,61 @@ void *evp_pkey_make_provided(EVP_PKEY *pk, OPENSSL_CTX *libctx,
...
+        if (pk->pkeys[i].keymgmt != NULL) {
+            tmp_keymgmt = pk->pkeys[i].keymgmt;
+            keydata = pk->pkeys[i].keydata;
+            goto end;
+        }
+
+        /*
+         * If we still don't have a keymgmt at this point, or the legacy
+         * key doesn't have an export function, just bail out */
+        if (tmp_keymgmt == NULL || pk->ameth->export_to == NULL)
```
mattcaswell: But `tmp_keymgmt` can never be NULL here. We already checked it was non-NULL all line 993

crypto/evp/p_lib.c
```diff
@@ -941,10 +990,61 @@ void *evp_pkey_make_provided(EVP_PKEY *pk, OPENSSL_CTX *libctx,
...
+         *
+         * TODO(3.0) Right now, we assume we have ample space.  We will
+         * have to think about a cache aging scheme, though, if |i| indexes
+         * outside the array.
+         */
+        i = evp_keymgmt_util_find_pkey_cache_index(pk, tmp_keymgmt);
+        if (!ossl_assert(i < OSSL_NELEM(pk->pkeys)))
+            goto end;
+        if (pk->pkeys[i].keymgmt != NULL) {
+            tmp_keymgmt = pk->pkeys[i].keymgmt;
```
mattcaswell: Huh? Surely if `pk->pkeys[i].keymgmt != NULL` then we know `evp_keymgmt_util_find_pkey_cache_index` found a valid index, and therefore this assignment is redundant.

levitte: Not quite, it will return the index of the cache entry having the keymgmt it was given, *or* the next free entry.  I therefore need to the check so `tmp_keymgmt` doesn't get assigned NULL.

Maybe it would be clearer if I had two separate functions, one specifically designed to find a matching keymgmt, and otherwise an index past the maximum index, and another one specifically designed to find the first free entry.  They will essentially do the exact same thing, but one will tune its result a bit.

crypto/evp/p_lib.c
levitte: Good point.  This is a remnant from earlier incantations of this code

crypto/evp/p_lib.c
levitte: Maybe I should start with documenting that function :wink:

crypto/evp/p_lib.c
mattcaswell: I think the word you were looking for was incarnation :-)

levitte: Possibly yeah

crypto/evp/p_lib.c
mattcaswell: I think you misunderstood my point. As you say:

> Not quite, evp_keymgmt_util_find_pkey_cache_index will return the index of the cache entry having the keymgmt it was given, or the next free entry

The "keymgmt it was given" was `tmp_keymgmt`. If the index of the cache entry is valid then `pk->pkeys[i].keymgmt != NULL`. If that is the case then we know that `pk->pkeys[i].keymgmt` and `tmp_keymgmt` are one and the same thing. So therefore this assignment is redundant, since `tmp_keymgmt` is already equal to `pk->pkeys[i].keymgmt`

levitte: ...

Ok, you do have a point there.  I can't recall why I had this assignment...  must be one of those *incarnations*...

crypto/evp/p_lib.c

crypto/evp/p_lib.c
```diff
@@ -932,6 +932,55 @@ void *evp_pkey_make_provided(EVP_PKEY *pk, OPENSSL_CTX *libctx,
...
+
+        /*
+         * If no keymgmt was given by the caller, we set it to the first
+         * that's cached, to become the keymgmt to re-export to if needed,
+         * or to have a token keymgmt to return on success.  Further checks
+         * are done further down.
+         *
+         * We need to carefully save the pointer somewhere other than in
+         * tmp_keymgmt, so the EVP_KEYMGMT_up_ref() below doesn't mistakenly
+         * increments the reference counter of a keymgmt given by the caller.
```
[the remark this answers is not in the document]

crypto/evp/p_lib.c
levitte: Line removed

crypto/evp/p_lib.c
```diff
@@ -932,6 +932,55 @@ void *evp_pkey_make_provided(EVP_PKEY *pk, OPENSSL_CTX *libctx,
...
+         * the cache.
+         */
+        if (pk->ameth->dirty_cnt(pk) != pk->dirty_cnt_copy) {
+            if (legacy_keymgmt != NULL && !EVP_KEYMGMT_up_ref(legacy_keymgmt))
+                goto end;
+            evp_keymgmt_util_clear_pkey_cache(pk);
+        }
+
+        /*
+         * If |legacy_keymgmt| was only given a value of |tmp_keymgmt| is
```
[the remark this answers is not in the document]
levitte: Edited

levitte:
Ping!

levitte:
Travis failure not relevant

levitte:
There are two other PRs waiting for this one, so please give it focus

crypto/evp/p_lib.c
```diff
@@ -941,10 +990,60 @@ void *evp_pkey_make_provided(EVP_PKEY *pk, OPENSSL_CTX *libctx,
...
+        if (!ossl_assert(i < OSSL_NELEM(pk->pkeys)))
+            goto end;
+        if (pk->pkeys[i].keymgmt != NULL) {
+            keydata = pk->pkeys[i].keydata;
+            goto end;
+        }
+
+        /*
+         * If we still don't have a keymgmt at this point, or the legacy
+         * key doesn't have an export function, just bail out */
```
mattcaswell: Nit: comment style

crypto/evp/keymgmt_lib.c
```diff
@@ -34,92 +47,66 @@ static int try_import(const OSSL_PARAM params[], void *arg)
...
+        return NULL;
+
+    /* Create space to import data into */
     if ((keydata = evp_keymgmt_newdata(keymgmt)) == NULL)
         return NULL;
 
-    if (pk->pkey.ptr != NULL) {
-        /* There is a legacy key, try to export that one to the provider */
+    /*
+     * Here, there is no legacy key, so we look at the already cached
```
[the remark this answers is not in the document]
slontis: I think because it is now only called from evp_pkey_make_provided() after it has already done the check?

crypto/evp/p_lib.c
```diff
@@ -941,10 +990,60 @@ void *evp_pkey_make_provided(EVP_PKEY *pk, OPENSSL_CTX *libctx,
...
+            goto end;
+        }
+
+        evp_keymgmt_util_cache_pkey(pk, i, tmp_keymgmt, keydata);
+
+        /* Synchronize the dirty count */
+        pk->dirty_cnt_copy = pk->ameth->dirty_cnt(pk);
+        goto end;
+    }
+#endif
```
slontis: #endif /* FIPS_MODE */

slontis:
I mainly get this code now that I went back to look at the code before the PR.
My brain is still however having a hard time understanding why we need an array of keymgmt/keydata for a single key. Is this to import into multiple providers which all have differents keymanagers (which is a case we dont currently have for say a DH key - i.e there is only one DH keymanager ??).

crypto/evp/keymgmt_lib.c
```diff
@@ -16,6 +16,19 @@
 #include "internal/provider.h"
 #include "evp_local.h"
 
+/*
+ * match_type() checks if two EVP_KEYMGMT are matching key types.  This
+ * function assumes that the caller has made all the necessary NULL checks.
+ */
+static int match_type(EVP_KEYMGMT *keymgmt1, EVP_KEYMGMT *keymgmt2)
+{
+    /* Unconstify the provider, because evp_first_name() demands it */
```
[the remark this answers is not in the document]

crypto/evp/keymgmt_lib.c
```diff
@@ -16,6 +16,19 @@
 #include "internal/provider.h"
 #include "evp_local.h"
 
+/*
+ * match_type() checks if two EVP_KEYMGMT are matching key types.  This
+ * function assumes that the caller has made all the necessary NULL checks.
+ */
+static int match_type(EVP_KEYMGMT *keymgmt1, EVP_KEYMGMT *keymgmt2)
```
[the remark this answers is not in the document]

crypto/evp/keymgmt_lib.c
```diff
@@ -34,92 +47,66 @@ static int try_import(const OSSL_PARAM params[], void *arg)
...
+    /* If we have an unassigned key, give up */
+    if (keymgmt == NULL && i < 1)
+        return NULL;
 
+    /* If we're already exported to the given keymgmt, no more to do */
+    if (keymgmt != NULL && keymgmt == pk->pkeys[i].keymgmt)
+        return pk->pkeys[i].keydata;
+
+    /* Make sure that the keymgmt key type matches the legacy NID */
+    if (!ossl_assert(match_type(pk->pkeys[0].keymgmt, keymgmt)))
```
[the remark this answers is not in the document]
levitte: I've added early checks to see if `keymgmt` or `pk->pkeys[0].keymgmt` are NULL, in which case this function just returns NULL.

crypto/evp/keymgmt_lib.c
```diff
@@ -34,92 +47,66 @@ static int try_import(const OSSL_PARAM params[], void *arg)
...
+
+    /* If we have an unassigned key, give up */
+    if (keymgmt == NULL && i < 1)
+        return NULL;
 
+    /* If we're already exported to the given keymgmt, no more to do */
+    if (keymgmt != NULL && keymgmt == pk->pkeys[i].keymgmt)
+        return pk->pkeys[i].keydata;
+
+    /* Make sure that the keymgmt key type matches the legacy NID */
```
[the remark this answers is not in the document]
levitte: Nothing.  Comment rewritten

crypto/evp/keymgmt_lib.c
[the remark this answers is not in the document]
levitte: @slontis is correct, the function in this file don't care at all about the legacy key, that's been moved entirely to p_lib.c and other files that handle legacy stuff.  My hope is that the day we drop legacy support, this file won't have to be touched.

I've edited this comment to not mention legacy at all.

crypto/evp/p_lib.c

crypto/evp/p_lib.c

levitte:
> My brain is still however having a hard time understanding why we need an array of keymgmt/keydata for a single key. Is this to import into multiple providers which all have differents keymanagers (which is a case we dont currently have for say a DH key - i.e there is only one DH keymanager ??).

The design is that the keymgmt we use is tightly coupled with the operations that use keys (i.e. they must exists in the same provider), so they can safely share key data.  That does mean that any new provider that implements (to take your example) DH will have to write their own keymgmt for DH keys, because they will probably have their own structure to organize the DH data in (for example, if they've implemented it on top of gmp).
This means that to *flexibly* use EVP_PKEYs with diverse providers that might support different operations, we need to be able to pass the key data between their respective keymgmt, and we hold a cache of the result to avoid having to do that more than once if we can help it.

richsalz:
I do not think "flexibly sharing keys" across providers is a good thing because it could lead to security issues. For example, caching keys is a CVE waiting to happen. I prefer to see a much simpler model of explicit export/import (or perhaps those two combined into one atomic thing). 

Doing a lot of work to support flexibility, which is *not* going to be a common path I believe, is the wrong trade-off to maek.

levitte:
The cache stores pointers to provider side keys. The EVP_PKEY itself doesn't store any key and never has. The security bits stay with the providers, and it's their choice if they want to export anything or not.

The answer back to you, @richsalz, is that if we make transfers of key material between providers explicit, a scenario with multiple providers will become quite cumbersome for the applications, and will most probably hinder trying to use several providers together.

richsalz:
> The answer back to you, @richsalz, is that if we make transfers of key material between providers explicit, a scenario with multiple providers will become quite cumbersome for the applications, and will most probably hinder trying to use several providers together.

And for this release, which is not an LTS, that is the right thing to do.  Once you get experience about applications working with multiple providers and trying to use several providers at the same time, **then** figure out how to support it. As of now, there isn't even a use-case for this.

levitte:
I guess we will see.

richsalz:
> I guess we will see.

No, we won't.  Because once this gets merged it's over.  And the possible mistakes, complication, security issues, and poor fit for real needs will live forever.

OpenSSL is really bad at predicting the future and this is another instance.

I know I'm going to lose this argument, as I usually do.  Shrug.

slontis:
> No, we won't. Because once this gets merged it's over.

Err while I agree with part of what you are saying in principle, this does seem like an overreaction. As far as I can see this code is all internal, so it can be changed under the hood without affecting external users.
Multiple providers for 3.0 are needed, caching could be deferred.
Either way it can be fixed if (when) there are issues..

richsalz:
Multiple providers are needed, I agree.

Do we need to implicitly/automatically share key material across those providers?

>Either way it can be fixed 

In years maybe.  But most likely it will work good enough for most people and there's new features to be added, and it will never happen.  I base this on decades of observation and years of participation.

I'm not a project member, shrug, do what you're gonna do.

levitte:
Let's start with a very simple fact: *no keymgmt is **obligated** to export*.  That means that if `pkeys[0].keymgmt` points at a keymgmt that doesn't supply an export function, operations with a different provider are simply not supported.  In other words, sharing key data is a voluntary function, the affected keymgmt must explicitly agree to it.

On to current existing practices, in very simple terms (this is *very* condensed code from `apps/pkeyutl.c`):

``` C
    /* pkey is an EVP_PKEY passed from...  somewhere */
    EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new(pkey, NULL);
    
    if (ctx == NULL
        || !EVP_PKEY_sign_init(ctx)
        || !EVP_PKEY_sign(ctx, out, &outlen, in, inlen))
        goto error;
```

(even in a legacy scenario, there's a lot that goes on implicitly, `EVP_PKEY_CTX_new()` might find an engine that has an `EVP_PKEY_METHOD` for the that key)

We want that code to continue to work in a provided scenario, right?

What happens under the hood in a provided scenario is that `EVP_PKEY_sign()` tries an implicit fetch of an `EVP_SIGNATURE` that matches the key type of the `pkey` that was passed to `EVP_PKEY_CTX_new()`, and if there is one, there must also be a matching keymgmt.

If the `pkey` contains a legacy key, i.e. was assigned one with functions/macros like `EVP_PKEY_assign_RSA()`, or came from a PEM read, or a `d2i_PrivateKey()`, then it seems pretty straight forward, we export the legacy key data to the keymgmt (assuming it has an import function).
*This is already a situation where the key data is shared between two backends*

If the `pkey` doesn't contain a legacy key, i.e. it was created with something like `EVP_PKEY_fromdata()`, but also possibly with a PEM read that creates an `EVP_PKEY` with provided keydata rather than a legacy key (this is looming in a future where legacy support is gone).  We have a very similar situation to what I describe in the previous paragraph, and *may* end up in a situation where the provider for the `EVP_SIGNATURE` is from a different provider than the one of the keymgmt used to create the `pkey`, which leads to the same need to export key data as above.

Two situations that are really exactly the same, the only thing that differs is where the `pkey` originated.

So what I hear is that such exports should be explicit rather than implicit.  Technically, that should be no problem whatsoever.  The question is, should that be done only when the `pkey` origin is a provided key, or also when the origin is a legacy key?

So let's imagine a situation with explicit export...  as far as I'm concerned, that is technically not very hard:

``` C
    EVP_PKEY *tmp_pkey = EVP_PKEY_export_to(keymgmt, pkey);
    EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new(tmp_pkey, NULL);

    if (ctx == NULL
        || !EVP_PKEY_sign_init(ctx)
        || !EVP_PKEY_sign(ctx, out, &outlen, in, inlen))
        goto error;
```

The situations where I see this happening are these:

1.  We do not support implicit export at all, not even for a `pkey` that contains a legacy key.
    *NOTE: this also means that we do not support using provided implementations with legacy keys*
2.  We do support implicit export for a `pkey` with legacy origins, but not for a `pkey` with provided origins (i.e. doesn't contain a legacy key), and this part of the application code doesn't know what the origin is.

Here are a few situations where the application might not be able to know what kind of keydata the `pkey` contains:

1.  It might be created with `EVP_PKEY_assign_RSA()`, but just as well with `EVP_PKEY_fromdata()`
2.  It might be loaded from some source, be it file, LDAP store, PKCS#11, whatever, that results in a `pkey` with provided keydata rather than legacy.  (even a provided OSSL_STORE `file:` loader would do that, meaning that certain PEM functions would end up producing a `pkey` with provided keydata)

Therefore, I suspect that all application would (sooner rather than later) simply add that extra `EVP_PKEY_export_to()` line everywhere an `EVP_PKEY_CTX` is created, or risk spurious errors.

We can do this, technically.

Now, here's something that we have repeated over and over; well-behaved application should just need a recompile and work as they are with OpenSSL 3.0 and the provider architecture (and any provider that plays along).  This is an objective of the 3.0 effort.
The code snippet without the explicit export is well-behaved code, so requiring an `EVP_PKEY_export_to()` in most if not all situations means breaking that objective.

For you to consider...

levitte:
Meanwhile, letting this PR stall means stalling #11025 (the first take tried to work without this refactoring, and the result was a horrible mess), an indirectly #10289 (because of tests that fail without #11025).  I would rather not see that happen.

If we do decide to go the explicit export route, I suggest letting that happen in the form of *another* refactoring PR.

t-j-h:
@richsalz  asked _Do we need to implicitly/automatically share key material across those providers?_

Yes. 

And one of the many reasons is that we want to move things around relatively freely between provides without the application code having to get explicitly involved in that happening.

As @levitte mentioned - a specific provider can elect to opt of out this - and place the burden on all applications that want to use keys independent of origin.

There is a much bigger mess potential here if we force all applications to change code any time we move an algorithm between providers - and that would be **highly undesirable** outcome. 

We can debate about how to design and implement something - but the objective that applications don't have to get involved in explicit changes simply because something moves between providers is pretty fundamental.

A lot of through and discussion went into the desired future state (i.e. beyond 3.0) and @levitte is keeping that in mind as he works on the current 3.0 code because we don't want to repeat mistakes of the past where we adopted something using a very short-term viewpoint of only solving the absolute bare minimum right-in-front-of-us context.

Of course it is a balance between too far in one direction, or two far in the other direction and different people have different views. But an alarmist approach in the discussion thread on this does not help in the communication style here. 

Your points have merit - but expressing them in the manner that you have to indicate that if things aren't done the way you think then the sky will fall in - that simply undermines your intent of asking for confirmation of a direction. 

That **you** feel that **you** don't need the capability in 3.0 for **your needs** is well understood - but you also need to appreciate that there are many other viewpoints and usage models that have to be considered when making these decisions. 

I don't think you fundamentally disagree about the ultimate end-state and that this is actually just a disagreement about the timing of the steps to get there - but that is just my assumption from what you have communicated. You haven't actually said that explicitly.

richsalz:
@t-j-h my responses are generally reasonable and reasoned; see https://github.com/openssl/openssl/pull/11074#issuecomment-588335370 https://github.com/openssl/openssl/pull/11074#issuecomment-588348609  Let me restate myself.

I think automatically converting keys among providers is not needed for this release. We don't have any information for how applications would want to do that, and our only obligations are to support the default and FIPS providers.  I think making applications explicitly import/export keys is safer -- no need to worry about a cache getting mis-used -- and not spending time on this concept is a better use of project resources. The schedule has already slipped once.

mattcaswell:
> not spending time on this concept is a better use of project resources. The schedule has already slipped once.

This PR is a refactor of code that is already present. It already does the caching. Ripping it out and dealing with all the subsequent problems that would ensue would be a bigger problem for the project in terms of schedule.

richsalz:
Then leave the code as-is and don't work on this PR until, say, the end of alpha?

t8m:
Other PRs depend on this.

richsalz:
>Other PR's depend on this

I am not positive that they can't be changed, but I don't know enough of the code.

It's hard when one person is busy writing code and another gives quick approvals, to take time to consider.  Oh well, that's the nature of the software business.

levitte:
> Other PRs depend on this.

Most of all #11025.  It's predecessor, #10807, was such a mess that even I had a hard time understanding what I was doing.  This PR is meant to make a clearer separation between EVP_PKEYs with legacy keys and EVP_PKEYs with just provided keys, and makes it possible for the cleaner #11025.

crypto/evp/exchange.c
```diff
@@ -303,6 +307,10 @@ int EVP_PKEY_derive_set_peer(EVP_PKEY_CTX *ctx, EVP_PKEY *peer)
 #ifdef FIPS_MODE
     return ret;
 #else
+    /*
+     * TODO(3.0) investigate the case where the operation is deemed legacy,
+     * but the given peer key is provided only.
```
t-j-h: "provider only"?

levitte: Fixed

levitte:
I'll merge later tonight, or early tomorrow

levitte:
Merged.

*However*...  the whole discussion about the key cache had me think that what it does is mightily unclear.  As a matter of fact, it has a dual role.  More on that in an upcoming PR, which also leads to a new and much simpler take on EVP_PKEY check, copy and compare, to replace #11025.

levitte:
See #11148

PR #11075, opened 2020-02-12
Add Fips_mode_set + add libctx into config loading.

slontis:
##### Checklist


slontis:
This PR started out as adding support for FIP_mode and FIPS_set_mode, but when I went to test I found I also needed the configuration loading using a libctx..
Note that I left the OPENSSL_CTX_new() call alone and added a second call that allows you to load a config using this libctx. It is done this way so that it will be able to handle putting a set_self_test_callback() in between the new and the load..

crypto/o_fips.c
```diff
@@ -1,24 +1,49 @@
...
  * this file except in compliance with the License.  You can obtain a copy
  * in the file LICENSE in the source distribution or at
  * https://www.openssl.org/source/license.html
  */
 
+#include "crypto/evp.h"
 #include "internal/cryptlib.h"
 
+/* Check that the global properties has the value "fips=yes" */
+int FIPS_mode_with_ctx(OPENSSL_CTX *libctx)
```
mattcaswell: We already have SSL_CTX_new_with_libctx, so probably we should go with `FIPS_mode_with_libctx`

crypto/o_fips.c
```diff
@@ -1,24 +1,49 @@
...
  * Licensed under the Apache License 2.0 (the "License").  You may not use
  * this file except in compliance with the License.  You can obtain a copy
  * in the file LICENSE in the source distribution or at
  * https://www.openssl.org/source/license.html
  */
 
+#include "crypto/evp.h"
 #include "internal/cryptlib.h"
 
+/* Check that the global properties has the value "fips=yes" */
```
mattcaswell: s/has/have/ ??

crypto/o_fips.c
```diff
@@ -1,24 +1,49 @@
...
+int FIPS_mode_set_with_ctx(OPENSSL_CTX *libctx, int on)
+{
+    int ret = 1;
+
+    if (!EVP_set_default_properties(libctx, on != 0 ? "fips=yes" : ""))
+        return 0;
+
+    if (on) {
+        EVP_MD *md = EVP_MD_fetch(libctx, "SHA-256", "fips=yes");
+        ret =( md != NULL);
```
mattcaswell: swap space and (

doc/man3/FIPS_mode.pod
```diff
@@ -0,0 +1,65 @@
...
+ int FIPS_mode_with_ctx(OPENSSL_CTX *libctx);
+
+ int FIPS_mode_set(int on);
+ int FIPS_mode(void);
+
+=head1 DESCRIPTION
+
+FIPS_mode_set() enables the global property "fips=true" if I<on> is non zero,
+otherwise the global property is set to "".
+Note that if additional global properties are required then
```
mattcaswell: Here we talk about "global properties". But the EVP_set_default_properties page talks about "default properties for all
future EVP algorithm fetches, implicit as well as explicit". So "global properties" seems like a new term. I think we need to get some consistency here.

doc/man3/FIPS_mode.pod
```diff
@@ -0,0 +1,65 @@
...
+
+ int FIPS_mode_set_with_ctx(OPENSSL_CTX *libctx, int on);
+ int FIPS_mode_with_ctx(OPENSSL_CTX *libctx);
+
+ int FIPS_mode_set(int on);
+ int FIPS_mode(void);
+
+=head1 DESCRIPTION
+
+FIPS_mode_set() enables the global property "fips=true" if I<on> is non zero,
```
mattcaswell: This describes `FIPS_mode_set()` but not `FIPS_mode_set_with_ctx`

doc/man3/FIPS_mode.pod
```diff
@@ -0,0 +1,65 @@
...
+recommended to use the FIPS_mode_set_with_ctx() and FIPS_mode_with_ctx() variants.
+
+=head1 RETURN VALUES
+
+FIPS_mode_set_with_ctx() and FIPS_mode_set() return 1 if successful,
+otherwise they return 0 on failure.
+If I<on> is non zero it will also fail if a digest can not be found that has the
+"fips=true" property.
+
+FIPS_mode_with_ctx() and FIPS_mode_with_ctx() return 1 if the global property
```
mattcaswell: Duplicate `FIPS_mode_with_ctx`

test/evp_fetch_prov_test.c
```diff
@@ -73,16 +76,38 @@ static int calculate_digest(const EVP_MD *md, const char *msg, size_t len,
     return ret;
 }
 
+static int test_fip_mode(OPENSSL_CTX *ctx, int loaded_fips)
+{
+    return TEST_int_eq(FIPS_mode_with_ctx(ctx), 0)
+           && TEST_int_eq(FIPS_mode_set_with_ctx(ctx, 1), loaded_fips)
+           && TEST_int_eq(FIPS_mode_with_ctx(ctx), 1)
```
mattcaswell: If `FIPS_mode_set_with_ctx` fails, then I would expect `FIPS_mode_with_ctx` to return 0. But here it seems to return 1 regardless.

doc/man3/FIPS_mode.pod
```diff
@@ -0,0 +1,65 @@
...
+
+ int FIPS_mode_set(int on);
+ int FIPS_mode(void);
+
+=head1 DESCRIPTION
+
+FIPS_mode_set() enables the global property "fips=true" if I<on> is non zero,
+otherwise the global property is set to "".
+Note that if additional global properties are required then
+L<EVP_set_default_properties(3)> should be used instead.
```
mattcaswell: The important different between FIPS_mode_set_ex() and EVP_set_default_properties() is that the former validates that the set has worked by doing a fetch. Should something about that be stated?

crypto/context.c
```diff
@@ -145,6 +146,16 @@ OPENSSL_CTX *OPENSSL_CTX_new(void)
     return ctx;
 }
 
+#ifndef FIPS_MODE
+int OPENSSL_CTX_load_config(OPENSSL_CTX *ctx, const char *config_file)
+{
+    if (!OPENSSL_init_crypto(OPENSSL_INIT_LOAD_CONFIG, NULL))
```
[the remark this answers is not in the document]
slontis: Because there are dependancies which we want to ensure are called once before loading..
i.e: openssl_config_int() calls OPENSSL_load_builtin_modules()

doc/man3/FIPS_mode.pod
slontis: global properties was the old term.. It no longer really applies - so changed to default properties instead.

test/evp_fetch_prov_test.c
slontis: Ok changed the code to restore the old state in this case.

crypto/context.c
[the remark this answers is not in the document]
slontis: Note that the way I have changed the code - this call it will only load the default configuration if the environment variable OPENSSL_CONF is set. This means that all existing test code works how it did before without breaking (since OPENSSL_CONF is used by quite a few tests).

slontis:
Ok I had to change the query for fips to use a new function because the match function I was using was not fit for that purpose. Also changed any =true to =yes in the queries, because it silently ignores =true and treats it the same as =false.

slontis:
Last commit fixes a memory leak if ossl_method_store_set_global_properties() is called multiple times with a non NULL prop_query.

mattcaswell:
> Also changed any =true to =yes in the queries, because it silently ignores =true and treats it the same as =false.

Huh??? What?? That doesn't sound right...is that a bug?

crypto/o_fips.c
```diff
@@ -1,24 +1,62 @@
...
+ * If on != 0 it also fetches the SHA256 algorithm with the property
+ * “fips=yes” and ensure that it successfully returns.
+ */
+int FIPS_mode_set_with_libctx(OPENSSL_CTX *libctx, int on)
+{
+    int ret = 1;
+    int last_on = FIPS_mode_with_libctx(libctx);
+
+    if (!EVP_set_default_properties(libctx, on != 0 ? FIPS_QUERY_ON :
+                                                      FIPS_QUERY_OFF))
```
mattcaswell: I've just been discussing this with @levitte. It think we have a problem here. Setting the default properties like this is going to mean that we cannot load any serializers - because they aren't in the FIPS provider.

paulidale:
> Huh??? What?? That doesn't sound right...is that a bug?

We've always used _yes_ and _no_ rather than _true_ or _false_.  Aliases have been mooted but not implemented.

slontis:
> We've always used _yes_ and _no_ rather than _true_ or _false_. Aliases have been mooted but not implemented.

I think its more of an issue that it silently ignores rather than its not implemented.

crypto/o_fips.c
slontis: This sounds like a fundamental design problem to me :)
If both the fips and default are always going to be present then how do we know its fetching the fips algorithm (which was the whole point of being able to specify it in the default properties).

mattcaswell: Yes - I agree there is a design issue here. It was always the plan that we would do it the way you're doing it in this PR. We just never appreciated the impact on serializers. @levitte is considering having a separate serializer provider - but this only goes part off the way to solving the problem. Probably an item to discuss on a call.

slontis: So it should not be too hard to change this function to do a merge with other properties instead of a straight set like it is doing now.. i.e leave other default properties alone and just turn on or off the fip=yes setting.
That wont stop someone calling EVP_set_default_properties() directly though.

mattcaswell: > So it should not be too hard to change this function to do a merge with other properties instead of a straight set like it is doing now.. i.e leave other default properties alone and just turn on or off the fip=yes setting.

But that doesn't entirely help either since the serializer provider won't have "fips=yes", so the serializers will never match the query properties. You could have it do "?fips"=yes" instead. But that would mean that (for example) if you had the default provider loaded and attempted to fetch a non-fipsable algorithm, then it would work - which is undesirable.

slontis: hmm will have to think about where the serializer stuff really belongs..

paulidale: Could we register the serialisers in the default provider including "fips=yes" in their properties?  The properties are specified per algorithm.

mattcaswell: That's possible but quite confusing. If we went that route, I'd suggest we put it in a "Serialisation" provider rather than the default provider (as mentioned above).

paulidale: I'd be fine with a serialisation provider too, it would still need to define _fips=yes_ for its implementations.

paulidale:
Where should we draw the line though?

1. true, 1, affirmative, да, on, positive, ja, 3, mandatory, はい, required;
1. false, 0, nein, negative, нет, off, banned, いいえ, disallowed, vetoed.

All of these will _silently_ be treated as something other than _yes_ or _no_.  I think correctly.

crypto/o_fips.c
mattcaswell: > That's possible but quite confusing.

I think the thing that is confusing about this is that we have properties such as "default=yes", "legacy=yes" and "fips=yes", which until now have always meant "the algorithm comes from the default, legacy or fips providers". Would it be better to have a single property for this "provider=legacy", "provider=default", "provider=fips"? Then we could have "fips=yes" as a separate property meaning "the algorithm is compatible for use in FIPS mode". This would make things a lot clearer I think.

mspncp: And it scales better when providers are added in the future.

richsalz:
>  All of these will silently be treated as something other than yes or no. I think correctly.

They should be an error if a boolean predicate has an invalid test-value.

crypto/o_fips.c
mattcaswell: I've made this change in #11097

crypto/o_fips.c
```diff
@@ -1,24 +1,62 @@
...
+ * “” for on == 0. i.e (fips is unset)
+ * If on != 0 it also fetches the SHA256 algorithm with the property
+ * “fips=yes” and ensure that it successfully returns.
+ */
+int FIPS_mode_set_with_libctx(OPENSSL_CTX *libctx, int on)
+{
+    int ret = 1;
+    int last_on = FIPS_mode_with_libctx(libctx);
+
+    if (!EVP_set_default_properties(libctx, on != 0 ? FIPS_QUERY_ON :
```
paulidale: A thought.  Instead of completely replacing the global property.  Would a merge be nicer.  For turning FIPS on, merge "fips=yes" over the default properties.  For turning off, merge "-fips".  The low level `ossl_property_merge` and `ossl_property_match_count` functions know what to do.

Just a thought.

paulidale: More specifically, replace `ossl_method_store_set_global_properties' with `ossl_method_store_merge_global_properties` defined like:
```C
int ossl_method_store_merge_global_properties(OSSL_METHOD_STORE *store,
                                              const char *prop_query) {
    int ret = 0;
    OSSL_PROPERTY_LIST *prop = NULL, *global;

    if (store == NULL)
        return 1;

    ossl_property_write_lock(store);
    ossl_method_cache_flush_all(store);
    if (prop_query == NULL) {
        ossl_property_free(store->global_properties);
        store->global_properties = NULL;
        goto success;
    }
    prop = ossl_parse_query(store->ctx, prop_query);
    if (prop == NULL)
        goto end;

    if (store->global_properties == NULL) {
        store->global_properties = prop;
        prop = NULL;
        goto success;
    }
    global = ossl_property_merge(prop, store->global_properties);
    if (global == NULL)
        goto end;
    ossl_property_free(store->global_properties);
    store->global_properties = global;
 success:
    ret = 1;
 end:
    ossl_property_unlock(store);
    ossl_property_free(prop);
    return ret;
}
```

t8m: -fips would not work probably? Because the serializers have fips=yes but you do not want to turn them off.

slontis: I have added this function in - thanks Pauli

slontis:
After discussions with others it was decided that FIP_mode and FIPS_set_mode should be deprecated and no longer used as it would be a confusing interface.
 There is now a merge method that allows fips to be turned on and off in the default property settings of a libctx.

Now that #11097 is merged this should be ready for review.

t8m:
> After discussions with others it was decided that FIP_mode and FIPS_set_mode should be deprecated and no longer used as it would be a confusing interface.

I'd ask for reconsidering this at least for the FIPS_mode() call. It is very useful to applications to have a simple interface to query whether the underlying crypto library is running in FIPS_mode or not. I suppose it could be easily emulated by querying the default properties. Of course it would also require a new API call which would allow supplying the libctx.

t8m:
I have no major problem with deprecation of FIPS_mode_set(), although it means that FIPS-aware applications that explicitly set the FIPS mode with the old FIPS module will require code changes. As the API stability promise is only from 1.1 to 3.0 and not 1.0 to 3.0, this is acceptable.

richsalz:
You also need config changes to get the FIPS module loaded, right?  So a single  change pointing to the right config file should do it, no?

t8m:
I thought it would be possible to load it somehow explicitly even without config changes. But that's no problem, as I said not having FIPS_mode_set is fine.

However there is another problem with the current commits - I do not think the FIPS_mode_set function should be there at all even as deprecated function returning error. There is no point in having a call that always returns error and not having it will actually properly signalize the callers that they must change which they have to anyway.

slontis:
> I thought it would be possible to load it somehow explicitly even without config changes. But that's no problem, as I said not having FIPS_mode_set is fine.
> 
> However there is another problem with the current commits - I do not think the FIPS_mode_set function should be there at all even as deprecated function returning error. There is no point in having a call that always returns error and not having it will actually properly signalize the callers that they must change which they have to anyway.

There are 2 new functions:
int EVP_default_properties_is_fips_enabled(OPENSSL_CTX *libctx);
int EVP_default_properties_enable_fips(OPENSSL_CTX *libctx, int enable);

The old functions still currently exist in 1_1_1..
I have no problem if the old functions are deleted - but that is not my decision to make.
The old functions no longer make any sense and are confusing now that there is a library context. Having a function that just tells you about the default context may not be that useful.

slontis:
> You also need config changes to get the FIPS module loaded, right? So a single change pointing to the right config file should do it, no?

I dont really understand what you are asking.

paulidale:
> I thought it would be possible to load it somehow explicitly even without config changes.

It isn't.  The integrity checksum lives in the config file.

kaduk:
I agree with @t8m  -- `FIPS_mode_set()` should not be a symbol in the library at all.  Yes, this means applications will need to do more than recompile if they're using it, but the behavior changes between the 1.0.2 semantics and the 3.0 semantics are significant enough that attempting to provide it as a "drop-in" will do our users a disservice.  If acting just on the default libctx suffices for a given application, the change is easy enough to make.

t8m:
> > I thought it would be possible to load it somehow explicitly even without config changes.
> 
> It isn't. The integrity checksum lives in the config file.

But is the path to a default config file somehow built-into the FIPS module or not? If not then this is perhaps something we will have to handle downstream in Fedora/RHEL - I do not want users to have to meddle with config files to enable FIPS mode.

And yes, we will patch-in support for automatic FIPS mode enablement based on /proc/sys/crypto/fips_enabled but I thought that just doing the equivalent of automatic call to EVP_default_properties_enable_fips() on initialization would be sufficient.

t8m:
> There are 2 new functions:
> int EVP_default_properties_is_fips_enabled(OPENSSL_CTX *libctx);
> int EVP_default_properties_enable_fips(OPENSSL_CTX *libctx, int enable);
> 
> The old functions still currently exist in 1_1_1..
> I have no problem if the old functions are deleted - but that is not my decision to make.

Then we will have to call a vote I suppose.

> The old functions no longer make any sense and are confusing now that there is a library context. Having a function that just tells you about the default context may not be that useful.

Why it would not be useful? Practically 99% of applications will not care about anything but the default context.

mattcaswell:
> But is the path to a default config file somehow built-into the FIPS module or not?

Well, not built into the FIPS module, but built into libcrypto. The config file should be automatically loaded without the end application having to do anything.

kaduk:
> Why it would not be useful? Practically 99% of applications will not care about anything but the default context.

"not useful" is perhaps too absolute.  A FIPS_mode() call that queries only the default library context is less useful than the FIPS_mode() API from 1.0.2, since the 1.0.2 semantics provide a solid compliance story.  If libcrypto reports that you're in FIPS mode, all calls to openssl are using the validated crypto, full stop.  With the default library context version, if you want to convince auditors that only the validated crypto is being used, then you now have to audit all your third-party dependencies to ensure that they are not creating additional library contexts (which would be using non-validated crypto).  The concern in this case is less about the local application than about the dependencies.

slontis:
This PR has now been simplified to only contain the contentious changes for the fips_mode_set and fips_mode API's.

Other code has been moved to #11239.

t8m:
> "not useful" is perhaps too absolute. A FIPS_mode() call that queries only the default library context is less useful than the FIPS_mode() API from 1.0.2, since the 1.0.2 semantics provide a solid compliance story. If libcrypto reports that you're in FIPS mode, all calls to openssl are using the validated crypto, full stop. With the default library context version, if you want to convince auditors that only the validated crypto is being used, then you now have to audit all your third-party dependencies to ensure that they are not creating additional library contexts (which would be using non-validated crypto). The concern in this case is less about the local application than about the dependencies.

I do not really see this as a problem. This is a matter of the FIPS module security policy.

kaduk:
> > "not useful" is perhaps too absolute. A FIPS_mode() call that queries only the default library context is less useful than the FIPS_mode() API from 1.0.2, since the 1.0.2 semantics provide a solid compliance story. If libcrypto reports that you're in FIPS mode, all calls to openssl are using the validated crypto, full stop. With the default library context version, if you want to convince auditors that only the validated crypto is being used, then you now have to audit all your third-party dependencies to ensure that they are not creating additional library contexts (which would be using non-validated crypto). The concern in this case is less about the local application than about the dependencies.
> 
> I do not really see this as a problem. This is a matter of the FIPS module security policy.

I'm not sure I understand your point properly, as I see it as a matter of the application's security policy (e.g., if the application's security policy is that all crypto must be FIPS).  With the 1.0.2 semantics this application policy is easy to enforce, by doing coarse-grained auditing that openssl is the only crypto, and then having libcrypto enforce the fips status.  This cannot be done within the fips module with the 3.0.0 semantics, since the dispatch to providers is done outside the fips provider.  In order to get the same enforcement of the application policy, now there must be a more specific audit that looks for use of non-default library contexts.

richsalz:
You can have a config that loads *only* the FIPS module and not the default.  Using things with a separate context or go outside your policy and load the default module, seem out of scope.

t8m:
I mean, the FIPS_mode() call has multiple uses and only one of them was something akin to ensurance that the application is using only FIPS validated crypto. BTW I do not think you could do away without code audits anyway if you wanted to be completely sure.

The other uses such as change of application behavior depending on the FIPS mode status are still valid.

t8m:
@openssl/otc @openssl/omc So I see two potential votes needed here:
1. Vote about full removal of FIPS_mode_set instead of crippling it.
2. Vote about keeping at least some backwards compatible behavior for FIPS_mode instead of crippling it.
And if 2 does not pass and 1 does I'll immediately propose removing FIPS_mode completely as well.
Now the question is whether these votes are for OTC or OMC.
As I am OTC member I can only propose the votest to OTC however if those decisions should be done by OMC (and at least 1 seems to me to be rather an OMC topic) it does not make sense to propose them to OTC.

t8m:
The vote for keeping the backwards compatible behavior of FIPS_mode did not pass. I would propose a vote to drop the functions completely because the crippled behavior of always returning an error does not make any sense. However it is rather an OMC vote and also I do not really care because I will be patching the FIPS_mode downstream anyway.

paulidale:
I've initiated an OMC discussion about removing these.

paulidale:
OMC vote is in progress to decide between this and #11680 so putting a hold on both.

paulidale:
The OMC vote decided in favour of removal rather than deprecation.
This PR will be closed.

PR #11078, opened 2020-02-12
Adapt i2d_PUBKEY for provider only keys

levitte:
It uses EVP_PKEY serializers to get the desired results.

### DIFF
--- a/crypto/x509/x_pubkey.c
+++ b/crypto/x509/x_pubkey.c
@@ -253,21 +253,52 @@ EVP_PKEY *d2i_PUBKEY(EVP_PKEY **a, const unsigned char **pp, long length)
 
 int i2d_PUBKEY(const EVP_PKEY *a, unsigned char **pp)
 {
-    X509_PUBKEY *xpk = NULL;
     int ret = -1;
 
     if (a == NULL)
         return 0;
-    if ((xpk = X509_PUBKEY_new()) == NULL)
-        return -1;
-    if (a->ameth != NULL && a->ameth->pub_encode != NULL
-        && !a->ameth->pub_encode(xpk, a))
-        goto error;
-    xpk->pkey = (EVP_PKEY *)a;
-    ret = i2d_X509_PUBKEY(xpk, pp);
-    xpk->pkey = NULL;
- error:
-    X509_PUBKEY_free(xpk);
+    if (a->ameth != NULL) {
+        X509_PUBKEY *xpk = NULL;
+
+        if ((xpk = X509_PUBKEY_new()) == NULL)
+            return -1;
+
+        /* pub_encode() only encode parameters, not the key itself */
+        if (a->ameth->pub_encode != NULL && a->ameth->pub_encode(xpk, a)) {
+            xpk->pkey = (EVP_PKEY *)a;
+            ret = i2d_X509_PUBKEY(xpk, pp);
+            xpk->pkey = NULL;
+        }
+        X509_PUBKEY_free(xpk);
+    } else if (a->pkeys[0].keymgmt != NULL) {
+        const char *serprop = OSSL_SERIALIZER_PUBKEY_TO_DER_PQ;
+        OSSL_SERIALIZER_CTX *ctx =
+            OSSL_SERIALIZER_CTX_new_by_EVP_PKEY(a, serprop);
+        BIO *out = BIO_new(BIO_s_mem());
+        BUF_MEM *buf = NULL;
+
+        if (ctx != NULL
+            && out != NULL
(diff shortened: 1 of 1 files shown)


crypto/x509/x_pubkey.c
```diff
@@ -263,9 +263,43 @@ int i2d_PUBKEY(const EVP_PKEY *a, unsigned char **pp)
...
+            OSSL_SERIALIZER_CTX_new_by_EVP_PKEY(a, serprop);
+        BIO *out = BIO_new(BIO_s_mem());
+
+        if (ctx != NULL
+            && out != NULL
+            && OSSL_SERIALIZER_CTX_get_serializer(ctx) != NULL
+            && OSSL_SERIALIZER_to_bio(ctx, out)) {
+            BUF_MEM *buf = NULL;
+
+            BIO_get_mem_ptr(out, &buf);
```
slontis: It looks like this wont return 0 - but maybe it should be doing the check anyway..

crypto/x509/x_pubkey.c
```diff
@@ -263,9 +263,43 @@ int i2d_PUBKEY(const EVP_PKEY *a, unsigned char **pp)
...
+            ret = buf->length;
+
+            if (pp != NULL) {
```
slontis: what is the point of calling this method if pp is NULL.. That looks like an error case to me.

crypto/x509/x_pubkey.c
```diff
@@ -263,9 +263,43 @@ int i2d_PUBKEY(const EVP_PKEY *a, unsigned char **pp)
...
+                if (*pp == NULL) {
```
slontis: This code looks very strange..
In the NULL case *pp points to the start of the buffer
In the other case it increments the buffer by ret.
Not sure how this works unless the calling code is doing equivalent code for the NULL case??

crypto/x509/x_pubkey.c
```diff
@@ -263,9 +263,43 @@ int i2d_PUBKEY(const EVP_PKEY *a, unsigned char **pp)
     if (a->ameth != NULL && a->ameth->pub_encode != NULL
```
slontis: maybe collapse this code into the new check below..

slontis: The xpk should also be inside the a->ameth != NULL block

crypto/x509/x_pubkey.c
```diff
@@ -263,9 +263,43 @@ int i2d_PUBKEY(const EVP_PKEY *a, unsigned char **pp)
...
+                    buf->data = NULL;
+                } else {
+                    memcpy(*pp, buf->data, ret);
+                    *pp += ret;
+                }
+            }
+        }
+        BIO_free(out);
+        OSSL_SERIALIZER_CTX_free(ctx);
+        if (ret >= 0)
```
slontis: this can just return if xpk = new is put in the right place..

crypto/x509/x_pubkey.c
levitte: Nope. And i2d function is still supposed to return the length of the would be encoding.

crypto/x509/x_pubkey.c
levitte: From the manual:

> **i2d_TYPE**() encodes the structure pointed to by _\*a_ into DER format. If _ppout_ is not NULL, it writes the DER encoded data to the buffer at _\*ppout_, and increments it to point after the data just written. If the return value is negative an error occurred, otherwise it returns the length of the encoded data.
> 
> If _\*ppout_ is NULL memory will be allocated for a buffer and the encoded data written to it. In this case _\*ppout_ is not incremented and it points to the start of the data just written.

levitte: You really should read the manual.  `perldoc -o man doc/man3/d2i_X509.pod`.  It has NOTES and WARNINGS sections that say quite a bit

slontis: yep - I know why its done (Do I think its good - not really :))
Thanks for the info : d2i_X509.pod is not the first place I would think of to look.

levitte:
BTW, @slontis...  if you're looking at this, you should also look at #11056.  These two PRs are virtually copies of each other.

crypto/x509/x_pubkey.c
levitte: Yeah, I just realised that

crypto/x509/x_pubkey.c
slontis: answer this one and we are done..

levitte: Done, I hope it's satisfactory

PR #11080, opened 2020-02-12
TEST: Create test specific output directories

levitte:
We had all tests run with test/test-runs/ as working directory, and
tests cleaned up after themselves...  which is well and good, until
you want to have a look at what went wrong when a complex test fails,
and you have to recreate everything it does manually.

To remedy this, we have OpenSSL::Test create the result directory
dynamically (and cleaning it up first if it's already there) and let
the test recipe have that as working directory.

Test recipes are now encouraged to name their diverse output files
uniquely, and not to clean them up, to allow a developer to have a
look at the files that were produced.

With continuous integration that allows this, the testrun directories
could also be archived and be left as a build artifact.


levitte:
I'm slowly modifying one test recipe after the other, hence WIP.

paulidale:
Rather than naming each file differently, could each test have its own subdirectory instead?
It would make it easier to locate all of the output from a failing test.

levitte:
Each test recipe gets its own subdirectory, but each test recipe may produce more than one file.

slontis:
travis failure looks valid..
15-test_mp_rsa.t                 (Wstat: 65280 Tests: 0 Failed: 0)
Non-zero exit status: 255
Parse errors: No plan found in TAP output

levitte:
So, the way I've currently implemented this is to let the test framework take the name given with `setup()` and create a directory `"testrun-$name"` in the test directory.

I'm not sure I like that too much, there's already a lot of stuff in the test directory, so I'm thinking that it might be better to create a separate directory tree, $BLDDIR/testruns, and then just create subdirectories in it that are exactly the same as the names given to `setup()`.
So for example, the test_mp_rsa test, which has `setup("test_mp_rsa");`, would end up writing its files in $BLDDIR/testruns/test_mp_rsa rather than what this PR currently does, $BLDDIR/test/testrun-test_mp_rsa.

Thoughts?

levitte:
> travis failure looks valid..

Fixedd

slontis:
I think not polluting the test folder is a good idea.. So I prefer the testruns folder being used as a base.

levitte:
> I think not _polluting_ the test folder is a good idea

That was the key word I was looking for.  Thank you.

levitte:
Change done.  There is now new directory created in the build tree top, `test-runs/`, under which there are subdirectories for each test recipe, do for example `test-runs/test_mp_rsa/` for 15-test_mp_rsa.t

levitte:
I've also updated CHANGES

paulidale:
Does `.gitignore` need an update too?

levitte:
> Does `.gitignore` need an update too?

Yup.  Done

levitte:
I've searched for all `unlink` calls and fixed those that needed fixing.  A few remain, but they are about SSL session files and inconsequential temporary files.

I have not checked every test recipe thoroughly to see if they overwrite their own output files in the same script, and do not intend to do so in this PR.  That's left as an exercise to anyone who has an interest in a particular recipe's output.

levitte:
Time for a reiew

levitte:
Argh.  The CMS tests were a bit more complex than I noticed.  Back to WIP

levitte:
I believe I finally got test_cms to bend to my will

levitte:
Hopefully, I'd fixed them now

levitte:
Yup, CIs are happy

romen:
Not at all to prevent this to be merged before #10631 ...

No, not really, as this is an idea for another PR that came to me while reviewing this: would it be possible to have a way to run the tests (or maybe the groups of tests) run in parallel rather than as a single sequential job?

I gather the main issue would be about contention of the console for warnings, errors and results, but it seems to me we already have in place some buffering capability in place to support the _verbose on failure_ option, so maybe we could devise something to solve the console contention in general and make sure that results are printed atomically?

What are your thoughts about it? Is it even worth considering?

levitte:
@romen, please create an issue.  There are mechanisms in Test::More to allow exactly what you're talking about.

PR #11081, opened 2020-02-12
dsatest: fix the non-deprecated version of the test run.

paulidale:
This test was broken and due to another error it always took the deprecated path and the break didn't reveal itself.


paulidale:
Merged to master, thanks.

PR #11082, opened 2020-02-13
EVP: deprecate the EVP_X_meth_ functions.

paulidale:


paulidale:
Very early WIP.

levitte:
Very early, you couldn't have said it better. I would suggest that this isn't merged before we have all provider features in place, including OSSL_STORE support. Why? Because this PR implicitly deprecates everything ENGINE.

paulidale:
It would be more motivational to merge this early :)

levitte:
Yeah, like we're not feeling the pressure already....

beldmit:
Being an engine developer, I think deprecating all this stuff in 3.0 is premature. Yes, I understand the difference between removing and deprecation :)

paulidale:
@levitte, any idea why only the ARM64 build is complaining about `doc-nits`?

t8m:
There is missing GENERATE="yes" in the failing ARM64 build, which means the openssl.pod is not updated from the .in files.

mattcaswell:
> Being an engine developer, I think deprecating all this stuff in 3.0 is premature. Yes, I understand the difference between removing and deprecation :)

I understand your point of view, but I disagree. The contortions we have had to put the code through to be able to continue to support engines and meths is awful. This is not long term sustainable. We're already going to have to live with it for significantly longer than I'd like due to our stability policy. We need to do everything we can to encourage people to move at the earliest opportunity.

paulidale:
Thanks @t8m.

paulidale:
Fix in #11607.

build.info
```diff
@@ -1,6 +1,10 @@
 # Note that some of these directories are filtered in Configure.  Look for
 # %skipdir there for further explanations.
-SUBDIRS=crypto ssl apps test util tools fuzz engines providers doc
+$DEPRECATED_SUBDIRS=
+IF[{- !$disabled{'deprecated-3.0'} -}]
+  $DEPRECATED_SUBDIRS=engines
+ENDIF
+SUBDIRS=crypto ssl apps test util tools fuzz providers doc $DEPRECATED_SUBDIRS
```
levitte, line 7: You forget that these things are cumulative, so you could simply to it like this:

```
SUBDIRS=crypto ssl apps test util tools fuzz providers doc
  SUBDIRS=engines
ENDIF
```

crypto/cms/cms_kari.c
```diff
@@ -429,6 +435,7 @@ static int cms_wrap_init(CMS_KeyAgreeRecipientInfo *kari,
         return 0;
     keylen = EVP_CIPHER_key_length(cipher);
     if ((EVP_CIPHER_flags(cipher) & EVP_CIPH_FLAG_GET_WRAP_CIPHER) != 0) {
+        /* TODO: make this not get a method we can call directly */
         ret = EVP_CIPHER_meth_get_ctrl(cipher)(NULL, EVP_CTRL_GET_WRAP_CIPHER,
                                                0, &kekcipher);
```
levitte, line 440: EVP_CTRL_GET_WRAP_CIPHER is new code in master, it seems...

test/build.info
```diff
@@ -27,14 +27,18 @@ IF[{- !$disabled{tests} -}]
   INCLUDE[libtestutil.a]=../include ../apps/include ..
   DEPEND[libtestutil.a]=../libcrypto
 
-  PROGRAMS{noinst}=\
+  IF[{- !$disabled{'deprecated-3.0'} -}]
+    $DEPRECATED_TESTS=enginetest
+  ENDIF
```
levitte, line 32:
```
    PROGRAMS{noinst}=enginetest
  ENDIF
```

levitte:
Apart from a couple of nits, this looks fine

build.info
paulidale: Okay, done.

crypto/cms/cms_kari.c
paulidale: More bizarre is that *we* don't appear to use it anywhere.

test/build.info
paulidale: Done.

paulidale:
Travis is relevant: `doc/man1/openssl-list.pod: list does not implement -public-key-methods`
It needs conditioning.

crypto/cms/cms_kari.c
paulidale: @beldmit, it looks like you added this but I can't find a reference to it either in OpenSSL or the GOST engine??

beldmit: Well... It seems to be a remnant of unsuccessful trying to add support KeyAgreement on creating the GOST CMS. As no way to pass the choice between KA and KT was found, I tried to clean up all the remnants but now I see that some still persist.

levitte: This was a side note, BTW.  The TODO comment is sufficient for now, we can revisit this in another PR

paulidale:
Travis is a timeout.

paulidale:
Merged to master.
Thanks for the patience.

PR #11087, opened 2020-02-13
include/openssl/whrlpool.h: correct unbalanced deprecation guards

levitte:


levitte:
This caused a build error for me when configured `enable-buildtest-c++` and `no-deprecated`

PR #11088, opened 2020-02-13
Fix VMS build [1.1.1 only]

levitte:
The restructuring of inclusion directories meant the descrip.mms template needed an adjustment.
Furthermore, the VMS entropy gatherer had a couple of small issues.


levitte:
My VMS builds with this patch applied went through smoothly, test 'n all

PR #11092, opened 2020-02-13
Enhance diagnostic messages of find-doc-nits

DDvO:
It happened to me more than once that `make doc-nits` terminated like this:
```
Malformed line 
```
and it took me a while to find out that this is because of a trailing empty line in `util/libcrypto.num` (which can easily produced by mistake when manually fixing a merge conflict in that file).

This PR changes the error output to something like
```
Malformed line 4931 in util/libcrypto.num: 
```

On this occasion, this PR also makes sure (In a separate commit) that error message strings consistently start with an uppercase letter.

### DIFF
--- a/util/find-doc-nits
+++ b/util/find-doc-nits
@@ -577,7 +577,7 @@ sub parsenum {
         next if /^#/;
         next if /\bNOEXIST\b/;
         my @fields = split();
-        die "Malformed line $_"
+        die "Malformed line $. in $file: $_"
             if scalar @fields != 2 && scalar @fields != 4;
         push @apis, $fields[0];
     }
--- a/util/find-doc-nits
+++ b/util/find-doc-nits
@@ -102,13 +102,13 @@ sub name_synopsis {
     return unless $contents =~ /=head1 NAME(.*)=head1 SYNOPSIS/ms;
     my $tmp = $1;
     $tmp =~ tr/\n/ /;
-    err($id, "trailing comma before - in NAME")
+    err($id, "Trailing comma before - in NAME")
         if $tmp =~ /, *-/;
     $tmp =~ s/ -.*//g;
     err($id, "POD markup among the names in NAME")
         if $tmp =~ /[<>]/;
     $tmp =~ s/  */ /g;
-    err($id, "missing comma in NAME")
+    err($id, "Missing comma in NAME")
         if $tmp =~ /[^,] /;
 
     my $dirname = dirname($filename);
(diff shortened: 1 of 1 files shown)


levitte:
It seems you haven't registered your personal address with your individual CLA...

DDvO:
@mattcaswell, unfortunately it seems my alternative email address is still causing issues with the CLA(s) on file. How if I simply sign a further ICLA containing that address?

DDvO:
Or can we persuade the CLA checker that via my PGP signature using my Siemens address on the new key associated with my alternative address carries over the ICLA as well :wink: ?

mattcaswell:
>  How if I simply sign a further ICLA containing that address?

That would probably make things easier (make sure to list both your addresses in the CLA).

DDvO:
Closing and re-opening this little PR to kick the CLA check.

DDvO:
Anyone volunteering to review these simple changes?

BTW, Travis CI is not really relevant here (and failed for the usual reason).

richsalz:
find-doc-nits needs its own nit-checker :)  consistency is good.  this should get approved.

util/find-doc-nits
```diff
@@ -533,7 +533,7 @@ sub check {
             my $count = 0;
             foreach my $line ( split /\n+/, $1 ) {
                 if ( $line =~ m@include <openssl/@ ) {
-                    err($id, "has multiple includes")
+                    err($id, "Has multiple includes")
```
mspncp: This sentence here seems to be lacking its subject. (cc @richsalz)

mspncp: Oh, I understand: The subject is "Malformed line ..."

richsalz:
"foo has multiple includes" is a valid sentence, no?

mspncp:
> "foo has multiple includes" is a valid sentence, no?

Sure. It was the 'foo' that I was missing, but meanwhile I found it. :-)

mspncp:
> > "foo has multiple includes" is a valid sentence, no?
> 
> Sure. It was the 'foo' that I was missing, but meanwhile I found it. :-)

Nevertheless, since "Has..." starts a new sentence, It looks gramatically wrong, one would expect something like "It has...". Also, it is not the line which has multiple includes, it's the pod file. So the message is a bit confusing IMHO.

How about just changing the message to the following?

> File has multiple include declarations

richsalz:
I think we have spent more time discussion an error message that will never escape past a PR :)

DDvO:
Thanks also to @richsalz and @mspncp for your comments.
Since I agree it's not worth discussing/fixing uppercase vs. lowercase in the error messages of this tool and we have already an OTC approval by @levitte I've just merged this without further ado.

DDvO:
Oops, sorry @mspncp I overlooked that there was a second approval by you.

mspncp:
Never mind. Things happen.

PR #11097, opened 2020-02-14
Introduce the provider property

mattcaswell:
Replace the properties default, fips and legacy with a single property
called "provider". So, for example, instead of writing "default=yes" to
get algorithms from the default provider you would instead write
"provider=default". We also have a new "fips" property to indicate that
an algorithm is compatible with FIPS mode. This applies to all the
algorithms in the FIPS provider, as well as any non-cryptographic
algorithms (currently only serializers).

### DIFF
--- a/crypto/property/property_parse.c
+++ b/crypto/property/property_parse.c
@@ -562,12 +562,9 @@ OSSL_PROPERTY_LIST *ossl_property_merge(const OSSL_PROPERTY_LIST *a,
 int ossl_property_parse_init(OPENSSL_CTX *ctx)
 {
     static const char *const predefined_names[] = {
-        "default",      /* Being provided by the default built-in provider */
-        "legacy",       /* Provided by the legacy provider */
-        "provider",     /* Name of provider (default, fips) */
+        "provider",     /* Name of provider (default, legacy, fips) */
         "version",      /* Version number of this provider */
-        "fips",         /* FIPS supporting provider */
-        "engine",       /* An old style engine masquerading as a provider */
+        "fips",         /* FIPS compatible algorithm */
         "format",       /* output format for serializers */
         "type",         /* output type for serializers */
     };
@@ -579,7 +576,11 @@ int ossl_property_parse_init(OPENSSL_CTX *ctx)
 
     /* Pre-populate the two Boolean values */
     if ((ossl_property_true = ossl_property_value(ctx, "yes", 1)) == 0
-        || (ossl_property_false = ossl_property_value(ctx, "no", 1)) == 0)
+        || (ossl_property_false = ossl_property_value(ctx, "no", 1)) == 0
+        /* Create values for the built-in providers */
+        || ossl_property_value(ctx, "default", 1) == 0
+        || ossl_property_value(ctx, "legacy", 1) == 0
+        || ossl_property_value(ctx, "fips", 1) == 0)
         goto err;
 
     return 1;
--- a/crypto/property/property_parse.c
+++ b/crypto/property/property_parse.c
@@ -576,11 +576,7 @@ int ossl_property_parse_init(OPENSSL_CTX *ctx)
 
     /* Pre-populate the two Boolean values */
     if ((ossl_property_true = ossl_property_value(ctx, "yes", 1)) == 0
-        || (ossl_property_false = ossl_property_value(ctx, "no", 1)) == 0
-        /* Create values for the built-in providers */
-        || ossl_property_value(ctx, "default", 1) == 0
-        || ossl_property_value(ctx, "legacy", 1) == 0
-        || ossl_property_value(ctx, "fips", 1) == 0)
+        || (ossl_property_false = ossl_property_value(ctx, "no", 1)) == 0)
         goto err;
 
     return 1;
--- a/crypto/property/property_parse.c
+++ b/crypto/property/property_parse.c
@@ -564,7 +564,7 @@ int ossl_property_parse_init(OPENSSL_CTX *ctx)
     static const char *const predefined_names[] = {
         "provider",     /* Name of provider (default, legacy, fips) */
         "version",      /* Version number of this provider */
-        "fips",         /* FIPS compatible algorithm */
+        "fips",         /* FIPS validated or FIPS supporting algorithm */
         "format",       /* output format for serializers */
         "type",         /* output type for serializers */
     };
--- a/doc/man7/EVP_MAC-CMAC.pod
+++ b/doc/man7/EVP_MAC-CMAC.pod
@@ -15,7 +15,7 @@ used with EVP_MAC_fetch():
 
 =over 4
 
-=item "CMAC", "provider=default"
+=item "CMAC", "provider=default" or "provider=fips"
 
 =back
 
--- a/doc/man7/EVP_MAC-CMAC.pod
+++ b/doc/man7/EVP_MAC-CMAC.pod
@@ -15,7 +15,7 @@ used with EVP_MAC_fetch():
 
 =over 4
 
-=item "CMAC", "default=yes"
+=item "CMAC", "provider=default"
 
 =back
 
--- a/doc/man7/EVP_MAC-GMAC.pod
+++ b/doc/man7/EVP_MAC-GMAC.pod
@@ -15,7 +15,7 @@ used with EVP_MAC_fetch():
 
 =over 4
 
-=item "GMAC", "default=yes"
+=item "GMAC", "provider=default"
 
 =back
 
--- a/doc/man7/EVP_MAC-GMAC.pod
+++ b/doc/man7/EVP_MAC-GMAC.pod
@@ -15,7 +15,7 @@ used with EVP_MAC_fetch():
 
 =over 4
 
-=item "GMAC", "provider=default"
+=item "GMAC", "provider=default" or "provider=fips"
 
 =back
 
--- a/doc/man7/EVP_MAC-HMAC.pod
+++ b/doc/man7/EVP_MAC-HMAC.pod
@@ -15,7 +15,7 @@ used with EVP_MAC_fetch():
 
 =over 4
 
-=item "HMAC", "provider=default"
+=item "HMAC", "provider=default" or "provider=fips"
 
 =back
 
--- a/doc/man7/EVP_MAC-HMAC.pod
+++ b/doc/man7/EVP_MAC-HMAC.pod
@@ -15,7 +15,7 @@ used with EVP_MAC_fetch():
 
 =over 4
 
-=item "HMAC", "default=yes"
+=item "HMAC", "provider=default"
 
 =back
 
--- a/doc/man7/EVP_MAC-KMAC.pod
+++ b/doc/man7/EVP_MAC-KMAC.pod
@@ -16,9 +16,9 @@ properties, to be used with EVP_MAC_fetch():
 
 =over 4
 
-=item "KMAC-128", "provider=default"
+=item "KMAC-128", "provider=default" or "provider=fips"
 
-=item "KMAC-256", "provider=default"
+=item "KMAC-256", "provider=default" or "provider=fips"
 
 =back
 
--- a/doc/man7/EVP_MAC-KMAC.pod
+++ b/doc/man7/EVP_MAC-KMAC.pod
@@ -16,9 +16,9 @@ properties, to be used with EVP_MAC_fetch():
 
 =over 4
 
-=item "KMAC-128", "default=yes"
+=item "KMAC-128", "provider=default"
 
-=item "KMAC-256", "default=yes"
+=item "KMAC-256", "provider=default"
 
 =back
 
--- a/doc/man7/property.pod
+++ b/doc/man7/property.pod
@@ -117,8 +117,8 @@ Ordering of optional clauses is not significant.
 =head2 Shortcut
 
 In order to permit a more concise expression of boolean properties, there
-is one short cut: a property name alone (e.g. "default") is
-exactly equivalent to "provider=default" in both definitions and queries.
+is one short cut: a property name alone (e.g. "myproperty") is
+exactly equivalent to "myproperty=yes" in both definitions and queries.
 
 =head2 Global and Local
 
--- a/doc/man7/property.pod
+++ b/doc/man7/property.pod
@@ -52,15 +52,15 @@ The full syntax for property definitions appears below.
 
 Each implementation of an algorithm can define any number of
 properties.
-For example, the default provider defines the property I<default=yes>
+For example, the default provider defines the property I<provider=default>
 for all of its algorithms.
-Likewise, the FIPS provider defines I<fips=yes> and the legacy provider
-defines I<legacy=yes> for all of their algorithms.
+Likewise, the FIPS provider defines I<provider=fips> and the legacy provider
+defines I<provider=legacy> for all of their algorithms.
 
 =head2 Queries
 
 A I<property query clause> is a single conditional test.
-For example, "fips=yes", "default!=yes" or "?iteration.count!=3".
+For example, "fips=yes", "provider!=default" or "?iteration.count!=3".
 The first two represent mandatory clauses, such clauses B<must> match
 for any algorithm to even be under consideration.
 The third clause represents an optional clause.
@@ -118,7 +118,7 @@ Ordering of optional clauses is not significant.
 
 In order to permit a more concise expression of boolean properties, there
 is one short cut: a property name alone (e.g. "default") is
-exactly equivalent to "default=yes" in both definitions and queries.
+exactly equivalent to "provider=default" in both definitions and queries.
 
 =head2 Global and Local
 
--- a/doc/man7/property.pod
+++ b/doc/man7/property.pod
@@ -117,8 +117,8 @@ Ordering of optional clauses is not significant.
 =head2 Shortcut
 
 In order to permit a more concise expression of boolean properties, there
-is one short cut: a property name alone (e.g. "myproperty") is
-exactly equivalent to "myproperty=yes" in both definitions and queries.
+is one short cut: a property name alone (e.g. "my.property") is
+exactly equivalent to "my.property=yes" in both definitions and queries.
 
 =head2 Global and Local
 
--- a/doc/man7/property.pod
+++ b/doc/man7/property.pod
@@ -54,8 +54,8 @@ Each implementation of an algorithm can define any number of
 properties.
 For example, the default provider defines the property I<provider=default>
 for all of its algorithms.
-Likewise, the FIPS provider defines I<provider=fips> and the legacy provider
-defines I<provider=legacy> for all of their algorithms.
+Likewise, OpenSSL's FIPS provider defines I<provider=fips> and the legacy
+provider defines I<provider=legacy> for all of their algorithms.
 
 =head2 Queries
 
--- a/doc/man7/provider.pod
+++ b/doc/man7/provider.pod
@@ -264,8 +264,8 @@ implementations of the same algorithms), the property "provider=default"
 can be used as a search criterion for these implementations. Some
 non-cryptographic algorithms (such as serializers for loading keys and
 parameters from files) are not FIPS algorithm implementations in themselves but
-are safe to use in "FIPS mode". The property "fips=yes" can be used to select
-such algorithms.
+support algorithms from the FIPS provider and are allowed for use in "FIPS
+mode". The property "fips=yes" can be used to select such algorithms.
 
 =head2 FIPS provider
 
--- a/doc/man7/provider.pod
+++ b/doc/man7/provider.pod
@@ -260,8 +260,12 @@ algorithm identifier to the appropriate fetching function.
 
 The default provider is built in as part of the F<libcrypto> library.
 Should it be needed (if other providers are loaded and offer
-implementations of the same algorithms), the property "default=yes"
-can be used as a search criterion for these implementations.
+implementations of the same algorithms), the property "provider=default"
+can be used as a search criterion for these implementations. Some
+non-cryptographic algorithms (such as serializers for loading keys and
+parameters from files) are not FIPS algorithm implementations in themselves but
+are safe to use in "FIPS mode". The property "fips=yes" can be used to select
+such algorithms.
 
 =head2 FIPS provider
 
@@ -269,8 +273,10 @@ The FIPS provider is a dynamically loadable module, and must therefore
 be loaded explicitly, either in code or through OpenSSL configuration
 (see L<config(5)>).
 Should it be needed (if other providers are loaded and offer
-implementations of the same algorithms), the property "fips=yes" can
-be used as a search criterion for these implementations.
+implementations of the same algorithms), the property "provider=fips" can
+be used as a search criterion for these implementations. All algorithm
+implementations in the FIPS provider can also be selected with the property
+"fips=yes".
 
 =head2 Legacy provider
 
@@ -278,7 +284,7 @@ The legacy provider is a dynamically loadable module, and must therefore
 be loaded explicitly, either in code or through OpenSSL configuration
 (see L<config(5)>).
 Should it be needed (if other providers are loaded and offer
-implementations of the same algorithms), the property "legacy=yes" can be
+implementations of the same algorithms), the property "provider=legacy" can be
 used as a search criterion for these implementations.
 
 =head1 EXAMPLES
@@ -300,21 +306,21 @@ Fetch any available implementation of AES-128-CBC in the default context:
 Fetch an implementation of SHA2-256 from the default provider in the default
 context:
 
- EVP_MD *md = EVP_MD_fetch(NULL, "SHA2-256", "default=yes");
+ EVP_MD *md = EVP_MD_fetch(NULL, "SHA2-256", "provider=default");
  ...
  EVP_MD_meth_free(md);
 
 Fetch an implementation of SHA2-256 that is not from the default provider in the
 default context:
 
- EVP_MD *md = EVP_MD_fetch(NULL, "SHA2-256", "default=no");
+ EVP_MD *md = EVP_MD_fetch(NULL, "SHA2-256", "provider!=default");
  ...
  EVP_MD_meth_free(md);
 
 Fetch an implementation of SHA2-256 from the default provider in the specified
 context:
 
- EVP_MD *md = EVP_MD_fetch(ctx, "SHA2-256", "default=yes");
+ EVP_MD *md = EVP_MD_fetch(ctx, "SHA2-256", "provider=default");
  ...
  EVP_MD_meth_free(md);
 
@@ -324,11 +330,11 @@ implementation of WHIRLPOOL from it:
  /* This only needs to be done once - usually at application start up */
  OSSL_PROVIDER *legacy = OSSL_PROVIDER_load(NULL, "legacy");
 
- EVP_MD *md = EVP_MD_fetch(NULL, "WHIRLPOOL", "legacy=yes");
+ EVP_MD *md = EVP_MD_fetch(NULL, "WHIRLPOOL", "provider=legacy");
  ...
  EVP_MD_meth_free(md);
 
-Note that in the above example the property string "legacy=yes" is optional
+Note that in the above example the property string "provider=legacy" is optional
 since, assuming no other providers have been loaded, the only implementation of
 the "whirlpool" algorithm is in the "legacy" provider. Also note that the
 default provider should be explicitly loaded if it is required in addition to
--- a/providers/defltprov.c
+++ b/providers/defltprov.c
@@ -20,7 +20,7 @@
 #include "prov/provider_util.h"
 #include "internal/nelem.h"
 
-#define ALGC(NAMES, FUNC, CHECK) { { NAMES, "default=yes", FUNC }, CHECK }
+#define ALGC(NAMES, FUNC, CHECK) { { NAMES, "provider=default", FUNC }, CHECK }
 #define ALG(NAMES, FUNC) ALGC(NAMES, FUNC, NULL)
 
 /* Functions provided by the core */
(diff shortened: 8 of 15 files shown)


providers/defltprov.c
```diff
@@ -331,134 +331,138 @@ static OSSL_ALGORITHM exported_ciphers[OSSL_NELEM(deflt_ciphers)];
...
     { NULL, NULL, NULL }
 };
 
+/*
+ * Unlike most algorithms in the default provider, the serializers are compatible
+ * for use in FIPS mode, and therefore have the "fips=yes" property.
+ */
 static const OSSL_ALGORITHM deflt_serializer[] = {
-    { "RSA", "default=yes,format=text,type=private",
+    { "RSA", "provider=default,fips=yes,format=text,type=private",
```
slontis: I am confused..
When we do a query with fips=n o doesnt that mean it wont find the serializer??

levitte: Er, I concur.  Why should the **default** provider have _anything_ with the property `fips=yes`?

mattcaswell: Because, "fips=yes" now means that the algorithm is compatible for use in FIPS mode. The serializers in the default provider *are* compatible and hence have the property. A query of "fips=no" means only algorithms that are not compatible for use in FIPS mode, and would rightly exclude the serializers. If you instead mean only those algorithms that are not in the FIPS *provider* then the query string should be "provider!=fips".

paulidale:
I need to consider this far more deeply.

providers/defltprov.c
paulidale: I cannot think of a case where _FIPS=no_ is used.  Perhaps non-US governments might but they'll not be asking for a NIST certified algorithm and thus they won't want a FIPS serialiser.

mspncp:
Let’s discuss this on Tuesday.

providers/defltprov.c
levitte: The absence of the fips property doesn't mean "fips=no", does it? I gotta admit that I'm a bit confused on that...

richsalz: I think knowing whether or not an algorithm is FIPS does not belong in OpenSSL; it belongs in the FIPS documents -- are you promising to keep current whenever they issue an update, for example? If it *does* end up in OpenSSL, I think it should be a separate stand-alone function, table-driven and global. It should not be part of the property lookup mechanism.

mattcaswell:
It seems like lots of people are getting very confused about what we're trying to do here. Let me restate the problem:

PR #11075, implements the function `FIPS_mode_set()`. It does this by setting the default property query "fips=yes". This means that any algorithm implementation that does not have this property against it will not be fetched. However this causes a problem because it means the serializers (which are not in the FIPS provider) will not be fetched - and yet it is perfectly ok for an application in fips mode to use those serializers. The problem is that the set of algorithms that are ok to use in FIPS mode is  not the same as the set of algorithms that are in the FIPS provider.  The set is bigger than just that.

Part of the problem is that our existing "fips" property means both "the set of algorithms loaded from the FIPS provider" *and* "the set of algorithms you can use in FIPS mode". This PR splits those two concepts.

levitte:
There's a problem with setting the property "fips=yes" in the default provider, though... It requires that the default provider gets loaded in the first place, and that's something we've explicitly avoided the need for whith the fips provider. This PR is not up to the point in its current form.

What's needed is a separate serializer provider. I would prefer to see it built in, for convenience. Do note that we want the same serializers available through the default provider too, for performance reasons. In the end, all that will differ with regards to the serializers will be which algorithm table we go through and the properties used in each.

mattcaswell:
> There's a problem with setting the property "fips=yes" in the default provider, though... It requires that the default provider gets loaded in the first place, and that's something we've explicitly avoided the need for whith the fips provider. This PR is not up to the point in its current form.

This PR does not attempt to solve the complete problem. It does solve *part* of the problem.

> What's needed is a separate serializer provider. I would prefer to see it built in, for convenience. Do note that we want the same serializers available through the default provider too...

All of this I fully agree with, and had that end goal in mind when writing this PR.

It should be possible to run in FIPS mode with either:
1) The FIPS and default provider loaded
2) The FIPS and separate serializer provider loaded
3) Just the FIPS provider and no serializer support

This PR allows (1) and (3), and will be needed for (2) to work.

levitte:
I may have overstated that a bit.  What I'm trying to say is that "fips=yes" in the default provider is incorrect, at least if we look at the objective that the default provider doesn't hold fips-usable algorithms, period.

levitte:
The consequence is that if you want serializers when in "fips mode", you will *have to* load the separate serializer module.  Of course.

mattcaswell:
>  at least if we look at the objective that the default provider doesn't hold fips-usable algorithms

Since when has that been an objective? Our objective has been to have a FIPS provider that can be validated. There is nothing in that which restricts what can go in the default provider.

> The consequence is that if you want serializers when in "fips mode", you will have to load the separate serializer module.

I don't agree with your premise, and therefore obviously I don't agree with your consequence.

levitte:
So if we're going to have "fips=yes" on the serializers that are in the default provider, then I don't see why we need a separate serializer provider, at all.  All that will do is duplicate almost exactly the same algorithm table for them, which seems quite unnecessary under these circumstances.

Note: in this "fips mode" vs "non-fips mode" debate, I don't see any importance in the provider name.  Yeah, they can still be nice to have if you do want to select algorithms based on that, and I do think that this PR fulfills that objective perfectly.

mattcaswell:
> So if we're going to have "fips=yes" on the serializers that are in the default provider, then I don't see why we need a separate serializer provider, at all.

Yes, I've considered this point. However I think there will be some people that will not want the default provider loaded at all - to avoid the possibility of "accidents". There will be others that may want the default provider loaded so that they can perform both FIPS operations and non-FIPS operations in the same application. So I think it is desirable to have both.

> All that will do is duplicate almost exactly the same algorithm table for them, which seems quite unnecessary under these circumstances.

Since they would both be "built-in" providers I think we could arrange things so that they both share the same physical table for the serializers - no duplication required.

> Note: in this "fips mode" vs "non-fips mode" debate, I don't see any importance in the provider name. Yeah, they can still be nice to have if you do want to select algorithms based on that, and I do think that this PR fulfills that objective perfectly.

Exactly.

providers/defltprov.c
beldmit: Sorry, I don't get a point. 

We have some algorithms that are allowed according to FIPS documents. 
We have various implementations of them. Some of them are FIPS-certified.

These 2 are different properties. What does "fips=yes" mean?

levitte:
> Since they would both be "built-in" providers I think we could arrange things so that they both share the same physical table for the serializers - no duplication required.

You're right! The only real difference would be the first downcall table, the one given back by the knit function.... and obviously and init function itself.

providers/defltprov.c
mattcaswell: "fips=yes" means it is allowed to be used in FIPS mode. This is different to it being an algorithm that is implemented in the FIPS provider (although it includes all of those). Essentially in FIPS mode you can use all the algorithms in the FIPS provider plus the serializers.

richsalz:
Based on https://github.com/openssl/openssl/pull/11097#issuecomment-586637695, PR #11075 is a mistake and should not be merged.

mattcaswell:
> Based on #11097 (comment), PR #11075 is a mistake and should not be merged.

#11075 is not the problem - it just exposed it. The problem exists anyway. This PR fixes the immediate issue. With this PR in place #11075 is perfectly fine to merge.

richsalz:
My main concern is at https://github.com/openssl/openssl/pull/11097#discussion_r379837254

slontis:
> Based on [#11097 (comment)](https://github.com/openssl/openssl/pull/11097#issuecomment-586637695), PR #11075 is a mistake and should not be merged.

The FIPS_mode_set in this PR calls an existing public API that can set properties per libctx (so yes the problem already existed).

richsalz:
> Part of the problem is that our existing "fips" property means both "the set of algorithms loaded from the FIPS provider" and "the set of algorithms you can use in FIPS mode". This PR splits those two concepts.

I view "fips" as meaning "this provider has been validated."  If my application cares about that the provider implements, it will either know out-of-band by reading the documentation (such as the OpenSSL architecture docs), or it will query to make sure the provider does the algorithms it wants.

It is a mistake to think of "FIPS mode" as something that still exists. What exists is providers, and which one(s) the application loads and uses.  That is also why I think "add FIPS_mode API" is wrong.  We have been saying that we don't *enforce* things in code, and FIPS-mode leads one to think that this is what is happening. Moving the "don't use non-FIPS implementations" is being left as administrative guidance.

richsalz:
> I may have overstated that a bit. What I'm trying to say is that "fips=yes" in the default provider is incorrect, at least if we look at the objective that the default provider doesn't hold fips-usable algorithms, period.

And I believe that this will lead to serious confusion and widespread mistakes amongst OpenSSL users who want FIPS-validated cryptography.

providers/defltprov.c
richsalz: >"fips=yes" means it is allowed to be used in FIPS mode.

NO.  This is wrong.  What is allowed to be used in FIPS mode is *FIPS-validated implementations.*  Algorithms that are acceptable are enumerated by NIST, but you can't use a non-validated implementation in FIPS mode.  This will cause immense confusion.

slontis: > I think knowing whether or not an algorithm is FIPS does not belong in OpenSSL; it belongs in the FIPS documents -- are you promising to keep current whenever they issue an update, for example?

You are entitled to your own view.
A fips provider is a list of certified algorithms..
e.g: https://csrc.nist.gov/projects/cryptographic-module-validation-program/Certificate/2473
It doesnt need to implement every single algorithm there is at any point in time - it can just be one algorithm if that is all that is needed.
Fips modules can also contain non fips algorithms (that can be accessed in non fips mode) if you really want.

You will also find the FIPS_mode() and FIPS_mode_set() listed in the design document.

richsalz:
> The FIPS_mode_set in this PR calls an existing public API that can set properties per libctx (so yes the problem already existed).

Yes, I get it, but it's still wrong to have that API with its *extensive* baggage present in this release. Since FOREVER, that function locked you into FIPS mode; it was a one-way switch, it couldn't be reversed. That's not what we're doing in this release, and there's no need for this outdated API, that is a convenience wrapper around an existing API, to exist with different semantics than what the entire OpenSSL FIPS-user community is used to.

providers/defltprov.c
richsalz: > You will also find the FIPS_mode() and FIPS_mode_set() listed in the design document.

And a wrapper mode for engines.  Your point being?

paulidale:
After more thought, I'm okay with this approach so I'm unblocking.

Another possibility would be to allow "fips=*" in a property definition.  The "*" being a wildcard that matches regardless of what is asked for.  I.e. it would match both "fips=yes" and "fips=no" queries.  I don't think I'd have it add to the score of an optional "?fips" however.  I don't think it is as streamlined as the solution here.

crypto/property/property_parse.c
```diff
@@ -579,7 +576,11 @@ int ossl_property_parse_init(OPENSSL_CTX *ctx)
 
     /* Pre-populate the two Boolean values */
     if ((ossl_property_true = ossl_property_value(ctx, "yes", 1)) == 0
-        || (ossl_property_false = ossl_property_value(ctx, "no", 1)) == 0)
+        || (ossl_property_false = ossl_property_value(ctx, "no", 1)) == 0
+        /* Create values for the built-in providers */
+        || ossl_property_value(ctx, "default", 1) == 0
```
paulidale: These aren't required.  Values are created on demand when parsing definitions.

The reason that _yes_ and _no_ are treated specially is to avoid lots of look ups when checking Booleans.  With this change we're moving away from Booleans and we could reconsider having these special values entirely.

doc/man7/property.pod
```diff
@@ -118,7 +118,7 @@ Ordering of optional clauses is not significant.
 
 In order to permit a more concise expression of boolean properties, there
 is one short cut: a property name alone (e.g. "default") is
-exactly equivalent to "default=yes" in both definitions and queries.
+exactly equivalent to "provider=default" in both definitions and queries.
```
paulidale: I'm quite uncomfortable with this short cut.  Booleans by default seems okay, providers a lot less so.

It is also not implemented :)

mattcaswell: This is actually more a case of me not properly reading the text I was updating rather than an intention to implement this short cut!

crypto/property/property_parse.c
mattcaswell: I backed this change out.

doc/man7/property.pod
mattcaswell: Fixed.

mattcaswell:
Fixup pushed addressing the latest feedback.

crypto/property/property_parse.c
```diff
@@ -562,12 +562,9 @@ OSSL_PROPERTY_LIST *ossl_property_merge(const OSSL_PROPERTY_LIST *a,
...
 {
     static const char *const predefined_names[] = {
-        "default",      /* Being provided by the default built-in provider */
-        "legacy",       /* Provided by the legacy provider */
-        "provider",     /* Name of provider (default, fips) */
+        "provider",     /* Name of provider (default, legacy, fips) */
         "version",      /* Version number of this provider */
-        "fips",         /* FIPS supporting provider */
-        "engine",       /* An old style engine masquerading as a provider */
+        "fips",         /* FIPS compatible algorithm */
```
levitte: Nit: I like "supporting" better than "compatible"

providers/defltprov.c
levitte: Regarding engines, we chose to leave them as they are.  A wrapper proved to be much harder than anticipated, and time runeth away.  That doesn't mean that we exclude a _future_ wrapper, though.

As far as I'm personally concerned, I wouldn't include `FIPS_mode()` or `FIPS_mode_set()` at all, but it was concluded very early (in the first design meeting) that there those that will want to keep using these.

So from an application point of view, ENGINEs are still going to be there in 3.0 as before, and `FIPS_mode()` / `FIPS_mode_set()` as well.

crypto/property/property_parse.c
mspncp: Or „FIPS conforming“?

providers/defltprov.c
mattcaswell: > NO. This is wrong. What is allowed to be used in FIPS mode is FIPS-validated implementations. Algorithms that are acceptable are enumerated by NIST, but you can't use a non-validated implementation in FIPS mode. This will cause immense confusion.

You do realise that what we are talking about here are the serializers? If you disable access to the serializers then you won't be able to:

- Save a generated private key for later use
- Send a public key to anyone
- Construct certificates
- Use TLS/CMS or any other similar protocol that transmits public keys between peers

FIPS without these capabilities is useless. AFAIK the OpenSSL FOM 2.0 did not disable access to these things.

levitte: There's a lot of side discussion in this comment thread.  My mistake picking up one of those side tracks...
The original concern seems to be resolved as far as I can tell, so...

crypto/property/property_parse.c
levitte: Not sure...  do the serializers conform to FIPS in any kind of way?

mattcaswell: I like "supporting"

levitte:
I've warmed up to this PR.  Looks fine to me

crypto/property/property_parse.c
mattcaswell: Fixed this, and tweaked the wording in a few other spots too. Fixup pushed.

levitte:
I would like to see another approval beside mine.  I know that it's more than what our criteria require, but safety in numbers, or something like that...

richsalz:
What does the "fips" property indicate?  Suppose there are two FIPS-validated providers that don't implement all the same algorithms?

mattcaswell:
> What does the "fips" property indicate?

An algorithm that is allowed for use in FIPS mode (either because it is a validated algorithm in itself, or is an allowed supporting algorithm). In practice it means a validated algorithm or a serializer.

The documentation in this PR that describes how "fips=yes" is used against algorithms in the default provider as follows:

"Some non-cryptographic algorithms (such as serializers for loading keys and
parameters from files) are not FIPS algorithm implementations in themselves but
support algorithms from the FIPS provider and are allowed for use in "FIPS
mode". The property "fips=yes" can be used to select such algorithms."

> Suppose there are two FIPS-validated providers that don't implement all the same algorithms?

I'm not too sure of the relevance of this.

doc/man7/property.pod
```diff
@@ -52,15 +52,15 @@ The full syntax for property definitions appears below.
 
 Each implementation of an algorithm can define any number of
 properties.
-For example, the default provider defines the property I<default=yes>
+For example, the default provider defines the property I<provider=default>
 for all of its algorithms.
-Likewise, the FIPS provider defines I<fips=yes> and the legacy provider
-defines I<legacy=yes> for all of their algorithms.
+Likewise, the FIPS provider defines I<provider=fips> and the legacy provider
```
richsalz: Is there only one FIPS provider?  Should that be "a FIPS provider"?  And didn't @slontis say that a FIPS-validated provider could have non-FIPS algorithms in it?

providers/defltprov.c
```diff
@@ -20,7 +20,7 @@
 #include "prov/provider_util.h"
 #include "internal/nelem.h"
 
-#define ALGC(NAMES, FUNC, CHECK) { { NAMES, "default=yes", FUNC }, CHECK }
+#define ALGC(NAMES, FUNC, CHECK) { { NAMES, "provider=default", FUNC }, CHECK }
```
richsalz: Is this syntactic sugar/shorthand really necessary?

richsalz:
Labelling a non-crytographic algorithm as "FIPS" is going to be confusing.  Labelling anything other than a FIPS-possible algorithm as "FIPS" is a bad idea.

You folks are too involved in this to see the big picture.

FIPS means validated implementation of a NIST-approved algorithm.  Do not redefine the term.  Create a new one if it's necessary.

t-j-h:
Actually "FIPS" means "The Federal Information Processing Standard" and there are a large number of those documents. It does not mean FIPS140-2 validated (or the prior FIPS140-1 or the pending FIPS140-3).  

Our use here is rather loose ... and this isn't aimed at your comment - just noting that use of "fips140" would be better ...

doc/man7/property.pod
levitte: It should probably say "OpenSSL's FIPS provider"

providers/defltprov.c
levitte: Yes. I have an upcoming script that needs something like this.

slontis:
> Actually "FIPS" means "Federal Information Processing Standard" and there are a large number of those documents.

Agreed..  The term 'fips module' has been used by openssl for quite a while, which implies this is the case.

paulidale:
With the potential opening of a major can of worms: might we consider provider dependencies?  Then the _default_ and _FIPS_ providers could both depend on _serialisers_.

I think this would be well out of scope for 3.0.

doc/man7/property.pod
paulidale: I'm okay leaving it as is or using @levitte's suggestion.

The FIPS provider can and I suspect will have non-validated algorithms, it is still the FIPS provider.

doc/man7/provider.pod
```diff
@@ -260,25 +260,31 @@ algorithm identifier to the appropriate fetching function.
...
 
 The FIPS provider is a dynamically loadable module, and must therefore
 be loaded explicitly, either in code or through OpenSSL configuration
 (see L<config(5)>).
 Should it be needed (if other providers are loaded and offer
-implementations of the same algorithms), the property "fips=yes" can
-be used as a search criterion for these implementations.
+implementations of the same algorithms), the property "provider=fips" can
+be used as a search criterion for these implementations. All algorithm
+implementations in the FIPS provider can also be selected with the property
```
paulidale: I'm still expecting e.g. MD5 to be in the FIPS provider (for TLS).  It can't be selected using "fips=yes".  It could be selected using "provider=fips" and then we're back to the beginning of this mess.

I don't think that this should prevent this PR moving forward.

mattcaswell:
> Do not redefine the term. Create a new one if it's necessary.

I'm open to considering alternative name suggestions (which could be done either in this PR or a follow up one)

doc/man7/provider.pod
richsalz: > I'm still expecting e.g. MD5 to be in the FIPS provider (for TLS)

I don't see how.  Looking at the IGhttps://csrc.nist.gov/csrc/media/projects/cryptographic-module-validation-program/documents/fips140-2/fips1402ig.pdf page 15, it says that if SHA1 is available than MD5 must not be used, for example.

paulidale: Several TLS versions require both SHA1 and MD5 to be present.

slontis: These fall into the category of "Non-FIPS Approved But Allowed Cryptographic Functions" See SP 800-135

richsalz:
It's not related to FIPS cryptographic algorithms, so a name like "fips-unrelated" or "not-cryptographic" might make sense.

mattcaswell:
> It's not related to FIPS cryptographic algorithms, so a name like "fips-unrelated" or "not-cryptographic" might make sense.

Its intended to convey the union of the algorithms in the (a?) FIPS module along with those other serializer type algorithms. So I'm not sure those names quite cover it.

richsalz:
I don't see a serializer as an algorithm.  FIPS means crypto.  Perhaps something along the lines of auxiliary or assistive works.

mattcaswell:
> I don't see a serializer as an algorithm. 

In OpenSSL terms they are. Providers provide algorithms communicated through an array of `OSSL_ALGORITHM` entries. There is one such `OSSL_ALGORITHM` entry for each serializer.

But whether you think of them as an algorithm or not doesn't really matter. The name we use should cover the union of algorithms in the FIPS module along with the serializers (whatever they are).

levitte:
So essentially, we make "fips=yes" mean "whatever is needed to enable FIPS validated algorithms, without introducing non-FIPS algorithms", i.e. it's not a strict definition like  what @richsalz tries to imply.

This comes down to practicality, and I can't see that as bad thing.

richsalz:
I don't need serializers to use a FIPS validated algorithms or provider.

You are taking a word which has a common meaning and extending the definition. I do not see why you think that's not a problem, why you do not think it will cause confusion for OpenSSL's users.

How about "facility" or "subsystem"?

A serializer is not an algorithm.

mattcaswell:
> You are taking a word which has a common meaning and extending the definition.

I'm not sure which word we're arguing about here "fips" or "algorithm".

If it's "algorithm" then that word is used extensively throughout IT. The Cambridge dictionary defines it as:

"a set of mathematical instructions or rules that, especially if given to a computer, will help to calculate an answer to a problem"

https://dictionary.cambridge.org/dictionary/english/algorithm

In this case the problem that serializers solve is "given an internal representation of a key/parameter give me back that key/parameter in well known format X". Our providers implement a set of instructions to solve that problem. They are very clearly algorithms in the common use of the term. They are probably not "cryptographic algorithms" in that they don't solve a cryptographic problem such as "key exchange", "signature", "encryption" etc.

levitte:
> I don't need serializers to use a FIPS validated algorithms or provider.

That's what you think...  it happens under the hood, any time there's a part of libcrypto that needs it for a key that only exists in the provider.  For example, a call of `i2d_PrivateKey` will call a provided serializer when the passed `EVP_PKEY` doesn't have an assigned legacy key, just an assigned provided key.

levitte:
`i2d_PrivateKey` was a bad example, as private keys are supposed not to escape from the FIPS provider.  However, there are serializers of public keys and parameters too, to text (used by functions like `EVP_PKEY_print_public()` and `EVP_PKEY_print_params()`, which use the text serializers, `X509_PUBKEY_set()`, which uses the public key to DER serializers, ...

So er, you might think that you don't need serializers, but...

mattcaswell:
> i2d_PrivateKey was a bad example, as private keys are supposed not to escape from the FIPS

Not so...its perfectly ok for private keys generated by the FIPS provider to be serialized into a file.

levitte:
> Not so...its perfectly ok for private keys generated by the FIPS provider to be serialized into a file.

Oh, right, then can be encrypted (PKCS#8 organization is assumed)

richsalz:
They *can* be used and perhaps they are used by the implementation in some corner cases, but my application does not need to use serializers.

As for terminology, I am saying that "FIPS algorithm" should not include seriaizers, but only mean crypto algorithms, and using the "fips" selector for this stuff is wrong.

mattcaswell:
> They can be used and perhaps they are used by the implementation in some corner cases, but my application does not need to use serializers.

Are you doing TLS? If so, you need them.

richsalz:
> Are you doing TLS? If so, you need them.

in my code?  Or does OpenSSL use them?

levitte:
> in my code? Or does OpenSSL use them?

libcrypto needs them, and by extension if not directly, so does libssl.  Does your application use those libraries?  Then the need extends to your application as well.

levitte:
You *could* of course make a study of exactly which of all possible code paths may or may not need some kind of serializations.  Chances are, though, that as soon as you deal with certificates and their public keys, you're getting there, because `X509_PUBKEY_set()`

richsalz:
I am sorry I am not explaining this well enough to make my point clear.

*MY* code does not use serializers.  *YOUR CODE* in its implementation of TLS does.  So what.  Why should the application care?  Any more than why should the application care about BIGNUM or ```bn->top```?  *it should not.*

We are, of course, arguging about pedantic side-points.  That's what programmers do.

"FIPS algorithm" means "crypto" and *only crypto.*  Saying otherwise is misleading to everyone not initimately familiar with the details of the project.  You are doing your users a dis-service here.

mattcaswell:
> in my code? Or does OpenSSL use them?

libssl will use the serializers. For example in a KeyExchange message or ClientHello key_share, libssl will ask the ephemeral DH/ECDH public key to serialize itself to send to the peer (not all of this code is there yet...but it will be by the time of 3.0).

levitte:
So @richsalz, if you dislike "fips=yes" so much for the purpose we (i.e. libcrypto and libssl) need, what do you suggest instead that's indicative enough?

(trying very hard to get out of the trenches)

richsalz:
I already did:
https://github.com/openssl/openssl/pull/11097#issuecomment-587025545
https://github.com/openssl/openssl/pull/11097#issuecomment-587006862

levitte:
@paulidale said:
> With the potential opening of a major can of worms: might we consider provider dependencies?

Actually, @mattcaswell and I were discussing an idea along those lines, but instead of introducing dependencies, which may indeed be a can of worms, we were thinking of a possible autoload feature, on a per operation basis if I recall correctly.  So for example, if nothing could be found for the operation `OSSL_OP_SERIALIZER` (i.e. no currently loaded provider would answer to that query), we would have programming indicating that the code should then try to load the "serializer" provider and try once more.

We even discussed making it possible to have the autoload feature configurable.

levitte:
> I already did:
> https://github.com/openssl/openssl/pull/11097#issuecomment-587025545
> https://github.com/openssl/openssl/pull/11097#issuecomment-587006862

Thanks.  Not sure what values would be expected for "assistive", "auxilliary", "facility" or "subsystem".  None of them are particularly intuitive...  But then, what about something far more generic?  "tag=*something*"?  "tag=fips"?  "tag=fips-support"?

doc/man7/provider.pod
t8m: Is TLS-1.0, 1.1 still something to worry about for FIPS? I would say that with the dependence of these protocol versions on SHA1 and prevalence of TLS-1.2 they do not have to be supported in FIPS mode anymore.

richsalz:
Autoload?  Doesn't this repeat the discussion of autoloading providers which we decided not to do by default?

richsalz:
> What about ...

Anything other than "fips"  The values would be which type of provider as in "assists=fips,default" for example.

t8m:
I have to say I am perfectly fine with fips=yes being indicative of something (crypto or non-crypto algorithm) that when used from OpenSSL is FIPS compliant. I do not know why the definition must be "fips approved crypto algorithm" and why any other definition would be confusing. And I am really scared by the @richsalz alternative proposals.

levitte:
> Autoload? Doesn't this repeat the discussion of autoloading providers which we decided not to do by default?

Kinda sorta...  we decided not to use built-in fallback providers when there are already loaded providers.  So this autoload feature would actually work in a similar manner, but on the more fine grained on-operation basis instead of on-whole-provider basis.

levitte:
@richsalz said:
> The values would be which type of provider as in "assists=fips,default" for example.

So a property query would have to be something that covers ( "fips=yes" xor "assist=fips" ) to satisfy you.  Trouble, we don't have the necessary property query syntax for it...  the closest we have would be "?fips=yes,?assist=yes", and I'm actually unsure what happens if none if the components match.  @paulidale?

mattcaswell:
> I have to say I am perfectly fine with fips=yes being indicative of something (crypto or non-crypto algorithm) that when used from OpenSSL is FIPS compliant. I do not know why the definition must be "fips approved crypto algorithm" and why any other definition would be confusing. And I am really scared by the @richsalz alternative proposals.

I could easily see a number of properties being defined, e.g.

fips_alg = a validated FIPS algorithm
fips_supporting = a non-FIPS algorithm that is allowed to be used in a FIPS compliant application
fips = The union of "fips_alg" and "fips_supporting"

I can't currently think of a better name for the union property other than just "fips". The suggestions so far don't seem right to me.

t8m:
> I could easily see a number of properties being defined, e.g.
> 
> fips_alg = a validated FIPS algorithm
> fips_supporting = a non-FIPS algorithm that is allowed to be used in a FIPS compliant application
> fips = The union of "fips_alg" and "fips_supporting"

I could certainly live with that although the usefulness of the two extra properties does not seem obvious to me.

levitte:
> the usefulness of the two extra properties does not seem obvious to me.

To me neither, but they are worth talking about to get out of the "you're wrong!  Am not!  Are too!" type of trench...

doc/man7/property.pod
mattcaswell: I fixed this using @levitte's suuggestion

mattcaswell:
I pushed a fixup to address the minor doc nit noted above. Please reconfirm!

richsalz:
The trench is my digging uphill to convince the project that using "fips" for this is wrong.  I lost.  That's okay, I lose most of my discussion points.  Still doesn't make me wrong nor you right. :)

paulidale:
> the closest we have would be "?fips=yes,?assist=yes"

This will match neither.  It prefers having the properties but not having them isn't a failure.

levitte:
That's what I suspected

t8m:
Still approved.

mattcaswell:
I was holding this one back to ensure #10631 could go in without merge conflicts. Now that has been merged, I've rebased and fixed the conflict issues.

@levitte / @t8m / @paulidale please can you reconfirm?

doc/man7/property.pod
```diff
@@ -117,8 +117,8 @@ Ordering of optional clauses is not significant.
 =head2 Shortcut
 
 In order to permit a more concise expression of boolean properties, there
-is one short cut: a property name alone (e.g. "default") is
-exactly equivalent to "default=yes" in both definitions and queries.
+is one short cut: a property name alone (e.g. "myproperty") is
```
paulidale: A thought: should _myproperty_ be changed to something that is legal.  E.g. _my.property_.  User defined properties need a dot in them.

This is completely optional.

providers/defltprov.c
```diff
@@ -331,136 +331,140 @@ static OSSL_ALGORITHM exported_ciphers[OSSL_NELEM(deflt_ciphers)];
...
+    { "EC:id-ecPublicKey", "provider=default", ec_keymgmt_functions },
+    { "X25519", "provider=default", x25519_keymgmt_functions },
+    { "X448", "provider=default", x448_keymgmt_functions },
 #endif
     { NULL, NULL, NULL }
 };
 
+/*
+ * Unlike most algorithms in the default provider, the serializers are allowed
+ * for use in FIPS mode, and therefore have the "fips=yes" property.
```
paulidale: Is it worth mentioning something along the lines of: _because they are not FIPS relevant_ here?

mattcaswell:
I addressed both issues raised by @paulidale.

test/recipes/30-test_evp_fetch_prov.t
```diff
@@ -60,28 +60,28 @@ unless ($no_fips) {
...
-              { args      => [ '-property', 'fips=yes' ],
-                message   => 'using property "fips=yes"' } ] }
+              { args      => [ '-property', 'provider!=default' ],
+                message   => 'using property "provider!=default"' },
+              { args      => [ '-property', 'provider=default' ],
+                message   => 'using property "provider=default"' },
+              { args      => [ '-property', 'provider!=fips' ],
+                message   => 'using property "provider!=fips"' },
+              { args      => [ '-property', 'provider=fips' ],
+                message   => 'using property "provider=fips"' } ] }
```
levitte: Actually, these tests aren't really very interesting. Don't get me wrong, of course we should test the boring properties, but the actually interesting one to test is "fips" with all its possible values, not to mention "-fips". With this change, that property, which be more crucial than any "provider=", is not tested at all.

In my opinion, this is subject for dismissal of approvals made so far.

mattcaswell: I've added  a bunch of new test cases to this test to cover the "fips" property.

mattcaswell:
@levitte - fixup pushed. Please take another look.

doc/man7/EVP_MAC-CMAC.pod
```diff
@@ -15,7 +15,7 @@ used with EVP_MAC_fetch():
 
 =over 4
 
-=item "CMAC", "default=yes"
+=item "CMAC", "provider=default"
```
levitte: Also "provider=fips"

doc/man7/EVP_MAC-GMAC.pod
```diff
@@ -15,7 +15,7 @@ used with EVP_MAC_fetch():
 
 =over 4
 
-=item "GMAC", "default=yes"
+=item "GMAC", "provider=default"
```
levitte: Also "provider=fips"

doc/man7/EVP_MAC-HMAC.pod
```diff
@@ -15,7 +15,7 @@ used with EVP_MAC_fetch():
 
 =over 4
 
-=item "HMAC", "default=yes"
+=item "HMAC", "provider=default"
```
levitte: Also "provider=fips"

doc/man7/EVP_MAC-KMAC.pod
```diff
@@ -16,9 +16,9 @@ properties, to be used with EVP_MAC_fetch():
 
 =over 4
 
-=item "KMAC-128", "default=yes"
+=item "KMAC-128", "provider=default"
```
levitte: Also "provider=fips"

doc/man7/EVP_MAC-KMAC.pod
```diff
@@ -16,9 +16,9 @@ properties, to be used with EVP_MAC_fetch():
...
 
-=item "KMAC-256", "default=yes"
+=item "KMAC-256", "provider=default"
```
levitte: Also "provider=fips"

levitte:
That was the last nits from my end, I think

mattcaswell:
All nits fixed and fixup pushed.

levitte:
@paulidale, @t8m, you might want to re-approve

mattcaswell:

mattcaswell:
Pushed. Thanks to everyone involved!

richsalz:
Sorry to write on a closed PR, does this require an update to the "EVP Configuration Module" section of doc/man5/config.pod ?

levitte:
I don't think so.

richsalz:
Hopefully someone who really understands what this PR did, can comment more definitively. :)

levitte:
Er... Not sure how to interpret what you just wrote.

The point is that "fips=yes" still means the same thing, at least mechanically. The philosophical discussion in "the meaning of FIPS" has done nothing to change that.

richsalz:
I didn't like "I don't think so."  If you're convinced no update is needed, that's fine.

levitte:
Is my expanded answer good enough?

richsalz:
> Is my expanded answer good enough?

Yes, it is definitive.

PR #11101, opened 2020-02-16
Coverity issues 1458438 and 1458439.

paulidale:
Fix Coverity problems 1458438 and 1458439

1. CID 1458438:  Memory - illegal accesses
1. CID 1458439:  Resource leaks

Also convert a sizeof(struct) to be sizeof(*pointer) in accordance with the coding style.

### DIFF
--- a/crypto/ffc/ffc_params.c
+++ b/crypto/ffc/ffc_params.c
@@ -15,7 +15,7 @@
 
 void ffc_params_init(FFC_PARAMS *params)
 {
-    memset(params, 0, sizeof(FFC_PARAMS));
+    memset(params, 0, sizeof(*params));
     params->pcounter = -1;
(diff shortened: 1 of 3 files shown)


crypto/ffc/ffc_params.c
```diff
@@ -15,7 +15,7 @@
 
 void ffc_params_init(FFC_PARAMS *params)
 {
-    memset(params, 0, sizeof(FFC_PARAMS));
+    memset(params, 0, sizeof(*params));
```
slontis: Hmm that's the second time I did that....

paulidale: It won't be the last for either of us.

paulidale:
Out of WIP.

paulidale:
Merged to master.  Thanks.

PR #11102, opened 2020-02-16
Improve small block cipher performance

kroeckx:
Avoid function calls we don't need to do.

In 1.1.1 we have:
aes-128-cbc     572267.80k   681197.08k   715430.74k   720508.59k   722359.64k   723004.07k

Current master:
aes-128-cbc     460663.70k   631125.66k   701283.58k   719794.52k   724732.59k   726668.63k

new:
aes-128-cbc     582057.64k   684288.62k   715721.90k   724856.15k   717578.24k   727176.53k


levitte:
@kroeckx, do you want someone else to merge for you?

kroeckx:
Feel free to merge it. I had other things I wanted to do, but I
guess there is no point in wating.

PR #11042, opened 2020-02-08
Memory code cleanup

davxy:
**Memory allocator generic code cleanup.**

----

The `CRYPTO_MEM_CHECK` modes (e.g. `CRYPTO_MEM_CHECK_OFF`) are supposed to be used only by the `CRYPTO_mem_ctrl()` function.

The `CRYPTO_mem_ctrl()` function is defined only if `OPENSSL_NO_CRYPTO_MDEBUG` is not
defined, thus makes sense to define the `CRYPTO_MEM_CHECK` modes under the same condition.

**Maybe the macros can be removed at all since the definitions are not used in the overall codebase:**
1. `CRYPTO_mem_ctrl(int type)` just returns -1 and ignores the type parameter
2. `CRYPTO_mem_ctr()` has been declared DEPRECATED from 3.0

----

Convenience typedefs for user-provided CRYPTO malloc, realloc and free functions to be used around the code.

----

The `crypto_ex_data_st` structure and macros were moved near the other CRYPTO_EX stuff and functions. Just to keep related stuff close to each other.

PR #11103, opened 2020-02-16
Mem sec cond code readability

davxy:
Mem -sec conditional code readability improvement.

Removed an apparently superfluous `if(1)` statement block.

-------

Commit #11042 has introduced a new, unused, `CRYPTO_EX_INDEX` macro to keep track of the number of available indices (`CRYPTO_EX_INDEX_COUNT`).

Maybe is better to keep the old `CRYPTO_EX_INDEX__COUNT` instead of the new introduced `CRYPTO_EX_INDEX_COUNT`

### DIFF
--- a/crypto/mem_sec.c
+++ b/crypto/mem_sec.c
@@ -227,9 +227,6 @@ size_t CRYPTO_secure_actual_size(void *ptr)
     return 0;
 #endif
 }
-/* END OF PAGE ...
-
-   ... START OF PAGE */
 
 /*
  * SECURE HEAP IMPLEMENTATION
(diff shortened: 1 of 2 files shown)


crypto/mem_sec.c
```diff
@@ -437,21 +434,19 @@ static int sh_init(size_t size, size_t minsize)
...
 #endif
     sh.map_size = pgsize + sh.arena_size + pgsize;
-    if (1) {
+
 #ifdef MAP_ANON
+    sh.map_result = mmap(NULL, sh.map_size,
+                         PROT_READ|PROT_WRITE, MAP_ANON|MAP_PRIVATE, -1, 0);
+#else
+    sh.map_result = MAP_FAILED;
+    /* Temporary use |ret| for file descriptor (assumes goto err returns 0) */
```
paulidale: Better to have a local variable just for this.
Somebody will eventually modify this code and not notice this double use.

davxy: Done

paulidale:
I can't review this anew since there are no changes.

paulidale:
Merged to master, thanks for the contribution.

PR #11106, opened 2020-02-17
Deprecate EC command line apps.

paulidale:
The _ec_ and _ecparam_ command line applications are deprecated.  Instead use the _pkey_, _genpkey_ and _pkeyparam_ commands.

### DIFF
--- a/apps/pkeyparam.c
+++ b/apps/pkeyparam.c
@@ -9,6 +9,7 @@
 
 #include <stdio.h>
 #include <string.h>
+#include <stdlib.h>
 #include "apps.h"
 #include "progs.h"
 #include <openssl/pem.h>
@@ -44,9 +45,11 @@ int pkeyparam_main(int argc, char **argv)
     ENGINE *e = NULL;
     BIO *in = NULL, *out = NULL;
     EVP_PKEY *pkey = NULL;
-    int text = 0, noout = 0, ret = 1, check = 0;
+    EVP_PKEY_CTX *ctx = NULL;
+    int text = 0, noout = 0, ret = EXIT_FAILURE, check = 0, r;
     OPTION_CHOICE o;
     char *infile = NULL, *outfile = NULL, *prog;
+    unsigned long err;
 
     prog = opt_init(argc, argv, pkeyparam_options);
     while ((o = opt_next()) != OPT_EOF) {
@@ -98,9 +101,6 @@ int pkeyparam_main(int argc, char **argv)
     }
 
     if (check) {
-        int r;
-        EVP_PKEY_CTX *ctx;
-
         ctx = EVP_PKEY_CTX_new(pkey, e);
         if (ctx == NULL) {
             ERR_print_errors(bio_err);
@@ -116,8 +116,6 @@ int pkeyparam_main(int argc, char **argv)
              * Note: at least for RSA keys if this function returns
              * -1, there will be no error reasons.
              */
-            unsigned long err;
-
             BIO_printf(out, "Parameters are invalid\n");
 
             while ((err = ERR_peek_error()) != 0) {
@@ -125,8 +123,8 @@ int pkeyparam_main(int argc, char **argv)
                            ERR_reason_error_string(err));
                 ERR_get_error(); /* remove err from error stack */
             }
+            goto end;
         }
-        EVP_PKEY_CTX_free(ctx);
     }
 
     if (!noout)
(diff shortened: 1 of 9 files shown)


apps/pkeyparam.c
```diff
@@ -125,6 +126,7 @@ int pkeyparam_main(int argc, char **argv)
                            ERR_reason_error_string(err));
                 ERR_get_error(); /* remove err from error stack */
             }
+            goto end;
```
levitte: Hmmm, free ctx?

paulidale: Yep, missed this one.  Fixed now.

test/recipes/15-test_ec.t
```diff
@@ -24,7 +24,7 @@ ok(run(test(["ectest"])), "running ectest");
 
 SKIP: {
     skip "Skipping EC conversion test", 3
-        if disabled("ec");
+        if disabled("ec") || disabled('deprecated-3.0');
```
romen: In addition to skipping this stanza, shouldn't we provide the same tests using `pkey`?

levitte: I'm of two minds on this. This recipe is primarily about the ec command, or so I thought. Isn't there another recipe that tests the pkey command?

paulidale: I've duplicated this stanza of tests, using `pkey` the second time around.

romen: @levitte I think it makes sense to have them in this test suite for now, while ec still exists, so that normally we test that the deprecated and replacement command do work.

We can then proceed to move all of this into a test_pkey/etc test once we are actually removing the ec command.

levitte: Will we remember to do that in 5+ years?
(hypothetically, 5+ years from now, "we" may be a different team, so "remember" also involves passing the plan on)

romen: We could add a comment about "this should be moved there once ec is removed", when ec is removed this file will cause build errors and whomever is removing ec will find the comment and can decide to ignore it completely because it does not make sense anymore ;P

paulidale: Only the first SKIP block will need removing when _ec_ is finally deleted.  I'd hope that anyone fixing things would at least look at the remaining tests and not touch them.

I don't see a pressing reason for a `test_pkey` move.  This one is called `test_ec` and that's what it is testing -- sure some uses the _ec_ command and the rest use _pkey_ but it is all _ec_ related.  Keeping the broad areas separate will help keep the noise down.

levitte: I had to look back at `test/README` to remind myself of the group numbers, and sure enough, this one is in the "individual asymmetric cipher algorithms" group, not "openssl commands" (which is 20-24).

So, er, never mind, but in that case, I'm starting to think that all commands used in this file could be changed to "pkey" ones.  Not duplicated, replaced.

paulidale: We'd lose the testing of the _ec_ command if this were done.

levitte: Well, here's a relevant question Emilia asked a few years ago: are we testing [this] or are we testing [that]. Here, translated to "are we testing the EC algorithm, or the EC command?" The underlying principle is that separate should test one aspect, not more.
We have been sloppy with that, but question is if we should keep being sloppy or if we should fix the sloppiness as we go. Or should we ditch that principle?

paulidale: We're testing EC algorithms here.  It is done via the command line using both of the two available commands.  This seems reasonable to me.

Might they be better in the 20-24 range?  Quite possibly.  Almost all of the 15- series tests are similar to this one in nature, if this moves they all should.

Would they be better separated into pkey and ec tests?  I'm less sure about that.  A pkey command test is going to be big and produce a lot of output, many parts of which aren't related.  The two commands used here are very closely related.

A refactoring of these tests into other places seems well beyond the scope of this PR.  An issue could be raised noting that we ought to go back and reconsider things.

levitte: I would be satisfied with a todo comment.

Also, considering we're moving to evp for everything, this test group seems less relevant for testing algorithms per se, since we have evp_test. 

We should probably have a 20-test_pkey.t, specifically to test `openssl pkey`. That's for another PR, though

paulidale: Okay, done.

paulidale:
Merged to master.  Rebase required fixing a conflict in CHANGES.
Thanks for the feedback.

PR #11114, opened 2020-02-18
Change DH_get_nid() to set the value of q if it is not already set

### LINKED ISSUE #11108: DH_generate_key requires q now
In the past, it was possible to call DH_generate_key when only (per documentation) the parameters p and g were set. Now, it seems that in some situations, q is required as well otherwise DH_generate_key fails with 0.

Pseudo code:
DH *dh = DH_new();
// p and g are provided, q is not
DH_set0_pqg(dh, p, NULL, g);
DH_generate_key(dh);

I was debugging this issue and this check in the generate_key function in dh_key.c https://github.com/openssl/openssl/blob/master/crypto/dh/dh_key.c#L249 fails because q is NULL:

            /*
             * The safe prime group code sets N = 2*s
             * (where s = max security strength supported).
             * N = dh->length (N = maximum bit length of private key)
             */
            if (dh->length == 0
                || dh->params.q == NULL
                || dh->length > BN_num_bits(dh->params.q))
                goto err;

DH_get_nid returns NID_modp_2048

slontis:
Fixes #11108.

It only sets q if a valid named group is found.
The function signature was recently changed to pass a non const DH pointer
in order to allow the nid to be cached internally. As an extension of this
the value of q can now also be set as q is always known for named groups.
If q is set then the length is also set for this case.

##### Checklist

### DIFF
--- a/crypto/dh/dh_group_params.c
+++ b/crypto/dh/dh_group_params.c
@@ -98,49 +98,74 @@ DH *DH_new_by_nid(int nid)
 
 int DH_get_nid(DH *dh)
 {
+    BIGNUM *q = NULL;
     int nid = dh->params.nid;
+    int nbits;
 
+    /* Just return if it is already cached */
     if (nid != NID_undef)
         return nid;
 
     if (BN_get_word(dh->params.g) != 2)
         return NID_undef;
-    if (!BN_cmp(dh->params.p, &_bignum_ffdhe2048_p))
+    if (!BN_cmp(dh->params.p, &_bignum_ffdhe2048_p)) {
         nid = NID_ffdhe2048;
-    else if (!BN_cmp(dh->params.p, &_bignum_ffdhe3072_p))
+        nbits = 225;
+    } else if (!BN_cmp(dh->params.p, &_bignum_ffdhe3072_p)) {
         nid = NID_ffdhe3072;
-    else if (!BN_cmp(dh->params.p, &_bignum_ffdhe4096_p))
+        nbits = 275;
+    } else if (!BN_cmp(dh->params.p, &_bignum_ffdhe4096_p)) {
         nid = NID_ffdhe4096;
-    else if (!BN_cmp(dh->params.p, &_bignum_ffdhe6144_p))
+        nbits = 325;
+    } else if (!BN_cmp(dh->params.p, &_bignum_ffdhe6144_p)) {
         nid = NID_ffdhe6144;
-    else if (!BN_cmp(dh->params.p, &_bignum_ffdhe8192_p))
+        nbits = 375;
+    } else if (!BN_cmp(dh->params.p, &_bignum_ffdhe8192_p)) {
         nid = NID_ffdhe8192;
+        nbits = 400;
 #ifndef FIPS_MODE
-    else if (!BN_cmp(dh->params.p, &_bignum_modp_1536_p))
+    } else if (!BN_cmp(dh->params.p, &_bignum_modp_1536_p)) {
         nid = NID_modp_1536;
+        nbits = 190;
 #endif
-    else if (!BN_cmp(dh->params.p, &_bignum_modp_2048_p))
+    } else if (!BN_cmp(dh->params.p, &_bignum_modp_2048_p)) {
         nid = NID_modp_2048;
-    else if (!BN_cmp(dh->params.p, &_bignum_modp_3072_p))
+        nbits = 225;
+    } else if (!BN_cmp(dh->params.p, &_bignum_modp_3072_p)) {
         nid = NID_modp_3072;
-    else if (!BN_cmp(dh->params.p, &_bignum_modp_4096_p))
+        nbits = 275;
+    } else if (!BN_cmp(dh->params.p, &_bignum_modp_4096_p)) {
         nid = NID_modp_4096;
-    else if (!BN_cmp(dh->params.p, &_bignum_modp_6144_p))
+        nbits = 325;
+    } else if (!BN_cmp(dh->params.p, &_bignum_modp_6144_p)) {
         nid = NID_modp_6144;
-    else if (!BN_cmp(dh->params.p, &_bignum_modp_8192_p))
+        nbits = 375;
+    } else if (!BN_cmp(dh->params.p, &_bignum_modp_8192_p)) {
         nid = NID_modp_8192;
-    else
+        nbits = 400;
+    } else {
         return NID_undef;
-
-    /* Verify q is correct if it exists - reset the nid if it is not correct */
-    if (dh->params.q != NULL) {
-        BIGNUM *q = BN_dup(dh->params.p);
-
-        /* Check q = p * 2 + 1 we already know q is odd, so just shift right */
-        if (q == NULL || !BN_rshift1(q, q) || (BN_cmp(dh->params.q, q) != 0))
-            nid = NID_undef;
-        BN_free(q);
     }
+    /* Set q = (p - 1) / 2 (p is known to be odd so just shift right ) */
+    q = BN_dup(dh->params.p);
+
+    if (q == NULL || !BN_rshift1(q, q)) {
+        nid = NID_undef; /* reset the nid if there was a failure */
+        goto err;
+    }
+    /* Verify q is correct if it exists */
+    if (dh->params.q != NULL) {
+        if (BN_cmp(dh->params.q, q) != 0)
+            nid = NID_undef;  /* reset the nid if q does not match */
+    } else {
+        /* assign the calculated q */
+        dh->params.q = q;
+        dh->length = nbits;
+        dh->dirty_cnt++;
+        q = NULL; /* set to NULL so it is not freed */
+    }
+err:
+    BN_free(q);
     dh->params.nid = nid; /* cache the nid */
     return nid;
 }
--- a/crypto/dh/dh_group_params.c
+++ b/crypto/dh/dh_group_params.c
@@ -22,6 +22,33 @@
 #include <openssl/objects.h>
 #include "crypto/bn_dh.h"
 #include "crypto/dh.h"
+#include "crypto/security_bits.h"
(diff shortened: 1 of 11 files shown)


crypto/dh/dh_group_params.c
```diff
@@ -126,15 +128,24 @@ int DH_get_nid(DH *dh)
     else
         return NID_undef;
 
-    /* Verify q is correct if it exists - reset the nid if it is not correct */
-    if (dh->params.q != NULL) {
-        BIGNUM *q = BN_dup(dh->params.p);
+    /* Calculate q = p * 2 + 1. We already know q is odd, so just shift right */
```
paulidale: The `q = p * 2 + 1` comment implies a left shift and increment.
Converting q to p would be a right shift.

slontis: Code is right as I compared it to the known values of q in the spec.. The comment is wrong.
It is correct in the code above. RFC7919 says q = (p-1) / 2..
The wiki on safe primes says the opposite :)

kroeckx:
I think this is wrong. I think the MODP groups are not safe
primes.

kroeckx:
They are probably safe primes anyway.

slontis:
> They are probably safe primes anyway.

According to this they do obey q = (p - 1) / 2
Not that there is any mention of it in the RFC.

https://crypto.stackexchange.com/questions/15819/are-the-rfc3526-modp-groups-schnorr-groups.

Thanks for scaring me :)

kroeckx:
There a are groups in RFCs that don't use safe primes for some
reason. I think there is the code assumes that if q is set that
it's a safe prime. We should probably document something related
to this.

slontis:
> I think there is the code assumes that if q is set that it's a safe prime. We should probably document something related to this.

I Temporarily added a prime test of q for all the modp and ffdhe groups that we have to verify this..

slontis:
@markt-asf let me know if this addresses the issue.

kroeckx:
The values you set for nbits are a little confusing.

slontis:
> The values you set for nbits are a little confusing.

Can you turn verbosity mode on :).
Is what you are asking is - are these the correct values?

Hmm I was assuming that whoever originally put the values into 
https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/dh/dh_rfc7919.c
got them right... So it doesnt look like they are. Is that your question??

Are you happy for me to change them to 2 * value in the tables in Sp800-56ar3?

Also SP800-56Ar3 does say that the modp group are safe primes.. (Thought I read it somewhere quite a while ago).

I tried using the horrible formula from 
https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Module-Validation-Program/documents/fips140-2/FIPS1402IG.pdf Section 7.5 (but I get the wrong answer)..

markt-asf:
Sorry, I still see the same issue. dh->length is still zero.

kroeckx:
I'm not sure what nbit is supposed to contain, but it's assigned to
length. I assume DH_get_length() returns this. According to the
manpage, that means:
The length parameter indicates the length of the secret exponent (private key) in bit

I didn't know that it returned that.

kroeckx:
So you now seem to assume that if q is set that length is also set. It might be want we want to do internally, but a user can set p and q, and not the length.

This is also a change in behavior if they only set p. By default we used to generate a private key that has 1 bit less then p, now we generate one that's twice as long as the security strength if we know p.

crypto/dh/dh_lib.c
```diff
@@ -205,11 +205,16 @@ int DH_set0_pqg(DH *dh, BIGNUM *p, BIGNUM *q, BIGNUM *g)
...
-
+    /*
+     * Check if this is a named group. If it finds a named group then the
+     * 'q' and 'length' value are either already set or are set by the
+     * call.
+     */
+    if (DH_get_nid(dh) == NID_undef) {
+        /* If its not a named group then set the 'length' if q is not NULL */
+        if (q != NULL)
+            dh->length = BN_num_bits(q);
```
kroeckx: This is not new behavior, but the code in 1.1.1 actually ignored the length if q was set, and uses BN_priv_rand_range(q). But from reading the manpage, this wasn't the behavior I expected, I expected it to remain 0. I'm also not sure why it's not using BN_priv_rand_range(p-1). So I think we should either not set dh->length in this case, or update the manpage.

doc/man3/DH_new_by_nid.pod
```diff
@@ -21,7 +21,11 @@ B<NID_modp_4096>, B<NID_modp_6144> or B<NID_modp_8192>.
 
 DH_get_nid() determines if the parameters contained in B<dh> match
 any named set. It returns the NID corresponding to the matching parameters or
-B<NID_undef> if there is no match.
+B<NID_undef> if there is no match. Internally it caches the nid, so that any
+subsequent calls can fetch the cached value.
+Note that if the value of parameter q is not set, then it is set to q = p * 2 + 1.
```
kroeckx: I assume that q is set to (p -1)/ 2. But I assume it would only do that in case it's a safe prime, which just happens to be the case.

kroeckx:
My previous comment is about the length not being set if q is set of course not the case, since the set0_pqg function did set it. But a user could override it and set it to 0 ...

doc/man3/DH_new_by_nid.pod
slontis: ok I see your point I will change it to be correct and put a comment about them all being safe primes.

slontis:
> My previous comment is about the length not being set if q is set of course not the case, since the set0_pqg function did set it. But a user could override it and set it to 0 ...

Its all a bit tricky isnt it.. I guess I then need to be a bit more careful about where I set it.. And cater for if maybe being zero in some places.

slontis:
> Sorry, I still see the same issue. dh->length is still zero.

Can you try again now..

markt-asf:
> > Sorry, I still see the same issue. dh->length is still zero.
> 
> Can you try again now..

Test passes now. Thanks for looking into this.

slontis:
@kroeckx Does this address all your comments now?

kroeckx:
You seem to have made some major changes on how you do this. I think my comments are addressed.

doc/man3/DH_new_by_nid.pod
```diff
@@ -24,8 +24,15 @@ B<NID_modp_1536>, B<NID_modp_2048>, B<NID_modp_3072>,
...
 
 DH_get_nid() determines if the parameters contained in B<dh> match
-any named set. It returns the NID corresponding to the matching parameters or
-B<NID_undef> if there is no match. This function is deprecated.
+any named safe prime group. It returns the NID corresponding to the matching
+parameters or B<NID_undef> if there is no match.
+Internally it caches the nid, so that any subsequent calls can fetch the
+cached value.
+Note that if the value of parameter q is not set, then it is set to
+q = (p - 1) / 2.
```
t8m: It should be clarified that the q is computed only if there is a match for the p:
```
If a matching p is found and if the value of parameter q is not set, then it is set to
q = (p - 1) / 2.
```

crypto/dh/dh_lib.c
t8m: Then I think we should correct the manpage.

doc/man3/DH_new_by_nid.pod
slontis: Updated

slontis:
> Then I think we should correct the manpage.

So you want DH_set0_pqg() to say something like?

This method checks if the parameters associated with I<p> and I<g> and optionally I<q> are associated with known safe prime groups. If it is a safe prime group then the value of q will be set to q = (p - 1 ) / 2 if q is NULL. For safe prime groups the optional length parameter is set to twice the value of the maximum_target_security_strength(BN_num_bits(p)) as listed in SP800-56Ar3 Table(s) 25 & 26.
If it is not a safe prime group then the optional length parameter will be set if q is not NULL to 
BN_num_bits(q).

levitte:
@t8m, ping

t8m:
Yes, the suggested text looks good to me.

slontis:
Add requested doc change.

doc/man3/DH_get0_pqg.pod
```diff
@@ -37,31 +37,38 @@ L<openssl_user_macros(7)>:
...
-the new values for B<p>, B<q> and B<g> as parameters to the function. Calling
+The I<p>, I<q> and I<g> values can be set by calling DH_set0_pqg() and passing
+the new values for I<p>, I<q> and I<g> as parameters to the function. Calling
 this function transfers the memory management of the values to the DH object,
 and therefore the values that have been passed in should not be freed directly
-after this function has been called. The B<q> parameter may be NULL.
+after this function has been called. The I<q> parameter may be NULL.
+DH_set0_pqg() also checks if the parameters associated with I<p> and I<g> and
+optionally I<q> are associated with known safe prime groups. If it is a safe
+prime group then the value of I<q> will be set to q = (p - 1 ) / 2 if I<q> is NULL.
```
t8m: Extraneous space `(p - 1 )`

doc/man3/DH_new_by_nid.pod
```diff
@@ -24,15 +24,22 @@ B<NID_modp_1536>, B<NID_modp_2048>, B<NID_modp_3072>,
...
+This function is deprecated.
 
 =head1 RETURN VALUES
 
 DH_new_by_nid() returns a set of DH parameters or B<NULL> if an error occurred.
 
-DH_get_nid() returns the NID of the matching set of parameters or
-B<NID_undef> if there is no match.
+DH_get_nid() returns the NID of the matching set of parameters for p and g
+and optionally q, otherwise it return B<NID_undef> if there is no match.
```
t8m: returns

slontis:
Updated

slontis:
thanks for reviewing.
merged to master

PR #11115, opened 2020-02-18
EdDSA support for DTLSv1.2 (1.1.1 branch)

biodranik:
This patch enables EdDSA support for DTLSv1.2 as specified in RFC 8422. While
the RFC does not explicitly state that these algorithms are applicable
to DTLS, the IANA registry contains DTLS-OK for the following entries:

  - Cipher Suites [1]

    TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 (RFC5289)
    TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA    (RFC8422)

  - Signature Algorithms [2]

    ed25519 (RFC8422)
    ed448   (RFC8422)

Testing:

  - TLSv1.2 handshake passes with the cipher suites of interest:

    $ openssl genpkey -algorithm ed25519 -out key.pem
    $ openssl req -new -out cert.csr -key key.pem
    $ openssl x509 -req -days 700 -in cert.csr -signkey key.pem -out cert.pem
    $ openssl s_server -tls1_2 -cert cert.pem -key key.pem -accept :12345
    $ openssl s_client -tls1_2 -connect localhost:12345

      New, TLSv1.2, Cipher is ECDHE-ECDSA-AES256-GCM-SHA384

  - DTLSv1.2 handshake fails with the cipher suites of interest

    $ openssl s_server -dtls1_2 -cert cert.pem -key key.pem -accept :12345
    $ openssl s_client -dtls1_2 -connect localhost:12345

      Server: [...]tls_post_process_client_hello:no shared cipher[...]
      Client: [...]dtls1_read_bytes:sslv3 alert handshake failure[...]

After the application of this patch DTLSv1.2 handshake succeeds.

[1] https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4
[2] https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-16

### DIFF
--- a/ssl/ssl_lib.c
+++ b/ssl/ssl_lib.c
@@ -3322,16 +3322,20 @@ void ssl_set_masks(SSL *s)
         if (ecdsa_ok)
             mask_a |= SSL_aECDSA;
     }
-    /* Allow Ed25519 for TLS 1.2 if peer supports it */
+    /* Allow Ed25519 for (D)TLS 1.2 if peer supports it */
     if (!(mask_a & SSL_aECDSA) && ssl_has_cert(s, SSL_PKEY_ED25519)
             && pvalid[SSL_PKEY_ED25519] & CERT_PKEY_EXPLICIT_SIGN
-            && TLS1_get_version(s) == TLS1_2_VERSION)
+            && (TLS1_get_version(s) == TLS1_2_VERSION
+                || (SSL_IS_DTLS(s)
+                    && DTLS_VERSION_GE(s->version, DTLS1_2_VERSION))))
             mask_a |= SSL_aECDSA;
 
-    /* Allow Ed448 for TLS 1.2 if peer supports it */
+    /* Allow Ed448 for (D)TLS 1.2 if peer supports it */
     if (!(mask_a & SSL_aECDSA) && ssl_has_cert(s, SSL_PKEY_ED448)
             && pvalid[SSL_PKEY_ED448] & CERT_PKEY_EXPLICIT_SIGN
-            && TLS1_get_version(s) == TLS1_2_VERSION)
+            && (TLS1_get_version(s) == TLS1_2_VERSION
+                || (SSL_IS_DTLS(s)
+                    && DTLS_VERSION_GE(s->version, DTLS1_2_VERSION))))
             mask_a |= SSL_aECDSA;
 #endif
 


kroeckx:
We asked the TLS list about this some months ago, but never got a
clear answer if the registry is correct or not.

mattcaswell:
Close/reopen to kick CLA bot

richsalz:
Try sending email to tls-reg-review@ietf.org

kaduk:
> We asked the TLS list about this some months ago, but never got a clear answer if the registry is correct or not.

Maybe I'm thinking of something else, but I thought we did get a clear answer ("the registry is correct").

biodranik:
@kroeckx @kaduk What to do with two opposite opinions? :)

kaduk:
@biodranik well, my opinion is more authoritative, clearly :)

In any case, I remembered that I can go look in the mailarchive from when the documents were getting approved, and while IANA did explicitly ask the authors of RFC 8422 about the DTLS-OK status for the hash algorithm entry '8', the ciphersuite registrations, being unchanged by 8422, did not get asked about.
The IANA review comments in the IETF datatracker for RFC 4492 (https://datatracker.ietf.org/doc/rfc4492/history/) do not say much on this topic, but do leave one with the conclusion that these ciphersuites have been DTLS-OK 'Y' in the registry for more than a decade.

mattcaswell:
> Maybe I'm thinking of something else, but I thought we did get a clear answer ("the registry is correct").

I don't feel that we've ever had a definitive response on this. IMO, at the moment RFC8422 and the registry do not agree with each other - specifically the entries 7 and 8 in the TLS Signature Algorithm registry are marked with DTLS-OK, but RFC8422 makes no mention of this.

kaduk:
> > Maybe I'm thinking of something else, but I thought we did get a clear answer ("the registry is correct").
> 
> I don't feel that we've ever had a definitive response on this. IMO, at the moment RFC8422 and the registry do not agree with each other - specifically the entries 7 and 8 in the TLS Signature Algorithm registry are marked with DTLS-OK, but RFC8422 makes no mention of this.

I checked with IANA.  They consulted with Yoav Nir (author of RFC 8422) when implementing the IANA actions for RFC 8422 and were explicitly told to use DTLS-OK of "Y" for the new signature algorithms.

mattcaswell:
> I checked with IANA. They consulted with Yoav Nir (author of RFC 8422) when implementing the IANA actions for RFC 8422 and were explicitly told to use DTLS-OK of "Y" for the new signature algorithms.

Ok. Well that seems fairly definitive about what the intent was - which should be good enough for this PR (although I think in that case the RFC should be updated to match this intent).

mattcaswell:
So, my next question is: why is this targeting the 1.1.1 branch? It seems like a new feature rather than a bug, and therefore doesn't seem appropriate for 1.1.1.

biodranik:
@mattcaswell because it's very useful now for many projects using DTLS/CoAP/OpenSSL. What version do you think is more appropriate?

mattcaswell:
> @mattcaswell because it's very useful now for many projects using DTLS/CoAP/OpenSSL. What version do you think is more appropriate?

As a matter of policy we do not add features to stable branches. Features can be added to the master branch (which is what will become OpenSSL 3.0).

richsalz:
> (although I think in that case the RFC should be updated to match this intent).

An errata to point out this bug in the RFC has been filed, https://www.rfc-editor.org/errata/eid6002  And if you go to the RFC and click on "errata" you can also see it.

kaduk:
The errata report is now marked as "Verified", too.

t8m:
Closing as we are not going to merge a new feature to 1.1.1 branch. Please open a PR against master if you're still interested in this.

PR #11119, opened 2020-02-18
Don't exclude quite so much in a no-sock build

mattcaswell:
We were excluding more code than we needed to in the OCSP/HTTP code in
the event of no-sock. We should also not assume that a BIO passed to our
API is socket based.

This fixes the no-sock build

### DIFF
--- a/crypto/http/http_client.c
+++ b/crypto/http/http_client.c
@@ -715,6 +715,7 @@ static BIO *HTTP_new_bio(const char *server, const char *server_port,
  end:
     return cbio;
 }
+#endif /* OPENSSL_NO_SOCK */
 
 static ASN1_VALUE *BIO_mem_d2i(BIO *mem, const ASN1_ITEM *it)
 {
@@ -731,6 +732,9 @@ static BIO *OSSL_HTTP_REQ_CTX_transfer(OSSL_HTTP_REQ_CTX *rctx)
 {
     int sending = 1;
     int rv;
+#ifndef OPENSSL_NO_SOCK
+    int issock = BIO_get_fd(rctx->rbio, NULL) > 0;
+#endif
 
     if (rctx == NULL) {
         HTTPerr(0, ERR_R_PASSED_NULL_PARAMETER);
@@ -743,9 +747,11 @@ static BIO *OSSL_HTTP_REQ_CTX_transfer(OSSL_HTTP_REQ_CTX *rctx)
             break;
         /* BIO_should_retry was true */
         sending = 0;
+#ifndef OPENSSL_NO_SOCK
         /* will not actually wait if rctx->max_time == 0 */
-        if (BIO_wait(rctx->rbio, rctx->max_time) <= 0)
+        if (issock && BIO_wait(rctx->rbio, rctx->max_time) <= 0)
             return NULL;
+#endif
     }
 
     if (rv == 0) {
@@ -840,12 +846,19 @@ BIO *OSSL_HTTP_transfer(const char *server, const char *port, const char *path,
 
     if (bio != NULL)
         cbio = bio;
+#ifndef OPENSSL_NO_SOCK
     else if ((cbio = HTTP_new_bio(server, port, proxy, proxy_port)) == NULL)
         return NULL;
+#else
+    else
+        return NULL;
+#endif
 
     (void)ERR_set_mark(); /* prepare removing any spurious libssl errors */
-    if (rbio == NULL && BIO_connect_retry(cbio, timeout) <= 0)
+#ifndef OPENSSL_NO_SOCK
+    if (bio == NULL && rbio == NULL && BIO_connect_retry(cbio, timeout) <= 0)
         goto end;
+#endif
     /* now timeout is guaranteed to be >= 0 */
 
     /* callback can be used to wrap or prepend TLS session */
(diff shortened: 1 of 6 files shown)


mattcaswell:
@DDvO please take a look at this. "no-sock" builds were failing in one of the tests. On investigation it seemed to me that we were excluding rather a lot of code in a "no-sock" build that actually had very little dependency on sockets. However there were a couple of places where we assumed that the underlying BIO we had been passed was based on a socket, which might not always be the case.

I've attempted to fix those issues, but I'm concerned that the test coverage of this seems light/non-existent. I suspect some/many of the codepaths I'm altering here have no real test coverage. Were you planning any additional tests in this area as part of the CMP contribution?

DDvO:
@mattcaswell, thanks for the investigation and the proposed fix. I've started reviewing it.

> On investigation it seemed to me that we were excluding rather a lot of code in a "no-sock" build that actually had very little dependency on sockets. However there were a couple of places where we assumed that the underlying BIO we had been passed was based on a socket, which might not always be the case.

Rright. The old HTTP client (which was in `crypto/ocsp/`) was always socket-based, and I had taken over this to the new one. Yet due to the generalizations you had requested the user may pass in the BIOs, which are not necessarily socket-based. I did not notice this and therefore had not adapted the usage of `#ifndef OPENSSL_NO_SOCK` accordingly.

> I've attempted to fix those issues, but I'm concerned that the test coverage of this seems light/non-existent. I suspect some/many of the codepaths I'm altering here have no real test coverage. Were you planning any additional tests in this area as part of the CMP contribution?

The tests that I've implemented in `test/http_test.c` cover nearly all aspects of the HTTP client _except_ socket-based I/O because this is a bit hard to test.
Yet the final chunk 12 of our CMP contribution will comprise of CLI-based tests, and those will include running a CMP mock server listening on a local port and interacting with the CMP client under test.

BTW, we faced some technical difficulties since the test hardness does not seem to support spawning another process in parallel to the application under test. We resorted to
```
sub start_mock_server {
 system ("LD_LIBRARY_PATH=../../ ../../apps/openssl cmp" .
            # " -config ../../test/default.cnf" .
            " -port $localport" .
...
            " $_[0]" . # optionally further CLI arguments
            " &")
}
```
and at the end of each test:
```
sub stop_mock_server {
    my $pid = `lsof -i | grep 'TCP .*:$localport' | awk '{ print \$2 }'`;
    system ("kill $pid") if $pid;
}
```
Is there a better/preferred way of doing this?

crypto/http/http_client.c
```diff
@@ -840,12 +846,19 @@ BIO *OSSL_HTTP_transfer(const char *server, const char *port, const char *path,
...
         return NULL;
+#else
+    else
+        return NULL;
+#endif
 
     (void)ERR_set_mark(); /* prepare removing any spurious libssl errors */
-    if (rbio == NULL && BIO_connect_retry(cbio, timeout) <= 0)
+#ifndef OPENSSL_NO_SOCK
+    if (bio == NULL && rbio == NULL && BIO_connect_retry(cbio, timeout) <= 0)
```
DDvO: Why did you add the `bio == NULL &&` part here? This leads to unintended behavior.
If `bio != NULL` and `rbio == NULL` this means, as documented, that the provided `bio` shall be used for connecting, writing request data, and reading response data.

crypto/http/http_client.c
```diff
@@ -1106,17 +1119,24 @@ int OSSL_HTTP_proxy_connect(BIO *bio, const char *server, const char *port,
     char *mbuf = OPENSSL_malloc(BUF_SIZE);
     char *mbufp;
     int read_len = 0;
-    int rv;
     int ret = 0;
     BIO *fbio = BIO_new(BIO_f_buffer());
+#ifndef OPENSSL_NO_SOCK
+    int rv;
```
DDvO: Better move the `int rv;` declaration inside the block guarded by `if (issocket)` since it is used only there.

crypto/http/http_client.c
```diff
@@ -1168,12 +1188,16 @@ int OSSL_HTTP_proxy_connect(BIO *bio, const char *server, const char *port,
...
     for (;;) {
         /* will not actually wait if timeout == 0 */
-        rv = BIO_wait(fbio, max_time);
-        if (rv <= 0) {
-            BIO_printf(bio_err, "%s: HTTP CONNECT %s\n", prog,
-                       rv == 0 ? "timed out" : "failed waiting for data");
-            goto end;
+#ifndef OPENSSL_NO_SOCK
+        if (issocket) {
+            rv = BIO_wait(fbio, max_time);
```
DDvO: Also here we now have the drawback of a busy loop if the I/O is not socket-based.

DDvO: This issue should at least be mentioned in the documentation.

DDvO:
Although the relaxation regarding `OPENSSL_NO_SOCK` does make sense,
it comes with the drawback that busy loops are incurred when using non-socket BIOs.

t8m:
Should there be at least some small sleep (`usleep()`) added to the busy loops?

mattcaswell:
I started to look at adding a usleep. This wouldn't be too difficult to do (we actually have some cross-platform usleep code in test/ssltestlib.c), but adding it started to turn this PR into something bigger than I had intended given time constraints working on the 3.0 tasks. So either I can:

1) Turn this PR over to someone else to pursue the usleep option

or

2) Implement the "simple" fix to correct no-sock builds for now (which is where I started this exercise). This would leave all the current code as being inaccessible in a no-sock build, which could be done at a later time.

DDvO:
@mattcaswell, I can take over.

I'd like to move the `ossl_sleep()` decls to some internal or maybe even external header. 
Which is best?

mattcaswell:
> @mattcaswell, I can take over.

Great. Would it be easiest if I close this PR, and you create a new one based on this?

> I'd like to move the ossl_sleep() decls to some internal or maybe even external header.
Which is best?

I'd say internal. Perhaps "internal/cryptlib.h"?

DDvO:
Fine.

crypto/http/http_client.c
```diff
@@ -1106,17 +1119,24 @@ int OSSL_HTTP_proxy_connect(BIO *bio, const char *server, const char *port,
...
+#endif
 
     if (bio == NULL || server == NULL || port == NULL
             || (bio_err != NULL && prog == NULL)) {
         HTTPerr(0, ERR_R_PASSED_NULL_PARAMETER);
         goto end;
     }
 
+#ifndef OPENSSL_NO_SOCK
+    issocket = BIO_get_fd(bio, NULL) > 0;
```
DDvO: @mattcaswell, were you sure to derive `issocket` from `bio`?
I think it should be from `fbio` instead.

mattcaswell: It shouldn't really make a difference. fbio is just a filter bio so should pass the BIO_C_GET_FD ctrl onto the underlying bio. Either should work.

mattcaswell: Of course if you  do it against fbio you have to do it *after* the `BIO_push(fbio, bio)` call below.

DDvO: Thanks for your quick response.
Meanwhile I think I found an even better solution that saves callers of `BIO_wait()` from worrying whether the BIO is a socket or not and whether sockets are supported.

mattcaswell:
Closing this in favour of #11134

PR #11120, opened 2020-02-18
DOC: Add documentation related to X509_LOOKUPs [1.1.1]

levitte:
Most of all, the base X509_LOOKUP functionality is now documented.
Furthermore, the names X509_LOOKUP_METHOD and X509_STORE are added for
reference.

Some functions were moved from X509_LOOKUP_meth_new.pod

### DIFF
--- a/doc/man3/X509_LOOKUP.pod
+++ b/doc/man3/X509_LOOKUP.pod
@@ -1,0 +1,179 @@
+=pod
+
+=head1 NAME
+
+X509_LOOKUP, X509_LOOKUP_TYPE,
+X509_LOOKUP_new, X509_LOOKUP_free, X509_LOOKUP_init,
+X509_LOOKUP_shutdown,
+X509_LOOKUP_set_method_data, X509_LOOKUP_get_method_data,
+X509_LOOKUP_ctrl,
+X509_LOOKUP_load_file, X509_LOOKUP_add_dir,
+X509_LOOKUP_get_store, X509_LOOKUP_by_subject,
+X509_LOOKUP_by_issuer_serial, X509_LOOKUP_by_fingerprint,
+X509_LOOKUP_by_alias
+- OpenSSL certificate lookup mechanisms
+
+=head1 SYNOPSIS
+
+ #include <openssl/x509_vfy.h>
+
+ typedef x509_lookup_st X509_LOOKUP;
+
+ typedef enum X509_LOOKUP_TYPE;
+
+ X509_LOOKUP *X509_LOOKUP_new(X509_LOOKUP_METHOD *method);
+ int X509_LOOKUP_init(X509_LOOKUP *ctx);
+ int X509_LOOKUP_shutdown(X509_LOOKUP *ctx);
+ void X509_LOOKUP_free(X509_LOOKUP *ctx);
+
+ int X509_LOOKUP_set_method_data(X509_LOOKUP *ctx, void *data);
+ void *X509_LOOKUP_get_method_data(const X509_LOOKUP *ctx);
+
+ int X509_LOOKUP_ctrl(X509_LOOKUP *ctx, int cmd, const char *argc,
+                      long argl, char **ret);
+ int X509_LOOKUP_load_file(X509_LOOKUP *ctx, char *name, long type);
+ int X509_LOOKUP_add_dir(X509_LOOKUP *ctx, char *name, long type);
+
+ X509_STORE *X509_LOOKUP_get_store(const X509_LOOKUP *ctx);
+
+ int X509_LOOKUP_by_subject(X509_LOOKUP *ctx, X509_LOOKUP_TYPE type,
+                            X509_NAME *name, X509_OBJECT *ret);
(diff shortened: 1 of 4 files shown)


levitte:
Ping?

doc/man3/X509_LOOKUP.pod
```diff
@@ -0,0 +1,179 @@
...
+                                const unsigned char *bytes, int len,
+                                X509_OBJECT *ret);
+ int X509_LOOKUP_by_alias(X509_LOOKUP *ctx, X509_LOOKUP_TYPE type,
+                          const char *str, int len, X509_OBJECT *ret);
+
+=head1 DESCRIPTION
+
+The B<X509_LOOKUP> structure holds the information needed to look up
+certificates and CRLs according to an associated L<X509_LOOKUP_METHOD(3)>.
+Multiple B<X509_LOOKUP> instances can be added to an L<X509_STORE(3)>
```
paulidale: Either use "a X509_STORE" or "an X509_STORE" consistently throughout.  Five lines down, "a" is used.  Here "an" is.

I think "a" is the more correct but I'd accept either so long as it is consistent.

doc/man3/X509_LOOKUP.pod
```diff
@@ -0,0 +1,179 @@
...
+its return value via I<*ret>.
+The meaning of the arguments depends on the I<cmd> number of the
+control command. In general, this function is not called directly, but
+wrapped by a macro call, see below.
+The control I<cmd>s known to OpenSSL are discussed in more depth
+in L</Control Commands>.
+
+X509_LOOKUP_load_file() passes a filename to be loaded immediately
+into the associated B<X509_STORE>.
+I<type> indicates what type of object is expected.
```
mattcaswell: We don't document what values are acceptable, i.e. we should talk about `X509_FILETYPE_PEM`, `X509_FILETYPE_ASN1` and `X509_FILETYPE_DEFAULT` here.

doc/man3/X509_LOOKUP.pod
```diff
@@ -0,0 +1,179 @@
...
+X509_LOOKUP_init() initializes the internal state and resources as
+needed by the given B<X509_LOOKUP> to do its work.
+
+X509_LOOKUP_shutdown() tears down the internal state and resources of
+the given B<X509_LOOKUP>.
+
+X509_LOOKUP_free() destructs the given B<X509_LOOKUP>.
+
+X509_LOOKUP_set_method_data() and X509_LOOKUP_get_method_data()
+associates and retrieves a pointer to application data to and from the
```
paulidale: Might this read better as:

```
X509_LOOKUP_set_method_data() associates an application data pointer with the given B<X509_LOOKUP>.

X509_LOOKUP_get_method_data() retrieves an application data pointer from the given B<X509_LOOKUP>.
```

That is, break it into two sentences.

It feels odd when I read it in its current state.

doc/man3/X509_LOOKUP.pod
levitte: I'm not English savvy enough to decide whether spelling according to sound ("ex five oh nine" starts with a vowel sound) or according to written letter ("X" being a consonant) is more correct.

So far, it seems that "an" (i.e. spell according to sound) is winning, but this question keeps coming back a little now and then.  So I'll go by that.

doc/man3/X509_LOOKUP.pod
levitte: I often juggle with compound statements like this, so it's become a habit.  I'll break them apart.

doc/man3/X509_LOOKUP.pod
richsalz: "an" is the correct as X sounds like a vowel.

levitte: So going by sound.  As I recall, this was also supported in a bunch of english discussion pages last time I did some research

mattcaswell: Agreed "an X509"

doc/man3/X509_LOOKUP.pod
levitte: Not quite...

https://github.com/openssl/openssl/blob/93c50f46809bc0bd6c39e8445a4140c034a4320d/include/openssl/x509_vfy.h#L47-L50

levitte: Ignore the now deleted comment...  *type* is ussed for different meanings in different functions, got me confused...

levitte: I've added a section on file types

doc/man3/X509_LOOKUP.pod
levitte: Fixed them all

doc/man3/X509_LOOKUP.pod
levitte: Done

levitte:
I've made all fixups a separate commit, for easy cherry pick up to master

PR #11122, opened 2020-02-18
Add *.d.tmp files to .gitignore

mattcaswell:
These are temporary files generated by the build process that should not
be checked in.


mattcaswell:
Pushed. Thanks.

Issue #10958, opened 2020-01-27
Unnecessary switch in bf_buff.c


eldakesh-ms:
Hi, I am a developer on the MSVC FE team and while porting some C++ warnings to C, our testing infrastructure found the warning triggered in this library. The offending construct is in `crypto/bio/bf_buff.c:421`, where there is a switch statement with no case labels and a default.
```
    switch (cmd) {
    default:
        ret = BIO_callback_ctrl(b->next_bio, cmd, fp);
        break;
    }
```
The fix is to just remove the switch statement. Just wanted to give a heads up.

mattcaswell:
Strange! Yes the fix seems straightforward.

I've marked this as a "good first issue". PRs welcome!

levitte:
This is an old copy-n-paste artifact. All ctrl functions are generally implemented as a switch statement, so this one simply follows an already set pattern.

RKinsey:
All but one of the xxxx_callback_ctrl functions (prefix_callback_ctrl in bf_prefix.c didn't) had default-only switches, I've removed them but left the early returns. However, I just noticed that bf_prefix uses BIO_next instead of directly calling b->next_bio like the others. I'm not sure if there's a reason for this that I'm missing or if it's okay to just swap over the other bf* files for consistency.

richsalz:
> bf_prefix uses BIO_next(b) instead of directly calling b->next_bio

It's okay to use internal data structures inside the module (BIO/bf in this case).  Being consistent would be helpful, if you want to fix that.

4cad:
I went ahead and added a PR that removes four more default switches from from this area.

WolfWings:
Was hunting through "Good First Issue" list, and looking into this one it appears all crypto/bio/* files have had constructs like this removed, so I think this issue can be closed?

t8m:
Yes, closing

PR #11125, opened 2020-02-19
Remove unneeded switch statement to fix warning

4cad:
https://github.com/openssl/openssl/issues/10958

This is a trivial change which has no functional impact, just to fix a couple of warnings.


paulidale:
Merged to master.   Thanks for checking the early change.

PR #11126, opened 2020-02-19
Refactor CRMF_poposigningkey_init() to work with provider keys

levitte:
The code in this function was almost entirely a copy of the
functionality in ASN1_item_sign(), so it gets refactored to actually
call ASN1_item_sign(), and thereby automatically gets support for
EVP_PKEYs with only provider side keys.


mattcaswell:
Looks ok, but it would be good to have an opinion from @DDvO

DDvO:
Would be good if `ASN1_item_sign()` and `ASN1_item_sign_ctx()` were documented.

levitte:
Yeah, I plan on doing that, just not nownownow.
Although, I really want to replace them with something better. They do too much in one function, and some bits are... a but clumsy.

PR #11130, opened 2020-02-19
Configuration: Restore linux-alpha-gcc+bwx

mattst88:
It was accidentally dropped in commit 7ead0c89185c ("Configure: fold
related configurations more aggressively and clean-up.") probably
because all but one of its bn_ops were removed (RC4_CHAR remains).

Bug: https://bugs.gentoo.org/697840

### DIFF
--- a/config
+++ b/config
@@ -498,10 +498,7 @@ case "$GUESSOS" in
 	OUT="ios64-cross" ;;
   alpha-*-linux2)
         ISA=`awk '/cpu model/{print$4;exit(0);}' /proc/cpuinfo`
-	case ${ISA:-generic} in
-	*[678])	OUT="linux-alpha+bwx-$CC" ;;
-	*)	OUT="linux-alpha-$CC" ;;
-	esac
+	OUT="linux-alpha-$CC"
 	if [ "$CC" = "gcc" ]; then
 	    case ${ISA:-generic} in
 	    EV5|EV45)		__CNF_CFLAGS="$__CNF_CFLAGS -mcpu=ev5"


mattst88:
I suspect the CLA is not needed, since I'm simply restoring bits that existed and were removed. Do I need to do anything to solve/avoid that?

paulidale:
You need to add a line to the commit message:
```
CLA: trivial
```

This sequence should do it:
```
git commit --amend   # edit the commit message adding the new line
git push -f
```

Configurations/10-main.conf
```diff
@@ -943,13 +943,24 @@ my %targets = (
...
         inherit_from     => [ "linux-generic64" ],
         lib_cppflags     => add("-DL_ENDIAN"),
         bn_ops           => "SIXTY_FOUR_BIT_LONG",
         asm_arch         => 'alpha',
         perlasm_scheme   => "void",
     },
+    "linux-alpha+bwx-gcc" => {
+        inherit_from     => [ "linux-generic64" ],
+        lib_cppflags     => add("-DL_ENDIAN"),
+        bn_ops           => "SIXTY_FOUR_BIT_LONG RC4_CHAR",
```
mattcaswell: This appears to be the same bn_ops as inherited from liunx-generic64, so this line seems redundant

t-j-h:
I don't think removal of that unusual combination was accidental - there is a comment saying use internal by config. I curious that anyone anywhere would be needing this specific combination ...

levitte:
The triggering line in `config` is this:

https://github.com/openssl/openssl/blob/ada66e78ef535fe80e422bbbadffe8e7863d457c/config#L502

In other words, `config` might very well set that as a target, and because that precise config target isn't present in `Configurations/*.conf`, it means that configuration on those machines (which are, if I understand correctly, the *last* and thereby *most current* Alpha models) fails.

From where I stand, this is very much a valid correction of an omission.

levitte:
The other option would be to remove the offending line from `config`

t-j-h:
I'd consider the error was then leaving that line in place rather than the removal.
Either way, don't feel my comment is a blocker at all on this - I though this was a deliberate cleanup but who knows.

levitte:
So there's a choice. In my book, removing that line in `config` is as acceptable as adding back the config target.

Someone choose.

Configurations/10-main.conf
mattst88: Thanks, fixed.

mattst88:
> To me this is sufficiently different from the original that I think a CLA is required.

I'm quite surprised, but okay.

> I noted that you are only targeting master. Probably you need a slightly different PR to make the same change in 1.1.1.

I was planning to open a pull request for 1.1.1 after this one is sorted.

> (which are, if I understand correctly, the last and thereby most current Alpha models)

Correct, EV[67] are the last alphas, so the most common configurations.

> The other option would be to remove the offending line from config

If we simply remove the `*[678]) OUT="linux-alpha+bwx-$CC" ;;` line from config, wouldn't that cause EV6/7 machines to get `linux-alpha-gcc` and lose `RC4_CHAR`? That doesn't seem like what we want.

levitte:
> If we simply remove the `*[678]) OUT="linux-alpha+bwx-$CC" ;;` line from config, wouldn't that cause EV6/7 machines to get `linux-alpha-gcc` and lose `RC4_CHAR`? That doesn't seem like what we want.

Your conclusion is correct.  As to what we want, the only Alphas I have to play with run OpenVMS, so it's difficult for me to judge which one of `RC4_CHAR` or `RC4_INT` is more optimal.  I assume that you have access to recent Alphas running Linux and can make a better judgement than me?  If `RC4_CHAR` is better with the latest EVs, then this PR does the right thing.

mattst88:
Thank you. I will do some benchmarks to confirm.

(It looks like I need to print/sign/scan/email the CLA?)

levitte:
> (It looks like I need to print/sign/scan/email the CLA?)

Yes.  Although, if it turns out that simply removing that single line in `config` is the move forward, I'd see that as trivial, so if a CLA feels demanding, I'd suggest waiting until you're done with your benchmarks.  On the other hand, if you plan on other contributions, you might as well deal with the CLA and be done with it.

mattst88:
Well, it looks like `RC4_CHAR` doesn't help, and maybe actually hurts vs `RC4_INT` on an EV7 CPU:

`RC4_CHAR`:
```
  Doing rc4 for 30s on 16 size blocks: 137605571 rc4's in 29.75s
  Doing rc4 for 30s on 64 size blocks: 36272843 rc4's in 29.75s
  Doing rc4 for 30s on 256 size blocks: 9188415 rc4's in 29.75s
  Doing rc4 for 30s on 1024 size blocks: 2306556 rc4's in 29.75s
  Doing rc4 for 30s on 8192 size blocks: 288619 rc4's in 29.75s
  Doing rc4 for 30s on 16384 size blocks: 144554 rc4's in 29.80s
  OpenSSL 1.1.1d  10 Sep 2019
  built on: Fri Feb 21 06:47:02 2020 UTC
  options:bn(64,64) rc4(char) des(int) aes(partial) idea(int) blowfish(ptr) 
  compiler: alpha-unknown-linux-gnu-gcc -fPIC -pthread -Wa,--noexecstack -Wall -O2 -mcpu=ev67 -mieee -pipe -fno-strict-aliasing -Wa,--noexecstack -DOPENSSL_USE_NODELETE -DL_ENDIAN -DOPENSSL_PIC -DOPENSSL_CPUID_O  BJ -DOPENSSL_BN_ASM_MONT -DSHA1_ASM -DGHASH_ASM -DZLIB -DNDEBUG  -DOPENSSL_NO_BUF_FREELISTS
  rc4              74003.93k    78027.21k    79061.51k    79384.23k    79469.30k    79483.93k
```

`RC4_INT`:
```
  Doing rc4 for 30s on 16 size blocks: 142539170 rc4's in 29.93s
  Doing rc4 for 30s on 64 size blocks: 37216875 rc4's in 29.98s
  Doing rc4 for 30s on 256 size blocks: 9403494 rc4's in 29.98s
  Doing rc4 for 30s on 1024 size blocks: 2355991 rc4's in 29.98s
  Doing rc4 for 30s on 8192 size blocks: 294927 rc4's in 29.98s
  Doing rc4 for 30s on 16384 size blocks: 147388 rc4's in 29.98s
  OpenSSL 1.1.1d  10 Sep 2019
  built on: Fri Feb 21 07:09:44 2020 UTC
  options:bn(64,64) rc4(int) des(int) aes(partial) idea(int) blowfish(ptr) 
  compiler: alpha-unknown-linux-gnu-gcc -fPIC -pthread -Wa,--noexecstack -Wall -O2 -mcpu=ev67 -mieee -pipe -fno-strict-aliasing -Wa,--noexecstack -DOPENSSL_USE_NODELETE -DL_ENDIAN -DOPENSSL_PIC -DOPENSSL_CPUID_O  BJ -DOPENSSL_BN_ASM_MONT -DSHA1_ASM -DGHASH_ASM -DZLIB -DNDEBUG  -DOPENSSL_NO_BUF_FREELISTS
  rc4              76189.54k    79442.55k    80284.96k    80459.73k    80584.57k    80538.07k
```

so it appears that simply removing the bwx configuration is the right way forward! I will update the patch tomorrow.

Thanks for the comments and help!

levitte:
Cool!  Thanks for doing the benchmark.

I frankly can't remember when or why the alpha+bwx config target was removed, and the commit (7ead0c89185c46378e3ed85c0012d083f4b3039b) only mentions an aggressive fold of related config targets, so it seems the omission was to forget to update `config` accordingly.

levitte:
Confirmed, it's trivial

mattst88:
Thanks! Is there anything else I need to do?

mattcaswell:
> Thanks! Is there anything else I need to do?

Nope.

CI failures are unexpected, but don't seem related to this PR.

Can someone else merge this please?

levitte:
Merged in master

19ded1a717b6c72c3db241f06787a353f1190755 config: Drop linux-alpha-gcc+bwx

levitte:
This should really be cherry-picked to 1.1.1 as well.  @mattcaswell, @paulidale, @t-j-h, do you agree?

mattcaswell:
I agree.

levitte:
Merged into 1.1.1

8d87aecfb2d15652fc4c2bb458b2379421305fca config: Drop linux-alpha-gcc+bwx

PR #11133, opened 2020-02-20
Add ECDH to fips provider

slontis:
Note: This PR has not attempted to move the curves into the provider dispatch table.
Mappings between the curve name / nid have been added to the inbuilt curve table.

##### Checklist

### DIFF
--- a/crypto/ec/ec_curve.c
+++ b/crypto/ec/ec_curve.c
@@ -20,6 +20,7 @@
 #include <openssl/obj_mac.h>
 #include <openssl/opensslconf.h>
 #include "internal/nelem.h"
+#include "e_os.h" /* strcasecmp required by windows */
 
 typedef struct {
     int field_type,             /* either NID_X9_62_prime_field or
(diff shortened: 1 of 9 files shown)


slontis:
Currently a WIP until the EC serializer has been added.

romen:
In the curve table, instead of manually inserting the strings, can't we use the defines? `SN_X9_62_prime256v1`, etc?

slontis:
> In the curve table, instead of manually inserting the strings, can't we use the defines? `SN_X9_62_prime256v1`, etc?

We could, but I dont see much point if we end up shifting this stuff into the dispatch table..
(This also took a while :))

levitte:
I assume that @romen meant you could use SN macros in ec_curve.c.  Considering you've put strings alongside actual NIDs, it's perfectly valid to use SN macros there.

slontis:
Can I do it as a seperate PR?

levitte:
> Can I do it as a seperate PR?

You may do it, or not (someone else might), in whatever way you like.  I've approved this regardless

slontis:
Rebased with minor changes to fix a duplicate function with a different name.

slontis:
Thanks. Merged to master.

PR #11134, opened 2020-02-20
 Don't exclude quite so much in a no-sock build, avoiding tight busy loops

DDvO:
This is the PR I've taken over from @mattcaswell's #11119:

> We were excluding more code than we needed to in the OCSP/HTTP code in
> the event of no-sock. We should also not assume that a BIO passed to our
> API is socket based.
> 
> This fixes the no-sock build.

The new solution also removes the exclusion on `BIO_wait()` and `BIO_connect_retry()`
and saves callers of `BIO_wait()` from worrying whether the BIO is a socket or not and whether sockets are supported at all.
Moreover, it helps avoiding tight busy loops when sockets are not available by resorting to some (system-specific) variant of sleep. The new HTTP client now makes use of this with a sleep time of 100 milliseconds.

For implementing these improvements I moved the definition of `ossl_sleep()` from `test/ssltestlib.c` to `include/internal/cryptlib.h` and moved `BIO_wait()` and `BIO_connect_retry()` from `crypto/bio/b_sock.c`  to `crypto/bio/bio_lib.c`.


DDvO:
Travis stalled and timed out after 10 minutes on one build - (how) can I force a retry?

mattcaswell:
This is an irritatingly common problem. Normally we just ignore it. You can just close/reopen to kick travis.

crypto/bio/bio_lib.c
```diff
@@ -784,3 +784,102 @@ void bio_cleanup(void)
...
+
+/* Internal variant of the below BIO_wait() not calling BIOerr() */
+static int bio_wait(BIO *bio, time_t max_time, unsigned long milliseconds)
+{
+    int fd;
+
+    if (max_time == 0)
+        return 1;
+
+    if (BIO_get_fd(bio, &fd) <= 0)
```
mattcaswell: Don't we want to do the ossl_sleep() in this case? This might happen with a non-socket based BIO.

crypto/bio/bio_lib.c
```diff
@@ -784,3 +784,102 @@ void bio_cleanup(void)
     CRYPTO_THREAD_lock_free(bio_type_lock);
     bio_type_lock = NULL;
 }
+
+/* Internal variant of the below BIO_wait() not calling BIOerr() */
+static int bio_wait(BIO *bio, time_t max_time, unsigned long milliseconds)
```
mattcaswell: Why "unsigned long" for milliseconds rather  than just "unsigned int"? We try to avoid "long".

test/ssltestlib.c
```diff
@@ -10,53 +10,21 @@
 #include <string.h>
 
 #include "internal/nelem.h"
+#include "internal/cryptlib.h" /* for ossl_sleep() */
```
mattcaswell: I'm surprised this actually works. I don't think `ossl_sleep` gets exported from libcrypto...so how are we able to use it in the test apps?

crypto/bio/bio_lib.c
DDvO: I'm not an export on sockets and file descriptors.
I thought that fds might be more generally applicable, but nevertheless I've changed it this way.

crypto/bio/bio_lib.c
DDvO: An amount of milliseconds can in general be quite hight, but for `bio_wait()` and `BIO_wait()` no very large numbers are needed. So I've changed this as well.

test/ssltestlib.c
DDvO: It works because `ossl_sleep()` is declared inline :)

DDvO: And BTW it originally was inline in `test/ssltestlib.c`, too.

DDvO:
@mattcaswell, requested changes are done. 
Travis just got stuck again on one of its builds, but all builds should be fine.

t8m:
Unfortunately this breaks API signature of BIO_wait. You'll have to add a new function and/or just use some default value.

DDvO:
No worries on compatiblity, `BIO_wait()` is a brand new function, introduced in #10667.

crypto/bio/bio_lib.c
```diff
@@ -784,3 +784,100 @@ void bio_cleanup(void)
...
+            BIOerr(0, rv == 0 ? BIO_R_CONNECT_TIMEOUT : BIO_R_CONNECT_ERROR);
+        } else {
+            rv = -1;
+            if (ERR_peek_error() == 0) /* missing error queue entry */
+                BIOerr(0, BIO_R_CONNECT_ERROR); /* workaround: general error */
+        }
+    }
+
+    return rv;
+}
```
levitte: `BIO_do_handshake()` is a macro closely related to `BIO_f_ssl()`, which means that `BIO_connect_retry()` is by extension.

Looking at previous patterns, we usually implement functions that are closely related to some BIO filter near the implementation of that filter.  In this case, that would be `ssl/bio_ssl.c`.  I would rather see this appear there.

levitte: `BIO_wait()` is more intuitive to have here, though...  what I would do to simplify matters is not having it record an error, but let the caller do it instead.  I agree that it seems cumbersome, but it will also end up being cumbersome to maintain an externally available internal function, so as trade-offs go...

t8m:
> No worries on compatiblity, `BIO_wait()` is a brand new function, introduced in #10667.

Ah, I am sorry for the false alarm.

crypto/bio/bio_lib.c
DDvO: I have the impression that the appearance of `BIO_do_handshake()` in `BIO_connect_retry()` is a bit misleading, and maybe I should revert that. Originally I had been using there `BIO_do_connect(), which is defined as `BIO_do_handshake()`, but the former for some reason is under `# ifndef OPENSSL_NO_SOCK`, so since moving `BIO_connect_retry()` out of that guard it cannot be used there, but  `BIO_do_handshake()` can.
I can imagine that `BIO_connect_retry()` does not fill well in `bio_lib.c`, but I don't think it fits in `bio_ssl.c` because the function is usable not only for SSL BIOs. Where else could I move it?

DDvO: Concerning where to throw the potential errors of `BIO_wait()`, it is not only a matter of convenience to do it within the function, but also a matter of clarity to the consumers of the error info: the error happened at the BIO level. Therefore I prefer to keep the error generation where it is now. I don't think it hurts to have a separate internal static function - I've seen a similar separation also in many other parts of the lib.

DDvO: How about moving the definition of `BIO_do_connect()` out of `#ifndef OPENSSL_NO_SOCK`?

test/ssltestlib.c
mattcaswell: Ah! Of course!

crypto/bio/bio_lib.c
levitte: I fear I may have had a too narrow view of this.  See, in *our* code, `BIO_do_handshake()` terminates in `crypto/bio/bss_conn.c` or `crypto/bio/bss_acpt.c`, and I failed to consider the unknown, BIOs in the wild that use entirely different mechanisms that we know nothing about.

And looking at this function, it's completely agnostic of the underlying BIO implementation, so it is more fitting here than I was thinking earlier.

In summary: never mind :wink:

DDvO:
Since this is already approved, shall I try forcing Travis again to re-do the stuck builds
although this would not be very environment friendly :wink: ?

mattcaswell:
> Since this is already approved, I wonder if I shall still try forcing Travis again to re-do the stuck builds

I wouldn't bother.

levitte:
I clicked it.  If it times out again, then I will bother no further with it

DDvO:
One of the builds timed out again :(
So I assume I can merge this after 24 hours, i.e., Friday early evening.

levitte:
> One of the builds timed out again :(

That's the same job that seems to be timing out all the time.  I've seen it in a gazillion PRs by now...

levitte:
> So I assume I can merge this after 24 hours, i.e., Friday early evening.

Yes

DDvO:
> That's the same job that seems to be timing out all the time. I've seen it in a gazillion PRs by now...

Hmm, then how about suspending that job from the Travis build list as long as it causes steady confusion?

DDvO:
Also independently of this PR, I still suggest moving the
```
#  define BIO_do_connect(b)       BIO_do_handshake(b)
```
out of `#ifndef OPENSSL_NO_SOCK` since it is inconsistent that 
```
# define BIO_do_handshake(b)     BIO_ctrl(b,BIO_C_DO_STATE_MACHINE,0,NULL)
```
is not under `#ifndef OPENSSL_NO_SOCK`.

DDvO:
Pushed. Thanks!

PR #11136, opened 2020-02-20
Support Linux Kernel TLS fro TLS 1.3 and new ciphers

junjunk:
Linux kernel 5.3 adds support for TLS 1.3 and AES_GCM256 and AES_CCM128 ciphers in Kernel TLS mode. This pull request adds support of that features in OpenSSL.

### DIFF
--- a/include/internal/ktls.h
+++ b/include/internal/ktls.h
@@ -111,6 +111,55 @@ static ossl_inline ossl_ssize_t ktls_sendfile(int s, int fd, off_t off,
 #  if defined(OPENSSL_SYS_LINUX)
 #   include <linux/version.h>
 
+#   if LINUX_VERSION_CODE < KERNEL_VERSION(4, 13, 0)
+
+struct tls_crypto_info {
+    unsigned short version;
+    unsigned short cipher_type;
+};
+
+#   else
+#    include <linux/tls.h>
+#   endif
+
+#   if LINUX_VERSION_CODE < KERNEL_VERSION(5, 2, 0)
+#    define TLS_CIPHER_AES_CCM_128				53
+#    define TLS_CIPHER_AES_CCM_128_IV_SIZE			8
+#    define TLS_CIPHER_AES_CCM_128_KEY_SIZE		16
+#    define TLS_CIPHER_AES_CCM_128_SALT_SIZE		4
+#    define TLS_CIPHER_AES_CCM_128_TAG_SIZE		16
+#    define TLS_CIPHER_AES_CCM_128_REC_SEQ_SIZE		8
+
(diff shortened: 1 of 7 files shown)


mattcaswell:
Close/reopen to kick CLA bot

include/internal/ktls.h
```diff
@@ -111,6 +111,55 @@ static ossl_inline ossl_ssize_t ktls_sendfile(int s, int fd, off_t off,
...
+#   if LINUX_VERSION_CODE < KERNEL_VERSION(4, 13, 0)
+
+struct tls_crypto_info {
+    unsigned short version;
+    unsigned short cipher_type;
+};
+
+#   else
+#    include <linux/tls.h>
+#   endif
```
kroeckx: I'm wondering if we need to include this, since we add everything we need if the headers are too old.

junjunk: I think it's better to include to prevent problems of possible ABI changes in newer versions of kernel

kroeckx: I have no idea why you think the ABI would be broken. If it does get broken, old applications will stop working, which is not acceptable.

junjunk: Well, do you think that it's better to remove this include and use own definition of structures?

kroeckx: Either we should just #include the file and not copy things from it, or have all the defines in our own source. My suggestion would be that we just #include the file, and that users will just need to compile to recent enough headers if they want to use it.

junjunk: Well, I was using the style that was introduced in the initial support for kTLS done by guys from Mellanox. Anyway, the case "just #include the file" forces us to make additional definition of what is supported on a kernel the library was built on, making it impossible to use new features on a newer kernel without rebuilding it from source. I mean, for example, Ubuntu have different versions of kernel for their 18.04 LTS release - 4.15.0 series, 4.18.0 series, 5.3.0 series. And change to another kernel version will force user to change or recompile the OpenSSL library. This is just an example of possible use-case of this strange definition. If you think that this is OK, please, confirm it in comment and I will rewrite the code with additional definitions of supported ciphers and TLS versions for kernel TLS case - I think it can be done fast.

junjunk:
>I think this also includes most of the changes from #10626 from what I can tell?
Yes, think work was done with #10626 in mind, so I've tried to make code reuse as possible as it can be.

t8m:
Is this replaced with #11589

junjunk:
Yes, I have made another PR and that's why I put this as a Draft

t8m:
So should it be closed?

junjunk:
Yes, I'll close it

Issue #11132, opened 2020-02-20
Compile fails on Solaris/x64 with syntax error in assembler 


jon-oracle:
I'm using Sun C v5.13 compiler and the configure line:

```
$ Configure --prefix=/export/home/nse/workspace/openssl-solaris10_x64_sunc513/install shared enable-trace -m64 solaris64-x86_64-cc
```
Which produced this configure output:

```
Command line (with current working directory = .):

    /export/home/nse/workspace/openssl-solaris10_x64_sunc513/local_perl/bin/perl ../Configure --prefix=/export/home/nse/workspace/openssl-solaris10_x64_sunc513/install shared enable-trace -m64 solaris64-x86_64-cc

Perl information:

    /export/home/nse/workspace/openssl-solaris10_x64_sunc513/local_perl/bin/perl
    5.24.0 for i86pc-solaris

Enabled features:

    aria
    asm
    async
    autoalginit
    autoerrinit
    autoload-config
    bf
    blake2
    camellia
    capieng
    cast
    chacha
    cmac
    cmp
    cms
    comp
    ct
    deprecated
    des
    dgram
    dh
    dsa
    dso
    dtls
    dynamic-engine
    ec
    ec2m
    ecdh
    ecdsa
    engine
    err
    filenames
    fips
    gost
    idea
    legacy
    md4
    mdc2
    module
    multiblock
    nextprotoneg
    pinshared
    ocb
    ocsp
    padlockeng
    pic
    poly1305
    posix-io
    psk
    rc2
    rc4
    rdrand
    rfc3779
    rmd160
    scrypt
    secure-memory
    seed
    shared
    siphash
    siv
    sm2
    sm3
    sm4
    sock
    srp
    srtp
    sse2
    ssl
    static-engine
    stdio
    tests
    threads
    tls
    trace
    ts
    ui-console
    whirlpool
    tls1
    tls1-method
    tls1_1
    tls1_1-method
    tls1_2
    tls1_2-method
    tls1_3
    dtls1
    dtls1-method
    dtls1_2
    dtls1_2-method

Disabled features:

    afalgeng            [not-linux]      OPENSSL_NO_AFALGENG
    asan                [default]        OPENSSL_NO_ASAN
    buildtest-c++       [default]        
    crypto-mdebug       [default]        OPENSSL_NO_CRYPTO_MDEBUG
    devcryptoeng        [default]        OPENSSL_NO_DEVCRYPTOENG
[...]

PR #1132, opened 2016-05-26
Avoid calling memcpy with lenght of 0

kroeckx:
We can call memcpy() with a pointer 1 past the last allocated byte and length
of 0 and you can argue that that's undefined behaviour.

Reported by tis-interpreter

PR #11137, opened 2020-02-20
x86_64: Replace .asciz "GNU" with .byte

hjl-tools:
Replace .asciz "GNU" with .byte since .asciz isn't supported on Solaris.
This should fix

https://github.com/openssl/openssl/issues/11132

##### Checklist

### DIFF
--- a/crypto/perlasm/x86_64-xlate.pl
+++ b/crypto/perlasm/x86_64-xlate.pl
@@ -115,7 +115,12 @@ if ($flavour =~ /elf/) {
 	.long 4f - 1f
 	.long 5
 0:
-	.asciz "GNU"
+	# "GNU" encoded with .byte, since .asciz isn't supported
+	# on Solaris.
+	.byte 0x47
+	.byte 0x4e
+	.byte 0x55
+	.byte 0
 1:
 	.p2align $p2align
 	.long 0xc0000002


crypto/perlasm/x86_64-xlate.pl
```diff
@@ -115,7 +115,10 @@
 	.long 4f - 1f
 	.long 5
 0:
-	.asciz "GNU"
+	.byte 0x47
```
t-j-h: a comment that this is "GNU" is appropriate to have here ... for future people who look at the code

hjl-tools: Done.

crypto/perlasm/x86_64-xlate.pl
```diff
@@ -115,7 +115,12 @@
 	.long 4f - 1f
 	.long 5
 0:
-	.asciz "GNU"
+	# Replace .asciz "GNU" with .byte since .asciz isn't supported
+	# on Solaris.
```
levitte: I understand that the comment is spur of the moment, but without the history, it looks odd.  May I suggest this?

```
	# "GNU" encoded with .byte, since .asciz isn't supported
	# on Solaris.

t-j-h: Approval stands with comment change. Good idea @levitte.

hjl-tools: Done.

slontis:
Thanks for addressing this so quickly...

Not quite sure on the format of the commit message.
It normally looks like 
Fixes #1132

It should not require reapproval if it is updated..

kroeckx:
Is it an option to just output this on platforms that understand it?

levitte:
> Is it an option to just output this on platforms that understand it?

It's only output on ELF platforms

slontis:
It doesnt look like this need to be applied to 1.1.1.

slontis:
Thanks @hjl-tools. Merged to master.

PR #11143, opened 2020-02-21
bugfix in cmac calculation example

asfakr:
The example never executes code inside of the while loop, as read()
returns bigger number than 0. Thus the end result is wrong.

CLA: trivial

##### Checklist


t8m:
Merged to master as 38e6c4907e

PR #11147, opened 2020-02-22
AES CTR-DRGB: do not leak timing information

p-steuer:
10.2.1.1 CTR_DRBG Internal State, NIST SP 800-90A Rev. 1 :
> The values of V and Key are the critical values of the internal state upon which the security of this DRBG mechanism depends (i.e., V and Key are the “secret values” of the internal state). 

So better not leak any timing information when incrementing V.

### DIFF
--- a/crypto/rand/drbg_ctr.c
+++ b/crypto/rand/drbg_ctr.c
@@ -12,6 +12,7 @@
 #include <openssl/crypto.h>
 #include <openssl/err.h>
 #include <openssl/rand.h>
+#include "crypto/modes.h"
 #include "internal/thread_once.h"
 #include "rand_local.h"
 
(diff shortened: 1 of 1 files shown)


slontis:
travis is not your friend :)

p-steuer:
missing include. build error should be fixed now.

p-steuer:
travis evp test fails probably not related to this PR

slontis:
Kicking off the builds again..

p-steuer:
travis red cross seems to be due to a time out

slontis:
Travis failure is just a timeout..

p-steuer:
Is this wanted for 1.1.1 ?

romen:
My approval does not count toward the threshold in this case, but personally I believe this fixes a minor potential vulnerability, so it would meet the requirements for 1.1.1 backporting.

If @paulidale agrees I would even think there is no need to restart the grace time counter on this.

t8m:
Travis failure is just an unrelated timeout.

paulidale:
For both branches too.

paulidale:
Merged to master and 1.1.1.  Apologies for letting this one slip through.

zz85:
this seems to be failing builds on Ubuntu (See #11487)

PR #11148, opened 2020-02-22
Rethink the EVP_PKEY cache of provider side keys

levitte:
The role of this cache was two-fold:

1.  It was a cache of key copies exported to providers with which an
    operation was initiated.
2.  If the EVP_PKEY didn't have a legacy key, item 0 of the cache was
    the corresponding provider side origin, while the rest was the
    actual cache.

This dual role for item 0 made the code a bit confusing, so we now
make a separate keymgmt / keydata pair outside of that cache, which is
the provider side "origin" key.

A hard rule is that an EVP_PKEY cannot hold a legacy "origin" and a
provider side "origin" at the same time.

This also adds evp_pkey_upgrade_to_provider(), for EVP_PKEY upgrades


levitte:
A small note on evp_pkey_upgrade_to_provider():  This should *never* become a public function, for the simple reason that it's temporary, even though it's obvious that it will be around for a few years.  The reason is that this duality between legacy keys and provider side keys is planned to go away when all the legacy stuff is gone, at which point the idea of an "upgrade" will be moot.

levitte:
This is WIP because I haven't updated the documentation yet.

crypto/evp/exchange.c
```diff
@@ -202,8 +202,8 @@ int EVP_PKEY_derive_init(EVP_PKEY_CTX *ctx)
 
     /* Ensure that the key is provided.  If not, go legacy */
```
slontis: probably update the comment

crypto/evp/exchange.c
```diff
@@ -202,8 +202,8 @@ int EVP_PKEY_derive_init(EVP_PKEY_CTX *ctx)
...
     tmp_keymgmt = ctx->keymgmt;
-    provkey = evp_pkey_make_provided(ctx->pkey, ctx->libctx,
-                                     &tmp_keymgmt, ctx->propquery);
+    provkey = evp_pkey_export_to_provider(ctx->pkey, ctx->libctx,
+                                          &tmp_keymgmt, ctx->propquery);
     if (provkey == NULL)
         goto legacy;
     if (!EVP_KEYMGMT_up_ref(tmp_keymgmt)) {
```
slontis: why is this one uprefed bu tthe set_peer one isnt?

crypto/evp/keymgmt_lib.c
```diff
@@ -47,33 +47,60 @@ void *evp_keymgmt_util_export_to_provider(EVP_PKEY *pk, EVP_KEYMGMT *keymgmt)
...
     if (keymgmt == NULL)
         return NULL;
 
     /* If we have an unassigned key, give up */
-    if (pk->pkeys[0].keymgmt == NULL)
+    if (pk->keymgmt == NULL)
         return NULL;
 
+    /* If |keymgmt| matches the "origin" |keymgmt|, no more to do */
+    if (pk->keymgmt == keymgmt)
```
slontis: If we had a provider with an single alternate implementation of an algorithm I could potentially see it wanting to share the same key manager code.. The code makes the assumption currently that each provider has its own key manager.

crypto/evp/keymgmt_lib.c
```diff
@@ -89,109 +116,107 @@ void *evp_keymgmt_util_export_to_provider(EVP_PKEY *pk, EVP_KEYMGMT *keymgmt)
...
-        /*
-         * The export function calls the callback (try_import), which does
-         * the import for us.  If successful, we're done.
-         */
-        if (evp_keymgmt_export(exp_keymgmt, exp_keydata,
-                               OSSL_KEYMGMT_SELECT_ALL,
-                               &try_import, &import_data))
-            break;
-
+    /*
```
slontis: Just a comment : This bit is a little bit cleaner with the seperation :)

crypto/evp/p_lib.c
```diff
@@ -990,34 +977,43 @@ void *evp_pkey_make_provided(EVP_PKEY *pk, OPENSSL_CTX *libctx,
         EVP_PKEY_CTX_free(ctx);
     }
 
+    /* If there's still no keymgmt to be haad, give up */
```
slontis: had

crypto/evp/p_lib.c
```diff
@@ -934,12 +945,10 @@ void *evp_pkey_make_provided(EVP_PKEY *pk, OPENSSL_CTX *libctx,
...
-     * the cache of provider keys.
+     * If there is a legacy "origin" key and it has changed, invalidate
+     * the operation cache of provider keys.
      */
     if (pk->pkey.ptr != NULL) {
-        EVP_KEYMGMT *legacy_keymgmt = NULL;
-
         /*
          * If there is no dirty counter, this key can't be used with
          * providers.
```
slontis: could ameth be null?

crypto/evp/p_lib.c
```diff
@@ -1059,3 +1075,96 @@ void *evp_pkey_make_provided(EVP_PKEY *pk, OPENSSL_CTX *libctx,
     EVP_KEYMGMT_free(allocated_keymgmt);
     return keydata;
 }
+
+#ifndef FIPS_MODE
+/*
+ * This differs from exporting in that releases the legacy key and assigns
```
slontis: that it releases

crypto/evp/p_lib.c
```diff
@@ -1059,3 +1075,96 @@ void *evp_pkey_make_provided(EVP_PKEY *pk, OPENSSL_CTX *libctx,
...
+     */
+    if (!ossl_assert(pk->keymgmt == NULL))
+        return NULL;
+
+    /* If the key is unassigned, give up */
+    if (pk->pkey.ptr == NULL)
+        return NULL;
+
+    /* If the legacy key doesn't have an export function, give up */
+    if (pk->ameth->export_to == NULL)
```
slontis: ameth ==  NULL check?

levitte:
I'm pondering if the separation between legacy "origin" and provider "origin" is a bit too strong. A slightly different model would be to have an upgrade simply use the legacy key as a template, and keep it around in case someone changes something in it...

crypto/evp/keymgmt_lib.c
richsalz: I don't think we should handle multiple implementations in a single provider, nor expect that such a mythical provider will have to externalize its key management between implementations.

At least not in this intial release.

levitte: Multiple implementations in a single provider? Where did you get that idea from?

richsalz: Did you not read Shane's comment just before mine?

levitte: Well, let's then say that we have a provider that implements SIGNATURE and KEYEXCH with some specific key. That means that it's reasonable to believe that they use the same key structure. Are you saying that each of them should implement their own handling of that same structure? Besides, what would an EVP_PKEY hang on to in such a scenario?

So yes, the current code not only assumes that every provider any asymmetric computation also comes with a keymgmt that holds on to the keys themselves, but *demands* it. 

This, BTW, has a corresponding thing in the legacy code: EVP_PKEY_ASN1_METHOD. There is one for every key type we support.

crypto/evp/p_lib.c
levitte: Fixed it

crypto/evp/p_lib.c
levitte: I've redone this part quite a bit, please look again

levitte:
My pondering lead to mods, not here, but in #11158.  Key upgrades must be used *very* carefully, the majority of cases only need to have a cached export.

crypto/evp/p_lib.c
levitte: This has been rewritten again, so a new look is probably best

crypto/evp/keymgmt_lib.c
levitte: Thanks :-)

crypto/evp/keymgmt_lib.c
levitte: > If we had a provider with an single alternate implementation of an algorithm I could potentially see it wanting to share the same key manager code..

So two providers where the OSSL_DISPATCH for the KEYMGMT happens to be the same, or even better, things like `->keymgmt-export` (to take an example)?  I've thought about the former, but haven't acted on it yet.  The latter gives me the heeby-jeebies for the moment, I need to investigate that line of thinking a little deeper.

levitte: Either way, I think that's for another PR, not this.

crypto/evp/exchange.c
levitte: Because the set_peer one doesn't init the ctx

crypto/evp/exchange.c
levitte: Done

crypto/evp/p_lib.c

levitte:
I did some squashing, for cleanup

levitte:
Docs have been added, this is no longer WIP

doc/internal/man3/evp_keymgmt_util_export_to_provider.pod
```diff
@@ -27,16 +30,24 @@ via a B<EVP_KEYMGMT> interface, if this hasn't already been done.
...
-To export a legacy key, use L<evp_pkey_make_provided(3)> instead, as
-this function deals purely with provider side keys and will not care
-to look at any legacy key.
+To export a legacy key, use L<evp_pkey_export_to_provider(3)> instead,
+as this function deals purely with provider side keys and will not
+care to look at any legacy key.
 
-evp_keymgmt_util_clear_pkey_cache() can be used to explicitly clear
-the cache of provider key references.
+evp_keymgmt_util_find_operation_cache_index() finds the location if
```
slontis: finds the index of 
I<keymgmt> in I<pk>'s cache of provider sides keys used for operations.

<Maybe this bit goes in the returns section??>
If I<keymgmt> is set to NULL then it returns the index of the first empty slot in the cache.

doc/internal/man3/evp_keymgmt_util_export_to_provider.pod
```diff
@@ -27,16 +30,24 @@ via a B<EVP_KEYMGMT> interface, if this hasn't already been done.
 It maintains a cache of provider key references in I<pk> to keep track
 of all provider side keys.
 
-To export a legacy key, use L<evp_pkey_make_provided(3)> instead, as
-this function deals purely with provider side keys and will not care
-to look at any legacy key.
+To export a legacy key, use L<evp_pkey_export_to_provider(3)> instead,
+as this function deals purely with provider side keys and will not
```
slontis: ignore any legacy key data

doc/internal/man3/evp_keymgmt_util_export_to_provider.pod
```diff
@@ -48,6 +59,11 @@ evp_keymgmt_export_to_provider() and evp_keymgmt_util_fromdata()
 return a pointer to the appropriate provider side key (created or
 found again), or NULL on error.
 
+evp_keymgmt_util_find_operation_cache_index() returns the index of the
+operation cache slot.  If there was no matching slot, it returns the
+index of the first matching slot, as if I<keymgmt> is NULL, or the
```
slontis: first matching slot is not right - maybe just use the text I had above (or something similar)

doc/internal/man3/evp_pkey_export_to_provider.pod
```diff
@@ -0,0 +1,72 @@
...
+
+=head1 DESCRIPTION
+
+evp_pkey_export_to_provider() exports the "origin" key contained in I<pk>
+to its operation cache to make it suitable for an B<EVP_KEYMGMT> given either
+with I<*keymgmt> or with an implicit fetch using I<libctx> (NULL means the
+default context), the name of the legacy type of I<pk>, and I I<propquery>
+(NULL means the default property query settings).
+
+If I<keymgmt> isn't NULL but I<*keymgmt> isn't, and the "origin" was
```
slontis: is NULL ??

doc/internal/man3/evp_pkey_export_to_provider.pod
```diff
@@ -0,0 +1,72 @@
...
+ void *evp_pkey_export_to_provider(EVP_PKEY *pk, OPENSSL_CTX *libctx,
+                                   EVP_KEYMGMT **keymgmt,
+                                   const char *propquery);
+ void *evp_pkey_upgrade_to_provider(EVP_PKEY *pk, OPENSSL_CTX *libctx,
+                                    EVP_KEYMGMT **keymgmt,
+                                    const char *propquery);
+
+=head1 DESCRIPTION
+
+evp_pkey_export_to_provider() exports the "origin" key contained in I<pk>
```
slontis: Its not clear what "origin" key is here (to someone new reading this)

doc/internal/man3/evp_pkey_export_to_provider.pod
```diff
@@ -0,0 +1,72 @@
...
+with I<*keymgmt> or with an implicit fetch using I<libctx> (NULL means the
+default context), the name of the legacy type of I<pk>, and I I<propquery>
+(NULL means the default property query settings).
+
+If I<keymgmt> isn't NULL but I<*keymgmt> isn't, and the "origin" was
+successfully exported, then I<*keymgmt> is assigned the implicitly fetched
+B<EVP_KEYMGMT>.
+
+evp_pkey_upgrade_to_provider() exports the legacy "origin" key contained in
+I<pk> to it's provided counterpart, then clears the legacy "origin" key along
```
slontis: I am not sure I like the use of the word 'provided'. maybe 'provider side'

doc/internal/man3/evp_pkey_export_to_provider.pod
```diff
@@ -0,0 +1,72 @@
...
+ void *evp_pkey_upgrade_to_provider(EVP_PKEY *pk, OPENSSL_CTX *libctx,
+                                    EVP_KEYMGMT **keymgmt,
+                                    const char *propquery);
+
+=head1 DESCRIPTION
+
+evp_pkey_export_to_provider() exports the "origin" key contained in I<pk>
+to its operation cache to make it suitable for an B<EVP_KEYMGMT> given either
+with I<*keymgmt> or with an implicit fetch using I<libctx> (NULL means the
+default context), the name of the legacy type of I<pk>, and I I<propquery>
```
slontis: and the I<propquery>

doc/internal/man3/evp_pkey_export_to_provider.pod
```diff
@@ -0,0 +1,72 @@
...
+B<EVP_KEYMGMT>.
+
+evp_pkey_upgrade_to_provider() exports the legacy "origin" key contained in
+I<pk> to it's provided counterpart, then clears the legacy "origin" key along
+with other legacy data, and resets all the caches.  Otherwise, it works like
+evp_pkey_export_to_provider().
+
+I<evp_pkey_upgrade_to_provider() must be used with great care, only if there's
+no other way.>
+Most of the times, it's sufficient to use evp_pkey_export_to_provider(), but
```
slontis: time

doc/internal/man3/evp_pkey_export_to_provider.pod
```diff
@@ -0,0 +1,72 @@
...
+
+=head1 RETURN VALUES
+
+evp_pkey_export_to_provider() and evp_pkey_upgrade_to_provider() both return
+the provider key data that was exported if the "origin" was successfully
+exported to its target.  Otherwise, NULL is returned.
+
+=head1 NOTES
+
+Some functions calling evp_pkey_make_provided() may have received a const
```
slontis: evp_pkey_make_provided() got renamed..

doc/internal/man7/EVP_PKEY.pod
```diff
@@ -0,0 +1,55 @@
...
+level key types, such as B<RSA>, B<DSA> and B<EC> --, but also the
+provided form of keys -- i.e. pointers to provider side key data.
+Those two forms are mutually exclusive; an B<EVP_PKEY> instance can't
+contain both a key in legacy form and in provided form.  Regardless of
+form, this key is commonly refered to as the "origin".
+
+An B<EVP_PKEY> also contains a cache of provided copies of the key,
+each adapted for the provider that is going to use that copy to
+perform some operation.  This does of course only the implementation
+of the "origin" key supports exporting the key data, otherwise the
```
slontis: This sentence is hard to understand..

This does of course only the implementation
of the "origin" key supports exporting the key data.

doc/internal/man7/EVP_PKEY.pod
```diff
@@ -0,0 +1,55 @@
...
+OP_keymgmt_export() must be implemented.  In all cases, the receiving
+B<EVP_KEYMGMT> must have an implemented OP_keygmt_import().
+
+=head1 SEE ALSO
+
+L<provider-keymgmt(7)>
+
+=head1 COPYRIGHT
+
+Copyright 2019 The OpenSSL Project Authors. All Rights Reserved.
```
slontis: 2020

crypto/evp/p_lib.c
```diff
@@ -866,28 +867,33 @@ int EVP_PKEY_up_ref(EVP_PKEY *pkey)
     return ((i > 1) ? 1 : 0);
 }
 
+static void evp_pkey_free_legacy(EVP_PKEY *x)
+{
+    if (x->ameth && x->ameth->pkey_free) {
```
slontis: == NULL && == NULL

crypto/evp/p_lib.c
```diff
@@ -866,28 +867,33 @@ int EVP_PKEY_up_ref(EVP_PKEY *pkey)
     return ((i > 1) ? 1 : 0);
 }
 
+static void evp_pkey_free_legacy(EVP_PKEY *x)
```
slontis: entire function should not exist in fips_mode probably.

crypto/evp/p_lib.c
```diff
@@ -1053,3 +1059,100 @@ void *evp_pkey_export_to_provider(EVP_PKEY *pk, OPENSSL_CTX *libctx,
...
+        if ((keydata = evp_keymgmt_newdata(tmp_keymgmt)) == NULL)
+            goto end;
+
+        if (!pk->ameth->export_to(pk, keydata, tmp_keymgmt)) {
+            evp_keymgmt_freedata(tmp_keymgmt, keydata);
+            keydata = NULL;
+            goto end;
+        }
+
+        EVP_KEYMGMT_up_ref(tmp_keymgmt);
```
slontis: Probably need to start checking these up_ref calls return results..

doc/internal/man3/evp_keymgmt_util_export_to_provider.pod
levitte: I rephrased a bit, both here and in RETURN VALUES.

doc/internal/man3/evp_keymgmt_util_export_to_provider.pod
levitte: Fixed

doc/internal/man3/evp_pkey_export_to_provider.pod
levitte: Fixed

doc/internal/man3/evp_pkey_export_to_provider.pod
levitte: Have a look at `doc/internal/man7/EVP_PKEY.pod`.  I'm adding a line here to refer to it.

doc/internal/man3/evp_pkey_export_to_provider.pod
levitte: Good point, especially since I use "provider side" elsewhere.  Thanks for the consistency heads up

doc/internal/man3/evp_pkey_export_to_provider.pod
levitte: Fixed

doc/internal/man3/evp_pkey_export_to_provider.pod
levitte: Fixed

doc/internal/man3/evp_pkey_export_to_provider.pod
levitte: Fixed

doc/internal/man7/EVP_PKEY.pod
levitte: Ah, my typical word drop (I seem to do that quite a bit...)

I've rephrased quite a bit, I hope it's become more understandable.

doc/internal/man7/EVP_PKEY.pod
levitte: Fixed

crypto/evp/p_lib.c

crypto/evp/p_lib.c
levitte: Good point

crypto/evp/p_lib.c

slontis:
travis errors are relevant..

levitte:
> travis errors are relevant..

Fixed

PR #11149, opened 2020-02-22
sec_mem: add note about the minimum size parameter.

paulidale:
Add a note indicating that the minimum size parameter to `CRYPTO_secure_malloc_init()` should be small.

### DIFF
--- a/crypto/mem_sec.c
+++ b/crypto/mem_sec.c
@@ -378,17 +378,33 @@ static int sh_init(size_t size, size_t minsize)
 
     memset(&sh, 0, sizeof(sh));
 
-    /* make sure size and minsize are powers of 2 */
+    /* make sure size is a powers of 2 */
     OPENSSL_assert(size > 0);
     OPENSSL_assert((size & (size - 1)) == 0);
-    OPENSSL_assert((minsize & (minsize - 1)) == 0);
     if (size == 0 || (size & (size - 1)) != 0)
         goto err;
-    if (minsize == 0 || (minsize & (minsize - 1)) != 0)
-        goto err;
 
-    while (minsize < sizeof(SH_LIST))
-        minsize *= 2;
+    if (minsize <= sizeof(SH_LIST)) {
+        OPENSSL_assert(sizeof(SH_LIST) <= 65536);
+        /*
+         * Compute the minimum possible allocation size.
+         * This must be a power of 2 and at least as large as the SH_LIST
+         * structure.
+         */
+        minsize = sizeof(SH_LIST) - 1;
+        minsize |= minsize >> 1;
+        minsize |= minsize >> 2;
+        if (sizeof(SH_LIST) > 16)
+            minsize |= minsize >> 4;
+        if (sizeof(SH_LIST) > 256)
+            minsize |= minsize >> 8;
+        minsize++;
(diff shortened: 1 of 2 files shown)


richsalz:
I think also saying it must be smaller than the first value.

paulidale:
@richsalz good suggestion.  Out of interest what values do you guys use?

I'm half tempted to ignore the value and use `sizeof(SH_LIST)` (rounded up to the nearest power of two) directly.

paulidale:
I've added a second commit for consideration.  This calculates the smallest legal `minsize` value, rather than relying on being passed something sane.

If this is acceptable, I'll revert the type change from #11003 and update the documentation.

richsalz:
I don't like enforcement of a policy for minsize because it really depends on what the app is doing.  Perhaps make "zero" be "please choose a default"?  Akamai proxy server uses 16 as minvalue.

paulidale:
Zero as default is reasonable.  The default will be twice the size of a pointer which will be 8 or 16 on most systems.

I still can't see a compelling reason for using anything bar the smallest value.  Fragmentation shouldn't be a problem, neither is the extra time cost.

richsalz:
If you have limited size for secure heap and you only want RSA/ECDSA keys and don't care about negotiated session keys.

paulidale:
I still don't see a reason :(

I'll just assume that there is one.

richsalz:
The secure heap is partitioned into "minsize" blocks, and record-keeping is done with standard heap structures.  If you have limited secure heap, you might as well make minsize by similar to your keysze.

paulidale:
Ah, the bitmap overhead.  That's reasonable.

paulidale:
Ping @openssl/committers for a review.

paulidale:
Rebased in the hopes that some of the build problems disappear.

paulidale:
Travis is the perpetual timing out build.  All good otherwise.

crypto/mem_sec.c
```diff
@@ -381,17 +381,33 @@ static int sh_init(size_t size, size_t minsize)
...
+        OPENSSL_assert(sizeof(SH_LIST) <= 65536);
+        /*
+         * Compute the minimum possible allocation size.
+         * This must be a power of 2 and at least as large as the SH_LIST
+         * structure.
+         */
+        minsize = sizeof(SH_LIST) - 1;
+        minsize |= minsize >> 1;
+        minsize |= minsize >> 2;
+        if (sizeof(SH_LIST) > 16)
```
vdukhovni: Actually, one thought.  Branches are much more expensive than a `shiff` and `or`.  Just remove the conditionals and always shift by 4 and 8.

paulidale: Any compiler that generates a branch here is really bad at optimising.  The expression is a constant.  The idea is to allow a compiler to completely remove the shift/or.  The structure contains two pointers, I very much doubt that this will be more than 16 bytes on any architecture we care about now or for a long time into the future.

vdukhovni: Fair enough, yes the condition is constant.

paulidale:
Merged to master after trivial rebase fix: `<=` changed to `==` and an assert relocated.

PR #11150, opened 2020-02-22
Build fix 20200222

levitte:
A couple of incorrect things went through in merges of diverse pieces of code.

URGENT!  We have builds failing because of this


levitte:
Turns out I overdid this PR, one fix cancelled out the other error.  It's now been reduced to something that should work

paulidale:
Travis timeout isn't relevant.
Merged to master.

levitte:
PING!

paulidale:
Ah, the PING! was deleted :)
I was feeling confused.

levitte:
Yeah, I got confused because the tab I had this in wasn't updated before I posted...

PR #11152, opened 2020-02-23
Remove double fetch of "OSSL_EX_DATA_GLOBAL" for global lock

davxy:
Fetch once and just pass the global ex_data to the "get_and_lock" static
function.

Removed a redundant null pointer check within the "get_and_lock" static
function (control already performed by the caller).

### DIFF
--- a/crypto/ex_data.c
+++ b/crypto/ex_data.c
@@ -24,19 +24,18 @@ int do_ex_data_init(OPENSSL_CTX *ctx)
 /*
  * Return the EX_CALLBACKS from the |ex_data| array that corresponds to
  * a given class.  On success, *holds the lock.*
+ * The |global| parameter is assumed to be non null (checked by the caller).
  */
-static EX_CALLBACKS *get_and_lock(OPENSSL_CTX *ctx, int class_index)
+static EX_CALLBACKS *get_and_lock(OSSL_EX_DATA_GLOBAL *global, int class_index)
 {
     EX_CALLBACKS *ip;
-    OSSL_EX_DATA_GLOBAL *global = NULL;
 
     if (class_index < 0 || class_index >= CRYPTO_EX_INDEX__COUNT) {
         CRYPTOerr(CRYPTO_F_GET_AND_LOCK, ERR_R_PASSED_INVALID_ARGUMENT);
         return NULL;
     }
 
-    global = openssl_ctx_get_ex_data_global(ctx);
-    if (global == NULL || global->ex_data_lock == NULL) {
+    if (global->ex_data_lock == NULL) {
         /*
          * If we get here, someone (who?) cleaned up the lock, so just
          * treat it as an error.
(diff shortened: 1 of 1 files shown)


crypto/ex_data.c
```diff
@@ -12,8 +12,9 @@
 
 int do_ex_data_init(OPENSSL_CTX *ctx)
 {
-    OSSL_EX_DATA_GLOBAL *global = openssl_ctx_get_ex_data_global(ctx);
+    OSSL_EX_DATA_GLOBAL *global;
```
richsalz: why did you split this into two lines?

crypto/ex_data.c
```diff
@@ -24,22 +25,21 @@ int do_ex_data_init(OPENSSL_CTX *ctx)
...
         return NULL;
     }
 
-    global = openssl_ctx_get_ex_data_global(ctx);
-    if (global == NULL || global->ex_data_lock == NULL) {
+    if (global->ex_data_lock == NULL) {
         /*
-         * If we get here, someone (who?) cleaned up the lock, so just
-         * treat it as an error.
+         * If we get here, someone cleaned up the lock, so just treat
```
richsalz: why did you remove the fact that we don't know who?  It's arguably a bug/TBD/TODO to find out.

crypto/ex_data.c
```diff
@@ -63,8 +63,9 @@ static void cleanup_cb(EX_CALLBACK *funcs)
 void crypto_cleanup_all_ex_data_int(OPENSSL_CTX *ctx)
 {
     int i;
-    OSSL_EX_DATA_GLOBAL *global = openssl_ctx_get_ex_data_global(ctx);
+    OSSL_EX_DATA_GLOBAL *global;
```
richsalz: again, why split?  please don't make needless changes if the code already follows the style.

crypto/ex_data.c
davxy: I believed it was obsolete. Do you suggest restoring it?

crypto/ex_data.c
davxy: I usually prefer to separate declaration from code involving function calls.

If the community prefer to adopt a single way to perform this, I suggest to put it in the coding guidelines to keep it consistent (sometimes is done in a way, sometimes in the other way round).

crypto/ex_data.c
davxy: See the comment above.

If you want I can put as was before, but (again). Put this in the coding guideline to keep it consistent.

davxy:
Looks like the CI appears to fail for reasons unrelated to the changes.

crypto/ex_data.c
davxy: restored the "inline" initialization

crypto/ex_data.c
richsalz: unless someone else on the project agrees with you, I think it should be restored.

davxy: Restored

slontis:
travis error is just a timeout that happens quite a bit

t8m:
The failure is just a timeout.

t8m:
Merged to master as 86cd42fbd3

PR #11158, opened 2020-02-24
EVP: Adapt EVP_PKEY checking, comparing and copying for provider keys, take 3

levitte:
This is an adaptation of the following functions for provided keys:

    EVP_PKEY_copy_parameters()
    EVP_PKEY_missing_parameters()
    EVP_PKEY_cmp_parameters()
    EVP_PKEY_cmp()

This replaces #11025 and is built on top of #11148

### DIFF
--- a/crypto/evp/keymgmt_meth.c
+++ b/crypto/evp/keymgmt_meth.c
@@ -335,3 +335,13 @@ const OSSL_PARAM *evp_keymgmt_export_types(const EVP_KEYMGMT *keymgmt,
         return NULL;
     return keymgmt->export_types(selection);
 }
+
+int evp_keymgmt_copy(const EVP_KEYMGMT *keymgmt,
+                     void *keydata_to, const void *keydata_from,
+                     int selection)
+{
+    /* We assume no copy if the implementation doesn't have a function */
+    if (keymgmt->copy == NULL)
+        return 0;
+    return keymgmt->copy(keydata_to, keydata_from, selection);
+}
--- a/crypto/evp/keymgmt_meth.c
+++ b/crypto/evp/keymgmt_meth.c
@@ -95,6 +95,10 @@ static void *keymgmt_from_dispatch(int name_id,
             if (keymgmt->validate == NULL)
                 keymgmt->validate = OSSL_get_OP_keymgmt_validate(fns);
             break;
+        case OSSL_FUNC_KEYMGMT_MATCH:
+            if (keymgmt->match == NULL)
+                keymgmt->match = OSSL_get_OP_keymgmt_match(fns);
+            break;
         case OSSL_FUNC_KEYMGMT_IMPORT:
             if (keymgmt->import == NULL) {
                 importfncnt++;
@@ -288,6 +292,16 @@ int evp_keymgmt_validate(const EVP_KEYMGMT *keymgmt, void *keydata,
     if (keymgmt->validate == NULL)
         return 1;
     return keymgmt->validate(keydata, selection);
+}
+
+int evp_keymgmt_match(const EVP_KEYMGMT *keymgmt,
+                      const void *keydata1, const void *keydata2,
+                      int selection)
+{
+    /* We assume no match if the implementation doesn't have a function */
+    if (keymgmt->match == NULL)
+        return 0;
+    return keymgmt->match(keydata1, keydata2, selection);
 }
 
 int evp_keymgmt_import(const EVP_KEYMGMT *keymgmt, void *keydata,
--- a/crypto/evp/p_lib.c
+++ b/crypto/evp/p_lib.c
@@ -124,30 +124,108 @@ int EVP_PKEY_missing_parameters(const EVP_PKEY *pkey)
     return 0;
 }
 
+/*
+ * This function is called for any mixture of keys except pure legacy pair.
+ * TODO When legacy keys are gone, we replace a call to this functions with
+ * a call to evp_keymgmt_util_match().
+ */
+static int evp_pkey_cmp_any(const EVP_PKEY *a, const EVP_PKEY *b,
+                            int selection)
+{
+    EVP_KEYMGMT *keymgmt1 = NULL, *keymgmt2 = NULL;
+    void *keydata1 = NULL, *keydata2 = NULL, *tmp_keydata = NULL;
+
+    /* If none of them are provided, this function shouldn't have been called */
+    if (!ossl_assert(a->keymgmt != NULL || b->keymgmt != NULL))
+        return -2;
+
+    /* For purely provided keys, we just call the keymgmt utility */
+    if (a->keymgmt != NULL && b->keymgmt != NULL)
+        return evp_keymgmt_util_match((EVP_PKEY *)a, (EVP_PKEY *)b, selection);
+
+    /*
+     * Here, we know that we have a mixture of legacy and provided keys.
+     * Try cross export and compare the resulting key data.
+     */
+    keymgmt1 = a->keymgmt;
+    keydata1 = a->keydata;
+    keymgmt2 = b->keymgmt;
+    keydata2 = b->keydata;
+
+    if ((keymgmt1 == NULL
+         && !EVP_KEYMGMT_is_a(keymgmt2, OBJ_nid2sn(a->type)))
+        || (keymgmt2 == NULL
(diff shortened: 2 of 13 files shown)


levitte:
@slontis, I know that I still need to add ec_match...

crypto/evp/p_lib.c
```diff
@@ -84,14 +84,34 @@ int EVP_PKEY_save_parameters(EVP_PKEY *pkey, int mode)
     return 0;
 }
 
+/*
+ * TODO(3.0) In case only one of the keys has a provider side "origin", the
+ * other one gets upgraded to have a provider side "origin" as well.  That
+ * means that it drops the pointers to any legacy "origin" is may have.  We
```
slontis: it may have.

crypto/evp/p_lib.c
```diff
@@ -84,14 +84,34 @@ int EVP_PKEY_save_parameters(EVP_PKEY *pkey, int mode)
...
+     * is gone.
+     */
+
+    /*
+     * Only check that type match this early when both keys are legacy.
+     * If either of them is provided, we let evp_keymgmt_util_copy()
+     * do this check, after having exported either of them that isn't
+     * provided.
+     */
+    if (!(to->keymgmt != NULL || from->keymgmt != NULL)) {
```
slontis: I think this is easier..
if ( x == NULL && y == NULL)  {

crypto/evp/p_lib.c
```diff
@@ -106,43 +126,163 @@ int EVP_PKEY_copy_parameters(EVP_PKEY *to, const EVP_PKEY *from)
...
+    /*
+     * If |from| is provided, we upgrade |to| to be provided as well.
+     * This drops the legacy key from |to|.
+     * evp_pkey_upgrade_to_provider() checks if |to| is already provided,
+     * we don't need to do that here.
+     */
+    if (from->keymgmt != NULL) {
+        EVP_KEYMGMT *tmp_keymgmt = from->keymgmt;
+
+        evp_pkey_upgrade_to_provider(to, NULL, &tmp_keymgmt, NULL);
```
slontis: Can this function not fail?

crypto/evp/p_lib.c
```diff
@@ -106,43 +126,163 @@ int EVP_PKEY_copy_parameters(EVP_PKEY *to, const EVP_PKEY *from)
...
+    }
+
+    /* For purely provided keys, we just call the keymgmt utility */
+    if (to->keymgmt != NULL && from->keymgmt != NULL)
+        return evp_keymgmt_util_copy(to, (EVP_PKEY *)from,
```
slontis: maybe one of the names for evp_keymgmt_util_copy / evp_keymgmt_copy should be updated to make it really clear what the difference is.

crypto/evp/p_lib.c
```diff
@@ -106,43 +126,163 @@ int EVP_PKEY_copy_parameters(EVP_PKEY *to, const EVP_PKEY *from)
...
+ * TODO When legacy keys are gone, we replace a call to this functions with
+ * a call to evp_keymgmt_util_match().
+ */
+static int evp_pkey_cmp_any(const EVP_PKEY *a, const EVP_PKEY *b,
+                            int selection)
+{
+    EVP_KEYMGMT *keymgmt1 = NULL, *keymgmt2 = NULL;
+    void *keydata1 = NULL, *keydata2 = NULL, *tmp_keydata = NULL;
+
+    /* If none of them are provided, this function should have been called */
```
slontis: should not ?

crypto/evp/p_lib.c
```diff
@@ -106,43 +126,163 @@ int EVP_PKEY_copy_parameters(EVP_PKEY *to, const EVP_PKEY *from)
...
+    if (!ossl_assert(a->keymgmt != NULL || b->keymgmt != NULL))
+        return -2;
+
+    /* For purely provided keys, we just call the keymgmt utility */
+    if (a->keymgmt != NULL && b->keymgmt != NULL)
+        return evp_keymgmt_util_match((EVP_PKEY *)a, (EVP_PKEY *)b, selection);
+
+    /*
+     * Here, we know that have a mixture of legacy and provided keys.
```
slontis: that we

crypto/evp/p_lib.c
levitte: Yes, but that shows in `to->keymgmt`, which remains NULL in that case.

crypto/evp/p_lib.c
levitte: That's what "util" is supposed to do.

slontis: Using the name 'util' doesnt make it obvious what the difference is. From the name I should get some indication of what the difference is.

crypto/evp/p_lib.c
slontis: I prefer not relying on side effects of function internals. The function could change over time (it appears to be changing quite a lot at the moment :)).

crypto/evp/p_lib.c
levitte: Suggest something other than "util" then

slontis: Describe to me what the difference is in 2 words or less :)

levitte: !util == wrapper

Did that help? :wink:

slontis: Not really - I give up ;)

crypto/evp/p_lib.c

crypto/evp/p_lib.c
levitte: Changed.
(there was good reason for this form, but that reason is gone now, so...)

crypto/evp/p_lib.c
levitte: Come to think of it, I misremembered this.  Check added.

crypto/evp/p_lib.c
levitte: What I meant to say is that `evp_keymgmt_util_` are helper / utility functions that deal with provider side keys only, while other `evp_keymgmt_` functions are wrappers for the keymgmt downcalls to the provider.

crypto/evp/p_lib.c
levitte: Yup

crypto/evp/p_lib.c

levitte:
ec_match was added a few days ago.

slontis:
travis failures again...

levitte:
> travis failures again...

Yup, that should fix itself as soon as #11148 is merged and this rebased on top of it.

levitte:
Now that #11148 is merged, I cleaned up the commits and rebased.  Let's see whaat the CIs say

crypto/evp/keymgmt_meth.c
```diff
@@ -321,3 +335,13 @@ const OSSL_PARAM *evp_keymgmt_export_types(const EVP_KEYMGMT *keymgmt,
...
     return keymgmt->export_types(selection);
 }
+
+int evp_keymgmt_copy(const EVP_KEYMGMT *keymgmt,
+                     void *keydata_to, const void *keydata_from,
+                     int selection)
+{
+    /* We assume no copy if the implementation doesn't have a function */
+    if (keymgmt->copy == NULL)
+        return 0;
```
romen: Why do we return 0 instead of -2 here?

levitte: -2 isn't a global value.  It isn't even consistent throughout all of EVP, there are some parts that return -1.  I did add a TODO in `keymgmt_lib.c`, in the middle of `evp_keymgmt_util_export_to_provider`, addressing exactly this question.  Looks like this:

``` C
    /* If the "origin" |keymgmt| doesn't support exporting, give up */
    /*
     * TODO(3.0) consider an evp_keymgmt_export() return value that indicates
     * that the method is unsupported.
     */
    if (pk->keymgmt->export == NULL)
        return NULL;
```

So *for the moment*, we're checking the function pointer before calling the `evp_keymgmt` wrapper function.  One day, we may get sick and tired of it and reconsider this question, but considering it's all internal functions, it doesn't have to be addressed today.

slontis:
Woohoo  (sense of relief that there is no take 4 that needs reviewing :) )

levitte:
> sense of relief that there is no take 4 that needs reviewing :)

Not to mention producing :wink:

PR #11166, opened 2020-02-25
WIP: cmdline app: add global `provider` command line option.

paulidale:
Add a global option framework.

Add a -provider option to allow providers to be loaded.  This option can be
specified multiple times.

Add a -provider_path option to allow the path to providers to be specified.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,10 @@
 
  Changes between 1.1.1 and 3.0.0 [xx XXX xxxx]
 
+  *) Add -provider and -provider_path options to the command line application.
+     The -provider option can be specified multiple times.
+     [Paul Dale]
+
   *) The test suite is changed to preserve results of each test recipe.
      A new directory test-runs/ with subdirectories named like the
      test recipes are created in the build tree for this purpose.
--- a/NEWS
+++ b/NEWS
@@ -23,6 +23,8 @@
       o Added 'openssl mac' that uses the EVP_MAC API.
       o Added 'openssl kdf' that uses the EVP_KDF API.
       o Add OPENSSL_info() and 'openssl info' to get built-in data.
+      o Add -provider and -provider_path options to the command line
+        applications.
       o Add support for enabling instrumentation through trace and debug
         output.
       o Changed our version number scheme and set the next major release to
--- a/apps/asn1pars.c
+++ b/apps/asn1pars.c
@@ -53,6 +53,8 @@ const OPTIONS asn1parse_options[] = {
     {"dump", OPT_DUMP, 0, "unknown data in hex form"},
     {"dlimit", OPT_DLIMIT, 'p',
      "dump the first arg bytes of unknown data in hex form"},
+
+    OPT_NO_GLOBAL_OPTS(),
     {NULL}
 };
 
--- a/apps/engine.c
+++ b/apps/engine.c
@@ -49,6 +49,8 @@ const OPTIONS engine_options[] = {
 
     OPT_PARAMETERS(),
     {"engine", 0, 0, "ID of engine(s) to load"},
+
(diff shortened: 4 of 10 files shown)


paulidale:
Currently missing are:
1. documentation
2. the -provider_path option needs to avoid `setenv(3)` and similar.
3. test cases

The documentation is going to touch virtually all the `man1/openssl-*` files.

richsalz:
> The documentation is going to touch virtually all the man1/openssl-* files.

If you bundle the provider apps into the "engine" variables, then the only change is in perlvars and openssl.pod

richsalz:
I don't like this.  (Yeah, I'll probably get overruled :)

Why can't the existing ```OPT_X_ENUM/CHOICES/etc``` mechanism be used?  Why the need to introduce something new?

paulidale:
The difference is that **all** applications need _provider_ support.  Only some of the applications have _engine_ support.  Okay, a couple of applications are good without the _provider_ option.

Adding a new `OPT_X_ENUM/CHOICES/etc` selection is an alternative possibility.  Is it preferable?  I don't know.

That's the point of this WIP PR, it's far from finished but feedback will help determine the path.

paulidale:
Compare with #11167 which is an approach similar to the existing shared options.

The third and least invasive option would be to add this to the shared engine options.  The negative here is that some applications require provider support but don't support engines.

levitte:
I don't see the value of having a `OPT_X_ENUM` that you *have* to spread all over all subcommands but a select few.

paulidale:
Is that a preference for this approach?

The global argument concept needs refinement.  At the moment it's providers but there will probably be other things that make sense globally.

levitte:
It is a preference for this solution over #11167, yeah. Even if this one has its own amount of hackiness...

paulidale:
I figured that this was the hacky alternative.

#11167 is pretty clean all things considered but it does require touching a lot of files.

I'm okay with either but I do think that application wide options will be beneficial in the future.  Things like `-in`, `-out`, `-help` and `-textout` are kind of crying out for such a treatment.  Also, the provider algorithm options: `-kdfopt`, `-macopt` and `-pkeyopt` coudl be squashed into a `-opt` and built into an OSSL_PARAM array automatically.

paulidale:
Regardless, I'll wait and see if anyone else expresses an opinion.  Currently: @richsalz -1, @levitte +1

levitte:
I don't think that we should globalise options that are only kinda sorta global. - in and - out are grey area at best, and there are other factors that may affect how they are treated. I suggest treading carefully.

paulidale:
I'll note that the proposed `-provider` option isn't global either.  There are a small number of apps where it doesn't make sense.

levitte:
It's global enough

richsalz:
Thanks for listening.  "global enough" is usually written as "common" :)

richsalz:
> I don't see the value of having a `OPT_X_ENUM` that you _have_ to spread all over all subcommands but a select few.

Do not think about the OpenSSL developers, but think about the orders-of-magnitude more developers who are looking at the code after 2-3 people have written it.

levitte:
I look at whoever writes a new subcommand and that must remember that oh right, have to add the global options...

richsalz:
Again: 2, 3, 4 people writing the app commands versus hundreds reading the code.  Wrong trade-off.

paulidale:
I'm thinking about using the #11167 approach.  It's consistent with what we've got currently which means it is not introducing any new complexity.

At some point we should reconsider the apps handling of command line options and clean things up.  I don't feel that 3.0 is a good place for this work.

I'll still wait for others input on direction for a while.

paulidale:
Closing this, going with the other.

PR #11160, opened 2020-02-24
Fix comment placement in ecp_nistp256.c

scottarc:
##### Checklist

- [ ] documentation is added or updated
- [ ] tests are added or updated

PR #11167, opened 2020-02-25
cmdline app: add provider commandline options.

paulidale:
Compare with #11160.

Add a -provider option to allow providers to be loaded. This option can be
specified multiple times.

Add a -provider_path option to allow the path to providers to be specified.

### DIFF
--- a/apps/crl.c
+++ b/apps/crl.c
@@ -24,7 +24,7 @@ typedef enum OPTION_choice {
     OPT_ISSUER, OPT_LASTUPDATE, OPT_NEXTUPDATE, OPT_FINGERPRINT,
     OPT_CRLNUMBER, OPT_BADSIG, OPT_GENDELTA, OPT_CAPATH, OPT_CAFILE, OPT_CASTORE,
     OPT_NOCAPATH, OPT_NOCAFILE, OPT_NOCASTORE, OPT_VERIFY, OPT_TEXT, OPT_HASH,
-    OPT_HASH_OLD, OPT_NOOUT, OPT_NAMEOPT, OPT_MD
+    OPT_HASH_OLD, OPT_NOOUT, OPT_NAMEOPT, OPT_MD, OPT_PROV_ENUM
 } OPTION_CHOICE;
 
 const OPTIONS crl_options[] = {
@@ -69,6 +69,7 @@ const OPTIONS crl_options[] = {
      "Do not load certificates from the default certificates directory"},
     {"no-CAstore", OPT_NOCASTORE, '-',
      "Do not load certificates from the default certificates store"},
+    OPT_PROV_OPTIONS,
     {NULL}
 };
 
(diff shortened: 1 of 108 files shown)


paulidale:
WIP because:

1. only two of nearly all apps have been updated.
1. no tests

t8m:
I like this approach more.

paulidale:
Travis isn't relevant, Windows is failing on `setenv` and `unsetenv` which need a better solution.

paulidale:
Moving this out of WIP.
Assuming the CI loops succeed.

I've deliberately not delved into the intricacies of the interactive command line application.

1. Should each entered command get its own library context or should one be shared?  If shared, unload needs to be supported.
1. Should provider options be accepted on the initial openssl command line (without a command name)?

richsalz:
> I've deliberately not delved into the intricacies of the interactive command line application.

Many things are not reset properly in "interactive mode" as I recall.  It would be nice to deprecate this.

paulidale:
ping.

paulidale:
Ping for review.

apps/crl.c
```diff
@@ -193,6 +194,10 @@ int crl_main(int argc, char **argv)
         case OPT_MD:
             if (!opt_md(opt_unknown(), &digest))
                 goto opthelp;
+        case OPT_PROV_CASES:
```
t8m: missing break; above.

apps/genpkey.c
```diff
@@ -131,6 +134,10 @@ int genpkey_main(int argc, char **argv)
                 BIO_printf(bio_err, "%s: cipher mode not supported\n", prog);
                 goto end;
             }
+        case OPT_PROV_CASES:
```
t8m: missing break

crypto/provider_core.c
```diff
@@ -384,6 +387,27 @@ int ossl_provider_add_parameter(OSSL_PROVIDER *prov,
...
 
+int OSSL_PROVIDER_set_default_search_path(OPENSSL_CTX *libctx, const char *path)
+{
+    struct provider_store_st *store;
+    char *p = NULL;
+
+    if (path != NULL) {
+        p = OPENSSL_strdup(path);
+        if (p == NULL)
+            return 0;
```
t8m: Should there be an error added?

doc/man1/openssl.pod
```diff
@@ -755,6 +755,26 @@ This file can be used in a subsequent command invocation.
 
 =back
 
+=head2 Provider Options
+
+All cryptographic operations in OpenSSL 3.0 are provider based.  These options
```
t8m: This is misleading. Even if we ignore legacy codepaths in EVP that will be removed before release, there are the engines and the legacy APIs.

doc/man3/OSSL_PROVIDER.pod
```diff
@@ -39,6 +43,11 @@ L<OPENSSL_CTX(3)> for further details.
 
 =head2 Functions
 
+OSSL_PROVIDER_set_default_search_path() specifies the default search B<path>
+that is to be used for looking for providers in the specified B<libctx>.
+If left unspecified, an environment variable and a fall back default value will
```
t8m: fall-back?

paulidale:
Rebased to address merge conflict (renaming of .conf to .cnf in documentation).

paulidale:
Merged to master.  Thanks.

Travis failure was the usual timeout.

PR #11170, opened 2020-02-25
Configuration: Add post module building check

### LINKED ISSUE #11020: fips.so is not self-contained
When we build fips.so we do so with the "-z defs" linker option (on Linux):

https://github.com/openssl/openssl/blob/d8d4e5fb32b3f7d9af28d21ce4c1c46cc1c7614c/Configurations/shared-info.pl#L31-L41

This is supposed to ensure that all symbols are resolved. Without this the link completes successfully even though some symbols are not defined in the resulting .so file. If that happens then I *think* those symbols get resolved from libcrypto at load time...which obviously would means that fips.so is not self-contained.

This used to work just fine. But it no longer seems to be the case. For example the symbol `_bignum_modp_1536_p` is defined here:

https://github.com/openssl/openssl/blob/d8d4e5fb32b3f7d9af28d21ce4c1c46cc1c7614c/crypto/bn/bn_dh.c#L630-L632

Since this is guarded with `ifndef FIPS_MODE` the symbol *only* exists in libcrypto and not in fips.so. The symbol is referenced from bn_const.c here:

https://github.com/openssl/openssl/blob/d8d4e5fb32b3f7d9af28d21ce4c1c46cc1c7614c/crypto/bn/bn_const.c#L85-L88

This code does get included in fips.so and the reference to `_bignum_modp_1536_p` is not guarded with `FIPS_MODE` guards like it should be. This should have resulted in a link failure - but it hasn't.

levitte:
We want an extra check that the FIPS provider module has resolved all
symbols except the standard library ones.  This is generalized with
the addition of a 'dso_post_cmd' config target attribute, that holds
the command line to do the check.

For the moment, we only do this for module building with the
'linux-shared' shared library scheme and its derivatives.

Fixes #11020

### DIFF
--- a/Configurations/shared-info.pl
+++ b/Configurations/shared-info.pl
@@ -29,14 +29,19 @@ my %shared_info;
         shared_sonameflag     => '-Wl,-soname=',
     },
     'linux-shared' => sub {
+        my $zdefs =
+            (grep /(?:^|\s)-fsanitize/,
+             @{$config{CFLAGS}}, @{$config{cflags}})
+            ? ''
+            : '-z defs';
         return {
             %{$shared_info{'gnu-shared'}},
             shared_defflag    => '-Wl,--version-script=',
-            dso_ldflags       =>
-                (grep /(?:^|\s)-fsanitize/,
-                 @{$config{CFLAGS}}, @{$config{cflags}})
-                ? ''
-                : '-z defs',
+            dso_ldflags       => $zdefs,
+            dso_post_cmd      =>
+                $zdefs
+                ? 'test `nm -u $@ | grep -v @@GLIBC | grep -v " _" | wc -l` = 0'
+                : undef,
         };
     },
     'bsd-gcc-shared' => sub { return $shared_info{'linux-shared'}; },
--- a/Configurations/unix-Makefile.tmpl
+++ b/Configurations/unix-Makefile.tmpl
@@ -1455,13 +1455,17 @@ EOF
                       fill_lines(' ', $COLUMNS - length($dso) - 2,
                                  @objs, @defs, @deps));
 
-      return <<"EOF";
+      my $out = <<"EOF";
 $dso: $deps
 	\$(CC) \$(DSO_CFLAGS) $linkflags\$(DSO_LDFLAGS) \\
 		-o $dso$shared_def \\
 		$objs \\
 		$linklibs\$(DSO_EX_LIBS)
 EOF
+      if (defined $target{dso_post_cmd}) {
+          $out .= "\t" . $target{dso_post_cmd} . "\n";
+      }
+      return $out;
   }
   sub obj2lib {
       my %args = @_;


slontis:
It might need to be a bit more specific about what target names it supports.. Not sure if this could be configured in a list somehow..

levitte:
Looking at https://travis-ci.org/openssl/openssl/jobs/654820137, it's clear this method doesn't work.  `nm` doesn't, of course, load all linked shared libraries (such as libcrypto), so it wrongly reports some of our engines, such as engines/afalg.so:

``` console
$ nm -u engines/afalg.so | grep -v '@@GLIBC'
                 U ASYNC_get_current_job@@OPENSSL_3.0.0
                 U ASYNC_get_wait_ctx@@OPENSSL_3.0.0
                 U ASYNC_pause_job@@OPENSSL_3.0.0
                 U ASYNC_WAIT_CTX_get_fd@@OPENSSL_3.0.0
                 U ASYNC_WAIT_CTX_set_wait_fd@@OPENSSL_3.0.0
                 U CRYPTO_set_mem_functions@@OPENSSL_3.0.0
                 U ENGINE_get_static_state@@OPENSSL_3.0.0
                 U ENGINE_set_ciphers@@OPENSSL_3.0.0
                 U ENGINE_set_destroy_function@@OPENSSL_3.0.0
                 U ENGINE_set_finish_function@@OPENSSL_3.0.0
                 U ENGINE_set_id@@OPENSSL_3.0.0
                 U ENGINE_set_init_function@@OPENSSL_3.0.0
                 U ENGINE_set_name@@OPENSSL_3.0.0
                 U ERR_get_next_error_library@@OPENSSL_3.0.0
                 U ERR_load_strings@@OPENSSL_3.0.0
                 U ERR_new@@OPENSSL_3.0.0
                 U ERR_set_debug@@OPENSSL_3.0.0
                 U ERR_set_error@@OPENSSL_3.0.0
                 U ERR_unload_strings@@OPENSSL_3.0.0
                 U EVP_CIPHER_CTX_cipher@@OPENSSL_3.0.0
                 U EVP_CIPHER_CTX_encrypting@@OPENSSL_3.0.0
                 U EVP_CIPHER_CTX_get_cipher_data@@OPENSSL_3.0.0
                 U EVP_CIPHER_CTX_iv_length@@OPENSSL_3.0.0
                 U EVP_CIPHER_CTX_iv_noconst@@OPENSSL_3.0.0
                 U EVP_CIPHER_CTX_iv@@OPENSSL_3.0.0
                 U EVP_CIPHER_CTX_key_length@@OPENSSL_3.0.0
                 U EVP_CIPHER_CTX_nid@@OPENSSL_3.0.0
                 U EVP_CIPHER_meth_free@@OPENSSL_3.0.0
                 U EVP_CIPHER_meth_new@@OPENSSL_3.0.0
                 U EVP_CIPHER_meth_set_cleanup@@OPENSSL_3.0.0
                 U EVP_CIPHER_meth_set_do_cipher@@OPENSSL_3.0.0
                 U EVP_CIPHER_meth_set_flags@@OPENSSL_3.0.0
                 U EVP_CIPHER_meth_set_impl_ctx_size@@OPENSSL_3.0.0
                 U EVP_CIPHER_meth_set_init@@OPENSSL_3.0.0
                 U EVP_CIPHER_meth_set_iv_length@@OPENSSL_3.0.0
                 w __gmon_start__
                 w _ITM_deregisterTMCloneTable
                 w _ITM_registerTMCloneTable
                 U OPENSSL_strlcpy@@OPENSSL_3.0.0
```

Back to the drawing table.

levitte:
(and closing this PR)

PR #11176, opened 2020-02-25
Cnf over conf

richsalz:
I know foo.conf is what we seem to prefer.  But foo.cnf is what the docs talk about, and openssl.cnf is the default.  I think it's less confusing to use .cnf as the extension for OpenSSL config files, so do the big renaming.


levitte:
It's a bit interesting to see how we struggle with these sort of things. After unabbreviating lcl -> locl -> local, we now go the other way in another part.

In my view, cnf vs conf is like htm vs html.

t8m:
There are conflicts, please rebase.

richsalz:
pesky makdown :)  rebased and pushed.

mspncp:
> It's a bit interesting to see how we struggle with these sort of things. After unabbreviating lcl -> locl -> local, we now go the other way in another part.
> 
> In my view, cnf vs conf is like htm vs html.

FWIW: If it weren't for `openssl.cnf`, I would have preferred the longer file extension. But in this case it make sense to go for the "more popular" file extension.

richsalz:
Don't know why the "ready to merge" robot didn't catch this, @iamamoose but this seems like 24 hours are done.

mspncp:
> Don't know why the "ready to merge" robot didn't catch this, @iamamoose but this seems like 24 hours are done.

IIUC, the bot refrains from setting the "ready to merge" state if someone posts additional comments after the approval. In this case it was probably [my post](https://github.com/openssl/openssl/pull/11176#issuecomment-594022261) which prevent it. The rationale behind it is that a comment after final approval might be an indication of an ongoing discussion.

t8m:
Pushed to master.

richsalz:
Thanks for the discusion and review!

iamamoose:
> IIUC, the bot refrains from setting the "ready to merge" state if someone posts additional comments after the approval. In this case it was probably [my post](https://github.com/openssl/openssl/pull/11176#issuecomment-594022261) which prevent it. The rationale behind it is that a comment after final approval might be an indication of an ongoing discussion.

Exactly that.  I don't want it to start second guessing or being too clever -- if there's any change to a PR since the Done label was set (such as a comment), then it assumes it needs a human decision.

richsalz:
"resolved - working as designed" :)  makes sense to me, just thought it was worth asking.

mspncp:
But OTOH we loose the comfort of being reminded. Let's assume @iamamoose that you change the bot to neglect commits after 'review done', then we still would have two options to prevent the automatism:

- Resetting the label from 'review done' to '* review pending' before the bot triggers
- Resetting the label from 'ready to merge' to a previous state after the bot triggers

Maybe that alternative would be more convenient?

richsalz:
Another possibility is to just change the message to "This appears to be ready to merge (the hold period has expired)"

mspncp:
I like that suggestion, but I would drop the part in the parentheses:

> This pull request appears to be ready to merge

@iamamoose what do you think about ignoring additional posts and triggering the timeout unconditionally, with a slightly modified (weaker) message?

iamamoose:
I'm not a fan of labels that can mean two things; it could mean "this is ready do merge" or "this is probably ready to merge once you've read all the recent changes and they're harmless".  

Two possibilities for "if there were changes made since the approval:done flag was set":

1. We could create a new label "ready to merge after review" and have the bot move those things to that label

2.  It could add a comment "24 hours has now elapsed since Approval:Done but since there were changes made it has not been automatically moved to Approval:Ready to Merge" and not do anything with labels.  That would help ping people (and stop them having to look to see if 24 hours is up) who could then manually set the label (or not if the comments required action first).

PR #11177, opened 2020-02-25
Update some FIPS module things

richsalz:
- The generated config for the module signature is "fipsmodule.conf",
since it's about the module.  This is a more sensible name for the user community.
- Add -q option to fipsinstall command, to stop chatty verbose status
messages. This is normally done at install/boot time, and we don't want extra status/noise messages.
- Document env var OPENSSL_CONF_INCLUDE. This is just a doc bug.

### DIFF
--- a/apps/fipsinstall.c
+++ b/apps/fipsinstall.c
@@ -31,12 +31,13 @@ static OSSL_CALLBACK self_test_events;
 static char *self_test_corrupt_desc = NULL;
 static char *self_test_corrupt_type = NULL;
 static int self_test_log = 1;
+static int quiet = 0;
 
 typedef enum OPTION_choice {
     OPT_ERR = -1, OPT_EOF = 0, OPT_HELP,
     OPT_IN, OPT_OUT, OPT_MODULE,
     OPT_PROV_NAME, OPT_SECTION_NAME, OPT_MAC_NAME, OPT_MACOPT, OPT_VERIFY,
-    OPT_NO_LOG, OPT_CORRUPT_DESC, OPT_CORRUPT_TYPE
+    OPT_NO_LOG, OPT_CORRUPT_DESC, OPT_CORRUPT_TYPE, OPT_QUIET
 } OPTION_CHOICE;
 
 const OPTIONS fipsinstall_options[] = {
@@ -60,6 +61,7 @@ const OPTIONS fipsinstall_options[] = {
     {"noout", OPT_NO_LOG, '-', "Disable logging of self test events"},
     {"corrupt_desc", OPT_CORRUPT_DESC, 's', "Corrupt a self test by description"},
     {"corrupt_type", OPT_CORRUPT_TYPE, 's', "Corrupt a self test by type"},
+    {"quiet", OPT_QUIET, '-', "No messages, just exit status"},
     {NULL}
 };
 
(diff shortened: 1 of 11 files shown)


richsalz:
ping @slontis for feedback.

mspncp:
Similar as in #11176 one can ask: why is  the OpenSSL configuration file called "openssl.cnf", but the fips configuration file "fipsmodule.conf"? Why not use the same file extension throughout?

slontis:
Looks ok to me..
I would argue that the -q option may not be a thing you want to turn off at installation time.
If paths are incorrect or the integrity of the module has been modified then it may be beneficial to know the reason for the failure. It should be noted however that it is highly unlikely that the actual self test KATS should ever fail.

richsalz:
There's a separate PR to rename *.conf to *.cnf (#11176 )

The output suppressed by -q is nothing useful, it just says pass/verified/failure.  When installing the same module on a few hundred thousand machines it's not needed.

apps/fipsinstall.c
```diff
@@ -424,12 +430,13 @@ int fipsinstall_main(int argc, char **argv)
...
             goto end;
-        BIO_printf(bio_out, "INSTALL PASSED\n");
+        if (!quiet)
+            BIO_printf(bio_out, "INSTALL PASSED\n");
     }
 
     ret = 0;
 end:
-    if (ret == 1) {
+    if (ret == 1 && !quiet) {
```
levitte: Why do you want to suppress error output?

richsalz: Most users will probably not use the -q flag. For me, the exit status suffices.

levitte: Yeah, but what we add in our source isn't just for you. What other command do you know that suppresses *error message output* with -q? `grep` doesn't (try `grep -q foo /`)

richsalz: We don't write it just for me?  :) Shrug.  Modified commit pushed so that only the redundant "VERIFY FAILED" is suppressed with -q.  I looked, and everything that does "goto err" prints out a diagnostic first.  Unless you get the options wrong:
```
; ./apps/openssl fipsinstall -h
fipsinstall: Unknown option: -h
fipsinstall: Use -help for summary.
INSTALL FAILED
        exit 1
; 
```

slontis: You must not of pushed the change yet?

richsalz:
Rebased; only change was to propagate .conf->.cnf changes.

richsalz:
ping for another review.

t8m:
The commit message seems a little bit obsolete, can you please amend it to reflect the actual patch?

richsalz:
Reworded the commit message, did I get your concern addressed?

t8m:
> Reworded the commit message, did I get your concern addressed?

Unfortunately no. I do not see any added documentation of OPENSSL_CONF_INCLUDE env var. And the configuration file is named fipsmodule.cnf not .conf.

richsalz:
I'm not going to touch this to avoid resetting the timers, and I assume you'll fix the commit message, @t8m

t8m:
@slontis I've forgot to ask for your reapproval. Does your approval still hold?

richsalz:
I rebased this, to capture some additional fipsinstall.cnf -> fipsmodule.cnf renamings that had to be done.

Fresh approvals needed. :(

t8m:
Merged to master. Thank you.

PR #11181, opened 2020-02-26
Avoid arm64 builds timeout due to silent make taking too long time

t8m:
The arm64 builds timeout because the silent make takes too long time before
it outputs anything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,6 @@ before_install:
 
 arch:
     - amd64
-    - arm64
 
 os:
     - linux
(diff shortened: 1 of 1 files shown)


t8m:
I have also reused one of the arm64 builds as a no-deprecated build because there was no no-deprecated build running.
I've also included a single ppc64le-build - there was none before.

t8m:
As this is a travis build environment-only change, I've added the urgent label.

.travis.yml
```diff
@@ -154,15 +160,6 @@ matrix:
...
           compiler: gcc
-        - arch: arm64
-          os: osx
-        # arm64 times out when using -fsanitize=address
-        # The amd64 build should be sufficient to test this.
-        # Note: the env line must exactly match the env line from the build
-        # matrix above which contains the `-fsanitize=address` option.
-        # TODO: come up with a better way of doing this that is less error prone.
-        - arch: arm64
-          env: CONFIG_OPTS="no-asm -Werror --debug no-afalgeng no-shared enable-rc5 enable-md2 -fsanitize=address" LSAN_OPTIONS="report_objects=1"
```
paulidale: This means we're losing the asan ARM build doesn't it?
Is there a possibility of keeping it by fiddling without options?  e.g. ditch the `no-asm` or `--debug` perhaps?

t8m: This build was already excluded. Is there any reason why this particular build would be so much useful when we have asan build for x86_64?

paulidale: With `no-asm` I guess they'll be pretty much the same.....

I err towards more testing rather than less.

t8m:
Merged to master as 7d72644025a

Issue #11165, opened 2020-02-24
OSSL_PARAM_REAL type


kroeckx:
The documentation says:
```
       OSSL_PARAM_REAL
           The parameter data is a floating point value in native form.
```

It seems that the code has switches all over checking sizeof(double). I think double is the only thing we support. It's probably more clear that we say double there instead of "floating point".

levitte:
Why? Both `float` and `double` are floating point types, just of different sizes.

kroeckx:
But we don't actually support either float or long double.

levitte:
So?

kroeckx:
If we have no intention to support anything other that double, I
think it should be clear that when we say OSSL_PARAM_REAL that
it's a double, and not any of the other floating point types. Or
everywhere that OSSL_PARAM_REAL is used, it should mention
only double is supported.

levitte:
Key word here is *currently*. I see no reason to lock that data type to one specific C type.

And the point is that any size is perfectly possible for the OSSL_PARAM structure itself, even though the OSSL_PARAM *functions* currently only support double.

paulidale:
I had `float` in the parameter code originally but it was decided to defer this until later.  Putting it back would be straightforward.

I'm less sure about `long double` though.  I have seen compilers where command line options change the size of this one which could cause problems with providers built differently to libcrypto.  A can of worms best left closed until we need it.

levitte:
Right. And we can with no changes to the OSSL_PARAM data types.

paulidale:
#11183 ??

PR #11183, opened 2020-02-26
Params: add float type parameters.

paulidale:
This adds single precision floating point parameters back again.

Inexact conversions from double to float are reported as failures, although they kind of make sense being successes since real values are almost always inexact.

Fixes: #11165

### DIFF
--- a/crypto/params.c
+++ b/crypto/params.c
@@ -13,6 +13,14 @@
 #include "internal/thread_once.h"
 #include "internal/numbers.h"
 
+/*
+ * Helper macros that know the number of mantassia bits in floats and doubles.
+ * Doubles have 53 bit mantissas (52 stored, 1 implicit).
+ * Floats have 24 bit mantissas (23 stored, 1 implicit).
+ */
+#define UINT_FITS_DOUBLE(u)  ((u >> 53) == 0)
+#define UINT_FITS_FLOAT(u)  ((u >> 24) == 0)
+
 OSSL_PARAM *OSSL_PARAM_locate(OSSL_PARAM *p, const char *key)
 {
     if (p != NULL && key != NULL)
@@ -156,6 +164,7 @@ int OSSL_PARAM_get_int32(const OSSL_PARAM *p, int32_t *val)
     uint32_t u32;
     uint64_t u64;
     double d;
+    float f;
 
     if (val == NULL || p == NULL )
         return 0;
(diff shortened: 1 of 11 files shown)


paulidale:
Travis is a time out during build and not relevant.

crypto/params.c
```diff
@@ -13,6 +13,14 @@
 #include "internal/thread_once.h"
 #include "internal/numbers.h"
 
+/*
+ * Helper macros that know the number of mantassia bits in floats and doubles.
+ * Doubles have 53 bit mantissas (52 stored, 1 implicit).
+ * Floats have 24 bit mantissas (23 stored, 1 implicit).
```
levitte: Are you sure this is universally true?

richsalz:
Why is this needed?  Pre-ANSI C always elevated float to double in params and math.  And what's the use-case for floating point in crypto operations?

crypto/params.c
paulidale: They should be on the conservative side.  The popular machines I remember where it wasn't were some IBM mainframes and the VAX.  The VAX has the F format which does have 23 mantissa bits but the D format has 56 rather than 53.  IBM used 24 and 55 bits and rounded in hexadecimal rather than binary :weary:.  Early Microsoft BASICs used 23 and 55 bits for its 32 and 64 bit formats.  Of these, I think that the only one that is currently supported by OpenSSL is the S390 by IBM.

IEEE 754 caught on very rapidly and very widely.  It was a really good standard.

There have been a [myriad of oddities](http://www.quadibloc.com/comp/cp0201.htm).  Microsoft BASIC has a 40 bit format, control data used 60 bit reals, Cray has a 64 bit real, etc.

paulidale: I think I can come up with an improved alternative.

levitte:
> And what's the use-case for floating point in crypto operations?

Could be my fault, I had the vague impression that we used doubles somewhere...  I just grepped for both float and double, and it turns out to be very little, and certainly nowhere near anything we want to transfer between libcrypto and providers.

With that in mind, I would opt for removing `OSSL_PARAM_REAL` and associated support functions.

paulidale:
Entropy / randomness will use them.

levitte:
> Entropy / randomness will use them.

That seems to be discouraged.

https://github.com/openssl/openssl/blob/30a4cda5e0c67b4e77da4f21b7c5f27991d3367a/crypto/rand/drbg_lib.c#L1290-L1322

Also, we're expressing it in a different manner to the entropy gatherers, as the denominator in 1/n, where n == "average number of data bits to get one bit of entropy", isn't that a reusable thing?

richsalz:
> Entropy / randomness will use them.

s/will/maybe might possibly in some vendors/ and I believe this. And if you thought OpenSSL needed it for it's DRBG implementation, this would have come up sooner.  And finally, double is good enough, right?

paulidale:
The integration of RNGs into providers hasn't happened yet...
We do need the ability to generate entropy in one provider for seeding another.

I'll close this since float support isn't required in addition to doubles.  Either one should be enough.

PR #11184, opened 2020-02-26
Run apps/progs.pl from the build tree

romen:
`apps/progs.pl` is optimized to be run from the build tree, but `make update` currently runs it from the src tree.

This can lead to failures in out-of-tree builds if the build tree is nested deeper than the src tree compared to the common ancestor node.

This commit fixes the Makefile template to run `apps/progs.pl` from the build tree instead.

### DIFF
--- a/Configurations/unix-Makefile.tmpl
+++ b/Configurations/unix-Makefile.tmpl
@@ -965,10 +965,10 @@ generate_apps:
 	( cd $(SRCDIR); $(PERL) VMS/VMSify-conf.pl \
 				< apps/openssl.cnf > apps/openssl-vms.cnf )
 	@ : {- output_off() if $disabled{apps}; "" -}
-	( b=`pwd`; cd $(SRCDIR); \
-	  $(PERL) -I$$b apps/progs.pl -H $(APPS_OPENSSL) > apps/progs.h )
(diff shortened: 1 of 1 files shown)


romen:
I am marking as `urgent` as my default build toolchain is now testing also out-of-tree builds and this can trigger failures, so I believe it fits the criteria of breing a fix for an issue breaking `master` builds.

Ping @levitte to evaluate if this fix matches what he has in mind and to evaluate if he agrees with the `urgent` label.

romen:
None of our CIs detected this failure, given that some scripts are adapted to be run from the src tree and some are adapted to be run from the build tree, should we have some build configuration target in which the out-of-tree builds are tested with different relative nesting depths for the src and build trees to verify that this kind of problem does not stay hidden in the future?

levitte:
Hahaha, I took this a step further, see #11185

levitte:
Good idea regarding the nesting levels.  That's a simple change in .travis.yml

levitte:
For a Travis fix, see #11186

romen:
Mine was just a local hack, while #11185 seems a more well-thought solution (even though I don't fully understand the changes): I am closing this in favor of #11185

PR #11185, opened 2020-02-26
build.info extensions: variable value substitutions and multi-item statements

levitte:
This started with a bug notification I got, that `apps/progs.pl` would fail at times.  It turns out that it should be run from the build tree, not from the source tree.

This also touched an older thought, that some files (the majority of them) that we currently generate with `make update` and thereby have versioned don't actually need it.  They can as well be generated in build time.

This PR changes things so `apps/progs.c` and `apps/progs.h` are generated at build time, and adds the necessary expansions of the build.info syntax to allow this to be done smoothly enough.

PR #11186, opened 2020-02-26
.travis.yml: where it matters, have build and source nesting levels differ

levitte:
Where we build out of source, the source directory was _srcdist and
the build directory was _build.  That gives the same nesting level for
both, which doesn't quite exercise all aspects of relative back
references from build to source tree.

Changing the build tree to be in _build/tree will challenge back
references a bit more, and ensure a bit more that we got it right.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -170,10 +170,10 @@ before_script:
     - if [ -n "$DESTDIR" ]; then
           sh .travis-create-release.sh $TRAVIS_OS_NAME;
           tar -xzf _srcdist.tar.gz;
-          mkdir _build;
-          cd _build;
-          srcdir=../_srcdist;
-          top=..;
+          mkdir -p _build/tree;
+          cd _build/tree;
+          srcdir=../../_srcdist;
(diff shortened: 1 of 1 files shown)


levitte:
I expect this to *fail* until #11184 or #11185 have been merged

romen:
> I expect this to _fail_ until #11184 or #11185 have been merged

This seems to be failing for the _wrong_ reason now for some reason!

I see travis reporting "No rule for target update" as if the Makefile was not correctly generated!

levitte:
Ah, I missed one spot.  Fixed

t8m:
You could drop the NOUPDATE variable support from .travis.yml that I've added in #11181 to workaround the progs update with no-deprecated.

t8m:
Ah, my last comment is on wrong PR.

levitte:
Rebased, and now it should work out fine.

levitte:
Ping!  CIs are fine, the Travis failure is a timeout

PR #11193, opened 2020-02-27
EVP: Check that key methods aren't foreign when exporting

### LINKED ISSUE #9915: dsa signature failure with external keys / master 2019-09-16
Issue is from e_nss (engine) regression tests - with current(2019-09-16) master "sign-dsa-dss1.test" fail.
Output file sign-dsa-sha1.out after (4) shows:
```
=== sign ===
engine "e_nss" set.
Error Signing Data
00:A7:5E:3F:7B:7F:00:00:error:80000085:E_NSS::Missing private key:/work/src/MyProjects/e_nss/anon-access/origin-local/engines/e_nss_dsa.c:42:
=== verify ===
Error reading signature file cfg-nss-dsa-sha1.sign
```

Commit history:
(4) Add support for verify/verify_recover functions to EVP_SIGNATURE:
commit 390acbebfa90500c79c5014e6659eacda861550c
(3) Implement DSA in the default provider:
commit 4889dadcb8511176c30888c748f1981adc38451d
(2) Add the ability to perform signatures in a provider
commit dfcb5d29b525f5d2b6bd80602dca5efe5fca77bb
(1) DH_check_pub_key_ex was accidentally calling DH_check, so results were undefined.
commit 2b95e8efcf8b99892106070d9ac745a0a369f503

(1) is lest know **good** version , (4) is **broken** version. With (2) or (3) as usually in "group" commits build fail.

### LINKED ISSUE #11179: Current master breaks libp11 RSA signature
MacOS Catalina 10.15.3, Xcode-11.3.1, installed OpenSSL-1.1.1d (reference) and (in a separate dir tree) OpenSSL-3.0 master.

Using `libp11` to perform RSA signature using hardware token accessible via PKCS#11. Fails with both a real hardware token, and SoftHSMv2.

With OpenSSL-1.1.1d (succeeds):
```
$ pkcs11-rsa-pss-sign-demo2
Generating ephemeral file /tmp/derive.89314.text to test RSA-PSS signature...

openssl rand -hex -out /tmp/derive.89314.text 5120

Signing file /tmp/derive.89314.text...
openssl dgst -engine pkcs11 -keyform engine -sign "pkcs11:manufacturer=piv_II;object=SIGN%20key;type=private" -sha384 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -out /tmp/derive.89314.text.sig /tmp/derive.89314.text
engine "pkcs11" set.
Enter PKCS#11 token PIN for xxxxxxxxx:
Signature is stored in /tmp/derive.89314.text.sig

Verifying signature:
openssl dgst -engine pkcs11 -keyform engine -sha384 -verify "pkcs11:manufacturer=piv_II;object=SIGN%20pubkey;type=public" -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -signature /tmp/derive.89314.text.sig  /tmp/derive.89314.text
engine "pkcs11" set.
Verified OK

$
```

With OpenSSL-3.0 (current master, fails):
```
$ pkcs11-rsa-pss-sign-demo3
Generating ephemeral file /tmp/derive.89324.text to test RSA-PSS signature...

openssl3 rand -hex -out /tmp/derive.89324.text 5120

Signing file /tmp/derive.89324.text...
openssl3 dgst -engine pkcs11 -keyform engine -sign "pkcs11:manufacturer=piv_II;object=SIGN%20key;type=private" -sha384 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -out /tmp/derive.89324.text.sig /tmp/derive.89324.text
engine "pkcs11" set.
Enter PKCS#11 token PIN for xxxxxxxxx:
Error Signing Data
C0:2D:B6:09:01:00:00:00:error:rsa routines:rsa_ossl_private_encrypt:missing private key:crypto/rsa/rsa_ossl.c:333:
C0:2D:B6:09:01:00:00:00:error:Provider routines:rsa_sign:RSA lib:providers/implementations/signature/rsa.c:410:

```

Also, see this `libp11` issue for details of a failure trying to sign using SoftHSMv2: https://github.com/OpenSC/libp11/issues/331

levitte:
The EVP_PKEY_ASN1_METHOD function export_to() must check that the key
we're trying to export has a known libcrypto method, i.e. is a built
in RSA_METHOD, DSA_METHOD, etc.  Otherwise, the method may be defined
by the calling application, by an engine, by another library, and we
simply cannot know all the quirks hidden behind that method, if we
have access to the key data, or much anything.

Such keys are simply deemed impossible to export to provider keys,
i.e. have export_to() return 0.  This cascades back to functions like
evp_pkey_export_to_provider() and evp_pkey_upgrade_to_provider() and
their callers.  In most cases, this is fine, but if these get mixed in
with provider side keys in any function, that function will fail.

Fixes #11179
Fixes #9915

### DIFF
--- a/crypto/ec/ec_ameth.c
+++ b/crypto/ec/ec_ameth.c
@@ -637,6 +637,13 @@ int ec_pkey_export_to(const EVP_PKEY *from, void *to_keydata,
             || (ecg = EC_KEY_get0_group(eckey)) == NULL)
         return 0;
 
+    /*
+     * If the EC_KEY method is foreign, then we can't be sure of anything,
+     * and can therefore not export or pretend to export.
+     */
+    if (EC_KEY_get_method(eckey) != EC_KEY_OpenSSL())
+        return 0;
+
     ossl_param_bld_init(&tmpl);
 
     /* export the domain parameters */
--- a/crypto/evp/pmeth_fn.c
+++ b/crypto/evp/pmeth_fn.c
@@ -38,7 +38,7 @@ static int evp_pkey_asym_cipher_init(EVP_PKEY_CTX *ctx, int operation)
      */
     ERR_set_mark();
 
-    if (ctx->keytype == NULL || ctx->engine != NULL)
+    if (ctx->engine != NULL || ctx->keytype == NULL)
         goto legacy;
 
     /*
--- a/crypto/rsa/rsa_ameth.c
+++ b/crypto/rsa/rsa_ameth.c
@@ -1092,6 +1092,13 @@ static int rsa_pkey_export_to(const EVP_PKEY *from, void *to_keydata,
     OSSL_PARAM *params = NULL;
     int rv = 0;
 
+    /*
+     * If the RSA method is foreign, then we can't be sure of anything, and
+     * can therefore not export or pretend to export.
+     */
+    if (RSA_get_method(rsa) != RSA_PKCS1_OpenSSL())
+        return 0;
+
     /* Public parameters must always be present */
     if (n == NULL || e == NULL)
         goto err;
--- a/crypto/dh/dh_ameth.c
+++ b/crypto/dh/dh_ameth.c
@@ -20,6 +20,7 @@
 #include "dh_local.h"
 #include <openssl/bn.h>
 #include "crypto/asn1.h"
+#include "crypto/dh.h"
 #include "crypto/evp.h"
 #include <openssl/cms.h>
 #include <openssl/core_names.h>
@@ -498,6 +499,13 @@ static int dh_pkey_export_to(const EVP_PKEY *from, void *to_keydata,
     const BIGNUM *priv_key = DH_get0_priv_key(dh);
     OSSL_PARAM *params;
     int rv;
+
+    /*
+     * If the DH method is foreign, then we can't be sure of anything, and
+     * can therefore not export or pretend to export.
+     */
+    if (dh_get_method(dh) != DH_OpenSSL())
+        return 0;
 
     if (p == NULL || g == NULL)
         return 0;
--- a/crypto/dh/dh_lib.c
+++ b/crypto/dh/dh_lib.c
@@ -45,6 +45,11 @@ int DH_set_method(DH *dh, const DH_METHOD *meth)
     return 1;
 }
 
+const DH_METHOD *dh_get_method(const DH *dh)
+{
+    return dh->meth;
+}
+
 DH *DH_new(void)
 {
     return dh_new_intern(NULL, NULL);
--- a/crypto/dsa/dsa_ameth.c
+++ b/crypto/dsa/dsa_ameth.c
@@ -528,6 +528,13 @@ static int dsa_pkey_export_to(const EVP_PKEY *from, void *to_keydata,
     OSSL_PARAM *params;
     int rv;
 
+    /*
+     * If the DSA method is foreign, then we can't be sure of anything, and
+     * can therefore not export or pretend to export.
+     */
+    if (DSA_get_method(dsa) != DSA_OpenSSL())
+        return 0;
+
     if (p == NULL || q == NULL || g == NULL)
         return 0;
 
--- a/crypto/evp/m_sigver.c
+++ b/crypto/evp/m_sigver.c
@@ -73,7 +73,7 @@ static int do_sigver_init(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
      */
     ERR_set_mark();
 
-    if (locpctx->keytype == NULL)
+    if (locpctx->engine != NULL || locpctx->keytype == NULL)
         goto legacy;
 
     /*
--- a/crypto/evp/signature.c
+++ b/crypto/evp/signature.c
@@ -359,7 +359,7 @@ static int evp_pkey_signature_init(EVP_PKEY_CTX *ctx, int operation)
      */
     ERR_set_mark();
 
-    if (ctx->keytype == NULL)
+    if (ctx->engine != NULL || ctx->keytype == NULL)
         goto legacy;
 
     /*
--- a/include/crypto/dh.h
+++ b/include/crypto/dh.h
@@ -23,3 +23,5 @@ int dh_get0_nid(const DH *dh);
 int dh_check_pub_key_partial(const DH *dh, const BIGNUM *pub_key, int *ret);
 int dh_check_priv_key(const DH *dh, const BIGNUM *priv_key, int *ret);
 int dh_check_pairwise(DH *dh);
+
+const DH_METHOD *dh_get_method(const DH *dh);


levitte:
While discussing on #11179, I found that this PR was incomplete.  I've just added a commit that completes it.

levitte:
> travis errors

D'oh.  Seems like I mixed up EVP_PKEY_CTX with EVP_PKEY...
Fixed by going back one commit.

levitte:
Travis error is an irrelevant timeout.  CIs are otherwise happy

levitte:
That being said, though, libp11 isn't happy,  The check is done at the wrong level, so this goes back to WIP for a bit

levitte:
Changes pushed, and tested as far as I have been able.  Now, it's up to @mouse07410 and @mtrojnar to try it out with libp11

mouse07410:
Please move the lines you're adding to that function to after this line https://github.com/openssl/openssl/blob/97ace6c2dae451ce8e3b099cf242968ecff128af/crypto/ec/ec_ameth.c#L638

Then all is good.

levitte:
> Please move the lines you're adding to that function to after this line 

Thanks.  Done

crypto/ec/ec_ameth.c
```diff
@@ -637,6 +637,13 @@ int ec_pkey_export_to(const EVP_PKEY *from, void *to_keydata,
             || (ecg = EC_KEY_get0_group(eckey)) == NULL)
         return 0;
 
+    /*
+     * If the DH method is foreign, then we can't be sure of anything, and
```
slontis: update comment to EC

crypto/evp/p_lib.c
```diff
@@ -30,6 +30,7 @@
 #include <openssl/serializer.h>
 #include <openssl/core_names.h>
 
+#include "crypto/dh.h"
```
slontis: Why?

crypto/rsa/rsa_ameth.c
```diff
@@ -1092,6 +1092,13 @@ static int rsa_pkey_export_to(const EVP_PKEY *from, void *to_keydata,
     OSSL_PARAM *params = NULL;
     int rv = 0;
 
+    /*
+     * If the DH method is foreign, then we can't be sure of anything, and
```
slontis: same here

crypto/evp/p_lib.c
levitte: Ah, it's a remnant from earlier incarnations of this PR, where the logic for checking the low level key method was done in this file, i.e. `dh_get_method()` was used.  Will remove,

crypto/ec/ec_ameth.c
levitte: Done

crypto/rsa/rsa_ameth.c
levitte: Fixed

crypto/evp/pmeth_fn.c
```diff
@@ -38,7 +38,7 @@ static int evp_pkey_asym_cipher_init(EVP_PKEY_CTX *ctx, int operation)
      */
     ERR_set_mark();
 
-    if (ctx->engine != NULL)
+    if (ctx->keytype != NULL)
```
mouse07410: ~~You **sure** about this?~~

Please revert this change. Everything else looks good, and with this one reverted all the OpenSSL and all the relevant `libp11` tests pass.

levitte: Specifically this one, *only* checking 'ctx->engine`, was plain wrong.

levitte:
Uhmm, the change that removes the check of `ctx->engine` should be handled by the checks in `export_to`, since an engine provided low level method is different than the libcrypto default, i.e. `export_to` should return zero, and the code that tries to export to providers should jump to legacy code.

crypto/evp/pmeth_fn.c
levitte: D'oh!  `!=` should be `==`...

mouse07410:
**Not good**. Without that `|| ctx->engine != NULL`  check several OpenSSL tests keep failing.

levitte:
> check several OpenSSL tests keep failing.

Which ones?  I see pkcs11-uri-without-token.softhsm and search-all-matching-tokens.softhsm failing, do you see something different?

mouse07410:
> Which ones? I see pkcs11-uri-without-token.softhsm and search-all-matching-tokens.softhsm failing, do you see something different?

Yes! OpenSSL's own tests are failing. Within `libp11` - your observation is correct, but I don't start testing it until "make test" completes in the OpenSSL directory.

That's why I think that the check the the engine must be added regardless - with it all the OpenSSL tests pass, **and** all the relevant `libp11` tests too. Those last two tests - I don't know. In theory, any operation on the token should involve cleanup in the end, so should trigger that crash. In practice - as you could see,  all the other tests pass and finish/exit normally. Since real tokens also seem to work correctly, it's probably ok.

levitte:
Ah, now I see what you mean.  Ok, will have a look at what happens when I'm done with some other stuff.

slontis:
Have you forgotten the fips sponsors meeting??

mouse07410:
@levitte may I suggest adding that check for now, add it fixes the OpenSSL tests and unbreaks libp11?

We can figure out later why what you thought should do that check earlier either failed or is not sufficient.

@slontis which one of us? :-) As I too have this week full of meetings. ;-)

slontis:
> @slontis which one of us? :-) As I too have this week full of meetings. ;-)

I could play along and say I know all about your meetings.. :)

levitte:
@mouse07410, you are right.  I just needed to confirm it to myself, to understand the mechanism that lead to that error.  Basically, it's perfectly possible for an ENGINE to have a different EVP_PKEY_METHOD but use the default low level key method underneath, which is undetectable for the `export_to` functions.

Thanks for your patience.

levitte:
This is no longer WIP

mouse07410:
@levitte thank you fur getting this issue solved.

slontis:
travis test is failing - so this is not quite there yet..

petrovr:
As pointed in #9915 default != openssl method. I expect checks to be based on openssl methods.
Current situation is not expected - see https://github.com/openssl/openssl/issues/9915#issuecomment-595053916 . It is expected sing to fail for all keys and verify to pass. This is not the case.

levitte:
> As pointed in #9915 default != openssl method.

Darnit.  You're right, of course.  Ok, I'll fix.

levitte:
I just modified the check to compare with the actual built-in methods.  @petrovr, does that look more like what you expected?  At the very least, I hope the CIs are happy now.

t8m:
>  At the very least, I hope the CIs are happy now.

Nope :disappointed:

levitte:
Ah!  I had missed one spot...

levitte:
Yay!  CIs are happy!

petrovr:
@levitte ,
Please add "Fixes #9915"

levitte:
> @levitte ,
> Please add "Fixes #9915"

Done.  For now, by adding an empty squash commit.

t8m:
The empty commit does not require re-approval.

PR #11036, opened 2020-02-07
Add data driven SELF TEST code for signatures and key agreement

slontis:
##### Checklist

- [x] documentation is added or updated
- [X] tests are added or updated

PR #11199, opened 2020-02-28
Add rsa sign to fips provider

slontis:
##### Checklist

### DIFF
--- a/crypto/provider_core.c
+++ b/crypto/provider_core.c
@@ -935,7 +935,7 @@ static const OSSL_DISPATCH core_dispatch_[] = {
     { OSSL_FUNC_BIO_READ_EX, (void (*)(void))BIO_read_ex },
     { OSSL_FUNC_BIO_FREE, (void (*)(void))BIO_free },
     { OSSL_FUNC_BIO_VPRINTF, (void (*)(void))BIO_vprintf },
-    { OSSL_FUNC_CORE_VSNPRINTF, (void (*)(void))BIO_vsnprintf },
+    { OSSL_FUNC_BIO_VSNPRINTF, (void (*)(void))BIO_vsnprintf },
     { OSSL_FUNC_SELF_TEST_CB, (void (*)(void))OSSL_SELF_TEST_get_callback },
 #endif
     { OSSL_FUNC_CRYPTO_MALLOC, (void (*)(void))CRYPTO_malloc },
--- a/crypto/provider_core.c
+++ b/crypto/provider_core.c
@@ -930,12 +930,12 @@ static const OSSL_DISPATCH core_dispatch_[] = {
     { OSSL_FUNC_CORE_CLEAR_LAST_ERROR_MARK,
       (void (*)(void))core_clear_last_error_mark },
     { OSSL_FUNC_CORE_POP_ERROR_TO_MARK, (void (*)(void))core_pop_error_to_mark },
-    { OSSL_FUNC_CORE_VSNPRINTF, (void (*)(void))BIO_vsnprintf },
     { OSSL_FUNC_BIO_NEW_FILE, (void (*)(void))BIO_new_file },
     { OSSL_FUNC_BIO_NEW_MEMBUF, (void (*)(void))BIO_new_mem_buf },
     { OSSL_FUNC_BIO_READ_EX, (void (*)(void))BIO_read_ex },
     { OSSL_FUNC_BIO_FREE, (void (*)(void))BIO_free },
     { OSSL_FUNC_BIO_VPRINTF, (void (*)(void))BIO_vprintf },
+    { OSSL_FUNC_CORE_VSNPRINTF, (void (*)(void))BIO_vsnprintf },
     { OSSL_FUNC_SELF_TEST_CB, (void (*)(void))OSSL_SELF_TEST_get_callback },
 #endif
     { OSSL_FUNC_CRYPTO_MALLOC, (void (*)(void))CRYPTO_malloc },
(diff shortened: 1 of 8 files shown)


crypto/rsa/rsa_pmeth.c
```diff
@@ -153,7 +154,9 @@ static int pkey_rsa_sign(EVP_PKEY_CTX *ctx, unsigned char *sig,
             if (ret <= 0)
                 return ret;
             ret = sltmp;
-        } else if (rctx->pad_mode == RSA_X931_PADDING) {
+        } else
+#endif /* FIPS_MODE */
+        if (rctx->pad_mode == RSA_X931_PADDING) {
```
levitte: Why are you changing this file at all?  It should *never* become part of the FIPS module, this is a legacy backend.

If there are things here that are missing in `providers/implementations/signature/rsa.c`, then they should be added there.

providers/implementations/signature/build.info
```diff
@@ -2,12 +2,10 @@
...
 
 IF[{- !$disabled{dsa} -}]
   SOURCE[$DSA_GOAL]=dsa.c
 ENDIF
 
-SOURCE[$RSA_GOAL]=rsa.c
-
-
+SOURCE[../../libfips.a]=rsa.c
+SOURCE[../../libnonfips.a]=rsa.c
```
levitte: Hmmmm, I know why, and it's not the cleanest...  some of the signing code could be placed in crypto/rsa/, especially considering how much is shared between crypto/rsa/rsa_pmeth.c and providers/implementations/signature/rsa.c...

I'm making a mental note on this for the moment being...

crypto/rsa/rsa_local.h
```diff
@@ -48,12 +48,12 @@ struct rsa_st {
     BIGNUM *dmp1;
     BIGNUM *dmq1;
     BIGNUM *iqmp;
+    RSA_PSS_PARAMS *pss;
```
levitte: Why did you leave the comment behind?

slontis: Oops..

slontis:
Note the test for this is deferred to PR #11036.. (i.e running the self test for rsa sign/verify confirmed that this is working in the fips module).

slontis:
Taken out of draft mode now that I have confirmed that the sign/verify functions correctly..

providers/fips/fipsprov.c
```diff
@@ -910,6 +912,9 @@ int OSSL_provider_init(const OSSL_PROVIDER *provider,
         case OSSL_FUNC_CORE_POP_ERROR_TO_MARK:
             c_pop_error_to_mark = OSSL_get_core_pop_error_to_mark(in);
             break;
+        case OSSL_FUNC_CORE_VSNPRINTF:
+            c_core_vsnprintf = OSSL_get_core_vsnprintf(in);
+            break;
         case OSSL_FUNC_CRYPTO_MALLOC:
             c_CRYPTO_malloc = OSSL_get_CRYPTO_malloc(in);
             break;
```
levitte: Note for future cleanup: we should really consider having common provider functions for this sort of thing.  I tried to start that with `providers/common/include/prov/bio.h` and `providers/common/bio_prov.c`

crypto/provider_core.c
```diff
@@ -893,8 +893,8 @@ static const OSSL_DISPATCH core_dispatch_[] = {
     { OSSL_FUNC_CORE_SET_ERROR_MARK, (void (*)(void))core_set_error_mark },
     { OSSL_FUNC_CORE_CLEAR_LAST_ERROR_MARK,
       (void (*)(void))core_clear_last_error_mark },
-    { OSSL_FUNC_CORE_POP_ERROR_TO_MARK,
-      (void (*)(void))core_pop_error_to_mark },
+    { OSSL_FUNC_CORE_POP_ERROR_TO_MARK, (void (*)(void))core_pop_error_to_mark },
+    { OSSL_FUNC_CORE_VSNPRINTF, (void (*)(void))BIO_vsnprintf },
```
levitte: Mind moving this line down a bit, so it's together with `OSSL_FUNC_BIO_VPRINTF`?

slontis: No I dont mind - but it doesnt have too much to do with BIO's

crypto/rsa/rsa_local.h
levitte: You misunderstood.  The comment on line 55 should be moved up.
(but yeah, removing the TODO was a good move nonetheless)

crypto/provider_core.c
levitte: It's in the BIO API...  but otherwise, you're remark is quite right.  However, they're the same kind of function, so...
(this is like discussing that while fprintf is a FILE function, printf isn't, and yet, they are usually presented together)

levitte:
Rebase needed.

slontis:
rebased..

crypto/rsa/rsa_local.h
levitte: You moved one comment too much.  The "be careful..." one belongs with `ex_data`

slontis: Ok moved it to the correct line (which is not where it was to start with) :)

slontis:
> Also, if you're only going to have a direct reference to `BIO_vsnprintf`, the number macro should

Renamed and moved any related code around to be near the other bio stuff.

slontis:
@levitte - this should address your comments..

slontis:
Thanks. Merged to master.

PR #11200, opened 2020-02-28
Use a correct OpenSSL logo

t8m:
And add a variant for dark backgrounds.


mspncp:
Thank you Tomas for doing the conversion. For the moment, I think it is legitimate to use the historical shape of the logo. But this does not change my general conviction that it is time to modernize our logo. I have a colleague, Daniel (@dsncp), who is a talented Web Designer and Web Programmer with a good eye for style. I managed to persuade him to participate in my current stealth project (OpenSSL Project Pages), which I will introduce officially in the next weeks. In that context we already discussed the logo and he has some promising ideas.

So maybe we will come up with some suggestions in the next days/weeks as a basis for a discussion. But that's a subject beyound this pull request which we don't need to rush.

For the moment I only have one change request: could you please try to increase the bounding box around the logo, currently it just looks huge [on the landing page](https://github.com/t8m/openssl/tree/fix-logo). GitHub automatically scales it to full width and there is no way rescale it to a smaller size using markdown only.

If you use Inkscape,  you can open 'File > Document Properties' (or press Shift+Ctrl+D) and then unfold the 'Resize page to content...' subdialog on the 'Page' tab. Here you can add a well defined border around the natural bounding box. I think the horizontal border should be about double the size as the vertical border. The values 100 and 200 are my initial tries.

image

t8m:
Better now?

mspncp:
Yes, much better. :)

For everybody else, here's a screenshot for your convenience. Alternatively, you can view [the landing page](https://github.com/t8m/openssl/tree/fix-logo) or the [README.md](https://github.com/t8m/openssl/blob/fix-logo/README.md) live on t8m's GitHub clone.

image

mattcaswell:
IMO the logo should be entirely based on the current shape of the logo as it appears on the website. We should not be reverting to the historical shape. Any decision to do anything else requires an OMC discussion. Unfortunately I'm not in a position to start that discussion at the moment (currently at the RSA conference and will be travelling home later today). For now I am putting this on hold.

mspncp:
> IMO the logo should be entirely based on the current shape of the logo as it appears on the website. We should not be reverting to the historical shape. Any decision to do anything else requires an OMC discussion.

Thank you Matt speaking up,it seems like we agree in two important points:

- Even if it was never officially announced, the _current_ logo on openssl.org is the de-facto logo and changing it would require an OMC vote.

- If we change the logo at some time, it should be towards a more modern style, and not towards restoring a retro-style.

Personally, I think the current OpenSSL logo needs a modernization, but we don't need to rush a decision. As I [wrote on openssl-project](https://mta.openssl.org/pipermail/openssl-project/2020-February/001848.html) to the openssl-project list), we could collect suggestions first and then have a contest.

mspncp:
> IMO the logo should be entirely based on the current shape of the logo as it appears on the website.

The current logo on master tries to get as close to to the web original as it can. Currently, it is implemented in html in [banner.shtml](https://github.com/openssl/web/blob/master/inc/banner.shtml#L4-L8).

```html
    <h1>
      <a href="/">
        <span id="header-open-text">Open</span><span id="header-ssl-text">SSL</span>
      </a>
    </h1>
```
According to [the style sheet](https://github.com/openssl/web/blob/master/inc/screen.css#L131-L158), the font-family is a list of allowed fonts,

```
font-family: "HelveticaNeue-Light", "Helvetica Neue Light", "Helvetica Neue", Helvetica, Arial, "Lucida Grande", sans-serif;
```

the fallback being sans-serif. This is exactly the font-family chosen in [openssl.svg](https://raw.githubusercontent.com/openssl/openssl/master/doc/images/openssl.svg) on master.

t8m:
I have to say that I do not like the current version of the logo on the OpenSSL website. I do not insist on returning to the old logo but the current one looks visually unpleasant to me and the old one (without any 3d effects) looks much better to me.

Anyway a PR is probably not the right place to discuss a logo changes so I am closing this.

PR #11206, opened 2020-02-29
Andoid cross compile: change ANDROID_NDK_HOME to ANDROID_NDK_ROOT

### LINKED ISSUE #11205: ANDROID_NDK_HOME should use ANDROID_NDK_ROOT
According to [`NOTES.ANDROID`](https://github.com/openssl/openssl/blob/master/NOTES.ANDROID):

>  Apart from PATH adjustment you need to set ANDROID_NDK_HOME
> environment to point at the NDK directory. If you're using a side-by-side
> NDK the path will look something like /some/where/android-sdk/ndk/<ver>

According to the NDK folks in [Recommended NDK Directory?](http://groups.google.com/group/android-ndk/browse_thread/thread/a998e139aca71d77), you should set two variables. The first is `ANDROID_SDK_ROOT`, and the second is `ANDROID_NDK_ROOT`. `ANDROID_SDK_ROOT` and `ANDROID_NDK_ROOT` are separate, independent environmental variables. Assuming there's some variable `ANDROID_NDK_HOME` that points to a SDK install and the NDK is some subfolder is just plain wrong.

In fact, setting `ANDROID_NDK_HOME` like that will break in another way because the Android Studio folks made their own shit up (instead of following the NDK folks) and used `ANDROID_NDK_HOME` for something else (iirc).

`ANDROID_NDK_HOME` is breaking Android builds for well configured environments.

levitte:
According to forum discussions with NDK developers, ANDROID_NDK_HOME
is used for something else.

Fixes #11205

### DIFF
--- a/Configurations/15-android.conf
+++ b/Configurations/15-android.conf
@@ -24,17 +24,17 @@
 
             my $ndk_var;
             my $ndk;
-            foreach (qw(ANDROID_NDK_HOME ANDROID_NDK)) {
+            foreach (qw(ANDROID_NDK_ROOT ANDROID_NDK)) {
                 $ndk_var = $_;
                 $ndk = $ENV{$ndk_var};
                 last if defined $ndk;
             }
-            die "\$ANDROID_NDK_HOME is not defined"  if (!$ndk);
+            die "\$ANDROID_NDK_ROOT is not defined"  if (!$ndk);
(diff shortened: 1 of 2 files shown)


levitte:
Ping?

levitte:
PING?

PR #11210, opened 2020-02-29
Fix drop of const qualifier

misery:
The parameter got "const" in 9fdcc21fdc9 but that was not added
to cast. So this throws a -Wcast-qual in user code.

error: cast from 'const DUMMY *' to 'ASN1_VALUE_st *' drops const qualifier [-Werror,-Wcast-qual]

CLA: trivial


paulidale:
Merged to master.  Thanks for the fix.

PR #11214, opened 2020-03-02
fix a minor bug of s_client

ningning-718:
-CAstore's option shoule be OPT_CASTORE, instead of OPT_CAFILE

##### Checklist


DDvO:
Thanks @ningning-718 for bringing this up.
There is something more broken with the `-CAstore` vs. `-CAfile`.

Users (including myself) have already been wondering what the difference between the two should be. It turns out that (before the fix proposed here) their effect was identical:
```
    {"CAfile", OPT_CAFILE, '<', "PEM format file of CA's"},
    {"CAstore", OPT_CAFILE, ':', "URI to store of CA's"},
```
although these are two different options with somewhat different descriptions.

It turns out that the proposed fix is insufficient because then the `-CAstore` argument ends up being ignored!

DDvO:
@ningning-718, it would have been better to continue with this PR 
rather than opening a new one   with an extended fix.

Closing this one in favor of the new one: #11215.

PR #11217, opened 2020-03-02
config, Configure: move the check of removed crypto/ sub-systems

levitte:
The 'config' script checked for a bunch of crypto/ sub-system
directories, and added 'no-' options if they weren't there.

We move it to 'Configure' in an effort to simplify 'config' for
further work.

Note: this is pretty much a historical thing.  In modern OpenSSL, it's
much simpler to edit the SUBDIRS statement in crypto/build.info.
However, it's been claimed the there are those who still remove some
of these sub-system sources.

### DIFF
--- a/Configure
+++ b/Configure
@@ -265,6 +265,19 @@ if (grep /^reconf(igure)?$/, @argvcopy) {
 
 $config{perlargv} = [ @argvcopy ];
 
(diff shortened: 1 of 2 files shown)


levitte:
Ping.  This helps following up on #11011

Issue #10725, opened 2019-12-29
The coding style guideline document should be updated


DDvO:
Every now and then when contributing code we come across situations where reviewers ask for changes according to coding style that hase become common practice within OpenSSL but for some reason is not reflected in the official published reference document https://www.openssl.org/policies/codingstyle.html (which apparently has not been updates since almost four years).

For instance, it (likely) should contain thinks like:
* Within function bodies have an empty line between local variable declarations and subsequent code.
* Do not place binary logical operators `&&` and `||` at the end of a line but at the beginning of the following line. They should have an extra indentation (of +4 characters).
* Comments referring to the code line before or after should be indented equally to that code line.
* If a line barely exceeds the 80-char width limit and adding a line break would clearly reduce readability, the line does not need to be split.
* Make (in-)equality checks with integer 0 (but not with Boolean 'false') explicit, e.g., write 'if (i != 0)' rather than 'if (i)'.
* Do not use needless parentheses, e.g., replace 'if ((ptr == NULL) && (!f(x)))' by 'if (ptr == NULL && !f(x))'.
* Avoid (needless) checks for NULL when calling free() functions.
* Avoid `goto err` when a simple return is sufficient.
* Use `const` in types wherever possible.
* Check for any potential error condition, in particular those potentially reported in the result value of (non-void) functions called.
* Be consistent with error reporting on a per-function basis: when adding errors to the OpenSSL error stack (using `XYZerr()`) then this should happen in all error paths. When some other function is called then assume this already has been achieved by that function, where one may add a comment such as "XYZerr already called" to make it clear that the omission of a specific error is deliberate.
* Avoid calling _dup() functions when `*_up_ref()` is sufficient, e.g., for `X509`.
* Avoid `#if 0` (as well as other forms of disabling or commenting out code) and `#if 1`.
* Do not use `enum` in public headers such as in `include/openssl/`.
* Make sure to spell-check comments and names.
* Use brief but descriptive names for types and functions.
* For getter and setter functions use names containing `get0_` or `get1_` (rather than `get_`) and `set0_` or `set1_` (rather than `set_`) and `push0_` or `push1_` (rather than `push_`) to point out if the function has aliasing or copying semantics.
* Align to RFC-defined terms and wording, including camelCase where appropriate.
* Use existing OpenSSL macros for declarations as far as possible, e.g., for function prototypes and type definitions.
[...]

PR #11220, opened 2020-03-02
DOC: Add more description of EVP_PKEY_fromdata(), and examples

### LINKED ISSUE #11131: DOC: EVP_PKEY_fromdata() documentation needs enhancing
To borrow the words that @kroeckx uttered in #11095:

> We seem to have an
> EVP_PKEY_fromdata() function, but it's not documented what that
> function expects as input.
> 
> A function like EVP_PKEY_param_fromdata_settable() is making
> little sense to me. That looks like something that needs to be
> in the documentation, rather that first writing a test program to
> see what parameters I can set.
> 
> That whole manpage needs examples.

levitte:
Fixes #11131

### DIFF
--- a/doc/man3/EVP_PKEY_CTX_new.pod
+++ b/doc/man3/EVP_PKEY_CTX_new.pod
@@ -32,7 +32,8 @@ The EVP_PKEY_CTX_new_from_name() function allocates a public key algorithm
 context using the library context I<libctx> (see L<OPENSSL_CTX(3)>), the
 key type specified by I<name> and the property query I<propquery>.  None
 of the arguments are duplicated, so they  must remain unchanged for the
-lifetime of the returned B<EVP_PKEY_CTX> or of any of its duplicates.
+lifetime of the returned B<EVP_PKEY_CTX> or of any of its duplicates.  Read
+further about the possible names in L</NOTES> below.
 
 The EVP_PKEY_CTX_new_from_pkey() function allocates a public key algorithm
 context using the library context I<libctx> (see L<OPENSSL_CTX(3)>) and the
@@ -52,22 +53,43 @@ If I<ctx> is NULL, nothing is done.
 
 =head1 NOTES
 
-=over 4
-
-=item 1.
+=head2 On B<EVP_PKEY_CTX>
 
 The B<EVP_PKEY_CTX> structure is an opaque public key algorithm context used
 by the OpenSSL high level public key API. Contexts B<MUST NOT> be shared between
 threads: that is it is not permissible to use the same context simultaneously
 in two threads.
 
-=item 2.
+=head2 On Key Types
 
 We mention "key type" in this manual, which is the same
 as "algorithm" in most cases, allowing either term to be used
 interchangeably.  There are algorithms where the I<key type> and the
 I<algorithm> of the operations that use the keys are not the same,
 such as EC keys being used for ECDSA and ECDH operations.
+
+Key types are given in two different manners:
+
+=over 4
+
+=item Legacy NID or EVP_PKEY type
+
+This is the I<id> used with EVP_PKEY_CTX_new_id().
+
+These are B<EVP_PKEY_RSA>, B<EVP_PKEY_RSA_PSS>, B<EVP_PKEY_DSA>,
+B<EVP_PKEY_DH>, B<EVP_PKEY_EC>, B<EVP_PKEY_SM2>, B<EVP_PKEY_X25519>,
+B<EVP_PKEY_X448>, and are used by legacy methods.
+
+=item Name strings
+
+This is the I<name> used with EVP_PKEY_CTX_new_from_name().
+
+These are names like "RSA", "DSA", and what's available depends on what
+providers are currently accessible.
+
+The OpenSSL providers offer a set of key types available this way, please
+see L<OSSL_PROVIDER-FIPS(7)> and L<OSSL_PROVIDER-default(7)> and related
+documentation for more information.
 
 =back
 
--- a/doc/man3/EVP_PKEY_fromdata.pod
+++ b/doc/man3/EVP_PKEY_fromdata.pod
@@ -139,7 +139,7 @@ TODO Write a set of cookbook documents and link to them.
      0x47
  };
  const OSSL_PARAM params[] = {
-     OSSL_PARAM_ulong("curve-name", "prime256v1"),
+     OSSL_PARAM_utf8_string("curve-name", "prime256v1"),
      OSSL_PARAM_BN("priv", priv, sizeof(priv));
      OSSL_PARAM_BN("pub", pub, sizeof(pub));
      OSSL_PARAM_END
--- a/doc/man3/EVP_PKEY_fromdata.pod
+++ b/doc/man3/EVP_PKEY_fromdata.pod
@@ -140,8 +140,8 @@ TODO Write a set of cookbook documents and link to them.
  };
  const OSSL_PARAM params[] = {
      OSSL_PARAM_utf8_string("curve-name", "prime256v1"),
-     OSSL_PARAM_BN("priv", priv, sizeof(priv));
-     OSSL_PARAM_BN("pub", pub, sizeof(pub));
+     OSSL_PARAM_BN("priv", priv, sizeof(priv)),
+     OSSL_PARAM_BN("pub", pub, sizeof(pub)),
      OSSL_PARAM_END
  };
 
@@ -204,7 +204,7 @@ TODO Write a set of cookbook documents and link to them.
  }
 
 The descriptor L<OSSL_PARAM(3)> returned by
-EVP_PKEY_key_fromdata_settable() may also be used programatically, for
+EVP_PKEY_key_fromdata_settable() may also be used programmatically, for
 example with L<OSSL_PARAM_allocate_from_text(3)>.
 
 =head1 SEE ALSO
--- a/doc/man3/EVP_PKEY_fromdata.pod
+++ b/doc/man3/EVP_PKEY_fromdata.pod
@@ -64,8 +64,8 @@ not supported by the public key algorithm.
 =head1 EXAMPLES
 
 These examples are very terse for the sake of staying on topic, which
-are the EVP_PKEY_fromdata() set of functions.  In real applications,
-BIGNUMs would be handled, and converted to byte arrays with
+is the EVP_PKEY_fromdata() set of functions.  In real applications,
+BIGNUMs would be handled and converted to byte arrays with
 BN_bn2nativepad(), but that's off topic here.
 
 =begin comment
@@ -115,7 +115,7 @@ TODO Write a set of cookbook documents and link to them.
   * These arrays represent large numbers, big endian organization.
   * In a real application, these would probably be bignums that get
   * converted to the native integer organization with BN_bn2nativepad().
-  * We're not doing that here, since this is not an of BIGNUM
+  * We're not doing that here, since this is not an example of BIGNUM
   * functionality, but an example of EVP_PKEY_fromdata().
   */
  #ifndef B_ENDIAN
@@ -165,7 +165,7 @@ TODO Write a set of cookbook documents and link to them.
  /* Program expects a key type as first argument */
  int main(int argc, char *argv[])
  {
-     EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_from_name(NULL, argv[0], NULL);
+     EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_from_name(NULL, argv[1], NULL);
      const *OSSL_PARAM *settable_params = NULL;
 
      if (ctx == NULL
--- a/doc/man3/EVP_PKEY_fromdata.pod
+++ b/doc/man3/EVP_PKEY_fromdata.pod
@@ -205,7 +205,7 @@ TODO Write a set of cookbook documents and link to them.
 
 The descriptor L<OSSL_PARAM(3)> returned by
 EVP_PKEY_key_fromdata_settable() may also be used programatically, for
-example with L<OSSL_PARAM_allocate_from_text()>.
+example with L<OSSL_PARAM_allocate_from_text(3)>.
 
 =head1 SEE ALSO
 
--- a/doc/man3/EVP_PKEY_fromdata.pod
+++ b/doc/man3/EVP_PKEY_fromdata.pod
@@ -147,7 +147,7 @@ TODO Write a set of cookbook documents and link to them.
 
  int main()
  {
-     EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_from_name(NULL, "RSA", NULL);
+     EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_from_name(NULL, "EC", NULL);
      EVP_PKEY *pkey = NULL;
 
      if (ctx == NULL
--- a/doc/man3/EVP_PKEY_fromdata.pod
+++ b/doc/man3/EVP_PKEY_fromdata.pod
@@ -18,7 +18,7 @@ EVP_PKEY_param_fromdata_settable, EVP_PKEY_key_fromdata_settable
 
 =head1 DESCRIPTION
 
-The function described here are used to create new keys from user
+The functions described here are used to create new keys from user
 provided key data, such as I<n>, I<e> and I<d> for a minimal RSA
 keypair.
 
@@ -63,7 +63,18 @@ not supported by the public key algorithm.
 
 =head1 EXAMPLES
 
-=head2 Creating an RSA keypair
+These examples are very terse for the sake of staying on topic, which
+are the EVP_PKEY_fromdata() set of functions.  In real applications,
+BIGNUMs would be handled, and converted to byte arrays with
+BN_bn2nativepad(), but that's off topic here.
+
+=begin comment
+
+TODO Write a set of cookbook documents and link to them.
+
+=end comment
+
+=head2 Creating an RSA keypair using raw key data
 
  #include <openssl/evp.h>
 
@@ -83,6 +94,57 @@ not supported by the public key algorithm.
      OSSL_PARAM_END
  };
      
+ int main()
+ {
+     EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_from_name(NULL, "RSA", NULL);
+     EVP_PKEY *pkey = NULL;
+
+     if (ctx == NULL
+         || !EVP_PKEY_key_fromdata_init(ctx)
+         || !EVP_PKEY_fromdata(ctx, &pkey, params))
+         exit(1);
+
+     /* Do what you want with |pkey| */
+ }
+
+=head2 Creating an ECC keypair using raw key data
+
+ #include <openssl/evp.h>
+
+ /*
+  * These arrays represent large numbers, big endian organization.
+  * In a real application, these would probably be bignums that get
+  * converted to the native integer organization with BN_bn2nativepad().
+  * We're not doing that here, since this is not an of BIGNUM
+  * functionality, but an example of EVP_PKEY_fromdata().
+  */
+ #ifndef B_ENDIAN
+ # error "We haven't prepared little endian arrays"
+ #endif
+ const unsigned char priv[] = {
+     0xb9, 0x2f, 0x3c, 0xe6, 0x2f, 0xfb, 0x45, 0x68,
+     0x39, 0x96, 0xf0, 0x2a, 0xaf, 0x6c, 0xda, 0xf2,
+     0x89, 0x8a, 0x27, 0xbf, 0x39, 0x9b, 0x7e, 0x54,
+     0x21, 0xc2, 0xa1, 0xe5, 0x36, 0x12, 0x48, 0x5d
+ };
+ const unsigned char pub[] = {
+     0x04, 0xcf, 0x20, 0xfb, 0x9a, 0x1d, 0x11, 0x6c,
+     0x5e, 0x9f, 0xec, 0x38, 0x87, 0x6c, 0x1d, 0x2f,
+     0x58, 0x47, 0xab, 0xa3, 0x9b, 0x79, 0x23, 0xe6,
+     0xeb, 0x94, 0x6f, 0x97, 0xdb, 0xa3, 0x7d, 0xbd,
+     0xe5, 0x26, 0xca, 0x07, 0x17, 0x8d, 0x26, 0x75,
+     0xff, 0xcb, 0x8e, 0xb6, 0x84, 0xd0, 0x24, 0x02,
+     0x25, 0x8f, 0xb9, 0x33, 0x6e, 0xcf, 0x12, 0x16,
+     0x2f, 0x5c, 0xcd, 0x86, 0x71, 0xa8, 0xbf, 0x1a,
+     0x47
+ };
+ const OSSL_PARAM params[] = {
+     OSSL_PARAM_ulong("curve-name", "prime256v1"),
+     OSSL_PARAM_BN("priv", priv, sizeof(priv));
+     OSSL_PARAM_BN("pub", pub, sizeof(pub));
+     OSSL_PARAM_END
+ };
+
  int main()
  {
      EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_from_name(NULL, "RSA", NULL);
--- a/doc/man3/EVP_PKEY_fromdata.pod
+++ b/doc/man3/EVP_PKEY_fromdata.pod
@@ -18,6 +18,17 @@ EVP_PKEY_param_fromdata_settable, EVP_PKEY_key_fromdata_settable
 
 =head1 DESCRIPTION
 
+The function described here are used to create new keys from user
+provided key data, such as I<n>, I<e> and I<d> for a minimal RSA
+keypair.
+
+These functions use an B<EVP_PKEY_CTX> context, which should primarly
+be created with L<EVP_PKEY_CTX_new_from_name(3)> or
+L<EVP_PKEY_CTX_new_id(3)>.
+
+The exact key data that the user can pass depends on the key type.
+These are passed as an L<OSSL_PARAM(3)> array.
+
 EVP_PKEY_param_fromdata_init() initializes a public key algorithm context
 for creating key parameters from user data.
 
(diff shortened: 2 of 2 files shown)


doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -18,6 +18,17 @@ EVP_PKEY_param_fromdata_settable, EVP_PKEY_key_fromdata_settable
...
+
+These functions use an B<EVP_PKEY_CTX> context, which should primarly
+be created with L<EVP_PKEY_CTX_new_from_name(3)> or
+L<EVP_PKEY_CTX_new_id(3)>.
+
+The exact key data that the user can pass depends on the key type.
+These are passed as an L<OSSL_PARAM(3)> array.
+
 EVP_PKEY_param_fromdata_init() initializes a public key algorithm context
 for creating key parameters from user data.
```
mattcaswell: I realised the reference below to `the "Built-in RSA Import/Export Types" section on the L<provider-keymgmt(7)> page` isn't quite right - because we specifically reference the RSA import/export types, but the text is referring to all key types. The wording probably needs to be adjusted to say something about that section and the following sections...or something along those lines.

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -50,9 +61,93 @@ EVP_PKEY_fromdata() return 1 for success and 0 or a negative value for
 failure.  In particular a return value of -2 indicates the operation is
 not supported by the public key algorithm.
 
+=head1 EXAMPLES
+
+=head2 Creating an RSA keypair
```
mattcaswell: I think an example for a second, different key type might be useful, so the reader can see what is the same and what is different between different types of key.

doc/man3/EVP_PKEY_fromdata.pod
levitte: We should have a *different* page for our providers.  `provider-*.pod` is not meant to hold the implementation details for our providers, they are meant as a description for those who want to write their own providers, which is different.

doc/man3/EVP_PKEY_fromdata.pod
levitte: Do you mean a key type that we know about, or the fictitious BLARGH?

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -18,6 +18,17 @@ EVP_PKEY_param_fromdata_settable, EVP_PKEY_key_fromdata_settable
 
 =head1 DESCRIPTION
 
+The function described here are used to create new keys from user
```
slontis: functions

doc/man3/EVP_PKEY_fromdata.pod
slontis: Creating an RSA keypair using raw key data

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -50,9 +61,93 @@ EVP_PKEY_fromdata() return 1 for success and 0 or a negative value for
...
+
+ #include <openssl/evp.h>
+
+ /*
+  * These are extremely small to make this example simple.  A real
+  * and secure application will not use such small numbers.  A real
+  * and secure application is expected to use BIGNUMs, and to build
+  * this array dynamically.
+  */
+ const unsigned long rsa_n = 0xbc747fc5;
```
slontis: If we are going to do this it should probably be a real case using proper BN instead of this invalid key data..

I think we also need to make the param_builder public..

doc/man3/EVP_PKEY_fromdata.pod
mattcaswell: These parameters sit on the interface between providers and applications. They are sent by applications and received by providers. It would seem very strange to have the same things documented twice. I don't mind too much *where* they are documented if you want to move them. But right at the moment they sit in `provider-keymgmt(7)` - so the reference should be right.

doc/man3/EVP_PKEY_fromdata.pod
mattcaswell: I'd suggest a real key type. Say X25519 or ECDH or...whatever.

mattcaswell:
Travis failures look relevant.

doc/man3/EVP_PKEY_fromdata.pod
levitte: I don't think it has to be valid data for this example. What will the input be? Will that add a function that builds BNs from raw data? That's extra stuff that's going to be distracting more than anything else.

We could consider a more complete cookbook, but that should be elsewhere.

slontis: Ok -  keep a mental note maybe that we should create a link here to the more complete one..

slontis:
make doc-nits failed

levitte:
> make doc-nits failed

```
/usr/bin/perl ./util/find-doc-nits -n -l -e
doc/man3/EVP_PKEY_fromdata.pod:1: Section missing in OSSL_PARAM_allocate_from_text()
*** ERROR: Apparent command =head1 not preceded by blank line at line 17 in file doc/internal/man3/s2i_ASN1_UTF8STRING.pod
Makefile:1752: recipe for target 'doc-nits' failed
make: *** [doc-nits] Error 1
```

So, er, how did *that* get the through?

doc/man3/EVP_PKEY_fromdata.pod
levitte: ECDH isn't a key type...

doc/man3/EVP_PKEY_fromdata.pod
levitte: Fixed

doc/man3/EVP_PKEY_fromdata.pod
levitte: Okei

mattcaswell: Well ok, EC then.

doc/man3/EVP_PKEY_fromdata.pod
levitte: I could add a comment saying so

levitte: I also added a bit of introdductionary text at the beginning of the EXAMPLES section:

> These examples are very terse for the sake of staying on topic, which
are the EVP_PKEY_fromdata() set of functions.  In real applications,
BIGNUMs would be handled, and converted to byte arrays with
BN_bn2nativepad(), but that's off topic here.

levitte:
> Travis failures look relevant.

Actually wasn't, but I've added a cleanup commit nonetheless.

doc/man3/EVP_PKEY_fromdata.pod
levitte: Added an ECC example

doc/man3/EVP_PKEY_fromdata.pod
levitte: I didn't say they should be duplicated...  they should indeed be moved.

kroeckx:
So I still do not get the point of EVP_PKEY_param_fromdata_settable(). Whatever that returns should be in the documentation.

kroeckx:
So we have a section that says:
`EVP_PKEY_fromdata() creates key parameters or a key, given data from I<params> and a context that's been initialized with EVP_PKEY_param_fromdata_init() or EVP_PKEY_key_fromdata_init().  The result is written to I<*ppkey>. The parameters that can be used for various types of key are as described in the "Built-in RSA Import/Export Types" section on the L<provider-keymgmt(7)> page.`

This needs a rewrite.

Clearly, the `various key types` is more than just RSA.

As far as I know, it doesn't "create" parameters or a key. I can at least confuse create with generate.

Since we pass a `EVP_PKEY **`, I assume an EVP_PKEY might be allocated (if *pkey is NULL?). I also assume that the EVP_PKEY will get associated with the EVP_PKEY_CTX.

I think that first sentence is also trying to put too much into one sentence.

kroeckx:
It might be obvious to you what the difference between parameters and the key is. But it would not hurt to repeat that, have an example with parameters. The provider-keymgmt could also clarify what is part of the key and what is a parameters.

kroeckx:
Is there a reason why we need a difference between parameters and keys in these functions?

levitte:
> Is there a reason why we need a difference between parameters and keys in these functions?

You're the second person to ask.  These functions are fairly close to key / param generation, the only real difference that the numbers are humanly generated rather than randomly, and key / param generation is separated.

Also, I can easily imagine that someone wants to use hand crafted FFC params and then use key generation to get a DSA key.  It's also not too hard to imagine that a backend might want to check that it got any key numbers at all if it was given a selector that includes OSSL_KEYMGMT_SELECT_KEYPAIR.  In that case, trying to create an EVP_PKEY with just hand crafted FFC params would fail without a function that uses a more limited selector.

And before you ask, I don't think we should make the OSSL_KEYMGMT selectors public (or turn them into EVP selectors).  That's an entirely new concept, and one I'm not keen to let out all the way.  We *might* somewhere in the future, but let's let it mature in a controlled way, yeah?

levitte:
> So I still do not get the point of EVP_PKEY_param_fromdata_settable(). Whatever that returns should be in the documentation.

I agree that it should be in the documentation, but not in *this* documentation.  This is comparable to, say, [ioctl(2)](http://man7.org/linux/man-pages/man2/ioctl.2.html), which doesn't describe any specific parameter in detail, but refers to other pages where they are enumerated.

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -50,9 +61,155 @@ EVP_PKEY_fromdata() return 1 for success and 0 or a negative value for
...
+ const OSSL_PARAM[] = {
+     OSSL_PARAM_ulong("n", &rsa_n),
+     OSSL_PARAM_ulong("e", &rsa_e),
+     OSSL_PARAM_ulong("d", &rsa_d),
+     OSSL_PARAM_END
+ };
+     
+ int main()
+ {
+     EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_from_name(NULL, "RSA", NULL);
```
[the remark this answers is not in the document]
levitte: You do know a teensie bit more...  it doesn't just allocate "something", it allocates a `EVP_PKEY_CTX` :wink:

Either way, that's not in scope for this PR.  PR welcome!

BTW, did you actually look?  https://github.com/openssl/openssl/blob/master/doc/man3/EVP_PKEY_CTX_new.pod

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -50,9 +61,155 @@ EVP_PKEY_fromdata() return 1 for success and 0 or a negative value for
...
+  * and secure application is expected to use BIGNUMs, and to build
+  * this array dynamically.
+  */
+ const unsigned long rsa_n = 0xbc747fc5;
+ const unsigned long rsa_e = 0x10001;
+ const unsigned long rsa_d = 0x7b133399;
+ const OSSL_PARAM[] = {
+     OSSL_PARAM_ulong("n", &rsa_n),
+     OSSL_PARAM_ulong("e", &rsa_e),
+     OSSL_PARAM_ulong("d", &rsa_d),
```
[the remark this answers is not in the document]
levitte: That documentation is incorrect.  PR welcome!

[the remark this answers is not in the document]
levitte: To clarify: the numbers are treated as BIGNUM on the receiving end, and we currently only support them as unsigned integers in the OSSL_PARAM structure.  This is a deliberate choice made very early on to avoid having to do what's done with an ASN1 INTEGER for positive numbers with the highest bit set, and after having concluded that all key numbers are always positive anyway.

levitte:
> So we have a section that says:
`EVP_PKEY_fromdata() creates key parameters or a key, given data from I<params> and a context that's been initialized with EVP_PKEY_param_fromdata_init() or EVP_PKEY_key_fromdata_init().  The result is written to I<*ppkey>. The parameters that can be used for various types of key are as described in the "Built-in RSA Import/Export Types" section on the L<provider-keymgmt(7)> page.`
> 
> This needs a rewrite.
> 
> Clearly, the `various key types` is more than just RSA.

Yes.  The way we're currently documenting the diverse key parameters in *our implementations* is an abomination in my mind.  I have an idea where to move them, though.  At that point, the reference here will have to change.  Can you tolerate that it looks bad for now, or would you prefer that the key parameters documentation gets moved first?

> As far as I know, it doesn't "create" parameters or a key. I can at least confuse create with generate.

Yes and no.  The backend creates an internal keydata structure, and a new `EVP_PKEY` gets created by libcrypto.  In that sense, "create" is quite correct.  Also, it *is* conceptually close to key generation, the only real difference is where the numbers come from (human vs random number device).  Also, have you noticed the title of this page?  (emphasis mine)

> =head1 NAME
> 
> EVP_PKEY_param_fromdata_init, EVP_PKEY_key_fromdata_init, EVP_PKEY_fromdata,
> EVP_PKEY_param_fromdata_settable, EVP_PKEY_key_fromdata_settable
> - *functions to create key parameters and keys from user data*

That being said, would this introduction work better for you?

`EVP_PKEY_fromdata() creates the structure to store key parameters or a key, ...`

> Since we pass a `EVP_PKEY **`, I assume an EVP_PKEY might be allocated (if *pkey is NULL?). I also assume that the EVP_PKEY will get associated with the EVP_PKEY_CTX.

Correct, the EVP_PKEY might be allocated (if *pkey == NULL).  It might also be overwritten (if *pkey != NULL).  Just like with key generation.

The EVP_PKEY_CTX has the same function as for key generation, i.e. it holds pointers to the backend, which is then used when actually creating the keydata structure.  It may also be used to set other parameters between fromdata_init and fromdata, which might affect aspects of keydata creation.  We *currently* do not have any such parameters (or well, I'm sure @romen could argue that the cofactor setting could be used here), but...

> I think that first sentence is also trying to put too much into one sentence.

I'm open for suggestions.

kroeckx:
On Fri, Mar 06, 2020 at 12:41:33AM -0800, Richard Levitte wrote:
> > So I still do not get the point of EVP_PKEY_param_fromdata_settable(). Whatever that returns should be in the documentation.
> 
> I agree that it should be in the documentation, but not in *this* documentation.  This is comparable to, say, [ioctl(2)](http://man7.org/linux/man-pages/man2/ioctl.2.html), which doesn't describe any specific parameter in detail, but refers to other pages where they are enumerated.

My point is that this function should not exist. The example
program you wrote for it just writes things that should be in the
documentation. Please provide an other usecase for this function
than printing what's already in the documentation.

doc/man3/EVP_PKEY_fromdata.pod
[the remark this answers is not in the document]
kroeckx: Yes, I looked. I actually also know that name is a key type, but I have no idea what propquery is, nor do I have an idea of what the supported key types are, and what passing NULL for propquey means. And the OPENSSL_CTX is even more mysterious.

doc/man3/EVP_PKEY_fromdata.pod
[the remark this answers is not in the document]
kroeckx: As far as I know, OSSL_PARAM doesn't do BIGNUM, the numbers we pass with OSSL_PARAM are in the native endian format. BIGNUM isn't  in the native endian format (on all platforms). OSSL_PARAM makes a difference between signed and unsigned integers.

It makes perfect sense for those parameters to only support unsigned integers, but the documentation currently says only signed integers are supported. If you don't want to fix that in this PR, please open an issue, I do not have the time to create PRs.

kroeckx:
> > Since we pass a `EVP_PKEY **`, I assume an EVP_PKEY might be allocated (if *pkey is NULL?). I also assume that the EVP_PKEY will get associated with the EVP_PKEY_CTX.
> 
> Correct, the EVP_PKEY might be allocated (if *pkey == NULL).  It might also be overwritten (if *pkey != NULL).  Just like with key generation.
> 
> The EVP_PKEY_CTX has the same function as for key generation, i.e. it holds pointers to the backend, which is then used when actually creating the keydata structure.  It may also be used to set other parameters between fromdata_init and fromdata, which might affect aspects of keydata creation.  We *currently* do not have any such parameters (or well, I'm sure @romen could argue that the cofactor setting could be used here), but...

Please document that.

doc/man3/EVP_PKEY_fromdata.pod
levitte: Done in #11270, that should resolve this particular concern, no?

doc/man3/EVP_PKEY_fromdata.pod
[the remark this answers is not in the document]
levitte: There's an update of the `EVP_PKDY_CTX_new_from_name()` documentation in #11270, would you mind looking there?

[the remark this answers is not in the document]
levitte: Sorry, confused the branches.  I've modified doc/man3/EVP_PKEY_CTX_new.pod as part of *this* PR.  Please have a look.

doc/man3/EVP_PKEY_fromdata.pod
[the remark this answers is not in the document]
levitte: Heh, we shouldn't compare our piles of PRs...  I think I win in backlog size.

But...  it's a fairly minimal change after all, so I guess I can include it here.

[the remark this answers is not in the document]
levitte: Darn.  There are *many* incorrectly documented numbers.  OK, different PR coming up.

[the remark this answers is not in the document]
levitte: See #11275

levitte:
> My point is that this function [EVP_PKEY_param_fromdata_settable()] should not exist.

`openssl provider` and `openssl list` will use this type of function to display what any chosen provider (not just *ours*) can handle.  Also, the array it returns can be use with functions like `OSSL_PARAM_allocate_from_text()`, should anyone want to write a program that can use named parameters (you know, the usual `-pkeyopt key:value` form) to create keys.

We've pretty much made it a pattern that any set_params() or get_params() function should come with a settable() or gettable() function as well.  Do you disagree with that principle, or is it rather that you see reasons not to follow that pattern in *this* case?

I agree that there should still be documentation, though.

levitte:
> Please document that.

Okie, will do

levitte:
> Please provide an other usecase for this function than printing what's already in the documentation.

I think I'd like to defer that to a cookbook kind of thing, which would involve a number of other functions described in other man-pages.  In other words, not in *this* PR

levitte:
Regarding a cookbook thing, which is something I think we *should* do, we need to discuss where and how.  That's a subject for the openssl-project list

paulidale:
> So I still do not get the point of EVP_PKEY_param_fromdata_settable(). Whatever that returns should be in the documentation.

This is up to each provider or algorithm to do.  It returns the list of settable parameters for an algorithm.  Writing much more than that would be difficult -- perhaps some universal parameters could be listed.  "key" for a cipher, "iv" for a mac.  I'm sceptical as to the benefit.

levitte:
"key" for a cipher?

doc/man3/EVP_PKEY_CTX_new.pod
```diff
@@ -52,23 +53,44 @@ If I<ctx> is NULL, nothing is done.
...
+
+=item Name strings
+
+This is the I<name> used with EVP_PKEY_CTX_new_from_name().
+
+These are names like "RSA", "DSA", and what's available depends on what
+providers are currently accessible.
+
+The OpenSSL providers offer a set of key types available this way, please
+see L<OSSL_PROVIDER-FIPS(7)> and L<OSSL_PROVIDER-default(7)> and related
```
kroeckx: The -default doesn't seem to exist

kroeckx: And the -FIPS manpage does not explain anything related to this, it only talks about self-testing.

doc/man3/EVP_PKEY_CTX_new.pod
```diff
@@ -52,23 +53,44 @@ If I<ctx> is NULL, nothing is done.
...
+
+=over 4
+
+=item Legacy NID or EVP_PKEY type
+
+This is the I<id> used with EVP_PKEY_CTX_new_id().
+
+These are B<EVP_PKEY_RSA>, B<EVP_PKEY_RSA_PSS>, B<EVP_PKEY_DSA>,
+B<EVP_PKEY_DH>, B<EVP_PKEY_EC>, B<EVP_PKEY_SM2>, B<EVP_PKEY_X25519>,
+B<EVP_PKEY_X448>, and are used by legacy methods.
```
kroeckx: I'm wondering since when NIDs are legacy, and does that mean the functions making use of an NID are deprecated?

kroeckx:
> `openssl provider` and `openssl list` will use this type of function to display what any chosen provider (not just *ours*) can handle.  Also, the array it returns can be use with functions like `OSSL_PARAM_allocate_from_text()`, should anyone want to write a program that can use named parameters (you know, the usual `-pkeyopt key:value` form) to create keys.
> 
> We've pretty much made it a pattern that any set_params() or get_params() function should come with a settable() or gettable() function as well.  Do you disagree with that principle, or is it rather that you see reasons not to follow that pattern in *this* case?

I have a problem with the pattern in all cases. All the providers
should come with documentation.

I have to think about the OSSL_PARAM_allocate_from_text() case,
but at the moment I don't see it.

doc/man3/EVP_PKEY_CTX_new.pod
levitte: See #11270

doc/man3/EVP_PKEY_CTX_new.pod
levitte: We have treated them as legacy for a long time now.  If you look at the `crypto/` code and search for `legacy:`, you'll find that it's all about old code that use NIDs as identifiers, while the provider based code do not use NIDs at all.

Note, however, that it's perfectly OK to use the constant NIDs we have _internally_.  We do that all over the place in our providers, quite visibly so in `providers/implementations/signature/`.  On the other hand, no NID is knowingly passing the libcrypto\<-\>provider interface, because that's a public interface.

levitte:
> All the providers should come with documentation.

No one says anything else.  #11270 is an attempt in that direction.

However, I don't see why documentation and the settable/gettable functions would be mutually exclusive.  As a matter of fact, we could use those functions via some commands to see that our implementations and the docs are in sync.

levitte:
Where are we on this one?

levitte:
PING!

doc/man3/EVP_PKEY_fromdata.pod
bernd-edlinger: @mattcaswell is your concern resolved?

doc/man3/EVP_PKEY_fromdata.pod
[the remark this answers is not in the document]
bernd-edlinger: @kroeckx are you satisfied with Richard's doc/man3/EVP_PKEY_CTX_new.pod ?

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -50,9 +61,156 @@ EVP_PKEY_fromdata() return 1 for success and 0 or a negative value for
...
+     
+ int main()
+ {
+     EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_from_name(NULL, "RSA", NULL);
+     EVP_PKEY *pkey = NULL;
+
+     if (ctx == NULL
+         || !EVP_PKEY_key_fromdata_init(ctx)
+         || !EVP_PKEY_fromdata(ctx, &pkey, params))
+         exit(1);
```
bernd-edlinger, line 105: by indenting the || above the exit is more visual, that was Andy's reasoning for the style.
Would you mind adding extra indents here, @levitte ?

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -50,9 +61,156 @@ EVP_PKEY_fromdata() return 1 for success and 0 or a negative value for
...
+     0xeb, 0x94, 0x6f, 0x97, 0xdb, 0xa3, 0x7d, 0xbd,
+     0xe5, 0x26, 0xca, 0x07, 0x17, 0x8d, 0x26, 0x75,
+     0xff, 0xcb, 0x8e, 0xb6, 0x84, 0xd0, 0x24, 0x02,
+     0x25, 0x8f, 0xb9, 0x33, 0x6e, 0xcf, 0x12, 0x16,
+     0x2f, 0x5c, 0xcd, 0x86, 0x71, 0xa8, 0xbf, 0x1a,
+     0x47
+ };
+ const OSSL_PARAM params[] = {
+     OSSL_PARAM_utf8_string("curve-name", "prime256v1"),
+     OSSL_PARAM_BN("priv", priv, sizeof(priv));
```
bernd-edlinger, line 143: semicolon here is that right?
did you compiler this sample?

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -50,9 +61,156 @@ EVP_PKEY_fromdata() return 1 for success and 0 or a negative value for
...
+
+ int main()
+ {
+     EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_from_name(NULL, "EC", NULL);
+     EVP_PKEY *pkey = NULL;
+
+     if (ctx == NULL
+         || !EVP_PKEY_key_fromdata_init(ctx)
+         || !EVP_PKEY_fromdata(ctx, &pkey, params))
+         exit(1);
```
bernd-edlinger, line 156: again, I did understand Andy in this point.  I took his idea and made it mine.

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -50,9 +61,156 @@ EVP_PKEY_fromdata() return 1 for success and 0 or a negative value for
...
+         printf("%s : %s ", settable_params->key, datatype);
+         if (settable_params->data_size == 0)
+             printf("(unlimited size)");
+         else
+             printf("(maximum size %zu)", settable_params->data_size);
+     }
+ }
+
+The descriptor L<OSSL_PARAM(3)> returned by
+EVP_PKEY_key_fromdata_settable() may also be used programatically, for
```
bernd-edlinger, line 207: programatically is't that derived from programming , thus double-m ?

doc/man3/EVP_PKEY_fromdata.pod
levitte: It wasn't Andy's idea. And yes I do mind, as it means I must either battle emacs' cc-mode indentation, or turn it off and then waste time indenting manually.

doc/man3/EVP_PKEY_fromdata.pod
levitte: Thanks.  Fixed

doc/man3/EVP_PKEY_fromdata.pod
levitte: Fixed.  Thanks

doc/man3/EVP_PKEY_fromdata.pod
levitte: Or oh wait...  in manual, yeahok, I have to do it manually anyway.

Mind you, though, that I find that style jarring, and also that even those who do this don't do it consistently, i.e. they do after `if` but not after `else if`

mspncp: > And yes I do mind, as it means I must either battle emacs' cc-mode indentation, or turn it off and then waste time indenting manually.

Same holds for me. Also, I think that the `|| ` prefix already adds sufficient extra indentation to the followup conditions.

bernd-edlinger: > it means I must either battle emacs' cc-mode indentation, or turn it off and then waste time indenting manually.

I am afraid you are gonna lose this battle...

bernd-edlinger: I think @DDvO pointed aout an Emacs solution for this problem, though.

doc/man3/EVP_PKEY_fromdata.pod
bernd-edlinger: and once, again, when I contribute to linux I use a totally different style
and follow that as pedantically as you know me already.
Also gcc and (i am still learning gdb) have their own also special style.
Which has nothing to do with the way I would write my programs
when I start a totally new project.

richsalz:
The cost of one person adding an extra four spaces, compared to the global cost of inconsistency and possible confusion.

doc/man3/EVP_PKEY_fromdata.pod
levitte: Reference please.  I haven't seen that solution.  I have said a few times that if someone comes up with a viable emacs solution, I will consider it.

doc/man3/EVP_PKEY_fromdata.pod
levitte: The extra indent of `if` condition expressions isn't mandated in our style.  We *allow it*, but the idea is so far ad-hoc, and has been agreed upon, because issues.

doc/man3/EVP_PKEY_fromdata.pod
levitte: I looked around in other manuals where we do have examples with multiline `if` conditions, and we don't have extra indents in any of them.

We hopefully write manuals _for others_, more than for ourselves, so I think it's wiser to use a style that's commonly reachable rather than one that requires some extra hackery to follow, be it in code or in mind.  So in the end, I'm not keen in making the change proposed here.

doc/man3/EVP_PKEY_fromdata.pod
mattcaswell: Yes

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -50,9 +61,156 @@ EVP_PKEY_fromdata() return 1 for success and 0 or a negative value for
 failure.  In particular a return value of -2 indicates the operation is
 not supported by the public key algorithm.
 
+=head1 EXAMPLES
+
+These examples are very terse for the sake of staying on topic, which
+are the EVP_PKEY_fromdata() set of functions.  In real applications,
```
paulidale, line 67: `s/are/is/`

"set" is singular.

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -50,9 +61,156 @@ EVP_PKEY_fromdata() return 1 for success and 0 or a negative value for
...
+BIGNUMs would be handled, and converted to byte arrays with
```
paulidale, line 68: Probably could drop the comma here.

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -50,9 +61,156 @@ EVP_PKEY_fromdata() return 1 for success and 0 or a negative value for
...
+
+=head2 Creating an ECC keypair using raw key data
+
+ #include <openssl/evp.h>
+
+ /*
+  * These arrays represent large numbers, big endian organization.
+  * In a real application, these would probably be bignums that get
+  * converted to the native integer organization with BN_bn2nativepad().
+  * We're not doing that here, since this is not an of BIGNUM
```
paulidale, line 118: this is not an _example_ of BIGNUM

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -50,9 +61,156 @@ EVP_PKEY_fromdata() return 1 for success and 0 or a negative value for
...
+ }
+
+=head2 Finding out params for an unknown key type
+
+ #include <openssl/evp.h>
+
+ /* Program expects a key type as first argument */
+ int main(int argc, char *argv[])
+ {
+     EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_from_name(NULL, argv[0], NULL);
```
paulidale, line 168: `argv[0]`???  I'm pretty sure this should be `argv[1]`.

doc/man3/EVP_PKEY_fromdata.pod
levitte: Fixed

doc/man3/EVP_PKEY_fromdata.pod
levitte: Fixed

doc/man3/EVP_PKEY_fromdata.pod
levitte: Fixed

doc/man3/EVP_PKEY_fromdata.pod
levitte: You're right.  Fixed.

doc/man3/EVP_PKEY_fromdata.pod
bernd-edlinger: @levitte have you considered https://github.com/openssl/openssl/issues/10725#issuecomment-585675825
does that resolve your concerns?

Anyway this unresolved conflict starts to poison the whole openssl project.
I do not really understand why this fight is necessay, without the
constant symmetry breaks in the coding style we could concentrate
to other matters, like the issue I raised on secrity@openssl.org two
days ago, and nobody took it serious... yet I believe that is a major
security issue.

levitte:
Merged.

476de2e5e598b46289c6087df04d1bc1886d6ef2 DOC: Add more description of EVP_PKEY_fromdata(), and examples
ccb47dbf47b36da849301aae38c32ed019e75a21 DOC: Extend the description of EVP_PKEY_CTX_new_from_name()

PR #11222, opened 2020-03-02
.gitignore: Add /apps/progs.{c,h}

CyberShadow:
These files were removed from the source tree in
fe909ee4aeb6eb64f6f31a1544c5d3c81c5fe1f1.

CLA: trivial

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -123,6 +123,8 @@ doc/man1/openssl-x509.pod
 /TAGS
 *.map
 *.ld
+/apps/progs.c
+/apps/progs.h
 
 # Windows (legacy)
 /tmp32


paulidale:
These files aren't built by the normal make process, they require a `make update`.  Having git ignore these files means that they need to be constructed as part of the make process.

CyberShadow:
> These files aren't built by the normal make process, they require a `make update`.

NAK:

```console
$ cd ~/work/extern/openssl
$ git checkout master
[...]
$ git clean -fdx
[...]
$ ./config && make -j`nproc`
[...]
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	apps/progs.c
	apps/progs.h

nothing added to commit but untracked files present (use "git add" to track)
```

>  Having git ignore these files means that they need to be constructed as part of the make process.

Sorry, I don't understand what you mean by this.

Generally, `.gitignore` files should be set up so that `git status` does not show any uncommited changes when there are no changes that ought to be committed.

levitte:
@paulidale uttered thusly:
>These files aren't built by the normal make process, they require a `make update`

Nope, not any more.  They are now generated during build.

https://github.com/openssl/openssl/blob/97ace6c2dae451ce8e3b099cf242968ecff128af/apps/build.info#L33-L38

levitte:
Trivial

slontis:
I am not waiting any longer to merge this one - its really annoying when you are constantly switching branches.

slontis:
Thanks @CyberShadow. Merged to master.

richsalz:
For the next time this happens :(, .git/info/exclude is a local .gitignore

PR #11224, opened 2020-03-02
spkac: Check return values of NETSCAPE_SPKI functions

CyberShadow:
Fixes silently producing an invalid SPKAC with non-RSA keys.

I'm not sure if non-RSA keys are expected to work.

Does this need a test?

##### Checklist

### DIFF
--- a/apps/spkac.c
+++ b/apps/spkac.c
@@ -138,8 +138,15 @@ int spkac_main(int argc, char **argv)
         if (challenge != NULL)
             ASN1_STRING_set(spki->spkac->challenge,
                             challenge, (int)strlen(challenge));
-        NETSCAPE_SPKI_set_pubkey(spki, pkey);
-        NETSCAPE_SPKI_sign(spki, pkey, EVP_md5());
+        if (!NETSCAPE_SPKI_set_pubkey(spki, pkey)) {
+            BIO_printf(bio_err, "Error setting public key\n");
+            goto end;
+        }
+        i = NETSCAPE_SPKI_sign(spki, pkey, EVP_md5());
+        if (i <= 0) {
+            BIO_printf(bio_err, "Error signing SPKAC\n");
+            goto end;
+        }
         spkstr = NETSCAPE_SPKI_b64_encode(spki);
         if (spkstr == NULL)
             goto end;


CyberShadow:
Does this need the CLA? If yes, can I sign it electronically? I don't have a printer or scanner.

apps/spkac.c
```diff
@@ -138,8 +138,16 @@ int spkac_main(int argc, char **argv)
         if (challenge != NULL)
             ASN1_STRING_set(spki->spkac->challenge,
                             challenge, (int)strlen(challenge));
-        NETSCAPE_SPKI_set_pubkey(spki, pkey);
-        NETSCAPE_SPKI_sign(spki, pkey, EVP_md5());
+        i = NETSCAPE_SPKI_set_pubkey(spki, pkey);
+        if (i <= 0) {
```
mattcaswell: `NETSCAPE_SPKI_set_pubkey` only ever returns 0 or 1, so this could just be written as:

````
if (!NETSCAPE_SPKI_set_pubkey(spki, pkey))
````

**Comment by CyberShadow** [comment]
> > Does this need the CLA?
> 
> Yes

OK, sent.

> `NETSCAPE_SPKI_set_pubkey` only ever returns 0 or 1, so this could just be written as:

Indeed, done.

> I'm not sure if non-RSA keys are expected to work.

I had a look at what happens with an ED25519 key. From what I understand, OpenSSL loads it as a legacy key, and then, when it tries to export it (`evp_pkey_export_to_provider`), that fails because the key object doesn't have a dirty counter / export function. I'm not sure why that is. Is that functionality not implemented?

**Comment by mattcaswell** [comment]
> I had a look at what happens with an ED25519 key. From what I understand, OpenSSL loads it as a legacy key, and then, when it tries to export it (evp_pkey_export_to_provider), that fails because the key object doesn't have a dirty counter / export function. I'm not sure why that is. Is that functionality not implemented?

This is work in progress (it's on my "to do" list to convert Ed25519 keys for the new provider architecture). However, it *should* just fallback to legacy handling if that fails. Does it not do that?

**Comment by mattcaswell** [comment]
Approved - needs second review.

File: apps/spkac.c [hunk 2]
```diff
@@ -138,8 +138,15 @@ int spkac_main(int argc, char **argv)
         if (challenge != NULL)
             ASN1_STRING_set(spki->spkac->challenge,
                             challenge, (int)strlen(challenge));
-        NETSCAPE_SPKI_set_pubkey(spki, pkey);
-        NETSCAPE_SPKI_sign(spki, pkey, EVP_md5());
+        if (!NETSCAPE_SPKI_set_pubkey(spki, pkey)) {
```
beldmit: Shouldn't be ```NETSCAPE_SPKI_set_pubkey(spki, pkey) == 0```?

mattcaswell: No - the return is a boolean (only ever 0 or 1), so treating as a boolean with `!` is fine

CyberShadow:
> This is work in progress (it's on my "to do" list to convert Ed25519 keys for the new provider architecture). However, it _should_ just fallback to legacy handling if that fails. Does it not do that?

It does fall back to the legacy path in `do_sigver_init`, however, that also fails (because `EVP_PKEY_CTX_ctrl` fails). I will have a closer look with an interactive debugger to see why exactly.

apps/spkac.c
beldmit: Got it, thanks!

CyberShadow:
> I will have a closer look with an interactive debugger to see why exactly.

OpenSSL does not seem to implement signing messages with Ed25519 keys:

https://github.com/openssl/openssl/blob/86fad9e730b266b967f3039ae4c82e4dcbebecba/crypto/ec/ecx_meth.c#L859-L864

Perhaps I'm missing some basic understanding of how signing works in OpenSSL or in general to make sense of this. In any case, this PR is still OK regardless of the above.

beldmit:
Merged. Thanks!

PR #11213, opened 2020-03-01
doc: document that 'openssl rand' is cryptographically secure

mspncp:
Alternative to #11202.

PR #11231, opened 2020-03-03
Correct two small documentation issues

t8m:
The find-doc-nits complains about non-zero word and about missing
line before =head1 which causes build failure.

##### Checklist


beldmit:
Approved if travis is happy.

t8m:
Merged to master as 4a98ad086a

mspncp:
Looks like this slipped in from #11213. Thanks for fixing it, @t8m!

PR #11239, opened 2020-03-04
 Add default property API's to enable and test for fips

slontis:
##### Checklist


crypto/evp/evp_fetch.c
```diff
@@ -325,10 +325,42 @@ int EVP_set_default_properties(OPENSSL_CTX *libctx, const char *propq)
...
+}
+
+int EVP_default_properties_is_fips_enabled(OPENSSL_CTX *libctx)
+{
+    return evp_default_property_is_enabled(libctx, "fips");
+}
+
+int EVP_default_properties_enable_fips(OPENSSL_CTX *libctx, int enable)
+{
+    const char *query = (enable != 0) ? "fips=yes" : "fips=no";
```
paulidale: Instead of `fips=no`, I think it should be `-fips`.

crypto/property/property_parse.c
```diff
@@ -453,6 +453,33 @@ int ossl_property_has_optional(const OSSL_PROPERTY_LIST *query)
...
+                             const OSSL_PROPERTY_LIST *prop_list)
+{
+    int i;
+    OSSL_PROPERTY_IDX name_id;
+    const PROPERTY_DEFINITION *prop = NULL;
+
+    if (prop_list == NULL)
+        return 0;
+
+    if (!parse_name(ctx, &property_name, 0,&name_id))
```
paulidale: Space between _,_ and _&_.

crypto/property/property_parse.c
```diff
@@ -453,6 +453,33 @@ int ossl_property_has_optional(const OSSL_PROPERTY_LIST *query)
...
+
+    prop = prop_list->properties;
+    for (i = 0; i < prop_list->n; ++i) {
+        if (prop[i].name_idx == name_id) {
+            return prop[i].optional
+                   || (prop[i].type == PROPERTY_TYPE_STRING
+                       && ((prop[i].oper == PROPERTY_OPER_EQ
+                                && prop[i].v.str_val == ossl_property_true)
+                           || (prop[i].oper == PROPERTY_OPER_NE
+                                && prop[i].v.str_val == ossl_property_false)));
```
paulidale: I think that this would be better as `prop[i].v.str_val != ossl_property_true`

crypto/property/property_parse.c
```diff
@@ -453,6 +453,33 @@ int ossl_property_has_optional(const OSSL_PROPERTY_LIST *query)
...
+    if (prop_list == NULL)
+        return 0;
+
+    if (!parse_name(ctx, &property_name, 0,&name_id))
+        return 0;
+
+    prop = prop_list->properties;
+    for (i = 0; i < prop_list->n; ++i) {
+        if (prop[i].name_idx == name_id) {
+            return prop[i].optional
```
paulidale: optional is a match?  [see note below]

crypto/property/property_parse.c
```diff
@@ -453,6 +453,33 @@ int ossl_property_has_optional(const OSSL_PROPERTY_LIST *query)
...
+                   || (prop[i].type == PROPERTY_TYPE_STRING
+                       && ((prop[i].oper == PROPERTY_OPER_EQ
+                                && prop[i].v.str_val == ossl_property_true)
+                           || (prop[i].oper == PROPERTY_OPER_NE
+                                && prop[i].v.str_val == ossl_property_false)));
+        }
+    }
+    return 0;
```
paulidale: If the property isn't mentioned in the query, should the return be 1 ?  Above at the [see note below], an optional match (either for or against) is treated as matching.  What's the difference between that case and this?

doc/man3/EVP_set_default_properties.pod
```diff
@@ -27,10 +30,22 @@ given with I<libctx> (NULL signifies the default library context).
 Any previous default property for the specified library context will
 be dropped.
 
+EVP_default_properties_enable_fips() sets the 'fips=yes' to be a default property
+if I<enable> is non zero, otherwise it clears 'fips=yes' as a default property
```
paulidale: `s/as a/from the/`
`s/$/query/`

Note: the code above does **not** clear the 'fips=yes' setting, it adds 'fips=no' which is quite different.

doc/man3/EVP_set_default_properties.pod
```diff
@@ -27,10 +30,22 @@ given with I<libctx> (NULL signifies the default library context).
...
+for the given I<libctx>. It merges the fips default property setting with any
```
paulidale: `s/setting/query/`

doc/man3/EVP_set_default_properties.pod
```diff
@@ -27,10 +30,22 @@ given with I<libctx> (NULL signifies the default library context).
...
+existing query strings that have been set via EVP_set_default_properties().
```
paulidale: Is it worth a note indicating that the change replaces any 'fips' query element that was there or is this sufficiently implied?

doc/man3/EVP_set_default_properties.pod
```diff
@@ -27,10 +30,22 @@ given with I<libctx> (NULL signifies the default library context).
...
+
+EVP_default_properties_is_fips_enabled() indicates if 'fips=yes' is a default
+property for the given I<libctx>.
```
paulidale: `s/for/query for/`

It isn't doing this, it also accepts `?fips=no`.

test/property_test.c
```diff
@@ -383,6 +384,34 @@ static int test_query_cache_stochastic(void)
...
+static int test_fips_mode(void)
+{
+    int ret = 0;
+    OPENSSL_CTX *ctx = NULL;
+
+    if (!TEST_ptr(ctx = OPENSSL_CTX_new()))
+        goto err;
+
+    ret = TEST_true(EVP_set_default_properties(ctx, "default=yes,fips=yes"))
+          && TEST_int_eq(EVP_default_properties_is_fips_enabled(ctx), 1)
```
paulidale: Why `TEST_int_eq` instead of `TEST_true`?

test/property_test.c
```diff
@@ -383,6 +384,34 @@ static int test_query_cache_stochastic(void)
...
+          && TEST_true(EVP_set_default_properties(ctx, "fips=no,default=yes"))
+          && TEST_int_eq(EVP_default_properties_is_fips_enabled(ctx), 0)
```
paulidale: Ditto but using `TEST_false`.

crypto/evp/evp_fetch.c
slontis: it seems to work how it is - what it the difference?

slontis:
Does anyone have any alternate names for the API's that would be better?

doc/man3/EVP_set_default_properties.pod
slontis: I thought that was what merges implies

crypto/evp/evp_fetch.c
levitte: If I understand correctly, "fips=no" will make all fips algos inaccessible, while "-fips" will make them accessible, just not mandatory (i.e "-fips" actually removes the property in a merge).

crypto/property/property_parse.c
```diff
@@ -453,6 +453,33 @@ int ossl_property_has_optional(const OSSL_PROPERTY_LIST *query)
...
+    const PROPERTY_DEFINITION *prop = NULL;
+
+    if (prop_list == NULL)
+        return 0;
+
+    if (!parse_name(ctx, &property_name, 0, &name_id))
+        return 0;
+
+    prop = prop_list->properties;
+    for (i = 0; i < prop_list->n; ++i) {
```
levitte: In crypto/serializer/serializer_pkey.c, there's a very simple parameter check, using `ossl_parse_query` and `ossl_property_match_count`. Why do you need a specially coded loop to do something similar?

crypto/evp/evp_fetch.c
slontis: It doesnt quite 'remove' the property.. but it does tag it as a different type.

slontis: Updated to use -

crypto/property/property_parse.c
slontis: Changed to !optional && ()

doc/man3/EVP_set_default_properties.pod
slontis: Its good enough..

crypto/property/property_parse.c
slontis: I tried this initially and it failed.. match_count returns 1 for optionals..

paulidale: If "?fips=yes" is in the default query, `ossl_property_match_count` will return 1 but FIPS mode isn't mandated.

Additionally, the matching code isn't designed to match a query against a query -- there are cases where it won't work sensibly.  I should have used two named types PROPERTY_DEFINITION_LIST and PROPERTY_QUERY_LIST to highlight the difference.

test/property_test.c
```diff
@@ -383,6 +384,34 @@ static int test_query_cache_stochastic(void)
...
+          && TEST_int_eq(EVP_default_properties_is_fips_enabled(ctx), 1)
+          && TEST_true(EVP_set_default_properties(ctx, "fips=no"))
+          && TEST_int_eq(EVP_default_properties_is_fips_enabled(ctx), 0)
+          && TEST_true(EVP_set_default_properties(ctx, "fips=no,default=yes"))
+          && TEST_true(EVP_default_properties_enable_fips(ctx, 1)
+          && TEST_int_eq(EVP_default_properties_is_fips_enabled(ctx), 1)
+          && TEST_true(EVP_default_properties_enable_fips(ctx, 0))
+          && TEST_int_eq(EVP_default_properties_is_fips_enabled(ctx), 0));
+err:
+    return ret;
```
slontis: @paulidale - you were so busy niting me that you didnt see the memory leak :)

paulidale: Meh, it's a test and automated tools are better at picking these up.

PS: they weren't nits for the most part but logic problems :)

crypto/property/property_parse.c
levitte: Ah, gotcha. In the serializer case, I'm matching against a property definition.
OK then.

crypto/property/property_parse.c
slontis: Ok this change combined with the -fips setting resulted in a nasty side effect..
It now also needs to check for override since the type is not set for these - and will result in uninitialized reads if accessed.

crypto/evp/evp_cnf.c
```diff
@@ -44,14 +44,11 @@ static int alg_module_init(CONF_IMODULE *md, const CONF *cnf)
...
-             * have this, so we assume that no default properties have
-             * been set before this.
+             * configurations.
              */
-            if (m > 0)
-                EVP_set_default_properties(NULL, "fips=yes");
+            EVP_default_properties_enable_fips(cnf->libctx, m > 0);
         } else if (strcmp(oval->name, "default_properties") == 0) {
-            EVP_set_default_properties(NULL, oval->value);
+            EVP_set_default_properties(cnf->libctx, oval->value);
```
mattcaswell: Should we be handling the return value (here and above)?

slontis: Ok changed..

t8m:
Could this be merged soon?  Please rebase. As for alternative names of the public functions I could propose `EVP_default_properties_has_fips` and `EVP_default_properties_set_fips` but the current names are OK as well.

slontis:
Rebased  : squashed and  updated generated files only.

crypto/property/property_parse.c
```diff
@@ -453,6 +453,35 @@ int ossl_property_has_optional(const OSSL_PROPERTY_LIST *query)
...
+        if (prop[i].name_idx == name_id) {
+            /* Do a seperate check for override as it does not set type */
```
t8m, line 472: Nit typo: separate

slontis:
Thanks. Merged to master.

FdaSilvaYY:
This merge breaks the build :
```
$ if ! $make update; then echo -e '\052\052 FAILED -- MAKE UPDATE'; travis_terminate 1; fi
EVP reason code 208 collision at EVP_R_SET_DEFAULT_PROPERTY_FAILURE
Makefile:2894: recipe for target 'errors' failed

```

bernd-edlinger:
I don't know why I am not surprised...

PR #11240, opened 2020-03-04
 Add support for passing the libctx to the config loader

slontis:
The self tests for the fips module are triggered on startup and they need to know the
core's libctx in order to function correctly. As the provider can be autoloaded via configuration
it then needs to propagate the callers libctx down to the provider via the config load.
    
Note that OPENSSL_init_crypto(OPENSSL_INIT_LOAD_CONFIG, ..) is still called, but will only load the default configuration if the OPENSSL_CONF environment variable is set.

##### Checklist


slontis:
travis failure is a timeout.

slontis:
@richsalz I think you are working on getting some fips tests running - so maybe this is useful?

richsalz:
I am reviewing run_tests and friends so that everything can be run with one or more providers.  The plan is to ".include" a FIPS module

Leo-Neat:
Note that CIFuzz found a bug in this branch request: 
https://github.com/openssl/openssl/runs/511776434?check_suite_focus=true

It did not report report it because it is in dry_run mode. I will submit a PR to change OpenSSL out of dry_run mode to assure that bugs like this are reported.

crypto/context.c
```diff
@@ -145,6 +146,16 @@ OPENSSL_CTX *OPENSSL_CTX_new(void)
     return ctx;
 }
 
+#ifndef FIPS_MODE
+int OPENSSL_CTX_load_config(OPENSSL_CTX *ctx, const char *config_file)
+{
+    if (!OPENSSL_init_crypto(OPENSSL_INIT_LOAD_CONFIG, NULL))
+        return 0;
```
mattcaswell: Why is this needed? It seems unnecessary?

crypto/conf/conf_mod.c
```diff
@@ -152,6 +155,25 @@ int CONF_modules_load_file(const char *filename, const char *appname,
...
+                           const char *appname, unsigned long flags)
+{
+    char *file = (char *)filename;
+    int ret = 0;
+
+    if (file == NULL) {
+        /* If we specify a default configuration to load try to load it */
+        file = CONF_get1_default_config_file();
+        /* dont fail if there is no default though */
+        if (file == NULL)
```
mattcaswell: Why do we need to do this? It seems to change the old behaviour. Previously, AFAICT, we only succeeded if the CONF_MFLAGS_IGNORE_RETURN_CODES flag was set. Now we succeed unconditionally. Why not just defer all of this to CONF_modules_load_file_with_libctx, and if the flag is set it succeeds?

mattcaswell:
> Note that CIFuzz found a bug in this branch request(under the run_fuzzers tab):

It looks to me like this issue is not related to this PR, but maybe pre-existing.

crypto/context.c
slontis: Because there are dependancies which we want to ensure are called once before loading..
i.e: openssl_config_int() calls OPENSSL_load_builtin_modules()

Take out the line and run evp_fetch_prov_test to see this in action.

crypto/conf/conf_mod.c
slontis: This is related to your other question below ...
OPENSSL_init_crypto(OPENSSL_INIT_LOAD_CONFIG) is needed to be called once..
which then calls openssl_config_int() which then calls CONF_modules_load_file().. So I am trying to deal with this case..
So maybe I need instead to do this code as a seperate copy that gets called inside the CONF_modules_load_file().

slontis: Note that the way I have changed the code - this call it will only load the default configuration if the environment variable OPENSSL_CONF is set. This means that all existing test code works how it did before without breaking (since OPENSSL_CONF is used by quite a few tests).

crypto/context.c
mattcaswell: Ah - right. Thanks, I understand now. However, I think this is misplaced. We should be making these init calls "just in time" IMO. So in this case the problem is we are trying to use built-in modules before they have been loaded. Therefore the function where we try to get the reference to the module should be where we do this just-in-time initialisation, i.e.

```` diff
diff --git a/crypto/conf/conf_mod.c b/crypto/conf/conf_mod.c
index 21d3b009bf..5c781ed33b 100644
--- a/crypto/conf/conf_mod.c
+++ b/crypto/conf/conf_mod.c
@@ -180,6 +180,12 @@ static int module_run(const CONF *cnf, const char *name, const char *value,
     CONF_MODULE *md;
     int ret;
 
+    /* Make sure any built-in modules are loaded before we try to find them */
+    if (!OPENSSL_init_crypto(OPENSSL_INIT_LOAD_CONFIG, NULL)) {
+        CONFerr(CONF_F_MODULE_RUN, CONF_R_MODULE_INITIALIZATION_ERROR);
+        return -1;
+    }
+
     md = module_find(name);
 
     /* Module not found: try to load DSO */
````

crypto/conf/conf_mod.c
mattcaswell: Sorry, I'm still not getting it...

> calls openssl_config_int() which then calls CONF_modules_load_file().

But `openssl_config_int()` calls `CONF_modules_load_file()` with the flags `DEFAULT_CONF_MFLAGS` which includes `CONF_MFLAGS_IGNORE_RETURN_CODES`. So this would mean `CONF_modules_load_file_with_libctx` shouldn't fail if there isn't a default config file anyway?

crypto/context.c
slontis: That is not going to work.
The OPENSSL_init_crypto() calls 
CONF_modules_load_file() which will call OPENSSL_init_crypto() - which then causes a deadlock.

crypto/conf/conf_mod.c
slontis: See comment about deadlock above, its related to the same problem..

slontis: This lazy init stuff is just plain nasty :)

crypto/context.c
mattcaswell: Good point. This actually implies to me that openssl_config_int() should not call OPENSSL_load_builtin_modules() at all. We could then just call OPENSSL_load_builtin_modules() from a RUNONCE in module_run.

slontis: Ahh ok that seems to work - glad you know what you are doing :)
That seems to clear up the dodgy bits of this PR that made me feel a bit dirty when I added them :)

crypto/conf/conf_mod.c
slontis: Ok so this function can now just call the libctx variation now that the dependancies are fixed up.

slontis:
Thanks. Merged to master.

PR #11246, opened 2020-03-04
This works around a gcc-9 crash

bernd-edlinger:
It happens when configured with ./config -ftest-coverage

see: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=94029

Master is not affected by this issue.

### DIFF
--- a/crypto/pem/pem_all.c
+++ b/crypto/pem/pem_all.c
@@ -85,8 +85,7 @@ IMPLEMENT_PEM_write_cb_const(RSAPrivateKey, RSA, PEM_STRING_RSA,
 
 IMPLEMENT_PEM_rw_const(RSAPublicKey, RSA, PEM_STRING_RSA_PUBLIC,
                        RSAPublicKey)
-    IMPLEMENT_PEM_rw(RSA_PUBKEY, RSA, PEM_STRING_PUBLIC,
-                     RSA_PUBKEY)
+IMPLEMENT_PEM_rw(RSA_PUBKEY, RSA, PEM_STRING_PUBLIC, RSA_PUBKEY)
 #endif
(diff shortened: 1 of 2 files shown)


t8m:
LGTM

crypto/pem/pem_all.c
```diff
@@ -84,9 +84,9 @@ IMPLEMENT_PEM_write_cb_const(RSAPrivateKey, RSA, PEM_STRING_RSA,
 
 
 IMPLEMENT_PEM_rw_const(RSAPublicKey, RSA, PEM_STRING_RSA_PUBLIC,
-                       RSAPublicKey) IMPLEMENT_PEM_rw(RSA_PUBKEY, RSA,
-                                                      PEM_STRING_PUBLIC,
-                                                      RSA_PUBKEY)
+                       RSAPublicKey)
+    IMPLEMENT_PEM_rw(RSA_PUBKEY, RSA, PEM_STRING_PUBLIC,
+                     RSA_PUBKEY)
```
levitte: It's still too indented...  it should be flush to the left

bernd-edlinger: the same funny indentation is in other places in this file too,
shall I fix them as well?

levitte: Please do

bernd-edlinger:
The unusual formatting triggered an assertion, about the line numbers looked bogus.

bernd-edlinger:
... and of course "line coverage" is pretty pointless, when we define multiple functions
on one macro in one line.
One might say these are a headache to debug as well.

t8m:
Still approved

bernd-edlinger:
Okay, merged to OpenSSL_1_1_1-stable bf42b99
Thanks!

Issue #10357, opened 2019-11-04
Move SM2 digest signature implementation to provider.


paulidale:
Move the SM2 digest signature code to `providers/implementation/signature`. The existing `dsa.c` file should give a general idea as to the bulk of the changes required.

This will probably need #10351 done first.

InfoHunter:
I will take this and #10358 after the EC-in-privoder part has been done

romen:
Fixed with #12536 (formerly #11248).

PR #11248, opened 2020-03-04
Add SM2 signature algorithm to default provider

InfoHunter:
Addressing #10357

This is still in a very early stage but the skeleton is almost done. I have several unclear things that need answers from the team:

1. ~how to pass values associated with the EVP_PKEY_CTX to the provider? (the SM2 identifier)~
2. ~how to turn the OID into DER format for the algorithm identifiers?~
3. I assume that the EC key management can be reused in SM2 case, or should I create a new SM2 keymgmt?

##### Checklist


InfoHunter:
Perhaps I should skip CIs at current stage...

providers/defltprov.c
```diff
@@ -388,6 +388,9 @@ static const OSSL_ALGORITHM deflt_signature[] = {
 #ifndef OPENSSL_NO_EC
     { "ED25519:Ed25519", "provider=default", ed25519_signature_functions },
     { "ED448:Ed448", "provider=default", ed448_signature_functions },
+# ifndef OPENSSL_NO_SM2
+    { "SM2:id-ecPublicKey", "provider=default", sm2_signature_functions },
```
levitte: Don't include id-ecPublicKey here.  That will pile it together with ECDH, and I'm pretty sure you don't want that.

levitte: (as a matter of fact, I'm not at all sure id-ecPublicKey should be an alias for ECDH either, but that's a separate issue)

InfoHunter: I don't think id-ecPublicKey should be the alias.

Leo-Neat:
Note that CIFuzz found a bug in this pull request. Please check the logs for details.

Leo-Neat:
Note that CIFuzz found a bug in this Pull request. It was not reported because CIfuzz was in dry_mode.

providers/implementations/signature/sm2sig.c
```diff
@@ -0,0 +1,472 @@
...
+        return EVP_MD_size(psm2ctx->md);
+    return 0;
+}
+#endif
+
+/* it seems SM2 doesn't need this */
+static int sm2sig_get_md_nid(const EVP_MD *md)
+{
+    /*
+     * Because the DSA library deals with NIDs, we need to translate.
```
slontis: DSA library?

InfoHunter: Ahh, a copy paste error...

InfoHunter:
> Note that CIFuzz found a bug in this Pull request. It was not reported because CIfuzz was in dry_mode.

Thanks and this PR is still not finished let's see if that exist after more commits are pushed.

InfoHunter:
> how to pass values associated with the EVP_PKEY_CTX to the provider? (the SM2 identifier)

Since PR #11302 has been merged so I don't think this is a problem anymore

levitte:
@InfoHunter said this in https://github.com/openssl/openssl/pull/11302#issuecomment-599492733:
> > Since PR #11302 has been merged so I don't think this is a problem anymore
> 
> I had a re-thought on this. It seems current design doesn't allow passing the dist-ID into a provider, so the user can only set the dist-ID in advance and then the calculation of the dist-ID is computed in the update process inside a provider. Is this what you originally intended? @levitte

Not sure I understand the issue.  The SM2 ID is set with `EVP_PKEY_CTX_set1_id()`, and can be used *after* `EVP_DigestSignInit()` since #11302 was merged in.  That's the intention.  It mist still, however, be used before `EVP_DigestSignUpdate()` and `EVP_DigestSignFinal()` is called.  Does that make sense?  Have you found a place where this doesn't quite work out?

InfoHunter:
Well, the calling sequence you mentioned requires the caller to set the ID by using `EVP_PKEY_CTX_set1_id()`, and currently if the ID is not set then the call will fail. I was wondering if some implementations (provided by providers) of SM2 would like to handle the ID set by the user *inside* the provider - for instance to adjust the value of the ID. I am not pretty sure about this use case, but just found that the ID is transparent to the provider since it's calculated outside.  Should this calculation be part of a provider implementation? It seems there is no way to intercept the EVP_PKEY_ctrl or custom_digest stuffs in a provider, right?

levitte:
You're right that `EVP_PKEY_CTX_set1_id()` currently only works with legacy implementations.  To work with providers, it needs to be rewritten in the same way that, for example, `EVP_PKEY_CTX_set_ecdh_cofactor_mode()` was rewritten.

The location for the EC control reimplementations is `crypto/ec/ec_evp_lib.c`

InfoHunter:
Actually I have some new ideas about the use case of handling the ID stuffs. It might be possible that some day in the future the compliance requirement changes to some new policies, say, the ID is required to less than 4 bytes..., so one can implement a 4-byte-ID-SM2 provider to comply with it... ;-)

InfoHunter:
But this is low priority - need to get this PR done first

InfoHunter:
Travis-CI failure is not relevant (timeout), considering to remove the `WIP` prologue

crypto/sm2/sm2_aid.c
```diff
@@ -0,0 +1,45 @@
...
+
+#define MD_CASE(name)                                                   \
+    case NID_##name:                                                    \
+        *len = sizeof(algorithmidentifier_##name##_der);                \
+        return algorithmidentifier_##name##_der
+
+const unsigned char *sm2_algorithmidentifier_encoding(int md_nid, size_t *len)
+{
+    switch (md_nid) {
+        MD_CASE(sm3);
```
mattcaswell: All of this seems a little overkill for one digest. Couldn't this be simplified on the assumption that there is only ever one interesting digest for use with SM2?

providers/implementations/signature/sm2sig.c
```diff
@@ -0,0 +1,502 @@
...
+    unsigned char aid[OSSL_MAX_ALGORITHM_ID_SIZE];
+    size_t  aid_len;
+
+    /* main digest */
+    EVP_MD *md;
+    EVP_MD_CTX *mdctx;
+    size_t mdsize;
+} PROV_SM2_CTX;
+
+/* TODO: it seems SM2 doesn't need this */
```
mattcaswell: Yes...agree. Seems a little too much and could be removed.

providers/implementations/signature/sm2sig.c
```diff
@@ -0,0 +1,502 @@
...
+static void *sm2sig_newctx(void *provctx)
+{
+    PROV_SM2_CTX *psm2ctx = OPENSSL_zalloc(sizeof(PROV_SM2_CTX));
+
+    if (psm2ctx == NULL)
+        return NULL;
+
+    psm2ctx->libctx = PROV_LIBRARY_CONTEXT_OF(provctx);
+    /* don't allow to change MD, and in fact there is no such need */
+    psm2ctx->flag_allow_md = 0;
```
mattcaswell: So, can we just delete this field? It seems unnecessary.

providers/implementations/signature/sm2sig.c
```diff
@@ -0,0 +1,502 @@
...
+        return 1;
+    }
+
+    if (sigsize < (size_t)ecsize)
+        return 0;
+
+    if (ctx->mdsize != 0 && tbslen != ctx->mdsize)
+        return 0;
+
+    /* I would like to make the name something like SM2_sign... */
```
mattcaswell: Why?

providers/implementations/signature/sm2sig.c
```diff
@@ -0,0 +1,502 @@
...
+    PROV_SM2_CTX *ctx = (PROV_SM2_CTX *)vpsm2ctx;
+    size_t algorithmidentifier_len = 0;
+    const unsigned char *algorithmidentifier;
+
+    free_md(ctx);
+
+    if (!sm2sig_signature_init(vpsm2ctx, ec))
+        return 0;
+
+    ctx->md = EVP_MD_fetch(ctx->libctx, mdname, props);
```
mattcaswell: Should we simply disallow anything that isn't SM3?

providers/implementations/signature/sm2sig.c
```diff
@@ -0,0 +1,502 @@
...
+
+    if (psm2ctx == NULL || params == NULL)
+        return 0;
+
+    p = OSSL_PARAM_locate(params, OSSL_SIGNATURE_PARAM_ALGORITHM_ID);
+    if (p != NULL
+        && !OSSL_PARAM_set_octet_string(p, psm2ctx->aid, psm2ctx->aid_len))
+        return 0;
+
+    p = OSSL_PARAM_locate(params, OSSL_SIGNATURE_PARAM_DIGEST);
```
mattcaswell: Does it make sense to have this as a gettable parameter if the only digest we can ever have is SM3?

providers/implementations/signature/sm2sig.c
```diff
@@ -0,0 +1,502 @@
...
+
+static int sm2sig_set_ctx_params(void *vpsm2ctx, const OSSL_PARAM params[])
+{
+    PROV_SM2_CTX *psm2ctx = (PROV_SM2_CTX *)vpsm2ctx;
+    const OSSL_PARAM *p;
+
+    if (psm2ctx == NULL || params == NULL)
+        return 0;
+
+    p = OSSL_PARAM_locate_const(params, OSSL_SIGNATURE_PARAM_DIGEST);
```
mattcaswell: Again - is this actually needed?

crypto/sm2/sm2_aid.c
InfoHunter: Well, I am struggling with this. It seems that SM2 should only be used with SM3 since the spec says SM2 should only the hash functions designated by the cryptography administration department and, the only hash function defined in SM ciphers is SM3.

But this is the thing that interesting. SM2 actually can use other hash algorithm other than SM3. So what I consider is that if there are educational or experimental use cases that require to use other hash functions - for instance, to have a performance benchmark among different hash functions with SM2

InfoHunter: So maybe we need to provide the flexibility of the hash function choice and let the provider to decide if it will to be complied with Chinese validation programs.

providers/implementations/signature/sm2sig.c
InfoHunter: Right, if we decide to limit the algorithm to SM2 only.

providers/implementations/signature/sm2sig.c
InfoHunter: Because initially when I was working on this file, I chose the name `sm2_sign` here and I found it conflicted with the internal `sm2_sign` - so I changed the `sm2_sign` into `sm2sig_sign`. I am not sure if other SM2 provider implementor will also confuse with this and feel it not convenient.  (Other internal functions mostly are named with the upper-case name like ECDSA_sign or so)

providers/implementations/signature/sm2sig.c
InfoHunter: Not sure yet, all depend on how we treat https://github.com/openssl/openssl/pull/11248/files#r394931352

providers/implementations/signature/sm2sig.c
InfoHunter: Same answer as above ;-)

providers/implementations/signature/sm2sig.c
InfoHunter: Still the same issue...

providers/implementations/signature/sm2sig.c
mattcaswell: > Other internal functions mostly are named with the upper-case name like ECDSA_sign or so

ECDSA_sign is (unfortunately) not internal. Generally speaking we prefer to reserve the capitilized form for external functions and non-capitalized for internal.

InfoHunter: So, it seems the name similar to `sm2_internal_sign` is more reasonable

InfoHunter:
@mattcaswell , for the SM3 considerations, I found in the legacy part there is not such limit, so do we need to limit that in default provider or should we just make it compatible with what legacy part does (OpenSSL 1.1.1)

InfoHunter:
The `WIP` is removed from this PR.

providers/implementations/signature/sm2sig.c
InfoHunter: I changed the name to `sm2_internal_sign` and `sm2_internal_verify` respectively

InfoHunter:
Ping?

InfoHunter:
What is the arrangement for this PR? Any dependencies yet?

InfoHunter:
Ping @openssl/otc

InfoHunter:
Should this be included in any alpha version? maybe the next alpha-2?

paulidale:
This ought to be included at some point.  It isn't essential for the alpha releases but it should be done before the first beta.

I do not expect that any of the core team will have time to do this, we're relying on you or a Chinese compatriot to convert it into a provider based solution.  The conversion isn't difficult, but it could be a bit tedious.

romen:
@InfoHunter can you rebase this on latest master? A lot of things changed since the last push (sorry I did not find the time to review this sooner) so it would be better to test this against latest master.

romen:
I believe it doesn't make much sense to include this without the corresponding keymgmt part

InfoHunter:
Okay, no problem, I would probably find some time later next week to handle this.

providers/implementations/signature/sm2sig.c
```diff
@@ -0,0 +1,501 @@
...
+/* TODO: it seems SM2 doesn't need this */
+static int sm2sig_get_md_nid(const EVP_MD *md)
+{
+    /*
+     * Because the EC library deals with NIDs, we need to translate.
+     * We do so using EVP_MD_is_a(), and therefore need a name to NID
+     * map.
+     */
+    static const OSSL_ITEM name_to_nid[] = {
+        { NID_sm2,   OSSL_DIGEST_NAME_SM3 },
```
levitte, line 93: .. this wasn't quite right

levitte:
> 3. I assume that the EC key management can be reused in SM2 case, or should I create a new SM2 keymgmt?

Kinda sorta.  We're treating SM2 as a separate key type, even though the underlying structure is the same.  You should be able to use most of the same functions for both, just make sure you keep track of what key type you're handling at all times, and there should be a separate `OSSL_DISPATCH` for SM2.

I've done that same sort of work, for separating plain RSA and RSA-PSS keys.  See #11751

levitte:
> @mattcaswell , for the SM3 considerations, I found in the legacy part there is not such limit, so do we need to limit that in default provider or should we just make it compatible with what legacy part does (OpenSSL 1.1.1)

I have personally no issues with that.  There's still the concept of the default digest type, which is up to the keymgmt to specify.

InfoHunter:
Have any of you encountered the duplicated symbol issue when building master?

~~~
duplicate symbol '_der_oid_id_shake128_len' in:
    providers/libnonfips.a(libnonfips-lib-der_digests.o)
    providers/libnonfips.a(libnonfips-lib-der_rsa.o)
~~~

levitte:
Hmmm, I'll look into that.  Thosse two may need a little bit of reorganization.

InfoHunter:
> I've done that same sort of work, for separating plain RSA and RSA-PSS keys. See #11751

It seems the PR number is not quite right?

mattcaswell:
> It seems the PR number is not quite right?

#11710

InfoHunter:
Thanks

levitte:
Refactoring of provider side DER stuff in #11868

InfoHunter:
Unfortunately, after the symbol conflicts are resolved, it seems there is still building problem with master.

~~~
Undefined symbols for architecture x86_64:
  "_BIO_get_data", referenced from:
      _bio_core_write_ex in libcommon.a(libcommon-lib-bio_prov.o)
      _bio_core_read_ex in libcommon.a(libcommon-lib-bio_prov.o)
  "_BIO_meth_free", referenced from:
      _bio_prov_init_bio_method in libcommon.a(libcommon-lib-bio_prov.o)
  "_BIO_meth_new", referenced from:
      _bio_prov_init_bio_method in libcommon.a(libcommon-lib-bio_prov.o)
  "_BIO_meth_set_create", referenced from:
      _bio_prov_init_bio_method in libcommon.a(libcommon-lib-bio_prov.o)
  "_BIO_meth_set_ctrl", referenced from:
      _bio_prov_init_bio_method in libcommon.a(libcommon-lib-bio_prov.o)
  "_BIO_meth_set_destroy", referenced from:
      _bio_prov_init_bio_method in libcommon.a(libcommon-lib-bio_prov.o)
  "_BIO_meth_set_gets", referenced from:
      _bio_prov_init_bio_method in libcommon.a(libcommon-lib-bio_prov.o)
  "_BIO_meth_set_puts", referenced from:
      _bio_prov_init_bio_method in libcommon.a(libcommon-lib-bio_prov.o)
  "_BIO_meth_set_read_ex", referenced from:
      _bio_prov_init_bio_method in libcommon.a(libcommon-lib-bio_prov.o)
  "_BIO_meth_set_write_ex", referenced from:
      _bio_prov_init_bio_method in libcommon.a(libcommon-lib-bio_prov.o)
  "_BIO_new", referenced from:
      _bio_new_from_core_bio in libcommon.a(libcommon-lib-bio_prov.o)
  "_BIO_set_data", referenced from:
      _bio_new_from_core_bio in libcommon.a(libcommon-lib-bio_prov.o)
  "_BIO_set_init", referenced from:
      _bio_core_new in libcommon.a(libcommon-lib-bio_prov.o)
      _bio_core_free in libcommon.a(libcommon-lib-bio_prov.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
~~~

providers/implementations/signature/sm2sig.c
levitte: This still isn't quite right

levitte:
@InfoHunter, would you mind telling me how you configured?

InfoHunter:
> @InfoHunter, would you mind telling me how you configured?

On MacOS (with clang), the config is: `config --prefix=/to/some/place enable-asan --strict-warnings`

InfoHunter:
Anyway, despite of that ld failure, I updated the SM2 signature algorithm provider code a bit.

InfoHunter:
... and rebased to master as well

crypto/sm2/sm2_aid.c
```diff
@@ -0,0 +1,45 @@
+/*
```
InfoHunter, line 1: It seems this whole file is not needed anymore, since the new DER part has already handled this.

levitte: The infrastructure is there, but you will need to add SM2 specific.

InfoHunter: Yes. SM2 specific stuffs are added in the latest commit - hope I have done right...

providers/implementations/signature/sm2sig.c
InfoHunter: Hmmm? I thought I have fixed this...Did you comment on an outdated code?

crypto/sm2/sm2_aid.c
levitte: I can help with those details this evening

InfoHunter: Thanks! One struggling thing is SM2 has no PublicKey OID. An SM2 key is encoded with `id_ecPublickey` IIRC at current stage, which is compatible with OpenSSL 1.1.1.

InfoHunter:
> I believe it doesn't make much sense to include this without the corresponding keymgmt part

@romen , I looked a bit in the ec_kmgmt.c, and I think it's possible just to wrap sm2_kmgmt around ec_kmgmt - with some minor modifications, like fixing the curve to SM2 curve...

romen:
> > I believe it doesn't make much sense to include this without the corresponding keymgmt part
> 
> @romen , I looked a bit in the ec_kmgmt.c, and I think it's possible just to wrap sm2_kmgmt around ec_kmgmt - with some minor modifications, like fixing the curve to SM2 curve...

@InfoHunter I left some notes regarding SM2 in the description of #10631 (which also references some important comments): maybe they are still useful?

I still think sm2_keymgmt needs to be distinct, yet most of it is going to be basically a duplicate of the ec_kmgmt with minor adjustments (e.g., as you say, maybe there is no need to include a curve name).

As a side note, I will leave it to your interpretation of it as you definitely have more insight on it, but regarding removing the flexibility of choosing this or that curve underneath SM2, [looking at the standards](https://github.com/openssl/openssl/issues/8435#issuecomment-600837510) before, I had the impression it is not unlikely that in the future one or more different curves could be used and recommended under the SM2 cryptosystem.
For example, considering PQC is coming to industry and it is not limited anymore to academic debate, I can see how there might be a push in the near future to recommend a >512 bit curve as the preferred one!

crypto/sm2/sm2_aid.c
levitte: Yes, I know. That is a bit of a problem, and means that certain assumptions need to be made...
I will have to look again how I solved that for the legacy implementation. I forget...

InfoHunter:
> I left some notes regarding SM2 in the description of #10631 (which also references some important comments): maybe they are still useful?

I reviewed the comments in #10631 and I finally recall something discussed then... Will try to resolve this SM2 key management stuff.

InfoHunter:
> For example, considering PQC is coming to industry and it is not limited anymore to academic debate, I can see how there might be a push in the near future to recommend a >512 bit curve as the preferred one!

I found this is reasonable and interesting as well!

InfoHunter:
Squashed and rebased to master - Although I still can't build the final binaries on my local machine but at least I am sure the compilation is OK ;-)

InfoHunter:
@romen , would you like to show me some examples on how `ec_kmgmt` is used?

levitte:
Hmmm, I'm actually ambivalent about having a distinct SM2 key type, as there's really *nothing* (apart from the default curve being SM2) that makes it distinct from an EC key.

The SM2 *operation* (signature), on the other hand, is distinct.

In legacy implementations, the operation type an the key type are strongly associated (which is the main reason why an `EVP_PKEY_set_alias_type()` call was needed, while they are less so with provider implementations.  I'm pondering if we can take advantage of that in a sane way.

romen:
- SM2 keygen is different (different ranges)
- it should be possible to run ECDH/DSA over the so-called SM2 curve
- it should be possible to rune SM2DH/SM2DSA over any curve

I believe there is no way around having 2 separate keymgmt given these conditions!

romen:
Their serialization looks the same: that is true, and it's basically standardization debacle that generates all the confusion and the weirdness around set_alias and friends.

romen:
> @romen , would you like to show me some examples on how `ec_kmgmt` is used?

@InfoHunter You mean how the provider interface calls into it? I think @levitte can be more helpful with that!

InfoHunter:
> SM2 keygen is different (different ranges)
Right. Current SM2 key generation follows this calling sequence:
~~~
        EC_KEY *eckey = EC_KEY_new();
        EC_KEY_set_group(eckey, group);
        EC_KEY_generate_key(eckey);
~~~
I remember there was a discussion that current OpenSSL SM2 key generation does not exactly follow the SM2 specification?

> 2. it should be possible to run ECDH/DSA over the so-called SM2 curve
> 3. it should be possible to rune SM2DH/SM2DSA over any curve

For 3, we can provide that flexibility. But for 2, I am not sure if SM2 curve is actually necessary to be used with ECDSA. (ECDH may have the use case)

InfoHunter:
> You mean how the provider interface calls into it?

Yes, exactly! I found some in the EVP code, but if there are more high level descriptions, that would be better.

InfoHunter:
> SM2 keygen is different (different ranges)

It seems there is no special treatment on the SM2 keygen in OpenSSL 1.1.1. It should just follow the standard `EC_GFp_mont_method->keygen` procedures.

Are you suggesting that we should fix this in provider implementation of SM2?

levitte:
> - SM2 keygen is different (different ranges)

Okie, that's an argument.

> - it should be possible to run ECDH/DSA over the so-called SM2 curve
> - it should be possible to rune SM2DH/SM2DSA over any curve

Sure, but those signature operations, not keymgmt.

> I believe there is no way around having 2 separate keymgmt given these conditions! 

Only the keygen concern seems relevant for having a separate SM2 key type.

levitte:
> It seems there is no special treatment on the SM2 keygen in OpenSSL 1.1.1. It should just follow the standard `EC_GFp_mont_method->keygen` procedures.
> 
> Are you suggesting that we should fix this in provider implementation of SM2?

Note that the legacy code for SM2 has changed in master, to avoid the dumber quirks that are still present in 1.1.1

romen:
> > * SM2 keygen is different (different ranges)
> 
> Okie, that's an argument.
> 
> > * it should be possible to run ECDH/DSA over the so-called SM2 curve
> > * it should be possible to rune SM2DH/SM2DSA over any curve
> 
> Sure, but those signature operations, not keymgmt.
> 
> > I believe there is no way around having 2 separate keymgmt given these conditions!
> 
> Only the keygen concern seems relevant for having a separate SM2 key type.

@levitte they are still valid arguments because of this

https://github.com/openssl/openssl/blob/93f99b681ab5a1cf7062053323e09b0cad5ff854/providers/implementations/keymgmt/ec_kmgmt.c#L54-L64

The keymgmt says what are the operations names, so SM2 cannot reuse the EC keymgmt and then USE different operations!

romen:
> > SM2 keygen is different (different ranges)
> 
> It seems there is no special treatment on the SM2 keygen in OpenSSL 1.1.1. It should just follow the standard `EC_GFp_mont_method->keygen` procedures.
> 
> Are you suggesting that we should fix this in provider implementation of SM2?

We have an open issue regarding the defect in the 1.1.1 implementation: https://github.com/openssl/openssl/issues/8435#issuecomment-481540010

The reason why it wasn't fixed yet is exactly because of how much in 1.1.1 SM2 and ec have been tied together due to the internal architecture and serialization/deserialization issues.
In 3.0.0 we should aim for a saner implementation where the 2 cryptosystems are handled distinctly and code duplication is avoided in smarter ways!

levitte:
> The keymgmt says what are the operations names, so SM2 cannot reuse the EC keymgmt and then USE different operations!

We might have to re-think the returned value, for example make that a NULL-terminated array of string pointers instead of just one string.  That would allow us to return `{ "ECDSA", "SM2", NULL }` for possible signature operations.  To be checked against a fetch EVP_SIGNATURE with EVP_SIGNATURE_is_a...  something like that.

levitte:
> In 3.0.0 we should aim for a saner implementation where the 2 cryptosystems are handled distinctly and code duplication is avoided in smarter ways!

We can only do so much!  If the keys in PKCS#8 form don't have any indication other than the SM2 curve, we only have an educated guess, and in doing so, we would remove any possibility to run ECDSA with the SM2 curve or SM2 signatures with other curves.

romen:
Deserialization is an issue bacuase the standardization of SM2 made them indistinguishable, true.

Somewhere in the deserialization step we probably will need to keep defaulting to "create an SM2 key if this is serialized as an EC key over the so-called-SM2-curve". 

Yet it should be possible from an application that can sidestep or implement its own deserialization stage, to ask a provider to generate an EC key on the SM2 curve, and viceversa to generate an SM2 key on this nist curve.

I don't see huge benefits in unifying again the SM2 and ec keymgmt, adding the complexity of multiple alternative options for the operations that a given key object supports.

levitte:
> Yet it should be possible from an application that can sidestep or implement its own deserialization stage, to ask a provider to generate an EC key on the SM2 curve, and viceversa to generate an SM2 key on this nist curve.

... and then it gets saved to file, and the next time it's read from that same file, it's suddenly potentially a different key type.  If that isn't a source of confusion, I don't know what is.

romen:
> > Yet it should be possible from an application that can sidestep or implement its own deserialization stage, to ask a provider to generate an EC key on the SM2 curve, and viceversa to generate an SM2 key on this nist curve.
> 
> ... and then it gets saved to file, and the next time it's read from that same file, it's suddenly potentially a different key type. If that isn't a source of confusion, I don't know what is.

Well, I don't think the library can solve a confusion that is embedded in the result of the standardization process.

Even having a single keymgmt for SM2 and EC will cause confusion if a user checks out the saved file: it is indistinguishable from an EC key on the SM2 curve.
Looking from this perspective, it is already confusing for a user in 1.1.1 to see that a key that looks like an EC key on disk does not compute ECDH or ECDSA when used for derive/sign!

romen:
This is my personal opinion (and I am biased because it's my proposal), but I think that:

- an EC keymgmt, with its own keygen, that is associated with ECDH/ECDSA operations
- a distinct SM2 keymgmt, with its own different keygen, that is associated with SM2DH/SM2DSA operations
- an EC deserializer, that defaults to calling the EC keymgmt, unless the so-called-SM2-curve OID is detected for which SM2 keymgmt is used instead

is going to be cleaner than

- a single keymgmt that does
   + EC keygen or SM2 keygen _(*)_
   + ECDH or SM2DH _(*)_
   + ECDSA or SM2DSA _(*)_
- an EC deserializer, that sets the internal flags that choose among EC operations or SM2 operations depending on detecting the so-called-SM2-curve OID.

_(*): depending on something internal and difficult to reach or override_


Yes, in both cases there will still be the issue of the confusion caused by the indistinguishability of the serialized format, but we can't really escape it until the SM2 spec is revised (and probably not even then, to retain compatibility with 1.1.1), but at least we avoid creating an overloaded key object standing for 2 different cryptosystems (keys are generated in different ways, they are associated with different operations) that only share (unluckily) the serialization format.

romen:
Maybe @bbbrumley can share his thoughts about this as well, given that he has been struggling with trying to fix #8435 in 1.1.1 and can have even more insights on what could be done better at this time.

InfoHunter:
> I don't see huge benefits in unifying again the SM2 and ec keymgmt, adding the complexity of multiple alternative options for the operations that a given key object supports.

One the advantages could be less code duplication. According to the discussions above, It seems the only difference we need to take care is the key generation process in the keymgmt scope.

levitte:
> One the advantages could be less code duplication. According to the discussions above, It seems the only difference we need to take care is the key generation process in the keymgmt scope.

Why do you need to duplicate code?  You could set a flag of some sort in the keydata (like we now do for the RSA sub-types), all that's needed then is a different constructor, and then pay attention to that flag in the few places where that's needed.  The only thing that would be duplicated is the OSSL_DISPATCH array.  I would much rather than a lot of code copying.

bbbrumley:
This is just my opinion as some random guy on the Internet.

> We might have to re-think the returned value, for example make that a NULL-terminated array of string pointers instead of just one string. That would allow us to return { "ECDSA", "SM2", NULL } for possible signature operations

@levitte I feel like this is similar to the line of thinking that led to the 111 mess with SM2. "Well let's just piggy back off this -- it's similar enough" and turns out it wasn't, right?

> and then it gets saved to file, and the next time it's read from that same file, it's suddenly potentially a different key type. If that isn't a source of confusion, I don't know what is.

I agree! And what you're doing in master is much better than the 111 state.

> This is my personal opinion (and I am biased because it's my proposal), but I think that

@romen I agree with this proposal.

Even if it means duplicate code here and there, keep them as separate as possible.

The way I look at it is this: what if SM2 had it's own OID that wasn't under the legacy EC hierarchy? Yes that is a standardization mistake, but put that aside. Step into fantasy land for a minute.

* Would you still be trying to tie in SM2 to the EC module at any level?
* Including key management?
* Does OpenSSL do that for ed25519? I can write that curve as a legacy curve, too ...

Ignore the standards mistake, and ask yourself these questions. I think you'll find the right decision for OpenSSL then.

richsalz:
> This is just my opinion as some random guy on the Internet.

Thanks for the chuckle in these trying times!

InfoHunter:
I just pushed a new commit containing the separated SM2 key management skeleton (and it's only a skeleton so please consider this as WIP at current stage). The current implementation is just a copy of the EC key management. According to what Nicola proposed in the emails during the weekend, I need to replace the content with SM2 implementation. We can have a discussion on this on the upcoming online meeting on Tuesday ;-)

romen:
We could house the SM2 keymgmt in `providers/implementations/keymgmt/ec_kmgmt.c` and do something similar to the RSA and RSA-PSS keymgmt:

https://github.com/openssl/openssl/blob/7e4f01d8ba9983b37758eb8842c64500ee0b29ca/providers/implementations/keymgmt/rsa_kmgmt.c#L584-L626

Most pointers in the dispatch table are identical, and only the ones that need to differ are actually different. This way we minimize duplication.

At a first examination, I think the SM2 keymgmt should differ from the EC one in the following:

- `OSSL_FUNC_KEYMGMT_GEN` (it should be a separate function, very similar to the EC one, but internally instead of `EC_KEY_generate_key(ec);` we should call a new internal function that behaves like it, but uses the proper range for picking the secret scalar)
- `OSSL_FUNC_KEYMGMT_QUERY_OPERATION_NAME` (pointing to the SM2 specific operations like you have in the current version of this PR)
- `OSSL_FUNC_KEYMGMT_GET_PARAMS`
  - `OSSL_PKEY_PARAM_SECURITY_BITS` might be defined differently for the SM2 curve?
    https://github.com/openssl/openssl/blob/7e4f01d8ba9983b37758eb8842c64500ee0b29ca/providers/implementations/keymgmt/ec_kmgmt.c#L484-L520

  - `OSSL_PKEY_PARAM_USE_COFACTOR_ECDH`, `OSSL_PKEY_PARAM_TLS_ENCODED_PT` shouldn't be supported
- `OSSL_FUNC_KEYMGMT_GETTABLE_PARAMS`, `OSSL_FUNC_KEYMGMT_SET_PARAMS`, `OSSL_FUNC_KEYMGMT_SETTABLE_PARAMS` should be similarly revised (seems like maybe there are no settable params leftover, but maybe there is the opportunity to add the SM2-specific `USER_ID` personalization string as a parameter here!)
- `OSSL_FUNC_KEYMGMT_IMPORT`, this could probably point to a wrapper around the existing internal `ec_import` function. It's only goal would be to filter import calls, so that the SM2 keymgmt only accepts keys/params that are on the so-called "SM2 curve" and fail otherwise.

InfoHunter:
I am going to draw back the latest commits containing the keymgmt code in this PR and open a new PR to address this.

InfoHunter:
> At a first examination, I think the SM2 keymgmt should differ from the EC one in the following:
> OSSL_FUNC_KEYMGMT_GEN (it should be a separate function, very similar to the EC one, but internally instead of EC_KEY_generate_key(ec); we should call a new internal function that behaves like it, but uses the proper range for picking the secret scalar)

And SM2 needn't to provide the ability of specifying curves, did we have a consensus on this yesterday?

romen:
@InfoHunter

> And SM2 needn't to provide the ability of specifying curves, did we have a consensus on this yesterday?

Regarding this I think the best course of action (and minimal work) is to leave the EC flexibility underneath, but filter so that if the curve id does not match `import` fails.

That way when another supported curve is added to part 5 of the SM2 standards we just have to add one more entry to the allowlist and now we don't need to refractor extra code in the keymgmt.

InfoHunter:
> but maybe there is the opportunity to add the SM2-specific USER_ID personalization string as a parameter here!

The SM2 user ID seems not a param with the key itself, this ID is involved in the SM2 signature computation progress which has been addressed by @levitte months ago IIRC. So it seems we don't need to consider supporting this in the key management part of SM2

InfoHunter:
As per https://github.com/openssl/openssl/pull/12536#issuecomment-672997935 and https://github.com/openssl/openssl/pull/12536#issuecomment-673010582

This PR is going to moved into PR #12536 , so this one can be closed now.

PR #11142, opened 2020-02-21
Chunk 8 of CMP contribution to OpenSSL: CMP server

DDvO:
Certificate Management Protocol (CMP, RFC 4210) extension to OpenSSL
Also includes CRMF (RFC 4211) and HTTP transfer (RFC 6712).
Adds the CMP and CRMF API to libcrypto and the "cmp" app to the CLI.
Adds extensive documentation and tests.
 
8th chunk: 
* generic CMP server code in `crypto/cmp/cmp_server.c`
* mock server for testing the CMP client in `apps/cmp_mock_srv.c`

and related files.

This PR is the continuation of https://github.com/mpeylo/cmpossl/pull/201.

Due to moving the mock server code out of libcrypto as discussed https://github.com/mpeylo/cmpossl/pull/201#issuecomment-581440768 
a couple of further CMP functions had to be exported.

This PR also includes various fixups to earlier CMP contribution chunks that came up during re-structuring the (test) server code and doing tests on the client code in upcoming chunk 9.
Among others, it slightly generalizes 
`static int ts_check_status_info(TS_RESP *response)` in `crypto/ts/ts_rsp_verify.c` to 
`sk_ASN1_UTF8STRING2text(STACK_OF(ASN1_UTF8STRING) *text, char sep)` and
makes it accessible within libcrypto by adding its declaration to `include/internal/cryptlib.h`.

- [x] documentation is added or updated
- [x] tests are added or updated

PR #11250, opened 2020-03-04
Don't compile commands if disabled

richsalz:
Rather than wrapping whole files in "ifndef OPENSSL_NO_xxx" we handle
the changes in build.info

Spurred by @mattcaswell's comment at https://github.com/openssl/openssl/pull/11142#discussion_r387906007 and because I'm waiting for some PR's to be merged and reviewed.

I will do another PR to fix the libs and remote that macro completely.

### DIFF
--- a/apps/build.info
+++ b/apps/build.info
@@ -13,15 +13,40 @@ ENDIF
 $OPENSSLSRC=\
         openssl.c progs.c \
         asn1pars.c ca.c ciphers.c cms.c crl.c crl2p7.c dgst.c \
-        ec.c ecparam.c enc.c engine.c errstr.c \
-        genpkey.c kdf.c mac.c nseq.c ocsp.c passwd.c pkcs12.c pkcs7.c \
+        enc.c errstr.c \
+        genpkey.c kdf.c mac.c nseq.c passwd.c pkcs7.c \
         pkcs8.c pkey.c pkeyparam.c pkeyutl.c prime.c rand.c req.c \
         s_client.c s_server.c s_time.c sess_id.c smime.c speed.c \
-        spkac.c srp.c ts.c verify.c version.c x509.c rehash.c storeutl.c \
+        spkac.c verify.c version.c x509.c rehash.c storeutl.c \
         list.c info.c provider.c fipsinstall.c
+IF[{- !$disabled{'des'} -}]
+  $OPENSSLSRC=$OPENSSLSRC pkcs12.c
+ENDIF
+IF[{- !$disabled{'ec'} -}]
+  $OPENSSLSRC=$OPENSSLSRC ec.c ecparam.c
+ENDIF
+IF[{- !$disabled{'ocsp'} -}]
+  $OPENSSLSRC=$OPENSSLSRC ocsp.c
+ENDIF
+IF[{- !$disabled{'srp'} -}]
+  $OPENSSLSRC=$OPENSSLSRC srp.c
+ENDIF
+IF[{- !$disabled{'ts'} -}]
+  $OPENSSLSRC=$OPENSSLSRC ts.c
+ENDIF
 IF[{- !$disabled{'deprecated-3.0'} -}]
-  $OPENSSLSRC=$OPENSSLSRC \
-        dhparam.c dsa.c dsaparam.c gendsa.c rsa.c rsautl.c genrsa.c
+  IF[{- !$disabled{'dh'} -}]
+    $OPENSSLSRC=$OPENSSLSRC dhparam.c
+  ENDIF
+  IF[{- !$disabled{'dsa'} -}]
+    $OPENSSLSRC=$OPENSSLSRC dsa.c dsaparam.c gendsa.c
+  ENDIF
+  IF[{- !$disabled{'engine'} -}]
+    $OPENSSLSRC=$OPENSSLSRC engine.c
+  ENDIF
+  IF[{- !$disabled{'rsa'} -}]
+    $OPENSSLSRC=$OPENSSLSRC rsa.c rsautl.c genrsa.c
+  ENDIF
 ENDIF
 IF[{- !$disabled{'cmp'} -}]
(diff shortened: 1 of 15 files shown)


richsalz:
I will *probably* do another commit, and it is likely that the macro will have to remain (cf crypto/ebcdic.c)

apps/build.info
```diff
@@ -13,15 +13,40 @@ ENDIF
...
+IF[{- !$disabled{'srp'} -}]
+  $OPENSSLSRC=$OPENSSLSRC srp.c
+ENDIF
+IF[{- !$disabled{'ts'} -}]
+  $OPENSSLSRC=$OPENSSLSRC ts.c
+ENDIF
 IF[{- !$disabled{'deprecated-3.0'} -}]
-  $OPENSSLSRC=$OPENSSLSRC \
-        dhparam.c dsa.c dsaparam.c gendsa.c rsa.c rsautl.c genrsa.c
+IF[{- !$disabled{'dh'} -}]
```
slontis: maybe indent these inner if's

richsalz: I wasn't sure if that would work.  Will it?

mattcaswell: Yes - this is common practice - and in fact is done later on in this same build.info file.

richsalz: thanks for the gentle nudge :)  fixup commit pushed.

levitte: Er, why wouldn't it work?

richsalz: A quick read of ```doc/internal/man7/build.info.pod``` only mentioned leading whitespace for comment characters.

richsalz:
Build failure is YATT

(yet another travis timeout)

apps/build.info
levitte: Oh, ok.

apps/build.info
```diff
@@ -13,15 +13,40 @@ ENDIF
...
-        ec.c ecparam.c enc.c engine.c errstr.c \
-        genpkey.c kdf.c mac.c nseq.c ocsp.c passwd.c pkcs12.c pkcs7.c \
+        enc.c errstr.c \
+        genpkey.c kdf.c mac.c nseq.c passwd.c pkcs7.c \
         pkcs8.c pkey.c pkeyparam.c pkeyutl.c prime.c rand.c req.c \
         s_client.c s_server.c s_time.c sess_id.c smime.c speed.c \
-        spkac.c srp.c ts.c verify.c version.c x509.c rehash.c storeutl.c \
+        spkac.c verify.c version.c x509.c rehash.c storeutl.c \
         list.c info.c provider.c fipsinstall.c
+IF[{- !$disabled{'des'} -}]
```
slontis: Err - I realise this was what it was already doing - but this just seems a bit crazy to me that it cant choose another default??

slontis:
needs a rebase.

apps/build.info
richsalz: I don't know of another way to do this, but I'd be happy to follow suggestions in a separate PR, not this one.

richsalz:
rebased.  the only change was apps/rsa.c removing pre-processor indents to fix the merge conflict.

richsalz:
>Given the age, a rebase and rerun of the CI is probably in order.

This sat here for a month.  And now I am being asked to rebase/wait because nobody else on the project had time or interest to review this.  GitHub claims there are no conflicts.  I'm not going to do it.  The project needs to understand the costs to the community of letting things just sit.  So, no.

bernd-edlinger:
Sorry Rich I am with you.
Just give me a few minutes for a nap.

bernd-edlinger:
do not rebase.
I will do that locally and send a [WIP] for that.
I dont like any re-base after an approval FWIW

bernd-edlinger:
I'll approve if #11506 succeeds. How about that?

bernd-edlinger:
Given that master never succeed, I'll approve if it does not get worse.
Is that also an 1.1.1 issue?

richsalz:
>I'll approve if #11506 succeeds. How about that?

It already has two approvals, so perhaps you -1 if it doesn't succeed.  Or feel free to take this over, or just let it drop.  (There are many PR's that are far older than this one, and way more important. The project team should also be looking at those.)

bernd-edlinger:
there was a force-push after Richards approval, how am I going to know?
Anyway, I will approve and this time it will count.

bernd-edlinger:
Relax, one thing is for sure, I will never let you drop.

richsalz:
> Is that also an 1.1.1 issue?

In most cases, the project hasn't considered "code cleanup" to valid for backporting.

Yes, @levitte needs to reconfirm.  Or someone else needs to approve.  When I force-pushed I explained what changed in https://github.com/openssl/openssl/pull/11250#issuecomment-596542020 but yes, I could be mistaken or deceitful.

bernd-edlinger:
No I need to approve, pauli did already approve that is 1+1=2

bernd-edlinger:
Please do not force push.  I can handle that. relax.

bernd-edlinger:
merged to master               1ae56f2 
Thanks!

bernd-edlinger:
do we need this in 1.1.1 too?

bernd-edlinger:
Request a review from me when you do that, I might miss this otherwise.

richsalz:
Thanks @bernd-edlinger .  This is code clean-up which has usually not been seen as a bugfix, so no backport IMO.

bernd-edlinger:
Maybe I disagree.

bernd-edlinger:
I see no code cleanup in 3.0 at all.

bernd-edlinger:
So what's the point in not trying to keep 1.1.1 maintainable since it is LTS ?

richsalz:
> So what's the point in not trying to keep 1.1.1 maintainable since it is LTS ?

That's not for me to say, it's up to the project to decide if they want this for 1.1.1 (and find someone to do it, I suppose).

PR #11261, opened 2020-03-05
Implement Ed25519 and Ed448 in the default provider

mattcaswell:
Implements Ed25519 and Ed448 in the default provider. In order to do this we have to add support for the "one shot" EVP_DigestSign() and EVP_DigestVerify() functions for providers. These two algorithms only support the one shot method.

Note: there is no support for serializers yet. That will come in a subsequent PR.

### DIFF
--- a/providers/implementations/signature/eddsa.c
+++ b/providers/implementations/signature/eddsa.c
@@ -1,0 +1,211 @@
+/*
+ * Copyright 2019 The OpenSSL Project Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License 2.0 (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
+ */
+
+#include <openssl/crypto.h>
+#include <openssl/core_numbers.h>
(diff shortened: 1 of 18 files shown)


mattcaswell:
Ping?

providers/implementations/signature/eddsa.c
```diff
@@ -0,0 +1,211 @@
+/*
+ * Copyright 2019 The OpenSSL Project Authors. All Rights Reserved.
```
slontis: 2020

mattcaswell:
Nit fixed. Fixup commit pushed. Please take another look.

mattcaswell:
Pushed. Thanks!

kroeckx:
Are there plans to add support for this in the FIPS provider?

I hope that FIPS 186-5 and SP 800-186 will be published soon.

mattcaswell:
> Are there plans to add support for this in the FIPS provider?

They're not currently on the list for inclusion. See Appendix 3 here:

https://www.openssl.org/docs/OpenSSL300Design.html

mattcaswell:
> They're not currently on the list for inclusion. See Appendix 3 here:

Although apparently #11371 is adding them to the FIPS provider as "non-FIPS algorithms"

PR #11262, opened 2020-03-05
Add NOTES.ZOS

p-steuer:
Hints on how to build OpenSSL 1.1.1 on z/OS that openssl-users
found to be helpful.

Adding a NOTES.ZOS file, as suggested by @richsalz . See also the corresponding thread on openssl-users.

### DIFF
--- a/NOTES.ZOS
+++ b/NOTES.ZOS
@@ -1,0 +1,73 @@
+
+ NOTES FOR THE Z/OS PLATFORM
+ ===========================
+
+ First of all note that OpenSSL is currently NOT supported on z/OS.
+
+ However, if you want to build OpenSSL on z/OS anyway, here are a few
+ hints on how to do that. Though these hints are intended to be helpful,
+ they might not work for you at all, depending on your build environment.
+
+ Building OpenSSL 1.1.1
+ ----------------------
+ (with xplink linkage, 64 bit, from EBCDIC sources)
+
+ 1. get a working EBCDIC perl version.
+
+ 2. gunzip the .tar file.
+
+ 3. use pax to extract from .tar. Example:
+
+      pax -ofrom=ISO8859-1,to=IBM-1047 -rvf openssl-1.1.1e.tar
+
+ 4. adjust the configuration files. Examples:
+
+ 4.1 10-main.conf
+
+     "OS390-Unix" => {
+         inherit_from     => [ "BASE_unix", asm("zos_asm") ],
+         cc               => "./tools/c99.sh",
+         cflags           => "-O -Wc,dll,XPLINK,exportall,hgpr,lp64 -Wa,'GOFF,SYSPARM(USE_XPLINK)' -qlongname -qlanglvl=extc99 -DB_ENDIAN -DCHARSET_EBCDIC -DNO_SYS_PARAM_H -D_ALL_SOURCE -D_OPEN_THREADS=2 -D_POSIX_SOURCE  -D_OPEN_MSGQ_EXT",
+         module_ldflags   => "-Wl,XPLINK,LP64",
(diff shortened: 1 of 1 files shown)


NOTES.ZOS
```diff
@@ -0,0 +1,73 @@
...
+ NOTES FOR THE Z/OS PLATFORM
+ ===========================
+
+ First of all note that OpenSSL is currently NOT supported on z/OS.
+
+ However, if you want to build OpenSSL on z/OS anyway, here are a few
+ hints on how to do that. Though these hints are intended to be helpful,
+ they might not work for you at all, depending on your build environment.
+
+ Building OpenSSL 1.1.1
```
levitte, line 11: Er.....  why do we want this on master?  If anything, I suggest these kind of "configuration packages" get added to our wiki.

For the suggested changes in 10-main.conf and 00-base-templates.conf, simply add them there, and add `c99.sh` as well...  although I would suggest `util/` rather than `tools/` since it's an internal utility.

NOTES.ZOS
```diff
@@ -0,0 +1,73 @@
...
+
+      pax -ofrom=ISO8859-1,to=IBM-1047 -rvf openssl-1.1.1e.tar
+
+ 4. adjust the configuration files. Examples:
+
+ 4.1 10-main.conf
+
+     "OS390-Unix" => {
+         inherit_from     => [ "BASE_unix", asm("zos_asm") ],
+         cc               => "./tools/c99.sh",
```
levitte, line 29: Value should be `'$(SRCDIR)/util/c99.sh'` to make it possible to build outside of the source tree.

NOTES.ZOS
```diff
@@ -0,0 +1,73 @@
...
+         shared_ldflags   => "-Wl,dll,XPLINK,LP64",
+         bn_ops           => "THIRTY_TWO_BIT RC4_CHAR",
+         thread_scheme    => "(unknown)",
+         perlasm_scheme   => "zOS64",
+     },
+
+ 4.2 00-base-templates.conf
+
+     # Copy and adjust the s390x_asm section by commenting out all asm
+     # or use a "no-asm" build right away.
```
levitte, line 41: This comment makes no sense at all.  What `s390x_asm` section?

levitte:
It *does* make sense to have a `NOTES.ZOS`, though, that might still explain how to configure on that platform

levitte:
Sorry for not seeing this earlier.  I don't see how this is ready to merge for the moment, so took away that label.

levitte:
As for the suggestion to add the 1.1.1 instructions to our wiki, I [just did](https://wiki.openssl.org/index.php/Configuration_%22packages%22)

NOTES.ZOS
p-steuer: Its was just meant to point out that all s390 specific (gnu-)asm wont compile on zos.

p-steuer:
Im fine with putting it in the wiki. Typo: zoss-unix.conf -> zos-unix.conf

NOTES.ZOS
p-steuer: Yes, its a 1.1.1 thing. Addind it to the wiki is probably a better solution than backporting this kind of ducomentation/configuration.

levitte:
So, that being done, do you have any plans on integrating OS390-Unix support in master?

mspncp:
ping

mattcaswell:
@p-steuer can this be merged now?

p-steuer:
Since it applies to 1.1.1 rather than master i agreed with @levitte to put it here instead https://wiki.openssl.org/index.php/Configuration_%22packages%22 - thanks.

From my side, it can be closed.

PR #11271, opened 2020-03-06
Some Serializer tweaks

mattcaswell:
If we fail to find a serializer then `ser` will be NULL. Don't try to deref it.

When searching for  a serializer, the first one we find might be from the desired provider. Don't keep looking for a desired serializer if we already found it.

### DIFF
--- a/crypto/serializer/serializer_pkey.c
+++ b/crypto/serializer/serializer_pkey.c
@@ -338,12 +338,14 @@ OSSL_SERIALIZER_CTX *OSSL_SERIALIZER_CTX_new_by_EVP_PKEY(const EVP_PKEY *pkey,
         OSSL_SERIALIZER_free(sel_data.first);
         OSSL_SERIALIZER_free(sel_data.desired);
 
-        current_props =
-            ossl_parse_property(libctx, OSSL_SERIALIZER_properties(ser));
-        if (ossl_property_match_count(check, current_props) > 0)
-            selection = OSSL_KEYMGMT_SELECT_ALL_PARAMETERS;
+        if (ser != NULL) {
+            current_props =
+                ossl_parse_property(libctx, OSSL_SERIALIZER_properties(ser));
+            if (ossl_property_match_count(check, current_props) > 0)
+                selection = OSSL_KEYMGMT_SELECT_ALL_PARAMETERS;
+            ossl_property_free(current_props);
+        }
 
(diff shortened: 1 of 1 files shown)


crypto/serializer/serializer_pkey.c
```diff
@@ -338,12 +338,14 @@ OSSL_SERIALIZER_CTX *OSSL_SERIALIZER_CTX_new_by_EVP_PKEY(const EVP_PKEY *pkey,
...
+            current_props =
+                ossl_parse_property(libctx, OSSL_SERIALIZER_properties(ser));
+            if (ossl_property_match_count(check, current_props) > 0)
+                selection = OSSL_KEYMGMT_SELECT_ALL_PARAMETERS;
+            ossl_property_free(current_props);
+        }
 
-        ossl_property_free(current_props);
         ossl_property_free(check);
     }
```
slontis: Do you want this to create a ctx with a NULL ser field - or is it supposed to be an error?

crypto/serializer/serializer_pkey.c
```diff
@@ -121,13 +121,13 @@ static void select_serializer(const char *name, void *data)
         return;
 
     if ((s = OSSL_SERIALIZER_fetch(d->libctx, name, d->propquery)) != NULL) {
-        if (d->first == NULL && s->serialize_data != NULL) {
-            d->first = s;
-        } else if (OSSL_SERIALIZER_provider(s) == d->desired_provider
+        if (OSSL_SERIALIZER_provider(s) == d->desired_provider
                    && s->serialize_object != NULL) {
```
levitte: Nit: Even for you, this is a little too much...

crypto/serializer/serializer_pkey.c
levitte: @slontis, it's not an error. See the comment above this function

crypto/serializer/serializer_pkey.c
mattcaswell: Fixed

mattcaswell:
Fixup pushed addressing the nit. Please take another look.

levitte:
Traviss failure is the usual timeout

levitte:
Ready to merge, approval stands

iamamoose:

oops. broken logic. fixing the bot.

mattcaswell:
Pushed. Thanks.

PR #11273, opened 2020-03-06
fix descriptions of credentials and verification app options

DDvO:
Looking for a way to provide extra untrusted certs for chain building during TLS server cert verification
I found that for both `s_client`  and `s_server` the documentation of the certificates given with the `-cert_chain` option attributes them as 'trusted' while in fact they are just untrusted additional certs to use for chain building. 

On the other hand the documentation of most of the `-chainCAfile`, `-chainCApath`, `-chainCAstore`, `-verifyCAfile`, `-verifyCApath`, and `-verifyCAstore` options failed to make clear that all the certificates addressed by them are trusted.

When correcting this I found and fixed some mistakes in the file format options and their descriptions.

I've also slightly improved the description of the `crl_download` and `nameopt` options. 

Moreover, the order of the description of these and related options was somewhat inconsistent, which is fixed as well in this PR.


DDvO:
BTW, how can one provide untrusted certs to help verifying peer certs, to use in case a peer does not send (all of) its cert chain?

DDvO:
Thanks @beldmit for your very swift approval!

Do you know if/how one can provide untrusted certs to help verifying peer TLS certs, to use in case a peer does not send (all of) its cert chain?
I found this possibility only for the `verify` app and direct use of `X509_STORE_CTX_init()`.

beldmit:
I don't know either :(
Not sure if it's a reasonable option because in real life (e. g. browsers) we don't have such a store so servers must include a full chain to avoid errors

DDvO:
I agree that for web scenarios pre-known untrusted certs make little sense,
but in managed (e.g., industrial) scenarios it can make much sense, 
in particular with limited bandwidth where it would be nice to avoid sending the same untrusted certs over and over.

richsalz:
This is going to be an RFC soon, it's just waiting for text editing: https://datatracker.ietf.org/doc/draft-ietf-tls-certificate-compression/?include_text=1

DDvO:
Nice, but this just compresses the certs being transmitted, but does not stop sending them.
And transmission size is not the only potential reason why a TLS node may need to augment the cert chain it receives from a peer.

DDvO:
Rebased and partly squashed.
Given that the approval was done two days before, can I merge this PR right away?

t8m:
It is missing OTC member approval.

kaduk:
Using extra certs for chain building can come into play when the server is sending a potentially valid chain but you want to use a different one, e.g., chaining up to a different or cross-signed root.

DDvO:
Ping for OTC approval

DDvO:
Rebased this PR. It still needs an OTC approval.

DDvO:
The review/approval by @jaym05700 does not count here.

DDvO:
Pinging again on a OTC member approval.

DDvO:
Merged - thanks!

DDvO:
I just found that the merge did not really happen, likely due to some misspelling or the like.
Now it's actually merged.

PR #11277, opened 2020-03-07
Check Port to Determine TLS Use

nathaniel-bennett:
I noticed that if one passes in a URL that doesn't have "https://" specified at the beginning, the function would not indicate the use of TLS, even if the port was specified as 443 (the standard port for HTTPS connections). So, I added a simple check to do this. If this is not in line with standard practices, feel free to discard it; I simply thought that it would be behavior a developer would expect out of the function.
Another thing to take into consideration--one unexpected side effect of my modification is that URLs such as http://example.com:443 would connect via https, despite the fact that only http was specified.

CLA: trivial

##### Checklist

### DIFF
--- a/crypto/http/http_lib.c
+++ b/crypto/http/http_lib.c
@@ -91,6 +91,8 @@ int OSSL_HTTP_parse_url(const char *url, char **phost, char **pport,
         goto err;
     if (pport != NULL && (*pport = OPENSSL_strdup(port)) == NULL)
         goto err;
+    if (pssl != NULL && strcmp(port, "443") == 0)
+        *pssl = 1;
 
     OPENSSL_free(buf);
     return 1;


richsalz:
It's not a side-effect, that's the *only* thing this PR does, right?  It assumes 443 is TLS even if the scheme doesn't say https.  Not sure that's the right thing to do.

kaduk:
> if one passes in a URL that doesn't have "https://" specified at the beginning, the function would not indicate the use of TLS, even if the port was specified as 443 (the standard port for HTTPS connections)

At a protocol level, it is the URL "scheme" (i.e., the "https:" part) that specifies what protocol is to be used to retrieve the indicated resource, including whether or not TLS is used to do so.  It is valid (albeit uncommon) protocol usage to speak http-not-s on port 443.

PR #11278, opened 2020-03-08
Add failure checks for UP_REF calls

slontis:
##### Checklist

### DIFF
--- a/crypto/cms/cms_env.c
+++ b/crypto/cms/cms_env.c
@@ -198,8 +198,10 @@ static int cms_RecipientInfo_ktri_init(CMS_RecipientInfo *ri, X509 *recip,
     if (!cms_set1_SignerIdentifier(ktri->rid, recip, idtype))
         return 0;
 
-    X509_up_ref(recip);
-    EVP_PKEY_up_ref(pk);
+    if (!X509_up_ref(recip))
+        return 0;
+    if (!EVP_PKEY_up_ref(pk))
+        return 0;
 
     ktri->pkey = pk;
     ktri->recip = recip;
--- a/crypto/ec/ec_mult.c
+++ b/crypto/ec/ec_mult.c
@@ -77,10 +77,13 @@ static EC_PRE_COMP *ec_pre_comp_new(const EC_GROUP *group)
 
 EC_PRE_COMP *EC_ec_pre_comp_dup(EC_PRE_COMP *pre)
 {
-    int i;
-    if (pre != NULL)
-        CRYPTO_UP_REF(&pre->references, &i, pre->lock);
-    return pre;
+    int ref = 0;
+
+    if (pre != NULL) {
+        if (!CRYPTO_UP_REF(&pre->references, &ref, pre->lock))
+            return NULL;
+    }
+    return ref > 0 ? pre : NULL;
 }
 
 void EC_ec_pre_comp_free(EC_PRE_COMP *pre)
@@ -90,7 +93,8 @@ void EC_ec_pre_comp_free(EC_PRE_COMP *pre)
     if (pre == NULL)
         return;
 
-    CRYPTO_DOWN_REF(&pre->references, &i, pre->lock);
+    if (!CRYPTO_DOWN_REF(&pre->references, &i, pre->lock))
+        return;
     REF_PRINT_COUNT("EC_ec", pre);
     if (i > 0)
         return;
--- a/crypto/ec/ecp_nistp256.c
+++ b/crypto/ec/ecp_nistp256.c
@@ -1875,10 +1875,13 @@ static NISTP256_PRE_COMP *nistp256_pre_comp_new(void)
 
 NISTP256_PRE_COMP *EC_nistp256_pre_comp_dup(NISTP256_PRE_COMP *p)
 {
-    int i;
-    if (p != NULL)
-        CRYPTO_UP_REF(&p->references, &i, p->lock);
-    return p;
+    int ref = 0;
+
+    if (p != NULL) {
+        if (!CRYPTO_UP_REF(&p->references, &ref, p->lock))
+            return NULL;
+    }
+    return ref > 0 ? p : NULL;
 }
 
 void EC_nistp256_pre_comp_free(NISTP256_PRE_COMP *pre)
@@ -1888,7 +1891,8 @@ void EC_nistp256_pre_comp_free(NISTP256_PRE_COMP *pre)
     if (pre == NULL)
         return;
 
-    CRYPTO_DOWN_REF(&pre->references, &i, pre->lock);
+    if (!CRYPTO_DOWN_REF(&pre->references, &i, pre->lock))
+        return;
     REF_PRINT_COUNT("EC_nistp256", x);
     if (i > 0)
         return;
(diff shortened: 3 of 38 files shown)


include/internal/refcount.h
```diff
@@ -171,4 +171,9 @@ typedef int CRYPTO_REF_COUNT;
 #  define REF_PRINT_COUNT(a, b)
 # endif
 
+
+#define OSSL_UP_REF(c)                                                         \
+int ref = 0;                                                                   \
+return CRYPTO_UP_REF(&c->refcnt, &ref, c->lock) <= 0 ? 0 : (ref > 0)
```
mspncp: I'm undecided whether this auxiliary define is really helpful or whether it just hides information in the locations where it is used. In particular, because it is a function-like macro which can't be used like a function (because of the extra declaration).

Also, IMHO the correct return statement is

```C
return CRYPTO_UP_REF(&c->refcnt, &ref, c->lock) ;
```
and not
```C
return CRYPTO_UP_REF(&c->refcnt, &ref, c->lock) <= 0 ? 0 : (ref > 0) ;
```
for two reasons:

* All variants of CRYPTO_UP_REF() (including the fallback which uses locks), return a boolean value (0 or 1).
* The check `ref > 0` adds extra semantics, which does not belong here.

slontis: There are quite a few existing ones that have the following pattern:

int DH_up_ref(DH *r)
{
    int i;

    if (CRYPTO_UP_REF(&r->references, &i, r->lock) <= 0)
        return 0;

    REF_PRINT_COUNT("DH", r);
    REF_ASSERT_ISNT(i < 2);
    return ((i > 1) ? 1 : 0);
}

slontis:
Added checks for down_ref calls.
Adding __owur could break someones code I would assume? So I am reluctant to do that..

crypto/ec/ecp_nistp256.c
```diff
@@ -1875,10 +1875,13 @@ static NISTP256_PRE_COMP *nistp256_pre_comp_new(void)
...
-    if (p != NULL)
-        CRYPTO_UP_REF(&p->references, &i, p->lock);
-    return p;
+    int ref = 0;
+
+    if (p != NULL) {
+        if (!CRYPTO_UP_REF(&p->references, &ref, p->lock))
+            return NULL;
+    }
+    return ref > 0 ? p : NULL;
```
levitte: Hmmm, thought experiment:

```
    int ref = 0;

    if (p == NULL
        || !CRYPTO_UP_REF(&p->references, &ref, p->lock)
        || ref <= 0)
        return NULL;
    return p;
```

I kinda find the single returns clearer.

Also, the check of `ref` on an UP_REF seems curious.  We *do* need to make such checks on DOWN_REF to see when further action is needed, but on UP_REF?

mspncp: The check for ref <= 0 is an overflow check, although I‘m not sure whether it is really necessary in practice.

paulidale: If we want to include an overflow check, it should be part of `CRYPTO_UP_REF` function.  That way it immediately goes everywhere.

Personally, I don't see a need for this but wouldn't object to it being put in.

bernd-edlinger:
Hmm, @slontis I see a lot of good ideas here,
but I think this is way too much for you to do alone.
If you don't mind, I would like to take this burden off
your shoulders, and make a half dozen PRs or so
from this and ask you to approve each one, since
that would bypass the stale-mate this project is in
now.

bernd-edlinger:
I wonder where the "hold" the openssl-machine away from this PR label is :-)

include/internal/refcount.h
bernd-edlinger: I never saw this funny check before.
I wonder if it is correct at all.  Will probably look at it
when I find time.  which may take some time....

crypto/cms/cms_env.c
```diff
@@ -198,8 +198,10 @@ static int cms_RecipientInfo_ktri_init(CMS_RecipientInfo *ri, X509 *recip,
     if (!cms_set1_SignerIdentifier(ktri->rid, recip, idtype))
         return 0;
 
-    X509_up_ref(recip);
-    EVP_PKEY_up_ref(pk);
+    if (!X509_up_ref(recip))
+        return 0;
+    if (!EVP_PKEY_up_ref(pk))
+        return 0;
```
kroeckx, line 204: I think the error handling is not correct. I expect that when this function returns failure that the effect is that no up_ref calls happened, including the other return 0 calls below.

kroeckx:
You might also want to look at doc/internal/man3/evp_generic_fetch.pod

kroeckx:
The return value of CRYPTO_atomic_add() is not documented, and documentation for CRYPTO_UP_REF() and CRYPTO_DOWN_REF() would be nice too.

include/internal/refcount.h
kroeckx: It seems that on _new(), references is set to 1, and it's debug code to check that you didn't do an up_ref for something where references was 0.

I think we should just replace it with: return CRYPTO_UP_REF(&r->references, &i, r->lock);

kroeckx:
> Adding __owur could break someones code I would assume? So I am reluctant to do that..

CRYPTO_UP_REF and CRYPTO_DOWN_REF are internal function, so you can just add it.

crypto/ec/ec_mult.c
```diff
@@ -77,10 +77,13 @@ static EC_PRE_COMP *ec_pre_comp_new(const EC_GROUP *group)
...
-    if (pre != NULL)
-        CRYPTO_UP_REF(&pre->references, &i, pre->lock);
-    return pre;
+    int ref = 0;
+
+    if (pre != NULL) {
+        if (!CRYPTO_UP_REF(&pre->references, &ref, pre->lock))
+            return NULL;
+    }
+    return ref > 0 ? pre : NULL;
```
kroeckx, line 86: Why not just return pre? If you get here and pre != NULL, the up_ref call was successful, ref was updated, and so ref should actually be > 1.

bernd-edlinger:
Thanks @kroeckx I will take these recommendations when I continue with this.
Do not expect this PR to be maintained any more, since I am stunned how
many files this touches in one go. I will recycle the ideas from this PR and create
smaller follow up PRs.  (when I find time)

crypto/cms/cms_env.c
bernd-edlinger: Yeah, absolutely

crypto/ec/ecp_nistp256.c
bernd-edlinger: I looked at doing that in CRYPTO_UP_REF, but it won't work becase
that is defined in tons of places, most of the time it is just a macro
calling CRYPTO_atomic_add, and while reference counts are never negative,
an atomic int is not a reference, and can therefore also be negative.
So I think the best we can do is like X509_up_ref:
```C
int X509_up_ref(X509 *x)
{
    int i;

    if (CRYPTO_UP_REF(&x->references, &i, x->lock) <= 0)
        return 0;

    REF_PRINT_COUNT("X509", x);
    REF_ASSERT_ISNT(i < 2);
    return ((i > 1) ? 1 : 0);
}
```

iamamoose:
> Thanks @kroeckx I will take these recommendations when I continue with this.
> Do not expect this PR to be maintained any more, since I am stunned how
> many files this touches in one go. I will recycle the ideas from this PR and create
> smaller follow up PRs. (when I find time)

therefore removing 'review pending' label (to help openssl-machine stop bugging)

PR #11279, opened 2020-03-08
DOCS: Fix the description of OSSL_PARAM_allocate_from_text()

### LINKED ISSUE #11276: OSSL_PARAM_allocate_from_text talks about non-existing varabiables
The OSSL_PARAM_allocate_from_text() manpage talks about `buf` and `buf_n`. I have no idea what is meant with them. OSSL_PARAM also doesn't contain such variables.

levitte:
Fixes #11276

### DIFF
--- a/doc/man3/OSSL_PARAM_allocate_from_text.pod
+++ b/doc/man3/OSSL_PARAM_allocate_from_text.pod
@@ -26,34 +26,74 @@ OpenSSL 3.0 introduces a new mechanism to do the same thing with an
 array of parameters that contain name, value, value type and value
 size (see L<OSSL_PARAM(3)> for more information).
 
-OSSL_PARAM_allocate_from_text() takes a control I<key>, I<value> and
-value size I<value_n>, and given a parameter descriptor array
-I<paramdefs>, it converts the value to something suitable for
-L<OSSL_PARAM(3)> and stores that in the buffer I<buf>, and modifies
-the parameter I<to> to match.
-I<buf_n>, if not NULL, will be assigned the number of bytes used in
-I<buf>.
-If I<buf> is NULL, only I<buf_n> will be modified, everything else is
-left untouched, allowing a caller to find out how large the buffer
-should be.
-I<buf> needs to be correctly aligned for the type of the B<OSSL_PARAM>
-I<key>.
-If <found> is not NULL, it is set to 1 if the parameter can be located and
-to 0 otherwise.
+OSSL_PARAM_allocate_from_text() uses I<key> to look up an item in
+I<paramdefs>.  If an item was found, it converts I<value> to something
+suitable for that item's I<data_type>, and stores the result in
+I<< to->data >> as well as its size in I<< to->data_size >>.
+I<< to->key >> and I<< to->data_type >> are assigned the corresponding
+values from the item that was found, and I<< to->return_size >> is set
+to zero.
 
-The caller must remember to free the data of I<to> when it's not
-useful any more.
+I<< to->data >> is always allocated using L<OPENSSL_zalloc(3)> and
+needs to be freed by the caller when it's not useful any more, using
+L<OPENSSL_free(3)>.
 
-For parameters having the type B<OSSL_PARAM_INTEGER>,
-B<OSSL_PARAM_UNSIGNED_INTEGER>, or B<OSSL_PARAM_OCTET_STRING>, both
-functions will interpret the I<value> differently if the key starts
-with "hex".
-In that case, the value is decoded first, and the result will be used
-as parameter value.
+If I<found> is not NULL, I<*found> is set to 1 if I<key> could be
+located in I<paramdefs>, and to 0 otherwise.
+
+=head2 The use of I<key> and I<value> in detail
+
+OSSL_PARAM_allocate_from_text() takes note if I<key> starts with
+"hex", and will only use the rest of I<key> to look up an item in
+I<paramdefs> in that case.  As an example, if I<key> is "hexid", "id"
+will be looked up in I<paramdefs>.
+
+When an item in I<paramdefs> has been found, I<value> is converted
+depending on that item's I<data_type>, as follows:
+
+=over 4
+
+=item B<OSSL_PARAM_INTEGER> and B<OSSL_PARAM_UNSIGNED_INTEGER>
+
+If I<key> started with "hex", I<value> is assumed to contain
+I<value_n> hexadecimal characters, which are decoded, and the
+resulting bytes become the number stored in the I<< to->data >>
+storage.
+
+If I<key> didn't start with "hex", I<value> is assumed to contain
+I<value_n> decimal characters, which are decoded, and the resulting
+bytes become the number stored in the I<< to->data >> storage.
+
+If I<value> contains characters that couldn't be decoded as
+hexadecimal or decimal characters, OSSL_PARAM_allocate_from_text()
+considers that an error.
+
+=item B<OSSL_PARAM_UTF8_STRING>
+
+If I<key> started with "hex", OSSL_PARAM_allocate_from_text()
+considers that an error.
+
+Otherwise, I<value> is considered a C string and is copied with no
+further checks to the I<< to->data >> storage.
+
+=item B<OSSL_PARAM_OCTET_STRING>
+
+If I<key> started with "hex", I<value> is assumed to contain
+I<value_n> hexadecimal characters, which are decoded, and the
+resulting bytes are stored in the I<< to->data >> storage.
+If I<value> contains characters that couldn't be decoded as
+hexadecimal or decimal characters, OSSL_PARAM_allocate_from_text()
+considers that an error.
+
+If I<key> didn't start with "hex", I<value_n> bytes from I<value> are
+copied to the I<< to->data >> storage.
+
+=back
 
 =head1 RETURN VALUES
 
-OSSL_PARAM_allocate_from_text() returns 1 on success, and 0 on error.
+OSSL_PARAM_allocate_from_text() returns 1 if I<key> was found in
+I<paramdefs> and there was no other failure, otherwise 0.
 
 =head1 NOTES
 
@@ -65,7 +105,7 @@ The following B<OSSL_PARAM> attributes are used:
 
 =item I<key>
 
-=item I<data>
+=item I<data_type>
 
 =item I<data_size>
 


doc/man3/OSSL_PARAM_allocate_from_text.pod
```diff
@@ -28,28 +28,21 @@ size (see L<OSSL_PARAM(3)> for more information).
...
 For parameters having the type B<OSSL_PARAM_INTEGER>,
-B<OSSL_PARAM_UNSIGNED_INTEGER>, or B<OSSL_PARAM_OCTET_STRING>, both
-functions will interpret the I<value> differently if the key starts
-with "hex".
-In that case, the value is decoded first, and the result will be used
-as parameter value.
+B<OSSL_PARAM_UNSIGNED_INTEGER>, or B<OSSL_PARAM_OCTET_STRING>,
+OSSL_PARAM_allocate_from_text() will interpret the I<value>
+differently if the key starts with "hex".  In that case, the value is
+hex decoded first, and the result will be used as parameter value.
```
kroeckx: I think this at least needs an example. I actually assume it start with "hex:", as in "hex:4F", and not "hex4F".

doc/man3/OSSL_PARAM_allocate_from_text.pod
```diff
@@ -28,28 +28,21 @@ size (see L<OSSL_PARAM(3)> for more information).
...
-should be.
-I<buf> needs to be correctly aligned for the type of the B<OSSL_PARAM>
-I<key>.
-If <found> is not NULL, it is set to 1 if the parameter can be located and
-to 0 otherwise.
-
-The caller must remember to free the data of I<to> when it's not
-useful any more.
+I<paramdefs>, it converts the value to something suitable for an
+L<OSSL_PARAM(3)> item, and stores the result in I<to>.
```
kroeckx: The "something suitable" gives the impression that it's not going to get stored in an OSSL_PARAM.

doc/man3/OSSL_PARAM_allocate_from_text.pod
```diff
@@ -28,28 +28,21 @@ size (see L<OSSL_PARAM(3)> for more information).
...
-left untouched, allowing a caller to find out how large the buffer
-should be.
-I<buf> needs to be correctly aligned for the type of the B<OSSL_PARAM>
-I<key>.
-If <found> is not NULL, it is set to 1 if the parameter can be located and
-to 0 otherwise.
-
-The caller must remember to free the data of I<to> when it's not
-useful any more.
+I<paramdefs>, it converts the value to something suitable for an
```
kroeckx: Maybe it should say something about looking up the value where key matching the key from paramdefs?

kroeckx:
What does error mean in this function? If the key wasn't found, is that an error? If so, why do we have that found variable?

doc/man3/OSSL_PARAM_allocate_from_text.pod
```diff
@@ -28,28 +28,21 @@ size (see L<OSSL_PARAM(3)> for more information).
...
+L<OSSL_PARAM(3)> item, and stores the result in I<to>.
+
+The I<data> field in I<to> allocated using L<OPENSSL_zalloc(3)> and
```
kroeckx: +is

doc/man3/OSSL_PARAM_allocate_from_text.pod
```diff
@@ -28,28 +28,21 @@ size (see L<OSSL_PARAM(3)> for more information).
...
+needs to be freed by the caller when it's not useful any omre, using
```
kroeckx: anymore, 1 word

kroeckx: Could also be 2 words, looks like anymore in 1 word might be used by some US people, and looks like everybody might be happy with the 2 words.

doc/man3/OSSL_PARAM_allocate_from_text.pod
levitte: Ah, thank you for pointing that out.

Well, yes and no. It's rather that this corresponds to things like `-pkeyopt id:A` vs `-pkeyopt hexid:41`. The key *starts with* "hex", and with OSSL_PARAM_allocate_from_text(), both forms will become an OSSL_PARAM item with the key "id" and exactly the same value.

doc/man3/OSSL_PARAM_allocate_from_text.pod
levitte: Er, how? *to* is a pointer to an OSSL_PARAM...

levitte:
> What does error mean in this function? If the key wasn't found, is that an error? If so, why do we have that found variable?

If the key wasn't found, it's an error. If memory couldn't be allocated, it's an error. However, those are two different types of errors, the former being recoverable by doing something else, while the latter isn't recoverable in the same way. This corresponds to the EVP_PKEY "magic" return value -2 for unsupported controls, vs 0 for "real errors". *found* is actually used in a ctrl_str function for exactly the purpose, to see if -2 or 0 should be returned.

doc/man3/OSSL_PARAM_allocate_from_text.pod
kroeckx: Just say it's going to get stored there, rather than making something that's suitable to store it there.

kroeckx:
It should at least document that not finding the key makes it return an error.

doc/man3/OSSL_PARAM_allocate_from_text.pod
levitte: Huh? What do you think "the result" refers to?

levitte:
> It should at least document that not finding the key makes it return an error.

Sure, no problem

doc/man3/OSSL_PARAM_allocate_from_text.pod
paulidale: `omre` should be `more`

doc/man3/OSSL_PARAM_allocate_from_text.pod
levitte: I've added a little more text that I hope clarifies this

doc/man3/OSSL_PARAM_allocate_from_text.pod
levitte: Looking up the value?  I don't understand...

doc/man3/OSSL_PARAM_allocate_from_text.pod
levitte: Fixed

doc/man3/OSSL_PARAM_allocate_from_text.pod
levitte: `omre` fixed

doc/man3/OSSL_PARAM_allocate_from_text.pod
kroeckx: I assume that in paramdefs, it's going to look for an OSSL_PARAM who's key == the key from the parameters

levitte: Yes. And it then uses that item's data type to figure out what to do with *value*. Is that what you meant?

kroeckx: Yes. And it's probably very obvious to you. But things like that should be documented.

levitte: Ok, I've rephrased that whole paragraph.  Does it seem better?

levitte: (and yeah, this *is* hard...  as you say, it's obvious to me, so it's hard to see what everyone else don't necessarily see.  So thank you for insisting)

doc/man3/OSSL_PARAM_allocate_from_text.pod
```diff
@@ -26,34 +26,39 @@ OpenSSL 3.0 introduces a new mechanism to do the same thing with an
...
-to 0 otherwise.
-
-The caller must remember to free the data of I<to> when it's not
-useful any more.
+OSSL_PARAM_allocate_from_text() looks up I<key> in I<paramdefs>.
+If an item was found, it converts I<value> (I<value_n> bytes) to
+something suitable for that item's I<data_type>, and stores the
+result in I<< to->data >> as well as its size in I<< to->data_size >>.
+I<< to->key >> and I<< to->data_type >> are copied from the item that
+was found, and I<< to->return_size >> is set to zero.
```
kroeckx: If to->key is copied, what does that mean for ownership of the pointer?

kroeckx: Later you document that if you pass "hexkey", it will contain "key", so I assume it's actually allocated, and needs to be freed too?

doc/man3/OSSL_PARAM_allocate_from_text.pod
```diff
@@ -26,34 +26,39 @@ OpenSSL 3.0 introduces a new mechanism to do the same thing with an
...
 For parameters having the type B<OSSL_PARAM_INTEGER>,
-B<OSSL_PARAM_UNSIGNED_INTEGER>, or B<OSSL_PARAM_OCTET_STRING>, both
-functions will interpret the I<value> differently if the key starts
-with "hex".
-In that case, the value is decoded first, and the result will be used
-as parameter value.
+B<OSSL_PARAM_UNSIGNED_INTEGER>, or B<OSSL_PARAM_OCTET_STRING>,
+OSSL_PARAM_allocate_from_text() will interpret the I<value>
+differently if the key starts with "hex".  In that case, the value is
+hex decoded first, and the result will be used as parameter value.
```
kroeckx: Is hex the only cases where things get converted, or are there other cases where it can convert something?

kroeckx: I guess it also converts from text to an integer ...

doc/man3/OSSL_PARAM_allocate_from_text.pod
levitte: Uhm, no. The pointer *to-\>key* becomes the same as the *key* pointer from the item that was found.

doc/man3/OSSL_PARAM_allocate_from_text.pod
levitte: I hear you, you want a detailed description of what gets converted to what

doc/man3/OSSL_PARAM_allocate_from_text.pod
levitte: Does it help if I say "assigned" instead of "copied"?  See the text change I made

doc/man3/OSSL_PARAM_allocate_from_text.pod
levitte: I've done another rewrite

doc/man3/OSSL_PARAM_allocate_from_text.pod
```diff
@@ -26,34 +26,74 @@ OpenSSL 3.0 introduces a new mechanism to do the same thing with an
...
+
+Otherwise, I<value> is considered a C string and is copied with no
+further checks to the I<< to->data >> storage.
+
+=item B<OSSL_PARAM_OCTET_STRING>
+
+If I<key> started with "hex", I<value> is assumed to contain
+I<value_n> hexadecimal characters, which are decoded, and the
+resulting bytes are stored in the I<< to->data >> storage.
+If I<value> containss characters that couldn't be decoded as
```
kroeckx: contains

kroeckx:
This looks much better

doc/man3/OSSL_PARAM_allocate_from_text.pod
levitte: Damn keyboard...  it randomly repeats certain characters (s and d are the most typical).  I guess it's showing age, but finding a new keyboard for a 5 year old laptop is hard.

levitte: Fixed

levitte:
> This looks much better

That makes me glad.  Thanks for your patience!

levitte:
So, final review anyone?

levitte:
Ping!

levitte:
Travis failure is the usual timeout

doc/man3/OSSL_PARAM_allocate_from_text.pod
```diff
@@ -26,34 +26,74 @@ OpenSSL 3.0 introduces a new mechanism to do the same thing with an
...
+
+If I<key> didn't start with "hex", I<value_n> bytes from I<value> are
+copied to the I<< to->data >> storage.
+
+=back
 
 =head1 RETURN VALUES
 
-OSSL_PARAM_allocate_from_text() returns 1 on success, and 0 on error.
+OSSL_PARAM_allocate_from_text() returns 1 on if I<key> was found in
```
mattcaswell: "on if"?

levitte: Fixed

levitte:
@t8m, does your approval stand?

t8m:
Yes.

PR #11282, opened 2020-03-09
Revert "Stop accepting certificates signed using SHA1 at security level 1" (1.1.1)

mattcaswell:
#10786 meant that we will stop accepting certificates signed using SHA1 at security level 1 in both master and 1.1.1.

The OMC subsequently voted on this issue, and did not vote in favour of backporting this change to 1.1.1. Therefore we should revert the change in that branch (it remains in master).


mattcaswell:
Pushed. Thanks.

PR #11285, opened 2020-03-09
fix check-format.pl false positive reporting '{1 stmt}' after multi-line if(expr)

### LINKED ISSUE #11284: util/check-format.pl
This code:

https://github.com/openssl/openssl/blob/99a16e0459e5089c2cfb92ee775f1221a51b8d05/providers/implementations/keymgmt/rsa_kmgmt.c#L400-L408

... generated this report:

```
providers/implementations/keymgmt/rsa_kmgmt.c:402:'for' { 1 stmt }:        ok = ok && rsa_validate_pairwise(rsa);
```

That doesn't seem quite right, since the `else` part contains multiple statements.

DDvO:
fixes #11284

This PR also marks `util/check-format.pl` as executable (new mode 100755).


DDvO:
I've added a fix for a further (minor) false positive
and extended the negative tests checking for false positives.

levitte:
I can't say that I quite understand the change, but the test files do show triggering examples, so...

DDvO:
Pushed - thanks!

PR #11288, opened 2020-03-09
Move legacy code blocks into seperate functions

slontis:
Experimental test to try to clean up legacy code.

##### Checklist

### DIFF
--- a/crypto/evp/digest.c
+++ b/crypto/evp/digest.c
@@ -1,5 +1,5 @@
 /*
- * Copyright 1995-2019 The OpenSSL Project Authors. All Rights Reserved.
+ * Copyright 1995-2020 The OpenSSL Project Authors. All Rights Reserved.
  *
  * Licensed under the Apache License 2.0 (the "License").  You may not use
  * this file except in compliance with the License.  You can obtain a copy
@@ -17,6 +17,7 @@
 #include "crypto/evp.h"
 #include "internal/provider.h"
 #include "evp_local.h"
+#include "legacy_meth.h"
 
 /* This call frees resources associated with the context */
 int EVP_MD_CTX_reset(EVP_MD_CTX *ctx)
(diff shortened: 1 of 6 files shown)


crypto/evp/digest.c
```diff
@@ -113,56 +90,15 @@ int EVP_DigestInit_ex(EVP_MD_CTX *ctx, const EVP_MD *type, ENGINE *impl)
...
 
-    if (type != NULL) {
-        /*
-         * Ensure an ENGINE left lying around from last time is cleared (the
-         * previous check attempted to avoid this if the same ENGINE and
-         * EVP_MD could be used).
-         */
-        ENGINE_finish(ctx->engine);
-        ctx->engine = NULL;
+        if (legacy_evp_digest_init_ex(ctx, type, impl, &ret))
```
levitte: I looked at `legacy_evp_digest_init_ex`, and noticed that it does look at fetched stuff, and that might be overdoing stuff a bit.  So it seems to me that instead of a "do everything" function, it might be better with two, kind of "prepare" and "execute", so to say.  In code, I would expect to something along the lines of:

``` C
        if ((type = legacy_evp_digest_get_impl(ctx, type, impl)) != NULL)
            return legacy_evp_digest_init_ex(ctx, type);
```

`legacy_evp_digest_get_impl()` would then be able to get an engine implementation if there is such a thing, for example from an engine, and if none is available, it would simply return `type`.  `legacy_evp_digest_init_ex()` would then execute the initialization itself.

levitte: Note that with that kind of division, there's also no need for `&ret`...

crypto/evp/digest.c
```diff
@@ -303,26 +162,27 @@ int EVP_DigestUpdate(EVP_MD_CTX *ctx, const void *data, size_t count)
...
 
-    if (ctx->digest == NULL
-            || ctx->digest->prov == NULL
-            || (ctx->flags & EVP_MD_CTX_FLAG_NO_INIT) != 0)
-        goto legacy;
+#ifndef FIPS_MODE
+    {
+        int ret = 0;
+
+        if (legacy_evp_digest_update(ctx, data, count, &ret))
```
levitte: Similarly, one might look at this as an alternative:

``` C
        if (legacy_evp_digest_is_legacy(ctx))
            return legacy_evp_digest_update(ctx, data, count);
```

levitte: ... and similarly for the final call, for ciphers, etc...

crypto/evp/digest.c
slontis: That doesnt really work for legacy_evp_digest_init_ex() since there is a whole skip_to_init: branch that doesnt use the type.

crypto/evp/digest.c
slontis: Even this simple one doesnt quite work... What if ctx->digest is NULL?

    if (ctx->digest == NULL)
        goto err;
    if (ctx->digest->prov != NULL)

crypto/evp/digest.c
levitte: Yeah I know, it's not the easiest code to form as we wish, and I'm not saying that for the lack of trying.

I'm way too tired to be constructive right now, so I'll get tomorrow, yeah?

slontis:
Probably not worth doing - so closing this PR.

PR #11296, opened 2020-03-10
If SOURCE_DATE_EPOCH is defined, use it for copyright year

boklm:
Using the date from SOURCE_DATE_EPOCH instead of the current date makes
it possible to reproduce a build that was built on a different year:
https://reproducible-builds.org/specs/source-date-epoch/

This is fixing an issue we had while building Tor Browser:
https://trac.torproject.org/projects/tor/ticket/33535

### DIFF
--- a/util/mkrc.pl
+++ b/util/mkrc.pl
@@ -25,7 +25,7 @@ if ( $filename =~ /openssl/i ) {
     $vft = "VFT_APP";
 }
 
-my $YEAR = [localtime()]->[5] + 1900;
+my $YEAR = [gmtime($ENV{SOURCE_DATE_EPOCH} || time())]->[5] + 1900;
 print <<___;
 #include <winver.h>
 


kroeckx:
I assume we also want this in the 1-1 branch.

Can someone explain which file contains the year?

mspncp:
> Can someone explain which file contains the year?

```
~/src/openssl$ git grep mkrc
apps/build.info:    GENERATE[openssl.rc]=../util/mkrc.pl openssl
build.info:  GENERATE[libcrypto.rc]=util/mkrc.pl libcrypto
build.info:  GENERATE[libssl.rc]=util/mkrc.pl libssl
```

util/mkrc.pl
```diff
@@ -25,7 +25,7 @@
     $vft = "VFT_APP";
 }
 
-my $YEAR = [localtime()]->[5] + 1900;
+my $YEAR = [localtime($ENV{SOURCE_DATE_EPOCH} || time())]->[5] + 1900;
```
kaduk, line 28: Doesn't localtime() depend on the ... local time zone?  I.e., is not actually reproducible for epoch values near a year transition, depending on the time zone currently in use?

richsalz: you're probably right but who cares. ;)

t8m: Then what about just replacing localtime() with gmtime()? Not that it matters much, just for a strict correctness.

boklm: I updated the commit to use gmtime() instead of localtime().

mattcaswell:
Pushed to master and 1.1.1.

PR #11299, opened 2020-03-10
doc: Update the reference from draft to RFC

Jakuje:
The [RFC 8410](https://tools.ietf.org/html/rfc8410) was published more than a year ago and referring to old draft versions is at least confusing.


levitte:
Formality: I agree this is trivial

kroeckx:
I agree it's trivial

t8m:
Merged to master (c08dea30d4) and 1.1.1 (9011225188). Thank you.

PR #11300, opened 2020-03-10
Chunk 9 of CMP contribution to OpenSSL: CMP client and related tests

DDvO:
Certificate Management Protocol (CMP, RFC 4210) extension to OpenSSL
Also includes CRMF (RFC 4211) and HTTP transfer (RFC 6712).
Adds the CMP and CRMF API to libcrypto and the "cmp" app to the CLI.
Adds extensive documentation and tests.
 
9th chunk: CMP client library code and related tests

##### Checklist


DDvO:
I just realized that the case
```
#if !defined(OPENSSL_NO_OCSP)
```
is not needed any more for the CMP client :)

crypto/cmp/cmp_client.c
```diff
@@ -0,0 +1,878 @@
...
+                if (time_left < check_after)
+                    check_after = time_left;
+                /* poll one last time just when timeout was reached */
+            }
+
+            OSSL_CMP_MSG_free(preq);
+            preq = NULL;
+            OSSL_CMP_MSG_free(prep);
+            prep = NULL;
+            sleep((unsigned int)check_after);
```
mattcaswell: Not sure I like a sleep here. It provides no opportunity for the application to do async processing and go off and do something else while it is waiting.

crypto/cmp/cmp_client.c
```diff
@@ -0,0 +1,878 @@
...
+{
+    OSSL_CMP_MSG *certConf = NULL;
+    OSSL_CMP_MSG *PKIconf = NULL;
+    int res = 0;
+
+    /* OSSL_CMP_certConf_new() also checks if all necessary options are set */
+    if ((certConf = ossl_cmp_certConf_new(ctx, fail_info, txt)) == NULL)
+        goto err;
+
+    res = send_receive_check(ctx, certConf, &PKIconf, OSSL_CMP_PKIBODY_PKICONF,
```
mattcaswell: It seems we throw away the returned value in the third argument in a few of these calls. I wonder whether it is worth allowing NULL to be provided there.

doc/man3/OSSL_CMP_exec_IR_ses.pod
```diff
@@ -0,0 +1,108 @@
...
+ X509 *OSSL_CMP_exec_KUR_ses(OSSL_CMP_CTX *ctx);
+ X509 *OSSL_CMP_exec_RR_ses(OSSL_CMP_CTX *ctx);
+ STACK_OF(OSSL_CMP_ITAV) *OSSL_CMP_exec_GENM_ses(OSSL_CMP_CTX *ctx;
+ int OSSL_CMP_certConf_cb(OSSL_CMP_CTX *ctx, X509 *cert, int fail_info,
+                          const char **text);
+
+=head1 DESCRIPTION
+
+This is the API for doing CMP (Certificate Management Protocol)  client-server
+message sequences using with OpenSSL.
```
mattcaswell: Just "using OpenSSL".

doc/man3/OSSL_CMP_exec_IR_ses.pod
```diff
@@ -0,0 +1,108 @@
...
+The default conveying protocol is HTTP.
+
+OSSL_CMP_exec_IR_ses() does the transaction which obtains an initial certificate
+from the CA.
+A populated OSSL_CMP_CTX structure has to be given as argument.
```
mattcaswell: "as *an* argument". Similarly below

test/cmp_client_test.c
```diff
@@ -0,0 +1,342 @@
...
+    ADD_TEST(test_exec_CR_ses);
+    ADD_TEST(test_exec_CR_ses_implicit_confirm);
+    ADD_TEST(test_exec_IR_ses);
+    ADD_TEST(test_exec_IR_ses_poll);
+    ADD_TEST(test_exec_IR_ses_poll_timeout);
+    ADD_TEST(test_exec_KUR_ses);
+    ADD_TEST(test_exec_P10CR_ses);
+    ADD_TEST(test_exec_GENM_ses);
+    ADD_TEST(test_exchange_certConf);
+    ADD_TEST(test_exchange_error);
```
mattcaswell: Does this require the tests to be run in exactly that order for this to work? Could be a problem if `OPENSSL_TEST_RAND_ORDER` is set.

crypto/cmp/cmp_client.c
```diff
@@ -0,0 +1,878 @@
...
+#include <openssl/bio.h>
+#include <openssl/cmp.h>
+#include <openssl/err.h>
+#include <openssl/evp.h>
+#include <openssl/x509v3.h>
+
+#include "openssl/cmp_util.h"
+
+#define IS_CREP(t) (t == OSSL_CMP_PKIBODY_IP || t == OSSL_CMP_PKIBODY_CP \
+                        || t == OSSL_CMP_PKIBODY_KUP)
```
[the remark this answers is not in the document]
DDvO: Done.

crypto/cmp/cmp_client.c
```diff
@@ -0,0 +1,878 @@
...
+/*
+ * Evaluate whether there's an standard-violating exception configured for
+ * handling errors without protection or with invalid protection
+ */
+static int unprotected_exception(const OSSL_CMP_CTX *ctx,
+                                 const OSSL_CMP_MSG *rep,
+                                 int invalid_protection, int expected_type)
+{
+    int rcvd_type = ossl_cmp_msg_get_bodytype(rep /* may be NULL */);
+    char *msg_type = NULL;
```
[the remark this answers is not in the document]
DDvO: Yep, done.

crypto/cmp/cmp_client.c
```diff
@@ -0,0 +1,878 @@
...
+    const char *req_type_string =
+        ossl_cmp_bodytype_to_string(ossl_cmp_msg_get_bodytype(req));
+    int msgtimeout;
+    int err = 0;
+    int bt;
+    OSSL_cmp_transfer_cb_t transfer_cb = ctx->transfer_cb;
+
+    if (transfer_cb == NULL) {
+        transfer_cb = NULL; /* TODO: will be OSSL_CMP_MSG_http_perform of chunk 10 */
+    }
```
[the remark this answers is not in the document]
DDvO: Fixed.

crypto/cmp/cmp_client.c
```diff
@@ -0,0 +1,878 @@
...
+
+    msgtimeout = ctx->msgtimeout; /* backup original value */
+    if ((IS_CREP(expected_type) || expected_type == OSSL_CMP_PKIBODY_POLLREP)
+            && ctx->totaltimeout > 0 /* total timeout is not infinite */) {
+        int64_t time_left = (int64_t)(ctx->end_time - time(NULL));
```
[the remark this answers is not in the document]
DDvO: Good spotting - this is a bug.
I've changed this to do the comparison beforehand.

crypto/cmp/cmp_client.c
```diff
@@ -0,0 +1,878 @@
...
+
+        if (time_left <= 0) {
+            CMPerr(0, CMP_R_TOTAL_TIMEOUT);
+            return 0;
+        }
+        if (ctx->msgtimeout == 0 || time_left < ctx->msgtimeout)
+            ctx->msgtimeout = (int)time_left;
```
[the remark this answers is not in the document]
DDvO: In this case the difference is guaranteed to be `< INT_MAX`.
I've made this clear adding an assertion and an explaining comment.

crypto/cmp/cmp_client.c
```diff
@@ -0,0 +1,878 @@
...
+    }
+
+    /* should print error queue since transfer_cb may call ERR_clear_error() */
+#ifndef OPENSSL_NO_STDIO
```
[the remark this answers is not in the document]
DDvO: Indeed meanwhile the guard is not needed any more; removed.

crypto/cmp/cmp_client.c
```diff
@@ -0,0 +1,878 @@
...
+    OSSL_CMP_CTX_print_errors(ctx);
+#endif
+
+    ossl_cmp_log1(INFO, ctx, "sending %s", req_type_string);
+
+    if ((*rep = (*transfer_cb)(ctx, req)) == NULL) {
+        err = ERR_GET_REASON(ERR_peek_error());
```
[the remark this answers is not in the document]
DDvO: Good spotting! 
I've added the following line to the documentation:
```
On error it should call B<CMPerr()> with a suitable reason code and return NULL.
```
and added to the code:
```
        if (err == 0) /* the application failed to provide a reason */
            err = CMP_R_TRANSFER_ERROR;
```

crypto/cmp/cmp_client.c
```diff
@@ -0,0 +1,878 @@
...
+    }
+    ctx->msgtimeout = msgtimeout; /* restore original value */
+
+    if (err != 0) {
+        if (err == CMP_R_FAILED_TO_RECEIVE_PKIMESSAGE
+                || err == CMP_R_READ_TIMEOUT
+                || err == CMP_R_ERROR_DECODING_MESSAGE) {
```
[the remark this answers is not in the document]
DDvO: I agree. I've simplified the error reporting.

crypto/cmp/cmp_client.c
```diff
@@ -0,0 +1,878 @@
...
+ * A total timeout may have been set in the context.  The function will continue
+ * to poll until the timeout is reached and then poll a last time even when that
+ * is before the "checkAfter" sent by the server. If ctx->totaltimeout is 0, the
+ * timeout is disabled.
+ *
+ * returns 1 on success, returns received PKIMESSAGE in *msg argument
+ * returns 0 on error or when timeout is reached without a received message
+ *
+ * TODO: handle multiple poll requests for multiple certificates
+ *       --> GitHub issue#67
```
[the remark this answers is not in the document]
DDvO: Ah, sorry I forgot to convert all those references in the same way as in `cmp_server.c`. Done.

crypto/cmp/cmp_client.c
DDvO: Unfortunately the way the API is designed so far there is not much choice.
I'll try coming up with an alternative that does not change too much...

crypto/cmp/cmp_client.c
DDvO: The response message is thrown away only in case of a `PKIConf` response, which has an empty body and is basically just an ACK.
I've checked: allowing a NULL pointer result variable would just complicate matters because then some case distinctions in `send_receive_check()` would be needed and in any case the response needs to be deallocated.

doc/man3/OSSL_CMP_exec_IR_ses.pod
DDvO: Fixed.

doc/man3/OSSL_CMP_exec_IR_ses.pod
DDvO: Fixed and simplified the wording in all these cases.

test/cmp_client_test.c
DDvO: Good thought. The tests are designed to be independent of each other.
I've tried this out reverting their order manually and everything still worked as expected.

DDvO:
Thank @mattcaswell for your swift and helpful review comments!
I've solved all issues - except for the deeper structural one: what to do while waiting for certificate responses - and rebased the PR.

crypto/cmp/cmp_client.c
DDvO: I plan to add a more flexible CMP cert enrollment function:
```
int OSSL_CMP_try_certreq(OSSL_CMP_CTX *ctx, int req_type, int wait);
```
Its input is the CMP context as usual, the type of cert request (ir/cr/p10cr/kur), 
and the remaining time to wait after a previous call, or `-1` for the initial call.
It returns `0` as soon as the requested cert is available (then the application can retrieve it using `OSSL_CMP_CTX_get0_newCert()`)
or else the amount of time to wait before trying again.
In the latter case the application can decide to sleep for some time and/or do whatever before retrying. Then it calls the same function again, giving any remaining time to wait before trying again or `0` for immediate retry.

@mattcaswell, how does this sound?

DDvO:
In the meantime: are there any other open issues with this PR?

crypto/cmp/cmp_client.c
DDvO: I've meanwhile conceived the following enhanced API design with improved usability and flexibility, which I'm going to implement today:
```
int OSSL_CMP_try_certreq(OSSL_CMP_CTX *ctx, int req_type);
```
Its input is the CMP context as usual and the type of cert request (ir/cr/p10cr/kur).
It returns 0 on error or -1 if the requested cert is available. Thenthe application can use `OSSL_CMP_CTX_get0_newCert()` to retrieve the new certificate (which will yield `NULL` in case an error occurred).
If no error occurred but no certificate is available yet then `OSSL_CMP_try_certreq()` remembers in the CMP context the point in time whan polling should be retried and returns a positive value indicating the number of seconds the application should let pass before trying again. In this case the application can decide to sleep for some time and/or do anything else before retrying. Then it calls the same function again with the same parameter values as before. In this case `OSSL_CMP_try_certreq()` sleeps until the point in time remembered has passed and then polls to see if meanwhile the requested certificate is available.
If the application decides to abort its pending certificate request it can use a different request type (or any other value, such as -1). In this case `OSSL_CMP_try_certreq()` aborts the CMP transaction by sending an error message to the server.

@mpeylo, I hope you also like this solution?

crypto/cmp/cmp_client.c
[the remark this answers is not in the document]
mattcaswell: Hmmmm......Macros like CMPerr are intended for use by internal code. The fact that they are public is entirely an accident of history which we should probably correct at some point. I am very hesitant to *encourage* end applications to call these macros. This does not seem like a good idea. Instead I would prefer to see some separate return code given back with macros defined for different types of errors that might occur. Peeking into the OpenSSL error stack like this doesn't seem like the right design.

crypto/cmp/cmp_client.c
mattcaswell: This sounds like a cleaner solution.

crypto/cmp/cmp_client.c
mattcaswell: Ok

crypto/cmp/cmp_client.c
```diff
@@ -0,0 +1,833 @@
...
+            goto err;
+        }
+        if ((cid = ossl_cmp_revrepcontent_get_CertId(rrep, rsid)) == NULL) {
+            result = NULL;
+            goto err;
+        }
+        if (X509_NAME_cmp(issuer, OSSL_CRMF_CERTID_get0_issuer(cid)) != 0
+                || ASN1_INTEGER_cmp(serial,
+                                    OSSL_CRMF_CERTID_get0_serialNumber(cid))
+            != 0) {
```
mattcaswell: This looks incorrectly indented

doc/man3/OSSL_CMP_SRV_CTX_new.pod
```diff
@@ -128,6 +128,11 @@ confirmation of newly enrolled certificates if requested.
 
 CMP is defined in RFC 4210 (and CRMF in RFC 4211).
 
+=head1 BUGS
+
+So far the CMP server implementation is limited to one request per CMP message
+(and consequently to at most one response component per CMP message).
```
mattcaswell: This isn't so much a bug as a design limitation. Probably this shouldn't be in the "BUGS" section - but just in the general description.

crypto/cmp/cmp_client.c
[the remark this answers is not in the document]
DDvO: I see. So I've reverted the addition to the documentation
and changed the code in the client to
```
    if ((*rep = (*transfer_cb)(ctx, req)) == NULL)
        CMPerr(0, CMP_R_TRANSFER_ERROR);
```
i.e. there is no dependency on the callback w.r.t elements on the error queue.

crypto/cmp/cmp_client.c
DDvO: Emacs and the format checker were fine with that, but I also had thought that it looks odd.
I've re-structured the code at this point, which as useful anyway to fix a misleading error reason, which made the issue disappear.

doc/man3/OSSL_CMP_SRV_CTX_new.pod
DDvO: Yeah, I had been looking for something like a LIMITS section, but did not find any,
so resorted to (ab-)using BUGS, which apparently had already done occasionally for other types of limitations.
I've just removed the BUGS section indicator, such that the text pointing out the limitation has become part of the NOTES section :)

DDvO:
> I've taken another scan through this PR and noted a couple of minor issues below.

Thanks - I've just fixed these, along with various minor naming inconsistencies I realizied.
Moreover I've extended the documentation of the new CMP client functions.

> As noted above I continue to have some concerns about the transfer_cb API and how errors are handled. 

I've addressed these as well.

> Other than that and the outstanding issue about what to do while waiting for certificate responses, I'm ok with this PR.

Very pleased to hear!
I was planning to re-structure the code to address the waiting issue today, 
but was held up by other things and will do by Monday...

crypto/cmp/cmp_client.c
DDvO: Pleased to hear. After discussion with a colleague I've slightly revised the design further according to my edit/update of the above description (because the application should be able to overrule/ignore the check-after time provided by the server).

mattcaswell:
Note - the AppVeyor failure looks relevant:

````
..\crypto\cmp\cmp_client.c(301): error C2220: warning treated as error - no 'object' file generated
..\crypto\cmp\cmp_client.c(301): warning C4244: 'function': conversion from 'int64_t' to 'unsigned long', possible loss of data
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual 
````

**Comment by DDvO** [comment]
> Note - the AppVeyor failure looks relevant:

Thx for the hint; this will be solved with my next update..

File: crypto/cmp/cmp_client.c (hunk 1 above)
Comment (DDvO) [reply crypto/cmp/cmp_client.c]: @mattcaswell and @Akretsch, finally the new function `OSSL_CMP_try_certreq()`, including documentation and tests, is ready for review.

File: crypto/cmp/cmp_client.c [hunk 17]
```diff
@@ -0,0 +1,877 @@
...
+    /* should print error queue since transfer_cb may call ERR_clear_error() */
+    OSSL_CMP_CTX_print_errors(ctx);
+
+    ossl_cmp_log1(INFO, ctx, "sending %s", req_type_str);
+
+    if ((*rep = (*transfer_cb)(ctx, req)) == NULL)
+        CMPerr(0, CMP_R_TRANSFER_ERROR);
+    ctx->msg_timeout = msg_timeout; /* restore original value */
+
+    if (*rep == NULL) {
```
Akretsch: rep != NULL ?

crypto/cmp/cmp_client.c
```diff
@@ -0,0 +1,877 @@
...
+    }
+    OPENSSL_free(subj);
+    return ret;
+}
+
+int OSSL_CMP_try_certreq(OSSL_CMP_CTX *ctx, int req_type)
+{
+    OSSL_CMP_MSG *req = NULL;
+    OSSL_CMP_MSG *rep = NULL;
+    int is_p10 = req_type == OSSL_CMP_PKIBODY_P10CR;
```
Akretsch: maybe

if (req_type == OSSL_CMP_PKIBODY_P10CR) {
...
} else {
...
}

is more clear

crypto/cmp/cmp_client.c
DDvO: No, error handling and return is needed when getting the response (`*rep`) did not work.

DDvO: Meanwhile I see what the actual issue was: duplicate error throwing -
thanks, fixed!

crypto/cmp/cmp_client.c
DDvO: Well, the proposed alternative would need 6 more lines, 
therefore I'd prefer the pretty succinct version given. 
As your 'maybe' suggests, it' a matter of taste.

DDvO:
Thanks @Akretsch for your two review comments, which I've handled. Did this finish your pass?

DDvO:
@mattcaswell, are you fine with the implementation (including tests and documentation) of the new `OSSL_CMP_try_certreq()` function?

crypto/cmp/cmp_client.c
Akretsch: Thanks!

crypto/cmp/cmp_client.c
Akretsch: OK for me!

doc/man3/OSSL_CMP_exec_IR_ses.pod
```diff
@@ -0,0 +1,168 @@
...
+OSSL_CMP_exec_P10CR_ses(), and OSSL_CMP_exec_KUR_ses() return a
+pointer to the newly obtained X509 certificate on success, B<NULL> on error.
+This pointer will be freed implicitly by OSSL_CMP_CTX_free() or
+CSSL_CMP_CTX_reinit().
+
+OSSL_CMP_try_certreq() returns -1 if the requested certificate is available
+via B<OSSL_CMP_CTX_get0_newCert()>, the checkAfter value (which is > 0)
+in case a 'waiting' status has been received (then B<OSSL_CMP_CTX_get0_newCert()>
+yields NULL),
+1 on successfully aborting a pending certificate request, and 0 on error.
```
mattcaswell: I don't particularly like overloading the return value to have all these different meanings. The certificate is available sounds like a success to me - so returning -1 to indicate it just seems odd and counter-intuitive. Perhaps better might be -1 for error, 0 for waiting and 1 for the certificate is available might be a more intuitive interface. In the event of "waiting" then there could be an additional function on the ctx to say how long is left until the application should poll again.

DDvO: Yeah, I agree that the various sorts of return values are a bit complicated.
I'll follow the -1 / 0 / +1 value you suggested, yet I'd like to a avoid adding a further function just for retrieving the checkAfter value - not only this would add one more function to the API but mainly because the meaning of the value could go wrong in case the app takes some time before retrieving it. Ok if I add instead an output parameter `int *checkAfter`?

mattcaswell: Yes - that would work too.

doc/man3/OSSL_CMP_exec_IR_ses.pod
```diff
@@ -0,0 +1,168 @@
...
+
+OSSL_CMP_exec_P10CR_ses() conveys a legacy PKCS#10 CSR requesting a certificate.
+
+OSSL_CMP_exec_KUR_ses() obtains an updated certificate.
+
+All these four types of certificate enrollment may be blocked by sleeping until the
+CAs or an intermedate PKI component can fully process and answer the request.
+
+OSSL_CMP_try_certreq() is an alternative to these four functions with
+more uniformity w.r.t. the certificate request type and
```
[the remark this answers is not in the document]
DDvO: I've changed this to:

OSSL_CMP_try_certreq() is an alternative to these four functions that is
more uniform regarding the type of the certificate request to use and
more flexible regarding what to do after receiving a checkAfter value.

doc/man3/OSSL_CMP_exec_IR_ses.pod
DDvO: Finished changing the result value structure.
It turns out that it is more straightforward to use - as usual - 1 for success and 0 for error, 
while -1 indicates the 'waiting' condition (in which case `*checkAfter` is assigned unless `checkAfter == NULL`).

crypto/cmp/cmp_client.c
```diff
@@ -0,0 +1,881 @@
...
+                > (sleep ? ULONG_MAX / 1000 : INT_MAX)) {
+                CMPerr(0, CMP_R_CHECKAFTER_OUT_OF_RANGE);
+                if (BIO_snprintf(str, OSSL_CMP_PKISI_BUFLEN, "value = %ld",
+                                 check_after) >= 0)
+                    ERR_add_error_data(1, str);
+                goto err;
+            }
+            if (ctx->total_timeout > 0) { /* timeout is not infinite */
+                const int exp = 5; /* expected max time per msg round trip */
+                int64_t time_left = (int64_t)(ctx->end_time - exp - time(NULL));
```
Akretsch: "-" followed by "-" might be confusing, maybe parenthesis could be useful for beginners?

crypto/cmp/cmp_client.c
```diff
@@ -0,0 +1,881 @@
...
+/* Send certConf for IR, CR or KUR sequences and check response */
+int ossl_cmp_exchange_certConf(OSSL_CMP_CTX *ctx, int fail_info,
+                               const char *txt)
+{
+    OSSL_CMP_MSG *certConf = NULL;
+    OSSL_CMP_MSG *PKIconf = NULL;
+    int res = 0;
+
+    /* OSSL_CMP_certConf_new() also checks if all necessary options are set */
+    if ((certConf = ossl_cmp_certConf_new(ctx, fail_info, txt)) == NULL)
```
Akretsch: previous initialization of certConf not really needed.

crypto/cmp/cmp_client.c
```diff
@@ -0,0 +1,881 @@
...
+                /* poll one last time just when timeout was reached */
+            }
+
+            if (pollRep->reason == NULL
+                    || (len = BIO_snprintf(str, OSSL_CMP_PKISI_BUFLEN,
+                                           " with reason = '")) < 0) {
+                *str = '\0';
+            } else {
+                char *text = sk_ASN1_UTF8STRING2text(pollRep->reason, ", ",
+                                                     sizeof(str) - len - 2);
```
Akretsch: len could be uninitilized here

crypto/cmp/cmp_client.c
```diff
@@ -0,0 +1,881 @@
...
+    OSSL_CMP_MSG *genm = NULL;
+    OSSL_CMP_MSG *genp = NULL;
+    STACK_OF(OSSL_CMP_ITAV) *rcvd_itavs = NULL;
+
+    if (ctx == NULL) {
+        CMPerr(0, CMP_R_INVALID_ARGS);
+        return 0;
+    }
+
+    if ((genm = ossl_cmp_genm_new(ctx)) == NULL)
```
Akretsch: previous initialization of genm not really needed.

crypto/cmp/cmp_client.c
DDvO: You mean, x - y - z being equal to x - (y + z) could be confusing?
I'd say it might be - to beginners of arithmetic at 7th grade in school ;-)

crypto/cmp/cmp_client.c
DDvO: Ok, removed.

crypto/cmp/cmp_client.c
DDvO: Removed as well.

crypto/cmp/cmp_client.c
DDvO: There is no issue with uninitialized `len` here because the `else` branch is not reached if `pollRep->reason == NULL` and otherwise `len` is assigned by the second disjunct of the `if` expression.

DDvO:
@mattcaswell, yesterday there was the usual false positive of Travis CI due to timeout; everything else went fine.
I've just done two cosmetic improvements according to @Akretsch's review and pushed the result, which gives the CI a new chance to report all-green.

Anything else to do on this chunk from your perspective?

DDvO:
Travis CI went fine this morning.
I've just rebased the PR and added two minor fixup commits that came up during rebasing #10587.

DDvO:
Rebased again, on the just merged #10504.

DDvO:
Great, thanks @mattcaswell 

In preparation of merging this chunk tomorrow I've squashed it as much as I believe makes sense.

DDvO:
Merged, closing :)

bernd-edlinger:
Sorry, can we please revert this one, it shot down my #11393 
which addresses the same issue but in a slightly better way.

mattcaswell:
Given the size of this PR compared to #11393, it makes more sense to adjust #11393.

bernd-edlinger:
Okay, I have not looked at the whole picture.

DDvO:
I deliberately kept my change to `strncpy` in a separate commit: 
b4ba2b7, so it could be reverted easily.
Yet I do not see that replacing the function by `memcpy` would be better.

bernd-edlinger:
But in general I do not understand why every PR addresses so many different
topics at the same time.  What is this good for?

DDvO:
It was a fixup of a CMP-related change.

bernd-edlinger:
strcpy implies zero-termination.  memcpy does not,
that is the reason why there was this warning in the first place.
also the strcpy below is funny, because it implies
okay. then I close my PR.  I  can impossibly keep up with your speed.

DDvO:
AFAICE using `memcpy` with the same length is wrong.
In order to copy over the NUL termination you need to copy `sep_len + 1` bytes/chars.

DDvO:
Semantically neither of the fixes is/was needed.
Just some compilers brought a (in this case misleading) warning. 
BTW, also clang did so, which motivated my change.

DDvO:
>  also the strcpy below is funny, because it implies

Indeed. Yet interestingly at least clang did not warn on
```
strncpy(p, (const char *)ASN1_STRING_get0_data(current), length);
```
For this line I agree that using `memcpy` would at least look better because apparently it is not guaranteed that that `ASN1_STRING_get0_data()` is NUL-terminated.
Yet, also here it is semantically not relevant due to the final assignment `*p = '\0'`.

bernd-edlinger:
The warning is not semantically misleading, but as I said, go with your change,
my life does not depend on the memcpy

DDvO:
BTW, can you give me a hint how to use gcc-10 on the latest Debian edition (10.3 "buster")?

bernd-edlinger:
build it from source. that is not yet released.

DDvO:
After getting around some build issues I meanwhile have a working `gcc (GCC) 10.0.1 20200326`,
and my SSD has some 9.5 GB less space.

OpenSSL builds (with `--strict-warnings` etc.) based on PR#11404 (which includes the merged commits of this PR#11300) with gcc-10.0.1 went fine without warnings.

Issue #1124, opened 2016-05-25
Cannot compile current trunk version on linux for windows with mingw


wolftobias:
I`m compiling on linux for windows with mingw, but with the current trunk version it fails.
```
Can't locate IPC/Cmd.pm in @INC (@INC contains: /usr/local/lib64/perl5 /usr/local/share/perl5 /usr/lib64/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib64/perl5 /usr/share/perl5 .) at ./Configure line 17.
BEGIN failed--compilation aborted at ./Configure line 17.
make: *** No targets specified and no makefile found.  Stop.
make: *** No rule to make target `install'.  Stop.
```

levitte:
What perl version do you use?

wolftobias:
I send you the output underneath. The compile is working fine with the latest beta version openssl-1.1.0-pre5.
```
perl --version

This is perl, v5.10.1 (*) built for x86_64-linux-thread-multi
```

levitte:
May I suggest you try with a fresher checkout?  The handling of IPC::Cmd was altered a few days ago, so we now have a fallback.

wolftobias:
Yes for sure. I'll try it next week Monday or Tuesday. Yesterday was a public holiday in Germany and today I'm out of office.

mattcaswell:
@wolftobias  any update on this?

mattcaswell:
Assuming this is resolved. Closing.

PR #11302, opened 2020-03-10
Refactor sm2_id

### LINKED ISSUE #11293: The special treatment of sm2_id shouldn't be necessary
Because of #11242, I had a closer look at how the SM2 ID is being set, and why.  It's a bit of a hack, really, and it *shouldn't* be necessary to give it this special treatment...  it should be something settable with `EVP_PKEY_CTX_ctrl()` or `EVP_PKEY_CTX_ctrl_str()` alongside any other key operation control, i.e. *after* `EVP_DigestSignInit()` and other similar functions.

By consequence, the apps *shouldn't* need separate `-sm2-id` options, they should able to specify the SM2 ID, with options such as `-sigopt sm2id:ABC` or `-sigopt hexsm2id:414243`.

I suspect that the major driving force behind this special treatment is this:

https://github.com/openssl/openssl/blob/99a16e0459e5089c2cfb92ee775f1221a51b8d05/crypto/evp/m_sigver.c#L249-L254

In other words, the call of `digest_custom` in `EVP_DigestSignInit()` forces everyone to have special knowledge about SM2 IDs (which makes it hard to have it as a configurable option among others).

This call could be moved to the corresponding Update (and Final in case Update wasn't called first) functions, and be subject of a flag set in the `EVP_PKEY_CTX` so it's only called once.

The above isn't very tricky at all.  `openssl verify` and `X509_verify()` are, however, because they don't touch keys, at least directly, so calling `EVP_PKEY_CTX_ctrl()` isn't the first thing that comes to mind.  An idea would be to have a similar mechanism to set optional data, something like `X509_STORE_CTX_ctrl()` and `X509_STORE_CTX_ctrl_str()`, and then have a `X509_verify_ctx()` that takes an extra `X509_STORE_CTX *` argument that it can pilfer data from.

Do note that with provider side implementations, there is *no* early processing support à la `digest_custom`, and I would rather see that it doesn't get added either.  This will force "early processing" to happen as part of the update (or final if no update was called) calls.

levitte:
The idea is to stop special-casing the SM2 ID.  In the center core, it means that `digest_custom()` is called slightly later, which allows *all* controls to be performed after EVP_DigestSignInit(), instead of having to call the SM2 ID setting control before and the rest after.

This allows us to remove the special `-sm2-id` and `-sm2-hex-id` flags in a number of apps.

For more information, see the commit comments.

Fixes #11293


levitte:
(some famously heard me say "I have a branch" during a vid-call this morning... this is what I was talking about, with some cleanup)

apps/lib/app_x509.c
```diff
@@ -0,0 +1,117 @@
...
+        return -1;
+    vtmp = strchr(stmp, ':');
+    if (vtmp != NULL) {
+        *vtmp = 0;
+        vtmp++;
+        vtmp_len = strlen(vtmp);
+    }
+
+    if (strcmp(stmp, "distid") == 0
+               /* GmSSL compatibility */
```
InfoHunter: For this comment and the following one, GmSSL is another project's name (https://github.com/guanzhi/GmSSL), and I don't think option `sm2_id` is compatible with GmSSL so this comment is confusing.

InfoHunter: Could be changed to some like `/* OpenSSL 1.1.1 compatibility or so */`

apps/x509.c
```diff
@@ -576,7 +578,7 @@ int x509_main(int argc, char **argv)
             BIO_printf(bio_err, "error unpacking public key\n");
             goto end;
         }
-        i = X509_REQ_verify(req, pkey);
+        i = do_X509_REQ_verify(req, pkey, sigopts);
```
InfoHunter: This is not right from my understandings. Users can only specify one `distid` in the sigopts, but the ID used when signing the CSR may not be the same as the one that is going to use when signing the real certificate from the CSR. Using one sigopts will lose the ability of this.

InfoHunter: Actually as the commands like `x509`, which deals with both verification and signing (another example is the `ca` command), should provide two separate `opts`: one sigopts and one vfyopts

apps/lib/app_x509.c
InfoHunter: Hmmm, I noticed that these sm2 options were added just after the 1.1.1 release, so it seems we don't need to consider backward compatibility at all

crypto/evp/m_sigver.c
```diff
@@ -301,6 +302,12 @@ int EVP_DigestSignUpdate(EVP_MD_CTX *ctx, const void *data, size_t dsize)
                                                       data, dsize);
 
  legacy:
+    /* do_sigver_init() checked that |digest_custom| is non-NULL */
+    if (pctx->flag_call_digest_custom
+        && !ctx->pctx->pmeth->digest_custom(ctx->pctx, ctx))
```
InfoHunter: Actually the reason why we put this part in the init phase back in 2018 was to avoid multi-part code copy - but if that is not possible anymore, then this seems fine as well.

InfoHunter: Well, one of the reasons...

apps/lib/app_x509.c
levitte: Nope. These string controls do not exist at all in 1.1.1. I checked. But you're right, gmssl don't use that, they use "signer_id" (which isn't a bad idea). 

Sooo, is "sm2_id" actually used by anyone other than fans of our master? If not, I'm considering simply removing it.

crypto/evp/m_sigver.c
levitte: I think this is justifiable by making it more manageable for the user. It's not like we're copying *large* chunks of code, this is, after all, just two lines (plus a comment)

apps/x509.c
levitte: Sure, I can add that

apps/lib/app_x509.c
InfoHunter: > Sooo, is "sm2_id" actually used by anyone other than fans of our master? If not, I'm considering simply removing it.

I think it's removable

levitte: Done

apps/x509.c
levitte: Done

levitte:
Rebased and hopefully last fight with picky compilers

apps/lib/app_x509.c
```diff
@@ -0,0 +1,120 @@
...
+                   "error: distinguishing ID allocation failed\n");
+    } else if (!ASN1_OCTET_STRING_set(v, value, value_len)) {
+        BIO_printf(bio_err,
+                   "error: setting distinguishing ID failed\n");
+        ASN1_OCTET_STRING_free(v);
+        v = NULL;
+    }
+    return v;
+}
+
```
beldmit: I'm not sure I like this diagnostic output. The function seems to be generic but the output is for a particular case.

InfoHunter:
> Rebased and hopefully last fight with picky compilers

Unfortunately it seems the red cross is relevant

levitte:
CI kicking close/re-open

levitte:
@InfoHunter, it seems that Travis didn't pick up on my last update

InfoHunter:
Ahh, so it is.

levitte:
Travis still out of sync, so I did some squashing and  pushed that.  Maybe that'll kick it hard enough...

InfoHunter:
Hmm, it remains...

InfoHunter:
Besides, considering the SM2 implementation changed recently , I think the man7/SM2.pod needs to get an update as well. Would you prefer to do that in this PR or I open a new PR fixing that, @levitte ?

levitte:
@beldmit made a good point, o I'm refactoring `apps/lib/app_x509.c`

As for `doc/man7/SM2.pod`, I think that a separate PR is due.  @InfoHunter, please feel free to work with it.

levitte:
... some days, it's hard to stand over zealous compilers.
[sigh]
another round...

levitte:
CIs are happy, the Travis failure is the usual timeout.

levitte:
@InfoHunter, thanks for the review!

Merge coming up

InfoHunter:
> Since PR #11302 has been merged so I don't think this is a problem anymore

I had a re-thought on this. It seems current design doesn't allow passing the dist-ID into a provider, so the user can only set the dist-ID in advance and then the calculation of the dist-ID is computed in the `update` process inside a provider. Is this what you originally intended? @levitte

InfoHunter:
Hmmm. The above comment was intended to post in PR https://github.com/openssl/openssl/pull/1124, and I posted to the wrong PR but nevertheless we can still discuss here...

levitte:
You mean #11248.  I would prefer we have the discussion there, as that's an open PR

PR #11310, opened 2020-03-11
Added signature length check for RSA PSS signature verification.

avgrygoriev:
Add length check for RSA PKCS1 Decryption and Signature Verification

According to RFC8017
https://tools.ietf.org/html/rfc8017

The first step of signature verification for RSASSA-PKCS1-v1_5 and RSASSA-PSS
1.  Length checking: If the length of the signature S is not k
            octets, output "invalid signature" and stop.

The first step of Decryption
RSAES-OAEP
1.  Length checking:
    b.  If the length of the ciphertext C is not k octets, output
        "decryption error" and stop.
RSAES-PKCS1-v1_5
1.  Length checking: If the length of the ciphertext C is not k
    octets (or if k < 11), output "decryption error" and stop.

This patch
- adds length check in rsa_ossl_public_decrypt
- adds length check in rsa_ossl_private_decrypt
- removes obselete SSL_R_WRONG_SIGNATURE_LENGTH
- removes obselete RSA_R_DATA_GREATER_THAN_MOD_LEN
- removes duplicated checks
- added test-case that verifies X509 cert with signature with trucated leading 0's

##### Checklist


t8m:
I do not think the CLA: trivial is appropriate for this kind of change. Also the Travis failure is relevant.

avgrygoriev:
I have submitted ICLA. Haven't received any confirmation.
Updated PR with a test-case.

mattcaswell:
> I have submitted ICLA. Haven't received any confirmation.

I processed it about 2 minutes ago :-)

Everything is good from a CLA perspective now.

crypto/rsa/rsa_pmeth.c
```diff
@@ -267,6 +267,10 @@ static int pkey_rsa_verify(EVP_PKEY_CTX *ctx,
                 return 0;
         } else if (rctx->pad_mode == RSA_PKCS1_PSS_PADDING) {
             int ret;
+            if (siglen != (size_t)RSA_size(rsa)) {
+                RSAerr(RSA_F_PKEY_RSA_VERIFY, RSA_R_WRONG_SIGNATURE_LENGTH);
+                return -1;
+            }
```
mattcaswell: This check is kind of already done by the `RSA_public_decrypt` call below:

````
    /*
     * This check was for equality but PGP does evil things and chops off the
     * top '0' bytes
     */
    if (flen > num) {
        RSAerr(RSA_F_RSA_OSSL_PUBLIC_DECRYPT, RSA_R_DATA_GREATER_THAN_MOD_LEN);
        goto err;
    }
````
So, it fails if the data is too long - but deliberately tolerates data that is too short and assumes it to have had leading 0's removed.

Comment (davidben) [reply crypto/rsa/rsa_pmeth.c]: Both PKCS#1 v1.5 and PSS have the same requirement in the specification, so `RSA_public_decrypt` also makes more sense as where the check should live. OAEP and PKCS#1 v1.5 decryption similarly require strict length checking, while `RSA_private_decrypt` has a similar weak check and comment.

FWIW, BoringSSL switched these to be spec-compliant back in 2014 and I don't think I've ever seen a report of issues.

Comment (avgrygoriev) [reply crypto/rsa/rsa_pmeth.c]: AFAIU, PKCS# v1.5 signature length is checked here
int_rsa_verify:wrong signature length:crypto/rsa/rsa_sign.c:297
```
    if (siglen != (size_t)RSA_size(rsa)) {
        RSAerr(RSA_F_INT_RSA_VERIFY, RSA_R_WRONG_SIGNATURE_LENGTH);
        return 0;
    }
    //...
    len = RSA_public_decrypt((int)siglen, sigbuf, d
```

That's why I decided to add similar check for PSS above ```RSA_public_decrypt```

Do you insist to move this check into ```RSA_public_decrypt``` (meaning ```rsa_ossl_public_decrypt```) ?

I have also viewed BoringSSL. It has this check in PSS verify function, not in generic decrypt function.

avgrygoriev:
Is that error "The job exceeded the maximum time limit for jobs, and has been terminated." something typical/random?
https://travis-ci.org/github/openssl/openssl/jobs/661647382?utm_medium=notification&utm_source=github_status

I was running same script on my machine, all tests passed.
Please, advice.

richsalz:
Yes it's not unexpected.

crypto/rsa/rsa_pmeth.c
davidben: > AFAIU, PKCS# v1.5 signature length is checked here
> int_rsa_verify:wrong signature length:crypto/rsa/rsa_sign.c:297

Ah, interesting. Hopefully it's then safe to tighten `RSA_public_decrypt` and perform the check in a more common place, since that means OpenSSL already rejects truncated PKCS#1 v1.5 signatures.

> I have also viewed BoringSSL. It has this check in PSS verify function, not in generic decrypt function.

It's right here:
https://boringssl.googlesource.com/boringssl/+/959c76d928c4df2827635ba68faac160edce10d8/crypto/fipsmodule/rsa/rsa_impl.c#594

mattcaswell:
Timeouts are an irritatingly frequent problem:

https://github.com/openssl/openssl/pull/11319

crypto/rsa/rsa_pmeth.c
mattcaswell: Yes - given that we already check this in PKCS1 v1.5, I think I favour just changing the check that is already there in RSA_public_decrypt which I noted above to be for strict equality and remove that comment. That comment has been there since before OpenSSL was born (i.e. from SSLeay) and is hence over 20 years old. I guess issues it was working around at the time have long since ceased to be a problem.

avgrygoriev: > > I have also viewed BoringSSL. It has this check in PSS verify function, not in generic decrypt function.
> 
> It's right here:
> https://boringssl.googlesource.com/boringssl/+/959c76d928c4df2827635ba68faac160edce10d8/crypto/fipsmodule/rsa/rsa_impl.c#594
> 
Oh! Didn't look here. Thank you very much.

Oh. I was talking about that code.
https://boringssl.googlesource.com/boringssl/+/959c76d928c4df2827635ba68faac160edce10d8/crypto/fipsmodule/rsa/rsa.c#645

I have updated patch and updated check to ```rsa_ossl_public_decrypt```

According to documentation ```RSA_private_decrypt```
> but may be smaller, when leading zero bytes are in the ciphertext
therefore, I didn't change it.

davidben: (Ah, yeah, that check is something else. That's about the *output* of the RSA operation and should never fail. That check is because `RSA_verify_PKCS1_PSS_mgf1` is terrible and uses an implicit length. For safety's sake, I like transitions from explicit to implicit length to be checked so it's locally obvious it's not a memory error.)

I would advocate fixing `RSA_private_decrypt` for the same reason (the spec mandates it) and fixing the documentation, but I'll defer to the OpenSSL folks whether they want to do that. BoringSSL's `RSA_private_decrypt` has also been spec-compliant here since 2014 without issues.

avgrygoriev: I have updated patch with added length check for ```rsa_ossl_private_decrypt```
Updated documentation and removed unused ```RSA_R_DATA_GREATER_THAN_MOD_LEN```

avgrygoriev:
Looking into travis failures.
So far could not reproduce locally.

mattcaswell:
OMC: Referring to OTC

t8m:
**OTC: The low level change in RSA code is OK for master, there should be a CHANGES entry**

(The high level checks removal is subject to normal review process.)

t8m:
Could you please rebase?

PR #11311, opened 2020-03-11
Refactor CPUID code

### LINKED ISSUE #11281: Build OpenSSL on ubuntu, compile failed in x86_32(providers/legacy.so)
Get the latest OpenSSL code of master branch, "./config; make; make install_sw;". compile it on ubuntu. find that the builds on x86-64,aarch32,aarch64 are success but fails on x86-32. The error log as below:
```
rm -f test/dsa_no_digest_size_test
${LDCMD:-gcc} -pthread -m32 -Wa,--noexecstack -Wall -O3 -fomit-frame-pointer -Werror   \
	-o test/dsa_no_digest_size_test \
	test/dsa_no_digest_size_test-bin-dsa_no_digest_size_test.o \
	test/libtestutil.a libcrypto.a -ldl -pthread 
rm -f test/dsatest
${LDCMD:-gcc} -pthread -m32 -Wa,--noexecstack -Wall -O3 -fomit-frame-pointer -Werror   \
	-o test/dsatest \
	test/dsatest-bin-dsatest.o \
	test/libtestutil.a libcrypto.a -ldl -pthread 
collect2: error: ld returned 1 exit status
Makefile:21332: recipe for target 'providers/legacy.so' failed
make[1]: *** [providers/legacy.so] Error 1
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory '/home/ent-user/ci-scripts/openssl-src'
Makefile:2899: recipe for target 'build_sw' failed
make: *** [build_sw] Error 2

```

levitte:
We were using CPUID coded in several modules, but it was unclear how
it actually got there, and could fail randomly.

To remedy that, this change separates the CPUID C code from the rest
of cryptlib.c, and ensures the right modules get both that and the
assembler sources explicitly.

Fixes #11281


paulidale:
I do wonder if propagating the CPUID is the way to go, would having a series of utility functions which query specific features be more readable?  It won't change what needs to be done here but it would avoid magic constants throughout the code.

levitte:
Do we want to make the CPUID functions public?  Considering that legacy.so links with libcrypto.so, that's what it'll take.  Or did you mean utility functions like replacing `(OPENSSL_ia32cap_P[2] & (1 << 29)) == 0` with `OPENSSL_ia32cap_has_SHAEXT()`?  I would straight up implement those as macros.

Anyway, that sounds like an effort for another PR

paulidale:
The latter.  Agreed, it is an enhancement.

mattcaswell:
Travis red cross is relevant

levitte:
I figured out one of the failures...  it's possible that the other one is fixed as well

t8m:
I suppose the travis failure is still relevant, also needs rebase.

levitte:
Rebased

slontis:
@levitte - travis is running now - it has some issues in liblegacy.

t8m:
Hmm, looking at the ppccap.c it seems to me it would make sense to rather refactor the algorithm-specific code out than to add everything to the legacy provider.

t8m:
Another option - more ugly but less invasive - #ifdef all the code not needed in the legacy provider so when the .c is built for the legacy provider this code is not compiled.

bernd-edlinger (dismissed):

t8m:
This is rather WIP - there are conflicts and issues in Travis jobs in legacy provider.

levitte:
> This is rather WIP - there are conflicts and issues in Travis jobs in legacy provider.

Yeah...  this has turned out to contain a building problem that I've been avoiding for some time.  I'll have to bite that bullet...

levitte:
I changed this back to milestone 3.0.0.  This is essentially to clear a building failure that appears under some conditions and some platforms.  I don't see that we must rush this before beta1

levitte:
Rebased on a fresher master.  Let's see what can be done with this

levitte:
I've done several tests lately, that were failing previously, but that now build and run correctly, given a fresher master.
With that in mind, I'm wondering if this fix is still necessary.  If someone still gets clashes when linking, *and* can give me a way to reproduce it *on Linux* (I run Debian Sid), then this may be worth another look.  If not, I'm considering closing this, to possibly pick up on another rainy day.

t8m:
I remember the problems were mostly on non x86 architectures. PPC was the most problematic IMO.

xkqian:
> I remember the problems were mostly on non x86 architectures. PPC was the most problematic IMO.

Seems it happened on x86 32bit platform.  I remember that Richard had one fix before, I verified it locally and the issue had gone.   But that fix seems cause Travis red cross. I will verify it locally to see whether the issue has gone now.

levitte:
> Seems it happened on x86 32bit platform.

Yes, and I could reproduce that...  then.

> I will verify it locally to see whether the issue has gone now.

Please do.

xkqian:
> > Seems it happened on x86 32bit platform.
> 
> Yes, and I could reproduce that... then.
> 
> > I will verify it locally to see whether the issue has gone now.
> 
> Please do.

Done, the issue has gone with the openssl master branch code, as least the commit: daa86f9

.travis.yml
```diff
@@ -59,6 +59,9 @@ jobs:
           arch: s390x
           compiler: gcc
           env: CONFIG_OPTS="--strict-warnings"
+        - os: linux-ppc64le
```
paulidale, line 62: I'm wondering why?

paulidale (approved):
One question but approved nonetheless.

paulidale:
I also removed the _[WIP]_ from the subject here.

levitte:
> I also removed the _[WIP]_ from the subject here.

... because?

.travis.yml
levitte: It's quite possible that this platform demonstrated the issue at the time.
Since we don't do Travis any more, that's a bit hard to test for the moment...  I'm not sure if it's even possible to specify specific hardware with GitHub Actions

levitte:
> Done, the issue has gone with the openssl master branch code, as least the commit: daa86f9

After a conversation with the rest of the team, it's been decided to close this for now.  If the issue turns out to still be present, we can pick it up then.

levitte:
Re-opening this, it's time to get it done, now that there's a workflow to test against (#14753)

PR #11312, opened 2020-03-11
Free unused ex data nodes

davxy:
Instead of using dummy callbacks and leave unused stuff around,  let the `CRYPTO_free_ex_index` free the malloced `EX_CALLBACK` structure.

---

Removed superfluous two steps EX_CALLBACK structure push in the EX_CALLBACKS stack.

---

CLA trivial

### DIFF
--- a/crypto/ex_data.c
+++ b/crypto/ex_data.c
@@ -152,13 +152,12 @@ int crypto_get_ex_new_index_ex(OPENSSL_CTX *ctx, int class_index, long argl,
     a->dup_func = dup_func;
     a->free_func = free_func;
 
-    if (!sk_EX_CALLBACK_push(ip->meth, NULL)) {
+    if (!sk_EX_CALLBACK_push(ip->meth, a)) {
         CRYPTOerr(CRYPTO_F_CRYPTO_GET_EX_NEW_INDEX_EX, ERR_R_MALLOC_FAILURE);
         OPENSSL_free(a);
         goto err;
     }
     toret = sk_EX_CALLBACK_num(ip->meth) - 1;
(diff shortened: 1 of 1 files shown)


richsalz:
I am concerned about the safety of this PR.  If one part of the program holds on to an old index while another part of the program free's it, things will go haywire here. The current behavior is a no-op.

davxy:
If another part of the program frees an index I see no issues. 

The user is using indexes via the interface functions and the internals are already checking for NULL pointers before use an ex_data node.

Correct me if I didn't get what you are suggesting.

Thanks

richsalz:
I have an exindex and I store "struct foo" in it.  A bug in my code says "I don't need that exindex any more, free it"  But I still have objects with that index in use.  Someone else (another developer, I library I am using, the application that uses the library I am developing, etc) gets a new exindex.  All objects that had the old one will not just leak their objects, but are likely to crash because "free struct foo" isn't the function being called any more.

davxy:
Understood. You are saying that the "others" belive that at a given index there is a certain "type" while someone else just stored another type there? This looks reasonable.

davxy:
I've just left the code that frees the structure. The indexes are now returned incrementally as before.

richsalz:
I don't understand why this is worth doing, but others may disagree.  Early free'ing of a small structure.  Shrug.

davxy:
@richsalz  is there ever a pull request that you find useful? I started studying the codebase from the ground up and when I find small things that deserve a small fix I feel compelled to do the fix. But every time your comments arrive on time that discourage the work done. I don't think this is the spirit on which a collaborative project in need of contributions such as OpenSSL must be founded. In any case, I will continue my path helping where I can. Thanks

PR #11314, opened 2020-03-12
dh: document what the PEM files in apps actually contain.

paulidale:
They were claimed to be the SKIP primes but they are really two of the MODP Diffie-Hellman groups for IKE.


paulidale:
Should this go back to 1.1.1 too?  The PEM files were last modified four years ago.
Should the `dh1024.pem` file be deleted and the reference in the documentation be removed?

I'm happy for both to happen.

doc/man3/SSL_CTX_set_tmp_dh_callback.pod
```diff
@@ -63,12 +63,11 @@ openssl L<openssl-dhparam(1)> application. This application
...
-These files can be converted into C code using the B<-C> option of the
-L<openssl-dhparam(1)> application. Generation of custom DH
-parameters during installation should still be preferred to stop an
-attacker from specializing on a commonly used group. File dh1024.pem
+version of the OpenSSL distribution contain two of the MODP Diffie-Hellman
+groups for IKE as per RFC 3526.  These files can be converted into C code
+using the B<-C> option of the L<openssl-dhparam(1)> application. Generation
+of custom DH parameters during installation should still be preferred to
+stop an attacker from specializing on a commonly used group. File dh1024.pem
 contains old parameters that must not be used by applications.
```
levitte: Why do we even have this paragraph in this manual?  Are we seriously telling people that they should go fetch the OpenSSL source to find some random files, *in the manual for `SSL_set_tmp_dh()` and friends???*

If we're serious about it, we should probably *install* those files, in something like $OPENSSLDIR/misc

levitte: Side note:

``` console
$ find . -name dh2048.pem
./apps/dh2048.pem
./crypto/dh/dh2048.pem
$ find . -name dh4096.pem
./apps/dh4096.pem
./crypto/dh/dh4096.pem
```

paulidale: I didn't notice the second copy.

paulidale: I agree that the manual page needs a major overhaul, my intention here was to correct a blatant factual error that took me way too long to figure out....

paulidale: The PEMs in question are not even meant to be used for TLS....

levitte: If we're not installing them, then this paragraph shouldn't be here at all

paulidale: I can do this.  I'll see if I can remove the files at the same time -- there is no point us having non-TLS DH parameters hanging around.

levitte: Uhmmm, I believe that the one in apps is used in some tests, so that would require rewriting them. I suggest doing that as a separate PR.

paulidale: I've removed all three of the .pem files and the paragraph from the documentation.

paulidale:
The dh*.pem files in crypto were also very old and unusued.  I've removed them as well.

Probably not for 1.1.1 given the file destruction.

paulidale:
Merged to master.  Thanks for the review.

PR #11315, opened 2020-03-12
Fix RSA structure

levitte:
The first field was called 'pad', but not for the reason one might
think.  It was really a padding int that was always zero, and was
placed first on purpose.  This is to pick up programming errors where
an RSA pointer was passed when an EVP_PKEY pointer should have been,
an makes it look like an EVP_PKEY structure with type EVP_PKEY_NONE,
which effectively avoids any further processing (and unintended
corruption of the RSA structure).

This is only relevant for legacy structure and EVP_PKEY_METHODs.  With
providers, EVP_PKEYs aren't passed to the backend anyway.


levitte:
I introduced a hashtag, `#legacy`, in the comment...  I propose that we use this consistently, as it makes it easy to grep for it later on.

Issue #11182, opened 2020-02-26
Current master fails two of the extended tests due to MD5-SHA1 signatures


t8m:
The current master fails two of the extended tests due to MD5-SHA1 digest.

One job fails with:
```

go test: FAILED (CertificateVerificationSucceed-Server-SSL3-Async-PackHandshakeFlight)
go test: unexpected failure: local error 'EOF', child error 'exit status 1', stdout:
go test: 
go test: stderr:
go test: Connection 1 failed.
go test: 40:17:77:7F:50:7F:00:00:error:digital envelope routines:EVP_MD_CTX_ctrl:message digest is null:crypto/evp/digest.c:661:
go test: 40:17:77:7F:50:7F:00:00:error:SSL routines::EVP lib:ssl/statem/statem_lib.c:509:

```
The other with:
```
ERROR in CLIENT
00:07:98:69:51:7F:00:00:error:digital envelope routines:EVP_MD_CTX_ctrl:message digest is null:crypto/evp/digest.c:661:
00:07:98:69:51:7F:00:00:error:SSL routines::EVP lib:ssl/statem/statem_lib.c:306:
client authentication
Doing handshakes=1 bytes=256
SSLv3, cipher (NONE) (NONE), 2048 bits RSA, temp key: 1024 bits DH, digest=MD5-SHA1
../../util/shlib_wrap.sh ../../test/ssltest_old -s_key keyU.ss -s_cert certU.ss -c_key keyU.ss -c_cert certU.ss -bio_pair -ssl3 -client_auth -CAfile certCA.ss => 1
    not ok 3 - test sslv3 with client authentication via BIO pair
    #   Failed test 'test sslv3 with client authentication via BIO pair'
    #   at test/recipes/80-test_ssl_old.t line 343.
ERROR in CLIENT
00:C7:58:AB:DE:7F:00:00:error:digital envelope routines:EVP_MD_CTX_ctrl:message digest is null:crypto/evp/digest.c:661:
00:C7:58:AB:DE:7F:00:00:error:SSL routines::EVP lib:ssl/statem/statem_lib.c:306:
client authentication
server authentication
Doing handshakes=1 bytes=256
depth=1 /C=AU/O=Dodgy Brothers/CN=Dodgy CA
depth=0 /C=AU/O=Dodgy Brothers/CN=Brother 1/CN=Brother 2
[...]

PR #11317, opened 2020-03-12
Handle the case where there is no digest in an EVP_MD_CTX

mattcaswell:
EVP_MD_CTX_ctrl assumes that a digest has always been set. However in a
non-legacy EVP_DigestSign* operation this is not the case because the
digest is handled entirely by the underlying signature implementation.

This fixes one of the travis failures on the master branch.

[extended tests]


mattcaswell:
As hoped the failing extended test is now passing. There is still one other failing extended test that this PR does not address.

t8m:
Just please amend the commit message with Fixes #11182 before merging.

mattcaswell:
Pushed. Thanks.

PR #11313, opened 2020-03-11
WIP: add some travis timeouts

richsalz:
Trying https://github.com/richsalz/openssl/pull/new/travis-work-arounds

PR #11319, opened 2020-03-12
Extend Travis build time-out

richsalz:
Replaces #11313 which I could not re-open because I force-pushed before reopening.  Who knew? :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,10 +11,10 @@ git:
 
 before_install:
     - if [ -n "$COVERALLS" ]; then
-          pip install --user cpp-coveralls;
+          travis_retry pip install --user cpp-coveralls;
       fi;
     - if expr "$CONFIG_OPTS" ":" ".*enable-external-tests" > /dev/null; then
-          git submodule update --init --recursive;
+          travis_retry git submodule update --init --recursive;
       fi;
     - eval "${MATRIX_EVAL}"
 
@@ -246,7 +246,7 @@ script:
           if [ -e krb5/src ]; then
               sudo apt-get -yq install bison dejagnu gettext keyutils ldap-utils libldap2-dev libkeyutils-dev python-cjson python-paste python-pyrad slapd tcl-dev tcsh;
           fi;
-          if ! HARNESS_VERBOSE_FAILURE=yes BORING_RUNNER_DIR=$top/boringssl/ssl/test/runner make test; then
+          if ! HARNESS_VERBOSE_FAILURE=yes BORING_RUNNER_DIR=$top/boringssl/ssl/test/runner travis_wait 60 make test; then
               echo -e '\052\052 FAILED -- MAKE TEST';
               travis_terminate 1;
           fi;
--- a/.travis.yml
+++ b/.travis.yml
@@ -246,7 +246,7 @@ script:
           if [ -e krb5/src ]; then
               sudo apt-get -yq install bison dejagnu gettext keyutils ldap-utils libldap2-dev libkeyutils-dev python-cjson python-paste python-pyrad slapd tcl-dev tcsh;
           fi;
-          if ! HARNESS_VERBOSE_FAILURE=yes BORING_RUNNER_DIR=$top/boringssl/ssl/test/runner travis_wait 60 make test; then
+          if ! HARNESS_VERBOSE_FAILURE=yes BORING_RUNNER_DIR=$top/boringssl/ssl/test/runner travis_wait 60 make _tests; then
               echo -e '\052\052 FAILED -- MAKE TEST';
               travis_terminate 1;
           fi;


levitte:
From reading this, it seems like `travis_wait` defaults to a 20 minute timeout, but you can specify something longer:

https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

richsalz:
Yes, 20 minutes is the default.  Let's see what happens because right now it cuts off at 10.

levitte:
... you mean at 30

It looks promising!

richsalz:
almost there.  updating to 40.

richsalz:
trying again with timeout set to (!!!) 60.

mattcaswell:
In spite of the commit message saying "Update timeout to 60", the timeout still seems to be 40! Although this time it passed.

slontis:
Might try this again just for fun...

richsalz:
forgot to add the file to the commit.  timeout now set at 60 and re-pushed.

richsalz:
I was going to try making the timeout be a shell variable, and then we could leave it at the default for all but the one problematic platform, but I don't think it's necessary. If every build takes its maximum of an hour, we'll notice the backlog I think.  So taking this out of WIP. Should be backported, too.

slontis:
Well that didn't quite work out as planned - 2 travis timeouts. (It still seemed to timeout after 50 minutes?)

beldmit:
I'm not sure that just increasing timeout is a decent way to deal with this problem...

levitte:
Hmmm, the timeouts we see seems to be largely due to `make test` causing a rebuild.  This is at least the case for [job #33142.4](https://travis-ci.org/github/openssl/openssl/jobs/662429085), and I *suspect* (it's not as clear) that the same goes for [job #33142.5](https://travis-ci.org/github/openssl/openssl/jobs/662429086).

This may mean that some dependencies get a later timestamp than expected, causing another recompile.  This is actually an old problem that isn't unique to us.  I recall that some GNU documentation recommend `make && make` for exactly this sort of reason when you use the automatic Makefile generation available in `gcc` and derivatives.

mspncp:
I encountered this behaviour locally too several times. At some time I suspected that it has to do with the conversion of som `*.pod` files to `*.pod.in` files and that it occurs only when I build in parallel (`-j 8`). But I never got around to invistigate that systematically.

levitte:
When you fiddle around with branches, it's mostly any changes in header file dependencies that may trigger an extra rebuild.

richsalz:
> Hmmm, the timeouts we see seems to be largely due to make test causing a rebuild. 

Since we did just build, then perhaps using _tests as the target makes sense.  Trying that now.

levitte:
> Since we did just build, then perhaps using _tests as the target makes sense. Trying that now.

I hope that it works out at all times and doesn't have us depend on luck...

richsalz:
I don't see what could go wrong (famous last words), but it worked.  So, ready for review.

levitte:
> I don't see what could go wrong (famous last words), but it worked. So, ready for review.

Considering the caches are probably PR specific, it's highly probably nothing goes wrong.  I've seen things going wrong, i.e. needing a second recompile, but that's been when I've jumped wildly between branches, so..

levitte:
I supect that things are different enough between 1.1.1 and master that 1.1.1 might need a separate PR

richsalz:
1.1.1 PR in #11331

richsalz:
Since this PR seems to address the near-constant CI timeouts, it seems helpful to get this merged sooner rather than waiting?

levitte:
Yup.
Second reviewer, please, @openssl

levitte:
@t8m, I propose "urgent".  Do you agree?

mspncp:
I squashed the two commits and slightly edited the joined commit message, see d18dae018278ed4883b6b2f795e315e9d89c7cd9.  @richsalz @levitte @t8m do I have your approval to push it like this?

```
Fix travis timeouts

- Add travis_wait to the build command
- And travis_retry to some apt-get commands.
- Use `make _tests` instead of `make test`

Reviewed-by: Matthias St. Pierre <Matthias.St.Pierre@ncp-e.com>
(Merged from https://github.com/openssl/openssl/pull/11319)
```

mspncp:
Oh, sorry, I should have used your pull request title as subject line. Hang on for a second...

```
Extend Travis build time-out

- Add travis_wait to the build command
- And travis_retry to some apt-get commands.
- Use `make _tests` instead of `make test`

Reviewed-by: Matthias St. Pierre <Matthias.St.Pierre@ncp-e.com>
(Merged from https://github.com/openssl/openssl/pull/11319)
```

mspncp:
Done, see fd32f91344da8d08d0ea91a3effccf0307b96d24.

richsalz:
fine with me.

mspncp:
Ok, then I'll push it now.

mspncp:
Merged as cde63b731575e6be41b3f94bcd9fa941cdc45488. Henceforth, thou shalt be called the Mighty Travis Timeout Slayer, @richsalz!

richsalz:
now you've jinxed it :)

mspncp:
> now you've jinxed it :)

Are you saying I should have kept my mouth shut until [Travis has completed](https://travis-ci.org/github/openssl/openssl/builds/663720437?utm_source=github_status&utm_medium=notification)? I hope, I didn't mess it up...

PR #11323, opened 2020-03-13
Add Android testing to Travis

noloader:
This PR adds Android testing for Travis.

There's not much to Android for Travis. First the Android SDK and NDK are installed for Android builds.

Second, `setenv-android.sh` is called to set `CC`, `CXX`, `AR`, `AS`, `RANLIB`, `CPPFLAGS`, `CFLAGS`, `CXXFLAGS`. `setenv-android.sh` also sets the `$PATH` so the toolchain is on path. `CFLAGS` and `CXXFLAGS` were taken from an Android sample project provided by the Android folks. They are the same flags used by `ndk-build`.

Third, the four Android platforms are configured, built and installed - `android-arm`, `android-aarch64`, `android-x64` and `android-x86_64`. The installation directory is `ANDROID_PREFIX="$HOME/android$ANDROID_API-$ANDROID_CPU"`.

`15-android.conf` was gutted and replaced with a streamlined version. Because `setenv-android.sh` sets the entire cross-compile environment the extra cruft could be removed.

`.travis.yml` was also gutted. Travis only needs to perform the Android builds so everything else was removed. Travis does perform two native Linux/amd64 builds to ensure things are working as expected.

I think there are two outstanding issues. First, `android-arm` has ASM disabled. Be careful if your enable it with, say, `perlasm_scheme => "linux32"`. The code will compile, but it segfaults on ARM devices in general (like a Beagleboard or a CubieTruck). It appears there is (or was) some bad code generation with Clang. Also see [Clang miscompiles ARMv7-a assembly module](https://bugs.llvm.org/show_bug.cgi?id=38133) in the LLVM bug tracker (it uses the Cryptogams code to demonstrate the problem).

The second issue is, `android-x86_64` has ASM disabled. Despite the choice of `bn_ops`, Clang uses 32-bit registers for `crypto/bn/asm/x86_64-gcc.c`. `x86_64-gcc.c` uses 64-bit inline asm, so it results in a compile error (like `addq %eax, %esi`).

`setenv-android.sh` is rooted in an old script Tim Hudson provided about 10 or 12 years ago. It was usurped and modified throughout the years. Now it has come full circle and made its way back to the project.

I looked at integrating into the existing `.travis.yml`, but the file is too busy. I did not feel comfortable changing it because the Android tests meant moving away from matrix expansion.

If OpenSSL does not want to add this PR to Master, then please consider keeping it in a separate branch and setting up a Travis CRON job to build the branch once a week or so. Also see [GitHub Help: Checking out pull requests locally](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally).

##### Checklist

### DIFF
--- a/doc/man7/Ed25519.pod
+++ b/doc/man7/Ed25519.pod
@@ -11,7 +11,7 @@ Ed448
 The B<Ed25519> and B<Ed448> EVP_PKEY implementation supports key generation,
 one-shot digest sign and digest verify using PureEdDSA and B<Ed25519> or B<Ed448>
 (see RFC8032). It has associated private and public key formats compatible with
-draft-ietf-curdle-pkix-04.
+RFC 8410.
 
 No additional parameters can be set during key generation, one-shot signing or
 verification. In particular, because PureEdDSA is used, a digest must B<NOT> be
--- a/doc/man7/X25519.pod
+++ b/doc/man7/X25519.pod
@@ -10,7 +10,7 @@ X448
 
 The B<X25519> and B<X448> EVP_PKEY implementation supports key generation and
 key derivation using B<X25519> and B<X448>. It has associated private and public
-key formats compatible with draft-ietf-curdle-pkix-03.
+key formats compatible with RFC 8410.
 
 No additional parameters can be set during key generation.
 


noloader:
I'm not sure why `demos/pkcs12/pkread.c` and `demos/pkcs12/pkwrite.c` are in this PR. I did not touch them, and I rebased off master before the PR.

More Git goodness...

mattcaswell:
> If OpenSSL does not want to add this PR to Master

I think the right answer is to get this PR up to a state where we can merge it into master. This will mean we see android issues at the same time as other issues and the dev team can see that it needs fixing if we break it. Having it in a separate branch is less than ideal - the branch will inevitably fall behind or break with merge conflicts and lose its usefulness - and will get less attention from the dev team.

In order for that to happen we can't just replace all the existing travis jobs with android ones - so we would need to integrate the two sets of jobs together into one travis.yml file.

I also wonder whether the work that setenv_android.sh is doing might be better done by the Configure script. Ideally - and as much as is possible - we don't want to have different build steps for different platforms. Perhaps @levitte might have some thoughts on this.

noloader:
Thanks @mattcaswell,

> In order for that to happen we can't just replace all the existing travis jobs with android ones

Lol... I know. I side-stepped it once I got a good look at the existing `.travis.yml`.

I started to update `.travis.yml` on a separate branch. Let me get back on it and see how things go.. Once the `.travis.yml` updates are done, integrating Android will be as easy as copy/paste from this PR.

> I also wonder whether the work that `setenv_android.sh` is doing might be better done by the Configure script...

OK, thanks. That may be a difference in philosophy. In my mind's eye, it is the user's responsibility to set the tools for configure and make. Once the tools are set, it is configure's job to build a working configuration to be made (like adding `armv4_aes.s`).

noloader:
Sorry Matt. I tried to sync with upstream and rebase but Git blew away my changes. That is one fucked up, miserable tool. It cannot perform the simplest of workflows developers need.

This PR is now dead.

mattcaswell:
> Sorry Matt. I tried to sync with upstream and rebase but Git blew away my changes. That is one fucked up, miserable tool. It cannot perform the simplest of workflows developers need.
>
> This PR is now dead.

Ah. That is a great shame.

richsalz:
The commit(s) might still be in your local repo.  Perhaps @mspncp can offer some git wizard tips to find them, and then "git show {commitid}" would work.

InfoHunter:
> The commit(s) might still be in your local repo. Perhaps @mspncp can offer some git wizard tips to find them, and then "git show {commitid}" would work.

I guess the magic could be related to some reflog stuffs...

mspncp:
@noloader it seems like you deleted your openssl repository clone? Normally, this would be no problem, because the openssl/openssl repository has references pointing to the heads of all pull requests, so you would be able to recover your branch using the following commands:

```
~/src/openssl$ git fetch https://github.com/openssl/openssl.git  refs/pull/11323/head
From https://github.com/openssl/openssl
 * branch                  refs/pull/11323/head -> FETCH_HEAD
 
~/src/openssl$ git checkout -b pr-11323-reloaded FETCH_HEAD
Switched to a new branch 'pr-11323-reloaded'

~/src/openssl$ git show --oneline --no-patch HEAD
c08dea30d4 (HEAD -> pr-11323-reloaded) doc: Update the reference from draft to RFC
```

Unfortunately, force-pushed to commit c08dea30d4d127412097b39d9974ba6090041a7c, which is a commit on the master branch. That's the reason why GitHub does not show any diffs anymore.

I looked in all my local working copies, but did not find an older version of your branch. But if you still have the local working copy which contained your branch, you might be able to recover the lost branch yourself using the git reflog. You find some instructions how to do it in [Section 10.7: Git Internals - Maintenance and Data Recovery](https://git-scm.com/book/en/v2/Git-Internals-Maintenance-and-Data-Recovery) of the Pro Git book  (scroll down to 'Data Recovery').

HTH

PR #11327, opened 2020-03-13
s_client: fix -starttls irc

ailin-nemui:
tested on chat.freenode.net:6667, irc.hackint.org:6667, and irc.afternet.org:6667

##### Checklist

### DIFF
--- a/apps/s_client.c
+++ b/apps/s_client.c
@@ -2408,16 +2408,14 @@ int s_client_main(int argc, char **argv)
         break;
     case PROTO_IRC:
         {
-            int numeric, pos, foundit = 0;
+            int numeric, pos;
             char *buf;
             char command[11] = { 0 };
             BIO *fbio = BIO_new(BIO_f_buffer());
 
-            BIO_printf(bio_c_out, "[sending plaintext] CAP LS 302\\r\\n"
-                       "STARTTLS\\r\\n\n");
+            BIO_printf(bio_c_out, "[sending plaintext] STARTTLS\n");
             BIO_push(fbio, sbio);
-            BIO_printf(fbio, "CAP LS 302\r\n"
-                       "STARTTLS\r\n");
+            BIO_printf(fbio, "STARTTLS\r\n");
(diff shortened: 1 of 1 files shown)


ailin-nemui:
I found that s_client -starttls cannot connect to IRC, it fails on the plaintext messages that precede the TLS connection. Is this change trivial enough?

mattcaswell:
> I found that s_client -starttls cannot connect to IRC, it fails on the plaintext messages that precede the TLS connection. Is this change trivial enough?

No, this change will require a CLA. Please see:

ailin-nemui:
I submitted icla

mattcaswell:
Close/reopen to kick CLA bot

apps/s_client.c
```diff
@@ -2408,11 +2408,16 @@ int s_client_main(int argc, char **argv)
...
+            char *buf;
+            char command[11] = { 0 };
             BIO *fbio = BIO_new(BIO_f_buffer());
 
+            BIO_printf(bio_c_out, "[sending plaintext] CAP LS 302\\r\\n"
+                       "STARTTLS\\r\\n\n");
             BIO_push(fbio, sbio);
-            BIO_printf(fbio, "STARTTLS\r\n");
+            BIO_printf(fbio, "CAP LS 302\r\n"
+                       "STARTTLS\r\n");
```
kroeckx: Can you explain why this is needed? There should be no need to list the IRC server capabilities first, it should always reply to STARTTLS if if supports it, see: https://ircv3.net/specs/extensions/tls-3.1.html

kroeckx:
So I can confirm that s_client doesn't work properly with those servers. But I can also confirm that if I just send STARTTLS as first command that they all do work properly. It just seems that s_client does not properly wait for the reply.

apps/s_client.c
ailin-nemui: I removed it since it does not seem to be strictly required

t8m:

t8m:

PR #11328, opened 2020-03-13
EC  param / key generation

levitte:
Both implemented at the same time, because they are so very closely related.

To get the same kind of functionality as with legacy keys, where we can basically switch back and forth between the EC and SM2 keytypes, this adds a `OP_keymgmt_gen_get_params`, which can really be used to get all kinds of information from the key generation, but *for the moment*, it's used to get the resulting keytype after generation.

UPDATE: SM2 key generation removed, and there's a hack in `crypto/evp/pmeth_gn.c` to handle them by downgrading to legacy keys when appropriate


levitte:
THIS IS A VERY EARLY DRAFT!

It is incomplete and doesn't even build right.  I just wanted to expose the idea of being able to switch the keytype of the EC key structure.

crypto/evp/pmeth_gn.c
```diff
@@ -181,6 +188,24 @@ int EVP_PKEY_gen(EVP_PKEY_CTX *ctx, EVP_PKEY **ppkey)
...
+     * SM2 curve resulting in a keytype "SM2" key by default.
+     */
+    {
+        const char *keytype = ctx->keytype;
+
+        ret = ret
+            && evp_keymgmt_util_gen_get_keytype(&ctx->keytype, ctx->keymgmt,
+                                                ctx->op.keymgmt.genctx)
+            && (keytype == NULL
+                || strcmp(keytype, ctx->keytype) == 0
```
slontis: do we care about case?

providers/implementations/keymgmt/ec_kmgmt.c
```diff
@@ -767,21 +777,210 @@ int ec_validate(void *keydata, int selection)
...
+    }
+    return gctx;
+}
+
+static void *sm2_gen_init(void *provctx, int selection)
+{
+    struct ec_gen_ctx *gctx = ec_gen_init(provctx, selection);
+
+    if (gctx != NULL) {
+        gctx->gen_group = EC_GROUP_new_by_curve_name(NID_sm2);
```
slontis: use the _ex variant that has a libctx

providers/implementations/keymgmt/ec_kmgmt.c
```diff
@@ -767,21 +777,210 @@ int ec_validate(void *keydata, int selection)
...
+{
+    if (group == NULL) {
+        ERR_raise(ERR_LIB_PROV, PROV_R_NO_PARAMETERS_SET);
+        return 0;
+    }
+    return EC_KEY_set_group(ec, group) > 0;
+}
+
+/*
+ * No callback defined for EC_KEY generation, so we don't use them.
```
slontis: Or you could just say :)

The callback arguments (osslcb & cbarg) are not used by EC_KEY generation

providers/implementations/keymgmt/ec_kmgmt.c
```diff
@@ -767,21 +777,210 @@ int ec_validate(void *keydata, int selection)
...
+    if ((gctx->selection & OSSL_KEYMGMT_SELECT_DOMAIN_PARAMETERS) != 0)
+        ret = ret && ec_gen_assign_group(ec, gctx->gen_group);
+    /* Whether you want it or not, you get a keypair, not just one half */
+    if ((gctx->selection & OSSL_KEYMGMT_SELECT_KEYPAIR) != 0)
+        ret = ret && EC_KEY_generate_key(ec);
+
+    if (ret)
+        return ec;
+
+    /* Something went wrong, through the key away */
```
slontis: throw

providers/implementations/keymgmt/ec_kmgmt.c
levitte: That'll help get past a build failure, at least...

providers/implementations/keymgmt/ec_kmgmt.c
levitte: ??!!!!  I know this, why do I type it wrong?  :confused:

crypto/evp/pmeth_gn.c
```diff
@@ -200,7 +200,7 @@ int EVP_PKEY_gen(EVP_PKEY_CTX *ctx, EVP_PKEY **ppkey)
             && evp_keymgmt_util_gen_get_keytype(&ctx->keytype, ctx->keymgmt,
                                                 ctx->op.keymgmt.genctx)
             && (keytype == NULL
-                || strcmp(keytype, ctx->keytype) == 0
+                || strcasecmp(keytype, ctx->keytype) == 0
```
slontis: normally have to include "e_os.h" for this..

levitte: Fixed

levitte:
Heh, it turns out that this must wait on #11248, since there are tests that generate an SM2 key in this PR, and then have signature failing because that hasn't been implemented yet.

InfoHunter:
> Heh, it turns out that this must wait on #11248, since there are tests that generate an SM2 key in this PR, and then have signature failing because that hasn't been implemented yet.

I expect #11248 will remove the `WIP` soon

romen:
I think we should discuss the approach of mixing SM2 and EC keygen.

- Why should the provider care about the legacy quirkiness of libcrypto in handling SM2 and EC as if they were the same? Wouldn't it be better if the provider implementation was simpler (at the cost of small repetition in non critical code)?
  - if framework asks for an SM2 key generation the SM2 key generation is performed
  - if framework asks for an EC key generation (even on the curve named SM2) an EC key generation is performed
  - push the "you asked for an EC key on SM2, take instead this SM2 key" logic inside libcrypto
- The current implementation of keygen for SM2 in 1.1.1 (and in this PR) is slightly incorrect: the ranges for the random scalar differ (see #8435). In 1.1.1 we got stuck because EC and SM2 share too much code about keygen, here in the provider keeping SM2 and EC keygen more separate could be better.

levitte:
I think we should discuss the approach of mixing SM2 and EC keygen.
@romen spoke thusly:
> - Why should the provider care about the legacy quirkiness of libcrypto in handling SM2 and EC as if they were the same? Wouldn't it be better if the provider implementation was simpler (at the cost of small repetition in non critical code)?
>   - if framework asks for an SM2 key generation the SM2 key generation is performed
>   - if framework asks for an EC key generation (even on the curve named SM2) an EC key generation is performed
>   - push the "you asked for an EC key on SM2, take instead this SM2 key" logic inside libcrypto

Generally speaking, it's a good question, and it basically comes down to "no surprise".  Essentially, the way the new legacy code works is by detection, i.e. change the key type based on curve name, *currently* only for SM2, but hang on, it'll be more interesting as I continue :wink:.

For the "push" that you mention, i.e. check at generation if the curve name was "SM2", and in that case, switch over to ask for an SM2 keymgmt instead of the EC keymgmt...  this involves much more, and I think that this also ties into what you've talked about elsewhere, to be able to have keytypes that contain a bit more information, for example to be able to specify, say, the keytype "EC-P239v1".
So basically, you're saying that `EVP_PKEY_gen()` should:

- detect that it's handling a key type that has possible sub-categories.  Right now, that would be detecting that the key type is "EC"
- figure out the sub-category.  Right now and if the key type is "EC", that involves getting the curve name, create a new key type composed of the key type "EC", a dash, and the curve name.  If the curve name is "P239v1", the resulting string is "EC-P239v1"

So question is, when happens with the next key type that has similar properties, but is otherwise different?  On the vid call today, you mentioned RSA (I supposed you meant things like RSASSA-PSS and RSAES-OAEP), so we'd have to hack in special knowledge about those.  And then there was something about DSA...

Do you see where I'm going with this, and how things could become quite "interesting" indeed?

> - The current implementation of keygen for SM2 in 1.1.1 (and in this PR) is slightly incorrect: the ranges for the random scalar differ (see #8435). In 1.1.1 we got stuck because EC and SM2 share too much code about keygen, here in the provider keeping SM2 and EC keygen more separate could be better.

I'll have to study that issue.  I wonder if that could even be the issue to kill `EVP_PKEY_set_alias_type()` (I thoroughly dislike that function, it's not well thought out)...

romen:
For SM2 my argument is slightly different than for ECDH/ECDSA over different curves.

- SM2 is its own cryptosystem, it has its specific keygen and its specific sign/verify algorithms.
- ECDSA is its own cryptosystem with its keygen and sign/verify algorithms.
- Both SM2 and ECDSA (and ECDH) are based on elliptic curve cryptography, so they use the `EC` module, the same way `RSA` and `DSA` use the `BN` module, but they are strictly independent from an operational point of view just as `RSA` and `DSA`.
- Both SM2 and ECDSA have as a parameter the elliptic curve on which to run the algorithms: for EC many different curves are standardized and commonly in use, for SM2 *at the moment* there is only one curve that the standard recommends for use (and it is what we informally refer to as the "SM2 curve"), but nothing prevents the definition of new SM2 recommended curves in the future (e.g. to increase the level of security one could expect the definition of a >512 bits curve, as the current one is around 256 bits), and already now the SM2 standard allows for arbitrary prime and binary curves with the full set of parameters we have for traditional ECDSA/ECDH (even though it is not clear to me how such keys would be serialized), nor we should prevent applications from running ECDSA/ECDH over the "SM2 curve".
- The only real point of contact between SM2 and ECDSA(/DH) is that SM2 does not specify its own distinct serialization format for keys
  - Instead implementations are reusing the SECG serialization used for ECDSA/ECDH keys, but instead of filling the "named curve" identifier member with a curve OID, OID `1.2.156.10197.1.301` for `"SM2" elliptic curve cryptography` is used
  - This OID is not properly a curve OID, and at the moment I don't even know how an SM2 key on another curve would be serialized: actually `1.2.156.10197.1.301` it's even weirder because usually it's leaf nodes in the "OID tree" that are used as identifiers, but in this case `301` represents a class as it has child nodes!
  - An EC curve over the "SM2 curve" at the moment could only be serialized using explicit parameters, because among [the many SM2 related OIDs](http://gmssl.org/docs/oid.html) none is officially identifying the curve itself

So, originally, the issue of coercing "EC(DSA/DH) over SM2 ECC" serialized objects into "SM2" memory objects, could have been an issue to be handled only on deserialization, but the way we handled SM2 in 1.1.1 has created a problem that we have to deal with:
- SM2 was treated as a subtype of ECDSA/ECDH: programmatically one created an `EC_KEY` on the "SM2 curve" and magically/hackishly the application ended up with a peculiar kind of `EC_KEY` object that instead of running ECDSA/ECDH like any other `EC_KEY` ran SM2 as the signing algorithm.
  - There might be applications out there that would expect this behavior of the API to keep working in 3.0
  - My belief is that providing compatibility with legacy somewhat quirky behaviour is a job for the framework and not for the providers that should register EC and/or SM2 keymgmt algorithms.
  - IMO we should avoid mixing again EC and SM2 keys and keep the provider interface simple, so if the framework asks an EC keymgmt for a key it should get back an EC key object, similarly an SM2 keymgmt should be able to generate only an SM2 key object: there shouldn't be the notion of EC keymgmt returning an SM2 object.


## Relevant standards

Using https://github.com/alipay/tls13-sm-spec and https://github.com/metanorma/gmt-0009-2012 for  English translations.
More details about different standardization bodies and the relationship between different standards can be found at https://github.com/alipay/tls13-sm-spec#chinese-algorithm-standards

| [CSTC](http://www.gmbz.org.cn/main/bzlb.html)                                                                           | [NISSTC](http://openstd.samr.gov.cn/bzgk/gb)                                                                         | ISO                  | Notes                 | Name                                                                                                      |
|-------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------|----------------------|-----------------------|-----------------------------------------------------------------------------------------------------------|
| GM/T 0003.1-2012                                                                                                        | [GB/T 32918.1-2016](https://github.com/alipay/tls13-sm-spec/raw/master/sm-en-pdfs/sm2/GBT.32918.1-2016.SM2-en.pdf)   |                      | SM2 Key generation    | Public Key Cryptograhpic Algorithm SM2 Based on Elliptic Curves - Part 1: General                         |
| GM/T 0003.2-2012                                                                                                        | [GB/T 32918.2-2016](https://github.com/alipay/tls13-sm-spec/raw/master/sm-en-pdfs/sm2/GBT.32918.2-2016.SM2-en.pdf)   | ISO/IEC 14888-3:2018 | SM2DSA                | Public Key Cryptograhpic Algorithm SM2 Based on Elliptic Curves - Part 2: Digitial Signature Algorithm    |
| GM/T 0003.3-2012                                                                                                        | [GB/T 32918.3-2016]( https://github.com/alipay/tls13-sm-spec/raw/master/sm-en-pdfs/sm2/GBT.32918.3-2016.SM2-en.pdf ) |                      |                       | Public Key Cryptograhpic Algorithm SM2 Based on Elliptic Curves - Part 3: Key Exchange Protocol           |
| GM/T 0003.4-2012                                                                                                        | [GB/T 32918.4-2016]( https://github.com/alipay/tls13-sm-spec/raw/master/sm-en-pdfs/sm2/GBT.32918.4-2016.SM2-en.pdf ) |                      |                       | Public Key Cryptograhpic Algorithm SM2 Based on Elliptic Curves - Part 4: Public Key Encryption Algorithm |
| GM/T 0003.5-2012                                                                                                        | [GB/T 32918.5-2016]( https://github.com/alipay/tls13-sm-spec/raw/master/sm-en-pdfs/sm2/GBT.32918.5-2016.SM2-en.pdf ) |                      |                       | Public Key Cryptograhpic Algorithm SM2 Based on Elliptic Curves - Part 5: Parameter Definition            |
| GM/T 0009-2012 [(unofficial translation)](https://github.com/metanorma/gmt-0009-2012/blob/master/gmt-0009-2012-en.adoc) | GB/T 35276-2017                                                                                                      |                      | SM2 objects encodings | SM2 Cryptographic Algorithm Application Specification                                                     |
|                                                                                                                         |                                                                                                                      |                      |                       |                                                                                                           |

Last one shows some ingredients for encoding private and public keys, but is not complete.
I couldn't find the sources for the registration of the various OIDs related to SM2, or at least not in English.

levitte:
> ECDSA is its own cryptosystem _with its keygen_ ...

Wait what?

> IMO we should avoid mixing again EC and SM2 keys and keep the provider interface simple, so if the framework asks an EC keymgmt for a key it should get back an EC key object, similarly an SM2 keymgmt should be able to generate only an SM2 key object: there shouldn't be the notion of EC keymgmt returning an SM2 object.

I wholeheartedly agree with this.  Upon a closer look, that won't be very much of a breaking change either.  Basically, `EVP_PKEY_set_alias_type()` will have to stop working.  All other SM2 things are in master only, so there's quite ample space to get things right.

romen:
> > ECDSA is its own cryptosystem _with its keygen_ ...
> 
> Wait what?
> 

Well, technically it is not false, it happens that both ECDH and ECDSA where defined to share the key generation algorithm.

I was oversimplifying, and focusing on the digital signature side of things, as we can see from the table above also SM2 can be declined not only in a digital signature algorithm, but also in key exchange/derivation (a la ECDH) and for public key encryption.

levitte:
I said:
> Upon a closer look, that won't be very much of a breaking change either.  Basically, `EVP_PKEY_set_alias_type()` will have to stop working.  All other SM2 things are in master only, so there's quite ample space to get things right.

I didn't look close enough, it's a bit more than that.

levitte:
Having looked around at the code in `crypto/ec`, I've decided to slowly and quietly back out from legacy SM2 stuff, and do what @romen proposes and keep that strictly separate in provider code.
(in other words, I'm removing the SM2 keygen from this PR and reworking it in another...  as to the legacy SM2 implementation, I frankly have zero desire to dive into that rabbit hole, especially in view of our schedule)

levitte:
In the same vein, I'm move away the keytype change feature from this PR, even though I see uses in the future (if we do decide to "upgrade" EC keys to composite "EC-{grp}" key types)...

levitte:
I reworked this to really only include provider side EC key generation, and let `EVP_PKEY_gen() detect the SM2 curve, and downgrade it to a legacy key.  This is as close to emulating what legacy EC key generation currently does as I can imagine.

levitte:
This is not a draft any more.  Comments welcome

mattcaswell:
> This is not a draft any more. Comments welcome

Is it still pending on other PRs? Is the WIP label still appropriate?

levitte:
> Is it still pending on other PRs?

At least #11358, which quickly fixes the issues with `EVP_PKEY_get0_EC`
I've removed the dependency on #11248, though

> Is the WIP label still appropriate?

I want to try this PR merged with #11358 to see that they want to play well together.

util/libcrypto.num
```diff
@@ -4984,3 +4984,5 @@ EVP_PKEY_gen                            ?	3_0_0	EXIST::FUNCTION:
 EVP_PKEY_CTX_set_rsa_keygen_bits        ?	3_0_0	EXIST::FUNCTION:RSA
 EVP_PKEY_CTX_set_rsa_keygen_pubexp      ?	3_0_0	EXIST::FUNCTION:RSA
 EVP_PKEY_CTX_set_rsa_keygen_primes      ?	3_0_0	EXIST::FUNCTION:RSA
+EVP_PKEY_CTX_set_ec_paramgen_curve_name ?	3_0_0	EXIST::FUNCTION:EC
+EVP_PKEY_CTX_set_ec_paramgen_curve_nid  ?	3_0_0	EXIST::FUNCTION:EC
```
mattcaswell: Do we need some tests for these newly added functions?

mattcaswell (changes requested):
Looks really nice!

providers/implementations/keymgmt/ec_kmgmt.c
```diff
@@ -766,8 +776,171 @@ int ec_validate(void *keydata, int selection)
...
+{
+    struct ec_gen_ctx *gctx = genctx;
+    EC_KEY *ec = templ;
+    const EC_GROUP *ec_group;
+
+    if (gctx == NULL || ec == NULL)
+        return 0;
+    if ((ec_group = EC_KEY_get0_group(ec)) == NULL)
+        return 0;
+    return ec_gen_set_group(gctx, EC_GROUP_get_curve_name(ec_group));
```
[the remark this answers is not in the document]
levitte: I totally agree with you.

util/libcrypto.num
levitte: Good point

providers/implementations/keymgmt/ec_kmgmt.c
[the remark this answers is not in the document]
romen: Me too! It has been a cause of confusion a lot of times for me in the past as well!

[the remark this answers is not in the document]
levitte: (I won't mention publicly the oozing stream of words that came out of my mouth the first time I saw it)

mattcaswell:
Looks like there are a whole bunch of ctrls that need to be handled in EVP_PKEYs that don't have an ameth.

ASN1_PKEY_CTRL_PKCS7_SIGN, ASN1_PKEY_CTRL_CMS_SIGN, ASN1_PKEY_CTRL_CMS_ENVELOPE, ASN1_PKEY_CTRL_CMS_RI_TYPE, ASN1_PKEY_CTRL_SET1_TLS_ENCPT, ASN1_PKEY_CTRL_GET1_TLS_ENCPT

levitte:
So, a good reason for keeping this one WIP was that I hadn't touched `legacy_ctrl_str_to_param()` and `legacy_ctrl_to_param()`...  and I'm not yet sure I got it right.

levitte:
> ASN1_PKEY_CTRL_PKCS7_SIGN, ASN1_PKEY_CTRL_CMS_SIGN, ASN1_PKEY_CTRL_CMS_ENVELOPE, ASN1_PKEY_CTRL_CMS_RI_TYPE, ASN1_PKEY_CTRL_SET1_TLS_ENCPT, ASN1_PKEY_CTRL_GET1_TLS_ENCPT

I'll deal with them later this evening

mattcaswell:
Also #11366 is likely to be a blocker in getting the SSL tests to pass with this PR.

levitte:
> Also #11366 is likely to be a blocker in getting the SSL tests to pass with this PR.

Ok, I'll have a look at that tomorrow

crypto/evp/pmeth_gn.c
```diff
@@ -20,6 +20,17 @@
 #include "crypto/evp.h"
 #include "evp_local.h"
 
+#ifndef FIPS_MODE
+# define TMP_SM2_HACK
+#endif
+
+/* TODO remove when provider SM2 key generation is implemented */
```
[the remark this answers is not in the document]
levitte: I have *no* idea when SM2 key generation is going to happen.  The SM2 situation is ...  not the best, see the debate in #8435.  So, marking a specific version is questionable.

levitte:
Merged with #11358 and #11394, this builds and tests fine with one exception, `test_cms`.  I'm currently working on that one, PR will appear soon

levitte:
The last bit that's needed is for PKCS#7 and CMS to start dealing with provider side keys.  I've started that effort in #11422

crypto/evp/pmeth_lib.c
```diff
@@ -611,6 +611,12 @@ int EVP_PKEY_CTX_get_params(EVP_PKEY_CTX *ctx, OSSL_PARAM *params)
             && ctx->op.ciph.cipher->get_ctx_params != NULL)
         return ctx->op.ciph.cipher->get_ctx_params(ctx->op.ciph.ciphprovctx,
                                                    params);
+    if (EVP_PKEY_CTX_IS_GEN_OP(ctx)
+        && ctx->op.keymgmt.genctx != NULL
+        && ctx->keymgmt != NULL
+        && ctx->keymgmt->gen_get_params != NULL)
+        return evp_keymgmt_gen_get_params(ctx->keymgmt, ctx->op.keymgmt.genctx,
```
slontis: Why do we need this method?

levitte: To pass key generation options, you know the `-pkeyopt` kind of parameters you can pass with `openssl genpkey`

levitte:
#11358 being merged, I rebased this, and squashed all fixups while I was at it

levitte:
I just realised that the CMS code, at least as far as our current tests go, currently never gets to see a provider side key, with only one exception:

https://github.com/openssl/openssl/blob/f5056577ba08b3694aab2722eae1c97bf00acc80/crypto/cms/cms_kari.c#L270-L298

... and of course, since ECDH is right in the middle of that, EC generation would hit CMS where it hurts the most.

In #11422, I recently added a few temporary hacks, simply because the EVP_PKEY_ASN1_METHOD code that `cms_RecipientInfo_kari_encrypt` and `CMS_RecipientInfo_kari_decrypt` use is much more complex than anticipated and requires more thought for rewriting in provider key terms.  I just tried with *only* those hacks added, and lo' and behold, the `test_cms` tests pass flawlessly.

So, instead of keeping this stuck on a PR that needs more effort, I'll have it depend on a much smaller PR that involves temporary hacks similar to what was done for TLS in #11358.

slontis:
> I just realised that the CMS code, at least as far as our current tests go, currently never gets to see a provider side key, with only one exception:

Which is the reason 1024 bit keys for DH were causing a issue when the key size restriction was changed to >= 2048 bit.

levitte:
#11501 is the temporary hack.  I've tested it merged with this PR, and it seems to work flawlessly.

levitte:
> Which is the reason 1024 bit keys for DH were causing a issue when the key size restriction was changed to >= 2048 bit.

Is there a reason we haven't implemented that change in legacy code?

util/libcrypto.num
mattcaswell: This is still outstanding...perhaps can be added later?

mattcaswell (changes requested):

crypto/evp/pmeth_gn.c
```diff
@@ -174,6 +191,77 @@ int EVP_PKEY_gen(EVP_PKEY_CTX *ctx, EVP_PKEY **ppkey)
...
+        char curve_name[OSSL_MAX_NAME_SIZE] = "";
+
+        if (EVP_PKEY_CTX_get_ec_paramgen_curve_name(ctx, curve_name,
+                                                    sizeof(curve_name)) < 1
+            || strcmp(curve_name, "SM2") != 0)
+            goto end;
+    }
+
+    /*
+     * The "magic" in downgrading a provider side key involves serializing
```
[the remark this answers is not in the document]
levitte: Oh, right!

util/libcrypto.num
levitte: Aren't they automatically tested by any EC key generation test we have already?

mattcaswell: Possibly they are

slontis:
> > Which is the reason 1024 bit keys for DH were causing a issue when the key size restriction was changed to >= 2048 bit.
> 
> Is there a reason we haven't implemented that change in legacy code?

Apart from the fact that it breaks backwards compatablity.. The change would be to have the code the same and regenerate data using 2048 bit keys.

crypto/evp/pmeth_gn.c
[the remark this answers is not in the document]
levitte: Reworked

crypto/evp/pmeth_gn.c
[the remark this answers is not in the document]
levitte: [shrug] done

levitte:
> Apart from the fact that it breaks backwards compatablity.. The change would be to have the code the same and regenerate data using 2048 bit keys.

Sooo...  accept 1024 bit keys but refuse to generate lower than 2048?  Not sure I understand what you're arguing, I just think it's *even worse* if we have different generation limits depending on if the EVP_PKEY_METHOD implementations or the provider implementation happens to be called.  Let's at the very least be consistent across the board!

mattcaswell (dismissed):
Approved assuming the CIs agree after rebase when #11501 goes in.

levitte:
Rebased

mattcaswell:
My original approval still stands. Since its only been rebased since then and more than 24 hours have elapsed I consider this ready for merge.

mattcaswell:
Darn. I spoke too soon. Travis failed :-(

There are a number of different problems.

levitte:
The memleak was interesting...  I wonder why we haven't seen it before.

crypto/evp/keymgmt_meth.c
```diff
@@ -38,7 +38,8 @@ static void *keymgmt_from_dispatch(int name_id,
                                    OSSL_PROVIDER *prov)
 {
     EVP_KEYMGMT *keymgmt = NULL;
-    int setparamfncnt = 0, getparamfncnt = 0, setgenparamfncnt = 0;
+    int setparamfncnt = 0, getparamfncnt = 0;
+    int setgenparamfncnt = 0, getgenparamfncnt = 0;;
```
mattcaswell, line 42: Double semi-colon

crypto/evp/pmeth_gn.c
```diff
@@ -20,6 +20,17 @@
 #include "crypto/evp.h"
 #include "evp_local.h"
 
+#ifndef FIPS_MODE
```
mattcaswell, line 23: `#if !defined(FIPS_MODE) && !defined(OPENSSL_NO_EC)`

mattcaswell:
Still a couple of travis failures remaining (issues noted above)

levitte:
I could have sworn I had pushed exactly those fixes, @mattcaswell.  Now done...

romen (approved):
I have been slowly testing during the long weekend and this LGTM.

mattcaswell (approved):

levitte:
Merged

2b9add696573131fc463d098ed8bcbff1b5829d9 KEYMGMT: Add functions to get param/key generation parameters
1f185f51a7899e1eddc9161d7781e3d5ae86ab78 PROV: Implement EC param / key generation
10d756a70e2aeaff0c08e86014075a8623f3e0ab EC: Refactor EVP_PKEY_CTX curve setting macros for param generation
813d31717853252e777b810cb74fa5793fcbb154 EVP: Add a temporary SM2 hack to key generation
49276c3569656a17c24517ff0781967ced2c9658 EVP: fix memleak in evp_pkey_downgrade()

PR #11334, opened 2020-03-16
WIP: TLSv1.2 connection using only FIPS crypto

mattcaswell:
This WIP PR is a whole bunch of miscellaneous tweaks and changes that I have made with the objective of getting a TLS connection working using only crypto from the FIPS module.

The sslprovider_test is used to demonstrate this.

Mostly this was about getting a TLS connection working using a non-default library context without using any crypto from the default library context. Once I got that working it was straight forward to demonstrate it using only FIPS loaded into the non-default library context.

My plan is to keep this as WIP, but pull out various of the commits in here into separate PRs - but I wanted to demonstrate the big picture of where I was heading with this.

Some things to note:

- It is TLSv1.2 only at the moment because I can only do RSA key exchange. We don't have the necessary provider EC/DH key gen/param gen in place yet for anything else. However, TLSv1.3 does not support RSA key exchange.

- This is dependent on @slontis's PR #11240 which enables setting up a non-default libctx with a config file. The commits from there are included here too.

- We currently have not support for provider based MAC generation using the old style EVP_PKEY interface (i.e via EVP_DigestSign). I have updated libssl to use the new style EVP_MAC interface instead. This had an unfortunate consequence however which was not clear to me until after I had implemented it: this completely breaks TLS using the gost engine. This is because you cannot use EVP_MAC with an engine, and so there is no support for this in the gost engine. (@beldmit)

- So far I've only tried a very simple handshake. I've not tried anything else like resumption, PSKs, etc. I expect there to be other issues for those things.


mattcaswell:
I should add that "using only FIPS crypto" might be over stating the case in one area: we don't currently have fetchable RAND. This means that where libssl generates random numbers directly (as opposed to say an algorithm generating random numbers itself as part of its operation), we use the libcrypto DRBG - not the one in the FIPS module. The underlying cipher is still fetched from the FIPS module, but the DRBG itself is not.

PR #11335, opened 2020-03-16
Refactor sm2_id - addendum

levitte:
A few pieces were missing.
Most of all, `test/evp_extra_test.c` is changed so any `EVP_PKEY_CTX_set1_if()` is now done after `EVP_DIgestSignInit()` / `EVP_DigestVerifyInit()`.
Also, `EVP_PKEY_is_a()` is added, to be able to check key types with provider side keys.

### DIFF
--- a/crypto/evp/p_lib.c
+++ b/crypto/evp/p_lib.c
@@ -755,6 +755,14 @@ int EVP_PKEY_base_id(const EVP_PKEY *pkey)
     return EVP_PKEY_type(pkey->type);
 }
 
+int EVP_PKEY_is_a(const EVP_PKEY *pkey, const char *name)
+{
+#ifndef FIPS_MODE
+    if (pkey->keymgmt == NULL)
+        return EVP_PKEY_type(pkey->type) == EVP_PKEY_type(OBJ_sn2nid(name));
+#endif
+    return EVP_KEYMGMT_is_a(pkey->keymgmt, name);
+}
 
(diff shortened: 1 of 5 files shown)


levitte:
Docs added

InfoHunter:
Why locating `EVP_PKEY_CTX_set1_id` after the `init` functions is necessary?

levitte:
> Why locating `EVP_PKEY_CTX_set1_id` after the `init` functions is necessary?

1.  It follows the pattern of *every other stuff you can set with an `EVP_PKEY_CTX`* (`EVP_PKEY_CTX_set1_id` was an [ugly duckling](https://en.wikipedia.org/wiki/The_Ugly_Duckling) in that regard)
2.  The keymgmt to be used is figured out with the `init` function, so before that, there isn't necessarily a backend to send the ID to.

The second point can possibly be refactored so the keymgmt is figured out when creating the `EVP_PKEY_CTX`...  that shouldn't even be very hard, I cannot remember for the moment why we do that so late.

levitte:
Let me get back to that second point in form of another PR

levitte:
@InfoHunter, #11343 may remove the need to reorder the `EVP_PKEY_CTX_set1_id()` calls.

InfoHunter:
Great, I will look into this tomorrow

levitte:
This is superceded by #11343.  Closing

PR #11336, opened 2020-03-16
Clean up after PR #10943

kaduk:
Don't leak the SSL_CTX pair in sslapitest by creating a new one on top
of them, and fix a typo in a comment.

[extended tests]

##### Checklist

### DIFF
--- a/test/sslapitest.c
+++ b/test/sslapitest.c
@@ -655,7 +655,7 @@ end:
  * Very focused test to exercise a single case in the server-side state
  * machine, when the ChangeCipherState message needs to actually change
  * from one cipher to a different cipher (i.e., not changing from null
- * encryption to reall encryption).
+ * encryption to real encryption).
  */
 static int test_ccs_change_cipher(void)
 {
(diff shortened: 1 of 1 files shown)


kaduk:
Marking urgent since this seems to be blocking the 1.1.1e release.
cc @mattcaswell

mattcaswell:
Repo is now frozen, but I will push this when the CIs finish.

kaduk:
Thanks!  Sorry for making the mess in the first place.

levitte:
I would like to suggest the following (entire) commit message:

```
test/sslapitest.c: Don't leak the SSL_CTX pair

Create a new one on top of them, and fix a typo in a comment.

[extended tests]
```

Referring to a PR in the commit message is kinda iffy in my opinion.

kaduk:
I'm happy to see the commit message change, but would propose instead

```
sslapitest: don't leak the SSL_CTX pair

We have no need for a new set of SSL_CTXs in test_ccs_change_cipher(), so just
keep using the original ones.  Also, fix a typo in a comment.

[extended tests]
```

I'm fine with Matt making that change while merging (and not re-running the CI).
Looking at what has run of the CI so far, travis seems to be thinking that the boring and krb5 tests are failing (IIRC, well-known), and also that (SSLv3?) client auth is breaking.  I don't remember offhand whether the latter is known (and am in a meeting at the moment so looking is hard)

levitte:
I'm fine with your proposed commit message, @kaduk

mattcaswell:
Pushed. Thanks.

PR #11340, opened 2020-03-17
Add EVP_PKEY API to retrieve raw OSSL_PARAM values from a provided key

slontis:
##### Checklist

### DIFF
--- a/crypto/evp/keymgmt_lib.c
+++ b/crypto/evp/keymgmt_lib.c
@@ -370,3 +370,82 @@ void *evp_keymgmt_util_gen(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
 
     return keydata;
 }
+
+struct export_todata_st {
+    OSSL_PARAM *params;
+};
+
+/*
+ * Some keys have multiple instances, so mark them as used so that they are only
+ * found once.
+ */
+static OSSL_PARAM *locate_used(const OSSL_PARAM *params, const char *key,\
+                               int used[], int used_len)
+{
+    int i;
+    OSSL_PARAM *p = (OSSL_PARAM *)params;
+
+    for (i = 0; p->key != NULL && i < used_len; p++, i++) {
+        if (!used[i] && strcmp(key, p->key) == 0) {
+            used[i] = 1;
+            return p;
+        }
+    }
+    return NULL;
+}
+
+#define OSSL_PARAM_TODATA_MAX 25
+
+static int todata(const OSSL_PARAM src[], void *cbarg)
+{
+    int i;
+    struct export_todata_st *data = cbarg;
+    OSSL_PARAM *p, *dst = data->params;
+    int used[OSSL_PARAM_TODATA_MAX];
+
+    memset(used, 0, sizeof(used));
+
+    for (i = 0; dst[i].key != NULL; ++i) {
+        p = locate_used(src, dst[i].key, used, OSSL_PARAM_TODATA_MAX);
+        if (p == NULL)
+            continue;
+        if (dst[i].data_type != p->data_type)
+            return 0;
+        if (p->return_size > dst[i].data_size)
+            return 0;
+        switch (dst[i].data_type) {
+        case OSSL_PARAM_UTF8_PTR:
+        case OSSL_PARAM_OCTET_PTR:
+            dst[i].data = p->data;
+            dst[i].data_size = p->data_size;
+            dst[i].return_size = p->return_size;
+            break;
+        case OSSL_PARAM_REAL:
+        case OSSL_PARAM_INTEGER:
+        case OSSL_PARAM_UNSIGNED_INTEGER:
+        case OSSL_PARAM_OCTET_STRING:
+        case OSSL_PARAM_UTF8_STRING:
+            memcpy(dst[i].data, p->data, p->data_size);
+            dst[i].data_size = p->data_size;
+            dst[i].return_size = p->return_size;
+            break;
+        default:
+            continue;
+        }
+    }
+    return 1;
+}
+
+int evp_keymgmt_util_todata(EVP_PKEY *pkey, int selection, OSSL_PARAM params[])
+{
+    struct export_todata_st exprt_todata;
+
+    if (pkey == NULL
+        || pkey->keymgmt == NULL
+        || pkey->keymgmt->export == NULL)
+        return 0;
+
+    exprt_todata.params = params;
+    return evp_keymgmt_export(pkey->keymgmt, pkey->keydata, selection,
+                              &todata, &exprt_todata);
+}
--- a/include/openssl/evp.h
+++ b/include/openssl/evp.h
@@ -1398,6 +1398,7 @@ int EVP_PKEY_CTX_set_signature_md(EVP_PKEY_CTX *ctx, const EVP_MD *md);
 # define EVP_PKEY_OP_DECRYPT             (1<<11)
 # define EVP_PKEY_OP_DERIVE              (1<<12)
 
+
 # define EVP_PKEY_OP_TYPE_SIG    \
         (EVP_PKEY_OP_SIGN | EVP_PKEY_OP_VERIFY | EVP_PKEY_OP_VERIFYRECOVER \
                 | EVP_PKEY_OP_SIGNCTX | EVP_PKEY_OP_VERIFYCTX)
@@ -1592,6 +1593,8 @@ int EVP_PKEY_key_fromdata_init(EVP_PKEY_CTX *ctx);
 int EVP_PKEY_fromdata(EVP_PKEY_CTX *ctx, EVP_PKEY **ppkey, OSSL_PARAM param[]);
 const OSSL_PARAM *EVP_PKEY_param_fromdata_settable(EVP_PKEY_CTX *ctx);
 const OSSL_PARAM *EVP_PKEY_key_fromdata_settable(EVP_PKEY_CTX *ctx);
+int EVP_PKEY_todata(EVP_PKEY *pkey, OSSL_PARAM param[]);
+
 int EVP_PKEY_paramgen_init(EVP_PKEY_CTX *ctx);
 int EVP_PKEY_paramgen(EVP_PKEY_CTX *ctx, EVP_PKEY **ppkey);
 int EVP_PKEY_keygen_init(EVP_PKEY_CTX *ctx);
--- a/test/evp_pkey_provided_test.c
+++ b/test/evp_pkey_provided_test.c
@@ -12,6 +12,7 @@
 #include <openssl/serializer.h>
 #include <openssl/provider.h>
 #include <openssl/params.h>
+#include <openssl/ec.h>
 #include <openssl/core_names.h>
 #include "crypto/ecx.h"
 #include "internal/nelem.h"
@@ -495,7 +496,7 @@ err:
     return ret;
 }
 
-static int test_fromdata_ec(void)
+static int test_fromdata_todata_ec(void)
 {
     int ret = 0;
     EVP_PKEY_CTX *ctx = NULL;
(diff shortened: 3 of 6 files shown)


include/openssl/evp.h
```diff
@@ -1398,6 +1398,7 @@ int EVP_PKEY_CTX_set_signature_md(EVP_PKEY_CTX *ctx, const EVP_MD *md);
 # define EVP_PKEY_OP_DECRYPT             (1<<11)
 # define EVP_PKEY_OP_DERIVE              (1<<12)
 
+
```
paulidale: ???

test/evp_pkey_provided_test.c
```diff
@@ -550,6 +554,37 @@ static int test_fromdata_ec(void)
...
+        goto err;
+
+    /* The public key is returned is in COMPRESSED FORMAT */
+    if (!TEST_true(pub[0] == (POINT_CONVERSION_COMPRESSED + 1))
+        || !TEST_mem_eq(pub + 1, todata_params[1].data_size - 1,
+                        ec_pub_keydata + 1, compressed_sz - 1))
+        goto err;
+
+    /*
+     * TODO(3.0): How do we compare bignums here?
```
paulidale: Convert to BN and use the test macro?

crypto/evp/keymgmt_lib.c
```diff
@@ -370,3 +370,82 @@ void *evp_keymgmt_util_gen(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+ * found once.
+ */
+static OSSL_PARAM *locate_used(const OSSL_PARAM *params, const char *key,\
+                               int used[], int used_len)
+{
+    int i;
+    OSSL_PARAM *p = (OSSL_PARAM *)params;
+
+    for (i = 0; p->key != NULL && i < used_len; p++, i++) {
+        if (!used[i] && strcmp(key, p->key) == 0) {
```
paulidale: Ugh :(

crypto/evp/keymgmt_lib.c
```diff
@@ -370,3 +370,82 @@ void *evp_keymgmt_util_gen(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+            used[i] = 1;
+            return p;
+        }
+    }
+    return NULL;
+}
+
+#define OSSL_PARAM_TODATA_MAX 25
+
+static int todata(const OSSL_PARAM src[], void *cbarg)
```
paulidale: I can't say I like this.  Passing the caller's OSSL_PARAM array to the export and filling the fields they want seems more natural than copying everything here.  Still, if the fields aren't known.....

Could an OSSL_PARAM array and its size be passed to the export and it gets filled but not overflowed without this callback being involved?

test/evp_pkey_provided_test.c
slontis: The data coming back is in reverse order - how does converting to BN help with this?

crypto/evp/keymgmt_lib.c
slontis: The export is there to do export 'everything'.. I have just hijacked it..
The callback is normally used for the import process (using the params that are built)

crypto/evp/keymgmt_lib.c
slontis: RSA is a good example of something with multiple fields that are the same, which is particularly ugly (i.e- how do we know how many it has before we do the call).

include/openssl/evp.h
slontis: I added a op and then deleted it - but not the newline

crypto/evp/keymgmt_lib.c
paulidale: Maybe a "query all possible params to export", a "clone a param list" and an "export" would simplify things.

slontis: The normal export is also a import

crypto/evp/keymgmt_lib.c
levitte: > (i.e- how do we know how many it has before we do the call).

find out by getting the value of the "primes" (**SSL_PKEY_PARAM_RSA_PRIMES**)

crypto/evp/keymgmt_lib.c
levitte: > The normal export is also a import

... depending on whose point of view.  What libcrypto exports, the provider imports, and vice versa

levitte:
Since these are values that the caller knows (or *should* know) the name of, I don't see why `OP_keymgmt_get_params()` isn't used for this?  And yeah, it means that we need to add cases for the get_params functions.

In a way, it means that OP_keymgmt_export() and OP_keymgmt_get_params() do more or less the same thing, but with a twist; OP_keymgmt_export() is meant for those who *do not know* and *do not need to know* the exact contents of the key beforehand, all the caller wants is a bunch of data, while OP_keymgmt_get_params() is meant for those who *do know*, and that may want to extract very specific data.

I've probably made the mistake of saying (even documenting) that op_keymgmt_get_params() should only be concerned with key meta-data -- I lost track of the need for what this PR is trying to achieve.  That was a mistake, there's really nothing that says that an asymmetric key implementation couldn't let OP_keymgmt_get_params() return *any data they want*, including the key data.

test/evp_pkey_provided_test.c
```diff
@@ -550,6 +554,37 @@ static int test_fromdata_ec(void)
...
+     * The value returned in priv is in BE order.
```
levitte: Er, what?  The value of priv *should* come in whatever byte order a native integer uses.

crypto/evp/keymgmt_lib.c
paulidale: No!

When we get a new algorithm that requires _n_ happy numbers to work, asking for primes isn't useful.

crypto/evp/keymgmt_lib.c
paulidale: I still believe that a query of all parameters makes sense.  I'd work with a query of all parameters and their values, plus a separate free function.

test/evp_pkey_provided_test.c
slontis: Could we have a seperate type for OSSL_PARAM_TYPE_BIGNUM that doesnt need to do native order....  Then we dont need to do this reversing bytes stuff that is confusing,
(I keep on saying its confusing because IT IS CONFUSING :)).
i.e- fromdata takes a non native unsigned char[], todata currently returns a unsigned char[] with reversed byte order (THIS IS NOT GOOD).

slontis:
> And yeah, it means that we need to add cases for the get_params functions.

I guess I was trying to prevent having to duplicate nearly identical code in multiple places.
(It may lead to some out of synch bugs as we move forward over time).

crypto/evp/keymgmt_lib.c
levitte: But supposedly, *the application* knows what to ask for.

"Know your provider" -- me

test/evp_pkey_provided_test.c
levitte: I don't think that `BIGNUM` is good at all.  However, since the data type is supposed to describe memory organization, I'd suggest something like `OSSL_PARAM_TYPE_INTEGER_LE`, `OSSL_PARAM_TYPE_UNSIGNED_INTEGER_LE`, `OSSL_PARAM_TYPE_INTEGER_BE` `OSSL_PARAM_TYPE_UNSIGNED_INTEGER_BE`, and either end that cares must deal with the difference.

The reason I don't want to treat BIGNUM as an essentially different type than any other number can be illustrated with RSA's *e*.  That's usually a small number, and it *should be perfectly legitimate for one end to treat it as an `unsigned int` and the other to treat it as a `BIGNUM` without having to jump through extra hoops*.

levitte:
> I guess I was trying to prevent having to duplicate nearly identical code in multiple places.

I haven't looked too closely, but I wouldn't be surprised if one can be made in terms of the other.
The other option could be to expose the export function to the public, including the callback.  I'm not too keen on that, but

test/evp_pkey_provided_test.c
levitte: `BN_native2bn` does the right thing.

levitte: And create a BIGNUM with `BN_bin2bn` for `ec_priv_keydata`, 'cause I suppose that one is in BE order, right?

richsalz:
Does requiring something like PEM serialization, which has a standard format for all keys, lessen the need for this?

levitte:
> Does requiring something like PEM serialization, which has a standard format for all keys, lessen the need for this?

DER serialization should suffice...  but no, not in the general case.  We want something that helps applications to extract data from a key, such as *n* from an RSA key, or *xyzzy* from a BLARGH key (the latter being fictional, of course, but serves the purpose of reminding us all that we need to be able to deal with what's unknown to us)

richsalz:
So perhaps for 3.0 we only support DER serialization, and once we get more data-points we try to solve the general case.

levitte:
That's easy, just tell the application authors serialize the key to DER using OSSL_SERIALIZER, then de serialize the result with the appropriate d2i function to convert it to whatever the type may be (DH is the popular example du jour). That can be done, here and now, yeah.

richsalz:
So you're saying that the problem we know about is solved?  And we can put this aside for 3.0?

levitte:
For once, yeah it's possible to do, limited to the key types we know internally.

test/evp_pkey_provided_test.c
slontis: I dont agree with this logic.. 
I think this is dealing with an edge case that is not needed.
If we want to put a constant into a bignum anywhere else in the code we convert it to a bignum.
For the todata operation - creating a unsigned char array shouldnt be a big deal.

It essentially boils down to this for me...
If I feed an array of bytes into a system, I expect it to be able to be extracted as an array of bytes in exactly the same order (not reversed due to some internal implementation details).

There is no way of knowing that a number is an actual bignum currently (only the end point API's convert to and from a  bignum).  While this isnt a problem at the moment we might find a case where it becomes a issue and then we cant fix it because we have shared the UNSIGNED_INT type.  I have no problem if it asserts if a int is assigned to a bignum type.

slontis: > I'd suggest something like OSSL_PARAM_TYPE_INTEGER_LE, OSSL_PARAM_TYPE_UNSIGNED_INTEGER_LE, 

You are talking about an internal representation which seems irrelevant.
What about OSSL_PARAM_TYPE_BIG_UNSIGNED_INTEGER?

slontis:
> So perhaps for 3.0 we only support DER serialization, and once we get more data-points we try to solve the general case.

Wouldnt recommend it.
The PEM format allows extra data to be added onto the end, which may be real useful (for key validation for example).

test/evp_pkey_provided_test.c
levitte: Why is BIG so important? Doesn't the data size tell you that it's big relative to whatever types you are using?

The point here is that you cannot know what the other end is doing, what kind of variables they're using etc. You *happen to know* when you're authoring both ends of the transferred data, but that's not always going to be the case. This is also why OSSL_PARAM is a public structure.
The data type is *exactly* about representation, although I call it memory organisation, there's nothing "internal" about that. This is by design.

levitte:
@slontis,what you quote says DER, why did you go back to PEM?

test/evp_pkey_provided_test.c
slontis: See the bit above that says 'It essentially boils down to this for me...'
That is my issue - the only way to solve that it to know its a bignum. I cant tell otherwise.
The internal representation is affecting the public interface - that is not good.

levitte: It's a number, represented as an array of bytes in native byte order. This goes for *any* number (integer, to be precise), regardless of size and regardless of the language type you're using for it.

slontis: Its a array that gets reversed when it comes back out.

levitte: The point here is that it seems you'd rather have OCTET_STRING to transfer these bits. And that's fair, I don't always see a point of insisting that the key data be seen as numbers, even though they are *internally* treated as such. The EC case is a bit absurd in that regard, with one half of the key being an octet string and the other being a number..

slontis: exactly - but then it still needs the type so it knows its a bignum right?

levitte: [sigh]
@slontis, you want something that's not represented in native byte order, but rather specifically in big endian order. I proposed data types for that. How about we go from there?

levitte: It being used as a BIGNUM on either end is internal information.

slontis:
Replaced by #11365

PR #11341, opened 2020-03-17
ASYNC: Remove funcargs copy of async function

uudiin:
It is not necessary to back up the funcargs of the ASYNC function.
Unless the parameters are in the stack of other coroutines, and the
stack loses data due to the function return, in fact this situation
will never happen. This also optimizes the performance of ASYNC to
a certain extent.

If there is no problem with this patch, the 'size' parameter of
ASYNC_start_job can also be removed.

Signed-off-by: Tianjia Zhang <tianjia.zhang@linux.alibaba.com>

##### Checklist

### DIFF
--- a/crypto/async/async.c
+++ b/crypto/async/async.c
@@ -91,7 +91,6 @@ static ASYNC_JOB *async_job_new(void)
 static void async_job_free(ASYNC_JOB *job)
 {
     if (job != NULL) {
-        OPENSSL_free(job->funcargs);
         async_fibre_free(&job->fibrectx);
         OPENSSL_free(job);
     }
@@ -138,7 +137,6 @@ static void async_release_job(ASYNC_JOB *job) {
         ERR_raise(ERR_LIB_ASYNC, ERR_R_INTERNAL_ERROR);
         return;
     }
-    OPENSSL_free(job->funcargs);
     job->funcargs = NULL;
     sk_ASYNC_JOB_push(pool->jobs, job);
 }
@@ -249,18 +247,7 @@ int ASYNC_start_job(ASYNC_JOB **job, ASYNC_WAIT_CTX *wctx, int *ret,
         if ((ctx->currjob = async_get_pool_job()) == NULL)
             return ASYNC_NO_JOBS;
 
-        if (args != NULL) {
-            ctx->currjob->funcargs = OPENSSL_malloc(size);
-            if (ctx->currjob->funcargs == NULL) {
-                async_release_job(ctx->currjob);
-                ctx->currjob = NULL;
-                return ASYNC_ERR;
-            }
-            memcpy(ctx->currjob->funcargs, args, size);
-        } else {
-            ctx->currjob->funcargs = NULL;
-        }
-
+        ctx->currjob->funcargs = args;
         ctx->currjob->func = func;
         ctx->currjob->waitctx = wctx;
         libctx = ossl_lib_ctx_get_concrete(NULL);


mattcaswell:
close/reopen to kick cla bot

uudiin:
@InfoHunter By the way, review this patch too, thanks.

InfoHunter:
Sure

uudiin:
In other words, the args parameter is transparent to the functions of the ASYNC series, as long as it is responsible for passing this parameter value, which also conforms to the traditional user data transfer rules. As for the specific data pointed to by args, it should be the caller To ensure its effectiveness, it should actually be done.
On the other hand, if there are pointers to the data in args elsewhere, and operations are performed on the data here, ASYNC copies a copy of the data here, there is no way to guarantee data consistency, which is also a potential defect.

InfoHunter:
One concern brought by this change is that this will break the API compatibility - although in OpenSSL 3.0.0 the compatibility is not mandatory but I am not sure if it's good to really break it.

uudiin:
> One concern brought by this change is that this will break the API compatibility - although in OpenSSL 3.0.0 the compatibility is not mandatory but I am not sure if it's good to really break it.

Yes, if this patch is acceptable, the ‘size’ of ASYNC_start_job() is a useless parameter. If it is not mandatory for compatibility, this parameter can be removed.

InfoHunter:
Another problem is the usage of the API has been changed as well. Say, someone can free the `args` after calling the `ASYNC_start_job()` but if he still does so with this patch, the caller might crash...

I don't know how current users (like Intel QAT engine) of ASYNC handle the `args`.

uudiin:
> Another problem is the usage of the API has been changed as well. Say, someone can free the `args` after calling the `ASYNC_start_job()` but if he still does so with this patch, the caller might crash...
> 
> I don't know how current users (like Intel QAT engine) of ASYNC handle the `args`.

I probably looked at the code of QAT. QAT does not directly call `ASYNC_start_job()` to create a coroutine. It seems that this operation should be responsible for the upper caller of QAT. I did not find `ASYNC_start_job()` in the code of QAT. But `ASYNC_pause_job()` is there.

InfoHunter:
Ping @mattcaswell , any thoughts on this?

mattcaswell:
This change looks problematic to me.

Imagine code like this:
````
int testfunc(ASYNC_JOB **job, ASYNC_WAIT_CTX *wctx, int *ret)
{
    char myarray[10] = { /* whatever */ };

    return ASYNC_start_job(job, wctx, ret, myasyncfunc, myarray, sizeof(myarray));
}
````
The passed in argument is stack allocated and will no longer be valid as soon as this function exits - but usage of the argument may continue for the duration of the async job. I think it is for this very reason that the copy occurs.

uudiin:
> This change looks problematic to me.
> 
> Imagine code like this:
> 
> ```
> int testfunc(ASYNC_JOB **job, ASYNC_WAIT_CTX *wctx, int *ret)
> {
>     char myarray[10] = { /* whatever */ };
> 
>     return ASYNC_start_job(job, wctx, ret, myasyncfunc, myarray, sizeof(myarray));
> }
> ```
> 
> The passed in argument is stack allocated and will no longer be valid as soon as this function exits - but usage of the argument may continue for the duration of the async job. I think it is for this very reason that the copy occurs.

Yes, it is really problematic to use it. I think it should not be used in this way. This usage should be restricted. I think it is more like a `pthread_create()` function. The caller should ensure the validity of `args`. Use `malloc()` function to allocate space instead of allocating it to the stack.

mattcaswell:
We can't modify the behaviour of existing functions in this way. That would break apps that already call this function in the manner I describe above. We *could* introduce a new function with a slightly different name that does the same thing. But that has the disadvantage of complicating the API so is not something to be done lightly. I'm not convinced its worth it.

uudiin:
> We can't modify the behaviour of existing functions in this way. That would break apps that already call this function in the manner I describe above. We _could_ introduce a new function with a slightly different name that does the same thing. But that has the disadvantage of complicating the API so is not something to be done lightly. I'm not convinced its worth it.

OK, thanks for your reply, I got it.
If you need to define a new API to achieve this function, I tend to split `ASYNC_start_job()` into two APIs, `ASYNC_job_create()` and `ASYNC_job_resume()`, which are used to create and run async tasks respectively. This can also fundamentally simplify the logic of ASYNC functions.

t8m:

t8m:
@uudiin How do you plan to address Matt's comments? This is really not mergeable in the current form.

uudiin:
> @uudiin How do you plan to address Matt's comments? This is really not mergeable in the current form.

@t8m @mattcaswell This modification does change the function behavior, but it does not affect compatibility, is opaque to users, and does not affect existing stock applications. In short, the patch only reduces some unnecessary memory allocations.

t8m:
I disagree that it is fully opaque. The comment https://github.com/openssl/openssl/pull/11341#issuecomment-614665769 describes the use case that will be broken.

t8m:

PR #11342, opened 2020-03-17
Update the NEWS file for the new release

mattcaswell:


mattcaswell:
Marked as urgent since this is needed for the release today

t8m:
Is this the only major change?

What about db943f43 57225c99 e9cd6e7 - they should be at least mentioned in CHANGES.

mattcaswell:
> Is this the only major change?

I didn't do a review of all the changes to see what updates should be made - on the assumption that we've updated CHANGES as we go (big assumption). I was more concerned with ensuring that NEWS was consistent. However I'll add some CHANGES entries for the items you pointed out.

mattcaswell:
New commit pushed adding the above entries to CHANGES.

NEWS
```diff
@@ -7,7 +7,8 @@
 
   Major changes between OpenSSL 1.1.1d and OpenSSL 1.1.1e [under development]
 
-      o
+      o Fixed an an overflow bug in the x64_64 Montgomery squaring procedure
```
InfoHunter: There is an extra `an`

mattcaswell: Oops. Fixed

mattcaswell:
Ping

mattcaswell:
Pushed. Thanks.

PR #11343, opened 2020-03-17
EVP: fetch the EVP_KEYMGMT earlier

levitte:
Instead of fetching the EVP_KEYMGMT in the init for every different
operation, do it when creating the EVP_PKEY_CTX.

This allows certain control functions to be called between the
creation of the EVP_PKEY_CTX and the call of the operation's init
function.

Use case: EVP_PKEY_CTX_set1_id(), which is allowed to be called very
early with the legacy implementation, this should still be allowed
with provider implementations.


include/crypto/evp_pkey.h
```diff
@@ -0,0 +1,144 @@
+/*
```
InfoHunter: what is the relation between this file and the `include/crypto/evp_pkey.h` one?

mattcaswell: Err...this is that file? Or did you mean evp_pkey_ctx.h?

mattcaswell: Although actually I don't understand what either of these files are being used for...and they seem to have some content duplicated between them.

levitte: Er..... I have no idea why this ended up in here...

levitte: Removed

levitte:
I've tested this merged with #11328, and the previous failures in `test/evp_extra test` went away.

include/crypto/evp_pkey.h
InfoHunter: > Err...this is that file? Or did you mean evp_pkey_ctx.h?

Right, I meant that file...

PR #11349, opened 2020-03-18
Update provider.pod

mattcaswell:
The provider.pod document was missing some references to more recent
operation types. Also, the section on explicit fetch was a little out of
date and did not cover how we do this with an EVP_PKEY_CTX.

### DIFF
--- a/doc/man7/provider.pod
+++ b/doc/man7/provider.pod
@@ -147,6 +147,30 @@ The number for this operation is B<OSSL_OP_KEYEXCH>.
 The functions the provider can offer are described in
 L<provider-keyexch(7)>
 
+=item Asymmetric Ciphers
+
+In the OpenSSL libraries, the corresponding method object is
+B<EVP_ASYM_CIPHER>.
+The number for this operation is B<OSSL_OP_ASYM_CIPHER>.
+The functions the provider can offer are described in
+L<provider-asym_cipher(7)>
+
+=item Signatures
+
+In the OpenSSL libraries, the corresponding method object is
+B<EVP_SIGNATURE>.
+The number for this operation is B<OSSL_OP_SIGNATURE>.
+The functions the provider can offer are described in
+L<provider-signature(7)>
+
+=item Key Management
+
+In the OpenSSL libraries, the corresponding method object is
+B<EVP_KEYMGMT>.
+The number for this operation is B<OSSL_OP_KEYMGMT>.
+The functions the provider can offer are described in
+L<provider-keymgmt(7)>
+
 =item Serialization
 
 In the OpenSSL libraries, the corresponding method object is
@@ -165,10 +189,9 @@ I<NOTE: This section is mostly interesting to OpenSSL users.>
 
 Users of the OpenSSL libraries never query the provider directly for
 its diverse implementations and dispatch tables.
-Instead, the diverse OpenSSL APIs often have fetching functions that
-do the work, and they return an appropriate method object back to the
-user.
-These functions usually have the name C<APINAME_fetch>, where
+Instead, the OpenSSL APIs often have fetching functions that do the work, and
+they return an appropriate method object back to the user.
+These functions often have the name C<APINAME_fetch>, where
 C<APINAME> is the name of the API, for example L<EVP_MD_fetch(3)>.
 
 These fetching functions follow a fairly common pattern, where three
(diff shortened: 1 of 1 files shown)


doc/man7/provider.pod
```diff
@@ -147,6 +147,30 @@ The number for this operation is B<OSSL_OP_KEYEXCH>.
 The functions the provider can offer are described in
 L<provider-keyexch(7)>
 
+=item Asymmetric Ciphers
+
+In the OpenSSL libraries, the corresponding method object is
+B<EVP_ASYM_CIPHER>.
```
levitte: In some manuals, for example `doc/man3/EVP_MAC.pod`, the type itself is one of the names of that manual.  The idea behind that is that if you want to know something about a type, you will be able to do `man EVP_MAC` and get at least a starting page.  If we did that consistently, `B<EVP_ASYM_CIPHER>` could be changed to `L<EVP_ASYM_CIPHER(3)>`, permitting easy links.

I would see that as an advantage, what are your thoughts?

doc/man7/provider.pod
```diff
@@ -147,6 +147,30 @@ The number for this operation is B<OSSL_OP_KEYEXCH>.
...
+The functions the provider can offer are described in
+L<provider-signature(7)>
+
+=item Key Management
+
+In the OpenSSL libraries, the corresponding method object is
+B<EVP_KEYMGMT>.
+The number for this operation is B<OSSL_OP_KEYMGMT>.
+The functions the provider can offer are described in
+L<provider-keymgmt(7)>
```
levitte: It might be worth saying already here that this type is seldom used alone, that it's normally paired with methods that use asymmetric keys, i.e. EVP_KEYEXCH, EVP_ASYM_CIPHER and EVP_SIGNATURE

doc/man7/provider.pod
```diff
@@ -204,6 +227,15 @@ an implementation.
 The method object that is fetched can then be used with diverse other
 functions that use them, for example L<EVP_DigestInit_ex(3)>.
 
+Operations that work with asymmetric keys via an B<EVP_PKEY_CTX> have a slightly
+different form. In those cases the fetch of the algorithm is done internally
```
levitte: Is there a difference between "internally" and "implicitly" here?  I view it as an implicit fetch, the only difference being that there's currently (*) no way to pass an explicitly fetched EVP_KEYMGMT.

(*) we *could* have a `EVP_PKEY_CTX_new_from_keymgmt()`...  I don't currently see an actual need, and it kinda defeats the purpose with how EVP_PKEY_CTX is created, even with legacy methods, as there's no way to create one from an `EVP_PKEY_METHOD` or an `EVP_PKEY_ASN1_METHOD` either...

doc/man7/provider.pod
mattcaswell: Well, it was actually writing this section that inspired we to raise the question in #11350. Because if we decide that they shouldn't be public at all then we'd need to tweak the wording again. However for the purposes of this PR I thought it should still be mentioned. I can add some words along the lines you suggest.

levitte: If those types are made non-public, *they shouldn't be mentioned at all in public documentation*

mattcaswell: > If those types are made non-public, they shouldn't be mentioned at all in public documentation

Oh, yes I agree. But as right now they are public, they should be mentioned. If at some later point in time they are no longer public then I expect a later PR to come and change this wording again.

doc/man7/provider.pod
levitte: Ok, that argument works for me
(as for new terms, if you still want to toy with that, how about "demi-explicit"? :wink:)

doc/man7/provider.pod
mattcaswell: Yes - I think it would be great if we had that.

doc/man7/provider.pod
levitte: Deal!

doc/man7/provider.pod
levitte: It's really just a matter of adding those names in the proper files.  I would suggest that they can be added in the corresponding *TYPE*_new.pod, and that file can be renamed to *TYPE*.pod.

richsalz: Strong suggestion to do "make doc-nits cmd-nits" before you push :)

slontis:
@mattcaswell ?

t8m:
Is this still relevant?

NostalgicGareth (approved):

mattcaswell:
No longer relevant.

PR #11353, opened 2020-03-18
Make it possible to easily specify a libctx for EVP_DigestSign*

mattcaswell:
EVP_DigestSignInit_ex and EVP_DigestVerifyInit_ex did not provide the
capability to specify an explicit OPENSSL_CTX parameter. It is still
possible by explicitly setting an EVP_PKEY_CTX - but in most cases it
would be much simpler to just specify it in the Init call. We add the
capability to do that.

### DIFF
--- a/crypto/evp/m_sigver.c
+++ b/crypto/evp/m_sigver.c
@@ -200,6 +200,9 @@ static int do_sigver_init(OPENSSL_CTX *libctx, EVP_MD_CTX *ctx,
      */
     ERR_pop_to_mark();
 
+    if (type == NULL && mdname != NULL)
+        type = EVP_get_digestbyname(mdname);
+
     if (ctx->pctx->pmeth == NULL) {
         EVPerr(0, EVP_R_OPERATION_NOT_SUPPORTED_FOR_THIS_KEYTYPE);
         return 0;
--- a/crypto/evp/m_sigver.c
+++ b/crypto/evp/m_sigver.c
@@ -36,10 +36,10 @@ static const char *canon_mdname(const char *mdname)
     return mdname;
 }
 
-static int do_sigver_init(OPENSSL_CTX *libctx, EVP_MD_CTX *ctx,
-                          EVP_PKEY_CTX **pctx, const EVP_MD *type,
-                          const char *mdname, const char *props, ENGINE *e,
-                          EVP_PKEY *pkey, int ver)
+static int do_sigver_init(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
+                          const EVP_MD *type, const char *mdname,
+                          const char *props, ENGINE *e, EVP_PKEY *pkey,
+                          OPENSSL_CTX *libctx, int ver)
 {
     EVP_PKEY_CTX *locpctx = NULL;
     EVP_SIGNATURE *signature = NULL;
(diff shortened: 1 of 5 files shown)


crypto/evp/m_sigver.c
```diff
@@ -59,8 +59,12 @@ static int do_sigver_init(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
...
     }
 
-    if (ctx->pctx == NULL)
-        ctx->pctx = EVP_PKEY_CTX_new(pkey, e);
+    if (ctx->pctx == NULL) {
+        if (libctx != NULL)
+            ctx->pctx = EVP_PKEY_CTX_new_from_pkey(libctx, pkey, props);
+        else
+            ctx->pctx = EVP_PKEY_CTX_new(pkey, e);
+    }
```
levitte: And what do you do in the `ctx->pctx != NULL && ctx->pctx->libctx != NULL && libctx != NULL` case?  Why is one chosen over the other?  Could that be problematic?

mattcaswell: The libctx from the existing pctx takes precedence. From the docs:

````
If the EVP_PKEY_CTX to be used is
created by EVP_DigestSignInit_ex then it will use the B<OPENSSL_CTX> specified
in I<libctx> and the property query string specified in I<props>.
````
So...libctx is only used if a pctx is created.

Comment (levitte) [reply crypto/evp/m_sigver.c]: Hmmm, ok.

Comment (t8m) [reply crypto/evp/m_sigver.c]: I am not sure in this case whether putting the libctx as first parameter isn't misleading. Basically as it is only auxiliary parameter which is not always used, it should not be the first parameter.

File: crypto/evp/m_sigver.c [hunk 2]
```diff
@@ -196,6 +200,9 @@ static int do_sigver_init(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
      */
     ERR_pop_to_mark();
 
+    if (type == NULL && mdname != NULL)
```
slontis: Why is this legacy path being hit by TLS?

mattcaswell: In `tls_psk_do_binder` we have to create an HMAC:

https://github.com/openssl/openssl/blob/8658feddea6aef5cf5cbb1cfbf6b1817fa432051/ssl/statem/extensions.c#L1597-L1605

We do so using EVP_DigestSignInit_ex() and pass the name of the digest to be used. In the test I was looking at that failed it was SHA384. Because we're actually creating a MAC and there are no "signature" implementations in the providers for MACs, we go down the legacy path and end up in crypto/evp/pkey_mac.c. This does the work of converting the call to EVP_MAC - and so eventually we will end up in a provider, via a legacy EVP_PKEY_METHOD.

Without this line type is NULL (because we actually called EVP_DigestSignInit_ex() and provided an mdname). This then resulted in us using the default digest for HMAC which is SHA256. Thus everything fell over because the MAC we generated wasn't the expected size.

mattcaswell:
I've moved the OPENSSL_CTX param to not be the first one in the list as suggested by @t8m.

Please take another look.

doc/man3/EVP_DigestVerifyInit.pod
```diff
@@ -9,9 +9,10 @@ EVP_DigestVerifyFinal, EVP_DigestVerify - EVP signature verification functions
 
  #include <openssl/evp.h>
 
- int EVP_DigestVerifyInit_ex(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
-                             const char *mdname, const char *props,
-                             EVP_PKEY *pkey, EVP_SIGNATURE *signature);
+ int EVP_DigestVerifyInit_ex(OPENSSL_CTX *libctx, EVP_MD_CTX *ctx,
```
slontis: these have to fixed too right?

mattcaswell: Oops. Fixed. Also updated the rest of the params because they were out of date and adjusted the text to match.

mattcaswell:
It turns out we need to use `evp_get_digestbyname_ex` rather than `EVP_get_digestbyname` in this PR. Otherwise we may not find the digest if the name being used is an alias that is unknown in the default libctx. Fixup pushed.

@t8m can you reconfirm?

t8m:
My approval holds.

mattcaswell:
Pushed. Thanks.

PR #11356, opened 2020-03-18
Use fetched cipher/digest for tickets

mattcaswell:
Make sure we use explicitly fetched cipher and digest when handling libssl tickets.

### DIFF
--- a/crypto/err/openssl.txt
+++ b/crypto/err/openssl.txt
@@ -3088,6 +3088,7 @@ SSL_R_EXTENSION_NOT_RECEIVED:279:extension not received
 SSL_R_EXTRA_DATA_IN_MESSAGE:153:extra data in message
 SSL_R_EXT_LENGTH_MISMATCH:163:ext length mismatch
 SSL_R_FAILED_TO_INIT_ASYNC:405:failed to init async
+SSL_R_ALGORITHM_FETCH_FAILED:295:algorithm fetch failed
 SSL_R_FRAGMENTED_CLIENT_HELLO:401:fragmented client hello
 SSL_R_GOT_A_FIN_BEFORE_A_CCS:154:got a fin before a ccs
 SSL_R_HTTPS_PROXY_REQUEST:155:https proxy request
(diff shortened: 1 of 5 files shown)


crypto/err/openssl.txt
```diff
@@ -3088,6 +3088,7 @@ SSL_R_EXTENSION_NOT_RECEIVED:279:extension not received
 SSL_R_EXTRA_DATA_IN_MESSAGE:153:extra data in message
 SSL_R_EXT_LENGTH_MISMATCH:163:ext length mismatch
 SSL_R_FAILED_TO_INIT_ASYNC:405:failed to init async
+SSL_R_FETCH_FAILED:295:fetch failed
```
t8m: Could the error be a little more descriptive? Something like "algorithm fetch failed" or "evp algorithm fetch failed".

ssl/statem/statem_srvr.c
```diff
@@ -3906,7 +3906,14 @@ static int construct_stateless_ticket(SSL *s, WPACKET *pkt, uint32_t age_add,
         }
         iv_len = EVP_CIPHER_CTX_iv_length(ctx);
     } else {
-        const EVP_CIPHER *cipher = EVP_aes_256_cbc();
+        EVP_CIPHER *cipher = EVP_CIPHER_fetch(s->ctx->libctx, "AES-256-CBC",
+                                              s->ctx->propq);
+
+        if (cipher == NULL) {
+            SSLfatal(s, SSL_AD_INTERNAL_ERROR, SSL_F_CONSTRUCT_STATELESS_TICKET,
+                     ERR_R_INTERNAL_ERROR);
```
t8m: And it could be used here as well?

crypto/err/openssl.txt
mattcaswell: I made it "algorithm fetch failed"

ssl/statem/statem_srvr.c
mattcaswell: Done

mattcaswell:
Please take another look

t8m:
Actually the Travis failure - a leak seems to be relevant.

t8m:
There is one return where the fetched sha256 is not freed.

mattcaswell:
Fixup pushed addressing the memory leak.

@t8m - please take another look.

mattcaswell:
Pushed. Thanks.

Matt

PR #11358, opened 2020-03-18
EVP & TLS: Add necessary EC_KEY data extraction functions, and use them

levitte:
libssl code uses EVP_PKEY_get0_EC_KEY() to extract certain basic data
from the EC_KEY.  We replace that with EVP_PKEY functions.

This may or may not be refactored later on.

### DIFF
--- a/crypto/evp/p_lib.c
+++ b/crypto/evp/p_lib.c
@@ -755,12 +755,16 @@ int EVP_PKEY_is_a(const EVP_PKEY *pkey, const char *name)
          */
         int type;
 
-        if (strcasecmp(name, "EC") == 0)
-            type = EVP_PKEY_EC;
-        else if (strcasecmp(name, "RSA") == 0)
+        if (strcasecmp(name, "RSA") == 0)
             type = EVP_PKEY_RSA;
+#ifndef OPENSSL_NO_EC
+        else if (strcasecmp(name, "EC") == 0)
+            type = EVP_PKEY_EC;
+#endif
+#ifndef OPENSSL_NO_DSA
         else if (strcasecmp(name, "DSA") == 0)
             type = EVP_PKEY_DSA;
+#endif
         else
             type = EVP_PKEY_type(OBJ_sn2nid(name));
         return EVP_PKEY_type(pkey->type) == type;
--- a/crypto/evp/p_lib.c
+++ b/crypto/evp/p_lib.c
@@ -33,8 +33,11 @@
 
 #include "crypto/asn1.h"
 #include "crypto/evp.h"
+#include "internal/evp.h"
 #include "internal/provider.h"
 #include "evp_local.h"
+
+#include "crypto/ec.h"
 
 static int pkey_set_type(EVP_PKEY *pkey, ENGINE *e, int type, const char *str,
                          int len, EVP_KEYMGMT *keymgmt);
@@ -779,6 +782,48 @@ int EVP_PKEY_can_sign(const EVP_PKEY *pkey)
     }
     return 0;
 }
+
+#ifndef OPENSSL_NO_EC
+/*
+ * TODO rewrite when we have proper data extraction functions
+ * Note: an octet pointer would be desirable!
+ */
+static OSSL_CALLBACK get_ec_curve_name_cb;
+static int get_ec_curve_name_cb(const OSSL_PARAM params[], void *arg)
+{
+    const OSSL_PARAM *p = NULL;
+
+    if ((p = OSSL_PARAM_locate_const(params, OSSL_PKEY_PARAM_EC_NAME)) != NULL)
+        return OSSL_PARAM_get_utf8_string(p, arg, 0);
+
+    /* If there is no curve name, this is not an EC key */
+    return 0;
+}
+
+int evp_pkey_get_EC_KEY_curve_nid(const EVP_PKEY *pkey)
+{
+    int ret = NID_undef;
+
+    if (pkey->keymgmt == NULL) {
+        if (EVP_PKEY_base_id(pkey) == EVP_PKEY_EC) {
+            EC_KEY *ec = EVP_PKEY_get0_EC_KEY(pkey);
+
+            ret = EC_GROUP_get_curve_name(EC_KEY_get0_group(ec));
+        }
+    } else if (EVP_PKEY_is_a(pkey, "EC") || EVP_PKEY_is_a(pkey, "SM2")) {
+        char *curve_name = NULL;
+
+        ret = evp_keymgmt_export(pkey->keymgmt, pkey->keydata,
+                                 OSSL_KEYMGMT_SELECT_DOMAIN_PARAMETERS,
+                                 get_ec_curve_name_cb, &curve_name);
+        if (ret)
+            ret = ec_curve_name2nid(curve_name);
+        OPENSSL_free(curve_name);
+    }
+
+    return ret;
+}
+#endif
 
 static int print_reset_indent(BIO **out, int pop_f_prefix, long saved_indent)
 {
--- a/crypto/evp/p_lib.c
+++ b/crypto/evp/p_lib.c
@@ -39,6 +39,9 @@
 
 #include "crypto/ec.h"
 
+/* TODO remove this when the EVP_PKEY_is_a() #legacy support hack is removed */
+#include "e_os.h"                /* strcasecmp on Windows */
+
 static int pkey_set_type(EVP_PKEY *pkey, ENGINE *e, int type, const char *str,
                          int len, EVP_KEYMGMT *keymgmt);
 static void evp_pkey_free_it(EVP_PKEY *key);
@@ -739,8 +742,29 @@ int EVP_PKEY_base_id(const EVP_PKEY *pkey)
 int EVP_PKEY_is_a(const EVP_PKEY *pkey, const char *name)
 {
 #ifndef FIPS_MODE
-    if (pkey->keymgmt == NULL)
-        return EVP_PKEY_type(pkey->type) == EVP_PKEY_type(OBJ_sn2nid(name));
+    if (pkey->keymgmt == NULL) {
+        /*
+         * These hard coded cases are pure hackery to get around the fact
+         * that names in crypto/objects/objects.txt are a mess.  There is
+         * no "EC", and "RSA" leads to the NID for 2.5.8.1.1, an OID that's
+         * fallen out in favor of { pkcs-1 1 }, i.e. 1.2.840.113549.1.1.1,
+         * the NID of which is used for EVP_PKEY_RSA.  Strangely enough,
+         * "DSA" is accurate...  but still, better be safe and hard-code
+         * names that we know.
+         * TODO Clean this away along with all other #legacy support.
+         */
+        int type;
+
+        if (strcasecmp(name, "EC") == 0)
+            type = EVP_PKEY_EC;
+        else if (strcasecmp(name, "RSA") == 0)
+            type = EVP_PKEY_RSA;
+        else if (strcasecmp(name, "DSA") == 0)
+            type = EVP_PKEY_DSA;
+        else
+            type = EVP_PKEY_type(OBJ_sn2nid(name));
+        return EVP_PKEY_type(pkey->type) == type;
+    }
 #endif
     return EVP_KEYMGMT_is_a(pkey->keymgmt, name);
 }
--- a/crypto/evp/p_lib.c
+++ b/crypto/evp/p_lib.c
@@ -24,6 +24,7 @@
 #include <openssl/rsa.h>
 #include <openssl/dsa.h>
 #include <openssl/dh.h>
+#include <openssl/ec.h>
 #include <openssl/cmac.h>
 #include <openssl/engine.h>
 #include <openssl/params.h>
@@ -732,6 +733,52 @@ int EVP_PKEY_base_id(const EVP_PKEY *pkey)
     return EVP_PKEY_type(pkey->type);
 }
 
+int EVP_PKEY_is_a(const EVP_PKEY *pkey, const char *name)
+{
+#ifndef FIPS_MODE
+    if (pkey->keymgmt == NULL)
+        return EVP_PKEY_type(pkey->type) == EVP_PKEY_type(OBJ_sn2nid(name));
+#endif
+    return EVP_KEYMGMT_is_a(pkey->keymgmt, name);
+}
+
+int EVP_PKEY_can_sign(const EVP_PKEY *pkey)
+{
+    if (pkey->keymgmt == NULL) {
+        switch (EVP_PKEY_base_id(pkey)) {
+        case EVP_PKEY_RSA:
+            return 1;
+#ifndef OPENSSL_NO_DSA
+        case EVP_PKEY_DSA:
+            return 1;
+#endif
+#ifndef OPENSSL_NO_EC
+        case EVP_PKEY_ED25519:
+        case EVP_PKEY_ED448:
+            return 1;
+        case EVP_PKEY_EC:        /* Including SM2 */
+            return EC_KEY_can_sign(pkey->pkey.ec);
+#endif
+        default:
+            break;
+        }
+    } else {
+        const OSSL_PROVIDER *prov = EVP_KEYMGMT_provider(pkey->keymgmt);
+        OPENSSL_CTX *libctx = ossl_provider_library_context(prov);
+        const char *supported_sig =
+            pkey->keymgmt->query_operation_name != NULL
+            ? pkey->keymgmt->query_operation_name(OSSL_OP_SIGNATURE)
+            : evp_first_name(prov, pkey->keymgmt->name_id);
+        EVP_SIGNATURE *signature = NULL;
+
+        signature = EVP_SIGNATURE_fetch(libctx, supported_sig, NULL);
+        if (signature != NULL) {
+            EVP_SIGNATURE_free(signature);
+            return 1;
+        }
+    }
+    return 0;
+}
 
 static int print_reset_indent(BIO **out, int pop_f_prefix, long saved_indent)
 {
--- a/doc/man3/EVP_PKEY_is_a.pod
+++ b/doc/man3/EVP_PKEY_is_a.pod
@@ -1,0 +1,72 @@
+=pod
+
+=head1 NAME
+
+EVP_PKEY_is_a, EVP_PKEY_can_sign
+- key type and capabilities functions
+
+=head1 SYNOPSIS
+
+ #include <openssl/evp.h>
+
+ int EVP_PKEY_is_a(const EVP_PKEY *pkey, const char *name);
+ int EVP_PKEY_can_sign(const EVP_PKEY *pkey);
+
+=head1 DESCRIPTION
+
+EVP_PKEY_is_a() checks if the key type of I<pkey> is I<name>.
+
+EVP_PKEY_can_sign() checks if the functionality for the key type of
+I<pkey> supports signing.  No other check is done, such as whether
+I<pkey> contains a private key.
+
+=head1 RETURN VALUES
+
+EVP_PKEY_is_a() returns 1 if I<pkey> has the key type I<name>,
+otherwise 0.
+
+EVP_PKEY_can_sign() returns 1 if the I<pkey> key type functionality
+supports signing, otherwise 0.
+
+=head1 EXAMPLES
+
+=head2 EVP_PKEY_is_a()
+
+The loaded providers and what key types they support will ultimately
+determine what I<name> is possible to use with EVP_PKEY_is_a().  We do know
+that the default provider supports RSA, DH, DSA and EC keys, so we can use
+this as an crude example:
+
+ #include <openssl/evp.h>
+
+ ...
+     /* |pkey| is an EVP_PKEY* */
+     if (EVP_PKEY_is_a(pkey, "RSA")) {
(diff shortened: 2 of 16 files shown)


levitte:
If nothing else, this can be an interim solution until #11340 has been merged.
#11328 depends on something like this.

mattcaswell:
Hmmm. I can't say I'm keen on this. This seems to just sidestep the problem of what to do with the `EVP_PKEY_get0_EC_KEY` et al functions - and fixes it *for us*. But does so by polluting the EVP_PKEY API with algorithm specific functions.

levitte:
Yes, that's true.  However, as soon as we get EC key generation, we get issues with this, so the question is how long that should be stuck waiting for "the right thing" to happen.

I *could* make those functions internal but still exported, if that helps matters

levitte:
This *is* an interim solution

levitte:
Marked as WIP for obvious reasons

mattcaswell:
> I could make those functions internal but still exported, if that helps matters

That would definitely help I think.

levitte:
Although, I think `EVP_PKEY_can_sign()` is more generally useful...

levitte:
There.  I've also added EVP_PKEY_is_a(), which has been floating a more than one branch I juggled with

doc/man3/EVP_PKEY_is_a.pod
```diff
@@ -0,0 +1,42 @@
...
+=head1 SYNOPSIS
+
+ #include <openssl/evp.h>
+
+ int EVP_PKEY_is_a(const EVP_PKEY *pkey, const char *name);
+ int EVP_PKEY_can_sign(const EVP_PKEY *pkey);
+
+=head1 DESCRIPTION
+
+EVP_PKEY_is_a() checks if the key type of I<pkey> is I<name>.
```
mattcaswell: I think some examples of valid names would be useful.

ssl/statem/statem_lib.c
```diff
@@ -1559,10 +1558,9 @@ static int is_tls13_capable(const SSL *s)
          * more restrictive so check that our sig algs are consistent with this
          * EC cert. See section 4.2.3 of RFC8446.
          */
-        eckey = EVP_PKEY_get0_EC_KEY(s->cert->pkeys[SSL_PKEY_ECC].privatekey);
-        if (eckey == NULL)
-            continue;
-        curve = EC_GROUP_get_curve_name(EC_KEY_get0_group(eckey));
+        curve = EVP_PKEY_get_EC_KEY_curve_nid(s->cert->pkeys[SSL_PKEY_ECC]
+                                              .privatekey);
+        fprintf(stderr, "DEBUG[%s] ECC curve = %d\n", __func__, curve);
```
mattcaswell: Err...fprintf?

ssl/statem/statem_lib.c
```diff
@@ -1559,10 +1558,9 @@ static int is_tls13_capable(const SSL *s)
          * more restrictive so check that our sig algs are consistent with this
          * EC cert. See section 4.2.3 of RFC8446.
          */
-        eckey = EVP_PKEY_get0_EC_KEY(s->cert->pkeys[SSL_PKEY_ECC].privatekey);
-        if (eckey == NULL)
-            continue;
-        curve = EC_GROUP_get_curve_name(EC_KEY_get0_group(eckey));
+        curve = EVP_PKEY_get_EC_KEY_curve_nid(s->cert->pkeys[SSL_PKEY_ECC]
```
mattcaswell: Shouldn't this be `evp_pkey_get_EC_KEY_curve_nid` (i.e. in lowercase)?

ssl/t1_lib.c
```diff
@@ -577,19 +577,15 @@ void tls1_get_formatlist(SSL *s, const unsigned char **pformats,
...
 
     /* If not an EC key nothing to check */
     if (EVP_PKEY_id(pkey) != EVP_PKEY_EC)
         return 1;
-    ec = EVP_PKEY_get0_EC_KEY(pkey);
-    grp = EC_KEY_get0_group(ec);
 
     /* Get required compression id */
-    if (EC_KEY_get_conv_form(ec) == POINT_CONVERSION_UNCOMPRESSED) {
+    if (EVP_PKEY_get_EC_KEY_conv_form(pkey) == POINT_CONVERSION_UNCOMPRESSED) {
```
mattcaswell: evp_pkey_...and so on for other occurrences

include/internal/evp.h
```diff
@@ -0,0 +1,18 @@
...
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
+ */
+
+#ifndef OSSL_INTERNAL_DSO_H
+# define OSSL_INTERNAL_DSO_H
+
+# include <openssl/evp.h>
+
+int evp_pkey_get_EC_KEY_curve_nid(const EVP_PKEY *pkey);
```
mattcaswell: Should there be some TODO comment somewhere saying this is an interim solution?

mattcaswell (changes requested):

include/internal/evp.h
```diff
@@ -0,0 +1,18 @@
...
+ * Copyright 2020 The OpenSSL Project Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License 2.0 (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
+ */
+
+#ifndef OSSL_INTERNAL_DSO_H
+# define OSSL_INTERNAL_DSO_H
```
mattcaswell: Should be `OSSL_INTERNAL_EVP_H`

ssl/statem/statem_lib.c
levitte: Oops, debug stuff.  Will remove

ssl/statem/statem_lib.c
levitte: Yes.  There are a few of these, I've already corrected them locally

include/internal/evp.h
levitte: Yup, already fixed locally

include/internal/evp.h
levitte: I thought I already added such a comment???  probably somewhere else, and yeah I can duplicate it here

doc/man3/EVP_PKEY_is_a.pod
levitte: Sure

mattcaswell:
Could you push the latest version of your local branch here? I'm experimenting with your ec key gen stuff...

levitte:
Done, and in #11328 as well

levitte:
Rebased on master, now that #11375 is in.

ssl/statem/statem_lib.c
levitte: Removed them all

include/internal/evp.h
levitte: Done

doc/man3/EVP_PKEY_is_a.pod
levitte: Done

levitte:
Ping!

levitte:
Travis failure re documentation:

```
doc/man3/EVP_PKEY_is_a.pod:1: EXAMPLE not EXAMPLES section.
```

That document contains *two* examples, so I'm starting to wonder if find-doc-nits isn't a little *too* picky in this case.

levitte:
Ping!

mattcaswell:
Needs rebase!

t8m:
There are now conflicts and the Travis failure is probably relevant

levitte:
Just rebased

ssl/t1_lib.c
```diff
@@ -577,19 +578,15 @@ void tls1_get_formatlist(SSL *s, const unsigned char **pformats,
 /* Check a key is compatible with compression extension */
 static int tls1_check_pkey_comp(SSL *s, EVP_PKEY *pkey)
```
mattcaswell: This function is inextricably linked with the point encoding/decoding. When we push the point encoding/decoding provider side then we will also have to push the logic about what point compression is supported provider side too, i.e. we should ask the provider what point compression is supported for any particular key - rather than trying to infer if from the field type. That works for own stuff because we know we support point compression - but we can't necessarily assume the same for provider owned keys. For now though we are explicitly downgrading EC keys prior to point encoding/decoding in order to use the ameth implementation. It seems we could do the same trick here and then just do it the old way - until such time as the whole thing goes provider side.

doc/man3/EVP_PKEY_is_a.pod
```diff
@@ -0,0 +1,74 @@
...
+
+EVP_PKEY_can_sign() checks if I<pkey> can be used for signing.  The
```
mattcaswell: I assume this returns false if the key can only be used for verifying? e.g. because it is a public key? Could we make that more explicit?

doc/man3/EVP_PKEY_set1_RSA.pod
```diff
@@ -51,52 +51,61 @@ EVP_PKEY_set1_engine, EVP_PKEY_get0_engine - EVP_PKEY assignment functions
...
 
 EVP_PKEY_set_alias_type() allows modifying a EVP_PKEY to use a
 different set of algorithms than the default.
 
+=head1 WARNINGS
+
+The following functions are only reliable with B<EVP_PKEY>s that have
+been assigned an internal key with EVP_PKEY_assign_*():
+
+EVP_PKEY_id(), EVP_PKEY_base_id(), EVP_PKEY_type(), EVP_PKEY_set_alias_type()
```
mattcaswell: Should we attempt a downgrade if someone calls these functions?

mattcaswell (changes requested):

crypto/evp/p_lib.c
```diff
@@ -732,6 +736,110 @@ int EVP_PKEY_base_id(const EVP_PKEY *pkey)
...
+        const char *supported_sig = NULL;
+        EVP_SIGNATURE *signature = NULL;
+
+        supported_sig =
+            pkey->keymgmt->query_operation_name != NULL
+            ? pkey->keymgmt->query_operation_name(OSSL_OP_SIGNATURE)
+            : evp_first_name(prov, pkey->keymgmt->name_id);
+        signature = EVP_SIGNATURE_fetch(libctx, supported_sig, NULL);
+        EVP_SIGNATURE_free(signature);
+        return signature != NULL;
```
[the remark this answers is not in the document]
levitte: A pointer is just a value just like any other value.  That value won't "magically" change because you freed the memory area its value points at.  The value in itself is perfectly usable as long as it's not dereferenced.

But, I can waste another variable to hold the truth value if you prefer...

crypto/evp/p_lib.c
```diff
@@ -732,6 +736,110 @@ int EVP_PKEY_base_id(const EVP_PKEY *pkey)
...
+        const OSSL_PROVIDER *prov = EVP_KEYMGMT_provider(pkey->keymgmt);
+        OPENSSL_CTX *libctx = ossl_provider_library_context(prov);
+        const char *supported_sig = NULL;
+        EVP_SIGNATURE *signature = NULL;
+
+        supported_sig =
+            pkey->keymgmt->query_operation_name != NULL
+            ? pkey->keymgmt->query_operation_name(OSSL_OP_SIGNATURE)
+            : evp_first_name(prov, pkey->keymgmt->name_id);
+        signature = EVP_SIGNATURE_fetch(libctx, supported_sig, NULL);
```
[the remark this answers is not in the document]
levitte: Looking at the legacy code, it seems we assume that method that supports signing also supports verifying.  Looking at what `EV_KEY_can_sign()` does, I found these indicators:

https://github.com/openssl/openssl/blob/20c98cd45399423f760dbd75d8912769c6b7b10e/crypto/ec/ec_local.h#L31-L32

https://github.com/openssl/openssl/blob/20c98cd45399423f760dbd75d8912769c6b7b10e/include/openssl/evp.h#L1402-L1404

crypto/evp/p_lib.c
```diff
@@ -732,6 +736,110 @@ int EVP_PKEY_base_id(const EVP_PKEY *pkey)
...
+            return 1;
+#ifndef OPENSSL_NO_DSA
+        case EVP_PKEY_DSA:
+            return 1;
+#endif
+#ifndef OPENSSL_NO_EC
+        case EVP_PKEY_ED25519:
+        case EVP_PKEY_ED448:
+        case EVP_PKEY_EC:        /* Including SM2 */
+            return EC_KEY_can_sign(pkey->pkey.ec);
```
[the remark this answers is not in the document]
levitte: Okie

crypto/evp/p_lib.c
```diff
@@ -732,6 +736,110 @@ int EVP_PKEY_base_id(const EVP_PKEY *pkey)
...
+            ? pkey->keymgmt->query_operation_name(OSSL_OP_SIGNATURE)
+            : evp_first_name(prov, pkey->keymgmt->name_id);
+        signature = EVP_SIGNATURE_fetch(libctx, supported_sig, NULL);
+        EVP_SIGNATURE_free(signature);
+        return signature != NULL;
+    }
+}
+
+#ifndef OPENSSL_NO_EC
+/* TODO rewrite when we have proper data extraction functions */
```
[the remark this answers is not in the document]
levitte: I wasn't ready to assume, so...

(I assume, BTW, that we're going to look at *all* TODO comments before final release, and that the presence of "(3.0)" only indicates if it's a "MUST do" or a "MAY do")

doc/man3/EVP_PKEY_is_a.pod
levitte: Ah, so having it verify that `evp_keymgmt_has(p->keymgmt, p->keydata, OSSL_KEYMGMT_SELECT_PRIVATE_KEY)` is true?  Yeahsure

doc/man3/EVP_PKEY_set1_RSA.pod
levitte: I'm unsure...  but as a side thought, I've been thinking that keys with a provider side "origin" should have the type `EVP_PKEY_KEYMGMT` rather than `EVP_PKEY_NONE`...

crypto/evp/p_lib.c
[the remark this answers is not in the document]
richsalz: FWIW, I agree that looking at the pointer after it's free'd, while technically legal, is confusing.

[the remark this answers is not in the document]
kaduk: I think that *technically* it's not actually legal.
It helps to recall that the C rules for pointers have their origins in the era of segmented architectures, where even trying to load a pointer value (not dereference it) for which the corresponding segment was not valid anymore, could fault.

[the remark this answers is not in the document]
levitte: Huh!  I didn't know that.

crypto/evp/p_lib.c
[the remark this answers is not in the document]
levitte: Should we care in the legacy section of this function?  How do we check if there's a private key there?

crypto/evp/p_lib.c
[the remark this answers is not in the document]
t8m: You could nicely do 
```
if (signature != NULL) {
    EVP_SIGNATURE_free(signature); 
    return 1; 
} return 0;
```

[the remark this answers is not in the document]
richsalz: I believe Ben is wrong; "free(p); look at p but not *p" is legal.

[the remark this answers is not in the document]
kaduk: See n1256.pdf appendix J.2, stanzas 8 and 9:

```
— An object is referred to outside of its lifetime (6.2.4).
— The value of a pointer to an object whose lifetime has ended is used (6.2.4).
```
The first one refers to dereferencing the pointer, and the second refers to only the value of the (undereferenced) pointer.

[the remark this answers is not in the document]
richsalz: I stand corrected, thanks.

ssl/t1_lib.c
levitte: Okie, done

doc/man3/EVP_PKEY_set1_RSA.pod
levitte: I'm leaning on no.  We risk too much uncontrolled downgrading.

crypto/evp/p_lib.c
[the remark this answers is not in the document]
levitte: The question here remains unanswered.  @mattcaswell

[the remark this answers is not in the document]
mattcaswell: It seems all this stems from confusion in my own head about what EVP_PKEY_can_sign() is supposed to do. Is it supposed to report whether the parameters associated with this key are capable of doing signing operations, or whether this specific key can do signing (i.e. it has a private key *and* it has suitable parameters). I was assuming the latter - but if this is supposed to be a drop-in replacement for EC_KEY_can_sign() then that seems to do the former - and therefore don't care whether this has a private key or not.

doc/man3/EVP_PKEY_is_a.pod
mattcaswell: See my comment above...

ssl/t1_lib.c
```diff
@@ -577,19 +578,26 @@ void tls1_get_formatlist(SSL *s, const unsigned char **pformats,
...
+    /*
+     * TODO(3.0) Remove this when we have routines that can extract the
+     * compression information from a provider side key.  We know that
+     * EVP_PKEY_get0() downgrades an EVP_PKEY to contain a legacy key.
+     *
+     * THIS IS TEMPORARY
+     */
+    EVP_PKEY_get0(pkey);
+    if (EVP_PKEY_id(pkey) == EVP_PKEY_NONE)
+        return 0;
```
mattcaswell: Now that we have downgraded we can keep all the original code below. The `EVP_PKEY_get0_EC_KEY()` call will just work and we don't need `evp_pkey_get_EC_KEY_conv_form` or `evp_pkey_get_EC_KEY_field_type`.

mattcaswell (changes requested):

crypto/evp/p_lib.c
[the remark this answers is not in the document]
levitte: I've designed it as a drop in replacement for `EC_KEY_can_sign()`.  It's possible that we want to make it stricter before releasing 3.0, though...  but maybe not immediately, as we seem to get ourselves into another philosophical rabbit hole.

ssl/t1_lib.c
levitte: At *some* point, we will need to get those things from provider side keys, so I'm apprehensive at removing those functions...  do you anticipate that the *final solution* will look substantially different?

crypto/evp/p_lib.c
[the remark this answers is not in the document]
levitte: Fixed

ssl/t1_lib.c
levitte: Ping @mattcaswell

mattcaswell: Yes. This whole dance with these calls is just to figure out what TLS compression id we need to use. I anticipate asking the provider directly for this compression id rather than inferring it from these other calls.

kaduk: (I assume "TLS compression id" means "EC point format", not "we're using something other than NULL compression".)

mattcaswell: Yes...the TLS EC point format. Sorry for the confusion.

levitte: In that case, I might as well restore this whole function to its original form, and just replace the EVP_PKEY_id() check with a EVP_PKEY_is_a() check.

levitte: Done

levitte:
Ah, the CI failure was due to a merge error with duplicate definition of the same local variable.
Fixed, and I took the opportunity to squash and rebase.

mattcaswell (approved):

mattcaswell:
Approved assuming CIs pass.

levitte:
... it only takes a rebase for hell to break loose :unamused:

levitte:
With @mattcaswell's assistance, I was able to put in a workable hack.  It turned out that legacy key support needed some hard coding.

mattcaswell (approved):

levitte:
One failure left...  I didn't account for `no-ec`

levitte:
... at least not everywhere

levitte:
Fixed

mattcaswell (approved):
Reconfirm assuming CIs are ok

levitte:
Apart from the usual timeout, CIs are fine

levitte:
Merged.

4f76d62f2384d3335bd1d043706995ae64b37348 EVP: add EVP_PKEY_is_a() and EVP_PKEY_can_sign()
c2041da8c15027ddde5afcf9809d8d3a975eb25b EVP & TLS: Add necessary EC_KEY data extraction functions, and use them
e3be0f4389fe1680e1fd547a8575ea71b4518b57 Fix export of provided EC keys
afce590b74159f7df1452fb2c4aa990a52536c38 TLS: Temporarly downgrade newly generated EVP_PKEYs to legacy

levitte:
Oh damn, *now* I see that the last approval came after the "approval: done" label was set, so I cut the grace period short...  by 3hrs, I hope it's not too bad.

PR #11359, opened 2020-03-18
Allow to continue with further checks on UNABLE_TO_VERIFY_LEAF_SIGNATURE

### LINKED ISSUE #11297: Continue on UNABLE_TO_VERIFY_LEAF_SIGNATURE
We like to request that it is possible (maybe by an explicit option) to continue certificate verification even if UNABLE_TO_VERIFY_LEAF_SIGNATURE is thrown.

We are implementing the official eID-Client of the German identity card. The whole system is defined by a technical guideline of the Federal Office for Information Security. [1]

The identity card uses NFC to establish a connection between server and the card itself. These connections are also secured by TLS. It is not possible to use root-CAs from system truststore as the German eID infrastructure uses it's own trust anchor. This trust anchor is a CVC (Card Verifiable Certificate) which is verified by the card itself. The CVC contains hash values of the allowed TLS certificates.

When we establish a secured connection via TLS we need to check certain things like certificate validity. But as we don't have a root CA we need to ignore "UNABLE_TO_VERIFY_LEAF_SIGNATURE" error. Even though the content of the certificate MIGHT be compromised we MUST do plausibility checks (like validity) because we can approve it LATER with our received CVC. However, openssl stops the whole validation. This breaks our workflow as it would allow that an expired certificate is not fatal.

We checked mbedtls and gnutls. Those implementations throw a similar error but don't stop the whole verification.

It is not possible to enforce all eID service providers to send the WHOLE certificate chain including the root-CA. See RFC5246 section-7.4.2.
```
   certificate_list
      This is a sequence (chain) of certificates.  The sender's
      certificate MUST come first in the list.  Each following
      certificate MUST directly certify the one preceding it.  Because
      certificate validation requires that root keys be distributed
      independently, the self-signed certificate that specifies the root
      certificate authority MAY be omitted from the chain, under the
      assumption that the remote end must already possess it in order to
      validate it in any case.
```

[1] https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Publikationen/TechnischeRichtlinien/TR03124/TR-03124-1.pdf?__blob=publicationFile&v=2

[2] https://tools.ietf.org/html/rfc5246#section-7.4.2

misery:
This unifies the behaviour of a single certificate with
an unknown CA certificate with a self-signed certificate.
The user callback can mask that error to retrieve additional
error information. So the user application can decide to
abort the connection instead to be forced by openssl.

This change in behaviour is backward compatible as user callbacks
who don't want to ignore UNABLE_TO_VERIFY_LEAF_SIGNATURE will
still abort the connection by default.

CLA: trivial
Fixes #11297

### DIFF
--- a/crypto/x509/x509_vfy.c
+++ b/crypto/x509/x509_vfy.c
@@ -1750,9 +1750,15 @@ static int internal_verify(X509_STORE_CTX *ctx)
             xs = xi;
             goto check_cert;
         }
-        if (n <= 0)
-            return verify_cb_cert(ctx, xi, 0,
-                                  X509_V_ERR_UNABLE_TO_VERIFY_LEAF_SIGNATURE);
+        if (n <= 0) {
+            if (!verify_cb_cert(ctx, xi, 0,
+                                X509_V_ERR_UNABLE_TO_VERIFY_LEAF_SIGNATURE))
+                return 0;
+
+            xs = xi;
+            goto check_cert;
+        }
+
         n--;
         ctx->error_depth = n;
         xs = sk_X509_value(ctx->chain, n);


misery:
Ping? Opinions?

misery:
ping?

DDvO (approved):
Nice improvement

misery (commented):
By the way... If n is 0 or lower. Is n-- still valid?

mspncp:
@misery I guess I completely misunderstood you question on first read, sorry. Indeed, what happens if `sk_X509_value()` is called with a negative value @DDvO?

https://github.com/openssl/openssl/blob/ad3bb58e90bd49e2f42021637bc5354082319377/crypto/x509/x509_vfy.c#L1724-L1726

misery:
> @misery I guess I completely misunderstood you question on first read, sorry. Indeed, what happens if `sk_X509_value()` is called with a negative value @DDvO?
> 
> https://github.com/openssl/openssl/blob/ad3bb58e90bd49e2f42021637bc5354082319377/crypto/x509/x509_vfy.c#L1724-L1726

Thanks for feedback! Maybe I need to set n to 0 after callback? Don't know openssl internal enough for that.

mspncp:
Please also replace the line
```
GH: #11297
```
by 
```
Fixes #11297
```
in your commit message.

mspncp:
> Thanks for feedback! Maybe I need to set n to 0 after callback? Don't know openssl internal enough for that.

I haven't actually been following the discussion #11297 and did not go into the details of the verification code. So I can't tell you offhand whether that  would be the correct fix. Maybe @DDvO can say more about it.

DDvO:
In this case (i.e., when the callback returns `1` on a chain that consists of just one cert) 
it does not hurt that `n` can become negative (or to be more specific, will be `-1`).

The `ctx->error_depth` becomes `-1` but this should be irrelevant because `internal_verify()` returns success.
The variable `xs` will be assigned to NULL due to `sk_X509_value(ctx->chain, n)` but is not used thereafter.
Yet note that due to the following `while (n >= 0)` the function will not do any further checks,
and after reading #11297 I found that this is not what you wanted.
I need to withdraw my quick approval because your PR so far essentially has no effect.

crypto/x509/x509_vfy.c
```diff
@@ -1718,8 +1718,9 @@ static int internal_verify(X509_STORE_CTX *ctx)
             goto check_cert;
         }
         if (n <= 0)
-            return verify_cb_cert(ctx, xi, 0,
-                                  X509_V_ERR_UNABLE_TO_VERIFY_LEAF_SIGNATURE);
+            if (!verify_cb_cert(ctx, xi, 0,
+                                X509_V_ERR_UNABLE_TO_VERIFY_LEAF_SIGNATURE))
+                return 0;
```
DDvO, line 1723: What you really need is
```
if (verify_cb_cert(ctx, xi, 0,
                   X509_V_ERR_UNABLE_TO_VERIFY_LEAF_SIGNATURE) {
    xs = xi;
    goto check_cert_time;
} else {
    return 0;
}

**Review by DDvO — CHANGES_REQUESTED** [review]
So far your PR has essentially no effect.
I've suggested a real solution.

**Comment by misery** [comment]
Well, why no effect? It works with my tests. Maybe it must be adjusted because of n.

		if (n <= 0) {
			if (!verify_cb_cert(ctx, xi, 0,
				X509_V_ERR_UNABLE_TO_VERIFY_LEAF_SIGNATURE))
					return 0;
			n = 0;
		} else
			n--;

**Comment by misery** [comment]
Thanks for your review!

**Comment by richsalz** [comment]
Get rid of the "else after goto"  And consider inverting the test:
```
if (!verify_cb_cert(ctx, xi, 0,
                   X509_V_ERR_UNABLE_TO_VERIFY_LEAF_SIGNATURE)
    return 0;
xs = xi;
goto check_cert_time;
```

**Comment by DDvO** [comment]
> Well, why no effect? It works with my tests.

Strange that it worked for you - if `n <= 0` and the callback says 'yes' then `internal_verify()` does not do any further checks.

> Maybe it must be adjusted because of n.
> 
> ```
> 	if (n <= 0) {
> 		if (!verify_cb_cert(ctx, xi, 0,
> 			X509_V_ERR_UNABLE_TO_VERIFY_LEAF_SIGNATURE))
> 				return 0;
> 		n = 0;
> 	} else
> 		n--;
> ```

This solution would have the same effect as mine because also with it `x509_check_cert_time()` is called while the self-signature check is certainly not done because in the given case the cert in question (`xs`, which is the same as `xi`) is not self-issued and thus cannot be self-signed.

**Comment by DDvO** [comment]
Please handle this Travis CI failure (due to a warning):
```
crypto/x509/x509_vfy.c: In function 'internal_verify':

crypto/x509/x509_vfy.c:1753:12: error: suggest explicit braces to avoid ambiguous 'else' [-Werror=parentheses]

         if (n <= 0)

            ^

crypto/x509/x509_vfy.c:1758:17: error: label 'check_cert_time' used but not defined

                 goto check_cert_time;

                 ^
```

**Comment by misery** [comment]
Thanks! Done...

**Comment by t8m** [comment]
This should be directly cherry-pickable to master.

**Comment by misery** [comment]
> This should be directly cherry-pickable to master.

Works for me.

**Review by DDvO — APPROVED** [review]
Good that you took over also @richsalz's suggestion,
which makes it even nicer.

**Comment by DDvO** [comment]
This still requires an OTC approval.

**Comment by DDvO** [comment]
We should also clean up the code in order to get rid of all those "goto check_cert(_time);" which make the control flow hard to follow.
We could do that in a separate PR or as part of this one, removing the label and simplifying the beginning of `internal_verify()` to:
```
    int n = sk_X509_num(ctx->chain) - 1;
    X509 *xi = sk_X509_value(ctx->chain, n);
    X509 *xs = xi;

    ctx->error_depth = n;
    /*
     * With DANE-verified bare public key TA signatures, it remains only to
     * check the timestamps of the top certificate.  We report the issuer as
     * NULL, since all we have is a bare key.
     */
    if (ctx->bare_ta_signed) {
        xi = NULL;
    } else if (!ctx->check_issued(ctx, xi, xi)
               /* exceptional case: last cert in the chain is not self-issued */
               && ((ctx->param->flags & X509_V_FLAG_PARTIAL_CHAIN) == 0)) {
        if (n > 0) {
            n--;
            ctx->error_depth = n;
            xs = sk_X509_value(ctx->chain, n);
        } else if (!verify_cb_cert(ctx, xi, 0,
                                   X509_V_ERR_UNABLE_TO_VERIFY_LEAF_SIGNATURE)) {
            return 0;
        }
    }
```

t8m (approved):

t8m:
Thinking about this more, this is really something that does not belong to 1.1.1 branch as it is strictly speaking not a bug fix.
No need to update the PR, we will cherry-pick it to master.

t8m:
The change is also borderline on the CLA:trivial designation. But I think it is still acceptable. @DDvO do you agree this is trivial change?

DDvO:
> The change is also borderline on the CLA:trivial designation. But I think it is still acceptable. @DDvO do you agree this is trivial change?

Thanks @t8m for your further comments, your approval, and your question.
I definitely agree with CLA:trivial here, in particular since the final form of this minor improvement was suggested by @richsalz and me who have signed a CLA.

DDvO:
I suggest adding "Fixes #11297" to the commit message when doing the merge.

mspncp:
> I suggest adding "Fixes #11297" to the commit message when doing the merge.

It's already in the commit message: https://github.com/openssl/openssl/pull/11359/commits/231c228f14d716d8981c25421dc247c08251de60

DDvO:
@misery, I wanted to merge this (or better to say, cherry-pick the commit on top of master),
but the label referred to in your commit, which is meant for 1.1.1, needs to be adapted for the master:
replace `goto check_cert` by `goto check_cert_time`.

DDvO:
@misery, would it be important for your application to have this change also backported on version 1.1.1?

misery:
> @misery, would it be important for your application to have this change also backported on version 1.1.1?

Yes, otherwise it is remotely possible to skip further checks like expiration date. The attacker could use an out-dated certificate. 

Normally that would be stopped by our own truststore. But there could be a slightly time window until we dropped the certificate from the truststore.

DDvO:
So it really would be good to have this little feature extension/fix also in 1.1.1.
@t8m, @mattcaswell, @levitte, is this acceptable?

t8m:
I suppose we would have to have an OTC vote for that.

t8m:
@misery could you please open a new PR against master? We can merge that one easily and then we can have a vote for the 1.1.1 exception for this PR.

misery:
Thanks! I created a new PR for master: https://github.com/openssl/openssl/pull/13083

DDvO:
> I suppose we would have to have an OTC vote for that.

@t8m, could you please bring this to the agenda there?

romen:
I updated the labels to reflect the current state.

I also applied the `hold: need otc decision` label on behalf of @t8m that already opened a [vote about it](https://www.mail-archive.com/openssl-project@openssl.org/msg02158.html).

romen:
@misery @DDvO @t8m I am not terribly familiar with this code, and I feel inclined to abstain with a `+0` at the moment.

I think it would help informing my opinion for the vote, and easily turn this into a `+1` if any of you could write a short argument on why this could be seen as a bug or security fix to support inclusion in 1.1.1.

DDvO:
> I suppose we would have to have an OTC vote for that.

@t8m, could you please bring this to the agenda there?

> I updated the labels to reflect the current state.
> 
> I also applied the `hold: need otc decision` label on behalf of @t8m that already opened a [vote about it](https://www.mail-archive.com/openssl-project@openssl.org/msg02158.html).

Thanks @t8m and @romen.

DDvO:
> I think it would help informing my opinion for the vote, and easily turn this into a `+1` if any of you could write a short argument on why this could be seen as a bug or security fix to support inclusion in 1.1.1.

Within the OpenSSL X.509 certificate chain verification code in `x509_vfy.c` 
there are currently two flavors of invocations of the optional certificate verification callback function:
1. Giving the callback the chance to decide to abort or continue with standard verification
2. Giving the callback the chance to decide to abort or take over any remaining verification activity

where the first flavor is used in the majority of cases.

The choice with of the two flavors is used at which places appears pretty ad-hoc and is not documented,
which can be seen as a bug at design/conceptual level.

This PR contributes to solving this by changing for one situation the choice from 2. to 1., which appears more appropriate.

t8m:
@romen it is not a security issue by itself. Of course if application supplying the callback and overriding the UNABLE_TO_VERIFY_LEAF_SIGNATURE error blindly decides to ignore that expiration checks are not done it would be a security issue of the application. This change would just make it to be avoided.

IMO the argument that it is a bug fix is that for self-signed certificates the expiration check is called although the signature is not verified either (there is no sense in verifying the self signature). So the behavior is inconsistent between self-signed and non-self-signed leaf certificates.

romen:
Thanks both @t8m and @DDvO .

Regarding the "breaking change" aspect, is my understanding that this change is basically opt-in and only affect applications that want to catch `UNABLE_TO_VERIFY_LEAF_SIGNATURE` in this particular case, and thus can only benefit from merging this?

If my understanding is not correct, how likely do you estimate this change could affect existing applications? And how badly?

romen:
Thinking about the potentially breaking change, even in the case it is only opt-in and can't have negative effects.

One thing that I might like to see included in this PR before merging if the vote passes, is an entry in `NEWS` or `CHANGES` to alert users of the change (this would probably apply to `master` as well): what are your thoughts about it?

t8m:
Well in theory there could be existing application that wants to waive the UNABLE_TO_VERIFY_LEAF_SIGNATURE check and does not care about checking the expiration time and does not care about self signed certs and is unable to handle (waive) further errors from the expiration time checks. This application would be broken by this change. It is also fairly convolute and improbable scenario that such application exists.

DDvO:
> Well in theory there could be existing application that wants to waive the UNABLE_TO_VERIFY_LEAF_SIGNATURE check and does not care about checking the expiration time and does not care about self signed certs and is unable to handle (waive) further errors from the expiration time checks. This application would be broken by this change. It is also fairly convolute and improbable scenario that such application exists.

If such an application exists, it could still waive any errors from the expiration time checks, using the same callback function.

t8m:
The OTC vote has passed so the PR can be now merged.

t8m:
Merged to 1.1.1 branch. Thank you for the contribution.

PR #11360, opened 2020-03-18
Fix EVP_PKEY_new_mac_key()

mattcaswell:
EVP_PKEY_new_mac_key() was failing if the specified MAC was not available in the default provider - even though that MAC is never actually needed to successfully complete the function. The resulting EVP_PKEY can then be used in some non-default libctx which *does* have the MAC loaded.

I ran into this while working on #11334, when libssl called `EVP_PKEY_new_mac_key()`. My solution there avoided the problem (for libssl only) by making libssl use the EVP_MAC API instead. This fixes the underlying problem.

It actually means that we can defer a decision (for now) on whether or not to change libssl to use EVP_MAC (since that is a breaking change).

This includes the commit from #11354, so please only look at the second commit.

### DIFF
--- a/crypto/evp/pkey_mac.c
+++ b/crypto/evp/pkey_mac.c
@@ -53,19 +53,25 @@ static int pkey_mac_init(EVP_PKEY_CTX *ctx)
     int nid = ctx->pmeth->pkey_id;
     EVP_MAC *mac = EVP_MAC_fetch(ctx->libctx, OBJ_nid2sn(nid), ctx->propquery);
 
-    if (mac == NULL) {
-        EVPerr(EVP_F_PKEY_MAC_INIT, EVP_R_FETCH_FAILED);
-        return 0;
-    }
+    /*
+     * mac == NULL may actually be ok in some situations. In an
+     * EVP_PKEY_new_mac_key() call a temporary EVP_PKEY_CTX is created with
+     * default libctx. We don't actually need the underlying MAC to be present
+     * to successfully set the key in that case. The resulting EVP_PKEY could
+     * then be used in some other libctx where the MAC *is* present
+     */
+
     if ((hctx = OPENSSL_zalloc(sizeof(*hctx))) == NULL) {
         EVPerr(EVP_F_PKEY_MAC_INIT, ERR_R_MALLOC_FAILURE);
         return 0;
     }
 
-    hctx->ctx = EVP_MAC_CTX_new(mac);
-    if (hctx->ctx == NULL) {
-        OPENSSL_free(hctx);
-        return 0;
+    if (mac != NULL) {
+        hctx->ctx = EVP_MAC_CTX_new(mac);
+        if (hctx->ctx == NULL) {
+            OPENSSL_free(hctx);
+            return 0;
+        }
     }
 
     if (nid == EVP_PKEY_CMAC) {
@@ -98,6 +104,12 @@ static int pkey_mac_copy(EVP_PKEY_CTX *dst, const EVP_PKEY_CTX *src)
 
     EVP_PKEY_CTX_set_data(dst, dctx);
     dst->keygen_info_count = 0;
+
+    if (sctx->ctx == NULL) {
+        /* This actually means the fetch failed during the init call */
+        EVPerr(0, EVP_R_FETCH_FAILED);
+        return 0;
+    }
 
     dctx->ctx = EVP_MAC_CTX_dup(sctx->ctx);
     if (dctx->ctx == NULL)
@@ -146,7 +158,7 @@ static void pkey_mac_cleanup(EVP_PKEY_CTX *ctx)
     MAC_PKEY_CTX *hctx = ctx == NULL ? NULL : EVP_PKEY_CTX_get_data(ctx);
 
     if (hctx != NULL) {
-        EVP_MAC *mac = EVP_MAC_CTX_mac(hctx->ctx);
+        EVP_MAC *mac = hctx->ctx != NULL ? EVP_MAC_CTX_mac(hctx->ctx) : NULL;
 
         switch (hctx->type) {
         case MAC_TYPE_RAW:
@@ -181,8 +193,15 @@ static int pkey_mac_keygen(EVP_PKEY_CTX *ctx, EVP_PKEY *pkey)
         break;
     case MAC_TYPE_MAC:
         {
-            EVP_MAC_CTX *cmkey = EVP_MAC_CTX_dup(hctx->ctx);
+            EVP_MAC_CTX *cmkey;
 
+            if (hctx->ctx == NULL) {
+                /* This actually means the fetch failed during the init call */
+                EVPerr(0, EVP_R_FETCH_FAILED);
+                return 0;
+            }
+
+            cmkey = EVP_MAC_CTX_dup(hctx->ctx);
             if (cmkey == NULL)
                 return 0;
             if (!EVP_MAC_up_ref(EVP_MAC_CTX_mac(hctx->ctx)))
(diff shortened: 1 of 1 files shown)


crypto/evp/pkey_mac.c
```diff
@@ -53,19 +53,25 @@ static int pkey_mac_init(EVP_PKEY_CTX *ctx)
...
     EVP_MAC *mac = EVP_MAC_fetch(ctx->libctx, OBJ_nid2sn(nid), ctx->propquery);
 
-    if (mac == NULL) {
-        EVPerr(EVP_F_PKEY_MAC_INIT, EVP_R_FETCH_FAILED);
-        return 0;
-    }
+    /*
+     * mac == NULL may actually be ok in some situations. In an
+     * EVP_PKEY_new_mac_key() call a temporary EVP_PKEY_CTX is created with
+     * default libctx. We don't actually need the underlying MAC to be present
```
slontis: Are you saying the default libctx is the fips provider for example (otherwise this makes not much sense to me).

crypto/evp/pkey_mac.c
```diff
@@ -99,6 +105,12 @@ static int pkey_mac_copy(EVP_PKEY_CTX *dst, const EVP_PKEY_CTX *src)
     EVP_PKEY_CTX_set_data(dst, dctx);
     dst->keygen_info_count = 0;
 
+    if (sctx->ctx == NULL) {
```
slontis: This is too late to check this..
There is a if (sctx->ctx->data == NULL) check further up which is bad and will get picked up in static analysis.

crypto/evp/pkey_mac.c
```diff
@@ -377,6 +410,14 @@ static int pkey_mac_ctrl(EVP_PKEY_CTX *ctx, int type, int p1, void *p2)
                                                       p2, p1);
                 params[params_n] = OSSL_PARAM_construct_end();
 
+                if (hctx->ctx == NULL) {
```
slontis: Does the one on line 377 also need a check?

crypto/evp/pkey_mac.c
```diff
@@ -95,6 +105,12 @@ static int pkey_mac_copy(EVP_PKEY_CTX *dst, const EVP_PKEY_CTX *src)
     EVP_PKEY_CTX_set_data(dst, dctx);
     dst->keygen_info_count = 0;
 
+    if (sctx->ctx == NULL) {
+        /* This actually means the fetch failed during the init call */
+        EVPerr(0, EVP_R_FETCH_FAILED);
```
levitte: There is a `ERR_R_INIT_FAIL`...

crypto/evp/pkey_mac.c
```diff
@@ -177,8 +193,15 @@ static int pkey_mac_keygen(EVP_PKEY_CTX *ctx, EVP_PKEY *pkey)
         break;
     case MAC_TYPE_MAC:
         {
-            EVP_MAC_CTX *cmkey = EVP_MAC_CTX_dup(hctx->ctx);
+            EVP_MAC_CTX *cmkey;
 
+            if (hctx->ctx == NULL) {
+                /* This actually means the fetch failed during the init call */
+                EVPerr(0, EVP_R_FETCH_FAILED);
```
levitte: `ERR_R_INIT_FAIL`

crypto/evp/pkey_mac.c
```diff
@@ -220,6 +243,12 @@ static int pkey_mac_signctx_init(EVP_PKEY_CTX *ctx, EVP_MD_CTX *mctx)
         hctx->type == MAC_TYPE_RAW
         && (ctx->pmeth->flags & EVP_PKEY_FLAG_SIGCTX_CUSTOM) != 0;
 
+    if (hctx->ctx == NULL) {
+        /* This actually means the fetch failed during the init call */
+        EVPerr(0, EVP_R_FETCH_FAILED);
```
levitte: `ERR_R_INIT_FAIL`

crypto/evp/pkey_mac.c
```diff
@@ -285,6 +314,14 @@ static int pkey_mac_ctrl(EVP_PKEY_CTX *ctx, int type, int p1, void *p2)
                                                      ciphname, 0);
                 params[params_n] = OSSL_PARAM_construct_end();
 
+                if (hctx->ctx == NULL) {
+                    /*
+                     * This actually means the fetch failed during the init call
+                     */
+                    EVPerr(0, EVP_R_FETCH_FAILED);
```
levitte: `ERR_R_INIT_FAIL`

crypto/evp/pkey_mac.c
```diff
@@ -373,6 +410,14 @@ static int pkey_mac_ctrl(EVP_PKEY_CTX *ctx, int type, int p1, void *p2)
                                                       p2, p1);
                 params[params_n] = OSSL_PARAM_construct_end();
 
+                if (hctx->ctx == NULL) {
+                    /*
+                     * This actually means the fetch failed during the init call
+                     */
+                    EVPerr(0, EVP_R_FETCH_FAILED);
```
levitte: `ERR_R_INIT_FAIL`

crypto/evp/pkey_mac.c
```diff
@@ -385,6 +430,12 @@ static int pkey_mac_ctrl(EVP_PKEY_CTX *ctx, int type, int p1, void *p2)
     case EVP_PKEY_CTRL_DIGESTINIT:
         switch (hctx->type) {
         case MAC_TYPE_RAW:
+            if (hctx->ctx == NULL) {
+                /* This actually means the fetch failed during the init call */
+                EVPerr(0, EVP_R_FETCH_FAILED);
```
levitte: `ERR_R_INIT_FAIL`

crypto/evp/pkey_mac.c
mattcaswell: In the test case that failed the default libctx had the legacy provider loaded (only), but there was an alternative libctx that needed to use the EVP_PKEY being created. That libctx had the FIPS provider loaded. So HMAC did not exist in the default libctx and EVP_PKEY_new_mac_key() failed - even though I was eventually going to use the EVP_PKEY in a libctx that did have HMAC.

mattcaswell:
> This propagation of NULL checks seems quite ugly.

Yes, its horrible - but I couldn't think of an alternative.

crypto/evp/pkey_mac.c
mattcaswell: Fixed

crypto/evp/pkey_mac.c
mattcaswell: Yes. Added.

crypto/evp/pkey_mac.c
mattcaswell: I think EVP_R_FETCH_FAILED is a more useful error. It actually tells you the underlying cause of the problem, and I would think is more likely to lead to the end user figuring out what to do to fix it.

mattcaswell:
I rebased this to pick up the fix to #11354, and to review feedback fixes. I've not made the change to the error code since I think EVP_R_FETCH_FAILED is much more useful.

crypto/evp/pkey_mac.c
slontis: I dont have an issue with this..

t8m: Alternatively there could be a new error reason code created for this - something like EVP_R_MAC_CTX_NOT_FETCHED or EVP_R_MAC_CTX_MISSING

paulidale: I'm fine with any of the three approaches.

mattcaswell:
Pushed. Thanks.

PR #11365, opened 2020-03-19
Add EVP_PKEY_get_params support for accessing key fields

slontis:
Currently only RSA, EC and ECX are supported (DH and DSA need to be added to the keygen
PR's seperately because the fields supported have changed significantly).
    
Also added EVP_PKEY_gettable_params. Neither of these functions are dependant on any
particular operation - but they do require the keys to be provider based.
    
Made the export and get_params functions share the same code by supplying
support functions that work for both a OSSL_PARAM_BLD as well as a OSSL_PARAM[].
This approach means that complex code is not required to build an
empty OSSL_PARAM[] with the correct sized fields before then doing a second
pass to populate the array.
    

##### Checklist


slontis:
This is a replacement of #11352. It started out as just adding EC support, but quite a few changes also applied to the RSA code so it is easier just to merge them into a single cleaned up PR.

slontis:
@levitte I have reverted the multiprime changes in this PR.

doc/man3/EVP_PKEY_gettable_params.pod
```diff
@@ -0,0 +1,108 @@
...
+I<*out_sz> is the returned size of the string if it is not NULL.
+
+EVP_PKEY_get_octet_string_param() copy a I<pkey>'s octet string value into a buffer
+I<buf> of maximum size I<max_buf_sz> associated with a name of I<key_name>.
+I<*out_sz> is the returned size of the buffer if it is not NULL.
+
+=head1 NOTES
+
+These functions only work with key management methods coming from a
+provider.
```
levitte:
```
These functions only work the B<EVP_PKEY>s that contain a provider side key.
```

levitte:
If you're introducing numbered OSSL_PARAM key names, then you should probably revert 45211c563fb12aca50771b3400b833da4095c6de

slontis:
> If you're introducing numbered OSSL_PARAM key names, then you should probably revert [45211c5](https://github.com/openssl/openssl/commit/45211c563fb12aca50771b3400b833da4095c6de)
> (MPRSA keys were uniquely the driving force)

Although this doesnt actually exclude it from being used that way.

doc/man3/EVP_PKEY_gettable_params.pod
slontis: updated

levitte:
> For the RSA case it didnt sit well to do it as an unordered array

I'm lost, what was the unordered part?

slontis:
> > For the RSA case it didnt sit well to do it as an unordered array
> 
> I'm lost, what was the unordered part?

You are correct - There was no unordered part for them :) 
Giving them unique names makes it a very simple interface for the end user (i.e we dont need a stack or index reference) as well as not requiring extra args being passed down to the keymanagement get_params method() (OR yet another method to do this). The trade off being that a table of const names is required (since the names are just pointed to during param building).

slontis:
ping

paulidale:
> I think the KDF's still do this also?

Several of the PRFs replicate the same named item.

providers/implementations/keymgmt/kmgmt_util.c
```diff
@@ -0,0 +1,152 @@
...
+                                 const char *key, const BIGNUM *bn,  size_t sz)
+{
+    if (bld != NULL)
+        return ossl_param_bld_push_BN_pad(bld, key, bn, sz);
+    p = OSSL_PARAM_locate(p, key);
+    if (p != NULL) {
+        if (sz > p->data_size)
+            return 0;
+        p->data_size = sz;
+        return OSSL_PARAM_set_BN(p, bn);
```
mattcaswell: Doesn't this skip the padding part of what this function is supposed to do?

slontis: setting the data_size here achieves this if you look inside OSSL_PARAM_set_BN

paulidale: I thought that the data_size was meant to be _const_ and only the _return_size_ changes.  I think a new function `OSSL_PARAM_set_BN_pad()` is in order.

paulidale: #11390

slontis: Err what has this PR got to do with this?

crypto/evp/p_lib.c
```diff
@@ -1312,3 +1312,103 @@ void *evp_pkey_upgrade_to_provider(EVP_PKEY *pk, OPENSSL_CTX *libctx,
...
+
+    if (pkey == NULL
+        || pkey->keymgmt == NULL
+        || pkey->keydata == NULL
+        || key_name == NULL
+        || bn == NULL)
+        return 0;
+
+    params[0] = OSSL_PARAM_construct_BN(key_name, buffer, sizeof(buffer));
+    params[1] =  OSSL_PARAM_construct_end();
```
levitte: Nit: extra space

Cudos for being consistent, though 😉

crypto/rsa/rsa_ameth.c
```diff
@@ -1148,24 +1148,21 @@ static int rsa_pkey_export_to(const EVP_PKEY *from, void *to_keydata,
         for (i = 0; i < numprimes; i++) {
             const BIGNUM *num = sk_BIGNUM_const_value(primes, i);
 
-            if (!ossl_param_bld_push_BN(&tmpl, OSSL_PKEY_PARAM_RSA_FACTOR,
-                                        num))
+            if (!ossl_param_bld_push_BN(&tmpl, rsa_factor_names[i], num))
```
levitte: I'm lost, where does that array come from?

doc/man3/EVP_PKEY_gettable_params.pod
```diff
@@ -0,0 +1,107 @@
...
+ int EVP_PKEY_get_utf8_string_param(EVP_PKEY *pkey, const char *key_name,
+                                    char *str, size_t max_buf_sz, size_t *out_sz);
+ int EVP_PKEY_get_octet_string_param(EVP_PKEY *pkey, const char *key_name,
+                                    unsigned char *buf, size_t max_buf_sz,
+                                    size_t *out_sz);
+
+=head1 DESCRIPTION
+
+EVP_PKEY_gettable_params) returns a constant list of I<params> indicating
+the names and types of key parameters that can be retrieved.
```
levitte: Please include `L<OSSL_PARAM(3)>` somewhere

crypto/rsa/rsa_ameth.c
slontis: providers/implementations/keymgmt/kmgmt_util.c

include/internal/kmgmt_util.h
```diff
@@ -0,0 +1,31 @@
...
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
+ */
+
+#include "internal/param_build.h"
+#include <openssl/safestack.h>
+
+int ossl_kmgmt_params_set_int(OSSL_PARAM_BLD *bld, OSSL_PARAM *p,
+                              const char *key, int num);
```
levitte: Re the names, there's nothing particularly keymgmt about these functions...

providers/implementations/keymgmt/ec_kmgmt.c
```diff
@@ -42,7 +42,6 @@ static OSSL_OP_keymgmt_query_operation_name_fn ec_query_operation_name;
 
 #define EC_POSSIBLE_SELECTIONS                 \
     (OSSL_KEYMGMT_SELECT_KEYPAIR | OSSL_KEYMGMT_SELECT_ALL_PARAMETERS )
-
```
levitte: Why remove this line?

include/internal/kmgmt_util.h
slontis: Nup but their use is pretty specifically related to keymanagement..
Where do you think they belong? Or what namespace should it use?

crypto/rsa/rsa_ameth.c
levitte: I finally found it... in the deep recesses of provider code. I'm not sure I agree with this. We *do* cheat some, but this brings it to a new level...

include/internal/kmgmt_util.h
levitte: They look like fairly general ossl_param routines to me

crypto/rsa/rsa_ameth.c
slontis: At what level should I store them? crypto??

levitte: You should avoid shared storage if this sort. The proper way of finding out is with a gettable function and pilfering through the returned OSSL_PARAM array, but before you say it, that's problematic too since there's other stuff there as well.

This is actually a reason why I went with multiple instances of the same name. I was in exactly the rabbit hole you're finding yourself in a while ago, 'cause I also considered numbered names back when I added MPRSA support.

If you really want to do this, yeah I would have those arrays in a separate source file in crypto/rsa, and make sure that file gets added to the source of libcrypto and libfips.a

slontis: thanks - will do

levitte: Generally speaking, I believe that code that is shared between crypto/FOO/FOO_ameth.c and whatever within providers/implementations should go into crypto/FOO. At some point, when all legacy backends are gone, we can move it all to the providers tree.

include/internal/kmgmt_util.h
slontis: so we already have OSSL_PARAM_set_int(), and I didnt want to get these confused..
I put them in keymgmt because I thought they were probably only going to be used by that subsystem..

levitte: It's assumptions like that that have us end up with a generic URL parser as an OCSP function....
Since this is centered around OSSL_PARAM_BLD, that may be a more usable prefix?

providers/implementations/keymgmt/kmgmt_util.c
slontis: Put a TODO here for now.

crypto/evp/p_lib.c
```diff
@@ -1459,3 +1459,131 @@ int evp_pkey_downgrade(EVP_PKEY *pk)
...
+        || pkey->keymgmt == NULL
+        || pkey->keydata == NULL
+        || key_name == NULL
+        || bn == NULL)
+        return 0;
+
+    memset(buffer, 0, sizeof(buffer));
+    params[0] = OSSL_PARAM_construct_BN(key_name, buffer, sizeof(buffer));
+    /* If the return_size is not changed from -1 then we know it was not found */
+    params[0].return_size = TERMINATOR;
```
slontis: Although this is ugly it is the nicest way of doing it. Trying to do this on the destination is not practical.

crypto/evp/p_lib.c
```diff
@@ -1459,3 +1459,131 @@ int evp_pkey_downgrade(EVP_PKEY *pk)
...
+/*
+ * This value is set into param->return_size to indicate if
+ * evp_keymgmt_get_params() ignored the parameter.
+ * The EVP_PKEY_get_XXXX_param() methods return 0 if the parameter was ignored.
+ */
+#define TERMINATOR ((size_t)-1)
+
+int EVP_PKEY_get_bn_param(EVP_PKEY *pkey, const char *key_name, BIGNUM **bn)
+{
+    unsigned char buffer[1024];
```
paulidale: This shouldn't be a fixed size buffer.  I'd not be surprised if BN's greater than 8192 bits were required in the future.  I've certainly seen RSA keys long that this.

richsalz: Strongly agree.

slontis: Doubled the size so it now supports 16K.
Also added a TODO - this is a general issue with passing BIGNUM's. All other types have their own data_size.

include/internal/kmgmt_util.h
slontis: moved

slontis:
rebased and addressed review comments.

slontis:
rebased and ended up in a bad place..
Collapsing the rebase resulted in a worse place :)

crypto/evp/p_lib.c
levitte: Hah, I'm doing something similar in another place, but I do `params[0].return_size = params[0].data_size + 1`, and later on determine if it's been touched with `!params[0].return_size != params[0].data_size + 1`.
The reason I did this is because theoretically, even though unlikely, `(size_t)-1` is a perfectly valid size, so if `params[0].data_size` happens to be that, you're screwed.
(with my formula, overflow isn't a concern)

levitte: You could use `sizeof(buffer) + 1` instead of `params[0].data_size + 1` if you want a compile time constant.

slontis: Ok done for all types..

slontis:
ping
It would be good to get this merged soon..

crypto/evp/p_lib.c
```diff
@@ -1459,3 +1459,142 @@ int evp_pkey_downgrade(EVP_PKEY *pk)
...
+{
+    /*
+     * TODO(3.0) make this buffer dynamic.
+     * The evp_keymgmt_get_params() call may fail if the buffer is too small.
+     * We could possibly use the return_size, alloc and call again if this
+     * happens.
+     */
+    unsigned char buffer[2048];
+    OSSL_PARAM params[2];
+    size_t not_set = sizeof(buffer) + 1;
```
t8m: Use const ? Also would it make sense to use rather (size_t) -1 so it is invariant even if the buffer is eventually made a dynamically allocated one?

crypto/evp/p_lib.c
```diff
@@ -1459,3 +1459,142 @@ int evp_pkey_downgrade(EVP_PKEY *pk)
...
+        return 0;
+    return OSSL_PARAM_get_BN(params, bn);
+}
+
+int EVP_PKEY_get_octet_string_param(EVP_PKEY *pkey, const char *key_name,
+                                    unsigned char *buf, size_t max_buf_sz,
+                                    size_t *out_sz)
+{
+    OSSL_PARAM params[2];
+    size_t not_set = max_buf_sz + 1;
```
t8m: Same here.

doc/man3/EVP_PKEY_gettable_params.pod
```diff
@@ -0,0 +1,108 @@
...
+See L<OSSL_PARAM(3)> for information about parameters.
+
+EVP_PKEY_get_int_param() retrieves a key I<pkey> integer value I<*out>
+associated with a name of I<key_name>.
+
+EVP_PKEY_get_size_t_param() retrieves a key I<pkey> size_t value I<*out>
+associated with a name of I<key_name>.
+
+EVP_PKEY_get_bn_param() retrieves a key I<pkey> BIGNUM value I<**bn>
+associated with a name of I<key_name>. If I<*bn > is NULL then the BIGNUM
```
t8m: Extra space in `I<*bn >`

crypto/evp/p_lib.c
slontis: done

crypto/evp/p_lib.c
slontis: I will just use const - if we need to handle the dynamic case then it will be probably not be a const. This should do for now.

providers/implementations/keymgmt/ec_kmgmt.c
```diff
@@ -170,21 +167,19 @@ int key_to_params(const EC_KEY *eckey, OSSL_PARAM_BLD *tmpl, int include_private
         if (ecbits <= 0)
             goto err;
         sz = (ecbits + 7 ) / 8;
-        if (!ossl_param_bld_push_BN_pad(tmpl,
-                                        OSSL_PKEY_PARAM_PRIV_KEY,
-                                        priv_key, sz))
+        if (!ossl_param_build_set_bn_pad(tmpl, params,
+                                         OSSL_PKEY_PARAM_PRIV_KEY, priv_key,
```
levitte: Nit: I actually find it nicer if `OSSL_PKEY_PARAM_PRIV_KEY` stands alone on its line, just as in the other calls.

crypto/evp/p_lib.c
```diff
@@ -1459,3 +1459,142 @@ int evp_pkey_downgrade(EVP_PKEY *pk)
...
+
+int EVP_PKEY_get_bn_param(EVP_PKEY *pkey, const char *key_name, BIGNUM **bn)
+{
+    /*
+     * TODO(3.0) make this buffer dynamic.
+     * The evp_keymgmt_get_params() call may fail if the buffer is too small.
+     * We could possibly use the return_size, alloc and call again if this
+     * happens.
+     */
+    unsigned char buffer[2048];
```
kaduk: I thought the comment was that using any fixed-size buffer (even a 16kbit one) was not great.
How will this fare if, e.g., we get a post-quantum algorithm with very large keys?

slontis: I think more than this one function may need changing for post quantum :)

levitte:
Sorry for forgetting to flip the label...

doc/man3/EVP_PKEY_gettable_params.pod
```diff
@@ -0,0 +1,108 @@
...
+ int EVP_PKEY_get_bn_param(EVP_PKEY *pkey, const char *key_name, BIGNUM **bn);
+ int EVP_PKEY_get_utf8_string_param(EVP_PKEY *pkey, const char *key_name,
+                                    char *str, size_t max_buf_sz, size_t *out_sz);
+ int EVP_PKEY_get_octet_string_param(EVP_PKEY *pkey, const char *key_name,
+                                    unsigned char *buf, size_t max_buf_sz,
+                                    size_t *out_sz);
+
+=head1 DESCRIPTION
+
+EVP_PKEY_gettable_params) returns a constant list of I<params> indicating
```
levitte: Nit: missing a `(`

providers/implementations/keymgmt/ec_kmgmt.c
slontis: updated

slontis:
Rebased.
Why is travis not showing up here???

slontis:
@levitte - has to rebase (also addressed your nits). Does your approval still hold?

levitte:
Did you need to adjust more than util/libcrypto.num? If not, it still stqnds

slontis:
Sigh - Got a merge error in ec_kmgmt.c - so it does need another review now

slontis:
Rebased yet again.
Modified EVP_PKEY_get_bn_param so that it optionally used an allocated buffer and added a test for this case.

slontis:
collapsed the commits so that if someone causes me yet another merge it wont be as bad.

Changed EVP_PKEY_get_bn_param() so it cleanses the buffers it uses (Since it could be retrieving private data).

slontis:
rebased to fix libcrypto.num - no reapproval is required.

slontis:
Thanks.. Merged.

PR #11368, opened 2020-03-19
[WIP] EVP: Fix EVP_PKEY_copy_parameters() for a newly allocated |to|

### LINKED ISSUE #11366: EVP_PKEY_copy_parameters() broken
EVP_PKEY_copy_parameters() appears to broken if the "to" key is a newly allocated EVP_PKEY, and the "from" key is a provider only key. It appears to bail out because it thinks the key types are not consistent with each other.

levitte:
We didn't account for that possibility, and tried to upgrade an empty key
to be provided.  evp_pkey_upgrade_to_provider() doesn't handle empty keys
too well, so we need to check that it doesn't get one.

Fixes #11366

### DIFF
--- a/crypto/evp/p_lib.c
+++ b/crypto/evp/p_lib.c
@@ -120,15 +120,13 @@ int EVP_PKEY_copy_parameters(EVP_PKEY *to, const EVP_PKEY *from)
     }
 
     /*
-     * If |from| is provided, we upgrade |to| to be provided as well.
-     * This drops the legacy key from |to|.
-     * evp_pkey_upgrade_to_provider() checks if |to| is already provided,
-     * we don't need to do that here.
+     * If |from| is provided and |to| isn't, we upgrade |to| to be provided
+     * as well.  This drops the legacy key from |to|.
      *
      * TODO(3.0) We should investigate if that's too aggressive and make
      * this scenario unsupported instead.
      */
-    if (from->keymgmt != NULL) {
+    if (from->keymgmt != NULL && to->type != EVP_PKEY_NONE) {
         EVP_KEYMGMT *tmp_keymgmt = from->keymgmt;
 
         /*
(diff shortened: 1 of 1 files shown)


slontis (changes requested):
Code looks reasonable.. Please add the test to confirm there are not more problems :).

mattcaswell (approved):
Approved.

Note: I think we may also have problems where we copy parameters between two keys of the same type in the same provider. None of our current keymgmt implementation seem to implement the "copy" function (OSSL_FUNC_KEYMGMT_COPY).

mattcaswell (changes requested):
Oh...actually. Just read @slontis's comment. A test would be good.

levitte:
Yup, test on its way.

levitte:
Ohhhhhhh...  copy hasn't been implemented everywhere.  This opens up for a side question, should it be mandatory to implement OP_keymgmt_copy?  Right now it's not, 'cause I didn't see it as fundamental function, but maybe it is?

levitte:
Oh, now I see that @mattcaswell said that same thing

levitte:
As to the question if OP_keymgmt_copy should be mandatory, I remembered that it's not necessary.  There's a fallback to using export/import.

mattcaswell:
> As to the question if OP_keymgmt_copy should be mandatory, I remembered that it's not necessary. There's a fallback to using export/import.

Hmmm. I'm fairly sure I've seen this fail if you don't have the copy there, so I'm not sure this is working.

levitte:
> Hmmm. I'm fairly sure I've seen this fail if you don't have the copy there, so I'm not sure this is working.

Yup.  The code in `evp_keymgmt_util_copy()`  is stoopid and checks if `from->keymgmt->copy` is non-NULL before it has a chance of trying the export/import method...  I'm dealing with that right now.

levitte:
Closing this in favour of #11375

PR #11371, opened 2020-03-20
ecx: add key generation support.

paulidale:
X25519, X448, Ed25519 and Ed448.


paulidale:
Very early WIP, TLS tests are completely broken.

slontis:
In travis it doesnt get past
> make update
Due to s390 related code..

slontis:
Failure in ./test/evp_test ./test/recipes/30-test_evp_data/evppkey.txt is in fips mode as these are not in the fips module.
You may need to seperate them out into another .txt file and add into the .t file.

slontis:
It looks like you broke it good :)

slontis:
Pauli just having a cursory look at a failure in ssl_test_old:

I think it generates one of your new keys and then runs..

static int add_key_share(SSL *s, WPACKET *pkt, unsigned int curve_id)
{
    ....
    /* Encode the public key. */
    encodedlen = EVP_PKEY_get1_tls_encodedpoint(key_share_key,
                                                &encoded_point);

which then calls     rv = evp_pkey_asn1_ctrl(pkey, ASN1_PKEY_CTRL_GET1_TLS_ENCPT, 0, ppt);

which probably then calls legacy_asn1_ctrl_to_param which will not handle ASN1_PKEY_CTRL_GET1_TLS_ENCPT and then return -2;

@mattcaswell you may already be aware of this?

paulidale:
I added these in FIPS mode.  I suspect there are things that have been missed so far.
Or I've missed thing that I shouldn't have.

mattcaswell:
> @mattcaswell you may already be aware of this?

Yes - I made the same note here wrt ec keymgmt:

https://github.com/openssl/openssl/pull/11328#issuecomment-601253691

paulidale:
Failing in the TLS tests, but finally passing the EVP ones.

paulidale:
This seems to be close now, the TLS tests are rubbish but otherwise it seems to work.  After building this branch:
```
$ ./util/opensslwrap.sh genpkey -algorithm ED25519 -out ed.pem
$ ./util/opensslwrap.sh pkeyutl -sign -rawin -in /tmp/f -inkey ed.pem -out sig
```
Then on a fresh 3.0 (master) build:
```
$ ./util/opensslwrap.sh pkeyutl -verify -rawin -in /tmp/f -inkey ed.pem -sigfile sig
```
I get:
> Signature Verified Successfully

Where `/tmp/f` is a trivial test file being verified.

mattcaswell:
Can this be rebased? Will the various fixes that @levitte put in help with the TLS failures now?

mattcaswell:
Just needs a rebase (to address the merge conflicts) and this fix, and all the tests will start passing:

```` diff
diff --git a/crypto/evp/pmeth_gn.c b/crypto/evp/pmeth_gn.c
index 67800282de..9a7cdcf591 100644
--- a/crypto/evp/pmeth_gn.c
+++ b/crypto/evp/pmeth_gn.c
@@ -155,8 +155,9 @@ int EVP_PKEY_gen(EVP_PKEY_CTX *ctx, EVP_PKEY **ppkey)
 
         if (keydata == NULL)
             goto not_supported;
-        ret = evp_keymgmt_gen_set_template(ctx->keymgmt,
-                                           ctx->op.keymgmt.genctx, keydata);
+        if (ctx->keymgmt->gen_set_template != NULL)
+            ret = evp_keymgmt_gen_set_template(ctx->keymgmt,
+                                            ctx->op.keymgmt.genctx, keydata);
     }
 
     /*
````

File: crypto/ec/ecx_meth.c [hunk 1]
```diff
@@ -92,15 +92,19 @@ static int ecx_key_op(EVP_PKEY *pkey, int id, const X509_ALGOR *palg,
             X25519_public_from_private(pubkey, privkey);
             break;
         case EVP_PKEY_ED25519:
-            ED25519_public_from_private(pubkey, privkey);
+            /*
+             * TODO(3.0): We set the library context to NULL for now. This will
+             * need to change or this code be removed.
+             */
+            ED25519_public_from_private(NULL, pubkey, privkey);
```
mattcaswell, line 99: Once #11494 is merged the libctx will be available and can be used here. Please check the return code and return an error if it fails

crypto/ec/ecx_meth.c
```diff
@@ -92,15 +92,19 @@ static int ecx_key_op(EVP_PKEY *pkey, int id, const X509_ALGOR *palg,
...
         case EVP_PKEY_X448:
             X448_public_from_private(pubkey, privkey);
             break;
         case EVP_PKEY_ED448:
             /*
              * TODO(3.0): We set the library context to NULL for now. This will
-             * need to change.
+             * need to change or this code be removed.
              */
             ED448_public_from_private(NULL, pubkey, privkey);
```
mattcaswell, line 109: Here too

providers/fips/fipsprov.c
```diff
@@ -477,6 +481,10 @@ static const OSSL_ALGORITHM fips_keymgmt[] = {
     { "RSA:rsaEncryption", "provider=fips,fips=yes", rsa_keymgmt_functions },
 #ifndef OPENSSL_NO_EC
     { "EC:id-ecPublicKey", "provider=fips,fips=yes", ec_keymgmt_functions },
+    { "X25519", "provider=fips,fips=no", x25519_keymgmt_functions },
+    { "X448", "provider=fips,fips=no", x448_keymgmt_functions },
+    { "ED25519", "provider=fips,fips=no", ed25519_keymgmt_functions },
+    { "ED448", "provider=fips,fips=no", ed448_keymgmt_functions },
```
mattcaswell, line 487: The `oneshot_hash` function in the Ed448 code uses SHAKE256, but you have not also included that digest in the FIPS provider. This will cause issues with loading Ed448 private keys.

paulidale: Good catch.

crypto/ec/ecx_meth.c
paulidale: I've added the test and anticipate a merge conflict to fix.

slontis:
We agreed last night to include shake also if ecx is being added to the fips module.
Matt was going to do this... I will let you figure out with him who does this work..

paulidale:
Shake as FIPS approved?

slontis:
Hmm probably same as ecx  which is unapproved.

richsalz:
Apparently so, per p190 of the FIPS-140 I.G. document.

paulidale:
IG A.11 discusses CAVS testing for SHAKE, so it is possible to validate it.

slontis:
Its a question of what will be validated by the lab.

paulidale:
Travis failures are from an asan memory leak.

slontis:
@mattcaswell this is in the tls related code.
I am assuming this is related to calls to ssl_evp_cipher_fetch() which can return a fetched cipher.
Unless there are explicit matching free calls for each of these then there will be memory leaks,

paulidale:
The direct leak looks like it is a key manager from a fetch.

paulidale:
A pkey is being assigned to `s->s3.tmp.pkey` at the end of `add_key_share()` in `statem/extensions_clnt.c`.  This pointer remains until `SSL_free()` is called and isn't freed there -- memory leak.

For the _recordlen_ test, freeing the pointer in `SSL_free()` stop the leaks.  This isn't the solution since other tests crash.  I suspect it should be freed earlier as part of the state machine.  @mattcaswell, I've no idea where.

mattcaswell:
> A pkey is being assigned to s->s3.tmp.pkey at the end of add_key_share() in statem/extensions_clnt.c. This pointer remains until SSL_free() is called and isn't freed there -- memory leak.

`SSL_free()` calls `s->method->ssl_free()` which calls `ssl3_free()` which does this:

````
    EVP_PKEY_free(s->s3.tmp.pkey);
    s->s3.tmp.pkey = NULL;
````

**Comment by paulidale** [comment]
Missed that code path.  An additional EVP_PKEY_free() call did fix the memory leak, so something else is holding onto a reference :(

**Comment by mattcaswell** [comment]
> Missed that code path. An additional EVP_PKEY_free() call did fix the memory leak, so something else is holding onto a reference :(

Looking now.

**Comment by mattcaswell** [comment]
> Travis failures are from an asan memory leak.

What travis failures? I'm not seeing any memory leak either locally or in the travis output (travis is green apart from a timeout in one build)? Appveyor is also green.

**Comment by slontis** [comment]
@mattcaswell - I can only assume one of your other PR's that has been merged today has addressed whatever was causing the issue?

**Comment by mattcaswell** [comment]
> I can only assume one of your other PR's that has been merged today has addressed whatever was causing the issue?

*Possibly*  this could be something to do with the private key handling and the `public_from_private` calls in this PR.

**Comment by mattcaswell** [comment]
It looks to me that this can come out of WIP

File: providers/fips/fipsprov.c [hunk 4]
```diff
@@ -457,6 +461,8 @@ static const OSSL_ALGORITHM fips_signature[] = {
 #endif
     { "RSA:rsaEncryption", "provider=fips,fips=yes", rsa_signature_functions },
 #ifndef OPENSSL_NO_EC
+    { "ED25519", "provider=fips,fips=no", ed25519_signature_functions },
+    { "ED448", "provider=fips,fips=no", ed448_signature_functions },
```
mattcaswell, line 465: The "fips=no" here makes sense - but might be unexpected for some.  We should document this somewhere.

providers/implementations/keymgmt/ecx_kmgmt.c
```diff
@@ -224,10 +390,421 @@ static const OSSL_PARAM *ecx_gettable_params(void)
...
 MAKE_KEYMGMT_FUNCTIONS(x25519)
 MAKE_KEYMGMT_FUNCTIONS(x448)
 MAKE_KEYMGMT_FUNCTIONS(ed25519)
 MAKE_KEYMGMT_FUNCTIONS(ed448)
+
+#ifdef S390X_EC_ASM
+# include "s390x_arch.h"
+# include "internal/constant_time.h"
+
+static void s390x_x25519_mod_p(unsigned char u[32])
```
mattcaswell, line 408: All this seems to be duplicated from ecx_meth.c. Do we really have to have 2 identical copies of this code?

paulidale: I'd ideally like remove the code in `ecx_meth.c`.  I'm not sure how that plays with legacy support -- the functionality won't change but performance will.  There is additional code in `ecx_meth.c` which we don't want to carry across, so things needed to be split.

The arguments to the top level functions were different, so some code needs duplication.

I can look into splitting off some the code in libcrypto instead rather than slating it for removal from `ecx_meth.c`.

providers/fips/fipsprov.c
paulidale: It will be mentioned in the security policy and on the FIPS certificate.  There is a non-approved algorithm section.

I've added a NEWS entry noting this. It seems more appropriate than a CHANGES entry.

paulidale:
The two address sanitizer builds were failing, they aren't now -- I did rebase a couple of times last night.

providers/fips/fipsprov.c
mattcaswell: I this this section in provider.pod at least should be updated:

> The FIPS provider is a dynamically loadable module, and must therefore be loaded explicitly, either in code or through OpenSSL configuration (see config(5)). Should it be needed (if other providers are loaded and offer implementations of the same algorithms), the property "provider=fips" can be used as a search criterion for these implementations. All algorithm implementations in the FIPS provider can also be selected with the property "fips=yes".

providers/implementations/keymgmt/ecx_kmgmt.c
mattcaswell: I think the code that needs to be in both places should be split out into a separate file. The relevant functions can be defined in some internal header file, and the code should be included in both libcrypto and the fips provider.

paulidale:
Travis is the usual time out.

providers/implementations/keymgmt/ecx_kmgmt.c
paulidale: Working on it...

paulidale: It's in now.

providers/fips/fipsprov.c
paulidale: Done.

crypto/evp/pmeth_gn.c
```diff
@@ -172,8 +172,9 @@ int EVP_PKEY_gen(EVP_PKEY_CTX *ctx, EVP_PKEY **ppkey)
 
         if (keydata == NULL)
             goto not_supported;
-        ret = evp_keymgmt_gen_set_template(ctx->keymgmt,
-                                           ctx->op.keymgmt.genctx, keydata);
+        if (ctx->keymgmt->gen_set_template != NULL)
+            ret = evp_keymgmt_gen_set_template(ctx->keymgmt,
+                                               ctx->op.keymgmt.genctx, keydata);
```
levitte, line 177: This is superceded by #11550

paulidale: This has been kept as a separate commit, so easy to drop.  The CI's break without it.

Unfortunately, I just approved #11550 despite the typo, so I'll be playing catchup.

paulidale:
I've removed the commit.  The CIs are expected to fail at this point.

Once #11550 is merged, I'll rebase and confirm the CIs pass.

paulidale:
Travis is happy, merged.  Thanks for all the input and assistance.

PR #11379, opened 2020-03-22
util/wrap.pl: Correct exit code when signalled

levitte:
On Unix, a caught signal that exits the process does so with an exit
code that is 'signal | 128'.  This modifies util/wrap.pl to mimic
that.

### DIFF
--- a/util/wrap.pl
+++ b/util/wrap.pl
@@ -35,5 +35,12 @@ my $waitcode = system @cmd;
 # (exitcode << 8 | signalcode)
 die "wrap.pl: Failed to execute '", join(' ', @cmd), "': $!\n"
     if $waitcode == -1;
-exit($? & 255) if ($? & 255) != 0;
+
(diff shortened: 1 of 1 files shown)


slontis:
Is this fixing a particular issue - or was this just an observed during testing?

levitte:
Observed during testing. All shells do this. If you have a process that does a sigsegv and you're back to the shell prompt, the shell's `$?` will be 139, not 11.

Do note that Windows and VMS do not have Unix signaling at all, so that part of the wait() state will always be zero.

slontis:
Maybe make it even more obvious by using (yeah I know the order or operations means it already works)
(signal & 255) | 128

levitte:
You do réalisé, right, that the order of operations doesn't matter in this case? 😉
But yeah, I can do that, no prob

levitte:
Done, and with extra commentary

PR #11380, opened 2020-03-22
i2b_PVK_bio: don't set PEM_R_BIO_WRITE_FAILURE in case of success

chris2511:
but in case of an error

CLA: trivial


slontis:
I assume this is not tested anywhere.
It looks like apps.c can load this format? Ideally we would load one of the keys to test it.

levitte:
PVKs, are they still around?

t8m:
Merged to master and 1.1.1

Thank you for the PR.

Issue #11301, opened 2020-03-10
CMP Fuzzer


kroeckx:
I think we should add new fuzz programs for the CMP related changes.

@DDvO: Is that something you can do?

mattcaswell:
This is an excellent idea.

DDvO:
I can give it a try when I have more time...
Where can I find documentation how to do that?

kroeckx:
On Wed, Mar 18, 2020 at 05:35:33AM -0700, David von Oheimb wrote:
> I can give it a try when I have more time...
> Where can I find documentation how to do that?

We have existing fuzzers in the fuzz directory. You create a new
file per fuzzer, and need to implement a FuzzerTestOneInput()
function. You then let the fuzzer try to parse that input.

You probably also need the FuzzerInitialize() and FuzzerCleanup()
functions, but they can be empty.

If you have any questions, or need some help, let me know.

richsalz:
One simple thing is add new ASN.1 types to fuzz/asn1.c

DDvO:
Thanks @kroeckx and @richsalz for your hints.
I've started a PR that solves this issue: #11137

DDvO:
Closing this issue since it has been solved meanwhile by #11386.

PR #4277, opened 2017-08-28
Generalize setup_engine of apps/apps.c to allow specific crypto methods

DDvO:
The function `ENGINE *setup_engine(const char *engine, int debug)`, defined in `apps/apps.c`, is used by many OpenSSL apps. So far it unconditionally sets all the crypto methods of the given engine as the application-wide default. This is likely what  is needed for most (simple) use cases, in particular if an application uses just one private key (and related set of algorithms) or just one engine (corresponding to one type of key store) for all its key material.

Yet when several implementations for the same type of algorithm need to be used where, for instance, one of them is taken from an engine and others should be the standard ones defined in the crypto lib or taken from a different engine, it is not appropriate to make (all) the methods of one engine the default for all crypto algorithms. This holds in particular for a new app I am currently enhancing for later inclusion with OpenSSL. [CMPforOpenSSL](https://sourceforge.net/p/cmpforopenssl/git/ci/cmp/tree/) is a CMP client that can use signatures for protecting messages both at application and at TLS level, where for example the private key for CMP-level protection is held in a smart card (and accessed via an engine) and the private key for the (optional) TLS channel is stored in a password-protected file (and the keys may even be of different type, say, RSA and ECC).

I tried to make use of the given implementation of `setup_engine()` and to unregister any conflicting crypto methods, but this does not work. The only (clean) solution is to use a variant of `setup_engine()` that does not automatically call `ENGINE_set_default` with argument `ENGINE_METHOD_ALL`. Then for each private key one has full flexibility which crypto engine and methods (or the default OpenSSL crypto lib implementation) to use.

Therefore I propose to generalize `setup_engine()` by adding an extra argument, `unsigned int flags`, to be handed to `ENGINE_set_default`. This allows to decide if `ENGINE_set_default()` has any effect at all and to choose for which types of methods the implementation given by the engine is taken as the default.
Backward compatibility with existing code can be achieved easily by providing a macro that uses `ENGINE_METHOD_ALL` as value for the `flags`.

PR #4136, opened 2017-08-10
Script to find undocumented options

johndoe31415:
This script browses the manpages and tries all command line options and
shows where the two differ: Undocumented options (i.e., where an option
the help page mentions, but that doesn't appear in the manpage) and
unsupported options (i.e., where an option appears in the manpage, but
not in the help page).

I'm aware that most other scripts are written in Perl, but I'm unable to code any perl and wrote this script just to find discrepancies in the docs. So I thought I'd give it a shot to anyone who is interested.

PR #4096, opened 2017-08-05
Simplify the function OPENSSL_isservice

xiaoyinl:
I am not sure why a union is needed here. It seems to me a simple static function pointer variable is OK.
Also added a comment to clarify why we want to call the exported function `_OPENSSL_isservice`.

PR #11386, opened 2020-03-23
CMP Fuzzer

DDvO:
This solves #11301 by adding fuzzing of all public CMP and CRMF data types.


DDvO:
Adding CMP (and CRMF) types to `fuzz/asn1.c`, as suggested by @richsalz  was straightforward, 
except that it turned out that for visibility reasons this can be done only for the public ones.

Is this already sufficient, or what should be done in addition?

@kroeckx mentioned adding a file that implements a `FuzzerTestOneInput()` function, 
yet `asn1.c` already does so generically for the given ASN.1 types.

kroeckx:
I think that adding it to asn1 isn't actually useful. The coverage
will ge up, the fuzzer will slow down, but it's really just
fuzzing duplicates of the same code.

(I actually had modified asn1.c here like that in November.)

One like cms, crl and x509 might be more useful.

But I think as part of your CMP work, you added things like an
HTTP client. I would like to see something that does something
with HTTP.

I don't know if there are other protocols involved. If there are,
it would be useful to test them.

DDvO:
> I think that adding it to asn1 isn't actually useful. The coverage will ge up, the fuzzer will slow down, but it's really just fuzzing duplicates of the same code. (I actually had modified asn1.c here like that in November.)

So shall I remove that addition?

> One like cms, crl and x509 might be more useful. 

I've just added `fuzz/cmp.c` according to the model of `fuzz/cms.c` because it is most closely related, but this is suspiciously simple (like adding types to `fuzz/asn1.c` was).
I also found that `fuzz/build.info` and `test/recipes/99-test_fuzz.t` need to be extended accordingly.

Yet when invoking
``` 
make tests TESTS="test_fuzz" V=1
```
I only get
```
    1..1
../../util/wrap.pl ../../fuzz/cmp-test ../../fuzz/corpora/cmp => 0
    ok 1
ok 1 - Fuzzing cmp
```
I suspect I need to set up `fuzz/corpora/cmp/` with all those hash-named files, but how?

A documentation/HOWTO for the OpenSSL fuzzing facility would be helpful here.

DDvO:
> But I think as part of your CMP work, you added things like an HTTP client. I would like to see something that does something with HTTP.

CMP, like CMS, is actually independent of the transfer protocol - HTTP is just the default.
So I think they are better tested independently.

HTTP is not an ASN.1-based protocol, and I have no clue how to fuzz that.
When adding the generic HTTP client support I've already added non-fuzzing tests for it:
```
test/http_test.c
test/recipes/80-test_http.t
```

> I don't know if there are other protocols involved. If there are, it would be useful to test them.

CMP itself does not involve any further protocols (while its data structures make use of CRMF and X509, which are formats, not protocols, so AFAICS do not play a direct role here).

kroeckx:
On Tue, Mar 24, 2020 at 03:07:08AM -0700, David von Oheimb wrote:
> A documentation/HOWTO for the OpenSSL fuzzing facility would be helpful here.

Have you looked at fuzz/README.md?


Kurt

DDvO:
Thanks @kroeckx for pointing at `fuzz/README.md`, which I had not noticed.
This explains how to fill `fuzz/corpora/cmp/`.

Yet `fuzz/helper.py cmp` has already been running for more than two days (on one CPU core), so far producing 13 MB data. Is there a way to see a forecast how much longer/more this is going to take?

kroeckx:
> Yet `fuzz/helper.py cmp` has already been running for more than two days (on one CPU core), so far producing 13 MB data. Is there a way to see a forecast how much longer/more this is going to take?

It will until run until you stop it. So just stop it, and add the
corpus.

DDvO:
> It will until run until you stop it. So just stop it, and add the corpus.

Oh! This important detail seems not mentioned in the README.

Actually I have data from two runs: 13.2 MB from the long run I mentioned, which I just stopped,
and 3.6 MB from an earlier run which I had cancelled for other reasons.

Which one is preferred? I presume there should be a balance between completeness and data size. Here is a comparison with the data sizes of the already included fuzzers:
```
du fuzz/corpora/* | sort -n

960	fuzz/corpora/ct
1928	fuzz/corpora/conf
1996	fuzz/corpora/bndiv
2184	fuzz/corpora/asn1parse
2196	fuzz/corpora/bignum
3600	fuzz/corpora/cmp_old
7152	fuzz/corpora/cms
8208	fuzz/corpora/server
9128	fuzz/corpora/crl
9816	fuzz/corpora/client
10168	fuzz/corpora/x509
13284	fuzz/corpora/cmp
15836	fuzz/corpora/asn1
```

kroeckx:
On Mon, Mar 30, 2020 at 12:41:10AM -0700, David von Oheimb wrote:
> > It will until run until you stop it. So just stop it, and add the corpus.
> 
> Oh! This important detail seems not mentioned in the README.
> 
> Actually I have data from two runs: 13.2 MB from the long run I mentioned, which I just stopped,
> and 3.6 MB from an earlier run which I had cancelled for other reasons.
> 
> Which one is preferred?

I recommend that you try to minimize it, that is run:
./cmp x509 -merge=1 new-dir old-dir

Where new-dir is an empty directory. You can put all the files you
have in the old-dir.


Kurt

DDvO:
> It will until run until you stop it. So just stop it, and add the corpus.

Oh! This important detail seems not mentioned in the README.

Actually I have data from two runs: 13.2 MB from the long run I mentioned, which I just stopped,
and 3.6 MB from an earlier run which I had cancelled for other reasons.

Which one is preferred? I presume there should be a balance between completeness and data size. Here is a comparison with the data sizes of the already included fuzzers:
```
du fuzz/corpora/* | sort -n

960	fuzz/corpora/ct
1928	fuzz/corpora/conf
1996	fuzz/corpora/bndiv
2184	fuzz/corpora/asn1parse
2196	fuzz/corpora/bignum
3600	fuzz/corpora/cmp_old
7152	fuzz/corpora/cms
8208	fuzz/corpora/server
9128	fuzz/corpora/crl
9816	fuzz/corpora/client
10168	fuzz/corpora/x509
13284	fuzz/corpora/cmp
15836	fuzz/corpora/asn1
```

DDvO:
> I recommend that you try to minimize it, that is run: ./cmp x509 -merge=1 new-dir old-dir Where new-dir is an empty directory. You can put all the files you have in the old-dir.

Good to know, will do.
Should I also run AFL mentioned in the README?
```
 afl-fuzz -i fuzz/corpora/$FUZZER -o fuzz/corpora/$FUZZER/out fuzz/$FUZZER
```

BTW, the copy of my last response was a mistake - my browser view of this page was inconsistent.

kroeckx:
> Should I also run AFL mentioned in the README?

No, libfuzzer alone is fine.

Once it's merged, Google will run that fuzzer with multiple
fuzzers including AFL. I will then download their corpus as some
point and merge it.

DDvO:
> I recommend that you try to minimize it, that is run:
> ./cmp x509 -merge=1 new-dir old-dir

> Where new-dir is an empty directory.
> You can put all the files you have in the old-dir.

This took about a day to complete, but apparently had no effect - `new-dir` is still empty :-(

Why to use the `x509` parameter here? Maybe this is wrong.

Should I better use, e.g., this? `./cmp new-dir -merge=1 old-dir`

kroeckx:
On Tue, Mar 31, 2020 at 11:44:35PM -0700, David von Oheimb wrote:
> > I recommend that you try to minimize it, that is run:
> > ./cmp x509 -merge=1 new-dir old-dir
> 
> > Where new-dir is an empty directory.
> > You can put all the files you have in the old-dir.
> 
> This took about a day to complete, but apparently had no effect - `new-dir` is still empty :-(
> 
> Why to use the `x509` parameter here? Maybe this is wrong.
> 
> Should I better use, e.g., this? `./cmp new-dir -merge=1 old-dir`

It clearly should just have been:
./cmp -merge=1 new-dir old-dir

I assume you have an x509 dir, and it put things in the x509 dir.


Kurt

DDvO:
> ./cmp -merge=1 new-dir old-dir

I'm running this now, let's see how this goes ...

> I assume you have an x509 dir, and it put things in the x509 dir.

Yes, `fuzz/corpora/x509/` exists also over here, 
but I had already checked: nothing new went in there.

DDvO:
Also
```
./cmp -merge=1 new-dir old-dir`
```
did not give any output and apparently terminated immediately, without effect.

DDvO:
In the meantime I've tentatively added the combined but non-merged/minimized data to `fuzz/corpora/cmp/`.
```
make tests TESTS="test_fuzz" V=1
```
now does include `fuzz/corpora/cmp/` and passes.

DDvO:
The cmp data added just about 2.5 seconds to the overall time (which is some 80 seconds on my machine) the fuzz tests took.
I cannot judge if this is a good sign or not.

kroeckx:
I just did a quick look at your fuzzer. Maybe you can pass
the msg to functions like:
- OSSL_CMP_SRV_process_request()
- OSSL_CMP_CTX_server_perform()

That is, write some basic applications that does something with
such a file.

I tried to minimize the corpus, and I had no problem with it:
$ ./fuzz/cmp -merge=1 fuzz/new/cmp fuzz/corpora/cmp/
INFO: Seed: 3695945473
INFO: Loaded 1 modules   (183405 inline 8-bit counters): 183405 [0x1d99ea0, 0x1dc6b0d),
INFO: Loaded 1 PC tables (183405 PCs): 183405 [0x1dc6b10,0x20931e0),
MERGE-OUTER: 4136 files, 0 in the initial corpus
MERGE-OUTER: attempt 1
INFO: Seed: 3696161283
INFO: Loaded 1 modules   (183405 inline 8-bit counters): 183405 [0x1d99ea0, 0x1dc6b0d),
INFO: Loaded 1 PC tables (183405 PCs): 183405 [0x1dc6b10,0x20931e0),
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 1048576 bytes
MERGE-INNER: using the control file '/tmp/libFuzzerTemp.1357149.txt'
MERGE-INNER: 4136 total files; 0 processed earlier; will process 4136 files now
#1      pulse  cov: 438 exec/s: 0 rss: 48Mb
#2      pulse  cov: 438 exec/s: 0 rss: 50Mb
#4      pulse  cov: 439 exec/s: 0 rss: 50Mb
#8      pulse  cov: 439 exec/s: 0 rss: 50Mb
#16     pulse  cov: 527 exec/s: 0 rss: 50Mb
#32     pulse  cov: 655 exec/s: 0 rss: 50Mb
#64     pulse  cov: 696 exec/s: 0 rss: 51Mb
#128    pulse  cov: 752 exec/s: 0 rss: 52Mb
#256    pulse  cov: 1462 exec/s: 0 rss: 56Mb
#512    pulse  cov: 1828 exec/s: 512 rss: 61Mb
#1024   pulse  cov: 1863 exec/s: 1024 rss: 69Mb
#2048   pulse  cov: 1987 exec/s: 409 rss: 87Mb
#4096   pulse  cov: 1987 exec/s: 341 rss: 162Mb
#4136   DONE  cov: 1987 exec/s: 344 rss: 166Mb
MERGE-OUTER: succesfull in 1 attempt(s)
MERGE-OUTER: the control file has 448086 bytes
MERGE-OUTER: consumed 0Mb (53Mb rss) to parse the control file
MERGE-OUTER: 1268 new files with 8138 new features added; 1987 new coverage edges

I've put that at https://www.roeckx.be/cmp.tar.gz


Kurt

DDvO:
>  I tried to minimize the corpus, and I had no problem with it: 
> $ ./fuzz/cmp -merge=1 fuzz/new/cmp fuzz/corpora/cmp/

Argh, meanwhile I found that in my builds of those excecutables like `fuzz/cmp` were garbled,
which explains their weird behavior.

I suspect they were (wrongly) produced by my experiments with AFL or the like.
How can one produce them correctly, as you did?
With a normal configuration they do not show up as make targets.

DDvO:
BTW, would be generally very helpful in my view if `Configure` (and thus also `config`) automatically did a `make clean` or the like to avoid inconsistencies with artifacts of earlier builds in the same directory.

DDvO:
> I've put that at https://www.roeckx.be/cmp.tar.gz Kurt

Thanks! I've replaced `fuzz/corpora/cmp/*` with that compressed data.
The fuzz tests also succeed with this data over here.

kroeckx:
On Wed, Apr 01, 2020 at 03:47:57AM -0700, David von Oheimb wrote:
> > I tried to minimize the corpus, and I had no problem with it: 
> > $ ./fuzz/cmp -merge=1 fuzz/new/cmp fuzz/corpora/cmp/
> 
> Argh, meanwhile I found that in my builds of those excecutables like `fuzz/cmp` were garbled,
> which explains their weird behavior.
> 
> I suspect they were (wrongly) produced by my experiments with AFL or the like.
> How can one produce them correctly, as you did?
> With a normal configuration they do not show up as make targets.

I did:
CC=clang-9 ./config enable-fuzz-libfuzzer --with-fuzzer-lib=/usr/lib/llvm-9/lib/clang/9.0.1/lib/linux/libclang_rt.fuzzer-x86_64.a -DPEDANTIC enable-asan enable-ubsan no-shared -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=fuzzer-no-link enable-ec_nistp_64_gcc_128 -fno-sanitize=alignment enable-weak-ssl-ciphers enable-rc5 enable-md2 enable-ssl3 enable-ssl3-method enable-nextprotoneg --debug
LDCMD=clang++-9 make -j4

You might need to change the version of clang.

The "enable-fuzz-libfuzzer" will change the Makefile to generate the
fuzz/cmp file.


Kurt

DDvO:
I see. I tried your new version of the config invocation
and found that I had to remove `-Wa,--noexecstack`.
My clang(++) version is 7.

Now I get on `LDCMD=clang++ make fuzz/cmp`:
```
rm -f fuzz/cmp
${LDCMD:-gcc} -pthread -m64 -fno-omit-frame-pointer -g  -Wall -O0 -g -DDEBUG_UNUSED -DPEDANTIC -pedantic -Wno-long-long -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -Wswitch -Wsign-compare -Wshadow -Wformat -Wtype-limits -Wundef -Werror -Wmissing-prototypes -Wstrict-prototypes -fstack-protector -fno-omit-frame-pointer -std=gnu90 -L.   \
	-o fuzz/cmp \
	fuzz/cmp-bin-cmp.o fuzz/cmp-bin-driver.o \
	-lcrypto -ldl -pthread 
/usr/bin/ld: /usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../../x86_64-linux-gnu/crt1.o: in function `_start':
(.text+0x20): undefined reference to `main'
```
Yet maybe I won't strictly need that `fuzz/cmp` executable.

kroeckx:
> (.text+0x20): undefined reference to `main'

Are you sure you passed the --with-fuzzer-lib? The library
provides the main function.

DDvO:
Now the corpora data merging works, with very similar result as your :-)

DDvO:
I've just extended `fuzz/README.md` according to the lessons I've learned today.

DDvO:
> I just did a quick look at your fuzzer. Maybe you can pass the msg to functions like:
> - OSSL_CMP_SRV_process_request()
> - OSSL_CMP_CTX_server_perform()

> That is, write some basic applications that does something with such a file.

I've extended `fuzz/cmp.c` with CMP client and server stubs trying to use the fuzzed CMP msg.
Is this what you meant?

Still the fuzzing went pretty fast and succeeded.

kroeckx:
On Wed, Apr 01, 2020 at 07:13:03AM -0700, David von Oheimb wrote:
> 
> I've extended `fuzz/cmp.c` with CMP client and server stubs trying to use the fuzzed CMP msg.
> Is this what you meant?

Yes. The coverage went up, it's finding lots of new inputs now.
(I have it running here.)

DDvO:
Pleased to hear. I'm going to further strengthen this by adapting the client and server state such that the messages can get past the initial protection/nonce/transactionID checks, which currently in all likelihood will block all input messages right away.

kroeckx:
Examples of real files could also be very useful

kroeckx:
Anyway, the FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION define can be
used to remove such checks, or ignore the result of the check.

DDvO:
@kroeckx, as you suggested I added to the corpora various "real" CMP messages (from other tests etc.) exemplifying all (supported) CMP message types.

I also was able to make good use of your `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` hint.
I've made sure and verified manually that, after "dismantling" some checks in this way,
* for each supported request message type the server skeleton handles it
* for each supported response message type the client follows some path accepting it.

Doing so I found and fixed a few cases there the error reason codes were not fully adequate.
I did not encounter any severe misbehavior of the client nor the server side.

So in my view the CMP fuzzing has become reasonably strong and is successful.
Consequently I've just removed the "WIP:" from the title of this PR.
Thanks Kurt for shepherding this development!

DDvO:
The CIFuzz run failed, and I don't know how to fix
```
FileNotFoundError: [Errno 2] No such file or directory: '/github/workspace/out/oss_fuzz_latest/openssl/cmp'
```

kroeckx:
I assume the problem is that it's a new fuzzer, and that it will
fix itself when it's merged. I suggest to ignore it.

I've been running it with the following patch applied:
--- a/fuzz/cmp.c
+++ b/fuzz/cmp.c
@@ -15,9 +15,14 @@
 #include <openssl/cmp.h>
 #include <openssl/err.h>
 #include "fuzzer.h"
+#include "rand.inc"

 int FuzzerInitialize(int *argc, char ***argv)
 {
+    OPENSSL_init_crypto(OPENSSL_INIT_LOAD_CRYPTO_STRINGS, NULL);
+    ERR_clear_error();
+    CRYPTO_free_ex_index(0, -1);
+    FuzzerSetRand();
     return 1;
 }

@@ -50,6 +55,7 @@ int FuzzerTestOneInput(const uint8_t *buf, size_t len)
         OSSL_CMP_CTX *client_ctx = OSSL_CMP_CTX_new();

         i2d_OSSL_CMP_MSG_bio(out, msg);
+        ASN1_item_print(out, (ASN1_VALUE *)msg, 4, ASN1_ITEM_rptr(OSSL_CMP_MSG), NULL);
         BIO_free(out);

         if (client_ctx != NULL)

And currently get:
#13522  INITED cov: 5852 ft: 20834 corp: 2428/1317Kb exec/s: 540 rss: 586Mb

I merged the existing files from the test suite, which greatly
increased the coverage. The fuzzer has found a lot of new coverage
(cov), covering a lot of new function since I've added the samples
from the test suite. If you can share more real files, I would love
to merge them too.

I'm not sure FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION is needed, I
think the fuzzer can actually find the correct input. It is at
least finding lots of inputs without it.

kroeckx:
I've put my current corpus at:
https://www.roeckx.be/cmp.tar.gz

kroeckx:
I've merged your new corpus, and now I'm finding lots of new
inputs again. I will let this run for an other day and send you a
new corpus again.

kroeckx:
It seems that your new version is now writing all kind of error
messages to stderr. I'm not sure why it's doing that, but it
should not output anything.

kroeckx:
I only seem to be getting all the error messages on stderr when
FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION is defined. I don't see
why FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION should have that
effect, so I assume it's non-modified code that does that.

In any case, the things you did with
FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION seem to improve the
coverage even more, so those changes look good.

kroeckx:
And it found the first issue, I have a NULL pointer dereference.
==1681722==ERROR: UndefinedBehaviorSanitizer: SEGV on unknown address 0x000000000006 (pc 0x0000006820ac bp 0x7fff2becb3a0 sp 0x7fff2becb350 T1681722)
==1681722==The signal is caused by a READ memory access.
==1681722==Hint: address points to the zero page.
    #0 0x6820ab in asn1_string_get_int64 crypto/asn1/a_int.c:325:10
    #1 0x681fe9 in ASN1_INTEGER_get_int64 crypto/asn1/a_int.c:515:12
    #2 0x4c96d8 in ossl_cmp_asn1_get_int crypto/cmp/cmp_asn.c:195:10
    #3 0x4dcb75 in suitable_rid crypto/cmp/cmp_msg.c:865:12
    #4 0x4dce22 in ossl_cmp_certrepmessage_get0_certresponse crypto/cmp/cmp_msg.c:925:13
    #5 0x4e9779 in check_msg_valid_cert_3gpp crypto/cmp/cmp_vfy.c:329:13
    #6 0x4e9dc2 in check_msg_with_certs crypto/cmp/cmp_vfy.c:370:25
    #7 0x4e98e3 in check_msg_all_certs crypto/cmp/cmp_vfy.c:399:9
    #8 0x4e79e7 in check_msg_find_cert crypto/cmp/cmp_vfy.c:462:16
    #9 0x4e6fbc in OSSL_CMP_validate_msg crypto/cmp/cmp_vfy.c:612:17
    #10 0x4e8be2 in ossl_cmp_msg_check_received crypto/cmp/cmp_vfy.c:662:14
    #11 0x4cc12e in send_receive_check crypto/cmp/cmp_client.c:186:15
    #12 0x4cf1a3 in OSSL_CMP_exec_GENM_ses crypto/cmp/cmp_client.c:884:10
    #13 0x42bbfe in cmp_client_process_response fuzz/cmp.c:77:15
    #14 0x42b8b9 in FuzzerTestOneInput fuzz/cmp.c:162:13
    #15 0x42c3f4 in LLVMFuzzerTestOneInput fuzz/driver.c:28:12
    #16 0x4684b1 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (fuzz/cmp+0x4684b1)
    #17 0x467cf5 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool*) (fuzz/cmp+0x467cf5)
    #18 0x469f97 in fuzzer::Fuzzer::MutateAndTestOne() (fuzz/cmp+0x469f97)
    #19 0x46acb5 in fuzzer::Fuzzer::Loop(std::Fuzzer::vector<fuzzer::SizedFile, fuzzer::fuzzer_allocator<fuzzer::SizedFile> >&) (fuzz/cmp+0x46acb5)
    #20 0x458a98 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) (fuzz/cmp+0x458a98)
    #21 0x481ee2 in main (fuzz/cmp+0x481ee2)
    #22 0x7fe108d78e0a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x26e0a)
    #23 0x409c89 in _start (fuzz/cmp+0x409c89)

UndefinedBehaviorSanitizer can not provide additional info.
==1681722==ABORTING
MS: 1 ChangeBit-; base unit: bf78c945dbccc4f52404eb701ebf3683ff2f9fb0
artifact_prefix='./'; Test unit written to ./crash-2727345000496842caca113d71a529f694da6536

File at:
https://www.roeckx.be/crash-2727345000496842caca113d71a529f694da6536

DDvO:
Thanks @kroeckx for all new comments!

> I only seem to be getting all the error messages on stderr when FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION is defined. I don't see why FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION should have that effect, so I assume it's non-modified code that does that. In any case, the things you did with FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION seem to improve the coverage even more, so those changes look good.

I've used `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` mostly to let the CMP client and server code ignore the results of various validity checks on incoming messages: on the message protection (signature- or password-based), various generic header field values (CMP version, transaction ID, and nonces), body type consistency, and the proof-of-possession (POPO) signature in cert requests. 

In order to make the client to proceed where according to its current state (in the CMP context) it would normally return an error, e.g, on a missing private key missing, I've also made some use of  `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` and added some CMP option settings at the beginning of `cmp_client_process_response()` in `fuzz/cmp.c`.

DDvO:
> I've put my current corpus at: https://www.roeckx.be/cmp.tar.gz

I tried amending the PR with that data, but apparently it was already included.

I added to the PR the patch you mentioned above.

kroeckx:
> I tried amending the PR with that data, but apparently it was already included.

The one I have now is much larger, I'll update it once it stops
finding new things for a while. The last new function it found
was PACKET_get_net_2, which I don't see being used outside ssl ...

kroeckx:
Before this can move on, I think at least the follow things need
to be fixed:
- There is a memory leak. It looks like the OSSL_CMP_MSG_dup() in
  transfer_cb() is never freed.
- The NULL pointer dererence that ossl_cmp_asn1_get_int caused

I would also like to see all the error messages that are printed
to stderr are no longer printed. The library should not be printing
errors to stderr, it's up to the application to decide what to do
with the error messages.

I think only the memory leak should actually get fixed in this PR,
the rest should probably be done in other PRs.

kroeckx:
Here is a crash with a different profile:
https://www.roeckx.be/crash-57e4fbcd139ec37e87aca0a695c194e68f1abcb7

    #0 0x5b37f0 in X509_NAME_cmp crypto/x509/x509_cmp.c:172:10
    #1 0x4ce983 in OSSL_CMP_exec_RR_ses crypto/cmp/cmp_client.c:835:13
    #2 0x42bbe0 in cmp_client_process_response fuzz/cmp.c:74:15
    #3 0x42b8b9 in FuzzerTestOneInput fuzz/cmp.c:162:13

DDvO:
> I'm not sure FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION is needed, I think the fuzzer can actually find the correct input. It is at least finding lots of inputs without it.

I'm meanwhile sure that `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` is necessary for significantly improving completeness since the fuzzer would have little chance guessing symmetric and private keys needed for protecting CMP messages in a way the (unmodified) CMP protection checker would accept. 
I also cannot imagine that it could guess transaction IDs and nonces.

DDvO:
> And it found the first issue, I have a NULL pointer dereference.
==1681722==ERROR: UndefinedBehaviorSanitizer: SEGV on unknown address 0x000000000006 (pc 0x0000006820ac bp 0x7fff2becb3a0 sp 0x7fff2becb350 T1681722) ==1681722==The signal is caused by a READ memory access. ==1681722==Hint: address points to the zero page. #0 0x6820ab in asn1_string_get_int64 crypto/asn1/a_int.c:325:10 #1 0x681fe9 in ASN1_INTEGER_get_int64 crypto/asn1/a_int.c:515:12 #2 0x4c96d8 in ossl_cmp_asn1_get_int crypto/cmp/cmp_asn.c:195:10 #3 0x4dcb75 in suitable_rid crypto/cmp/cmp_msg.c:865:12 #4 0x4dce22 in ossl_cmp_certrepmessage_get0_certresponse crypto/cmp/cmp_msg.c:925:13 #5 0x4e9779 in check_msg_valid_cert_3gpp crypto/cmp/cmp_vfy.c:329:13

BTW, how can I reproduce such crashes?

> - The NULL pointer dererence that ossl_cmp_asn1_get_int caused

I believe this would have already been fixed by #11448, which unfortunately is not yet reviewed and merged.

For simplicity and such that we can proceed this fuzzing PR I've cerry-picked the essential commit https://github.com/openssl/openssl/pull/11448/commits/1da812778f9098286b8f069391106d0d2eb05017 of that PR to this one.

DDvO:
> There are some situations where this leads the client to output errors via `OSSL_CMP_CTX_print_errors(ctx)` or the like. If this is a problem I can of course suppress that.

I've meanwhile suppressed such error queue output while fuzzing.

DDvO:
> Here is a crash with a different profile: https://www.roeckx.be/crash-57e4fbcd139ec37e87aca0a695c194e68f1abcb7
 #0 0x5b37f0 in X509_NAME_cmp crypto/x509/x509_cmp.c:172:10
 #1 0x4ce983 in OSSL_CMP_exec_RR_ses crypto/cmp/cmp_client.c:835:13
 #2 0x42bbe0 in cmp_client_process_response fuzz/cmp.c:74:15
 #3 0x42b8b9 in FuzzerTestOneInput fuzz/cmp.c:162:13

This must have been an artifact due to insufficient `ctx->oldCert` initialization in `cmp_client_process_response()` in `fuzz/cmp.c` - fixed.

DDvO:
> Before this can move on, I think at least the follow things need to be fixed: - There is a memory leak. It looks like the OSSL_CMP_MSG_dup() in transfer_cb() is never freed.

I had included the `dup()` to prevent double frees that would occur during fuzzing.
I was aware of that potential fuzzing mem leak artifact but had hoped that this would have been harmless/ignored.

I've meanwhile solved this issue differently, by guarding all respective `OSSL_CMP_MSG_free()` calls in `crypto/cmp/cmp_client.c` with `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`, which is a bit ugly but more to the point.

DDvO:
> I believe this would have already been fixed by #11448, which unfortunately is not yet reviewed and merged.
> 
> For simplicity and such that we can proceed this fuzzing PR I've cerry-picked the essential commit [1da8127](https://github.com/openssl/openssl/commit/1da812778f9098286b8f069391106d0d2eb05017) of that PR to this one.

Thanks @mattcaswell for quickly reviewing&approving #11448.
Thus I've reverted the cherry-pick done here a little earlier,
keeping just the fuzzing-related suppression of error queue printing that would (re-)appear here when rebasing on that PR after it has been merged (planned for tomorrow noon CEST).

kroeckx:
> BTW, how can I reproduce any (potential future) such crashes you report?

As the README says, you run: cmp-test $file

> I've meanwhile suppressed such error queue output while fuzzing.

I'm not sure how you did that, but if it's only suppressed while fuzzing, the solution seems wrong to me.

(I don't see any update in this PR?)

DDvO:
> (I don't see any update in this PR?)

Sorry, my last push went by mistake not to the right repo. I've just corrected that.

> > BTW, how can I reproduce any (potential future) such crashes you report?
> 
> As the README says, you run: cmp-test $file

Ah, indeed!

This way I've been able to confirm that crash and also that the fix done in commit 03f6514f0fab6 was sufficient: when building from that state (or one of the later ones before the reversal done in commit cbb46bddbd43) the crash disappears.

> > I've meanwhile suppressed such error queue output while fuzzing.
> 
> I'm not sure how you did that, but if it's only suppressed while fuzzing, the solution seems wrong to me.

I'm not sure if I used the right wording in the sentence you quoted here.
What I meant is that I've guarded the error output statements as follows:
```
#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
    OSSL_CMP_CTX_print_errors(ctx);
#endif
```

kroeckx:
So I'm still seeing a memory leak, now it says it's the
d2i_OSSL_CMP_MSG_bio() call.

kroeckx:
Why don't you set up a callback that does nothing with the error messages?

DDvO:
> Why don't you set up a callback that does nothing with the error messages?

Good point! Done.

DDvO:
> So I'm still seeing a memory leak, now it says it's the d2i_OSSL_CMP_MSG_bio() call.

Hmm, I cannot spot manually why this happens and unfortunately can't get any mem leak errors
although I configure with `enable-crypto-mdebug enable-crypto-mdebug-backtrace --debug`
and run the tests with `OPENSSL_DEBUG_MEMORY=on`.

DDvO:
BTW, is there a way of restricting the fuzz tests to CMP? I usually call
```
 make tests TESTS="test_fuzz" V=1
```
which takes pretty long mostly because of the ASN.1 fuzz tests.

kroeckx:
> BTW, is there a way of restricting the fuzz tests to CMP? I usually call
> ```
>  make tests TESTS="test_fuzz" V=1
> ```
> which takes pretty long mostly because of the ASN.1 fuzz tests.

It looks like recipes/99-test_fuzz.t doesn't support that.

You can just run: ./fuzz/cmp-test fuzz/corpora/cmp/

If built using asan, it should report the memory leaks.

kroeckx:
I suggest that you remove your changes to fuzz/asn1.c, I don't
think they are useful, just slow things down.

kroeckx:
So I've updated https://www.roeckx.be/cmp.tar.gz

DDvO:
> I suggest that you remove your changes to fuzz/asn1.c, I don't think they are useful, just slow things down.

All right, removed those.

DDvO:
> So I've updated https://www.roeckx.be/cmp.tar.gz

Thanks, I've included them.

DDvO:
>> BTW, is there a way of restricting the fuzz tests to CMP? I usually call 
``` make tests TESTS="test_fuzz" V=1 ```
which takes pretty long mostly because of the ASN.1 fuzz tests.

> It looks like recipes/99-test_fuzz.t doesn't support that.
You can just run: `./fuzz/cmp-test fuzz/corpora/cmp/`

> If built using asan, it should report the memory leaks.

Good hints!

This way I've been able to find the mem leak. It was a bit tricky because only part of a `msg` structure was not freed. Fixed in `fuzz/cmp.c`.

Since I found that during fuzz tests also the CMP server side produced error queue outputs I disabled them as well (using the same no-op callback as for the client side).

**Comment by kroeckx** [comment]
So that still leaves the:
crypto/x509/x509_cmp.c:172:13: runtime error: member access within null pointer of type 'const X509_NAME' (aka 'const struct X509_name_st')

**Comment by kroeckx** [comment]
I find it strange that you don't do the OSSL_CMP_MSG_free() in
case of FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION. It looks like
an API problem to me.

**Comment by DDvO** [comment]
> I find it strange that you don't do the OSSL_CMP_MSG_free() in case of FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION. It looks like an API problem to me.

I agree that it smells like that. Yet was due to my so far ad-hoc use (with forth and back changes) of the CMP client API in `fuzz/cmp.c`.
After having a close look I've found a much simpler and cleaner solution.
Good that you brought this up again.

**Comment by DDvO** [comment]
> So that still leaves the: 
```
crypto/x509/x509_cmp.c:172:13: runtime error: member access within null pointer of type 'const X509_NAME' (aka 'const struct X509_name_st')
```

Hmm, I have not seen this in the above discussion and cannot reproduce it.
Still I had a look at the source location you provided and I'm pretty sure that this is because some comparison of certs is done where the second does not have an issuer name.

The trivial 'test' cert I provided in `fuzz/cmp.c`:
```
    ctx->oldCert = X509_new(); /* satisfy crm_new() and ossl_cmp_rr_new() */
```
does not have any subject, issuer, serial number, etc, and functions like `X509_cmp()` do not safely deal with that (exceptional, but not outruled) case.
I've provided a tentative fix of a couple such potential NULL situations in `crypto/x509/x509_cmp.c`.
Does this fix the crash you obtained?

**Comment by kroeckx** [comment]
> Does this fix the crash you obtained?

I thought I pointed to such a file before. But I can't reproduce
any of the reported issues I had before with the current version.

I'm wondering why there are still a few:
+#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
     OSSL_CMP_CTX_print_errors(ctx);
+#endif

**Comment by DDvO** [comment]
> I thought I pointed to such a file before. But I can't reproduce any of the reported issues I had before with the current version.

Meanwhile I think I've found the crash you were referring to (it was folded away by this GitHub page): https://www.roeckx.be/crash-57e4fbcd139ec37e87aca0a695c194e68f1abcb7

Then that crash must already have been fixed by the 3GPP-related fixup commit a1e4c8ef8125bfff403d49d60a625f75b3074a79 I had mentioned. This was merged yesterday and I've rebased this PR on it, which seems to explain why you now cannot reproduce it either.

Despite of this, WDYT about the additional tentative fix commit a7539162ce9fee9840f66d68ffc9dc981d30dd42
which makes some functions in `x509_cmp.c` more robust?

**Comment by DDvO** [comment]
> I'm wondering why there are still a few: 
```
+#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
    OSSL_CMP_CTX_print_errors(ctx); 
+#endif
```
Oops, they must have crept in again when rebasing. Got rid of them.

**Comment by kroeckx** [comment]
I think all the changes look good, but I would like someone else to review it.

**Comment by DDvO** [comment]
> I think all the changes look good, but I would like someone else to review it.

Thanks again Kurt for your constant and very constructive feedback on this.

Any other OMC member willing to do the official review?
Maybe the code contains an Easter egg :wink:

**Comment by kroeckx** [comment]
> Any other OMC member willing to do the official review?

It's an OTC member that you need.

**Comment by DDvO** [comment]
Pinging for an OTC member review.

It's actually not very much code, and @kroeckx already made sure that from the fuzzing perspective it makes much sense.

**Comment by mattcaswell** [comment]
Travis failure looks relevant

File: fuzz/README.md [hunk 1]
```diff
@@ -31,14 +31,16 @@ to the `libFuzzer` library file while configuring; this is represented as
...
+    make clean
+    LDCMD=clang++ make -j4
 
 Finally, perform the actual fuzzing:
 
-    $ fuzz/helper.py $FUZZER
+    fuzz/helper.py $FUZZER
 
 where $FUZZER is one of the executables in `fuzz/`.
+It will until run until you stop it.
```
mattcaswell, line 43: s/until run/run until/

fuzz/cmp.c
```diff
@@ -0,0 +1,203 @@
...
+#include <openssl/bio.h>
+#include <openssl/cmp.h>
+#include "../crypto/cmp/cmp_local.h"
+#include <openssl/err.h>
+#include "fuzzer.h"
+#include "rand.inc"
+
+int FuzzerInitialize(int *argc, char ***argv)
+{
+    OPENSSL_init_crypto(OPENSSL_INIT_LOAD_CRYPTO_STRINGS, NULL);
```
mattcaswell, line 23: Not sure why this is necessary.

kroeckx: To make the result more reproducible. All fuzzers do that.

mattcaswell: Ok

fuzz/README.md
DDvO: Fixed

DDvO:
> Looks good. A couple of nits I spotted below.

Pleased to hear!
I've just fixed the remaining (minor) issues. The respective commits can be found in https://github.com/mpeylo/cmpossl/tree/cmp_fuzzer_reviewed

> How do you intend to structure the commits - and with what commit message?

I've meanwhile rebased and partly squashed this PR. 
The new commit structure is my proposal.

DDvO:
Merged - thanks @kroeckx and @mattcaswell

mattcaswell:
This PR seems to have broken Travis. All master builds have been red since it was merged.

DDvO:
Hmm, are you sure the Travis failures are due to merging this PR? 
Can you please provide more detail supporting this?

For instance, #4277 which I rebased to master today,
fails at a point unrelated to CMP and fuzzing:
```
 Looks like you failed 40 tests of 54.
30-test_evp_fetch_prov.t ........... 
Dubious, test returned 40 (wstat 10240, 0x2800)
```
according to https://travis-ci.org/github/openssl/openssl/jobs/677138716

mattcaswell:
I'm looking at this build:

https://travis-ci.org/github/openssl/openssl/jobs/676658702

Which fails in cmp_vfy_test:

````
    # ERROR: (int) 'fixture->expected == ossl_cmp_verify_popo(fixture->msg, fixture->additional_arg)' failed @ test/cmp_vfy_test.c:101
    # [0] compared to [1]
    not ok 2 - test_verify_popo_bad
````

This is from the first build after commit e0331e - and is the first point that travis goes red. I think there are some subsequent other problems from later commits that are unrelated to this.

mattcaswell:
I see test failures locally in cmp_vfy_test if I use `-DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`, as is done in the build in question. Although, strangely, the errors I see are slightly different to the ones in the failing build.

DDvO:
Thx for the details; now I understand what's going on.
I need to disable some at least the above test in case `-DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`
(while I wouldn't have expected that the normal tests are run while that symbol is defined).
I'll provided a fixup PR shortly...

mattcaswell:
Please run the extended tests in your fixup PR - the failing job only gets run when extended tests are enabled.

DDvO:
How to run the extended tests?

kroeckx:
Add a line in the commit message of the last commit that says:
[extended tests]

DDvO:
Done in #11585

PR #11390, opened 2020-03-24
Param builder: make the OSSL_PARAM_BLD APIs public.

paulidale:
The catalyst for this is the difficult of passing BNs through the other OSSL_PARAM APIs.

Two `git mv` and two `sed -i`.


paulidale:
This currently exposes the param builder struct.  It's probably worth the effort making this opaque.  However, the decision to accept OSSL_PARAM_BLD as part of our public APIs is the first step.

include/openssl/param_build.h
```diff
@@ -39,38 +39,38 @@ typedef struct {
     OSSL_PARAM_BLD_DEF params[OSSL_PARAM_BLD_MAX];
 } OSSL_PARAM_BLD;
```
levitte: Are you sure the structures are stable enough for public view for the foreseeable future?

paulidale: See the comment:

> This currently exposes the param builder struct. It's probably worth the effort making this opaque.

So no, I don't think they should be public.  They are probably stable enough to be since OSSL_PARAM isn't going to change.

slontis:
Should we keep OSSL_PARAM_construct_BN()?

include/openssl/param_build.h
paulidale: The question at this point is should the param builder be public or not?

We're using it quite a bit in the providers.

paulidale:
@slontis, that's a different question for a different PR.

Given that nothing in the current tree apart from the tests use `OSSL_PARAM_construct_BN()` and when I've tried it was overly burdensome, I'm in favour.

richsalz:
If this is made publc can we hide the existing lower-level API?  Two ways to do the same thing ... this isn't perl :)

paulidale:
I'd suggest not.

1. the low level API is being used by (our) providers to communicate with libcrypto
1. the low level API involves less copying and avoids memory allocations.

paulidale:
OSSL_PARAM_BLD is now both public and opaque.  Review?

crypto/param_build.c
```diff
@@ -12,11 +12,47 @@
...
 #include <openssl/cryptoerr.h>
 #include <openssl/params.h>
+#include <openssl/types.h>
 #include "internal/cryptlib.h"
 #include "openssl/param_build.h"
 
+/*
+ * The number of OSSL_PARAM elements a builder will allow.
+ */
+#define OSSL_PARAM_BLD_MAX          25
```
t8m: This seems to be a little bit too small (when used for example for the RSA multi-prime support) and as it is now allocated on heap it should be fine to raise it. Unfortunately structs with flexible arrays at the end are C99 construct.

paulidale: 25 should be plenty for the short term -- MP RSA is limited to 5 primes maximum.

I'd prefer to rework the param builder to remove the limit entirely -- that's a larger task and time is short for 3.0.

t8m: > 25 should be plenty for the short term -- MP RSA is limited to 5 primes maximum.

The providers/implementations/keymgmt/rsa_kmgmt.c supports up to 10 primes actually so that would be over the 25 limit for all the primes, exponents and coefficients.

> I'd prefer to rework the param builder to remove the limit entirely -- that's a larger task and time is short for 3.0.

If the OSSL_PARAM_BLD_new() api required a maximum and there was no "dynamic reallocation" support the change could be pretty minimal I suppose.

But all this is just a note. I do not block the PR.

paulidale: I bit the bullet and converted the build to use safestack.  Limit removed :)

paulidale:
Travis problem is a full disc.

NEWS.md
```diff
@@ -31,6 +31,7 @@ OpenSSL 3.0
   * enable-crypto-mdebug and enable-crypto-mdebug-backtrace were mostly
     disabled; the project uses address sanitize/leak-detect instead.
   * Added OSSL_SERIALIZER, a generic serializer API.
+  * Added OSSL_PARAM_BLD, an easier to use API to OSSL_PARAM.
```
levitte: Maybe a CHANGES.md entry as well, with a bit more fleshing out?

paulidale: Apart from RTFM, what else needs to be said?

levitte: A short summary of the FM is often seen.  You might want to look at some entries in said file for inspiration.

paulidale: Added a note the CHANGES.md

paulidale:
Given the large number of other outstanding PRs impacted by this, should this be considered "urgent" or should they?  It's mostly @slontis's PRs :)

crypto/param_build.c
```diff
@@ -66,83 +101,107 @@ static int param_push_num(OSSL_PARAM_BLD *bld, const char *key,
...
+            OPENSSL_free(r);
+            r = NULL;
+        }
+    }
+    return r;
+}
+
+static void free_all_params(OSSL_PARAM_BLD *bld)
+{
+    int i, n = sk_OSSL_PARAM_BLD_DEF_num(bld->params);
```
romen: Nit: This assignment should happen after variable declarations.

crypto/param_build.c
```diff
@@ -244,12 +303,12 @@ static OSSL_PARAM *param_bld_convert(OSSL_PARAM_BLD *bld, OSSL_PARAM *param,
                                      OSSL_PARAM_BLD_BLOCK *blk,
                                      OSSL_PARAM_BLD_BLOCK *secure)
 {
-    size_t i;
+    int i, num = sk_OSSL_PARAM_BLD_DEF_num(bld->params);
```
romen: Nit: This assignment should happen after variable declarations, unless this is a macro statically resolved at compile time.

crypto/param_build.c
```diff
@@ -288,11 +347,12 @@ static OSSL_PARAM *param_bld_convert(OSSL_PARAM_BLD *bld, OSSL_PARAM *param,
     return param + i;
 }
 
-OSSL_PARAM *ossl_param_bld_to_param(OSSL_PARAM_BLD *bld)
+OSSL_PARAM *OSSL_PARAM_BLD_to_param(OSSL_PARAM_BLD *bld)
 {
     OSSL_PARAM_BLD_BLOCK *blk, *s = NULL;
     OSSL_PARAM *params, *last;
-    const size_t p_blks = bytes_to_blocks((1 + bld->curr) * sizeof(*params));
+    const int num = sk_OSSL_PARAM_BLD_DEF_num(bld->params);
```
romen: Nit: This assignment should happen after variable declarations, unless this is a compile-time value.

crypto/rsa/rsa_ameth.c
```diff
@@ -1084,7 +1084,7 @@ static int rsa_pkey_export_to(const EVP_PKEY *from, void *to_keydata,
                               EVP_KEYMGMT *to_keymgmt)
 {
     RSA *rsa = from->pkey.rsa;
-    OSSL_PARAM_BLD tmpl;
+    OSSL_PARAM_BLD *tmpl = OSSL_PARAM_BLD_new();
```
romen: Nit: This assignment should happen after variable declarations.

doc/man3/OSSL_PARAM_BLD.pod
```diff
@@ -143,37 +141,39 @@ For both, the predefined key variables are:
 This example shows how to create an OSSL_PARAM array that contains an RSA
 private key.
 
-    OSSL_PARAM_BLD bld;
+    OSSL_PARAM_BLD *bld = OSSL_PARAM_BLD_new();
```
romen: Nit: This assignment should happen after variable declarations.

Well, maybe, this is just a code example..

doc/man3/OSSL_PARAM_BLD.pod
```diff
@@ -143,37 +141,39 @@ For both, the predefined key variables are:
...
-    ossl_param_bld_free(params);
+    OSSL_PARAM_BLD_free_params(params);
 
 =head2 Example 2
 
 This example shows how to create an OSSL_PARAM array that contains an RSA
 public key.
 
-    OSSL_PARAM_BLD bld;
+    OSSL_PARAM_BLD *bld = OSSL_PARAM_BLD_new();
```
romen: Nit: This assignment should happen after variable declarations.

Well, maybe, this is just an example..

providers/implementations/keymgmt/dsa_kmgmt.c
```diff
@@ -153,26 +153,26 @@ static int dsa_export(void *keydata, int selection, OSSL_CALLBACK *param_cb,
                       void *cbarg)
 {
     DSA *dsa = keydata;
-    OSSL_PARAM_BLD tmpl;
+    OSSL_PARAM_BLD *tmpl = OSSL_PARAM_BLD_new();
```
romen: Nit: This assignment should happen after variable declarations.

test/param_build_test.c
```diff
@@ -10,13 +10,13 @@
...
 #include <openssl/params.h>
-#include "internal/param_build.h"
+#include "openssl/param_build.h"
 #include "internal/nelem.h"
 #include "testutil.h"
 
 static int template_public_test(void)
 {
-    OSSL_PARAM_BLD bld;
+    OSSL_PARAM_BLD *bld = OSSL_PARAM_BLD_new();
```
romen: Nit: This assignment should happen after variable declarations.

test/param_build_test.c
```diff
@@ -106,7 +107,7 @@ static int template_private_test(void)
 {
     static int data1[] = { 2, 3, 5, 7, 11, 15, 17 };
     static unsigned char data2[] = { 2, 4, 6, 8, 10 };
-    OSSL_PARAM_BLD bld;
+    OSSL_PARAM_BLD *bld = OSSL_PARAM_BLD_new();
```
romen: Nit: This assignment should happen after variable declarations.

test/param_build_test.c
```diff
@@ -184,15 +185,60 @@ static int template_private_test(void)
...
     BN_free(bn);
     BN_free(bn_res);
     return res;
 }
 
+static int builder_limit_test(void)
+{
+    const int n = 100;
+    char names[100][3];
+    OSSL_PARAM_BLD *bld = OSSL_PARAM_BLD_new();
```
romen: Nit: This assignment should happen after variable declarations.

romen:
Another note: given that we can expect external providers to be among the main users of this API, shouldn't this API and also the low level OSSL_PARAM API provided to providers as pointers in the dispatch table passed during provider initialization?

Without exposing the 2 APIs in the dispatch table we are basically forcing external providers to link against libcrypto or to rewrite their own wrappers for handling OSSL_PARAMs (which is definitely quite error prone and not entirely trivial in effort).

test/evp_pkey_provided_test.c
```diff
@@ -596,7 +596,7 @@ static int test_fromdata_ec(void)
     int ret = 0;
     EVP_PKEY_CTX *ctx = NULL;
     EVP_PKEY *pk = NULL, *copy_pk = NULL;
-    OSSL_PARAM_BLD bld;
+    OSSL_PARAM_BLD *bld = OSSL_PARAM_BLD_new();
```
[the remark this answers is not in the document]
paulidale: No.  Why?

paulidale:
External providers are fine to link against this or not.  The underlying OSSL_PARAM mechanism is completely public -- that's what gets transferred.  This and the OSSL_PARAM_[sg]et* (etc) calls are helpers.  Our providers can use them because either:

1. they link against libcrypto or
1. the FIPS provider pulls a copy of the code inside its boundary.

I don't see a need for these to go via the callback dispatch table.  As mentioned, the underlying OSSL_PARAM structure needs to be public so both sides can handle it -- the dispatch table has nothing to do with that.  If it did, we'd have a much cleaner and more usable API...

test/evp_pkey_provided_test.c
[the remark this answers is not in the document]
paulidale: We allow function calls during variable initialisation.  It isn't always appropriate but it is allowed.

romen:
The OSSL_PARAM struct while public is not really a user-friendly API, and handling params by directly manipulating the struct has a lot of caveats.

If our vision is that external providers could be compiled without linking to a specific version of libcrypto so that they don't need to be recompiled/relinked when libcrypto is updated, we are basically forcing each external provider to rewrite the equivalent of OSSL_PARAM_[gs]et utility functions and in some cases of the OSSL_PARAM_BLD functions.

If we felt the need to use these wrappers to have some form of maintenability when dealing with OSSL_PARAMs we can expect the same to be true for external providers as well.

I believe we have the wrappers in place because we recognize that direct manipulation of the OSSL_PARAM structs is cumbersome and error prone, leads to a lot of unnecessary repetition and in general scores very low in any usability and best-practices metric.

I see the value of having the option of manipulating the structs directly, especially when/if the external provider is very "alien" compared to libcrypto (e.g. written in a language different than C), but I don't think we should endorse it as the primary way of communication between external providers and the framework.

paulidale:
The current situation is the outcome of the intentional design effort.  Param builder is simply a helper over the top of the raw OSSL_PARAM structure.  It is no more than an assistant and no provider is mandated to use it -- as you wrote, they can implement this themselves or do something different that matches the definition.

That we're using it internally in several places is an indication that it might be use more widely.  If you disagree, feel free to down vote this and I'll close it.

The lowest common denominator is and always will be OSSL_PARAM.  The intention is that this will work across versions pretty much forever.

romen:
I don't want to stop this PR, I just want to bring it one step closer to fruition by provider authors.

I will open a separate PR to try and discuss the matter of making these functions available to providers through the dispatch table without linking to libcrypto.

test/evp_pkey_provided_test.c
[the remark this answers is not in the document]
romen: Never mind then!

slontis:
Pauli - do you have any idea what merge chaos this will cause in my keygen PR's?

paulidale:
Having slept on this, it does seem sensible for this to be public **and** available via the callbacks.  

I do descending sense of dread as to what this invokes.

The first is BN.  The BN structure needs to remain private and different providers could use differently implementations of BNs. Reconciling this with wanting calls to make BN easy is a problem will be difficult.

The second is memory ownership.  By making them callbacks, this could get blurred.  Some of the helpers will allocate memory if none is given.

paulidale:
> Pauli - do you have any idea what merge chaos this will cause in my keygen PR's?

Lots of fun I expect :)

Looking at the changes involved per file, it isn't bad.  Add a `*` in the variable definition.  Change the `_init` call to `_new` with an assignment.  Capitalise the start of the function names, possibly remove a '&' in each call and add the `OSSL_PARAM_BLD_free` call.

paulidale:
I'll offer to fixup any relevant PRs if that helps.

romen:
> Having slept on this, it does seem sensible for this to be public **and** available via the callbacks.
> 
> I do descending sense of dread as to what this invokes.
> 
> The first is BN. The BN structure needs to remain private and different providers could use differently implementations of BNs. Reconciling this with wanting calls to make BN easy is a problem will be difficult.
> 
> The second is memory ownership. By making them callbacks, this could get blurred. Some of the helpers will allocate memory if none is given.

Regarding the BN builder, I am thinking that maybe that is the one that does not need to be exposed via callback too: if a provider author is using BIGNUM from libcrypto the provider will inevitably be linked against libcrypto anyway, so the public BLD API can be used, if they are using some other multiprecision solution, they will need to write their own auxiliary functions anyway to massage such objects into an OSSL_PARAM.

But having tried to handle get_params and keymgmt in an external provider without linking to libcrypto, I am quite convinced that both the BLD and the locate/get/set functions for handling at least the non-openssl-specific types should probably exposed also as callbacks in the exported dispatch table.
It is not a requirement, because as you say one could manipulate the OSSL_PARAMs directly, but doing so shouldn't be the only option if we care that the library has a good usability for provider authors, and forcing every provider author to either link against libcrypto or write their own wrappers seems quite harsh.

paulidale:
Merged to master.  Thanks for the reviews.

@romen, I'm not sure what to do with BIGNUM's, they are the main problem here.

levitte:
> I am quite convinced that both the BLD and the locate/get/set functions for handling at least the non-openssl-specific types should probably exposed also as callbacks in the exported dispatch table.

Not sure I understand why.  In that case, linking with libcrypto should be workable, should it not?  We are, after all, handling an *open structure that shall remain open and unchanging', so there should be no dependence on libcrypto version for that stuff...

richsalz:
> @romen, I'm not sure what to do with BIGNUM's, they are the main problem here.

Why do you ahve to solve it in the 3.0 release?

t8m:
The problem of third party providers not wanting to link libcrypto but having to duplicate this code could be also solved by having a separate simple (possibly statically linked only) library that would provide an implementation of this and other utility things that would be useful to providers.

romen:
@t8m this was an alternative I was considering, but from the project side this should be done minimizing duplication of code.

Also, what about licensing? If such a utility static library was release under Apache, would there be restrictions on external binary providers that would statically link that code into them?

romen:
> > @romen, I'm not sure what to do with BIGNUM's, they are the main problem here.
> 
> Why do you ahve to solve it in the 3.0 release?

@richsalz I believe the provider API (and what kind of callbacks are available in the imported dispatch table) should be already stable enough within 3.0: otherwise we will force external providers to be a mess of ifdefity to do essential things differently depending on the release version of openssl.

t8m:
> Also, what about licensing? If such a utility static library was release under Apache, would there be restrictions on external binary providers that would statically link that code into them?

Apache license is not a copyleft license so it does not enforce you to apply it to all the derivative work.

paulidale:
We don't have to solve BIBNUM for 3.0.

PR #11391, opened 2020-03-24
EC: Refactor ec_curve_name2nid() to accept NIST curve names

levitte:
We can find no reason why everyone should have to call both
EC_curve_nist2nid() and ec_curve_name2nid() to find the NID for a
name, and it's too easy to forget EC_curve_nist2nid(), so we make life
simpler.

One could argue that FIPS only allows a limited set of names, but that
now gets handled internally, and those who really want to be really
sure to only get the NIST names can still do so with EC_curve_nist2nid()

### DIFF
--- a/crypto/ec/ec_curve.c
+++ b/crypto/ec/ec_curve.c
@@ -3117,25 +3117,65 @@ static const ec_list_element curve_list[] = {
 
 #define curve_list_length OSSL_NELEM(curve_list)
 
+static const ec_list_element *ec_curve_nid2curve(int nid)
+{
+    size_t i;
+
+    if (nid <= 0)
+        return NULL;
+
+    for (i = 0; i < curve_list_length; i++) {
+        if (curve_list[i].nid == nid)
+            return &curve_list[i];
+    }
+    return NULL;
+}
+
+static const ec_list_element *ec_curve_name2curve(const char *name)
+{
+    size_t i;
+
+    for (i = 0; i < curve_list_length; i++) {
+        if (strcasecmp(curve_list[i].name, name) == 0)
+            return &curve_list[i];
+    }
+    return NULL;
+}
+
 const char *ec_curve_nid2name(int nid)
 {
-    int i;
+    /*
+     * TODO(3.0) Figure out if we should try to find the nid with
+     * EC_curve_nid2nist() first, i.e. make it a priority to return
+     * NIST names if there is one for the NID.  This is related to
+     * the TODO comment in ec_curve_name2nid().
+     */
+    const ec_list_element *curve = ec_curve_nid2curve(nid);
 
-    for (i = 0; i < (int)curve_list_length; i++) {
-        if (curve_list[i].nid == nid)
-            return curve_list[i].name;
-    }
+    if (curve != NULL)
+        return curve->name;
     return NULL;
 }
 
 int ec_curve_name2nid(const char *name)
 {
-    int i;
+    const ec_list_element *curve = NULL;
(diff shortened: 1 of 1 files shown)


levitte:
A question is if `ec_curve_nid2name` and `ec_curve_name2nid` should be made public (i.e. become `EC_curve_nid2name` and `EC_curve_name2nid`).  After all, they are part of constructing EC_KEYs, and constructors are not deprecated, so...

paulidale:
I'm sceptical that they should be public.

crypto/ec/ec_curve.c
```diff
@@ -3117,25 +3117,65 @@ static const ec_list_element curve_list[] = {
...
 
-const char *ec_curve_nid2name(int nid)
+static const ec_list_element *ec_curve_nid2curve(int nid)
 {
     int i;
 
+    if (nid <= 0)
+        return NULL;
+
     for (i = 0; i < (int)curve_list_length; i++) {
```
mattcaswell: Perhaps just make `i` a `size_t` and get rid of the cast?

crypto/ec/ec_curve.c
```diff
@@ -3117,25 +3117,65 @@ static const ec_list_element curve_list[] = {
...
     }
     return NULL;
 }
 
-int ec_curve_name2nid(const char *name)
+static const ec_list_element *ec_curve_name2curve(const char *name)
 {
     int i;
 
     for (i = 0; i < (int)curve_list_length; i++) {
```
mattcaswell: And again

crypto/ec/ec_curve.c
levitte: Done

crypto/ec/ec_curve.c
levitte: Done

levitte:
> I'm sceptical that they should be public.

Then we need to figure out another way to deliver a set of construction routines that aren't crippled.  Or deprecate constructors.

Choose.

paulidale:
3. Live with the current state of _crippled_ construction routines.

If they really are that crippled, nobody will be using them and they can be deprecated.

paulidale:
Travis is a timeout and not relevant.

levitte:
> If they really are that crippled, nobody will be using them and they can be deprecated.

The crippling part is that, as things currently stand, you can only specify the NIST names by name (string) using public function.

Another viable solution would be to have the new constructors, `EC_GROUP_new_by_curve_name_ex()` and `EC_KEY_new_by_curve_name_ex()`, take a string instead of an int, which would be more fitting with their names anyway (how anyone could think that it would be a good idea to use `name` in the name of a function that takes a nid is beyond me).  Considering we're doing exactly that elsewhere (for example `EVP_PKEY_CTX_new_from_name()`), it would be quite fitting.

h-vetinari:
This PR didn't make it into [the 3.0 project](https://github.com/openssl/openssl/projects/2) somehow.

levitte:
> This PR didn't make it into [the 3.0 project](https://github.com/openssl/openssl/projects/2) somehow.

That's because it's no specifically New Core or FIPS

h-vetinari:
Well, who am I to argue? 😅
I just thought that as a requirement for PRs within that project, it would necessarily be part of the project as well. Never mind then, sorry for the noise.

levitte:
If you go to the [projects tab](https://github.com/openssl/openssl/projects), that text might make it a bit clearer.  Otherwise, feel free to suggest a better clarification.

PR #11375, opened 2020-03-21
[WIP, pending on #11374] EVP: Downgrade keys rather than upgrade

levitte:
Upgrading EVP_PKEYs from containing legacy keys to containing provider
side keys proved to be risky, with a number of unpleasant corner
cases, and with functions like EVP_PKEY_get0_DSA() failing
unexpectedly.

We therefore change course, and instead of upgrading legacy internal
keys to provider side internal keys, we downgrade provider side
internal keys to legacy ones.  To be able to do this, we add
|import_from| and make it a callback function designed for
evp_keymgmt_export().

This means that evp_pkey_upgrade_to_provider() is replaced with
evp_pkey_downgrade().

-----

This also clarifies the interaction between certain key fields in EVP_PKEY.

PR #11394, opened 2020-03-24
PROV: Fix EC_KEY exporters to allow domain parameter keys

levitte:
The provider key export functions for EC_KEY assumed that a public key
is always present, and would fail if not.  This blocks any attempt to
export a key structure with only domain parameters.

This is similar to earlier work done in EVP_PKEY_ASN1_METHODs.

### DIFF
--- a/crypto/ec/ec_backend.c
+++ b/crypto/ec/ec_backend.c
@@ -63,7 +63,7 @@ int ec_set_param_ecdh_cofactor_mode(EC_KEY *ec, const OSSL_PARAM *p)
  */
 int ec_key_fromdata(EC_KEY *ec, const OSSL_PARAM params[], int include_private)
 {
-    const OSSL_PARAM *param_priv_key, *param_pub_key;
+    const OSSL_PARAM *param_priv_key = NULL, *param_pub_key = NULL;
     BN_CTX *ctx = NULL;
     BIGNUM *priv_key = NULL;
     unsigned char *pub_key = NULL;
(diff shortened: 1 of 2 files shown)


levitte:
The previous similar work is #11374, now merged

providers/implementations/keymgmt/ec_kmgmt.c
```diff
@@ -165,27 +165,28 @@ int params_to_key(EC_KEY *ec, const OSSL_PARAM params[], int include_private)
...
+    if (param_priv_key != NULL && param_pub_key == NULL)
+        goto err;
+
+    if (param_pub_key == NULL)
+        if (!OSSL_PARAM_get_octet_string(param_pub_key,
+                                         (void **)&pub_key, 0, &pub_key_len)
             || (pub_point = EC_POINT_new(ecg)) == NULL
-            || !EC_POINT_oct2point(ecg, pub_point,
-                                   pub_key, pub_key_len, ctx))
+            || !EC_POINT_oct2point(ecg, pub_point, pub_key, pub_key_len, ctx)p)
```
mattcaswell: What is this stray `p` on the end

providers/implementations/keymgmt/ec_kmgmt.c
```diff
@@ -165,27 +165,28 @@ int params_to_key(EC_KEY *ec, const OSSL_PARAM params[], int include_private)
...
-     */
-    if (param_pub_key == NULL
-            || !OSSL_PARAM_get_octet_string(param_pub_key,
-                                            (void **)&pub_key, 0, &pub_key_len)
+
+    /* OpenSSL decree: If there's a private key, there must be a public key */
+    if (param_priv_key != NULL && param_pub_key == NULL)
+        goto err;
+
+    if (param_pub_key == NULL)
```
mattcaswell: Shouldn't this be `!=`

mattcaswell:
The title of this PR mentions DSA - but I see no DSA changes

providers/implementations/keymgmt/ec_kmgmt.c
levitte: Fixedd

providers/implementations/keymgmt/ec_kmgmt.c
levitte: Fixed

levitte:
> The title of this PR mentions DSA - but I see no DSA changes

Thanks for the reminder...  yeah it turned out that DSA was already doing it right!

levitte:
Ping

levitte:
Some of the work done here happened as part of #11375, but some still remained, so this PR isn't entirely lost yet.

PING

crypto/ec/ec_backend.c
```diff
@@ -76,24 +76,25 @@ int ec_key_fromdata(EC_KEY *ec, const OSSL_PARAM params[], int include_private)
     if (ecg == NULL)
         return 0;
 
-    param_priv_key =
-        OSSL_PARAM_locate_const(params, OSSL_PKEY_PARAM_PRIV_KEY);
     param_pub_key =
         OSSL_PARAM_locate_const(params, OSSL_PKEY_PARAM_PUB_KEY);
+    if (include_private)
+        param_priv_key =
```
slontis: you need to set this to NULL when it is inited..

levitte: Thanks!  Fixed

PR #11396, opened 2020-03-25
Params: Add OSSL_PARAM_set_BN_pad() function.

paulidale:
This allows constant width number to be passed more easily across the
provider boundary without compromising the constness of the data_size field.

### DIFF
--- a/crypto/params.c
+++ b/crypto/params.c
@@ -608,13 +608,15 @@ OSSL_PARAM OSSL_PARAM_construct_size_t(const char *key, size_t *buf)
 int OSSL_PARAM_get_BN(const OSSL_PARAM *p, BIGNUM **val)
 {
     BIGNUM *b;
+    size_t len;
 
     if (val == NULL
         || p == NULL
         || p->data_type != OSSL_PARAM_UNSIGNED_INTEGER)
         return 0;
 
-    b = BN_native2bn(p->data, (int)p->data_size, *val);
+    len = p->return_size > 0 ? p->return_size : p->data_size;
+    b = BN_native2bn(p->data, (int)len, *val);
     if (b != NULL) {
         *val = b;
         return 1;
@@ -623,6 +625,13 @@ int OSSL_PARAM_get_BN(const OSSL_PARAM *p, BIGNUM **val)
 }
 
 int OSSL_PARAM_set_BN(OSSL_PARAM *p, const BIGNUM *val)
+{
+    if (p == NULL)
+        return 0;
+    return OSSL_PARAM_set_BN_pad(p, val, p->data_size);
+}
+
+int OSSL_PARAM_set_BN_pad(OSSL_PARAM *p, const BIGNUM *val, size_t sz)
 {
     size_t bytes;
 
@@ -638,11 +647,14 @@ int OSSL_PARAM_set_BN(OSSL_PARAM *p, const BIGNUM *val)
 
     bytes = (size_t)BN_num_bytes(val);
     p->return_size = bytes;
+    if (sz < bytes)
+        return 0;
+
     if (p->data == NULL)
         return 1;
-    if (p->data_size >= bytes) {
-        p->return_size = p->data_size;
-        return BN_bn2nativepad(val, p->data, p->data_size) >= 0;
+    if (p->data_size >= sz) {
+        p->return_size = sz;
+        return BN_bn2nativepad(val, p->data, sz) >= 0;
     }
     return 0;
 }
--- a/test/params_api_test.c
+++ b/test/params_api_test.c
@@ -401,6 +401,34 @@ err:
     return ret;
 }
 
+static int test_param_bignum_pad(void)
+{
+    unsigned char buf[MAX_LEN];
+    OSSL_PARAM param;
+    BIGNUM *bn, *out = NULL;
+    size_t i;
+    int ret = 0;
+
+    param = OSSL_PARAM_construct_BN("bn", buf, sizeof(buf));
+    if (!TEST_ptr(bn = BN_new())
+            || !TEST_ptr(out = BN_new())
+            || !TEST_true(BN_set_word(bn, 54321)))
+        goto err;
+    for (i = sizeof(BN_ULONG); i < MAX_LEN; i++) {
+        if (!TEST_true(OSSL_PARAM_set_BN_pad(&param, bn, i))
+                || !TEST_size_t_eq(param.return_size, i)
+                || !TEST_true(OSSL_PARAM_get_BN(&param, &out))
+                || !TEST_BN_eq(bn, out)
+                || !TEST_true(BN_set_word(out, 999)))
+            goto err;
+    }
+    ret = 1;
+err:
+    BN_free(out);
+    BN_free(bn);
+    return ret;
+}
+
 static int test_param_real(void)
 {
     double p;
@@ -575,6 +603,7 @@ int setup_tests(void)
     ADD_ALL_TESTS(test_param_int64, OSSL_NELEM(raw_values));
     ADD_ALL_TESTS(test_param_uint64, OSSL_NELEM(raw_values));
     ADD_ALL_TESTS(test_param_bignum, OSSL_NELEM(raw_values));
+    ADD_TEST(test_param_bignum_pad);
     ADD_TEST(test_param_real);
     ADD_TEST(test_param_construct);
     return 1;
--- a/doc/man3/OSSL_PARAM_int.pod
+++ b/doc/man3/OSSL_PARAM_int.pod
@@ -25,9 +25,9 @@ OSSL_PARAM_get_octet_ptr,
 OSSL_PARAM_set_double, OSSL_PARAM_set_int, OSSL_PARAM_set_int32,
 OSSL_PARAM_set_int64, OSSL_PARAM_set_long, OSSL_PARAM_set_size_t,
 OSSL_PARAM_set_uint, OSSL_PARAM_set_uint32, OSSL_PARAM_set_uint64,
-OSSL_PARAM_set_ulong, OSSL_PARAM_set_BN, OSSL_PARAM_set_utf8_string,
-OSSL_PARAM_set_octet_string, OSSL_PARAM_set_utf8_ptr,
-OSSL_PARAM_set_octet_ptr
+OSSL_PARAM_set_ulong, OSSL_PARAM_set_BN, OSSL_PARAM_set_BN_pad,
+OSSL_PARAM_set_utf8_string, OSSL_PARAM_set_octet_string,
+OSSL_PARAM_set_utf8_ptr, OSSL_PARAM_set_octet_ptr
 - OSSL_PARAM helpers
 
 =head1 SYNOPSIS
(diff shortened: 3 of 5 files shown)


crypto/params.c
```diff
@@ -638,11 +645,14 @@ int OSSL_PARAM_set_BN(OSSL_PARAM *p, const BIGNUM *val)
 
     bytes = (size_t)BN_num_bytes(val);
     p->return_size = bytes;
+    if (sz < bytes)
```
slontis: I think return_size should only be set on success

crypto/params.c
```diff
@@ -638,11 +645,14 @@ int OSSL_PARAM_set_BN(OSSL_PARAM *p, const BIGNUM *val)
...
+        return 0;
+
     if (p->data == NULL)
         return 1;
-    if (p->data_size >= bytes) {
-        p->return_size = p->data_size;
-        return BN_bn2nativepad(val, p->data, p->data_size) >= 0;
+    if (p->data_size >= sz) {
+        p->return_size = sz;
```
slontis: And here also

crypto/params.c
paulidale: It is typically set on failure too as an indication as to what space is required.

slontis: 0 is returned for multiple different types of error.
I don't think it is a good idea to assume it was a not enough space error.
If they want to figure out how much space they need - then the NULL data argument should be passed instead.

levitte: I think that it really doesn't matter. If there's an error, the caller should act on the error, not try to second guess that maybe just maybe it was about space. There's is a mechanism for finding out the space requirement, by setting the *data* field to NULL. That doesn't change even if we do set *return_size* on error.

Granted, though, that this particular check could be positioned before the assignment of `p->return_size`

crypto/params.c
levitte: Here's a spot where I think that's taking purity too far.

crypto/params.c
paulidale: Do we want the two `OSSL_PARAM_set_BN` calls to be inconsistent with all the rest of the set calls?

crypto/params.c
paulidale: This is **exactly** where the _return_size_ **must** be set.

This is the number of bytes of BN that are being returned.  With the padding, the allocated space in the OSSL_PARAM might not be fully utilised and the get function needs to know this.  Witness the change to `OSSL_PARAM_get_BN()`.

crypto/params.c
```diff
@@ -608,13 +608,15 @@ OSSL_PARAM OSSL_PARAM_construct_size_t(const char *key, size_t *buf)
...
     BIGNUM *b;
+    size_t len;
 
     if (val == NULL
         || p == NULL
         || p->data_type != OSSL_PARAM_UNSIGNED_INTEGER)
         return 0;
 
-    b = BN_native2bn(p->data, (int)p->data_size, *val);
+    len = p->return_size > 0 ? p->return_size : p->data_size;
```
levitte: You do remember that this is dicy, right? The "other" end may not have initialized *return_size* at all...

levitte: It would be better to have a separate function for the OSSL_PARAM array owner. I haven't been able to figure out a suitable name, though, and mostly figured it would just be an alias for `BN_native2bn` in the end...

crypto/params.c
paulidale: I'll also note that by this point, the only possible error is 'not enough space'.  It seems reasonable to return the required size in this case.

All other errors set _return_size_ to zero, so they can be distinguished.

crypto/params.c
paulidale: Please suggest an alternative way for a OSSL_PARAM with a 50 byte buffer to accept a 20 byte padded BN?

I could have `OSSL_PARAM_set_BN_pad()` zero fill the OSSL_PARAM buffer but that's doesn't seem like constant time which is the entire point of wanting this new function.

crypto/params.c
levitte: Hmmmm, so this is an error, not so much between the OSSL_PARAM owner and the responder, but within the responder.
That's... interesting.

crypto/params.c
levitte: I agree that is the right spot. My comment was to what @slontis wrote

paulidale: Apologies, I thought the comment was directed at the change :(

levitte: No prob, this is the usual one thread depth problem.  I could have specified it in my comment

crypto/params.c
levitte: I have *no* idea what you're commenting on here.  This is `OSSL_PARAM_get_bn()`, which is, or at least was designed for the "receiving end" (i.e. the type of function that receives an `OSSL_PARAM` pointer as an argument) of the `OSSL_PARAM` array.  As such, it should only rely on *data_size*.

I assume you're again trying to redesign it for double purposes, i.e. to become something for the "sending end" (i.e. the `OSSL_PARAM` owner) to use.  I'm only pointing out that there's no guarantee that the "sending end" has zeroed *return_size*, and the "receiving end" shouldn't have to depend on that.

test/params_api_test.c
```diff
@@ -401,6 +401,34 @@ static int test_param_bignum(int n)
...
+
+    param = OSSL_PARAM_construct_BN("bn", buf, sizeof(buf));
+    if (!TEST_ptr(bn = BN_new())
+            || !TEST_ptr(out = BN_new())
+            || !TEST_true(BN_set_word(bn, 54321)))
+        goto err;
+    for (i = sizeof(BN_ULONG); i < MAX_LEN; i++) {
+        if (!TEST_true(OSSL_PARAM_set_BN_pad(&param, bn, i))
+                || !TEST_size_t_eq(param.return_size, i)
+                || !TEST_true(OSSL_PARAM_get_BN(&param, &out))
```
levitte: I see where your change of `OSSL_PARAM_get_BN` comes from.  It it were me, I would just `!TEST_true(BN_native2bn(buf, param->return_size, out))`

paulidale: This complete circumvents the problem......

The param owner is permitted to call `OSSL_PARAM_set_BN` with an undersized BN and then pass it over.

levitte: `OSSL_PARAM_set_BN_pad()` is designed for the "receiving end", just as all the other `set` and `get` functions, is it not?

The `OSSL_PARAM` owner calls `OSSL_PARAM_construct_BN()` with the appropriate buffer size.

See, I think you're confusing yourself with this test, where you're clearly mixing the "sending end" and the "receiving end"...  which is fine, *if you're clear about which call belongs in what end*.  I would suggest dividing up this test in two parts to make things clearer for yourself.

test/params_api_test.c
```diff
@@ -401,6 +401,34 @@ static int test_param_bignum(int n)
...
 
+static int test_param_bignum_pad(void)
+{
+    unsigned char buf[MAX_LEN];
+    OSSL_PARAM param;
+    BIGNUM *bn, *out = NULL;
+    size_t i;
+    int ret = 0;
+
+    param = OSSL_PARAM_construct_BN("bn", buf, sizeof(buf));
```
levitte: This is on the "sending end"

test/params_api_test.c
```diff
@@ -401,6 +401,34 @@ static int test_param_bignum(int n)
...
+    if (!TEST_ptr(bn = BN_new())
+            || !TEST_ptr(out = BN_new())
+            || !TEST_true(BN_set_word(bn, 54321)))
+        goto err;
+    for (i = sizeof(BN_ULONG); i < MAX_LEN; i++) {
+        if (!TEST_true(OSSL_PARAM_set_BN_pad(&param, bn, i))
```
levitte: This is on the "receiving end"

test/params_api_test.c
levitte: This is no the "sending end"

paulidale: I view `OSSL_PARAM_set_BN_pad()` as being applicable on **both** the sending and receiving sides.  Unlike all the other types, we cannot create an OSSL_PARAM containing a BN without having extra storage.  The _to native_ conversion mandates a memory copy, this isn't true for the string types or the basic numeric types -- these can be their own storage.

I think it quite reasonable for the sender to use `OSSL_PARAM_construct_BN()` to create a PARAM and `OSSL_PARAM_set_BN_pad()` to populate it.

I'm beginning to think that a BN PARAM type might avoid a lot of the gymnastics we're being forced through.

levitte: > I'm beginning to think that a BN PARAM type might avoid a lot of the gymnastics we're being forced through.

How?  Are you going to just point directly to the BIGNUM, i.e. force everyone to use that specific storage format for big numbers?  In that case, *we need to have the format very well documented*, and the BIGNUM type will have to continue to be an open and unchanging structure for the foreseeable future.  That, or lock every in on our BIGNUM implementation.

Whichever way we turn with BIGNUMs, there are consequences to consider.

levitte: Of course, "lock in" is relative.  When we made the nCipher (CHIL) engine, we had to deal with exactly that, the result being a couple of conversion routines for juggling big numbers back and forth between OpenSSL's BN and nCipher's MPI.  Look for `mpi2bn` and `bn2mpi` in https://github.com/engine-corner/engine-chil/blob/master/e_chil.c to see

levitte: However, back to `OSSL_PARAM_set_BN()` / `OSSL_PARAM_set_BN_pad()` for the "sending" end, I'll suggest again to make it a separate function.  I've said this what feels like a gazillion times already.  That's the only way I can see avoiding dicy corners.

paulidale: I was thinking of exposing the BN_ULONG array as the passing mechanism.  Plus the length of the array.  We don't need the sign.  The flags might be a loss.

I'm tempted to ditch this PR as too difficult.

levitte:
i'm hearing your thoughts on giving this up, @paulidale, but before you do, I wonder if you might consider what determines the possibility to pad at all, and all I can see is that this lies with the `OSSL_PARAM` owner (the "sending" end).  In other words, the difference between `OSSL_PARAM_set_BN()` and `OSSL_PARAM_set_BN_pad()` could be this:

-   `OSSL_PARAM_set_BN()` only populates as much of `p->data_size` bytes as is determined by `BN_num_bytes()`, so it ends up calling `BN_bn2nativepad(bn, p->data, BN_num_bytes())`, and sets `p->return_size = BN_num_bytes()`.
-   `OSSL_PARAM_set_BN_pad()` always populates all of `p->data_size` bytes regardless, so it ends up calling `BN_bn2nativepad(bn, p->data, p->data_size)`, and sets `p->return_size = p->data_size`.

So basically, leave the power of determining the buffer size to the `OSSL_PARAM` owner.  Just as with everything else.

paulidale:
That would be a significant behavioural difference to what we do now.  `OSSL_PARAM_set_BN()` pads to the OSSL_PARAM size already.

I'm not convinced that we need `OSSL_PARAM_set_BN_pad()` -- if the OSSL_PARAM size is larger, it will be padded.  If it is smaller, we can't transfer the BIGNUM.

The width of the (padded) BIGNUM is never secret (that I remember), clearing extra bytes to a transfer width thus wouldn't be an information leak.

levitte:
> That would be a significant behavioural difference to what we do now.  `OSSL_PARAM_set_BN()` pads to the OSSL_PARAM size already.

Oh!  I had forgotten about that.  In that case yeah, I'm not sure I understand the need for a separate "pad" function that would essentially do same same...

paulidale:
The pad function would allow under sizing to a known width.  Since the width generally is public, copying to a larger buffer shouldn't leak anything....

PR #11398, opened 2020-03-25
Fix --strict-warnings build

p-steuer:
Appease -Werror=stringop-truncation

```
crypto/asn1/asn1_lib.c: In function 'sk_ASN1_UTF8STRING2text':
crypto/asn1/asn1_lib.c:429:13: error: 'strncpy' output truncated before terminating nul copying as many bytes from a string as its length [-Werror=stringop-truncation]
  429 |             strncpy(p, sep, sep_len);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~
crypto/asn1/asn1_lib.c:412:15: note: length computed here
  412 |     sep_len = strlen(sep);
      |               ^~~~~~~~~~~


```

I think the warning is right: strncpy(target,source,strlen(source)) makes no sense. You probably want a simple memcpy here.

### DIFF
--- a/crypto/asn1/asn1_lib.c
+++ b/crypto/asn1/asn1_lib.c
@@ -426,7 +426,7 @@ char *sk_ASN1_UTF8STRING2text(STACK_OF(ASN1_UTF8STRING) *text, const char *sep,
         current = sk_ASN1_UTF8STRING_value(text, i);
         length = ASN1_STRING_length(current);
         if (i > 0 && sep_len > 0) {
-            strncpy(p, sep, sep_len);
+            memcpy(p, sep, sep_len);
             p += sep_len;
         }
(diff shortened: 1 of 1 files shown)


mattcaswell:
Is this master only - or 1.1.1 as well?

p-steuer:
Its master only.

p-steuer:
Merge conflict due to b4ba2b7ce0933bede5d3b59a5abbde8fa3de2228 which changed exactly this line to strncpy(target,source,strlen(source)+1), making the warnings disappear because the \0 will be copied.

Both strncpys inside the for loop would not need to copy the \0 tough, because it is over written by the next iteration appending to the string and the final \0 is added after the loop anyway. So still both strncpys could imo be replaced by memcpys since using strncpy with a length depending on the source operand makes no sense.

However, Im fine with the solution, feel free to close this one if you agree.

PR #11400, opened 2020-03-25
The revert of EOF detection change in libssl for 1.1.1

t8m:
##### Checklist


kroeckx:
I think we should not remove things from the public headers.

t8m:
> I think we should not remove things from the public headers.

If we do 1.1.1f release ASAP I think it should not be a problem in this case. Especially that the presence of SSL_R_UNEXPECTED_EOF_WHILE_READING define can be used in applications to check the proper EOF handling in the OpenSSL 3.0.

t8m:
If you think we should keep the BIO_FLAGS_IN_EOF I would be OK with that, but then we need to drop the revert of all the BIO changes. So please tell me what do you prefer.

levitte:
I see no issue keeping BIO_FLAGS_IN_EOF...  but it could be perceived as an added feature...

mattcaswell:
> I see no issue keeping BIO_FLAGS_IN_EOF... but it could be perceived as an added feature...

Or a bug that it wasn't there in the first place?

kroeckx:
Maybe in case we get the unexpected EOF, we should just simulate
the old behaviour?

t8m:
> Maybe in case we get the unexpected EOF, we should just simulate the old behaviour?

That's what the revert of the changes in libssl does.

t8m:
> I see no issue keeping BIO_FLAGS_IN_EOF... but it could be perceived as an added feature...

We already did this for 1.1.1e because it was needed for the proper EOF handling, so I do not see an issue if we just keep it.

t8m:
So I've changed the PR to revert only the libssl changes.

kroeckx:
I wonder if s_client can detect the unexpected EOF, and if we
should still let it return an error.

kroeckx:
So I think the answer to that is, just it can detect it, it now
prints this "Success" error message. I think we should turn that
into a proper error message.

t8m:
> So I think the answer to that is, just it can detect it, it now prints this "Success" error message. I think we should turn that into a proper error message.

That should be a different PR.

t8m:
Merged to 1.1.1 as 0cd2ee64bffcdece599c3e4b5fac3830a55dc0fa and 30d190caf311d534867df97e26b552e628cb7d85

PR #11332, opened 2020-03-15
Add dh keygen to provider

slontis:
##### Checklist

- [x] documentation is added or updated
- [ ] tests are added or updated

PR #11401, opened 2020-03-25
TLSv1.2 connection using only FIPS crypto (v2)

mattcaswell:
This PR contains the final commits from #11334 to get a TLS connection working using only crypto from the FIPS module.

Some things to note:

- It is TLSv1.2 only at the moment because I can only do RSA key exchange. We don't have the necessary provider EC/DH key gen/param gen in place yet for anything else. However, TLSv1.3 does not support RSA key exchange.

- This only does a very simple handshake, i.e. it doesn't include resumption, PSKs, etc. 

- In #11334 I included changes to libssl to use the new EVP_MAC API. This had the unexpected impact of breaking the GOST engine. I have not included those commits here because I have solved the problem I was trying to get over in a different way - so we don't need to make those changes just yet. However, we might want to still consider making that change in 3.0 - but not right now.

- We don't currently have fetchable RAND. This means that where libssl generates random numbers directly (as opposed to say an algorithm generating random numbers itself as part of its operation), we use the libcrypto DRBG - not the one in the FIPS module. The underlying cipher is still fetched from the FIPS module, but the DRBG itself is not.

In order to get more tests working using only FIPS we need to get EC/ECX/DH/DSA key generation in (#11328, #11371, #11332). I have a "hacked" local branch which is enabling me to make some progress towards this goal, so I expect to be raising various libssl fixes needed for the more complete libssl with FIPS support.

### DIFF
--- a/ssl/s3_lib.c
+++ b/ssl/s3_lib.c
@@ -4723,19 +4723,33 @@ EVP_PKEY *ssl_generate_pkey_group(SSL *s, uint16_t id)
      */
 # ifndef OPENSSL_NO_DH
     if (gtype == TLS_GROUP_FFDHE)
+#  if 0
+        pctx = EVP_PKEY_CTX_new_from_name(s->ctx->libctx, "DH", s->ctx->propq);
+#  else
         pctx = EVP_PKEY_CTX_new_id(EVP_PKEY_DH, NULL);
+#  endif
 #  ifndef OPENSSL_NO_EC
     else
-#  endif
-# endif
+#  endif /* OPENSSL_NO_EC */
+# endif /* OPENSSL_NO_DH */
 # ifndef OPENSSL_NO_EC
     {
+        /*
+         * TODO(3.0): When provider based EC key gen is present we can enable
+         * this code.
+         */
         if (gtype == TLS_GROUP_CURVE_CUSTOM)
             pctx = EVP_PKEY_CTX_new_id(ginf->nid, NULL);
         else
+#  if 0
+            pctx = EVP_PKEY_CTX_new_from_name(s->ctx->libctx, "EC",
+                                              s->ctx->propq);
+#  else
             pctx = EVP_PKEY_CTX_new_id(EVP_PKEY_EC, NULL);
+#  endif
+
     }
-# endif
+# endif /* OPENSSL_NO_EC */
     if (pctx == NULL) {
         SSLfatal(s, SSL_AD_INTERNAL_ERROR, SSL_F_SSL_GENERATE_PKEY_GROUP,
                  ERR_R_MALLOC_FAILURE);
@@ -4801,7 +4815,11 @@ EVP_PKEY *ssl_generate_param_group(SSL *s, uint16_t id)
     EVP_PKEY_CTX *pctx = NULL;
     EVP_PKEY *pkey = NULL;
     const TLS_GROUP_INFO *ginf = tls1_group_id_lookup(id);
+#if 0
+    const char * pkey_ctx_name;
+#else
     int pkey_ctx_id;
+#endif
 
     if (ginf == NULL)
         goto err;
@@ -4819,9 +4837,16 @@ EVP_PKEY *ssl_generate_param_group(SSL *s, uint16_t id)
      * s->ctx->libctx and s->ctx->propq when paramgen has been updated to be
      * provider aware.
      */
+#if 0
+    pkey_ctx_name = (ginf->flags & TLS_GROUP_FFDHE) ? "DH" : "EC";
+    pctx = EVP_PKEY_CTX_new_from_name(s->ctx->libctx, pkey_ctx_name,
+                                      s->ctx->propq);
+#else
     pkey_ctx_id = (ginf->flags & TLS_GROUP_FFDHE)
                         ? EVP_PKEY_DH : EVP_PKEY_EC;
     pctx = EVP_PKEY_CTX_new_id(pkey_ctx_id, NULL);
+#endif
+
     if (pctx == NULL)
         goto err;
     if (EVP_PKEY_paramgen_init(pctx) <= 0)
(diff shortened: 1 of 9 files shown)


richsalz:
Congratulations; this is a major milestone!

ssl/s3_lib.c
```diff
@@ -4723,19 +4723,33 @@ EVP_PKEY *ssl_generate_pkey_group(SSL *s, uint16_t id)
      */
 # ifndef OPENSSL_NO_DH
     if (gtype == TLS_GROUP_FFDHE)
+#  if 0
```
paulidale: #if 0 ??

test/sslprovidertest.c
```diff
@@ -7,32 +7,49 @@
...
+
+    if (!TEST_true(OPENSSL_CTX_load_config(libctx, configfile)))
         goto end;
-#endif
+
+    prov = OSSL_PROVIDER_load(libctx, modulename);
+    if (!TEST_ptr(prov)
+               /* Check we have the provider available */
+            || !TEST_true(OSSL_PROVIDER_available(libctx, modulename))
+               /* Check the default provider is not available */)
```
paulidale: Comment should be below the goto shouldn't it?

paulidale:
Travis failure is relevant.

ssl/s3_lib.c
mattcaswell: The "#if 0" code is stuff we're going to need, but I can't enable yet because EC/DH keygen is not yet in place. As soon as it is this code should be made active.

test/sslprovidertest.c
mattcaswell: Oops. An artefact from an earlier code iteration. Fixed now.

mattcaswell:
> Travis failure is relevant.

Fixup pushed hopefully addressing travis failure and above feedback.

mattcaswell:
Ping?

ssl/s3_lib.c
```diff
@@ -4723,19 +4723,33 @@ EVP_PKEY *ssl_generate_pkey_group(SSL *s, uint16_t id)
...
+        pctx = EVP_PKEY_CTX_new_from_name(s->ctx->libctx, "DH", s->ctx->propq);
+#  else
         pctx = EVP_PKEY_CTX_new_id(EVP_PKEY_DH, NULL);
+#  endif
 #  ifndef OPENSSL_NO_EC
     else
-#  endif
-# endif
+#  endif /* OPENSSL_NO_EC */
```
kaduk: We appear to have an ifndef/else/endif trio with no content in either branch?

ssl/statem/statem_lib.c
```diff
@@ -272,7 +272,9 @@ int tls_construct_cert_verify(SSL *s, WPACKET *pkt)
         goto err;
     }
 
-    if (EVP_DigestSignInit(mctx, &pctx, md, NULL, pkey) <= 0) {
+    if (EVP_DigestSignInit_ex(mctx, &pctx,
+                              md == NULL ? NULL : EVP_MD_name(md),
```
kaduk: Did we think about having `EVP_MD_name()` accept `NULL` and return `NULL`?  This ternary appears a few times already...

ssl/s3_lib.c
mattcaswell: No - we have ifndef/endif with "else" as the content!!

ssl/statem/statem_lib.c
mattcaswell: Hmmmm.....maybe......not sure. But I'd say out of scope for this PR.

ssl/s3_lib.c
kaduk: Totally not confusing :)
Thanks for the cluebat.

ssl/s3_lib.c
```diff
@@ -4801,7 +4815,11 @@ EVP_PKEY *ssl_generate_param_group(SSL *s, uint16_t id)
     EVP_PKEY_CTX *pctx = NULL;
     EVP_PKEY *pkey = NULL;
     const TLS_GROUP_INFO *ginf = tls1_group_id_lookup(id);
+#if 0
+    const char * pkey_ctx_name;
```
kaduk: `*pkey_ctx_name` (no space)?

ssl/s3_lib.c
```diff
@@ -4819,9 +4837,16 @@ EVP_PKEY *ssl_generate_param_group(SSL *s, uint16_t id)
      * s->ctx->libctx and s->ctx->propq when paramgen has been updated to be
      * provider aware.
      */
+#if 0
+    pkey_ctx_name = (ginf->flags & TLS_GROUP_FFDHE) ? "DH" : "EC";
```
kaduk: I forget if we want a `!= 0` check for the flags, in our style guide.

test/sslprovidertest.c
```diff
@@ -7,32 +7,48 @@
...
     int testresult = 0;
     OPENSSL_CTX *libctx = OPENSSL_CTX_new();
+    OSSL_PROVIDER *prov = NULL;
 
-/* TODO(3.0): Re-enable this code. See comment in setup_tests() */
-#if 0
-    /* Verify that the default provider in the default libctx is not available */
-    if (!TEST_false(OSSL_PROVIDER_available(NULL, "default")))
+    /*
+     * Verify that the default and fips providers in the default libctx iare not
```
kaduk: s/iare/are/

test/sslprovidertest.c
```diff
@@ -114,15 +136,14 @@ int setup_tests(void)
         return 0;
     }
 
+    defctxlegacy = OSSL_PROVIDER_load(NULL, "legacy");
```
kaduk: Should the block comment move down here along with the `OSSL_PROVIDER_load(NULL, .)`?  It seems to still be most relevant here.

ssl/s3_lib.c
mattcaswell: Fixed.

ssl/s3_lib.c
mattcaswell: Yes, we should. Done.

test/sslprovidertest.c
mattcaswell: Fixed

test/sslprovidertest.c
mattcaswell: Done

mattcaswell:
Fixups pushed addressing the feedback above. Please take another look.

mattcaswell:
Pushed. Thanks!

PR #11402, opened 2020-03-25
Use fetched ciphers in a few more spots in libssl

mattcaswell:
There were a couple more spots in libssl where we were not using fetched ciphers, so this PR addresses those locations.


mattcaswell:
Pushed. Thanks.

PR #11403, opened 2020-03-25
TLS Cipher Suite 0xC102 Support

NMorozxov:
For GOST2012-GOST8912-GOST8912 was used 0xFF85 identifier,
but new identifier 0xc102 was assigned.
Because of old software we will support both numbers.

https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-2
https://datatracker.ietf.org/doc/draft-smyshlyaev-tls12-gost-suites/

##### Checklist

### DIFF
--- a/ssl/t1_trce.c
+++ b/ssl/t1_trce.c
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2018 The OpenSSL Project Authors. All Rights Reserved.
+ * Copyright 2012-2020 The OpenSSL Project Authors. All Rights Reserved.
  *
  * Licensed under the Apache License 2.0 (the "License").  You may not use
  * this file except in compliance with the License.  You can obtain a copy
@@ -427,6 +427,7 @@ static const ssl_trace_tbl ssl_ciphers_tbl[] = {
     {0xC0AD, "TLS_ECDHE_ECDSA_WITH_AES_256_CCM"},
     {0xC0AE, "TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8"},
     {0xC0AF, "TLS_ECDHE_ECDSA_WITH_AES_256_CCM_8"},
+    {0xC102, "IANA-GOST2012-GOST8912-GOST8912"},
     {0xCCA8, "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256"},
     {0xCCA9, "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256"},
     {0xCCAA, "TLS_DHE_RSA_WITH_CHACHA20_POLY1305_SHA256"},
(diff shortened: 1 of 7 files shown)


kaduk:
I am wondering whether there would be a desire for a knob to provide "strictly compliant" behavior (i.e., only recognize 0xc102 and not 0xff85), and what such a knob would look like.
We usually do not have knobs for such narrow behavior changes, though, so it seems somewhat unlikely that we would add such a thing.

mattcaswell:
This change will cause backwards compat problems. Because we rename the old cipher to `LEGACY-GOST2012-GOST8912-GOST8912` this means that any existing configs will automatically start using the new definition and cipher id for `GOST2012-GOST8912-GOST8912`. This will fail to interoperate with any existing installations that don't yet understand the new cipher-id. This will force all users to amend their configs to list both the old and the new values.

One possibility is to arrange things so that if clients have `GOST2012-GOST8912-GOST8912` configured then they always send both cipher ids in the ClientHello (with the new one before the old one). On the server side, if `GOST2012-GOST8912-GOST8912` is configured then it could accept either cipher id.

NMorozxov:
How i can fo this arrangement?
1) i can combine LEGACY-GOST2012-GOST8912-GOST8912 and NEW-GOST2012-GOST8912-GOST8912 to GOST2012-GOST8912-GOST8912 to a group, but there is no group expext STRONG, may be i must add a new one
2) I can fix code in place where the ciphers is sended

NMorozxov:
Or i can leave the old one, and add a new one IANA-GOST2012-GOST8912-GOST8912 or something like this

mattcaswell:
> i can combine LEGACY-GOST2012-GOST8912-GOST8912 and NEW-GOST2012-GOST8912-GOST8912 to GOST2012-GOST8912-GOST8912 to a group, but there is no group expext STRONG, may be i must add a new one

This sounds plausible - but perhaps do it as an alias. See cipher_aliases in ssl_ciph.c

beldmit:
@NMorozxov, do you have plans to implement Matt's suggestion?

NMorozxov:
@beldmit I have pushed yesterday. Two chippers and alias. What suggestions did i miss ?

beldmit:
Sorry, I missed the commit. Thanks!

ssl/t1_trce.c
```diff
@@ -427,6 +427,7 @@ static const ssl_trace_tbl ssl_ciphers_tbl[] = {
     {0xC0AD, "TLS_ECDHE_ECDSA_WITH_AES_256_CCM"},
     {0xC0AE, "TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8"},
     {0xC0AF, "TLS_ECDHE_ECDSA_WITH_AES_256_CCM_8"},
+    {0xC102, "GOST2012-GOST8912-GOST8912"},
```
beldmit: Worth mentioning IANA and legacy code here too.

beldmit: Thanks for pushing the fix!
But could you also please write a note about it when done?

NMorozxov:
@mattcaswell @beldmit fix submited

beldmit:
Approve if Travis is happy

beldmit:
Merged. Many thanks!

PR #11404, opened 2020-03-25
Chunk 10 of CMP contribution to OpenSSL: CMP HTTP client and related tests

DDvO:
Chunk 10 of CMP contribution to OpenSSL: CMP HTTP client and related tests
    
Also improves the [generic HTTP client](https://github.com/openssl/openssl/pull/10667) w.r.t. proxy and no_proxy options,
makes HTTP server/proxy address and port parameters more consistent,
and include minor other improvements of the generic HTTP client.
Makes use of the generic HTTP client for the CMP client implementation.

##### Checklist


DDvO:
Strange that Travis still shows yellow in the CI status overview above although its details indicate that all its runs have passed 16 hours ago.

crypto/http/http_lib.c
```diff
@@ -114,3 +120,46 @@ int OSSL_HTTP_parse_url(const char *url, char **phost, char **pport,
...
+                             const char *server, int use_ssl)
+{
+    const int http_len = strlen(OSSL_HTTP_PREFIX);
+    const int https_len = strlen(OSSL_HTTPS_PREFIX);
+
+    if (proxy == NULL)
+        proxy = getenv(use_ssl ? "https_proxy" : "http_proxy");
+    if (proxy == NULL)
+        proxy = getenv(use_ssl ? "HTTPS_PROXY" : "HTTP_PROXY");
+    if (proxy != NULL && strncmp(proxy, OSSL_HTTP_PREFIX, http_len) == 0)
```
Akretsch: Maybe its simpler:

if (proxy == NULL) 
    return NULL;
Then drop "proxy != NULL" from next lines.

DDvO: Good point - done.

crypto/cmp/cmp_http.c
```diff
@@ -0,0 +1,66 @@
...
+        CMPerr(0, CMP_R_NULL_ARGUMENT);
+        return 0;
+    }
+
+    if (!X509V3_add_value("Pragma", "no-cache", &headers))
+        return NULL;
+
+    BIO_snprintf(server_port, sizeof(server_port), "%d", ctx->serverPort);
+
+    res = (OSSL_CMP_MSG *)
```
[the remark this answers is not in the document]
DDvO: Yes; removed.
Doing so I found that the port parameter default value use of this function, as well as of `OSSL_HTTP_transfer()` and friends, can be made more flexible. Done as well.

DDvO:
Travis again brought the usual false postitive (due to timeout).

DDvO:
Neat that the number of this PR ends in 404 - kind of suitable for an HTTP topic :wink:

crypto/http/http_lib.c
```diff
@@ -22,8 +25,11 @@ int OSSL_HTTP_parse_url(const char *url, char **phost, char **pport,
 {
     char *p, *buf;
     char *host;
-    char *port = "80";
+    const char *port = OSSL_HTTP_PORT;
+    size_t https_len = strlen(OSSL_HTTPS_NAME);
 
+    if (!ossl_assert(https_len >= strlen(OSSL_HTTP_NAME)))
```
mattcaswell: This seems pointless given the line just above.

crypto/http/http_lib.c
```diff
@@ -114,3 +120,49 @@ int OSSL_HTTP_parse_url(const char *url, char **phost, char **pport,
...
+    const char *found = NULL;
+
+    if (!ossl_assert(server != NULL))
+        return 0;
+    sl = strlen(server);
+
+    if (no_proxy == NULL)
+        no_proxy = getenv("no_proxy");
+    if (no_proxy == NULL)
+        no_proxy = getenv("NO_PROXY");
```
mattcaswell: Should this be `OPENSSL_NO_PROXY` or similar? Or perhaps better `OPENSSL_NO_HTTPS_PROXY`?

Not sure why we have both case variants. We don't do that for any other env variable AFAIK.

crypto/http/http_lib.c
```diff
@@ -114,3 +120,49 @@ int OSSL_HTTP_parse_url(const char *url, char **phost, char **pport,
...
+const char *http_adapt_proxy(const char *proxy, const char *no_proxy,
+                             const char *server, int use_ssl)
+{
+    const int http_len = strlen(OSSL_HTTP_PREFIX);
+    const int https_len = strlen(OSSL_HTTPS_PREFIX);
+
+    if (proxy == NULL)
+        proxy = getenv(use_ssl ? "https_proxy" : "http_proxy");
+    if (proxy == NULL)
+        proxy = getenv(use_ssl ? "HTTPS_PROXY" : "HTTP_PROXY");
```
mattcaswell: As above

crypto/cmp/cmp_ctx.c
```diff
@@ -36,7 +34,6 @@ X509_STORE *OSSL_CMP_CTX_get0_trustedStore(const OSSL_CMP_CTX *ctx)
  * Set certificate store containing trusted (root) CA certs and possibly CRLs
  * and a cert verification callback function used for CMP server authentication.
  * Any already existing store entry is freed. Given NULL, the entry is reset.
- * returns 1 on success, 0 on error
```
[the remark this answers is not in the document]
DDvO: Just because I felt these comments are pretty redundant.
Nearly all functions in this file are public and thus fully documented.
The return values are documented in the pod files and obvious in the sources.
Anyway the ones remove here are just the usual `1` for success and `0` for error.

crypto/http/http_lib.c
DDvO: This just makes a precondition of a later pointer calculation obvious
and checks it to be on the safe side (in case the values of `OSSL_HTTP_NAME` or 'OSSL_HTTPS_NAME` changes, yeah, pretty unlikely for the next 50+ years).

crypto/http/http_lib.c
DDvO: Good point to add `OPENSSL_NO_PROXY`, done.

Regarding the lower-case variants, I found that some projects do so for compatibility - I can kick them out here if you prefer.

crypto/http/http_lib.c
DDvO: I've added `OPENSSL_HTTP_PROXY`  and `OPENSSL_HTTPS_PROXY`.

DDvO:
Thanks @mattcaswell for your review!
All comments addressed.

DDvO:
Travis failure is the usual timeout.

crypto/http/http_lib.c
DDvO: It seems that there is no normative standard on the proxy environment variables.

Here is a nice description of what Git does: https://github.com/request/request#controlling-proxy-behaviour-using-environment-variables
and this PR follows that model.

Curl is weird in the sense that it only supports the lower_case `http_proxy` but both lower- and upper-case `HTTPS_PROXY` - see https://ec.haxx.se/usingcurl/usingcurl-proxies

BTW, apparently no implementation uses `NO_HTTPS_PROXY` or the like - the only env vars typically used for exclusion are `NO_PROXY` and/or `no_proxy` - these are used for both HTTP and HTTPS.

https://en.wikibooks.org/wiki/Guide_to_Unix/Environment_Variables writes

> The names of environment variables are case-sensitive; by convention they are uppercase.

Unfortunately it is not guaranteed that `getenv` is case insensitive - see,  e.g., https://www.php.net/manual/en/function.getenv.php

crypto/http/http_lib.c
mattcaswell: Ah...actually my comment was because I hadn't spotted that one was referring to `OSSL_HTTPS_NAME` and one was `OSSL_HTTP_NAME`...the extra `S` had passed me by. So...ok.

crypto/http/http_lib.c
```diff
@@ -114,3 +120,50 @@ int OSSL_HTTP_parse_url(const char *url, char **phost, char **pport,
...
+    const char *found = NULL;
+
+    if (!ossl_assert(server != NULL))
+        return 0;
+    sl = strlen(server);
+
+    if (no_proxy == NULL)
+        no_proxy = getenv("no_proxy");
+    if (no_proxy == NULL)
+        no_proxy = getenv(OPENSSL_NO_PROXY);
```
mattcaswell, line 136: I had actually been thinking that the environment variable itself (i.e. the string) should be called `OPENSSL_NO_PROXY` - but what I hand't appreciated was that these names were deliberately chosen for compatibility with other software and aren't OpenSSL specific. Given that it might be worth a comment explaining that. This also seems like a reasonable justification for accepting the lowercase variant where we don't do that anywhere else for OpenSSL environment variables.

DDvO: I see.
I've added respective source comments to `http_lib.c` and a corresponding NOTES section to `OSSL_HTTP_transfer.pod`.

DDvO:
Great that this already approved - thanks @mattcaswell for your swift responses!

I'm going to merge this soon after the 24 hours have passed.
Just weird that GitHub still shows the Travis run as non-finished here, while the details indicate that all runs went fine yesterday.

DDvO:
Rebased, squashed to two commits, and merged. Thanks!

PR #11405, opened 2020-03-25
Provide better diagnostics if a fetch fails

mattcaswell:
Fetch failures are a common problem. If that happens it is useful to have more detailed information about the fetch, such as the algorithm that was being fetched, along with the properties and whether we were using the default libctx or not.

### DIFF
--- a/crypto/evp/evp_fetch.c
+++ b/crypto/evp/evp_fetch.c
@@ -294,9 +294,25 @@ void *evp_generic_fetch(OPENSSL_CTX *libctx, int operation_id,
                         int (*up_ref_method)(void *),
                         void (*free_method)(void *))
 {
-    return inner_evp_generic_fetch(libctx,
-                                   operation_id, 0, name, properties,
-                                   new_method, up_ref_method, free_method);
+    void *ret = inner_evp_generic_fetch(libctx,
+                                        operation_id, 0, name, properties,
+                                        new_method, up_ref_method, free_method);
+
+    if (ret == NULL) {
+        EVPerr(0, EVP_R_FETCH_FAILED);
+#ifndef FIPS_MODE
+        ERR_add_error_data(6,
+                           openssl_ctx_is_default(libctx)
+                           ? "Default library context"
+                           : "Non-default library context",
+                           ", Algorithm (",
+                           name == NULL ? "<null>" : name,
+                           "), Properties (",
+                           properties == NULL ? "<null>" : properties,
+                           ")");
+#endif
+    }
+    return ret;
 }
 
 /*
@@ -314,9 +330,32 @@ void *evp_generic_fetch_by_number(OPENSSL_CTX *libctx, int operation_id,
                                   int (*up_ref_method)(void *),
                                   void (*free_method)(void *))
 {
-    return inner_evp_generic_fetch(libctx,
-                                   operation_id, name_id, NULL, properties,
-                                   new_method, up_ref_method, free_method);
+    void *ret = inner_evp_generic_fetch(libctx,
+                                        operation_id, name_id, NULL,
+                                        properties, new_method, up_ref_method,
+                                        free_method);
+
+    if (ret == NULL) {
+        EVPerr(0, EVP_R_FETCH_FAILED);
+#ifndef FIPS_MODE
+        {
+            OSSL_NAMEMAP *namemap = ossl_namemap_stored(libctx);
+            const char *name = (namemap == NULL)
+                               ? NULL
+                               : ossl_namemap_num2name(namemap, name_id, 0);
+            ERR_add_error_data(6,
+                               openssl_ctx_is_default(libctx)
+                               ? "Default library context"
+                               : "Non-default library context",
+                               ", Algorithm (",
+                               name == NULL ? "<null>" : name,
+                               "), Properties (",
+                               properties == NULL ? "<null>" : properties,
+                               ")");
+        }
+#endif
+    }
+    return ret;
 }
 
 int EVP_set_default_properties(OPENSSL_CTX *libctx, const char *propq)
--- a/crypto/evp/evp_fetch.c
+++ b/crypto/evp/evp_fetch.c
@@ -299,17 +299,18 @@ void *evp_generic_fetch(OPENSSL_CTX *libctx, int operation_id,
                                         new_method, up_ref_method, free_method);
 
     if (ret == NULL) {
-        EVPerr(0, EVP_R_FETCH_FAILED);
-#ifndef FIPS_MODE
-        ERR_add_error_data(6,
-                           openssl_ctx_is_default(libctx)
-                           ? "Default library context"
-                           : "Non-default library context",
-                           ", Algorithm (",
-                           name == NULL ? "<null>" : name,
-                           "), Properties (",
-                           properties == NULL ? "<null>" : properties,
-                           ")");
+        int code = EVP_R_FETCH_FAILED;
+
+#ifdef FIPS_MODE
+        ERR_raise(ERR_LIB_EVP, code);
+#else
+        ERR_raise_data(ERR_LIB_EVP, code,
+                       "%s, Algorithm (%s), Properties (%s)",
+                       (openssl_ctx_is_default(libctx)
+                        ? "Default library context"
+                        : "Non-default library context"),
+                       name = NULL ? "<null>" : name,
+                       properties == NULL ? "<null>" : properties);
 #endif
     }
     return ret;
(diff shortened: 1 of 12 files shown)


crypto/evp/evp_fetch.c
```diff
@@ -294,9 +294,25 @@ void *evp_generic_fetch(OPENSSL_CTX *libctx, int operation_id,
...
+        ERR_add_error_data(6,
+                           openssl_ctx_is_default(libctx)
+                           ? "Default library context"
+                           : "Non-default library context",
+                           ", Algorithm (",
+                           name == NULL ? "<null>" : name,
+                           "), Properties (",
+                           properties == NULL ? "<null>" : properties,
+                           ")");
+#endif
```
levitte:
``` C
        int code = EVP_R_FETCH_FAILED;

#ifdef FIPS_MODE
        ERR_raise(ERR_LIB_EVP, code);
#else
        ERR_raise_data(ERR_LIB_EVP, code,
                       "%s, Algorithm (%s), Properties (%s)",
                       (openssl_ctx_is_default(libctx)
                        ? "Default library context"
                        : "Non-default library context"),
                       name = NULL ? "<null>" : name,
                       properties == NULL ? "<null>" : properties);
#endif
```

Dunno about you, but I find that a tad more readable...

crypto/evp/evp_fetch.c
```diff
@@ -314,9 +330,32 @@ void *evp_generic_fetch_by_number(OPENSSL_CTX *libctx, int operation_id,
...
+                                        free_method);
+
+    if (ret == NULL) {
+        EVPerr(0, EVP_R_FETCH_FAILED);
+#ifndef FIPS_MODE
+        {
+            OSSL_NAMEMAP *namemap = ossl_namemap_stored(libctx);
+            const char *name = (namemap == NULL)
+                               ? NULL
+                               : ossl_namemap_num2name(namemap, name_id, 0);
```
levitte: There's a `evp_first_name()`...

crypto/evp/exchange.c
```diff
@@ -229,9 +229,12 @@ int EVP_PKEY_derive_init(EVP_PKEY_CTX *ctx)
 
     /*
      * Because we cleared out old ops, we shouldn't need to worry about
-     * checking if exchange is already there.
+     * checking if exchange is already there. The fetch can fail in the legacy
+     * case.
      */
+    ERR_set_mark();
     exchange = EVP_KEYEXCH_fetch(ctx->libctx, supported_exch, ctx->propquery);
+    ERR_pop_to_mark();
```
levitte: If you look, you'll find `ERR_set_mark()` and `ERR_pop_to_mark()` are already used, but not so tightly around this specific call.

levitte: That goes for all places you added them

providers/common/provider_util.c
```diff
@@ -76,12 +76,17 @@ int ossl_prov_cipher_load_from_params(PROV_CIPHER *pc,
...
     pc->cipher = pc->alloc_cipher = EVP_CIPHER_fetch(ctx, p->data, propquery);
     /* TODO legacy stuff, to be removed */
 #ifndef FIPS_MODE /* Inside the FIPS module, we don't support legacy ciphers */
     if (pc->cipher == NULL)
         pc->cipher = EVP_get_cipherbyname(p->data);
 #endif
+    if (pc->cipher != NULL)
+        ERR_pop_to_mark();
+    else
+        ERR_clear_last_mark();
```
levitte: Ok, perhaps not *all* places...

ssl/ssl_ciph.c
```diff
@@ -322,6 +322,7 @@ int ssl_load_ciphers(SSL_CTX *ctx)
     const ssl_cipher_table *t;
 
     disabled_enc_mask = 0;
+
```
levitte: Spurious?

ssl/ssl_lib.c
```diff
@@ -5862,8 +5864,11 @@ const EVP_CIPHER *ssl_evp_cipher_fetch(OPENSSL_CTX *libctx,
     }
 #endif
 
-    /* Otherwise we do an explicit fetch */
-    return EVP_CIPHER_fetch(libctx, OBJ_nid2sn(nid), properties);
+    /* Otherwise we do an explicit fetch. This may fail and that could be ok */
```
levitte: Uhm, why is that ok?

crypto/evp/evp_fetch.c
mattcaswell: Ok - I changed it to that.

crypto/evp/evp_fetch.c
mattcaswell: That requires you to have a provider reference - which we don't here.

crypto/evp/exchange.c
mattcaswell: Ah - thanks. I missed that. I rolled this change back - and a number of the others.

ssl/ssl_ciph.c
mattcaswell: Removed

ssl/ssl_lib.c
mattcaswell: Because in libssl we sometimes try to get a cipher which we know is not available on all platforms.

mattcaswell:
Fixups pushed addressing the comments. Please take a look.

crypto/evp/evp_fetch.c
levitte: Ah, right...

ssl/ssl_lib.c
levitte: Ah, all right

crypto/evp/exchange.c
levitte: It seems you forgot about this

crypto/evp/m_sigver.c
```diff
@@ -110,10 +110,13 @@ static int do_sigver_init(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
...
     /*
      * Because we cleared out old ops, we shouldn't need to worry about
-     * checking if signature is already there.
+     * checking if signature is already there. The fetch can fail in the legacy
+     * case.
      */
+    ERR_set_mark();
     signature = EVP_SIGNATURE_fetch(locpctx->libctx, supported_sig,
                                     locpctx->propquery);
+    ERR_pop_to_mark();
```
levitte: ... and this...

crypto/evp/signature.c
```diff
@@ -391,10 +391,12 @@ static int evp_pkey_signature_init(EVP_PKEY_CTX *ctx, int operation)
 
     /*
      * Because we cleared out old ops, we shouldn't need to worry about
-     * checking if signature is already there.
+     * checking if signature is already there. This can fail in the legacy case.
      */
+    ERR_set_mark();
     signature =
         EVP_SIGNATURE_fetch(ctx->libctx, supported_sig, ctx->propquery);
+    ERR_pop_to_mark();
```
levitte: ... and here as well...

crypto/evp/pmeth_fn.c
```diff
@@ -71,10 +71,12 @@ static int evp_pkey_asym_cipher_init(EVP_PKEY_CTX *ctx, int operation)
 
     /*
      * Because we cleared out old ops, we shouldn't need to worry about
-     * checking if cipher is already there.
+     * checking if cipher is already there. This can fail in the legacy case.
      */
+    ERR_set_mark();
     cipher =
         EVP_ASYM_CIPHER_fetch(ctx->libctx, supported_ciph, ctx->propquery);
+    ERR_pop_to_mark();
```
levitte: ... oh, and here too

ssl/ssl_lib.c
```diff
@@ -5913,7 +5920,10 @@ const EVP_MD *ssl_evp_md_fetch(OPENSSL_CTX *libctx,
 #endif
 
     /* Otherwise we do an explicit fetch */
-    return EVP_MD_fetch(libctx, OBJ_nid2sn(nid), properties);
+    ERR_set_mark();
+    md = EVP_MD_fetch(libctx, OBJ_nid2sn(nid), properties);
+    ERR_pop_to_mark();
+    return md;
```
levitte: If `md` is NULL at this point, is that an error or not?  If it is, `ERR_set_mark()` / `ERR_pop_to_mark()` is wrong

xkqian:
Does the travis-ci failure of 33398.5 affect the result? I have one new patch which has the same travis-ci failure with 33398.5. Even my patch targets on the aarch64, I am no idea why it fails on x86-64...

providers/common/provider_util.c
```diff
@@ -76,12 +76,17 @@ int ossl_prov_cipher_load_from_params(PROV_CIPHER *pc,
         return 0;
 
     EVP_CIPHER_free(pc->alloc_cipher);
+    ERR_set_mark();
```
levitte: Travis tells that you need to include `<openssl/err.h>` in this file.

mattcaswell:
It helps if I actually commit the changes that I made. Sorry!

@levitte - please look again

ssl/ssl_lib.c
mattcaswell: No. It's not an error. Sometimes we attempt to get an MD that isn't available everywhere (e.g. GOST mds). We don't know if it's there until we try and get hold of it. If this function returns NULL we just ignore the digest and move on to the next one.

mattcaswell:
> Does the travis-ci failure of 33398.5 affect the result? I have one new patch which has the same travis-ci failure with 33398.5. Even my patch targets on the aarch64, I am no idea why it fails on x86-64...

No. Its a timeout on that job. Its a recurring problem that we haven't figured out how to fix yet.

mattcaswell:
@levitte - any more comments on this?

ssl/ssl_lib.c
levitte: Okie, fair enough

mattcaswell:
Pushed. Thanks.

PR #11406, opened 2020-03-25
Display errors if a test TLS connection fails

mattcaswell:
If a connection fails it is useful during debugging to have any errors that are on the OpenSSL error stack at the time of the failure.

### DIFF
--- a/test/ssltestlib.c
+++ b/test/ssltestlib.c
@@ -938,6 +938,8 @@ int create_bare_ssl_connection(SSL *serverssl, SSL *clientssl, int want,
 
         if (!clienterr && retc <= 0 && err != SSL_ERROR_WANT_READ) {
             TEST_info("SSL_connect() failed %d, %d", retc, err);
+            if (want != SSL_ERROR_SSL)
+                TEST_openssl_errors();
             clienterr = 1;
         }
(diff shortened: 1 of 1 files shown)


test/ssltestlib.c
```diff
@@ -938,6 +938,8 @@ int create_bare_ssl_connection(SSL *serverssl, SSL *clientssl, int want,
 
         if (!clienterr && retc <= 0 && err != SSL_ERROR_WANT_READ) {
             TEST_info("SSL_connect() failed %d, %d", retc, err);
+            if (want != SSL_ERROR_SSL)
```
slontis: A description of 'want' at the top might be worth doing at some point..

mattcaswell:
I added a description of the `want` parameter. Please take another look.

mattcaswell:
> The idea of the SSL_ERROR_SSL check is that the caller expects a failure and so we shouldn't be clearing the error queue?

We expect a failure and so we shouldn't be printing it - that will just confuse the output and make it look like an error occurred - when if fact that is expected test behaviour.

mattcaswell:
...and yes we want to retain the reason code on the error stack.

mattcaswell:
Pushed. Thanks.

PR #11407, opened 2020-03-25
Add OCSP_RESPID_set_by_key_ex() and OCSP_RESPID_match_ex()

mattcaswell:
OCSP_RESPID_set_by_key() calculates a SHA1 hash of the supplied
certificate. We need to be able to specify which libctx and property
query string is used to fetch that algorithm so we introduce
OCSP_RESPID_set_by_key_ex() which does the same thing but enables you to
speicfy the library context and propery query string explicitly.

OCSP_RESPID_match() matches with certificates based on the SHA1 hash.
Therefore for the same reason we introduce OCSP_RESPID_match_ex().

This will be needed in a later PR where we will make ssapitest run with a non-default libctx.

### DIFF
--- a/crypto/ocsp/ocsp_srv.c
+++ b/crypto/ocsp/ocsp_srv.c
@@ -259,45 +259,67 @@ int OCSP_RESPID_set_by_name(OCSP_RESPID *respid, X509 *cert)
     return 1;
 }
 
-int OCSP_RESPID_set_by_key(OCSP_RESPID *respid, X509 *cert)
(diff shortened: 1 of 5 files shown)


mattcaswell:
> Needs a rebase of the generated file

I rebased, and I'm assuming that your approval still holds.

slontis:
Yes. rebasing a generated file should be fine.

mattcaswell:
Pushed. Thanks.

PR #11409, opened 2020-03-25
Explicitly cache X509v3 extensions in libssl

mattcaswell:
A by product of processing the X509v3 extensions in an X509 object results in a SHA1 hash being calculated and stored for later use. Lots of X509 functions end up in the implicit processing of the X509v3 extensions and could result in this SHA1 hash being calculated. Prior to this PR the SHA1 hash would just use the default libctx and property query. However libssl needs to be able to specify a non-default libctx. Rather than modifying large numbers of X509 functions to add an OPENSSL_CTX parameter, we expose a new function to explicitly process the X509v3 extensions. This can be called up-front before other X509 functions are called.

### DIFF
--- a/crypto/x509/v3_purp.c
+++ b/crypto/x509/v3_purp.c
@@ -15,8 +15,6 @@
 #include "crypto/x509.h"
 #include "internal/tsan_assist.h"
 
-static void x509v3_cache_extensions(X509 *x);
-
 static int check_ssl_ca(const X509 *x);
 static int check_purpose_ssl_client(const X509_PURPOSE *xp, const X509 *x,
                                     int ca);
@@ -80,8 +78,7 @@ int X509_check_purpose(X509 *x, int id, int ca)
     int idx;
     const X509_PURPOSE *pt;
 
-    x509v3_cache_extensions(x);
-    if (x->ex_flags & EXFLAG_INVALID)
+    if (!X509v3_cache_extensions(x, NULL, NULL))
         return -1;
 
     /* Return if side-effect only call */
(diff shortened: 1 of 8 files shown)


crypto/x509/v3_purp.c
```diff
@@ -361,21 +358,25 @@ static void x509v3_cache_extensions(X509 *x)
     EXTENDED_KEY_USAGE *extusage;
     X509_EXTENSION *ex;
     int i;
+    EVP_MD *sha1;
 
 #ifdef tsan_ld_acq
     /* fast lock-free check, see end of the function for details. */
     if (tsan_ld_acq((TSAN_QUALIFIER int *)&x->ex_cached))
-        return;
+        return (x->ex_flags & EXFLAG_INVALID) == 0;
```
richsalz: I wonder if calling this bit out separately as an atomic-int field is worthwhile.

mattcaswell: The `ex_flags` can be obtained by application code using `X509_get_extension_flags()`, so for backwards compat reasons we'd still have to keep the `EXFLAG_INVALID` flag anyway. If we were designing this from scratch then what you say might make sense - but probably not now.

richsalz: You can keep the API, just merge both fields as the return from the "get-flags" method.

I am slightly concerned about the thread-safety of non-atomic (bitwise) operations, but oh well.

mattcaswell: > You can keep the API, just merge both fields as the return from the "get-flags" method.

We could do, but I don't think its worth it.

mattcaswell:
I rebased this to pick up the latest libcrypto.num changes.

Still looking for a reviewer for this. Ping!

kaduk:
[fixed up approval labels per @t8m's approval]

mattcaswell:
Pushed. Thanks.

PR #11410, opened 2020-03-25
Make SRP library context aware

mattcaswell:
libssl uses SRP. Since libssl is library context aware we need to make everything it uses (where applicable) library context aware too. In the case of SRP this means making a number of new functions that take a library context and property query string and calling those variants from libssl instead.

We also update sslapitest to use these new variants. At the moment we just pass NULL for the library context - but a future PR will pass a non-default library context.

Note: SRP is the most awful API in the history of all APIs. I'm not sure how widely used it is but I would love to kill it at some point. If we must keep it then the APIs should be ditched and some fetchable EVP version created instead. For now though I've just done the minimum necessary to get it to work.


mattcaswell:
Pushed with your suggested nit fix. I also rebased to pick up the latest util/libcrypto.num

PR #11415, opened 2020-03-26
s_server: Properly indicate ALPN protocol mismatch

t8m:
Return SSL_TLSEXT_ERR_ALERT_FATAL from alpn_select_cb so that
an alert is sent to the client on ALPN protocol mismatch.


kaduk:
How does this fit into the general model of "`s_server` (and `s_client`) are debugging tools that try to continue past things that would normally be errors"?

t8m:
In most cases they do not continue in cases where a protocol violation would have to occur to continue. I do not really see harm in this fix. Also it seems to me to be more useful to be able to test the proper ALPN handling than making the ALPN protocol just optional.

t8m:
Merged to master as 9e885a707d

Issue #10069, opened 2019-10-02
API to send NewSessionTicket on demand


kaduk:
TLS 1.3 allows for the NewSessionTicket message to be sent at any time after the handshake.
OpenSSL offers an SSL_SESSION_set1_ticket_appdata() API to let the application include data in that session ticket, and the normal course of execution of a program might cause the application to wish to update the data included in the session ticket used by the client.  In order to do so, the application would need a way to cause libssl to generate a NewSessionTicket message on demand; libssl currently only emits (SSL_set_num_tickets()) tickets immediately after a handshake completes, and never after that.  An SSL_write_ticket() API was proposed in #7948 as one of many potential routes towards solving that problem, but it may have utility in other scenarios as well.

This issue is for discussion of whether we would want to support this general class of functionality, and if so, what the appropriate API contract would be.  I haven't started to write any code since this would necessarily interact with the handshake state machine, which always requires some delicateness.  Probably we would need to generalize the mechanism we use to re-enter handshake mode for update-requested KeyUpdate scenarios in order to handle the requesut to send NewSessionTicket as well, but that's just from memory and not based on any experimentation.

mattcaswell:
This seems like a reasonable request, and something that I always expected we would do anyway.

kaduk:
Belatedly closing, since #11416 was merged

PR #11416, opened 2020-03-26
Add API to generate NewSessionTicket on demand

kaduk:
Add a TLS-1.3-server-only API to allow the application to request that a NewSessionTicket message be generated the next time we are about to start writing a data record.

Fixes: https://github.com/openssl/openssl/issues/10069

##### Checklist


test/sslapitest.c
```diff
@@ -2082,6 +2082,89 @@ static int test_psk_tickets(void)
...
+    new_called = 0;
+    if (!TEST_true(SSL_new_session_ticket(serverssl))
+            || !TEST_true(SSL_new_session_ticket(serverssl))
+            || !TEST_true(SSL_write_ex(serverssl, &c, 0, &nbytes))
+            || !TEST_int_eq(0, nbytes)
+            || !TEST_int_eq(2, new_called)
+            || !TEST_false(SSL_read_ex(clientssl, buf, sizeof(buf), &nbytes))
+            || !TEST_int_eq(4, new_called))
+        goto end;
+
```
kaduk: I'd love to also be able to test the logic that delays writing tickets while there's a partial write pending, but as far as I can tell the mempacket BIO that we use in this test harness does not have a way to leave an incomplete write, at present.

ssl/ssl_lib.c
```diff
@@ -2295,6 +2295,14 @@ int SSL_renegotiate_pending(const SSL *s)
     return (s->renegotiate != 0);
 }
 
+int SSL_new_session_ticket(SSL *s)
+{
+	if (SSL_in_init(s) || !s->server || !SSL_IS_TLS13(s))
```
kaduk: I feel like there should be another thing to check here, but this matches the documentation and I am not sure what the other check would be...

ssl/statem/statem_srvr.c
```diff
@@ -723,7 +730,7 @@ WORK_STATE ossl_statem_server_pre_work(SSL *s, WORK_STATE wst)
         return WORK_FINISHED_CONTINUE;
 
     case TLS_ST_SW_SESSION_TICKET:
-        if (SSL_IS_TLS13(s) && s->sent_tickets == 0) {
+        if (SSL_IS_TLS13(s) && s->sent_tickets == 0 && s->ext.extra_tickets_expected == 0) {
```
kaduk: This check worries me a bit, since it implicitly relies on the invariant that `s->ext.extra_tickets_expected` is zero during the initial handshake.  (A potential design to implement "only send tickets on first application data" on top of this would violate that invariant.)

ssl/statem/statem_srvr.c
```diff
@@ -527,7 +532,9 @@ static WRITE_TRAN ossl_statem_server13_write_transition(SSL *s)
          * Following an initial handshake we send the number of tickets we have
          * been configured for.
          */
-        if (s->hit || s->num_tickets <= s->sent_tickets) {
+        if (s->ext.extra_tickets_expected > 0) {
+            s->ext.extra_tickets_expected--;
```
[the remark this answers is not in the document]
paulidale: Would conditioning on having completed the handshake make sense?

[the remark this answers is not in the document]
kaduk: That would probably make sense, yes.  Thanks!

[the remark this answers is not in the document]
kaduk: Unfortunately, neither `SSL_in_init()` nor `SSL_is_init_finished()` suffice for this call site (at least in my current model for the API), since in order to write new handshake messages (tickets) we go back "in init", and for the case when we want to write more than one ticket in a row, we have to *stay there* for multiple passes through this logic.  So when I tried it, the "send two tickets in a row" test only sent one ticket.

kaduk:
> I'd approve if my TLS 1.3 knowledge were sufficiently advanced.

I expect to ask Matt for review, but only after 3.0.0alpha1 is out -- this change is lower priority than the release process.

kaduk:
Travis error was just a timeout:

```
Still running (42 of 60): make _tests
```

kaduk:
@mattcaswell now that #11520 is closed, is it appropriate to put a bit of time into looking at this one?

kaduk:
Oops, I just saw https://github.com/openssl/openssl/pull/11470#issuecomment-616647970, so nevermind

kaduk:
@mattcaswell do you think you'll want to review this one?

ssl/statem/statem_srvr.c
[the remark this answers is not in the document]
mattcaswell: `SSL_IS_FIRST_HANDSHAKE` might do what you need to figure out whether the original handshake has completed or not

test/sslapitest.c
mattcaswell: We have `bio_s_always_retry()` does that help? See `test_key_update_in_write` for an example

doc/man3/SSL_CTX_set_num_tickets.pod
```diff
@@ -40,14 +42,30 @@ handshake then SSL_set_num_tickets() can be called again prior to calling
...
 
+To issue tickets after other events (such as application-layer changes),
+SSL_new_session_ticket() is used to request that a new ticket be sent when
+it is safe to do so.  New tickets are only allowed to be sent in this manner
+after the initial handshake has completed, and only for TLS 1.3 connections.
+The ticket generation and transmission are delayed until the server is
+starting a new write operation, so that it is bundled with other application
+data being written and properly aligned to a record boundary.
+SSL_new_session_ticket() can be called more than once to request additional
+tickets be sent; all such requests are queued until and the multiple tickets
```
mattcaswell, line 53: "queued until". Doesn't quite make sense

doc/man3/SSL_CTX_set_num_tickets.pod
```diff
@@ -40,14 +42,30 @@ handshake then SSL_set_num_tickets() can be called again prior to calling
...
 SSL_CTX_get_num_tickets() and SSL_get_num_tickets() return the number of
 tickets set by a previous call to SSL_CTX_set_num_tickets() or
 SSL_set_num_tickets(), or 2 if no such call has been made.
 
 =head1 RETURN VALUES
 
-SSL_CTX_set_num_tickets() and SSL_set_num_tickets() return 1 on success or 0 on
-failure.
+SSL_CTX_set_num_tickets(), SSL_set_num_tickets(), and
+SSL_new_session_ticket() return 1 on success or 0 on failure.
```
mattcaswell, line 68: Update the HISTORY section?

ssl/ssl_lib.c
mattcaswell: During the early data phase we come out of init for a while - but the first handshake has not yet completed. Therefore we might want a check of `SSL_IS_FIRST_HANDSHAKE` here (which returns true if still in the first handshake regardless of the "init" value).

ssl/statem/statem_srvr.c
mattcaswell: Perhaps an ossl_assert here to check this invariant is appropriate?

test/sslapitest.c
```diff
@@ -2082,6 +2082,91 @@ static int test_psk_tickets(void)
...
+    /* Now try a (real) write to actually send the tickets */
+    c = '1';
+    if (!TEST_true(SSL_write_ex(serverssl, &c, 0, &nbytes))
+            || !TEST_size_t_eq(0, nbytes)
+            || !TEST_int_eq(idx * 2 + 2, new_called)
+            || !TEST_false(SSL_read_ex(clientssl, buf, sizeof(buf), &nbytes))
+            || !TEST_int_eq(idx * 2 + 4, new_called)
+            /* || !TEST_int_eq(sizeof(buf), nbytes)*/
+            /* || !TEST_int_eq(c, buf[0]) */
```
mattcaswell, line 2128: Why the commented out code?

test/sslapitest.c
```diff
@@ -2082,6 +2082,91 @@ static int test_psk_tickets(void)
...
+               /* Check we got the number of tickets we were expecting */
+            || !TEST_int_eq(idx * 2, new_called)
+            || !TEST_true(SSL_new_session_ticket(serverssl))
+            || !TEST_true(SSL_new_session_ticket(serverssl))
+            || !TEST_int_eq(idx * 2, new_called))
+        goto end;
+
+    /* Now try a (real) write to actually send the tickets */
+    c = '1';
+    if (!TEST_true(SSL_write_ex(serverssl, &c, 0, &nbytes))
```
mattcaswell, line 2122: You talk about doing it again below using "dummy" writes...but this also looks like a dummy write

test/sslapitest.c
```diff
@@ -2082,6 +2082,91 @@ static int test_psk_tickets(void)
...
+    SSL_CTX *sctx = NULL, *cctx = NULL;
+    SSL *serverssl = NULL, *clientssl = NULL;
+    int testresult = 0;
+    size_t nbytes;
+    unsigned char c, buf[1];
+
+    new_called = 0;
+    do_cache = 1;
+
+    if (!setup_ticket_test(0, idx, &sctx, &cctx))
```
mattcaswell, line 2097: Should we also be testing this with stateful tickets (i.e. the first param here set to 1)?

kaduk:
Thanks for all the comments!  These are really good and will give me some more bits to work on for the next update.  (`SSL_IS_FIRST_HANDSHAKE()` is already looking promising; I'm not sure why I failed to find that when I first looked for it.)
With respect to the "dummy writes", they should be asking to write zero bytes, which is why I called them "dummy".

doc/man3/SSL_CTX_set_num_tickets.pod
kaduk: It was once "queued until it is safe to do so" but had a bad edit; fixed.

ssl/ssl_lib.c
kaduk: That is almost certainly the other thing I was thinking of; thanks!

ssl/statem/statem_srvr.c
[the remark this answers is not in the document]
kaduk: I added the check here, though later testing indicates that in a normal 1.3 handshake `SSL_IS_FIRST_HANDSHAKE()` will return 0 by the time we are writing tickets.  (Which makes sense, since it's looking at finished lengths, and NST happens after Finished.)

ssl/statem/statem_srvr.c
kaduk: I don't think I can actually assert anything at this location, since (as mentioned above) `SSL_IS_FIRST_HANDSHAKE()` is false already by this point.  Preventing calls to `SSL_new_session_ticket()` should be enough, though if we do end up adding an option to defer the initial batch of tickets until the first write we'll need to check whether `sent_tickets` should be updated for all tickets (not just initial ones)

test/sslapitest.c
kaduk: I do recall deciding to not test stateful tickets, but I don't remember why, so it clearly can't have been a good reason.
(added.)

test/sslapitest.c
kaduk: Seems to be an artifact of some changes I made while debugging the initial work; it should be a real 1-byte write now.

test/sslapitest.c
kaduk: Same debugging artifact -- I can't check the value that I read when the read fails because the write was zero-length!

test/sslapitest.c
kaduk: It does; thanks!

doc/man3/SSL_CTX_set_num_tickets.pod
```diff
@@ -40,14 +42,29 @@ handshake then SSL_set_num_tickets() can be called again prior to calling
...
+The ticket generation and transmission are delayed until the server is
+starting a new write operation, so that it is bundled with other application
+data being written and properly aligned to a record boundary.
+SSL_new_session_ticket() can be called more than once to request additional
+tickets be sent; all such requests are queued and written together when it is
+safe to do so.  Note that a successful return from SSL_new_session_ticket()
+indicates only that the request to send a ticket was processed, not that the
+ticket itself was sent.  To be notified when the ticket itself is sent, a
+new-session callback can be registered with L<SSL_CTX_sess_set_new_cb(3)> that
+will be invoked as the ticket or tickets are generated.
```
mattcaswell, line 58: It's kind of implied, but perhaps make it clearer that this is only valid on the server side - you can't use this on the client side to request the server send you a ticket!

ssl/record/rec_layer_s3.c
```diff
@@ -387,7 +387,8 @@ int ssl3_write_bytes(SSL *s, int type, const void *buf_, size_t len,
      * If we are supposed to be sending a KeyUpdate then go into init unless we
      * have writes pending - in which case we should finish doing that first.
```
mattcaswell, line 388: Update this comment to say we also go into init if we have pending tickets to send?

ssl/statem/statem_srvr.c
```diff
@@ -437,6 +437,11 @@ static WRITE_TRAN ossl_statem_server13_write_transition(SSL *s)
             st->hand_state = TLS_ST_SW_CERT_REQ;
             return WRITE_TRAN_CONTINUE;
         }
+        if (s->ext.extra_tickets_expected > 0) {
+            st->hand_state = TLS_ST_SW_SESSION_TICKET;
+            s->ext.extra_tickets_expected--;
```
mattcaswell, line 442: Is it feasible to move this decrement into the ticket processing itself?  I've tried (not 100% successfully) to avoid side effects in the transition functions as much as possible.

ssl/statem/statem_srvr.c
```diff
@@ -527,7 +532,9 @@ static WRITE_TRAN ossl_statem_server13_write_transition(SSL *s)
          * Following an initial handshake we send the number of tickets we have
          * been configured for.
          */
-        if (s->hit || s->num_tickets <= s->sent_tickets) {
+        if (!SSL_IS_FIRST_HANDSHAKE(s) && s->ext.extra_tickets_expected > 0) {
+            s->ext.extra_tickets_expected--;
```
mattcaswell, line 536: Similarly here it would be good to avoid side effects if possible.

ssl/statem/statem_srvr.c
kaduk: Sure; I had gone this route since I didn't want to put the "random side effect" into the ticket-generation function, but avoiding side-effects in the transition functions seems like a better thing to prioritize.

kaduk:
New (final?) fixup commit pushed.

test/sslapitest.c
kaduk: I'll note for posterity that adding this test revealed a bug in the code that checked if writes were pending that mis-grouped the boolean expressions.

kaduk:
Squashed and pushed to master.
I took the liberty of fixing a stale comment in sslapitest.c that claimed cctx still had the `new_cachesession_cb` but it has actually been changed to use `new_session_cb`.

PR #11421, opened 2020-03-27
ASYNC: Fixes for nested job creation

uudiin:
In such a scenario, calling ASYNC_start_job() in a job again to create
a new job will cause the program to crash. This scenario is common in
coroutines, such as a web server, a coroutine listener, and a new client,
a new coroutine service, after accept () is created.

The current design of ASYNC does not support this scenario. Each coroutine
eventually falls back to the dispatcher. There are multiple coroutines,
and there is only one context for saving the coroutine exit, which is the
dispatcher. At this time, the context will be Being covered, this also
determines that only one coroutine can be running at the same time. And it
is dispatched by the dispatcher, which limits the application scenario
of ASYNC to a certain extent. In the traditional sense of coroutines,
dispatching is done by the users themselves, and there is no need for a
so-called dispatching center, dispatcher.

This patch solves this problem. For each coroutine, a context called back
is used for exit or pause (), so the dispatcher is not needed. This patch
does not change the ASYNC API.

There is also a simple example in the test code, which fully illustrates the
problems I described.

Signed-off-by: Tianjia Zhang <tianjia.zhang@linux.alibaba.com>

##### Checklist

### DIFF
--- a/crypto/async/arch/async_win.c
+++ b/crypto/async/arch/async_win.c
@@ -22,32 +22,13 @@ int ASYNC_is_capable(void)
 
 void async_local_cleanup(void)
 {
-    async_ctx *ctx = async_get_ctx();
-    if (ctx != NULL) {
-        async_fibre *fibre = &ctx->dispatcher;
-        if (fibre != NULL && fibre->fibre != NULL && fibre->converted) {
-            ConvertFiberToThread();
-            fibre->fibre = NULL;
-        }
-    }
+    if (GetCurrentFiber())
+        ConvertFiberToThread();
 }
 
-int async_fibre_init_dispatcher(async_fibre *fibre)
+int async_fibre_init_dispatcher(async_ctx *ctx)
 {
-# if defined(_WIN32_WINNT) && _WIN32_WINNT >= 0x600
-    fibre->fibre = ConvertThreadToFiberEx(NULL, FIBER_FLAG_FLOAT_SWITCH);
-# else
-    fibre->fibre = ConvertThreadToFiber(NULL);
-# endif
-    if (fibre->fibre == NULL) {
-        fibre->converted = 0;
-        fibre->fibre = GetCurrentFiber();
-        if (fibre->fibre == NULL)
-            return 0;
-    } else {
-        fibre->converted = 1;
-    }
-
+    ConvertThreadToFiber(NULL);
     return 1;
 }
 
--- a/crypto/async/arch/async_win.h
+++ b/crypto/async/arch/async_win.h
@@ -25,7 +25,7 @@ typedef struct async_fibre_st {
 } async_fibre;
 
 # define async_fibre_swapcontext(o,n,r) \
-        (SwitchToFiber((n)->fibre), 1)
+        ((o)->fibre = GetCurrentFiber(), SwitchToFiber((n)->fibre), 1)
 
 # if defined(_WIN32_WINNT) && _WIN32_WINNT >= 0x600
 #   define async_fibre_makecontext(c) \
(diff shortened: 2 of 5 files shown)


uudiin:
@InfoHunter @mattcaswell Could you give me some feedback on this patch?

InfoHunter:
I still don't quite understand why the nested jobs are needed...

uudiin:
> I still don't quite understand why the nested jobs are needed...

As I described above, it will be used in the ssl server scenario:
Open a job for `listen`, after `accept` a client and then open a job to communicate with the client

InfoHunter:
Why don't you just start a new job after `accept` without opening a job for `listen`?

uudiin:
> Why don't you just start a new job after `accept` without opening a job for `listen`?

`accept` is also a blocking operation, to achieve asynchronous, you need yield coroutine, which is `ASYNC_pause_job()`, so it is necessary to let the combination of `listen + accept` run in a async job.

uudiin:
@mattcaswell @iamamoose ping

mattcaswell:
IMO, nested async jobs are not a good idea. The concept was deliberately excluded from the original design. It adds too much complexity (high risk of something going wrong) - with little benefit.

I don't think this PR should be accepted.

t8m:
> IMO, nested async jobs are not a good idea. The concept was deliberately excluded from the original design. It adds too much complexity (high risk of something going wrong) - with little benefit.
> 
> I don't think this PR should be accepted.

I agree. Would you put OTC hold on it and call a vote?

mattcaswell:
> Would you put OTC hold on it and call a vote?

I put the hold on it. Do we actually need to call a vote to reject a PR? I'd say a vote is only required if there is disagreement within the committers.

paulidale:
I don't think a vote is required unless a committer (or alternatively a committee member) wants this to go in.  In that case, they could call the vote.

t8m:
I do not think a vote to reject a PR is always needed. For example if some PR is some total nonsense or the contributor does not respond to requests for changes it is clearly unneeded.

But in this concrete case where the PR is seemingly valid but we do not want to support the feature for technical reasons it would make sense to me.
The vote can be more general than just about closing this PR.  For example it could be a vote that we are not going to support nested ASYNC jobs.

mattcaswell:
> But in this concrete case where the PR is seemingly valid but we do not want to support the feature for technical reasons it would make sense to me.

Seems like we need to discuss policy. Lets see if we have time at the next OTC meeting.

InfoHunter:
> I don't think a vote is required unless a committer (or alternatively a committee member) wants this to go in

Personally, I think this PR needs more actions other than just rejecting it:

1. Do we really need this feature, if not:
2. Do we need to forbid this nested usage explicitly other than let the application crashes.

BTW, the author of this PR and me are in the affiliated companies (Alibaba & Ant Group), so perhaps my comment should only be consider as a note instead of a series opinion ;-)

mattcaswell:
> Do we really need this feature, if not:

I've not yet heard a compelling argument for it, so I don't think so.

> Do we need to forbid this nested usage explicitly other than let the application crashes.

No, it shouldn't crash. I think a PR that failed gracefully in that scenario would be acceptable.

mattcaswell:
OTC: Moving this to the Post 3.0 milestone

uudiin:
@mattcaswell @t8m @paulidale This PR has already rebased on master and combine the PR https://github.com/openssl/openssl/pull/16703

InfoHunter:
Ping @openssl/committers @openssl/otc , this is stale for a very long time, any ideas on this guys?

t8m:
IMO there was a relevant reply by Matt here: https://github.com/openssl/openssl/pull/11421#issuecomment-733574384

mattcaswell:
As I've previously noted in https://github.com/openssl/openssl/pull/11421#issuecomment-725984292 I don't think this PR should be accepted.

Issue #11413, opened 2020-03-26
CMS SignerInfo signatureAlgorithm, what gives?


levitte:
I'm a little lost on what the signatureAlgorithm is supposed to be, and it seems to vary depending on key algorithm, at least looking at our implementation.  Looking at the code we produce a Subject Public Key Algorithm for RSA (https://tools.ietf.org/html/rfc3279#section-2.3.1), while we produce a Signature Algorithm for DSA (https://tools.ietf.org/html/rfc3279#section-2.2.2) and ECDSA (https://tools.ietf.org/html/rfc3279#section-2.2.3).

Evidence for RSA:

https://github.com/openssl/openssl/blob/8158cf209792f7a92f0812ac89a9f54950e8453b/crypto/rsa/rsa_ameth.c#L812-L815
https://github.com/openssl/openssl/blob/8158cf209792f7a92f0812ac89a9f54950e8453b/crypto/rsa/rsa_ameth.c#L822

Evidence for DSA (`snid` is the composed digest-sign NID):

https://github.com/openssl/openssl/blob/8158cf209792f7a92f0812ac89a9f54950e8453b/crypto/dsa/dsa_ameth.c#L492-L497

Evidence for ECDSA:

https://github.com/openssl/openssl/blob/8158cf209792f7a92f0812ac89a9f54950e8453b/crypto/ec/ec_ameth.c#L493-L498

So I wonder, why the difference?  And most of all, I wonder why the digest-sign algid should be used at all, considering that the SignerInfo structure has a separate digestAlgorithm and [the Message Signature Verification Process isn't interest in a digest-sign algorithm at all](https://tools.ietf.org/html/rfc2630#section-5.5)

Is this possibly a bug?  Otherwise, what am I missing?  Am I reading the code wrong?
If this is indeed an correct inconsistency, is this documented in a standard somewhere?

levitte:
@DDvO, you might be the guru du jour, can you shed some light on this?

levitte:
I also looked at the GOST implementation, and it seems to produce Subject Public Key Algorithm (https://tools.ietf.org/html/rfc4491#section-2.3).  Evidence:

https://github.com/gost-engine/engine/blob/7c30097805cba0c62555493df6dad9f0c5d1d0f3/gost_ameth.c#L258-L261

levitte:
It's also a bit interesting to find diverse comments referring to implementations using signature OIDs as "broken":

[...]

PR #11422, opened 2020-03-27
[WIP] PKCS#7 and CMS with provider side keys

levitte:
Provider side keys are a bit of a challenge for code that relies on being able to use EVP_PKEY_METHOD and EVP_PKEY_ASN1_METHOD functions directly.

However, it turns out that for the PKCS#7 and CMS, the controls and functions that existed specifically for them were all only used to create relevant `X509_ALGOR` when and where needed.  This was mainly enabled by passing `libcrypto` structures directly to the method code and letting them deal with those as they wish (potentially opening up for all sorts of more or less dirty hackery).

With providers, it's not possible to share `libcrypto` internal structures with the provider implementations, so we need another mechanism to achieve the objective, i.e. getting relevant `X509_ALGOR` for the `EVP_PKEY` or operation (via contexts) that's used.

This mechanism was already introduced in 6f4b7663150e2ee264836dbc5b338bd979828312, but wasn't explained in general terms (or at all), nor was it yet made for general use.  The mechanism involves getting a DER encoded AlgorithmIdentifier from the provider implementation as an `OSSL_PARAM` with the name `"algorithm-id"` (macro **OSSL_PKEY_PARAM_ALGORITHM_ID**) and decoding it with `d2i_X509_ALGOR()` in `libcrypto`.  In this PR, we introduce general internal functionality to get the `X509_ALGOR` for any operation.

Related to this is the question asked in #11413, where it turns out that CMS has its own set of standards for diverse cases when an `X509_ALGOR` is needed, which may or may not match what you'd normally get for X.509, so we need a different parameter name specifically for CMS; for the moment, I've chosen `"algorithm-id:CMS"` (macro **OSSL_CMS_PARAM_ALGORITHM_ID**).  The CMS code is made to try that first, and uses the "normal" parameter name as a fallback (because some CMS standards have signature algorithms that match the X.509 standards, so might as well).

##### Checklist


levitte:
This is currently in WIP because I haven't yet modified the relevant provider implementations to respond to a requested algorithm identifier.

levitte:
Regardless of WIP or Travis failures, I'm interested in early comments on the ideas introduced in this PR

crypto/cms/cms_sd.c
```diff
@@ -223,10 +224,39 @@ int cms_SignerIdentifier_cert_cmp(CMS_SignerIdentifier *sid, X509 *cert)
...
+     * evp_pkey_ctx_get_algid().  When the OID should specify the keyalg
+     * only, the response is allowed to come from either that or
+     * evp_pkey_get_algid().
+     */
+    alg = evp_pkey_ctx_get_algid(pctx, OSSL_CMS_PARAM_ALGORITHM_ID);
+    if (alg == NULL)
+        alg = evp_pkey_get_algid(pkey, OSSL_CMS_PARAM_ALGORITHM_ID);
+    /* Try the common X509_ALGOR as a fallback */
+    if (alg == NULL)
+        alg = evp_pkey_ctx_get_algid(pctx, OSSL_PKEY_PARAM_ALGORITHM_ID);
```
mattcaswell: So we ask the pctx using the cms specific param, then the pkey using the cms specific param, and then the pctx using the generic param. But we never ask the pkey using the generic param. Why do we never ask the pkey for the generic param?

mattcaswell (changes requested):

crypto/cms/cms_env.c
```diff
@@ -1031,26 +1073,57 @@ BIO *cms_EnvelopedData_init_bio(CMS_ContentInfo *cms)
...
+    EVP_PKEY_CTX_free(pctx);
+    return r;
+#endif
 }
 
 int cms_pkey_is_ri_type_supported(EVP_PKEY *pk, int ri_type)
 {
     int supportedRiType;
 
+#if 0                            /* Why is this needed? */
```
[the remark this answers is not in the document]
beldmit: Sure.

New specification of Russian GOST allows more than one mechanism of the shared key transmission: KeyTransport, KeyAgreement, KeyEncryption. Not all of them are mandatory for implementation so this is a way to request whether the implementation used supports this mechanism.

crypto/cms/cms_env.c
```diff
@@ -1031,26 +1073,57 @@ BIO *cms_EnvelopedData_init_bio(CMS_ContentInfo *cms)
  */
 int cms_pkey_get_ri_type(EVP_PKEY *pk)
 {
+#if 0                            /* This should just be removed */
```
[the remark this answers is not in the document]
levitte: This is my way to open that very question, i.e. comments, acclamations, etc welcome

crypto/cms/cms_env.c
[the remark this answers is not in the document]
levitte: What I'm thinking is that each of these mechanisms use different methods, so probing with diverse inits should give just as good an answer, and releaves the need to keep details in sync for the provider author.

crypto/cms/cms_sd.c
levitte: I'm wondering if that's needed. How's an RSA key response for CMS (OID rsaEncryption) going to differ a response for X.509 (OID rsaEncryption)? Do we expect to see anything where that will differ?

But on the other, maybe as well cover all bases just to be safe, is that what you're thinking?

crypto/asn1/a_sign.c
```diff
@@ -157,47 +159,28 @@ int ASN1_item_sign_ctx(const ASN1_ITEM *it,
...
-        if (EVP_PKEY_CTX_get_params(pctx, params) <= 0)
-            goto err;
-
-        if ((aid_len = params[0].return_size) == 0) {
-            ASN1err(ASN1_F_ASN1_ITEM_SIGN_CTX,
-                    ASN1_R_DIGEST_AND_KEY_TYPE_NOT_SUPPORTED);
+        if (((alg1 =
+              evp_pkey_ctx_get_algid(pctx, OSSL_PKEY_PARAM_ALGORITHM_ID))
+             == NULL)
+            || (algor2 != NULL && (alg2 = X509_ALGOR_dup(alg1)) == NULL)) {
```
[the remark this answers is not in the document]
levitte: As far as I have seen, all legacy backends that do the corresponding thing do exactly this. I have no idea why we're asking for two nowadays. I think there was an idea that some structure might contain two different algorithmidentifiers for the same key, but this doesn't seem to have turned up, or asked for that way.

I have it my mind that these functions need deeper refactoring. I hope to get there before b1.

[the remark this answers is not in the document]
mattcaswell: ok

crypto/cms/cms_env.c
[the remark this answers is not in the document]
mattcaswell: It seems reasonable. @beldmit does this work for you?

crypto/cms/cms_sd.c
mattcaswell: > Do we expect to see anything where that will differ?

Perhaps not. I'm not sure.

> But on the other, maybe as well cover all bases just to be safe, is that what you're thinking?

Well yes - it was just that it stuck out as being a bit strange. I could live with the answer being that we don't think that is ever needed...as long we have actually thought about it and believe that is most likely to be the case.

crypto/cms/cms_env.c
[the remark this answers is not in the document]
mattcaswell: Probing seems reasonable. Again I'd like to hear from @beldmit.

crypto/cms/cms_env.c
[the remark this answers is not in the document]
beldmit: Will check.

crypto/cms/cms_sd.c
levitte: I can flesh out the comment if that helps

mattcaswell: ok

levitte: Actually, I'm doubling back on this.  The more I look at it, the more I see it as an omission.

doc/man3/EVP_PKEY_size.pod
```diff
@@ -37,11 +41,26 @@ of cryptographic length is specific to the key cryptosystem.
 EVP_PKEY_security_bits() returns the number of security bits of the given
 I<pkey>, bits of security is defined in NIST SP800-57.
 
+EVP_PKEY_get_params() gets arbitrary parameters from the I<pkey>.  The
```
slontis: Hmm - not sure we should keep this method.. The individual accessors are a lot less error prone.

kaduk:
https://tools.ietf.org/html/draft-housley-lamps-cms-update-alg-id-protect-00 (and list discussion thereof) may provide some additional clarity for the use of multiple algorithm IDs within CMS.

beldmit:
Well. I have many failing tests for CMS.

beldmit:
The diagnostics is 

```
Error decrypting CMS using private key
00:C1:24:36:68:7F:00:00:error:CMS routines:CMS_decrypt_set1_pkey_and_peer:no matching recipient:crypto/cms/cms_smime.c:677:
```

crypto/cms/cms_env.c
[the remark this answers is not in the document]
beldmit: Uncommenting this block fixes some tests.

beldmit:
I see two sets of problems for GOST.

1. GOST engine I use supports both key derivation and key encryption. The code https://github.com/openssl/openssl/pull/11422/files#diff-47b8520df0a2f8fa0433ff37611596f3R1092-R1099 enforces selection according to supported operation implicitly. 

Reordering the options (checking first for encryption, then for derivation) does the trick for GOST.
As support of `KeyTransRecipientInfo` is mandatory according to specification (see below quote in Russian, unfortunately, no text in English yet)
```
Совместимые реализации, оперирующие сообщениями с типом данных «конверт данных», должны поддерживать следующий функционал при зашифровании и расшифровании:
    • Обработка шифрованных сообщений с использованием эфемерного протокола Диффи-Хеллмана;
    • **Закодирование и раскодирование информации о получателях шифрованного сообщения в виде структур типа KeyTransRecipientInfo;**
```
I'd suggest to leave it this way.

2. The 2nd problem seems unrelated to this one. It prevents decryption for key agreement and I'm in process of investigations.

beldmit:
The 2nd problem is related to cms_pkey_is_ri_type_supported function. Now it returns the result of comparison with one exact value instead of checking for more than one in ctrl function.

@levitte, feel free to ask for more information. I currently have a reasonable test installation for its work.

crypto/cms/cms_env.c
```diff
@@ -1031,26 +1073,57 @@ BIO *cms_EnvelopedData_init_bio(CMS_ContentInfo *cms)
...
+     * after the other.
+     */
+    EVP_PKEY_CTX *pctx = EVP_PKEY_CTX_new(pk, NULL);
+    int r;
+
+    ERR_set_mark();
+    if (EVP_PKEY_derive_init(pctx) > 0)
+        r = CMS_RECIPINFO_AGREE;
+    else if (EVP_PKEY_encrypt_init(pctx) > 0)
+        r = CMS_RECIPINFO_TRANS;
```
beldmit: I suggest at least set this branch first.

beldmit (changes requested):

levitte:
#11450 is in, I've rebased, now I just gotta finish up

romen:
The travis failure seems relevant

```
$ if test -n "$CHECKDOCS" && ! $make doc-nits; then echo -e '\052\052 FAILED -- MAKE DOC-NITS'; travis_terminate 1; fi
/usr/bin/perl ./util/find-doc-nits -n -l -e
doc/man3/EVP_PKEY_size.pod:1: EVP_PKEY_gettable_params(3) also in NAME section of doc/man3/EVP_PKEY_gettable_params.pod
doc/man3/EVP_PKEY_size.pod:1: The following exist as other .pod files: EVP_PKEY_gettable_params
*** WARNING: No items in =over (at line 22) / =back list at line 27 in file doc/man3/X509_get0_distinguishing_id.pod
*** WARNING: No items in =over (at line 12) / =back list at line 52 in file doc/man7/EVP_MD-common.pod
*** WARNING: No items in =over (at line 19) / =back list at line 25 in file doc/man7/OSSL_PROVIDER-FIPS.pod
*** WARNING: No items in =over (at line 18) / =back list at line 22 in file doc/man7/OSSL_PROVIDER-default.pod
*** WARNING: No items in =over (at line 21) / =back list at line 25 in file doc/man7/OSSL_PROVIDER-legacy.pod
Makefile:1812: recipe for target 'doc-nits' failed
make: *** [doc-nits] Error 1
** FAILED -- MAKE DOC-NITS
```

crypto/cms/cms_env.c
```diff
@@ -53,19 +54,60 @@ static CMS_EnvelopedData *cms_enveloped_data_init(CMS_ContentInfo *cms)
...
+    }
+
+    if (pkey->keymgmt == NULL)
+        goto legacy;
+
+    /* Compensate for not setting this up in the CMS_RECIPINFO_TRANS case */
+    if (pctx == NULL) {
+        if ((pctx = alloc_pctx = EVP_PKEY_CTX_new(pkey, NULL)) == NULL
+            || EVP_PKEY_encrypt_init(pctx) <= 0)
+            goto end;
```
romen, line 83: Should we raise some special error for this case?

crypto/cms/cms_env.c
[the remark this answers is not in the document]
romen: The approach proposed in the `#else` branch seems fine to me, we should probably just go ahead and drop this `#if 0`.




> This is my way to open that very question, i.e. comments, _acclamations_, etc welcome

@levitte Hip hip hooray !!!

crypto/cms/cms_env.c
romen: So it seems GOST case at least would be fixed by checking for encrypt first and derive later. Can we rationalize this and leave a comment explaining why a specific order is preferable?

crypto/cms/cms_env.c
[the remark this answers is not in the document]
romen: @beldmit do you have an update on this?

crypto/cms/cms_sd.c
```diff
@@ -327,8 +359,15 @@ CMS_SignerInfo *CMS_add1_signer(CMS_ContentInfo *cms,
...
-    if (!(flags & CMS_KEY_PARAM) && !cms_sd_asn1_ctrl(si, 0))
-        goto err;
+    if (!(flags & CMS_KEY_PARAM)) {
+        EVP_MD_CTX *tmp_mctx = EVP_MD_CTX_new();
+        int ok = (EVP_DigestSignInit(tmp_mctx, NULL, md, NULL, pk) > 0);
+
+        ok = ok && cms_sd_asn1_ctrl(si, EVP_MD_CTX_pkey_ctx(tmp_mctx), 0);
+        EVP_MD_CTX_free(tmp_mctx);
+        if (!ok)
+            goto err;
```
romen, line 369: Should we raise a specific error for this case?

doc/man3/EVP_PKEY_size.pod
```diff
@@ -2,7 +2,8 @@
 
 =head1 NAME
 
-EVP_PKEY_size, EVP_PKEY_bits, EVP_PKEY_security_bits
+EVP_PKEY_size, EVP_PKEY_bits, EVP_PKEY_security_bits,
+EVP_PKEY_get_params, EVP_PKEY_gettable_params
```
romen, line 6: Why in this file rather than adding to `EVP_PKEY_gettable_params.pod`?

I am not opposed to unify them in a single pod file, but I am not sure I see the logical connection between `EVP_PKEY_size` and `EVP_PKEY_get_params`.

If we unify, EVP_PKEY_gettable_params.pod` should be removed.

doc/man3/EVP_PKEY_size.pod
```diff
@@ -37,11 +41,26 @@ of cryptographic length is specific to the key cryptosystem.
...
+
 =head1 RETURN VALUES
 
 EVP_PKEY_size(), EVP_PKEY_bits() and EVP_PKEY_security_bits() return a
 positive number, or 0 if this size isn't available.
 
+EVP_PKEY_get_params() returns 1 for success or 0 otherwise.
+EVP_PKEY_gettable_params() returns an OSSL_PARAM array on success or NULL on
+error.
+
```
romen, line 63: This file is missing a `HISTORY` section: if we unify the documentation into this `.pod` file we should add one and document that the new functions are added in `3.0`

include/openssl/core_names.h
```diff
@@ -182,6 +183,29 @@ extern "C" {
...
+ * to RFC 3279 with updates as well as other similar standards.  They may be
+ * used with a KEYMGMT to get the Subject Public Key Algorithm as well as
+ * with diverse other PKEY operations, such as getting a Signature Algorithm
+ * from a SIGNATURE operation.
+ *
+ * The value is expected to be an octet string with the DER encoded
+ * AlgorithmIdentifier for that operation or keypair.
+ *
+ * Supporting this parameter is not mandatory, but will determine whether
+ * it can be used for certain purposes.
```
romen, line 197: Nit: This phrasing seems a bit vague (and I get this is intentional). I think it might benefit both us in the future and authors of external providers if we at least add some kind of example to give an idea of what kind of purposes would benefit from enabling this.

My proposal would be for example to add, just before the end `(e.g., certain CMS operations might require support of this parameter or OSSL_CMS_PARAM_ALGORITHM_ID, as described further down).`

romen (changes requested):
I really like this refactoring!

I have a few comments, and a ping for @beldmit for comments to see what we can do to move this forward.

beldmit:
@levitte, @romen - could you please remind me how to fetch the current state of this PR to make tests? There were hints at the top of the page when a review was requested, but they've disappeared :(
Many thanks!

beldmit:
Found the commands in history.
```
git fetch origin refs/pull/11422/head
git checkout -b levitte/cms-with-providers FETCH_HEAD
```

beldmit:
After introducing the change I suggested earlier I get two failures on decrypting messages with KeyAgreement. Diagnostics is 

```
STDERR CONTENTS:
Error decrypting CMS using private key
00:B1:5A:6F:2A:7F:00:00:error:CMS routines:CMS_decrypt_set1_pkey_and_peer:no matching recipient:crypto/cms/cms_smime.c:677:
```
My engine provides decryption for both KeyTransport and KeyAgreement, and encryption in KeyTransport

doc/man3/EVP_PKEY_size.pod
levitte: `EVP_PKEY_gettable_params()` was developed separately by both @slontis and me, and our approaches regarding the manual were different.  I haven't dealt with it, 'cause I'm ambivalent which way to go.

levitte:
@beldmit, if you look at the bottom box at the end of any PR, where you see all the CIs among others, you will see the text "or view command line instructions" near the bottom of that box, just by the merge button.  Click that link and you have very official instructions.

doc/man3/EVP_PKEY_size.pod
levitte: I don't see that one needs to be gone in favor of the other...  while the individual accessors are absolutely fine if all you need is one item, I suspect that the churn is quite high as soon as you need to get more than one item in one go.

crypto/cms/cms_env.c
levitte: I have nothing against switching them around, but what will happen then with other implementations that have stronger derive than encrypt / decrypt support?  It's basically the same problem as @beldmit describes, but the other way around.

Although, I don't know CMS very well, so it's possible encrypt should always be used first and derive as fallback...  if someone tells me that's always the case, I'll happily switch.

Either way, I could use some help formulating a comment to explain the order we choose.

levitte:
@romen:
> I really like this refactoring!

Thanks :-)

crypto/cms/cms_env.c
levitte: Because?  Mind you, there are error codes to say that this and that other library had an error and serves as a trace...  unfortunately, I recently discovered that those codes are messed up...

crypto/cms/cms_sd.c
levitte: This is a situation where we usually count on the called function to use that already...

crypto/cms/cms_env.c
beldmit: Well, the problem is deeper.
We potentially have more than one supported format both for encryption and description. In the case of decryption, I think it's more reasonable to delegate the support to an engine/provider. For encryption, we currently don't have appropriate control function call to choose the preferred way and don't have any implementation that requires such a control.

levitte: > In the case of decryption, I think it's more reasonable to delegate the support to an engine/provider.

You mean to still ask the engine/provider the direct question instead of probing?  I'm *not* too keen on insert CMS specific hackery in providers, this PR is an effort to get away from that...  but will have to give this some though.  A capability probe maybe?

If I understood a little better what possibilities there are and on what conditions, that might help.

beldmit: I'm not aware of all the possible options either. RFC 4490 describes both variants for GOST, but in practice, we met only KeyTransport (partially because KeyAgreement was not implemented in OpenSSL at all). Do I understand correctly that RSA, DSA, and EC provide only one variant?

BTW, it's worth mentioning that capability of key encryption/derivation does not say us anything about CMS support.

levitte: I'm not sure what "CMS support" should be needed on the provider side.  After all, key derivation with the same algorithms is done the same way regardless, and so is encryption, right?  After all, looking at our EVP_PKEY_ASN1_METHODs for DH, DSA, EC and RSA, their ASN1_PKEY_CTRL_CMS_RI_TYPE responses are, respectively, CMS_RECIPINFO_AGREE, CMS_RECIPINFO_NONE, CMS_RECIPINFO_AGREE and CMS_RECIPINFO_TRANS.  To me, those look like the absolutely obvious answer, which could just as well have been probed by looking if the corresponding EVP_PKEY_METHOD has a non-NULL `encrypt_init` or `derive_init`...

Looking at the old code, this function defaults to KeyTransport if there's no response on ASN1_PKEY_CTRL_CMS_RI_TYPE, so I guess that's how this new code differs, and since there a number of functions in gost_ameth.c that support both  derivation and encryption, and that's likely to be the case in a provider as well, then the sudden change of primary choice may come as a bit of a surprise, I get that.

So I guess that's the hard question to ask, then...  in case of ambiguity, who should have the power to decide which method should take priority?

levitte: I'll ponder this over-night...  or I guess that'll be in the morning, 'cause I'm not very awake for the moment.

levitte: In the interest of expedience, I've simply switched them around so KeyTransport is probed first, and added a TODO comment that we should investigate that further.  That will allow us to move forward with this PR, and perhaps let our brains ruminate on the problem for some time and revisit it later.

levitte:
@beldmit, the diagnostics you're showing, they are with the current GOST engine, what's available on github, right?  Is that the test suite?

doc/man3/EVP_PKEY_size.pod
levitte: Fixed

include/openssl/core_names.h
levitte: Added

beldmit:
> @beldmit, the diagnostics you're showing, they are with the current GOST engine, what's available on github, right? Is that the test suite?

Unfortunately no, it's currently available in the commercial implementation and tests are run semi-manually.

mattcaswell:
Given this PR is on the critical path for alpha1 we need to get it to a state where our own tests are passing. Any other issues should be marked as a "known issue" for future work to fix.

beldmit:
> Given this PR is on the critical path for alpha1 we need to get it to a state where our own tests are passing. Any other issues should be marked as a "known issue" for future work to fix.

I can live with it if we change the order of checks for now and fix the 2nd problem before the release.

levitte:
> I can live with it if we change the order of checks for now and fix the 2nd problem before the release.

At the very least, I hope that this PR will give us a foundation to work from.

mattcaswell:
Is this still WIP?

levitte:
> Is this still WIP?

Yes.  I'm currently testing it merged with #11328, I need to ensure that works first.

levitte:
However, please feel welcome to comment what you see!

levitte:
Oh...  wow!  The code that's in right now works flawlessly when merged with #11328

levitte:
> Oh... wow! The code that's in right now works flawlessly when merged with #11328

... heh, that says nothing.  Since it's merged with EC key generation, and we only tell EC with key derivation, for which I've added an EVP_PKEY downgrade hack, it reduces the whole test suite to legacy key only.  I wonder what happens if I merge in the DSA key generation PR, which is around, er, somewhere...  yes, in #11303.

slontis (commented):
Richard I am not going to block this PR,, but have some concerns about whether EVP_PKEY_get_params() should be public or even exist at all..
The one case where you are using it currently it uses a single parameter?

There were multiple reason(s) why I did not add it when I added the individual accessors.
(which I did send around in a mail when I was doing the work).
i.e:
  - It has some issues related to BIGNUM's (to me this is a fairly significant issue).
  - figuring out which params have been silently ignored may be important. The individual API's all set the return size internally. I see you have done this trick internally where you have used it. (A end user may not be so fortunate to understand this intricacy)
  - When it fails it may be harder to figure out which entry failed.

levitte:
I think we need to discuss the issues with OSSL_PARAM / BIGNUM elsewhere.  Let's no hijack this PR (or any other PR) for that discussion.

paulidale:
Elsewhere we have `EVP_XX_CTX_[gs]et_params`.  Here the names are ass-about.  Can we work on a better name for a start?  Just because the underlying naming is rotten does not mean we need to perpetuate the rot.

The BIGNUM concerns do need a separate discussion.  Agreed that it will not happen here.  BIGNUMs are **_difficult_** with the existing OSSL_PARAM framework.  The framework **mandates** extra buffers of the right size before knowing what the actual values are.  My opinion is that OSSL_PARAM is broken with respects to this :(  I understand why we're byte reordering but it's pain for minimal benefit.

Figuring out which OSSL_PARAMs have been used is a two directional affair, we will **need** to address this at some point, again not here.

levitte:
> BIGNUMs are **_difficult_**

I think that sums it up.  It comes down to lack of language support.

levitte:
... damnit.  NOT hijack this PR, @levitte...

paulidale:
> ... damnit. NOT hijack this PR, @levitte...

We understood :)

Time is too pressing to hold things up.

h-vetinari:
Is this a topic for beta1...?

t8m:
Is this still relevant? IMO not.

PR #11425, opened 2020-03-27
Give a better error if we can't find a sig alg

mattcaswell:
Some scenarios where we could not find a suitable sig alg just
gave "internal error" as the reason - which isn't very helpful. A
more suitable reason code already exists - so we use that.

Found while trying to make sslapitest work with the FIPS provider - but the same issue exists in 1.1.1 and can be applied there too.


slontis:
Does this break compatibility?

paulidale:
Error codes have not been considered part of the API.  The error existing is, but the specifics not.

paulidale:
Good for 1.1.1 as well.

t8m:
Well in some API call cases the error code could be considered part of the API (or at least the change of it should have a very good reason). However in this particular case I think the change is pretty safe.

mattcaswell:
Pushed. Thanks.

PR #11426, opened 2020-03-27
Don't attempt to up-ref an EVP_CIPHER if it is NULL

mattcaswell:
EVP_CIPHERs in the ssl_cipher_methods table can be NULL if
they are not available. We shouldn't attempt to up-ref a
cipher if it is NULL.


mattcaswell:
Pushed. Thanks.

PR #11427, opened 2020-03-27
Fix off-by-1 bug on provider_activate with custom error strings

romen:
Starting `cnt` from 1 would work if we weren't using cnt itself to access elements of the array returned calling the provider callback.

As it is before this commit, we have 2 problems:
- first, in the unlikely case that the incoming array was "empty" (only contains the terminator item) we would skip past it and potentially end up with oob reads;
- otherwise, at the end of the while loop, `cnt` will be equal to the number of items in the input array, not 1 more. We then add 1 more to the zalloc call to account for the library name item, and we fill all of it (relying on zalloc to have zeroed the terminator item).
  The first read access that will read the list up to the terminator will result in a OOB read as we did not allocate enough space to also contain the terminator.

##### Checklist

### DIFF
--- a/crypto/provider_core.c
+++ b/crypto/provider_core.c
@@ -540,12 +540,13 @@ static int provider_activate(OSSL_PROVIDER *prov)
          * with the error library number, so we need to make a copy of that
          * array either way.
          */
-        cnt = 1;                 /* One for the terminating item */
+        cnt = 0;
         while (reasonstrings[cnt].id != 0) {
             if (ERR_GET_LIB(reasonstrings[cnt].id) != 0)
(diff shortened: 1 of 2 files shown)


romen:
Ping to @p-steuer that brought this item to my attention while working on testing with external providers!

@levitte @paulidale I believe one of you might want to review this!

romen:
Merged to `master` with:
- 551543e5e2 Add test for providers exposing OSSL_FUNC_PROVIDER_GET_REASON_STRINGS
- fd03868b34 Fix off-by-1 bug on provider_activate with custom error strings

Issue #8731, opened 2019-04-11
EC key checking needs random data


dlegaultbbry:
Deep in the EC key checking code (eventually ends up calling ec_wNAF_mul), the library wants to have access to random data. I was using openssl to check that a public key is sane on a system where /dev/[u]random is not available. And this fails miserably because DRBG just falls flat.

This is using 1.1.1b.

Any suggestions on being able to check EC keys for validity without the need for random numbers? Perhaps just skip that part if random numbers aren't available?

bernd-edlinger:
This is a coordinate blinding at work, in the case of a public key it would be unnecessary,
but this function does not know if the scalar is secret or not.
What function do you use to validate the key?

dlegaultbbry:
The call stack I could figure out is:

EVP_PKEY_public_check -> EC_KEY_check_key -> ec_key_simple_check_key -> EC_POINT_mul -> EC_POINTs_mul -> ec_wNAF_mul -> somewhere under here random numbers are required

bernd-edlinger:
would this work for you?
```
diff --git a/crypto/ec/ec_mult.c b/crypto/ec/ec_mult.c
index 755d644..c442fda 100644
--- a/crypto/ec/ec_mult.c
+++ b/crypto/ec/ec_mult.c
@@ -441,7 +441,7 @@ int ec_wNAF_mul(const EC_GROUP *group, EC_POINT *r, const BIGNUM *scalar,
          * scalar multiplication implementation based on a Montgomery ladder,
          * with various timing attack defenses.
          */
-        if ((scalar != NULL) && (num == 0)) {
+        if ((scalar != group->order) && (scalar != NULL) && (num == 0)) {
             /*-
              * In this case we want to compute scalar * GeneratorPoint: this
              * codepath is reached most prominently by (ephemeral) key
```

bernd-edlinger:
well better make that:
```
diff --git a/crypto/ec/ec_mult.c b/crypto/ec/ec_mult.c
index 755d644..968125f 100644
--- a/crypto/ec/ec_mult.c
+++ b/crypto/ec/ec_mult.c
@@ -441,7 +441,7 @@ int ec_wNAF_mul(const EC_GROUP *group, EC_POINT *r, const BIGNUM *scalar,
          * scalar multiplication implementation based on a Montgomery ladder,
          * with various timing attack defenses.
          */
-        if ((scalar != NULL) && (num == 0)) {
[...]

PR #11439, opened 2020-03-30
[crypto/ec] blind coordinates in ec_wNAF_mul for robustness

bbbrumley:
See #11435. This PR makes the `ec_wNAF_mul` algorithm state less predictable. It's a good cheap hardening against bug attacks.

Deja vu -- [it's 2011 all over again](https://marc.info/?l=openssl-dev&m=131194808413635).

### DIFF
--- a/crypto/ec/ec_mult.c
+++ b/crypto/ec/ec_mult.c
@@ -746,6 +746,20 @@ int ec_wNAF_mul(const EC_GROUP *group, EC_POINT *r, const BIGNUM *scalar,
                     if (r_is_at_infinity) {
                         if (!EC_POINT_copy(r, val_sub[i][digit >> 1]))
                             goto err;
+
+                        /*-
+                         * Apply coordinate blinding for EC_POINT.
+                         *
+                         * The underlying EC_METHOD can optionally implement this function:
+                         * ec_point_blind_coordinates() returns 0 in case of errors or 1 on
+                         * success or if coordinate blinding is not implemented for this
+                         * group.
+                         */
+                        if (!ec_point_blind_coordinates(group, r, ctx)) {
+                            ECerr(EC_F_EC_WNAF_MUL, EC_R_POINT_COORDINATES_BLIND_FAILURE);
+                            goto err;
+                        }
+
                         r_is_at_infinity = 0;
                     } else {
                         if (!EC_POINT_add
--- a/crypto/ec/ecp_smpl.c
+++ b/crypto/ec/ecp_smpl.c
@@ -1442,36 +1442,38 @@ int ec_GFp_simple_blind_coordinates(const EC_GROUP *group, EC_POINT *p,
     temp = BN_CTX_get(ctx);
     if (temp == NULL) {
         ECerr(EC_F_EC_GFP_SIMPLE_BLIND_COORDINATES, ERR_R_MALLOC_FAILURE);
-        goto err;
+        goto end;
     }
 
-    /* make sure lambda is not zero */
+    /*-
+     * Make sure lambda is not zero.
+     * If the RNG fails, we cannot blind but nevertheless want
+     * code to continue smoothly and not clobber the error stack.
+     */
     do {
-        if (!BN_priv_rand_range_ex(lambda, group->field, ctx)) {
-            ECerr(EC_F_EC_GFP_SIMPLE_BLIND_COORDINATES, ERR_R_BN_LIB);
-            goto err;
+        ERR_set_mark();
+        ret = BN_priv_rand_range_ex(lambda, group->field, ctx);
+        ERR_pop_to_mark();
+        if (ret == 0) {
+            ret = 1;
+            goto end;
         }
     } while (BN_is_zero(lambda));
 
     /* if field_encode defined convert between representations */
-    if (group->meth->field_encode != NULL
-        && !group->meth->field_encode(group, lambda, lambda, ctx))
-        goto err;
-    if (!group->meth->field_mul(group, p->Z, p->Z, lambda, ctx))
-        goto err;
-    if (!group->meth->field_sqr(group, temp, lambda, ctx))
-        goto err;
-    if (!group->meth->field_mul(group, p->X, p->X, temp, ctx))
-        goto err;
-    if (!group->meth->field_mul(group, temp, temp, lambda, ctx))
-        goto err;
-    if (!group->meth->field_mul(group, p->Y, p->Y, temp, ctx))
-        goto err;
-    p->Z_is_one = 0;
+    if ((group->meth->field_encode != NULL
+         && !group->meth->field_encode(group, lambda, lambda, ctx))
+        || !group->meth->field_mul(group, p->Z, p->Z, lambda, ctx)
+        || !group->meth->field_sqr(group, temp, lambda, ctx)
+        || !group->meth->field_mul(group, p->X, p->X, temp, ctx)
+        || !group->meth->field_mul(group, temp, temp, lambda, ctx)
+        || !group->meth->field_mul(group, p->Y, p->Y, temp, ctx))
+        goto end;
 
+    p->Z_is_one = 0;
     ret = 1;
 
- err:
+ end:
     BN_CTX_end(ctx);
     return ret;
 }


bbbrumley:
The CI failure doesn't really make sense to me.

t8m:
> The CI failure doesn't really make sense to me.

That's just an usual timeout unrelated to this PR.

mspncp:
> Deja vu -- [it's 2011 all over again](https://marc.info/?l=openssl-dev&m=131194808413635).

Thanks for the historical side note :-)

romen:
I don't think confirmation from an OMC member is required for backporting this PR to 1.1.1, as it is pretty clear that it is a security fix.

I want to know though from @mattcaswell if, once this PR and/or #11435 are ready to merge, we should prioritize them for merging into 1.1.1f or if we will use the normal 24h grace-period and miss the merge window.

crypto/ec/ec_mult.c
```diff
@@ -753,6 +753,20 @@ int ec_wNAF_mul(const EC_GROUP *group, EC_POINT *r, const BIGNUM *scalar,
...
+
+                        /*-
+                         * Apply coordinate blinding for EC_POINT.
+                         *
+                         * The underlying EC_METHOD can optionally implement this function:
+                         * ec_point_blind_coordinates() returns 0 in case of errors or 1 on
+                         * success or if coordinate blinding is not implemented for this
+                         * group.
+                         */
+                        if (!ec_point_blind_coordinates(group, r, ctx)) {
```
bernd-edlinger: I hope public keys do not need the blinding?
I remember I fixed a bug, where signature validation was failing because of failed
random number generator

bernd-edlinger: see 3051bf2a

bernd-edlinger: So, what I mean is, multiplication by group order never secret,
so at least those do not need blinding.

bernd-edlinger: Okay, unless I am mistaken this will break my fix above, since the case where
the exponent is the group order will not use the montgomery ladder, but
will instead be blinded (and fail if the random source is not available)
please do at least add a similar precaution, that the exponent is not the group
order, it need not be a sophisticated compare, a simple pointer compare is enough.

bbbrumley: > So, what I mean is, multiplication by group order never secret,
> so at least those do not need blinding.

Bug attacks aren't necessarily about secrets ;) They are about finding clever inputs that trigger bugs and do interesting things. Yes, one of these interesting things is recovering secrets. But others could be
1. Fooling signature verification, i.e. creating a forgery using the bug
2. Fooling key validation, like we did [in our first real world bug attack](https://eprint.iacr.org/2011/633) (independent from us stealing keys), [CVE-2011-4354](https://nvd.nist.gov/vuln/detail/CVE-2011-4354)
3. Other stuff I'm not clever enough to think of.

I wonder if there is a better solution here than disabling coordinate randomization? Does switching from `BN_priv_rand_range_ex` to something less hangy help? These blinding values are not long term keys -- so while ideally they are generated similarly, in practice the requirements are much less strict.

bernd-edlinger: Obviously I would be satisfied, if the randomization fails because of the
random generator fails, then just continue, because otherwise you
cannot even read in a public key file

bernd-edlinger: Nevertheless I think that it does not hurt to check the scalar in this way,
I do not propose to compare the value, I propose to compare the pointer,
so when openssl mutliplies P*order, I am pretty sure P is already the
public point.  an order is not by chance the same value, but really the
same object.

bbbrumley: > mutliplies P*order,

My point is order*P is a potential attack vector. (See above -- "fooling key validation")

bernd-edlinger: I need some time to think about your reaoning. hold the line.

bernd-edlinger: Okay, I see your point now,
What do you think of  3051bf2  ?
Probably it would be okay to keep that and do the coordinate-blinding
here instead, but if that fails with the "good" random generator 
also non-cryptographic one would be fine, or so, I believe I just
want to make sure that the software survives every possible error
condition from out of memory to out of entropy, to .... ?

bernd-edlinger: .. user not checking error codes, etc. etc...

romen: @bernd-edlinger I guess your point is that you want to make sure #8731 does not happen: users should be able to verify stuff without access to a random number generator.

Considering bug attacks, randomization is useful also during key validation or signature verification, so I believe the solution is not to avoid using randomness for certain corner cases. 

I have no experience with platforms without /dev/(u)random what does OpenSSL do in those cases? Does it hang indefinitely? Does it detect the condition and return an error?

I believe the problem should be centrally solved, not exempted for some cases.

bernd-edlinger: It does not hang forever, it start to fail in all possible ways, it is very likely
that functions that under normal conditions never fail do fail, and the
error condition is either ignored, because not anticipated, or the
untested error handling - in user code - makes everything even worse.
We openssl developers do never make mistakes, of course ;-)

bernd-edlinger: I think the coord-blinding function should try harder, when the normal random
number generator fails, it should just get desperate and use rand() for instance.

bernd-edlinger: @paulidale what did we use when the crypto engines failed to load the random provider?
I think you had some explicit bsd-type random generator, maybe we can use that
for coord-blinding if a similar error happens ?

t8m: We had the RAND_pseudo_bytes() for that purpose. Having call to rand() in the openssl library would probably trigger "interesting" questions from various 3rd party auditors or automated tools that review insecure library calls.

bernd-edlinger: RAND_pseudo_bytes can and will fail when there is no entropy to start the engine.

bernd-edlinger: furthermore it seems to be `#if OPENSSL_API_COMPAT < 0x10100000L`

bbbrumley: https://github.com/openssl/openssl/blob/beb958ccd87b95f1a86bfe2b879492b62e58d80e/crypto/ec/ecp_smpl.c#L1448-L1454

If I catch the error there and remove it from the stack, and just return success, is that an acceptable solution?

bernd-edlinger: Yes.

bernd-edlinger: You may also consider if what I did in 3051bf2 is still okay, of if you want to
have a similar solution in the ladder code.

bernd-edlinger: > an acceptable solution?

of course with a comment explaining the reasoning, so it wont be "fixed" in the future.

bbbrumley: > You may also consider if what I did in [3051bf2](https://github.com/openssl/openssl/commit/3051bf2afab7ac8b7b9c64e68755d1addd2fb8ff) is still okay,

I thought about that. But IMO 3051bf2 is still OK -- I'm a little hesitant to add "silent fails" to that ladder code only to handle this corner case.

richsalz:
```RAND_psuedo_bytes``` was made equivalant to ```RAND_bytes``` in 1.1.0

File: crypto/ec/ec_mult.c (hunk 1 above)
Comment (bernd-edlinger) [reply crypto/ec/ec_mult.c:765]: Thanks @bbbrumley

**Comment by romen** [comment]
Should we relabel err to end?

**Comment by bbbrumley** [comment]
> Thanks!

np.

I don't know what error to look for so peeking with `ERR_peek_last_error` and checking does not really help. But maybe it doesn't matter -- just let the code move forward.

**Comment by bernd-edlinger** [comment]
I think it does not matter, we are good now.

**Comment by bbbrumley** [comment]
> Should we relabel err to end?

Fixed in 902775a9379cf6d1679d67e73b2d2a268ff3ac6e.



> I think it does not matter, we are good now.

There was some other cleanup to do there when I took a closer look, so I went ahead and did the change.

**Comment by bernd-edlinger** [comment]
Okay, no problem, I really like what I see.  I hope this will not be a merge conflict,
but we can approve that quickly if necesary.

File: crypto/ec/ecp_smpl.c [hunk 2]
```diff
@@ -1442,36 +1442,38 @@ int ec_GFp_simple_blind_coordinates(const EC_GROUP *group, EC_POINT *p,
...
-    if (!group->meth->field_sqr(group, temp, lambda, ctx))
-        goto err;
-    if (!group->meth->field_mul(group, p->X, p->X, temp, ctx))
-        goto err;
-    if (!group->meth->field_mul(group, temp, temp, lambda, ctx))
-        goto err;
-    if (!group->meth->field_mul(group, p->Y, p->Y, temp, ctx))
-        goto err;
-    p->Z_is_one = 0;
+        || !group->meth->field_mul(group, p->Z, p->Z, lambda, ctx)
```
bernd-edlinger, line 1466: I am not sure about the indentation of || recently discussion was a bit ambivalent,
some do indent this || by 4 spaces, some dont, I thin here with the && above and the ||
the extra indentation would look funny.

crypto/ec/ecp_smpl.c
```diff
@@ -1442,36 +1442,38 @@ int ec_GFp_simple_blind_coordinates(const EC_GROUP *group, EC_POINT *p,
...
+        if (ret == 0) {
+            ret = 1;
+            goto end;
         }
     } while (BN_is_zero(lambda));
 
     /* if field_encode defined convert between representations */
-    if (group->meth->field_encode != NULL
+    if ((group->meth->field_encode != NULL
         && !group->meth->field_encode(group, lambda, lambda, ctx))
```
bernd-edlinger, line 1465: If we are pedantic, at least this && need to be indented

bernd-edlinger: by one space.

bbbrumley: Done ;)

bernd-edlinger:
BTW: is there also a ec2m blind coords ?

romen:
> BTW: is there also a ec2m blind coords ?

@bernd-edlinger Yes and no.

In the EC_METHOD in ec2_smpl.c only supports affine coordinates for its implementation of point addition and point doubling.
For single point multiplication or double point multiplication we use ec_scalar_mul_ladder (which uses ladder_pre, ladder_step and ladder_post): laddr_pre in this case includes blinding because ladder_step and ladder_post use a projective representation.
When we instead hit ec_wnaf_mul we rely on the pointer for blind_coordinates in the EC_METHOD, which for ec2_smpl.c. C is NULL because regular point addition and doubling used by the wnaf implementation can only work in affine.

crypto/ec/ec_mult.c
t8m: > RAND_pseudo_bytes can and will fail when there is no entropy to start the engine.

I know that. That's the reason why I used the past tense. And I am not sure we want to revive it just for this.

bernd-edlinger: If we want to use a non-cryptographic error source we should
Use @paulidale 's random source for instance.  I remember
he invented something for to be used *before* the random provider
is loaded.

paulidale: Wouldn't a "predicable" RNG defeat the blinding?

I used the [32 bit xorshift](https://doi.org/10.18637/jss.v008.i14) for the property cache flushing.

mspncp: For the records, it was done in commit f06cf3c4149c271d498764c2a071cb68b3d9a431.

paulidale: It would be easy enough to extract the xorshift RNG and its basic seeding if it would help.
Anyway, it would be a different PR if so.

bernd-edlinger: Maybe it would, after all this PR is to avoid potential bugs in the EC point multipication
and if you can hit a bug with one blinding value, maybe, but the next one will not.
So I guess you can still get make the attack much more difficult this way.

@bbbrumley ?

bbbrumley: @bernd-edlinger yes I think your instinct is correct.

But if I'm fully honest, I feel that if your RNG is failing, you have much bigger problems than bug attacks. So I don't feel the need to have a one-off solution here.

@romen said

> I believe the problem should be centrally solved, not exempted for some cases.

and I agree that spinning a solution just for this seems silly. If a `RAND_bytes_hopefully_but_at_least_something` function emerges at some point, I'm happy to make the change then.

mspncp: At least we already have a cool name for this function.

paulidale: `RAND_bytes_hopefully_but_at_least_something` is essentially what's being discussed.  Against bug attacks, it seems like it would be of use (caveat: some kind of seeding).  For some reason I was thinking timing.

I agree that a central solution is the way to go.

paulidale: > At least we already have a cool name for this function.

Yep, `RAND_psuedo_bytes` :grin:

bernd-edlinger: or RAND_paranoid_bytes :grinning:

paulidale: I don't think paranoid is a good choice for a function that mightn't be random.
`OSSL_blinding_bytes` avoiding the RAND connection.

richsalz:
The problem with ```RAND_psuedo_bytes```, ```RAND_do_not_use_this_we_mean_it```, etc., is that people will use it incorrectly. They will be insecure, CVE's will be found, and OpenSSL will catch some of the blame.

Thinking of the larger user community, a single, cryptographically-strong, PRNG cannot be misused and should be the only thing we provide.

paulidale:
@richsalz agreed.  The name needs to have nothing to do with randomness if we even attempt something.

bernd-edlinger:
Agree as well.  AND It should be an internal function.

paulidale:
There is no way we should export anything less than a cryptographically secure RNG.  @richsalz is perfectly correct on this front.

bernd-edlinger:
The side discussion is not about this PR but about possible future PRs
@romen please proceed merging this PR when you have time for it.

crypto/ec/ec_mult.c
bernd-edlinger: I just mean, we are a bit paranoid, to assume a bug-attack and a random-generator
failure at the same time.  But I think that is about the security level you need in
an atomic power plant.

bernd-edlinger: And I hope I did not infect you :), oops.

bernd-edlinger:
@romen 
> For single point multiplication or double point multiplication we use ec_scalar_mul_ladder 

Hmhmhm....
with my 3051bf2a I of course made an exception for scalar = group->order,
so they won't reach ec_scalar_mul_ladder.  However I don't care much about
ec2m I don't see that being used anywhere.

How about 25519, is that using blinding either?

romen:
@bernd-edlinger 

> How about 25519, is that using blinding either?

It doesn't seem like it looking at [curve25519.c](https://github.com/openssl/openssl/blob/3cb55fe47c3398b81956e4fe20c4004524d47519/crypto/ec/curve25519.c)

But curve25519/curve448 are "special" in a way, because of the way they are defined in their spec: although based on ECC, they are defined as functions rather than in terms of EC operations.

Anyway, my _personal_ take on 3051bf2afab7ac8b7b9c64e68755d1addd2fb8ff is that, once we solve somewhere else, for the whole library, the problem of "what to do about _optional_ randomization in environments where no entropy is available", we should revert it: the flow chart of which code path is taken for "legacy" scalar multiplication is already complicated enough as it is, adding even more special cases as a workaround for an unrelated "lack of entropy" problem makes things just worse!

romen:
@paulidale can you suggest a way in a Linux x86-64 machine to "disable" entropy so I can properly test the special case code?

romen:
About backporting to 1.1.1, this is the extent of the changes needed for backporting:

```diff
--- master.patch        2020-04-06 13:09:21.409228209 +0200
+++ 111.patch   2020-04-06 13:08:56.753228934 +0200
@@ -1,14 +1,16 @@
-commit 4a00e1b82ebf737aaec67ccc4f37ac9ec27a40b0
+commit b0d5544bbb0c098212903cbf49c41da664d419bd
 Author: Billy Brumley <bbrumley@gmail.com>
 Date:   Wed Apr 1 21:15:58 2020 +0300
 
     [crypto/ec] blind coordinates in ec_wNAF_mul for robustness
+    
+    (cherry picked from commit 4a00e1b82ebf737aaec67ccc4f37ac9ec27a40b0)
 
 diff --git a/crypto/ec/ec_mult.c b/crypto/ec/ec_mult.c
-index 2d3fc50acf..c66276efce 100644
+index d9b18b82de..d2e4773270 100644
 --- a/crypto/ec/ec_mult.c
 +++ b/crypto/ec/ec_mult.c
-@@ -746,6 +746,20 @@ int ec_wNAF_mul(const EC_GROUP *group, EC_POINT *r, const BIGNUM *scalar,
+@@ -740,6 +740,20 @@ int ec_wNAF_mul(const EC_GROUP *group, EC_POINT *r, const BIGNUM *scalar,
                      if (r_is_at_infinity) {
                          if (!EC_POINT_copy(r, val_sub[i][digit >> 1]))
                              goto err;
@@ -30,10 +32,10 @@
                      } else {
                          if (!EC_POINT_add
 diff --git a/crypto/ec/ecp_smpl.c b/crypto/ec/ecp_smpl.c
-index 9eaa887f21..0c25816c72 100644
+index cb9be38fc1..6903db58ff 100644
 --- a/crypto/ec/ecp_smpl.c
 +++ b/crypto/ec/ecp_smpl.c
-@@ -1442,36 +1442,38 @@ int ec_GFp_simple_blind_coordinates(const EC_GROUP *group, EC_POINT *p,
+@@ -1432,36 +1432,38 @@ int ec_GFp_simple_blind_coordinates(const EC_GROUP *group, EC_POINT *p,
      temp = BN_CTX_get(ctx);
      if (temp == NULL) {
          ECerr(EC_F_EC_GFP_SIMPLE_BLIND_COORDINATES, ERR_R_MALLOC_FAILURE);
@@ -48,11 +50,11 @@
 +     * code to continue smoothly and not clobber the error stack.
 +     */
      do {
--        if (!BN_priv_rand_range_ex(lambda, group->field, ctx)) {
+-        if (!BN_priv_rand_range(lambda, group->field)) {
 -            ECerr(EC_F_EC_GFP_SIMPLE_BLIND_COORDINATES, ERR_R_BN_LIB);
 -            goto err;
 +        ERR_set_mark();
-+        ret = BN_priv_rand_range_ex(lambda, group->field, ctx);
++        ret = BN_priv_rand_range(lambda, group->field);
 +        ERR_pop_to_mark();
 +        if (ret == 0) {
 +            ret = 1;
```

Unless someone will explicitly object in the next 24h I will assume everyone agrees these are trivial changes that do not require a separate approval or reapproval.

romen:
I will revert the ready to merge label to restart the grace period counter.

paulidale:
My initial thought is along the lines of:

```sudo rm -f /dev/random
sudo ln /dev/zero /dev/random
```

Failing that there are build time options to disable everything -- I don't remember the specifics off hand.  We're going to need an entropy callback of some kind for the FIPS testing, this might require a dummy RNG which you could hook for such things -- it's not ready at this point.

bernd-edlinger:
Further discussion is affirmative

bernd-edlinger:
@romen should I merge on your behalf, if you are too busy to respond withing 24H
I will go ahead and do it?

romen:
I can do this!

romen:
Merged to `1.1.1` with:
- 43c242b0f6 [crypto/ec] blind coordinates in ec_wNAF_mul for robustness

Merged to `master` with:
- c61ced5ec5 [crypto/ec] blind coordinates in ec_wNAF_mul for robustness


Thanks!

bernd-edlinger:
Thanks Nicola!

PR #11440, opened 2020-03-30
GOST cipher names adjustment

beldmit:
The erroneously introduced names grasshopper-* replaced with
kuznyechik-* according to official algorithm name translation.

Too long symbolic names replaced with human-enterable ones.

##### Checklist


include/openssl/obj_mac.h
```diff
@@ -4430,23 +4430,23 @@
...
 
-#define SN_grasshopper_cfb              "grasshopper-cfb"
-#define NID_grasshopper_cfb             1016
+#define SN_kuznyechik_cfb               "kuznyechik-cfb"
+#define NID_kuznyechik_cfb              1228
 
-#define SN_grasshopper_mac              "grasshopper-mac"
-#define NID_grasshopper_mac             1017
+#define SN_kuznyechik_mac               "kuznyechik-mac"
+#define NID_kuznyechik_mac              1229
```
mattcaswell: Don't all the changes in this file represent an API break? There are suddenly lots of macros that were there before - but no loner are?

beldmit:
I don't think the are API break, but they definitely break the backward compatibility. I did not expect so many changes though

The most important part of the proposed patch for me is renaming grasshopper to "kuznyechik". It seems worth breaking the backward compatibility in 3.0 release. 

The second part is my try to provide human-writable names instead of the formal ones present now. I'd prefer to replace only values for SN/LNs, not their names, but don't know whether it's possible.

beldmit:
@mattcaswell, @openssl/otc - any advices to make this PR acceptable are welcome. I'd prefer to fix these names here and now when we almost don't have implementations referring to them...

mattcaswell:
> I don't think the are API break, but they definitely break the backward compatibility. 

Hmmmm...if it breaks backwards compatibility isn't that an API break??

Can we rename them but somehow provide some compatibility macros for the old names?

beldmit:
At least some of the previous names aren't worth having compatibility macros. BTW, what is the relevant file to place them?

include/openssl/obj_mac.h
mattcaswell: Well since the original defines are in obj_mac.h then I would expect to see the compatibility macros in obj_mac.h....but that is a generated file. You might need to modify objects.pl to insert them.

beldmit: OK, thanks!

beldmit:
Backward compatibility restored, obj_mac.h generation changed.
Irrelevant gost names provided as deprecated in 3.0

t8m:
Travis failure seems relevant.

beldmit:
Fixed. Thanks!
Would you mind to review the PR, @t8m? :)

t8m:
Would it make sense to edit the obj_dat.num so the old deprecated nids are just renamed if there is 1-1 mapping and no new nids are allocated? It should make the gaps in nid_objs to disappear.

beldmit:
Let's try

beldmit:
> That looks much better. Approved if CI passes. Please squash the commits when merging.

Thank you very much, it was a brilliant idea.

Configurations/unix-Makefile.tmpl
```diff
@@ -983,6 +983,7 @@ generate_crypto_objects:
 				crypto/objects/obj_mac.num \
 				crypto/objects/obj_xref.txt \
 				> crypto/objects/obj_xref.h )
+	( cd $(SRCDIR);	cat crypto/objects/obj_compat.h >> include/openssl/obj_mac.h )
```
mattcaswell, line 986: Doesn't this mean that anything not on Unix, won't get this?

beldmit: According to https://github.com/openssl/openssl/blob/7165593ce5a07a6860d4d408ad640ee707172936/Configurations/unix-Makefile.tmpl#L941
all these targets are available only on Unix. Am I wrong?

paulidale: I suspect the filename gives away the secret...

beldmit: This PR opens a way to deprecate erroneous names in a normal way.

beldmit:
@mattcaswell, could you please approve if you are satisfied with the explanation?

Configurations/unix-Makefile.tmpl
levitte: `generate_crypto_objects` is one of the `make update` targets, so really not an issue at all for other platforms.

beldmit:
Merged. Thanks!

PR #11454, opened 2020-04-01
Fix a printf format error in cmp_client.c

bernd-edlinger:
The value is of type uint64 but the format
%ld is not suitable for that, need to use %jd.

### DIFF
--- a/crypto/cmp/cmp_client.c
+++ b/crypto/cmp/cmp_client.c
@@ -284,7 +284,7 @@ static int poll_for_response(OSSL_CMP_CTX *ctx, int sleep, int rid,
             if (check_after < 0 || (uint64_t)check_after
                 > (sleep ? ULONG_MAX / 1000 : INT_MAX)) {
                 CMPerr(0, CMP_R_CHECKAFTER_OUT_OF_RANGE);
-                if (BIO_snprintf(str, OSSL_CMP_PKISI_BUFLEN, "value = %ld",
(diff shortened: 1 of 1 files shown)


bernd-edlinger:
Luckily I did not review this code. :smiley:

crypto/cmp/cmp_client.c
```diff
@@ -284,7 +284,7 @@ static int poll_for_response(OSSL_CMP_CTX *ctx, int sleep, int rid,
             if (check_after < 0 || (uint64_t)check_after
                 > (sleep ? ULONG_MAX / 1000 : INT_MAX)) {
                 CMPerr(0, CMP_R_CHECKAFTER_OUT_OF_RANGE);
-                if (BIO_snprintf(str, OSSL_CMP_PKISI_BUFLEN, "value = %ld",
+                if (BIO_snprintf(str, OSSL_CMP_PKISI_BUFLEN, "value = %jd",
```
t8m, line 287: For clarity I think there should be cast `(intmax_t)check_after` although this just an alias.

bernd-edlinger: Yes, probably, I wonder how many other places need that.
OTOH, I use %ld mostly in my own code, because it is more portable,
and cast the value to (long) and live with the possible data loss.

bernd-edlinger:
merged to master: b5f7aa5
Thanks!

PR #11458, opened 2020-04-01
sslapitest: only compile test when it will be used

kaduk:
The test_ccs_change_cipher() test routine is used only when TLS 1.2
is enabled; to fix the strict-warnings build we should not try to
compile it when TLS 1.2 is disabled, either.

### DIFF
--- a/test/sslapitest.c
+++ b/test/sslapitest.c
@@ -649,7 +649,6 @@ end:
 
     return testresult;
 }
-#endif
 
 /*
  * Very focused test to exercise a single case in the server-side state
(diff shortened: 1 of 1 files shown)


kaduk:
(Not sure if this is 'urgent'; it should fix a run-checker build.)

t8m:
@kaduk Can you please merge?

kaduk:
@t8m thanks for the reminder; I think I forgot that this was my submission by the time the openssl-machine reminder came in :(

kaduk:
Merged to master and 1.1.1; closing.

PR #11461, opened 2020-04-02
Expand the XTS documentation

mattcaswell:
Explain that XTS does not support streaming, and that the IV value is the
tweak.


mattcaswell:
Pushed. Thanks.

PR #11462, opened 2020-04-02
Enable WPACKET to write to the end of the buffer first

mattcaswell:
This adds the ability to WPACKET to write to the end of the buffer first and fill the buffer backwards. It also adds the ability to support DER style length encoding.

### DIFF
--- a/crypto/packet.c
+++ b/crypto/packet.c
@@ -65,8 +65,11 @@ int WPACKET_reserve_bytes(WPACKET *pkt, size_t len, unsigned char **allocbytes)
         if (BUF_MEM_grow(pkt->buf, newlen) == 0)
             return 0;
     }
-    if (allocbytes != NULL)
+    if (allocbytes != NULL) {
         *allocbytes = WPACKET_get_curr(pkt);
+        if (pkt->endfirst)
+            *allocbytes -= len;
+    }
 
     return 1;
 }
@@ -74,6 +77,9 @@ int WPACKET_reserve_bytes(WPACKET *pkt, size_t len, unsigned char **allocbytes)
 int WPACKET_sub_reserve_bytes__(WPACKET *pkt, size_t len,
                                 unsigned char **allocbytes, size_t lenbytes)
 {
+    if (pkt->endfirst && lenbytes > 0)
+        return 0;
+
     if (!WPACKET_reserve_bytes(pkt, lenbytes + len, allocbytes))
         return 0;
 
@@ -131,8 +137,23 @@ int WPACKET_init_static_len(WPACKET *pkt, unsigned char *buf, size_t len,
     pkt->staticbuf = buf;
     pkt->buf = NULL;
     pkt->maxsize = (max < len) ? max : len;
+    pkt->endfirst = 0;
 
     return wpacket_intern_init_len(pkt, lenbytes);
+}
+
+int WPACKET_init_der(WPACKET *pkt, unsigned char *buf, size_t len)
+{
+    /* Internal API, so should not fail */
+    if (!ossl_assert(buf != NULL && len > 0))
+        return 0;
+
+    pkt->staticbuf = buf;
+    pkt->buf = NULL;
+    pkt->maxsize = len;
+    pkt->endfirst = 1;
+
+    return wpacket_intern_init_len(pkt, 0);
 }
 
 int WPACKET_init_len(WPACKET *pkt, BUF_MEM *buf, size_t lenbytes)
@@ -144,6 +165,7 @@ int WPACKET_init_len(WPACKET *pkt, BUF_MEM *buf, size_t lenbytes)
     pkt->staticbuf = NULL;
     pkt->buf = buf;
     pkt->maxsize = maxmaxsize(lenbytes);
+    pkt->endfirst = 0;
 
     return wpacket_intern_init_len(pkt, lenbytes);
 }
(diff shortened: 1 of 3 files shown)


test/wpackettest.c
```diff
@@ -349,6 +353,43 @@ static int test_WPACKET_memcpy(void)
...
+               /* Test writing data in a length prefixed sub-packet */
+            || !TEST_true(WPACKET_start_sub_packet(&pkt))
+            || !TEST_true(WPACKET_memcpy(&pkt, testdata, sizeof(testdata)))
+            || !TEST_true(WPACKET_close(&pkt))
+            || !TEST_true(WPACKET_put_bytes_u8(&pkt, 0xfc))
+            || !TEST_true(WPACKET_finish(&pkt))
+            || !TEST_true(WPACKET_get_total_written(&pkt, &written))
+            || !TEST_mem_eq(WPACKET_get_curr(&pkt), written, simpleder,
+                            sizeof(simpleder)))
+        return cleanup(&pkt);
```
levitte, line 376: Would you mind writing in a comment what ASN.1 structure you're writing here?  I do not understand what you're doing here from a structural point of view, or how to draw actual examples from this.

levitte:
> WPACKET could really do with something in doc/internal/man3...

Frankly, that's putting it lightly.  It *must* be documented.

crypto/packet.c
```diff
@@ -232,6 +255,19 @@ static int wpacket_intern_close(WPACKET *pkt, WPACKET_SUB *sub, int doclose)
...
+
+        while ((tmplen = tmplen >> 8) > 0)
+            numlenbytes++;
+        if (!WPACKET_put_bytes__(pkt, packlen, numlenbytes))
+            return 0;
+        if (packlen > 0x7f) {
+            numlenbytes |= 0x80;
+            if (!WPACKET_put_bytes_u8(pkt, numlenbytes))
+                return 0;
+        }
```
levitte, line 270: So, uhmmmm, do I understand this right, that I need to make a sub-packet for *every* *damn* *item* in an ASN.1 structure?  Seriously?

mattcaswell: I expect there to be DER helper methods for the different DER types. This is just the underlying WPACKET support.

crypto/packet.c
```diff
@@ -65,15 +65,21 @@ int WPACKET_reserve_bytes(WPACKET *pkt, size_t len, unsigned char **allocbytes)
         if (BUF_MEM_grow(pkt->buf, newlen) == 0)
             return 0;
     }
-    if (allocbytes != NULL)
+    if (allocbytes != NULL) {
         *allocbytes = WPACKET_get_curr(pkt);
+        if (pkt->endfirst)
```
levitte, line 70: Should be `if (pkt->endfirst && *allocbytes != NULL)`

levitte:
I haven't looked too closely, so it's possible that only the DER support stuff needs testing with no buffer (see https://github.com/openssl/openssl/pull/11462#discussion_r402534397), but it's also possible that you need to test that more thoroughly in a more general sense.  I suggest you take an extra look.

crypto/packet.c
mattcaswell: Yes - I actually had this fix in my other temporary branch that I shared with you - but it hadn't made it back into this PR

mattcaswell:
Fixup pushed addressing the feedback above. It also adds a `WPACKET_init_null_der` function - which does the same as `WPACKET_init_null` but calculates lengths properly for DER based length encoding.

The NULL packet encoding needs a test, so I've put this into WIP until I've written that - hopefully tomrrow.

crypto/packet.c
```diff
@@ -232,6 +265,19 @@ static int wpacket_intern_close(WPACKET *pkt, WPACKET_SUB *sub, int doclose)
...
+        if (!WPACKET_put_bytes__(pkt, packlen, numlenbytes))
+            return 0;
+        if (packlen > 0x7f) {
+            numlenbytes |= 0x80;
+            if (!WPACKET_put_bytes_u8(pkt, numlenbytes))
+                return 0;
+        }
     }
 
     if (doclose) {
```
levitte, line 283: Ok, `packlen` is being written to the buffer, fine, but then what?  At least in DER terms, it needs to be accumulated in the parent packet, otherwise we end up with too short lengths all over the place.  Remember that the length of a SEQUENCE, SET or CHOICE is the sum of the lengths of all its elements, including their tags and lengths...

levitte: Never mind this, I was looking at the wrong byte...

levitte:
I've used this merged with the latest #11450, and it now works flawlessly.

mattcaswell:
Now updated with some additional test code to cover the `WPACKET_init_null_der` function. Taken out of WIP. Please take a look.

levitte:
Travis failure is the usual timeout...  why is this still happening?

richsalz:
> Travis failure is the usual timeout... why is this still happening?

Because *one hour* is still not enough time.

This build should be removed.

levitte:
> This build should be removed.

PR welcome :wink:

richsalz:
I would do remove the build if I could, but I don't know enough about travis configuration matrices, sorry.

mattcaswell:
> This build should be removed.

I suggest instead a stripped back build that runs a subset of the tests.

mattcaswell:
Pushed. Thanks.

Issue #11456, opened 2020-04-01
upgrading from 1.1.1.e and 1.1.1.f causes openvpn connections using ssl to fail.


baelish:
Apologies in advance for what is probably a poor report as it's affected me only indirectly through openvpn.

```
moiraine$>openvpn --version
OpenVPN 2.4.8 [git:makepkg/3976acda9bf10b5e+] x86_64-pc-linux-gnu [SSL (OpenSSL)] [LZO] [LZ4] [EPOLL] [PKCS11] [MH/PKTINFO] [AEAD] built on Jan  3 2020
library versions: OpenSSL 1.1.1e  17 Mar 2020, LZO 2.10
Originally developed by James Yonan
Copyright (C) 2002-2018 OpenVPN Inc <sales@openvpn.net>
Compile time defines: enable_async_push=no enable_comp_stub=no enable_crypto=yes enable_crypto_ofb_cfb=yes enable_debug=yes enable_def_auth=yes enable_dlopen=unknown enable_dlopen_self=unknown enable_dlopen_self_static=unknown enable_fast_install=yes enable_fragment=yes enable_iproute2=yes enable_libtool_lock=yes enable_lz4=yes enable_lzo=yes enable_management=yes enable_multihome=yes enable_pam_dlopen=no enable_pedantic=no enable_pf=yes enable_pkcs11=yes enable_plugin_auth_pam=yes enable_plugin_down_root=yes enable_plugins=yes enable_port_share=yes enable_selinux=no enable_server=yes enable_shared=yes enable_shared_with_static_runtimes=no enable_small=no enable_static=yes enable_strict=no enable_strict_options=no enable_systemd=yes enable_werror=no enable_win32_dll=yes enable_x509_alt_username=yes with_aix_soname=aix with_crypto_library=openssl with_gnu_ld=yes with_mem_check=no with_sysroot=no
```
^ Works

```
(1/1) Arming ConditionNeedsUpdate...

~/Downloads
moiraine$>openvpn --version
OpenVPN 2.4.8 [git:makepkg/3976acda9bf10b5e+] x86_64-pc-linux-gnu [SSL (OpenSSL)] [LZO] [LZ4] [EPOLL] [PKCS11] [MH/PKTINFO] [AEAD] built on Jan  3 2020
library versions: OpenSSL 1.1.1f  31 Mar 2020, LZO 2.10
Originally developed by James Yonan
Copyright (C) 2002-2018 OpenVPN Inc <sales@openvpn.net>
[...]

PR #11463, opened 2020-04-02
Allow certificates with Basic Constraints CA:false, pathlen:0

t8m:
Do not mark such certificates with EXFLAG_INVALID although they
violate the RFC 5280, they are syntactically correct and
openssl itself can produce such certificates without any errors
with command such as:

openssl x509 -req -signkey private.pem -in csr.pem -out cert.pem \
  -extfile <(echo "basicConstraints=CA:FALSE,pathlen:0")

With the commit ba4356ae4002a04e28642da60c551877eea804f7 the
EXFLAG_INVALID causes openssl to not consider such certificate
even as leaf self-signed certificate which is breaking existing
installations.

Fixes: 11456


t8m:
This is a proposal. There might be other ways to handle this problem.
Alternatively we might apply this change only to 1.1.1 branch and keep master as more strict.

bernd-edlinger:
Nice commit message.
Could you make X509_V_FLAG_X509_STRICT check that error condition instead ??

t8m:
What about this approach - I've just added a check into the check_chain_extensions() in addition to the original patch.

crypto/x509/x509_vfy.c
```diff
@@ -509,6 +509,12 @@ static int check_chain_extensions(X509_STORE_CTX *ctx)
                 ret = 1;
             break;
         }
+        if ((x->ex_flags & EXFLAG_CA) == 0
+            && x->ex_pathlen != -1
+            && ctx->param->flags & X509_V_FLAG_X509_STRICT) {
```
bernd-edlinger, line 514: please parentheses around & here

bernd-edlinger:
Yes, I like it this way, two commit also okay,
commit message also okay.

crypto/x509/x509_vfy.c
```diff
@@ -509,6 +509,12 @@ static int check_chain_extensions(X509_STORE_CTX *ctx)
...
+            ctx->error = X509_V_ERR_INVALID_EXTENSION;
```
bernd-edlinger, line 515: do we need a verify callback here?

bernd-edlinger: (I dont know in the moment) the answer is open.

bernd-edlinger:
This needs a test case.

t8m:
Thank you for the review.

> This needs a test case.

Yes, I'll add one tomorrow. In the mean time it would be nice if @vdukhovni and/or @mattcaswell could also provide input on whether this approach is correct.

kaduk:
I have very mixed feelings about using "openssl lets you generate such a thing" to justify "openssl should allow this thing as valid".  Just because I can't come up with a specific reason it's insecure to do so isn't a guarantee that it's safe.

bernd-edlinger:
ack.

bernd-edlinger:
Really, we should make an attempt to generate that insanity fail,
but it will break something as well.

crypto/x509/x509_vfy.c
vdukhovni: Perhaps, though applications would be wise to not ignore such errors, especially non-leaf certificates, since we can't provide much context here... For now, we can leave this the way it is (no callback).

vdukhovni:
The rationale isn't that OpenSSL can generate it, but rather that it is actually observed to break deployments, and is a behaviour change from 1.1.1d in 1.1.1[ef].

@t8m can you provide some more detail on the context in which the problems were observed, what type of servers were found to have the problem certificates?

kroeckx:
Did you read #11456?

vdukhovni:
> Did you read #11456?

Thanks for drawing my attention to that detail. I did now, that's useful background.

t8m:
Test case added. @vdukhovni @bernd-edlinger  please re-review.

t8m:
Merged to master.
Merged to 1.1.1 (after resolving trivial conflict in test/recipes/25-test_verify.t).

PR #11468, opened 2020-04-03
Travis matrix adjustments to avoid duplication and frequent timeouts

t8m:
The intention is to re-enable some clang build on linux and replace the constantly timeouting osx build with it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
-dist: xenial
-sudo: required
+dist: bionic
 
 osx_image: xcode9.3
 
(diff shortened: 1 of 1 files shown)


t8m:
As you can see here https://travis-ci.com/github/t8m/openssl?utm_medium=notification&utm_source=email the only build not passing within the extended tests is the one with external krb5 test. That is no regression.

kaduk:
> As you can see here https://travis-ci.com/github/t8m/openssl?utm_medium=notification&utm_source=email the only build not passing within the extended tests is the one with external krb5 test. That is no regression.

I thought we had the krb5 test working fairly recently, to while it may not be a local regression I am still interested to know how it came to be.  I will go kick off a local build with the krb5 tests, at least...

t8m:
> > As you can see here https://travis-ci.com/github/t8m/openssl?utm_medium=notification&utm_source=email the only build not passing within the extended tests is the one with external krb5 test. That is no regression.
> 
> I thought we had the krb5 test working fairly recently, to while it may not be a local regression I am still interested to know how it came to be. I will go kick off a local build with the krb5 tests, at least...

The last commit where krb5 test was passing is: f33ca114c1f8a1ea07948fec6e175806d69274ef
The first commit where it is failing is: 8efc4a9c656b8b3eabeac64c54efdeb07dd42f8e

However there were a few (rather innocently looking) commits in between where Travis did not run.

t8m:
Actually the commits are not that innocent I suppose - most probably the krb5 build is broken due to some mix-up of system and local openssl after the change from shlib_wrap.sh to wrap.pl.

t8m:
@kaduk I've reported is as #11492

kaduk:
Thanks!  That does  feel like a likely culprit, given what things look like in my local testing.
(And I agree that it's a separate issue that should not hold up this PR.)

t8m:
Merged to master as fbc6efb

PR #11472, opened 2020-04-05
Ensure ECDSA_size always returns >= 0

### LINKED ISSUE #10484: ECDSA_size can return negative value on error
https://www.openssl.org/docs/man1.1.1/man3/ECDSA_size.html states that "ECDSA_size() returns the maximum length signature or 0 on error", but in fact, it can also return a negative value because it calls: 
 ret = ASN1_object_size(1, i, V_ASN1_SEQUENCE);
which can return -1 under two error conditions.

wbrawner:
Fixes #10484

Signed-off-by: William Brawner <me@wbrawner.com>


mattcaswell:
Needs 2nd review.

t8m:
Merged to 1.1.1 as f9f2e60. Thank you.

PR #6279, opened 2018-05-17
fixing integer overflow in ASN1_STRING_set

jkrshnmenon:
ASN1_STRING data types are not supposed to contain strings longer than INT_MAX bytes.
Simply returning 0 to the caller might suffice.

PR #11473, opened 2020-04-05
Integer overflow in ASN1_STRING_set.

paulidale:
Addressing a potential integer overflow condition.

Alternative implementation for #6279.

### DIFF
--- a/crypto/asn1/asn1_lib.c
+++ b/crypto/asn1/asn1_lib.c
@@ -275,18 +275,29 @@ ASN1_STRING *ASN1_STRING_dup(const ASN1_STRING *str)
     return ret;
 }
 
-int ASN1_STRING_set(ASN1_STRING *str, const void *_data, int len)
+int ASN1_STRING_set(ASN1_STRING *str, const void *_data, int len_in)
 {
     unsigned char *c;
     const char *data = _data;
+    size_t len;
 
-    if (len < 0) {
+    if (len_in < 0) {
         if (data == NULL)
             return 0;
-        else
-            len = strlen(data);
+        len = strlen(data);
+    } else {
+        len = (size_t)len_in;
     }
-    if ((str->length <= len) || (str->data == NULL)) {
+    /*
+     * Veryif that the length fits within an integer for assignment to
+     * str->length below.  The additional 1 is subtracted to allow for the
+     * \0 terminator even though this isn't strictly necessary.
+     */
+    if (len > INT_MAX - 1) {
+        ASN1err(0, ASN1_R_TOO_LARGE);
+        return 0;
+    }
+    if ((size_t)str->length <= len || str->data == NULL) {
         c = str->data;
         str->data = OPENSSL_realloc(c, len + 1);
         if (str->data == NULL) {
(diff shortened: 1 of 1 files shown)


crypto/asn1/asn1_lib.c
```diff
@@ -275,18 +275,24 @@ ASN1_STRING *ASN1_STRING_dup(const ASN1_STRING *str)
...
+    if (len_in < 0) {
         if (data == NULL)
             return 0;
-        else
-            len = strlen(data);
+        len = strlen(data);
+    } else {
+        len = (size_t)len_in;
+    }
+    if (len > SIZE_MAX - 1) {
```
t8m, line 291: This should be INT_MAX-1 actually.

paulidale: Why exactly?  The `memcpy(2)` allows a `size_t` and passing a negative length value means calculate the length and hasn't limited it in the past.

BTW: I'm fine reducing this to `INT_MAX - 1` if it is desirable.

t8m: Because then on line 304 you're assigning the value to int value str->length.

paulidale: Good point, I'll fix.

paulidale: Fixed now :)

crypto/asn1/asn1_lib.c
```diff
@@ -275,18 +275,24 @@ ASN1_STRING *ASN1_STRING_dup(const ASN1_STRING *str)
...
+    if (len_in < 0) {
         if (data == NULL)
             return 0;
-        else
-            len = strlen(data);
+        len = strlen(data);
+    } else {
+        len = (size_t)len_in;
+    }
+    if (len > INT_MAX - 1) {
```
levitte, line 291: `- 1` is to leave space for an ending NUL byte, right?  You might want to add a quick comment to that effect

t8m: Actually it would not necessarily have to be there because the length does not cover the extra NUL byte however I think it is reasonable to have it there "just-in-case" someone uses the length value with int variable and tries to increment it.

paulidale: I figured out that the `-1` wasn't vital but allowing the user to `strlen()+1` and store into an integer seemed worthwhile as a surprise avoider.

I've added a comment explaining the situation.

crypto/asn1/asn1_lib.c
```diff
@@ -275,18 +275,29 @@ ASN1_STRING *ASN1_STRING_dup(const ASN1_STRING *str)
...
         if (data == NULL)
             return 0;
-        else
-            len = strlen(data);
+        len = strlen(data);
+    } else {
+        len = (size_t)len_in;
+    }
+    /*
+     * Veryif that the length fits within an integer for assignment to
```
t8m, line 292: typo: Veryif

t8m:
OK now.

paulidale:
Merged to 1.1.1 and master.

z00z004242:
Hi everyone. 
This issue has been fixed in latest 1.1.1g release. Why this issue didn't receive any CVE ?

Regards, 
W

PR #11474, opened 2020-04-06
params: avoid a core dump with a null pointer and a get string call

paulidale:
Rather than seg faulting, set the param size and return a failure.

### DIFF
--- a/crypto/params.c
+++ b/crypto/params.c
@@ -778,6 +778,8 @@ static int get_string_internal(const OSSL_PARAM *p, void **val, size_t max_len,
 
     if (sz == 0)
         return 1;
+    if (p->data == NULL)
+        return 0;
 
     if (*val == NULL) {
         char *const q = OPENSSL_malloc(sz);


paulidale:
Is @slontis hit this while coding something a reasonable excuse? :)

I'll amend the commit message with a note about avoiding a NULL dereference.

t8m:
I might sound like Andy :D but in case this is a 'cannot happen' situation sometimes the segfault on NULL poiner dereference (with a proper backtrace) can provide more information to the developer that made a mistake than error return that can be silently ignored and causing weird behavior much later. 
So I wanted to see the rationale in form of "yes, we need this because this could be a legitimate situation that we can hit and so we need to handle it gracefully".

paulidale:
This is definitely a _can happen_ situation, as I discovered earlier today.  Returning the desired size is consistent with what the other OSSL_PARAM helper functions do with NULL.

t8m:
OK, thank you for the updated commit message. Still approved.

levitte:
> I might sound like Andy :D but in case this is a 'cannot happen' situation sometimes the segfault on NULL poiner dereference...

I was thinking the same, but figured that these are public functions, and the NULL pointer may come all the way from the user, so...

paulidale:
Travis is a time out and not relevant.

paulidale:
Merged to master, thanks for the comments.

PR #11480, opened 2020-04-07
params: add a warning about the PTR types.

paulidale:
The warning is to deter the unsure -- if in doubt the PTR type is almost certainly
**not** what you should be using.

### DIFF
--- a/doc/man3/OSSL_PARAM.pod
+++ b/doc/man3/OSSL_PARAM.pod
@@ -159,6 +159,9 @@ The parameter data is a pointer to a printable string.
 The difference between this and B<OSSL_PARAM_UTF8_STRING> is that I<data>
 doesn't point directly at the data, but to a pointer that points to the data.
 
+If there is any uncertainty about which to use, B<OSSL_PARAM_UTF8_STRING> is
+almost certainly the correct choice.
+
 This is used to indicate that constant data is or will be passed,
 and there is therefore no need to copy the data that is passed, just
 the pointer to it.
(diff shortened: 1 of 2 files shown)


t8m:
The man page text seems to be OK. The header comment text seems to me to be patronizing too much. I'd use something like `If you are not completely sure you most likely want to use the ...other type`.

paulidale:
I've changed the text.

I'm not sure that this can be blatant enough -- I think that **_everyone_** (except perhaps @levitte) has tried to use PTR types incorrectly.

If you think you should use them, you shouldn't :)

t8m:
Then why do we even provide such thing? :smile:

t8m:
Should the types be named OSSL_PARAM_UTF8_FOR_LEVITTE_ONLY ? :smile:

paulidale:
> Should the types be named OSSL_PARAM_UTF8_FOR_LEVITTE_ONLY ? 😄

Yes.

levitte:
I'd like to hear the background story, i.e how did someone shoot themselves in the foot this time?

paulidale:
It's more an accumulation of feet shot than a single incident.

I've been meaning to add a note like this for a while.

levitte:
> OSSL_PARAM_UTF8_FOR_LEVITTE_ONLY

while I'm kinda flattered (... ... ... I think :pensive:), my self inflation isn't grand enough.  So no...

paulidale:
Merged to master.  Thanks for the comments.

PR #11481, opened 2020-04-07
Minor SSL_CTX_sessions DOC improvement

EasySec:
use LHASH_OF(TYPE) macro to make the example consistent with the declaration in ssl.h

### DIFF
--- a/doc/man3/SSL_CTX_sessions.pod
+++ b/doc/man3/SSL_CTX_sessions.pod
@@ -8,7 +8,7 @@ SSL_CTX_sessions - access internal session cache
 
  #include <openssl/ssl.h>
 
- struct lhash_st *SSL_CTX_sessions(SSL_CTX *ctx);
+ LHASH_OF(SSL_SESSION) *SSL_CTX_sessions(SSL_CTX *ctx);
 
 =head1 DESCRIPTION
 
--- a/util/find-doc-nits
+++ b/util/find-doc-nits
@@ -351,6 +351,7 @@ sub name_synopsis {
 
         my $sym;
         my $is_prototype = 1;
+        $line =~ s/LHASH_OF\([^)]+\)/int/g;
         $line =~ s/STACK_OF\([^)]+\)/int/g;
         $line =~ s/SPARSE_ARRAY_OF\([^)]+\)/int/g;
         $line =~ s/__declspec\([^)]+\)//;


kaduk:
Hmm, the travis failures are doc-nits complaining.  But I don't remember what we've done to appease it in the past for similar issues...

richsalz:
Looks like find-doc-nits might have a bug to complain about the LHASH_OF.
The other errors are because someone merged a broken file.

mattcaswell:
Yes, travis failure is relevant unfortunately. As @richsalz points out there seems to be a bug in doc-nits  that this change is hitting.

t8m:
I'd recommend redoing the PR from scratch by cherry-picking. This rebase went certainly seriously wrong.

EasySec:
> 
> 
> I'd recommend redoing the PR from scratch by cherry-picking. This rebase went certainly seriously wrong.

I think so, sorry for the mess.

t8m:
Unfortunately the find-doc-nits deficiency would have to be fixed first to merge this.

t8m:
One option would be to expand the macro in the manpage. I.e. to use 'struct lhash_st_SSL_SESSION *` instead.

EasySec:
> 
> 
> One option would be to expand the macro in the manpage. I.e. to use 'struct lhash_st_SSL_SESSION *` instead.

IMHO, it would be better to promote the use of the macro

t8m:
Sure, but that requires fixing the find-doc-nits first. Are you able to do that? If not you can open an issue and this PR will be blocked on that issue. Unfortunately I cannot promise this will get fixed any time soon.

richsalz:
This seems to fix it:
```
diff --git a/util/find-doc-nits b/util/find-doc-nits
index 60a2b8c6a0..a5ea78706d 100755
--- a/util/find-doc-nits
+++ b/util/find-doc-nits
@@ -351,6 +351,7 @@ sub name_synopsis {
 
         my $sym;
         my $is_prototype = 1;
+        $line =~ s/LHASH_OF\([^)]+\)/int/g;
         $line =~ s/STACK_OF\([^)]+\)/int/g;
         $line =~ s/SPARSE_ARRAY_OF\([^)]+\)/int/g;
         $line =~ s/__declspec\([^)]+\)//;
; 
```

EasySec:
> 
> 
> This seems to fix it:
> 
And super-Rich save us again. Thanks ! (I wouldn't have been able to fix anything)
Can I assume there is nothing more to do now with that PR except waiting ?
(and there is not urgency ;-)

t8m:
Well, the fix needs to be included - you could just commit it here as a separate commit with `--author="Rich Salz <rsalz@akamai.com>"`

richsalz:
I don't need credit.  If you must add the one-line fix to your  commit and say "find-doc-nits fix courtesy Rich Salz" but it's really not necessary.

EasySec:
> 
> 
> Well, the fix needs to be included - you could just commit it here as a separate commit with `--author="Rich Salz <rsalz@akamai.com>"`

Well, I did not even realize before that you were speaking about a perl script in the source tree...
Will do it.

EasySec:
Ooops, looks not good. Did I miss something ?

t8m:
The check_update failure is unrelated.

t8m (approved):

kaduk (approved):
The doc fix is fine.
IIRC, the API itself is a bit dicey, since you get access to the structure but not the lock that protects updates to it.  (It is really bad if you start doing things like #2397, but I think still problematic even as-is.)

paulidale (approved):

paulidale:
Merged to master.  Thanks for the improvement.

PR #11482, opened 2020-04-07
TLSEXT_SIGALG_gostr34102012 0x0840 and 0x0841 support

NMorozxov:
For GOST2012-GOST8912-GOST8912 IANA introduce signature Signature Algorithm parametrs
https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-16

##### Checklist

### DIFF
--- a/ssl/ssl_local.h
+++ b/ssl/ssl_local.h
@@ -2127,6 +2127,8 @@ typedef enum downgrade_en {
 #define TLSEXT_SIGALG_dsa_sha512                                0x0602
 #define TLSEXT_SIGALG_dsa_sha224                                0x0302
 #define TLSEXT_SIGALG_dsa_sha1                                  0x0202
+#define TLSEXT_SIGALG_gostr34102012_256_intrinsic               0x0840
+#define TLSEXT_SIGALG_gostr34102012_512_intrinsic               0x0841
 #define TLSEXT_SIGALG_gostr34102012_256_gostr34112012_256       0xeeee
 #define TLSEXT_SIGALG_gostr34102012_512_gostr34112012_512       0xefef
 #define TLSEXT_SIGALG_gostr34102001_gostr3411                   0xeded
--- a/ssl/t1_trce.c
+++ b/ssl/t1_trce.c
@@ -569,6 +569,8 @@ static const ssl_trace_tbl ssl_sigalg_tbl[] = {
     {TLSEXT_SIGALG_dsa_sha512, "dsa_sha512"},
     {TLSEXT_SIGALG_dsa_sha224, "dsa_sha224"},
     {TLSEXT_SIGALG_dsa_sha1, "dsa_sha1"},
+    {TLSEXT_SIGALG_gostr34102012_256_intrinsic, "gost2012_256"},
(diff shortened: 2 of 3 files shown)


ssl/ssl_local.h
```diff
@@ -2115,6 +2115,8 @@ typedef enum downgrade_en {
 #define TLSEXT_SIGALG_rsa_pss_rsae_sha384                       0x0805
 #define TLSEXT_SIGALG_rsa_pss_rsae_sha512                       0x0806
 #define TLSEXT_SIGALG_rsa_pss_pss_sha256                        0x0809
+#define TLSEXT_SIGALG_gostr34102012_256_intrinsic               0x0840
```
t8m, line 2118: This is really a nit however why are you adding the gostr sigalgs in between rsa_pss ones?

ssl/t1_trce.c
```diff
@@ -557,6 +557,8 @@ static const ssl_trace_tbl ssl_sigalg_tbl[] = {
     {TLSEXT_SIGALG_rsa_pss_rsae_sha384, "rsa_pss_rsae_sha384"},
     {TLSEXT_SIGALG_rsa_pss_rsae_sha512, "rsa_pss_rsae_sha512"},
     {TLSEXT_SIGALG_rsa_pss_pss_sha256, "rsa_pss_pss_sha256"},
+    {TLSEXT_SIGALG_gostr34102012_256_intrinsic, "gost2012_256"},
```
t8m, line 560: The same nit as above. Please put them along the other gost sigalgs.

ssl/ssl_local.h
NMorozxov: sorted by value

ssl/t1_trce.c
NMorozxov: Ok, no problem i will but them along the gost in both places

ssl/ssl_local.h
NMorozxov: ohhh...
this list isn't sorted, i miss

NMorozxov:
@t8m @beldmit fix commited

beldmit:
> Dmitry, please re-approve.

Done.

NMorozxov:
ping @beldmit @t8m

beldmit:
I don't understand why the label was not applied automatically.

mattcaswell:
As per the comment  from openssl-machine above "this PR has been updated in that time the label 'approval: ready to merge' is not being automatically set." - you added a comment after "approval done".

beldmit:
This behavior seems a bit strange - no code changes/change requests... Thanks for the clarification!

beldmit:
Merged. Thanks!

mattcaswell:
> This behavior seems a bit strange - no code changes/change requests... Thanks for the clarification!

The bot doesn't know the difference between an innocuous "Looks good to me" comment from an approval status relevant "This code is rubbish, I think this PR should be closed without merge". So it errs on the side of caution. Any update whatsoever results in the tag not being applied automatically - instead you just get the message above.

PR #11485, opened 2020-04-07
WIP: FIPS-validatable RAND API

richsalz:
This PR is for making it possible to FIPS-validate the RAND "get-bytes" API. This work has been called "fetchable RAND" which is shorter and more evocative, but it also could be seen as covering too much work.  The goal here (as opposed to #10144) is that this PR is doing less but still meeting the requirements.


richsalz:
I do not like the "../../../crypto/rand/rand_local.h" that appears in the source files. It follows what the SipHash provider-based code does. Should I move rand_local.h to internal?  Please advise.

slontis:
> I do not like the "../../../crypto/rand/rand_local.h" that appears in the source files. It follows what the SipHash provider-based code does. Should I move rand_local.h to internal? Please advise.

Maybe just move whatever calls we need exposed to internal? If that turns out to be 1-1 then the local file then disappears.

mattcaswell:
> This PR is for making it possible to FIPS-validate the RAND "get-bytes" API. This work has been called "fetchable RAND" which is shorter and more evocative, but it also could be seen as covering too much work. The goal here (as opposed to #10144) is that this PR is doing less but still meeting the requirements.

It's not clear to me from what you've got here so far what you are attempting to achieve. This just seems to move some files around but I'm not seeing how it addresses the problem of making the RAND "get-bytes" API validatable?

BTW I get link errors when I try to build this.

richsalz:
This is a WIP, and there is one commit that talks about moving files into the provider subtree. It's far from ready.  I am trying to make progress in the PR and get comments so that (unlike my other PR's) this doesn't sit around for months with nobody looking at it.  It's an experiment.

So far, I found one issue, that @slontis mentioned.

What link errors do you get?  I get none. My config flags are:
```
./config -d --strict-warnings --prefix=/usr/local/openssl -Qunused-arguments -fsanitize=leak -fsanitize=address enable-md2 enable-ec_nistp_64_gcc_128
```

paulidale:
Can we agree on requirements just so they are clear:

- a DRBG that can be fetched from a provider for a specified a library context.
- The DRBG and the seeding source must stay within a single provider
- There must be hooks to permit FIPS related testing
- The CRNG test must remain (this uses the entropy DRBG callbacks)
- The public RAND functions must operate as before
- The public DRBG functions should operate as before

I'm sure I've missed a few.

Keeping the `rand_meth_st` public isn't on this list.  It will break some thing but it should have been made opaque with 1.1.

richsalz:
Yes, those are the requirements.  I would like ```rand_method_t``` to work for legacy but it's not a requirement.

richsalz:
I'm moving ```crypto/rand/rand_local.h``` to ```include/internal/rand.h```.  There is already ```include/crypto/rand.h```, so I need to rationalize internal and crypto rand.h

richsalz:
I have talked about this with the FIPS team, and have had multiple email threads. I am closing this since @paulidale started to address this. I wish he had told me.   The start of the design doc, which only includes the rationale, is at https://docs.google.com/document/d/1pW5O4GBVoskdqYY7Bd1VVE-xpJCKaN6lCYYSpOIqd4I/edit?usp=sharing

PR #11488, opened 2020-04-08
Add a null provider which implements no algorithms.

paulidale:
By loading the null provider into the default context, it is possible
to verify that it is not accidentally being used.


paulidale:
Merged to master.  Thanks.

PR #11493, opened 2020-04-08
Enable export_to functions to have access to the libctx

mattcaswell:
The EC export_to function calls EC_POINT_point2buf that can later
generate a random number in some circumstances. Therefore we pass in a
BN_CTX associated with the library context. This means we have to change
the export_to function signature to accept the library context.

This is needed for alpha1.


kaduk:
Travis failure is just the usual timeout, so setting approval:done per the existing approvals

mattcaswell:
Pushed. Thanks.

PR #11494, opened 2020-04-08
Make PrivateKey loading libctx aware

mattcaswell:
During loading of a private key it is possible that the implementation will need to use a cryptographic algorithm Notably in decoding an Ed448 private key we automatically derive the public  key (because it is not encoded along with the private key). To do so we need to use SHAKE256. This means that we have to make the private key loading functions OPENSSL_CTX aware.

We also update libssl to use the new private key loading functions.

This is needed for alpha1


crypto/asn1/ameth_lib.c
```diff
@@ -283,6 +283,18 @@ void EVP_PKEY_asn1_copy(EVP_PKEY_ASN1_METHOD *dst,
...
+
+    dst->set_priv_key = src->set_priv_key;
+    dst->set_pub_key = src->set_pub_key;
+    dst->get_priv_key = src->get_priv_key;
+    dst->get_pub_key = src->get_pub_key;
+
+    dst->dirty_cnt = src->dirty_cnt;
+    dst->export_to = src->export_to;
+
+    dst->priv_decode_with_libctx = src->priv_decode_with_libctx;
```
levitte, line 297: It would be much safer to just `*dst = *(EVP_PKEY_ASN1_METHOD *)src` and call it a day...  well, not entirely, we need to preserve `pkey_id`, `pkey_base_id`, `pkey_flags`, `pem_str`, `info`, but that seems easier than to remember a gazillion function pointers...

crypto/evp/evp_pkey.c
```diff
@@ -18,7 +18,8 @@
 
 /* Extract a private key from a PKCS8 structure */
 
-EVP_PKEY *EVP_PKCS82PKEY(const PKCS8_PRIV_KEY_INFO *p8)
+EVP_PKEY *evp_pkcs82pkey_int(const PKCS8_PRIV_KEY_INFO *p8, OPENSSL_CTX *libctx,
+                             const char *propq)
```
levitte, line 22: So while you're making new public functions out of the others, this one you're making internal...  it seems inconsistent, wanna tell me why?

mattcaswell: It wasn't clear to me whether we wanted to encourage the use of `EVP_PKCS82PKEY` or not. While I wasn't sure I erred on the side of caution thinking it was easier to make things public later than the other way around.

crypto/asn1/ameth_lib.c
mattcaswell: Good idea. Done.

mattcaswell:
Fixup pushed addressing the comment above. I also rebased to resolve a conflict with master. Please look again.

slontis:
travis error is relevant..

include/openssl/pem.h:353:34: error: unknown type name 'FILE'
 EVP_PKEY *PEM_read_PrivateKey_ex(FILE *fp, EVP_PKEY **x, pem_password_cb *cb,

crypto/asn1/ameth_lib.c
```diff
@@ -250,39 +250,20 @@ EVP_PKEY_ASN1_METHOD *EVP_PKEY_asn1_new(int id, int flags,
...
-
-    dst->pkey_check = src->pkey_check;
-
+    int pkey_id = dst->pkey_id;
+    int pkey_base_id = dst->pkey_base_id;
+    unsigned long pkey_flags = dst->pkey_flags;
+    char *pem_str = dst->pem_str;
+    char *info = dst->info;
+
+    *dst = *src;
```
slontis, line 259: The old code was not copying the newer provider related methods.. So this is a good thing.

crypto/ec/ecx_meth.c
```diff
@@ -35,7 +35,8 @@ typedef enum {
 
 /* Setup EVP_PKEY using public, private or generation */
 static int ecx_key_op(EVP_PKEY *pkey, int id, const X509_ALGOR *palg,
-                      const unsigned char *p, int plen, ecx_key_op_t op)
+                      const unsigned char *p, int plen, ecx_key_op_t op,
+                      OPENSSL_CTX *libctx, const char *propq)
```
slontis, line 39: why pass propq if it is not used?

paulidale: It's prudent to keep the libctx, propq pair I suspect.
There are currently some places the property query doesn't reach down to but the libctx does.

t8m: Which makes me to ask - wouldn't it make sense to make the propq part of the libctx instead of having to add two parameters everywhere? @mattcaswell @levitte ?

mattcaswell: No. We already have a global libctx wide propq. The idea is to be able to override it for specific calls.

mattcaswell: > It's prudent to keep the libctx, propq pair I suspect.
There are currently some places the property query doesn't reach down to but the libctx does.

Right. I think in general we should always be passing the pair. We might have to go back and fix a few spots where we haven't done that at some point

mattcaswell:
I've rebased  this and pushed a fixup to (hopefully) resolve the travis issue. Please take another look.

mattcaswell:
Rebased again to fix conflicts with master. Also another fix up to address travis issues.

Ping?

mattcaswell:
Pushed. Thanks.

slontis:
Do you have more PR's that you are about to merge? libcrypto.num is a bit of a problem.

PR #11496, opened 2020-04-08
Enable Ed25519 signing/verifying to use the libctx

mattcaswell:
Ed25519 needs to fetch a digest and so needs to use the correct libctx.

This is needed for alpha1


mattcaswell:
Travis red cross is due to timeout.

mattcaswell:
> LGTM. What's a little bit weird is that propq is never used with non NULL value. Will this change with future 3.0 work?

It *may* do. I thought it better to add it.

paulidale:
Merge conflicts ahead for me :)

mattcaswell:
Pushed. Thanks!

PR #11498, opened 2020-04-08
Fix EVP_DigestSign interface when used with DES CMAC

p-steuer:
DES implementations were missing the dup/copy ctx routines
required by CMAC implementation. A regression test is added.


p-steuer:
needs a rebase due to https://github.com/openssl/openssl/commit/f5056577ba08b3694aab2722eae1c97bf00acc80

p-steuer:
Rebased. please Re-review.

I noticed during rebase that a lot of header files are missing include guards (e.g. all headers in providers/implementations/ciphers/ but also include/crypto/modes.h ..). I wonder if they were just forgotten or if there is a reason for it ?

mattcaswell:
> I noticed during rebase that a lot of header files are missing include guards (e.g. all headers in providers/implementations/ciphers/ but also include/crypto/modes.h ..). I wonder if they were just forgotten or if there is a reason for it ?

I suspect this is just an error - but I've not looked into it.

p-steuer:
merged .

master: https://github.com/openssl/openssl/commit/abfc73f374c2b80e15f6c9d0b3a7437b3e94afa8

PR #11501, opened 2020-04-09
CMS KARI: Temporarly downgrade newly generated EVP_PKEYs to legacy

levitte:
The EVP_PKEY_ASN1_METHOD code used by CMS_RecipientInfo_kari_decrypt()
and cms_RecipientInfo_kari_encrypt() is quite complex and needs more
careful thought to work with provider side keys.  Unfortunately, we
need to get key generation in place, among others for ECC keys, so we
add a temporary hack, similar to what's already done in TLS code, that
downgrades a provider side EVP_PKEY to become EVP_PKEY_ASN1_METHOD /
EVP_PKEY_METHOD based.

-----

This is very similar to what was done for TLS in #11358


levitte:
#11501 is the new dependency

levitte:
Oops, `make update` needed.  Done

crypto/evp/pmeth_lib.c
```diff
@@ -890,7 +890,8 @@ static int legacy_ctrl_to_param(EVP_PKEY_CTX *ctx, int keytype, int optype,
         case EVP_PKEY_CTRL_CMS_DECRYPT:
         case EVP_PKEY_CTRL_CMS_ENCRYPT:
 # endif
-            if (ctx->pmeth->pkey_id != EVP_PKEY_RSA_PSS)
+            /* TODO (3.0) Temporary hack, this should probe */
+            if (!EVP_PKEY_is_a(EVP_PKEY_CTX_get0_pkey(ctx), "RSASSA-PSS"))
```
romen, line 894: Can get0_pkey return NULL? Will EVP_PKEY_is_a handle ti gracefully?

Do we care about a far fetched case in a temporary solution?

levitte: > Do we care about a far fetched case in a temporary solution?

I've seen crashes because `ctx->pmeth == NULL`

levitte: > Can get0_pkey return NULL? Will EVP_PKEY_is_a handle ti gracefully?

Too answer that, you need to look here:

https://github.com/openssl/openssl/blob/fbc6efbbc16fc1410f776431395ea2b2ad691d4c/crypto/evp/pmeth_lib.c#L913-L920

See how it checks that an operation is ongoing?  The basic assumption, as far as I know, is that you can't have an initialized operation without a key.

levitte:
Uhmmm, we do have some CI reactions, but they are completely unrelated to this PR.

t8m:
> Uhmmm, we do have some CI reactions, but they are completely unrelated to this PR.

Apparently the test coverage before #11468 was not particularly good.

levitte:
> Apparently the test coverage before #11468 was not particularly good.

Ah, so you're to blame :wink:
Still unrelated to this PR...  and I'm actually quite surprised that perl issues weren't caught earlier

levitte:
Ah, darn...  OpenSSL::OID uses `List::Util::pairs`, which came with List::Util 1.29, which was included with perl v5.19.3.  In the failing case, the perl version is 5.18.2.

Okie, will fix that in a separate PR

levitte:
OpenSSL::OID issue fixedd, see #11503.

mattcaswell:
Travis failures are strange but do not seem related to this PR

levitte:
Merged

aec8de1a5f0b3f6e6279266b45836d6c9f6878df CMS KARI: Temporarly downgrade newly generated EVP_PKEYs to legacy
cc572c25647f6ab0c026a4a3057dc706e4a4a412 EVP: legacy_ctrl_to_param() to handle provider side keys

PR #11503, opened 2020-04-09
OpenSSL::OID: Don't use List::Util

levitte:
It turns out that the pairwise functions of List::Util came into perl
far later than 5.10.0.  We can't use that under those conditions, so
must revert to a quick internal implementation of the functions we're
after.


levitte:
> Also I think this is urgent.

Agreed.

levitte:
Merged

8519b244bc6c38f265bf9bad80c52bd7c0ff469f OpenSSL::OID: Don't use List::Util

PR #11504, opened 2020-04-09
[URGENT] Add common internal crypto/ modules in liblegacy.a

levitte:
Just as for the FIPS module, there's code in the legacy module that need
this.

### DIFF
--- a/crypto/build.info
+++ b/crypto/build.info
@@ -80,6 +80,7 @@ SOURCE[../libcrypto]=$UTIL_COMMON \
         o_fopen.c getenv.c o_init.c o_fips.c init.c trace.c provider.c \
         $UPLINKSRC
 SOURCE[../providers/libfips.a]=$UTIL_COMMON
+SOURCE[../providers/liblegacy.a]=$UTIL_COMMON
 
 # Implementations are now spread across several libraries, so the defines
 # need to be applied to all affected libraries and modules.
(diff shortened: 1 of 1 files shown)


levitte:
This fixes, I hope, the other build issue that was encountered at the same time as what fixed in #11503

t8m:
Should this be conditional on $disabled{'deprecated-3.0'} ?

slontis:
Thanks @levitte that seemed to fix the no-asm no-deprecated configuration..

t8m:
> Should this be conditional on $disabled{'deprecated-3.0'} ?

To expand on this - without no-deprecated the code that depends on CPUID and other objects will be linked directly from the shared libcrypto so these object files will be in legacy provider unnecessarily.
On the other hand it is a question why the linkage model of legacy provider should differ on with or without no-deprecated.

levitte:
> To expand on this - without no-deprecated the code that depends on CPUID and other objects will be linked directly from the shared libcrypto so these object files will be in legacy provider unnecessarily.

No.  They aren't necessarily exported by libcrypto.  If you wanna dive deeper and make a more precise selection, do feel free.

levitte:
> On the other hand it is a question why the linkage model of legacy provider should differ on with or without no-deprecated.

That's more about what will remain in libcrypto.  Without `no-deprecated`, all the ciphers, digests et al remain in libcrypto and are exported.  With `no-deprecated`, they are *removed* from libcrypto, entirely, and then need to end up in the legacy provider.

t8m:
> > To expand on this - without no-deprecated the code that depends on CPUID and other objects will be linked directly from the shared libcrypto so these object files will be in legacy provider unnecessarily.
> 
> No. They aren't necessarily exported by libcrypto. If you wanna dive deeper and make a more precise selection, do feel free.

I mean currently this is not needed for builds without no-deprecated at all because nothing in legacy.so depends on these things directly, only transitionally via the moved-in crypto implementations with no-deprecated. However I understand things might eventually change in future.

t8m (approved):

slontis (approved):
Assuming the Appveyor tests pass.

levitte:
All CIs happy.  This goes in now

levitte:
Merged.

09fafd06065b4de17ec8b2ae0004002f2a252dc8 Add common internal crypto/ modules in liblegacy.a

PR #11505, opened 2020-04-09
Fix for #7717

### LINKED ISSUE #7717: BIO_do_accept returns closed fd
OpenSSL version: 1.1.1

```C
{
    BIO         *acc, *client;
    THREAD_TYPE tid;
 
    init_OpenSSL(  );
 
    acc = BIO_new_accept(PORT);
    if (!acc)
        int_error("Error creating server socket");
  
    if (BIO_do_accept(acc) <= 0)
        int_error("Error binding server socket");
  
    for (;;)
    {
        if (BIO_do_accept(acc) <= 0)
            int_error("Error accepting connection");
 
        client = BIO_pop(acc);
        THREAD_CREATE(tid, (void *)server_thread, client);
    }
 
    BIO_free(acc);
    return 0;
}
```
In case of 'address already in use' error, the first ```BIO_do_accept``` returns the closed socket rather than -1, which cause the second ```BIO_do_accept``` have a chance to run, producing the following:

```
$ ./server 
** server.c:56 Error accepting connection
139656693395264:error:02006062:system library:bind:Address already in use:crypto/bio/b_sock2.c:159:
139656693395264:error:20093075:BIO routines:BIO_bind:unable to bind socket:crypto/bio/b_sock2.c:160:
139656693395264:error:0200F009:system library:getsockopt:Bad file descriptor:crypto/bio/b_sock2.c:218:
139656693395264:error:2008B086:BIO routines:BIO_listen:getting socktype:crypto/bio/b_sock2.c:219:
```
The fix is one liner:
```C
        case ACPT_S_LISTEN:
            {
                if (!BIO_listen(c->accept_sock,
                                BIO_ADDRINFO_address(c->addr_iter),
                                c->bind_mode)) {
                    BIO_closesocket(c->accept_sock);
                    + ret = -1;
                    goto exit_loop;
                }
            }
```

blumf:
Fixes #7717
CLA: trivial

`BIO_do_accept` was returning incorrect values when unable to bind to a port.

Problem in `acpt_state` assigning a temp value to the return var and not resetting it.

### DIFF
--- a/crypto/bio/bss_acpt.c
+++ b/crypto/bio/bss_acpt.c
@@ -150,7 +150,7 @@ static int acpt_free(BIO *a)
 static int acpt_state(BIO *b, BIO_ACCEPT *c)
 {
     BIO *bio = NULL, *dbio;
-    int s = -1, ret = -1;
+    int s = -1, ret = -1, sock = -1;
 
     for (;;) {
         switch (c->state) {
@@ -222,18 +222,18 @@ static int acpt_state(BIO *b, BIO_ACCEPT *c)
             break;
 
         case ACPT_S_CREATE_SOCKET:
-            ret = BIO_socket(BIO_ADDRINFO_family(c->addr_iter),
-                             BIO_ADDRINFO_socktype(c->addr_iter),
-                             BIO_ADDRINFO_protocol(c->addr_iter), 0);
-            if (ret == (int)INVALID_SOCKET) {
+            sock = BIO_socket(BIO_ADDRINFO_family(c->addr_iter),
+                              BIO_ADDRINFO_socktype(c->addr_iter),
+                              BIO_ADDRINFO_protocol(c->addr_iter), 0);
+            if (sock == (int)INVALID_SOCKET) {
                 ERR_raise_data(ERR_LIB_SYS, get_last_socket_error(),
                                "calling socket(%s, %s)",
                                 c->param_addr, c->param_serv);
                 BIOerr(BIO_F_ACPT_STATE, BIO_R_UNABLE_TO_CREATE_SOCKET);
                 goto exit_loop;
             }
-            c->accept_sock = ret;
-            b->num = ret;
+            c->accept_sock = sock;
+            b->num = sock;
             c->state = ACPT_S_LISTEN;
             break;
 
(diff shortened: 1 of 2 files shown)


t8m:
Please amend the commit message.
1. The subject of the message should be more descriptive such as "BIO_do_accept: correct error return value"
2. The Fixes: #7717 and CLA: trivial must be on separate lines.

I think this is borderline trivial. It would be better if you could submit CLA, but I think this is still acceptable with CLA: trivial.

blumf:
Don't think I can change the subject name now (I think whoever merges it can)

If anyone is desperate, I could submit a CLA, but it really is a small change and I doubt I'll be sending any more patches.

crypto/bio/bss_acpt.c
```diff
@@ -150,7 +150,7 @@ static int acpt_free(BIO *a)
 static int acpt_state(BIO *b, BIO_ACCEPT *c)
 {
     BIO *bio = NULL, *dbio;
-    int s = -1, ret = -1;
+    int s = -1, ret = -1, sock = -1;
```
levitte, line 153: Why add another variable for the socket when you have `s`?

slontis:
> Don't think I can change the subject name now (I think whoever merges it can)


```git commit --amend```

Then edit the commit message (see @t8m steps 1 & 2 above)

```git push -f```

File: crypto/bio/bss_acpt.c (hunk 1 above)
Comment (blumf) [reply crypto/bio/bss_acpt.c]: I think you're right. Amended.

File: crypto/bio/bss_acpt.c [hunk 2]
```diff
@@ -222,19 +222,20 @@ static int acpt_state(BIO *b, BIO_ACCEPT *c)
...
                 BIOerr(BIO_F_ACPT_STATE, BIO_R_UNABLE_TO_CREATE_SOCKET);
                 goto exit_loop;
             }
-            c->accept_sock = ret;
-            b->num = ret;
+            c->accept_sock = s;
+            b->num = s;
             c->state = ACPT_S_LISTEN;
+            s = -1;
             break;
```
kaduk, line 239: Do we need to set `ret` to something other than the `-1` it was initialized to?

t8m: @kaduk No, the next state is ACPT_S_LISTEN and that will take care of it properly. Actually clobbering the -1 in ACPT_S_CREATE_SOCKET was the issue we are fixing here.

t8m:
Setting ready to merge directly because the required approvals were already in 4 days ago.

t8m:
Merged to master as 0437435.
Merged to 1.1.1 after resolving a trivial conflict as 5f62ff4.

PR #11507, opened 2020-04-09
Make X509_verify() libctx aware

mattcaswell:
X509_verify() validates the signature in an X509 cert. We should make sure we use a library ctx and property query string for fetching any algorithms. Therefore we introduce X509_verify_ex().

I've also included a second loosely related commit which gives us an internal version of X509_check_issued() which is library ctx aware. We don't actually *need* this to be public (you can achieve the same thing via other means) - but it makes our internal code a bit neater doing things this way.

This is required for alpha1.

### DIFF
--- a/doc/man3/X509_sign.pod
+++ b/doc/man3/X509_sign.pod
@@ -13,8 +13,8 @@ X509_CRL_sign_ctx, X509_CRL_verify
 
  int X509_sign(X509 *x, EVP_PKEY *pkey, const EVP_MD *md);
  int X509_sign_ctx(X509 *x, EVP_MD_CTX *ctx);
- int X509_verify_ex(X509 *a, EVP_PKEY *pkey, OPENSSL_CTX *libctx, const char *propq);
- int X509_verify(X509 *a, EVP_PKEY *pkey;
+ int X509_verify_ex(X509 *x, EVP_PKEY *pkey, OPENSSL_CTX *libctx, const char *propq);
+ int X509_verify(X509 *x, EVP_PKEY *pkey;
 
  int X509_REQ_sign(X509_REQ *x, EVP_PKEY *pkey, const EVP_MD *md);
  int X509_REQ_sign_ctx(X509_REQ *x, EVP_MD_CTX *ctx);
--- a/doc/man3/X509_sign.pod
+++ b/doc/man3/X509_sign.pod
@@ -86,7 +86,8 @@ L<X509_NAME_get_index_by_NID(3)>,
 L<X509_NAME_print_ex(3)>,
 L<X509_new(3)>,
 L<X509V3_get_d2i(3)>,
-L<X509_verify_cert(3)>
+L<X509_verify_cert(3)>,
+L<OPENSSL_CTX(3)>
 
 =head1 HISTORY
 
--- a/doc/man3/X509_sign.pod
+++ b/doc/man3/X509_sign.pod
@@ -2,9 +2,10 @@
 
 =head1 NAME
 
-X509_sign, X509_sign_ctx, X509_verify, X509_REQ_sign, X509_REQ_sign_ctx,
-X509_REQ_verify, X509_CRL_sign, X509_CRL_sign_ctx, X509_CRL_verify -
-sign or verify certificate, certificate request or CRL signature
+X509_sign, X509_sign_ctx, X509_verify_ex, X509_verify, X509_REQ_sign,
+X509_REQ_sign_ctx, X509_REQ_verify_ex, X509_REQ_verify, X509_CRL_sign,
+X509_CRL_sign_ctx, X509_CRL_verify
+- sign or verify certificate, certificate request or CRL signature
 
 =head1 SYNOPSIS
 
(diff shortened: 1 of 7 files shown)


doc/man3/X509_sign.pod
```diff
@@ -28,11 +32,16 @@ X509_sign() signs certificate B<x> using private key B<pkey> and message
 digest B<md> and sets the signature in B<x>. X509_sign_ctx() also signs
 certificate B<x> but uses the parameters contained in digest context B<ctx>.
 
-X509_verify() verifies the signature of certificate B<x> using public key
-B<pkey>. Only the signature is checked: no other checks (such as certificate
-chain validity) are performed.
+X509_verify_ex() verifies the signature of certificate B<x> using public key
+B<pkey>. Any cryptographic algorithms required for the verification are fetched
```
t8m, line 36: `certificate I<a> .... public key I<r> ` and/or fix the prototypes to use more descriptive parameters.

doc/man3/X509_sign.pod
```diff
@@ -2,21 +2,25 @@
...
+X509_CRL_sign_ctx, X509_CRL_verify
+- sign or verify certificate, certificate request or CRL signature
 
 =head1 SYNOPSIS
 
  #include <openssl/x509.h>
 
  int X509_sign(X509 *x, EVP_PKEY *pkey, const EVP_MD *md);
  int X509_sign_ctx(X509 *x, EVP_MD_CTX *ctx);
+ int X509_verify_ex(X509 *a, EVP_PKEY *r, OPENSSL_CTX *libctx, const char *propq);
```
t8m, line 16: Nit: move it after the X509_verify.

doc/man3/X509_sign.pod
mattcaswell: Fixed

doc/man3/X509_sign.pod
mattcaswell: Hmm. I actually think `X509_verify_ex` should come first. That is the way it is described in the text below. So, for consistency I also moved `X509_REQ_verify_ex` to be before `X509_REQ_verify`

mattcaswell:
Fixups pushed. I also rebased to fix a conflict with master.

mattcaswell:
Travis failure is strange but unrelated to this PR.

mattcaswell:
Ping?

doc/man3/X509_sign.pod
```diff
@@ -55,10 +64,10 @@ X509_sign(), X509_sign_ctx(), X509_REQ_sign(), X509_REQ_sign_ctx(),
...
-X509_verify(), X509_REQ_verify() and X509_CRL_verify() return 1 if the
-signature is valid and 0 if the signature check fails. If the signature
-could not be checked at all because it was invalid or some other error
-occurred then -1 is returned.
+X509_verify_ex(), X509_verify(), X509_REQ_verify_ex(), X509_REQ_verify() and
+X509_CRL_verify() return 1 if the signature is valid and 0 if the signature
+check fails. If the signature could not be checked at all because it was invalid
+or some other error occurred then -1 is returned.
 
 =head1 SEE ALSO
```
slontis, line 72: Should we have a libctx reference here also?

mattcaswell: Done!

mattcaswell:
Fixup pushed addressing the nit. Also rebased to resolve conflicts with master.

doc/man3/X509_sign.pod
```diff
@@ -2,37 +2,46 @@
...
-digest B<md> and sets the signature in B<x>. X509_sign_ctx() also signs
-certificate B<x> but uses the parameters contained in digest context B<ctx>.
+X509_sign() signs certificate I<x> using private key I<pkey> and message
+digest I<md> and sets the signature in I<x>. X509_sign_ctx() also signs
+certificate I<x> but uses the parameters contained in digest context I<ctx>.
 
-X509_verify() verifies the signature of certificate B<x> using public key
-B<pkey>. Only the signature is checked: no other checks (such as certificate
-chain validity) are performed.
+X509_verify_ex() verifies the signature of certificate I<x> using public key
```
t8m, line 35: You still have `a` above for the X509 parameter and here you have `I<x>`

doc/man3/X509_sign.pod
```diff
@@ -2,37 +2,46 @@
...
  int X509_REQ_sign_ctx(X509_REQ *x, EVP_MD_CTX *ctx);
- int X509_REQ_verify(X509_REQ *a, EVP_PKEY *r);
+ int X509_REQ_verify_ex(X509_REQ *a, EVP_PKEY *pkey, OPENSSL_CTX *libctx,
+                        const char *propq);
+ int X509_REQ_verify(X509_REQ *a, EVP_PKEY *pkey);
 
  int X509_CRL_sign(X509_CRL *x, EVP_PKEY *pkey, const EVP_MD *md);
  int X509_CRL_sign_ctx(X509_CRL *x, EVP_MD_CTX *ctx);
- int X509_CRL_verify(X509_CRL *a, EVP_PKEY *r);
+ int X509_CRL_verify(X509_CRL *a, EVP_PKEY *pkey);
```
t8m, line 27: Hmm... should we have also X509_CRL_verify_ex?

doc/man3/X509_sign.pod
mattcaswell: Fixed

doc/man3/X509_sign.pod
mattcaswell: Perhaps, and I did consider it. However, X509_CRL_verify is  implemented differently to the other 2 I updated and it looked more complicated to change. Since I don't *need* to do that now for alpha1 I opted to exclude it from this PR.

mattcaswell:
Fixup commit pushed addressing feedback above. I also rebased (twice!) to fix conflicts with master. Please take another look.

mattcaswell:
Pushed. Thanks!

PR #11508, opened 2020-04-09
Run sslapi test with the FIPS module

mattcaswell:
We run sslapitest twice: Once with a non-default library context with the default provider loaded into it, and once with a non-default library context with the FIPS provider loaded into it. In both cases we load the "null" provider into the default context to make sure we don't accidentally pick up algorithms from there.

This will fail when running the sslapitest since it requires all the key gen PRs to be merged first (and I have not included them here). However, aside from the dependencies this should be fairly complete and can be reviewed.

### DIFF
--- a/ssl/ssl_local.h
+++ b/ssl/ssl_local.h
@@ -1760,6 +1760,7 @@ typedef struct sigalg_lookup_st {
 
 typedef struct tls_group_info_st {
     int nid;                    /* Curve NID */
+    const char *keyname;
     int secbits;                /* Bits of security (from SP800-57) */
     uint32_t flags;             /* For group type and applicable TLS versions */
     uint16_t group_id;          /* Group ID */
--- a/ssl/ssl_local.h
+++ b/ssl/ssl_local.h
@@ -1760,7 +1760,7 @@ typedef struct sigalg_lookup_st {
 
 typedef struct tls_group_info_st {
     int nid;                    /* Curve NID */
-    const char *keyname;
+    const char *keytype;
     int secbits;                /* Bits of security (from SP800-57) */
     uint32_t flags;             /* For group type and applicable TLS versions */
     uint16_t group_id;          /* Group ID */
--- a/test/recipes/90-test_sslapi.t
+++ b/test/recipes/90-test_sslapi.t
@@ -19,25 +19,19 @@ use lib srctop_dir('Configurations');
 use lib bldtop_dir('.');
 use platform;
 
+my $no_fips = disabled('fips') || ($ENV{NO_FIPS} // 0);
 
 plan skip_all => "No TLS/SSL protocols are supported by this OpenSSL build"
     if alldisabled(grep { $_ ne "ssl3" } available_protocols("tls"));
 
-plan tests => 3;
+plan tests =>
+    ($no_fips ? 0 : 2)          # FIPS install test + sslapitest with fips
+    + 1;                        # sslapitest with default provider
 
 (undef, my $tmpfilename) = tempfile();
 
-
 $ENV{OPENSSL_MODULES} = bldtop_dir("providers");
 $ENV{OPENSSL_CONF_INCLUDE} = bldtop_dir("providers");
-
-ok(run(app(['openssl', 'fipsinstall',
-            '-out', bldtop_file('providers', 'fipsinstall.cnf'),
-            '-module', bldtop_file('providers', platform->dso('fips')),
-            '-provider_name', 'fips', '-mac_name', 'HMAC',
-            '-macopt', 'digest:SHA256', '-macopt', 'hexkey:00',
-            '-section_name', 'fips_sect'])),
-   "fipsinstall");
 
 ok(run(test(["sslapitest", srctop_dir("test", "certs"),
              srctop_file("test", "recipes", "90-test_sslapi_data",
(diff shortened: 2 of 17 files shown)


mattcaswell:
Fixup pushed containing some code that I had locally but somehow managed to miss including in this PR

test/sslapitest.c
```diff
@@ -3763,9 +3774,9 @@ static int test_ciphersuite_change(void)
 # if !defined(OPENSSL_NO_CHACHA) && !defined(OPENSSL_NO_POLY1305)
```
slontis, line 3774: These ifdef's are no longer relevant if CHACHA is removed here..

test/sslapitest.c
```diff
@@ -3740,9 +3747,13 @@ static int test_ciphersuite_change(void)
...
     /* Create a session based on SHA-256 */
-    if (!TEST_true(create_ssl_ctx_pair(TLS_server_method(), TLS_client_method(),
-                                       TLS1_VERSION, 0,
+    if (!TEST_true(create_ssl_ctx_pair(libctx, TLS_server_method(),
+                                       TLS_client_method(), TLS1_VERSION, 0,
                                        &sctx, &cctx, cert, privkey))
+            || !TEST_true(SSL_CTX_set_ciphersuites(sctx,
+                                                   "TLS_AES_128_GCM_SHA256:"
+                                                   "TLS_AES_256_GCM_SHA384:"
+                                                   "TLS_AES_128_CCM_SHA256"))
```
slontis, line 3756: Are these extra cipher suites needed here?

test/sslapitest.c
```diff
@@ -4927,6 +4955,9 @@ static int test_export_key_mat(int tst)
     if (tst == 1)
         return 1;
 #endif
+    /* TODO(3.0): No TLSv1.0/TLSv1.1 support with FIPS at the moment */
```
slontis, line 4958: I thought we agreed to not supporting v1.X in FIPS MODE?

test/sslapitest.c
kaduk: They're not extra -- this is fewer than the default list (e.g., chacha/poly is excluded)

levitte:
You might want to edit the subject, as #11328, #11303, and #11332 are now merged

mattcaswell:
> You might want to edit the subject, as #11328, #11303, and #11332 are now merged

Done

test/sslapitest.c
mattcaswell: This test is about resuming a session where the ciphersuite changes. In TLSv1.3 this is allowed, as long as the hash remains the same. So we need 2 ciphersuites available which both support SHA256. I also wanted one other non-SHA256 ciphersuite. We need to explicitly configure these because the CCM one is not available by default.

test/sslapitest.c
mattcaswell: Fixed

test/sslapitest.c
mattcaswell: I removed the comment

mattcaswell:
Updated to address comments above.

mattcaswell:
I've rebased this now that #11371 has gone in and taken it out of WIP. I also had to add an additional  commit to fixup llibssl a little bit. That commit was included in #11520, but hadn't been included here.

Please take a look!

ssl/ssl_local.h
```diff
@@ -1760,6 +1760,7 @@ typedef struct sigalg_lookup_st {
 
 typedef struct tls_group_info_st {
     int nid;                    /* Curve NID */
+    const char *keyname;
```
levitte, line 1763: We're consistently calling it "keytype" elsewhere, aren't we?

mattcaswell: Are we? Ok. I'll change.

mattcaswell:
Fixup pushed addressing the comment from @levitte above.

slontis:
I have added a change to skip the fips tests is "fips" is disabled.

test/recipes/90-test_sslapi.t
```diff
@@ -19,36 +19,40 @@ use lib srctop_dir('Configurations');
...
                          "passwd.txt"), $tmpfilename, "default",
              srctop_file("test", "default.cnf")])),
              "running sslapitest");
 
-ok(run(test(["sslapitest", srctop_dir("test", "certs"),
-             srctop_file("test", "recipes", "90-test_sslapi_data",
-                         "passwd.txt"), $tmpfilename, "fips",
-             srctop_file("test", "fips.cnf")])),
-             "running sslapitest");
+unless ($no_fips) {
```
paulidale, line 42: Would a SKIP block here be more sensible?  It avoids computing the number of tests in the plan.

slontis: we do it this way in quite a few other places..

slontis: > As for resetting the timer, I'm unsure.

I dont think it needs to restart - all that got changed is that a test that is not supposed to run doesnt run now.

paulidale: I'm fine either way, there isn't a lot of different.

mattcaswell:
> I have added a change to skip the fips tests if "fips" is disabled.

This is fixing up the wrong commit - but that can be addressed during merge

mattcaswell:
Apparently the github UI won't let me approve my own PR. But anyway for the record I approve Shane's last commit - and will sort out during the push to make sure it fixes up the correct commit.

mattcaswell:
Pushed! Thanks.

Issue #11509, opened 2020-04-09
inconsistent API's xxx_with_libctx vs xxx_ex


richsalz:
There are six functions with the suffix ```_with_new_libctx```:
```
;  grep new_with_libctx util/*.num   
util/libcrypto.num:NCONF_new_with_libctx                   ?	3_0_0	EXIST::FUNCTION:
util/libcrypto.num:X509_STORE_CTX_new_with_libctx          ?	3_0_0	EXIST::FUNCTION:
util/libcrypto.num:CT_POLICY_EVAL_CTX_new_with_libctx      ?	3_0_0	EXIST::FUNCTION:CT
util/libcrypto.num:CTLOG_new_with_libctx                   ?	3_0_0	EXIST::FUNCTION:CT
util/libcrypto.num:CTLOG_STORE_new_with_libctx             ?	3_0_0	EXIST::FUNCTION:CT
util/libssl.num:SSL_CTX_new_with_libctx                 ?	3_0_0	EXIST::FUNCTION:
```

But most additional functions added to support a libctx have an "_ex" suffix.  I think these should be renamed. There are no conflicts if that is done.

t8m:
This was resolved with #11512

PR #11512, opened 2020-04-09
Rename XXX_new_with_libctx to XXX_new_ex

richsalz:
Fixes: #11509

### DIFF
--- a/crypto/conf/conf_lib.c
+++ b/crypto/conf/conf_lib.c
@@ -174,7 +174,7 @@ int CONF_dump_bio(LHASH_OF(CONF_VALUE) *conf, BIO *out)
  * the "CONF classic" functions, for consistency.
  */
 
-CONF *NCONF_new_with_libctx(OPENSSL_CTX *libctx, CONF_METHOD *meth)
+CONF *NCONF_new_ex(OPENSSL_CTX *libctx, CONF_METHOD *meth)
 {
     CONF *ret;
 
@@ -193,7 +193,7 @@ CONF *NCONF_new_with_libctx(OPENSSL_CTX *libctx, CONF_METHOD *meth)
 
 CONF *NCONF_new(CONF_METHOD *meth)
 {
-    return NCONF_new_with_libctx(NULL, meth);
+    return NCONF_new_ex(NULL, meth);
 }
 
 void NCONF_free(CONF *conf)
--- a/crypto/conf/conf_mod.c
+++ b/crypto/conf/conf_mod.c
@@ -125,7 +125,7 @@ int CONF_modules_load_file_with_libctx(OPENSSL_CTX *libctx,
     CONF *conf = NULL;
     int ret = 0;
 
-    conf = NCONF_new_with_libctx(libctx, NULL);
+    conf = NCONF_new_ex(libctx, NULL);
     if (conf == NULL)
         goto err;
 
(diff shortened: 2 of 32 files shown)


mattcaswell:
It was an explicit decision following review feedback quite some while ago (I don't recall which PR) to use "with_libctx" in preference to "_ex" as it is more descriptive. Since then I've used it for all "new" type functions. In other functions the "libctx" seemed more incidental and I stuck with "_ex". I dislike "_ex" but have an imagination failure every time I try to think of a new function name. I could live with a decision to make all "_ex" functions that take a libctx be called "with_libctx". Or a decision to just rename any existing "_new_ex" functions that take a libctx to "_new_with_libctx" (I actually think that *should* be done). What I don't think is a good idea is to rename existing "with_libctx" functions to "_ex".

richsalz:
I think it is more important to pick *one scheme* and use it.  Historically, OpenSSL has added the ```_ex``` suffix whenever it added a similar API with new parameters.  It has not added descriptive suffixes.

"Join us" :)  No, seriously, you shouldn't add a new naming convention.  I mean, you're the one who created "RAND_DRBG_new_ex" ...

slontis:
> I think it is more important to pick _one scheme_ and use it. Historically, OpenSSL has added the `_ex` suffix whenever it added a similar API with new parameters. It has not added descriptive suffixes.
> 
> "Join us" :) No, seriously, you shouldn't add a new naming convention. I mean, you're the one who created "RAND_DRBG_new_ex" ...

So what happens when there is already an _ex??, 
ex2?
ex_ex

richsalz:
*   So what happens when there is already an _ex??

I don’t know.  Presumably the OTC decides on a policy. But none of the functions here (nor OPENSSL_CTX_get0_master_drbg etc) have that problem.

richsalz:
Pushed commit to rename OPENSSL_CTX_get0_public_drbg to RAND_DRBG_get0_master_ex, etc.

mattcaswell:
I disagree with the premise behind this PR.

richsalz:
I am working on making DRBG FIPS-compatible.  I just happened to stumble across the OPENSSL_CTX_get0_master_drbg.

Consistent naming conventions help the user, but *also help the OpenSSL developer.*

Sorry you disagree with the premise: consistency is important.

mattcaswell:
> Sorry you disagree with the premise: consistency is important.

I don't disagree with consistency. I disagree with your interpretation of what consistency should look like.

richsalz:
Given an existing API, that has a history of adding "_ex" suffix when additional parameters are added... I am curious to know what you think consistency should look like?  This PR changes two families of functions to match the existing convention.  Clearly all three should not exist.  Which should change, and to what?

richsalz:
rebased to address conflict in libcrypto.num

levitte:
`_ex` was, I think, inspired from Microsoft :astonished: 

https://stackoverflow.com/questions/4039054/what-does-ex-stand-for-in-windows-api-function-names

I'm one of those that find it perfectly non-descript, and worse, if you look at the English language where "ex" usually implies "previous", confusing. 

Also, the issue with needing to change a function that already has that suffix remains unanswered... and we already have a few cases where we went for more descriptive names for that reason among others, the new variants of `EVP_PKEY_CTX_new`.

kaduk:
> inspired from Microsoft

I believe I've even seen some `Ex3` APIs from Microsoft.  But maybe it was "just" a data structure and not an API; it's been a while.

PR #11513, opened 2020-04-09
PROV: Remove duplicate in .../ciphers/build.com

levitte:
TDES files were specified in two places, and had cipher_tdes_common.c
in duplicate.

### DIFF
--- a/providers/implementations/ciphers/build.info
+++ b/providers/implementations/ciphers/build.info
@@ -55,7 +55,7 @@ ENDIF
 
 IF[{- !$disabled{des} -}]
   SOURCE[$TDES_GOAL]=\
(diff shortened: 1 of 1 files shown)


levitte:
This was triggered by a build of my own, where Configure complained that cipher_tdes_common.c appeared twice in the same library.  This is something that some static library programs dislike.

I need to investigate further, as @slontis is right that this fix doesn't make sense.  I was too quick.

oneton:
The same error occurs on 2 of the Travis builds for the most recent master commits. For example https://travis-ci.org/github/openssl/openssl/jobs/673201497

t8m:
I have hopefully correct fix in #11544

PR #11514, opened 2020-04-10
Fix missed fields in EVP_PKEY_meth_copy and  EVP_PKEY_asn1_copy

dev-aaront-org:
I bumped in to these while working on the libp11 PKCS#11 engine.

##### Checklist


crypto/asn1/ameth_lib.c
```diff
@@ -282,7 +285,17 @@ void EVP_PKEY_asn1_copy(EVP_PKEY_ASN1_METHOD *dst,
...
+    dst->pkey_param_check = src->pkey_param_check;
+
+    dst->set_priv_key = src->set_priv_key;
+    dst->set_pub_key = src->set_pub_key;
+    dst->get_priv_key = src->get_priv_key;
+    dst->get_pub_key = src->get_pub_key;
 
+    dst->dirty_cnt = src->dirty_cnt;
+    dst->export_to = src->export_to;
+    dst->import_from = src->import_from;
```
mattcaswell, line 298: There is a slightly neater fix for this in #11494

dev-aaront-org: Nice. I've updated that patch to follow that pattern for EVP_PKEY_meth_copy. It also makes the tests I added pretty redundant so I took those out.

beldmit:
Merged. Thanks!

Issue #8863, opened 2019-05-02
Add exdata to FIPS (etc) crypto objects?


richsalz:
There are a number of ways to use ex_data (as in `CRYPTO_get_ex_new_index.pod`) to cryptographic keys including the generic method using the `CRYPTO_EX_INDEX_xxx` values in `crypto.h`, and key-specific methods such as `RSA_set_ex_data()`, `EVP_MD_meth_set_app_datasize()` (sic), `EVP_PKEY_CTX_set_app_data()` and so on.

Some of those will not be available, of course, as `RSA` type isn't part of the EVP interface (via `EVP_PKEY_keygen()`), but some should still work, right?

Does the "index" value used to store type-specific exdata now vary depending on which provider or core is used?  That seems yucky, and is kind of an ABI break, since applications which used to init and have a single exdata index, now have to have more than one and know when to use each.

My suggestion is to make exdata index and instance pointer lists be a core service, but some disagree.  This discussion has wandered over multiple PR's, and @levitte rightfully suggested opening a new issue.

mattcaswell:
> Does the "index" value used to store type-specific exdata now vary depending on which provider or core is used?

Nope. This remains exactly as it is now.

richsalz:
> Nope. This remains exactly as it is now.

If ```CRYPTO_get_ex_new_index``` is inside the module, then this seems incorrect but maybe because that function isn't module-aware yet.  And looking at that code, since it only has the *class* index there seems to be no way to make it provider-aware.

Since the type-specific functions get a type object (e.g., ```RSA_[sg]et_ex_data``` to use our favorite example) I see how those can be module aware and dispatch to the right ```CRYPTO_[sg]et_ex_data``` but I still dont see how to set the index and callback functions done by ```CRYPTO_get_ex_new_index```

levitte:
A side note: `EVP_MD_meth_set_app_datasize()` and `EVP_PKEY_CTX_set_app_data()` have absolutely nothing to do with exdata.

richsalz:
You're right, I was confused by the similar names. I should have known better :)

levitte:
> And looking at that code, since it only has the class index there seems to be no way to make it provider-aware.

@mattcaswell is making it library context aware in #8857, so if such context is allocated in the provider, it automatically becomes provider aware

levitte:
[...]

PR #11515, opened 2020-04-10
Add ex_data to EVP_PKEY

dev-aaront-org:
I'm looking in to adding EdDSA support to the libp11 PCKS#11 engine and I'll need a place in EVP_PKEY to stash a reference to the corresponding PKCS#11 object. The existing algorithms use the ex_data on the low-level key but that's not an option for ECX keys.

This was also discussed in issue #8863.

##### Checklist


levitte:
Why not add the ex_data to the ECX structure?

mattcaswell:
> Why not add the ex_data to the ECX structure?

Isn't it better to have it on the EVP_PKEY? If we have provider only keys what do you attach the ex_data to?

include/openssl/evp.h
```diff
@@ -1616,6 +1616,11 @@ int EVP_PKEY_param_check(EVP_PKEY_CTX *ctx);
 int EVP_PKEY_private_check(EVP_PKEY_CTX *ctx);
 int EVP_PKEY_pairwise_check(EVP_PKEY_CTX *ctx);
 
+# define EVP_PKEY_get_ex_new_index(l, p, newf, dupf, freef) \
+    CRYPTO_get_ex_new_index(CRYPTO_EX_INDEX_EVP_PKEY, l, p, newf, dupf, freef)
```
mattcaswell, line 1620: Probably we enclose each of these params in "()"

richsalz:
Also please look at #10997 which was waiting for the team to decide whether or not to do this for "all" objects.

include/openssl/evp.h
dev-aaront-org: It's not actually necessary since comma is the lowest precedence operator, and if one of the arguments has a comma, the caller has to put it in () themselves in order to pass it as a single argument. None of the other TYPE_get_ex_new_index macros put the args in ()s so I was just following that. I don't have a problem making the change if you prefer, but I'd change them all so they're consistent.

dev-aaront-org:
> > Why not add the ex_data to the ECX structure?
> 
> Isn't it better to have it on the EVP_PKEY? If we have provider only keys what do you attach the ex_data to?

Agreed. And this is where the discussion seemed to land in #8863, with opaque provider data and ex_data as two separate attributes in EVP_PKEY.

levitte:
> Isn't it better to have it on the EVP_PKEY? If we have provider only keys what do you attach the ex_data to?

You don't at all for the purpose that @dev-aaront-org mentions here, because that is something the backend does. It would have to be included in whatever provider side structure that `pkey->keydata` points at.

mattcaswell:
> You don't at all for the purpose that @dev-aaront-org mentions here, because that is something the backend does. It would have to be included in whatever provider side structure that pkey->keydata points at.

Agreed - it doesn't help on the provider side.l @dev-aaront-org talks about an engine not a provider though - so, if we're talking about engines then placing ex_data on the EVP_PKEY makes sense. Even forgetting engines though it makes sense to me just for any application.

mattcaswell:
Needs second review.

beldmit:
Not sure I understand why this is used only when `#ifndef FIPS_MODE`

The 2nd question - do we need copying this data? If yes, shouldn't the copying method be somewhere in EVP_PKEY_METHOD?

mattcaswell:
> Not sure I understand why this is used only when #ifndef FIPS_MODE

crypto/evp/evp_plib.c gets compiled twice. Once for inclusion in libcrypto and once for inclusion in the FIPS provider. The version that gets included in libcrypto is the version that is available for use by end user applications (regardless of whether they are using the FIPS provider or not). This version gets compiled without `FIPS_MODE` defined. The version that gets included in fips.so is purely for internal use within the FIPS provider itself, i.e. where code inside the FIPS provider is itself making EVP calls. The EVP capabilities inside the FIPS provider are a somewhat restricted version of that included in libcrypto - and does not include support for crypto_ex. This should make no difference to end user applications.

richsalz:
Thoughts on #10997 which starts to do this for more objects?

beldmit:
> Thoughts on #10997 which starts to do this for more objects?

Sorry, I can't say whether it's necessary in that case...

levitte:
> Not sure I understand why this is used only when `#ifndef FIPS_MODE`

Because `FIPS_MODE` is only defined when building the FIPS module, where the application doesn't reach, so ex_data isn't relevant.

I'm starting to think that `FIPS_MODE` was a bad name, and that maybe we should rename it to `FIPS_MODULE`

richsalz:
>I'm starting to think that FIPS_MODE was a bad name, and that maybe we should rename it to FIPS_MODULE

YES!

mattcaswell:
> I'm starting to think that FIPS_MODE was a bad name, and that maybe we should rename it to FIPS_MODULE

Yes - that would be a lot less confusing.

beldmit:
Merged. Many thanks!

PR #11516, opened 2020-04-10
Developer scripts: Release script

levitte:
The old release script that exists in another repository has aged, and
risks becoming messy beyond maintainability if it's made to deal with
multiple OpenSSL version schemes.

A solution, which has been seen in other projects, is to have the
release script as part of the versioned source tree, and ensure it's
adapted for the ongoing version scheme in that source tree.

This introduces dev/, a directory of OpenSSL developer "stuff".  We
may expand it with other practical scripts to easy development setup
and other similar things that developers may need.  For now, it's the
release script dev/release.sh, with auxilliary files in dev/release-aux/.
The script is self describing, the manual is available by running the
command `./dev/release.sh --manual`.

The dev/ directory shall never appear in a source distribution.

##### Checklist

### DIFF
--- a/dev/release-aux/openssl-announce-pre-release.tmpl
+++ b/dev/release-aux/openssl-announce-pre-release.tmpl
@@ -1,21 +1,21 @@
 
-   OpenSSL version $release ($LABEL)
-   ===========================================
+   OpenSSL version $release_text released
+   ======================================
 
    OpenSSL - The Open Source toolkit for SSL/TLS
    https://www.openssl.org/
 
-   OpenSSL $MAJOR.$MINOR.$PATCH is currently in $LABEL.
+   OpenSSL $series is currently in $label.
 
-   OpenSSL $release has now been made available.  For details of
+   OpenSSL $release_text has now been made available.  For details of
    changes and known issues see the release notes at:
 
-        https://www.openssl.org/news/openssl-$MAJOR.$MINOR-notes.html
+        https://www.openssl.org/news/openssl-$series-notes.html
 
    Note: This OpenSSL pre-release has been provided for testing ONLY.
    It should NOT be used for security critical purposes.
 
-   The $LABEL release is available for download via HTTPS and FTP from the
+   The $label release is available for download via HTTPS and FTP from the
    following master locations (you can find the various FTP mirrors under
    https://www.openssl.org/source/mirror.html):
 
--- a/dev/release-aux/openssl-announce-pre-release.tmpl
+++ b/dev/release-aux/openssl-announce-pre-release.tmpl
@@ -1,0 +1,49 @@
+
+   OpenSSL version $release ($LABEL)
+   ===========================================
+
+   OpenSSL - The Open Source toolkit for SSL/TLS
+   https://www.openssl.org/
+
+   OpenSSL $MAJOR.$MINOR.$PATCH is currently in $LABEL.
+
+   OpenSSL $release has now been made available.  For details of
+   changes and known issues see the release notes at:
+
+        https://www.openssl.org/news/openssl-$MAJOR.$MINOR-notes.html
+
+   Note: This OpenSSL pre-release has been provided for testing ONLY.
+   It should NOT be used for security critical purposes.
+
+   The $LABEL release is available for download via HTTPS and FTP from the
+   following master locations (you can find the various FTP mirrors under
+   https://www.openssl.org/source/mirror.html):
+
+     * https://www.openssl.org/source/
+     * ftp://ftp.openssl.org/source/
+
+   The distribution file name is:
+
+    o $tarfile
+      Size: $length
+      SHA1 checksum: $sha1hash
+      SHA256 checksum: $sha256hash
+
+   The checksums were calculated using the following commands:
+
+    openssl sha1 $tarfile
+    openssl sha256 $tarfile
+
+   Please download and check this $LABEL release as soon as possible.
+   To report a bug, open an issue on GitHub:
+
+    https://github.com/openssl/openssl/issues
+
+   Please check the release notes and mailing lists to avoid duplicate
+   reports of known issues. (Of course, the source is also available
+   on GitHub.)
+
+   Yours,
+
+   The OpenSSL Project Team.
+
--- a/dev/release-aux/release-state-fn.sh
+++ b/dev/release-aux/release-state-fn.sh
@@ -1,0 +1,202 @@
+#! /bin/sh
+# Copyright 2020 The OpenSSL Project Authors. All Rights Reserved.
+#
+# Licensed under the Apache License 2.0 (the "License").  You may not use
+# this file except in compliance with the License.  You can obtain a copy
+# in the file LICENSE in the source distribution or at
+# https://www.openssl.org/source/license.html
+
+# This will increase the version number and pre-release tag, according to the
+# current state of the source tree, and the function's first argument (called
+# |next| internally), which is how the caller tells what the next step should
+# be.
+#
+# The possible current source tree states are:
+# ''            The source is in a released state.
+# 'dev'         The source is in development.  This is the normal state.
+# 'alpha', 'alphadev'
+#               The source is undergoing a series of alpha releases.
+# 'beta', 'betadev'
+#               The source is undergoing a series of beta releases.
+# These states are computed from $PRE_LABEL and $TYPE
+#
+# The possible |next| values are:
+# 'alpha'       The source tree should move to an alpha release state, or
+#               stay there.  This trips the alpha / pre-release counter.
+# 'beta'        The source tree should move to a beta release state, or
+#               stay there.  This trips the beta / pre-release counter.
+# 'final'       The source tree should move to a final release (assuming it's
+#               currently in one of the alpha or beta states).  This turns
+#               off the alpha or beta states.
+# ''            The source tree should move to the next release.  The exact
+#               meaning depends on the current source state.  It may mean
+#               tripping the alpha / beta / pre-release counter, or increasing
+#               the PATCH number.
+#
+# 'minor'       The source tree should move to the next minor version.  This
+#               should only be used in the master branch when a release branch
+#               has been created.
+#
+# This function expects there to be a function called fixup_version(), which
+# SHOULD take the |next| as first argument, and SHOULD increase the label
+# counter or the PATCH number accordingly, but only when the current
+# state is "in development".
+
+next_release_state () {
+    local next="$1"
+    local today="$(date '+%-d %b %Y')"
+    local retry=true
+
+    local before="$PRE_LABEL$TYPE"
+
+    while $retry; do
+        retry=false
+
+        $DEBUG >&2 "DEBUG[next_release_state]: BEGIN: \$before=$before"
+        $DEBUG >&2 "DEBUG[next_release_state]: BEGIN: \$next=$next"
+        $DEBUG >&2 "DEBUG[next_release_state]: BEGIN: \$MAJOR=$MAJOR"
+        $DEBUG >&2 "DEBUG[next_release_state]: BEGIN: \$MINOR=$MINOR"
+        $DEBUG >&2 "DEBUG[next_release_state]: BEGIN: \$PATCH=$PATCH"
+        $DEBUG >&2 "DEBUG[next_release_state]: BEGIN: \$TYPE=$TYPE"
+        $DEBUG >&2 "DEBUG[next_release_state]: BEGIN: \$PRE_LABEL=$PRE_LABEL"
+        $DEBUG >&2 "DEBUG[next_release_state]: BEGIN: \$PRE_NUM=$PRE_NUM"
+        $DEBUG >&2 "DEBUG[next_release_state]: BEGIN: \$RELEASE_DATE=$RELEASE_DATE"
+
+        case "$before+$next" in
+            # MAKING ALPHA RELEASES ##################################
+            
+            # Alpha releases can't be made from beta versions or real versions
+            beta*+alpha | +alpha )
+                echo >&2 "Invalid state for an alpha release"
+                echo >&2 "Try --beta or --final, or perhaps nothing"
+                exit 1
+                ;;
+            # For alpha releases, the tag update is dev => alpha or
+            # alpha dev => alpha for the release itself, and
+            # alpha => alpha dev for post release.
+            dev+alpha | alphadev+alpha )
+                TYPE=
+                RELEASE_DATE="$today"
+                fixup_version "alpha"
+                ;;
+            alpha+alpha )
+                TYPE=dev
+                RELEASE_DATE=
+                fixup_version "alpha"
+                ;;
+
+            # MAKING BETA RELEASES ###################################
+
+            # Beta releases can't be made from real versions
+            +beta )
+                echo >&2 "Invalid state for beta release"
+                echo >&2 "Try --final, or perhaps nothing"
+                exit 1
+                ;;
+            # For beta releases, the tag update is dev => beta1, or
+            # alpha{n}-dev => beta1 when transitioning from alpha to
+            # beta, or beta{n}-dev => beta{n} for the release itself,
+            # or beta{n} => beta{n+1}-dev for post release.
+            dev+beta | alphadev+beta | betadev+beta )
+                TYPE=
+                RELEASE_DATE="$today"
+                fixup_version "beta"
+                ;;
+            beta+beta )
+                TYPE=dev
+                RELEASE_DATE=
+                fixup_version "beta"
+                ;;
+            # It's possible to switch from alpha to beta in the
+            # post release.  That's what --next-beta does.
+            alpha+beta )
+                TYPE=dev
+                RELEASE_DATE=
+                fixup_version "beta"
+                ;;
+
+            # MAKING FINAL RELEASES ##################################
+
+            # Final releases can't be made from the main development branch
+            dev+final)
+                echo >&2 "Invalid state for final release"
+                echo >&2 "This should have been preceded by an alpha or a beta release"
+                exit 1
+                ;;
+            # For final releases, the starting point must be a dev state
+            alphadev+final | betadev+final )
+                TYPE=
+                RELEASE_DATE="$today"
+                fixup_version "final"
+                ;;
+            # The final step of a final release is to switch back to
+            # development
+            +final )
+                TYPE=dev
+                RELEASE_DATE=
+                fixup_version "final"
+                ;;
+
+            # SWITCHING TO THE NEXT MINOR RELEASE ####################
+
+            *+minor )
+                TYPE=dev
+                RELEASE_DATE=
+                fixup_version "minor"
+                ;;
+
+            # MAKING DEFAULT RELEASES ################################
+
+            # If we're coming from a non-dev, simply switch to dev.
+            # fixup_version() should trip up the PATCH number.
+            + )
+                TYPE=dev
+                fixup_version ""
+                ;;
+
+            # If we're coming from development, switch to non-dev, unless
+            # the PATCH number is zero.  If it is, we force the caller to
+            # go through the alpha and beta release process.
+            dev+ )
+                if [ "$PATCH" = "0" ]; then
+                    echo >&2 "Can't update PATCH version number from 0"
+                    echo >&2 "Please use --alpha or --beta"
+                    exit 1
+                fi
+                TYPE=
+                RELEASE_DATE="$today"
+                fixup_version ""
+                ;;
+
+            # If we're currently in alpha, we continue with alpha, as if
+            # the user had specified --alpha
+            alpha*+ )
+                next=alpha
+                retry=true
+                ;;
+
+            # If we're currently in beta, we continue with beta, as if
+            # the user had specified --beta
+            beta*+ )
+                next=beta
+                retry=true
+                ;;
+
+            *)
+                echo >&2 "Invalid combination of options"
+                exit 1
+                ;;
+        esac
+
+        $DEBUG >&2 "DEBUG[next_release_state]: END: \$before=$before"
+        $DEBUG >&2 "DEBUG[next_release_state]: END: \$next=$next"
+        $DEBUG >&2 "DEBUG[next_release_state]: END: \$MAJOR=$MAJOR"
+        $DEBUG >&2 "DEBUG[next_release_state]: END: \$MINOR=$MINOR"
+        $DEBUG >&2 "DEBUG[next_release_state]: END: \$PATCH=$PATCH"
+        $DEBUG >&2 "DEBUG[next_release_state]: END: \$TYPE=$TYPE"
+        $DEBUG >&2 "DEBUG[next_release_state]: END: \$PRE_LABEL=$PRE_LABEL"
+        $DEBUG >&2 "DEBUG[next_release_state]: END: \$PRE_NUM=$PRE_NUM"
+        $DEBUG >&2 "DEBUG[next_release_state]: END: \$RELEASE_DATE=$RELEASE_DATE"
+    done
+}
+
--- a/dev/release.sh
+++ b/dev/release.sh
@@ -224,9 +224,10 @@ fi
 
 echo "== Initializing work tree"
 
+get_version
+
 # Generate a cloned directory name
-clone_branch=$(eval $(git cat-file blob HEAD:VERSION)
-               echo "openssl-$MAJOR.$MINOR.x")
+clone_branch="openssl-$SERIES.x"
 release_clone="$clone_branch-release-tmp"
 
 echo "== Work tree will be in $release_clone"
@@ -238,9 +239,10 @@ if ! [ -d "$release_clone" ]; then
 fi
 cd "$release_clone"
 
+get_version
+
 current_branch="$(git rev-parse --abbrev-ref HEAD)"
-new_branch=$(eval $(git cat-file blob HEAD:VERSION)
-             echo "openssl-$MAJOR.$MINOR.x")
+new_branch="openssl-$SERIES.x"
 
 # Check that we're still on the same branch, or on a release branch
 if [ "$current_branch" = "$orig_branch" ]; then
@@ -268,7 +270,6 @@ $DEBUG >&2 "DEBUG: Source directory is $SOURCEDIR"
 
 # Release ############################################################
 
-get_version
 # We always expect to start from a state of development
 if [ "$TYPE" != 'dev' ]; then
     echo >&2 "Not in a development branch"
@@ -315,11 +316,13 @@ fi
 # Write the version information we updated
 set_version
 
-if [ -n "$PRE_RELEASE_TAG" ]; then
-    release=$MAJOR.$MINOR.$PATCH-$PRE_RELEASE_TAG$BUILD_METADATA
+if [ -n "$PRE_LABEL" ]; then
+    release="$VERSION-$PRE_RELEASE_TAG$BUILD_METADATA"
+    release_text="$SERIES$BUILD_METADATA $PRE_LABEL $PRE_NUM"
     announce_template=openssl-announce-pre-release.tmpl
 else
-    release=$MAJOR.$MINOR.$PATCH$BUILD_METADATA
+    release="$VERSION$BUILD_METADATA"
+    release_text="$release"
     announce_template=openssl-announce-release.tmpl
 fi
 tag="openssl-$release"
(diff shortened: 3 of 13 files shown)


levitte:
This is in draft because it's not entirely complete yet.  It doesn't upload anything for the moment, and lacks a few environment checks that the old script has.  However, other than that, it should go through the motions correctly.

@mattcaswell, please run some tests.  There should be no danger to anything you do, as this script clones whatever workspace you start from into a subdirectory, and reverting only involves removing that directory and its contents.

dev/release.sh
```diff
@@ -0,0 +1,614 @@
...
+. ./dev/release-aux/release-version-fn.sh
+. ./dev/release-aux/release-state-fn.sh
+
+# Make sure it's a branch we recognise
+current_branch=$(git rev-parse --abbrev-ref HEAD)
+if (echo "current_branch" \
+        | grep -E -q \
+               -e '^master$' \
+               -e '^OpenSSL_[0-9]+_[0-9]+_[0-9]+[a-z]*-stable$' \
+               -e '^openssl_[0-9]+\.[0-9]+\.x$'); then
```
mattcaswell, line 185: Why .x?

dev/release.sh
```diff
@@ -0,0 +1,614 @@
...
+fi
+
+# Initialize #########################################################
+
+echo "== Initializing work tree"
+
+# Generate a release branch name.  We also use this as part of the cloned
+# directory name
+new_branch=$(eval $(git cat-file blob HEAD:VERSION)
+             echo "openssl-$MAJOR.$MINOR.x")
```
mattcaswell, line 202: .x?

dev/release.sh
```diff
@@ -0,0 +1,614 @@
...
+set_version
+
+if [ -n "$PRE_RELEASE_TAG" ]; then
+    release=$MAJOR.$MINOR.$PATCH-$PRE_RELEASE_TAG$BUILD_METADATA
+    announce_template=openssl-announce-pre-release.tmpl
+else
+    release=$MAJOR.$MINOR.$PATCH$BUILD_METADATA
+    announce_template=openssl-announce-release.tmpl
+fi
+tag="openssl-$release"
```
mattcaswell, line 286: Why the change in tag format? Historically we've used a format like "OpenSSL_1_1_1c"

dev/release.sh
```diff
@@ -0,0 +1,614 @@
...
+B<--verbose> |
+B<--debug> |
+B<--help> |
+B<--manual> |
+]
+
+=head1 DESCRIPTION
+
+B<release.sh> creates an OpenSSL release, given current worktree conditions.
+It will refuses to work unless the current branch is C<master> or a release
```
mattcaswell, line 458: s/refuses/refuse/

dev/release.sh
```diff
@@ -0,0 +1,614 @@
...
+B<--alpha> is only acceptable if the I<PATCH> version number is zero and
+the current state is "in development" or that alpha releases are ongoing.
+
+B<--beta> is only acceptable if the I<PATCH> version number is zero and
+that alpha or beta releases are ongoing.
+
+=item B<--next-beta>
+
+Use together with B<--alpha> to switch to beta releases after the current
+release is done.
```
mattcaswell, line 495: What if we later decide to do another alpha?

dev/release.sh
```diff
@@ -0,0 +1,614 @@
...
+
+=item B<--branch>
+
+Create a branch specific for the I<MAJOR>.I<MINOR>.x release series, if it
+doesn't already exist, and switch to it.  The exact branch name will be
+C<< openssl-I<MAJOR>.I<MINOR>.x >>.
+
+=item B<--no-upload>
+
+Don't upload the produce files.
```
mattcaswell, line 512: s/produce/produced/

dev/release.sh
```diff
@@ -0,0 +1,614 @@
...
+Prior to OpenSSL 3.0, the release branches were named
+C<< OpenSSL_I<version>-stable >>, and the release tags were named
+C<< OpenSSL_I<version> >> for regular releases, or
+C<< OpenSSL_I<version>-preI<n> >> for pre-releases.
+
+From OpenSSL 3.0 ongoing, the release branches are named
+C<< openssl-I<MAJOR>.I<MINOR>.x >>, and the release tags are named
+C<< openssl-I<MAJOR>.I<MINOR>.I<PATCH> >> for regular releases, or
+C<< openssl-I<MAJOR>.I<MINOR>.I<PATCH>-alphaI<n> >> for alpha releases
+and C<< openssl-I<MAJOR>.I<MINOR>.<PATCH>-betaI<n> >> for beta releases.
```
mattcaswell, line 552: Why the change of" OpenSSL_" to "openssl-"?

dev/release.sh
```diff
@@ -0,0 +1,614 @@
...
+    git checkout $git_quiet -b "$new_branch"
+fi
+
+echo "== Configuring OpenSSL for update and release.  This make take a bit of time"
+
+./Configure cc | while read L; do $VERBOSE "> $L"; done
+
+$VERBOSE "== Checking source file updates"
+
+make update | while read L; do $VERBOSE "> $L"; done
```
mattcaswell, line 268: Should we abort if Configure or make fail?

dev/release.sh
```diff
@@ -0,0 +1,614 @@
...
+--no-update     Don't perform 'make update'.
+--verbose       Verbose output.
+--debug         Include debug output.  Implies --no-upload.
+
+--force         Force execution
+
+--help          This text
+--manual        The manual
+
+If none of --major, --minor, --alpha, --beta, or --final are given, this script
```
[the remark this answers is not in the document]
levitte: Thanks, they will disappear

dev/release-aux/release-state-fn.sh
```diff
@@ -0,0 +1,202 @@
...
+#               has been created.
+#
+# This function expects there to be a function called fixup_version(), which
+# SHOULD take the |next| as first argument, and SHOULD increase the label
+# counter or the PATCH number accordingly, but only when the current
+# state is "in development".
+
+next_release_state () {
+    local next="$1"
+    local today=$(date '+%-d %b %Y')
```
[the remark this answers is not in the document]
levitte: Ah.  What shell did you use?  I might have inadvertently used a bash-ism

dev/release.sh
```diff
@@ -0,0 +1,614 @@
...
+while true; do
+    case $1 in
+        --alpha | --beta | --final )
+            next_method=$(echo "x$1" | sed -e 's|^x--||')
+            if [ -z "$next_method2" ]; then
+                next_method2=$next_method
+            fi
+            shift
+            if [ "$next_method" = 'final' ]; then
+                do_branch=true
```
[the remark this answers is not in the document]
levitte: Lines around 206.  A warning, but otherwise inconsequential.  So what I hear is that maybe I should use a different variable name for the implied flags...

dev/release.sh
```diff
@@ -0,0 +1,614 @@
...
+    exit 1
+fi
+
+# Load version functions
+. ./dev/release-aux/release-version-fn.sh
+. ./dev/release-aux/release-state-fn.sh
+
+# Make sure it's a branch we recognise
+current_branch=$(git rev-parse --abbrev-ref HEAD)
+if (echo "current_branch" \
```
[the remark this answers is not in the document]
levitte: Yes it should!

dev/release.sh
```diff
@@ -0,0 +1,614 @@
...
+        | grep -E -q \
+               -e '^master$' \
+               -e '^OpenSSL_[0-9]+_[0-9]+_[0-9]+[a-z]*-stable$' \
```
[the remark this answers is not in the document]
levitte: I made these scripts in such a way that it should be easy to port to 1.1.1 in case there's an interest.  The only things that *should* need a rewrite is `dev/release-aux/release-version-fn.sh`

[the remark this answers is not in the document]
levitte: ... so while I haven't yet planned *to* backport this, I've planned *for the possibility*

dev/release.sh
levitte: It's a branch that's going to have all the PATCH releases, just like we have letter releases in pre-3.0.  `x` represents the series of PATCH numbers, and distinguishes the *branch* from releases that would contain numbers only.

I was inspired by other projects I've been involved in which do exactly this.  https://github.com/darktable-org/darktable is one example.

dev/release.sh
levitte: The old tag format is derived from what CVS allowed back in the days.  CVS branches and tags were quite limited, [they only allow letters, digits, dashes and underscores](http://web.mit.edu/gnu/doc/html/cvs_8.html).  Git branches and tags, on the other hand, allow periods, and actually most characters (only a few, like `:`, are disallowed, for syntactic reasons).  So there's really no reason to continue with the period-to-underscore conversion.

Also, I see a benefit in having the same name as the tar file.  Copy-n-paste, just add `.tar.gz` and voilà!

dev/release.sh
levitte: Tough...  I actually think we need to be strict with ourselves.  But either way, it's also still perfectly possible to simply edit `VERSION` and commit it, with good justification.

dev/release.sh
levitte: Like I said in https://github.com/openssl/openssl/pull/11516/files#r406868833, having tags match the tar file could be beneficial.

dev/release-aux/openssl-announce-pre-release.tmpl
```diff
@@ -0,0 +1,49 @@
...
+
+   OpenSSL $release has now been made available.  For details of
+   changes and known issues see the release notes at:
+
+        https://www.openssl.org/news/openssl-$MAJOR.$MINOR-notes.html
+
+   Note: This OpenSSL pre-release has been provided for testing ONLY.
+   It should NOT be used for security critical purposes.
+
+   The $LABEL release is available for download via HTTP and FTP from the
```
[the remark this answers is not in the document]
levitte: Hah!  Good catch...  this text is an exact copy of what we have in the old script (apart from some variable names), and we never noticed...

dev/release-aux/openssl-announce-pre-release.tmpl
```diff
@@ -0,0 +1,49 @@
...
+
+     * https://www.openssl.org/source/
+     * ftp://ftp.openssl.org/source/
+
+   The distribution file name is:
+
+    o $tarfile
+      Size: $length
+      SHA1 checksum: $sha1hash
+      SHA256 checksum: $sha256hash
```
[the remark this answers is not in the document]
levitte: I'll have a closer look at the space thing.  There is a hack that detects series of `=` and tries to make it the same length as the previous line, so I guess it's a bug there.

dev/release-aux/release-state-fn.sh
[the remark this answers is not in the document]
levitte: Ooooohhhh!  No, that wasn't it.  I have to surround the `$(date ... )` part with quotes.  What that line gave you today was basically this:

```
    local today=10 Jan 2020
```

`Jan` is, of course, a perfectly valid variable name, but `2020` isn't...

dev/release.sh
levitte: I wrote a bit more about it here: https://mta.openssl.org/pipermail/openssl-project/2020-April/001931.html

levitte:
```
cipher_tdes_common.o appears 2 times for the product libcrypto
Died at ./Configure line 2344.
make: *** No rule to make target 'update'. Stop.
```

Yes, thanks, so it wasn't just me.  I tried to fix it, but my fix turned out to be wrong...  fix will come up, but probably past this weekend.

dev/release.sh
levitte: I suggest you look at the very top line...  The `-e` will ensure that *any* error aborts.

dev/release.sh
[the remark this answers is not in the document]
levitte: Done

dev/release.sh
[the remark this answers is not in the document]
levitte: Done, I simply added another variable, `warn_branch`, to tell if the warning should be displayed or not.

dev/release.sh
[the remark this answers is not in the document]
levitte: If we decide that we won't backport this, then it's easy to remove that line.

dev/release.sh
levitte: Fixed

dev/release.sh
levitte: Fixed

dev/release-aux/release-state-fn.sh
```diff
@@ -0,0 +1,202 @@
...
+# 'alpha', 'alphadev'
+#               The source is undergoing a series of alpha releases.
+# 'beta', 'betadev'
+#               The source is undergoing a series of beta releases.
+# These states are computed from $PRE_LABEL and $TYPE
+#
+# The possible |next| values are:
+# 'alpha'       The source tree should move to an alpha release state, or
+#               stay there.  This trips the alpha / pre-release counter.
+# 'beta'        The source tree should move to an beta release state, or
```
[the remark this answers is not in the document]
levitte: Thanks, fixed

dev/release.sh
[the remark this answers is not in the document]
levitte: Fixed

dev/release-aux/release-state-fn.sh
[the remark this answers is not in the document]
levitte: Fixed

dev/release.sh
```diff
@@ -0,0 +1,614 @@
...
+# Update the version information.  This won't save anything anywhere, yet,
+# but does check for possible next_method errors before we do bigger work.
+next_release_state "$next_method"
+
+if $do_branch; then
+    $VERBOSE "== Creating a release branch: $new_branch"
+    git checkout $git_quiet -b "$new_branch"
+fi
+
+echo "== Configuring OpenSSL for update and release.  This make take a bit of time"
```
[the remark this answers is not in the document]
levitte: Fixed

dev/release-aux/openssl-announce-pre-release.tmpl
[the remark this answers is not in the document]
levitte: Fixed, in both places

dev/release.sh
```diff
@@ -0,0 +1,614 @@
+#! /bin/sh -e
+# Copyright 2020 The OpenSSL Project Authors. All Rights Reserved.
+#
+# Licensed under the Apache License 2.0 (the "License").  You may not use
+# this file except in compliance with the License.  You can obtain a copy
+# in the file LICENSE in the source distribution or at
+# https://www.openssl.org/source/license.html
+
+usage () {
+    cat <<EOF
```
richsalz, line 10: Don't you want ```\EOF``` or similar to avoid quote-eating?

dev/release.sh
```diff
@@ -0,0 +1,614 @@
...
+              -l 'no-upload,no-update' \
+              -l 'verbose,debug' \
+              -l 'force' \
+              -l 'help,manual' \
+              -n release.sh -- - "$@")
+eval set -- "$TEMP"
+while true; do
+    case $1 in
+        --alpha | --beta | --final )
+            next_method=$(echo "x$1" | sed -e 's|^x--||')
```
richsalz, line 69: like C switch/case ... undent the body of the case/esac?

dev/release-aux/openssl-announce-pre-release.tmpl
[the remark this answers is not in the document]
levitte: Ouch!  `read` reads words, not lines, not even with just one variable...  so any starting or ending spaces are removed.  I've rewritten the title fixing hack in perl.

dev/release.sh
levitte: Do you know how to make `sh-mode` in emacs do that?

dev/release.sh
levitte: Did you actually check that quotes are eaten?  Here's what `./dev/release.sh --help` displayed to me:

```
: ; ./dev/release.sh --help
Usage: release.sh [ options ... ]

--alpha         Start or increase the "alpha" pre-release tag.
--next-beta     Switch to the "beta" pre-release tag after alpha release.
                It can only be given with --alpha.
--beta          Start or increase the "beta" pre-release tag.
--final         Get out of "alpha" or "beta" and make a final release.
                Implies --branch.

--branch        Create a release branch 'openssl-{major}.{minor}.x',
                where '{major}' and '{minor}' are the major and minor version
                numbers.

--no-upload     Don't upload to upload@dev.openssl.org.
--no-update     Don't perform 'make update'.
--verbose       Verbose output.
--debug         Include debug output.  Implies --no-upload.

--force         Force execution

--help          This text
--manual        The manual

If none of --major, --minor, --alpha, --beta, or --final are given, this script
tries to figure out the next step.
```

dev/release.sh
levitte: ...  it turned out to be easy...

``` elisp
(setq sh-indent-for-case-label 0)
(setq sh-indent-for-case-alt '+)
```

richsalz: almost as easy as "!vi" :)

richsalz:
I have to say I'm surprised you're writing this in shell.  I think perl would be shorter and more understandable.  Yes, seriously.

dev/release.sh
richsalz: the joys of shells.  good.

levitte:
> I have to say I'm surprised you're writing this in shell. I think perl would be shorter and more understandable. Yes, seriously.

I disagree.  I did consider it, but in this case, I actually found shell clearer.  If nothing else, this:

``` sh
    git commit -m 'something'
```

compared to this:

``` perl
    system("git commit -m 'whatever'");
    die "yada yada" if $?;
```

I guess I'm more diverse than you give me credit for

dev/release.sh
levitte: I have no idea what that means...

richsalz:
> I guess I'm more diverse than you give me credit for

No criticism intended.  I was just surprised.  And your explanation makes sense.

dev/release.sh
richsalz: It was a joke.  Spawn a subshell and run vi :)

dev/release.sh
mattcaswell: ...but it didn't

levitte:
> No criticism intended.

Oh, didn't take it as that.  My remark on diversity was fairly light-hearted :wink:

levitte:
I've added the necessary code to upload as well as to specify a non-default signing key.

levitte:
I've added the fixup code for CHANGES.md and NEWS.md, and done some other fixes.  I believe it's ready for a new review, @mattcaswell

dev/release-aux/openssl-announce-pre-release.tmpl
```diff
@@ -0,0 +1,49 @@
+
+   OpenSSL version $release ($LABEL)
```
mattcaswell, line 2: Does it make sense to show the patch number here? We also end up with "alpha" twice. This ends up as:

> OpenSSL version 3.0.0-alpha1 (alpha)

Wouldn't this be better as:

> OpenSSL version 3.0 (alpha1)

dev/release-aux/openssl-announce-pre-release.tmpl
```diff
@@ -0,0 +1,49 @@
...
+   ===========================================
+
+   OpenSSL - The Open Source toolkit for SSL/TLS
+   https://www.openssl.org/
+
+   OpenSSL $MAJOR.$MINOR.$PATCH is currently in $LABEL.
```
mattcaswell, line 8: Similarly here. Not sure about showing the patch number. This ends up as:

>  OpenSSL 3.0.0 is currently in alpha.

To my eye, this is better as:

> OpenSSL 3.0 is currently in alpha

dev/release-aux/openssl-announce-pre-release.tmpl
```diff
@@ -0,0 +1,49 @@
...
+   The $LABEL release is available for download via HTTPS and FTP from the
+   following master locations (you can find the various FTP mirrors under
+   https://www.openssl.org/source/mirror.html):
+
+     * https://www.openssl.org/source/
+     * ftp://ftp.openssl.org/source/
+
+   The distribution file name is:
+
+    o $tarfile
```
mattcaswell, line 27: I'm even unsure whether the patch number belongs in the filename for pre-releases. This comes out as:

> openssl-3.0.0-alpha1.tar.gz

But, is this better as:

> openssl-3.0-alpha1.tar.gz

dev/release.sh
```diff
@@ -0,0 +1,695 @@
...
+        echo >&2 "Internal option error ($next_method, $next_method2)"
+        exit 1
+        ;;
+esac
+
+# Verbosity feed for certain commands
+VERBOSITY_FIFO=/tmp/openssl-$$.fifo
+mkfifo -m 600 $VERBOSITY_FIFO
+( cat $VERBOSITY_FIFO | while read L; do $VERBOSE "> $L"; done ) &
+exec 42>$VERBOSITY_FIFO
```
mattcaswell, line 178: I get this error at this point:

````
dev/release.sh: 178: exec: 42: not found
````

File: dev/release-aux/openssl-announce-pre-release.tmpl (hunk 20 above)
Comment (t8m) [reply dev/release-aux/openssl-announce-pre-release.tmpl:2]: But we are not dropping the last version field in the new versioning scheme, just the letter. So `OpenSSL version 3.0.0 (alpha1)`?

File: dev/release-aux/openssl-announce-pre-release.tmpl (hunk 22 above)
Comment (t8m) [reply dev/release-aux/openssl-announce-pre-release.tmpl:27]: I would not recommend doing 3.0 in pre-release and switching to 3.0.0 after that. It would be confusing.

File: dev/release-aux/openssl-announce-pre-release.tmpl (hunk 21 above)
Comment (t8m) [reply dev/release-aux/openssl-announce-pre-release.tmpl:8]: Here it makes sense because this is rather the name of the release stream.

File: dev/release-aux/openssl-announce-pre-release.tmpl (hunk 20 above)
Comment (mattcaswell) [reply dev/release-aux/openssl-announce-pre-release.tmpl:2]: Right - but does the patch number hold any significance here? Throughout the documentation, when referring to the series of versions we refer to 3.0 - never 3.0.0. In much the same way as we say 1.1.1 as shorthand when referring to 1.1.1a, 1.1.1b, 1.1.1c etc. To me having a specific patch number makes no sense here. It's not an alpha release for a specific patch - its an alpha release for the whole 3.0 series.

Comment (levitte) [reply dev/release-aux/openssl-announce-pre-release.tmpl:2]: Curious, btw, the label should be just "alpha"...  However, I don't really see why we must separate it out, so I would rather throw away the parenthesis altogether.

File: dev/release-aux/openssl-announce-pre-release.tmpl (hunk 22 above)
Comment (levitte) [reply dev/release-aux/openssl-announce-pre-release.tmpl:27]: No, but it would make sense to have openssl-3.0.tar.gz as the final release.  The patch releases that follow it would of course be openssl-3.0.1.tar.gz, openssl-3.0.2.tar.gz etc

File: dev/release.sh (hunk 23 above)
Comment (levitte) [reply dev/release.sh:178]: What shell are you using?  It clearly doesn't follow POSIX specs in that case...

Comment (mattcaswell) [reply dev/release.sh:178]: bash

Comment (levitte) [reply dev/release.sh:178]: See https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_07_05

Comment (levitte) [reply dev/release.sh:178]: Also https://pubs.opengroup.org/onlinepubs/009695399/utilities/exec.html

Comment (levitte) [reply dev/release.sh:178]: Er...  I use bash as well, and famous last words, it works for me.

File: dev/release-aux/openssl-announce-pre-release.tmpl (hunk 22 above)
Comment (t8m) [reply dev/release-aux/openssl-announce-pre-release.tmpl:27]: In theory, yes, but it would be still confusing.

File: dev/release-aux/openssl-announce-pre-release.tmpl (hunk 20 above)
Comment (t8m) [reply dev/release-aux/openssl-announce-pre-release.tmpl:2]: It would be really confusing if we released the final version as 3.0 and then the patch levels were 3.0.1 and so on. And yes the 3.0 refers to the whole series but a concrete release should not lose and gain the patch-level number. At least it is very uncommon among the open-source projects to do so.

File: dev/release.sh (hunk 23 above)
Comment (levitte) [reply dev/release.sh:178]: @mattcaswell and I found that /bin/sh is often linked to dash on Debian-based systems, and that dash isn't quite POSIXly correct, so I'm switching the shebang to specify bash.

File: dev/release-aux/openssl-announce-pre-release.tmpl (hunk 22 above)
Comment (levitte) [reply dev/release-aux/openssl-announce-pre-release.tmpl:27]: I'm sorry, what part was confusing now?

File: dev/release-aux/openssl-announce-pre-release.tmpl (hunk 20 above)
Comment (levitte) [reply dev/release-aux/openssl-announce-pre-release.tmpl:2]: I hear you.  I just had a quick look at diverse software version histories, and indeed, those following a semantic version scheme (or semantic version-ish) do present all three numbers at all times, and only use MAJOR.MINOR to represent the series (i.e. 3.0 would be equivalent to 3.0.x).

Comment (levitte) [reply dev/release-aux/openssl-announce-pre-release.tmpl:2]: The benefit with a stricter three-number release version is that it makes the script simpler.

File: dev/release-aux/openssl-announce-pre-release.tmpl (hunk 22 above)
Comment (levitte) [reply dev/release-aux/openssl-announce-pre-release.tmpl:27]: > I'm sorry, what part was confusing now?

Never mind that...

File: dev/release-aux/openssl-announce-pre-release.tmpl (hunk 21 above)
Comment (levitte) [reply dev/release-aux/openssl-announce-pre-release.tmpl:8]: Related to that, @t8m, would you rather the release branch be called `openssl-3.0.x` or would you prefer `openssl-3.0`?  I do prefer the former, but can go with the latter if that's what everyone else prefers.

Comment (t8m) [reply dev/release-aux/openssl-announce-pre-release.tmpl:8]: I do not really have a preference here, both are OK.

Comment (levitte) [reply dev/release-aux/openssl-announce-pre-release.tmpl:8]: I think I'll keep `.x`, as it makes it more distinct

File: dev/release-aux/openssl-announce-pre-release.tmpl (hunk 20 above)
Comment (levitte) [reply dev/release-aux/openssl-announce-pre-release.tmpl]: I've made changes that clarify all this

**Comment by levitte** [comment]
Per a review discussion, the latest fixup makes it more consistent if we talk in terms of a release series or a specific release version.  Alpha and beta releases are always considered to be for whole releases (we never do alphas and betas within a series, so that makes sense).  The master branch is always considered to be the development of the next series (i.e. the next minor or major release).

I believe I've been able to reflect that line of thinking.

**Comment by t8m** [comment]
LGTM, I'll leave formal approval to someone from OMC as this seems to me to be somewhat within their area.

**Comment by levitte** [comment]
> LGTM, I'll leave formal approval to someone from OMC as this seems to me to be somewhat within their area.

I *am* someone from OMC :wink:

**Comment by t8m** [comment]
I know, but I mean the versioning could be a little bit controversial topic so having someone else from OMC other than the author shouldn't hurt.

File: dev/release.sh [hunk 24]
```diff
@@ -0,0 +1,711 @@
...
+. $HERE/dev/release-aux/release-version-fn.sh
+. $HERE/dev/release-aux/release-state-fn.sh
+
+# Make sure it's a branch we recognise
+orig_branch=$(git rev-parse --abbrev-ref HEAD)
+if (echo "$orig_branch" \
+        | grep -E -q \
+               -e '^master$' \
+               -e '^OpenSSL_[0-9]+_[0-9]+_[0-9]+[a-z]*-stable$' \
+               -e '^openssl_[0-9]+\.[0-9]+\.x$'); then
```
mattcaswell, line 213: This should be "openssl-" instead of "openssl_" (hyphen not underscore). The tar files are using a hyphen for the version name, and the "--branch" option creates it with a hyphen.

dev/release-aux/openssl-announce-pre-release.tmpl
```diff
@@ -0,0 +1,49 @@
...
+
+   OpenSSL - The Open Source toolkit for SSL/TLS
+   https://www.openssl.org/
+
+   OpenSSL $series is currently in $label.
+
+   OpenSSL $release_text has now been made available.  For details of
+   changes and known issues see the release notes at:
+
+        https://www.openssl.org/news/openssl-$series-notes.html
```
mattcaswell, line 13: Just to note that this page (perhaps not unsurprisingly) does not yet exist for 3.0. Do we need to update something to make sure that it is?

levitte: Yes we do, but that's in openssl/web.

dev/release.sh
levitte: Yeah, that was a blunder.  Thanks!

levitte: Fixed

dev/release.sh
kaduk: [obligatory note that assuming bash is present is a linux-ism]

paulidale: OS/X has moved from bash to zsh, I wouldn't be entirely surprised if bash disappeared after a couple more releases.

levitte: I try to write POSIXly correctly. If zsh is POSIXly correct, I don't anticipate problems.

levitte:
Have we settled everything here?

dev/release-aux/openssl-announce-pre-release.tmpl
```diff
@@ -0,0 +1,49 @@
...
+   OpenSSL version $release_text released
+   ======================================
+
+   OpenSSL - The Open Source toolkit for SSL/TLS
+   https://www.openssl.org/
+
+   OpenSSL $series is currently in $label.
+
+   OpenSSL $release_text has now been made available.  For details of
+   changes and known issues see the release notes at:
```
mattcaswell, line 11: Adding a link to the wiki notes would be a useful addition. Either in this PR or a subsequent one.

levitte: Other PR

levitte:
Merged.

b0b0b6a41dd4418bab19cdd28c01ce248fee6af7 Developer scripts: Release script

PR #5860, opened 2018-04-03
Don't trust the OPENSSL_ia32cap envvar more than the cpuid instruction

bernd-edlinger:
Fixes test failures with illegal values in OPENSSL_ia32cap, i.e.:
OPENSSL_ia32cap=0xFFFFFFFFFFFFFFFF:0xFFFFFFFFFFFFFFFF make test

@dot-asm this one's for you ;)

PR #11525, opened 2020-04-12
Dont trust ia32cap envvar more than cpuid 1.1.1

bernd-edlinger:
This is just there to have extended CI tests in 1.1.1 run with #5860 

~It is WIP becasue I don't want it approved by you.~

### DIFF
--- a/crypto/ppccap.c
+++ b/crypto/ppccap.c
@@ -250,41 +250,51 @@ static unsigned long getauxval(unsigned long key)
 #define HWCAP_VEC_CRYPTO        (1U << 25)
 #define HWCAP_ARCH_3_00         (1U << 23)
 
-# if defined(__GNUC__) && __GNUC__>=2
+#if defined(__GNUC__) && __GNUC__>=2
 __attribute__ ((constructor))
-# endif
+#endif
 void OPENSSL_cpuid_setup(void)
 {
+#ifndef OPENSSL_PPCCAP_OVERRIDE
     char *e;
     struct sigaction ill_oact, ill_act;
     sigset_t oset;
+    volatile unsigned int mask;
+#endif
     static int trigger = 0;
 
     if (trigger)
         return;
     trigger = 1;
 
-    if ((e = getenv("OPENSSL_ppccap"))) {
-        OPENSSL_ppccap_P = strtoul(e, NULL, 0);
-        return;
-    }
+#ifdef OPENSSL_PPCCAP_OVERRIDE
+    OPENSSL_ppccap_P = OPENSSL_PPCCAP_OVERRIDE;
+#else
+    if ((e = getenv("OPENSSL_ppccap")))
+        mask = (unsigned int)strtoul(e, NULL, 0);
+    else
+        mask = 0xFFFFFFFF;
 
     OPENSSL_ppccap_P = 0;
 
-#if defined(_AIX)
-    OPENSSL_ppccap_P |= PPC_FPU;
-
+# if defined(_AIX)
     if (sizeof(size_t) == 4) {
         struct utsname uts;
-# if defined(_SC_AIX_KERNEL_BITMODE)
-        if (sysconf(_SC_AIX_KERNEL_BITMODE) != 64)
+#  if defined(_SC_AIX_KERNEL_BITMODE)
+        if (sysconf(_SC_AIX_KERNEL_BITMODE) != 64) {
+            OPENSSL_ppccap_P = mask & PPC_FPU;
             return;
-# endif
-        if (uname(&uts) != 0 || atoi(uts.version) < 6)
+        }
+#  endif
+        if (uname(&uts) != 0 || atoi(uts.version) < 6) {
+            OPENSSL_ppccap_P = mask & PPC_FPU;
             return;
+        }
     }
 
-# if defined(__power_set)
+#  if defined(__power_set)
+    OPENSSL_ppccap_P |= PPC_FPU;
+
     /*
      * Value used in __power_set is a single-bit 1<<n one denoting
      * specific processor class. Incidentally 0xffffffff<<n can be
@@ -309,11 +319,12 @@ void OPENSSL_cpuid_setup(void)
     if (__power_set(0xffffffffU<<17))           /* POWER9 and later */
         OPENSSL_ppccap_P |= PPC_MADD300;
 
+    OPENSSL_ppccap_P &= mask;
     return;
+#  endif
 # endif
-#endif
 
-#if defined(__APPLE__) && defined(__MACH__)
+# if defined(__APPLE__) && defined(__MACH__)
     OPENSSL_ppccap_P |= PPC_FPU;
 
     {
@@ -331,11 +342,12 @@ void OPENSSL_cpuid_setup(void)
                 OPENSSL_ppccap_P |= PPC_ALTIVEC;
         }
 
+        OPENSSL_ppccap_P &= mask;
         return;
     }
-#endif
+# endif
 
-#ifdef OSSL_IMPLEMENT_GETAUXVAL
+# ifdef OSSL_IMPLEMENT_GETAUXVAL
     {
         unsigned long hwcap = getauxval(HWCAP);
         unsigned long hwcap2 = getauxval(HWCAP2);
@@ -364,15 +376,17 @@ void OPENSSL_cpuid_setup(void)
         if (hwcap2 & HWCAP_ARCH_3_00) {
             OPENSSL_ppccap_P |= PPC_MADD300;
         }
+
+        OPENSSL_ppccap_P &= mask;
     }
-#endif
+# endif
 
     sigfillset(&all_masked);
     sigdelset(&all_masked, SIGILL);
     sigdelset(&all_masked, SIGTRAP);
-#ifdef SIGEMT
+# ifdef SIGEMT
     sigdelset(&all_masked, SIGEMT);
-#endif
+# endif
     sigdelset(&all_masked, SIGFPE);
     sigdelset(&all_masked, SIGBUS);
     sigdelset(&all_masked, SIGSEGV);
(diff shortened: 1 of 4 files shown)


bernd-edlinger:
OpenSSL> dgst -sha512 sample3.txt
SHA512(sample3.txt)= 861b6a036e3e2f79e02df261a687b04dc0a359ad36d5cb4d7c821390e43cf190c48cbfb30f98c85e6aa81d076615dec2167a5216b1bb49f798c1f8e35cb284d1
OpenSSL> dgst -sha256 sample3.txt
SHA256(sample3.txt)= eb65278bc9a2f29921c598ebb268d19844d7e6beb53699355cd9d617864bd931
OpenSSL> dgst -sha512 sample4.txt
SHA512(sample4.txt)= cbf24dd4ab1337a213b764f7a5a51c2e1fb5192f945f4ae76b870b84c29b872f8f1ce68196f2f1ab2a0e79cac59d7ff0a274253b7b5f784c3c834c09d5541ad9
OpenSSL> dgst -sha256 sample4.txt
SHA256(sample4.txt)= 31225421ae89c1abbcac87de98c31d09e7cbf4f376f9c7bc00b67ccb74b8a4de
OpenSSL> dgst -sha512 sample5.txt
SHA512(sample5.txt)= f6da079c81917a0e28e79693b7000e77a6d12e8b3aade5c67cd20aea26344954d3f4a14634c8a039c381aec45ec0c6d694254c595b1db47d907ad21b1da50555
OpenSSL> dgst -sha256 sample5.txt
SHA256(sample5.txt)= 6b10ee2e43f19ebf72ec8f9857a73ea3089cea60ebc54b66283b44d4572352ab

crypto/ppccap.c
```diff
@@ -360,50 +374,68 @@ void OPENSSL_cpuid_setup(void)
...
 
-        if (sizeof(size_t) == 4) {
-# ifdef __linux
-            struct utsname uts;
-            if (uname(&uts) == 0 && strcmp(uts.machine, "ppc64") == 0)
-# endif
+            if (sizeof(size_t) == 4 && (mask & PPC_FPU64)) {
+#  ifdef __linux
+                struct utsname uts;
+                if (uname(&uts) == 0 && strcmp(uts.machine, "ppc64") == 0)
```
paulidale, line 386: Style nit: blank line above.

doc/man3/OPENSSL_ia32cap.pod
```diff
@@ -151,6 +153,16 @@ the problem by making build procedure use following script:
 instead of real clang. In which case it doesn't matter which clang
 version is used, as it is GNU assembler version that will be checked.
 
+The environment value is limited to whatever the cpuid instruction
+allows, thus it cannot enable features that are not available by
+the hardware detection.  Previous versions allowed the environment
+variable to enable unavailable hardware features, and thus cause the
```
paulidale, line 159: Suggestion: `s/, and thus cause/ causing/`

doc/man3/OPENSSL_ia32cap.pod
```diff
@@ -151,6 +153,16 @@ the problem by making build procedure use following script:
...
+application to fail.
+
+If the compile option is used, the environment variable is ignored.
```
paulidale, line 162: If the compile _time_ option is used...

levitte:
#5860 got quite a bit of push-back, to the point that you closed it...  do you expect a different outcome here?

bernd-edlinger:
can we talk?

bernd-edlinger:
I only closed #5860 since a major merge conflict was imminent.

paulidale:
I hadn't looked at #5860 much.  Thanks for the reference @levitte 

I'm unsure about 1.1.1 -- this feels like a feature.

t8m:
My opinion on this one is clear - I do not think it is needed in this form.

paulidale:
I can see the use case for freezing the settings at compile time.  That makes some sense.  After reading the discussions, I completely fail to understand why there is any need for masking.

An attack via an environment variable to cause a denial of service to yourself isn't justification.  This would be a local mis-configuration which we shouldn't be trying to mitigate.

An attacker breaking into your system and setting such an environment variable, likewise doesn't justify this.  You've got a lot more to worry about than a denial of service.

t8m:
Yes, possibility to do compile-time freezing would make sense and/or completely disable the CAP envvar support as build-time option for paranoid use-cases.

bernd-edlinger:
This is my proof of concept:
```
ed@w-ed:/tmp$ export OPENSSL_ia32cap=0:0
ed@w-ed:/tmp$ wget https://marc.info/?l=linux-kernel&m=158697302506474&q=mbox&OPENSSL_ia32cap=0xFFFFFFFFFFFFFFFF:0xFFFFFFFFFFFFFFFF
[1] 11378
[2] 11379
[3] 11380
ed@w-ed:/tmp$ --2020-04-20 07:24:11--  https://marc.info/?l=linux-kernel
Auflösen des Hostnamen »marc.info (marc.info)«... 71.33.226.59, 205.134.191.174
Verbindungsaufbau zu marc.info (marc.info)|71.33.226.59|:443... verbunden.
HTTP-Anforderung gesendet, warte auf Antwort... 200 OK
Länge: nicht spezifiziert [text/html]
In »»index.html?l=linux-kernel«« speichern.

    [    <=>                                                                                    ] 101.170      160KB/s   in 0,6s   

2020-04-20 07:24:13 (160 KB/s) - »index.html?l=linux-kernel« gespeichert [101170]


[1]   Fertig                  wget https://marc.info/?l=linux-kernel
[2]-  Fertig                  m=158697302506474
[3]+  Fertig                  q=mbox
ed@w-ed:/tmp$ openssl dgst -sha256 index.html\?l\=linux-kernel 
Ungültiger Maschinenbefehl (Speicherabzug geschrieben)
ed@w-ed:/tmp$ echo $OPENSSL_ia32cap 
0xFFFFFFFFFFFFFFFF:0xFFFFFFFFFFFFFFFF
ed@w-ed:/tmp$ 
```

bernd-edlinger:
Sorry, I am unable to reach you on security@openssl.org,
but this is really most important to be fixed.  It allows an attack
on any system that sets OPENSSL_ia32cap.
The discussion here is pretty much attempting to distract you
from the fact, that this is fixed in ubunu, but they keep the fix
secret.

paulidale:
Shouldn't it be:
```
wget 'https://marc.info/?l=linux-kernel&m=158697302506474&q=mbox&OPENSSL_ia32cap=0xFFFFFFFFFFFFFFFF:0xFFFFFFFFFFFFFFFF'
```

Without the quotes, you are wgetting something and setting m, q and OPENSSL_ia32cap to values.  m and q will be discarded since they are in sub-shells.  The OPENSSL_ia32cap is set.

If it was your intention to not include the quotes, it is user error.  We can't protect against user error.

paulidale:
The urgent label is for fixes to build breaking changes.  I don't think it applies here.

bernd-edlinger:
@paulidale you must be an NSA agent.  This is the proof.

bernd-edlinger:
Of course Paul Dale, is not @paulidale 
Paul Dale is my friend, and he can ask Matthias who has my number.

paulidale:
Proof of what?  That a user error causes problems for OpenSSL?  That's not a surprise.

Try it with this command instead:

```sh
wget https://marc.info/?l=linux-kernel&m=158697302506474&OPENSSL_ia32cap=0xFFFFFFFFFFFFFFFF:0xFFFFFFFFFFFFFFFF&q=mbox
```

I've enabled shell syntax highlighting to indicate the special characters your shell is finding.  Each `&` is acting as a command separator like `;` does.

bernd-edlinger:
Ordinary users don't even know what OPENSSL_ia32cap means.

That is vodoo for them, they are totally suprised by that nonsensical behavior.
Why do you think that is not a CVSS severity 9.0 bug ?

bernd-edlinger:
@openssl/omc please help.

paulidale:
The non-sensical behaviour is the shell's fault not OpenSSL's.  I also don't think that ordinary users run a Unix like system and they don't run commands from the shell.  Caveat emptor.

bernd-edlinger:
No, that bug in the shell was also planted by agents, as this openssl bug.

bernd-edlinger:
I am pretty sure the real target is the TOR browser, which depends on OpenSSL,
I broke it already once, and Matt fixed that, while I was too sick to do it myself.
I can do it again.

bernd-edlinger:
or what do you really want?

bernd-edlinger:
So, decipher this, Pauli Dale.  This is my last e-mail from 18.4.2020:

```
Hi Pauli,

On 4/17/20 6:43 AM, Dr Paul Dale wrote:
> Bernd,
>
> I didn’t find your photograph or birthday.  I didn’t spend a lot of time on it.
>
> Pauli
>

Okay, that is fine, because I had no idea if that is one of the results
that is hidden from me, when I do google from germany, but this means, that
the hidden result is not my picture or my birthday,

So they still belong to myself  and not google (no idea about facebook,
I am not on f.book, but lots of my family, but obviously they have forgotten
about me, becuase g+f=1. QED.

Now I have a another riddle for you  hope you enjoy riddles just as I do.
It is encrypted with AES, and you have the key already, I just added blabla
to the IV, I wonder if you are able to decrypt:


rFoiD81vVuipY0q3NHNF9vPnYTBAv19oVVw6kKSzJZNCz60YxvYguGuwP/t18y97
T9Y0LB8P3XXPtJX/Naywp7rFqQbMAmtlGXmdqKQ9QB1KLUSLjgk189WLaU7abDca
TVymWX01lJeV+FvuZT+Ojk4W3s/Ao6abx+W5QQALUf04agc4AdxQvnoWA1wAmfAx
aa4l5as1RBDjlAwsDF0NqiBburXf14RNM2PJMeHDALeSqoik1SI0xdwpGigaYES4
BSI+CLXZTyp02OFJq7sI3AMWmSiEWGMspR0o3BSU9Oesnec+4J08WNLQ4US2+PNU
s14+Y6CoiCQMkRuGvSrTw35de3PlYV4Yc4S/fzRiraS8AVgtiFdOazB8D0UoTp2R
S6f8+BHXvM7ROajcauq6fUiAJlsm+U2PXbpwMpYRsniywhSFR9C/IrcC4YensNbY
t9zTYfBB8AMWcTIwztq5+6wPvWi2dE6lPhsVPL7fMN1J5ixyob18F7JCwS2o6bal
IFlQ0+UgRvw9YTKPJWpimsKrmQuaul1iU5C20qaDT3zXGVCEmH3/US1Q9KMhj1yn
auKR/8fpjpr1tsLVmNRWLKUrBGRaP/O5pdWZo+gXypffXJMGcBwY7MRLf/u3ypc5
fyJOKE9TKOcTMM9tmdZuEE+MLF2caIC8zOEMsYiemx8Yv/kPiYgqmwFbMh03lQv6
1DyQGiKyXC9Ki7K0GacGwbodS5TXCH14hMQoFlaVm84kYcuHh05nzZuS4Whil90x
yPpuCRUu7DgQJME+N+52sWeYN16t+IUIHOH8N4roXajaaQnv82KAOQVP8S7KxjFm
4lxz+t+guqsojtg3h7RoS1TIyo8SKs4hHEjQderC6oXzF1KqbP+rFuj8bEI5Tk49
AIOTZPBtIhmJaoiAeGk0txqw3B3LecM7t43QCkZbzEjMmbB1gqEL7lbowkfKipXe
hfVoAbrB/LOYfgfiVuSvXYKcohoZHNY5KAShnGVVOeMLFwHMGWHEZWs8PA0K+dPZ
kgtverrIp0y24XKsxWKLWc1rkDAG+kfQ1GL0z6PKHEhDGt9PSkvSxaFGSoHN4jsS
2+wS02PAEPM82Oas8XHUf+uT9/uVHrrBKt/B32mTilSg3PDFOTrRPxNs2MA+gbMW
VDhaT+/iOB3gRyq8TByMExq6OyIa9Kurit+podHrik+ufR4DRgzkGfgEZ8sDzSNJ
BZXtt1esmSlfxntv5T/B0dStvc5sUYarA5gNvjWUDln8r3IZ3BCXItyBqr597WHV
sN/p+HgRHhKu5LNcnu8vl3Tj9bTlFt7gsIZUUSBoia7hWAZ+cTvnkcw1f2oeLEds
OrpdQDNof/Av7ZeoZh9Y7kiEiL2xaNHrTaqdsym/poCByIJE0sg1oSGWUPsQJbcv
9ugjQEyUZzFN32rewVy06iuk4Pn/Ol376wx0Kl+wZPErXyTIU8M1ArgneXzo7v6u
gnsh9RwRf/bgTzycCW7/3e0stvLqUBupN4ySggIUtRPNS4WJrKJsu6ODlxsdQUn/


If you need further hints feel free to ask 

By the way, it is not signed, I will know if you received the
unmodified cipher text if you are able to send a reply to me,
and I will only be able to decrypt it,
if there was no man in the middle attack.


See you on Tuesday,
Bernd.
```

t8m:
Bernd, this is simply a nonsense. If you can trick an user to use wget with a random unquoted URL with a shell, you can trick him to install a root kit for example. Setting OPENSSL_ia32cap is only a single random not-that-much evil thing that you can do via such broken command from unlimited numbers of other many more much more evil examples that one could invent.

kroeckx:
So I think there are 2 ways to look at this:
- Anything that can be used in an attack should be fixed
- We don't see a way this can be exploited without being
  vulnerable to other kind of attacks

You can argue that there are problems with the 2nd way of looking
at things:
- An attacker might find a different way to look at things that
  you currently can't think of
- You might end up in a situation that where you don't fix one
  thing because the other isn't fixed, and the other way around.

So I think that in general, as long as it's reasonable, we should
fix anything an attacker can use. Note that even after the fix,
an attacker will still be able to abuse it by making things
slower, but can no longer prevent the correct working.

That does not mean I consider this a security issue, just
something that should get fixed. And it should get fixed in master
first.

richsalz:
What does compile-time freezing mean?  I hope it doesn't mean that the build machine must exactly match the deployed machines.

slontis:
I seem to remember @mattcaswell discussing this in last week's meeting.
It looks like an unnecessary change to me.

z00z004242:
Hi everyone. 
Is this issue security-related and will be handled as a vulnerability  ? 
Regards, 
W

paulidale:
No, it is not considered a security issue.

paulidale:
> What does compile-time freezing mean?

You specify the CPU capabilities on the configure line.  Those capabilities are always used.

richsalz:
> You specify the CPU capabilities on the configure line. Those capabilities are always used.

That could be an option, I suppose, but there is no way that should be the default.  It would prevent the current run-time checking that OpenSSL does to optimize the assembler code. I am sure any multi-platform provider (heck, even Stepping numbers within X86) would scream about this if they knew.

paulidale:
It's not default.  The use case is embedded devices where the capabilities are known and there is no point probing for them.

richsalz:
Thanks for the explanation, @paulidale  I missed where that was documented.  Can you point to it?

paulidale:
It's in the code at the moment.

kroeckx:
This PR adds it, so it's documented in this PR.

richsalz:
So it's in the code as a comment (here? https://github.com/openssl/openssl/pull/11525/files#diff-aa4b8bc0c4623928e19882b06d2825dbR99) , and a note in *run time* manpage, with no description in the configuration description.  Ugh.

romen:
I just removed the `urgent` API as what is being discussed definitely does not fit into the criteria for the `urgent` treatment.

bernd-edlinger:
@richsalz 
> What does compile-time freezing mean? I hope it doesn't mean that the build machine must exactly match the deployed machines.

No, exactly opposite, the use case is embeded target and/or cross-complile settup
where the build machine has nothing in common with the target, and the target
is exactly defined.  Every part is exactly defined, each chip is fixed, and already
known.  There is no need to add assembler modules that will never run.
But the memory overhead of assembler modules is too high.   Therefore
one is often forced to use no-asm configurations.

bernd-edlinger:
> even Stepping numbers within X86

And embedded target would not change stepping numbers without
thorough testing -- at least not if they are not insane.

richsalz:
Thank you @bernd-edlinger for your excellent description in https://github.com/openssl/openssl/pull/11525#issuecomment-618428973  This info should show up more prominently to users than just PR comments :)

bernd-edlinger:
@t8m 
> Bernd, this is simply a nonsense. If you can trick an user to use wget with a random unquoted URL 

I can't help to wonder if you are trying to protect a back-door here?

bernd-edlinger:
Obviously it is not a problem by itself, but together with the undocumented
behaviour of the bash this makes it a vulnerabiltiy, maybe it would help
if I assign a CVE number for this one on my own?

iamamoose:
This is not a security vulnerability in OpenSSL, therefore does not qualify for a CVE name

bernd-edlinger:
Fine. so we do not fix it?

bernd-edlinger:
oops, wrong button  :-/

romen:
Reading the conversation, it seems like after the discussion we changed the target of the PR.

It seems like it could be valuable to add a build feature that allows fixing the target capabilities at build time and ship only the code supported by the fixed capabilities with the library.
The potential scenario that could most benefit from this feature is (cross-)compiling for constrained-resources targets.

Being a new feature this should be in a PR targeting `master`, I suggest closing this PR and opening a dedicated one with a fitting description.

(I am not excluding that we can have a vote to then backport the feature to 1.1.1, but the process should be to get it approved and merged to master and then approve its backport)

bernd-edlinger:
No.

t8m:
Closing as this is a feature - not for merging to 1.1.1.

PR #11520, opened 2020-04-10
WIP: Alpha1 Test merge

mattcaswell:
This is a test merge of all the key outstanding alpha1 PRs to make sure everything works together. I don't ever plan to merge this specific PR. I will close it when all the PRs it is using have been merged (or when its no longer useful).

While doing this I uncovered a number of problems - many of which I've noted in the outstanding PRs where applicable. In other cases I've had to implement new fixes and workarounds to get it working - which I will raise as new PRs.

This includes the following PRs:
#11328 
#11303 
#11332 
#11371 
#11494 
#11507 
#11508 
#11511 

This means that, with this PR, we are finally able to to sslapitest and ssl_test_new run all the way through using the FIPS provider. I would still like to tackle ssl_test_old.

All tests are passing locally. Lets see what the CIs make of it.

PR #11535, opened 2020-04-13
When calling EC_POINT_point2buf we must use a libctx

mattcaswell:
In a similar way to commit 76e23fc5 we must ensure that we use a libctx
whenever we call EC_POINT_point2buf because it can end up using crypto
algorithms.

This is one of the fixes that I mentioned I had to make in #11520.


mattcaswell:
Pushed. Thanks.

PR #11537, opened 2020-04-13
Make sure we always send an alert in libssl if we hit a fatal error

mattcaswell:
We had a spot where a fatal error was occurring but we hadn't sent an
alert. This results in a later assertion failure.


mattcaswell:
Pushed. Thanks.

PR #11539, opened 2020-04-13
Rename FIPS_MODE to FIPS_MODULE

### LINKED ISSUE #11538: Use FIPS_MODULE not FIPS_MODE for ifdef's
Consensus seems to agree that using FIPS_MODE wasn't a great choice for ifdef'ing code, and that it should be FIPS_MODULE.

levitte:
This macro is used to determine if certain pieces of code should
become part of the FIPS module or not.  The old name was confusing.

Fixes #11538

### DIFF
--- a/crypto/bn/bn_conv.c
+++ b/crypto/bn/bn_conv.c
@@ -46,8 +46,8 @@ char *BN_bn2hex(const BIGNUM *a)
     return buf;
 }
 
-#ifndef FIPS_MODE
-/* No BIO_snprintf in FIPS_MODE */
+#ifndef FIPS_MODULE
+/* No BIO_snprintf in FIPS_MODULE */
 /* Must 'OPENSSL_free' the returned data */
 char *BN_bn2dec(const BIGNUM *a)
 {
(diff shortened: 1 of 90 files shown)


richsalz:
A quick compatibility hack is to do this in e_os.h:
```
#ifdef FIPS_MODULE
# define FIPS_MODE
```
for a time.

mspncp:
@richsalz I think @slontis is more worried about unnecessary merge conflicts caused by all the renamings. Since this change here is a perl (or sed) one-liner which can easily reapplied, I would say it's only fair to respect @slontis' wish.

levitte:
> @richsalz I think @slontis is more worried about unnecessary merge conflicts caused by all the renamings

Most of all, I think we're looking at harsh triage of stuff that *must* go into alpha1 vs what is fairly cosmetic, with approval grace periods and all that.  I see no reason to rush this one...

paulidale:
Is this safe to go in now that alpha1 is released?

slontis:
I am fine with it going in as long as it is up to date..

levitte:
Updated

levitte:
Merged

f844f9eb44186df2f8b0cfd3264b4eb003d8c61a Rename FIPS_MODE to FIPS_MODULE

PR #11542, opened 2020-04-13
Fix an assertion (and a comment) of evp_method_id()

mspncp:


crypto/evp/evp_fetch.c
```diff
@@ -85,7 +85,7 @@ static OSSL_METHOD_STORE *get_evp_method_store(OPENSSL_CTX *libctx)
  */
 static uint32_t evp_method_id(unsigned int operation_id, int name_id)
 {
-    if (!ossl_assert(name_id < (1 << 24) || operation_id < (1 << 8))
+    if (!ossl_assert(name_id < (1 << 24) && operation_id < (1 << 8))
```
richsalz, line 88: IMO this is better broken into separate assertions so user can see what's wrong.
```
if (!ossl_assert(name_id == 0 || operation_id == 0)
    || !ossl_assert((name_id&0xFFFFFF != name_id)
    || !ossl_assert((operation_id & 0xFF) != operation_id))
       return 0;

Comment (mspncp) [reply crypto/evp/evp_fetch.c:88]: I think it's a matter of taste how to group these logical conditions and personally I don't think your expression is more intuitive. In view of the above comment, it is natural to see a range check here. In fact, the only improvement I would do is to rearrange the logical conditions to have a separate range assert for each variable:

```
    if (!ossl_assert(name_id > 0 && name_id < (1 << 24))
        || !ossl_assert(operation_id > 0 && operation_id < (1 << 8)))
        return 0;
```

Then in the (unlikely) case of a failure it will be obvious whether `name_id` or `operation_id` was out of range, which should be enough information for the developer.

Comment (mspncp) [reply crypto/evp/evp_fetch.c:88]: Oh, and it seems like your condition contains a mistake: the first part
```
    !ossl_assert(name_id == 0 || operation_id == 0)
```
inverts the original assertion
```
!ossl_assert(name_id > 0 && operation_id > 0)
```

Comment (mspncp) [reply crypto/evp/evp_fetch.c]: See 55164d68086e534e421ec4ac9753982114664358.

File: crypto/evp/evp_fetch.c [hunk 2]
```diff
@@ -85,8 +85,8 @@ static OSSL_METHOD_STORE *get_evp_method_store(OPENSSL_CTX *libctx)
  */
 static uint32_t evp_method_id(unsigned int operation_id, int name_id)
```
mspncp, line 86: BTW: @levitte I am a little bit irritated about the order of the function parameters `name_id` and `operation_id`, which  is reverted compared to their order of appearance in the comment and the assertions. 

https://github.com/openssl/openssl/blob/55164d68086e534e421ec4ac9753982114664358/crypto/evp/evp_fetch.c#L75-L86

levitte: Feel free to change it

mspncp: Done, see c94bbd7fe661610dcc1c45c5258b4090289adaf3.

mspncp: Ups, there was a mistake.  Commit replaced by 13367d3e947e205350c5a46a27cac8466f9bb3a7.

levitte:
Have a look at Travis, something is really wrong...

mspncp:
> Have a look at Travis, something is really wrong...

Argggh, classical rookie mistake! I had fixed this already, but then I did a `git commit --amend` and forgot to add the `-a` as well. :-/
 
```diff
diff --git a/crypto/evp/evp_fetch.c b/crypto/evp/evp_fetch.c
index c0f52984fa..757b2dca7c 100644
--- a/crypto/evp/evp_fetch.c
+++ b/crypto/evp/evp_fetch.c
@@ -83,7 +83,7 @@ static OSSL_METHOD_STORE *get_evp_method_store(OPENSSL_CTX *libctx)
  * |      name identity     | op id  |
  * +------------------------+--------+
  */
-static uint32_t evp_method_id(int name_idint name_id, unsigned int operation_id)
+static uint32_t evp_method_id(int name_id, unsigned int operation_id)
 {
     if (!ossl_assert(name_id > 0 && name_id < (1 << 24))
         || !ossl_assert(operation_id > 0 && operation_id < (1 << 8)))
```

mspncp:
Merged to master, thanks!

b418980c3f Swap parameters of evp_method_id()
64f849f439 Fix an assertion (and a comment) of evp_method_id()

PR #11544, opened 2020-04-14
Fix no-pic static builds

t8m:
The cipher_tdes_common.c is not needed for single DES and
causes build failure as being duplicated in libcrypto static builds.

### DIFF
--- a/providers/implementations/ciphers/build.info
+++ b/providers/implementations/ciphers/build.info
@@ -63,8 +63,12 @@ IF[{- !$disabled{des} -}]
       cipher_tdes_default.c cipher_tdes_default_hw.c \
       cipher_tdes_wrap.c cipher_tdes_wrap_hw.c
   SOURCE[$DES_GOAL]=\
-      cipher_desx.c cipher_desx_hw.c cipher_tdes_common.c\
+      cipher_desx.c cipher_desx_hw.c \
       cipher_des.c cipher_des_hw.c
+ IF[{- !$disabled{module} -}]
+   SOURCE[$DES_GOAL]=\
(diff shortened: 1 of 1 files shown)


slontis:
desx hijacks it currently - which is why it was needed here..
It may require a bit more work to try to seperate these.

t8m:
OK, let's try to make it conditional.

t8m:
It seems the conditional works.

providers/implementations/ciphers/build.info
```diff
@@ -63,8 +63,12 @@ IF[{- !$disabled{des} -}]
       cipher_tdes_default.c cipher_tdes_default_hw.c \
       cipher_tdes_wrap.c cipher_tdes_wrap_hw.c
   SOURCE[$DES_GOAL]=\
-      cipher_desx.c cipher_desx_hw.c cipher_tdes_common.c\
+      cipher_desx.c cipher_desx_hw.c \
       cipher_des.c cipher_des_hw.c
+ IF[{- !$disabled{shared} -}]
```
levitte, line 68: Please use the same condition as is used in `providers/build.info` to determine if the legacy provider is going to be a module or part of libcrypto.

t8m: OK, done.

t8m:
Proposing as urgent although it is fixing only the extended tests. But it is trivial anyway.

levitte:
Thanks for figuring it out!

t8m:
Merged to master as ca59b00bbd701b9e5e7ce213f44a4d7577d6d2db

PR #11545, opened 2020-04-14
mkbuildinf: Strip fdebug-prefix-map from saved CFLAGS

sebastianas:
The saved `CFLAGS` may contain the `-fdebug-prefix-map=` which is used
to strip the build path from the debug info. If this information remains
recorded then the binary is no longer reproducible if the build
directory changes.

Replace the recorded CFLAG parameter `-fdebug-prefix-map=$OLD=$NEW` with
`-fdebug-prefix-map=<REPRODUCIBLE>`.

##### Checklist

### DIFF
--- a/doc/man3/OpenSSL_version.pod
+++ b/doc/man3/OpenSSL_version.pod
@@ -124,7 +124,9 @@ The value of B<OPENSSL_FULL_VERSION_STR>
 
 The compiler flags set for the compilation process in the form
 C<compiler: ...>  if available, or C<compiler: information not available>
-otherwise.
+otherwise. The path specified in C<-fdebug-prefix-map=>,
+C<-fmacro-prefix-map=> and C<-ffile-prefix-map=> is replaced with a
+fixed string to avoid recording the build path.
 
 =item OPENSSL_BUILT_ON
 
--- a/util/mkbuildinf.pl
+++ b/util/mkbuildinf.pl
@@ -10,6 +10,10 @@ use strict;
 use warnings;
 
 my ($cflags, $platform) = @ARGV;
+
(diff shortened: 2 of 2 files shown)


sebastianas:
ping

sebastianas:
@kroeckx

sebastianas:
A gentle ping.

vagrantc:
You'll want to update this to also handle -fmacro-prefix-map and -ffile-prefix-map (which is a combination of -fdebug-prefix-map and -fmacro-prefix-map). Since Debian's bullseye release, -ffile-prefix-map has been in the defaults that builds using dpkg-buildflags will use.

doc/man3/OpenSSL_version.pod
```diff
@@ -99,7 +99,8 @@ The value of B<OPENSSL_FULL_VERSION_STR>
 
 The compiler flags set for the compilation process in the form
 C<compiler: ...>  if available, or C<compiler: information not available>
-otherwise.
+otherwise. The path specified in C<-fdebug-prefix-map=> is replaced with a
+fix string to avoid recording the build path.
```
paulidale, line 103: `s/fix/fixed/`

kroeckx:
@sebastianas This seems to be waiting for changes from  you

paulidale (commented):
I think that this should either be enabled via a configure option or disabled for debug builds.  I'm not sure which is the better choice.  Suggestions anyone?

sebastianas:
What would be the scenario for this to be useful? That would probably make it easier to decide.
I can't think of one so I didn't think that it is worth preserving.

paulidale:
This option breaks debuggers.  I think that that is a very bad thing to do by _default_.

I know that debuggers can be told the correct source address but I'm confident that a lot of folks wouldn't know how to do this.  By making this an enableable (but disabled by default) option, the user is making the choice to break their debugger.  Likewise, by not enabling this in debug builds, it's less invasive.

PR #11550, opened 2020-04-15
EVP: Fix calls to evp_pkey_export_to_provider()

### LINKED ISSUE #11549: EVP_PKEY_keygen from EVP_PKEY_CTX_new seems to be broken
After commit 2972af109e10c5ce30e548190e3eee28327d6043 there is error in `EVP_PKEY_keygen` if `ctx` is created by `EVP_PKEY_CTX_new`. While it works good if `ctx` is created by `EVP_PKEY_CTX_new_id`.

Reproduction example (based on `man EVP_PKEY_keygen`):
```
#include <openssl/evp.h>
#include <openssl/rsa.h>

int main(void)
{
        EVP_PKEY_CTX *ctx;
        EVP_PKEY *pkey = NULL;
#if 0
        ctx = EVP_PKEY_CTX_new_id(EVP_PKEY_RSA, NULL);
#else
        EVP_PKEY *tkey;
        tkey = EVP_PKEY_new();
        EVP_PKEY_set_type(tkey, EVP_PKEY_RSA);
        ctx = EVP_PKEY_CTX_new(tkey, NULL);
#endif
        if (!ctx)
                OpenSSLDie(__FILE__, __LINE__, "EVP_PKEY_CTX_new_id");
        if (EVP_PKEY_keygen_init(ctx) <= 0)
                OpenSSLDie(__FILE__, __LINE__, "EVP_PKEY_keygen_init");
        if (EVP_PKEY_keygen(ctx, &pkey) <= 0)
                OpenSSLDie(__FILE__, __LINE__, "EVP_PKEY_keygen");
        return 0;
}
```
Produces an error:
```
OpenSSL internal error: EVP_PKEY_keygen
```

This error is mentioned in issue #11510 and found by gost-engine test suite.
@beldmit @levitte @slontis

levitte:
The calls weren't quite right, as this function has changed its behaviour.
We also change the internal documentation of this function, and document
evp_pkey_downgrade().

Fixes #11549

### DIFF
--- a/crypto/evp/pmeth_gn.c
+++ b/crypto/evp/pmeth_gn.c
@@ -170,10 +170,15 @@ int EVP_PKEY_gen(EVP_PKEY_CTX *ctx, EVP_PKEY **ppkey)
             evp_pkey_export_to_provider(ctx->pkey, ctx->libctx,
                                         &tmp_keymgmt, ctx->propquery);
 
-        if (keydata == NULL)
+        if (tmp_keymgmt == NULL)
             goto not_supported;
-        ret = evp_keymgmt_gen_set_template(ctx->keymgmt,
-                                           ctx->op.keymgmt.genctx, keydata);
+        /*
+         * If there is no keydata, then the given pkey is empty, so
+         * there is no template to set
+         */
+        if (keydata != NULL)
+            ret = evp_keymgmt_gen_set_template(ctx->keymgmt,
+                                               ctx->op.keymgmt.genctx, keydata);
     }
 
     /*
--- a/crypto/evp/pmeth_gn.c
+++ b/crypto/evp/pmeth_gn.c
@@ -173,12 +173,11 @@ int EVP_PKEY_gen(EVP_PKEY_CTX *ctx, EVP_PKEY **ppkey)
         if (tmp_keymgmt == NULL)
             goto not_supported;
         /*
-         * If there is no keydata, then the given pkey is empty, so
-         * there is no template to set
+         * It's ok if keydata is NULL here.  The backend is expected to deal
+         * with that as it sees fit.
          */
-        if (keydata != NULL)
-            ret = evp_keymgmt_gen_set_template(ctx->keymgmt,
-                                               ctx->op.keymgmt.genctx, keydata);
+        ret = evp_keymgmt_gen_set_template(ctx->keymgmt,
+                                           ctx->op.keymgmt.genctx, keydata);
     }
 
     /*
--- a/test/evp_extra_test.c
+++ b/test/evp_extra_test.c
@@ -1561,7 +1561,7 @@ static int test_keygen_with_empty_template(int n)
             goto err;
         break;
     case 1:
-        /* Here we create an empty RSA key that serves as ou template */
+        /* Here we create an empty RSA key that serves as our template */
         if (!TEST_ptr(tkey = EVP_PKEY_new())
             || !TEST_true(EVP_PKEY_set_type(tkey, EVP_PKEY_RSA))
             || !TEST_ptr(ctx = EVP_PKEY_CTX_new(tkey, NULL)))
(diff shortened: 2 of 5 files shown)


levitte:
Please suggest a spot for such a test...  or do you expect a whole new test program?

(I was toying with the idea of making a library to collect test cases like this, and a program `test/corners.c` that goes through them all, as well as the recipe `test/recipes/90-test_corner_cases.t`)

mattcaswell:
> Please suggest a spot for such a test

evp_extra_test?

levitte:
> evp_extra_test?

Ah!  That's the pile of random tests!  Thanks :-)

levitte:
Test added

slontis:
The new test fails :)

levitte:
> The new test fails :)

I was tired when I wrote that :-/

Fixed, I hope

test/evp_extra_test.c
```diff
@@ -1540,6 +1540,47 @@ static int test_EVP_PKEY_set1_DH(void)
...
+    int ret = 0;
+
+    switch (n) {
+    case 0:
+        /* We do test with no template at all as well */
+        if (!TEST_ptr(ctx = EVP_PKEY_CTX_new_id(EVP_PKEY_RSA, NULL)))
+            goto err;
+        break;
+    case 1:
+        /* Here we create an empty RSA key that serves as ou template */
```
paulidale, line 1564: `s/ ou / our /`

levitte: Fixed

crypto/evp/pmeth_gn.c
```diff
@@ -153,10 +153,15 @@ int EVP_PKEY_gen(EVP_PKEY_CTX *ctx, EVP_PKEY **ppkey)
...
             goto not_supported;
-        ret = evp_keymgmt_gen_set_template(ctx->keymgmt,
-                                           ctx->op.keymgmt.genctx, keydata);
+        /*
+         * If there is no keydata, then the given pkey is empty, so
+         * there is no template to set
+         */
+        if (keydata != NULL)
+            ret = evp_keymgmt_gen_set_template(ctx->keymgmt,
+                                               ctx->op.keymgmt.genctx, keydata);
```
levitte, line 164: My [comment in #11371](https://github.com/openssl/openssl/pull/11371#pullrequestreview-394386991) got me wondering if it's correct to make the call conditional on a non-NULL `keydata`...
This is compatible with the `EVP_PKEY_METHOD` key generators we have, which simply ignore the template EVP_PKEY unless they are interested in its parameters, but that's arguably a bug (an omission) rather than an intended feature.  In that respect, the code in #11549 is actually broken, as RSA doesn't support this sort of template (i.e. it's  ignored in the `EVP_PKEY_METHOD` key generators).

Either way, these lines are only partially backward compatible.  If the template has been populated with actual key data, `evp_keymgmt_gen_set_template` will fail if OP_keymgmt_gen_set_template hasn't been implemented.

So the question remains, should `evp_keymgmt_gen_set_template` return 1 rather than 0 if OP_keymgmt_gen_set_template isn't implemented?  That would be fully backward compatible, but arguably wrong...

slontis: I think you are asking a question that may not have a good answer.
Potentially breaking existing applications (even if it is crappy code) doesnt sound like a great idea.

levitte: Ok, so basically, your answer is that `evp_keymgmt_gen_set_template` should return 1 if OP_keymgmt_gen_set_template isn't implemented, right?

However, looking at what for example `pkey_dsa_keygen` (`crypto/dsa/dsa_pmeth.c`) does, it's perfectly acceptable to pass a NULL `keydata`, so the call to `evp_keymgmt_gen_set_template` should be unconditional, subject to the backend's judgement, if OP_keymgmt_gen_set_template is implemented.

slontis:
Pauli you wanna merge this?

paulidale:
Yep, merging.

paulidale:
Merged.

PR #11553, opened 2020-04-16
INSTALL: document 'no-ui-console' rather than 'no-ui'

### LINKED ISSUE #11551: "no-ui-console" configuration option is not documented in the INSTALL file
I noticed this with OpenSSL 1.1.1f. In the CHANGES file it says:

     To disable, configure with 'no-ui-console'.  'no-ui' is still
     possible to use as an alias.  Check at compile time with the
     macro OPENSSL_NO_UI_CONSOLE.  The macro OPENSSL_NO_UI is still
     possible to check and is an alias for OPENSSL_NO_UI_CONSOLE.

This sounds to me that `no-ui-console` configuration option should be used going forward and that `no-ui` is only available for backward compatibility. But only `no-ui` is documented in the INSTALL file.

levitte:
The UI interface itself is never disabled, but the console backend may
be.  'no-ui' is a deprecated backward compatibility alias for
'no-ui-console'.

Fixes #11551

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -861,11 +861,11 @@ never be used in production environments.  It will only work when used with gcc
 or clang and should be used in conjunction with the `-DPEDANTIC` option
 (or the `--strict-warnings` option).
 
-### no-ui ###
+### no-ui-console ###
 
-Don't build with the User Interface (UI) capability
+Don't build with the User Interface (UI) console method
 
-The User Interface is the set of features enabling text based prompts.
+The User Interface console method enables text based console prompts.
 
 ### enable-unit-test ###
 


levitte:
Should this be backported to 1.1.1?

levitte:
1.1.1 backport in #11554

levitte:
Merged

301ea192c7e0df8a58bf418a147ee8bdb7f1305e INSTALL: document 'no-ui-console' rather than 'no-ui'

richsalz:
What happens, such as to the commands, if the UI console is disabled?

levitte:
> What happens, such as to the commands, if the UI console is disabled?

The commands will simply not prompt, just tell the user that a password is required, and the user is forced to use the `-pass` type command options.

richsalz:
My attempted Socratic question was that this should have been in the docs.  not worth a new PR for that.

PR #11554, opened 2020-04-16
INSTALL: document 'no-ui-console' rather than 'no-ui' [1.1.1]

### LINKED ISSUE #11551: "no-ui-console" configuration option is not documented in the INSTALL file

levitte:
The UI interface itself is never disabled, but the console backend may
be. 'no-ui' is a deprecated backward compatibility alias for
'no-ui-console'.

Fixes #11551

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -535,9 +535,9 @@
                    conjunction with the "-DPEDANTIC" option (or the
                    --strict-warnings option).
 
-  no-ui
-                   Don't build with the "UI" capability (i.e. the set of
-                   features enabling text based prompts).
+  no-ui-console
+                   Don't build with the "UI" console method (i.e. the "UI"
+                   method that enables text based console prompts).
 
   enable-unit-test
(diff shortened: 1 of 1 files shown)


levitte:
Merged

18fefa35f6a28bc12918fc3b3ed881696d3f438b INSTALL: document 'no-ui-console' rather than 'no-ui'

PR #11555, opened 2020-04-16
s_time: Allow using -CAfile option as in other commands

t8m:
The s_time command in difference from all the other similar
commands supported -cafile option instead of -CAfile.
Add the -CAfile option and keep -cafile only for backwards
compatibility.

##### Checklist


t8m:
Merged to 1.1.1 as a38369db32

PR #11565, opened 2020-04-18
TEST: make and use a fipsinstall script

levitte:
We have copies of the exact same fipsinstall call in several test
recipes.  This refactors those calls into a single simple script.

### DIFF
--- a/test/fipsinstall.pl
+++ b/test/fipsinstall.pl
@@ -1,0 +1,53 @@
+#! /usr/bin/env perl
+
+use strict;
+use warnings;
+
+use File::Spec;
+
+use if $^O eq "VMS", "VMS::Filespec";
+
+my $bldtop_dir;
+
+# First script argument MUST be the build top directory
+BEGIN {
+    $bldtop_dir = $ARGV[0];
+    # 'use lib' needs Unix-ish paths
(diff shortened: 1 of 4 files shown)


paulidale:
Assuming the CIs agree.

levitte:
Merged

ae6b654b669638882a6ddce012ff55adc7cf6a82 TEST: make and use a fipsinstall script

t8m:
@levitte The Travis breakage on OS/X was relevant.

levitte:
I'm trying to investigate that.
(if someone would toss a working Mac my way, that would be very helpful :wink:)

test/fipsinstall.pl
```diff
@@ -0,0 +1,53 @@
...
+    $bldtop_dir = VMS::Filespec::unixpath($bldtop_dir) if $^O eq "VMS";
+}
+
+use lib $bldtop_dir;
+use FindBin;
+use lib "$FindBin::Bin/../Configurations";
+use platform;
+
+my @providers = ($bldtop_dir, 'providers');
+my $fips_cnf = File::Spec->catfile(@providers, 'fipsinstall.cnf');
```
richsalz, line 25: Sigh.  This is a a really bad name for the config file.  It is about configuration of the FIPS *module*.  I will update my "rename things" PR. :(

PR #11566, opened 2020-04-18
Build files: add module installation targets

### LINKED ISSUE #11559: The files fips.so and legacy.so are never installed
Running "make install" does not seem to install the fips.so and legacy.so files anywhere.

Should "make install" also run "openssl fipsinstall"? Or is that better left as an explicit step for the user?

levitte:
We only installed engines, now we also install other modules.

Fixes #11559

### DIFF
--- a/Configurations/descrip.mms.tmpl
+++ b/Configurations/descrip.mms.tmpl
@@ -50,6 +50,10 @@
       grep { !$unified_info{attributes}->{modules}->{$_}->{noinst}
              && $unified_info{attributes}->{modules}->{$_}->{engine} }
       @{$unified_info{modules}};
+  our @install_modules =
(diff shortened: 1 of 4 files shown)


paulidale:
As part of `make install` is a good suggestion but wouldn't be applicable to all platforms.  E.g. one of the binary based Linux distros, they don't `make install` and cannot have the FIPS install done ahead of time.

It's probably okay to do it and have an ability to just uninstall the FIPS install resides.

levitte:
Let's figure things out an do that as a separate PR, yeah?

levitte:
Merged

4114964865435edc475c9ba49a7fa2b78956ab76 Build files: add module installation targets

PR #11572, opened 2020-04-20
Add ACVP tests

slontis:
For FIPS validation purposes - Automated Cryptographic Validation Protocol (ACVP) tests need to be performed. (See https://github.com/usnistgov/ACVP). These tests are very similiar to the old CAVS tests.

This PR uses a hardwired subset of these test vectors to perform similiar operations,
to show the usage and prove that the API's are able to perform the required operations.
It may also help with communication with the lab (i.e- The lab could add a test here to show
a unworking use case - which we can then address).

The EVP layer performs these tests instead of calling lower level API's
as was done in the old FOM.
Some of these tests require access to internals that are not normally allowed/required.

The config option 'acvp_tests' (enabled by default) has been added so that this
access may be removed.

The mechanism has been implemented as additional OSSL_PARAM values that can be set and get. A callback mechanism did not seem to add any additional benefit.
These params will not be added to the gettables lists.

##### Checklist


slontis:
This is not is a reviewable state yet - just up here as a reference while I rebase.

test/rsa_sp800_56b_test.c
```diff
@@ -33,63 +33,6 @@ int setup_tests(void)
 static const unsigned char cav_e[] = {
     0x01,0x00,0x01
 };
-static const unsigned char cav_Xp[] = {
```
slontis, line 36: This test has been moved to the ACVP tests

slontis:
rebased..

ping

test/acvp_test.c
```diff
@@ -0,0 +1,1370 @@
...
+        goto err;
+
+    if (!TEST_ptr(key_ctx = EVP_PKEY_CTX_new_from_pkey(libctx, pkey, ""))
+        || !TEST_int_gt(EVP_PKEY_sign_init(key_ctx), 0)
+        || (params != NULL
+            && !TEST_true(EVP_PKEY_CTX_set_params(key_ctx, params)))
+        || !TEST_int_gt(EVP_PKEY_sign(key_ctx, NULL, &sig_len, digest,
+                                      digest_len), 0)
+        || !TEST_ptr(sig = OPENSSL_zalloc(sig_len))
+        || !TEST_int_gt(EVP_PKEY_sign(key_ctx, sig, &sig_len, digest,
```
mattcaswell, line 104: Why are we not just using EVP_DigestSign* here?

test/acvp_test.c
```diff
@@ -0,0 +1,1370 @@
...
+
+    if (!TEST_ptr(md = EVP_MD_fetch(libctx, tst->digest_alg, ""))
+        || !TEST_true(EVP_Digest(tst->msg, tst->msg_len, digest, &digest_len,
+                                 md, NULL)))
+        goto err;
+
+    if (!TEST_int_gt((sig_len = i2d_ECDSA_SIG(sign, &sig)), 0)
+        || !TEST_ptr(ctx = EVP_PKEY_CTX_new_from_pkey(libctx, pkey, ""))
+        || !TEST_int_gt(EVP_PKEY_verify_init(ctx), 0)
+        || !TEST_int_eq(EVP_PKEY_verify(ctx, sig, sig_len, digest, digest_len),
```
mattcaswell, line 319: Why not EVP_DigestVerify*?

test/acvp_test.c
```diff
@@ -0,0 +1,1370 @@
...
+    unsigned char *priv = NULL, *pub = NULL;
+    size_t priv_len = 0, pub_len = 0;
+    const struct dsa_paramgen_st *tst = &dsa_keygen_data[id];
+
+    if (!TEST_ptr(param_key = dsa_paramgen(tst->L, tst->N))
+        || !TEST_ptr(keygen_ctx = EVP_PKEY_CTX_new_from_pkey(libctx, param_key,
+                                                             NULL))
+        || !TEST_int_gt(EVP_PKEY_keygen_init(keygen_ctx), 0))
+        goto err;
+    for (i = 0; i < 2; ++i) {
```
mattcaswell, line 405: Why do we do this twice?

test/acvp_test.c
```diff
@@ -0,0 +1,1370 @@
...
+
+    ret = 1;
+err:
+    BN_CTX_free(bn_ctx);
+    EVP_PKEY_free(param_key);
+    EVP_PKEY_CTX_free(key_ctx);
+    return ret;
+}
+
+/* Extract r and s  from a dsa signature */
```
mattcaswell, line 566: Nit: spurious extra space here

test/acvp_test.c
```diff
@@ -0,0 +1,1370 @@
...
+
+    if (!TEST_ptr(md = EVP_MD_fetch(libctx, tst->digest_alg, ""))
+        || !TEST_true(EVP_Digest(tst->msg, tst->msg_len,
+                                 digest, &digest_len, md, NULL)))
+        goto err;
+
+    if (!TEST_int_gt((sig_len = i2d_DSA_SIG(sign, &sig)), 0)
+        || !TEST_ptr(ctx = EVP_PKEY_CTX_new_from_pkey(libctx, pkey, ""))
+        || !TEST_int_gt(EVP_PKEY_verify_init(ctx), 0)
+        || !TEST_int_eq(EVP_PKEY_verify(ctx, sig, sig_len, digest, digest_len),
```
mattcaswell, line 670: EVP_DigestVerify?

test/acvp_test.c
```diff
@@ -0,0 +1,1370 @@
...
+        || !TEST_true(EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_AEAD_SET_TAG, tag_len,
+                                          enc ? NULL : (void *)tag))
+        || !TEST_true(EVP_CipherInit_ex(ctx, NULL, NULL, key, iv, enc))
+        || !TEST_true(EVP_CIPHER_CTX_set_padding(ctx, 0))
+        || !TEST_true(EVP_CipherUpdate(ctx, NULL, &len, NULL, pt_len))
+        || !TEST_true(EVP_CipherUpdate(ctx, NULL, &len, aad, aad_len))
+        || !TEST_int_eq(EVP_CipherUpdate(ctx, out, &len, pt, pt_len), pass))
+        goto err;
+
+    if (!pass) {
```
mattcaswell, line 766: Maybe I'm misunderstanding the intent here. Shouldn't we always go ahead and verify the tag? But if pass is 0 we expect the tag verification to fail. Here we just don't do the tag verification at all.

test/acvp_test.c
```diff
@@ -0,0 +1,1370 @@
...
+
+    return aes_ccm_enc_dec(tst->alg, tst->pt, tst->pt_len,
+                           tst->key, tst->key_len,
+                           tst->iv, tst->iv_len, tst->aad, tst->aad_len,
+                           tst->ct, ct_len, tag, tag_len, enc, pass)
+            && aes_ccm_enc_dec(tst->alg, tst->ct, ct_len,
+                               tst->key, tst->key_len,
+                               tst->iv, tst->iv_len, tst->aad, tst->aad_len,
+                               tst->pt, tst->pt_len, tag, tag_len, !enc, pass)
+            /* test that it fails it the tag is incorrect */
```
mattcaswell, line 810: s/it the/if the/

test/acvp_test.c
```diff
@@ -0,0 +1,1370 @@
...
+            && aes_ccm_enc_dec(tst->alg, tst->ct, ct_len,
+                               tst->key, tst->key_len,
+                               tst->iv, tst->iv_len, tst->aad, tst->aad_len,
+                               tst->pt, tst->pt_len,
+                               tag - 1, tag_len, !enc, !pass);
```
mattcaswell, line 815: Does `tag - 1` here assume that `ct_len >= 1`? Can it ever be 0?

test/acvp_test.c
```diff
@@ -0,0 +1,1370 @@
...
+    }
+    if (!TEST_true(EVP_CipherInit_ex(ctx, NULL, NULL, key, iv, enc))
+        || !TEST_true(EVP_CIPHER_CTX_set_padding(ctx, 0))
+        || !TEST_true(EVP_CipherUpdate(ctx, NULL, &len, aad, aad_len))
+        || !TEST_true(EVP_CipherUpdate(ctx, out, &len, pt, pt_len)))
+        goto err;
+
+    if (!TEST_int_eq(EVP_CipherFinal_ex(ctx, out + len, &out_len), pass))
+        goto err;
+    if (!pass) {
```
mattcaswell, line 859: This seems the way I would have expected CCM to work.

test/acvp_test.c
```diff
@@ -0,0 +1,1370 @@
...
+#ifndef OPENSSL_NO_RSA
+static EVP_PKEY *rsa_keygen(int bits)
+{
+    EVP_PKEY *key = NULL;
+    EVP_PKEY_CTX *keygen_ctx = NULL;
+
+    if (!TEST_ptr(keygen_ctx = EVP_PKEY_CTX_new_from_name(libctx, "RSA", NULL))
+        || !TEST_int_gt(EVP_PKEY_keygen_init(keygen_ctx), 0)
+        || !TEST_true(EVP_PKEY_CTX_set_rsa_keygen_bits(keygen_ctx, bits))
+/*      || !TEST_true(EVP_PKEY_CTX_set_rsa_keygen_pubexp(keygen_ctx, exp)) */
```
mattcaswell, line 1020: Why is this commented out?

test/acvp_test.c
```diff
@@ -0,0 +1,1370 @@
...
+            break;
+        case OPT_TEST_CASES:
+           break;
+        default:
+        case OPT_ERR:
+            return 0;
+        }
+    }
+
+    libctx = OPENSSL_CTX_new();
```
mattcaswell, line 1324: Should we load "null" into the default ctx?

include/openssl/core_names.h
```diff
@@ -350,6 +349,19 @@ extern "C" {
 #define OSSL_GEN_PARAM_POTENTIAL            "potential" /* integer */
 #define OSSL_GEN_PARAM_ITERATION            "iteration" /* integer */
 
+/* ACVP Test parameters : These should not be used normally */
```
mattcaswell, line 352: Is this intended as public interface, and therefore should they be documented?

mattcaswell:
Travis failure appears relevant.

test/acvp_test.c
slontis: Because the CAVS keygen tests actually run multiple times - and I just wanted to make sure it was working.

test/acvp_test.c
slontis: The test above is !TEST_int_eq(EVP_CipherUpdate(ctx, out, &len, pt, pt_len), pass))..
So doing a final and getting the tag is not the correct thing to do.

test/acvp_test.c
slontis: Put an extra check in - it wont currently happen with the given data.

test/acvp_test.c
slontis: I have no control over that expectation :)

test/acvp_test.c
slontis: It is not needed for this test - but was put there so a user would know that they can add it if required. (That function is used in other tests in this file).

include/openssl/core_names.h
slontis: Hmm that is a tricky question...
They are public really just for the labs validation testing, so they should not be used by 99% of users.

test/acvp_test.c
slontis: Will add - null provider did not exist when I wrote this.

crypto/rsa/rsa_lib.c
```diff
@@ -162,6 +162,11 @@ void RSA_free(RSA *r)
     BN_clear_free(r->dmp1);
     BN_clear_free(r->dmq1);
     BN_clear_free(r->iqmp);
+
+#if defined(FIPS_MODULE) && !defined(OPENSSL_NO_ACVP_TESTS)
+    rsa_acvp_test_free(r->acvp_test);
+#endif
+    /* TODO(3.0): Support PSS in FIPS_MODULE */
```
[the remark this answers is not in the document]
slontis: Not when I wrote it - added pss tests

test/acvp_test.c
mattcaswell: Err...why are we expecting the update to fail?

crypto/dh/dh_lib.c
```diff
@@ -511,6 +511,19 @@ int EVP_PKEY_CTX_set_dhx_rfc5114(EVP_PKEY_CTX *ctx, int gen)
     return EVP_PKEY_CTX_set_dh_rfc5114(ctx, gen);
 }
 
+int EVP_PKEY_CTX_set_dh_group_name(EVP_PKEY_CTX *ctx, const char *group_name)
```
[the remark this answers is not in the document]
mattcaswell: This is in #11914

test/acvp_test.c
mattcaswell: In that case perhaps a comment would be appropriate.

include/openssl/core_names.h
mattcaswell: I tend to think it should still be documented. Some users may engage other labs, and those labs will need to understand this interface.

test/acvp_test.c
slontis: removed

providers/implementations/keymgmt/rsa_kmgmt.c
```diff
@@ -451,6 +455,14 @@ static int rsa_gen_set_params(void *genctx, const OSSL_PARAM params[])
         && !pss_params_fromdata(&gctx->pss_params, params, gctx->rsa_type,
                                 gctx->libctx))
         return 0;
+#if defined(FIPS_MODULE) && !defined(OPENSSL_NO_ACVP_TESTS)
+    /*
+     * Any acvp test related parameters are copied into a params[],
+     * As this is only used for testing it is just a shallow copy.
```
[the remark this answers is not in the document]
slontis: Changed the name to _new and added a _free also - so the data is now memduped.

include/openssl/core_names.h
slontis: Ok.

test/acvp_test.c
slontis: So for GCM the final calls CRYPTO_gcm128_finish which does a memcmp and then it fails..

test/acvp_test.c
slontis: Ok so it happens since the last update passes the ciphertext and does the decrypt - because it does the decrypt is also does the tag check - Since the decrypted data should only be returned IF AND ONLY IF the tag is correct.

crypto/dh/dh_lib.c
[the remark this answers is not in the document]
slontis: Ok I will just remove the function altogether and use setparams..

slontis:
Updated PR to address review comments.

slontis:
rebased and removed doc for setting DH_group_name

mattcaswell:
Travis failure looks relevant (md nits)

test/acvp_test.c
mattcaswell: Is this behaviour specific to CCM mode in general, or specific to CCM mode with the FIPS provider? I thought it was the case (generally) that we did not check the tag until the "Final" call. Otherwise how can we know that we've finished receiving data (there could be more "Update" calls to come with more ciphertext).

slontis: I went back to 1_1_1 and debugged evp_test with CCM input. EVP_Final does nothing. There is even a comment saying final does nothing.
Note that for CCM the MAC is performed on the plaintext.
The test doesnt do fragging for CCM so it is basically a one shot operation.

mattcaswell:
Travis failure is now not relevant.

test/acvp_test.c
```diff
@@ -0,0 +1,1366 @@
...
+    EVP_CIPHER_CTX *ctx;
+    EVP_CIPHER *cipher = NULL;
+    int out_len, len;
+    unsigned char out[1024];
+
+    TEST_note("%s : %s : expected to %s", alg, enc ? "encrypt" : "decrypt",
+              pass ? "pass" : "fail");
+    /*
+     * TODO(3.0): RAND needs to be fetchable - so the IV can be set
+     * the IV should not be set outside the boundary as it is now.
```
mattcaswell, line 821: I don't understand this comment. Why do we need fetchable RAND in order to set the IV inside the boundary? RAND is already available inside the boundary (you don't need it to be fetchable for that).

test/recipes/30-test_acvp.t
```diff
@@ -0,0 +1,43 @@
...
+
+plan skip_all => "ACVP is not supported by this test"
+    if $no_fips || disabled("acvp_tests");
+
+use lib srctop_dir('Configurations');
+use lib bldtop_dir('.');
+use platform;
+
+$ENV{OPENSSL_MODULES} = bldtop_dir("providers");
+$ENV{OPENSSL_CONF_INCLUDE} = bldtop_dir("providers");
```
mattcaswell, line 29: As of 6ed34b3eff68 this should  no longer be necessary

test/acvp_test.c
slontis: What I am trying to say here is that the iv needs to be fed in as a dummy entropy source instead of being supplied via EVP_CipherInit_ex().

test/recipes/30-test_acvp.t
slontis: Ok removed.

test/acvp_test.c
slontis: Changed the text.

slontis:
ping

crypto/ffc/ffc_params_generate.c
```diff
@@ -754,59 +759,63 @@ int ffc_params_FIPS186_4_gen_verify(OPENSSL_CTX *libctx, FFC_PARAMS *params,
...
+
 #ifdef FIPS_MODULE
     /*
      * FIPS 186-4 states that validation can only be done for this pair.
      * (Even though the original spec allowed L = 512 + 64*j (j = 0.. 8))
      */
     if (L != 1024 || N != 160) {
         *res = FFC_CHECK_BAD_LN_PAIR;
-        return FFC_PARAMS_RET_STATUS_FAILED;
+        return FFC_PARAM_RET_STATUS_FAILED;
```
mattcaswell, line 810: Doesn't `md` get leaked here? And at similar return below.

slontis: Well spotted - It is done properly in the very similiar function for 186_4.

slontis:
Thanks. Merged to master.

PR #11575, opened 2020-04-20
doc: remove deprecation notes for apps that are staying.

paulidale:
The apps that are staying are: dhparam, dsa, dsaparam, ec, ecparam, gendsa and
rsa.

The rsautl app remains deprecated as does the -dsaparam option to dhparam also remains deprecated.

This is not required for ɑ1.


levitte:
I thoroughly dislike this move. It means we maintain the burden of double maintenance.

-1

mattcaswell:
> I thoroughly dislike this move. It means we maintain the burden of double maintenance.

It looks to me like we might need an OMC vote for this then.

levitte:
Yup

t8m:
What would the maintenance burden be if we explicitly declare these "legacy" commands as "no more new features"? The commands are convenient and easy to use. We should deprecate them only if the algorithms they support are deprecated as well. I could imagine deprecating them once we are in the post-quantum-crypto world.

I suppose there will be some minimal maintenance burden with them for things like apps code refactorings or cleanups, but that seems to me to be acceptable. And the deprecation actually does not help with it, only when the actual removal happens.

levitte:
It's a burden to have to keep in mind that there are at least two spots to fix when a fix, for code that is more or less duplicated between these commands and their `pkey` counterparts.  We have the unfortunate history of duplication, with the later follow-up fixes that are done in one corner, but not another.

paulidale:
The vote to include this PR passed, removing the hold.

paulidale:
Merged to master.

PR #11573, opened 2020-04-20
evptest: set the DESX provider to be the legacy provider.

paulidale:
This fixes the first of the two Mac build errors.  In this case the legacy provider isn't being used for DESX and the load fails.  It's not clear why this isn't the case everywhere else.

- [ ] documentation is added or updated
- [x] tests are added or updated

PR #11576, opened 2020-04-20
 EVP: Fix EVP_Digest{Sign,Verify}Init() to handle no default digest 

### LINKED ISSUE #11571: EVP_DigestSignInit fails with "ED25519" provider key
OpenSSL master branch.

When a ED25519 key is created with:
EVP_PKEY_CTX_new_from_name(NULL, SN_ED25519, NULL)
then EVP_DigestSignInit(ctx, NULL, NULL, NULL, key) fails.

**Analysis**
The problem is that in do_sigver_init, if mdname is NULL, it is obtained using EVP_PKEY_get_default_digest_name:
https://github.com/openssl/openssl/blob/master/crypto/evp/m_sigver.c#L156

EVP_PKEY_get_default_digest_name returns an empty string ""
https://github.com/openssl/openssl/blob/master/crypto/evp/p_lib.c#L1029
Which is then passed to canon_mdname
canon_mdname returns the same empty string (and not NULL):
https://github.com/openssl/openssl/blob/master/crypto/evp/m_sigver.c#L33

mdname with value "" is then passed to eddsa_digest_signverify_init, which then fails with PROV_R_INVALID_DIGEST,
because it is expecting mdname to be NULL and not an empty string:
https://github.com/openssl/openssl/blob/master/providers/implementations/signature/eddsa.c#L59

**Possible Fix**
Make canon_mdname return NULL if the string is empty. 
Or make EVP_PKEY_get_default_digest_name return "UNDEF" (through the provider param OSSL_PKEY_PARAM_DEFAULT_DIGEST):
https://github.com/openssl/openssl/blob/master/crypto/evp/p_lib.c#L1012

levitte:
EVP_DigestSignInit() and EVP_DigestVerifyInit() would detect if there
is no default digest when using legacy (EVP_PKEY_ASN1_METHOD)
implementations.  However, it doesn't do that when provider side keys
are used.

Furthermore, because EVP_PKEY_get_default_digest_name() was used in
the portion of the code that uses the provider implementation, the
EVP_PKEY_ASN1_METHOD would be used if the key has one attached.  This
is now changed to use evp_keymgmt_util_get_deflt_digest_name()
instead.

Finally, we make sure to detect if the provider implementation
supports the digest name parameters (default or mandatory), and
returns with error if not.  This is what the legacy portion of the

Fixes #11571

### DIFF
--- a/crypto/evp/keymgmt_lib.c
+++ b/crypto/evp/keymgmt_lib.c
@@ -412,3 +412,51 @@ void *evp_keymgmt_util_gen(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
 
     return keydata;
 }
+
+/*
+ * Returns the same numbers as EVP_PKEY_get_default_digest_name()
+ * When the string from the EVP_KEYMGMT implementation is "", we use
+ * SN_undef, since that corresponds to what EVP_PKEY_get_default_nid()
+ * returns for no digest.
+ */
+int evp_keymgmt_util_get_deflt_digest_name(EVP_KEYMGMT *keymgmt,
+                                           void *keydata,
+                                           char *mdname, size_t mdname_sz)
+{
+    OSSL_PARAM params[3];
+    char mddefault[100] = "";
+    char mdmandatory[100] = "";
+    char *result = NULL;
+    int rv = -2;
+
+    params[0] =
+        OSSL_PARAM_construct_utf8_string(OSSL_PKEY_PARAM_DEFAULT_DIGEST,
+                                         mddefault, sizeof(mddefault));
+    params[0].return_size = sizeof(mddefault) + 1;
+    params[1] =
+        OSSL_PARAM_construct_utf8_string(OSSL_PKEY_PARAM_MANDATORY_DIGEST,
+                                         mdmandatory,
+                                         sizeof(mdmandatory));
+    params[1].return_size = sizeof(mdmandatory) + 1;
+    params[2] = OSSL_PARAM_construct_end();
+
+    if (!evp_keymgmt_get_params(keymgmt, keydata, params))
+        return 0;
+
+    if (params[1].return_size != sizeof(mdmandatory) + 1) {
+        if (params[1].return_size == 1) /* Only a NUL byte */
+            result = SN_undef;
+        else
+            result = mdmandatory;
+        rv = 2;
+    } else if (params[0].return_size != sizeof(mddefault) + 1) {
+        if (params[0].return_size == 1) /* Only a NUL byte */
+            result = SN_undef;
+        else
+            result = mddefault;
+        rv = 1;
+    }
+    if (rv > 0)
+        OPENSSL_strlcpy(mdname, result, mdname_sz);
+    return rv;
+}
--- a/doc/man3/EVP_PKEY_get_default_digest_nid.pod
+++ b/doc/man3/EVP_PKEY_get_default_digest_nid.pod
@@ -18,8 +18,8 @@ EVP_PKEY_get_default_digest_nid, EVP_PKEY_get_default_digest_name
 EVP_PKEY_get_default_digest_name() fills in the default message digest
 name for the public key signature operations associated with key
 I<pkey> into I<mdname>, up to at most I<mdname_sz> bytes including the
-ending NUL byte.  The name may turn out to be C<"UNDEF">, signifying
-that no digest should be used.
+ending NUL byte.  The name could be C<"UNDEF">, signifying that no digest
+should be used.
 
 EVP_PKEY_get_default_digest_nid() sets I<pnid> to the default message
 digest NID for the public key signature operations associated with key
--- a/doc/man3/EVP_PKEY_get_default_digest_nid.pod
+++ b/doc/man3/EVP_PKEY_get_default_digest_nid.pod
@@ -18,7 +18,8 @@ EVP_PKEY_get_default_digest_nid, EVP_PKEY_get_default_digest_name
 EVP_PKEY_get_default_digest_name() fills in the default message digest
 name for the public key signature operations associated with key
 I<pkey> into I<mdname>, up to at most I<mdname_sz> bytes including the
-ending NUL byte.
+ending NUL byte.  The name may turn out to be C<"UNDEF">, signifying
+that no digest should be used.
 
 EVP_PKEY_get_default_digest_nid() sets I<pnid> to the default message
 digest NID for the public key signature operations associated with key
--- a/providers/implementations/keymgmt/ecx_kmgmt.c
+++ b/providers/implementations/keymgmt/ecx_kmgmt.c
@@ -207,6 +207,17 @@ static int ecx_get_params(void *key, OSSL_PARAM params[], int bits, int secbits,
     return key_to_params(ecx, NULL, params);
 }
 
+static int ed_get_params(void *key, OSSL_PARAM params[])
+{
+    OSSL_PARAM *p;
+
+    if ((p = OSSL_PARAM_locate(params,
+                               OSSL_PKEY_PARAM_MANDATORY_DIGEST)) != NULL
+        && !OSSL_PARAM_set_utf8_string(p, ""))
+        return 0;
+    return 1;
+}
+
 static int x25519_get_params(void *key, OSSL_PARAM params[])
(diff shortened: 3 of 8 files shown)


levitte:
This turned out to be a logical exercise as well.  Basically, the previous logic in `do_sigver_init()` seemed to be that if the provider implementation didn't specify any default or mandatory digest (i.e. that parameter is unimplemented), then no digest must be used.  However, that doesn't seem to be the logic for legacy implementations, where the interpretation rather seems to be that if the default md nid ctrl (ASN1_PKEY_CTRL_DEFAULT_MD_NID) is unsupported, then the caller *must* specify a digest, i.e. `type` is not allowed to be NULL.

So essentially, we're turning the provider implementation logic around to be legacy compatible, i.e. the provider implementation must provide a default or mandatory digest name if there is such a thing, and the empty string gets the special meaning that no digest must be used with this algorithm.

crypto/evp/keymgmt_lib.c
```diff
@@ -412,3 +412,51 @@ void *evp_keymgmt_util_gen(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+    OSSL_PARAM params[3];
+    char mddefault[100] = "";
+    char mdmandatory[100] = "";
+    char *result = NULL;
+    int rv = -2;
+
+    params[0] =
+        OSSL_PARAM_construct_utf8_string(OSSL_PKEY_PARAM_DEFAULT_DIGEST,
+                                         mddefault, sizeof(mddefault));
+    params[0].return_size = sizeof(mddefault) + 1;
```
paulidale, line 435: Setting the return size here will have no effect.  The return size is set by the recipient.

This isn't a great solution to the _did the other end set this_ question that we've thus far not been able to answered.

crypto/evp/keymgmt_lib.c
```diff
@@ -412,3 +412,51 @@ void *evp_keymgmt_util_gen(EVP_PKEY *target, EVP_KEYMGMT *keymgmt,
...
+    params[1] =
+        OSSL_PARAM_construct_utf8_string(OSSL_PKEY_PARAM_MANDATORY_DIGEST,
+                                         mdmandatory,
+                                         sizeof(mdmandatory));
+    params[1].return_size = sizeof(mdmandatory) + 1;
```
paulidale, line 440: Likewise.

doc/man3/EVP_PKEY_get_default_digest_nid.pod
```diff
@@ -18,7 +18,8 @@ EVP_PKEY_get_default_digest_nid, EVP_PKEY_get_default_digest_name
 EVP_PKEY_get_default_digest_name() fills in the default message digest
 name for the public key signature operations associated with key
 I<pkey> into I<mdname>, up to at most I<mdname_sz> bytes including the
-ending NUL byte.
+ending NUL byte.  The name may turn out to be C<"UNDEF">, signifying
```
paulidale, line 21: How about:

> The name could be C<"UNDEF"> signifying

providers/implementations/keymgmt/ecx_kmgmt.c
```diff
@@ -222,13 +233,15 @@ static int x448_get_params(void *key, OSSL_PARAM params[])
...
+                          ED25519_KEYLEN)
+        && ed_get_params(key, params);
 }
 
 static int ed448_get_params(void *key, OSSL_PARAM params[])
 {
     return ecx_get_params(key, params, ED448_BITS, ED448_SECURITY_BITS,
-                          ED448_KEYLEN);
+                          ED448_KEYLEN)
+        && ed_get_params(key, params);
```
mattcaswell, line 244: I think you will also need to adjust gettable_params

crypto/evp/keymgmt_lib.c
levitte: > Setting the return size here will have no effect. The return size is set by the recipient.

... *if* the recipient supports this parameter, and in that case, it will be set to something other than this initial value.

providers/implementations/keymgmt/ecx_kmgmt.c
levitte: Ah, correct

levitte: Done

doc/man3/EVP_PKEY_get_default_digest_nid.pod
levitte: Done

crypto/evp/keymgmt_lib.c
paulidale: I noted this, I don't like this way of detecting a parameter being set.

It's error prone -- the return size is set but the buffer not if the return size is larger than the data length.  So a legitimate return of size `sizeof(mddefault) + 1` could be a problem -- it isn't here because of another check but I think it is a bad precedent to set.

I'd consider using `(size_t)-1` would be better.  Best would be for the helper functions to support what is required.  The construct calls clearing whatever sentinel is used, the set calls flag it and an additional function to check.

levitte: > So a legitimate return of size `sizeof(mddefault) + 1` could be a problem

If `p.data_size` is `sizeof(mddefault)`, I don't see how that would be a legitimate return size...  that would mean that the receiving end has overwritten the buffer.

> I'd consider using `(size_t)-1` would be better.

I've considered that, but couldn't shake the feeling that there may be platforms where that's a perfectly legitimate size that doesn't necessarily eat all available memory (i.e. where `size_t` is smaller in size than `void *`).  It's *unlikely* that we'll hit that kind of size, at least for now, but...

> Best would be for the helper functions to support what is required.  The construct calls clearing whatever sentinel is used, the set calls flag it and an additional function to check.

I have nothing against that idea.

levitte:
> Do we need a test for this?

That should be a generalized test.  Not in this PR

crypto/evp/keymgmt_lib.c
paulidale: > If p.data_size is sizeof(mddefault), I don't see how that would be a legitimate return size...

If the return size is larger than the buffer size, the return size is set but no data is returned.  The function also returns 0, so it's not an issue here.

I'll try make a stab at an API tomorrow.  No promises.

levitte: Don't make this a blocker for the upcoming alpha.

t8m:
The Travis breakage is relevant.

kaduk:
There seems to be some missing text in the commit message for "EVP: Fix EVP_Digest{Sign,Verify}Init() to handle no default digest" (and the pull request description), that cuts off after:

```
Finally, we make sure to detect if the provider implementation
supports the digest name parameters (default or mandatory), and
returns with error if not.  This is what the legacy portion of the
```

crypto/evp/keymgmt_lib.c
kaduk: Interestingly, a compliant C99 implementation is allowed to set `SIZE_MAX` as small as 65535.

levitte:
> The Travis breakage is relevant.

It turns out that EC keys came with no default digest...

levitte:
> There seems to be some missing text in the commit message for "EVP: Fix EVP_Digest{Sign,Verify}Init() to handle no default digest" (and the pull request description), that cuts off after:

Reworded

crypto/evp/keymgmt_lib.c
paulidale: I've used processors where even that would be way too large :)

paulidale: Kind of moot given the delay but refer #11588 for an alternative

levitte: I've come to terms with the `(size_t)-1` value.  It will probably be eons before we see a legitimate size get even close...

levitte:
The remaining Travis failures are not relevant here (one is fixed in #11573)

levitte:
Rebased.  The #11592 fix should result in a clean Travis run

levitte:
@paulidale, if you could remember the "approval: done" label, that would be quit helpful.

paulidale:
Oops, sorry.

mattcaswell:
Pushed!

PR #11578, opened 2020-04-20
Fix no-des

mattcaswell:
Disable a recently added CMS test in the case of no-des.

### DIFF
--- a/test/recipes/80-test_cms.t
+++ b/test/recipes/80-test_cms.t
@@ -640,8 +640,8 @@ subtest "CMS Decrypt message encrypted with OpenSSL 1.1.1\n" => sub {
     plan tests => 1;
 
     SKIP: {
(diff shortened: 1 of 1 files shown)


test/recipes/80-test_cms.t
```diff
@@ -641,7 +641,7 @@ subtest "CMS Decrypt message encrypted with OpenSSL 1.1.1\n" => sub {
 
     SKIP: {
         skip "EC isn't supported in this build", 1
```
InfoHunter, line 643: Should you also changed this description to something like "EC or DES isn't supported..."

mattcaswell: Good point. Fixed.

mattcaswell:
Pushed. Thanks.

PR #11580, opened 2020-04-20
Fix a few more tests to use the new fipsinstall script

mattcaswell:
Commit ae6b654b6 introduced a new script to make running fipsinstall
easier. A few tests were added around the same time as that commit that
still do it the old way - so we fix those up.

##### Checklist


paulidale:
Assuming CIs agree

slontis:
Maybe this should be titled - 'break a few more tests on mac' :).

mattcaswell:
Closing due to #11592

PR #11581, opened 2020-04-20
Fix no-ec

mattcaswell:
This was missed by Travis because, although it has a no-ec build, the test
that failed only runs in a debug build. The Travis job with no-ec is not
a debug build and so the test was skipped.


mattcaswell:
The Travis red cross is unrelated to this PR.

mattcaswell:
Pushed. Thanks.

PR #11582, opened 2020-04-20
Fix no-dh and no-dsa

mattcaswell:
Add some missing OPENSSL_NO_DH guards.

Also, one of the sub-tests in the fipsinstall test corrupts a DH test to
confirm that fipsinstall fails. However that is never noticed in a
no-dh build - so we just skip that test in a no-dh build.


mattcaswell:
I also just added a commit to fix a closely related failure in no-dsa builds.

slontis:
Thats an impressive amount of travis timeouts.

Why were some of these not being found as 'unresolved' in travis builds?

mattcaswell:
Pushed. Thanks.

PR #11583, opened 2020-04-20
Fix no-deprecated tests where the GOST engine is present

mattcaswell:
The GOST test only runs if it detects that the GOST engine is present.
This is the case for the run-checker environment. The GOST engine uses
some deprecated functions, so we disable that test in a no-deprecated
build.


mattcaswell:
Pushed. Thanks.

PR #11584, opened 2020-04-20
Fix no-err

mattcaswell:
This fixes an assertion failure that can occur in the CMP code in the
event of a no-err build. The "improve_location_name" function assumed
that the fallback argument was always populated with something. However
in a no-err build this is not the case.


t8m:
The Travis failure is unrelated.

mattcaswell:
Pushed.

PR #11585, opened 2020-04-20
Fix test_cmp_vfy failures with FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION

DDvO:
Should fix issue reported today in #11386

- [x] tests are added or updated

PR #11586, opened 2020-04-20
Fix ssl_test_new and ssl_test_old when SSLv3 is enabled

mattcaswell:
The ssl_test_new and ssl_test_old tests are failing when SSLv3 is enabled.

There are a couple of fetches, that only occur in those codepaths and weren't using the correct libctx. Also we need to disable some tests in ssl_test_old when using the FIPS provider.

Marking this as urgent because it is one of the causes for the current travis failures (other fixes for travis failures in #11585 and #11573.

I've marked this with "extended tests". I do expect some of those tests to fail because of the other issues - but hopefully it won't fail due to ssl_test_new/ssl_test_old.

### DIFF
--- a/test/recipes/80-test_ssl_old.t
+++ b/test/recipes/80-test_ssl_old.t
@@ -359,7 +359,7 @@ sub testssl {
 
       SKIP: {
 	  skip "SSLv3 is not supported by this OpenSSL build", 4
-	      if disabled("ssl3");
+	      if disabled("ssl3") || $provider eq "fips";
 
 	  ok(run(test([@ssltest, "-bio_pair", "-ssl3"])),
 	     'test sslv3 via BIO pair');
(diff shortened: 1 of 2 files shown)


test/recipes/80-test_ssl_old.t
```diff
@@ -359,7 +359,7 @@ sub testssl {
 
       SKIP: {
 	  skip "SSLv3 is not supported by this OpenSSL build", 4
-	      if disabled("ssl3");
+	      if disabled("ssl3") || $provider eq "fips";
```
t8m, line 362: A cosmetic issue - perhaps there should be a separate skip statement? "SSLv3 is not supported by fips provider" ?

mattcaswell: Good point. Fixed.

mattcaswell:
> Approved if the SSLv3 related test failures are fixed and no new ones are introduced.

Although some builds failed due to unrelated issues, there aren't any TLS test failures in so I will push this now.

mattcaswell:
Pushed. Thanks.

PR #11588, opened 2020-04-21
param: Param add a modified indicator to OSSL_PARAM.

paulidale:
Using the _return_size_ field as a sentinel for change detection.

### DIFF
--- a/crypto/params.c
+++ b/crypto/params.c
@@ -36,8 +36,19 @@ static OSSL_PARAM ossl_param_construct(const char *key, unsigned int data_type,
     res.data_type = data_type;
     res.data = data;
     res.data_size = data_size;
-    res.return_size = 0;
+    res.return_size = OSSL_PARAM_UNMODIFIED;
     return res;
+}
+
+int OSSL_PARAM_modified(const OSSL_PARAM *p)
+{
+    return p != NULL && p->return_size != OSSL_PARAM_UNMODIFIED;
+}
+
+void OSSL_PARAM_set_unmodified(OSSL_PARAM *p)
+{
+    if (p != NULL)
+        p->return_size = OSSL_PARAM_UNMODIFIED;
 }
 
 int OSSL_PARAM_get_int(const OSSL_PARAM *p, int *val)
--- a/doc/man3/OSSL_PARAM_int.pod
+++ b/doc/man3/OSSL_PARAM_int.pod
@@ -27,7 +27,7 @@ OSSL_PARAM_set_int64, OSSL_PARAM_set_long, OSSL_PARAM_set_size_t,
 OSSL_PARAM_set_uint, OSSL_PARAM_set_uint32, OSSL_PARAM_set_uint64,
 OSSL_PARAM_set_ulong, OSSL_PARAM_set_BN, OSSL_PARAM_set_utf8_string,
 OSSL_PARAM_set_octet_string, OSSL_PARAM_set_utf8_ptr,
-OSSL_PARAM_set_octet_ptr
+OSSL_PARAM_set_octet_ptr, OSSL_PARAM_UNMODIFIED
 - OSSL_PARAM helpers
 
 =head1 SYNOPSIS
@@ -51,6 +51,8 @@ OSSL_PARAM_set_octet_ptr
  #define OSSL_PARAM_utf8_ptr(key, address, size)
  #define OSSL_PARAM_octet_ptr(key, address, size)
  #define OSSL_PARAM_END
+
+ #define OSSL_PARAM_UNMODIFIED
 
  OSSL_PARAM OSSL_PARAM_construct_TYPE(const char *key, TYPE *buf);
  OSSL_PARAM OSSL_PARAM_construct_BN(const char *key, unsigned char *buf,
@@ -90,6 +92,9 @@ OSSL_PARAM_set_octet_ptr
                               size_t *used_len);
  int OSSL_PARAM_set_octet_ptr(OSSL_PARAM *p, const void *val,
                               size_t used_len);
+
+ int OSSL_PARAM_modified(const OSSL_PARAM *p);
+ void OSSL_PARAM_set_unmodified(OSSL_PARAM *p);
 
 =head1 DESCRIPTION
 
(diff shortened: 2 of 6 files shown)


doc/man3/OSSL_PARAM_int.pod
```diff
@@ -27,7 +27,7 @@ OSSL_PARAM_set_int64, OSSL_PARAM_set_long, OSSL_PARAM_set_size_t,
 OSSL_PARAM_set_uint, OSSL_PARAM_set_uint32, OSSL_PARAM_set_uint64,
 OSSL_PARAM_set_ulong, OSSL_PARAM_set_BN, OSSL_PARAM_set_utf8_string,
 OSSL_PARAM_set_octet_string, OSSL_PARAM_set_utf8_ptr,
-OSSL_PARAM_set_octet_ptr
+OSSL_PARAM_set_octet_ptr, OSSL_PARAM_USED
```
levitte, line 30: Looks misspellt.

paulidale: Worse, that's an older name I started with.  Fix is in.

paulidale:
I wonder if it would be better to not document how the modification detection works.

doc/man3/OSSL_PARAM_int.pod
```diff
@@ -261,6 +276,8 @@ OSSL_PARAM_locate() and OSSL_PARAM_locate_const() return a pointer to
 the matching OSSL_PARAM object.  They return B<NULL> on error or when
 no object matching B<key> exists in the B<array>.
 
+OSSL_PARAM_used() returns B<1> if the parameter was set and B<0> otherwise.
```
levitte, line 279: Looks misspellt...

levitte:
I'm wondering if you shouldn't mention what the value of `OSSL_PARAM_UNMODIFIED` is, and declare it "reserved".  If nothing else, those who worry that their particular huge buffer size might have the same value will feel informed.

doc/man3/OSSL_PARAM_int.pod
paulidale: I need a break I think...

levitte:
> I wonder if it would be better to not document how the modification detection works.

Why?

paulidale:
> Why?

To give us some flexibility to change it without incurring breakage.  I can't guarantee that it will never need changing.  By not spelling it out, we can change it because the helper functions and macros hide the details.

romen:
> > Why?
> 
> To give us some flexibility to change it without incurring breakage. I can't guarantee that it will never need changing. By not spelling it out, we can change it because the helper functions and macros hide the details.

Would it be more future proof to have a separate flag variable to track the state of an OSSL_PARAM?

paulidale:
> Would it be more future proof to have a separate flag variable to track the state of an OSSL_PARAM?

Yes, but it was shot down in the past.

Stealing a bit from the type field would be another possibility (although this would be far too invasive to consider now).

OSSL_PARAM is a public structure and we want to keep it plain and simple.  A version number and some flags wouldn't go astray but...

levitte:
> OSSL_PARAM is a public structure 

... and that's exactly the reason to document the unmodified trick

paulidale:
> Should this be also cleaning up where we already use the return_size in this fashion. There are quite a few of them.

We should clean them up.  I'd prefer to get this in and to then address the other uses.  I didn't release that such uses were widespread.

levitte:
> I didn't release that such uses were widespread.

They aren't yet.

```
: ; git grep 'return_size =.*not_set'
crypto/evp/p_lib.c:    params[0].return_size = not_set;
crypto/evp/p_lib.c:        if (params[0].return_size == not_set
crypto/evp/p_lib.c:    if (params[0].return_size == not_set)
crypto/evp/p_lib.c:    params[0].return_size = not_set;
crypto/evp/p_lib.c:    if (params[0].return_size == not_set)
crypto/evp/p_lib.c:    params[0].return_size = not_set;
crypto/evp/p_lib.c:    if (params[0].return_size == not_set)
crypto/evp/p_lib.c:    params[0].return_size = not_set;
crypto/evp/p_lib.c:    if (params[0].return_size == not_set)
crypto/evp/p_lib.c:    params[0].return_size = not_set;
crypto/evp/p_lib.c:    if (params[0].return_size == not_set)
```

crypto/params.c
```diff
@@ -36,10 +36,21 @@ static OSSL_PARAM ossl_param_construct(const char *key, unsigned int data_type,
...
 }
 
+int OSSL_PARAM_modified(const OSSL_PARAM *p)
+{
+    return p != NULL && p->return_size != OSSL_PARAM_UNMODIFIED;
+}
+
+void OSSL_PARAM_set_unmodified(OSSL_PARAM *p)
+{
+    if (p != NULL)
```
richsalz, line 50: dislike the NULL checks here and above FWIW.

richsalz:
Since the structure is public, and since the define's are public, and since openssl considers the contents of the header files, and not just the documentation of them, to be part of the API/ABI, might as well document what things currently do.

paulidale:
Merged to master, thanks.

kaduk:
> since openssl considers the contents of the header files, and not just the documentation of them, to be part of the API/ABI

I thought I remembered the stance being a bit more subtle than that; do you have a pointer handy ("no" is a fine answer)?

richsalz:
Sorry, I don't.  Perhaps @t-j-h might reply here, since he was an advocate of that view.

levitte:
Generally speaking, we consider the public headers to define the API. However, I would argue that some of those headers or APIs have become public by accident, so it's possible to justify that some of them should stop (see parts of the CONF API that was recently made unpublic).

OSSL_PARAM is intentionally a public structure, however, and very much part of the libcrypto<->provider interface, so it's safe to say that it should be considered part of the public API.

kroeckx:
> OSSL_PARAM is intentionally a public structure, however, and very much part of the libcrypto<->provider interface, so it's safe to say that it should be considered part of the public API. 

It's not because it's used for the interface between libcrypto and
the providers, that it should also be used in the interface
between the application and the libcrypto.

PR #11589, opened 2020-04-21
Support Linux Kernel TLS for TLSv1.3 and new ciphers

junjunk:
Linux kernel 5.3 adds support for TLS 1.3 and AES_GCM256 and AES_CCM128 ciphers in Kernel TLS mode. This pull request adds support of that features in OpenSSL.

- [x ] tests are added or updated

This is version 2 of patchset. Previous PR #11136
Changes:
- Do not redefine kernel structures, use only defined for current kernel
- Rebase on top of current master

### DIFF
--- a/doc/man3/SSL_CTX_set_record_padding_callback.pod
+++ b/doc/man3/SSL_CTX_set_record_padding_callback.pod
@@ -16,7 +16,7 @@ SSL_set_block_padding - install callback to specify TLS 1.3 record padding
  #include <openssl/ssl.h>
 
  void SSL_CTX_set_record_padding_callback(SSL_CTX *ctx, size_t (*cb)(SSL *s, int type, size_t len, void *arg));
- void SSL_set_record_padding_callback(SSL *ssl, size_t (*cb)(SSL *s, int type, size_t len, void *arg));
+ int SSL_set_record_padding_callback(SSL *ssl, size_t (*cb)(SSL *s, int type, size_t len, void *arg));
 
  void SSL_CTX_set_record_padding_callback_arg(SSL_CTX *ctx, void *arg);
  void *SSL_CTX_get_record_padding_callback_arg(const SSL_CTX *ctx);
@@ -32,6 +32,8 @@ SSL_set_block_padding - install callback to specify TLS 1.3 record padding
 SSL_CTX_set_record_padding_callback() or SSL_set_record_padding_callback()
 can be used to assign a callback function I<cb> to specify the padding
 for TLS 1.3 records. The value set in B<ctx> is copied to a new SSL by SSL_new().
+Kernel TLS is not possible if the record padding callback is set, and the callback
+function cannot be set if Kernel TLS is already configured for the current SSL object.
 
 SSL_CTX_set_record_padding_callback_arg() and SSL_set_record_padding_callback_arg()
 assign a value B<arg> that is passed to the callback when it is invoked. The value
@@ -64,6 +66,9 @@ indicates no padding will be added. A return value that causes the record to
 exceed the maximum record size (SSL3_RT_MAX_PLAIN_LENGTH) will pad out to the
 maximum record size.
 
+The SSL_CTX_get_record_padding_callback_arg() function returns 1 on success or 0 if
+the callback function is not set because Kernel TLS is configured for the SSL object.
+
 =head1 NOTES
 
 The default behavior is to add no padding to the record.
@@ -83,6 +88,9 @@ L<ssl(7)>, L<SSL_new(3)>
 =head1 HISTORY
 
 The record padding API was added for TLS 1.3 support in OpenSSL 1.1.1.
+
+The return type of SSL_CTX_set_record_padding_callback() function was
+changed to int in OpenSSL 3.0.
 
 =head1 COPYRIGHT
 
--- a/include/internal/ktls.h
+++ b/include/internal/ktls.h
@@ -309,6 +309,103 @@ static ossl_inline int ktls_read_record(int fd, void *data, size_t length)
 }
 
 #   endif /* OPENSSL_NO_KTLS_RX */
+
+/* Function to check supported ciphers in Linux */
+static ossl_inline int ktls_check_supported_cipher(const EVP_CIPHER *c,
+                                            const EVP_CIPHER_CTX *dd)
+{
+    /* check that cipher is AES_GCM_128, AES_GCM_256, AES_CCM_128 */
+    switch (EVP_CIPHER_nid(c))
+    {
+#  ifdef OPENSSL_KTLS_AES_CCM_128
+    case NID_aes_128_ccm:
+        if (EVP_CIPHER_CTX_tag_length(dd) != EVP_CCM_TLS_TAG_LEN)
+          return 0;
+#  endif
+#  ifdef OPENSSL_KTLS_AES_GCM_128
+    case NID_aes_128_gcm:
+#  endif
+#  ifdef OPENSSL_KTLS_AES_GCM_256
+    case NID_aes_256_gcm:
+#  endif
+        return 1;
+    default:
+        return 0;
+    }
+}
+
+/* Function to configure kernel TLS structure */
+static ossl_inline int ktls_configure_crypto(const EVP_CIPHER *c, int tls_version,
+                                EVP_CIPHER_CTX *dd, void *rl_sequence,
+                                struct tls_crypto_info_all *crypto_info,
+                                unsigned char **rec_seq, unsigned char *iv,
+                                unsigned char *key)
+{
+    unsigned char geniv[12];
+    unsigned char *iiv = iv;
+
+    if (tls_version == TLS1_2_VERSION &&
+        EVP_CIPHER_mode(c) == EVP_CIPH_GCM_MODE) {
+        EVP_CIPHER_CTX_ctrl(dd, EVP_CTRL_GET_IV,
+                            EVP_GCM_TLS_FIXED_IV_LEN + EVP_GCM_TLS_EXPLICIT_IV_LEN,
+                            geniv);
+        iiv = geniv;
+    }
+
+    memset(crypto_info, 0, sizeof(*crypto_info));
+    switch (EVP_CIPHER_nid(c))
+    {
+#  ifdef OPENSSL_KTLS_AES_GCM_128
+    case NID_aes_128_gcm:
+        crypto_info->gcm128.info.cipher_type = TLS_CIPHER_AES_GCM_128;
+        crypto_info->gcm128.info.version = tls_version;
+        crypto_info->tls_crypto_info_len = sizeof(crypto_info->gcm128);
+        memcpy(crypto_info->gcm128.iv, iiv + EVP_GCM_TLS_FIXED_IV_LEN,
+                TLS_CIPHER_AES_GCM_128_IV_SIZE);
+        memcpy(crypto_info->gcm128.salt, iiv, TLS_CIPHER_AES_GCM_128_SALT_SIZE);
+        memcpy(crypto_info->gcm128.key, key, EVP_CIPHER_key_length(c));
+        memcpy(crypto_info->gcm128.rec_seq, rl_sequence,
+                TLS_CIPHER_AES_GCM_128_REC_SEQ_SIZE);
+        if (rec_seq != NULL)
+            *rec_seq = crypto_info->gcm128.rec_seq;
+        return 1;
+#  endif
+#  ifdef OPENSSL_KTLS_AES_GCM_256
+    case NID_aes_256_gcm:
+        crypto_info->gcm256.info.cipher_type = TLS_CIPHER_AES_GCM_256;
+        crypto_info->gcm256.info.version = tls_version;
+        crypto_info->tls_crypto_info_len = sizeof(crypto_info->gcm256);
+        memcpy(crypto_info->gcm256.iv, iiv + EVP_GCM_TLS_FIXED_IV_LEN,
+                TLS_CIPHER_AES_GCM_256_IV_SIZE);
+        memcpy(crypto_info->gcm256.salt, iiv, TLS_CIPHER_AES_GCM_256_SALT_SIZE);
+        memcpy(crypto_info->gcm256.key, key, EVP_CIPHER_key_length(c));
+        memcpy(crypto_info->gcm256.rec_seq, rl_sequence,
+                TLS_CIPHER_AES_GCM_256_REC_SEQ_SIZE);
+        if (rec_seq != NULL)
+            *rec_seq = crypto_info->gcm256.rec_seq;
+        return 1;
+#  endif
+#  ifdef OPENSSL_KTLS_AES_CCM_128
+    case NID_aes_128_ccm:
+        crypto_info->ccm128.info.cipher_type = TLS_CIPHER_AES_CCM_128;
+        crypto_info->ccm128.info.version = tls_version;
+        crypto_info->tls_crypto_info_len = sizeof(crypto_info->ccm128);
+        memcpy(crypto_info->ccm128.iv, iiv + EVP_CCM_TLS_FIXED_IV_LEN,
+                TLS_CIPHER_AES_CCM_128_IV_SIZE);
+        memcpy(crypto_info->ccm128.salt, iiv, TLS_CIPHER_AES_CCM_128_SALT_SIZE);
+        memcpy(crypto_info->ccm128.key, key, EVP_CIPHER_key_length(c));
+        memcpy(crypto_info->ccm128.rec_seq, rl_sequence,
+                TLS_CIPHER_AES_CCM_128_REC_SEQ_SIZE);
+        if (rec_seq != NULL)
+            *rec_seq = crypto_info->ccm128.rec_seq;
+        return 1;
+#  endif
+    default:
+        return 0;
+    }
+
+}
+
 #  endif /* OPENSSL_SYS_LINUX */
 # endif /* HEADER_INTERNAL_KTLS */
 #else /* defined(OPENSSL_NO_KTLS) */
--- a/include/internal/ktls.h
+++ b/include/internal/ktls.h
@@ -7,6 +7,19 @@
  * https://www.openssl.org/source/license.html
  */
 
+#if defined(OPENSSL_SYS_LINUX)
+# ifndef OPENSSL_NO_KTLS
+#  include <linux/version.h>
+#  if LINUX_VERSION_CODE < KERNEL_VERSION(4, 13, 0)
+#   define OPENSSL_NO_KTLS
+#   ifndef PEDANTIC
+#    warning "KTLS requires Kernel Headers >= 4.13.0"
+#    warning "Skipping Compilation of KTLS"
+#   endif
+#  endif
+# endif
+#endif
+
 #ifndef OPENSSL_NO_KTLS
 # ifndef HEADER_INTERNAL_KTLS
 #  define HEADER_INTERNAL_KTLS
@@ -109,93 +122,57 @@ static ossl_inline ossl_ssize_t ktls_sendfile(int s, int fd, off_t off,
 #  endif                         /* __FreeBSD__ */
 
 #  if defined(OPENSSL_SYS_LINUX)
-#   include <linux/version.h>
 
-#   define K_MAJ   4
-#   define K_MIN1  13
-#   define K_MIN2  0
-#   if LINUX_VERSION_CODE < KERNEL_VERSION(K_MAJ, K_MIN1, K_MIN2)
-
+#   include <linux/tls.h>
+#   if LINUX_VERSION_CODE < KERNEL_VERSION(4, 17, 0)
+#    define OPENSSL_NO_KTLS_RX
 #    ifndef PEDANTIC
-#     warning "KTLS requires Kernel Headers >= 4.13.0"
-#     warning "Skipping Compilation of KTLS"
+#     warning "KTLS requires Kernel Headers >= 4.17.0 for receiving"
+#     warning "Skipping Compilation of KTLS receive data path"
 #    endif
+#   endif
+#   define OPENSSL_KTLS_AES_GCM_128
+#   if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 1, 0)
+#    define OPENSSL_KTLS_AES_GCM_256
+#    define OPENSSL_KTLS_TLS13
+#    if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 2, 0)
+#     define OPENSSL_KTLS_AES_CCM_128
+#    endif
+#   endif
 
-#    define TLS_TX                  1
+#   include <sys/sendfile.h>
+#   include <netinet/tcp.h>
+#   include <linux/socket.h>
+#   include "openssl/ssl3.h"
+#   include "openssl/tls1.h"
+#   include "openssl/evp.h"
+
+#   ifndef SOL_TLS
+#    define SOL_TLS 282
+#   endif
+
+#   ifndef TCP_ULP
+#    define TCP_ULP 31
+#   endif
+
+#   ifndef TLS_RX
 #    define TLS_RX                  2
+#   endif
 
-#    define TLS_CIPHER_AES_GCM_128                          51
-#    define TLS_CIPHER_AES_GCM_128_IV_SIZE                  8
-#    define TLS_CIPHER_AES_GCM_128_KEY_SIZE                 16
-#    define TLS_CIPHER_AES_GCM_128_SALT_SIZE                4
-#    define TLS_CIPHER_AES_GCM_128_TAG_SIZE                 16
-#    define TLS_CIPHER_AES_GCM_128_REC_SEQ_SIZE             8
-
-#    define TLS_SET_RECORD_TYPE     1
-
-struct tls_crypto_info {
-    unsigned short version;
-    unsigned short cipher_type;
+struct tls_crypto_info_all {
+    union {
+#   ifdef OPENSSL_KTLS_AES_GCM_128
+        struct tls12_crypto_info_aes_gcm_128 gcm128;
+#   endif
+#   ifdef OPENSSL_KTLS_AES_GCM_256
+        struct tls12_crypto_info_aes_gcm_256 gcm256;
+#   endif
+#   ifdef OPENSSL_KTLS_AES_CCM_128
+        struct tls12_crypto_info_aes_ccm_128 ccm128;
+#   endif
+    };
+    size_t tls_crypto_info_len;
 };
-
-struct tls12_crypto_info_aes_gcm_128 {
-    struct tls_crypto_info info;
-    unsigned char iv[TLS_CIPHER_AES_GCM_128_IV_SIZE];
-    unsigned char key[TLS_CIPHER_AES_GCM_128_KEY_SIZE];
-    unsigned char salt[TLS_CIPHER_AES_GCM_128_SALT_SIZE];
-    unsigned char rec_seq[TLS_CIPHER_AES_GCM_128_REC_SEQ_SIZE];
-};
-
-/* Dummy functions here */
-static ossl_inline int ktls_enable(int fd)
-{
-    return 0;
-}
-
-static ossl_inline int ktls_start(int fd, void *crypto_info, 
-                                  size_t len, int is_tx)
-{
-    return 0;
-}
-
-static ossl_inline int ktls_send_ctrl_message(int fd, unsigned char record_type,
-                                              const void *data, size_t length)
-{
-    return -1;
-}
-
-static ossl_inline int ktls_read_record(int fd, void *data, size_t length)
-{
-    return -1;
-}
-
-static ossl_inline ossl_ssize_t ktls_sendfile(int s, int fd, off_t off, size_t size, int flags)
-{
-    return -1;
-}
-
-#   else                        /* KERNEL_VERSION */
-
-#    include <sys/sendfile.h>
-#    include <netinet/tcp.h>
-#    include <linux/tls.h>
-#    include <linux/socket.h>
-#    include "openssl/ssl3.h"
-#    include "openssl/tls1.h"
-#    include "openssl/evp.h"
-
-#    ifndef SOL_TLS
-#     define SOL_TLS 282
-#    endif
-
-#    ifndef TCP_ULP
-#     define TCP_ULP 31
-#    endif
-
-#    ifndef TLS_RX
-#     define TLS_RX                  2
-#    endif
-
 /*
  * When successful, this socket option doesn't change the behaviour of the
  * TCP socket, except changing the TCP setsockopt handler to enable the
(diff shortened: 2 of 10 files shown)


junjunk:
> for me _OPENSSL_cleanse(key, sizeof(key))_ is better than _OPENSSL_cleanse(key, EVP_MAX_KEY_LENGTH)_. In future the declaration of key can change, and if this place is forgotten to update, thus produce a bug or a segfault

It seems to me that _OPENSSL_cleanse(key, sizeof(key))_ is not possible in _derive_secret_key_and_iv_ because _sizeof(key)_ is _sizeof(unsigned char *)_
That's why it should be caller response to clean the key.
From security point of view in case of any error in _tls13_change_cipher_state_ key must be cleaned as well as _secret_.
Thanks for pointing this out, I will update and rebase it soon.

junjunk:
> The same old was better

Do you mean _old_ - previous Pull Request?

NMorozxov:
I missed some place to comment. I mean than there is many places when you replace sizeof() to fixed size
> > The same old was better
> 
> Do you mean _old_ - previous Pull Request?

junjunk:
> I missed some place to comment. I mean than there is many places when you replace sizeof() to fixed size
> 
> > > The same old was better
> > 
> > 
> > Do you mean _old_ - previous Pull Request?

Fixed the one in _derive_secret_key_and_iv_ and didn't find any other in my patch set. Could you please suggest the places to fix?

NMorozxov:
All is fixed
> Fixed the one in _derive_secret_key_and_iv_ and didn't find any other in my patch set. Could you please suggest the places to fix?

junjunk:
@kroeckx Can you please review this one - I have changed it based on your suggestions from #11136

ssl/tls13_enc.c
```diff
@@ -764,12 +772,132 @@ int tls13_change_cipher_state(SSL *s, int which)
...
+skip_ktls:
+# endif
+#endif
     ret = 1;
  err:
     if ((which & SSL3_CC_EARLY) != 0) {
         /* We up-refed this so now we need to down ref */
         ssl_evp_cipher_free(cipher);
     }
+    OPENSSL_cleanse(key, sizeof(key));
```
bsdjhb, line 900: Moving the cleanse here to the function that allocates the key is the right thing to do.

junjunk:
>I'm also fine with rebasing #10626 on top of this if this is likely to get merged soon.

But will it get merged? I still can't see any tags or review results

ssl/tls13_enc.c
```diff
@@ -469,7 +471,6 @@ static int derive_secret_key_and_iv(SSL *s, int sending, const EVP_MD *md,
 
     return 1;
  err:
```
mattcaswell, line 473: Might as well remove all the "goto err" and just return 0.

ssl/tls13_enc.c
```diff
@@ -764,12 +772,132 @@ int tls13_change_cipher_state(SSL *s, int which)
...
+        || ((which & SSL3_CC_WRITE) && (s->mode & SSL_MODE_NO_KTLS_TX)))
+        goto skip_ktls;
+
+    /* ktls supports only the maximum fragment size */
+    if (ssl_get_max_send_fragment(s) != SSL3_RT_MAX_PLAIN_LENGTH)
+        goto skip_ktls;
+
+    /* check whether cipher is known */
+    if (cipher == NULL)
+        goto skip_ktls;
```
mattcaswell, line 788: `cipher` shouldn't ever be NULL here I think. We already de-ref'd it in the `derive_secret_key_and_iv` call.

ssl/tls13_enc.c
```diff
@@ -764,12 +772,132 @@ int tls13_change_cipher_state(SSL *s, int which)
...
+
+    /* check that cipher is AES_GCM_128, AES_GCM_256, AES_CCM_128 */
+    switch (EVP_CIPHER_nid(cipher))
+    {
+#   ifdef OPENSSL_KTLS_AES_GCM_128
+    case NID_aes_128_gcm:
+        if (EVP_CIPHER_mode(cipher) != EVP_CIPH_GCM_MODE ||
+            EVP_CIPHER_key_length(cipher) != TLS_CIPHER_AES_GCM_128_KEY_SIZE)
```
mattcaswell, line 796: Again I'm not sure why we need more than just checking the nid.

ssl/tls13_enc.c
```diff
@@ -764,12 +772,132 @@ int tls13_change_cipher_state(SSL *s, int which)
         s->statem.enc_write_state = ENC_WRITE_STATE_WRITE_PLAIN_ALERTS;
     else
         s->statem.enc_write_state = ENC_WRITE_STATE_VALID;
+#ifndef OPENSSL_NO_KTLS
+# if defined(OPENSSL_KTLS_TLS13)
+#  ifndef __FreeBSD__
+    if (!(which & SSL3_CC_WRITE)
```
mattcaswell, line 778: So we only support transmit for TLSv1.3?

ssl/tls13_enc.c
```diff
@@ -764,12 +772,132 @@ int tls13_change_cipher_state(SSL *s, int which)
...
+                 ERR_R_INTERNAL_ERROR);
+        goto err;
+    }
+
+    /* All future data will get encrypted by ktls. Flush the BIO or skip ktls */
+    if (BIO_flush(bio) <= 0)
+        goto skip_ktls;
+
+    /* ktls doesn't support renegotiation */
+    if (BIO_get_ktls_send(bio)) {
```
mattcaswell, line 832: I'm trying to reconcile the comment, the code, and the fact that TLSv1.3 doesn't support renegotiation - and I'm failing.

ssl/tls13_enc.c
```diff
@@ -764,12 +772,132 @@ int tls13_change_cipher_state(SSL *s, int which)
...
+            TLS_CIPHER_AES_CCM_128_IV_SIZE);
+        memcpy(crypto_info.ccm128.salt, iv, TLS_CIPHER_AES_GCM_128_SALT_SIZE);
+        memcpy(crypto_info.ccm128.key, key, EVP_CIPHER_key_length(cipher));
+        memcpy(crypto_info.ccm128.rec_seq, &s->rlayer.write_sequence,
+            TLS_CIPHER_AES_GCM_128_REC_SEQ_SIZE);
+        break;
+#   endif
+    default:
+        goto skip_ktls;
+    }
```
mattcaswell, line 883: This looks very similar to the equivalent TLSv1.2 block of code. Is there some opportunity to pull this out into a function that can be shared - rather than all the code duplication?

ssl/tls13_enc.c
```diff
@@ -764,12 +772,132 @@ int tls13_change_cipher_state(SSL *s, int which)
...
+
+    /* ktls works with user provided buffers directly */
+    if (BIO_set_ktls(bio, &crypto_info, which & SSL3_CC_WRITE)) {
+        ssl3_release_write_buffer(s);
+        SSL_set_options(s, SSL_OP_NO_RENEGOTIATION);
```
mattcaswell, line 888: Irrelevant in TLSv1.3

test/sslapitest.c
```diff
@@ -1164,85 +1157,409 @@ static int test_ktls_sendfile(void)
...
+#ifndef OPENSSL_KTLS_AES_GCM_128
+   if(tst == 1) return 1;
+#endif
+#ifndef OPENSSL_KTLS_AES_CCM_128
+   if(tst == 2) return 1;
+#endif
+#ifndef OPENSSL_KTLS_AES_GCM_256
+   if(tst == 3) return 1;
+#endif
+   return test_ktls_sendfile(TLS1_2_VERSION,cipher[tst]);
```
mattcaswell, line 1501: Nit: missing space after ,

test/sslapitest.c
```diff
@@ -1164,85 +1157,409 @@ static int test_ktls_sendfile(void)
...
+    case 11:
+       return test_ktls_client_no_tx_server(TLS1_3_VERSION);
+    case 12:
+       return test_ktls_no_txrx_client_server(TLS1_3_VERSION);
+    case 13:
+       return test_ktls_no_rx_client_server(TLS1_3_VERSION);
+    case 14:
+       return test_ktls_no_tx_client_server(TLS1_3_VERSION);
+    case 15:
+       return test_ktls_client_server(TLS1_3_VERSION);
```
mattcaswell, line 1541: test_ktls_tls13 seems to be identical to test_ktls_tls12 except for the parameter sent to all of the functions. Surely there is come opportunity to rationalise this by merging the two functions.

include/internal/ktls.h
```diff
@@ -338,8 +308,35 @@ static ossl_inline int ktls_read_record(int fd, void *data, size_t length)
     return ret;
 }
 
-#    endif
 #   endif
 #  endif
 # endif
+#else
```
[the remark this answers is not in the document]
junjunk: Ok, I've added some comments

ssl/t1_enc.c
```diff
@@ -432,13 +437,36 @@ int tls1_change_cipher_state(SSL *s, int which)
...
-    if (EVP_CIPHER_nid(c) != NID_aes_128_gcm
-        || EVP_CIPHER_mode(c) != EVP_CIPH_GCM_MODE
-        || EVP_CIPHER_key_length(c) != TLS_CIPHER_AES_GCM_128_KEY_SIZE)
+    /* check that cipher is AES_GCM_128, AES_GCM_256, AES_CCM_128 */
+    switch (EVP_CIPHER_nid(c))
+    {
+#  ifdef OPENSSL_KTLS_AES_GCM_128
+    case NID_aes_128_gcm:
+        if (EVP_CIPHER_mode(c) != EVP_CIPH_GCM_MODE ||
+            EVP_CIPHER_key_length(c) != TLS_CIPHER_AES_GCM_128_KEY_SIZE)
```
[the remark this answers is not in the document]
junjunk: Do not exactly know about mode checks, but key length checks are made to be sure that we will not overwrite any memory later in memcpy(). This checks were cloned from previous code of kernel TLS support. Additional check of tag len in CCM mode is needed because NID_aes_128_ccm is used for bith AES128-CCM and AES128-CCM8 but latter is not supported in kernel.

ssl/tls13_enc.c
junjunk: Should this check be moved earlier, just before `derive_secret_key_and_iv` or we assume that NULL value of `cipher` is only possible in error path?

ssl/tls13_enc.c
junjunk: fixed

ssl/tls13_enc.c
junjunk: Yes, right now I'm not sure how to correctly implement RX path for TLSv1.3 because it have a lot of dependencies used in TLSv1.2 code. I'll try it later in another PR.

ssl/tls13_enc.c
junjunk: That check means that if `BIO_CTRL_GET_KTLS_SEND` flag is already set in BIO then `tls13_change_cipher_state` is called in case of renegotiation but not in initial handshake and that is not possible in Kernel TLS mode. The same check exists in TLSv1.2 code:
`    /* ktls doesn't support renegotiation */
    if ((BIO_get_ktls_send(s->wbio) && (which & SSL3_CC_WRITE)) ||
        (BIO_get_ktls_recv(s->rbio) && (which & SSL3_CC_READ))) {
        SSLfatal(s, SSL_AD_NO_RENEGOTIATION, SSL_F_TLS1_CHANGE_CIPHER_STATE,
                 ERR_R_INTERNAL_ERROR);
        goto err;
    }
`
It's a bit comples because of RX and TX path checks, in TLSv1.3 with support of only TX path the check is simplified.

ssl/tls13_enc.c
junjunk: Yes, you are right. Will make it as a function in `ktls.h`

ssl/tls13_enc.c
junjunk: Sure, will fix it

ssl/t1_enc.c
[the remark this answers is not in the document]
mattcaswell: But both mode and key length are given in the nid! If we get a NID of NID_aes_128_gcm and the EVP_CIPHER_mode() returns anything other than `EVP_CIPH_GCM_MODE` then something has gone fundamentally wrong. Similarly with the key length. This code looks redundant to me. I realise it was there in the original code - but now we're duplicating this same redundant code everywhere, which seems wrong.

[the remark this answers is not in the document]
mattcaswell: Tag length checks make sense

ssl/tls13_enc.c
mattcaswell: Either we just remove it, or we could add this just before `derive_secret_key_and_iv`:

````
    if(!ossl_assert(cipher != NULL))
        goto err;
````

File: test/sslapitest.c (hunk 9 above)
Comment (junjunk) [reply test/sslapitest.c]: fixed

File: test/sslapitest.c (hunk 10 above)
Comment (junjunk) [reply test/sslapitest.c]: Refactored. The same change done to test_ktls_sendfile_tls{12,13} functions.

File: ssl/tls13_enc.c (hunk 8 above)
Comment (junjunk) [reply ssl/tls13_enc.c]: removed `SSL_set_options(s, SSL_OP_NO_RENEGOTIATION);`

File: ssl/tls13_enc.c (hunk 7 above)
Comment (junjunk) [reply ssl/tls13_enc.c]: Made two functions - `ktls_check_supported_cipher` and `ktls_configure_crypto` and refactored code of TLSv1.2 and TLSv1.3 to use this functions. The only question right now is do we really need to call `EVP_CIPHER_CTX_ctrl(..., EVP_CTRL_GET_IV, ..)` in case of TLSv1.2 and GCM ciphers? This code was in initial support for Kernel TLS, but in the same case code for FreeBSD uses `iv` that was used in `EVP_CIPHER_CTX_ctrl(..,EVP_CTRL_GCM_SET_IV_FIXED,..)`

File: ssl/tls13_enc.c (hunk 6 above)
Comment (junjunk) [reply ssl/tls13_enc.c]: removed this check.

File: ssl/tls13_enc.c (hunk 4 above)
Comment (junjunk) [reply ssl/tls13_enc.c]: removed this check in `ktls_check_supported_cipher`

File: ssl/t1_enc.c (hunk 12 above)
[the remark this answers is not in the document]
Comment (junjunk) [reply ssl/t1_enc.c]: Redundant checks are removed and the hole check is moved to `ktls_check_supported_cipher` in `ktls.h`

File: ssl/tls13_enc.c (hunk 3 above)
Comment (junjunk) [reply ssl/tls13_enc.c]: Made this check as you suggested

File: ssl/tls13_enc.c (hunk 7 above)
Comment (junjunk) [reply ssl/tls13_enc.c]: @borisPis-zz Can you please explain this call `EVP_CIPHER_CTX_ctrl(..., EVP_CTRL_GET_IV, ..)` for Kernel TLS configuration ?

**Comment by junjunk** [comment]
@mattcaswell It seems to me that all requests are fixed, can you please look at this once more?

File: ssl/record/rec_layer_s3.c [hunk 13]
```diff
@@ -938,6 +938,7 @@ int do_ssl3_write(SSL *s, int type, const unsigned char *buf,
         }
 
         if (SSL_TREAT_AS_TLS13(s)
+                && !BIO_get_ktls_send(s->wbio)
```
mattcaswell, line 941: Probably we not go down the KTLS path if `record_padding_cb` is set. Also if the max_fragment_length extension is in use then ssl_get_max_send_fragment() could return something smaller than normal - so we probably shouldn't use KTLS in that scenario either (I wonder if we have considered this in the TLSv1.2 case?)

ssl/tls13_enc.c
```diff
@@ -764,12 +774,53 @@ int tls13_change_cipher_state(SSL *s, int which)
...
         s->statem.enc_write_state = ENC_WRITE_STATE_VALID;
+#ifndef OPENSSL_NO_KTLS
+# if defined(OPENSSL_KTLS_TLS13)
+#  ifndef __FreeBSD__
+    if (!(which & SSL3_CC_WRITE)
+        || ((which & SSL3_CC_WRITE) && (s->mode & SSL_MODE_NO_KTLS_TX)))
+        goto skip_ktls;
+
+    /* ktls supports only the maximum fragment size */
+    if (ssl_get_max_send_fragment(s) != SSL3_RT_MAX_PLAIN_LENGTH)
```
mattcaswell, line 785: Ah!! We never enter KTLS if max_fragment_length is in use. Probably we also need to consider the `record_padding_cb` here. Strictly speaking the `record_padding_cb` could be set at any time (even after the handshake). Probably we should fail to set it if KTLS is in use?? (Would need to update docs in that case)

ssl/record/rec_layer_s3.c
junjunk: Yes, that is checked on configuration stage

ssl/tls13_enc.c
junjunk: Additional check is added in configuration.
And also added checks to SSL_set_record_padding_callback. Checks are not needed in SSL_CTX_set_record_padding_callback because this value is not propagated into already established SSLs.

junjunk:
@mattcaswell Hi again, can you please check it one more time - I have added an additional check as you mentioned, and changed the documentation

ssl/ssl_lib.c
```diff
@@ -4624,11 +4624,18 @@ int SSL_CTX_set_block_padding(SSL_CTX *ctx, size_t block_size)
...
-void SSL_set_record_padding_callback(SSL *ssl,
+int SSL_set_record_padding_callback(SSL *ssl,
                                      size_t (*cb) (SSL *ssl, int type,
                                                    size_t len, void *arg))
 {
-    ssl->record_padding_cb = cb;
+    BIO *b;
+
+    b = SSL_get_wbio(ssl);
+    if (!b || !BIO_get_ktls_send(b)) {
```
mattcaswell, line 4634: Change `!b` to `b == NULL`

junjunk: Done

doc/man3/SSL_CTX_set_record_padding_callback.pod
```diff
@@ -32,6 +32,8 @@ SSL_set_block_padding - install callback to specify TLS 1.3 record padding
 SSL_CTX_set_record_padding_callback() or SSL_set_record_padding_callback()
 can be used to assign a callback function I<cb> to specify the padding
 for TLS 1.3 records. The value set in B<ctx> is copied to a new SSL by SSL_new().
+Kernel TLS is not possible if record padding callback is set. And also callback
+function cannot be set if Kernel TLS is already configured on current SSL.
```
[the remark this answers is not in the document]
junjunk: Done

doc/man3/SSL_CTX_set_record_padding_callback.pod
```diff
@@ -32,6 +32,8 @@ SSL_set_block_padding - install callback to specify TLS 1.3 record padding
 SSL_CTX_set_record_padding_callback() or SSL_set_record_padding_callback()
 can be used to assign a callback function I<cb> to specify the padding
 for TLS 1.3 records. The value set in B<ctx> is copied to a new SSL by SSL_new().
+Kernel TLS is not possible if record padding callback is set. And also callback
```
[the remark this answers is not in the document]
junjunk: Done

doc/man3/SSL_CTX_set_record_padding_callback.pod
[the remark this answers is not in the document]
junjunk: Done

doc/man3/SSL_CTX_set_record_padding_callback.pod
```diff
@@ -32,6 +32,9 @@ SSL_set_block_padding - install callback to specify TLS 1.3 record padding
 SSL_CTX_set_record_padding_callback() or SSL_set_record_padding_callback()
 can be used to assign a callback function I<cb> to specify the padding
 for TLS 1.3 records. The value set in B<ctx> is copied to a new SSL by SSL_new().
+Kernel TLS is not possible if the record padding callback is set. And also the
```
bsdjhb, line 35: Maybe adjust: s/set. And also the/set, and the/.

include/internal/ktls.h
```diff
@@ -338,8 +308,132 @@ static ossl_inline int ktls_read_record(int fd, void *data, size_t length)
     return ret;
 }
 
-#    endif
-#   endif
+#   endif /* OPENSSL_NO_KTLS_RX */
+
+/* Function to check supported ciphers in Linunx */
```
bsdjhb, line 313: s/Linunx/Linux/

doc/man3/SSL_CTX_set_record_padding_callback.pod
```diff
@@ -64,6 +67,9 @@ indicates no padding will be added. A return value that causes the record to
 exceed the maximum record size (SSL3_RT_MAX_PLAIN_LENGTH) will pad out to the
 maximum record size.
 
+The SSL_CTX_get_record_padding_callback_arg() function returns 1 on sccess or 0 if
```
t8m, line 70: Typo: sccess.

doc/man3/SSL_CTX_set_record_padding_callback.pod
junjunk: Applied, thanks

include/internal/ktls.h
junjunk: Fixed typo, thanks

doc/man3/SSL_CTX_set_record_padding_callback.pod
junjunk: Fixed typo, thanks

doc/man3/SSL_CTX_set_record_padding_callback.pod
```diff
@@ -16,7 +16,7 @@ SSL_set_block_padding - install callback to specify TLS 1.3 record padding
  #include <openssl/ssl.h>
 
  void SSL_CTX_set_record_padding_callback(SSL_CTX *ctx, size_t (*cb)(SSL *s, int type, size_t len, void *arg));
- void SSL_set_record_padding_callback(SSL *ssl, size_t (*cb)(SSL *s, int type, size_t len, void *arg));
+ int SSL_set_record_padding_callback(SSL *ssl, size_t (*cb)(SSL *s, int type, size_t len, void *arg));
```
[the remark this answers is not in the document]
mattcaswell: I started a discussion on openssl-project about this. So far the feedback seems to be that people are comfortable with this change.

doc/man3/SSL_CTX_set_record_padding_callback.pod
```diff
@@ -84,6 +89,9 @@ L<ssl(7)>, L<SSL_new(3)>
 
 The record padding API was added for TLS 1.3 support in OpenSSL 1.1.1.
 
+The return type of SSL_CTX_set_record_padding_callback() function was
+changed to int in OpenSSL 3.0.0.
```
mattcaswell, line 93: Please make that just "OpenSSL 3.0"

doc/man3/SSL_CTX_set_record_padding_callback.pod
junjunk: fixed

doc/man3/SSL_CTX_set_record_padding_callback.pod
junjunk: Changed

ssl/tls13_enc.c
junjunk: @mattcaswell Can you please look at my comments? If you can please explain me the need to refetch IV in TLSv1.2 GCM mode?

include/internal/ktls.h
```diff
@@ -309,6 +309,103 @@ static ossl_inline int ktls_read_record(int fd, void *data, size_t length)
...
+    unsigned char geniv[12];
+    unsigned char *iiv = iv;
+
+    if (tls_version == TLS1_2_VERSION &&
+        EVP_CIPHER_mode(c) == EVP_CIPH_GCM_MODE) {
+        EVP_CIPHER_CTX_ctrl(dd, EVP_CTRL_GET_IV,
+                            EVP_GCM_TLS_FIXED_IV_LEN + EVP_GCM_TLS_EXPLICIT_IV_LEN,
+                            geniv);
+        iiv = geniv;
+    }
```
junjunk, line 353: @mattcaswell I mean this part of code, it was taken from initial commit of Kernel TLS

bsdjhb: It looks like it is reading the value OpenSSL computed for the explicit part of the IV in TLS 1.2.  The IV for GCM (and I believe for CCM) is split into an implicit salt computed from the PRF, an explicit 8 byte counter that increments for each TLS PDU, and a 4-byte block counter that increments for each block within a PDU.  The 8 byte explicit value is on the wire, so the sender is free to choose any value with the constraint as I understand that the explicit IV must be unique and never reused with the same session key and salt.  One option permitted by the RFC is to reuse the TLS sequence number as the explicit IV.  OpenSSL generates a random 8 byte value that it then increments for each TLS PDU.  In FreeBSD's KTLS, we only pass in the salt from userland to the kernel and the kernel then generates its own random starting value that it increments for each TLS PDU.  It seems here that the existing code is reusing the random value generated by OpenSSL for the starting value of the explicit IV rather than relying on the kernel to generate its own random starting value (or just using the TLS record sequence number as the explicit IV).

So, the real question is how does the Linux kernel expect to deal with the explicit 8 byte IV for TLS 1.2?  Does it expect the socket option to initialize the starting value, or does it generate its own?  As it is, right now for CCM you are passing the next 8 bytes of the PRF result after the IV as the explicit IV's starting value.  That might be ok as long as the kernel keeps incrementing the value, (and the value is public on the wire anyway), but technically you are reading beyond the bounds of 'iiv'.  Most likely you need to fetch the full IV for CCM as well for TLS 1.2 to avoid that out-of-bounds read.

For TLS 1.3, the entire 12 byte nonce is implicit from the PRF.

junjunk: John, thank you very much for clarification. I have found the place in the code where explicit part of IV is filled with random data in case of GCM (it is in EVP_CTRL_GCM_SET_IV_FIXED section), and that's it should be read. But for CCM mode only implicit salt is copied to CIPHER_CTX and explicit part is untouched. The iv is taken from s->s3.tmp.key_block, and that key_block is allocated and computed from PRF. The size of s->s3.tmp.key_block is computed with full size of IV, so there is no out-of-bounds read, and the value is permitted by the RFC 6655.
The Linux kernel expects that socket option must initialize the starting value and then it just keeps incrementing it.
So, again, thank you very much for clarification, now I fully understand the need for this part of code.

junjunk:
@mattcaswell Is it possible to merge this PR before OpenSSL 3.0 release?

mattcaswell:
> Needs second review.

Ping @openssl/committers 

> @mattcaswell Is it possible to merge this PR before OpenSSL 3.0 release?

I fully expect this to be in OpenSSL 3.0 - although it is unlikely to be in our next alpha (this week).

junjunk:
Fixed a problem in TLSv1.3 code with Kernel TLS. During setup of TLSv1.3 connection the **tls13_change_cipher_state()** is called twice, once with SSL3_CC_HANDSHAKE flag and again with SSL3_CC_APPLICATION. We cannot reconfigure kernel state machine with new configuration so we must configure Kernel TLS only in case of SSL3_CC_APPLICATION.

bsdjhb:
FWIW, this looks ok to me.  I'm not an OpenSSL committer, but I am somewhat familiar with the KTLS bits at least.

junjunk:
Ping @openssl/committers
Please review this PR

ssl/t1_enc.c
```diff
@@ -109,6 +109,7 @@ static int tls1_generate_key_block(SSL *s, unsigned char *km, size_t num)
   * record layer. If read_ahead is enabled, then this might be false and this
   * function will fail.
   */
+#ifndef OPENSSL_NO_KTLS_RX
```
paulidale, line 112: needs a space between the `#` and the `ifndef`.
likewise on the closing `#endif`

ssl/tls13_enc.c
```diff
@@ -454,7 +456,7 @@ static int derive_secret_key_and_iv(SSL *s, int sending, const EVP_MD *md,
     if (!tls13_derive_key(s, md, secret, key, keylen)
             || !tls13_derive_iv(s, md, secret, iv, ivlen)) {
         /* SSLfatal() already called */
-        goto err;
+        return 0;
```
paulidale, line 459: If the second call here fails, won't `key` contain sensitive information which isn't being cleansed?

I guess with the connection dropping, it's not that important.

test/sslapitest.c
```diff
@@ -1164,85 +1157,381 @@ static int test_ktls_sendfile(void)
     return testresult;
 }
 
-static int test_ktls_no_txrx_client_no_txrx_server(void)
+static int test_ktls_no_txrx_client_no_txrx_server(int tlsver)
 {
-    return execute_test_ktls(0, 0, 0, 0);
+    int testresult = 1;
+#ifdef OPENSSL_KTLS_AES_GCM_128
```
paulidale, line 1163: Blank line before.

test/sslapitest.c
```diff
@@ -1164,85 +1157,381 @@ static int test_ktls_sendfile(void)
...
+#endif
+    return testresult;
 }
 
-static int test_ktls_no_rx_client_no_txrx_server(void)
+static int test_ktls_no_rx_client_no_txrx_server(int tlsver)
 {
-    return execute_test_ktls(1, 0, 0, 0);
+    int testresult = 1;
+#ifdef OPENSSL_KTLS_AES_GCM_128
```
paulidale, line 1181: Blank line before.

test/sslapitest.c
```diff
@@ -1164,85 +1157,381 @@ static int test_ktls_sendfile(void)
...
+#endif
+    return testresult;
 }
 
-static int test_ktls_no_tx_client_no_txrx_server(void)
+static int test_ktls_no_tx_client_no_txrx_server(int tlsver)
 {
-    return execute_test_ktls(0, 1, 0, 0);
+    int testresult = 1;
+#ifdef OPENSSL_KTLS_AES_GCM_128
```
paulidale, line 1199: Blank line before.

test/sslapitest.c
```diff
@@ -1164,85 +1157,381 @@ static int test_ktls_sendfile(void)
...
+#endif
+    return testresult;
 }
 
-static int test_ktls_client_no_txrx_server(void)
+static int test_ktls_client_no_txrx_server(int tlsver)
 {
-    return execute_test_ktls(1, 1, 0, 0);
+    int testresult = 1;
+#ifdef OPENSSL_KTLS_AES_GCM_128
```
paulidale, line 1217: Blank line before.

And more cases below.

ssl/t1_enc.c
junjunk: fixed

test/sslapitest.c
junjunk: fixed all style problems in tests

ssl/tls13_enc.c
junjunk: It will be cleaned by the callers of `derive_secret_key_and_iv` in case of any error.

paulidale: Okay, Thanks.  I figured something like this would be the case.

junjunk:
rebased on top of master tree

test/sslapitest.c
paulidale: Thanks.

mattcaswell:
Pushed. Thanks for your contribution!

PR #11591, opened 2020-04-21
Make test/fipsinstall.pl when running via the test harness

levitte:


levitte:
Hmmmmmmmmmmm...  the result I expected was something like this:

```
dyld: Library not loaded: /usr/local/lib/libssl.3.dylib
  Referenced from: /Users/travis/build/openssl/openssl/_build/tree/test-runs/test_evp_fetch_prov/../../apps/openssl
  Reason: image not found
# fipsinstall.pl: ../../apps/openssl fipsinstall -out "../../providers/fipsinstall.cnf" -module "../../providers/fips.so" -provider_name "fips" -mac_name "HMAC" -macopt "digest:SHA256" -macopt "hexkey:00" -section_name "fips_sect"  => 0
../../util/wrap.pl /usr/bin/perl ../../../../_srcdist/test/fipsinstall.pl ../.. => 0
```

... but all I got was:

```
dyld: Library not loaded: /usr/local/lib/libssl.3.dylib
  Referenced from: /Users/travis/build/openssl/openssl/_build/tree/test-runs/test_evp_fetch_prov/../../apps/openssl
  Reason: image not found
../../util/wrap.pl /usr/bin/perl ../../../../_srcdist/test/fipsinstall.pl ../.. => 0
```

I can only conclude that environment variables aren't always passed down?  That's at least a tentative guess, and that would explain the MacOS issue we see, as it means that the modifications to DYLD_LIBRARY_PATH and others done by wrap.pl may also be lost, which would explain why it tries to load libssl.3.dylib from /usr/local/lib.

Note that this is a *tentative guess*...  I need to do some research before saying anything definitive.

levitte:
BTW, help would be highly appreciated from anyone who listens.  I'm fumbling in the dark.

levitte:
I added a temporary commit, bringing out the big guns: debug prints...

levitte:
Ok, something is quite screwed up.  With the debug prints currently in place, I can see that `util/shlib_wrap.pl` (which is called by `util/wrap.pl` sets LD_LIBRARY_PATH, DYLD_LIBRARY_PATH, SHLIB_PATH and LIBPATH, and does export them, but `test/fipsinstall.pl` doesn't get to see them.  At.  All.

Interestingly enough, PATH comes through???

These lines show this in action:

https://travis-ci.org/github/openssl/openssl/jobs/677614888#L4733-L4881

levitte:
It seems that others have run into the same problem, and have solved it with diverse hackery:

https://github.com/nteract/nteract/issues/1523
https://github.com/Perl/perl5/issues/15057
https://cdcvs.fnal.gov/redmine/issues/18942
https://www.perlmonks.org/?node_id=1156743

So at this point, I'm thinking that `test/fipsinstall.pl` is not worth spending more time on...  sadly :disappointed: .  Let's throw it away.

paulidale:
It's by design :(

https://developer.apple.com/library/archive/documentation/Security/Conceptual/System_Integrity_Protection_Guide/RuntimeProtections/RuntimeProtections.html

levitte:
So basically, pure luck that we haven't wrapped anything perly with `util/shlib_wrap.sh` before now...  I can only assume that the standard installation of perl is as a protected binary.

levitte:
We *could* re-enable the DYLD_INSERT_LIBRARIES feature for MacOS X...  but nah, won't help, that would only pre-load the libraries for perl itself, not for anything the script runs...

paulidale:
That would be my guess.  `/usr` (except `/usr/local`) is protected.  Perl is `/usr/bin/perl`.

Locally, I've got perl in `/usr/local` so it won't be protected.  However, the script first executes `/usr/bin/env` which is protected.  The perl shipped with OS/X used to be too old (in OS/X 10.15.4 perl is v5.18.4 so should be okay).

A search brings up other projects that have been stung by this :(

Good detective work BTW.

levitte:
> Good detective work BTW.

Thanks :smile:

levitte:
Remova in #11592.  Closing this one...

paulidale:
Could we pass the path through in a different env variable and have `run` set it properly?
I doubt anything we build is protected.

levitte:
> Could we pass the path through in a different env variable and have `run` set it properly?

Not worth it for the moment, but let's keep that in mind for the day we hit a snag on something we just can't undo

t8m:
> Could we pass the path through in a different env variable and have `run` set it properly?
> I doubt anything we build is protected.

Yeah, the SHLIB_PATH and LIBPATH is passed fine. However I agree that undoing the fipsinstall  at this point is probably right.

PR #11593, opened 2020-04-21
Mention the 3.0 wiki page in the release announcement emails

mattcaswell:


dev/release-aux/openssl-announce-pre-release.tmpl
```diff
@@ -8,13 +8,18 @@
...
 
         https://www.openssl.org/news/openssl-$series-notes.html
 
    Note: This OpenSSL pre-release has been provided for testing ONLY.
    It should NOT be used for security critical purposes.
 
+   Specific notes on upgrading to OpenSSL $series from previous versions, as well
+   as known issues are available on the OpenSSL Wiki, here:
+
+        https://wiki.openssl.org/index.php/OpenSSL_$series
```
levitte, line 21: I see future plans in the making 😉

dev/release-aux/openssl-announce-release.tmpl
```diff
@@ -7,10 +7,15 @@
 
    The OpenSSL project team is pleased to announce the release of
    version $release of our open source toolkit for SSL/TLS.
-   For details of changes and known issues see the release notes at:
+   For details of changes see the release notes at:
 
         https://www.openssl.org/news/openssl-$series-notes.html
 
+   Specific notes on upgrading to OpenSSL $series from previous versions, as well
+   as known issues are available on the OpenSSL Wiki, here:
```
levitte, line 15: The status bit was meant for pre-releases (alpha & beta). Not sure I see a reason to keep it afterwards

dev/release-aux/openssl-announce-pre-release.tmpl
```diff
@@ -8,13 +8,18 @@
    OpenSSL $series is currently in $label.
 
    OpenSSL $release_text has now been made available.  For details of
-   changes and known issues see the release notes at:
+   changes see the release notes at:
```
paulidale, line 11: _the_ changes?

dev/release-aux/openssl-announce-release.tmpl
```diff
@@ -7,10 +7,15 @@
 
    The OpenSSL project team is pleased to announce the release of
    version $release of our open source toolkit for SSL/TLS.
-   For details of changes and known issues see the release notes at:
+   For details of changes see the release notes at:
```
paulidale, line 10: I'd put a comma after changes.  Not a biggie either way.

dev/release-aux/openssl-announce-release.tmpl
levitte: Er....  this is not a side note

romen: > The status bit was meant for pre-releases (alpha & beta). Not sure I see a reason to keep it afterwards

@levitte I don't understand your comment in reference to the highlighted lines!
Can you elaborate please?

levitte: "as well as known issues are available on the OpenSSL Wiki" is the part that I was only thinking was for alpha/beta releases, to alert those testing that we know about it and intend to correct for the final release.

For "real" releases, I don't see that we need to preserve the STATUS part of the wiki page (other than possibly for historical reference)...

levitte: If you do see value for the STATUS bit beyond the alpha/beta releases, then you need to ask yourself if you're willing to maintain it.  It *is* extra work, and we're not fantastic at keeping things like that up to date for the longer term.

romen: Ah ok, I have to admit that even after rereading it after your feedback I don't see a 1:1 mapping between "known issues" and the STATUS section of the wiki.

For me that part of the text is not necessarily referred to the development status, but more broadly to known issue related to updating to the new version (which in the case of a development release also includes the STATUS section about things that are not entirely there yet).

In an announcement for the final 3.0 release or a future 6.0 release this phrasing would still make sense even in the absence of a STATUS section in the lined wiki.

romen: E.g. "caveat: if you maintain engines you are encouraged to migrate them into providers, otherwise all hell will break loose and loading your engine won't play nice with applications trying to use the new provider features" is a "known issue" that should survive even after all status is 100% and that section is removed.

mattcaswell: @romen's argument makes sense to me

levitte: Hmm, ok.

mattcaswell:
Pushed. Thanks!

PR #6074, opened 2018-04-24
Avoid undefined behavior with unaligned accesses

bernd-edlinger:
Fixes: #4983

PR #11596, opened 2020-04-21
[crypto/ec] ec_nistp_64_gcc_128 enable-ubsan fixes

bbbrumley:
@romen told me some of these tests were failing. I didn't know what `enable-ubsan` does but anyway here are the fixes.

```
$ ./config -d no-shared enable-ec_nistp_64_gcc_128 enable-ubsan -Werror

1..16
crypto/ec/ecp_nistp521.c:147:15: runtime error: load of misaligned address 0x7ffd39ea9aa7 for type 'limb', which requires 8 byte alignment
0x7ffd39ea9aa7: note: pointer points here
 c2 31 7e 7e f9  9b 42 6a 85 c1 b3 48 33  de a8 ff a2 27 c1 1d fe  28 59 e7 ef 77 5e 4b a1  ba 3d 4d
             ^ 

1..16
crypto/ec/ecp_nistp521.c:187:28: runtime error: load of misaligned address 0x7ffe68fe96f7 for type 'limb', which requires 8 byte alignment
0x7ffe68fe96f7: note: pointer points here
 c2 31 7e 7e 01  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00
             ^ 

crypto/ec/ecp_nistp224.c:317:15: runtime error: load of misaligned address 0x7ffcf66ee6f7 for type 'const uint64_t', which requires 8 byte alignment
0x7ffcf66ee6f7: note: pointer points here
 11 d6 80 32 34  22 11 c2 56 d3 c1 03 4a  b9 90 13 32 7f bf b4 6b  bd 0c 0e b7 2d 56 00 00  10 5b 6b
             ^ 
```

I think the possible unaligned loads/stores are undefined behavior and will crash on some obscure architecture that someone will drag out of a cave.

PS The tests are failing all over the place with `enable-ubsan` so it's probably something to look into.

PPS I reckon the P-256 version of this one will also throw errors but the foolish logic in `ec_curve.c` ignores when users build with `enable-ec_nistp_64_gcc_128` and chooses `nistz` anyway when it can. So I'm not going to check it because that's a dumb design decision. (Do you know how many distribution packages are built with that? I wonder if they are aware.)

### DIFF
--- a/crypto/ec/ecp_nistp521.c
+++ b/crypto/ec/ecp_nistp521.c
@@ -43,6 +43,11 @@
 #include <openssl/err.h>
 #include "ec_local.h"
 
+#define STRICT_ALIGNMENT 1
+#ifndef PEDANTIC
+# undef STRICT_ALIGNMENT
+#endif
+
 #if defined(__SIZEOF_INT128__) && __SIZEOF_INT128__==16
   /* even with gcc, the typedef won't work for 32-bit platforms */
 typedef __uint128_t uint128_t;  /* nonstandard; implemented by gcc on 64-bit
@@ -143,6 +148,18 @@ static const limb bottom58bits = 0x3ffffffffffffff;
  */
 static void bin66_to_felem(felem out, const u8 in[66])
 {
+#ifdef STRICT_ALIGNMENT
+    memcpy(out, in, 66);
+    out[8] = (out[8] << 48 | out[7] >> 16) & bottom57bits;
+    out[7] = (out[7] << 42 | out[6] >> 22) & bottom58bits;
+    out[6] = (out[6] << 36 | out[5] >> 28) & bottom58bits;
+    out[5] = (out[5] << 30 | out[4] >> 34) & bottom58bits;
+    out[4] = (out[4] << 24 | out[3] >> 40) & bottom58bits;
+    out[3] = (out[3] << 18 | out[2] >> 46) & bottom58bits;
+    out[2] = (out[2] << 12 | out[1] >> 52) & bottom58bits;
+    out[1] = (out[1] << 6 | out[0] >> 58) & bottom58bits;
+    out[0] &= bottom58bits;
+#else
     out[0] = (*((limb *) & in[0])) & bottom58bits;
     out[1] = (*((limb *) & in[7]) >> 2) & bottom58bits;
     out[2] = (*((limb *) & in[14]) >> 4) & bottom58bits;
@@ -152,6 +169,7 @@ static void bin66_to_felem(felem out, const u8 in[66])
     out[6] = (*((limb *) & in[43]) >> 4) & bottom58bits;
     out[7] = (*((limb *) & in[50]) >> 6) & bottom58bits;
     out[8] = (*((limb *) & in[58])) & bottom57bits;
+#endif
 }
 
 /*
@@ -160,6 +178,19 @@ static void bin66_to_felem(felem out, const u8 in[66])
  */
 static void felem_to_bin66(u8 out[66], const felem in)
 {
+#ifdef STRICT_ALIGNMENT
+    felem squeeze;
+    squeeze[0] = in[0] | in[1] << 58;
+    squeeze[1] = in[2] << 52 | in[1] >> 6;
+    squeeze[2] = in[3] << 46 | in[2] >> 12;
+    squeeze[3] = in[4] << 40 | in[3] >> 18;
+    squeeze[4] = in[5] << 34 | in[4] >> 24;
+    squeeze[5] = in[6] << 28 | in[5] >> 30;
+    squeeze[6] = in[7] << 22 | in[6] >> 36;
+    squeeze[7] = in[8] << 16 | in[7] >> 42;
+    squeeze[8] = in[8] >> 48;
+    memcpy(out, squeeze, 66);
+#else
     memset(out, 0, 66);
     (*((limb *) & out[0])) = in[0];
     (*((limb *) & out[7])) |= in[1] << 2;
@@ -170,6 +201,7 @@ static void felem_to_bin66(u8 out[66], const felem in)
     (*((limb *) & out[43])) |= in[6] << 4;
     (*((limb *) & out[50])) |= in[7] << 6;
     (*((limb *) & out[58])) = in[8];
+#endif
 }
 
 /* BN_to_felem converts an OpenSSL BIGNUM into an felem */
--- a/crypto/ec/ecp_nistp224.c
+++ b/crypto/ec/ecp_nistp224.c
@@ -43,6 +43,11 @@
 #include <openssl/err.h>
 #include "ec_local.h"
 
+#define STRICT_ALIGNMENT 1
+#ifndef PEDANTIC
+# undef STRICT_ALIGNMENT
+#endif
+
 #if defined(__SIZEOF_INT128__) && __SIZEOF_INT128__==16
   /* even with gcc, the typedef won't work for 32-bit platforms */
 typedef __uint128_t uint128_t;  /* nonstandard; implemented by gcc on 64-bit
@@ -313,10 +318,18 @@ const EC_METHOD *EC_GFp_nistp224_method(void)
  */
 static void bin28_to_felem(felem out, const u8 in[28])
 {
+#ifdef STRICT_ALIGNMENT
+    memset(out, 0, sizeof(felem));
+    memcpy(out, in, 7);
+    memcpy(&out[1], in + 7, 7);
+    memcpy(&out[2], in + 14, 7);
+    memcpy(&out[3], in + 21, 7);
+#else
     out[0] = *((const uint64_t *)(in)) & 0x00ffffffffffffff;
     out[1] = (*((const uint64_t *)(in + 7))) & 0x00ffffffffffffff;
     out[2] = (*((const uint64_t *)(in + 14))) & 0x00ffffffffffffff;
     out[3] = (*((const uint64_t *)(in+20))) >> 8;
+#endif
 }
 
 static void felem_to_bin28(u8 out[28], const felem in)


romen:
The other tests failing when configuring with `enable-ubsan` are also due to alignment and disappear if compiling with `-DPEDANTIC`:

https://github.com/openssl/openssl/blob/07caec83b81859ea9aa2d5075a394aa48c4e5fae/include/crypto/modes.h#L30-L39

https://github.com/openssl/openssl/blob/07caec83b81859ea9aa2d5075a394aa48c4e5fae/crypto/modes/cbc128.c#L86-L111


@bbbrumley  Do you believe it is worth to keep the old code and guard these changes with `#ifdef STRICT_ALIGNMENT`

And add something like this in the preamble?

```C
#define STRICT_ALIGNMENT 1
#ifndef PEDANTIC
# undef STRICT_ALIGNMENT
#endif
```

@mattcaswell what is your take on the matter?

bbbrumley:
> Do you believe it is worth to keep the old code and guard these changes with `#ifdef STRICT_ALIGNMENT`

Yes -- hacking up the code like this for irrelevant architectures is a disgrace!

mattcaswell:
> And add something like this in the preamble?

Makes sense to me

bbbrumley:
> Makes sense to me

Fixed in force push 2aef3aeeba22cc8cacafe09b1c9e35c6785f04da

romen:
@mattcaswell this technically fixes a build problem so it might be considered as matching the `urgent` requirements.

I am going to add the label now, and leave it to you to consider if it should be part of the alpha1 release.

In any case I volunteer to take care of backporting to 1.1.1 after the release is done.

romen:
No need for the `urgent` label anymore, as release window is over and the grace period is almost over as well.

bernd-edlinger:
@romen 

I have not followed all here, but please do not back-port the STRICT_ALIGNMENT change

romen:
@bernd If you have some concern about this please review!

I will wait to merge to both branches.

bernd-edlinger:
FYI I had an old PR, where I tried to fix the ubsan errors, which are real errors
in ARMv4 arch.  #6074 but I gave up already a long time ago.

bernd-edlinger:
Yes, I have concerns.

-1 for 1.1.1 at least.

crypto/ec/ecp_nistp521.c
```diff
@@ -43,6 +43,11 @@
 #include <openssl/err.h>
 #include "ec_local.h"
 
+#define STRICT_ALIGNMENT 1
```
bernd-edlinger, line 46: Hmm, this is x86 only, right?
but STRICT_ALIGNMENT is ARMv4 and PowerPC.

bernd-edlinger: So I would kindly ask for a different name of this define.

romen: This is how it is done elsewhere, whitelisting most platforms but not all of them:
https://github.com/openssl/openssl/blob/07caec83b81859ea9aa2d5075a394aa48c4e5fae/include/crypto/modes.h#L30-L39

Would this be acceptable to you?

bernd-edlinger:
Is enable-ubsan an 1.1.1 option?

romen:
> The other tests failing when configuring with `enable-ubsan` are also due to alignment and disappear if compiling with `-DPEDANTIC`:
> 
> https://github.com/openssl/openssl/blob/07caec83b81859ea9aa2d5075a394aa48c4e5fae/include/crypto/modes.h#L30-L39
> 
> https://github.com/openssl/openssl/blob/07caec83b81859ea9aa2d5075a394aa48c4e5fae/crypto/modes/cbc128.c#L86-L111
> 
> @bbbrumley Do you believe it is worth to keep the old code and guard these changes with `#ifdef STRICT_ALIGNMENT`
> 
> And add something like this in the preamble?
> 
> ```c
> #define STRICT_ALIGNMENT 1
> #ifndef PEDANTIC
> # undef STRICT_ALIGNMENT
> #endif
> ```
> 
> @mattcaswell what is your take on the matter?



> Is enable-ubsan an 1.1.1 option?

-pedantic is!

kroeckx:
> Is enable-ubsan an 1.1.1 option?

It is, but I don't think that's relevant

bernd-edlinger:
It is, this PR is about enable-ubsan false positive warnings.

bernd-edlinger:
@romen what do you mean with -pedantic ?

kroeckx:
Either we consider the misaligned access a bug, and we fix it in
1.1.1, or we don't. That enable-ubsan as an option exists or not
does not have an effect on it being a bug or not.

crypto/ec/ecp_nistp521.c
bernd-edlinger: Well, I thought about that, but this moule module is only x86 ?

bernd-edlinger:
It is a performance regression.

kroeckx:
If PEDANTIC is defined, we don't do misaligned accesses, even if
the archicture supports it.

bernd-edlinger:
you move too fast.  I would suggest to slow down.

romen:
I added the otc hold label to avoid inadvertently merging this before we figure out how to address this.

I don't understand completely @bernd-edlinger concerns, as this code anyway is not reachable in arm4 as `enable-ec_nistp_64_gcc_128` has requirements that are not met by that platform.

Also the logic for 

~~~C
#define STRICT_ALIGNMENT 1
#ifndef PEDANTIC
# undef STRICT_ALIGNMENT
#endif
~~~

is the same that is already in place elsewhere (and is required both for builds with `enable-ubsan` and for builds with the `-pedantic` option, so it applies to both 1.1.1 and master): 
https://github.com/openssl/openssl/blob/07caec83b81859ea9aa2d5075a394aa48c4e5fae/include/crypto/modes.h#L30-L39


We could add the explicit whitelisting but I am quite sure that the conditions required to be able to use `enable-ec_nistp_64_gcc_128` are already more restrictive than that list.

bernd-edlinger:
I admit it may be possible that I am wrong,
but how I see it, is that this code is currently only x86 but
I fail to see why that cannot be other achitectures as well.
Maybe we should just move the "STRICT_ALIGNMENT"
define to a common include file, then it would not look so
odd, like an slippery place waiting for someone to fall over it.

bernd-edlinger:
BTW, while it is rather unfortunate that GCC does not have a predefined
preprocessor macro for "strict" alignment, it has a define for little endian
and big endian, maybe we have also defines for that ?

bernd-edlinger:
I think the -DPEDANTIC is not automatically set with enable-ubsan,
nor with -pedantic (which is just a GCC option)

see Configure
```
unless ($disabled{ubsan}) {
    # -DPEDANTIC or -fnosanitize=alignment may also be required on some
    # platforms.
    push @{$config{cflags}}, "-fsanitize=undefined", "-fno-sanitize-recover=all";
}
```
So while -fno-sanitize-revocer=all is just a bad idea,  using -fno-sanitize=alignment
sems like a much better way to fix this.

kroeckx:
> So while -fno-sanitize-revocer=all is just a bad idea,  using -fno-sanitize=alignment
> sems like a much better way to fix this.

We currently do:
unless ($disabled{ubsan} || defined $detected_sanitizers{ubsan}) {
    # -DPEDANTIC or -fnosanitize=alignment may also be required on some
    # platforms.
    push @{$config{cflags}}, "-fsanitize=undefined", "-fno-sanitize-recover=all";
}

(The condition looks wrong to me.)

kroeckx:
I was clearly little confused.

Anyway, in fuzz/README.md, we say to use -fno-sanitize=alignment

romen:
How do you enable `-fno-sanitize=alignment` just for this file?

How does using `-fno-sanitize=alignment` fix the fact that a build with `-DPEDANTIC -pedantic enable-ec_nistp_64_gcc_128` will fail because `-pedantic` will check for unaligned memory accesses and fail?

Why using 
https://github.com/openssl/openssl/blob/07caec83b81859ea9aa2d5075a394aa48c4e5fae/include/crypto/modes.h#L30-L39
is not an issue but doing the same here is cause for concern?

bernd-edlinger:
Oops, not sure if that is really okay.
this should not mix things together
a strict alignment platform has per definition
one where int* need has to be 4-byte aligned,
or otherwise an exception occurs.
now when this is in a header file, that is not
good at all.

romen:
My impression is that there is some confusion, maybe related to naming conventions internal to GCC and the name of the macro used by us.

I'll summarize here what I believe is happening, but I might be wrong, so please correct me so that I can better understand your arguments.

- As far as I can tell, defining a macro called `STRICT_ALIGNMENT` is not redefining a reserved keyword and does not have any effect in included headers.
- Inside GCC implementation a macro with the same name seems to be used to identify some specific platforms.
- Inside OpenSSL we use a macro called `STRICT_ALIGNMENT` set to 1 to indicate that implementations should avoid violating address alignment, even in platforms where strict alignment is not a requirement.

We need a macro to enable this behavior, because we want to be able to use static and dynamic analyzers (the `-pedantic` compiler option or the `enable-ubsan` `Configure` option), and those often tend to complain about misaligned addressing (even on platforms where it is not necessarily unsupported) because it is not 100% compliant with C.

At a global level we use a macro called `PEDANTIC` to enable all kind of workarounds to force behavior compatible with the `-pedantic` option, even in platforms where for regular builds the "_pedantic behavior_" is not required.

Instead of overloading the `PEDANTIC` macro with all kinds of semantics, for the misalignment issue we decided to use a dedicated macro called `STRICT_ALIGNMENT`, that is disabled/undefined/set-to-false on platforms where strict alignment is not a requirement, __unless__ `PEDANTIC` is defined.

Currently `STRICT_ALIGNMENT` is used exclusively in `crypto/modes/` and in `crypto/whrlpool` and there isn't any configuration target that enables `STRICT_ALIGNMENT` globally.
As far as I can tell, it is always defined internally to each relevant translation unit, using the scheme of https://github.com/openssl/openssl/blob/07caec83b81859ea9aa2d5075a394aa48c4e5fae/include/crypto/modes.h#L30-L39

In this PR we use the same approach, but we skip the platform whitelisting because anyway this file cannot be compiled in the platforms where `STRICT_ALIGNMENT` shouldn't be undefined, as they don't meet the requirements for `enable-ec_nistp_64_gcc_128`:
https://github.com/openssl/openssl/blob/96ebe52e897dea29664683e138877fb5eb995e4d/INSTALL.md#L630-L641

romen:
> It is a performance regression.

@bernd-edlinger this shouldn't really be a concern here for the following reasons:
1. Defining `PEDANTIC` is not meant for production use anyway, and already causes all kinds of performance regressions, __intentionally__.
2. `STRICT_ALIGNMENT` is never defined globally per target platform in the `Configure` step; it's always redefined and undefined internally to single translation units, so there isn't a risk of propagating it by mistake.
3. Even if we didn't guard the pedantic version of the code of this PR and we used it by default (even if ugly), in this particular case the performance regression would be minimal, as the code we are discussing is called once during setup and once after the operation is performed, just to convert to/from the internal representation.

romen:
> I was clearly little confused. Anyway, in fuzz/README.md, we say to use -fno-sanitize=alignment

@kroeckx I believe once this goes in (or as an amendment in this PR) we can update `fuzz/README.md`:
the requirement for `-fno-sanitize=alignment` seems to be exactly caused by `enable-ec_nistp_64_gcc_128`, apparently because the files modified by this PR are the only ones that even in presence of `-DPEDANTIC` still do unaligned accesses.

I believe it is better to locally fix the individual files where unaligned accesses are carefully and intentionally used, rather than using a global `-fno-sanitize=alignment` blanket option that would hide errors/warnings if this behavior were added outside of them, and we forgot to properly guard/whitelist those sections by the `STRICT_ALIGNMENT`/`PEDANTIC` logic.

bernd-edlinger:
@romen 

> > It is a performance regression.
> 
> @bernd-edlinger this shouldn't really be a concern here for the following reasons:
> 

Ah, right, I mixed that up.  in my memory the other uses of STRICT_ALIGNMENT
were already questionable but I found it impossible to win the fight with Andy
to change these places.  But once this goes in, someone will fix the symmetry
break in the wrong way.  THAT is the problem I am concerned of.

The reasons you gave, are correct here, but they need at least be in the
comment so nobody cones along and copies this wrong use of "STRICT_ALIGNMMENT"
everywhere else.

Also in GCC the meaning of STRICT_ALIGNMENT is not at all clear.

Also in part since ARM is moving towards non-strict-alignment.
While ARMv4 is completely strict align ARMv7 is mostly non-strict-align.
But from GCC's STRICT_ALIGNMENT the whole arch is strict align.

Similarly X86 is non-strict-alignment, nevertheless certain vector-ops
are stict alignmen, and GCC optimizes this based on the assumption
that strict alignment does not happen.

I discussed that matter in depth in the other PR and nobody wanted to
listen.  Currently GCC does not have a way to optimize this code to use
vector ops but that can change any time.

romen:
> But once this goes in, someone will fix the symmetry
> break in the wrong way. THAT is the problem I am concerned of.

@bernd-edlinger can you elaborate on that?
Maybe I did not have enough tea yet to wake up fully! 😅
What do you mean by symmetry break? 

I also had another idea: would it improve the situation if we renamed the macro as a verb?
E.g. FORCE_STRICT_ALIGNMENT?

romen:
@bernd-edlinger do you have an update on this?

romen:
Ping

bernd-edlinger:
I have a probably related PR pending since over a year.
#11780
maybe this misses to address this ec_nistp_64_gcc_128 code (since it is often disabled)
and I do not have the impression that this code is faster than the assembler modules,
at least for P-256
Maybe we could use some ideas from there?

romen:
> I have a probably related PR pending since over a year.
> #11780
> maybe this misses to address this ec_nistp_64_gcc_128 code (since it is often disabled)
> and I do not have the impression that this code is faster than the assembler modules,
> at least for P-256
> Maybe we could use some ideas from there?

Is #11780 the right number? I can't see how they are related.

bernd-edlinger:
No sorry, this was the wrong one.
#11781   (+/-1 error)

bbbrumley:
> maybe this misses to address this ec_nistp_64_gcc_128 code (since it is often disabled)

I do not care about this PR at all. And it is somehow shocking to me that such an insignificant PR that is basically dead code has generated almost 50 comments. But if we're gonna make progress, it should be based on facts.

`enable-ec_nistp_64_gcc_128` is the most common configuration on common linux distributions. @bernd-edlinger go check the `rules` file of the Debian package.

romen:
For reference, here are the current `rules` for the debian build of OpenSSL: https://salsa.debian.org/debian/openssl/-/blob/debian/unstable/debian/rules#L44-46

Most debian derived distro, including Ubuntu (and flavours) enable it as well, so yes, `ec_nistp_64_gcc_128` is not disabled that often.

I haven't run recent benchmarks, but in all the works on optimizing the EC layer in the previous months, performance-wise (and security-wise) on supported platforms `ec_nistp_64_gcc_128` did make a difference!

kroeckx:
Being the Debian maintainer, I would enable it on more platforms
than just amd64 if I could. I didn't look at this for many years,
I guess I'll take a look again at which platforms I can enable it.

kroeckx:
So looking at this, with the current requirement that it needs
unaligned access, I think I'm limited to amd64, ppc64el and arm64.
If the unaligned access is restricted to those platforms that
support it, and doesn't make it slower, I would be able to enable
it on all 64 bit little endian platforms. (As far as I know, none
of the 32 bit platforms support __uint128_t.)

I've just tested it on arm64, most speeds didn't change, nistp224
and nistp521 got about 3 times faster.

bernd-edlinger:
That is what I also wanted to say, I only care about P-256
P-521 is irrelevant for me, as well as P-224
P-256 has a very fast assembler support, which is almost as
fast as X25519.

Regarding the alignment issue in this PR,
my proposal is this: 841f5f1afd61beef2acd0bfe7424d88a544c7919

I tested this with
`./Configure linux-x86_64 no-pic no-dso enable-md2 enable-rc5 enable-ec_nistp_64_gcc_128 no-asm -g -fsanitize=address,undefined -fno-sanitize-recover=all -fno-omit-frame-pointer`

and all tests pass without problem.  I believe telling the compiler that
the u64 is unaligned when it is unaligned should not need to be conditional,
as the compiler is known to be gcc, and this tells the sanitizer that the access
is unaligned, which is just the plain truth.

romen:
> So looking at this, with the current requirement that it needs unaligned access, I think I'm limited to amd64, ppc64el and arm64. If the unaligned access is restricted to those platforms that support it, and doesn't make it slower, I would be able to enable it on all 64 bit little endian platforms. (As far as I know, none of the 32 bit platforms support __uint128_t.) I've just tested it on arm64, most speeds didn't change, nistp224 and nistp521 got about 3 times faster.

Yes enabling this in theory affects nistp225 , nistp256 and nistp521, but the changes on nistp256 are shadowed by the nistz256 implementation that is even faster and picked over the implementation enabled by `enable-ec_nistp_64_gcc_128`.

romen:
@bernd-edlinger does 841f5f1afd61beef2acd0bfe7424d88a544c7919 have an impact on performance?

bernd-edlinger:
> @bernd-edlinger does 841f5f1 have an impact on performance?

I did not check, but I can't imagine any performance impact, since
the compiler just knows which assembler instruction to choose, once
it is known how the alignment is.  This should even work for armv4
which may allow the compiler to choose the best available assembler
instruction for each of the unaligned accesses.

bernd-edlinger:
>  but the changes on nistp256 are shadowed by the nistz256 implementation that is even faster and picked over the implementation 

note I tested with no-asm and there were no unaligned access in nistp256.

bbbrumley:
> This should even work for armv4

We had the conversation way, way previously in this thread. These entire `EC_METHOD`s are not reachable on armv4 or any 32-bit architecture.

romen:
So, in the interest of trying to move things forward, let's recap the current situation:

- holding back a solution to this issue prevents from compiling with `-d --strict-warnings no-shared enable-ec_nistp_64_gcc_128 enable-ubsan`. (I personally consider this a problem as I always build with this feature on, as it matches what is delivered to a huge portion of Linux binary distributions of OpenSSL).
- the proposed solution uses the same logic in https://github.com/openssl/openssl/blob/07caec83b81859ea9aa2d5075a394aa48c4e5fae/include/crypto/modes.h#L30-L39
  yet here it is cause of concern. (I don't fully understand why, and I can use more feedback on that to understand better). Consider that this logic has been in use for [at least the past 6 years](https://github.com/openssl/openssl/commit/021e5043e524b1cb28a929ef902548a987c16e65#diff-734c8c087acc3f8f7863e3e5cd301b55)!
- here, usage of `STRICT_ALIGNMENT` has nothing to do with `armv4` and/or other platforms where misaligned memory references are not tolerated or underperform (they are clearly excluded in the listed conditions for this feature), it's just about appeasing the `-pedantic` and `enable-ubsan` that raise warnings/errors for this undefined behavior, even on platforms that would tolerate this
- `-fno-sanitize=address` is the current documented solution, but it is suboptimal, as it has the potential of hiding places where such warnings/errors might be relevant as they could be reached compiling for unsupported platforms
- @bernd-edlinger has an alternative solution in 841f5f1afd61beef2acd0bfe7424d88a544c7919 , but it is in a PR targeting only 1.1.1, against our policy of approving on `master` first and then backporting to the other stable branches.
- Arguably 841f5f1afd61beef2acd0bfe7424d88a544c7919 is more elegant than the currently proposed solution:
  - `+` it's more concise
  - `+` apparently it shouldn't affect performance
  - `?` does it use GCC specific extensions?

The last item is relevant because, despite the option name, GCC (or a specific GCC version) is not currently listed among the requirements for this option:

https://github.com/openssl/openssl/blob/96ebe52e897dea29664683e138877fb5eb995e4d/INSTALL.md#L630-L641

If the last concern is relevant I would prefer the current solution of this PR, otherwise maybe we could close this and go with the solution proposed in 841f5f1afd61beef2acd0bfe7424d88a544c7919, but that PR is currently stalled.

To move things forward, could we instead approve this, merge it and then let a separate PR to improve the status quo about `STRICT_ALIGNMENT` (that would involve these changes and also existing patterns from [`include/crypto/modes.h` (in `master`)](https://github.com/openssl/openssl/blob/07caec83b81859ea9aa2d5075a394aa48c4e5fae/include/crypto/modes.h#L30-L39)/[`crypto/modes/modes_local.h` (in `1.1.1`)](https://github.com/openssl/openssl/blob/e512efe0894481679a5d3c57d10bf4ea97046c2a/crypto/modes/modes_local.h#L29-L38))?

bernd-edlinger:
FYI I have meanwhile re-based #6074 the original PR targeted at master.
and included the proposed solution to include the ec_nistp_64_gcc_128.

It covers *all* unaligned accesses that happen in the test cases, without need for
-DPEDANTIC and different code paths in sanitized vs. non-sanitized.

I am not aware of a compiler that would not support the __attribute((aligned))
gcc extension, and still have the required non-standard gcc types and defines.

romen:
> FYI I have meanwhile re-based #6074 the original PR targeted at master.
> and included the proposed solution to include the ec_nistp_64_gcc_128.
> 
> It covers _all_ unaligned accesses that happen in the test cases, without need for
> -DPEDANTIC and different code paths in sanitized vs. non-sanitized.
> 
> I am not aware of a compiler that would not support the __attribute((aligned))
> gcc extension, and still have the required non-standard gcc types and defines.

Apologies @bernd-edlinger , I did not notice!

Your answer addresses all my concerns, so I think it would be fine to solve the issue with #6074 and its backport: the only reason why I am not asking yet @bbbrumley if we can close this is that #6074 seemed to be controversial, even though it's not clear to me what was left to do after the whole discussion.

If once more an agreement on approving #6074 cannot be found (hopefully this won't be the case!!!), this would still be a second best solution.

romen:
In the meantime, I want to apologize also to @bbbrumley , as he started looking into this on my input, while I did not realize that #6074 was already there to try and address similar problems!

bbbrumley:
@bernd-edlinger FWIW I really like the attribute aligned trick -- it is slick and for sure I prefer it over this PR

> I am not aware of a compiler that would not support the __attribute((aligned))
> gcc extension, and still have the required non-standard gcc types and defines.

I'm leaning that way, too. And in the worse case, someone will open an issue for some obscure compiler arch/compiler combo we didn't think of, and we can tackle it then. Even better -- in the worst case it just won't compile so it can't lead to any security issue.

What about `icc`? Anyone willing to try? I think they have `__uint128_t` but I can't say about the alignment attributed and I don't have `icc` to test.

romen:
> @bernd-edlinger FWIW I really like the attribute aligned trick -- it is slick and for sure I prefer it over this PR
> 
> > I am not aware of a compiler that would not support the __attribute((aligned))
> > gcc extension, and still have the required non-standard gcc types and defines.
> 
> I'm leaning that way, too. And in the worse case, someone will open an issue for some obscure arch/compiler combo we didn't think of, and we can tackle it then. Even better -- in the worst case it just won't compile so it can't lead to any security issue.
> 
> What about `icc`? Anyone willing to try? I think they have `__uint128_t` but I can't say about the alignment attribute and I don't have `icc` to test.

I don't know about `icc`, but I see 2 spots where we use different ways of signalling unusual alignment based on the compiler:

https://github.com/openssl/openssl/blob/f7201301ef001b70109d7007a37525e233d30907/crypto/ec/ecp_nistz256.c#L40-L46

https://github.com/openssl/openssl/blob/f7201301ef001b70109d7007a37525e233d30907/crypto/ec/ecp_nistz256_table.c#L24-L30

Maybe we need to do something similar also in #6074 : I'll comment about it there

romen:
#6074 and #11781 are now merged and solve the initial issue together with other undefined behaviour issues with non-aligned accesses: we can close this as already fixed.

Thanks @bbbrumley for working on this and for the discussion.

PR #11598, opened 2020-04-21
dsa/dh_validate_private. return early if there is no private key

matbech:
When a private key is validated and there is no private key, return early.
Affected functions: dsa_validate_private and dh_validate_private.

### DIFF
--- a/providers/implementations/keymgmt/dh_kmgmt.c
+++ b/providers/implementations/keymgmt/dh_kmgmt.c
@@ -322,6 +322,8 @@ static int dh_validate_public(DH *dh)
     const BIGNUM *pub_key = NULL;
 
     DH_get0_key(dh, &pub_key, NULL);
+    if (pub_key == NULL)
+        return 0;
     return DH_check_pub_key_ex(dh, pub_key);
 }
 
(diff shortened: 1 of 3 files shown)


levitte:
Would you mind doing the same for the public key?  DSA and DH structure with neither public or private key are perfectly possible, they may contain domain parameters only...

matbech:
@levitte Done. Thank you for the suggestion.

slontis:
Maybe add a test (somewhere like evp_pkey_provided_test.c).

```
    if (!TEST_ptr(ctx = EVP_PKEY_CTX_new_from_name(NULL, "DSA", NULL)))
        || !TEST_false(EVP_PKEY_check(key_ctx))
        || !TEST_false(EVP_PKEY_public_check(key_ctx))
        || !TEST_false(EVP_PKEY_private_check(key_ctx))
        || !TEST_false(EVP_PKEY_pairwise_check(key_ctx)))  
       goto err;
```

slontis:
Oops.. You may want to change key_ctx to ctx and see if it works locally first :)

levitte:
Slightly squashed and merged

2fc2e37b282cb6570760e9c837599dd51f239ca1 When a private key is validated and there is no private key, return early.
90113096186e185cd07ee1c4c7267d0e68e67793 Add a test for EVP_PKEY_*_check functions for "DSA" keys

I took the liberty to edit the first commit message to this:

```
    When a private key is validated and there is no private key, return early.
    
    Affected functions:
    
    dsa_validate_public
    dsa_validate_private
    dh_validate_public
    dh_validate_private
```

PR #11599, opened 2020-04-21
Add support for unusal 'othername' subjectAltNames

dirkx:
Increasingly certificates seem to have special things in the subjectAltName that have arbitrary strings in them. Right now these print as the not very useful othername:<unsupported>. Which is a bit annoying as you often want to check these values / use them in webservers for client auth.

CE.g. some (now) common in EU export certificates and, for a few years now, certificates issued to medical doctors (in for example the netherlands, the full spec is https://www.uziregister.nl/Media/Default/PDF/20200325%20CA%20model%20pasmodel%20certificaatprofielen%20v10_0.pdf, section 4.8, page 16 for an example of one country).

And example of such a certificate has been added to test (this one mimics a practitioner who is licensed to practice medicine as a GP). The customs-export certificates are much the same; an IA5String containing the EORI number (EU tax number).

##### Checklist


crypto/x509/v3_alt.c
```diff
@@ -121,6 +121,17 @@ STACK_OF(CONF_VALUE) *i2v_GENERAL_NAME(X509V3_EXT_METHOD *method,
                 return NULL;
             break;
         default:
+             /* check if the value is something printable */
+            if (gen->d.otherName->value->type == V_ASN1_IA5STRING)
+                 if (X509V3_add_value("othername",
+                             gen->d.otherName->value->value.ia5string->data,
+                             &ret))
+                 return ret;
```
beldmit, line 129: Nit: alignment. 
Id' also used curved brackets around the internal if.

crypto/x509/v3_alt.c
```diff
@@ -121,6 +121,17 @@ STACK_OF(CONF_VALUE) *i2v_GENERAL_NAME(X509V3_EXT_METHOD *method,
...
+            if (gen->d.otherName->value->type == V_ASN1_UTF8STRING)
+                 if (X509V3_add_value("othername",
+                             gen->d.otherName->value->value.utf8string->data,
+                             &ret))
+                 return ret;
```
beldmit, line 134: Same here.

crypto/x509/v3_alt.c
```diff
@@ -121,6 +121,17 @@ STACK_OF(CONF_VALUE) *i2v_GENERAL_NAME(X509V3_EXT_METHOD *method,
                 return NULL;
             break;
         default:
+             /* check if the value is something printable */
+            if (gen->d.otherName->value->type == V_ASN1_IA5STRING)
+                 if (X509V3_add_value("othername",
```
t8m, line 126: The Travis failure is relevant. Please use X509V3_add_value_uchar(). Also you're losing the information about the type_id - perhaps using OBJ_obj2txt() to add it would make sense?

dirkx: understood. fixing. 

Not using the suggested OBJ_OBJ_obj2txt() as these particular strings are reall that- textual strings which we want to regex/match on; for example in an Apache/webserver TLS lockdown or when doing a RFC 3161 signature. And you'd end up having to match on a <ascii-ordinal>.<ascii-ordinal>.<32><2E> .. etc 

The latter which is not really openSSL its fault - the design behind these othernames is abit silly (encoding OIDs as ascii strings that is and then ascii-concatenating them with a dash)

crypto/x509/v3_alt.c
dirkx: done

crypto/x509/v3_alt.c
dirkx: done. Sorry !

crypto/x509/v3_alt.c
t8m: But I am talking about the type_id not about the value itself. I do not understand - is the information in type_id duplicated in the value itself?

dirkx: In this case it is devoid of info; in the case of the customs document it is just '.1/ISO'. Let me check a few others to see if there is a useful value sometimes.

crypto/x509/v3_alt.c
```diff
@@ -121,6 +121,21 @@ STACK_OF(CONF_VALUE) *i2v_GENERAL_NAME(X509V3_EXT_METHOD *method,
                 return NULL;
             break;
         default:
+             /* check if the value is something printable */
+            if (gen->d.otherName->value->type == V_ASN1_IA5STRING) {
+                 if (X509V3_add_value_uchar("othername",
```
beldmit, line 126: Nit: should be shifted 1 pos left

crypto/x509/v3_alt.c
```diff
@@ -121,6 +121,21 @@ STACK_OF(CONF_VALUE) *i2v_GENERAL_NAME(X509V3_EXT_METHOD *method,
...
+                             gen->d.otherName->value->value.ia5string->data,
+                             &ret)) {
+                     return ret;
```
beldmit, line 129: Nit: Here we don't need curved brackets, as it is one-line term

crypto/x509/v3_alt.c
dirkx: Ok - so I found some other cases where it is useful. Below would work generic enough; and with a bit of a fiddle with snprintf to leave the `othername:` in place.

```
@@ -79,6 +79,7 @@ STACK_OF(CONF_VALUE) *i2v_GENERAL_NAME(X509V3_EXT_METHOD *method,
                                        STACK_OF(CONF_VALUE) *ret)
 {
     unsigned char *p;
+    char othername[256];
     char oline[256], htmp[5];
     int i;
 
@@ -121,22 +122,27 @@ STACK_OF(CONF_VALUE) *i2v_GENERAL_NAME(X509V3_EXT_METHOD *method,
                 return NULL;
             break;
         default:
+             if (OBJ_obj2txt(oline, sizeof(oline), gen->d.otherName->type_id, 0)) 
+                 snprintf(othername, sizeof(othername), "othername: %s:", oline);
+             else
+                 strncpy(othername, "othername", sizeof(othername));
+
              /* check if the value is something printable */
             if (gen->d.otherName->value->type == V_ASN1_IA5STRING) {
-                 if (X509V3_add_value_uchar("othername",
+                 if (X509V3_add_value_uchar(othername,
                              gen->d.otherName->value->value.ia5string->data,
                              &ret)) {
                      return ret;
                 };
             };
             if (gen->d.otherName->value->type == V_ASN1_UTF8STRING) {
-                if (X509V3_add_value_uchar("othername",
+                if (X509V3_add_value_uchar(othername,
                              gen->d.otherName->value->value.utf8string->data,
                              &ret)) {
                      return ret;
                 };
             };
-            if (!X509V3_add_value("othername", "<unsupported>", &ret))
+            if (!X509V3_add_value(othername, "<unsupported>", &ret))
                 return NULL;
             break;
         }
```

as that would pretty much remove the need for all of them (as those OIDs nicely resolve through OBJ_obj2txt).
And this preserves the double :: of the NAIRealm, UNP and others above it.

But I am wondering if it makes sense to also make the same change to all those cases

```
  case NID_ms_upn:
            if (gen->d.otherName->value->type != V_ASN1_UTF8STRING
                    || !X509V3_add_value_uchar("othername: UPN:",
                            gen->d.otherName->value->value.utf8string->data,
                            &ret))
                return NULL;
            break;
   
```

Is that in line with what you where thinkng ?

dirkx: It would just remove the need for that whole case statement ?

t8m: It would change the output for the UPN and SmtpUTF8Mailbox so these would have to be probably special-cased. However yes, it could be simplified like this.

dirkx: fair point - and I found one Linux "API" in the /etc/cloud which then breaks. So lets not do this.

kaduk:
I suggest a more nuanced subject to the commit message, perhaps "Improve support for" or "Add printing support for some SANs" or such.

dirkx:
@kaduk changes it to `Add support for unusal 'othername' subjectAltNames` and squashed the commits.

crypto/x509/v3_alt.c
```diff
@@ -121,7 +122,25 @@ STACK_OF(CONF_VALUE) *i2v_GENERAL_NAME(X509V3_EXT_METHOD *method,
...
+            else
+                strncpy(othername, "othername", sizeof(othername));
+
+            /* check if the value is something printable */
+            if (gen->d.otherName->value->type == V_ASN1_IA5STRING) {
+                if (X509V3_add_value_uchar(othername,
+                             gen->d.otherName->value->value.ia5string->data,
+                             &ret)) 
+                    return ret;
+            };
```
FdaSilvaYY, line 136: useless semi-comma after }.

May cause failure of some pedantic builds.

dirkx-reno-2020: done.

crypto/x509/v3_alt.c
```diff
@@ -121,7 +122,25 @@ STACK_OF(CONF_VALUE) *i2v_GENERAL_NAME(X509V3_EXT_METHOD *method,
                 return NULL;
             break;
         default:
-            if (!X509V3_add_value("othername", "<unsupported>", &ret))
+            if (OBJ_obj2txt(oline, sizeof(oline), gen->d.otherName->type_id, 0))
```
t8m, line 125: OBJ_obj2txt can return negative value on error. So please use > 0.

crypto/x509/v3_alt.c
```diff
@@ -121,7 +122,25 @@ STACK_OF(CONF_VALUE) *i2v_GENERAL_NAME(X509V3_EXT_METHOD *method,
                 return NULL;
             break;
         default:
-            if (!X509V3_add_value("othername", "<unsupported>", &ret))
+            if (OBJ_obj2txt(oline, sizeof(oline), gen->d.otherName->type_id, 0)) 
+                snprintf(othername, sizeof(othername), "othername: %s:", oline);
+            else
+                strncpy(othername, "othername", sizeof(othername));
```
t8m, line 128: Why not `"othername:"`?

dirkx-reno-2020: well - because it looked nicer. But you are right - it is not identical to the old behaviour. fixed.

crypto/x509/v3_alt.c
dirkx-reno-2020: fixed

t8m:
@beldmit please re-approve

crypto/x509/v3_alt.c
```diff
@@ -121,7 +122,25 @@ STACK_OF(CONF_VALUE) *i2v_GENERAL_NAME(X509V3_EXT_METHOD *method,
                 return NULL;
             break;
         default:
-            if (!X509V3_add_value("othername", "<unsupported>", &ret))
+            if (OBJ_obj2txt(oline, sizeof(oline), gen->d.otherName->type_id, 0)>0)
```
t8m, line 125: Oh, I am sorry - a formatting nit here - spaces around >

dirkx-reno-2020: done

dirkx-reno-2020: fixed - and assumed you wanted it squashed with the other to typo changes ?

beldmit:
Merged. Thanks!

mattcaswell:
I'm getting this error after this PR:

````
crypto/x509/v3_alt.c: In function 'i2v_GENERAL_NAME':
crypto/x509/v3_alt.c:129:68: error: '%s' directive output may be truncated writing up to 255 bytes into a region of size 245 [-Werror=format-truncation=]
  129 |                snprintf(othername, sizeof(othername), "othername: %s:", oline);
      |                                                                   ^~    ~~~~~

crypto/x509/v3_alt.c:129:17: note: 'snprintf' output between 13 and 268 bytes into a destination of size 256
  129 |                 snprintf(othername, sizeof(othername), "othername: %s:", oline);
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
````

levitte:
@mattcaswell, fixed in #11656

PR #11600, opened 2020-04-21
WIP: Additional tests for CVE-2020-1967

kaduk:
Add some tests for the client-side part of CVE-2020-1967.
In order to make them work, some bugfixes and improvements to TLSProxy are also necessary.
Furthermore, while the bug was specific to TLS 1.3, the scenarios being tested should succeed regardless of TLS version, so remove the specificity from the tests.

This is WIP because the last commit reverts a change that was needed for 3.0 (making things libctx-aware) but the tests fail without it.  When using an Ed25519 client cert (for `SSL_check_chain()`), the computation of public key from private key produces an incorrect result and so the client fails the handshake for reasons unrelated to the test scenario, but none of the relevant code in https://github.com/openssl/openssl/pull/11371/commits/bde75d75028ce551d383d0ec9d3275836ffd27f7 seems obviously erroneous to me.

##### Checklist

### DIFF
--- a/crypto/ec/curve25519.c
+++ b/crypto/ec/curve25519.c
@@ -5582,18 +5582,8 @@ int ED25519_public_from_private(OPENSSL_CTX *ctx, uint8_t out_public_key[32],
 {
     uint8_t az[SHA512_DIGEST_LENGTH];
     ge_p3 A;
-    int r;
-    EVP_MD *sha512 = NULL;
 
-    sha512 = EVP_MD_fetch(ctx, SN_sha512, NULL);
-    if (sha512 == NULL)
-        return 0;
-    r = EVP_Digest(private_key, 32, az, NULL, sha512, NULL);
-    EVP_MD_free(sha512);
-    if (!r) {
-        OPENSSL_cleanse(az, sizeof(az));
-        return 0;
-    }
+    SHA512(private_key, 32, az);
 
     az[0] &= 248;
     az[31] &= 63;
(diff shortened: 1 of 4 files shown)


kaduk:
ccing @paulidale who wrote the bits that are apparently failing for me, in case he has any insight

kaduk:
Hmm, the travis log is a little odd, but suggests that the client-side tests do need skipping, but for no-ec rather than no-tls13

mattcaswell:
> This is WIP because the last commit reverts a change that was needed for 3.0 (making things libctx-aware) but the tests fail without it. When using an Ed25519 client cert (for SSL_check_chain()), the computation of public key from private key produces an incorrect result and so the client fails the handshake for reasons unrelated to the test scenario, but none of the relevant code in bde75d7 seems obviously erroneous to me.

I think this is actually a TLSProxy problem. The TLSProxy uses the ossltest engine to replace certain algorithm primitives with test versions. With digests, the digests always give the same answer no matter what you feed to them. This is so that we have the ability to modify the handshake without the handshake actually failing during the Finished message check. One of the digests it modifies is SHA512.

The previous Ed25519 codepath before bde75d7502 used the low level SHA512 API which completely avoids the use of an Engine. After bde75d7502 we are using EVP which spots that we have an Engine in place for SHA512 and routes it to ossltest...which obviously gives the wrong result.

kaduk:
> I think this is actually a TLSProxy problem

that makes a lot of sense!  I can investigate further given that hint (and should probably close #11621 )

t8m:

t8m:

PR #11601, opened 2020-04-22
Use the param modified sentinel in existing code.

paulidale:
Update existing code so that the param modified sentinel is using instead of fiddling with the return size directly.

Also added the sentinel to the param builder.

### DIFF
--- a/crypto/evp/evp_lib.c
+++ b/crypto/evp/evp_lib.c
@@ -76,13 +76,15 @@ int EVP_CIPHER_param_to_asn1(EVP_CIPHER_CTX *c, ASN1_TYPE *type)
             goto err;
 
         /* ... but, we should get a return size too! */
-        if (params[0].return_size != 0
+        if (OSSL_PARAM_modified(params)
+            && params[0].return_size != 0
             && (der = OPENSSL_malloc(params[0].return_size)) != NULL) {
             params[0].data = der;
             params[0].data_size = params[0].return_size;
-            params[0].return_size = 0;
+            OSSL_PARAM_set_all_unmodified(params);
             derp = der;
             if (EVP_CIPHER_CTX_get_params(c, params)
+                && OSSL_PARAM_modified(params)
                 && d2i_ASN1_TYPE(&type, (const unsigned char **)&derp,
                                  params[0].return_size) != NULL) {
                 ret = 1;
--- a/crypto/params.c
+++ b/crypto/params.c
@@ -45,10 +45,11 @@ int OSSL_PARAM_modified(const OSSL_PARAM *p)
     return p != NULL && p->return_size != OSSL_PARAM_UNMODIFIED;
 }
 
-void OSSL_PARAM_set_unmodified(OSSL_PARAM *p)
+void OSSL_PARAM_set_all_unmodified(OSSL_PARAM *p)
 {
     if (p != NULL)
-        p->return_size = OSSL_PARAM_UNMODIFIED;
+        while (p->key != NULL)
+            p++->return_size = OSSL_PARAM_UNMODIFIED;
 }
 
 int OSSL_PARAM_get_int(const OSSL_PARAM *p, int *val)
--- a/test/params_api_test.c
+++ b/test/params_api_test.c
@@ -69,7 +69,11 @@ static int test_param_type_extra(OSSL_PARAM *param, const unsigned char *cmp,
     const int sizet = bit32 && sizeof(size_t) > sizeof(int32_t);
     const int signd = param->data_type == OSSL_PARAM_INTEGER;
 
-    OSSL_PARAM_set_all_unmodified(param);
+    /* 
+     * Set the unmodified sentinal directly because there is no param array
+     * for these tests.
+     */
+    param->return_size = OSSL_PARAM_UNMODIFIED;
     if (signd) {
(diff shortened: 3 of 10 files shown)


paulidale:
#11576 is open and is the reason for this.  I didn't notice the other ones going in or I'd have complained then :)

All instances which might be setting the return size can be found using: `grep -r 'return_size *=[^=]'` and I'd missed two which are now fixed :(

crypto/evp/evp_lib.c
```diff
@@ -76,13 +76,15 @@ int EVP_CIPHER_param_to_asn1(EVP_CIPHER_CTX *c, ASN1_TYPE *type)
...
 
         /* ... but, we should get a return size too! */
-        if (params[0].return_size != 0
+        if (OSSL_PARAM_modified(params)
+            && params[0].return_size != 0
             && (der = OPENSSL_malloc(params[0].return_size)) != NULL) {
             params[0].data = der;
             params[0].data_size = params[0].return_size;
-            params[0].return_size = 0;
+            OSSL_PARAM_set_unmodified(params);
```
levitte, line 84: Style nit: I think the intention is clearer if you make the argument `&params[0]`

paulidale: Ugh.

paulidale: Really???  I'd prefer to change the previous lines to `*params`.

levitte: well, when I first saw `OSSL_PARAM_set_unmodified(params)`, the *first* thing that came to my mind was "wait what?  Is that function modifying the whole params array?"

Unfortunately, this is a fallacy in C, it's unclear if a passed pointer of an array is meant to just affect the first element or the whole array.

Maybe I'm overthinking this and it's just a matter of remembering what the function does...

paulidale: It could modify the whole array if we want....

I think that's overkill but might be easier for users.

paulidale: I'm warming to the idea of `OSSL_PARAM_set_unmodified` resetting the whole params array.

richsalz: There's no need for this function, since you're already directly setting three of the fields.

crypto/params.c
```diff
@@ -47,8 +47,8 @@ int OSSL_PARAM_modified(const OSSL_PARAM *p)
 
 void OSSL_PARAM_set_unmodified(OSSL_PARAM *p)
 {
-    if (p != NULL)
-        p->return_size = OSSL_PARAM_UNMODIFIED;
+    while (p != NULL)
```
t8m, line 50: It should be p->key != NULL

t8m: or `p != NULL && p->key != NULL`

crypto/evp/evp_lib.c
levitte: You do understand, I hope, that I have the same hiccup at `OSSL_PARAM_modified(param)`...

I'm not sure it's wise to have an API where the related functions behave so very differently.  `OSSL_PARAM_set_modified()` and `OSSL_PARAM_modified()` should work in tandem, that's simply less confusing API design.

paulidale: It doesn't seem like there is a win here.  Do it one way, complaints.  Do it the other, complaints.

I'll change the API to `OSSL_PARAM_set_all_unmodified()` which makes it clearer.

crypto/params.c
paulidale: Lesson to self: don't try to rework code when very tired.
It's never stuck before, I doubt it will this time either :(

Thanks.

paulidale: Fixed now.

crypto/evp/evp_lib.c
levitte: Yeah well, this got to bigger proportions than I had anticipated for a simple style nit. I wish I had shut the hell up...

paulidale: We all wish that at times :)

I think the new call is an improvement nonetheless.

levitte: Okie.  At the very least, it cleared some confusion :wink:

test/params_api_test.c
```diff
@@ -69,7 +69,7 @@ static int test_param_type_extra(OSSL_PARAM *param, const unsigned char *cmp,
     const int sizet = bit32 && sizeof(size_t) > sizeof(int32_t);
     const int signd = param->data_type == OSSL_PARAM_INTEGER;
 
-    OSSL_PARAM_set_unmodified(param);
+    OSSL_PARAM_set_all_unmodified(param);
```
levitte, line 72: The flip side is that `param` now *must* be a properly terminated OSSL_PARAM array.  Because `test_param_int()` makes a singular OSSL_PARAM, this call barfs all over Travis for the moment...

levitte: You could revert this back to `param->return_size = OSSL_PARAM_UNMODIFIED`...

paulidale: I've done this with a comment and added a test that uses the API call properly on an array.

paulidale:
Merged to master after removing a single end of line space.

Thanks for the very constructive input.  I agree, this is much better than when it began.

PR #11602, opened 2020-04-22
FIPS: remove algorithms that are not being validated.

paulidale:
Several MACs and one KDF are included in the FIPS provider with the property
"fips=yes" set but are not listed as being part of the OpenSSL validation.

This removes them from the FIPS provider.

### DIFF
--- a/providers/fips/fipsprov.c
+++ b/providers/fips/fipsprov.c
@@ -367,12 +367,6 @@ static const OSSL_ALGORITHM fips_digests[] = {
     { "SHA3-256", "provider=fips,fips=yes", sha3_256_functions },
     { "SHA3-384", "provider=fips,fips=yes", sha3_384_functions },
     { "SHA3-512", "provider=fips,fips=yes", sha3_512_functions },
-    /*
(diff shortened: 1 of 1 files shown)


paulidale:
I've amended this to also include KMAC.
Re-review required.

slontis:
Hmm SSKDF supports hash, HMAC or KMAC.. (Guess it is ok to only do a subset)

paulidale:
I don't see much point including KMAC for SSKDF in the FIPS provider unless it (KMAC) is validated.

richsalz:
Can you explain why "fips=no" doesn't work?  And how does just editing the table *remove* things from the provider?  I'm not questioning that this is a way to do it, but I think that eventually we will need the kind of documentation that I am seeking here.  Thanks.

paulidale:
"fips=no" might work but it increases the size of the FIPS provider.  The ecx are only in because the sponsors' expected and insisted it was.

Taking them out of the table removes them because of `ld` doing its job.

mattcaswell:
Pushed!

PR #11607, opened 2020-04-22
travis: add GENERATE=yes to the ARM64 build.

paulidale:
This fixes builds that change the `libcrypto.num` and other generated and compared files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ env:
     # Note: env entry here must exactly match the value in the exclude: table below that contains env:, otherwise it will not find a match.
     - CONFIG_OPTS="" DESTDIR="_install"
     - CONFIG_OPTS="no-asm -Werror --debug no-afalgeng no-shared enable-rc5 enable-md2 -fsanitize=address" LSAN_OPTIONS="report_objects=1"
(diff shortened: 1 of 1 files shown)


t8m:
Actually I am not sure this is right. The build without GENERATE=yes seems to be useful as we need the changes to the generated files to be committed eventually, no? We cannot drop them from git because they are needed on non-unix platforms to build.

paulidale:
The changes to `libcrypto.num` are committed in #11082 but this build fails, that's what confusing me.

t8m:
Ah, you're right. Actually it does not make sense to run no-deprecated with CHECKDOCS="yes" without GENERATE="yes".

t8m:
But still it might be useful to have some run with CHECKDOCS and without GENERATE. So what about dropping GENERATE="yes" from lines 35 and 44?4

paulidale:
I've removed it on those two, let's see what the CIs think.

t8m:
Ah, no-makedepend will be needed for the ARM build.

paulidale:
Ping for review?

paulidale:
Merged to master, thanks.

PR #11610, opened 2020-04-23
Add documentation for core_names.h fields and update code based on usage

slontis:
##### Checklist


slontis:
Still quit a bit to do.

doc/man7/provider-base.pod
```diff
@@ -220,22 +220,52 @@ useless without at least provider_query_operation(), and
 provider_gettable_params() is fairly useless if not accompanied by
 provider_get_params().
 
+=head2 Provider parameters
+
+provider_get_params() must return the following provider parameters to the core.
```
paulidale, line 225: _must_ or _should_ or _can_ ?

The name is mandatory, the rest might not be.

doc/man7/provider-base.pod
```diff
@@ -220,22 +220,52 @@ useless without at least provider_query_operation(), and
...
+
+=over 4
+
+=item "name" (B<OSSL_PROV_PARAM_NAME>) <UTF8_ptr>
+
+This points to a string that should give a unique description of the provider.
```
paulidale, line 231: s/description of/name for/

doc/man7/provider-base.pod
```diff
@@ -220,22 +220,52 @@ useless without at least provider_query_operation(), and
...
+
+=item "version" (B<OSSL_PROV_PARAM_VERSION>) <UTF8_ptr>
+
+This points to a string that is a version number associated with this provider.
+OpenSSL in-built providers use OPENSSL_VERSION_STR, but this may be different
+for any third party provider.
```
paulidale, line 237: What about adding: _This string is only printed and the format used for the version is left to the provider._ ?

doc/man7/provider-base.pod
```diff
@@ -220,22 +220,52 @@ useless without at least provider_query_operation(), and
...
+
+=item "buildinfo" (B<OSSL_PROV_PARAM_BUILDINFO>) <UTF8_ptr>
+
+This points to a string that is a build number associated with this provider.
```
paulidale, line 241: s/number/???? _I don't know, but it's not a number_ ????/

doc/man7/provider-base.pod
```diff
@@ -220,22 +220,52 @@ useless without at least provider_query_operation(), and
...
+OpenSSL in-built providers use OPENSSL_FULL_VERSION_STR, but this may be
+different for any third party provider.
+
+=back 
+
+provider_gettable_params() should contain the above parameter types.
```
paulidale, line 247: s/contain/return/
s/parameter types/parameters/

doc/man7/provider-base.pod
```diff
@@ -220,22 +220,52 @@ useless without at least provider_query_operation(), and
...
-OpenSSL libraries' full version string, i.e. the string expanded from
-the macro B<OPENSSL_VERSION_STR>.
+This points to the OpenSSL libraries' idea of what the calling provider is named.
 
-=item "provider-name"
+=item "module-filename" (B<OSSL_PROV_PARAM_CORE_MODULE_FILENAME>) <UTF8_ptr>
 
-This is a B<OSSL_PARAM_UTF8_PTR> type of parameter, pointing at the
-OpenSSL libraries' idea of what the calling provider is called.
+This points to a string containing the full filename of the providers
```
paulidale, line 267: s/providers/provider's/

Is this a file name or a path or either?

doc/man7/provider-base.pod
levitte: I wonder if these is a good idea.  A provider author can't necessarily know what other providers there are.  It would probably make more sense to use the module full path on the libcrypto side of things.

doc/man7/provider-base.pod
levitte: It can be anything the provider author choses.  It *could* be a build number, but that's not necessarily the case.

doc/man7/provider-base.pod
slontis: It opens this - so its a filename.

slontis:
Ready for feedback.

slontis:
ping

doc/man7/EVP_PKEY-EC.pod
```diff
@@ -2,7 +2,11 @@
 
 =head1 NAME
 
-EVP_PKEY-EC - EVP_PKEY EC keytype and algorithm support
+EVP_PKEY-EC,
+EVP_PKEY-ECDH,
+EVP_PKEY-ECDSA,
```
levitte, line 7: Er...  as far as I know, there is no ECDH or ECDSA key type, so this sounds plain wrong to me.  It may be that certain curves are better suited for use with DH while others are better suited for DSA, but that's a different matter.  Whether a discussion on that should be in `EVP_PKEY-EC(7)` or `EVP_KEYEXCH-ECDH(7)` / `EVP_SIGNATURE-ECDSA(7)`, I cannot really say, but please do not claim in this way that there are separate ECDH and ECDSA keys, that's a rabbit hole OpenSSL got *out* of ages ago.

doc/man7/EVP_PKEY-FFC.pod
```diff
@@ -0,0 +1,202 @@
+=pod
+
+=head1 NAME
+
+EVP_PKEY-FFC - EVP_PKEY FFC key type and algorithm support
```
levitte, line 5: FFC isn't a key type per se.  It's not really anything tangible in this way on it's own, it's a component of the key types DH and DSA, so should rather be presented as such.  I'm not at all sure what to do with this, naming wise.  It possible that we should leave it as `FFC(7)`, but I'm unsure...

doc/man7/provider-keyexch.pod
```diff
@@ -152,79 +152,27 @@ See L<OSSL_PARAM(3)> for the use of B<OSSL_PARAM> as parameter descriptor.
...
 
 See L<OSSL_PARAM(3)> for further details on the parameters structure used by
 the OP_keyexch_set_ctx_params() and OP_keyexch_get_ctx_params() functions.
 
-Parameters currently recognised by built-in key exchange algorithms are as
-follows.
-Not all parameters are relevant to, or are understood by all key exchange
-algorithms:
+Common parameters currently recognised by built-in key exchange algorithms are
+as follows.
```
levitte, line 161: You should do this rephrase in `provider-signature.pod` as well...

levitte:
Apart from the concerns I just mentioned, this is great work.  I'm especially happy to see the `EVP_SIGNATURE` and `EVP_KEYEXCH` pages appear.  Yay :-)

doc/man7/EVP_PKEY-FFC.pod
slontis: What if I call it something more like
EVP_PKEY-DH_and_DSA_FFC_params.
And then change the description?

doc/man7/EVP_PKEY-EC.pod
slontis: Removed the tags.

doc/man7/EVP_PKEY-FFC.pod
slontis: I have just changed the description for now.

slontis:
rebased

slontis:
another rebase

slontis:
travis error is s390 resource unavailable..

crypto/ec/ec_backend.c
```diff
@@ -53,6 +48,15 @@ int ec_set_param_ecdh_cofactor_mode(EC_KEY *ec, const OSSL_PARAM *p)
...
 
+int ec_set_param_ecdh_cofactor_mode(EC_KEY *ec, const OSSL_PARAM *p)
+{
+    int mode;
+
+    if (!OSSL_PARAM_get_int(p, &mode))
+        return 0;
+    return ec_set_ecdh_cofactor_mode(ec, mode);
+}
+
```
levitte, line 59: Instead of adding this small thing, may I suggest this?

``` diff
diff --git a/crypto/ec/ec_backend.c b/crypto/ec/ec_backend.c
index 98dd0ecf5d..73dd4d5903 100644
--- a/crypto/ec/ec_backend.c
+++ b/crypto/ec/ec_backend.c
@@ -224,8 +220,12 @@ int ec_key_otherparams_fromdata(EC_KEY *ec, const OSSL_PARAM params[])
         return 0;
 
     p = OSSL_PARAM_locate_const(params, OSSL_PKEY_PARAM_USE_COFACTOR_ECDH);
-    if (p != NULL && !ec_set_param_ecdh_cofactor_mode(ec, p))
-        return 0;
+    if (p != NULL) {
+        int mode;
 
+        if (!OSSL_PARAM_get_int(p, &mode)
+            || !ec_set_param_ecdh_cofactor_mode(ec, p))
+            return 0;
+    }
     return 1;
 }
diff --git a/providers/implementations/keymgmt/ec_kmgmt.c b/providers/implementations/keymgmt/ec_kmgmt.c
index a48b279547..bc66c7c1e4 100644
--- a/providers/implementations/keymgmt/ec_kmgmt.c
+++ b/providers/implementations/keymgmt/ec_kmgmt.c
@@ -542,14 +542,7 @@ const OSSL_PARAM *ec_settable_params(void)
 static
 int ec_set_params(void *key, const OSSL_PARAM params[])
 {
-    EC_KEY *eck = key;
-    const OSSL_PARAM *p;
-
-    p = OSSL_PARAM_locate_const(params, OSSL_PKEY_PARAM_USE_COFACTOR_ECDH);
-    if (p != NULL && !ec_set_param_ecdh_cofactor_mode(eck, p))
-        return 0;
-
-    return 1;
+    return ec_key_otherparams_fromdata(key, params);
 }
 
 static
```
```

File: doc/man3/EVP_KDF.pod [hunk 11]
```diff
@@ -66,6 +66,53 @@ B<EVP_KDF_CTX> is a context type that holds the algorithm inputs.
 
 EVP_KDF_fetch() fetches an implementation of a KDF I<algorithm>, given
 a library context I<libctx> and a set of I<properties>.
+
+The OpenSSL default provider supports the following algorithms:
```
levitte, line 70: This adds a duplicate of the lists that are already in `doc/man7/OSSL_PROVIDER-*.pod`.  I would rather just have a reference, like this:

```
See L<OSSL_PROVIDER-default.pod/Key Derivation Function (KDF)> for the lists of
algorithms supported by the default provider.
```

doc/man3/EVP_MAC.pod
```diff
@@ -80,6 +80,45 @@ rely on an underlying computation algorithm.
 
 EVP_MAC_fetch() fetches an implementation of a MAC I<algorithm>, given
 a library context I<libctx> and a set of I<properties>.
+
+The OpenSSL default provider supports the following algorithms:
```
levitte, line 84: This adds a duplicate of the lists that are already in `doc/man7/OSSL_PROVIDER-*.pod`.  I would rather just have a reference, like this:

```
See L<OSSL_PROVIDER-default.pod/Message Authentication Code (MAC)> for the lists of
algorithms supported by the default provider.
```

doc/man7/EVP_KEYEXCH-ECDH.pod
```diff
@@ -0,0 +1,138 @@
...
+
+=head1 NAME
+
+EVP_KEYEXCH-ECDH - ECDH Key Exchange algorithm support
+
+=head1 DESCRIPTION
+
+Key exchange support for the B<ECDH> key type.
+
+
```
levitte, line 11: Style: two empty lines?

doc/man7/EVP_KEYEXCH-ECDH.pod
```diff
@@ -0,0 +1,138 @@
...
+B<OSSL_PKEY_PARAM_USE_COFACTOR_ECDH> parameter that can be set on a
+per-key basis.
+
+=item "kdf-type" (B<OSSL_EXCHANGE_PARAM_KDF_TYPE>) <utf8_string>
+
+Sets or gets the Key Derivation Function type to apply within the associated key
+exchange ctx.
+
+
+
```
levitte, line 41: *Three* empty lines?

doc/man7/EVP_KEYEXCH-ECDH.pod
```diff
@@ -0,0 +1,138 @@
...
+Key Derivation Function associated with the given key exchange ctx.
+
+=item "kdf-ukm-len" (B<OSSL_EXCHANGE_PARAM_KDF_UKM_LEN>) <size_t>
+
+Gets the size of the User Key Material to be used as part of the selected
+Key Derivation Function associated with the given key exchange ctx.
+
+=back
+
+
```
levitte, line 75: Two empty line?

doc/man7/EVP_PKEY-DH.pod
```diff
@@ -0,0 +1,231 @@
...
+Used for DH generation of safe primes using the old generator code.
+It is recommended to use a named safe prime group instead, if domain parameter
+validation is required. The default value is 2.
+
+These are not named safe prime groups so setting this value for the OpenSSL FIPS
+provider will instead choose a named safe prime group based on the size of I<p>.
+
+=back
+
+
```
levitte, line 59: Two empty lines?

doc/man7/EVP_PKEY-EC.pod
```diff
@@ -47,10 +48,67 @@ An B<EVP_PKEY> context can be obtained by calling:
...
+    EVP_PKEY_CTX_set_params(gctx, params);
+
+    EVP_PKEY_gen(gctx, &key);
+
+    EVP_PKEY_print_private(bio_out, key, 0, NULL);
+    ...
+    EVP_PKEY_free(key);
+    EVP_PKEY_CTX_free(gctx);
+
+An B<EVP_PKEY> ECCDH key can be generated with a "K-571" named group by calling:
```
levitte, line 73: "ECDH" should be spellt "ECDH" :wink:

doc/man7/EVP_PKEY-FFC.pod
```diff
@@ -0,0 +1,203 @@
...
+For B<DSA> (and B<DH> that is not a named group) the FIPS186-4 standard
+specifies that the values used for FFC parameter generation are also required
+for parameter validation.
+This means that optional FFC domain parameter values for I<seed>, I<pcounter>
+and I<gindex> may need to be stored for validation purposes.
+For B<DH> the I<seed> and I<pcounter> can be stored in ASN1 data
+(but the I<gindex> is not). For B<DSA> however, these fields are not stored in
+the ASN1 data so they need to be stored externally if validation is required.
+
+
```
levitte, line 30: More of two empty lines...

providers/implementations/signature/rsa.c
```diff
@@ -917,15 +915,18 @@ static int rsa_get_ctx_params(void *vprsactx, OSSL_PARAM *params)
...
             case RSA_PSS_SALTLEN_MAX:
-                if (!OSSL_PARAM_set_utf8_string(p, "max"))
+                if (!OSSL_PARAM_set_utf8_string(p,
+                        OSSL_PKEY_RSA_PSS_SALT_LEN_MAX))
                     return 0;
                 break;
             case RSA_PSS_SALTLEN_AUTO:
-                if (!OSSL_PARAM_set_utf8_string(p, "auto"))
+                if (!OSSL_PARAM_set_utf8_string(p,
+                        OSSL_PKEY_RSA_PSS_SALT_LEN_AUTO))
```
levitte, line 929: While I personally have nothing against this break of indentation style, I wonder what our current code style guide has to say about this, not to mention `util/check-format.pl`...  one way to get around that problem could be to refactor the code to this:

``` C
        } else if (p->data_type == OSSL_PARAM_UTF8_STRING) {
            const *value = NULL;

            switch (prsactx->saltlen) {
            case RSA_PSS_SALTLEN_DIGEST:
                value = OSSL_PKEY_RSA_PSS_SALT_LEN_DIGEST;
                break;
            case RSA_PSS_SALTLEN_MAX:
                value = OSSL_PKEY_RSA_PSS_SALT_LEN_MAX;
                break;
            case RSA_PSS_SALTLEN_AUTO:
                value = OSSL_PKEY_RSA_PSS_SALT_LEN_AUTO;
                break;
            default:
                if (BIO_snprintf(p->data, p->data_size, "%d", prsactx->saltlen)
                    <= 0)
                    return 0;
                break;
            }
            if (value != NULL
                && !OSSL_PARAM_set_utf8_string(p, value))
                return 0;
```

Side note: shouldn't the default case set `p->return_size`?

levitte:
I'm generally happy with what I see, there were just a few nits.  I'll make another sweep later today to see that I haven't missed something, but other than that and nit fixups, this should be good to go.

doc/man7/EVP_PKEY-EC.pod
slontis: For this example it is actually ECCDH not ECDH - The extra C is for the cofactor.

providers/implementations/signature/rsa.c
slontis: err.. this is setting utf8 strings - so I dont know what you mean :)

doc/man7/EVP_PKEY-EC.pod
slontis: Hmm maybe I am confused.. Looks like that c means computational.

What I am trying to refer to is that this example uses the 
Elliptic   Curve   Cryptography   Cofactor   Diffie-Hellman   (ECC   CDH) Primitive
As defined in https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf Section 5.7.1.2

providers/implementations/signature/rsa.c
slontis: I have updated to use value... Thanks.

doc/man7/EVP_PKEY-EC.pod
levitte: Gotcha...  although, "ECCDH" mostly looks like a kind of in-between, and doesn't "click" with me.  Maybe it would be better to actually write `ECC CDH`...

providers/implementations/signature/rsa.c
levitte: > err.. this is setting utf8 strings - so I dont know what you mean :)

`OSSL_PARAM_set_utf8_string()` sets `return_size`.  The default case, however, doesn't use that function at all, it goes directly at the raw data, but should still set `return_size`.  So, something like this?

``` C
            default:
                {
                    int len = BIO_snprintf(p->data, p->data_size, "%d", prsactx->saltlen);

                    if (len <= 0)
                        return 0;
                    p->return_size = len;
                }
                break;
            }
```

doc/man3/EVP_KDF.pod
```diff
@@ -248,14 +251,16 @@ supported by the KDF algorithm.
...
+L<EVP_KDF-KB(7)>
+L<EVP_KDF-KRB5KDF(7)>
+L<EVP_KDF-PBKDF2(7)>
+L<EVP_KDF-SCRYPT(7)>
 L<EVP_KDF-SS(7)>
 L<EVP_KDF-SSHKDF(7)>
-L<EVP_KDF-X963(7)>
+L<EVP_KDF-TLS1_PRF(7)>
 L<EVP_KDF-X942(7)>
+L<EVP_KDF-X963(7)>
```
levitte, line 263: So here's a question from a maintainability point of view...  should we have this duplicated list, or should we simply have `L<OSSL_PROVIDER-default(7)>`?

doc/man7/EVP_SIGNATURE-ED25519.pod
```diff
@@ -54,25 +49,32 @@ since version 1.1.1.
...
- EVP_PKEY_CTX *pctx = EVP_PKEY_CTX_new_id(EVP_PKEY_ED25519, NULL);
- EVP_PKEY_keygen_init(pctx);
- EVP_PKEY_keygen(pctx, &pkey);
- EVP_PKEY_CTX_free(pctx);
- PEM_write_PrivateKey(stdout, pkey, NULL, NULL, 0, NULL, NULL);
+    EVP_DigestSign(md_ctx, sig, &sig_len, msg, msg_len);
+    ...
+    OPENSSL_free(sig);
+    EVP_MD_CTX_free(md_ctx);
+    }
```
levitte, line 72: The indentation in this example looks horrible on github.  Could it be that there are tabs to be converted to 8 spaces?

doc/man7/EVP_SIGNATURE-ED25519.pod
```diff
@@ -54,25 +49,32 @@ since version 1.1.1.
 Valid algorithm names are B<ed25519>, B<ed448> and B<eddsa>. If B<eddsa> is
 specified, then both Ed25519 and Ed448 are benchmarked.
 
+
```
levitte, line 52: Remove this empty line

doc/man7/EVP_KEYEXCH-ECDH.pod
```diff
@@ -0,0 +1,134 @@
...
+                                                  ukm, ukm_len);
+    params[5] = OSSL_PARAM_construct_end();
+    EVP_PKEY_CTX_set_params(dctx, params);
+
+    EVP_PKEY_derive_set_peer(dctx, peer_pub_key);
+    EVP_PKEY_derive(dctx, secret, &secret_len);
+    ...
+    OPENSSL_clear_free(secret, secret_len);
+    EVP_PKEY_CTX_free(dctx);
+    }
```
levitte, line 114: Indentation is weird here too

doc/man7/EVP_KEYEXCH-DH.pod
```diff
@@ -0,0 +1,99 @@
...
+
+    /* Get the size by passing NULL as the buffer */
+    EVP_PKEY_derive(dctx, NULL, &secret_len);
+    secret = OPENSSL_zalloc(secret_len);
+
+    EVP_PKEY_derive(dctx, secret, &secret_len);
+    ...
+    OPENSSL_clear_free(secret, secret_len);
+    EVP_PKEY_CTX_free(dctx);
+    }
```
levitte, line 75: Indentation weird

doc/man7/EVP_PKEY-EC.pod
richsalz: I think "ECC DH" is more clear.  And "spellt" should be spelled like "spelt" (sic)

levitte: > I think "ECC DH" is more clear.

In what way?

> And "spellt" should be spelled like "spelt" (sic)

Ah!  Thanks

richsalz: In fact, I think it should be "Elliptic Curve Cofactor DH" since ECDH is common, but the openssl community is probably not used to cofactor.

levitte: Yeah, but this is confusing, as ECCDH could match Elliptic Curve Cryptography Diffie-Hellman just as much as Elliptic Curve Cofactor Diffie-Hellman.  And especially the separate "ECC" is, as far as I know, much more commonly known to mean Elliptic Curve Cryptography.  We *might* get away with "EC CDH", if you really want to get rid of one C...  after all, the cofactor thing is much more tied to the DH aspect (isn't it?)...

doc/man7/OSSL_PROVIDER-legacy.pod
```diff
@@ -20,7 +20,7 @@ defined:
```
richsalz, line 20: "this" property defined, not "these property"

doc/man7/provider-base.pod
```diff
@@ -234,22 +234,52 @@ useless without at least provider_query_operation(), and
 provider_gettable_params() is fairly useless if not accompanied by
 provider_get_params().
 
+=head2 Provider parameters
+
+provider_get_params() can return the following provider parameters to the core.
```
richsalz, line 239: why no colon here, but there is one at line 266.

doc/man7/EVP_PKEY-EC.pod
slontis: will change to EC CDH (Cofactor Diffie-Hellman)

doc/man3/EVP_KDF.pod
slontis: Changed to L<OSSL_PROVIDER-default(7)/Key Derivation Function (KDF)>

levitte: Works for me :-)

levitte:
This is good work, @slontis!  Thanks for your patience

slontis:
Cheers.. Thanks for reviewing - I hope to get better at documenting with less NITS at some point..

slontis:
Thanks. Merged to master.

PR #11612, opened 2020-04-23
FIPS: set the ECX algotihms to "fips=yes".

paulidale:
It seems likely that these can be vendor affirmed and if so, they are better being
tagged as FIPS approved.

### DIFF
--- a/providers/fips/fipsprov.c
+++ b/providers/fips/fipsprov.c
@@ -369,7 +369,7 @@ static const OSSL_ALGORITHM fips_digests[] = {
     { "SHA3-512", "provider=fips,fips=yes", sha3_512_functions },
 
     /* Non-FIPS algorithm to support oneshot_hash in the Ed448 code */
-    { "SHAKE-256:SHAKE256", "provider=fips,fips=no", shake_256_functions },
+    { "SHAKE-256:SHAKE256", "provider=fips,fips=yes", shake_256_functions },
     { NULL, NULL, NULL }
 };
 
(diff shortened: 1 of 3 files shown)


providers/fips/fipsprov.c
```diff
@@ -375,7 +375,7 @@ static const OSSL_ALGORITHM fips_digests[] = {
     { "KECCAK-KMAC-256:KECCAK-KMAC256", "provider=fips,fips=yes", keccak_kmac_256_functions },
 
     /* Non-FIPS algorithm to support oneshot_hash in the Ed448 code */
-    { "SHAKE-256:SHAKE256", "provider=fips,fips=no", shake_256_functions },
+    { "SHAKE-256:SHAKE256", "provider=fips,fips=yes", shake_256_functions },
```
mattcaswell, line 378: Are we also vendor affirming this one?

mattcaswell:
Should this be in alpha1? If so needs the urgent label.

providers/fips/fipsprov.c
paulidale: I think it will need to be validated -- this is a consequence of having Ed448.

At this stage, let's assume it is in.  We can remove it later if the lab disagrees.

paulidale:
Yes, I think it should be in the alpha.

mattcaswell:
> Yes, I think it should be in the alpha.

Agreed. I think there will be some updates required to the 3.0 wiki page as a result of this.

paulidale:
I'll update the wiki.

paulidale:
Wiki is done.

providers/fips/fipsprov.c
paulidale: I've emailed the lab to ask this specific question.

richsalz:
"It seems likely" is a weak justification.  We should (a) ask the lab; and (b) point this out very explicitly in the FIPS guide.  Since we don't have that yet, perhaps we need to collect issues for it?

paulidale:
The lab has responded and has answered in the affirmative.

The vendor affirmed will be very clearly stated in the security policy.  It doesn't need to go anywhere else because it will be FIPS approved and usable in FIPS mode.

paulidale:
This one was overlooked for the alpha :(

paulidale:
Merged to master.

PR #11614, opened 2020-04-23
Fix type cast in SSL_CTX_set1_groups macro

schwabe:
The macro casts the glist parameter to char*
instead of (int *) like the documentation of the function suggest.

Also the function tls1_set_groups that is called from SSL_CTX_ctrl
takes an int * argument. This looks like a copy&paste error from
SSL_CTX_set1_groups_list function.

CLA: trivial
Signed-off-by: Arne Schwabe <arne@rfc2549.org>


t8m:
Merged to master and 1.1.1.

Thank you for the PR.

PR #11615, opened 2020-04-23
Fix a typo in the makefile templates

mattcaswell:
The fips.so and legacy.so providers were not being installed because of
a typo in the makefile templates.


mattcaswell:
I'm marking  this as "urgent" because I believe this should be in the alpha 1 release today.

mattcaswell:
Pushed. Thanks.

PR #11626, opened 2020-04-23
Remove aes bitslice code

bernd-edlinger:
Remove bit-sliced CTR code since it is unclear
if this patent claim here is compatible with OpenSSL CLA

https://www.federalregister.gov/documents/2009/06/23/E9-14734/jointly-owned-invention-available-for-non-exclusive-royalty-free-licensing-for-advanced-encryption


richsalz:
The code was written by Andy Polyakov who licensed all his code under the CLA.
The NIST article gives rights to use the technique for AES, and that is the only place where the technique is used. Why do you think it's not acceptable?

bernd-edlinger:
Sorry, Rich, this section of code was written by me.
But I used an algorithm that was suggested by @davidben 
Unlike the non-bitsliced algorithm which is not patented
as far as I know, I would even claim I invented it myself.
The bitsliced cirquit is patented.
The patent appears to be held by the NSA.
I have no idea if the patent is free enough to be compatible
with the OpenSSL license.
But what I know is what I signed, in section 3 of my ICLA.
That is about patent issues, and I am not a lawyer, therefore
I must insist in removing that secution of code.
Because I cannot grant you a patent license that I do not own.

bernd-edlinger:
BTW: I am not sure if above link is not containing a VIRUS,
you know the NSA is your friend :-)

richsalz:
I mis-read the git history, thanks for the correction @bernd-edlinger

bernd-edlinger:
Does this need an approval ?
I just want to remove a contribution of my own,
I realized potential patent issues, I do that just
to protect you (@openssl/omc).

randombit:
Why not just use another sbox circuit, such as the one used in bsaes-x86_64.pl or as described in papers such as https://eprint.iacr.org/2011/332.pdf or https://www.iacr.org/archive/ches2005/032.pdf

crypto/evp/e_aes.c
```diff
@@ -130,11 +130,6 @@ void bsaes_xts_decrypt(const unsigned char *inp, unsigned char *out,
                        size_t len, const AES_KEY *key1,
                        const AES_KEY *key2, const unsigned char iv[16]);
 #endif
-#if !defined(AES_ASM) && !defined(AES_CTR_ASM)			\
-	&& defined(OPENSSL_AES_CONST_TIME)			\
-	&& !defined(OPENSSL_SMALL_FOOTPRINT)
-# define AES_CTR_ASM
-#endif
 #ifdef AES_CTR_ASM
```
levitte, line 133: Is this defined anywhere else?  If not, there's no point keeping this `#ifdef` either...

levitte:
I don't see that we *depend* on this code existing, so I can't see the harm with removing it on your request.

levitte:
I would like to see another approval beside mine.  If for nothing else, then for awareness.

levitte:
I removed the "urgent" label, 'cause it was misused here.  This PR is not *urgent* in that sense.

bernd-edlinger:
I am not a lawyer, but I am just concerned about my personal risk
or will you pay a lawyer to defend me if I am sued bythe patent owner ?

levitte:
I agree, @t8m, but I'm also considering what's the friendly thing to when someone feels bad about their code. This is a case, I think, where friendliness outweighs the interpretation of the law.

If this was code that we depended more heavily upon, we would have a different discussion, I'm sure.

bernd-edlinger:
Of course I could leave the code as is, if you pay my expenses in case
I get sued because of this perpetual thing in the  ICLA, I mean
lawyer and possible license fees I (or my apologies I don't know the English
word for it "Erben") will have to pay.
So if you want something perpetual from me, you need to offer
me something perpetual in return.

bernd-edlinger:
Merged.

Issue #11627, opened 2020-04-24
NULL dereference in EVP_PKEY_get_raw_private_key with provider side keys


matbech:
When EVP_PKEY_get_raw_private_key or EVP_PKEY_get_raw_public_key is called with a provider side key, there is a NULL dereference.

pkey->ameth for provider side keys is NULL:
https://github.com/openssl/openssl/blob/master/crypto/evp/p_lib.c#L384

A check for NULL could be added:
https://github.com/openssl/openssl/pull/11628

mattcaswell:
This highlighted a whole heap of problems with ECX provider keys. My fix for this issue and other related ones in #11635.

PR #11628, opened 2020-04-24
Fix NULL dereference in EVP_PKEY_get_raw_private_key

matbech:
When EVP_PKEY_get_raw_private_key or EVP_PKEY_get_raw_public_key is called with a provider side key, there is a NULL dereference.

Fixes https://github.com/openssl/openssl/issues/11627

### DIFF
--- a/crypto/evp/p_lib.c
+++ b/crypto/evp/p_lib.c
@@ -381,7 +381,8 @@ int EVP_PKEY_get_raw_private_key(const EVP_PKEY *pkey, unsigned char *priv,
                                  size_t *len)
 {
     /* TODO(3.0) Do we need to do anything about provider side keys? */
-     if (pkey->ameth->get_priv_key == NULL) {
+     if (pkey->ameth == NULL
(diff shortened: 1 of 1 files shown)


mattcaswell:
This fixes the NULL deref, but still means that EVP_PKEY_get_raw_private_key/EVP_PKEY_get_raw_public_key return an error for provider side keys. 

Probably we should just call `evp_keymgmt_export` for provider keys and make these APIs function correctly.

levitte:
> This fixes the NULL deref, but still means that EVP_PKEY_get_raw_private_key/EVP_PKEY_get_raw_public_key return an error for provider side keys.
> 
> Probably we should just call `evp_keymgmt_export` for provider keys and make these APIs function correctly.

EVP_PKEY_new_raw_private_key and EVP_PKEY_new_raw_public_key must be considered too in that case.  Please don't make just half a fix.

mattcaswell:
Thanks for your PR. This prompted me to dive into this deeper and I found numerous other issues. Alternative fix in #11635.

PR #11630, opened 2020-04-24
dev/release.sh: Add --reviewer to set reviewers

levitte:
Doing this is kind of contrary to how we normally do things, as this
constitutes a kind of pre-approval.  However, without this, the normal
review process will modify the reviewed commits, and render the
annotated release tag invalid, which forces the person doing the
release to re-tag manually.


levitte:
@mattcaswell, if you want this for the next alpha release, please review.

mattcaswell:
Pushed.

PR #11635, opened 2020-04-24
Fix raw provider keys

### LINKED ISSUE #11627: NULL dereference in EVP_PKEY_get_raw_private_key with provider side keys

mattcaswell:
The functions EVP_PKEY_get_raw_private_key and EVP_PKEY_get_raw_public_key fail with a NULL pointer deref if you pass a provider key to them. This also highlighted that EVP_PKEY_new_raw_private_key and EVP_PKEY_new_raw_public_key do not create provider side keys as you might expect.

This PR fixes both of those issues. Along the way a number of other issues were highlighted.

This is WIP because the tests are failing. At least one of those failures (possibly all of them) are because EVP_PKEY_cmp is not working with ECX provider side keys.

Fixes #11627

### DIFF
--- a/crypto/ec/ecx_backend.c
+++ b/crypto/ec/ecx_backend.c
@@ -9,6 +9,8 @@
 
 #include <openssl/core_names.h>
 #include <openssl/params.h>
+#include <openssl/ec.h>
+#include <openssl/err.h>
 #include "crypto/ecx.h"
 #include "ecx_backend.h"
 
@@ -17,6 +19,31 @@
  * for legacy backends (EVP_PKEY_ASN1_METHOD and EVP_PKEY_METHOD) and provider
  * implementations alike.
  */
+
+int ecx_public_from_private(ECX_KEY *key)
+{
+    switch (key->type) {
+    case ECX_KEY_TYPE_X25519:
+        X25519_public_from_private(key->pubkey, key->privkey);
+        break;
+    case ECX_KEY_TYPE_ED25519:
+        if (!ED25519_public_from_private(key->libctx, key->pubkey, key->privkey)) {
+            ECerr(0, EC_R_FAILED_MAKING_PUBLIC_KEY);
+            return 0;
+        }
+        break;
+    case ECX_KEY_TYPE_X448:
+        X448_public_from_private(key->pubkey, key->privkey);
+        break;
+    case ECX_KEY_TYPE_ED448:
+        if (!ED448_public_from_private(key->libctx, key->pubkey, key->privkey)) {
+            ECerr(0, EC_R_FAILED_MAKING_PUBLIC_KEY);
+            return 0;
+        }
+        break;
+    }
+    return 1;
+}
 
 int ecx_key_fromdata(ECX_KEY *ecx, const OSSL_PARAM params[],
                      int include_private)
@@ -32,11 +59,8 @@ int ecx_key_fromdata(ECX_KEY *ecx, const OSSL_PARAM params[],
     if (include_private)
         param_priv_key =
             OSSL_PARAM_locate_const(params, OSSL_PKEY_PARAM_PRIV_KEY);
-    /*
-     * If a private key is present then a public key must also be present.
-     * Alternatively we've just got a public key.
-     */
-    if (param_pub_key == NULL)
+
+    if (param_pub_key == NULL && param_priv_key == NULL)
         return 0;
 
     if (param_priv_key != NULL
@@ -46,13 +70,17 @@ int ecx_key_fromdata(ECX_KEY *ecx, const OSSL_PARAM params[],
         return 0;
 
     pubkey = ecx->pubkey;
-    if (!OSSL_PARAM_get_octet_string(param_pub_key,
-                                     (void **)&pubkey,
-                                     sizeof(ecx->pubkey), &pubkeylen))
+    if (param_pub_key != NULL
+        && !OSSL_PARAM_get_octet_string(param_pub_key,
+                                        (void **)&pubkey,
+                                         sizeof(ecx->pubkey), &pubkeylen))
         return 0;
 
-    if (pubkeylen != ecx->keylen
+    if ((param_pub_key != NULL && pubkeylen != ecx->keylen)
         || (param_priv_key != NULL && privkeylen != ecx->keylen))
+        return 0;
+
+    if (param_pub_key == NULL && !ecx_public_from_private(ecx))
         return 0;
 
     ecx->haspubkey = 1;
--- a/crypto/ec/ecx_backend.c
+++ b/crypto/ec/ecx_backend.c
@@ -48,7 +48,7 @@ int ecx_public_from_private(ECX_KEY *key)
 int ecx_key_fromdata(ECX_KEY *ecx, const OSSL_PARAM params[],
                      int include_private)
 {
-    size_t privkeylen = 0, pubkeylen;
+    size_t privkeylen = 0, pubkeylen = 0;
     const OSSL_PARAM *param_priv_key = NULL, *param_pub_key;
     unsigned char *pubkey;
 
--- a/crypto/ec/ecx_key.c
+++ b/crypto/ec/ecx_key.c
@@ -10,13 +10,14 @@
 #include <openssl/err.h>
 #include "crypto/ecx.h"
 
-ECX_KEY *ecx_key_new(ECX_KEY_TYPE type, int haspubkey)
+ECX_KEY *ecx_key_new(OPENSSL_CTX *libctx, ECX_KEY_TYPE type, int haspubkey)
 {
     ECX_KEY *ret = OPENSSL_zalloc(sizeof(*ret));
 
     if (ret == NULL)
         return NULL;
 
+    ret->libctx = libctx;
     ret->haspubkey = haspubkey;
     switch (type) {
     case ECX_KEY_TYPE_X25519:
--- a/crypto/evp/p_lib.c
+++ b/crypto/evp/p_lib.c
@@ -494,18 +494,61 @@ EVP_PKEY *EVP_PKEY_new_raw_public_key(int type, ENGINE *e,
     return new_raw_key_int(NULL, NULL, NULL, type, e, pub, len, 0);
 }
 
+struct raw_key_details_st
+{
+    unsigned char **key;
+    size_t *len;
+    int selection;
+};
+
+static OSSL_CALLBACK get_raw_key_details;
+static int get_raw_key_details(const OSSL_PARAM params[], void *arg)
+{
+    const OSSL_PARAM *p = NULL;
+    struct raw_key_details_st *raw_key = arg;
+
+    if (raw_key->selection == OSSL_KEYMGMT_SELECT_PRIVATE_KEY) {
+        if ((p = OSSL_PARAM_locate_const(params, OSSL_PKEY_PARAM_PRIV_KEY))
+                != NULL)
+            return OSSL_PARAM_get_octet_string(p, (void **)raw_key->key,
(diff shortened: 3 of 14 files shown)


crypto/evp/p_lib.c
```diff
@@ -317,78 +318,234 @@ int EVP_PKEY_cmp(const EVP_PKEY *a, const EVP_PKEY *b)
...
+                                                          key, len)) {
+                EVPerr(0, EVP_R_KEY_SETUP_FAILED);
+                goto err;
+            }
+
+            params = OSSL_PARAM_BLD_to_param(tmpl);
+            if (params == NULL) {
+                EVPerr(0, ERR_R_MALLOC_FAILURE);
+                goto err;
+            }
```
levitte, line 392: Er, I fail to see why you go through all the trouble with OSSL_PARAM_BUILD here...

``` C
        OSSL_PARAM params[] = { OSSL_PARAM_END, OSSL_PARAM_END };

        params[0] = OSSL_PARAM_construct_octet_string(priv ? OSSL_PKEY_PARAM_PRIV_KEY
                                                           : OSSL_PKEY_PARAM_PUB_KEY,
                                                      key, len);
```

Done.

crypto/evp/p_lib.c
```diff
@@ -317,78 +318,234 @@ int EVP_PKEY_cmp(const EVP_PKEY *a, const EVP_PKEY *b)
...
-               EVP_R_OPERATION_NOT_SUPPORTED_FOR_THIS_KEYTYPE);
+    if (!pkey_set_type(pkey, e, nidtype, strtype, -1, NULL)) {
+        /* EVPerr already called */
         goto err;
     }
 
-    if (!ret->ameth->set_priv_key(ret, priv, len)) {
-        EVPerr(EVP_F_EVP_PKEY_NEW_RAW_PRIVATE_KEY, EVP_R_KEY_SETUP_FAILED);
+    if (pkey->ameth == NULL) {
+        /* Shouldn't happen */
```
levitte, line 422: This could be an assertion, then, i.e. an extra check that `pkey_set_type()` did its job.

util/libcrypto.num
```diff
@@ -5079,3 +5079,5 @@ X509_REQ_set0_signature                 ?	3_0_0	EXIST::FUNCTION:
 X509_REQ_set1_signature_algo            ?	3_0_0	EXIST::FUNCTION:
 OSSL_PARAM_modified                     ?	3_0_0	EXIST::FUNCTION:
 OSSL_PARAM_set_unmodified               ?	3_0_0	EXIST::FUNCTION:
+EVP_PKEY_new_raw_private_key_with_libctx ?	3_0_0	EXIST::FUNCTION:
+EVP_PKEY_new_raw_public_key_with_libctx ?	3_0_0	EXIST::FUNCTION:
```
levitte, line 5083: Please document them

crypto/evp/p_lib.c
mattcaswell: Yes - that's much better. Done.

crypto/evp/p_lib.c
mattcaswell: Done

util/libcrypto.num
mattcaswell: Yes, of course. Done.

mattcaswell:
> I had no time yet to review, but I wanted to start asking if we could also add a test for THESE new APIs so that we can catch further problems also with other key types?

Of course. We already had some tests for the non-libctx versions so I extended them to test both variants.

mattcaswell:
All feedback responded to and updates applied. I have rebased to resolve some conflicts with master and added some commits to deal with the test failures.

This is now out of WIP. Please take another look.

slontis:
There are at least 3 different travis errors.

crypto/ec/ecx_backend.c
```diff
@@ -18,6 +20,31 @@
...
  */
 
+int ecx_public_from_private(ECX_KEY *key)
+{
+    switch (key->type) {
+    case ECX_KEY_TYPE_X25519:
+        X25519_public_from_private(key->pubkey, key->privkey);
+        break;
+    case ECX_KEY_TYPE_ED25519:
+        if (!ED25519_public_from_private(key->libctx, key->pubkey, key->privkey)) {
```
slontis, line 30: Should then functions be passing a property_query string also?

crypto/ec/ecx_key.c
```diff
@@ -10,13 +10,14 @@
 #include <openssl/err.h>
 #include "crypto/ecx.h"
 
-ECX_KEY *ecx_key_new(ECX_KEY_TYPE type, int haspubkey)
+ECX_KEY *ecx_key_new(OPENSSL_CTX *libctx, ECX_KEY_TYPE type, int haspubkey)
```
slontis, line 13: prop_query also??

crypto/evp/p_lib.c
```diff
@@ -318,78 +319,210 @@ int EVP_PKEY_cmp(const EVP_PKEY *a, const EVP_PKEY *b)
...
+        if (strtype != NULL)
+            ameth = EVP_PKEY_asn1_find_str(&tmpe, strtype, -1);
+        else if (nidtype != EVP_PKEY_NONE)
+            ameth = EVP_PKEY_asn1_find(&tmpe, nidtype);
+
+        /* If tmpe is NULL then no engine is claiming to support this type */
+        if (tmpe == NULL)
+            ameth = NULL;
+
+        ENGINE_finish(tmpe);
```
slontis, line 350: Engine API's need to be ifdef

crypto/evp/p_lib.c
```diff
@@ -318,78 +319,210 @@ int EVP_PKEY_cmp(const EVP_PKEY *a, const EVP_PKEY *b)
...
+            EVPerr(0, ERR_R_MALLOC_FAILURE);
+            goto err;
+        }
+        /* May fail if no provider available */
+        ERR_set_mark();
+        if (EVP_PKEY_key_fromdata_init(ctx) == 1) {
+            OSSL_PARAM params[] = { OSSL_PARAM_END, OSSL_PARAM_END };
+
+            ERR_clear_last_mark();
+            params[0] = OSSL_PARAM_construct_octet_string(
```
slontis, line 372: Will this actually work as an octet string???
ecx is the only one that uses octet string for both priv and pub.
ec uses BN for priv.
dsa and dh use BN for both.

doc/man3/EVP_PKEY_new.pod
```diff
@@ -46,16 +58,34 @@ EVP_PKEY_up_ref() increments the reference count of B<key>.
...
-B<type> argument indicates what kind of key this is. The value should be a NID
-for a public key algorithm that supports raw private keys, i.e. one of
-B<EVP_PKEY_HMAC>, B<EVP_PKEY_POLY1305>, B<EVP_PKEY_SIPHASH>, B<EVP_PKEY_X25519>,
-B<EVP_PKEY_ED25519>, B<EVP_PKEY_X448> or B<EVP_PKEY_ED448>. B<key> points to the
-raw private key data for this B<EVP_PKEY> which should be of length B<keylen>.
-The length should be appropriate for the type of the key. The public key data
-will be automatically derived from the given private key data (if appropriate
-for the algorithm type).
+EVP_PKEY_new_raw_private_key_with_libctx() allocates a new B<EVP_PKEY>. Unless
+an engine should be used for the key type a provider for the key is found using
```
slontis, line 62: type,

crypto/evp/p_lib.c
```diff
@@ -318,78 +319,210 @@ int EVP_PKEY_cmp(const EVP_PKEY *a, const EVP_PKEY *b)
...
-                                       size_t len)
+
+static EVP_PKEY *new_raw_key_int(OPENSSL_CTX *libctx,
+                                 const char *strtype,
+                                 const char *propq,
+                                 int nidtype,
+                                 ENGINE *e,
+                                 const unsigned char *key,
+                                 size_t len,
+                                 int priv)
```
slontis, line 330: maybe rename this priv param to priv_enabled. If looks like a private key pointer otherwise.

crypto/evp/p_lib.c
slontis: Ahh I see the docs say only weird types are used :)

doc/man3/EVP_PKEY_new.pod
```diff
@@ -46,16 +58,34 @@ EVP_PKEY_up_ref() increments the reference count of B<key>.
...
+also use "HMAC" which is not a public key algorithm but is treated as such by
+some OpenSSL APIs. You are encouraged to use the EVP_MAC APIs instead for HMAC
+(see L<EVP_MAC(3)>). I<key> points to the raw private key data for this
+B<EVP_PKEY> which should be of length I<keylen>. The length should be
+appropriate for the type of the key. The public key data will be automatically
+derived from the given private key data (if appropriate for the algorithm type).
+
+EVP_PKEY_new_raw_private_key() does the same as
+EVP_PKEY_new_raw_private_key_with_libctx() except that the default library
+context is and default property query is used instead. If B<e> is non-NULL then
```
slontis, line 76: the default library context and default property query string are used.

crypto/ec/ecx_backend.c
mattcaswell: This is definitely a non-trivial task. The problem is that this code is called provider side. If we're going to use the property query here then we'd have to pass it through as an OSSL_PARAM, and presumably have an `ecx_set_params` implementation in order to be able to associate that query with the key in some way. I suspect this problem is bigger than just ECX (we're quite inconsistent about where we pass both a libctx and propq and where we just pass a libctx) - so I suspect we're going to need to address that as a separate issue. Solving this in this case is straying quite far from my original scope for this PR - so I'd rather not try and deal with this here.

crypto/ec/ecx_key.c
mattcaswell: As above

crypto/evp/p_lib.c
mattcaswell: Fixed

crypto/evp/p_lib.c
mattcaswell: Yes...this is for raw keys which are always just a string of bytes.

doc/man3/EVP_PKEY_new.pod
mattcaswell: Fixed

crypto/evp/p_lib.c
mattcaswell: I renamed it too `key_is_priv`

doc/man3/EVP_PKEY_new.pod
mattcaswell: Fixed

mattcaswell:
Fixups pushed addressing the feedback above. Hopefully should also fix the Travis issues.

crypto/ec/ecx_backend.c
paulidale: We pass property queries without the libctx in places.  KDF definitely but I think MAC too.

mattcaswell:
Unfortunately there was a conflict with master in ecx_meth.c - updated and force pushed.

@slontis please can you reconfirm?

slontis:
reconfirmed

mattcaswell:
Pushed. Thanks!

PR #11638, opened 2020-04-24
Partially sync netbsd changes

zoulasc:
- Fix non-vsx capable PPC sha3
- Cosmetics (ctype casts, printf annotations)
- Aliasing warning fixes
- Lint fixes
- Use snprintf instead of sprintf
- Don't override alloca definion when stack protector
- Fix mips64 assembly warning
- Tidy up arm builds (for different families)


crypto/bio/bss_log.c
```diff
@@ -280,7 +280,7 @@ static void xsyslog(BIO *bp, int priority, const char *string)
         break;
     }
 
-    sprintf(pidbuf, "[%lu] ", GetCurrentProcessId());
+    snprintf(pidbuf, sizeof(pidbuf), "[%u] ", GetCurrentProcessId());
```
levitte, line 283: Why the change to %u?

zoulasc: This is a mistake, %lu was correct.

include/openssl/safestack.h
```diff
@@ -32,104 +32,104 @@ extern "C" {
     typedef t3 * (*sk_##t1##_copyfunc)(const t3 *a); \
     static ossl_unused ossl_inline int sk_##t1##_num(const STACK_OF(t1) *sk) \
     { \
-        return OPENSSL_sk_num((const OPENSSL_STACK *)sk); \
+        return OPENSSL_sk_num((const OPENSSL_STACK *)(const void *)sk); \
```
mattcaswell, line 35: Why is this double cast necessary? Same question for all the similar occurrences below.

crypto/bio/bss_log.c
zoulasc: fixed.

apps/s_client.c
```diff
@@ -2765,6 +2765,16 @@ int s_client_main(int argc, char **argv)
     for (;;) {
         FD_ZERO(&readfds);
         FD_ZERO(&writefds);
+        int fdin = fileno_stdin();
+        if (fdin < 0) {
+            BIO_printf(bio_err,"bad fileno for stdin\n");
+            goto shut;
+        }
+        int fdout = fileno_stdout();
```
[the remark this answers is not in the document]
zoulasc: fixed.

crypto/ppccap.c
```diff
@@ -171,6 +171,37 @@ void ecp_nistz256_from_mont(unsigned long res[4], const unsigned long in[4])
...
+
+size_t SHA3_absorb(uint64_t A[5][5], const unsigned char *inp, size_t len,
+    size_t r)
+{
+    return OPENSSL_ppccap_P & PPC_CRYPTO207
+        ? SHA3_absorb_vsx(A, inp, len, r)
+        : SHA3_absorb_default(A, inp, len, r);
+}
+
+void SHA3_squeeze(uint64_t A[5][5], unsigned char *out, size_t len, size_t r)
```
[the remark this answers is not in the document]
zoulasc: Where would you like me to put it? Is there a particular order you want the algorithms?

include/openssl/safestack.h
zoulasc: These are needed for -fstrict-aliasing

zoulasc:
Is there anything else you need me to do?

krytarowski:
It would be nice to marge this.

crypto/ppccap.c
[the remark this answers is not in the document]
mattcaswell: Looks like this code is better placed in the cypto/sha/keccak1600.c file

include/openssl/safestack.h
mattcaswell: Ok

crypto/ppccap.c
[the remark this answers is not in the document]
zoulasc: I am happy to put it there, but this code calls the _vsx variants of the code if the processor has support for them, which are powerpc specific. If you look above in ppccap.c, the same is done for chacha, poly1305, and ECP_NISTZ256_ASM. I can move it, but I was just following suit. Please confirm.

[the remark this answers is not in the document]
zoulasc: Should I change it or leave it?

crypto/ppccap.c
```diff
@@ -171,6 +171,37 @@ void ecp_nistz256_from_mont(unsigned long res[4], const unsigned long in[4])
...
+{
+    OPENSSL_ppccap_P & PPC_CRYPTO207
+        ? SHA3_absorb_vsx(A, out, len, r)
```
vdukhovni, line 199: This is plausibly a cut/paste error?  Should it be `SHA3_squeeze_vsx`?

zoulasc: Indeed. that is a cut/paste error. The mips change is for mips64 where we get from gcc:

Avoid branch delay spot warnings: Warning: macro instruction expanded into
multiple instructions in a branch delay slot

crypto/bn/asm/mips.pl
```diff
@@ -488,8 +488,9 @@
 	$ST	$ta3,-2*$BNSZ($a0)
 
 	.set	noreorder
-	bgtz	$ta0,.L_bn_sqr_words_loop
 	$ST	$ta2,-$BNSZ($a0)
+	bgtz	$ta0,.L_bn_sqr_words_loop
+	 nop
```
vdukhovni, line 493: Does `$ST` really expand into multiple instructions?  Or is gcc issuing a misleading warning?
(Ditto for `$LD` below...)

zoulasc: I am not a mips expert but see http://math-atlas.sourceforge.net/devel/assembly/mips-iv.pdf section A 2.1.2.

vdukhovni: Well, I'm not an assembly programmer on any CPU, so don't really know what the right answer is, but if the GCC warning is legit, then a comment could be useful here.  Or else a comment about why it is OK to ignore the GCC warning.  Perhaps
@levitte will have more constructive feedback here.  The rest of the PR looks fine (module the cut/paste that I expect will be fixed soon).

crypto/ppccap.c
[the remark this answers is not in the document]
mattcaswell: Looking at the other *cap.c files none of them (with one exception) seem to have algorithm specific code in them. The one exception is sparcv9cap.c which does have a BN function defined in it. This looks similarly misplaced to me.

I think the code should be moved out of this file.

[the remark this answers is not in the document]
zoulasc: I am looking at:
https://github.com/openssl/openssl/blob/master/crypto/ppccap.c#L105 (chacha)
and
https://github.com/openssl/openssl/blob/master/crypto/ppccap.c#L132 (poly)

Which are doing the same kind of function selection I am doing. I am I missing something?

crypto/bn/asm/mips.pl
vdukhovni: @dot-asm I hope you might find a few minutes to comment the GCC warnings about the MIPS branch delay slot. Is there in fact a potential issue with:
```
    bgtz	$ta0,.L_bn_sqr_words_loop
    $ST	$ta2,-$BNSZ($a0)
```
that is mitigated by instead writing:
```
    $ST	$ta2,-$BNSZ($a0)
    bgtz	$ta0,.L_bn_sqr_words_loop
    nop
```
on some MIPS systems?

dot-asm: Just in case to eliminate ambiguity, $ST is always **one** mnemonic, `sw` or `sd` depending on processor bit-ness, so it's not about script, but assembler itself. Secondly, even though MIPS assemblers are known to support synthetic mnemonics, such that get expanded into multiple instructions, loads and stores are **not** such mnemonics, they map to actual instructions 1-to-1. Now, I don't quite catch one thing. It sounds like it's about warning. But if I try to trigger it [with GNU assembler up to 2.34], I fail with all ABI variants. So is there any specific assembler flag that was used? In other words how was it compiled more specifically?

On side note one has to recognize that if it was an actual error, then it wouldn't stand 10 years of not failing (even though it wasn't tested as frequently since I retired my SGI workstation).

crypto/armcap.c
```diff
@@ -41,6 +41,7 @@ static void ill_handler(int sig)
  * Following subroutines could have been inlined, but it's not all
  * ARM compilers support inline assembler...
  */
+#if __ARM_MAX_ARCH__>=7
```
dot-asm, line 44: ??? This is in `#else` to `#if __ARM_MAX_ARCH<7`...

crypto/modes/gcm128.c
```diff
@@ -669,33 +669,37 @@ void gcm_gmult_4bit_x86(u64 Xi[2], const u128 Htable[16]);
 void gcm_ghash_4bit_x86(u64 Xi[2], const u128 Htable[16], const u8 *inp,
                         size_t len);
 #  endif
-# elif defined(__arm__) || defined(__arm) || defined(__aarch64__)
+# elif (defined(__arm__) || defined(__arm) || defined(__aarch64__)) && defined(GHASH_ASM)
 #  include "arm_arch.h"
 #  if __ARM_MAX_ARCH__>=7
 #   define GHASH_ASM_ARM
 #   define GCM_FUNCREF_4BIT
-#   define PMULL_CAPABLE        (OPENSSL_armcap_P & ARMV8_PMULL)
```
dot-asm, line 677: Just because it appears counter-intuitive it doesn't mean it's wrong. 32-bit code can take advantage of ARMv8 extensions, and that's what this placement is about.

crypto/modes/gcm128.c
```diff
@@ -669,33 +669,37 @@ void gcm_gmult_4bit_x86(u64 Xi[2], const u128 Htable[16]);
...
                   size_t len);
+#   endif
 #  endif
 # elif defined(__sparc__) || defined(__sparc)
 #  include "sparc_arch.h"
-#  define GHASH_ASM_SPARC
-#  define GCM_FUNCREF_4BIT
+#  if defined(__arch64__)
+#   define GHASH_ASM_SPARC
+#   define GCM_FUNCREF_4BIT
```
dot-asm, line 696: Again, just because it's counter-intuitive, it doesn't mean it's wrong [or unintentional]. Though reason is a bit more subtle than that in ARM case. Just like in ARM case, 32-bit application can take advantage of the module in question, **vendor** compiler unfortunately doesn't pre-define any additional macros if you ask for 32-bit code that can execute on V9 processor. As result, you're stuck with `__sparc` macro as the only option. On additional note, Linux also implements dual-ABI setup, so that throwing in `__arch64__` cripples 32-bit applications. In other words if this actually solves some problem somewhere, it creates problem elsewhere.

crypto/sha/asm/keccak1600-ppc64.pl
```diff
@@ -409,10 +409,10 @@
 	.long	0
 .size	dword_le_load,.-dword_le_load
 
-.globl	SHA3_absorb
-.type	SHA3_absorb,\@function
+.globl	SHA3_absorb_vsx
+.type	SHA3_absorb_vsx,\@function
```
dot-asm, line 413: ??? So that **both** assembly modules export SHA3_absorb_vsx... Either way, it makes lesser sense to deploy vector keccak1600p8-ppc.pl, because benefits are not that clear. It's only nominally faster than scalar module on POWER8 but significantly slower on POWER9. Because POWER9 is better equipped to handle non-vector code. On related note, there is version optimized for little-endian, as well as 32-bit version. [And since MIPS was mentioned, there is even MIPS module...]

zoulasc: Ok, I will disable it instead

crypto/modes/gcm128.c
zoulasc: Thanks for the explanation, I will revert this.

crypto/bn/asm/mips.pl
zoulasc: I was unable to trigger it either for both mips and mips 64. I will revert.

crypto/ppccap.c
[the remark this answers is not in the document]
mattcaswell: I can understand your thinking here. IMO this is still misplaced - and the other algorithm specific code in this file is also misplaced. However, it is perhaps a little much to ask you to resolve this more general issue in this PR. So I'm ok with it being left here for now. It would be nice at some point in the future if someone could properly refactor this and put things in the right place - but that doesn't have to be done now.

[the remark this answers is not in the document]
zoulasc: Thanks @mattcaswell , but I am going to undo the whole thing since @dot-asm wrote below that the performance impact is small and we are better off turning off the assembly for powerpc.

crypto/sha/asm/keccak1600-ppc64.pl
zoulasc: I have disabled it in the NetBSD build, but I am not familiar with the autoconfiguration mechanism to disable it in the standard build.

zoulasc:
Is there anything else outstanding?

apps/s_client.c
```diff
@@ -2765,8 +2765,20 @@ int s_client_main(int argc, char **argv)
     }
 
     for (;;) {
+        int fdin, fdout;
+
         FD_ZERO(&readfds);
         FD_ZERO(&writefds);
+	fdin = fileno_stdin();
```
vdukhovni, line 2772: There's a TAB here, we now always use spaces. Do we care?

vdukhovni: This appears to need a rebase. Perhaps the TAB can be expanded while rebasing?

krytarowski: Yes.

zoulasc: @vdukhovni are you going to take care of it, or should I put another commit on top that fixes it?

vdukhovni: > @vdukhovni are you going to take care of it, or should I put another commit on top that fixes it?

If you don't mind rebasing it, and expanding the TAB while you're at it, that'd be great...

vdukhovni:
Thanks for the whitespace fix, but what about rebasing on the latest upstream commit?

vdukhovni:
This PR has merge-commits.  Figuring out what's old, what's new and how to rebase this to master is a PITA.  I am afraid, I'm going to punt. @zoulasc you're going to have to produce a strictly linear sequence of commits, with reverts squashed, and NO MERGE COMMITS. :-(

krytarowski:
@vdukhovni can you just cherry-pick things that are good?

vdukhovni:
> @vdukhovni can you just cherry-pick things that are good?

No sorry, this PR has bitrotted beyond repair.  If someone from NetBSD wishes to submit a new PR, rebased on a reasonably recent snapshot, I'll review it promptly.

vdukhovni:
I'm closing this PR as abandoned.

PR #11651, opened 2020-04-26
coverity: fixes for latest batch of static analysis problems.

paulidale:
Address 42 of the 45 coverity issues.  The remaining three are:
- 1462557
- 1462559
- 1462568

### DIFF
--- a/apps/lib/app_x509.c
+++ b/apps/lib/app_x509.c
@@ -106,12 +106,18 @@ static int do_x509_ctrl_string(int (*ctrl)(void *object, int cmd,
         cmd = EVP_PKEY_CTRL_SET1_ID; /* ... except we put it in X509 */
 #endif
     } else if (strcmp(stmp, "hexdistid") == 0) {
+        vtmp = "";
+        vtmp_len = 0;
+    } else {
         long hexid_len = 0;
-        void *hexid = OPENSSL_hexstr2buf((const char *)vtmp, &hexid_len);
+        void *hexid;
 
-        OPENSSL_free(stmp);
-        stmp = vtmp = hexid;
-        vtmp_len = (size_t)hexid_len;
+        if (vtmp == NULL) {
+            hexid = OPENSSL_hexstr2buf((const char *)vtmp, &hexid_len);
+            OPENSSL_free(stmp);
+            stmp = vtmp = hexid;
+            vtmp_len = (size_t)hexid_len;
+        }
 #ifdef EVP_PKEY_CTRL_SET1_ID
         cmd = EVP_PKEY_CTRL_SET1_ID; /* ... except we put it in X509 */
 #endif
--- a/apps/lib/app_x509.c
+++ b/apps/lib/app_x509.c
@@ -106,22 +106,17 @@ static int do_x509_ctrl_string(int (*ctrl)(void *object, int cmd,
         cmd = EVP_PKEY_CTRL_SET1_ID; /* ... except we put it in X509 */
 #endif
     } else if (strcmp(stmp, "hexdistid") == 0) {
-        if (vtmp == NULL) {
-            vtmp = "";
-            vtmp_len = 0;
-        } else {
-            if (vtmp == NULL) {
-                long hexid_len = 0;
-                void *hexid;
+        if (vtmp != NULL) {
+            void *hexid;
+            long hexid_len = 0;
 
-                hexid = OPENSSL_hexstr2buf((const char *)vtmp, &hexid_len);
-                OPENSSL_free(stmp);
-                stmp = vtmp = hexid;
-                vtmp_len = (size_t)hexid_len;
-            }
+            hexid = OPENSSL_hexstr2buf((const char *)vtmp, &hexid_len);
+            OPENSSL_free(stmp);
+            stmp = vtmp = hexid;
+            vtmp_len = (size_t)hexid_len;
         }
 #ifdef EVP_PKEY_CTRL_SET1_ID
-            cmd = EVP_PKEY_CTRL_SET1_ID; /* ... except we put it in X509 */
+        cmd = EVP_PKEY_CTRL_SET1_ID; /* ... except we put it in X509 */
 #endif
     }
 
--- a/apps/lib/app_x509.c
+++ b/apps/lib/app_x509.c
@@ -106,20 +106,22 @@ static int do_x509_ctrl_string(int (*ctrl)(void *object, int cmd,
         cmd = EVP_PKEY_CTRL_SET1_ID; /* ... except we put it in X509 */
 #endif
     } else if (strcmp(stmp, "hexdistid") == 0) {
-        vtmp = "";
-        vtmp_len = 0;
-    } else {
-        long hexid_len = 0;
-        void *hexid;
-
         if (vtmp == NULL) {
-            hexid = OPENSSL_hexstr2buf((const char *)vtmp, &hexid_len);
-            OPENSSL_free(stmp);
-            stmp = vtmp = hexid;
-            vtmp_len = (size_t)hexid_len;
+            vtmp = "";
+            vtmp_len = 0;
+        } else {
+            if (vtmp == NULL) {
+                long hexid_len = 0;
+                void *hexid;
+
+                hexid = OPENSSL_hexstr2buf((const char *)vtmp, &hexid_len);
+                OPENSSL_free(stmp);
+                stmp = vtmp = hexid;
+                vtmp_len = (size_t)hexid_len;
+            }
         }
 #ifdef EVP_PKEY_CTRL_SET1_ID
-        cmd = EVP_PKEY_CTRL_SET1_ID; /* ... except we put it in X509 */
+            cmd = EVP_PKEY_CTRL_SET1_ID; /* ... except we put it in X509 */
 #endif
     }
 
--- a/crypto/cms/cms_kari.c
+++ b/crypto/cms/cms_kari.c
@@ -385,6 +385,9 @@ int cms_RecipientInfo_kari_init(CMS_RecipientInfo *ri,  X509 *recip, EVP_PKEY *r
         /* Use originator key */
         CMS_OriginatorIdentifierOrKey *oik = ri->d.kari->originator;
 
+        if (originatorPrivKey == NULL || originator == NULL)
+            return 0;
+
         if (flags & CMS_USE_ORIGINATOR_KEYID) {
              oik->type = CMS_OIK_KEYIDENTIFIER;
              oik->d.subjectKeyIdentifier = ASN1_OCTET_STRING_new();
--- a/crypto/cms/cms_kari.c
+++ b/crypto/cms/cms_kari.c
@@ -412,7 +412,7 @@ static int cms_wrap_init(CMS_KeyAgreeRecipientInfo *kari,
 {
     EVP_CIPHER_CTX *ctx = kari->ctx;
     const EVP_CIPHER *kekcipher;
-    int keylen = EVP_CIPHER_key_length(cipher);
+    int keylen;
     int ret;
 
     /* If a suitable wrap algorithm is already set nothing to do */
@@ -422,8 +422,10 @@ static int cms_wrap_init(CMS_KeyAgreeRecipientInfo *kari,
             return 0;
         return 1;
     }
-    else if (cipher != NULL
-         && (EVP_CIPHER_flags(cipher) & EVP_CIPH_FLAG_GET_WRAP_CIPHER)) {
+    if (cipher == NULL)
+        return 0;
+    keylen = EVP_CIPHER_key_length(cipher);
+    if ((EVP_CIPHER_flags(cipher) & EVP_CIPH_FLAG_GET_WRAP_CIPHER) != 0) {
         ret = EVP_CIPHER_meth_get_ctrl(cipher)(NULL, EVP_CTRL_GET_WRAP_CIPHER,
                                                0, &kekcipher);
         if (ret <= 0)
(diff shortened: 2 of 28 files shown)


paulidale:
Travis is a timeout waiting for the S39X build to start rather than a failure.

crypto/x509/v3_tlsf.c
```diff
@@ -135,5 +135,6 @@ static TLS_FEATURE *v2i_TLS_FEATURE(const X509V3_EXT_METHOD *method,
 
  err:
     sk_ASN1_INTEGER_pop_free(tlsf, ASN1_INTEGER_free);
+    ASN1_INTEGER_free(ai);
```
mattcaswell, line 138: I think this could lead to a double free if we go around the "for" loop multiple times, and then get an error before the line where we reassign `ai` to a new integer. Probably we need to additionally set `ai = NULL;` after the successful push.

paulidale: Good point.

paulidale: Done

paulidale:
Ping?

crypto/evp/pkey_mac.c
```diff
@@ -496,10 +496,17 @@ static int pkey_mac_ctrl_str(EVP_PKEY_CTX *ctx,
                               const char *type, const char *value)
 {
     MAC_PKEY_CTX *hctx = EVP_PKEY_CTX_get_data(ctx);
-    const EVP_MAC *mac = EVP_MAC_CTX_mac(hctx->ctx);
+    const EVP_MAC *mac;
     OSSL_PARAM params[2];
     int ok = 0;
 
+    if (hctx->ctx == NULL) {
```
slontis, line 503: maybe also do hctx == NULL ||

crypto/evp/keymgmt_lib.c
```diff
@@ -384,11 +384,6 @@ int evp_keymgmt_util_copy(EVP_PKEY *to, EVP_PKEY *from, int selection)
         return 0;
     }
 
-    if (to->keymgmt == NULL
```
slontis, line 387: Why is this completely deleted??

crypto/evp/p_lib.c
```diff
@@ -1212,10 +1212,12 @@ static int pkey_set_type(EVP_PKEY *pkey, ENGINE *e, int type, const char *str,
          * to the |save_type| field, because |type| is supposed to be set
          * to EVP_PKEY_NONE in that case.
          */
-        if (keymgmt != NULL)
-            pkey->save_type = ameth->pkey_id;
-        else if (pkey->ameth != NULL)
-            pkey->type = ameth->pkey_id;
+        if (ameth != NULL) {
```
slontis, line 1215: Should it return 0 in this case?

crypto/rsa/rsa_gen.c
```diff
@@ -462,9 +462,6 @@ static int rsa_keygen_pairwise_test(RSA *rsa, OSSL_CALLBACK *cb, void *cbarg)
                                        padding);
     if (ciphertxt_len <= 0)
         goto err;
-    if (ciphertxt_len == plaintxt_len
```
slontis, line 465: I think the actual bug needs to be fixed for this - not just delete the code.
i.e- The ciphertext must be tested to be not equal to the plaintext.

crypto/evp/keymgmt_lib.c
paulidale: We've already de-referenced `to->keymgmt` in all paths leading to here, it cannot be NULL.

slontis: @levitte probably should review this?

crypto/evp/p_lib.c
paulidale: I'm not sure.  Is it mandatory for a PKEY to have an ameth?  FIPS mode doesn't, why not others?

crypto/evp/pkey_mac.c
paulidale: Added with its own error.

crypto/evp/p_lib.c
slontis: This whole section is a !FIPS_MODE

apps/lib/app_x509.c
```diff
@@ -107,8 +107,13 @@ static int do_x509_ctrl_string(int (*ctrl)(void *object, int cmd,
...
     } else if (strcmp(stmp, "hexdistid") == 0) {
         long hexid_len = 0;
-        void *hexid = OPENSSL_hexstr2buf((const char *)vtmp, &hexid_len);
+        void *hexid;
 
+        if (vtmp == NULL) {
+            OPENSSL_free(stmp);
+            return -1;
+        }
+        hexid = OPENSSL_hexstr2buf((const char *)vtmp, &hexid_len);
```
levitte, line 116: Alternately, if vtmp is NULL, hexid could be the empty string, let ctrl deal with it...

crypto/rsa/rsa_gen.c
paulidale: I'm pretty sure this is a copy paste in the wrong place problem.  The same block of code appears three lines further down after `decoded` has been initialised.

It's coverity 1462561 Uninitialized scalar variable:
```C
unsigned char *ciphertxt = NULL;
443    const unsigned char plaintxt[16] = {0};
   1. var_decl: Declaring variable decoded without initializer.
444    unsigned char decoded[256];
445    unsigned int decoded_len;
446    unsigned int plaintxt_len = (unsigned int)sizeof(plaintxt_len);
447    int padding = RSA_PKCS1_PADDING;
448    OSSL_SELF_TEST *st = NULL;
449
450    st = OSSL_SELF_TEST_new(cb, cbarg);
   2. Condition st == NULL, taking false branch.
451    if (st == NULL)
452        goto err;
453    OSSL_SELF_TEST_onbegin(st, OSSL_SELF_TEST_TYPE_PCT,
454                           OSSL_SELF_TEST_DESC_PCT_RSA_PKCS1);
455
456    ciphertxt_len = RSA_size(rsa);
457    ciphertxt = OPENSSL_zalloc(ciphertxt_len);
   3. Condition ciphertxt == NULL, taking false branch.
458    if (ciphertxt == NULL)
459        goto err;
460
461    ciphertxt_len = RSA_public_encrypt(plaintxt_len, plaintxt, ciphertxt, rsa,
462                                       padding);
   4. Condition ciphertxt_len <= 0, taking false branch.
463    if (ciphertxt_len <= 0)
464        goto err;
   5. Condition ciphertxt_len == plaintxt_len, taking true branch.
   CID 1462561 (#1 of 1): Uninitialized scalar variable (UNINIT)6. uninit_use_in_call: Using uninitialized element of array decoded when calling memcmp.
465    if (ciphertxt_len == plaintxt_len
466        && memcmp(decoded, plaintxt, plaintxt_len) == 0)
467        goto err;
468
469    OSSL_SELF_TEST_oncorrupt_byte(st, ciphertxt);
470
471    decoded_len = RSA_private_decrypt(ciphertxt_len, ciphertxt, decoded, rsa,
472                                      padding);
473    if (decoded_len != plaintxt_len
474        || memcmp(decoded, plaintxt,  decoded_len) != 0)
475        goto err;
476
477    ret = 1;
478 err:
479    OSSL_SELF_TEST_onend(st, ret);
480    OSSL_SELF_TEST_free(st);
481    OPENSSL_free(ciphertxt);
482
483    return ret;
```

crypto/evp/keymgmt_lib.c
levitte: I'll have a look.

crypto/rsa/rsa_gen.c
slontis: The intention for the last bit of code is to make sure the decrypted ciphertext matches the original input plaintext.
The intention of the middle check is that the encrypted cipher text does not match the plaintext i.e - the encryption actually does something. (its one of the check it says it should be doing).

apps/lib/app_x509.c
paulidale: Much of a muchness.

I've changed the code over to pass a null string instead.

crypto/rsa/rsa_gen.c
paulidale: I guess it could be checking that the ciphertext is different to the plaintext???

crypto/evp/p_lib.c
levitte: This is correct

crypto/rsa/rsa_gen.c
paulidale: Put in a check that the ciphertxt is different to the plaintxt instead.

crypto/evp/p_lib.c
paulidale: Thanks :)

crypto/x509/v3_ist.c
```diff
@@ -102,9 +102,6 @@ static int i2r_issuer_sign_tool(X509V3_EXT_METHOD *method,
         return 0;
     }
     if (ist->signTool != NULL) {
-        if (new_line == 1) {
-            BIO_write(out, "\n", 1);
-        }
```
levitte, line 107: This is silly...

paulidale: You want the dead code put back in?
Why?

paulidale: The code is back and I've marked coverity as a false positive for this one.

levitte: I figured it showed the pattern to follow, should anyone want to add another printed line before this one. But never mind, the code fragments are small enough to make it obvious.

paulidale: I've reverted this already, might as well leave it in.

crypto/evp/keymgmt_lib.c
levitte: Actually, it's quite unclear what happens, and what *should* happen when `to->keymgmt == NULL`...  This "quick fix" does not resolve that question.

levitte: I find it interesting, though, that coverity didn't catch the completely unused `alloc_keydata`.

paulidale: > Actually, it's quite unclear what happens, and what _should_ happen when `to->keymgmt == NULL`... This "quick fix" does not resolve that question.

I know what happens: you get a core dump earlier in the function :)
As for what _should_ happen, I've no idea.

levitte: That being said, coverity hasn't actually determined that `to->keymgmt` can't be NULL at this point, it only *assumes* this because we've dereferenced it all over the place.  For some reason, it didn't report that we might dereference NULL in one of those code branches.

levitte: See #11668

paulidale: I'll ditch this change then.

paulidale: It's gone.   Now to figure out which other amendment broke the CIs :(

apps/lib/app_x509.c
paulidale: The alternative causes memory leaks in the tests :(

levitte: Er, so deal with it?  Do you want me to take over this case?

paulidale: Is supplying an argument `hexdistid` without the colon or hex string really that meaningful?

paulidale: Something is wrong here.  The `sm2_pmeth` function that handles this repeated the `OPENSSL_hexstr2buf()` call but using the converted string.

levitte: Hmmm, perhaps not.  I had to look around how we do this elsewhere to be sure...

Okie then

levitte: > Something is wrong here. The `sm2_pmeth` function that handles this repeated the `OPENSSL_hexstr2buf()` call but using the converted string.

Huh...  let me have a look

levitte: > Something is wrong here.  The `sm2_pmeth` function that handles this repeated the `OPENSSL_hexstr2buf()` call but using the converted string.

I don't understand how this happens...  it should not be called at all?

paulidale: The solution seems to be to pass `NULL` not `""`.

paulidale: I deleted my mistaken comment...

crypto/cms/cms_kari.c
```diff
@@ -385,9 +385,6 @@ int cms_RecipientInfo_kari_init(CMS_RecipientInfo *ri,  X509 *recip, EVP_PKEY *r
          /* Use originator key */
          CMS_OriginatorIdentifierOrKey *oik = ri->d.kari->originator;
 
-         if (originatorPrivKey == NULL && originator == NULL)
```
t8m, line 388: Shouldn't it be actually kept and just `&&` changed to `||`?

ssl/tls13_enc.c
```diff
@@ -608,6 +608,7 @@ int tls13_change_cipher_state(SSL *s, int which)
                 SSLfatal(s, SSL_AD_INTERNAL_ERROR,
                          SSL_F_TLS13_CHANGE_CIPHER_STATE,
                          SSL_R_ALGORITHM_FETCH_FAILED);
+                    EVP_MD_CTX_free(mdctx);
```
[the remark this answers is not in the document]
paulidale: Done.

crypto/cms/cms_kari.c
paulidale: I'm not sure: CMS isn't something I know much about.  Except to know that I don't want to know more if I can help if :)  Still, it seems reasonable since these are dereferenced without checking.

The indentation was off by one here too.

paulidale:
Merged to master.  Thanks for the help.

PR #11655, opened 2020-04-27
include/openssl/x509v3.h: restore previous stack definition arrangement

levitte:
It turned out that configuration options may affect the definition and
use of diverse stacks and how they relate to the underlying types.
For example, the configuration option 'no-rfc3779' results in a build
error around STACK_OF(IPAddressFamily) and related stacks.

Previously, STACK_OF definitions were located near the definition of
the underlying type, which are also affected by configuration options,
which made this easier to maintain.  We relocate the new stack
definitions back to those locations for that reason.


levitte:
> Approved, but I think the same change needs to be made in other headers too.

Hmmmm, I can't see that those lead to build errors...  however, consistency has value

levitte:
> Approved, but I think the same change needs to be made in other headers too.

Added.

t8m:
The Travis failure is relevant

levitte:
> The Travis failure is relevant

Hopefully fixed

richsalz:
So only the #ifndef OPENSSL_NO_RFC3779 is a bugfix, the rest are just esthetics it seems.  Is that right?

levitte:
> So only the #ifndef OPENSSL_NO_RFC3779 is a bugfix, the rest are just esthetics it seems. Is that right?

... and consistency, and making future maintenance easier by having things grouped together (-ish, I must admit...  some stack definitions are kinda all over the place)

richsalz:
I am all in favor of consistency.  I just wanted to make sure I understand what bug *I* created.  Thanks.

levitte:
CIs are happy.  It would be good if someone did a final review, there are reports of build failures that are related to this.

levitte:
@richsalz, this resolved more than I recalled when I answered you above.  See these threads:

https://mta.openssl.org/pipermail/openssl-users/2020-April/012353.html
https://mta.openssl.org/pipermail/openssl-users/2020-April/012350.html

levitte:
Merged

01659135a1d8e042c6777886a74b682b32b28161 include/openssl/x509v3.h: restore previous stack definition arrangement
90a37ce389182cfdbf347c3a85317524a2d0e39a include/openssl/ts.h: clean away a misplaced EVP_MD stack definition

richsalz:
Tnx for the additional details, @levitte.
I guess I need to test big header changes with a series of no-xxx builds?

t8m:
Unfortunately it is not just about no-xxx I think. It is also coming from the fact that we must be very careful when moving things around headers as some application code might depend on something implicitly declared by inclusion of a header from a different header they already include.

richsalz:
In the "define not declare" PR's, I was very careful to *keep the declarations.*

markt-asf:
For the record, the Apache httpd (mod_ssl) and Apache Tomcat (Native) CI builds both broke and are fixed now this has been reverted. I didn't get as far as figuring out which includes I needed to add to fix them.

t8m:
> In the "define not declare" PR's, I was very careful to _keep the declarations._

It was not a problem that the declarations were dropped but moved around.

richsalz:
> It was not a problem that the declarations were dropped but moved around.

yes, noted. thanks.

PR #11656, opened 2020-04-27
crypto/x509/v3_alt.c: make 'othername' a bit bigger

levitte:
We want to fill 'othername' with the contents of 'oline' (256 bytes)
plus some additional text.  We need to ensure that 'othername' is
large enough to contain this.


paulidale (approved):
I was tempted to classify this coverity issue as a false positive.  The copies to the buffer are well guarded.

levitte:
I had a compiler warning telling me that the contents of `oline` may be truncated...

mattcaswell:
I'm marking this as "urgent". I'm getting build errors because of this problem.

mattcaswell (approved):

levitte:
Merged

d0bcad73253b7ea0a6f8a4ebd2aa567257cd2101 crypto/x509/v3_alt.c: make 'othername' a bit bigger

PR #11666, opened 2020-04-28
Add vpaes_ctr32_encrypt_blocks for x86_64

bernd-edlinger:
This adds an analog to the (currently unused) "2x" optimization from
vpaes-armv8.pl and a ctr128_f implementation.

##### Checklist

### DIFF
--- a/crypto/evp/e_aes.c
+++ b/crypto/evp/e_aes.c
@@ -115,6 +115,14 @@ void vpaes_cbc_encrypt(const unsigned char *in,
                        unsigned char *out,
                        size_t length,
                        const AES_KEY *key, unsigned char *ivec, int enc);
+# if defined(__x86_64) || defined(__x86_64__) \
+     || defined(_M_AMD64) || defined(_M_X64)
+void vpaes_ctr32_encrypt_blocks(const unsigned char *in, unsigned char *out,
+                                size_t len, const AES_KEY *key,
+                                const unsigned char ivec[16]);
+# else
+#  define vpaes_ctr32_encrypt_blocks NULL
+# endif
 #endif
 #ifdef BSAES_ASM
 void bsaes_cbc_encrypt(const unsigned char *in, unsigned char *out,
@@ -186,9 +194,6 @@ extern unsigned int OPENSSL_ia32cap_P[];
 
 # ifdef VPAES_ASM
 #  define VPAES_CAPABLE   (OPENSSL_ia32cap_P[1]&(1<<(41-32)))
-# endif
-# ifdef BSAES_ASM
-#  define BSAES_CAPABLE   (OPENSSL_ia32cap_P[1]&(1<<(41-32)))
 # endif
 /*
  * AES-NI section
@@ -2681,8 +2686,11 @@ static int aes_init_key(EVP_CIPHER_CTX *ctx, const unsigned char *key,
         ret = vpaes_set_encrypt_key(key, EVP_CIPHER_CTX_key_length(ctx) * 8,
                                     &dat->ks.ks);
         dat->block = (block128_f) vpaes_encrypt;
-        dat->stream.cbc = mode == EVP_CIPH_CBC_MODE ?
-            (cbc128_f) vpaes_cbc_encrypt : NULL;
+        dat->stream.cbc = NULL;
+        if (mode == EVP_CIPH_CBC_MODE)
+            dat->stream.cbc = (cbc128_f) vpaes_cbc_encrypt;
+        else if (mode == EVP_CIPH_CTR_MODE)
+            dat->stream.ctr = (ctr128_f) vpaes_ctr32_encrypt_blocks;
     } else
 #endif
     {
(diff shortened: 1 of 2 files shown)


crypto/evp/e_aes.c
```diff
@@ -115,6 +115,14 @@ void vpaes_cbc_encrypt(const unsigned char *in,
                        unsigned char *out,
                        size_t length,
                        const AES_KEY *key, unsigned char *ivec, int enc);
+# if defined(__x86_64) || defined(__x86_64__) \
+     || defined(_M_AMD64) || defined(_M_X64)
```
davidben, line 119: Not for this PR, but grepping for `__x86_64` in the source tree, I see a lot of variations of this ifdef. Perhaps there's something to be said for having some common architecture ifdefs in some common file?

crypto/evp/e_aes.c
```diff
@@ -2686,8 +2694,12 @@ static int aes_init_key(EVP_CIPHER_CTX *ctx, const unsigned char *key,
         ret = vpaes_set_encrypt_key(key, EVP_CIPHER_CTX_key_length(ctx) * 8,
                                     &dat->ks.ks);
         dat->block = (block128_f) vpaes_encrypt;
-        dat->stream.cbc = mode == EVP_CIPH_CBC_MODE ?
-            (cbc128_f) vpaes_cbc_encrypt : NULL;
+        dat->stream.cbc = NULL;
+        if (mode == EVP_CIPH_CBC_MODE)
+            dat->stream.cbc = (cbc128_f) vpaes_cbc_encrypt;
+        else
+        if (mode == EVP_CIPH_CTR_MODE)
```
davidben, line 2701: Huh, is this the usually OpenSSL style? I thought you all usually had `else` and `if` on the same line.

crypto/evp/e_aes.c
bernd-edlinger: I don't even know which compiler made this mess necessary.
I could find out for GCC, but there are lots of other compilers
where I cannot look at the source code.  I do not understand
if that was just s stupid idea from the beginning. or was there
a reason for this multiple defines?  There is no comment
anywhere, giving the rationale, therefore I personally think
that was driven by irrational fear about possible bugs in
unknown / ancient no longer supported compiler versions.

crypto/evp/e_aes.c
bernd-edlinger: Okay, I will look if the same style is nearby, I think that I would not break the local style
since style-cleanups tend to cause merge conflicts, but yes, sometimes I mix up the
style guides from openssl gcc linux OPC/UA as I often switch between different styles ;-)

bernd-edlinger:
@davidben FYI: This is a back-port of your contribution to #9677
The main body of that PR was already merged to 1.1.1 in 87bea6550ae0dda7c40937cff2e86cc2b0b09491

So there is no BSAES any more for X86_64, therefore this PR should make
a difference.  There is still BSAES for ARMv8 IIRC, therfore I did not remove
BSAES completely.
But I can try to add a "int3" to the assembler module, and see if I can
trigger that.

I do not understand why that was approved for 1.1.1 but the same PR for
master is stuck, and I will probably close it soon.

Frankly in the moment I feel like creating my own personal totallyBoringSSL branch.
What were your reasons for ceating BoringSSL?

bernd-edlinger:
I just confirmed,
```
diff --git a/crypto/aes/asm/vpaes-x86_64.pl b/crypto/aes/asm/vpaes-x86_64.pl
index c15618d..6e66642 100644
--- a/crypto/aes/asm/vpaes-x86_64.pl
+++ b/crypto/aes/asm/vpaes-x86_64.pl
@@ -92,6 +92,7 @@ $code.=<<___;
 .align 16
 _vpaes_encrypt_core:
 .cfi_startproc
+int3
        mov     %rdx,   %r9
        mov     \$16,   %r11
        mov     240(%rdx),%eax
```
OPENSSL_ia32cap=2199023255552 make test
gives many test errors. without the int3 the above test is good.

davidben:
> So there is no BSAES any more for X86_64, therefore this PR should make a difference. There is still BSAES for ARMv8 IIRC, therfore I did not remove BSAES completely.
> But I can try to add a "int3" to the assembler module, and see if I can trigger that.

Ah, sorry, I'd missed that! Ignore what I said about this being a no-op then. Is the x86 `BSAES_CAPABLE` definition still needed then?

(Regarding BoringSSL, I didn't start the project, but the README has some comment about how it came about. I got involved because I work on Chrome and was distractable. :-) )

bernd-edlinger:
No reason to feel sorry about that.  I am happy you looked at this PR.
So it is probably my fault to not start the conversation with
"Thank you very much for your review" :-)

bernd-edlinger:
> Is the x86 BSAES_CAPABLE definition still needed then?

good point I will check that.

bernd-edlinger:
> regarding "Chrome"

isn't that a toxic element?
By the way does it alse have the "omni.ja" file, I found recently in my firefox/thunderbird
installation, where hidden extensions are installed that are next to impossible to uninstall?

PR #11668, opened 2020-04-28
EVP: Fix evp_keymgmt_util_copy() for to->keymgmt == NULL

levitte:
evp_keymgmt_util_copy() didn't treat the case to->keymgmt correctly.
The proper change is to use from->keymgmt when to->keymgmt is NULL.

Fixes coverity #1462553

### DIFF
--- a/crypto/evp/keymgmt_lib.c
+++ b/crypto/evp/keymgmt_lib.c
@@ -346,10 +346,19 @@ int evp_keymgmt_util_copy(EVP_PKEY *to, EVP_PKEY *from, int selection)
     if (from == NULL || from->keydata == NULL)
         return 0;
 
+    /*
+     * If |to| is unassigned, ensure it gets the same KEYMGMT as |from|,
+     * Note that the final setting of KEYMGMT is done further down, with
+     * EVP_PKEY_set_type_by_keymgmt(); we don't want to do that prematurely.
+     */
+    if (to_keymgmt == NULL)
+        to_keymgmt = from->keymgmt;
+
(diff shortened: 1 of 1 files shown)


levitte:
Note that this change really doesn't matter *for the moment*, 'cause we're only calling this function when both `to->keymgmt != NULL` and `from->keymgmt != NULL`.  However, when legacy code is gone, I expect that this will be called unconditionally by `EVP_PKEY_copy_parameters()`.

(I really should get the "no-legacy-code" config PR started, that's going to demonstrate what we can expect things to look like in the farther future)

crypto/evp/keymgmt_lib.c
```diff
@@ -369,16 +378,15 @@ int evp_keymgmt_util_copy(EVP_PKEY *to, EVP_PKEY *from, int selection)
         import_data.selection = selection;
 
         if (!evp_keymgmt_export(from->keymgmt, from->keydata, selection,
-                                &try_import, &import_data)) {
-            evp_keymgmt_freedata(to_keymgmt, alloc_keydata);
+                                &try_import, &import_data))
             return 0;
```
slontis, line 382: Memory leak?? evp_keymgmt_freedata(to_keymgmt, alloc_keydata);

levitte: Ah, thanks

levitte: Restored

levitte:
Merged

036ee3706352970a15300b5cd4bf0b2cb370e12a EVP: Fix evp_keymgmt_util_copy() for to->keymgmt == NULL

Issue #6640, opened 2018-07-03
BIGNUM code is not constant-time due to bn_correct_top


davidben:
This is a write-up of some work we did recently in BoringSSL. It addresses a long-standing timing leak in `BIGNUM`. Unfortunately, the issue is in a core `BIGNUM` invariant, so it won't be a small fix. But I thought it'd be good to have a write-up available for when folks are ready to tackle it.

---

Most of OpenSSL's asymmetric crypto code is not constant-time because, among other things, `bn_correct_top` chops off leading zeros. This is to satisfy an internal invariant where all `BIGNUM`s are minimal representation.

This invariant makes sense for calculators, but it is [incompatible with cryptography](https://mta.openssl.org/pipermail/openssl-dev/2017-December/009887.html). While p and q in RSA have public bit widths (half of the bit width of n), d has a secret bit width. It is only publicly bounded by n. Field elements and scalars in EC are only publicly bounded by the field size and group order, respectively. Zero as a field element in P-256 and P-384 should have different representations. Leaking information about the minimal bit width of RSA private key operation result, for example, can give a private key oracle, per Manger's attack. (Although, for RSA keys that are a whole number of words, particularly on 64-bit, the probability of leaking anything is low enough that this is probably mostly theoretical.)

Unfortunately, many `BIGNUM` functions internally rely on this invariant. E.g. `BN_abs_is_word` and `BN_is_zero` here:
https://github.com/openssl/openssl/blob/dfee8626a8f6c1e23ab270a6fc20b4d1ba145392/crypto/bn/bn_lib.c#L840-L848

Or this logic from `BN_cmp`, which assumes a wider number always has larger magnitude.
https://github.com/openssl/openssl/blob/dfee8626a8f6c1e23ab270a6fc20b4d1ba145392/crypto/bn/bn_lib.c#L574-L577

## Approach

We finally found a successful approach for BoringSSL. We've been running with it for most of the year now and it seems fine.

First off, the invariant needs to go. That's unavoidable, which means that logic like the above needs to be rewritten. We relaxed the invariant slightly so that `bn->top`, which we renamed to `bn->width` (see below for why) refers not to the *minimal* word width, but the *public* width. This is a public upper bound on the value and also the size of `bn->d`. If `bn->width` is 4, the value is less than `2^(4*BN_BITS2)`. But it *may* fit in fewer words.

This means there are many different representations for a given value. E.g. any number of zero words means zero.

The calling convention for `BIGNUM` functions is then:
[...]

Issue #5158, opened 2018-01-24
Replace setters and empty states with better constructors


davidben:
Ideally, this sort of thing should be done in two stages: first add APIs in 1.1.x for the world to switch to them, then remove the existing APIs in 1.2.0.

OpenSSL 1.1.0 opaquified structs, but did so in a way that does not help much in reducing complexity for future changes. Every object has an unnecessary empty state. For example, to construct an RSA key, one calls `RSA_new` and then fills in the components piecemeal. To hash something, one calls `EVP_MD_CTX_new` and then separately calls `EVP_DigestInit_ex` to initialize it.

This is bad for both users and OpenSSL. For consumers, it is users to initialize every object in two stages, both of which must be checked for failures. The library is more complex and thus more error-prone to use.

For OpenSSL, this adds a lot of complexity. Every object has a public "empty" state, which every function must now check for. Additionally, every property, even if it really only needs to be touched on construction, is mutable. This adds a lot of extra complexity.

For instance, `EVP_MD_CTX_new` and `EVP_DigestInit_ex` should have been combined into:

```
EVP_MD_CTX *EVP_MD_CTX_new_digest(const EVP_MD *type, ENGINE *engine);
```

This is how most libraries with heap-allocated hash contexts work. The split is a remnant of stack-allocated `EVP_MD_CTX`s, where there are separate allocation and construction stages. There is no reason to split them when heap-allocating because `NULL` is already a zero state.

In this API, `EVP_DigestUpdate` does not need to worry about the callers passing in an empty `EVP_MD_CTX` because it is impossible to vend such a thing. Additionally, `EVP_DigestInit_ex` does not need to worry about resizing the hash context when the hash changes, because that's impossible. The caller would just make a new `EVP_MD_CTX`.

(It even allows a small optimization. One could allocate the `EVP_MD_CTX` and hash context in a single allocation.)

On the RSA side, this is especially pronounced. Beyond the problems above of (multiple!) partially-constructed states, there is no easy point for an `RSA` object to "finish" initializing. `RSA` objects have a pre-computed `BN_MONT_CTX`. Right now, these are filled in lazily via `BN_MONT_CTX_set_locked`. That makes the implementation more complex and carries a confusing API contract for the user: `RSA` objects are mutable, but you should not call those functions after you've used it, because we'll set up caches and get the answer wrong. (One could imagine making the setters drop those caches, but that's even more corner cases for a use case that does not actually exist; it is simpler to just make a new `RSA` object.)

[...]

PR #11669, opened 2020-04-28
Undeprecate DH, DSA and RSA _bits() functions.

paulidale:
These were deemed information and useful and that they should not be
deprecated.


CHANGES.md
```diff
@@ -101,36 +101,34 @@ OpenSSL 3.0
...
-   RSA_meth_get0_app_data, RSA_meth_set0_app_data, RSA_meth_get_pub_enc,
-   RSA_meth_set_pub_enc, RSA_meth_get_pub_dec, RSA_meth_set_pub_dec,
-   RSA_meth_get_priv_enc, RSA_meth_set_priv_enc, RSA_meth_get_priv_dec,
-   RSA_meth_set_priv_dec, RSA_meth_get_mod_exp, RSA_meth_set_mod_exp,
-   RSA_meth_get_bn_mod_exp, RSA_meth_set_bn_mod_exp, RSA_meth_get_init,
-   RSA_meth_set_init, RSA_meth_get_finish, RSA_meth_set_finish,
-   RSA_meth_get_sign, RSA_meth_set_sign, RSA_meth_get_verify,
-   RSA_meth_set_verify, RSA_meth_get_keygen, RSA_meth_set_keygen,
-   RSA_meth_get_multi_prime_keygen and RSA_meth_set_multi_prime_keygen.
+    RSA_new_method, RSA_size, RSA_security_bits, RSA_get0_pss_params,
```
levitte, line 104: Why the extra indentation?

paulidale: No idea where that came from.  Fixed now.

paulidale:
`make update` done now.

davidben:
This isn't a review comment, just a question for my own edification: what is OpenSSL 3.0's plan w.r.t. the DH, DSA, and RSA objects? I thought all those modules were to be deprecated in favor of EVP, but this PR suggests otherwise. Is the idea then that you'd still be able to get an RSA object out of an EVP_PKEY and poke at it?

kroeckx:
It was at least my impression that we would deprecate them,
and would sort of expect all function in files like dh.h, dsa.h,
rsa.h, aes.h, md*.h, sha.h and so on to be deprecated. Looking
at those files, a lot of the functions seem to be deprecated,
but it's not clear why some aren't to me. Other files seem to be
untouched.

paulidale:
We still need some of the functions for key creation purposes.
The deprecations have removed the ability to do _computations_ with these things.

paulidale:
Merged to master, thanks for the comments and suggestions.

mattcaswell:
> This isn't a review comment, just a question for my own edification: what is OpenSSL 3.0's plan w.r.t. the DH, DSA, and RSA objects? I thought all those modules were to be deprecated in favor of EVP, but this PR suggests otherwise. Is the idea then that you'd still be able to get an RSA object out of an EVP_PKEY and poke at it?

You'll still be able to create DH, DSA and RSA objects and get some basic values (p, q, g, etc) out of them - and then assign them to an EVP_PKEY. Any functions that actually use those objects directly to do some kind of crypto operations will be deprecated.

davidben:
Ah, interesting. Key creation isn't covered by `EVP_PKEY_fromdata`? I guess setting up a bunch of `OSSL_PARAM`s is mildly tedious?

The reason I ask is that we're looking at adding new constructor functions for `RSA`, in light of all the problems we've experienced with the OpenSSL 1.1.1 APIs (#5158), both for callers and for the implementation. I figured you all wouldn't be looking to do that given this new EVP and provider-centric future, but if `RSA` objects still have a future, then it might be worth getting those improvements into OpenSSL as well.

mattcaswell:
`EVP_PKEY_fromdata` is certainly the intended way ahead. We won't be adding any *new* objects like RSA/DH/DSA - and I also don't expect to be encouraging users to use them. But actually deprecating them completely was considered a step too far at this stage. I'm not sure we'd want to add new constructors...but would rather encourage users to go to `EVP_PKEY_fromdata` instead.

davidben:
Sounds good. `EVP_PKEY_fromdata` does avoid the flaw from OpenSSL 1.1.x, namely that the RSA components are set piecemeal. That means there's no point where you can check the parameters or compute derived values like Montgomery information (and thus `RSA` objects have this oddball lock inside them). Indeed mutating an `RSA` object often silently gives the wrong answer because of those derived values.

It's tricky to take advantage of this until the old APIs are gone, but glancing through the current implementation, it doesn't look like `EVP_PKEY_fromdata` runs `RSA_check_key`. I would recommend doing that and not creating an invalid key. (`RSA_check_key` being a separate operation was a weird consequence of old exposed struct, carried over via 1.1.x's API mistakes. It really should be one combined operation.)

levitte:
> it doesn't look like `EVP_PKEY_fromdata` runs `RSA_check_key`

I assume that you mean to say that `RSA_check_key()` should be run in the provider implementation.  However, that could be tricky, because...

>  `EVP_PKEY_fromdata` does avoid the flaw from OpenSSL 1.1.x, namely that the RSA components are set piecemeal.

Does it?

``` C
    EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_from_name(NULL, "RSA", NULL);
    EVP_PKEY *pkey = NULL;

    EVP_PKEY_key_fromdata_init(ctx);
    EVP_PKEY_fromdata(ctx, &pkey, some_pieces);
    EVP_PKEY_fromdata(ctx, &pkey, some_other_pieces);
```

This is permitted by the code.  I can't currently tell you how the provider code reacts to that (too early in the morning for that sort of digging :wink:), but the libcrypto code definitely allows this.

The responsibility to check the validity of the keys falls back on the user, and there is a set of functions to do exactly that:

```
: ; grep -E 'EVP_PKEY_[^_]*_check\(' include/openssl/evp.h 
int EVP_PKEY_public_check(EVP_PKEY_CTX *ctx);
int EVP_PKEY_param_check(EVP_PKEY_CTX *ctx);
int EVP_PKEY_private_check(EVP_PKEY_CTX *ctx);
int EVP_PKEY_pairwise_check(EVP_PKEY_CTX *ctx);
```

mattcaswell:
> This is permitted by the code

Should we allow that? Or should we take the opportunity to disallow it?

levitte:
I'm not sure I see much benefit for fromdata, really. However, we might want to have a look at validation when a key is to be used, and when it's loaded from somewhere. This could even be done entirely in the provider, with the result cached in the key object.

mattcaswell:
> I'm not sure I see much benefit for fromdata

Well the benefit is that you can then never have a key in a half completed state. It's either entirely empty or complete. This has the potential to simplify code elsewhere where we have to have to allow for some fields being filled in and other fields being empty. It also means we know when a key is "finished" and we can run validation checks etc on it.

levitte:
Who's "we"? Let me put it another way, when is it important that a key is "finished"? I have assumed that validation would be the test to see if a key is "finished", was that an incorrect assumption?

mattcaswell:
"We" = OpenSSL developers.
I don't think we would want to run validation checks on a key every time we've had some partial update just to find out if it was finished. But we might want to automatically run validation checks if we believe it is finished and reject the key if validation fails.

levitte:
BTW, I misremembered things...  Every time `EVP_PKEY_fromdata` is called, a new internal key is created.  So my example in https://github.com/openssl/openssl/pull/11669#issuecomment-622247013 is false.

davidben:
> I don't think we would want to run validation checks on a key every time we've had some partial update just to find out if it was finished. But we might want to automatically run validation checks if we believe it is finished and reject the key if validation fails.

Right, if your API allows partial updates, input validation is impossible. But there's no reason to allow partial updates in the first place. Keys are rather atomic things. You aren't going to get a useful key by swapping out the public modulus and leaving the rest alone. If you need a new key, just make a new one.

It's not just validity, but also setting up the key itself. RSA implementations use Montgomery reduction, which requires setting up some state. In a normal API, this is straightforward: it's part of the process of importing a key.

However, in OpenSSL's API, there is no time to set up this state. It needs to be deferred (see `BN_MONT_CTX_set_locked`) to the first operation, with thread synchronization and everything. If it was done in construction, none of this complexity would be needed.

If you all fix #6640, you'll encounter another version of this: `d`, `dmp1`, `dmq1`, and `iqmp` should be sized relative to `n`, `p`, `q`, and `p`, respectively. They may not be sized that way on input from the caller. (Indeed even the standard ASN.1 serialization mis-sizes them.) They should be normalized on key input so that the operations themselves aren't affected. But if the caller configures `dmp1` before `p`, you can't do anything here.

We ultimately needed to extend the `BN_MONT_CTX_set_locked` lazy init to a `freeze_private_key` function to compute the actual key. But since that's done inside an RSA operation, it can no longer mutate the usual copy of `dmp1`. (It is valid for one thread to call `RSA_get0_dmp1` and read the `BIGNUM` while another thread calls `RSA_sign`.) So we make a `d_fixed` copy and, as a result, RSA keys take around twice as much memory as they should.

On top of this, partial updates don't even work. I believe the following sequence of events in OpenSSL produces the wrong answer. The cached Montgomery values aren't reset.
```
RSA *rsa = RSA_new();
RSA_set0_key(rsa, ...);  // Configure a key.
RSA_set0_factors(rsa, ...);
RSA_set0_crt_params(rsa, ...);
RSA_sign(rsa, ...);  // Make a signature.
RSA_set0_key(rsa, ...);  // I changed my mind! Different key.
RSA_set0_factors(rsa, ...);
RSA_set0_crt_params(rsa, ...);
RSA_sign(rsa, ...);  // Make another signature.
```

All this mess can be avoid by simply setting all the parameters at once and making `RSA` objects immutable after creation. The above use case is still fine: the caller just calls `RSA_free` on the old key and makes a new one.

> BTW, I misremembered things... Every time EVP_PKEY_fromdata is called, a new internal key is created. So my example in #11669 (comment) is false.

Cool. My suggestion then still stands: perform input checks at key import time, because that's the natural time to do it, and you're in a position to set the expectations for a new API. Deferring 
to the first operation adds unnecessary complexity.

Fixing the other consequences of the historical APIs is something you can do internally without affecting callers, and is easier once `RSA_set0_*` goes away. Whether you want better RSA constructors is I guess a question of whether you believe the ecosystem will quickly move off of `RSA` objects, or if adding a local fix is worthwhile to get rid of `RSA_set0_*` faster. (I expect we'll keep the low-level APIs around for a while, and we have a tighter feedback loop with downstream code, so it's worthwhile for us. Whether it's worthwhile for OpenSSL, I dunno.)

But you need the input checks part of the API contract from the start, otherwise you'll run into problems. For instance, Montgomery reduction only works for odd moduli. Even moduli are blatantly invalid, so that's fine. But it becomes much messier to set up Montgomery reduction on import if callers don't expect `EVP_PKEY_fromdata` to perform input checks. (To compare, while `RSA_set0_*` technically can return failure, it only does so on programmer error. At this point it would be a breaking change to fix that because no one checks the return value.)

Issue #11648, opened 2020-04-26
Support sha512-256 with RSA PSS signatures


alex:
Currently attempting to call `EVP_PKEY_CTX_set_signature_md` for RSA signatures with an `EVP_MD*` that's for SHA-512/256 results in an error. It'd be great if this was fully supported! Not sure what additional information is required here.

mattcaswell:
Did you get anything on the OpenSSL error stack? My guess is it needs an update here:

https://github.com/openssl/openssl/blob/70411a50d4b207e6d0352c12588e93c0c206ca39/crypto/rsa/rsa_pmeth.c#L378-L399

And probably here too (although provider side RSA-PSS support isn't quite there yet):

https://github.com/openssl/openssl/blob/70411a50d4b207e6d0352c12588e93c0c206ca39/providers/implementations/signature/rsa.c#L126-L142

levitte:
Correct, @mattcaswell.  Baically, have a look [in RFC 8017 A.2](https://tools.ietf.org/html/rfc8017#appendix-A.2) and copy what you find missing from there.

slontis:
I will add this - (some of the entries like ripe are probably questionable ?)

slontis:
So we have an RSA.asn1 table that doesnt have
        md5_sha1, 
        md4, 
        mdc2,
        ripemd160

Do we want these also or should they be removed from rsa.c?

levitte:
> So we have an RSA.asn1 table that doesnt have
> ...
> Do we want these also or should they be removed from rsa.c?

It may come as a surprise if anyone still uses those digests with RSA signatures...
Deprecation comes to mind, but I can't currently imagine what that should look like.
In the mean time, I'd say keep them for the moment, but possibly attach a TODO cleanup comment.

richsalz:
I would like to see MD4 MDC2 and RIPEMD160 removed.

alex:
It seems @slontis fixed the subject of the original issue here several years ago. Good enough for me!

PR #11681, opened 2020-04-30
Add RSA SHA512 truncated digest support.

slontis:
Partial Fix for #11648.

Some additional work still needs to be done to support RSA-PSS mode.

RSA legacy digests will be addressed in another PR.

##### Checklist


slontis:
ping

levitte:
Travis failure not relevant here

slontis:
Thanks. Merged to master.

bbolli:
It seems to me that the `evppkey.txt` change adds the same test for SHA512-224 twice, and none for SHA512-256.

slontis:
> It seems to me that the `evppkey.txt` change adds the same test for SHA512-224 twice, and none for SHA512-256.

Not quite - one is a sign test the other is a verify of the same data. An extended test that interops with another toolkit would be more useful.

bbolli:
You're right. Sorry for not looking hard enough.

PR #11683, opened 2020-04-30
Remove gen_get_params & gen_gettable_params from keygen operation

slontis:
EVP_PKEY_CTX_gettable_params() was missing code for the keygen operation.
After adding it it was noticed that it is probably not required for this type, so instead
the gen_get_params and gen_gettable_params have been remnoved from the provider interface.
gen_get_params was only implemented for ec to get the curve name. This seems redundant
since normally you would set parameters into the keygen_init() and then generate a key.
Normally you would expect to extract data from the key - not the object that we just set up
to do the keygen.

Added a simple settable and gettable test into a test that does keygen.

##### Checklist

### DIFF
--- a/crypto/evp/pmeth_lib.c
+++ b/crypto/evp/pmeth_lib.c
@@ -613,12 +613,6 @@ int EVP_PKEY_CTX_get_params(EVP_PKEY_CTX *ctx, OSSL_PARAM *params)
             && ctx->op.ciph.cipher->get_ctx_params != NULL)
         return ctx->op.ciph.cipher->get_ctx_params(ctx->op.ciph.ciphprovctx,
                                                    params);
-    if (EVP_PKEY_CTX_IS_GEN_OP(ctx)
-        && ctx->op.keymgmt.genctx != NULL
-        && ctx->keymgmt != NULL
-        && ctx->keymgmt->gen_get_params != NULL)
-        return evp_keymgmt_gen_get_params(ctx->keymgmt, ctx->op.keymgmt.genctx,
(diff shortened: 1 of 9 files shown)


levitte:
This is incomplete. You need to deal with the SM2 hack in crypto/evp/pmeth_gn.c as well.

slontis:
> This is incomplete. You need to deal with the SM2 hack in crypto/evp/pmeth_gn.c as well.

Ahh ok. Updated the hack :)

crypto/evp/pmeth_lib.c
```diff
@@ -656,8 +648,7 @@ const OSSL_PARAM *EVP_PKEY_CTX_settable_params(EVP_PKEY_CTX *ctx)
             && ctx->op.ciph.cipher->settable_ctx_params != NULL)
         return ctx->op.ciph.cipher->settable_ctx_params();
     if (EVP_PKEY_CTX_IS_GEN_OP(ctx)
-            && ctx->keymgmt != NULL
-            && ctx->keymgmt->gen_settable_params != NULL)
+            && ctx->keymgmt != NULL)
```
levitte, line 651: Er, why?

slontis: I knew you would ask that - the util function that it calls does this check. It is different to some of the others that just call the function directly if the value is not NULL.

levitte: Ah!

slontis:
Thanks. Merged to master.

PR #11689, opened 2020-04-30
rand_unix.c: Prefer KERN_ARND over /dev/urandom on NetBSD and fix usage

### LINKED ISSUE #11687: libcrypto fails to use KERN_ARND on NetBSD due to sys/sysctl.h not being included, and always falls back to reading from /dev/urandom
This is fixed in NetBSD's vendored copy of OpenSSL. I'd like to help fix it here but I don't have access to a scanner to complete a CLA - can I sign it digitally?

`rand_unix.c` not only keeps usage of `sysctl(KERN_ARND)` behind `if defined(__FreeBSD__) || defined(__NetBSD__)`, it also checks if the `KERN_ARND` symbol is defined on these platforms. This is fragile. It makes what should be a hard error into a silent one. The syscall is available on all supported NetBSD and FreeBSD versions going back a decade.

[OpenSSL isn't including sys/sysctl.h on NetBSD](https://github.com/openssl/openssl/blob/master/crypto/rand/rand_unix.c#L30) when the code obviously assumes it's included. So, KERN_ARND never gets defined, and the syscall never gets used due to the silent failure.  The fallback to /dev/urandom introduces potential problems because opening and reading from it can fail, while `sysctl(KERN_ARND)` never fails if used correctly. `sys/sysctl.h` should be included on NetBSD in `rand_unix.c`.

[This hack for old FreeBSD versions](https://github.com/openssl/openssl/blob/master/crypto/rand/rand_unix.c#L224) is enabled unconditionally. It should probably be enabled conditionally just in case it means we can't get entropy on NetBSD.

`sysctl(KERN_ARND)` supports non-blocking reads of up to 256 bytes per call. Requests for higher numbers will fail. [If buflen exceeds 256 bytes, len should be set to 256 bytes here](https://github.com/openssl/openssl/blob/master/crypto/rand/rand_unix.c#L251). I'm not sure if OpenSSL will request more than 256 bytes in practice, but it's something to be aware of.

niacat:
Fixes #11687

### DIFF
--- a/crypto/rand/rand_unix.c
+++ b/crypto/rand/rand_unix.c
@@ -250,7 +250,7 @@ static ssize_t sysctl_random(char *buf, size_t buflen)
     mib[1] = KERN_ARND;
 
     do {
-        len = buflen;
+        len = buflen > 256 ? 256 : buflen;
         if (sysctl(mib, 2, buf, &len, NULL, 0) == -1)
             return done > 0 ? done : -1;
         done += len;
--- a/crypto/rand/rand_unix.c
+++ b/crypto/rand/rand_unix.c
@@ -227,10 +227,12 @@ static ssize_t sysctl_random(char *buf, size_t buflen)
      * when the sysctl returns long and we want to request something not a
      * multiple of longs, which should never be the case.
      */
+#if   defined(__FreeBSD__)
     if (!ossl_assert(buflen % sizeof(long) == 0)) {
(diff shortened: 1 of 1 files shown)


mattcaswell:
Close/reopen to kick CLA bot

mattcaswell:
Try again...

mattcaswell:

PR #11691, opened 2020-04-30
Centralise Environment Variables for the tests

mattcaswell:
The test_includes test was failing if OPENSSL_CONF_INCLUDE happened to
be set in the user's environment. To ensure that no tests accidentally
use this or other enviroment variables from the user's environment we
automatically set them centrally for all tests.

##### Checklist

### DIFF
--- a/test/recipes/80-test_ct.t
+++ b/test/recipes/80-test_ct.t
@@ -11,7 +11,7 @@ use OpenSSL::Test qw/:DEFAULT srctop_file srctop_dir/;
 use OpenSSL::Test::Simple;
 
 setup("test_ct");
-$ENV{CTLOG_FILE} = srctop_file("test", "ct", "log_list.cnf");
+
 $ENV{CT_DIR} = srctop_dir("test", "ct");
 $ENV{CERTS_DIR} = srctop_dir("test", "certs");
 simple_test("test_ct", "ct_test", "ct", "ec");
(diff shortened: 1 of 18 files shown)


test/recipes/80-test_ct.t
```diff
@@ -11,7 +11,7 @@ use OpenSSL::Test qw/:DEFAULT srctop_file srctop_dir/;
 use OpenSSL::Test::Simple;
 
 setup("test_ct");
-$ENV{CTLOG_FILE} = srctop_file("test", "ct", "log_list.cnf");
+
 $ENV{CT_DIR} = srctop_dir("test", "ct");
```
richsalz, line 15: Why aren't these two env vars centralized?

richsalz:
Should probably have something in test/README about this.

test/recipes/80-test_ct.t
mattcaswell: These 2 are only ever set by and read by this test - so there is no possibility of accidentally getting a wrong value from the user's environment. Also since they are only relevant to this test it seems pointless to centralise them.

mattcaswell:
I updated test/README with some text about this. I have *no* idea how the changes in this PR could cause the Travis failure and I can't reproduce it locally. Lets see if that failure re-occurs after this latest push.

Please take a look.

mattcaswell:
> I have no idea how the changes in this PR could cause the Travis failure and I can't reproduce it locally. Lets see if that failure re-occurs after this latest push.

Unfortunately it did re-occur. It seems there was a latent memory leak in the CONF ".include" handling. We weren't hitting it before, but now the tests are run with OPENSSL_CONF_INCLUDE set the fuzz test found the leak. I've added a commit to address that issue. Lets see what travis makes of it now.

mattcaswell:
Travis is now green.

Ping?

mattcaswell:
Pushed. Thanks.

PR #11696, opened 2020-05-01
Add OIDS for md4 and ripemd160 to der_rsa

slontis:
##### Checklist

### DIFF
--- a/providers/common/der/RSA.asn1
+++ b/providers/common/der/RSA.asn1
@@ -86,3 +86,14 @@ id-rsassa-pkcs1-v1_5-with-sha3-224 OBJECT IDENTIFIER ::= { sigAlgs 13 }
 id-rsassa-pkcs1-v1_5-with-sha3-256 OBJECT IDENTIFIER ::= { sigAlgs 14 }
 id-rsassa-pkcs1-v1_5-with-sha3-384 OBJECT IDENTIFIER ::= { sigAlgs 15 }
 id-rsassa-pkcs1-v1_5-with-sha3-512 OBJECT IDENTIFIER ::= { sigAlgs 16 }
+
+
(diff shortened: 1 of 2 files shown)


providers/common/der/RSA.asn1
```diff
@@ -86,3 +86,14 @@ id-rsassa-pkcs1-v1_5-with-sha3-224 OBJECT IDENTIFIER ::= { sigAlgs 13 }
 id-rsassa-pkcs1-v1_5-with-sha3-256 OBJECT IDENTIFIER ::= { sigAlgs 14 }
 id-rsassa-pkcs1-v1_5-with-sha3-384 OBJECT IDENTIFIER ::= { sigAlgs 15 }
 id-rsassa-pkcs1-v1_5-with-sha3-512 OBJECT IDENTIFIER ::= { sigAlgs 16 }
+
+
+-- -------------------------------------------------------------------
+-- These OID's exist in the codebase but may need to be deprecated at some point.
+-- mdc2 and md5_sha1 have been omitted as they do not look like valid entries.
```
levitte, line 93: Would you mind adding a reference to where you found this?

slontis: Richard,

I used one of the oid repos for RIPE.
The closest thing I can find is https://homes.esat.kuleuven.be/~bosselae/ripemd160.html (which references iso standards which I cant see).
I can see some RFC https://www.ietf.org/rfc/rfc4051.txt that has references to dead links - so that is not really great either.
So the short answer is no :)
It is referred to in PKCS11 and CMS so it is probably being used by someone somewhere.

levitte: Okie, I can deal with that.

slontis:
Thanks. Merged to master.

PR #11697, opened 2020-05-01
Add solaris assembler fixes for legacy provider

slontis:
The legacy provider contains assembler references. Most code is automagically pulled in from libcrypto - but the platform specific assembler functions will not be visible in the symbol table. Copying BNASM and DESASM into liblegacy seems to be a better solution than exposing platform specific function in libcrypto.num.

Added a missing call in the des_cbc code for sparc.

##### Checklist


slontis:
Thanks. Merged to master.

PR #11698, opened 2020-05-01
Fix aix compile error in cmp_ctx_test.c

slontis:
Errors were of the form 1506-226 (S) The ":" operator is not allowed between "int" and "char*".
I think it is valid syntax the way it was written, But just rewrote so it compiled.
The aix compiler must be looking at the type of blah() when doing test ? (blah(), NULL) : X.

##### Checklist


levitte:
I find this one to be clearer than #11705

slontis:
Thanks. Merged to master.

PR #11700, opened 2020-05-01
 Don't offer or accept ciphersuites that we can't support

mattcaswell:
We were not correctly detecting whether TLSv1.3 ciphersuites could
actually be supported by the available provider implementations. For
example a FIPS client would still offer CHACHA20-POLY1305 based
ciphersuites even though it couldn't actually use them. Similarly on
the server would try to use CHACHA20-POLY1305 and then fail the
handshake.

### DIFF
--- a/ssl/ssl_ciph.c
+++ b/ssl/ssl_ciph.c
@@ -1596,8 +1596,16 @@ STACK_OF(SSL_CIPHER) *ssl_create_cipher_list(const SSL_METHOD *ssl_method,
 
     /* Add TLSv1.3 ciphers first - we always prefer those if possible */
     for (i = 0; i < sk_SSL_CIPHER_num(tls13_ciphersuites); i++) {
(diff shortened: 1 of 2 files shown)


mattcaswell:
Fixup pushed to address the travis failure.

Ping?

t8m:
The Travis failure is now unrelated.

IMO, on 1.1.1 there is no FIPS so there should be no issue - either the cipher is built-in and then it works or isn't and then the related ciphersuite is not present at all.

mattcaswell:
> IMO, on 1.1.1 there is no FIPS so there should be no issue - either the cipher is built-in and then it works or isn't and then the related ciphersuite is not present at all.

Correct. The difference is that in master the availability of algorithms can be determined as runtime based on what providers are loaded. In 1.1.1 we don't have the situation with the TLSv1.3 ciphers where they could be available at compile time, but not at runtime.

mattcaswell:
Pushed. Thanks.

PR #11703, opened 2020-05-01
WPACKET: don't write DER length when we don't want to

levitte:
With endfirst writing, it could be that we want to abandon any zero
length sub-packet.  That's what WPACKET_FLAGS_ABANDON_ON_ZERO_LENGTH
was supposed to make happen, but the DER length writing code didn't
look at that flag.  Now it does.

### DIFF
--- a/crypto/packet.c
+++ b/crypto/packet.c
@@ -267,7 +267,8 @@ static int wpacket_intern_close(WPACKET *pkt, WPACKET_SUB *sub, int doclose)
             return 0;
     } else if (pkt->endfirst && sub->parent != NULL
                && (packlen != 0
-                   || (sub->flags & WPACKET_FLAGS_NON_ZERO_LENGTH) == 0)) {
+                   || (sub->flags
+                       & WPACKET_FLAGS_ABANDON_ON_ZERO_LENGTH) == 0)) {
         size_t tmplen = packlen;
         size_t numlenbytes = 1;
 
--- a/crypto/packet.c
+++ b/crypto/packet.c
@@ -265,7 +265,9 @@ static int wpacket_intern_close(WPACKET *pkt, WPACKET_SUB *sub, int doclose)
                 && !put_value(&buf[sub->packet_len], packlen,
                               sub->lenbytes))
             return 0;
-    } else if (pkt->endfirst && sub->parent != NULL) {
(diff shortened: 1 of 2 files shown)


mattcaswell (dismissed):
LGTM. Approved assuming CIs agree and you fix the commit message to actually make sense ;-)

levitte:
> ... and you fix the commit message to actually make sense ;-)

Better?

mattcaswell:
> Better now?

Yes - thanks!

levitte:
> Yes - thanks!

I actually had a good laugh when I re-read it.  Complete gibberish!

kaduk (changes requested):
The commit message talks about `WPACKET_FLGAS_ABANDON_ON_ZERO_LENGTH` but the code checks `WPACKET_FLAGS_NON_ZERO_LENGTH`.

mattcaswell:
> The commit message talks about WPACKET_FLGAS_ABANDON_ON_ZERO_LENGTH but the code checks WPACKET_FLAGS_NON_ZERO_LENGTH.

Oops. Good catch!

crypto/packet.c
```diff
@@ -265,7 +265,9 @@ static int wpacket_intern_close(WPACKET *pkt, WPACKET_SUB *sub, int doclose)
                 && !put_value(&buf[sub->packet_len], packlen,
                               sub->lenbytes))
             return 0;
-    } else if (pkt->endfirst && sub->parent != NULL) {
+    } else if (pkt->endfirst && sub->parent != NULL
+               && (packlen != 0
+                   || (sub->flags & WPACKET_FLAGS_NON_ZERO_LENGTH) == 0)) {
```
levitte, line 270: Hold on, this isn't right...

levitte: Ah, I see others discovered the same thing.

levitte: Fixed

levitte:
Travis failure not relevant here.  Fix in #11708

kaduk (approved):
That looks better now, thanks.

mattcaswell (approved):

levitte:
Merged

12cbb8e0497bc4990cfb02f1b9ebe23df9b53a2c WPACKET: don't write DER length when we don't want to

PR #11704, opened 2020-05-01
util/perl/OpenSSL/OID.pm: remove the included unit test

levitte:
The unit test uses features that appeared in perl 5.12, and is
therefore a source of trouble when building.


levitte:
Merged

6d81bb2676f6a210f15f17324ab6852f52cc7f55 util/perl/OpenSSL/OID.pm: remove the included unit test

romen:
@levitte is there a way to retrieve this test by making it conditional to the perl version?

Or is there a good reason why this test is not important anymore?

levitte:
We never ran that test suite, but the original that this comes from (see comment at the beginning of the module) does.

levitte:
But to answer your question, I don't think it's possible to have part of a module considered separately depending on the perl version. It's an interesting idea, but...

romen:
Ah great, thanks for answering!

richsalz:
It might be worthwhile putting it into the regular test suite.

levitte:
I dunno.  After all, it's run during the build, so testing it as part of the test suite becomes an afterthought...

PR #11705, opened 2020-05-01
Fix test macro compile errors on AIX

nrgiii:
Make SSL tests compile with the IBM xlc compiler.
Resolve (several) errors like this by adding casts:
"test/cmp_ctx_test.c", line 723.1: 1506-226 (S) The ":" operator is not allowed between "int" and "char*".


mattcaswell:
Same as #11698.

slontis:
I am going to close this one as #11698 has been merged.. Thank you for doing the fix.

PR #11708, opened 2020-05-02
Fix reason code clash

levitte:
EVP_R_NULL_MAC_PKEY_CTX vs EVP_R_SET_DEFAULT_PROPERTY_FAILURE

PR #11709, opened 2020-05-02
SSL_OP_DISABLE_TLSEXT_CA_NAMES option implementation

NMorozxov:
##### Checklist

### DIFF
--- a/apps/s_server.c
+++ b/apps/s_server.c
@@ -761,7 +761,7 @@ typedef enum OPTION_choice {
     OPT_SRTP_PROFILES, OPT_KEYMATEXPORT, OPT_KEYMATEXPORTLEN,
     OPT_KEYLOG_FILE, OPT_MAX_EARLY, OPT_RECV_MAX_EARLY, OPT_EARLY_DATA,
     OPT_S_NUM_TICKETS, OPT_ANTI_REPLAY, OPT_NO_ANTI_REPLAY, OPT_SCTP_LABEL_BUG,
-    OPT_HTTP_SERVER_BINMODE,
+    OPT_HTTP_SERVER_BINMODE, OPT_NOCANAMES,
     OPT_R_ENUM,
     OPT_S_ENUM,
     OPT_V_ENUM,
@@ -952,6 +952,8 @@ const OPTIONS s_server_options[] = {
     {"anti_replay", OPT_ANTI_REPLAY, '-', "Switch on anti-replay protection (default)"},
     {"no_anti_replay", OPT_NO_ANTI_REPLAY, '-', "Switch off anti-replay protection"},
     {"http_server_binmode", OPT_HTTP_SERVER_BINMODE, '-', "opening files in binary mode when acting as http server (-WWW and -HTTP)"},
+    {"no_ca_names", OPT_NOCANAMES, '-',
+     "Disable TLS Extension CA Names"},
     {"stateless", OPT_STATELESS, '-', "Require TLSv1.3 cookies"},
 #ifndef OPENSSL_NO_SSL3
     {"ssl3", OPT_SSL3, '-', "Just talk SSLv3"},
@@ -1089,6 +1091,7 @@ int s_server_main(int argc, char *argv[])
     const char *keylog_file = NULL;
     int max_early_data = -1, recv_max_early_data = -1;
     char *psksessf = NULL;
+    int no_ca_names = 0;
 #ifndef OPENSSL_NO_SCTP
     int sctp_label_bug = 0;
 #endif
@@ -1655,6 +1658,9 @@ int s_server_main(int argc, char *argv[])
         case OPT_HTTP_SERVER_BINMODE:
             http_server_binmode = 1;
             break;
+        case OPT_NOCANAMES:
+            no_ca_names = 1;
+            break;
         case OPT_SENDFILE:
 #ifndef OPENSSL_NO_KTLS
             use_sendfile = 1;
@@ -1899,6 +1905,10 @@ int s_server_main(int argc, char *argv[])
 
     if (async) {
         SSL_CTX_set_mode(ctx, SSL_MODE_ASYNC);
+    }
+
+    if (no_ca_names) {
+        SSL_CTX_set_options(ctx, SSL_OP_DISABLE_TLSEXT_CA_NAMES);
     }
 
     if (max_send_fragment > 0
--- a/doc/man3/SSL_CTX_set_options.pod
+++ b/doc/man3/SSL_CTX_set_options.pod
@@ -67,6 +67,12 @@ The following B<bug workaround> options are available:
 Don't prefer ECDHE-ECDSA ciphers when the client appears to be Safari on OS X.
 OS X 10.8..10.8.3 has broken support for ECDHE-ECDSA ciphers.
 
+=item SSL_OP_DISABLE_TLSEXT_CA_NAMES
+
+Disable TLS Extension CA Names. You may want to disable it for security reasons
+or for compatibility with some Windows TLS implementations crashing when this
+extension is larger than 1024 bytes.
+
 =item SSL_OP_DONT_INSERT_EMPTY_FRAGMENTS
 
 Disables a countermeasure against a SSL 3.0/TLS 1.0 protocol
@@ -378,7 +384,7 @@ The B<SSL_OP_NO_EXTENDED_MASTER_SECRET> option was added in OpenSSL 3.0.
 
 =head1 COPYRIGHT
 
-Copyright 2001-2019 The OpenSSL Project Authors. All Rights Reserved.
+Copyright 2001-2020 The OpenSSL Project Authors. All Rights Reserved.
 
 Licensed under the Apache License 2.0 (the "License").  You may not use
 this file except in compliance with the License.  You can obtain a copy
--- a/doc/man1/openssl-s_server.pod.in
+++ b/doc/man1/openssl-s_server.pod.in
@@ -46,6 +46,7 @@ B<openssl> B<s_server>
 [B<-www>]
 [B<-WWW>]
 [B<-http_server_binmode>]
+[B<-no_ca_names>]
 [B<-servername>]
 [B<-servername_fatal>]
 [B<-tlsextdebug>]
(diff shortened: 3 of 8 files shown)


doc/man3/SSL_CTX_set_options.pod
```diff
@@ -67,6 +67,12 @@ The following B<bug workaround> options are available:
 Don't prefer ECDHE-ECDSA ciphers when the client appears to be Safari on OS X.
 OS X 10.8..10.8.3 has broken support for ECDHE-ECDSA ciphers.
 
+=item SSL_OP_DISABLE_TLSEXT_CA_NAMES
+
+Disable TLS Extention CA Names. You can want to disable it for security reasons.
+Another usage: some Windows TLS Intplementation crushes if this extention is
+bigger than 1024 bytes.
```
beldmit, line 74: Better wording is smth like

Disable TLS Extension CA Names. You may want to disable it for security reasons or for compatibility with some Windows TLS implementations crashing when this extension is larger than 1024 bytes.

NMorozxov: fixed

NMorozxov:
@beldmit fix commited

beldmit:
Travis failures seem irrelevant. 
Could you please take a look whether similar options are tested and, if yes, add a test for your option?

NMorozxov:
Ok i will try
> Travis failures seem irrelevant.
> Could you please take a look whether similar options are tested and, if yes, add a test for your option?

richsalz:
Does this need to be exposed at the s_client, etc., level?  Via SSL_conf_cmd or similar?

NMorozxov:
SSL_CTX_set_options(ctx, SSL_OP_DISABLE_TLSEXT_CA_NAMES);
when creating SSL Context
> Does this need to be exposed at the s_client, etc., level? Via SSL_conf_cmd or similar?

richsalz:
Can you use s_client to set this new option?  I do not think so.  I think it should be possible to use s_client to set this option. The way to do that is to add it to the tables (and documentation) for ```SSL_conf_cmd()/SSL_CTX_conf_cmd()```

NMorozxov:
Ok i understand. Fixed
> Can you use s_client to set this new option? I do not think so. I think it should be possible to use s_client to set this option. The way to do that is to add it to the tables (and documentation) for `SSL_conf_cmd()/SSL_CTX_conf_cmd()`

NMorozxov:
@beldmit @richsalz fix commited

NMorozxov:
**EVP reason code 208 collision at EVP_R_SET_DEFAULT_PROPERTY_FAILURE**
is relevant but doen't belong to my PR
i just pull a clean repo and easy got this error after make update
Some ones other PR broke this

NMorozxov:
on the fixing PR
https://github.com/openssl/openssl/pull/11708

kaduk:
I would be interested in hearing more about the scenarios where this is seen to be needed.  While I do recall the stories of misconfigured systems that send out (e.g.) all of /etc/ssl/certs using lots of bandwidth, it's not immediately clear to me that an SSL option is better than just correcting the configuration to only send a few (or zero) certificates at all.
In general, I am reluctant to add configuration options to specifically disable a given TLS extension, since the TLS ecosystem is supposed to be robust to unknown extensions and we have other ways to influence the contents of this one in particular.  (It was, unfortunately, necessary to add options for extended-master-secret and encrypt-then-mac because of deployed implementations that choke badly when they receive them.  I'm not aware of such implementation bugs surrounding the certificate-authorities extension (hence my desire to know more).

NMorozxov:
Imagine this scenario: I want to make a web server with TLS authentication using GOST12-GOST89-GOST8912 cjipersuite, that allowed ONLY for users from Russian government certificated CA. Check the users after TLS connection established is prohibited by  moderator.
The solution:
1) I can build nginx http://nginx.org/ with OpenSSL
2) I can download the CA certs from http://e-trust.gosuslugi.ru/CA/DownloadTSL?schemaVersion=0
3) I add all certificates in nginx ssl_trusted_certificate and set ssl_verify_client on
4) I works, but some Windows can't connect. Disabling this extention will allow them to connect.

NMorozxov:
ping @beldmit @kaduk

beldmit:
I think it also worth adding an option to the `s_server` application.

kaduk:
Thanks for the clear description of the problematic situation.  It seems like there would be a few other avenues that could (at a technical level) be tweaked to allow things to work, so I personally do not want to endorse this mechanism.  Having said that, I will not object if someone else approves it and it goes into the tree.

beldmit:
@kaduk, I am not sure I understand which other avenues you mean.

NMorozxov:
@beldmit option to s_server commited

apps/s_server.c
```diff
@@ -952,6 +952,8 @@ const OPTIONS s_server_options[] = {
     {"anti_replay", OPT_ANTI_REPLAY, '-', "Switch on anti-replay protection (default)"},
     {"no_anti_replay", OPT_NO_ANTI_REPLAY, '-', "Switch off anti-replay protection"},
     {"http_server_binmode", OPT_HTTP_SERVER_BINMODE, '-', "opening files in binary mode when acting as http server (-WWW and -HTTP)"},
+    {"no-CANames", OPT_NOCANAMES, '-',
```
beldmit, line 955: Not minus but underscore?

NMorozxov:
@beldmit fix commited
but there is 
{"no-CAfile", OPT_NOCAFILE, '-',

beldmit:
Travis failures are relevant.

NMorozxov:
@beldmit fix commited

kaduk:
> @kaduk, I am not sure I understand which other avenues you mean.

Mostly it relates to "Check the users after TLS connection established is prohibited by moderator." being a policy (non-technical) requirement.  I don't know why it's there, or whether it would also prohibit checking the users during (not after) TLS connection establishment.  Based on my current knowledge it appears as just an arbitrary unneeded restriction.
There's also an option to tell people with old, unsupported, non-protocol-compliant windows clients that they will need to find something that implements the spec properly, though I do not actually expect anyone to pick this route.
I also don't know the details of the certification process, who is eligible to get certified, how many stale entries there are in the presented corpus, or whether it's practical to enforce a consolidation to a small subset of the certified CAs.

apps/s_server.c
t8m: Also I'd put all in lower case. I do not think the CAfile no-CAfile example is a good thing to follow.

apps/s_server.c
```diff
@@ -952,6 +952,8 @@ const OPTIONS s_server_options[] = {
     {"anti_replay", OPT_ANTI_REPLAY, '-', "Switch on anti-replay protection (default)"},
     {"no_anti_replay", OPT_NO_ANTI_REPLAY, '-', "Switch off anti-replay protection"},
     {"http_server_binmode", OPT_HTTP_SERVER_BINMODE, '-', "opening files in binary mode when acting as http server (-WWW and -HTTP)"},
+    {"no_CANames", OPT_NOCANAMES, '-',
```
t8m, line 955: Use `no_ca_names` please. And please  rename the variable as well.

NMorozxov: fixed

NMorozxov:
@t8m fix commited

t8m:
My approval is pending until the CI passes.

NMorozxov:
Very strange travis job fails, can you restart it

beldmit:
Travis is happy.

apps/s_server.c
richsalz: Using no-CA{file,path,store} is common -- 27 cases, so please use that convention.

beldmit: Well, this option has a different scope so I'm not sure that a convention is relevant here.

richsalz: you are correct and I was wrong.

kaduk:
> > @kaduk, I am not sure I understand which other avenues you mean.
> 
> Mostly it relates to "Check the users after TLS connection established is prohibited by moderator." being a policy (non-technical) requirement. I don't know why it's there, or whether it would also prohibit checking the users during (not after) TLS connection establishment. Based on my current knowledge it appears as just an arbitrary unneeded restriction.
> There's also an option to tell people with old, unsupported, non-protocol-compliant windows clients that they will need to find something that implements the spec properly, though I do not actually expect anyone to pick this route.
> I also don't know the details of the certification process, who is eligible to get certified, how many stale entries there are in the presented corpus, or whether it's practical to enforce a consolidation to a small subset of the certified CAs.

In particular, it seems like it should be possible to have the initial SSL_CTX/SSL configuration not include CA names and only load the CA configuration during the verify callback.  This would result in no configured CAs while the CertificateRequest is being constructed (and so the extension would not be sent), but (IIUC) would still allow using the configured CAs in the vallidation process.  Yes, it's pretty messy, and would require a bit of care to avoid a lot of IO on each connection, but it seems technically possible.

beldmit:
It seems to be an overengineering to me :(

NMorozxov:
For me my option looks like a automation for this
> > > @kaduk, I am not sure I understand which other avenues you mean.
> > 
> > 
> > Mostly it relates to "Check the users after TLS connection established is prohibited by moderator." being a policy (non-technical) requirement. I don't know why it's there, or whether it would also prohibit checking the users during (not after) TLS connection establishment. Based on my current knowledge it appears as just an arbitrary unneeded restriction.
> > There's also an option to tell people with old, unsupported, non-protocol-compliant windows clients that they will need to find something that implements the spec properly, though I do not actually expect anyone to pick this route.
> > I also don't know the details of the certification process, who is eligible to get certified, how many stale entries there are in the presented corpus, or whether it's practical to enforce a consolidation to a small subset of the certified CAs.
> 
> In particular, it seems like it should be possible to have the initial SSL_CTX/SSL configuration not include CA names and only load the CA configuration during the verify callback. This would result in no configured CAs while the CertificateRequest is being constructed (and so the extension would not be sent), but (IIUC) would still allow using the configured CAs in the vallidation process. Yes, it's pretty messy, and would require a bit of care to avoid a lot of IO on each connection, but it seems technically possible.

beldmit:
Merged. Thanks!

PR #11710, opened 2020-05-02
PSS pack: Add provider support for PSS parameters

levitte:
PSS parameter support is sorely lacking in our providers.  This PR seeks to remedy that.
Please read the separate commits for better understanding.


levitte:
This is currently incomplete, the EVP_SIGNATURE implementation still needs to be reworked, and serialization doesn't quite kick in as I expected.

In spite of that, *early comments welcome!*

levitte:
The serializer issue was a non-issue, apart from a missing dependency...

crypto/rsa/rsa_ameth.c
```diff
@@ -1104,65 +1112,34 @@ static int rsa_pkey_export_to(const EVP_PKEY *from, void *to_keydata,
...
-        }
-
-        for (i = 0; i < numcoeffs && rsa_mp_coeff_names[i] != NULL; i++) {
-            const BIGNUM *num = sk_BIGNUM_const_value(coeffs, i);
-
-            if (!OSSL_PARAM_BLD_push_BN(tmpl, rsa_mp_coeff_names[i], num))
-                goto err;
-        }
+    /*
+     * These methods only use rsa->pss.
```
slontis, line 1126: I have no idea what this comment means

crypto/rsa/rsa_backend.c
```diff
@@ -7,10 +7,18 @@
...
+#include <string.h>
 #include <openssl/core_names.h>
 #include <openssl/params.h>
+#include <openssl/evp.h>
+#include "internal/sizes.h"
+#include "internal/param_build_set.h"
 #include "crypto/rsa.h"
 
+#ifdef _WIN32
+# define strcasecmp _stricmp
```
slontis, line 19: Why are you doing this when it is defined somewhere already?

crypto/rsa/rsa_lib.c
```diff
@@ -185,6 +186,13 @@ int RSA_up_ref(RSA *r)
     return i > 1 ? 1 : 0;
 }
 
+#ifndef FIPS_MODULE
```
slontis, line 189: I dont think we do the ifdef elsewhere for other key types??

crypto/rsa/rsa_lib.c
```diff
@@ -637,7 +645,32 @@ const BIGNUM *RSA_get0_iqmp(const RSA *r)
...
+#else
     return r->pss;
+#endif
+}
+
+/* Internal */
+const RSA_PSS_PARAMS_30 *rsa_get0_pss_params_30(RSA *r)
+{
+#ifdef FIPS_MODULE
+    return NULL;
```
slontis, line 659: ???

crypto/rsa/rsa_local.h
```diff
@@ -50,8 +61,17 @@ struct rsa_st {
...
     BIGNUM *iqmp;
-    /* If a PSS only key this contains the parameter restrictions */
+#ifndef FIPS_MODULE
+    /*
+     * If a PSS only key this contains the parameter restrictions.
+     * There are two structures for the same thing, used in different cases.
+     */
+
+    /* This is used uniquely by OpenSSL provider implementations. */
+    RSA_PSS_PARAMS_30 *pss_params;
```
slontis, line 71: Shouldnt this be outside the #ifndef??

levitte:
@slontis, I'm confused, I seem to have imagined that RSASSA-PSS shouldn't be part of the fips module. Thanks for pointing it out.

crypto/rsa/rsa_lib.c
levitte: Fixed

crypto/rsa/rsa_lib.c
levitte: Fixed

crypto/rsa/rsa_local.h
levitte: Yes.  Fixed

crypto/rsa/rsa_backend.c
```diff
@@ -97,3 +105,152 @@ int rsa_fromdata(RSA *rsa, const OSSL_PARAM params[])
...
+    }
+
+    if (*rsa_pss_params == NULL)
+        pss_params = alloc_pss_params = rsa_pss_params_30_new();
+    else
+        pss_params = *rsa_pss_params;
+    if (pss_params == NULL)
+        return 0;
+
+    if ((param_md != NULL && !OSSL_PARAM_get_utf8_string(param_md, &pmdname,
```
slontis, line 225: I think I prefer checking the type and pointing to the string data directly without having to copy to a temporary local buffer for these.. I wonder if there should be an API variant to do this common case..

crypto/rsa/rsa_local.h
```diff
@@ -50,8 +61,18 @@ struct rsa_st {
...
+    /*
+     * If a PSS only key this contains the parameter restrictions.
+     * There are two structures for the same thing, used in different cases.
+     */
+    /* This is used uniquely by OpenSSL provider implementations. */
+    RSA_PSS_PARAMS_30 *pss_params;
+#ifndef FIPS_MODULE
+    /* This is used uniquely by rsa_ameth.c and rsa_pmeth.c. */
     RSA_PSS_PARAMS *pss;
+#endif
```
slontis, line 74: Maybe just collapse this by removing line 74 & 76

crypto/rsa/rsa_pss.c
```diff
@@ -256,6 +256,66 @@ int RSA_padding_add_PKCS1_PSS_mgf1(RSA *rsa, unsigned char *EM,
...
+    rsa_pss_params->mask_gen.hash_algorithm_nid = maskgenhashalg_nid;
+    rsa_pss_params->salt_len = saltlen;
+    rsa_pss_params->trailer_field = trailerfield;
+    return 1;
+}
+
+int rsa_pss_params_30_hashalg(const RSA_PSS_PARAMS_30 *rsa_pss_params)
+{
+    if (rsa_pss_params == NULL)
+        return 0;
```
slontis, line 287: NID_undef

crypto/rsa/rsa_schemes.c
```diff
@@ -0,0 +1,86 @@
...
+    size_t i;
+
+    if (meth != NULL)
+        for (i = 0; i < items_n; i++)
+            if (meth_is_a(meth, items[i].ptr))
+                return (int)items[i].id;
+    return NID_undef;
+}
+
+static const char *nid2name(int meth, const OSSL_ITEM *items, size_t items_n)
```
slontis, line 30: meth is probably not the best name here..

crypto/rsa/rsa_schemes.c
```diff
@@ -0,0 +1,86 @@
...
+ *     { OID id-sha224     PARAMETERS NULL }|
+ *     { OID id-sha256     PARAMETERS NULL }|
+ *     { OID id-sha384     PARAMETERS NULL }|
+ *     { OID id-sha512     PARAMETERS NULL }|
+ *     { OID id-sha512-224 PARAMETERS NULL }|
+ *     { OID id-sha512-256 PARAMETERS NULL },
+ *     ...  -- Allows for future expansion --
+ * }
+ */
+static const OSSL_ITEM oaeppss_name_nid_map[] = {
```
slontis, line 55: Hmmm I am looking at this old file https://tools.ietf.org/rfc/rfc6931.txt which seems to have 
2.3.10.  RSASSA-PSS without Parameters
      http://www.w3.org/2007/05/xmldsig-more#sha3-224-rsa-MGF1
      http://www.w3.org/2007/05/xmldsig-more#sha3-256-rsa-MGF1
      http://www.w3.org/2007/05/xmldsig-more#sha3-384-rsa-MGF1
      http://www.w3.org/2007/05/xmldsig-more#sha3-512-rsa-MGF1
      http://www.w3.org/2007/05/xmldsig-more#md2-rsa-MGF1
      http://www.w3.org/2007/05/xmldsig-more#md5-rsa-MGF1
      http://www.w3.org/2007/05/xmldsig-more#sha1-rsa-MGF1
      http://www.w3.org/2007/05/xmldsig-more#sha224-rsa-MGF1
      http://www.w3.org/2007/05/xmldsig-more#sha256-rsa-MGF1
      http://www.w3.org/2007/05/xmldsig-more#sha384-rsa-MGF1
      http://www.w3.org/2007/05/xmldsig-more#sha512-rsa-MGF1
      http://www.w3.org/2007/05/xmldsig-more#ripemd128-rsa-MGF1
      http://www.w3.org/2007/05/xmldsig-more#ripemd160-rsa-MGF1
      http://www.w3.org/2007/05/xmldsig-more#whirlpool-rsa-MGF1

slontis: These are all dead links now.. (but I would assume at least sh3 should be there)

providers/implementations/signature/rsa.c
```diff
@@ -201,14 +218,48 @@ static int rsa_signature_init(void *vprsactx, void *vrsa)
...
+    case RSA_FLAG_TYPE_RSA:
+        prsactx->pad_mode = RSA_PKCS1_PADDING;
+        break;
+    case RSA_FLAG_TYPE_RSASSAPSS:
+        prsactx->pad_mode = RSA_PKCS1_PSS_PADDING;
+
+        {
+            const RSA_PSS_PARAMS_30 *pss =
+                rsa_get0_pss_params_30(prsactx->rsa);
+            int md_nid = NID_sha1, mgf1md_nid = NID_sha1;
```
slontis, line 236: Maybe add #defines to the top of the file for these default values.

providers/implementations/signature/rsa.c
```diff
@@ -177,6 +177,23 @@ static int rsa_check_padding(int mdnid, int padding)
...
 }
 
+static int rsa_check_parameters(EVP_MD *md, PROV_RSA_CTX *prsactx)
+{
+    if (prsactx->pad_mode == RSA_PKCS1_PSS_PADDING) {
+        int max_saltlen;
+
+        /* See if minimum salt length exceeds maximum possible */
+        max_saltlen = RSA_size(prsactx->rsa) - EVP_MD_size(md);
+        if ((RSA_bits(prsactx->rsa) & 0x7) == 1)
```
slontis, line 187: So if there were 9 bits it subtracts 1, but for 10..15 bits is doesnt?

providers/implementations/serializers/serializer_rsa.c
```diff
@@ -109,10 +111,160 @@ int ossl_prov_print_rsa(BIO *out, RSA *rsa, int priv)
...
+    case RSA_FLAG_TYPE_RSA:
+        if (rsa_get0_pss_params_30(rsa) != NULL) {
+            if (ossl_prov_bio_printf(out, "(INVALID PSS PARAMETERS)\n") <= 0)
+                goto err;
+        }
+        break;
+    case RSA_FLAG_TYPE_RSASSAPSS:
+        {
+            const RSA_PSS_PARAMS_30 *pss_params = rsa_get0_pss_params_30(rsa);
+            int hashalg_nid = NID_sha1;
```
slontis, line 125: I see those hardwired defaults in this file also in multiple places - so they should probably be in a header and shared

providers/implementations/serializers/serializer_rsa.c
```diff
@@ -109,10 +111,160 @@ int ossl_prov_print_rsa(BIO *out, RSA *rsa, int priv)
...
+                                     (maskgenalg_nid == NID_mgf1
+                                      && maskgenhashalg_nid == NID_sha1
+                                      ? " (default)" : "")) <= 0)
+                goto err;
+            if (ossl_prov_bio_printf(out, "  Minimum Salt Length: %d%s\n",
+                                     saltlen,
+                                     (saltlen == 20 ? " (default)" : "")) <= 0)
+                goto err;
+            /*
+             * TODO(3.0) Should we show the ASN.1 trailerField value, or
```
slontis, line 159: Looks like it is only ever this value - so it is probably not needed.

providers/implementations/keymgmt/rsa_kmgmt.c
```diff
@@ -431,14 +488,38 @@ static void *rsa_gen(void *genctx, OSSL_CALLBACK *osslcb, void *cbarg)
...
+    RSA_set_flags(rsa_tmp, gctx->rsa_type);
+
+    switch (gctx->rsa_type & RSA_FLAG_TYPE_MASK) {
+    case RSA_FLAG_TYPE_RSA:
+        break;
+    case RSA_FLAG_TYPE_RSASSAPSS:
+        if (gctx->pss_params == NULL) {
+            if ((gctx->pss_params = rsa_pss_params_30_new()) == NULL
+                || !rsa_pss_params_30_set0_items(gctx->pss_params, NID_sha1,
+                                                 NID_mgf1, NID_sha1, 20, 1))
```
slontis, line 506: Ahh the famous default values are in yet another place.. :)

crypto/rsa/rsa_ameth.c
```diff
@@ -1104,65 +1112,34 @@ static int rsa_pkey_export_to(const EVP_PKEY *from, void *to_keydata,
...
+     * rsa->pss_params is used by our providers only
+     */
+    if (rsa->pss != NULL) {
+        const EVP_MD *md = NULL, *mgf1md = NULL;
+        int min_saltlen = 20;
```
slontis, line 1131: magic default number

crypto/rsa/rsa_schemes.c
slontis: Also will have to figure out if a seperate table is needed for oeap

levitte: Uhm, no, not here.  Note that this is the set of digests specifically permitted for restricted OAEP and PSS use, not the list of digests for general PKCS#1 use.

providers/implementations/signature/rsa.c
levitte: Frankly, I have *no* idea why this is done, I copied this from here:

https://github.com/openssl/openssl/blob/e307e616f25a6b7b0f343fc1e62a35b2cba888f3/crypto/rsa/rsa_pmeth.c#L829-L836

So if you tell me this is wrong, I'll take your word for it, but in that case, we should look at the rsa_pmeth.c code as well.

providers/implementations/serializers/serializer_rsa.c
levitte: *So far*, yes.

crypto/rsa/rsa_pss.c
```diff
@@ -256,6 +256,142 @@ int RSA_padding_add_PKCS1_PSS_mgf1(RSA *rsa, unsigned char *EM,
...
+ *             saltLength          20,
+ *             trailerField        trailerFieldBC
+ *         }
+ *     }
+ *
+ *     RSASSA-AlgorithmIdentifier ::= AlgorithmIdentifier {
+ *         {PKCS1Algorithms}
+ *     }
+ */
+static const RSA_PSS_PARAMS_30 default_RSASSA_PSS_params = {
```
slontis, line 281: Nice!

crypto/rsa/rsa_pss.c
levitte: I haad a different idea...

crypto/rsa/rsa_schemes.c
levitte: I have no ideaa why RFC 6931 specifies all those hashes while referring to RFC 3447.  The latter specifies a much small set: https://tools.ietf.org/html/rfc3447#appendix-A.2.1

providers/implementations/signature/rsa.c
levitte: Done

providers/implementations/serializers/serializer_rsa.c
levitte: Done

crypto/rsa/rsa_pss.c
levitte: Thanks! :-)

levitte:
Note that this depends on #11750.  With that merged in, all tests succeed in my setup, except test_ssl_new...

crypto/rsa/rsa_backend.c
levitte: e_os.h is a mess, it seems unnecessary to include it for just this.

crypto/rsa/rsa_backend.c
levitte: Done

crypto/rsa/rsa_local.h
levitte: Different topics, I think I prefer to keep them separate despite the physical closeness

crypto/rsa/rsa_backend.c
richsalz: bleah, that's a bad reason.  fix e_os.h if that's the core problem, don't monkey-patch around it.

levitte: Sure, it's in my backlog, after the stuff that I need to get ready before beta

richsalz: But the point is that this simple three-line hack adds to technical debt.  You know have to find all the cases where the define is used.

And especially, if you intend to work on e_os.h then this is misguided.

Yes, this is a lot of commentary for a simple hack.  That's what makes us programmers ;)

mattcaswell:
I've taken a look at the test_ssl_new failure. It's failing in the cert selection tests which use a PSS cert. Specifically:

test/certs/server-pss-cert.pem
test/certs/server-pss-key.pem

````
$ openssl x509 -in server-pss-cert.pem -noout -text
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 2 (0x2)
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: CN = Root CA
        Validity
            Not Before: Sep 14 12:58:41 2017 GMT
            Not After : Sep 15 12:58:41 2117 GMT
        Subject: CN = RSASSA-PSS
        Subject Public Key Info:
            Public Key Algorithm: rsassaPss
                RSA-PSS Public-Key: (2048 bit)
                Modulus:
                    00:9c:c8:93:7e:81:ba:b4:4b:b9:15:ac:30:75:33:
                    72:6a:0d:0d:81:19:c3:85:4c:37:46:4d:54:65:18:
                    8f:b4:49:b4:23:fa:c0:a0:5c:40:49:07:bb:ae:bb:
                    27:75:c6:c5:49:b7:44:74:fc:10:1b:f6:47:e5:ca:
                    71:55:fe:54:e7:3c:49:26:f9:fa:4d:7f:40:21:d7:
                    17:5a:9a:98:d1:5a:9f:bc:2d:62:38:1b:01:78:80:
                    d3:f5:a0:f7:27:65:85:fc:23:e6:0a:78:9f:9d:7a:
                    f0:b7:3f:0e:79:05:05:dc:8c:4b:29:4d:38:44:94:
                    be:d2:b2:bb:cd:75:97:e7:9c:6a:47:a2:75:46:d4:
                    9c:b1:85:d9:76:ea:c8:ad:56:2a:9d:7c:b7:59:f5:
                    ac:96:22:ff:ea:0c:9b:cc:68:24:f3:2d:5a:3a:02:
                    db:38:ad:df:04:f5:93:cf:bd:94:29:71:59:91:61:
                    2e:60:c5:cb:49:66:4f:c1:40:56:f3:0e:93:86:3f:
                    13:ed:9e:43:4b:11:ad:5f:91:12:c7:f2:14:4a:32:
                    b6:6a:06:2a:5c:0e:f7:3d:a0:2b:99:f7:59:a2:71:
                    f9:8d:79:13:29:0c:1b:a0:ac:8c:1a:7c:e2:20:a2:
                    16:47:7a:2c:e3:43:9d:74:01:04:b0:2f:9a:0d:42:
                    8b:c5
                Exponent: 65537 (0x10001)
                No PSS parameter restrictions
        X509v3 extensions:
            X509v3 Subject Key Identifier: 
                6F:B5:A1:A1:27:97:70:6C:E8:4D:AF:1C:57:56:C1:46:34:A9:D9:61
            X509v3 Authority Key Identifier: 
                70:7F:2E:AE:83:68:59:98:04:23:2A:CD:EB:3E:17:CD:24:DD:01:49
            X509v3 Basic Constraints: 
                CA:FALSE
            X509v3 Extended Key Usage: 
                TLS Web Server Authentication
            X509v3 Subject Alternative Name: 
                DNS:RSASSA-PSS
    Signature Algorithm: sha256WithRSAEncryption
    Signature Value:
        87:53:2b:28:ac:1d:a6:c8:d3:ed:0f:a0:7f:20:17:41:18:30:
        07:3d:79:9b:4f:1e:66:2a:7b:af:39:f3:89:e1:34:f8:fb:97:
        68:64:13:f6:33:84:99:22:93:b4:41:dd:d3:b9:89:d1:83:22:
        8d:d1:af:22:da:d9:6a:87:54:57:4a:ec:71:d0:85:77:89:54:
        7b:8b:9c:b6:2a:e3:54:c2:14:38:a0:0b:67:b1:ec:a1:b5:af:
        29:69:e8:6f:61:91:40:9f:ef:4c:1e:e9:95:47:9c:94:58:57:
        d4:5b:24:88:6e:e4:87:9d:1c:31:e7:8b:d8:ec:bb:3b:2e:1c:
        b1:fe:de:d9:a6:f9:ab:57:58:ae:2d:58:93:b1:d2:8b:f5:d7:
        1b:0a:26:bd:6c:fc:8c:6a:94:4b:d5:06:26:b4:10:10:da:e8:
        82:0e:54:7b:7d:79:bb:7d:a5:6d:1f:8d:ee:04:c1:ef:0d:ca:
        8a:bb:e9:3f:ec:33:26:08:93:a8:c2:7d:06:e1:a8:f2:f4:a8:
        5c:0d:ea:ae:60:5a:61:2f:c5:43:bc:78:81:0f:3a:25:bd:28:
        2d:1b:93:8b:29:e2:58:2f:ef:03:87:b1:63:5f:62:33:f8:fa:
        fc:84:16:33:82:18:ee:2b:8d:01:4d:c3:30:67:57:38:79:2e:
        c6:bc:a2:cc
````

Note that there are "No PSS parameter restrictions".

The test loads the cert and key via config, which ultimately calls `SSL_CTX_use_PrivateKey_file()`. Using the debugger I can see that the EVP_PKEY that this loads contains an RSA object with a NULL pss field (`pkey->pkey->rsa->pss == NULL`).

During the handshake the server attempts to sign the ServerKeyExchange. The call to `EVP_DigestSignInit_ex` here fails:

https://github.com/openssl/openssl/blob/257e9d03b028402089c9f98f3acb25ba668c09af/ssl/statem/statem_srvr.c#L2824-L2831

This init function calls `evp_pkey_export_to_provider` which calls `rsa_pss_pkey_export_to` which calls `rsa_in_export_to`. That function sees that `rsa->pss` is NULL and therefore does not attempt to populate the `RSA_PSS_PARAMS_30` structure, nor does it set `selection` to include `OSSL_KEYMGMT_SELECT_OTHER_PARAMETERS`. Therefore on the provider side `rsa_import` never calls `pss_params_fromdata`.

Eventually we arrive at the `rsa_signature_init` function and run this code:

````
            const RSA_PSS_PARAMS_30 *pss =
                rsa_get0_pss_params_30(prsactx->rsa);
            int md_nid = rsa_pss_params_30_hashalg(pss);
            int mgf1md_nid = rsa_pss_params_30_maskgenhashalg(pss);
            int min_saltlen = rsa_pss_params_30_saltlen(pss);
            const char *mdname, *mgf1mdname;

            mdname = rsa_oaeppss_nid2name(md_nid);
            mgf1mdname = rsa_oaeppss_nid2name(mgf1md_nid);
            prsactx->min_saltlen = min_saltlen;

            if (mdname == NULL || mgf1mdname == NULL)
                return 0;
````

The pss structure is empty and therefore `md_nid` is 0, and `mdname` is NULL. Therefore the function fails.

File: crypto/rsa/rsa_ameth.c [hunk 18]
```diff
@@ -1104,65 +1112,36 @@ static int rsa_pkey_export_to(const EVP_PKEY *from, void *to_keydata,
...
-        numexps = sk_BIGNUM_const_num(exps);
-        numcoeffs = sk_BIGNUM_const_num(coeffs);
+    /*
+     * These methods only use rsa->pss.
+     * rsa->pss_params is used by our providers only
+     */
+    if (rsa->pss != NULL) {
+        const EVP_MD *md = NULL, *mgf1md = NULL;
+        int md_nid, mgf1md_nid, saltlen;
+        RSA_PSS_PARAMS_30 pss_params;
```
mattcaswell, line 1132: This looks suspicious. You create a `pss_params` structure, populate it, and then AFAICT throw it away without using it.

mattcaswell:
@levitte can you re-enable the PSS tests in 20-cert-select.cnf.in which were disabled because of a lack of this support:

https://github.com/openssl/openssl/blob/257e9d03b028402089c9f98f3acb25ba668c09af/test/ssl-tests/20-cert-select.cnf.in#L603-L606

levitte:
> I've taken a look at the test_ssl_new failure. It's failing in the cert selection tests which use a PSS cert. Specifically:
[etc etc etc]

Ah, one spot where I didn't check for an unrestricted RSA-PSS key.  That's something I got in fairly late, so I kinda felt that I must have missed a spot.  Thanks for finding it!

crypto/rsa/rsa_ameth.c
levitte: The series of `rssa_psss_params_30_set_*` calls followed by a `rsa_pss_params_30_todata` do the work.

mattcaswell: Ah - I missed that.

mattcaswell:
Travis errors  are relevant.

Is this ready to come out of WIP now that #11750 is merged?

crypto/rsa/rsa_schemes.c
kaduk: RFC 3447 is of course obsoleted by RFC 8017, though even that does not include the ripemd, whirlpool, or mdN offerings.  I surmise that it was an attempt to define identifiers for thost algorithms as part of RFC 6931 itself, though defining XML identifiers for algorithms without the corresponding ASN.1 codepoint is somewhat unusual.

slontis:
@levitte?

levitte:
Rebase on a fresh master.  Unfortunately, I till have problems with test_ssl_new...

levitte:
> Rebase on a fresh master. Unfortunately, I till have problems with test_ssl_new...

.... and it's more of what @mattcaswell pointed out.  I'm clearing that up now.

levitte:
It seems that, generally speaking, we have pretty thorough key generation and signature tests in `test_ssl_new`, but not as tests that are specifically about key generation, signing and so on...

levitte:
I've spent most of today trying to find out what's going on with the `test_ssl_new` test, with zero success.  It got me to add a few more error records, though, that can't be a bad thing...  but suffice to say, I need help...  again.

levitte:
Looking again now in the morning, it occurred to me that the part of `test_sssl_new` that failed was FIPS related, so I tried this little fix, and 'lo an behold, everything now tests fine on my machine!

```
diff --git a/providers/fips/fipsprov.c b/providers/fips/fipsprov.c
index faf74831eb..9c8a5be620 100644
--- a/providers/fips/fipsprov.c
+++ b/providers/fips/fipsprov.c
@@ -469,6 +469,7 @@ static const OSSL_ALGORITHM fips_keymgmt[] = {
     { "DSA", "provider=fips,fips=yes", dsa_keymgmt_functions },
 #endif
     { "RSA:rsaEncryption", "provider=fips,fips=yes", rsa_keymgmt_functions },
+    { "RSA-PSS:RSASSA-PSS", "provider=default", rsapss_keymgmt_functions },
 #ifndef OPENSSL_NO_EC
     { "EC:id-ecPublicKey", "provider=fips,fips=yes", ec_keymgmt_functions },
     { "X25519", "provider=fips,fips=no", x25519_keymgmt_functions },
```

It worries me a bit that I don't quite understand what's going on, though.  If we only have a FIPS provider, and it doesn't implement RSA-PSS keys, any use should fall back to legacy and continue to work...  shouldn't it?

slontis:
> It worries me a bit that I don't quite understand what's going on, though. If we only have a FIPS provider, and it doesn't implement RSA-PSS keys, any use should fall back to legacy and continue to work... shouldn't it?

I think the tests were setup for FIPS testing so that only the FIPS module is loaded - so this is exactly what we want to happen. i.e- We dont want to be asking for FIPS and then magically use the default for these tests. (otherwise it will hide issues - like the one you just found).
The tests have been modified so that they chose fips algorithms when testing fips mode.

levitte:
@slontis, I didn't write "default", I wrote "legacy", i.e. the EVP_PKEY_METHODs / EVP_PKEY_ASN1_METHODs.  As far as I know, they aren't removed from libcrypto, so libcrypto should currently fall back to using the `rsa_ameth.c` / `rsa_pmeth.c` code if it can't fetch a KEYMGMT for RSASSA-PSS.

now, I understand that if the FIPS module makes an internal fetch of RSASSA-PSS KEYMGMT and fails, that will cause a problem, because internally in the FIPS module, the legacy code isn't present.  Are you saying that's what's happening, that it's the internal query that's failing?  That would certainly  explain things a little better.

slontis:
> now, I understand that if the FIPS module makes an internal fetch of RSASSA-PSS KEYMGMT and fails, that will cause a problem, because internally in the FIPS module, the legacy code isn't present. Are you saying that's what's happening, that it's the internal query that's failing? That would certainly explain things a little better.

You wrote the fetch code - so I assume you know more about how it works than I do :).
I was referring to evp_generic_fetch.

levitte:
...

Someone added `-Werror` in a build, when we really should use `--strict-warnings`

levitte:
As far as I'm concerned, this is no longer WIP

levitte:
Apparently, even with `--strict-warnings`, some Travis builds report this:

```
providers/common/der/der_rsa.c:424:28: error: variable 'der_aid_sha224Identifier' is not needed and will not be emitted [-Werror,-Wunneeded-internal-declaration]
static const unsigned char der_aid_sha224Identifier[] = {
                           ^
providers/common/der/der_rsa.c:434:28: error: variable 'der_aid_sha256Identifier' is not needed and will not be emitted [-Werror,-Wunneeded-internal-declaration]
static const unsigned char der_aid_sha256Identifier[] = {
                           ^
providers/common/der/der_rsa.c:444:28: error: variable 'der_aid_sha384Identifier' is not needed and will not be emitted [-Werror,-Wunneeded-internal-declaration]
static const unsigned char der_aid_sha384Identifier[] = {
                           ^
providers/common/der/der_rsa.c:454:28: error: variable 'der_aid_sha512Identifier' is not needed and will not be emitted [-Werror,-Wunneeded-internal-declaration]
static const unsigned char der_aid_sha512Identifier[] = {
                           ^
providers/common/der/der_rsa.c:464:28: error: variable 'der_aid_sha512_224Identifier' is not needed and will not be emitted [-Werror,-Wunneeded-internal-declaration]
static const unsigned char der_aid_sha512_224Identifier[] = {
                           ^
providers/common/der/der_rsa.c:474:28: error: variable 'der_aid_sha512_256Identifier' is not needed and will not be emitted [-Werror,-Wunneeded-internal-declaration]
static const unsigned char der_aid_sha512_256Identifier[] = {
                           ^
6 errors generated.
```

This is problem, and what's worse, not true, as these arrays are at the very least used with `sizeof()`.  I'm not at all ready to hard-code numbers instead...

paulidale:
I'm not sure that globally adding the `-Wno-unneeded-internal-declaration` option is the way to go here.  What about dropping the `static` instead.  This forces them to be emitted.

I can understand the compiler optimising away arrays where `sizeof()` is their only access.

levitte:
> I can understand the compiler optimising away arrays where `sizeof()` is their only access.

So I wonder how one can have that without defining the array the way I do.  Maybe this is a thing one can do with modern C, but not in ANSI C...

> I'm not sure that globally adding the `-Wno-unneeded-internal-declaration` option is the way to go here. What about dropping the `static` instead. This forces them to be emitted.

I was considering that...  That requires adding a declaration, which can be done, of course, just is a bit ugly.

paulidale:
How does:
```C
#include <stdlib.h>

static int x;

size_t derf(void) { return sizeof(x); }
```
compile?

paulidale:
```
$ clang -c -Wall /tmp/x.c
/tmp/x.c:3:12: warning: variable 'x' is not needed and will not be emitted [-Wunneeded-internal-declaration]
static int x;
           ^
1 warning generated.
```

paulidale:
No need for an array here.

You can't tell the size of an array without the contents.  `n * sizeof(element)` is as good as can be hoped for but determining `n` could be a challenge.

levitte:
> You can't tell the size of an array without the contents.  `n * sizeof(element)` is as good as can be hoped for but determining `n` could be a challenge.

That reasons for the arrays, which do get filled with content, is exatly to avoid having to determine `n` by hand.

levitte:
Anyway, I've done the non-static change.

crypto/rsa/rsa_ameth.c
```diff
@@ -1078,18 +1092,12 @@ static size_t rsa_pkey_dirty_cnt(const EVP_PKEY *pkey)
     return pkey->pkey.rsa->dirty_cnt;
 }
 
-DEFINE_SPECIAL_STACK_OF_CONST(BIGNUM_const, BIGNUM)
-
-static int rsa_pkey_export_to(const EVP_PKEY *from, void *to_keydata,
-                              EVP_KEYMGMT *to_keymgmt, OPENSSL_CTX *libctx,
-                              const char *propq)
+static int rsa_int_export_to(const EVP_PKEY *from, int rsa_type,
```
slontis, line 1095: I fail to see where this rsa_type is getting used in this function.. I guess it could do a sanity check maybe?

crypto/rsa/rsa_backend.c
```diff
@@ -7,10 +7,18 @@
...
 
+#include <string.h>
 #include <openssl/core_names.h>
 #include <openssl/params.h>
+#include <openssl/evp.h>
+#include "internal/sizes.h"
+#include "internal/param_build_set.h"
 #include "crypto/rsa.h"
 
+#ifdef _WIN32
```
slontis, line 18: Please do
#include "e_os.h" /* strcasecmp */

This header file is included in more than 50 other .c files. ifdef of a platform specific define is a way worse thing to start to block copy.

slontis (commented):
Apart from a few NITs this seems good.

providers/implementations/signature/rsa.c
slontis: I have no idea if it is correct.. Guess we will assume it is ok fo rnow.

crypto/rsa/rsa_schemes.c
slontis: ?

crypto/rsa/rsa_ameth.c
levitte: One could argue that we could drop `rsa_type` and trust the call path, but on other hand, we could do with some hardening too...  the latter requires more coding in the rest of the ameth functions, which I think should be done separately, it's not strictly necessary no.

I'm adding a TODO comment on this.

crypto/rsa/rsa_ameth.c
levitte: I've removed the whole comment...  it's aged and not quite necessary any more
(you have no idea how many iterations I went through with the PSS parameter handling)

crypto/rsa/rsa_backend.c
levitte: [sigh] ok, fine

levitte:
All nits addressed

crypto/rsa/rsa_ameth.c
slontis: I have some idea - I have reviewed this quite a few times :)

slontis (approved):
Good job!

levitte:
Merged

484d1a73c70000ad07b156f04368b3922f9910b7 RSA: Add RSA key types
645a541a3fdabd32cb8cbda48651f4150486189d RSA: Extract much of the rsa_pkey_export_to() code to a separate function
e9d6186e0507fb814310c5230293ff62310c5f9d RSA: Add rsa_schemes.c, to store scheme data and translator functions
15671090f46364a0e92456b32ead7b4714ae0b5e RSA: Add a less loaded PSS-parameter structure
967cc3f9390740f76f6ef3c91f2aeceab1902b19 RSA: Add PSS-parameter processing in EVP_PKEY_ASN1_METHOD functions
2275ff656c6d2043b40663686ec6627613d68318 DER writer: Add the possibility to abandon empty SEQUENCEs
36a2a551d7dd5628a3533f7f23b1f3827f9535f7 PROV: Refactor the RSA DER support
0ec36bf117b2c79f5663effd638410f1676c38dd PROV: Refactor the RSA SIGNATURE implementation for better param control
8a758e96f2865f0aee417025630626f095bb3ae3 PROV & KEYMGMT: Add PSS-parameter support in the RSA KEYMGMT implementation
ea297dca509b16190ad0a915f1324777b08ed8d8 PROV & SERIALIZER: Adapt the RSA serializers for PSS-parameters
2c6094baca6476d8b024dc7d9f461dae597ae797 EVP: For SIGNATURE operations, pass the propquery early
2d5536609ba92481daf42614a36bafb4e1d99293 PROV & SIGNATURE: Adapt the RSA signature code for PSS-parameters
e25761b10d48abb36a5863b087152be81ea28466 EVP: Refactor the RSA-PSS key generation controls for providers
106ec30bc74d5cd3086a3b959a11d73e46d7b9d8 PROV & ASYM_CIPHER: Adapt the RSA asymmetric cipher code for PSS-parameters
d59b7a54a5332303c42c3d097db5764dc809ecc9 test/evp_pkey_provided_test.c: Display first, compare after
f63f3b72949e5437a9bd6de67e9301a8c24c731e test/ssl-tests/20-cert-select.cnf.in: Re-enable RSA-PSS related tests
d49be019d257149d61b9061be83602ec51fa9812 test/recipes/15-test_rsapss.t: Add test with unrestricted signature
16e3588d98b313701a55ab1337b1d30ba7b08872 .travis.yml: never use -Werror, use --strict-warnings instead
90ad284f4e76254f8d67686ae3a5d6c576037091 PROV: make some DER AID arrays non-static, to avoid clang complaints

PR #11723, opened 2020-05-04
Use fips=yes consistently in documentation

tiran:
The documentation for ``EVP_default_properties_is_fips_enabled()`` uses
``fips=yes`` in one place and ``fips=true`` in another place. Stick to
``fips=yes`` like everywhere else.

Signed-off-by: Christian Heimes <christian@python.org>

##### Checklist

- [X] documentation is added or updated


t8m:
Merged to master as 6763f9c7e6. Thank you.

PR #11725, opened 2020-05-04
Configure: Avoid SIXTY_FOUR_BIT for linux-mips64

orgads:
This is a 32-bit ABI build (as opposed to linux64-mips64).

Setting SIXTY_FOUR_BIT breaks hardware optimizations, at least on octeon processors.


mattcaswell:
Pushed.

PR #11728, opened 2020-05-04
Fix up whitespace nits introduced by PR #11416

kaduk:
I can only surmise that I was too slow in getting my editor config on my new machine.


kaduk:
Merged to master; closing.

PR #11731, opened 2020-05-05
Remove cipher table lookup from EVP_CipherInit_ex

slontis:
This lookup should no longer be needed now that all ciphers are in the built in provider.
If it is needed for some reason then the logic should be inverted to only include the cipher that is legacy.

##### Checklist


beldmit:
Will it still work for the engine-provided ciphers?

slontis:
> Will it still work for the engine-provided ciphers?
There is still a
if (using an engine)
   goto legacy

so it should work..

slontis:
Thanks. Merged to master.

Issue #11209, opened 2020-02-29
EOF handling in s_client faliure


beldmit:
s_server:
```
openssl s_server -cert cert.pem -key key.pem -www -cipher ECDHE-RSA-AES256-SHA
```
s_client:
```
openssl s_client -connect localhost:4433 -CAfile manul-gost/sslCA/cacert.pem -state -cipher ECDHE-RSA-AES256-SHA -ign_eof

GET / \n\n
```
result:
(some html)

```
</pre></BODY></HTML>

SSL3 alert write:fatal:decode error
00:71:81:B1:E1:7F:00:00:error:SSL routines::unexpected eof while reading:ssl/record/rec_layer_s3.c:307:
```
This behavior was introduced in https://github.com/openssl/openssl/commit/db943f43a6
Found by @vt-alt

beldmit:
If I understand correctly, now we have a violation of the comment here:
https://github.com/openssl/openssl/blob/e32c608e0733d5b295c9aa119153133413c5d744/apps/s_client.c#L2939-L2940

and need a more thorough study of the error setting it to zero when BIO_eof occurs.

beldmit:
And also there is a problem with the result code of s_client, which I think should be 0 in this case.

vt-alt:
@mattcaswell @t8m
Developers should note that this is stable branch, not development branch. So, this is sorta important.

kroeckx:
I'm not sure what you're complaining about. s_client gave a very
confusing error "read:errno=0". With the current version, it
properly prints "unexpected eof while reading", which is normal.

s_server does not properly shut down a connection when it should,
it just closes the connection. So s_client should give an error
message.

If you modify s_server to properly shut down the connection, s_client
will not return an error. With a modified s_server, both the old
and new s_client will print "closed".

beldmit:
Hmmm. The idea to fix it from the server side is interesting.

mattcaswell:
> If you modify s_server to properly shut down the connection, s_client
will not return an error. With a modified s_server, both the old
and new s_client will print "closed".

Yes - this was my immediate thought too. If s_server is closing the connection without shutting it down properly then this seems more like an s_server bug than an s_client one.

beldmit:
I tried to add SSL_shutdown (in a loop) at the end of www_body, and the result appears to be the same.

beldmit:
I want to mention that the error appears when using s_client to connect to IIS-based server. So I think it should be retested and maybe rethought.

[...]

PR #11734, opened 2020-05-05
s_server normal shutdown

beldmit:
Partially fixes #11209

s_server application did not perform shutdown normally.

##### Checklist

- [ ] documentation is added or updated
- [ ] tests are added or updated

PR #11735, opened 2020-05-05
Ssl op allow eof

beldmit:
This is the 2nd part of fixing #11209

##### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,6 +30,12 @@ OpenSSL 3.0
 
    *Shane Lontis*
 
+ * The SSL option SSL_OP_IGNORE_UNEXPECTED_EOF is introduced. If that option
+   is set, an unexpected EOF is ignored, it pretends a close notify was received
+   instead and so the returned error becomes SSL_ERROR_ZERO_RETURN.
+
+   *Dmitry Belyavskiy*
+
  * Deprecated EC_POINT_set_Jprojective_coordinates_GFp() and
    EC_POINT_get_Jprojective_coordinates_GFp(). These functions are not widely
    used and applications should instead use the
--- a/apps/s_client.c
+++ b/apps/s_client.c
@@ -576,7 +576,7 @@ typedef enum OPTION_choice {
     OPT_READ_BUF, OPT_KEYLOG_FILE, OPT_EARLY_DATA, OPT_REQCAFILE,
     OPT_V_ENUM,
     OPT_X_ENUM,
-    OPT_S_ENUM,
+    OPT_S_ENUM, OPT_IGNORE_UNEXPECTED_EOF,
     OPT_FALLBACKSCSV, OPT_NOCMDS, OPT_PROXY, OPT_PROXY_USER, OPT_PROXY_PASS,
     OPT_DANE_TLSA_DOMAIN,
 #ifndef OPENSSL_NO_CT
@@ -718,6 +718,8 @@ const OPTIONS s_client_options[] = {
      "Do not send the server name (SNI) extension in the ClientHello"},
     {"tlsextdebug", OPT_TLSEXTDEBUG, '-',
      "Hex dump of all TLS extensions received"},
+    {"ignore_unexpected_eof", OPT_IGNORE_UNEXPECTED_EOF, '-',
+     "Do not treat lack of close_notify from a peer as an error"},
 #ifndef OPENSSL_NO_OCSP
     {"status", OPT_STATUS, '-', "Request certificate status from server"},
 #endif
@@ -1001,6 +1003,7 @@ int s_client_main(int argc, char **argv)
 #ifndef OPENSSL_NO_SCTP
     int sctp_label_bug = 0;
 #endif
+    int ignore_unexpected_eof = 0;
 
     FD_ZERO(&readfds);
     FD_ZERO(&writefds);
@@ -1179,6 +1182,9 @@ int s_client_main(int argc, char **argv)
         case OPT_X_CASES:
             if (!args_excert(o, &exc))
                 goto end;
+            break;
+        case OPT_IGNORE_UNEXPECTED_EOF:
+            ignore_unexpected_eof = 1;
             break;
         case OPT_PREXIT:
             prexit = 1;
@@ -1775,6 +1781,9 @@ int s_client_main(int argc, char **argv)
     if (max_version != 0
         && SSL_CTX_set_max_proto_version(ctx, max_version) == 0)
         goto end;
+
+    if (ignore_unexpected_eof)
+        SSL_CTX_set_options(ctx, SSL_OP_IGNORE_UNEXPECTED_EOF);
 
     if (vpmtouched && !SSL_CTX_set1_param(ctx, vpm)) {
         BIO_printf(bio_err, "Error setting verify params\n");
--- a/apps/s_server.c
+++ b/apps/s_server.c
@@ -761,7 +761,7 @@ typedef enum OPTION_choice {
     OPT_SRTP_PROFILES, OPT_KEYMATEXPORT, OPT_KEYMATEXPORTLEN,
     OPT_KEYLOG_FILE, OPT_MAX_EARLY, OPT_RECV_MAX_EARLY, OPT_EARLY_DATA,
     OPT_S_NUM_TICKETS, OPT_ANTI_REPLAY, OPT_NO_ANTI_REPLAY, OPT_SCTP_LABEL_BUG,
-    OPT_HTTP_SERVER_BINMODE, OPT_NOCANAMES,
+    OPT_HTTP_SERVER_BINMODE, OPT_NOCANAMES, OPT_IGNORE_UNEXPECTED_EOF,
     OPT_R_ENUM,
     OPT_S_ENUM,
     OPT_V_ENUM,
@@ -851,6 +851,8 @@ const OPTIONS s_server_options[] = {
      "Disable caching and tickets if ephemeral (EC)DH is used"},
     {"www", OPT_WWW, '-', "Respond to a 'GET /' with a status page"},
     {"WWW", OPT_UPPER_WWW, '-', "Respond to a 'GET with the file ./path"},
+    {"ignore_unexpected_eof", OPT_IGNORE_UNEXPECTED_EOF, '-',
+     "Do not treat lack of close_notify from a peer as an error"},
     {"tlsextdebug", OPT_TLSEXTDEBUG, '-',
      "Hex dump of all TLS extensions received"},
     {"HTTP", OPT_HTTP, '-', "Like -WWW but ./path includes HTTP headers"},
@@ -1095,6 +1097,7 @@ int s_server_main(int argc, char *argv[])
 #ifndef OPENSSL_NO_SCTP
     int sctp_label_bug = 0;
 #endif
+    int ignore_unexpected_eof = 0;
 
     /* Init of few remaining global variables */
     local_argc = argc;
@@ -1666,6 +1669,9 @@ int s_server_main(int argc, char *argv[])
             use_sendfile = 1;
 #endif
             break;
+        case OPT_IGNORE_UNEXPECTED_EOF:
+            ignore_unexpected_eof = 1;
+            break;
         }
     }
     argc = opt_num_rest();
@@ -1866,7 +1872,6 @@ int s_server_main(int argc, char *argv[])
             goto end;
         }
     }
-
 #ifndef OPENSSL_NO_SCTP
     if (protocol == IPPROTO_SCTP && sctp_label_bug == 1)
         SSL_CTX_set_mode(ctx, SSL_MODE_DTLS_SCTP_LABEL_LENGTH_BUG);
@@ -1909,6 +1914,9 @@ int s_server_main(int argc, char *argv[])
     if (no_ca_names) {
         SSL_CTX_set_options(ctx, SSL_OP_DISABLE_TLSEXT_CA_NAMES);
     }
+
+    if (ignore_unexpected_eof)
+        SSL_CTX_set_options(ctx, SSL_OP_IGNORE_UNEXPECTED_EOF);
 
     if (max_send_fragment > 0
         && !SSL_CTX_set_max_send_fragment(ctx, max_send_fragment)) {
--- a/doc/man3/SSL_CTX_set_options.pod
+++ b/doc/man3/SSL_CTX_set_options.pod
@@ -217,6 +217,20 @@ not propose, and servers will not accept the extension.
 Disable all renegotiation in TLSv1.2 and earlier. Do not send HelloRequest
 messages, and ignore renegotiation requests via ClientHello.
 
+=item SSL_OP_IGNORE_UNEXPECTED_EOF
+
+Some TLS implementations do not send the mandatory close_notify alert on
+shutdown. If the application tries to wait for the close_notify alert but the
+peer closes the connection without sending it, an error is generated. When this
+option is enabled the peer does not need to send the close_notify alert and a
+closed connection will be treated as if the close_notify alert was received.
+
+You should only enable this option if the protocol running over TLS
+can detect a truncation attack itself, and that the application is checking for
+that truncation attack.
+
+For more information on shutting down a connection, see L<SSL_shutdown(3)>.
+
 =item SSL_OP_ALLOW_NO_DHE_KEX
 
 In TLSv1.3 allow a non-(ec)dhe based key exchange mode on resumption. This means
@@ -367,7 +381,7 @@ secure renegotiation and 0 if it does not.
 
 =head1 SEE ALSO
 
-L<ssl(7)>, L<SSL_new(3)>, L<SSL_clear(3)>,
+L<ssl(7)>, L<SSL_new(3)>, L<SSL_clear(3)>, L<SSL_shutdown(3)>
 L<SSL_CTX_set_tmp_dh_callback(3)>,
 L<SSL_CTX_set_min_proto_version(3)>,
 L<openssl-dhparam(1)>
@@ -380,7 +394,8 @@ OpenSSL 0.9.8m.
 The B<SSL_OP_PRIORITIZE_CHACHA> and B<SSL_OP_NO_RENEGOTIATION> options
 were added in OpenSSL 1.1.1.
 
-The B<SSL_OP_NO_EXTENDED_MASTER_SECRET> option was added in OpenSSL 3.0.
+The B<SSL_OP_NO_EXTENDED_MASTER_SECRET> and B<SSL_OP_IGNORE_UNEXPECTED_EOF>
+options were added in OpenSSL 3.0.
 
 =head1 COPYRIGHT
 
--- a/doc/man3/SSL_get_error.pod
+++ b/doc/man3/SSL_get_error.pod
@@ -25,6 +25,15 @@ other OpenSSL function calls should appear in between.  The current
 thread's error queue must be empty before the TLS/SSL I/O operation is
 attempted, or SSL_get_error() will not work reliably.
 
+=head1 NOTES
+
+Some TLS implementations do not send a close_notify alert on shutdown.
+
+On an unexpected EOF, versions before OpenSSL 3.0 returned
+B<SSL_ERROR_SYSCALL>, nothing was added to the error stack, and errno was 0.
+Since OpenSSL 3.0 the returned error is B<SSL_ERROR_SSL> with a meaningful
+error on the error stack.
+
 =head1 RETURN VALUES
 
 The following return values can currently occur:
@@ -43,6 +52,9 @@ close_notify alert.
 No more data can be read.
 Note that B<SSL_ERROR_ZERO_RETURN> does not necessarily
 indicate that the underlying transport has been closed.
+
+This error can also appear when the option B<SSL_OP_IGNORE_UNEXPECTED_EOF>
+is set. See L<SSL_CTX_set_options(3)> for more details.
 
 =item SSL_ERROR_WANT_READ, SSL_ERROR_WANT_WRITE
 
--- a/doc/man3/SSL_shutdown.pod
+++ b/doc/man3/SSL_shutdown.pod
@@ -83,8 +83,10 @@ message, otherwise an unexpected EOF will be reported.
 There are implementations that do not send the required close_notify alert.
 If there is a need to communicate with such an implementation, and it's clear
 that all data has been received, do not wait for the peer's close_notify alert.
-Waiting for the close_notify alert when the peer just closes the connection will
-result in an error being generated.
+Waiting for the close_notify alert when the peer just closes the connection
+will result in an error being generated.
+The error can be ignored using the B<SSL_OP_IGNORE_UNEXPECTED_EOF>.
+For more information see L<SSL_CTX_set_options(3)>.
 
 =head2 First to close the connection
 
@@ -159,7 +161,7 @@ It can also occur when not all data was read using SSL_read().
 
 L<SSL_get_error(3)>, L<SSL_connect(3)>,
 L<SSL_accept(3)>, L<SSL_set_shutdown(3)>,
-L<SSL_CTX_set_quiet_shutdown(3)>,
+L<SSL_CTX_set_quiet_shutdown(3)>, L<SSL_CTX_set_options(3)>
 L<SSL_clear(3)>, L<SSL_free(3)>,
 L<ssl(7)>, L<bio(7)>
 
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -324,14 +324,9 @@ typedef int (*SSL_async_callback_fn)(SSL *s, void *arg);
 /* Allow initial connection to servers that don't support RI */
 # define SSL_OP_LEGACY_SERVER_CONNECT                    0x00000004U
 
-/* Reserved value (until OpenSSL 3.0.0)                  0x00000008U */
 # define SSL_OP_TLSEXT_PADDING                           0x00000010U
-/* Reserved value (until OpenSSL 3.0.0)                  0x00000020U */
 # define SSL_OP_SAFARI_ECDHE_ECDSA_BUG                   0x00000040U
-/*
- * Reserved value (until OpenSSL 3.0.0)                  0x00000080U
- * Reserved value (until OpenSSL 3.0.0)                  0x00000100U
- */
+# define SSL_OP_IGNORE_UNEXPECTED_EOF                    0x00000080U
 
 # define SSL_OP_DISABLE_TLSEXT_CA_NAMES                  0x00000200U
 
--- a/test/sslapitest.c
+++ b/test/sslapitest.c
@@ -6673,6 +6673,61 @@ static int test_ticket_callbacks(int tst)
 }
 
 /*
+ * Test incorrect shutdown.
+ * Test 0: client does not shutdown properly,
+ *         server does not set SSL_OP_IGNORE_UNEXPECTED_EOF,
+ *         server should get SSL_ERROR_SSL
+ * Test 1: client does not shutdown properly,
+ *         server sets SSL_OP_IGNORE_UNEXPECTED_EOF,
+ *         server should get SSL_ERROR_ZERO_RETURN
+ */
+static int test_incorrect_shutdown(int tst)
+{
+    SSL_CTX *cctx = NULL, *sctx = NULL;
+    SSL *clientssl = NULL, *serverssl = NULL;
+    int testresult = 0;
+    char buf[80];
+    BIO *c2s;
+
+    if (!TEST_true(create_ssl_ctx_pair(libctx, TLS_server_method(),
+                                       TLS_client_method(), 0, 0,
+                                       &sctx, &cctx, cert, privkey)))
+        goto end;
+
+    if (tst == 1)
+        SSL_CTX_set_options(sctx, SSL_OP_IGNORE_UNEXPECTED_EOF);
+
+    if (!TEST_true(create_ssl_objects(sctx, cctx, &serverssl, &clientssl,
+                                            NULL, NULL)))
+        goto end;
+
+    if (!TEST_true(create_ssl_connection(serverssl, clientssl,
+                                              SSL_ERROR_NONE)))
+        goto end;
+
+    c2s = SSL_get_rbio(serverssl);
+    BIO_set_mem_eof_return(c2s, 0);
+
+    if (!TEST_false(SSL_read(serverssl, buf, sizeof(buf))))
+        goto end;
+
+    if (tst == 0 && !TEST_int_eq(SSL_get_error(serverssl, 0), SSL_ERROR_SSL) )
+        goto end;
+    if (tst == 1 && !TEST_int_eq(SSL_get_error(serverssl, 0), SSL_ERROR_ZERO_RETURN) )
+        goto end;
+
+    testresult = 1;
+
+ end:
+    SSL_free(serverssl);
+    SSL_free(clientssl);
+    SSL_CTX_free(sctx);
+    SSL_CTX_free(cctx);
+
+    return testresult;
+}
+
+/*
  * Test bi-directional shutdown.
  * Test 0: TLSv1.2
  * Test 1: TLSv1.2, server continues to read/write after client shutdown
@@ -7685,6 +7740,7 @@ int setup_tests(void)
     ADD_ALL_TESTS(test_ssl_get_shared_ciphers, OSSL_NELEM(shared_ciphers_data));
     ADD_ALL_TESTS(test_ticket_callbacks, 16);
     ADD_ALL_TESTS(test_shutdown, 7);
+    ADD_ALL_TESTS(test_incorrect_shutdown, 2);
     ADD_ALL_TESTS(test_cert_cb, 6);
     ADD_ALL_TESTS(test_client_cert_cb, 2);
     ADD_ALL_TESTS(test_ca_names, 3);
--- a/doc/man1/openssl-s_client.pod.in
+++ b/doc/man1/openssl-s_client.pod.in
@@ -78,6 +78,7 @@ B<openssl> B<s_client>
 [B<-split_send_frag>]
 [B<-max_pipelines>]
 [B<-read_buf>]
+[B<-ignore_unexpected_eof>]
 [B<-bugs>]
 [B<-comp>]
 [B<-no_comp>]
@@ -576,6 +577,15 @@ The default read buffer size to be used for connections. This will only have an
 effect if the buffer size is larger than the size that would otherwise be used
 and pipelining is in use (see L<SSL_CTX_set_default_read_buffer_len(3)> for
 further information).
+
+=item B<-ignore_unexpected_eof>
+
+Some TLS implementations do not send the mandatory close_notify alert on
+shutdown. If the application tries to wait for the close_notify alert but the
+peer closes the connection without sending it, an error is generated. When this
+option is enabled the peer does not need to send the close_notify alert and a
+closed connection will be treated as if the close_notify alert was received.
+For more information on shutting down a connection, see L<SSL_shutdown(3)>.
 
 =item B<-bugs>
 
--- a/doc/man1/openssl-s_server.pod.in
+++ b/doc/man1/openssl-s_server.pod.in
@@ -47,6 +47,7 @@ B<openssl> B<s_server>
 [B<-WWW>]
 [B<-http_server_binmode>]
 [B<-no_ca_names>]
+[B<-ignore_unexpected_eof>]
 [B<-servername>]
 [B<-servername_fatal>]
 [B<-tlsextdebug>]
@@ -417,6 +418,15 @@ by the client in binary mode.
 Disable TLS Extension CA Names. You may want to disable it for security reasons
 or for compatibility with some Windows TLS implementations crashing when this
 extension is larger than 1024 bytes.
+
+=item B<-ignore_unexpected_eof>
+
+Some TLS implementations do not send the mandatory close_notify alert on
+shutdown. If the application tries to wait for the close_notify alert but the
+peer closes the connection without sending it, an error is generated. When this
+option is enabled the peer does not need to send the close_notify alert and a
+closed connection will be treated as if the close_notify alert was received.
+For more information on shutting down a connection, see L<SSL_shutdown(3)>.
 
 =item B<-id_prefix> I<val>
 
--- a/ssl/record/rec_layer_s3.c
+++ b/ssl/record/rec_layer_s3.c
@@ -303,8 +303,13 @@ int ssl3_read_n(SSL *s, size_t n, size_t max, int extend, int clearold,
             if (ret <= 0
                     && !BIO_should_retry(s->rbio)
                     && BIO_eof(s->rbio)) {
-                SSLfatal(s, SSL_AD_DECODE_ERROR, SSL_F_SSL3_READ_N,
-                         SSL_R_UNEXPECTED_EOF_WHILE_READING);
+                if (s->options & SSL_OP_IGNORE_UNEXPECTED_EOF) {
+                    SSL_set_shutdown(s, SSL_RECEIVED_SHUTDOWN);
+                    s->s3.warn_alert = SSL_AD_CLOSE_NOTIFY;
+                } else {
+                    SSLfatal(s, SSL_AD_DECODE_ERROR, SSL_F_SSL3_READ_N,
+                             SSL_R_UNEXPECTED_EOF_WHILE_READING);
+                }
             }
         } else {
             SSLfatal(s, SSL_AD_INTERNAL_ERROR, SSL_F_SSL3_READ_N,


doc/man3/SSL_CONF_cmd.pod
```diff
@@ -85,6 +85,13 @@ its preference list. This usually indicates a client without AES hardware
 acceleration (e.g. mobile) is in use. Equivalent to B<SSL_OP_PRIORITIZE_CHACHA>.
 Only used by servers. Requires B<-serverpref>.
 
+=item B<-allow_ssl_eof>
+
+Some TLS implementation (such as L<openssl-s_server(1)> before the OpenSSL 3.0
+or IIS) do not send close_notify on shutdown.  Such behaviour is treated as
```
mattcaswell, line 91: I would drop the information in the parens here. Such things tend to get out-of-date very quickly.

s/implementation/implementations/

doc/man3/SSL_CONF_cmd.pod
```diff
@@ -85,6 +85,13 @@ its preference list. This usually indicates a client without AES hardware
...
+error since OpenSSL 3.0. Enabling this option allows to treat such behaviour as
```
mattcaswell, line 92: Strictly speaking its still an error even if using this option. It always was. It was just historically reported as SSL_ERROR_SYSCALL with a 0 errno (which is against the documented behaviour for that SSL_ERROR_SYSCALL). What we have changed is that the error is now reported as SSL_ERROR_SSL with a meaningful error on the error stack. So really this option is about getting the legacy error reporting back rather than the new error reporting. The docs should probably explain all of this. I'd also suggest calling the option SSL_OP_LEGACY_SSL_EOF, or similar. It's not really about allowing it - just changing how it is handled.

include/openssl/ssl.h
```diff
@@ -328,8 +328,9 @@ typedef int (*SSL_async_callback_fn)(SSL *s, void *arg);
 # define SSL_OP_TLSEXT_PADDING                           0x00000010U
 /* Reserved value (until OpenSSL 3.0.0)                  0x00000020U */
 # define SSL_OP_SAFARI_ECDHE_ECDSA_BUG                   0x00000040U
+/* Reserved value (until OpenSSL 3.0.0)                  0x00000080U */
+# define SSL_OP_ALLOW_SSL_EOF                            0x00000080U
```
mattcaswell, line 332: All these "Reserved value (until OpenSSL 3.0.0)" comments can be removed. The next release *is* 3.0 and therefore these values are no longer reserved!! They were only there for ABI compatibility - but we're not concerned with ABI compatibility for 3.0.

doc/man3/SSL_CONF_cmd.pod
beldmit: Fixed.

doc/man3/SSL_CONF_cmd.pod
beldmit: Fixed.

include/openssl/ssl.h
beldmit: Done.

apps/include/opt.h
```diff
@@ -185,6 +186,8 @@
             "Disallow session resumption on renegotiation"}, \
         {"no_legacy_server_connect", OPT_S_NOLEGACYCONN, '-', \
             "Disallow initial connection to servers that don't support RI"}, \
+        {"legacy_ssl_eof", OPT_S_LEGACY_SSL_EOF, '-', \
+            "Do not treat lack of close_notify as error"}, \
```
mattcaswell, line 190: Use legacy error handling if there is no close_notify

doc/man3/SSL_CONF_cmd.pod
```diff
@@ -85,6 +85,15 @@ its preference list. This usually indicates a client without AES hardware
 acceleration (e.g. mobile) is in use. Equivalent to B<SSL_OP_PRIORITIZE_CHACHA>.
 Only used by servers. Requires B<-serverpref>.
 
+=item B<-legacy_ssl_eof>
+
+Some TLS implementations do not send close_notify on shutdown. Before OpenSSL
+3.0 the SSL_ERROR_SYSCALL with errno=0 was reported in such cases. This
```
mattcaswell, line 91: Too many "the"s:
Before OpenSSL 3.0, SSL_ERROR_SYSCALL

doc/man3/SSL_CONF_cmd.pod
```diff
@@ -85,6 +85,15 @@ its preference list. This usually indicates a client without AES hardware
...
+behaviour cotradicts the B<SSL_ERROR_SYSCALL> documentation. Since OpenSSL 3.0
+the returned error is B<SSL_ERROR_SSL> with a meaningful error on the error
+stack. This option allows to get the legacy error in such cases. Equivalent to
```
mattcaswell, line 94: This option gives the legacy error in such cases.

doc/man3/SSL_CONF_cmd.pod
```diff
@@ -484,6 +493,9 @@ B<EncryptThenMac>: use encrypt-then-mac extension, enabled by
 default. Inverse of B<SSL_OP_NO_ENCRYPT_THEN_MAC>: that is,
 B<-EncryptThenMac> is the same as setting B<SSL_OP_NO_ENCRYPT_THEN_MAC>.
 
+B<LegacySSLEOF>: do not treat lack of peer's close_notify on shutdown as error.
```
mattcaswell, line 496: use legacy error handling if the peer does not send close_notify

doc/man3/SSL_CTX_set_options.pod
```diff
@@ -211,6 +211,12 @@ not propose, and servers will not accept the extension.
 Disable all renegotiation in TLSv1.2 and earlier. Do not send HelloRequest
 messages, and ignore renegotiation requests via ClientHello.
 
+=item SSL_OP_LEGACY_SSL_EOF
+
+Some TLS implementations do not send close_notify on shutdown. Such behaviour
+is treated as error since OpenSSL 3.0. Enabling this option allows to treat
+such behaviour as normal. See more details in L<SSL_CONF_cmd(3)>.
```
mattcaswell, line 218: I'd copy the wording from -legacy_ssl_eof here.

include/openssl/ssl.h
```diff
@@ -324,12 +324,10 @@ typedef int (*SSL_async_callback_fn)(SSL *s, void *arg);
...
 
-/* Reserved value (until OpenSSL 3.0.0)                  0x00000008U */
 # define SSL_OP_TLSEXT_PADDING                           0x00000010U
-/* Reserved value (until OpenSSL 3.0.0)                  0x00000020U */
 # define SSL_OP_SAFARI_ECDHE_ECDSA_BUG                   0x00000040U
+# define SSL_OP_LEGACY_SSL_EOF                           0x00000080U
 /*
- * Reserved value (until OpenSSL 3.0.0)                  0x00000080U
  * Reserved value (until OpenSSL 3.0.0)                  0x00000100U
  * Reserved value (until OpenSSL 3.0.0)                  0x00000200U
```
mattcaswell, line 332: Might as well remove all of these

apps/include/opt.h
beldmit: Fixed. Thanks!

doc/man3/SSL_CONF_cmd.pod
beldmit: Fixed. Thanks!

doc/man3/SSL_CONF_cmd.pod
beldmit: Fixed. Thanks!

doc/man3/SSL_CTX_set_options.pod
beldmit: Fixed. Thanks!

include/openssl/ssl.h
beldmit: Done.

mattcaswell:
Oh...there probably should be a CHANGES entry for this

beldmit:
> Oh...there probably should be a CHANGES entry for this

Done. Please reapprove.

include/openssl/ssl.h
kaduk: Or maybe leave them in as "unallocated"?  I guess that might go stale more easily than just not defining something.

mattcaswell:
Looks like the travis failure is relevant

beldmit:
Fixed and Travis is happy. @mattcaswell could you please reconfirm?

beldmit:
Ping @openssl/otc

kroeckx:
I wonder if it's not more useful to have an option that instead gives a different distinct error message. I think the "SSL_ERROR_SYSCALL with errno=0" really doesn't tell much, and isn't something we want as an API that people need to check. If this is actually the API we do want, I think that special case should get documented in the SSL_get_error() manpage, and maybe also the SSL_shutdown() manpage.

I'm also wondering if we should backport this to the 1.1.1 branch. If we go with the current patch, it would just be a define and documentation for 1.1.1.

beldmit:
People from Nginx I had a conversation think that the change was not necessary at all and the best solution would be to revert this patch because otherwise they will have to support both variants.

Personally I plan to use this feature as an option to test against known "broken" (omitting close_notify) implementations.

kroeckx:
So what do the nginx people intend to do?

beldmit:
I asked them to respond here.

mdounin:
As already mentioned by Dmitry, we here at nginx don't think the change was necessary. As Matt already said above in the comments to SSL_CONF_cmd.pod change, the error was always reported. The only issue is that SSL_ERROR_SYSCALL with a 0 errno is not properly documented. On the other hand, the behaviour was present since ancient OpenSSL versions, and actually tested in various software using OpenSSL library, including nginx. A better solution, in our opinion, would be to document the error instead.

Right now the situation in OpenSSL 3.0 is that the error reporting behaviour was changed, and, if we are going to support OpenSSL 3.0, we have to introduce specific error testing for OpenSSL 3.0. And at the same time we have to support previous error reporting, since we support OpenSSL versions starting from OpenSSL 0.9.8, as well as other libraries such as BoringSSL and LibreSSL, which still report connection close with SSL_ERROR_SYSCALL with a 0 errno.

For obvious reasons we don't want to support multiple code paths to test for the same error. Especially keeping in mind that due to BoringSSL and LibreSSL we probably have to support these multiple code paths forever.

It would be really helpful if the change in question was reverted and the existing behaviour (that is, SSL_ERROR_SYSCALL with a 0 errno) was documented instead.

kroeckx:
Note that there are actually 2 differences in master:
- The error code that is returned changed
- We properly deal with it as an error, including marking the
  session as non-reusable.

For some applications, it is a real error. And we currently can't
tell if it's a real error for the application or not.

The only solutions I currently see is:
- The application somehow tells us if we should treat it like an
  error or not. The question then is what should the default be.
- The application avoids getting that error.

kroeckx:
> Note that there are actually 2 differences in master:
> - The error code that is returned changed
> - We properly deal with it as an error, including marking the
>   session as non-reusable.

The description of this patch really only talks about the first
thing, while the patch changes the behaviour of both things, and
I'm not sure if that's something we want. If it's something we
want, it also needs to be documented as such.

t8m:
I agree with @kroeckx that the documentation should be improved in this PR. I would also say that the name of the SSL_OP is not a particularly good one. Perhaps something like SSL_OP_ALLOW_UNCLEAN_EOF would be more descriptive.

On the other hand I think the SSL_OP should be provided and nginx and other HTTPS servers and clients can easily use it with an `#ifdef`. So I would not be that much in favor of making this OP a default for 3.0. At most we might think of including it into the SSL_OP_ALL.

There is IMO no point in backporting the OP to 1.1.1 as we reverted the change there and the OP would be a no-op there. :grin:

beldmit:
I'm ready to reasonably improve this PR when the consensus is reached. I'm still not sure we don't need to revert this change at all replacing with the documentation.

kroeckx:
> I'm ready to reasonably improve this PR when the consensus is reached.

I'm writing the -project list about this.

CHANGES.md
```diff
@@ -30,6 +30,11 @@ OpenSSL 3.0
 
    *Shane Lontis*
 
+ * The SSL option SSL_OP_IGNORE_UNEXPECTED_EOF restoring old behaviour when peer
+   does not send close_notify is introduced.
```
kroeckx, line 34: The behaviour you get is not the old behaviour.

kroeckx: The current implementation probably does what you document, but it's not what was discussed. If that option is set, the unexpected EOF is ignored, it pretends close notify was received instead and so returns SSL_ERROR_ZERO_RETURN.

doc/man3/SSL_CTX_set_options.pod
```diff
@@ -217,6 +217,15 @@ not propose, and servers will not accept the extension.
...
 messages, and ignore renegotiation requests via ClientHello.
 
+=item SSL_OP_IGNORE_UNEXPECTED_EOF
+
+Some TLS implementations do not send close_notify on shutdown. Before OpenSSL
+3.0, SSL_ERROR_SYSCALL with errno=0 was reported in such cases. This
+behaviour contradicts the B<SSL_ERROR_SYSCALL> documentation. Since OpenSSL 3.0
+the returned error is B<SSL_ERROR_SSL> with a meaningful error on the error
+stack. This option gives the legacy error in such cases. Equivalent to
+B<SSL_OP_IGNORE_UNEXPECTED_EOF>.
```
kroeckx, line 227: This should come with a warning that you should only enable this option if the protocol that's running over TLS can detect a truncation attack itself, and that the application is checking for that truncation attack.

CHANGES.md
beldmit: Does it mean we still need to update the code? Sorry, I don't fill familiar enough with it.

kroeckx:
This needs test suite changes that verify we actually do what we document.

The SSL_get_error() manpage should probably get updated.

CHANGES.md
kroeckx: Yes, the code needs to change. I think at least s->shutdown needs to set SSL_RECEIVED_SHUTDOWN and s->s3.warn_alert needs to be set to SSL_AD_CLOSE_NOTIFY

doc/man3/SSL_CONF_cmd.pod
```diff
@@ -85,6 +85,15 @@ its preference list. This usually indicates a client without AES hardware
 acceleration (e.g. mobile) is in use. Equivalent to B<SSL_OP_PRIORITIZE_CHACHA>.
 Only used by servers. Requires B<-serverpref>.
 
+=item B<-ignore_unexpected_eof>
+
+Some TLS implementations do not send close_notify on shutdown. Before OpenSSL
+3.0, SSL_ERROR_SYSCALL with errno=0 was reported in such cases. This
+behaviour contradicts the B<SSL_ERROR_SYSCALL> documentation. Since OpenSSL 3.0
+the returned error is B<SSL_ERROR_SSL> with a meaningful error on the error
+stack. This option gives the legacy error in such cases. Equivalent to
```
t8m, line 94: `This option gives the legacy error in such cases.` This is incorrect - SSL_ERROR_ZERO_RETURN should be returned with this option. Also please add some warning that using this options is safe only when the underlying protocol detects truncation by itself.

doc/man3/SSL_CONF_cmd.pod
```diff
@@ -484,6 +493,9 @@ B<EncryptThenMac>: use encrypt-then-mac extension, enabled by
 default. Inverse of B<SSL_OP_NO_ENCRYPT_THEN_MAC>: that is,
 B<-EncryptThenMac> is the same as setting B<SSL_OP_NO_ENCRYPT_THEN_MAC>.
 
+B<IgnoreUnexpectedEOF>: use legacy error handling if the peer does not send
```
t8m, line 496: Using the term `legacy error handling` is wrong.

doc/man3/SSL_CTX_set_options.pod
```diff
@@ -217,6 +217,16 @@ not propose, and servers will not accept the extension.
 Disable all renegotiation in TLSv1.2 and earlier. Do not send HelloRequest
 messages, and ignore renegotiation requests via ClientHello.
 
+=item SSL_OP_IGNORE_UNEXPECTED_EOF
+
+Some TLS implementations do not send close_notify on shutdown. Before OpenSSL
+3.0, SSL_ERROR_SYSCALL with errno=0 was reported in such cases. This
+behaviour contradicts the B<SSL_ERROR_SYSCALL> documentation. Since OpenSSL 3.0
+the returned error is B<SSL_ERROR_SSL> with a meaningful error on the error
+stack. This option gives the legacy error in such cases. You should only enable
```
t8m, line 226: As above.

doc/man3/SSL_CONF_cmd.pod
beldmit: Fixed. Thanks!

doc/man3/SSL_CONF_cmd.pod
beldmit: Fixed. Thanks!

doc/man3/SSL_CTX_set_options.pod
beldmit: Fixed. Thanks!

doc/man3/SSL_CONF_cmd.pod
beldmit: Fixed. Thanks!

doc/man3/SSL_CONF_cmd.pod
beldmit: Fixed. Thanks!

doc/man3/SSL_CTX_set_options.pod
beldmit: Fixed. Thanks!

CHANGES.md
```diff
@@ -30,6 +30,12 @@ OpenSSL 3.0
 
    *Shane Lontis*
 
+ * The SSL option SSL_OP_IGNORE_UNEXPECTED_EOF is introduced. If that option
+   is set, the unexpected EOF is ignored, it pretends close notify was received
+   instead and so returned error becomes SSL_ERROR_ZERO_RETURN.
```
kroeckx, line 35: I think it should be "an unexpected EOF", "a close notify", and "the returned error".

doc/man3/SSL_CONF_cmd.pod
```diff
@@ -85,6 +85,17 @@ its preference list. This usually indicates a client without AES hardware
 acceleration (e.g. mobile) is in use. Equivalent to B<SSL_OP_PRIORITIZE_CHACHA>.
 Only used by servers. Requires B<-serverpref>.
 
+=item B<-ignore_unexpected_eof>
+
+Some TLS implementations do not send close_notify on shutdown. Before OpenSSL
```
kroeckx, line 90: a close_notify alert

doc/man3/SSL_CONF_cmd.pod
```diff
@@ -85,6 +85,17 @@ its preference list. This usually indicates a client without AES hardware
...
+3.0, SSL_ERROR_SYSCALL with errno=0 was reported in such cases. This
+behaviour contradicts the B<SSL_ERROR_SYSCALL> documentation. Since OpenSSL 3.0
+the returned error is B<SSL_ERROR_SSL> with a meaningful error on the error
+stack. This option gives the B<SSL_ERROR_ZERO_RETURN> error in such cases.
```
kroeckx, line 94: I suggest you make that something like: When this option is enabled an B<SSL_ERROR_ZERO_RETURN> error is returned in case an EOF is detected and the close_notify alert was not received.

kroeckx:
This is still missing tests.

At first look, this might work. But I'm not convinced that it's the right function to set that the shutdown was received. I'm not familiar enough with the code for that. I think when it's in that function, you can have things like a frame is partially received, and then get an EOF, for which we still might want to give an error. On the other hand, the documentation does say that the application should detect any errors like that.

doc/man3/SSL_CTX_set_options.pod
```diff
@@ -217,6 +217,17 @@ not propose, and servers will not accept the extension.
...
+=item SSL_OP_IGNORE_UNEXPECTED_EOF
+
+Some TLS implementations do not send close_notify on shutdown. Before OpenSSL
+3.0, SSL_ERROR_SYSCALL with errno=0 was reported in such cases. This
+behaviour contradicts the B<SSL_ERROR_SYSCALL> documentation. Since OpenSSL 3.0
+the returned error is B<SSL_ERROR_SSL> with a meaningful error on the error
+stack. This option gives the B<SSL_ERROR_ZERO_RETURN> error in such cases.
+You should only enable this option if the protocol running over TLS can detect
+a truncation attack itself, and that the application is checking for that
+truncation attack.
```
kroeckx, line 229: If you have the same text at 2 location, please update them both

beldmit:
> This is still missing tests.

I've tested it manually (partially rolled back #11734 for this purpose). It works according to the current documentation.

doc/man3/SSL_CTX_set_options.pod
beldmit: Done. Somewhen this documentation should be united...

CHANGES.md
beldmit: Done. Thanks!

doc/man3/SSL_CONF_cmd.pod
beldmit: Done. Thanks!

doc/man3/SSL_CONF_cmd.pod
beldmit: Done. Thanks!

kroeckx:
I think the SSL_get_error() manpage should also be updated:
- SSL_ERROR_ZERO_RETURN should mention that you can also get it when the option is turned on
- It should probably document the old behaviour in a NOTES section.

beldmit:
Done.

mattcaswell:
> I've tested it manually (partially rolled back #11734 for this purpose). It works according to the current documentation.

An automated test should be straight forward. Take a look at sslapitest.c for plenty of examples of creating connections between a client and server. I'd suggest 2 iterations of the test. 1 without setting the option and without shutting down cleanly - confirm that you get the "unexpected eof" error. Then a second iteration where we do set the option and confirm that we get SSL_ERROR_ZERO_RETURN

doc/man3/SSL_get_error.pod
```diff
@@ -25,6 +25,13 @@ other OpenSSL function calls should appear in between.  The current
 thread's error queue must be empty before the TLS/SSL I/O operation is
 attempted, or SSL_get_error() will not work reliably.
 
+=head1 NOTES
+
+Some TLS implementations do not send a close_notify alert on shutdown. Before
+OpenSSL 3.0, B<SSL_ERROR_SYSCALL> with errno=0 was reported in such cases.
+Since OpenSSL 3.0 the returned error is B<SSL_ERROR_SSL> with a meaningful
+error on the error stack. See the B<SSL_ERROR_ZERO_RETURN> below.
```
kroeckx, line 33: It's not only in case the peer never sends a close_notify alert, but any unexpected EOF.

I would suggest that you either remove the last sentence, or talk about peers that never send it, and then just point them to directly to the SSL_OP_IGNORE_UNEXPECTED_EOF option.

doc/man3/SSL_CONF_cmd.pod
```diff
@@ -85,6 +85,18 @@ its preference list. This usually indicates a client without AES hardware
...
 
+=item B<-ignore_unexpected_eof>
+
+Some TLS implementations do not send a close_notify alert on shutdown. Before
+OpenSSL 3.0, B<SSL_ERROR_SYSCALL> with errno=0 was reported in such cases. This
+behaviour contradicts the B<SSL_ERROR_SYSCALL> documentation. Since OpenSSL 3.0
+the returned error is B<SSL_ERROR_SSL> with a meaningful error on the error
+stack. When this option is enabled an B<SSL_ERROR_ZERO_RETURN> error is
+returned in case an EOF is detected and the close_notify alert was not
+received. You should only enable this option if the protocol running over TLS
```
kroeckx, line 96: Here is a replacement suggestion:
Some TLS implementations do not send the mandatory close_notify alert on shutdown. If the application tries to wait for the close_notify alert but the peer closes the connection without sending it, an error is generated.  When this option is enabled the peer does not need to send the close_notify alert and a closed connection will be treated as if the close_notify alert was received. For more information on shutting down a connection, see L<SSL_shutdown>.

You should only [...]

SSL_shutdown() should probably mention the option.

kroeckx:
You made an option to be able to set this in a config file. I'm
not sure we want that. I think this option should only be set in
the program after actually confirming that it's ok to set the
option.

t8m:
> You made an option to be able to set this in a config file. I'm not sure we want that. I think this option should only be set in the program after actually confirming that it's ok to set the option.

On the other hand it would allow "workarounding" breakage in existing client/server apps that implement protocols that are known to require this op - such as https - without any code changes in the actual application. I am not saying it is absolutely needed however it has some pros. Also if we decide to remove this support, we will need to implement the option for s_client/s_server explicitly.

t8m:
> At first look, this might work. But I'm not convinced that it's the right function to set that the shutdown was received. I'm not familiar enough with the code for that. I think when it's in that function, you can have things like a frame is partially received, and then get an EOF, for which we still might want to give an error. On the other hand, the documentation does say that the application should detect any errors like that.

This is what the existing code in 1.1.1 already does (apart from returning the different error). I do not think we want to make this option to produce fatal error in some circumstances and zero return in others depending on the situation when the EOF exactly happens. It would only lead to confusion and further bug reports.

kroeckx:
> > You made an option to be able to set this in a config file. I'm not sure we want that. I think this option should only be set in the program after actually confirming that it's ok to set the option.
> 
> On the other hand it would allow "workarounding" breakage in existing client/server apps that implement protocols that are known to require this op - such as https - without any code changes in the actual application.

We are introducing the change in behaviour at an ABI/soname
change. So applications that have this problem should get this
around the time someone decided to recompile it, at which time
it should be easy to apply a software change too.

I'm concerned that people will modify the default openssl.cfg
file, because that's the only file where they can write such a
thing in most cases. And that will break the security of other
applications.

t8m:
> We are introducing the change in behaviour at an ABI/soname change. So applications that have this problem should get this around the time someone decided to recompile it, at which time it should be easy to apply a software change too. I'm concerned that people will modify the default openssl.cfg file, because that's the only file where they can write such a thing in most cases. And that will break the security of other applications.

Yes, that concern is definitely valid. Perhaps it could be partially mitigated by documentation? But I do not insist on having this as config file option so I am ok with dropping the config support. It should still be possible to enable/disable this via s_client/s_server option though.

doc/man3/SSL_get_error.pod
beldmit: Done.

doc/man3/SSL_CONF_cmd.pod
beldmit: Done.

beldmit:
I'd prefer to allow this option to be set via configuration file at least for test purposes.

beldmit:
Travis is happy.

BTW, the implemented behaviour differs from the old one and contradicts @mdounin (and Nginx as whole, AFAIK) request. Anyway, it allows to workaround when the peer does not send close_notify.

kroeckx:
> I'd prefer to allow this option to be set via configuration file at least for test purposes.

Can you explain what kind of test purposes?

beldmit:
> I'd prefer to allow this option to be set via configuration file at least for test purposes.
> Can you explain what kind of test purposes?

I'd like to provide a supplementary config to s_client.

Well, if this is the only showstopper, I can stand removing this option.

kroeckx:
> I'd like to provide a supplementary config to s_client.

I don't see how that is easier that just having an option for
s_client.

> Well, if this is the only showstopper, I can stand removing this option.

I will not approve it with that option, but that shouldn't prevent
someone else from approving it.

beldmit:
> Well, if this is the only showstopper, I can stand removing this option.
> I will not approve it with that option, but that shouldn't prevent someone else from approving it.

I'd prefer to finalize this PR so I removed this config option.

doc/man3/SSL_shutdown.pod
```diff
@@ -80,11 +80,12 @@ It should only be done when the peer has a way to make sure all
...
 message, otherwise an unexpected EOF will be reported.
 
-There are implementations that do not send the required close_notify alert.
-If there is a need to communicate with such an implementation, and it's clear
-that all data has been received, do not wait for the peer's close_notify alert.
-Waiting for the close_notify alert when the peer just closes the connection will
-result in an error being generated.
+There are implementations that do not send the required close_notify alert. If
+there is a need to communicate with such an implementation, and it's clear that
+all data has been received, do not wait for the peer's close_notify alert.
```
kroeckx, line 85: It's unclear to me why you reformat the above.

Note that the old one had the "." at the end of a line, and then the next sentence on the next line. pod2man recognized that "." at the end of a line. At least some people will format all the manpages like that, but I don't think we have a general style guide about it (yet).

doc/man3/SSL_shutdown.pod
```diff
@@ -80,11 +80,12 @@ It should only be done when the peer has a way to make sure all
...
+Waiting for the close_notify alert when the peer just closes the connection
+will result in an error being generated. See more details in
+L<SSL_CTX_set_options(3)>, the option B<SSL_OP_IGNORE_UNEXPECTED_EOF>.
```
kroeckx, line 88: I suggest: The error can be ignored using the B<SSL_OP_IGNORE_UNEXPECTED_EOF>. For more information see L<SSL_CTX_set_options(3)>.

doc/man3/SSL_shutdown.pod
beldmit: Removed the reformatting. Thanks!

doc/man3/SSL_shutdown.pod
beldmit: Done. Thanks!

test/sslapitest.c
```diff
@@ -6684,7 +6684,77 @@ static int test_ticket_callbacks(int tst)
...
+    char buf[80];
+    size_t written, readbytes;
+    SSL_SESSION *sess;
+    BIO *c2s;
+
+    if (!TEST_true(create_ssl_ctx_pair(libctx, TLS_server_method(),
+                                       TLS_client_method(),
+                                       TLS1_VERSION,
+                                       (tst <= 1) ? TLS1_2_VERSION
+                                                  : TLS1_3_VERSION,
```
kroeckx, line 6710: Just pass 0, 0 for the versions.

test/sslapitest.c
```diff
@@ -6684,7 +6684,77 @@ static int test_ticket_callbacks(int tst)
...
+            || !TEST_true(SSL_SESSION_is_resumable(sess)))
+        goto end;
+
+    c2s = SSL_get_rbio(serverssl);
+    BIO_set_mem_eof_return(c2s, 0);
+
+    if (!TEST_true(written = SSL_write(clientssl, msg, sizeof(msg))))
+        goto end;
+    if (!TEST_true(readbytes = SSL_read(serverssl, buf, sizeof(buf))))
+        goto end;
```
kroeckx, line 6735: I'm not sure what you want to do here, but you seem to set written and readbytes, then don't do anything with it.

test/sslapitest.c
```diff
@@ -6684,7 +6684,77 @@ static int test_ticket_callbacks(int tst)
...
+
+    SSL_set_shutdown(clientssl, SSL_SENT_SHUTDOWN | SSL_RECEIVED_SHUTDOWN);
```
kroeckx, line 6737: The SSL_CTX_set_quiet_shutdown() and SSL_set_shutdown() call don't actually seem relevant to the test.

doc/man3/SSL_CONF_cmd.pod
```diff
@@ -85,6 +85,15 @@ its preference list. This usually indicates a client without AES hardware
...
 
+=item B<-ignore_unexpected_eof>
+
+Some TLS implementations do not send the mandatory close_notify alert on
+shutdown. If the application tries to wait for the close_notify alert but the
+peer closes the connection without sending it, an error is generated. When this
+option is enabled the peer does not need to send the close_notify alert and a
+closed connection will be treated as if the close_notify alert was received.
+For more information on shutting down a connection, see L<SSL_shutdown(3)>.
+
```
kroeckx, line 96: I think this should get removed?

doc/man3/SSL_CTX_set_options.pod
```diff
@@ -217,6 +217,18 @@ not propose, and servers will not accept the extension.
...
+
+Some TLS implementations do not send a close_notify alert on shutdown. Before
+OpenSSL 3.0, B<SSL_ERROR_SYSCALL> with errno=0 was reported in such cases. This
+behaviour contradicts the B<SSL_ERROR_SYSCALL> documentation. Since OpenSSL 3.0
+the returned error is B<SSL_ERROR_SSL> with a meaningful error on the error
+stack. When this option is enabled an B<SSL_ERROR_ZERO_RETURN> error is
+returned in case an EOF is detected and the close_notify alert was not
+received. You should only enable this option if the protocol running over TLS
+can detect a truncation attack itself, and that the application is checking for
+that truncation attack.
```
kroeckx, line 230: This is missing the changes you did in ignore_unexpected_eof above

ssl/ssl_conf.c
```diff
@@ -656,6 +656,7 @@ static const ssl_conf_cmd_tbl ssl_conf_cmds[] = {
     SSL_CONF_CMD_SWITCH("no_renegotiation", 0),
     SSL_CONF_CMD_SWITCH("no_resumption_on_reneg", SSL_CONF_FLAG_SERVER),
     SSL_CONF_CMD_SWITCH("no_legacy_server_connect", SSL_CONF_FLAG_SERVER),
+    SSL_CONF_CMD_SWITCH("ignore_unexpected_eof", 0),
```
kroeckx, line 659: This, and the other change in this file, should get removed

doc/man3/SSL_get_error.pod
```diff
@@ -25,6 +25,15 @@ other OpenSSL function calls should appear in between.  The current
...
 attempted, or SSL_get_error() will not work reliably.
 
+=head1 NOTES
+
+Some TLS implementations do not send a close_notify alert on shutdown. In such
+cases and in other cases of unexpected EOF, before OpenSSL 3.0,
+B<SSL_ERROR_SYSCALL> with errno=0 was reported.
+
+Since OpenSSL 3.0 the returned error is B<SSL_ERROR_SSL> with a meaningful
+error on the error stack. See the B<SSL_ERROR_ZERO_RETURN> below.
```
kroeckx, line 35: I suggest:
On an unexpected EOF, versions before OpenSSL 3.0 returned B<SSL_ERROR_SYSCALL>, nothing was added to the error stack, and errno was 0. Since OpenSSL 3.0 the returned error is B<SSL_ERROR_SSL> with a meaningful
error on the error stack.

ssl/ssl_conf.c
beldmit: Removing these changes causes an error when the ignore_unexpected_eof is provided via command line.

kroeckx: To me that mean that it needs to be written in a different way.

test/sslapitest.c
beldmit: Done. Thanks!

test/sslapitest.c
beldmit: Done. Thanks!

apps/include/opt.h
```diff
@@ -185,6 +186,8 @@
             "Disallow session resumption on renegotiation"}, \
         {"no_legacy_server_connect", OPT_S_NOLEGACYCONN, '-', \
             "Disallow initial connection to servers that don't support RI"}, \
+        {"ignore_unexpected_eof", OPT_S_IGNORE_UNEXPECTED_EOF, '-', \
+            "Use legacy error handling if there is no close_notify"}, \
```
kroeckx, line 190: So this needs to be removed, and you need to write actual code in s_client/s_server and call SSL_(CTX_)set_options.

test/sslapitest.c
beldmit: Removed. Thanks!

doc/man3/SSL_get_error.pod
beldmit: Applied. Thanks!

doc/man3/SSL_CTX_set_options.pod
beldmit: Sorry, could you please specify the exact place to inherit the changes from?

kroeckx: The one in SSL_CONF_cmd.pod

doc/man3/SSL_CONF_cmd.pod
beldmit: Done. Thanks!

doc/man3/SSL_CTX_set_options.pod
beldmit: Done. Thanks!

ssl/ssl_conf.c
beldmit: Done. Thanks!

apps/include/opt.h
beldmit: Done. Thanks!

doc/man3/SSL_CTX_set_options.pod
```diff
@@ -217,6 +217,25 @@ not propose, and servers will not accept the extension.
...
+peer closes the connection without sending it, an error is generated. When this
+option is enabled the peer does not need to send the close_notify alert and a
+closed connection will be treated as if the close_notify alert was received.
+
+Before OpenSSL 3.0, B<SSL_ERROR_SYSCALL> with errno=0 was reported in such cases.
+This behaviour contradicts the B<SSL_ERROR_SYSCALL> documentation. Since OpenSSL 3.0
+the returned error is B<SSL_ERROR_SSL> with a meaningful error on the error
+stack. When this option is enabled an B<SSL_ERROR_ZERO_RETURN> error is
+returned in case an EOF is detected and the close_notify alert was not
+received. You should only enable this option if the protocol running over TLS
```
kroeckx, line 233: I think starting from "Before" until "received." can just be removed.

test/sslapitest.c
```diff
@@ -6684,7 +6684,63 @@ static int test_ticket_callbacks(int tst)
...
+        SSL_CTX_set_options(sctx, SSL_OP_IGNORE_UNEXPECTED_EOF);
+
+    if (!TEST_true(create_ssl_objects(sctx, cctx, &serverssl, &clientssl,
+                                            NULL, NULL)))
+        goto end;
+
+    if (!TEST_true(create_ssl_connection(serverssl, clientssl,
+                                              SSL_ERROR_NONE))
+            || !TEST_ptr_ne(sess = SSL_get_session(clientssl), NULL)
+            || !TEST_true(SSL_SESSION_is_resumable(sess)))
```
kroeckx, line 6719: The session things don't seem useful in this test

doc/man3/SSL_CTX_set_options.pod
beldmit: Done. Thanks!

test/sslapitest.c
beldmit: Removed. Thanks!

beldmit:
@kroeckx, all your change requests are processed and Travis will be happy soon

beldmit:
Travis is happy.

apps/s_server.c
```diff
@@ -851,6 +851,8 @@ const OPTIONS s_server_options[] = {
      "Disable caching and tickets if ephemeral (EC)DH is used"},
     {"www", OPT_WWW, '-', "Respond to a 'GET /' with a status page"},
     {"WWW", OPT_UPPER_WWW, '-', "Respond to a 'GET with the file ./path"},
+    {"ignore_unexpected_eof", OPT_IGNORE_UNEXPECTED_EOF, '-',
+     "Use legacy error handling if there is no close_notify"},
```
mattcaswell, line 855: This text doesn't seem quite right. The error handling isn't like it was in legacy

test/sslapitest.c
```diff
@@ -6684,7 +6684,60 @@ static int test_ticket_callbacks(int tst)
...
  *                  doesn't read it
+ * Test 7: TLSv1.3, client does not shutdown properly,
+ *                  server does not set SSL_OP_IGNORE_UNEXPECTED_EOF,
+ *                  server should get SSL_ERROR_SSL
+ * Test 8: TLSv1.3, client does not shutdown properly,
+ *                  server sets SSL_OP_IGNORE_UNEXPECTED_EOF,
+ *                  server should get SSL_ERROR_ZERO_RETURN
  */
+
+static int test_incorrect_shutdown(int tst)
```
mattcaswell, line 6695: Since the codepath is entirely different I would prefer this to be an independent top-level test instead of a sub-flow of `test_shutdown`

apps/s_server.c
t8m: And the same applies to the option in s_client.

test/sslapitest.c
beldmit: Done. Thanks!

apps/s_server.c
beldmit: Fixed both places. Thanks!

test/sslapitest.c
beldmit: Done. Thanks!

beldmit:
Travis is happy.

The incorrect shutdown tests are separated, the description of the -ignore_unexpected_eof option is adjusted.

apps/s_client.c
```diff
@@ -718,6 +718,8 @@ const OPTIONS s_client_options[] = {
      "Do not send the server name (SNI) extension in the ClientHello"},
     {"tlsextdebug", OPT_TLSEXTDEBUG, '-',
      "Hex dump of all TLS extensions received"},
+    {"ignore_unexpected_eof", OPT_IGNORE_UNEXPECTED_EOF, '-',
+     "Do not treat lack of close_notify from peer as error"},
```
t8m, line 722: I'm sorry for a nit: `from a peer as an error`

beldmit: Fixed. Thanks!

test/sslapitest.c
```diff
@@ -6672,6 +6672,61 @@ static int test_ticket_callbacks(int tst)
     return testresult;
 }
 
+/*
+ * Test incorrect shutdown.
+ * Test 0: TLSv1.3, client does not shutdown properly,
+ *                  server does not set SSL_OP_IGNORE_UNEXPECTED_EOF,
+ *                  server should get SSL_ERROR_SSL
+ * Test 1: TLSv1.3, client does not shutdown properly,
```
mattcaswell, line 6680: It says TLSv1.3 here - but actually you haven't specified any particular protocol version in the test below (the min/max protocol version parameters to `create_ssl_ctx_pair` are set to 0. So if TLSv1.3 is enabled we will use it. Otherwise we will use TLSv1.2 (or below of TLSv1.2  is also disabled). Should we test both TLSv1.2 *and* TLSv1.3? Or is one sufficient? Either way this comment isn't quite right.

beldmit: I think the version is irrelevant here. Updated the comment.

beldmit:
Merged. Many thanks to all involved!

PR #11736, opened 2020-05-05
 Generalize HTTP server code from apps/ocsp.c to apps/lib/http_server.c

DDvO:
This is a further spin-off PR of #11470 as requested by @mattcaswell: https://github.com/openssl/openssl/pull/11470#discussion_r419925336.
    
This adds `apps/lib/http_server.c` and `apps/include/http_server.h`
simplifying `apps/ocsp.c`. This going to be used also for `apps/cmp.c`
and will be helpful as well for any future app acting as an HTTP server.

This PR also contains a couple of related fixes as separate commits:

* Fix `bio_wait()` in `crypto/bio/bio_lib.c` in case `OPENSSL_NO_SOCK`
* Constify '`req`' parameter of `OSSL_HTTP_post_asn1()`
* Fix bug in `OSSL_CMP_SRV_process_request()` on transaction renewal

### DIFF
--- a/apps/include/http_server.h
+++ b/apps/include/http_server.h
@@ -1,0 +1,102 @@
+/*
+ * Copyright 1995-2020 The OpenSSL Project Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License 2.0 (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
+ */
+
+#ifndef OSSL_HTTP_SERVER_H
+# define OSSL_HTTP_SERVER_H
+
+# include "apps.h"
+
+# ifndef HAVE_FORK
+#  if defined(OPENSSL_SYS_VMS) || defined(OPENSSL_SYS_WINDOWS)
+#   define HAVE_FORK 0
+#  else
+#   define HAVE_FORK 1
+#  endif
+# endif
+
+# if HAVE_FORK
+#  undef NO_FORK
+# else
+#  define NO_FORK
+# endif
+
+# if !defined(NO_FORK) && !defined(OPENSSL_NO_SOCK) \
+    && !defined(OPENSSL_NO_POSIX_IO)
+#  define HTTP_DAEMON
+#  include <sys/types.h>
+#  include <sys/wait.h>
+#  include <syslog.h>
+#  include <signal.h>
+#  define MAXERRLEN 1000 /* limit error text sent to syslog to 1000 bytes */
+# else
+#  undef LOG_INFO
+#  undef LOG_WARNING
+#  undef LOG_ERR
+#  define LOG_INFO      0
+#  define LOG_WARNING   1
+#  define LOG_ERR       2
+# endif
+
+/*-
+ * Log a message to syslog if multi-threaded HTTP_DAEMON, else to bio_err
+ * prog: the name of the current app
+ * level: the severity of the message, e.g., LOG_ERR
+ * fmt: message with potential extra parameters like with printf()
+ * returns nothing
+ */
+void log_message(const char *prog, int level, const char *fmt, ...);
+
+# ifndef OPENSSL_NO_SOCK
+/*-
+ * Initialize an HTTP server by setting up its listening BIO
+ * prog: the name of the current app
+ * port: the port to listen on
+ * returns a BIO for accepting requests, NULL on error
+ */
+BIO *http_server_init_bio(const char *prog, const char *port);
+/*-
+ * Accept an ASN.1-formatted HTTP request
+ * it: the expected request ASN.1 type
+ * preq: pointer to variable where to place the parsed request
+ * pcbio: pointer to variable where to place the BIO for sending the response to
+ * acbio: the listening bio (typically as returned by http_server_init_bio())
+ * prog: the name of the current app
+ * accept_get: wheter to accept GET requests (in addition to POST requests)
+ * timeout: connection timeout (in seconds), or 0 for none/infinite
+ * returns 0 in case caller should retry, then *preq == *pcbio == NULL
+ * returns -1 on fatal error; also in this case *preq == *pcbio == NULL
+ * returns 1 otherwise. In this case it is guaranteed that *pcbio != NULL
+ *                      while *preq == NULL if and only if request is invalid
+ */
+int http_server_get_asn1_req(const ASN1_ITEM *it, ASN1_VALUE **preq,
+                        BIO **pcbio, BIO *acbio,
+                        const char *prog, int accept_get, int timeout);
+/*-
+ * Send an ASN.1-formatted HTTP response
+ * cbio: destination BIO (typically as returned by http_server_get_asn1_req())
+ *       note: cbio should not do an encoding that changes the output length
+ * content_type: string identifying the type of the response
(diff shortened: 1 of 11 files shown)


apps/ocsp.c
```diff
@@ -1336,209 +1144,32 @@ static char **lookup_serial(CA_DB *db, ASN1_INTEGER *ser)
...
-
-    *preq = req;
-
-out:
-# ifdef OCSP_DAEMON
-    if (timeout > 0)
-        alarm(0);
-    acfd = (int)INVALID_SOCKET;
-# endif
     return 1;
```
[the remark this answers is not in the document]
DDvO: Opos, this was a glitch. Fixed.

crypto/cmp/cmp_server.c
```diff
@@ -489,9 +489,9 @@ OSSL_CMP_MSG *OSSL_CMP_SRV_process_request(OSSL_CMP_SRV_CTX *srv_ctx,
                           tid);
             OPENSSL_free(tid);
         }
-        /* start of a new transaction, set transactionID and senderNonce */
-        if (!OSSL_CMP_CTX_set1_transactionID(ctx, hdr->transactionID)
-                || !ossl_cmp_ctx_set1_recipNonce(ctx, hdr->senderNonce))
+        /* start of a new transaction, reset transactionID and senderNonce */
+        if (!OSSL_CMP_CTX_set1_transactionID(ctx, NULL)
+                || !OSSL_CMP_CTX_set1_senderNonce(ctx, NULL))
```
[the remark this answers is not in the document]
DDvO: This change is part of the commit with the message 
```
Fix bug in OSSL_CMP_SRV_process_request() on transaction renewal
```
I found the issue while testing the new HTTP server code with CMP.
So its server-related but only relevant for CMP.

I thought it's not worth setting up an extra PR for it.
Or shall I move this commit to #11470?

apps/lib/http_server.c
```diff
@@ -0,0 +1,399 @@
...
+    return acbio;
+
+ err:
+    BIO_free_all(acbio);
+    BIO_free(bufbio);
+    return NULL;
+}
+
+/*
+ * Decode %xx URL-decoding in-place. Ignores mal-formed sequences.
```
vdukhovni, line 233: s/mal-formed/malformed/

apps/lib/http_server.c
```diff
@@ -0,0 +1,399 @@
...
+    return ret;
+}
+
+int http_server_send_resp(BIO *cbio, const char *content_type,
+                          const ASN1_ITEM *it, const ASN1_VALUE *val)
+{
+    int ret = BIO_printf(cbio, "HTTP/1.0 200 OK\r\nContent-type: %s\r\n"
+                         "Content-Length: %d\r\n\r\n", content_type,
+                         ASN1_item_i2d(val, NULL, it)) > 0
+            && ASN1_item_i2d_bio(it, cbio, val) > 0;
```
vdukhovni, line 394: There's an assumption here that `cbio` is not performing a non-trivial encoding that changes the output length.  That should perhaps be noted in comments above the function.

apps/lib/http_server.c
```diff
@@ -0,0 +1,399 @@
...
+    if (timeout > 0)
+        alarm(0);
+    acfd = (int)INVALID_SOCKET;
+# endif
+    return ret;
+}
+
+int http_server_send_resp(BIO *cbio, const char *content_type,
+                          const ASN1_ITEM *it, const ASN1_VALUE *val)
+{
```
vdukhovni, line 390: This function should perhaps have "asn1" somewhere in its name. One might imagine other HTTP server interfaces that send something other than ASN1 data.

apps/include/http_server.h
```diff
@@ -0,0 +1,65 @@
...
+
+void log_message(const char *prog, int level, const char *fmt, ...);
+
+# ifndef OPENSSL_NO_SOCK
+BIO *http_server_init(const char *prog, const char *port);
+int http_server_get_req(const ASN1_ITEM *it, ASN1_VALUE **preq,
+                        BIO **pcbio, BIO *acbio,
+                        const char *prog, int accept_get, int timeout);
+int http_server_send_resp(BIO *cbio, const char *content_type,
+                          const ASN1_ITEM *it, const ASN1_VALUE *val);
```
vdukhovni, line 54: Though these are internal interfaces, they deserve some internal documentation, in the form of block comments with interface docs, describing semantics, input requirements, ...  Now that they're not just internal functions in `ocsp.c`

apps/lib/http_server.c
DDvO: done

apps/lib/http_server.c
DDvO: I've extended the name of this function accordingly, and similarly for two other functions.

apps/lib/http_server.c
DDvO: I've added a respective comment/note to the implementation and in the header file.

apps/include/http_server.h
DDvO: I've placed such documenting comments in apps/include/http_server.h

DDvO:
Thanks @mattcaswell and @vdukhovni for your reviews and comments.
I've addressed all of them.

apps/include/http_server.h
```diff
@@ -0,0 +1,102 @@
...
+ * prog: the name of the current app
+ * port: the port to listen on
+ * returns a BIO for accepting requests, NULL on error
+ */
+BIO *http_server_init_bio(const char *prog, const char *port);
+/*-
+ * Accept an ASN.1-formatted HTTP request
+ * it: the expected request ASN.1 type
+ * preq: pointer to variable where to place the parsed request
+ * pcbio: pointer to variable where to place to BIO for sending the response to
```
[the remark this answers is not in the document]
DDvO: fixed

DDvO:
> Approved assuming the minor nit is fixed.

Thanks for your swift reaction!
I've fixed the doc nit and partly squashed the commit structure for merging tomorrow.

vdukhovni:
It would have been cool to leverage a ubiquitous light-weight HTTP server library in C, with a suitable license instead of rolling our own, but the only that comes to mind (libmicrohttpd) is LGPL.  Mind you, we only use the HTTP server in "apps", it wouldn't be quite so bad to have it be runtime linked to an LGPL library, but we'd still need to something on systems where that's not available.  Anyone know of suitable alternatives?

vdukhovni:
> It would have been cool to leverage a ubiquitous light-weight HTTP server library in C, with a suitable license instead of rolling our own, but the only one that comes to mind (libmicrohttpd) is LGPL. Mind you, we only use the HTTP server in "apps", it wouldn't be quite so bad to have it be runtime linked to an LGPL library, but we'd still need to something on systems where that's not available. Anyone know of suitable alternatives?

Anyone have any experience good/bad with [libhttp](https://www.libhttp.org/)? The license is MIT, so not a problem...

DDvO:
Merged - thanks @mattcaswell and @vdukhovni!

DDvO:
> It would have been cool to leverage a ubiquitous light-weight HTTP server library in C, with a suitable license instead of rolling our own

I agree this would make sense at least instead of trying to further complete this very rudimentary HTTP server implementation, which would be a mayor endeavor.
The same I'd say holds for the HTTP client merged in #10667, which is somewhat more advanced but also far from being complete and already took quite some effort.

Issue #11694, opened 2020-04-30
Segmentation fault compiling openssl-3.0.0-alpha1 on RHEL 6.10


satya-kotla:
I am compiling openssl-3.0.0-alpha1 on RHEL 6.10 server and getting Segmentation fault below. Can some one suggest a solution for this issue.

./config --prefix=/app/digital/openssl shared
Operating system: x86_64-whatever-linux2
Configuring OpenSSL version 3.0.0-alpha1 for target linux-x86_64
Using os-specific seed configuration
./config: line 910: 19681 Segmentation fault      (core dumped) /usr/bin/env __CNF_CPPDEFINES='' __CNF_CPPINCLUDES='' __CNF_CPPFLAGS='' __CNF_CFLAGS='' __CNF_CXXFLAGS='' __CNF_LDFLAGS='' __CNF_LDLIBS='' /usr/bin/perl ./Configure linux-x86_64 '--prefix=/app/digital/openssl' 'shared'

mattcaswell:
What perl version are you running?

satya-kotla:
perl version is 5.10.1

mattcaswell:
I don't have access to RHEL 6.10 but this seems to be caused by a bug in older versions of perl. I am seeing a very similar crash in an older perl on a different platform, so it might be the same thing or might not.

Please could you try out the following workaround procedure and see if it works for you?

From a *clean* source tree please edit the file `apps/build.info` to remove these lines:

````
  DEPEND[${OPENSSLSRC/.c/.o}]=progs.h
  GENERATE[progs.c]=progs.pl -C $(APPS_OPENSSL)
  GENERATE[progs.h]=progs.pl -H $(APPS_OPENSSL)
  # progs.pl tries to read all 'openssl' sources, including progs.c, so we make
  # sure things are generated in the correct order.
  DEPEND[progs.h]=progs.c
````
Then run `config`/`Configure` as you normally would.

Before running make and while standing in the top of the source tree, execute the following:

````
touch apps/progs.c
perl apps/progs.pl -C apps/openssl >apps/progs.c
perl apps/progs.pl -H apps/openssl >apps/progs.h
````
Now run `make` and `make test` as normal.

Please let me know if this works.

levitte:
Interesting workaround.
@mattcaswell, have you tried `perl -d Configure ... ` ? I'm not sure if it catches a core dump, but if it does, it would be nice to see exactly where that happens...

mattcaswell:
[...]

PR #11737, opened 2020-05-05
Configure: avoid perl regexp bugs

levitte:
It seems that in older perl versions '(?P' doesn't interact very well
with '(?|' or '(?:'.

Since we make extensive use of '(?P' in build.info parsing, we avoid
combining that with '(?|' and '(?:' when parsing build.info variables,
and end up parsing variable modifier twice (first generally, and then
parse that result into the modifier components).

All perl versions up to and including 5.12.5 are affected.


mspncp:
Are there any references to the OpenSSL issue reports and/or the Perl bug?

mattcaswell:
> Are there any references to the OpenSSL issue reports and/or the Perl bug?

See #11694

levitte:
> See #11694

Thanks for the reminder!

t8m:
There was only commit message change in the force-push.

levitte:
Merged

4975e8b4d2cfab923d522840533334a1bbd754b8 Configure: avoid perl regexp bugs

PR #11739, opened 2020-05-05
Enable markdownlint

richsalz:
testing.


t8m:
Just a comment - I'd probably add it to some singleton build, not to the build matrix.

richsalz:
>  I'd probably add it to some singleton build, not to the build matrix.

Clues or pointers?

.travis.yml
```diff
@@ -32,7 +32,7 @@ env:
     # Note: env entry here must exactly match the value in the exclude: table below that contains env:, otherwise it will not find a match.
     - CONFIG_OPTS="" DESTDIR="_install"
     - CONFIG_OPTS="no-asm -Werror --debug no-afalgeng no-shared enable-rc5 enable-md2 -fsanitize=address" LSAN_OPTIONS="report_objects=1"
-    - CONFIG_OPTS="no-asm no-makedepend enable-buildtest-c++ --strict-warnings -D_DEFAULT_SOURCE" BUILDONLY="yes" CHECKDOCS="yes" CPPFLAGS="-ansi"
+    - CONFIG_OPTS="no-asm no-makedepend enable-buildtest-c++ --strict-warnings -D_DEFAULT_SOURCE" BUILDONLY="yes" CHECKDOCS="yes" MARKDOWNLINT="yes" CPPFLAGS="-ansi"
```
t8m, line 35: Instead of adding MARKDOWNLINT here, I would add it on line 138 - that is the "disable all that's possible" build which should be already quite fast so running something else in there should not be a problem

richsalz: thanks, good idea.

ACKNOWLEDGEMENTS.md
```diff
@@ -1,5 +1,4 @@
-Acknowlegements
-===============
+# Acknowlegements #
```
t8m, line 1: Is the change from `===` to `##` mandatory for the lint to pass? Can it be excepted? The reason the underlining is used is that the .md file is supposed to be well readable in the pure textual form.

CHANGES.md
```diff
@@ -5976,15 +5971,15 @@ OpenSSL 1.0.1
    If the BIT STRING containing the signature has non zero unused bits reject
    the signature. All current signature algorithms require zero unused bits.
 
-   2. Check certificate algorithm consistency.
+   1. Check certificate algorithm consistency.
```
t8m, line 5974: This looks like a wrong fix. The numbering should be kept, probably the text needs to be reformatted to keep all the text as part of the list item.

CHANGES.md
```diff
@@ -6419,12 +6414,12 @@ OpenSSL 1.0.1
 
    1. Do not use record version number > TLS 1.0 in initial client
       hello: some (but not all) hanging servers will now work.
-   2. If we set OPENSSL_MAX_TLS1_2_CIPHER_LENGTH this will truncate
+   1. If we set OPENSSL_MAX_TLS1_2_CIPHER_LENGTH this will truncate
```
t8m, line 6417: Same as above.

CHANGES.md
```diff
@@ -12466,7 +12457,7 @@ s-cbc           3624.96k     5258.21k     5530.91k     5624.30k     5628.26k
    *Steve Henson*
 
  * New subcommands for 'openssl ca':
-   'openssl ca -status <serial>' prints the status of the cert with
+   'openssl ca -status &lt;serial>' prints the status of the cert with
```
t8m, line 12460: :disappointed: This also worsens readability of the text as a pure text.

t8m:
> So I see three issues with the changes to conform to the markdownlint that worsen the readability as pure text:
> 
>     * `#` instead of underlines - ugly
> 

Here this can be solved by using `rule 'MD003', :style => :setext_with_atx` and changing the lower level headers to conform to the style.

>     * numbered lists renumbered to 1 - that is a major thing

I would simply disable the rule MD029.
 
>     * `&lt;` - not nice either

This can be solved by using code block instead of `&lt;` to avoid violating bare HTML rule.

richsalz:
I will do all of those, thanks for the review and suggested solutions!

richsalz:
All done, thanks.
(Now hopefully this will not take as long as other doc cleanup PR's)

.travis.yml
```diff
@@ -204,6 +204,12 @@ script:
           echo -e '\052\052 FAILED -- MAKE DOC-NITS';
           travis_terminate 1;
       fi
+    - if test -n "$MARKDOWNLINT" ; then
+          echo -e "====START MARKDOWNLINT====";
+          gem install mdl || travis_terminate 1;
+          mdl -s util/style.rb README.md || travis_terminate 1;
```
t8m, line 210: You run the lint only on README.md here.

util/style.rb
```diff
@@ -0,0 +1,17 @@
...
+
+exclude_rule 'MD004' # Unordered list style TODO(fix?)
+exclude_rule 'MD005' # Inconsistent indentation for list items at the same level
+exclude_rule 'MD006' # Consider starting bulleted lists at the beginning of the line
+exclude_rule 'MD012' # Multiple consecutive blank lines
+exclude_rule 'MD014' # Dollar signs used before commands without showing output
+exclude_rule 'MD024' # Multiple headers with the same content
+exclude_rule 'MD025' # Multiple top level headers in the same document
+exclude_rule 'MD029' # Ordered list item prefix
+exclude_rule 'MD036'  # Emphasis used instead of a header
```
t8m, line 17: Nit: extra space before comment

t8m:
Do you prefer if the commits are merged as they are or are you OK with squashing them during the merge?

richsalz:
I am happy to have them merged into one.  It was useful during development to show each set of changes so that, e.g., you could make good suggestions :)  I will fix the travis file.  I have one more cleanup commit coming.

mspncp:
> Do you prefer if the commits are merged as they are or are you OK with squashing them during the merge?

I would prefer to see two commits: one which enables markdownlint, and one which fixes the markdown files.

.travis.yml
richsalz: fixed.

util/style.rb
richsalz: Tht space is there so you can easily add a comma if you add more rule exlusions,
while still keeping the comment-character lined up :)

CHANGES.md
```diff
@@ -981,14 +978,12 @@ OpenSSL 3.0
 
    *Boris Pismenny*
 
-
 OpenSSL 1.1.1
 -------------
 
-### Changes between 1.1.1e and 1.1.1f [xx XXX xxxx] ###
+### Changes between 1.1.1e and 1.1.1f [xx XXX xxxx]
```
mspncp, line 984: @mattcaswell does this change affect the regex of the release script?

CHANGES.md
```diff
@@ -22,7 +21,7 @@ OpenSSL Releases
 OpenSSL 3.0
 -----------
 
-### Changes between 1.1.1 and 3.0 [xx XXX xxxx] ###
+### Changes between 1.1.1 and 3.0 [xx XXX xxxx]
```
mspncp, line 24: @mattcaswell does this change affect the regex of the release script?

CHANGES.md
```diff
@@ -4365,10 +4358,10 @@ OpenSSL 1.0.2
...
 
- * Fix memory issues in BIO_*printf functions
+ * Fix memory issues in `BIO_*printf` functions
 
    The internal |fmtstr| function used in processing a "%s" format string in
-   the BIO_*printf functions could overflow while calculating the length of a
+   the `BIO_*printf` functions could overflow while calculating the length of a
    string and cause an OOB read when printing very long strings.
 
    Additionally the internal |doapr_outch| function can attempt to write to an
```
mspncp, line 4367: There are a lot of places where pipes are used as delimiters for code code, like ` |doapr_outch|`.  Mind replacing them with backticks? For example, this change note contains a lot of those:

```
 * Fix BN_hex2bn/BN_dec2bn NULL pointer deref/heap corruption

   In the BN_hex2bn function the number of hex digits is calculated using an
   int value |i|. Later |bn_expand| is called with a value of |i * 4|. For
   large values of |i| this can result in |bn_expand| not allocating any
   memory because |i * 4| is negative. This can leave the internal BIGNUM data
   field as NULL leading to a subsequent NULL ptr deref. For very large values
   of |i|, the calculation |i * 4| could be a positive value smaller than |i|.
   In this case memory is allocated to the internal BIGNUM data field, but it
   is insufficiently sized leading to heap corruption. A similar issue exists
   in BN_dec2bn. This could have security consequences if BN_hex2bn/BN_dec2bn
   is ever called by user applications with very large untrusted hex/dec data.
   This is anticipated to be a rare occurrence.
```

CHANGES.md
```diff
@@ -9401,7 +9377,7 @@ OpenSSL 0.9.x
 
    *Bodo Moeller; problem pointed out by Yutaka Oiwa (Research Center
    for Information Security, National Institute of Advanced Industrial
-   Science and Technology [AIST*, Japan)]
+   Science and Technology [AIST, Japan)]*
```
mspncp, line 9380: Unbalanced parentheses; I would suggest changing

    Science and Technology [AIST, Japan)]*

to

    Science and Technology [AIST], Japan)*

CHANGES.md
```diff
@@ -13483,7 +13446,7 @@ ndif
    instead.  BIO_gethostbyname() does not know what timeouts are
    appropriate, so entries would stay in cache even when they have
    become invalid.
-   *Bodo Moeller; problem pointed out by Rich Salz <rsalz@zolera.com>
+   *Bodo Moeller; problem pointed out by Rich Salz <rsalz@zolera.com>*
```
mspncp, line 13449: @richsalz you worked for a car rental company?

CHANGES.md
```diff
@@ -14044,7 +14007,7 @@ ndif
    *Bodo Moeller; problem reported by Eric Day <eday@concentric.net>*
 
  * In RSA_eay_public_{en,ed}crypt and RSA_eay_mod_exp (rsa_eay.c),
```
mspncp, line 14009: Backquotes `` `RSA_eay_public_{en,ed}crypt` ``

CHANGES.md
```diff
@@ -14544,8 +14506,8 @@ ndif
 
    *Steve Henson*
 
- * crypto/err.c locking bugfix: Use write locks (CRYPTO_w_[un]lock),
-   not read locks (CRYPTO_r_[un]lock).
+ * crypto/err.c locking bugfix: Use write locks (`CRYPTO_w_[un]lock)`,
```
mspncp, line 14509: Last tick is misplaced: `` (`CRYPTO_w_[un]lock)` ``  ->  `` (`CRYPTO_w_[un]lock`) ``

CHANGES.md
```diff
@@ -15107,7 +15069,7 @@ ndif
 
  * Change names of new functions to the new get1/get0 naming
    convention: After 'get1', the caller owns a reference count
```
mspncp, line 15071: Backticks around `` `get1` ``?

mspncp: (same for `` `get0` `` and `` `get` `` below)

CHANGES.md
```diff
@@ -16864,7 +16822,7 @@ ndif
  * Create a duplicate of the SSL_CTX's CERT in SSL_new instead of
    copying pointers.  The cert_st handling is changed by this in
    various ways (and thus what used to be known as ctx->default_cert
-   is now called ctx->cert, since we don't resort to s->ctx->[default_]cert
+   is now called ctx->cert, since we don't resort to `s->ctx->[default_]cert`
```
mspncp, line 16825: Why did you backtick only the second code expression and not the first? Is there some system behind it? (E.g. only those code fragments which  contain wildcard-like characters such as `*[]{}`?)

CHANGES.md
```diff
@@ -16873,7 +16831,7 @@ ndif
 
    Note that using the SSL API in certain dirty ways now will result
    in different behaviour than observed with earlier library versions:
-   Changing settings for an SSL_CTX *ctx after having done s = SSL_new(ctx)
+   Changing settings for an `SSL_CTX *ctx` after having done s = SSL_new(ctx)
```
mspncp, line 16834: Same here: why only the first, not the second?

t8m:
I would not insist on Rich to fix all the possible issues in the changelogs. :) It could be a separate PR. However the point about release scripts is important.

t8m:
The fixup-CHANGES.md-postrelease.pl and fixup-NEWS.md-postrelease.pl need to be adjusted according to the new style.

CHANGES.md
richsalz: I was going to look at that next.

richsalz:
Thanks, @t8m!  Additional commit pushed.

CHANGES.md
richsalz: Yes, exactly. I did just what was necessary for vim-syntax to not think there was [emphasis] or other markdown markup. :)  It was easy, paging through the file looking for misplaced highlight blocks :)

CHANGES.md
richsalz: another PR to fix the pipe and quote "markup" is a good idea.  maybe someone will do that.  maybe i will :)  i would like my other doc update PR's to get merged first.

CHANGES.md
richsalz: I restored the numbering.

CHANGES.md
richsalz: > you worked for a car rental company?

No, it was a startup that tried to provide a digital signature signing device.  It failed. :)

CHANGES.md
richsalz: > Science and Technology [AIST], Japan)*
fixed.

CHANGES.md
richsalz: > Last tick is misplaced: (`CRYPTO_w_[un]lock)` -> (`CRYPTO_w_[un]lock`)
fixed.

CHANGES.md
richsalz: > Backquotes `` `RSA_eay_public_{en,ed}crypt` ``

fixed.

richsalz:
I created #11749 to capture the still-to-do-work.

As far as merging,smooshing,re-ordering the commits, I leave it to whoever does the merge; I don't care. :)

t8m:
If you take care of the merge, then you can just do it when merging.

mspncp:
Merged to master, thanks!

257e9d03b0 Fix issues reported by markdownlint
4ef0ddc9d8 travis: enable markdownlint checks

PR #11484, opened 2020-04-07
Fix PEM certificate loading that sometimes failes

blenk92:
As described in https://github.com/openssl/openssl/issues/9187 the loading of PEM certificates sometimes fails if the line base64 content is in one line and the length of the line is a multiple of 254.
The problem is in get_header_and_data(). When the last line has a length of 254 (or a multiple) the next read will only read a newline. Due to this get_header_and_data() expects to be in the header not in the data area. This commit fixes this by introducing the new state MAYBE_POST_HEADER. This state ensures that the loaded content is treated as data, if such a newline is the last thing that could be read from the string. If there is additional content after the content already read will be treated as data.

CLA: trivial

Fixes #9187

I'm not sure about the CLA thingy, please tell me if you need it ;-)

##### Checklist

- [x] tests are added or updated

Issue #9187, opened 2019-06-18
Unable to load certificate - sometimes


pgporada:
Hi,

I am experiencing this on Fedora 29, Fedora 30, and Ubuntu 19.04 which all use OpenSSL `1.1.1`. CentOS7 does not exhibit this issue, but it has version `1.0.2`. Debian 9.9 does not exhibit this either, but it has version `1.1.0`
```
$ openssl version
OpenSSL 1.1.1c FIPS  28 May 2019

$ uname -a
Linux work 4.19.50-1.pvops.qubes.x86_64 #1 SMP Sat Jun 15 22:10:40 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

The issue I'm encountering is that openssl is unable to parse a specific certificate.
```
$ openssl x509 -in cert.127.pem -noout
unable to load certificate

$ cat cert.127.pem 
-----BEGIN CERTIFICATE-----
[...]

PR #11741, opened 2020-05-05
Fix PEM certificate loading that sometimes fails

blenk92:
As described in https://github.com/openssl/openssl/issues/9187, the
loading of PEM certificates sometimes fails if a line of base64
content has the length of a multiple of 254.
The problem is in get_header_and_data(). When such a line with a
length of 254 (or a multiple) has been read, the next read will
only read a newline. Due to this get_header_and_data() expects to be
in the header not in the data area. This commit fixes that by checking
if lines have been read completely or only partially. In case of a
previous partial read, a newline will be ignored.

Created a new PR as adviced during submitting (I|C)CLA
Old PR: https://github.com/openssl/openssl/pull/11484

##### Checklist

- [x ] tests are added or updated


crypto/pem/pem_lib.c
```diff
@@ -828,6 +828,12 @@ static int get_header_and_data(BIO *bp, BIO **header, BIO **data, char *name,
             goto err;
         }
 
+        /* Check if line has been read completely or if only part of the line
+         * has been read. Keep the previous value to ignore newlines that
+         * appear due to reading a line up until the char before the newline */
```
t8m, line 833: Style nit - the end of multiline comment should be on a separate line with the `*` aligned.

crypto/pem/pem_lib.c
```diff
@@ -838,13 +844,17 @@ static int get_header_and_data(BIO *bp, BIO **header, BIO **data, char *name,
 
         /* Check for end of header. */
         if (linebuf[0] == '\n') {
-            if (got_header == POST_HEADER) {
-                /* Another blank line is an error. */
-                PEMerr(PEM_F_GET_HEADER_AND_DATA, PEM_R_BAD_END_LINE);
-                goto err;
+            /* If previous line has been read only partially this newline is a
+             * regular newline at the end of a line and not an empty line */
```
t8m, line 848: As above.

crypto/pem/pem_lib.c
blenk92: okay, done

crypto/pem/pem_lib.c
blenk92: done, as well ;-)

crypto/pem/pem_lib.c
kaduk: Shouldn't the opening `/*` also be on a line of its own (for a multiline comment)?

crypto/pem/pem_lib.c
```diff
@@ -828,6 +828,13 @@ static int get_header_and_data(BIO *bp, BIO **header, BIO **data, char *name,
             goto err;
         }
 
+        /* Check if line has been read completely or if only part of the line
+         * has been read. Keep the previous value to ignore newlines that
+         * appear due to reading a line up until the char before the newline
```
kaduk, line 833: If you make other changes (e.g., to the formatting of the start of the comment), please also add a period to end this sentence.

blenk92:
> Thanks again for putting this together with the extra test cases -- it looks good (modulo the comment style for which @t8m and I need to check what is right)

okay, thanks for the fast reviews :-)

let me know if something needs to be changed...

kaduk:
Chapter 8 at https://www.openssl.org/policies/codingstyle.html says that the opening `/*` should be on its own line.  I don't think that these comments have particular formatting, so the `-` in `/*-` is not needed.  @t8m please confirm

blenk92:
> Chapter 8 at https://www.openssl.org/policies/codingstyle.html says that the opening `/*` should be on its own line. I don't think that these comments have particular formatting, so the `-` in `/*-` is not needed. @t8m please confirm

Alright then, i guess if it's written there it makes sense to change it...

kaduk:
Thanks; I reconfirm it looks good to me.

t8m:
Yeah, I forgot about that rule. Still approved after the comment style change.

kaduk:
It's been more than 24 hours since @t8m reconfirmed, so I'll go ahead and start merging this one.  I will keep @beldmit as a reviewer given the trivial nature of the changes since his approval

kaduk:
Merged manually to master and 1.1.1, so closing.
Thank you again for the fix!!

PR #11744, opened 2020-05-06
Read MIDR_EL1 system register on aarch64

zorrorffm:
MIDR_EL1 system register exposes microarchitecture information so that people can make micro-arch related optimization such as exposing as much instruction level parallelism as possible.

MIDR_EL1 register can be read only if HWCAP_CPUID feature is supported.

We are going to optimize some algorithms (chacha20, AES-GCM etc.) based on this change which can get performance uplifted significantly. 

##### Checklist

### DIFF
--- a/crypto/armcap.c
+++ b/crypto/armcap.c
@@ -18,6 +18,7 @@
 #include "arm_arch.h"
 
 unsigned int OPENSSL_armcap_P = 0;
+unsigned int OPENSSL_arm_midr = 0;
 
 #if __ARM_MAX_ARCH__<7
 void OPENSSL_cpuid_setup(void)
@@ -48,6 +49,7 @@ void _armv8_sha256_probe(void);
 void _armv8_pmull_probe(void);
 # ifdef __aarch64__
 void _armv8_sha512_probe(void);
+unsigned int _armv8_cpuid_probe(void);
 # endif
 uint32_t _armv7_tick(void);
 
@@ -95,6 +97,7 @@ void OPENSSL_cpuid_setup(void) __attribute__ ((constructor));
 #  define HWCAP_CE_PMULL         (1 << 4)
 #  define HWCAP_CE_SHA1          (1 << 5)
 #  define HWCAP_CE_SHA256        (1 << 6)
+#  define HWCAP_CPUID            (1 << 11)
 #  define HWCAP_CE_SHA512        (1 << 21)
 # endif
 
@@ -155,6 +158,9 @@ void OPENSSL_cpuid_setup(void)
 #  ifdef __aarch64__
         if (hwcap & HWCAP_CE_SHA512)
             OPENSSL_armcap_P |= ARMV8_SHA512;
+
+        if (hwcap & HWCAP_CPUID)
+            OPENSSL_armcap_P |= ARMV8_CPUID;
 #  endif
     }
 # endif
@@ -210,5 +216,10 @@ void OPENSSL_cpuid_setup(void)
 
     sigaction(SIGILL, &ill_oact, NULL);
     sigprocmask(SIG_SETMASK, &oset, NULL);
+
+# ifdef __aarch64__
+    if (OPENSSL_armcap_P & ARMV8_CPUID)
+        OPENSSL_arm_midr = _armv8_cpuid_probe();
+# endif
 }
 #endif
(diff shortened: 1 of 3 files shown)


zorrorffm:
Does anyone take a look at this CL as well as the related one #12397? Thanks

paulidale:
Ping for review.

crypto/armcap.c
```diff
@@ -198,6 +204,10 @@ void OPENSSL_cpuid_setup(void)
             _armv8_sha512_probe();
             OPENSSL_armcap_P |= ARMV8_SHA512;
         }
+        if (sigsetjmp(ill_jmp, 1) == 0) {
+            _armv8_cpuid_probe();
+            OPENSSL_armcap_P |= ARMV8_CPUID;
+        }
 #  endif
```
ardbiesheuvel, line 211: I think we can drop this hunk: this code is not expected to execute at an exception level where MIDR_EL1 is accessible, and so the only way we can read it is through the CPUID emulation, which is a Linux specific feature, and which is advertised via the CPUID hwcap.

paulidale: I did wonder if it was necessary but figured it wasn't harmful.

ardbiesheuvel: Actually, it could be harmful, and it would be better to remove this code entirely at some point.

The problem is that the OS may decide to hide certain capabilities from user space, either because of silicon errata if there is a problem with the feature, or in the case of heterogeneous SMP (big.LITTLE), where a task may migrate from a CPU that has support for certain instructions to one that does not.

In this particular case, I think it is harmless, since the only possible way to access MIDR_EL1 is via the emulation, which will always produce the value the OS thinks is the correct one. But in general, we should abide by the hwcap features only, and avoid SIGILL trapping for feature discovery.

kroeckx: If the OS supports reading hardware capabilities, that is the only information we should use. But it's not available on all platforms, so we need fallback code.

The OS might not always provide enough information for us to make a good decision about which implementation to use. While a CPU might support 2 implementations we have, newer generation of CPUs might preform better using the other that worked best for the older generation. There can also be differences depending on the range of CPU: mobile vs. server. And as far as I know, we don't really either have code nor benchmark results for it to pick the best implementation.

There is 1 PR making use of this PR, which has a performance again of about 3%. I think we've previously considered 10% performance increase the minimum we should see to consider such changes.

ardbiesheuvel: In a sense, those are even worse: if sha256 is supported by the big CPUs, but not by the little CPUs, and the program happens to be started on big one, you might experience SIGILL crashes if the program gets migrated to a little CPU by the OS scheduler.

paulidale (approved):
Still approved.  Needs second review.

ardbiesheuvel:
Not sure if my review counts or not, but I am fine with the patch as is. *However*, I am still not convinced we need this patch at all, given that the patch that relies on it is still under discussion, and is unlikely to be merged in its current form given the limited performance gain. So please don't merge this yet.

zorrorffm:
> 
> 
> Not sure if my review counts or not, but I am fine with the patch as is. _However_, I am still not convinced we need this patch at all, given that the patch that relies on it is still under discussion, and is unlikely to be merged in its current form given the limited performance gain. So please don't merge this yet.

I agree that we can't merge a PR with limited performance uplift.  What I want to point out is that the original optimization on chacha20  is dedicated to ThunderX2, the author commented that his implementation is not friendly to most platforms but ThunderX2 because he thought ThunderX2 is server oriented, other platforms are client oriented. I believed that the author compromised because there wasn't a approach to detect the platform at run time. So actually this PR provide the possibility making chacha20 running at best performance on various platforms. Moreover we can see the performance data from his implementation, A53 gets 9% performance regression, this is to say, based on this PR, #12397 can revert the performance on A53. 

Another point I would like to mention is that this PR will benefit to  micro architecture  related optimization which maybe bring significant performance gain. Actually original chacha20 optimization on ThunderX2 is such an example, this optimization dedicated to ThunderX2 makes the performance 37% uplift.   Also for further ARM micro architecture like Zeus we will have more opportunities to optimize crypto algorithms based on this PR.  So I thinks this PR is valuable.

zorrorffm:
Just ping...
This PR provides the opportunity to balance the performance on all platform.  Based on this PR,  #12397 makes all the performance on different platforms(A53, A55, A57, A72) uplifted while no performance regression on ThunderX2 occurs.

Thanks.

paulidale:
ping @openssl/committers for review?

t8m (approved):

t8m:
Merged to master. Thank you for the contribution.

PR #7390, opened 2018-10-12
WIP: apps: Switch to using OSSL_STORE for loading keys, certs, ...

levitte:
This makes all the format specifiers largely irrelevant, except for
the ENGINE format when a key is given as a key ID only.  However, with
the engine loader, it's possible to change this:

    -engine {engineid} -key {keyid} -keyformat ENGINE

To this:

    -key engine://{engineid}/{keyid}

We retain the ENGINE format for the sake of compatibility.  All other
input formats are simply ignored.  Note that output formats is a
different story.

Fixes #1963

PR #11745, opened 2020-05-06
Use OSSL_STORE for load_key() and load_pubkey() in apps/lib/apps.c

DDvO:
This is a further spin-off of PR #11470 according to https://github.com/openssl/openssl/pull/11470#discussion_r416669564 by @mattcaswell.

This also adds
STORE_load_key_cert_crl(), STORE_load_pkey(), 
STORE_load_cert(), STORE_load_crl(),
get_passwd(), cleanse(), and clear_free()
to be used also in `apps/cmp.c` etc.

### DIFF
--- a/apps/lib/apps.c
+++ b/apps/lib/apps.c
@@ -29,6 +29,7 @@
 #include <openssl/x509.h>
 #include <openssl/x509v3.h>
 #include <openssl/pem.h>
+#include <openssl/store.h>
 #include <openssl/pkcs12.h>
 #include <openssl/ui.h>
 #include <openssl/safestack.h>
@@ -208,6 +209,24 @@ int wrap_password_callback(char *buf, int bufsiz, int verify, void *userdata)
 
 
 static char *app_get_pass(const char *arg, int keepbio);
+
+char *get_passwd(const char *pass, const char *desc)
+{
+    char *result = NULL;
+
+    if (desc == NULL)
+        desc = "<unknown>";
+    if (!app_passwd(pass, NULL, &result, NULL))
+        BIO_printf(bio_err, "Error getting password for %s\n", desc);
+    if (pass != NULL && result == NULL) {
+        BIO_printf(bio_err,
+                   "Trying plain input string (better precede with 'pass:')\n");
+        result = OPENSSL_strdup(pass);
+        if (result == NULL)
+        BIO_printf(bio_err, "Out of memory getting password for %s\n", desc);
+    }
+    return result;
+}
 
 int app_passwd(const char *arg1, const char *arg2, char **pass1, char **pass2)
 {
@@ -419,16 +438,13 @@ static int load_pkcs12(BIO *in, const char *desc,
     const char *pass;
     char tpass[PEM_BUFSIZE];
     int len, ret = 0;
+    const char *error = NULL;
     PKCS12 *p12;
 
     p12 = d2i_PKCS12_bio(in, NULL);
-    if (p12 == NULL) {
-        if (desc != NULL)
-            BIO_printf(bio_err, "Error loading PKCS12 file for %s\n", desc);
-        else
-            BIO_printf(bio_err, "Error loading PKCS12 file\n");
-        goto die;
-    }
+    error = "Error loading PKCS12 file";
+    if (p12 == NULL)
+        goto end;
     /* See if an empty password will do */
     if (PKCS12_verify_mac(p12, "", 0) || PKCS12_verify_mac(p12, NULL, 0)) {
         pass = "";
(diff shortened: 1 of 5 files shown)


DDvO:
I'm going to generalize `load_cert()` in the same manner later today...

apps/lib/apps.c
```diff
@@ -551,71 +572,96 @@ X509_REQ *load_csr(const char *file, int format, const char *desc)
...
+        if (!maybe_stdin) {
+            BIO_printf(bio_err, "No filename or uri specified for loading %s\n",
+                       desc);
+            goto end;
+        }
         unbuffer(stdin);
-        key = dup_bio_in(format);
-    } else {
-        key = bio_open_default(file, 'r', format);
+        bio = dup_bio_in(format);
```
levitte, line 635: It's possible to use the STORE API with stdin.  The function is calledd `OSSL_STORE_attach` and is to be used in place of `OSSL_STORE_open` when you want it to handle a BIO using some chosen scheme.  The OSSL_STORE functionality knows how to handle streamed data, it uses the cache buffer to catch the first 4KiB and uses that to find out (guess) the format.

apps/lib/apps.c
```diff
@@ -629,103 +675,119 @@ EVP_PKEY *load_key(const char *file, int format, int maybe_stdin,
...
+    if (uri == NULL){
+        if (!maybe_stdin) {
+            BIO_printf(bio_err, "No filename or uri specified for loading %s\n",
+                       desc);
+        }
         unbuffer(stdin);
-        key = dup_bio_in(format);
-    } else {
-        key = bio_open_default(file, 'r', format);
+        bio = dup_bio_in(format);
```
levitte, line 721: Again, `OSSL_STORE_attach`.  That will probably also simplify your code, as you really shouldn't need to care about the format at all, except for ENGINE.

apps/lib/apps.c
```diff
@@ -847,6 +906,111 @@ int load_crls(const char *file, STACK_OF(X509_CRL) **crls, int format,
     return load_certs_crls(file, format, pass, desc, NULL, crls);
 }
 
+/*
+ * Load those types of credentials for which the result pointer is not NULL.
+ * For each type the first credential found in the store is loaded.
+ * May yield partial result even if rv == 0.
+ */
+int STORE_load_key_cert_crl(const char *uri, const char *pass, const char *desc,
+                            EVP_PKEY **ppkey, X509 **pcert, X509_CRL **pcrl)
```
levitte, line 915: I don't see why you're making duplicates of the already existing `load_` functions.  They aren't precious in an absolute stability manner, libapps is an *internal* library.

apps/lib/apps.c
DDvO: Hmm, would be nice, but `OSSL_STORE_attach()` does not exist, just undocumented internal functions `ossl_store_attach_pem_bio()` and `ossl_store_file_attach_pem_bio_int()`.

How if I export something like
```
OSSL_STORE_attach(bio, ui_method, ui_data) = 
ossl_store_attach_pem_bio(bio, ui_method, ui_data, NULL, NULL)
```
?

apps/lib/apps.c
DDvO: Yes, and for the exotic legacy formats FORMAT_MSBLOB and FORMAT_PVK.

levitte: Why?

apps/lib/apps.c
DDvO: Good point, so I'll merge them, and tend to add/"export" just 
```
load_key_cert_crl(const char *uri, const char *pass, const char *desc,
                            EVP_PKEY **ppkey, X509 **pcert, X509_CRL **pcrl)
```
for its flexibility loading multiple credentials from the same PEM or PKCS#12 file.

apps/lib/apps.c
levitte: Ah, right, it *could* be that it's because I looked at my own effort to do the same thing, where that is indeed already done, #7390

I should probably split that one out a bit.

apps/lib/apps.c
DDvO: Oops, these two legacy formats FORMAT_MSBLOB and FORMAT_PVK are relevant for load_key, while for load_pubkey it is only FORMAT_MSBLOB.
Or does OSSL_STORE support these as well?

mattcaswell:
Errr.....why did openssl-machine close this and delete the branch????

mattcaswell:
Ah! @DDvO - did you accidentally create this branch in the main github openssl repo rather than in your private fork? It seems that's where it was and got deleted when openssl-machine mirrored the main repo.

DDvO:
Argh, indeed I accidentally had pushed this PR to the wrong remote :-(
Unfortunately I cannot change the repo reference in this PR, so I need to open a new PR, right?

This is the new (meanwhile extended) commit: https://github.com/siemens/openssl/commit/8f636f170d10c66b55ca355dfb8f4944b64f3fd5

mattcaswell:
> Unfortunately I cannot change the repo reference in this PR, so I need to open a new PR, right?

Unfortunately, I think so.

DDvO:
So I've resurrected this as #11755, which already includes further improvements.

apps/lib/apps.c
DDvO: @levitte, I've found a much simpler solution than using `OSSL_STORE_attach()` (which also would require including part of #7390) - see the tiny commit 8b1080a in #11755.

apps/lib/apps.c
DDvO: I've verified that the legacy format are not supported by OSSL_STORE,
so in #11755 I kept their support in `apps/lib/apps.c`.

apps/lib/apps.c
DDvO: This is also done in #11755.

PR #11750, opened 2020-05-06
EVP: when setting the operation to EVP_PKEY_OP_UNDEFINED, clean up!

levitte:
There were a few instances where we set the EVP_PKEY_CTX operation to
EVP_PKEY_OP_UNDEFINED, but forgot to clean up first.  After the
operation is made undefined, there's no way to know what should be
cleaned away, so that must be done first, in all spots.


levitte:
Merged

c7fa92979c5964966efa298bf2a40ff451ee7482 EVP: when setting the operation to EVP_PKEY_OP_UNDEFINED, clean up!

PR #11752, opened 2020-05-07
Add 'on demand self test' and status test to providers

slontis:
The default and legacy providers currently return 1 for status and self test checks.
Added test to show the 3 different stages the self test can be run (for installation, loading and on demand).

For the fips provider:
 - If the on demand self test fails, then any subsequent fetches should also fail. To implement this the
   cached algorithms are flushed on failure.
 - getting the self test callback in the fips provider is a bit complicated since the callback hangs off the core
   libctx (as it is set by the application) not the actual fips library context. Also the callback can be set at
   any time not just during the OSSL_provider_init() so it is calculated each time before doing any self test.

##### Checklist

### DIFF
--- a/crypto/property/property.c
+++ b/crypto/property/property.c
@@ -394,10 +394,19 @@ fin:
     return ret;
 }
 
-static void impl_cache_flush_alg(ossl_uintmax_t idx, ALGORITHM *alg)
+static void impl_cache_flush_alg(ossl_uintmax_t idx, ALGORITHM *alg, void *arg)
 {
+    SPARSE_ARRAY_OF(ALGORITHM) *algs = arg;
+
     lh_QUERY_doall(alg->cache, &impl_cache_free);
-    lh_QUERY_flush(alg->cache);
+    if (algs != NULL) {
+        sk_IMPLEMENTATION_pop_free(alg->impls, &impl_free);
+        lh_QUERY_free(alg->cache);
+        OPENSSL_free(alg);
+        ossl_sa_ALGORITHM_set(algs, idx, NULL);
+    } else {
+        lh_QUERY_flush(alg->cache);
+    }
 }
 
 static void ossl_method_cache_flush(OSSL_METHOD_STORE *store, int nid)
@@ -406,14 +415,16 @@ static void ossl_method_cache_flush(OSSL_METHOD_STORE *store, int nid)
 
     if (alg != NULL) {
         store->nelem -= lh_QUERY_num_items(alg->cache);
-        impl_cache_flush_alg(0, alg);
+        impl_cache_flush_alg(0, alg, NULL);
     }
 }
 
-void ossl_method_store_flush_cache(OSSL_METHOD_STORE *store)
+void ossl_method_store_flush_cache(OSSL_METHOD_STORE *store, int all)
 {
+    void *arg = (all != 0 ? store->algs : NULL);
+
     ossl_property_write_lock(store);
-    ossl_sa_ALGORITHM_doall(store->algs, &impl_cache_flush_alg);
+    ossl_sa_ALGORITHM_doall_arg(store->algs, &impl_cache_flush_alg, arg);
     store->nelem = 0;
     ossl_property_unlock(store);
 }
--- a/crypto/provider.c
+++ b/crypto/provider.c
@@ -10,6 +10,7 @@
 #include <openssl/err.h>
 #include <openssl/cryptoerr.h>
 #include <openssl/provider.h>
+#include <openssl/core_names.h>
 #include "internal/provider.h"
 
 OSSL_PROVIDER *OSSL_PROVIDER_load(OPENSSL_CTX *libctx, const char *name)
@@ -67,6 +68,11 @@ const OSSL_ALGORITHM *OSSL_PROVIDER_query_operation(const OSSL_PROVIDER *prov,
 void *OSSL_PROVIDER_get0_provider_ctx(const OSSL_PROVIDER *prov)
 {
     return ossl_provider_prov_ctx(prov);
+}
+
+int OSSL_PROVIDER_self_test(const OSSL_PROVIDER *prov)
+{
+    return ossl_provider_self_test(prov);
 }
 
 int OSSL_PROVIDER_get_capabilities(const OSSL_PROVIDER *prov,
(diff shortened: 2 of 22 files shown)


crypto/property/property.c
```diff
@@ -440,10 +449,14 @@ int ossl_method_store_merge_global_properties(OSSL_METHOD_STORE *store,
...
 
-static void impl_cache_flush_alg(ossl_uintmax_t idx, ALGORITHM *alg)
+static void impl_cache_flush_alg(ossl_uintmax_t idx, ALGORITHM *alg, void *arg)
 {
+    SPARSE_ARRAY_OF(ALGORITHM) *algs = arg;
+
     lh_QUERY_doall(alg->cache, &impl_cache_free);
     lh_QUERY_flush(alg->cache);
+    if (algs != NULL)
+        ossl_sa_ALGORITHM_set(algs, idx, NULL);
```
paulidale, line 459: Does this leak the hash table?

`lh_QUERY_free(alg->cache)` instead of `lh_QUERY_flush(alg->cache)` ??

slontis: So you mean:
if (algs != NULL) {
    lh_QUERY_free(alg->cache);
    ossl_sa_ALGORITHM_set(algs, idx, NULL);
} else {
    lh_QUERY_flush(alg->cache);
}

paulidale: Yeah.

slontis: It needed to do a bit more to cleanup than this as this was the only pointer to the algorithm.

slontis:
rebased.

slontis:
ping

slontis:
Rebase again.
ping

providers/fips/fipsprov.c
```diff
@@ -235,7 +252,16 @@ static int fips_get_params(void *provctx, OSSL_PARAM params[])
...
     if (p != NULL && !OSSL_PARAM_set_utf8_ptr(p, OPENSSL_FULL_VERSION_STR))
         return 0;
-
+    p = OSSL_PARAM_locate(params, OSSL_PROV_PARAM_STATUS);
+    if (p != NULL && !OSSL_PARAM_set_uint(p, FIPS_is_running()))
+        return 0;
+    p = OSSL_PARAM_locate(params, OSSL_PROV_PARAM_SELF_TEST);
+    if (p != NULL) {
+        set_self_test_cb(FIPS_get_core_handle(selftest_params.libctx));
+        status = SELF_TEST_post(&selftest_params, 1) ? 1 : 0;
```
mattcaswell, line 261: I find it quite strange that running the self test is a side-effect of "getting" parameters. I would have expected a separate dedicated function to trigger this.

crypto/provider.c
```diff
@@ -54,7 +56,19 @@ const OSSL_PARAM *OSSL_PROVIDER_gettable_params(const OSSL_PROVIDER *prov)
...
 int OSSL_PROVIDER_get_params(const OSSL_PROVIDER *prov, OSSL_PARAM params[])
 {
-    return ossl_provider_get_params(prov, params);
+    OSSL_PARAM *p;
+    int ret;
+    unsigned int status = 0;
+
+    ret = ossl_provider_get_params(prov, params);
+
+    p = OSSL_PARAM_locate(params, OSSL_PROV_PARAM_SELF_TEST);
```
mattcaswell, line 65: This special purpose code is quite ugly here.

crypto/provider.c
```diff
@@ -54,7 +56,19 @@ const OSSL_PARAM *OSSL_PROVIDER_gettable_params(const OSSL_PROVIDER *prov)
...
+    if (p != NULL) {
+        OSSL_PARAM_get_uint(p, &status);
+        if (status == 0)
+            evp_method_store_flush(ossl_provider_library_context(prov));
```
mattcaswell, line 69: Is this flushing good enough? It doesn't actually prevent the FIPS module being used if algorithms have been pre-fetched - which is a very reasonable strategy and is something libssl does. In the libssl case I could imagine that you could get a self test failure and libssl would carry on quite happily still using the FIPS module.

providers/fips/fipsprov.c
```diff
@@ -753,23 +774,24 @@ int fips_intern_provider_init(const OSSL_CORE_HANDLE *handle,
...
-    if (c_get_libctx == NULL)
+    if (c_internal_get_libctx == NULL)
         return 0;
 
     if ((*provctx = PROV_CTX_new()) == NULL)
         return 0;
+
     /*
      * Using the parent library context only works because we are a built-in
      * internal provider. This is not something that most providers would be
```
mattcaswell, line 797: Aren't we supposed to be calling `c_internal_get_libctx` below?

crypto/provider.c
slontis: This was discussed in the initial design meetings, Not sure there is much we can do if they are prefetched..
The old FOm had a test (or at least tried to and failed) in every operation to see. Not sure that we want to head down that route do we?

providers/fips/fipsprov.c
slontis: Err maybe - this has probably been like this for quite some time..

providers/fips/fipsprov.c
slontis: Ok changed.

providers/fips/fipsprov.c
mattcaswell: No - it was correct before your changes. Your changes renamed `c_get_libctx` to `c_internal_get_libctx` everywhere except in the section below.

mattcaswell: No - it was correct before your changes. Your changes renamed c_get_libctx to c_internal_get_libctx everywhere except in the section below.

crypto/provider.c
slontis: @paulidale - care to comment?

slontis:
Rebased again.

crypto/provider.c
slontis: @mattcaswell - I think this is outside the scope of this PR. A separate PR will be needed to address this.

mattcaswell: Ok

slontis:
rebased to fix generated file.

slontis:
Thanks. Merged to master.

PR #11754, opened 2020-05-07
Warn that objects with opaque types must never be passed on.

levitte:
If libcrypto passes an opaque object pointer to a function of a
provider that's linked with a different libcrypto, there's no saying
how that will work, as an opaque object may change at any time, even
between PATCH level.  Therefore, such objects must only be used as
arguments to upcalls, which are guaranteed to exist in the same
libcrypto that the objects came from.

### DIFF
--- a/doc/man7/provider.pod
+++ b/doc/man7/provider.pod
@@ -52,6 +52,8 @@ the provider.
 For the provider itself, it may hold some interesting information,
 and is also passed to some of the functions given in the dispatch
 array I<in>.
+If the provider is linked with OpenSSL libraries, objects like this
+must be treated carefully, see L</WARNINGS> below.
 
 I<in> is a dispatch array of base functions offered by the OpenSSL
 libraries, and the available functions are further described in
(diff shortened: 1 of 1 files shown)


levitte:
This came up in a recent discussion with a colleague, where we discussed possible [ahem] deep hacks.

Part of the design criteria is that a provider module may be linked with a different libcrypto version than the one that calls it, thereby attempting to guarantee both forward and backward compatibility.
However, it seems that even we who work with OpenSSL development every day forget this criterion, so better put it in writing.

mattcaswell:
I'd suggest holding back on this for the moment. I have a half-written PR that should address many of the concerns. Once I've got that done we can assess whether this warning is still needed or not.

levitte:
I frankly don't understand how anything will be able to change that warning.  If libcrypto (built from 3.1.0, say) passes an opaque pointer to a provider linked with a different libcrypto (built from 3.0.3, say), it would be *dangerous* if that provider chooses to pass that pointer to a function in its libcrypto (the 3.0.3 one).

levitte:
However, fine, there's no rush  for this PR

doc/man7/provider.pod
```diff
@@ -351,6 +353,21 @@ other providers:
...
+=head1 WARNINGS
+
+For a dynamically loadable provider module, any object with an opaque
+type passed from the OpenSSL libraries to any provider funtion I<must
+only> be used as arguments to OpenSSL library upcalls, i.e. the
+functions available through the I<in> dispatch table passed to the
+initialization functions described in L</Provider> above.  Any other
+use of such objects is forbidden.
+
+A provider that's built-in to the application or OpenSSL's F<libcrypto>
```
paulidale, line 365: `s.that's.that is.`

mattcaswell:
> Are there any instances of passing a libcrypto opaque type to a provider? The params APIs do not support this.

Yes. There are several.

> I'd prefer that this be banned than having restrictions placed upon it.

I'm working on it.

levitte:
> Are there any instances of passing a libcrypto opaque type to a provider? The params APIs do not support this.

The serializers currently call things like `i2d_PKCS8_bio`...  I plan to change that, though.

> I'd prefer that this be banned than having restrictions placed upon it.

I have no idea what you mean with that...

mattcaswell:
See #11758

levitte:
Does everyone think that this PR is now unnecessary?

mattcaswell:
Once #11758 is in, it doesn't seem necessary.

levitte:
Okie.  Since #11758 is well on its way, I see no reason for keeping this around.  Kill

PR #11758, opened 2020-05-07
Maintain strict type discipline between the core and providers

mattcaswell:
A provider could be linked against a different version of libcrypto than
the version of libcrypto that loaded the provider. Different versions of
libcrypto could define opaque types differently. It must never occur that
a type created in one libcrypto is used directly by the other libcrypto.
This will cause crashes.

We can "cheat" for "built-in" providers that are part of libcrypto itself,
because we know that the two libcrypto versions are the same - but not for
other providers.

To ensure this does not occur we use different types names for the handful
of opaque types that are passed between the core and providers.

### DIFF
--- a/doc/man7/provider-base.pod
+++ b/doc/man7/provider-base.pod
@@ -16,15 +16,15 @@ provider-base
   */
 
  /* Functions offered by libcrypto to the providers */
- const OSSL_ITEM *core_gettable_params(const OSSL_PROVIDER *prov);
+ const OSSL_ITEM *core_gettable_params(const OSSL_CORE_PROVIDER *prov);
  int core_get_params(const OSSL_PROVIDER *prov, OSSL_PARAM params[]);
- int core_thread_start(const OSSL_PROVIDER *prov,
+ int core_thread_start(const OSSL_CORE_PROVIDER *prov,
                        OSSL_thread_stop_handler_fn handfn);
- OPENSSL_CTX *core_get_library_context(const OSSL_PROVIDER *prov);
- void core_new_error(const OSSL_PROVIDER *prov);
- void core_set_error_debug(const OSSL_PROVIDER *prov,
+ OPENSSL_CORE_CTX *core_get_library_context(const OSSL_CORE_PROVIDER *prov);
+ void core_new_error(const OSSL_CORE_PROVIDER *prov);
+ void core_set_error_debug(const OSSL_CORE_PROVIDER *prov,
                            const char *file, int line, const char *func);
- void core_vset_error(const OSSL_PROVIDER *prov,
+ void core_vset_error(const OSSL_CORE_PROVIDER *prov,
                       uint32_t reason, const char *fmt, va_list args);
 
  /*
@@ -52,7 +52,19 @@ provider-base
                                const char *file, int line);
  int CRYPTO_secure_allocated(const void *ptr);
  void OPENSSL_cleanse(void *ptr, size_t len);
- unsigned char *OPENSSL_hexstr2buf(const char *str, long *len);
+
+ OSSL_CORE_BIO * BIO_new_file(const char *filename, const char *mode)
+ OSSL_CORE_BIO * BIO_new_membuf(const void *buf, int len)
+ int BIO_read_ex(OSSL_CORE_BIO *bio, void *data, size_t data_len,
+                 size_t *bytes_read))
+ int BIO_write_ex(OSSL_CORE_BIO *bio, const void *data, size_t data_len,
+                  size_t *written)
+ int BIO_free(OSSL_CORE_BIO *bio))
+ int BIO_vprintf(OSSL_CORE_BIO *bio, const char *format, va_list args)
+ int BIO_vsnprintf(char *buf, size_t n, const char *fmt, va_list args)
+
+ void self_test_cb(OPENSSL_CORE_CTX *ctx, OSSL_CALLBACK **cb, void **cbarg)
+
 
  /* Functions offered by the provider to libcrypto */
  void provider_teardown(void *provctx);
@@ -114,7 +126,6 @@ provider):
  BIO_free                       OSSL_FUNC_BIO_FREE
  BIO_vprintf                    OSSL_FUNC_BIO_VPRINTF
  OPENSSL_cleanse                OSSL_FUNC_OPENSSL_CLEANSE
- OPENSSL_hexstr2buf             OSSL_FUNC_OPENSSL_HEXSTR2BUF
  OSSL_SELF_TEST_set_callback    OSSL_FUNC_SELF_TEST_CB
 
 For I<*out> (the B<OSSL_DISPATCH> array passed from the provider to
@@ -188,7 +199,10 @@ BIO_new_file(), BIO_new_mem_buf(), BIO_read_ex(), BIO_free(),
 BIO_vprintf(), OPENSSL_cleanse(), and OPENSSL_hexstr2buf()
 correspond exactly to the public functions with the same name.
 As a matter of fact, the pointers in the B<OSSL_DISPATCH> array are
-direct pointers to those public functions.
+direct pointers to those public functions. Note that the BIO functions take an
+B<OSSL_CORE_BIO> type rather than the standard B<BIO> type. This is to ensure
+that a provider does not mix BIOs from the core with BIOs used on the provider
+side (the two are not compatible).
 OSSL_SELF_TEST_set_callback() is used to set an optional callback that can be
 passed into a provider. This may be ignored by a provider.
 
--- a/doc/man7/provider-base.pod
+++ b/doc/man7/provider-base.pod
@@ -88,7 +88,7 @@ function pointer from a B<OSSL_DISPATCH> element named
 B<OSSL_get_{name}>.
 For example, the "function" core_gettable_params() has these:
 
- typedef OSSL_ITEM *
+ typedef OSSL_PARAM *
      (OSSL_core_gettable_params_fn)(const OSSL_CORE_HANDLE *handle);
  static ossl_inline OSSL_NAME_core_gettable_params_fn
      OSSL_get_core_gettable_params(const OSSL_DISPATCH *opf);
@@ -148,14 +148,14 @@ parameters.
 
 =for comment core_thread_start() TBA
 
-core_get_library_context() retrieves the library context in which the
-B<OSSL_CORE_HANDLE> object I<handle> is stored.
+core_get_library_context() retrieves the library context in which the library
+object for the current provider is stored, accessible through the I<handle>.
 This may sometimes be useful if the provider wishes to store a
 reference to its context in the same library context.
 
 core_new_error(), core_set_error_debug() and core_set_error() are
 building blocks for reporting an error back to the core, with
-reference to the handle object I<handle>.
+reference to the I<handle>.
 
 =over 4
 
--- a/doc/man7/provider-base.pod
+++ b/doc/man7/provider-base.pod
@@ -16,15 +16,15 @@ provider-base
   */
 
  /* Functions offered by libcrypto to the providers */
- const OSSL_ITEM *core_gettable_params(const OSSL_CORE_PROVIDER *prov);
- int core_get_params(const OSSL_PROVIDER *prov, OSSL_PARAM params[]);
- int core_thread_start(const OSSL_CORE_PROVIDER *prov,
+ const OSSL_ITEM *core_gettable_params(const OSSL_CORE_HANDLE *handle);
+ int core_get_params(const OSSL_CORE_HANDLE *handle, OSSL_PARAM params[]);
+ int core_thread_start(const OSSL_CORE_HANDLE *handle,
                        OSSL_thread_stop_handler_fn handfn);
- OPENSSL_CORE_CTX *core_get_library_context(const OSSL_CORE_PROVIDER *prov);
- void core_new_error(const OSSL_CORE_PROVIDER *prov);
- void core_set_error_debug(const OSSL_CORE_PROVIDER *prov,
+ OPENSSL_CORE_CTX *core_get_library_context(const OSSL_CORE_HANDLE *handle);
+ void core_new_error(const OSSL_CORE_HANDLE *handle);
+ void core_set_error_debug(const OSSL_CORE_HANDLE *handle,
                            const char *file, int line, const char *func);
- void core_vset_error(const OSSL_CORE_PROVIDER *prov,
+ void core_vset_error(const OSSL_CORE_HANDLE *handle,
                       uint32_t reason, const char *fmt, va_list args);
 
  /*
@@ -89,7 +89,7 @@ B<OSSL_get_{name}>.
 For example, the "function" core_gettable_params() has these:
 
  typedef OSSL_ITEM *
-     (OSSL_core_gettable_params_fn)(const OSSL_PROVIDER *prov);
+     (OSSL_core_gettable_params_fn)(const OSSL_CORE_HANDLE *handle);
  static ossl_inline OSSL_NAME_core_gettable_params_fn
      OSSL_get_core_gettable_params(const OSSL_DISPATCH *opf);
 
@@ -142,20 +142,20 @@ F<libcrypto>):
 core_gettable_params() returns a constant array of descriptor
 B<OSSL_PARAM>, for parameters that core_get_params() can handle.
 
-core_get_params() retrieves I<prov> parameters from the core.
+core_get_params() retrieves parameters from the core for the given I<handle>.
 See L</Core parameters> below for a description of currently known
 parameters.
 
 =for comment core_thread_start() TBA
 
 core_get_library_context() retrieves the library context in which the
-B<OSSL_PROVIDER> object I<prov> is stored.
+B<OSSL_CORE_HANDLE> object I<handle> is stored.
 This may sometimes be useful if the provider wishes to store a
 reference to its context in the same library context.
 
 core_new_error(), core_set_error_debug() and core_set_error() are
 building blocks for reporting an error back to the core, with
-reference to the provider object I<prov>.
+reference to the handle object I<handle>.
 
 =over 4
 
@@ -341,7 +341,7 @@ operation C<BAR>.
 
  /* Provider context */
  struct prov_ctx_st {
-     OSSL_PROVIDER *prov;
+     OSSL_CORE_HANDLE *handle;
  }
 
  /* operation context for the algorithm FOO */
@@ -357,7 +357,7 @@ operation C<BAR>.
      if (fooctx != NULL)
          fooctx->provctx = provctx;
      else
-         c_put_error(provctx->prov, E_MALLOC, __FILE__, __LINE__);
+         c_put_error(provctx->handle, E_MALLOC, __FILE__, __LINE__);
      return fooctx;
  }
 
@@ -433,7 +433,7 @@ operation C<BAR>.
      { 0, NULL }
  };
 
- int OSSL_provider_init(const OSSL_PROVIDER *provider,
+ int OSSL_provider_init(const OSSL_CORE_HANDLE *handle,
                         const OSSL_DISPATCH *in,
                         const OSSL_DISPATCH **out,
                         void **provctx)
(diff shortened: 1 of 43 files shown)


levitte:
> We can "cheat" for "built-in" providers that are part of libcrypto itself, because we know that the two libcrypto versions are the same - but not for other providers.

Providers built into applications can also "cheat"

doc/man7/provider.pod
```diff
@@ -42,7 +42,7 @@ If it's built-in, the initialization function may have any name.
 
 The initialization function must have the following signature:
 
- int NAME(const OSSL_PROVIDER *provider,
+ int NAME(const OSSL_CORE_PROVIDER *provider,
```
levitte, line 45: I wonder if it shouldn't be called something like `OSSL_CORE_HANDLE`...  there is no "core provider", and the only use this type has anyway is as a handle to the core for some of the upcalls...

include/openssl/core.h
```diff
@@ -25,6 +25,11 @@ extern "C" {
  * to communicate data between them.
  */
 
+/* Opaque handles used by functions from the core */
```
levitte, line 28: Hmmm...  "Opaque handles to be used with core upcall functions"

providers/common/bio_prov.c
```diff
@@ -94,3 +109,110 @@ int ossl_prov_bio_printf(BIO *bio, const char *format, ...)
...
+static const OPENSSL_CTX_METHOD bio_prov_ossl_ctx_method = {
+    bio_prov_ossl_ctx_new,
+    bio_prov_ossl_ctx_free,
+};
+
+BIO *BIO_new_from_core_bio(OPENSSL_CTX *libctx, OSSL_CORE_BIO *corebio)
+{
+    BIO_PROV_GLOBAL *bpgbl = openssl_ctx_get_data(libctx,
+                                                  OPENSSL_CTX_BIO_PROV_INDEX,
+                                                  &bio_prov_ossl_ctx_method);
```
levitte, line 208: Hmmmm, do you really expect the code to deliver several core OSSL_DISPATCH (the *in* argument to the provider init function)?  If not, I'm not sure I understand why you can't have just a single provider global `bio_prov_ossl_ctx_method` that you populate once.

If you really want to tie it to the provider context, you could simply add a BIO method to the provider context itself instead of (unnecessarily) going through the library context.  Just make it an actual structure:

``` C
struct provctx_st {
    OPENSSL_CTX *libctx;
    BIO_METHOD *biocoremeth;
}
```

doc/man7/provider.pod
mattcaswell: I've done this rename

include/openssl/core.h
mattcaswell: Done

providers/common/bio_prov.c
mattcaswell: > Hmmmm, do you really expect the code to deliver several core OSSL_DISPATCH (the in argument to the provider init function)? If not, I'm not sure I understand why you can't have just a single provider global bio_prov_ossl_ctx_method that you populate once.

We've gone around this loop (several times) before. Global variables are problematic in the providers, because you have to figure out how to free them again when the provider unloads (ref-counting does not work). This is a minefield. Lets not go around that loop again.

> If you really want to tie it to the provider context, you could simply add a BIO method to the provider context itself instead of (unnecessarily) going through the library context. Just make it an actual structure:

Its simple to say, but not so simple to do. Do-able - but introducing a provider context object which isn't just the libctx is probably a whole PR in itself. We assume all over the place that the provctx is the library context. Now you could just put the libctx first in the provctx structure - and that assumption would still hold. But it is really ugly. I really don't want to go down that path in this PR.

mattcaswell:
While doing the latest updates I stumbled across a new problem. Namely this code in the legacy provider:

````
    /*
     * We want to make sure that all calls from this provider that requires
     * a library context use the same context as the one used to call our
     * functions.  We do that by passing it along as the provider context.
     */
    *provctx = c_get_libctx(handle);
````
So the provider context is the library context from core. This just doesn't work in the legacy provider because it is not built-in - so the application and the provider could end up using different versions of libcrypto.

The right answer is not use that libctx and create a new one instead. This means (for example) that any fetches done within the legacy provider may have different providers in it to the application - which could lead to unexpected results. In practice, right now, I don't think this is a problem - because I don't think there are any such fetches. However it is likely to become a problem in the future.

In implementing the "right answer" I came across another problem. For some reason calling `OPENSSL_CTX_new()` from the legacy provider results in a linker error. I have no idea why. I'm hoping @levitte might be able to shed some  light.

File: providers/common/bio_prov.c [hunk 4]
```diff
@@ -94,3 +109,110 @@ int ossl_prov_bio_printf(BIO *bio, const char *format, ...)
...
+static int bio_core_write_ex(BIO *bio, const char *data, size_t data_len,
+                             size_t *written)
+{
+    return ossl_prov_bio_write_ex(BIO_get_data(bio), data, data_len, written);
+}
+
+static long bio_core_ctrl(BIO *bio, int cmd, long num, void *ptr)
+{
+    /* We don't support this */
+    assert(0);
```
levitte, line 131: This may be foolish.  What if, for example, someone wants to `BIO_eof(bio)`?  It would be perfectly reasonable to expect that to work properly...

Considering that the BIO ctrl commands have been stable for eons, I don't expect this to become a can of worms.

providers/common/bio_prov.c
levitte: > We assume all over the place that the provctx is the library context.

I sure hope that *we do not*.  There is a central spot where the provider context would be defined, `providers/common/include/prov/provider_ctx.h`, and there is a macro to extract the library context from the provider context - `PROV_LIBRARY_CONTEXT_OF`, *exactly so we can avoid that kind of assumption*.  Why do you think I pushed for that macro in the first place?

providers/legacyprov.c
```diff
@@ -184,11 +184,34 @@ int OSSL_provider_init(const OSSL_PROVIDER *provider,
...
      * a library context use the same context as the one used to call our
      * functions.  We do that by passing it along as the provider context.
      */
-    *provctx = c_get_libctx(provider);
+
+    /*
+     * TODO(3.0): FIXME FIXME FIXME DANGER DANGER DANGER
+     * 
+     * We should really replace this with the code below - but for some reason
+     * I'm getting a linker error when I do so.
```
levitte, line 198: Interesting...  I'm about to look at the same thing, so I can investigate as well.

providers/common/bio_prov.c
levitte: Ah, I found *one spot* where that kind of assumption is made:

https://github.com/openssl/openssl/blob/8c30dfee3ea038b71f339f193331ed6ac11e3055/providers/fips/fipsprov.c#L652

levitte:
> In implementing the "right answer" I came across another problem. For some reason calling `OPENSSL_CTX_new()` from the legacy provider results in a linker error. I have no idea why. I'm hoping @levitte might be able to shed some light.

Is this the linking error you get?

```
/usr/bin/ld: providers/liblegacy.a(liblegacy-lib-context.o): in function `context_init':
/home/levitte/gitwrk/openssl.net/official/_build/../master/crypto/context.c:81: undefined reference to `ossl_property_parse_init'
```

That one is fixed like this:

``` diff
diff --git a/crypto/property/build.info b/crypto/property/build.info
index bfa1f0602f..56f26760c6 100644
--- a/crypto/property/build.info
+++ b/crypto/property/build.info
@@ -2,3 +2,4 @@ LIBS=../../libcrypto
 $COMMON=property_string.c property_parse.c property.c defn_cache.c
 SOURCE[../../libcrypto]=$COMMON property_err.c
 SOURCE[../../providers/libfips.a]=$COMMON
+SOURCE[../../providers/liblegacy.a]=$COMMON
```

levitte:
Following https://github.com/openssl/openssl/pull/11758#discussion_r422452084, I went investigating what a proper provider context structure would actually mean, and stumbled on some bad code in the process.

PRs coming up!

doc/man7/provider-base.pod
```diff
@@ -327,7 +341,7 @@ operation C<BAR>.
 
  /* Provider context */
  struct prov_ctx_st {
-     OSSL_PROVIDER *prov;
+     OSSL_CORE_HANDLE *handle;
```
levitte, line 344: It's interesting that we document that providers should do this, but we don't do this ourselves!

providers/common/bio_prov.c
levitte: Have a look at #11803

providers/common/bio_prov.c
mattcaswell: This is not public API. This is about providing a BIO for our own providers to use. If we need support for additional ctrls then we can add it at that point. Probably I should have called the BIO creation function `bio_new_from_core_bio` i.e. with a lowercase `bio` to emphasise that point.

levitte: The trouble is that *something* inside the provider, or in libcrypto that the provider calls, might assume a normal BIO.

Oh well, I guess we'll get there some day...

include/openssl/core.h
```diff
@@ -25,6 +25,11 @@ extern "C" {
  * to communicate data between them.
  */
 
+/* Opaque handles to be used with core upcall functions */
+typedef void *OSSL_CORE_HANDLE;
```
t8m, line 29: I think this is a wrong way how to declare an opaque handle as it won't provide type safety. The right way IMO is:

`typedef struct ossl_core_handle OSSL_CORE_HANDLE;`

levitte: Oh, now that I see this, I would advice against pointer typedefs. They can be very confusing at times. And I do agree with @t8m.

mattcaswell: Fixed

providers/common/bio_prov.c
mattcaswell: Ok, I rebased this on top of #11803

mattcaswell:
I rebased this on top of #11803 and updated various things according the feedback received so far. Unfortunately I'm now getting a linker error that I don't know how to solve.

providers/common/bio_prov.c
```diff
@@ -7,12 +7,15 @@
...
 
+#include <assert.h>
 #include <openssl/core_numbers.h>
+#include "internal/cryptlib.h"
 #include "prov/bio.h"
 
 static OSSL_BIO_new_file_fn *c_bio_new_file = NULL;
 static OSSL_BIO_new_membuf_fn *c_bio_new_membuf = NULL;
 static OSSL_BIO_read_ex_fn *c_bio_read_ex = NULL;
+static OSSL_BIO_write_ex_fn *c_bio_write_ex = NULL;
```
slontis, line 18: Why is this needed?
The initial reason for adding the bio calls was for fips verification.

include/openssl/core.h
t8m: @mattcaswell You've probably forgot to push the change?

providers/common/bio_prov.c
mattcaswell: Its used by the serializers.

include/openssl/core.h
```diff
@@ -25,6 +25,11 @@ extern "C" {
...
+typedef void *OPENSSL_CORE_CTX;
+typedef void *OSSL_CORE_BIO;
```
levitte, line 31: Likewise for these, please avoid building in the pointer indicator (`*`) in the typedef.  You're obviously also confusing yourself, considering that you're using `OPENSSL_CORE_CTX *` and `OSSL_CORE_BIO *` all over the place :wink: (even though casting "fixes" it, it's still potentially dangerous practice)

providers/legacyprov.c
```diff
@@ -184,11 +184,34 @@ int OSSL_provider_init(const OSSL_PROVIDER *provider,
...
+            return 0;
+
+        /*
+         * We would really like to use the same libctx that the application uses
+         * - but since the application may be linked against a different version
+         * of libcrypto to us we cannot.
+         */
+        *provctx = libctx;
+    }
+#endif
```
levitte, line 215: Conidering I've done the latter in #11803, I assume that this particular change is simply going to disappear

mattcaswell:
Pushed for real this time. I was stupidly working in a hurry last night because I had another commitment in the evening and wanted to finish before then. In doing that I managed to mess up and lose my work :-(.

Anyway, please take another look. Ignore the changes from #11803

providers/legacyprov.c
mattcaswell: Correct

include/openssl/core.h
mattcaswell: Fixed

include/openssl/types.h
```diff
@@ -229,6 +229,11 @@ typedef long ossl_intmax_t;
 typedef unsigned long ossl_uintmax_t;
 #endif
 
+/* Opaque handles to be used with core upcall functions from providers */
+typedef struct ossl_core_handle_st OSSL_CORE_HANDLE;
+typedef struct openssl_core_ctx_st OPENSSL_CORE_CTX;
+typedef struct ossl_core_bio_st OSSL_CORE_BIO;
+
```
levitte, line 236: Hmmm...  I understand in a way, but at the same time, `core.h` contains all types that are specific to the libcrypto\<-\>provider interface.  I'm a bit ambivalent regarding spreading them out like this.  Everyone that have use of these types include `<openssl/core.h>` anyway...

mattcaswell: I've been dithering in my own mind between what is the correct place to have it - and have had it in both at various points. I could live with either. So I'll move it back again if you prefer.

providers/common/provider_ctx.c
```diff
@@ -9,14 +9,31 @@
...
+
+# ifndef FIPS_MODULE
+    ctx->corebiometh = bio_prov_init_bio_method();
+    if (ctx->corebiometh == NULL) {
+        OPENSSL_free(ctx);
+        return NULL;
+    }
+# endif
+
+    return ctx;
```
levitte, line 29: This is a case where I believe it would be simpler if each provider init function did this on their own as needed.

``` C
    if ((libctx = [... whatever ...]) == NULL
        || !(tmpprovctx = PROV_CTX_new()) == NULL
        || !(biometh = bio_prov_init_bio_method()) == NULL) {
        BIO_meth_free(biometh);
        PROV_CTX_free(tmpprovctx);
        [... whatever ...](libctx);
        return 0;
    }
    tmpprovctx->handle = provider;
    tmpprovctx->libctx = libctx;
    tmpprovctx->corebiometh = biometh;
    *provctx = tmpprovctx;
```

The internal FIPS provider wouldn't do the biometh part at all.
This is the pattern I set forth in #11803, and it was knowingly done this way to avoid exactly the kind of linking problem you got yourself into...

providers/common/provider_ctx.c
```diff
@@ -9,14 +9,31 @@
...
 }
 
 void PROV_CTX_free(PROV_CTX *ctx)
 {
+# ifndef FIPS_MODULE
+    BIO_meth_free(ctx->corebiometh);
+# endif
```
levitte, line 36: Likewise, I would leave this to the teardown functions

include/openssl/types.h
levitte: I think I do prefer that.

mattcaswell: Done

providers/common/provider_ctx.c
mattcaswell: Done

providers/common/provider_ctx.c
mattcaswell: I kind of feels wrong that PROV_CTX_free() does not know how to free its own internal members and it is the responsibility of something external to do that. However, for now, I've done it like this. I think we might want to revisit this in the future.

providers/common/bio_prov.c
richsalz: explicit initialization to null/zero/0.0/etc for globals, yuk.  I think the project should come up with a style policy on this.

mattcaswell:
I rebased this on a fresh master now that #11803 has gone in.

providers/defltprov.c
```diff
@@ -585,15 +587,23 @@ int ossl_default_provider_init(const OSSL_PROVIDER *provider,
...
         return 0;
-    PROV_CTX_set0_library_context(*provctx, c_get_libctx(provider));
-    PROV_CTX_set0_provider(*provctx, provider);
+
+    corebiometh = bio_prov_init_bio_method();
+    if (corebiometh == NULL) {
+        PROV_CTX_free(*provctx);
+        *provctx = NULL;
+        return 0;
+    }
```
levitte, line 603: You *could* have a cascade...

``` C
    if ((*provctx = PROV_CTX_new()) == NULL
        || (corebiometh = bio_prov_init_bio_method()) == NULL) {
        PROV_CTX_free(*provctx);
        *provctx = NULL;
        return 0;
    }
```

providers/fips/fipsprov.c
```diff
@@ -727,8 +727,12 @@ int fips_intern_provider_init(const OSSL_PROVIDER *provider,
 
     if ((*provctx = PROV_CTX_new()) == NULL)
         return 0;
-    PROV_CTX_set0_library_context(*provctx, c_get_libctx(provider));
-    PROV_CTX_set0_provider(*provctx, provider);
+    /*
+     * Only works because we are a built-in internal provider. This is not
```
levitte, line 731: "Using the parent library context only works..."

levitte:
Apart from those nits, this looks good

providers/defltprov.c
mattcaswell: Done

providers/fips/fipsprov.c
mattcaswell: Done

mattcaswell:
Nits fixed.

doc/man7/provider-base.pod
```diff
@@ -77,7 +89,7 @@ B<OSSL_get_{name}>.
 For example, the "function" core_gettable_params() has these:
 
  typedef OSSL_ITEM *
-     (OSSL_core_gettable_params_fn)(const OSSL_PROVIDER *prov);
+     (OSSL_core_gettable_params_fn)(const OSSL_CORE_HANDLE *handle);
```
levitte, line 92: This isn't quite right any more...  the returned type is `const OSSL_PARAM *`.  Probably an omission on my part, but would you mind changing that while you're at it?

doc/man7/provider-base.pod
```diff
@@ -131,20 +142,20 @@ F<libcrypto>):
...
-core_get_params() retrieves I<prov> parameters from the core.
+core_get_params() retrieves parameters from the core for the given I<handle>.
 See L</Core parameters> below for a description of currently known
 parameters.
 
 =for comment core_thread_start() TBA
 
 core_get_library_context() retrieves the library context in which the
-B<OSSL_PROVIDER> object I<prov> is stored.
+B<OSSL_CORE_HANDLE> object I<handle> is stored.
```
levitte, line 152: `... retrieves the library context in which the library object for the current provider is stored, accessible through the I<handle>.`

doc/man7/provider-base.pod
```diff
@@ -131,20 +142,20 @@ F<libcrypto>):
...
 This may sometimes be useful if the provider wishes to store a
 reference to its context in the same library context.
 
 core_new_error(), core_set_error_debug() and core_set_error() are
 building blocks for reporting an error back to the core, with
-reference to the provider object I<prov>.
+reference to the handle object I<handle>.
```
levitte, line 158: `... with reference to the I<handle>`

doc/man7/provider-base.pod
mattcaswell: Done

doc/man7/provider-base.pod
mattcaswell: Done

doc/man7/provider-base.pod
mattcaswell: Done

mattcaswell:
All comments addressed and new fixup pushed. Please take another look.

mattcaswell:
There were some non-trivial conflicts in the RSA serializer code as a result of the RSA-PSS work. I've had to rebase this and fix a few things up. @levitte please can you reconfirm? Taking the "approval done" label off until you have.

Unfortunately this means that this PR isn't going to make it into alpha 2 :-(

levitte:
Thanks for making that conflict resolution a separate fixup commit.  That made it very easy to re-review.

mattcaswell:
> Thanks for making that conflict resolution a separate fixup commit. That made it very easy to re-review.

Oh - actually there were two sets of changes. One was "real" conflicts which I resolved during the rebase - these were in `rsa_pub_der` and `rsa_pub_pem`. Then separate to that I had some compilation failures after I finished the rebase which I fixed in a separate fixup commit.

levitte:
Okie.  I just did a quick sweep and things still look good.  My approval stands.

mattcaswell:
Pushed. Thanks.

PR #11765, opened 2020-05-08
Update RSA keygen to use sp800-56b by default and fix infinite loop.

### LINKED ISSUE #11742: EVP_PKEY_keyGen for RSA hangs if pub exponent invalid
Setting rsa public exponent to invalid value such as 1 or 2 or 4 causes EVP_PKEY_keygen to hang.

```
    BIGNUM *bn = NULL;
    EVP_PKEY *pkey = NULL;
    EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_id(NID_rsaEncryption, NULL);
    EVP_PKEY_keygen_init(ctx);
    BN_hex2bn(&bn, "02");
    EVP_PKEY_CTX_set_rsa_keygen_pubexp(ctx, bn);
    EVP_PKEY_keygen(ctx, &pkey);
```

### LINKED ISSUE #11764: RSA keygen regression
beldmit@manul$ LD_LIBRARY_PATH=`pwd` apps/openssl genpkey  -algorithm rsa -pkeyopt rsa_keygen_bits:512 -out key.pem
Error generating key

beldmit@manul$ git bisect bad
5e77b79a8c47f0801f656cfccfcbaaa3ca1035b4 is the first bad commit
commit 5e77b79a8c47f0801f656cfccfcbaaa3ca1035b4
Author: Shane Lontis <shane.lontis@oracle.com>
Date:   Thu May 7 15:31:05 2020 +1000

    Remove gen_get_params & gen_gettable_params from keygen operation
    
    EVP_PKEY_CTX_gettable_params() was missing code for the keygen operation.
    After adding it it was noticed that it is probably not required for this type, so instead
    the gen_get_params and gen_gettable_params have been remnoved from the provider interface.
    gen_get_params was only implemented for ec to get the curve name. This seems redundant
    since normally you would set parameters into the keygen_init() and then generate a key.
    Normally you would expect to extract data from the key - not the object that we just set up
    to do the keygen.
    
    Added a simple settable and gettable test into a test that does keygen.
    
    (Merged from https://github.com/openssl/openssl/pull/11683)

slontis:
Fixes #11742
Fixes #11764

The newer RSA sp800-56b algorithm is being used for the normal case of a non multiprime key of at least length 2048.
Insecure key lengths and multiprime RSA will use the old method.

Bad public exponents are no longer allowed (i.e values less than 65537 or even). Values such as 2 that would cause a infinite loop now result in an error. The value of 3 has been marked as deprecated but is still allowed for legacy purposes.

##### Checklist

### DIFF
--- a/apps/genrsa.c
+++ b/apps/genrsa.c
@@ -33,7 +33,10 @@ static int genrsa_cb(EVP_PKEY_CTX *ctx);
 
 typedef enum OPTION_choice {
     OPT_ERR = -1, OPT_EOF = 0, OPT_HELP,
-    OPT_3, OPT_F4, OPT_ENGINE,
+#ifndef OPENSSL_NO_DEPRECATED_3_0
+    OPT_3,
+#endif
+    OPT_F4, OPT_ENGINE,
     OPT_OUT, OPT_PASSOUT, OPT_CIPHER, OPT_PRIMES, OPT_VERBOSE,
     OPT_R_ENUM, OPT_PROV_ENUM
 } OPTION_CHOICE;
@@ -48,7 +51,9 @@ const OPTIONS genrsa_options[] = {
 #endif
 
     OPT_SECTION("Input"),
+#ifndef OPENSSL_NO_DEPRECATED_3_0
     {"3", OPT_3, '-', "(deprecated) Use 3 for the E value"},
+#endif
     {"F4", OPT_F4, '-', "Use F4 (0x10001) for the E value"},
     {"f4", OPT_F4, '-', "Use F4 (0x10001) for the E value"},
 
@@ -100,9 +105,11 @@ opthelp:
             ret = 0;
             opt_help(genrsa_options);
             goto end;
+#ifndef OPENSSL_NO_DEPRECATED_3_0
         case OPT_3:
             f4 = RSA_3;
             break;
+#endif
         case OPT_F4:
             f4 = RSA_F4;
             break;
--- a/apps/genrsa.c
+++ b/apps/genrsa.c
@@ -54,8 +54,8 @@ const OPTIONS genrsa_options[] = {
 #ifndef OPENSSL_NO_DEPRECATED_3_0
     {"3", OPT_3, '-', "(deprecated) Use 3 for the E value"},
 #endif
-    {"F4", OPT_F4, '-', "Use F4 (0x10001) for the E value"},
-    {"f4", OPT_F4, '-', "Use F4 (0x10001) for the E value"},
+    {"F4", OPT_F4, '-', "Use the Fermat number F4 (0x10001) for the E value"},
+    {"f4", OPT_F4, '-', "Use the Fermat number F4 (0x10001) for the E value"},
 
     OPT_SECTION("Output"),
     {"out", OPT_OUT, '>', "Output the key to specified file"},
--- a/apps/genrsa.c
+++ b/apps/genrsa.c
@@ -48,7 +48,7 @@ const OPTIONS genrsa_options[] = {
 #endif
 
     OPT_SECTION("Input"),
-    {"3", OPT_3, '-', "Use 3 for the E value"},
+    {"3", OPT_3, '-', "(deprecated) Use 3 for the E value"},
     {"F4", OPT_F4, '-', "Use F4 (0x10001) for the E value"},
     {"f4", OPT_F4, '-', "Use F4 (0x10001) for the E value"},
 
--- a/crypto/rsa/rsa_gen.c
+++ b/crypto/rsa/rsa_gen.c
@@ -70,16 +70,10 @@ int RSA_generate_multi_prime_key(RSA *rsa, int bits, int primes,
     return rsa_keygen(NULL, rsa, bits, primes, e_value, cb, 0);
 }
 
-static int rsa_keygen(OPENSSL_CTX *libctx, RSA *rsa, int bits, int primes,
-                      BIGNUM *e_value, BN_GENCB *cb, int pairwise_test)
+#ifndef FIPS_MODULE
+static int rsa_multiprime_keygen(RSA *rsa, int bits, int primes,
+                                 BIGNUM *e_value, BN_GENCB *cb)
 {
-    int ok = -1;
-#ifdef FIPS_MODULE
-    if (primes != 2)
-        return 0;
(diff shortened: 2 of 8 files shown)


beldmit:
Unfortunately it does not work :(

```
beldmit@manul$ LD_LIBRARY_PATH=. apps/openssl genpkey  -algorithm rsa -pkeyopt rsa_keygen_bits:2048 -out key.pem
Error generating key
```

So could you please add this test (and maybe, test for 512 bits until it is forbidden totally)?

apps/genrsa.c
```diff
@@ -48,7 +48,7 @@ const OPTIONS genrsa_options[] = {
 #endif
 
     OPT_SECTION("Input"),
-    {"3", OPT_3, '-', "Use 3 for the E value"},
+    {"3", OPT_3, '-', "(deprecated) Use 3 for the E value"},
     {"F4", OPT_F4, '-', "Use F4 (0x10001) for the E value"},
     {"f4", OPT_F4, '-', "Use F4 (0x10001) for the E value"},
```
levitte, line 53: At some point, someone will have to explain how F4 is the same as 0x10001...

apps/genrsa.c
```diff
@@ -48,7 +48,7 @@ const OPTIONS genrsa_options[] = {
 #endif
 
     OPT_SECTION("Input"),
-    {"3", OPT_3, '-', "Use 3 for the E value"},
+    {"3", OPT_3, '-', "(deprecated) Use 3 for the E value"},
```
paulidale, line 51: Guard this with `#ifndef OPENSSL_NO_DEPRECATED_3_0` perhaps?

beldmit:
@slontis does pkeygen works for you?

slontis:
> @slontis does pkeygen works for you?

yes..

beldmit:
I confirm that #11764 works now (without applying this changes)

apps/genrsa.c
slontis: :)

apps/genrsa.c
slontis: Done

slontis:
> So could you please add this test (and maybe, test for 512 bits until it is forbidden totally)?

With the changes in this PR it now chooses 512 as the keysize in 15-test_genrsa.t.
Added a test case specifically for 2048 bits also so that it covers the normal case also.

apps/genrsa.c
t8m: What about adjusting the text to say "Use Fermat number 4 (0x10001) for the E value" then?

paulidale: @t8m's suggestion is good.  I suggest using it or something similar.

There might be confusion around F₄ being the _fifth_ Fermat number (they start with F₀).  What about: **Use the Fermat number F4 (0x10001) for the E value** instead?

t8m: That's why I did not say "Use fourth Fermat number" :grin: . But I am also ok with @paulidale text, sure.

bernd-edlinger: F4 is the fourth fermat number, right?

paulidale: It's the fifth 😀

F₀, F₁, F₂, F₃, F₄ are 3, 5, 17, 257, 65537 respectively.

bernd-edlinger: So yes, but Fermat Numbers count from 0, see
https://oeis.org/A000215
`Fermat numbers: a(n) = 2 ^ (2^n) + 1`

bernd-edlinger: And even you said F₀, F₁, F₂, F₃, F₄ 
Q.E..D :sweat_smile: :

levitte: I vote for @paulidale's variant

kroeckx:
It's my understanding that the method we implement (B.3.6 Generation of Probable Primes with Conditions Based on Auxiliary Probable Primes) is required for smaller keys, but that for keys starting at 2048 bit we can use B.3.3 Generation of Random Primes that are Probably Prime. And we're only allowed to generate 2048 or larger keys in FIPS mode.

B.3.3 is very close to what we do now, with 2 exceptions:
- We round sqrt(2) to 2 bits, which the B.3.6 implementation also did initially.
- If it's not a prime, instead of generating a new random number, we add 2. It's at least my understanding that this does not generate a bias, but it's clearly not what FIPS 186-4 has.

So it surprises me that we're going to do just the opposite, use B.3.3 for short keys, while we should use B.3.6 for short keys. So I suggest that we always use B.3.6, except for multi prime.

kroeckx:
If we're only going to use the old for multi prime, we can
consider changing the prime generation to only set 1 top bit, and
then check that the multiprime actually has enough enough bits.

slontis:
The logic was very simple for the change. i.e:

Use the newer keygen code for fips and the normal case where possible.
For other existing cases such as multiprime (and weak keys that should not really be used) use the old method.

Please submit a PR if you wish to change the old code.

slontis:
travis error is 390 resource is unavailable

crypto/rsa/rsa_gen.c
```diff
@@ -407,8 +408,29 @@ static int rsa_keygen(OPENSSL_CTX *libctx, RSA *rsa, int bits, int primes,
...
+    /*
+     * Only multi-prime keys or insecure keys with a small key length will use
+     * the older rsa_multiprime_keygen().
+     */
+    if (primes == 2 && bits >= 2048)
+        ok = rsa_sp800_56b_generate_key(rsa, bits, e_value, cb);
+#ifndef FIPS_MODULE
+    else
+        ok = rsa_multiprime_keygen(rsa, bits, primes, e_value, cb);
+#endif /* FIPS_MODULE */
```
t8m, line 429: I am missing here putting some error value on the error stack in case of the FIPS module. Or will some error value be added in upper levels?

slontis: This comment currently applies to a lot of things.. I think a seperate PR is in order to clean up the error handling for the keymanagement keys..

t8m:
The travis error is infrastructure glitch and unrelated.

slontis:
I did a rebase since it was a while ago that this was approved. No manual merging was required, so it should still be ok.

slontis:
Thanks. Merged to master.

PR #11768, opened 2020-05-08
Correct alignment calculation in ssl3_setup_write

### LINKED ISSUE #11766: SSL3_ALIGN_PAYLOAD calculation goes wrong during do_ssl3_write
OpenSSL version 1.1.1d

During SSL_write, it is seen that the write buffer is allocated in ssl3_setup_write_buffer. When the buffer is allocated, it is considered that the align payload is of 3 bytes, and a total of 16472 bytes are allocated (16384 data + 80 (SSL3_RT_SEND_MAX_ENCRYPTED_OVERHEAD) +5 (headerlen)+align). 

When the data is copied to the buffer in do_ssl3_write, the alignment is calculated again. Here, the alignment calculation is based on the address that is allocated for the send buffer. If the memory address of the allocated buffer ends with 0x4 or 0xC, then the calculated align will be of 7bytes in do_ssl3_write function and if the address of the buffer ends with 0x0 or 0x8, then the align will be of 3bytes. 
 
In case, if the data to be sent is of 16384 bytes and if the align is 7 bytes, then the write fails in do_ssl3_write function as the data cannot be fit in the buffer of size 16472 bytes (16384+ 5(header) + 80(SSL3_RT_SEND_MAX_ENCRYPTED_OVERHEAD) +7 bytes align).

Thanks,
Nagesh

mattcaswell:
The alignment calculation in ssl3_setup_write incorrectly results in a
value of 3. The maximum alignment actually used in do_ssl3_write is 7
bytes - so there is a potential to overrun the buffer by up to 4 bytes.

Fortunately, the encryption overhead allowed for is 80 bytes which
consists of 16 bytes for the cipher block size and 64 bytes for the MAC
output. However the biggest MAC that we ever produce is HMAC-384 which is
48 bytes - so we have a headroom of 16 bytes (i.e. more than the 4 bytes
of potential overrun).

Fixes #11766

### DIFF
--- a/ssl/record/ssl3_buffer.c
+++ b/ssl/record/ssl3_buffer.c
@@ -94,7 +94,7 @@ int ssl3_setup_write_buffer(SSL *s, size_t numwpipes, size_t len)
             headerlen = SSL3_RT_HEADER_LENGTH;
 
 #if defined(SSL3_ALIGN_PAYLOAD) && SSL3_ALIGN_PAYLOAD!=0
-        align = (-SSL3_RT_HEADER_LENGTH) & (SSL3_ALIGN_PAYLOAD - 1);
+        align = SSL3_ALIGN_PAYLOAD - 1;
 #endif
 
         len = ssl_get_max_send_fragment(s)


richsalz:
As it's just a "lucky accident" that this isn't a real (dangerous) bug, should the commit explicitly thank the person who reported it?

mattcaswell:
> As it's just a "lucky accident" that this isn't a real (dangerous) bug, should the commit explicitly thank the person who reported it?

Good point. Done.

@nageshshegde - if you would rather have your real name in the commit rather than your github id - please let me know what it is!

nageshshegde:
Thanks for the fix. 

Regards,
Nagesh Hegde

mattcaswell:
I updated the commit to mention Nagesh's proper name.

kaduk:
(Travis failure is the unrelated CHANGES.md markdown nit.)
Matt, are you planning to do the merge (note that I had some comments on the commit message)?

mattcaswell:
> (I don't think it would always be 3.)

Err - yes it will?

The align calculation was previously:

    align = (-SSL3_RT_HEADER_LENGTH) & (SSL3_ALIGN_PAYLOAD - 1);

These are fixed constant values* which simplify to:

    align = (-5) & (8 - 1) = 0xffff...fb & 0x07 = 3

(* Strictly speaking the header files do allow you to define your own value of SSL3_ALIGN_PAYLOAD - but I'm not expecting that anyone actually does this - or at least very few). I could rewrite the commit message in terms of these macros if it makes things clearer, i.e. change the first para to this:

> The alignment calculation in ssl3_setup_write incorrectly results in an alignment allowance of (-SSL3_RT_HEADER_LENGTH) & (SSL3_ALIGN_PAYLOAD - 1) bytes. This equals 3 in almost all cases. The maximum alignment actually used in do_ssl3_write is (SSL3_ALIGN_PAYLOAD - 1). This equals 7 bytes in almost all cases. So there is a potential to overrun the buffer by up to 4 bytes.

kaduk:
Right, I meant that I could define `SSL3_ALIGN_PAYLOAD` as, um, 2, if I was feeling particularly obtuse.
I would be happier with the new proposal for the commit message, but you can use my approval with either old or new version.

mattcaswell:
Pushed to master and 1.1.1. Thanks.

PR #10545, opened 2019-11-30
Revamp the docs using minimalistic markdown

mspncp:
The Landing Page
================

On GitHub the README file of a project serves as its landing page, it is displayed immediately below the directory listing. Currently, the OpenSSL README is a raw text file, which gives a rather dusty impression, compared to all these fancy landing pages you find nowadays on GitHub.

Current Landing Page
--------------------

This is how our landing page looks currenty.

current landing page

New Landing Page
----------------

Wouldn't it be much nicer if it would look something like that?

new landing page

For a long time I've dreamed of such a fancy landing page, but it turned out not to be so easily done.

The Problem
-----------

My first attempt to replace the simple text README file by a fancy markdown file ended up in a file which was not very well readable anymore in a text editor. This was a serious disadvantage for developers downloading the source tarball and trying to get started.

One possibility to avoid this problem would have been to add a separate markdown version of the README file as `.github/README.md`, which would have been favoured by GitHub. But having two different README files is not really a good solution. After trying different options for several weeks, the solution to the problem came to my mind.

Minimalistic Markdown
=====================

The idea is to have a single README file which does not only look pretty when viewed in the browser, but is also well formatted and easy to read when you view the raw file in your favourite text editor. Wasn't that exactly the goal of the original [markdown][] implementation of John Gruber und Aaron Swartz? So I tried to go back to the roots using minimalistic markdown.

With 'minimalistic' markdown, I don't mean a certain markdown dialect. It's just the fact that I try to add as little markup as possible, which could interfere with the reading flow int the text file.

For example,

 * avoid [ATX headings][] and use [setext headings][] instead (works for `<h1>` and `<h2>` headings only).
 * avoid [inline links][] and use [reference links][] instead.
 * avoid [fenced code blocks][], use [indented code blocks][] instead.

[markdown]:             https://daringfireball.net/projects/markdown/
[ATX headings]:         https://github.github.com/gfm/#atx-headings
[setext headings]:      https://github.github.com/gfm/#setext-headings
[inline links]:         https://github.github.com/gfm/#inline-link
[reference links]:      https://github.github.com/gfm/#reference-link
[fenced code blocks]:   https://github.github.com/gfm/#fenced-code-blocks
[indented code blocks]: https://github.github.com/gfm/#indented-code-blocks

Show Cases
==========

### The Landing Page ###

[...]

PR #11770, opened 2020-05-08
Add "md-nits" make target

richsalz:
Also fix a nit in recent CHANGES.md update.

A followup to #11739.  If you have ruby and mdl installed, you can do "make md-nits"
This also cleans up the travis integration a bit.

Ping @mspncp @t8m

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -208,11 +208,11 @@ script:
           if ! gem install mdl ; then
               echo -e '\052\052 FAILED -- GEM INSTALL MDL';
               travis_terminate 1;
-          fi
+          fi;
           if ! $make md-nits ; then
               echo -e '\052\052 FAILED -- MAKE MD-NITS';
               travis_terminate 1;
-          fi
+          fi;
       fi
     - if ! $make2; then
           echo -e '\052\052 FAILED -- MAKE';
--- a/.travis.yml
+++ b/.travis.yml
@@ -205,10 +205,14 @@ script:
           travis_terminate 1;
       fi
     - if test -n "$MARKDOWNLINT" ; then
-          echo -e "====START MARKDOWNLINT====";
-          gem install mdl || travis_terminate 1;
-          mdl -s util/markdownlint.rb . || travis_terminate 1;
-          echo -e "====END MARKDOWNLINT====";
+          if ! gem install mdl ; then
+              echo -e '\052\052 FAILED -- GEM INSTALL MDL';
+              travis_terminate 1;
+          fi
+          if ! $make md-nits ; then
+              echo -e '\052\052 FAILED -- MAKE MD-NITS';
+              travis_terminate 1;
+          fi
       fi
     - if ! $make2; then
           echo -e '\052\052 FAILED -- MAKE';
--- a/Configurations/unix-Makefile.tmpl
+++ b/Configurations/unix-Makefile.tmpl
@@ -519,7 +519,7 @@ clean: libclean
 	-$(RM) `find . -name '*{- platform->depext() -}' \! -name '.*' \! -type d -print`
 	-$(RM) `find . -name '*{- platform->objext() -}' \! -name '.*' \! -type d -print`
 	$(RM) core
-	$(RM) tags TAGS doc-nits cmd-nits
+	$(RM) tags TAGS doc-nits cmd-nits md-nits
 	$(RM) -r test/test-runs
 	$(RM) openssl.pc libcrypto.pc libssl.pc
 	-$(RM) `find . -type l \! -name '.*' -print`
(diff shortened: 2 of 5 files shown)


Configurations/unix-Makefile.tmpl
```diff
@@ -982,13 +982,17 @@ update: generate errors ordinals
...
 
-.PHONY: doc-nits cmd-nits
+.PHONY: doc-nits cmd-nits md-nits
 doc-nits: build_generated
 	$(PERL) $(SRCDIR)/util/find-doc-nits -n -l -e
 
 cmd-nits: build_generated apps/openssl
 	$(PERL) $(SRCDIR)/util/find-doc-nits -c
 
+# Needs ruby and "gem install mdl"
```
levitte, line 992: Another variant is to use the snap store: https://snapcraft.io/install/mdl/debian

richsalz: Do you want me to add that comment?

levitte: Yes.  There are enough installations out there that use general repositories like snap (when there isn't a distribution-specific package) rather than language specific installers to warrant adding that information

levitte: (**I** would like that information, especially after all the scars from battles with gem, npm and what not)

richsalz: okay, I added more comments that anyone will ever want or need :)

.travis.yml
```diff
@@ -205,10 +205,14 @@ script:
...
       fi
     - if test -n "$MARKDOWNLINT" ; then
-          echo -e "====START MARKDOWNLINT====";
-          gem install mdl || travis_terminate 1;
-          mdl -s util/markdownlint.rb . || travis_terminate 1;
-          echo -e "====END MARKDOWNLINT====";
+          if ! gem install mdl ; then
+              echo -e '\052\052 FAILED -- GEM INSTALL MDL';
+              travis_terminate 1;
+          fi
```
levitte, line 211: Terminate with a semicolon.

.travis.yml
```diff
@@ -205,10 +205,14 @@ script:
...
+          if ! $make md-nits ; then
+              echo -e '\052\052 FAILED -- MAKE MD-NITS';
+              travis_terminate 1;
+          fi
```
levitte, line 215: Terminate with a semicolon.

richsalz:
semicolons added.

richsalz:
Pushed an additional commit.  Needs re-review.

levitte:
> Pushed an additional commit. Needs re-review.

Are you sure that you did push?  Github gives *no* indication of an additional commit or a force-push...

richsalz:
NOW I'm sure I pushed :)

levitte:
Damn! mdl is *harsh*

richsalz:
> Damn! mdl is _harsh_

:)  Well, I think it's more a mix of a couple of factors.  First, markdown has no full formal specification as it's designed to be informal. As a result different variants exists and mdl tries to help authors make a portable subset. Second, the project doesn't have a lot of experience with markdown (until @mspncp's yoeman work there were no real md files and his PR was merged at the end of November), and that will grow over time.

But yeah, it's strict :)

levitte:
> First, markdown has no full formal specification

That [isn't entirely true](https://commonmark.org/)

richsalz:
See, you made my point:  That's _one_ version of markdown, and it says that the original was never really specified :)

mspncp:
> > First, markdown has no full formal specification
> 
> That [isn't entirely true](https://commonmark.org/)

I was essentially following the [GitHub Flavored Markdown (GFM) Spec](https://github.github.com/gfm), using a 'minimalistic' markdown style, as I called it in  #10545:


> With 'minimalistic' markdown, I don't mean a certain markdown dialect. It's just the fact that I try to add as little markup as possible, which could interfere with the reading flow int the text file.
> 
> For example,
> 
> * avoid [ATX headings](https://github.github.com/gfm/#atx-headings) and use [setext headings](https://github.github.com/gfm/#setext-headings) instead (works for `<h1>` and `<h2>` headings only).
> * avoid [inline links](https://github.github.com/gfm/#inline-link) and use [reference links](https://github.github.com/gfm/#reference-link) instead.
> * avoid [fenced code blocks](https://github.github.com/gfm/#fenced-code-blocks), use [indented code blocks](https://github.github.com/gfm/#indented-code-blocks) instead.

t8m:
Needs second review.

richsalz:
Thanks for the kind words, @mspncp but really, you did the hard work with your initial conversion!  I just did fluff/OCD nits :)

mspncp:
Merged to master as

43a70f0202 Fix all MD036 (emphasis used instead of heading)
a51f225d0d Add "md-nits" make target

_(The second commit of the three was a fixup for the first one, so I squashed those two.)_

PR #11775, opened 2020-05-09
Fix maximum possible length of cipher text

LujunWeng:
For EVP_EncryptUpdate(), the amount of data written my be up to (inl + cipher_block_size) instead of (inl + cipher_block_size - 1). At least for AES128CBC, when "inl" is divisible by cipher_block_size (16), the space needed is (inl + cipher_block_size).

### DIFF
--- a/doc/man3/EVP_EncryptInit.pod
+++ b/doc/man3/EVP_EncryptInit.pod
@@ -212,7 +212,7 @@ writes the encrypted version to B<out>. This function can be called
 multiple times to encrypt successive blocks of data. The amount
 of data written depends on the block alignment of the encrypted data:
 as a result the amount of data written may be anything from zero bytes
-to (inl + cipher_block_size - 1) so B<out> should contain sufficient
+to (inl + cipher_block_size) so B<out> should contain sufficient
 room. The actual number of bytes written is placed in B<outl>. It also
 checks if B<in> and B<out> are partially overlapping, and if they are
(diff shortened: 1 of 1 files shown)


mattcaswell:
Close/reopen to kick CLA bot

doc/man3/EVP_EncryptInit.pod
```diff
@@ -212,7 +212,7 @@ writes the encrypted version to B<out>. This function can be called
 multiple times to encrypt successive blocks of data. The amount
 of data written depends on the block alignment of the encrypted data:
 as a result the amount of data written may be anything from zero bytes
-to (inl + cipher_block_size - 1) so B<out> should contain sufficient
+to (inl + cipher_block_size) so B<out> should contain sufficient
```
mattcaswell, line 215: This change doesn't seem correct to me? In what circumstance would we write out a whole extra block in an update call?

LujunWeng: Yes. I misunderstood here. Paddings are encrypted in EVP_EncryptFinal_ex and it states the space should be enough for one cipher block. I used (inl+cipher_block_size-1) for both "update" and "final", hence I got the error. If it were (inl+cipher_block_size-1 + cipher_block_size), all would be good.

mattcaswell: Ok - no problem. In that I case I'll close this PR.

PR #11777, opened 2020-05-09
Fix some misunderstandings in our providers' main modules

levitte:
This started with adding forward declarations of all provider side
interface functions, and fixing all compiler errors.

Furthermore, diminish the faulty assumption that the provider context
is and always will be just a library context.  That means adding a
teardown function in all providers that aren't necessarily built into
libcrypto.

### DIFF
--- a/providers/fips/fipsprov.c
+++ b/providers/fips/fipsprov.c
@@ -34,6 +34,15 @@
 #include "prov/provider_util.h"
 #include "self_test.h"
 
+/*
+ * Forward declarations to ensure that interface functions are correctly
+ * defined.
+ */
+static OSSL_provider_teardown_fn fips_teardown;
+static OSSL_provider_gettable_params_fn fips_gettable_params;
+static OSSL_provider_get_params_fn fips_get_params;
+static OSSL_provider_query_operation_fn fips_query;
+
 #define ALGC(NAMES, FUNC, CHECK) { { NAMES, "provider=fips,fips=yes", FUNC }, CHECK }
 #define ALG(NAMES, FUNC) ALGC(NAMES, FUNC, NULL)
 
@@ -208,12 +217,12 @@ static int dummy_evp_call(void *provctx)
     return ret;
 }
 
-static const OSSL_PARAM *fips_gettable_params(const OSSL_PROVIDER *prov)
+static const OSSL_PARAM *fips_gettable_params(void *provctx)
 {
     return fips_param_types;
 }
 
-static int fips_get_params(const OSSL_PROVIDER *prov, OSSL_PARAM params[])
+static int fips_get_params(void *provctx, OSSL_PARAM params[])
 {
     OSSL_PARAM *p;
 
@@ -479,9 +488,8 @@ static const OSSL_ALGORITHM fips_keymgmt[] = {
     { NULL, NULL, NULL }
 };
 
-static const OSSL_ALGORITHM *fips_query(OSSL_PROVIDER *prov,
-                                         int operation_id,
-                                         int *no_cache)
+static const OSSL_ALGORITHM *fips_query(void *provctx, int operation_id,
+                                        int *no_cache)
 {
     *no_cache = 0;
     switch (operation_id) {
@@ -506,13 +514,14 @@ static const OSSL_ALGORITHM *fips_query(OSSL_PROVIDER *prov,
     return NULL;
 }
 
+static void fips_teardown(void *provctx)
+{
+    OPENSSL_CTX_free(PROV_LIBRARY_CONTEXT_OF(provctx));
+}
+
 /* Functions we provide to the core */
 static const OSSL_DISPATCH fips_dispatch_table[] = {
-    /*
-     * To release our resources we just need to free the OPENSSL_CTX so we just
-     * use OPENSSL_CTX_free directly as our teardown function
-     */
-    { OSSL_FUNC_PROVIDER_TEARDOWN, (void (*)(void))OPENSSL_CTX_free },
+    { OSSL_FUNC_PROVIDER_TEARDOWN, (void (*)(void))fips_teardown },
     { OSSL_FUNC_PROVIDER_GETTABLE_PARAMS, (void (*)(void))fips_gettable_params },
     { OSSL_FUNC_PROVIDER_GET_PARAMS, (void (*)(void))fips_get_params },
     { OSSL_FUNC_PROVIDER_QUERY_OPERATION, (void (*)(void))fips_query },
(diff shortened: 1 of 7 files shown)


levitte:
*It should be considered **good coding practice** to always have forward declarations of provider interface functions that are implemented*
That ensures that we're coding them correctly.

providers/legacyprov.c
```diff
@@ -27,19 +36,19 @@ static OSSL_core_gettable_params_fn *c_gettable_params = NULL;
 static OSSL_core_get_params_fn *c_get_params = NULL;
 
 /* Parameters we provide to the core */
-static const OSSL_ITEM legacy_param_types[] = {
-    { OSSL_PARAM_UTF8_PTR, OSSL_PROV_PARAM_NAME },
-    { OSSL_PARAM_UTF8_PTR, OSSL_PROV_PARAM_VERSION },
-    { OSSL_PARAM_UTF8_PTR, OSSL_PROV_PARAM_BUILDINFO },
-    { 0, NULL }
+static const OSSL_PARAM legacy_param_types[] = {
+    OSSL_PARAM_utf8_string(OSSL_PROV_PARAM_NAME, NULL, 0),
```
slontis, line 40: These should be ptrs.. They are pointers everywhere else. I assume this was a conversion mistake.

levitte: Ah, yes, that was mistake.

levitte:
I rebased to get rid of the silly faults in CHANGES.md

providers/fips/fipsprov.c
```diff
@@ -639,35 +647,32 @@ int OSSL_provider_init(const OSSL_PROVIDER *provider,
...
+    if (!dummy_evp_call(libctx))
+        goto err;
 
     *out = fips_dispatch_table;
-    *provctx = ctx;
+    *provctx = libctx;
 
     return 1;
+ err:
+    fips_teardown(*provctx);
```
slontis, line 674: This will leak the libctx.. provctx will not be setup on the err path.

slontis:
> I rebased to get rid of the silly faults in CHANGES.md

Er what (CHANGES.md isnt in this PR)?

levitte:
> Er what (CHANGES.md isnt in this PR)?

Ah, never mind, the CHANGES.md fault is still there.  So no, *this* PR hasn't changed CHANGES.md, but 31b069ecea2c567de22b3874c8e71cc37c921ec9 contains a new entry that the new md-nits check errors on (why that wasn't corrected in #11739, I do not know), thus creating a Travis failure.

providers/fips/fipsprov.c
levitte: Fixed

providers/fips/fipsprov.c
```diff
@@ -667,11 +669,11 @@ int OSSL_provider_init(const OSSL_PROVIDER *provider,
         goto err;
 
     *out = fips_dispatch_table;
-    *provctx = libctx;
 
     return 1;
  err:
     fips_teardown(*provctx);
+    *provctx = NULL;
```
slontis, line 676: I would actually perfer it to use libctx here if it can?

levitte: I don't understand.  if the init function fails, it should clean up the provider context it created, shouldn't it?

slontis: I was thinking that the last thing it should do it assign it to the provctx just before it returns when it passes. On the error case I dont like it setting up the pointer so much..

levitte: ... which means that when we have a proper provider CTX that's its own structure that *contains* a libctx, we will have two teardown procedures, one "special" in this init function, and one that's a downcall.  In other words, code copying all over again, with all the problems that may follow.

No thank you.

slontis: Fair enough - but what we want to make sure is that NO information leaks back to the core if the self test fails. If it does then that is a compliance failure. (That is all I am concerned about with this).

levitte: I completely understand that concern.  Are you trying to say that attaching `libctx` to `*provctx` early is a risk of such leakage?  That has me thinking of these lines:

https://github.com/openssl/openssl/blob/3f2a8d971a44b9aac30a5725ae44cfcb1d4156f1/crypto/provider_core.c#L489-L491

As one can see, the address to `prov->provctx` is passed directly.  We could easily change that code to use an intermediary temporary variable, this only attaching the provider context to the provider after the init function has returned.  Will that be enough to clear your concerns?  I would rather do that.

slontis: Its probably ok with what you have now..

slontis: I just saw your teardown is a one line function which is not really much of a duplication of code.. But anyways..

levitte: Well, I added such a commit anyway.  Others may be equally concerned, so it's just as well that we try to lessen that concern in a general fashion.

levitte:
Did commit b2f8ec3 arrive before or after that approval, @slontis?  If you didn't see it in your review, please re-confirm

richsalz:
#11770 fixes the CHANGES.md nit that was introduced while two PR's collided.

levitte:
Uhmmm, #11770 doesn't fix the current issue, #11788 does...

richsalz:
Yes, #11770 did fix the CHANGES.md issue (see the comment at the top of the PR).  But I rebased so that change is now lost.

levitte:
> Yes, #11770 did fix the CHANGES.md issue

This was the error reported before #11788:

```
: ; mdl -s util/markdownlint.rb CHANGES.md
CHANGES.md:26: MD030 Spaces after list markers

A detailed description of the rules is available at https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md
```

I don't see `MD030` mentioned in `util/markdownlint.rb` and I don't see that #11770 adds it either.

richsalz:
Okay, I'll beat this dead horse into the ground.

MD030 was part of the first markdown cleanup, #11739.
While preparing #11770, I noticed that there was a nit that slipped into CHANGES.md. *As the comment at the top of that PR says:*
```
Also fix a nit in recent CHANGES.md update.
```

So yes, my PR (which you commented on, but did not notice or ignored or forgot) also fixed the build-breaking CHANGES.md nit.  I cannot prove it to you, because I since rebased and pushed the update that removes that change and only updates the ci/make stuff.  If we need to keep at this, perhaps @mspncp can tell me how to create a branch when I only have the last commit before a push.

Or, ya know, you could just agree that I also found and fixed this, but didn't mark it as urgent :)

levitte:
Ok, sorry.  Either way, the rest of #11770 isn't invalidated just because I quickly slipped the CHANGES.md fix.

richsalz:
Oh look, the horse isn't dead :)  No big deal.  Looking forward to review and merge of #11770.

levitte:
I've spent most of today trying to find out what's going on with the `test_ssl_new` test, with zero success.  It got me to add a few more error records, though, that can't be a bad thing...  but suffice to say, I need help.

levitte:
Merged.

b0f3c594083b22f082057719f7bb1aa575e7d5a1 CORE: Fix the signature of OSSL_provider_query_operation_fn
fdaad3f1b31df6827554c378dd8385695a1deed4 Fix some misunderstandings in our providers' main modules
914db66d2337d560b042ac710817c69b89045d52 CORE: Attach the provider context to the provider late

PR #5427, opened 2018-02-21
Rework the partially overlapping handling

bernd-edlinger:
The problem here is that in the default case of EVP_EncryptUpdate
the in and out buffers can have different sizes:

Suppose we have a block_size=16 and already buf_len=15.
EVP_EncryptUpdate is called with in==out and inl==2.

In this case, is_partially_overlapping(in+15, in, 2)
returns false, because the intervals are not overlapping at all.
But ctx->cipher->do_cipher(ctx, out, ctx->buf, bl) clobbers the
second byte before it is used.

PR #11780, opened 2020-05-09
Rework the partially overlapping handling

bernd-edlinger:
The problem here is that in the default case of EVP_EncryptUpdate
the in and out buffers can have different sizes:

Suppose we have a block_size=16 and already buf_len=15.
EVP_EncryptUpdate is called with in==out and inl==2.

In this case, is_partially_overlapping(in+15, in, 2)
returns false, because the intervals are not overlapping at all.
But ctx->cipher->do_cipher(ctx, out, ctx->buf, bl) clobbers the
second byte before it is used.

Make partially overlapping more permissive, i.e.
allow essentially all kinds of overlapping that "work"

Fix error handling in aes_ocb_cipher

Avoid undefined behavior in des_ede3_unwrap

Add a sentence about in-place encryption to the doc
Add a statement about the output buffer size for WRAP mode ciphers
Add a definition for partially overlapping

Add test cases for invalid overlapping
Add test cases for overlapping encryption
Add AES-xxx-CFB1/8 test vectors from NIST SP800-38A
Reuse the second DES-EDE3-CFB1 as BitLen test

##### Checklist

Backport of #5427 to 1.1.1


romen:
> Backport of #5427 to 1.1.1

This is a bit misleading: #5427 was not merged, so these changes are not being backported from `master` as we require.

Also there might be a need to discuss if all the changes in here are fixes or if some are new features (which are not allowed in 1.1.1 without a vote to make an exception).

PR #11787, opened 2020-05-11
Fix d2i_PrivateKey_ex() to work as documented

levitte:
d2i_PrivateKey(), and thereby d2i_PrivateKey_ex(), is documented to
return keys of the type given as first argument |type|, unconditionally.
Most specifically, the manual says this:

> An error occurs if the decoded key does not match type.

However, when faced of a PKCS#8 wrapped key, |type| was ignored, which
may lead to unexpected results.

### DIFF
--- a/crypto/asn1/d2i_pr.c
+++ b/crypto/asn1/d2i_pr.c
@@ -58,6 +58,8 @@ EVP_PKEY *d2i_PrivateKey_ex(int type, EVP_PKEY **a, const unsigned char **pp,
                 goto err;
             EVP_PKEY_free(ret);
             ret = tmp;
+            if (EVP_PKEY_type(type) != EVP_PKEY_base_id(ret))
+                goto err;
         } else {
             ASN1err(0, ERR_R_ASN1_LIB);
             goto err;


levitte:
> However, when faced of a PKCS#8 wrapped key, |type| was ignored, which may lead to unexpected results.

Very notably, it leads to issues with the 'file' OSSL_STORE loader, when it's asked to parse a PKCS#8 wrapped key.  Instead of getting *one* match for the correct tested key type, it's gets matches for *all* tested key types.  See the discussion in #11755

levitte:
> LGTM if the tests pass.

If they don't, I'll eat my virtual hat.

slontis:
> If they don't, I'll eat my virtual hat.

Look forward to the video :)

levitte:
Looks good.  The Travis failure is about CHANGES.md, which is fixed in #11788. 
(and no, that's not subject for hat eating :wink:)

levitte:
Why did you put this in the "3.0 New Core + FIPS" project?  This change has *nothing* to do with that?

levitte:
Rebased to get rid of the md-nits failure

levitte:
Merged

b2952366dd0248bf35c83e1736cd203033a22378 Fix d2i_PrivateKey_ex() to work as documented

levitte:
Baackport to 1.1.1 in #11888

Issue #11743, opened 2020-05-06
Error string format does not match documentation


jaimeebrown:
Error string format output by `ERR_print_errors` has changed from previous versions. Documentation still says:
```
 [pid]:error:[error code]:[library name]:[function name]:[reason string]:[filename]:[line]:[optional text message]
```
However, example output from master is:

```
00:47:4F:BB:77:7F:00:00:error:rsa routines:(unknown function):key size too small:../crypto/rsa/rsa_gen.c:94:
```
Either the format is incorrect, or documentation needs to be updated.

levitte:
The format is indeed incorrect. I

PR #11789, opened 2020-05-11
Fix ERR_print_errors() format to match documentation.

slontis:
##### Checklist

### DIFF
--- a/crypto/err/err_prn.c
+++ b/crypto/err/err_prn.c
@@ -15,6 +15,7 @@
 #include <openssl/crypto.h>
 #include <openssl/buffer.h>
 #include <openssl/err.h>
+#include "crypto/err.h"
 #include "err_local.h"
 
 #define ERR_PRINT_BUF_SIZE 4096
@@ -35,8 +36,9 @@ void ERR_print_errors_cb(int (*cb) (const char *str, size_t len, void *u),
             func = "unknown function";
         if ((flags & ERR_TXT_STRING) == 0)
             data = "";
-        hex = OPENSSL_buf2hexstr((const unsigned char *)&tid, sizeof(tid));
-        BIO_snprintf(buf, sizeof(buf), "%s:error:%s:%s:%s:%s:%d:%s\n",
+        hex = openssl_buf2hexstr_sep((const unsigned char *)&tid, sizeof(tid),
+                                     '\0');
+        BIO_snprintf(buf, sizeof(buf), "%s:error::%s:%s:%s:%s:%d:%s\n",
                      hex == NULL ? "<null>" : hex, lib, func, reason, file,
                      line, data);
         OPENSSL_free(hex);
--- a/crypto/err/err_prn.c
+++ b/crypto/err/err_prn.c
@@ -15,7 +15,6 @@
 #include <openssl/crypto.h>
 #include <openssl/buffer.h>
 #include <openssl/err.h>
-#include "crypto/err.h"
 #include "err_local.h"
 
 #define ERR_PRINT_BUF_SIZE 4096
--- a/crypto/o_str.c
+++ b/crypto/o_str.c
@@ -10,7 +10,6 @@
 #include "e_os.h"
 #include <limits.h>
 #include <openssl/crypto.h>
-#include "crypto/err.h"
 #include "internal/cryptlib.h"
 
 #define DEFAULT_SEPARATOR ':'
--- a/crypto/o_str.c
+++ b/crypto/o_str.c
@@ -13,7 +13,7 @@
 #include "crypto/err.h"
 #include "internal/cryptlib.h"
 
-#define DEFAULT_SEPERATOR ':'
+#define DEFAULT_SEPARATOR ':'
 #define CH_ZERO '\0'
 
 char *CRYPTO_strdup(const char *str, const char* file, int line)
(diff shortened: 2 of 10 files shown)


slontis:
Fixes #11743

levitte:
Side note: I've been pondering refactoring all possible variants of hexdump output into a BIO filter...  that's an effort in itself and won't happen now, but might take away the growth of a gazillion different hexdump functions.

(no, I do not intend to block this PR...  if nothing else, it's a necessary temporary measure)

crypto/o_str.c
```diff
@@ -143,26 +143,24 @@ int OPENSSL_hexstr2buf_ex(unsigned char *buf, size_t buf_n, size_t *buflen,
 
     for (p = (const unsigned char *)str, q = buf, cnt = 0; *p; ) {
         ch = *p++;
-        if (ch == ':')
+        /* A seperator of CH_ZERO means there is no seperator */
```
t-j-h, line 146: separator not seperator

t8m:
The actual test source seems to be missing? Also do I get it right that the added hexstr2buf variant is solely for the test purposes?

levitte:
> Also do I get it right that the added hexstr2buf variant is solely for the test purposes?

... or maybe it's for feature completeness (I've complained about the lack of feature completeness, so I can understand if efforts are being made)

crypto/o_str.c
slontis: > Side note: I've been pondering refactoring all possible variants of hexdump output into a BIO

In that case it sounds like I should just make these functions internal for now.

slontis:
> Also do I get it right that the added hexstr2buf variant is solely for the test purposes?

It is not really a one way function, so if a separator is added it should work both ways.

doc/man3/OPENSSL_hexchar2int.pod
```diff
@@ -19,6 +19,7 @@ OPENSSL_buf2hexstr_ex, OPENSSL_buf2hexstr
                            const unsigned char *buf, long buflen);
  char *OPENSSL_buf2hexstr(const unsigned char *buf, long buflen);
 
+
```
levitte, line 22: Remove this extra empty line

doc/man3/OPENSSL_hexchar2int.pod
```diff
@@ -59,12 +60,12 @@ or -1 on error.
...
 
-OPENSSL_buf2hexstr_ex() and OPENSSL_hexstr2buf_ex() return 1 on
-success, or 0 on error.
+OPENSSL_buf2hexstr_ex() and OPENSSL_hexstr2buf_ex()
+return 1 on success, or 0 on error.
 
 =head1 COPYRIGHT
 
-Copyright 2016-2019 The OpenSSL Project Authors. All Rights Reserved.
+Copyright 2016-2020 The OpenSSL Project Authors. All Rights Reserved.
```
levitte, line 68: Actually, you can simply undo all changes in this file.

include/crypto/err.h
```diff
@@ -15,4 +15,7 @@ void err_cleanup(void);
 int err_shelve_state(void **);
 void err_unshelve_state(void *);
 
+char *openssl_buf2hexstr_sep(const unsigned char *buf, long buflen, char sep);
+unsigned char *openssl_hexstr2buf_sep(const char *str, long *buflen,
+                                      const char sep);
```
levitte, line 20: This is wrongly placed, as these are not ERR functions per se.  Please move these declarations to `cryptlib.h` (side note: that header should never have been moved to `include/internal`, as it was meant as declarations of libcrypto private functions)

crypto/err/err_prn.c
```diff
@@ -15,6 +15,7 @@
 #include <openssl/crypto.h>
 #include <openssl/buffer.h>
 #include <openssl/err.h>
+#include "crypto/err.h"
```
levitte, line 18: Should not be needed at this point

crypto/o_str.c
```diff
@@ -10,8 +10,12 @@
 #include "e_os.h"
 #include <limits.h>
 #include <openssl/crypto.h>
+#include "crypto/err.h"
```
levitte, line 13: Should not be needed

doc/man3/OPENSSL_hexchar2int.pod
```diff
@@ -59,8 +59,8 @@ or -1 on error.
 OPENSSL_buf2hexstr() and OPENSSL_hexstr2buf()
 return a pointer to allocated memory, or NULL on error.
 
-OPENSSL_buf2hexstr_ex() and OPENSSL_hexstr2buf_ex() return 1 on
-success, or 0 on error.
+OPENSSL_buf2hexstr_ex() and OPENSSL_hexstr2buf_ex() return 1 on success,
+or 0 on error.
```
levitte, line 63: If you undo this too, this file can drop entirely from your commit list,

crypto/err/err_prn.c
slontis: These last 3 were already addressed :)

slontis:
I am just removing the err include from the test also..

levitte:
My approval stands

slontis:
Thanks. Merged to master.

PR #11802, opened 2020-05-12
Fix AIX test error due to self test startup failure.

### LINKED ISSUE #11762: test failures on AIX
building the 3.0 alpha on AIX 6.1 with perl 5.24.0 I get massive test failures like this:
```
normg@sky>gmake test
        make depend && make _tests
        ( SRCTOP=.  BLDTOP=.
PERL="/export/localnew/RISC6000.AIX/perl-5.24.0/bin/perl" EXE_EXT= 
/export/localnew/RISC6000.AIX/perl-5.24.0/bin/perl ./test/run_tests.pl  )
01-test_abort.t ....................
# The results of this test will end up in test-runs/test_abort
1..1
Use of uninitialized value in concatenation (.) or string at
../../util/wrap.pl line 14.
Use of uninitialized value in concatenation (.) or string at
../../util/wrap.pl line 14.
Unmatched ) in regex; marked by <-- HERE in m/ '') <-- HERE  eq '' && -d
../../util/../engines;
 = ../../util/../providers
    if ( / at ../../util/wrap.pl line 14.
../../util/wrap.pl ../../test/aborttest => 255
ok 1 - Testing that abort is caught correctly
ok
01-test_sanity.t ...................
# The results of this test will end up in test-runs/test_sanity
1..1
Use of uninitialized value in concatenation (.) or string at
../../util/wrap.pl line 14.
Use of uninitialized value in concatenation (.) or string at
../../util/wrap.pl line 14.
Unmatched ) in regex; marked by <-- HERE in m/ '') <-- HERE  eq '' && -d
../../util/../engines;
 = ../../util/../providers
    if ( / at ../../util/wrap.pl line 14.
../../util/wrap.pl ../../test/sanitytest => 255
not ok 1 - running sanitytest
```
The tests get MUCH further with this edit to wrap.pl
```
index fd24c42c..4aba02ea 100755
--- a/util/wrap.pl
+++ b/util/wrap.pl
@@ -12,9 +12,9 @@ my $std_providers = catdir($there, 'providers');
 my $unix_shlib_wrap = catfile($there, 'util/shlib_wrap.sh');
 
 $ENV{OPENSSL_ENGINES} = $std_engines
-    if ($ENV{OPENSSL_ENGINES} // '') eq '' && -d $std_engines;
+    if !defined($ENV{OPENSSL_ENGINES}) && -d $std_engines;
 $ENV{OPENSSL_MODULES} = $std_providers
-    if ($ENV{OPENSSL_MODULES} // '') eq '' && -d $std_providers;
+    if !defined($ENV{OPENSSL_MODULES}) && -d $std_providers;
 ```
After the above edit here is the result:
```
Test Summary Report
-------------------
03-test_fipsinstall.t            (Wstat: 512 Tests: 12 Failed: 2)
  Failed tests:  3-4
  Non-zero exit status: 2
30-test_evp.t                    (Wstat: 256 Tests: 27 Failed: 1)
  Failed test:  1
  Non-zero exit status: 1
30-test_evp_fetch_prov.t         (Wstat: 10496 Tests: 54 Failed: 41)
  Failed tests:  2, 9-28, 35-54
  Non-zero exit status: 41
80-test_ssl_new.t                (Wstat: 7424 Tests: 31 Failed: 29)
  Failed tests:  1-22, 24-29, 31
  Non-zero exit status: 29
80-test_ssl_old.t                (Wstat: 512 Tests: 12 Failed: 2)
  Failed tests:  1, 8
  Non-zero exit status: 2
90-test_sslapi.t                 (Wstat: 512 Tests: 3 Failed: 2)
  Failed tests:  2-3
  Non-zero exit status: 2
90-test_sslprovider.t            (Wstat: 512 Tests: 3 Failed: 2)
  Failed tests:  1, 3
  Non-zero exit status: 2
Files=197, Tests=1971, 3348 wallclock secs ( 7.94 usr  1.02 sys + 1844.35 cusr 71.37 csys = 1924.68 CPU)
Result: FAIL
make: The error code from the last command is 1.
```

### PR DESCRIPTION
**Pull request by slontis**
Fixes #11762

DEP entry points for init and fini do not work for _AIX.
As the platform does not support this mechanism the best thing we can do is just progress the FIPS_state in the run_once section. I have made this the default behaviour for all unknown platforms. (Since it may be better to be non compliant than just falling over on other platforms).

##### Checklist

### DIFF
--- a/providers/fips/self_test.c
+++ b/providers/fips/self_test.c
@@ -37,9 +37,16 @@ static int FIPS_state = FIPS_STATE_INIT;
 static CRYPTO_RWLOCK *self_test_lock = NULL;
 static unsigned char fixed_key[32] = { 0 };
 
+#if !defined(_WIN32) && !defined(__CYGWIN__)                                   \
+    && !defined(_AIX) && !defined(__sun)                                       \
+    && !defined(__hpux) && !defined(__GNUC__)
+# error "Platform does not support fips: to build add 'disable-fips' to the config options"
+#endif
+
 static CRYPTO_ONCE fips_self_test_init = CRYPTO_ONCE_STATIC_INIT;
 DEFINE_RUN_ONCE_STATIC(do_fips_self_test_init)
 {
+    int ret = 0;
     /*
      * This lock gets freed in platform specific ways that may occur after we
      * do mem leak checking. If we don't know how to free it for a particular
@@ -47,7 +54,16 @@ DEFINE_RUN_ONCE_STATIC(do_fips_self_test_init)
      * mem leak checking while we allocate this.
      */
     self_test_lock = CRYPTO_THREAD_lock_new();
-    return self_test_lock != NULL;
+    ret = (self_test_lock != NULL);
+
+    /*
+     * For platforms that have no support for a DEP such as AIX we progress the
+     * state in the same manner as the DEP would.
+     */
+#if defined(_AIX)
+    FIPS_state = FIPS_STATE_SELFTEST;
+#endif
+    return ret;
 }
 
 #define DEP_DECLARE()                                                          \

### CODE REVIEW

**Comment by mattcaswell** [comment]
> I have made this the default behaviour for all unknown platforms.

I'm not sure if that is a good idea. By making this the default behaviour we will never discover different mechanisms for different platforms, i.e. there may well be a way to do a DEP on some platforms - but we just haven't implemented it. By making this change it will "just work" without a DEP and hence be silently non-compliant.

**Comment by slontis** [comment]
> By making this the default behaviour we will never discover different mechanisms for different platforms,

In order to be actually FIPS compliant it needs to go thru a lab validation process anyway.
So breaking the build for people running different platforms doesnt seem like the best approach to me.

**Comment by richsalz** [comment]
>In order to be actually FIPS compliant it needs to go thru a lab validation process anyway.

Not quite.  IF 3.0 follows previous validations, then downstream users can claim vendor-affirmed. But only for platforms that OpenSSL provides. So for platforms that openssl doesn't provide FIPS validation, almost anything can go.

**Comment by slontis** [comment]
> OpenSSL provides. So for platforms that openssl doesn't provide FIPS validation, almost anything can go.

They are not fips validated when 'anything goes'.
They can claim vendor affirmed if it is the same platform - not for different platforms (and from the discussions we have had so far it now needs to be specific hardware).
i.e there is a reason you were asked for non-generic platform specs for fips validation.

If we are going to head down this path, then it should be a compile error - not fail during testing.

**Comment by paulidale** [comment]
I think that a compile error would be better than a test failure.

**Comment by slontis** [comment]
I have changed the code to just add the _AIX case..
Other unknown platforms will cause a compile error unless disable-fips is used.

**Comment by mattcaswell** [comment]
> Other unsupported platforms will cause a compile error unless disable-fips is used.

I think that is absolutely fine - especially during the alpha/beta phase. During that phase we definitely want to know about platforms that can't do this. As we approach final release we can review this and decide what is best moving forwards.

**Comment by richsalz** [comment]
I agree with what was done here.  When I said "anything goes" I meant that the *project* could do anything it wanted. :)

**Comment by mspncp** [comment]
Editorial side note: I stumbled over the 'DEP' acronym in your commit message and it took me a little while to figure out what exactly it expands to. Maybe you could (when merging) change

    DEP entry points for init and fini ...
  
to

    Default Entry Points (DEP) for init and fini ...

**Comment by makr** [comment]
I just stumbled over this one. There is a way to specify init and fini routines on AIX. You can do this at link time with "-binitfini:init:cleanup". See [ld](https://www.ibm.com/support/knowledgecenter/ssw_aix_71/l_commands/ld.html) - Binder section.
Reading again, I am not quite sure, if this is actually relevant, though. Just ignore me then.

**Comment by t-j-h** [comment]
My understanding is that AIX actually supports DEP - just in a different manner ... called __start

**Comment by slontis** [comment]
It looks like __start is for executables.
If someone has more time to play around with this then another PR can be made to update it.

**Comment by mattcaswell** [comment]
Just taking my approval off this for the moment. 

> I just stumbled over this one. There is a way to specify init and fini routines on AIX. You can do this at link time with "-binitfini:init:cleanup". See ld - Binder section.

This seems relevant. If AIX does in fact support DEPs then we should not progress the `FIPS_state` on that platform. If we don't have time to investigate it, it would be better to just keep the compilation failure change for that platform.

**Comment by makr** [comment]
I am testing with HEAD of master. To verify the functionality test_fipsinstall should not fail anymore, right?

**Comment by makr** [comment]
Two things need to be edited:

1. In `providers/fips/self_test.c` line 95:
```C
#elif defined(__sun) || defined(_AIX)
```
so that the pragmas init and fini declare init and cleanup. This will work for the aix{,64}-cc targets. I don't have any gcc available on the AIX machines, so cannot test these targets.

2. This one is difficult for me to spot, hence no PR. In the final `Makefile` where fips.so is constructed the options
```
-Wl,-bE:providers/fips.ld
```
need to be extended to
```
-Wl,-binitfini:init:cleanup,-bE:providers/fips.ld
```
If I am guessing that right, this is `$shared_defflag`, but it should only be used on shared modules that actually have any `init()` and `cleanup()`. Any suggestions how to fit the second case into the build system will lead to a PR :smile: ...

With these changes the whole test suite passes.

slontis:
I did try 1) by itself and it compiled but never ran.
@levitte any suggestions for implementing part 2.

makr:
I think I've found a way. Will need to test this on Monday on an AIX machine and will then create a PR.

slontis:
Closing in favor of #11950

PR #11803, opened 2020-05-12
PROV: Add a proper provider context structure for OpenSSL providers

levitte:
The provider context structure is made to include the following information:

- The core provider handle (first argument to the provider init
  function).  This handle is meant to be used in all upcalls that need
  it.

- A library context, used for any libcrypto calls that need it, done in
  the provider itself.

Regarding the library context, that's generally only needed if the
provider makes any libcrypto calls, i.e. is linked with libcrypto.  That
happens to be the case for all OpenSSL providers, but is applicable for
other providers that use libcrypto internally as well.

The normal thing to do for a provider init function is to create its own
library context.  For a provider that's meant to become a dynamically
loadable module, this is what MUST be done.
However, we do not do that in the default provider; it uses the library
context associated with the core provider handle instead.  This is
permissible, although generally discouraged, as long as the provider in
question is guaranteed to be built-in, into libcrypto or into the
application that uses it.

### DIFF
--- a/crypto/property/build.info
+++ b/crypto/property/build.info
@@ -2,3 +2,4 @@ LIBS=../../libcrypto
 $COMMON=property_string.c property_parse.c property.c defn_cache.c
 SOURCE[../../libcrypto]=$COMMON property_err.c
 SOURCE[../../providers/libfips.a]=$COMMON
+SOURCE[../../providers/liblegacy.a]=$COMMON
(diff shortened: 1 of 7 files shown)


levitte:
This is WIP because it currently includes everything from #11777.  As soon as that one is merged, I will rebase this and take it out of WIP.

This PR is related to a discussion over in #11758: https://github.com/openssl/openssl/pull/11758/files#r421464885

levitte:
#11777 got merged just now, so I rebased this and took it out of WIP

mattcaswell:
Approval of course assumes that the CIs pass

levitte:
Of course, there had to be one small nit...  re-approval needed

mattcaswell:
I just pushed this in order that I can progress #11758

Issue #8332, opened 2019-02-25
EC interface changes


bbbrumley:
Opening a discussion about reducing the EC interface in future versions where API changes are allowed. Tagging @romen if there's something I've missed.

* ~~Entire concept of `EC_METHOD` should not be public facing.~~ done in #11928
* ~~Deprecate `EC_GROUP_precompute_mult` `EC_GROUP_have_precompute_mult` `EC_KEY_precompute_mult`. The `speed` app is the only software I've ever found using them. (Besides, with recent EC changes most use cases don't have access to this type of precomp anyway. Only the hard coded constants from the custom `EC_METHOD`s.)~~ done in #11851
* ~~Deprecate `EC_POINTs_mul`. The `aG + bP` interface of `EC_POINT_mul` is enough for crypto.~~ done in #11807
* ~~Deprecate `EC_POINT_set_Jprojective_coordinates_GFp` `EC_POINT_get_Jprojective_coordinates_GFp`. Whether a point has 3, 4, 5, ... coordinates, and what coordinate system the `EC_METHOD` uses, should not be public facing.~~ done in #11527
* ~~Deprecate `EC_POINT_make_affine` and `EC_POINTs_make_affine` since the `EC_POINT` structure has been opaque for some time now.~~ done in #11874
* Investigate deprecating `EC_GROUP_set_curve` since it seems uncommon to reuse an `EC_GROUP` that way instead of other direct constructors for `EC_GROUP`. Suggested in #11928

davidben:
+1 to each of these. As a data point, we've dropped all of these from BoringSSL without trouble. In practice, people don't seem to really use them.

The one footnote is here is `EC_GROUP_method_of` + `EC_METHOD_get_field_type` is currently the only way to check whether a group is GFp or GF2m. See also issue #6275. (As a compatibility hack, BoringSSL currently returns a dummy pointer out of `EC_GROUP_method_of` and makes `EC_METHOD_get_field_type` always return `NID_X9_62_prime_field` because we don't do GF2m at all.)

bbbrumley:
> See also issue #6275

Great! They are good arguments.

> `EC_GROUP_method_of` + `EC_METHOD_get_field_type` is currently the only way to check whether a group is GFp or GF2m

[...]

PR #11807, opened 2020-05-13
deprecate EC_POINTs_mul function

bbbrumley:
Reiterating from #8332:

* Deprecate `EC_POINTs_mul`. The `aG + bP` interface of `EC_POINT_mul` is enough for crypto.

Since I started poking at the `ec` module in 2007, `test/ectest.c` is the only linking application I've ever seen using `EC_POINTs_mul`.

I started out trying to remove the calls from `ectest.c` but the tests are honestly such a mess, and those parts that exercise `EC_POINTs_mul` have not been updated in forever. So I chickened out and used the suggestion from @mattcaswell. I'll revisit those tests (hopefully with more deprecations) after 3.0. Especially when I have time to change the internal `mul` interface in `EC_METHOD`.


bbbrumley:
Oh I see from the CI failure -- lots of checks in `EC_POINTs_mul`. Let me take care of that now.

mattcaswell:
Looks like Travis still has some relevant failures.

bbbrumley:
> Looks like Travis still has some relevant failures.

I mean it's clear it will fail if you try to compile without the deprecated symbols. I guess I was hoping `# OPENSSL_SUPPRESS_DEPRECATED` and the test framework would somehow figure it out.

I don't really know what to do.

Should I wrap the deprecated calls somehow with `ifdef` blocks? (Maybe the tests will still function, I'm not sure. It's a bunch of spaghetti testing right now so it is hard to extract chunks.)

mattcaswell:
> I mean it's clear it will fail if you try to compile without the deprecated symbols. I guess I was hoping `# OPENSSL_SUPPRESS_DEPRECATED` and the test framework would somehow figure it out

`OPENSSL_SUPPRESS_DEPRECATED` just suppresses compiler warnings if you attempt to use deprecated functions. 

You need to separately decide what to do if the symbols are not present at all. Options are:

1) Don't compile or run the entire test in that case (seems a bit over the top)
2) Rewrite the code using the deprecated interface to not use it
3) Don't compile the bits of the test that use the deprecated API

Of these (2) is the preferred option. The main reason for keeping the old function call there would be if the entire purpose of the test is to test the deprecated function (in which case do (3)). If the use of the deprecated function is incidental to the thing being tested, then we should rewrite it.

bbbrumley:
> You need to separately decide what to do if the symbols are not present at all

Got it -- thanks @mattcaswell . Unfortunately in this case the test functions are mixed between "normal" testing and also directly testing this deprecated functionality. So I don't think (2) is an option given the time frame. Maybe I can make (3) work. I'll have a go this evening.

bbbrumley:
> Looks like Travis still has some relevant failures.

OK that wasn't as hackish as I though it would be. Fixed in c3c7060c589852494adffced6cbb1d467b98e523 hopefully.

bbbrumley:
Haha so mixed declaration and code failures -- I obviously don't even understand what that _really_ means because:

https://github.com/openssl/openssl/blob/master/test/ectest.c#L66-L68

and that's why I moved those in the first place. Anyway I'll take a look now ...

crypto/err/openssl.txt
```diff
@@ -696,6 +696,7 @@ EC_F_EC_POINT_INVERT:210:EC_POINT_invert
 EC_F_EC_POINT_IS_AT_INFINITY:118:EC_POINT_is_at_infinity
 EC_F_EC_POINT_IS_ON_CURVE:119:EC_POINT_is_on_curve
 EC_F_EC_POINT_MAKE_AFFINE:120:EC_POINT_make_affine
+EC_F_EC_POINT_MUL:309:
```
bbbrumley, line 699: @mattcaswell out of curiosity, what is going on here after `make update`? I'm not great with the error side of OpenSSL, but looking at the diff I'd expect the function name on the RHS.

richsalz: OpenSSL no longer displays the function name, so that field is now blank.

mattcaswell:
Pushed. Thanks!

PR #11808, opened 2020-05-13
Add OPENSSL_CTX to CMP

slontis:
##### Checklist


slontis:
Nothing to see here - move along..

slontis:
Thats weird - I have gotten quite a few other peoples branches that way.. Whatever works for you..
Feel free to force push today - since I will probably go offline soon.

slontis:
Good work.. we are nearly finished :)

DDvO:
I've just tested that I can also force-push :)

DDvO:
Thanks for setting this up - good night to down under!

DDvO:
Oops, I did not mean to close this PR, but pruning the two test commits did that automatically.
Yet I'm confident that pushing new (real) commits will re-open it...

slontis:
If not we should be able to reopen manually once there is a commit..

DDvO:
I just wanted to push my first commit with the message
```
add OSSL_CMP_CTX_new_with_libctx() including documentation
```
but now I'm not allowed to push any more :(
Seems you need to revive the branch.

slontis:
I added my bad commit back in after a rebase..

DDvO:
Thanks.
I've rebased my commit on that, then removed your initial commit, and force-pushed mine.
This worked fine :)

DDvO:
Quoting from https://github.com/openssl/openssl/pull/11470#issuecomment-627774125
>>> See if you can do
>>> `git fetch git@github.com:slontis/openssl.git  cmp_provider_support:{yourbranchname}`

>> This did not work directly, but as follows:
>> ```
>> git clone git@github.com:slontis/openssl.git slontis
>> cd slontis
>> git fetch origin cmp_provider_support:cmp_provider_libext
>> ```

> @DDvO Thats weird - I have gotten quite a few other peoples branches that way.

Just for the record: I just found on the top of this page that
GitHub suggests checking out your branch locally as follows:
```
git fetch origin refs/pull/11808/head
git checkout -b slontis/cmp_provider_support FETCH_HEAD
```

which I've been able to follow successfully, adapting it as follows:
```
git fetch upstream refs/pull/11808/head
git checkout -b cmp_provider_support FETCH_HEAD
```

crypto/cmp/cmp_ctx.c
```diff
@@ -123,9 +128,16 @@ OSSL_CMP_CTX *OSSL_CMP_CTX_new(void)
 
  err:
     OSSL_CMP_CTX_free(ctx);
+    X509err(0, ERR_R_MALLOC_FAILURE);
     return NULL;
 }
 
+OSSL_CMP_CTX *OSSL_CMP_CTX_new(void)
```
mattcaswell, line 135: Do we need this variant. I'd be tempted to just have the one above and call *that* one `OSSL_CMP_CTX_new`. We couldn't do that with other APIs because of backwards compatibility - but that isn't an issue with CMP since its never appeared in a stable release.

DDvO: All right, done.
I've added the two new parameters also to `OSSL_CMP_SRV_CTX_new()`.

DDvO: @slontis, therefore I also had to adapt the invocations of these `new()` functions in the tests.
For now I've just put the defaults there: `(NULL, NULL)`.

DDvO:
@slontis, I hope to finish the remaining functions tomorrow.

slontis:
I have added some support to cmp_server_test in a similiar manner to the server_tests..
It will be a similiar process for any other tests that we want to do. With these changes we can pick up any fetches that are not using the libctx.

The 3 tests that are run now have extra params which are:

"none" - This will just work how it did before - which has the default provider using the default library context of NULL.

"default", "default.cnf" - This sets up the default library context to use a NULL provider, and a second library context which gets loaded with the default library. This means that any calls that fetch with a libctx = NULL will fail (as there is no algorithms in the default NULL lib ctx).

"fips", "fips.cnf" - This is similiar to the "default" case above except it uses the fips provider.
If you try to fetch an algorithm that is not in the fips provider then it will fail.

Note that most of the nastiness in the .t file is related to fips.
For fips it also runs a
```openssl fipsinstall ```
application that generates some additional configuration file data required by the fips provider.

You will now see fetches failing with these changes.
OSSL_CRMF_pbm_new() is the first one..
EVP_get_digestbyobj() needs to be replaced with a EVP_MD_FETCH(libctx, )
There is also a EVP_MAC_fetch(libctx) in this function)

**Comment by mattcaswell** [comment]
> "none" - This will just work how it did before - which has the default provider using the default library context of NULL.

Not entirely sure that this run is necessary. I did this with `test_ssl_new` because there were a few tests that were not otherwise run in the `default` and `fips` cases because of some stuff that wasn't fully transferred to providers at the time (such as PSS). However for other tests I just do 2 passes so as not to unduly lengthen the test run time.

**Comment by DDvO** [comment]
> @slontis, I hope to finish the remaining functions tomorrow.

Finally, I'm done with that. I hope I caught all places that needed to be extended.

This was much more work than expected, 
since it turned out that the general support for signing and verifying ASN.1 structures also was lacking those `OPENSSL_CTX *libctx, const char *propq` parameters. So I added to `crypto/asn1/`:
```
int ASN1_item_sign_ex(const ASN1_ITEM *it, X509_ALGOR *algor1,
                      X509_ALGOR *algor2, ASN1_BIT_STRING *signature,
                      const void *data, const ASN1_OCTET_STRING *id,
                      EVP_PKEY *pkey, const EVP_MD *md,
                      OPENSSL_CTX *libctx, const char *propq);
int ASN1_item_verify_ex(const ASN1_ITEM *it, const X509_ALGOR *alg,
                        const ASN1_BIT_STRING *signature, const void *data,
                        const ASN1_OCTET_STRING *id, EVP_PKEY *pkey,
                        OPENSSL_CTX *libctx, const char *propq);
```
along with some aux functions etc., 
which also simplified `X509_verify_ex()` and `X509_REQ_verify_ex()` etc. in `crypto/x509/x_all.c` 
and `ossl_cmp_calc_protection()` in `crypto/cmp/cmp_protect.c`.

**Comment by DDvO** [comment]
> I have added some support to cmp_server_test

@slontis, where can I find this?

**Comment by slontis** [comment]
> @slontis, where can I find this?

It looks like you force pushed before pulling. I thought that might happen :).

**Comment by slontis** [comment]
I think I might just give you the patch.

From 785624168c69f7d6f34bfa30a4c8af5fe10970ab Mon Sep 17 00:00:00 2001
From: Shane Lontis <shane.lontis@someplace.com>
Date: Thu, 14 May 2020 12:32:44 +1000
Subject: [PATCH] Add libctx/provider support to cmp_server_test

---
 test/cmp_server_test.c            | 41 ++++++++++++++++++++++++++++++++++++---
 test/recipes/65-test_cmp_server.t | 40 +++++++++++++++++++++++++++++++++++---
 2 files changed, 75 insertions(+), 6 deletions(-)

diff --git a/test/cmp_server_test.c b/test/cmp_server_test.c
index dc52a25..44c1c08 100644
--- a/test/cmp_server_test.c
+++ b/test/cmp_server_test.c
@@ -10,6 +10,7 @@
  */
 
 #include "cmp_testlib.h"
+#include <openssl/provider.h>
 
 typedef struct test_fixture {
     const char *test_case_name;
@@ -18,6 +19,8 @@ typedef struct test_fixture {
     OSSL_CMP_MSG *req;
 } CMP_SRV_TEST_FIXTURE;
 
+static OPENSSL_CTX *libctx = NULL;
+static OSSL_PROVIDER *default_null_provider = NULL, *provider = NULL;
 static OSSL_CMP_MSG *request = NULL;
 
 static void tear_down(CMP_SRV_TEST_FIXTURE *fixture)
@@ -67,7 +70,7 @@ static int execute_test_handle_request(CMP_SRV_TEST_FIXTURE *fixture)
     OSSL_CMP_ERRORMSGCONTENT *errorContent;
     int res = 0;
 
-    if (!TEST_ptr(client_ctx = OSSL_CMP_CTX_new(NULL, NULL))
+    if (!TEST_ptr(client_ctx = OSSL_CMP_CTX_new(libctx, NULL))
             || !TEST_true(OSSL_CMP_CTX_set_transfer_cb_arg(client_ctx, ctx)))
         goto end;
 
@@ -119,23 +122,55 @@ static int test_handle_request(void)
 void cleanup_tests(void)
 {
     OSSL_CMP_MSG_free(request);
+    OSSL_PROVIDER_unload(default_null_provider);
+    OSSL_PROVIDER_unload(provider);
     return;
 }
 
+OPT_TEST_DECLARE_USAGE("cmp_server_test CR_protected_PBM_1234.der modulename"
+                       " [fips_conf_file]\n")
+
 int setup_tests(void)
 {
     const char *request_f;
+    const char *module_name;
 
     if (!test_skip_common_options()) {
         TEST_error("Error parsing test options\n");
         return 0;
     }
 
-    if (!TEST_ptr(request_f = test_get_argument(0))) {
-        TEST_error("usage: cmp_server_test CR_protected_PBM_1234.der\n");
+    if (!TEST_ptr(request_f = test_get_argument(0))
+            || !TEST_ptr(module_name = test_get_argument(1))) {
+        TEST_error("usage: cmp_server_test CR_protected_PBM_1234.der "
+                   "modulename [fips_conf_file]\n");
         return 0;
     }
 
+    if (strcmp(module_name, "none") != 0) {
+        const char *config_fname = test_get_argument(2);
+
+        default_null_provider = OSSL_PROVIDER_load(NULL, "null");
+        libctx = OPENSSL_CTX_new();
+        if (!TEST_ptr(libctx)) {
+            TEST_error("Failed to create libctx\n");
+            return 0;
+        }
+
+        if (config_fname != NULL
+                && !TEST_true(OPENSSL_CTX_load_config(libctx, config_fname))) {
+            TEST_error("Error loading configfile %s\n", config_fname);
+            return 0;
+        }
+
+        provider = OSSL_PROVIDER_load(libctx, module_name);
+        if (!TEST_ptr(provider)) {
+            TEST_error("Failed to load provider %s\n", module_name);
+            return 0;
+        }
+    }
+
     if (!TEST_ptr(request = load_pkimsg(request_f))) {
         cleanup_tests();
         return 0;
diff --git a/test/recipes/65-test_cmp_server.t b/test/recipes/65-test_cmp_server.t
index 87dbdb1..c94aa39 100644
--- a/test/recipes/65-test_cmp_server.t
+++ b/test/recipes/65-test_cmp_server.t
@@ -9,10 +9,22 @@
 
 use strict;
-use OpenSSL::Test qw/:DEFAULT data_file/;
+use OpenSSL::Test qw/:DEFAULT data_file srctop_dir bldtop_file bldtop_dir srctop_file/;
+
 use OpenSSL::Test::Utils;
 
+BEGIN {
 setup("test_cmp_server");
+}
+
+use lib srctop_dir('Configurations');
+use lib bldtop_dir('.');
+use platform;
+
+my $no_fips = disabled('fips') || ($ENV{NO_FIPS} // 0);
+
+$ENV{OPENSSL_MODULES} = bldtop_dir("providers");
+$ENV{OPENSSL_CONF_INCLUDE} = bldtop_dir("providers");
 
 plan skip_all => "This test is not supported in a no-cmp build"
     if disabled("cmp");
@@ -20,7 +32,29 @@ plan skip_all => "This test is not supported in a no-cmp build"
 plan skip_all => "This test is not supported in a no-ec build"
     if disabled("ec");
 
-plan tests => 1;
+plan tests => 2
+              + ($no_fips ? 0 : 2); #fips install + fips test
 
 ok(run(test(["cmp_server_test",
-             data_file("CR_protected_PBM_1234.der")])));
+             data_file("CR_protected_PBM_1234.der"),
+             "none"])));
+
+ok(run(test(["cmp_server_test",
+             data_file("CR_protected_PBM_1234.der"),
+             "default",
+             srctop_file("test", "default.cnf")])));
+
+unless ($no_fips) {
+    ok(run(app(['openssl', 'fipsinstall',
+                '-out', bldtop_file('providers', 'fipsmodule.cnf'),
+                '-module', bldtop_file('providers', platform->dso('fips')),
+                '-provider_name', 'fips', '-mac_name', 'HMAC',
+                '-macopt', 'digest:SHA256', '-macopt', 'hexkey:00',
+                '-section_name', 'fips_sect'])),
+       "fipsinstall");
+
+    ok(run(test(["cmp_server_test",
+                 data_file("CR_protected_PBM_1234.der"),
+                 "fips",
+                 srctop_file("test", "fips.cnf")])));
+}
\ No newline at end of file
-- 
2.7.4

File: crypto/crmf/crmf_pbm.c [hunk 2]
```diff
@@ -148,7 +155,7 @@ int OSSL_CRMF_pbm_new(const OSSL_CRMF_PBMPARAMETER *pbmp,
      * compute the key used in the MAC process.  All implementations MUST
      * support SHA-1.
      */
-    if ((m = EVP_get_digestbyobj(pbmp->owf->algorithm)) == NULL) {
+    if ((owf = EVP_get_digestbyobj(pbmp->owf->algorithm)) == NULL) {
```
slontis, line 158: I will have to look into what to do with this call...
See https://github.com/openssl/openssl/issues/11820

DDvO:
> > @slontis, where can I find this?
> 
> It looks like you force pushed before pulling. I thought that might happen :).

Oh, indeed - sorry for that.
I rebased on upstream/master to merge `apps/cmp.c` and had overlooked your push to commit 7856241 and had expected that you were at first working on a different branch: `cmp_test_fix` but meanwhile I found that that one is obsolete.

slontis:
> Oh, indeed - sorry for that.

No problems. I did exactly the same thing, when I tried to work on something with Richard :)

DDvO:
> I think I might just give you the patch.

Thanks. 
I wonder why I cannot simply get your commit 785624168c69f7d6f34bfa30a4c8af5fe10970ab - 
I fetched from all potential repos but still get "bad object" although GitHub obviously still has it.

slontis:
Not sure - might be that force pushing messes with it.

slontis:
I didnt want to try to push and then blat your changes also :)

slontis:
I am trying to do the same kind of changes to cms/pkcs7 at the moment.

DDvO:
> I didnt want to try to push and then blat your changes also :)

:)

> Let me know if the patch is no good.

Including the patch went fine, thanks.

I just was fighting other small nits in auto-generated err files and in doc files that made Travis fail. I've just pushed; hope these are gone now.

I'm still chasing the cause of a test failure test_ssl_old that seems to be due to my commit fc1423d9b526c6bf4a95253981da93ec0b72e7e1 ..

DDvO:
@slontis, I've just force-pushed again to clean up the branch and restore your original author and date entries for
```
[PATCH] Add libctx/provider support to cmp_server_test
```

crypto/crmf/crmf_pbm.c
DDvO: I've been able to solve this; `test_cmp_server` now passes.

DDvO:
> I'm still chasing the cause of a test failure test_ssl_old that seems to be due to my commit [fc1423d](https://github.com/openssl/openssl/commit/fc1423d9b526c6bf4a95253981da93ec0b72e7e1)..

Meanwhile this disappeared - not sure why.

DDvO:
@mattcaswell, in order to make several Travis CI runs happy I had to do insert some ugly guards in the new `evp_md_ctx_new_ex()` in `crypto/evp/digest.c`:
```
#ifndef OPENSSL_NO_SM2
# ifndef FIPS_MODULE /* TODO(3.0): EVP_PKEY_CTX_ctrl() not yet supported in FIPS module */
#  ifndef OPENSSL_NO_EC /* TODO why is EVP_PKEY_CTX_set1_id only available then? */
    if (id != NULL && EVP_PKEY_CTX_set1_id(pctx, id->data, id->length) <= 0) {
        X509err(0, ERR_R_MALLOC_FAILURE);
        goto err;
    }
#  endif
# endif
#endif

#ifndef FIPS_MODULE /* TODO(3.0): EVP_DigestSign* not yet supported in FIPS module */
    EVP_MD_CTX_set_pkey_ctx(ctx, pctx);
#else
    EVP_PKEY_CTX_free(pctx);
#endif
```

DDvO:
The Travis failure "No space left on device" has come up again :(

crypto/evp/digest.c
```diff
@@ -70,6 +70,42 @@ int EVP_MD_CTX_reset(EVP_MD_CTX *ctx)
...
+    EVP_MD_CTX *ctx;
+    EVP_PKEY_CTX *pctx = NULL;
+
+    if ((ctx = EVP_MD_CTX_new()) == NULL
+        || (pctx = EVP_PKEY_CTX_new_from_pkey(libctx, pkey, propq)) == NULL) {
+        X509err(0, ERR_R_MALLOC_FAILURE);
+        goto err;
+    }
+
+#ifndef OPENSSL_NO_SM2
```
mattcaswell, line 85: I'm not sure why this is required?

crypto/evp/digest.c
```diff
@@ -70,6 +70,42 @@ int EVP_MD_CTX_reset(EVP_MD_CTX *ctx)
...
+# ifndef FIPS_MODULE /* TODO(3.0): EVP_PKEY_CTX_ctrl() not yet supported in FIPS module */
+#  ifndef OPENSSL_NO_EC /* TODO why is EVP_PKEY_CTX_set1_id only available then? */
```
mattcaswell, line 87: `EVP_PKEY_CTX_set1_id` is defined in the ec.h header file. It seems misplaced there to me. It's an EVP function and therefore presumably should be in evp.h. I guess the reason was that there are a few other EVP_* macros in that header which are EC specific and the SM2 authors just followed that practice. To my eyes all of those macros are misplaced there.

Currently EVP_PKEY_CTX_set1_id is only useful for SM2 - which is based on EC so I can see why it might have ended up there...but still it seems wrong. Obviously because this is in ec.h it inherits the "no-ec" aspect of that. That would go away if this was instead defined in evp.h.....but not necessarily something we need to sort out as part of this PR.

crypto/evp/digest.c
```diff
@@ -70,6 +70,42 @@ int EVP_MD_CTX_reset(EVP_MD_CTX *ctx)
...
 
+EVP_MD_CTX *evp_md_ctx_new_ex(EVP_PKEY *pkey, const ASN1_OCTET_STRING *id,
+                              OPENSSL_CTX *libctx, const char *propq)
+{
+    EVP_MD_CTX *ctx;
+    EVP_PKEY_CTX *pctx = NULL;
+
+    if ((ctx = EVP_MD_CTX_new()) == NULL
+        || (pctx = EVP_PKEY_CTX_new_from_pkey(libctx, pkey, propq)) == NULL) {
+        X509err(0, ERR_R_MALLOC_FAILURE);
```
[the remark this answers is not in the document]
DDvO: One of my favorite copy&paste glitches when moving around function bodies. Fixed.

crypto/evp/digest.c
```diff
@@ -70,6 +70,42 @@ int EVP_MD_CTX_reset(EVP_MD_CTX *ctx)
...
+        goto err;
+    }
+
+#ifndef OPENSSL_NO_SM2
+# ifndef FIPS_MODULE /* TODO(3.0): EVP_PKEY_CTX_ctrl() not yet supported in FIPS module */
```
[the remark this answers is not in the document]
DDvO: `evp_md_ctx_new_ex` can't be made static because it's also used by a_verify.c.
But good thought to move it to `crypto/asn/` - it now lives in a_verify.c,
and its declaration is now in include/crypto/asn1.h

crypto/evp/digest.c
DDvO: I inferred this from a failed Travis run, but likely this is not needed anymore, see below...

crypto/evp/digest.c
DDvO: Thanks for your thoughts and explanation.
Also the documentation suggests that  `#define EVP_PKEY_CTX_set1_id()` and its two siblings should be in evp.h - I've simply moved them there.
Looks like all those ugly guards are not needed any more :)

DDvO: There was still a build issue with no-sm2, which seems to imply no-ec.
Upon a closer look I found that the `#define`s for EVP_PKEY_CTRL_SET1_ID and EVP_PKEY_CTRL_GET1_ID{,_LEN} were wrong and misplaced - fixed in commit c0646ecad65f38c1f4e6a086f8ec92dc9a5834e5

kaduk:
> I wonder why I cannot simply get your commit [7856241](https://github.com/openssl/openssl/commit/785624168c69f7d6f34bfa30a4c8af5fe10970ab) -
> I fetched from all potential repos but still get "bad object" although GitHub obviously still has it.

My understanding is that the git protocol requires some kind of name in order to fetch an object and won't let you fetch an object directly by hash.  Github, on the other hand, does expose some APIs for viewing by hash.
So you could have shane to a `git branch -f purplemonkeydishwasher 785624168c69f7d6f34bfa30a4` and push that somewhere and fetch by that name (or make a tag or whatever), the hash itself is in some sense inside the abstraction barrier of his local repo and github's local repo.

crypto/asn1/a_verify.c
```diff
@@ -21,6 +21,32 @@
...
 
+EVP_MD_CTX *evp_md_ctx_new_ex(EVP_PKEY *pkey, const ASN1_OCTET_STRING *id,
+                              OPENSSL_CTX *libctx, const char *propq)
+{
+    EVP_MD_CTX *ctx;
+    EVP_PKEY_CTX *pctx = NULL;
+
+    if ((ctx = EVP_MD_CTX_new()) == NULL
+        || (pctx = EVP_PKEY_CTX_new_from_pkey(libctx, pkey, propq)) == NULL) {
+        ASN1err(0, ERR_R_MALLOC_FAILURE);
```
slontis, line 32: I wonder if this function should be somewhere else.. It doesnt really do much with ASN1. It just gets data and a length.

crypto/cmp/cmp_ctx.c
```diff
@@ -88,16 +88,38 @@ int OSSL_CMP_CTX_set1_untrusted_certs(OSSL_CMP_CTX *ctx, STACK_OF(X509) *certs)
     return 0;
 }
 
+static int cmp_ctx_set_md(OSSL_CMP_CTX *ctx, EVP_MD **pmd, int nid)
+{
+    const EVP_MD *const_md = EVP_get_digestbynid(nid);
```
slontis, line 93: Not sure you need to do the fetch this early - you could just store the const md
and then fetch when you use it. For the CMS code - I tried to fetch just before the Init operation and then free it straight after.

crypto/cmp/cmp_util.c
```diff
@@ -260,27 +260,29 @@ int ossl_cmp_X509_STORE_add1_certs(X509_STORE *store, STACK_OF(X509) *certs,
 }
 
 /*-
- * Builds up the certificate chain of certs as high up as possible using
- * the given list of certs containing all possible intermediate certificates and
- * optionally the (possible) trust anchor(s). See also ssl_add_cert_chain().
+ * Builds up the chain of intermediate CA certificates
+ * starting from of the given certificate <cert> as high up as possible using
```
slontis, line 264: from of ?

slontis:
Could you apply similiar changes to the other tests as were done in server tests to add the libctx.
The code should be identical and may pick up more instances of the libctx not being used.
Or do you think the server test will cover all cases?

crypto/cmp/cmp_hdr.c
```diff
@@ -135,26 +135,16 @@ int ossl_cmp_hdr_update_messageTime(OSSL_CMP_PKIHEADER *hdr)
...
     int res = 0;
 
-    if (src == NULL) { /* generate a random value if src == NULL */
-        if ((bytes = OPENSSL_malloc(len)) == NULL)
-            goto err;
-        if (RAND_bytes(bytes, len) <= 0) {
-            CMPerr(0, CMP_R_FAILURE_OBTAINING_RANDOM);
-            goto err;
-        }
+    if (bytes == NULL || RAND_bytes_ex(ctx->libctx, bytes, len) <= 0)
```
FdaSilvaYY, line 144: Oups: if 'RAND_bytes_ex', the code is no more calling `ossl_cmp_asn1_octet_string_set1_bytes`

crypto/cmp/cmp_hdr.c
```diff
@@ -135,26 +135,16 @@ int ossl_cmp_hdr_update_messageTime(OSSL_CMP_PKIHEADER *hdr)
...
+        CMPerr(0, CMP_R_FAILURE_OBTAINING_RANDOM);
+    else
```
FdaSilvaYY, line 146: So, I think this `else` should be replaced by a `return ...`

slontis:
I just tried adding libctx to cmp_client_test.
It needs to share some code that was written for the CMS PR that handles adding a library context to X509 certificates (since they do a fetch of SHA1 when verifiying). I will start looking at that.

DDvO:
> @DDvO : any chance you can revisit this PR?

Thanks @slontis for the reminder.

I still had the provider support topic in the back of my head,
but recently unfortunately I've been overly busy with other topics,
and from (later) today until the end of next week I'll be travelling on vacation,
where I'll likely only be able to do a few urgent & quick things.

So I could resume with this likely in week 26.

slontis:
Ok thanks for the heads up..

It looks like there are quite a few X509 objects in the CMP ASN1 objects..
I have added a X509_set0_libctx() for the CMS bits - so that after the d2i I then go thru the loaded objects and find all the X509 bits and call the function for each of them..

I started adding a d2i_XX_with_libctx() function to handle this, but that just got really nasty and I ended up with lots of d2i functions(). So the resolve seems to be cleaner..  But it is still a bit of work to get this working.

DDvO:
@slontis, unfortunately I did not find the time this week to work in the provider topic; hopefully next week...

slontis:
> unfortunately I did not find the time this week to work in the provider topic; hopefully next week...

Thats ok.. It may not be worth doing until #12153 is resolved.

slontis:
@DDvO Any chance you can have another look at this now?
Now that PR #12153 has been merged this can proceed..

I started trying to continue this PR but realized it will take me quite a bit of time to get my head around what the code does/what has changed recently/what has been done in this PR already that will have merge conflicts)

The X509 object can now be created with a libctx/propq pair via X509_new_with_libctx().
You will also notice some extra APIs such as OSSL_STORE_open_with_libctx() (so that certs can be loaded with a libctx/propq).

If you are loading ASN1 data for cmp then the verify chain will fail currently when a libctx is used.
In the CMS code I have changed the d2i_CMS_bio() to also do a resolve 
(See https://github.com/openssl/openssl/blob/33232925ced5bc4ff932fdcdba07343287ffc7ae/crypto/cms/cms_io.c)
so that it goes through all the loaded objects certs and sets up the libctx/propq for each one.


NOTE: OPENSSL_CTX_set0_default() could be used but I would prefer not to propagate this method if we can avoid it.

DDvO:
> @DDvO Any chance you can have another look at this now?

I'm still on vacation this week, so unfortunately have no time to do anything larger by end-July.

slontis:
@DDvO  - any chance you can have a look this week?
We are about a week away from beta, so the API's may be locked down at that point for 3.0.
It would be good to get any libctx's required in before this date for this new CMP interface.

DDvO:
> @DDvO - any chance you can have a look this week?

Yes! Last week I was busy catching up with other things after my vacation,
so I was about to let you know that I can restart tackling this today.

> We are about a week away from beta, so the API's may be locked down at that point for 3.0.

Good to know...

> It would be good to get any libctx's required in before this date for this new CMP interface.

Indeed.

crypto/asn1/a_verify.c
DDvO: Good point - so I've moved the function from `crypto/asn1/a_verify.c` to `crypto/evp/digest.c`.

crypto/cmp/cmp_ctx.c
DDvO: I don't remember why I switched to doing the MD object fetching in advance.
At least it has the advantage that an unsupported or wrong NID is caught early.

crypto/cmp/cmp_util.c
DDvO: Fixed typo.

crypto/cmp/cmp_hdr.c
DDvO: Thanks @FdaSilvaYY for your two comments.
The change in the semantics of the function was on purpose.
This is also reflected to some extent by the changed function name.

crypto/cmp/cmp_hdr.c
DDvO: See above.

DDvO:
@slontis, phew, today I've managed at least to rebase this PR and
(hopefully) fix a number of non-trivial merge conflicts etc.
Hope I did not introduce errors this way.
I've also handled the review comments you gave in May.

DDvO:
A couple of CMP tests currently fail; I did not have the time today to analyze why,
in particular whether these are new issues somehow related to the rebase.
@slontis, if you have any hints how to solve them please let me know.
I'm going to continue with this PR, trying to extend the changes to the CMP client, tomorrow.

slontis:
> A couple of CMP tests currently fail; I did not have the time today to analyze why,
> in particular whether these are new issues somehow related to the rebase.

I tried to rebase it before and got lot in the changes - so I know what you mean :)
It looks like the fipsinstall is failing for some reason - I will have a look.

crypto/evp/digest.c
```diff
@@ -73,6 +74,34 @@ int EVP_MD_CTX_reset(EVP_MD_CTX *ctx)
     return 1;
 }
 
+EVP_MD_CTX *evp_md_ctx_new_ex(EVP_PKEY *pkey, const ASN1_OCTET_STRING *id,
```
slontis, line 77: It appears the only place this is called is via asn1 API's. These sit outside the providers - so this function probably needs a #ifndef FIPS_MODULE around it..

The error you are currently seeing is because EVP_MD_CTX_set_pkey_ctx is not inside the fips module. So putting this function inside the provider means it also tries to find that method inside the provider and fails.

test/recipes/65-test_cmp_server.t
```diff
@@ -9,18 +9,52 @@
...
+ok(run(test(["cmp_server_test",
+             data_file("CR_protected_PBM_1234.der"),
+             "default",
+             srctop_file("test", "default.cnf")])));
+
+unless ($no_fips) {
+    ok(run(app(['openssl', 'fipsinstall',
+                '-out', bldtop_file('providers', 'fipsmodule.cnf'),
+                '-module', bldtop_file('providers', platform->dso('fips')),
+                '-provider_name', 'fips', '-mac_name', 'HMAC',
```
slontis, line 51: A lot of these parameters are set by default now.. (The hexkey has a different default than 00 so this will cause a error).
The fix is to simplify this to

```
    ok(run(app(['openssl', 'fipsinstall',
                '-out', bldtop_file('providers', 'fipsmodule.cnf'),
                '-module', bldtop_file('providers', platform->dso('fips'))])),
       "fipsinstall");
```

doc/man7/property.pod
```diff
@@ -42,9 +42,9 @@ property names like
 A I<property> is a I<name=value> pair.
 A I<property definition> is a sequence of comma separated properties.
 There can be any number of properties in a definition.
-For example: "" defines a null property definition; "my.foo=bar"
-defines a property named I<my.foo> which has a string value I<bar> and
-"iteration.count=3" defines a property named I<iteration.count> which
+For example: "" defines an empty property definition (i.e., no restriction);
```
slontis, line 45: Did you change this??

include/openssl/asn1.h
```diff
@@ -678,6 +678,42 @@ void *ASN1_dup(i2d_of_void *i2d, d2i_of_void *d2i, const void *x);
...
+                                 unsigned char *md, unsigned int *len))
+
+DEPRECATEDIN_3_0(int ASN1_sign(i2d_of_void *i2d, X509_ALGOR *algor1,
+                               X509_ALGOR *algor2, ASN1_BIT_STRING *signature,
+                               char *data, EVP_PKEY *pkey, const EVP_MD *type))
+
+int ASN1_item_sign(const ASN1_ITEM *it, X509_ALGOR *algor1, X509_ALGOR *algor2,
+                   ASN1_BIT_STRING *signature, const void *data,
+                   EVP_PKEY *pkey, const EVP_MD *md);
+int ASN1_item_sign_ex(const ASN1_ITEM *it, X509_ALGOR *algor1,
```
slontis, line 699: Anywhere where we have a libctx added use 'with_libctx' e.g. 
ASN1_item_sign_with_libctx() and
ASN1_item_verify_with_libctx()

Also can we really move these form another public header... That feels like a braking change?

crypto/evp/digest.c
DDvO: I see - thanks for the explanation!
Fixed.

test/recipes/65-test_cmp_server.t
DDvO: Very good hint - done.

doc/man7/property.pod
DDvO: Yes - see commit b3e63f1eb512f34f3984769754a5cdcd8b7da40c with the message
```
Re-word null->empty property; improve iteration.count example in property.pod
```
In case the change is wrong, how best to fix it?
My understanding of the whole provider mechanism including properties etc. is still pretty rudimentary.

include/openssl/asn1.h
DDvO: > Anywhere where we have a libctx added use 'with_libctx' e.g.
> ASN1_item_sign_with_libctx() and
> ASN1_item_verify_with_libctx()

Done this rename.

DDvO: > Also can we really move these from another public header... That feels like a breaking change?

Well, could be, so I've reverted this cleanup.

DDvO:
Now there seems to be just one test issue left:
```
apps/openssl cmp -config ../Mock/test.cnf -section 'Mock credentials' -proxy '' -no_proxy 127.0.0.1 -cert signer.crt -key signer.p12 -keypass 'pass:12345' -digest md2
```
succeeds although md2 should be disabled because it is legacy, right?

slontis:
> succeeds although md2 should be disabled because it is legacy, right?

you might be able to use -provider default -provider legacy to fix this

DDvO:
> > succeeds although md2 should be disabled because it is legacy, right?
> 
> you might be able to use -provider default -provider legacy to fix this

Thanks for the hint, but it had no effect - the following still tries to reach the server rather than failing:
```
cd test/recipes/81-test_cmp_cli_data/Mock
../../../../util/wrap.pl ../../../../apps/openssl cmp -config ../Mock/test.cnf -section 'Mock credentials' -proxy '' -no_proxy 127.0.0.1 -cert signer.crt -key signer.p12 -keypass 'pass:12345' -digest md2 -provider default -provider legacy
```

DDvO:
> > > succeeds although md2 should be disabled because it is legacy, right?
> > 
> > 
> > you might be able to use -provider default -provider legacy to fix this
> 
> Thanks for the hint, but it had no effect - the following still tries to reach the server rather than failing:
> 
> ```
> cd test/recipes/81-test_cmp_cli_data/Mock
> ../../../../util/wrap.pl ../../../../apps/openssl cmp -config ../Mock/test.cnf -section 'Mock credentials' -proxy '' -no_proxy 127.0.0.1 -cert signer.crt -key signer.p12 -keypass 'pass:12345' -digest md2 -provider default -provider legacy
> ```

I suspect that this is because `apps/cmp.c` does not make sufficient use of the libctx.
Adding `app_get0_libctx()` and `app_get0_propq()`:
```
OSSL_CMP_CTX_new(app_get0_libctx(), app_get0_propq())
```
appears necessary but still was not sufficient.

slontis:
> appears necessary but still was not sufficient.

Have you set the environment variable also ?

slontis:
The way I was finding issues was by putting a breakpoint inside evp_generic_fetch() line 313..
And then going back up the callstack to see what got called.

slontis:
If you are creating certs in the app OR loading them as a embedded blob in another ASN1 object then the X509 objects need to be set up with a libctx.
Use X509_new_with_libctx() for any that you create instead of X509_new() and for the blobs loaded with i2d find the X509 objects and set the libctx.

slontis:
(FYI. The initial beta date was 2 weeks after the last alpha so I think that is at the end of next week, not this week - and that date has not been agreed on yet by the OMC - so you have a bit more time than I thought).

test/testutil/provider.c
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright 2018-2020 The OpenSSL Project Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License 2.0 (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
+ */
+
+#include "../testutil.h"
```
slontis, line 10: you might want to see what app_create_libctx() does instead..
this gets enabled if export OPENSSL_TEST_LIBCTX="1"

I have used -config in other apps combined with -provider.

DDvO:
Thanks @slontis for your various hints, which were very helpful moving on.
This way I've been able to add libctx/provider support to cmp_client_test.

DDvO:
Additions to the other tests are still to be done,
and I'm not yet sure about the use of `OPENSSL_TEST_LIBCTX="1"` for the CLI.

Still I suppose that the bulk of the provider support for CMP is in place meanwhile
so would be nice if the current state can be reviewed already now and hopefully merged soon.

Mr1262:
slontis:cmp_provider_support

kaduk:
Hmm, an organization admin should probably block the spam.  (See also the openssl-book repo.)

mattcaswell:
> Hmm, an organization admin should probably block the spam. (See also the openssl-book repo.)

Done

DDvO:
> Now there seems to be just one test issue left:
> 
> ```
> apps/openssl cmp -config ../Mock/test.cnf -section 'Mock credentials' -proxy '' -no_proxy 127.0.0.1 -cert signer.crt -key signer.p12 -keypass 'pass:12345' -digest md2
> ```
> 
> succeeds although md2 should be disabled because it is legacy, right?

I've meanwhile found that this was simply because after changing the CMP_CTX fields from holding digest NIDs
to holding `EVP_MD` references I forgot to check in `apps/cmp.c` the result of calling
`OSSL_CMP_CTX_set_option(ctx, OSSL_CMP_OPT_DIGEST_ALGNID, digest)`, which may now fail. 
Fixed.

crypto/x509/x_x509.c
```diff
@@ -152,13 +152,18 @@ X509 *X509_new_with_libctx(OPENSSL_CTX *libctx, const char *propq)
...
     cert = (X509 *)ASN1_item_new((X509_it()));
-    if (cert != NULL) {
-        cert->libctx = libctx;
-        cert->propq = propq;
-    }
+    (void)x509_set0_libctx(cert, libctx, propq);
     return cert;
 }
 
+X509 *x509_dup_with_libctx(const X509 *cert)
```
mattcaswell, line 159: Seems strange to have a "with_libctx" function that doesn't take a libctx as a parameter. Shouldn't this just be "X509_dup" - and that function should do the right thing with the libctx in the src X509 if it is present?

doc/man3/OSSL_CRMF_MSG_get0_tmpl.pod
```diff
@@ -26,7 +26,7 @@ OSSL_CRMF_MSG_get_certReqId
 
  X509
  *OSSL_CRMF_ENCRYPTEDVALUE_get1_encCert(const OSSL_CRMF_ENCRYPTEDVALUE *ecert,
-                                        EVP_PKEY *pkey);
+                                        OPENSSL_CTX *libctx, EVP_PKEY *pkey);
```
mattcaswell, line 29: Strange to have no propq

test/cmp_client_test.c
```diff
@@ -321,7 +324,8 @@ static int test_exchange_certConf(void)
 {
     SETUP_TEST_FIXTURE(CMP_SES_TEST_FIXTURE, set_up);
     fixture->expected = 0; /* client should not send certConf immediately */
-    if (!ossl_cmp_ctx_set0_newCert(fixture->cmp_ctx, X509_dup(client_cert))) {
+    if (!ossl_cmp_ctx_set0_newCert(fixture->cmp_ctx,
+                                   x509_dup_with_libctx(client_cert))) {
```
mattcaswell, line 328: As mentioned above - I think X509_dup should just do the right thing.

test/recipes/65-test_cmp_client.t
```diff
@@ -9,19 +9,43 @@
...
+}
+
+use lib srctop_dir('Configurations');
+use lib bldtop_dir('.');
+use platform;
+
+my $no_fips = disabled('fips') || ($ENV{NO_FIPS} // 0);
+
+$ENV{OPENSSL_MODULES} = bldtop_dir("providers");
+$ENV{OPENSSL_CONF_INCLUDE} = bldtop_dir("providers");
```
mattcaswell, line 26: This should not be necessary. These are already set by test/run_tests.pl

test/recipes/65-test_cmp_server.t
```diff
@@ -9,18 +9,41 @@
...
+}
+
+use lib srctop_dir('Configurations');
+use lib bldtop_dir('.');
+use platform;
+
+my $no_fips = disabled('fips') || ($ENV{NO_FIPS} // 0);
+
+$ENV{OPENSSL_MODULES} = bldtop_dir("providers");
+$ENV{OPENSSL_CONF_INCLUDE} = bldtop_dir("providers");
```
mattcaswell, line 26: Not necessary

include/openssl/crmf.h
```diff
@@ -67,9 +67,11 @@ DECLARE_ASN1_FUNCTIONS(OSSL_CRMF_MSGS)
 typedef struct ossl_crmf_optionalvalidity_st OSSL_CRMF_OPTIONALVALIDITY;
 
 /* crmf_pbm.c */
-OSSL_CRMF_PBMPARAMETER *OSSL_CRMF_pbmp_new(size_t slen, int owfnid,
-                                           int itercnt, int macnid);
-int OSSL_CRMF_pbm_new(const OSSL_CRMF_PBMPARAMETER *pbmp,
+OSSL_CRMF_PBMPARAMETER *OSSL_CRMF_pbmp_new(OPENSSL_CTX *ctx, size_t slen,
```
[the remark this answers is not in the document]
DDvO: We only need the `libctx` here because no real crypto alg is used but just `RAND_bytes_ex()`, which does not use `propq`.

include/openssl/ec.h
```diff
@@ -1500,9 +1500,6 @@ int EVP_PKEY_CTX_get0_ecdh_kdf_ukm(EVP_PKEY_CTX *ctx, unsigned char **ukm);
 #  define EVP_PKEY_CTRL_GET_EC_KDF_OUTLEN             (EVP_PKEY_ALG_CTRL + 8)
 #  define EVP_PKEY_CTRL_EC_KDF_UKM                    (EVP_PKEY_ALG_CTRL + 9)
 #  define EVP_PKEY_CTRL_GET_EC_KDF_UKM                (EVP_PKEY_ALG_CTRL + 10)
-#  define EVP_PKEY_CTRL_SET1_ID                       (EVP_PKEY_ALG_CTRL + 11)
-#  define EVP_PKEY_CTRL_GET1_ID                       (EVP_PKEY_ALG_CTRL + 12)
-#  define EVP_PKEY_CTRL_GET1_ID_LEN                   (EVP_PKEY_ALG_CTRL + 13)
```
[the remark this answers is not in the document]
DDvO: It may be considered a breaking change since the (corrected) `#define` statements are moved to a different public header.
To be on the safe side for now I reverted the move (and added a TODO as reminder to do the move later).

crypto/cmp/cmp_msg.c
```diff
@@ -994,6 +996,8 @@ X509 *ossl_cmp_certresponse_get1_certificate(EVP_PKEY *privkey,
     }
     if (crt == NULL)
         CMPerr(0, CMP_R_CERTIFICATE_NOT_FOUND);
+    else
+        (void)x509_set0_libctx(crt, ctx->libctx, ctx->propq);
```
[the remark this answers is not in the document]
DDvO: In general I'd agree that it should be done as early as possible,
but in this case it's simpler to do it once at the given place 
because otherwise it would be needed not only in the `OSSL_CMP_CERTORENCCERT_ENCRYPTEDCERT` case but also in the OSSL_CMP_CERTORENCCERT_CERTIFICATE case (where `X509_dup(coec->value.certificate)` is called).

crypto/cmp/cmp_protect.c
```diff
@@ -181,24 +177,24 @@ int ossl_cmp_msg_add_extraCerts(OSSL_CMP_CTX *ctx, OSSL_CMP_MSG *msg)
...
     ASN1_STRING *pbm_str = NULL;
 
     if (!ossl_assert(ctx != NULL))
-        return NULL;
+        return 0;
 
-    alg = X509_ALGOR_new();
-    pbm = OSSL_CRMF_pbmp_new(ctx->pbm_slen, ctx->pbm_owf, ctx->pbm_itercnt,
+    if (*alg == NULL)
+        *alg = X509_ALGOR_new();
```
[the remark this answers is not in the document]
DDvO: This indeed seems strange, though in the given context it does not hurt because `alg` points to a member of a larger structure which will eventually be freed in any case.
Nevertheless I've moved the allocation of `*alg` (in case it was NULL before) to a later point such that on failure of the function there is no need to free `*alg` (in case it was NULL before).

crypto/cmp/cmp_protect.c
```diff
@@ -208,17 +204,45 @@ static X509_ALGOR *create_pbmac_algor(OSSL_CMP_CTX *ctx)
...
-    return NULL;
+    return 0;
+}
+
+static int set_sig_algor(const OSSL_CMP_CTX *ctx, X509_ALGOR **alg)
+{
+    int nid = 0;
+    ASN1_OBJECT *algo = NULL;
+
+    if (*alg == NULL && (*alg = X509_ALGOR_new()) == NULL)
```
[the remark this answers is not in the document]
DDvO: Yes; solution is as above.

crypto/x509/x_x509.c
DDvO: Really strange, but at least while extending the CMP msg and client tests with provider support it looked like `X509_dup()` does not copy the `libctx` and `propq` fields of `x509_st` and this appeared to be the reason why the extended tests. I cannot find the actual implementation of that function, likely because it is implemented purely generically as for many other ASN.1-based structures, but I suspect that non-ASN.1-fields are not actually copied.
Still i tried again without the new function  `X509_dup()_with_libctx()`, and for whatever reason it is not actually needed any more. So I removed it again.

doc/man3/OSSL_CRMF_MSG_get0_tmpl.pod
DDvO: Good spotting - it turns out that the `propq` parameter was not missing in the code but only in its doc - fixed.

test/cmp_client_test.c
DDvO: Yep; removed.

test/recipes/65-test_cmp_client.t
DDvO: Yes - maybe it was needed initially when @slontis started this PR, but indeed it is not needed now, so I removed it.

test/recipes/65-test_cmp_server.t
DDvO: Yes - removed as well.

DDvO:
Thanks @mattcaswell for your new comments of Friday!
I've meanwhile handled them.

crypto/asn1/a_sign.c
```diff
@@ -115,40 +115,50 @@ int ASN1_sign(i2d_of_void *i2d, X509_ALGOR *algor1, X509_ALGOR *algor2,
...
+}
+
+int ASN1_item_sign_with_libctx(const ASN1_ITEM *it, X509_ALGOR *algor1,
+                               X509_ALGOR *algor2, ASN1_BIT_STRING *signature,
+                               const void *data, const ASN1_OCTET_STRING *id,
+                               EVP_PKEY *pkey, const EVP_MD *md,
+                               OPENSSL_CTX *libctx, const char *propq)
+{
+    int rv = 0;
+    EVP_MD_CTX *ctx = evp_md_ctx_new_ex(pkey, id, libctx, propq);
```
slontis, line 133: with_libctx instead of ex?

crypto/asn1/a_sign.c
```diff
@@ -115,40 +115,50 @@ int ASN1_sign(i2d_of_void *i2d, X509_ALGOR *algor1, X509_ALGOR *algor2,
...
 
     if (ctx == NULL) {
-        ASN1err(ASN1_F_ASN1_ITEM_SIGN, ERR_R_MALLOC_FAILURE);
-        return 0;
-    }
-    if (!EVP_DigestSignInit(ctx, NULL, type, NULL, pkey)) {
-        EVP_MD_CTX_free(ctx);
+        ASN1err(ASN1_F_ASN1_ITEM_SIGN_WITH_LIBCTX, ERR_R_MALLOC_FAILURE);
```
slontis, line 136: Replace ASN1_F_ASN1_ITEM_SIGN_WITH_LIBCTX with 0. They are not used anymore so we should not be creating new ones.

crypto/cmp/cmp_vfy.c
```diff
@@ -24,29 +24,23 @@
...
-    size_t prot_part_der_len = 0;
-    unsigned char *prot_part_der = NULL;
     BIO *bio = BIO_new(BIO_s_mem()); /* may be NULL */
     int res = 0;
 
     if (!ossl_assert(cmp_ctx != NULL && msg != NULL && cert != NULL))
         return 0;
 
+    /* prevent misleading error later in case x509v3_cache_extensions() fails */
+    if (!x509v3_cache_extensions(cert)) {
```
slontis, line 40: Hmm - this should be being done when the cert is resolved.. I saw another one of these above also.

doc/internal/man3/ossl_cmp_msg_protect.pod
```diff
@@ -2,19 +2,44 @@
...
+                                STACK_OF(X509) *certs, X509 *cert);
+ ASN1_BIT_STRING *ossl_cmp_calc_protection(const OSSL_CMP_CTX *ctx,
+                                           const OSSL_CMP_MSG *msg);
+ int ossl_cmp_msg_protect(OSSL_CMP_CTX *ctx, OSSL_CMP_MSG *msg);
+ int ossl_cmp_msg_add_extraCerts(OSSL_CMP_CTX *ctx, OSSL_CMP_MSG *msg);
 
 =head1 DESCRIPTION
 
+ossl_cmp_build_cert_chain() builds up the chain of intermediate CA certificates
+starting from of the given certificate B<cert> as high up as possible using
```
slontis, line 26: remove of

doc/internal/man3/ossl_cmp_pkisi_get_status.pod
```diff
@@ -50,6 +53,10 @@ ossl_cmp_pkisi_check_pkifailureinfo
 
 =head1 DESCRIPTION
 
+ossl_cmp_certresponse_get1_cert() returns a pointer to a copy of the newly
+enrolled certificate from the given certResponse B<crep>, or NULL on error.
```
slontis, line 57: These should be I<>?

doc/man3/OSSL_CRMF_MSG_get0_tmpl.pod
```diff
@@ -48,8 +49,9 @@ OSSL_CRMF_CERTID_get0_issuer retrieves the issuer name
 of the given CertId B<cid>, which must be of ASN.1 type GEN_DIRNAME.
 
 OSSL_CRMF_ENCRYPTEDVALUE_get1_encCert() decrypts the certificate in the given
-encryptedValue B<ecert>, using the private key B<pkey>.
-This is needed for the indirect PoP method as in RFC 4210 section 5.2.8.2.
+encryptedValue B<ecert>, using the private key B<pkey> and library context
```
slontis, line 52: , instead of and

doc/man3/OSSL_CRMF_MSG_set0_validity.pod
```diff
@@ -59,11 +61,13 @@ OSSL_CRMF_MSG_push0_extension() pushes the X509 extension I<ext> to the
 extensions in the certTemplate of I<crm>.  Consumes I<ext>.
 
 OSSL_CRMF_MSG_create_popo() creates and sets the Proof-of-Possession (POPO)
-according to the method I<ppmtd> in I<crm>.
+according to the method I<meth> in I<crm>.
+The library context I<libctx> and property query string I<propq>,
+may be NULL to select the defaults.
 In case the method is OSSL_CRMF_POPO_SIGNATURE the POPO is calculated
-using the private I<pkey> and the digest algorithm NID I<dgst>.
+using the private B<pkey> and the digest algorithm I<digest>.
```
slontis, line 68: I<pkey>

doc/man3/OSSL_CRMF_pbmp_new.pod
```diff
@@ -8,39 +8,49 @@ OSSL_CRMF_pbmp_new
...
-  OSSL_CRMF_PBMPARAMETER *OSSL_CRMF_pbmp_new(size_t saltlen, int owfnid,
-                                             int itercnt, int macnid);
+ OSSL_CRMF_PBMPARAMETER *OSSL_CRMF_pbmp_new(OPENSSL_CTX *libctx, size_t saltlen,
+                                            int owfnid, size_t itercnt,
+                                            int macnid);
 
 =head1 DESCRIPTION
 
 OSSL_CRMF_pbm_new() generates a PBM (Password-Based MAC) based on given PBM
 parameters B<pbmp>, message B<msg>, and secret B<sec>, along with the respective
```
slontis, line 26: I<>

doc/man3/OSSL_CRMF_pbmp_new.pod
```diff
@@ -8,39 +8,49 @@ OSSL_CRMF_pbmp_new
...
 
-The iteration count must be at least 100, as stipulated by RFC 4211, and is
-limited to at most 100000 to avoid DoS through manipulated or otherwise
-malformed input.
-
-OSSL_CRMF_pbmp_new() initializes and returns a new PBMParameter
-structure with a new random salt of given length B<saltlen>, OWF (one-way
-function) NID B<owfnid>, iteration count B<itercnt>, and MAC NID B<macnid>.
+OSSL_CRMF_pbmp_new() initializes and returns a new B<PBMParameter> structure
+with a new random salt of given length B<saltlen>,
```
slontis, line 36: I<> there are a few of them here.

doc/man7/property.pod
```diff
@@ -60,7 +60,7 @@ provider defines I<provider=legacy> for all of their algorithms.
 =head2 Queries
 
 A I<property query clause> is a single conditional test.
-For example, "fips=yes", "provider!=default" or "?iteration.count!=3".
+For example, "fips=yes", "provider!=default" or "?iteration.count=3".
```
slontis, line 63: @paulidale can you check this please.

paulidale: It looks okay.

crypto/asn1/a_sign.c
DDvO: Done.

crypto/asn1/a_sign.c
DDvO: Done.

crypto/cmp/cmp_vfy.c
DDvO: > Hmm - this should be being done when the cert is resolved.. I saw another one of these above also.

I'm not sure if I understand what you mean here.
I do these new sort of checks basically as soon as the cert in question is known,
yet in this case it was done a bit later than that and it fits better in `cert_acceptable()`, so I've moved it there.

doc/internal/man3/ossl_cmp_msg_protect.pod
DDvO: Done

doc/internal/man3/ossl_cmp_pkisi_get_status.pod
DDvO: Right, done.

doc/man3/OSSL_CRMF_MSG_get0_tmpl.pod
DDvO: Done

doc/man3/OSSL_CRMF_MSG_set0_validity.pod
DDvO: Done

doc/man3/OSSL_CRMF_pbmp_new.pod
DDvO: Done

doc/man3/OSSL_CRMF_pbmp_new.pod
DDvO: Done

DDvO:
Thanks @slontis for your new comments. I've handled all of them.

> There are still quite a few tests doing OSSL_CMP_CTX_new(NULL, NULL).
> Should they all be using the libctx in case there are any edge cases?

I've just done the exercise to at the libctx also to cmp_{msg,vfy,protect}_test, which did not bring up any new cases.
I don't think it would be worth doing so for cmp_{ctx,hdr} test because they are not about using crypto.

So I'm pretty confident that for the CMP+CRMF implementation we have meanwhile caught all relevant cases.

test/cmp_msg_test.c
```diff
@@ -31,6 +32,18 @@ typedef struct test_fixture {
     OSSL_CMP_PKISI *si;
 } CMP_MSG_TEST_FIXTURE;
 
+static OPENSSL_CTX *libctx = NULL;
+static OSSL_PROVIDER *default_null_provider = NULL, *provider = NULL;
+
+static X509 *X509_dup_with_libctx(const X509 *cert)
+{
+    X509 *dup = X509_dup(cert);
```
slontis, line 40: Shouldnt cert already have  a libctx set?
X509_dup should be copying the libctx, propq if it doesnt already..

test/cmp_msg_test.c
```diff
@@ -283,10 +296,11 @@ static int test_cmp_create_certconf(void)
     fixture->fail_info = 0;
     fixture->expected = 1;
     if (!TEST_true(ossl_cmp_ctx_set0_newCert(fixture->cmp_ctx,
-                                             X509_dup(cert)))) {
+                                             X509_dup_with_libctx(cert)))) {
         tear_down(fixture);
         fixture = NULL;
     }
+    x509_set0_libctx(fixture->cmp_ctx->newCert, libctx, NULL);
```
slontis, line 303: newCert should already be resolved with a libctx here?

test/cmp_testlib.c
```diff
@@ -39,6 +39,7 @@ X509 *load_pem_cert(const char *file)
         (void)TEST_ptr(cert = PEM_read_bio_X509(bio, NULL, NULL, NULL));
```
slontis, line 39: x509_set0_libctx() should only be used as a last resort for resolving..

The better way to do this is 

```
TEST_ptr(cert = X509_new_with_libctx(libctx, propq);
..
PEM_read_bio_X509(bio, &cert, NULL, NULL)
..
```
If you can think of other places where you might of done this it would be good if they were changed..


If you have a look at d2i_X509() you will notice that it calls x509v3_cache_extensions() if the cert is passed in..
This makes it a bit more of a immutable operation.. This was also the reason why I was a bit wary of some of the 
x509v3_cache_extensions() calls that you had added in a few places.

So the rules I used for the CMS changes were..
Use X509_new_with_libctx() everywhere instead of X509_new().
If a ASN1 object got loaded that contains certs.. resolve them straight afterwards..

crypto/cmp/cmp_msg.c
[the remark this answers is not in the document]
slontis: Which means that coec->value.certificate should also already be setup straight after it is loaded also..

crypto/crmf/crmf_lib.c
```diff
@@ -629,7 +627,7 @@ X509 *OSSL_CRMF_ENCRYPTEDVALUE_get1_encCert(const OSSL_CRMF_ENCRYPTEDVALUE *ecer
     }
     cikeysize = EVP_CIPHER_key_length(cipher);
     /* first the symmetric key needs to be decrypted */
-    pkctx = EVP_PKEY_CTX_new(pkey, NULL);
+    pkctx = EVP_PKEY_CTX_new_from_pkey(libctx, pkey, propq);
```
slontis, line 630: I would like to see a X509_new_with_libctx() used here which is then passed to d2i_X509

test/cmp_msg_test.c
DDvO: >  X509_dup should be copying the libctx, propq if it doesnt already..

Yes, this is what one would expect, but as mentioned before in https://github.com/openssl/openssl/pull/11808#discussion_r471092440 this is not really the case, so I added this again as a workaround.
I confirmed this bug by adding `printf` to that function and using this test with and without the workaround.
I've just opened an issue on this bug: #12680.

crypto/crmf/crmf_lib.c
DDvO: > I would like to see a X509_new_with_libctx() used here which is then passed to d2i_X509

I don't get what you mean here.
This change is not about `X509` but `EVP_PKEY_CTX`.
If any change is actually needed here please let me know the concrete code.

test/cmp_testlib.c
DDvO: Thanks for your explanation.
I've changed this as you mentioned, and I did not find any further such occurrences.

I still think the calls of  `x509v3_cache_extensions()` I added at some places are needed there
because otherwise I had situations where I got reported misleading errors.

The CMP+CRMF code does not (directly) use `X509_new()`.

crypto/crmf/crmf_lib.c
slontis: Err I cant mark the line but look at line 686 which should be
```
cert = X509_new_with_libctx(libctx, propq);
if (cert == NULL) {
}
if (d2i_X509(&cert, &p, outlen) == NULL) {
```

slontis: This will do the cache_extensions as I have described above..

test/cmp_msg_test.c
DDvO: Oops, this was a leftover. Removed.

test/cmp_testlib.c
slontis: See below :)

test/cmp_msg_test.c
slontis: I would probably add something like
/* TODO(3.0) Clean this up - See issue #12680 */

DDvO: Done.

crypto/crmf/crmf_lib.c
DDvO: Ah, so your comment was referring not to the right position in the code.
Thanks for the code suggestion; I've used it (in a more compact form).

test/cmp_testlib.c
DDvO: Meanwhile I understand better what you mean with doing any needed`x509v3_cache_extensions()` as early as possible.
I've just removed one of those because I found it's done elsewhere.
The remaining two are used in cases where certs are input by a CMP_CTX_set1 function or by the CMP request/response transfer callback which in general is not under the control of the CMP implementation, neither.

slontis: So the question then becomes - what bit of code is currently loading the offending cert - and make the point where it is loaded where the resolve is done..

slontis:
travis failure is relevant - maybe a no-ec case?

DDvO:
> travis failure is relevant - maybe a no-ec case?

Ah, yes, the recently added test key file contained an EC key - changed that to RSA.

test/cmp_testlib.c
DDvO: > So the question then becomes - what bit of code is currently loading the offending cert - and make the point where it is loaded where the resolve is done..

The `x509v3_cache_extensions()` call in `cmp_ctx()` is right in `OSSL_CMP_CTX_set1_##FIELD` where a cert is being imported by the crypto lib user. In the tests the cert has been loaded using the `PEM_read_bio_X509()` call in `cmp_testlib.c` which we have meanwhile handled. So this should be fine now.

The other remaining `x509v3_cache_extensions()` call is in `cmp_vfy()` where the cert has been determined by a non-trivial process called via `ossl_cmp_msg_check_update()` (and then `OSSL_CMP_validate_msg()`) which typically selects it from the `extraCerts` field in a received CMP message. In the test these are loaded via `load_pkimsg()` in `cmp_testlib.c`. This indirectly calls `d2i_OSSL_CMP_MSG_bio()` where you wrote that this invokes  `x509v3_cache_extensions()` for any certs included.

So I'd say the two remaining explicit `x509v3_cache_extensions()` call in the lib are good as precautions in case a lib user does not properly augment the certs (s)he provides with any needed libctx etc.

DDvO:
@slontis, hope that this PR can finally be approved now, such that it makes it in the beta1 release?

In case any further (subtle) issues come up later we can still handle them when needed.

slontis:
So I approve.. This was my branch (even though you did all the work).. So if you approve also then it is good to go.

slontis (commented):
Approved.

mspncp (changes requested):
Unless you intend to squash everything into a single commit, please clean up the timeline of your 22 commits and group them logically. In particular, the "[PATCH] ..." should be removed from the subject line of commit 103e86de4558da8d54fa9bed3ce1bd7dddba012a and the "...fix omission bug..." commit de55e77f8bf930de639dd89f0166fe932a89ad19 be squashed.

Ideally, you would do an interactive rebase without moving the merge base (`git rebase -i --keep-base master`), because then GitHub's force-push diff would prove that no other changes were made.

_(edit: older git versions might not have the `--keep-base` option yet. In this case, you can use `git rebase -i HEAD~22`)_

DDvO:
> Unless you intend to squash everything into a single commit,

> please clean up the timeline of your 22 commits and group them logically. 

I had already done various squashing before and in my view the remaining commits are already structured reasonably well.
Also the commit messages are IMO to the point.

> In particular, the "[PATCH] ..." should be removed from the subject line of commit [103e86d](https://github.com/openssl/openssl/commit/103e86de4558da8d54fa9bed3ce1bd7dddba012a)

Done.

> and the "...fix omission bug..." commit [de55e77](https://github.com/openssl/openssl/commit/de55e77f8bf930de639dd89f0166fe932a89ad19) be squashed.

Not really, because this fixes an independent (pre-existing) bug I noticed during the preparation of this PR.
I've just removed the word "omission" from the commit message since it may be more confusing than helpful.
 
> Ideally, you would do an interactive rebase without moving the merge base (`git rebase -i --keep-base master`), because then GitHub's force-push diff would prove that no other changes were made.
> 
> _(edit: older git versions might not have the `--keep-base` option yet. In this case, you can use `git rebase -i HEAD~22`)_

Thanks for this hint. Interesting option,but my git version 2.20.1 does not support it.

DDvO (approved):
Re-approving for formal reasons after two minor commit message improvements.

DDvO:
> So I approve.. This was my branch (even though you did all the work).. So if you approve also then it is good to go.

Thanks @slontis!
So unless GitHub really prohibits this, would be nice if you can formally add your approval as well.

DDvO:
If this PR gets merged in about 24 hours from now, is this still in time for the beta release?
BTW, I cannot find the current actual (which differs from the official) plan for the alpha and beta releases.

DDvO:
Or may we do the merge already today, helping #12621?

DDvO:
BTW, @mattcaswell, this PR still contains your stale change request.

mspncp:
> Or may we do the merge already today, helping #12621?

There's no hurry. Current ETA for beta1 is approximately 2020-08-27. Repo freeze may happen a few days earlier, but your pull request seems to be in time to go in before that.

mspncp:
> BTW, @mattcaswell, this PR still contains your stale change request.

Many committers don't care to remove stale change requests after they are adressed. Formally, only the approvals count. But if they disturb them, feel free to dismiss them yourself (with a comment like "change requests addressed"), or alternatively re-request the reviews.

mspncp:
> Re-approving for formal reasons after two minor commit message improvements.

It's ok for me if you count the grace period starting from @slontis' approval in https://github.com/openssl/openssl/pull/11808#pullrequestreview-471297265, which was eight hours ago.

mspncp:
> I've just removed the word "omission" from the commit message since it may be more confusing than helpful.

Are you shure you pushed those changes? I still see the old messages. Anyway, it's ok for me if these two fixes are made while merging.

DDvO:
> > BTW, @mattcaswell, this PR still contains your stale change request.
> 
> Many committers don't care to remove stale change requests after they are adressed. Formally, only the approvals count. But if they disturb ~them~ you, feel free to dismiss them yourself (with a comment like "change requests addressed"), or alternatively re-request the reviews.

Thanks @mspncp for your four helpful responses.

In this case Matt's change request were addressed, which should be documented by my responses to them.
I've just tried to formally dismiss the stale change request using the GitHub buttons next to the review overview, but GitHub claims it cannot find the respective commits, likely because they were outdated by a forced push. 
Good to know that in the end it does not matter as long as approvals are there.

DDvO:
> > I've just removed the word "omission" from the commit message since it may be more confusing than helpful.
> 
> Are you shure you pushed those changes? I still see the old messages. Anyway, it's ok for me if these two fixes are made while merging.

Ah, I did a mistake (not for the first time), namely to edit the commit messages in the overview while interactive rebasing, not actually using the r/reword command. Now it's done (and pushed).

slontis:
> So unless GitHub really prohibits this, would be nice if you can formally add your approval as well.

Because it is my branch I cant do that..

Reconfirmed approval..

DDvO:
Merged - thanks a lot @slontis for your very competent, diligent, and patient shepherding of this involved task!

slontis:
> Merged - thanks a lot @slontis for your very competent, diligent, and patient shepherding of this involved task!

You did all the hard work :) Thanks for doing it..

PR #8965, opened 2019-05-21
Configure: for '-z defs', also check $config{cflags}

levitte:
When sanitize options are added as 'enable-msan' or similar, the
-fsanitize C flags is set in $config{cflags} rather than
$config{CFLAGS}, so we need to check both.

PR #8892, opened 2019-05-07
Don't use '-z defs' if sanitizers are used

levitte:
There are quite a number of sanitizers for clang that aren't
documented in the clang user documentation.  This makes it impossible
to be selective about what sanitizers to look at to determine if
'-z defs' should be used of not.

Under these circumstances, the sane thing to do is to just look for
any sanitizer specification and not use '-z defs' if there's one
present.

Fixes #8735

PR #8749, opened 2019-04-15
Don't use '-z defs' with Clang's sanitizers

levitte:
The clang documentation in all sanitizers we currently use says this:

    When linking shared libraries, the {flavor}Sanitizer run-time is
    not linked, so -Wl,-z,defs may cause link errors (don’t use it
    with {flavor}Sanitizer)

(in our case, {flavor} is one of Address, Memory, or UndefinedBehavior)

Therefore, we turn off that particular flag specifically when using
the sanitizers.

Fixes #8735

PR #11815, opened 2020-05-13
Pass "-z defs" to the linker via "-Wl,-z,defs"

tomiii00:
Older versions of gcc do not support the -z flag. Use -Wl,-z,defs instead.

CLA: trivial

### DIFF
--- a/Configurations/shared-info.pl
+++ b/Configurations/shared-info.pl
@@ -36,7 +36,7 @@ my %shared_info;
                 (grep /(?:^|\s)-fsanitize/,
                  @{$config{CFLAGS}}, @{$config{cflags}})
                 ? ''
-                : '-z defs',
+                : '-Wl,-z,defs',
         };
     },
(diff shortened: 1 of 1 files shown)


mattcaswell:
This PR has the CLA: trivial label. I agree it is trivial.

@levitte - do you also agree?

levitte:
Ah, yes, I agree

mattcaswell:
Pushed to master only.

@tomiii00 - are you able  to create a backport PR for the 1.1.1 branch? This does not cherry-pick cleanly there.

levitte:
I just had a look at a backport, and it's non-trivial.  There's a slew of other PRs that aren't back-ported before we would be able to cherry-pick this one: #8537, #8749, #8892, #8965.

I'm thinking that isn't worth the effort.  We don't *really* need `-z defs` in pre-3.0.

levitte:
I've made a judgement call to remove the 1.1.1 label and leave it at that.

PR #11824, opened 2020-05-14
Build: make apps/progs.c depend on configdata.pm

levitte:
The reason for this change is this, which happened in a build on my machine:

```
../master/apps/cmp.c:3008:5: error: no previous prototype for 'cmp_main' [-Werror=missing-prototypes]
 3008 | int cmp_main(int argc, char **argv)
      |     ^~~~~~~~
```


levitte:
Merged

c6601bd2d728d4c61711a016c6267fb45910e7cd Build: make apps/progs.c depend on configdata.pm

PR #11826, opened 2020-05-14
Fail if we fail to fetch the EVP_KEYMGMT

mattcaswell:
If we failed to fetch an EVP_KEYMGMT then we were falling back to legacy.
This is because some algorithms (such as MACs and KDFs used via an old
style EVP_PKEY) have not been transferred to providers.

Unfortunately this means that you cannot stop some algorithms from being
used by not loading the provider.

For example if you wanted to prevent RSA from being used, you might expect
to just not load any providers that make it available. Unfortunately that
doesn't work because we simply fall back to legacy if we fail to fetch
the EVP_KEYMGMT.

Instead we should fail *unless* the key type is one of those legacy key
types that we have not transferred.

### DIFF
--- a/crypto/evp/pmeth_lib.c
+++ b/crypto/evp/pmeth_lib.c
@@ -137,6 +137,40 @@ EVP_PKEY_METHOD *EVP_PKEY_meth_new(int id, int flags)
 }
 #endif /* FIPS_MODULE */
 
+static int legacyalg(int id, const char *keytype)
+{
+#ifndef FIPS_MODULE
+    /* Certain EVP_PKEY keytypes are only available in legacy form */
+    if (id == -1) {
+        id = OBJ_sn2nid(keytype);
+        if (id == NID_undef)
+            id = OBJ_ln2nid(keytype);
+        if (id == NID_undef)
+            return  0;
+    }
+    switch (id) {
+    /*
+     * TODO(3.0): Remove SM2 and DHX when they are converted to have provider
+     * support
+     */
+    case EVP_PKEY_SM2:
+    case EVP_PKEY_DHX:
+    case EVP_PKEY_SCRYPT:
+    case EVP_PKEY_TLS1_PRF:
+    case EVP_PKEY_HKDF:
+    case EVP_PKEY_CMAC:
+    case EVP_PKEY_HMAC:
+    case EVP_PKEY_SIPHASH:
+    case EVP_PKEY_POLY1305:
+        return 1;
+    default:
+        return 0;
+    }
+#else
+    return 0;
+#endif
+}
+
 static EVP_PKEY_CTX *int_ctx_new(OPENSSL_CTX *libctx,
                                  EVP_PKEY *pkey, ENGINE *e,
                                  const char *keytype, const char *propquery,
@@ -228,10 +262,20 @@ static EVP_PKEY_CTX *int_ctx_new(OPENSSL_CTX *libctx,
      * implementation.
      */
     if (e == NULL && keytype != NULL) {
-        /* This could fail so ignore errors */
-        ERR_set_mark();
+        int legacy = legacyalg(id, keytype);
+
+        if (legacy) {
+            /* This could fail so ignore errors */
+            ERR_set_mark();
+        }
+
         keymgmt = EVP_KEYMGMT_fetch(libctx, keytype, propquery);
-        ERR_pop_to_mark();
+        if (legacy) {
+            ERR_pop_to_mark();
+        } else if (keymgmt == NULL) {
+            EVPerr(EVP_F_INT_CTX_NEW, EVP_R_FETCH_FAILED);
+            return NULL;
+        }
     }
 
     ret = OPENSSL_zalloc(sizeof(*ret));
--- a/crypto/evp/pmeth_lib.c
+++ b/crypto/evp/pmeth_lib.c
@@ -137,7 +137,7 @@ EVP_PKEY_METHOD *EVP_PKEY_meth_new(int id, int flags)
 }
 #endif /* FIPS_MODULE */
 
-static int legacyalg(int id, const char *keytype)
+static int is_legacy_alg(int id, const char *keytype)
 {
 #ifndef FIPS_MODULE
     /* Certain EVP_PKEY keytypes are only available in legacy form */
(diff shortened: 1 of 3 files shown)


crypto/evp/pmeth_lib.c
```diff
@@ -137,6 +137,40 @@ EVP_PKEY_METHOD *EVP_PKEY_meth_new(int id, int flags)
...
+        id = OBJ_sn2nid(keytype);
+        if (id == NID_undef)
+            id = OBJ_ln2nid(keytype);
+        if (id == NID_undef)
+            return  0;
+    }
+    switch (id) {
+    /*
+     * TODO(3.0): Remove SM2 and DHX when they are converted to have provider
+     * support
```
levitte, line 154: I'm surprised to see the KDFs-by-PKEY here...  the *do* have provider support, all of them...  no?

mattcaswell: They have EVP_KDF support - but there is no EVP_KEYMGMT for them. The *legacy* implementation does the mapping to EVP_KDF. The same goes for the EVP_MAC support.

crypto/evp/pmeth_lib.c
```diff
@@ -137,6 +137,40 @@ EVP_PKEY_METHOD *EVP_PKEY_meth_new(int id, int flags)
...
+    case EVP_PKEY_SCRYPT:
+    case EVP_PKEY_TLS1_PRF:
+    case EVP_PKEY_HKDF:
+    case EVP_PKEY_CMAC:
+    case EVP_PKEY_HMAC:
+    case EVP_PKEY_SIPHASH:
+    case EVP_PKEY_POLY1305:
+        return 1;
+    default:
+        return 0;
```
levitte, line 167: Hmmm, so what happens with an EVP_PKEY_METHOD provided by an ENGINE?  "GOST R 34.10-2001" comes to mind...

mattcaswell: That should all "just work". `int_ctx_new` does the work of checking if there is an ENGINE available or not. If there is we never attempt to fetch an EVP_KEYMGMT and don't go down this codepath.

levitte:
I'm not so sure about this solution...  again, it assumes that libcrypto has universal knowledge, which isn't true.

So I wonder if we shouldn't pick on certain markers that tell if an EVP_PKEY_METHOD has a provider replacement, and make the judgement call on that.  The presence of `export_to` and `import_from` would be such a marker, shouldn't they?

mattcaswell:
> I'm not so sure about this solution... again, it assumes that libcrypto has universal knowledge, which isn't true.

Under what circumstances is this *not* true (given that this codepath only occurs after we've already checked for ENGINE implementations)?

> So I wonder if we shouldn't pick on certain markers that tell if an EVP_PKEY_METHOD has a provider replacement, and make the judgement call on that. The presence of export_to and import_from would be such a marker, shouldn't they?

This sounded like a good idea - so I looked into it. Unfortunately it doesn't work. The relevant functions are in `EVP_PKEY_ASN1_METHOD` - which only applies if you actually have a legacy key present. Unfortunately many of these algorithms don't actually use an EVP_PKEY at all...they just use an EVP_PKEY_CTX!! For example the KDFs do a "derive" operation without having an EVP_PKEY in place.

test/evp_extra_test.c
```diff
@@ -1623,6 +1623,47 @@ static int test_keygen_with_empty_template(int n)
     return ret;
 }
 
+/*
+ * Test that we fail if we attempt to use an algorithm that is not available
+ * in the current library context (unless we are using an algorithm that should
+ * be made available via legacy codepaths).
```
levitte, line 1629: So essentially, we're looking for legacy implementations that *should* have a provider implementation.  Would you mind writing the comment in those terms?  I find the current comment harder to grasp.

levitte:
I really thoroughly *dislike* those switch tables of hard-coded IDs, especially when they're coded in terms of "we know that only these are legacy".

What I can figure is that apart from the markers I already mentioned, you can also check if the pmeth matches one of our well known bridges.  An easy way would be like this:

``` C
    if (pmeth->export_to != NULL
        || (pmeth->init == tls1_prf_pkey_method()->init)
        || (pmeth->init == hmac_pkey_method()->init))
        return 0; /* not legacy */
```

(no, I haven't checked the correctness of those symbols, I hope that it's clear enough anyway)
That constitutes what we *actually* know, i.e. things we have very specifically coded.

Regarding what we actually know vs ENGINE, there's also the possibility that the calling application has internal EVP_PKEY_METHODs and EVP_PKEY_ASN1_METHODs.  They don't have an attached ENGINE.  That's why I'm saying that libcrypto cannot have universal knowledge.

mattcaswell:
> What I can figure is that apart from the markers I already mentioned, you can also check if the pmeth matches one of our well known bridges. An easy way would be like this:

What you are asking for is to identify markers that say whether specific ctxs are definitely *not* legacy and always fallback if they are not. None of the markers you have so far mentioned help with that:

- export_to/import_from etc are only present in the ameth. An EVP_PKEY_CTX *does not have an ameth* - it only has a pmeth. It does have an *optional* pkey which may have an ameth - but that doesn't help if you don't have a pkey (which is common).

- Checking (for example) whether the init function is a well known init function only helps you identify whether it is definitely legacy. A failure to match doesn't tell you it definitely isn't legacy - so this doesn't help either.

mattcaswell:
Note #11834 is pending on this PR.

crypto/evp/pmeth_lib.c
```diff
@@ -137,6 +137,40 @@ EVP_PKEY_METHOD *EVP_PKEY_meth_new(int id, int flags)
 }
 #endif /* FIPS_MODULE */
 
+static int legacyalg(int id, const char *keytype)
```
levitte, line 140: Nit: `is_legacy_alg` ?

mattcaswell: Fixed

mattcaswell:
Nit fixed. I also included a fixup to handle a travis failure. Previously we were successfully managing to load an X25519 key in a no-ec build via the legacy path. This is no longer the case so evp_test needed a small tweak.

@levitte - please can you reconfirm?

mattcaswell:
Ping @levitte. Please reconfirm?

mattcaswell:
Pushed. Thanks!

PR #11832, opened 2020-05-14
Delete the sslprovider test

mattcaswell:
This was added before the changes to the sslapi/ssl_new/ssl_old tests which
run those tests with a non-default library context. It no longer adds
anything that those tests don't already do, so it can be deleted.

This also fixes a number of run-checker build failures which were failing
in this test if TLSv1.2 was disabled.


t8m:
Travis failure not relevant.

mattcaswell:
Pushed. Thanks.

PR #11834, opened 2020-05-14
Check Signature Algorithms are available before offering or accepting them

mattcaswell:
Different providers may support different signature and hash schemes, and we cannot guarantee that the default provider is always loaded. Therefore we should check our signature algorithms to ensure that the providers we have loaded do actually support them before we offer or accept them.

For now we "cheat" a little bit. We check that the hash is available, and we check that the signature scheme is available. However we don't check that the combination is available (e.g. imagine a provider that only supports RSA with SHA256, even though other digests are available from other providers). That is quite a bit harder to solve.

This is quite tricky to test because it requires the ability to switch on and off the availability of some algorithms. In order to do this I have implemented a "filter" provider within the test suite that provides this functionality. It essentially sits in front of the default provider and enables you to configure which algorithms are made available. This required the addition of a couple of new `OSSL_PROVIDER_*` functions.

This is marked as WIP because it includes the commits  from #11826.

### DIFF
--- a/crypto/provider.c
+++ b/crypto/provider.c
@@ -57,6 +57,15 @@ int OSSL_PROVIDER_get_params(const OSSL_PROVIDER *prov, OSSL_PARAM params[])
     return ossl_provider_get_params(prov, params);
 }
 
+
+const OSSL_ALGORITHM *OSSL_PROVIDER_query_operation(const OSSL_PROVIDER *prov,
+                                                    int operation_id,
+                                                    int *no_cache)
+{
+    return ossl_provider_query_operation(prov, operation_id, no_cache);
+}
+
+
 int OSSL_PROVIDER_add_builtin(OPENSSL_CTX *libctx, const char *name,
                               OSSL_provider_init_fn *init_fn)
 {
--- a/crypto/provider.c
+++ b/crypto/provider.c
@@ -65,7 +65,7 @@ const OSSL_ALGORITHM *OSSL_PROVIDER_query_operation(const OSSL_PROVIDER *prov,
     return ossl_provider_query_operation(prov, operation_id, no_cache);
 }
 
-void *OSSL_PROVIDER_get_provider_ctx(const OSSL_PROVIDER *prov)
+void *OSSL_PROVIDER_get0_provider_ctx(const OSSL_PROVIDER *prov)
 {
     return ossl_provider_prov_ctx(prov);
 }
--- a/crypto/provider.c
+++ b/crypto/provider.c
@@ -65,6 +65,10 @@ const OSSL_ALGORITHM *OSSL_PROVIDER_query_operation(const OSSL_PROVIDER *prov,
     return ossl_provider_query_operation(prov, operation_id, no_cache);
 }
 
+void *OSSL_PROVIDER_get_provider_ctx(const OSSL_PROVIDER *prov)
+{
+    return ossl_provider_prov_ctx(prov);
+}
 
 int OSSL_PROVIDER_add_builtin(OPENSSL_CTX *libctx, const char *name,
                               OSSL_provider_init_fn *init_fn)
--- a/crypto/provider_core.c
+++ b/crypto/provider_core.c
@@ -761,7 +761,10 @@ const char *ossl_provider_module_path(const OSSL_PROVIDER *prov)
 
 void *ossl_provider_prov_ctx(const OSSL_PROVIDER *prov)
 {
-    return prov->provctx;
+    if (prov != NULL)
+        return prov->provctx;
+
+    return NULL;
 }
 
 OPENSSL_CTX *ossl_provider_library_context(const OSSL_PROVIDER *prov)
--- a/crypto/provider_core.c
+++ b/crypto/provider_core.c
@@ -759,6 +759,11 @@ const char *ossl_provider_module_path(const OSSL_PROVIDER *prov)
 #endif
 }
 
+void *ossl_provider_prov_ctx(const OSSL_PROVIDER *prov)
+{
+    return prov->provctx;
+}
+
 OPENSSL_CTX *ossl_provider_library_context(const OSSL_PROVIDER *prov)
 {
     /* TODO(3.0) just: return prov->libctx; */
--- a/doc/man3/OSSL_PROVIDER.pod
+++ b/doc/man3/OSSL_PROVIDER.pod
@@ -6,8 +6,8 @@ OSSL_PROVIDER_set_default_search_path,
 OSSL_PROVIDER, OSSL_PROVIDER_load, OSSL_PROVIDER_unload,
 OSSL_PROVIDER_available, OSSL_PROVIDER_do_all,
 OSSL_PROVIDER_gettable_params, OSSL_PROVIDER_get_params,
-OSSL_PROVIDER_query_operation, OSSL_PROVIDER_add_builtin,
-OSSL_PROVIDER_name - provider routines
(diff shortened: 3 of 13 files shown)


mattcaswell:
Rebased now that #11826 was merged. I've also taken this out of WIP.

This is ready for review. Please take a look.

mattcaswell:
Ping

crypto/provider.c
```diff
@@ -57,6 +57,19 @@ int OSSL_PROVIDER_get_params(const OSSL_PROVIDER *prov, OSSL_PARAM params[])
...
 
+
+const OSSL_ALGORITHM *OSSL_PROVIDER_query_operation(const OSSL_PROVIDER *prov,
+                                                    int operation_id,
+                                                    int *no_cache)
+{
+    return ossl_provider_query_operation(prov, operation_id, no_cache);
+}
+
+void *OSSL_PROVIDER_get_provider_ctx(const OSSL_PROVIDER *prov)
```
slontis, line 68: Are the public interfaces supposed to use get0 ?

crypto/provider_core.c
```diff
@@ -759,6 +759,11 @@ const char *ossl_provider_module_path(const OSSL_PROVIDER *prov)
 #endif
 }
 
+void *ossl_provider_prov_ctx(const OSSL_PROVIDER *prov)
+{
+    return prov->provctx;
```
slontis, line 764: NULL check here maybe?

ssl/t1_lib.c
```diff
@@ -910,16 +910,72 @@ static const uint16_t tls_default_sigalg[] = {
...
+        /*
+         * Check hash is available.
+         * TODO(3.0): This test is not perfect. A provider could have support
+         * for a signature scheme, but not a particular hash. However the hash
+         * could be available from some other loaded provider. In that case it
+         * could be that the signature is available, and the hash is available
+         * independently - but not as a combination. We ignore this for now.
+         */
+        if (lu->hash != NID_undef
+                && ctx->ssl_digest_methods[lu->hash_idx] == NULL) {
```
slontis, line 941: Could this do a fetch here?

ssl/t1_lib.c
```diff
@@ -910,16 +910,72 @@ static const uint16_t tls_default_sigalg[] = {
...
+    return ret;
+}
+
 /* Lookup TLS signature algorithm */
-static const SIGALG_LOOKUP *tls1_lookup_sigalg(uint16_t sigalg)
+static const SIGALG_LOOKUP *tls1_lookup_sigalg(const SSL *s, uint16_t sigalg)
 {
     size_t i;
-    const SIGALG_LOOKUP *s;
+    const SIGALG_LOOKUP *lu;
```
slontis, line 971: legacy_rsa_sigalg is ignored? Is that intended?

test/sslapitest.c
```diff
@@ -7664,6 +7671,131 @@ static int test_servername(int tst)
...
+
+    if (!TEST_true(create_ssl_ctx_pair(libctx, TLS_server_method(),
+                                       TLS_client_method(),
+                                       TLS1_VERSION,
+                                       0,
+                                       &sctx, &cctx, cert, privkey)))
+        goto end;
+
+    if (idx < 3) {
+        if (!SSL_CTX_set1_sigalgs_list(cctx,
```
slontis, line 7742: This looks like it should be using sctx?

test/sslapitest.c
```diff
@@ -7664,6 +7671,131 @@ static int test_servername(int tst)
...
+                                       "rsa_pss_rsae_sha384"
+                                       ":rsa_pss_rsae_sha256")
+                || !SSL_CTX_set1_sigalgs_list(cctx,
+                                              "rsa_pss_rsae_sha384"
+                                              ":rsa_pss_rsae_sha256"))
+            goto end;
+    } else {
+        if (!SSL_CTX_set1_sigalgs_list(cctx, "rsa_pss_rsae_sha256:ECDSA+SHA256")
```
slontis, line 7750: And here also sctx?

test/sslapitest.c
```diff
@@ -37,6 +38,10 @@
 #include "internal/ktls.h"
 #include "../ssl/ssl_local.h"
 
+/* Defined in filterprov.c */
+OSSL_provider_init_fn filter_provider_init;
+int filter_provider_set_filter(int operation, const char *name);
```
slontis, line 43: This almost makes me wonder if there should be a setparams for providers.

doc/man3/OSSL_PROVIDER.pod
```diff
@@ -82,6 +87,16 @@ The caller must prepare the B<OSSL_PARAM> array before calling this
 function, and the variables acting as buffers for this parameter array
 should be filled with data when it returns successfully.
 
+OSSL_PROVIDER_query_operation() calls the provider's I<query_operation>
+function (see L<provider(7)>), if the provider has one. It should return an
```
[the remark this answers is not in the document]
mattcaswell: Done

doc/man3/OSSL_PROVIDER.pod
```diff
@@ -82,6 +87,16 @@ The caller must prepare the B<OSSL_PARAM> array before calling this
...
+array of I<OSSL_ALGORITHM> for the given I<operation_id> terminated by an all
+NULL OSSL_ALGORITHM entry. This is considered a low-level function that most
+applications should not need to call.
+
+If it is permissible to cache references to this array then I<*no_store> is set
+to 0 or 1 otherwise. If the array is not cacheable then it should be assumed to
```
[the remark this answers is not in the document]
mattcaswell: Done

test/filterprov.c
```diff
@@ -0,0 +1,199 @@
+/*
+ * Copyright 2019-2020 The OpenSSL Project Authors. All Rights Reserved.
```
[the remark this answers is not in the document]
mattcaswell: Fixed

crypto/provider.c
mattcaswell: Yes. Fixed.

crypto/provider_core.c
mattcaswell: Done.

ssl/t1_lib.c
mattcaswell: That's effectively what its doing. The check of ssl_digest_methods confirms that we have the given hash available (which it has previously checked by fetching it). The subsequent call of `EVP_PKEY_CTX_new_from_pkey` does a fetch of the keymgmt for the signature.

ssl/t1_lib.c
mattcaswell: `legacy_rsa_sigalg` is a kind of "fallback if all else fails". If we've fallen back and the fallback fails then we're doomed anyway, so I don't think we need to worry about checking availability of those sigalgs up front.

test/sslapitest.c
mattcaswell: Oops. Good catch. Fixed.

test/sslapitest.c
mattcaswell: Fixed.

test/sslapitest.c
mattcaswell: Perhaps...but not for this PR!!

mattcaswell:
All feedback addressed. Please take another look.

mattcaswell:
Pushed. Thanks.

PR #11837, opened 2020-05-15
CORE: Fix a couple of bugs in algorithm_do_this()

levitte:
The call of ossl_provider_query_operation() used |data->operation_id|,
when |cur_operation| should be used.

If any ossl_provider_query_operation() call returned NULL, the loop
was stopped, when it should just continue on to the next operation.


levitte:
Merged

827f04d5105e9bec0af214c42b8ad799fba5bb0d CORE: Fix a couple of bugs in algorithm_do_this()

PR #11842, opened 2020-05-15
CORE: query for operations only once per provider (unless no_store is true)

### LINKED ISSUE #11814: Provider query function called many times
While running a test in the debugger I noticed that the provider query function was being called many times - even though `no_cache` is set to 0.

I *think* what is happening is that the algorithms that are present are caching. But algorithms that are not present result in a fresh query to the underlying provider every time. But for most providers (unless `no_cache` is set to 1) then the result is not going to change every time you query it. We really should remember whether we have previously queried a provider and not ask again if we haven't. If we've already previously asked all the available providers, and the algorithm isn't in the cache - then we don't have it and should fail.

levitte:
When a desired algorithm wasn't available, we didn't register anywhere
that an attempt had been made, with the result that next time the same
attempt was made, the whole process would be done again.

To avoid this churn, we register a bit for each operation that has
been queried in the libcrypto provider object, and test it before
trying the same query and method construction loop again.

If course, if the provider has told us not to cache, we don't register
this bit.

Fixes #11814

### DIFF
--- a/crypto/core_algorithm.c
+++ b/crypto/core_algorithm.c
@@ -16,8 +16,11 @@
 struct algorithm_data_st {
     OPENSSL_CTX *libctx;
     int operation_id;            /* May be zero for finding them all */
+    int (*pre)(OSSL_PROVIDER *, int operation_id, void *data, int *result);
     void (*fn)(OSSL_PROVIDER *, const OSSL_ALGORITHM *, int no_store,
                void *data);
+    int (*post)(OSSL_PROVIDER *, int operation_id, int no_store, void *data,
+                int *result);
     void *data;
 };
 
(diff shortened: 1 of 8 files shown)


levitte:
Note: this incorporates #11837, and is therefore WIP until it can be rebased

mattcaswell:
LGTM

t8m:
There are some doc-nits errors introduced as seen in the Travis build

levitte:
Documentation nit fixed

levitte:
Squashed fixups, then rebased.  Not WIP any more, as #11837 is now merged

paulidale:
I'm too tired to comprehend the code at present.

Does it only cache negative outcomes if there are zero _no_cache_ outcomes?
A positive outcome can be cached in the presence of _no_cache_ results since we don't specify which of several matches are returned.  However, unmatched optional properties confound this.

mattcaswell:
> Does it only cache negative outcomes if there are zero no_cache outcomes?

This is kind of the wrong question. This doesn't change the caching or results directly - it just remembers whether we have previously called a provider's query function for a given operation or not. If we have then we don't call it again *unless* no_cache is set.

levitte:
Merged

5a29b6286f8ccafc2ed9a026b0e8d4bd6d0396e6 CORE: query for operations only once per provider (unless no_store is true)

PR #11846, opened 2020-05-16
Add cpu info str to speed, version and info command

bernd-edlinger:
This is cherry-pick of
47ca8338358b01ef429a3801ce6173f7a0791674
096978f09908ba0f679ff905b0db4861a57eb1c8
363e941ed43c648adf4d6d0874077ddd80041e1f

to 1.1.1

##### Checklist

### DIFF
--- a/apps/version.c
+++ b/apps/version.c
@@ -33,7 +33,7 @@
 
 typedef enum OPTION_choice {
     OPT_ERR = -1, OPT_EOF = 0, OPT_HELP,
-    OPT_B, OPT_D, OPT_E, OPT_F, OPT_O, OPT_P, OPT_V, OPT_A, OPT_R
+    OPT_B, OPT_D, OPT_E, OPT_M, OPT_F, OPT_O, OPT_P, OPT_V, OPT_A, OPT_R
 } OPTION_CHOICE;
 
 const OPTIONS version_options[] = {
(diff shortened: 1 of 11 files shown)


bernd-edlinger:
I wonder what happened to appveyor CI:
interesting which error happens in the
following build `Environment: VSVER=15; Configuration: shared; Platform: x64 `
https://ci.appveyor.com/project/openssl/openssl/builds/32924846/job/8p0t1b4m4c4u8be2:
`ObjectNotFound: (Microsoft.Power....`

the same error happens for PRs againt master and 1.1.1 so it must be a general problem.

apps/version.c
```diff
@@ -42,29 +42,21 @@ const OPTIONS version_options[] = {
     {"b", OPT_B, '-', "Show build date"},
     {"d", OPT_D, '-', "Show configuration directory"},
     {"e", OPT_E, '-', "Show engines directory"},
+    {"m", OPT_M, '-', "Show modules directory"},
```
levitte, line 45: There is no such thing in 1.1.1...

levitte:
`openssl info` is a new feature in 3.0, I don't see that it's viable for 1.1.1

apps/version.c
bernd-edlinger: right, and it just prints N/A but I see that as a useful feature for scripts,
instead of having to figure out which version it is,

bernd-edlinger:
The main motivation for this PR is the need for proper CPUINFO
in speed and similar commands, I just needed some other commits since otherwise
the the code would not be merged cleanly, as these were committed before
the CPUINFO.  I think we need to have this information so users of
the LTS branch are able to report their CPU flags, in case we ask them to do so.
Having more information in openssl version and openssl info just looks like the
right thing to do.

PR #11848, opened 2020-05-17
Fix egd and devrandom source configs

bernd-edlinger:
./config --with-rand-seed=egd

need to defines OPENSSL_RAND_SEED_EGD and OPENSSL_NO_EGD
so get rid of OPENSSL_NO_EGD (compiles but I did not really test EGD)

./config --with-rand-seed=devrandom

does not work since wait_random_seeded works under the assumption
that OPENSSL_RAND_SEED_GETRANDOM is supposed to be enabled as well,
that is usually the case, but not when only devrandom is enabled.
Skip the wait code in this special case.

##### Checklist

### DIFF
--- a/crypto/rand/rand_unix.c
+++ b/crypto/rand/rand_unix.c
@@ -391,7 +391,8 @@ static struct random_device {
 } random_devices[OSSL_NELEM(random_device_paths)];
 static int keep_random_devices_open = 1;
 
-#   if defined(__linux) && defined(DEVRANDOM_WAIT)
+#   if defined(__linux) && defined(DEVRANDOM_WAIT) \
+       && defined(OPENSSL_RAND_SEED_GETRANDOM)
 static void *shm_addr;
 
 #    if !defined(FIPS_MODULE)
--- a/crypto/rand/rand_unix.c
+++ b/crypto/rand/rand_unix.c
@@ -475,7 +475,7 @@ static int wait_random_seeded(void)
     }
     return seeded;
 }
-#   else /* defined __linux */
+#   else /* defined __linux && DEVRANDOM_WAIT && OPENSSL_RAND_SEED_GETRANDOM */
 static int wait_random_seeded(void)
 {
     return 1;
--- a/Configure
+++ b/Configure
@@ -1124,6 +1124,9 @@ if (scalar(@seed_sources) == 0) {
     print "Using os-specific seed configuration\n";
     push @seed_sources, 'os';
 }
+if (scalar(grep { $_ eq 'egd' } @seed_sources) > 0) {
+    delete $disabled{'egd'};
+}
 if (scalar(grep { $_ eq 'none' } @seed_sources) > 0) {
     die "Cannot seed with none and anything else" if scalar(@seed_sources) > 1;
     warn <<_____ if scalar(@seed_sources) == 1;


bernd-edlinger:
Hope you don't mind that this fixes 2 issues at once.
@paulidale you are right, in the moment this will indeed cherry-pick
from master to 1.1.1 which proves that I can be wrong from time to time,

bernd-edlinger:
I think the CI failure is only a stress induced one.
The CI failure has nothing to do with this PR, since, we do not even try
to test --with-rand-seed in the CI, if I am correct, which happens
rather often.

bernd-edlinger:
@paulidale one part of this PR fixes an oversight with the kernel version check.
I ran into this issue, when I used linux 4.20+ as kernel, and configured
./config --with-rand-seed=devrandom
and was puzzled that openssl speed rand does print funny errors,
because the devrandom seeding was failing, our check assumed that
getentropy is enabled at the same time, and that it must succeed.
I think if that devrandom is configured explicitly we do not need this
waiting for randomness at all, since we cannot be sure that it works
with unknown linux versions.
The other part, is simply the egd random source, I already noted
that with 1.1.1d but at that time I did not realize that --with-rand-seed=egd
is defeated by the default-enabled no-egd option

crypto/rand/rand_unix.c
```diff
@@ -391,7 +391,8 @@ static struct random_device {
 } random_devices[OSSL_NELEM(random_device_paths)];
 static int keep_random_devices_open = 1;
 
-#   if defined(__linux) && defined(DEVRANDOM_WAIT)
+#   if defined(__linux) && defined(DEVRANDOM_WAIT) \
+       && defined(OPENSSL_RAND_SEED_GETRANDOM)
```
levitte, line 395: The comment at the `#else` line (line 478) needs a bit of editing...

levitte:
@openssl, would it be possible to refactor `rand_unix.c` into seed source specific source files?  The ifdef sallad is very hard to read!

bernd-edlinger:
Yes, that is hard to follow.
I think this will be moved to a provider anyway, that would be an opportunity for
more re-factoring.

paulidale:
Weren't you the one who said #ifdef mazes were normal and okay????  I've started some refactoring in #11682 but I've not touched on the individual source files yet.

bernd-edlinger:
I would pretend amnesia, since I cannot remember having said that.
But yes, that ifdef puzzle is in fact openssl style.

bernd-edlinger:
Oh, nice, now I get also spam in my inbox from CIFuzz:

```
Run failed for fix_random_sources (94310a5)

Repository: openssl/openssl
Workflow: CIFuzz
Duration: 5 minutes and 3.0 seconds
Finished: 2020-05-18 07:06:40 UTC

View results: https://github.com/openssl/openssl/actions/runs/107873782

Jobs:
  * Fuzzing failed (2 annotations)

-- 
You are receiving this because this workflow ran on your branch.
Manage your GitHub Actions notifications: https://github.com/settings/notifications
```

Have you got those e-mails also,
I did not ask for them.

paulidale:
@bernd-edlinger, I was addressing @levitte.  I didn't notice your comment betwixt.

bernd-edlinger:
I knew that, but I couldn't resist making a joke.

levitte:
Er, I can't remember taking a personal stand of the sort. That being said, openssl has a history of ifdef sallads, to the point of it becoming "normal", maybe I alluded to that?

bernd-edlinger:
Hi RND-experts, @mspncp @paulidale : 
I would be surprised, if it was the case, but may I ask is this PR controversial ?

bernd-edlinger:
Merged to master as ddec332 and 1.1.1 as 5f10fce.
Thanks!

paulidale:
Thanks to @mspncp's efforts in fixing file moves, this didn't cause a merge conflict!

mspncp:
I doubt you would have even gotten a merge conflict. But Bernd's change would have ended up in the old location and would have been at risk to be dropped inadvertently, unless manually applied to the module at the new location.

paulidale:
If I'd gone as far as removing the file, it definitely would have conflicted :)
I'm well past that point now.

PR #11850, opened 2020-05-17
Fix some places where X509_up_ref is used

bernd-edlinger:
without error handling.

This takes up the ball from #11278
without trying to solve everything at once.

[extended tests]

##### Checklist


crypto/x509/x509_vfy.c
```diff
@@ -370,11 +376,15 @@ static int check_issued(X509_STORE_CTX *ctx, X509 *x, X509 *issuer)
 static int get_issuer_sk(X509 **issuer, X509_STORE_CTX *ctx, X509 *x)
 {
     *issuer = find_issuer(ctx, ctx->other_ctx, x);
-    if (*issuer) {
-        X509_up_ref(*issuer);
-        return 1;
-    } else
-        return 0;
+
+    if (!*issuer || !X509_up_ref(*issuer))
```
mattcaswell, line 380: Could you make this `*issuer != NULL`?

crypto/x509/x_pubkey.c
```diff
@@ -219,8 +219,10 @@ EVP_PKEY *X509_PUBKEY_get0(X509_PUBKEY *key)
 EVP_PKEY *X509_PUBKEY_get(X509_PUBKEY *key)
 {
     EVP_PKEY *ret = X509_PUBKEY_get0(key);
```
mattcaswell, line 221: Please add a blank line after this declaration

crypto/x509/x_pubkey.c
```diff
@@ -219,8 +219,10 @@ EVP_PKEY *X509_PUBKEY_get0(X509_PUBKEY *key)
...
-    if (ret != NULL)
-        EVP_PKEY_up_ref(ret);
+    if (ret != NULL && !EVP_PKEY_up_ref(ret)) {
+        X509err(X509_F_X509_PUBKEY_GET0, ERR_R_INTERNAL_ERROR);
```
mattcaswell, line 223: Wrong F code here. You can just make it 0, since the F codes are no longer used.

crypto/x509/x509_vfy.c
bernd-edlinger: Yes, good point.

crypto/x509/x_pubkey.c
bernd-edlinger: ok.

crypto/x509/x_pubkey.c
bernd-edlinger: Incredibly Travis found this also.  So I will fix this, but it means
this will not cherry-pick cleanly to 1.1.1, sigh...

include/openssl/x509err.h
```diff
@@ -101,8 +103,8 @@ int ERR_load_X509_strings(void);
 # define X509_R_BAD_X509_FILETYPE                         100
 # define X509_R_BASE64_DECODE_ERROR                       118
 # define X509_R_CANT_CHECK_DH_KEY                         114
-# define X509_R_CERT_ALREADY_IN_HASH_TABLE                101
 # define X509_R_CERTIFICATE_VERIFICATION_FAILED           139
+# define X509_R_CERT_ALREADY_IN_HASH_TABLE                101
```
bernd-edlinger, line 107: I wonder why these lines always change, does the sort order somehow depend on my
perl version or do everybody else, do something different, (i did `make update`)

mattcaswell: I don't know but I've seen it many times before. It's annoying.

bernd-edlinger:
> Please make a different PR for 1.1.1 since this won't cherry-pick.

yeah, already at it.

bernd-edlinger:
1.1.1 version in #11852

bernd-edlinger:
Pushed to master as e9e7b5d 
Thanks!

PR #9781, opened 2019-09-05
[crypto/ec] for ECC parameters with NULL or zero cofactor, compute it (1.1.1)

bbbrumley:
Backport of [#9727](https://github.com/openssl/openssl/pull/9727) to 1.1.1.

The regression test is pretty clear in this regard, but anyway for due diligence I checked the debugger with an explicit params PEM with the `dgst` cmd to sign. It early exits as expected to the SCA hardened code path.

PR #11851, opened 2020-05-17
deprecate EC precomputation functionality

bbbrumley:
Reiterating from #8332:

* Deprecate `EC_GROUP_precompute_mult` `EC_GROUP_have_precompute_mult` and `EC_KEY_precompute_mult`. The `speed` app is the only software I've ever found using them. (Besides, with recent EC changes most use cases don't have access to this type of precomp anyway. Only the hard coded constants from the custom `EC_METHOD`s.)

Although I've dug several times in the past for linking applications using this functionality, I took another stab today. I found [python-cryptography](https://github.com/pyca/cryptography/commit/6639dd15a7847fae731b29ce01f98cba79757838#diff-c0b27d970c896efabfcb87ec21548a2d) used to have the bindings available, but as you can see they were removed because they weren't used.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,13 @@ OpenSSL 3.0
 
 ### Changes between 1.1.1 and 3.0 [xx XXX xxxx]
 
+* Deprecated EC_GROUP_precompute_mult(), EC_GROUP_have_precompute_mult(), and
+  EC_KEY_precompute_mult() These functions are not widely used and applications
+  should instead switch to named curves which OpenSSL has hardcoded lookup
+  tables for.
+
+  *Billy Bob Brumley*
+
  * Deprecated EC_POINTs_mul(). This function is not widely used and applications
    should instead use the L<EC_POINT_mul(3)> function.
 
--- a/apps/speed.c
+++ b/apps/speed.c
@@ -3049,7 +3049,6 @@ int speed_main(int argc, char **argv)
             rsa_count = 1;
         } else {
             for (i = 0; i < loopargs_len; i++) {
-                EC_KEY_precompute_mult(loopargs[i].ecdsa[testnum], NULL);
                 /* Perform ECDSA signature test */
                 EC_KEY_generate_key(loopargs[i].ecdsa[testnum]);
                 st = ECDSA_sign(0, loopargs[i].buf, 20, loopargs[i].buf2,
--- a/crypto/ec/ec_key.c
+++ b/crypto/ec/ec_key.c
@@ -781,12 +781,14 @@ void EC_KEY_set_asn1_flag(EC_KEY *key, int flag)
         EC_GROUP_set_asn1_flag(key->group, flag);
 }
 
+#ifndef OPENSSL_NO_DEPRECATED_3_0
 int EC_KEY_precompute_mult(EC_KEY *key, BN_CTX *ctx)
 {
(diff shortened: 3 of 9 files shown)


beldmit:
I'm not fond of this idea. I use `EC_GROUP_precompute_mult` for GOST-related apps.

beldmit:
For my case (GOST-only systems) this precomputation brings a significant performance boost (> 10%) comparing to constructing EC_GROUP on demand and not doing the precomputation. I'd like to leave this functionality.

romen:
@beldmit is the 10% improvement measured against recent versions?

I can see how there could still be some benefit from precomputation on the signature verification path, but I would expect no effect on keygen, ecdh and signature generation given that for those cases we are currently overriding the scalar multiplication routine before going into the wnaf algorithm that used the precomputation data.

Would it be possible for you to run benchmarks with and without the precomputation call to measure the difference at this time?

beldmit:
These measurements were done for 10000 signatures and verifications using 1.1.1d
Were there any significant changes in the underlying code?

romen:
#11435 and #11439 have happened after 1.1.1d and might show some effect on your measurements, but the hijacking of the scalar mul to not use wnaf (and the precomputation data) should have been included in 1.1.1d already, if I am not mistaken.

Was the 10% improvement related only to signature verifications? I really fail to see how it could be happening for the code paths that should be hardened! 

If you see 10% improvement across both verification and keygen/dh/siggen it is worth investigating with the debugger what code path is being followed and why!


Can it be that the tests were run leaving the cofactor undefined? That could explain why the insecure code path was being taken, but than I would strongly recommend revising the curve parameters and setting them all to avoid security risks!

beldmit:
Should I test the difference in performance in master or 1.1.1g is enough?

bbbrumley:
> Should I test the difference in performance in master or 1.1.1g is enough?

1.1.1g should be enough.

It would be helpful to have independent measurements for EVP `keygen`, `derive`, `sign`, and `verify` so we can understand the performance diff.

bbbrumley:
> Can it be that the tests were run leaving the cofactor undefined? That could explain why the insecure code path was being taken, but than I would strongly recommend revising the curve parameters and setting them all to avoid security risks!

I agree with always providing the cofactor for good measure, but after #9781 this shouldn't matter. (I don't recall which 111 flavor this intersected -- it was a backport.)

beldmit:
Benchmark, part 1, cached groups, EC_GROUP_precompute_mult, 20000 ops:
verify:  1457.832204/s
sign: 1240.310078/s
derive: 1310.186702/s

Benchmark, part 2, allocating groups on-demand, no precomputing mult:
sign: 1199.184555/s
verify: 1376.557230/s
derive: 1260.239445/s

Benchmark, part 3, preallocating groups, no precomputing mult:
sign: 1233.121647/s
verify: 1410.735699/s
derive: 1310.959622/s

So I was wrong and the performance degradation is not related to EC_GROUP_precompute_mult

romen:
I'd like if there was an explicit approval by @beldmit as GOST was probably the only other user of the precomputation functionality.

romen:
> LGTM.
> 
> Sorry for the disturbance.

No disturbance whatsoever!!
It was good to have external testing and a discussion on this!

beldmit:
Squashed and merged. Many thanks!

PR #11852, opened 2020-05-17
Fix some places where X509_up_ref is used [1.1.1]

bernd-edlinger:
without error handling.

This takes up the ball from #11278
without trying to solve everything at once.

[extended tests]

##### Checklist


bernd-edlinger:
@mattcaswell I hope the error macros are now right, again something totally unexpected
has changed.  Maybe that happen due to cherry-picks from different versions ?

include/openssl/pemerr.h
```diff
@@ -11,9 +11,7 @@
 #ifndef HEADER_PEMERR_H
 # define HEADER_PEMERR_H
 
-# ifndef HEADER_SYMHACKS_H
-#  include <openssl/symhacks.h>
-# endif
+# include <openssl/symhacks.h>
```
bernd-edlinger, line 14: again, funny, this was probably cherry-picked, I see it in several places, but `make update` reverts it.

kroeckx: make update does not result in any changes for me in the 1.1.1-stable branch

include/openssl/pemerr.h
```diff
@@ -32,17 +30,19 @@ int ERR_load_PEM_strings(void);
 # define PEM_F_D2I_PKCS8PRIVATEKEY_FP                     121
 # define PEM_F_DO_B2I                                     132
 # define PEM_F_DO_B2I_BIO                                 133
-# define PEM_F_DO_BLOB_HEADER                             134
+# define PEM_F_DO_BLOB_HEADER                             147
```
bernd-edlinger, line 33: Interesting this seems to be 134 since 2017 at least.
But the script changes it to 147.

include/openssl/pemerr.h
bernd-edlinger: yes, confirmed. But the script needs a trigger to do something.

bernd-edlinger: try `make update ERROR_REBUILD=-rebuild`

include/openssl/pemerr.h
kroeckx: If I revert your changes to the auto generated files, run make update, things work for me.

kroeckx: Did you maybe cherry-pick the auto generated files from master?

include/openssl/pemerr.h
kroeckx: I do get that change too after applying your patch

include/openssl/pemerr.h
bernd-edlinger: magic, It may have started because I cherry-picked, but the define
was missing in the header file, so X509err did not compile

bernd-edlinger:
I deleted all changes on generated files, and did again a make update,
this time the script picks
` define X509_F_X509_PUBKEY_GET                           161`
instead of 166 which was in master, but that is probably okay,
and makes less noise....

bernd-edlinger:
@kroeckx does it look better now?
BTW: this kind of proves my point that we should forbid cherry-picks
between master and 1.1.1, it would work, if they are more closely related,
like 1.1.0 and 1.1.1 , the difference is 0.01, but
between 1.1.1 and 3.0 the difference is 2.89, that is logical from a mathematical
point of view. But a majority decision can easily end up with 1+1 = 3  :-)

kroeckx:
I think auto generated files should always be generated in their own branch.

kroeckx:
But if we have multiple stable branches with the same soname, we
should check that they actually get the same number in those
branches.

bernd-edlinger:
I one approval is 2 days old, the other 22 h,
@t-j-h I will set the ready-to-merge manually, as the median of the approvals age is > 24 h,
okay?

bernd-edlinger:
merged to OpenSSL_1_1_1-stable as 5cea584
Thanks!

mattcaswell:
> I one approval is 2 days old, the other 22 h,
@t-j-h I will set the ready-to-merge manually, as the median of the approvals age is > 24 h,
okay?

ready-to-merge can be set 24 hours after the point where you received sufficient approvals for merge. Subsequent approvals do not reset the timer. So in this case this PR was eligible for the "ready-to-merge" label 24 hours after @kroeckx gave his approval.

PR #11857, opened 2020-05-18
Cast to at least 32 bit unsigned type when shifting left 24 places

### LINKED ISSUE #11853: Master build is K.O. :  runtime error: left shift of 184 by 24 places cannot be represented in type 'int'
See  https://travis-ci.org/github/openssl/openssl/builds/687826428

Config: 
```
EXTENDED_TEST="yes" CONFIG_OPTS="no-asm enable-ubsan enable-rc5 enable-md2 enable-ssl3 enable-ssl3-method enable-nextprotoneg no-shared enable-buildtest-c++ -fno-sanitize=alignment -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -Wno-unused-command-line-argument" CXX="clang++" 
```

Ouput :
```
read DSA key
crypto/pem/pvkfmt.c:42:18: runtime error: left shift of 184 by 24 places cannot be represented in type 'int'
../../util/wrap.pl ../../apps/openssl dsa -pubin -pubout -in msb-f.d -inform d -out msb-ff.dd -outform d => 1
    not ok 5 - d -> d

crypto/pem/pvkfmt.c:42:18: runtime error: left shift of 184 by 24 places cannot be represented in type 'int'
../../util/wrap.pl ../../apps/openssl dsa -pubin -pubout -in msb-f.d -inform d -out msb-ff.dp -outform p => 1
...
    not ok 8 - d -> p
    #   Failed test 'd -> p'
    #   at ../../test/recipes/tconversion.pl line 70.
```

t8m:
Fixes #11853

Implicit C promotion rules apply so the value is promoted to int otherwise and that might not be big enough when the value is then left-shifted 24 places.

### DIFF
--- a/crypto/bn/bn_mpi.c
+++ b/crypto/bn/bn_mpi.c
@@ -45,7 +45,7 @@ BIGNUM *BN_mpi2bn(const unsigned char *d, int n, BIGNUM *ain)
     int neg = 0;
     BIGNUM *a = NULL;
 
-    if (n < 4) {
+    if (n < 4 || (d[0] & 0x80) != 0) {
         BNerr(BN_F_BN_MPI2BN, BN_R_INVALID_LENGTH);
         return NULL;
     }
--- a/crypto/pem/pvkfmt.c
+++ b/crypto/pem/pvkfmt.c
@@ -36,10 +36,10 @@ static unsigned int read_ledword(const unsigned char **in)
 {
     const unsigned char *p = *in;
     unsigned int ret;
-    ret = *p++;
-    ret |= (*p++ << 8);
-    ret |= (*p++ << 16);
-    ret |= (*p++ << 24);
+    ret = (unsigned int)*p++;
+    ret |= (unsigned int)*p++ << 8;
+    ret |= (unsigned int)*p++ << 16;
+    ret |= (unsigned int)*p++ << 24;
     *in = p;
     return ret;
 }
--- a/ssl/t1_trce.c
+++ b/ssl/t1_trce.c
@@ -665,8 +665,10 @@ static int ssl_print_random(BIO *bio, int indent,
 
     if (*pmsglen < 32)
         return 0;
-    tm = ((unsigned int)p[0] << 24) | ((unsigned int)p[1] << 16)
-         | ((unsigned int)p[2] << 8) | (unsigned int)p[3];
+    tm = ((unsigned int)p[0] << 24)
+         | ((unsigned int)p[1] << 16)
+         | ((unsigned int)p[2] << 8)
+         | (unsigned int)p[3];
     p += 4;
     BIO_indent(bio, indent, 80);
     BIO_puts(bio, "Random:\n");
@@ -1369,8 +1371,10 @@ static int ssl_print_ticket(BIO *bio, int indent, const SSL *ssl,
     }
     if (msglen < 4)
         return 0;
-    tick_life = ((unsigned int)msg[0] << 24) | ((unsigned int)msg[1] << 16)
-                | ((unsigned int)msg[2] << 8) | (unsigned int)msg[3];
+    tick_life = ((unsigned int)msg[0] << 24)
+                | ((unsigned int)msg[1] << 16)
+                | ((unsigned int)msg[2] << 8)
+                | (unsigned int)msg[3];
     msglen -= 4;
     msg += 4;
     BIO_indent(bio, indent + 2, 80);
@@ -1381,8 +1385,10 @@ static int ssl_print_ticket(BIO *bio, int indent, const SSL *ssl,
         if (msglen < 4)
             return 0;
         ticket_age_add =
-            ((unsigned int)msg[0] << 24) | ((unsigned int)msg[1] << 16)
-            | ((unsigned int)msg[2] << 8) | (unsigned int)msg[3];
+            ((unsigned int)msg[0] << 24)
+            | ((unsigned int)msg[1] << 16)
+            | ((unsigned int)msg[2] << 8)
+            | (unsigned int)msg[3];
         msglen -= 4;
         msg += 4;
         BIO_indent(bio, indent + 2, 80);
--- a/ssl/t1_trce.c
+++ b/ssl/t1_trce.c
@@ -665,7 +665,8 @@ static int ssl_print_random(BIO *bio, int indent,
 
     if (*pmsglen < 32)
         return 0;
-    tm = (p[0] << 24) | (p[1] << 16) | (p[2] << 8) | p[3];
+    tm = ((unsigned int)p[0] << 24) | ((unsigned int)p[1] << 16)
+         | ((unsigned int)p[2] << 8) | (unsigned int)p[3];
     p += 4;
     BIO_indent(bio, indent, 80);
     BIO_puts(bio, "Random:\n");
@@ -870,8 +871,10 @@ static int ssl_print_extension(BIO *bio, int indent, int server,
             break;
         if (extlen != 4)
             return 0;
-        max_early_data = (ext[0] << 24) | (ext[1] << 16) | (ext[2] << 8)
-                         | ext[3];
+        max_early_data = ((unsigned int)ext[0] << 24)
+                         | ((unsigned int)ext[1] << 16)
+                         | ((unsigned int)ext[2] << 8)
+                         | (unsigned int)ext[3];
         BIO_indent(bio, indent + 2, 80);
         BIO_printf(bio, "max_early_data=%u\n", max_early_data);
         break;
@@ -1366,7 +1369,8 @@ static int ssl_print_ticket(BIO *bio, int indent, const SSL *ssl,
     }
     if (msglen < 4)
         return 0;
-    tick_life = (msg[0] << 24) | (msg[1] << 16) | (msg[2] << 8) | msg[3];
+    tick_life = ((unsigned int)msg[0] << 24) | ((unsigned int)msg[1] << 16)
+                | ((unsigned int)msg[2] << 8) | (unsigned int)msg[3];
     msglen -= 4;
     msg += 4;
     BIO_indent(bio, indent + 2, 80);
@@ -1377,7 +1381,8 @@ static int ssl_print_ticket(BIO *bio, int indent, const SSL *ssl,
         if (msglen < 4)
             return 0;
         ticket_age_add =
-            (msg[0] << 24) | (msg[1] << 16) | (msg[2] << 8) | msg[3];
+            ((unsigned int)msg[0] << 24) | ((unsigned int)msg[1] << 16)
+            | ((unsigned int)msg[2] << 8) | (unsigned int)msg[3];
         msglen -= 4;
         msg += 4;
         BIO_indent(bio, indent + 2, 80);


levitte:
As a matter of experimenting, would `*p++ << 24U` work as well?

opensignature:
imho, casting should be enough for 24 and not be necessary for the other values
ret | = (* p ++ << 16);
ret | = ((unsigned int) * p ++ << 24);

levitte:
Why not be consistent?

opensignature:
> Why not be consistent?

ok, but it's not even nice to disseminate code of (unsigned int).
The problem is not C, but the -fsanitize = undefined parameter of gcc.
If I compile with gcc only i have no error

t8m:
> As a matter of experimenting, would `*p++ << 24U` work as well?

I do not know, and I am not sure it makes sense to even try, because that is much more ugly.

t8m:
> > Why not be consistent?
> 
> ok, but it's not even nice to disseminate code of (unsigned int).
> The problem is not C, but the -fsanitize = undefined parameter of gcc.
> If I compile with gcc only i have no error

That does not mean that the undefined behavior of overflowing the value into the sign is not triggered. Of course in reality it will work because the shift into the sign will produce exactly the same value when finally cast to unsigned int.

t8m:
> ok, but it's not even nice to disseminate code of (unsigned int).

And to this - feel free to submit a later PR that replaces all the usage of unsigned int in that source file with something else. But that is clearly for a different PR than this one that fixes a concrete build breakage.

crypto/bn/bn_mpi.c
```diff
@@ -45,7 +45,7 @@ BIGNUM *BN_mpi2bn(const unsigned char *d, int n, BIGNUM *ain)
     int neg = 0;
     BIGNUM *a = NULL;
 
-    if (n < 4) {
+    if (n < 4 || d[0] & 0x80) {
```
bernd-edlinger, line 48: this looks wrong

kroeckx:
> Of course in reality it will work because the shift into the sign will produce exactly the same value when finally cast to unsigned int.

Being undefined, it does not need to be the same value. It might
preserve the sign, or not, depending on the platform.

On the other hand, a right shift on a signed type is implementation
defined, and we should avoid that too.

crypto/bn/bn_mpi.c
t8m: Why? Could you be please more specific? I am trying to avoid values which could not be represented in signed long type (after the shift) which is used in the following code.

opensignature:
> And to this - feel free to submit a later PR 

https://github.com/openssl/openssl/pull/11860

crypto/bn/bn_mpi.c
bernd-edlinger: First long is usually longer than 4 bytes.
But also d[0..3] is probably not a signed quantity,
d[4] is the sign bit.

bernd-edlinger: I think the problem is it uses long, probably unsigned long would be more appropriate?

bernd-edlinger: could it be that this is not the problem you wanted to address, since this code
has lots more undefined behavior than this?

crypto/bn/bn_mpi.c
```diff
@@ -45,7 +45,7 @@ BIGNUM *BN_mpi2bn(const unsigned char *d, int n, BIGNUM *ain)
...
         BNerr(BN_F_BN_MPI2BN, BN_R_INVALID_LENGTH);
         return NULL;
     }
```
bernd-edlinger, line 51: Oouch, the line below is
```C
len = ((long)d[0] << 24) | ((long)d[1] << 16) | ((int)d[2] << 8) | (int)
    d[3];
```
This was apparently written in the days when int was a 16 bit short value.
So if you like please rewrite the whole stuff, in a PR of its own.

crypto/bn/bn_mpi.c
kaduk: Wouldn't we want to compare the result of the bitwise-and against zero anyway, as a matter of style?

kaduk:
> > As a matter of experimenting, would `*p++ << 24U` work as well?
> 
> I do not know, and I am not sure it makes sense to even try, because that is much more ugly.

It would not work, FWIW.

crypto/bn/bn_mpi.c
bernd-edlinger: to be completely correct you should add `&& sizeof(long) == 4`

bernd-edlinger:
I would prefer to fix here only what is necessary to make the test case work.

crypto/pem/pvkfmt.c
```diff
@@ -36,10 +36,10 @@ static unsigned int read_ledword(const unsigned char **in)
...
     const unsigned char *p = *in;
     unsigned int ret;
-    ret = *p++;
-    ret |= (*p++ << 8);
-    ret |= (*p++ << 16);
-    ret |= (*p++ << 24);
+    ret = (unsigned int)*p++;
+    ret |= ((unsigned int)*p++ << 8);
+    ret |= ((unsigned int)*p++ << 16);
+    ret |= ((unsigned int)*p++ << 24);
```
paulidale, line 42: Personally, I'd ditch the outer parentheses on these three.

bernd-edlinger: Yes, I agree

ssl/t1_trce.c
```diff
@@ -665,7 +665,8 @@ static int ssl_print_random(BIO *bio, int indent,
 
     if (*pmsglen < 32)
         return 0;
-    tm = (p[0] << 24) | (p[1] << 16) | (p[2] << 8) | p[3];
+    tm = ((unsigned int)p[0] << 24) | ((unsigned int)p[1] << 16)
+         | ((unsigned int)p[2] << 8) | (unsigned int)p[3];
```
bernd-edlinger, line 669: Please do that in another PR.
Reason: there is no test coverage for the undefined behavior.
so no test gets fixed by this.  please correct me if I am wrong.

crypto/bn/bn_mpi.c
bernd-edlinger, line 48: Also this change may be an improvement but there is no test coverage for it,
which can be seen by the fact that the reported build failure is not affected by this change

ssl/t1_trce.c
```diff
@@ -870,8 +871,10 @@ static int ssl_print_extension(BIO *bio, int indent, int server,
             break;
         if (extlen != 4)
             return 0;
-        max_early_data = (ext[0] << 24) | (ext[1] << 16) | (ext[2] << 8)
-                         | ext[3];
+        max_early_data = ((unsigned int)ext[0] << 24)
+                         | ((unsigned int)ext[1] << 16)
+                         | ((unsigned int)ext[2] << 8)
+                         | (unsigned int)ext[3];
```
bernd-edlinger, line 877: Likewise here, no test coverage for the previously undefined behavior.

ssl/t1_trce.c
```diff
@@ -1366,7 +1369,8 @@ static int ssl_print_ticket(BIO *bio, int indent, const SSL *ssl,
     }
     if (msglen < 4)
         return 0;
-    tick_life = (msg[0] << 24) | (msg[1] << 16) | (msg[2] << 8) | msg[3];
+    tick_life = ((unsigned int)msg[0] << 24) | ((unsigned int)msg[1] << 16)
+                | ((unsigned int)msg[2] << 8) | (unsigned int)msg[3];
```
bernd-edlinger, line 1373: and here, if I am right

ssl/t1_trce.c
```diff
@@ -1377,7 +1381,8 @@ static int ssl_print_ticket(BIO *bio, int indent, const SSL *ssl,
         if (msglen < 4)
             return 0;
         ticket_age_add =
-            (msg[0] << 24) | (msg[1] << 16) | (msg[2] << 8) | msg[3];
+            ((unsigned int)msg[0] << 24) | ((unsigned int)msg[1] << 16)
+            | ((unsigned int)msg[2] << 8) | (unsigned int)msg[3];
```
bernd-edlinger, line 1385: and here

crypto/bn/bn_mpi.c
t8m: So we want to support a more than 2GB large BIGNUMs on systems that allow this? I think it is better to be consistent and have the limit be the same on all supported architectures.

bernd-edlinger: good argument but dor a denial of service attack also 100.000 bit BIGNUMS would
be a threat.

t8m:
@bernd-edlinger I am going to split these into separate commits, but I really think this should be kept as a single PR.

crypto/bn/bn_mpi.c
t8m: The line below is technically correct albeit a little bit ugly.

crypto/bn/bn_mpi.c
t8m: > good argument but dor a denial of service attack also 100.000 bit BIGNUMS would
> be a threat.

Now that would be for a different PR. But I do not think this function is used in a way that a DoS attack would apply.

ssl/t1_trce.c
t8m: I disagree with doing a different PR just for this reason.

bernd-edlinger: yeah we disagree.

bernd-edlinger:
@t8m I approve your commit 32cf8906d6d45bbc8466155b9bf819d486e1b62a

Note I replaced my #11867 PR with your commit.
The problem with the other commits is they make the [extended tests]
being ignored.

But if it is in a PR of its own the extended tests will be honored by the CI engines.

levitte:
@t8m, you need to add `[extended tests]` to the most recent commit's message for it to take effect.

Here's where it's detected (it's not a github feature):

https://github.com/openssl/openssl/blob/5a29b6286f8ccafc2ed9a026b0e8d4bd6d0396e6/.travis.yml#L143-L145

crypto/bn/bn_mpi.c
bernd-edlinger: you are of course right that this needs improvements.
But if you change the de-serialization to choke at 2GB you should
also change the inverse function above.  IT has undefined behavior
at 2GB, due to integer overflow.

bernd-edlinger: any way, it is fun for me to discuss these issues with you.
and disagreement is not per se a bad thing.

levitte:
> the unrelated changes are in principle okay, but belong to another PR,

Er...  as far as I can see, these are *related* changes, they all deal with the possible overflow because of promotion to `int`.

bernd-edlinger:
Yes, I agree that this PR make good progress.
If that would be the complete list, where undefined behavior of that kind happens,
I would be rather surprised.  Have you done something like lint or so?
Or are those places just spotted my chance?

t8m:
> Yes, I agree that this PR make good progress.
> If that would be the complete list, where undefined behavior of that kind happens,
> I would be rather surprised. Have you done something like lint or so?
> Or are those places just spotted by chance?

I've just grepped for `<< 24`

t8m:
Fixup commit to fix the formatting.

t8m:
The previously broken extended tests already passed in the Travis.

levitte:
Travis failures is just the usual s390x timeout

t8m:
So anybody willing to give an approval?

opensignature:
> Travis failures is just the usual s390x timeout

mmmh Travis, you need a larger container ...

error: could not write config file /home/travis/build/openssl/openssl/.git/config: No space left on device

bernd-edlinger:
> Travis failures is just the usual s390x timeout

I remember some of the s390x failures recently were because
of "no space left on disk device"

romen:


None of the comments since the `approval: done` label was assigned was intended to invalidate the approval: I set the `approval: ready to merge` status manually.

t8m:
Merged to master.

romen:
@t8m what about the `1.1.1` label?

romen:
Sorry, I mis-clicked on `Comment` too quickly!

What I meant was:

> @t8m what about the 1.1.1 label?

Should we have a separate PR before merging just to test against the 1.1.1 CI with `[extended test]`?

t8m:
I am sorry about forgetting that. It cherry-picked cleanly so I did merged it to 1.1.1 as well.

PR #11860, opened 2020-05-18
Cast to at least 32 bit unsigned type when shifting left 24 places

opensignature:
##### Checklist

### DIFF
--- a/crypto/pem/pvkfmt.c
+++ b/crypto/pem/pvkfmt.c
@@ -39,7 +39,7 @@ static unsigned int read_ledword(const unsigned char **in)
     ret = *p++;
     ret |= (*p++ << 8);
     ret |= (*p++ << 16);
-    ret |= ((*p++ + 0U) << 24);
+    ret |= ((uint32_t)*p++ << 24);
     *in = p;
     return ret;
 }
--- a/crypto/pem/pvkfmt.c
+++ b/crypto/pem/pvkfmt.c
@@ -39,7 +39,7 @@ static unsigned int read_ledword(const unsigned char **in)
     ret = *p++;
     ret |= (*p++ << 8);
     ret |= (*p++ << 16);
-    ret |= (*p++ << 24);
+    ret |= ((*p++ + 0U) << 24);
     *in = p;
     return ret;
 }
--- a/crypto/pem/pvkfmt.c
+++ b/crypto/pem/pvkfmt.c
@@ -39,7 +39,7 @@ static unsigned int read_ledword(const unsigned char **in)
     ret = *p++;
     ret |= (*p++ << 8);
     ret |= (*p++ << 16);
-    ret |= ((uint32_t)*p++ << 24);
+    ret |= ((unsigned int)*p++ << 24);
     *in = p;
     return ret;
 }
--- a/ssl/t1_trce.c
+++ b/ssl/t1_trce.c
@@ -665,7 +665,7 @@ static int ssl_print_random(BIO *bio, int indent,
 
     if (*pmsglen < 32)
         return 0;
-    tm = ((uint32_t)p[0] << 24) | (p[1] << 16) | (p[2] << 8) | p[3];
+    tm = ((unsigned int)p[0] << 24) | (p[1] << 16) | (p[2] << 8) | p[3];
     p += 4;
     BIO_indent(bio, indent, 80);
     BIO_puts(bio, "Random:\n");
@@ -870,7 +870,7 @@ static int ssl_print_extension(BIO *bio, int indent, int server,
             break;
         if (extlen != 4)
             return 0;
-        max_early_data = ((uint32_t)ext[0] << 24) | (ext[1] << 16) | (ext[2] << 8)
+        max_early_data = ((unsigned int)ext[0] << 24) | (ext[1] << 16) | (ext[2] << 8)
                          | ext[3];
         BIO_indent(bio, indent + 2, 80);
         BIO_printf(bio, "max_early_data=%u\n", max_early_data);
@@ -1366,7 +1366,7 @@ static int ssl_print_ticket(BIO *bio, int indent, const SSL *ssl,
     }
     if (msglen < 4)
         return 0;
-    tick_life = ((uint32_t)msg[0] << 24) | (msg[1] << 16) | (msg[2] << 8) | msg[3];
+    tick_life = ((unsigned int)msg[0] << 24) | (msg[1] << 16) | (msg[2] << 8) | msg[3];
     msglen -= 4;
     msg += 4;
     BIO_indent(bio, indent + 2, 80);
@@ -1377,7 +1377,7 @@ static int ssl_print_ticket(BIO *bio, int indent, const SSL *ssl,
         if (msglen < 4)
             return 0;
         ticket_age_add =
-            ((uint32_t)msg[0] << 24) | (msg[1] << 16) | (msg[2] << 8) | msg[3];
+            ((unsigned int)msg[0] << 24) | (msg[1] << 16) | (msg[2] << 8) | msg[3];
         msglen -= 4;
         msg += 4;
         BIO_indent(bio, indent + 2, 80);
--- a/ssl/t1_trce.c
+++ b/ssl/t1_trce.c
@@ -665,7 +665,7 @@ static int ssl_print_random(BIO *bio, int indent,
 
     if (*pmsglen < 32)
         return 0;
-    tm = ((p[0] + 0U) << 24) | (p[1] << 16) | (p[2] << 8) | p[3];
+    tm = ((uint32_t)p[0] << 24) | (p[1] << 16) | (p[2] << 8) | p[3];
     p += 4;
     BIO_indent(bio, indent, 80);
     BIO_puts(bio, "Random:\n");
@@ -870,7 +870,7 @@ static int ssl_print_extension(BIO *bio, int indent, int server,
             break;
         if (extlen != 4)
             return 0;
-        max_early_data = ((ext[0] + 0U) << 24) | (ext[1] << 16) | (ext[2] << 8)
+        max_early_data = ((uint32_t)ext[0] << 24) | (ext[1] << 16) | (ext[2] << 8)
                          | ext[3];
         BIO_indent(bio, indent + 2, 80);
         BIO_printf(bio, "max_early_data=%u\n", max_early_data);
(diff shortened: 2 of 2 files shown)


bernd-edlinger:
I agree with @t8m this solution is really more ugly.

levitte:
*That's* what it took? Eh gods...

opensignature:
> I agree with @t8m this solution is really more ugly.

"De gustobus non est disputandum"
I prefer to write less code.
An unsigned char (8 bit) can never go overboard if it is under 24 places. It's math baby.

richsalz:
There should be a comment otherwise someone will come along later and remove all the "needless" ```+ 0U``` parts.

bernd-edlinger:
FYI we have now two competing PRs with the same name #11857 and this one.

bernd-edlinger:
> De gustobus non est disputandum

Sic est.

kaduk:
> _That's_ what it took? Eh gods...

You have to get the intermediate calculation (shift) done on an unsigned type.  Whether you do that by explicit cast or creating an arithmetic expression that is forced to promote to the unsigned type by means of an artificial constant with forced type is just a matter of style...

opensignature:
>  is just a matter of style...

I agree. The question is to do 4 intermediate calculations (of which three useless) or only one.
I agree with @richsalz, if we have to insert comments, then better (unsigned int)

levitte:
@paulidale, did you just say that `p[0]` instead of `*p` fixes the issue?

paulidale:
> @paulidale, did you just say that `p[0]` instead of `*p` fixes the issue?

No.  Adding the cast fixes the issue -- the problem has both forms `x[0]` and `*x`.

The cast can be an explicit `(uint32_t)` or a `u` suffix to the shift distance.  I prefer the former.

levitte:
> or a `u` suffix to the shift distance. I prefer the former.

Uhm no, that has already been determined not to be true.  Just for kicks, I tried:

``` C
#include <stdlib.h>

int main()
{
    unsigned char buf[] = { 2, 7, 127, 128 };
    unsigned char *p = buf;
    unsigned int ret;

    ret = p[0];
    ret |= (p[1] << 8);
    ret |= (p[2] << 16);
    ret |= (p[3] << 24u);

    exit(ret);
}
```

``` console
$ cc -fsanitize=undefined foo2.c -o foo2
$ ./foo2
foo2.c:12:18: runtime error: left shift of 128 by 24 places cannot be represented in type 'int'
```

t8m:
I am sorry, but I really prefer my PR.

crypto/pem/pvkfmt.c
```diff
@@ -39,7 +39,7 @@ static unsigned int read_ledword(const unsigned char **in)
     ret = *p++;
     ret |= (*p++ << 8);
     ret |= (*p++ << 16);
-    ret |= (*p++ << 24);
+    ret |= ((uint32_t)*p++ << 24);
```
t8m, line 42: So you cast here to uint32_t and then the result is implicitly cast to unsigned int. That's ugly.

opensignature: you are right :)

t8m:
So everywhere else in the code in similar shift/or cases the explicit casts are done for all the bytes or-ed. I have strong opinion that we should follow this here as well.

opensignature:
> So everywhere else in the code in similar shift/or cases the explicit casts are done for all the bytes or-ed. I have strong opinion that we should follow this here as well.

This does not mean that the past cannot be changed. In my opinion it is useful to reflect on the performance of implicit and explicit casting.
If there is little or no difference then ok explicit, it is more consistent, otherwise it is better not to abuse the explicit one and remove it in all cases where it is not needed.

opensignature:
I did some performance tests, explicit is also faster, I close PR

bernd-edlinger:
> I did some performance tests, explicit is also faster, I close PR

Thanks for testing, I am a bit surpirsed that your compiler does not
elide an exiplict +0U.  Which compiler + version was that?

opensignature:
gcc version 7.5.0

this is test:
`
#include <stdio.h>
#include <time.h>
#include <stdint.h>

struct timespec start, stop;
int main() {
unsigned int a;
unsigned char p;
int b, delta;
clock_gettime(CLOCK_MONOTONIC_RAW, &start);
for (b = 0; b < 200000000; b++) {
p=2;
a = p << 24;
}
clock_gettime(CLOCK_MONOTONIC_RAW, &stop);
delta = (stop.tv_sec - start.tv_sec) * 1000000 + (stop.tv_nsec - start.tv_nsec) / 1000;
printf("Implicit %d\n",delta);

clock_gettime(CLOCK_MONOTONIC_RAW, &start);
for (b = 0; b < 200000000; b++) {
p=2;
a = (unsigned int)p << 24;
}
clock_gettime(CLOCK_MONOTONIC_RAW, &stop);
delta = (stop.tv_sec - start.tv_sec) * 1000000 + (stop.tv_nsec - start.tv_nsec) / 1000;
printf("Explicit %d\n",delta);

clock_gettime(CLOCK_MONOTONIC_RAW, &start);
for (b = 0; b < 200000000; b++) {
p=2;
a = (p + 0U) << 24;
}
clock_gettime(CLOCK_MONOTONIC_RAW, &stop);
delta = (stop.tv_sec - start.tv_sec) * 1000000 + (stop.tv_nsec - start.tv_nsec) / 1000;
printf("+0U %d\n",delta);
}
`

kaduk:
I share Bernd's surprise.
@opensignature what was the compiler invocation?  E.g., was any `-O2` or similar passed?

bernd-edlinger:
When I look at the test case, I would change p to be volatile,
otherwise an expression like `a = (p + 0U) << 24);" is evaluated as a
constant.  and it will probably moved out if the loop....

paulidale:
There is a `*` in the expression like that here.

bernd-edlinger:
aehm, what did you mean?

levitte:
I suppose that the cast makes sure that the runtime check isn't emitted, so of course it's faster.

bernd-edlinger:
if you cast a constant the value can be folded.

bernd-edlinger:
you can trust me, I am a gcc maintainer :-)

bernd-edlinger:
and, if you add 0 to something, expect the compiler to remove that, unless
volatile is involved.
and if you do that in a loop, expect the compiler to move constant evaluations
out of the loop.

bernd-edlinger:
again, volatile can prevent this or a memory barrier of course.

bernd-edlinger:
> I suppose that the cast makes sure that the runtime check isn't emitted, so of course it's faster

so, when I compile the test case above with GCC: (GNU) 11.0.0 20200511 (experimental)
at -O3 I get the following assembler:
```
main:
.LFB11:
	.cfi_startproc
	pushq	%rbx
	.cfi_def_cfa_offset 16
	.cfi_offset 3, -16
	movl	$start, %esi
	movl	$4, %edi
	movabsq	$2361183241434822607, %rbx
	call	clock_gettime
	movl	$stop, %esi
	movl	$4, %edi
	call	clock_gettime
	movq	stop+8(%rip), %rcx
	movl	$.LC0, %edi
	subq	start+8(%rip), %rcx
	movq	%rcx, %rax
	movq	stop(%rip), %rsi
	sarq	$63, %rcx
	subq	start(%rip), %rsi
	imulq	%rbx
	xorl	%eax, %eax
	imull	$1000000, %esi, %esi
	sarq	$7, %rdx
	subq	%rcx, %rdx
	addl	%edx, %esi
	call	printf
	movl	$start, %esi
	movl	$4, %edi
	call	clock_gettime
	movl	$stop, %esi
	movl	$4, %edi
	call	clock_gettime
	movq	stop+8(%rip), %rcx
	movl	$.LC1, %edi
	subq	start+8(%rip), %rcx
	movq	%rcx, %rax
	movq	stop(%rip), %rsi
	sarq	$63, %rcx
	subq	start(%rip), %rsi
	imulq	%rbx
	xorl	%eax, %eax
	imull	$1000000, %esi, %esi
	sarq	$7, %rdx
	subq	%rcx, %rdx
	addl	%edx, %esi
	call	printf
	movl	$start, %esi
	movl	$4, %edi
	call	clock_gettime
	movl	$stop, %esi
	movl	$4, %edi
	call	clock_gettime
	movq	stop+8(%rip), %rcx
	movl	$.LC2, %edi
	subq	start+8(%rip), %rcx
	movq	%rcx, %rax
	movq	stop(%rip), %rsi
	sarq	$63, %rcx
	subq	start(%rip), %rsi
	imulq	%rbx
	imull	$1000000, %esi, %esi
	sarq	$7, %rdx
	subq	%rcx, %rdx
	xorl	%eax, %eax
	addl	%edx, %esi
	call	printf
	xorl	%eax, %eax
	popq	%rbx
	.cfi_def_cfa_offset 8
	ret
	.cfi_endproc
```
You see no branch whatsoever, the loops are optimized away.
```
$ ./a.out 
Implicit 1
Explicit 0
+0U 0
```

paulidale:
What I meant was the only instance in the changes is `((uint32_t)*p++ << 24)` which is not a constant expression because the pointer is being de-referenced.

Agreed, the sample code above includes it without a de-reference and the optimiser will have fun -- I'm kind of surprised that the output is so long to be honest.

bernd-edlinger:
Yes, I also only asked because if the expression `((uint32_t)*p++ << 24)`
would generate different code than `(*p++ + 0U) << 24` that would have
been a compiler error, but gcc 7.5 is out of support now, so even if that
may have have emitted different code, we can't do nothing about it.

opensignature:
```
  10:test.c         **** ret = ((unsigned int)*p++ << 24);
  26              		.loc 1 10 0
  27 0023 488B45E8 		movq	-24(%rbp), %rax
  28 0027 488D5001 		leaq	1(%rax), %rdx
  29 002b 488955E8 		movq	%rdx, -24(%rbp)
  30 002f 0FB600   		movzbl	(%rax), %eax
  31 0032 0FB6C0   		movzbl	%al, %eax
  32 0035 C1E018   		sall	$24, %eax
  33 0038 8945E4   		movl	%eax, -28(%rbp)
  34 003b B8000000 		movl	$0, %eax
```
```
  10:test.c        **** ret = (*p++ + 0U) << 24;
  26                            .loc 1 10 0
  27 0023 488B45E8              movq    -24(%rbp), %rax
  28 0027 488D5001              leaq    1(%rax), %rdx
  29 002b 488955E8              movq    %rdx, -24(%rbp)
  30 002f 0FB600                movzbl  (%rax), %eax
  31 0032 0FB6C0                movzbl  %al, %eax
  32 0035 C1E018                sall    $24, %eax
  33 0038 8945E4                movl    %eax, -28(%rbp)
  34 003b B8000000              movl    $0, %eax
```

opensignature:
> I share Bernd's surprise.
> @opensignature what was the compiler invocation? E.g., was any `-O2` or similar passed?

I put -O0 to cancel the optimization.
I could also have added + rand() at a = p << 24 to avoid it.
However the result does not change because, as seen above the assembler code is identical.
I didn't understand which compilation error you are referring to.
I also tried with gcc 11, but it doesn't give me any errors.

bernd-edlinger:
> I didn't understand which compilation error you are referring to.

I meant a "missing" optimization (with -O2 or -O3) since we do only care
about the speed of optimized code, and the optimization level is usually -O3
for production code.
But as your experiment shows, even with -O0 sometimes surprising
optimizations do happen.

PR #11867, opened 2020-05-19
Avoid undefined behavior in pvkfmt.c

bernd-edlinger:
Alternate more lean fix for #11853

this time with [extended tests]
in the commit message, so we can actually see the effect.

##### Checklist

### DIFF
--- a/crypto/pem/pvkfmt.c
+++ b/crypto/pem/pvkfmt.c
@@ -36,10 +36,10 @@ static unsigned int read_ledword(const unsigned char **in)
 {
     const unsigned char *p = *in;
     unsigned int ret;
-    ret = *p++;
-    ret |= (*p++ << 8);
-    ret |= (*p++ << 16);
-    ret |= (*p++ << 24);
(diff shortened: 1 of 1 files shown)


t8m:
This is identical to the first commit of my PR.

bernd-edlinger:
> This is identical to the first commit of my PR.

nice, but did you also add the [extended tests] ?

bernd-edlinger:
As you see all travis tests are green with this simple PR.
So the commit message Fixes #11853 is misleading when
you fix more than 11853 ?

levitte:
> So the commit message Fixes #11853 is misleading when you fix more than 11853 ?

That's only attached to one of the commits in #11857, so it's fine.  It's true that the PR description is a little bit confusing, but that can be changed to say that one of the commits fixes #11853.  In the end, for auto-closing purposes, github pays attention to commit messages, not to PR descriptions.

bernd-edlinger:
Since @t8m did the same commit now, with a slightly better commit message,
I took his commit directly in this PR.

PR #11869, opened 2020-05-19
rsa_padding_add_PKCS1_OAEP_mgf1_with_libctx(): fix check of |md|

### LINKED ISSUE #11865: Openssl 3.0.0 alpha3 fips provider: OAEP encryption fails
Observed with Openssl 3.0.0-alpha3 NULL library context configured to use fips provider only.

OAEP encryption always fails with error during EVP_PKEY_encrypt.  (Btw decryption succeeds)

```
#include <openssl/provider.h>
#include <openssl/crypto.h>
#include <openssl/evp.h>
#include <openssl/err.h>
#include <openssl/rsa.h>


{
   EVP_PKEY *pkey = NULL;
   EVP_PKEY_CTX *gen_ctx;
   EVP_PKEY_CTX *enc_ctx;
   char *pval; 
   size_t outlen;
   char *data = "hello";
   int ret;
   char *out;

    printf("Generating RSA key\n");
    gen_ctx = EVP_PKEY_CTX_new_id(NID_rsaEncryption, NULL);   
        
    printf("... done generating RSA key\n");


    printf("Setting up for OAEP encrypt\n");

    pval = OPENSSL_malloc(3);
    if (pval == NULL) goto end;

    enc_ctx = EVP_PKEY_CTX_new(pkey, NULL);
    if (enc_ctx == NULL) goto end;

    if (EVP_PKEY_encrypt_init(enc_ctx) <= 0) goto end;

    if (EVP_PKEY_CTX_set_rsa_padding(enc_ctx, RSA_PKCS1_OAEP_PADDING) <= 0) goto end;
    
    if (EVP_PKEY_CTX_set_rsa_oaep_md(enc_ctx, EVP_sha256()) <= 0) goto end;
    
    if (EVP_PKEY_CTX_set_rsa_mgf1_md(enc_ctx, EVP_sha1()) <= 0) goto end;
    
    if (EVP_PKEY_CTX_set0_rsa_oaep_label(enc_ctx, pval, 3) <= 0) goto end;
    
    if (EVP_PKEY_encrypt(enc_ctx, NULL, &outlen, data, 5) <= 0) 
        goto end;
    
    out = OPENSSL_malloc(outlen);
    if (out == NULL) goto end;
    
    printf("... done\n. Doing the encrypt\n");
    if (EVP_PKEY_encrypt(enc_ctx, out, &outlen, data, 5) <= 0) 
        goto end;
    printf(".... done. Succeeded (%ld bytes output) \n", outlen);
    
end:
   printf ("end:\n");
   ERR_print_errors_fp(stderr);
   EVP_PKEY_free(pkey);
   EVP_PKEY_CTX_free(gen_ctx);
   EVP_PKEY_CTX_free(enc_ctx);
   return 1;
}
```
Output for fips provider:
```
default provider available: 0
fips provider available: 1
Generating RSA key
... done generating RSA key
Setting up for OAEP encrypt
... done
. Doing the encrypt
end:
80:7B:1B:81:C6:7F:00:00:error:rsa routines:(unknown function):passed a null parameter:../crypto/rsa/rsa_oaep.c:74:
```
Output (success) for default provider:
```
default provider available: 1
fips provider available: 0
Generating RSA key
... done generating RSA key
Setting up for OAEP encrypt
... done
. Doing the encrypt
.... done. Succeeded (256 bytes output) 
end:
```

levitte:
In the FIPS module, the code as written generate an unconditional
error.

Fixes #11865

### DIFF
--- a/crypto/rsa/rsa_oaep.c
+++ b/crypto/rsa/rsa_oaep.c
@@ -67,13 +67,14 @@ int rsa_padding_add_PKCS1_OAEP_mgf1_with_libctx(OPENSSL_CTX *libctx,
     unsigned char seedmask[EVP_MAX_MD_SIZE];
     int mdlen, dbmask_len = 0;
 
+    if (md == NULL) {
 #ifndef FIPS_MODULE
-    if (md == NULL)
         md = EVP_sha1();
 #else
         RSAerr(0, ERR_R_PASSED_NULL_PARAMETER);
         return 0;
 #endif
+    }
     if (mgf1md == NULL)
         mgf1md = md;
 


mattcaswell:
Why didn't our tests pick this issue up?

levitte:
> Why didn't our tests pick this issue up?

I have *no* idea.  Please someone investigate

bernd-edlinger:
yeah I will look.

bernd-edlinger:
This is test test case for the rsa pading:
```
$ ./util/shlib_wrap.sh test/rsa_test
1..4
    # Subtest: test_rsa_pkcs1
    1..3
    ok 1 - iteration 1
    ok 2 - iteration 2
    ok 3 - iteration 3
ok 1 - test_rsa_pkcs1
    # Subtest: test_rsa_sslv23
    1..3
    ok 1 - iteration 1
    ok 2 - iteration 2
    ok 3 - iteration 3
ok 2 - test_rsa_sslv23
    # Subtest: test_rsa_oaep
    1..3
    ok 1 - iteration 1
    ok 2 - iteration 2
    ok 3 - iteration 3
ok 3 - test_rsa_oaep
    # Subtest: test_rsa_security_bit
    1..12
    ok 1 - iteration 1
    ok 2 - iteration 2
    ok 3 - iteration 3
    ok 4 - iteration 4
    ok 5 - iteration 5
    ok 6 - iteration 6
    ok 7 - iteration 7
    ok 8 - iteration 8
    ok 9 - iteration 9
    ok 10 - iteration 10
    ok 11 - iteration 11
    ok 12 - iteration 12
ok 4 - test_rsa_security_bit
```
note I even added a test for SSLV23_PADDING when I
reworked that constant time RSA padding,
I would say it should be impossible how this test succeeds.

mattcaswell:
> I would say it should be impossible how this test succeeds.

My guess is its only testing the default provider not the FIPS provider (which is where this issue occurs).

mattcaswell:
> My guess is its only testing the default provider not the FIPS provider (which is where this issue occurs).

Although I would have expected evp_test to test this padding - and that should cover both providers.

bernd-edlinger:
and this compiles, so FIPS_MODULE is never defined:
```C
diff --git a/crypto/rsa/rsa_oaep.c b/crypto/rsa/rsa_oaep.c
index 8ffde9f..fdfdf92 100644
--- a/crypto/rsa/rsa_oaep.c
+++ b/crypto/rsa/rsa_oaep.c
@@ -71,6 +71,7 @@ int rsa_padding_add_PKCS1_OAEP_mgf1_with_libctx(OPENSSL_CTX *libctx,
     if (md == NULL)
         md = EVP_sha1();
 #else
+ssss
         RSAerr(0, ERR_R_PASSED_NULL_PARAMETER);
         return 0;
 #endif
```

bernd-edlinger:
... Now I'm lost, the FIPS code is probably dead alltogrther,
and should be removed.

levitte:
I have no idea what you're doing, @bernd-edlinger. When I add an "ssss" line like you fid, my compiler does complain!

bernd-edlinger:
hmm.
I have master  e9e7b5d
here, and the code in question does compile perfectly for me.
```
diff --git a/crypto/rsa/rsa_oaep.c b/crypto/rsa/rsa_oaep.c
index 8ffde9f..05dc5d8 100644
--- a/crypto/rsa/rsa_oaep.c
+++ b/crypto/rsa/rsa_oaep.c
@@ -67,10 +67,12 @@ int rsa_padding_add_PKCS1_OAEP_mgf1_with_libctx(OPENSSL_CTX *libctx,
     unsigned char seedmask[EVP_MAX_MD_SIZE];
     int mdlen, dbmask_len = 0;
 
+asm("int3");
 #ifndef FIPS_MODULE
     if (md == NULL)
         md = EVP_sha1();
 #else
+ssss
         RSAerr(0, ERR_R_PASSED_NULL_PARAMETER);
         return 0;
 #endif


make
gcc  -I. -Iinclude -Iproviders/common/include -Iproviders/implementations/include -Icrypto/include  -DSTATIC_LEGACY -pthread -m64 -Wall -O0 -g -DOPENSSL_USE_NODELETE -DL_ENDIAN -DOPENSSLDIR="\"/usr/local/ssl\"" -DENGINESDIR="\"/usr/local/lib/engines-3\"" -DMODULESDIR="\"/usr/local/lib/ossl-modules\"" -DOPENSSL_BUILDING_OPENSSL  -MMD -MF crypto/rsa/libcrypto-lib-rsa_oaep.d.tmp -MT crypto/rsa/libcrypto-lib-rsa_oaep.o -c -o crypto/rsa/libcrypto-lib-rsa_oaep.o crypto/rsa/rsa_oaep.c
```
but the int3 makes the test/rsa_test fail:
```
$ test/rsa_test 
1..4
    # Subtest: test_rsa_pkcs1
    1..3
    ok 1 - iteration 1
    ok 2 - iteration 2
    ok 3 - iteration 3
ok 1 - test_rsa_pkcs1
    # Subtest: test_rsa_sslv23
    1..3
    ok 1 - iteration 1
    ok 2 - iteration 2
    ok 3 - iteration 3
ok 2 - test_rsa_sslv23
    # Subtest: test_rsa_oaep
    1..3
Trace/Breakpoint ausgelöst (Speicherabzug geschrieben)
```
I have configured with `./config -d no-asm no-pic no-dso`

levitte:
With `no-pic`, you don't get a FIPS module...

richsalz:
> With no-pic, you don't get a FIPS module...

Perhaps Configure should output a message about that.

bernd-edlinger:
Yeah, if I want to debug something, I do usually try to build everthing statically,
debugging shared object is of course more fun....

bernd-edlinger:
OTOH, when there is not FIPS which test case should complain about that ?

bernd-edlinger:
`03-test_fipsinstall.t .............. skipped: Test only supported in a fips build`
that is probably the hint I overlooked :-)

bernd-edlinger:
but how can it be, that the no-pic version with int3 in the oaep code path
has only
```
15-test_rsa.t ...................... Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/10 subtests 
	(less 2 skipped subtests: 7 okay)
80-test_cms.t ...................... Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/7 subtests 
```
but @mattcaswell you said that the evp test should also trigger?
why doesn't it trigger here?

bernd-edlinger:
Okay, now shared object: `./config -d`
```
diff --git a/crypto/rsa/rsa_oaep.c b/crypto/rsa/rsa_oaep.c
index 8ffde9f..2a0e428 100644
--- a/crypto/rsa/rsa_oaep.c
+++ b/crypto/rsa/rsa_oaep.c
@@ -71,6 +71,7 @@ int rsa_padding_add_PKCS1_OAEP_mgf1_with_libctx(OPENSSL_CTX *libctx,
     if (md == NULL)
         md = EVP_sha1();
 #else
+asm("int3");
         RSAerr(0, ERR_R_PASSED_NULL_PARAMETER);
         return 0;
 #endif
```
@levitte yes, the ssss version does not compile but int3 does not cause any test failures.
so again the result is this is dead code.
the code in crypto/rsa/rsa_oaep.c is only used for cms  and the deprecated RSA API,
and those are not available in FIPS.

That explains why this defect cant be found in our tests.

bernd-edlinger:
The same duplication I already saw with the AES code.
The new code is in the providers, but the old code is still
needed for the CMS and probably engines to use.
Nobody can maintain a code base that is designed that way.
Which alpha release will remove the duplication ?

bernd-edlinger:
Hmm, is it possible that the deprecated legacy AES API is
also not using the provider @slontis ?

levitte:
It's possible that our OAEP tests aren't performed with the FIPS module. That doesn't make the code dead.

bernd-edlinger:
I wouldn't call it a good sign either.
How is this code supposed to be tested?

mattcaswell:
> How is this code supposed to be tested?

There are a number of OAEP tests in test/recipes/30-test_evp_data/evppkey.txt. The EVP tests are supposed to be run by both the default provider and the FIPS provider, so I would have expected those tests to pick up this issue. I haven't investigated why it didn't.

bernd-edlinger:
Ah, the answer is easy:
```
grep crypt.*OAEP 30-test_evp_data/evppkey.txt
Decrypt=RSA-OAEP-1
Decrypt=RSA-OAEP-1
Decrypt=RSA-OAEP-1
Decrypt=RSA-OAEP-1
Decrypt=RSA-OAEP-1
Decrypt=RSA-OAEP-1
Decrypt=RSA-OAEP-2
Decrypt=RSA-OAEP-2
Decrypt=RSA-OAEP-2
Decrypt=RSA-OAEP-2
Decrypt=RSA-OAEP-2
Decrypt=RSA-OAEP-2
Decrypt=RSA-OAEP-3
Decrypt=RSA-OAEP-3
Decrypt=RSA-OAEP-3
Decrypt=RSA-OAEP-3
Decrypt=RSA-OAEP-3
Decrypt=RSA-OAEP-3
Decrypt=RSA-OAEP-4
Decrypt=RSA-OAEP-4
Decrypt=RSA-OAEP-4
Decrypt=RSA-OAEP-4
Decrypt=RSA-OAEP-4
Decrypt=RSA-OAEP-4
Decrypt=RSA-OAEP-5
Decrypt=RSA-OAEP-5
Decrypt=RSA-OAEP-5
Decrypt=RSA-OAEP-5
Decrypt=RSA-OAEP-5
Decrypt=RSA-OAEP-5
Decrypt=RSA-OAEP-6
Decrypt=RSA-OAEP-6
Decrypt=RSA-OAEP-6
Decrypt=RSA-OAEP-6
Decrypt=RSA-OAEP-6
Decrypt=RSA-OAEP-6
Decrypt=RSA-OAEP-7
Decrypt=RSA-OAEP-7
Decrypt=RSA-OAEP-7
Decrypt=RSA-OAEP-7
Decrypt=RSA-OAEP-7
Decrypt=RSA-OAEP-7
Decrypt=RSA-OAEP-8
Decrypt=RSA-OAEP-8
Decrypt=RSA-OAEP-8
Decrypt=RSA-OAEP-8
Decrypt=RSA-OAEP-8
Decrypt=RSA-OAEP-8
Decrypt=RSA-OAEP-9
Decrypt=RSA-OAEP-9
Decrypt=RSA-OAEP-9
Decrypt=RSA-OAEP-9
Decrypt=RSA-OAEP-9
Decrypt=RSA-OAEP-9
Decrypt=RSA-OAEP-10
Decrypt=RSA-OAEP-10
Decrypt=RSA-OAEP-10
Decrypt=RSA-OAEP-10
Decrypt=RSA-OAEP-10
Decrypt=RSA-OAEP-10
```

bernd-edlinger:
You can't have a known answer test when the encyption is randomized.

mattcaswell:
> You can't have a known answer test when the encyption is randomized.

We should perhaps add a roundtrip (i.e. encrypt then decrypt) test in evp_extra_test for OAEP.

levitte:
Could we add encryption stanzas for evp_test?

mattcaswell:
> Could we add encryption stanzas for evp_test?

Well, I think the problem there is as @bernd-edlinger points out "You can't have a known answer test when the encyption is randomized."

slontis:
> "You can't have a known answer test when the encyption is randomized."

You can if you supply the entropy.

mattcaswell:
> You can if you supply the entropy.

Do we have that capability in evp_test?

slontis:
not yet :)

bernd-edlinger:
now, I am curious, are there FIPS test vectors for that?

paulidale:
> Do we have that capability in evp_test?

We do at the moment but not in a nice manner.  The DRBG tests implement something that is adequate.

The changes in #11682 will break what is currently done but the capability needs to be available and it is on the TODO list.

slontis:
> now, I am curious, are there FIPS test vectors for that?

I cant see any..
Interop tests with another toolkit would probably be better.
(Or debug another toolkit to grab its internal data).

levitte:
Merged

e637d47c9122d74d4f3a40a8cbe867de29468ba3 rsa_padding_add_PKCS1_OAEP_mgf1_with_libctx(): fix check of |md|

PR #11873, opened 2020-05-19
APPS: Make it possible to load_cert() from stdin again

### LINKED ISSUE #11871: x509 can't read from stdin
The man page for x509 says:
>        -in filename
>            This specifies the input filename to read a certificate from or
>            standard input if this option is not specified.

The following command works with OpenSSL 3.0.0-alpha2:
```
cat cert.pem | openssl x509 -text
```
Using the current master branch on GitHub, the same command returns an error:
```
Could not open file or uri  for loading Certificate
80:1B:4D:82:29:7F:00:00:error:system library:file_open:No such file or directory:crypto/store/loader_file.c:882:calling stat()
Unable to load Certificate
```
The current master branch on GitHub does work, however, if the command line is changed to:
```
cat cert.pem | openssl x509 -text -in /dev/stdin
```
My guess is that PR #11755 is the reason that the command no longer reads from standard input if the `-in` option is not specified.

levitte:
Fixes #11871

### DIFF
--- a/apps/lib/apps.c
+++ b/apps/lib/apps.c
@@ -453,7 +453,7 @@ X509 *load_cert_pass(const char *uri, int maybe_stdin,
 /* the format parameter is meanwhile not needed anymore and thus ignored */
 X509 *load_cert(const char *uri, int format, const char *desc)
 {
-    return load_cert_pass(uri, 0, NULL, desc);
+    return load_cert_pass(uri, 1, NULL, desc);
 }
 
 /* the format parameter is meanwhile not needed anymore and thus ignored */
--- a/apps/lib/apps.c
+++ b/apps/lib/apps.c
@@ -438,10 +438,6 @@ X509 *load_cert_pass(const char *uri, int maybe_stdin,
 
     if (desc == NULL)
         desc = "certificate";
-    if (uri == NULL) {
-        unbuffer(stdin);
-        uri = "";
-    }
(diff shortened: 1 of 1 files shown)


dcooper16:
Hi @levitte,

Thanks for looking into this so quickly. I tried the change in this PR and it didn't work for me. In order to make things work I also needed to change load_cert_pass():
```
    if (uri == NULL) {
        unbuffer(stdin);
        uri = "";
    }
```
It seems that uri needs to be NULL rather than "" in order for `load_key_cert_crl()` to read from stdin. Also, while I don't know what `unbuffer(stdin);` does, I wonder if it is necessary, since `load_key_cert_crl()` would also make that same call.

vdukhovni:
> It seems that uri needs to be NULL rather than "" in order for `load_key_cert_crl()` to read from stdin

Correct, the line setting `uri = ""` needs to be removed.

levitte:
@dcooper16, thanks for your feedback.  I finally got my focus back here, and could quickly confirm that you're exactly right.  I'm solving it by removing the lines you mention.

levitte:
Merged

0808a75e4db421f0f33d866b0abfe10db4f472f2 APPS: Make it possible to load_cert() from stdin again
f5e23fe88904fb5309e5b199bf5254b06dd4f883 APPS: Remove unnecessary NULL check of uri in load_cert_pass()

PR #11874, opened 2020-05-19
deprecate EC_POINT_make_affine and EC_POINTs_make_affine

bbbrumley:
Related to what's going on in #8332.

* Deprecate `EC_POINTs_make_affine`. I have only ever seen one caller to this -- internally in `ec_mult.c`.
* Deprecate `EC_POINT_make_affine`. Since there is no longer any visibility inside an `EC_POINT` -- there is no noticeable state change for linking applications. Applications no longer reach in like `P->X` to get at the `BIGNUM`, but use the API like `EC_POINT_get_affine_coordinates` to fetch them -- where this conversion happens internally by the `EC_METHOD` automatically.

Apologies -- I fat fingered the `CHANGES.md` in the last PR. My glasses have been in my work office since quarantine started and the alignment was a little off :nerd_face:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,12 +23,18 @@ OpenSSL 3.0
 
 ### Changes between 1.1.1 and 3.0 [xx XXX xxxx]
 
-* Deprecated EC_GROUP_precompute_mult(), EC_GROUP_have_precompute_mult(), and
-  EC_KEY_precompute_mult() These functions are not widely used and applications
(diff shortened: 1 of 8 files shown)


romen:
Travis failures are unrelated: some timeout on Xenial builds when fetching required packages with `apt`

t8m:
Merged to master. Thank you for the patch!

PR #11876, opened 2020-05-19
There is no -signreq option in CA.pl

mettacrawler:
CLA: trivial

##### Checklist

### DIFF
--- a/doc/man1/CA.pl.pod
+++ b/doc/man1/CA.pl.pod
@@ -113,7 +113,7 @@ written to standard output.
 
 =item B<-signCA>
 
-This option is the same as the B<-signreq> option except it uses the
+This option is the same as the B<-sign> option except it uses the
 configuration file section B<v3_ca> and so makes the signed request a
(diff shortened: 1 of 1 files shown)


t8m:
Unfortunately for 1.1.1 there should be a separate PR because the -signreq is mentioned at one additional place there.

t8m:
Merged to master, would you like to create a PR for the 1.1.1 branch?

mettacrawler:
@t8m is it normal to close a branch with un-merged commits?

t8m:
It is actually merged, the e1c6f76 is your commit, rebased and with the commit message adjusted so it contains the reviewed-by and link to the PR.

mettacrawler:
@t8m PR for 1.1.1 https://github.com/openssl/openssl/pull/11897

richsalz:
> is it normal to close a branch with un-merged commits?

yes, because the master source for openssl is not on github, but on an openssl server which the pushes to github as a read-only copy.

PR #11878, opened 2020-05-19
Adjust length of some strncpy() calls

romen:
This fixes warnings detected by `-Wstringop-truncation`.

### DIFF
--- a/crypto/LPdir_unix.c
+++ b/crypto/LPdir_unix.c
@@ -121,7 +121,7 @@ const char *LP_find_file(LP_DIR_CTX **ctx, const char *directory)
 
 #ifdef __VMS
     strncpy((*ctx)->previous_entry_name, (*ctx)->entry_name,
-            sizeof((*ctx)->previous_entry_name) - 1);
+            sizeof((*ctx)->previous_entry_name));
 
  again:
 #endif
--- a/crypto/LPdir_unix.c
+++ b/crypto/LPdir_unix.c
@@ -121,7 +121,7 @@ const char *LP_find_file(LP_DIR_CTX **ctx, const char *directory)
 
 #ifdef __VMS
     strncpy((*ctx)->previous_entry_name, (*ctx)->entry_name,
-            sizeof((*ctx)->previous_entry_name));
+            sizeof((*ctx)->previous_entry_name) - 1);
 
  again:
 #endif
--- a/crypto/x509/v3_alt.c
+++ b/crypto/x509/v3_alt.c
@@ -128,7 +128,7 @@ STACK_OF(CONF_VALUE) *i2v_GENERAL_NAME(X509V3_EXT_METHOD *method,
                 BIO_snprintf(othername, sizeof(othername), "othername: %s:",
                              oline);
             else
-                strncpy(othername, "othername:", sizeof(othername) - 1);
+                OPENSSL_strlcpy(othername, "othername:", sizeof(othername));
 
             /* check if the value is something printable */
             if (gen->d.otherName->value->type == V_ASN1_IA5STRING) {
--- a/crypto/x509/v3_alt.c
+++ b/crypto/x509/v3_alt.c
@@ -128,7 +128,7 @@ STACK_OF(CONF_VALUE) *i2v_GENERAL_NAME(X509V3_EXT_METHOD *method,
                 BIO_snprintf(othername, sizeof(othername), "othername: %s:",
                              oline);
             else
-                strncpy(othername, "othername:", sizeof(othername));
+                strncpy(othername, "othername:", sizeof(othername) - 1);
 
             /* check if the value is something printable */
             if (gen->d.otherName->value->type == V_ASN1_IA5STRING) {
--- a/providers/implementations/signature/rsa.c
+++ b/providers/implementations/signature/rsa.c
@@ -336,9 +336,20 @@ static int rsa_signature_init(void *vprsactx, void *vrsa, int operation)
                     return 0;
                 }
 
-                strncpy(prsactx->mdname, mdname, sizeof(prsactx->mdname) - 1);
-                strncpy(prsactx->mgf1_mdname, mgf1mdname,
-                        sizeof(prsactx->mgf1_mdname) - 1);
+                if (OPENSSL_strlcpy(prsactx->mdname, mdname,
+                                    sizeof(prsactx->mdname))
+                        >= sizeof(prsactx->mdname)) {
+                    ERR_raise_data(ERR_LIB_PROV, PROV_R_INVALID_DIGEST,
+                                   "hash algorithm name too long");
+                    return 0;
+                }
+                if (OPENSSL_strlcpy(prsactx->mgf1_mdname, mgf1mdname,
+                                sizeof(prsactx->mgf1_mdname))
+                        >= sizeof(prsactx->mgf1_mdname)) {
+                    ERR_raise_data(ERR_LIB_PROV, PROV_R_INVALID_DIGEST,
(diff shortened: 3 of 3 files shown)


crypto/LPdir_unix.c
```diff
@@ -121,7 +121,7 @@ const char *LP_find_file(LP_DIR_CTX **ctx, const char *directory)
 
 #ifdef __VMS
     strncpy((*ctx)->previous_entry_name, (*ctx)->entry_name,
-            sizeof((*ctx)->previous_entry_name));
+            sizeof((*ctx)->previous_entry_name) - 1);
```
romen, line 124: This one was detected by `grep`ping for `strncpy` usage combined with `sizeof`, and I cannot test compilation on VMS.

romen: If accepted, this change would also apply to `1.1.1`, the others are `master`-specific.

I would open a separate PR for this.

kroeckx: This change actually looks wrong, the old code was correct. Both previous_entry_name and entry_name are the same size.

crypto/x509/v3_alt.c
```diff
@@ -128,7 +128,7 @@ STACK_OF(CONF_VALUE) *i2v_GENERAL_NAME(X509V3_EXT_METHOD *method,
                 BIO_snprintf(othername, sizeof(othername), "othername: %s:",
                              oline);
             else
-                strncpy(othername, "othername:", sizeof(othername));
+                strncpy(othername, "othername:", sizeof(othername) - 1);
```
kroeckx, line 131: Since it's: char othername[300], this doesn't seem to fix anything. But if you want to fix something, you would need to make sure it's actually NUL terminated.

romen: `strncpy` [should](https://en.cppreference.com/w/c/string/byte/strncpy)

> Zeroes out the rest of the destination buffer

so there shouldn't be a need to explicitly NUL-terminate it.

Or am I missing your point about it?

---

I can see how there is no real risk of exceeding 300 when copying the shorter `"othername:"`, but conceptually shouldn't we just avoid the anti-pattern `strncpy(dst_buf, src_buf, sizeof(dst_buf))` as error prone?

providers/implementations/signature/rsa.c
```diff
@@ -336,9 +336,9 @@ static int rsa_signature_init(void *vprsactx, void *vrsa, int operation)
                     return 0;
                 }
 
-                strncpy(prsactx->mdname, mdname, sizeof(prsactx->mdname));
+                strncpy(prsactx->mdname, mdname, sizeof(prsactx->mdname) - 1);
                 strncpy(prsactx->mgf1_mdname, mgf1mdname,
-                        sizeof(prsactx->mgf1_mdname));
+                        sizeof(prsactx->mgf1_mdname) - 1);
                 prsactx->saltlen = min_saltlen;
```
kroeckx, line 342: This seems to be a change that in practice doesn't actually fix anything. mdname seems to be OSSL_MAX_NAME_SIZE, 50, and the longest we're going to put into currently looks like it's going to be 12+1.

If you want to future proof this, I think we need to check that it actually fits in it, and that if it's too small generate an error. If we don't mind truncation, I suggest you just replace strncpy by OPENSSL_strlcpy() and keep the size.

crypto/x509/v3_alt.c
kroeckx: The point is, if you use sizeof()-1, you don't actually zero terminate it if your source string was >= sizeof()-1. You should make sure it's terminated by writing the '\0' to [sizeof()-1], because you didn't write anything to that. Or you could just use strlcpy().

crypto/LPdir_unix.c
romen: ok, reverting!

Thanks for the feedback @kroeckx I was indeed quite misguided working on this the way I approached it!

crypto/x509/v3_alt.c
romen: Right! Thanks for the feedback!

crypto/LPdir_unix.c
bernd-edlinger: but you did not push the revert?

bernd-edlinger:
@romen we have a better function for that purpose:
OPENSSL_strlcpy
I use it whenever these warnings turn up,
It guarantees the zero-termination and avoids the pitfalls.

But it is nice that someone other than me jumps in and
fixes those warnings.

crypto/LPdir_unix.c
romen: > but you did not push the revert?
> but you did not push the revert?

Yeah, I got lost while trying to figure out why I was seeing other failures, before realizing that those are happening for me even before these changes!

Pushed now a fixup commit, changes to this file are dropped.

crypto/x509/v3_alt.c
romen: In the latest fixup commit I just replaced `strncpy()` with `OPENSSL_strlcpy()`, I am not checking the returned length as it seems like we were already ok with truncating it here.

providers/implementations/signature/rsa.c
romen: For these ones I am using `OPENSSL_strlcpy()` and raising an error if truncation happens.

providers/implementations/signature/rsa.c
```diff
@@ -336,9 +336,20 @@ static int rsa_signature_init(void *vprsactx, void *vrsa, int operation)
                     return 0;
                 }
 
-                strncpy(prsactx->mdname, mdname, sizeof(prsactx->mdname));
-                strncpy(prsactx->mgf1_mdname, mgf1mdname,
-                        sizeof(prsactx->mgf1_mdname));
+                if (OPENSSL_strlcpy(prsactx->mdname, mdname,
+                                    sizeof(prsactx->mdname))
+                        >= sizeof(prsactx->mdname)) {
```
romen, line 341: I am unsure about how to split the line in this `if`, is the current solution acceptable or do you have a better recommendation?

bernd-edlinger: Hmm. I ususally tend to indent only lines beginning with || and && by another 4 spaces,
otherwise this looks like
if (a
    op b)
so if a is this rather long strcpy I would not indent the >= by 4 spaces.
But it is sometimes better not to do too much in the if-expression.
That is a general remark, there are much more complicated expressions,
which have no reason for being in an if construct

bernd-edlinger: one semantic difference between strncpy and our OPENSSL_strlcpy is,
that strncpy does not need a zero-termination in the input, while strlcpy
does need it (but only for the return value)

romen: I went ahead and reformatted it using an auxiliary `size_t` variable to store the result of `OPENSSL_strlcpy()` which then I compare with the size of the buffer.

Seems cleaner now, what do you think?

bernd-edlinger: Yeah, I like it better too.

romen:
I noticed that in `providers/implementation/signature/rsa.c` there were already other uses of `OPENSSL_strlcpy()` but we weren't checking for truncation.

With https://github.com/openssl/openssl/pull/11878/commits/aaa9c622f2af6ffd8e2fe3d093f797cf4a7933b3 I tried to address that.

romen:
@bernd-edlinger @kroeckx would you be available to review this again after the last commits?

romen:
Merged to `master` as 2de64666a07cccf8477e6483de62ae31f463df64

Thanks all for the feedback, the process definitely had a very good impact on my initial submission! ;)

PR #10757, opened 2020-01-04
s_client: Show algorithms & validity period with "Certificate Chain"

MarcT512:
Add certificate validity period (v) and public key & signature algorithms (a) to the s_client "Certificate Chain" output.

Example of connecting to www.google.com:443 showing new (a) & (v) lines:
```
Certificate chain
 0 s:C = US, ST = California, L = Mountain View, O = Google LLC, CN = www.google.com
   i:C = US, O = Google Trust Services, CN = GTS CA 1O1
   a:PKEY: id-ecPublicKey, 256 (bit); sigalg: RSA-SHA256
   v:NotBefore: Dec  3 14:49:26 2019 GMT; NotAfter: Feb 25 14:49:26 2020 GMT
 1 s:C = US, O = Google Trust Services, CN = GTS CA 1O1
   i:OU = GlobalSign Root CA - R2, O = GlobalSign, CN = GlobalSign
   a:PKEY: rsaEncryption, 2048 (bit); sigalg: RSA-SHA256
   v:NotBefore: Jun 15 00:00:42 2017 GMT; NotAfter: Dec 15 00:00:42 2021 GMT
```

PR #7975, opened 2019-01-03
Proxyauth

MarcT512:
RFC: A first attempt at adding basic (base64) proxy authentication to s_client.
CLA: Was initially trivial - now submitted.

##### Checklist

- [x] documentation is added or updated
- [ ] tests are added or updated

PR #11880, opened 2020-05-20
s_client: Fix -proxy flag regression

### LINKED ISSUE #11879: s_client: Connection via HTTP proxy is broken in openssl-3.0.0-alpha2
s_client: connection via an HTTP proxy broke somewhere prior to openssl-3.0.0-alpha2.

`openssl s_client -connect <target> -proxy <proxy_host:proxy_port>`

Results in s_client making a TCP connection to proxy_host:proxy_port and then issuing an HTTP CONNECT to the proxy, instead of the target.

I have a 'works-for-me' fix, which I'll try to link below.

MarcT512:
s_client: connection via an HTTP proxy broke somewhere prior to openssl-3.0.0-alpha2.

`openssl s_client -connect <target> -proxy <proxy_host:proxy_port>`

Results in s_client making a TCP connection to proxy_host:proxy_port and then issuing an HTTP CONNECT to the proxy, instead of the target.

Fixes #11879 

##### Checklist

### DIFF
--- a/apps/s_client.c
+++ b/apps/s_client.c
@@ -1602,7 +1602,7 @@ int s_client_main(int argc, char **argv)
 #endif
 
     if (connectstr != NULL) {
-        int res = 0;
+        int res;
         char *tmp_host = host, *tmp_port = port;
 
         res = BIO_parse_hostserv(connectstr, &host, &port, BIO_PARSE_PRIO_HOST);
@@ -1619,17 +1619,17 @@ int s_client_main(int argc, char **argv)
     }
 
     if (proxystr != NULL) {
-        int res = 0;
+        int res;
         char *tmp_host = host, *tmp_port = port;
 
-        if (connectstr == NULL) {
+        if (host == NULL || port == NULL) {
             BIO_printf(bio_err, "%s: -proxy requires use of -connect or target parameter\n", prog);
             goto opthelp;
         }
 
         /* Retain the original target host:port for use in the HTTP proxy connect string */
-        thost=OPENSSL_strdup(host);
-        tport=OPENSSL_strdup(port);
+        thost = OPENSSL_strdup(host);
+        tport = OPENSSL_strdup(port);
         if (thost == NULL || tport == NULL) {
             BIO_printf(bio_err, "%s: out of memory\n", prog);
             goto end;
--- a/apps/s_client.c
+++ b/apps/s_client.c
@@ -1599,8 +1599,26 @@ int s_client_main(int argc, char **argv)
         goto opthelp;
     }
 #endif
+
+    if (connectstr != NULL) {
+        int res = 0;
+        char *tmp_host = host, *tmp_port = port;
+            res = BIO_parse_hostserv(connectstr, &host, &port,
+                                     BIO_PARSE_PRIO_HOST);
+        if (tmp_host != host)
+            OPENSSL_free(tmp_host);
+        if (tmp_port != port)
+            OPENSSL_free(tmp_port);
+        if (!res) {
+            BIO_printf(bio_err,
+                       "%s: -connect argument or target parameter malformed or ambiguous\n",
+                       prog);
+            goto end;
+        }
+    }
+
     if (proxystr != NULL) {
-        int res;
+        int res = 0;
         char *tmp_host = host, *tmp_port = port;
         if (connectstr == NULL) {
             BIO_printf(bio_err, "%s: -proxy requires use of -connect or target parameter\n", prog);
@@ -1614,22 +1632,6 @@ int s_client_main(int argc, char **argv)
         if (!res) {
             BIO_printf(bio_err,
                        "%s: -proxy argument malformed or ambiguous\n", prog);
-            goto end;
-        }
-    } else {
-        int res = 1;
-        char *tmp_host = host, *tmp_port = port;
-        if (connectstr != NULL)
-            res = BIO_parse_hostserv(connectstr, &host, &port,
-                                     BIO_PARSE_PRIO_HOST);
-        if (tmp_host != host)
-            OPENSSL_free(tmp_host);
-        if (tmp_port != port)
-            OPENSSL_free(tmp_port);
-        if (!res) {
-            BIO_printf(bio_err,
-                       "%s: -connect argument or target parameter malformed or ambiguous\n",
-                       prog);
             goto end;
         }
     }
@@ -2389,9 +2391,19 @@ int s_client_main(int argc, char **argv)
         }
         break;
     case PROTO_CONNECT:
-        if (!OSSL_HTTP_proxy_connect(sbio, host, port, proxyuser, proxypass,
-                                     0 /* no timeout */, bio_err, prog))
-            goto shut;
+	{
+	    /* Need to parse the actual connect target now */
+            char *thost, *tport;
+            /* connectstr was already validated, so we can assume success below */
+            BIO_parse_hostserv(connectstr, &thost, &tport,
+                                         BIO_PARSE_PRIO_HOST);
+
+            if (!OSSL_HTTP_proxy_connect(sbio, thost, tport, proxyuser, proxypass,
+                                         0 /* no timeout */, bio_err, prog))
+                goto shut;
+            OPENSSL_free(thost);
+            OPENSSL_free(tport);
+        }
         break;
     case PROTO_IRC:
         {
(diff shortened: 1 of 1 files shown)


MarcT512:
NB: I do have a signed CLA, but the cla-check doesn't seem to like the github.com email address.

See also https://github.com/openssl/openssl/pull/10757 & https://github.com/openssl/openssl/pull/7975.

mattcaswell:
@DDvO - can you take a look?

apps/s_client.c
```diff
@@ -2389,9 +2391,19 @@ int s_client_main(int argc, char **argv)
...
         break;
     case PROTO_CONNECT:
-        if (!OSSL_HTTP_proxy_connect(sbio, host, port, proxyuser, proxypass,
-                                     0 /* no timeout */, bio_err, prog))
-            goto shut;
+	{
+	    /* Need to parse the actual connect target now */
+            char *thost, *tport;
+            /* connectstr was already validated, so we can assume success below */
+            BIO_parse_hostserv(connectstr, &thost, &tport,
```
t8m, line 2398: To avoid re-parsing why not just save the host and port that you already have above when parsing the proxystr?

MarcT512: Fair enough - I was trying to make minimal changes to prove/fix the issue.
In the version below I strdup() the original values of port & host from connectstr, for use in the HTTP connect string sent to the proxy. They're freed when done.

richsalz: strdup isn't available on all platforms, so you should do ```OPENSSL_strdup``` and ```OPENSSL_free```.

MarcT512: Apologies  - I did use `OPENSSL_strdup` & `OPENSSL_free` in the patch, but didn't write it up correctly above.

richsalz: I apologize for not looking at your code.  Nice work :)

apps/s_client.c
```diff
@@ -1599,37 +1600,44 @@ int s_client_main(int argc, char **argv)
...
+
+    if (connectstr != NULL) {
+        int res = 0;
         char *tmp_host = host, *tmp_port = port;
-        if (connectstr == NULL) {
-            BIO_printf(bio_err, "%s: -proxy requires use of -connect or target parameter\n", prog);
-            goto opthelp;
-        }
-        res = BIO_parse_hostserv(proxystr, &host, &port, BIO_PARSE_PRIO_HOST);
+            res = BIO_parse_hostserv(connectstr, &host, &port,
```
t8m, line 1607: The indentation seems off here. Also please separate declarations from the rest of the code by an empty line.

apps/s_client.c
```diff
@@ -3138,6 +3149,8 @@ int s_client_main(int argc, char **argv)
     OPENSSL_free(bindstr);
     OPENSSL_free(host);
     OPENSSL_free(port);
+    if (thost != NULL) OPENSSL_free(thost);
```
t8m, line 3152: Please drop the ifs, the free call checks it.

apps/s_client.c
```diff
@@ -2389,9 +2397,12 @@ int s_client_main(int argc, char **argv)
         }
         break;
     case PROTO_CONNECT:
-        if (!OSSL_HTTP_proxy_connect(sbio, host, port, proxyuser, proxypass,
-                                     0 /* no timeout */, bio_err, prog))
-            goto shut;
+	{
```
t8m, line 2400: No need to add a block - there are no declarations here.

apps/s_client.c
```diff
@@ -1599,37 +1600,44 @@ int s_client_main(int argc, char **argv)
...
-        if (connectstr != NULL)
-            res = BIO_parse_hostserv(connectstr, &host, &port,
-                                     BIO_PARSE_PRIO_HOST);
+        if (connectstr == NULL) {
+            BIO_printf(bio_err, "%s: -proxy requires use of -connect or target parameter\n", prog);
+            goto opthelp;
+        }
+
+        /* Retain the original target host:port for use in the HTTP proxy connect string */
+        thost=OPENSSL_strdup(host);
```
t8m, line 1630: Please check for NULL return and error out.

apps/s_client.c
MarcT512: You're right. Fixed.

apps/s_client.c
MarcT512: Done.

apps/s_client.c
MarcT512: I did put in some declarations then removed them, block now removed.

apps/s_client.c
MarcT512: Done.

apps/s_client.c
```diff
@@ -1599,37 +1600,49 @@ int s_client_main(int argc, char **argv)
...
+    }
+
+    if (proxystr != NULL) {
+        int res = 0;
         char *tmp_host = host, *tmp_port = port;
-        if (connectstr != NULL)
-            res = BIO_parse_hostserv(connectstr, &host, &port,
-                                     BIO_PARSE_PRIO_HOST);
+
+        if (connectstr == NULL) {
```
t8m, line 1625: Better to check host and port here because this is what really matters. If you specify connectstr in way that does not include host or port part (i.e. you use `:port` or `host:`) you would get the misleading out of memory error from the below check.

apps/s_client.c
```diff
@@ -1599,37 +1600,49 @@ int s_client_main(int argc, char **argv)
         goto opthelp;
     }
 #endif
-    if (proxystr != NULL) {
-        int res;
+
+    if (connectstr != NULL) {
+        int res = 0;
```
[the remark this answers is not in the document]
MarcT512: Fair enough - done.

apps/s_client.c
```diff
@@ -1599,37 +1600,49 @@ int s_client_main(int argc, char **argv)
...
-                                     BIO_PARSE_PRIO_HOST);
+
+        if (connectstr == NULL) {
+            BIO_printf(bio_err, "%s: -proxy requires use of -connect or target parameter\n", prog);
+            goto opthelp;
+        }
+
+        /* Retain the original target host:port for use in the HTTP proxy connect string */
+        thost=OPENSSL_strdup(host);
+        tport=OPENSSL_strdup(port);
```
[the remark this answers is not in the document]
MarcT512: Done

apps/s_client.c
```diff
@@ -1599,37 +1600,49 @@ int s_client_main(int argc, char **argv)
...
+                       "%s: -connect argument or target parameter malformed or ambiguous\n",
+                       prog);
             goto end;
         }
-    } else {
-        int res = 1;
+    }
+
+    if (proxystr != NULL) {
+        int res = 0;
```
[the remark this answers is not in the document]
MarcT512: Done

apps/s_client.c
MarcT512: That is a very good point. I should have spotted that.
Fixed.

apps/s_client.c
DDvO: Each of the tree such strings (connectstr, proxystr, and bindstr) is parsed only once, so I don't see a re-parsing issue here.

apps/s_client.c
DDvO: There is meanwhile a tool that can be used to check the coding style (formatting etc.) automatically: `util/check-format.pl`.

apps/s_client.c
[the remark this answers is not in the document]
DDvO: Also this would have been caught by `util/check-format.pl`.

DDvO:
> @DDvO - can you take a look?

@mattcaswell, unfortunately I often miss such an addressing in the sheer number of messages automatically sent by GitHub. I need to find a way to flag them prominently, maybe using a Thunderbird message filter rule.
This time Hendrik pointed me to this issue.
Thanks @t8m for stepping in in the meantime.

It turns out that the bug fixed here crept in when I carved out the code of the new `OSSL_HTTP_proxy_connect()` in commit 29f178bddfdbd11218fbcba0b8060297696968e3, so I really was the right person to address here.

apps/s_client.c
DDvO: (I just found that my above comment was outdated.)

t8m:
The cla required label is incorrect. Marc has CLA.

t8m:
Merged to master as f7201301ef . Thank you for the PR!

PR #11884, opened 2020-05-20
Add libctx support to CMS/PKCS7

slontis:
##### Checklist


slontis:
Documentation still needs to be done.

crypto/cms/cms_kari.c
```diff
@@ -413,13 +429,15 @@ int cms_RecipientInfo_kari_init(CMS_RecipientInfo *ri,  X509 *recip, EVP_PKEY *r
...
     EVP_CIPHER_CTX *ctx = kari->ctx;
-    const EVP_CIPHER *kekcipher;
+    EVP_CIPHER *kekcipher;
+    const char *kekcipher_name;
     int keylen;
     int ret;
 
     /* If a suitable wrap algorithm is already set nothing to do */
-    kekcipher = EVP_CIPHER_CTX_cipher(ctx);
+    kekcipher = (EVP_CIPHER *)EVP_CIPHER_CTX_cipher(ctx);
```
bukka, line 440: I'm sure it has got a reason but was just wondering why is it necessary to cast from const...?

crypto/cms/cms_lib.c
```diff
@@ -20,18 +20,65 @@ DEFINE_STACK_OF(CMS_RevocationInfoChoice)
...
+{
+    CMS_ContentInfo *ci;
+
+    ci = (CMS_ContentInfo *)ASN1_item_new(ASN1_ITEM_rptr(CMS_ContentInfo));
+    cms_set0_libctx(ci, NULL, NULL);
+    return ci;
+}
+
+CMS_ContentInfo *CMS_ContentInfo_new_ex(OPENSSL_CTX *libctx,
+                                        const char *propq)
```
bukka, line 46: CS: might not need to be wrapped (looks like 80 on single line)

crypto/cms/cms_lib.c
```diff
@@ -20,18 +20,65 @@ DEFINE_STACK_OF(CMS_RevocationInfoChoice)
...
+{
+    ASN1_item_free((ASN1_VALUE *)cms, ASN1_ITEM_rptr(CMS_ContentInfo));
+}
+
+const CMS_CTX *CMS_get0_ctx(const CMS_ContentInfo *cms)
+{
+    return cms != NULL ? &cms->ctx : NULL;
+}
+
+void CMS_set0_libctx(CMS_ContentInfo *ci, OPENSSL_CTX *libctx, const char *propq)
```
bukka, line 64: CS: this is over 80 :)

doc/man3/CMS_set0_libctx.pod
```diff
@@ -0,0 +1,52 @@
...
+
+ #include <openssl/cms.h>
+
+ void CMS_set0_libctx(CMS_ContentInfo *ci, OPENSSL_CTX *libctx,
+                      const char *propq);
+
+=head1 DESCRIPTION
+
+CMS_set0_libctx() may be called after loading CMS_ContentInfo
+(See CMSMIME_read_CMS(), PEM_read_bio_CMS(), d2i_CMS_bio()) to set the library
```
bukka, line 17: typo: should be `SMIME_read_CMS()`

crypto/cms/cms_kari.c
slontis: I was reusing the variable down further - will add it as a seperate non cost variable instead.

apps/lib/app_provider.c
```diff
@@ -69,6 +89,8 @@ int opt_provider(int opt)
         return opt_provider_load(opt_arg());
     case OPT_PROV_PROVIDER_PATH:
         return opt_provider_path(opt_arg());
+    case OPT_PROV_PROVIDER_CONFIG_FILE:
+        return opt_libctx_config_path(opt_arg());
```
mattcaswell, line 93: I'm not too sure about the semantics of this. Effectively this introduces the side effect of creating a non-default libctx. This is not at all obvious and means that `-provider default -provider_config conf.cnf` means something different to `-provider_config conf.cnf -provider default`. This seems quite non-intuitive. It also means the option will be ignored for any app that we don't update to being "libctx" aware.

A possible alternative is that these options never create a new libctx themselves. They only use whatever the value of `libctx` is. Applications that are libctx aware could call a `setup_libctx` function which creates a non-default value for libctx, and loads the "null" provider into the default ctx.

apps/include/opt.h
```diff
@@ -271,17 +271,20 @@
...
         OPT_PROV_PROVIDER, OPT_PROV_PROVIDER_PATH, \
+        OPT_PROV_PROVIDER_CONFIG_FILE, \
         OPT_PROV__LAST
 
 # define OPT_PROV_OPTIONS \
         OPT_SECTION("Provider"), \
         { "provider_path", OPT_PROV_PROVIDER_PATH, 's', "Provider load path (must be before 'provider' argument if required)" }, \
-        { "provider", OPT_PROV_PROVIDER, 's', "Provider to load (can be specified multiple times)" }
+        { "provider", OPT_PROV_PROVIDER, 's', "Provider to load (can be specified multiple times)" }, \
+        { "provider_config", OPT_PROV_PROVIDER_CONFIG_FILE, 's', "Provider config file to load (can be specified multiple times)." }
```
mattcaswell, line 281: This option needs documentation

crypto/asn1/a_digest.c
```diff
@@ -38,30 +38,40 @@ int ASN1_digest(i2d_of_void *i2d, const EVP_MD *type, char *data,
...
     i = ASN1_item_i2d(asn, &str, it);
-    if (!str)
+    if (str == NULL)
         return 0;
 
-    if (!EVP_Digest(str, i, md, len, type, NULL)) {
-        OPENSSL_free(str);
-        return 0;
-    }
+    if (!EVP_MD_provider(md))
```
mattcaswell, line 60: `EVP_MD_provider == NULL`

crypto/asn1/asn_mime.c
```diff
@@ -231,14 +232,16 @@ static int asn1_write_micalg(BIO *out, STACK_OF(X509_ALGOR) *mdalgs)
 
 /* SMIME sender */
 
-int SMIME_write_ASN1(BIO *bio, ASN1_VALUE *val, BIO *data, int flags,
-                     int ctype_nid, int econt_nid,
-                     STACK_OF(X509_ALGOR) *mdalgs, const ASN1_ITEM *it)
+int SMIME_write_ASN1_ex(BIO *bio, ASN1_VALUE *val, BIO *data, int flags,
+                        int ctype_nid, int econt_nid,
+                        STACK_OF(X509_ALGOR) *mdalgs, const ASN1_ITEM *it,
+                        OPENSSL_CTX *libctx)
```
mattcaswell, line 238: Should we also pass the propq to all these functions? We're quite inconsistent, but I tend to think we should always pass both.

crypto/cms/cms_dd.c
```diff
@@ -17,11 +17,13 @@
...
 
-CMS_ContentInfo *cms_DigestedData_create(const EVP_MD *md)
+CMS_ContentInfo *cms_DigestedData_create(const EVP_MD *md,
+                                         OPENSSL_CTX *libctx, const char *propq)
 {
     CMS_ContentInfo *cms;
     CMS_DigestedData *dd;
-    cms = CMS_ContentInfo_new();
+
+    cms = CMS_ContentInfo_new_ex(libctx, propq);
```
mattcaswell, line 26: `CMS_ContentInfo_new_with_libctx`?

crypto/cms/cms_env.c
```diff
@@ -756,10 +812,12 @@ static int cms_RecipientInfo_kekri_encrypt(const CMS_ContentInfo *cms,
     r = 1;
 
  err:
+    EVP_CIPHER_free(cipher);
     if (!r)
         OPENSSL_free(wkey);
     EVP_CIPHER_CTX_free(ctx);
 
+
```
mattcaswell, line 820: Too many blank lines?

crypto/cms/cms_ess.c
```diff
@@ -30,30 +30,30 @@ int CMS_get1_ReceiptRequest(CMS_SignerInfo *si, CMS_ReceiptRequest **prr)
...
 
-CMS_ReceiptRequest *CMS_ReceiptRequest_create0(unsigned char *id, int idlen,
-                                               int allorfirst,
-                                               STACK_OF(GENERAL_NAMES)
-                                               *receiptList, STACK_OF(GENERAL_NAMES)
-                                               *receiptsTo)
+CMS_ReceiptRequest *CMS_ReceiptRequest_create0_ex(
+    unsigned char *id, int idlen, int allorfirst,
+    STACK_OF(GENERAL_NAMES) *receiptList, STACK_OF(GENERAL_NAMES) *receiptsTo,
+    OPENSSL_CTX *libctx)
```
mattcaswell, line 56: propq?

crypto/cms/cms_lib.c
```diff
@@ -20,18 +20,65 @@ DEFINE_STACK_OF(CMS_RevocationInfoChoice)
...
+{
+    ASN1_item_free((ASN1_VALUE *)cms, ASN1_ITEM_rptr(CMS_ContentInfo));
+}
+
+const CMS_CTX *CMS_get0_ctx(const CMS_ContentInfo *cms)
+{
+    return cms != NULL ? &cms->ctx : NULL;
+}
+
+void CMS_set0_libctx(CMS_ContentInfo *ci, OPENSSL_CTX *libctx,
```
mattcaswell, line 63: Should this be `CMS_ContentInfo_set0_libctx`? Same for the get0 above

crypto/x509/x509_cmp.c
```diff
@@ -163,6 +163,14 @@ int X509_cmp(const X509 *a, const X509 *b)
     return rv;
 }
 
+int X509_cmp_ex(X509 *a, X509 *b,
+                OPENSSL_CTX *libctx, const char *propq)
+{
+    X509_set0_libctx(a, libctx, propq);
+    X509_set0_libctx(b, libctx, propq);
```
mattcaswell, line 170: Surely the whole point of adding the libctx to the x509 structure is that we don't have to change all the `X509_*` calls to additionally take a libctx. So, I'm not sure why this one gets singled out for special treatment.

crypto/x509/x_x509.c
```diff
@@ -116,6 +116,14 @@ ASN1_SEQUENCE_ref(X509, x509_cb) = {
 IMPLEMENT_ASN1_FUNCTIONS(X509)
 IMPLEMENT_ASN1_DUP_FUNCTION(X509)
 
+void X509_set0_libctx(X509 *x509, OPENSSL_CTX *libctx, const char *propq)
```
mattcaswell, line 119: If we're going to take this approach, I wonder if we need the newly added `X509_verify_ex()` function. But so far I'm wondering why we need it at all...

crypto/cms/cms_kari.c
```diff
@@ -342,18 +351,25 @@ static int cms_kari_set_originator_private_key(CMS_KeyAgreeRecipientInfo *kari,
...
+    if (ri->d.kari == NULL)
         return 0;
     ri->type = CMS_RECIPINFO_AGREE;
 
     kari = ri->d.kari;
     kari->version = 3;
+    kari->cms_ctx = ctx;
+
+    X509_set0_libctx(recip, ctx->libctx, ctx->propq);
+    X509_set0_libctx(originator, ctx->libctx, ctx->propq);
```
mattcaswell, line 372: We seem to do this a lot and I'm not convinced this is correct. Can't an X509 object have a lifetime beyond a single CMS message, and may be used in multiple CMS messages? So tying to set the libctx to the one used by the CMS_CTX somehow doesn't seem right. I would expect the libctx associated with the X509 to be set shortly after it is created, and then never modified again. If the X509 object is created as part of the CMS operation, then it seems reasonable to set its libctx (and do so immediately after its creation) - but if it is passed in through an external API then this makes less sense.

crypto/cms/cms_sd.c
```diff
@@ -267,11 +276,13 @@ CMS_SignerInfo *CMS_add1_signer(CMS_ContentInfo *cms,
     if (!si)
         goto merr;
     /* Call for side-effect of computing hash and caching extensions */
+    X509_set0_libctx(signer, ctx->libctx, ctx->propq);
```
mattcaswell, line 279: Should the caller be responsible for doing this here?

crypto/x509/v3_purp.c
```diff
@@ -83,7 +83,7 @@ int X509_check_purpose(X509 *x, int id, int ca)
     int idx;
     const X509_PURPOSE *pt;
 
-    if (!X509v3_cache_extensions(x, NULL, NULL))
+    if (!X509v3_cache_extensions(x, x->libctx, x->propq))
```
mattcaswell, line 86: I wonder whether we should now remove the `X509v3_cache_extensions` function from the public API. It was recently added. Previously it was an internal function without the libctx/propq parameters...but now we have the ability to store the libctx in the X509 itself, so it seems pointless to pass through a different one.

doc/man3/X509_set0_libctx.pod
```diff
@@ -0,0 +1,56 @@
...
+
+X509_set0_libctx() may be called before calling other X509 functions that need 
+a library context I<libctx> and property query I<propq> to fetch algorithms
+from providers. These values are both NULL by default which results in the
+default (NULL) library context being used.
+
+=head1 NOTES
+
+The X509_set0_libctx() should be used by certificate objects if the library
+context is required. The following functions use the values set by
```
mattcaswell, line 23: Compare with X509v3_cache_extensions. I don't think we need both of those functions.

crypto/cms/cms_dd.c
slontis: Have we actually decided that that is the name we want to propagate?

crypto/cms/cms_sd.c
slontis: The caller is the app. Are you sure you want every app to have to set this themselves? More than likely it will fall over in a fecth and they wont have any idea why it happened.

crypto/cms/cms_dd.c
mattcaswell: I thought we had

slontis: Ok no problem I thought we still had a few new_ex about the place..

apps/lib/app_provider.c
slontis: Hmm I just tried doing this in #11948..
So having the app always have the libctx is also not good (at least initially as it propogates to every other instance of the app being called). if we have just one case where the libctx is not valid (e.g: SM2) then another test will break.. Maybe a -use_libctx option is better?

crypto/cms/cms_dd.c
mattcaswell: > Ok no problem I thought we still had a few new_ex about the place..

Having made the decision and still having some `new_ex` about the place are two different things! I thought the general consensus was that this was the preferred name form. That doesn't mean that all existing name forms have been changed.

apps/lib/app_provider.c
mattcaswell: Yes, that could work.

apps/include/opt.h
levitte: Like I said in #11948, I don't see why this option is necessary.  We have `-config` already.

crypto/asn1/asn_mime.c
slontis: ok it only uses it for rand - will change

crypto/cms/cms_lib.c
slontis: No - the CMS_ is used all through tthe APIS's to mean contentinfo..

apps/lib/app_provider.c
slontis: ok done.

crypto/x509/x509_cmp.c
slontis: Fixed

crypto/x509/x_x509.c
slontis: Removed the function.

crypto/cms/cms_kari.c
slontis: fixed..

crypto/cms/cms_sd.c
slontis: Removed

crypto/x509/v3_purp.c
slontis: Removed

doc/man3/X509_set0_libctx.pod
slontis: Removed X509v3_cache_extensions()

apps/include/opt.h
slontis: changed to use -config

slontis:
This should be ready for review again.

slontis:
ping

slontis:
Changed to WIP: as it is now dependant on bits from #11948 and #12153 are merged..
It will be rebased once those dependant PR's are merged, This should make it slightly easier to review.

slontis:
@mattcaswell 
Taken out of work in progress now that the dependent PR's have been merged.
It required quite a bit of rework since the base PR for X509 changed quite a bit,
so I have also broken it up into separate logical commits.

crypto/cms/cms_io.c
```diff
@@ -71,19 +76,34 @@ int SMIME_write_CMS(BIO *bio, CMS_ContentInfo *cms, BIO *data, int flags)
...
         mdalgs = cms->d.signedData->digestAlgorithms;
     else
         mdalgs = NULL;
 
-    return SMIME_write_ASN1(bio, (ASN1_VALUE *)cms, data, flags,
-                            ctype_nid, econt_nid, mdalgs,
-                            ASN1_ITEM_rptr(CMS_ContentInfo));
+    return SMIME_write_ASN1_with_libctx(bio, (ASN1_VALUE *)cms, data, flags,
+                               ctype_nid, econt_nid, mdalgs,
+                               ASN1_ITEM_rptr(CMS_ContentInfo),
```
mattcaswell, line 88: Indentation is now off

crypto/cms/cms_lib.c
```diff
@@ -14,24 +14,107 @@
...
+
+CMS_ContentInfo *CMS_ContentInfo_new_with_libctx(OPENSSL_CTX *libctx,
+                                                 const char *propq)
+{
+    CMS_ContentInfo *ci;
+
+    ci = (CMS_ContentInfo *)ASN1_item_new(ASN1_ITEM_rptr(CMS_ContentInfo));
+    if (ci != NULL) {
+        ci->ctx.libctx = libctx;
+        ci->ctx.propq = propq;
```
mattcaswell, line 55: Shouldn't we dup the propq here?

crypto/cms/cms_sd.c
```diff
@@ -395,14 +408,18 @@ CMS_SignerInfo *CMS_add1_signer(CMS_ContentInfo *cms,
...
             if (EVP_PKEY_sign_init(si->pctx) <= 0)
                 goto err;
             if (EVP_PKEY_CTX_set_signature_md(si->pctx, md) <= 0)
                 goto err;
-        } else if (EVP_DigestSignInit(si->mctx, &si->pctx, md, NULL, pk) <= 0)
+        } else if (EVP_DigestSignInit_with_libctx(si->mctx, &si->pctx,
+                                                  EVP_MD_name(md),
+                                                  ctx->libctx, ctx->propq,
+                                                  pk) <= 0)
             goto err;
```
mattcaswell, line 423: {} needed here

crypto/cms/cms_smime.c
```diff
@@ -460,35 +496,38 @@ int CMS_verify_receipt(CMS_ContentInfo *rcms, CMS_ContentInfo *ocms,
...
     return cms_Receipt_verify(rcms, ocms);
 }
 
-CMS_ContentInfo *CMS_sign(X509 *signcert, EVP_PKEY *pkey,
-                          STACK_OF(X509) *certs, BIO *data,
-                          unsigned int flags)
+CMS_ContentInfo *CMS_sign_with_libctx(X509 *signcert, EVP_PKEY *pkey,
+                             STACK_OF(X509) *certs, BIO *data,
+                             unsigned int flags,
+                             OPENSSL_CTX *libctx, const char *propq)
```
mattcaswell, line 510: Indentation is off

doc/man3/CMS_EnvelopedData_create.pod
```diff
@@ -24,6 +27,11 @@ L<CMS_add0_recipient_key(3)>.
 The B<CMS_ContentInfo> structure needs to be finalized using L<CMS_final(3)>
 and then freed using L<CMS_ContentInfo_free(3)>.
 
+
+CMS_EnvelopedData_create_with_libctx() is similar to CMS_EnvelopedData_create()
```
mattcaswell, line 31: I've tended to put the `with_libctx` version first in the docs and then describe how the non-libctx version is different.

crypto/pkcs7/pk7_asn1.c
```diff
@@ -62,7 +63,33 @@ ASN1_NDEF_SEQUENCE_cb(PKCS7, pk7_cb) = {
...
+    return ASN1_item_i2d((const ASN1_VALUE *)a, out, (PKCS7_it()));\
+}
+
+PKCS7 *PKCS7_new_with_libctx(OPENSSL_CTX *libctx, const char *propq)
+{
+    PKCS7 *pkcs7 = PKCS7_new();
+
+    if (pkcs7 != NULL) {
+        pkcs7->ctx.libctx = libctx;
+        pkcs7->ctx.propq = propq;
```
mattcaswell, line 89: dup the propq?

doc/man3/X509_dup.pod
```diff
@@ -334,6 +336,10 @@ to generate the function bodies.
 B<I<TYPE>_new>() allocates an empty object of the indicated type.
 The object returned must be released by calling B<I<TYPE>_free>().
 
+B<I<TYPE>_new_ex>() is similiar to B<I<TYPE>_new>() but also passes the library
```
mattcaswell, line 339: Don't you mean new_with_libctx here?

util/missingcrypto.txt
```diff
@@ -1120,8 +1123,10 @@ SHA256_Transform(3)
 SHA512_Transform(3)
 SMIME_crlf_copy(3)
 SMIME_read_ASN1(3)
+SMIME_read_ASN1_ex(3)
 SMIME_text(3)
 SMIME_write_ASN1(3)
+SMIME_write_ASN1_with_libctx(3)
```
mattcaswell, line 1129: Why are we adding new undocumented functions?

crypto/cms/cms_env.c
```diff
@@ -430,19 +480,21 @@ static int cms_RecipientInfo_ktri_decrypt(CMS_ContentInfo *cms,
     if (cms->d.envelopedData->encryptedContentInfo->havenocert
             && !cms->d.envelopedData->encryptedContentInfo->debug) {
         X509_ALGOR *calg = ec->contentEncryptionAlgorithm;
-        const EVP_CIPHER *ciph = EVP_get_cipherbyobj(calg->algorithm);
+        const char *name = OBJ_nid2sn(OBJ_obj2nid(calg->algorithm));
 
+        ciph = EVP_CIPHER_fetch(ctx->libctx, name, ctx->propq);
```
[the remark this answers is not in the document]
slontis: We need to either call the fetch OR evp_get_cipherbyname_ex() here so that it handles the libctx.

crypto/cms/cms_lib.c
slontis: Done

doc/man3/CMS_EnvelopedData_create.pod
slontis: ok

doc/man3/X509_dup.pod
slontis: yep..

util/missingcrypto.txt
slontis: Because it is derived from an existing undocumented function.

slontis:
I had to search for with_libctx to fix up all the indentation and documentation. Hopefully I got them all now.

slontis:
The last 5 commits are the fix ups after a rebase.

util/missingcrypto.txt
mattcaswell: So lets document it!

slontis: I was afraid you might say that - now I have to figure out what they really do..

slontis: Ok added docs now.

crypto/cms/cms_env.c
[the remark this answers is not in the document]
mattcaswell: Ok

doc/man3/CMS_EncryptedData_decrypt.pod
```diff
@@ -0,0 +1,49 @@
...
+
+ #include <openssl/cms.h>
+
+ int CMS_EncryptedData_decrypt(CMS_ContentInfo *cms,
+                               const unsigned char *key, size_t keylen,
+                               BIO *dcont, BIO *out, unsigned int flags);
+
+=head1 DESCRIPTION
+
+CMS_EncryptedData_decrypt() decrypts an I<cms> EncryptedData object using the
```
mattcaswell, line 18: *a* cms?

doc/man3/EVP_SignInit.pod
```diff
@@ -28,20 +33,30 @@ EVP_SignUpdate() hashes I<cnt> bytes of data at I<d> into the
...
 will be written to the integer at I<s>, at most C<EVP_PKEY_size(pkey)> bytes
 will be written.
 
+EVP_SignFinal() is similar to EVP_SignFinal_with_libctx() but uses default
+values of NULL for the library context I<libctx> and the property query I<propq>.
+
+ but also passes the
+library context I<libctx> and the property query I<propq> to use when creating
+a context to use with the key I<pkey>. EVP_SignFinal() uses NULL values for
+these fields.
```
mattcaswell, line 51: This text seems misplaced

slontis:
Looks like fuzzing test is generally broken at the moment.

apps/cms.c
```diff
@@ -978,8 +999,9 @@ int cms_main(int argc, char **argv)
                 goto end;
         }
     } else if (operation == SMIME_ENCRYPTED_ENCRYPT) {
-        cms = CMS_EncryptedData_encrypt(in, cipher,
-                                        secret_key, secret_keylen, flags);
+        cms = CMS_EncryptedData_encrypt_with_libctx(in, cipher,
+                                           secret_key, secret_keylen, flags,
+                                           libctx, propq);
```
mattcaswell, line 1004: Indentation is off

include/openssl/asn1.h
```diff
@@ -845,7 +845,13 @@ int PEM_write_bio_ASN1_stream(BIO *out, ASN1_VALUE *val, BIO *in, int flags,
 int SMIME_write_ASN1(BIO *bio, ASN1_VALUE *val, BIO *data, int flags,
                      int ctype_nid, int econt_nid,
                      STACK_OF(X509_ALGOR) *mdalgs, const ASN1_ITEM *it);
+int SMIME_write_ASN1_with_libctx(BIO *bio, ASN1_VALUE *val, BIO *data, int flags,
+                        int ctype_nid, int econt_nid,
+                        STACK_OF(X509_ALGOR) *mdalgs, const ASN1_ITEM *it,
+                        OPENSSL_CTX *libctx, const char *propq);
```
mattcaswell, line 851: Indentation is off

slontis:
rebased again : only change was a rename of an API from "_ex" to "_with_libctx".

slontis:
Thanks for reviewing. Finally merged to master.

PR #11888, opened 2020-05-20
Fix d2i_PrivateKey() to work as documented [1.1.1]

levitte:
d2i_PrivateKey() is documented to return keys of the type given as
first argument |type|, unconditionally.  Most specifically, the manual
says this:

> An error occurs if the decoded key does not match type.

However, when faced of a PKCS#8 wrapped key, |type| was ignored, which
may lead to unexpected results.

(cherry picked from commit b2952366dd0248bf35c83e1736cd203033a22378)


levitte:
Merged

176eb406691f14d560cf7619365830a4d033ee28 Fix d2i_PrivateKey() to work as documented

PR #11892, opened 2020-05-21
Coverity fixes

paulidale:
Fixes for recent Coverity issues:

- 1463571: Null pointer dereferences (FORWARD_NULL)
- 1463574: Null pointer dereferences (REVERSE_INULL)
- 1463576: Error handling issues (CHECKED_RETURN)
- 1463258: Incorrect expression (EVALUATION_ORDER)

### DIFF
--- a/crypto/x509/v3_ncons.c
+++ b/crypto/x509/v3_ncons.c
@@ -197,7 +197,7 @@ static int print_nc_ipadd(BIO *bp, ASN1_OCTET_STRING *ip)
     int len2 = ip->length - len1;
     char *ip1 = ipaddr_to_asc(ip->data, len1);
     char *ip2 = ipaddr_to_asc(ip->data + len1, len2);
-    int ret = ret = ip1 != NULL && ip2 != NULL
+    int ret = ip1 != NULL && ip2 != NULL
         && BIO_printf(bp, "IP:%s/%s", ip1, ip2) > 0;
 
     OPENSSL_free(ip1);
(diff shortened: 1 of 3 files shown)


providers/implementations/signature/rsa.c
```diff
@@ -337,9 +337,9 @@ static int rsa_signature_init(void *vprsactx, void *vrsa, int operation)
                     return 0;
                 }
 
-                strncpy(prsactx->mdname, mdname, sizeof(prsactx->mdname));
+                strncpy(prsactx->mdname, mdname, sizeof(prsactx->mdname) - 1);
                 strncpy(prsactx->mgf1_mdname, mgf1mdname,
-                        sizeof(prsactx->mgf1_mdname));
+                        sizeof(prsactx->mgf1_mdname) - 1);
```
kroeckx, line 342: This is being fixed in #11878 already.

kroeckx: I suggest you read the discussion there, because this doesn't actually fix anything.

paulidale: Thanks, I'll drop this part.

paulidale: And gone...

kroeckx:
Should that be REVERSE_NULL, without the I?

paulidale:
It was a copy paste from the Coverity issue, so I guess so.

t8m:
It could mean 'is NULL'.

paulidale:
Merged to master, thanks for the review.

PR #11893, opened 2020-05-21
Fix missing dependencies

zorrorffm:
Error will be reported when compiling code using "make -jxx" command.
There are some missing dependencies. Fix it by making dependencies between
*.c files and corresponding *.h files in providers/common/der directory.


mattcaswell:
Seems reasonable....@levitte?

slontis:
travis error is a timeout

levitte:
I'm a bit worried about repeated file names...  and didn't see this before now.  So [ahem], there's another variant in #11906...

Issue #11870, opened 2020-05-19
X509_PUBKEY_cmp() seems incorrect


vt-alt:
There is `static` function `X509_PUBKEY_cmp` used by `OSSL_CRMF_MSGS_verify_popo` in `crypto/crmf/crmf_lib.c"`. I wanted to copy-paste it for some gost-engine tests (since it's not exported), but found it's being strange:
```
/* returns 0 for equal, -1 for a < b or error on a, 1 for a > b or error on b */
static int X509_PUBKEY_cmp(X509_PUBKEY *a, X509_PUBKEY *b)
{
    X509_ALGOR *algA = NULL, *algB = NULL;
    int res = 0;

    if (a == b)
        return 0;
    if (a == NULL || !X509_PUBKEY_get0_param(NULL, NULL, NULL, &algA, a)
            || algA == NULL)
        return -1;
    if (b == NULL || !X509_PUBKEY_get0_param(NULL, NULL, NULL, &algB, b)
            || algB == NULL)
        return 1;
    if ((res = X509_ALGOR_cmp(algA, algB)) != 0)
        return res;
    return EVP_PKEY_cmp(X509_PUBKEY_get0(a), X509_PUBKEY_get0(b));
}
```
It should return 0 if keys match, **but** `EVP_PKEY_cmp` will return 1 if keys match.

Man: `The function EVP_PKEY_cmp_parameters() and EVP_PKEY_cmp() return 1 if the keys match, 0 if they don't match, -1 if the key types are different and -2 if the operation is not supported.`.

bernd-edlinger:
Yeah, that makes no sense at all.
Fortunatetly that X509_PUBKEY_cmp is not in the LTS branch.

@DDvO I apologize for not finding this while I was reviewing this
commit, I should not have done that review so quickly  :smile:

bernd-edlinger:
Okay, I looked at the original review of this function, and found one
unresolved review comment:
https://github.com/openssl/openssl/pull/7646/files#r234418540
I said "X509_PUBKEY_get0 can return NULL, and EVP_PKEY_cmp does not check for NULL.
That means you have to."
@DDvO said "Done" and I beleived him, and therefore approved.
@mattcaswell said on 11th dec 2018:
 https://github.com/openssl/openssl/pull/7646/files#r240675299
"Done?? I don't see it here?"
And @DDvO answerded:
"

This was done in

commit 413ed461c4edbd0b0ee5ac338361b962776ec8c4
[...]

Issue #9296, opened 2019-07-02
EVP_PKEY_meth_set_copy API break on master


mattcaswell:
Commit 9fdcc21f made lots of constification changes. One of those was to make the parameter of `EVP_PKEY_CTX_dup()` a const. That in itself is fine and is API compatible, but that function calls the "copy" function of the underlying EVP_PKEY_METHOD. Therefore that commit also changed the function pointer parameter to `EVP_PKEY_meth_set_copy`:

````
 void EVP_PKEY_meth_set_copy(EVP_PKEY_METHOD *pmeth,
                             int (*copy) (EVP_PKEY_CTX *dst,
-                                         EVP_PKEY_CTX *src));
+                                         const EVP_PKEY_CTX *src));
````
*This* change is *not* API compatible and so represents an API break with 1.1.1 (something we are explicitly trying to avoid as much as possible).

Should we revert this aspect of the change (not sure what the consequences of doing that might be since this change was made for a reason)?

**Comment by mattcaswell** [comment]
@DDvO

**Comment by DDvO** [comment]
@mattcaswell, just very recently I came across that issue where you mentioned me.
Apparently it is very similar to https://github.com/openssl/openssl/pull/11894#discussion_r430288241.

I fear that if we revert the type of the `copy` parameter of  `EVP_PKEY_meth_set_copy()`, strictly speaking the constification of most (if not all) of the `*_dup()` functions done in #8029 and #9347 would have to be reverted, too.

And it looks like since last July no user complained about this change?

**Comment by DDvO** [comment]
As already parly mentioned above, there is a similar API incompatibility on
```
void EVP_PKEY_asn1_set_public(EVP_PKEY_ASN1_METHOD *ameth,
                              int (*pub_decode) (EVP_PKEY *pk,
                                                 const X509_PUBKEY *pub),
                              int (*pub_encode) (X509_PUBKEY *pub,
                                                 const EVP_PKEY *pk),
                              int (*pub_cmp) (const EVP_PKEY *a,
                                              const EVP_PKEY *b),
                              int (*pub_print) (BIO *out,
                                                const EVP_PKEY *pkey,
                                                int indent, ASN1_PCTX *pctx),
                              int (*pkey_size) (const EVP_PKEY *pk),
                              int (*pkey_bits) (const EVP_PKEY *pk));
```
because the `pub` argument of the `pub_decode` function pointer argument now is `const`.
[...]

PR #11894, opened 2020-05-21
Fix X509_PUBKEY_cmp()

DDvO:
As discussed in #11870:
Fix `X509_PUBKEY_cmp()`, move to crypto/x509/x_pubkey.c, export, and document it-
Solves #11870.

### DIFF
--- a/crypto/asn1/ameth_lib.c
+++ b/crypto/asn1/ameth_lib.c
@@ -277,7 +277,7 @@ void EVP_PKEY_asn1_free(EVP_PKEY_ASN1_METHOD *ameth)
 
 void EVP_PKEY_asn1_set_public(EVP_PKEY_ASN1_METHOD *ameth,
                               int (*pub_decode) (EVP_PKEY *pk,
-                                                 X509_PUBKEY *pub),
+                                                 const X509_PUBKEY *pub),
                               int (*pub_encode) (X509_PUBKEY *pub,
                                                  const EVP_PKEY *pk),
                               int (*pub_cmp) (const EVP_PKEY *a,
--- a/crypto/x509/x_pubkey.c
+++ b/crypto/x509/x_pubkey.c
@@ -472,3 +472,24 @@ ASN1_BIT_STRING *X509_get0_pubkey_bitstr(const X509 *x)
         return NULL;
     return x->cert_info.key->public_key;
 }
+
+/* Returns 1 for equal, 0, for non-equal, < 0 on error */
+int X509_PUBKEY_eq(const X509_PUBKEY *a, const X509_PUBKEY *b)
+{
+    X509_ALGOR *algA, *algB;
+    EVP_PKEY *pA, *pB;
+
+    if (a == b)
+        return 1;
+    if (a == NULL || b == NULL)
+        return 0;
+    if (!X509_PUBKEY_get0_param(NULL, NULL, NULL, &algA, a) || algA == NULL
+        || !X509_PUBKEY_get0_param(NULL, NULL, NULL, &algB, b) || algB == NULL)
+        return -2;
+    if (X509_ALGOR_cmp(algA, algB) != 0)
+        return 0;
+    if ((pA = X509_PUBKEY_get0(a)) == NULL
+        || (pB = X509_PUBKEY_get0(b)) == NULL)
+        return -2;
+    return EVP_PKEY_cmp(pA, pB);
+}
--- a/crypto/x509/x_pubkey.c
+++ b/crypto/x509/x_pubkey.c
@@ -30,7 +30,7 @@ struct X509_pubkey_st {
     EVP_PKEY *pkey;
 };
 
-static int x509_pubkey_decode(EVP_PKEY **pk, X509_PUBKEY *key);
+static int x509_pubkey_decode(EVP_PKEY **pk, const X509_PUBKEY *key);
 
 /* Minor tweak to operation: free up EVP_PKEY */
 static int pubkey_cb(int operation, ASN1_VALUE **pval, const ASN1_ITEM *it,
@@ -151,7 +151,7 @@ int X509_PUBKEY_set(X509_PUBKEY **x, EVP_PKEY *pkey)
  */
 
 
-static int x509_pubkey_decode(EVP_PKEY **ppkey, X509_PUBKEY *key)
+static int x509_pubkey_decode(EVP_PKEY **ppkey, const X509_PUBKEY *key)
 {
     EVP_PKEY *pkey = EVP_PKEY_new();
 
@@ -188,7 +188,7 @@ static int x509_pubkey_decode(EVP_PKEY **ppkey, X509_PUBKEY *key)
     return 0;
 }
 
-EVP_PKEY *X509_PUBKEY_get0(X509_PUBKEY *key)
+EVP_PKEY *X509_PUBKEY_get0(const X509_PUBKEY *key)
 {
     EVP_PKEY *ret = NULL;
 
@@ -216,7 +216,7 @@ EVP_PKEY *X509_PUBKEY_get0(X509_PUBKEY *key)
     return NULL;
 }
 
-EVP_PKEY *X509_PUBKEY_get(X509_PUBKEY *key)
+EVP_PKEY *X509_PUBKEY_get(const X509_PUBKEY *key)
 {
     EVP_PKEY *ret = X509_PUBKEY_get0(key);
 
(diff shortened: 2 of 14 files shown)


mspncp:
Nit: please add `Solves #11870` to the body of the commit message.

t8m:
Hmm... this number of different return values in different cases makes the semantics seriously convoluted. Especially for a function that is now being public. Is this function used for sorting? Does actually make sorting by pubkey any sense? If not, would it make sense to make the different return values just internal thing or collapse them (at least partially). Otherwise we might not be able to fulfil the API contract in future if we make some changes.

crypto/x509/x_pubkey.c
```diff
@@ -472,3 +472,32 @@ ASN1_BIT_STRING *X509_get0_pubkey_bitstr(const X509 *x)
...
+            || algB == NULL)
+        return 1;
+    if ((res = X509_ALGOR_cmp(algA, algB)) != 0)
+        return res;
+    if ((pA = X509_PUBKEY_get0(a)) == NULL)
+        return -1;
+    if ((pB = X509_PUBKEY_get0(b)) == NULL)
+        return 1;
+    res = EVP_PKEY_cmp(pA, pB);
+    return res == 1 ? 0 : res - 2;
```
bernd-edlinger, line 502: No this can't be right, the return values of EVP_PKEY_cmp
are :
1  = equel
0  = non-equal
-1 = key type mismatch
-2 = error

so no way how to return 1 for a > b here.
Anyway this returning +/- 1 based on numeric difference of private key data
looks more or less like a oracle.

vt-alt:
@t8m JFYI: it's partly discussed since there: https://github.com/openssl/openssl/issues/11870#issuecomment-631504517

crypto/x509/x_pubkey.c
```diff
@@ -472,3 +472,32 @@ ASN1_BIT_STRING *X509_get0_pubkey_bitstr(const X509 *x)
         return NULL;
     return x->cert_info.key->public_key;
 }
+
+/*
+ * Returns 0 for equal, -1 for a < b or error on a, 1 for a > b or error on b,
```
bernd-edlinger, line 477: please simplify this function to return equal / nonequal / error or so.

crypto/x509/x_pubkey.c
t8m: It is actually public key data. But anyway, I agree that there seems to be no point in other than equal/non-equal/error results.

DDvO:
If we are only interested (in-)equality we should  return 1 for equal, 0 for non-equal, and < 0 on error and the function should not be called `*_cmp()` but `*_eq()`.

If we are interested in a comparison operation that assumes some order on the given datatype (even if it is just based in `memcmp()`) and can be used for building sorted container data types as done for `X509` we should return 1 for larger, 0 for equal, -1 for smaller, and < - 1 on error and the function should be named `_cmp()`.

In the improved version of `X509_PUBKEY_cmp()` a couple of days ago I tried to retain as much comparison information as possible, but the weird return value design of `EVP_PKEY_cmp()`, used to 'compare' public key data, results in the pretty awkward return values criticized above.
I'll try to find a real solution such that `X509_PUBKEY_cmp()` is really a comparison operator in the above sense, like `X509_cmp` for instance.

bernd-edlinger:
Hmm.  X509_cmp is a rather bad example.
We should not follow it at all.
I think it should be fixed as well, since
actually it looks completely broken.
The only use that works is check for == 0 means equal.
-2 can also be the return value of memcmp !
here:
https://github.com/openssl/openssl/blob/6e15b81c340e8124e45a92a2dbcd923c2ba4e79f/crypto/x509/x509_cmp.c#L145-L153

Note to myself: I must blame myself for trying to check the error of X509_check_purpose :-)

DDvO:
I meanwhile agree that the equality semantics makes more sense for `X509_PUBPKEY`, 
so I'll do that and consequently rename the function to `X509_PUBPKEY_eq()`.

crypto/x509/x_pubkey.c
DDvO: So I've changed the semantics from order comparison to equality, and renamed the function accordingly to `X509_PUBKEY_eq()`.

crypto/x509/x_pubkey.c
DDvO: Done

DDvO:
Merged - thanks to all who contributed to the discussion on the best solution.

crypto/asn1/ameth_lib.c
```diff
@@ -277,7 +277,7 @@ void EVP_PKEY_asn1_free(EVP_PKEY_ASN1_METHOD *ameth)
 
 void EVP_PKEY_asn1_set_public(EVP_PKEY_ASN1_METHOD *ameth,
                               int (*pub_decode) (EVP_PKEY *pk,
-                                                 X509_PUBKEY *pub),
+                                                 const X509_PUBKEY *pub),
```
romen, line 280: Do we consider this an API breaking change?

External applications will be required to change their source code, see this failure: https://travis-ci.org/github/romen/openssl/jobs/691216006

I am unsure if the fix is to update the gost engine to use the constified version (but this would also mean that sharing the codebase with 1.1.1 and 3.0.0-dev, this would require `#ifdef` guards) or to undo the constification.

romen: Ping @mattcaswell @t8m

beldmit: Current gost-engine upstream is not compatible with 1.1.1 already

romen: > Current gost-engine upstream is not compatible with 1.1.1 already

Ok, so this particular failure could be fixed by updating the gost source.

Yet, I want to raise the issue that any external app or engine that was using `EVP_PKEY_asn1_set_public()` will require a source change after the constification: it's not like we don't allow this for 3.0.0, but usually we have a conversation (if not a vote) about breaking API changes, and I don't think the matter was raised here.

It is perfectly acceptable to say that we will stick with the API change, I just want it to be acknowledged and approved explicitly.

beldmit: I agree with you.

bernd-edlinger: Actually I don't understand what renaming X509_PUBKEY_cmp to X509_PUBKEY_eq
has to do at all with constification of pub_decode functions ?

DDvO: This was needed  to allow 
`int X509_PUBKEY_eq(const X509_PUBKEY *a, const X509_PUBKEY *b)`
rather than
`int X509_PUBKEY_eq(X509_PUBKEY *a, X509_PUBKEY *b)`
and also independently of that function the constification should make much sense.

DDvO: It's really a pity that for so many OpenSSL API functions `const` types for read-only parameters were not done right from the beginning.

romen: > This was needed to allow
> `int X509_PUBKEY_eq(const X509_PUBKEY *a, const X509_PUBKEY *b)`
> rather than
> `int X509_PUBKEY_eq(X509_PUBKEY *a, X509_PUBKEY *b)`
> and also independently of that function the constification should make much sense.

I understand, and often constifying the arguments of a function do not cause source code compatibility issues!
But in this case, by changing the argument of a function pointer declaration we end up with a breaking API change.

I don't disagree with constification, and I wouldn't vote against the breaking change in this case, and I share the sentiment that it would have been better if it was const to begin with!

I just want to raise the issue that, as happened with the gost source, external applications will get failures out of this change and normally we require the reviewers to acknowledge and approve this, or even call a vote in some cases! 

> It's really a pity that for so many OpenSSL API functions `const` types for read-only parameters were not done right from the beginning.

Agreed!

DDvO: How to deal with the API incompatibility here? Should there a voting?
Fortunately `EVP_PKEY_asn1_set_public()` is an 'expert-only' function.
Shall I document that API-breaking change, and how best to do that?

DDvO: I just found that I missed to add the `const` in a further line of `EVP_PKEY_ASN1_METHOD.pod`:
```
 int (*pub_decode) (EVP_PKEY *pk, X509_PUBKEY *pub);
```

DDvO: BTW strange that for `priv_decode()` and `param_decode()` the input parameter was already `const` but not for `pub_decode()`.

At the latest when documenting those functions in commit 751148e2b0d1048522d754d55319166435cdda86
the `const` issue introduced in commit 0f113f3ee4d should have been rectified.

t8m: Or could we just cast-away the const when the callback is called? I know, ugly and should be documented at least.

DDvO: Again my question: How shall we proceed now with this (in my view minor) API incompatibility?
Is this something that we should decide/fix urgently before the release?

DDvO: BTW, a week ago I realized that @mattcaswell had brought up a very similar issue 11  months ago, which is still open, too: https://github.com/openssl/openssl/issues/9296

t8m:
As the pull request is merged, it should be closed. I'd suggest adding the new API break to the already existing #9296 and we can call a vote on how it should be handled.

PR #11895, opened 2020-05-21
Enable legacy AES API to use AESNI or VPAES if available

bernd-edlinger:
When no assembler support is available, we fall back
to either the constant time implementation or the non-constant time
C code implementation.

This is controlled by -DOPENSSL_AES_CONST_TIME.
So this makes the legacy API completely constant time,
if OPENSSL_AES_CONST_TIME is defined, otherwise it uses constant time
assembler implementaions when available, and may fall back to the
non-constant time implementaion.

This works so far only for intel and aarch64 CPUs.

[extended tests]

##### Checklist

### DIFF
--- a/crypto/aes/aes_cbc.c
+++ b/crypto/aes/aes_cbc.c
@@ -10,6 +10,13 @@
 #include <openssl/aes.h>
 #include <openssl/modes.h>
 
+#if defined(OPENSSL_CPUID_OBJ) && !defined(AES_ASM)
+void aes_cbc_encrypt(const unsigned char *in, unsigned char *out,
+                     size_t len, const AES_KEY *key,
+                     unsigned char *ivec, const int enc);
+# define AES_cbc_encrypt aes_cbc_encrypt
+#endif
+
 void AES_cbc_encrypt(const unsigned char *in, unsigned char *out,
                      size_t len, const AES_KEY *key,
                      unsigned char *ivec, const int enc)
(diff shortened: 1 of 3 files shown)


bernd-edlinger:
Travis failure is unrelated.

bernd-edlinger:
Note that this is for 1.1.1 only at the moment, but it depends on #9677 which
has not made any progress since more than 6 months.
I abandoned that PR by now, since I have other things to do.

bernd-edlinger:
With the second commit also the AES_cbc_encrypt API is able to use AESNI or VPAES.
This is very suprising by how much the LEGACY API is outperforming the EVP API:
LEGACY API: `openssl speed aes`
```
aes-128 cbc     214020.15k   216315.86k   235227.05k   240226.30k   241702.23k   241696.77k
aes-192 cbc     182358.41k   184561.07k   197459.03k   200981.85k   201981.95k   202020.18k
aes-256 cbc     158889.29k   160572.27k   170269.87k   172872.36k   173618.52k   173626.71k
```
EVP API: `openssl speed -evp aes-128-cbc`
```
aes-128-cbc     148546.79k   193050.69k   227180.46k   238088.87k   241426.43k   241423.70k
aes-192-cbc     131881.39k   164745.30k   191568.47k   199358.46k   201667.93k   201736.19k
aes-256-cbc     119544.65k   144924.27k   165435.73k   171423.06k   167753.05k   169213.95k
```
```

PR #11896, opened 2020-05-21
Prevent use after free of global_engine_lock

t8m:
If buggy application calls engine functions after cleanup of engines
already happened the global_engine_lock will be used although
already freed.

See for example:
https://bugzilla.redhat.com/show_bug.cgi?id=1831086


bernd-edlinger:
Travis failure is unrelated.

bernd-edlinger:
sorry wrong PR.

t8m:
Merged to master and 1.1.1.

PR #11868, opened 2020-05-19
Refactor the provider side DER constants and writers

levitte:
This splits up all the providers/common/der/*.c.in so the generated
portion is on its own and all related DER writing routines are in
their own files.  This also ensures that the DIGEST consstants aren't
reproduced in several files (resulting in symbol clashes).

Finally, the production of OID macros is moved to the generated header
files, allowing other similar macros, or DER constant arrays, to be
built on top of them.

PR #11906, opened 2020-05-22
Fix omissions in providers/common/der/build.info

levitte:
Dependencies on generated files must be declared explicitly.  When
refactoring the DER code in providers/common/der, a few of those
dependency declaration were omitted, which may lead to build errors in
a parallel build.

Some cleanup and extensive used of build.info variables is done while
at it, to avoid unnecessary repetition.

### DIFF
--- a/providers/common/der/build.info
+++ b/providers/common/der/build.info
@@ -1,36 +1,64 @@
-$FIPSABLE=\
-        der_rsa_gen.c der_rsa_key.c der_rsa_sig.c \
-        der_dsa_gen.c der_dsa_key.c der_dsa_sig.c \
-        der_ec_gen.c der_ec_key.c der_ec_sig.c \
-        der_digests_gen.c
+#----- Digests
+$DER_DIGESTS_H=../include/prov/der_digests.h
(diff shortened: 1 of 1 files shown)


levitte:
The issue was mentioned here: https://github.com/openssl/openssl/pull/11868#issuecomment-631838296

kroeckx:
I think OSS-Fuzz is actually having problems with this issue, so I've marked it as urgent.

levitte:
Okie. Now, all that's missing is an approvql

levitte:
Ping!

levitte:
> Is the $DER_RSA_FIPSABLE separated from $COMMON because it will never go into libcommon.a? It is not clear to me. However approving anyway.

Yes.  If you grep for `# *ifn?def  *FIPS_MODULE`, that's the only file in this directory that matches.
I *currently* can't place the others in libcommon, 'cause the link between that and libcrypto is not as clear as it should be, and that requires some intricate changes in the core of the build system.  It's on my mind.

levitte:
Merged

3f5ea7dc0ca4affb1fbe5c9f6d25add8aa3535b3 Fix omissions in providers/common/der/build.info

PR #11908, opened 2020-05-22
fips: add AES CFB & OFB mode ciphers to FIPS provider.

paulidale:
Put the AES-CFB & AES-OFB ciphers into the FIPS provider after confirmation from the FIPS lab that this is okay.


paulidale:
I'm not sure there is a place.  The relevant manual page (`man7/OSSL_PROVIDER-FIPS.pod`) only has this to say about AES modes:

```
=item AES, see L<EVP_CIPHER-AES(7)>
```

test/recipes/30-test_evp_data/evpciph.txt
```diff
@@ -259,31 +259,27 @@ Ciphertext = B2EB05E2C39BE9FCDA6C19078C6A9D1B
 #   AES-bits-CFB:key:IV/ciphertext':plaintext:ciphertext:encdec
 # CFB128-AES128.Encrypt
 Cipher = AES-128-CFB
-Availablein = default
```
levitte, line 262: Should anyone try this with the legacy provider only, the test will fail miserably.  Therefore, instead of removing this line, I'd rather add another with `Availablein = fips`

paulidale: The other AES cipher modes (except OFB) don't include an `Availablein` line at all.  Why is CFB special?

mattcaswell:
> =item AES, see L<EVP_CIPHER-AES(7)>

And that reference doesn't seem to even exist. That seems to indicate a whole in our documentation. Not to be resolved here though.

paulidale:
Yup :(

test/recipes/30-test_evp_data/evpciph.txt
paulidale: There is one instance of "Availablein = fips" in the EVP tests: to verify AES-XTS mode with duplicate keys errors out properly.

levitte:
> And that reference doesn't seem to even exist. That seems to indicate a whole in our documentation.

Yup, that's because no one has done the documentation move for EVP_CIPHER yet. I assume provider-cipher.pod still contains implementation specific details.

paulidale:
Travis failures are timeouts.

paulidale:
ping?

paulidale:
Merged to master, thanks for the input.

Issue #1102, opened 2016-05-20
OOB read and leak in b_print.c fmtfp()


guidovranken:
PoC:

```c
#include "bio.h"

int main(void)
{
    BIO *bio;
    bio = BIO_new_fp(stdout, BIO_NOCLOSE);
    BIO_printf(bio, "%f\n", 9223372036854775808.0);
}
```

This causes the ```intpart``` and ```fracpart``` inside ```fmtfp``` (crypto/bio/b_print.c) to be negative, thus causing OOB reads on:
```c
         iconvert[iplace++] = "0123456789"[intpart % 10];
```

and

```c
675         fconvert[fplace++] = "0123456789"[fracpart % 10];
```

You can put printf's there to see that array index is in fact negative and/or compile with ```clang -fsanitize=undefined```.

Guido

mattcaswell:
Thanks for the report. Fixed by commit 1b62d880b. Closing.

PR #11909, opened 2020-05-22
Cast ULONG_MAX to LDOUBLE before comparing

kroeckx:
clang-10 gives the following error:
crypto/bio/b_print.c:638:19: error: implicit conversion from 'unsigned long' to 'double' changes value from 18446744073709551615 to 18446744073709551616 [-Werror,-Wimplicit-int-float-conversion]
    if (ufvalue > ULONG_MAX) {
                ~ ^~~~~~~~~
/usr/lib/llvm-10/lib/clang/10.0.0/include/limits.h:57:37: note: expanded from macro 'ULONG_MAX'
                   ~~~~~~~~~~~~~~~~~^~~~

A double can only reperesent integers up to 2^53-1, a long double 2^65-1.

<!--
Thank you for your pull request. Please review these requirements:

Contributors guide: https://github.com/openssl/openssl/blob/master/CONTRIBUTING.md

Other than that, provide a description above this comment if there isn't one already

If this fixes a github issue, make sure to have a line saying 'Fixes #XXXX' (without quotes) in the commit message.
-->

##### Checklist
<!-- Remove items that do not apply. For completed items, change [ ] to [x]. -->

### DIFF
--- a/crypto/bio/b_print.c
+++ b/crypto/bio/b_print.c
@@ -635,7 +635,7 @@ fmtfp(char **sbuffer,
             fvalue = tmpvalue;
     }
     ufvalue = abs_val(fvalue);
-    if (ufvalue > ULONG_MAX) {
+    if (ufvalue > (LDOUBLE)ULONG_MAX) {
         /* Number too big */
         return 0;
     }


kroeckx:
I still need to check that this is actually correct

kroeckx:
There is also: https://github.com/openssl/openssl/blob/4d55122ee782ebd306ef492f50c9b41e41a56244/crypto/params.c#L411

crypto/bio/b_print.c
```diff
@@ -635,7 +635,7 @@ fmtfp(char **sbuffer,
             fvalue = tmpvalue;
     }
     ufvalue = abs_val(fvalue);
-    if (ufvalue > ULONG_MAX) {
+    if (ufvalue > (LDOUBLE)ULONG_MAX) {
```
paulidale, line 638: I don't think this is a proper solution.  `sizeof(long double) == sizeof(double)` on many platforms and the problem remains.

`ULONG_MAX + 1` can be exactly represented as a double.  Unfortunately, `ULONG_MAX + 1` isn't calculable directly.

I'll have a think about solutions....

kroeckx: Yes, I was just checking that. At least on my arm it's the same.

paulidale:
I'm unsure if the ones in `params.c` are genuine issues.
That the value increases by 1 wouldn't matter if the comparisons didn't include equality.  I'd expect the compiler to change `<=` to `<` if the RHS gets increased.

crypto/bio/b_print.c
kroeckx: I think changing it from > to >= will actually fix it, the implicit conversion seems to change it to the value we want.

kroeckx:
I don't expect the compiler to change the <= to <, it just casts the number.

crypto/bio/b_print.c
kroeckx: But that also doesn't actually fix anything. I think we just need to check for it being >= 1<<53, or something like that.

kroeckx: But it gets cast to an integer, and the the integer part gets subtracted again ... So maybe the patch will actually do the correct thing. I guess I currently don't understand the code good enough.

paulidale:
> I don't expect the compiler to change the <= to <, it just casts the number.

It is allowed to change e.g. `x <= 15` into `x < 16` which would be the equivalent if the integer size was four bits.  It is thus possible to emit correct code in spite of the accuracy loss.


It's too late to think too much here, I can think of some possible approaches:

* Use `frexp(3)` to decompose the floating point number and compare the parts against the range of integer values.  I think this ends up being a comparison of the exponent alone since the returned mantissa is always in the interval [½, 1).
* Use `ldexp(3)` to get an exact representation for `ULONG_MAX + 1` for comparison purposes.  This value is `ldexp(1.0, sizeof(ULONG_MAX) * 8)`.
* Use `ldexp(3)` to break up the ULONG_MAX and do the comparison in two steps.  For example, `ldexp(ULONG_MAX >> 32, 32)` would return a double that is representable without risking integer overflow.
* Compare logarithms of the values.  Calculating `ceil(log₂(ULONG_MAX))` isn't difficult and gives the same result as `log₂(ULONG_MAX + 1)`.

I kind of expect a sleepless night ahead while my mind mulls over these and no doubt other possibilities.

kroeckx:
I have to wonder why we refuse to print floating points that are larger than some value.

kroeckx:
So it seems this was added as a result of #1102, and was added in commit 1b62d880b5190de8c49a01837d96501cecf2a111

paulidale:
The reason is that we print reals as an unsigned long.  Without the range check, we produce incorrect output.

My TODO list has an item for implementing proper floating point output formatting.  Given how long this has been a TODO item, I'm not expecting to get to it any time soon.  @vdukhovni sent me a reference to a paper (which I can't find quickly) describing an efficient way to do this.

paulidale:
A quick solution would be to reduce the range allowed so that the conversion doesn't fail.

crypto/bio/b_print.c
paulidale: > But that also doesn't actually fix anything. I think we just need to check for it being >= 1<<53, or something like that.

Changing the comparison to 1<<53 seems reasonable.

It's not quite a fix: 1<<60 is exactly representable by both a double and a long integer.  Still, I think it's good enough for now.

paulidale:
I'll look into the OSSL_PARAM issues in a different PR.

PR #11912, opened 2020-05-22
Backport enhancement to OSSL_STORE_close()

DDvO:
as requested by @t8m in https://github.com/openssl/openssl/pull/11733#pullrequestreview-414357821

### DIFF
--- a/doc/man3/OSSL_STORE_open.pod
+++ b/doc/man3/OSSL_STORE_open.pod
@@ -94,6 +94,7 @@ OSSL_STORE_eof() shows that the end of data has been reached.
 OSSL_STORE_close() takes a B<OSSL_STORE_CTX>, closes the channel that was opened
 by OSSL_STORE_open() and frees all other information that was stored in the
 B<OSSL_STORE_CTX>, as well as the B<OSSL_STORE_CTX> itself.
+If B<ctx> is NULL it does nothing.
 
 =head1 SUPPORTED SCHEMES
 
--- a/doc/man3/OSSL_STORE_open.pod
+++ b/doc/man3/OSSL_STORE_open.pod
@@ -46,21 +46,22 @@ OSSL_STORE_close() to work together.
 
 =head2 Functions
 
-OSSL_STORE_open() takes a uri or path B<uri>, password UI method
-B<ui_method> with associated data B<ui_data>, and post processing
-callback B<post_process> with associated data B<post_process_data>,
+OSSL_STORE_open() takes a uri or path I<uri>, password UI method
+I<ui_method> with associated data I<ui_data>, and post processing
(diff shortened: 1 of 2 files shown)


doc/man3/OSSL_STORE_open.pod
```diff
@@ -94,6 +94,7 @@ OSSL_STORE_eof() shows that the end of data has been reached.
 OSSL_STORE_close() takes a B<OSSL_STORE_CTX>, closes the channel that was opened
 by OSSL_STORE_open() and frees all other information that was stored in the
 B<OSSL_STORE_CTX>, as well as the B<OSSL_STORE_CTX> itself.
+If B<ctx> is NULL it does nothing.
```
t8m, line 97: I wonder if this should be in a HISTORY section and mark there it is since 1.1.1h.
@levitte what do you think?

levitte: That sort of information is never wrong in my opinion

t8m: So perhaps keep this, and add HISTORY with a sentence such as `Handling of NULL I<ctx> in the B<OSSL_STORE_CTX> was introduced in OpenSSL 1.1.1h.`
Also please change `B<ctx>` to `I<ctx>`.

levitte: Er, if you change `B<ctx>` to `I<ctx>`, then I think the whole file should be cleaned up, not just this instance.

@DDvO, the guiding rule is that the things named in the NAME section should be bold (pod2xxx takes care of that for all function names if they are followed with `()`), all type and macro names should be **bold** except for NULL, which should have no marking at all, and all function parameter names should be *italics*.  See [man-pages(7)](http://man7.org/linux/man-pages/man7/man-pages.7.html)
(and no, we don't follow that entirely to the letter.  Most specifically, our SYNOPSIS sections are verbatim code snippets...  changing that is a *huge* job that I don't think we're ready for right now)

DDvO: Thanks @levitte for the explanation on the `B< >` doc styles. 
I've added the requested remark in the HISTORY section,
replaced `B< >` by `I< >` for arg names, and replaced `B<NULL>` by `NULL`.
I've also back-ported a remark on the use of `ui_data` 
and fixed a typo: `ui_data_data` -> `ui_data`.

levitte:
For commit messages, I would prefer it if " as in PR#11733" was removed from it.

DDvO:
> For commit messages, I would prefer it if " as in PR#11733" was removed from it.

Done

DDvO:
Pushed to OpenSSL_1_1_1-stable - thanks!

PR #11913, opened 2020-05-22
Re-introduce legacy EVP_PKEY types for provided keys

### LINKED ISSUE #11823: EVP_PKEY_base_id always returns 0 for generated pkey on master
```
#include <stdio>
#include <openssl/evp.h>
#include <openssl/rsa.h>

void main() {
    EVP_PKEY_CTX *ctx;
    EVP_PKEY *pkey = NULL;

    ctx = EVP_PKEY_CTX_new_id(EVP_PKEY_RSA, NULL);
    if (!ctx) {
        printf("1\n");
        exit(1);
    }
        
    if (EVP_PKEY_keygen_init(ctx) <= 0) {
        printf("2\n");
        exit(1);
    }

    if (EVP_PKEY_CTX_set_rsa_keygen_bits(ctx, 2048) <= 0) {
        printf("3\n");
        exit(1);
    }

    /* Generate key */
    if (EVP_PKEY_keygen(ctx, &pkey) <= 0) {
        printf("4\n");
        exit(1);
    }

    printf("key_type %d\n", EVP_PKEY_base_id(pkey));
}
```

This prints `0` for master branch but correctly prints `6` for previous versions.

levitte:
EVP_PKEYs with provider side internal keys got the key type
EVP_PKEY_NONE.  This turned out to be too disruptive, so we try
instead to find a matching EVP_PKEY_ASN1_METHOD and use whatever
EVP_PKEY type it uses.

To make internal coding easier, we introduce a few internal macros to
distinguish what can be expected from a EVP_PKEY:

- evp_pkey_is_unknown(), to detect an unassigned EVP_PKEY.
- evp_pkey_is_typed(), to detect that an EVP_PKEY has been assigned a
  type, which may be an old style type number or a EVP_KEYMGMT method.
- evp_pkey_is_assigned(), to detect that an EVP_PKEY has been assigned
  an key value.
- evp_pkey_is_legacy(), to detect that the internal EVP_PKEY key is a
  legacy one, i.e. will be handled via an EVP_PKEY_ASN1_METHOD and an
  EVP_PKEY_METHOD.
- evp_pkey_is_provided(), to detect that the internal EVP_PKEY key is
  a provider side one, i.e. will be handdled via an EVP_KEYMGMT and
  other provider methods.

This also introduces EVP_PKEY_KEYMGMT, to indicate that this EVP_PKEY
contains a provider side key for which there are no known
EVP_PKEY_ASN1_METHODs or EVP_PKEY_METHODs, i.e. these can only be
handled via EVP_KEYMGMT and other provider methods.

Fixes #11823

### DIFF
--- a/crypto/evp/pmeth_lib.c
+++ b/crypto/evp/pmeth_lib.c
@@ -158,7 +158,7 @@ static EVP_PKEY_CTX *int_ctx_new(OPENSSL_CTX *libctx,
      * If the key doesn't contain anything legacy, then it must be provided,
      * so we extract the necessary information and use that.
      */
-    if (pkey != NULL && pkey->type == EVP_PKEY_NONE) {
+    if (pkey != NULL && evp_pkey_is_provided(pkey)) {
         /* If we have an engine, something went wrong somewhere... */
         if (!ossl_assert(e == NULL))
             return NULL;
--- a/crypto/evp/pmeth_lib.c
+++ b/crypto/evp/pmeth_lib.c
@@ -155,8 +155,8 @@ static EVP_PKEY_CTX *int_ctx_new(OPENSSL_CTX *libctx,
         goto common;
 
     /*
-     * If the key doesn't contain anything legacy, then it must be provided,
-     * so we extract the necessary information and use that.
+     * If the internal key is provided, we extract the keytype from its
+     * keymgmt and skip over the legacy code.
      */
     if (pkey != NULL && evp_pkey_is_provided(pkey)) {
         /* If we have an engine, something went wrong somewhere... */
--- a/include/crypto/evp.h
+++ b/include/crypto/evp.h
@@ -526,7 +526,7 @@ const EVP_CIPHER *EVP_##cname##_ecb(void) { return &cname##_ecb; }
  *
  *     keymgmt != NULL
  */
-#define evp_pkey_is_unknown(pk)                                 \
+#define evp_pkey_is_blank(pk)                                   \
     ((pk)->type == EVP_PKEY_NONE && (pk)->keymgmt == NULL)
 #define evp_pkey_is_typed(pk)                                   \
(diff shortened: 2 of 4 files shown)


crypto/evp/pmeth_lib.c
```diff
@@ -158,7 +158,7 @@ static EVP_PKEY_CTX *int_ctx_new(OPENSSL_CTX *libctx,
      * If the key doesn't contain anything legacy, then it must be provided,
      * so we extract the necessary information and use that.
```
mattcaswell, line 159: This comment could probably do with an update

include/crypto/evp.h
```diff
@@ -518,9 +518,25 @@ const EVP_CIPHER *EVP_##cname##_ecb(void) { return &cname##_ecb; }
...
- * The easiest way to detect a provider side key is:    keymgmt != NULL
+ * The easiest way to detect a legacy key is:
+ *
+ *     keymgmt == NULL && type != EVP_PKEY_NONE
+ *
+ * The easiest way to detect a provider side key is:
+ *
+ *     keymgmt != NULL
  */
+#define evp_pkey_is_unknown(pk)                                 \
```
mattcaswell, line 529: "unknown" suggests to me that it is something, but we don't know what it is. The test here suggests that the key is blank rather than unknown. Perhaps a different name?

levitte: Hmm, so `evp_pkey_is_blank()`?

crypto/evp/pmeth_lib.c
levitte: Done

include/crypto/evp.h
levitte: Done

levitte:
The Travis failures are not relevant for this PR

levitte:
Merged

5e5bc836fbc5b1c0af428864f5286bbb225f7baf Re-introduce legacy EVP_PKEY types for provided keys

PR #11916, opened 2020-05-22
Fix auto-gen names to ignore

richsalz:
Not build-breaking but annoying.


t8m:
Merged to master as e919166927
Thank you!

PR #11920, opened 2020-05-22
Pass zero-length parameters to providers

kaduk:
This was reported to me as a failure of HKDF with empty IKM (that appears to be a regression from 1.1.1), but also can manifest in other places, such as the RSA OAEP CMS test that I had to provide a fix for.

In general, we have to be able to differentiate between parameter-absent and parameter-present-but-zero-length: some protocols exist that mandate different behavior in these cases.  This approach currently feels the most natural to me, though I could be persuaded that an alternate approach is preferred, as I also think that there are aspects of this approach that are ... not exactly pretty.

### DIFF
--- a/crypto/params.c
+++ b/crypto/params.c
@@ -788,8 +788,6 @@ static int get_string_internal(const OSSL_PARAM *p, void **val, size_t max_len,
     if (used_len != NULL)
         *used_len = sz;
 
-    if (sz == 0)
-        return 1;
     if (p->data == NULL)
         return 0;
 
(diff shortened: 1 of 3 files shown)


kaduk:
Oops, the "regression from 1.1.1 for HKDF" seems to be premature (and I'll need to adjust the commit message); I had started my development on master but forgotten the details of the report I actually received.
That said, the main thrust of the change still seems valid.

crypto/params.c
```diff
@@ -788,21 +788,22 @@ static int get_string_internal(const OSSL_PARAM *p, void **val, size_t max_len,
...
     if (*val == NULL) {
-        char *const q = OPENSSL_malloc(sz);
+        char *const q = OPENSSL_malloc(sz > 0 ? sz : 1);
 
         if (q == NULL)
             return 0;
         *val = q;
-        memcpy(q, p->data, sz);
+        if (sz == 0)
+            *q = '\0';
```
levitte, line 804: This isn't necessary, as only octet strings may have zero length.  UTF8 string lengths should include a terminating NUL byte, so the "empty" UTF8 string is 1 byte, not 0.

kaduk: It was only intended as "don't pass uninitialized memory anywhere", though I could be convinced that if it stays inside a single provider it doesn't matter.

kaduk:
force-pushed with updated commit messages and the change that @levitte requested.

kaduk:
Merged to master; closing

PR #11921, opened 2020-05-22
[crypto/ec] respect enable-ec_nistp_64_gcc_128 option

bbbrumley:
Quoting from `INSTALL.md`:

```
### enable-ec_nistp_64_gcc_128

Enable support for optimised implementations of some commonly used NIST
elliptic curves.
```

But changes from 4d3fa06fce52682bfbc503c7ded2d0289e3f8cde trump that user option, which is misleading.

If I explictly say `enable-ec_nistp_64_gcc_128`, I expect to get the `EC_METHOD` from `ecp_nistp256.c`.

Getting the `EC_METHOD` in `ecp_nistz256.c` (if your arch supports it) is easy -- just don't pass `enable-ec_nistp_64_gcc_128` and it defaults to false. Making OpenSSL not trump your explicit configure option is not.

If users prefer tried and true pure C implementations versus manually written assembly, you should respect that option.

### DIFF
--- a/crypto/ec/ec_curve.c
+++ b/crypto/ec/ec_curve.c
@@ -2858,12 +2858,12 @@ static const ec_list_element curve_list[] = {
     {"prime192v1", NID_X9_62_prime192v1, &_EC_NIST_PRIME_192.h, 0,
      "NIST/X9.62/SECG curve over a 192 bit prime field"},
     {"prime256v1", NID_X9_62_prime256v1, &_EC_X9_62_PRIME_256V1.h,
-# if defined(ECP_NISTZ256_ASM)
+# if !defined(OPENSSL_NO_EC_NISTP_64_GCC_128)
+     EC_GFp_nistp256_method,
+# elif defined(ECP_NISTZ256_ASM)
      EC_GFp_nistz256_method,
 # elif defined(S390X_EC_ASM)
      EC_GFp_s390x_nistp256_method,
-# elif !defined(OPENSSL_NO_EC_NISTP_64_GCC_128)
-     EC_GFp_nistp256_method,
 # else
      0,
 # endif


romen:
If we do this is there an easy way to let the user configure a build so that they end up with the following?

- `EC_GFp_nistp224_method` for P-224
- `EC_GFp_nistz256_method` for P-256
- `EC_GFp_nistp521_method` for P-521

This combination is the current default in builds that do `enable-ec_nistp_64_gcc_128` and also have the nistz256 implementation available, and I suppose is currently the optimal one from a performance point of view (and is sound also from a security PoV).

kroeckx:
This PR seems to change the prefered if both are enabled from nistz256 to nistp256. My understanding is that if both are available that we should prefer nistz256, so I don't understand the motivation for this PR.

kroeckx:
I guess I should properly read your comment. I think the option to prefer C over assembler is called `no-asm`. To me the `enable-ec_nistp_64_gcc_128` means my compiler and platform supports that C version and I want to build it so I can use it for the cases where it's the best option. The option is called "enable", not "use".

bbbrumley:
> and I want to build it so I can use it for the cases where it's the best option.

It's actually currently the only option if you want [formal verification guarantees](https://www.iis.sinica.edu.tw/~bywang/papers/ase19.pdf) for P-224, P-256, and P-521.

So that is really what grinds my gears about this chunk of code.

* User: "I want that formally verified code."
* OpenSSL: "No, silently take this hand coded assembly instead."

I can't be the only one who sees the problem with that security wise?

I understand the `no-asm` argument but that is library wide and very heavy handed to reach these very specific code paths isolated to the `ec` module.

bbbrumley:
> If we do this is there an easy way to let the user configure a build so that they end up with the following

It's a good point. I don't know the answer. Although I disagree with

> sound also from a security PoV

because it's a formally verified vs not metric.

romen:
> Although I disagree with
> 
>> sound also from a security PoV
>
> because it's a formally verified vs not metric.

That's a very good point! 


What if instead we were to add a `NO_ECP_NISTZ256_ASM` define guard so that the user has an option to explicitly disable nistz also on the platforms that have support for it?

romen:
Or we can think of a 'PREFER_ECP_NISTZ256_ASM` to make the default behave as preferring the formally verified implementation, while giving an option to the user to pick NISTZ256 over the C implementation.

In the end I agree that `ECP_NISTZ256_ASM` doesn't offer the same guarantees as `ec_nistp_64_gcc_128`, but users might still want to choose the latter for performance in the absence of reported flaws in it.

kroeckx:
It's the first time I've read that paper, and that I've heard about crytoline (https://github.com/fmlab-iis/cryptoline). In there examples they actually seem to have nitz256 too.

kroeckx:
I think this is the relevant paper about it: https://www.iis.sinica.edu.tw/papers/byyang/21728-F.pdf

bbbrumley:
> I think this is the relevant paper

Thanks Kurt -- that one is new to me. Let me have a read through it over the weekend.

kroeckx:
There is also: http://precision.moscito.org/by-publ/recent/verify-gcc.pdf

kroeckx:
So I've been playing with the tools that are available. The one
for C looks like it might become useful. The one for assembler is
disappointing. If I had the time for it, I think I can write
something that's better and more useful for assembler.

bbbrumley:
@kroeckx So I flipped through these. I wasn't aware of any formal verification effort surrounding `ecp_nistz256` -- thanks again for the pointers.

Given that fact, I think the current configuration in OpenSSL makes sense and is trying to do the right thing security wise, not just performance wise as I originally though.

So I'm closing this one. Thanks for the discourse

kroeckx:
I actually started writing my own software to verify assembler.
So far it actually doesn't look hard to write it. I'm just wondering
if the SMT solver will be able to do it in a reasonable time.

PR #11926, opened 2020-05-23
CORE: make sure activated fallback providers stay activated

levitte:
Calling 'OSSL_PROVIDER_available(NULL, "default")' would search for
the "default" provider, and in doing so, activate it if necessary,
thereby detecting that it's available...  and then immediately free
it, which could deactivate that provider, even though it should stay
available.

We solve this by incrementing the refcount for activated fallbacks one
extra time, thereby simulating an explicit OSSL_PROVIDER_load(), and
compensate for it with an extra ossl_provider_free() when emptying the
provider store.

### DIFF
--- a/crypto/provider_core.c
+++ b/crypto/provider_core.c
@@ -104,6 +104,23 @@ struct provider_store_st {
     unsigned int use_fallbacks:1;
 };
 
+/*
+ * provider_deactivate_free() is a wrapper around ossl_provider_free()
+ * that also makes sure that activated fallback providers are deactivated.
+ * This is simply done by freeing them an extra time, to compensate for the
+ * refcount that provider_activate_fallbacks() gives them.
+ * Since this is only called when the provider store is being emptied, we
+ * don't need to care about any lock.
+ */
+static void provider_deactivate_free(OSSL_PROVIDER *prov)
+{
+    int extra_free = (prov->flag_initialized && prov->flag_fallback);
+
+    if (extra_free)
+        ossl_provider_free(prov);
+    ossl_provider_free(prov);
+}
+
 static void provider_store_free(void *vstore)
 {
     struct provider_store_st *store = vstore;
@@ -111,7 +128,7 @@ static void provider_store_free(void *vstore)
     if (store == NULL)
         return;
     OPENSSL_free(store->default_path);
-    sk_OSSL_PROVIDER_pop_free(store->providers, ossl_provider_free);
+    sk_OSSL_PROVIDER_pop_free(store->providers, provider_deactivate_free);
     CRYPTO_THREAD_lock_free(store->lock);
     OPENSSL_free(store);
 }
@@ -654,13 +671,18 @@ static void provider_activate_fallbacks(struct provider_store_st *store)
             OSSL_PROVIDER *prov = sk_OSSL_PROVIDER_value(store->providers, i);
 
             /*
-             * Note that we don't care if the activation succeeds or not.
-             * If it doesn't succeed, then any attempt to use any of the
-             * fallback providers will fail anyway.
+             * Activated fallback providers get an extra refcount, to
+             * simulate a regular load.
+             * Note that we don't care if the activation succeeds or not,
+             * other than to maintain a correct refcount.  If the activation
+             * doesn't succeed, then any future attempt to use the fallback
+             * provider will fail anyway.
              */
             if (prov->flag_fallback) {
                 activated_fallback_count++;
-                provider_activate(prov);
+                if (ossl_provider_up_ref(prov)
+                    && !provider_activate(prov))
+                    ossl_provider_free(prov);
             }
         }
 
--- a/crypto/provider_core.c
+++ b/crypto/provider_core.c
@@ -681,13 +681,14 @@ static void provider_activate_fallbacks(struct provider_store_st *store)
              * provider will fail anyway.
              */
             if (prov->flag_fallback) {
-                activated_fallback_count++;
-
-                if (ossl_provider_up_ref(prov)
-                    && !provider_activate(prov))
-                    ossl_provider_free(prov);
-                else
-                    prov->flag_activated_as_fallback = 1;
+                if (ossl_provider_up_ref(prov)) {
+                    if (!provider_activate(prov)) {
+                        ossl_provider_free(prov);
+                    } else {
+                        prov->flag_activated_as_fallback = 1;
+                        activated_fallback_count++;
+                    }
+                }
             }
         }
 
--- a/crypto/provider_core.c
+++ b/crypto/provider_core.c
@@ -42,7 +42,8 @@ struct provider_store_st;        /* Forward declaration */
 struct ossl_provider_st {
     /* Flag bits */
     unsigned int flag_initialized:1;
-    unsigned int flag_fallback:1;
+    unsigned int flag_fallback:1; /* Can be used as fallback */
+    unsigned int flag_activated_as_fallback:1;
 
     /* OpenSSL library side data */
     CRYPTO_REF_COUNT refcnt;
@@ -114,7 +115,8 @@ struct provider_store_st {
  */
 static void provider_deactivate_free(OSSL_PROVIDER *prov)
 {
-    int extra_free = (prov->flag_initialized && prov->flag_fallback);
+    int extra_free = (prov->flag_initialized
+                      && prov->flag_activated_as_fallback);
 
     if (extra_free)
         ossl_provider_free(prov);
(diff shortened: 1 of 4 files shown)


levitte:
My branch includes a tag 'provider-fallback-test', which includes only the test, not the fix.  You can use that to see the test in action before and after the fix.

levitte:
Added a couple of fixups...  it turned out that my first attempt made fallback cleanup to work right, but started failing on explicit loads of the default provider.  As a consequence, the added test now also tests that.  The failure was ugly, as it resulted in access of already freed memory.

slontis:
windoze doesnt seem to like unsetenv

levitte:
> windoze doesnt seem to like unsetenv

[sigh]

According to Microsoft documentation, it seems like `_putenv()` can be used to remove an environment variable (the same trick can't be used on Unix, I checked).

levitte:
That did fix it, it sseems.

levitte:
Ping?

crypto/provider_core.c
```diff
@@ -654,13 +673,21 @@ static void provider_activate_fallbacks(struct provider_store_st *store)
...
              */
             if (prov->flag_fallback) {
                 activated_fallback_count++;
-                provider_activate(prov);
+
+                if (ossl_provider_up_ref(prov)
+                    && !provider_activate(prov))
+                    ossl_provider_free(prov);
+                else
+                    prov->flag_activated_as_fallback = 1;
```
mattcaswell, line 690: So we set this to 1 if `ossl_provider_up_ref(prov)` returns false? That doesn't seem quite right

test/provider_fallback_test.c
```diff
@@ -0,0 +1,73 @@
...
+static int test_provider(OPENSSL_CTX *ctx)
+{
+    EVP_KEYMGMT *rsameth = NULL;
+    const OSSL_PROVIDER *prov = NULL;
+    int ok;
+
+    ok = TEST_true(OSSL_PROVIDER_available(ctx, "default"))
+        && TEST_ptr(rsameth = EVP_KEYMGMT_fetch(ctx, "RSA", NULL))
+        && TEST_ptr(prov = EVP_KEYMGMT_provider(rsameth))
+        && TEST_str_eq(OSSL_PROVIDER_name(prov), "default");
```
mattcaswell, line 24: Have you tried running this test without your patch? Wouldn't this pass anyway?, i.e. `OSSL_PROVIDER_available()` would load the fallback provider, and then unload it again. But then wouldn't the `EVP_KEYMGMT_fetch` just reload it again because we still have no providers loaded? However if we had loaded (say) the legacy provider between checking  the availability of "default" and fetching RSA, then no fallbacks will be auto-loaded after the `OSSL_PROVIDER_available` check.

levitte: Yes I have tested.  If you look, you will see that the test commit comes first.  I also added a tag, as mentioned in https://github.com/openssl/openssl/pull/11926#issuecomment-633076426, exactly so you all could easily test for yourselves.

> i.e. `OSSL_PROVIDER_available()` would load the fallback provider, and then unload it again. But then wouldn't the `EVP_KEYMGMT_fetch` just reload it again because we still have no providers loaded?

No.  The way it's made is that there's *one* chance per library context to load the fallbacks.  As soon as an attempt has been made to activate the fallbacks, the provider store's `use_fallbacks` is zeroed, and remains that way for the lifetime of that store.
You expect something else...  so let me ask you, if you load the FIPS provider, use it for a bit, and then unload it, do you expect the fallbacks to be possible to activate again?  That's an expectation we've never talked about...

crypto/provider_core.c
levitte: Ah, no, that logic wasn't quite right.  Fixed, I hope it makes more sense now.

levitte:
Ping.

There's a question hanging in review, which I'll repeat here:

Currently, the activation of fallback providers is subject to a flag in the provider store (`use_fallbacks`), which is initially true, and is set to false as soon as a provider is loaded, and remains that way for the remainder of that store's life (i.e. the lifetime of the library context in which the provider store is attached).  In other words, it's a one-shot deal.

A different way to do this would be to allow fallback providers to be activated if the store currently has zero activated providers.  In other words, this would allow someone to load/activate a provider (subject of a library context), do some operations with it, then unload/deactivate it, and if anything is fetched after that, the fallback providers are again on-line for activation.

It's not terribly difficult to modify the core code to allow the latter alternative.

Thoughts?

slontis:
Sounds reasonable to me. The use_fallbacks flag could be a user settable flag only (enabled by default but able to be turned off - and then the null provider would not be needed at all).

```
if (use_fallbacks && provider_count == 0) {
}
```

test/provider_fallback_test.c
mattcaswell: Now that you've explained it that makes sense.

test/provider_fallback_test.c
```diff
@@ -0,0 +1,73 @@
...
+
+int setup_tests(void)
+{
+    /*
+     * We must ensure that there is no OPENSSL_CONF defined.  Otherwise,
+     * we risk that the configuration file contains statements that load
+     * providers, which defeats the purpose of this test.
+     */
+#ifndef _WIN32
+    unsetenv("OPENSSL_CONF");
```
mattcaswell, line 59: This seems unnecessary:

https://github.com/openssl/openssl/blob/a5a87011baeef71c86938a2bae54f89fbe99e5dc/test/run_tests.pl#L30

levitte: Some day, someone will make a change. I'd rather be extra safe.

mattcaswell: > Some day, someone will make a change. I'd rather be extra safe.

And in that event the test will likely fail, and that person will fix it. Should we do this with all our tests? I don't think so.

levitte: This test is *particularly sensitive* in this regard, as it tests a specific condition (fallbacks).  Not all our tests do that, specifically.

mattcaswell: There are lots of tests that would start failing under these conditions. I don't think this one is any more sensitive in that regards.

levitte: Huh?  The other tests might fail if `OPENSSL_CONF` is undefined, this test will most likely not be able to check what it's supposed to if `OPENSSL_CONF` *is **defined***!  How do you not see the difference?

levitte: For example, what happens if we decide to globally define `OPENSSL_CONF` with `test/default-and-legacy.cnf`?  Will the other tests fail?  What will happen with this one?

levitte: ...

I guess that the other alternative is to ensure I have OPENSSL_CONF defined to the NUL device, in the test recipe.

levitte: Done

mattcaswell: > For example, what happens if we decide to globally define OPENSSL_CONF with test/default-and-legacy.cnf? Will the other tests fail?

I thought that was an interesting question, so I tried it out:

````
Test Summary Report
-------------------
20-test_provider.t               (Wstat: 256 Tests: 9 Failed: 1)
  Failed test:  9
  Non-zero exit status: 1
90-test_sslapi.t                 (Wstat: 512 Tests: 3 Failed: 2)
  Failed tests:  1, 3
  Non-zero exit status: 2
Files=198, Tests=1993, 309 wallclock secs ( 4.74 usr  0.50 sys + 295.25 cusr 18.45 csys = 318.94 CPU)
Result: FAIL
````

So, it turns out, that my "lots of tests that would start failing under these conditions" was an exaggeration - although some do. However, I much prefer your new solution.

**Comment by levitte** [comment]
BTW, here's a recipe to run the test without the change in `crypto/provider_core.c` with a modern enough git (2.23):

``` console
$ git reset $(git merge-base --fork-point origin/master) -- crypto/provider_core.c
Unstaged changes after reset:
M	crypto/provider_core.c
# That git reset leaves the reset file in the staged index, and the modified file
# in the work tree.  The latter must be restored to its pre-change shape
$ git restore -W crypto/provider_core.c
```

It's easily brought back to this branch with a `git restore -S crypto/provider_core.c`.

test/provider_fallback_test.c
levitte: The trouble now is that if someone wants to try that particular test program manually, they must remember to define `OPENSSL_CONF` "properly".

(this is a pet peeve of mine, btw, that when running the tests verbosely, we don't display the values of diverse important environment variables, which makes manual testing a royal pain...  Hmmm, I should probably raise an issue for that)

levitte:
Merged

f995e5bdcd5d4c2143323ff78e4dcc0b95c5e024 TEST: Add provider_fallback_test, to test aspects of fallback providers
c8567c392c1dc3dd15651c0d2746a5b87b5a88dd CORE: make sure activated fallback providers stay activated

PR #11928, opened 2020-05-23
move EC_METHOD to internal-only

bbbrumley:
Last TODO item from #8332! Quoting from there:

* Entire concept of `EC_METHOD` should not be public facing.

It was surprisingly easy to update the tests.

~~It's WiP because I'll document everything when I get some high level confirmation that:~~

1. ~~It's OK to deprecate all these functions~~
2. ~~It's OK to add `EC_GROUP_get_field_type`~~

Edit: finally out of WiP after ea1cd6bfbce0a10102a275fa73536c91896e990f. Thanks @levitte @t8m and @romen for the feedback.

### DIFF
--- a/crypto/ec/ec_curve.c
+++ b/crypto/ec/ec_curve.c
@@ -3195,7 +3195,7 @@ static EC_GROUP *ec_group_new_from_data(OPENSSL_CTX *libctx,
 
     /* If no curve data curve method must handle everything */
     if (curve.data == NULL)
-        return EC_GROUP_new_ex(libctx,
+        return ec_group_new_ex(libctx,
                                curve.meth != NULL ? curve.meth() : NULL);
 
     if ((ctx = BN_CTX_new_ex(libctx)) == NULL) {
@@ -3218,7 +3218,7 @@ static EC_GROUP *ec_group_new_from_data(OPENSSL_CTX *libctx,
 
     if (curve.meth != 0) {
         meth = curve.meth();
-        if (((group = EC_GROUP_new_ex(libctx, meth)) == NULL) ||
+        if (((group = ec_group_new_ex(libctx, meth)) == NULL) ||
             (!(group->meth->group_set_curve(group, p, a, b, ctx)))) {
             ECerr(EC_F_EC_GROUP_NEW_FROM_DATA, ERR_R_EC_LIB);
             goto err;
@@ -3388,17 +3388,13 @@ int ec_curve_nid_from_params(const EC_GROUP *group, BN_CTX *ctx)
     unsigned char *param_bytes = NULL;
     const EC_CURVE_DATA *data;
     const EC_POINT *generator = NULL;
-    const EC_METHOD *meth;
     const BIGNUM *cofactor = NULL;
     /* An array of BIGNUMs for (p, a, b, x, y, order) */
     BIGNUM *bn[NUM_BN_FIELDS] = {NULL, NULL, NULL, NULL, NULL, NULL};
 
-    meth = EC_GROUP_method_of(group);
-    if (meth == NULL)
-        return -1;
     /* Use the optional named curve nid as a search field */
     nid = EC_GROUP_get_curve_name(group);
-    field_type = EC_METHOD_get_field_type(meth);
+    field_type = EC_GROUP_get_field_type(group);
     seed_len = EC_GROUP_get_seed_len(group);
     seed = EC_GROUP_get0_seed(group);
     cofactor = EC_GROUP_get0_cofactor(group);
--- a/crypto/ec/ec_lib.c
+++ b/crypto/ec/ec_lib.c
@@ -23,7 +23,7 @@
 
 /* functions for EC_GROUP objects */
 
-EC_GROUP *EC_GROUP_new_ex(OPENSSL_CTX *libctx, const EC_METHOD *meth)
+EC_GROUP *ec_group_new_ex(OPENSSL_CTX *libctx, const EC_METHOD *meth)
 {
     EC_GROUP *ret;
 
@@ -65,11 +65,13 @@ EC_GROUP *EC_GROUP_new_ex(OPENSSL_CTX *libctx, const EC_METHOD *meth)
     return NULL;
 }
 
-#ifndef FIPS_MODULE
+#ifndef OPENSSL_NO_DEPRECATED_3_0
+# ifndef FIPS_MODULE
 EC_GROUP *EC_GROUP_new(const EC_METHOD *meth)
 {
-    return EC_GROUP_new_ex(NULL, meth);
+    return ec_group_new_ex(NULL, meth);
 }
+# endif
 #endif
 
 void EC_pre_comp_free(EC_GROUP *group)
@@ -255,7 +257,7 @@ EC_GROUP *EC_GROUP_dup(const EC_GROUP *a)
     if (a == NULL)
         return NULL;
 
-    if ((t = EC_GROUP_new_ex(a->libctx, a->meth)) == NULL)
+    if ((t = ec_group_new_ex(a->libctx, a->meth)) == NULL)
         return NULL;
     if (!EC_GROUP_copy(t, a))
         goto err;
@@ -270,6 +272,7 @@ EC_GROUP *EC_GROUP_dup(const EC_GROUP *a)
         return t;
 }
 
+#ifndef OPENSSL_NO_DEPRECATED_3_0
 const EC_METHOD *EC_GROUP_method_of(const EC_GROUP *group)
 {
     return group->meth;
@@ -279,6 +282,7 @@ int EC_METHOD_get_field_type(const EC_METHOD *meth)
 {
     return meth->field_type;
 }
+#endif
 
 static int ec_precompute_mont_data(EC_GROUP *);
 
@@ -475,6 +479,11 @@ const BIGNUM *EC_GROUP_get0_field(const EC_GROUP *group)
     return group->field;
 }
 
+int EC_GROUP_get_field_type(const EC_GROUP *group)
+{
+    return group->meth->field_type;
+}
+
 void EC_GROUP_set_asn1_flag(EC_GROUP *group, int flag)
 {
     group->asn1_flag = flag;
@@ -602,8 +611,7 @@ int EC_GROUP_cmp(const EC_GROUP *a, const EC_GROUP *b, BN_CTX *ctx)
 #endif
 
     /* compare the field types */
-    if (EC_METHOD_get_field_type(EC_GROUP_method_of(a)) !=
-        EC_METHOD_get_field_type(EC_GROUP_method_of(b)))
+    if (EC_GROUP_get_field_type(a) != EC_GROUP_get_field_type(b))
         return 1;
     /* compare the curve name (if present in both) */
     if (EC_GROUP_get_curve_name(a) && EC_GROUP_get_curve_name(b) &&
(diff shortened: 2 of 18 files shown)


levitte:
This looks good to me.

bbbrumley:
I think the travis red cross is relevant, and to be honest when I submitted the PR I had no idea how this was actually working at all since I marked those public `EC_METHOD`s as deprecated. Yet both

1. `./config -d no-shared no-deprecated -Wall -Werror -Wl,--fatal-warnings`
2. `./config -d no-shared -Wall -Werror -Wl,--fatal-warnings`

succeeded, so I blissfully ignored.

But now I think I see the trick going on internally -- is this the right fix for the red cross?

```diff
diff --git a/crypto/ec/ecp_s390x_nistp.c b/crypto/ec/ecp_s390x_nistp.c
index 75c8475e69..edbad15cdd 100644
--- a/crypto/ec/ecp_s390x_nistp.c
+++ b/crypto/ec/ecp_s390x_nistp.c
@@ -7,6 +7,12 @@
  * https://www.openssl.org/source/license.html
  */
 
+/*
+ * EC_METHOD low level APIs are deprecated for public use, but still ok for
+ * internal use.
+ */
+#include "internal/deprecated.h"
+
 #include <stdlib.h>
 #include <string.h>
 #include <openssl/err.h>
```

levitte:
Yes, that's the trick.

bbbrumley:
> Yes, that's the trick.

Thanks @levitte! All green after 0bf4176966b0463a4c7580940f38b6501d93e4ba.

crypto/ec/ec_lib.c
```diff
@@ -475,6 +479,11 @@ const BIGNUM *EC_GROUP_get0_field(const EC_GROUP *group)
     return group->field;
 }
 
+int EC_GROUP_get_field_type(const EC_GROUP *group)
+{
+    return group->meth->field_type;
```
romen, line 484: Related: I think it is fine here not to check if group->meth is NULL before dereferencing it.

test/ectest.c
```diff
@@ -836,8 +802,7 @@ static int char2_curve_test(int n)
         || !TEST_true(BN_hex2bn(&p, test->p))
         || !TEST_true(BN_hex2bn(&a, test->a))
         || !TEST_true(BN_hex2bn(&b, test->b))
-        || !TEST_true(group = EC_GROUP_new(EC_GF2m_simple_method()))
-        || !TEST_true(EC_GROUP_set_curve(group, p, a, b, ctx))
```
romen, line 840: Sidenote: do you have thoughts about deprecating also EC_GROUP_set_curve? (as another PR in case it makes sense)

test/ectest.c
```diff
@@ -2170,7 +2119,7 @@ static int cardinality_test(int n)
 
     if (!TEST_ptr(ctx = BN_CTX_new())
         || !TEST_ptr(g1 = EC_GROUP_new_by_curve_name(nid))
-        || !TEST_ptr(g2 = EC_GROUP_new(EC_GROUP_method_of(g1)))) {
+        || !TEST_ptr(g2 = EC_GROUP_new_by_curve_name(nid))) {
```
romen, line 2122: I guess this test could be just dropped now?
I think it was explicitly testing EC_GROUP_method_of, now is not doing much really!

test/ectest.c
romen: To elaborate more:

Once `EC_GROUP_new(some_method)` function is deprecated, I don't see leftover reasons to set parameters after instantiation.

crypto/ec/ec_curve.c
```diff
@@ -3388,17 +3388,15 @@ int ec_curve_nid_from_params(const EC_GROUP *group, BN_CTX *ctx)
...
     const EC_CURVE_DATA *data;
     const EC_POINT *generator = NULL;
-    const EC_METHOD *meth;
     const BIGNUM *cofactor = NULL;
     /* An array of BIGNUMs for (p, a, b, x, y, order) */
     BIGNUM *bn[NUM_BN_FIELDS] = {NULL, NULL, NULL, NULL, NULL, NULL};
 
-    meth = EC_GROUP_method_of(group);
-    if (meth == NULL)
+    if (group->meth == NULL)
```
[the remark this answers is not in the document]
bbbrumley: I agree. It's just old habits die hard and it was quick edits preserving what was there :) But I'll update it

test/ectest.c
bbbrumley: I ... think I wrote that line. Further down in the function, it's manipulating `g2` to muck with the parameters, then recalculating them and checking vs `g1`. So `g1` is kind of the "control" `EC_GROUP` for the test.

romen: Oh, yeah, I missed how g2 was being used!

test/ectest.c
bbbrumley: I agree in the sense that ... the only time I've written `EC_GROUP_set_curve` in code is exactly the file you're looking at. To be able to reuse an `EC_GROUP` without re-initializing it.

But outside of a test, I don't see why anyone would do that.

That is ofc just my subjective opinion -- I have not checked for applications using `EC_GROUP_set_curve`.

test/ectest.c
bbbrumley: But yea you're right I was lazy there -- it would be better to not instantiate `g1` and `g2` literally from the same call, to keep them as independent as possible for a valid control. I'll adjust it

bbbrumley:
Right so latest red cross is because I switched `nistp_single_test` from a method-specific test per NIST GFp curve to a curve NID-specific test per NIST GFp curve. And the `ecp_s390x_nistp.c` `EC_METHOD`s do not support precomp. I'll push a fixup.

test/ectest.c
romen: Maybe we can add this item to the parent issue you were using to track all these related PRs?
We can try to do an analysis of that and see if deprecation can be recommended.

test/ectest.c
```diff
@@ -1396,7 +1344,7 @@ static int nistp_single_test(int idx)
         || !TEST_false(EC_GROUP_have_precompute_mult(NISTP))
         /* now repeat all tests with precomputation */
         || !TEST_true(EC_GROUP_precompute_mult(NISTP, ctx))
-        || !TEST_true(EC_GROUP_have_precompute_mult(NISTP))
+        || !TEST_int_ge(EC_GROUP_have_precompute_mult(NISTP), 0)
```
romen, line 1347: We can effectively strip this, right?
We are testing that the value is 0 or greater for a function that should return either 0 or 1.

test/ectest.c
```diff
@@ -1430,7 +1378,7 @@ static int nistp_single_test(int idx)
         || !TEST_int_eq(0, EC_POINT_cmp(NISTP, Q, G, ctx)))
       goto err;
 
-    r = group_order_tests(NISTP);
+    r = 1;
```
romen, line 1381: We are removing this, because it is effectively already tests in `prime_field_tests()` and this now doesn't add any value anymore, right?

bbbrumley: Yes exactly

test/ectest.c
bbbrumley: I added `EC_GROUP_set_curve` to the #8332 list for analysis.

test/ectest.c
bbbrumley: done in ab7aa90af9c1ceb4c8df71d33c4b4e5805efd6e2

bbbrumley:
> A few more comments

Thanks. I will start on the `pod` and `CHANGES.md` entries hopefully today.

romen:
Thanks for all the work so far!

crypto/ec/ec_lib.c
```diff
@@ -23,6 +23,7 @@
 
 /* functions for EC_GROUP objects */
 
+#ifndef OPENSSL_NO_DEPRECATED_3_0
 EC_GROUP *EC_GROUP_new_ex(OPENSSL_CTX *libctx, const EC_METHOD *meth)
 {
```
t8m, line 28: It does not make sense to make a new 3.0 function immediately deprecated. Should it be added at all?

include/openssl/ec.h
```diff
@@ -113,15 +115,15 @@ const EC_METHOD *EC_GF2m_simple_method(void);
  *  \param   meth   EC_METHOD to use
  *  \return  newly created EC_GROUP object or NULL in case of an error.
  */
-EC_GROUP *EC_GROUP_new_ex(OPENSSL_CTX *libctx, const EC_METHOD *meth);
+DEPRECATEDIN_3_0(EC_GROUP *EC_GROUP_new_ex(OPENSSL_CTX *libctx, const EC_METHOD *meth))
```
t8m, line 118: As said above - it does not make sense to immediately deprecate a new function so it should be converted to an internal function instead.

crypto/ec/ec_lib.c
romen: Good point!

@bbrumley I believe this will also affect your documentation edits!

include/openssl/ec.h
bbbrumley: tyvm @t8m ! I will now sanity check all of these, since that never even crossed my mind ...

bbbrumley: > it does not make sense to immediately deprecate a new function so it should be converted to an internal function instead

Done in 17468cdf115b92d47ecb29952ec3dbd14d5b7ebd. I had to rebase due to conflict but I didn't squash yet just to keep the progress.

I sanity checked the rest of the functions for good measure. They're all ancient, so this was the only one.

Thanks for catching it :)

util/libcrypto.num
```diff
@@ -4682,7 +4682,7 @@ ERR_set_error                           ?	3_0_0	EXIST::FUNCTION:
 ERR_vset_error                          ?	3_0_0	EXIST::FUNCTION:
 X509_get0_authority_issuer              ?	3_0_0	EXIST::FUNCTION:
 X509_get0_authority_serial              ?	3_0_0	EXIST::FUNCTION:
-EC_GROUP_new_ex                         ?	3_0_0	EXIST::FUNCTION:DEPRECATEDIN_3_0,EC
+EC_GROUP_new_ex                         ?	3_0_0	NOEXIST::FUNCTION:DEPRECATEDIN_3_0,EC
```
bbbrumley, line 4685: @t8m well above is `make update` result but I don't think it really makes sense?

t8m: If you remove manually the line and re-run make update, it should not reappear.

util/libcrypto.num
```diff
@@ -5094,4 +5093,6 @@ EVP_PKEY_CTX_set_rsa_pss_keygen_saltlen ?	3_0_0	EXIST::FUNCTION:RSA
 EVP_PKEY_CTX_set_rsa_pss_keygen_mgf1_md ?	3_0_0	EXIST::FUNCTION:RSA
 EVP_PKEY_CTX_set_rsa_pss_keygen_mgf1_md_name ?	3_0_0	EXIST::FUNCTION:RSA
 OSSL_PROVIDER_do_all                    ?	3_0_0	EXIST::FUNCTION:
+EC_GROUP_get_field_type                 ?	3_0_0	EXIST::FUNCTION:EC
+OSSL_CRMF_CERTID_dup                    ?	3_0_0	EXIST::FUNCTION:CRMF
```
t8m, line 5097: Apparently in some other PR a make update commit was missing? These additions should go into a separate commit.

bbbrumley: It's just an artifact of the rebase. It'll fall out when I squash stuff. (I also noted and looked into the same before I pushed.)

util/libcrypto.num
bbbrumley: perfect ty that works.

I'll do the `pod` and `CHANGES.md` edits now, then squash and this one should be ready.

bbbrumley:
So I added the doc and squashed. Renamed to take out of WiP. Let me know if there's anything else.

t8m:
Would you please rebase to fix the CHANGES.md conflict?

bbbrumley:
> Would you please rebase to fix the CHANGES.md conflict?

Done :)

romen:
Merging this now!

romen:
Merge to `master` as 23ccae80bd58adfe89e3e345414684eb82bdb531

Thanks all!

PR #11933, opened 2020-05-24
[WIP] Remove OPENSSL_ia32cap overrides in various test scripts

bernd-edlinger:
The removed override was: OPENSSL_ia32cap=~0x200000200000000
which disables AESNI codepaths and PCLMULQDQ (useful for ghash).
It is unclear why this was done, but it probably just hides bugs.

[extended tests]

It is unclear to me why some test cases are failing due to the AESNI bit.
@mattcaswell any ideas?

PR #11934, opened 2020-05-24
Remove OPENSSL_ia32cap overrides in various test scripts [1.1.1]

bernd-edlinger:
The removed override was: OPENSSL_ia32cap=~0x200000200000000
which disables AESNI codepaths and PCLMULQDQ (useful for ghash).
It is unclear why this was done, but it probably just hides bugs.

[extended tests]

@mattcaswell FYI: 1.1.1 has the same problem, just in case you thing that is due
to the new provider design.

### DIFF
--- a/engines/e_ossltest.c
+++ b/engines/e_ossltest.c
@@ -319,6 +319,7 @@ static int bind_ossltest(ENGINE *e)
         return 0;
     }
 
+    OBJ_NAME_remove("AES-128-CBC-HMAC-SHA1", OBJ_NAME_TYPE_CIPHER_METH);
     return 1;
 }
 
--- a/test/recipes/70-test_comp.t
+++ b/test/recipes/70-test_comp.t
@@ -27,7 +27,6 @@ plan skip_all => "$test_name needs the sock feature enabled"
 plan skip_all => "$test_name needs TLSv1.3 or TLSv1.2 enabled"
     if disabled("tls1_3") && disabled("tls1_2");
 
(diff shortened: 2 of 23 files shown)


bernd-edlinger:
Taking out of WIP, the straight forward fix to remove AES-128-CBC-HMAC-SHA1
and AES-128-CBC-HMAC-SHA256 in the e_ossltest engine works for 1.1.1!
The remaining CI failures are unrelated (out of diskspace or so),
but unfortunately the same trick does not work for master.
As @kroeckx pointed out the e_ossltest engine must first be rewritten from
engine to provider, so I think that can be left as a TODO for now.

t8m:
Is there equivalent PR for master branch? I do not think this is cleanly cherry-pickable and it cannot be merged to 1.1.1 unless done in master first.

levitte:
> It is unclear why this was done, but it probably just hides bugs.

To test that things work when those features aren't available, I presume

t8m:
> To test that things work when those features aren't available, I presume

I do not think so. It could be like that if the testcases were run both with the override and without it but apparently they are run only with the overrride.

bernd-edlinger:
Yes, indeed. This is all about the AES-128-CBC-HMAC-SHA1 cihper, which gets in the way of the ossltest engine
doing the plain text encryption.
The trouble is that something is preventing the master version #11933 to work correctly.
I can imagine that it is not possible to remove a cipher from the default provider,
but I do not understand why it does not even work if I implement a complete
AES-128-CBC-HMAC-SHA1 replacement as in #16693 while the exact identical #16694 works.

PR #11941, opened 2020-05-24
Coverity 1463830: Resource leaks (RESOURCE_LEAK)

paulidale:
Free the stack if it's been allocated on error path.


paulidale:
Not for 1.1.1, this isn't an issue in 1.1.1.  The return code checking for the up_ref call was added in master only.

paulidale:
Merged to master thanks for the reviews.

bernd-edlinger:
> Not for 1.1.1, this isn't an issue in 1.1.1. The return code checking for the up_ref call was added in master only.

No I know for sure, since I introduced this code just a few days ago,
and I cherry-picked it to 1.1.1.

Please cherry-pick it to 1.1.1

bernd-edlinger:
Although, I needed to do some hand-work, after the cherry-pick,
so it was a PR of it's own. see 5cea5841 and #11852

paulidale:
You're correct, my 1.1.1 branch hadn't been updated recently enough.
I've cherrypicked this to 1.1.1 as well.  It merged cleanly.

PR #11943, opened 2020-05-24
bio printf: Safely range check real overflow on output.

paulidale:
There is a problem casting ULONG_MAX to double which gcc-10 is warning about.
ULONG_MAX typically cannot be exactly represented as a double.  ULONG_MAX + 1
can be and this fix uses the latter, however constructing this number cannot
be done using integer arithmetic.

This is an alternative fix to #11909 

The same problem exists in 1.1.1.

### DIFF
--- a/crypto/bio/b_print.c
+++ b/crypto/bio/b_print.c
@@ -9,6 +9,7 @@
 
 #include <stdio.h>
 #include <string.h>
+#include <math.h>
 #include "internal/cryptlib.h"
 #include "crypto/ctype.h"
 #include "internal/numbers.h"
@@ -635,10 +636,16 @@ fmtfp(char **sbuffer,
             fvalue = tmpvalue;
     }
     ufvalue = abs_val(fvalue);
-    if (ufvalue > ULONG_MAX) {
-        /* Number too big */
+    /*
+     * Check if the real value is too large to be displayed via an
+     * unsigned long int.
+     *
+     * The obvious comparison ufvalue > ULONG_MAX doesn't work because
+     * ULONG_MAX cannot be converted to a double without accuracy loss.
+     */
+    if (ufvalue >= ldexp(1., sizeof(unsigned long) * 8))
         return 0;
-    }
+
     intpart = (unsigned long)ufvalue;
 
     /*


crypto/bio/b_print.c
```diff
@@ -635,10 +636,16 @@ fmtfp(char **sbuffer,
...
-    if (ufvalue > ULONG_MAX) {
-        /* Number too big */
+    /*
+     * Check if the real value is too large to be displayed via an
+     * unsigned long int.
+     *
+     * The obvious comparison ufvalue > ULONG_MAX doesn't work because
+     * ULONG_MAX cannot be converted to a double without accuracy loss.
+     */
+    if (ufvalue >= ldexp(1., sizeof(unsigned long) * 8))
```
levitte, line 646: I hope that function is available everywhere...

paulidale: It's present in C89 and I think it was in K&R.

What I'm not certain about is compilers optimising away the call and avoiding libm being a dependency.

bernd-edlinger: did you also consider `if (ufvalue >= 18446744073709551616.0) {` ?

bernd-edlinger: Hmm, we don't know for sure what ULONG_MAX is, but would this work?
`if (ifValue >= (double)(ULONG_MAX - 65535) + 65536.0) {`

Idea: ULONG_MAX-65535 cancels 16 bits, so that should be
representable as a 53 bit float and adding 65536.0 is also an exact result.

richsalz: C89?  You sure? "    The ldexp() functions conform to ISO/IEC 9899:2011."

Adding -lm as a dependency is a deal-breaker, IMO.

richsalz: Yes, I did some checking and ldexp is in v7 unix but in -lm.

Maybe change the docs/code to say "doesn't work within 1K of MAX values"

bernd-edlinger: I would also like to avoid a new linker dependency whatever it costs.
Especially when this should be back-ported to the 1.1.1 LTS branch.
You must at least subtract 2K-1 from ULONG_MAX, to get the next lower
double number.  I think IEEE double has 53 bit mantissa, while u_long has 64 bit, so
64 bit - 53 bit = 11, so 2K = 2^11 would make sense.
FYI my last comment was not meant as a joke ;-)
I think that should really do the trick.

kroeckx:
Note that it was clang-10, not gcc-10, that generated that warning.

crypto/bio/b_print.c
levitte: > C89? You sure? " The ldexp() functions conform to ISO/IEC 

Different manuals say different things.
The OpenBSD manual says C89: https://man.openbsd.org/ldexp.3
The Linux manual says C99 and POSIX-1: http://man7.org/linux/man-pages/man3/ldexp.3.html

richsalz: It was in V7, according to https://s3.amazonaws.com/plan9-bell-labs/7thEdMan/vol1/man3.bun.txt

petrovr:
Pauli wrote:
> There is a problem casting ULONG_MAX to double which gcc-10 is warning about.
> ULONG_MAX typically cannot be exactly represented as a double.  ULONG_MAX + 1
> can be and this fix uses the latter, however constructing this number cannot
> be done using integer arithmetic.
>
> This is an alternative fix to #11909

Both patches does not look correct.
Idea to add extra dependency to "math" library just for a warning is awful. More over real issue could be only for some 64-bit cases.

Let see number 0x1p+64 . If long is 8 bytes mathematically this number is equal to ULONG_MAX + 1.
The problem is that double is with 52 bit mantissa. So let see the number 0x1.fffffffffffffp+63 where all bits in mantissa are set one. This is the double that could be converted to 8 bytes long unsigned long. It is equal to ULONG_MAX - ;) .

If long is 4 bytes. If I remember well 64 bit windows also user 32 bit long.
There is no issue as even 0x1.fffffffffffffp+31 is save to be converted.


I'm not sure what is best way to correct issue without to check size of long and length of mantissa.


Regards,
Roumen Petrov

bernd-edlinger:
My alternate attempt to fix: #11955

paulidale:
Deferring to #11955.

crypto/bio/b_print.c
paulidale: > did you also consider `if (ufvalue >= 18446744073709551616.0) {`

Yes.


I do remember the `ldexp()` function being around very early.  I don't have a copy of K&R anymore so couldn't check.  The _float_ and _long double_ versions are far more recent.

PR #11946, opened 2020-05-24
WIP: Build: add a make target to build the openssl application

### LINKED ISSUE #11932: The openssl app can't be built directly from a clean state
In https://github.com/openssl/openssl/pull/11682#issuecomment-631241508, @paulidale made a remark that it is not possible to build `apps/openssl` directly from a clean state:

```
make clean && ./config && make apps/openssl
```
This made me wonder whether this actually worked in master and 1.1.1, and it turned out this was not the case either.

**master**

```
~/src/openssl$ make apps/openssl
/usr/bin/perl apps/progs.pl -C "apps/openssl" > apps/progs.c
/usr/bin/perl apps/progs.pl -H "apps/openssl" > apps/progs.h
gcc  -Iapps -I. -Iinclude -Iapps/include  -pthread -m64 -Wa,--noexecstack -Wall -O0 -g -DOPENSSL_BUILDING_OPENSSL  -MMD -MF apps/openssl-bin-asn1pars.d.tmp -MT apps/openssl-bin-asn1pars.o -c -o apps/openssl-bin-asn1pars.o apps/asn1pars.c
In file included from ./e_os.h:14,
                 from apps/include/apps.h:13,
                 from apps/asn1pars.c:13:
include/openssl/opensslconf.h:13:10: fatal error: openssl/configuration.h: No such file or directory
          ^~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:22463: apps/openssl-bin-asn1pars.o] Error 1
```

**1.1.1**

```
~/src/openssl-1.1.1$ make apps/openssl
/usr/bin/perl apps/progs.pl apps/openssl > apps/progs.h
gcc  -I. -Iinclude -Iapps -pthread -m64 -Wa,--noexecstack -Wall -O0 -g -DDEBUG_UNUSED -DPEDANTIC -pedantic -Wno-long-long -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -Wswitch -Wsign-compare -Wshadow -Wformat -Wtype-limits -Wundef -Werror -Wmissing-prototypes -Wstrict-prototypes   -MMD -MF apps/asn1pars.d.tmp -MT apps/asn1pars.o -c -o apps/asn1pars.o apps/asn1pars.c
In file included from apps/apps.h:13,
                 from apps/asn1pars.c:13:
./e_os.h:14:11: fatal error: openssl/opensslconf.h: No such file or directory
 # include <openssl/opensslconf.h>
           ^~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:6380: apps/asn1pars.o] Error 1
```

### A missing dependency

In both cases, the compilation fails because the generated header files are missing. During normal builds, the files are generated in time, but not if one builds `apps/openssl` directly. While this is not an issue for regular users, it is inconvenient for developers.

The problem can be fixed by making the app's modules depend on the `$(GENERATED_MANDATORY)` files. The `$(GENERATED_MANDATORY)` list is generated from the empty DEPEND entries and actually the comment indicates that those dependencies should be built before anything else.

**master**

https://github.com/openssl/openssl/blob/5e5bc836fbc5b1c0af428864f5286bbb225f7baf/build.info#L10-L14

**1.1.1**

https://github.com/openssl/openssl/blob/e512efe0894481679a5d3c57d10bf4ea97046c2a/build.info#L16-L19

This works for `make`, but not for `make apps/openssl`. I was able to fix the build by adding the dependencies manually to the Makefile, but I failed to do it properly in the OpenSSL build system. But for @levitte it should be an easy fix, right?

On a related note it seems inconvenient that the `GENERATED_MANDATORY` list has proliferated a lot recently. Originally, in 1.1.1 it used to contain only the generated header files

```
GENERATED_MANDATORY=include/crypto/bn_conf.h include/crypto/dso_conf.h include/openssl/opensslconf.h
```

but in master a lot of generated pod files (and one cnf file) have been added

```
GENERATED_MANDATORY=doc/man1/openssl-asn1parse.pod doc/man1/openssl-ca.pod \
                    doc/man1/openssl-ciphers.pod doc/man1/openssl-cmds.pod \
                    doc/man1/openssl-cmp.pod doc/man1/openssl-cms.pod \
                    doc/man1/openssl-crl.pod doc/man1/openssl-crl2pkcs7.pod \
                    doc/man1/openssl-dgst.pod doc/man1/openssl-dhparam.pod \
                    doc/man1/openssl-dsa.pod doc/man1/openssl-dsaparam.pod \
                    doc/man1/openssl-ec.pod doc/man1/openssl-ecparam.pod \
                    doc/man1/openssl-enc.pod doc/man1/openssl-engine.pod \
                    doc/man1/openssl-errstr.pod \
                    doc/man1/openssl-fipsinstall.pod \
                    doc/man1/openssl-gendsa.pod doc/man1/openssl-genpkey.pod \
                    doc/man1/openssl-genrsa.pod doc/man1/openssl-info.pod \
                    doc/man1/openssl-kdf.pod doc/man1/openssl-list.pod \
                    doc/man1/openssl-mac.pod doc/man1/openssl-nseq.pod \
                    doc/man1/openssl-ocsp.pod doc/man1/openssl-passwd.pod \
                    doc/man1/openssl-pkcs12.pod doc/man1/openssl-pkcs7.pod \
                    doc/man1/openssl-pkcs8.pod doc/man1/openssl-pkey.pod \
                    doc/man1/openssl-pkeyparam.pod \
                    doc/man1/openssl-pkeyutl.pod doc/man1/openssl-prime.pod \
                    doc/man1/openssl-provider.pod doc/man1/openssl-rand.pod \
                    doc/man1/openssl-rehash.pod doc/man1/openssl-req.pod \
                    doc/man1/openssl-rsa.pod doc/man1/openssl-rsautl.pod \
                    doc/man1/openssl-s_client.pod \
                    doc/man1/openssl-s_server.pod \
                    doc/man1/openssl-s_time.pod doc/man1/openssl-sess_id.pod \
                    doc/man1/openssl-smime.pod doc/man1/openssl-speed.pod \
                    doc/man1/openssl-spkac.pod doc/man1/openssl-srp.pod \
                    doc/man1/openssl-storeutl.pod doc/man1/openssl-ts.pod \
                    doc/man1/openssl-verify.pod doc/man1/openssl-version.pod \
                    doc/man1/openssl-x509.pod \
                    doc/man7/openssl_user_macros.pod \
                    include/crypto/bn_conf.h include/crypto/dso_conf.h \
                    include/openssl/configuration.h \
                    include/openssl/opensslv.h \
                    test/provider_internal_test.cnf
```

Would it be possible to separate those from the generated headers by adding a new kind of dependency like `DEPEND[doc]` `DEPEND[test]` for those?

https://github.com/openssl/openssl/blob/5e5bc836fbc5b1c0af428864f5286bbb225f7baf/doc/man1/build.info#L2-L56

https://github.com/openssl/openssl/blob/5e5bc836fbc5b1c0af428864f5286bbb225f7baf/test/build.info#L731

mspncp:
This commit adds a `build_the_app` make target, which essentially
builds `apps/openssl`, respecting all dependencies.
This target is mainly intended to be used by developers in order
to speed up compilation time.

Fixes #11932

### DIFF
--- a/Configurations/descrip.mms.tmpl
+++ b/Configurations/descrip.mms.tmpl
@@ -145,8 +145,8 @@ HTMLDOCS3={- join(", ", map { "-\n\t".$_ } @{$unified_info{htmldocs}->{man3}}) -
 HTMLDOCS5={- join(", ", map { "-\n\t".$_ } @{$unified_info{htmldocs}->{man5}}) -}
 HTMLDOCS7={- join(", ", map { "-\n\t".$_ } @{$unified_info{htmldocs}->{man7}}) -}
 
-APPS_OPENSSL="{- use File::Spec::Functions;
-                 catfile("apps","openssl") -}"
+APPS_OPENSSL={- use File::Spec::Functions;
+                 catfile("apps","openssl") -}
 
 # DESTDIR is for package builders so that they can configure for, say,
 # SYS$COMMON:[OPENSSL] and yet have everything installed in STAGING:[USER].
@@ -413,11 +413,13 @@ NODEBUG=@
 {- dependmagic('build_libs'); -} : build_libs_nodep
 {- dependmagic('build_modules'); -} : build_modules_nodep
 {- dependmagic('build_programs'); -} : build_programs_nodep
+{- dependmagic('build_the_app'); -} : build_the_app_nodep
 
 build_generated : $(GENERATED_MANDATORY)
 build_libs_nodep : $(LIBS), $(SHLIBS)
 build_modules_nodep : $(MODULES)
 build_programs_nodep : $(PROGRAMS), $(SCRIPTS)
+build_the_app_nodep : $(APPS_OPENSSL), $(SCRIPTS)
 
 build_docs: build_html_docs
 build_html_docs: $(HTMLDOCS1) $(HTMLDOCS3) $(HTMLDOCS5) $(HTMLDOCS7)
--- a/Configurations/unix-Makefile.tmpl
+++ b/Configurations/unix-Makefile.tmpl
@@ -203,8 +203,8 @@ MANDOCS7={-
              fill_lines(" ", $COLUMNS - 9, map { platform->bin($_) }
                         @{$unified_info{mandocs}->{man7}})) -}
 
-APPS_OPENSSL="{- use File::Spec::Functions;
-                 catfile("apps","openssl") -}"
+APPS_OPENSSL={- use File::Spec::Functions;
+                 catfile("apps","openssl") -}
 
 # DESTDIR is for package builders so that they can configure for, say,
 # /usr/ and yet have everything installed to /tmp/somedir/usr/.
@@ -445,6 +445,7 @@ LANG=C
 {- dependmagic('build_libs'); -}: build_libs_nodep
 {- dependmagic('build_modules'); -}: build_modules_nodep
 {- dependmagic('build_programs'); -}: build_programs_nodep
+{- dependmagic('build_the_app'); -}: build_the_app_nodep
 
 build_docs: build_man_docs build_html_docs
 build_man_docs: $(MANDOCS1) $(MANDOCS3) $(MANDOCS5) $(MANDOCS7)
@@ -454,6 +455,7 @@ build_generated: $(GENERATED_MANDATORY)
 build_libs_nodep: libcrypto.pc libssl.pc openssl.pc
 build_modules_nodep: $(MODULES)
 build_programs_nodep: $(PROGRAMS) $(SCRIPTS)
+build_the_app_nodep: $(APPS_OPENSSL) $(SCRIPTS)
 
 # Kept around for backward compatibility
 build_apps build_tests: build_programs
--- a/Configurations/windows-makefile.tmpl
+++ b/Configurations/windows-makefile.tmpl
@@ -145,8 +145,8 @@ HTMLDOCS5_BLDDIRS={- my %dirs = map { dirname($_) => 1 } @HTMLDOCS5;
 HTMLDOCS7_BLDDIRS={- my %dirs = map { dirname($_) => 1 } @HTMLDOCS7;
                      join(' ', sort keys %dirs) -}
 
-APPS_OPENSSL="{- use File::Spec::Functions;
-                 catfile("apps","openssl") -}"
+APPS_OPENSSL={- use File::Spec::Functions;
+                 catfile("apps","openssl") -}
 
 # Do not edit these manually. Use Configure with --prefix or --openssldir
 # to change this!  Short explanation in the top comment in Configure
@@ -376,6 +376,7 @@ PROCESSOR= {- $config{processor} -}
 {- dependmagic('build_libs'); -}: build_libs_nodep
 {- dependmagic('build_modules'); -}: build_modules_nodep
 {- dependmagic('build_programs'); -}: build_programs_nodep
+{- dependmagic('build_the_app'); -}: build_the_app_nodep
 
 build_docs: build_html_docs
 build_html_docs: $(HTMLDOCS1) $(HTMLDOCS3) $(HTMLDOCS5) $(HTMLDOCS7)
@@ -384,6 +385,7 @@ build_generated: $(GENERATED_MANDATORY)
 build_libs_nodep: $(LIBS) {- join(" ",map { platform->sharedlib_import($_) // () } @{$unified_info{libraries}}) -}
 build_modules_nodep: $(MODULES)
 build_programs_nodep: $(PROGRAMS) $(SCRIPTS)
+build_the_app_nodep: $(APPS_OPENSSL) $(SCRIPTS)
 
 # Kept around for backward compatibility
 build_apps build_tests: build_programs
--- a/apps/build.info
+++ b/apps/build.info
@@ -62,8 +62,8 @@ IF[{- !$disabled{apps} -}]
   DEPEND[openssl]=libapps.a ../libssl
 
   DEPEND[${OPENSSLSRC/.c/.o}]=progs.h
-  GENERATE[progs.c]=progs.pl -C $(APPS_OPENSSL)
-  GENERATE[progs.h]=progs.pl -H $(APPS_OPENSSL)
+  GENERATE[progs.c]=progs.pl -C "$(APPS_OPENSSL)"
+  GENERATE[progs.h]=progs.pl -H "$(APPS_OPENSSL)"
   # progs.pl tries to read all 'openssl' sources, including progs.c, so we make
   # sure things are generated in the correct order.
   DEPEND[progs.h]=progs.c


mspncp:
**Note**

Originally, I wanted to call the target just `build_app`, until I noticed that there is a legacy target called `build_apps` (see commit commit 1e3d16b0a62a9523b44cfcd65abc5998949ed9ab). After some thought, I called it `build_the_app`. If you don't like the name, I'm open for suggestions  (`build_tool`, `build_osslapp`, `build_apps_openssl`, ...).

levitte:
I don't like how this is done.  It adds "magic knowledge" to the build file templates, where such knowledge should *diminish*.

If you do want something like this, it should be governed by information from the build.info files.  You can take the way I separate different classes of scripts as an example.  So for example, I would do this:

``` diff
diff --git a/apps/build.info b/apps/build.info
index 8bfcec65d0..d73fc84855 100644
--- a/apps/build.info
+++ b/apps/build.info
@@ -56,7 +56,7 @@ IF[{- !$disabled{'cmp'} -}]
 ENDIF
 
 IF[{- !$disabled{apps} -}]
-  PROGRAMS=openssl
+  PROGRAMS{apps}=openssl
   SOURCE[openssl]=$INITSRC $OPENSSLSRC
   INCLUDE[openssl]=.. ../include include
   DEPEND[openssl]=libapps.a ../libssl
```

And similarly, I would use `PROGRAMS{tests}` in `test/build.info`.

When that's done, you could add two variables in the build files APPS and TESTS, which collect the two program variants, then change around `build_apps`, `build_tests` and `build_programs` like this:

``` diff
diff --git a/Configurations/unix-Makefile.tmpl b/Configurations/unix-Makefile.tmpl
index 126180ace7..100f1991f8 100644
--- a/Configurations/unix-Makefile.tmpl
+++ b/Configurations/unix-Makefile.tmpl
@@ -444,7 +444,9 @@ LANG=C
 {- dependmagic('build_sw'); -}: build_libs_nodep build_modules_nodep build_programs_nodep link-utils
 {- dependmagic('build_libs'); -}: build_libs_nodep
 {- dependmagic('build_modules'); -}: build_modules_nodep
-{- dependmagic('build_programs'); -}: build_programs_nodep
+{- dependmagic('build_apps'); -}: build_apps_nodep
+{- dependmagic('build_tests'); -}: build_tests_nodep
+{- dependmagic('build_programs'); -}: build_apps_nodep build_tests_nodep
 
 build_docs: build_man_docs build_html_docs
 build_man_docs: $(MANDOCS1) $(MANDOCS3) $(MANDOCS5) $(MANDOCS7)
@@ -453,10 +455,6 @@ build_html_docs: $(HTMLDOCS1) $(HTMLDOCS3) $(HTMLDOCS5) $(HTMLDOCS7)
 build_generated: $(GENERATED_MANDATORY)
 build_libs_nodep: libcrypto.pc libssl.pc openssl.pc
 build_modules_nodep: $(MODULES)
-build_programs_nodep: $(PROGRAMS) $(SCRIPTS)
-
-# Kept around for backward compatibility
-build_apps build_tests: build_programs
 
 # Convenience target to prebuild all generated files, not just the mandatory
 # ones
```

(and similarly in the others)

mspncp:
Ah ok, thanks. I'll try that.

richsalz:
Why add a new target, and not just update the dependencies?

mspncp:
> Why add a new target, and not just update the dependencies?

I did try several variants before discovering the 'dependmagic' targets. The most successful one was the following. It succeeded when I `make apps/openssl` but fails with `make -j8 apps/openssl`:

```diff
commit 19e1713f990bfe9a780f8376895647c79558b763
Author: Dr. Matthias St. Pierre <matthias.st.pierre@ncp-e.com>
Date:   Sun May 24 15:52:27 2020 +0200

    Build: make apps/progs.c depend on generated header files
    
    Fixes #11932

diff --git a/apps/build.info b/apps/build.info
index 8bfcec65d0..9f898ab3bb 100644
--- a/apps/build.info
+++ b/apps/build.info
@@ -68,8 +68,9 @@ IF[{- !$disabled{apps} -}]
   # sure things are generated in the correct order.
   DEPEND[progs.h]=progs.c
   # Because the files to look through may change (depends on $OPENSSLSRC),
-  # always depend on a changed configuration.
-  DEPEND[progs.c]=../configdata.pm
+  # always depend on a changed configuration.  Also, ensure that the
+  # generated header files are present when building.
+  DEPEND[progs.c]=../configdata.pm ../build_generated
 
   IF[{- $config{target} =~ /^(?:Cygwin|mingw|VC-)/ -}]
     GENERATE[openssl.rc]=../util/mkrc.pl openssl
```

richsalz:
Maybe it's harder than just adding a new target, but it seems like getting the dependency right is the correct thing to do.  Shrug.

mspncp:
The main targets all invoke two recursive `make` calls: one to setup the dependencies and one to build them. I don't see how you would be able to achieve this in a single `make` instance.

```
build_sw: build_generated
	$(MAKE) depend && $(MAKE) _build_sw
```

(With gnu make, you might be able to do it, but what about nmake for example?)

richsalz:
I'm not going to try to solve this, but ... why doesn't apps/openssl explicitly depend on the libraries, for example?

mspncp:
Good question.

levitte:
> I'm not going to try to solve this, but ... why doesn't apps/openssl explicitly depend on the libraries, for example?

Come again?

``` console
: ; grep -F 'DEPEND[openssl]' apps/build.info 
  DEPEND[openssl]=libapps.a ../libssl
: ; grep -F 'DEPEND[libssl]' build.info 
DEPEND[libssl]=libcrypto
: ; grep -A30 'apps/openssl:' Makefile
apps/openssl: apps/openssl-bin-asn1pars.o apps/openssl-bin-ca.o \
              apps/openssl-bin-ciphers.o apps/openssl-bin-cmp.o \
              apps/openssl-bin-cmp_mock_srv.o apps/openssl-bin-cms.o \
              apps/openssl-bin-crl.o apps/openssl-bin-crl2p7.o \
              apps/openssl-bin-dgst.o apps/openssl-bin-dhparam.o \
              apps/openssl-bin-dsa.o apps/openssl-bin-dsaparam.o \
              apps/openssl-bin-ec.o apps/openssl-bin-ecparam.o \
              apps/openssl-bin-enc.o apps/openssl-bin-engine.o \
              apps/openssl-bin-errstr.o apps/openssl-bin-fipsinstall.o \
              apps/openssl-bin-gendsa.o apps/openssl-bin-genpkey.o \
              apps/openssl-bin-genrsa.o apps/openssl-bin-info.o \
              apps/openssl-bin-kdf.o apps/openssl-bin-list.o \
              apps/openssl-bin-mac.o apps/openssl-bin-nseq.o \
              apps/openssl-bin-ocsp.o apps/openssl-bin-openssl.o \
              apps/openssl-bin-passwd.o apps/openssl-bin-pkcs12.o \
              apps/openssl-bin-pkcs7.o apps/openssl-bin-pkcs8.o \
              apps/openssl-bin-pkey.o apps/openssl-bin-pkeyparam.o \
              apps/openssl-bin-pkeyutl.o apps/openssl-bin-prime.o \
              apps/openssl-bin-progs.o apps/openssl-bin-provider.o \
              apps/openssl-bin-rand.o apps/openssl-bin-rehash.o \
              apps/openssl-bin-req.o apps/openssl-bin-rsa.o \
              apps/openssl-bin-rsautl.o apps/openssl-bin-s_client.o \
              apps/openssl-bin-s_server.o apps/openssl-bin-s_time.o \
              apps/openssl-bin-sess_id.o apps/openssl-bin-smime.o \
              apps/openssl-bin-speed.o apps/openssl-bin-spkac.o \
              apps/openssl-bin-srp.o apps/openssl-bin-storeutl.o \
              apps/openssl-bin-ts.o apps/openssl-bin-verify.o \
              apps/openssl-bin-version.o apps/openssl-bin-x509.o \
              apps/libapps.a libssl.so libcrypto.so
	rm -f apps/openssl
	$${LDCMD:-$(CC)} $(BIN_CFLAGS) -L. $(BIN_LDFLAGS) \
```

richsalz:
Well, if the dependencies were correct then it would build.  So something is missing.

levitte:
Well, the problem is manifold.

The most central technique we use is automatic dependency generation, i.e. the output generated by options such as gcc's -MM + -MF.  However, since they are generated during the build, those dependencies will simply not be present the first time around.  For a pristine source checkout where you build everything, that's not a problem, things will be built anyway, and as long as files aren't move around and `#include`s aren't changed, neither is the second build, is it will pick up the (correct) dependencies.  The issue only really occurs when files are moved around, or some `#include`s are changed, because it means that the current automatically generated dependencies will be out of sync, for one build...  that one build will generated new dependencies, so things should be in sync after that.

As for the need for `build_generated`, *which is a different issue*, we *could* have explicit dependencies where approriate.  Let's take, say, `include/crypto/dso_conf.h`...  do you know what object files depend on that one, directly or indirectly?  Are you willing to find out?  How should we deal with those in a more general sense without driving ourselves crazy?
(*one* way to deal with that could actually be to generate a build.info that finds everything that depends specifically on the generated header files...  this would be something for `make update`.  That does mean using makedepend all over again, but it would be a viable option)

mspncp:
@levitte I would like to take the opportunity for a clarification and big compliment: we're optimizing nits here. The unified build scheme which you created for 1.1.0+ is lightyears better than what we used to have back in 1.0.2. (I can say that, because we're still using it.)

levitte:
Sure.  It seems, though, that @richsalz needs an explanation or a reminder, and I'm happy to serve.

richsalz:
I never meant to imply that this isn't miles better than what it replaced.

If you tell me that apps/openssl cannot have explicit dependencies so that things "just work" well, okay.  I will continue to push a bit.  For example, the build_generated phase could create a build.inc that goes lists apps/*.o dependencies on generated files.

But if I'm in the minority here, so be it.

levitte:
> For example, the build_generated phase could create a build.inc that goes lists apps/*.o dependencies on generated files.

What will that accomplish, and how?  And why should the application be treated specially?  What if someone wants to `make libcrypto.so`?

> But if I'm in the minority here, so be it.

Most of all, you're throwing in a seemingly half-baked idea, or if it's not half-baked, you don't fully share the idea.

PR #11948, opened 2020-05-25
Test Rsa oaep in fips mode

slontis:
##### Checklist


slontis:
Dependent on #11927 for memory leak fix.

Also depends on #11884 which documents the openssl application changes.

test/recipes/15-test_rsaoaep.t &
apps/pkeyutil.c are the only files that should be in this PR.

test/recipes/15-test_rsaoaep.t
```diff
@@ -0,0 +1,156 @@
...
+
+use strict;
+use warnings;
+
+use OpenSSL::Test qw(:DEFAULT data_file bldtop_dir srctop_file srctop_dir bldtop_file);
+use OpenSSL::Test::Utils;
+use File::Compare qw/compare_text/;
+
+BEGIN {
+setup("test_rsaoaep");
```
levitte, line 18: Nit: indent

test/recipes/15-test_rsaoaep.t
```diff
@@ -0,0 +1,156 @@
...
+   "RSA OAEP Encryption should generate different encrypted data");
+
+ok(run(app(['openssl', 'pkeyutl',
+            @prov,
+            '-decrypt',
+            '-inkey', $key_file,
+            '-pkeyopt', 'pad-mode:oaep',
+            '-pkeyopt', 'oaep-label:123',
+            '-in', $enc_file,
+            '-out', $dec_file]))
```
levitte, line 96: Please use different output files for every run.  That allows looking at them after the test is run too re-play or find out what went wrong when a test fails (the output files of this test are found in `test-runs/test_rssaoaep`)

apps/include/opt.h
```diff
@@ -271,17 +271,20 @@
...
         OPT_PROV_PROVIDER, OPT_PROV_PROVIDER_PATH, \
+        OPT_PROV_PROVIDER_CONFIG_FILE, \
         OPT_PROV__LAST
 
 # define OPT_PROV_OPTIONS \
         OPT_SECTION("Provider"), \
         { "provider_path", OPT_PROV_PROVIDER_PATH, 's', "Provider load path (must be before 'provider' argument if required)" }, \
-        { "provider", OPT_PROV_PROVIDER, 's', "Provider to load (can be specified multiple times)" }
+        { "provider", OPT_PROV_PROVIDER, 's', "Provider to load (can be specified multiple times)" }, \
+        { "provider_config", OPT_PROV_PROVIDER_CONFIG_FILE, 's', "Provider config file to load (can be specified multiple times)." }
```
levitte, line 281: The parenthesis should be dropped, that's not how you have implemented it.

apps/lib/app_provider.c
```diff
@@ -44,6 +46,8 @@ static int opt_provider_load(const char *provider)
 static void provider_free(OSSL_PROVIDER *prov)
 {
     OSSL_PROVIDER_unload(prov);
+    OPENSSL_CTX_free(libctx);
+    libctx = NULL;
```
levitte, line 50: These two lines should be in `app_providers_cleanup()`

slontis: removed this file - its in another bigger PR.

apps/include/opt.h
slontis: Removed this file also..  Multiple configs could be loaded - so the parenthesis are valid.

test/recipes/15-test_rsaoaep.t
slontis: Done - shuffled the tests around also so they are all grouped together for the file that is used.

apps/pkeyutl.c
```diff
@@ -244,6 +247,7 @@ int pkeyutl_main(int argc, char **argv)
     argc = opt_num_rest();
     if (argc != 0)
         goto opthelp;
+    libctx = app_libctx();
```
levitte, line 250: Where is `app_libctx()` defined, let alone declared?

slontis: My initial comment described the dependancy on #11884.

slontis: Ok I have added a revised libctx back in, to try to progress this forward.

apps/lib/app_provider.c
```diff
@@ -41,6 +44,43 @@ static int opt_provider_load(const char *provider)
...
+    if (libctx == NULL)
+        return 1;
+
+    if (!OSSL_PROVIDER_do_all(libctx, provider_count, data))
+        return 0;
+
+    if (count == 0) {
+        if (!opt_provider_load("default"))
+            return 0;
+    }
```
levitte, line 80: Why do you do this, for something that should work through fallbacks?

apps/lib/app_provider.c
```diff
@@ -41,6 +44,43 @@ static int opt_provider_load(const char *provider)
     return 1;
 }
 
+/*
+ * Add this call to the start of any application that support a library context.
+ * NOTE: app_load_default_provider() should also be called after the options
+ * have been processed.
+ */
+OPENSSL_CTX *app_new_libctx(void)
```
levitte, line 52: This function *could* be called once and for all in `apps/opensl.c`'s `main()`...

apps/lib/app_provider.c
```diff
@@ -50,13 +90,28 @@ void app_providers_cleanup(void)
 {
     sk_OSSL_PROVIDER_pop_free(app_providers, provider_free);
     app_providers = NULL;
+    OPENSSL_CTX_free(libctx);
+    libctx = NULL;
```
levitte, line 94: If you call `app_new_libctx()` from `main()`, these two lines should move there as well, of course...

apps/lib/app_provider.c
slontis: If every app uses a non default libctx then that it true.. otherwise it is not.

apps/lib/app_provider.c
slontis: Err assuming that the NULL context has been loaded with the NULL provider I thought this will not work.

apps/lib/app_provider.c
slontis: see above

apps/lib/app_provider.c
levitte: This code won't work either, then, will it?  The "null" provider counts as *one*, doesn't it?  So what I see you doing here is to load the "default" provider when no other providers (not even the "null" provider) have been loaded, which is exactly what the fallback mechanism does, no?

levitte: Oh, hold on, `opt_provider_load` only touches the default library context.  Quite frankly, though, I have a hard time understanding what you want to accomplish with this...  after all, it's not like doing anything with a non-default library context will magically fall back on whatever's in the default provider, will it?  At least conceptually...

slontis: Removed this code - you are right the fallback would handle it.

apps/lib/app_provider.c
slontis: Fixed.

slontis:
Updated docs.

levitte:
On a per-command basis, I'm not sure I like the whole library context fiddling, at all.  It seems...  unnecessary.

Note, though, that I don't think that a library context per se is unnecessary, but it should be made on a global level.  Think of the interactive mode, where we currently have each command affecting the rest regarding libcrypto global data.  If each command was executed within its own library context, the problem of one command affecting the next should ultimately go away.

Remains electively loading the null provider into the default provider...  I assume that's an interim thing, to purposefully make things break if they rely on the default library context when they shouldn't.  I'm not sure I'd make that subject of a per-command option, but rather of an environment variable (mostly because we don't have global options).  `OPENSSL_NO_DEFAULT_CONTEXT` maybe?

I've had a global app library context in the back of my mind for a while, so I have ideas (but for once, not a branch yet).  Do you want to tackle that or shall I?

slontis:
I would like to see less environment variables, not more of them :)
Putting the null provider into the default context and using the fips provider in a created library context is a fundamental requirement. The library context was introduced to solve a particular problem. Using the fips provider in the default library context is avoiding the issue that the library context was added for.

slontis:
As I just discovered with pkeyutl a library context is going to break certain operations that are not currently compatible (such as SM2). Which is why I gone with the per command -use_libctx option.

levitte:
> I would like to see less environment variables, not more of them :)

Yes, me too, but I would like *even less* to see interim / unnecessary options

> Putting the null provider into the default context and using the fips provider in a created library context is a fundamental requirement.

Er, what???  Why?

> The library context was introduced to solve a particular problem. Using the fips provider in the default library context is avoiding the issue that the library context was added for.

Yes, almost everything that's currently global in libcrypto should move to be stored in a library context, and we're currently not there.  That's the reason I'm calling all these new options "interim", because as soon as we've finished that move, they will not be necessary any more.  That's also why I seriously question the "fundamental requirement" that the FIPS provider must be loaded in a non-default provider...  it sounds much more like a band-aid than a "fundamental requirement".

apps/include/opt.h
```diff
@@ -270,17 +270,22 @@
...
+        OPT_PROV_PROVIDER, OPT_PROV_PROVIDER_PATH, OPT_PROV_LOAD_CONFIG, \
         OPT_PROV__LAST
 
+# define OPT_LIBCTX_OPTIONS \
+        { "use_libctx", OPT_LIBCTX, '-', "Create a library context to load providers into (use before provider options)" }
+
 # define OPT_PROV_OPTIONS \
         OPT_SECTION("Provider"), \
         { "provider_path", OPT_PROV_PROVIDER_PATH, 's', "Provider load path (must be before 'provider' argument if required)" }, \
+        { "provider_config", OPT_PROV_LOAD_CONFIG, 's', "Load provider using a configuration file" }, \
```
levitte, line 282: We already have `-config`, why add another option that essentially does the same thing?

slontis: I can have a look..

-config is not global either.   

And then it does this block copy in the apps that use it..
 if ((conf = load_config_file(configfile)) == NULL)
        goto end;
    if (configfile != default_config_file && !app_load_modules(conf))
        goto end;

apps/lib/app_provider.c
```diff
@@ -41,6 +46,24 @@ static int opt_provider_load(const char *provider)
     return 1;
 }
 
+OPENSSL_CTX *opt_provider_libctx(void)
+{
+    /*
+     * Load the NULL provider into the default library context and create a
+     * library context which will then be used for any OPT_PROV options.
+     */
+    if (libctx == NULL) {
```
paulidale, line 55: Should there be an error if `libctx != NULL`?
I tend to think so.

slontis:
> I assume that's an interim thing, to purposefully make things break if they rely on the default library context when they shouldn't.

I am concerned if we dont always test it this way-  then it will get broken every time new code is added that doesnt handle the library context. (which would be easy to miss).

doc/man1/openssl.pod
```diff
@@ -780,6 +780,18 @@ multiple time to load more than one provider.
...
+
+Load a provider using a configuration I<filename>.  This option can be used by
+any provider that requires additional configuration data in order to load.
+
+=item B<-use_libctx>
+
+Create the library context that providers can be loaded into, and load the
+"null" provider into the default (NULL) library context.
+This option should appear before any other provider options.
+Most openssl applications do not currently support this option.
```
paulidale, line 793: What about a note like: _This option allows detection algorithm use where the libctx is omitted._

test/recipes/15-test_rsaoaep_data/plain_text
```diff
@@ -0,0 +1 @@
+Hello world
```
paulidale, line 1: Deliberately not including the final new line or not?

paulidale:
The testing to catch errors is compelling IMO.

I also think that loading the FIPS provider into a non-default context is a requirement.  Imagine an application with supports FIPS and not simultaneously using a library context and the default context.  Insisting that FIPS is in the default context would work but it kind of obtuse.

Also consider the application linked with two different libraries (say A & B) each using OpenSSL -- how does a library A guarantee it is FIPS and that the default context isn't contaminated by B?

apps/include/opt.h
levitte: > -config is not global either.

Well, no it isn't.  We don't support global options, that's well known.

> And then it does this block copy in the apps that use it..
> if ((conf = load_config_file(configfile)) == NULL)
> goto end;
> if (configfile != default_config_file && !app_load_modules(conf))
> goto end;

Yes?  The app library isn't holy, it may be changed.  However, I wonder if `OPENSSL_CTX_load_config()` shouldn't be made to take a `CONF *` instead of a file name, and if there shouldn't be a `CONF_modules_load_with_libctx()` that it could use.  That would be much more on par with already existing functionality.

levitte:
> Insisting that FIPS is in the default context would work but it kind of obtuse.

Insisting that it must be loaded in a non-default context is just as obtuse.

paulidale:
> Insisting that it must be loaded in a non-default context is just as obtuse.

Which nobody is suggesting.  Both must be available.

levitte:
> Also consider the application linked with two different libraries (say A & B) each using OpenSSL -- how does a library A guarantee it is FIPS and that the default context isn't contaminated by B?

It doesn't, so an application that does this would be *wise* to create library contexts for them.  However, this is also one specific scenario, and I agree that with that scenario, the application should do so.  *our* application isn't one of those, as far as I know.

levitte:
> > Insisting that it must be loaded in a non-default context is just as obtuse.
> 
> Which nobody is suggesting.  Both must be available.

In that case, we understand "fundamental requirement" differently.  For me, those are strong words, basically saying "everyone *must* do this, there's no alternative".

paulidale:
Ummm, how does an application control the library contexts in third party libraries?  This was the primary driver behind library contexts.

Sure, in time we'd hope that each third party library would create its own library context.

levitte:
> Ummm, how does an application control the library contexts in third party libraries? This was the primary driver behind library contexts.

Uhmmm, it doen't.  The idea was that those libraries would create their own library contexts, to insulate themselves from the application.

levitte:
> how does an application control the library contexts in third party libraries?

I get seriously worried by this question.  What third party libraries do is out of our control, let's not try and pretend otherwise.

paulidale:
We need to provide the capability for third parties to do the right thing.  We can't force them to of course.

levitte:
> We need to provide the capability for third parties to do the right thing. We can't force them to of course.

I haven't said anything else.  Of course we should.  What I'm currently concerned with is adding on options to the *application **(!)*** that are quite obviously going to be interim (like I said, the interactive mode will need to create a new library context for each command).

richsalz:
Just to make sure that this point doesn't get lost.  We want the default context to be "normal" OpenSSL, with all the non-FIPS algorithms. For some connections, we want to use the FIPS context and we are going to exlicitly specify the context for those.

I am sure there are applications who are doing the exact opposite.

Both are still possible, right?

levitte:
> Both are still possible, right?

Yes. The dispute is mainly about the app's ui at this point.

apps/lib/app_provider.c
slontis: Moved it into app.. This is allowed to be NULL.

apps/include/opt.h
slontis: Ok rewritten..

doc/man1/openssl.pod
slontis: What about no - this makes no sense :)

test/recipes/15-test_rsaoaep_data/plain_text
slontis: Added

slontis:
@levitte. I have updated the code. 
I need the 'use_libctx' option to test this as well as any other fips tests (see CMS). So even though it is 'just for testing'. It needs to always be there in order to do the tests. So this is not really a temporary config option.

I have reused -config instead of using provider_config (but this also now loads any modules).

slontis:
The size of PR #11884 is getting quite large - so I have tried to separate out some bits into smaller PR's to make it easier to review.

slontis:
ping @levitte

slontis:
ping @levitte

slontis:
ping @levitte

slontis:
ping @levitte

apps/pkeyutl.c
```diff
@@ -547,13 +567,19 @@ static EVP_PKEY_CTX *init_ctx(const char *kdfalg, int *pkeysize,
...
     } else {
         if (pkey == NULL)
             goto end;
 
         *pkeysize = EVP_PKEY_size(pkey);
-        ctx = EVP_PKEY_CTX_new(pkey, impl);
+        if (libctx != NULL)
+            ctx = EVP_PKEY_CTX_new_from_pkey(libctx, pkey, propq);
+        else
+            ctx = EVP_PKEY_CTX_new(pkey, impl);
```
t8m, line 582: Same as above.

doc/man1/openssl.pod
levitte: I can still not make sense of this option.  Yeah, I get that *for us*, it has some interest for testing purposes, but for the average user?

apps/pkeyutl.c
```diff
@@ -547,13 +567,19 @@ static EVP_PKEY_CTX *init_ctx(const char *kdfalg, int *pkeysize,
                 goto end;
             }
         }
-        ctx = EVP_PKEY_CTX_new_id(kdfnid, impl);
+        if (libctx != NULL)
+            ctx = EVP_PKEY_CTX_new_from_name(libctx, kdfalg, propq);
+        else
+            ctx = EVP_PKEY_CTX_new_id(kdfnid, impl);
```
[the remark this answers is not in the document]
slontis: It is fine for the libctx to be NULL..
That then just uses the default context..
The same is true for the engine pointer also.

It could at best do
```
if (impl == NULL)
   new_from_pkey()
else
  new( , impl)
```

slontis:
Changed apps to use an environment variable 'OPENSSL_TEST_LIBCTX'
instead of the -use_libctx option.

@levitte or @mattcaswell  - can someone review please.

slontis:
Needed to rebase and remove the default options for tests call to fipsinstall.

richsalz:
you could have left the redundant default flags, as a test.  When I added the defaults, I deliberately did that in one test.

slontis:
Either way.. having hexkey:00 was not good though :)

slontis:
Wow the tests ran without a timeout...

slontis:
Thanks.. Merged to master.

PR #11953, opened 2020-05-25
Rename EVP_PKEY_cmp() to EVP_PKEY_eq() and EVP_PKEY_cmp_parameters()  to EVP_PKEY_parameters_eq()

DDvO:
As discussed in #11870

### DIFF
--- a/crypto/evp/exchange.c
+++ b/crypto/evp/exchange.c
@@ -368,13 +368,13 @@ int EVP_PKEY_derive_set_peer(EVP_PKEY_CTX *ctx, EVP_PKEY *peer)
 
     /*
      * For clarity.  The error is if parameters in peer are
-     * present (!missing) but don't match.  EVP_PKEY_cmp_parameters may return
+     * present (!missing) but don't match.  EVP_PKEY_parameters_eq may return
      * 1 (match), 0 (don't match) and -2 (comparison is not defined).  -1
      * (different key types) is impossible here because it is checked earlier.
      * -2 is OK for us here, as well as 1, so we can check for 0 only.
      */
     if (!EVP_PKEY_missing_parameters(peer) &&
-        !EVP_PKEY_cmp_parameters(ctx->pkey, peer)) {
+        !EVP_PKEY_parameters_eq(ctx->pkey, peer)) {
         EVPerr(EVP_F_EVP_PKEY_DERIVE_SET_PEER, EVP_R_DIFFERENT_PARAMETERS);
         return -1;
     }
--- a/crypto/evp/keymgmt_lib.c
+++ b/crypto/evp/keymgmt_lib.c
@@ -236,8 +236,8 @@ int evp_keymgmt_util_has(EVP_PKEY *pk, int selection)
  * but also in the operation cache to see if there's any common keymgmt that
  * supplies OP_keymgmt_match.
  *
- * evp_keymgmt_util_match() adheres to the return values that EVP_PKEY_cmp()
- * and EVP_PKEY_cmp_parameters() return, i.e.:
+ * evp_keymgmt_util_match() adheres to the return values that EVP_PKEY_eq()
+ * and EVP_PKEY_parameters_eq() return, i.e.:
  *
  *  1   same key
  *  0   not same key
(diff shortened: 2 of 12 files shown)


DDvO:
All right, done.

romen:
Should these changes be reflected also on the provider interface?

levitte:
> Should these changes be reflected also on the provider interface?

The name in that interface is `OP_keymgmt_match()`.  Semantically speaking, I don't see that it needs to be changed.

vt-alt:
Is it correct just to replace `EVP_PKEY_cmp` with `EVP_PKEY_eq`? I thought it's decided they should return different values.

```
-        ret = EVP_PKEY_cmp(xk, k);
+        ret = EVP_PKEY_eq(xk, k);
```

DDvO:
@vt-alt, no, it would be very hard to change the semantics of `EVP_PKEY_cmp()` to properly do an order-like comparison. Instead we make clear that it basically just checks for equality.
OTOH, the name and return values of `X509_PUBKEY_cmp()`, now `X509_PUBKEY_eq()`, will change due to #11894, and will be consistent with `EVP_PKEY_eq()`.

vt-alt:
I don't say to change `EVP_PKEY_cmp`, but isn't `EVP_PKEY_eq` should return 1 for a match? Now it will return 0, isn't it?

DDvO:
On equal input, `EVP_PKEY_eq()`, which is the same as the former `EVP_PKEY_cmp()`, continues to return 1, and `X509_PUBKEY_eq()`, which is the fixed `X509_PUBKEY_cmp()`, will return also 1 (making direct internal use of `EVP_PKEY_eq/cmp()`, which is now correct, see https://github.com/openssl/openssl/pull/11894/files#diff-a4d3e5a924de45dc8e27b1f12c1907ad).

vt-alt:
@DDvO Ah, yes! Thanks.

t8m:
Of course this will require update after #11894 is merged.

levitte:
Sorry for throwing this in so late, but I'm not sure `eq` is much better than `cmp`.  What does "equality" mean, precisely?  For example, `EVP_PKEY_cmp()` compares key parameters and the public key, does that make two EVP_PKEYs equal?  One could contain the private half while the other doesn't, are they equal?

How about "match" instead of "eq"?

(I'll totally blame @romen for reminding me of the name used in the provider interface :wink:)

DDvO:
> Of course this will require update after #11894 is merged.

Right - done.

DDvO:
> Sorry for throwing this in so late, but I'm not sure `eq` is much better than `cmp`. What does "equality" mean, precisely? For example, `EVP_PKEY_cmp()` compares key parameters and the public key, does that make two EVP_PKEYs equal? One could contain the private half while the other doesn't, are they equal?
> 
> How about "match" instead of "eq"?


I agree that "match" typially has a broader meaning than "eq" (equality). Both should be instances of an equivalence relation (mathematically speaking).

Both "match" and "eq" are certainly better here than "cmp" (compare), which typically involves an ordering relation (such as "<" or ">="), not an equivalence relation.

Also the meaning of equality is generally not as narrow as one might naively expect, not only in mathematics/logics (where AFAICS any equivalence relation would do), but think also of user-defined equality operations that many OO programming languages offer.

In our concrete case here in my view "eq" would be broad enough to cover also the case that a private key is not present while the public keys are the same, and in particiular the difference cannot be much because if the public keys are equal the corresponding private keys (most likely) are equal, too.
Moreover, "eq" is shorter than "match" :wink:

levitte:
I see you're not much acquainted with [lisp](https://stackoverflow.com/questions/547436/whats-the-difference-between-eq-eql-equal-and-equalp-in-common-lisp) :wink:

But yeahok, I can accept that argument

t8m:
You should also rename the EVP_PKEY_cmp call from X509_PUBKEY_eq().

DDvO:
> You should also rename the EVP_PKEY_cmp call from X509_PUBKEY_eq().

Ah, I thought I had already done this, but my `--amend` missed that  `x_pubkey.c` file.
Now it's really pushed.

DDvO:
Merged after sufficient discussion :)

PR #11955, opened 2020-05-25
bio printf: Avoid using rounding errors in range check

bernd-edlinger:
There is a problem casting ULONG_MAX to double which clang-10 is warning about.
ULONG_MAX typically cannot be exactly represented as a double.  ULONG_MAX + 1
can be and this fix uses the latter, however since ULONG_MAX cannot be
represented exactly as a double number we subtract 65535 from this number,
and the result has at most 48 leading one bits, and can therefore be
represented as a double integer without rounding error.  By adding
65536.0 to this number we achive the correct result, which should avoid the
warning.

The addresses a symptom of the underlying problem: we print doubles via an
unsigned long integer.  Doubles have a far greater range and should be printed
better.

This is an alternate fix avoiding dependency to libm and using
inexact floating point numbers with rounding errors.

### DIFF
--- a/crypto/bio/b_print.c
+++ b/crypto/bio/b_print.c
@@ -635,7 +635,7 @@ fmtfp(char **sbuffer,
             fvalue = tmpvalue;
     }
     ufvalue = abs_val(fvalue);
-    if (ufvalue > ULONG_MAX) {
+    if (ufvalue >= (double)(ULONG_MAX - 65535) + 65536.0) {
         /* Number too big */
         return 0;
     }
--- a/crypto/bio/b_print.c
+++ b/crypto/bio/b_print.c
@@ -635,7 +635,13 @@ fmtfp(char **sbuffer,
             fvalue = tmpvalue;
     }
     ufvalue = abs_val(fvalue);
-    if (ufvalue >= (double)(ULONG_MAX - 65535) + 65536.0) {
+    /*
+     * By subtracting 65535 (2^16-1) we canel the low order 15 bits
+     * of ULONG_MAX to avoid using imprecise floating point values.
+     * The second condition is necessary to catch NaN values.
+     */
+    if (ufvalue >= (double)(ULONG_MAX - 65535) + 65536.0
+            || ufvalue != ufvalue /* NaN */) {
         /* Number too big */
         return 0;
     }
--- a/crypto/bio/b_print.c
+++ b/crypto/bio/b_print.c
@@ -636,12 +636,12 @@ fmtfp(char **sbuffer,
     }
     ufvalue = abs_val(fvalue);
     /*
-     * By subtracting 65535 (2^16-1) we canel the low order 15 bits
+     * By subtracting 65535 (2^16-1) we cancel the low order 15 bits
      * of ULONG_MAX to avoid using imprecise floating point values.
      * The second condition is necessary to catch NaN values.
      */
     if (ufvalue >= (double)(ULONG_MAX - 65535) + 65536.0
-            || ufvalue != ufvalue /* NaN */) {
+            || !(ufvalue == ufvalue) /* NaN */) {
         /* Number too big */
         return 0;
     }
--- a/test/bioprinttest.c
+++ b/test/bioprinttest.c
@@ -244,7 +244,7 @@ static int test_fp(int i)
 static int test_big(void)
 {
     char buf[80];
-    double d, z, inf, nan;
+    volatile double d, z, inf, nan;
 
     /* Test excessively big number. Should fail */
     if (!TEST_int_eq(BIO_snprintf(buf, sizeof(buf),
--- a/test/bioprinttest.c
+++ b/test/bioprinttest.c
@@ -241,10 +241,12 @@ static int test_fp(int i)
     return r;
 }
 
+double zero_value = 0.0;
+
 static int test_big(void)
 {
     char buf[80];
-    volatile double d, z, inf, nan;
+    double d, z, inf, nan;
 
     /* Test excessively big number. Should fail */
     if (!TEST_int_eq(BIO_snprintf(buf, sizeof(buf),
@@ -252,7 +254,7 @@ static int test_big(void)
         return 0;
 
     d = 1.0;
-    z = 0.0;
+    z = zero_value;
     inf = d / z;
     nan = z / z;
 
--- a/test/bioprinttest.c
+++ b/test/bioprinttest.c
@@ -244,11 +244,42 @@ static int test_fp(int i)
 static int test_big(void)
 {
     char buf[80];
+    double d, z, inf, nan;
 
     /* Test excessively big number. Should fail */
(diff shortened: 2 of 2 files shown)


paulidale:
Note: I don't think we'll be able to do proper real formatting without either:

1. _knowing_ the format of `double` intimately or
1. using _libm_.

bernd-edlinger:
Yes, indeed, this just fixes a warning.
Are there other real formats that IEEE 754 ?
I would be surprised when any other format ever existed.
But IEEE 754 has exceptional values, +/- INF they also work
and are greater than any finite number.  So theses are okay,
and the check works for them.

BUT the NaN (not - a - number, result of 0/0) has undefined results for
comparison, so I do not know what the result of this comparison is.
However the result of the cast of that to u_long should be ULONG_MAX.
If I remember correctly (but that also only works using IEEE 754 semantics).

bernd-edlinger:
I just tried what I said above:
print a NaN and debug.
Result: ufValue = `-nan(0x8000000000000`
`if (ufvalue > ULONG_MAX) {`  is not taken,
intpart = (unsigned long)ufvalue;
is `0x8000000000000000`
%f prints: "9223372036854775809.775808"
gcc 4.8.4 here.  will try with gcc-11 again.

paulidale:
There are **lots** of other floating point formats.  Most are historic these days.

Any comparison involving a NaN is defined to be false.  This leads to `!(x == x)` being used to check if `x` is a NaN.

bernd-edlinger:
nice idea, I could give it a try.

paulidale:
GPUs used to define their own short floating point formats.  IBM uses a base 16 floating point format on their big machines.  I can't remember the call but one of the Unix system calls used a funny floating point format for values it returned.

paulidale:
I remembered it: process accounting.

Look at `man 5 acct`, inside you'll find the `comp_t` type which has a 3 bit base 8 exponent with a 13 bit mantissa.

bernd-edlinger:
Nice, I would like to add a test case, and also your idea for checking
NaN numbers.

Would you think that is okay in this PR, or should I do that in another PR?

bernd-edlinger:
@paulidale I added some test cases, and fixed NaN handling,
please re-confirm.

Thanks
Bernd.

crypto/bio/b_print.c
```diff
@@ -635,7 +635,13 @@ fmtfp(char **sbuffer,
             fvalue = tmpvalue;
     }
     ufvalue = abs_val(fvalue);
-    if (ufvalue > ULONG_MAX) {
+    /*
+     * By subtracting 65535 (2^16-1) we canel the low order 15 bits
```
paulidale, line 639: cancel not canel

crypto/bio/b_print.c
```diff
@@ -635,7 +635,13 @@ fmtfp(char **sbuffer,
...
+     * of ULONG_MAX to avoid using imprecise floating point values.
+     * The second condition is necessary to catch NaN values.
+     */
+    if (ufvalue >= (double)(ULONG_MAX - 65535) + 65536.0
+            || ufvalue != ufvalue /* NaN */) {
```
paulidale, line 644: Stylistic, I'd use: `!(ufvalue == ufvalue)`.
I think it catches the reader's attention better.

test/bioprinttest.c
```diff
@@ -244,11 +244,42 @@ static int test_fp(int i)
...
 
     /* Test excessively big number. Should fail */
     if (!TEST_int_eq(BIO_snprintf(buf, sizeof(buf),
                                   "%f\n", 2 * (double)ULONG_MAX), -1))
         return 0;
+
+    d = 1.0;
+    z = 0.0;
+    inf = d / z;
+    nan = z / z;
```
paulidale, line 257: Would a sanity check: `TEST_false(nan == nan)` make sense?

It might be better in `sanitytest.c` and isn't all that relevant to cryptography (at least cryptography as it stands now).

Anyway, not in this PR.

crypto/bio/b_print.c
bernd-edlinger: Ok.

crypto/bio/b_print.c
bernd-edlinger: Ok, will do.

test/bioprinttest.c
bernd-edlinger: Yes, I have to agree it is not at all obvious for a crypto library to have a snprintf function...

bernd-edlinger:
oops, msvc does not like my test:
```
bioprinttest.c
c:\projects\openssl\test\bioprinttest.c(256) : error C2220: warning treated as error - no 'object' file generated
c:\projects\openssl\test\bioprinttest.c(256) : warning C4723: potential divide by 0
c:\projects\openssl\test\bioprinttest.c(257) : warning C4723: potential divide by 0
```

bernd-edlinger:
Sorry, I had to change the test case to fix a MSVC warning about 
"possible" division by zero.
Please re-confirm.

test/bioprinttest.c
```diff
@@ -244,11 +244,42 @@ static int test_fp(int i)
 static int test_big(void)
 {
     char buf[80];
+    volatile double d, z, inf, nan;
```
paulidale, line 247: What about making z a non-static global variable instead?

bernd-edlinger: could work. I can give it a try.

bernd-edlinger: I will name it "zero_value" so it is less likely to collide with other globals.

paulidale: It will also need `extern zero_value;` to appease the compiler warnings.

bernd-edlinger: it compiles on linux this way, I  dont see the reason why?

bernd-edlinger: Ouch clang is silly.

paulidale: Clang isn't silly.  It's pedantic.  It's the only compiler I've found that produces good code for:
```C
unsigned magic(unsigned n, unsigned d, int *o) {
    unsigned int r = n * d;

    *o = r / d != n;
    return r;
}
```

bernd-edlinger: Hmm, "-Wmissing-variable-declarations" is only in clang_devteams waring,
gcc would also implement this warning, so maybe that should be enabled
for gcc (unless we don't want this warning)

bernd-edlinger: > Clang isn't silly. It's pedantic. It's the only compiler I've found that produces good code for:

I checked the generated code, it optimizes magic(0,0,&r)
unter the assumption that 0/0 == 0.

paulidale: More specifically, it omits the division entirely.

bernd-edlinger:
Merged master    082c041, OpenSSL_1_1_1-stable 7d76c1f.
Thanks!

PR #11961, opened 2020-05-26
Deprecate SHA and MD5 again.

paulidale:
This reverts commit a978dc3bffb63e6bfc40fe6955e8798bdffb4e7e.

### DIFF
--- a/crypto/md5/build.info
+++ b/crypto/md5/build.info
@@ -17,6 +17,7 @@ ENDIF
 $COMMON=md5_dgst.c md5_one.c md5_sha1.c $MD5ASM
 SOURCE[../../libcrypto]=$COMMON
 SOURCE[../../providers/libimplementations.a]=$COMMON
+SOURCE[../../providers/liblegacy.a]=$COMMON
 
 # Implementations are now spread across several libraries, so the defines
 # need to be applied to all affected libraries and modules.
--- a/crypto/md5/build.info
+++ b/crypto/md5/build.info
@@ -17,7 +17,9 @@ ENDIF
 $COMMON=md5_dgst.c md5_one.c md5_sha1.c $MD5ASM
 SOURCE[../../libcrypto]=$COMMON
 SOURCE[../../providers/libimplementations.a]=$COMMON
-SOURCE[../../providers/liblegacy.a]=$COMMON
+IF[{- !$disabled{dso} -}]
+  SOURCE[../../providers/liblegacy.a]=$COMMON
+ENDIF
 
 # Implementations are now spread across several libraries, so the defines
 # need to be applied to all affected libraries and modules.
--- a/crypto/md5/build.info
+++ b/crypto/md5/build.info
@@ -17,6 +17,13 @@ ENDIF
 $COMMON=md5_dgst.c md5_one.c md5_sha1.c $MD5ASM
 SOURCE[../../libcrypto]=$COMMON
 SOURCE[../../providers/libimplementations.a]=$COMMON
+
+# A no-deprecated no-shared build ends up with double function definitions
+# without conditioning this on dso. The issue is MD5 which is needed in the
+# legacy provider for one of the spliced algorithms, however it resides in the
+# default provider.  A no-deprecated build removes the external definition from
+# libcrypto and this means that the code needs to be in liblegacy.  However,
+# when building without 'dso', liblegacy is included in libcrypto.
 IF[{- !$disabled{dso} -}]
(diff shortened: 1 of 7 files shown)


paulidale:
RC4_HMAC_MD5 in the legacy provider needs to access the MD5 low level calls, this caused the Travis failures.

paulidale:
Before this can go in, libssl needs to lose the special CBC mode implementation which calls the digests directly.

paulidale:
This depends on #12094.

paulidale:
Rebased to current master resolving conflicts in `libcrypto.num`.
Assuming the CI's suceed, I'll assume that the approval still holds.

paulidale:
This will need to be reviewed again even though the only change is `libcrypto.num`.

I dismissed the older review anticipating changes would be required after the TLS CBC code was moved but no changes were required.

levitte:
Travis failures because libssl.so refers directly to the now deprecated functions.  I have lost track on the effort to correct that, perhaps a rebase is sufficient?

paulidale:
It requires a fix in libssl to remove the direct references.  @mattcaswell is doing this.

levitte:
I'd suggest making this WIP in the mean time, and perhaps noting the PR this depends on, if there is one

paulidale:
No PR number yet.

paulidale:
With the TLS work finished, this one is ready for prime time again.

paulidale:
Should be ready for review now.

paulidale:
Ping?  This is an API deprecation.

crypto/md5/build.info
```diff
@@ -17,11 +17,15 @@ ENDIF
 $COMMON=md5_dgst.c md5_one.c md5_sha1.c $MD5ASM
 SOURCE[../../libcrypto]=$COMMON
 SOURCE[../../providers/libimplementations.a]=$COMMON
+IF[{- !$disabled{dso} -}]
```
kaduk, line 20: I see `deprecated-3.0` checks (and not `dso`) ones for, e.g., md2 and md4 (as well as comments supposedly explaining why).  I'm not nsure I understand why DSO is the right conditional here.

ssl/record/ssl3_record.c
```diff
@@ -1307,6 +1307,25 @@ int tls1_enc(SSL *s, SSL3_RECORD *recs, size_t n_recs, int sending,
     return 1;
 }
 
+/*
+ * ssl3_cbc_record_digest_supported returns 1 iff |ctx| uses a hash function
+ * which ssl3_cbc_digest_record supports.
+ */
+char ssl3_cbc_record_digest_supported(const EVP_MD_CTX *ctx)
```
kaduk, line 1314: Kind of bizzare that the return type is `char` (when it is going to get promoted to `int` anyway), but this is just moved code...

ssl/record/ssl3_record.c
```diff
@@ -1335,6 +1354,9 @@ int n_ssl3_mac(SSL *ssl, SSL3_RECORD *rec, unsigned char *md, int sending)
     if (!sending &&
         EVP_CIPHER_CTX_mode(ssl->enc_read_ctx) == EVP_CIPH_CBC_MODE &&
         ssl3_cbc_record_digest_supported(hash)) {
+#ifdef OPENSSL_NO_DEPRECATED_3_0
+        return 0;
```
kaduk, line 1358: It's a little weird that the commit message talks about MD5/SHA1 but the helper claims that SHA-2 is supported, but IIUC this is just used for sslv3 and thus the SHA-2 support is vestigal

kaduk (commented):
md5 moves to liblegacy but sha1 does not?

crypto/md5/build.info
paulidale: A `no-deprecated` `no-shared` build ends up with double function definitions otherwise.

The issue is MD5 which is needed in the legacy provider for one of the spliced algorithms, however it resides in the default provider.  A `no-deprecated` build removes the external definition from libcrypto and this means that the code needs to be in liblegacy.  However, when building without 'dso', liblegacy is included in libcrypto.

kaduk (approved):
Thanks for the extra explanation.
It would be nice to add that as a comment in the build.info, though.

paulidale:
I've added my comment to the file.

kaduk:
Thanks! (Reconfirm +1)

paulidale:
Thanks for the review and comments.  Finally merged.

PR #11963, opened 2020-05-26
Add header file docs for openssl/core_numbers.h and openssl/core_names.h

levitte:


mspncp:
As a side note related to this pr: looking at the content of `<openssl/core_numbers.h>` I wonder whether the header name `<openssl/core_functions.h>` wouldn't be more appropriate? It's all about defining provider _functions_ and the _number_, (i.e. the function ID) is just a part of their definition.

https://github.com/openssl/openssl/blob/e847085914476636d75ee1874b78e1c0e983da6e/include/openssl/core_numbers.h#L90-L128

Also, there are parts in the header where the IDs are intertwined with the function declarations, and other parts where the IDs precede the declarations blockwise.

https://github.com/openssl/openssl/blob/e847085914476636d75ee1874b78e1c0e983da6e/include/openssl/core_numbers.h#L130-L150

I find the latter variant more readable. Whichever variant you choose, it should be consistent.

Tymbur:
I don’t know why I received this it looks like it belongs to someone else
and it happened because of the github iOS app it looks like a security
issue to me there are enumerated blobs and shit in here .  This is me
saying report this error .  I don’t know if this were someone else besides
me if it would be a big deal or not but it looks like a personal email I’m
not sure.

mspncp:
@Tymbur I have no idea why you received the mail. It was a regular post of mine on pull request #11963. To  prevent further email notifications, you can follow the link to the pull request and unsubscribe from the thread.

Tymbur:
Sure thing just wanted to make sure it wasn’t something serious

levitte:
> As a side note related to this pr: looking at the content of `<openssl/core_numbers.h>` I wonder whether the header name `<openssl/core_functions.h>` wouldn't be more appropriate?

Not sure that's the best replacement names either, the name "core_functions" would make me expect to see a number of declarations for functions offered by the core of libcrypto...  regarding the name "core_numbers", I would argue that these *are* dispatch numbers offered by the core, along with the function types that the provider need to associate with those numbers.

levitte:
> Also, there are parts in the header where the IDs are intertwined with the function declarations, and other parts where the IDs precede the declarations blockwise.
> ...
> I find the latter variant more readable. Whichever variant you choose, it should be consistent.

I would say that it depends.  The functions offered by the core (!!!) is a looong list, that's a lot of jumping back and forth to see what function number macro corresponds to what function signature.
Although, one might argue that some operations have a long list of numbers/function signatures as well, where we separate the number block from the function signature block, so we're not quite consistent there.

levitte:
Now that I've commented on stuff that isn't about this PR itself at all, anyone want to review?

slontis:
There are some make doc-nits errors in travis.

doc/man7/openssl-core_names.h.pod
```diff
@@ -0,0 +1,49 @@
...
+openssl/core_names.h - OpenSSL provider parameter names
+
+=head1 SYNOPSIS
+
+ #include <openssl/core_names.h>
+
+=head1 DESCRIPTION
+
+The F<< <openssl/core_names.h> >> header defines a multitude of macros
+for L<OSSL_PARAM(3)> names, algorithm names and other known names used
```
slontis, line 14: It also contains some param values..

doc/man7/openssl-core_names.h.pod
```diff
@@ -0,0 +1,49 @@
...
+with OpenSSL's providers, made available for practical purposes only.
+
+Existing names are further described in the manuals for OpenSSL's
+providers (see L</SEE ALSO>) and the manuals for each algorithm they
+provide (listed in those provider manuals).
+
+=head1 SEE ALSO
+
+L<OSSL_PROVIDER-default(7)>, L<OSSL_PROVIDER-FIPS(7)>
```
slontis, line 23: legacy and NULL also?

doc/man7/openssl-core_names.h.pod
```diff
@@ -0,0 +1,49 @@
...
+
+=head1 HISTORY
+
+The macros described here were added in OpenSSL 3.0.
+
+=head1 CAVEATS
+
+=begin comment
+
+CAVEATS is a section name borrowed from OpenBSD, where mdoc(7)
```
slontis, line 33: Is this comment really needed?

doc/man7/openssl-core_names.h.pod
```diff
@@ -0,0 +1,49 @@
...
+=end comment
+
+I<This header file does not constitute a general registry of names>.
+Providers that implement new algorithms are to be responsible for
+their own parameter names.
+
+However, authors of provider that implement their own variants of
+algorithms that OpenSSL providers support will want to pay attention
+to the names provided in this header to work in a compatible manner.
+
```
slontis, line 49: missing the stuff normally at the ned 

=cut

doc/man7/openssl-core_numbers.h.pod
```diff
@@ -0,0 +1,56 @@
+=pod
+
+=head1 NAME
+
+openssl/core_numbers.h
+- OpenSSL provider interface numbers and function types
```
slontis, line 6: for numbers maybe?

doc/man7/openssl-core_names.h.pod
levitte: I know some nit-picky people that are likely to ask.

doc/man7/openssl-core_names.h.pod
levitte: Ah, thanks!

doc/man7/openssl-core_numbers.h.pod
levitte: Er, no?  But perhaps "dispatch numbers" is better than "interface numbers"

doc/man7/openssl-core_names.h.pod
levitte: Aren't they "other known names"?

doc/man7/openssl-core_names.h.pod
slontis: Yeah I understand why you did it :)
Maybe move to the git comments instead?

doc/man7/openssl-core_names.h.pod
slontis: ok

doc/man7/openssl-core_numbers.h.pod
slontis: identifiers and function wrappers?

doc/man7/openssl-core_names.h.pod
levitte: The null provider doesn't provide much, so I don't see why it should be mentioned here.  I've added the legacy provider to the list, though.

doc/man7/openssl-core_numbers.h.pod
levitte: ... especially since I use "dispatch number" everywhere else.  Done

levitte: > identifiers and function wrappers?

Those are new terms that we have never used before in this context...  and "wrappers"?  that's doesn't connect with what they are, at least to me.

slontis: Ok.. I cant think of anything better.

doc/man7/openssl-core_names.h.pod
levitte: Oh, I forgot about this.  I did as you suggested

levitte:
Care to re-approve with the squash commit I just added?

richsalz:
I dislike adding a new CAVEATS section.  OpenSSL has always used NOTES, which seems quite appropriate here.

Don't blaze new trails.  Be boring.  Be  consistent.

levitte:
You often want to collapse NOTES into DESCRIPTION too... :wink:

I think this particular one does deserve a bit more "voice" than a mere note...

richsalz:
Yes, I often collapse NOTES into DESCRIPTION because it doesn't deserve a separate call-out, and folks get lazy by just adding NOTES after the main manpage is written.

Look at other notes sections.  This is no different.

I'm sure I'll lose this point, sigh, but you're still wrong.

levitte:
Merged.

329b2a2cde480fbfe1cf43bf8ecf1c75a16cf6bf DOCS: add openssl-core_numbers.h(7)
f438f53a4e57462216be271c1c965550b6ff9941 DOCS: add openssl-core_names.h(7)

PR #11944, opened 2020-05-24
ossl_shim: add deprecation guards around the -use-ticket-callback option.

paulidale:
The ticket callback is deprecated in 3.0 and can't be used in a no-deprecated build.

Fixes #11942

- [ ] documentation is added or updated
- [x] tests are added or updated

PR #11966, opened 2020-05-26
ossl_shim: use the correct ticket key call back.

paulidale:
Fixing the bad fix in #11944.


paulidale:
Merged to master.  Thanks.

PR #11967, opened 2020-05-27
Fix errtest for older compilers

slontis:
Some older compilers use "unknown function" if they dont support __func, so the
test using ERR_PUT_error needed to compensate for this when comparing against the
expected value.

##### Checklist


slontis:
Thanks. Merged to master.

PR #10457, opened 2019-11-15
AES CTR-DRGB: use platform specific AES-CTR support

p-steuer:
Next try to get HW supported AES-CTR for CTR-DRBG's generate operation.

Related:

- My original 2017 PR: https://github.com/openssl/openssl/pull/4530
(was rejected/stalled due to not using EVP and worries about internal state's counter increment not exactly matching the NIST pseudo-code/reference implementation)

- Today's PR: https://github.com/openssl/openssl/pull/10456
(tries to do the same only for x86)

This one uses EVP, keeps internal state counter in sync with reference implementation and should already pass the test suite.

TODO:
- simplyfiy if-statement in generate for loop - done.
- look at performance especially for small chunks - done.
- add test case(s) that cover counter wrapping (think my old PR had that one) - done.

PR #11968, opened 2020-05-27
AES CTR-DRGB: performance improvement

bernd-edlinger:
Optimize the the AES-based implementation of the CTR_DRBG
construction, see 10.2.1 in [1].
Due to the optimizations, the code may deviate (more) from the
pseudocode in [1], but it is functional equivalence being decisive
for compliance:

"All DRBG mechanisms and algorithms are described in this document
in pseudocode, which is intended to explain functionality.
The pseudocode is not intended to constrain real-world
implementations." [9 in [1]].

The following optimizations are done:

- Replace multiple plain AES encryptions by a single AES-ECB
  encryption of a corresponding pre-initialized buffer, where
  possible.
  This allows platform-specific AES-ECB support to
  be used and reduces the overhead of multiple EVP calls.

- Replace the generate operation loop (which is a counter
  increment followed by a plain AES encryption) by a
  loop which does a plain AES encryption followed by
  a counter increment. The latter loop is just a description
  of AES-CTR, so we replace it by a single AES-CTR
  encryption.
  This allows for platform-specific AES-CTR support to be used
  and reduces the overhead of multiple EVP calls.
  This change, that is, going from a pre- to a post- counter
  increment, requires the counter in the internal state
  to be kept at "+1" (compared to the pseudocode in [1])
  such that it is in the correct state, when a generate
  operation is called.
  That in turn also requires all other operations to be
  changed from pre- to post-increment to keep functional
  equivalence.

[1] NIST SP 800-90A Revision 1

Signed-off-by: Patrick Steuer <patrick.steuer@de.ibm.com>

(Merged from https://github.com/openssl/openssl/pull/10457)

(cherry picked from commit 28bdbe1aaa474ae8cd83e520d02e463e46ce89d9)

##### Checklist

@p-steuer FYI, I cherry-picked your commit and fixed a few trivial conflicts,
I hope you don't mind.

### DIFF
--- a/crypto/rand/drbg_ctr.c
+++ b/crypto/rand/drbg_ctr.c
@@ -63,15 +63,15 @@ static void ctr_XOR(RAND_DRBG_CTR *ctr, const unsigned char *in, size_t inlen)
  * Process a complete block using BCC algorithm of SP 800-90A 10.3.3
  */
 __owur static int ctr_BCC_block(RAND_DRBG_CTR *ctr, unsigned char *out,
-                                const unsigned char *in)
+                                const unsigned char *in, int len)
 {
     int i, outlen = AES_BLOCK_SIZE;
 
-    for (i = 0; i < 16; i++)
+    for (i = 0; i < len; i++)
         out[i] ^= in[i];
 
-    if (!EVP_CipherUpdate(ctr->ctx_df, out, &outlen, out, AES_BLOCK_SIZE)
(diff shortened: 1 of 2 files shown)


p-steuer:
@bernd-edlinger Im fine with it .

bernd-edlinger:
Merged: OpenSSL_1_1_1-stable 53eb05b 
Thanks!

PR #11969, opened 2020-05-27
util/mkpod2html.pl: Fix unbalanced quotes

levitte:


levitte:
I'm classifying it as urgent, because trying to use the output in OpenSSL's manmaster tree has it look ... quite bad.  Current example: https://beta.openssl.org/docs/manmaster/man1/openssl-asn1parse.html

levitte:
Merged

bb90f9fee1a3055f8a9cd63aad691b1bfae3e53e util/mkpod2html.pl: Fix unbalanced quotes

Issue #11449, opened 2020-03-31
BIO_connect_retry will  Endless loop in BIO_CONN_S_CONNECT  return ETIMEDOUT


yangyangtiantianlonglong:
```javascript
OpenSSL> version
OpenSSL 3.0.0-dev xx XXX xxxx
```
https://github.com/openssl/openssl/blob/f4c88073091592b1ff92ba12c894488ff7d03ece/crypto/bio/bss_conn.c#L156-L162
https://github.com/openssl/openssl/blob/f4c88073091592b1ff92ba12c894488ff7d03ece/crypto/bio/bio_lib.c#L853-L868
issue: sock connect don't stop
1.   The func conn_state  lines 158 BIO_connect return 0, get_last_sys_error return  ETIMEDOUT.
2.   The func BIO_connect_retry  lines 868 will goto retry.
3.    BIO_do_handshake --> conn_state -->BIO_do_handshake will endless loop.

yangyangtiantianlonglong:
Why the system ETIMEDOUT does not gradually reduce the timeout time？and always endless loop connect.

DDvO:
> Why the system ETIMEDOUT does not gradually reduce the timeout time？and always endless loop connect.

The "goto retry" is needed because of some strange behavior of `BIO_do_handshake()` at least in the blocking case.
Good point that (in the non-bocking case) the timeout condition should be checked.
I'm going to provide a fix for that.

DDvO:
@yangyangtiantianlonglong, does the following patch solve the problem for you?
```
diff --git a/crypto/bio/bio_lib.c b/crypto/bio/bio_lib.c
index 1579f7c366..ed893e2ce2 100644
--- a/crypto/bio/bio_lib.c
+++ b/crypto/bio/bio_lib.c
@@ -867,8 +867,8 @@ int BIO_connect_retry(BIO *bio, int timeout)
              * unless using BIO_reset(), blocking next connect() may crash and
              * non-blocking next BIO_do_handshake() will fail
              */
-            goto retry;
-        } else if (BIO_should_retry(bio)) {
+        }
+        if (get_last_sys_error() == ETIMEDOUT || BIO_should_retry(bio)) {
             /* will not actually wait if timeout == 0 (i.e., blocking BIO) */
             rv = bio_wait(bio, max_time, 100 /* milliseconds */);
             if (rv > 0)

```

DDvO:
This issue should be fixed by #11986.

PR #11971, opened 2020-05-27
Make BIO connect to allow multiple IP addresses

DDvO:
When doing CLI-based tests with the new CMP client and mock server I faced strange 
`Connection refused` errors when using non-blocking I/O with timeout enabled.

It took me quite a while to find out that this only happened when connecting to, e.g., `localhost` (but not when using IPv4 addresses such as `127.0.0.1`) and only when `/etc/hosts` contains both IPv4 and IPv6 addresses for `localhost` and when IPv6 is preferred.

It turns out that `conn_state()` in `crypto/bio/bss_conn.c` is simply not flexible enough to handle alternative IP addresses on connect (`BIO_C_DO_STATE_MACHINE`). This made `BIO_connect_retry()` fail on the first retry when the host is `localhost` and the `timeout` parameter is non-zero.

The patch given fixes this problem by making sure that *all* IP addresses (of any type) are tried in case a hostname (i.e., DNS name) is resolved to more than one IP address.

A workaround at least in my case is to comment out in `/etc/hosts/` the line
```
::1             localhost                                                                                                                                           
```

### DIFF
--- a/crypto/bio/bss_conn.c
+++ b/crypto/bio/bss_conn.c
@@ -188,8 +188,17 @@ static int conn_state(BIO *b, BIO_CONNECT *c)
 
         case BIO_CONN_S_BLOCKED_CONNECT:
             i = BIO_sock_error(b->num);
-            if (i) {
+            if (i != 0) {
                 BIO_clear_retry_flags(b);
+                if ((c->addr_iter = BIO_ADDRINFO_next(c->addr_iter)) != NULL) {
+                    /*
+                     * if there are more addresses to try, do that first
+                     */
+                    BIO_closesocket(b->num);
+                    c->state = BIO_CONN_S_CREATE_SOCKET;
+                    ERR_clear_error();
+                    break;
+                }
                 ERR_raise_data(ERR_LIB_SYS, i,
                                "calling connect(%s, %s)",
                                 c->param_hostname, c->param_service);


levitte:
I find it weird to do this in the blocked state...  and I have a hard time understanding why you'd have a configuration where one IP version is *blocked* while the other isn't.  But there's new stuff going on that I frankly don't always understand...  back when I was doing network management, a block was a block was a block.

DDvO:
> I find it weird to do this in the blocked state... >

For whatever reason, I got the issue only in non-blocking mode (when `BIO_connect_retry()` is given a timeout), but there might be situations where the fix is of interest also for blocking mode, and at least it should not hurt then.

> and I have a hard time understanding why you'd have a configuration where one IP version is _blocked_ while the other isn't.

That's not the point here. 
I'm not really a network guy, but it appears to me that the very basic HTTP server in `apps/lib/http_server.c` accepts only IPv4 connections, or there must be some other reason why at least on some Linux systems IPv6 connections to the loopback device are not accepted while IPv4 works fine.

DDvO:
>  But there's new stuff going on that I frankly don't always understand... back when I was doing network management, a block was a block was a block.

BTW, `bss_conn.c` and siblings is not the most beautiful and readable piece of code.
And I have a hard time understanding why all that retry business is needed after all,
in particular why a blocking connect sometimes needs to be retried, which required some ugly workarounds and an apparent bug in `BIO_connect_retry()`. See also #11449.

crypto/bio/bss_conn.c
```diff
@@ -188,7 +188,16 @@ static int conn_state(BIO *b, BIO_CONNECT *c)
...
+                if ((c->addr_iter = BIO_ADDRINFO_next(c->addr_iter)) != NULL) {
+                    /*
+                     * if there are more addresses to try, do that first
+                     */
+                    BIO_closesocket(b->num);
+                    c->state = BIO_CONN_S_CREATE_SOCKET;
+                    ERR_clear_error();
+                    break;
+                }
                 BIO_clear_retry_flags(b);
```
bernd-edlinger, line 201: In general I think this makes sense, if the blocking connect fails the
error handling is here:
https://github.com/openssl/openssl/blob/c74aaa3920f116fe4c1003153838144c37d6e527/crypto/bio/bss_conn.c#L167-L175
but if the non-blocking connect does not fail immediately
the error is in the SO_ERROR and we end up here.
Not sure if the call to BIO_clear_retry_flags should move up, since
the connect was retried and failed.  This flag might be visible in the callback.

bernd-edlinger:
I think this should be back-ported to 1.1.1

crypto/bio/bss_conn.c
DDvO: Thanks @bernd-edlinger for having a look at this and commenting.
When I debugged this I found that for non-blocking BIOs first `BIO_CONN_S_CONNECT` is reached and fails, and on the first retry (done by `BIO_connect_retry()` `BIO_CONN_S_BLOCKED_CONNECT` is reached.
I don't understand why this complicated retry is needed at all, but it works.

bernd-edlinger: Yeah, a name resolution can yield several addresses, this tries each address,
until the first succeeds.
In my code when I write a server I usually let the user decide on IPv4 or IPv6 first,
and then try in that sequence for instance.  I think we have added s_client option:
```
 -4                         Use IPv4 only
 -6                         Use IPv6 only
```
does that work for you?

kroeckx:
I think you're just using the API wrong. On failure, the application calling BIO_connect() should use the next address, and that it what for instance s_client does in init_client(). It iterates over the addresses.

kroeckx:
This also most likely seem to cause errors, since the address family of the socket you've created with BIO_socket() doesn't actually match with what BIO_connect() is trying to do now

kroeckx:
Or am mixing up the APIs

bernd-edlinger:
I may be wrong as well....
I think the error handling in the blocking case seems to pick the next address in a list.
The same should happen in the non-blocking case.

crypto/bio/bss_conn.c
DDvO: I certainly could add such options to the CMP client, but I see not much motivation to do so. And the corrected retry behavior may be needed also for other reasons, not only IP v4 vs. v6.

kroeckx:
Clearly I was mixing up APIs.

That code at least looks weird. I think it's missing a BIO_sock_should_retry() call. I think we can be in 3 states: success, failure, still waiting. And it seems now it's only success or failure.

DDvO:
@kroeckx, `BIO_connect()` seems to operate on a specific address only, while `BIO_do_connect()` should be more flexible AFAICS, but their documentation is pretty poor. Nothing is written there about the iterations potentially needed over several IP addresses, and it would be pretty strange to me if the application had to do the switch from one such address to the next. And even if so, there should be a properly documented API function for doing that.

DDvO:
So it seems we agree the IP address iteration should be handled internally (like the fix now make sure)?

bernd-edlinger:
see the case BIO_CONN_S_CONNECT:
that is where the BIO_connect happens,
in the case when BIO_sock_should_retry returns true,
enter the BIO_CONN_S_BLOCKED_CONNECT state.
unclear why the s_client does not use this API at all.

kroeckx:
I think the BIO_CONN_S_BLOCKED_CONNECT is at least a confusing name, I think it's waiting for the connect to succeed there. I was also expecting that if you ran out of addresses you'd go to BIO_CONN_S_CONNECT_ERROR

bernd-edlinger:
Yes, agree, maybe move the `BIO_clear_retry_flags(b);` before the
 `if ((c->addr_iter = BIO_ADDRINFO_next(c->addr_iter)) != NULL) {`

bernd-edlinger:
However while a connection to localhost will always fail quickly,
so no noticeable delay, when the wrong protocol is tried,
a connection to a remote node, may take longer, usually 1 minute,
until it fails, and then you will get complaints...

DDvO:
A quick look at `s_client.c` does not really give enlightenment.
But meanwhile I found the iteration of `BIO_connect()` @kroeckx has been referring to: it's buried in `apps/lib/s_socket.c` and rather ugly, so not something a normal app should be doing.

bernd-edlinger:
Yeah, s_client does not use this API, but it would be a good idea to use this there.
s_client just picks the first IPv4/IPv6 address, so it solves the same problem,
in a different way.

DDvO:
s_client also does try alternative addresses, via `init_client()`, which is in `s_socket.c` as I just wrote.

crypto/bio/bss_conn.c
bernd-edlinger: There must be a way to set connect_family to BIO_FAMILY_IPV4
BIO_C_SET_CONNECT: /num == 3
this is rotten bad code without any comments.

bernd-edlinger: BIO_set_conn_ip_family(b,BIO_FAMILY_IPV4)

DDvO:
> Yeah, s_client does not use this API, but it would be a good idea to use this there.

I also have the impression it would be better if s_client used `BIO_do_connect()`, with the improvement given here. Or even better, used the rather new `BIO_connect_retry()`, which I introduced along the the generalized HTTP client.

crypto/bio/bss_conn.c
bernd-edlinger: I can only say, that it is really worth it to give the user a way to prefer
IPv4 or IPv6, not only in this project.
Of course lots of other things can go wrong but then someone else
gets the blame...

bernd-edlinger:
I think the commit message should mention BIO_do_connect,
to avoid the impression that BIO_connect should try different addresses.

bernd-edlinger:
> Or even better, used the rather new BIO_connect_retry()

You should fix the BIO_socket_wait function before you do that:
This `tv.tv_sec = (long)(max_time - now); /* might overflow */ `
cannot overflow, IMHO, and I don't understand why the type case is necessary at all.
But `select(fd + 1`  can overflow,
and especrially this `openssl_fdset(fd, &confds);` 
will crash if fd >= 1023, and why not check for fd == -1 ?

bernd-edlinger:
Hmm, actually, I don't really see this function in libcrypto.
What is the point in doing non-blocking I/O when the function is designed to
only handle one socket at a time?

bernd-edlinger:
Anyway, this PR is okay, also for 1.1.1,
please change the commit message, to mention BIO_do_connect and
move BIO_clear_retry_flags(b); to line 192.

kroeckx:
I also don't see the point of BIO_connect_retry(). It seems to
turn the non-blocking case in a blocking case. Maybe the behaviour
should just change depending on BIO_set_nbio()?

crypto/bio/bss_conn.c
DDvO: Of course for apps like `s_client` focusing on network connections it makes much sense to have `-4` and `-6` options, while for the specific CMP client use case I'm still a bit reluctant to add such rather network-special options, in particular since this app already has tons of options.

On the other hand, the four HTTP client functions declared in `include/openssl/http.h`:
```
BIO *OSSL_HTTP_get(const char *url, const char *proxy, const char *no_proxy,
                   BIO *bio, BIO *rbio,
                   OSSL_HTTP_bio_cb_t bio_update_fn, void *arg,
                   const STACK_OF(CONF_VALUE) *headers,
                   int maxline, unsigned long max_resp_len, int timeout,
                   const char *expected_content_type, int expect_asn1);
ASN1_VALUE *OSSL_HTTP_get_asn1(const char *url,
                               const char *proxy, const char *no_proxy,
                               BIO *bio, BIO *rbio,
                               OSSL_HTTP_bio_cb_t bio_update_fn, void *arg,
                               const STACK_OF(CONF_VALUE) *headers,
                               int maxline, unsigned long max_resp_len,
                               int timeout, const char *expected_content_type,
                               const ASN1_ITEM *it);
ASN1_VALUE *OSSL_HTTP_post_asn1(const char *server, const char *port,
                                const char *path, int use_ssl,
                                const char *proxy, const char *no_proxy,
                                BIO *bio, BIO *rbio,
                                OSSL_HTTP_bio_cb_t bio_update_fn, void *arg,
                                const STACK_OF(CONF_VALUE) *headers,
                                const char *content_type,
                                const ASN1_VALUE *req, const ASN1_ITEM *req_it,
                                int maxline, unsigned long max_resp_len,
                                int timeout, const char *expected_ct,
                                const ASN1_ITEM *rsp_it);
BIO *OSSL_HTTP_transfer(const char *server, const char *port, const char *path,
                        int use_ssl, const char *proxy, const char *no_proxy,
                        BIO *bio, BIO *rbio,
                        OSSL_HTTP_bio_cb_t bio_update_fn, void *arg,
                        const STACK_OF(CONF_VALUE) *headers,
                        const char *content_type, BIO *req_mem,
                        int maxline, unsigned long max_resp_len, int timeout,
                        const char *expected_ct, int expect_asn1,
                        char **redirection_url);
```
are meant to be as generally applicable as possible.
Shall I add to those an `int family` parameter?

DDvO:
> I think the commit message should mention BIO_do_connect,
> to avoid the impression that BIO_connect should try different addresses.

Done.

DDvO:
> > Or even better, used the rather new BIO_connect_retry()
> 
> You should fix the BIO_socket_wait function before you do that:
> This `tv.tv_sec = (long)(max_time - now); /* might overflow */ `
> cannot overflow, IMHO, and I don't understand why the type case is necessary at all.

As far as I recall, the type cast was needed for Windows builds (AppVeyor),
and depending on the size of `long` and `time_t` in general this might overflow.

> But `select(fd + 1` can overflow,
> and especrially this `openssl_fdset(fd, &confds);`
> will crash if fd > 1023, and why not check for fd == -1 ?

So I've added the following input check to the function:
```
    if (fd < 0 || fd >= FD_SETSIZE - 1)
        return -1;
```

crypto/bio/b_sock.c
```diff
@@ -387,6 +387,8 @@ int BIO_socket_wait(int fd, int for_read, time_t max_time)
     struct timeval tv;
     time_t now;
 
+    if (fd >= FD_SETSIZE - 1)
```
bernd-edlinger, line 390: fd >= FD_SETSIZE but please in a separate PR.

DDvO: All right, will do in a separate PR.
Yet don't I need to check for `fd >= FD_SETSIZE - 1` since the function uses `fd + 1`?

crypto/bio/bss_conn.c
bernd-edlinger: It is just my experience, with OPC/UA Clients that it will probably be helpful.
But when I look at those parameters, they are easy to confuse, 90% of the paramters
are just how to establish a tcp connection, but what about udp?
Maybe also add a protocol parameter?
And  put all the common how-to-connect parameters into a structure, maybe
with a parser, that parses a connection string and fill in all the params?

DDvO:
> I also don't see the point of BIO_connect_retry(). It seems to turn the non-blocking case in a blocking case. Maybe the behaviour should just change depending on BIO_set_nbio()?

The point of `BIO_connect_retry()` is two-fold:
* add a timeout feature (which internally needs non-blocking)
* deal with strange behavior of `BIO_do_connect()` (aka `BIO_do_handshake()`) as detailed in the comments included in its code

This is mentioned in its documentation:
```
BIO_connect_retry() connects via the given B<bio>, retrying BIO_do_connect()
until success or a timeout or error condition is reached.
```

crypto/bio/bss_conn.c
bernd-edlinger: but also please as another PR.

crypto/bio/b_sock.c
bernd-edlinger: no.

crypto/bio/bss_conn.c
DDvO: Yep, will do as soon as I have more bandwidth again.
I also was thinking that the many function parameter need to be abstracted/structured.

bernd-edlinger, line 201: Okay, I'd still move this BIO_clear_retry_flags up before the if,
since the previous connect is finished unsucessfully now.
(at least unless that does not work for any reason)

DDvO: Done

DDvO:
Merged - thanks all involved!

richsalz:
This seems like a change that would be noticed.  manpage and/or CHANGES entry?

DDvO:
> This seems like a change that would be noticed. manpage and/or CHANGES entry?

I've made a new issue since this PR is closed: https://github.com/openssl/openssl/pull/12017

crypto/bio/bss_conn.c
```diff
@@ -188,8 +188,17 @@ static int conn_state(BIO *b, BIO_CONNECT *c)
...
         case BIO_CONN_S_BLOCKED_CONNECT:
             i = BIO_sock_error(b->num);
-            if (i) {
+            if (i != 0) {
                 BIO_clear_retry_flags(b);
+                if ((c->addr_iter = BIO_ADDRINFO_next(c->addr_iter)) != NULL) {
+                    /*
+                     * if there are more addresses to try, do that first
+                     */
+                    BIO_closesocket(b->num);
```
bernd-edlinger, line 197: This is copied from above.
b->num should be initialized to INVALID_SOCKET.
since it takes a risk that the socket is close twice. 
I think both places should clear that value since the
BIO_socket could fail, and the following BIO_reset
will close the socket a second time.

t8m: We need either new PR or add it to #12017

PR #11972, opened 2020-05-27
Ensure CMAC_CTX is correctly initialised

### LINKED ISSUE #11957: CMAC EVP_aes_128_xts() outputs uninitialized memory
OSS-Fuzz #22453

```c
#include <openssl/cmac.h>
int main(void)
{
    OPENSSL_init_crypto(OPENSSL_INIT_LOAD_CRYPTO_STRINGS, NULL);
    unsigned char key[32] = {0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20,
        0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0xff};
    unsigned char cleartext[32] = { 0 };
    CMAC_CTX* ctx = CMAC_CTX_new();
    CF_CHECK_EQ(CMAC_Init(ctx, key, sizeof(key), EVP_aes_128_xts(), NULL), 1);
    CF_CHECK_EQ(CMAC_Update(ctx, cleartext, sizeof(cleartext)), 1);
    {
        size_t len = 0;
        uint8_t out[EVP_MAX_MD_SIZE];
        CF_CHECK_EQ(CMAC_Final(ctx, out, &len), 1);
        if ( len ) {
            printf("%02X\n", out[0]);
        }
    }
end:
    return 0;
}
```

```out[0]``` will be uninitialized upon printing it, which you can verify with valgrind.

### PR DESCRIPTION
**Pull request by mattcaswell**
The function EVP_Cipher() is very low level and has some rather odd return value semantics (see the docs). Essentially we need to treat any return value from that of 0 or less as an error - however the CMAC code was treating it as a boolean.

In fixing this issue I discovered that although we have EVP level CMAC tests we have no tests for the low level API. So I have created a simple test, heavily based on the existing hmactest. This actually uncovered a couple more bugs in CMAC initialisation, where we would attempt to use a CMAC_CTX that has only been partially initialised.

Fixes #11957

### DIFF
--- a/crypto/cmac/cmac.c
+++ b/crypto/cmac/cmac.c
@@ -125,12 +125,18 @@ int CMAC_Init(CMAC_CTX *ctx, const void *key, size_t keylen,
         return 1;
     }
     /* Initialise context */
-    if (cipher && !EVP_EncryptInit_ex(ctx->cctx, cipher, impl, NULL, NULL))
-        return 0;
+    if (cipher != NULL) {
+        /* Ensure we can't use this ctx until we also have a key */
+        ctx->nlast_block = -1;
+        if (!EVP_EncryptInit_ex(ctx->cctx, cipher, impl, NULL, NULL))
+            return 0;
+    }
     /* Non-NULL key means initialisation complete */
-    if (key) {
+    if (key != NULL) {
         int bl;
 
+        /* If anything fails then ensure we can't use this ctx */
+        ctx->nlast_block = -1;
         if (!EVP_CIPHER_CTX_cipher(ctx->cctx))
             return 0;
         if (!EVP_CIPHER_CTX_set_key_length(ctx->cctx, keylen))
--- a/crypto/cmac/cmac.c
+++ b/crypto/cmac/cmac.c
@@ -145,7 +145,7 @@ int CMAC_Init(CMAC_CTX *ctx, const void *key, size_t keylen,
             return 0;
         if ((bl = EVP_CIPHER_CTX_block_size(ctx->cctx)) < 0)
             return 0;
-        if (!EVP_Cipher(ctx->cctx, ctx->tbl, zero_iv, bl))
+        if (EVP_Cipher(ctx->cctx, ctx->tbl, zero_iv, bl) <= 0)
             return 0;
         make_kn(ctx->k1, ctx->tbl, bl);
         make_kn(ctx->k2, ctx->k1, bl);
@@ -186,12 +186,12 @@ int CMAC_Update(CMAC_CTX *ctx, const void *in, size_t dlen)
             return 1;
         data += nleft;
         /* Else not final block so encrypt it */
-        if (!EVP_Cipher(ctx->cctx, ctx->tbl, ctx->last_block, bl))
+        if (EVP_Cipher(ctx->cctx, ctx->tbl, ctx->last_block, bl) <= 0)
             return 0;
     }
     /* Encrypt all but one of the complete blocks left */
     while (dlen > (size_t)bl) {
-        if (!EVP_Cipher(ctx->cctx, ctx->tbl, data, bl))
+        if (EVP_Cipher(ctx->cctx, ctx->tbl, data, bl) <= 0)
             return 0;
         dlen -= bl;
         data += bl;
--- a/test/cmactest.c
+++ b/test/cmactest.c
@@ -1,0 +1,216 @@
+/*
+ * Copyright 1995-2020 The OpenSSL Project Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License 2.0 (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by mattcaswell** [comment]
This needs backport to 1.1.1, but I'll do that in a separate PR once this is approved.

**Comment by levitte** [comment]
Side note: I don't understand what makes `EVP_Cipher()` more low-level than the other EVP functions...

**Comment by mattcaswell** [comment]
> Side note: I don't understand what makes EVP_Cipher() more low-level than the other EVP functions...

It goes straight to the underlying "cipher" function. It doesn't handle anything around padding or partial blocks etc.

**Comment by levitte** [comment]
Okie. So not quite as low level as, for example, using AES functions without going through EVP (which is where I place the bar on "low level"). How many levels of low-level are we going to end up with, say? :wink:

**Comment by slontis** [comment]
The travis errors are valid.. So this will need reapproval.

File: test/cmactest.c [hunk 1]
```diff
@@ -0,0 +1,216 @@
...
+
+#include "internal/nelem.h"
+
+#include <openssl/cmac.h>
+#include <openssl/aes.h>
+#include <openssl/evp.h>
+
+#include "testutil.h"
+
+const char xtskey[32] = {
```
levitte, line 28: Make this static too and Travis will get happier

crypto/cmac/cmac.c
```diff
@@ -180,12 +186,12 @@ int CMAC_Update(CMAC_CTX *ctx, const void *in, size_t dlen)
...
         data += nleft;
         /* Else not final block so encrypt it */
-        if (!EVP_Cipher(ctx->cctx, ctx->tbl, ctx->last_block, bl))
+        if (EVP_Cipher(ctx->cctx, ctx->tbl, ctx->last_block, bl) <= 0)
             return 0;
     }
     /* Encrypt all but one of the complete blocks left */
     while (dlen > (size_t)bl) {
-        if (!EVP_Cipher(ctx->cctx, ctx->tbl, data, bl))
+        if (EVP_Cipher(ctx->cctx, ctx->tbl, data, bl) <= 0)
```
bernd-edlinger, line 194: Is EVP_Cipher ever expeced to return < 0 ?

crypto/cmac/cmac.c
```diff
@@ -139,7 +145,7 @@ int CMAC_Init(CMAC_CTX *ctx, const void *key, size_t keylen,
             return 0;
         if ((bl = EVP_CIPHER_CTX_block_size(ctx->cctx)) < 0)
             return 0;
-        if (!EVP_Cipher(ctx->cctx, ctx->tbl, zero_iv, bl))
+        if (EVP_Cipher(ctx->cctx, ctx->tbl, zero_iv, bl) <= 0)
```
bernd-edlinger, line 148: I ask because I wondered why the test case did not exhibit the same behavior on 1.1.1
since this EVP_Cipher call returned 0, but it seems that may have changed while
reworking everything at once.  Now I see it is possible to return -1 when an error happens.

crypto/cmac/cmac.c
mattcaswell: Yes it is documented to do so in some situations (and indeed does do it in the issue that sparked this PR in the first place #11957)

bernd-edlinger: In this test case the return value of EVP_Cipher was 0 in 1.1.1
and -1 in master, which one was right ?
And where is this documented?

mattcaswell: I'll investigate the discrepancy between 1.1.1 and master.

The `EVP_Cipher` man page says this:

````
EVP_Cipher() returns the amount of encrypted / decrypted bytes, or -1
on failure, if the flag B<EVP_CIPH_FLAG_CUSTOM_CIPHER> is set for the
cipher.  EVP_Cipher() returns 1 on success or 0 on failure, if the flag
B<EVP_CIPH_FLAG_CUSTOM_CIPHER> is not set for the cipher.
````

mattcaswell: > In this test case the return value of EVP_Cipher was 0 in 1.1.1
and -1 in master, which one was right ?

It turns out they are both correct behaviour. In master the provided xts implementation has the EVP_CIPH_FLAG_CUSTOM_CIPHER flag defined against it (due to f7397f0d58). But in 1.1.1 this is not the case. The presence or absence of this flag changes the return value semantics of the EVP_Cipher() function (as I noted above). Therefore both master and 1.1.1 are behaving in accordance with the documentation.

test/cmactest.c
mattcaswell: Fixed

mattcaswell:
@levitte - please reconfirm

kroeckx:
So this fails on a no-cmac build

mattcaswell:
> So this fails on a no-cmac build

Darn. Fixed now.

@levitte - can you reconfirm?

PR #11973, opened 2020-05-27
Add some fies to .gitignore

mattcaswell:
The previous attempt to fix .gitignore didn't quite get it right.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,10 @@
 doc/man1/openssl-*.pod
 
 # Auto generated der files
+providers/common/der/der_dsa.c
+providers/common/der/der_ec.c
+providers/common/der/der_rsa.c
+providers/common/der/der_digests.c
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -29,6 +29,10 @@
 doc/man1/openssl-*.pod
 
 # Auto generated der files
+providers/common/der/der_dsa.c
+providers/common/der/der_ec.c
+providers/common/der/der_rsa.c
+providers/common/der/der_digests.c
```
levitte, line 35: They shouldn't exist at all at this point, as they got renamed. Should .gitignore contain historical records

mattcaswell: Hmm. No. My mistake - I saw these files showing in git and assumed they had been missed from .gitignore. They must have been hanging around from a build on a branch without your renaming updates.

PR #11982, opened 2020-05-28
APPS: Create a library context in the main app, and pass it to commands 

levitte:
This turned out to have some consequences...  the NCONF functionality is slightly modified to give better flexibility.

Note that none of the sub-commands are modified to use the create library context, but it's there.  The subcommands can now be modified gradually to use it.

### DIFF
--- a/apps/openssl.c
+++ b/apps/openssl.c
@@ -287,18 +287,14 @@ int main(int argc, char *argv[])
     /* first check the program name */
     f.name = pname;
     fp = lh_FUNCTION_retrieve(prog, &f);
-    if (fp != NULL) {
-        argv[0] = pname;
-        if (fp->deprecated_alternative != NULL)
-            warn_deprecated(fp);
-        ret = fp->func(argc, argv);
-        goto end;
-    }
-
-    /* If there is stuff on the command line, run with that. */
-    if (argc != 1) {
+    if (fp == NULL) {
+        /* We assume we've been called as 'openssl cmd' */
         argc--;
         argv++;
+    }
+
+    /* If there's a command, run with that. */
+    if (argc > 0) {
         ret = do_cmd(prog, argc, argv);
         if (ret < 0)
             ret = 0;
@@ -400,7 +396,7 @@ const OPTIONS help_options[] = {
 };
 
 
-int help_main(int argc, char **argv)
+int help_main(OPENSSL_CTX *libctx, int argc, char **argv)
 {
     FUNCTION *fp;
     int i, nl;
@@ -485,9 +481,21 @@ static int do_cmd(LHASH_OF(FUNCTION) *prog, int argc, char *argv[])
         }
     }
     if (fp != NULL) {
+        int ret = 1;
+        OPENSSL_CTX *libctx = NULL;
+
         if (fp->deprecated_alternative != NULL)
             warn_deprecated(fp);
-        return fp->func(argc, argv);
+
+        if ((libctx = OPENSSL_CTX_new()) != NULL
+            && OPENSSL_CTX_load_config(libctx, NULL)) {
+            ret = fp->func(libctx, argc, argv);
+        } else {
+            BIO_printf(bio_err, "Error creating library context");
+            ERR_print_errors(bio_err);
+        }
+        OPENSSL_CTX_free(libctx);
+        return ret;
     }
     if ((strncmp(argv[0], "no-", 3)) == 0) {
         /*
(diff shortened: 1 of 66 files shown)


richsalz:
I'm sure this is going to cause yet more conflicts with #10632.  It sure would have been nice if the project looked at that PR.

levitte:
Just by eyeballing, I can't see any conflict with #10632...

apps/openssl.c
```diff
@@ -287,18 +287,14 @@ int main(int argc, char *argv[])
     /* first check the program name */
     f.name = pname;
     fp = lh_FUNCTION_retrieve(prog, &f);
-    if (fp != NULL) {
-        argv[0] = pname;
-        if (fp->deprecated_alternative != NULL)
```
slontis, line 292: Why has this been removed?

slontis:
Thanks for doing this.
It does seem backwards to me though.

So the -provider options currently work on the NULL libctx.. Is that the intention?

What you have supplied currently is the ability to add a libctx that gets loaded with either the  default config that can possibly be overriden via an environment variable.
So I guess the fips conf can be loaded with this.

Having to set and unset an environment variable for a series of test seems fairly horrible
(See #11884 80-test-cms)

I have put #11948 on hold also. Maybe this is a discussion point for Tuesday?

levitte:
Not sure I understand your concern, @slontis. The environment variable `OPENSSL_CONF` is checked by libcrypto as a default. However, there's nothing to stop any subcommand from loading another, some subcommands already do, with the option `-config`.

What I've done here is something we talked about a long time ago, that one suitable application of library contexts would be to have throw-away library context for each execution of a subcommand, especially important for interactive mode (you know, when you just run `openssl` with no arguments and get a `OpenSSL>` prompt).

levitte:
Tuesday discussion point sounds fine by me.

richsalz:
> However, there's nothing to stop any subcommand from loading another, some subcommands already do, with the option -config.

An *additional* (another) config, or a *different* config?

levitte:
> An *additional* (another) config, or a *different* config?

Additional.  This can be mitigated with `OPENSSL_CONF=/dev/null` if that's desirable.

richsalz:
That should be better documented.  openssl.pod?  Where do you think?

levitte:
> That should be better documented. openssl.pod? Where do you think?

In 1.1.1, that's exactly where it's documented, in the "COMMAND SUMMARY" section.  That documentation isn't very good at all, really.

I'm actually pondering if we should review the interaction between OPENSSL_CONF and -config more in depth.  I can't remember, but am going to guess that -config came first, as this was originally only used with configuration of the req, x509 and ca commands.  Then came engines, and we suddenly had the need for global, always present configuration, and I guess that's when OPENSSL_CONF came to be, so we could have the configuration loaded in *one* *place* instead of having to duplicate the same sample app code everywhere.

I believe that the *intention* is that OPENSSL_CONF is meant to act as a default file, and that -config can be used to give another configuration file to be used instead of the default.  That is, however, *not* how things are currently coded, because at least the engine and provider definitions are used immediately when loading the configuration file...  so when there's both a default configuration file and a -config given, both files are loaded. and the engine and provider definitions from each are added to libcrypto internals / library context, thereby being *additions*, and meanwhile, the settings for req, ca and x509 are *replacements*, read from the configuration file that came last.  It's a bit of a mess.

beldmit:
From my point of view, the OPENSSL_CONF/compiled-in config file name is the default config loaded somewhen at startup. As some algorithms can be enabled via the config-loaded engines/providers, it should be loaded as early as possible.

The supplementary configs, if I understand correctly, were designed for providing some non-permanent data (such as cert/req DNs), but, being parsed via the same mechanism, work both as additions and replacements. Such behavior potentially causes problems, e.g. engines can be initialized twice.

levitte:
I just looked over the git history, and found that I was wrong, at least as far as OpenSSL history goes.  We've inherited *both* `OPENSSL_CONF` (then called `SSLEAY_CONF`) and `-config` from SSLeay, and available history doesn't tell what came first.

However, at the time, there was nothing much "global" about the config file.  `OPENSSL_CONF` simply provided a default for the openssl commands to use, and it was up to each command to do whatever they wanted with it (including having a `-config` option that would provide a replacement file).  So in practice, `-config` replaced  the value of `OPENSSL_CONF`, which replaced the built in default.

That being said, ENGINEs changed all that, as they brought along global stuff that became configurable with the config files, and at that point, "replacement" became "replacement or addition", all depending on what part of the config file you look at.

levitte:
Closing this, because #12023

PR #11983, opened 2020-05-28
Misc apps fixes [2020-05-28]

levitte:
`make_config_name()` is a duplicate of `CONF_get1_default_config_file()`, so drop it.

`apps/progs.pl` has strange results when it's running in a national locale.  For example, `w` isn't matched by the regular expression `[a-z]` in the Swedish locale (which means `passwd_main` isn't matched, for example).  To avoid that sort of surprise, we set the "C" locale.

### DIFF
--- a/apps/openssl.c
+++ b/apps/openssl.c
@@ -222,7 +222,6 @@ int main(int argc, char *argv[])
     arg.size = 0;
 
     /* Set up some of the environment. */
-    default_config_file = CONF_get1_default_config_file();
     bio_in = dup_bio_in(FORMAT_TEXT);
     bio_out = dup_bio_out(FORMAT_TEXT);
     bio_err = dup_bio_err(FORMAT_TEXT);
@@ -262,6 +261,10 @@ int main(int argc, char *argv[])
         goto end;
     }
     pname = opt_progname(argv[0]);
+
+    default_config_file = CONF_get1_default_config_file();
+    if (default_config_file == NULL)
+        app_bail_out("%s: could not get default config file\n", pname);
 
     /* first check the program name */
     f.name = pname;
--- a/apps/openssl.c
+++ b/apps/openssl.c
@@ -80,27 +80,6 @@ static void apps_shutdown(void)
     destroy_ui_method();
 }
 
-static char *make_config_name(void)
-{
-    const char *t;
-    size_t len;
-    char *p;
-
-    if ((t = getenv("OPENSSL_CONF")) != NULL)
-        return OPENSSL_strdup(t);
-
-    t = X509_get_default_cert_area();
-    len = strlen(t) + 1 + strlen(OPENSSL_CONF) + 1;
-    p = app_malloc(len, "config filename buffer");
-    strcpy(p, t);
-#ifndef OPENSSL_SYS_VMS
-    strcat(p, "/");
-#endif
-    strcat(p, OPENSSL_CONF);
-
-    return p;
-}
-
 
 #ifndef OPENSSL_NO_TRACE
 typedef struct tracedata_st {
@@ -243,7 +222,7 @@ int main(int argc, char *argv[])
     arg.size = 0;
 
     /* Set up some of the environment. */
-    default_config_file = make_config_name();
+    default_config_file = CONF_get1_default_config_file();
     bio_in = dup_bio_in(FORMAT_TEXT);
     bio_out = dup_bio_out(FORMAT_TEXT);
     bio_err = dup_bio_err(FORMAT_TEXT);
--- a/crypto/conf/conf_mod.c
+++ b/crypto/conf/conf_mod.c
@@ -516,25 +516,23 @@ void CONF_module_set_usr_data(CONF_MODULE *pmod, void *usr_data)
 
 char *CONF_get1_default_config_file(void)
 {
+    const char *t;
     char *file, *sep = "";
-    int len;
+    int size;
 
     if ((file = ossl_safe_getenv("OPENSSL_CONF")) != NULL)
         return OPENSSL_strdup(file);
 
-    len = strlen(X509_get_default_cert_area());
+    t = X509_get_default_cert_area();
 #ifndef OPENSSL_SYS_VMS
-    len++;
     sep = "/";
 #endif
-    len += strlen(OPENSSL_CONF);
-
-    file = OPENSSL_malloc(len + 1);
+    size = strlen(t) + strlen(sep) + strlen(OPENSSL_CONF) + 1;
+    file = OPENSSL_malloc(size);
 
     if (file == NULL)
         return NULL;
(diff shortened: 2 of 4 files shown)


richsalz:
Explicitly running in the "C" locale seems like it will be bad for some users, e.g., @beldmit and @InfoHunter and their countrymen.  Are you sure that's the right thing to do, rather than add a NOTE (sic) in the documentation?

beldmit:
Explicit setting locale to "C" is often a way to find out an understandable (or at least googlable) diagnostic output in Russia :)

levitte:
In my case, *not* running under the "C" locale was bad. We are, after all, trying to match ASCII characters, so the "C" locale is the best way to ensure `[a-z]` matches them all.

apps/openssl.c
```diff
@@ -243,7 +222,7 @@ int main(int argc, char *argv[])
     arg.size = 0;
 
     /* Set up some of the environment. */
-    default_config_file = make_config_name();
+    default_config_file = CONF_get1_default_config_file();
```
bernd-edlinger, line 225: this change is problematic, while make_config_name uses app_malloc
was at least in the past guaranteed to return non-zero.
I see OPENSSL_strdup was added and breaks that contract.
CONF_get1_default_config_file does not kill the process when out of memory happens.
so just add a fatal error handling here?

levitte: Good point, will add that.

levitte:
I've dropped the locale change in apps/progs.pl, couldn't reproduce what went wrong when I tried again.

apps/openssl.c
levitte: Fixed

apps/openssl.c
```diff
@@ -243,7 +222,6 @@ int main(int argc, char *argv[])
     arg.size = 0;
 
     /* Set up some of the environment. */
-    default_config_file = make_config_name();
     bio_in = dup_bio_in(FORMAT_TEXT);
     bio_out = dup_bio_out(FORMAT_TEXT);
     bio_err = dup_bio_err(FORMAT_TEXT);
```
bernd-edlinger, line 227: also unclear if the return value of dup_bio_xxx should be checked likewise.

richsalz: or do what app_malloc does and fail/exit into dup_bio routine.

crypto/conf/conf_mod.c
```diff
@@ -516,25 +516,23 @@ void CONF_module_set_usr_data(CONF_MODULE *pmod, void *usr_data)
 
 char *CONF_get1_default_config_file(void)
 {
+    const char *t;
     char *file, *sep = "";
-    int len;
+    int size;
```
bernd-edlinger, line 521: should be size_t

levitte: Thanks.  Fixed

apps/openssl.c
bernd-edlinger: Note there are already calls in apps/lib/apps.c
where the return value is checked.

richsalz: Something like this, and remove the one :) place it's checked:
```
diff --git a/apps/lib/apps.c b/apps/lib/apps.c
index dbc9be344d..38ff5adfe3 100644
--- a/apps/lib/apps.c
+++ b/apps/lib/apps.c
@@ -2363,6 +2363,17 @@ int raw_write_stdout(const void *buf, int siz)
 }
 #endif
 
+static BIO *bio_check(BIO *bp, const char *what, int format)
+{
+    if (bp == NULL) {
+        BIO_printf(bio_err, "%s: Could not duplicate %s (mode 0x%x)\n",
+                opt_getprog(), what, format);
+        ERR_print_errors(bio_err);
+        exit(1);
+    }
+    return bp;
+}
+
 /*
  * Centralized handling of input and output files with format specification
  * The format is meant to show what the input and output is supposed to be,
@@ -2372,14 +2383,15 @@ int raw_write_stdout(const void *buf, int siz)
  */
 BIO *dup_bio_in(int format)
 {
-    return BIO_new_fp(stdin,
-                      BIO_NOCLOSE | (FMT_istext(format) ? BIO_FP_TEXT : 0));
+    int mode = BIO_NOCLOSE | (FMT_istext(format) ? BIO_FP_TEXT : 0); 
+
+    return bio_check(BIO_new_fp(stdin, mode), "stdin", mode);
 }
 
 BIO *dup_bio_out(int format)
 {
-    BIO *b = BIO_new_fp(stdout,
-                        BIO_NOCLOSE | (FMT_istext(format) ? BIO_FP_TEXT : 0));
+    int mode = BIO_NOCLOSE | (FMT_istext(format) ? BIO_FP_TEXT : 0);
+    BIO *b = BIO_new_fp(stdout, mode);
     void *prefix = NULL;
 
 #ifdef OPENSSL_SYS_VMS
@@ -2393,7 +2405,7 @@ BIO *dup_bio_out(int format)
         BIO_set_prefix(b, prefix);
     }
 
-    return b;
+    return bio_check(b, "stdout", mode);
 }
 
 BIO *dup_bio_err(int format)
```

Fixing ```dup_bio_err``` is left as an exercise for the reader :)

bernd-edlinger: @richsalz This would render the errorhandling in app_get_pass pointless:
```C
        } else if (strcmp(arg, "stdin") == 0) {
            pwdbio = dup_bio_in(FORMAT_TEXT);
            if (pwdbio == NULL) {
                BIO_printf(bio_err, "Can't open BIO for stdin\n");
                return NULL;
            }
```

bernd-edlinger: to keep it simple, I would just prefer
```C
if (bio_in == NULL || bio_out == NULL || bio_err == NULL)
  app_bail_out(...);
```

richsalz: yes it would make that pointless, shrug.  it would also fix engine.c, two cases in each of s_client an s_server. And then look at the callers of bio_open_default.  Seems worth it to me.

bernd-edlinger: good point.

bernd-edlinger: @levitte if this still cherry-picks to 1.1.1 I'd approve it now,
and let that for another PR.

levitte: Other PR sounds good to me

richsalz: If I do it, will it get reviewed?  Or are you going to do it.

bernd-edlinger: Please do it.  I will review.

levitte: Oh do feel free!

t8m:
Approved 5 days ago by Bernd already.

levitte:
Merged

e306f83c8cfc7ac970d04a36c90634ab8573a594 APPS: Remove make_config_name, use CONF_get1_default_config_file instead

PR #11986, opened 2020-05-28
Add range check for 'fd' argument of BIO_socket_wait()

DDvO:
as discussed in https://github.com/openssl/openssl/pull/11971#discussion_r431817997.

### DIFF
--- a/crypto/bio/b_sock.c
+++ b/crypto/bio/b_sock.c
@@ -387,6 +387,8 @@ int BIO_socket_wait(int fd, int for_read, time_t max_time)
     struct timeval tv;
     time_t now;
 
+    if (fd < 0 || fd >= FD_SETSIZE)
+        return -1;
     if (max_time == 0)
         return 1;
 
--- a/crypto/bio/bio_lib.c
+++ b/crypto/bio/bio_lib.c
@@ -786,41 +786,49 @@ void bio_cleanup(void)
 }
 
 /* Internal variant of the below BIO_wait() not calling BIOerr() */
-static int bio_wait(BIO *bio, time_t max_time, unsigned int milliseconds)
+static int bio_wait(BIO *bio, time_t max_time, unsigned int nap_milliseconds)
 {
 #ifndef OPENSSL_NO_SOCK
     int fd;
 #endif
+    long sec_diff;
 
-    if (max_time == 0)
+    if (max_time == 0) /* no timeout */
         return 1;
 
 #ifndef OPENSSL_NO_SOCK
-    if (BIO_get_fd(bio, &fd) > 0)
+    if (BIO_get_fd(bio, &fd) > 0 && fd < FD_SETSIZE)
         return BIO_socket_wait(fd, BIO_should_read(bio), max_time);
 #endif
-    if (milliseconds > 1000) {
-        long sec_diff = (long)(max_time - time(NULL)); /* might overflow */
+    /* fall back to polling since no sockets are available */
 
-        if (sec_diff <= 0)
-            return 0; /* timeout */
-        if ((unsigned long)sec_diff < milliseconds / 1000)
-            milliseconds = (unsigned long)sec_diff * 1000;
+    sec_diff = (long)(max_time - time(NULL)); /* might overflow */
+    if (sec_diff < 0)
+        return 0; /* clearly timeout */
+
+    /* now take a nap at most the given number of milliseconds */
+    if (sec_diff == 0) { /* we are below the 1 seconds resolution of max_time */
+        if (nap_milliseconds > 1000)
+            nap_milliseconds = 1000;
+    } else { /* for sec_diff > 0, take min(sec_diff * 1000, nap_milliseconds) */
+        if ((unsigned long)sec_diff * 1000 < nap_milliseconds)
+            nap_milliseconds = (unsigned int)sec_diff * 1000;
     }
-    ossl_sleep(milliseconds);
+    ossl_sleep(nap_milliseconds);
     return 1;
 }
 
-/*
+/*-
  * Wait on (typically socket-based) BIO at most until max_time.
- * Succeed immediately if max_time == 0. If sockets are not available succeed
- * after waiting at most given milliseconds in order to avoid a tight busy loop.
- * Call BIOerr(...) unless success.
+ * Succeed immediately if max_time == 0.
+ * If sockets are not available support polling: succeed after waiting at most
+ * the number of nap_milliseconds in order to avoid a tight busy loop.
+ * Call BIOerr(...) on timeout or error.
  * Returns -1 on error, 0 on timeout, and 1 on success.
  */
-int BIO_wait(BIO *bio, time_t max_time, unsigned int milliseconds)
+int BIO_wait(BIO *bio, time_t max_time, unsigned int nap_milliseconds)
 {
-    int rv = bio_wait(bio, max_time, milliseconds);
+    int rv = bio_wait(bio, max_time, nap_milliseconds);
 
     if (rv <= 0)
         BIOerr(0, rv == 0 ? BIO_R_TRANSFER_TIMEOUT : BIO_R_TRANSFER_ERROR);
@@ -828,13 +836,16 @@ int BIO_wait(BIO *bio, time_t max_time, unsigned int milliseconds)
 }
 
 /*
- * Connect via given BIO using BIO_do_handshake() until success/timeout/error.
- * Parameter timeout == 0 means infinite, < 0 leads to immediate timeout error.
+ * Connect via given BIO using BIO_do_connect() until success/timeout/error.
+ * Parameter timeout == 0 means no timeout, < 0 means exactly one try.
+ * For non-blocking and potentially even non-socket BIOs perform polling with
+ * the given density: between polls sleep nap_milliseconds using BIO_wait()
+ * in order to avoid a tight busy loop.
  * Returns -1 on error, 0 on timeout, and 1 on success.
  */
-int BIO_connect_retry(BIO *bio, int timeout)
+int BIO_do_connect_retry(BIO *bio, int timeout, int nap_milliseconds)
 {
-    int blocking = timeout == 0;
+    int blocking = timeout <= 0;
     time_t max_time = timeout > 0 ? time(NULL) + timeout : 0;
     int rv;
 
@@ -843,40 +854,47 @@ int BIO_connect_retry(BIO *bio, int timeout)
         return -1;
     }
 
-    if (timeout < 0) {
-        BIOerr(0, BIO_R_CONNECT_TIMEOUT);
-        return 0;
-    }
+    if (nap_milliseconds < 0)
+        nap_milliseconds = 100;
+    BIO_set_nbio(bio, !blocking);
 
-    if (!blocking)
-        BIO_set_nbio(bio, 1);
+ retry:
+    rv = BIO_do_connect(bio); /* This may indirectly call ERR_clear_error(); */
 
- retry: /* it does not help here to set SSL_MODE_AUTO_RETRY */
-    rv = BIO_do_handshake(bio); /* This indirectly calls ERR_clear_error(); */
+    if (rv <= 0) { /* could be timeout or retryable error or fatal error */
+        int err = ERR_peek_last_error();
+        int reason = ERR_GET_REASON(err);
+        int do_retry = BIO_should_retry(bio); /* may be 1 only if !blocking */
 
-    if (rv <= 0) {
-        if (get_last_sys_error() == ETIMEDOUT) {
-            /*
-             * if blocking, despite blocking BIO, BIO_do_handshake() timed out
-             * when non-blocking, BIO_do_handshake() timed out early
-             * with rv == -1 and get_last_sys_error() == 0
-             */
-            ERR_clear_error();
-            (void)BIO_reset(bio);
-            /*
-             * unless using BIO_reset(), blocking next connect() may crash and
-             * non-blocking next BIO_do_handshake() will fail
-             */
-            goto retry;
-        } else if (BIO_should_retry(bio)) {
-            /* will not actually wait if timeout == 0 (i.e., blocking BIO) */
-            rv = bio_wait(bio, max_time, 100 /* milliseconds */);
+        if (ERR_GET_LIB(err) == ERR_LIB_BIO) {
+            switch (reason) {
+            case ERR_R_SYS_LIB:
+                /*
+                 * likely retryable system error occurred, which may be
+                 * EAGAIN (resource temporarily unavailable) some 40 secs after
+                 * calling getaddrinfo(): Temporary failure in name resolution
+                 * or a premature ETIMEDOUT, some 30 seconds after connect()
+                 */
+            case BIO_R_CONNECT_ERROR:
+            case BIO_R_NBIO_CONNECT_ERROR:
+                /* some likely retryable connection error occurred */
+                (void)BIO_reset(bio); /* often needed to avoid retry failure */
+                do_retry = 1;
+                break;
+            default:
+                break;
+            }
+        }
+        if (timeout >= 0 && do_retry) {
+            ERR_clear_error(); /* using ERR_pop_to_mark() would be cleaner */
+            /* will not actually wait if timeout == 0 (i.e., blocking BIO): */
+            rv = bio_wait(bio, max_time, nap_milliseconds);
             if (rv > 0)
                 goto retry;
             BIOerr(0, rv == 0 ? BIO_R_CONNECT_TIMEOUT : BIO_R_CONNECT_ERROR);
         } else {
             rv = -1;
-            if (ERR_peek_error() == 0) /* missing error queue entry */
+            if (err == 0) /* missing error queue entry */
                 BIOerr(0, BIO_R_CONNECT_ERROR); /* workaround: general error */
         }
     }
--- a/crypto/bio/bss_conn.c
+++ b/crypto/bio/bss_conn.c
@@ -438,12 +438,13 @@ static long conn_ctrl(BIO *b, int cmd, long num, void *ptr)
     case BIO_C_SET_CONNECT:
         if (ptr != NULL) {
             b->init = 1;
-            if (num == 0) {
+            if (num == 0) { /* BIO_set_conn_hostname */
                 char *hold_service = data->param_service;
                 /* We affect the hostname regardless.  However, the input
                  * string might contain a host:service spec, so we must
                  * parse it, which might or might not affect the service
                  */
+
                 OPENSSL_free(data->param_hostname);
                 data->param_hostname = NULL;
                 ret = BIO_parse_hostserv(ptr,
@@ -452,19 +453,29 @@ static long conn_ctrl(BIO *b, int cmd, long num, void *ptr)
                                          BIO_PARSE_PRIO_HOST);
                 if (hold_service != data->param_service)
                     OPENSSL_free(hold_service);
-            } else if (num == 1) {
+            } else if (num == 1) { /* BIO_set_conn_port */
                 OPENSSL_free(data->param_service);
-                data->param_service = OPENSSL_strdup(ptr);
-            } else if (num == 2) {
+                if ((data->param_service = OPENSSL_strdup(ptr)) == NULL)
+                    ret = 0;
+            } else if (num == 2) { /* BIO_set_conn_address */
                 const BIO_ADDR *addr = (const BIO_ADDR *)ptr;
+                char *host = BIO_ADDR_hostname_string(addr, 1);
+                char *service = BIO_ADDR_service_string(addr, 1);
+
+                ret = host != NULL && service != NULL;
                 if (ret) {
-                    data->param_hostname = BIO_ADDR_hostname_string(addr, 1);
-                    data->param_service = BIO_ADDR_service_string(addr, 1);
+                    OPENSSL_free(data->param_hostname);
+                    data->param_hostname = host;
+                    OPENSSL_free(data->param_service);
+                    data->param_service = service;
                     BIO_ADDRINFO_free(data->addr_first);
                     data->addr_first = NULL;
                     data->addr_iter = NULL;
+                } else {
+                    OPENSSL_free(host);
+                    OPENSSL_free(service);
                 }
-            } else if (num == 3) {
+            } else if (num == 3) { /* BIO_set_conn_ip_family */
                 data->connect_family = *(int *)ptr;
             } else {
                 ret = 0;
--- a/crypto/bio/bss_acpt.c
+++ b/crypto/bio/bss_acpt.c
@@ -433,8 +433,10 @@ static long acpt_ctrl(BIO *b, int cmd, long num, void *ptr)
                 b->init = 1;
             } else if (num == 1) {
                 OPENSSL_free(data->param_serv);
-                data->param_serv = OPENSSL_strdup(ptr);
-                b->init = 1;
+                if ((data->param_serv = OPENSSL_strdup(ptr)) == NULL)
+                    ret = 0;
+                else
+                    b->init = 1;
             } else if (num == 2) {
                 data->bind_mode |= BIO_SOCK_NONBLOCK;
             } else if (num == 3) {
--- a/crypto/http/http_client.c
+++ b/crypto/http/http_client.c
@@ -814,7 +814,7 @@ static int update_timeout(int timeout, time_t start_time)
  *   BIO *(*OSSL_HTTP_bio_cb_t) (BIO *bio, void *arg, int conn, int detail);
  * The callback may modify the HTTP BIO provided in the bio argument,
  * whereby it may make use of any custom defined argument 'arg'.
- * During connection establishment, just after BIO_connect_retry(),
+ * During connection establishment, just after BIO_do_connect_retry(),
  * the callback function is invoked with the 'conn' argument being 1
  * 'detail' indicating whether a HTTPS (i.e., TLS) connection is requested.
  * On disconnect 'conn' is 0 and 'detail' indicates that no error occurred.
@@ -873,7 +873,7 @@ BIO *OSSL_HTTP_transfer(const char *server, const char *port, const char *path,
     /* remaining parameters are checked indirectly by the functions called */
 
     (void)ERR_set_mark(); /* prepare removing any spurious libssl errors */
-    if (rbio == NULL && BIO_connect_retry(cbio, timeout) <= 0)
+    if (rbio == NULL && BIO_do_connect_retry(cbio, timeout, -1) <= 0)
         goto end;
     /* now timeout is guaranteed to be >= 0 */
 
--- a/doc/man3/BIO_socket_wait.pod
+++ b/doc/man3/BIO_socket_wait.pod
@@ -4,8 +4,8 @@
 
 BIO_socket_wait,
 BIO_wait,
-BIO_connect_retry
-- BIO socket utility functions
+BIO_do_connect_retry
+- BIO connection utility functions
 
 =head1 SYNOPSIS
 
@@ -14,8 +14,8 @@ BIO_connect_retry
  #ifndef OPENSSL_NO_SOCK
  int BIO_socket_wait(int fd, int for_read, time_t max_time);
  #endif
- int BIO_wait(BIO *bio, time_t max_time, unsigned int milliseconds);
- int BIO_connect_retry(BIO *bio, long timeout);
+ int BIO_wait(BIO *bio, time_t max_time, unsigned int nap_milliseconds);
+ int BIO_do_connect_retry(BIO *bio, int timeout, int nap_milliseconds);
 
 =head1 DESCRIPTION
 
@@ -23,27 +23,38 @@ BIO_socket_wait() waits on the socket B<fd> for reading if B<for_read> is not 0,
 else for writing, at most until B<max_time>.
 It succeeds immediately if B<max_time> == 0 (which means no timeout given).
 
-BIO_wait() waits at most until B<max_time> on the given B<bio>,
-which is typically socket-based,
-for reading if B<bio> is supposed to read, else for writing.
+BIO_wait() waits at most until B<max_time> on the given (typically socket-based)
+B<bio>, for reading if B<bio> is supposed to read, else for writing.
+It is used by BIO_do_connect_retry() and can be used together L<BIO_read(3)>.
 It succeeds immediately if B<max_time> == 0 (which means no timeout given).
-If sockets are not available it succeeds after waiting at most given
-B<milliseconds> in order to help avoiding a tight busy loop at the caller.
+If sockets are not available it supports polling by succeeding after sleeping
+at most the given B<nap_milliseconds> in order to avoid a tight busy loop.
+Via B<nap_milliseconds> the caller determines the polling granularity.
 
-BIO_connect_retry() connects via the given B<bio>, retrying BIO_do_connect()
-until success or a timeout or error condition is reached.
+BIO_do_connect_retry() connects via the given B<bio>.
+It retries BIO_do_connect() as far as needed to reach a definite outcome,
+i.e., connection succeeded, timeout has been reached, or an error occurred.
+For non-blocking and potentially even non-socket BIOs it polls
+every B<nap_milliseconds> and sleeps in between using BIO_wait().
+If B<nap_milliseconds> is < 0 then a default value of 100 ms is used.
 If the B<timeout> parameter is > 0 this indicates the maximum number of seconds
-to wait until the connection is established. A value of 0 enables waiting
-indefinitely, while a value < 0 immediately leads to a timeout condition.
+to wait until the connection is established or a definite error occurred.
+A value of 0 enables waiting indefinitely (i.e, no timeout),
+while a value < 0 means that BIO_do_connect() is tried only once.
+The function may, directly or indirectly, invoke ERR_clear_error().
 
 =head1 RETURN VALUES
 
-BIO_socket_wait(), BIO_wait(), and BIO_connect_retry()
+BIO_socket_wait(), BIO_wait(), and BIO_do_connect_retry()
 return -1 on error, 0 on timeout, and 1 on success.
+
+=head1 SEE ALSO
+
+L<BIO_do_connect(3)>, L<BIO_read(3)>
 
 =head1 HISTORY
 
-BIO_socket_wait(), BIO_wait(), and BIO_connect_retry()
+BIO_socket_wait(), BIO_wait(), and BIO_do_connect_retry()
 were added in OpenSSL 3.0.
 
 =head1 COPYRIGHT
--- a/doc/man3/OSSL_CMP_CTX_new.pod
+++ b/doc/man3/OSSL_CMP_CTX_new.pod
@@ -337,7 +337,7 @@ function, which has the prototype
 The callback may modify the BIO B<bio> provided by OSSL_CMP_MSG_http_perform(),
 whereby it may make use of a custom defined argument B<ctx>
 stored in the OSSL_CMP_CTX by means of OSSL_CMP_CTX_set_http_cb_arg().
-During connection establishment, just after calling BIO_connect_retry(),
+During connection establishment, just after calling BIO_do_connect_retry(),
 the function is invoked with the B<connect> argument being 1 and the B<detail>
 argument being 1 if HTTPS is requested, i.e., SSL/TLS should be enabled. On
 disconnect B<connect> is 0 and B<detail> is 1 in case no error occurred, else 0.
--- a/doc/man3/OSSL_HTTP_transfer.pod
+++ b/doc/man3/OSSL_HTTP_transfer.pod
@@ -154,7 +154,7 @@ B<bio_update_fn> is a BIO connect/disconnect callback function with prototype
 The callback may modify the HTTP BIO provided in the B<bio> argument,
 whereby it may make use of a custom defined argument B<arg>,
 which may for instance refer to an I<SSL_CTX> structure.
-During connection establishment, just after calling BIO_connect_retry(),
+During connection establishment, just after calling BIO_do_connect_retry(),
 the function is invoked with the B<connect> argument being 1 and the B<detail>
 argument being 1 if HTTPS is requested, i.e., SSL/TLS should be enabled.
 On disconnect B<connect> is 0 and B<detail> is 1 if no error occurred, else 0.
--- a/include/openssl/bio.h
+++ b/include/openssl/bio.h
@@ -664,8 +664,8 @@ int BIO_sock_should_retry(int i);
 int BIO_sock_non_fatal_error(int error);
 int BIO_socket_wait(int fd, int for_read, time_t max_time);
 # endif
-int BIO_wait(BIO *bio, time_t max_time, unsigned int milliseconds);
-int BIO_connect_retry(BIO *bio, int timeout);
+int BIO_wait(BIO *bio, time_t max_time, unsigned int nap_milliseconds);
+int BIO_do_connect_retry(BIO *bio, int timeout, int nap_milliseconds);
 
 int BIO_fd_should_retry(int i);
 int BIO_fd_non_fatal_error(int error);
--- a/util/libcrypto.num
+++ b/util/libcrypto.num
@@ -4927,7 +4927,7 @@ RAND_DRBG_set_callback_data             ?	3_0_0	EXIST::FUNCTION:
 RAND_DRBG_get_callback_data             ?	3_0_0	EXIST::FUNCTION:
 BIO_socket_wait                         ?	3_0_0	EXIST::FUNCTION:SOCK
 BIO_wait                                ?	3_0_0	EXIST::FUNCTION:
-BIO_connect_retry                       ?	3_0_0	EXIST::FUNCTION:
+BIO_do_connect_retry                    ?	3_0_0	EXIST::FUNCTION:
 ERR_load_HTTP_strings                   ?	3_0_0	EXIST::FUNCTION:
 OSSL_HTTP_get                           ?	3_0_0	EXIST::FUNCTION:
 OSSL_HTTP_get_asn1                      ?	3_0_0	EXIST::FUNCTION:


crypto/bio/b_sock.c
```diff
@@ -387,6 +387,8 @@ int BIO_socket_wait(int fd, int for_read, time_t max_time)
     struct timeval tv;
     time_t now;
 
+    if (fd < 0 || fd >= FD_SETSIZE)
```
bernd-edlinger, line 390: Okay, but I'd like to see the `fd >= FD_SETSIZE` also in bio_wait, and defer to the default handling
in that case.

DDvO: Good point - done.

On this occasion I've also renamed the `millisecondss` param of `BiO_wait()` to `fallback_milliseconds` for clarity.

levitte:
This looks good to me, but I'll defer to @bernd-edlinger to actually approve

crypto/bio/bio_lib.c
```diff
@@ -796,31 +796,32 @@ static int bio_wait(BIO *bio, time_t max_time, unsigned int milliseconds)
         return 1;
```
bernd-edlinger, line 796: I wonder why the condition above this line `if (max_time == 0) `
makes sense, and why the return value is not 0 (timeout) in this case.

crypto/bio/b_sock.c
```diff
@@ -387,6 +387,8 @@ int BIO_socket_wait(int fd, int for_read, time_t max_time)
...
+        return -1;
     if (max_time == 0)
         return 1;
```
bernd-edlinger, line 393: likewise here,  what is so special with max_time==0 that this
returns 1, which means socket is readable?

crypto/bio/bio_lib.c
```diff
@@ -796,31 +796,32 @@ static int bio_wait(BIO *bio, time_t max_time, unsigned int milliseconds)
...
-    if (milliseconds > 1000) {
+    if (fallback_millisec > 1000) {
         long sec_diff = (long)(max_time - time(NULL)); /* might overflow */
 
         if (sec_diff <= 0)
             return 0; /* timeout */
-        if ((unsigned long)sec_diff < milliseconds / 1000)
-            milliseconds = (unsigned long)sec_diff * 1000;
+        if ((unsigned long)sec_diff < fallback_millisec / 1000)
+            fallback_millisec = (unsigned long)sec_diff * 1000;
```
bernd-edlinger, line 808: why are you casting to unsigned long, when fallback_millisec is unsigned int ?

crypto/bio/bio_lib.c
```diff
@@ -796,31 +796,32 @@ static int bio_wait(BIO *bio, time_t max_time, unsigned int milliseconds)
         return 1;
 
 #ifndef OPENSSL_NO_SOCK
-    if (BIO_get_fd(bio, &fd) > 0)
+    if (BIO_get_fd(bio, &fd) > 0 && fd < FD_SETSIZE)
         return BIO_socket_wait(fd, BIO_should_read(bio), max_time);
 #endif
-    if (milliseconds > 1000) {
+    if (fallback_millisec > 1000) {
```
bernd-edlinger, line 802: Strictly speaking this if (fallback_millisec > 1000) looks wrong.
Why is the due time not considered when fallback millisec is <= 1000 ?

crypto/bio/bio_lib.c
DDvO: As written in the documentation of `BIO_socket_wait()`, `BIO_wait()`, and `BIO_connect_retry()`, the meaning of `max_time == 0` is that no timeout is wanted (i.e., blocking behavior has been selected).
This, as well as the 'success' return in this case, is for convenient use for instance by `BIO_connect_retry()` and the HTTP client, where `max_time` has the same semantics.

crypto/bio/b_sock.c
DDvO: See above.
As written in the code comments of `BIO_connect_retry()`, there are some strange situations where also in the blocking case `BIO_do_connect()` returns with a retry condition, and for uniformity with the non-blocking case (with non-zero timeout/max_time) it is convenient to return with 1 (success) of `max_time == 0`.

crypto/bio/bio_lib.c
DDvO: This is a leftover from an earlier version where the type of `(fallback_)millisec` was `long` and not needed any more. Removed.

crypto/bio/b_sock.c
bernd-edlinger: I do not understand what strange condition that might be.
BUT I think I don't like libcrypto calling into libssl (BIO_do_handshake)

bernd-edlinger: sorry mixed it up with SSL_do_handshake :-)

crypto/bio/bio_lib.c
DDvO: Here we are in the fallback case where proper waiting until max_time or input is available is not possible.
The idea is that we need to poll but not too tightly (to prevent a busy loop).
So we wait typically for the given number of milliseconds (such that the granularity of the polling can be determined by the caller) but at most for one second, which is the resolution of `max_time`.
This is also (to some extent) mentioned in the doc:
```
If sockets are not available it succeeds after waiting at most given
B<fallback_milliseconds> in order to avoid a tight busy loop at the caller.
```

bernd-edlinger: but the time out check is in the if bloc,
so it will never happen.

bernd-edlinger: how I see this you wait for a small time,
try to read, that returns EAGAIN, wait more
in a loop, that while loop has a time out.
But Why is max_time only compared against time()
when the "small time" is > 1000 ???

bernd-edlinger: And, sorry, for what is this function useful when OPENSSL_NO_SOCK is defined?

DDvO: Good point - there was a bug.
And my above description was not right regarding waiting at most one second.
I've corrected the code and extended the commenting inside `bio_wait()` and the doc of `BIO_wait()`.
Hope things are clear and correct now.

DDvO: > And, sorry, for what is this function useful when OPENSSL_NO_SOCK is defined?

As I wrote, the function is designed to be useful also in the case no sockets are available, where one needs to resort to polling. The caller can decide via `fallback_milliseconds` how tight the busy loop shall be.

crypto/bio/bio_lib.c
```diff
@@ -786,41 +786,48 @@ void bio_cleanup(void)
...
-            milliseconds = (unsigned long)sec_diff * 1000;
+    sec_diff = (long)(max_time - time(NULL)); /* might overflow */
+    if (sec_diff < 0)
+        return 0; /* clearly timeout */
+
+    /* now wait at most the given number of milliseconds */
+    if (fallback_millisec > 1000) {
+
+        if (sec_diff < fallback_millisec / 1000)
+            fallback_millisec = sec_diff * 1000;
```
bernd-edlinger, line 813: I did not mean no typecast at all.
I just wanted to say that I would be more appropriate to do
```C
if (sec_diff < fallback_millisec / 1000)
    fallback_millisec = (unsigned int) sec_diff * 1000;
```

crypto/bio/bio_lib.c
bernd-edlinger: as a user I would expect to exclude socket-related code when I configure with no-sock
WTF is the use case, where you would use this function instead of a simple usleep,
when it ignores the passed in bio ?

t8m:
The AppVeyor failure is relevant.

crypto/bio/bio_lib.c
bernd-edlinger: whee, so make that
```C
if ((unsigned long)sec_diff < fallback_millisec / 1000)
    fallback_millisec = (unsigned int) sec_diff * 1000;
```

DDvO: Done

crypto/bio/bio_lib.c
```diff
@@ -856,21 +863,21 @@ int BIO_connect_retry(BIO *bio, int timeout)
 
     if (rv <= 0) {
         if (get_last_sys_error() == ETIMEDOUT) {
-            /*
-             * if blocking, despite blocking BIO, BIO_do_handshake() timed out
-             * when non-blocking, BIO_do_handshake() timed out early
-             * with rv == -1 and get_last_sys_error() == 0
+            /*-
+             * If blocking, BIO_do_handshake() timed out despite blocking BIO.
```
bernd-edlinger, line 867: This is probably a misunderstanding how a blocking connect works.
If the server or a firewall does not send RST packets the connect
will time-out after 1-2 minutes.  So a connection error may be
immediately due to a RST packet, or after a timeout due to no SYN-ACK packet.
In both cases the next address in this list should be tried,
I wonder if the BIO_reset resets that address.

crypto/bio/bio_lib.c
```diff
@@ -856,21 +863,21 @@ int BIO_connect_retry(BIO *bio, int timeout)
...
+             * If non-blocking, BIO_do_handshake() timed out early with
+             *                  rv == -1 and get_last_sys_error() == 0.
              */
             ERR_clear_error();
             (void)BIO_reset(bio);
             /*
-             * unless using BIO_reset(), blocking next connect() may crash and
-             * non-blocking next BIO_do_handshake() will fail
+             * Unless using BIO_reset(), blocking next connect() may crash and
+             * non-blocking next BIO_do_handshake() would fail.
```
bernd-edlinger, line 875: I am a bit unsure if that is a symptom of a deeper problem.
Do you have some details what "crash" may have happened here?

crypto/bio/bio_lib.c
```diff
@@ -856,21 +863,21 @@ int BIO_connect_retry(BIO *bio, int timeout)
```
bernd-edlinger, line 863: I wonder if I look at this code above
```C
    if (!blocking)
        BIO_set_nbio(bio, 1);
```

is that garanteed that the bio is always in blocking mode so
wouldn't it be safer to call `BIO_set_nbio(bio, !blocking);` here ?

crypto/bio/bio_lib.c
DDvO: I've meanwhile extended the documentation and comments of both `BIO_wait()` and `BIO_connect_retry()`, which should make their inter-working and use more clear.

Note that not all BIOs are socket-based, and even blocking BIOs sometimes return early (for unclear reasons) and ask for retry. For the user all that should not matter.
For non-blocking behavior a timeout value may need to be respected.

Therefore `BIO_connect_retry()` uses `BIO_wait()` to implement this transparently, for any type of BIO.
For non-blocking and potentially even non-socket BIOs it polls every 100 milliseconds and sleeps in between, avoiding a busy loop.

Moreover, the pretty convenient `BIO_wait()` is used also in `crypto/http/http_client.c` to transparently wait for new input in between `BIO_read()` or `BIO_gets()`.

bernd-edlinger: Hmm, but the BIO_connect_retry is only for sockets.

crypto/bio/bio_lib.c
```diff
@@ -856,21 +863,21 @@ int BIO_connect_retry(BIO *bio, int timeout)
...
-             * unless using BIO_reset(), blocking next connect() may crash and
-             * non-blocking next BIO_do_handshake() will fail
+             * Unless using BIO_reset(), blocking next connect() may crash and
+             * non-blocking next BIO_do_handshake() would fail.
              */
             goto retry;
         } else if (BIO_should_retry(bio)) {
             /* will not actually wait if timeout == 0 (i.e., blocking BIO) */
-            rv = bio_wait(bio, max_time, 100 /* milliseconds */);
+            rv = bio_wait(bio, max_time, 100 /* milliseconds to sleep */);
```
bernd-edlinger, line 880: I think you mentioned that this is sometimes called for non-blocking bio,
is that right?

crypto/bio/bio_lib.c
DDvO: It originally was meant for use with sockets.
Yet there might be also BIOs not involving sockets (well, of course not those implemented by `bss_conn.c`).
And as mentioned the polling is needed for sure also for non-blocking socket-based BIOs.

bernd-edlinger: Can you give more details for when we needed polling for non-blocking socket BIOs?

crypto/bio/bio_lib.c
DDvO: Good question. I had inferred that from programming examples.
Looking at the documentation of `BIO_set_nbio()` confirms this:
```
Blocking I/O is the default.
```

crypto/bio/bio_lib.c
DDvO: Yes, this is meant for non-blocking BIOs.
As written in the comment it `bio_wait()` has no effect in the blocking case (timeout == 0).

crypto/bio/bio_lib.c
DDvO: Thanks for that explanation.
AFAICS from `bio_conn.c` BIO_reset() resets the list of addresses to try, so does too much, right?

crypto/bio/bio_lib.c
bernd-edlinger: Yes, unless the BIO is initialized otherwise, and when you say,
the blocking sockets return with BIO_should_retry that makes me
curious, what triggers this behavior.

crypto/bio/bio_lib.c
DDvO: Unfortunately I don't recall the details.
It was about two years back when I had trouble implementing a stable HTTP connect
and asked the OpenSSL gurus for help (as far as I remember on the mailing list).
As I did not receive it I experimented until I found a working solution, as given here.

crypto/bio/bio_lib.c
bernd-edlinger: Could be possible

crypto/bio/bio_lib.c
bernd-edlinger: I wouldn't block this PR for it, as it is just a range check, but this
explanation given in this comment raises more questions than it answers.
So probably need more investigations.

crypto/bio/bio_lib.c
DDvO: Connection attempts for non-blocking (socket-based or not) BIOs typically return rather quickly. If retry is requested this would lead to a busy loop re-doing the retry immediately over and over.
I am sure about this because before I implemented the wait/sleep mechanism my CPU load went to 100% until the request max time was reached (timeout).

crypto/bio/bio_lib.c
```diff
@@ -856,21 +863,21 @@ int BIO_connect_retry(BIO *bio, int timeout)
 
     if (rv <= 0) {
         if (get_last_sys_error() == ETIMEDOUT) {
```
bernd-edlinger, line 865: I would be surprised if that does not reset the errno,
the last call `i = BIO_sock_error(b->num);`
is not an error, so should not *change* errno,
in the end I believe using errno in this way uses undefined data?

crypto/bio/bio_lib.c
DDvO: So I've just added for now the following comment:
```
            /*
             * TODO BIO_reset() maybe needs to be replaced since it likely
             * does too much, reseting the list of IP addresses to try
             */
```

crypto/bio/bio_lib.c
bernd-edlinger: you only tried localhost ?
for connections involving possibly non-reachable network addresses
things are much different.

crypto/bio/bio_lib.c
DDvO: Would be great if you anyone else with more network programming experience than me could investigate into that. Shall we open an issue?

bernd-edlinger: If you can setup something reproducible, yes, why not.
Currently it looks foggy to me.

crypto/bio/bio_lib.c
DDvO: Originally I had used `errno` but due to code review I replaced this by `get_last_sys_error()`. 
According to `e_os.h` it (usually) does not reset the errno:
```
# define get_last_sys_error()    errno
```

crypto/bio/bio_lib.c
DDvO: I agree, as for localhost it usually does not help much waiting for connection establishment.
I definitely tried for non-local severs and remember unplugging and plugging my laptop's ethernet cable (with WiFi disable) for trying this out :smile:

crypto/bio/bio_lib.c
DDvO: All right, then I won't open an issue now because I currently cannot reproduce the crash,
and I think the newly added TODO should be sufficient for now.

crypto/bio/bio_lib.c
bernd-edlinger: Yes, of course but errno is totally volatile, and the
value is *undefined* at least when the 
BIO_CONN_S_BLOCKED_CONNECT code path is
executed.
see `SYSerr(SYS_F_CONNECT, i);`

bernd-edlinger: For instance errno is undefined when the error
is retrieved from getsockopt(s, SOL_SOCKET, SO_ERROR,...)

https://github.com/openssl/openssl/blob/082c041b4233b17b80129d4ac6b33a28014442b0/crypto/bio/bss_conn.c#L190-L193

crypto/bio/bio_lib.c
bernd-edlinger: I wonder why a blocking BIO should set BIO_should_retry, that is unexpected.

crypto/bio/bio_lib.c
```diff
@@ -786,41 +786,48 @@ void bio_cleanup(void)
...
-            milliseconds = (unsigned long)sec_diff * 1000;
+    sec_diff = (long)(max_time - time(NULL)); /* might overflow */
+    if (sec_diff < 0)
+        return 0; /* clearly timeout */
+
+    /* now sleep at most the given number of milliseconds */
+    if (polling_millisec > 1000) {
+
+        if ((unsigned long)sec_diff < polling_millisec / 1000)
+            polling_millisec = (unsigned int)sec_diff * 1000;
```
bernd-edlinger, line 813: This does not work right, if polling_mllisec > 1000 
this will wait 1 seconds, until sec_diff == 0, then
since sec_diff < polling_millisec / 1000
set polling_millisec = 0,
then it will skip the wait below and create 100% CPU load
for the last second....

crypto/bio/bio_lib.c
```diff
@@ -848,29 +857,20 @@ int BIO_connect_retry(BIO *bio, int timeout)
...
-            /*
-             * unless using BIO_reset(), blocking next connect() may crash and
-             * non-blocking next BIO_do_handshake() will fail
-             */
-            goto retry;
-        } else if (BIO_should_retry(bio)) {
-            /* will not actually wait if timeout == 0 (i.e., blocking BIO) */
-            rv = bio_wait(bio, max_time, 100 /* milliseconds */);
+        if (BIO_should_retry(bio)
+                || get_last_sys_error() == EAGAIN /* Temporary gai error */) {
```
bernd-edlinger, line 871: What is a "gai" error?
and if we end up here with EAGAIN and !BIO_should_retry,
that looks like a bug, so why not simply fix the bug?

crypto/bio/bio_lib.c
```diff
@@ -848,29 +857,20 @@ int BIO_connect_retry(BIO *bio, int timeout)
         return 0;
     }
 
+    if (polling_milliseconds < 0)
+        polling_milliseconds = 100;
     if (!blocking)
         BIO_set_nbio(bio, 1);
```
bernd-edlinger, line 863: I'd suggest mak this BIO_set_nbio(bio, !blocking);
You can't be sure what the internal state of this bio is?

crypto/bio/bio_lib.c
```diff
@@ -848,29 +857,20 @@ int BIO_connect_retry(BIO *bio, int timeout)
...
 
- retry: /* it does not help here to set SSL_MODE_AUTO_RETRY */
-    rv = BIO_do_handshake(bio); /* This indirectly calls ERR_clear_error(); */
+ retry: /* it does not always help to set SSL_MODE_AUTO_RETRY */
+    clear_sys_error();
+    rv = BIO_do_connect(bio); /* This indirectly calls ERR_clear_error(); */
```
bernd-edlinger, line 867: looking at It bss_conn.c it don't look like always calling ERR_clear_error,
I only see that happening when the next address is tried (blocking mode)
or when a non-blocking connect is started.

bernd-edlinger:
@DDvO I just spotted another issue in the bss_conn.c file:

https://github.com/openssl/openssl/blob/60d3b5b9ffb8c1273af0cc0338ec1c98f464f4b1/crypto/bio/bss_conn.c#L446-L457

You see case 1 is  BIO_set_conn_port, this should really be a comment here.
but it calls strdup and does not check the result succeeds
And case 2 is BIO_set_conn_address, here is a resource leak since the original
param_hostname and param_service need to be free'd.
Moreover the if(ret) is useless, since ret is just initialized to 1.
But also here is no checking if the operation succeeds.

crypto/bio/bio_lib.c
```diff
@@ -786,41 +786,48 @@ void bio_cleanup(void)
...
 #endif
+    long sec_diff;
 
-    if (max_time == 0)
+    if (max_time == 0) /* no timeout */
         return 1;
 
 #ifndef OPENSSL_NO_SOCK
-    if (BIO_get_fd(bio, &fd) > 0)
+    if (BIO_get_fd(bio, &fd) > 0 && fd < FD_SETSIZE)
```
bernd-edlinger, line 800: Actually the strategy to fall back to polling can
work for read/write but wont work for connect.
Reason is that in the non-blocking case, you have
to wait until the socket is in writable state, but when
the select can't be used for that, you can't re-try
the connect, but have to use getsockopt(s, SOL_SOCKET, SO_ERROR, ...)
and that returns != 0 if an error happened,
but 0 when the connect succeeds, but also when the connect is
not yet completed.  You can assume the connect succeeds,
and continue with send/receive data which will fail, but we dont
have the data to send, and instead want to try the next
address if the previous one fails.

bernd-edlinger: It would make the bio_wait function more useful if it tries
to use the poll(2) API if it is available, and just fail if it is
not available.  I would recommend not to fall back the sleep
and also not to expose this API at all when NO_SOCK is in effect.

crypto/bio/bio_lib.c
bernd-edlinger: I think in case of blocking mode this should just call BIO_do_connect
there is no need for any retires in that case.
What exactly is the motivation for the non-blocking mode here,
put a time-out how long the connect may take?

crypto/bio/bio_lib.c
DDvO: I've meanwhile taken the time to re-do my connection error experiments and made adaptations in this PR accordingly.
* For non-blocking connects the first call to `BIO_do_connect()` always returns with "BIO should retry" and `errno=115` (Operation now in progress) and obviously needs to be retried, which works fine.
* For local ports (e.g., 127.0.0.1 or localhost:8000) that no service listens to there is an immediate "Connection refused" error, which I'd say is fine and there is likely no need to retry connecting until timeout.
* If there is no route to a non-local addresses (e.g., 1.2.3.4:8000 or test.com:8000) there is an immediate "No route to host" error, which I'd say is fine and there is likely no need to retry connecting until timeout.
* For unreachable non-local addresses (e.g., 1.2.3.4:8000 or test.com:8000) an internal timeout occurs after 30 seconds and `get_last_socket_error()` or `BIO_sock_error()` yields `110` (Connection timed out), where a blocking connect sets `errno=110` while a non-blocking connect for some reason retains `errno=115. In both blocking and non-blocking case `BIO_connect_retry()` should retry (including retry of any applicable IP addresses).
If the retry is done directly I cannot reproduce a crash, but a connect error is thrown again immediately. Therefore `ERR_clear_error(); (void)BIO_reset(bio);` are really necessary.
For enabling `BIO_connect_retry()` to correctly detecting these internal timeout (`ETIMEDOUT`) situations I've added for both the blocking and non-blocking case a differentiation of the error reason thrown: `BIO_R_CONNECT_TIMEOUT` and `BIO_R_NBIO_CONNECT_TIMEOUT`, respectively.
* For DNS trouble on non-local addresses (e.g., 1.2.3.4:8000 or test.com:8000) an internal timeout occurs after 40 seconds and `BIOerr(BIO_F_BIO_LOOKUP_EX, ERR_R_SYS_LIB)` is thrown and `get_last_sys_error()` yields `EAGAIN` (resource temporarily unavailable) and `BIO_connect_retry()` should retry.

The adaptations lead to major cleanup of `BIO_connect_retry()`.

crypto/bio/bio_lib.c
DDvO: Meanwhile I found the email exchange I referred to 3 days ago: 
https://mta.openssl.org/pipermail/openssl-dev/2017-September/009697.html

crypto/bio/bio_lib.c
DDvO: By "gai" error I meant the following `getaddrinfo()` error: Temporary failure in name resolution.
I've improved the commenting and the handling of this case in my most recent push.
This is all I can do today; anything else needs to wait until tomorrow.

crypto/bio/bio_lib.c
bernd-edlinger: > while a non-blocking connect for some reason retains errno=115.

I think that is basically what I claim that the errno is undefined when
the non-blocking socket connect fails:
```C
           i = BIO_sock_error(b->num);
            if (i != 0) {
                BIO_clear_retry_flags(b);
                if ((c->addr_iter = BIO_ADDRINFO_next(c->addr_iter)) != NULL) {
                    /*
                     * if there are more addresses to try, do that first
                     */
                    BIO_closesocket(b->num);
                    c->state = BIO_CONN_S_CREATE_SOCKET;
                    ERR_clear_error();
                    break;
                }
                ERR_raise_data(ERR_LIB_SYS, i,
                               "calling connect(%s, %s)",
                                c->param_hostname, c->param_service);
                BIOerr(BIO_F_CONN_STATE, BIO_R_NBIO_CONNECT_ERROR);
                ret = 0;
                goto exit_loop;
```
does nowere set errno, ETIMEDOUT
while `i = BIO_sock_error(b->num)` should return ETIMEDOUT,
so errno is still 115 EINPROGRESS.

crypto/bio/bio_lib.c
```diff
@@ -848,35 +857,44 @@ int BIO_connect_retry(BIO *bio, int timeout)
...
+        if (ERR_GET_LIB(err) == ERR_LIB_BIO) {
+            if (reason == BIO_R_NBIO_CONNECT_TIMEOUT
+                    || reason == BIO_R_CONNECT_TIMEOUT) {
+                /* premature timeout has occurred, typically after 30 seconds */
+                (void)BIO_reset(bio); /* needed to avoid failure on retry */
+                ERR_clear_error();
+                do_retry = 1;
+            }
+            else if (reason == ERR_R_SYS_LIB && get_last_sys_error()
+                     == EAGAIN /* resource temporarily unavailable */) {
```
bernd-edlinger, line 882: I personally, would stop this loop when the address resolution fails.
Technically it is problemati since when we come here, lots of code has run and could clobber errno,
would it be possible to get the original errno from there:
```C
       switch ((gai_ret = getaddrinfo(host, service, &hints, res))) {
# ifdef EAI_SYSTEM
        case EAI_SYSTEM:
            ERR_raise_data(ERR_LIB_SYS, get_last_socket_error(),
                           "calling getaddrinfo()");
            BIOerr(BIO_F_BIO_LOOKUP_EX, ERR_R_SYS_LIB);
```

crypto/bio/bio_lib.c
DDvO: Yes, so I've meanwhile made sure that `BIO_do_retry()` is independent of `errno`,
which in my view is cleaner anyway.

crypto/bio/bio_lib.c
DDvO: The comments that - justifiably - raised questions are gone meanwhile,
due to the re-structuring of the retry checks that I did yesterday.
Are you fine with the new version?

crypto/bio/bio_lib.c
DDvO: Right, `BIO_should_retry()` should be needed only for the non-blocking case.
I've meanwhile added a respective comment.

crypto/bio/bio_lib.c
DDvO: This issue has been solved meanwhile, by avoiding the use of `errno`, as just mentioned above.

crypto/bio/bio_lib.c
DDvO: > I'd suggest mak this BIO_set_nbio(bio, !blocking);
> You can't be sure what the internal state of this bio is?

Done.

crypto/bio/bio_lib.c
bernd-edlinger: I think this is not completely right, this should not be skipped if polling_millisec <= 1000,
since sec_diff can be 0, but you do not sleep(0)

crypto/bio/bio_lib.c
DDvO: > I think in case of blocking mode this should just call BIO_do_connect
> there is no need for any retires in that case.
> What exactly is the motivation for the non-blocking mode here,
> put a time-out how long the connect may take?

The retry is still needed in the non-blocking case, as meanwhile made clear in the revised code and its comments:
Even for blocking connects non-local connection attempts may return a premature error in case there a system-internal timeout occurs (e.g., when the intended host is currently not reachable) and in case a domain name resolution fails temporarily.
I've confirmed these situation in my yesterday's experiments.

crypto/bio/bio_lib.c
```diff
@@ -786,53 +786,62 @@ void bio_cleanup(void)
...
-    if (milliseconds > 1000) {
-        long sec_diff = (long)(max_time - time(NULL)); /* might overflow */
+    /* fall back to polling since no sockets are available */
 
-        if (sec_diff <= 0)
-            return 0; /* timeout */
-        if ((unsigned long)sec_diff < milliseconds / 1000)
-            milliseconds = (unsigned long)sec_diff * 1000;
+    sec_diff = (long)(max_time - time(NULL)); /* might overflow */
+    if (sec_diff < 0)
```
bernd-edlinger, line 806: previously this was if (sec_diff <= 0) was this change intentional?

crypto/bio/bio_lib.c
DDvO: Correct. Yet my comment was not meant as an assertion that `ERR_clear_error()` is called in any case but as a warning that it may be called, such that any previous error queue contents may be lost.
I've meanwhile made this clear, also in the documentation, in my today's fixup commit.

crypto/bio/bio_lib.c
DDvO: No, if `polling_millisec is > 1000` then the function determines `max(polling_millisec, 1000*sec_def)` and waits accordingly, as mentioned in the comment above that code.

crypto/bio/bss_conn.c
```diff
@@ -202,15 +198,27 @@ static int conn_state(BIO *b, BIO_CONNECT *c)
...
+                    BIOerr(BIO_F_CONN_STATE, BIO_R_NBIO_CONNECT_ERROR);
                 ret = 0;
                 goto exit_loop;
             } else
                 c->state = BIO_CONN_S_OK;
             break;
 
         case BIO_CONN_S_CONNECT_ERROR:
-            BIOerr(BIO_F_CONN_STATE, BIO_R_CONNECT_ERROR);
+            i = get_last_socket_error();
```
bernd-edlinger, line 212: the callback is called, and could do anything with the errno.

crypto/bio/bio_lib.c
bernd-edlinger: but in last second of the timeout, 1000*sec_def is null, and then you have a busy loop.
jut for the last second, but I think that is not what you indend.

bernd-edlinger: And actually this computes `polling_millisec = MIN(polling_millisec, 1000*sec_def)` 
not MAX. so it can be much smaller than polling_millisec.

crypto/bio/bio_lib.c
bernd-edlinger: > Done.

but not pushed :-)

> The retry is still needed 

So you re-try the connect even if it is a hard failure.
Now I understand at least what the code should do...

crypto/bio/bio_lib.c
DDvO: Meanwhile I see that caling `get_last_socket_error()` is essentially the same as reading `errno`.

In my yesterday's experiments I confirmed that the (previous and current) implementation works fine:
If I use `BIO_do_connect_retry()` in blocking mode or with a sufficiently large timeout value and at first the DNS does not work but after some retries it starts working, the function gets to learn this fact in that way, and the connection succeeds.

crypto/bio/bio_lib.c
```diff
@@ -848,35 +857,44 @@ int BIO_connect_retry(BIO *bio, int timeout)
...
-        } else if (BIO_should_retry(bio)) {
-            /* will not actually wait if timeout == 0 (i.e., blocking BIO) */
-            rv = bio_wait(bio, max_time, 100 /* milliseconds */);
+        int err = ERR_peek_last_error();
+        int reason = ERR_GET_REASON(err);
+        int do_retry = BIO_should_retry(bio);
+
+        if (ERR_GET_LIB(err) == ERR_LIB_BIO) {
+            if (reason == BIO_R_NBIO_CONNECT_TIMEOUT
+                    || reason == BIO_R_CONNECT_TIMEOUT) {
```
bernd-edlinger, line 875: With the BIO_R_CONNECT_TIMEOUT this is really an endless loop.
the blocking connect is only tried when timeout == 0
I would also not differentiate between NBIO_CONNECT_TIMEOUT and NBIO_CONNECT_ERROR,
if the server does not listen, you can get a CONNECT_ERROR and you can of course
try that connection again after a short nap.

crypto/bio/bss_conn.c
DDvO: Sure a callback could do all kinds of things, but it should mess with `errno` only if there is good reason.
And how else would you suggest that one can detect the `getaddrinfo: Temporary failure in name resolution` situation?

bernd-edlinger: Yes, the getaddrinfo is something I hadn't considered before.
I don't really know, but wouldn't it be possible to get the 
errno value back what was entered to the error queue, right
when the getaddrinfo failed?

But I don't fully agree that this change is an improvement,
the difference when a connect returns TIMEOUT and when
it returns a different error, is not very big.  Just an lost
packet for instance.  But if a server is not online, you can
also try again it might just starting.
For the do_connect_retry it would be okay to retry as long
as the time is over.  I had not considered that possibility
but that can be done, but then I would just do that.

crypto/bio/bio_lib.c
DDvO: > With the BIO_R_CONNECT_TIMEOUT this is really an endless loop.
the blocking connect is only tried when timeout == 0

Indeed it could result in an endless loop (yet not in a busy loop, due to `BIO_wait`),
but if the user sets `timeout == 0` this is exactly what he wants: keep trying to connect as long as whatever needed (even infinitely) as long as the connection cannot be established.

> I would also not differentiate between NBIO_CONNECT_TIMEOUT and NBIO_CONNECT_ERROR,
if the server does not listen, you can get a CONNECT_ERROR and you can of course
try that connection again after a short nap.

Interesting thought. Yet I would assume that `NBIO_CONNECT_ERROR` is something fatal, while `NBIO_CONNECT_TIMEOUT` according to my experiments is a non-fatal error just indicating that within the strange 30 s internal grace period no connection could be established, so when the user is patient enough it makes sense to retry.

crypto/bio/bio_lib.c
DDvO: Oops, I did actually mean: min.

DDvO: If polling_millisec < 1000 the function does not wait zero time even if sec_diff is 0,
but for the intended nap period.

bernd-edlinger: Yes, that case is of course more reasonable.
You wouldn't get that sort of problem if you
consider sec_diff = 0 as timeout.

crypto/bio/bio_lib.c
bernd-edlinger: It is your function.

But again, I did not assume that timeout=0 means forever, I did just assume,
try the connection, to all addresses that match to the request exactly once.

For the blocking connect, you get only the error code of the last connect
attempt, the first few could have a CONNECT_ERROR and then a few
CONNECT_TIMEOUT and then it is an endless loop.
But if the first connect has a CONNECT_TIMEOUT and then a CONNECT_ERROR,
then you stop the loop.

bernd-edlinger: Same in the non-blocking connect.
You get here only when the *last* address that getaddrinfo returned
had a TIMEOUT.  That looks bogus to take that as an indication
when to terminate the loop.

crypto/bio/bio_lib.c
DDvO: Oops, that commit went wrong. Now it's in and pushed :)

crypto/bio/bio_lib.c
DDvO: Yes, because if `sec_diff == 0` the calculation of the minimum would go wrong and the sec_diff resolution insufficient if `polling_millisec` < 1000.

crypto/bio/bss_conn.c
DDvO: > Yes, the getaddrinfo is something I hadn't considered before.
> I don't really know, but wouldn't it be possible to get the
> errno value back what was entered to the error queue, right
> when the getaddrinfo failed?

Indeed it would be possible and likely more safe to consult `errno` just after `getaddrinfo` failed, and then add a to-the-point entry/reason to the error queue

> But I don't fully agree that this change is an improvement,
> the difference when a connect returns TIMEOUT and when
> it returns a different error, is not very big. Just an lost
> packet for instance. But if a server is not online, you can
> also try again it might just starting.
> For the do_connect_retry it would be okay to retry as long
> as the time is over. I had not considered that possibility
> but that can be done, but then I would just do that.

You mean: as long the time is *not* over.
Good point, so we could at least to some extent lump timeout and error.
For some cases (`ETIMEDOUT`) I found that we need to call `BIO_reset()` to prevent immediate failure of the retry, but maybe such a call does not hurt also for all other error cases? Then much of the distinctions I added could be removed.

crypto/bio/bio_lib.c
DDvO: You just convinced me that it is more consistent - and also simpler of course - not to distinguish the timeout and error cases. 
I'll give that a try...

crypto/bio/bss_conn.c
bernd-edlinger: > Indeed it would be possible and likely more safe to consult errno just after getaddrinfo failed, and then add a to-the-point entry/reason to the error queue

I really don't know but isn't the errno here entered into the error queue?
https://github.com/openssl/openssl/blob/a5a87011baeef71c86938a2bae54f89fbe99e5dc/crypto/bio/b_addr.c#L705-L708

It would feel much safer to get the errno back from the error queue,
but if that is not possible, if the DNS says no this name is not known *now*
you could as well try again after the short nap.

It would this connect_retry less surprising when there function simply retries
what it is doing until that timeout is over or it succeeds then it leaves the loop.

>  I found that we need to call BIO_reset() 

I think I begin to understand what this does.  in the case when
the connect fails, the socket is *not* closed, just the *connect*
failed.  and the state stays at BIO_CONN_S_CONNECT_ERROR.
Without that reset the state machine will not continue with
BIO_CONN_S_BEFORE.

Yes, if that is the correct behaviour for your application you just
can reset the socket and try to connect until that time is over.
I just think that this may be a parameter, since some clients,
may just want to connect once, and let the user decide what to
do next.

crypto/bio/bio_lib.c
```diff
@@ -848,35 +858,42 @@ int BIO_connect_retry(BIO *bio, int timeout)
...
 
-    if (!blocking)
-        BIO_set_nbio(bio, 1);
+    if (polling_milliseconds < 0)
+        polling_milliseconds = 100;
+    BIO_set_nbio(bio, !blocking);
 
- retry: /* it does not help here to set SSL_MODE_AUTO_RETRY */
-    rv = BIO_do_handshake(bio); /* This indirectly calls ERR_clear_error(); */
+ retry: /* it does not always help to set SSL_MODE_AUTO_RETRY */
```
bernd-edlinger, line 865: The comment mentions SSL_MODE_AUTO_RETRY
This error handling is for BIO_TYPE_CONNECT,
not for BIO_TYPE_SSL, right?

crypto/bio/bio_lib.c
DDvO: So I've unified the handling of  CONNECT_TIMEOUT and CONNECT_ERROR,
which is most likely an improvement, at least in terms of consistency.

It turns out that this does not yet cover the system error cases, in particular `EAGAIN` of `getaddrinfo` (Temporary failure in name resolution), so I've added the `ERR_R_SYS_LIB` case to cover all system errors, assuming that it makes sense to retry on any of them.

Experiments after these changes went fine. 
So it looks like we really can treat all non-fatal errors like internal  timeouts: reset the BIO, clear the error queue, and retry.

crypto/bio/bio_lib.c
DDvO: I also think that SSL_MODE_AUTO_RETRY is specifically for SSL-type BIOs, and this is one of the reasons why in general it is not sufficient.

crypto/bio/bio_lib.c
DDvO: Yes, but as documented I'd like to give the connect a last chance in case the remaining time before timeout is less than the resolution of the timeout value, which is one second.

crypto/bio/bio_lib.c
DDvO: > Actually the strategy to fall back to polling can
> work for read/write but wont work for connect.
> Reason is that in the non-blocking case, you have
> to wait until the socket is in writable state, but when
> the select can't be used for that, you can't re-try
> the connect, but have to use getsockopt(s, SOL_SOCKET, SO_ERROR, ...)
> and that returns != 0 if an error happened,
> but 0 when the connect succeeds, but also when the connect is
> not yet completed. You can assume the connect succeeds,
> and continue with send/receive data which will fail, but we dont
> have the data to send, and instead want to try the next
> address if the previous one fails.

> It would make the bio_wait function more useful if it tries
> to use the poll(2) API if it is available, and just fail if it is
> not available. I would recommend not to fall back the sleep
> and also not to expose this API at all when NO_SOCK is in effect.

I've tried several times to understand what exactly you meant here.
When resorting in the non-blocking case to polling then this is because the BIO is not socket-based and then one likely cannot use `getsockopt()` nor the poll API since they work on a socket.
For non-socket based connects it should still make sense to nap and poll again on a failed connect - AFAICS you don't have other chances than doing that (or simply failing potentially prematurely).

Please note also that `BIO_wait()` is also used for read/write, and at least there we agree that falling back to polling does make sense.

crypto/bio/bss_conn.c
DDvO: > I really don't know but isn't the errno here entered into the error queue?

Yes, it is entered by `ERR_raise_data(ERR_LIB_SYS, get_last_socket_error(), "calling getaddrinfo()");` but then it is buried by `BIOerr(BIO_F_BIO_LOOKUP_EX, ERR_R_SYS_LIB);`.
Meanwhile `BIO_do_connect_retry()`, as you suggested, does not any more depend on `errno`.

DDvO: > Yes, if that is the correct behaviour for your application you just
can reset the socket and try to connect until that time is over.

This is meanwhile implemented.

DDvO: > I just think that this may be a parameter, since some clients,
may just want to connect once, and let the user decide what to
do next.

I like the idea to offer this option, although in this case the function degenerates to a normal `BIO_do_connect() call`.
Nevertheless I have implemented it. I did so not as an extra parameter but in case `timeout < 0`, which so far had no real use for `BIO_do_connect_retry()`.
This adds to uniformity and even simplifies the implementation.

crypto/bio/bio_lib.c
bernd-edlinger: I don't understand your reasoning, as you say `if sec_diff == 0 the calculation of the minimum would go wrong`
I think you never want a sleep of 0 ms.  The easiest fix would be to consider sec_diff == 0 as time out.

crypto/bio/bio_lib.c
```diff
@@ -786,55 +786,65 @@ void bio_cleanup(void)
...
-        if (sec_diff <= 0)
-            return 0; /* timeout */
-        if ((unsigned long)sec_diff < milliseconds / 1000)
-            milliseconds = (unsigned long)sec_diff * 1000;
+    sec_diff = (long)(max_time - time(NULL)); /* might overflow */
+    if (sec_diff < 0)
+        return 0; /* clearly timeout */
+
+    /* now take a nap at most the given number of milliseconds */
+    if (nap_milliseconds > 1000) {
```
bernd-edlinger, line 810: This condition is redundant, if nap_milliseconds < 1000 the following
condition is sec_diff < 0 so never satisfied.
if nap_milliseconds == 1000 the following condition is only satisfied,
if sec_diff < 1, but this changes 1000ms to 0ms delay.
That is not what you want.

bernd-edlinger: ...and it does not happen when nap_milliseconds = 999, then you have
a sleep 999, in the last second.

crypto/bio/bss_conn.c
```diff
@@ -202,15 +198,27 @@ static int conn_state(BIO *b, BIO_CONNECT *c)
...
+                else
+                    BIOerr(BIO_F_CONN_STATE, BIO_R_NBIO_CONNECT_ERROR);
                 ret = 0;
                 goto exit_loop;
             } else
                 c->state = BIO_CONN_S_OK;
             break;
 
         case BIO_CONN_S_CONNECT_ERROR:
-            BIOerr(BIO_F_CONN_STATE, BIO_R_CONNECT_ERROR);
```
bernd-edlinger, line 213: please drop this change.

crypto/bio/bss_conn.c
```diff
@@ -202,15 +198,27 @@ static int conn_state(BIO *b, BIO_CONNECT *c)
                 ERR_raise_data(ERR_LIB_SYS, i,
                                "calling connect(%s, %s)",
                                 c->param_hostname, c->param_service);
-                BIOerr(BIO_F_CONN_STATE, BIO_R_NBIO_CONNECT_ERROR);
+                if (i == ETIMEDOUT)
```
bernd-edlinger, line 201: please drop this change.

crypto/bio/bss_conn.c
```diff
@@ -452,19 +461,26 @@ static long conn_ctrl(BIO *b, int cmd, long num, void *ptr)
...
-            } else if (num == 2) {
+                if ((data->param_service = OPENSSL_strdup(ptr)) == NULL)
+                    ret = 0;
+            } else if (num == 2) { /* BIO_set_conn_address */
                 const BIO_ADDR *addr = (const BIO_ADDR *)ptr;
+                char *host = BIO_ADDR_hostname_string(addr, 1);
+                char *service = BIO_ADDR_service_string(addr, 1);
+
+                ret = host != NULL && service != NULL;
                 if (ret) {
```
bernd-edlinger, line 474: remove this if, ret is always 1 here.

bernd-edlinger: oops. disregard the previous comment.
instead you need to add an else block
and free the host and service in case only
one was NULL.

crypto/bio/bss_conn.c
```diff
@@ -438,12 +446,13 @@ static long conn_ctrl(BIO *b, int cmd, long num, void *ptr)
     case BIO_C_SET_CONNECT:
         if (ptr != NULL) {
             b->init = 1;
-            if (num == 0) {
+            if (num == 0) { /* BIO_set_conn_hostname */
```
bernd-edlinger, line 449: Thanks for the comments.

This part would be good to have in the 1.1.1 branch.

crypto/bio/bss_conn.c
bernd-edlinger: Yes.  I think that is getting better now.

crypto/bio/bio_lib.c
bernd-edlinger: I just want to make sure this comment is not understood that this logic here is
prepared to handle BIO_TYPE_SSL, those have other special wait conditions,
for instance BIO_get_retry_reason() == BIO_RR_SSL_X509_LOOKUP

DDvO:
> @DDvO I just spotted another issue in the bss_conn.c file:
> 
> https://github.com/openssl/openssl/blob/60d3b5b9ffb8c1273af0cc0338ec1c98f464f4b1/crypto/bio/bss_conn.c#L446-L457
> 
> You see case 1 is BIO_set_conn_port, this should really be a comment here.
> but it calls strdup and does not check the result succeeds
> And case 2 is BIO_set_conn_address, here is a resource leak since the original
> param_hostname and param_service need to be free'd.
> Moreover the if(ret) is useless, since ret is just initialized to 1.
> But also here is no checking if the operation succeeds.

Although these issues are rather far from the topic of this PR I've fixed them along with it, in commit 76d9cc01cdbed5eba472c2983cdca7a4c535528e.

crypto/bio/bss_conn.c
DDvO: Ah yes, good that you reminded me of that case. Done.

crypto/bio/bss_conn.c
DDvO: Argh, I just did this backport and by mistake directly pushed it to `OpenSSL_1_1_1-stable` :(
See #12034.

crypto/bio/bss_conn.c
DDvO: Done. Also for the blocking case.

crypto/bio/bss_conn.c
DDvO: I dropped the differentiation w.r.t. `ETIMEDOUT`, but the move of
```
                   ERR_raise_data(ERR_LIB_SYS, get_last_socket_error(),
                                  "calling connect(%s, %s)",
                                   c->param_hostname, c->param_service);
```
to the new code location is for consistency with the non-blocking case,
and the addition of the guard `i = get_last_socket_error(); if (i != 0)` not only is for consistency but likely also a necessary safeguard. At least for some (good?) reason this guard is present for the non-blocking case.

crypto/bio/bss_conn.c
DDvO: The actual backport PR is now in #12036.

crypto/bio/bio_lib.c
DDvO: By `would go wrong` if `sec_diff == 0` I meant that in this case `min(sec_diff * 1000, nap_milliseconds)` would be 0 ms, and this is what the code was intended to prevent. 
But the calculation was really buggy - I've just revised it.
Thanks for having a close look at that!

crypto/bio/bio_lib.c
DDvO: I've revised the calculation. Does it make sense now?

crypto/bio/bio_lib.c
DDvO: Do you have a suggestion how to phrase the comment better,
or should I simply drop it because it may give more confusion than explanation?

bernd-edlinger: Yes please drop the comment.

crypto/bio/bss_conn.c
```diff
@@ -174,9 +173,6 @@ static int conn_state(BIO *b, BIO_CONNECT *c)
                     ERR_clear_error();
                     break;
                 } else {
-                    ERR_raise_data(ERR_LIB_SYS, get_last_socket_error(),
-                                   "calling connect(%s, %s)",
-                                    c->param_hostname, c->param_service);
```
bernd-edlinger, line 179: please undo this change.

crypto/bio/bss_conn.c
```diff
@@ -8,7 +8,6 @@
  */
 
 #include <stdio.h>
-#include <errno.h>
```
bernd-edlinger, line 11: maybe undo this too.

crypto/bio/bss_conn.c
```diff
@@ -210,6 +206,12 @@ static int conn_state(BIO *b, BIO_CONNECT *c)
             break;
 
         case BIO_CONN_S_CONNECT_ERROR:
+            i = get_last_socket_error();
```
bernd-edlinger, line 209: please revert this change, reason that errno is likely no preseved over the callback.

crypto/bio/bio_lib.c
```diff
@@ -843,40 +854,47 @@ int BIO_connect_retry(BIO *bio, int timeout)
...
+            case BIO_R_NBIO_CONNECT_ERROR:
+                /* some likely retryable connection error occurred */
+                (void)BIO_reset(bio); /* often needed to avoid retry failure */
+                do_retry = 1;
+                break;
+            default:
+                break;
+            }
+        }
+        if (do_retry || timeout < 0) {
```
bernd-edlinger, line 888: do_retry && timeout >= 0 ?

crypto/bio/bio_lib.c
```diff
@@ -786,55 +786,66 @@ void bio_cleanup(void)
...
         BIOerr(0, rv == 0 ? BIO_R_TRANSFER_TIMEOUT : BIO_R_TRANSFER_ERROR);
     return rv;
 }
 
 /*
- * Connect via given BIO using BIO_do_handshake() until success/timeout/error.
- * Parameter timeout == 0 means infinite, < 0 leads to immediate timeout error.
+ * Connect via given BIO using BIO_do_connect() until success/timeout/error.
+ * Parameter timeout == 0 means no timeout, < 0 means exactly one try.
+ * For non-blocking and potentially even non-socket BIOs perform polling with
```
bernd-edlinger, line 841: which non-socket BIOs have you in mind?
have you tested this code with one of those?

crypto/bio/bio_lib.c
bernd-edlinger: yes.

crypto/bio/bss_conn.c
bernd-edlinger: maybe I missed something, but you should never move the errno away from the
place where the error happened, and especially not call a callback in between.

crypto/bio/bio_lib.c
DDvO: Done

crypto/bio/bss_conn.c
DDvO: All right, I've reverted this.
Likely the guard `i = get_last_socket_error(); if (i != 0)` is not critical.

crypto/bio/bss_conn.c
DDvO: Done

crypto/bio/bss_conn.c
DDvO: Done (though the #include is not really needed, 
but `get_last_socket_error()` indirectly uses `errno`).

crypto/bio/bss_conn.c
DDvO: Good point, done.

crypto/bio/bio_lib.c
DDvO: Oops, yes!
Corrected.

crypto/bio/bss_conn.c
bernd-edlinger: I wear special roentgen glasses that look throu macros :-)

crypto/bio/bio_lib.c
DDvO: I don't have such a non-socket BIO at hand, so could not test this.
That's part of the reason I wrote "potentially even non-socket BIOs".
Anyway, for non-blocking socket-based BIOs (i.e., timeout > 0) polling is needed.

DDvO:
Thanks a lot @bernd-edlinger for diligently working with me through the rather fresh high-level BIO connect code. This was much more effort that I would have expected, but certainly worth while because now it is corrected and much cleaner, and also the documentation has improved.

We also fixed some more or less related issues in bss_acpt.c and bss_conn.c.

I've rebased and squashed this PR as much as it made sense.

crypto/bio/bio_lib.c
bernd-edlinger: Well, that's why I'd initially restrict the scope of such a function,
to the domain where it is used, and extend that later
when other non-socket use become available.
I have the impression that must be invented first.

bernd-edlinger:
suggestions for further improvement:
use poll(2) if available,
make whole api depend on OPENSSL_NO_SOCK
unless other concrete use cases can be found.
thus reduce code size.

DDvO:
> This looks good to me, but I'll defer to @bernd-edlinger to actually approve

@levitte, now since @bernd-edlinger approved,
 I think this also needs your (formal) blessing.

DDvO:
Also here I still need an OTC approval.

This PR should fix by the way #11449.

DDvO:
@levitte, is the state of this PR okay for you as well?

DDvO:
Merged - thanks again!

PR #11988, opened 2020-05-28
fix a docs typo

oconnor663:
Correct "EC_KEY_point2buf" to "EC_POINT_point2buf". The former does not exist.

##### Checklist

### DIFF
--- a/doc/man3/EC_KEY_new.pod
+++ b/doc/man3/EC_KEY_new.pod
@@ -146,7 +146,7 @@ Modern versions should instead switch to named curves which OpenSSL has
 hardcoded lookup tables for.
 
 EC_KEY_oct2key() and EC_KEY_key2buf() are identical to the functions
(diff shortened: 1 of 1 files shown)


oconnor663:
This probably counts as a trivial change that doesn't require signing a CLA. That said, I might as well go ahead with the CLA process. Is there a way to do it that doesn't require a printer?

t8m:
If you want, you can amend the commit message to mark the commit as trivial. See
https://github.com/openssl/openssl/blob/master/CONTRIBUTING.md

oconnor663:
Ok, I added `CLA: trivial`.

kaduk:
FWIW several people have used some form of e-signature for their CLA (i.e., no printer), but I don't know the details of what was used, since I'm just watching from the sidelines for that part of things.

t8m:
Merged to master [a5a87011ba] and 1.1.1 [315170f662]. Thank you for the PR.

PR #11989, opened 2020-05-28
Make BIO_do_connect() and friends handle multiple IP addresses also in 1.1.1

DDvO:
This backports #11971 as requested: https://github.com/openssl/openssl/pull/11971#issuecomment-634679197

### DIFF
--- a/crypto/bio/bss_conn.c
+++ b/crypto/bio/bss_conn.c
@@ -186,8 +186,17 @@ static int conn_state(BIO *b, BIO_CONNECT *c)
 
         case BIO_CONN_S_BLOCKED_CONNECT:
             i = BIO_sock_error(b->num);
-            if (i) {
+            if (i != 0) {
                 BIO_clear_retry_flags(b);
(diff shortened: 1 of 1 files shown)


DDvO:
BTW, I cannot any more locally build `OpenSSL_1_1_1-stable`.
Even after `make distclean` and `./config no-dso` I get:
```
/usr/bin/ld: ./libcrypto.so: undefined reference to `DSO_METHOD_openssl'
```

bernd-edlinger:
cannot reproduce your build problem, try make distclean and let use git clean afterwards,
maybe something is lingering.

DDvO:
@bernd-edlinger, `git clean` was an excellent hint!
It showed that (from checking out `master` in between) there were some remnant
`crypto/bn_conf.h crypto/dso_conf.h crypto/opensslconf.h`
and after removing those the build went fine again.

bernd-edlinger:
Yeah, moreover make distclean is broken in master.
But I tried to fix the PR stalled for months.  Now I have given up

DDvO:
Merged - thanks @bernd-edlinger and @t8m!

PR #4930, opened 2017-12-14
Various fixes regarding PKCS#12 input and related cleanup of apps, doc, and tests

DDvO:
***UPDATE:*** Originally this PR was about extending support for PKCS#12 input in apps.
I've meanwhile carved out the most interesting pieces of that and contributed them separately.
This is the leftovers fixing several corner cases in PKCS#12 input and its error handling.
There are also some rather unrelated fixes to several apps and their documentation, which I could separate if requested.     

##### Checklist

- [x] documentation is added or updated
- [x] tests are added or updated

PR #11993, opened 2020-05-29
Centralize bio error

richsalz:
The apps make use of some common library routines to do things like "dup stdout" or "open a file for writing"  The callers were not consistent in error-checking.  So, we do like app_malloc does and return a valid BIO or print the error stack and exit.

There were a couple of places where we needed to "silently fail" -- in cases where we *tried* to open a file (config file) and didn't care if it failed. That was handled be rewriting code to do the fopen first, and then failing if ```BIO_new_fp``` failed.  I also did the same thing for cmp.c, but I'm not sure if that's needed, so pinging @ddvo to look at that.  Also pinging @mspncp and @levitte who promised to review this. :)


apps/lib/apps.c
```diff
@@ -2370,16 +2366,29 @@ int raw_write_stdout(const void *buf, int siz)
...
+
+    return check_bio(BIO_new_fp(stdin, mode), "stdin", mode);
 }
 
 BIO *dup_bio_out(int format)
 {
-    BIO *b = BIO_new_fp(stdout,
-                        BIO_NOCLOSE | (FMT_istext(format) ? BIO_FP_TEXT : 0));
+    int mode = BIO_NOCLOSE | (FMT_istext(format) ? BIO_FP_TEXT : 0);
+    BIO *b = BIO_new_fp(stdout, mode);
```
bernd-edlinger, line 2391: should we checke that b != NULL ?

apps/req.c
```diff
@@ -887,8 +887,6 @@ int req_main(int argc, char **argv)
                            keyout != NULL && outfile != NULL &&
                            strcmp(keyout, outfile) == 0 ? 'a' : 'w',
```
bernd-edlinger, line 888: oh, this wont work in windows...

apps/lib/apps.c
richsalz: Yes, adding check.

apps/req.c
richsalz: well, I didn't touch that code.  I only removed the "if" check after that.

bernd-edlinger: never mind.  maybe better leave some things for later...

apps/lib/apps.c
```diff
@@ -2518,46 +2528,21 @@ BIO *bio_open_owner(const char *filename, int format, int private)
     return NULL;
```
bernd-edlinger, line 2528: Could it be that bio_open_owner should also follow the new model?
in one code path it returns bio_open_default and the other path
requires the caller to check for errors?
What do you think?

richsalz: I took a quick look and it didn't seem easy, so I'm not opposed but I'd rather do it in a separate PR.  Can't do the fopen/new_fp trick because of what it does, but yeah, making it return or fail/exit makes sense.

apps/lib/apps.c
```diff
@@ -2370,41 +2366,58 @@ int raw_write_stdout(const void *buf, int siz)
...
+    int mode = BIO_NOCLOSE | (FMT_istext(format) ? BIO_FP_TEXT : 0);
+    BIO *b = BIO_new_fp(stdout, mode);
     void *prefix = NULL;
 
+    /* Only care about NULL error-exit. */
+    check_bio(b, "stdout", mode);
+
 #ifdef OPENSSL_SYS_VMS
     if (FMT_istext(format))
         b = BIO_push(BIO_new(BIO_f_linebuffer()), b);
```
bernd-edlinger, line 2399: I think this nees another check_bio, @levitte may be able to test.

apps/lib/apps.c
```diff
@@ -2370,41 +2366,58 @@ int raw_write_stdout(const void *buf, int siz)
...
-    return b;
+    return check_bio(b, "stdout", mode);
 }
 
 BIO *dup_bio_err(int format)
 {
-    BIO *b = BIO_new_fp(stderr,
-                        BIO_NOCLOSE | (FMT_istext(format) ? BIO_FP_TEXT : 0));
+    int mode = BIO_NOCLOSE | (FMT_istext(format) ? BIO_FP_TEXT : 0);
+    BIO *b = BIO_new_fp(stderr, mode);
```
bernd-edlinger, line 2414: likewise checks missing here

apps/lib/apps.c
bernd-edlinger: More important is probably 1.1.1 back-port, most of the time that is forgotten nowadays.

apps/lib/apps.c
```diff
@@ -2518,46 +2531,21 @@ BIO *bio_open_owner(const char *filename, int format, int private)
...
-        if (ret != NULL)
-            return ret;
-        BIO_printf(bio_err,
-                   "Can't open %s for %s, %s\n",
+    /* Using stdin or stdout? */
+    if (filename == NULL || strcmp(filename, "-") == 0)
+        return mode == 'r' ? dup_bio_in(format) : dup_bio_out(format);
+
+    ret = BIO_new_file(filename, modestr(mode, format));
+    if (ret == NULL) {
```
bernd-edlinger, line 2543: a job for your new check function?

apps/lib/apps.c
richsalz: Actually, restructuring the code and it was easy.  Will be in an additional commit shortly.

apps/lib/apps.c
richsalz: the code does "if first-param == NULL return b" so test isn't strictly needed.

apps/lib/apps.c
richsalz: same reason, if you push a NULL bio you get the original BIO back so no check needed.

apps/lib/apps.c
richsalz: Okay, but |mode| doesn't quite fit, going to leave it as 'r' if that's okay.

apps/lib/apps.c
bernd-edlinger: Hmm, but if BIO_new fails, then BIO_push returns b,
so checking the return value of BIO_push is probably not enough.

richsalz:
Additional commit pushed adding more calls; you convinced me.

apps/lib/apps.c
```diff
@@ -2504,28 +2493,19 @@ BIO *bio_open_owner(const char *filename, int format, int private)
...
-    if (fp)
-        fclose(fp);
-    else if (fd >= 0)
-        close(fd);
-    return NULL;
+    if (fd < 0
+            || (fp = fdopen(fd, modestr('w', format))) == NULL
+            || (b = BIO_new_fp(fp, bflags)) == NULL) {
+        BIO_printf(bio_err, "%s: Can't open \"%s\" for writing, %s\n",
+                opt_getprog(), filename, strerror(errno));
```
bernd-edlinger, line 2505: indentation wrong

apps/lib/apps.c
bernd-edlinger: the check is still needed, BIO_push would happily push NULL as second parameter.
unclear what the sematic would be only the BIO_CTRL_PUSH could do something.

apps/lib/apps.c
richsalz: fixup commit pushed.

apps/lib/apps.c
richsalz: fixup commit added.

apps/lib/apps.c
```diff
@@ -2370,41 +2366,60 @@ int raw_write_stdout(const void *buf, int siz)
...
-        && (prefix = getenv("HARNESS_OSSL_PREFIX")) != NULL) {
-        b = BIO_push(BIO_new(BIO_f_prefix()), b);
+            && (prefix = getenv("HARNESS_OSSL_PREFIX")) != NULL) {
+        b = BIO_push(check_bio(BIO_new(BIO_f_prefix()), "prefix stdout", 0),
+                     b);
         BIO_set_prefix(b, prefix);
     }
 
-    return b;
+    return check_bio(b, "stdout", mode);
```
bernd-edlinger, line 2408: this may be omitted, if you like.

apps/lib/apps.c
bernd-edlinger: I think the check is missing here.

apps/lib/apps.c
richsalz: Yeah, I know it can be omitted.  Kept for consistency.

apps/lib/apps.c
```diff
@@ -2370,41 +2366,60 @@ int raw_write_stdout(const void *buf, int siz)
...
+
 #ifdef OPENSSL_SYS_VMS
     if (FMT_istext(format))
-        b = BIO_push(BIO_new(BIO_f_linebuffer()), b);
+        b = BIO_push(check_bio(BIO_new(BIO_f_linebuffer()),
+                               "linebuffer stderr", 0),
+                     b);
 #endif
-    return b;
+    return check_bio(b, "stderr", mode);;
```
bernd-edlinger, line 2422: could be omitted, unless for clarity.

richsalz: Kept for clarity.

richsalz:
amended fixup pushed to catch missing check call.

richsalz:
are we having fun yet? :)

bernd-edlinger:
I hope that is not a show stopper,
but does this mean that the openssl command shell will exit
when the first file can't be read?

richsalz:
you mean interactive mode?  yes, it will exit.  that could be a show-stopper if folks think interactive mode is (still) important.  it could be fixed using setjmp/longjmp.

t8m:
I must say it is a showstopper for me unless the interactive mode is dropped completely.

Why using setjmp/longjmp is better than just keeping the if (bio==NULL) in individual apps?

Memory allocation errors are quite some different beast than failure to open a file in the sense that they are usually unrecoverable so for these immediate exit from an application is somewhat acceptable.

richsalz:
Since #12023 is on the fast track, although the body isn't even cold yet, I assume that there's no show-stoppers any more.

apps/lib/apps.c
bernd-edlinger: the check_bio is still missing here.

bernd-edlinger: (or in th SYS_VMS ifdef block)

richsalz: forgot to push.

t8m:
Yeah, no more show-stoppers. I still a little bit dislike this approach but I am definitely not blocking this.

apps/cmp.c
```diff
@@ -809,7 +809,11 @@ static int load_certs_preliminary(const char *file, STACK_OF(X509) **certs,
     int ret = 0;
 
     if (format == FORMAT_PKCS12) {
-        BIO *bio = bio_open_default(file, 'r', format);
+        FILE *fp = fopen(file, "r");
+        BIO *bio = NULL;
+
+        if (fp != NULL)
+            bio = BIO_new_fp(fp, format);
```
DDvO, line 816: If `BIO_new_fp(fp, format)` fails (because `BIO_new(BIO_s_file()` fails) then `fp` needs to be closed.

DDvO, line 816: Yes, such a change is needed.
The question would have not arisen if the review of #4930 had been finalized and the changes merged, because then the whole `load_certs_preliminary()` function would be gone.

apps/lib/apps.c
```diff
@@ -2370,41 +2366,60 @@ int raw_write_stdout(const void *buf, int siz)
...
-    BIO *b = BIO_new_fp(stdout,
-                        BIO_NOCLOSE | (FMT_istext(format) ? BIO_FP_TEXT : 0));
+    int mode = BIO_NOCLOSE | (FMT_istext(format) ? BIO_FP_TEXT : 0);
+    BIO *b = check_bio(BIO_new_fp(stdout, mode), "stdout", mode);
     void *prefix = NULL;
 
 #ifdef OPENSSL_SYS_VMS
     if (FMT_istext(format))
-        b = BIO_push(BIO_new(BIO_f_linebuffer()), b);
+        b = BIO_push(check_bio(BIO_new(BIO_f_linebuffer()),
```
DDvO, line 2396: If `BIO_new(BIO_f_linebuffer()` fails then, strictly speaking, `b` should be freed, 
but it does not matter in practice because the program is exited then anyway. Maybe add a comment on this?
The same comment applies also to more such cases below.

apps/lib/apps.c
```diff
@@ -2370,41 +2366,60 @@ int raw_write_stdout(const void *buf, int siz)
  * does impact behavior on some platforms, such as differentiating between
  * text and binary input/output on non-Unix platforms
  */
+static BIO *check_bio(BIO *bp, const char *what, int mode)
+{
+    if (bp == NULL) {
+        BIO_printf(bio_err, "%s: Could not open %s in mode 0x%x\n",
+                opt_getprog(), what, mode);
+        if (bio_err != NULL)
+            ERR_print_errors(bio_err);
```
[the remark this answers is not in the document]
richsalz: Hm, that's undocumented behavior, but okay.

apps/lib/apps.c
```diff
@@ -2370,41 +2366,60 @@ int raw_write_stdout(const void *buf, int siz)
  * does impact behavior on some platforms, such as differentiating between
  * text and binary input/output on non-Unix platforms
  */
+static BIO *check_bio(BIO *bp, const char *what, int mode)
```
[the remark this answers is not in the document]
richsalz: it can't be a char, because it's an bitmask (e.g., "BIO_F_ NOCLOSE|" is often or'd in).

apps/lib/apps.c
```diff
@@ -2370,41 +2366,60 @@ int raw_write_stdout(const void *buf, int siz)
...
+{
+    if (bp == NULL) {
+        BIO_printf(bio_err, "%s: Could not open %s in mode 0x%x\n",
+                opt_getprog(), what, mode);
```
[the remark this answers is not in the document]
richsalz: I disagree, the flags are a bitmask.

apps/lib/apps.c
richsalz: comment added to before check_bio.

richsalz:
Thanks for the review!  I just resolved comments where I agreed. Fixup commit pushed.

apps/lib/apps.c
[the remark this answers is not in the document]
DDvO: This `ERR_print_errors(NULL)` behavior is very widely used, also in `apps.c`.
Why not document it on this occasion.

apps/lib/apps.c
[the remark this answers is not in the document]
DDvO: Hmm, I guess you meant `BIO_NOCLOSE`, which happens to be `0x00`,
yet also `BIO_FP_TEXT` is often or'ed in, which is non-zero,
to be used as argument to
```
BIO *BIO_new_fp(FILE *stream, int flags);
```

I meanwhile found that in  `check_bio` you are mixing those BIO flags not only with the `fopen()` mode,
having values like `'r'`, but also with  `O_WRONLY`, optionally or'ed with `O_BINARY` etc.,
which is pretty unclean.

I recommend changing the function at least to
```
static BIO *check_bio(BIO *bp, const char *what, char mode, int flags);
```
to avoid confusion and for properly printing the fopen mode (if non-NUL).

apps/lib/apps.c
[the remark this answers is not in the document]
DDvO: It depends. In `check_bio(ret, mode == 'r' ? "stdin" : "stdout", mode);` you are using the parameter for the fopen mode, which is a `char`.

apps/lib/apps.c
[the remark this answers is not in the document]
DDvO: An alternative would be to rename the parameter to the less confusing
```
static BIO *check_bio(BIO *bp, const char *what, int flags);
```
and make sure that the function is called only with some flavor of flags but not with the fopen mode (such as `'r'`).

apps/lib/apps.c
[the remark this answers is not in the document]
richsalz: okay; additional commit pushed.

apps/lib/apps.c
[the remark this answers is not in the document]
richsalz: okay, done.  additional commit pushed.

apps/lib/apps.c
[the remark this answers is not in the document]
DDvO: Apparently you forgot to correct the indentation?
`util/format-check.pl` certainly would have caught this.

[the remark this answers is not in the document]
richsalz: indent fixed via fixup commit, thanks.

apps/lib/apps.c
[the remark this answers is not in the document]
DDvO: Thanks, but the confusing/wrong name `mode` still appears in the error message
and as local var name in `dup_bio_*()` and `app_load_config_quiet()`.

apps/lib/apps.c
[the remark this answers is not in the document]
DDvO: In the `check_bio(b, "stdout", format)` call in `bio_open_default()` 
the `flags` parameter is now abused in new way: to hold the format.

apps/lib/apps.c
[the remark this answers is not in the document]
richsalz: It's called "format" throughout the apps.  Changing it here would break the larger consistency.

apps/lib/apps.c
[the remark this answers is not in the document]
richsalz: I don't understand this comment.

apps/lib/apps.c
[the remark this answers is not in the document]
richsalz: commit pushed to fix the error message.

apps/lib/apps.c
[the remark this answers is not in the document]
DDvO: Having a look at
```
BIO *bio_open_default(const char *filename, char mode, int format)
{
    BIO *b;

    /* Using stdin or stdout? */
    if (filename == NULL || strcmp(filename, "-") == 0)
        return mode == 'r' ? dup_bio_in(format) : dup_bio_out(format);

    b = BIO_new_file(filename, modestr(mode, format));
    if (mode == 'r')
        return check_bio(b, "stdin", format);
    return check_bio(b, "stdout", format);
```
the last two lines call `check_bio(...,  ..., format)` where the last argument is not some type of flags as for the other calls, but the file contents format specifier, such at `FORMAT_PEM`, which is not a set of flags. 

To cure this new type of mix-up, I suggest replacing these two lines by 
```
check_bio(ret, mode == 'r' ? "stdin" : "stdout",
               mode == 'r' ? O_RDONLY : O_WRONLY);
```

apps/lib/apps.c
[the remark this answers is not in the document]
DDvO: > It's called "format" throughout the apps. Changing it here would break the larger consistency.

Not really - did you mean "mode"?

There are many correct uses of the names "mode" and "format" in (hitherto) apps.c:
```
static const char *modestr(char mode, int format)
static const char *modeverb(char mode)
static BIO *bio_open_default_(const char *filename, char mode, int format,
                              int quiet)
BIO *bio_open_default(const char *filename, char mode, int format)
BIO *bio_open_default_quiet(const char *filename, char mode, int format)

```

So far there has been only one function in apps.c that confusingly has been using the name "mode" where "flags" would have been appropriate:
```
BIO *bio_open_owner(const char *filename, int format, int private)
{
...
    mode = O_WRONLY;
#ifdef O_CREAT
    mode |= O_CREAT;
#endif
#ifdef O_TRUNC
    mode |= O_TRUNC;
#endif
    textmode = FMT_istext(format);
    if (!textmode) {
#ifdef O_BINARY
        mode |= O_BINARY;
#elif defined(_O_BINARY)
        mode |= _O_BINARY;
#endif
    }
...
        fd = open(filename, mode, 0600);

```
Clearly `mode` should better have been named `flags` in this function, 
not only to avoid confusion with `char mode` used elsewhere in the file,
but also for consistency with `int open(const char *pathname, int flags);`
See also, e.g., https://linux.die.net/man/3/open

apps/lib/apps.c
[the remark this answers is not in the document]
richsalz: But that change is *wrong*  It's the ```FORMAT_xxx``` flags.

apps/lib/apps.c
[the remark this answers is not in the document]
richsalz: I renamed mode to flags.

And I meant "format" in my comment above.  As in ```-outform DER``` and so on.

apps/lib/apps.c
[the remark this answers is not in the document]
DDvO: Sorry, but you still seem to confuse/equate the `int flags` parameter of
```
BIO *BIO_new_fp(FILE *stream, int flags);

BIO_new_fp() creates a file BIO wrapping stream. Flags can be: BIO_CLOSE, BIO_NOCLOSE (the close flag) BIO_FP_TEXT
       (sets the underlying stream to text mode, default is binary: this only has any effect under Win32).
```
and the `int format` parameter of many file read/write functions, being the output of
```
int opt_format(const char *s, unsigned long flags, int *result)
```
which are two somewhat related but still different things.

For instance, `BIO_FP_TEXT ==  0x10` while `FORMAT_TEXT == 0x8000`.

apps/lib/apps.c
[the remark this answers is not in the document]
DDvO: Nice. The same would be good also for 
- `int mode = BIO_FP_TEXT;`
- `static BIO *check_bio(BIO *bp, const char *what, int mode)`
- `int mode = BIO_NOCLOSE | (FMT_istext(format) ? BIO_FP_TEXT : 0);`  (3 times)

apps/lib/apps.c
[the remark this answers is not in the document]
richsalz: The issue, I guess, is that check_bio's diagnostic print is getting two different flags semantics.  So printing it as hex is good enough.  The alternative is having two/three check functions,  1 for fopen mode, 2 for open flags, 3 for BIO flags.  That's not worth doing in my view. Or to use the common phrase "for another PR"

apps/lib/apps.c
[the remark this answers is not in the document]
richsalz: See comment below about mode/format confusion and my unwillingness to fix it in this PR.

apps/lib/apps.c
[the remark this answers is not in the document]
DDvO: > The issue, I guess, is that check_bio's diagnostic print is getting two different flags semantics.

Right, `check_bio()` for the most part gets handed in BIO opening flags such as `BIO_NOCLOSE` (though they are still confusingly called 'mode' as mentioned in my other comment above) while via `bio_open_default()` at this point `check_bio()` gets handed in a format (i.e, file content type information such as `FORMAT_PKCS12`).

> So printing it as hex is good enough.

The diagnostic message `Could not open BIO(%s, 0x%x)` meanwhile lets it open how to interpret the hex value printed. This is not wrong but also not very helpful to the user because (s)he cannot know which way the value printed should be interpreted.

> The alternative is having two/three check functions, 1 for fopen mode, 2 for open flags, 3 for BIO flags.

The alternative I proposed would have been simpler and been to the point, namely to let the 
`check_bio()` parameter `flags` always be a BIO opening flag, and this could have been achieved simply by changing 
```
    if (mode == 'r')
        return check_bio(b, "stdin", format);
    return check_bio(b, "stdout", format);
```
to
```
check_bio(ret, mode == 'r' ? "stdin" : "stdout",
               mode == 'r' ? O_RDONLY : O_WRONLY);
```
Then the diagnostic message of `check_bio()` could be more informative and non-confusing: 
`Could not open BIO(filename=%s, flags=0x%x)`.

> That's not worth doing in my view. Or to use the common phrase "for another PR"

I agree that this can be sorted out later and is not worth the time to be discussed further in this PR.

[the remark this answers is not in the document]
richsalz: Opened https://github.com/openssl/openssl/issues/12056 to capture this.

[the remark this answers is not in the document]
DDvO: Fine

richsalz:
There was a merge conflict (new function app_bail_out) so I rebased and merged/squashed many of the commits.  Any new feedback comes in, I'll just push fixups :)

doc/man3/ERR_print_errors.pod
```diff
@@ -16,12 +16,15 @@ ERR_print_errors, ERR_print_errors_fp, ERR_print_errors_cb
 
 =head1 DESCRIPTION
 
-ERR_print_errors() is a convenience function that prints the error
+ERR_print_errors() prints the contents of the current
+error stack to the BIO B<bp>, and empties the error queue.
 strings for all errors that OpenSSL has recorded to B<bp>, thus
 emptying the error queue.
```
DDvO, line 22: These two leftover lines should be deleted.

richsalz: fixup commit pushed.

richsalz:
@bernd-edlinger back from vacation yet? :)

apps/cmp.c
```diff
@@ -809,7 +809,14 @@ static int load_certs_preliminary(const char *file, STACK_OF(X509) **certs,
     int ret = 0;
 
     if (format == FORMAT_PKCS12) {
-        BIO *bio = bio_open_default(file, 'r', format);
+        FILE *fp = fopen(file, "r");
```
levitte, line 812: `"r"` should be `"rb"`.  `FORMAT_PKCS12` is a binary format, and some platforms (such as Windows) need to know.

However, I fail to understand what was wrong with the `bio_open_default` call.  What's up with this special case?

apps/lib/apps.c
```diff
@@ -2377,42 +2374,61 @@ int raw_write_stdout(const void *buf, int siz)
  * and is therefore a show of intent more than anything else.  However, it
  * does impact behavior on some platforms, such as differentiating between
  * text and binary input/output on non-Unix platforms
+ *
+ * The routines below call check_bio() in case of failure, and therefore
+ * don't free up other resources on the error path because they know it will
+ * be handled when the process exits.
  */
+static BIO *check_bio(BIO *bp, const char *what, int mode)
```
levitte, line 2382: I've noticed that you're sending in `format` for mode in certain places.  They are not at all the same thing!
Also, I much prefer it if the error said text what was wanted with this particular BIO, i.e. pass a describing string, like the ones that `modeverb()` generated: "appending", "reading", "writing".  The error message could print like this:

``` C
        app_bail_out("%s: Could not open %s for %s\n",
                     opt_getprog(), what, reason);
```

apps/cmp.c
richsalz: The code there allows failure.  I was hoping @ddvo could answer your question, but he went off and gave detailed feedback on everything else :)

apps/lib/apps.c
richsalz: Yes, it's mixing number-spaces in a common error.  See the subthread at https://github.com/openssl/openssl/pull/11993#discussion_r434717452

apps/cmp.c
DDvO: I already commented on this above (https://github.com/openssl/openssl/pull/11993#discussion_r435013465):
> Yes, such a change was needed.
The question would have not arisen if the review of #4930 had been finalized and the changes merged, because then the whole load_certs_preliminary() function would be gone.

DDvO: To answer @levitte's new question with a little more detail: 
The change is needed (as long as #4930 is not merged) because `load_certs_autofmt()` may try loading certs via `load_certs_preliminary()` with two different formats, so the app must not terminate if the first try fails.

apps/lib/apps.c
levitte: Yeah, and you misunderstand.  The original error message would say something like this:

> Can't open foo.something for writing, ...

("writing" could as well be "reading" or "appending")

With this PR it's saying:

> prog: Could not open foo.something in mode <magicnumber>

Which of the two do you think is more useful to the user?  Words like "writing", "appending", "reading", or a magic number?

apps/lib/apps.c
```diff
@@ -2498,74 +2503,37 @@ BIO *bio_open_owner(const char *filename, int format, int private)
...
-{
-    return bio_open_default_(filename, mode, format, 1);
+    /* Using stdin or stdout? */
+    if (filename == NULL || strcmp(filename, "-") == 0)
+        return mode == 'r' ? dup_bio_in(format) : dup_bio_out(format);
+
+    b = BIO_new_file(filename, modestr(mode, format));
+    if (mode == 'r')
+        return check_bio(b, "stdin", format);
+    return check_bio(b, "stdout", format);
```
levitte, line 2536: Here, you pass `format` to `check_bio`.  Why?

DDvO: I was also complaining about the mixup of format, file mode, and BIO flags values, but @richsalz  did not want to fix this as part of this PR.
Yet actually this PR would be a better place to do it correctly right away.

DDvO: > Here, you pass `format` to `check_bio`. Why?

This issue has been postponed - see #12056.

apps/lib/apps.c
richsalz: Sure, I get it.  (I am the one who wrote and used fmt2str, for example).  The problem with now is that mode and format are overloaded, as discussed upthread.  Untangle those, then fix the error messages.

apps/lib/apps.c
richsalz: Because ```format``` is what was used in the BIO function.

apps/lib/apps.c
levitte: I have no idea why you would mix them in the first place.  If you want to be able to have both in this function (not sure I understand why), *then have them as two separate arguments*

richsalz: At first it was just a ```FMT_``` style argument.  Then during various reviews, others pointed out places where ```check_bio``` could be added, and it wasn't clear that this was an open(2) ```mode``` argument until other reviews pointed it out.

I'll get around to it shortly.

levitte: You know what, I think it'll be easier to simply show you how I mean it.

richsalz: You don't have to spend any effort on this unless you don't have anything better to do.  I know what you and David want.  I didn't want to do it as part of this PR, but I will when I get time.  Or you can just push commits to this PR if you prefer.

levitte:
I just added a fixup commit that I hope will settle the `check_bio` dispute

richsalz:
I don't think your commit addresses @ddvo's concern, but I will be happy if I am wrong.

I added another commit that replaces most "mode" variables with "bflags" when they are BIO flags.

apps/lib/apps.c
```diff
@@ -2433,8 +2433,8 @@ BIO *dup_bio_out(int format)
 
 BIO *dup_bio_err(int format)
 {
-    int mode = BIO_NOCLOSE | (FMT_istext(format) ? BIO_FP_TEXT : 0);
-    BIO *b = check_bio(BIO_new_fp(stderr, mode), "stderr", 'w', format);
+    int bflags = BIO_NOCLOSE | (FMT_istext(format) ? BIO_FP_TEXT : 0);
+    BIO *b = check_bio(BIO_new_fp(stderr, bflags), "stderr", 'w', format);
```
levitte, line 2437: `format` should be `bflags`

levitte:
Quite frankly, I'm not sure what use the BIO flags or the format number is for the user...  what are they supposed to do with it, go dive in header files?

apps/lib/apps.c
richsalz: replaced 3 occurrences; pushed fixup commit.

richsalz:
I err on the side of more diagnostic info, even if not useful to the end-user.

apps/lib/apps.c
```diff
@@ -2379,56 +2392,57 @@ int raw_write_stdout(const void *buf, int siz)
...
  * be handled when the process exits.
  */
-static BIO *check_bio(BIO *bp, const char *what, int mode)
+static BIO *check_bio(BIO *bp, const char *what, char mode, int bflags)
 {
     if (bp == NULL)
-        app_bail_out("%s: Could not open %s in mode 0x%x\n",
-                     opt_getprog(), what, mode);
+        app_bail_out("%s: Could not open %s for %s (BIO flags 0x%x)\n",
+                     opt_getprog(), what, modeverb(mode), bflags);
```
DDvO, line 2399: This is definitely an improvement along the lines I had asked for some two weeks back.

apps/lib/apps.c
```diff
@@ -75,7 +75,7 @@ typedef struct {
     unsigned long mask;
 } NAME_EX_TBL;
 
-static BIO *check_bio(BIO *bp, const char *what, int flags);
+static BIO *check_bio(BIO *bp, const char *what, char mode, int format);
```
DDvO, line 78: You are still mixing format and BIO flags - please separate them as well.
This won't be much work ;)

apps/lib/apps.c
```diff
@@ -2379,56 +2392,57 @@ int raw_write_stdout(const void *buf, int siz)
...
-                     opt_getprog(), what, mode);
+        app_bail_out("%s: Could not open %s for %s (BIO flags 0x%x)\n",
+                     opt_getprog(), what, modeverb(mode), bflags);
     return bp;
 }
 
 BIO *dup_bio_in(int format)
 {
-    int mode = BIO_NOCLOSE | (FMT_istext(format) ? BIO_FP_TEXT : 0);
+    int bflags = BIO_NOCLOSE | (FMT_istext(format) ? BIO_FP_TEXT : 0);
```
DDvO: Good that you also did such renamings, which avoid confusion.

DDvO:
> I don't think your commit addresses @DDvO's concern, but I will be happy if I am wrong.

Recent changes go in the right direction by separating the file opening mode (r/w/a) from other detail, but there is still a mix of file content format and BIO flags, which should be separated as well.

richsalz:
I pushed a commit that does the separation, at least what I think is needed.  If it's not right, please add a commit here.

DDvO:
> I pushed a commit that does the separation, at least what I think is needed. If it's not right, please add a commit here.

Thanks! Now the code is clean and the flags/format output is meaningful.

I would have expected that `check_bio()` would get both parameters (rather than adding a second function, `check_bio_fmt()`), which would be a bit less redundant and can be more informative in case both the flags and the format is present, but in case only one of them is available then the solution with two separate functions is of course more natural.

richsalz:

DDvO:
> https://mta.openssl.org/pipermail/openssl-project/2020-July/002091.html

@richsalz, are you sure you did not confuse issues?
What has #12089 to do with this PR?

DDvO:
It appears to me that this PR has been closed and branch deleted by mistake!
Can you restore it?

levitte:
@richsalz is absolutely free to withdraw his submissions as he sees fit and for whatever reason, even though we may find that unfortunate. He has stated his reasons in the openssl-users he linked.

DDvO:
Oh - meanwhile I noticed what the reason/background of this deletion is!
Sad that a merited contributor leaves, for an IMHO unessential reason.
And a waste of resources to kill PRs, on which also others have invested.

PR #11900, opened 2020-05-21
[WIP] Openssl 111 features

bernd-edlinger:
This contains 1.1.1 features I have
#2986 
#6516 
#11525 
#11577 
#11666 
#11780 
#11782 
#11846 
#11895 
#11934 
#11956 
#11995 
#12079

PR #11995, opened 2020-05-30
Remove getenv(OPENSSL_FIPS) in openssl command

bernd-edlinger:
This is left over from the past.

##### Checklist

### DIFF
--- a/apps/openssl.c
+++ b/apps/openssl.c
@@ -261,11 +261,6 @@ int main(int argc, char *argv[])
     setup_trace(getenv("OPENSSL_TRACE"));
 #endif
 
-    if (getenv("OPENSSL_FIPS")) {
-        BIO_printf(bio_err, "FIPS mode not supported.\n");
-        return 1;
-    }
-
     if (!apps_startup()) {
         BIO_printf(bio_err,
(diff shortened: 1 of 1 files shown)


bernd-edlinger:
My concern about this code, is that it might be re-used for the 3.0 FIPS provider,
and then in consequence be set in environment variables, which will then
1.1.1 fail unexpectedly.

paulidale:
I'm unsure about 1.1.1 for this.

bernd-edlinger:
Why, this is no documented behavior, and it plays a prank on the user.

levitte:
This environment variable was most relevant with the FIPS OpenSSL Module 2.0 (compatible with OpenSSL 1.0.2).  The reason we have the error is for users that have upgrade to the 1.1.x series must not be lead to believe that OPENSSL_FIPS still has the meaning they might think.

So I see nothing against removing this in 3.0, we have moved far enough from 1.0.2 to simply drop the check.  For 1.1.1, I'm just as dubious as @paulidale.

bernd-edlinger:
Okay, then I'll add this to my 1.1.1 feature branch #11900

bernd-edlinger:
Merged to master as 32df134

PR #11997, opened 2020-05-30
mac: make function naming consistent.

paulidale:
EVP_MAC_CTX functions have been renamed to the EVP_MAC namespace for consistency.
The EVP_MAC header file is now separate from evp.h but still included by evp.h

Before 3.0 is released, the names are mutable and this prevents more inconsistencies being introduced.

There are no functional or code changes. Just the renaming and a little reformatting.


levitte:
I wholeheartedly disagree with this.

richsalz:
Can you explain why you disagree?  I mean I understand the argument for consistency (see also #11512), but I don't see the argument for *not* renaming things.

levitte:
I'll gladly explain.  It's a bit long, but this is a dispute that has been going on for a while behind the scene, and has got me thinking about all this for quite a while.

### Already established patterns

First of all, we already have an established pattern, which you can see with the functions surrounding `EVP_CIPHER`, `EVP_DIGEST`, `EVP_PKEY`, that functions that have `_CTX` in their name, are all constructors, destructors, direct setters and direct getters of the associated context structure, while those without `_CTX` in the name revolve around the operation they're supposed to perform (they *use* the context structure, but they aren't constructors, destructors, direct setters or direct getters).

This pattern is age old, and what our users are used to see.

### Meaning of names

I knew I'd get to see this: `EVP_MAC_CTX_size()`

What's that, the size of the context structure?  That's at least the first thing that come to my mind.  And while one might argue that `EVP_MAC_size()` could be understood as the size of the `EVP_MAC` structure, I'd think that's less likely, among others considering that there's no constructor for it, only a fetcher...  and we have earlier patterns to lean on (`EVP_MD_size()`, for example)

### So what about consistent function name prefixes?

While I've claimed that we have already established patterns, I see no reason to stop thinking about alternatives, and I absolutely hear the desire for name consistency.  An alternative to sticking `_CTX` everywhere to the point of confusion, would be to remove it!  My reasoning here is that the functions revolve around an operation that we want to perform, not around its context - the context is just a carrier of information - so for the EVP_MAC interface, if we choose to break with established patterns, here's what I suggest (this is a rewrite of what's currently in include/openssl/evp.h):

``` C
EVP_MAC *EVP_MAC_fetch(OPENSSL_CTX *libctx, const char *algorithm,
                       const char *properties);
int EVP_MAC_up_ref(EVP_MAC *mac);
void EVP_MAC_free(EVP_MAC *mac);
int EVP_MAC_number(const EVP_MAC *mac);
int EVP_MAC_is_a(const EVP_MAC *mac, const char *name);
const OSSL_PROVIDER *EVP_MAC_provider(const EVP_MAC *mac);
int EVP_MAC_get_params(EVP_MAC *mac, OSSL_PARAM params[]);

EVP_MAC_CTX *EVP_MAC_new_ctx(EVP_MAC *mac);
void EVP_MAC_free_ctx(EVP_MAC_CTX *ctx);
EVP_MAC_CTX *EVP_MAC_dup_ctx(const EVP_MAC_CTX *src);
EVP_MAC *EVP_MAC_get_ctx_mac(EVP_MAC_CTX *ctx);
int EVP_MAC_get_ctx_params(EVP_MAC_CTX *ctx, OSSL_PARAM params[]);
int EVP_MAC_set_ctx_params(EVP_MAC_CTX *ctx, const OSSL_PARAM params[]);

size_t EVP_MAC_size(EVP_MAC_CTX *ctx);
int EVP_MAC_init(EVP_MAC_CTX *ctx);
int EVP_MAC_update(EVP_MAC_CTX *ctx, const unsigned char *data, size_t datalen);
                  unsigned char *out, size_t *outl, size_t outsize);
const OSSL_PARAM *EVP_MAC_gettable_params(const EVP_MAC *mac);
const OSSL_PARAM *EVP_MAC_gettable_ctx_params(const EVP_MAC *mac);
const OSSL_PARAM *EVP_MAC_settable_ctx_params(const EVP_MAC *mac);

void EVP_MAC_do_all_provided(OPENSSL_CTX *libctx,
                             void (*fn)(EVP_MAC *mac, void *arg),
                             void *arg);
void EVP_MAC_names_do_all(const EVP_MAC *mac,
                          void (*fn)(const char *name, void *data),
                          void *data);

```

This also makes it more legible in English, when "ctx" becomes a noun (or adjective?) in the phrase that the function name expresses.

levitte:
Let's remember that there's no perfect name scheme, but there's nuances that can be termed "better" and "worse", and my claim is that sticking in `_CTX` is "worse" than the currently established pattern, and removing it is "better" (or at least not "worse") if we are to change the pattern.

levitte:
(as for #11512, I consider that a different bullet that's worth its own discussion)

paulidale:
The _already established patterns_ is a furphy.  I'll reword it more precisely: _we've done things badly in the past, thus we should do them badly moving forwards_.  We've always been at war with Eurasia.....oh wrong context, it's Eastasia.

I would be willing to accept the _consistent function name prefixes_ option.  I.e. get rid of the EVP_*_CTX functions completely.  They'll be a few "fun" name conversions I suspect but this seems reasonable.

I agree, #11512 is a different matter.

levitte:
I assume that with getting rid of the `EVP_*_CTX` functions, you mean rename them to names without `_CTX`. We seem to agree.

Mind you, the more I think of it, the more I like the idea of breaking the old pattern, 'cause I agree that it's, er, not the best. My biggest reason for this dispute is that I really don't want a change to something worse. 

Question is, should we add corresponding aliases for the older interfaces, to align them with a redesigned pattern? I would argue that we should at least give it some thought.

levitte:
Side note: thanks for expanding my very little knowledge of Oz slang.  "Furphy" was new to me :-)

levitte:
There are a couple of details in my proposed modified EVP_MAC interface that I feel unsure about, and it's sticking `_ctx` in so many places.  In other words, these:

``` C
EVP_MAC *EVP_MAC_get_ctx_mac(EVP_MAC_CTX *ctx);
int EVP_MAC_get_ctx_params(EVP_MAC_CTX *ctx, OSSL_PARAM params[]);
int EVP_MAC_set_ctx_params(EVP_MAC_CTX *ctx, const OSSL_PARAM params[]);
```

could as well be:

``` C
EVP_MAC *EVP_MAC_get_mac(EVP_MAC_CTX *ctx);
int EVP_MAC_get_params(EVP_MAC_CTX *ctx, OSSL_PARAM params[]);
int EVP_MAC_set_params(EVP_MAC_CTX *ctx, const OSSL_PARAM params[]);
```

Question is what to make of these:

``` C
int EVP_MAC_get_params(EVP_MAC *mac, OSSL_PARAM params[]);
const OSSL_PARAM *EVP_MAC_gettable_params(const EVP_MAC *mac);
```

I could see folding them and simply stating that you'll get any parameters via the context, even those that are "class attributes" rather than "instance attributes".  I could also see renaming them to, er, something else.

richsalz:
Thanks for the interesting discussion.

t8m:
I think these are OK.
```
EVP_MAC *EVP_MAC_get_ctx_mac(EVP_MAC_CTX *ctx);
int EVP_MAC_get_ctx_params(EVP_MAC_CTX *ctx, OSSL_PARAM params[]);
int EVP_MAC_set_ctx_params(EVP_MAC_CTX *ctx, const OSSL_PARAM params[]);
```
I do not think we necessarily have to eliminate the `ctx` from the names completely.

paulidale:
Updated to match the suggested naming convention.

paulidale:
For the record, the rename command used was:
```bash
sed -i -e 's/EVP_MAC_CTX_new/EVP_MAC_new_ctx/g'               \
       -e 's/EVP_MAC_CTX_free/EVP_MAC_free_ctx/g'             \
       -e 's/EVP_MAC_CTX_dup/EVP_MAC_dup_ctx/g'               \
       -e 's/EVP_MAC_CTX_mac/EVP_MAC_get_ctx_mac/g'           \
       -e 's/EVP_MAC_CTX_get_params/EVP_MAC_get_ctx_params/g' \
       -e 's/EVP_MAC_CTX_set_params/EVP_MAC_set_ctx_params/g' \
    `egrep -rl 'EVP_MAC_CTX_new|EVP_MAC_CTX_free|EVP_MAC_CTX_dup|EVP_MAC_CTX_mac|EVP_MAC_CTX_get_params|EVP_MAC_CTX_set_params'`
```

crypto/err/openssl.txt
```diff
@@ -843,8 +843,8 @@ EVP_F_EVP_KEYEXCH_FETCH:245:EVP_KEYEXCH_fetch
 EVP_F_EVP_KEYEXCH_FROM_DISPATCH:244:evp_keyexch_from_dispatch
 EVP_F_EVP_MAC_CTRL:209:EVP_MAC_ctrl
 EVP_F_EVP_MAC_CTRL_STR:210:EVP_MAC_ctrl_str
-EVP_F_EVP_MAC_CTX_DUP:211:EVP_MAC_CTX_dup
-EVP_F_EVP_MAC_CTX_NEW:213:EVP_MAC_CTX_new
+EVP_F_EVP_MAC_CTX_DUP:211:EVP_MAC_dup_ctx
+EVP_F_EVP_MAC_CTX_NEW:213:EVP_MAC_new_ctx
```
levitte, line 847: Considering that you removed them in code, feel free to just remove these lines.

include/openssl/evp_mac.h
```diff
@@ -0,0 +1,52 @@
+/*
```
levitte, line 1: About this file name, wouldn't `mac.h` be enough?  Other headers have paved the way already, such as `kdf.h`, and I really don't see a need to have `evp_` in the header file name.

levitte:
Apart from a couple of nits, this looks good

include/openssl/evp_mac.h
paulidale: Can do.

paulidale:
Merged, thanks all for the input.

mattcaswell:
Sorry, I didn't look at this PR before. I find the new names totally surprising and confusing.

levitte:
> Sorry, I didn't look at this PR before. I find the new names totally surprising and confusing.

You should have seen what the first incarnation of this PR was like...

mattcaswell:
I think we should revert this. This makes things *less* consistent IMO (unlike what the name of this PR suggests)

paulidale:
Blame @levitte :)  Originally, it was pretty different.

I'm fine reverting this BTW, it's another of pass using _sed_.

mattcaswell:
Revert in #12186.

PR #12001, opened 2020-05-30
Undeprecate load locations

t-j-h:
##### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -542,13 +542,6 @@ OpenSSL 3.0
    - SSL_CTX_load_verify_dir()
    - SSL_CTX_load_verify_store()
 
-   Also, the following functions are now deprecated:
-
-   - X509_STORE_load_locations() (use X509_STORE_load_file(),
-     X509_STORE_load_path() or X509_STORE_load_store() instead)
-   - SSL_CTX_load_verify_locations() (use SSL_CTX_load_verify_file(),
-     SSL_CTX_load_verify_dir() or SSL_CTX_load_verify_store() instead)
-
    *Richard Levitte*
 
  * Added a new method to gather entropy on VMS, based on SYS$GET_ENTROPY.
--- a/crypto/x509/x509_d2.c
+++ b/crypto/x509/x509_d2.c
@@ -73,8 +73,6 @@ int X509_STORE_load_store(X509_STORE *ctx, const char *uri)
     return 1;
 }
 
-/* Deprecated */
-#ifndef OPENSSL_NO_DEPRECATED_3_0
 int X509_STORE_load_locations(X509_STORE *ctx, const char *file,
                               const char *path)
 {
@@ -86,4 +84,3 @@ int X509_STORE_load_locations(X509_STORE *ctx, const char *file,
         return 0;
     return 1;
 }
-#endif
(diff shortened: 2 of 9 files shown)


mattcaswell:
There was a CHANGES entry related to this change which is now misleading. I think the references to the deprecation should be removed from that entry.

t-j-h:
I don't think it is appropriate to ever edit a CHANGES entry - I could add a new one indicating the undoing of the deprecation if you would like - but editing the change history seems like a slippery slope to go down.

mattcaswell:
> I don't think it is appropriate to ever edit a CHANGES entry - I could add a new one indicating the undoing of the deprecation if you would like - but editing the change history seems like a slippery slope to go down.

I totally disagree. If it were in a published release I would agree. Once published the CHANGES record should be an accurate reflection of all the changes that have occurred since the last release. People should be able to read that list to gain an understanding of important differences between releases. Contradictory entries would be very confusing. The CHANGES document is *not* a record of the history of the development. That is what the git log is for. It is a list of all the CHANGES since the last release. If we alter those changes before we release then we should absolutely feel free to edit the CHANGES document.

t-j-h:
Commit added - wasn't thinking correctly about the CHANGES context.

levitte:
I frankly don't understand why we should maintain the functions that are replaced.  Do we want to retain them "permanently"?  Why?

Deprecation is, after all, only a marker that things will eventually go away, with a compiler warning thereof.  NOT deprecating can be seen as a promise that the old functions will continue to be maintained "indefinitely".

t-j-h:
Because we have *users* who use those APIs and get zero benefit from changing from one name to a slightly different one as views change over time. It was documented and the functionality remains.

Depreciation warnings should be for things that need to change where users need to migrate. 

Swapping one name for another and adding a NULl check around the new version is a pointless exercise for the *user*. 

This one is unlike the majority of the others that have been deprecated where we want the users to move off to a better alternative so we can clean up or restructure or get users out of out internals. This certainly isn't that situation.

levitte:
> Swapping one name for another

There's a *little bit more* than a mere name change in this API change.

> This one is unlike the majority of the others that have been deprecated where we want the users to move off to a better alternative so we can clean up or restructure or get users out of out internals. This certainly isn't that situation.

And this is where you and I disagree.  I do think that the itemized functions are the better alternative, and very clearly are the *only* choice as soon as new location forms (such as a URI for use with `OSSL_STORE` functionality) is to be passed.

t-j-h:
I have no issue with adding the itemised functions. But deprecating something that is widely used is not the right approach - this is not what we do to our users without a very good reason - and that isn't this case. If you want other functionality you have to call other functions - that too is a given. It doesn't mean you force users to change to calling two other functions simply because there are now other options which they user would also have to call.

levitte:
So pile the functions on, keep the "replaced" ones permanently (as far as we can see for the moment), is that the basic message?  ...  within the scope of these fairly simple things, there are other parts of OpenSSL where I assume we can justify the need much more strongly.

Boy, this will be messy, but ok, I can align with that line of thinking.  There are bigger fish :wink:

levitte:
.... and close

Issue #12007, opened 2020-06-01
More bytes are generated for Key Material than needed for AEAD ciphers


maximmasiutin:
### Bug description

There is a bug that, when generating the Key Material (key_block), OpenSSL generates more data via the PRF than needed for AEAD ciphers.

Please find tls1_setup_key_block()  function in the "t1_enc.c" file.

There is the following code:

    num = EVP_CIPHER_key_length(c) + mac_secret_size + EVP_CIPHER_iv_length(c);
    num *= 2;

later it calls tls1_generate_key_block() to generate “num” of pseudo-random bytes. However, for AEAD ciphers like AES-GCM, the resulting "num" is larger than needed.

Besides that, the order of addition “EVP_CIPHER_key_length(c) + mac_secret_size + EVP_CIPHER_iv_length(c)” is misleading, because according the RFC, the MAC secret key comes first, not second. An easier to understand code would be:

    num = mac_key_length + EVP_CIPHER_key_length(c) + EVP_CIPHER_iv_length(c);

You can add a line like the following

        BIO_printf(trc_out, "key block length: %ld\n", num);

to actually see that more key bytes are generated then needed. 

If you wish to see which of the constituents lead to a decreased key block length, add the line like the following:

        BIO_printf(trc_out, "key block length: %ld (mac_secret_size: %ld, EVP_CIPHER_key_length: %ld, EVP_CIPHER_iv_length: %ld)\n", num, mac_secret_size, EVP_CIPHER_key_length(c), EVP_CIPHER_iv_length(c));

### Steps to reproduce

To reproduce this bug and see it actually, add the above "BIO_printf" line to t1_enc.c after the first entry of OSSL_TRACE_BEGIN(TLS) in tls1_setup_key_block and compile OpenSSL with tracing enabled, and run a TLS 1.2 session with AES256-GCM-SHA384.

When building OpenSSL, add the "enable-trace" parameter to "Configure". For example, for 64-bit Windows, call:

    perl Configure enable-trace enable-ssl-trace VC-WIN64A
    nmake

    set OPENSSL_TRACE=TLS
    openssl.exe s_server

On client, run the following:

openssl.exe s_client -host 127.0.0.1 -port 4433 -tls1_2 -cipher AES256-GCM-SHA384

It will start accepting TLS connections on port 4433. Connect to it with your client with any AEAD cipher suite like AES_GCM and you will see on your console trace data like this:

    TRACE[08:26:00:00]:TLS: Premaster Secret:
[...]

PR #12008, opened 2020-06-01
This fix shows actual Key Block size bug

maximmasiutin:
This fix shows actual Key Block size when tracing to help demonstrate the bug described in https://github.com/openssl/openssl/issues/12007

[x] Tracing shows Key Block size


maximmasiutin:
Could you please consider this commit "Trivial", to save from filing the agreement?

PR #12013, opened 2020-06-02
undeprecate EVP_PKEY_cmp and EVP_PKEY_cmp_parameters

t-j-h:
##### Checklist

### DIFF
--- a/doc/man3/EVP_PKEY_copy_parameters.pod
+++ b/doc/man3/EVP_PKEY_copy_parameters.pod
@@ -46,6 +46,9 @@ Since OpenSSL private keys contain public key components too the function
 EVP_PKEY_eq() can also be used to determine if a private key matches
 a public key.
 
+EVP_PKEY_cmp() and EVP_PKEY_cmp_parameters() differ in their return values 
+compared to other _cmp() functions.
+
 =head1 RETURN VALUES
 
 The function EVP_PKEY_missing_parameters() returns 1 if the public key
@@ -55,7 +58,8 @@ doesn't use parameters.
 These functions EVP_PKEY_copy_parameters() returns 1 for success and 0 for
 failure.
 
-The function EVP_PKEY_parameters_eq() and EVP_PKEY_eq() return 1 if their
+The functions EVP_PKEY_cmp_parameters(), EVP_PKEY_parameters_eq(), 
+EVP_PKEY_cmp() and EVP_PKEY_eq() return 1 if their
 inputs match, 0 if they don't match, -1 if the key types are different and
 -2 if the operation is not supported.
 
--- a/doc/man3/EVP_PKEY_copy_parameters.pod
+++ b/doc/man3/EVP_PKEY_copy_parameters.pod
@@ -47,7 +47,9 @@ EVP_PKEY_eq() can also be used to determine if a private key matches
 a public key.
 
 EVP_PKEY_cmp() and EVP_PKEY_cmp_parameters() differ in their return values 
-compared to other _cmp() functions.
+compared to other _cmp() functions. They are aliases for EVP_PKEY_eq() and 
+EVP_PKEY_parameters_eq() functions provided for backwards compatibility 
+with existing applications.
 
 =head1 RETURN VALUES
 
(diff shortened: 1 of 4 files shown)


t8m:
I strongly believe this undeprecation is wrong. The function name is seriously confusing due to the return value difference from all the other _cmp functions. This is very much different from other cases.

t-j-h:
It is a long standing API. 
It doesn't follow the same pattern - but it also isn't sensible to require users to change their code to a new name. 
Documentation should warn of the difference and we should stop using it ourselves - but depreciation isn't the right answer for this sort of thing. We should reserve depreciation for when there is a clear need to move users off an interface.

This isn't one of those circumstances - we just poorly named the functions and didn't follow an undocumented pattern. That is unfortunate. But it doesn't mean we should force exisiting users to change their code to get warning free 3.0 builds.

beldmit:
> I strongly believe this undeprecation is wrong. The function name is seriously confusing due to the return value difference from all the other _cmp functions. This is very much different from other cases.

I totally agree with @t8m

levitte:
> ... undocumented pattern

Depending on how broadly you look at the pattern, it's not that undocumented.  Broadly speaking (not just staying withing OpenSSL) "cmp" is part of function names that all have the same semantics, they return \< 0, 0 or \> 0.  I believe this is so deeply ingrained in most programmers' minds that we assume those semantics unconsciously, even when we've read the documentation of these particular functions (and we who work with OpenSSL are no exception).

doc/man3/EVP_PKEY_copy_parameters.pod
```diff
@@ -55,7 +58,8 @@ doesn't use parameters.
 These functions EVP_PKEY_copy_parameters() returns 1 for success and 0 for
 failure.
 
-The function EVP_PKEY_parameters_eq() and EVP_PKEY_eq() return 1 if their
+The functions EVP_PKEY_cmp_parameters(), EVP_PKEY_parameters_eq(), 
+EVP_PKEY_cmp() and EVP_PKEY_eq() return 1 if their
```
t8m, line 62: Please drop this hunk. We should make the _eq as the preferred name.

doc/man3/EVP_PKEY_copy_parameters.pod
```diff
@@ -46,6 +46,9 @@ Since OpenSSL private keys contain public key components too the function
 EVP_PKEY_eq() can also be used to determine if a private key matches
 a public key.
 
+EVP_PKEY_cmp() and EVP_PKEY_cmp_parameters() differ in their return values 
+compared to other _cmp() functions.
+
```
t8m, line 51: You could add here something like "....cmp and cmp_parameters" functions differ in their return values 
compared to other _cmp() functions, they are just aliases for the _eq and parameters_eq functions provided for backwards compatibility with existing applications."

richsalz:
> But it doesn't mean we should force exisiting users to change their code to get warning free 3.0 builds.

I thought they *DO* get warning-free builds unless they configure with special flag.  Am I wrong?

kaduk:
> But it doesn't mean we should force exisiting users to change their code to get warning free 3.0 builds.

I'm not sure I follow all the steps from the premise "3.0 will be 99% API-compatible with 1.1.1" to the conclusion that users need to specifically have a *warning-free* build with 3.0.  Warnings are not errors, and can be suppressed.  If you're building your own openssl you'll need to change your build system to get the new openssl and can apply `-Wno-error` at that time; if you get your openssl from the system the OS is going to give you lots of compat lead time across a SONAME transition.

Apologies if this was already discussed and I missed/forgot about it.

doc/man3/EVP_PKEY_copy_parameters.pod
t-j-h: I have added a new commit adjusting the documentation. Please confirm this is in line with your suggested update.

t8m: Perfect now.

t8m:
@DDvO would you please submit a new PR which would mention the peculiarity of EVP_PKEY_cmp and the new EVP_PKEY_eq in CHANGES.md?

DDvO:
> @DDvO would you please submit a new PR which would mention the peculiarity of EVP_PKEY_cmp and the new EVP_PKEY_eq in CHANGES.md?

All right, done in #12026.

kroeckx:
@t-j-h: Any reason this didn't get merged yet?

levitte:
Knowing that @t-j-h's time can be quite precious, nothing stops anyone else from doing the merge.
I'll do it now

levitte:
Merged

eb750219f2ab7886f174a071f880b31cedeeb0a8 undeprecate EVP_PKEY_cmp and EVP_PKEY_cmp_parameters

PR #12016, opened 2020-06-02
Fix a buffer overflow in drbg_ctr_generate

### LINKED ISSUE #12012: heap buffer overflow in random generation
https://oss-fuzz.com/testcase-detail/5444665703923712 reports an unreproducible problem with a reading past the end of a buffer. For OSS-Fuzz it's not reproducible because it's all with random numbers.

The problem is that bnrand() does:
https://github.com/openssl/openssl/blob/32df13497a1f1d7ef90a9c0b3128f65d18ab0086/crypto/bn/bn_rand.c#L43

And then:
https://github.com/openssl/openssl/blob/32df13497a1f1d7ef90a9c0b3128f65d18ab0086/crypto/bn/bn_rand.c#L50-L51

This buffer, in the example 8 bytes large, makes it's way to drbg_ctr_generate: https://github.com/openssl/openssl/blob/32df13497a1f1d7ef90a9c0b3128f65d18ab0086/crypto/rand/drbg_ctr.c#L376

This is clearly an incorrect use of the `out` buffer that's not large enough for the data that needs to be generated. It looks like in some cases `buflen` is the larger than `outlen`, the next multiple of the blocksize, while the buffer isn't that large. In the other cases where buflen == outlen, it's misusing the API, it will try to write a multiple of whole blocks and so can again be too small. 

In 1.1.1 we had a special buffer for requests smaller than 16 bytes, but that got removed in commit 28bdbe1aaa4.

bernd-edlinger:
This can happen if the 32-bit counter overflows
and the last block is not a multiple of 16 bytes.

Fixes #12012

[extended tests]

##### Checklist

### DIFF
--- a/crypto/rand/drbg_ctr.c
+++ b/crypto/rand/drbg_ctr.c
@@ -369,8 +369,8 @@ __owur static int drbg_ctr_generate(RAND_DRBG *drbg,
             if (ctr32 != 0) {
                 blocks -= ctr32;
                 buflen = blocks * 16;
+                ctr32 = 0;
             }
-            ctr32 = 0;
             ctr96_inc(ctr->V);
         }
         PUTU32(ctr->V + 12, ctr32);
--- a/crypto/rand/drbg_ctr.c
+++ b/crypto/rand/drbg_ctr.c
@@ -366,8 +366,10 @@ __owur static int drbg_ctr_generate(RAND_DRBG *drbg,
         ctr32 = GETU32(ctr->V + 12) + blocks;
         if (ctr32 < blocks) {
             /* 32-bit counter overflow into V. */
-            blocks -= ctr32;
-            buflen = blocks * 16;
+            if (ctr32 != 0) {
+                blocks -= ctr32;
+                buflen = blocks * 16;
+            }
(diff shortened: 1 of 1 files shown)


crypto/rand/drbg_ctr.c
```diff
@@ -366,8 +366,10 @@ __owur static int drbg_ctr_generate(RAND_DRBG *drbg,
         ctr32 = GETU32(ctr->V + 12) + blocks;
         if (ctr32 < blocks) {
             /* 32-bit counter overflow into V. */
-            blocks -= ctr32;
-            buflen = blocks * 16;
+            if (ctr32 != 0) {
+                blocks -= ctr32;
+                buflen = blocks * 16;
+            }
             ctr32 = 0;
```
t8m, line 373: Should the ctr32 = 0 be moved inside the if?

bernd-edlinger: yes. good point.

bernd-edlinger:
Merged to master               42fa3e6
OpenSSL_1_1_1-stable e6a80cb
Thanks!

PR #12017, opened 2020-06-02
Consolidate doc of BIO_do_connect() and its alias BIO_do_handshake()

DDvO:
As suggested by @richsalz https://github.com/openssl/openssl/pull/11971#issuecomment-636923583 this documents that these functions meanwhile try all IP addresses resolved for a given domain name.

Doing so I found that the documentation of these two functions, although they are just aliases for the same `BIO_ctrl(b,BIO_C_DO_STATE_MACHINE,0,NULL)` call, was differing.
This PR includes my suggestion for consolidating their description.

Should we explicitly mention that `BIO_do_connect()` and `BIO_do_handshake()` are essentially the same?

##### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -158,7 +158,8 @@ OpenSSL 3.0
    *David von Oheimb*
 
  * BIO_do_connect and BIO_do_handshake have been extended:
-   if domain name resolution yields multiple IP addresses all of them are tried.
+   If domain name resolution yields multiple IP addresses all of them are tried
+   after connect() failures.
 
    *David von Oheimb*
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -157,6 +157,11 @@ OpenSSL 3.0
 
    *David von Oheimb*
 
+ * BIO_do_connect and BIO_do_handshake have been extended:
+   if domain name resolution yields multiple IP addresses all of them are tried.
+
+   *David von Oheimb*
+
  * All of the low level RSA functions have been deprecated including:
 
    RSA_new_method, RSA_size, RSA_security_bits, RSA_get0_pss_params,
--- a/doc/man3/BIO_f_ssl.pod
+++ b/doc/man3/BIO_f_ssl.pod
@@ -98,7 +98,8 @@ pointer.
 BIO_do_handshake() attempts to complete an SSL handshake on the
 -supplied BIO and establish the SSL connection.
 For non-SSL BIOs the connection is done typically at TCP level.
-If domain name resolution yields multiple IP addresses all of them are tried.
+If domain name resolution yields multiple IP addresses all of them are tried
+after connect() failures.
 The function returns 1 if the connection was established successfully.
 A zero or negative value is returned if the connection could not be established.
 The call BIO_should_retry() should be used for non-blocking connect BIOs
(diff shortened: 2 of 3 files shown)


levitte:
> Should we explicitly mention that BIO_do_connect() and BIO_do_handshake() are essentially the same?

No. That's an implementation detail, and it may change. 

Also, in describing these two calls, you should look at intended use more than implementation details.

richsalz:
Hmm.
```
#  define BIO_do_connect(b)       BIO_do_handshake(b)
#  define BIO_do_accept(b)        BIO_do_handshake(b)
```

Perhaps the talks and say that the accept and connect are currently the same as do-handshake.

I still think the change in behavior should be in CHANGES.

DDvO:
So I've restored the original doc intro sentence on `BIO_do_handshake()` since it is meant for SSL.

I've meanwhile mentioned the extension of `BIO_do_connect()` and `BIO_do_handshake()` also in `CHANGES.md`.

CHANGES.md
```diff
@@ -157,6 +157,11 @@ OpenSSL 3.0
 
    *David von Oheimb*
 
+ * BIO_do_connect and BIO_do_handshake have been extended:
+   if domain name resolution yields multiple IP addresses all of them are tried.
```
t8m, line 161: Isn't if -> If more correct gramatically?
Also I would like to see something like: "If domain name resolution yields multiple IP addresses all of them are tried after connect() failures." or "until some succeeds." or "before returning an error."

doc/man3/BIO_f_ssl.pod
```diff
@@ -96,12 +96,14 @@ chain and calling SSL_shutdown() on its internal SSL
...
 BIO_do_handshake() attempts to complete an SSL handshake on the
-supplied BIO and establish the SSL connection. It returns 1
-if the connection was established successfully. A zero or negative
-value is returned if the connection could not be established, the
-call BIO_should_retry() should be used for non blocking connect BIOs
-to determine if the call should be retried. If an SSL connection has
-already been established this call has no effect.
+-supplied BIO and establish the SSL connection.
+For non-SSL BIOs the connection is done typically at TCP level.
+If domain name resolution yields multiple IP addresses all of them are tried.
```
t8m, line 101: Same here.

CHANGES.md
DDvO: > Isn't if -> If more correct grammatically?

It turns out that it depends on BE vs. AE and likely has become a matter of taste:
https://www.grammarly.com/blog/capitalization-after-colons/

BTW, which is the preferred flavor of English in OpenSSL these days? 
Looks like it used to be British, but in general American seems to dominate also the IT world.

DDvO: > Also I would like to see something like: "If domain name resolution yields multiple IP addresses all of them are tried after connect() failures."

Done, also for the other two occurrences.

doc/man3/BIO_f_ssl.pod
DDvO: Yep

CHANGES.md
DDvO: @t8m, is this fine now?

DDvO:
Merged - thanks!

PR #12019, opened 2020-06-02
[crypto/ec] Remove unreachable AVX2 code in NISTZ256 implementation

romen:
`crypto/ec/ecp_nistz256.c` contains code sections guarded by a `ECP_NISTZ256_AVX2` define.

The relevant comment reads:

>     /*
>      * Note that by default ECP_NISTZ256_AVX2 is undefined. While it's great
>      * code processing 4 points in parallel, corresponding serial operation
>      * is several times slower, because it uses 29x29=58-bit multiplication
>      * as opposite to 64x64=128-bit in integer-only scalar case. As result
>      * it doesn't provide *significant* performance improvement. Note that
>      * just defining ECP_NISTZ256_AVX2 is not sufficient to make it work,
>      * you'd need to compile even asm/ecp_nistz256-avx.pl module.
>      */

Without diminishing the quality of the original submission, it's evident that this code has been basically unreachable without modifications to the library source code and is under-tested.

This PR removes these sections from the codebase.

This is an alternative to #10616 .

### Branches

This PR is against `master`.

This can in theory apply, probably cleanly, to 1.1.1 as well: yet I prefer to open a separate dedicated PR about it if this gets accepted as, although unreachable, I would like an OMC explicit approval about removing code from the stable LTS branch.


romen:
@bernd-edlinger @mattcaswell could you comment on this?

Ping @bbbrumley that was also involved in the #10616 discussion that triggered this.

t8m:
I really think this is no-brainer even on 1.1.1 branch and does not require formal approval. The removed code is broken and unused. There is really no point in keeping it.

If someone downstream uses that code and patches the bugs out they can as well re-add the code back. But I think the probability of such downstream use is practically zero.

romen:
Merged to `master` as 00da0f69890874feaa555fafb99b967b861e9118 !

I will soon open a dedicate backport PR for 1.1.1

PR #12020, opened 2020-06-03
fips: add additional algorithms to the FIPS provider.

paulidale:
Discussions are ongoing but the OMC has approved the in-principle addition
of these algorithms to the upcoming FIPS validation.


paulidale:
Thanks for the review.

PR #12018, opened 2020-06-02
ossl_shim: include core_names.h to resolve undeclared symbols

paulidale:
Fixes #11987

- [x] tests are added or updated

PR #12022, opened 2020-06-03
ossl_shim: const cast the param arguments to avoid errors

paulidale:
Use const_cast<> to remove the _const_ nature of arguments to the OSSL_PARAM calls.

Refer #12018

### DIFF
--- a/test/ossl_shim/ossl_shim.cc
+++ b/test/ossl_shim/ossl_shim.cc
@@ -393,8 +393,10 @@ static int TicketKeyCallback(SSL *ssl, uint8_t *key_name, uint8_t *iv,
     return 0;
   }
 
-  *p++ = OSSL_PARAM_construct_utf8_string(OSSL_MAC_PARAM_DIGEST, "SHA256", 0);
-  *p++ = OSSL_PARAM_construct_octet_string(OSSL_MAC_PARAM_KEY, kZeros,
+  *p++ = OSSL_PARAM_construct_utf8_string(OSSL_MAC_PARAM_DIGEST,
+                                          const_cast<char *>("SHA256"), 0);
+  *p++ = OSSL_PARAM_construct_octet_string(OSSL_MAC_PARAM_KEY,
+                                           (void *)kZeros,
                                            sizeof(kZeros));
   *p = OSSL_PARAM_construct_end();
 


t8m:
Please use [extended tests]

paulidale:
I've added [extended tests].

test/ossl_shim/ossl_shim.cc
```diff
@@ -393,8 +393,10 @@ static int TicketKeyCallback(SSL *ssl, uint8_t *key_name, uint8_t *iv,
     return 0;
   }
 
-  *p++ = OSSL_PARAM_construct_utf8_string(OSSL_MAC_PARAM_DIGEST, "SHA256", 0);
-  *p++ = OSSL_PARAM_construct_octet_string(OSSL_MAC_PARAM_KEY, kZeros,
+  *p++ = OSSL_PARAM_construct_utf8_string(OSSL_MAC_PARAM_DIGEST,
+                                          const_cast<char *>("SHA256"), 0);
+  *p++ = OSSL_PARAM_construct_octet_string(OSSL_MAC_PARAM_KEY,
+                                           const_cast<void *>(kZeros),
```
bernd-edlinger, line 399: You need to change this API to use const char *, and const void*,
Otherwise this API is unusable from C++

paulidale: We can't change this API like that.  When constructing params, we've no idea if they will be read or written.  Thus the pointers passed in cannot be _const_.

t8m:
The void* cast does not work, probably should just be `const_cast<uint8_t *>`

test/ossl_shim/ossl_shim.cc
t8m: Hmm, could this be actually API deficiency? Should we have also read-only params?

paulidale:
Thanks @t8m.  I'd forgotten how much I disliked a few of C++'s nuances :)

test/ossl_shim/ossl_shim.cc
paulidale: We can do read only params, we can't use the construct calls to make them.  To use the helpers, casts need to be introduced.

Adding a duplicate set of `OSSL_PARAM_construct_read_only_XXX` calls would be a possibility.  It's a lot of calls and they'd need to cast internally.

t8m: Yes, I understand that we would have to have duplicate helper calls and I do not know whether it is worth it. I am just thinking as the read-only params would be quite common and the use of them with const char arrays would be also fairly common, whether this is something we should have or not. I suppose the construct calls could be added later, however implementing the read-only flag in the OSSL_PARAM would probably need to be done in 3.0 already. @levitte @mattcaswell what do you think? Is it worth the hassle?

paulidale:
Merged to master.  Hopefully this fixes the build.
Thanks for the feedback and help.

t8m:
Heh I wanted to provide approval ex-post but it was not possible. I suppose I'd have to reopen. Also @levitte should just re-approve if he agrees with the patch as it seems Paul merged it without proper approval.

paulidale:
I caught the window between approval and dismissal.
@levitte needs to approve this.

PR #12032, opened 2020-06-03
Fixes the bug "More bytes are generated for the key_block than needed for AEAD ciphers"

maximmasiutin:
Will make a new pull request to merge multiple commits.


ssl/t1_enc.c
```diff
@@ -175,6 +175,20 @@ int tls_provider_set_tls_params(SSL *s, EVP_CIPHER_CTX *ctx,
     return 1;
 }
 
+
+int EVP_CIPHER_iv_length_within_key_block(const EVP_CIPHER *c)
+{
+    int k;
```
richsalz, line 181: You can remove "k", and just replace the assignments with return statements, then remove the "else"

maximmasiutin:
This fixes the issue #12007 (More bytes are generated for Key Material than needed for AEAD ciphers). Reused existing code that was already calculating IV length within the key block. 

See the discussion at https://github.com/openssl/openssl/issues/12007 for more details.

It was the idea of @jwalch to use existing code.

@richsalz has proposed to optimize the code that I have taken using the "extract function" refactoring pattern. The proposal was to remove "k", and just replace the assignments with return statements, then remove the "else".

maximmasiutin:
@mattcaswell - this commit fixes the bug so no extra bytes are generated on AEAD.

maximmasiutin:
This pull request is basically about the two things:
1) Enables tracing of the key block length to show the bug (commits on Jun 3, 2020) by adding 'BIO_printf(trc_out, "key block length: %ld\n", num);'
2) Fixes the bug (commits on Sep 29, 2020) by creating a function 'EVP_CIPHER_iv_length_within_key_block' from existing code (extract function refactoring pattern) and calling it from that place and from a second place where IV length was taking regardless its context within the key_block.

maximmasiutin (commented):
I have changed the order of addendums to correspond to one given in the RFC. The old order was somewhat counter-intuitive.
Old:
    num = EVP_CIPHER_key_length(c) + mac_secret_size + EVP_CIPHER_iv_length(c);

New:
    num = mac_secret_size + EVP_CIPHER_key_length(c) + EVP_CIPHER_iv_length_within_key_block(c);


Here is the order in the RFC:

      client_write_MAC_key[SecurityParameters.mac_key_length]
      server_write_MAC_key[SecurityParameters.mac_key_length]
      client_write_key[SecurityParameters.enc_key_length]
      server_write_key[SecurityParameters.enc_key_length]
      client_write_IV[SecurityParameters.fixed_iv_length]
      server_write_IV[SecurityParameters.fixed_iv_length]


So, MAC_key comes first, then come cipher key and then the IV. That's why I put "mac_secret_size" first.

maximmasiutin:
Fixed the clang fatal waring: ssl/t1_enc.c:179:5: error: no previous prototype for function EVP_CIPHER_iv_length_within_key_block [-Werror,-Wmissing-prototypes]

ssl/t1_enc.c
maximmasiutin: Thank you, I have followed your advice.

maximmasiutin:
Fixes #12007

ssl/t1_enc.c
```diff
@@ -175,6 +175,18 @@ int tls_provider_set_tls_params(SSL *s, EVP_CIPHER_CTX *ctx,
     return 1;
 }
 
+
+static int EVP_CIPHER_iv_length_within_key_block(const EVP_CIPHER *c)
```
mattcaswell, line 179: Don't prefix this function with `EVP_CIPHER_`. We typically use this capitalisation to indicate that it is an external function. Since this is a tls specific calculation, I'd prefer something like `tls_iv_length_within_key_block`...or perhaps more simply `tls_fixed_iv_length`

ssl/t1_enc.c
```diff
@@ -583,6 +588,9 @@ int tls1_setup_key_block(SSL *s)
     s->s3.tmp.key_block = p;
 
     OSSL_TRACE_BEGIN(TLS) {
+
```
mattcaswell, line 591: Remove this blank line

mattcaswell (changes requested):
Not sure this is worth fixing in the 1.1.1 branch. Therefore I've just applied the "master" label.

maximmasiutin:
@mattcaswell, thank you for reviewing the code and proposing the changes. I have implemented and committed them.

mattcaswell (approved):
LGTM. Needs second review.

maximmasiutin:
Since I cannot edit the "Linked issues" field (maybe insufficient privileges), can you please manually link this pull request to the issue #12007 on the GitHub, so the linkage will be visible in the right top part of the web interface?

mattcaswell:
Just insert the text "Fixes #12007" as a line on its own in the commit description body (not the subject line).

On looking at the commits in this PR I notice that you have merge commits which are not allowed. Please can you squash all the commits into a single commit (removing any merge commits), and rebase this onto the current master? You will have to "force push" (i.e. "git push -f") to your github branch once your done.

maximmasiutin:
I will make a new clean pull request that has all the changes as one commit.

PR #12037, opened 2020-06-04
Remove disabled TLS 1.3 ciphers from the SSL(_CTX)

kaduk:
In ssl_create_cipher_list() we make a pass through the ciphers to
remove those which are disabled in the current libctx.  We are
careful to not include such disabled TLS 1.3 ciphers in the final
consolidated cipher list that we produce, but the disabled ciphers
are still kept in the separate stack of TLS 1.3 ciphers associated
with the SSL or SSL_CTX in question.  This leads to confusing
results where a cipher is present in the tls13_cipherlist but absent
from the actual cipher list in use.  Keep the books in order and
remove the disabled ciphers from the 1.3 cipherlist at the same time
we skip adding them to the active cipher list.

### DIFF
--- a/ssl/ssl_ciph.c
+++ b/ssl/ssl_ciph.c
@@ -1625,8 +1625,11 @@ STACK_OF(SSL_CIPHER) *ssl_create_cipher_list(SSL_CTX *ctx,
         if ((sslc->algorithm_enc & disabled_enc) != 0
                 || (ssl_cipher_table_mac[sslc->algorithm2
                                          & SSL_HANDSHAKE_MAC_MASK].mask
-                    & ctx->disabled_mac_mask) != 0)
+                    & ctx->disabled_mac_mask) != 0) {
+            sk_SSL_CIPHER_delete(tls13_ciphersuites, i);
+            i--;
             continue;
+        }
 
         if (!sk_SSL_CIPHER_push(cipherstack, sslc)) {
             sk_SSL_CIPHER_free(cipherstack);


kaduk:
I'm not sure whether or not this should go back to 1.1.1

mattcaswell:
> This leads to confusing
results where a cipher is present in the tls13_cipherlist but absent
from the actual cipher list in use. 

I'm wondering what confusing results you have experienced? AFAIK tls13_cipherlist is only ever used to construct the active cipher list and is never returned back to the application.

> I'm not sure whether or not this should go back to 1.1.1

I don't think there are any situations in 1.1.1 where discrepancies could occur anyway. The code to exclude TLSv1.3 disabled ciphers doesn't even exist there. We are interested here in ciphers that are enabled at compile time, but are disabled at runtime. If they are disabled at compile time then they don't even make it into the global `tls13_ciphers` list in the first place (and the only one this applies to anyway is TLS_CHACHA20_POLY1305_SHA256 - we don't support disabling AES at compile time). This is mostly about ciphers that may or may not be available via ENGINEs. I don't think you *can* disable AES or CHACHA at runtime in 1.1.1, but still have them enabled at compile time. At least I can't think how you could do that.

In 3.0 its a different story. Ciphers may or may not be available depending on what providers have been loaded.

With that in mind, while reviewing the code in order to respond to this PR, I just spotted a bug in ssl_load_ciphers in 3.0. It seems we update `disabled_enc_mask` et al in this function. Those seem to be *global* variables - but in 3.0 ssl_load_ciphers is called per SSL_CTX (in 1.1.1 its called once globally). The available ciphers may differ from one SSL_CTX to another, so what is available in one should not affect another SSL_CTX.

mattcaswell:
>  I just spotted a bug in ssl_load_ciphers in 3.0

I raised issue #12040 for that.

kaduk:
> > This leads to confusing
> > results where a cipher is present in the tls13_cipherlist but absent
> > from the actual cipher list in use.
> 
> I'm wondering what confusing results you have experienced? AFAIK tls13_cipherlist is only ever used to construct the active cipher list and is never returned back to the application.

Well, to be honest the confusion happened due to an interaction with some old code that we're maintaining in Akamai's OpenSSL derivative (but IIRC did not try to push upstream because it's been superseded by other advances in OpenSSL, e.g., server-preference).  In essence, we put some additional substructure within the (non-1.3) cipher list, and need to track that substructure across modifications to the overall list (including changes to both the 1.3 ciphersuites and the non-1.3 cipher list).  We've been using the cardinality of the `tls13_cipherlist` as an indication of how many 1.3 ciphersuites are at the start of the consolidated stack, and making corrections to our internal cipherlist structure on that assumption.  When there's three ciphers in the `tls13_cipherlist` but only two in the consolidated list, such logic is wrong, and I got test failures to point it out (in our internal tests for this particular feature, to be clear).

So while I'm not sure there's currently any user-visible consequence to this "mismatch" using just stock OpenSSL code without extra patches, it still feels like an internal inconsistency to me, and removing the inconsistency does not incur significant cost.  Thus, I am proposing this change for purposes of internal hygeine, and to be prepared if there is some future development that does start exposing this information.

> > I'm not sure whether or not this should go back to 1.1.1
> 
> I don't think there are any situations in 1.1.1 where discrepancies could occur anyway. The code to exclude TLSv1.3 disabled ciphers doesn't even exist there. We are interested here in ciphers that are enabled at compile time, but are disabled at runtime. If they are disabled at compile time then they don't even make it into the global `tls13_ciphers` list in the first place (and the only one this applies to anyway is TLS_CHACHA20_POLY1305_SHA256 - we don't support disabling AES at compile time). This is mostly about ciphers that may or may not be available via ENGINEs. I don't think you _can_ disable AES or CHACHA at runtime in 1.1.1, but still have them enabled at compile time. At least I can't think how you could do that.
> 

I agree that I don't see a path for such discrepancies in 1.1.1 to begin with.  I only encountered issues with chacha/poly being disabled in the fips provider.

> In 3.0 its a different story. Ciphers may or may not be available depending on what providers have been loaded.
> 
> With that in mind, while reviewing the code in order to respond to this PR, I just spotted a bug in ssl_load_ciphers in 3.0. It seems we update `disabled_enc_mask` et al in this function. Those seem to be _global_ variables - but in 3.0 ssl_load_ciphers is called per SSL_CTX (in 1.1.1 its called once globally). The available ciphers may differ from one SSL_CTX to another, so what is available in one should not affect another SSL_CTX.

Good point.  I had noticed it was a little weird that we also have `disabled_enc` et all, that are initialized from `disabled_enc_mask` in `ssl_create_cipher_list()` but had missed the glaring race condition there.

kaduk:
@mattcaswell I rebased past the per-ctx disabled mask changes that you made as a result of starting to review this change.
It would be good to get a sense for whether this should just be rejected as not user-visible or not so we can close it out one way or the other.

mattcaswell (approved):
I don't strongly feel the need for this PR - but I also don't object so I'm fine with you merging this if you want.

kaduk:
Pushed to master; thanks for the review.

PR #12039, opened 2020-06-04
some tweaks for the provider cipher implementation

kaduk:
Don't error out if asking for an IV using a buffer that's larger than the IV in question (we do this from libcrypto!).
Also, be more consistent about IV handling, since they can be requested both as an octet string and an octet pointer.

### DIFF
--- a/providers/implementations/ciphers/cipher_aes_cbc_hmac_sha.c
+++ b/providers/implementations/ciphers/cipher_aes_cbc_hmac_sha.c
@@ -229,7 +229,8 @@ static int aes_get_ctx_params(void *vctx, OSSL_PARAM params[])
     }
     p = OSSL_PARAM_locate(params, OSSL_CIPHER_PARAM_IV);
     if (p != NULL
-        && !OSSL_PARAM_set_octet_string(p, ctx->base.oiv, ctx->base.ivlen)) {
+        && !OSSL_PARAM_set_octet_string(p, ctx->base.oiv, ctx->base.ivlen)
+        && !OSSL_PARAM_set_octet_ptr(p, &ctx->base.oiv, ctx->base.ivlen)) {
         ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_SET_PARAMETER);
         return 0;
     }
--- a/providers/implementations/ciphers/cipher_aes_ocb.c
+++ b/providers/implementations/ciphers/cipher_aes_ocb.c
@@ -401,7 +401,7 @@ static int aes_ocb_get_ctx_params(void *vctx, OSSL_PARAM params[])
 
     p = OSSL_PARAM_locate(params, OSSL_CIPHER_PARAM_IV);
     if (p != NULL) {
-        if (ctx->base.ivlen != p->data_size) {
+        if (ctx->base.ivlen > p->data_size) {
             ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_IV_LENGTH);
             return 0;
         }
--- a/providers/implementations/ciphers/cipher_aes_ocb.c
+++ b/providers/implementations/ciphers/cipher_aes_ocb.c
@@ -405,7 +405,8 @@ static int aes_ocb_get_ctx_params(void *vctx, OSSL_PARAM params[])
             ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_IV_LENGTH);
             return 0;
         }
-        if (!OSSL_PARAM_set_octet_string(p, ctx->base.oiv, ctx->base.ivlen)) {
+        if (!OSSL_PARAM_set_octet_string(p, ctx->base.oiv, ctx->base.ivlen)
+            && !OSSL_PARAM_set_octet_ptr(p, &ctx->base.oiv, ctx->base.ivlen)) {
             ERR_raise(ERR_LIB_PROV, PROV_R_FAILED_TO_SET_PARAMETER);
             return 0;
         }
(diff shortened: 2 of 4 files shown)


providers/implementations/ciphers/cipher_aes_ocb.c
```diff
@@ -401,11 +401,12 @@ static int aes_ocb_get_ctx_params(void *vctx, OSSL_PARAM params[])
 
     p = OSSL_PARAM_locate(params, OSSL_CIPHER_PARAM_IV);
     if (p != NULL) {
-        if (ctx->base.ivlen != p->data_size) {
+        if (ctx->base.ivlen > p->data_size) {
```
mattcaswell, line 404: Isn't the data_size the size of the pointer in the case of octet_ptr?

kaduk:
> Where do we request them as an octet pointer? I recall there were concerns with over using octet ptr...but I don't immediately recall what the issue was.

I also remember some discussion, but haven't found exactly where yet (that's a TODO for another change I've got pending, the shape of which will depend on those previous discussions).

As far as requesting an octet pointer, see https://github.com/openssl/openssl/blob/19d15f97b3932d14e17d0b61115b664169ca6159/crypto/evp/evp_lib.c#L463-L475

providers/implementations/ciphers/cipher_aes_ocb.c
kaduk: I confess I'm not 100% clear on that; in the case I link to (`EVP_CIPHER_CTX_iv_noconst`) we pass the size of the `iv[]` array, at least.

kaduk: I don't see anything in either OSSL_PARAM_BLD.pod or OSSL_PARAM_int.pod that suggests that the size should be the size of the pointer.
The implementation of `OSSL_PARAM_BLD_push_octet_ptr()`, however, does:
https://github.com/openssl/openssl/blob/00da0f69890874feaa555fafb99b967b861e9118/crypto/param_build.c#L287-L302
Could you say a bit more about why you believe the pointer size is correct? In some sense it seems like the value is unused, since we're not providing a buffer, and are instead getting a live pointer to someone else's allocated storage of the proper length.

providers/implementations/ciphers/cipher_aes_cbc_hmac_sha.c
```diff
@@ -229,7 +229,8 @@ static int aes_get_ctx_params(void *vctx, OSSL_PARAM params[])
     }
     p = OSSL_PARAM_locate(params, OSSL_CIPHER_PARAM_IV);
     if (p != NULL
-        && !OSSL_PARAM_set_octet_string(p, ctx->base.oiv, ctx->base.ivlen)) {
+        && !OSSL_PARAM_set_octet_string(p, ctx->base.oiv, ctx->base.ivlen)
+        && !OSSL_PARAM_set_octet_ptr(p, &ctx->base.oiv, ctx->base.ivlen)) {
```
levitte, line 233: Oooooh, I never thought of doing it this way!  Nice!

providers/implementations/ciphers/cipher_aes_ocb.c
levitte: In OSSL_PARAM.doc, there's this section (rendered here):

>   _data_size_ must be set to the size of the data, not the size of the
>   pointer to the data.  If this is used in a parameter request,
>   _data_size_ is not relevant.  However, the _responder_ will set
>   _return_size_ to the size of the data.

levitte: I recall that we discussed what the data size for a indirect data pointer (which is what a `OSSL_PARAM_OCTET_PTR` _data_ really is) should be the pointer size or the size of the value being (indirectly) pointed at.  @mattcaswell may simply misremember how we concluded that discussion.

providers/implementations/ciphers/cipher_aes_cbc_hmac_sha.c
kaduk: I was just cribbing from https://github.com/openssl/openssl/blob/6985b0e3deaee2f6e83a670ce7b33cf9ee47933a/providers/implementations/ciphers/ciphercommon.c#L361-L367 , for which it seems we can blame ... you (in 06c8331c51797f43881e5262b4c114aacc6d0ca0) :)
Shane moved it around a bunch in follow-up commits, though.  (Also, passing `&ctx->iv` for both octet-string and octet-ptr feels weird to me.)

providers/implementations/ciphers/cipher_aes_ocb.c
kaduk: So, Richard, it seems that you are saying that Matt was incorrect to request changes here?
(Does anyone care to push the 'approve' button? ;)

kaduk: Or is the claim that I should change it to only check the length in the octet-string case?

kaduk: @levitte some clarification would be appreciated

kaduk:
@levitte I'm waiting for further input on this one, if you get a chance.

providers/implementations/ciphers/cipher_aes_ocb.c
levitte: I can only repeat what I just pointed out, that `data_size` should be set to the size of the data, i.e. the size of the storage that the pointer that `data` points at is pointing at.

(verbal indirect reference... it just had to be done :wink:)

providers/implementations/ciphers/cipher_aes_cbc_hmac_sha.c
levitte: Oh, uhm, I guess I've forgotten stuff I've done.

kaduk:
Thanks for the review!
Pushed to master; closing.

PR #12041, opened 2020-06-04
Silence gcc false positive warning on alpn_protos_len in test/handshake_helper.c in 1.1.1

DDvO:
As agreed with @bernd-edlinger in https://github.com/openssl/openssl/pull/12036#discussion_r435153089

PR #12042, opened 2020-06-04
Silence gcc false positive warning on alpn_protos_len in test/handshake_helper.c

DDvO:
As suggested by @bernd-edlinger, same as #12041 but for current master.


DDvO:
@mattcaswell, wanna approve this variant for the current master as well?

DDvO:
Consequently also pushed this variant of #12041 for the 3.0 master -
Thanks @bernd-edlinger, @mattcaswell and @t8m.

PR #12046, opened 2020-06-04
[bp-1.1.1][crypto/ec] Remove unreachable AVX2 code in NISTZ256 implementation

romen:
`crypto/ec/ecp_nistz256.c` contains code sections guarded by a `ECP_NISTZ256_AVX2` define.

The relevant comment reads:

>     /*
>      * Note that by default ECP_NISTZ256_AVX2 is undefined. While it's great
>      * code processing 4 points in parallel, corresponding serial operation
>      * is several times slower, because it uses 29x29=58-bit multiplication
>      * as opposite to 64x64=128-bit in integer-only scalar case. As result
>      * it doesn't provide *significant* performance improvement. Note that
>      * just defining ECP_NISTZ256_AVX2 is not sufficient to make it work,
>      * you'd need to compile even asm/ecp_nistz256-avx.pl module.
>      */

Without diminishing the quality of the original submission, it's evident that this code has been basically unreachable without modifications to the library source code and is under-tested.

This PR removes these sections from the codebase.

### Branches

This PR is the 1.1.1 backport of  #12019 .

00da0f69890874feaa555fafb99b967b861e9118 cherry-picked cleanly (excluding a warning on file removal), but I prefer to open it as a separate dedicated PR so that we can gather an OMC explicit approval about removing code from the stable LTS branch (probably unnecessary, as this code is anyway unreachable normally, but still I prefer to err on the side of caution).

Also, this gives the opportunity to run the extended tests on 1.1.1 before merging.


romen:
Merged as 49eebbc408535e8670cb62ea661b696ec5819f15

Thanks!

PR #12052, opened 2020-06-04
Fix NULL pointer access caused by X509_ATTRIBUTE_create()

ischwarze:
When X509_ATTRIBUTE_create() receives an invalid NID (e.g., -1), return
failure rather than silently constructing a broken X509_ATTRIBUTE object
that might cause NULL pointer accesses later on.  This matters because
X509_ATTRIBUTE_create() is used by API functions like PKCS7_add_attribute(3)
and the NID comes straight from the user.

This bug was found while working on LibreSSL documentation.

Reviewed-by: Theo Buehler <tb@openbsd.org>

CLA: trivial

Original commit: https://cvsweb.openbsd.org/src/lib/libcrypto/asn1/x_attrib.c#rev1.14
Regression test available: https://cvsweb.openbsd.org/src/regress/lib/libcrypto/x509/x509attribute.c

I expect you also need to merge this to 1.1.1-stable.

### DIFF
--- a/crypto/x509/x_attrib.c
+++ b/crypto/x509/x_attrib.c
@@ -39,10 +39,13 @@ X509_ATTRIBUTE *X509_ATTRIBUTE_create(int nid, int atrtype, void *value)
 {
     X509_ATTRIBUTE *ret = NULL;
     ASN1_TYPE *val = NULL;
+    ASN1_OBJECT *oid;
 
+    if ((oid = OBJ_nid2obj(nid)) == NULL)
+        return NULL;
     if ((ret = X509_ATTRIBUTE_new()) == NULL)
         return NULL;
-    ret->object = OBJ_nid2obj(nid);
+    ret->object = oid;
     if ((val = ASN1_TYPE_new()) == NULL)
         goto err;
     if (!sk_ASN1_TYPE_push(ret->set, val))


ischwarze:
mattcaswell asked:

> Were you the original author of the commit in Libressl?

Yes.  I found the bug while auditing code in order to write documentation, and i also developed, tested, and committed the patch for LibreSSL.

Thanks to both Matt and Tomáš for checking the patch!

richsalz:
If there's only one way to do things, then it's generally not copyrightable, @t8m.  And hey, @ischwarze nice to see you!

ischwarze:
Hi @richsalz,

> If there's only one way to do things, then it's generally not copyrightable, @t8m. And hey, @ischwarze nice to see you!

Same here!  :-)  You still didn't come round to systematically merge any substantial part of all the documentation work i have been doing, right?  On the other hand, given that you still have a viable license on 1.1.1-stable, i could still merge various of the fixes from your tree.  At some point, this will inevitably grow into a problem (so far, we have been lucky that it didn't already), in both directions: i can no longer take documentation improvements from your master branch because the new license is tainted with contract and patent law and no longer free enough, and once i find any non-trivial bug, you will have a problem because you can no longer take non-trivial patches from outsiders, by your own choice of requiring legalese paperwork (from a brief glance, your legalese is not quite as bad as what the FSF concocted and which i rejected outright even though they invited me as a developer to one of their projects - but i'm still reluctant to waste my time on such paperwork, apart from the additional legal risk that i perceive in signing any such fine print...).

t-j-h:
There is more than one way to implement this check so don't see "only one way" as meaningfully applying to this situation.

@ischwarze  I appreciate your public comments on the CLA paperwork (always good to get feedback) and one thing we did do was basically re-use the ASF approach which is also used on a very large number of open source projects. It makes any evaluation of your willingness to accept such an approach broadly applicable rather than having to see OpenSSL as its own unique complicated context (which is what it was prior to this change).

richsalz:
@ischwarze I owe you an apology.  I'm sorry that I never followed-up to say that the mdoc/perldoc comparison become too hard for me to reconcile, and I let your changes fall through the cracks.  Perhaps now I could try again, if you're willing to resend me the changes.

ischwarze:
@richsalz wrote:

> @ischwarze I owe you an apology.

No problem at all, everybody is busy!

> I'm sorry that I never followed-up to say that the mdoc/perldoc
> comparison become too hard for me to reconcile, and I let your
> changes fall through the cracks.

IIRC, you actually did mention that, which is why i ventured to send another ping, suspecting that the situation might eventually improve.  Of course, you might also just switch to mdoc(7) which would improve the quality of your markup and make merging easier going forward...

> Perhaps now I could try again, if you're willing to resend me the changes.

Sure.  After some years of work, it would of course be hard to make sure that all patches i'm including are indeed relevant for you, so i preferred to err to the side of including a bit too many patches rather than including too few.  I guess you can just skip those that don't matter for you.  But i did exclude the original POD->MDOC conversion even though various improvements were already sneaked into that step, but they would be very hard to extract, and including the raw conversion would make this patchset even bulkier.

So, here are the patches (LibreSSL manual page changes only):
  http://www.studis.de/privat/schwarze/Tmp/richsalz.patches

   $ wc richsalz.patches
  182031  845141 6020508 richsalz.patches
   $ grep -c ^commit richsalz.patches
  945
   $ grep -c ^diff richsalz.patches
  3343
   $ grep -c '^+[^+]' richsalz.patches
  76322
   $ grep -c '^-[^-]' richsalz.patches
  21532
   $ grep ^Author richsalz.patches  | sort | uniq -c | sort -nr
 803 Author: schwarze 
  81 Author: jmc 
  27 Author: tb 
   8 Author: jsing 
   8 Author: beck 
   2 Author: sobrado 
   2 Author: reyk 
   2 Author: bcook 
   1 Author: mmcc 
   1 Author: millert 
   1 Author: jsg 
   1 Author: jca 
   1 Author: inoguchi 
   1 Author: guenther 
   1 Author: dtucker 
   1 Author: dlg 
   1 Author: djm 
   1 Author: deraadt 
   1 Author: claudio 
   1 Author: benno 

All our changes are under the ISC license, so including them should be no problem for you.

Do you think you can handle that?

Yours,
  Ingo

richsalz:
I have the file, thanks.  Love all the grep/wc metrics!
Can you ping me off-line?  rsalz at akamai.com

t8m:
Merged to master and 1.1.1 branch. Thank you for the contribution.

PR #12055, opened 2020-06-05
Fix a CMS_get0_signers() description in CMS_verify.pod

olszomal:
The original wording suggests that it is required to execute CMS_get0_signers() after CMS_verify(), while it is CMS_get0_signers() that requires prior successful invocation of CMS_verify().

[skip ci]
CLA: trivial

### DIFF
--- a/doc/man3/CMS_verify.pod
+++ b/doc/man3/CMS_verify.pod
@@ -24,7 +24,7 @@ present in B<cms>. The content is written to B<out> if it is not NULL.
 B<flags> is an optional set of flags, which can be used to modify the verify
 operation.
 
-CMS_get0_signers() retrieves the signing certificate(s) from B<cms>, it must
+CMS_get0_signers() retrieves the signing certificate(s) from B<cms>, it may only
 be called after a successful CMS_verify() operation.
 
 =head1 VERIFY PROCESS


olszomal:
CLA: trivial

slontis:
Both wordings seems to indicate that CMS_get0_signers() can only be called after CMS_verify().
So I am not sure the change is necessary?

mtrojnar:
> Both wordings seems to indicate that CMS_get0_signers() can only be called after CMS_verify().

I respectfully disagree.  "B must be called after A" means much more than "B requires A".  It also literally means that every call of A must be followed by a call of B.

slontis:
So your objection is really to the word 'must'? (Which really depends how you read it).
So replacing must with 'may only' would work too right?

mtrojnar:
@slontis IMHO 'may only' is sufficiently clear.  @olszomal?

slontis:
I agree that this is trivial and does not require a CLA.

t8m:
This should be cherry-pickable to 1.1.1. @slontis Do you agree with merging to 1.1.1 as well?

t8m:
I take the thumbs-up as yes.

t8m:
Merged to master and 1.1.1. Thank you for the PR!

PR #12057, opened 2020-06-05
TESTUTIL: Separate TAP output and other output by BIO filter

### LINKED ISSUE #12054: evp_pkey_provided_test fails randomly sometimes
The evp_pkey_provided_test randomly fails sometimes. It occurs because it dumps PEM files to stdout. If those PEM files happen to contain a line that starts with the text "ok" then it confuses the test harness and the test fails.

Example failing build:

https://travis-ci.org/github/openssl/openssl/jobs/694280520

levitte:
Output that's supposed to be understood by a TAP parser gets its own
BIOs (|tap_out| and |tap_err|), and is only used internally within
testutils.  |bio_out| and |bio_err| is now only used for output that
shouldn't be parsed by the TAP parser, and all output written to those
BIOs are therefore always made to look like comments (it gets prefixed
with "# ").

Indentation and prefixing with "# " is reworked to use BIO_f_prefix(),
which allows us to throw away the internal BIO_f_tap().

The indentation level is now adjusted via a special function.

Fixes #12054


test/testutil/basic_output.c
```diff
@@ -14,27 +14,51 @@
 #include <openssl/crypto.h>
 #include <openssl/bio.h>
 
+/* The are available for any test program */
```
mattcaswell, line 17: s/The/These/

test/testutil/basic_output.c
```diff
@@ -14,27 +14,51 @@
...
 BIO *bio_out = NULL;
 BIO *bio_err = NULL;
 
+/* The are available for TAP output only (internally) */
```
mattcaswell, line 21: and here

test/testutil/basic_output.c
levitte: Fixed

test/testutil/basic_output.c
levitte: Fixed

levitte:
Merged

8dce4aa2d974fa357bb3f7d8a77e581d487de732 TESTUTIL: Separate TAP output and other output by BIO filter
66d8bd4b37c759e2db51324f5614438bf3db1dbd TESTUTIL: Adjust the rest of testutil
c4683009ad6e8d64e80112fd689921f6c169bd20 TEST: Adjust test/bioprinttest.c to behave like the testutil routines

PR #12061, opened 2020-06-05
use safe primes in ssl_get_auto_dh()

### LINKED ISSUE #12060: Insecure DH parameters selected by default
When using openssl s_server with RSA keys that are 1024 bit or 2048 bit long OpenSSL will select DH parameters from RFC 5114, a.k.a group 22 and group 23 for IKE.

Those parameters do not use a safe primes, they do have small subgroups, see also https://jhalderm.com/pub/papers/subgroup-ndss16.pdf

This affects both current 1.1.1 (8354f53d407) and master (c7f837cfcc5)

   <details><summary>Reproducer:</summary><p>

```
openssl req -x509 -newkey rsa:2048 -keyout key.pem -out cert.pem -days 365 -nodes -subj "/CN=localhost"
openssl s_server -key key.pem -cert cert.pem -www &> openssl.err &
openssl_pid=$!
git clone https://github.com/tomato42/tlsfuzzer
pushd tlsfuzzer
git checkout ffdh-param-matching  # won't be needed after https://github.com/tomato42/tlsfuzzer/pull/668 is merged
git clone https://github.com/tomato42/tlslite-ng .tlslite-ng
ln -s .tlslite-ng/tlslite tlslite
git clone https://github.com/warner/python-ecdsa .python-ecdsa
ln -s .python-ecdsa/ecdsa ecdsa
PYTHONPATH=. python scripts/test-ffdhe-expected-params.py
popd
kill $openssl_pid
```

   </p></details>

   <details><summary>tlsfuzzer output:</summary><p>

```
sanity ...
OK

FFDH parameters check ...
Error encountered while processing node <tlsfuzzer.expect.ExpectServerKeyExchange object at 0x7f68949a6e10> (child: <tlsfuzzer.expect.ExpectServerHelloDone object at 0x7f68949a6e50>) with last message being: <tlslite.messages.Message object at 0x7f68949acc90>
Error while processing
Traceback (most recent call last):
  File "scripts/test-ffdhe-expected-params.py", line 217, in main
    runner.run()
  File "/home/hkario/dev/tlsfuzzer/tlsfuzzer/runner.py", line 239, in run
    node.process(self.state, msg)
  File "/home/hkario/dev/tlsfuzzer/tlsfuzzer/expect.py", line 1147, in process
    self._checkParams(server_key_exchange)
  File "/home/hkario/dev/tlsfuzzer/tlsfuzzer/expect.py", line 1084, in _checkParams
    "received: {0}".format(name))
AssertionError: DH parameters not from valid set, received: RFC5114 group 23

sanity ...
OK

Test to check if server selects the expected DH parameters
See tlslite.mathtls.FFDHE_PARAMETERS for supported names

Test end
====================
version: 1
====================
TOTAL: 3
SKIP: 0
PASS: 2
XFAIL: 0
FAIL: 1
XPASS: 0
====================
FAILED:
        'FFDH parameters check'
```

   </p></details>

Using a larger RSA key, like 3072, will cause OpenSSL to select the secure RFC 3526 group 15 parameters.

(side note: it's possible to specify expected parameters using the `--named-ffdh`, e.g.: `PYTHONPATH=. python scripts/test-ffdhe-expected-params.py --named-ffdh "RFC3526 group 15"`)

tomato42:
DH_get_1024_160() and DH_get_2048_224() return parameters from
RFC5114. Those parameters include primes with known small subgroups,
making them unsafe. Change the code to use parameters from
RFC 2409 and RFC 3526 instead (group 2 and 14 respectively).

This patch also adds automatic selection of 4096 bit params for 4096 bit
RSA keys

fixes #12060 
Should be applied both to master and cherry-picked to 1.1.1

##### Checklist

### DIFF
--- a/ssl/t1_lib.c
+++ b/ssl/t1_lib.c
@@ -2646,46 +2646,48 @@ int SSL_check_chain(SSL *s, X509 *x, EVP_PKEY *pk, STACK_OF(X509) *chain)
 #ifndef OPENSSL_NO_DH
 DH *ssl_get_auto_dh(SSL *s)
 {
+    DH *dhp;
+    BIGNUM *p, *g;
     int dh_secbits = 80;
-    if (s->cert->dh_tmp_auto == 2)
-        return DH_get_1024_160();
-    if (s->s3.tmp.new_cipher->algorithm_auth & (SSL_aNULL | SSL_aPSK)) {
-        if (s->s3.tmp.new_cipher->strength_bits == 256)
-            dh_secbits = 128;
-        else
-            dh_secbits = 80;
-    } else {
-        if (s->s3.tmp.cert == NULL)
-            return NULL;
-        dh_secbits = EVP_PKEY_security_bits(s->s3.tmp.cert->privatekey);
+    if (s->cert->dh_tmp_auto != 2) {
+        if (s->s3.tmp.new_cipher->algorithm_auth & (SSL_aNULL | SSL_aPSK)) {
+            if (s->s3.tmp.new_cipher->strength_bits == 256)
+                dh_secbits = 128;
+            else
+                dh_secbits = 80;
+        } else {
+            if (s->s3.tmp.cert == NULL)
+                return NULL;
+            dh_secbits = EVP_PKEY_security_bits(s->s3.tmp.cert->privatekey);
+        }
     }
 
-    if (dh_secbits >= 128) {
-        DH *dhp = DH_new();
-        BIGNUM *p, *g;
-        if (dhp == NULL)
-            return NULL;
-        g = BN_new();
-        if (g == NULL || !BN_set_word(g, 2)) {
-            DH_free(dhp);
-            BN_free(g);
-            return NULL;
-        }
-        if (dh_secbits >= 192)
-            p = BN_get_rfc3526_prime_8192(NULL);
-        else
-            p = BN_get_rfc3526_prime_3072(NULL);
-        if (p == NULL || !DH_set0_pqg(dhp, p, NULL, g)) {
-            DH_free(dhp);
-            BN_free(p);
-            BN_free(g);
-            return NULL;
-        }
-        return dhp;
+    dhp = DH_new();
+    if (dhp == NULL)
+        return NULL;
+    g = BN_new();
+    if (g == NULL || !BN_set_word(g, 2)) {
+        DH_free(dhp);
+        BN_free(g);
+        return NULL;
     }
-    if (dh_secbits >= 112)
-        return DH_get_2048_224();
-    return DH_get_1024_160();
+    if (dh_secbits >= 192)
+        p = BN_get_rfc3526_prime_8192(NULL);
+    else if (dh_secbits >= 152)
+        p = BN_get_rfc3526_prime_4096(NULL);
+    else if (dh_secbits >= 128)
+        p = BN_get_rfc3526_prime_3072(NULL);
+    else if (dh_secbits >= 112)
+        p = BN_get_rfc3526_prime_2048(NULL);
+    else
+        p = BN_get_rfc2409_prime_1024(NULL);
+    if (p == NULL || !DH_set0_pqg(dhp, p, NULL, g)) {
+        DH_free(dhp);
+        BN_free(p);
+        BN_free(g);
+        return NULL;
+    }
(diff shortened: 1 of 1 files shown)


kroeckx:
I'm trying to understand what dh_tmp_auto means. I think there was some case we set that internally, but I can't seem to find it.

kroeckx:
As "documented" in commit 09599b52d4e295c380512ba39958a11994d63401, == 2 means only 1024 bit, because some peers can't handle such big keys.

kroeckx:
Is there a reason not to set q? If q is set, we generate a key in the range of 2 - q-1.

kroeckx:
I think we should just drop the case where dh_tmp_auto == 2

kroeckx:
Looking closer at what happens, the code is at least confusing. In DH_set0_pqg(), if q != NULL, we set length = BN_num_bits(q). But in DH_compute_key(), if q is set, we generate in the range 2 - q-1, and otherwise look if length is set and use that.

tomato42:
> Is there a reason not to set q? If q is set, we generate a key in the range of 2 - q-1.

setting q is only relevant when you're not dealing with safe primes, to check if the peer shared key matches the generator. For safe primes just checking for 0, 1 and -1 is enough to exclude all small subgroups

> I think we should just drop the case where dh_tmp_auto == 2

there are at least two implementations that can't handle primes larger than 1024 bit, and while old, people may care about being able to communicate with them at all, even if it's insecure

> Looking closer at what happens, the code is at least confusing. In DH_set0_pqg(), if q != NULL, we set length = BN_num_bits(q). But in DH_compute_key(), if q is set, we generate in the range 2 - q-1, and otherwise look if length is set and use that.

I don't see how this is relevant. Setting just the g and p is exactly what a TLS client has to do when calculating its key share and then shared key. Writing down just the g and p is exactly what `openssl dhparam` will do when asked to generate parameters. If this is wrong, we have much bigger problems than use of potentially back-doored primes.

kroeckx:
It's my understanding that you don't want to generate the public
key of 1, so avoid the private key of 0 and q. Setting
q will prevent it from generating q.

I guess I was looking at the 1.1.1 branch before. In master
we now actually have code to see if it's one of the known groups,
where p and g are checked against a list of known groups, and sets
q if it's not set. And the rfc3526 groups you use are actually
in the list, so it doesn't matter that you set q or not in master,
you'll end up generating a value in the range 1 - q-1.

But I guess you can also argue that generating q is very
unlikely and we shouldn't bother checking it.

ssl/t1_lib.c
```diff
@@ -2647,45 +2647,46 @@ int SSL_check_chain(SSL *s, X509 *x, EVP_PKEY *pk, STACK_OF(X509) *chain)
...
-            p = BN_get_rfc3526_prime_3072(NULL);
-        if (p == NULL || !DH_set0_pqg(dhp, p, NULL, g)) {
-            DH_free(dhp);
-            BN_free(p);
-            BN_free(g);
-            return NULL;
-        }
-        return dhp;
+    DH *dhp = DH_new();
+    BIGNUM *p, *g;
```
kroeckx, line 2664: Variable declarations need to be at the start of a block

kroeckx:
Other than the place of the variable declaration, it look good

tomato42:
> It's my understanding that you don't want to generate the public key of 1, so avoid the private key of 0 and q. Setting q will prevent it from generating q. I guess I was looking at the 1.1.1 branch before. 
> But I guess you can also argue that generating q is very unlikely and we shouldn't bother checking it.

actually it's more that that, it's not worth because it's literally selecting a 1 in 2^1024 number (something so unlikely that the security of RSA depends on no two people selecting the same /prime/ from a smaller range),
and secondly, it will cause generation of malformed key shares, as g^q == -1 mod p, for g that is a primitive root or g^q == 1 mod p if it's not, both of which lead to invalid key share and invalid shared secret (in case it would be accepted by peer)

t8m:
Merged to master as 7646610. Thank you for the PR.

Unfortunately this does not cherry-pick cleanly to 1.1.1 branch. @tomato42 Would you please prepare a PR for 1.1.1? Or I can do that if you want.

tomato42:
@t8m I may do it later this week, I won't mind if you do it yourself

PR #12048, opened 2020-06-04
EVP: Let EVP_PKEY_gen() initialize ctx->keygen_info

levitte:
In EVP_PKEY_METHOD code, the backend initializes ctx->keygen_info.
With provider side code, it's not possible to reach back into the
EVP_PKEY_CTX in the same manner, so we need to make that
initialization in the central generation function, EVP_PKEY_gen().

This isn't quite compatible with the idea that keygen_info could have
an arbitrary amount of elements, but since all our legacy backends use
exactly two elements, that's what we go for.

Fixes #12047

## APPS: Fix 'openssl dsaparam -genkey'

Using a parameter EVP_PKEY for key generation with EVP_PKEY routines
works a little differently than the raw DSA routines that were used
before.

While fixing that, clean away all remaining use of the DSA type, which
simplifies the code a bit more.

## APPS: Fix 'openssl dhparam'

'dhparam' can't be completely rewritten in terms of EVP_PKEY functions
yet, because we lack X9.42 support.  However, we do when generating,
but forgot to extract a DH pointer with EVP_PKEY_get0_DH().

Issue #12047, opened 2020-06-04
dsaparam and dhparm are broken


DDvO:
Trying to repair the tests in `test/recipes/90-test_store.t` for #11733 
I've come across nice crashes of `dsaparam`.

Calling it with a key size acceptable these days, e.g., 2048:
```
apps/openssl dsaparam  -out dsaparam.pem 2048
```
or
```
apps/openssl dsaparam -genkey -out dsakey.pem 2048
```
one gets for instance
```
==11521==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7f6239ee36d8 bp 0x7fff02e120f0 sp 0x7fff02e12050 T0)
==11521==The signal is caused by a WRITE memory access.
==11521==Hint: address points to the zero page.
    #0 0x7f6239ee36d8 in ossl_callback_to_pkey_gencb crypto/evp/pmeth_gn.c:136
    #1 0x7f623a127516 in dsa_gencb providers/implementations/keymgmt/dsa_kmgmt.c:487
    #2 0x7f6239d002f5 in BN_GENCB_call crypto/bn/bn_prime.c:115
    #3 0x7f6239ef4091 in generate_q_fips186_4 crypto/ffc/ffc_params_generate.c:290
    #4 0x7f6239ef54e5 in ffc_params_FIPS186_4_gen_verify crypto/ffc/ffc_params_generate.c:634
    #5 0x7f6239ef71b3 in ffc_params_FIPS186_4_generate crypto/ffc/ffc_params_generate.c:982
    #6 0x7f6239db6b1f in dsa_generate_ffc_parameters crypto/dsa/dsa_gen.c:46
    #7 0x7f623a1279b5 in dsa_gen providers/implementations/keymgmt/dsa_kmgmt.c:528
    #8 0x7f6239ec5b58 in evp_keymgmt_gen crypto/evp/keymgmt_meth.c:335
    #9 0x7f6239ec3587 in evp_keymgmt_util_gen crypto/evp/keymgmt_lib.c:413
    #10 0x7f6239ee3cf8 in EVP_PKEY_gen crypto/evp/pmeth_gn.c:190
    #11 0x7f6239ee4358 in EVP_PKEY_paramgen crypto/evp/pmeth_gn.c:276
    #12 0x44daa7 in dsaparam_main apps/dsaparam.c:184
    #13 0x465fc1 in do_cmd apps/openssl.c:467
    #14 0x46528c in main apps/openssl.c:279
```
[...]

PR #12062, opened 2020-06-05
Improve DH and DSA error output

DDvO:
As discussed in https://github.com/openssl/openssl/issues/12047#issuecomment-638967154 and https://github.com/openssl/openssl/pull/12048#issuecomment-639472423, there is no to-the-point error output in case DH or DSA key sizes are considered unsuitable according to FIPS 186 FFC.

Maybe similar related issues should be handled as well, 
maybe the placement of the `DHerr` and `DSAerr` calls is not yet optimal,
and maybe the wording of the reason codes needs improvement,
but this PR is at least a starting point.


apps/dhparam.c
```diff
@@ -240,20 +240,20 @@ int dhparam_main(int argc, char **argv)
                        num, g);
             BIO_printf(bio_err, "This is going to take a long time\n");
             if (!EVP_PKEY_paramgen_init(ctx)) {
+                ERR_print_errors(bio_err);
                 BIO_printf(bio_err,
                            "Error, unable to initialise DH param generation\n");
-                ERR_print_errors(bio_err);
```
levitte, line 245: Why the order change?

crypto/dh/dh_gen.c
```diff
@@ -57,8 +57,12 @@ int dh_generate_ffc_parameters(DH *dh, int type, int pbits,
         ret = ffc_params_FIPS186_4_generate(dh->libctx, &dh->params,
                                             FFC_PARAM_TYPE_DH,
                                             pbits, qbits, md, &res, cb);
-    if (ret > 0)
+    if (ret > 0) {
         dh->dirty_cnt++;
+    } else {
+        if ((res & FFC_CHECK_BAD_LN_PAIR) != 0)
+            DHerr(0, DH_R_BAD_FFC_PARAMETERS);
+    }
```
levitte, line 65: Why not add errors in `ffc_params_FIPS186_2_generate` and `ffc_params_FIPS186_4_generate`?

apps/dhparam.c
DDvO: > Why the order change?

Because usually the low-level errors are printed before/above the high-level ones.
Not following this pattern (which is natural due to causality) is confusing.

In the cases handled here (`dhparam.c` and `dsaparam.c`) there is a further motivation, 
which I mentioned explicitly in the commit message: consistency.
Both apps were partly calling `ERR_print_errors()` before `BIO_printf()` and partly after.

crypto/dh/dh_gen.c
DDvO: > Why not add errors in ffc_params_FIPS186_2_generate and ffc_params_FIPS186_4_generate?

This would have been my preferred solution, but so far there have been no `FFCerr()` calls and `FFC_R_*` codes. 
Shall I introduce those, or move the `DHerr(0, DH_R_BAD_FFC_PARAMETERS)` and `DSAerr(0, DSA_R_BAD_FFC_PARAMETERS)` calls into the appropriate places?

BTW, as mentioned in https://github.com/openssl/openssl/issues/12047#issuecomment-638967154 the best place for the err() calls in my view would be directly in `ffc_params_generate.c`, for instance:
```
static int ffc_validate_LN(size_t L, size_t N, int type)
{
    if (type == FFC_PARAM_TYPE_DH) {
        /* Valid DH L,N parameters from SP800-56Ar3 5.5.1 Table 1 */
        if (L == 2048 && (N == 224 || N == 256))
            return 112;
#ifndef OPENSSL_NO_DH
        DHerr(0, DH_R_BAD_FFC_PARAMETERS);
#endif
    } else if (type == FFC_PARAM_TYPE_DSA) {
        /* Valid DSA L,N parameters from FIPS 186-4 Section 4.2 */
        if (L == 1024 && N == 160)
            return 80;
        if (L == 2048 && (N == 224 || N == 256))
            return 112;
        if (L == 3072 && N == 256)
            return 128;
#ifndef OPENSSL_NO_DSA
        DSAerr(0, DSA_R_BAD_FFC_PARAMETERS);
#endif
    }
    return 0;
}
```

DDvO: I've tentatively moved the error queue entry generation there.
Or should it actually better be in `ffc_params_FIPS186_2_generate` and `ffc_params_FIPS186_4_generate`?

DDvO: We could also unify the names `DSA_R_BAD_FFC_PARAMETERS` and `DH_R_BAD_FFC_PARAMETERS` - if so, to which one?

apps/dhparam.c
levitte: If you look around in other commands, I think you will discover that they print the application level message first and then dump the error queue.  As a matter of fact, some of them have this generic thing at the end of their main function:

``` C
    if (ret != 0)
        ERR_print_errors(bio_err);
```

... which I actually find more appealing than littering the source with a gazillion  `ERR_print_errors(bio_err)` calls.

So in terms of consistency, what you're doing is not.

crypto/dh/dh_gen.c
levitte: I would suggest not using WHATEVERerr(), but use ERR_raise() or ERR_raise_data() instead.  There's much less to deal with there.

I leave it to you to use ERR_LIB_DSA and ERR_LIB_DH, or create ERR_LIB_FFC and use that if you prefer.

levitte: Personally, I'm perfectly OK with using `DSA_R_BAD_FFC_PARAMETERS` for the DSA stuff and `DH_R_BAD_FFC_PARAMETERS` for the DH stuff, and am OK with the error messages in `ffc_validate_LN`

levitte: Actually, I just looked one more time, and am happy with what you did

crypto/ffc/ffc_params_generate.c
```diff
@@ -27,9 +27,13 @@
 #include <string.h> /* memset */
 #include <openssl/sha.h> /* SHA_DIGEST_LENGTH */
 #include <openssl/rand.h>
+#include <openssl/err.h>
+#include <openssl/dherr.h>
+#include <openssl/dsaerr.h>
 #include "crypto/bn.h"
 #include "internal/ffc.h"
 
+
```
levitte, line 36: Unnecessary empty line

DDvO: Oops, removed.

apps/dhparam.c
DDvO: I see your point with the consistency with other apps.

>  As a matter of fact, some of them have this generic thing at the end of their main function:

    if (ret != 0)
        ERR_print_errors(bio_err);

> ... which I actually find more appealing than littering the source with a gazillion ERR_print_errors(bio_err) calls.

This is really a strong argument. Done for these two apps as well.
Also added a couple of missing high-level error messages.

apps/dhparam.c
```diff
@@ -199,7 +199,7 @@ int dhparam_main(int argc, char **argv)
             BN_GENCB *cb  = BN_GENCB_new();
 
             if (cb == NULL) {
-                ERR_print_errors(bio_err);
+                BIO_printf(bio_err, "Out of memory\n");
```
levitte, line 202: I don't think this one is necessary

apps/dhparam.c
```diff
@@ -212,23 +212,23 @@ int dhparam_main(int argc, char **argv)
...
+                           "Error, unable t generate DSA parameters\n");
                 goto end;
             }
 
             dh = DSA_dup_DH(dsa);
             DSA_free(dsa);
             BN_GENCB_free(cb);
             if (dh == NULL) {
-                ERR_print_errors(bio_err);
+                BIO_printf(bio_err, "Out of memory\n");
```
levitte, line 224: ... or this

apps/dhparam.c
```diff
@@ -272,15 +269,14 @@ int dhparam_main(int argc, char **argv)
...
-                ERR_print_errors(bio_err);
+                BIO_printf(bio_err, "Error, unable to load DSA parameters\n");
                 goto end;
             }
 
             dh = DSA_dup_DH(dsa);
             DSA_free(dsa);
             if (dh == NULL) {
-                ERR_print_errors(bio_err);
+                BIO_printf(bio_err, "Out of memory\n");
```
levitte, line 279: ... or this

apps/dhparam.c
DDvO: All right, removed.

apps/dhparam.c
DDvO: Yep

apps/dhparam.c
DDvO: This as well, and a fourth one.

apps/dhparam.c
```diff
@@ -212,23 +210,20 @@ int dhparam_main(int argc, char **argv)
                                                cb)) {
                 DSA_free(dsa);
                 BN_GENCB_free(cb);
-                ERR_print_errors(bio_err);
+                BIO_printf(bio_err, "Error, unable t generate DSA parameters\n");
```
levitte, line 213: t => to

apps/dhparam.c
```diff
@@ -314,7 +302,6 @@ int dhparam_main(int argc, char **argv)
 
     if (check) {
         if (!EVP_PKEY_param_check(ctx) /* DH_check(dh, &i) */) {
-            ERR_print_errors(bio_err);
             BIO_printf(bio_err, "ERROR: Invalid parameters generated\n");
```
levitte, line 305: For consistency, would you mind changing `ERROR:` to `Error,`?

DDvO: Good point, done as well, also for the 2nd such thing.

apps/dhparam.c
DDvO: fixed

DDvO:
Is this good to go now, or should we add/improve any more related error reporting?

DDvO:
@levitte, your new PR #12072 overlaps with the error reporting cleanup I already did in one of the two commits of this PR: https://github.com/openssl/openssl/pull/12062/commits/c77e4e5aaba119fc291aa3d3ddad7d9bf2c10ba9. How to deal with that?

levitte:
> How to deal with that?

I believe this PR will be merged before #12072, so I guess I'll rebase after that?

Either way, it seems that this PR conflicts with the changes from #12048, so you need to do a little bit of rebasing as well :wink:

DDvO:
> I believe this PR will be merged before #12072, so I guess I'll rebase after that?

Sounds good.

> Either way, it seems that this PR conflicts with the changes from #12048, so you need to do a little bit of rebasing as well :wink:

Yeah, right - done.
CI has been happy already.
Happy to approve this now? :smile:

levitte:
> Happy to approve this now?

Obviously

DDvO:
Merged -thanks @levitte
you can now rebase #12072.

PR #12063, opened 2020-06-05
improve SSL_CTX_set_tlsext_ticket_key_cb ref impl

gstrauss:
improve reference implementation code in
  SSL_CTX_set_tlsext_ticket_key_cb man page

change EVP_aes_128_cbc() to EVP_aes_256_cbc(), with the implication
of requiring longer keys.  Updating this code brings the reference
implementation in line with implementation in openssl committed in 2016:
commit 05df5c20
Use AES256 for the default encryption algoritm for TLS session tickets

add comments where user-implementation is needed to complete code

CLA: trivial

##### Checklist

### DIFF
--- a/doc/man3/SSL_CTX_set_tlsext_ticket_key_cb.pod
+++ b/doc/man3/SSL_CTX_set_tlsext_ticket_key_cb.pod
@@ -159,6 +159,7 @@ Reference Implementation:
                                      EVP_MAC_CTX *hctx, int enc)
  {
      OSSL_PARAM params[3];
+     your_type_t *key; /* something that you need to implement */
 
      if (enc) { /* create new session */
          if (RAND_bytes(iv, EVP_MAX_IV_LENGTH) <= 0)
@@ -178,10 +179,10 @@ Reference Implementation:
          }
          memcpy(key_name, key->name, 16);
 
-         EVP_EncryptInit_ex(&ctx, EVP_aes_128_cbc(), NULL, key->aes_key, iv);
+         EVP_EncryptInit_ex(&ctx, EVP_aes_256_cbc(), NULL, key->aes_key, iv);
 
          params[0] = OSSL_PARAM_construct_octet_string(OSSL_MAC_PARAM_KEY,
-                                                       key->hmac_key, 16);
+                                                       key->hmac_key, 32);
          params[1] = OSSL_PARAM_construct_utf8_string(OSSL_MAC_PARAM_DIGEST,
                                                       "sha256", 0);
          params[2] = OSSL_PARAM_construct_end();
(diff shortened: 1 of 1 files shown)


gstrauss:
Should a similar patch be prepared for OpenSSL_1_1_1-stable ?

kaduk:
It's probably best to wait to prepare the 1.1.1 version until the version for master is done/approved, in case anything ends up changing in it.

Also, it's interesting that the commit you linked only changed from AES128 to AES256, but left SHA256 instead of going up to SHA384 or SHA512 as is typically paired with AES256.

gstrauss:
This matches the algorithms used in `ssl/statem/statem_srvr.c:construct_stateless_ticket()`
https://github.com/openssl/openssl/blob/master/ssl/statem/statem_srvr.c#L4015
If stronger algorithms should be recommended, then should the algorithms used in `construct_stateless_ticket()` also be reviewed?

kaduk:
> If stronger algorithms should be recommended, then should the algorithms used in `construct_stateless_ticket()` also be reviewed?

Yes, they should be reviewed.  @mattcaswell please take a look.  I'm inclined to swap out sha256 for sha384, though to be honest I'm not entirely convinced we need to be using AES256 here.  If we were all TLS 1.3 I'd be more solidly in the "go back to AES128" camp, but given how stateless tickets are used in earlier versions we may want the extra protection, to prevent future computing advances from allowing for decryption of previously captured encrypted traffic.

kaduk:
Also, thank you for noticing the discrepancy and submitting the patch!  It is good that we are revisiting what we are actually using in `construct_stateless_ticket()` as a result.

mattcaswell:
>  I'm inclined to swap out sha256 for sha384

That is my inclination too. Let's go with that.

t8m:
> It's probably best to wait to prepare the 1.1.1 version until the version for master is done/approved, in case anything ends up changing in it.
> 
> Also, it's interesting that the commit you linked only changed from AES128 to AES256, but left SHA256 instead of going up to SHA384 or SHA512 as is typically paired with AES256.

If we are talking about HMAC here, we are talking about preimage or second preimage attacks and there is absolutely no point in using anything stronger than SHA256 as that has 256 bit security against preimage and second preimage attacks.

gstrauss:
Is there a consensus on next steps?

Would you like me to extend this patch to update `construct_stateless_ticket()` and the code example to use stronger algorithms, or, as @t8m suggested, is this patch to the code example acceptable?

kaduk:
Sorry for making this more complicated of a process than it needed to be!

gstrauss:
CLA emailed to legal@opensslfoundation

Please let me know if you would like me to update the commit message in this pull request to remove the CLA: trivial tag

Thank you.

mattcaswell:
Close/reopen to kick CLA bot.

mattcaswell:
Close/reopen to kick CLA bot.

mattcaswell:
Pushed to master. Thanks! @gstrauss - if you'd like to prepare a version for 1.1.1 then that would be helpful.

PR #12067, opened 2020-06-06
Fix crypto/evp/pmeth_check.c copyright years

levitte:
This file was created in 2020, not 2006


kroeckx:
It comes from pmeth_gn.c, which existed since 2006. Some of the code in this file is not new, and exist at least since 2017.

slontis:
I think this PR should be closed.

levitte:
Okie

PR #12070, opened 2020-06-06
APPS: Fix invoking openssl without a command

### LINKED ISSUE #12069: master: The commadn "openssl" segfaults
Since commit 
   eca4713913781 ("APPS: Drop interactive mode in the 'openssl' program")

invoking just "openssl" segfaults:
>(gdb) bt
>#0  __strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:65
>#1  0x00005555555ebfd9 in opt_progname (argv0=0x0) at ../apps/lib/opt.c:105
>#2  0x00005555555ecc28 in opt_init (ac=<optimized out>, av=0x7fffffffddd0, o=0x555555621700 <help_options>) at ../apps/lib/opt.c:129
>#3  0x00005555555b4478 in help_main (argc=<optimized out>, argv=<optimized out>) at ../apps/openssl.c:312
>#4  0x0000555555592fba in main (argc=0, argv=0x7fffffffddd0) at ../apps/openssl.c:269

sebastianas:
Invoking help with "empty" argc leads to a segfault.
Invoke do_cmd() with help as argument which invokes help_main() with
proper argv.

Fixes #12069

### DIFF
--- a/apps/openssl.c
+++ b/apps/openssl.c
@@ -205,6 +205,8 @@ static void setup_trace(const char *str)
 }
 #endif /* OPENSSL_NO_TRACE */
 
+static char *help_argv[] = { "help" };
+
 int main(int argc, char *argv[])
 {
     FUNCTION f, *fp;
@@ -268,7 +270,7 @@ int main(int argc, char *argv[])
     /* If there's a command, run with that, otherwise "help". */
     ret = argc > 0
         ? do_cmd(prog, argc, argv)
-        : help_main(argc, argv);
+        : do_cmd(prog, 1, help_argv);
 
  end:
     app_providers_cleanup();


levitte:
Good catch!  Thank you

MarcT512:
Whilst it works, I'm not 100% sure this is the "correct" solution.The backtrace shows:
```
(gdb) bt
#0  0x00007ffff6d63ff6 in strlen () from /lib64/libc.so.6
#1  0x000000000048a957 in opt_progname (argv0=0x0) at apps/lib/opt.c:105
#2  0x000000000048aa03 in opt_init (ac=0, av=0x7fffffffdda0, o=0x4a7ba0 <help_options>) at apps/lib/opt.c:129
#3  0x00000000004495f9 in help_main (argc=0, argv=0x7fffffffdda0) at apps/openssl.c:312
#4  0x0000000000449550 in main (argc=0, argv=0x7fffffffdda0) at apps/openssl.c:271
```
...it segfaults in opt_progname because argv was incremented (incorrectly?) in apps/openssl.c main() at around line 271, and so argv[0] is now null (rather than the program name).

With this PR, help_main() will think the utility was invoked as ./help (rather than ./openssl). Whether this matters, I'm not sure...

I fixed it locally here by doing:
```
-        : help_main(argc, argv);
+        : help_main(1, (fp == NULL) ? --argv:argv);
```
This was also a hack, but means argv[0] is correct in help_main().

levitte:
> With this PR, help_main() will think the utility was invoked as ./help (rather than ./openssl). Whether this matters, I'm not sure...

It's actually correct, that's what it should get (or well, it will get "help", not "./help")

levitte:
Merged

96786ad12308215621142f7cb2ef21a9aca30665 APPS: Fix invoking openssl without a command

sebastianas:
Ehm. I just noticed that commit b75f08cb13ca2 ("Fix segfault in openssl app called with no args.") fixed the same thing differently. So maybe this should be reverted since we don't need both?

mattcaswell:
> Ehm. I just noticed that commit b75f08c ("Fix segfault in openssl app called with no args.") fixed the same thing differently. So maybe this should be reverted since we don't need both?

We now seem to have failures coming from Travis/Appveyor in this area on master. Not sure if it is directly related to this change or not.

levitte:
#12258 fixes this, should anyone wonder.

PR #12072, opened 2020-06-06
[WIP] APPS: Refactoring dsaparam and dhparam

levitte:
## APPS: Refactoring dsaparam and dhparam - error printing

Move all 'ERR_print_errors(bio_err)' calls to one spot, make sure it's
only called on failure.

## APPS: Refactoring dsaparam and dhparam - set EVP_PKEY_CTX params after init

Calling any EVP_PKEY_CTX_set function before the EVP_PKEY_CTX has been
properly initialized (in this case with EVP_PKEY_paramgen_init()) is
dodgy.  The recommended order is init, set all extra params, perform.

While we're at it, we set the param generation generator.

## APPS: Refactoring dhparam - make sure |pkey| is always set

|pkey| would be assigned the DH key when it was generated, but not
otherwise.  This ensures it's always assigned the DH key.

## APPS: Refactoring dhparam - make sure |ctx| is correct when checking

|ctx| was the context that |pkey| was created with, but wouldn't contain
|pkey| as a result.  When using |ctx| to validate the params, it must be
recreated from the |pkey|.

## APPS: Refactoring dhparam - refactor in EVP_PKEY and EVP_PKEY_CTX terms

This also implements do_gendsa_params(), a function shared between
dsaparam.c and dhparam.c.
For good measure, we also add do_gendh_params(), but leave it to be
for dhparam.c only for now.

### DIFF
--- a/apps/dhparam.c
+++ b/apps/dhparam.c
@@ -310,6 +310,13 @@ int dhparam_main(int argc, char **argv)
         EVP_PKEY_print_params(out, pkey, 4, NULL);
 
     if (check) {
+        EVP_PKEY_CTX_free(ctx);
+        ctx = EVP_PKEY_CTX_new(pkey. NULL);
+        if (ctx == NULL) {
+            BIO_printf(bio_err,
+                       "Error, DH key check context allocation failed\n");
+            goto end;
+        }
         if (!EVP_PKEY_param_check(ctx) /* DH_check(dh, &i) */) {
             BIO_printf(bio_err, "Error, invalid parameters generated\n");
             goto end;
--- a/apps/dhparam.c
+++ b/apps/dhparam.c
@@ -297,6 +297,15 @@ int dhparam_main(int argc, char **argv)
     if (dh == NULL)
         goto end;
 
+    /* If |pkey| is non-NULL, we know it contains |dh| */
+    if (pkey == NULL) {
+        if ((pkey = EVP_PKEY_new()) == NULL
+            || !EVP_PKEY_assign_DH(pkey, dh))
+            goto end;
+    }
+
+    /* At this point, |pkey| contains |dh|, without a doubt */
+
     if (text)
         EVP_PKEY_print_params(out, pkey, 4, NULL);
 
--- a/apps/dhparam.c
+++ b/apps/dhparam.c
@@ -32,10 +32,9 @@
 
 #define DEFBITS 2048
 
-#if !defined(OPENSSL_NO_DSA) && !defined(OPENSSL_NO_DEPRECATED_3_0)
-static int dh_cb(int p, int n, BN_GENCB *cb);
-#endif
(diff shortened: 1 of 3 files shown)


levitte:
This includes the current commits from #12048.  Will be rebased as soon as that one's in

levitte:
I hope this gets rid of all strange bugs that the previous refactoring left behind.

levitte:
#12048 is in, so this got rebased and is not WIP any more

levitte:
Ah, need to rebase.  I'll do some squashing while I'm at it.

romen:
I have no idea why it is not failing in Travis, but I am getting a failure on `make cmd-nits`, probably unrelate, but still..

~~~bash
; make cmd-nits
/usr/bin/perl ./util/find-doc-nits -c
doc/man1/openssl-rsa.pod: rsa does not implement -provider
doc/man1/openssl-rsa.pod: rsa does not implement -provider_path
Makefile:3499: recipe for target 'cmd-nits' failed
make: *** [cmd-nits] Error 1
~~~

Also, do these changes need to be applied similarly to `ecparam`?

levitte:
> Also, do these changes need to be applied similarly to `ecparam`?

That one hasn't been touched at all, so no.

Regarding the `rsa` command thing, I have no idea why that appears.  It might go away if I rebase...

romen:
> Regarding the `rsa` command thing, I have no idea why that appears. It might go away if I rebase...

It wasn't, it's a thing that comes out when trying to build with `no-dsa`!

Fix is in #12127

apps/dhparam.c
```diff
@@ -281,49 +241,70 @@ int dhparam_main(int argc, char **argv)
...
 
+#if !defined(OPENSSL_NO_DSA) && !defined(OPENSSL_NO_DEPRECATED_3_0)
+    if (dsaparam) {
+        DSA *dsa;
+        DH *dh;
+
+        dsa = EVP_PKEY_get0_DSA(params);
+        if (dsa == NULL)
+            goto end;
+        dh = DSA_dup_DH(dsa);
```
InfoHunter, line 271: This is worth a `NULL` check although it will not cause problem without the check - but it will result in unnecessary function calls below.

levitte:
I did this work because some of the code was badly ordered...  my changes went a bit further than I really think is worth it, especially since these commands have long been deemed antiquated and to be deprecated at some point.

I'll have a look again, but not immediately

mattcaswell:
> I'll have a look again, but not immediately

@kroeckx added this to the 3.0 beta 1 milestone. Can you at least determine whether that milestone is appropriate for this PR?

levitte:
I've moved the milestone, and marked this WIP.  This is intended as a bug-fix, I can't see anything in here that says "new feature" and warrants being rushed for beta1

levitte:
This has been done better in other PRs

Issue #11650, opened 2020-04-26
3.0-alpha1: "openssl speed" complains about md4/ripemd160 +EdDSA


sebastianas:
Built alpha1 with ../config.

 "openssl speed" reports:

>  Doing md4 for 3s on 16 size blocks: md4 error!
>  40:07:14:74:7B:7F:00:00:error:digital envelope routines:evp_generic_fetch:fetch failed:../crypto/evp/evp_fetch.c:307:Default library context, Algorithm (MD4), Properties ()
>  40:07:14:74:7B:7F:00:00:error:digital envelope routines:EVP_DigestInit_ex:initialization error:../crypto/evp/digest.c:175:
>  Doing md5 for 3s on 16 size blocks: 27556753 md5's in 3.00s
> …
>  Doing whirlpool for 3s on 16384 size blocks: 36288 whirlpool's in 3.00s
>  Doing rmd160 for 3s on 16 size blocks: rmd160 error!
>  40:07:14:74:7B:7F:00:00:error:digital envelope routines:evp_generic_fetch:fetch failed:../crypto/evp/evp_fetch.c:307:Default library context, Algorithm (RIPEMD160), Properties ()
>  40:07:14:74:7B:7F:00:00:error:digital envelope routines:EVP_DigestInit_ex:initialization error:../crypto/evp/digest.c:175:
>  Doing rc4 for 3s on 16 size blocks: 96622041 rc4's in 3.00s
> …
>  Doing 253 bits  ecdh's for 10s: 288312 253-bits ECDH ops in 10.00s
>  Doing 448 bits  ecdh's for 10s: 25217 448-bits ECDH ops in 10.00s
>  Doing 253 bits sign Ed25519's for 10s: 240352 253 bits Ed25519 signs in 10.00s
>  EdDSA verify failure.  No EdDSA verify will be done.
>  40:07:14:74:7B:7F:00:00:error:digital envelope routines:EVP_DigestSignUpdate:called a function you should not call:../crypto/evp/m_sigver.c:312:
>  Doing 456 bits sign Ed448's for 10s: 37818 456 bits Ed448 signs in 10.00s
>  EdDSA verify failure.  No EdDSA verify will be done.
>  40:07:14:74:7B:7F:00:00:error:digital envelope routines:EVP_DigestSignUpdate:called a function you should not call:../crypto/evp/m_sigver.c:312:
[...]

PR #12074, opened 2020-06-06
openssl speed fixes and documentation updates regarding legacy provider

sebastianas:
This addresses the warnings I noticed in `openssl speed' and reported in #11650.
I also added a note about the "legacy provider" to the EVP_$hash(3) man page.

### DIFF
--- a/apps/speed.c
+++ b/apps/speed.c
@@ -575,6 +575,7 @@ typedef struct loopargs_st {
     EC_KEY *ecdsa[ECDSA_NUM];
     EVP_PKEY_CTX *ecdh_ctx[EC_NUM];
     EVP_MD_CTX *eddsa_ctx[EdDSA_NUM];
+    EVP_MD_CTX *eddsa_ctx2[EdDSA_NUM];
 # ifndef OPENSSL_NO_SM2
     EVP_MD_CTX *sm2_ctx[SM2_NUM];
     EVP_MD_CTX *sm2_vfy_ctx[SM2_NUM];
@@ -1242,7 +1243,7 @@ static int EdDSA_verify_loop(void *args)
 {
     loopargs_t *tempargs = *(loopargs_t **) args;
     unsigned char *buf = tempargs->buf;
-    EVP_MD_CTX **edctx = tempargs->eddsa_ctx;
+    EVP_MD_CTX **edctx = tempargs->eddsa_ctx2;
     unsigned char *eddsasig = tempargs->buf2;
     size_t eddsasigsize = tempargs->sigsize;
     int ret, count;
@@ -3389,6 +3390,11 @@ int speed_main(int argc, char **argv)
                 st = 0;
                 break;
             }
+            loopargs[i].eddsa_ctx2[testnum] = EVP_MD_CTX_new();
+            if (loopargs[i].eddsa_ctx2[testnum] == NULL) {
+                st = 0;
+                break;
+            }
 
             if ((ed_pctx = EVP_PKEY_CTX_new_id(ed_curves[testnum].nid, NULL))
                     == NULL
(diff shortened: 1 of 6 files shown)


doc/man3/EVP_md2.pod
```diff
@@ -16,6 +16,10 @@ EVP_md2
 MD2 is a cryptographic hash function standardized in RFC 1319 and designed by
 Ronald Rivest.
 
+=head2 Identity
```
mattcaswell, line 19: I'm not really sure I understand what "Identity" means in this context. I would probably drop this, and just make the sentence below part of the description (and for the other similar occurrence below)

apps/speed.c
```diff
@@ -3397,6 +3412,13 @@ int speed_main(int argc, char **argv)
                 EVP_PKEY_free(ed_pkey);
                 break;
             }
+	    if (!EVP_DigestVerifyInit(loopargs[i].eddsa_ctx2[testnum], NULL,
```
mattcaswell, line 3415: Indentation seems to be off here

mattcaswell: I have to wonder if this ever worked....

doc/man3/EVP_md2.pod
sebastianas: I followed the flow in "doc/man7/EVP_MD-MD2.pod" so it  looks the same.
You want me drop it?

apps/speed.c
sebastianas: So git bisect says here:
> good: [8a5cb59601fa9892e22e26337917cf513d57c473] TEST: Add a test of keygen with >an empty template in test/evp_extra_test.c
> git bisect good 8a5cb59601fa9892e22e26337917cf513d57c473
> bad: [c1e48c5171affd99111fd08e7a0a7e1a76121138] s390: ECX key generation fixes.
> git bisect bad c1e48c5171affd99111fd08e7a0a7e1a76121138
> bad: [43cd37014ef4433ae8e82ba64bddc42cf0bd618a] ecx: add key generation support.
> git bisect bad 43cd37014ef4433ae8e82ba64bddc42cf0bd618a
> skip: [1ee1e551148d963b566a69c2e6e9a583646112fc] Add ECX to FIPS provider as >non-FIPS algorithms
> git bisect skip 1ee1e551148d963b566a69c2e6e9a583646112fc
> only skipped commits left to test
> possible first bad commit: [43cd37014ef4433ae8e82ba64bddc42cf0bd618a] ecx: add key >generation support.
> possible first bad commit: [1ee1e551148d963b566a69c2e6e9a583646112fc] Add ECX to >FIPS provider as non-FIPS algorithms
> c1e48c5171aff s390: ECX key generation fixes.
> 43cd37014ef44 (refs/bisect/bad) ecx: add key generation support.
> 1ee1e551148d9 (refs/bisect/skip-1ee1e551148d963b566a69c2e6e9a583646112fc) Add ECX >to FIPS provider as non-FIPS algorithms
> 8a5cb59601fa9 (refs/bisect/good-8a5cb59601fa9892e22e26337917cf513d57c473) TEST: >Add a test of keygen with an empty template in test/evp_extra_test.c

Would you consider this as a regression then?

mattcaswell: Well, this dubious code seems to have been there since it was first committed in d3a9fb10e...but I've not actually tested to see if it *did* actually somehow work then.

doc/man3/EVP_md2.pod
mattcaswell: Ah, but there the full sentence is:

    This implementation is only available with the legacy provider, and is identified with the name "MD2".

So, there the "identity" is the name "MD2" and so the title makes sense. Here you have dropped the bit from the sentence which is the identity (because it is not relevant in this context) and so the title no longer makes sense. Please just drop the title.

apps/speed.c
sebastianas: Well, according to git bisect either 43cd370 or 1ee1e55 (which could not be tested) led to change in behavior using the same ctx for sign & verify. Since then it is either or.

mattcaswell: Well...either way, its still a bug that should be fixed. Please fix the indentation here, and otherwise this PR looks good.

beldmit:
Merged. Thanks!

Issue #12058, opened 2020-06-05
Typo in doc string? bn_dh.c


tomato42:
`bn_dh.c` refers to RFC 2312:
https://github.com/openssl/openssl/blob/7b2d995ed79f0653defd682e5097ed937f4cabbc/crypto/bn/bn_dh.c#L32

but I can't find anything about DH primes in [RFC 2312](https://tools.ietf.org/html/rfc2312), so I'm guessing it's a typo. Any idea what is the correct RFC reference?

kaduk:
Looks like RFC 2412 (OAKLEY groups)

PR #12076, opened 2020-06-06
fix doc typo in bn_dh.c

tomato42:
while RFC 2312 refers to S/MIME it doesn't actually declare any groups,
RFC 2412 actually talks about DH extensively and the group
defined in the code below is defined on page 47 of it

see also #12058

##### Checklist


beldmit:
Merged. Thanks!

PR #12079, opened 2020-06-07
Avoid use after free in wrap and ocb ciphers

### LINKED ISSUE #12073: OpenSSL 1.1.1 AES key wrap ciphers use-after-free
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23014

This OSS-Fuzz bug triggered only in LibreSSL, not OpenSSL, because OpenSSL master branch is not affected, but the 1.1.1 branch is.

This is a different bug than https://github.com/openssl/openssl/issues/12014

Compile the following poc against the ```OpenSSL_1_1_1``` branch.

```c
#include <openssl/evp.h>

#define CF_CHECK_GTE(expr, res) if ( (expr) < (res) ) { goto end; }
int main(void)
{
const unsigned char key[16] = { 0 };
const unsigned char iv[8] = { 0 };
unsigned char cleartext[16] = { 0 };
unsigned char out[1024];
int len = -1;
EVP_CIPHER_CTX* ctx = EVP_CIPHER_CTX_new();
EVP_CIPHER_CTX* ctx2 = NULL;
const EVP_CIPHER* cipher = EVP_aes_128_wrap();
OPENSSL_init_crypto(OPENSSL_INIT_LOAD_CRYPTO_STRINGS, NULL);
EVP_CIPHER_CTX_set_flags(ctx, EVP_CIPHER_CTX_FLAG_WRAP_ALLOW);

CF_CHECK_EQ(EVP_EncryptInit_ex(ctx, cipher, NULL, NULL, NULL), 1);

CF_CHECK_EQ(EVP_EncryptInit_ex(ctx, NULL, NULL, key, iv), 1);

/* Duplicate the ctx */
ctx2 = EVP_CIPHER_CTX_new();
CF_CHECK_EQ(EVP_CIPHER_CTX_copy(ctx2, ctx), 1);
EVP_CIPHER_CTX_free(ctx);
ctx = ctx2;


/* "the amount of data written may be anything from zero bytes to (inl + cipher_block_size - 1)" */
CF_CHECK_GTE(sizeof(out), sizeof(cleartext) + EVP_CIPHER_block_size(cipher) - 1);
CF_CHECK_EQ(EVP_EncryptUpdate(ctx, out, &len, cleartext, sizeof(cleartext)), 1);
end:
return 0;
}
```

bernd-edlinger:
Fixes #12073

##### Checklist

### DIFF
--- a/crypto/evp/e_aes.c
+++ b/crypto/evp/e_aes.c
@@ -87,7 +87,6 @@ typedef struct {
     int key_set;                /* Set if key initialised */
     int iv_set;                 /* Set if an iv is set */
     OCB128_CONTEXT ocb;
-    unsigned char *iv;          /* Temporary IV store */
     unsigned char tag[16];
     unsigned char data_buf[16]; /* Store partial data blocks */
     unsigned char aad_buf[16];  /* Store partial AAD blocks */
@@ -506,7 +505,7 @@ static int aesni_ocb_init_key(EVP_CIPHER_CTX *ctx, const unsigned char *key,
          * If we have an iv we can set it directly, otherwise use saved IV.
          */
         if (iv == NULL && octx->iv_set)
-            iv = octx->iv;
+            iv = ctx->iv;
         if (iv) {
             if (CRYPTO_ocb128_setiv(&octx->ocb, iv, octx->ivlen, octx->taglen)
                 != 1)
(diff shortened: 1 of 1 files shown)


bernd-edlinger:
The test cases do not execute the changed functions in master, therefore I think removal
of the whole e_aes.c would be better, but my previous experience shows this will not work
either.
I have no idea how this re-design is supposed to work, instead of moving algorithms
to a provider create duplicated code for the provider and leave the original versions
as dead code, maybe used be cms maybe not, maybe used by engines maybe not,
but never tested.

paulidale:
The original code isn't dead.  There are legacy paths which reach it.
It will be removed once we reach the point of being able to remove some of the functions that have been deprecated.

PR #12080, opened 2020-06-07
Test genpkey app for EC keygen with various args

romen:
This PR adds a new recipe to test EC key generation with the `genpkey` CLI app.

For each built-in curve, it tests key generation with text output, in PEM and in DER format, using `explicit` and `named_curve` for parameters encoding.

The list of built-in curves is static at the moment, as this allows to differentiate between prime curves and binary curves to avoid failing when `ec2m` is disabled.

## Issue

At this time these new tests, that are passing in 1.1.1 (#12085) are failing due to changes introduced in #11328.

Issue #12102 tracks the progress on solving the issue.

### `TODO: {}` block

For now, the failing tests are annotated within a [`TODO: {}` block](https://manpages.debian.org/stable/perl-doc/Test::More.3perl.en.html) that informs the testing harness that the tests are expected to fail because development is not yet completed.

The PR fixing #12102 will address also the removal of this `TODO: {}` annotation.

## Credits

Thanks to @bbbrumley and the rest of the team behind the [OpenSSL Triggerflow CI](https://gitlab.com/nisec/openssl-triggerflow-ci) ([paper](https://eprint.iacr.org/2019/366)) that detected and reported this problem!

## Checklist

- [x] issue is created (#12102)
- [x] add more curve aliases

### DIFF
--- a/test/recipes/15-test_genec.t
+++ b/test/recipes/15-test_genec.t
@@ -115,15 +115,16 @@ my @curve_aliases = qw(
 );
 
 my @curve_list = ();
-push(@curve_list, @prime_curves);
-push(@curve_list, @binary_curves)
-    if !disabled("ec2m");
-push(@curve_list, @other_curves);
+#push(@curve_list, @prime_curves);
+#push(@curve_list, @binary_curves)
+#    if !disabled("ec2m");
+#push(@curve_list, @other_curves);
 push(@curve_list, @curve_aliases);
 
 my @params_encodings = ('named_curve', 'explicit');
 
-my @output_formats = ('PEM', 'DER');
+#my @output_formats = ('PEM', 'DER');
+my @output_formats = ();
 
 plan tests => scalar(@curve_list) * scalar(@params_encodings)
     * (1 + scalar(@output_formats)) # Try listed @output_formats and text output
--- a/test/recipes/15-test_genec.t
+++ b/test/recipes/15-test_genec.t
@@ -111,8 +111,24 @@ push(@other_curves, 'SM2')
     if !disabled("sm2");
 
 my @curve_aliases = qw(
+    P-192
+    P-224
     P-256
+    P-384
+    P-521
 );
+push(@curve_aliases, qw(
+    B-163
+    B-233
+    B-283
+    B-409
+    B-571
+    K-163
+    K-233
+    K-283
+    K-409
+    K-571
+)) if !disabled("ec2m");
 
 my @curve_list = ();
 push(@curve_list, @prime_curves);
--- a/test/recipes/15-test_genec.t
+++ b/test/recipes/15-test_genec.t
@@ -131,15 +131,17 @@ push(@curve_aliases, qw(
 )) if !disabled("ec2m");
 
 my @curve_list = ();
-push(@curve_list, @prime_curves);
-push(@curve_list, @binary_curves)
-    if !disabled("ec2m");
-push(@curve_list, @other_curves);
-push(@curve_list, @curve_aliases);
+# push(@curve_list, @prime_curves);
+# push(@curve_list, @binary_curves)
+#     if !disabled("ec2m");
+# push(@curve_list, @other_curves);
+# push(@curve_list, @curve_aliases);
+push(@curve_list, 'P-256');
 
 my @params_encodings = ('named_curve', 'explicit');
 
-my @output_formats = ('PEM', 'DER');
+# my @output_formats = ('PEM', 'DER');
+my @output_formats = ();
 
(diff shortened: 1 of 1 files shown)


romen:
I need @levitte's feedback on how to address the issue.

Currently we have this https://github.com/openssl/openssl/blob/da560600ca83b90cd2a1dbdfe2ee521bad27fb51/crypto/evp/pmeth_lib.c#L1000-L1015

but this was from before further developments for serializers.

If we don't provide a way for the framework to keep track of encoding settings, what are the alternatives?

- modify apps to downgrade the key to a legacy key and use deprecated functions?
- accept the breaking change and alter the documentation saying that `ec_param_enc` in OpenSSL 3.0  is not supported anymore?

romen:
I am not very familiar with the serializer code. But this is another tentative plan for a solution.

- Is serializing a separate operation that can have its own ctx?
- If yes, could this be solved by passing both `pkeyopt`s to both the `keygen` and the `serialize` ctx?
  - `keygen` could then use `ec_paramgen_curve` and ignore `ec_param_enc`
  - `serialize` could then use `ec_param_enc:explicit` (or ignore it if it does not support it) and ignore `ec_paramgen_curve`
  - `apps/genpkey` could detect if `ec_param_enc` has been passed and barf an error if the provider of `serialize` does not support it

romen:
The last set of commits cherry-picks the additional tests (and a fix) included in #12085 (the 1.1.1 counterpart of this PR).

The very last commit disables anyway most of these tests to keep only `P-256` in the list of tested curves, as every curve is currently failing in `master` upon `ec_param_enc:explicit` (and the extra curves are just cluttering the test output with no extra benefit).

levitte:
@romen, #12097 contains stuff to mark up certain tests as TODOs and let the recipe pass

romen:
> @romen, #12097 contains stuff to mark up certain tests as TODOs and let the recipe pass

@levitte I cherry-picked your amendment in this PR, and added a small change to move the subroutines towards the beginning of the file (as it seems more readable to me).

Can you find in the documentation if the TODO annotation in `Test::More` is supported in all the Perl versions we care about?

levitte:
I looked through early Test::More changelog, and it shows that the TODO feature has been around for a *long* time (12 years back at least)

test/recipes/15-test_genec.t
```diff
@@ -0,0 +1,211 @@
...
+push(@curve_list, @prime_curves);
+push(@curve_list, @binary_curves)
+    if !disabled("ec2m");
+push(@curve_list, @other_curves);
+push(@curve_list, @curve_aliases);
+
+my %params_encodings =
+    (
+     'named_curve'      => \&supported,
+     'explicit'         => \&unsupported
```
levitte, line 168: Another way would be this:

```suggestion
     'named_curve'      => sub { my $str = shift;
                                 ok(run(@_), $str); },
     'explicit'         => sub { my $str = shift;
                                 TODO: { local $TODO = "Currently not supported";
                                         ok(run(@_), $str); }
                               },
```

levitte: That would get rid of the function definitions higher up.

romen: I think the current approach is fine, using `\&<name>`.

What I find a bit unreadable of having the subroutines at the end is that when you first encounter `\&supported` or `\&unsupported` reading from top to bottom, there is very little context to understand what is going on.

Having them first, with the comment describing what is happening, makes it easier to grasp what is going on in the associative arrays and the loops even for Perl-scared people like me!

romen:
I took this out of `WIP`:
- [x] title edited
- [x] `draft` status removed in GitHub
- [x] description updated
- [x] separate issue to track the defect created (#12102)

romen:
@levitte would you be willing to review this?

romen:
@levitte would this require further changes?
Do you prefer the lambda option to the current one?

romen:
Ping @levitte

levitte:
Why was this added to the "New Core + FIPS" project?

romen:
> Why was this added to the "New Core + FIPS" project?

It's adding tests for something we broke during the development of the new core+providers architecture.
It's not enough to qualify?

levitte:
> It's adding tests for something we broke during the development of the new core+providers architecture.
> It's not enough to qualify?

Not in my opinion.  It's *affected* by new core / fips work, but itself doesn't touch anything that's distinctly new core / fips.

romen:
Merged to `master` with:

- 0c2bddb76a Test genpkey app for EC keygen with various args
- c65b1d0252 TEST: Add TODO segments in test/recipes/15-test_genec.t


Thanks!

PR #12085, opened 2020-06-08
[1.1.1] Test genpkey app for EC keygen with various args

romen:
This is the equivalent of #12080 for the 1.1.1 branch.

The difference here, and the reason why this should be evaluated on its own and is already out of WIP is that in 1.1.1 these new tests are already passing, while #12080 is held back until the issue they highlight is fixed.

Notice also that this PR already adds new aliases and makes testing on SM2 dependent on enabling the sm2 feature during config.

## Credits

Thanks to @bbbrumley and the rest of the team behind the [OpenSSL Triggerflow CI](https://gitlab.com/nisec/openssl-triggerflow-ci) ([paper](https://eprint.iacr.org/2019/366)) that detected and reported the original issue that led to the development of these new tests.

## Checklist

- [x] add more curve aliases

### DIFF
--- a/test/recipes/15-test_genec.t
+++ b/test/recipes/15-test_genec.t
@@ -111,8 +111,24 @@ push(@other_curves, 'SM2')
     if !disabled("sm2");
 
 my @curve_aliases = qw(
+    P-192
+    P-224
     P-256
+    P-384
+    P-521
 );
+push(@curve_aliases, qw(
+    B-163
+    B-233
+    B-283
+    B-409
+    B-571
+    K-163
+    K-233
+    K-283
+    K-409
+    K-571
+)) if !disabled("ec2m");
 
 my @curve_list = ();
 push(@curve_list, @prime_curves);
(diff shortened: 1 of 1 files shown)


romen:
Can I get also another @openssl/otc confirmation that it is fine in this case to go ahead and include these new tests in `1.1.1` even if they are not included in `master` yet?

As stated in the description, the reason for the exemption from the normal procedure of _"first in master, then backport"_ is that `master` is currently broken w.r.t. the aspect of CLI usage tested by the new tests, so #12080 will stay in `WIP` until we decide how to fix `master`.

romen:
So, we discussed #12080 in a conf call: for now the new tests that are failing are going to be annotated with a `TODO` label, and a separate issue will be opened to address the issue before the final release of 3.0.0.

The TODO annotation is not useful in 1.1.1, so that change won't be included in this PR, which effectively becomes independent from #12080 for now.

So this is now ready to merge!

Issue #12064, opened 2020-06-06
OCSP sign does not RSA_METHOD_FLAG_NO_CHECK


ashman-p:
OCSP_basic_sign_ctx() in ocsp_srv.c , does not check for RSA_METHOD_FLAG_NO_CHECK. If RSA_set_flags() to enable RSA_METHOD_FLAG_NO_CHECK, then OCSP sign operations can fail because the X509_check_private_key().
A similar condition could exist in ocsp_cl.c. But I have not run into a failure there.
See attached patch file containing a possible solution.
[ocsp_srv.c.diff.txt](https://github.com/openssl/openssl/files/4739442/ocsp_srv.c.diff.txt)

 ```
apps/openssl version -a 
OpenSSL 1.1.1h-dev  xx XXX xxxx
built on: Sat Jun  6 05:57:21 2020 UTC
platform: darwin64-x86_64-cc
options:  bn(64,64) rc4(int) des(int) idea(int) blowfish(ptr) 
compiler: cc -fPIC -arch x86_64 -O3 -Wall -DL_ENDIAN -DOPENSSL_PIC -D_REENTRANT -DNDEBUG
OPENSSLDIR: "/usr/local/ssl"
ENGINESDIR: "/usr/local/lib/engines-1.1"
Seeding source: os-specific
 ```
 ```
perl configdata.pm --dump

Command line (with current working directory = .):

    perl ./Configure darwin64-x86_64-cc no-asm

Perl information:

    perl
    5.18.4 for darwin-thread-multi-2level

Enabled features:

    aria
    async
    autoalginit
    autoerrinit
    autoload-config
    bf
    blake2
    buildtest-c\+\+
    camellia
    capieng
    cast
    chacha
    cmac
    cms
    comp
    ct
    deprecated
    des
    dgram
    dh
    dsa
    dso
    dtls
    dynamic-engine
    ec
    ec2m
    ecdh
    ecdsa
    engine
    err
    filenames
    gost
    hw(-.+)?
    idea
    makedepend
    md4
    mdc2
    multiblock
    nextprotoneg
    pinshared
    ocb
    ocsp
    pic
    poly1305
    posix-io
    psk
    rc2
    rc4
    rdrand
    rfc3779
    rmd160
    scrypt
    seed
    shared
    siphash
    sm2
    sm3
    sm4
    sock
    srp
    srtp
    sse2
    ssl
    static-engine
    stdio
    tests
    threads
    tls
    ts
    ui-console
[...]

PR #12087, opened 2020-06-08
OCSP sign does not RSA_METHOD_FLAG_NO_CHECK

ashman-p:
OCSP_basic_sign_ctx() in ocsp_srv.c , and
OCSP_request_sign() in ocsp_cl.c does not check for RSA_METHOD_FLAG_NO_CHECK. If RSA_set_flags() was csalled to enable RSA_METHOD_FLAG_NO_CHECK, then OCSP sign operations can fail (because the X509_check_private_key()).

##### Checklist

### DIFF
--- a/crypto/rsa/rsa_ameth.c
+++ b/crypto/rsa/rsa_ameth.c
@@ -122,14 +122,11 @@ static int rsa_pub_cmp(const EVP_PKEY *a, const EVP_PKEY *b)
      * Don't check the public/private key, this is mostly for smart
      * cards.
      */
-    if (((a->type == EVP_PKEY_RSA)
-            && (RSA_flags(a->pkey.rsa) & RSA_METHOD_FLAG_NO_CHECK))
-            ||
-        ((b->type == EVP_PKEY_RSA)
-            && (RSA_flags(b->pkey.rsa) & RSA_METHOD_FLAG_NO_CHECK))
-       )
-        ; /* no-op */
-    else if (BN_cmp(b->pkey.rsa->n, a->pkey.rsa->n) != 0
+    if (((RSA_flags(a->pkey.rsa) & RSA_METHOD_FLAG_NO_CHECK))
+       || (RSA_flags(b->pkey.rsa) & RSA_METHOD_FLAG_NO_CHECK))
+        return 1;
+
+    if (BN_cmp(b->pkey.rsa->n, a->pkey.rsa->n) != 0
         || BN_cmp(b->pkey.rsa->e, a->pkey.rsa->e) != 0)
         return 0;
     return 1;
--- a/crypto/rsa/rsa_ameth.c
+++ b/crypto/rsa/rsa_ameth.c
@@ -118,7 +118,18 @@ static int rsa_pub_decode(EVP_PKEY *pkey, X509_PUBKEY *pubkey)
 
 static int rsa_pub_cmp(const EVP_PKEY *a, const EVP_PKEY *b)
 {
-    if (BN_cmp(b->pkey.rsa->n, a->pkey.rsa->n) != 0
+    /*
+     * Don't check the public/private key, this is mostly for smart
+     * cards.
+     */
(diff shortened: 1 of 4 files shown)


ashman-p:
The is related to issue https://github.com/openssl/openssl/issues/12064

mattcaswell:
It seems to me that the current handling of `RSA_METHOD_FLAG_NO_CHECK` is just totally wrong - and implemented at the wrong level of abstraction. It is a flag set on the RSA key which is only ever respected by libssl and gets checked 3 times - twice before calls to X509_check_private_key() and once before a call to EVP_PKEY_cmp(). That final instance seems to more-or-less replicate the logic inside X509_check_private_key() which ends up calling EVP_PKEY_cmp() to actually check the consistency of the public and private key.

So everything always ends up in EVP_PKEY_cmp(). For RSA keys this is implemented by `rsa_pub_cmp` which looks like this:

https://github.com/openssl/openssl/blob/7f699cb663741a73cfe95214d4a39a1078c94294/crypto/rsa/rsa_ameth.c#L119-L125

This means that we will have to propagate this flag check everywhere where a consistency check is made to be sure we've got every instance.

A possible alternative approach would be to strip the `RSA_METHOD_FLAG_NO_CHECK` checks from everywhere, and instead move them to `rsa_pub_cmp` causing it to always return 1 if the flag is set. This will automatically ensure we do the right thing everywhere, and I can't see why it would break anything.

kaduk:
My (admittedly quite vague) recollection is that the `RSA_METHOD_FLAG_NO_CHECK` mechanism was supposed to be superseded by the `RSA_FLAG_EXT_PKEY` flag.

mattcaswell:
> My (admittedly quite vague) recollection is that the RSA_METHOD_FLAG_NO_CHECK mechanism was supposed to be superseded by the RSA_FLAG_EXT_PKEY flag.

These flags seem related but different:

````
# define RSA_METHOD_FLAG_NO_CHECK        0x0001/* don't check pub/private
                                                * match */
````
Compared to
````
/*
 * This flag means the private key operations will be handled by rsa_mod_exp
 * and that they do not depend on the private key components being present:
 * for example a key stored in external hardware. Without this flag
 * bn_mod_exp gets called when private key components are absent.
 */
# define RSA_FLAG_EXT_PKEY               0x0020
````

File: crypto/rsa/rsa_ameth.c [hunk 1]
```diff
@@ -118,7 +118,18 @@ static int rsa_pub_decode(EVP_PKEY *pkey, X509_PUBKEY *pubkey)
 
 static int rsa_pub_cmp(const EVP_PKEY *a, const EVP_PKEY *b)
 {
-    if (BN_cmp(b->pkey.rsa->n, a->pkey.rsa->n) != 0
+    /*
+     * Don't check the public/private key, this is mostly for smart
+     * cards.
+     */
+    if (((a->type == EVP_PKEY_RSA)
```
mattcaswell, line 125: This check is probably unnecessary. By virtue of being in this function we must have an RSA key type.

crypto/rsa/rsa_ameth.c
```diff
@@ -118,7 +118,18 @@ static int rsa_pub_decode(EVP_PKEY *pkey, X509_PUBKEY *pubkey)
...
+            && (RSA_flags(a->pkey.rsa) & RSA_METHOD_FLAG_NO_CHECK))
+            ||
+        ((b->type == EVP_PKEY_RSA)
+            && (RSA_flags(b->pkey.rsa) & RSA_METHOD_FLAG_NO_CHECK))
+       )
+        ; /* no-op */
```
mattcaswell, line 131: I would prefer simply to return here, and drop the "else" from the beginning of the line below.

ashman-p:
CLA and ICLA updated and emailed.

ashman-p:
Replaced by new updated PR 
OCSP sign does not RSA_METHOD_FLAG_NO_CHECK 2 #12419

Issue #12077, opened 2020-06-07
openssl command not responding when create key in linux


rangpec:
I tried to create key in my linux server below openssl command, I don't get any response from openssl. It is waiting for longtime. I don't get any error message. Even if I wait for 1hr I don't see any error/response message.

```
openssl req -new -newkey rsa:4096 -days 365 -nodes -x509 -subj "/C=""/ST=""/L=""/O=""/CN=sim.nsr.ssg.tnsi.com" -keyout server.key -out server.cert
Generating a RSA private key

```

I have attached the output of "perl configdata.pm --dump" command
Here is openssl version intalled:

```

[root@asb-dev-idh-rs01 openssl-OpenSSL_1_1_1-stable]# openssl version
OpenSSL 1.1.1h-dev  xx XXX xxxx
[root@asb-dev-idh-rs01 openssl-OpenSSL_1_1_1-stable]# openssl version -a
OpenSSL 1.1.1h-dev  xx XXX xxxx
built on: Wed May 27 07:28:07 2020 UTC
platform: linux-x86_64
options:  bn(64,64) rc4(16x,int) des(int) idea(int) blowfish(ptr)
compiler: gcc -fPIC -pthread -m64 -Wa,--noexecstack -Wall -O3 -DOPENSSL_USE_NODELETE -DL_ENDIAN -DOPENSSL_PIC -DOPENSSL_CPUID_OBJ -DOPENSSL_IA32_SSE2 -DOPENSSL_BN_ASM_MONT -DOPENSSL_BN_ASM_MONT5 -DOPENSSL_BN_ASM_GF2m -DSHA1_ASM -DSHA256_ASM -DSHA512_ASM -DKECCAK1600_ASM -DRC4_ASM -DMD5_ASM -DAESNI_ASM -DVPAES_ASM -DGHASH_ASM -DECP_NISTZ256_ASM -DX25519_ASM -DPOLY1305_ASM -DNDEBUG
OPENSSLDIR: "/usr/local/ssl"
ENGINESDIR: "/usr/local/lib64/engines-1.1"
Seeding source: os-specific
[root@asb-dev-idh-rs01 openssl-OpenSSL_1_1_1-stable]#

```
Could someone help me why I am unable to generate key?
Actually I trying to upgrade my python script to support TLS1.2 version. my python script hangs while creating SSLContext(). I suspect this issue also due to openssl. 

I need really help to resolve this openssl issue. Advance Thanks!!

[...]

PR #12095, opened 2020-06-09
APPS key/param gen: Add warning on delay due to missing entropy; clean up error handling

DDvO:
This has been motivated by #12077.

This PR also introduces `app_keygen()` and `app_paramgen()`,
factoring out the respective low-level calls and the diagnostics around them
in `dgst`, `dhparam`, `dsaparam`, `gendsa`, `genpkey`, `genrsa`, and `req`,
and cleans up error reporting and coding style in a couple of these apps.

### DIFF
--- a/apps/dgst.c
+++ b/apps/dgst.c
@@ -106,7 +106,7 @@ int dgst_main(int argc, char **argv)
     const char *md_name = NULL;
     OPTION_CHOICE o;
     int separator = 0, debug = 0, keyform = FORMAT_UNDEF, siglen = 0;
-    int i, ret = 1, out_bin = -1, want_pub = 0, do_verify = 0;
+    int i, ret = EXIT_FAILURE, out_bin = -1, want_pub = 0, do_verify = 0;
     int xoflen = 0;
     unsigned char *buf = NULL, *sigbuf = NULL;
     int engine_impl = 0;
@@ -125,7 +125,7 @@ int dgst_main(int argc, char **argv)
             goto end;
         case OPT_HELP:
             opt_help(dgst_options);
-            ret = 0;
+            ret = EXIT_SUCCESS;
             goto end;
         case OPT_LIST:
             BIO_printf(bio_out, "Supported digests:\n");
@@ -134,7 +134,7 @@ int dgst_main(int argc, char **argv)
             OBJ_NAME_do_all_sorted(OBJ_NAME_TYPE_MD_METH,
                                    show_digests, &dec);
             BIO_printf(bio_out, "\n");
-            ret = 0;
+            ret = EXIT_SUCCESS;
             goto end;
         case OPT_C:
             separator = 1;
@@ -244,10 +244,8 @@ int dgst_main(int argc, char **argv)
 
     in = BIO_new(BIO_s_file());
     bmd = BIO_new(BIO_f_md());
-    if ((in == NULL) || (bmd == NULL)) {
-        ERR_print_errors(bio_err);
+    if (in == NULL || bmd == NULL)
         goto end;
-    }
 
     if (debug) {
         BIO_set_callback(in, BIO_debug_callback);
@@ -272,7 +270,7 @@ int dgst_main(int argc, char **argv)
         goto end;
 
     if ((!(mac_name == NULL) + !(keyfile == NULL) + !(hmac_key == NULL)) > 1) {
-        BIO_printf(bio_err, "MAC and Signing key cannot both be specified\n");
+        BIO_printf(bio_err, "MAC and signing key cannot both be specified\n");
         goto end;
     }
 
@@ -302,31 +300,24 @@ int dgst_main(int argc, char **argv)
 
     if (mac_name != NULL) {
         EVP_PKEY_CTX *mac_ctx = NULL;
-        int r = 0;
+
         if (!init_gen_str(&mac_ctx, mac_name, impl, 0, NULL, NULL))
-            goto mac_end;
+            goto end;
         if (macopts != NULL) {
-            char *macopt;
             for (i = 0; i < sk_OPENSSL_STRING_num(macopts); i++) {
-                macopt = sk_OPENSSL_STRING_value(macopts, i);
+                char *macopt = sk_OPENSSL_STRING_value(macopts, i);
+
                 if (pkey_ctrl_string(mac_ctx, macopt) <= 0) {
-                    BIO_printf(bio_err,
-                               "MAC parameter error \"%s\"\n", macopt);
-                    ERR_print_errors(bio_err);
-                    goto mac_end;
+                    EVP_PKEY_CTX_free(mac_ctx);
+                    BIO_printf(bio_err, "MAC parameter error \"%s\"\n", macopt);
+                    goto end;
                 }
             }
         }
-        if (EVP_PKEY_keygen(mac_ctx, &sigkey) <= 0) {
-            BIO_puts(bio_err, "Error generating key\n");
-            ERR_print_errors(bio_err);
-            goto mac_end;
-        }
-        r = 1;
- mac_end:
+
+        sigkey = app_keygen(mac_ctx, mac_name, 0, 0 /* not verbose */);
+        /* Verbose output would make external-tests gost-engine fail */
         EVP_PKEY_CTX_free(mac_ctx);
-        if (r == 0)
-            goto end;
     }
 
     if (hmac_key != NULL) {
@@ -342,28 +333,27 @@ int dgst_main(int argc, char **argv)
     if (sigkey != NULL) {
         EVP_MD_CTX *mctx = NULL;
         EVP_PKEY_CTX *pctx = NULL;
-        int r;
+        int res;
+
         if (!BIO_get_md_ctx(bmd, &mctx)) {
             BIO_printf(bio_err, "Error getting context\n");
-            ERR_print_errors(bio_err);
             goto end;
         }
         if (do_verify)
-            r = EVP_DigestVerifyInit(mctx, &pctx, md, impl, sigkey);
+            res = EVP_DigestVerifyInit(mctx, &pctx, md, impl, sigkey);
         else
-            r = EVP_DigestSignInit(mctx, &pctx, md, impl, sigkey);
-        if (!r) {
+            res = EVP_DigestSignInit(mctx, &pctx, md, impl, sigkey);
+        if (res == 0) {
             BIO_printf(bio_err, "Error setting context\n");
-            ERR_print_errors(bio_err);
             goto end;
         }
         if (sigopts != NULL) {
-            char *sigopt;
             for (i = 0; i < sk_OPENSSL_STRING_num(sigopts); i++) {
-                sigopt = sk_OPENSSL_STRING_value(sigopts, i);
+                char *sigopt = sk_OPENSSL_STRING_value(sigopts, i);
+
                 if (pkey_ctrl_string(pctx, sigopt) <= 0) {
-                    BIO_printf(bio_err, "parameter error \"%s\"\n", sigopt);
-                    ERR_print_errors(bio_err);
+                    BIO_printf(bio_err, "Signature parameter error \"%s\"\n",
+                               sigopt);
                     goto end;
                 }
             }
@@ -374,23 +364,21 @@ int dgst_main(int argc, char **argv)
         EVP_MD_CTX *mctx = NULL;
         if (!BIO_get_md_ctx(bmd, &mctx)) {
             BIO_printf(bio_err, "Error getting context\n");
-            ERR_print_errors(bio_err);
             goto end;
         }
         if (md == NULL)
             md = (EVP_MD *)EVP_sha256();
         if (!EVP_DigestInit_ex(mctx, md, impl)) {
             BIO_printf(bio_err, "Error setting digest\n");
-            ERR_print_errors(bio_err);
             goto end;
         }
     }
 
     if (sigfile != NULL && sigkey != NULL) {
         BIO *sigbio = BIO_new_file(sigfile, "rb");
+
         if (sigbio == NULL) {
             BIO_printf(bio_err, "Error opening signature file %s\n", sigfile);
-            ERR_print_errors(bio_err);
             goto end;
         }
         siglen = EVP_PKEY_size(sigkey);
@@ -399,7 +387,6 @@ int dgst_main(int argc, char **argv)
         BIO_free(sigbio);
         if (siglen <= 0) {
             BIO_printf(bio_err, "Error reading signature file %s\n", sigfile);
-            ERR_print_errors(bio_err);
             goto end;
         }
     }
@@ -431,27 +418,28 @@ int dgst_main(int argc, char **argv)
                     siglen, NULL, md_name, "stdin");
     } else {
         const char *sig_name = NULL;
-        if (!out_bin) {
+
+        if (out_bin == 0) {
             if (sigkey != NULL)
                 sig_name = EVP_PKEY_get0_type_name(sigkey);
         }
-        ret = 0;
+        ret = EXIT_SUCCESS;
         for (i = 0; i < argc; i++) {
-            int r;
             if (BIO_read_filename(in, argv[i]) <= 0) {
                 perror(argv[i]);
-                ret++;
+                ret = EXIT_FAILURE;
                 continue;
             } else {
-                r = do_fp(out, buf, inp, separator, out_bin, xoflen,
-                          sigkey, sigbuf, siglen, sig_name, md_name, argv[i]);
+                if (do_fp(out, buf, inp, separator, out_bin, xoflen,
+                          sigkey, sigbuf, siglen, sig_name, md_name, argv[i]))
+                    ret = EXIT_FAILURE;
             }
-            if (r)
-                ret = r;
             (void)BIO_reset(bmd);
         }
     }
  end:
+    if (ret != EXIT_SUCCESS)
+        ERR_print_errors(bio_err);
     OPENSSL_clear_free(buf, BUFSIZE);
     BIO_free(in);
     OPENSSL_free(passin);
@@ -537,14 +525,13 @@ int do_fp(BIO *out, unsigned char *buf, BIO *bp, int sep, int binout, int xoflen
           const char *file)
 {
     size_t len = BUFSIZE;
-    int i, backslash = 0, ret = 1;
+    int i, backslash = 0, ret = EXIT_FAILURE;
     unsigned char *allocated_buf = NULL;
 
     while (BIO_pending(bp) || !BIO_eof(bp)) {
         i = BIO_read(bp, (char *)buf, BUFSIZE);
         if (i < 0) {
-            BIO_printf(bio_err, "Read Error in %s\n", file);
-            ERR_print_errors(bio_err);
+            BIO_printf(bio_err, "Read error in %s\n", file);
             goto end;
         }
         if (i == 0)
@@ -557,14 +544,13 @@ int do_fp(BIO *out, unsigned char *buf, BIO *bp, int sep, int binout, int xoflen
         if (i > 0) {
             BIO_printf(out, "Verified OK\n");
         } else if (i == 0) {
-            BIO_printf(out, "Verification Failure\n");
+            BIO_printf(out, "Verification failure\n");
             goto end;
         } else {
-            BIO_printf(bio_err, "Error Verifying Data\n");
-            ERR_print_errors(bio_err);
+            BIO_printf(bio_err, "Error verifying data\n");
             goto end;
         }
-        ret = 0;
+        ret = EXIT_SUCCESS;
         goto end;
     }
     if (key != NULL) {
@@ -573,8 +559,7 @@ int do_fp(BIO *out, unsigned char *buf, BIO *bp, int sep, int binout, int xoflen
 
         BIO_get_md_ctx(bp, &ctx);
         if (!EVP_DigestSignFinal(ctx, NULL, &tmplen)) {
-            BIO_printf(bio_err, "Error Signing Data\n");
-            ERR_print_errors(bio_err);
+            BIO_printf(bio_err, "Error getting maximum length of signed data\n");
             goto end;
         }
         if (tmplen > BUFSIZE) {
@@ -583,8 +568,7 @@ int do_fp(BIO *out, unsigned char *buf, BIO *bp, int sep, int binout, int xoflen
             buf = allocated_buf;
         }
         if (!EVP_DigestSignFinal(ctx, buf, &len)) {
-            BIO_printf(bio_err, "Error Signing Data\n");
-            ERR_print_errors(bio_err);
+            BIO_printf(bio_err, "Error signing data\n");
             goto end;
         }
     } else if (xoflen > 0) {
@@ -600,15 +584,12 @@ int do_fp(BIO *out, unsigned char *buf, BIO *bp, int sep, int binout, int xoflen
 
         if (!EVP_DigestFinalXOF(ctx, buf, len)) {
             BIO_printf(bio_err, "Error Digesting Data\n");
-            ERR_print_errors(bio_err);
             goto end;
         }
     } else {
         len = BIO_gets(bp, (char *)buf, BUFSIZE);
-        if ((int)len < 0) {
-            ERR_print_errors(bio_err);
+        if ((int)len < 0)
             goto end;
-        }
     }
 
     if (binout) {
@@ -643,7 +624,7 @@ int do_fp(BIO *out, unsigned char *buf, BIO *bp, int sep, int binout, int xoflen
         BIO_printf(out, "\n");
     }
 
-    ret = 0;
+    ret = EXIT_SUCCESS;
  end:
     if (allocated_buf != NULL)
         OPENSSL_clear_free(allocated_buf, len);
--- a/apps/dhparam.c
+++ b/apps/dhparam.c
@@ -217,11 +217,7 @@ int dhparam_main(int argc, char **argv)
             }
         }
 
-        if (!EVP_PKEY_paramgen(ctx, &tmppkey)) {
-            BIO_printf(bio_err, "Error, %s generation failed\n", alg);
-            goto end;
-        }
-
+        tmppkey = app_paramgen(ctx, alg);
         EVP_PKEY_CTX_free(ctx);
         ctx = NULL;
         if (dsaparam) {
--- a/apps/dsaparam.c
+++ b/apps/dsaparam.c
@@ -11,6 +11,7 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#include "apps.h"
 #include <time.h>
 #include <string.h>
 #include "apps.h"
@@ -176,10 +177,7 @@ int dsaparam_main(int argc, char **argv)
                        "Error, DSA key generation setting bit length failed\n");
             goto end;
         }
-        if (EVP_PKEY_paramgen(ctx, &params) <= 0) {
-            BIO_printf(bio_err, "Error, DSA key generation failed\n");
-            goto end;
-        }
+        params = app_paramgen(ctx, "DSA");
     } else {
         params = load_keyparams(infile, informat, 1, "DSA", "DSA parameters");
     }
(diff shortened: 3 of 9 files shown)


DDvO:
According to a hint by @kroeckx I've revised this to issue the warning only if `RAND_status()` returns `false`.

So as expected on my system the warning disappears. 
Hope it is still given whenever a system has poor entropy.

kroeckx:
On Tue, Jun 09, 2020 at 02:30:38AM -0700, David von Oheimb wrote:
> According to a hint by @kroeckx I've revised this to issue the warning only if `RAND_status()` returns `false`.

I'm not sure just calling it is enough. It could be that on a
system that don't have enough entropy, the current drbg_status()
will just hang itself. If that's the case, we should probably look
into fixing that.

DDvO:
>> I've revised this to issue the warning only if `RAND_status()` returns `false`.

> I'm not sure just calling it is enough. It could be that on a system that don't have enough entropy, the current drbg_status() will just hang itself. If that's the case, we should probably look into fixing that.

Good point. So I've marked this PR as WIP and will await this being sorted out.

apps/genpkey.c
```diff
@@ -99,7 +100,7 @@ int genpkey_main(int argc, char **argv)
                 goto end;
             break;
         case OPT_ALGORITHM:
-            if (!init_gen_str(&ctx, opt_arg(), e, do_param))
+            if (!init_gen_str(&ctx, algname = opt_arg(), e, do_param))
```
richsalz, line 103: Please make this two lines.

apps/lib/apps.c
```diff
@@ -2720,3 +2721,30 @@ void app_params_free(OSSL_PARAM *params)
...
+    return 0;
+}
+
+int app_paramgen(EVP_PKEY_CTX *ctx, EVP_PKEY **pkey)
+{
+    if (!RAND_status())
+        BIO_printf(bio_err, "Warning: ganerating random key parameters may take long time if the system has a poor entropy source\n");
+    if (EVP_PKEY_paramgen(ctx, pkey) > 0)
+        return 1;
+    BIO_puts(bio_err, "Generating key parameters failed\n");
```
richsalz, line 2748: should these follow #11993 and even use app_bail_out?

apps/req.c
```diff
@@ -490,7 +490,6 @@ int req_main(int argc, char **argv)
             if (oid_bio == NULL) {
```
richsalz, line 490: Why try to proceed if this fails?  The user specified an OID file and you're not opening it.  Should use ```bio_open_default```, right?

apps/req.c
```diff
@@ -598,41 +597,41 @@ int req_main(int argc, char **argv)
     if (newreq && (pkey == NULL)) {
         app_RAND_load_conf(req_conf, section);
 
-        if (!NCONF_get_number(req_conf, section, BITS, &newkey)) {
-            newkey = DEFAULT_KEY_LENGTH;
+        if (!NCONF_get_number(req_conf, section, BITS, &newkey_len)) {
```
richsalz, line 600: extra curly braces.

apps/req.c
```diff
@@ -799,10 +790,8 @@ int req_main(int argc, char **argv)
             }
 
             i = do_X509_sign(x509ss, pkey, digest, sigopts);
-            if (!i) {
-                ERR_print_errors(bio_err);
+            if (!i)
```
richsalz, line 793: explicit test against zero?

apps/req.c
```diff
@@ -868,14 +855,13 @@ int req_main(int argc, char **argv)
 
         i = do_X509_REQ_verify(req, tpubkey, vfyopts);
 
-        if (i < 0) {
+        if (i < 0)
             goto end;
-        } else if (i == 0) {
+        else if (i == 0)
```
richsalz, line 860: else after goto is an issue for me, FWIW.

apps/lib/apps.c
```diff
@@ -2720,3 +2721,30 @@ void app_params_free(OSSL_PARAM *params)
...
+{
+    if (alg != NULL) {
+        if (bits > 0)
+            BIO_printf(bio_err, "Generating a %s key, with %d bits\n",
+                       alg, bits);
+        else
+            BIO_printf(bio_err, "Generating a %s key\n", alg);
+    }
+    if (!RAND_status())
+        BIO_printf(bio_err, "Warning: generating random key material may take long time if the system has a poor entropy source\n");
```
kaduk, line 2735: This line is quite long; should we add a linebreak (both a printed one and a code one)?
Also, nit, "a long time".

apps/lib/apps.c
```diff
@@ -2720,3 +2721,30 @@ void app_params_free(OSSL_PARAM *params)
...
+    if (EVP_PKEY_keygen(ctx, pkey) > 0)
+        return 1;
```
kaduk, line 2737: Will the keygen callback be pushing errors on the stack that we need to print?  There are lots of code paths that call these helpers now but don't seem to have an `ERR_print_errors()` anymore.

apps/lib/apps.c
```diff
@@ -2720,3 +2721,30 @@ void app_params_free(OSSL_PARAM *params)
...
+    BIO_puts(bio_err, "Generating private key failed\n");
+    return 0;
+}
+
+int app_paramgen(EVP_PKEY_CTX *ctx, EVP_PKEY **pkey)
+{
+    if (!RAND_status())
+        BIO_printf(bio_err, "Warning: ganerating random key parameters may take long time if the system has a poor entropy source\n");
```
kaduk, line 2745: (same as above)

apps/req.c
```diff
@@ -255,7 +255,7 @@ int req_main(int argc, char **argv)
     int informat = FORMAT_PEM, outformat = FORMAT_PEM, keyform = FORMAT_PEM;
     int modulus = 0, multirdn = 0, verify = 0, noout = 0, text = 0;
     int nodes = 0, newhdr = 0, subject = 0, pubkey = 0, precert = 0;
-    long newkey = -1;
+    long newkey_len = -1;
```
kaduk, line 258: This kind of mechanical rename is a little easier to verify if it's done in a dedicated commit.  No need to change this one, since it's already done, but maybe in the future if it comes up again.

apps/req.c
```diff
@@ -644,25 +643,17 @@ int req_main(int argc, char **argv)
...
-            BIO_printf(bio_err, "Generating a %s private key\n", keyalgstr);
-        }
-
         EVP_PKEY_CTX_set_cb(genctx, genpkey_cb);
         EVP_PKEY_CTX_set_app_data(genctx, bio_err);
 
-        if (EVP_PKEY_keygen(genctx, &pkey) <= 0) {
-            BIO_puts(bio_err, "Error Generating Key\n");
+        if (!app_keygen(genctx, &pkey,
+                        pkey_type == EVP_PKEY_EC ? "EC": keyalgstr, newkey_len))
```
kaduk, line 655: I note that the purpose of 57358a83a401ef469353b7ebdae0cf3b870a4d5e was to get the bits right, not something specific about EC as the algorithm name.

apps/req.c
```diff
@@ -1675,7 +1652,6 @@ static int do_x509_req_init(X509_REQ *x, STACK_OF(OPENSSL_STRING) *opts)
         char *opt = sk_OPENSSL_STRING_value(opts, i);
         if (x509_req_ctrl_string(x, opt) <= 0) {
             BIO_printf(bio_err, "parameter error \"%s\"\n", opt);
-            ERR_print_errors(bio_err);
             return 0;
```
kaduk, line 1655: I'm not entirely sure I understand why this app gets 'return 0' directoy instead of the 'goto end' with `ERR_print_errors()` that we see in some of the others.

apps/dgst.c
```diff
@@ -234,10 +234,8 @@ int dgst_main(int argc, char **argv)
 
     in = BIO_new(BIO_s_file());
     bmd = BIO_new(BIO_f_md());
-    if ((in == NULL) || (bmd == NULL)) {
-        ERR_print_errors(bio_err);
+    if ((in == NULL) || (bmd == NULL))
```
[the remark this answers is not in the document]
DDvO: Done.

apps/dgst.c
```diff
@@ -262,7 +260,7 @@ int dgst_main(int argc, char **argv)
         goto end;
 
     if ((!(mac_name == NULL) + !(keyfile == NULL) + !(hmac_key == NULL)) > 1) {
-        BIO_printf(bio_err, "MAC and Signing key cannot both be specified\n");
+        BIO_printf(bio_err, "MAC and signing key cannot both be specified\n");
```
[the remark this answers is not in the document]
DDvO: > I would say it's best to leave that to a separate PR (but agree it is an improvement).

I agree with @kaduk.

apps/dgst.c
```diff
@@ -293,25 +291,23 @@ int dgst_main(int argc, char **argv)
     if (mac_name != NULL) {
         EVP_PKEY_CTX *mac_ctx = NULL;
         int r = 0;
+
         if (!init_gen_str(&mac_ctx, mac_name, impl, 0))
             goto mac_end;
         if (macopts != NULL) {
             char *macopt;
```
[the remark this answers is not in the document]
DDvO: Yes, this variable could not be moved into the scope, but after the next set of improvements it disappeared anyway :)

apps/dgst.c
```diff
@@ -341,16 +336,14 @@ int dgst_main(int argc, char **argv)
             r = EVP_DigestSignInit(mctx, &pctx, md, impl, sigkey);
         if (!r) {
```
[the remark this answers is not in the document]
DDvO: Done

apps/dgst.c
```diff
@@ -341,16 +336,14 @@ int dgst_main(int argc, char **argv)
...
             BIO_printf(bio_err, "Error setting context\n");
-            ERR_print_errors(bio_err);
             goto end;
         }
         if (sigopts != NULL) {
             char *sigopt;
```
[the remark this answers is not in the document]
DDvO: Done

apps/dgst.c
```diff
@@ -377,7 +368,6 @@ int dgst_main(int argc, char **argv)
         BIO *sigbio = BIO_new_file(sigfile, "rb");
```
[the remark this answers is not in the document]
DDvO: Done

apps/genpkey.c
DDvO: Done

apps/lib/apps.c
DDvO: Good thought! Done.
This lead to saving a couple of code lines and error conditions.

apps/req.c
DDvO: > Why try to proceed if this fails? The user specified an OID file and you're not opening it. Should use `bio_open_default`, right?

I don't know either. 
We might ask whoever did that.

apps/req.c
DDvO: Removed

apps/req.c
DDvO: Fixed

apps/req.c
DDvO: Yeah, funny style. Fixed.

apps/dgst.c
```diff
@@ -562,16 +550,13 @@ int do_fp(BIO *out, unsigned char *buf, BIO *bp, int sep, int binout,
             buf = sigbuf;
         }
         if (!EVP_DigestSignFinal(ctx, buf, &len)) {
-            BIO_printf(bio_err, "Error Signing Data\n");
-            ERR_print_errors(bio_err);
+            BIO_printf(bio_err, "Error signing data\n");
```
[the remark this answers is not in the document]
DDvO: Good point - I've differentiated the error messages.

apps/dsaparam.c
```diff
@@ -266,11 +264,8 @@ int dsaparam_main(int argc, char **argv)
             ERR_print_errors(bio_err);
             goto end;
         }
-        if (!EVP_PKEY_keygen(ctx, &pkey)) {
-            BIO_printf(bio_err, "unable to generate key\n");
-            ERR_print_errors(bio_err);
+        if (!app_keygen(ctx, &pkey, verbose ? "DSA" : NULL, numbits))
             goto end;
```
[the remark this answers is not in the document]
DDvO: The point of being able to remove all the inner `ERR_print_errors()` is that
there is now a global one just before exit (also in the bail out case).

apps/dhparam.c
```diff
@@ -251,11 +251,8 @@ int dhparam_main(int argc, char **argv)
                 ERR_print_errors(bio_err);
                 goto end;
             }
-            if (!EVP_PKEY_paramgen(ctx, &pkey)) {
-                BIO_printf(bio_err, "Error, DH generation failed\n");
-                ERR_print_errors(bio_err);
+            if (!app_paramgen(ctx, &pkey))
                 goto end;
```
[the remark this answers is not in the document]
DDvO: See above

apps/lib/apps.c
DDvO: Done both improvements.

apps/lib/apps.c
DDvO: See my remark on `ERR_print_errors()` above.

apps/lib/apps.c
DDvO: Yep, done.

apps/req.c
DDvO: You are right, will do in the future.

apps/req.c
DDvO: Meanwhile the `newkey_len` appears to be set correctly also in the EC case.

apps/req.c
DDvO: See above

apps/req.c
DDvO: Ah yes, and the special case for the key alg name is not (any more) needed either, so removed that.

DDvO:
Thanks @richsalz and @kroeckx for your review comments!
Finally I had some breath today for handling them.
Anything more to do?

kroeckx:
> >> I've revised this to issue the warning only if `RAND_status()` returns `false`.
> 
> > I'm not sure just calling it is enough. It could be that on a system that don't have enough entropy, the current drbg_status() will just hang itself. If that's the case, we should probably look into fixing that.
> 
> Good point. So I've marked this PR as WIP and will await this being sorted out.

So has this been sorted out?

DDvO:
> >> I've revised this to issue the warning only if `RAND_status()` returns `false`. > I'm not sure just calling it is enough. It could be that on a system that don't have enough entropy, the current drbg_status() will just hang itself. If that's the case, we should probably look into fixing that. Good point. So I've marked this PR as WIP and will await this being sorted out.
> So has this been sorted out?

I don't think so. But since over the last three weeks nobody spoke up with an idea/solution for this
and on the other hand this PR already contains a number of (more or less related) improvements
I've changed my mind and would like to merge this PR already as it is, 
and then post an issue about the remaining problem, referring to the current state.
Otherwise I fear that this PR would be blocked indefinitely.

apps/dsaparam.c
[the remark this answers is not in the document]
kaduk: I understood that that was the intent; I just failed to find the corresponding `ERR_print_errors()` call for this app.  Which is weird, because it seems like it's always been there.
Sorry for the noise.

DDvO:
@kaduk, I just realized that it was not @kroeckx but you who gave (along with @richsalz) reviewer's comments here - thanks.

Is this now acceptable, modulo the `RAND_status()` problem for which I'm going to open a follow-up issue?

DDvO:
@kaduk, I've just rebased this PR, solving two minor merge conflicts.
Do you have some this for a final look at this before I leave on Thursday?

apps/dgst.c
```diff
@@ -417,21 +403,20 @@ int dgst_main(int argc, char **argv)
...
-            int r;
             if (BIO_read_filename(in, argv[i]) <= 0) {
                 perror(argv[i]);
                 ret++;
                 continue;
             } else {
-                r = do_fp(out, buf, inp, separator, out_bin, sigkey, sigbuf,
-                          siglen, sig_name, md_name, argv[i]);
+                ret = do_fp(out, buf, inp, separator, out_bin, sigkey, sigbuf,
+                            siglen, sig_name, md_name, argv[i]);
```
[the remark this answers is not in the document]
DDvO: > This chunk means we only report the status from the last iteration of the loop (in terms of `do_fp()`); the old one would latch after any failure.

Yes. Yet this is an issue this PR is is not aiming to handle.
So I've added your remark as a comment there.

apps/dsaparam.c
```diff
@@ -262,10 +260,7 @@ int dsaparam_main(int argc, char **argv)
                        "Error, unable to initialise for key generation\n");
             goto end;
         }
-        if (!EVP_PKEY_keygen(ctx, &pkey)) {
-            BIO_printf(bio_err, "Error, unable to generate key\n");
-            goto end;
-        }
+        pkey = app_keygen(ctx, verbose ? "DSA" : NULL, numbits);
```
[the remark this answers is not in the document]
DDvO: The `verbose` option is used for printing application-level trace output;
any error diagnostics is not affected by it.

apps/genrsa.c
```diff
@@ -191,13 +191,7 @@ int genrsa_main(int argc, char **argv)
         BIO_printf(bio_err, "Error setting number of primes\n");
         goto end;
     }
-    if (verbose)
-        BIO_printf(bio_err, "Generating RSA private key, %d bit long modulus (%d primes)\n",
```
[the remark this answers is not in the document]
DDvO: Yes, we lose it on error output, but it's printed with `verbose`.

apps/req.c
```diff
@@ -826,10 +811,8 @@ int req_main(int argc, char **argv)
                 goto end;
             }
             i = do_X509_REQ_sign(req, pkey, digest, sigopts);
-            if (!i) {
-                ERR_print_errors(bio_err);
+            if (!i)
```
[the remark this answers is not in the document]
DDvO: > And if we're changing to explicit tests against zero, this is also one.

The point of the two deletions quoted above is not to change the program flow,
but just to remove (needless) intermediate calls of `ERR_print_errors(bio_err)`,
which is printed on exit anyway (like for other apps).

apps/req.c
```diff
@@ -868,14 +851,13 @@ int req_main(int argc, char **argv)
 
         i = do_X509_REQ_verify(req, tpubkey, vfyopts);
 
-        if (i < 0) {
+        if (i < 0)
             goto end;
-        } else if (i == 0) {
+        if (i == 0)
             BIO_printf(bio_err, "verify failure\n");
-            ERR_print_errors(bio_err);
```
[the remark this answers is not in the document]
DDvO: As mentioned before, in this PR all error queue printing is moved to app exit (like for other apps).

apps/req.c
```diff
@@ -1354,7 +1333,6 @@ static int add_attribute_object(X509_REQ *req, char *text, const char *def,
     if (!X509_REQ_add1_attr_by_NID(req, nid, chtype,
                                    (unsigned char *)buf, -1)) {
         BIO_printf(bio_err, "Error adding attribute\n");
-        ERR_print_errors(bio_err);
```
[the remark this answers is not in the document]
DDvO: I agree.

DDvO:
Thanks @kaduk for you comments!

> The for-loop in `dgst_main` is the only thing that seems particularly problematic in this batch of comments.

Yes, but since that issue is not the topic of this PR I've converted it to a FIXME comment in the code.

> (Note that I'm not on the OTC, so my slow review wasn't keeping this from getting merged anyway, but sorry that I didn't get to it in the requested timeframe regardless.)

After my vacation I forgot to handle your comments, but finally I've handled them.
So we still need an OTC review here, but if you approve this would likely speed up things.

DDvO:
Now there is one Travis failure (on test_cms), which seems unrelated.

paulidale:
I am expecting `RAND_status()` to block waiting for entropy.  At least, it will once #12714 is fixed.

DDvO:
> I am expecting `RAND_status()` to block waiting for entropy. At least, it will once #12714 is fixed.

Thanks for this warning, which is in the same direction as https://github.com/openssl/openssl/pull/12095#issuecomment-641168726.

Which non-blocking function can be used alternatively to check if currently available entropy should be sufficient?

paulidale:
We don't have a function to check if entropy can be obtained without blocking.  I'm not sure such is generally possible.

DDvO:
> We don't have a function to check if entropy can be obtained without blocking. I'm not sure such is generally possible.

That's a pity.
Then I tend to withdraw the essence of this PR, keeping just the general code cleanup I've done along with it.
Since this is not urgent I'm just marking this as WIP for now.

DDvO:
Rebased to fix merge conflicts.

DDvO:
Is it meanwhile clear whether adding a warning on potential delays due to missing entropy makes sense after all?
If it does not make sense (anymore), I'll remove the proposed warnings from this PR and keep just the cleanup of error handling.

apps/dgst.c
```diff
@@ -571,8 +554,7 @@ int do_fp(BIO *out, unsigned char *buf, BIO *bp, int sep, int binout, int xoflen
 
         BIO_get_md_ctx(bp, &ctx);
         if (!EVP_DigestSignFinal(ctx, NULL, &tmplen)) {
-            BIO_printf(bio_err, "Error Signing Data\n");
-            ERR_print_errors(bio_err);
+            BIO_printf(bio_err, "Error getting length of signed data\n");
```
paulidale, line 557: Is this the only failure mode?

paulidale (commented):
One of the changed error messages seems awry but otherwise good.

paulidale:
I think we should have the warning.

apps/dgst.c
DDvO: AKAICS yes because `EVP_DigestSignFinal(ctx, NULL, &tmplen)` just asks for the size.
The `EVP_DigestSignFinal(ctx, buf, &len)` below returns the signature, 
and I believe the actual signature calculation is done along with that.
Yet `EVP_DigestSignInit.pod` is not entirely clear to me on this.

DDvO: Just verified my suspicion on `dsa_sign()` and `ecdsa_sign()`, so: yes.

DDvO: I'm proposing a respective doc improvement in #15354.

paulidale: Now add a 3rd party provider into the mix.  Is this still the only failure mode?

The message needs to less specific.

t8m:
IMO the text is OK. Even with a third party provider the failure of EVP_DigestSignFinal with NULL signature parameter cannot mean anything else than error getting maximum signature length. This is the only sensible semantics. Of course the reason for the error might be various but the error is just this. The call is not supposed to do anything else.

20. 5. 2021 2:15, 2:15, Pauli ***@***.***> napsal/a:
***@***.*** commented on this pull request.

t8m:
Although the obtained length is a maximum length. It doesn't necessarily have to be the exact length so perhaps some fine tunning of the message should be done.

20. 5. 2021 2:15, 2:15, Pauli ***@***.***> napsal/a:
***@***.*** commented on this pull request.

apps/dgst.c
DDvO: A 3rd party provider should behave according to the expectations that hopefully are sufficiently documented. My above reasoning was just to convince myself of what those expectations should be, and doing so I even updated the documentation what the users of `EVP_DigestSignFinal()` can expect.

The response meanwhile given below by @t8m confirms that for the caller of `EVP_DigestSignFinal(ctx, NULL, &tmplen)` the point is just that the `dgst` app obtains the (maximum) length of the signed data, and the error message should refer to that. 
I've made the error message more precise in this way: `"Error getting maximum length of signed data\n"`
And IMHO this is all what can be said reasonably in this error message.

Note that also in the call further down `EVP_DigestSignFinal(ctx, buf, &len)`
everyone is happy with `"Error signing data\n"`, although theoretically a provider could in addition do any other/further stuff like mining bitcoins, which also could result in a failure of the function that would be unrelated to signing the data, but of course we do not take such things into account in the error message.

I don't think it's worth spending much effort in discussing such error messages at length and in detail.
My original aim when replacing `"Error signing data\n"` was just to correct that statement because it was plainly wrong/misleading, and this correction is certainly achieved now with the new formulation.

DDvO: @paulidale, is this okay now?

t8m (approved):

apps/dgst.c
```diff
@@ -435,21 +421,20 @@ int dgst_main(int argc, char **argv)
...
         for (i = 0; i < argc; i++) {
-            int r;
             if (BIO_read_filename(in, argv[i]) <= 0) {
                 perror(argv[i]);
                 ret++;
                 continue;
             } else {
-                r = do_fp(out, buf, inp, separator, out_bin, xoflen,
-                          sigkey, sigbuf, siglen, sig_name, md_name, argv[i]);
+                ret = do_fp(out, buf, inp, separator, out_bin, xoflen,
```
t8m, line 429: Isn't this changing semantics? This will happily overwrite the ret if one do_fp() call passes where previously there might have been failures already.

t8m (changes requested):

t8m:
Also the external tests failure seems relevant.

apps/dgst.c
DDvO: Oh, good spotting!
Fixed.

DDvO:
> Also the external tests failure seems relevant.

Yes, unfortunately. This was hard to debug and get right.
It turns out that the `gost-engine` tests are rather sensitive to changes of the console output.
It took me a while to find out what it expects.
In the end I found forced to add a workaround on the GOST algorithm name printed by `dgst`
in order to avoid mismatched like the following:
```
 # Failed test 'GOST MAC - parameters 2012'
# at /home/runner/work/openssl/openssl/gost-engine/test/02-mac.t line 39.
# +---------------------------------+----+---------------------------------+
# | GOT                             | OP | CHECK                           |
# +---------------------------------+----+---------------------------------+
# | Generating a gost-mac-12 key\n  | eq | GOST-MAC-12-gost-mac-12(testdat |
# | GOST-MAC-12-gost-mac-12(testdat |    | a.dat)= be4453ec\n              |
# | a.dat)= be4453ec\n              |    |                                 |
# +---------------------------------+----+---------------------------------+
```

t8m:
This workaround is very ugly. I would consider other options. What about printing these messages to stderr instead of stdout? IMO this would be more appropriate for such kind of auxiliary messages.

Or, maybe not printing them at all for this hmac keygen case.

But I suspect that there will be other apps or scripts broken by this an it will inevitably lead to stuff like requests for new -quiet options.

21. 5. 2021 19:11, 19:11, David von Oheimb ***@***.***> napsal/a:

DDvO:
> This workaround is very ugly.

In between it was even more ugly :stuck_out_tongue_winking_eye:

> I would consider other options. What about printing these messages to stderr instead of stdout?

Originally it was `stderr`, but had essentially the same issue. Apparently the test harness merges their output.

> IMO this would be more appropriate for such kind of auxiliary messages.

I would have thought the other way round, but all right, followed your preference.

> Or, maybe not printing them at all for this hmac keygen case.

Done this now - which actually is the original behavior at this point.

apps/lib/apps.c
```diff
@@ -3258,3 +3259,35 @@ void app_params_free(OSSL_PARAM *params)
...
+            BIO_printf(bio_err, "Generating a %s key, with %d bits\n",
+                       alg, bits);
+        else
+            BIO_printf(bio_err, "Generating a %s key\n", alg);
+    }
+    if (!RAND_status())
+        BIO_printf(bio_err, "Warning: generating random key material may take a long time\n"
+                   "if the system has a poor entropy source\n");
+    if (EVP_PKEY_keygen(ctx, &res) <= 0)
+        app_bail_out("%s: Error generating %s key\n", opt_getprog(), alg);
```
t8m, line 3278: This does not handle well the case where alg == NULL. Perhaps it would be better to use an additional argument instead of alg == NULL to indicate the verbosity? IMO the verbose message `Generating a .... key ...` should not be added for existing cases where there was no verbose message printed now.

t8m (changes requested):

apps/lib/apps.c
DDvO: Good points. Done.
Adding a `verbose` parameter makes things cleaner. 
Made sure that also `app_paramgen()` handles the case `alg == NULL` well.

t8m (approved):

PR #12096, opened 2020-06-09
[test] ectest: check custom generators

bbbrumley:
I was doing some custom `EC_METHOD` work that should've obviously failed because it was oblivious to `EC_GROUP_set_generator`.

But it passed all tests.

So here's a PR for that.

A handful of the NIST curves have similar tests in `ectest.c` that I'll deprecate as part of a larger cleanup of this test harness in the future.

### DIFF
--- a/test/ectest.c
+++ b/test/ectest.c
@@ -2337,6 +2337,89 @@ static int ec_point_hex2point_test(int id)
     return ret;
 }
 
+/*
+ * check the EC_METHOD respects the supplied EC_GROUP_set_generator G
+ */
+static int custom_generator_test(int id)
+{
+    int ret = 0, nid, bsize;
+    EC_GROUP *group = NULL;
+    EC_POINT *G2 = NULL, *Q1 = NULL, *Q2 = NULL;
+    BN_CTX *ctx = NULL;
+    BIGNUM *k = NULL;
+    unsigned char *b1 = NULL, *b2 = NULL;
+
+    /* Do some setup */
+    nid = curves[id].nid;
+    TEST_note("Curve %s", OBJ_nid2sn(nid));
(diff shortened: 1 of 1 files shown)


romen:
I missed this before!

This is supposed to be also in `1.1.1` right? Can you check if it cherry-picks cleanly?

bbbrumley:
This is holding up #12201, which still needs effort after this gets merged. Can I get a second review please?

beldmit:
Merged to both branches. Thanks!

test/ectest.c
```diff
@@ -2337,6 +2337,89 @@ static int ec_point_hex2point_test(int id)
...
+
+    BN_CTX_start(ctx);
+
+    if (!TEST_ptr(group = EC_GROUP_new_by_curve_name(nid)))
+        goto err;
+
+    /* expected byte length of encoded points */
+    bsize = (EC_GROUP_get_field_type(group) == NID_X9_62_prime_field) ?
+             BN_num_bytes(EC_GROUP_get0_field(group)) :
+             (EC_GROUP_get_degree(group) + 7) / 8;
```
romen, line 2366: @bbbrumley is there a reason why we need to prefer this over simply:

~~~C
bsize = (EC_GROUP_get_degree(group) + 7) / 8;
~~~

(in `1.1.1` there is no getter for `group->field` and `EC_GROUP_get_field_type()` needs to be locally defined as a wrapper around `EC_METHOD_get_field_type()`, but on deeper inspection I don't see why we can't just use the result of `get_degree()` always and convert bits to bytes as the manpage says:

> EC_GROUP_get_degree() gets the degree of the field. For Fp fields this will be the number of bits in p. For F2^m fields this will be the value m.

)

romen: See #12432 and #12433

PR #12097, opened 2020-06-09
Amendment to #12080

levitte:
For @romen.

## TEST: Add TODO segments in test/recipes/15-test_genec.t

There currently do not support 'ec_param_enc:explicit' with provider
side key generation.  Reflect that by encoding the expected failure
with a Test::More TODO section for those particular tests.

Because the tests in this recipe are data driven, we implement this
mechanism with two functions, one for stuff that's supported and one
for stuff that isn't.

### DIFF
--- a/test/recipes/15-test_genec.t
+++ b/test/recipes/15-test_genec.t
@@ -115,15 +115,16 @@ my @curve_aliases = qw(
 );
 
 my @curve_list = ();
-push(@curve_list, @prime_curves);
-push(@curve_list, @binary_curves)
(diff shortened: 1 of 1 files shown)


romen:
Thanks @levitte ! (this is going to be useful in general for some more TDD I am doing at the moment)

I am closing this as I cherry-picked your commit on #12080 and added a small cosmetic change moving the subroutines towards the beginning of the file to improve readability (at least to the eyes of someone that is not that fluent in Perl and spent a lot of time to understand the reference to the subroutines in the associative array!).

Issue #12068, opened 2020-06-06
Windows 10 - Build Problems - ActivePerl missing Win32/Console.pm


Tockra:
So Guys,

I'm very mad. Your build instructions are very bad.
You describe the dependencies here: https://github.com/openssl/openssl/blob/master/NOTES.WIN

I installed ActiveState, nasm and VS2019. 
I tried to run `perl Configure VC-Win64A` from a "Visual Studio 2019 Developer Command Prompt v16.6.1" what isn't explained somewhere. But I think it is necessary to use this command prompt...

Then I get the error: 
```bash
C:\Users\TiTan\openssl-master>perl Configure VC-WIN64A
Configuring OpenSSL version 1.1.1g (0x1010107fL) for VC-WIN64A
Using os-specific seed configuration
Can't locate Win32/Console.pm in @INC (you may need to install the Win32::Console module) (@INC contains: C:/Users/TiTan/openssl-master/util/perl C:/Perl64/site/lib C:/Perl64/lib) at C:/Perl64/site/lib/ActivePerl/Config.pm line 400.
Compilation failed in require at C:/Perl64/site/lib/ExtUtils/MakeMaker.pm line 9.
BEGIN failed--compilation aborted at C:/Perl64/site/lib/ExtUtils/MakeMaker.pm line 9.
Compilation failed in require at C:/Perl64/site/lib/IPC/Cmd.pm line 235.
```

So no problem. I tried `cpan -i Win32::Console`. But then: 
```bash
C:\Users\TiTan\openssl-master>cpan -i Win32::Console
Loading internal logger. Log::Log4perl recommended for better logging
Reading 'C:\Perl64\cpan\Metadata'
  Database was generated on Fri, 05 Jun 2020 03:17:03 GMT
Running install for module 'Win32::Console'
CPAN: Digest::SHA loaded ok (v6.01)
CPAN: Compress::Zlib loaded ok (v2.092)
Checksum for C:\Perl64\cpan\sources\authors\id\J\JD\JDB\Win32-Console-0.10.tar.gz ok
CPAN: Archive::Tar loaded ok (v2.30)
Win32-Console-0.10/
Win32-Console-0.10/Changes
Win32-Console-0.10/Console.pm
Win32-Console-0.10/Console.xs
Win32-Console-0.10/docs/
Win32-Console-0.10/docs/index.html
Win32-Console-0.10/docs/intro.html
Win32-Console-0.10/docs/license.html
[...]

PR #12098, opened 2020-06-09
Update and extend NOTES.WIN, adding 'Quick start' subsection

DDvO:
This PR has been motivated by a hick-up with recent AcitvePerl-5.28, see #12068. It
* makes ActivePerl second choice, after Strawberry Perl,
* adds a little more info on the Perl modules needed and where to get them,
* tweaks a couple of sentences, and
* adds a `Quick start` section with steps taken from @mattcaswell's answer from https://stackoverflow.com/questions/45494630/how-to-build-openssl-on-windows-with-visual-studio-2017/45495686#45495686.

### DIFF
--- a/NOTES.WIN
+++ b/NOTES.WIN
@@ -35,14 +35,11 @@
    via https://platform.activestate.com/ActiveState and then download it.
 
  - Microsoft Visual C compiler.
-   Since these are proprietry and ever-changing and we cannot test them all,
-   there is unavoidable uncertainty about which versions are supported.
-   Latest version along with couple of previous are certainly supported.
+   Since these are proprietary and ever-changing we cannot test them all.
+   Older versions may not work. Use a recent version wherever possible.
 
  - Netwide Assembler (NASM), available from https://www.nasm.us
    Note that NASM is the only supported assembler.
-   Even though Microsoft provided assembler is NOT supported, contemporary
-   64-bit version is exercised through CI of the VC-WIN64A-masm target.
 
  Quick start
  -----------
@@ -150,7 +147,7 @@
 
  * It is also possible to build mingw[64] on Linux or Cygwin.
 
-   In this case configure with corresponding --cross-compile-prefix= option.
+   In this case configure with the corresponding --cross-compile-prefix= option.
    For example
 
      ./Configure mingw --cross-compile-prefix=i686-w64-mingw32- ...
@@ -187,8 +184,8 @@
 	}
 
  If you link with OpenSSL .DLLs, then you're expected to include into
- your application code small "shim" snippet, which provides glue between
- the OpenSSL BIO layer and your compiler run-time.
+ your application code a small "shim" snippet, which provides
+ the glue between the OpenSSL BIO layer and your compiler run-time.
  See also the OPENSSL_Applink manual page.
 
 
--- a/NOTES.WIN
+++ b/NOTES.WIN
@@ -128,20 +128,23 @@
 
    - MSYS2 shell, from https://www.msys2.org/
 
+   - Perl, at least version 5.10.0, which usually comes pre-installed with MSYS2
+
+   - make, installed using "pacman -S make" into the MSYS2 environment
+
    - MinGW[64] compiler: mingw-w64-i686-gcc and/or mingw-w64-x86_64-gcc.
      These compilers must be on your MSYS2 $PATH.
      A common error is to not have these on your $PATH.
      The MSYS2 version of gcc will not work correctly here.
-
-   - Perl, at least version 5.10.0, installed into the MSYS2 environment
-
-   - make, installed into the MSYS2 environment
 
    In the MSYS2 shell do the configuration depending on the target architecture:
 
      ./Configure mingw ...
    or
      ./Configure mingw64 ...
+   or
+     ./config ...
+   for the default architecture.
 
    Apart from that, follow the Unix / Linux instructions in INSTALL.md.
 
@@ -154,8 +157,8 @@
    or
      ./Configure mingw64 --cross-compile-prefix=x86_64-w64-mingw32- ...
 
-   This naturally requires that you've installed corresponding add-on packages.
-
+   This requires that you've installed the necessary add-on packages for
+   mingw[64] cross compilation.
 
  Linking your application
  ========================
@@ -185,7 +188,7 @@
 
  If you link with OpenSSL .DLLs, then you're expected to include into
  your application code small "shim" snippet, which provides glue between
- OpenSSL BIO layer and your compiler run-time.
+ the OpenSSL BIO layer and your compiler run-time.
  See also the OPENSSL_Applink manual page.
 
 
--- a/NOTES.WIN
+++ b/NOTES.WIN
@@ -35,24 +35,51 @@
  In addition to the requirements and instructions listed in INSTALL,
  these are required as well:
 
- - Perl. We recommend ActiveState Perl, available from
-   https://www.activestate.com/ActivePerl. Another viable alternative
-   appears to be Strawberry Perl, http://strawberryperl.com.
-   You also need the perl module Text::Template, available on CPAN.
-   Please read NOTES.PERL for more information.
+ - Perl.
+   We recommend Strawberry Perl, available from http://strawberryperl.com.
+   Please read NOTES.PERL for more information, including the use of CPAN.
+   An alternative is ActiveState Perl, https://www.activestate.com/ActivePerl
+   for which you may need to explicitly select the Perl module Win32/Console.pm
+   available via https://platform.activestate.com/ActiveState.
 
- - Microsoft Visual C compiler. Since we can't test them all, there is
-   unavoidable uncertainty about which versions are supported. Latest
-   version along with couple of previous are certainly supported. On
-   the other hand oldest one is known not to work. Everything between
-   falls into best-effort category.
+ - Microsoft Visual C compiler.
+   Since these are proprietry and ever-changing and we cannot test them all,
+   there is unavoidable uncertainty about which versions are supported.
+   Latest version along with couple of previous are certainly supported.
+   On the other hand oldest one is known not to work.
+   Everything between falls into best-effort category.
 
- - Netwide Assembler, a.k.a. NASM, available from https://www.nasm.us,
-   is required. Note that NASM is the only supported assembler. Even
-   though Microsoft provided assembler is NOT supported, contemporary
-   64-bit version is exercised through continuous integration of
-   VC-WIN64A-masm target.
+ - Netwide Assembler, a.k.a. NASM, available from https://www.nasm.us.
+   Note that NASM is the only supported assembler.
+   Even though Microsoft provided assembler is NOT supported, contemporary
+   64-bit version is exercised through CI of the VC-WIN64A-masm target.
 
+ Quick start
+ -----------
+
+ 1. Install Perl
+
+ 2. Install NASM
+
+ 3. Make sure both Perl and NASM are on your %PATH%
+
+ 4. Run a Visual Studio Developer Command Prompt with administrative privileges
+    (make sure you use the 32-bit one if you are building 32-bit OpenSSL,
+    or the 64-bit one if you are building 64-bit OpenSSL).
+    This sets up the environment variables needed to find make.exe, cl.exe, etc.
+
+ 5. From the root of the OpenSSL source directory enter
+    perl Configure VC-WIN32    if you want 32-bit OpenSSL or
+    perl Configure VC-WIN64A   if you want 64-bit OpenSSL
+
+ 6. nmake
+
+ 7. nmake test
+
+ 8. nmake install
+
+For the full installation instructions, of if anything goes wrong at any stage,
+check the INSTALL.md file.
 
  Installation directories
  ------------------------
--- a/NOTES.WIN
+++ b/NOTES.WIN
@@ -2,29 +2,12 @@
  NOTES FOR THE WINDOWS PLATFORMS
  ===============================
 
- Windows targets can be classified as "native", ones that use Windows API
- directly, and "hosted" which rely on POSIX-compatible layer. "Native"
- targets are VC-* (where "VC" stems from abbreviating Microsoft Visual C
- compiler) and mingw[64]. "Hosted" platforms are Cygwin and MSYS[2]. Even
- though the latter is not directly supported by OpenSSL Team, it's #1
- popular choice for building MinGW targets. In the nutshell MinGW builds
- are always cross-compiled. On Linux and Cygwin they look exactly as such
- and require --cross-compile-prefix option. While on MSYS[2] it's solved
- rather by placing gcc that produces "MinGW binary" code 1st on $PATH.
- This is customarily source of confusion. "Hosted" applications "live" in
- emulated file system name space with POSIX-y root, mount points, /dev
- and even /proc. Confusion is intensified by the fact that MSYS2 shell
- (or rather emulated execve(2) call) examines the binary it's about to
- start, and if it's found *not* to be linked with MSYS2 POSIX-y thing,
- command line arguments that look like file names get translated from
- emulated name space to "native". For example '/c/some/where' becomes
- 'c:\some\where', '/dev/null' - 'nul'. This creates an illusion that
- there is no difference between MSYS2 shell and "MinGW binary", but
- there is. Just keep in mind that "MinGW binary" "experiences" Windows
- system in exactly same way as one produced by VC, and in its essence
- is indistinguishable from the latter. (Which by the way is why
- it's referred to in quotes here, as "MinGW binary", it's just as
- "native" as it can get.)
+ Windows targets can be classified as "native", ones that use the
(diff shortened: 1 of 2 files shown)


NOTES.WIN
```diff
@@ -35,24 +35,51 @@
...
  these are required as well:
 
- - Perl. We recommend ActiveState Perl, available from
-   https://www.activestate.com/ActivePerl. Another viable alternative
-   appears to be Strawberry Perl, http://strawberryperl.com.
-   You also need the perl module Text::Template, available on CPAN.
+ - Perl.
+   We recommend Strawberry Perl, available from http://strawberryperl.com.
+   An alternative is ActiveState Perl, https://www.activestate.com/ActivePerl. 
+   You also need the perl modules Text::Template and Win32::Console,
```
levitte, line 41: "You *may* also...", and as far as we know, that's only for newer versions of ActivePerl

NOTES.WIN
```diff
@@ -35,24 +35,51 @@
...
+    perl Configure VC-WIN64A   if you want 64-bit OpenSSL
+
+ 6. nmake
+
+ 7. nmake test
+
+ 8. nmake install
+
+If anything goes wrong at any stage,
+check the INSTALL file and the NOTES.WIN file.
```
levitte, line 82: INSTALL.md, and this is already NOTES.WIN, that self referential looks odd

NOTES.WIN
DDvO: done

NOTES.WIN
DDvO: Ah, sure, changed to `check the INSTALL.md file.`

DDvO:
BTW, any chance to rename the `NOTES.WIN` file to, e.g., `NOTES-WIN.txt`?
Since the Windows file explorer does not know how to open a `WIN File` :wink:

NOTES.WIN
```diff
@@ -35,24 +35,50 @@
...
  these are required as well:
 
- - Perl. We recommend ActiveState Perl, available from
-   https://www.activestate.com/ActivePerl. Another viable alternative
-   appears to be Strawberry Perl, http://strawberryperl.com.
-   You also need the perl module Text::Template, available on CPAN.
+ - Perl.
+   We recommend Strawberry Perl, available from http://strawberryperl.com.
+   An alternative is ActiveState Perl, https://www.activestate.com/ActivePerl.
+   You may also need the perl modules Text::Template and Win32::Console,
```
levitte, line 41: Actually, to make it clear that this is for ActivePerl, so may I suggest joining the sentences, i.e. like this?

```
 - Perl.
   We recommend Strawberry Perl, available from http://strawberryperl.com.
   An alternative is ActiveState Perl, https://www.activestate.com/ActivePerl,
   for which you may need the perl modules Text::Template and Win32::Console,
   available on CPAN or https://platform.activestate.com/ActiveState.
```

DDvO: The reference to `Text::Template` was already there, and I suspect that at least that module is needed (but already included) with Strawberry Perl. 
CPAN is likely not helpful for ActivePerl.

levitte: I'm frankly not sure why Text::Template is needed, as we include it in our tarballs...  and Win32::Console is currently only needed for ActivePerl.  The way this is currently phrased, it sounds like both are needed for Strawberry as well, and I see no reason why we should send our users on that goose chase unnecessarily.

DDvO: I've just changed this to:
```
An alternative is ActiveState Perl, https://www.activestate.com/ActivePerl
for which you may need to explicitly select the Perl module Win32/Console.pm
available via https://platform.activestate.com/ActiveState.
```

DDvO:
I've just found that `NOTES.PERL` also needed an update regarding ActivePerl; done.

NOTES.WIN
levitte: Yes!  Really good.  Thanks for your patience

DDvO: And thank you for the swift feedback & approval.

mattcaswell:
I wonder if we could re-write the opening section in NOTES.WIN. The "wall of text" is off putting IMO.

NOTES.WIN
```diff
@@ -35,24 +35,50 @@
...
+    perl Configure VC-WIN32    if you want 32-bit OpenSSL or
+    perl Configure VC-WIN64A   if you want 64-bit OpenSSL
+
+ 6. nmake
+
+ 7. nmake test
+
+ 8. nmake install
+
+If anything goes wrong at any stage, check the INSTALL.md file.
```
mattcaswell, line 81: Perhaps:

"For the full installation instructions, or if anything goes wrong at any stage check the INSTALL.md file"

levitte:
> I wonder if we could re-write the opening section in NOTES.WIN. The "wall of text" is off putting IMO.

I would take it away.

NOTES.WIN
DDvO: Done.

DDvO:
> I wonder if we could re-write the opening section in NOTES.WIN. The "wall of text" is off putting IMO.

I've tentatively moved it down to where it belongs: the MinGW section.

NOTES.WIN
```diff
@@ -35,24 +18,51 @@
...
 
- - Netwide Assembler, a.k.a. NASM, available from https://www.nasm.us,
-   is required. Note that NASM is the only supported assembler. Even
-   though Microsoft provided assembler is NOT supported, contemporary
-   64-bit version is exercised through continuous integration of
-   VC-WIN64A-masm target.
+ 4. Run a Visual Studio Developer Command Prompt with administrative privileges
+    (make sure you use the 32-bit one if you are building 32-bit OpenSSL,
+    or the 64-bit one if you are building 64-bit OpenSSL).
+    This sets up the environment variables needed to find make.exe, cl.exe, etc.
```
levitte, line 52: nmake.exe

richsalz:
Renaming the file, as pointed out in https://github.com/openssl/openssl/pull/12098#issuecomment-641255497, seems like a real aide to newcomers.

NOTES.WIN
DDvO: Done

NOTES.WIN
```diff
@@ -2,29 +2,12 @@
...
- system in exactly same way as one produced by VC, and in its essence
- is indistinguishable from the latter. (Which by the way is why
- it's referred to in quotes here, as "MinGW binary", it's just as
- "native" as it can get.)
+ Windows targets can be classified as "native", ones that use the
+ Windows API directly, and "hosted" which rely on POSIX-compatible layer.
+ "Native" targets are VC-* (where "VC" stems from Microsoft Visual C compiler)
+ and mingw[64]. "Hosted" platforms are Cygwin and MSYS[2].
+ Even though the latter is not directly supported by OpenSSL Team,
+ it's #1 popular choice for building MinGW targets.
```
mattcaswell, line 10: This text is actually quite confusing - although I understand what the original author was trying to say. Part of the confusion comes from the fact that you can build "native" targets (via mingw), using a "hosted" build environment (msys2). I think we can simplify some of this. How about:

> OpenSSL can be built in a number of different ways on the Windows platform. There are two main types: "native" and "hosted". A native build of OpenSSL uses the Windows APIs directly. A hosted build of OpenSSL relies on an external compatibility layer.
> 
> To build a native OpenSSL you have two build environment options. You can either use:
> 
> 1) a Visual C++ based environment
> or
> 2) a MinGW based environment
> 
> To build a "Hosted" OpenSSL you can use Cygwin.

levitte:
What about the other notes?
Also, they are largely markdown formatted, is .txt an appropriate extension? In that case, should we look at the extension for README and INSTALL one more time?

NOTES.WIN
```diff
@@ -93,9 +103,28 @@
...
+ start, and if it's found *not* to be linked with MSYS2 POSIX-y thing,
+ command line arguments that look like file names get translated from
+ emulated name space to "native". For example '/c/some/where' becomes
+ 'c:\some\where', '/dev/null' - 'nul'. This creates an illusion that
+ there is no difference between MSYS2 shell and "MinGW binary", but
+ there is. Just keep in mind that "MinGW binary" "experiences" Windows
+ system in exactly same way as one produced by VC, and in its essence
+ is indistinguishable from the latter. (Which by the way is why
+ it's referred to in quotes here, as "MinGW binary", it's just as
+ "native" as it can get.)
```
mattcaswell, line 126: I think most of this is unnecessary. How about:

> MinGW offers an alternative way to build a native OpenSSL. 
> 
>  Requirement details
>  --------------------------
> 
> - MSYS2 shell. You must install the MSYS2 environment and start an MSYS2 shell. MSYS2 provides a Unix like command prompt, and a UNIX compatibility layer for applications. However in this context it is only used for building OpenSSL. The resulting OpenSSL that is created does not rely on MSYS2 to run and is fully native.
> 
> - MinGW or MinGW64 compiler. These compilers must be on your MSYS2 $PATH. A common error is to not have these on your $PATH and instead have the MSYS2 version of gcc. This will not work correctly.
> 
> - Perl, installed into the MSYS2 environment
>
> - make installed into the MSYS2 environment

NOTES.WIN
```diff
@@ -93,9 +103,28 @@
...
+
```
mattcaswell, line 127: I'd be tempted to move  the section below about building for mingw from Linux/Cygwin into a separate "Cross compiling" section.

DDvO:
> What about the other notes?
> Also, they are largely markdown formatted, is .txt an appropriate extension? In that case, should we look at the extension for README and INSTALL one more time?

I'd take the `.md` extension for all MarkDown formatted files,
and `.txt` for everything else, in particular the NOTES.* files.
Or even better, convert the latter ones to MarkDown.

richsalz:
Strong agreement with @ddvo's comment https://github.com/openssl/openssl/pull/12098#issuecomment-641343640  Convert to md and use the .md extension else use the .txt extension and fix the names of those files.

levitte:
Does Windows understand the .md extension out of the box?
Anyway, the NOTES files are already largely md (you could call it md-ish)

DDvO:
> Does Windows understand the .md extension out of the box?

Sigh, no. So I suggest using the `.txt` extension for the `NOTES.WIN` file (only).

> Anyway, the NOTES files are already largely md (you could call it md-ish)

Yes, so it's not a big effort to check if they look fine after renaming them to `.md` 
and potentially adapt their contents as far as needed. I can do that when agreed.

DDvO:
> Strong agreement with @DDvO's comment [#12098 (comment)](https://github.com/openssl/openssl/pull/12098#issuecomment-641343640) Convert to md and use the .md extension else use the .txt extension and fix the names of those files.

Ah, so I think I can take this as agreement.
Will do by tomorrow (I have to leave now).

DDvO:
> My i suggest that the general modification be made as _another_ PR?

No problem - I can even do this in a parallel PR since that one won't touch the contents of `NOTES.WIN`.

levitte:
> No problem - I can even do this in a parallel PR since that one won't touch the contents of NOTES.WIN.

Thank you.  I was afraid that this PR would grow indefinitely as new ideas are tossed at you, and I'm sure you've noticed, big PRs have a disproportionally longer processing time (for understandable reasons, human fallacy 'n all that)

NOTES.WIN
DDvO: > This text is actually quite confusing

I agree. Thanks @mattcaswell for your clarifications,
which were of helped also to me to understand the various possibilities better.
I've taken over your suggestion for improved structuring and extended it.
This was more work than expected, but I think it was worth while.

NOTES.WIN
DDvO: > I think most of this is unnecessary.

Yes, and also according to @levitte's comment I remove most of the MSYS details
and combined your suggesting with part of the remaining MinGW related text.

NOTES.WIN
DDvO: > I'd be tempted to move the section below about building for mingw from Linux/Cygwin into a separate "Cross compiling" section.

Please check the new MinGW section, which I revised entirely. Ok now?

NOTES.WIN
```diff
@@ -83,41 +102,60 @@
...
+   - Perl, at least version 5.10.0, installed into the MSYS2 environment
+
+   - make, installed into the MSYS2 environment
+
+   In the MSYS2 shell do the configuration depending on the target architecture:
 
+     ./Configure mingw ...
    or
+     ./Configure mingw64 ...
```
mattcaswell, line 144: You should just be able to do `./config` here. It *should* auto-detect the environment correctly (but I don't have easy access to a Windows environment at the moment to confirm its still working...my Windows environment died).

NOTES.WIN
```diff
@@ -83,41 +102,60 @@
...
+   In this case configure with corresponding --cross-compile-prefix= option.
+   For example
+
+     ./Configure mingw --cross-compile-prefix=i686-w64-mingw32- ...
+   or
      ./Configure mingw64 --cross-compile-prefix=x86_64-w64-mingw32- ...
 
-   This naturally implies that you've installed corresponding add-on
-   packages.
+   This naturally requires that you've installed corresponding add-on packages.
```
mattcaswell, line 157: This requires that you've installed the necessary add-on packages for mingw[64] cross compilation.

NOTES.WIN
```diff
@@ -147,27 +185,28 @@
 
  If you link with OpenSSL .DLLs, then you're expected to include into
  your application code small "shim" snippet, which provides glue between
- OpenSSL BIO layer and your compiler run-time. See the OPENSSL_Applink
- manual page for further details.
+ OpenSSL BIO layer and your compiler run-time.
```
mattcaswell, line 188: ...include into your application *a* small "shim" snippet, which provides glue between *the* OpenSSL BIO layer and your compiler run-time.

NOTES.WIN
levitte: I can confirm that it does

DDvO: > You should just be able to do ./config here. It should auto-detect the environment correctly 

Yes, it picks the default architecture (likely depending on the MSYS2 variant installed).
I've added this as a third option.

> I don't have easy access to a Windows environment at the moment to confirm its still working...my Windows environment died

Mine is also a bit crippled these days. I have not (yet re-)installed MSVC after the last fresh system installation, but the configuration scripts can be tried out regardless.

> I can confirm that it does

Me too. Yet strangely with MSYS2 the configuration takes several minutes here.
Also other perl and make runs (e.g., for converting the POD files) are painfully slow.
Looks like there is an issue with the MSYS2 file system access, which may be a peculiarity with my company's Windows additions like the Trend Micro virus scanner.

NOTES.WIN
DDvO: Done

NOTES.WIN
DDvO: Done

NOTES.WIN
```diff
@@ -1,58 +1,77 @@
...
+
+ - Microsoft Visual C compiler.
+   Since these are proprietry and ever-changing and we cannot test them all,
+   there is unavoidable uncertainty about which versions are supported.
+   Latest version along with couple of previous are certainly supported.
+
+ - Netwide Assembler (NASM), available from https://www.nasm.us
+   Note that NASM is the only supported assembler.
+   Even though Microsoft provided assembler is NOT supported, contemporary
+   64-bit version is exercised through CI of the VC-WIN64A-masm target.
```
mattcaswell, line 45: This sentence seems fairly meaningless for most people. I think we should just drop this.

NOTES.WIN
```diff
@@ -1,58 +1,77 @@
...
+   We recommend Strawberry Perl, available from http://strawberryperl.com/
+   Please read NOTES.PERL for more information, including the use of CPAN.
+   An alternative is ActiveState Perl, https://www.activestate.com/ActivePerl
+   for which you may need to explicitly build the Perl module Win32/Console.pm
+   via https://platform.activestate.com/ActiveState and then download it.
+
+ - Microsoft Visual C compiler.
+   Since these are proprietry and ever-changing and we cannot test them all,
+   there is unavoidable uncertainty about which versions are supported.
+   Latest version along with couple of previous are certainly supported.
```
mattcaswell, line 40: How about:

Since these are proprietary and ever-changing we cannot test them all. Older versions may not work. Use a recent version wherever possible.

NOTES.WIN
```diff
@@ -83,41 +102,63 @@
...
+     ./Configure mingw64 ...
+   or
+     ./config ...
+   for the default architecture.
+
+   Apart from that, follow the Unix / Linux instructions in INSTALL.md.
+
+ * It is also possible to build mingw[64] on Linux or Cygwin.
+
+   In this case configure with corresponding --cross-compile-prefix= option.
```
mattcaswell, line 153: the corresponding

NOTES.WIN
```diff
@@ -147,27 +188,28 @@
 
  If you link with OpenSSL .DLLs, then you're expected to include into
  your application code small "shim" snippet, which provides glue between
```
mattcaswell, line 190: a small "shim"
the glue

DDvO:
Merged - thanks @levitte, @richsalz, and @mattcaswell

PR #12099, opened 2020-06-09
Do not allow dropping Extended Master Secret extension on renegotiaton [1.1.1]

t8m:
This is a cherry-pick of the #12045 for the 1.1.1 branch.

The test cannot be added as it depends on SSL_OP_NO_EXTENDED_MASTER_SECRET which we do not have on 1.1.1.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,10 @@
 
  Changes between 1.1.1g and 1.1.1h [xx XXX xxxx]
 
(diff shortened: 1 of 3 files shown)


t8m:
Matt, does your approval still hold?

t8m:
Merged to 1.1.1 as 4b70970253

Issue #9335, opened 2019-07-10
SSL_CTX_set_ciphersuites does not ignore unknown cipher string


Posolsvetla:
Hi,

the statement in the issue title is well according to the docs for _SSL_CTX_set_ciphersuites_ at https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_cipher_list.html
> SSL_CTX_set_ciphersuites() and SSL_set_ciphersuites() return 1 if the requested ciphersuite list was configured, and 0 otherwise.

**However, imagine this scenario:**

Prerequisites:
1. An application (APP) which uses OpenSSL provided by the OS (or other 3rd party)
2. APP calls _SSL_CTX_set_ciphersuites_ with (e.g.) the cipher string "TLS_CHACHA20_POLY1305_SHA256"

Steps:
1. A new "SUPER_SECURE_NEW_CIPHER" is added into OpenSSL version x.y.z (might be very well 1.1.1)
2. APP is updated to call _SSL_CTX_set_ciphersuites_ with "TLS_CHACHA20_POLY1305_SHA256:SUPER_SECURE_NEW_CIPHER" with the intention to use the new cipher ASAP
3. APP is deployed to the public, but OpenSSL stays not updated (the developer of APP does not have any influence on the deployment of OpenSSL) 
4. As call to _SSL_CTX_set_ciphersuites_ failes, APP could (as a consequence), either:
4.1. stop using TLS 1.3
4.2. continue using TLS 1.3 with the default ciphers
neither of which is ideal

**This behaviour basically prevents deployment of APP which would use "SUPER_SECURE_NEW_CIPHER".**

Possible solutions:
1. _SSL_CTX_set_ciphersuites_ would not fail, if at least one cipher could be set up
2. APP would (at the beginning of step 2) enumerate the ciphers provided by OpenSSL and would not try to set up the ones which are not present

Ad solution 1:
Current behaviour seems to be implemented from the very beginning in #5359, but I didn't find any reasoning for it, just a small mention of BoringSSL in https://github.com/openssl/openssl/pull/7759#issuecomment-457697535
Therefore I don't know if the solution 1 is acceptable or not.
Ad solution 2:
It happens that I already enumerate the ciphers provided by OpenSSL, for a different reason (there is, and won't be, no other possibility, unfortunately). The code is rather silly (although better than in 1.0.2) and I think such code should not be needed in APP before calling of _SSL_CTX_set_ciphersuites_.

[...]

PR #12100, opened 2020-06-09
Fix set_ciphersuites ignore unknown ciphers.

OttoHollmann:
Fix SSL_CTX_set_ciphersuites() and SSL_set_ciphersuites() to ignore unknown cipehrs.
These functions now should behave like SSL_CTX_set_cipher_list() and SSL_set_cipher_list().

This commit fixes bug/feature https://github.com/openssl/openssl/issues/9335

### DIFF
--- a/ssl/ssl_ciph.c
+++ b/ssl/ssl_ciph.c
@@ -1300,6 +1300,8 @@ static int ciphersuite_cb(const char *elem, int len, void *arg)
     if (cipher == NULL) {
         ERR_raise(ERR_LIB_SSL, SSL_R_NO_CIPHER_MATCH);
         return 0;
+        /* Ciphersuite not found but return 1 to parse rest of the list */
+        return 1;
     }
 
     if (!sk_SSL_CIPHER_push(ciphersuites, cipher)) {
@@ -1319,7 +1321,8 @@ static __owur int set_ciphersuites(STACK_OF(SSL_CIPHER) **currciphers, const cha
 
     /* Parse the list. We explicitly allow an empty list */
     if (*str != '\0'
-            && !CONF_parse_list(str, ':', 1, ciphersuite_cb, newciphers)) {
+            && (CONF_parse_list(str, ':', 1, ciphersuite_cb, newciphers) <= 0
+                || sk_SSL_CIPHER_num(newciphers) == 0 )) {
         sk_SSL_CIPHER_free(newciphers);
         return 0;
     }
--- a/ssl/ssl_ciph.c
+++ b/ssl/ssl_ciph.c
@@ -1318,7 +1318,7 @@ static __owur int set_ciphersuites(STACK_OF(SSL_CIPHER) **currciphers, const cha
     /* Parse the list. We explicitly allow an empty list */
     if (*str != '\0'
             && (CONF_parse_list(str, ':', 1, ciphersuite_cb, newciphers) <= 0
-                || sk_SSL_CIPHER_num(newciphers) == 0 )) {
+                || sk_SSL_CIPHER_num(newciphers) == 0)) {
         ERR_raise(ERR_LIB_SSL, SSL_R_NO_CIPHER_MATCH);
         sk_SSL_CIPHER_free(newciphers);
         return 0;
--- a/ssl/ssl_ciph.c
+++ b/ssl/ssl_ciph.c
@@ -1288,21 +1288,17 @@ static int ciphersuite_cb(const char *elem, int len, void *arg)
     /* Arbitrary sized temp buffer for the cipher name. Should be big enough */
     char name[80];
 
-    if (len > (int)(sizeof(name) - 1)) {
-        ERR_raise(ERR_LIB_SSL, SSL_R_NO_CIPHER_MATCH);
-        return 0;
-    }
+    if (len > (int)(sizeof(name) - 1))
+        /* Anyway return 1 so we can parse rest of the list */
+        return 1;
 
     memcpy(name, elem, len);
     name[len] = '\0';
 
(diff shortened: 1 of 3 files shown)


kaduk:
I had this vague recollection that the ciphersuites API was intentionally made to be strict and fail if something unknown was passed (maybe to avoid the risk of weird behavior when all components of the passed list were unknown?), but if Matt doesn't remember such a discussion I'm inclined to believe that I'm imagining things.

OttoHollmann:
There is another option. Keep this ciphersuites API as is and allow TLS 1.3 ciphersuites to be entered in cipher_list. See my first PR https://github.com/openssl/openssl/pull/10590
Then users can decide which API they want to use.

mattcaswell:
> but if Matt doesn't remember such a discussion I'm inclined to believe that I'm imagining things.

...or that I can't even remember what day of the week it is, let alone what previous discussions we might have had on this issue.

mattcaswell:
> There is another option. Keep this ciphersuites API as is and allow TLS 1.3 ciphersuites to be entered in cipher_list. See my first PR #10590

I think this is problematic. The whole reason we didn't go down this route in the first place is that it is possible that legacy configurations have cipher lists which, if applied to the TLSv1.3 ciphersuites, would accidentally disable them. Given that TLSv1.3 ciphersuites are so different to TLSv1.2 ciphersuites we decided to separate them. You can argue the pros and cons of that decision - but I don't think we should be back-tracking on it now.

OttoHollmann:
> I think this is problematic. The whole reason we didn't go down this route in the first place is that it is possible that legacy configurations have cipher lists which, if applied to the TLSv1.3 ciphersuites, would accidentally disable them. Given that TLSv1.3 ciphersuites are so different to TLSv1.2 ciphersuites we decided to separate them. You can argue the pros and cons of that decision - but I don't think we should be back-tracking on it now.

I know that this is very controversial theme, but I'm still think, that my improved cipher_list will deal with accidentally disabling ciphersuites. That's the reason why I added new parameters for specifying cipheruites version.
(Maybe we should move our further discussion into appropriate PR.)

ssl/ssl_ciph.c
```diff
@@ -1311,7 +1311,8 @@ static int ciphersuite_cb(const char *elem, int len, void *arg)
     cipher = ssl3_get_cipher_by_std_name(name);
     if (cipher == NULL) {
         SSLerr(SSL_F_CIPHERSUITE_CB, SSL_R_NO_CIPHER_MATCH);
-        return 0;
+        /* Ciphersuite not found but return 1 to parse rest of the list */
+        return 1;
```
mattcaswell, line 1315: Why is it that this instance returns 0, but the other "no cipher match" failure in this function a few lines above returns 0. Shouldn't that one also return 1?

ssl/ssl_ciph.c
```diff
@@ -1311,7 +1311,8 @@ static int ciphersuite_cb(const char *elem, int len, void *arg)
     cipher = ssl3_get_cipher_by_std_name(name);
     if (cipher == NULL) {
         SSLerr(SSL_F_CIPHERSUITE_CB, SSL_R_NO_CIPHER_MATCH);
```
mattcaswell, line 1313: Setting this error here no longer seems appropriate. Since we're going to continue to process the rest of the list we might get a subsequent cipher match. Probably we should delay raising this error until we notice that the returned stack has no elements in it in `set_ciphersuites`.

mattcaswell (changes requested):

ssl/ssl_ciph.c
OttoHollmann: Yes, it should return 1.

ssl/ssl_ciph.c
OttoHollmann: Yes, error there is no longer appropriate. I've moved this error into function set_ciphersuites().

I was originally thinking about some "warnings", so user should know, that some of entered ciphersuites can't be recognized., that was reason why I keeped this error.

mattcaswell (approved):
Needs second review

t8m:
Please note there is a merge commit in the PR. We do not allow merge commits so the merge commit needs to be dropped from the PR and the PR needs to be properly rebased.

OttoHollmann:
Rebased.

ssl/ssl_ciph.c
```diff
@@ -1331,7 +1329,9 @@ static __owur int set_ciphersuites(STACK_OF(SSL_CIPHER) **currciphers, const cha
 
     /* Parse the list. We explicitly allow an empty list */
     if (*str != '\0'
-            && !CONF_parse_list(str, ':', 1, ciphersuite_cb, newciphers)) {
+            && (CONF_parse_list(str, ':', 1, ciphersuite_cb, newciphers) <= 0
+                || sk_SSL_CIPHER_num(newciphers) == 0 )) {
```
t8m, line 1333: Nit: extra space after 0

t8m (changes requested):
One trivial nit.

ssl/ssl_ciph.c
OttoHollmann: Removed

t8m (approved):
LGTM. @mattcaswell is it still approved?

mattcaswell (approved):
Still good.

t8m:
I'm sorry we neglected this PR for so long time.

I've rebased the branch to fix conflicts. @mattcaswell can you please re-approve?

mattcaswell (approved):
Still good.

t8m:
Merged to master. Unfortunately I forgot to squash the commits which I intended to do originally. Anyway thank you for your contribution and patience.

PR #12103, opened 2020-06-09
Limit sequence number for GOST TLS 1.2 ciphers

NMorozxov:
##### Checklist

### DIFF
--- a/crypto/err/openssl.txt
+++ b/crypto/err/openssl.txt
@@ -3170,6 +3170,7 @@ SSL_R_LENGTH_TOO_LONG:404:length too long
 SSL_R_LENGTH_TOO_SHORT:160:length too short
 SSL_R_LIBRARY_BUG:274:library bug
 SSL_R_LIBRARY_HAS_NO_CIPHERS:161:library has no ciphers
+SSL_R_MAX_SEQUENCE_NUMBER_EXCEEDED:296:max sequence number exceeded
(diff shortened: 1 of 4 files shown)


beldmit:
I'm not sure it's the best possible solution. I try to avoid GOST-specific patches as much as possible and do this on the engine level. 

We have a TLSTREE ctrl call accepting sequence, and I think that it's better to place comparison there. And we have, e.g., OpenVPN with its own sequence generation: the engine seems to be a more relevant place for such a check.

NMorozxov:
ping @openssl/committers need some one else review

NMorozxov:
I think my ping not working

NMorozxov:
@beldmit Can you set the "review required" label. May be  this will attention other developers

beldmit:
Done

t8m:
I agree with Dmitry that this should be more general check - either in the engine/provider or it should be a property of a TLS ciphersuite and checked for all supported ciphersuites where limiting the number makes sense - I suppose all 64bit-block ciphersuites should have limit.

NMorozxov:
@t8m ciphersuite property sounds like good idea

kaduk:
> I agree with Dmitry that this should be more general check - either in the engine/provider or it should be a property of a TLS ciphersuite and checked for all supported ciphersuites where limiting the number makes sense - I suppose all 64bit-block ciphersuites should have limit.

Even some of the AES modes (128-bit blocksize) have reachable limits, too.

PR #12106, opened 2020-06-10
Use rd instead rmdir because if rmdir.exe from cygwin or msys is pres…

mturk:
nmake clean fails if rmdir.exe is present in the path.
Use rd (which is synonym from rmdir) as cmd.exe command


levitte:
It would make sense to change all `mkdir` to `md` as well...

mturk:
Probably would, but since mkdir does not use any of the Microsoft’s options (actually the are none) so no errors are reported. Well, if mkdir.exe does not handle Windows paths with backslashes, it can break. But yes, using MD would be much safer

mturk:
Actually it’s a weird thing. For example cmd.exe will use his internal rmdir and mkdir unless you explicitly call rmdir.exe, but NMAKE first looks in the PATH, and then uses internal cmd.exe’s command. Having git or using msys perl, you are bound with those posix utils

levitte:
Uhmm, what's this static-crt that got sneaked into this PR?

t8m:
Please rebase and drop the commit that apparently does not belong to this PR.

PR #12107, opened 2020-06-10
Ensure CMAC_CTX is correctly initialised (1.1.1)

mattcaswell:
This is a backport of #11972 to the 1.1.1 branch. That PR had the following description:

The function EVP_Cipher() is very low level and has some rather odd return value semantics (see the docs). Essentially we need to treat any return value from that of 0 or less as an error - however the CMAC code was treating it as a boolean.

In fixing this issue I discovered that although we have EVP level CMAC tests we have no tests for the low level API. So I have created a simple test, heavily based on the existing hmactest. This actually uncovered a couple more bugs in CMAC initialisation, where we would attempt to use a CMAC_CTX that has only been partially initialised.


mattcaswell:
Ping?

mattcaswell:
Pushed. Thanks.

PR #12122, opened 2020-06-11
Configure: Try to enable ec_nistp_64_gcc_128 by default

sebastianas:
As far as I undertand ec_nistp_64_gcc_128 isn't enabled by default
because the compiler does not always support it.

Invoke the compiler and look for __SIZEOF_INT128__ in its pre-defines
which is set by gcc if supported.

### DIFF
--- a/Configure
+++ b/Configure
@@ -494,7 +494,6 @@ our %disabled = ( # "what"         => "comment"
                   "crypto-mdebug"       => "default",
                   "crypto-mdebug-backtrace" => "default",
                   "devcryptoeng"        => "default",
-                  "ec_nistp_64_gcc_128" => "default",
                   "egd"                 => "default",
                   "external-tests"      => "default",
                   "fuzz-libfuzzer"      => "default",
@@ -1614,6 +1613,17 @@ unless ($disabled{ktls}) {
 }
 
 push @{$config{openssl_other_defines}}, "OPENSSL_NO_KTLS" if ($disabled{ktls});
+
+unless ($disabled{ec_nistp_64_gcc_128}) {
+    $config{ec_nistp_64_gcc_128}="";
+    my $cc = $config{CROSS_COMPILE}.$config{CC};
+    my $cc_cflags = $config{CFLAGS}[0];
+
+    system("$cc -dM -E $cc_cflags - < /dev/null | grep __SIZEOF_INT128__ > /dev/null");
+    if ($? != 0) {
+        disable('not-supported-by-compiler', 'ec_nistp_64_gcc_128');
+   }
+}
 
 # Get the extra flags used when building shared libraries and modules.  We
 # do this late because some of them depend on %disabled.


mattcaswell:
Hmm...I'm not sure I understand whats going on with the Travis failure here.

bbbrumley:
I don't believe you need to invoke the compiler. Just deprecate the option like you've done, but deprecate the internal macro:

https://github.com/openssl/openssl/blob/7f81aed4bccbeb0ad491b68dd601d576c4956c78/crypto/ec/ec_curve.c#L2832

and switch to `#ifdef __SIZEOF_INT128__` instead :)

kroeckx:
That the compiler supports it is not enough. INSTALL.md lists the
following additional requirements:
 - with little-endian storage of non-byte types
 - that tolerate misaligned memory references

There are only a few targets that tolerate misaligned memory
access. And then I think we should only do it if it's actually
faster.

As discussed in some other PR, I think this really are the only
targets that might support it:
- x86_64 / amd64
- ppc64el
- aarch64 / arm64

I think I've tried arm64, it worked as was faster. I did not try
ppc64el.

bbbrumley:
> That the compiler supports it is not enough

Oh shoot ofc you're right. Thanks Kurt for the sanity check.

FYI my team is working on an automated tool that produces (semi) formally verified, constant time ECC stacks similar to these `EC_METHOD`s. The branch [is here](https://github.com/bbbrumley/openssl/tree/bbb_ecshim). Everything is 100% computer generated, including the OpenSSL rigging. It supports either endian and has no assumptions on alignment.

The work is pretty cool because it auto generates a full stack, both 64-bit and 32-bit, for any curve you want.

We are seeing about 4x speedup on keygen, derive and sign, and 2x on verify compared to generic method. And that will only get better.

richsalz:
@bbbrumley that is *awesome*  Can't wait!

bbbrumley:
> that is _awesome_ Can't wait!

Yea that makes two of us :)

Joint work with @JJChiDguez @luinxz @beldmit and Igor Ustinov. I'm super pumped about it.

@sebastianas sorry not trying to hijack your thread here. If you can find a way to match the restrictions from @kroeckx, this is a very useful PR.

beldmit:
My role is very limited - just GOST curves benchmarking :)

mattcaswell:
> FYI my team is working on an automated tool that produces (semi) formally verified, constant time ECC stacks similar to these EC_METHODs. The branch is here. Everything is 100% computer generated, including the OpenSSL rigging. It supports either endian and has no assumptions on alignment.
>
> The work is pretty cool because it auto generates a full stack, both 64-bit and 32-bit, for any curve you want.

COOL!! Is this likely to hit before the 3.0 beta1 feature freeze?

p-steuer:
I confim it does not work on s390 due to endian issues. I havent looked into it since it dont expect it to outperform s390 HW acceleration.

PR #12123, opened 2020-06-12
property: default properties don't work with a NULL query.

### LINKED ISSUE #11594: EVP_set_default_properties() does not work
The function `EVP_set_default_properties()` does not seem to be working. I created the following test program:

````
#include <stdio.h>
#include <openssl/evp.h>

int main(void)
{
    EVP_MD *md;

    if (!EVP_set_default_properties(NULL, "provider=wibble"))
        printf("Failed to set default properties\n");

    md = EVP_MD_fetch(NULL, "SHA2-256", NULL);
    if (md == NULL)
        printf("Failed to fetch the digest\n");
    else
        printf("Succeeded fetching the digest\n");

    EVP_MD_free(md);

    return 0;
}

````

The output I received was "Succeeded fetching the digest". But I expected it to fail, because no implementations match the specified criteria.

I also note that `EVP_set_default_properties()` does not exist in the test suite - so we should fix that.

### PR DESCRIPTION
**Pull request by paulidale**
Specifying an algorithm with a NULL query ignores the default properties.

Fixes #11594

### DIFF
--- a/crypto/property/property.c
+++ b/crypto/property/property.c
@@ -60,7 +60,6 @@ struct ossl_method_store_st {
     OPENSSL_CTX *ctx;
     size_t nelem;
     SPARSE_ARRAY_OF(ALGORITHM) *algs;
-    OSSL_PROPERTY_LIST *global_properties;
     int need_flush;
     CRYPTO_RWLOCK *lock;
 };
@@ -74,7 +73,34 @@ typedef struct {
 DEFINE_SPARSE_ARRAY_OF(ALGORITHM);
 
 static void ossl_method_cache_flush(OSSL_METHOD_STORE *store, int nid);
-static void ossl_method_cache_flush_all(OSSL_METHOD_STORE *c);
+
+/* Global properties are stored per library context */
+static void ossl_ctx_global_properties_free(void *vstore)
+{
+    OSSL_PROPERTY_LIST **plp = vstore;
+
+    if (plp != NULL) {
+        ossl_property_free(*plp);
+        OPENSSL_free(plp);
+    }
+}
+
+static void *ossl_ctx_global_properties_new(OPENSSL_CTX *ctx)
+{
+    return OPENSSL_zalloc(sizeof(OSSL_PROPERTY_LIST **));
+}
+
+
+static const OPENSSL_CTX_METHOD ossl_ctx_global_properties_method = {
+    ossl_ctx_global_properties_new,
+    ossl_ctx_global_properties_free,
+};
+
+OSSL_PROPERTY_LIST **ossl_ctx_global_properties(OPENSSL_CTX *libctx)
+{
+    return openssl_ctx_get_data(libctx, OPENSSL_CTX_GLOBAL_PROPERTIES,
+                                &ossl_ctx_global_properties_method);
+}
 
 static int ossl_method_up_ref(METHOD *method)
 {
@@ -166,7 +192,6 @@ void ossl_method_store_free(OSSL_METHOD_STORE *store)
     if (store != NULL) {
         ossl_sa_ALGORITHM_doall(store->algs, &alg_cleanup);
         ossl_sa_ALGORITHM_free(store->algs);
-        ossl_property_free(store->global_properties);
         CRYPTO_THREAD_lock_free(store->lock);
         OPENSSL_free(store);
     }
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: crypto/property/property.c [hunk 1]
```diff
@@ -324,27 +324,31 @@ int ossl_method_store_fetch(OSSL_METHOD_STORE *store, int nid,
...
     if (prop_query == NULL) {
-        if ((impl = sk_IMPLEMENTATION_value(alg->impls, 0)) != NULL) {
-            best_method = &impl->method;
-            ret = 1;
+        if (store->global_properties == NULL) {
+            if ((impl = sk_IMPLEMENTATION_value(alg->impls, 0)) != NULL) {
+                best_method = &impl->method;
+                ret = 1;
+            }
+            goto fin;
```
levitte, line 332: Having this bit smack in the middle of property query processing makes for a confusing read, at least to me.  May I suggest this rearrangement?

``` C
    if (prop_query != NULL) {
        p2 = pq = ossl_parse_query(store->ctx, prop_query);
    }
    if (store->global_properties != NULL) {
        if (pq == NULL) {
            pq = store->global_properties;
        } else {
            p2 = ossl_property_merge(pq, store->global_properties);
            if (p2 == NULL)
                goto fin;
            ossl_property_free(pq);
            pq = p2;
        }
    }

    if (pq == NULL) {
        if ((impl = sk_IMPLEMENTATION_value(alg->impls, 0)) != NULL) {
            best_method = &impl->method;
            ret = 1;
        }
        goto fin;
    }
    optional = ossl_property_has_optional(pq);
    for (j = 0; j < sk_IMPLEMENTATION_num(alg->impls); j++) {
```
```

File: crypto/property/property.c [hunk 2]
```diff
@@ -358,7 +362,7 @@ int ossl_method_store_fetch(OSSL_METHOD_STORE *store, int nid,
     else
         ret = 0;
     ossl_property_unlock(store);
-    ossl_property_free(pq);
+    ossl_property_free(p2);
```
levitte, line 365: Make sure to initialise `p2`

paulidale: `p2` is initialised to NULL above.

I welcome suggestions and would love a patch commit to correct my mis-comprehensions.

The bottom line is that I understand the property code well enough but I do not understand the higher levels.

paulidale: I.e. add a patch here or create a new PR.

levitte: D'oh, I'm blind...

test/evp_extra_test.c
```diff
@@ -471,6 +471,35 @@ static EVP_PKEY *load_example_hmac_key(void)
     return pkey;
 }
 
+static int test_EVP_set_default_properties(void)
+{
+    OPENSSL_CTX *ctx;
+    EVP_MD *md;
```
levitte, line 477: Travis tells us that `md` needs to be initialised

levitte:
> I welcome suggestions and would love a patch commit to correct my misscomprehensions.

Done

paulidale:
@levitte I'm beyond being able to reason about the code paths tonight.  Feel free to push changes to this branch and I'll review them tomorrow.

levitte:
> @levitte I'm beyond being able to reason about the code paths tonight. Feel free to push changes to this branch and I'll review them tomorrow.

Will do

crypto/property/property.c
```diff
@@ -323,28 +323,33 @@ int ossl_method_store_fetch(OSSL_METHOD_STORE *store, int nid,
...
-            goto fin;
-        ossl_property_free(pq);
-        pq = p2;
-    }
     optional = ossl_property_has_optional(pq);
     for (j = 0; j < sk_IMPLEMENTATION_num(alg->impls); j++) {
         impl = sk_IMPLEMENTATION_value(alg->impls, j);
         score = ossl_property_match_count(pq, impl->properties);
-        if (score > best) {
+        if (score > best && score >= 0) {
```
paulidale, line 352: Note to self, the `score >= 0` check is redundant.  Negative scores mean no match found and this equals the starting value for best -- thus the condition is always false.

paulidale:
Thanks for the rework.  Much appreciated.

paulidale:
I'm not entirely sure I like this fix (the final commit), it boils down to the default properties being associated with the store not the library context.

levitte:
I agree. The default props should be stored directly in the library context. I think that was already mentioned in #11594

paulidale:
I'm happy to pull the last commit in favour of a better fix.

paulidale:
The default property query is now per library context instead of per store.

paulidale:
Ping? @levitte ?

paulidale:
Ping @openssl/committers

paulidale:
Merged to master thanks.

PR #12125, opened 2020-06-12
util/find-doc-nits: Do not read "missing" files when -u is given

### LINKED ISSUE #12117: find-doc-nits is broken
Ever since #11476, the command `find-doc-nits -u` no longer works. It returns no output:

````
$ util/find-doc-nits -u
$
````
Optimistically I might think that means we've documented everything...but somehow I doubt that :-)

### PR DESCRIPTION
**Pull request by levitte**
Fixes #12117

### CODE REVIEW

**Comment by mattcaswell** [comment]
That doesn't seem to fix it:

````
$ util/find-doc-nits -u
# 165 macros undocumented (count is approximate)
````

i.e. I still don't get the number of undocumented functions in libssl or libcrypto.

Also, by the way, I note that `find-doc-nits -c` also doesn't work:

````
$ util/find-doc-nits -c
$
````

**Comment by levitte** [comment]
Also, by the way, I note that `find-doc-nits -c` also doesn't work

Can you actually point at something it should list?  I tried just removing something from a .pod, and got this:

``` console
$ ./util/find-doc-nits -c
doc/man1/openssl-asn1parse.pod: undocumented option -in
```

levitte:
Ping!

mattcaswell:
> Can you actually point at something it should list? I tried just removing something from a .pod, and got this:

Oops. Looks like I'm wrong about the -c option. The -u still isn't working though.

levitte:
I hope the new commit fixes -u well enough

paulidale:
Travis failures are likely relevant since they are doc-nits related.

mattcaswell:
> Travis failures are likely relevant since they are doc-nits related.

The travis failures are due to newly added functions that have not been documented. These should have caused travis to fail before, but the failure was masked because find-doc-nits is broken!

It does not seem right for this PR to have to fix all of those. I'd suggest this PR (once approved) should go in as is. That will break travis but we should then fix it by providing the  missing documentation as a matter of urgency.

levitte:
Uhmmm, I could add those entries in the appropriate "missing" files as part of this PR.  That's no biggie

mattcaswell:
> Uhmmm, I could add those entries in the appropriate "missing" files as part of this PR. That's no biggie

Ah, yes ok. While I thoroughly dislike adding new "missing" entries, it seems appropriate in this case.

levitte:
> While I thoroughly dislike adding new "missing" entries, it seems appropriate in this case.

I agree with you.  Thanks

levitte:
Added

levitte:
> Perhaps we should add an issue highlighting the missing documentation for these newly added functions?

Good idea.  I'll do that

levitte:
Merged

e3ce33b3b5ab773bbf4156074d9b37b33ac3744d util/find-doc-nits: Do not read "missing" files when -u is given
c4de5d22aa189d357aec18e4a744e3747bd545af util/find-doc-nits: Modernise printem()
3fd16304f4b33814ea481fe965d79675a5f4f2c3 Missing documentation missing, let's note that down

The "Small cleanup of doc/man7/EVP_PKEY-DH.pod" was dropped, it was apparently fixed elsewhere.

Issue #12050, opened 2020-06-04
Build failure of OpenSSL 3.0.0 alpha3 on FreeBSD/i386


tetlowgm:
Build is failing on FreeBSD/i386.

`uname` information:
```
$ uname -a
FreeBSD xxx-i386.freebsd.org 12.1-STABLE FreeBSD 12.1-STABLE #0 r356856: Sat Jan 18 03:52:01 UTC 2020     xxx@xxx.freebsd.org:/usr/obj/usr/src/sys/CLUSTER12  i386
```

Commands run:
`./config`
`make`

Error:
```
cc -fPIC -pthread -Wa,--noexecstack -Qunused-arguments -Wall -O3 -fomit-frame-pointer -shared -nostdlib    -o libcrypto.so.3  <giant list of .o files> providers/libimplementations.a providers/libcommon.a providers/libnonfips.a  -pthread
ld: error: can't create dynamic relocation R_386_32 against symbol: CAST_S_table0 in readonly segment; recompile object files with -fPIC or pass '-Wl,-z,notext' to allow text relocations in the output
>>> defined in crypto/cast/libcrypto-shlib-c_skey.o
>>> referenced by crypto/cast/libcrypto-shlib-cast-586.o:(CAST_encrypt)

ld: error: can't create dynamic relocation R_386_32 against symbol: CAST_S_table1 in readonly segment; recompile object files with -fPIC or pass '-Wl,-z,notext' to allow text relocations in the output
>>> defined in crypto/cast/libcrypto-shlib-c_skey.o
>>> referenced by crypto/cast/libcrypto-shlib-cast-586.o:(CAST_encrypt)

ld: error: can't create dynamic relocation R_386_32 against symbol: CAST_S_table2 in readonly segment; recompile object files with -fPIC or pass '-Wl,-z,notext' to allow text relocations in the output
>>> defined in crypto/cast/libcrypto-shlib-c_skey.o
>>> referenced by crypto/cast/libcrypto-shlib-cast-586.o:(CAST_encrypt)

ld: error: can't create dynamic relocation R_386_32 against symbol: CAST_S_table3 in readonly segment; recompile object files with -fPIC or pass '-Wl,-z,notext' to allow text relocations in the output
>>> defined in crypto/cast/libcrypto-shlib-c_skey.o
>>> referenced by crypto/cast/libcrypto-shlib-cast-586.o:(CAST_encrypt)

[...]

PR #12128, opened 2020-06-13
Fix logic error for building x86 CAST assembly

kaduk:
The assembly code is not PIC, so we should only try to build it
when the configuration has disabled PIC, not the other way around.

Noted by @kroeckx at https://github.com/openssl/openssl/issues/12050#issuecomment-640253738


kaduk:
Pushed to master; closing

PR #12142, opened 2020-06-13
Refactor BN_R_NO_INVERSE logic in internal functions

### LINKED ISSUE #12129: About BN_R_NO_INVERSE issue
Hi，
    It seems this function BIGNUM *BN_mod_inverse(BIGNUM *in） missed some logic int following line.
https://github.com/openssl/openssl/blob/ffb6c86663f8d1048b8bac4023c0e41d3ec5891d/crypto/bn/bn_gcd.c#L49
   When int_bn_mod_inverse go into the following logic, https://github.com/openssl/openssl/blob/ffb6c86663f8d1048b8bac4023c0e41d3ec5891d/crypto/bn/bn_gcd.c#L49
 *pnoinv = 0  always execute, if errors happen, this *pnoinv is always 0, even though BN_R_NO_INVERSE error set.
So maybe we should pass pnoinv to 
static BIGNUM *BN_mod_inverse_no_branch(BIGNUM *in, const BIGNUM *a, const BIGNUM *n,BN_CTX *ctx)  
   function and Set pnoinv to 1 at
https://github.com/openssl/openssl/blob/ffb6c86663f8d1048b8bac4023c0e41d3ec5891d/crypto/bn/bn_gcd.c#L507

romen:
Closes #12129

_**NOTE for reviewers**: I strongly suggest reviewing commit by commit, as one of the separate commits is reordering function definitions, which makes it look like a big set of changes even when the changes are quite minimal._

As described in https://github.com/openssl/openssl/issues/12129 the readability of the internal functions providing the two alternative implementations for `BN_mod_inverse()` is a bit lacking.

Both these functions are now completely internal, so we have the flexibility needed to slightly improve readability and remove unnecessary NULL checks.

The main changes here are:
- rename `BN_mod_inverse_no_branch()` as `bn_mod_inverse_no_branch()`: this function is `static` so it is not even visible within the rest of libcrypto. By convention upcase prefixes are reserved for public functions.
- remove `if (pnoinv == NULL)` checks in `int_bn_mod_inverse()`: this function is internal to the BN module and we can guarantee that all callers pass non-NULL arguments.
- `bn_mod_inverse_no_branch()` takes an extra `int *pnoinv` argument, so that it can signal if no inverse exists for the given inputs: in this way the caller is in charge of raising `BN_R_NO_INVERSE` as it is the case for the non-consttime implementation of `int_bn_mod_inverse()`.

I kept the following 3 commits separated as they might require further discussion.

## Create a temporary `BN_CTX` if needed in the public function

> commit 1824fadc9dc2e1bcaf8cc217c445efbf5134d476
> 
>     - `BN_mod_inverse()` is a public function and must guarantee that the
>       internal functions providing the actual implementation receive valid
>       arguments. If the caller passes a NULL `BN_CTX` we create a temporary
>       one for internal use.

## Reorder function definitions to avid forward declarations

> commit f22d828c5c708b5276c756a4f0523479d1bd85a9
> 
>     @ This commit only reorders the existing functions, no other edit is
>     @ included.
>     @
>     - reorder function definitions in `crypto/bn/bn_gcd.c` to avoid forward
>       declaration of `static` functions (in preparation for inlining)

## Inline `bn_mod_inverse_no_branch()`

> commit 7e7cb711b5585827e1722736c85aeffef5fa6b23
> 
>     - inline `bn_mod_inverse_no_branch()`.

### DIFF
--- a/crypto/bn/bn_gcd.c
+++ b/crypto/bn/bn_gcd.c
@@ -17,9 +17,10 @@
  * This is a static function, we ensure all callers in this file pass valid
  * arguments: all passed pointers here are non-NULL.
  */
-static BIGNUM *bn_mod_inverse_no_branch(BIGNUM *in,
-                                        const BIGNUM *a, const BIGNUM *n,
-                                        BN_CTX *ctx, int *pnoinv)
+static ossl_inline
+BIGNUM *bn_mod_inverse_no_branch(BIGNUM *in,
+                                 const BIGNUM *a, const BIGNUM *n,
+                                 BN_CTX *ctx, int *pnoinv)
 {
     BIGNUM *A, *B, *X, *Y, *M, *D, *T, *R = NULL;
     BIGNUM *ret = NULL;
--- a/crypto/bn/bn_gcd.c
+++ b/crypto/bn/bn_gcd.c
@@ -10,31 +10,182 @@
 #include "internal/cryptlib.h"
 #include "bn_local.h"
 
+/*
+ * bn_mod_inverse_no_branch is a special version of BN_mod_inverse. It does
+ * not contain branches that may leak sensitive information.
+ *
+ * This is a static function, we ensure all callers in this file pass valid
+ * arguments: all passed pointers here are non-NULL.
+ */
 static BIGNUM *bn_mod_inverse_no_branch(BIGNUM *in,
                                         const BIGNUM *a, const BIGNUM *n,
-                                        BN_CTX *ctx, int *pnoinv);
-
-/* solves ax == 1 (mod n) */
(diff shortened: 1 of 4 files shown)


romen:
## Backport to 1.1.1

Most of the changes here should apply trivially also to 1.1.1, but I expect some required changes related to the internal header and to the error function codes.

I will open a separate PR once this is approved dedicated to 1.1.1 and will update this comment.

romen:
Merged to `master` as 5d8b3a3ef2941b8822523742a0408ca6896aa65d

Thanks!

romen:
See #12169 for a dedicated backport PR.

PR #12144, opened 2020-06-13
EVP: allow empty strings to EVP_Decode* functions

### LINKED ISSUE #12143: EVP_DecodeBlock dereferences *f pointer even if n == 0
https://github.com/openssl/openssl/blob/e7fb44e7c3f7a37ff83a6b69ba51a738e549bf5c/crypto/evp/encode.c#L426

Does it expect to find a null terminator if `n == 0` ?
Maybe move the check `(n > 0)` to the left side of condition?

levitte:
This is a simple check order correction.

Fixes #12143

### DIFF
--- a/crypto/evp/encode.c
+++ b/crypto/evp/encode.c
@@ -423,7 +423,7 @@ static int evp_decodeblock_int(EVP_ENCODE_CTX *ctx, unsigned char *t,
         table = data_ascii2bin;
 
     /* trim white space from the start of the line. */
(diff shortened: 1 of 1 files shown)


levitte:
I've checked; this PR cherry-picks cleanly in 1.1.1

kaduk:
Setting ready-to-merge directly, since Pauli's review is more than 24 hours old.

paulidale:
Merged to master and 1.1.1.

PR #12153, opened 2020-06-15
Add X509 related libctx changes.

slontis:
- In order to not add many X509_XXXX_with_libctx() functions the libctx and propq may be stored in the X509 object via a call to X509_new_with_libctx().
- Loading via PEM_read_bio_X509() or d2i_X509() should pass in a created cert using X509_new_with_libctx().
- Renamed some XXXX_ex() to XXX_with_libctx() for X509 API's.
- Removed the extra parameters in check_purpose..
- X509_digest() has been modified so that it expects a const EVP_MD object() and then internally it does the fetch when it needs to (via ASN1_item_digest_with_libctx()).
- Added API's that set the libctx when they load such as X509_STORE_new_with_libctx() so that the cert chains can be verified.

##### Checklist


slontis:
The size of PR #11884 is getting quite large - so I have tried to separate out some bits into smaller PR's to make it easier to review.

crypto/x509/x_all.c
```diff
@@ -443,19 +438,19 @@ int X509_pubkey_digest(const X509 *data, const EVP_MD *type,
     return EVP_Digest(key->data, key->length, md, len, type, NULL);
 }
 
-int X509_digest(const X509 *data, const EVP_MD *type, unsigned char *md,
+int X509_digest(const X509 *cert, const EVP_MD *md, unsigned char *data,
                 unsigned int *len)
 {
-    if (type == EVP_sha1() && (data->ex_flags & EXFLAG_SET) != 0
-            && (data->ex_flags & EXFLAG_INVALID) == 0) {
+    if (md == EVP_sha1() && (cert->ex_flags & EXFLAG_SET) != 0
```
mattcaswell, line 444: EVP_MD_is_a()?

doc/man3/X509_STORE_new.pod
```diff
@@ -2,22 +2,27 @@
...
+ X509_STORE *X509_STORE_new_with_libctx(OPENSSL_CTX *libctx, const char *propq);
  void X509_STORE_free(X509_STORE *v);
  int X509_STORE_lock(X509_STORE *v);
  int X509_STORE_unlock(X509_STORE *v);
  int X509_STORE_up_ref(X509_STORE *v);
 
 =head1 DESCRIPTION
 
-The X509_STORE_new() function returns a new X509_STORE.
+The X509_STORE_new_with_libctx() function returns a new X509_STORE and sets
```
mattcaswell, line 22: Seems to be half a sentence

crypto/asn1/a_digest.c
```diff
@@ -48,20 +48,35 @@ int ASN1_digest(i2d_of_void *i2d, const EVP_MD *type, char *data,
...
-    if (!str)
+    if (str == NULL)
         return 0;
 
-    if (!EVP_Digest(str, i, md, len, type, NULL)) {
-        OPENSSL_free(str);
-        return 0;
-    }
+    if (EVP_MD_provider(md) == NULL)
+         fetched_md = EVP_MD_fetch(libctx, EVP_MD_name(md), propq);
```
mattcaswell, line 65: What if `md` is an ENGINE provided implementation? Should we prioritise the ENGINE implementation over a fetched one? In some cases (e.g. GOST??) there might not even be a fetched implementation.

ssl/s3_lib.c
```diff
@@ -4000,7 +4000,8 @@ long ssl3_ctx_ctrl(SSL_CTX *ctx, int cmd, long larg, void *parg)
                 return 0;
             }
         }
-        if (!X509v3_cache_extensions((X509 *)parg, ctx->libctx, ctx->propq)) {
+        X509_set0_libctx((X509 *)parg, ctx->libctx, ctx->propq);
+        if (!X509v3_cache_extensions((X509 *)parg)) {
```
mattcaswell, line 4004: I think none of these X509v3_cache_extensions() calls are required any more in libssl. They were added recently because of a lack of X509_set0_libctx.

include/openssl/x509v3.h
```diff
@@ -571,7 +571,7 @@ GENERAL_NAME *v2i_GENERAL_NAME_ex(GENERAL_NAME *out,
                                   X509V3_CTX *ctx, CONF_VALUE *cnf,
                                   int is_nc);
 
-int X509v3_cache_extensions(X509 *x, OPENSSL_CTX *libctx, const char *propq);
+int X509v3_cache_extensions(X509 *x);
```
mattcaswell, line 574: Documentation update for this?

ssl/statem/statem_clnt.c
```diff
@@ -1897,6 +1897,7 @@ MSG_PROCESS_RETURN tls_process_server_certificate(SSL *s, PACKET *pkt)
                      ERR_R_MALLOC_FAILURE);
             goto err;
         }
+        X509_set0_libctx(x, s->ctx->libctx, s->ctx->propq);
```
mattcaswell, line 1900: Probably we should do this straight after the d2i. We're already using `x` in the `tls_parse_all_extensions` call so it should have the libctx set by the time we do that.

ssl/statem/statem_srvr.c
```diff
@@ -3717,6 +3717,7 @@ MSG_PROCESS_RETURN tls_process_client_certificate(SSL *s, PACKET *pkt)
                      ERR_R_MALLOC_FAILURE);
             goto err;
         }
+        X509_set0_libctx(x, s->ctx->libctx, s->ctx->propq);
```
mattcaswell, line 3720: Similiarly, I think this should be straight after the d2i.

ssl/s3_lib.c
slontis: Not sure I quite understand the original intent..
X509v3_cache_extensions() is doing a lot of checks.. Are you saying they will get done later when needed?? So they were only being done here so the libctx was known?

slontis: Ahh ok I get it..

crypto/asn1/a_digest.c
slontis: What are you suggesting this code should look like? Do you mean this??
     if (fetched_md == NULL)
         fetched_md = md;

mattcaswell: Perhaps this?:

````
if (EVP_MD_provider(md) == NULL && ENGINE_get_digest_engine(EVP_MD_type(md)) == NULL)
    fetched_md = EVP_MD_fetch(libctx, EVP_MD_name(md), propq);
````

Comment (slontis) [reply crypto/asn1/a_digest.c:65]: Ok I have changed it..
I would imagine this is not going to be the only case where this is going to happen.

File: ssl/s3_lib.c [hunk 8]
```diff
@@ -4000,10 +4000,7 @@ long ssl3_ctx_ctrl(SSL_CTX *ctx, int cmd, long larg, void *parg)
                 return 0;
             }
         }
-        if (!X509v3_cache_extensions((X509 *)parg, ctx->libctx, ctx->propq)) {
-            SSLerr(0, ERR_LIB_X509);
-            return 0;
-        }
+        X509_set0_libctx((X509 *)parg, ctx->libctx, ctx->propq);
```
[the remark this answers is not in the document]
levitte: I gotta say that I agree with Matt.  Most of all, if an application has tied the X509 to a libctx (if that's the inherent meaning), then it's a damn surprise to the application if we start mucking around with that!

slontis:
I have temporarily merged in @levitte code for OPENSSL_CTX_set0_default().
The commit message needs changing and docs need to be added (once we are happy with the API's that have been added). 

X509_set0_libctx() has been replaced by X509_new_with_libctx().
The new is important for some of the X509 read functions...
eg.. 
```
    cert = X509_new_with_libctx(libctx, NULL);

    PEM_read_bio_X509(in, &cert, ...);
       OR
    d2i_X509(&cert, in, len);

NOTE that d2i_X509() calls X509v3_cache_extensions if the cert object is passed in.


```

ssl/s3_lib.c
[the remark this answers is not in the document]
slontis: Ok done what was suggested.. see what you think now..

doc/man3/X509_STORE_new.pod
```diff
@@ -2,22 +2,28 @@
...
  void X509_STORE_free(X509_STORE *v);
  int X509_STORE_lock(X509_STORE *v);
  int X509_STORE_unlock(X509_STORE *v);
  int X509_STORE_up_ref(X509_STORE *v);
 
 =head1 DESCRIPTION
 
-The X509_STORE_new() function returns a new X509_STORE.
+The X509_STORE_new_with_libctx() function returns a new X509_STORE and sets the
+objects library context I<libctx> and property query I<propq>.
```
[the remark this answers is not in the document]
slontis: Ok cleaned up. The only ones remaining now are of the form X509_STORE_load_XXX_with_libctx

crypto/x509/x509_lu.c
```diff
@@ -350,6 +357,7 @@ static int x509_store_add(X509_STORE *store, void *x, int crl) {
     } else {
         obj->type = X509_LU_X509;
         obj->data.x509 = (X509 *)x;
+        X509_set0_libctx(obj->data.x509, store->libctx, store->propq);
```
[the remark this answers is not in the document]
slontis: removed...

slontis:
Rebased and fixed up documentation.
ping

doc/man3/X509_sign.pod
```diff
@@ -32,16 +31,13 @@ X509_sign() signs certificate I<x> using private key I<pkey> and message
 digest I<md> and sets the signature in I<x>. X509_sign_ctx() also signs
 certificate I<x> but uses the parameters contained in digest context I<ctx>.
 
-X509_verify_ex() verifies the signature of certificate I<x> using public key
+X509_verify() verifies the signature of certificate I<x> using public key
 I<pkey>. Any cryptographic algorithms required for the verification are fetched
 using the library context I<libctx> and the property query string I<propq>. Only
```
mattcaswell, line 36: There is now no  libctx passed to this function.

test/sslapitest.c
```diff
@@ -799,11 +799,17 @@ static int execute_test_large_message(const SSL_METHOD *smeth,
 
     if (!TEST_ptr(certbio = BIO_new_file(cert, "r")))
         goto end;
-    chaincert = PEM_read_bio_X509(certbio, NULL, NULL, NULL);
+
+    chaincert = X509_new_with_libctx(libctx, NULL);
+    if (chaincert == NULL) {
+        SSLerr(0, ERR_R_MALLOC_FAILURE);
```
mattcaswell, line 805: Err....`SSLerr`? In a test? Should be `TEST_ptr` - which you seem to have removed below?

doc/man3/X509_set0_libctx.pod
```diff
@@ -0,0 +1,58 @@
...
+
+=head1 NAME
+
+X509_set0_libctx - set X509 certificates library context
+
+=head1 SYNOPSIS
+
+ #include <openssl/x509.h>
+
+ void X509_set0_libctx(X509 *x509, OPENSSL_CTX *libctx, const char *propq);
```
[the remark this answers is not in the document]
slontis: I have removed this function for now.. There is a X509_new_with_libctx() instead.

include/openssl/x509v3.h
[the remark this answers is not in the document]
slontis: Ok renamed

crypto/x509/x509_local.h
```diff
@@ -75,9 +75,10 @@ struct x509_lookup_method_st {
     int (*init) (X509_LOOKUP *ctx);
     int (*shutdown) (X509_LOOKUP *ctx);
     int (*ctrl) (X509_LOOKUP *ctx, int cmd, const char *argc, long argl,
-                 char **ret);
+                 char **ret, OPENSSL_CTX *libctx, const char *propq);
     int (*get_by_subject) (X509_LOOKUP *ctx, X509_LOOKUP_TYPE type,
-                           const X509_NAME *name, X509_OBJECT *ret);
+                           const X509_NAME *name, X509_OBJECT *ret,
+                           OPENSSL_CTX *libctx, const char *propq);
```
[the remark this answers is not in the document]
slontis: ok - I will add another variant get_by_subject_with_libctx annd leave this one alone.. Thanks.

slontis:
@mattcaswell - Some X509 functionality has changed in master which caused a unpleasant rebase.
I have collapsed the reviewed code into the first commit.
The remaining fixup commits are new to address the comments and rebase to master.

include/openssl/store.h
```diff
@@ -84,6 +84,11 @@ int OSSL_STORE_vctrl(OSSL_STORE_CTX *ctx, int cmd, va_list args);
  */
 OSSL_STORE_INFO *OSSL_STORE_load(OSSL_STORE_CTX *ctx);
 
+OSSL_STORE_INFO *OSSL_STORE_load_with_libctx(OSSL_STORE_CTX *ctx,
+                                             OPENSSL_CTX *libctx,
+                                             const char *propq);
+
```
levitte, line 90: This doesn't make much sense, really, if you consider OSSL_STORE for providers, and that scheme based methods will be fetched by `OSSL_STORE_open()`.  What I did was this:

```
+OSSL_STORE_CTX *OSSL_STORE_open_with_libctx(OPENSSL_CTX *libctx,
+                                   const char *uri, const char *properties,
+                                   const UI_METHOD *ui_method, void *ui_data,
+                                   OSSL_STORE_post_process_info_fn post_process,
+                                   void *post_process_data);
```

There should obviously be a corresponding `OSSL_STORE_open_with_libctx_fn` that takes a libctx and propq and saves them away in the loader context.

slontis: Not sure I understand.. Do you mean the function signature is this??

```
OSSL_STORE_CTX *OSSL_STORE_open_with_libctx(OPENSSL_CTX *libctx,
                                            const char *uri, const char *properties,
                                            const UI_METHOD *ui_method, void *ui_data,
                                            OSSL_STORE_post_process_info_fn post_process,
                                            void *post_process_data,
                                            OSSL_STORE_open_with_libctx_fn);
```

slontis: And will that work in PEM_read_bio_PrivateKey_ex
Which does an OSSL_STORE_attach() followed by OSSL_STORE_load()

levitte: > Not sure I understand.. Do you mean the function signature is this??

No.  Why would I pass the method function pointer?  It should be in the `OSSL_STORE_LOADER` that it gets from the URI scheme.

> And will that work in PEM_read_bio_PrivateKey_ex
> Which does an OSSL_STORE_attach() followed by OSSL_STORE_load()

`OSSL_STORE_attach()` already takes a libctx and propq (and propagates them to the OSSL_STORE_LOADER` `attach` method, something you removed?????), so I fail to see the issue.

include/openssl/store.h
```diff
@@ -249,8 +254,6 @@ int OSSL_STORE_LOADER_set_open(OSSL_STORE_LOADER *loader,
 typedef OSSL_STORE_LOADER_CTX *(*OSSL_STORE_attach_fn)(const OSSL_STORE_LOADER
                                                        *loader,
                                                        BIO *bio,
-                                                       OPENSSL_CTX *libctx,
-                                                       const char *propq,
```
levitte, line 253: Why remove these?

include/openssl/store.h
levitte: So basically, what I'm trying to say is that I'd prefer specifying the libctx and propq on init (open or attach) rather than on load.  That corresponds pretty much with how we operate elsewhere.

slontis: Fair enough..
I think the original problem I had was that the libctx and propq are sitting in a object that is not passed to other lower level objects. As they dont know who the parent is they also dont know what the libctx and propq are either. I will have to code a bit to try to remember what exactly was the problem (this was done many weeks ago).

include/openssl/store.h
slontis: @levitte
Inside loader_file.c we have
```
file_load(OSSL_STORE_LOADER_CTX *ctx, const UI_METHOD *ui_method, void *ui_data);
```
And this function needs the libctx/propq.

So are you happy having a libctx/propq stored in ossl_store_loader_ctx_st,
that gets copied during the file_open()?

slontis: ... and file_attach() also

levitte: > So are you happy having a libctx/propq stored in ossl_store_loader_ctx_st, that gets copied during the file_open()?

Yes, that's exactly where I think we should go.

> (I thought I remember doing this before and then you may have complained that there was a libctx in there..)

It's not *specifically* about this.  I believe it's a mistake not to treat structures like RSA as free objects, not attached to a library context...  that was a losing battle though, so here we are, and we do need to pass libctx and propq in ways I didn't expect at the time.  This is one of them.  In other word, the protest I made at the time isn't valid with the code we currently have.

mattcaswell:
Travis failures are relevant.

apps/lib/apps.c
```diff
@@ -613,9 +613,10 @@ static int load_certs_crls(const char *file, int format,
     if (bio == NULL)
         return 0;
 
-    xis = PEM_X509_INFO_read_bio(bio, NULL,
+    xis = PEM_X509_INFO_read_bio_with_libctx(bio, NULL,
                                  (pem_password_cb *)password_callback,
-                                 &cb_data);
+                                 &cb_data,
+                                 app_get0_libctx(), app_get0_propq());
```
mattcaswell, line 619: Indentation looks off now

crypto/x509/x509_vfy.c
```diff
@@ -118,33 +118,32 @@ static int null_callback(int ok, X509_STORE_CTX *e)
  * to match issuer and subject names (i.e., the cert being self-issued) and any
  * present authority key identifier to match the subject key identifier, etc.
  */
-static int x509_self_signed_ex(X509 *cert, int verify_signature,
-                               OPENSSL_CTX *libctx, const char *propq)
+static int x509_self_signed(X509 *cert, int verify_signature)
```
mattcaswell, line 121: I wonder why we need this as a separate function now. Can't this just be collapsed into `X509_self_signed()`

crypto/x509/x509_local.h
```diff
@@ -75,7 +75,7 @@ struct x509_lookup_method_st {
     int (*init) (X509_LOOKUP *ctx);
     int (*shutdown) (X509_LOOKUP *ctx);
     int (*ctrl) (X509_LOOKUP *ctx, int cmd, const char *argc, long argl,
-                 char **ret);
+                 char **ret, OPENSSL_CTX *libctx, const char *propq);
```
mattcaswell, line 78: As with get_by_subject, you also need to retain this one for backwards compat and introduce a ctrl_with_libctx

slontis: Yep - silly me..

slontis:
Fixed memory leak..
Addressed comments related to store open/attach/load.

include/openssl/store.h
```diff
@@ -120,11 +120,12 @@ int OSSL_STORE_close(OSSL_STORE_CTX *ctx);
...
  */
-OSSL_STORE_CTX *OSSL_STORE_attach(BIO *bio, OPENSSL_CTX *libctx,
-                                  const char *scheme, const char *propq,
+OSSL_STORE_CTX *OSSL_STORE_attach(BIO *bio,
+                                  const char *scheme,
                                   const UI_METHOD *ui_method, void *ui_data,
                                   OSSL_STORE_post_process_info_fn post_process,
-                                  void *post_process_data);
+                                  void *post_process_data,
+                                  OPENSSL_CTX *libctx, const char *propq);
```
levitte, line 128: The previous pattern was modelled around the typical fetch call, i.e. starting with the tuple \< libctx, name, propq \>.  Is there a reason to shift that around?  Should we remodel the fetch calls accordingly?

I believe the logic for the fetch call model is "fetch within the *library context* something called *name*, using the given *properties*"

levitte: And yeah, I was expecting the new opening function to look like this:

``` C

OSSL_STORE_CTX *OSSL_STORE_open_with_libctx(
    OPENSSL_CTX *libctx, const char *uri, const char *propq,
    const UI_METHOD *ui_method, void *ui_data,
    OSSL_STORE_post_process_info_fn post_process, void *post_process_data);
```

include/openssl/store.h
```diff
@@ -57,6 +57,11 @@ OSSL_STORE_CTX *OSSL_STORE_open(const char *uri, const UI_METHOD *ui_method,
                                 OSSL_STORE_post_process_info_fn post_process,
                                 void *post_process_data);
 
+OSSL_STORE_CTX *OSSL_STORE_open_with_libctx(
+    const char *uri, const UI_METHOD *ui_method, void *ui_data,
+    OSSL_STORE_post_process_info_fn post_process, void *post_process_data,
+    OPENSSL_CTX *libctx, const char *propq);
```
levitte, line 63: Source code style wise, I believe this isn't OK, as parenthesised things should be aligned with the left parenthesis.

However, I see nothing wrong with this, and I believe it fits our coding style:

``` C
OSSL_STORE_CTX *OSSL_STORE_open_with_libctx
    (const char *uri, const UI_METHOD *ui_method, void *ui_data,
     OSSL_STORE_post_process_info_fn post_process, void *post_process_data,
     OPENSSL_CTX *libctx, const char *propq);
```

slontis: Err.. this style is done all over the place and is (arguably) more readable than the proposed alternative.

Just as this is more readable (also arguably)
```
if (....) {
}
```

include/openssl/store.h
slontis: I would like to stuff them all on the end for all API's (plus find any remaining _ex ones)..
I don't think it changes the language of the fetch much.
i.e: "fetch something called name with the given library context and properties".
@paulidale also mentioned maybe having a struct that contains both the libctx and propq and passing that around instead (since it is almost always done as a pair).

I think @mspncp also mentioned putting the libctx on the end of the list also.
Fixing the existing ones up is another PR though - probably best left to just before beta (when hopefully big PR's are settling down).

slontis:
Tests pass now - look like it is getting closer.

slontis:
Had to rebase and apply a small change now that #11948 has been merged (which added the env variable to set the app_lib_ctx.

include/openssl/store.h
slontis: @levitte are you going to block on either of the above two comments?

include/openssl/store.h
levitte: I have no problems re-arranging them into a \< name, libctx, propq \> tuple, but please keep them together at least.

levitte: In other words, something like this then:

``` C

OSSL_STORE_CTX *OSSL_STORE_open_with_libctx(
    const char *uri, OPENSSL_CTX *libctx, const char *propq,
    const UI_METHOD *ui_method, void *ui_data,
    OSSL_STORE_post_process_info_fn post_process, void *post_process_data);
```

levitte: (fetch functions can easily be remodelled like that)

levitte: As for a struct, that invites quite clumsy programming for not much extra value.  I'm not so keen on that.

include/openssl/store.h
levitte: I have no idea how you think a function declaration compares to an if statement.  They are completely different in my mind (apart from using parentheses in their respective syntax)

levitte: I feel strongly about the \< name, libctx, propq \> tuple staying together.  The code style issue, less so, but I *am* worried that we stray more and more from the style guide, and re-create the mess we had before.

(the coding style needs an update, and there's been several proposals, but not final decisions...)

include/openssl/store.h
slontis: It does until we need to add yet another parameter associated with a libctx and then end up with
_with_libctx_ex :)

slontis:
Addressed @levitte's comments.

include/openssl/store.h
t8m: Unless you make the struct non-public with getters and setters which would be a really horrible to use by callers, it will not help for such kind of API break.

crypto/store/loader_file.c
```diff
@@ -969,39 +970,62 @@ static OSSL_STORE_LOADER_CTX *file_open(const OSSL_STORE_LOADER *loader,
...
+
 static OSSL_STORE_LOADER_CTX *file_attach(const OSSL_STORE_LOADER *loader,
-                                          BIO *bp, OPENSSL_CTX *libctx,
-                                          const char *propq,
+                                          BIO *bp,
                                           const UI_METHOD *ui_method,
-                                          void *ui_data)
+                                          void *ui_data,
+                                          OPENSSL_CTX *libctx,
+                                          const char *propq)
```
levitte, line 1001: Please restore.  Basically, the tuple here is \< bp, libctx, propq \>

include/openssl/store.h
levitte: Please change this signature to, so it's consistent with OSSL_STORE_open_with_libctx:

``` C
OSSL_STORE_CTX *OSSL_STORE_attach(BIO *bio, const char *scheme,
                                  OPENSSL_CTX *libctx, const char *propq,
                                  const UI_METHOD *ui_method, void *ui_data,
                                  OSSL_STORE_post_process_info_fn post_process,
                                  void *post_process_data);
```

slontis:
NOTE that
```
make update
```
 does not like 

```
void func
   (......
    .......)
{
    ...
    OSSL_STOREerr(0, ERR_R_MALLOC_FAILURE);
    ...
}
```
I assume it was looking for the '(' on the same line to indicate a function.. So there is still one in store_lib.c
that is doing
```
void func(
    ....
    ....)
```

slontis:
Also removed the dead functions from include/crypto/store.h

include/openssl/store.h
slontis: > I have no idea how you think a function declaration compares to an if statement.

As far as I am aware the reason for this format
```
if () {
}
```
Is related to scan-ability of the line.
So I am not sure why you think this can't also apply to () when it needs to span multiple lines.

slontis: I have complied in this PR.. It doesn't mean I have to agree :)

slontis:
@levitte signatures changed now.. Will merge in a few hours in there are no more objections.

include/openssl/store.h
```diff
@@ -57,6 +57,11 @@ OSSL_STORE_CTX *OSSL_STORE_open(const char *uri, const UI_METHOD *ui_method,
                                 OSSL_STORE_post_process_info_fn post_process,
                                 void *post_process_data);
 
+OSSL_STORE_CTX *OSSL_STORE_open_with_libctx
+    (const char *uri, OPENSSL_CTX *libctx, const char *propq,
+     const UI_METHOD *ui_method, void *ui_data,
+     OSSL_STORE_post_process_info_fn post_process, void *post_process_data);
```
levitte, line 63: Yes!  Thank you

include/openssl/store.h
```diff
@@ -115,8 +120,8 @@ int OSSL_STORE_close(OSSL_STORE_CTX *ctx);
  * Note that this function is considered unsafe, all depending on what the
  * BIO actually reads.
  */
-OSSL_STORE_CTX *OSSL_STORE_attach(BIO *bio, OPENSSL_CTX *libctx,
-                                  const char *scheme, const char *propq,
+OSSL_STORE_CTX *OSSL_STORE_attach(BIO *bio, const char *scheme,
+                                  OPENSSL_CTX *libctx, const char *propq,
```
levitte, line 124: Yes!  Thank you

include/openssl/store.h
```diff
@@ -239,21 +244,20 @@ const ENGINE *OSSL_STORE_LOADER_get0_engine(const OSSL_STORE_LOADER *loader);
...
-                                                     const char *uri,
-                                                     const UI_METHOD *ui_method,
-                                                     void *ui_data);
+typedef OSSL_STORE_LOADER_CTX *(*OSSL_STORE_open_fn)
+    (const OSSL_STORE_LOADER *loader, const char *uri,
+     const UI_METHOD *ui_method, void *ui_data);
+typedef OSSL_STORE_LOADER_CTX *(*OSSL_STORE_open_with_libctx_fn)
+    (const OSSL_STORE_LOADER *loader,
+     const char *uri, OPENSSL_CTX *libctx, const char *propq,
+     const UI_METHOD *ui_method, void *ui_data);
```
levitte, line 253: Yes!  Thank you

include/openssl/store.h
```diff
@@ -239,21 +244,20 @@ const ENGINE *OSSL_STORE_LOADER_get0_engine(const OSSL_STORE_LOADER *loader);
...
-                                                       BIO *bio,
-                                                       OPENSSL_CTX *libctx,
-                                                       const char *propq,
-                                                       const UI_METHOD
-                                                       *ui_method,
-                                                       void *ui_data);
+typedef OSSL_STORE_LOADER_CTX *(*OSSL_STORE_attach_fn)
+    (const OSSL_STORE_LOADER *loader, BIO *bio,
+     OPENSSL_CTX *libctx, const char *propq,
+     const UI_METHOD *ui_method, void *ui_data);
```
levitte, line 260: Yes!  Thank you

include/openssl/store.h
```diff
@@ -272,6 +276,10 @@ typedef OSSL_STORE_INFO *(*OSSL_STORE_load_fn)(OSSL_STORE_LOADER_CTX *ctx,
                                                void *ui_data);
 int OSSL_STORE_LOADER_set_load(OSSL_STORE_LOADER *loader,
                                OSSL_STORE_load_fn load_function);
+typedef OSSL_STORE_INFO *(*OSSL_STORE_load_with_libctx_fn)
+    (OSSL_STORE_LOADER_CTX *ctx, const UI_METHOD *ui_method, void *ui_data,
+     OPENSSL_CTX *libctx, const char *propq);
```
levitte, line 281: [sigh] there's a pattern throughout OSSL_STORE, where the UI stuff comes last, and diverse callbacks come even more last...

So here's a choice, change this signature in this PR, or leave it to me to do so in #12512.  I would prefer the former, but this PR has been waiting for a while, so...

slontis: another PR would be a good idea :)

levitte: If you choose to change it, here's what I'd prefer to see:

``` C
typedef OSSL_STORE_INFO *(*OSSL_STORE_load_with_libctx_fn)
    (OSSL_STORE_LOADER_CTX *ctx, OPENSSL_CTX *libctx, const char *propq,
     const UI_METHOD *ui_method, void *ui_data);
```

levitte: > another PR would be a good idea :)

Hahaha, I think that making a whole new PR just to re-arrange this parameters is a bit over the top.

slontis: Ahh ok this is a leftover in the header - load_with_libctx_fn is not needed now. Will remove.

slontis: Removed

slontis:
Thanks for reviewing. Merged to master.

PR #12160, opened 2020-06-15
use safe primes in ssl_get_auto_dh()

tomato42:
DH_get_1024_160() and DH_get_2048_224() return parameters from
RFC5114. Those parameters include primes with known small subgroups,
making them unsafe. Change the code to use parameters from
RFC 2409 and RFC 3526 instead (group 2 and 14 respectively).

This patch also adds automatic selection of 4096 bit params for 4096 bit
RSA keys

backport of 7646610 to 1.1.1

see also #12061 and #12060

Signed-off-by: Hubert Kario <hkario@redhat.com>

##### Checklist

### DIFF
--- a/ssl/t1_lib.c
+++ b/ssl/t1_lib.c
@@ -2439,46 +2439,48 @@ int SSL_check_chain(SSL *s, X509 *x, EVP_PKEY *pk, STACK_OF(X509) *chain)
 #ifndef OPENSSL_NO_DH
 DH *ssl_get_auto_dh(SSL *s)
 {
+    DH *dhp = NULL;
+    BIGNUM *p = NULL, *g = NULL;
     int dh_secbits = 80;
-    if (s->cert->dh_tmp_auto == 2)
-        return DH_get_1024_160();
-    if (s->s3->tmp.new_cipher->algorithm_auth & (SSL_aNULL | SSL_aPSK)) {
-        if (s->s3->tmp.new_cipher->strength_bits == 256)
-            dh_secbits = 128;
-        else
-            dh_secbits = 80;
-    } else {
-        if (s->s3->tmp.cert == NULL)
-            return NULL;
-        dh_secbits = EVP_PKEY_security_bits(s->s3->tmp.cert->privatekey);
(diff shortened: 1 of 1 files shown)


ssl/t1_lib.c
```diff
@@ -2440,45 +2440,46 @@ int SSL_check_chain(SSL *s, X509 *x, EVP_PKEY *pk, STACK_OF(X509) *chain)
...
-            p = BN_get_rfc3526_prime_3072(NULL);
-        if (p == NULL || !DH_set0_pqg(dhp, p, NULL, g)) {
-            DH_free(dhp);
-            BN_free(p);
-            BN_free(g);
-            return NULL;
-        }
-        return dhp;
+    DH *dhp = DH_new();
+    BIGNUM *p, *g;
```
kroeckx, line 2457: You have the variables in the middle of a function again, they need to be at the start of a block

ssl/t1_lib.c
```diff
@@ -2440,45 +2440,46 @@ int SSL_check_chain(SSL *s, X509 *x, EVP_PKEY *pk, STACK_OF(X509) *chain)
...
+    if (dhp == NULL)
+        return NULL;
+    g = BN_new();
+    if (g == NULL || !BN_set_word(g, 2)) {
+        DH_free(dhp);
+        BN_free(g);
+        return NULL;
+    }
+    if (dh_secbits >= 192)
+        p = BN_get_rfc3526_prime_8192(NULL);
```
slontis, line 2467: Can you use RFC7919 here instead?

ssl/t1_lib.c
slontis: Set them all at the top with values of NULL.

ssl/t1_lib.c
tomato42: why do we want different behaviour in 1.1.1 and master?

ssl/t1_lib.c
tomato42: done

t8m:
Merged to 1.1.1. Thank you for the PR.

PR #12162, opened 2020-06-16
  Add fix for RSA keygen in FIPS using keysizes 2048 < bits < 3072 

### LINKED ISSUE #11863: Openssl 3.0.0alpha3 fips provider - RSA keygen doesn't allow 2048 < modulus < 3072
Expect to be able to generate RSA keys with bitlen >2048 and <3072 as it is FIPS valid. 

Observation during RSA key gen using old API (EVP_PKEY_CTX_set_rsa_keygen_bits) but with fips provider loaded into NULL context. 

- As expected, setting bits = 2048, 3072 or 4096 succeed no problem.
- Setting bits > 3072 succeeds also.
- But setting bits > 2048 and less than 3072 returns error (with no error info)

```
#include <openssl/provider.h>
#include <openssl/crypto.h>
#include <openssl/evp.h>
#include <openssl/err.h>
#include <openssl/rsa.h>


{
    EVP_PKEY *pkey = NULL;
    EVP_PKEY_CTX *gen_ctx;
    int ret;
    int bits = 2080; /* example failing bits value */

    OSSL_PROVIDER_set_default_search_path(NULL, "/home/user/ossl-install/main/lib/ossl-modules/");
    if (!OPENSSL_CTX_load_config(NULL, "/home/user/ossl-install/cnf/openssl.cnf")) goto end;

    printf("Generating RSA key ( %d bits) \n", bits);
    gen_ctx = EVP_PKEY_CTX_new_id(NID_rsaEncryption, NULL);   

    if (EVP_PKEY_CTX_set_rsa_keygen_bits(gen_ctx, bits) <= 0) goto end;

    printf("... done generating RSA key\n");

end:
    printf ("err:\n");
    ERR_print_errors_fp(stderr);
    EVP_PKEY_free(pkey);
    EVP_PKEY_CTX_free(gen_ctx);
    return 1;
}
```
And the openssl.cnf contains

```
openssl_conf = openssl_init

.include /home/user/ossl-install/cnf/fips.cnf
[openssl_init]
providers = provider_sect
alg_section = algorithm_sect

[provider_sect]
fips = fips_sect

[algorithm_sect]
default_properties = fips=yes
```
And failing output

```
default provider available: 0
fips provider available: 1
Generating RSA key ( 2080 bits) 
err:
```

slontis:
Fixes #11863

##### Checklist

### DIFF
--- a/apps/genpkey.c
+++ b/apps/genpkey.c
@@ -21,13 +21,15 @@
 # include <openssl/engine.h>
 #endif
 
-static int init_keygen_file(EVP_PKEY_CTX **pctx, const char *file, ENGINE *e);
+static int init_keygen_file(EVP_PKEY_CTX **pctx, const char *file, ENGINE *e,
+                            OPENSSL_CTX *libctx, const char *propq);
 static int genpkey_cb(EVP_PKEY_CTX *ctx);
 
 typedef enum OPTION_choice {
     OPT_ERR = -1, OPT_EOF = 0, OPT_HELP,
     OPT_ENGINE, OPT_OUTFORM, OPT_OUT, OPT_PASS, OPT_PARAMFILE,
     OPT_ALGORITHM, OPT_PKEYOPT, OPT_GENPARAM, OPT_TEXT, OPT_CIPHER,
+    OPT_CONFIG,
     OPT_PROV_ENUM
 } OPTION_CHOICE;
 
@@ -41,6 +43,7 @@ const OPTIONS genpkey_options[] = {
     {"algorithm", OPT_ALGORITHM, 's', "The public key algorithm"},
     {"pkeyopt", OPT_PKEYOPT, 's',
      "Set the public key algorithm option as opt:value"},
+     OPT_CONFIG_OPTION,
 
     OPT_SECTION("Output"),
     {"out", OPT_OUT, '>', "Output file"},
@@ -60,6 +63,7 @@ const OPTIONS genpkey_options[] = {
 
 int genpkey_main(int argc, char **argv)
 {
+    CONF *conf = NULL;
     BIO *in = NULL, *out = NULL;
     ENGINE *e = NULL;
     EVP_PKEY *pkey = NULL;
@@ -69,6 +73,8 @@ int genpkey_main(int argc, char **argv)
     OPTION_CHOICE o;
     int outformat = FORMAT_PEM, text = 0, ret = 1, rv, do_param = 0;
     int private = 0;
+    OPENSSL_CTX *libctx = app_get0_libctx();
+    const char *propq = app_get0_propq();
 
     prog = opt_init(argc, argv, genpkey_options);
     while ((o = opt_next()) != OPT_EOF) {
@@ -98,11 +104,11 @@ int genpkey_main(int argc, char **argv)
         case OPT_PARAMFILE:
             if (do_param == 1)
                 goto opthelp;
-            if (!init_keygen_file(&ctx, opt_arg(), e))
+            if (!init_keygen_file(&ctx, opt_arg(), e, libctx, propq))
                 goto end;
             break;
         case OPT_ALGORITHM:
-            if (!init_gen_str(&ctx, opt_arg(), e, do_param))
+            if (!init_gen_str(&ctx, opt_arg(), e, do_param, libctx, propq))
                 goto end;
             break;
         case OPT_PKEYOPT:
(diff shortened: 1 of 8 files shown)


slontis:
Please ignore the first commit (which is the dependent PR #11948)

slontis:
Note that currently genpkey passes when the generated pkey is not successfully written to a file. Does this need to be fixed?

NOTE: until the serializer can by seperated out as another provider, this test will not write out the key.

apps/genpkey.c
```diff
@@ -273,25 +296,27 @@ int init_gen_str(EVP_PKEY_CTX **pctx,
...
+    if (libctx == NULL || e != NULL) {
+        ameth = EVP_PKEY_asn1_find_str(&tmpeng, algname, -1);
 
 #ifndef OPENSSL_NO_ENGINE
-    if (!ameth && e)
-        ameth = ENGINE_get_pkey_asn1_meth_str(e, algname, -1);
+        if (!ameth && e)
+            ameth = ENGINE_get_pkey_asn1_meth_str(e, algname, -1);
 #endif
+        if (!ameth) {
```
mattcaswell, line 306: `ameth == NULL`

test/recipes/15-test_genrsa.t
```diff
@@ -105,3 +115,25 @@ ok(!run(app([ 'openssl', 'genpkey', '-algorithm', 'RSA',
...
+                '-provider_name', 'fips', '-mac_name', 'HMAC',
+                '-macopt', 'digest:SHA256', '-macopt', 'hexkey:00',
+                '-section_name', 'fips_sect'])),
+       "fipsinstall");
+
+    ok(run(app(['openssl', 'genpkey',
+                @prov,
+               '-algorithm', 'RSA',
+               '-pkeyopt', 'bits:2048',
+               '-out', 'genrsatest2080.pem'])),
```
mattcaswell, line 137: Shouldn't this say `bits:2080`?

slontis: You are correct - I know I tested it locally :)

slontis:
The genpkey test is now failing in fips mode - as there is no way to find the serializer to write out the key currently.

slontis:
Now that the base provider works, this uncovered another issue where the RSA object differs between the FIPS_MODULE and the base.. Which is not so good for i2d which uses offsets into this RSA object.

crypto/serializer/serializer_pkey.c
```diff
@@ -150,7 +150,8 @@ static int serializer_EVP_PKEY_to_bio(OSSL_SERIALIZER_CTX *ctx, BIO *out)
     if (ctx->ser == NULL)
         return 0;
 
-    if (ctx->ser->serialize_object == NULL) {
+    if (ctx->ser->serialize_object == NULL
+        || OSSL_SERIALIZER_provider(ctx->ser) != EVP_KEYMGMT_provider(keymgmt)) {
```
mattcaswell, line 154: Side note: I provided the same patch to @romen recently in relation to #12291.

romen: Matt's patch somehow broke the tests for EC in FIPS mode in #12291 that are currently working.

I wish we could merge 12291 before fixing things that break other things!

slontis: So that implies that the export is not working correctly for some edge case in ec and will need to be fixed.
You may notice that this PR is 60 PR's older than your one :) and it  is actually green.

romen: I am not opposed to merge this of course, but I would answer that we have been merging a lot of functionality without adding tests to make sure everything added is covered though.

That one is not green because it only adds tests that should have been showstoppers for previous PRs that got already merged!

slontis: Are you going to tell me that after your tests work that we have covered everything? :)

romen: Obviously 😂

OK, I am rereading my previous comment and it can indeed be read way more aggressively than I intended it to be!

slontis: Thanks for adding extra tests, I cant always think of every possible scenario..

slontis:
Before trying to merge - I noticed that openssl-genpkey.pod.in contained some libctx related fields that were replaced by an environment variable. So a small change was required to delete these.

slontis:
Thanks.. Merged to master.

PR #12164, opened 2020-06-16
Fix segfault in openssl app called with no args.

slontis:
This is a result of removal of interactive mode.
Redirected it to now use 'openssl help'.

##### Checklist


slontis:
Thanks. Merged to master.

PR #12167, opened 2020-06-16
Flag RSA secret BNs as consttime on keygen and checks

romen:
<https://github.com/openssl/openssl/pull/11765> switched the default code path for keygen.

External testing through TriggerFlow highlighted that in several places we failed (once more!) to set the `BN_FLG_CONSTTIME` flag on critical secret values (either long term or temporary values).

This commit tries to make sure that the secret BN values inside the `rsa struct` are always flagged on creation, and that temporary values derived from these secrets are flagged when allocated from a BN_CTX.

Acknowledgments
---------------

Thanks to @Voker57, @bbbrumley, @sohhas, @cpereida for the [OpenSSL Triggerflow CI] ([paper]) through which this defect was detected and tested, and for providing early feedback to fix the issue!

[OpenSSL Triggerflow CI]: https://gitlab.com/nisec/openssl-triggerflow-ci
[paper]: https://eprint.iacr.org/2019/366

### DIFF
--- a/crypto/rsa/rsa_sp800_56b_check.c
+++ b/crypto/rsa/rsa_sp800_56b_check.c
@@ -37,7 +37,14 @@ int rsa_check_crt_components(const RSA *rsa, BN_CTX *ctx)
     r = BN_CTX_get(ctx);
     p1 = BN_CTX_get(ctx);
     q1 = BN_CTX_get(ctx);
-    ret = (q1 != NULL)
+    if (q1 != NULL) {
+        BN_set_flags(p1, BN_FLG_CONSTTIME);
+        BN_set_flags(q1, BN_FLG_CONSTTIME);
+        ret = 1;
+    } else {
+        ret = 0;
+    }
+    ret = ret
           /* p1 = p -1 */
           && (BN_copy(p1, rsa->p) != NULL)
           && BN_sub_word(p1, 1)
@@ -138,7 +145,14 @@ int rsa_check_prime_factor(BIGNUM *p, BIGNUM *e, int nbits, BN_CTX *ctx)
     BN_CTX_start(ctx);
     p1 = BN_CTX_get(ctx);
     gcd = BN_CTX_get(ctx);
-    ret = (gcd != NULL)
+    if (gcd != NULL) {
+        BN_set_flags(p1, BN_FLG_CONSTTIME);
+        BN_set_flags(gcd, BN_FLG_CONSTTIME);
+        ret = 1;
+    } else {
+        ret = 0;
+    }
+    ret = ret
           /* (Step 5d) GCD(p-1, e) = 1 */
           && (BN_copy(p1, p) != NULL)
           && BN_sub_word(p1, 1)
@@ -172,7 +186,16 @@ int rsa_check_private_exponent(const RSA *rsa, int nbits, BN_CTX *ctx)
     lcm = BN_CTX_get(ctx);
     p1q1 = BN_CTX_get(ctx);
     gcd = BN_CTX_get(ctx);
-    ret = (gcd != NULL
+    if (gcd != NULL) {
+        BN_set_flags(p1, BN_FLG_CONSTTIME);
+        BN_set_flags(q1, BN_FLG_CONSTTIME);
+        BN_set_flags(lcm, BN_FLG_CONSTTIME);
+        BN_set_flags(gcd, BN_FLG_CONSTTIME);
+        ret = 1;
+    } else {
+        ret = 0;
+    }
+    ret = (ret
           /* LCM(p - 1, q - 1) */
           && (rsa_get_lcm(ctx, rsa->p, rsa->q, lcm, gcd, p1, q1, p1q1) == 1)
           /* (Step 6a) d < LCM(p - 1, q - 1) */
@@ -241,6 +264,12 @@ int rsa_get_lcm(BN_CTX *ctx, const BIGNUM *p, const BIGNUM *q,
                 BIGNUM *lcm, BIGNUM *gcd, BIGNUM *p1, BIGNUM *q1,
                 BIGNUM *p1q1)
 {
+    BN_set_flags(lcm, BN_FLG_CONSTTIME);
+    BN_set_flags(gcd, BN_FLG_CONSTTIME);
+    BN_set_flags(p1, BN_FLG_CONSTTIME);
+    BN_set_flags(q1, BN_FLG_CONSTTIME);
+    BN_set_flags(p1q1, BN_FLG_CONSTTIME);
+
     return BN_sub(p1, p, BN_value_one())    /* p-1 */
            && BN_sub(q1, q, BN_value_one()) /* q-1 */
            && BN_mul(p1q1, p1, q1, ctx)     /* (p-1)(q-1) */
--- a/crypto/rsa/rsa_sp800_56b_check.c
+++ b/crypto/rsa/rsa_sp800_56b_check.c
@@ -38,6 +38,7 @@ int rsa_check_crt_components(const RSA *rsa, BN_CTX *ctx)
     p1 = BN_CTX_get(ctx);
     q1 = BN_CTX_get(ctx);
     if (q1 != NULL) {
+        BN_set_flags(r, BN_FLG_CONSTTIME);
         BN_set_flags(p1, BN_FLG_CONSTTIME);
         BN_set_flags(q1, BN_FLG_CONSTTIME);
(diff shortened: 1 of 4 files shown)


romen:
@cpereida @bbbrumley can I also ask for your feedback on the final version of this?

crypto/rsa/rsa_sp800_56b_check.c
```diff
@@ -172,7 +186,16 @@ int rsa_check_private_exponent(const RSA *rsa, int nbits, BN_CTX *ctx)
     lcm = BN_CTX_get(ctx);
     p1q1 = BN_CTX_get(ctx);
     gcd = BN_CTX_get(ctx);
-    ret = (gcd != NULL
+    if (gcd != NULL) {
+        BN_set_flags(p1, BN_FLG_CONSTTIME);
+        BN_set_flags(q1, BN_FLG_CONSTTIME);
+        BN_set_flags(lcm, BN_FLG_CONSTTIME);
```
romen, line 192: I missed p1q1 here!

romen: Pushed fixup commit: triggerflow did not detect this because I was anyway setting the `BN_FLG_CONSTTIME` flag twice (once where the temp values are allocated from the `BN_CTX` and once entering `rsa_get_lcm()`, which happened to be the first call).

Although it leads to more lines of code, I believe it is cleaner, less error prone and more future-proof, if we recommend that the "owner" of the variables sets the flag as soon as they are allocated, rather than relying on the fact that `rsa_get_lcm()` is currently the first function call that uses those `BN`.

The fixup commit does so and removes the second flag setting batch from within `rsa_get_lcm()` documenting in the function description that setting the flag is the caller's responsibility.

romen:
@slontis can you reapprove?

crypto/rsa/rsa_sp800_56b_check.c
slontis: fair enough - its only called in 2 places.

slontis: r doesnt seem to need it currently- would it potentially need it in the future? - (it has a expand in the call somewhere in the divide)

romen: `r`?

slontis: Line 37..

romen: Ah, I was looking at the wrong function!

Well, I am not sure about `r`, at the moment anyway the fact that `rsa->d` and `lcm` have the flag force things through the right paths.

I assumed (but without doing too much analysis) that given that `r` is the only value not being cleared it shouldn't ever contain confidential results (not even in the case `r != 1` that makes the validation fail.
If this is not the case, we should clear `r` as otherwise bug attacks could be exploited to try and recover information about the secret exponent or the primes.

romen: Also in https://github.com/openssl/openssl/blob/3f651554453fbfd10557d87a93e1a0c625b3fd3d/crypto/rsa/rsa_sp800_56b_check.c#L16-L76

`r` at the moment does not "need" to be flagged because the operands of the operations "returning" `r` are flagged.

`r` here also is not cleared, so I am assuming that even in the case of failed validation it or if 
any of
https://github.com/openssl/openssl/blob/3f651554453fbfd10557d87a93e1a0c625b3fd3d/crypto/rsa/rsa_sp800_56b_check.c#L64
https://github.com/openssl/openssl/blob/3f651554453fbfd10557d87a93e1a0c625b3fd3d/crypto/rsa/rsa_sp800_56b_check.c#L67
https://github.com/openssl/openssl/blob/3f651554453fbfd10557d87a93e1a0c625b3fd3d/crypto/rsa/rsa_sp800_56b_check.c#L70

fails, `r` won't contain sensitive data.

If this assumption is wrong we should make sure that `r` is cleared at the end of this function as well.

slontis: It would contain the result of (qInv . q) mod p.
I dont think it leaks anything - maybe we should just clear it anyway.

romen: Maybe @bbbrumley or @cpereida can double check the risks of not clearing `r` in these 2 functions (and if we should preemptively set the flag also on `r`).

crypto/rsa/rsa_sp800_56b_gen.c
```diff
@@ -353,6 +374,7 @@ int rsa_sp800_56b_pairwise_test(RSA *rsa, BN_CTX *ctx)
     k = BN_CTX_get(ctx);
     if (k == NULL)
         goto err;
+    BN_set_flags(k, BN_FLG_CONSTTIME);
```
mattcaswell, line 377: What about `tmp` here?

cpereida:
Checking through this, I noticed that the function `rsa_sp800_56b_check_keypair` is reachable through the wrapper https://github.com/openssl/openssl/blob/3f651554453fbfd10557d87a93e1a0c625b3fd3d/crypto/rsa/rsa_chk.c#L243
*only* when setting `FIPS_MODULE`.

I wonder if this should be made default following the spirit from #11765
If so, probably it requires its own PR but I thought it's worth mentioning in here.

crypto/rsa/rsa_sp800_56b_gen.c
cpereida: I quickly checked and it seems that functions using `tmp` do not check any `BN_FLG_CONSTTIME`, so I believe is ok.

crypto/rsa/rsa_sp800_56b_check.c
cpereida: If the cost of clearing it (and preemptively adding the flag) is not high, I'd recommend doing it so.

crypto/rsa/rsa_sp800_56b_gen.c
mattcaswell: Hmm....`tmp` is used in the call to `BN_mod_exp` below, which then calls `BN_mod_exp_mont`, which then calls `BN_mod_exp_mont_consttime` if the flag is set...or am I misreading the logic somewhere? In practice `rsa->d` should already have the flag set, so it should still go down that path...but still....

crypto/rsa/rsa_sp800_56b_check.c
romen: I tend to agree, I will push a fixup this evening

crypto/rsa/rsa_sp800_56b_gen.c
romen: I actually originally flagged tmp and later removed it:

From a logic perspective k and rsa->d here are the secrets, tmp is the public encrypted value and e and n are public.

Flagging tmp here could mask more serious mistakes from being detected by the CI, that was my rationale.

mattcaswell: Ok - makes sense.

romen:
> Checking through this, I noticed that the function `rsa_sp800_56b_check_keypair` is reachable through the wrapper
> 
> https://github.com/openssl/openssl/blob/3f651554453fbfd10557d87a93e1a0c625b3fd3d/crypto/rsa/rsa_chk.c#L243
> 
> 
> _only_ when setting `FIPS_MODULE`.
> I wonder if this should be made default following the spirit from #11765
> If so, probably it requires its own PR but I thought it's worth mentioning in here.

Let's see what @slontis says about it, I agree it should be a separate PR (maybe assigned to @slontis :stuck_out_tongue:)

crypto/rsa/rsa_sp800_56b_check.c
romen: See 31b9b166b9ee3e35bc309332eafb5bde0d714322

romen:
All feedback should have been addressed now.

slontis:
> I wonder if this should be made default following the spirit from #11765
> If so, probably it requires its own PR but I thought it's worth mentioning in here.

In fips mode it needs to only run what it runs now..
In NON fips mode do you want it to fail if it finds a multiprime key or a weak key? (That sounds like a breaking change?).

romen:
@mattcaswell can you give formal approval or dismiss your previous request for changes?
@slontis can you reapprove?

romen:
Merged to `master` as:

- d4bf0d57a8 Flag RSA secret BNs as consttime on keygen and checks


Thanks!

PR #12171, opened 2020-06-16
Fix test_ocsp with installed certificates

sebastianas:
First patch fixes the -no-CAstore for "openssl ocsp" (and "verify" while at it).
The second patch uses the new option in the test so it passes on system which has the certificates installed in the default path.

Sebastian


levitte:
Merged

00493490ddd824932d2019de4a73175cb50d7b95 APPS: Properly pass -no-CAstore
a4e440d6477b40b04ce825dae14c7f27778ca8f1 TEST: Pass -no-CAstore in 80-test_ocsp.t

PR #12159, opened 2020-06-15
Ensure we use the passed in libctx when loading EC private keys

mattcaswell:
If the public key is not present in an PKCS8 EC private key we can end up not using the libctx.

Fixes #12150

PR #12176, opened 2020-06-17
Add multiple fixes for ffc key generation using invalid p,q,g params

### LINKED ISSUE #11864: Openssl 3.0.0alpha3 fips provider: DSA key gen with some invalid p, q, g causes core dumps or hangs
Observed with 3.0.0alpha3 configured with NULL lib context using fips provider only. Observed with older (non-fetch) APIs.
Some validity checking of p, q, g seems to be missing because for selected values, core dumps or hangs occur. 

Consider valid parameters (pval, qval, gval). The following behavior was observed with the following expected negative cases:
- set (p, q, g) = (qval, qval, qval) => key gen hangs
- set (p, q, g) = (qval, gval, qval) => Core dump [ free(): invalid next size ]
- set (p, q, q) = (pval,, qval, qval) => Core dump [ free(): invalid pointer ]
- set (p, q, g) = (qval, pval, qval) => Core dump [ free(): invalid next size ]
- set (p, q, g) = (pval, gval, gval) => Core dump [ free(): invalid next size ]

```
#include <openssl/provider.h>
#include <openssl/crypto.h>
#include <openssl/evp.h>
#include <openssl/err.h>
#include <openssl/dsa.h>


{
   EVP_PKEY *pkey_parm = NULL;
   EVP_PKEY *pkey = NULL;
   DSA *dsa = NULL;
   EVP_PKEY_CTX *gen_ctx = NULL;
   int ret;
   BIGNUM *p = NULL;
   BIGNUM *q = NULL;
   BIGNUM *g = NULL;
   const char *pstr = "00F82CD0B121DF91E2F9D1A84A9A89402A40B9544184E1FDBD27B045D122D719BF1CB7188330EA0E866D3DD2E779C81146316D7280DA9E09FFEA58F4219484B8E7C606F8C6C15F5BD87C21730CC83484495EF991980DCE1D704C6FFB7330B691CCEE948F39935BDF4A1E6CEDAAA6EF37C83868EA0FFA529537384E3595D14F50FF044F9BA38CED5AB1B291D29C8DD2DA43C711E662666FE0C241835E2100C08210FBF0E180F7941CD12C8D98BE70CD68FCC7F57D40EB447D68BA269F6A36E6672D232B59077AD48933C95924E81C524775C7EB5E4D2996C21D7714DA89CF76C91C6E48E3678B80C75CE90437B3C8608886BB9595876C200CA77E554E6E0F724B39";
   const char *qstr = "00879D04B33B22C098583DE711AF3C6CEAB0BEBB0AAC1B5B5203154EEB";
   const char *gstr = "25895D1722207B2E06032D0269587DFDA581800D5510A5605888A7E9868BCFE625CFB6CFF9641AE18BDF0595CC3A7668F014D7E9A818006F7A6E63B1919A25E41389249F0880A968CB5E63714CA3B7CACFFB1C27BE121F7E4122FB711FCEA26F7FE2645799A9AF6007D00F846B04242A1A9664F084BD06762C66BF2BB1E42CFDF5CAE58BD4796272150A304115ACF499FACF41F57225CA6EEDFCB909F0331B9719E5B80F18399A677D0574BED3FDEA92BD05524B0FBCED902B73A203E26A864C99994B19B7C93959E58D5623480349B4468B47975C0F8676F05A429DFE31D7FB9A100F73C8B17C151391C63E814F93F7F249B7E861C7958DF56D021063DEA150";

    dsa = DSA_new();
    if (dsa == NULL) goto end;

    BN_hex2bn(&p, pstr);
    if (p == NULL) goto end;
    
    BN_hex2bn(&q, qstr);
    if (q == NULL) goto end;
    
    BN_hex2bn(&g, qstr);  /* changed from gstr to qstr => causes core dump */
    if (g == NULL) goto end;

    if (DSA_set0_pqg(dsa, p, q, g) <= 0)
        goto end;
    pkey_parm = EVP_PKEY_new();
    if (pkey_parm == NULL)
        goto end;
    if (EVP_PKEY_assign_DSA(pkey_parm, dsa) <= 0)
        goto end;
    
    printf("Generating DSA key \n");
    gen_ctx = EVP_PKEY_CTX_new(pkey_parm, NULL);   
    printf("... done generating DSA key\n");

end:
   printf ("err:\n");
   ERR_print_errors_fp(stderr);
   EVP_PKEY_free(pkey_parm);
   
   EVP_PKEY_free(pkey);
   EVP_PKEY_CTX_free(gen_ctx);
   return 1;
}
```
Output received:
```
default provider available: 0
fips provider available: 1
Generating DSA key 
free(): invalid pointer
Aborted (core dumped)
```

slontis:
Fixes #11864

- The dsa keygen assumed valid p, q, g values were being passed. If this is not correct then it is
possible that dsa keygen can either hang or segfault.
The fix was to do a partial validation of p, q, and g inside the keygen.
- Fixed a potential double free in the dsa keypair test in the case when in failed (It should never fail!).
It freed internal object members without setting them to NULL.
- Changed the FFC key validation to accept 1024 bit keys in non fips mode.
- Added tests that use the fips provider to test these cases.

##### Checklist

### DIFF
--- a/crypto/dh/dh_key.c
+++ b/crypto/dh/dh_key.c
@@ -287,6 +287,10 @@ static int generate_key(DH *dh)
             } else
 #endif
             {
+                /* Do a partial check for invalid p, q, g */
+                if (!ffc_params_simple_validate(dh->libctx, &dh->params,
+                                                FFC_PARAM_TYPE_DH))
+                    goto err;
                 /*
                  * For FFC FIPS 186-4 keygen
                  * security strength s = 112,
--- a/crypto/dsa/dsa_key.c
+++ b/crypto/dsa/dsa_key.c
@@ -74,6 +74,11 @@ static int dsa_keygen(DSA *dsa, int pairwise_test)
         priv_key = dsa->priv_key;
     }
 
+    /* Do a partial check for invalid p, q, g */
+    if (!ffc_params_simple_validate(dsa->libctx, &dsa->params,
+                                    FFC_PARAM_TYPE_DSA))
+        goto err;
+
     /*
      * For FFC FIPS 186-4 keygen
      * security strength s = 112,
@@ -110,6 +115,8 @@ static int dsa_keygen(DSA *dsa, int pairwise_test)
         if (!ok) {
             BN_free(dsa->pub_key);
             BN_clear_free(dsa->priv_key);
+            dsa->pub_key = NULL;
+            dsa->priv_key = NULL;
             BN_CTX_free(ctx);
             return ok;
         }
--- a/crypto/ffc/ffc_params_generate.c
+++ b/crypto/ffc/ffc_params_generate.c
@@ -39,6 +39,11 @@
  */
 static int ffc_validate_LN(size_t L, size_t N, int type)
 {
+#ifndef FIPS_MODULE
+    if (L == 1024 && N == 160)
+        return 80;
+#endif
+
     if (type == FFC_PARAM_TYPE_DH) {
         /* Valid DH L,N parameters from SP800-56Ar3 5.5.1 Table 1 */
         if (L == 2048 && (N == 224 || N == 256))
@@ -498,6 +503,7 @@ int ffc_params_FIPS186_4_gen_verify(OPENSSL_CTX *libctx, FFC_PARAMS *params,
     EVP_MD *md = NULL;
     int verify = (mode == FFC_PARAM_MODE_VERIFY);
     unsigned int flags = verify ? params->flags : 0;
+    const char *def_name;
 
     *res = 0;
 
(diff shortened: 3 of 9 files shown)


test/recipes/30-test_libctx.t
```diff
@@ -0,0 +1,46 @@
+#! /usr/bin/env perl
```
t8m, line 1: I am confused by the test name here. I suppose this is meant to be more general test in future but wouldn't it be somewhat duplicate with the evp_extra_test2 from #12159 ?

slontis: This one will test both fips and non fips modes..
Unravelling the mess in evp_extra_test & evp_extra_test2 was outside the scope of what I am trying to do here.

slontis: Another PR will need to resolve this..
Some of the tests in evp_extra will need to run in non fips mode only.

t8m: Could you please at least rename the test to something like test_evp_libctx or something like that so it is clear what the intention of the test is?

slontis: @t8m Updated the test name and added description in the file.

beldmit:
These changes look good to me per se.

slontis:
Thanks. merged to master.

PR #12177, opened 2020-06-17
Fix potential double free in rsa_keygen pairwise test.

slontis:
It should never hit this branch of code, so there is no feasible test.
Found due to a similar issue in PR #12176.

##### Checklist

### DIFF
--- a/crypto/rsa/rsa_gen.c
+++ b/crypto/rsa/rsa_gen.c
@@ -445,6 +445,12 @@ static int rsa_keygen(OPENSSL_CTX *libctx, RSA *rsa, int bits, int primes,
             BN_clear_free(rsa->dmp1);
             BN_clear_free(rsa->dmq1);
             BN_clear_free(rsa->iqmp);
+            rsa->d = NULL;
+            rsa->p = NULL;
+            rsa->q = NULL;
+            rsa->dmp1 = NULL;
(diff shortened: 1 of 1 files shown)


crypto/rsa/rsa_gen.c
```diff
@@ -445,6 +445,12 @@ static int rsa_keygen(OPENSSL_CTX *libctx, RSA *rsa, int bits, int primes,
             BN_clear_free(rsa->dmp1);
             BN_clear_free(rsa->dmq1);
             BN_clear_free(rsa->iqmp);
+            rsa->d = NULL;
+            rsa->p = NULL;
+            rsa->q = NULL;
+            rsa->dmp1 = NULL;
+            rsa->dmq1 = NULL;
+            rsa->iqmp = NULL;
```
InfoHunter, line 453: Why are these fields called 'intermediate results'?

slontis: I think the term is applicable to these values in this context.. For this operation sensitive values must be cleared on failure. (This term comes from the specs somewhere)
Normally we free the RSA object when doing this type of operation.

levitte:
Merged

7905806c94b17b1907d5352ceb047dd8d859288c Fix potential double free in rsa_keygen pairwise test.

PR #12179, opened 2020-06-17
Fix the s_server -dtls option

mattcaswell:
DTLS in s_server was failing due to problems with the cookie generate callback which was recently converted to use EVP_MAC. The new code was failing to return the length of the generated cookie, and so libssl aborted the connection.

While I was at this fix I also corrected the value of the DTLS1_COOKIE_LENGTH value in the public header. This was known to be off-by-one, but could not be fixed due to ABI compat concerns. Since 3.0 is not ABI compatible we can fix it.


mattcaswell:
Pushed,  including the suggested tweak. Thanks.

PR #12180, opened 2020-06-17
Fix some SSL_dup issues and provide better documentation for it

mattcaswell:
SSL_dup() is a very strange function and it doesn't behave the way one might expect it to, and not in accordance with the man page. There are two primary code paths:

1) If the SSL object is in a state where the first handshake has started or completed then we up ref the SSL and return back a reference to *the same* SSL, i.e. it does *not*  do a dup.

2) If the SSL object has not yet started the first handshake then
    - We create a new SSL object from the same SSL_CTX
    - We copy some selected settings from the src SSL object to the dest SSL object

Probably when SSL_dup was first written it dup'd all of the setting in the SSL object. But it doesn't do that any more and hasn't done so for a very long time. For example a selection of settings that it doesn't "dup" include:

tlsext_host_name
tlsext_debug_callback
tlsext_status_type
mtu
psk_client_callback
psk_server_callback
psk_find_session_callback
psk_use_session_callback
max_early_data
sigalgs
client_sigalgs
min_proto_version
max_proto_version
groups
max_pipelines
...
(this is just a small selection the list goes on...)

Keeping SSL_dup maintained and current is a hard job. In practice I believe this function to be rarely used. IMO the best way forward is to simply document what it does do and not attempt to  fix all the various settings.

Potentially we might consider deprecating it entirely. However I have not taken that step in this PR, as I thought that might be controversial. That could be a follow up PR.

One setting that I have fixed is that for min_proto_version/max_proto_version. This probably really should be "dup'd". Thanks to Rebekah Johnson for reporting this issue.

While looking into this problem I realised that the current BIO handling in SSL_dup is broken to the extent that if you call SSL_dup before a handshake has started, but after you have set BIOs on your SSL object, then it is likely to crash, or at least fail in some unexpected way:

Firstly the SSL_dup code was passing a BIO ** as the destination argument for BIO_dup_state. However BIO_dup_state expects a BIO * for that parameter. Any attempt to use this will either (1) fail silently, (2) crash or fail in some other strange way.

Secondly many BIOs do not implement the BIO_CTRL_DUP ctrl required to make this work.

Thirdly, if rbio == wbio in the original SSL object, then an attempt is made to up-ref the BIO in the new SSL object - even though it hasn't been set yet and is NULL. This results in a crash.

This appears to have been broken for a very long time with at least some of the problems described above coming from SSLeay. The simplest approach is to just remove this BIO dup capability from the function.

Once this is approved, I will create a 1.1.1 backport PR.

### DIFF
--- a/doc/man3/SSL_new.pod
+++ b/doc/man3/SSL_new.pod
@@ -73,9 +73,6 @@ L<SSL_set_info_callback(3)>
 
 =item any configured Cipher List
 
-=item any BIOs configured on I<s> will have new BIO's created and the BIO state
-duplicated via BIO_dup_state().
-
 =item initial accept (server) or connect (client) state
 
 =item the max cert list value set via L<SSL_set_max_cert_list(3)>
--- a/doc/man3/SSL_new.pod
+++ b/doc/man3/SSL_new.pod
@@ -82,7 +82,7 @@ L<SSL_set_info_callback(3)>
 =item application specific data set via L<SSL_set_ex_data(3)>
 
 =item any CA list or client CA list set via L<SSL_set0_CA_list(3)>,
-L<SSL_set0_client_CA_list(3)> or similar functions
+SSL_set0_client_CA_list() or similar functions
 
 =item any security level settings or callbacks
 
--- a/doc/man3/SSL_new.pod
+++ b/doc/man3/SSL_new.pod
@@ -26,16 +26,9 @@ structure are freed.
 SSL_up_ref() increments the reference count for an
 existing B<SSL> structure.
 
-SSL_dup() duplicates an existing B<SSL> structure into a new allocated one
-or just increments the reference count if the connection is active. All
-settings are inherited from the original B<SSL> structure. Dynamic data (i.e.
-existing connection details) are not copied, the new B<SSL> is set into an
-initial accept (server) or connect (client) state.
-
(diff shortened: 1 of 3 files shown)


mattcaswell:
Travis eventually ran and reported a doc-nits issue. I've rebased this to resolve a conflict with master, and pushed a fixup to address the travis failure.

@t8m please reconfirm?

doc/man3/SSL_new.pod
```diff
@@ -26,16 +26,9 @@ structure are freed.
 SSL_up_ref() increments the reference count for an
 existing B<SSL> structure.
 
-SSL_dup() duplicates an existing B<SSL> structure into a new allocated one
-or just increments the reference count if the connection is active. All
```
kaduk, line 30: We no longer document the "just increments the reference count" behavior but it is still there?

mattcaswell: It is documented. That aspect was duplicated:

"For connections that are not in their initial state SSL_dup() just increments an internal
reference count and returns the I<same> handle."

kaduk: Oops, I guess I missed that.  Thanks for the clarification, and sorry for the noise.

mattcaswell:
Pushed to master. Backport to 1.1.1 in #12245.

PR #12181, opened 2020-06-18
namemap: change ossl_namemap_empty() to do what the documentation says.

paulidale:
The `ossl_namemap_empty()` function is documented as returning 1 when passed a NULL argument.  Instead it core dumps.  Added a unit test for this and of behaviours of `ossl_namemap_empty()`.

Additionally, a performance improvement is incorporated.  The namemap
max_number field is only ever compared against zero and incremented.
The zero comparison grabs a lock specifically for this check.  This change
uses TSAN operations instead if they are available.

### DIFF
--- a/crypto/core_namemap.c
+++ b/crypto/core_namemap.c
@@ -11,6 +11,7 @@
 #include "internal/namemap.h"
 #include <openssl/lhash.h>
 #include "crypto/lhash.h"      /* openssl_lh_strcasehash */
+#include "internal/tsan_assist.h"
 
 /*-
  * The namenum entry
@@ -34,7 +35,12 @@ struct ossl_namemap_st {
 
     CRYPTO_RWLOCK *lock;
     LHASH_OF(NAMENUM_ENTRY) *namenum;  /* Name->number mapping */
-    int max_number;                    /* Current max number */
+
+#ifdef tsan_ld_acq
+    TSAN_QUALIFIER int max_number;     /* Current max number TSAN version */
+#else
+    int max_number;                    /* Current max number plain version */
+#endif
 };
 
 /* LHASH callbacks */
@@ -91,14 +97,21 @@ static const OPENSSL_CTX_METHOD stored_namemap_method = {
 
 int ossl_namemap_empty(OSSL_NAMEMAP *namemap)
 {
-    int rv = 0;
+#ifdef tsan_ld_acq
+    /* Have TSAN support */
+    return namemap == NULL || tsan_load(&namemap->max_number) == 0;
+#else
+    /* No TSAN support */
+    int rv;
+
+    if (namemap == NULL)
+        return 1;
 
     CRYPTO_THREAD_read_lock(namemap->lock);
-    if (namemap->max_number == 0)
-        rv = 1;
+    rv = namemap->max_number == 0;
     CRYPTO_THREAD_unlock(namemap->lock);
-
     return rv;
+#endif
 }
 
 typedef struct doall_names_data_st {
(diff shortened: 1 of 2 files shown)


paulidale:
An alternative would be to make this function static and remove it from the (internal) documentation.  I'm happy to change it over.

crypto/core_namemap.c
```diff
@@ -216,7 +229,7 @@ int ossl_namemap_add_name_n(OSSL_NAMEMAP *namemap, int number,
         goto err;
 
     namenum->number = tmp_number =
-        number != 0 ? number : ++namemap->max_number;
+        number != 0 ? number : 1 + tsan_counter(&namemap->max_number);
```
levitte, line 232: Side note: I had to look up what `tsan_counter` does...  the name is not at all obvious!  Could it be renamed to `tsan_incr`?  In another PR, of course.

crypto/core_namemap.c
```diff
@@ -91,14 +97,21 @@ static const OPENSSL_CTX_METHOD stored_namemap_method = {
 
 int ossl_namemap_empty(OSSL_NAMEMAP *namemap)
 {
-    int rv = 0;
+#ifdef tsan_ld_acq
+    /* Have TSAN support */
+    return namemap == NULL || tsan_load(&namemap->max_number) == 0;
+#else
+    /* No TSAN support */
```
levitte, line 104: So this means we still have to be careful about possible locking issues, but that we might not catch that issue because most of us (as well as Travis) run our tests in an environment where we have access to atomics and thereby have "proper" TSAN support...  While I like this change, that's something that we should be *very* clear about.

crypto/core_namemap.c
paulidale: Yes, only one increment is possible.  Pre-increment isn't available at this point, we rely on the compiler recognising it.

crypto/core_namemap.c
paulidale: Seriously?????????????  You were the person who pointed out this exact problem in my previous PR.  We've used TSAN primitives without locking for a while now.  They aren't safe.

As a solution, I think we should wrapper up the TSAN code (in a struct) so that it locks if the TSAN calls are not available.  Currently, we use TSAN for _int_, _unsigned int_ and _unsigned long int_.  The final flavour will disappear with @richsalz's hash changes.  Still, this needs to be sorted out before 3.0.

crypto/core_namemap.c
levitte: Uhmm, I only asked for a rename.  I understand that it's going to be behave like a post-increment, just like `tsan_decr` behaves like a post-decrement.

crypto/core_namemap.c
levitte: I haven't been aware of TSAN before I saw your changes in that other PR.  Looking at the TSAN macros now, I find it striking how much (not exactly, but very much) they do the same thing as the CRYPTO_UP_REF / CRYPTO_DOWN_REF functions.  As a matter of fact, we could easily reimplement much of the latter in TSAN terms.

The fact remains, though, that the more we let locking mechanisms be for "legacy" platforms only, the more we risk getting bug reports that we have a hard time debugging.  We might want to consider a knob that turns off our use of atomics for that very purpose.

richsalz:
BTW, similar tsan optimization for LHASH is in #10632

crypto/core_namemap.c
paulidale: The legacy locking is the reason I'd like to wrap these up more nicely.  A user (i.e. us) shouldn't care if locks need to be used or not.

paulidale:
#10632 doesn't provide an alternative locking path for systems that don't have tsan support.  Unsigned longs are going to be _non-atomic_ on many platforms.  Fortunately, I don't think it matters for lhash statistics collection or I'd have raised it in the PR.

crypto/core_namemap.c
paulidale: Another PR for the rename -- perhaps as part of the rationalisation and wrapper I'd like to get done.

paulidale:
Ping @openssl/committers

paulidale:
Merged to master, thanks.

PR #11996, opened 2020-05-30
kdf: make function naming consistent.

paulidale:
The EVP_KDF_CTX functions have been renamed to bring consistency to the KDF naming.

Before 3.0 is released, the names are mutable and this prevents more inconsistencies being introduced.

There are no functional or code changes.  Just the renaming and a little reformatting.

- [x] documentation is added or updated
- [x] tests are added or updated

PR #12186, opened 2020-06-18
Revert KDF/MAC renaming

mattcaswell:
This reverts commits d9c2fd51e and 765d04c9.

Those commits made the following changes (amongst other similar ones) to the MAC APIs (and similar ones to the KDF APIs):

EVP_MAC_CTX_new -> EVP_MAC_new_ctx
EVP_MAC_CTX_free -> EVP_MAC_free_ctx

In master at the moment I count 26 occurrences of the FOO_CTX_new pattern, and 2 of the FFO_new_ctx pattern (the 2 recently introduced ones that would be reverted by this commit). This is similarly true with the "free" versions.

Those commits do *not* make things *more* consistent they do the opposite. We should not be breaking this well established pattern. This is a recipe for introducing more confusion for our users not less.


paulidale:
As an aside, there was a discussion about the name changes.

Agreed, these changes make the function names different to the rest of the APIs.  That doesn't make the rest of the APIs good.....

paulidale:
I'm not going to block the reversion.  I'm ambivalent as to the changes.

mattcaswell:
> Agreed, these changes make the function names different to the rest of the APIs. That doesn't make the rest of the APIs good.....

"We wouldn't start from here" is something I could say about lots of the OpenSSL API. But its what we have. Trying to change it now will be a source of HUGE confusion.

levitte:
I'm actually opposed to this reversal. The dispute about what functions should be prefixed WITH FOO_CTX and which ones should be prefixed with FOO has felt endless, and the prefixing of certain getters have been a gray area with the same recurring question of how they should be prefixed (answer: "weeeell, that depends", which I can understand doesn't help).

Having the same prefix on all related functions does clear quite a lot of that up. 

(mind you, I was originally opposed to changing the old pattern at all... it's been a long process)

mattcaswell:
> Having the same prefix on all related functions does clear quite a lot of that up.

But it does so entirely in isolation just for these specific APIs. We have not agreed any strategy for some kind of wholesale renaming of the APIs (and I would argue against that). Whilst the current naming convention has its quirks it is a least broadly consistent across types with how we construct and destruct ctx objects. I see this as breaking what consistency we do have. Far from clearing things up it make things much more murky IMO.

*If* we want to change the naming conventions then we should do so only after agreeing what those conventions should be, and agreeing a strategy for how we are going to apply them. It should *not* be done piecemeal (and hidden away in a PR that supposedly made things more consistent but in reality did the opposite).

t8m:
I believe that if we do not apply this reversal, we should also create new name aliases for the existing EVP_MD_CTX, EVP_CIPHER_CTX and similar functions, that would go along with the new naming convention. This would allow everyone following the new naming conventions to do it consistently across the objects. But of course they would not be used much from start by applications as they usually have to work with both 1.1.1 and 3.0 APIs.

levitte:
> We have not agreed any strategy for some kind of wholesale renaming of the APIs

Although, naming strategies have been brought up before...  and received disappointingly litle attention.  The latest (and most open, most discussions have been informal) I could find is https://mta.openssl.org/pipermail/openssl-project/2019-March/001280.html
So I guess that #11997 and #11996 got some much needed attention to the matter :wink:

> *If* we want to change the naming conventions then we should do so only after agreeing what those conventions should be, and agreeing a strategy for how we are going to apply them. It should *not* be done piecemeal (and hidden away in a PR that supposedly made things more consistent but in reality did the opposite).

So let's discuss and see if we can find agreement.  At this point, I would like to see that discussion and conclusion happen *before* this reversal, lest we want to see a reversal possibly followed by a counter reversal.

levitte:
> I believe that if we do not apply this reversal, we should also create new name aliases for the existing EVP_MD_CTX, EVP_CIPHER_CTX and similar functions, that would go along with the new naming convention. This would allow everyone following the new naming conventions to do it consistently across the objects. But of course they would not be used much from start by applications as they usually have to work with both 1.1.1 and 3.0 APIs.

That shouldn't even be hard.  @paulidale has described #11996 and #11997 as fairly simple sed jobs (plus, I assume, a few last nudges), it shouldn't be too hard to replicate that for the others.

(note, though, that specifically for ciphers and digests, I would like to get away from CamelCase while we're at it)

levitte:
I've put a hold on this, because I believe the OTC needs to discuss this first, and to avoid the possibility of a reversal followed by a counter reversal.

levitte:
Had @mattcaswell noticed #11996 and #11997 earlier, I assume we would have seen an OTC hold on those, and that would have been fine.  I'm only acting on the current situation as it is.

h-vetinari:
> @t8m: I believe that if we do not apply this reversal, we should also create new name aliases for the existing EVP_MD_CTX, EVP_CIPHER_CTX and similar functions, that would go along with the new naming convention.

Sounds like a major version bump would be a good time to do this...? :wink:

levitte:
After much deliberation, we've concluded that this new pattern was too early and not thought through enough.  See https://mta.openssl.org/pipermail/openssl-project/2020-July/002093.html

I'm lifting the hold

levitte:
This needs to be rebased, unfortunately

levitte:
I'll have another look tomorrow morning.  My brain is too fried to deal with this right now.  It would, after all, be better if the "new scheme" didn't slip out in the upcoming alpha just to be renamed back in the one after...

levitte:
It turned out not being terribly difficult to resolve the conflicts

levitte:
Argh...  unfortunately, the new names proliferated elsewhere...

```
gcc  -I. -Icrypto -Iinclude -Iproviders/implementations/include -Iproviders/common/include -I../master -I../master/crypto -I../master/include -I../master/providers/implementations/include -I../master/providers/common/include  -fPIC -pthread -m64 -fno-omit-frame-pointer -g -Wa,--noexecstack -Wall -O0 -g -DDEBUG_UNUSED -DPEDANTIC -pedantic -Wno-long-long -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -Wswitch -Wsign-compare -Wshadow -Wformat -Wtype-limits -Wundef -Werror -Wmissing-prototypes -Wstrict-prototypes -fsanitize=address -DOPENSSL_USE_NODELETE -DL_ENDIAN -DOPENSSL_PIC -DOPENSSLDIR="\"/usr/local/ssl\"" -DENGINESDIR="\"/usr/local/lib/engines-3\"" -DMODULESDIR="\"/usr/local/lib/ossl-modules\"" -DOPENSSL_BUILDING_OPENSSL  -MMD -MF providers/implementations/rands/libnonfips-lib-drbg_hmac.d.tmp -MT providers/implementations/rands/libnonfips-lib-drbg_hmac.o -c -o providers/implementations/rands/libnonfips-lib-drbg_hmac.o ../master/providers/implementations/rands/drbg_hmac.c
../master/providers/implementations/rands/drbg_hmac.c: In function 'do_hmac':
../master/providers/implementations/rands/drbg_hmac.c:67:10: error: implicit declaration of function 'EVP_MAC_set_ctx_params'; did you mean 'EVP_RAND_set_ctx_params'? [-Werror=implicit-function-declaration]
   67 |     if (!EVP_MAC_set_ctx_params(ctx, params)
      |          ^~~~~~~~~~~~~~~~~~~~~~
      |          EVP_RAND_set_ctx_params
../master/providers/implementations/rands/drbg_hmac.c: In function 'drbg_hmac_free':
../master/providers/implementations/rands/drbg_hmac.c:318:9: error: implicit declaration of function 'EVP_MAC_free_ctx'; did you mean 'EVP_MAC_free'? [-Werror=implicit-function-declaration]
  318 |         EVP_MAC_free_ctx(hmac->ctx);
      |         ^~~~~~~~~~~~~~~~
      |         EVP_MAC_free
```

The fix isn't terribly difficult:

``` diff
diff --git a/providers/implementations/rands/drbg_hmac.c b/providers/implementations/rands/drbg_hmac.c
index b73fe958b0..7ddfae1568 100644
--- a/providers/implementations/rands/drbg_hmac.c
+++ b/providers/implementations/rands/drbg_hmac.c
@@ -64,7 +64,7 @@ static int do_hmac(PROV_DRBG_HMAC *hmac, unsigned char inbyte,
 
     *params = OSSL_PARAM_construct_octet_string(OSSL_MAC_PARAM_KEY, hmac->K,
                                                 hmac->blocklen);
-    if (!EVP_MAC_set_ctx_params(ctx, params)
+    if (!EVP_MAC_CTX_set_params(ctx, params)
             || !EVP_MAC_init(ctx)
             /* K = HMAC(K, V || inbyte || [in1] || [in2] || [in3]) */
             || !EVP_MAC_update(ctx, hmac->V, hmac->blocklen)
@@ -78,7 +78,7 @@ static int do_hmac(PROV_DRBG_HMAC *hmac, unsigned char inbyte,
    /* V = HMAC(K, V) */
     *params = OSSL_PARAM_construct_octet_string(OSSL_MAC_PARAM_KEY, hmac->K,
                                                 hmac->blocklen);
-    return EVP_MAC_set_ctx_params(ctx, params)
+    return EVP_MAC_CTX_set_params(ctx, params)
            && EVP_MAC_init(ctx)
            && EVP_MAC_update(ctx, hmac->V, hmac->blocklen)
            && EVP_MAC_final(ctx, hmac->V, NULL, sizeof(hmac->V));
@@ -220,7 +220,7 @@ static int drbg_hmac_generate(PROV_DRBG *drbg,
     for (;;) {
         *params = OSSL_PARAM_construct_octet_string(OSSL_MAC_PARAM_KEY,
                                                     hmac->K, hmac->blocklen);
-        if (!EVP_MAC_set_ctx_params(ctx, params)
+        if (!EVP_MAC_CTX_set_params(ctx, params)
             || !EVP_MAC_init(ctx)
             || !EVP_MAC_update(ctx, temp, hmac->blocklen))
             return 0;
@@ -315,7 +315,7 @@ static void drbg_hmac_free(void *vdrbg)
     PROV_DRBG_HMAC *hmac;
 
     if (drbg != NULL && (hmac = (PROV_DRBG_HMAC *)drbg->data) != NULL) {
-        EVP_MAC_free_ctx(hmac->ctx);
+        EVP_MAC_CTX_free(hmac->ctx);
         ossl_prov_digest_reset(&hmac->digest);
         OPENSSL_secure_clear_free(hmac, sizeof(*hmac));
     }
```

... but it demands re-approval.

I've rebased this PR and added a commit for the above diff

levitte:
I would *like* to see this get merged urgently, as we're releasing the next alpha in less than 24 hours.  If you agree, set the "urgent" label.

levitte:
Ping?

t8m:
I agree with it being urgent. We should not keep dead-end API names public in another alpha.

t8m:
One more thing - to avoid bad surprises, please amend the last commit with `[extended tests]`.

We should get extended Travis run passed before merging this.

beldmit:
I definitely know the gost-engine build will fail after the rollback of the renaming. But we have a kind of catch-22 here...

t8m:
Yeah, of course the external tests that depend on the new names might fail, that would be an expected failure.

levitte:
> One more thing - to avoid bad surprises, please amend the last commit with `[extended tests]`.

Done...  but I understand correctly, that we will ignore name failures in external tests, right?

t8m:
Yes, at least for the gost test, for sure.

levitte:
Of course, both Travis and Appveyor have quite a backlog for us...  timing is of the essence, eh?

t8m:
Yeah, :disappointed:

levitte:
Appveyor has come through now, and seems to be green across all builds so far.

levitte:
I've run a few tests on my machine, all successful.  Quite frankly, I don't think that the external tests should matter for this PR, as we haven't changed any *functionality* in this PR, only the names.  To my judgement, it's hard to see how this would cause other surprises...  So unless you tell me not to, I will merge this, based on Appveyor success alone if need be.

levitte:
@romen also performed some test and found it passing, so...

Merged

865adf97c9b8271788ee7293ecde9e8a643a1c45 Revert "The EVP_MAC functions have been renamed for consistency.  The EVP_MAC_CTX_*"
660c534435e238c6bd8065c1d544a1c4d3c555a3 Revert "kdf: make function naming consistent."
e4162f86d7fd92058e5558bd81ce9dbc615fec3f DRBG: Fix the renamed functions after the EVP_MAC name reversal

PR #12187, opened 2020-06-18
Fix wrong return value check of mmap function

trisbauer:
The mmap function never returns NULL. If an error occurs, the function returns MAP_FAILED.

CLA: trivial

### DIFF
--- a/crypto/mem_sec.c
+++ b/crypto/mem_sec.c
@@ -514,7 +514,7 @@ static void sh_done(void)
     OPENSSL_free(sh.freelist);
     OPENSSL_free(sh.bittable);
     OPENSSL_free(sh.bitmalloc);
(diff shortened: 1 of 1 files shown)


t8m:
> The mmap function never returns NULL. If an error occurs, the function returns MAP_FAILED.

Yes, but the sh_done must also handle situation when mmap was not called yet and in that case the value is NULL.

t8m:
> > The mmap function never returns NULL. If an error occurs, the function returns MAP_FAILED.
> 
> Yes, but the sh_done must also handle situation when mmap was not called yet and in that case the value is NULL.

To correct myself - that should be handled already by the test for sh.map_size.

paulidale:
Agreed, trivial.

paulidale:
Merged to master and 1.1.1.  Thanks for the submission.

PR #12206, opened 2020-06-19
Cms print

beldmit:
cms -cmsout -print currently has no option to print strings in human-readable representation.
This PR adds `-nameopt` option for this purpose.

##### Checklist

### DIFF
--- a/apps/cms.c
+++ b/apps/cms.c
@@ -1168,7 +1168,7 @@ int cms_main(int argc, char **argv)
                         ASN1_PCTX_set_nm_flags(pctx, get_nameopt());
                     }
                 }
-                 CMS_ContentInfo_print_ctx(out, cms, 0, pctx);
+                CMS_ContentInfo_print_ctx(out, cms, 0, pctx);
                 ASN1_PCTX_free(pctx);
             }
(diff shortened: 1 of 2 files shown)


beldmit:
BTW, if it is possible to print the certificates included in the CMS structure using `X509_print_ex`, it would be much better...

apps/cms.c
```diff
@@ -1152,8 +1158,19 @@ int cms_main(int argc, char **argv)
         }
     } else {
         if (noout) {
-            if (print)
-                CMS_ContentInfo_print_ctx(out, cms, 0, NULL);
+            if (print) {
+                ASN1_PCTX *pctx = NULL;
+                if (get_nameopt() != XN_FLAG_ONELINE) {
+                    pctx = ASN1_PCTX_new();
+                    if (pctx) { /* Print anyway if malloc failed */
```
paulidale, line 1165: Missing `!= NULL`

beldmit: Fixed. Thanks!

Will squash fixups on merging.

beldmit:
Squashed and merged. Thanks!

PR #12212, opened 2020-06-20
Windows build: (1) string pooling only for release (2) default stack …

ideafarm:
…probes

Without these changes, Visual Studio's debugger cannot be used to step through code for "openssl s_server -WWW".  These changes fix that.  With string pooling, code is so optimized that execution stepping is confusing and ineffective.  The use of /Gs0 is contrary to Microsoft documentation, which states that nondefault stack probing should only be used if you know exactly why you need to do it.  Worse, /Gs0 appears to break Visual Studio's debugger.  (The workaround is to use the disassembler window and step into and through _chkstk.)

### DIFF
--- a/Configurations/10-main.conf
+++ b/Configurations/10-main.conf
@@ -1293,8 +1293,8 @@ my %targets = (
         inherit_from     => [ "VC-common" ],
         template         => 1,
         CFLAGS           => add(picker(debug   => '/Od',
-                                       release => '/O2')),
-        cflags           => add(picker(default => '/Gs0 /GF /Gy',
+                                       release => '/O2 /GF')),
+        cflags           => add(picker(default => ' /Gy',
                                        debug   =>
                                        sub {
                                            ($disabled{shared} ? "" : "/MDd");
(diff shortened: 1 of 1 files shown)


Configurations/10-main.conf
```diff
@@ -1293,8 +1293,8 @@ my %targets = (
         inherit_from     => [ "VC-common" ],
         template         => 1,
         CFLAGS           => add(picker(debug   => '/Od',
-                                       release => '/O2')),
-        cflags           => add(picker(default => '/Gs0 /GF /Gy',
+                                       release => '/O2 /GF')),
```
makr, line 1296: I think this is unnecessary as /O2 enables /GF anyway, [see here](https://docs.microsoft.com/en-us/cpp/build/reference/gf-eliminate-duplicate-strings?view=vs-2019).

levitte:
I guess the most important modification is `/Gs0`...  it's something that came along a long time ago, I found it as far back as the first import of SSLeay to become OpenSSL, back in December 1998...  so the reason for it, as far as we're concerned, is lost in history.

Re `/GF`, I'm amazed that it trips the debugger...  after all, it only means that any address to the exact same string copy will point at the exactly same place in memory instead of others...  but ok, if that trips the debugger, Microsoft must do some funnies that we simply don't fathom.

Configurations/10-main.conf
ideafarm: I agree so would eliminate /GF altogether, since it is implied by /O2.

ideafarm:
Perhaps there is agreement.  If so, the only remaining issue is the cla-check.  I've submitted my CLA twice, with no effect, even though I received a reply that seemed to indicate that it was accepted.

ideafarm:
I am going to close this PR and submit a new one that eliminates /GF.  Perhaps the CLA will be effective now.

levitte:
@ideafarm, your CLA contains a different address than the one you use in your commits.  The one used in your commits seems to be typical of commits done directly in github.  That's why the CLA checker fails.

ideafarm:
@levitte Thanks.  I'll take care of that before opening a PR revised per this conversation.

PR #12222, opened 2020-06-21
Make the naming scheme for dispatched functions more consistent

mspncp:
While digging through the fetchable RAND implementation (#11682) it occurred to me (not for the first time) that the current naming conventions for the cryptographic operations and their dispatched functions exhibit some inconsistensies and I found the scheme a bit unintuitive and hard to remember. The FIPS core developers of you might think differently about it, but maybe it's just because you have got accustomed to the current scheme over the time.

After thinking about it for a while, I came up with the following proposal for straightening out the naming scheme. I am aware that my proposal comes late, but it's not too late before the final freeze. And OTOH, in the middle of the heavy replumbing it would not have been easy to make such a change either. So please give it at least some sympathetic consideration, @openssl/otc.

```
Make the naming scheme for dispatched functions more consistent

The new naming scheme consistently usese the `OSSL_FUNC_` prefix for all
functions which are dispatched between the core and providers.

This change includes in particular all up- and downcalls, i.e., the
dispatched functions passed from core to provider and vice versa.

- OSSL_core_  -> OSSL_FUNC_core_
- OSSL_provider_ -> OSSL_FUNC_core_

For operations and their function dispatch tables, the following convention
is used:

  Type                 | Name (evp_generic_fetch(3))       |
  ---------------------|-----------------------------------|
  operation            | OSSL_OP_FOO                       |
  function id          | OSSL_FUNC_FOO_FUNCTION_NAME       |
  function "name"      | OSSL_FUNC_foo_function_name       |
  function typedef     | OSSL_FUNC_foo_function_name_fn    |
  function ptr getter  | OSSL_FUNC_foo_function_name       |
```

In the following, I'll try to give some more explanations

### Current naming convention

The inconsistency I am talking about becomes evident if you compare the current documentation on master (evp_generic_fetch(3)) with the way how it's currently implemented:

Type                 | Documentation (evp_generic_fetch(3)) |  Implementation (example)       |
---------------------|--------------------------------------|---------------------------------|
operation id         | OSSL_OP_FOO                          | OSSL_OP_SIGNATURE               |
function id          | OSSL_OP_FOO_FUNCTION_NAME            | OSSL_FUNC_SIGNATURE_SIGN_INIT   |
function "name"      | OP_foo_function_name                 | OP_signature_sign_init          |
function typedef     | OSSL_OP_foo_function_name_fn         | OSSL_OP_signature_sign_init_fn  |
function ptr getter  | OSSL_get_OP_foo_function_name        | OSSL_get_OP_signature_sign_init |

For example,

- while the documentation uses the `OSSL_OP_` prefix throughout (except for the function "name", where the `OSSL_` prefix is missing) for both the operations and their dispatched "functions",
- the actual implementation uses the `OSSL_FUNC_` prefix instead of `OSSL_OP_` for the function id (which is consistent with the name of the `OSSL_CORE_MAKE_FUNC()` macro), but retains the `OP_` for the remaining names.

This inconsistency can be observed in the source code by grepping for `signature_sign_init`

```make
crypto/evp/evp_local.h:169:    OSSL_OP_signature_sign_init_fn *sign_init;
crypto/evp/signature.c:68:            signature->sign_init = OSSL_get_OP_signature_sign_init(fns);

doc/man7/provider-keymgmt.pod:125:operations, such as OP_signature_sign_init() (see

doc/man7/provider-signature.pod:26: int OP_signature_sign_init(void *ctx, void *provkey);

doc/man7/provider-signature.pod:110: OP_signature_sign_init              OSSL_FUNC_SIGNATURE_SIGN_INIT

doc/man7/provider-signature.pod:146:=item OP_signature_sign_init and OP_signature_sign
doc/man7/provider-signature.pod:189:OP_signature_sign_init() initialises a context for signing given a provider side

include/openssl/core_functions.h:593:OSSL_CORE_MAKE_FUNC(int, OP_signature_sign_init, (void *ctx, void *provkey))
```

### New naming convention

The new naming convention which I propose uses the `OSSL_OP_` prefix for the operations only and the `OSSL_FUNC_` prefix (without exceptions) for the operation. 

Type                 | Documentation (evp_generic_fetch(3)) |  Implementation (example)        |
---------------------|--------------------------------------|----------------------------------|
operation            | OSSL_OP_FOO                          | OSSL_OP_SIGNATURE                |
function id          | OSSL_FUNC_FOO_FUNCTION_NAME          | OSSL_FUNC_SIGNATURE_SIGN_INIT    |
function "name"      | OSSL_FUNC_foo_function_name          | OSSL_FUNC_signature_sign_init    |
function typedef     | OSSL_FUNC_foo_function_name_fn       | OSSL_FUNC_signature_sign_init_fn |
function ptr getter  | OSSL_FUNC_foo_function_name          | OSSL_FUNC_signature_sign_init    |

What I like most about the naming convention is:  the function's "name" coincides with the name of the getter for obtaining its function pointer. So, in a sense, `OSSL_FUNC_signature_sign_init(ctx, provkey)` is a "sloppy" notation for `OSSL_FUNC_signature_sign_int(fns)(ctx, provkey)`. Isn't that nicely intuitive and consistent?

As mentioned in the commit message, the consistency goes beyond operations:  _all_ dispatched functions, including the up- and downcalls between core and providers, consistently have the `OSSL_FUNC_` prefix

- OSSL_FUNC_core_
- OSSL_FUNC_provider_


romen:
@mspncp nice work!
I would totally support applying this set of changes at the cost of having to rebase many of the pending PRs!

mspncp:
> @mspncp nice work!
> I would totally support applying this set of changes at the cost of having to rebase many of the pending PRs!

Thanks @romen! FWIW: I wouldn't mind assisting with the rebase of the PRs. I have some routine doing the changes.

levitte:
Correction to make: in the new standdard, `OSSL_FUNC_SIGNATURE_SIGN` should be `OSSL_FUNC_SIGNATURE_SIGN_INIT`

mspncp:
> Correction to make: in the new standdard, `OSSL_FUNC_SIGNATURE_SIGN` should be `OSSL_FUNC_SIGNATURE_SIGN_INIT`

You are right, I'll correct that. Originally, my example contained the  'sign' function, then I switched to 'sign_init'.  In the source code, it should be correct.

apps/provider.c
```diff
@@ -18,7 +18,7 @@
 #include <openssl/safestack.h>
 #include <openssl/provider.h>
 #include <openssl/core.h>
-#include <openssl/core_numbers.h>
+#include <openssl/core_functions.h>
```
levitte, line 21: This name doesn't translate well in my mind.  While that header does include function typedefs for core functions offered to the provider, the majority of it includes function typedefs that *providers* may offer.

The reason I chose "core_numbers" was that I viewed the numbers as the primary data and the associated typedefs as auxiliary data at the time, and those numbers *are* offered by the core.  Not sure I'd view it as categorically today, though, but my imagination hasn't been enough to find a new name that I'm satisfied with...

crypto/core_algorithm.c
```diff
@@ -29,7 +29,7 @@ static int algorithm_do_this(OSSL_PROVIDER *provider, void *cbdata)
     struct algorithm_data_st *data = cbdata;
     int no_store = 0;    /* Assume caching is ok */
     int first_operation = 1;
-    int last_operation = OSSL_OP__HIGHEST;
+    int last_operation = OSSL_OP_HIGHEST;
```
levitte, line 32: The reason for the double underscore is so that it's not mixed up with the usual operation numbers...  that is, unless you want to claim that there is a "HIGHEST" operation :wink:

doc/man7/provider-asym_cipher.pod
```diff
@@ -60,73 +60,73 @@ provider_query_operation() function
...
-For example, the "function" OP_asym_cipher_newctx() has these:
+B<OSSL_FUNC_{name}>.
+For example, the "function" OSSL_FUNC_asym_cipher_newctx() has these:
 
- typedef void *(OSSL_OP_asym_cipher_newctx_fn)(void *provctx);
- static ossl_inline OSSL_OP_asym_cipher_newctx_fn
-     OSSL_get_OP_asym_cipher_newctx(const OSSL_DISPATCH *opf);
+ typedef void *(OSSL_FUNC_asym_cipher_newctx_fn)(void *provctx);
+ static ossl_inline OSSL_FUNC_asym_cipher_newctx_fn
+     OSSL_FUNC_asym_cipher_newctx(const OSSL_DISPATCH *opf);
```
levitte, line 68: So, while I see the charm with having this getter named being exactly the same as the function "name", I'm wondering if it won't be confusing about, 'cause when we're talking about them in text, how will we do so and still make sure that it's understood which one we talk about?  What is the nomenclature going to be?  What extra words will we have to add as context?

include/openssl/core.h
```diff
@@ -185,15 +185,15 @@ typedef void (*OSSL_thread_stop_handler_fn)(void *arg);
  *              provider needs it.  This value is passed to other provider
  *              functions, notably other context constructors.
  */
-typedef int (OSSL_provider_init_fn)(const OSSL_CORE_HANDLE *handle,
+typedef int (OSSL_FUNC_provider_init_fn)(const OSSL_CORE_HANDLE *handle,
```
levitte, line 188: Absolutely not!  The provider init function isn't a dispatched function, it's a module/provider entry point.

apps/provider.c
mspncp: > This name doesn't translate well in my mind.

Well, I anticipated that this would be the most controversial change, so I left it in a separate commit from the beginning, to be able to change or drop it. :-)

I understand your concerns that the term "functions" is too narrowed down for what the header contains. But "numbers" is quite the opposite, it is vague and leaves a lot of room for imagination, too. After all, the fact that we are using natural numbers for representing the IDs of the functions in the dispatch table is an implementation detail, isn't it?

Speeking of dispatch tables: how would the name `<openssl/core_dispatch.h> translate in your mind`? Would that fit better? If you like the name, then I would suggest moving the declaration of `struct ossl_dispatch_st` from core.h into core_dispatch.h as well, then the name makes even more sense.

https://github.com/openssl/openssl/blob/fe3c5424ec5dffbeca5d71164aff3eb9a8cc008c/include/openssl/core.h#L33-L42

crypto/core_algorithm.c
mspncp: I guessed that this was your intention, but do you _really_ think such a precaution is necessary?

doc/man7/provider-asym_cipher.pod
mspncp: I think I should add some more detailed explanation of this duality to `evp_generic_fetch(3)`, along the lines of what I wrote in the pr description.

Also, in practice I believe that there is no risk of confusion, because people will call these function through saved function pointers anyway.

include/openssl/core.h
mspncp: Agreed. Is this the only exceptional case which needs to be reverted?

include/openssl/core.h
```diff
@@ -185,15 +185,15 @@ typedef void (*OSSL_thread_stop_handler_fn)(void *arg);
...
                                     const OSSL_DISPATCH *in,
                                     const OSSL_DISPATCH **out,
                                     void **provctx);
 # ifdef __VMS
 #  pragma names save
 #  pragma names uppercase,truncated
 # endif
-extern OSSL_provider_init_fn OSSL_provider_init;
+extern OSSL_FUNC_provider_init_fn OSSL_FUNC_provider_init;
```
mspncp, line 196: (This line will be reverted, too)

doc/man7/provider-asym_cipher.pod
mspncp: The name fits nicely to the existing description...
```pod
=head1 NAME

openssl/core_dispatch.h
- OpenSSL provider dispatch numbers and function types
```

mspncp: > I think I should add some more detailed explanation of this duality to `evp_generic_fetch(3)`, along the lines of what I wrote in the pr description.

Or maybe `provider-base(7)` would be a better place for such general explanations? :thinking:

mspncp: > ... how will we do so and still make sure that it's understood which one we talk about? What is the nomenclature going to be?

Placing "function" in quotes as a reminder, as it is done currently (`git grep -F '"function' --  doc/`), is an option but not ideal IMO. Maybe we should use the term *dispatch function* instead?

apps/provider.c
levitte: Hmmm, `core_dispatch.h` is ok, but I would prefer to define the central types in `core.h`

crypto/core_algorithm.c
levitte: Yes

doc/man7/provider-asym_cipher.pod
levitte: That does remind me that I've used the term *dispatched function*  at times...  yeah, that can work

mspncp: _dispatch function_ or _dispatched function_ are both ok. Whatever you prefer.

mspncp: No, _dispatched function_ is better.

mspncp:
To keep changes overseeable, this [force-push](https://github.com/openssl/openssl/compare/fe3c5424ec5dffbeca5d71164aff3eb9a8cc008c..35c16698ef29833a83cbeb6ec4159f82dbaa16e6) contains only a replacement of the first commit which renames `core_numbers.h` to `core_dispatch.h` (instead of `core_functions.h`).

apps/provider.c
mspncp: Done, see [8a94c5ed1](https://github.com/openssl/openssl/pull/12222/commits/8a94c5ed1e78bb6db0079b722dce0151cf2b5e27).

include/openssl/core.h
mspncp: [Reverted](https://github.com/openssl/openssl/pull/12222/commits/ae887295945c9e1aea224f8176c321653dc629b6).

crypto/core_algorithm.c
mspncp: [Reverted](https://github.com/openssl/openssl/pull/12222/commits/cae2cfdff71d9315e8f42227b3a684e1d0c40d2e).

doc/man7/provider-asym_cipher.pod
mspncp: Still tbd...

mspncp:
@levitte commit f118ee5f22c00013558ea77e1088f8fb60ba14aa is an experimental attempt to improve the documentation of the dispatched functions. Let me know what you think about the idea to add curly braces around the function names in the prototypes, and whether those curly braces should be used throughout the pod file, or only in the prototypes themselves.

doc/man7/provider-base.pod
```diff
@@ -7,168 +7,186 @@ provider-base
...
- #include <openssl/core_numbers.h>
+ #include <openssl/core_dispatch.h>
 
  /*
-  * None of these are actual functions, but are displayed like this for
-  * the function signatures for functions that are offered as function
-  * pointers in OSSL_DISPATCH arrays.
+  * The following function prototypes don't represent actual C functions
+  * but rather the signatures of so-called 'dispatched' functions.
+  * To indicate the difference, curly {braces} have been added around
```
levitte, line 15: I *like* that.  Wouldn't mind seeing that consistently in all provider pods.

mspncp: Is it ok if I split the pod changes into a separate pr? This would speed up the merging of this pr.

levitte: I'm ok with that

mspncp: Ok, I dropped the experimental commit and rebased the pull request (without further changes) to the tip of master, just to make sure I didn't miss any new dispatched functions on master in the mean time. (This was not the case.)  Ready for review.

mspncp:
The documentation update for the dispatched functions will be a followup pull request, which I'll try to submit as soon as possible.

mspncp:
> Note: I'm a little surprised to see remaining pod changes, but it appears to just be the renaming, so I look forward to the more involved changes in the follow up PR

I noticed that too and briefly considered to detach the doc changes entirely. But then the documentation would have been wrong on master, because it still would have contained the OP_ and OSSL_OP_ prefixes. That's why I left the renaming here. So documentation currently _correct_ but it will be _improved_ by the followup pr.

mspncp:
> The documentation update for the dispatched functions will be a followup pull request, which I'll try to submit as soon as possible.

See draft pull request #12250.

levitte:
Oops...  rebase needed

mspncp:
@levitte the rebase conflict was trivial to resolve, see diff below. The only significant change which requires approval is in the last commmit, where I applied my changes to @paulidale's dispatched rand functions. I would suggest, that your reapproval and a successful CI build is sufficient for merging, without resetting the grace period again.

```diff
diff --cc crypto/evp/evp_local.h
index 132534464f,95a046c2ed..0000000000
--- a/crypto/evp/evp_local.h
+++ b/crypto/evp/evp_local.h
@@@ -9,8 -9,7 +9,14 @@@
  
  /* EVP_MD_CTX related stuff */
  
++<<<<<<< HEAD
 +#include <openssl/core_numbers.h>
 +#include "internal/refcount.h"
++||||||| merged common ancestors
++#include <openssl/core_numbers.h>
++=======
+ #include <openssl/core_dispatch.h>
++>>>>>>> Rename <openssl/core_numbers.h> -> <openssl/core_dispatch.h>
  
  #define EVP_CTRL_RET_UNSUPPORTED -1
```

levitte:
Yes, my approval stands

mspncp:
Travis builds succeeded, only test failures due to #12246. AppVeyor the sloth is still queued...  I'm going to merge now.

mspncp:
Merged, thanks!

androidb88:
Admin key

PR #12228, opened 2020-06-22
CORE: Add OPENSSL_CTX_set0_default(), to set a default library context

levitte:
Applications may want to set their own default library context,
possibly per-thread.  OPENSSL_CTX_set0_default() does that.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,24 @@ OpenSSL 3.0
 
 ### Changes between 1.1.1 and 3.0 [xx XXX xxxx]
 
+ * Added a library context that applications as well as other
+   libraries can use to form a separate context within which libcrypto
+   operations are performed.
+
+   There are two ways this can be used:
+
+   - Directly, by passing a library context to functions that take
+     such an argument, such as `EVP_CIPHER_fetch` and similar algorithm
+     fetching functions.
+   - Indirectly, by creating a new library context and then assigning
+     it as the new default, with `OPENSSL_CTX_set0_default`.
+
+   All public OpenSSL functions that take an `OPENSSL_CTX` pointer,
+   apart from the functions directly related to `OPENSSL_CTX`, accept
+   NULL to indicate that the default library context should be used.
+
+   *Richard Levitte*
+
  * Handshake now fails if Extended Master Secret extension is dropped
    on renegotiation.
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -39,6 +39,10 @@ OpenSSL 3.0
    apart from the functions directly related to `OPENSSL_CTX`, accept
    NULL to indicate that the default library context should be used.
 
+   Library code that changes the default library context using
+   `OPENSSL_CTX_set0_default` should take care to restore it with a
+   second call before returning to the caller.
+
    *Richard Levitte*
 
  * Handshake now fails if Extended Master Secret extension is dropped
--- a/crypto/async/async.c
+++ b/crypto/async/async.c
@@ -252,7 +252,7 @@ int ASYNC_start_job(ASYNC_JOB **job, ASYNC_WAIT_CTX *wctx, int *ret,
 
         ctx->currjob->func = func;
         ctx->currjob->waitctx = wctx;
-        libctx = get_default_context();
+        libctx = openssl_ctx_get_concrete(NULL);
         if (!async_fibre_swapcontext(&ctx->dispatcher,
                 &ctx->currjob->fibrectx, 1)) {
             ASYNCerr(ASYNC_F_ASYNC_START_JOB, ASYNC_R_FAILED_TO_SWAP_CONTEXT);
--- a/crypto/async/async.c
+++ b/crypto/async/async.c
@@ -170,6 +170,7 @@ int ASYNC_start_job(ASYNC_JOB **job, ASYNC_WAIT_CTX *wctx, int *ret,
                     int (*func)(void *), void *args, size_t size)
 {
     async_ctx *ctx;
+    OPENSSL_CTX *libctx;
 
     if (!OPENSSL_init_crypto(OPENSSL_INIT_ASYNC, NULL))
         return ASYNC_ERR;
@@ -203,6 +204,11 @@ int ASYNC_start_job(ASYNC_JOB **job, ASYNC_WAIT_CTX *wctx, int *ret,
 
             if (ctx->currjob->status == ASYNC_JOB_PAUSED) {
                 ctx->currjob = *job;
+                /*
+                 * Restore the default libctx to what it was the last time the
+                 * fibre ran
+                 */
+                libctx = OPENSSL_CTX_set0_default(ctx->currjob->libctx);
                 /* Resume previous job */
                 if (!async_fibre_swapcontext(&ctx->dispatcher,
                         &ctx->currjob->fibrectx, 1)) {
@@ -210,6 +216,12 @@ int ASYNC_start_job(ASYNC_JOB **job, ASYNC_WAIT_CTX *wctx, int *ret,
                              ASYNC_R_FAILED_TO_SWAP_CONTEXT);
                     goto err;
                 }
+                /*
+                 * In case the fibre changed the default libctx we set it back
+                 * again to what it was originally, and remember what it had
+                 * been changed to.
+                 */
+                ctx->currjob->libctx = OPENSSL_CTX_set0_default(libctx);
                 continue;
             }
 
@@ -240,11 +252,17 @@ int ASYNC_start_job(ASYNC_JOB **job, ASYNC_WAIT_CTX *wctx, int *ret,
 
         ctx->currjob->func = func;
         ctx->currjob->waitctx = wctx;
+        libctx = get_default_context();
         if (!async_fibre_swapcontext(&ctx->dispatcher,
                 &ctx->currjob->fibrectx, 1)) {
             ASYNCerr(ASYNC_F_ASYNC_START_JOB, ASYNC_R_FAILED_TO_SWAP_CONTEXT);
             goto err;
         }
+        /*
+         * In case the fibre changed the default libctx we set it back again
+         * to what it was, and remember what it had been changed to.
+         */
+        ctx->currjob->libctx = OPENSSL_CTX_set0_default(libctx);
     }
 
 err:
--- a/crypto/context.c
+++ b/crypto/context.c
@@ -138,7 +138,7 @@ static OPENSSL_CTX *get_thread_default_context(void)
     return CRYPTO_THREAD_get_local(&default_context_thread_local);
 }
 
-OPENSSL_CTX *get_default_context(void)
+static OPENSSL_CTX *get_default_context(void)
 {
     OPENSSL_CTX *current_defctx = get_thread_default_context();
 
--- a/crypto/context.c
+++ b/crypto/context.c
@@ -138,7 +138,7 @@ static OPENSSL_CTX *get_thread_default_context(void)
     return CRYPTO_THREAD_get_local(&default_context_thread_local);
 }
 
-static OPENSSL_CTX *get_default_context(void)
+OPENSSL_CTX *get_default_context(void)
 {
     OPENSSL_CTX *current_defctx = get_thread_default_context();
 
--- a/crypto/context.c
+++ b/crypto/context.c
@@ -214,6 +214,15 @@ int openssl_ctx_is_default(OPENSSL_CTX *ctx)
     return 0;
 }
 
+int openssl_ctx_is_global_default(OPENSSL_CTX *ctx)
+{
+#ifndef FIPS_MODULE
+    if (openssl_ctx_get_concrete(ctx) == &default_context_int)
+        return 1;
+#endif
+    return 0;
+}
+
 static void openssl_ctx_generic_new(void *parent_ign, void *ptr_ign,
                                     CRYPTO_EX_DATA *ad, int index,
                                     long argl_ign, void *argp)
--- a/crypto/context.c
+++ b/crypto/context.c
@@ -39,13 +39,6 @@ struct openssl_ctx_st {
     struct openssl_ctx_onfree_list_st *onfreelist;
 };
 
-#ifndef FIPS_MODULE
-static OPENSSL_CTX default_context_int;
-
-/* Always points at default_context_int if it has been initialised */
-static OPENSSL_CTX *default_context = NULL;
-#endif
-
 static int context_init(OPENSSL_CTX *ctx)
 {
     size_t i;
@@ -120,18 +113,46 @@ static int context_deinit(OPENSSL_CTX *ctx)
 }
 
 #ifndef FIPS_MODULE
-void openssl_ctx_default_deinit(void)
-{
-    context_deinit(default_context);
-}
+/* The default default context */
+static OPENSSL_CTX default_context_int;
 
 static CRYPTO_ONCE default_context_init = CRYPTO_ONCE_STATIC_INIT;
-DEFINE_RUN_ONCE_STATIC(do_default_context_init)
-{
-    if (context_init(&default_context_int))
-        default_context = &default_context_int;
+static CRYPTO_THREAD_LOCAL default_context_thread_local;
 
-    return 1;
+DEFINE_RUN_ONCE_STATIC(default_context_do_init)
+{
+    return CRYPTO_THREAD_init_local(&default_context_thread_local, NULL)
+        && context_init(&default_context_int);
+}
+
+void openssl_ctx_default_deinit(void)
+{
+    context_deinit(&default_context_int);
+}
+
+static OPENSSL_CTX *get_thread_default_context(void)
+{
+    if (!RUN_ONCE(&default_context_init, default_context_do_init))
+        return NULL;
+
+    return CRYPTO_THREAD_get_local(&default_context_thread_local);
+}
+
+static OPENSSL_CTX *get_default_context(void)
+{
+    OPENSSL_CTX *current_defctx = get_thread_default_context();
+
+    if (current_defctx == NULL)
+        current_defctx = &default_context_int;
+    return current_defctx;
+}
+
+static int set_default_context(OPENSSL_CTX *defctx)
+{
+    if (defctx == &default_context_int)
+        defctx = NULL;
+
+    return CRYPTO_THREAD_set_local(&default_context_thread_local, defctx);
 }
 #endif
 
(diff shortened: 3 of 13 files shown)


levitte:
I'd like to write a test, but fail to figure out how to make it representative enough.  Ideas?

mattcaswell:
A simple test would be something like this:

````
    OSSL_PROVIDER nullprov = OSSL_PROVIDER_load(NULL, "null");
    OPENSSL_CTX *mainctx= OPENSSL_CTX_new();
    OPENSSL_CTX *oldctx;
    EVP_MD sha256;

    sha256 = EVP_MD_fetch(NULL, "SHA2-256", NULL);
    if (!TEST_ptr_null(sha256))
        goto err;
    oldctx = OPENSSL_CTX_set0_default(mainctx);
    sha256 = EVP_MD_fetch(NULL, "SHA2-256", NULL);
    if (!TEST_ptr(sha256))
        goto err;
    EVP_MD_free(sha256);
    sha256 = NULL;
    if (!TEST_ptr_eq(OPENSSL_CTX_set0_default(oldctx), mainctx))
        goto err;
    sha256 = EVP_MD_fetch(NULL, "SHA2-256", NULL);
    if (!TEST_ptr_null(sha256))
        goto err;
````

That doesn't test the thread local aspect of course. To do that you'd need to have multiple threads running and check that changing the libctx in one doesn't affect the other one.

**Comment by levitte** [comment]
That's a good enough start.  Thanks!

**Comment by levitte** [comment]
That test helped me find some bugs too, thanks!

File: crypto/context.c [hunk 1]
```diff
@@ -160,22 +181,32 @@ void OPENSSL_CTX_free(OPENSSL_CTX *ctx)
...
-            return 0;
-        return default_context;
-    }
+    if (ctx == NULL)
+        return get_default_context();
 #endif
     return ctx;
 }
 
 int openssl_ctx_is_default(OPENSSL_CTX *ctx)
```
paulidale, line 206: Given that this function is only used in two places (both in `evp_fetch()`) and both are raising an error indicating default vs non-default library context, perhaps it might indicate that a default library context was used but that it wasn't the default default one:

``` C
const char *openssl_ctx_is_default(OPENSSL_CTX *ctx)
{
    OPENSSL_CTX *dflt = get_default_context();

    if (ctx == NULL)
        ctx = dflt;
    if (ctx == &default_context_int)
        return "Default library context";
    if (ctx == dflt)
        return "Non-default default library context";
    return "Non-default library context";
}
```

crypto/context.c
```diff
@@ -120,18 +113,46 @@ static int context_deinit(OPENSSL_CTX *ctx)
...
-    if (context_init(&default_context_int))
-        default_context = &default_context_int;
+    OPENSSL_CTX *current_defctx;
 
-    return 1;
+    if (!RUN_ONCE(&default_context_init, default_context_do_init))
+        return NULL;
+
+    current_defctx = CRYPTO_THREAD_get_local(&default_context_thread_local);
+    if (current_defctx == (OPENSSL_CTX *)-1)
```
paulidale, line 136: Why the -1 cast and test here.  Doesn't `CRYPTO_THREAD_get_local()` return NULL if not found?

levitte: I copied code from elsewhere, and misremembered how that was done...  It infuriates me, though, that NULL wouldn't be a valid value.

crypto/context.c
levitte: Er, so you want the caller to compare magic text to find out?

crypto/context.c
levitte: Fixed, I simply removed this check

crypto/context.c
levitte: Point taken, though, I solved that with one more function

levitte: Done

crypto/evp/evp_fetch.c
```diff
@@ -306,7 +306,9 @@ void *evp_generic_fetch(OPENSSL_CTX *libctx, int operation_id,
 #else
         ERR_raise_data(ERR_LIB_EVP, code,
                        "%s, Algorithm (%s), Properties (%s)",
-                       (openssl_ctx_is_default(libctx)
+                       (openssl_ctx_is_global_default(libctx)
+                        ? "Global default library context"
+                        : openssl_ctx_is_default(libctx)
                         ? "Default library context"
```
paulidale, line 312: Shouldn't his be indented?

crypto/evp/evp_fetch.c
```diff
@@ -350,7 +352,9 @@ void *evp_generic_fetch_by_number(OPENSSL_CTX *libctx, int operation_id,
 
             ERR_raise_data(ERR_LIB_EVP, code,
                            "%s, Algorithm (%s), Properties (%s)",
-                           (openssl_ctx_is_default(libctx)
+                           (openssl_ctx_is_global_default(libctx)
+                            ? "Global default library context"
+                            : openssl_ctx_is_default(libctx)
                             ? "Default library context"
```
paulidale, line 358: Indentation here too.

crypto/context.c
paulidale: NULL is fine here.  I thought this through when I asked about the cast.

crypto/context.c
paulidale: There is no need to compare text, the output is only ever being printed into an error message.

crypto/evp/evp_fetch.c
levitte: "Should" is a matter of taste in this case.  We have said nothing about this particular type of condition cascade.  That being said, I can add parentheses around the last condition if you think that's better.

levitte: Ping?

CHANGES.md
```diff
@@ -23,6 +23,24 @@ OpenSSL 3.0
...
 ### Changes between 1.1.1 and 3.0 [xx XXX xxxx]
 
+ * Added a library context that applications as well as other
+   libraries can use to form a separate context within which libcrypto
+   operations are performed.
+
+   There are two ways this can used:
+
+   - Directly, by passing a library context to functions that take
+     such a argument, such as `EVP_CIPHER_fetch` and similar algorithm
```
mattcaswell, line 33: typo: an argument

crypto/evp/evp_fetch.c
mattcaswell, line 312: Perhaps the string should be something like "Thread Local default library context" to more clearly distinguish it from the global default.

doc/man3/OPENSSL_CTX.pod
```diff
@@ -14,12 +15,13 @@ OPENSSL_CTX, OPENSSL_CTX_new, OPENSSL_CTX_free, OPENSSL_CTX_load_config
...
+ OPENSSL_CTX *OPENSSL_CTX_set0_default(OPENSSL_CTX *libctx);
 
 =head1 DESCRIPTION
 
-C<OPENSSL_CTX> is an internal OpenSSL library context type.
-Applications may allocate their own, but may also use C<NULL> to use
-the internal default context with functions that take a C<OPENSSL_CTX>
+B<OPENSSL_CTX> is an internal OpenSSL library context type.
+Applications may allocate their own, but may also use NULL to use
+the internal default context with functions that take a B<OPENSSL_CTX>
```
mattcaswell, line 24: I would say "take an OPENSSL_CTX", ie. "an" rather than "a"

CHANGES.md
levitte: Fixed

crypto/evp/evp_fetch.c
levitte: That's sensible.  Done

doc/man3/OPENSSL_CTX.pod
levitte: Fixed

crypto/evp/evp_fetch.c
levitte: I ended up refactoring a bit more so there would be only one place where these strings are defined.  It's currently static in this file, but could easily be moved to a libcrypto-internal function if need be (subject for another PR, meybe?)

crypto/evp/evp_fetch.c
```diff
@@ -286,6 +286,18 @@ inner_evp_generic_fetch(OPENSSL_CTX *libctx, int operation_id,
...
 }
 
+#ifndef FIPS_MODULE
+static const char *libctx_descriptor(OPENSSL_CTX *libctx)
+{
+    return
+        openssl_ctx_is_global_default(libctx)
+        ? "Global default library context"
+        : openssl_ctx_is_default(libctx)
+        ? "Thread-local default library context"
```
paulidale, line 296: This should still be indented two spaces right.

richsalz: This code is way too clever; it's not necessary to do JAPH-style one-liners.  Even the original author can't get the indentation right.

richsalz: After all, it's just a return statement:
```
if (openssl_ctx_is_global_default(libctx))
    return "Global default library context";
if (openssl_ctx_is_default(libctx))
    return "Thread-local default library context";
return "Non-default library context";
```

levitte: @paulidale, there's no basis in the coding style for "should", this is an unprecedented (?) case.
I'll convert this to what @richsalz proposes, that'll be the sane way to avoid a dispute.

levitte: Done

doc/man3/OPENSSL_CTX.pod
```diff
@@ -32,19 +34,23 @@ OPENSSL_CTX_load_config() loads a configuration file using the given C<ctx>.
 This can be used to associate a libctx with providers that are loaded from
 a configuration.
 
-OPENSSL_CTX_free() frees the given C<ctx>.
+OPENSSL_CTX_free() frees the given I<ctx>.
+
+OPENSSL_CTX_set0_default() sets the default OpenSSL library context to be
+I<libctx> in the current thread.  The previous default library context is
+returned.
```
slontis, line 41: Maybe we should recommend that it is restored to its old value via a subsequent call to OPENSSL_CTX_set0_default() using the previously returned libctx.

levitte: Done

CHANGES.md
```diff
@@ -23,6 +23,24 @@ OpenSSL 3.0
 
 ### Changes between 1.1.1 and 3.0 [xx XXX xxxx]
 
+ * Added a library context that applications as well as other
+   libraries can use to form a separate context within which libcrypto
+   operations are performed.
+
+   There are two ways this can used:
```
mattcaswell, line 30: Typo: can be used

crypto/context.c
```diff
@@ -120,18 +113,47 @@ static int context_deinit(OPENSSL_CTX *ctx)
...
+    if (defctx == &default_context_int)
+        defctx = NULL;
+
+    return CRYPTO_THREAD_set_local(&default_context_thread_local, defctx);
+}
+
+void openssl_ctx_default_deinit(void)
+{
+    context_deinit(get_default_context());
+    set_default_context(NULL);
```
mattcaswell, line 156: I'm not convinced this function is quite right. This is called during auto de-init via atexit. I think it should always be de-initing the global default ctx. This begs the question - what happens if we enter de-init with the default ctx for the current thread being something other than the global default ctx? And should that be allowed - or is that considered a programmer bug?

slontis:
Hmm, I have just added to a PR..

```
EVP_PBE_scrypt_with_libctx
PKCS5_PBKDF2_HMAC_with_libctx
```

Should we start using OPENSSL_CTX_set0_default for these type of cases?

levitte:
> Should we start using OPENSSL_CTX_set0_default for these type of cases?

Probably...  especially if you have reason to believe that the library context is needed further down the code path

crypto/context.c
levitte: I can change that back

CHANGES.md
levitte: Fixed

crypto/context.c
levitte: Done

mattcaswell:
A thought about this. It's going to go horribly wrong when doing ASYNC stuff....

mattcaswell:
I think you're going to need some logic in ASYNC_start_job to handle it, i.e. before and after `async_fibre_swapcontext()` calls.

crypto/evp/evp_fetch.c
paulidale: @richsalz's suggestion is best of all.

I'd argue that the `X ? Y : Z` in `A ? B : X ? Y : Z` is its own expression and the ?: should indent under the X.  More readable would have been writing it `A ? B : (X ? Y : Z)` where the indentation is unambiguous.

slontis: I think this took a tern for the worse. (sorry couldn't resist..)

crypto/context.c
```diff
@@ -155,27 +176,48 @@ int OPENSSL_CTX_load_config(OPENSSL_CTX *ctx, const char *config_file)
...
+    if (openssl_ctx_is_default(ctx))
+        return;
+
+    context_deinit(ctx);
     OPENSSL_free(ctx);
 }
 
+OPENSSL_CTX *OPENSSL_CTX_set0_default(OPENSSL_CTX *libctx)
+{
+#ifndef FIPS_MODULE
```
slontis, line 188: I am not in a position to say if this is going to bite us in fips mode yet or not.. I guess we will have to try it and see if there are any scenarios related to this.

slontis:
Any chance you can squash this to make it easier to cherrypick?

levitte:
> Any chance you can squash this to make it easier to cherrypick?

Uhmm, sure.  But I can give you another tip too: merge

levitte:
squashing rebase done

levitte:
@mattcaswell agreed to assist re the async stuff, as I have a hard time wrapping my head around its mechanisms.  I won't do more changes until that's done.

slontis:
This seems to be hitting the spot in addressing some of the issues with the X509 libctx problems.

levitte:
> This seems to be hitting the spot in addressing some of the issues with the X509 libctx problems.

I fully expected that...  as a matter of fact, I've started to think that there are a number of issues we could have avoided, had we done this earlier.

mattcaswell:
I pushed 2 additional commits into this PR: One adding a test for how the ASYNC code behaves wrt OPENSSL_CTX_set0_default(), and one fixing the issues.

mattcaswell:
Oh...and just added a small fixup for something I missed in the test.

mattcaswell:
Incidentally I find it surprising that asynctest does not seem to have been converted to use our internal testlib stuff. I haven't attempted to address that issue in this PR.

crypto/async/async.c
```diff
@@ -240,11 +252,17 @@ int ASYNC_start_job(ASYNC_JOB **job, ASYNC_WAIT_CTX *wctx, int *ret,
 
         ctx->currjob->func = func;
         ctx->currjob->waitctx = wctx;
+        libctx = get_default_context();
```
levitte, line 255: There's already a non-static function  you can use:

``` C
        libctx = openssl_ctx_get_concrete(NULL);
```

I would rather not open up `get_default_context`

levitte: Oh... Right, my PR, I can fix as I please :wink:

mattcaswell: Right!

levitte: Done

levitte:
I also added a bit more documentation in `OPENSSL_CTX.pod` regarding the interaction between default library contexts and async jobs.  I hope it's understandable.

crypto/context.c
```diff
@@ -155,27 +176,48 @@ int OPENSSL_CTX_load_config(OPENSSL_CTX *ctx, const char *config_file)
 
 void OPENSSL_CTX_free(OPENSSL_CTX *ctx)
 {
-    if (ctx != NULL)
-        context_deinit(ctx);
+    if (openssl_ctx_is_default(ctx))
+        return;
```
mattcaswell, line 180: I wonder about this. Does this indicate a programmer error?

levitte: Well, I have documented it like this:

``` pod
OPENSSL_CTX_free() frees the given I<ctx>, unless it happens to be the
default OpenSSL library context.

OPENSSL_CTX_set0_default() sets the default OpenSSL library context to be
I<ctx> in the current thread.  The previous default library context is
returned.  Care should be taken by the caller to restore the previous
default library context with a subsequent call of this function.
```

Basically, it's there to avoid shot-in-foot failure.

mattcaswell: I do wonder whether it is better to allow the shot in the foot - which will presumably eventually fail due to use-after-free. Or (possibly) make this an assertion. Basically if this happens the programmer has gone wrong.

levitte: I can certainly *wrap* the condition with `ossl_assert`...

mattcaswell: Yeah. That would do it. Although, you'll need to separate out the NULL check - since ctx can be NULL in normal operation.

levitte: So, uhm, there's a choice...  either this:

``` C
    /* Passing the default library context is wrong */
    if (ossl_assert(!openssl_ctx_is_default(ctx))) {
        context_deinit(ctx);
        OPENSSL_free(ctx);
    }
```

or this:

``` C
    /*
     * Passing the default library context is wrong, but we deinit regardless.
     * If the caller passed it here, the next use of it will simply explodde in a use-after-free.
     */
    context_deinit(ctx);
    OPENSSL_free(ctx);
```

...  you know, I'm not too sure about the latter, that's a recipe for hard-to-find bugs.

levitte: ... actually, I strongly dislike both variants.  They add absolutely nothing in a release build, except for a possible use-after-free in the latter form, which may happen at any time and without making it obvious or even easy to figure out where the actual bug is.
IF we're to add an assertion, that only that's reasonable is an unconditional one.

mattcaswell: Unconditional is fine in this context I think, but the former would also be ok with me.

levitte: How would you document it?

mattcaswell: Actually, I'm having second thoughts. This is an API error. Normally in such cases we return a failure - and reserve an assert for internal errors - but of course this is a void function. So perhaps just leave it.

CHANGES.md
```diff
@@ -23,6 +23,24 @@ OpenSSL 3.0
...
+   libraries can use to form a separate context within which libcrypto
+   operations are performed.
+
+   There are two ways this can be used:
+
+   - Directly, by passing a library context to functions that take
+     such an argument, such as `EVP_CIPHER_fetch` and similar algorithm
+     fetching functions.
+   - Indirectly, by creating a new library context and then assigning
+     it as the new default, with `OPENSSL_CTX_set0_default`.
```
kaduk, line 36: The way in which we write the intro above combines with this bullet to suggest that it's okay for an "other library" to use `OPENSSL_CTX_set0_default()`, which is very much not the case.

mattcaswell: Err...yes it is?? As long as it sets it back again before control returns to the application then that should be fine.

kaduk: So I think we should say something about save/restore here.  An application could perhaps just set and forget, but a library should not.

levitte: Point received...  I just pushed a fixup that adds a few lines on the matter.

levitte: @mattcaswell, does that reset the clock?

kaduk: Thank you!

levitte:
Merged

5a9752756b68632320c5aed7b6eb38e63a8ebf31 CORE: Add OPENSSL_CTX_set0_default(), to set a default library context
3bd65f9b5b4731acae395d045dea63d7fdfd507b Update NEWS and CHANGES
e31eda006faaadc8bbe0bb809dbbaba0e4323fe6 TEST: Add test to exercise OPENSSL_CTX_set0_default()
cfbd76c1a9737617d4cf90d3a1af704241c97d98 CORE: Add an internal function to distinguish the global default context
6c689e58f70726cfa5533bc10f9669abce199fb8 Make the ASYNC code default libctx aware
92db29e5e8c26af79f740dfd3bddfeae9b5345a8 Add a test to make sure ASYNC aware code gets the right default libctx

PR #12229, opened 2020-06-22
Don't forget our provider ctx when resetting

### LINKED ISSUE #12225: EVP_PKEY KDF's do not work if the fips provider is in the default library context,
After running

```
make TESTS="test_evp" tests V=1
```
Run the evp_test using the fips module in the default library context.

```
export OPENSSL_CONF_INCLUDE=providers
export OPENSSL_MODULES=providers
OPENSSL_CONF=./test/fips.cnf ./test/evp_test/evppkey_kdf.txt
```

This fails when doing a internal fetch because the library context is NULL..

When setting the digest via pkey_kdf_ctrl_str() the call stack is
```
  EVP_KDF_set_ctx_params()
    kdf_tls1_prf_set_ctx_params()
      ossl_prov_macctx_load_from_params() 
         EVP_MAC_fetch(libctx = NULL, macname = “HMAC", properties = NULL);
```

-------------

If using the default provider it really only works due to the trickery related to the default library context..

i.e: the inner_fetch does calls like this...

```
static OSSL_METHOD_STORE *get_evp_method_store(OPENSSL_CTX *libctx)
{
    return openssl_ctx_get_data(libctx, OPENSSL_CTX_EVP_METHOD_STORE_INDEX,
                                &evp_method_store_method);
}
```
Which then calls openssl_ctx_get_concrete() which does special things if the libctx is NULL.

Should we be doing this trickery also if ONLY the fips module is loaded somehow?
Or do we really just not care about this scenario..

This type of pkey really goes pear shaped earlier on when it does the call to 
is_legacy_alg() when calling..
```
ctx = EVP_PKEY_CTX_new_id(kdf_nid, NULL);
```
Which basically tells it to ignore the fetch failure and continue on without a libctx being associated with the returned ctx. 
Should it fail early in the case of the default provider not being loaded??

mattcaswell:
kdf_tls1_prf_reset and kdf_hkdf_reset were resetting a little too much.

We also update evp_test to run the KDF tests for the FIPS provider as well.

Fixes #12225

### DIFF
--- a/providers/implementations/kdfs/hkdf.c
+++ b/providers/implementations/kdfs/hkdf.c
@@ -90,12 +90,14 @@ static void kdf_hkdf_free(void *vctx)
 static void kdf_hkdf_reset(void *vctx)
 {
     KDF_HKDF *ctx = (KDF_HKDF *)vctx;
+    void *provctx = ctx->provctx;
 
     ossl_prov_digest_reset(&ctx->digest);
     OPENSSL_free(ctx->salt);
     OPENSSL_clear_free(ctx->key, ctx->key_len);
     OPENSSL_cleanse(ctx->info, ctx->info_len);
     memset(ctx, 0, sizeof(*ctx));
+    ctx->provctx = provctx;
 }
 
 static size_t kdf_hkdf_size(KDF_HKDF *ctx)
--- a/providers/implementations/kdfs/kbkdf.c
+++ b/providers/implementations/kdfs/kbkdf.c
@@ -122,6 +122,7 @@ static void kbkdf_free(void *vctx)
 static void kbkdf_reset(void *vctx)
 {
     KBKDF *ctx = (KBKDF *)vctx;
+    void *provctx = ctx->provctx;
 
     EVP_MAC_free_ctx(ctx->ctx_init);
     OPENSSL_clear_free(ctx->context, ctx->context_len);
@@ -129,6 +130,7 @@ static void kbkdf_reset(void *vctx)
     OPENSSL_clear_free(ctx->ki, ctx->ki_len);
     OPENSSL_clear_free(ctx->iv, ctx->iv_len);
     memset(ctx, 0, sizeof(*ctx));
+    ctx->provctx = provctx;
 }
 
 /* SP800-108 section 5.1 or section 5.2 depending on mode. */
--- a/providers/implementations/kdfs/krb5kdf.c
+++ b/providers/implementations/kdfs/krb5kdf.c
@@ -78,11 +78,13 @@ static void krb5kdf_free(void *vctx)
 static void krb5kdf_reset(void *vctx)
 {
     KRB5KDF_CTX *ctx = (KRB5KDF_CTX *)vctx;
+    void *provctx = ctx->provctx;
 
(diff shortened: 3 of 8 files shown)


t8m:
Other KDF implementations do the same. Would you mind fixing them as well?

mattcaswell:
Done.

test/evp_test.c
```diff
@@ -2092,9 +2092,14 @@ static int kdf_test_init(EVP_TEST *t, const char *name)
 
     kdf = EVP_KDF_fetch(NULL, name, NULL);
     if (kdf == NULL) {
-        OPENSSL_free(kdata);
-        return 0;
+        /*
+         * Don't fail immediately. This algorithm might not be available in this
```
slontis, line 2096: I have a PR coming that does 2 things related to this:
(1) make the 'availablein' the first entry (before the command). This then means we can completely skip the init method.
(2) create some is_XX_disabled() methods that can be called before the fetches instead of making the fetch fail everywhere..

mattcaswell: @slontis - am I best to remove the test aspects of this PR then - and leave it all to your PR? It sounds like yours is broader than mine.

slontis: Yes that would be good :)

slontis:
There is a mem leak in travis  - so it is not ready to merge..

test/evp_test.c
```diff
@@ -2092,9 +2092,14 @@ static int kdf_test_init(EVP_TEST *t, const char *name)
...
+         * provider. Keep going and see if we have an "Availablein" line. If
+         * not we will fail later.
+         */
```
slontis, line 2099: This is where the mem leak - I am happy for you to remove the test bits if it passes.

test/evp_test.c
```diff
@@ -2254,8 +2259,12 @@ static int pkey_kdf_test_init(EVP_TEST *t, const char *name)
         return 0;
     kdata->ctx = EVP_PKEY_CTX_new_id(kdf_nid, NULL);
     if (kdata->ctx == NULL) {
-        OPENSSL_free(kdata);
-        return 0;
+        /*
+         * Don't fail immediately. This algorithm might not be available in this
+         * provider. Keep going and see if we have an "Availablein" line. If
+         * not we will fail later.
```
slontis, line 2265: This will also leak..

test/evp_test.c
```diff
@@ -2254,8 +2259,12 @@ static int pkey_kdf_test_init(EVP_TEST *t, const char *name)
         return 0;
     kdata->ctx = EVP_PKEY_CTX_new_id(kdf_nid, NULL);
     if (kdata->ctx == NULL) {
-        OPENSSL_free(kdata);
-        return 0;
+        /*
+         * Don't fail immediately. This algorithm might not be available in this
```
slontis, line 2263: This line looks bad to me -  if the ctx is NULL, pkey_kdf_test_parse() would fall over..

mattcaswell:
I remove the test related commits from this PR on the expectations that @slontis's PR will do the same thing. Please take another look.

slontis:
travis error is unrelated.

mattcaswell:
Pushed. Thanks.

PR #12231, opened 2020-06-22
Fix four Coverity findings in apps/cmp.c

DDvO:
Reported by Coverity on 18 May 2020 at 1:58:04 am AEST.

### DIFF
--- a/apps/cmp.c
+++ b/apps/cmp.c
@@ -2042,12 +2042,11 @@ static int handle_opt_geninfo(OSSL_CMP_CTX *ctx)
         return 0;
     }
 
-    aint = ASN1_INTEGER_new();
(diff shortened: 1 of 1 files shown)


paulidale:
Merged to master, thanks.

DDvO:
As discussed by email with @paulidale, I found that the Coverity findings
```
CID 1463570:    (USE_AFTER_FREE)
CID 1463570:    (USE_AFTER_FREE)
Passing freed pointer "e" as an argument to "release_engine".
```
are not really false positives but due to a deep bug in the engine layer,
which prematurely frees the engine in case there is a key loading error in `ENGINE_load_private_key()`.

So the addition of the following line to `apps/cmp.c` was actually needless:
```
     if (engine != NULL) /* workaround for Coverity false positive */
```

PR #12175, opened 2020-06-17
Fix CMP CLI tests

DDvO:
* Fix CMP -days option range checking and test failing with enable-ubsan, reported by @romen https://github.com/openssl/openssl/pull/11998#issuecomment-643693856
* Fix #12156 by doing connections to 127.0.0.1 (e.g., Mock server) without proxy
* Fix #12145 by disabling some test in case of fuzzing
* etc. (see my https://github.com/openssl/openssl/pull/12175#issuecomment-647009784 further below)

PR #12109, opened 2020-06-10
Rename NOTES*, README*, VERSION, HACKING, LICENSE to .md or .txt

DDvO:
This implements https://github.com/openssl/openssl/pull/12098#issuecomment-641355034.

For the files with names changed to `.md`, numerous small layout issues had to be solved because those files are now interpreted as MarkDown. It's likely that I missed some of them but at least all those files look reasonable now.

I also had to update various references to the renamed files - hope I caught all of them (I noticed and corrected some cases where similar references to other files renamed earlier had not been updated). Where it made most sense I converted them to proper (clickable) MarkDown-style references.

PR #12232, opened 2020-06-22
Move test-related detail from INSTALL.md to new test/README.md

DDvO:
as discussed in https://github.com/openssl/openssl/pull/12175#issuecomment-647131775.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -46,7 +46,7 @@ Prerequisites
 
 To install OpenSSL, you will need:
 
- * A make implementation
+ * A "make" implementation
  * Perl 5 with core modules (please read [NOTES.PERL](NOTES.PERL))
  * The Perl module Text::Template (please read [NOTES.PERL](NOTES.PERL))
  * an ANSI C compiler
@@ -820,7 +820,7 @@ This only has an impact when not built "shared".
 
 ### no-stdio
 
-Don't use anything from the C header file "stdio.h" that makes use of the "FILE"
+Don't use anything from the C header file `stdio.h` that makes use of the `FILE`
 type.  Only libcrypto and libssl can be built in this way.  Using this option will
 suppress building the command line applications.  Additionally, since the OpenSSL
 tests also use the command line applications, the tests will also be skipped.
@@ -984,7 +984,7 @@ on all platforms and are confined to the configuration scripts only.
 These assignments override the corresponding value in the inherited environment,
 if there is one.
 
-The following variables are used as "make variables" and can be used as an
+The following variables are used as "`make` variables" and can be used as an
 alternative to giving preprocessor, compiler and linker options directly as
 configuration.  The following variables are supported:
 
@@ -1135,14 +1135,15 @@ run:
 If your system isn't listed, you will have to create a configuration
 file named Configurations/{{ something }}.conf and add the correct
 configuration for your system.  See the available configs as examples
-and read Configurations/README and Configurations/README.design for
-more information.
+and read [Configurations/README](Configurations/README)
+and [Configurations/README.design](Configurations/README.design)
+for more information.
 
 The generic configurations "cc" or "gcc" should usually work on 32 bit
 Unix-like systems.
 
-Configure creates a build file ("Makefile" on Unix, "makefile" on Windows
-and "descrip.mms" on OpenVMS) from a suitable template in Configurations,
+Configure creates a build file (`Makefile` on Unix, `makefile` on Windows
+and `descrip.mms` on OpenVMS) from a suitable template in Configurations,
 and defines various macros in include/openssl/configuration.h (generated
 from include/openssl/configuration.h.in).
 
@@ -1205,8 +1206,7 @@ be tested.  Run:
 **Warning:** you MUST run the tests from an unprivileged account (or disable
 your privileges temporarily if your platform allows it).
 
-If some tests fail, take a look at the [Test Failures](#test-failures)
-subsection of the [Troubleshooting](#troubleshooting) section.
+See the file [test/README.md](test/README.md) for further details.
 
 Install OpenSSL
 ---------------
@@ -1289,8 +1289,8 @@ Package builders who want to configure the library for standard locations,
 but have the package installed somewhere else so that it can easily be
 packaged, can use
 
-  $ make DESTDIR=/tmp/package-root install         # Unix
-  $ mms/macro="DESTDIR=TMP:[PACKAGE-ROOT]" install ! OpenVMS
+    $ make DESTDIR=/tmp/package-root install         # Unix
+    $ mms/macro="DESTDIR=TMP:[PACKAGE-ROOT]" install ! OpenVMS
 
 The specified destination directory will be prepended to all installation
 target paths.
@@ -1465,64 +1465,11 @@ described here.  Examine the Makefiles themselves for the full list.
 Running Selected Tests
 ----------------------
 
-The make variable TESTS supports a versatile set of space separated tokens
-with which you can specify a set of tests to be performed.  With a "current
-set of tests" in mind, initially being empty, here are the possible tokens:
+You can specify a set of tests to be performed
+using the `make` variable `TESTS`.
 
-     alltests      The current set of tests becomes the whole set of available
-                   tests (as listed when you do 'make list-tests' or similar).
-
-     xxx           Adds the test 'xxx' to the current set of tests.
-
-    -xxx           Removes 'xxx' from the current set of tests.  If this is the
-                   first token in the list, the current set of tests is first
-                   assigned the whole set of available tests, effectively making
-                   this token equivalent to TESTS="alltests -xxx".
-
-     nn            Adds the test group 'nn' (which is a number) to the current
-                   set of tests.
-
-    -nn            Removes the test group 'nn' from the current set of tests.
-                   If this is the first token in the list, the current set of
-                   tests is first assigned the whole set of available tests,
-                   effectively making this token equivalent to
-                   TESTS="alltests -xxx".
-
-Also, all tokens except for "alltests" may have wildcards, such as *.
-(on Unix and Windows, BSD style wildcards are supported, while on VMS,
-it's VMS style wildcards)
-
-### Examples
-
-Run all tests except for the fuzz tests:
-
-    $ make TESTS=-test_fuzz test
-
-or, if you want to be explicit:
-
-    $ make TESTS='alltests -test_fuzz' test
-
-Run all tests that have a name starting with "test_ssl" but not those
-starting with "test_ssl_":
-
-    $ make TESTS='test_ssl* -test_ssl_*' test
-
-Run only test group 10:
-
-    $ make TESTS='10'
-
-Run all tests except the slow group (group 99):
-
-    $ make TESTS='-99'
-
-Run all tests in test groups 80 to 99 except for tests in group 90:
-
-    $ make TESTS='[89]? -90'
-
-To stochastically verify that the algorithm that produces uniformly distributed
-random numbers is operating correctly (with a false positive rate of 0.01%):
-
-    $ ./util/wrap.sh test/bntest -stochastic
+See the section [Running Selected Tests of
+test/README.md](test/README.md#running-selected-tests).
 
 Troubleshooting
 ===============
@@ -1634,57 +1581,20 @@ Test Failures
 
 If some tests fail, look at the output.  There may be reasons for the failure
 that isn't a problem in OpenSSL itself (like an OS malfunction or a Perl issue).
-You may want increased verbosity, that can be accomplished like this:
 
-Full verbosity (`make` macro `VERBOSE` or `V`):
+You may want increased verbosity, that can be accomplished as described in
+section [Test Failures of test/README.md](test/README.md#test-failures).
 
-    $ make V=1 test                                  # Unix
-    $ mms /macro=(V=1) test                          ! OpenVMS
-    $ nmake V=1 test                                 # Windows
-
-Verbosity on test failure (`VERBOSE_FAILURE` or `VF´, Unix example shown):
-
-    $ make test VF=1
-
-Verbosity on failed (sub-)tests only (`VERBOSE_FAILURES_ONLY` or `VFO`):
-
-    $ make test VFO=1
-
-Verbosity on failed (sub-)tests, in addition progress on succeeded (sub-)tests
-(`VERBOSE_FAILURES_PROGRESS` or `VFP`):
-
-    $ make test VFP=1
-
-If you want to run just one or a few specific tests, you can use
-the make variable TESTS to specify them, like this:
-
-    $ make TESTS='test_rsa test_dsa' test            # Unix
-    $ mms/macro="TESTS=test_rsa test_dsa" test       ! OpenVMS
-    $ nmake TESTS='test_rsa test_dsa' test           # Windows
-
-And of course, you can combine (Unix examples shown):
-
-    $ make test TESTS='test_rsa test_dsa' VF=1
-    $ make test TESTS="test_cmp_*" VFO=1
-
-You can find the list of available tests like this:
-
-    $ make list-tests                                # Unix
-    $ mms list-tests                                 ! OpenVMS
-    $ nmake list-tests                               # Windows
-
-Have a look at the manual for the perl module Test::Harness to
-see what other HARNESS_* variables there are.
+You may want to selectively specify which test(s) to perform. This can be done
+sing the `make` variable `TESTS` as described in section [Running Selected Tests
+of test/README.md](test/README.md#running-selected-tests).
 
 If you find a problem with OpenSSL itself, try removing any
-compiler optimization flags from the CFLAGS line in Makefile and
-run "make clean; make" or corresponding.
+compiler optimization flags from the `CFLAGS` line in the Makefile and
+run `make clean; make` or corresponding.
 
 To report a bug please open an issue on GitHub, at
 <https://github.com/openssl/openssl/issues>.
-
-For more details on how the make variables TESTS can be used,
-see section [Running Selected Tests](#running-selected-tests) below.
 
 Notes
 =====
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -130,8 +130,8 @@ determined by the user.
 
     [[ options ]]
 
-Note that the notation assumes spaces around {, }, [, ], {{, }} and
-[[, ]].  This is to differentiate from OpenVMS directory
+Note that the notation assumes spaces around `{`, `}`, `[`, `]`, `{{`, `}}` and
+`[[`, `]]`.  This is to differentiate from OpenVMS directory
 specifications, which also use [ and ], but without spaces.
 
 Quick Installation Guide
@@ -175,10 +175,10 @@ and issue the following commands to build OpenSSL.
 As mentioned in the [Choices](#choices) section, you need to pick one
 of the four Configure targets in the first command.
 
-Most likely you will be using the VC-WIN64A target for 64bit Windows
-binaries (AMD64) or VC-WIN32 for 32bit Windows binaries (X86).
-The other two options are VC_WIN64I (Intel IA64, Itanium) and
-VC-CE (Windows CE) are rather uncommon nowadays.
+Most likely you will be using the `VC-WIN64A` target for 64bit Windows
+binaries (AMD64) or `VC-WIN32` for 32bit Windows binaries (X86).
+The other two options are `VC-WIN64I` (Intel IA64, Itanium) and
+`VC-CE` (Windows CE) are rather uncommon nowadays.
 
 Installing OpenSSL
 ------------------
@@ -262,7 +262,7 @@ for 32bit binaries on 64bit Windows (WOW64).
 #### Installing to a different location
 
 To install OpenSSL to a different location (for example into your home
-directory for testing purposes) run Configure as shown in the following
+directory for testing purposes) run `Configure` as shown in the following
 examples.
 
 On Unix:
@@ -281,10 +281,10 @@ in otherwise unexpected ways.
 Configuration Options
 =====================
 
-There are several options to ./Configure to customize the build (note that
-for Windows, the defaults for `--prefix` and `--openssldir` depend in what
+There are several options to `./Configure` to customize the build (note that
+for Windows, the defaults for `--prefix` and `--openssldir` depend on what
 configuration is used and what Windows implementation OpenSSL is built on.
-More notes on this in NOTES.WIN):
+More notes on this in [NOTES.WIN](NOTES.WIN)):
 
 API Level
 ---------
@@ -307,12 +307,12 @@ If `--api` isn't given, it defaults to the current (minor) OpenSSL version.
 Cross Compile Prefix
 --------------------
 
-    --cross-compile-prefix=PREFIX
+    --cross-compile-prefix=<PREFIX>
 
-The PREFIX to include in front of commands for your toolchain.
+The `<PREFIX>` to include in front of commands for your toolchain.
 
-It is likely to have to end with dash, e.g.  a-b-c- would invoke GNU compiler
-as a-b-c-gcc, etc.  Unfortunately cross-compiling is too case-specific to put
+It is likely to have to end with dash, e.g. `a-b-c-` would invoke GNU compiler
+as `a-b-c-gcc`, etc.  Unfortunately cross-compiling is too case-specific to put
 together one-size-fits-all instructions.  You might have to pass more flags or
 set up environment variables to actually make it work.  Android and iOS cases
 are discussed in corresponding `Configurations/15-*.conf` files.  But there are
@@ -348,9 +348,9 @@ Directories
 
 The name of the directory under the top of the installation directory tree
 (see the `--prefix` option) where libraries will be installed.  By default
-this is "lib". Note that on Windows only static libraries (`*.lib`) will
+this is `lib/`. Note that on Windows only static libraries (`*.lib`) will
 be stored in this location. Shared libraries (`*.dll`) will always be
-installed to the "bin" directory.
+installed to the `bin/` directory.
 
 ### openssldir
 
@@ -406,12 +406,12 @@ If not provided the system library path will be used.
 
 **On Windows:** this is the filename of the zlib library (with or
 without a path).  This flag must be provided if the
-[zlib-dynamic](#zlib-dynamic) option is not also used.  If zlib-dynamic is used
-then this flag is optional and defaults to "ZLIB1" if not provided.
+[zlib-dynamic](#zlib-dynamic) option is not also used. If `zlib-dynamic` is used
+then this flag is optional and defaults to `ZLIB1` if not provided.
 
 **On VMS:** this is the filename of the zlib library (with or without a path).
-This flag is optional and if not provided then "GNV$LIBZSHR", "GNV$LIBZSHR32"
-or "GNV$LIBZSHR64" is used by default depending on the pointer size chosen.
+This flag is optional and if not provided then `GNV$LIBZSHR`, `GNV$LIBZSHR32`
+or `GNV$LIBZSHR64` is used by default depending on the pointer size chosen.
 
 Seeding the Random Generator
 ----------------------------
@@ -436,8 +436,8 @@ Use the [getrandom(2)][man-getrandom] or equivalent system call.
 
 ### devrandom
 
-Use the first device from the DEVRANDOM list which can be opened to read
-random bytes.  The DEVRANDOM preprocessor constant expands to
+Use the first device from the `DEVRANDOM` list which can be opened to read
+random bytes.  The `DEVRANDOM` preprocessor constant expands to
 
     "/dev/urandom","/dev/random","/dev/srandom"
 
@@ -449,7 +449,7 @@ Check for an entropy generating daemon.
 
 ### rdcpu
 
-Use the RDSEED or RDRAND command if provided by the CPU.
+Use the `RDSEED` or `RDRAND` command if provided by the CPU.
 
 ### librandom
 
@@ -468,15 +468,15 @@ at the end of this document.
 Enable and Disable Features
 ---------------------------
 
-Feature options always come in pairs, an option to enable feature xxxx, and
+Feature options always come in pairs, an option to enable feature `xxxx`, and
 and option to disable it:
 
     [ enable-xxxx | no-xxxx ]
 
 Whether a feature is enabled or disabled by default, depends on the feature.
 In the following list, always the non-default variant is documented: if
-feature xxxx is disabled by default then enable-xxxx is documented and
-if feature xxxx is enabled by default then no-xxxx is documented.
+feature `xxxx` is disabled by default then `enable-xxxx` is documented and
+if feature `xxxx` is enabled by default then `no-xxxx` is documented.
 
 ### no-afalgeng
 
@@ -532,8 +532,8 @@ Don't automatically load all supported ciphers and digests.
 Typically OpenSSL will make available all of its supported ciphers and digests.
 For a statically linked application this may be undesirable if small executable
 size is an objective.  This only affects libcrypto.  Ciphers and digests will
-have to be loaded manually using EVP_add_cipher() and EVP_add_digest() if this
-option is used.  This option will force a non-shared build.
+have to be loaded manually using `EVP_add_cipher()` and `EVP_add_digest()`
+if this option is used.  This option will force a non-shared build.
 
 ### no-autoerrinit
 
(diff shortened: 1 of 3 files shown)


levitte:
Two points:

1.  There is a `test/README`, which is more like advice/reminders to ourselves when we write test recipes
2.  I think we need to ask the question "is this helpful for the user?", as the stuff that you're moving is intended for use consumption.  Does it help them to have to go looking in another file when it comes to testing details?

paulidale:
Travis doesn't seem relevant.

paulidale:
With the hyperlinks, going across to the other file is easy.  I'm fine with this.  @levitte, dismiss my review if you feel strongly.

DDvO:
@levitte,

>  1. There is a `test/README`, which is more like advice/reminders to ourselves when we write test recipes.

Ah, yes! I somehow recalled that there was a README already, but then I overlooked it.
I suggest that I'll append that more internal info to the end of the new `test/README.md`, but to do this after this PR and #12109 have been merged because otherwise there would be a heavy merge conflict. 
I suggest moving also the contents of `NOTES.VALGRIND` into `test/README.md` on that occasion.

BTW, would be good if #12109 gets approved soon..

>  2. I think we need to ask the question "is this helpful for the user?", as the stuff that you're moving is intended for use consumption.  Does it help them to have to go looking in another file when it comes to testing details?

I agree with @paulidale that due to the hyperlinks going to the other file should be convenient to do. Moreover, many users will expect info on using the tests in `test/` folder, not within `./INSTALL.md`.

DDvO:
Yesterday I had started cleaning up of references and code/symbol quotation layout in `INSTALL.md`, and I've just completed this, adding a new commit for that.
Can you please re-approve?

levitte:
> hyperlinks

When you get a tarball and there's an `WHATEVER.md`, how do you read it?  I usually run `less` or similar on it.  Hyperlinks aren't the best in that case.  A browser?  If it knows a way to render markdown...

levitte:
> Moreover, many users will expect info on using the tests in `test/` folder, not within `./INSTALL.md`.

I do not know what users you speak for.  To me, that signals that testing isn't that important.

DDvO:
> > hyperlinks
> 
> When you get a tarball and there's an `WHATEVER.md`, how do you read it? I usually run `less` or similar on it. 

Me too.

> Hyperlinks aren't the best in that case. 

Agreed.
An alternative would be to put the same text in both places, which is even more sub-optimal in my view.

> A browser? If it knows a way to render markdown...

If I use Firefox for file browsing with a URL like this: `file:///home/david/openssl/prepare/`
and then click on a `.md` file (or anything else with 'unknown' file extension) that silly thing does not even attempt rendering this as text. Instead, it asks me to open the file in an editor.
Chrome/Chromium does better here: it renders such files as plain text.

Yet when browsing the file via GitHub everything works fine.

DDvO:
> > Moreover, many users will expect info on using the tests in `test/` folder, not within `./INSTALL.md`.
> 
> I do not know what users you speak for. To me, that signals that testing isn't that important.

When I as a user wanna know details about testing, I'd expect them in the `test/` folder README, not buried under `./INSTALL.md`. I did not say that *all* users think this way :wink: 

And we do still stress the importance of running the tests in `INSTALL.md` at the points where it makes most sense:
```
Test OpenSSL

After a successful build, and before installing, the libraries should be tested. Run:

make test                                      # Unix
mms test                                       ! OpenVMS
nmake test                                     # Windows

Warning: you MUST run the tests from an unprivileged account (or disable your privileges temporarily if your platform allows it).

See the file test/README.md for further details.
```
and
```
Running Selected Tests

You can specify a set of tests to be performed using the make variable TESTS.

See the section Running Selected Tests of test/README.md.
```
and
```
Test Failures

If some tests fail, look at the output. There may be reasons for the failure that isn't a problem in OpenSSL itself (like an OS malfunction or a Perl issue).

You may want increased verbosity, that can be accomplished as described in section Test Failures of test/README.md.

If you find a problem with OpenSSL itself, try removing any compiler optimization flags from the CFLAGS line in the Makefile and run make clean; make or corresponding.

To report a bug please open an issue on GitHub, at https://github.com/openssl/openssl/issues.

For more details on how the make variables TESTS can be used, see section Running Selected Tests of test/README.md.
```

Note also that I did not remove any of the sections regarding test but just moved further detail (namely on selecting specific tests and choosing verbosity) to `test/README.md`.

DDvO:
@romen since you brought up the idea for such a change, WDYT about the current state of this PR?

levitte:
> An alternative would be to put the same text in both places, which is even more sub-optimal in my view.

No, please, that's a maintenance issue waiting to happen.

> Yet when browsing the file via GitHub, e.g., via
https://github.com/siemens/openssl/blob/add_test_README.md/INSTALL.md#test-failures
everything works fine.

Yeeeeeah, and the user will intuitively know to go there when they see INSTALL.md in their source tree.  Right.

DDvO:
> > An alternative would be to put the same text in both places, which is even more sub-optimal in my view.
> 
> No, please, that's a maintenance issue waiting to happen.

Good that we agree here.

> > Yet when browsing the file via GitHub, e.g., via
> > https://github.com/siemens/openssl/blob/add_test_README.md/INSTALL.md#test-failures
> > everything works fine.
> 
> Yeeeeeah, and the user will intuitively know to go there when they see INSTALL.md in their source tree. Right.

Sure :wink:

Of course this is not what we can always expect.
Yet whenever a user starts looking at the project via the natural entry point https://github.com/openssl/openssl all those `.md` files are rendered well, including their hyperlinks.

There is the inherent problem after we started switching to `.md` files that *local file* browser support is (so far) bad. Non-clickable references, being discussed here, is just one of the resulting nuisances.

levitte:
> Yet whenever a user starts looking at the project via the natural entry point https://github.com/openssl/openssl all those `.md` files are rendered well, including their hyperlinks.

Please stop.  We announce tarballs when releasing, they are downloadable from https://www.openssl.org/source/, that's what most users are going to go for.  From that point of view, there's nothing "natural" about the github repo.

DDvO:
> > Yet whenever a user starts looking at the project via the natural entry point https://github.com/openssl/openssl all those `.md` files are rendered well, including their hyperlinks.
> 
> Please stop. We announce tarballs when releasing, they are downloadable from https://www.openssl.org/source/, that's what most users are going to go for. From that point of view, there's nothing "natural" about the github repo.

Well, I was not thinking about the way users obtain the sources (I agree that for most of them it will be via the sources linked from https://www.openssl.org/, but could also be via git etc.),
but how new users would approach OpenSSL. When googling for 'openssl', the GitHub repo is already the 3nd hit after the normal project home page and the Wikipedia entry), and more and more folks will be using that for its convenience.

You are right we cannot generally expect people to use the online browsing approach.
BUT, following this logic, we should not have converted *any* of the doc/text files to MarkDown.
As mentioned, their offline file browsing support is not yet wide-spread,
and this pertains not only to hyperlinks but also to their general rendering (layout etc.).
This is an inherent problem, by far not only affecting the changes proposed in this PR.

levitte:
> BUT, following this logic, we should not have converted `any` of the doc/text files to MarkDown.

Why not?  They are still readable as plain text, that's actually the most central feature with MarkDown!

DDvO:
> Why not? They are still readable as plain text, that's actually the most central feature with MarkDown!

Ok, for most part they are pretty well readable.
Yet already the existing (intra- and inter-file) cross-references, are hard to read and use manually, in particular at the beginning of `INSTALL.md`:

```
Table of Contents
=================

 - [Prerequisites](#prerequisites)
 - [Notational Conventions](#notational-conventions)
 - [Quick Installation Guide](#quick-installation-guide)
   - [Building OpenSSL](#building-openssl)
   - [Installing OpenSSL](#installing-openssl)
 - [Configuration Options](#configuration-options)
   - [API Level](#api-level)
   - [Cross Compile Prefix](#cross-compile-prefix)
   - [Build Type](#build-type)
   - [Directories](#directories)
   - [Compiler Warnings](#compiler-warnings)
   - [ZLib Flags](#zlib-flags)
   - [Seeding the Random Generator](#seeding-the-random-generator)
   - [Enable and Disable Features](#enable-and-disable-features)
   - [Displaying configuration data](#displaying-configuration-data)
 - [Installation Steps in Detail](#installation-steps-in-detail)
   - [Configure](#configure-openssl)
   - [Build](#build-openssl)
   - [Test](#test-openssl)
   - [Install](#install-openssl)
 - [Advanced Build Options](#advanced-build-options)
   - [Environment Variables](#environment-variables)
   - [Makefile Targets](#makefile-targets)
   - [Running Selected Tests](#running-selected-tests)
 - [Troubleshooting](#troubleshooting)
   - [Configuration Problems](#configuration-problems)
   - [Build Failures](#build-failures)
   - [Test Failures](#test-failures)
 - [Notes](#notes)
   - [Notes on multi-threading](#notes-on-multi-threading)
   - [Notes on shared libraries](#notes-on-shared-libraries)
   - [Notes on random number generation](#notes-on-random-number-generation)

Prerequisites
=============

To install OpenSSL, you will need:

 * A make implementation
 * Perl 5 with core modules (please read [NOTES.PERL](NOTES.PERL))
 * The Perl module Text::Template (please read [NOTES.PERL](NOTES.PERL))
 * an ANSI C compiler
 * a development environment in the form of development libraries and C
   header files
 * a supported operating system

For additional platform specific requirements, solutions to specific
issues and other details, please read one of these:

 * [NOTES.UNIX](NOTES.UNIX) - notes for Unix like systems
 * [NOTES.VMS](NOTES.VMS) - notes related to OpenVMS
 * [NOTES.WIN](NOTES.WIN) - notes related to the Windows platform
 * [NOTES.DJGPP](NOTES.DJGPP) - building for DOS with DJGPP
 * [NOTES.ANDROID](NOTES.ANDROID) - building for Android platforms (using NDK)
 * [NOTES.VALGRIND](NOTES.VALGRIND) - testing with Valgrind
 * [NOTES.PERL](NOTES.PERL) - some notes on Perl
```

levitte:
It's a matter of taste...  the titles in the table of contents match the headings further down, and it's true that we have separate NOTES files for the non-common stuff.

This still doesn't change my opinion on moving the test specific stuff (which is common for all platforms, I might add)

INSTALL.md
```diff
@@ -566,7 +566,7 @@ This option will be forced if on a platform that does not support CAPI.
 
 ### no-cmp
 
-Don't build support for Certificate Management Protocol (CMP).
+Don't build support for Certificate Management Protocol (CMP) and CRMF.
```
romen, line 569: Nit: for coherence should we expand the CRMF acronym?

INSTALL.md
```diff
@@ -854,8 +855,8 @@ Don't build Time Stamping (TS) Authority support.
 Build with the Undefined Behaviour sanitiser (UBSAN).
 
 This is a developer option only.  It may not work on all platforms and should
-never be used in production environments.  It will only work when used with gcc
-or clang and should be used in conjunction with the `-DPEDANTIC` option
+never be used in production environments.  It will only work when used with
+"gcc" or "clang" and should be used in conjunction with the `-DPEDANTIC` option
```
romen, line 859: Backticks?

INSTALL.md
```diff
@@ -1036,17 +1038,17 @@ for the following:
...
+For example, the following command will not see `-DBAR`:
 
     CPPFLAGS=-DBAR ./config -DCOOKIE
 
 However, the following will see both set variables:
 
     CC=gcc CROSS_COMPILE=x86_64-w64-mingw32- ./config -DCOOKIE
 
-If CC is set, it is advisable to also set CXX to ensure both the C and C++
+If `CC `is set, it is advisable to also set `CXX` to ensure both the C and C++
```
romen, line 1049: Space should be outside the backticks around CC

INSTALL.md
```diff
@@ -1112,54 +1114,55 @@ please use the appropriate form for your platform.
...
 or
 
-    $ perl Configure                                 # All other platforms
+    perl Configure                                 # All other platforms
 
 For the remainder of this text, the Unix form will be used in all examples.
 Please use the appropriate form for your platform.
 
 Pick a suitable name from the list that matches your system.  For most
 operating systems there is a choice between using "cc" or "gcc".
```
romen, line 1143: Backticks for cc and gcc?

INSTALL.md
```diff
@@ -1112,54 +1114,55 @@ please use the appropriate form for your platform.
...
 Unix-like systems.
 
-Configure creates a build file ("Makefile" on Unix, "makefile" on Windows
-and "descrip.mms" on OpenVMS) from a suitable template in Configurations,
-and defines various macros in include/openssl/configuration.h (generated
-from include/openssl/configuration.h.in).
+`Configure` creates a build file (`Makefile` on Unix, `makefile` on Windows
+and `descrip.mms` on OpenVMS) from a suitable template in `Configurations/`,
+and defines various macros in `include/openssl/configuration.h` (generated
+from `nclude/openssl/configuration.h.in`.
```
romen, line 1165: Typo on include

INSTALL.md
```diff
@@ -1221,30 +1224,29 @@ Test OpenSSL
...
+    make install                                   # Unix
+    mms install                                    ! OpenVMS
+    nmake install                                  # Windows
 
 Note that in order to perform the install step above you need to have
 appropriate permissions to write to the installation directory.
 
 The above commands will install all the software components in this
-directory tree under PREFIX (the directory given with `--prefix` or
+directory tree under <PREFIX> (the directory given with `--prefix` or
```
romen, line 1249: Maybe backticks around <PREFIX> or there is the risk some markdown renderer might treat it as inline html

INSTALL.md
```diff
@@ -1312,8 +1314,8 @@ Package builders who want to configure the library for standard locations,
 but have the package installed somewhere else so that it can easily be
 packaged, can use
 
-  $ make DESTDIR=/tmp/package-root install         # Unix
-  $ mms/macro="DESTDIR=TMP:[PACKAGE-ROOT]" install ! OpenVMS
+    make DESTDIR=/tmp/package-root install         # Unix
+    mms/macro="DESTDIR=TMP:[PACKAGE-ROOT]" install ! OpenVMS
```
romen, line 1318: Spurious extra space at the beginning (5 spaces, rather than 4)

INSTALL.md
```diff
@@ -1643,9 +1592,9 @@ build.  Use this command:
     nmake clean                                    # Windows
 
 Assembler error messages can sometimes be sidestepped by using the
-"no-asm" configuration option.
+`no-asm` configuration option.
 
-Compiling parts of OpenSSL with gcc and others with the system compiler will
+Compiling parts of OpenSSL with "gcc" and others with the system compiler will
```
romen, line 1597: Backticks?

NOTES.VALGRIND
```diff
@@ -18,21 +18,21 @@ Requirements
 2. Valgrind installed on the platform
    See: http://valgrind.org/downloads/current.html
 3. OpensSSL compiled
-   See: INSTALL
+   See: [INSTALL.md](INSTALL.md)
```
romen, line 21: [INSTALL.md] is equivalent here (and in similar other places where the link URI and the link label are identical)

NOTES.VALGRIND
```diff
@@ -51,20 +51,20 @@ used to control what capabilities OpenSSL uses.
 As of valgrind-3.15.0 on Linux/x86_64, instructions up to AVX2 are
 supported. Setting the following disables instructions beyond AVX2:
 
-   OPENSSL_ia32cap=":0"
+    OPENSSL_ia32cap=":0"
 
 This variable may need to be set to something different based on the
 processor and Valgrind version you are running tests on. More information
-may be found in docs/man3/OPENSSL_ia32cap.pod.
+may be found in [docs/man3/OPENSSL_ia32cap.pod](docs/man3/OPENSSL_ia32cap.pod).
```
romen, line 58: Just keep the square brackets, no need to repeat the URI

NOTES.VALGRIND
```diff
@@ -51,20 +51,20 @@ used to control what capabilities OpenSSL uses.
...
 
-Additional variables (i.e. VERBOSE and TESTS) are described in the
-INSTALL file in the root of the OpenSSL source tree.
+Additional variables (such as `VERBOSE` and `TESTS`) are described in the
+file [test/README.md]/(test/README.md).
```
romen, line 61: Spurious /

Also, there shouldn't be a need to specify the URI in round brackets.

test/README.md
```diff
@@ -0,0 +1,129 @@
...
+Verbosity on failed (sub-)tests only (`VERBOSE_FAILURES_ONLY` or `VFO`):
+
+    make test VFO=1
+
+Verbosity on failed (sub-)tests, in addition progress on succeeded (sub-)tests
+(`VERBOSE_FAILURES_PROGRESS` or `VFP`):
+
+    make test VFP=1
+
+If you want to run just one or a few specific tests, you can use
```
romen, line 42: Doesn't the rest of this paragraph belong to Running selected tests?

test/README.md
```diff
@@ -0,0 +1,129 @@
...
+                   If this is the first token in the list, the current set of
+                   tests is first assigned the whole set of available tests,
+                   effectively making this token equivalent to
+                   TESTS="alltests -xxx".
+
+Also, all tokens except for "alltests" may have wildcards, such as *.
+(on Unix and Windows, BSD style wildcards are supported, while on VMS,
+it's VMS style wildcards)
+
+### Examples
```
romen, line 99: Some markdown renderers will complain if you mix different styles for the headings.

My preference goes for

    # first level heading
    ## second level heading 
    ### third level heading

as I never remember with the alternative "underlining" style which character is used for which heading level

test/README.md
```diff
@@ -0,0 +1,129 @@
...
+    make TESTS='-99'
+
+Run all tests in test groups 80 to 99 except for tests in group 90:
+
+    make TESTS='[89]? -90'
+
+To stochastically verify that the algorithm that produces uniformly distributed
+random numbers is operating correctly (with a false positive rate of 0.01%):
+
+    ./util/wrap.sh test/bntest -stochastic
```
romen, line 129: @levitte is this example current? Or is it now `wrap.pl`?

romen:
Re: hyperlinks

I might be weird, but even when reading markdown on the terminal, if I find an hyperlink to another file in the Repo I have no problem in pointing my editor/viewer/pager to the file I am pointed to, and actually I prefer to have logical separation among different files (when it is worth it) instead of a monolithic single long file.

E.g. I'd rather avoid to merge the current content of test/README or test/NOTES.VALGRIND into test/README.md and I would rather provide links.

My preference for individual files is that I tend to read the whole file only once (case in which split files is annoying) and many more times I am more likely to be looking for some details about a specific topic. In the latter case having separate files means that I can still use `git grep` (or a recursive grep/ack/ag/pt/whatever if dealing with a source tarball without git info) if memory fails me and I want to quickly check which file covered which topic, and grepping for keywords within that file leads to less false positives than grepping for a pattern in a single monolithic file.

romen:
Re: readability of links

`[foo/bar.md](foo/bar.md)` is equivalent to just `[foo/bar.md]`, but in many instances we are still using the explicit link form: readability without markdown renderers could be improved switching to the implicit form.

INSTALL.md
```diff
@@ -1,4 +1,4 @@
-Build and Install
+yyBuild and Install
```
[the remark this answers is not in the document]
DDvO: Oops, fixed.

INSTALL.md
```diff
@@ -46,7 +46,7 @@ Prerequisites
 
 To install OpenSSL, you will need:
 
- * A make implementation
+ * A "make" implementation
```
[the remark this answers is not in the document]
DDvO: Here I deliberately did not use backticks because this is not a code quotation but refers to 'make' in a broader sense, including also, e.g., `nmake`.

INSTALL.md
```diff
@@ -346,9 +346,9 @@ Directories
 
 The name of the directory under the top of the installation directory tree
 (see the `--prefix` option) where libraries will be installed.  By default
-this is "lib". Note that on Windows only static libraries (`*.lib`) will
+this is `lib/`. Note that on Windows only static libraries (`*.lib`) will
```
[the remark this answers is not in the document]
DDvO: Possibly, but at least for Unix users it immediately makes clear that this is a directory name.

INSTALL.md
```diff
@@ -346,9 +346,9 @@ Directories
...
 be stored in this location. Shared libraries (`*.dll`) will always be
-installed to the "bin" directory.
+installed to the `bin/` directory.
```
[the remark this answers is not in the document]
DDvO: I'm not so sure about this. Even on Windows one can use `/` (as a less well-known alternative to that terrible backslash notation `\`). 
Also note that we already use the forward slash universally in many places, such as `include/openssl/configuration.h`.

INSTALL.md
```diff
@@ -380,7 +380,7 @@ Compiler Warnings
     --strict-warnings
 
 This is a developer flag that switches on various compiler options recommended
-for OpenSSL development.  It only works when using gcc or clang as the compiler.
+for OpenSSL development.  It only works when using "gcc" or "clang" as the compiler.
```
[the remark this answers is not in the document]
DDvO: It was like this already at some places, and I just made it consistent here.
In these occurrences we do not talk about code portions but about compilers with those names.
Maybe it would be even better to drop the `"` around them in all occurrences?

INSTALL.md
```diff
@@ -404,12 +404,12 @@ If not provided the system library path will be used.
 
 **On Windows:** this is the filename of the zlib library (with or
 without a path).  This flag must be provided if the
-[zlib-dynamic](#zlib-dynamic) option is not also used.  If zlib-dynamic is used
-then this flag is optional and defaults to "ZLIB1" if not provided.
+[zlib-dynamic](#zlib-dynamic) option is not also used. If `zlib-dynamic` is used
+then this flag is optional and defaulsts to `ZLIB1` if not provided.
```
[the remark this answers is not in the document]
DDvO: Fixed

INSTALL.md
```diff
@@ -498,7 +498,7 @@ Build with the Address sanitiser.
 
 This is a developer option only.  It may not work on all platforms and should
 never be used in production environments.  It will only work when used with
-gcc or clang and should be used in conjunction with the [no-shared](#no-shared)
+"gcc" or "clang" and should be used in conjunction with the [no-shared](#no-shared)
```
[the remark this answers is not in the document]
DDvO: See above

INSTALL.md
DDvO: Good point - done.

INSTALL.md
DDvO: See above

INSTALL.md
DDvO: Fixed

INSTALL.md
DDvO: See above

INSTALL.md
DDvO: Fixed

INSTALL.md
DDvO: Good thought - done.

INSTALL.md
DDvO: This is just a diff artifact - indentation is 4 as intended.

INSTALL.md
DDvO: See above

NOTES.VALGRIND
DDvO: Thanks for this hint - so I was able to remove quite some clutter from many links.

NOTES.VALGRIND
DDvO: Yep

NOTES.VALGRIND
DDvO: Yes :)

test/README.md
DDvO: Not really - at this point we have just one example of using `TESTS`, mostly to make the point in the following example that it may be combine with the verbosity specifiers which are the topic of this section.

test/README.md
DDvO: Sounds good. Yet when I tried changing to using `#`, `make md-nits` complained:
```
./test/README.md:1: MD003 Header style
./test/README.md:15: MD003 Header style
./test/README.md:67: MD003 Header style
```
So I've kept it as it was because then mdl does not complain and the rendering is fine at least with Firefox.

DDvO:
> Left some comments

Thanks @romen! I've just handled them.
Please check those I left as unresolved with some comment.

DDvO:
> Also, maybe we should rename the old test/README as test/AUTHORING_TESTS.md or something similar, and add a link somewhere in the new test/README.md that is more user oriented.

I would prefer to simply append the current contents of test/README to test/README.md, 
but in any case wait until #12109 is finished (I'm still waiting for the completion of the reviews begun there) to avoid further merge conflicts.

DDvO:
> Re: readability of links
> 
> `[foo/bar.md](foo/bar.md)` is equivalent to just `[foo/bar.md]`, but in many instances we are still using the explicit link form: readability without markdown renderers could be improved switching to the implicit form.

Argh, after doing is in all of `INSTALL.md` I found that although this seems to be possible according to https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links
both Firefox and Chromium don't implement this, so I need to revert this :frowning_face:

DDvO:
> Re: hyperlinks
> 
> I might be weird, but even when reading markdown on the terminal, if I find an hyperlink to another file in the Repo I have no problem in pointing my editor/viewer/pager to the file I am pointed to,

Interesting - with which offline tool does this work for you? 
It does not work for me with `less` run in a Bash shell, 
neither with Emacs within the terminal or on a separate window.

> and actually I prefer to have logical separation among different files (when it is worth it) instead of a monolithic single long file.

Me too. This is another reason I don't like all those details on tests in `INSTALL.md`.
 
> E.g. I'd rather avoid to merge the current content of test/README or test/NOTES.VALGRIND into test/README.md and I would rather provide links.

Fine by me.

> My preference for individual files is that I tend to read the whole file only once (case in which split files is annoying) and many more times I am more likely to be looking for some details about a specific topic. In the latter case having separate files means that I can still use `git grep` (or a recursive grep/ack/ag/pt/whatever if dealing with a source tarball without git info) if memory fails me and I want to quickly check which file covered which topic, and grepping for keywords within that file leads to less false positives than grepping for a pattern in a single monolithic file.

Good point.

INSTALL.md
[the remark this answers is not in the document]
levitte: Isn't "The name of the directory" clear enough?

INSTALL.md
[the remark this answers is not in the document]
levitte: > Even on Windows one can use `/`

That totally depends.  In CMD, you cannot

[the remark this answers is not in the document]
levitte: > Also note that we already use the forward slash universally in many places, such as `include/openssl/configuration.h`.

CL is one of those programs

test/README.md
levitte: It is current.

levitte:
> `[foo/bar.md](foo/bar.md)` is equivalent to just `[foo/bar.md]`

Let's put that to the test, shall we?

- `[foo/bar.md](foo/bar.md)` : [foo/bar.md](foo/bar.md)
- `[foo/bar.md]` : [foo/bar.md]

INSTALL.md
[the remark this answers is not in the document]
DDvO: In this case it would be clear enough.

levitte:
> I would prefer to simply append the current contents of test/README to test/README.md,

Two very different audiences.  I would rather not.

INSTALL.md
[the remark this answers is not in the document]
DDvO: > That totally depends. In CMD, you cannot necessarily do so, it very much depends on the program that's being invoked.

What you cannot do with CMD is path completion involving a `/` (using the Tab key).
Yet you can do, e.g., `cd include/openssl/`.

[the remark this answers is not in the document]
DDvO: So even Windows users should not be too surprised about a trailing `/`, 
and if so, they can learn something new about their wonderful OS :wink:

DDvO:
> > I would prefer to simply append the current contents of test/README to test/README.md,
> 
> Two very different audiences. I would rather not.

All right.

levitte:
> > Re: readability of links
> > 
> > `[foo/bar.md](foo/bar.md)` is equivalent to just `[foo/bar.md]`, but in many instances we are still using the explicit link form: readability without markdown renderers could be improved switching to the implicit form.
> 
> Argh, after doing this in all of `INSTALL.md` I found that although this seems to be possible according to https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links

That's because you misunderstand.  `[link text itself]` is a [reference link](https://spec.commonmark.org/0.29/#reference-link), the actual link is a few lines down, looking like this:

``` markdown
[link text itself]: http://www.reddit.com
```

Note that we have external links in this form

DDvO:
> > > Re: readability of links
> > > `[foo/bar.md](foo/bar.md)` is equivalent to just `[foo/bar.md]`, but in many instances we are still using the explicit link form: readability without markdown renderers could be improved switching to the implicit form.
> > 
> > 
> > Argh, after doing this in all of `INSTALL.md` I found that although this seems to be possible according to https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links
> 
> That's because you misunderstand. `[link text itself]` is a [reference link](https://spec.commonmark.org/0.29/#reference-link), the actual link is a few lines down, looking like this:
> 
> ```gfm
> [link text itself]: http://www.reddit.com
> ```
> 
> Note that we have external links in this form

Got it - thanks @levitte for the clarification!

romen:
I am sorry for the confusion and the extra work I caused about implicit links!

I was clearly mistaken!

test/README.md
romen: mdl must have been setup explicitly to allow mixing the two styles for 3d level: https://github.com/DavidAnson/markdownlint/blob/HEAD/doc/Rules.md#md003---heading-style

This seems deliberate, so forget my comment!

INSTALL.md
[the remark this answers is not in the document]
romen: I don't feel strongly about this, so I will accept whatever you think it's best.

My rule of thumb usually is to use backticks for inline code, paths, and _commands_. That is why I would go for backticks here and elsewhere. But this is just personal preference.

INSTALL.md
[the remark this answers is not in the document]
romen: I was under the impression than in most non-platform-specific sections there was a deliberate intent to avoid using platform-specific things, like path separators.

But then again, as you noted we are not consistent about this and I don't think this should block this PR!

So unless @levitte feels strongly otherwise I would approve it as it is.

INSTALL.md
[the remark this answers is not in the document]
romen: Very good point!

INSTALL.md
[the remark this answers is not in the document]
DDvO: > I don't feel strongly about this, so I will accept whatever you think it's best.

So then I've just generally simplified "gcc" to gcc and "clang" to clang,
using the backticks e.g. `gcc` only where it is used to denote a config item name, which I'd consider as a king of code rather than the compiler name.

DDvO:
> All my feedback was addressed and the leftover items are fine by me if they stay as currently proposed or not.
 
Thanks @romen for your further comments and approval.

I just changed the compiler name references as mentioned above,
and reverted the removal of the `$ ` command prefix that I had done also in this PR because @levitte and others did not like that (as discussed e.g. in #12257).

DDvO:
What still remains open here:

@levitte, you questioned the idea of moving (as I clarified, just the details of) the information about testing out of `INSTALL.md` into `test/README.md`. After having discussed this a bit, is this meanwhile acceptable to you?

DDvO:
I've rebased this PR to solve merge conflicts on `INSTALL.md`
and did a further little improvement on its `Test Failures` section.

Could any OTC member please re-approve.

levitte:
> @levitte, you questioned the idea of moving (as I clarified, just the details of) the information about testing out of `INSTALL.md` into `test/README.md`. After having discussed this a bit, is this meanwhile acceptable to you?

I haven't changed my mind, but am clearly a minority, and I don't care enough to fight harder.  Que sera, sera...

DDvO:
> > @levitte, you questioned the idea of moving (as I clarified, just the details of) the information about testing out of `INSTALL.md` into `test/README.md`. After having discussed this a bit, is this meanwhile acceptable to you?
> 
> I haven't changed my mind, but am clearly a minority, and I don't care enough to fight harder. Que sera, sera...

Thanks - so I'm taking this as an implicit approval, or at least no objection against the two existing ones :)

DDvO:
Rebased, solved conflicts, and merged.
Thanks to all who took part discussing this.

PR #12233, opened 2020-06-22
Restore `EVP_CIPHER_CTX` "running IV" access

kaduk:
In https://github.com/openssl/openssl/pull/10026 it was asserted that there's no need to expose the "running IV" to applications (note: not strictly an *I*V anymore) and that there should only be a need for the "original IV" in applications, and that if there's a need for the running IV that can be added later.

It turns out there is a need for the "running IV", for when a cipher context needs to be serialized, as is done by kTLS (though note that AFAIK kTLS only currently supports AEAD cipher modes, for which there is not a distinction between "original" and "running" IV).

However, the existing APIs that have (historically, prior to #10026) exposed the "running IV" are bad and deliberately undocumented -- they provide a pointer directly into the internal implementation buffer.  When this buffer was in the `EVP_CIPHER_CTX` itself, that was workable (but still unpleasant) since it was in a known location and always available if the `EVP_CIPHER_CTX` was non-NULL, but in a provider-centric model this can only make sense if the returned pointer is into the *provider*'s internal data store, and that is flat-out unsustainable.  For one, it requires that the provider implements things with the IV stored as a contiguous byte array, which is a weird and opaque ABI constraint between libcrypto and provider, and additionally, there's not even a requirement that the provider's internal data store be accessible from the same address space as libcrypto!

To make the best of a bad situation, introduce a pair of new APIs that copy the initial or "running" IV into a caller-supplied buffer, which will eventually let us remove the octet-pointer parameter usage for the IV data, and document the old APIs as deprecated.

Test the new APIs for several cipher modes of AES, confirming that the "running IV" or, as I call it for the new APIs, the "IV state", diverges from the initial IV for CBC and OFB, but not for GCM, CCM, and OCB.

In order to test GCM, adjust `gcm_get_ctx_params()` to use a slightly different check for when it is safe to return a value: previously we could only retrieve an IV value that was generated internally to the library, but not one that was explicitly provided at context initialization.


kaduk:
Note that I did not (yet) do so, but I believe it would be permitted to remove the `EVP_CTRL_GET_IV` `EVP_CIPHER_CTX_ctrl()` operation in favor of the new APIs, since that control has not appeared in a release yet.  I don't know how disruptive such a change would be to the kTLS implementations, though (cc @bsdjhb @borisPis-zz @junjunk )

bsdjhb:
One thing I will note is that kTLS on FreeBSD does indeed support some cipher suites using AES-CBC.  However, the kTLS hooks use the IV directly from the PRF result at the start of a key session rather than fetching it from the context, so I believe they are not affected by this change.

For an example of kTLS fetching the IV, see https://github.com/openssl/openssl/blob/master/ssl/t1_enc.c#L446 where `iv` is a pointer into the session key block (https://github.com/openssl/openssl/blob/master/ssl/t1_enc.c#L313)

kaduk:
Thanks for the insight, John!
It seems that the Linux kTLS path does not use that control either, as of 4ffccf6c4deebd6f3ab5d46234c64f957ff1c1c5, so maybe I should look into removing it.

kaduk:
[travis is quite lit up about the part where we use now-deprecated functions internally; I'll have to look up the incantation to appeast it]

paulidale:
`#define OPENSSL_SUPPRESS_DEPRECATED` before including any headers.

kaduk:
Marking as WIP until I have a chance to address the no-deprecated stuff

kaduk:
I pushed quite a few additional commits and rebased onto the latest master (only one minor merge conflict, over whitespace in core_names.h).  For most of the libcrypto EVP implementations it seemed better to just inline the (old) `EVP_CIPHER_CTX_iv_noconst()` accesses to use the `iv` array in the `EVP_CIPHER_CTX` structure itself, rather than round-tripping to a dummy provider context that was just providing storage and not doing anything; I currently have one commit per file (roughly) in case that's easier to review, but am happy to squash or rejigger how that's distributed.  For the ENGINE implementations I just suppressed the deprecation warnings, and there were a couple places in the tests and library to adapt to use non-deprecated functions, too.  And `EVP_CTRL_GET_IV` is removed, as expected (it seems there was one remaining kTLS consumer that I had missed the previous time I looked).

@levitte please take another look when you get a chance.

kaduk:
(Also, took out of WIP)

levitte:
After what @bsdjhb wrote, I'm at a loss as to why the IV state is still needed.

kaduk:
> After what @bsdjhb wrote, I'm at a loss as to why the IV state is still needed.

The proxmiate cause is that I have need to serialize the cipher state; it happens to be for a non-kTLS purpose, but it's not clear that the exact use-case matters.  (I can talk about it if you want, of course, but is it needed?)  I can do this in 1.1.1 with the existing public APIs.  However, it is quite clear that the current state of master is an API break from 1.1.1, since the data that you can access from the `EVP_CIPHER_CTX_iv()` family of functions is fundamentally different data.  This sort of undocumented API break is exactly what we don't want people to have to deal with in the upgrade from 1.1.1 to 3.0.  In my opinion, this PR provides much clearer documentation on what is happening, and provides a cleaner API that is consistent with the 1.1.1 API and gives us a way to remove an (undocumented!) obscure restriction on the provider implementation and external-facing interface.

Now, I *could* replicate access to that data by caching the last blocks of (plaintext and) ciphertext [0], and then grabbing those off from the side where I currently use the IV state, but that's both needlessly complicated and duplicates what's already being done by the implementation.

[0] The CBC IV state is the last full block of ciphertext, and the OFB state is the keystream used to encrypt the last block of ciphertext, thus recoverable by (plaintext xor ciphertext)

levitte:
Thank you.

A reason I'm asking these things is the unfortunate fact that before 1.1.0, almost *everything* was public. I have always thought of the running IV as internal, and belonging with the backends, and use of it by any other a kinda dubious hack. But if there are legitimate uses of it after all, I can understand.

kaduk:
FWIW I'm going to have to rebase past #12226 in order to get a clean no-deprecated build.
I'll go ahead and squash the fixup commits at the same time.

kaduk:
CI looks clean (the one failed Travis job is just a timeout, that I think we've been seeing fairly regularly).

kaduk:
I'd be interested to know what @p-steuer thinks about ec681a305ee58c70c9a1a319a104ba4df9abc9d1 -- I think the best thing to do would be to teach the appropriate `_get_ctx_params()` routine how to find these values, but I don't know that such a routine currently exists.  Trying to write one from scratch without access to hardware is not something I'm confident I would succeed at, so I just put in this workaround for now.

kaduk:
Rebased past the libcrypto.num conflict.
Stell pending @openssl/otc review.

p-steuer:
> I'd be interested to know what @p-steuer thinks about ec681a3 -- I think the best thing to do would be to teach the appropriate  _get_ctx_params() routine how to find these values, but I don't know that such a routine currently exists. Trying to write one from scratch without access to hardware is not something I'm confident I would succeed at, so I just put in this workaround for now.

I think theres nothing wrong with sync-ing the PROV_CIPHER_CTX for now. If the additional memcpy turns out to affect performance, i ll look into a get_ctx_params solution. 

At least it looks to me like the sync is missing in case of s390x's ccm and gcm though. Lets see if CI agrees.

kaduk:
> At least it looks to me like the sync is missing in case of s390x's ccm and gcm though. Lets see if CI agrees.

The CCM and GCM modes don't have any "IV state" that gets updated after each operation, so the CI should be okay (in that regard at least).  Whether or not I consciously thought about that and decided to not put in "useless" memcpys is another matter, though...

kaduk:
Looking through the travis logs, several builds complain about leaks in test_rand, the GOST engine doesn't build, the boringssl tests are complaining about some things, and test_serializer_deserializer is unhappy on S390x. The test_rand leaks are in codepaths that these changes should not be relevant for, and I've seen other mentions of the GOST and boringssl issues being preexisting, but I'm less sure about the serializer/deserializer bits for S390x.  I guess I can run a separate CI build with no changes to get a better reference point.
Ah, the GOST engine got fixed in the intervening period by #12599 already, and it seems that #12509 fixed the test_rand leaks as well. @t8m do you think I should rebase to get a cleaner picture on the CI results?  I think I still have to do a separate CI run (in a "dummy" PR to know about the S390x status, though.

t8m:
As there is now also a trivial conflict in libcrypto.num, it would be better to rebase I think. However I do not think any of the existing test failures are relevant for this PR.

kaduk:
Okay, rebased once more (past libcrypto.num)

kaduk:
And Travis seems to be consistent with the current baseline.
(The boringssl tests failing is kind of unfortunate; I may try to look at that at some point.)

t8m:
My approval still holds.

t8m:
There is another libcrypto.num conflict, but as that is trivial, please resolve it when merging.

kaduk:
Resolved the libcrypto.num conflict/rebased, and pushed to master.
Thanks for all the review!

PR #11592, opened 2020-04-21
Revert "TEST: make and use a fipsinstall script"

levitte:
Unfortunately, this won't work on MacOS because of system integrity
measures on that platform, which clears DYLD_LIBRARY_PATH before
starting a sub-process executable.

Ref: https://developer.apple.com/library/archive/documentation/Security/Conceptual/System_Integrity_Protection_Guide/RuntimeProtections/RuntimeProtections.html

This reverts commit ae6b654b669638882a6ddce012ff55adc7cf6a82.

PR #12235, opened 2020-06-23
Add --fips-key configuration parameter

richsalz:
This allows a FIPS HMAC key to be specified at configuration time.

### DIFF
--- a/Configurations/descrip.mms.tmpl
+++ b/Configurations/descrip.mms.tmpl
@@ -454,7 +454,7 @@ test : tests
 
 list-tests :
         @ ! {- output_off() if $disabled{tests}; "" -}
-        @ DEFINE SRCTOP {- sourcedir() -}
+        @ DEFINE SRCTOP "$(SRCDIR)"
         @ $(PERL) {- sourcefile("test", "run_tests.pl") -} list
         @ DEASSIGN SRCTOP
         @ ! {- if ($disabled{tests}) { output_on(); } else { output_off(); } "" -}
--- a/Configurations/descrip.mms.tmpl
+++ b/Configurations/descrip.mms.tmpl
@@ -441,6 +441,7 @@ test : tests
         @ ! {- output_off() if $disabled{tests}; "" -}
         DEFINE SRCTOP {- sourcedir() -}
         DEFINE BLDTOP {- builddir() -}
+        DEFINE FIPSKEY {- $config{FIPSKEY} -}
         IF "$(VERBOSE)" .NES. "" THEN DEFINE VERBOSE "$(VERBOSE)"
         $(PERL) {- sourcefile("test", "run_tests.pl") -} $(TESTS)
         DEASSIGN BLDTOP
--- a/Configurations/descrip.mms.tmpl
+++ b/Configurations/descrip.mms.tmpl
@@ -106,6 +106,7 @@ OPTIONS={- $config{options} -}
 CONFIGURE_ARGS=({- join(", ",quotify_l(@{$config{perlargv}})) -})
 SRCDIR={- $config{sourcedir} -}
 BLDDIR={- $config{builddir} -}
+FIPSKEY={- $config{FIPSKEY} -}
 
 # Allow both V and VERBOSE to indicate verbosity.  This only applies
 # to testing.
@@ -439,13 +440,14 @@ all : build_sw build_docs
 test : tests
 {- dependmagic('tests'); -} : build_programs_nodep, build_modules_nodep copy-utils
         @ ! {- output_off() if $disabled{tests}; "" -}
-        DEFINE SRCTOP {- sourcedir() -}
-        DEFINE BLDTOP {- builddir() -}
-        DEFINE FIPSKEY {- $config{FIPSKEY} -}
+        DEFINE SRCTOP "$(SRCDIR)"
+        DEFINE BLDTOP "$(BLDDIR)"
+        DEFINE FIPSKEY "$(FIPSKEY)"
         IF "$(VERBOSE)" .NES. "" THEN DEFINE VERBOSE "$(VERBOSE)"
         $(PERL) {- sourcefile("test", "run_tests.pl") -} $(TESTS)
         DEASSIGN BLDTOP
         DEASSIGN SRCTOP
+        DEASSIGN FIPSKEY
         @ ! {- if ($disabled{tests}) { output_on(); } else { output_off(); } "" -}
         @ WRITE SYS$OUTPUT "Tests are not supported with your chosen Configure options"
         @ ! {- output_on() if !$disabled{tests}; "" -}
--- a/Configure
+++ b/Configure
@@ -238,7 +238,9 @@ my $local_config_envname = 'OPENSSL_LOCAL_CONFIG_DIR';
 
 $config{sourcedir} = abs2rel($srcdir);
 $config{builddir} = abs2rel($blddir);
-$config{FIPSKEY} = '00';
+# echo -n 'holy hand grenade of antioch' | openssl sha256
+$config{FIPSKEY} =
+    'f4556650ac31d35461610bac4ed81b1a181b2d8a43ea2854cbae22ca74560813';
 
 # Collect reconfiguration information if needed
 my @argvcopy=@ARGV;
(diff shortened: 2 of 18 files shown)


slontis:
I am thinking that inside the fipsinstall app the option for '-macopt hexkey:x'  should be made to be optional..
i.e if it is not specified it does a sk_OPENSSL_STRING_push() of the default value from the .h file..
We will probably set the default key to SHA256 of something...

richsalz:
Making "-macopt hexkey:xxxx" be optional and defaulting to FIPSKEY makes sense.  Not thrilled with scannign for "hexkey" option, but willing to do it -- should I?

slontis:
> Making "-macopt hexkey:xxxx" be optional and defaulting to FIPSKEY makes sense. Not thrilled with scannign for "hexkey" option, but willing to do it -- should I?

Yeah I think it is worth doing.. I guess it could be 'key' or 'hexkey'

richsalz:
Actually, it was easy to use the default FIPSKEY; additional commit pushed.

slontis:
Could you make the default key..
f4556650ac31d35461610bac4ed81b1a181b2d8a43ea2854cbae22ca74560813

i.e: SHA256("holy hand grenade of antioch")

Configure
```diff
@@ -238,7 +238,8 @@ my $local_config_envname = 'OPENSSL_LOCAL_CONFIG_DIR';
 
 $config{sourcedir} = abs2rel($srcdir);
 $config{builddir} = abs2rel($blddir);
-$config{FIPSKEY} = '00';
+$config{FIPSKEY} =
+    'f4556650ac31d35461610bac4ed81b1a181b2d8a43ea2854cbae22ca74560813';
```
mspncp, line 242: May I suggest adding the adding the reference as a comment?
```suggestion
$config{FIPSKEY} =
#   SHA256("holy hand grenade of antioch")
    'f4556650ac31d35461610bac4ed81b1a181b2d8a43ea2854cbae22ca74560813';
```

richsalz: Updated commit pushed.

Configure
```diff
@@ -914,6 +917,16 @@ while (@argvcopy)
                             push @seed_sources, $x;
                             }
                         }
+                elsif (/^--fips-key=(.*)$/)
+                        {
+                        $user{FIPSKEY}=lc($1);
+                        die "Non-hex character in FIPS key\n"
+                           if $user{FIPSKEY} =~ /[^a-f0-9]/;
```
mspncp, line 924: Maybe allow uppercase hex letters, too?

test/recipes/03-test_fipsinstall.t
```diff
@@ -27,75 +27,76 @@ plan skip_all => "Test only supported in a fips build" if disabled("fips");
 plan tests => 12;
 
 my $infile = bldtop_file('providers', platform->dso('fips'));
+my $fipskey = $ENV{FIPSKEY} // '00';
```
mspncp, line 30: _(Shouldn't this default to the holy handgrenade, too?)_

Configure
richsalz: line 922 converts to lowercase.

mspncp: Ah, ok, I overlooked that.  Two letters, very unix-ish.

test/recipes/03-test_fipsinstall.t
richsalz: I could go either way.  As it is right now, it catches a "make" error where the env var isn't set.

mspncp: I have no preference, so just leave it for the moment. Maybe @slontis has an opinion about it when he reapproves.

slontis: As long as it runs I dont really mind..

test/recipes/03-test_fipsinstall.t
```diff
@@ -27,75 +27,76 @@ plan skip_all => "Test only supported in a fips build" if disabled("fips");
...
 
 # fail if no module name
 ok(!run(app(['openssl', 'fipsinstall', '-out', 'fips.cnf', '-module',
              '-provider_name', 'fips',
-             '-macopt', 'digest:SHA256', '-macopt', 'hexkey:00',
+             '-macopt', 'digest:SHA256', '-macopt', "hexkey:$fipskey",
```
slontis, line 35: Maybe add one more test that removes most of the options..
I am starting to think using macopt was a bad idea (and the digest should work as a default also).

slontis: I can do that in a seperate PR if you dont want to do it.
It would probably be easier just to handle -hexkey & -digest

richsalz: Or you can push a commit to this PR if that's simpler.

slontis: That normally ends up with someone forgetting to pull and then doing a force push :)
Will do in a seperate PR :)

slontis:
It looks like this change has broken travis (ignore the test_app one).

Configurations/descrip.mms.tmpl
```diff
@@ -441,6 +441,7 @@ test : tests
         @ ! {- output_off() if $disabled{tests}; "" -}
         DEFINE SRCTOP {- sourcedir() -}
         DEFINE BLDTOP {- builddir() -}
+        DEFINE FIPSKEY {- $config{FIPSKEY} -}
```
levitte, line 444: With all the other build files, you create a make variable, which you refer to for the tests.  the same should be done here.

A make variable looks exactly the same in `descrip.mms`, i.e.:

```
FIPSKEY={- $config{FIPSKEY} -}
```

The DEFINE line would then look quite unsurprising:

```
        DEFINE FIPSKEY "$(FIPSKEY)"
```

levitte:
> It looks like this change has broken travis (ignore the test_app one).

That's because the `test-fipsinstall` change needs to be reproduced in these files as well:

``` console
$ git grep -l fipsinstall test/recipes
test/recipes/03-test_fipsinstall.t
test/recipes/30-test_acvp.t
test/recipes/30-test_evp.t
test/recipes/30-test_evp_fetch_prov.t
test/recipes/80-test_ssl_new.t
test/recipes/80-test_ssl_old.t
test/recipes/90-test_sslapi.t
```

(one of the perks with code copying, you get to do the same fix in multiple places...)

Configurations/descrip.mms.tmpl
richsalz: Okay.  I fixed SRCTOP/BLDTOP while I was there so that nobody else is led astray in the future :)

richsalz:
Is it worth putting those (two lines, seven places) into a new single utility routine?

I think removing the macopt/hexkey arguments, since fipsinstall now defaults to the configured key, is the way to go for these other keys (keep it in fipsinstall for flexibility/testing), so that's what I'm doing.

test/recipes/03-test_fipsinstall.t
richsalz: I removed all uses of -macopt and -mac_name, except for fipsinstall.

richsalz:
Needs a re-review, non-trivial changes made.

Configurations/descrip.mms.tmpl
```diff
@@ -439,13 +440,14 @@ all : build_sw build_docs
 test : tests
 {- dependmagic('tests'); -} : build_programs_nodep, build_modules_nodep copy-utils
         @ ! {- output_off() if $disabled{tests}; "" -}
-        DEFINE SRCTOP {- sourcedir() -}
-        DEFINE BLDTOP {- builddir() -}
-        DEFINE FIPSKEY {- $config{FIPSKEY} -}
+        DEFINE SRCTOP "$(SRCDIR)"
```
mspncp, line 443: What about [this location here](https://github.com/openssl/openssl/blob/018c5a1a9d/Configurations/descrip.mms.tmpl#L455-L457)? Should it be changed accordingly?

```
list-tests :
        @ ! {- output_off() if $disabled{tests}; "" -}
        @ DEFINE SRCTOP {- sourcedir() -}
```

https://github.com/openssl/openssl/blob/018c5a1a9d/Configurations/descrip.mms.tmpl#L457

richsalz: It's come to this, cleanup up MMS files. :)

slontis:
> Is it worth putting those (two lines, seven places) into a new single utility routine?
> 
> I think removing the macopt/hexkey arguments, since fipsinstall now defaults to the configured key, is the way to go for these other keys (keep it in fipsinstall for flexibility/testing), so that's what I'm doing.

@levitte tried doing this in #11592 but ran into some problems on mac.

richsalz:
No, I meant creating like "call_fipsinstall()" as a perl subroutine.

levitte:
Ah, so a perl fipsinstall *module* basically?  That's doable...

richsalz:
> Ah, so a perl fipsinstall module basically? That's doable...

Yes, but no longer necessary since most invocations use default values.

slontis:
Thanks Rich. merged to master.

PR #12236, opened 2020-06-23
Add evp_test fixes.

slontis:
Changed many tests so they also test fips (and removed 'availablein = default' from some tests).
Seperated the monolithic evppkey.txt file into smaller maintainable groups.
Changed the availablein option so it must be first - this then skips the entire test before any fetching happens.
Changed the code so that all the OPENSSL_NO_XXXX tests are done in code via methods such as is_cipher_disabled(alg), before the fetch happens.

##### Checklist


test/evp_test.c
```diff
@@ -337,21 +335,159 @@ typedef struct digest_data_st {
     int pad_type;
 } DIGEST_DATA;
 
+
+/* Done as macros to avoid compiler warnings about unused functions  */
+#define STR_STARTS_WITH(str, pre) strncmp(pre, str, strlen(pre)) == 0
```
richsalz, line 340: I don't understand the comment.  And shouldn't that by strncasecmp?

slontis: If I do this as a function then I need to do #if defined(X) || defined{Y} || define(Z) || .......
otherwise it will produce a error that the function is not used in the case where a user turns everything off with no- options.

richsalz: Still don't get it, but that's okay.

but strncasecmp?

slontis: Will do..

If you search for STR_STARTS_WITH() you will see each one is inside a #ifdef e.g:

```
#ifdef OPENSSL_NO_POLY1305
    if (STR_STARTS_WITH(name, "Poly1305"))
        return 1;
#endif
```
if STR_STARTS_WITH is a function instead of a macro then the function would be unused if all the OPENSSL_NO_POLY1305 lines are undefined.

slontis: So the function would need to be wrapped with a nasty #ifdef to pick up this case.

richsalz: or a dummy case provided.  shrug.  doesn't matter to me.

crypto/sm2/sm2_crypt.c
```diff
@@ -366,7 +378,12 @@ int sm2_decrypt(const EC_KEY *key,
         goto done;
     }
 
-    if (!EVP_DigestInit(hash, digest)
+    fetched_digest = EVP_MD_fetch(libctx, EVP_MD_name(digest), propq);
```
mattcaswell, line 381: This doesn't look correct to me. The passed digest may already be a "fetched" digest (e.g. if EVP_PKEY_CTRL_MD has been used to set a custom digest). Should the fix instead be in `pkey_sm2_encrypt`, to not use the non-fetched `EVP_sm3()`? Or is it actually the fact that SM2 has not been converted to be provider aware that is at the root of the problem?

test/evp_test.c
```diff
@@ -2644,40 +2477,31 @@ typedef struct pkey_kdf_data_st {
...
-    if (kdata->ctx == NULL) {
-        OPENSSL_free(kdata);
-        return 0;
-    }
-    if (EVP_PKEY_derive_init(kdata->ctx) <= 0) {
-        EVP_PKEY_CTX_free(kdata->ctx);
-        OPENSSL_free(kdata);
-        return 0;
-    }
+    kdata->ctx = EVP_PKEY_CTX_new_id_with_libctx(kdf_nid, libctx, NULL);
```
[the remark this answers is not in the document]
slontis: See below comment.

crypto/evp/pmeth_lib.c
```diff
@@ -411,6 +411,12 @@ EVP_PKEY_CTX *EVP_PKEY_CTX_new_id(int id, ENGINE *e)
     return int_ctx_new(NULL, NULL, e, NULL, NULL, id);
 }
 
+EVP_PKEY_CTX *EVP_PKEY_CTX_new_id_with_libctx(int id, OPENSSL_CTX *libctx,
```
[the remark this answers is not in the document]
slontis: This is related to the issue you are currently looking into .. i.e the EVP_PKEY KDF problem.
i.e. We have a legacy EVP_PKEY so it needs to use the id to create it, but when we do the fetch it needs the library context.

include/openssl/evp.h
```diff
@@ -1258,6 +1258,11 @@ int PKCS5_PBKDF2_HMAC_SHA1(const char *pass, int passlen,
 int PKCS5_PBKDF2_HMAC(const char *pass, int passlen,
                       const unsigned char *salt, int saltlen, int iter,
                       const EVP_MD *digest, int keylen, unsigned char *out);
+int PKCS5_PBKDF2_HMAC_with_libctx(const char *pass, int passlen,
```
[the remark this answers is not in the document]
slontis: Ok so I backed out this change..
You do realize this stuff is entangled in the PKCS12 code also..
I have made a internal version of PKCS5_PBKDF2_HMAC_with_libctx as it is called form the providers scrypt.

test/evp_test.c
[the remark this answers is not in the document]
slontis: I have got rid of this API.. For now the tests have changed to just use OPENSSL_CTX_set0_default()
for these cases..

crypto/sm2/sm2_crypt.c
slontis: SM2 is still a legacy key..
EVP_PKEY_CTX_ctrl_str() will go thru the legacy path and call EVP_PKEY_CTX_md() which
calls EVP_get_digestbyname()
it then calls EVP_PKEY_CTX_ctrl() and hits the legacy part of that which calls 
EVP_PKEY_CTRL_MD..

I will change pkey_sm2_encrypt() and pkey_sm2_decrypt() to do the fetch.

crypto/evp/pmeth_lib.c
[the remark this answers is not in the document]
slontis: Added a call to OPENSSL_CTX_set0_default() instead.

slontis:
rebased with API renames + new evp_test files + broke aes_cbc_cts into 2 parts..

test/evp_test.c
```diff
@@ -3569,3 +3578,171 @@ int setup_tests(void)
...
     return 1;
 }
+
+void cleanup_tests(void)
+{
+    OSSL_PROVIDER_unload(prov_null);
+    OPENSSL_CTX_free(libctx);
+}
+
+/* Done as macros to avoid compiler warnings about unused functions  */
```
mattcaswell, line 3588: I don't understand this comment

mattcaswell:
Oh....except the travis failure looks relevant.

test/evp_test.c
slontis: I will delete the comment since @richsalz didnt understand it either :)

slontis:
Fixed memory leak in evp_test.
Removed a comment.

slontis:
@mattcaswell this is passing the CI loop now.

crypto/sm2/sm2_pmeth.c
```diff
@@ -135,16 +139,24 @@ static int pkey_sm2_encrypt(EVP_PKEY_CTX *ctx,
             return 1;
     }
 
-    return sm2_encrypt(ec, md, in, inlen, out, outlen);
+    fetched_md = EVP_MD_fetch(libctx, EVP_MD_name(md), 0);
```
levitte, line 142: Er...  if `md = dctx->md`, isn't it already fetched?

slontis: Not sure: it does this.. How does this tell you its fetched?    
  case EVP_PKEY_CTRL_MD:
        smctx->md = p2;
        return 1;

slontis:
Did a rebase. Not much changed apart from any new changes to .txt files had to be moved into their new .txt file

slontis:
Rebased again with no changes.

slontis:
Thanks. Merged to master.

PR #12250, opened 2020-06-23
WIP: Improve documentation of dispatched functions

mspncp:
This pull request continues work started in pull request #12222.

~~The branch is currently still based on #12222, only the last commit is relevant at the moment.~~

### DIFF
--- a/doc/man7/provider-base.pod
+++ b/doc/man7/provider-base.pod
@@ -3,172 +3,202 @@
 =head1 NAME
 
 provider-base
-- The basic OpenSSL library E<lt>-E<gt> provider functions
+- The basic OpenSSL library E<lt>-E<gt> provider dispatch interface
 
 =head1 SYNOPSIS
 
  #include <openssl/core_dispatch.h>
 
  /*
-  * None of these are actual functions, but are displayed like this for
-  * the function signatures for functions that are offered as function
-  * pointers in OSSL_DISPATCH arrays.
+  * The following function prototypes don't represent actual C functions
+  * but rather the signatures of so-called 'dispatched' functions.
+  * To indicate the difference, curly {braces} have been added around
+  * the function names.
+  *
+  * See the L<Dispatched Functions> section for an explanation how to
+  * interpret those prototypes.
   */
 
  /* Functions offered by libcrypto to the providers */
- const OSSL_ITEM *core_gettable_params(const OSSL_CORE_HANDLE *handle);
- int core_get_params(const OSSL_CORE_HANDLE *handle, OSSL_PARAM params[]);
- int core_thread_start(const OSSL_CORE_HANDLE *handle,
-                       OSSL_thread_stop_handler_fn handfn);
- OPENSSL_CORE_CTX *core_get_library_context(const OSSL_CORE_HANDLE *handle);
- void core_new_error(const OSSL_CORE_HANDLE *handle);
- void core_set_error_debug(const OSSL_CORE_HANDLE *handle,
-                           const char *file, int line, const char *func);
- void core_vset_error(const OSSL_CORE_HANDLE *handle,
+ const OSSL_ITEM * {core_gettable_params}(const OSSL_CORE_HANDLE *handle);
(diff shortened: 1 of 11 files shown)


doc/man7/provider-base.pod
```diff
@@ -10,165 +10,193 @@ provider-base
...
+ {CRYPTO_secure_free}             OSSL_FUNC_CRYPTO_SECURE_FREE
+ {CRYPTO_secure_clear_free}       OSSL_FUNC_CRYPTO_SECURE_CLEAR_FREE
+ {CRYPTO_secure_allocated}        OSSL_FUNC_CRYPTO_SECURE_ALLOCATED
+ {BIO_new_file}                   OSSL_FUNC_BIO_NEW_FILE
+ {BIO_new_mem_buf}                OSSL_FUNC_BIO_NEW_MEMBUF
+ {BIO_read_ex}                    OSSL_FUNC_BIO_READ_EX
+ {BIO_free}                       OSSL_FUNC_BIO_FREE
+ {BIO_vprintf}                    OSSL_FUNC_BIO_VPRINTF
+ {OPENSSL_cleanse}                OSSL_FUNC_OPENSSL_CLEANSE
+ {OSSL_SELF_TEST_set_callback}    OSSL_FUNC_SELF_TEST_CB
```
mspncp, line 159: This is an error: instead of the exported symbol `OSSL_SELF_TEST_set_callback` this line should contain the name of the dispatched function:

```
 {self_set_cb}         OSSL_FUNC_SELF_TEST_CB
```

doc/man7/provider-base.pod
```diff
@@ -156,7 +156,7 @@ provider):
  {BIO_free}                       OSSL_FUNC_BIO_FREE
  {BIO_vprintf}                    OSSL_FUNC_BIO_VPRINTF
  {OPENSSL_cleanse}                OSSL_FUNC_OPENSSL_CLEANSE
- {OSSL_SELF_TEST_set_callback}    OSSL_FUNC_SELF_TEST_CB
+ {self_test_cb}                   OSSL_FUNC_SELF_TEST_CB
```
mspncp, line 159: We could also call it `SELF_TEST_set_callback` (without OSSL_) which corresponds to 

- OSSL_FUNC_SELF_TEST_SET_CALLBACK
- OSSL_FUNC_SELF_TEST_set_callback()

levitte: This set of functions is a bit special, ass they are upcalls that are direct pointers to public OpenSSL functions of the exact same name.  For example, the one just above, `OPENSSL_cleanse`, is an actual OpenSSL function.  So, following the same scheme with this callback setter, it should be:

```
{OSSL_SELF_TEST_set_callback}    OSSL_FUNC_OSSL_SELF_TEST_SET_CALLBACK
```

mspncp: I understand that, but we are not forced to name the fetched upcall _exactly_ the same way as the original function. The old types `CRYPTO`, `BIO`, etc. are different, they did not have an `SSL_` prefix enforced yet, contrary to `SELF_TEST`. Do we really want those prefixes to add up?

The helper function would become

```
OSSL_FUNC_OSSL_SELF_TEST_set_callback()

```
which looks a bit ridiculous, instead of

```
OSSL_FUNC_SELF_TEST_set_callback()
```

which would be sufficient IMHO. So I would prefer naming it

```
{SELF_TEST_set_callback}    OSSL_FUNC_SELF_TEST_SET_CALLBACK
```

mspncp: Thinking in C++, the `SSL_` prefix is a global namespace, so it shouldn't accumulate.

mspncp:
Rebased on master to drop  the dependency on PR #12222

doc/man7/provider-base.pod
mspncp: I reverted the renaming of the self test callback for the moment. The names need some straightening out, but I'll keep that for another pr.

t8m:
@mspncp is this still relevant?

PR #12251, opened 2020-06-23
Add support to zeroize plaintext in S3 record layer

melshuber:
Some applications want even all plaintext copies beeing
zeroized. However, currently plaintext residuals are kept in rbuf
within the s3 record layer.

This patch add the option SSL_OP_CLEANSE_PLAINTEXT to its friends to
optionally enable cleansing of decrypted plaintext data.

* This patch is a followup of the discussion with Matt on openssl-users
https://mta.openssl.org/pipermail/openssl-users/2020-June/012604.html

##### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1060,8 +1060,8 @@ OpenSSL 3.0
  * The SSL option SSL_OP_CLEANSE_PLAINTEXT is introduced. If that
    option is set, openssl cleanses (zeroize) plaintext bytes from
    internal buffers after delivering them to the application. Note,
-   the application is still responsible to cleanse other copies (e.g.:
-   data received by SSL_read(3)).
+   the application is still responsible for cleansing other copies
+   (e.g.: data received by SSL_read(3)).
 
    *Martin Elshuber*
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1057,6 +1057,12 @@ OpenSSL 3.0
 
    *Boris Pismenny*
 
+ * The SSL option SSL_OP_CLEANSE_PLAINTEXT is introduced. If that
+   option is set, openssl cleanses plaintext bytes after delivering
+   them to the application.
+
+   *Martin Elshuber*
+
 OpenSSL 1.1.1
 -------------
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1058,8 +1058,10 @@ OpenSSL 3.0
    *Boris Pismenny*
 
  * The SSL option SSL_OP_CLEANSE_PLAINTEXT is introduced. If that
-   option is set, openssl cleanses plaintext bytes after delivering
-   them to the application.
+   option is set, openssl cleanses (zeroize) plaintext bytes from
+   internal buffers after delivering them to the application. Note,
+   the application is still responsible to cleanse other copies (e.g.:
+   data received by SSL_read(3)).
 
    *Martin Elshuber*
 
--- a/doc/man3/SSL_CTX_set_options.pod
+++ b/doc/man3/SSL_CTX_set_options.pod
@@ -267,12 +267,12 @@ clients.
 
 =item SSL_OP_CLEANSE_PLAINTEXT
 
-By default TLS connections keep a copy of received plaintext data in a
-static buffer until it is overwritten by the next portion of
-data. When enabling SSL_OP_CLEANSE_PLAINTEXT this plaintext is
-cleansed by calling OPENSSL_cleanse(3) after passing data to the
-application. Data is also cleansed when releasing the connection
-(eg. L<SSL_free(3)>).
+By default TLS connections keep a copy of received plaintext
+application data in a static buffer until it is overwritten by the
+next portion of data. When enabling SSL_OP_CLEANSE_PLAINTEXT
+deciphered application data is cleansed by calling OPENSSL_cleanse(3)
+after passing data to the application. Data is also cleansed when
+releasing the connection (eg. L<SSL_free(3)>).
 
 Since OpenSSL only cleanses internal buffers, the application is
 still responsible for cleansing all other buffers. Most notable this
--- a/doc/man3/SSL_CTX_set_options.pod
+++ b/doc/man3/SSL_CTX_set_options.pod
@@ -265,6 +265,14 @@ functionality is not required. Those applications can turn this feature off by
 setting this option. This is a server-side opton only. It is ignored by
 clients.
 
+=item SSL_OP_CLEANSE_PLAINTEXT
+
+By default TLS connections keep a copy of received plaintext data in a
+static buffer until it is overwritten by followup data. When enabling
+SSL_OP_CLEANSE_PLAINTEXT this plaintext is cleansed by calling
+OPENSSL_cleanse(3) after passing data to the application. Data is also
+cleansed when releasing the connection (eg. L<SSL_free(3)>).
+
 =back
 
 The following options no longer have any effect but their identifiers are
--- a/doc/man3/SSL_CTX_set_options.pod
+++ b/doc/man3/SSL_CTX_set_options.pod
@@ -275,9 +275,9 @@ after passing data to the application. Data is also cleansed when
 releasing the connection (eg. L<SSL_free(3)>).
 
 Since OpenSSL only cleanses internal buffers, the application is still
-responsible for cleansing all other buffers. Most notable this applies
-to buffers passed to functions like L<SSL_read(3)>,
-L<SSL_peek(3)|SSL_read(3)> but also like L<SSL_write(3)>.
+responsible for cleansing all other buffers. Most notably, this
+applies to buffers passed to functions like L<SSL_read(3)>,
+L<SSL_peek(3)> but also like L<SSL_write(3)>.
 
 =back
 
--- a/doc/man3/SSL_CTX_set_options.pod
+++ b/doc/man3/SSL_CTX_set_options.pod
@@ -268,10 +268,16 @@ clients.
 =item SSL_OP_CLEANSE_PLAINTEXT
 
 By default TLS connections keep a copy of received plaintext data in a
-static buffer until it is overwritten by followup data. When enabling
-SSL_OP_CLEANSE_PLAINTEXT this plaintext is cleansed by calling
-OPENSSL_cleanse(3) after passing data to the application. Data is also
-cleansed when releasing the connection (eg. L<SSL_free(3)>).
+static buffer until it is overwritten by the next portion of
+data. When enabling SSL_OP_CLEANSE_PLAINTEXT this plaintext is
+cleansed by calling OPENSSL_cleanse(3) after passing data to the
+application. Data is also cleansed when releasing the connection
+(eg. L<SSL_free(3)>).
+
+Since OpenSSL only cleanses internal buffers, the application is
+still responsible for cleansing all other buffers. Most notable this
+applies to buffers passed to functions like L<SSL_read(3)>,
+SSL_peek(3) but also like L<SSL_write(3)>.
 
 =back
 
--- a/doc/man3/SSL_CTX_set_options.pod
+++ b/doc/man3/SSL_CTX_set_options.pod
@@ -274,10 +274,10 @@ deciphered application data is cleansed by calling OPENSSL_cleanse(3)
 after passing data to the application. Data is also cleansed when
 releasing the connection (eg. L<SSL_free(3)>).
 
-Since OpenSSL only cleanses internal buffers, the application is
-still responsible for cleansing all other buffers. Most notable this
-applies to buffers passed to functions like L<SSL_read(3)>,
-SSL_peek(3) but also like L<SSL_write(3)>.
+Since OpenSSL only cleanses internal buffers, the application is still
+responsible for cleansing all other buffers. Most notable this applies
+to buffers passed to functions like L<SSL_read(3)>,
+L<SSL_peek(3)|SSL_read(3)> but also like L<SSL_write(3)>.
 
 =back
 
--- a/ssl/record/rec_layer_d1.c
+++ b/ssl/record/rec_layer_d1.c
@@ -74,6 +74,8 @@ void DTLS_RECORD_LAYER_clear(RECORD_LAYER *rl)
 
     while ((item = pqueue_pop(d->processed_rcds.q)) != NULL) {
         rdata = (DTLS1_RECORD_DATA *)item->data;
+        if (rl->s->options & SSL_OP_CLEANSE_PLAINTEXT)
+            OPENSSL_cleanse(rdata->rbuf.buf, rdata->rbuf.len);
         OPENSSL_free(rdata->rbuf.buf);
         OPENSSL_free(item->data);
         pitem_free(item);
@@ -81,6 +83,8 @@ void DTLS_RECORD_LAYER_clear(RECORD_LAYER *rl)
 
     while ((item = pqueue_pop(d->buffered_app_data.q)) != NULL) {
         rdata = (DTLS1_RECORD_DATA *)item->data;
+        if (rl->s->options & SSL_OP_CLEANSE_PLAINTEXT)
+            OPENSSL_cleanse(rdata->rbuf.buf, rdata->rbuf.len);
         OPENSSL_free(rdata->rbuf.buf);
         OPENSSL_free(item->data);
         pitem_free(item);
@@ -120,18 +124,14 @@ void DTLS_RECORD_LAYER_set_write_sequence(RECORD_LAYER *rl, unsigned char *seq)
 static int dtls1_copy_record(SSL *s, pitem *item)
 {
     DTLS1_RECORD_DATA *rdata;
-    SSL3_BUFFER *rbuf;
 
     rdata = (DTLS1_RECORD_DATA *)item->data;
-    rbuf = &s->rlayer.rbuf;
 
-    if (s->options & SSL_OP_CLEANSE_PLAINTEXT)
-        OPENSSL_cleanse(rbuf->buf, rbuf->len);
-    SSL3_BUFFER_release(rbuf);
+    SSL3_BUFFER_release(&s->rlayer.rbuf);
 
     s->rlayer.packet = rdata->packet;
     s->rlayer.packet_length = rdata->packet_length;
-    memcpy(rbuf, &(rdata->rbuf), sizeof(SSL3_BUFFER));
+    memcpy(&s->rlayer.rbuf, &(rdata->rbuf), sizeof(SSL3_BUFFER));
     memcpy(&s->rlayer.rrec, &(rdata->rrec), sizeof(SSL3_RECORD));
 
     /* Set proper sequence number for mac calculation */
--- a/ssl/record/rec_layer_d1.c
+++ b/ssl/record/rec_layer_d1.c
@@ -120,14 +120,18 @@ void DTLS_RECORD_LAYER_set_write_sequence(RECORD_LAYER *rl, unsigned char *seq)
 static int dtls1_copy_record(SSL *s, pitem *item)
 {
     DTLS1_RECORD_DATA *rdata;
+    SSL3_BUFFER *rbuf;
 
     rdata = (DTLS1_RECORD_DATA *)item->data;
+    rbuf = &s->rlayer.rbuf;
 
-    SSL3_BUFFER_release(&s->rlayer.rbuf);
+    if (s->options & SSL_OP_CLEANSE_PLAINTEXT)
+        OPENSSL_cleanse(rbuf->buf, rbuf->len);
+    SSL3_BUFFER_release(rbuf);
 
     s->rlayer.packet = rdata->packet;
     s->rlayer.packet_length = rdata->packet_length;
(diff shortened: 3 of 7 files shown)


CHANGES.md
```diff
@@ -1057,6 +1057,12 @@ OpenSSL 3.0
 
    *Boris Pismenny*
 
+ * The SSL option SSL_OP_CLEANSE_PLAINTEXT is introduced. If that
+   option is set, openssl cleanses plaintext bytes after delivering
```
mattcaswell, line 1061: Probably we should clarify that this means we cleanse *internal* buffers. The application is still responsible for cleansing its own buffers.

ssl/record/rec_layer_s3.c
```diff
@@ -1477,6 +1477,9 @@ int ssl3_read_bytes(SSL *s, int type, int *recvd_type, unsigned char *buf,
                 n = len - totalbytes;
 
             memcpy(buf, &(rr->data[rr->off]), n);
+            if (s->options & SSL_OP_CLEANSE_PLAINTEXT)
```
mattcaswell, line 1480: This shouldn't be done if `peek` is set

test/sslapitest.c
```diff
@@ -1582,6 +1582,88 @@ static int test_large_message_tls_read_ahead(void)
...
+    static unsigned char sbuf[16000];
+
+    if (!TEST_ptr(certbio = BIO_new_file(cert, "r")))
+        goto end;
+    chaincert = PEM_read_bio_X509(certbio, NULL, NULL, NULL);
+    BIO_free(certbio);
+
+    certbio = NULL;
+    if (!TEST_ptr(chaincert))
+        goto end;
```
mattcaswell, line 1605: The block above seems unnecessary. `chaincert` is never used.

test/sslapitest.c
```diff
@@ -1582,6 +1582,88 @@ static int test_large_message_tls_read_ahead(void)
...
+
+    if (!TEST_true(create_ssl_connection(serverssl, clientssl,
+                                        SSL_ERROR_NONE)))
+        goto end;
+
+    for (i = 0; i < sizeof(cbuf); i++) {
+        cbuf[i] = i & 0xff;
+    }
+
+    if (!TEST_true(SSL_write(clientssl, cbuf, sizeof(cbuf)) == sizeof(cbuf)))
```
mattcaswell, line 1629: `TEST_int_eq` seems more appropriate than `TEST_true` here.

test/sslapitest.c
```diff
@@ -1582,6 +1582,88 @@ static int test_large_message_tls_read_ahead(void)
...
+        goto end;
+
+    while ((err = SSL_read(serverssl, &sbuf, sizeof(sbuf))) != sizeof(sbuf)) {
```
mattcaswell, line 1632: The loop here seems unnecessary. We know the data is going to be there because there is no network being used here.

mattcaswell, line 1632: I think we should do a peek first,

mattcaswell:
Please can you submit a CLA:

doc/man3/SSL_CTX_set_options.pod
```diff
@@ -265,6 +265,14 @@ functionality is not required. Those applications can turn this feature off by
 setting this option. This is a server-side opton only. It is ignored by
 clients.
 
+=item SSL_OP_CLEANSE_PLAINTEXT
+
+By default TLS connections keep a copy of received plaintext data in a
+static buffer until it is overwritten by followup data. When enabling
```
beldmit, line 271: followup => next portion of?

melshuber:
> Please can you submit a CLA:
> 
> https://www.openssl.org/policies/cla.html

Sure the CCLA is already in my mailbox, the ICLA I am singing tomorrow (my printer is dead) and planning to submit them together

melshuber:
are you preferring to amend the commit or a followup and squash later?

melshuber:
I am wondering, if we should add a similar behaviour in in dtls1_read_bytes() as well. But since I am not sure about the peek/read behaviour for datagrams I am a little bit hesitant.

mattcaswell:
> are you preferring to amend the commit or a followup and squash later?

Preferred is to add additional commits fixing up previous ones using `git commit --fixup`. That way whoever merges can just use `git rebase -i --autosquash`

mattcaswell:
> I am wondering, if we should add a similar behaviour in in dtls1_read_bytes() as well. But since I am not sure about the peek/read behaviour for datagrams I am a little bit hesitant.

Oh, yes. Good point - we should do that. It should be virtually the same.

mattcaswell:
Hmm. The travis failures look odd - but I don't think they are related to this change.

melshuber:
> Hmm. The travis failures look odd - but I don't think they are related to this change.

Well that concerns me as well. I run the testsuite on my local machine and there all looks good.
However since travis checks more this is just an indication.

The odd thing is that we see red Xes all over the place even on merged commits. I did not check in detail but the logs look similar.

However since it least the 'if' statements are executed in many configurations, I am not sure what to do about this.

melshuber:
> > Please can you submit a CLA:
> > https://www.openssl.org/policies/cla.html
> 
> Sure the CCLA is already in my mailbox, the ICLA I am singing tomorrow (my printer is dead) and planning to submit them together

I sent the ICLA and CCLA a couple of ours ago, please tell if something is missing

melshuber:
To summarize the delta:

* Addressed your comments
* Added zeroization for DTLS (and test)
* Test: Added checks to explicitly check internal buffer contents after SSL_peek, and after SSL_read
* Test: Moved test functions down, it was in between the large_message tests

melshuber:
> > are you preferring to amend the commit or a followup and squash later?
> 
> Preferred is to add additional commits fixing up previous ones using `git commit --fixup`. That way whoever merges can just use `git rebase -i --autosquash`

So hold back with rebasing myself for now

mattcaswell:
Close/reopen to kick CLA bot

mattcaswell:
Travis is now passing :-)

ssl/record/rec_layer_d1.c
```diff
@@ -120,14 +120,18 @@ void DTLS_RECORD_LAYER_set_write_sequence(RECORD_LAYER *rl, unsigned char *seq)
...
 
-    SSL3_BUFFER_release(&s->rlayer.rbuf);
+    if (s->options & SSL_OP_CLEANSE_PLAINTEXT)
+        OPENSSL_cleanse(rbuf->buf, rbuf->len);
+    SSL3_BUFFER_release(rbuf);
 
     s->rlayer.packet = rdata->packet;
     s->rlayer.packet_length = rdata->packet_length;
-    memcpy(&s->rlayer.rbuf, &(rdata->rbuf), sizeof(SSL3_BUFFER));
+    memcpy(rbuf, &(rdata->rbuf), sizeof(SSL3_BUFFER));
```
mattcaswell, line 134: Here we're copying plaintext out of `rdata`, which comes from `item->data`. There are two callers of `dtls1_copy_record`, and in both cases `item->data` is immediately freed, so we probably also need to cleanse that. This function could be called for any type of record - not just application data. Which makes me wonder whether this option is supposed to cleanse *all* plaintext (e.g. including handshake/alert plaintext), or just application data plaintext. If the former then I think there are some spots that might have been missed  (see further comments below).

melshuber:
> 
> 
> Travis is now passing :-)

Feels better this way :)

melshuber:
Since there are still changes requested. Is something missing from my side?

ssl/record/rec_layer_s3.c
```diff
@@ -1483,6 +1483,8 @@ int ssl3_read_bytes(SSL *s, int type, int *recvd_type, unsigned char *buf,
                 if (SSL3_RECORD_get_length(rr) == 0)
                     SSL3_RECORD_set_read(rr);
             } else {
+                if (s->options & SSL_OP_CLEANSE_PLAINTEXT)
+                    OPENSSL_cleanse(&(rr->data[rr->off]), n);
```
mattcaswell, line 1487: This cleanses record data that we are expecting to receive. Sometimes we received records of other types, e.g. alerts, or new handshake records (see the sections later in this function for alert handling, and handshake fragment/unexpected handshake messages). We're not cleansing the plaintext for those types of records. Although this in itself begs another question - how far do you go? Information that we learn from plaintext handshake records is stored in the SSL object all over the place. Should we cleanse all of those too (if so then the scope suddenly became a lot more complicated)? Alternatively we clarify that the option is only about application data plaintext and restrict ourselves to worrying about that only. (All of this comment applies to the same section of the DTLS code)

test/sslapitest.c
```diff
@@ -1595,6 +1595,119 @@ static int test_large_message_dtls(void)
...
+        cbuf[i] = i & 0xff;
+    }
+
+    if (!TEST_int_eq(SSL_write(clientssl, cbuf, sizeof(cbuf)), sizeof(cbuf)))
+        goto end;
+
+    if (!TEST_int_eq(SSL_peek(serverssl, &sbuf, sizeof(sbuf)), sizeof(sbuf)))
+        goto end;
+
+    if (memcmp(cbuf, sbuf, sizeof(cbuf)))
```
mattcaswell, line 1640: Use `TEST_mem_eq` instead of `memcmp`. You get better diagnostic info if the test ever fails.

test/sslapitest.c
```diff
@@ -1595,6 +1595,119 @@ static int test_large_message_dtls(void)
...
+    rr = serverssl->rlayer.rrec;
+    zbuf = &rr->data[rr->off];
+    if (!TEST_int_eq(rr->length, sizeof(cbuf)))
+        goto end;
+
+    /*
+     * After SSL_peek() the plaintext must still be stored in the
+     * record.
+     */
+    if (memcmp(cbuf, zbuf, sizeof(cbuf)))
```
mattcaswell, line 1657: `TEST_mem_eq`...and so on below

ssl/record/rec_layer_d1.c
melshuber: will recheck and update

test/sslapitest.c
melshuber: will do

test/sslapitest.c
melshuber: will do

ssl/record/rec_layer_s3.c
melshuber: I think clearly we should stop at application data. The whole point is to extend residual information protection to application data, which is often a requirement when using secure channels. 

Applications transport all kind of items within application data of the secure channel (e.g.: Passwords, PSKs, PINs, ...). This feature aims to support the kind of requirements stating that resuduals thereof must be removed as well. These requirements are not that uncommon.

Further, the current implementation only clenses the application record payload, but not the header.

Regarding handshake data: I think in TLS1.2 most of it is sent plaintext anyways. TLS1.3 is different, but to my understanding this is done mainly to protect the from monitoring metainformation (e.g.: certificates used).

Regarding handshake data within the secure channel (e.g.: KEY-UPDATE): prove me wrong, but i do not think that sensitive information is contained in those (other than metainformation).

We could add an "if" to exclude records other than application data, but I do not think that clansing a little bit too much notably more hurts perfomance when comparing it to application data cleansing.

So regarding: 
(1) "if so then the scope suddenly became a lot more complicated" and 
(2) "Alternatively we clarify that the option is only about application data plaintext and restrict ourselves to worrying about that only."

ad (1) totally correct -> I will be a never ending story.
ad (2) I think this is the way to got: I will clarify this in the manpage.

ssl/record/rec_layer_d1.c
melshuber: I am trying to get around where DTLS maintains copies of records.
 Sof far i found three queues
* _unprocessed_rcds.q_: 
  I think this contains ciphertext records (no cleansing required)
* _processed_rcds.q_: I think this contains plaintext records, which can be either app data or handshake. 
  Cleansing is required only when we free it and there is still data left (_DTLS_RECORD_LAYER_clear_)
* _buffered_app_data.q_: Bufferd but not yet deliverd plaintext application data records. Must be cleansed in (DTLS_RECORD_LAYER_clear, in dtls1_read_bytes)

Is it correct that each of these three queues contain DTLS1_RECORD_DATA items where each of those convey it own 16K buffer for a single record?

To my understanding dtls1_copy_record takes an item from one of those queues and replaces s->rlayer.rbuf, with the queued item. The old record is freed (and cleansed) in this process?

Note however, as i understand it _dtls1_copy_record_, does not copy the record it self, in only copies metadata (pointers and stuff).

Keeping this in mind I think we can revert the cleanse modification in _dtls1_copy_record_, since it **only cleanses application data after it is delivered (and already cleansed) to the application**, but we need to add cleansing of leftofer items in _buffered_app_data_ and _processed_rcds_ _DTLS_RECORD_LAYER_clear_.

I am really not totaly sure about my analysis, please let me know if it is correct. I already have prepared a fixup containing the following changes:

* Add cleansing in _DTLS_RECORD_LAYER_clear_
* Revert cleansing in _dtls1_copy_record_

ssl/record/rec_layer_d1.c
```diff
@@ -514,6 +518,8 @@ int dtls1_read_bytes(SSL *s, int type, int *recvd_type, unsigned char *buf,
             if (SSL3_RECORD_get_length(rr) == 0)
                 SSL3_RECORD_set_read(rr);
         } else {
+            if (s->options & SSL_OP_CLEANSE_PLAINTEXT)
+                OPENSSL_cleanse(&(SSL3_RECORD_get_data(rr)[SSL3_RECORD_get_off(rr)]), n);
```
melshuber, line 522: Depending on the read sematics for datagram connections i think the _n_ should be replaced by _SSL3_RECORD_get_length(rr)_

Since when for datagram sockets when reading *up to _n_* bytes from a datagram with a *length less than _n_**. The remaining bytes are dropped and never delivered to the application.

melshuber: Sorry _n_ need to be replaced by:
_len - SSL3_RECORD_get_off(rr)_

However reading the code twice i think OpenSSL does not drop the data, so _n_ seems fine anyways

ssl/record/rec_layer_d1.c
mattcaswell: > Is it correct that each of these three queues contain DTLS1_RECORD_DATA items where each of those convey it own 16K buffer for a single record?

Yes

mattcaswell: > The old record is freed (and cleansed) in this process?

It is not cleansed. However, I don't think it needs to be. If the old record contains plaintext application data then we'll only copy in a new record if we've already read that data - and if we've read it then (as a result of this PR) we will have already cleansed that application data.

> I am really not totaly sure about my analysis, please let me know if it is correct. I already have prepared a fixup containing the following changes:

I agree with your analysis

CHANGES.md
```diff
@@ -1057,6 +1057,14 @@ OpenSSL 3.0
 
    *Boris Pismenny*
 
+ * The SSL option SSL_OP_CLEANSE_PLAINTEXT is introduced. If that
+   option is set, openssl cleanses (zeroize) plaintext bytes from
+   internal buffers after delivering them to the application. Note,
+   the application is still responsible to cleanse other copies (e.g.:
```
mattcaswell, line 1063: A better phrasing might "...is still responsible for cleansing other copies"

melshuber:
Fixed proposal in CHANGES.md

doc/man3/SSL_CTX_set_options.pod
```diff
@@ -265,6 +265,20 @@ functionality is not required. Those applications can turn this feature off by
...
+
+By default TLS connections keep a copy of received plaintext
+application data in a static buffer until it is overwritten by the
+next portion of data. When enabling SSL_OP_CLEANSE_PLAINTEXT
+deciphered application data is cleansed by calling OPENSSL_cleanse(3)
+after passing data to the application. Data is also cleansed when
+releasing the connection (eg. L<SSL_free(3)>).
+
+Since OpenSSL only cleanses internal buffers, the application is
+still responsible for cleansing all other buffers. Most notable this
```
slontis, line 278: NIT: notably (not sure if it should have a , or not)

doc/man3/SSL_CTX_set_options.pod
```diff
@@ -265,6 +265,20 @@ functionality is not required. Those applications can turn this feature off by
...
+applies to buffers passed to functions like L<SSL_read(3)>,
+SSL_peek(3) but also like L<SSL_write(3)>.
```
slontis, line 280: L<SSL_peek(3)>

test/sslapitest.c
```diff
@@ -1595,6 +1595,119 @@ static int test_large_message_dtls(void)
...
+
+    if (!TEST_int_eq(SSL_peek(serverssl, &sbuf, sizeof(sbuf)), sizeof(sbuf)))
+        goto end;
+
+    if (!TEST_mem_eq(cbuf, sizeof(cbuf), sbuf, sizeof(sbuf)))
+        goto end;
+
+    /*
+     * Since we called SSL_peek(), we know the data is stored as
+     * plaintext record. We can gather the pointer to check for
```
slontis, line 1645: a

test/sslapitest.c
```diff
@@ -1595,6 +1595,119 @@ static int test_large_message_dtls(void)
...
+     * zerozation after SSL_read().
```
slontis, line 1646: zeroization

doc/man3/SSL_CTX_set_options.pod
melshuber: Not sure either, but

https://thecriticalreader.com/because-since-and-commas/ and
https://ell.stackexchange.com/questions/38051/should-i-add-a-comma-before-since#:~:text=When%20since%20is%20used%20as,starts%20after%20the%20main%20clause.

state the ',' should stay

melshuber:
@slontis: thx for the review,

I added update according to to your proposial (except the one i commeneted on).

doc/man3/SSL_CTX_set_options.pod
```diff
@@ -265,6 +265,20 @@ functionality is not required. Those applications can turn this feature off by
...
+
+By default TLS connections keep a copy of received plaintext
+application data in a static buffer until it is overwritten by the
+next portion of data. When enabling SSL_OP_CLEANSE_PLAINTEXT
+deciphered application data is cleansed by calling OPENSSL_cleanse(3)
+after passing data to the application. Data is also cleansed when
+releasing the connection (eg. L<SSL_free(3)>).
+
+Since OpenSSL only cleanses internal buffers, the application is still
+responsible for cleansing all other buffers. Most notable this applies
```
mattcaswell, line 278: @slontis made two comments about this. I think you missed the one "notable" -> "notably"

doc/man3/SSL_CTX_set_options.pod
```diff
@@ -265,6 +265,20 @@ functionality is not required. Those applications can turn this feature off by
...
+to buffers passed to functions like L<SSL_read(3)>,
+L<SSL_peek(3)|SSL_read(3)> but also like L<SSL_write(3)>.
```
mattcaswell, line 280: Just `L<SSL_peek(3)>` should be sufficient here (and is consistent with what we normally do). There are symlinks in place to resolve where multiple functions are defined in a single file.

melshuber: I tried with L<SSL_peek(3)> and this leads to a dead link in the html man page.
therefor i did not add it in the beginning

doc/man3/SSL_CTX_set_options.pod
melshuber: ups sorry

doc/man3/SSL_CTX_set_options.pod
melshuber: at least when using links2(1)

doc/man3/SSL_CTX_set_options.pod
melshuber: So @slontis also meant the comma after the notably.

I am not sure about this either.
What do you think?

doc/man3/SSL_CTX_set_options.pod
mattcaswell: Hmm. Strange. But this *should* work, so if it isn't then we've got that problem throughout the man pages. Not an issue for this PR to resolve - so I think we should just have it as`L<SSL_peek(3)>` for consistency.

melshuber: Ok,

JFI: Here is what i tried:

```
./Configure --prefix=$(pwd)/tmp linux-x86_64
make install
links2 tmp/share/doc/openssl/html/man3/SSL_CTX_set_options.html
```
there are no symlinks in `tmp/share/doc/openssl/html/man3/`

mattcaswell: @levitte...any thoughts on that?

doc/man3/SSL_CTX_set_options.pod
melshuber: @slontis; sorry i misread your comment

fixed it as well

mattcaswell: I'm not sure about the comma - so whatever.

beldmit:
Squashed and merged. Thanks!

melshuber:
Thanks all

PR #12252, opened 2020-06-23
TEST: Provide a default openssl.cnf for tests

sebastianas:
In Debian the apps/openssl.cnf gets patched to provide a system wide
policy for TLSv1.2+ and SECLEVEL=2.
Somewhere between alpha1 and alpha3 the testsuite started failing
because the tests requiring < TLSv1.2 failed due to the restriction in
the modified .cnf file.
As far as I can tell, there is no need use the whole .cnf file from
apps which contains mainly certificate related options. Also the
testsuite should not depend on files outside of the testsuite.

Provide an empty .cnf file for the testsuite if none is specified. This
is enough to pass the testsuite.

### DIFF
--- a/test/run_tests.pl
+++ b/test/run_tests.pl
@@ -31,7 +31,7 @@ my $bldtop = $ENV{BLDTOP} || $ENV{TOP};
 my $recipesdir = catdir($srctop, "test", "recipes");
 my $libdir = rel2abs(catdir($srctop, "util", "perl"));
 
-$ENV{OPENSSL_CONF} = rel2abs(catdir($srctop, "apps", "openssl.cnf"));
+$ENV{OPENSSL_CONF} = rel2abs(catdir($srctop, "test", "default.cnf"));
 $ENV{OPENSSL_CONF_INCLUDE} = rel2abs(catdir($bldtop, "providers"));
 $ENV{OPENSSL_MODULES} = rel2abs(catdir($bldtop, "providers"));
 $ENV{OPENSSL_ENGINES} = rel2abs(catdir($bldtop, "engines"));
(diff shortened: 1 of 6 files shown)


mattcaswell:
Looks reasonable to me. Needs second review.

The travis/appveyor failures seem to be unrelated and coming from master.

test/run_tests.pl
```diff
@@ -31,7 +31,7 @@ BEGIN
 my $recipesdir = catdir($srctop, "test", "recipes");
 my $libdir = rel2abs(catdir($srctop, "util", "perl"));
 
-$ENV{OPENSSL_CONF} = rel2abs(catdir($srctop, "apps", "openssl.cnf"));
+$ENV{OPENSSL_CONF} = rel2abs(catdir($srctop, "test", "openssl.cnf"));
```
richsalz, line 34: should it be "default.cnf" ?
There should be a comment in openssl.cnf explaining that it is empty on purpose.

sebastianas: The empty file has a comment now.
There was already a default.cnf so I renamed it to default-provider.cnf (because it appears to be provider related) so I can have a default.cnf which is empty (minus the comment).

slontis:
It looks like travis did not like these changes..

levitte:
> Also the testsuite should not depend on files outside of the testsuite.

Agreed...  however, the "standard" config file *should* be tested...  I believe that we have a few tests that intentionally does that, and if not, we should consider adding them.

h-vetinari:
This was (seemingly) auto-closed - since it belonged to the 3.0 project, is this still relevant?

PR #9444, opened 2019-07-23
Add Argon2 KDF support.

ckalina:
##### Checklist

- [x] tests are added or updated

Hello,

this is an initial RFC pull request aiming at getting Argon2 support in OpenSSL. Currently only single threaded Argon2 mode is enforced. I tried to keep as close to [reference implementation](https://github.com/P-H-C/phc-winner-argon2) (which is dual licensed under the [CC0 License](https://creativecommons.org/about/cc0) and the [Apache 2.0 License](http://www.apache.org/licenses/LICENSE-2.0) and as such should be compatible with OpenSSL) as possible to aid in future backports.

I'd appreciate any feedback you could provide (especially regarding threading).

Related: https://github.com/openssl/openssl/issues/4091#

Thank you!

Best regards,
Cestmir

PR #12256, opened 2020-06-24
KDF: Introduce Argon2i, Argon2d, Argon2id

ckalina:
Rework of https://github.com/openssl/openssl/pull/9444
Threading has since been singled out into a separate MR here: https://github.com/openssl/openssl/pull/12255 and is required for this set to go in.

Currently no RFC is available; IETF draft: https://datatracker.ietf.org/doc/draft-irtf-cfrg-argon2/

##### Checklist


guidovranken:
I tested this again with Cryptofuzz. Verifying against Botan's ARGON2 implementation. So far, output appears to be correct in all cases.

It found a minor flaw:

```
providers/implementations/kdfs/argon2.c:1128:12: runtime error: null pointer passed as argument 1, which is declared to never be null
/usr/include/string.h:43:28: note: nonnull attribute specified here
```

You can add `return 1` to `kdf_argon2_derive` if `outlen` is 0 to circumvent this.

Once this is merged, long term fuzzing at OSS-Fuzz should bring to light any remaining issues.

ckalina:
@guidovranken, nice catch. Fixed (although it should `return 0` as Argon2 requires `outlen >= 4`).

ckalina:
Rebased changes from https://github.com/openssl/openssl/pull/12255 (threading support MR), changed a handful of `crypto_thread*` function calls to conform to changes made therein.

include/openssl/kdf.h
```diff
@@ -79,6 +79,13 @@ void EVP_KDF_names_do_all(const EVP_KDF *kdf,
 # define EVP_KDF_CTRL_SET_UKM                0x16 /* unsigned char *, size_t */
 # define EVP_KDF_CTRL_SET_CEK_ALG            0x17 /* char * */
 # define EVP_KDF_CTRL_SET_SHARED_INFO        EVP_KDF_CTRL_SET_SSKDF_INFO
+# define EVP_KDF_CTRL_SET_ARGON2_LANES       0xf1 /* uint32 */
```
slontis, line 82: Unless there is a good reason, I think these CTRL's should not be needed (The controls were there for the old API's, OSSL_PARAMS should be sufficient for any new KDF's).

include/crypto/evp.h
```diff
@@ -191,6 +191,9 @@ struct evp_kdf_st {
     OSSL_OP_kdf_set_ctx_params_fn *set_ctx_params;
 };
 
+extern const EVP_KDF argon2i_kdf_meth;
```
slontis, line 194: These don't look like they should be here..

doc/man7/EVP_KDF-ARGON2.pod
```diff
@@ -0,0 +1,83 @@
...
+associated data, s is the digest length, t the number of threads.
+
+You may use f to control whether or not B<OPENSSL_cleanse> is called on
+password and/or secret during initial hash phase, by setting applying
+1 or 2 mask, respectively.
+
+t, s, l, m, and f are of type B<uint32_t>.
+
+=back
+
```
slontis, line 64: Can you add sections for 
  NOTES
  EXAMPLES
  CONFORMING TO
  HISTORY

(Similiar to EVP_KDF_SS.pod)

include/openssl/evperr.h
```diff
@@ -152,6 +152,7 @@ int ERR_load_EVP_strings(void);
 #  define EVP_F_S390X_AES_GCM_CTRL                         0
 #  define EVP_F_S390X_AES_GCM_TLS_CIPHER                   0
 #  define EVP_F_SCRYPT_ALG                                 0
+#  define EVP_F_ARGON2_ALG                                 0
```
slontis, line 155: Remove this..
Just use (0, X) instead in the code - these flags are dead now.

doc/man7/EVP_KDF-ARGON2.pod
```diff
@@ -0,0 +1,83 @@
...
+=item "l" (B<OSSL_KDF_PARAM_ARGON2_LANES>) <unsigned integer>
+
+=item "m" (B<OSSL_KDF_PARAM_ARGON2_MEMCOST>) <unsigned integer>
+
+=item "f" (B<OSSL_KDF_PARAM_ARGON2_FLAGS>) <unsigned integer>
+
+These parameters configure the Argon2.
+
+m is the memory size argument, l the number of lanes, a is the optional
+associated data, s is the digest length, t the number of threads.
```
slontis, line 55: maybe move these into the items they are associated with..

doc/man7/EVP_KDF-ARGON2.pod
```diff
@@ -0,0 +1,83 @@
...
+
+You may use f to control whether or not B<OPENSSL_cleanse> is called on
+password and/or secret during initial hash phase, by setting applying
```
slontis, line 58: or applying?

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+#include <internal/worker.h>
+#include <internal/numbers.h>
+#include <prov/implementations.h>
+#include <prov/provider_ctx.h>
+#include <prov/providercommonerr.h>
+#include <prov/blake2.h>
+
+#ifndef OPENSSL_NO_ARGON2
+
+# define ARGON2_MIN_LANES UINT32_C(1)
```
slontis, line 33: Is UINT32_C c99? (That might be a problem).

test/evp_kdf_test.c
```diff
@@ -160,6 +162,180 @@ static int test_kdf_pbkdf2(void)
     return ret;
 }
 
+#ifndef OPENSSL_NO_ARGON2
+static int test_kdf_argon2(void)
+{
+    int ret;
+    EVP_KDF_CTX *kctx;
+    OSSL_PARAM params1[3], params2[5], params3[6], params4[5], *p;
```
slontis, line 170: why use seperate arrays?

test/evp_kdf_test.c
```diff
@@ -160,6 +162,180 @@ static int test_kdf_pbkdf2(void)
...
+    *p = OSSL_PARAM_construct_end();
+
+    ret =
+        TEST_ptr(kctx = get_kdfbyname(SN_argon2d))
+        && TEST_true(EVP_KDF_set_ctx_params(kctx, params1))
+        && TEST_int_eq(EVP_KDF_derive(kctx, out, sizeof(out)), 1)
+        && TEST_mem_eq(out, sizeof(out), expected_res1, sizeof(expected_res1));
+
+    if (ret == 0)
+        return 0;
```
slontis, line 200: these returns throughout are all going to leak.

test/evp_kdf_test.c
```diff
@@ -160,6 +162,180 @@ static int test_kdf_pbkdf2(void)
...
+    threads = 2;
+    p = params4;
+    *p++ = OSSL_PARAM_construct_octet_string(OSSL_KDF_PARAM_PASSWORD,
+                                             (char *)"1234567890", 10);
+    *p++ = OSSL_PARAM_construct_octet_string(OSSL_KDF_PARAM_SALT,
+                                             (char *)"saltsalt", 8);
+    *p++ = OSSL_PARAM_construct_int(OSSL_KDF_PARAM_THREADS, &threads);
+    *p++ = OSSL_PARAM_construct_int(OSSL_KDF_PARAM_ARGON2_LANES, &threads);
+    *p++ = OSSL_PARAM_construct_int(OSSL_KDF_PARAM_ARGON2_MEMCOST, &m_cost);
+    *p = OSSL_PARAM_construct_end();
```
slontis, line 302: Its a bit messy lumping all 3 tests in here..
It may be cleaner to do them as seperate tests.

You can also use ADD_ALL_TESTS and data drive it..

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+    if (outlen > UINT32_MAX || out == NULL || outlen == 0)
+        goto fail;
+
+    if (outlen > BLAKE2B_OUTBYTES)
+        blake_outlen = BLAKE2B_OUTBYTES;
+
+    /* Ensure little-endian byte order */
+    store32(outlen_bytes, (uint32_t)outlen);
+
+    if ((mac = EVP_MAC_fetch(NULL, "blake2bmac", NULL)) == NULL)
```
slontis, line 924: You might be better off doing this fetch once and then just passing the mac pointer about..

Either way the EVP_MAC_fetch() needs to be passed the libctx anywhere where it is used.

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+    ctx->out = NULL;
+    return 0;
+}
+
+static void kdf_argon2_reset(void *vctx)
+{
+    KDF_ARGON2 *ctx_wrap;
+    ARGON2_CTX *ctx;
+    ARGON2_TYPE type;
+
```
slontis, line 1149: Like the other KDF reset functions there is a bug where the 
ctx->provctx is being NULLED here..
It needs to be saved at the start and restored between the memset and init()

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+        goto fail;
+
+    if (EVP_MAC_init(ctx) != 1)
+        goto fail;
+
+    if (EVP_MAC_update(ctx, in, inlen) != 1)
+        goto fail;
+
+    if (EVP_MAC_final(ctx, out, (size_t *) &out_written, outlen) != 1)
+        goto fail;
```
slontis, line 886: Collapse these into either
```
if (X
        || Y
        || Z)
    goto fail;
```

OR

ret = A
        && B
        && C
        && D;
```

File: providers/implementations/kdfs/argon2.c [hunk 14]
```diff
@@ -0,0 +1,1488 @@
...
+    par_n = 0;
+
+    if (out == NULL || outlen == 0)
+        goto fail;
+
+    if ((mac = EVP_MAC_fetch(NULL, "blake2bmac", NULL)) == NULL)
+        goto fail;
+
+    if ((ctx = EVP_MAC_new_ctx(mac)) == NULL)
+        goto fail;
```
slontis, line 868: collapse as described below

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+# define ARGON2_MIN_PWD_LENGTH UINT32_C(0)
+# define ARGON2_MAX_PWD_LENGTH UINT32_C(0xFFFFFFFF)
+# define ARGON2_MIN_AD_LENGTH UINT32_C(0)
+# define ARGON2_MAX_AD_LENGTH UINT32_C(0xFFFFFFFF)
+# define ARGON2_MIN_SALT_LENGTH UINT32_C(8)
+# define ARGON2_MAX_SALT_LENGTH UINT32_C(0xFFFFFFFF)
+# define ARGON2_MIN_SECRET UINT32_C(0)
+# define ARGON2_MAX_SECRET UINT32_C(0xFFFFFFFF)
+# define ARGON2_DEFAULT_FLAGS UINT32_C(0)
+# define ARGON2_FLAG_CLEAR_PASSWORD (UINT32_C(1) << 0)
```
slontis, line 57: These should probably be public..

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+    return ret;
+}
+
+static void kdf_argon2_init(KDF_ARGON2 *ctx, ARGON2_TYPE type)
+{
+    ARGON2_CTX *c;
+
+    c = &ctx->ctx;
+    memset(c, 0, sizeof(*c));
+    c->outlen = 64;
```
slontis, line 986: These magic numbers should be #defined (with the tag DEFAULT in the name)

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+            x[2 * i + 96], x[2 * i + 97], x[2 * i + 112], x[2 * i + 113]       \
+        );                                                                     \
+    } while ((void)0, 0)
+
+typedef struct {
+    uint64_t v[ARGON2_QWORDS_IN_BLOCK];
+} block;
+
+typedef enum {
+    ARGON2_VERSION_10 = 0x10,
```
slontis, line 122: Do you see any need for the version 10 variant?
If so add a setter - otherwise maybe just remove it..

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+        OPENSSL_clear_free(ctx->ad, ctx->adlen);
+        ctx->ad = NULL;
+        ctx->adlen = 0U;
+    }
+
+    if (!OSSL_PARAM_get_octet_string(p, (void **)&ctx->ad, 0, &buflen))
+        return 0;
+
+    if (buflen+1 < ARGON2_MIN_AD_LENGTH+1 || buflen > ARGON2_MAX_AD_LENGTH
+            || buflen > UINT32_MAX) {
```
slontis, line 1324: Isnt this sufficient?
```
buflen < ARGON2_MIN_AD_LENGTH
   || buflen > ARGON2_MAX_AD_LENGTH
```

slontis: Same question applies to cases above also.

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+static int blake2b(void *out, size_t outlen, const void *in, size_t inlen,
+                   const void *key, size_t keylen)
+{
+    int ret;
+    EVP_MAC *mac;
+    EVP_MAC_CTX *ctx;
+    OSSL_PARAM par[3];
+    size_t par_n, out_written;
+
+    mac = NULL;
```
slontis, line 856: It is arguably more readable to move these up.

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+    }
+
+    /* Hash the result */
+    store_block(blockhash_bytes, &blockhash);
+    blake2b_long(ctx->out, ctx->outlen, blockhash_bytes, ARGON2_BLOCK_SIZE);
+    OPENSSL_cleanse(blockhash.v, ARGON2_BLOCK_SIZE);
+    OPENSSL_cleanse(blockhash_bytes, ARGON2_BLOCK_SIZE);
+
+    if (ctx->type != ARGON2_D)
+        OPENSSL_secure_clear_free(ctx->memory, ctx->memory_blocks*sizeof(block));
```
slontis, line 842: NIT: spacing on the operator ' * '

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+    ARGON2_MIN(UINT32_C(32), (sizeof(void *) * CHAR_BIT - 10 - 1))
+# define ARGON2_MAX_MEMORY                                                      \
+    ARGON2_MIN(UINT32_C(0xFFFFFFFF), UINT64_C(1) << ARGON2_MAX_MEMORY_BITS)
+# define ARGON2_MIN_TIME UINT32_C(1)
+# define ARGON2_MAX_TIME UINT32_C(0xFFFFFFFF)
+# define ARGON2_MIN_PWD_LENGTH UINT32_C(0)
+# define ARGON2_MAX_PWD_LENGTH UINT32_C(0xFFFFFFFF)
+# define ARGON2_MIN_AD_LENGTH UINT32_C(0)
+# define ARGON2_MAX_AD_LENGTH UINT32_C(0xFFFFFFFF)
+# define ARGON2_MIN_SALT_LENGTH UINT32_C(8)
```
slontis, line 52: Maybe this should be 16?

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+        c = c + d + 2 * mul_lower(c, d);                                       \
+        b = rotr64(b ^ c, 24);                                                 \
+        a = a + b + 2 * mul_lower(a, b);                                       \
+        d = rotr64(d ^ a, 16);                                                 \
+        c = c + d + 2 * mul_lower(c, d);                                       \
+        b = rotr64(b ^ c, 63);                                                 \
+    } while ((void)0, 0)
+
+# undef BLAKE2_ROUND
+# define BLAKE2_ROUND(v0, v1, v2, v3, v4, v5, v6, v7, v8, v9, v10, v11,        \
```
slontis, line 82: Not sure about this name.. Its like a blake_round..

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+        G(v1, v6, v11, v12);                                                   \
+        G(v2, v7, v8, v13);                                                    \
+        G(v3, v4, v9, v14);                                                    \
+    } while ((void)0, 0)
+
+# undef BLAKE2_ROUND_COLUMN
+# define BLAKE2_ROUND_COLUMN(x, i)                                             \
+    do {                                                                       \
+        BLAKE2_ROUND(                                                          \
+            x[16 * i],      x[16 * i + 1],  x[16 * i + 2],  x[16 * i + 3],     \
```
slontis, line 99: I assume this gets optimized but you could maybe cache the 16*i bit (or put it in a pointer)

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+            x[2 * i],      x[2 * i + 1],  x[2 * i + 16],  x[2 * i + 17],       \
+            x[2 * i + 32], x[2 * i + 33], x[2 * i + 48],  x[2 * i + 49],       \
+            x[2 * i + 64], x[2 * i + 65], x[2 * i + 80],  x[2 * i + 81],       \
+            x[2 * i + 96], x[2 * i + 97], x[2 * i + 112], x[2 * i + 113]       \
+        );                                                                     \
+    } while ((void)0, 0)
+
+typedef struct {
+    uint64_t v[ARGON2_QWORDS_IN_BLOCK];
+} block;
```
slontis, line 119: Uppercase this name.

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+    uint64_t w = 0;
+    unsigned char i;
+
+    if (is_endian.little) {
+        memcpy(&w, src, sizeof(w));
+        return w;
+    }
+
+    for (i = 0; i < 8; i++)
+        w |= ((uint64_t)src[i] << (8*i));
```
slontis, line 263: spaces around *

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+# define ARGON2_FLAG_CLEAR_PASSWORD (UINT32_C(1) << 0)
+# define ARGON2_FLAG_CLEAR_SECRET (UINT32_C(1) << 1)
+# define ARGON2_BLOCK_SIZE 1024
+# define ARGON2_QWORDS_IN_BLOCK ARGON2_BLOCK_SIZE / 8
+# define ARGON2_OWORDS_IN_BLOCK ARGON2_BLOCK_SIZE / 16
+# define ARGON2_HWORDS_IN_BLOCK ARGON2_BLOCK_SIZE / 32
+# define ARGON2_512BIT_WORDS_IN_BLOCK ARGON2_BLOCK_SIZE / 64
+# define ARGON2_ADDRESSES_IN_BLOCK 128
+# define ARGON2_PREHASH_DIGEST_LENGTH 64
+# define ARGON2_PREHASH_SEED_LENGTH 72
```
slontis, line 66: # define ARGON2_PREHASH_SEED_LENGTH
(ARGON2_PREHASH_DIGEST_LENGTH + (2 * sizeof()))

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+            ref_area_sz = ctx->lane_length - ctx->segment_length +
+                ((index == 0) ? (-1) : 0);
+        if (slice != ARGON2_SYNC_POINTS - 1)
+            start_pos = (slice + 1) * ctx->segment_length;
+        break;
+    }
+
+    rel_pos = pseudo_rand;
+    rel_pos = rel_pos * rel_pos >> 32;
+    rel_pos = ref_area_sz - 1 - (ref_area_sz * rel_pos>> 32);
```
slontis, line 455: NIT spacing on operator >>

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+    start_idx = 0;
+
+    /* We've generated the first two blocks. Generate the 1st block of addrs. */
+    if ((pass == 0) && (slice == 0)) {
+        start_idx = 2;
+        if (data_indep_addressing(ctx, pass, slice))
+            next_addresses(&address_block, &input_block, &zero_block);
+    }
+
+    curr_offset = lane * ctx->lane_length+slice * ctx->segment_length+start_idx;
```
slontis, line 496: more operator spacing NITS :)

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+        ref_block = ctx->memory + ctx->lane_length * ref_lane + ref_index;
+        curr_block = ctx->memory + curr_offset;
+        if (ARGON2_VERSION_10 == ctx->version) {
+            /* Version 1.2.1 and earlier: overwrite, not XOR */
+            fill_block(ctx->memory + prev_offset, ref_block, curr_block, 0);
+            continue;
+        }
+
+        if(pass == 0)
+            fill_block(ctx->memory + prev_offset, ref_block, curr_block, 0);
```
slontis, line 537: maybe just use this instead of the if
```
, pass == 0 ? 0 : 1
```

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+            }
+            for (l = ctx->lanes - ctx->threads; l < ctx->lanes; ++l) {
+                if (crypto_thread_join(ctx->openssl_ctx, t[l], NULL) == 0)
+                    goto fail;
+                t[l] = NULL;
+            }
+        }
+    }
+
+    crypto_thread_clean(ctx->openssl_ctx, NULL);
```
slontis, line 600: What would happen here if something else called crypto_thread_clean(ctx->openssl_ctx, NULL) during this function??

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+    if (ctx->m_cost < 8 * ctx->lanes)
+        return 0;
+    if (ctx->t_cost < ARGON2_MIN_TIME)
+        return 0;
+    if (ctx->t_cost > ARGON2_MAX_TIME)
+        return 0;
+    if (ctx->lanes < ARGON2_MIN_LANES)
+        return 0;
+    if (ctx->lanes > ARGON2_MAX_LANES)
+        return 0;
```
slontis, line 697: collapsing this to just do a single return 0 would be cleaner.

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+    if (ctx->salt == NULL && ctx->saltlen != 0)
+        return 0;
+    if (ctx->saltlen < ARGON2_MIN_SALT_LENGTH)
+        return 0;
+    if (ctx->saltlen > ARGON2_MAX_SALT_LENGTH)
+        return 0;
+    if (ctx->secret == NULL && ctx->secretlen != 0)
+        return 0;
+    if (ctx->secret != NULL && ARGON2_MIN_SECRET+1 > ctx->secretlen+1)
```
slontis, line 673: Not sure I have seen this +1 stuff anywhere else in the code.

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+    };
+
+    if (ctx == NULL || blockhash == NULL)
+        return;
+
+    mdctx = EVP_MD_CTX_create();
+    if (mdctx == NULL)
+        goto fail;
+
+    if (EVP_DigestInit_ex(mdctx, EVP_blake2b512(), NULL) != 1)
```
slontis, line 734: This is not library context friendly.
blake2b512 should be fetched using the ctx->openssl_ctx

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+
+    if (outlen > UINT32_MAX)
+        return 0;
+
+    if (outlen == 0)
+        return 0;
+
+    if (outlen != ctx->outlen)
+        kdf_argon2_ctx_set_digest_length(ctx, (uint32_t) outlen);
+
```
slontis, line 1098: maybe use
```
sz = ctx->outlen + 1;
```
Then it can be used in the cleanup also.

providers/implementations/macs/blake2_mac_impl.c
```diff
@@ -78,13 +78,10 @@ static int blake2_mac_init(void *vmacctx)
...
     struct blake2_mac_data_st *macctx = vmacctx;
 
-    /* Check key has been set */
-    if (macctx->params.key_length == 0) {
-        ERR_raise(ERR_LIB_PROV, PROV_R_NO_KEY_SET);
-        return 0;
-    }
-
-    return BLAKE2_INIT_KEY(&macctx->ctx, &macctx->params, macctx->key);
+    if (macctx->params.key_length == 0)
```
slontis, line 81: Is this change relevant to the normal case or just for the argon usage?

slontis:
travis errors are relevant.

providers/implementations/kdfs/argon2.c
richsalz: not thrilled with the null-pointer check.  if you give a size for something, and a NULL pointer, you did something wrong and a segfault seems reasonable and the best way to find and fix the error.

slontis: An assert would be better than segfaulting.

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+    long one;
+    char little;
+} is_endian = { 1 };
+
+typedef struct {
+    void *provctx;
+    ARGON2_CTX ctx;
+} KDF_ARGON2;
+
+static OSSL_OP_kdf_newctx_fn kdf_argon2i_new;
```
slontis, line 181: When you rebase off master these names will have to change to OSSL_FUNC_kdf_xxx_fn

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1488 @@
...
+#include <stdarg.h>
+#include <limits.h>
+#include <string.h>
+#include <crypto/evp.h>
+#include <openssl/evp.h>
+#include <openssl/objects.h>
+#include <openssl/crypto.h>
+#include <openssl/kdf.h>
+#include <openssl/err.h>
+#include <openssl/core_names.h>
```
slontis, line 22: When you rebase off master 'core_names.h' has been removed..

levitte: Uhm, no it hasn't.  *`core_numbers.h`* was renamed to `core_dispatch.h`

levitte:
@ckalina, are you around?

richsalz:
Is this the same algorithm in https://datatracker.ietf.org/doc/draft-irtf-cfrg-argon2/ ?  Looks like it needs some edits, and it will be some months before published as an RFC.

richsalz:
Quoting the IRTF (research) chair:
```
There are some very minor comments to address, then this should be ready for IESG conflict review and publication.

Colin
```

mbroz:
@richsalz Yes, the RFC describes Argon2 that is part of this merge request (I linked it somewhere but perhaps it was the previous pull request).
The primary motivation here was that we use it in cryptsetup (embedded Argon2 reference implemenatation from Argon2 authors - also authors of RFC draft). Because our default crypto backend is OpenSSL we would like to swich to natiove OpenSSL implementation.

providers/implementations/kdfs/argon2.c
ckalina: Renamed, thanks!

providers/implementations/kdfs/argon2.c
ckalina: Fixed, thanks!

providers/implementations/kdfs/argon2.c
ckalina: This X+1 > Y+1 was used to appease some overzealous Travis checks that would've triggered a warning otherwise.

providers/implementations/kdfs/argon2.c
ckalina: crypto_thread_clean locks a mutex to avoid issues of that sort.

providers/implementations/kdfs/argon2.c
ckalina: Opted for the latter approach. Thanks!

providers/implementations/kdfs/argon2.c
ckalina: Went with PERMUTATION_P (and PERMUTATION_P_{COLUMN,ROW}) in accordance with the Argon2 paper naming.

providers/implementations/kdfs/argon2.c
ckalina: 16 is recommended, not minimal (see https://github.com/P-H-C/phc-winner-argon2/blob/master/argon2-specs.pdf)

providers/implementations/kdfs/argon2.c
ckalina: You're right, that was an artifact from a pre-providers version of the implementation. Removed, thanks!

providers/implementations/kdfs/argon2.c
ckalina: Setter added. Code-wise, the distinction is trivial enough ... we might as well support it.

providers/implementations/kdfs/argon2.c
ckalina: Defined.

providers/implementations/kdfs/argon2.c
ckalina: The flags field was split into separate flags setters with their own semantics, thus made public.

providers/implementations/kdfs/argon2.c
ckalina: Passing mac around in ctx.

test/evp_kdf_test.c
ckalina: Split it into multiple functions -- test cases.

test/evp_kdf_test.c
ckalina: Not relevant anymore, since the tests were separated into distinct functions.

providers/implementations/kdfs/argon2.c
ckalina: To my knowledge, yes (https://busybox.net/~landley/c99-draft.html#B.17).

doc/man7/EVP_KDF-ARGON2.pod
ckalina: Added separate flags.

ckalina:
@slontis Thank you for the review! Addressed.
@levitte Yes, was mostly busy with https://github.com/openssl/openssl/pull/12255. Rebased.
@richsalz Yes, it is the same algorithm.

mattcaswell:
IMO this should not go into 3.0. Removing that beta 1 milestone.

tarcieri:
If anyone is interested in trying to specify an ASN.1 module which would allow Argon2 to be used with PKCS#5 (since RFC9106 didn't seem to specify one), I opened an issue about that here:

https://github.com/P-H-C/phc-winner-argon2/issues/348

providers/implementations/digests/blake2_prov.c
```diff
@@ -35,7 +35,75 @@ IMPLEMENT_digest_functions(blake2s256, BLAKE2S_CTX,
...
-IMPLEMENT_digest_functions(blake2b512, BLAKE2B_CTX,
-                           BLAKE2B_BLOCKBYTES, BLAKE2B_DIGEST_LENGTH, 0,
-                           ossl_blake2b512_init, ossl_blake2b_update,
-                           ossl_blake2b_final)
+
+static OSSL_FUNC_digest_init_fn blake2b512_internal_init;
+static OSSL_FUNC_digest_newctx_fn blake2b512_newctx;
+static OSSL_FUNC_digest_freectx_fn blake2b512_freectx;
+static OSSL_FUNC_digest_dupctx_fn blake2b512_dupctx;
+static OSSL_FUNC_digest_final_fn blake2b512_internal_final;
```
paulidale, line 43: I would expect to also see forward declarations for
`blake2b512_get_params`, `ossl_blake2b_set_ctx_params`, `ossl_blake2b_settable_ctx_params` as well.

providers/implementations/digests/blake2_prov.c
```diff
@@ -35,7 +35,75 @@ IMPLEMENT_digest_functions(blake2s256, BLAKE2S_CTX,
...
+{
+    return ossl_prov_is_running() && ossl_blake2b_set_ctx_params(ctx, params)
+        && ossl_blake2b512_init(ctx);
+}
+
+static void *blake2b512_newctx(void *prov_ctx)
+{
+    struct blake2b_md_data_st *ctx =
+        ossl_prov_is_running()? OPENSSL_zalloc(sizeof(*ctx)) : NULL;
+    return ctx;
```
paulidale, line 55: We normally put a blank line between declarations and code.

I guess it would be rewritten:
```c
return ossl_prov_is_running() ? OPENSSL_zalloc(sizeof(struct blake2b_md_data_st)) : NULL;
```

But I think it looks better with just a blank line inserted.

providers/implementations/digests/blake2_prov.c
```diff
@@ -35,7 +35,75 @@ IMPLEMENT_digest_functions(blake2s256, BLAKE2S_CTX,
...
+}
+
+static void blake2b512_freectx(void *vctx)
+{
+    struct blake2b_md_data_st *ctx = (struct blake2b_md_data_st *)vctx;
+    OPENSSL_clear_free(ctx, sizeof(*ctx));
```
paulidale, line 61: Blank line again & more below.

doc/man7/EVP_KDF-ARGON2.pod
```diff
@@ -0,0 +1,142 @@
...
+
+EVP_KDF-ARGON2 - The Argon2 EVP KDF implementation
+
+=head1 DESCRIPTION
+
+Support for computing the B<argon2> password-based KDF through the B<EVP_KDF>
+API.
+
+The EVP_KDF-ARGON2 algorithm implements the Argon2 password-based key
+derivation function, as described in IETF Argon2 draft.  It is memory-hard in
```
paulidale, line 13: It's not a draft any longer is it?

doc/man7/EVP_KDF-ARGON2.pod
```diff
@@ -0,0 +1,142 @@
...
+ *p++ = OSSL_PARAM_uint(OSSL_KDF_PARAM_THREADS, &threads);
+ *p++ = OSSL_PARAM_uint32(OSSL_KDF_PARAM_ARGON2_LANES, &threads);
+ *p++ = OSSL_PARAM_uint32(OSSL_KDF_PARAM_ARGON2_MEMCOST, &m_cost);
+ *p = OSSL_PARAM_construct_end();
+
+ if (EVP_KDF_CTX_set_params(kctx, p) <= 0) {
+     error("EVP_KDF_CTX_set_params");
+ }
+
+ if (EVP_KDF_derive(kctx, out, sizeof(out)) <= 0) {
```
paulidale, line 105: EVP_KDF_derive gained a params argument.  These five lines can be reduced to:
```c
if (EVP_KDF_derive(kctx, out, sizeof(out), params) >= 0) {
```
You lose the separate error reporting.

doc/man7/EVP_KDF-ARGON2.pod
```diff
@@ -0,0 +1,142 @@
...
+ *p++ = OSSL_PARAM_octet_string(OSSL_KDF_PARAM_PASSWORD,
+                                          (char *)"1234567890", 10);
+ *p++ = OSSL_PARAM_octet_string(OSSL_KDF_PARAM_SALT,
+                                          (char *)"saltsalt", 8);
+ *p++ = OSSL_PARAM_uint(OSSL_KDF_PARAM_THREADS, &threads);
+ *p++ = OSSL_PARAM_uint32(OSSL_KDF_PARAM_ARGON2_LANES, &threads);
+ *p++ = OSSL_PARAM_uint32(OSSL_KDF_PARAM_ARGON2_MEMCOST, &m_cost);
+ *p = OSSL_PARAM_construct_end();
+
+ if (EVP_KDF_CTX_set_params(kctx, p) <= 0) {
```
paulidale, line 101: The `p` here should be `params` or it does nothing.  But see below.

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1551 @@
...
+    ARGON2_POS pos;
+    ARGON2_CTX *ctx;
+} ARGON2_THREAD_DATA;
+
+union endianity_test {
+    long one;
+    char little;
+};
+
+static union endianity_test is_endian = { 1 };
```
paulidale, line 192: OpenSSL has an endian test, but to use them I think.

See include/internal/endian.h
`DECLARE_IS_ENDIAN;` declares it
`IS_LITTLE_ENDIAN` and `IS_BIG_ENDIAN` test it.

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1551 @@
...
+} KDF_ARGON2;
+
+static OSSL_FUNC_kdf_newctx_fn kdf_argon2i_new;
+static OSSL_FUNC_kdf_newctx_fn kdf_argon2d_new;
+static OSSL_FUNC_kdf_newctx_fn kdf_argon2id_new;
+static OSSL_FUNC_kdf_freectx_fn kdf_argon2_free;
+static OSSL_FUNC_kdf_reset_fn kdf_argon2_reset;
+static OSSL_FUNC_kdf_derive_fn kdf_argon2_derive;
+static OSSL_FUNC_kdf_settable_ctx_params_fn kdf_argon2_settable_ctx_params;
+static OSSL_FUNC_kdf_set_ctx_params_fn kdf_argon2_set_ctx_params;
```
paulidale, line 206: Looks like forward declarations for: `kdf_argon2_get_ctx_params` and `kdf_argon2_gettable_ctx_params` are missing./

paulidale (commented):
I expected to see some unit test cases for evp_test for this.

providers/implementations/digests/blake2_prov.c
ckalina: `blake2b512_get_params` was present elsewhere in the file, moved up

`ossl_blake2b_set_ctx_params` and `ossl_blake2b_settable_ctx_params` are declared in `providers/implementations/include/prov/blake2.h` that gets included. Would you prefer to include a forward declaration in the file anyway?

doc/man7/EVP_KDF-ARGON2.pod
ckalina: Fixed, thanks!

providers/implementations/kdfs/argon2.c
ckalina: They should be present a couple of lines below, will move them up.

ckalina:
Since GitHub won't let me comment on an "outdated" comment, replying here:

Tests migrated from evp_kdf_test to evp_test. It required a small change to evp_test to accept the number of threads in the test spec 
```
Threads = N
```
and calls to enable/disable threading.

mattcaswell:
> Is there anything blocking this PR now when threads are in?

It would be great to get this rebased so we can continue the review work. I'm not aware of any other blockers.

mbroz:
Could we move this forward, please? What is blocking this PR now when threads are in?

doc/man7/EVP_KDF-ARGON2.pod
```diff
@@ -0,0 +1,150 @@
...
+L<EVP_KDF(3)>,
+L<EVP_KDF_CTX_new(3)>,
+L<EVP_KDF_CTX_free(3)>,
+L<EVP_KDF_CTX_set_params(3)>,
+L<EVP_KDF_derive(3)>,
+L<EVP_KDF(3)/PARAMETERS>
+
+=head1 HISTORY
+
+This functionality was added to OpenSSL 3.0.
```
romen, line 139: This should be amended to say 3.2 probably.

test/evp_test.c
```diff
@@ -2762,6 +2763,55 @@ static int kdf_test_ctrl(EVP_TEST *t, EVP_KDF_CTX *kctx,
...
+        t->skip = 1;
+        goto end;
+    }
+
+    if (strcmp(name, "memcost") == 0
+        && OSSL_PARAM_locate_const(defs, name) == NULL) {
+        TEST_info("skipping, setting 'memcost' is unsupported");
+        t->skip = 1;
+        goto end;
+    }
```
romen, line 2792: The memcost check is duplicated

providers/implementations/kdfs/argon2.c
romen: @ckalina @slontis is this comment still relevant?

romen (changes requested):
Thanks @ckalina for your efforts. 
This PR seems mostly fine to me.

I left a few minor comments.

Also I think this PR should include a CHANGES entry to document the addition of new supported KFDs.

Configure
```diff
@@ -668,6 +669,9 @@ my @disable_cascades = (
     "threads"           => [ "thread-pool" ],
     "thread-pool"       => [ "default-thread-pool" ],
 
+    "blake2"            => [ "argon2" ],
+    sub { !$disabled{"fips"} } => [ "argon2" ],
```
t8m, line 673: This does not look right. Enabling fips provider should not disable argon2.

crypto/objects/objects.txt
```diff
@@ -1689,6 +1689,11 @@ secg-scheme 14 3 : dhSinglePass-cofactorDH-sha512kdf-scheme
 1 3 6 1 4 1 311 60 2 1 2	: jurisdictionST	: jurisdictionStateOrProvinceName
 1 3 6 1 4 1 311 60 2 1 3	: jurisdictionC		: jurisdictionCountryName
 
+# RFC 9106 Argon2 (see https://www.rfc-editor.org/rfc/rfc9106.txt)
+                            : ARGON2I  : argon2i
+                            : ARGON2D  : argon2d
+                            : ARGON2ID : argon2id
+
```
t8m, line 1696: If I understand it right the Argon2 NIDs aren't used. Could you please remove them? As there do not seem to be any official Argon2 object identifiers, there is no point in adding it to the object database.

doc/man7/EVP_KDF-ARGON2.pod
```diff
@@ -0,0 +1,150 @@
...
+
+=item "f1" (B<OSSL_KDF_PARAM_ARGON2_FLAG_CLEAR_PWD>) <unsigned integer>
+
+If set (nonzero), B<OPENSSL_cleanse> is called on password during initial
+hash phase.
+
+=item "f2" (B<OSSL_KDF_PARAM_ARGON2_FLAG_CLEAR_SECRET>) <unsigned integer>
+
+If set (nonzero), B<OPENSSL_cleanse> is called on secret during initial
+hash phase.
```
t8m, line 68: These flags have different names in the core_names.h file. Also I'm not sure we want them. What is the rationale? No other KDF implementation has anything like this.

doc/man7/EVP_KDF-ARGON2.pod
```diff
@@ -0,0 +1,150 @@
...
+The number of threads. This can only be used with built-in thread support.
+Threading must be explicitly enabled. See EXAMPLES section for more information.
+
+=item "s" (B<OSSL_KDF_PARAM_ARGON2_SZ>) <unsigned integer>
+
+Digest length.
+
+=item "a" (B<OSSL_KDF_PARAM_ARGON2_AD>) <octet string>
+
+Optional associated data.
```
t8m, line 46: Maybe add some more description to what this is used for.

doc/man7/EVP_KDF-ARGON2.pod
```diff
@@ -0,0 +1,150 @@
...
+
+=item "pass" (B<OSSL_KDF_PARAM_PASSWORD>) <octet string>
+
+=item "salt" (B<OSSL_KDF_PARAM_SALT>) <octet string>
+
+=item "secret" (B<OSSL_KDF_PARAM_SECRET>) <octet string>
+
+=item "iter" (B<OSSL_KDF_PARAM_ITER>) <unsigned integer>
+
+These parameters work as described in L<EVP_KDF(3)/PARAMETERS>.
```
t8m, line 33: It is not clear how secret vs. pass is used within Argon2.

providers/implementations/digests/blake2b_prov.c
```diff
@@ -17,9 +17,49 @@
...
+#include <openssl/proverr.h>
+#include <openssl/err.h>
 #include "blake2_impl.h"
 #include "prov/blake2.h"
 
+static const OSSL_PARAM known_blake2b_settable_ctx_params[] = {
+    {OSSL_DIGEST_PARAM_XOFLEN, OSSL_PARAM_UNSIGNED_INTEGER, NULL, 0, 0},
+    OSSL_PARAM_END
+
+};
```
t8m, line 30: I'd like to see at least one or two testcases in blake2b evp test file that will test the XOFLEN parameter.

t8m (changes requested):

doc/man7/EVP_KDF-ARGON2.pod
```diff
@@ -0,0 +1,150 @@
...
+using 2 lanes, 2 threads, and memory cost of 65536:
+
+ int test(void)
+ {
+    EVP_KDF *kdf = NULL;
+    EVP_KDF_CTX *kctx = NULL;
+    OSSL_PARAM params[11], *p = params;
+    uint8_t result[128];
+
+    int outlen, lanes = 2, threads = 2, memcost = 65536;
```
matbech, line 84: outlen and out is unused.

doc/man7/EVP_KDF-ARGON2.pod
ckalina: Would simply specifying `Secret, or pepper, is optional.` under the `secret` =item work?

doc/man7/EVP_KDF-ARGON2.pod
ckalina: Would something like `Optional associated data, may be used to "tag" a group of keys, or tie them to a particular public key, without having to modify salt.` work?

vanessalcataldo:
How many times do I need to keep unsubscribing to something I never received before? Stop spamming me!


Vanessa Cataldo 

On Tuesday, November 29, 2022, 8:09 AM, ckalina ***@***.***> wrote:




@ckalina commented on this pull request.

In doc/man7/EVP_KDF-ARGON2.pod:
> +
+=head2 Supported parameters
+
+The supported parameters are:
+
+=over 4
+
+=item "pass" (B<OSSL_KDF_PARAM_PASSWORD>) <octet string>
+
+=item "salt" (B<OSSL_KDF_PARAM_SALT>) <octet string>
+
+=item "secret" (B<OSSL_KDF_PARAM_SECRET>) <octet string>
+
+=item "iter" (B<OSSL_KDF_PARAM_ITER>) <unsigned integer>
+
+These parameters work as described in L<EVP_KDF(3)/PARAMETERS>.

Would simply specifying Secret, or pepper, is optional. under the secret =item work?

doc/man7/EVP_KDF-ARGON2.pod
t8m: Yeah that would WFM. Please also note there is a bug in the param name - it is `ad` not `a`.

doc/man7/EVP_KDF-ARGON2.pod
t8m: What about `Secret is an optional secret data used along the password.`?

t8m:
> How many times do I need to keep unsubscribing to something I never received before? Stop spamming me

@vanessalcataldo you have to unsubscribe from the openssl repository if you watch it.

doc/man7/EVP_KDF-ARGON2.pod
ckalina: The rationale for these flags is:

These inputs are used to calculate the initial hash H_0 during `kdf_argon2_derive` in `initial_hash` and they are not required afterwards for the current derivation. They get ultimately zeroed with `OPENSSL_clear_free` in `kdf_argon2_free` on free, `kdf_argon2_reset` on reset, or `kdf_argon2_ctx_set_pwd` on that param change.

If one wishes to derive a new key with, for example, a different ad, one may just provide a new octet string param `OSSL_KDF_PARAM_ARGON2_AD` and call derive again. This saves the user from storing the password or secret after having initialized the Argon2 context.

If that's not required one only has to explicitly set these clear flags and `OPENSSL_cleanse` is called right after initialization, not after any of the above operations which may occur substantially later.

t8m: Could you please put some of the rationale into the manual page text?  Something like: `This can be useful if multiple keys with different ad value are to be generated from a single password and secret.`

ckalina: Certainly, would something like this work?
```
=item "f1" (B<OSSL_KDF_PARAM_ARGON2_FLAG_CLEAR_PWD>) <unsigned integer>

=item "f2" (B<OSSL_KDF_PARAM_ARGON2_FLAG_CLEAR_SECRET>) <unsigned integer>

If set (non-zero), password (or secret) stored in Argon2 context is zeroed
early during initial hash computation, as soon as it's not needed.

It's zeroed along the rest of Argon2 context data on clear, free, reset,
password (or secret) param assignment operations otherwise.

This can be useful if, for example, multiple keys with different ad value
are to be generated from a single password and secret.
```

doc/man7/EVP_KDF-ARGON2.pod
```diff
@@ -0,0 +1,150 @@
...
+
+    if (EVP_KDF_derive(kctx, &result, 128, params) != 1)
+        goto fail;
+
+    printf("Output = %s\n", OPENSSL_buf2hexstr(result, 128));
+
+ fail:
+    EVP_KDF_free(kdf);
+    EVP_KDF_CTX_free(kctx);
+    CRYPTO_THREAD_disable(NULL);
```
ckalina, line 114: Replace CRYPTO_THREAD_{enable,disable} -> OSSL_set_max_threads.

doc/man7/EVP_KDF-ARGON2.pod
t8m: Hmm, I am still not convinced it is that useful because if I am not going to reuse the KDF CTX to derive multiple keys I'll probably call EVP_KDF_CTX_free() just after the derive call and thus the stored password/secret is zeroized immediately after the derivation anyway. But if I want to reuse the context to derive multiple keys later I cannot use these flags. Or do I misunderstand?

ckalina: You're right, if you want to re-use the context, you have to do it before a call to EVP_KDF_CTX_free is made and you cannot use these flags.

If low memory cost/iter count is used, derive will likely finish quickly, however, derive can take quite a while otherwise (especially with large memcost). I think that's when you might want to use these, if you don't plan on genering multiple keys, no point for the data to hang around.

ckalina: (I suppose we could merge them into a single flag that could be re-used elsewhere, if adding too many flags is a concern.)

providers/implementations/digests/blake2b_prov.c
ckalina: https://github.com/openssl/openssl/pull/12256/commits/20f84d7f8d6d08393104e3246aa637a4ec5e5ac8 https://github.com/openssl/openssl/pull/12256/commits/d7425e74e904868de357720a4f6188fadf1b42ef

providers/implementations/kdfs/argon2.c
ckalina: `ctx->outlen` is set to a default value of 64 during initialization and left unchanged provided `OSSL_KDF_PARAM_SIZE` isn't set.

This value is not the minimal supported, e.g., `outlen = 32` passed to derive is still ok as far as Argon2 is concerned. In this case, if `sz` means `outlen`, we might write past `out` in some cases.

I changed it a bit here: https://github.com/openssl/openssl/blob/b94350f527e540e6262069c16f09b8db6f0c6db8/providers/implementations/kdfs/argon2.c#L1108

@slontis Do you think there's anything to be done about this hunk?

doc/man7/EVP_KDF-ARGON2.pod
ckalina: Refactored the two tags into a generic `OSSL_KDF_PARAM_EARLY_CLEAN`.

providers/implementations/kdfs/argon2.c
ckalina: Unless there are objections, marking as resolved. Please feel free to unresolve if there's anything to do about this hunk.

CHANGES.md
```diff
@@ -253,6 +253,11 @@ OpenSSL 3.2
 
    *Maxim Mikityanskiy*
 
+ * Support for Argon2d, Argon2i, Argon2id KDFs has been added along with
+   basic thread pool implementation for select platforms.
+
+   *Cestmir Kalina*
```
t8m, line 259: You can use UTF-8 characters in your name. No need to stay with pure ASCII.

doc/man7/EVP_KDF-ARGON2.pod
```diff
@@ -0,0 +1,169 @@
...
+
+Support for computing the B<argon2> password-based KDF through the B<EVP_KDF>
+API.
+
+The EVP_KDF-ARGON2 algorithm implements the Argon2 password-based key
+derivation function, as described in IETF RFC 9106.  It is memory-hard in
+the sense that it deliberately requires a significant amount of RAM for efficient
+computation. The intention of this is to render brute forcing of passwords on
+systems that lack large amounts of main memory (such as GPUs or ASICs)
+computationally infeasible.
```
t8m, line 17: I am missing at least short mention of the ARGON2I/D/ID variants and at least some minimal description of the differences.

doc/man7/EVP_KDF-ARGON2.pod
```diff
@@ -0,0 +1,169 @@
...
+Output size.
+
+=item "ad" (B<OSSL_KDF_PARAM_ARGON2_AD>) <octet string>
+
+Optional associated data, may be used to "tag" a group of keys, or tie them
+to a particular public key, without having to modify salt.
+
+=item "lanes" (B<OSSL_KDF_PARAM_ARGON2_LANES>) <unsigned integer>
+
+Number of lanes.
```
t8m, line 53: I'd like to see some description of relation between the `lanes` and `threads` parameter. I.e., that for maximum efficiency of the computation lanes == threads. But threads should not exceed the number of CPU cores available, otherwise it does not bring any speed up. (If I understand the Argon2 design right.)

doc/man7/EVP_KDF-ARGON2.pod
ckalina: Added:
```
Argon2d (Argon2i) uses data-dependent (data-independent) memory access and primary seek to address trade-off (side-channel) attacks.

Argon2id is a hybrid construction which, in the first two slices of the first pass, generates reference addresses data-independently as in Argon2i, whereas in in later slices and next passess it generates them data-dependently as in Argon2d.

Sbox-hardened version Argon2ds is not supported.

For more information, please refer to RFC 9106.
```

Does that work?

t8m: LGTM

doc/man7/EVP_KDF-ARGON2.pod
t8m: Yes, good

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1531 @@
...
+    c->outlen = ARGON2_DEFAULT_OUTLEN;
+    c->t_cost = ARGON2_DEFAULT_T_COST;
+    c->m_cost = ARGON2_DEFAULT_M_COST;
+    c->lanes = ARGON2_DEFAULT_LANES;
+    c->threads = ARGON2_DEFAULT_THREADS;
+    c->version = ARGON2_DEFAULT_VERSION;
+    c->early_clean = 0;
+    c->type = type;
+    c->mac = EVP_MAC_fetch(c->libctx, "blake2bmac", NULL);
+    c->md = EVP_MD_fetch(c->libctx, "blake2b512", NULL);
```
t8m, line 996: These fetches should be checked for NULL return. Also there should be a way to set the propquery as it is done for other KDFs when fetching the used digests or ciphers. I.e. the propquery is being passed in as a parameter OSSL_ALG_PARAM_PROPERTIES see ossl_prov_digest_load_from_params(). It's different in this case that the digest itself is not changeable but the propquery stil should be IMO.

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1531 @@
...
+    type = ctx->type;
+    libctx = ctx->libctx;
+    mac = ctx->mac;
+    md = ctx->md;
+
+    if (mac == NULL)
+        mac = EVP_MAC_fetch(libctx, "blake2bmac", NULL);
+
+    if (md == NULL)
+        md = EVP_MD_fetch(ctx->libctx, "blake2b512", NULL);
```
t8m, line 1180: This refetch is probably not useful here.

doc/man7/EVP_KDF-ARGON2.pod
```diff
@@ -0,0 +1,192 @@
...
+computation. The intention of this is to render brute forcing of passwords on
+systems that lack large amounts of main memory (such as GPUs or ASICs)
+computationally infeasible.
+
+Argon2d (Argon2i) uses data-dependent (data-independent) memory access and
+primary seek to address trade-off (side-channel) attacks.
+
+Argon2id is a hybrid construction which, in the first two slices of the first
+pass, generates reference addresses data-independently as in Argon2i, whereas
+in in later slices and next passess it generates them data-dependently as in
```
FdaSilvaYY, line 24: nit: in in

providers/implementations/digests/blake2_prov.c
```diff
@@ -35,7 +35,78 @@ IMPLEMENT_digest_functions(blake2s256, BLAKE2S_CTX,
...
+{
+    return ossl_prov_is_running() && ossl_blake2b_set_ctx_params(ctx, params)
+        && ossl_blake2b512_init(ctx);
+}
+
+static void *blake2b512_newctx(void *prov_ctx)
+{
+    struct blake2b_md_data_st *ctx;
+
+    ctx = ossl_prov_is_running()? OPENSSL_zalloc(sizeof(*ctx)) : NULL;
```
hlandau, line 56: nit: space before `?`

providers/implementations/digests/blake2b_prov.c
```diff
@@ -17,9 +17,49 @@
...
+#include <openssl/core_names.h>
+#include <openssl/proverr.h>
+#include <openssl/err.h>
 #include "blake2_impl.h"
 #include "prov/blake2.h"
 
+static const OSSL_PARAM known_blake2b_settable_ctx_params[] = {
+    {OSSL_DIGEST_PARAM_XOFLEN, OSSL_PARAM_UNSIGNED_INTEGER, NULL, 0, 0},
+    OSSL_PARAM_END
+
```
hlandau, line 29: nit: blank line

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1594 @@
...
+# define ARGON2_MAX_LANES 0xFFFFFFu
+# define ARGON2_MIN_THREADS 1u
+# define ARGON2_MAX_THREADS 0xFFFFFFu
+# define ARGON2_SYNC_POINTS 4u
+# define ARGON2_MIN_OUT_LENGTH 4u
+# define ARGON2_MAX_OUT_LENGTH 0xFFFFFFFFu
+# define ARGON2_MIN_MEMORY (2 * ARGON2_SYNC_POINTS)
+# define ARGON2_MIN(a, b) ((a) < (b) ? (a) : (b))
+# define ARGON2_MAX_MEMORY_BITS                                               \
+    ARGON2_MIN(32u, (sizeof(void *) * CHAR_BIT - 10 - 1))
```
hlandau, line 59: `sizeof(void *)` seems wrong here, this is not the same as a word (e.g. `sizeof(void *)` can be 128 bits).

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1594 @@
...
+# define ARGON2_MIN_PWD_LENGTH 0u
+# define ARGON2_MAX_PWD_LENGTH 0xFFFFFFFFu
+# define ARGON2_MIN_AD_LENGTH 0u
+# define ARGON2_MAX_AD_LENGTH 0xFFFFFFFFu
+# define ARGON2_MIN_SALT_LENGTH 8u
+# define ARGON2_MAX_SALT_LENGTH 0xFFFFFFFFu
+# define ARGON2_MIN_SECRET 0u
+# define ARGON2_MAX_SECRET 0xFFFFFFFFu
+# define ARGON2_BLOCK_SIZE 1024
+# define ARGON2_QWORDS_IN_BLOCK ARGON2_BLOCK_SIZE / 8
```
hlandau, line 73: Add parens for safety?

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1594 @@
...
+    if (IS_LITTLE_ENDIAN) {
+        memcpy(&w, src, sizeof(w));
+        return w;
+    }
+
+    for (i = 0; i < 8; i++)
+        w |= ((uint64_t)src[i] << (8 * i));
+
+    return w;
+}
```
hlandau, line 290: The following code does not require testing endianness and generates equally performant code on modern GCC, perhaps this would be better.

```
uint64_t load64(const uint8_t *src)
{
    return 
      (((uint64_t)src[0]) << 0)
    | (((uint64_t)src[1]) << 8)
    | (((uint64_t)src[2]) << 16)
    | (((uint64_t)src[3]) << 24)
    | (((uint64_t)src[4]) << 32)
    | (((uint64_t)src[5]) << 40)
    | (((uint64_t)src[6]) << 48)
    | (((uint64_t)src[7]) << 56);
}
```

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1594 @@
...
+    uint8_t *p = (uint8_t *)dst;
+
+    if (IS_LITTLE_ENDIAN) {
+        memcpy(dst, &w, sizeof(w));
+        return;
+    }
+
+    for (i = 0; i < 4; i++)
+        p[i] = (uint8_t)(w >> (8 * i));
+}
```
hlandau, line 304: Likewise for this

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1594 @@
...
+    uint8_t *p = (uint8_t *)dst;
+
+    if (IS_LITTLE_ENDIAN) {
+        memcpy(dst, &w, sizeof(w));
+        return;
+    }
+
+    for (i = 0; i < 8; i++)
+        p[i] = (uint8_t)(w >> (8 * i));
+}
```
hlandau, line 318: Likewise for this

hlandau (changes requested):

providers/implementations/kdfs/argon2.c
ckalina: Added `ERR_raise_data(ERR_LIB_PROV, PROV_R_INVALID_MEMORY_SIZE, "cannot allocate required memory");` on memory allocation failure.

`ARGON2_MAX_MEMORY_BITS` didn't really provide a tight upper-bound in the first place, so I think it's best removed altogether. Kept the Argon2 mem upper limit to adhere to specs. Does this work better you think?

hlandau: Sure.

doc/man7/EVP_KDF-ARGON2.pod
```diff
@@ -0,0 +1,192 @@
...
+     /* argon2 params, please refer to RFC9106 for recommended defaults */
+     uint32_t lanes = 2, threads = 2, memcost = 65536;
+     char pwd[] = "1234567890", salt[] = "saltsalt";
+
+     /* derive result */
+     size_t outlen = 128;
+     unsigned char result[outlen];
+
+     /* required if threads > 1 */
+     if (OSSL_set_max_threads(NULL, threads) != 1)
```
docularxu, line 126: Sorry to jump in. I found this pull request because I traced down the internal API: ossl_crypto_thread_start()/_join(), which were merged into master branch some time ago. Actually I'm thinking about another use case of multi-threads. Then, it leads me to this question: how can you be sure Argon2 is the only user of threads? how can you be so sure to set the max number of threads only according to only Argon2's need?

What if there is other users (maybe internal ossl frameworks) who set the max number already? How to address such conflicts?

kroeckx: Not having looked at any of the implementations, and the manpage doesn't say anything about it, but I would expect the OSSL_set_max_threads() call to set a limit per libctx. So you can create a different libctx for other usecases, and that call should not use NULL for the libctx.

docularxu: > the OSSL_set_max_threads() call to set a limit per libctx.

Right, it sets a limit per libctx. This means within one libctx, there can be only one 'usecase' of the threadpool. How could this be justified? Suppose one method from one provider uses threadpool, and another method from another provider uses threadpool too. Then, these two providers cannot be loaded into the same libctx. So, here we are setting _implicit_  rules about loading providers.

I don't think this is expected, isn't it?

mattcaswell: We have to assume that the user and creator of the libctx knows what it is going to be used for and what sensible limits should be enforced.

> Then, these two providers cannot be loaded into the same libctx.

Err...yes they can? They may be sharing a thread pool but that isn't a problem per se if the pool has been adequately sized.

Although that does make me think of a different problem. Many (most?) providers will create their own libctx independent of the user libctx (or a child libctx based on the user one). Those libctx's will have a separate thread pool - which is fine (and means different providers won't interfere with each other) but does mean then end user has no real control over the "total" size of the thread pool.

Anyway this is kind of tangential to this Argon2 PR so discussion should probably be in some separate issue.

docularxu: > They may be sharing a thread pool but that isn't a problem per se if the pool has been adequately sized.

Still, they can steal the thread (passing the per user's upper limit, not max) from each other, right? I put more details in a new issue: https://github.com/openssl/openssl/issues/20059

> Anyway this is kind of tangential to this Argon2 PR so discussion should probably be in some separate issue.

Agree, and I submitted a new issue https://github.com/openssl/openssl/issues/20059

ckalina: If that's fine with you, @docularxu, I am resolving this conversation and we can continue the conversation in the two other issues.

ckalina:
Thank you for your reviews. All of the outstanding comments/issues have been resolved and code rebased.

mbroz:
I there a plan in which OpenSSL release this PR will land?

I see it mentioned in https://www.openssl.org/roadmap.html but without details.

mbroz:
Any update here, or it is just waiting for review? (@hlandau ?)

Sorry for poking again, just I see another issues with emebedded Argon2 code in cryptsetup and I wish I can finally use a proper crypto library API for it...

providers/implementations/kdfs/argon2.c
```diff
@@ -0,0 +1,1584 @@
...
+    return 1;
+}
+
+static int kdf_argon2_ctx_set_out_length(KDF_ARGON2 *ctx, uint32_t outlen)
+{
+    /* ARGON2_MAX_OUT_LENGTH == max outlen value, so upper bounds checks
+     * are always satisfied; to suppress compiler if statement tautology
+     * warnings, these checks are skipped; however, to ensure that these
+     * limits are met and implementation conforming to Argon2 RFC, we need
+     * to fix the type */
```
t8m, line 1247: Nit: Please fix the formatting of the multiline comment (beginning and end marker on separate line)

t8m (commented):
Apart from these nits this looks good

ckalina:
Thank you, resolved and rebased.

t8m:
@ckalina the CI failure is relevant, please fix.

t8m (approved):

ckalina:
@t8m errrors causing test failure were regenerated

test/evp_test.c
```diff
@@ -3948,6 +4002,12 @@ static int parse(EVP_TEST *t)
                 return 0;
             }
             t->reason = take_value(pp);
+        } else if (strcmp(pp->key, "Threads") == 0) {
```
paulidale, line 4005: Should "Threads" be lowercase like the rest?

paulidale (changes requested):
One possibly typographic issue.

There should be a mention of this in `doc/man7/OSSL_PROVIDER-default.pod`.  It's just this added (with blank lines in the _Key Derivation Function_ section:

```pod
=item ARGON2, see L<EVP_KDF-ARGON2(7)>
```

Approved these are addressed, I'll approve.

A marathon effort, thanks for your persistence.

test/evp_test.c
t8m: Nope, the others are ctrl names which are lowercase for other algorithms too. But this one is a new key name for test parameters and these are always with a capital letter at the beginning.

paulidale: Okay, no need to update.

paulidale (approved):
Finally!

t8m (approved):

paulidale:
And merged.  Thanks for the perseverance on this.

mbroz:
Thanks everyone!

For the record how long marathon this was, it started with @ckalina [bachelor thesis](https://is.muni.cz/th/uvuxj/?lang=en) in 2020 ... Just a few years and we are finally there :-)

mattcaswell:
Good news!! Thanks @ckalina for all your efforts!

PR #12257, opened 2020-06-24
INSTALL.md: Restore $ as command prompt indicator

levitte:
We have a notational convention in INSTALL.md, which says this among
others:

> Any line starting with a dollar sign is a command line.
>
>     $ command
>
> The dollar sign indicates the shell prompt and is not to be entered as
> part of the command.

That notation exists to make it clear what is a command line and
what's output from that command line.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1026,7 +1026,7 @@ configuration.  The following variables are supported:
 These cannot be mixed with compiling/linking flags given on the command line.
 In other words, something like this isn't permitted.
 
-    ./config -DFOO CPPFLAGS=-DBAR -DCOOKIE
+    $ ./config -DFOO CPPFLAGS=-DBAR -DCOOKIE
 
 Backward compatibility note:
 
@@ -1038,11 +1038,11 @@ for the following:
 
 For example, the following command will not see -DBAR:
 
-    CPPFLAGS=-DBAR ./config -DCOOKIE
+    $ CPPFLAGS=-DBAR ./config -DCOOKIE
 
 However, the following will see both set variables:
 
-    CC=gcc CROSS_COMPILE=x86_64-w64-mingw32- ./config -DCOOKIE
+    $ CC=gcc CROSS_COMPILE=x86_64-w64-mingw32- ./config -DCOOKIE
 
 If CC is set, it is advisable to also set CXX to ensure both the C and C++
 compiler are in the same "family".  This becomes relevant with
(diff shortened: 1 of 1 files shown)


mattcaswell (approved):

richsalz:
Does OpenSSL have the issue in other places, like manpages?  "git grep '^ '" might be useful.

DDvO:
Repeating some comments of mine that apparently were ignored being in https://github.com/openssl/openssl/pull/12175/files#r444780398 and below it:

I am not happy about that `$ ` notation, mostly because when there is a number of shell commands in a row, such as in
```
    $ mkdir /var/tmp/openssl-build
    $ cd /var/tmp/openssl-build
    $ whatever
```
you cannot simply copy&paste that series of commands, while with
```
    mkdir /var/tmp/openssl-build
    cd /var/tmp/openssl-build
    whatever
```
you can.

And I suppose there are just a few cases where, leaving it out, command input and output could be confused.
In such cases we could, e.g., insert a blank line to help understanding this distinction.

Moreover, the `$` symbol is not knows to Windows users, where typically a `>` is used as prompt symbol.

richsalz:
Some examples intermix command and output and the "$" prompt is used to mark off those.  

I notice that "!" is used as a comment char for the OpenVMS line and I'm not sure it should get special treatment.

DDvO:
Is there a way in MarkDown to render those `$ ` in a way that they are not copied when selecting and copying text including them?
In HTML for instance there are some tricks for making text non-selectable.

richsalz:
There is no way to make something non-selectable in markdown.

richsalz:
(FWIW, my prompt is a semicolon

DDvO:
> There is no way to make something non-selectable in markdown.

I see :-( but thanks for the info!

mspncp (approved):
Many of the code examples are really just examples and are not be meant to be copy&pasted literally. Some people might be tempted to blindly copy&paste large code blocks without checking the code. So having to edit the lines gives them a little extra time to switch on their brain. Like in the following code block:

```
    $ make clean                                     # Unix
    $ mms clean                                      ! (or mmk) OpenVMS
    $ nmake clean                                    # Windows
```

_P.S: programmers usually tend to have smart editors which are capable of block editing.  ;-)_

levitte:
Merged

270540fd5413b00a746a581e8939c30862c689b1 INSTALL.md: Restore $ as command prompt indicator

PR #12258, opened 2020-06-24
apps: avoid memory overrun.

paulidale:
NULL terminate the built in "help" `argv` array to avoid reading beyond the end.

Fixes #12246

PR #12259, opened 2020-06-24
apps/openssl: Fix buffer-overflow for command with no arguments

levitte:
The 'openssl' command without arguments was intended to behave like
'openssl help'.  However, the array used to synthesize the 'help'
command lacked a terminating NULL, making it prone to buffer
overflow.

We also remove code in help_main() that duplicates this case, which is
now handled in main().

### DIFF
--- a/apps/openssl.c
+++ b/apps/openssl.c
@@ -205,7 +205,7 @@ static void setup_trace(const char *str)
 }
 #endif /* OPENSSL_NO_TRACE */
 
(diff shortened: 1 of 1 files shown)


levitte:
Didn't notice #12258 before now.  This also contains a bit of cleanup, though, so I'll keep it open

mspncp:
I would approve the rest, but the commit message needs to be adopted.

levitte:
Closed because #12258

Issue #11918, opened 2020-05-22
Bugs found by static tests


tester7632:
Running the latest git version 2de6466 through sonarqube revealed some potential bugs:

https://sonarcloud.io/project/issues?id=tester7632_openssl&resolved=false&severities=BLOCKER&types=BUG

There are quite a number of "bugs" reported by the tool, but as far as I can see most are only "not best practice" kind of bugs and not worth the time to fix them. The one listed as BLOCKER look like they should get a closer look though.

Complete report:
https://sonarcloud.io/dashboard?id=tester7632_openssl

kroeckx:
This looks like yet an other tool with 99% false positives.

richsalz:
I looked at one of the "blocker" items, https://sonarcloud.io/project/issues?id=tester7632_openssl&open=AXI9Kb9LAi29hCJ8RXj4&resolved=false&severities=BLOCKER&types=BUG.  It is wrong, complaining about a condition that can't happen.

The project uses Coverity static scanning.  I think someone will have to review all of the issues here and open an issue for the ones that are valid.  I doubt the team is interested in doing this.

kroeckx:
You really randomly looked at the same issue I did? :)

richsalz:
I looked at code that I originally wrote :)  And I didn't see your comment until after I posted mine.

tester7632:
Yes this tool reports a lot of false positives and everything that is not "best practice" 
Thats why I only pointed out the blocker items as I can not reliably tell if everything listed there is bogus and in the case it is not ...

BenBE:
Just had a closer look at
https://sonarcloud.io/project/issues?id=tester7632_openssl&open=AXI9KcWbAi29hCJ8RZQN&resolved=false&severities=BLOCKER&types=BUG
(OOB memory access in `opt_help`)
which is a OOB stack write after the array `start` if `MAX_OPT_HELP_WIDTH > 80` (plus some other conditions). Not an immediate issue AFAICS, as the passed in data at the point seems to be constant.

HAven't looked at any other of the reports, but from my experience, once those tools can tell you some consistent steps to follow along, there's usually also some minor detail, that the code misses. In this case, the static buffer size for `start`.

nhorman:
Marking as inactive, to be closed at the end of 3.4 dev, barring further input

t8m:
Closing as the issue was labeled as inactive for extended period of time and there was no additional input.
Please add a comment if you have any further information to support reopening the issue.

PR #12265, opened 2020-06-24
Avoid potential OOB if width > sizeof(start)

BenBE:
This can't currently happen due to `sizeof(start)` being way larger than `MAX_OPT_HELP_WIDTH`, but wasn't (properly) checked for previously. With this patch there still remains one (static) OOB, when the length of the option name and the `valtype2param` string for that argument together  overflow the buffer in `opt_print`. This is kinda unlikely, unless someone intentionally crafts a
long option name, in which case this would become some trivial stack buffer overrun with
possibility to overwrite the pointer to the `OPTIONS` structure (a long `o->name` is the critical path here). All current documentation for options seem to be sane here ATM.

I sincerely hope we trust our built-in documentation to not exploit ourselves.

##### Checklist

### DIFF
--- a/apps/lib/opt.c
+++ b/apps/lib/opt.c
@@ -1111,53 +1111,62 @@ static void opt_print(const OPTIONS *o, int doingparams, int width)
     char start[80 + 1];
     char *p;
 
-        help = o->helpstr ? o->helpstr : "(No additional info)";
-        if (o->name == OPT_HELP_STR) {
-            opt_printf_stderr(help, prog);
-            return;
-        }
-        if (o->name == OPT_SECTION_STR) {
-            opt_printf_stderr("\n");
-            opt_printf_stderr(help, prog);
-            return;
-        }
-        if (o->name == OPT_PARAM_STR) {
-            opt_printf_stderr("\nParameters:\n");
-            return;
-        }
+    /* Avoid OOB if width is beyond the buffer size of start */
+    if (width >= (int)sizeof(start))
+        width = (int)sizeof(start) - 1;
 
-        /* Pad out prefix */
-        memset(start, ' ', sizeof(start) - 1);
-        start[sizeof(start) - 1] = '\0';
+    help = o->helpstr ? o->helpstr : "(No additional info)";
+    if (o->name == OPT_HELP_STR) {
+        opt_printf_stderr(help, prog);
+        return;
+    } else if (o->name == OPT_SECTION_STR) {
+        opt_printf_stderr("\n");
+        opt_printf_stderr(help, prog);
+        return;
+    } else if (o->name == OPT_PARAM_STR) {
+        opt_printf_stderr("\nParameters:\n");
+        return;
+    }
 
-        if (o->name == OPT_MORE_STR) {
-            /* Continuation of previous line; pad and print. */
-            start[width] = '\0';
-            opt_printf_stderr("%s  %s\n", start, help);
-            return;
-        }
+    /* Pad out prefix */
+    memset(start, ' ', sizeof(start) - 1);
+    start[sizeof(start) - 1] = '\0';
 
-        /* Build up the "-flag [param]" part. */
-        p = start;
-        *p++ = ' ';
-        if (!doingparams)
-            *p++ = '-';
-        if (o->name[0])
-            p += strlen(strcpy(p, o->name));
-        else
-            *p++ = '*';
-        if (o->valtype != '-') {
-            *p++ = ' ';
-            p += strlen(strcpy(p, valtype2param(o)));
-        }
-        *p = ' ';
-        if ((int)(p - start) >= MAX_OPT_HELP_WIDTH) {
-            *p = '\0';
-            opt_printf_stderr("%s\n", start);
-            memset(start, ' ', sizeof(start));
-        }
+    if (o->name == OPT_MORE_STR) {
+        /* Continuation of previous line; pad and print. */
         start[width] = '\0';
         opt_printf_stderr("%s  %s\n", start, help);
+        return;
+    }
+
+    /* Build up the "-flag [param]" part. */
+    p = start;
+
+    *p++ = ' ';
+
+    if (!doingparams)
+        *p++ = '-';
+
+    if (o->name[0])
+        p += strlen(strcpy(p, o->name));
+    else
+        *p++ = '*';
+
+    if (o->valtype != '-') {
+        *p++ = ' ';
+        p += strlen(strcpy(p, valtype2param(o)));
+    }
+
+    *p = ' ';
+
+    if ((int)(p - start) >= MAX_OPT_HELP_WIDTH) {
+        *p = '\0';
+        opt_printf_stderr("%s\n", start);
+        memset(start, ' ', sizeof(start));
+    }
+
+    start[width] = '\0';
+    opt_printf_stderr("%s  %s\n", start, help);
 }
 
 void opt_help(const OPTIONS *list)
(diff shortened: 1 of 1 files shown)


BenBE:
NB: Patch created based on the issue I commented on in #11918.

slontis:
I think this is really another case of a false positive..
opt_print() is 'currently' only accessed by opt_help().
And the options are all a fixed size.

You will see that opt_help() also has a buffer  char start[80 + 1];
and does OPENSSL_assert(i < (int)sizeof(start));

So would it be sufficient to just change that code to do the following?

```
#define MAX_HELP_OPT_SIZE (80 + 1)

opt_print(...)
{
    char start[MAX_HELP_OPT_SIZE];

    OPENSSL_assert(width < (int)sizeof(start));
   ...
   ...
}

opt_help(...)
{
    char start[MAX_HELP_OPT_SIZE];
}

```

BenBE:
No, as this won't fix the issue in case `MAX_OPT_HELP_WIDTH > MAX_OPT_HELP_SIZE`.

I see there is little incentive in fixing it as all inputs to those functions are static. But better do something properly and avoid problems in the future. I put the "I hope we trust our documentation" for a reason. ;-)

And as shown in the patches, there are two OOB issues here. The `width>sizeof(buf)` one seems unexploitable, as the necessary checks are there, but the second one (using the large `opt->name` is there and easily translates into RCE, if you just fancied to write some option string long enough to overwrite the function return address.

Also does the current setup allocate the memory buffer `start` twice on the stack (once in `opt_help` and once in `opt_print`). You can avoid inconsistencies between those routines by using a define, or avoid this source of inconsistency altogether by eliminating the second stack buffer, which is unused (apart from the length check).

My rewrite of those routines avoids those issues completely, and if format strings in `BIO_printf` support width formatting (`%*s`) you could even get rid of that `start` buffer completely AFAIC. Haven't checked this, but the patch for this would be easy.

IMO leaving the current implementation as is leaves a taste of bad code hygiene.

mattcaswell:
> No, as this won't fix the issue in case MAX_OPT_HELP_WIDTH > MAX_OPT_HELP_SIZE.

Perhaps define `MAX_OPT_HELP_SIZE` in terms of `MAX_OPT_HELP_WIDTH` so that this can never be the case?

BenBE:
> > No, as this won't fix the issue in case MAX_OPT_HELP_WIDTH > MAX_OPT_HELP_SIZE.
> 
> Perhaps define `MAX_OPT_HELP_SIZE` in terms of `MAX_OPT_HELP_WIDTH` so that this can never be the case?

Doesn't make much sense in my eyes, as they are independent of each other:
```
 -name_of_very_long_option_that_may_overflow_the_buffer_as_this_name_here_is_not_bounded_by_anything_in_the_option_definition_and_thus_can_be_extend_to_infinity_or_until_I_run_out_of_ideas_to_make_this_ridiculous_option_name_even_longer PEM|DER@ENGINE // <-- MAX_OPT_HELP_SIZE
<-- - - - - MAX_OPT_HELP_WIDTH - - - - -->  This is the help text for this fancy option
```

The first line overwrites (unchecked) in both strcat calls for the name AND the valtype.

Fixing the relation as suggested fixes the OOB noted in the first patch, while leaving the OOB that is fixed in the second patch.

FWIW: @richsalz suggested limitting length of `option->name` in `list_options_for_command` (https://github.com/openssl/openssl/blob/master/apps/list.c#L420) by causing a CI failure.

IMO this is a somewhat lazy approach, instead of fixing the underlaying issue(s).

slontis:
Here is an alternative that uses the same code to calculate the width and do the print.
I think it is a bit cleaner?

```
static int put_char(char *out, char ch, int max_len, int *cur_index)
{
    if (*cur_index >= max_len)
        return 0;
    if (out != NULL)
        out[*cur_index] = ch;
    *cur_index += 1;
    return 1;
}

static int put_string(char *out, const char *str, int max_len, int *cur_index)
{
    int len = (int)strlen(str);
    int remain = max_len - 1 - *cur_index;

    if (remain <= 0)
        return 0;

    if (out != NULL) {
        if (remain < len)
            len = remain;
        memcpy(out + *cur_index, str, len);
    }
    *cur_index += len;
    return 1;
}

static int do_print(const OPTIONS *o, int doingparams, char *out, size_t out_len)
{
    int ret = 1, width = 0;
    const char *help;

    help = o->helpstr ? o->helpstr : "(No additional info)";
    if (o->name == OPT_HELP_STR) {
        if (out != NULL)
            opt_printf_stderr(help, prog);
        return 0;
    }
    if (o->name == OPT_SECTION_STR) {
        if (out != NULL) {
            opt_printf_stderr("\n");
            opt_printf_stderr(help, prog);
        }
        return 0;
    }
    if (o->name == OPT_PARAM_STR) {
        if (out != NULL)
            opt_printf_stderr("\nParameters:\n");
        return 0;
    }

    /* Fill with spaces */
    if (out != NULL) {
        memset(out, ' ', out_len - 1);
        out[out_len - 1] = '\0';
    }
    if (o->name == OPT_MORE_STR) {
        if (out != NULL)
            opt_printf_stderr("%s  %s\n", out, help);
        return 0;
    }

    ret = put_char(out, ' ', out_len, &width);
    if (!doingparams)
        ret = ret && put_char(out, '-', out_len, &width);

    if (o->name[0])
        ret = put_string(out, o->name, out_len, &width);
    else
        ret = put_char(out, '*', out_len, &width);

    if (o->valtype != '-') {
        ret = ret && put_char(out, ' ', out_len, &width);
        ret = ret && put_string(out, valtype2param(o), out_len, &width);
    }

    if (out != NULL)
        opt_printf_stderr("%s  %s\n", out, help);

    if (width != 0)
        width++;    /* Include the NUL terminator in the count */
    return width;
}

void opt_help(const OPTIONS *list)
{
    const OPTIONS *o;
    int sawparams = 0, w, width = 5;
    char buffer[MAX_OPT_HELP_WIDTH + 1];

    /* Starts with its own help message? */
    if (list[0].name != OPT_HELP_STR) {
        opt_printf_stderr("Usage: %s [options]\n", prog);
        if (list[0].name != OPT_SECTION_STR)
            opt_printf_stderr("Valid options are:\n", prog);
    }

    /* Find the widest help. */
    for (o = list; o->name; o++) {
        if (o->name == OPT_PARAM_STR)
            sawparams = 1;
        w = do_print(o, sawparams, NULL, MAX_OPT_HELP_WIDTH + 1);
        if (w > width)
            width = w;
    }

    /* Now let's print. */
    sawparams = 0;
    for (o = list; o->name; o++) {
        if (o->name == OPT_PARAM_STR)
            sawparams = 1;
        do_print(o, sawparams, buffer, width);
    }
}

```

BenBE:
Well, in https://github.com/openssl/openssl/pull/12265/commits/0eb18590ac37d68a509ac0e441774f05128479b6 a changed the overall code in a way that completely gets rid of the string concatenation stuff.

Unfortunately it seems to have been stuck in the CI; if somebody could kick the CI to retest …

t8m (approved):

t8m:
IMO we should just take this.

apps/lib/opt.c
```diff
@@ -870,63 +870,67 @@ void opt_print(const OPTIONS *o, int doingparams, int width)
...
-            return;
-        }
-        if (o->name == OPT_PARAM_STR) {
-            opt_printf_stderr("\nParameters:\n");
-            return;
-        }
+    int linelen, printlen;
+
+    /* Avoid OOB if width is beyond the buffer size of start */
+    if(width >= (int)sizeof(start))
```
t8m, line 876: Nit: - please add a space after `if`

t8m (changes requested):
One minor nit

apps/lib/opt.c
BenBE: Fixed. ;-)

mattcaswell (approved):

t8m (approved):

paulidale (approved):

paulidale:
Merged, thanks for this.

PR #12266, opened 2020-06-24
Force ssl/tls protocol flags to use stream sockets

BenBE:
Prior to this patch doing something like `openssl s_client -dtls1 -tls1 ...` could cause s_client to speak TLS on a UDP socket which does not normally make much sense.

NB: Bug exists since at least OpenSSL 1.0.1c (originally found there).


t8m:
Yes, approval holds for 1.1.1 as well.

paulidale:

Issue #12145, opened 2020-06-14
Extended tests are failing cmp tests


kroeckx:
In travis, the extended tests are failing with:
```
    # ERROR: (int) 'fixture->expected == ossl_cmp_msg_check_update(fixture->cmp_ctx, fixture->msg, NULL, 0)' failed @ test/cmp_vfy_test.c:127
    # [0] compared to [1]
    # check_msg_find_cert:crypto/cmp/cmp_vfy.c:538:CMP info: trying to verify msg signature with a valid cert that..
    # check_msg_find_cert:crypto/cmp/cmp_vfy.c:540:CMP info: matches msg sender    = /O=openssl_cmp
    # check_msg_find_cert:crypto/cmp/cmp_vfy.c:544:CMP info: while msg header does not contain senderKID
    # check_msg_all_certs:crypto/cmp/cmp_vfy.c:448:CMP info: trying first normal mode using trust store
    # check_msg_with_certs:crypto/cmp/cmp_vfy.c:401:CMP warning: no extraCerts
    # check_msg_with_certs:crypto/cmp/cmp_vfy.c:401:CMP warning: no untrusted certs
    # cert_acceptable:crypto/cmp/cmp_vfy.c:275:CMP info:  considering self-issued cert from certs in trusted store with..
    # cert_acceptable:crypto/cmp/cmp_vfy.c:277:CMP info:   subject = /O=openssl_cmp
    # cert_acceptable:crypto/cmp/cmp_vfy.c:296:CMP warning: cert has expired
    not ok 3 - test_validate_msg_signature_trusted_expired
```
[...]

```
65-test_cmp_vfy.t                (Wstat: 256 Tests: 1 Failed: 1)
  Failed test:  1
  Non-zero exit status: 1
81-test_cmp_cli.t                (Wstat: 768 Tests: 7 Failed: 3)
  Failed tests:  4-5, 7
  Non-zero exit status: 3
```

@DDvO: Can you look at this?

romen:
@mpeylo would you be able to look into this?

This comment might also be related: https://github.com/openssl/openssl/pull/11998#issuecomment-643693856

DDvO:
I'm on vacation this week but will try to deal with this by tomorrow...

DDvO:
How to reproduce these 'extended' Tests?

DDvO:
[...]

PR #12275, opened 2020-06-25
Fix some CMP related msan failures

mattcaswell:
The cmp_cli test is failing in some travis builds on master due to msan failures. There are two separate problems fixed by this PR:

1) Ensure a string is properly terminated in http_client.c

In HTTP_new_bio(), if the host has a trailing '/' we took a copy of the hostname but failed to terminate it properly.

2) If an empty password is supplied still try to use it

If an empty password was supplied we ignored it and were trying to use the fallback method to read the password instead (i.e. read from stdin). However if that failed (which it always does if the cmp option -batch is used) then we were reporting that we had successfully read the password without actually setting one.

Instead, if an empty password is explicitly provided we should use it. If no password is supplied explicitly and we have no fallback method then we assume the empty password.

[extended tests]

### DIFF
--- a/apps/lib/apps_ui.c
+++ b/apps/lib/apps_ui.c
@@ -37,7 +37,7 @@ static int ui_read(UI *ui, UI_STRING *uis)
             {
                 const char *password =
                     ((PW_CB_DATA *)UI_get0_user_data(ui))->password;
-                if (password && password[0] != '\0') {
+                if (password != NULL) {
                     UI_set_result(ui, uis, password);
                     return 1;
                 }
@@ -54,6 +54,8 @@ static int ui_read(UI *ui, UI_STRING *uis)
     reader = UI_method_get_reader(ui_fallback_method);
     if (reader)
         return reader(ui, uis);
+    /* Default to the empty password if we've got nothing better */
+    UI_set_result(ui, uis, "");
     return 1;
 }
 
--- a/apps/lib/apps_ui.c
+++ b/apps/lib/apps_ui.c
@@ -20,7 +20,7 @@ static int ui_open(UI *ui)
 {
     int (*opener)(UI *ui) = UI_method_get_opener(ui_fallback_method);
 
-    if (opener)
+    if (opener != NULL)
         return opener(ui);
     return 1;
 }
@@ -37,6 +37,7 @@ static int ui_read(UI *ui, UI_STRING *uis)
             {
                 const char *password =
                     ((PW_CB_DATA *)UI_get0_user_data(ui))->password;
+
                 if (password != NULL) {
                     UI_set_result(ui, uis, password);
                     return 1;
@@ -52,7 +53,7 @@ static int ui_read(UI *ui, UI_STRING *uis)
     }
 
     reader = UI_method_get_reader(ui_fallback_method);
-    if (reader)
+    if (reader != NULL)
         return reader(ui, uis);
     /* Default to the empty password if we've got nothing better */
     UI_set_result(ui, uis, "");
@@ -84,7 +85,7 @@ static int ui_write(UI *ui, UI_STRING *uis)
     }
 
     writer = UI_method_get_writer(ui_fallback_method);
-    if (writer)
+    if (writer != NULL)
         return writer(ui, uis);
     return 1;
 }
@@ -93,7 +94,7 @@ static int ui_close(UI *ui)
 {
     int (*closer)(UI *ui) = UI_method_get_closer(ui_fallback_method);
 
-    if (closer)
+    if (closer != NULL)
         return closer(ui);
     return 1;
 }
@@ -114,7 +115,7 @@ int setup_ui_method(void)
 
 void destroy_ui_method(void)
 {
-    if (ui_method) {
+    if (ui_method != NULL) {
         UI_destroy_method(ui_method);
         ui_method = NULL;
     }
--- a/crypto/http/http_client.c
+++ b/crypto/http/http_client.c
@@ -712,10 +712,15 @@ static BIO *HTTP_new_bio(const char *server /* optionally includes ":port" */,
     }
 
     host_end = strchr(host, '/');
-    if (host_end != NULL && (size_t)(host_end - host) < sizeof(host_name)) {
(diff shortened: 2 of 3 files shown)


mattcaswell:
@DDvO

apps/lib/apps_ui.c
```diff
@@ -37,7 +37,7 @@ static int ui_read(UI *ui, UI_STRING *uis)
             {
                 const char *password =
                     ((PW_CB_DATA *)UI_get0_user_data(ui))->password;
-                if (password && password[0] != '\0') {
+                if (password != NULL) {
```
DDvO, line 40: Please add an empty line after the var decl.

DDvO: While you are at it: 
there are three cases in this source file where a pointer test is written without `!= NULL`,
which may be improved along with the other improvements: `if (opener)` above and `if (reader)` and `if (closer)` below.

apps/lib/apps_ui.c
```diff
@@ -54,6 +54,8 @@ static int ui_read(UI *ui, UI_STRING *uis)
     reader = UI_method_get_reader(ui_fallback_method);
     if (reader)
         return reader(ui, uis);
+    /* Default to the empty password if we've got nothing better */
+    UI_set_result(ui, uis, "");
```
DDvO, line 58: I'm not an expert on the password UI but according to the explanation you gave this looks good.

crypto/http/http_client.c
```diff
@@ -712,10 +712,15 @@ static BIO *HTTP_new_bio(const char *server /* optionally includes ":port" */,
...
-        /* chop trailing string starting with '/' */
-        strncpy(host_name, host, host_end - host + 1);
-        host = host_name;
+    if (host_end != NULL) {
+        size_t host_len = host_end - host;
+
+        if (host_len < sizeof(host_name)) {
+            /* chop trailing string starting with '/' */
+            strncpy(host_name, host, host_len);
+            host_name[host_len] = '\0';
```
DDvO, line 721: Fix confirmed.

DDvO:
Extended test still fail for the Travis CI build witth configuration 
`no-asm enable-ubsan enable-rc5 enable-md2 enable-ssl3 enable-ssl3-method enable-nextprotoneg no-shared enable-buildtest-c++ -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -Wno-unused-command-line-argument`.

It turns out that this is (as in #12145) due to some CMP verification steps being omitted to allow for more aggressive fuzzing in case `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` is enabled.
In such cases `test_cmp_cli` does not really make sense.

I had already tried disabling such tests in `test/recipes/81-test_cmp_cli.t` in #12175 as follows:
```
plan skip_all => "These tests are not supported in a fuzz build"
    if !disabled("fuzz-libfuzzer") || !disabled("fuzz-afl");
```
but this turns out to be insufficient here.

Can we somehow check using Perl in that script if the configuration has been done with 
`-D FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`?
The above condition `!disabled("fuzz-libfuzzer") || !disabled("fuzz-afl")` should be replaced by that new check.

DDvO:
This PR is currently blocked by the following question:
> Can we somehow check using Perl in that script if the configuration has been done with
`-D FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`?

richsalz:
Not sure the best way to do it, but this probably works:
```
use lib bldtop_dir('.');
import configdata;
if ( $config{options} =~ /-DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION/ ) {
 ...
}
```

DDvO:
> Not sure the best way to do it, but this probably works:
> 
> ```
> use lib bldtop_dir('.');
> import configdata;
> if ( $config{options} =~ /-DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION/ ) {
>  ...
> }
> ```

Thanks @richsalz - the idea is good,
I just had to adapt this because `bldtop_dir` is not yet available when importing libs.

@mattcaswell, if you add to `test/recipes/81-test_cmp_cli.t` the import of `bldtop_file` and replace
```
plan skip_all => "These tests are not supported in a fuzz build"
    if !disabled("fuzz-libfuzzer") || !disabled("fuzz-afl");
```
by
```
open(CONFIG, bldtop_file("configdata.pm"));
plan skip_all => "These tests are not supported in a fuzz build"
    if (grep { /-DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION/ } <CONFIG>);
close CONFIG;

```
the exclusion of fuzz-related builds should work fine.

mattcaswell:
Updated with a new commit that should address the travis failures. I was able to tweak @richsalz's variant a bit more to get it to work.

mattcaswell:
I just realised I forgot to address @DDvO's style comments. Fix up pushed addressing those.

DDvO:
@mattcaswell, this is ready to merge. Shall I do?

mattcaswell:
> @mattcaswell, this is ready to merge. Shall I do?

It's not been re-approved since the latest changes - so unfortunately not yet ready for merge.

t8m:
Unfortunately it is not clear whether all the extended tests are fixed because the fixup commit makes them not to run.

mattcaswell:
Darn...I remembered to add "extended tests" to my earlier fixup - but not to the subsequent one. Will add shortly.

t8m:
I've restarted the Travis build on the previous commit to find out as the fixup commit should not change anything in regards to tests.

mattcaswell:
> I've restarted the Travis build on the previous commit to find out as the fixup commit should not change anything in regards to tests

Ah! Ok - perfect. Thanks.

mattcaswell:
> I've restarted the Travis build on the previous commit to find out as the fixup commit should not change anything in regards to tests.

The Travis run passed.

mattcaswell:
Pushed. Thanks.

PR #12279, opened 2020-06-26
Improve run-tests.pl modes VFO and VFP, make VFO replace VF

DDvO:
Add a visual separator after the output of each failed test in VFO and VFP mode.

Also add a warning in case VFO (HARNESS_VERBOSE_FAILURES_ONLY) or VFP (HARNESS_VERBOSE_FAILURES_PROGRESS) is used along with VF or V overriding it,
and in case VF (HARNESS_VERBOSE_FAILURE) is used along with V (VERBOSE) overriding it.

### DIFF
--- a/test/run_tests.pl
+++ b/test/run_tests.pl
@@ -10,13 +10,15 @@ use strict;
 use warnings;
 
 # Recognise VERBOSE aka V which is common on other projects.
-# Additionally, recognise VERBOSE_FAILURE aka VF.
-# and recognise VERBOSE_FAILURE_PROGRESS aka VFP.
+# Additionally, recognise VERBOSE_FAILURE aka VF aka REPORT_FAILURES
+# and recognise VERBOSE_FAILURE_PROGRESS aka VFP aka REPORT_FAILURES_PROGRESS.
 BEGIN {
     $ENV{HARNESS_VERBOSE} = "yes" if $ENV{VERBOSE} || $ENV{V};
-    $ENV{HARNESS_VERBOSE_FAILURE} = "yes" if $ENV{VERBOSE_FAILURE} || $ENV{VF};
+    $ENV{HARNESS_VERBOSE_FAILURE} = "yes"
+        if $ENV{VERBOSE_FAILURE} || $ENV{VF} || $ENV{REPORT_FAILURES};
     $ENV{HARNESS_VERBOSE_FAILURE_PROGRESS} = "yes"
-        if $ENV{VERBOSE_FAILURE_PROGRESS} || $ENV{VFP};
+        if ($ENV{VERBOSE_FAILURE_PROGRESS} || $ENV{VFP}
+            || $ENV{REPORT_FAILURES_PROGRESS});
 }
 
 use File::Spec::Functions qw/catdir catfile curdir abs2rel rel2abs/;
--- a/test/run_tests.pl
+++ b/test/run_tests.pl
@@ -9,15 +9,14 @@
 use strict;
 use warnings;
 
-# Recognise VERBOSE and V which is common on other projects.
-# Additionally, also recognise VERBOSE_FAILURE and VF.
+# Recognise VERBOSE aka V which is common on other projects.
+# Additionally, recognise VERBOSE_FAILURE aka VF.
+# and recognise VERBOSE_FAILURE_PROGRESS aka VFP.
 BEGIN {
     $ENV{HARNESS_VERBOSE} = "yes" if $ENV{VERBOSE} || $ENV{V};
     $ENV{HARNESS_VERBOSE_FAILURE} = "yes" if $ENV{VERBOSE_FAILURE} || $ENV{VF};
-    $ENV{HARNESS_VERBOSE_FAILURES_ONLY} = "yes"
-        if $ENV{VERBOSE_FAILURES_ONLY} || $ENV{VFO};
-    $ENV{HARNESS_VERBOSE_FAILURES_PROGRESS} = "yes"
-        if $ENV{VERBOSE_FAILURES_PROGRESS} || $ENV{VFP};
+    $ENV{HARNESS_VERBOSE_FAILURE_PROGRESS} = "yes"
+        if $ENV{VERBOSE_FAILURE_PROGRESS} || $ENV{VFP};
 }
 
 use File::Spec::Functions qw/catdir catfile curdir abs2rel rel2abs/;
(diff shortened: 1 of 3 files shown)


DDvO:
Are there any suggestions for potentially making the names
`VFO` (`HARNESS_VERBOSE_FAILURES_ONLY`) and
`VFP` (`HARNESS_VERBOSE_FAILURES_PROGRESS`)
less cryptic (i.e., more intuitive)?

DDvO:
How about using in Travis builds by default not `VF=1` but `VFO=1`,
such that in case a test fails specifically only the output of those test cases that failed is printed?

This would save us from manually searching through potentially huge amounts of test output for sections ending with `not ok`.

romen:
> Are there any suggestions for potentially making the names
> `VFO` (`HARNESS_VERBOSE_FAILURES_ONLY`) and
> `VFP` (`HARNESS_VERBOSE_FAILURES_PROGRESS`)
> less cryptic (i.e., more intuitive)?

I don't have a better suggestion, just wanted to note that `VFO` is particularly unfortunate as the uppercase `O` can be confused for a `0` depending on the font of the user!

romen:
> How about using in Travis builds by default not `VF=1` but `VFO=1`,
> such that in case a test fails specifically only the output of those test cases that failed is printed?
> 
> This would save us from manually searching through potentially huge amounts of test output for sections ending with `not ok`.

This I would really like, unless there is the risk of losing information relevant for debugging in not seeing the output of successes around the failure.
I personally can't think of such occurrences where this would be the case, but maybe @levitte can provide feedback on this!

test/run_tests.pl
```diff
@@ -177,6 +189,7 @@ sub find_matching_tests {
                         print $output_buffer if !$is_ok;
                         print "\n".$self->as_string
                             if !$is_ok || $failure_verbosity == 2;
+                        print "\n--------------------------------------------------------------------------------" if !$is_ok;
```
levitte, line 192: The basis for the perl framework is [TAP](http://testanything.org/) (Test Anything Protocol), and while not *currently* mandatory, it's encouraged to display everything that's not strictly part of the protocol as comments (i.e. lines starting with `#`).  We haven't been absolutely strict ourselves (it's been a learning process), but are getting better...  this lines goes in the other direction.

DDvO: Good point - no problem, now the separator starts with `# `.

levitte:
I would take it one step further and have VF mean VFO. That was my intention with VF from the get go, I just didn't dig deep enough to make it that finely tuned.

DDvO:
> I would take it one step further and have VF mean VFO. That was my intention with VF from the get go, I just didn't dig deep enough to make it that finely tuned.

Excellent - done.
This way the improved semantics of (hitherto) VFO automatically becomes the default for Travis builds,
and cryptic and error-prone name `VFO` is gone.

test/run_tests.pl
```diff
@@ -150,7 +153,7 @@ sub find_matching_tests {
                     if defined $fh;
 
                 my $failure_verbosity = $openssl_args{failure_verbosity};
-                if ($failure_verbosity == 3) {
+                if ($failure_verbosity == 3) { # old VF semantics
                     push @failure_output, $self->as_string;
                 } elsif ($failure_verbosity > 0) {
```
levitte, line 158: Nah, dump the old VF semantics...

``` diff
-                if ($failure_verbosity == 3) {
-                    push @failure_output, $self->as_string;
-                } elsif ($failure_verbosity > 0) {
+                if ($failure_verbosity > 0) {

Comment (DDvO) [reply test/run_tests.pl]: Done.

**Comment by richsalz** [comment]
How about "REPORT" instead of "HARNESS_VERBOSE_FAILURES" ?

**Comment by DDvO** [comment]
> How about "REPORT" instead of "HARNESS_VERBOSE_FAILURES" ?

This would be nice and short, but at least I would not understand from the name what it means.
How about `REPORT_FAILURES` or just  `FAILURES`?

I've tentatively added the latter as synonym to `VERBOSE_FAILURE` aka `VF`.

**Comment by richsalz** [comment]
When would REPORT be used?  When doing "make test"  context makes it clear to me, but YMMV.

**Comment by DDvO** [comment]
> When would REPORT be used? When doing "make test" context makes it clear to me, but YMMV.

Sure we are in the context of running tests, but 'report' could also mean getting statistics or other info.
In my view the name should make clear that it's about showing/reporting *failures*.

**Comment by richsalz** [comment]
I meant "REPORT_FAILURES"  That is ```s/HARNESS_VERBOSE/REPORT/```  Apologies for not being clear and then compounding my error :)

**Comment by DDvO** [comment]
Ah, @richsalz then I think we are on the same page :-)
So I've changed the tentative alias of `HARNESS_VERBOSE_FAILURE` (aka `VERBOSE_FAILURE` aka `VF`)
from `FAILURES` to `REPORT_FAILURES`.

**Comment by DDvO** [comment]
@levitte, are you fine with the alias as well, 
and if so, can you please re-approve?

**Comment by DDvO** [comment]
Rebased in order to handle merge conflicts.

File: test/run_tests.pl [hunk 3]
```diff
@@ -9,15 +9,15 @@
...
     $ENV{HARNESS_VERBOSE} = "yes" if $ENV{VERBOSE} || $ENV{V};
-    $ENV{HARNESS_VERBOSE_FAILURE} = "yes" if $ENV{VERBOSE_FAILURE} || $ENV{VF};
-    $ENV{HARNESS_VERBOSE_FAILURES_ONLY} = "yes"
-        if $ENV{VERBOSE_FAILURES_ONLY} || $ENV{VFO};
-    $ENV{HARNESS_VERBOSE_FAILURES_PROGRESS} = "yes"
-        if $ENV{VERBOSE_FAILURES_PROGRESS} || $ENV{VFP};
+    $ENV{HARNESS_VERBOSE_FAILURE} = "yes"
+        if $ENV{VERBOSE_FAILURE} || $ENV{VF} || $ENV{REPORT_FAILURES};
+    $ENV{HARNESS_VERBOSE_FAILURE_PROGRESS} = "yes"
+        if $ENV{VERBOSE_FAILURE_PROGRESS} || $ENV{VFP};
```
levitte, line 20: Should there be a `REPORT_FAILURE_PROGRESS` as well?

DDvO: Good point to be consistent here. 
Added `VERBOSE_FAILURES_PROGRESS`.

DDvO:
@levitte, could you please confirm your approval?
The only change relative to your earlier one is the two new alias names.

DDvO:
Since I did a minor extension of this PR (adding two env. var. aliases) after the approval by @levitte,
is a re-approval strictly needed in this case or can I go ahead and merge this?

I wanna push this ASAP because all Travis CI users will thus benefit from the
(then by default) restriction of verbose test output to those test cases that failed
(plus the general test info header and footer including some statistics as usual).

mattcaswell:
> Since I did a minor extension of this PR (adding two env. var. aliases) after the approval by @levitte,
is a re-approval strictly needed in this case or can I go ahead and merge this?

Unfortunately, yes - a re-approval is required.

DDvO:
> > Since I did a minor extension of this PR (adding two env. var. aliases) after the approval by @levitte,
> > is a re-approval strictly needed in this case or can I go ahead and merge this?
> 
> Unfortunately, yes - a re-approval is required.

I see, thanks.

So @levitte, could you please re-approve soon?
Sorry for being (for the reason given above) that pushy :wink: on being allowed to push this.

DDvO:
Merged - thanks!

PR #12280, opened 2020-06-26
Improve test/recipes/81-test_cmp_cli.t

DDvO:
* Correct Perl subroutine `quote_spc_empty` and its use
* Streamline `{start,stop}_mock_server` and improve server port setting

### DIFF
--- a/test/recipes/81-test_cmp_cli.t
+++ b/test/recipes/81-test_cmp_cli.t
@@ -55,7 +55,6 @@ my @cmp_basic_tests = (
 my $rsp_cert = "signer_only.crt";
 my $outfile = "test.cert.pem";
 my $secret = "pass:test";
-my $localport = 1700;
 
 # this uses the mock server directly in the cmp app, without TCP
 sub use_mock_srv_internally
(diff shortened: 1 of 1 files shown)


DDvO:
Can on OTC member please have a look at this little PR?

test/recipes/81-test_cmp_cli.t
```diff
@@ -222,7 +221,7 @@ indir data_dir() => sub {
                 test_cmp_cli_aspect($server_name, $aspect, $tests);
             };
         };
-        stop_mock_server() if $launch_mock;
+        stop_mock_server($pid) if $pid;
```
mattcaswell, line 224: Isn't the value of `$pid` undefined if `$server_name` isn't "Mock"?

DDvO: > Isn't the value of `$pid` undefined if `$server_name` isn't "Mock"?

Yes (because it is a scalar variable that has not been initialized in this case), 
and then it is considered 'false' in the `if` context - 
I've just confirmed the latter for myself at <https://perlmaven.com/boolean-values-in-perl>.

mattcaswell: Ah - ok fair enough.

DDvO:
Merged - thanks @mattcaswell

Issue #12276, opened 2020-06-25
alpha3: Testsuite fails on hurd due to truncated error code


sebastianas:
The test 04-test_err.t fails on hurd:

> 04-test_err.t ...................... 
>  The results of this test will end up in test-runs/test_err
> 1..1
>     # Subtest: ../../test/errtest
>     1..4
>     ok 1 - preserves_system_error
>     ok 2 - vdata_appends
>     ok 3 - raised_error
>     # ERROR: (string) 'expected == p' failed @ ../test/errtest.c:56
>     # --- expected
>     # +++ p
>     #    0:- ':error::system library:test_print_error_format:Operation not per'
>     #    0:+ ':error::system library:test_print_error_format:(os/kern) invalid'
>     #                                                        ^^^^^^^^^ ^^^^^^^
>     #   64:- 'mitted:errtest.c:30:'
>     #   64:+ ' address:errtest.c:3'
>     #         ^^^^^^^^^^^^^^^^^^^^
>     # 
 >    not ok 4 - test_print_error_format
 >../../util/wrap.pl ../../test/errtest => 1
 >not ok 1 - running errtest

errtest.c does:
>     ERR_PUT_error(ERR_LIB_SYS, 0, 1, "errtest.c", 30);

where 1 should be EPERM. And EPERM on hurd is 
> bigeasy@exodar:~$ grep EPERM /usr/include/i386-gnu/bits/errno.h
>  EPERM                          = 0x40000001,  /* Operation not permitted */
> #define EPERM                          0x40000001

So it is an int and it does not fit in 24bit which I believe is the what the code relies on judging by looking at ERR_PACK() and ERR_GET_REASON(). 
The errno number one is EKERN_INVALID_ADDRESS which explains the actual printed error message.

I don't think that this macros / kind of error handling is something new. I noticed that this test is behind 
OPENSSL_NO_DEPRECATED_3_0. Is this "reducing errno to 24bits" deprecated or just this set of functions?
Any idea what we can do about it?

levitte:
Duplicate of #12217

levitte:
This was reported once already, but does fill in details I asked about there.  Thank you.

> I don't think that this macros / kind of error handling is something new. I noticed that this test is behind
OPENSSL_NO_DEPRECATED_3_0. Is this "reducing errno to 24bits" deprecated or just this set of functions?
> Any idea what we can do about it?

[...]

PR #9452, opened 2019-07-24
Implement ERR_raise() / ERR_raise_data()

levitte:
`ERR_raise()` and `ERR_raise_data` is new functionality to raise errors, with this intended signature:

``` C
void ERR_raise(int lib, int reason);
void ERR_raise_data(int lib, int reason, const char *fmt, ...);
```

The intention with this is to have functionality that makes it easier to report errors with additional text, instead of the sometimes cumbersome error handling with `ERR_set_error_data()`.

`ERR_raise()` and `ERR_raise_data` are implemented as macros which works with a few building blocks:

- `ERR_new()` which simply allocates a new error record in the error queue.
- `ERR_set_debug()` which sets debugging information in the current error record.  "debugging information" are `__FILE__`, `__LINE__` and `__func__` or variants thereof (in reality, we use `OPENSSL_FILE`, `OPENSSL_LINE` and `OPENSSL_FUNC`, which are defined appropriately for the toolchain that builds this)
- `ERR_set_error()` which sets the basic error information in the current error record; library number, reason code, and a format string followed by arguments that form the additional data.

All the `FOOerr` macros are reimplemented in terms of `ERR_raise()` / `ERR_raise_data`, including `FUNCerr`, which also means that `ERR_put_func_error()` lost its meaning.

-----

This functionality was briefly discussed in [this thread on openssl-project](https://mta.openssl.org/pipermail/openssl-project/2019-June/001426.html)

PR #9441, opened 2019-07-23
Add script to convert XXXerr() to XXXerror

richsalz:
As discussed in the comment thread at https://github.com/openssl/openssl/pull/9391#discussion_r305412893, this adds new XXXerror() defines that gets rid of the now-unused function name and deprecates the XXXerr() defines.

This is FYI to give a "taste" of what the conversion would be like. In numeric order, the ones remaining are shown below.

Does the OMC want to do this?  If so, I'll continue with the conversion.
```
14 PROPerr(
16 OBJerr(
18 ASYNCerr(
34 PROVerr(
36 UIerr(
41 OCSPerr(
47 FUNCerr(
49 CONFerr(
50 CRYPTOerr(
50 DSAerr(
51 RANDerr(
57 CTerr(
58 CRMFerr(
60 DHerr(
63 OSSL_STOREerr(
63 PKCS12err(
64 KDFerr(
73 BNerr(
73 SM2err(
77 TSerr(
86 ENGINEerr(
86 PEMerr(
99 DSOerr(
108 BIOerr(
118 PKCS7err(
151 X509err(
184 CMSerr(
185 RSAerr(
195 X509V3err(
283 EVPerr(
295 ASN1err(
323 SSLerr(
470 ECerr(
```

PR #12289, opened 2020-06-26
Completely remove "func" from ERR state.

### LINKED ISSUE #12217: hurd failure
On hurd we're getting the following error:
```
04-test_err.t ...................... 
# The results of this test will end up in test-runs/test_err
1..1
    # Subtest: ../../test/errtest
    1..4
    ok 1 - preserves_system_error
    ok 2 - vdata_appends
    ok 3 - raised_error
    # ERROR: (string) 'expected == p' failed @ ../test/errtest.c:56
    # --- expected
    # +++ p
    #    0:- ':error::system library:test_print_error_format:Operation not per'
    #    0:+ ':error::system library:test_print_error_format:(os/kern) invalid'
    #                                                        ^^^^^^^^^ ^^^^^^^
    #   64:- 'mitted:errtest.c:30:'
    #   64:+ ' address:errtest.c:3'
    #         ^^^^^^^^^^^^^^^^^^^^
    # 
    not ok 4 - test_print_error_format
../../util/wrap.pl ../../test/errtest => 1
not ok 1 - running errtest
```

richsalz:
Use EPERM not "1" in errtest.
Add comments on error code format, update documentation.
Clean up some code in err library.

This started because I gave an obscure comment https://github.com/openssl/openssl/issues/12276#issuecomment-650166747 and I was challenged to make a PR.  So here it is.  I think this fixes #12217, but it does it by fully removing
the function name (which we removed internally anyway).

### DIFF
--- a/include/openssl/err.h
+++ b/include/openssl/err.h
@@ -168,13 +168,14 @@ struct err_state_st {
  * library (L), function (F, not used as of 3.0), and reason (R). It is
  * split as follows:
  *      LL FFF RRR
- * The top two bits of LL are zero and fatal-error-flag.
- * This is part of the ABI, and therefore an implied part of the API.
- * Since the function is not used, a future release might re-partition the code.
+ * This is an implied part of the API. Since the function sub-part is not
+ * used, a future release might re-partition the code.  The first two bigs
+ * of the first nibble of RRR are reserved for ERR_R_FATAL and another
+ * flag that is currently zero.
  */
-# define ERR_MAX_LIB             0x3F
+# define ERR_MAX_LIB             0xFF
 # define ERR_MAX_FUNC            0
-# define ERR_MAX_REASON          0xFFF
+# define ERR_MAX_REASON          0x3FF
 # define ERR_R_FATAL             0x40
 
 # define ERR_PACK(l,f,r) ( \
--- a/include/openssl/err.h
+++ b/include/openssl/err.h
@@ -163,12 +163,25 @@ struct err_state_st {
 #  define X509err(f, r) ERR_raise_data(ERR_LIB_X509, (r), NULL)
 # endif
 
+/*
+ * Error codes are packed into a single 32-bit number, and have three parts,
+ * library (L), function (F, not used as of 3.0), and reason (R). It is
+ * split as follows:
+ *      LL FFF RRR
+ * The top two bits of LL are zero and fatal-error-flag.
+ * This is part of the ABI, and therefore an implied part of the API.
+ * Since the function is not used, a future release might re-partition the code.
+ */
+# define ERR_MAX_LIB             0x3F
+# define ERR_MAX_FUNC            0
+# define ERR_MAX_REASON          0xFFF
+# define ERR_R_FATAL             0x40
+
 # define ERR_PACK(l,f,r) ( \
         (((unsigned int)(l) & 0x0FF) << 24L) | \
-        (((unsigned int)(f) & 0xFFF) << 12L) | \
         (((unsigned int)(r) & 0xFFF)       ) )
 # define ERR_GET_LIB(l)          (int)(((l) >> 24L) & 0x0FFL)
-# define ERR_GET_FUNC(l)         (int)(((l) >> 12L) & 0xFFFL)
+# define ERR_GET_FUNC(l)         0
 # define ERR_GET_REASON(l)       (int)( (l)         & 0xFFFL)
 # define ERR_FATAL_ERROR(l)      (int)( (l)         & ERR_R_FATAL)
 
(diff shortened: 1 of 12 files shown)


richsalz:
This also uses the symbolic name EPERM instead of "1" and skips the test if the value of EPERM is too big.

include/openssl/err.h
```diff
@@ -284,7 +296,6 @@ unsigned long ERR_get_error(void);
  * additional data.
  */
 unsigned long ERR_get_error_line(const char **file, int *line);
-unsigned long ERR_get_error_func(const char **func);
```
levitte, line 287: You must *replace* it, not just remove it, since it's a public function.  A macro would work, but lacks the deprecation attribute, so it would be preferable if the functions were properly deprecated, and you left an implementation that just returns zero.

levitte: Oh hold on, they're *new* in master...  Never mind then!

include/openssl/err.h
```diff
@@ -163,12 +163,25 @@ struct err_state_st {
 #  define X509err(f, r) ERR_raise_data(ERR_LIB_X509, (r), NULL)
 # endif
 
+/*
+ * Error codes are packed into a single 32-bit number, and have three parts,
+ * library (L), function (F, not used as of 3.0), and reason (R). It is
+ * split as follows:
+ *      LL FFF RRR
+ * The top two bits of LL are zero and fatal-error-flag.
```
levitte, line 171: bits => nibbles

levitte: The fatal-error-flag is part of the RRR (somewhere in the middle)...  and is never used.  It actually looks like an unfinished idea that just hung around for eternity.

include/openssl/err.h
```diff
@@ -163,12 +163,25 @@ struct err_state_st {
...
+ * This is part of the ABI, and therefore an implied part of the API.
```
levitte, line 172: Actually, since 3.0 only needs to be source compatible, you don't need to worry about the ABI

levitte: ... in other word, it should be safe to simply ditch the function number altogether, and thereby make the packed error code change to this structure:

    LL RRRRRR

include/openssl/err.h
```diff
@@ -225,7 +238,6 @@ struct err_state_st {
 # define ERR_R_MISSING_ASN1_EOS                  63
 
 /* fatal error */
-# define ERR_R_FATAL                             64
```
levitte, line 228: This was wrongly moved

include/openssl/err.h
richsalz: You're right, 0x40, not shifted.  So reason codes are limited to 0x3FF.  I rewrote the comment and am pushing a fixup commit.  Should I remove/deprecate ```ERR_FLAG_FATAL``` or just leave it?

include/openssl/err.h
richsalz: Ditching the function number was more than I first wanted to do, but will do so if you think I should.  I would save it as ```FLLRRRRR``` where is a flags nibble, currently must be zero.

include/openssl/err.h
richsalz: I disagree, because it's part of the composition of an error code, so it belongs with the other func/lib/reason defines.

richsalz:
I don't think function strings should remain.  We don't use them, and we don't have much evidence that people do anything with OpenSSL's error code stuff in their own code.

include/openssl/err.h
levitte: It's part of the composition of *reason codes* (hence `_R_`)

levitte:
> I don't think function strings should remain. We don't use them,

Er, we don't? So if the error output shows you a function name, that's not useful? I wholeheartedly disagree.
We added this so we wouldn't have to keep using `_F_` macros and their associated strings, and switched to use `_func_` and similar instead. Have you forgotten?

richsalz:
>>We don't use function names

>Er, we don't?

No, we don't.   We made a conscious decision to stop putting internal OpenSSL fuction names.  The func-name stuff isn't used by openssl, and won't be because of that decision.  It *might possibly* by useful, but there are other ways -- error-data -- to get the same thing.

include/openssl/err.h
richsalz: It is also a part of the partitioning of the error-code space, so it should be with the other shifts and masks that define the partitioning.  It's a matter of taste, I guess.

levitte:
@richsalz, look back at #9452.  That's the main PR that *replaced* the old function indicators (`_F_` macros with associated strings) with the new (based on `OPENSSL_FUNC`).  We didn't decide to remove function indicators altogether, we only decided that we want to get rid of the old.

Now, for some reason, *you* want to get rid of function indicators altogether...  but that doesn't mean that *"we"* want that.

richsalz:
I'll see your #9452, and raise you #9441 :)

> This is what you want .. not what *we* want.

Aren't many PR's what the poster wants? Shrug.  

If you don't like this, just take the one small fix to test/errtest.c

levitte:
#9441 doesn't in any way contradict what I've said here.

richsalz:
And I don't see how #9452 adds to your argument, but that's okay.  You disagree with what I'm doing here, and that's okay.  You're a committer, I'm not, and I'm okay with that.  Doesn't mean that I'm wrong. :)

Almost every time OpenSSL adds something because "someone might need it," as opposed to doing something that OpenSSL *itself* needs it, the project gets it wrong.  [YAGNI](https://www.martinfowler.com/bliki/Yagni.html), y'all.

richsalz:

PR #12292, opened 2020-06-26
Fix OSSL_PROVIDER_get_capabilities()

### LINKED ISSUE #12286: SSL_CTX_new(meth) returning NULL with legacy provider loaded
In OpenSSL 3.0.0-alpha4 the following code now fails with a null `SSL_CTX *`:

```c
#include <openssl/ssl.h>
#include <openssl/provider.h>
#include <assert.h>

int main(void) {
    SSL_library_init();
    SSL_load_error_strings();
    OSSL_PROVIDER_load(NULL, "legacy");
    OSSL_PROVIDER_load(NULL, "default");
    const SSL_METHOD *meth = TLS_method();
    assert(meth != NULL);
    SSL_CTX *ctx = SSL_CTX_new(meth);
    assert(ctx != NULL);
    return 0;
}
```

In alpha3 and below this worked without error. Removing the `OSSL_PROVIDER_load` for the legacy provider fixes the error (but obviously then we can't use legacy algorithms).

mattcaswell:
It is not a failure to call OSSL_PROVIDER_get_capabilities() with a
provider loaded that has no capabilities.

Fixes #12286

While I was at it I fixed a capabilities related typo in the man pages pointed out by @romen. Also a minor logic flow problem in an error path.

### DIFF
--- a/crypto/provider_core.c
+++ b/crypto/provider_core.c
@@ -831,7 +831,7 @@ int ossl_provider_get_capabilities(const OSSL_PROVIDER *prov,
                                    void *arg)
 {
     return prov->get_capabilities == NULL
-        ? 0 : prov->get_capabilities(prov->provctx, capability, cb, arg);
+        ? 1 : prov->get_capabilities(prov->provctx, capability, cb, arg);
 }
 
 
--- a/providers/common/capabilities.c
+++ b/providers/common/capabilities.c
@@ -97,26 +97,34 @@ static const TLS_GROUP_CONSTANTS group_list[35] = {
 
 static const OSSL_PARAM param_group_list[][10] = {
 #ifndef OPENSSL_NO_EC
+# ifndef OPENSSL_NO_EC2M
     TLS_GROUP_ENTRY("sect163k1", "sect163k1", "EC", 0),
+# endif
 # ifndef FIPS_MODULE
     TLS_GROUP_ENTRY("sect163r1", "sect163r1", "EC", 1),
 # endif
+# ifndef OPENSSL_NO_EC2M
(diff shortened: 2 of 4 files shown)


mattcaswell:
Hmm. Travis failure appears relevant.

levitte:
> Travis failure appears relevant.

I would even go so far that it's hard to find a more relevant failure :laughing:

mattcaswell:
> Hmm. Travis failure appears relevant.

I think this could (maybe) be another instance of #12293. At least I'm seeing some unexpected KEYMGMT fetch behaviour.

slontis:
For the failing test I can see that it tries to do a EVP_KEYMGMT_fetch() inside add_provider_groups in order to add the "XORgroup".

This fails when it tries to process the query for the legacy provider..

```
provider_forall_loaded()
   for (i = 0; i < num_provs; i++) { 
      ...    
      if (!(ret = cb(prov, cbdata)))  / * legacy provider will return a failure as it does not handle the operation */
          break;
  } 
}
```

slontis:
For some reason this function is behaving different between the working config and the -no config:

inner_evp_generic_fetch() {
 ...

/*
  In the non working case name_id returns 0 when name is "XOR" which means it goes down a different path,
after the call.. In the working case the above code doesnt get activated as the method_id is already available.
*/
name_id = ossl_namemap_name2num(namemap, name);
....
}

mattcaswell:
I've rebased this now that the fix for #12293 has gone in. Lets see if that improves things.

mattcaswell:
The travis errors no longer look relevant to this PR.

mattcaswell:
Pushed. Thanks.

PR #12295, opened 2020-06-27
apps/openssl: clean-up of unused fallback code

levitte:
Remove code in help_main() that duplicates the case when 'openssl' is
called with no arguments, which is now handled in main().


levitte:
Merged

96e0445195e22f27c661ca4cd57c1caa52b6abb4 apps/openssl: clean-up of unused fallback code

PR #12296, opened 2020-06-27
Improve error diagnostics of X509V3 conf parsing, CMP, and other apps

### LINKED ISSUE #12268: Please fix/correct the text of the error message
https://github.com/openssl/openssl/blob/19765f5bcfcb6dc181c3a562c32ddd489291502a/apps/cmp.c#L1886

I assume it means to say "could **not**".

DDvO:
This fixes #12268
and various other small issues I recently found using the CMP CLI and other apps:

* Error messages on parsing X.509v3 extension specs in config files where often weird. For instance for
```
extendedKeyUsage = clientAuth, serverAuth, wrong
```
the diagnostics was:
```
asn1 encoding routines:crypto/asn1/a_object.c:73:CMP error: first num too large
X509 V3 routines:crypto/x509/v3_extku.c:98:CMP error: invalid object identifier : section:<NULL>,name:unknown_key_usage_name,value:<NULL>
X509 V3 routines:crypto/x509/v3_conf.c:50:CMP error: error in extension : name=extendedKeyUsage, value=clientAuth, serverAuth, unknown_key_usage_name
```
With the fixes done here, this becomes:
```
object identifier routines:crypto/objects/obj_dat.c:376:CMP error: unknown object name
X509 V3 routines:crypto/x509/v3_extku.c:98:CMP error: invalid object identifier : unknown_key_usage_name
X509 V3 routines:crypto/x509/v3_conf.c:50:CMP error: error in extension : section=myKeyUsages, name=extendedKeyUsage, value=clientAuth, serverAuth, unknown_key_usage_name
```

* Correct and extend the diagnostics of `parse_name()` in `apps/lib/apps.c`
* Defer the so far premature diagnostic output of `apps/cmp.c` on server+proxy to be contacted.


apps/lib/apps.c
```diff
@@ -1644,8 +1648,8 @@ X509_NAME *parse_name(const char *cp, long chtype, int canmulti)
             *bp++ = *cp++;
         if (*cp == '\0') {
             BIO_printf(bio_err,
-                       "%s: Hit end of string before finding the '='\n",
-                       opt_getprog());
+                       "%s: Hit end of %d name string before finding the '='\n",
```
paulidale, line 1651: There ought to be a better word that "Hit".  What about "Reached"?

crypto/objects/obj_dat.c
```diff
@@ -372,6 +372,10 @@ ASN1_OBJECT *OBJ_txt2obj(const char *s, int no_name)
         if (((nid = OBJ_sn2nid(s)) != NID_undef) ||
             ((nid = OBJ_ln2nid(s)) != NID_undef))
             return OBJ_nid2obj(nid);
+        if (*s < '0' || *s > '9') {
```
paulidale, line 375: Might `if (!ossl_isdigit(*s))` be safer since it is character set aware?

crypto/x509/v3_cpols.c
```diff
@@ -116,11 +117,10 @@ static STACK_OF(POLICYINFO) *r2i_certpol(X509V3_EXT_METHOD *method,
     ia5org = 0;
     for (i = 0; i < num; i++) {
         cnf = sk_CONF_VALUE_value(vals, i);
-
-        if (cnf->value || !cnf->name) {
+        if (cnf->value != NULL || cnf->name == NULL) {
             X509V3err(X509V3_F_R2I_CERTPOL,
                       X509V3_R_INVALID_POLICY_IDENTIFIER);
-            X509V3_conf_err(cnf);
+            X509V3_conf_err2(cnf);
```
paulidale, line 123: Is there a better name than `X509V3_conf_err2`?

include/openssl/x509v3.h
```diff
@@ -331,10 +331,6 @@ struct ISSUING_DIST_POINT_st {
 /* onlysomereasons present */
 # define IDP_REASONS     0x40
 
-# define X509V3_conf_err(val) ERR_add_error_data(6, \
```
paulidale, line 334: This is a breaking API change.

apps/lib/apps.c
DDvO: Yes, in this case we can simply (and more concisely) replace "Hit end of" by "End of" - done that.

crypto/objects/obj_dat.c
DDvO: Good point - done.

crypto/x509/v3_cpols.c
DDvO: How about `X509v3_conf_add_error_name_value`?

include/openssl/x509v3.h
DDvO: Well, that macro is exported but undocumented.
Shall I move its definition back to `include/openssl/x509v3.h`?

crypto/x509/v3_cpols.c
DDvO: I just did this tentatively.

include/openssl/x509v3.h
DDvO: Just did this tentatively as well.

apps/lib/apps.c
```diff
@@ -1648,7 +1648,7 @@ X509_NAME *parse_name(const char *cp, int chtype, int canmulti,
             *bp++ = *cp++;
         if (*cp == '\0') {
             BIO_printf(bio_err,
-                       "%s: Hit end of %d name string before finding the '='\n",
+                       "%s: End of %d name string before finding the '='\n",
```
levitte, line 1651: `"%s: %d name string doesn't contain a '='\n"`

DDvO: > `"%s: %d name string doesn't contain a '='\n"`

This is not general enough because there may be occurrences of `=` (for other RDNs) before."
Just improved this to:
```
            BIO_printf(bio_err,
                       "%s: Missing '=' after RDN type string '%s' in %s name string\n",
                       opt_getprog(), typestr, desc);

```
(which required terminating the `typestr` just before that).

levitte: Even better!

DDvO:
Thanks @paulidale for reviewing this.
Dues this look okay now?

DDvO:
For whatever reason, the Travis builds have been queued, for whatever strange reason, for two days.
I've just rebased, squashed, and force-pushed on this PR; hope Travis will do better this time.

DDvO:
@paulidale, can you still have a quick look at the changes you requested and my reaction and approve that before you leave?

In any case, have a nice and restful vacation!

DDvO:
Merged - thanks @paulidale

PR #12300, opened 2020-06-27
Add X509_self_signed(), splitting doc/man3/X509_sign.pod

DDvO:
This off-loads some topics that can be handled independently of PR #10587:
* Move doc of `X509{,_REQ,_CRL}_verify{,_ex}()` from `X509_sign.pod` to new file `X509_verify.pod`
* Add `X509_self_signed()`, extending and improving documentation and tests and using it in `cmp_util.c`


DDvO:
It looks like that this PR won't be needed because the full PR #10587 will be approved soon...

PR #12305, opened 2020-06-28
[apps/genpkey] exit status should not be 0 on output errors

romen:
## TL;DR

If the key is to be serialized or printed as text and the framework returns an error, the app should signal the failure to the user using a non-zero exit status.

## History

From `1.1.1-stable` (this snippet did not change in `master`, only shifted by a few lines):
https://github.com/openssl/openssl/blob/7bdf1ee8ccb69a743e29e3d1a72194c30e8583ae/apps/genpkey.c#L180-L199

The code above has been unchanged (apart from formatting) for the last 14 years.

In case of an error writing the key to a file or in textual form to the terminal, the exit status of the command is still reset to 0.

## Branches

This should be fixed both in `master` and in `1.1.1`

### DIFF
--- a/apps/genpkey.c
+++ b/apps/genpkey.c
@@ -192,6 +192,7 @@ int genpkey_main(int argc, char **argv)
     if (rv <= 0) {
         BIO_puts(bio_err, "Error writing key\n");
         ERR_print_errors(bio_err);
+        goto end;
     }
 
     if (text) {
@@ -203,6 +204,7 @@ int genpkey_main(int argc, char **argv)
         if (rv <= 0) {
             BIO_puts(bio_err, "Error printing key\n");
             ERR_print_errors(bio_err);
+            goto end;
         }
     }
 
--- a/apps/genpkey.c
+++ b/apps/genpkey.c
@@ -189,10 +189,12 @@ int genpkey_main(int argc, char **argv)
         goto end;
     }
 
+    ret = 0;
+
     if (rv <= 0) {
         BIO_puts(bio_err, "Error writing key\n");
         ERR_print_errors(bio_err);
-        goto end;
+        ret++;
     }
 
     if (text) {
@@ -204,11 +206,9 @@ int genpkey_main(int argc, char **argv)
         if (rv <= 0) {
             BIO_puts(bio_err, "Error printing key\n");
             ERR_print_errors(bio_err);
-            goto end;
+            ret++;
         }
     }
-
-    ret = 0;
 
  end:
     EVP_PKEY_free(pkey);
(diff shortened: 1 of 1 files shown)


romen:
@mattcaswell this command has been "misbehaving" this way for a very long time and across releases.

Yet I don't think these 2 error paths have been triggered often, as excluding the current status of `master` with the serializer implementation still being finished, my guts say errors that could have resulted in either of these 2 steps failing would have been caught within a `goto end` block, before reaching this part of `genpkey_main`.

Anyway, one could argue that fixing this bug is changing an established behavior of `genpkey`:
- do we require a discussion and vote to change this in `1.1.1`?
- do we require a discussion and vote to change this in `master`?

levitte:
I would rather move `ret = 0` to line 192 and replace your `goto end` with `ret++`

romen:
> I would rather move `ret = 0` to line 192 and replace your `goto end` with `ret++`

I like it! I actually fixed it like that first, but later changed to use `goto` just for conformity with our usual handling and to shorten the discussion!

Fixup pushed!

romen:
Travis failure is due to #12306 (the failure was already there, both in 1.1.1 and master, this PR just bring it to the surface!)

richsalz:
```ret++```  Why not ```ret = 1```?

romen:
> `ret++` Why not `ret = 1`?

@richsalz either works for me, I guess both 1 and 2 are valid "failure" exit statuses, but I agree that it does not add much more information than is already reported on the screen to know which of the two is failing.

richsalz:
I only brought it up because there are *very* few instances where an exit code is meaningful.  (Verify, maybe?)

romen:
@paulidale Yes, the failures are due to #12306

I'll soon open PRs to fix it in both `master` and `1.1.1`, and I will update this PR and the `1.1.1` one to rebase on top of those and to further test out these problematic curves!

romen:
@paulidale 

Travis failures are expected: #12305, #12312 and #12307 depend on each other.

Check #12307 (based on top of the other 2) for (hopefully) green Travis.

romen:
CI failures are expected, and fixed in #12307 / #12308  which have been approved.

They will be merged together

romen:
Merged to `master` as:

- 466d30c0d7fa861a5fcbaebd2e2010a8c2aea322 `[apps/genpkey] exit status should not be 0 on output errors`

Merged to `1.1.1` as:

- 1940c092a52afd8bc919b8faa5f3d51004503f3a `[apps/genpkey] exit status should not be 0 on output errors`

Thanks!

Issue #12277, opened 2020-06-25
Coverity 1463570 Use after free


paulidale:
Use after free of an engine.

DDvO:
It this issue gone?
I cannot find it (any more) on >https://scan5.coverity.com/reports.htm#v53814/p10222>.

DDvO:
See also my comment on this issue at https://github.com/openssl/openssl/pull/12231#issuecomment-650032345.

paulidale:
I flagged it as a false positive.  It looked like Coverity got confused by the reference counting.  There was an up ref in the same block as the _first_ free -- this means it cannot have been freed there.  Look in `load_key` in apps.c: it calls `init_engine` which ups the count frees it at the end of the block which reduces the count -- the free cannot occur.

PR #12309, opened 2020-06-28
rand: Coverity fixes

paulidale:
Mostly fixing problems with the provider friendly rand work.

Also includes a fix for a null pointer dereference in properties and the removal of an unnecessary NULL check in apps/cmp.c

### DIFF
--- a/apps/cmp.c
+++ b/apps/cmp.c
@@ -3186,8 +3186,7 @@ int cmp_main(int argc, char **argv)
     X509_STORE_free(OSSL_CMP_CTX_get_certConf_cb_arg(cmp_ctx));
     OSSL_CMP_CTX_free(cmp_ctx);
     X509_VERIFY_PARAM_free(vpm);
-    if (engine != NULL) /* workaround for Coverity false positive */
-        release_engine(engine);
+    release_engine(engine);
 
     NCONF_free(conf); /* must not do as long as opt_... variables are used */
     OSSL_CMP_log_close();
--- a/apps/lib/apps.c
+++ b/apps/lib/apps.c
@@ -1156,9 +1156,8 @@ ENGINE *setup_engine_methods(const char *id, unsigned int methods, int debug)
 void release_engine(ENGINE *e)
 {
 #ifndef OPENSSL_NO_ENGINE
-    if (e != NULL)
-        /* Free our "structural" reference. */
-        ENGINE_free(e);
+    /* Free our "structural" reference. */
+    ENGINE_free(e);
 #endif
 }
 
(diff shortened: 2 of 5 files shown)


paulidale:
Ping for review please.

apps/cmp.c
```diff
@@ -3186,8 +3186,7 @@ int cmp_main(int argc, char **argv)
     X509_STORE_free(OSSL_CMP_CTX_get_certConf_cb_arg(cmp_ctx));
     OSSL_CMP_CTX_free(cmp_ctx);
     X509_VERIFY_PARAM_free(vpm);
-    if (engine != NULL) /* workaround for Coverity false positive */
-        release_engine(engine);
+    release_engine(engine);
```
DDvO, line 3189: It's correct to remove this guard because issue #12277 needs to be solved differently,
as I note here: https://github.com/openssl/openssl/pull/12231#issuecomment-650032345

DDvO: BTW, is that issue #12277 maybe already solved?

At least it does not look like that this PR contains a fix for `OPENSSL_free(e)` being called prematurely in `cyrpto/engine/eng_lib.c`, and I cannot find any more the corresponding Coverity issue 1463570.

paulidale: It's fine to remove the guard since it is done one level down:
``` C
void release_engine(ENGINE *e)
{
#ifndef OPENSSL_NO_ENGINE
    if (e != NULL)
        /* Free our "structural" reference. */
        ENGINE_free(e);
#endif
}
```

paulidale: See #12277 for comment about why its a false positive.

slontis:
Looks ok to me.. Would approve if I could :)

apps/cmp.c
DDvO: > It's fine to remove the guard since it is done one level down:
> 
> ```c
> void release_engine(ENGINE *e)
> {
> #ifndef OPENSSL_NO_ENGINE
>     if (e != NULL)
>         /* Free our "structural" reference. */
>         ENGINE_free(e);
> #endif
> }
> ```

Actually also there the guard can be removed,
since for any (regular) `*_free()` functions calling with a `NULL` arg is a no-op.

DDvO:
> Looks ok to me.. Would approve if I could :)

As @mattcaswell taught me:
Even where a normal committer's approval does not formally count, it's a helpful indication to others.
So please feel encouraged to give approvals from your perspective whenever you convinced yourself that a PR is fine.

paulidale:
@slontis and I work for the same employer, we don't approve each other's PRs.  Doing so would risk merging happening.

apps/cmp.c
paulidale: Good catch.  Fixed now.

paulidale:
Merged to master, thanks for the feedback.

PR #12321, opened 2020-06-29
rand: fix recursive locking issue.

### LINKED ISSUE #12290: RAND_status() appears to be broken in OpenSSL 3.0.0-alpha
Greetings!  I'm trying to port Tor so that it will work with OpenSSL 3.0.0 when it's released.  Unfortunately, our code currently expects `RAND_status()` to return 1 when the prng is seeded, and that feature appears to be broken in the alphas.

Here is a short program to demonstrate:
```
#include <openssl/rand.h>
#include <stdio.h>

int main(int argc, char **argv)
{
	int r1 = RAND_poll();
	printf("RAND_poll() said %d\n", r1);
	unsigned char byte;
	int r2 = RAND_bytes(&byte, 1);
	printf("RAND_bytes() said %d\n", r2);
	int r3 = RAND_status();
	printf("RAND_status() said %d\n", r3);
	if (r3 == 1) {
		puts("RAND_status is okay!");
	} else {
		puts("RAND_status said we aren't seeded?");
	}
	return 0;
}
```

With OpenSSL 1.1.1g, the output is:
```
RAND_poll() said 1
RAND_bytes() said 1
RAND_status() said 1
RAND_status is okay!
```

But with OpenSSL 3.0.0-alpha4, the output is:
```
RAND_poll() said 1
RAND_bytes() said 1
RAND_status() said 0
RAND_status said we aren't seeded?
```

I believe that this behavior is an error: if the PRNG isn't seeded, then why did RAND_bytes() report success?

paulidale:
The calls to query the DRBG strength, state and maximum output size all used nested locks.  This removes the nesting.

Fixes #12290

### DIFF
--- a/crypto/evp/evp_rand.c
+++ b/crypto/evp/evp_rand.c
@@ -339,30 +339,43 @@ EVP_RAND *EVP_RAND_CTX_rand(EVP_RAND_CTX *ctx)
     return ctx->meth;
 }
 
+static int evp_rand_get_ctx_params_locked(EVP_RAND_CTX *ctx,
+                                          OSSL_PARAM params[])
+{
+    return ctx->meth->get_ctx_params(ctx->data, params);
+}
+
 int EVP_RAND_get_ctx_params(EVP_RAND_CTX *ctx, OSSL_PARAM params[])
 {
     int res;
 
     if (!evp_rand_lock(ctx))
         return 0;
-    res = ctx->meth->get_ctx_params(ctx->data, params);
+    res = evp_rand_get_ctx_params_locked(ctx, params);
     evp_rand_unlock(ctx);
     return res;
 }
 
+static int evp_rand_set_ctx_params_locked(EVP_RAND_CTX *ctx,
+                                          const OSSL_PARAM params[])
+{
+    /* Clear out the cache state because the values can change on a set */
+    ctx->strength = 0;
+    ctx->max_request = 0;
+
+    if (ctx->meth->set_ctx_params != NULL)
+        return ctx->meth->set_ctx_params(ctx->data, params);
+    return 1;
+}
+
 int EVP_RAND_set_ctx_params(EVP_RAND_CTX *ctx, const OSSL_PARAM params[])
 {
-    int res = 1;
+    int res;
 
-    if (ctx->meth->set_ctx_params != NULL) {
-        if (!evp_rand_lock(ctx))
-            return 0;
-        res = ctx->meth->set_ctx_params(ctx->data, params);
-        evp_rand_unlock(ctx);
-        /* Clear out the cache state because the values can change on a set */
-        ctx->strength = 0;
-        ctx->max_request = 0;
-    }
+    if (!evp_rand_lock(ctx))
+        return 0;
+    res = evp_rand_set_ctx_params_locked(ctx, params);
+    evp_rand_unlock(ctx);
     return res;
 }
 
@@ -380,7 +393,7 @@ const OSSL_PARAM *EVP_RAND_gettable_ctx_params(const EVP_RAND *rand)
 const OSSL_PARAM *EVP_RAND_settable_ctx_params(const EVP_RAND *rand)
 {
     return rand->settable_ctx_params == NULL ? NULL
-                                             :rand->settable_ctx_params();
+                                             : rand->settable_ctx_params();
 }
 
 void EVP_RAND_do_all_provided(OPENSSL_CTX *libctx,
@@ -400,6 +413,14 @@ void EVP_RAND_names_do_all(const EVP_RAND *rand,
         evp_names_do_all(rand->prov, rand->name_id, fn, data);
 }
 
+static int evp_rand_instantiate_locked
+    (EVP_RAND_CTX *ctx, unsigned int strength, int prediction_resistance,
+     const unsigned char *pstr, size_t pstr_len)
+{
+    return ctx->meth->instantiate(ctx->data, strength, prediction_resistance,
+                                  pstr, pstr_len);
+}
+
 int EVP_RAND_instantiate(EVP_RAND_CTX *ctx, unsigned int strength,
                          int prediction_resistance,
                          const unsigned char *pstr, size_t pstr_len)
@@ -408,10 +429,15 @@ int EVP_RAND_instantiate(EVP_RAND_CTX *ctx, unsigned int strength,
 
     if (!evp_rand_lock(ctx))
         return 0;
-    res = ctx->meth->instantiate(ctx->data, strength, prediction_resistance,
-                                 pstr, pstr_len);
+    res = evp_rand_instantiate_locked(ctx, strength, prediction_resistance,
+                                      pstr, pstr_len);
     evp_rand_unlock(ctx);
     return res;
+}
+
+static int evp_rand_uninstantiate_locked(EVP_RAND_CTX *ctx)
+{
+    return ctx->meth->uninstantiate(ctx->data);
 }
 
 int EVP_RAND_uninstantiate(EVP_RAND_CTX *ctx)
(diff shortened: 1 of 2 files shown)


crypto/evp/evp_rand.c
```diff
@@ -440,7 +439,8 @@ int EVP_RAND_generate(EVP_RAND_CTX *ctx, unsigned char *out, size_t outlen,
         params[0] = OSSL_PARAM_construct_size_t(OSSL_DRBG_PARAM_MAX_REQUEST,
                                                 &chunk);
         params[1] = OSSL_PARAM_construct_end();
-        if (!EVP_RAND_get_ctx_params(ctx, params) || chunk == 0) {
+        /* Cannot call EVP_RAND_get_ctx_params() since we have the lock */
+        if (!ctx->meth->get_ctx_params(ctx->data, params) || chunk == 0) {
```
slontis, line 443: err yuk :)

paulidale: Why?  It's just calling a method.

slontis: I dont think you have any other options here :)

mspncp: > I dont think you have any other options here :)

There is _always_ another option:) This change here fixes a single recursive lock at the prise of readability. The next recursive lock might be lurking elsewhere. IMHO there is a more generic and consistent solution to avoid taking recursive locks: identify _all_ EVP_RAND functions which take the lock and split them into a lowercase private (inline?) function which does the actual work

- `evp_rand_generate()`, `evp_rand_get_ctx_params()`, ...

and a public uppercase function which wraps the locking around the lowercase function.

- `EVP_RAND_generate()`, `EVP_RAND_get_ctx_params()`, ...

Then stick to rule to always call the lowercase functions internally. That' s how I did it in RAND_DRBG. It comes at the prise of an extra function call which in many cases the compiler can inline away.

mspncp: > identify all EVP_RAND functions

ok, not really all, but all funcitons which need to be called internally. It's not worth to split `EVP_RAND_generate()` if it does not get called internally anyway.

paulidale: If this solution is adopted, it should apply to *all* EVP_RAND functions, this should include `EVP_RAND_generate()`.

BTW: the mixed case functions tripped me up a couple of times.

This sounds like it should be an OTC coding style decision.

paulidale: All refactored.  I used lower case function names with a __locked_ suffix to better indicate what the functions do.

crypto/evp/evp_rand.c
```diff
@@ -338,32 +339,43 @@ EVP_RAND *EVP_RAND_CTX_rand(EVP_RAND_CTX *ctx)
...
+    evp_rand_unlock(ctx);
     return res;
 }
 
+static int evp_rand_set_ctx_params_locked(EVP_RAND_CTX *ctx,
+                                          const OSSL_PARAM params[])
+{
+    /* Clear out the cache state because the values can change on a set */
+    ctx->strength = 0;
+    ctx->max_request = 0;
```
mattcaswell, line 364: Should this only happen if `set_ctx_params` is non-NULL?

crypto/evp/evp_rand.c
```diff
@@ -194,6 +194,7 @@ static void *evp_rand_from_dispatch(int name_id,
             if (rand->get_ctx_params != NULL)
                 break;
             rand->get_ctx_params = OSSL_FUNC_rand_get_ctx_params(fns);
+            fnctxcnt++;
```
mspncp, line 197: Are you fixing an unrelated error here en passant?

crypto/evp/evp_rand.c
```diff
@@ -220,7 +221,7 @@ static void *evp_rand_from_dispatch(int name_id,
      * locking functions.
      */
     if (fnrandcnt != 3
-            || fnctxcnt != 2
+            || fnctxcnt != 3
```
mspncp, line 224: > Are you fixing an unrelated error here en passant?

same question here

mspncp:
>  The next recursive lock might be lurking elsewhere. 

Gotcha!

As I predicted, here's another one. And the capitalization (EVP_RAND_...() vs. evp_rand_..()) immediately identifies the culprit:

`#0`  CRYPTO_THREAD_write_lock (lock=0x555555581bc0) at crypto/threads_pthread.c:83
`#1`  0x00007ffff7eac3be in drbg_lock (vctx=0x55555557cd80) at providers/implementations/rands/drbg.c:51
`#2`  0x00007ffff7d5c9cf in evp_rand_lock (rand=0x55555557c4f0) at crypto/evp/evp_rand.c:78
`#3`  0x00007ffff7d5d37e in **EVP_RAND_get_ctx_params** (ctx=0x55555557c4f0, params=0x7fffffffdef0) at crypto/evp/evp_rand.c:352
`#4`  0x00007ffff7d5dc80 in evp_rand_state_locked (ctx=0x55555557c4f0) at crypto/evp/evp_rand.c:581
`#5`  0x00007ffff7d5dcbf in EVP_RAND_state (ctx=0x55555557c4f0) at crypto/evp/evp_rand.c:592
`#6`  0x00007ffff7dc693b in RAND_status () at crypto/rand/rand_lib.c:243
`#7`  0x00005555555551d1 in main (argc=1, argv=0x7fffffffe0a8) at rand-status.c:11


_(*) BTW: Not sure whether I do like the `_locked` suffix. I still prefer the more generic UPPER_CASE_function() vs. lower_case_function() pattern which simply means public vs. internal function doing the same thing. Maybe a comment at the internal function would have been sufficient. But YMMV._

crypto/evp/evp_rand.c
paulidale: Only if `get_ctx_params` is non-NULL which is guaranteed by the changes you noted later.

crypto/evp/evp_rand.c
paulidale: This guarantees that `get_context_params` is non-NULL.  The entire EVP_RAND breaks (well returns errors for just about everything) if it is NULL -- can't get _status_ or _strength_ otherwise.

paulidale:
> Gotcha!

It isn't one you predicted, its one that got added making the change to doubled functions :)

crypto/evp/evp_rand.c
paulidale: These two variables simply cache the values in the provider rand.
It's tempting to remove them and call through all the time.

paulidale: I've gone ahead and removed these two.  They are now always queried from the underlying PROV_RAND object.
This will avoid any possible cache incoherency problems.

mspncp: I don't quite get the reason for removing the caching. Originally, it was a speed optimization, I guess. Now it has been removed on the ground taht it avoids cache incoherency, even though `evp_rand_set_ctx_params_locked` is holding the rand lock (if enabled).

paulidale:
I'm more worried about the value becoming incorrect.  There is no signal from the PROV_RAND to EVP_RAND to indicate that something has changed.  It's a bit out there but consider an HRNG which consists of multiple generators which are concatenated.  If one such generator fails, the strength could reduce.

The `max_request` doesn't need to be cached: it was only queried before generate which will be relatively expensive.
There might be some reason to cache `strength` still but let's wait for some profile data.

I'll definitely merge with this commit kept separate.  Does it really need to be a separate PR?

mspncp:
> I'll definitely merge with this commit kept separate. Does it really need to be a separate PR?

Ok, you’re right. It‘s sufficient to keep it a separate commit.

paulidale:
Merged to master.  Thanks for the input.

PR #12322, opened 2020-06-30
doc: remove reference to the predecessor of SHA-1.

### LINKED ISSUE #8407: Documentation bug: SHA is no longer supported in 1.1.1
Have just built OpenSSL 1.1.1b, and unit tests in other software failed to build because SHA support is no longer present.

https://www.openssl.org/docs/man1.1.1/man3/SHA1_Init.html states "The predecessor of SHA-1, SHA, is also implemented, but it should be used only when backward compatibility is required."

paulidale:
Removal of stale documentation.

Fixes #8407


paulidale:
Merged to master and 1.1.1, thanks.

PR #12267, opened 2020-06-24
rand: fix CPU randomness source

paulidale:
Fixes #12261

PR #12325, opened 2020-06-30
rand: detect if FIPS approved randomness sources are being used.

paulidale:
As mentioned in #12267, the EGD source doesn't build with FIPS enabled.

The only permitted sources for the FIPS provider are the operating system sources and RDRAND.  All other sources are not available in the FIPS module.

### DIFF
--- a/providers/implementations/rands/seeding/rand_unix.c
+++ b/providers/implementations/rands/seeding/rand_unix.c
@@ -37,6 +37,24 @@
 # include <sys/param.h>
 #endif
 
+/*
+ * Provide a compile time error if the FIPS module is being build and none
+ * of the supported entropy sources are available.
+ */
+#if defined(FIPS_MODULE)
+# if !defined(OPENSSL_RAND_SEED_GETRANDOM) \
+     && !defined(OPENSSL_RAND_SEED_DEVRANDOM) \
+     && !defined(OPENSSL_RAND_SEED_RDCPU) \
+     && !defined(OPENSSL_RAND_SEED_OS)
+#  error FIPS mode without supported randomness source
+# endif
+/* Remove the sources that are not permitted in FIPS */
+# undef OPENSSL_RAND_SEED_LIBRANDOM
+# undef OPENSSL_RAND_SEED_RDTSC
+# undef OPENSSL_RAND_SEED_EGD
+# undef OPENSSL_RAND_SEED_NONE
+#endif
+
 #if (defined(OPENSSL_SYS_UNIX) && !defined(OPENSSL_SYS_VXWORKS)) \
      || defined(__DJGPP__)
 # include <sys/types.h>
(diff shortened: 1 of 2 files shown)


richsalz:
I understand why quietly ignoring non-compliant options is a reasonable thing to do.  But I think this needs some documentation in INSTALL.md under the random number section.  (And also, why two sections?  Seems needless, but I digress...)

paulidale:
Good point, I've added some words to INSTALL.md.

paulidale:
Ping @openssl/committers for review

paulidale:
Reping @openssl/committers for review

providers/implementations/rands/seeding/rand_unix.c
```diff
@@ -37,6 +37,24 @@
 # include <sys/param.h>
 #endif
 
+/*
```
mspncp, line 40: being _built_

providers/implementations/rands/seeding/rand_unix.c
```diff
@@ -37,6 +37,24 @@
...
+ */
+#if defined(FIPS_MODULE)
+# if !defined(OPENSSL_RAND_SEED_GETRANDOM) \
+     && !defined(OPENSSL_RAND_SEED_DEVRANDOM) \
+     && !defined(OPENSSL_RAND_SEED_RDCPU) \
+     && !defined(OPENSSL_RAND_SEED_OS)
+#  error FIPS mode without supported randomness source
+# endif
+/* Remove the sources that are not permitted in FIPS */
+# undef OPENSSL_RAND_SEED_LIBRANDOM
```
mspncp, line 52: How about adding a `#warning` for each if it is defined?

providers/implementations/rands/seeding/rand_unix.c
```diff
@@ -695,7 +715,7 @@ size_t prov_pool_acquire_entropy(RAND_POOL *pool)
         return entropy_available;
 #   endif
 
-#   if defined(OPENSSL_RAND_SEED_EGD)
+#   if defined(OPENSSL_RAND_SEED_EGD) && !defined(FIPS_MODULE)
```
mspncp, line 718: Why the extra check after explicitly undef‘ing EGD?

(same question for the other defines)

providers/implementations/rands/seeding/rand_unix.c
paulidale: This would break warnings as errors builds where the sources are wanted outside of FIPS mode.

providers/implementations/rands/seeding/rand_unix.c
paulidale: There is no need for the extra check, I've removed it.

providers/implementations/rands/seeding/rand_unix.c
paulidale: Okay.

paulidale:
Comments addressed.

paulidale:
Ping @mspncp

t8m:
The Travis failure is the usual unrelated one. I wonder why is it really happenning. Unsure how to investigate it.

paulidale:
Merged to master.  Thanks for the reviews and input.

PR #12326, opened 2020-06-30
Run tests in parallel

romen:
The environment variable `HARNESS_JOBS` can be used to control how many
jobs to run in parallel. It defaults to 4.

This commit does not define custom `rules`, and different versions of
`TAP::Harness` come with different strategies regarding the default
`rules` that define which test recipes can be run in parallel.
In recent versions of Perl, unless specified otherwise any task can be
run in parallel.

##### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -173,6 +173,7 @@ before_script:
           fi;
       fi
     - ./configdata.pm --dump
+    - export HARNESS_JOBS=${HARNESS_JOBS:-4}
     - cd $top
 
 script:
--- a/test/README.md
+++ b/test/README.md
@@ -114,15 +114,15 @@ starting with "test_ssl_":
 
 Run only test group 10:
 
-    $ make TESTS='10'
+    $ make TESTS='10' test
 
 Run all tests except the slow group (group 99):
 
-    $ make TESTS='-99'
+    $ make TESTS='-99' test
 
 Run all tests in test groups 80 to 99 except for tests in group 90:
 
-    $ make TESTS='[89]? -90'
+    $ make TESTS='[89]? -90' test
 
 To stochastically verify that the algorithm that produces uniformly distributed
 random numbers is operating correctly (with a false positive rate of 0.01%):
--- a/test/README.md
+++ b/test/README.md
@@ -134,10 +134,10 @@ Running Tests in Parallel
 
 By default the test harness will execute the selected tests sequentially.
 Depending on the platform characteristics, running more than one test job in
-parallel may speed up test execution. When desirable it is possible to run test
-jobs in parallel via the `HARNESS_JOBS` environment variable: its value should
-be set to the number (a positive integer equal or greater than 1) of maximum
-test jobs to run in parallel.
+parallel may speed up test execution.
+This can be requested by setting the `HARNESS_JOBS` environment variable to a
+positive integer value. This specifies the maximum number of test jobs to run in
+parallel.
 
 Depending on the Perl version different strategies could be adopted to select
 which test recipes can be run in parallel.  In recent versions of Perl, unless
--- a/test/README.md
+++ b/test/README.md
@@ -132,9 +132,12 @@ random numbers is operating correctly (with a false positive rate of 0.01%):
 Running Tests in Parallel
 -------------------------
 
-By default the test harness will execute the selected tests sequentially. When
-desirable it is possible to run test jobs in parallel via the `HARNESS_JOBS`
-environment variable.
+By default the test harness will execute the selected tests sequentially.
+Depending on the platform characteristics, running more than one test job in
+parallel may speed up test execution. When desirable it is possible to run test
+jobs in parallel via the `HARNESS_JOBS` environment variable: its value should
+be set to the number (a positive integer equal or greater than 1) of maximum
+test jobs to run in parallel.
 
 Depending on the Perl version different strategies could be adopted to select
 which test recipes can be run in parallel.  In recent versions of Perl, unless
--- a/test/README.md
+++ b/test/README.md
@@ -128,3 +128,19 @@ To stochastically verify that the algorithm that produces uniformly distributed
 random numbers is operating correctly (with a false positive rate of 0.01%):
 
     $ ./util/wrap.sh test/bntest -stochastic
+
+Running Tests in Parallel
+-------------------------
+
+By default the test harness will execute the selected tests sequentially. When
+desirable it is possible to run test jobs in parallel via the `HARNESS_JOBS`
+environment variable.
+
+Depending on the Perl version different strategies could be adopted to select
+which test recipes can be run in parallel.  In recent versions of Perl, unless
+specified otherwise, any task can be run in parallel. Consult the documentation
+for `TAP::Harness` to know more.
+
+To run up to four tests in parallel at any given time:
+
+    $ make HARNESS_JOBS=4 test
--- a/test/run_tests.pl
+++ b/test/run_tests.pl
@@ -20,6 +20,11 @@ BEGIN {
         if $ENV{VERBOSE_FAILURES_PROGRESS} || $ENV{VFP};
 }
 
+# Support running jobs in parallel
+BEGIN {
+    $ENV{HARNESS_JOBS} = "4" if (!defined $ENV{HARNESS_JOBS});
+}
+
 use File::Spec::Functions qw/catdir catfile curdir abs2rel rel2abs/;
 use File::Basename;
 use FindBin;
@@ -42,6 +47,7 @@ my %tapargs =
       lib               => [ $libdir ],
       switches          => '-w',
       merge             => 1,
+      jobs              => $ENV{HARNESS_JOBS},
     );
 
 # Additional OpenSSL special TAP arguments.  Because we can't pass them via
--- a/test/run_tests.pl
+++ b/test/run_tests.pl
@@ -20,11 +20,6 @@ BEGIN {
         if $ENV{VERBOSE_FAILURES_PROGRESS} || $ENV{VFP};
 }
 
-# Support running jobs in parallel
-BEGIN {
-    $ENV{HARNESS_JOBS} = "4" if (!defined $ENV{HARNESS_JOBS});
-}
-
 use File::Spec::Functions qw/catdir catfile curdir abs2rel rel2abs/;
 use File::Basename;
 use FindBin;
--- a/test/run_tests.pl
+++ b/test/run_tests.pl
@@ -35,6 +35,7 @@ my $srctop = $ENV{SRCTOP} || $ENV{TOP};
 my $bldtop = $ENV{BLDTOP} || $ENV{TOP};
 my $recipesdir = catdir($srctop, "test", "recipes");
 my $libdir = rel2abs(catdir($srctop, "util", "perl"));
+my $jobs = $ENV{HARNESS_JOBS};
 
 $ENV{OPENSSL_CONF} = rel2abs(catdir($srctop, "apps", "openssl.cnf"));
 $ENV{OPENSSL_CONF_INCLUDE} = rel2abs(catdir($bldtop, "providers"));
@@ -47,8 +48,9 @@ my %tapargs =
       lib               => [ $libdir ],
       switches          => '-w',
       merge             => 1,
-      jobs              => $ENV{HARNESS_JOBS},
     );
+
+$tapargs{jobs} = $jobs if defined $jobs;
 
 # Additional OpenSSL special TAP arguments.  Because we can't pass them via
 # TAP::Harness->new(), they will be accessed directly, see the


romen:
@levitte @DDvO thoughts?

levitte:
I'm not terribly keen on making this a default thing.  It should be at the discretion of the user.  In other words, I would prefer that we document that it's possible to specify the number of jobs in some way, same as we do for V, VF, etc.  The run-tests.pl code could then look like this:

``` perl
my $jobs = $ENV{HARNESS_JOBS}
$tapargs{jobs} = $jobs if defined $jobs;
```

paulidale:
Do we know if there are any dependencies between tests?
Running the test suite backwards (99 - 1) would likely find any.

romen:
@paulidale 
> Do we know if there are any dependencies between tests?
> Running the test suite backwards (99 - 1) would likely find any.


Testing on my [private branch](https://github.com/romen/openssl/tree/features/3.0.0/sequential_reversed_tests):

- [Travis](https://github.com/romen/openssl/runs/822492065)
- [AppVeyor](https://ci.appveyor.com/project/romen/openssl/builds/33821802)

romen:
@levitte I pushed a few commits to do as you suggested:

- 8e65ddee21 squash! Run tests in parallel  _(your logic + documentation)_
- 4d2e7095f0 Travis: default to HARNESS_JOBS=4
- 44cd1195f9 [test/README.md] minor fix of examples missing the test target _(fixes some trivial things in the examples that we recently overlooked, this is unrelated to the rest of this PR, @DDvO might also want to comment on it)_

levitte:
> Do we know if there are any dependencies between tests?

We wouldn't be able to run individual tests if that was the case.
*There must be no dependencies between test recipes!*

levitte:
From the looks of it, this looks good.
Because you added this for Travis as well, I'd like to wait for the CIs before rubber-stamping

test/README.md
```diff
@@ -114,17 +114,33 @@ starting with "test_ssl_":
...
 To stochastically verify that the algorithm that produces uniformly distributed
 random numbers is operating correctly (with a false positive rate of 0.01%):
 
     $ ./util/wrap.sh test/bntest -stochastic
+
+Running Tests in Parallel
+-------------------------
+
+By default the test harness will execute the selected tests sequentially. When
+desirable it is possible to run test jobs in parallel via the `HARNESS_JOBS`
```
DDvO, line 136: Instead of "When desirable" I'd say: "For speeding up test execution"

test/run_tests.pl
```diff
@@ -20,6 +20,11 @@ BEGIN
         if $ENV{VERBOSE_FAILURES_PROGRESS} || $ENV{VFP};
 }
 
+# Support running jobs in parallel
+BEGIN {
+    $ENV{HARNESS_JOBS} = "4" if (!defined $ENV{HARNESS_JOBS});
```
DDvO, line 25: This looks like 4 parallel jobs still becomes the default here, which @levitte did not recommend.

BTW, I would actually prefer having this as the default.

DDvO: If we keep it, the following would be simpler: 
```
$ENV{HARNESS_JOBS} = $ENV{HARNESS_JOBS} // "4";
```

DDvO: or even simpler: ``ENV{HARNESS_JOBS} //= "4";

levitte: Just remove this whole BEGIN section.  I thought it was gone already, must have given it a blind eye...

DDvO:
@romen, I would appreciate if you bring forward this (nearly finished) PR quickly,
since it's going to save all of us quite some time when running (the whole collection of) tests.

DDvO:
Can we make this parallelization of test runs the default (not for all `make test` calls but) for the Travis CI builds?

test/run_tests.pl
romen: I also thought I got rid of this! Apologies!

test/README.md
romen: Well, it is debatable that this speeds up things. Running the jobs in parallel when you don't have enough cores or resources for it can actually slow it down!

.travis.yml
```diff
@@ -173,6 +173,7 @@ before_script:
           fi;
       fi
     - ./configdata.pm --dump
+    - export HARNESS_JOBS=${HARNESS_JOBS:-4}
```
romen, line 176: @DDvO this should make it the default for all Travis builds (in theory)

test/README.md
DDvO: Why else would one desire to run (part of) the tests in parallel?

.travis.yml
romen: The thought is that individual targets can override this default by defining a different HARNESS_JOBS value in their env property

test/README.md
DDvO: At least in my experience with building OpenSSL (even when using ccache) using a dual-core CPU with hyperthreading (such that it virtually can run 4 parallel threads) is that `make -j4` is *much* faster than `make`.

romen: Yeah that is the goal, the problem I see with "For speeding up test execution it is possible to run test jobs in parallel" is that an unexperienced reader might wrongly conclude that enabling parallelization of tests will always deliver a speedup when the opposite can be true depending on the platform characteristics!

.travis.yml
DDvO: >  this should make it the default for all Travis builds (in theory)

Excellent.

test/README.md
romen: (I also have the exact experience, and the tests are usually even less resource hungry than the compilation steps. Problem is if someone tries to enable this to speed up their testing on a single core ARM CPU or a container/vm restrained to a single core, attempting to multiplex might make things worse!

DDvO: I now see your point. Then how about this?

When desirable it is possible to run test jobs in parallel, which may speed up test execution depending on the platform characteristics, ...

romen: Fixup commit coming soon to address the feedback!

romen: @DDvO I went for this:

~~~diff
diff --git a/test/README.md b/test/README.md
index d046765494..b5080a3a6b 100644
--- a/test/README.md
+++ b/test/README.md
@@ -132,9 +132,12 @@ random numbers is operating correctly (with a false positive rate of 0.01%):
Running Tests in Parallel
-------------------------

By default the test harness will execute the selected tests sequentially.
{+Depending on the platform characteristics, running more than one test job in+}
{+parallel may speed up test execution.+} When desirable it is possible to run test
jobs in parallel via the `HARNESS_JOBS` environment [-variable.-]{+variable: its value should+}
{+be set to the number (a positive integer equal or greater than 1) of maximum+}
{+test jobs to run in parallel.+}

Depending on the Perl version different strategies could be adopted to select
which test recipes can be run in parallel.  In recent versions of Perl, unless
~~~

thoughts?

beldmit:
Could we somehow detect a number of cores and use it as a reasonable default?

romen:
> Could we somehow detect a number of cores and use it as a reasonable default?

@beldmit I would leave that to a separate PR.

I see it as potentially complex and not worth it: even just in Linux across different distros tools to do this differ, and looking into `/proc` might not give the full picture due to `cpuset`/`namespace`/containerization or just the user trying to stick some parts of the processes tree to some set of cores via affinity and whatnot.

Doing this across different OS seems to be even more risky and complex: maybe we can add a CHANGE or NEWS entry to highlight this change so that more users are aware of it and can try to benefit from it, but so far I have refrained from doing so because this change is limited to test execution and might not be newsworthy!

beldmit:
> @beldmit I would leave that to a separate PR.

Definitely. And even if the proposed solution is Linux-only, it still will make sense.

romen:
My initial tradeoff was on making it default to 4 testing jobs, as I expect more users would benefit from this than those that will suffer performance degradation.
For those that would be negative impacted by this, setting `HARNESS_JOBS` to `1` (or a greater value for those that can support an even higher level of parallelization) seemed a low price to pay.

But @levitte recommended to stick the default to 1, and I think he has a better perspective than me to decide on this tradeoff, that is why I went and applied the suggested change rather than defend my original proposition.

test/README.md
DDvO: I propose further adaptations to:

By default the test harness will execute the selected tests sequentially.
Depending on the platform characteristics, running test jobs in parallel may speed up test execution.
This can be requested by setting the `HARNESS_JOBS` environment variable to a positive integer value.
This specifies the maximum number of test jobs to run in parallel.

DDvO:
> My initial tradeoff was on making it default to 4 testing jobs, as I expect more users would benefit from this than those that will suffer performance degradation.
> For those that would be negative impacted by this, setting `HARNESS_JOBS` to `1` (or a greater value for those that can support an even higher level of parallelization) seemed a low price to pay.

I agree on this.

> But @levitte recommended to stick the default to 1, and I think he has a better perspective than me to decide on this tradeoff, that is why I went and applied the suggested change rather than defend my original proposition.

I thought that @levitte was just against making 4 the general default for `make test`, right?
I would not expect a risk for performance degradation on the machine(s) running our CI runs.

romen:
I just pushed 2 fixup commits:
- the first removes the leftover `BEGIN {}` section that was messing up the `HARNESS_JOBS` env var (I could have sweared I did remove that already, but `git` does not lie :smile: )
- the second touches up the documentation to address @DDvO feedback on that.

I also went to compare execution time on Travis of this PR (in the state before these fixup commits) with some randomly picked PR builds on `openssl/openssl`: there is quite a lot of variance, and this was a low effort attempt (statistical value: 0, it was basically very coarse eyeballing), but in the 5 or so successful builds I compared timings were higher than the timings recorded for this PR, so it looks promising.

In an attempt to be more rigorous about validating 4 as the default for Travis I pushed a branch on my personal fork, running the same commits, but setting the default back to 1.
I don't know if PR builds on `openssl/openssl` might have been assigned more resources from Travis, so it might be better to compare 2 Travis branch builds on my personal `romen/openssl` fork to set the field equally.

Here they are:

- https://travis-ci.com/github/romen/openssl/builds/173850679 ([defaults to 4], the source branch of this PR)
- https://travis-ci.com/github/romen/openssl/builds/173851664 ([defaults to 1])

[defaults to 4]: https://github.com/romen/openssl/tree/features/3.0.0/parallel_tests
[defaults to 1]: https://github.com/romen/openssl/tree/features/3.0.0/parallel_tests_not_in_Travis

DDvO:
> the second touches up the documentation to address @DDvO feedback on that.

Did you see my new suggestion for further improving that paragraph?

romen:
> @paulidale
> 
> > Do we know if there are any dependencies between tests?
> > Running the test suite backwards (99 - 1) would likely find any.
> 
> Testing on my [private branch](https://github.com/romen/openssl/tree/features/3.0.0/sequential_reversed_tests):
> 
> * [Travis](https://github.com/romen/openssl/runs/822492065)
> * [AppVeyor](https://ci.appveyor.com/project/romen/openssl/builds/33821802)

@paulidale Running the tests in reversed order succeeds (travis has some failures, but are unrelated)

romen:
> > the second touches up the documentation to address @DDvO feedback on that.
> 
> Did you see my new suggestion for further improving that paragraph?

I didn't before!
Just pushed another fixup commit with your latest proposal: 57ea094fdc31440ccd956962abdd6388cf7f2c05

test/README.md
romen: Addressed in 57ea094fdc31440ccd956962abdd6388cf7f2c05

romen:
> > But @levitte recommended to stick the default to 1, and I think he has a better perspective than me to decide on this tradeoff, that is why I went and applied the suggested change rather than defend my original proposition.
> 
> I thought that @levitte was just against making 4 the _general_ default for `make test`, right?
> I would not expect a risk for performance degradation on the machine(s) running our _CI runs_.

Yes, he was talking about not making it default to 4 for any run of `make test`, so I adjusted it as suggested to behave exactly as before (current default of `TAP::Harness` is 1 job), unless `HARNESS_JOBS` is defined, in which case we pass that value to `TAP:Harness`.

For people running our CI runs, specifically our Travis CI runs, they will get the default of 4 test jobs, as it is now set in the `before_script:` step which should apply to all jobs in the Travis matrix.

I am not familiar with Appveyor: does it make sense to alter the default for Appveyour? How do you suggest to do so?

romen:
> I just pushed 2 fixup commits:
> 
> * the first removes the leftover `BEGIN {}` section that was messing up the `HARNESS_JOBS` env var (I could have sweared I did remove that already, but `git` does not lie smile )
> * the second touches up the documentation to address @DDvO feedback on that.
> 
> I also went to compare execution time on Travis of this PR (in the state before these fixup commits) with some randomly picked PR builds on `openssl/openssl`: there is quite a lot of variance, and this was a low effort attempt (statistical value: 0, it was basically very coarse eyeballing), but in the 5 or so successful builds I compared timings were higher than the timings recorded for this PR, so it looks promising.
> 
> In an attempt to be more rigorous about validating 4 as the default for Travis I pushed a branch on my personal fork, running the same commits, but setting the default back to 1.
> I don't know if PR builds on `openssl/openssl` might have been assigned more resources from Travis, so it might be better to compare 2 Travis branch builds on my personal `romen/openssl` fork to set the field equally.
> 
> Here they are:
> 
> * https://travis-ci.com/github/romen/openssl/builds/173850679 ([defaults to 4](https://github.com/romen/openssl/tree/features/3.0.0/parallel_tests), the source branch of this PR)
> * https://travis-ci.com/github/romen/openssl/builds/173851664 ([defaults to 1](https://github.com/romen/openssl/tree/features/3.0.0/parallel_tests_not_in_Travis))

There are unrelated failures, but the experiment suggests that defaulting to 4 test jobs for Travis is indeed beneficial.

romen:
(I added the approval done label on the basis of @DDvO approval)

DDvO:
I've just done some experiments: the acceleration is very good!

It turns out I was wrong on the data of my CPU (an i5-8365U): 
it actually has 4 cores supporting 2 threads each, 
so `HARNESS_JOBS` up to should make sense,
and the experiments yield exactly that:

sequential
real	14m40.176s
user	13m47.235s
sys	1m1.817s

2 parallel
real	8m34.269s
user	14m4.113s
sys	1m5.732s

4 parallel
real	6m26.297s
user	15m53.536s
sys	1m19.386s

8 parallel
real	5m53.235s
user	18m45.684s
sys	1m47.853s

16 parallel
real	6m24.378s
user	20m57.687s
sys	1m56.835s

romen:
@levitte do you also want to officially rubber stamp?

paulidale:
> @levitte do you also want to officially rubber stamp?

This isn't required, the 24 hour grace period before merging is for comments by others.
Nice to ask though.

romen:
> > @levitte do you also want to officially rubber stamp?
> 
> This isn't required, the 24 hour grace period before merging is for comments by others.
> Nice to ask though.

Sure, not required, but he expressed an intention to do so once the CI was done, and I know he has a lot on his plate!
This was meant as a courtesy ping once I noticed all the checks were green!

romen:
Merged to `master` as:

- a20c9075d6 Run tests in parallel
- af3e8c298a Travis: default to HARNESS_JOBS=4
- 9576c498ca [test/README.md] minor fix of examples missing the test target


Thanks all for reviews and feedback!

romen:
@mattcaswell @levitte please have a look at https://github.com/openssl/tools/pull/71 to make `run-checker` use 4 test jobs in parallel.

DDvO:
Can this be backported to the v1.1.1 test machinery?

romen:
@DDvO 

> Can this be backported to the v1.1.1 test machinery?

See #12362, it's WIP because in `1.1.1` test jobs are currently not as isolated as they are in `master`.

P.S. I find it pleasing that #12362 is the backport of #12326 , but I swear it was not done intentionally!

DDvO:
> P.S. I find it pleasing that #12362 is the backport of #12326 , but I swear it was not done intentionally!

We could make this an intentional rule :rofl:

PR #1320, opened 2016-07-15
zero pad DHE public key in ServerKeyExchange message for interop

russor:
Some versions of the Microsoft TLS stack have problems when the DHE public key
is encoded with fewer bytes than the DHE prime.

There's some public acknowledgement of the bug at these links:

https://connect.microsoft.com/IE/feedback/details/1253526/tls-serverkeyexchange-with-1024-dhe-may-encode-dh-y-as-127-bytes-breaking-internet-explorer-11
https://connect.microsoft.com/IE/feedback/details/1104905/wininet-calculation-of-mac-in-tls-handshake-intermittently-fails-for-dhe-rsa-key-exchange

This encoding issue also causes the same errors with 2048-bit DHE, if the
public key is encoded in fewer than 256 bytes and includes the TLS stack on
Windows Phone 8.x.

Pull request against master, but can be easily backported to 1.0.2 and 1.0.1 (if eligible), in those versions the logic is in ssl/s3_srvr.c and the added lines could be copied as-is.

PR #12331, opened 2020-06-30
also zero pad DHE public key in ClientKeyExchange message for interop

heinzelotto:
The Windows SChannel TLS stack has problems when the DHE pub key is encoded with fewer bytes than the DHE prime.

In https://github.com/openssl/openssl/pull/1320 padding has already been added to work around this problem for the OpenSSL _server_ case, and I experimented a little and Microsoft seems to have also fixed this in their _client_ code.

I guess that OpenSSL server <-> windows client is the more common case, but the opposite case still causes handshakes to randomly fail when a DHE suite is selected. By adding padding for the OpenSSL client key exchange I hope to fix that.

This is my first contribution to OpenSSL, I tried to follow the style guide and conventions. Any corrections are much appreciated.

### DIFF
--- a/ssl/statem/statem_clnt.c
+++ b/ssl/statem/statem_clnt.c
@@ -3057,9 +3057,9 @@ static int tls_construct_cke_dhe(SSL *s, WPACKET *pkt)
 {
 #ifndef OPENSSL_NO_DH
     DH *dh_clnt = NULL;
-    const BIGNUM *pub_key;
     EVP_PKEY *ckey = NULL, *skey = NULL;
     unsigned char *keybytes = NULL;
+    int np;
 
     skey = s->s3.peer_tmp;
     if (skey == NULL) {
@@ -3089,15 +3089,19 @@ static int tls_construct_cke_dhe(SSL *s, WPACKET *pkt)
     }
 
     /* send off the data */
-    DH_get0_key(dh_clnt, &pub_key, NULL);
-    if (!WPACKET_sub_allocate_bytes_u16(pkt, BN_num_bytes(pub_key),
-                                        &keybytes)) {
+    np = BN_num_bytes(DH_get0_p(dh_clnt));
+    /*
+     * for interoperability with some versions of the Microsoft TLS
+     * stack, we need to zero pad the DHE pub key to the same length
+     * as the prime, so use the length of the prime here
+     */
+    if (!WPACKET_sub_allocate_bytes_u16(pkt, np, &keybytes)) {
         SSLfatal(s, SSL_AD_INTERNAL_ERROR, SSL_F_TLS_CONSTRUCT_CKE_DHE,
                  ERR_R_INTERNAL_ERROR);
         goto err;
     }
 
-    BN_bn2bin(pub_key, keybytes);
+    BN_bn2binpad(DH_get0_pub_key(dh_clnt), keybytes, np);
     EVP_PKEY_free(ckey);
 
     return 1;
--- a/ssl/statem/statem_clnt.c
+++ b/ssl/statem/statem_clnt.c
@@ -3095,13 +3095,13 @@ static int tls_construct_cke_dhe(SSL *s, WPACKET *pkt)
      * stack, we need to zero pad the DHE pub key to the same length
      * as the prime, so use the length of the prime here.
      */
-    if (!WPACKET_sub_allocate_bytes_u16(pkt, prime_len, &keybytes)) {
+    if (!WPACKET_sub_allocate_bytes_u16(pkt, prime_len, &keybytes)
+            || BN_bn2binpad(DH_get0_pub_key(dh_clnt), keybytes, prime_len) < 0) {
         SSLfatal(s, SSL_AD_INTERNAL_ERROR, SSL_F_TLS_CONSTRUCT_CKE_DHE,
                  ERR_R_INTERNAL_ERROR);
         goto err;
     }
 
-    BN_bn2binpad(DH_get0_pub_key(dh_clnt), keybytes, prime_len);
     EVP_PKEY_free(ckey);
 
     return 1;
--- a/ssl/statem/statem_clnt.c
+++ b/ssl/statem/statem_clnt.c
@@ -3059,7 +3059,7 @@ static int tls_construct_cke_dhe(SSL *s, WPACKET *pkt)
     DH *dh_clnt = NULL;
     EVP_PKEY *ckey = NULL, *skey = NULL;
     unsigned char *keybytes = NULL;
-    int np;
+    int prime_len;
 
     skey = s->s3.peer_tmp;
     if (skey == NULL) {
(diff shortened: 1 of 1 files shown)


ssl/statem/statem_clnt.c
```diff
@@ -3089,14 +3092,26 @@ static int tls_construct_cke_dhe(SSL *s, WPACKET *pkt)
...
         goto err;
     }
 
+    npad = (BN_num_bytes(p) - BN_num_bytes(pub_key));
+    for(j = 0; j < npad; ++j) {
+        *keybytes = 0;
+        ++keybytes;
+    }
+
     BN_bn2bin(pub_key, keybytes);
```
davidben, line 3115: You may want to look at `BN_bn2binpad`.

heinzelotto: Thank you, good idea! I need to close and reopen the pull request because of cla-check.

ssl/statem/statem_clnt.c
```diff
@@ -3089,15 +3091,21 @@ static int tls_construct_cke_dhe(SSL *s, WPACKET *pkt)
     }
 
     /* send off the data */
-    DH_get0_key(dh_clnt, &pub_key, NULL);
-    if (!WPACKET_sub_allocate_bytes_u16(pkt, BN_num_bytes(pub_key),
-                                        &keybytes)) {
+    DH_get0_pqg(dh_clnt, &p, NULL, NULL);
+    np = BN_num_bytes(p);
```
davidben, line 3095: `BN_num_bytes(DH_get0_p(dh_clnt))` is probably shorter.

ssl/statem/statem_clnt.c
```diff
@@ -3089,15 +3091,21 @@ static int tls_construct_cke_dhe(SSL *s, WPACKET *pkt)
...
+     */
+    if (!WPACKET_sub_allocate_bytes_u16(pkt, np, &keybytes)) {
         SSLfatal(s, SSL_AD_INTERNAL_ERROR, SSL_F_TLS_CONSTRUCT_CKE_DHE,
                  ERR_R_INTERNAL_ERROR);
         goto err;
     }
 
-    BN_bn2bin(pub_key, keybytes);
+    DH_get0_key(dh_clnt, &pub_key, NULL);
+    BN_bn2binpad(pub_key, keybytes, np);
```
davidben, line 3108: Likewise, `BN_bn2binpad(DH_get0_pub_key(dh_clnt, ...))`

heinzelotto:
@mattcaswell are you familiar with the WPACKET API and would you taking a look?

ssl/statem/statem_clnt.c
```diff
@@ -3089,15 +3089,19 @@ static int tls_construct_cke_dhe(SSL *s, WPACKET *pkt)
...
 
     /* send off the data */
-    DH_get0_key(dh_clnt, &pub_key, NULL);
-    if (!WPACKET_sub_allocate_bytes_u16(pkt, BN_num_bytes(pub_key),
-                                        &keybytes)) {
+    np = BN_num_bytes(DH_get0_p(dh_clnt));
+    /*
+     * for interoperability with some versions of the Microsoft TLS
+     * stack, we need to zero pad the DHE pub key to the same length
+     * as the prime, so use the length of the prime here
```
kaduk, line 3096: If we have to make other changes, we could use a capital 'F' to start the comment and a period at the end, to make it a complete sentence, but it's not worth changing just to fix that.
(And the ServerKeyExchange case doesn't do either, anyway.)

ssl/statem/statem_clnt.c
```diff
@@ -3057,9 +3057,9 @@ static int tls_construct_cke_dhe(SSL *s, WPACKET *pkt)
 {
 #ifndef OPENSSL_NO_DH
     DH *dh_clnt = NULL;
-    const BIGNUM *pub_key;
     EVP_PKEY *ckey = NULL, *skey = NULL;
     unsigned char *keybytes = NULL;
+    int np;
```
kaduk, line 3062: I'd consider a different name for this variable, maybe `prime_len` -- I'm not actually sure what "np" is supposed to indicate or stand for.

kaduk:
I'm going to dismiss @davidben 's review, since it was before the force-push.
(In some sense it doesn't matter, since we're both committers and this needs an OTC review in order to get merged.)

kaduk:
(the travis failures are the spurious "../../_srcdir/config: no such file or directory" that we've been seeing lately)

heinzelotto:
Thank you @kaduk for your comments. I wanted to ask whether there were any necessary steps on my side once marked as "approval: otc review pending"? And I figured that if I am going to do an additional round trip anyway, I will just add the changes you suggested. (Hopefully I did the fixup! correctly)

kaduk:
@heinzelotto the fixup looks good, thank you.
With the "approval: otc review pending" set we are not waiting on anything else from you (unless that review produces additional feedback/requested changes).  Unfortunately, OTC time is in high demand right now, with the work towards the 3.0 release, so I don't have a great estimate for you.

ssl/statem/statem_clnt.c
```diff
@@ -3089,15 +3089,19 @@ static int tls_construct_cke_dhe(SSL *s, WPACKET *pkt)
...
+     * as the prime, so use the length of the prime here.
+     */
+    if (!WPACKET_sub_allocate_bytes_u16(pkt, prime_len, &keybytes)) {
         SSLfatal(s, SSL_AD_INTERNAL_ERROR, SSL_F_TLS_CONSTRUCT_CKE_DHE,
                  ERR_R_INTERNAL_ERROR);
         goto err;
     }
 
-    BN_bn2bin(pub_key, keybytes);
+    BN_bn2binpad(DH_get0_pub_key(dh_clnt), keybytes, prime_len);
```
t8m, line 3104: Please add check for the return value. Although it is in practice "can't happen" condition, if prime_len is lower than the length of the actual pub_key, the function will fail. You can move the call to the if condition above.

t8m (approved):
LGTM

t8m:
Merged to master. Thank you for your contribution and patience when waiting for the reviews.

paveljanik:
This makes openssl as the client send out zero padded DHE pubkey in TLSv1.2 which means it breaks the standard as leading zeros should be stripped in TLSv1.2 (as opposite to TLSv1.3 where they should be left in the stream!). Is this really what you need/want?

davidben:
The spec says the resulting secret has leading zeros removed. I believe RFC5246 is silent as to whether the public key is padded. (Which is a mistake. It should have picked one, and the correct one to pick is fixed-width. But a bit late to fix that now, so implementations have to pick one.)

paveljanik:
So on the server side, you have to support (try...) both because your clients could have picked padding and not-padding? ;-)

davidben:
No, just have your parser tolerate leading zeros, as OpenSSL does. The resulting secret is different because it's a KDF input.

(Alternatively, stop doing TLS 1.2 DHE as it's beyond saving.)

PR #12333, opened 2020-07-01
Add FIPS related configuration data stubs to the default openssl app config file.

slontis:
##### Checklist


apps/openssl.cnf
```diff
@@ -11,9 +13,31 @@
...
+# List of providers to load
+[provider_sect]
+default = default_sect
+# The fips section name should match the section name inside the
+# included fipsmodule.cnf.
+# fips = fips_sect
+
+[default_sect]
+# activate = 1
+oid_section = new_oids
```
levitte, line 37: Wait, what?  This line should remain in the default section.

slontis: For some reason that made some of the tests actually hang..

slontis: I will debug it to see what is going on..

slontis: Not sure why but today it is working.. All I did was a make update and a rebuild.. weird.

levitte:
We have a misunderstanding, it seems, emanating from sudden double use if "default" in configuration terms.

There is such a thing as a "default section", which has nothing to do with the default provider, but is rather anything that comes before the first named section (i.e. the "default section" is usually unnamed in the config file).

Unfortunately, the way you have currently placed the provider related stuff, the following have been moved from the default section to the *default provider* section:

```
# Extra OBJECT IDENTIFIER info:
#oid_file		= $ENV::HOME/.oid
oid_section		= new_oids

# To use this configuration file with the "-extfile" option of the
# "openssl x509" utility, name here the section containing the
# X.509v3 extensions to use:
# extensions		=
# (Alternatively, use a configuration file that has only
# X.509v3 extensions in its main [= default] section.)
```

So, regarding location, the `openssl_conf = openssl_init` line should indeed be somewhere in the default section (i.e. anywhere before the `[ new_oids ]` section, but the rest must be placed after the whole default section (i.e. just before `[ new_oids ]` or anywhere further down the file).

levitte:
Side note: the default configuration section does have a name.  `[ default ]`.  It should be possible to actually name it, although that's something I've never experimented with...

slontis:
config.pod seems to list the following..

 [openssl_init]
 oid_section = oids
 providers = providers
 alg_section = evp_properties
 ssl_conf = ssl_configuration
 engines = engines

Should it have the oid_section in here then? (That was where I got the info from)..

apps/openssl-vms.cnf
```diff
@@ -1,7 +1,9 @@
 #
 # OpenSSL example configuration file.
-# This is mostly being used for generation of certificate requests.
+# See doc/man5/config.pod for more info.
```
levitte, line 3: Oh!  You should probably refer to `openssl-fipsinstall(1)` in the SEE ALSO in `doc/man5/config.pod`

apps/openssl.cnf
```diff
@@ -1,7 +1,9 @@
 #
 # OpenSSL example configuration file.
-# This is mostly being used for generation of certificate requests.
+# See doc/man5/config.pod for more info.
 #
+# This is mostly being used for generation of certificate requests,
```
richsalz, line 5: Is this really true?  Lots of TS stuff, for example.

slontis: I am open to suggestions.. tsa is mainly certificate related too right?

apps/openssl-vms.cnf
slontis: I added that in this PR..

apps/openssl.cnf
richsalz: not a big deal, maybe just delete the comment.  or leave as-is.

slontis: I think I will leave it as is for now.

levitte:
Merged

e45d943665e806ff49d06cfbdd566a8e2d57d56d Add FIPS related configuration data to the default openssl application configuration file

PR #12334, opened 2020-07-01
Fix build issue in generated fipskey.h when building with c++

jon-oracle:
When configuring/building with these options the compile fails due to a missing closing brace in fipskey.h:

./Configure no-deprecated enable-buildtest-c++ --strict-warnings -D_DEFAULT_SOURCE


paulidale:
Merged to master, thanks.

PR #12339, opened 2020-07-01
util/perl/OpenSSL/config.pm: move misplaced Windows and VMS entries

levitte:
OpenSSL::config::guess_system() is supposed to return system triplets.
However, for Windows and VMS, it returned the final OpenSSL config
target instead.  We move the entries for them to the table that
OpenSSL::config::map_guess() uses, so it can properly convert the
input triplet to an OpenSSL config target.

### DIFF
--- a/NOTES.WIN
+++ b/NOTES.WIN
@@ -59,7 +59,8 @@
 
  5. From the root of the OpenSSL source directory enter
     perl Configure VC-WIN32    if you want 32-bit OpenSSL or
-    perl Configure VC-WIN64A   if you want 64-bit OpenSSL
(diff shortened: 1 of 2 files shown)


levitte:
Oops.
We're not testing the target guessing code anywhere, as far as I can see...  maybe we should.

levitte:
Merged

b2bed3c6e53fd8e439bae6a66860d31ce994bb70 util/perl/OpenSSL/config.pm: move misplaced Windows and VMS entries
16328e9f6cc8bcd5ff5fef09b78374134de1f1e1 NOTE.WIN: suggest the audetecting configuration variant as well

PR #12343, opened 2020-07-01
Refactor ERR codes 2 (without the Hurd specific bits)

### LINKED ISSUE #12217: hurd failure

### LINKED ISSUE #12276: alpha3: Testsuite fails on hurd due to truncated error code

### LINKED ISSUE #12354: 3.0.0-alpha4 Platform specific error string test in errtest.c
File test/errtest.c tests against the string "Operation not permitted" for errno h, which is correct for Linux, but on Solaris 10 Sparc errno 1 gives "Not owner".

levitte:
This fixes a series of current problems with the ERR sub-system:

-   reason codes overlaps
-   known issues with recorded system errors (they get truncated to 24 bits, previously even fewer), which we now detect and skip over
-   look through a larger range of system errors, to try and catch them all (even though POSIX allows any number that fits in an **int**, most systems use a range within 1..255)

Fixes #12276
Fixes #12217
Fixes #12354

### DIFF
--- a/crypto/err/err.c
+++ b/crypto/err/err.c
@@ -173,92 +173,6 @@ static ERR_STRING_DATA *int_err_get_item(const ERR_STRING_DATA *d)
     return p;
 }
 
-#ifndef OPENSSL_NO_ERR
-/*
- * SYS_str_reasons is filled with pointers to copies of strerror() results
- * at initialization. Although POSIX permits 'errno' to have any positive
- * number that fits in an 'int', most systems use a range starting at 1,
- * and the highest number we have observed is well below 255, which is
- * our current upper limit.  Other codes will be displayed numerically
- * by ERR_error_string.
- * It is crucial that we have something for each reason code that occurs in
- * ERR_str_reasons, or bogus reason strings will be returned for SYSerr(),
- * which always gets an errno value and never one of those 'standard' reason
- * codes.
- */
-# define NUM_SYS_STR_REASONS 255
-
-/*
- * Russian and Ukrainian locales on Linux required more than 6,5 kB for the
- * errno range 1..127 (observed 2019-05-21).
- * We defined the string space to be 8 KiB at the time.  Now that we double
- * the range, we double the string space.
- */
-# define SPACE_SYS_STR_REASONS (16 * 1024)
-
-static void build_SYS_str_reasons(void)
-{
-    /* OPENSSL_malloc cannot be used here, use static storage instead */
-    static ERR_STRING_DATA SYS_str_reasons[NUM_SYS_STR_REASONS + 1];
-    static char strerror_pool[SPACE_SYS_STR_REASONS];
-    char *cur = strerror_pool;
-    size_t cnt = 0;
-    static int init = 1;
-    int i;
-    int saveerrno = get_last_sys_error();
-
-    CRYPTO_THREAD_write_lock(err_string_lock);
-    if (!init) {
-        CRYPTO_THREAD_unlock(err_string_lock);
-        return;
-    }
-
-    for (i = 1; i <= NUM_SYS_STR_REASONS; i++) {
-        ERR_STRING_DATA *str = &SYS_str_reasons[i - 1];
-
-        str->error = ERR_PACK(ERR_LIB_SYS, 0, i);
-        /*
-         * If we have used up all the space in strerror_pool,
-         * there's no point in calling openssl_strerror_r()
-         */
-        if (str->string == NULL && cnt < sizeof(strerror_pool)) {
-            if (openssl_strerror_r(i, cur, sizeof(strerror_pool) - cnt)) {
-                size_t l = strlen(cur);
-
-                str->string = cur;
-                cnt += l;
-                cur += l;
-
-                /*
-                 * VMS has an unusual quirk of adding spaces at the end of
-                 * some (most? all?) messages. Lets trim them off.
-                 */
-                while (cur > strerror_pool && ossl_isspace(cur[-1])) {
-                    cur--;
-                    cnt--;
-                }
-                *cur++ = '\0';
-                cnt++;
-            }
-        }
-        if (str->string == NULL)
-            str->string = "unknown";
-    }
-
-    /*
-     * Now we still have SYS_str_reasons[NUM_SYS_STR_REASONS] = {0, NULL}, as
-     * required by ERR_load_strings.
-     */
-
-    init = 0;
-
-    CRYPTO_THREAD_unlock(err_string_lock);
-    /* openssl_strerror_r could change errno, but we want to preserve it */
-    set_sys_error(saveerrno);
-    err_load_strings(SYS_str_reasons);
-}
-#endif
-
 static void ERR_STATE_free(ERR_STATE *s)
 {
     int i;
@@ -330,7 +244,6 @@ int ERR_load_ERR_strings(void)
 
     err_load_strings(ERR_str_libraries);
     err_load_strings(ERR_str_reasons);
-    build_SYS_str_reasons();
 #endif
     return 1;
 }
@@ -577,8 +490,8 @@ static unsigned long get_error_values(ERR_GET_ACTION g,
 
 void ERR_error_string_n(unsigned long e, char *buf, size_t len)
 {
-    char lsbuf[64], rsbuf[64];
-    const char *ls, *rs;
+    char lsbuf[64], rsbuf[256];
+    const char *ls, *rs = NULL;
     unsigned long f = 0, l, r;
 
     if (len == 0)
@@ -591,8 +504,13 @@ void ERR_error_string_n(unsigned long e, char *buf, size_t len)
         ls = lsbuf;
     }
 
-    rs = ERR_reason_error_string(e);
     r = ERR_GET_REASON(e);
+    if (ERR_SYSTEM_ERROR(e)) {
+        if (openssl_strerror_r(r, rsbuf, sizeof(rsbuf)))
+            rs = rsbuf;
+    } else {
+        rs = ERR_reason_error_string(e);
+    }
     if (rs == NULL) {
         BIO_snprintf(rsbuf, sizeof(rsbuf), "reason(%lu)", r);
         rs = rsbuf;
@@ -649,6 +567,9 @@ const char *ERR_reason_error_string(unsigned long e)
     if (!RUN_ONCE(&err_string_init, do_err_strings_init)) {
         return NULL;
     }
+
+    if (ERR_SYSTEM_ERROR(e))
+        return NULL;
 
     l = ERR_GET_LIB(e);
     r = ERR_GET_REASON(e);
--- a/crypto/err/err.c
+++ b/crypto/err/err.c
@@ -187,7 +187,6 @@ static ERR_STRING_DATA *int_err_get_item(const ERR_STRING_DATA *d)
  * codes.
  */
 # define NUM_SYS_STR_REASONS 255
-static ERR_STRING_DATA SYS_str_reasons[NUM_SYS_STR_REASONS + 1];
 
 /*
  * Russian and Ukrainian locales on Linux required more than 6,5 kB for the
@@ -195,11 +194,12 @@ static ERR_STRING_DATA SYS_str_reasons[NUM_SYS_STR_REASONS + 1];
  * We defined the string space to be 8 KiB at the time.  Now that we double
  * the range, we double the string space.
  */
-# define SPACE_SYS_STR_REASONS 16 * 1024
+# define SPACE_SYS_STR_REASONS (16 * 1024)
 
 static void build_SYS_str_reasons(void)
 {
     /* OPENSSL_malloc cannot be used here, use static storage instead */
+    static ERR_STRING_DATA SYS_str_reasons[NUM_SYS_STR_REASONS + 1];
     static char strerror_pool[SPACE_SYS_STR_REASONS];
     char *cur = strerror_pool;
     size_t cnt = 0;
--- a/crypto/err/err.c
+++ b/crypto/err/err.c
@@ -174,20 +174,28 @@ static ERR_STRING_DATA *int_err_get_item(const ERR_STRING_DATA *d)
 }
 
 #ifndef OPENSSL_NO_ERR
-/* 2019-05-21: Russian and Ukrainian locales on Linux require more than 6,5 kB */
-# define SPACE_SYS_STR_REASONS 8 * 1024
-# define NUM_SYS_STR_REASONS 127
-
-static ERR_STRING_DATA SYS_str_reasons[NUM_SYS_STR_REASONS + 1];
 /*
- * SYS_str_reasons is filled with copies of strerror() results at
- * initialization. 'errno' values up to 127 should cover all usual errors,
- * others will be displayed numerically by ERR_error_string. It is crucial
- * that we have something for each reason code that occurs in
+ * SYS_str_reasons is filled with pointers to copies of strerror() results
+ * at initialization. Although POSIX permits 'errno' to have any positive
+ * number that fits in an 'int', most systems use a range starting at 1,
+ * and the highest number we have observed is well below 255, which is
+ * our current upper limit.  Other codes will be displayed numerically
+ * by ERR_error_string.
+ * It is crucial that we have something for each reason code that occurs in
  * ERR_str_reasons, or bogus reason strings will be returned for SYSerr(),
  * which always gets an errno value and never one of those 'standard' reason
  * codes.
  */
+# define NUM_SYS_STR_REASONS 255
+static ERR_STRING_DATA SYS_str_reasons[NUM_SYS_STR_REASONS + 1];
+
+/*
+ * Russian and Ukrainian locales on Linux required more than 6,5 kB for the
+ * errno range 1..127 (observed 2019-05-21).
+ * We defined the string space to be 8 KiB at the time.  Now that we double
+ * the range, we double the string space.
+ */
+# define SPACE_SYS_STR_REASONS 16 * 1024
 
 static void build_SYS_str_reasons(void)
 {
--- a/crypto/err/err.c
+++ b/crypto/err/err.c
@@ -504,6 +504,12 @@ void ERR_error_string_n(unsigned long e, char *buf, size_t len)
         ls = lsbuf;
     }
 
+    /*
+     * ERR_reason_error_string() can't safely return system error strings,
+     * since it would call openssl_strerror_r(), which needs a buffer for
+     * thread safety.  So for system errors, we call openssl_strerror_r()
+     * directly instead.
+     */
     r = ERR_GET_REASON(e);
     if (ERR_SYSTEM_ERROR(e)) {
         if (openssl_strerror_r(r, rsbuf, sizeof(rsbuf)))
@@ -568,6 +574,11 @@ const char *ERR_reason_error_string(unsigned long e)
         return NULL;
     }
 
+    /*
+     * ERR_reason_error_string() can't safely return system error strings,
+     * since openssl_strerror_r() needs a buffer for thread safety, and we
+     * haven't got one that would serve any sensible purpose.
+     */
     if (ERR_SYSTEM_ERROR(e))
         return NULL;
 
--- a/include/openssl/err.h
+++ b/include/openssl/err.h
@@ -195,51 +195,52 @@ struct err_state_st {
  */
 
 /* Macros to help decode recorded system errors */
-# define ERR_SYSTEM_FLAG        (1UL << (sizeof(unsigned int) * CHAR_BIT - 1))
-# define ERR_SYSTEM_MASK        (ERR_SYSTEM_FLAG - 1)
+# define ERR_SYSTEM_FLAG                ((unsigned int)INT_MAX + 1)
+# define ERR_SYSTEM_MASK                ((unsigned int)INT_MAX)
 
 /* Macros to help decode recorded OpenSSL errors */
-# define ERR_LIB_OFFSET         23L
-# define ERR_LIB_MASK           0xFF
-# define ERR_RFLAGS_OFFSET      19L
-# define ERR_RFLAGS_MASK        0xF
-# define ERR_REASON_MASK        0X7FFFFF
+# define ERR_LIB_OFFSET                 23L
+# define ERR_LIB_MASK                   0xFF
+# define ERR_RFLAGS_OFFSET              19L
+# define ERR_RFLAGS_MASK                0xF
+# define ERR_REASON_MASK                0X7FFFFF
 
 /*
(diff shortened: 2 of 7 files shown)


crypto/err/err.c
```diff
@@ -174,20 +174,28 @@ static ERR_STRING_DATA *int_err_get_item(const ERR_STRING_DATA *d)
...
+ * and the highest number we have observed is well below 255, which is
+ * our current upper limit.  Other codes will be displayed numerically
+ * by ERR_error_string.
+ * It is crucial that we have something for each reason code that occurs in
  * ERR_str_reasons, or bogus reason strings will be returned for SYSerr(),
  * which always gets an errno value and never one of those 'standard' reason
  * codes.
  */
+# define NUM_SYS_STR_REASONS 255
+static ERR_STRING_DATA SYS_str_reasons[NUM_SYS_STR_REASONS + 1];
```
richsalz, line 190: This can be static inside the function.

crypto/err/err.c
```diff
@@ -174,20 +174,28 @@ static ERR_STRING_DATA *int_err_get_item(const ERR_STRING_DATA *d)
...
+
+/*
+ * Russian and Ukrainian locales on Linux required more than 6,5 kB for the
+ * errno range 1..127 (observed 2019-05-21).
+ * We defined the string space to be 8 KiB at the time.  Now that we double
+ * the range, we double the string space.
+ */
+# define SPACE_SYS_STR_REASONS 16 * 1024
```
richsalz, line 198: probably should be parenthesized.

include/openssl/err.h
```diff
@@ -163,14 +163,43 @@ struct err_state_st {
...
+ * Where LL is the library code, r is the reason flags, and rRRRRR is the
+ * reason code.
+ * Do note that the reason flags is part of the reason code, and could as
+ * well be seen as a section of all possible reason codes.  We do this for
+ * backward compatibility reasons, i.e. how ERR_R_FATAL was implemented.
+ *
+ * System errors (ERR_LIB_SYS) are structured the same way, except they
+ * don't have any reason flag.
+ *
+ * LL RRRRRR
```
richsalz, line 180: so this is special encoding for a specific LL value; maybe put that value here?
And consider making ```ERR_LIB_SYS``` by define'd to zero.

test/errtest.c
```diff
@@ -36,25 +36,37 @@ static int test_print_error_format(void)
     char *out = NULL, *p = NULL;
     int ret = 0, len;
     BIO *bio = NULL;
+    int code = EPERM;
```
richsalz, line 39: Why have the code variable, just use EPERM throughout.

test/recipes/02-test_errstr.t
```diff
@@ -40,79 +40,52 @@ plan skip_all => 'This is unsupported on MSYS/MinGW or MSWin32'
...
+      };
+
+      # We know that the system reasons are in OpenSSL error library 2
+      my @oerr = run(app([ qw(openssl errstr), sprintf("2%06x", $errnum) ]),
+                     capture => 1);
+      $oerr[0] =~ s|\R$||;
+      @oerr = split_error($oerr[0]);
+      # Optional, if libcrypto didn't capture the text
+      my $rerr = "reason($errnum)";
+      ok($oerr[3] eq $perr || $oerr[3] eq $rerr,
```
richsalz, line 78: This would be much easier to read if it were a single if statement with two different ok calls.

crypto/err/err.c
levitte: Good point

crypto/err/err.c
levitte: Okie

include/openssl/err.h
levitte: Using zero is difficult, as that value is used internally to reach ERR_R_ reasons.  Happens here:

https://github.com/openssl/openssl/blob/94941cada25433a7dca35b5b9f8cbb751ab65ab3/crypto/err/err.c#L649-L652

So, using zero for the lib part in recorded system errors as well comes with a (very) likely risk of clashes, and misleading error text.

test/errtest.c
levitte: So someone changes EPERM to some other E-macro in one spot but forgets another...  Why can't one spot be enough?

test/recipes/02-test_errstr.t
levitte: It's a trade-off.  If you want to quickly check that you've calculated the plan right, it's easier to see that there's only one `ok` per code and not two if there's *actually* only one `ok` there.

include/openssl/err.h
levitte: Actually, I have other thoughts on system errors, see https://github.com/openssl/openssl/pull/12314#issuecomment-652801232

test/recipes/02-test_errstr.t
richsalz:
```
if (something) { ok ... }
else { ok ... }
```
seems pretty obvious to me.  The proposed code is far from obvious to me.

test/errtest.c
richsalz: If someone changes it in one place, the test will fail.  This level of indirection makes the code less clear. It cascades, because now you have an auxiliary "code2" variable.  code2 is not a clear name either.  And so on.

You have to at least make code "const int" no?

levitte:
The ideas mentioned in https://github.com/openssl/openssl/pull/12314#issuecomment-652801232 are now implemented here.  @DDvO, please have a look.

crypto/err/err.c
```diff
@@ -583,8 +504,13 @@ void ERR_error_string_n(unsigned long e, char *buf, size_t len)
         ls = lsbuf;
     }
 
-    rs = ERR_reason_error_string(e);
     r = ERR_GET_REASON(e);
+    if (ERR_SYSTEM_ERROR(e)) {
```
richsalz, line 508: A one-line comment would be helpful.

test/errtest.c
```diff
@@ -36,25 +36,40 @@ static int test_print_error_format(void)
...
+        TEST_skip("libcrypto didn't register EPERM (%d) correctly (%d).  "
+                  "This is a known limitation",
+                  syserr, reasoncode);
+        ERR_clear_error();
+    } else {
+        char buf[256];
+
+        OPENSSL_strlcpy(buf, strerror(syserr), sizeof(buf));
+        BIO_snprintf(expected, sizeof(expected), expected_format,
+                     OPENSSL_FUNC, buf);
```
DDvO, line 55: You don't really need the `buf`, simply taking `strerror(syserr)` as arg of `BIO_snprintf()`.

test/errtest.c
```diff
@@ -36,25 +36,40 @@ static int test_print_error_format(void)
...
+
+        if (!TEST_int_gt(len = BIO_get_mem_data(bio, &out), 0))
+            goto err;
+        /* Skip over the variable thread id at the start of the string */
+        for (p = out; *p != ':' && *p != 0; ++p) {
+            if (!TEST_true(IS_HEX(*p)))
+                goto err;
+        }
+        if (!TEST_true(*p != 0)
+            || !TEST_strn_eq(expected, p, strlen(expected)))
```
DDvO, line 70: I suggest pointing out in a comment that the string is compared only up to the reason string (and subsequent `:`).

test/recipes/02-test_errstr.t
```diff
@@ -40,84 +34,31 @@ plan skip_all => 'This is unsupported on MSYS/MinGW or MSWin32'
...
 }
 
-my @after = run(app([ qw(openssl errstr 2000080) ]), capture => 1);
-$after[0] =~ s|\R$||;
-@after = split_error($after[0]);
-ok($after[3] eq "reason(128)", "(128) '$after[3]' == 'reason(128)'");
+# OpenSSL library 1 is the "unknown" library
+&ok(match_opensslerr_reason(0x1 << 23 | 0x100, "reason(256)"));
+# Reason code 0 of any library gives the library name as reason
+&ok(match_opensslerr_reason(0x1 << 23 |  0x00, "unknown library"));
```
DDvO, line 59: Maybe use symbolic const names here for `0x01` and `23`?

test/recipes/02-test_errstr.t
```diff
@@ -132,3 +73,58 @@ sub split_error {
...
+        $desc = "match '$first' ($desc) with '$strings[0]'";
+    }
+
+    return ( scalar( grep { $first eq $_ } @strings ) > 0,
+             $desc );
+}
+
+sub match_opensslerr_reason {
+    my $errnum = shift;
+    my @strings = @_;
```
DDvO, line 100: This variable actually obtains `$errnum` (which is likely not intended, see comment below at the end of `match_syserr_reason`) and the two strings computed in the below function.

test/recipes/02-test_errstr.t
```diff
@@ -132,3 +73,58 @@ sub split_error {
...
+    } else {
+        $desc = "match '$first' ($desc) with '$strings[0]'";
+    }
+
+    return ( scalar( grep { $first eq $_ } @strings ) > 0,
+             $desc );
+}
+
+sub match_opensslerr_reason {
+    my $errnum = shift;
```
DDvO, line 99: Better name maybe: `$errcode`

test/recipes/02-test_errstr.t
```diff
@@ -132,3 +73,58 @@ sub split_error {
...
+          $!
+    };
+    # The OpenSSL fallback string
+    push @strings, "reason($errnum)";
+
+    # We know that the system reasons are recorded by OpenSSL as an 'int'
+    # with an added set high bit, which always is POSIX::INT_MAX + 1.
+    my $syserr_flag = INT_MAX + 1;
+
+    return match_opensslerr_reason($syserr_flag | $errnum, $errnum, @strings);
```
DDvO, line 129: I think the second arg, `$errnum, should be left out here.

test/errtest.c
```diff
@@ -36,25 +36,40 @@ static int test_print_error_format(void)
...
+    ERR_PUT_error(ERR_LIB_SYS, 0, syserr, "errtest.c", 30);
+    reasoncode = ERR_GET_REASON(ERR_peek_error());
 
-    if (!TEST_ptr(bio = BIO_new(BIO_s_mem())))
-        return 0;
+    if (reasoncode != syserr) {
+        TEST_skip("libcrypto didn't register EPERM (%d) correctly (%d).  "
+                  "This is a known limitation",
+                  syserr, reasoncode);
+        ERR_clear_error();
```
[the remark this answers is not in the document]
levitte: EPERM is only "raised" to check that `ERR_print_errors()` outputs what we expect it to, it's not meant to remain in the error queue.

However, I'm thinking that the check that the system error code came back correctly should be a proper test, not something that's being skipped...

include/openssl/err.h
```diff
@@ -163,14 +164,82 @@ struct err_state_st {
...
+          : (l) & ERR_RFLAG_FATAL)
+
+/*
+ * ERR_PACK is a helper macro to properly pack OpenSSL error codes and may
+ * only be used for that purpose.  System errors are packed internally.
+ */
+/* ERR_PACK takes reason flags and reason code combined in |r| */
+# define ERR_PACK(l,f,r)                                                \
+    ( (((unsigned int)(l) & ERR_LIB_MASK) << ERR_LIB_OFFSET) |          \
+      (((unsigned int)(r) & ERR_REASON_MASK)) )
```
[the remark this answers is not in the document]
levitte: I actually have no clue why `unsigned int` is used here, other than making it unsigned...  maybe someone thought that we should keep with some sort of minimum type that could fit 32 bits and let the compiler to whatever expansion it needed when needed.  I can certainly align this with the types used in `ERR_STATE and ERR_STRING_DATA...

test/errtest.c
levitte: Good point

crypto/err/err.c
levitte: Done...  although it became more than one line.

crypto/err/err.c
```diff
@@ -569,8 +490,8 @@ static unsigned long get_error_values(ERR_GET_ACTION g,
 
 void ERR_error_string_n(unsigned long e, char *buf, size_t len)
 {
-    char lsbuf[64], rsbuf[64];
-    const char *ls, *rs;
+    char lsbuf[64], rsbuf[256];
+    const char *ls, *rs = NULL;
     unsigned long f = 0, l, r;
```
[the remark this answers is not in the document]
levitte: If you don't mind, I prefer to make that a separate cleanup PR.  There's a *lot* of this in the ERR sub-system.

include/openssl/err.h
```diff
@@ -163,14 +164,82 @@ struct err_state_st {
...
+ *                   +-------+-------------------------------------+
+ *                   | rflags|                reason               |
+ *                   +-------+-------------------------------------+
+ *
+ * We have the reason flags being part of the overall reason code for
+ * backward compatibility reasons, i.e. how ERR_R_FATAL was implemented.
+ */
+
+/* Macros to help decode recorded system errors */
+# define ERR_SYSTEM_FLAG        (1UL << (sizeof(unsigned int) * CHAR_BIT - 1))
```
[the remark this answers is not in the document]
levitte: Note sure why one is less robust than the other (I distrust headers when I can get the same answer directly from the compiler)...  but sure, that's probably more readable.

include/openssl/err.h
```diff
@@ -163,14 +164,82 @@ struct err_state_st {
...
+# define ERR_SYSTEM_ERROR(l)    (((l) & ERR_SYSTEM_FLAG) != 0)
+
+# define ERR_GET_LIB(l)                                         \
+    (int)(ERR_SYSTEM_ERROR(l)                                   \
+          ? ERR_LIB_SYS                                         \
+          : ((l) >> ERR_LIB_OFFSET) & ERR_LIB_MASK)
+# define ERR_GET_FUNC(l)        0
+# define ERR_GET_RFLAGS(l)                                      \
+    (int)(ERR_SYSTEM_ERROR(l)                                   \
+          ? 0                                                   \
```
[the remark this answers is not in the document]
levitte: We have never done that...  I'd suggest leaving that decision for another PR

include/openssl/err.h
```diff
@@ -163,14 +164,82 @@ struct err_state_st {
...
+    (int)(ERR_SYSTEM_ERROR(l)                                   \
+          ? (l) & 0                                             \
+          : (l) & ERR_RFLAG_FATAL)
+
+/*
+ * ERR_PACK is a helper macro to properly pack OpenSSL error codes and may
+ * only be used for that purpose.  System errors are packed internally.
+ */
+/* ERR_PACK takes reason flags and reason code combined in |r| */
+# define ERR_PACK(l,f,r)                                                \
```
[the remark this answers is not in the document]
levitte: Done

include/openssl/err.h
```diff
@@ -163,14 +164,82 @@ struct err_state_st {
...
+ *                   | rflags|                reason               |
+ *                   +-------+-------------------------------------+
+ *
+ * We have the reason flags being part of the overall reason code for
+ * backward compatibility reasons, i.e. how ERR_R_FATAL was implemented.
+ */
+
+/* Macros to help decode recorded system errors */
+# define ERR_SYSTEM_FLAG        (1UL << (sizeof(unsigned int) * CHAR_BIT - 1))
+# define ERR_SYSTEM_MASK        (ERR_SYSTEM_FLAG - 1)
```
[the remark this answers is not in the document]
levitte: Done

include/openssl/err.h
```diff
@@ -163,14 +164,82 @@ struct err_state_st {
...
+ * number.
+ */
+# define ERR_RFLAG_FATAL        (0x1 << ERR_RFLAGS_OFFSET)
+
+# define ERR_SYSTEM_ERROR(l)    (((l) & ERR_SYSTEM_FLAG) != 0)
+
+# define ERR_GET_LIB(l)                                         \
+    (int)(ERR_SYSTEM_ERROR(l)                                   \
+          ? ERR_LIB_SYS                                         \
+          : ((l) >> ERR_LIB_OFFSET) & ERR_LIB_MASK)
```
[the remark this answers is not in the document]
levitte: Since these macros are completely rewritten, I decided to use longer names in this section specifically.

test/errtest.c
```diff
@@ -36,25 +36,40 @@ static int test_print_error_format(void)
     char *out = NULL, *p = NULL;
     int ret = 0, len;
     BIO *bio = NULL;
+    int syserr = EPERM;
+    int reasoncode;
 
-    BIO_snprintf(expected, sizeof(expected), expected_format, OPENSSL_FUNC);
+    ERR_PUT_error(ERR_LIB_SYS, 0, syserr, "errtest.c", 30);
```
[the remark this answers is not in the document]
levitte: I've moved code around to use those macros, and generally trying to clarify a little bit.

test/errtest.c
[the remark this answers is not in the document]
levitte: Done

include/openssl/err.h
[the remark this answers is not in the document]
levitte: Done

test/errtest.c
levitte: Done

test/errtest.c
levitte: Uhmmm, but that's not true

test/recipes/02-test_errstr.t
levitte: Sure

test/recipes/02-test_errstr.t
levitte: Yup, you're right.  That was a left-over from previous coding.  Gone

test/recipes/02-test_errstr.t
levitte: Okie

test/recipes/02-test_errstr.t
DDvO: At least three people think that that `ok(P || Q, P? ... : ...)` call is needlessly hard to read.
On the other hand, well, it's "just" part of a test script.

include/openssl/err.h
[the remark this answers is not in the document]
DDvO: Also here you need to trust a header since `CHAR_BIT` is defined in `<limits.h>` as well :stuck_out_tongue:
So let's better go for the more readable variant.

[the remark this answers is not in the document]
DDvO: Ah, now I see you already did the change :smile:

test/errtest.c
DDvO: This is only comparing those strings up to `strlen(expected)`, and the one in `p` is longer.

DDvO:
Travis failures (on tests) are relevant.

BTW, nice to see the improved (by #12279) semantics of `VF=1` at work also for the Travis CI :)

levitte:
Ewwww, that parallel testing output looks quite ugly.  Hardly legible

levitte:
Ah, I see why.  Some macros need to be converted to static inline functions to avoid multiple evaluation of the argument

levitte:
Fixed.  i think

DDvO:
> Ewwww, that parallel testing output looks quite ugly. Hardly legible

This is to be expected on arbitrarily interleaved output of parallel runs.
We should be able to improve on this by buffering all per-test output and then printing it atomically.
Looking at the current state of `run-tests.pl`, this should already be mostly so for for `VF=1` and `VFP=1`,
while for `V=1`, which is not explicitly printed there, I did not check.

My comment above comment was referring to a different aspect: verbose output on failed test cases only.

DDvO:
> Ah, I see why. Some macros need to be converted to static inline functions to avoid multiple evaluation of the argument

Just noticed this - I can have a look at the improved version later ...

levitte:
Merged

163b2bcd8b2e5cd149dfc8dce1ca096805559379 ERR: refactor global error codes
71f2994b151f5de0c7bc14592c84795ff98256c1 ERR: special case system errors
17b7f8968481aa99c622080ac73879f42fb8c4ae TEST: fix test/errtest.c
fa7a807435edbcdcddf809398c59a60650315981 SSL: fix misuse of ERR_LIB_SYS
1b726e9b91a032298dc96ad117b23e18e1583246 TEST: update 02-test_errstr.t to have better tests

PR #12359, opened 2020-07-02
test/run_tests.pl: Do slowest tests first for parallel runs

DDvO:
This a follow-up on @romen's grand improvement on test execution time in #12326.
When being able to run four tests in parallel at a time, that reduces test run time to some 44% (compared to the 100% before).

It turns out that in our test suite there are two individual tests that take *much* longer than the others:
`80-test_ssl_new` and `99-test_fuzz`. According to my experiments, *each* of them takes some 25% of the total time!

Since these have rather high ordinals, they are started rather late, which is on purpose for sequential runs.
However, when running all tests in parallel with `HARNESS_JOBS` being `4` or some other number that is much less than the number of tests, this leads to the undesirable effect that those two tests are started rather late and continue to run way beyond the finish time of the other (much faster) tests, which considerably reduces the actual gain through parallelism.

This PR makes sure that - as far as any of those slow tests is selected to run - these slow tests are started first,
such that effectively all the time they run in parallel to the other, much faster tests.
In particular, when 4 CPU cores are available to run the tests, *for most of the total test execution time*
* one core runs `ssl_test` (aka `80-test_ssl_new`), 
* another one runs `asn1-test` or other parts of `99-test_fuzz` such as `client-test` and `server-test`, 
* while the two remaining cores run all the other tests.

According to what I just measured, 
with `HARNESS_JOBS=4` the gain in overall execution time is some further 18% (compared to not re-ordering the tests).
In other words, compared to sequential execution, the re-ordering reduces the execution time not only to some 44% 
but to some 36% (taking the time for executing all tests sequentially as 100%).

Consequently, with this further enhancement of #12326, running all tests now typically takes only around a third of the time,
compared with the time the test suite took when running all tests sequentially.


richsalz:
Another possibility is to split test-fuzz into parts like fuzz-odd and fuzz-even which run the same program but pick odd or even tests.

DDvO:
> Another possibility is to split test-fuzz into parts like fuzz-odd and fuzz-even which run the same program but pick odd or even tests.

Splitting the monolithic 25% of the total run time consumed by `test_fuzz` into 2 chunks of 12,5% each would not give a large benefit
because, due to their ordinal being 99, those two chunks would still be started after all other tests have been started
and because the `test_ssl_new` would still be a monolith and commence much later than all other tests,
still causing a major loss of effective parallelism.
So rather than gaining some further 18% due to the enhancement proposed here, I estimate the gain would be around 6%.

richsalz:
Sorry, I didn't mean "only split"  I meant "also split"

DDvO:
> Sorry, I didn't mean "only split" I meant "also split"

So, you mean as a further enhancement on top of the other two.
Can you suggest a simple way of implementing that, at least for quick experimental purposes?

richsalz:
>Can you suggest a simple way of implementing that, at least for quick experimental purposes?

See attached.

[patch.txt](https://github.com/openssl/openssl/files/4866193/patch.txt)

test/run_tests.pl
```diff
@@ -65,6 +65,16 @@ BEGIN
 my @alltests = find_matching_tests("*");
 my %tests = ();
 
+sub reorder {
+    my $key = $_;
+
+    # for parallel test runs, do slow tests first
+    if ($jobs >= 1 && $key =~ m/test_ssl_new|test_fuzz/) {
```
romen, line 72: `if $jobs > 1`, right?

DDvO: > `if $jobs > 1`, right?

Oops, yes.
For some experiments I had changed `>` to `>=` and forgot to restore that. Done.

DDvO:
> > Can you suggest a simple way of implementing that, at least for quick experimental purposes?
> 
> See attached.
> 
> [patch.txt](https://github.com/openssl/openssl/files/4866193/patch.txt)

Thanks @richsalz for the suggestion how to split the fuzz tests.
I did various experiments using - and further adapting - it.

It turns out that also the fuzz tests have a very uneven run time distribution:
Each of `asn1`, `client`, and `server` take some 30% of the total fuzzing time,
while all the others take almost negligible time.

I first did the suggested split into "even" and "odd", 
then also tried out splitting into asn1, client, server, and the sequence all others, 
mostly using `HARNESS_JOBS=4` according to my machine having a quad-core CPU.
There was no real gain doing so, even when I additionally prioritized the three CPU-intense fuzz tests,
which just had the effect that initially all four cores had 100% load for a while.

So in the end I see no advantage in splitting up the fuzz tests,
but at least the suggestion was a good motivation and basis for cleaning up `99-test_fuzz.t` and 
re-structuring it a little such that one could easily do any splitting of the fuzz tests if desired.

DDvO:
I'd say the way of parallelizing the tests as implemented in this PR (based on #12326) should be rather optimal for use with 2, 3, or 4 cores because 
* the two by far most long running tests `test_ssl_new` and `test_fuzz` get started first and will keep 2 cores pretty evenly busy for long time, and
* using 2 cores, all other tests are run thereafter with pretty even distribution
* using 3 cores, all other tests are run in 1 parallel thread next to `test_ssl_new` and `test_fuzz` for long time,
and then until the end of the overall test run the remaining tests run in 3 parallel threads with pretty even distribution.
* using 4 cores, all other tests are run in 2 parallel threads next to `test_ssl_new` and `test_fuzz` for most of the time,
and towards the end of the overall test run the rather small number of remaining tests run in 4 parallel threads with pretty even distribution.

Of course, the changes take only effect if `HARNESS_JOBS` is set to a value > 1, 
i.e., they do not change anything for sequential runs.

test/run_tests.pl
```diff
@@ -82,15 +92,15 @@ BEGIN
     if ($arg eq 'alltests') {
 	warn "'alltests' encountered, ignoring everything before that...\n"
 	    unless $initial_arg;
-	%tests = map { $_ => basename($_) } @alltests;
+	%tests = map { reorder($_) => $_ } @alltests;
```
levitte, line 95: This isn't right. Look for `sort keys %tests`, change it to `sort { ... } keys %tests`, where '...' is your sorting code.

DDvO: No, it does exactly what I meant it to do:
`map { reorder($_) => $_ } @alltests` transforms an array of test script names into a hash table
where each element associates with the string `reorder($_)` the original name `$_`.

The `=>` may look confusing here because it does not mean a lamba abstraction that typically may be used to define a function on-the-fly.

Note that the `reorder` subroutine just 'renames' some of the test scritps, and together with the construct explained above it effectively re-orders the names where needed.

I had already considered using the optional `{ ... }` argument with `sort`, but this has a different semantics: compare two values and yield -1, 0, or +1.

levitte: I have no idea why I'd interpret the list pair operator (that's really what `=>` is) as a lambda abstraction.  Mixed up languages, perhaps?

I don't quite understand your retort,  The "reorder" function simply rewrites the keys to get a different *sorting order*, so why that would be impossible to use *when sorting*, I do not understand.  The following (diff from master) should give the exact same results.

``` diff
diff --git a/test/run_tests.pl b/test/run_tests.pl
index d47f7cf1e6..864bc7a5b4 100644
--- a/test/run_tests.pl
+++ b/test/run_tests.pl
@@ -70,6 +70,16 @@ open $openssl_args{'tap_copy'}, ">$outfilename"
 my @alltests = find_matching_tests("*");
 my %tests = ();
 
+sub reorder {
+    my $key = pop;
+
+    # for parallel test runs, do slow tests first
+    if (defined $jobs && $jobs > 1 && $key =~ m/test_ssl_new|test_fuzz/) {
+        $key =~ s/(\d+)-/00-/;
+    }
+    return $key;
+}
+
 my $initial_arg = 1;
 foreach my $arg (@ARGV ? @ARGV : ('alltests')) {
     if ($arg eq 'list') {
@@ -281,7 +291,7 @@ unless (defined $eres) {
 my $harness = $package->new(\%tapargs);
 my $ret =
     $harness->runtests(map { [ abs2rel($_, rel2abs(curdir())), $tests{$_} ] }
-                       sort keys %tests);
+                       sort { reorder($a) cmp reorder($b) } keys %tests);
 
 # $ret->has_errors may be any number, not just 0 or 1.  On VMS, numbers
 # from 2 and on are used as is as VMS statuses, which has severity encoded
```

DDvO: > I have no idea why I'd interpret the list pair operator (that's really what `=>` is) as a lambda abstraction. Mixed up languages, perhaps?

At least to me the `=>` is confusing, from the background of functional languages or the lambda calculus underlying them.

> I don't quite understand your retort, The `reorder` function simply rewrites the keys to get a different _sorting order_, 

That's correct. My "no" was just on your strict "This isn't right.", which I don't agree on.

> so why that would be impossible to use _when sorting_, I do not understand.

I did not see that it was possible also via the `{ ...}` argument to `sort`.

> The following (diff from master) should give the exact same results.
> ```
>      $harness->runtests(map { [ abs2rel($_, rel2abs(curdir())), $tests{$_} ] }
> -                       sort keys %tests);
> +                       sort { reorder($a) cmp reorder($b) } keys %tests);
>  
>  # $ret->has_errors may be any number, not just 0 or 1.  On VMS, numbers
>  # from 2 and on are used as is as VMS statuses, which has severity encoded
> ```

You are right, this works as well - thanks for pointing this out.
And it's likely easier to understand, so I've taken over your solution.

Still I've kept moving the hitherto three calls of `basename()` to one place,
namely into the `map { ... }` argument just above calling `sort`.
This is a simplification of the original code (and is independent of the re-ordering).

levitte: > And it's likely easier to understand

That is precisely the reason why I suggested this.  I'm sorry that it came out harsher.

> Still I've kept moving the hitherto three calls of `basename()` to one place,

I suspected you would.  That's fine :wink:

> At least to me the `=>` is confusing, from the background of functional languages or the lambda calculus underlying them.

I know, I've seen it in Standard ML and in Javascript.  That's why I asked if you got languages mixed up.

But yeah, in perl, it's just syntactic sugar for the comma operator in a list context.

Side note: Out of interest, I had a look through https://en.wikipedia.org/wiki/Anonymous_function, and it turns that `->` is more common than `=>` for a lambda constructor (even in C++11, which is *terribly* confusing)

test/run_tests.pl
```diff
@@ -82,15 +92,15 @@ BEGIN
     if ($arg eq 'alltests') {
 	warn "'alltests' encountered, ignoring everything before that...\n"
 	    unless $initial_arg;
-	%tests = map { $_ => basename($_) } @alltests;
+	%tests = map { $_ => $_ } @alltests;
     } elsif ($arg =~ m/^(-?)(.*)/) {
```
levitte, line 96: There's something wrong with the indentation here...  could it be that we let TABs sneak in?  Would you mind untabifying?

test/run_tests.pl
```diff
@@ -280,8 +290,8 @@ sub find_matching_tests {
 
 my $harness = $package->new(\%tapargs);
 my $ret =
-    $harness->runtests(map { [ abs2rel($_, rel2abs(curdir())), $tests{$_} ] }
-                       sort keys %tests);
+    $harness->runtests(map { [ abs2rel($_, rel2abs(curdir())), basename($tests{$_}) ] }
+                       sort { reorder($a) cmp reorder($b) } keys %tests);
```
levitte, line 294: You *could* simplify further by changing things further:

-   all `%tests = map { $_ => $_ } @alltests;` to `%tests = map { $_ => 1 } @alltests;`
-   all `$tests{$test} = $test;` to `$tests{$test} = 1;`
-   these lines to:
    ``` perl
        $harness->runtests(map { [ abs2rel($_, rel2abs(curdir())), basename($_) ] }
                           sort { reorder($a) cmp reorder($b) } keys %tests);
    ```

The only reason to still have a `%tests` is that it guarantees uniqueness, and the `delete` line is much clearer than grepping stuff away from a list.

test/run_tests.pl
DDvO: Thanks @levitte for clarifying!

test/run_tests.pl
DDvO: No problem, done.

util/perl/OpenSSL/Test.pm
```diff
@@ -330,6 +330,25 @@ sub app {
...
+            plan skip_all => "No directory fuzz/corpora/$f" unless @dir;
+            plan tests => scalar @dir; # likely 1
+
+            foreach (@dir) {
+                ok(run(fuzz(["$f-test", $_])));
+            }
+        }
+    }
+}
+
```
levitte, line 351: This function is not locate where I would expect.  I frankly don't see any viable reason why it shouldn't stay in `99-test_fuzz.t`, but if you must, I would much rather see it in `util/perl/OpenSSL/Test/Utils.pm`

test/run_tests.pl
DDvO: Interesting simplification - done.

util/perl/OpenSSL/Test.pm
DDvO: The motivation to have `fuzz_tests` in a lib was that this way fuzzers could be partitioned easily,
splitting the calls of the resulting partitions of sub-tests into several `recipes/*.t` files.
So I've just moved it to the other lib source file, `Test/Utils.pm`.

DDvO:
I've squashed and rebased this PR.

DDvO:
This two Travis issues are unrelated.
@levitte, can this be approved now?

util/perl/OpenSSL/Test.pm
levitte: Sorry for being picky, but I'm rethinking this...  it doesn't feel like such a generic utility, but rather specialized.  So following what we do with `tconversion`, let me suggest that this function end up in, say, `test/recipes/fuzz.pl`...  don't forget to end that module with a `1;` (yes, after everything!), and then have this in the appropriate recipes:

``` perl
require_ok(srctop_file('test','recipes','fuzz.pl'));
```

levitte: `require_ok` counts as a test in itself, so remember to increment the plan by 1

DDvO: All right, done.
On this occasion I also considered moving the fuzz-related Perl code from `Test.pm` to the new `fuzz.pl`,
but it looks like that code is strongly related to the rest of `Test.pm`.

DDvO:
Merged - thanks @levitte

PR #12360, opened 2020-07-02
Configuration and build:  Fix solaris tags

### LINKED ISSUE #12356: 3.0.0-alpha4: Configurations name mismatch for solaris
File Configurations/10-main.conf refers to

shared_target    => "solaris-gcc"

but in file Configurations/shared-info.pl the entry is named "solaris-gcc-shared". When running the build, all settings from solaris-gcc-shared are not applied, eg. no symbolic linking is done and the SONAME of the created libs is mising the version suffix.

A naive fix of using "solaris-gcc-shared" instead of "solaris-gcc" in Configurations/10-main.conf leads to the next problem: in the generated Makefile the script mkdef.pl gets called with "--OS solaris-gcc", but mkdef.pl only knows about "solaris". So I added an alias line

"solaris-gcc" => "solaris"

to util/mkdef.pl.

It is possible, that other platforms will have similar problems. The following additional names occur in Configurations/10-main.conf as shared_target, without a definition in Configurations/shared-info.pl:

- aix
- self (irix-common, hpux-common)
- vms-shared
- win-shared

Regards,

Rainer

levitte:
The shared_target attrribute for Solaris built with gcc wasn't right
and shared libraries couldn't be properly built.

Fixes #12356

### DIFF
--- a/Configurations/10-main.conf
+++ b/Configurations/10-main.conf
@@ -225,7 +225,7 @@ my %targets = (
     "solaris-common-gcc" => {
         inherit_from     => [ "solaris-common" ],
         template         => 1,
-        shared_target    => "solaris-gcc", # The rest is on shared_info.pl
+        shared_target    => "solaris-gcc-shared", # The rest is on shared_info.pl
     },
 #### Solaris x86 with GNU C setups
     "solaris-x86-gcc" => {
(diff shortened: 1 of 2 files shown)


slontis:
@rainerjung Can you confirm that this addresses your iusse please.

rainerjung:
Confirmed, looks good to me!

levitte:
Merged

3a19f1a9dd16f5855a7e2a42cd3e7f136f849bd3 Configuration and build:  Fix solaris tags

PR #12362, opened 2020-07-03
[WIP][bp-1.1.1] Run tests in parallel

romen:
This is a tentative backport of #12326 to `1.1.1`.

~~~
Run tests in parallel

The environment variable `HARNESS_JOBS` can be used to control how many
jobs to run in parallel.  The default is still to run jobs sequentially.

This commit does not define custom `rules`, and different versions of
`TAP::Harness` come with different strategies regarding the default
`rules` that define which test recipes can be run in parallel.
In recent versions of Perl, unless specified otherwise, any task can be
run in parallel.

(cherry picked from commit a20c9075d6a72f484d1f27d99a54483a7e96fc51)
~~~

~~~
Travis: default to HARNESS_JOBS=4

We can run tests in parallel by setting the HARNESS_JOBS environment
variable.

(cherry picked from commit af3e8c298ab9fcee47d2d4a54f8be084990f9382)
~~~

### WIP

Notice that while in `master` each test job gets its own work directory in `/test-runs/<test_name>`, in `1.1.1` at the moment all test jobs get executed inside the same work directory `/test/test-runs/`.
This means that running test jobs in parallel can lead to conflicts, and that there might be hidden dependencies between tests so that the order in which they are run matters.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -218,6 +218,7 @@ before_script:
           fi;
       fi
     - ./configdata.pm --dump
+    - export HARNESS_JOBS=${HARNESS_JOBS:-4}
     - cd $top
 
 script:
--- a/INSTALL
+++ b/INSTALL
@@ -1206,6 +1206,25 @@
 
  $ make TESTS='[89]? -90'
 
+ Running Tests in Parallel
+ -------------------------
+
+ By default the test harness will execute the selected tests sequentially.
+ Depending on the platform characteristics, running more than one test job in
+ parallel may speed up test execution.
+ This can be requested by setting the `HARNESS_JOBS` environment variable to a
+ positive integer value. This specifies the maximum number of test jobs to run
+ in parallel.
+
+ Depending on the Perl version different strategies could be adopted to select
+ which test recipes can be run in parallel.  In recent versions of Perl, unless
+ specified otherwise, any task can be run in parallel. Consult the documentation
+ for `TAP::Harness` to know more.
+
+ To run up to four tests in parallel at any given time:
+
+ $ make HARNESS_JOBS=4 test
+
  Note on multi-threading
  -----------------------
 
--- a/test/run_tests.pl
+++ b/test/run_tests.pl
@@ -27,6 +27,7 @@ my $srctop = $ENV{SRCTOP} || $ENV{TOP};
 my $bldtop = $ENV{BLDTOP} || $ENV{TOP};
 my $recipesdir = catdir($srctop, "test", "recipes");
 my $libdir = rel2abs(catdir($srctop, "util", "perl"));
+my $jobs = $ENV{HARNESS_JOBS};
 
 $ENV{OPENSSL_CONF} = catdir($srctop, "apps", "openssl.cnf");
 
@@ -36,6 +37,8 @@ my %tapargs =
       switches  => '-w',
       merge     => 1
     );
+
+$tapargs{jobs} = $jobs if defined $jobs;
 
 my @alltests = find_matching_tests("*");
 my %tests = ();


romen:
@levitte can you suggest the least invasive way of making sure tests can't conflict with each other?

romen:
@DDvO I don't feel confident about merging this into `1.1.1` without solving the problem of potential conflicts!

DDvO:
> @DDvO I don't feel confident about merging this into `1.1.1` without solving the problem of potential conflicts!

I understand. 

I did not look myself - are there tmp/output files you know to overlap?

romen:
> > @DDvO I don't feel confident about merging this into `1.1.1` without solving the problem of potential conflicts!
> 
> I understand.
> 
> I did not look myself - are there tmp/output files you know to overlap?

I don't have too much time to check this in depth now but, assuming we have a similar set of tests in `1.1.1` and `master`, I see that in `master` I get the following duplicates that mark potential conflicts in `1.1.1`:

~~~bash
; for f in $(find test-runs -type f ); do echo "$f\t\t$(basename $f)"; done | sort -k2 | uniq -f1 -D                                                                                                 
test-runs/test_ca/demoCA/cacert.pem             cacert.pem
test-runs/test_store/store_21018/cacert.pem             cacert.pem
test-runs/test_ssl/certCA.srl           certCA.srl
test-runs/test_verify_store/certCA.srl          certCA.srl
test-runs/test_ssl/certCA.ss            certCA.ss
test-runs/test_verify_store/certCA.ss           certCA.ss
test-runs/test_ssl/certU.ss             certU.ss
test-runs/test_verify_store/certU.ss            certU.ss
test-runs/test_ssl/keyCA.ss             keyCA.ss
test-runs/test_verify_store/keyCA.ss            keyCA.ss
test-runs/test_ssl/keyU.ss              keyU.ss
test-runs/test_verify_store/keyU.ss             keyU.ss
test-runs/test_ec/pkey-f.d              pkey-f.d
test-runs/test_rsa/pkey-f.d             pkey-f.d
test-runs/test_ec/pkey-ff.dd            pkey-ff.dd
test-runs/test_rsa/pkey-ff.dd           pkey-ff.dd
test-runs/test_ec/pkey-ff.dp            pkey-ff.dp
test-runs/test_rsa/pkey-ff.dp           pkey-ff.dp
test-runs/test_ec/pkey-fff.p            pkey-fff.p
test-runs/test_rsa/pkey-fff.p           pkey-fff.p
test-runs/test_ec/pkey-ff.pd            pkey-ff.pd
test-runs/test_rsa/pkey-ff.pd           pkey-ff.pd
test-runs/test_ec/pkey-ff.pp            pkey-ff.pp
test-runs/test_rsa/pkey-ff.pp           pkey-ff.pp
test-runs/test_ec/pkey-f.p              pkey-f.p
test-runs/test_rsa/pkey-f.p             pkey-f.p
test-runs/test_ssl/req2CA.ss            req2CA.ss
test-runs/test_verify_store/req2CA.ss           req2CA.ss
test-runs/test_ssl/reqCA.ss             reqCA.ss
test-runs/test_verify_store/reqCA.ss            reqCA.ss
test-runs/test_ssl/reqU.ss              reqU.ss
test-runs/test_verify_store/reqU.ss             reqU.ss
test-runs/test_dgst/testdsa.sig         testdsa.sig
test-runs/test_pkeyutl/testdsa.sig              testdsa.sig
test-runs/test_dgst/testec-p256.sig             testec-p256.sig
test-runs/test_pkeyutl/testec-p256.sig          testec-p256.sig
test-runs/test_dgst/testrsa.sig         testrsa.sig
test-runs/test_pkeyutl/testrsa.sig              testrsa.sig
~~~

Also, even if we didn't have duplicates right now in `1.1.1`, there is a chance of creating duplicates in the future!

And I think the worst thing that can happen is that test results become unreliable, as failure/success/false positives/false negatives might be triggered simply by the order of execution of concurrent test jobs!

DDvO:
> The list above is probably partial, as tests might be cleaning up after themselves, with potential conflicts during the execution with temporary files, and such conflicts wouldn't be visible doing the check I ran above!

We could get more clarity as follows:
While running all tests, let each file open-for-write operation (at libcrypto and Perl level) generate a per-test log entry and then check the log for overlaps.

> Also, even if we didn't have duplicates right now in 1.1.1, there is a chance of creating duplicates in the future!

> And I think the worst thing that can happen is that test results become unreliable, as failure/success/false positives/false negatives might be triggered simply by the order of execution of concurrent test jobs!

Well, if those parallel runs were the only ones, I would see a risk, too, 
but not really if the tests get run sequentially at least from time to time.

The simplest way out should be to support parallel runs but not to make this the default for CI runs, neither.
Still when developing locally backports to v1.1.1, we could take advantage of quick test runs without actual risk.

romen:
@levitte is there an easy way to make it so that also in 1.1.1 each test runs in its own folder?

romen:
> @levitte is there an easy way to make it so that also in 1.1.1 each test runs in its own folder?

ping @levitte 

Somehow I believe that achieving that shouldn't be too hard, and it's just a question of summoning the Perl wizard!! ;)

t8m:
I do not think this is really worth it. The changes on 1.1.1 should be pretty minimal so there is no need to run the tests as quickly as possible. This risks having intermittent test failures due to test conflicts.

t8m:
Closing as this is not acceptable to 1.1.1

PR #12089, opened 2020-06-08
Better words

richsalz:
Remove some offensive/archaic terminology from OpenSSL.

PR #4402, opened 2017-09-22
Add master DRBG for reseeding

mspncp:
##### Checklist

- [x] documentation is added or updated  (as comments)
- [x] tests are added or updated

# The master DRBG

Up to now, there was an asymmetry between the public and the private global DRBG. Only the former was reseeded by the `RAND_add()` call. This pull request introduces a master DRBG which is not publicly accessible. So there are now three global DRBGs:

```
 <os entropy sources>
        |
     <master>          \
      /   \            | three global DRBGs with locking
<public>  <private>    /
    |
  <ssl>
```
The master DRBG is the only DRBG that pulls entropy from the trusted entropy sources, and that is reseeded by the `RAND_add()` call. Three of the DRBGs, namely master, public and private DRBG, have a locking mechanism, while the DRBGs chained to the public DRBG are intended to be used by a single thread and have no locking.

## Naming

In order to get a consistent naming of the three global DRBGs, the two existing DRBGs had to be renamed. This includes the `RAND_DRBG_get0_*()` method which are exported from libcrypto to libssl, but that's no problem because these functions were introduced in master only recently and are not part of an official API yet.

### Module globals

```C
static RAND_DRBG drbg_master;  /* The master DRBG. Seeded from the os */
static RAND_DRBG drbg_public;  /* The global public DRBG. */
static RAND_DRBG drbg_private; /* The global private-key DRBG. */
``` 

### Accessors

```C
RAND_DRBG *RAND_DRBG_get0_master(void);
RAND_DRBG *RAND_DRBG_get0_public(void);
RAND_DRBG *RAND_DRBG_get0_private(void);
```

## Seed Propagation

Instead of having `RAND_add()` reseed all global DRBGs (which would require taking all three locks), we only reseed the master DRBG (because it holds the lock already) and use a simple mechanism to propagate the reseeding to to all chained DRBGs: Every DRBG has a seed count which is incremented whenever the DRBG pulls fresh random input. Any chained DRBG compares in the `RAND_DRBG_generate()` call its seed count with the one of its parent. If there is a mismatch, it seeds itself automatically.

This propagation works for regular reseedings, too, not only for the ones triggered by `RAND_add()`.

**Update:** _As suggested by @paulidale, seed propagation will not proceed to the DRBGs which are owned by a thread or an SSL context, because this would make `RAND_add()` a very expensive operation. Only the global public and private DRBGs will be automatically restarted._  

[...]

PR #12366, opened 2020-07-04
Improving DRBG taxonomy

paulidale:
Since #12089 failed to get it in totality, this is an attempt to get one part of it in.

The term _master_ is part of the public API, it cannot easily be removed or changed.

### DIFF
--- a/crypto/rand/drbg_lib.c
+++ b/crypto/rand/drbg_lib.c
@@ -43,6 +43,8 @@ typedef struct drbg_global_st {
      *
      * There are three shared DRBG instances: <primary>, <public>, and
      * <private>.  The <public> and <private> DRBGs are secondary ones.
+     * These are used for non-secret (e.g. nonces) and secret
+     * (e.g. private keys) data respectively.
      */
     CRYPTO_RWLOCK *lock;
 
@@ -735,7 +737,7 @@ int RAND_DRBG_set_reseed_time_interval(RAND_DRBG *drbg, time_t interval)
  * Set the default values for reseed (time) intervals of new DRBG instances
  *
  * The default values can be set independently for primary DRBG instances
- * (without a parent) and normal DRBG instances (with parent).
+ * (without a parent) and secondary DRBG instances (with parent).
  *
  * Returns 1 on success, 0 on failure.
  */
@@ -849,7 +851,7 @@ static int drbg_bytes(unsigned char *out, int count)
 /* Implements the default OpenSSL RAND_add() method */
 static int drbg_add(const void *buf, int num, double randomness)
 {
-    RAND_DRBG *drbg = RAND_DRBG_get0_primary();
+    RAND_DRBG *drbg = RAND_DRBG_get0_master();
 
     if (drbg == NULL || num <= 0)
         return 0;
@@ -867,7 +869,7 @@ static int drbg_seed(const void *buf, int num)
 static int drbg_status(void)
 {
     int ret;
-    RAND_DRBG *drbg = RAND_DRBG_get0_primary();
+    RAND_DRBG *drbg = RAND_DRBG_get0_master();
 
     if (drbg == NULL)
         return 0;
@@ -903,7 +905,7 @@ RAND_DRBG *OPENSSL_CTX_get0_primary_drbg(OPENSSL_CTX *ctx)
     return dgbl->primary_drbg;
 }
 
-RAND_DRBG *RAND_DRBG_get0_primary(void)
+RAND_DRBG *RAND_DRBG_get0_master(void)
 {
     return OPENSSL_CTX_get0_primary_drbg(NULL);
 }
--- a/crypto/rand/drbg_lib.c
+++ b/crypto/rand/drbg_lib.c
@@ -41,29 +41,30 @@ typedef struct drbg_global_st {
     /*
      * The three shared DRBG instances
      *
-     * There are three shared DRBG instances: <master>, <public>, and <private>.
+     * There are three shared DRBG instances: <primary>, <public>, and
+     * <private>.  The <public> and <private> DRBGs are secondary ones.
      */
     CRYPTO_RWLOCK *lock;
 
     /*
-     * The <master> DRBG
+     * The <primary> DRBG
      *
      * Not used directly by the application, only for reseeding the two other
      * DRBGs. It reseeds itself by pulling either randomness from os entropy
      * sources or by consuming randomness which was added by RAND_add().
      *
-     * The <master> DRBG is a global instance which is accessed concurrently by
+     * The <primary> DRBG is a global instance which is accessed concurrently by
      * all threads. The necessary locking is managed automatically by its child
      * DRBG instances during reseeding.
      */
-    RAND_DRBG *master_drbg;
+    RAND_DRBG *primary_drbg;
     /*
      * The <public> DRBG
      *
      * Used by default for generating random bytes using RAND_bytes().
      *
-     * The <public> DRBG is thread-local, i.e., there is one instance per
-     * thread.
+     * The <public> secondary DRBG is thread-local, i.e., there is one instance
+     * per thread.
      */
     CRYPTO_THREAD_LOCAL public_drbg;
     /*
@@ -71,43 +72,44 @@ typedef struct drbg_global_st {
      *
      * Used by default for generating private keys using RAND_priv_bytes()
      *
-     * The <private> DRBG is thread-local, i.e., there is one instance per
-     * thread.
+     * The <private> secondary DRBG is thread-local, i.e., there is one
+     * instance per thread.
      */
     CRYPTO_THREAD_LOCAL private_drbg;
 } DRBG_GLOBAL;
 
 #define RAND_DRBG_TYPE_FLAGS    ( \
-    RAND_DRBG_FLAG_MASTER | RAND_DRBG_FLAG_PUBLIC | RAND_DRBG_FLAG_PRIVATE )
+    RAND_DRBG_FLAG_PRIMARY | RAND_DRBG_FLAG_PUBLIC | RAND_DRBG_FLAG_PRIVATE )
 
-#define RAND_DRBG_TYPE_MASTER                     0
+#define RAND_DRBG_TYPE_PRIMARY                    0
 #define RAND_DRBG_TYPE_PUBLIC                     1
 #define RAND_DRBG_TYPE_PRIVATE                    2
 
 /* Defaults */
 static int rand_drbg_type[3] = {
-    RAND_DRBG_TYPE, /* Master */
+    RAND_DRBG_TYPE, /* Primary */
     RAND_DRBG_TYPE, /* Public */
     RAND_DRBG_TYPE  /* Private */
 };
 static unsigned int rand_drbg_flags[3] = {
-    RAND_DRBG_FLAGS | RAND_DRBG_FLAG_MASTER, /* Master */
-    RAND_DRBG_FLAGS | RAND_DRBG_FLAG_PUBLIC, /* Public */
-    RAND_DRBG_FLAGS | RAND_DRBG_FLAG_PRIVATE /* Private */
+    RAND_DRBG_FLAGS | RAND_DRBG_FLAG_PRIMARY, /* Primary */
+    RAND_DRBG_FLAGS | RAND_DRBG_FLAG_PUBLIC,  /* Public */
+    RAND_DRBG_FLAGS | RAND_DRBG_FLAG_PRIVATE  /* Private */
 };
 
-static unsigned int master_reseed_interval = MASTER_RESEED_INTERVAL;
-static unsigned int slave_reseed_interval  = SLAVE_RESEED_INTERVAL;
+static unsigned int primary_reseed_interval = PRIMARY_RESEED_INTERVAL;
+static unsigned int secondary_reseed_interval = SECONDARY_RESEED_INTERVAL;
 
-static time_t master_reseed_time_interval = MASTER_RESEED_TIME_INTERVAL;
-static time_t slave_reseed_time_interval  = SLAVE_RESEED_TIME_INTERVAL;
+static time_t primary_reseed_time_interval = PRIMARY_RESEED_TIME_INTERVAL;
+static time_t secondary_reseed_time_interval = SECONDARY_RESEED_TIME_INTERVAL;
 
 /* A logical OR of all used DRBG flag bits (currently there is only one) */
 static const unsigned int rand_drbg_used_flags =
     RAND_DRBG_FLAG_CTR_NO_DF | RAND_DRBG_FLAG_HMAC | RAND_DRBG_TYPE_FLAGS;
 
 
-static RAND_DRBG *drbg_setup(OPENSSL_CTX *ctx, RAND_DRBG *parent, int drbg_type);
+static RAND_DRBG *drbg_setup(OPENSSL_CTX *ctx, RAND_DRBG *parent,
+                             int drbg_type);
 
 static int get_drbg_params(int type, unsigned int flags, const char **name,
                            OSSL_PARAM params[3])
@@ -229,7 +231,7 @@ static void drbg_ossl_ctx_free(void *vdgbl)
         return;
 
     CRYPTO_THREAD_lock_free(dgbl->lock);
-    RAND_DRBG_free(dgbl->master_drbg);
+    RAND_DRBG_free(dgbl->primary_drbg);
     CRYPTO_THREAD_cleanup_local(&dgbl->private_drbg);
     CRYPTO_THREAD_cleanup_local(&dgbl->public_drbg);
 
@@ -293,16 +295,16 @@ int RAND_DRBG_set(RAND_DRBG *drbg, int type, unsigned int flags)
     int use_df;
 
     if (type == 0 && flags == 0) {
-        type = rand_drbg_type[RAND_DRBG_TYPE_MASTER];
-        flags = rand_drbg_flags[RAND_DRBG_TYPE_MASTER];
+        type = rand_drbg_type[RAND_DRBG_TYPE_PRIMARY];
+        flags = rand_drbg_flags[RAND_DRBG_TYPE_PRIMARY];
     }
 
     if (drbg->parent == NULL) {
-        reseed_interval = master_reseed_interval;
-        reseed_time_interval = master_reseed_time_interval;
+        reseed_interval = primary_reseed_interval;
+        reseed_time_interval = primary_reseed_time_interval;
     } else {
-        reseed_interval = slave_reseed_interval;
-        reseed_time_interval = slave_reseed_time_interval;
+        reseed_interval = secondary_reseed_interval;
+        reseed_time_interval = secondary_reseed_time_interval;
     }
     *p++ = OSSL_PARAM_construct_uint(OSSL_DRBG_PARAM_RESEED_REQUESTS,
                                      &reseed_interval);
@@ -371,9 +373,10 @@ int RAND_DRBG_set_defaults(int type, unsigned int flags)
     }
 
     all = ((flags & RAND_DRBG_TYPE_FLAGS) == 0);
-    if (all || (flags & RAND_DRBG_FLAG_MASTER) != 0) {
-        rand_drbg_type[RAND_DRBG_TYPE_MASTER] = type;
-        rand_drbg_flags[RAND_DRBG_TYPE_MASTER] = flags | RAND_DRBG_FLAG_MASTER;
+    if (all || (flags & RAND_DRBG_FLAG_PRIMARY) != 0) {
+        rand_drbg_type[RAND_DRBG_TYPE_PRIMARY] = type;
+        rand_drbg_flags[RAND_DRBG_TYPE_PRIMARY] = flags
+                                                  | RAND_DRBG_FLAG_PRIMARY;
     }
     if (all || (flags & RAND_DRBG_FLAG_PUBLIC) != 0) {
         rand_drbg_type[RAND_DRBG_TYPE_PUBLIC]  = type;
@@ -381,7 +384,8 @@ int RAND_DRBG_set_defaults(int type, unsigned int flags)
     }
     if (all || (flags & RAND_DRBG_FLAG_PRIVATE) != 0) {
         rand_drbg_type[RAND_DRBG_TYPE_PRIVATE] = type;
-        rand_drbg_flags[RAND_DRBG_TYPE_PRIVATE] = flags | RAND_DRBG_FLAG_PRIVATE;
+        rand_drbg_flags[RAND_DRBG_TYPE_PRIVATE] = flags
+                                                  | RAND_DRBG_FLAG_PRIVATE;
     }
     return 1;
 }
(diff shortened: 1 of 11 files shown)


levitte:
Hmmm, could we have *one* replacement, not three?  I've seen:

-   "normal"
-   "t" (supposedly for "test")
-   "leaf"

(my preference is "leaf", should that matter)

paulidale:
`t` is a mistake.
`leaf` is in the test case where there is a chain of three DRBGs.  They might be better numbered than named.
`normal` is for the public and private DRBGs -- these are both chains of two so _leaf_ doesn't seem appropriate.

paulidale:
I've addressed `t` and `leaf`.  `normal` remains.

test/drbgtest.c
```diff
@@ -983,84 +983,84 @@ static int test_multi_thread(void)
...
-    /* During a normal reseed, only the slave DRBG should be reseed */
-    inc_reseed_counter(i);
-    mreseed = reseed_counter(m);
-    ireseed = reseed_counter(i);
-    sreseed = reseed_counter(s);
-    if (!TEST_true(RAND_DRBG_reseed(s, NULL, 0, 0))
-        || !TEST_int_eq(reseed_counter(m), mreseed)
-        || !TEST_int_eq(reseed_counter(i), ireseed)
-        || !TEST_int_gt(reseed_counter(s), sreseed))
+    /* During a normal reseed, only the leaf DRBG should be reseed */
```
mattcaswell, line 1003: should be reseeded?

crypto/rand/drbg_lib.c
```diff
@@ -97,10 +97,10 @@ static unsigned int rand_drbg_flags[3] = {
 };
 
 static unsigned int master_reseed_interval = MASTER_RESEED_INTERVAL;
-static unsigned int slave_reseed_interval  = SLAVE_RESEED_INTERVAL;
+static unsigned int normal_reseed_interval  = NORMAL_RESEED_INTERVAL;
```
mspncp, line 100: Please remove the extra space before the `=` to align the two equal signs.

crypto/rand/drbg_lib.c
```diff
@@ -97,10 +97,10 @@ static unsigned int rand_drbg_flags[3] = {
...
 
 static time_t master_reseed_time_interval = MASTER_RESEED_TIME_INTERVAL;
-static time_t slave_reseed_time_interval  = SLAVE_RESEED_TIME_INTERVAL;
+static time_t normal_reseed_time_interval  = NORMAL_RESEED_TIME_INTERVAL;
```
mspncp, line 103: same here

richsalz:
> The term master is part of the public API, it cannot easily be removed or changed.

Sure it can, the project has an existing deprecation mechanism. You, in fact, have multiple PR's that use it.

vdukhovni:
Let's not rush into merging this. Finding a more accurate term should be given a bit of time.

richsalz:
> The term master is part of the public API, it cannot easily be removed or changed.

This is not true.  Seven percent have already been deprecated
```
; grep DEPRE.*3 util/libcrypto.num | wc
    408    1632   35882
; wc util/libcrypto.num 
  5138  20552 361292 util/libcrypto.num
; 
```
Commit c5f8713443c8d606ff149b12c478cd6162f2a7ac alone deprecates nearly 100; check out the author.

vdukhovni:
> If we're really looking for more accurate names, then `trunk` vs. `leaf` (or perhaps `branch`) is I think much more clear than `master` vs. `normal`. The trunk DRBG evolves linearly, while `leaf` (or `branch`) DRBG are split off for independently evolving instances.

If we keep `master` for the trunk DRBG, then `derived` is better than `normal`.  The `master` DRBG generates the master secrets (seeds) from which the `derived` DRBGs compute their seed values.  TLS has a master secret, and derived keys, so essentially the pattern is already in use.

test/drbgtest.c
paulidale: Will do.

crypto/rand/drbg_lib.c
paulidale: Okay.

paulidale:
Updates made as suggested, I'm assuming re-review isn't required.  Still, I'll wait for the discussion about improved nomenclature to play out.

vdukhovni:
> Updates made as suggested, I'm assuming re-review isn't required. Still, I'll wait for the discussion about improved nomenclature to play out.

I really would like to see something that approximately captures the relationship between the types of DRBG, and `master` vs. `normal` is not particularly compelling, there's no indication that the latter is derived from or descends from the former. There is no improvement in clarity, perhaps even a regression. Let's find terms that hint at the right mental model. I really don't think we're there yet.

t-j-h:
_normal_ really isn't the right semantic label for this - the comments show _parent_ used and a dependency relationship so you could perhaps just use _master_ and _dependent_

mspncp:
I‘m putting an OTC hold on this because I think this question should not be considered in isolation but rather as part of a general concept for the future relationship between the three random APIs, RAND, RAND_DRBG, and EVP_RAND. Since the current plan seems to be to deprecate (almost) the entire public RAND_DRBG interface anyway, we are not tied to the term _master_ in the future anymore.

paulidale:
I don't like the chances of RAND/DRBG default setup making it into 3.0.  I suspect we're stuck with the master/public/private arrangement until 3.1, although it's likely to be deprecated for public access.

Note: applications can override RAND still and DRBGs can be chained together fine.

DDvO:
Just my 2 cents (not being a native speaker) on the discussion on re-wording "slave" and potentially re-wording "master":

These (and other 'debatable') words are generally used in the IT world without any intention to discriminate against anybody.

Still I am aware of the fear that some people may still feel offended by the word "slave".
Wherever we can avoid this name without major technical trouble (as done in this PR) I can agree with changing it.

I do not see any real problem with using the word "master", which in my view has a neutral or positive spin.
I see two major meanings of master:
* "lord" (or in German: "Herr"): a person who has authority. This does not imply that the authority is exercised in a negative way. While of course there are many cases of bad masters, there are also many cases of good masters.
* "champion" (or in German: "Meister"): a person who is (verifyably) good at doing something.

If we were to deprecate and rename "master", then for the same reason(s)
* all Git repositories should rename their main branch (which by default is called `master`) and
* all people holding an academic title like "Master of X" would get affected - and most likely offended.

mspncp:
I agree with @DDvO that the term 'master' does not have a negative attitude per se. But as part of the antagonism 'master <-> slave' it clearly does. I'm not opposed to just replacing the word 'slave', but currently it seems like we are having problems finding a suitable replacement. So all I was saying: if we find a better replacement without the word 'master', then let's deprecate it. 

The reason why I placed the hold is because I agree with @vdukhovni:

> Let's not rush into merging this. Finding a more accurate term should be given a bit of time. 

I'm not requesting a formal vote (at least at this moment) but I have the impression that there is no consensus between the OTC members yet. After all, that's the meaning of "need otc decision", isn't it? At least for me it doesn't mean: "let's stop talking and just have a vote". As soon as we found a suitable compromise I'll remove the hold again. I intend to make a suggestion myself, but I don't want to be rushed into it.

vdukhovni:
> _normal_ really isn't the right semantic label for this - the comments show _parent_ used and a dependency relationship so you could perhaps just use _master_ and _dependent_

That's roughly the same as my suggestion of `derived`, but I'd like to suggest that `derived` is closer, because the `dependence` is momentary, at the point at which the DRBG is split off the master instance, it then runs independently.  So, while dependent is closer to mark, something that suggests a point-in-time evolution from the parent DRBG, and then independent evolution thereafter would be IMHO be better.  It could be `parent/child`, or `master/derived` or `trunk/branch` (I see some arguments against `leaf` above)...

vdukhovni:
> Still I am aware of the fear that some people may still feel offended by the word "slave".
> Therefore, wherever we can avoid this word without major technical trouble (as done in this PR), I can agree with changing it.

We should not be in the business of retroactively scrubbing text that was written without hostile intent, and whose meaning is plain and correct in context.  We will offend as many or more people when we transgress into policing Newspeak, as we might include by correcting a couple of identifiers in some corner of the API.

Our documentation should first and foremost be clear and correct.  In new contributions, where otherwise equally good or better alternatives are available, we can give appropriate weight to potential negative secondary meanings, and suggest rephrasing some text as needed.  We should otherwise refrain from applying dialectical materialism to OpenSSL documentation.

In this case, the original words are not particularly apt terms for the types of DRBGs, and it is fine to seek better terms on that basis.

richsalz:
Being pedantically correct, rather than replacing words that hurt some people, is the wrong choice to make.

This is not about "scrubbing words that were written with no hostile intent."  It is about removing archaic terminology to make the project more welcome.  It is about living up to your own code of conduct.

Primary/secondary are fine terms, and seem apt.  #12089 did that.

mspncp:
## Some historical background

Before making my suggestion, let me quickly give you some historical background to how the term _slave_ entered the code and why that name was a bad choice from the very beginning. Not because of political (in-)correctness but for technical reasons. And trust me, I know the story. Because I happened to be the culprit who actually introduced the term when I added the master DRBG in commit a93ba405b0327db9106f8f224112a2b64fb264e7:

    Author: Dr. Matthias St. Pierre <Matthias.St.Pierre@ncp-e.com>
    Date:   Fri Nov 24 14:59:58 2017 +0100

    Add master DRBG for reseeding
    
    A third shared DRBG is added, the so called master DRBG. Its sole purpose
    is to reseed the two other shared DRBGs, the public and the private DRBG.
    The randomness for the master DRBG is either pulled from the os entropy
    sources, or added by the application using the RAND_add() call.
    
    The master DRBG reseeds itself automatically after a given number of generate
    requests, but can also be reseeded using RAND_seed() or RAND_add().
    A reseeding of the master DRBG is automatically propagated to the public
    and private DRBG. This construction fixes the problem, that up to now
    the randomness provided by RAND_add() was added only to the public and
    not to the private DRBG.
    
    Signed-off-by: Dr. Matthias St. Pierre <Matthias.St.Pierre@ncp-e.com>
    
    Reviewed-by: Paul Dale <paul.dale@oracle.com>
    Reviewed-by: Kurt Roeckx <kurt@roeckx.be>
    (Merged from https://github.com/openssl/openssl/pull/4402)

### child vs. parent DRBG

The terms _child_ and _parent_ DRBG had a well defined meaning from the very beginning. Every DRBG instance needs a random source to (re-)seed in a regular base. There are two cases:

- The DRBG instance (re-)seeds from the entropy source provided by the operating system (or whatever else OpenSSL is configured to use).

- The DRBG (re-)seeds from another DRBG instance, in which case the two DRBGs are called the _child_ and the _parent_ DRBG, respectively. The _parent_ of a DRBG can only be configured before it is instantiated and normally never changes.

The default reseeding interval for all DRBGs was 2^24.

https://github.com/openssl/openssl/blob/56d362881eb9173d74f89aa8c4c4b42bc3397a17/crypto/rand/drbg_rand.c#L360
https://github.com/openssl/openssl/blob/56d362881eb9173d74f89aa8c4c4b42bc3397a17/crypto/rand/rand_lcl.h#L24


### public, private, and master DRBG

To reduce the number of system calls for reseeding, we introduced the concept of a singleton DRBG, which would be the only OpenSSL DRBG to (re-)seed from the operating system and serving as a random source for all other DRBGs, which were chained (directly or indirectly) to it. THis singleton was called the _master_ DRBG. So the term _master_ was originally meant in the sense of _main_ or _root_ only. 


###  slave vs. master

When you look at commit a93ba405b0, you will notice that the term _slave_ was used solely in two locations, in the context of defining the reseed_interval:

```
msp@debian:~/src/openssl$ git grep -n -i slave a93ba405b0
a93ba405b0:crypto/rand/drbg_lib.c:623:        drbg->reseed_interval = SLAVE_RESEED_INTERVAL;
a93ba405b0:crypto/rand/rand_lcl.h:28:# define SLAVE_RESEED_INTERVAL                   (1 << 16)
```

With the introduction of the master DRBG, the public and private DRBG would have received fresh OS entropy only after approximately 2^48 generate requests. That meant, practically never.

So I decided to lower the reseeding time, and out of fear to reseed too often, I split the 2^24 unequally

https://github.com/openssl/openssl/blob/a93ba405b0327db9106f8f224112a2b64fb264e7/crypto/rand/rand_lcl.h#L23-L28

The intention was to use a default value 2^8 for the _master_ DRBG only and 2^16 for _every other_ DRBG.

https://github.com/openssl/openssl/blob/a93ba405b0327db9106f8f224112a2b64fb264e7/crypto/rand/drbg_lib.c#L619-L624

So the term _slave_ happened to be an awkward choice with the meaning: "all DRBGs, except the master DRBG". Maybe I should have called it the _default_ reseed interval instead.

```
MASTER_RESEED_INTERVAL     # default for master DRBG only
DEFAULT_RESEED_INTERVAL    # default for all other DRBGs
```

### Thread local DRBGs

For performance reasons, every thread received its own pair of public and private DRBGs, which had the advantage that only the master DRBG needed to have a locking mechanism.

The downside was that it was not so easy anymore to set the default reseeding parameters of the public and private DRBG at application startup anymore. This is when `RAND_DRBG_set_reseed_defaults()` came into place, so finally the term _slave_ also ended up in the public API.


## My Suggestion

After digging into the source code and thinking everything over, my favourite choice is to use the following naming conventions:

* Use the terms _child_ and _parent_ for two DRBG instances where the former (re-)seeds from the latter.
* Use the term _primary_ for a DRBG which has no parent, i.e., reseeds from the OS entropy source
* Use the term _secondary_ for a DRBG which has a parent.


### Rationale

In hindsight, the choice **master vs. slave** was a mistake, and **master vs. default** is actually not much better because it's not intuitive.

Of all suggestions, I like **primary vs. secondary** best, which was coined by @rsalz. A _primary_ DRBG is a DRBG which has no dependency on another DRBG (i.e., no parent), and a _secondary_ DRBG has such a dependency. But this naming convention would only be consistent if the master DRBG would be renamed to primary DRBG.

If there is no majority for dropping the name master, then @t-j-h's suggestion **master vs. dependent** comes second best IMO.

### Compatibility and Depracation

As for compatibility: all we promised for OpenSSL 3.0 is that applications need only recompile. So it is legal to simply rename the public symbol from e.g. `RAND_DRBG_get0_master()` to `RAND_DRBG_get0_primary()` and add a compatility define

```
#define RAND_DRBG_get0_master() #define RAND_DRBG_get0_primary()
```

which could be marked deprecated immediately in version 3.0.

mspncp:
And to everybody else of you who still tries to reduce the discussion to a purely technical queston I would like to say: The choice of words _does_ matter, even if it was made unintentionally. I regret having made that choice without giving it a deeper thought and apologize for that.

_(If you're still not convinced, take a look at this excellent [talk about welcoming nomenclature](https://youtu.be/hZuFeFuazwo) which was [shared by](https://twitter.com/iamamoose/status/1279309543990800384) @iamamoose.)_

paulidale:
Unless there are objections, I'll switch over to primary and secondary.
This lets us introduce tertiary at some point in the future.

paulidale:
I've changed to the primary/secondary nomenclature now and dismissed all reviews.

crypto/rand/drbg_lib.c
```diff
@@ -41,73 +41,75 @@ typedef struct drbg_global_st {
     /*
      * The three shared DRBG instances
      *
-     * There are three shared DRBG instances: <master>, <public>, and <private>.
+     * There are three shared DRBG instances: <primary>, <public>, and
+     * <private>.  The <public> and <private> DRBGs are secondary ones.
```
vdukhovni, line 45: Here (though it is mentioned below), I'd be tempted to add: 
```
They're used for non-secret (e.g. nonces) and secret (e.g. private keys) data respectively.
```
So that it is immediately clear what "public" vs. "private" are about.  This is however a source file, and so not essential.

crypto/rand/drbg_lib.c
```diff
@@ -730,32 +734,32 @@ int RAND_DRBG_set_reseed_time_interval(RAND_DRBG *drbg, time_t interval)
 /*
  * Set the default values for reseed (time) intervals of new DRBG instances
  *
- * The default values can be set independently for master DRBG instances
- * (without a parent) and slave DRBG instances (with parent).
+ * The default values can be set independently for primary DRBG instances
+ * (without a parent) and normal DRBG instances (with parent).
```
vdukhovni, line 738:
```suggestion
 * (without a parent) and secondary DRBG instances (with parent).
```

test/drbgtest.c
```diff
@@ -983,84 +983,84 @@ static int test_multi_thread(void)
...
-    /* During a normal reseed, only the slave DRBG should be reseed */
-    inc_reseed_counter(i);
-    mreseed = reseed_counter(m);
-    ireseed = reseed_counter(i);
-    sreseed = reseed_counter(s);
-    if (!TEST_true(RAND_DRBG_reseed(s, NULL, 0, 0))
-        || !TEST_int_eq(reseed_counter(m), mreseed)
-        || !TEST_int_eq(reseed_counter(i), ireseed)
-        || !TEST_int_gt(reseed_counter(s), sreseed))
+    /* During a normal reseed, only the leaf DRBG should be reseeded */
```
vdukhovni, line 1003:
```suggestion
    /* During a normal reseed, only the secondary DRBGs should be reseeded */
```

test/drbgtest.c
```diff
@@ -983,84 +983,84 @@ static int test_multi_thread(void)
...
-    /* During a normal generate, only the slave DRBG should be reseed */
-    inc_reseed_counter(i);
-    mreseed = reseed_counter(m);
-    ireseed = reseed_counter(i);
-    sreseed = reseed_counter(s);
-    if (!TEST_true(RAND_DRBG_generate(s, buf1, sizeof(buf1), 0, NULL, 0))
-        || !TEST_int_eq(reseed_counter(m), mreseed)
-        || !TEST_int_eq(reseed_counter(i), ireseed)
-        || !TEST_int_gt(reseed_counter(s), sreseed))
+    /* During a normal generate, only the leaf DRBG should be reseed */
```
vdukhovni, line 1025:
```suggestion
    /* During a normal generate, only the secondary DRBGs should be reseed */
```

test/drbgtest.c
```diff
@@ -983,84 +983,84 @@ static int test_multi_thread(void)
...
-    /* Verify that a normal reseed still only reseeds the slave DRBG */
-    inc_reseed_counter(i);
-    mreseed = reseed_counter(m);
-    ireseed = reseed_counter(i);
-    sreseed = reseed_counter(s);
-    if (!TEST_true(RAND_DRBG_reseed(s, NULL, 0, 0))
-        || !TEST_int_eq(reseed_counter(m), mreseed)
-        || !TEST_int_eq(reseed_counter(i), ireseed)
-        || !TEST_int_gt(reseed_counter(s), sreseed))
+    /* Verify that a normal reseed still only reseeds the leaf DRBG */
```
vdukhovni, line 1048:
```suggestion
    /* Verify that a normal reseed still only reseeds the secondary DRBGs */
```

doc/man3/RAND_DRBG_get0_primary.pod
```diff
@@ -2,10 +2,10 @@
 
 =head1 NAME
 
-OPENSSL_CTX_get0_master_drbg,
+OPENSSL_CTX_get0_primary_drbg,
 OPENSSL_CTX_get0_public_drbg,
 OPENSSL_CTX_get0_private_drbg,
-RAND_DRBG_get0_master,
+RAND_DRBG_get0_primary,
```
mattcaswell, line 8: I am unconvinced by the necessity for renaming this particular API. We've decided to change our internal naming - fine. And it makes sense to rename `OPENSSL_CTX_get0_master_drbg` accordingly - since that has never appeared in a stable build. But haven't we more-or-less decided to deprecate the RAND_DRBG APIs? Why introduce a new function in an API which we're about to deprecate? Just let it be deprecated and it will (eventually) be removed from the source anyway.

mspncp: The motivation of my suggestion was that renaming the public symbol and adding a compatibility #define would be sufficient to satisfy our promise (recompile and it will work) and will not break the API. Deprecating a #define is much simpler than having two public symbols for an interim phase and deprecating the old public symbol.

mspncp:
I'm removing the otc hold again, because I've got the impression that we are on a good way to find a consensus now.

doc/man3/RAND_DRBG_get0_primary.pod
mattcaswell: > The motivation of my suggestion was that renaming the public symbol and adding a compatibility #define would be sufficient to satisfy our promise (recompile and it will work) and will not break the API. Deprecating a #define is much simpler than having two public symbols for an interim phase and deprecating the old public symbol.

I'm not sure I understand the point you're trying to make. My argument is that there shouldn't be a `RAND_DRBG_get0_primary` at all. Don't have have it as a new symbol or a define. Just deprecate `RAND_DRBG_get0_master`. That should be sufficient to discourage new users from using the API and won't break existing users (but may give them deprecation warnings to encourage them to change to the new OPENSSL_CTX_get0_primary() way of doing things). Why would we introduce `RAND_DRBG_get0_primary` if we're going to immediately deprecate it and say don't use this use this use `OPENSSL_CTX_get0_primary()` instead?

richsalz:
Someone is deleting my posts.  I am discussing a factual error in this PR.  The intro text says "The term master is part of the public API, it cannot easily be removed or changed."  This is false.  The next release has several hundred deprecated API's.  If the author is confused about how to do that, take a look at c5f8713443c8d606ff149b12c478cd6162f2a7ac.  Please don't delete this.

doc/man3/RAND_DRBG_get0_primary.pod
mspncp: Ah, I understand. I thought that `RAND_DRBG_get0_master()` was still used internally and needed a replacement. But if @paulidale decides to use `OPENSSL_CTX_get0_primary()` internally then you are right, it's enought to deprecate `RAND_DRBG_get0_master()`.

mspncp: > Ah, I understand. I thought that `RAND_DRBG_get0_master()` was still used internally and needed a replacement. But if @paulidale decides to use `OPENSSL_CTX_get0_primary()` internally then you are right, it's enough to deprecate `RAND_DRBG_get0_master()`.

test/drbgtest.c
paulidale: This introduces a semantic error.  The test sets up a three long DRBG chain.  There are two secondaries in the chain but only the last is expected to be reseeded.  I've added clarification to indicate this.

test/drbgtest.c
paulidale: Same comment as before but the clarification is simplified to _last_.

doc/man3/RAND_DRBG_get0_primary.pod
paulidale: I think deprecation of the old calls and removal / internalisation of the new is the way to go here.
I'm not that all of this should be done in this PR.  I've reverted _primary_ to _master_ for this one call.

paulidale:
Travis is timeouts rather than errors.

t8m:
> Travis is timeouts rather than errors.

Not really timeouts but some Travis environment setup errors. But irrelevant to the PR anyway.

doc/man3/RAND_DRBG_get0_public.pod
```diff
@@ -80,9 +80,12 @@ L<RAND_DRBG(7)>
 
 =head1 HISTORY
 
-The OPENSSL_CTX_get0_master_drbg(), OPENSSL_CTX_get0_public_drbg() and
+The OPENSSL_CTX_get0_primary_drbg(), OPENSSL_CTX_get0_public_drbg() and
 OPENSSL_CTX_get0_private_drbg() functions were added in OpenSSL 3.0.
 
+The RAND_DRBG_get0_master() function was renamed to RAND_DRBG_get0_master()
```
t8m, line 86: This paragraph is superfluous.

paulidale: Yes it's superfluous.  Removed.

doc/man3/RAND_DRBG_get0_primary.pod
richsalz: Why don't you handle this the way other deprecations have been handled?  Replace all uses in the code base, and have a compatibility define.  I am puzzled.

paulidale:
Merged to master (a little early).
Thanks for all the comments and reviews.

PR #12369, opened 2020-07-04
Configuration: darwin64-arm64-cc for Apple silicon

stuartcarnie:
The PR adds a new configuration, `darwin64-arm64-cc` for targeting Apple silicon (tested on DTK).

### DIFF
--- a/Configurations/10-main.conf
+++ b/Configurations/10-main.conf
@@ -1557,6 +1557,14 @@ my %targets = (
         bn_ops           => "SIXTY_FOUR_BIT_LONG",
         perlasm_scheme   => "macosx",
     },
+    "darwin64-arm64-cc" => {
+        inherit_from     => [ "darwin-common", asm("aarch64_asm") ],
+        CFLAGS           => add("-Wall"),
+        cflags           => add("-arch arm64"),
+        lib_cppflags     => add("-DL_ENDIAN"),
+        bn_ops           => "SIXTY_FOUR_BIT_LONG",
+        perlasm_scheme   => "ios64",
+    },
 
 ##### GNU Hurd
     "hurd-x86" => {
--- a/CHANGES
+++ b/CHANGES
@@ -15,6 +15,9 @@
 
   *) The Oracle Developer Studio compiler will start reporting deprecated APIs
 
+  *) Add support for Apple Silicon M1 Macs with the darwin64-arm64-cc target.
+     [Stuart Carnie]
+
  Changes between 1.1.1f and 1.1.1g [21 Apr 2020]
 
   *) Fixed segmentation fault in SSL_check_chain()
--- a/config
+++ b/config
@@ -253,11 +253,8 @@ case "${SYSTEM}:${RELEASE}:${VERSION}:${MACHINE}" in
 	    Power*)
 		echo "ppc-apple-darwin${VERSION}"
 		;;
-	    x86_64)
-		echo "x86_64-apple-darwin${VERSION}"
-		;;
 	    *)
-		echo "i686-apple-darwin${VERSION}"
+		echo "${MACHINE}-apple-darwin${VERSION}"
 		;;
 	esac
 	exit 0
@@ -497,6 +494,9 @@ case "$GUESSOS" in
 	else
 	    OUT="darwin64-x86_64-cc"
 	fi ;;
+  $MACHINE-apple-darwin*)
+	OUT="darwin64-$MACHINE-cc"
+	;;
   armv6+7-*-iphoneos)
 	__CNF_CFLAGS="$__CNF_CFLAGS -arch armv6 -arch armv7"
 	__CNF_CXXFLAGS="$__CNF_CXXFLAGS -arch armv6 -arch armv7"


beldmit:
(Just for clarity)
This is a port to newly announced Apple processors for new Macs, isn't it?

beldmit (dismissed):
LGTM

levitte:
Before this goes through, a CLA is required.

stuartcarnie:
@levitte how do I sign the CLA?

stuartcarnie:
@beldmit that is correct, this is for Apple's newly announced ARM hardware

stuartcarnie:
No worries - I found and sent of a signed copy of the agreement to legal.

MonkeybreadSoftware:
For the time bing, we use the following flags here:

 -mmacosx-version-min=11.0 -isysroot /Applications/Xcode-beta.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk

as arm64 on MacOS is not available before 11.0, we can set the minimum there.
And as for my Mac the default Xcode is 11.x, the 12.0 beta is explicit referenced by path here for the Mac SDK.

MonkeybreadSoftware:
I can confirm it builds.
But we have to specify the -isysroot in the variables here for both linker and compiler to use the SDK in the Xcode beta and link with arm64 libraries.

From my build script:

export CC="clang"
export CXX="clang"
export CFLAGS=" -O2 -mmacosx-version-min=11.0 -arch arm64 "
export LDFLAGS=" -arch arm64 -isysroot /Applications/Xcode-beta.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk "
export CPPFLAGS=" -I.. -I/Users/cs/Development/Libs/zlib/macArm -I/Users/cs/Development/Libs/zlib/zlib -isysroot /Applications/Xcode-beta.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk "

./Configure shared enable-rc5 zlib darwin64-arm64-cc

mattcaswell:
Close/reopen to kick CLA bot.

shepmaster:
I cherry picked this into the code used by rust-openssl and it allowed native builds. 👍

lucasderraugh:
When might we see this merged in and added to a release?

mattcaswell:
Looks like this has the necessary approvals - except there is a discrepancy above.

@levitte - you've labelled this for "master" but the PR is targeting 1.1.1.

Do you intend that it should be applied to both?

hjmallon:
Since an almost identical change to cover this has been applied to `master` now will this issue be accepted to 1.1.1 stable? An accepted 1.1.1 stable patch would be useful for package managers to reference, even if there isn't an immediate requirement for a 1.1.1 release.

kaduk:
The 1.1.1 stable branch is in "bugfix-only" mode, and the thread at https://mta.openssl.org/pipermail/openssl-project/2020-June/002044.html doesn't seem to have reached a clear conclusion about whether hardware-support patches fall on the "feature" or "bugfix" side of the line.
(Also, we seem to actually be overdue for a 1.1.1 release, based on the "approximately three months" schedule since 1.1.1g, which came out on April 21, 2020.)

hjmallon:
I see what you mean, and it is completely up to the maintainers to make that decision. I would just like to say:

* Package managers will need to pick up this patch or similar.
  * It's obviously prefereable to get OpenSSL straight from the source, without extra patches / forks.
* OpenSSL 3.0.0 is not slated to [release](https://wiki.openssl.org/index.php/OpenSSL_3.0_Release_Schedule) until "2020 early Q4".
* We expect consumer Apple Silicon hardware before the end of the year, dev hardware is already out there.
* OpenSSL 3.0.0 is a API bump, so could require more work for some users than 'drop it in'.

I think it is completely fair to have 1.1.1 be bugfix only, but only when there is a release of OpenSSL3 to point people towards, before then there could be some leniency towards minor patches for hardware enablement. For now I will merge the patch on 1.1.1 for the package manager I use for OpenSSL.

MonkeybreadSoftware:
I would hope this patch goes into an 1.1.1 update as everyone will want to just rebuild apps for the new Apple Mac platform.

fxcoudert:
This is already being shipped through Homebrew, but we would welcome it being merged and released as a 1.1.1 update, so all users on macOS get the exact same (and definitive) version.

kaduk:
Attempted to make the labels reflect reality.
AFAICT the code has not changed since @levitte and @beldmit approved, so setting "approval:done", but the question of whether this is appropriate for 1.1.1 needs OTC action.

mattcaswell:
Removing the "beta 1" milestone, since this PR targets 1.1.1.

shepmaster:
Now that 1.1.1h has shipped without this, and based on the previously quoted [approximately three months](https://github.com/openssl/openssl/pull/12369#issuecomment-671600487) release cadence, is it correct to assume that this support will not be shipped in 2020?

kaduk:
> Now that 1.1.1h has shipped without this, and based on the previously quoted [approximately three months](https://github.com/openssl/openssl/pull/12369#issuecomment-671600487) release cadence, is it correct to assume that this support will not be shipped in 2020?

Per https://mta.openssl.org/pipermail/openssl-project/2020-September/002221.html there should be one more 1.1.1 release in 2020 ... but this PR is blocking on an OTC decision, for which I have not seen indications of a particular timeline.

fxcoudert:
Hi, Homebrew maintainer here: for those of us not knowledgable about the OpenSSL governance, is there a way to request/spur an OTC decision, or at least to know whether this is being considered or silently ignored?

kaduk:
@levitte could you opine on the question of starting an OTC vote, since you approved the changes?
(It seems that I am the one who added the "hold: need OTC decision" label and thus could technically remove it, but that doesn't feel right, since I'm not on the OTC...)

mattcaswell:
> @levitte could you opine on the question of starting an OTC vote, since you approved the changes?

OTC can only consider whether this meets current policy or not. Current policy says only bug and security fixes are allowed in stable branches. I think it is a stretch to say that this qualifies as a bug fix.

In order to *change* the policy in order to allow a broader set of changes to be backported I think that you have to have an OMC decision. There might be some support for that. See the recent discussion in #11188.

kaduk:
@mattcaswell thanks for clarifying.  What do you think the correct label state is for this PR at this time?  (Please feel empowered to remove the "OTC hold" label that I presumptively added.)

mattcaswell:
Changed this to an OMC hold.

tawmoto:
Seems approved, does it need something else before merge? thanks

mattcaswell:
> Seems approved, does it need something else before merge? thanks

This PR has an OMC hold on it, so its awaiting an OMC decision. OMC policy does not currently allow the backport of this change to 1.1.1 (it's considered a new feature, and new features are not allowed to be introduced in stable releases). However there is an ongoing discussion to change that policy, which would allow the addition of new platforms into a stable release.

See:

https://mta.openssl.org/pipermail/openssl-project/2020-October/002380.html

tawmoto:
@mattcaswell thanks a lot for your explanations, what do you mean by "backport to 1.1.1"? where should be added if not in 1.1.1?
Because version 3 is not stable yet.

mattcaswell:
> @mattcaswell thanks a lot for your explanations, what do you mean by "backport to 1.1.1"? where should be added if not in 1.1.1?
Because version 3 is not stable yet and companies use only stable versions. (1.1.1 in openssl's case)
cheers!

There is a similar change already in master (which will become 3.0). New features are only allowed in master. This PR backports that functionality from master to 1.1.1.

felixbuenemann:
Just to let you know, this issue has been duped by #13476, if anyone wants to chime in on the bugfix vs feature discussion.

felixbuenemann:
As suggested by @richiksc, here's a copy of my summary from #13476, after arguing that this is a bugfix and not a feature:

* When adding patches to Homebrew Formulae, the patches have to be pushed upstream (happening here by proxy)
* This change is configuration only, the chance to break existing targets is zero.
* The existing aarch64 assembly code is tested on iOS via ios64 assembly target.
* The M1 chip is an evolution of those ARMv8 chips (it is ARMv8.4 compatible) and can even run existing iOS apps without emulation, showing it is indeed fully compatible.

felixbuenemann:
@mattcaswell I have looked through the LTS+ discussion and it sounds like this will only add confusion and more maintenance work to handle different releases. The same could be achieved by altering the LTS rules instead of adding another release type.

Many existing projects like Ubuntu regularly update their LTS releases with new kernels to support new hardware, so adding hardware support as part of LTS is common practice.

The whole goal of LTS is to provide a version that can be used and depended on over an extended period of time which also means it _needs_ to support new hardware, otherwise people will have to port their software to new versions before they can upgrade or switch a server.

**I think what _really_ matters is how risky the change is to break existing code/platforms**.

Back on topic — in the LTS+ mailing list thread a bug fix is defined as:

> - **Fixes for defects that impact behaviour that do not impact the API or the ABI**
> - Minor fixes to existing code to correct major performance regressions (…)
> - Inclusion of new API functions to correct missing accessor functions

It could be argued that existing assembly for the CPU not being used is a defect (extremely poor performance) and the fix for that (adding a build target) is not impacting the API or ABI. So in that sense it is a bug fix.

Under the LTS+ idea this would be explicitly allowed as a "platform support update", but these can't really break the ABI or API either, so these only differ from a bug fix if they _also_ introduce new code under the existing ABI/API.

richiksc:
Please correct me if I'm wrong, because I'm unfamiliar with the structure of the project, but adding a new compile target in configuration has virtually zero risk of breaking functionality on other, pre-existing platforms, right? This is no different to adding another compile target in a buildscript for a project? In that scenario, the only risk would be having the built binaries for `darwin64-arm64-cc` not be fully functional, which would not be a regression. If that is the case, I don't see why it wouldn't be an accepted patch in an LTS release according to the traditional definition of LTS.

In addition, the ios64 asm has already been tested on processors that use largely the same architecture as the M-series SoCs on arm64 Macs, as @felixbuenemann pointed out.

mattcaswell:
> I think what really matters is how risky the change is to break existing code/platforms.

I understand your point of view.  However this change to add a new platform is probably on the least risky end of what adding a new platform often looks like. Typically adding a new platform is usually more involved and invasive. Often requiring amendments to various platform specific files, perhaps adding integration to a platform specific seed source, etc (e.g. take a look at some of @rsbecker's PRs adding support for the non stop platform). So, in general, adding a new platform is not considered an "ultra low" risk activity. It does comes with some impact (not to mention the potential for additional support requests that rolling out to a platform that has not had support might bring).

Until now the OMC has had a simple rule to follow: no new features in stable releases - with new platforms counted as new features. This is simple and easy to explain. If we start accepting some new platforms but not others then suddenly things start become a lot more murky: where is the dividing line between what is acceptable and what is not?

Added to this we have some users who are very conservative about what they want from an LTS release. They have no interest in adding additional platforms. Such users prioritise stability above all else.

And yet, we have other users for whom adding new platforms and performance improvements is a perfectly acceptable risk - and indeed highly desirable.

Our current rules are too strict IMO. But in our internal discussions it has become clear that just loosening them for all users is not going to be an acceptable answer. That is where the LTS+ proposal is coming from. It keeps both camps of users happy at the cost of a slightly raised maintenance overhead. Although in reality I expect the two branches to track fairly closely to each other - so I don't expect the overhead to be that large.

Now, with specific reference to *this* PR it is clear that it does not currently conform to the rules laid out by the OMC. So, if you want to see it merged that gives 3 possible routes to achieve this (that I can think of):

1) Seek a one off exception from the OMC
2) Change the rules for what is acceptable in an LTS release
3) Introduce the LTS+ release type

Having sat in the some very long discussions on this topic I can tell you that (2) is unlikely happen any time soon. (1) is a possibility - but I suspect that the OMC would hold any decision on that until the LTS+ question has been resolved one way or the other. However if the LTS+ proposal fails this is probably the best route (although I don't have a high degree of confidence that such an exception would be accepted).

felixbuenemann:
@mattcaswell Spinning this from another angle: What *is* a new platform?

This change adds neither a new operating system, since it reuses darwin64 code nor a new CPU, since it reuses ios64 ASM code for armv8. So in the end it's just adding a missing configure entry for a platform that is already there.

If it is not accepted, the same patch will be used downstream in various projects anyways, since there's no viable alternative.
Having it upstream prevents fragmentation and is good open source practice.

I think seeking an exception is the way to go here and does not depend on the LTS+ decision, since that handles more complicated additions, like the one you mentioned.

**What would be the process to ask for such an exception from the OMC?**

mattcaswell:
> What would be the process to ask for such an exception from the OMC?

The first step would be to decide what decision it is exactly that you are asking the OMC to make. This needs to be in the form of a proposed vote. You might check the openssl-project email archives where there are numerous examples of various OTC and OMC vote proposals (this would be an OMC vote, but OTC votes work in a very similar way):

https://mta.openssl.org/pipermail/openssl-project/

You'd also need to get an OMC member willing to start your vote. I would probably do that assuming you came up with some reasonable vote text (note I don't necessarily guarantee to vote in favour...I'll need to give that some thought and look at the exact vote text to decide that).

Ideally proposed votes should be discussed on the openssl-project list before the vote actually starts (in order to incorporate any feedback into the vote text). That's a moderated list (you must join the list in order to post), but I'm a moderator and usually let through any on-topic posts.

The actual vote itself will take place on the private OMC list with the results reported back to openssl-project once done.

rsbeckerca:
Having lived through this, it almost seems that the criteria analogous to `CLA: Trivial` might be a reasonable deciding factor on a new platform at this stage. That definitely would not apply in the case of NonStop, which actually had a surprisingly small number of high impact changes that cause the builds to break a few times. But if a platform could be added requiring a one or two line configuration file with no code changes (qualifying for a `CLA: Trivial` PR), perhaps that might be acceptable, since CLA is a well quantified policy. But if the new platform required additional touch-points in the code, that would escalate the requirement to a larger change (but I don't really get a vote, just an opinion).

felixbuenemann:
@mattcaswell I've written a proposal and posted it at https://gist.github.com/ee30b5c8f52e030629dc2de95f81d8b, you can add comments there if anything needs to be changed, so we don't clutter this thread.

richiksc:
@felixbuenemann Link is 404. Are you sure the gist is public?

felixbuenemann:
> @felixbuenemann Link is 404. Are you sure the gist is public?

Link is fixed, the last char was missing.

mattcaswell:
> @mattcaswell I've written a proposal and posted it at https://gist.github.com/ee30b5c8f52e030629dc2de95f81d8b1, you can add comments there if anything needs to be changed, so we don't clutter this thread.

Nice write up. The proposal looks fine to me, but you should post it to the openssl-project list asking for feedback on the vote text. You can state that I have agreed to raise it with OMC if you like (normally vote proposals come from OMC/OTC members - but there is no reason why someone else can't ask for a vote to be considered). Normally I like to leave starting a vote for a few days to give people a chance to feedback on the wording.

stuartcarnie:
@felixbuenemann great write up, thanks!

felixbuenemann:
@mattcaswell OK, I've posted the proposal to the openssl-project list, pending moderation.

mattcaswell:
> @mattcaswell OK, I've posted the proposal to the openssl-project list, pending moderation.

Yep, I saw. I let it through:

https://mta.openssl.org/pipermail/openssl-project/2020-November/002457.html

Configurations/10-main.conf
```diff
@@ -1557,6 +1557,14 @@ my %targets = (
         bn_ops           => "SIXTY_FOUR_BIT_LONG",
         perlasm_scheme   => "macosx",
     },
+    "darwin64-arm64-cc" => {
+        inherit_from     => [ "darwin-common", asm("aarch64_asm") ],
+        CFLAGS           => add("-Wall"),
+        cflags           => add("-arch arm64"),
+        lib_cppflags     => add("-DL_ENDIAN"),
+        bn_ops           => "SIXTY_FOUR_BIT_LONG",
+        perlasm_scheme   => "ios64",
```
mlindner, line 1566: Curiosity question for my own self interest, why is the perlasm_scheme listed as ios64?

felixbuenemann: Because it is the same assembly used for iOS 64-Bit devices. The M1 is basically an evolution of the A-Series chips used in iPhones and iPads.

mattcaswell:
OMC vote has started. I'll let you know the result once I know.

mattcaswell:
The OMC vote has now completed. It passed:

accepted:  yes  (for: 6, against: 1, abstained: 0, not voted: 0)

Therefore I am lifting the OMC hold on this PR.

mattcaswell:
@stuartcarnie - I think this change deserves a mention in the CHANGES file. Please could you add on to this PR?

felixbuenemann:
> @stuartcarnie - I think this change deserves a mention in the CHANGES file. Please could you add one to this PR?

@stuartcarnie I had one on the (dupe) PR #13476 in commit 105be6e21f2f33b687a28b81c6d62e85d54ea860, you can just copy it and change the name.

stuartcarnie:
@felixbuenemann I've updated the CHANGES

t-j-h:
Note that #13517 should be folded in here too so that ./config gets the correct platform (otherwise it will select i686-apple-darwin  as the platform (incorrectly).

mattcaswell:
Given that we just announced that 1.1.1i will be released on 8th December it would be great to get this PR merged before then.

> Note that #13517 should be folded in here too so that ./config gets the correct platform (otherwise it will select i686-apple-darwin as the platform (incorrectly).

@stuartcarnie - are you able to include that change here?

stuartcarnie:
@mattcaswell yes, I can transfer that commit over now

stuartcarnie:
@mattcaswell that fix is based of master; there is no `config.pm` in the 1.1.1 branch. Any suggestions?

stuartcarnie:
I have found the guessing logic in `config`, but that will require I do some testing on my DTK to reimplement it, which I will do later today.

levitte:
Do you want help with the 1.1.1 `config` script?  I know it inside out, so I'm guessing it takes me 5 minutes tops

stuartcarnie:
@levitte yes please! 👏🏻👏🏻

t-j-h:
```
--- config-     2020-12-02 08:16:21.393694700 +1000
+++ config      2020-12-02 08:16:25.397069100 +1000
@@ -253,11 +253,8 @@
            Power*)
                echo "ppc-apple-darwin${VERSION}"
                ;;
-           x86_64)
-               echo "x86_64-apple-darwin${VERSION}"
-               ;;
            *)
-               echo "i686-apple-darwin${VERSION}"
+               echo "${MACHINE}-apple-darwin${VERSION}"
                ;;
        esac
        exit 0
```

t-j-h:
That is the equivalent patch to apply ...

t-j-h:
And another update having actually tested this on a apple silicon device so we have to pick this up in one more location to translate correctly.

```
--- config-     2020-12-02 08:16:21.393694700 +1000
+++ config      2020-12-02 08:32:21.806134400 +1000
@@ -253,11 +253,8 @@
            Power*)
                echo "ppc-apple-darwin${VERSION}"
                ;;
-           x86_64)
-               echo "x86_64-apple-darwin${VERSION}"
-               ;;
            *)
-               echo "i686-apple-darwin${VERSION}"
+               echo "${MACHINE}-apple-darwin${VERSION}"
                ;;
        esac
        exit 0
@@ -497,6 +494,9 @@
        else
            OUT="darwin64-x86_64-cc"
        fi ;;
+  $MACHINE-apple-darwin*)
+        OUT="darwin64-$MACHINE-cc"
+       ;;
   armv6+7-*-iphoneos)
        __CNF_CFLAGS="$__CNF_CFLAGS -arch%20armv6 -arch%20armv7"
        __CNF_CXXFLAGS="$__CNF_CXXFLAGS -arch%20armv6 -arch%20armv7"

```

levitte:
@t-j-h beat me to it :wink:

stuartcarnie:
Thanks! Applying now and will push another commit

stuartcarnie:
@t-j-h I can't apply that patch:

```sh
$ pbpaste | git apply -
error: patch failed: config:253
error: config: patch does not apply
```

levitte (approved):

t-j-h (approved):

beldmit (approved):
LGTM

mattcaswell (approved):

felixbuenemann:
Will this be in 1.1.1i or is it too late for that?

mattcaswell:
> Will this be in 1.1.1i or is it too late for that?

It should be fine for inclusion. The next step is we have to wait 24 hours in case of any objections. But assuming none are raised we can merge - which is in plenty of time for 1.1.1i.

levitte:
Merged.  I took the freedom to squash all the commits into one.

61168b5b8dde03f3b77ddf5e4b1b81c338c01746 Configuration: darwin64-arm64-cc for Apple silicon

mattcaswell:
Thanks everyone who contributed to getting this in.

udiakasnuke:
I am sure 1.1.1 must now be building for arm64(Apple Silicon).
I recently got the need to support arm64.
I took the latest stable branch of 1.1.1. 
./Configure shared enable-rc5 darwin64-arm64-cc
make

Now when i do make it fails also giving supported architecture error, obviously i am using an x86-64 mac.
 
In file included from crypto/armcap.c:10:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/stdio.h:64:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/_stdio.h:75:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/_types/_va_list.h:32:9: error: unknown type name '__darwin_va_list'
typedef __darwin_va_list va_list;
        ^
fatal error: too many errors emitted, stopping now [-ferror-limit=]

Am i running wrong somewhere, Please !! Thanks!

richiksc:
@udiakasnuke To cross compile for arm64 Macs on an x86_64 Mac, you need to have the Xcode 12 CLT and the macOS 11 SDK installed, and then make sure you are using that SDK as the active sdk by setting these two environment variables:

```
SDKROOT=$(xcrun -sdk macosx11.0 --show-sdk-path)
MACOSX_DEPLOYMENT_TARGET=$(xcrun -sdk macosx11.0 --show-sdk-platform-version)
```

udiakasnuke:
@richiksc - Yeah!! Thanks!! I almost forgot that.... Just for summary, i did

1. SDKROOT=$(xcrun -sdk macosx11.1 --show-sdk-path)                 -    I had Xcode 12.3 hence 11,.1
2. MACOSX_DEPLOYMENT_TARGET=$(xcrun -sdk macosx11.1 --show-sdk-platform-version)
3. ./Configure shared enable-rc5 zlib darwin64-arm64-cc
4. make

and i could see the static and dynamic library of libcrypto(.a / .dylib) and libssl(.a / .dylib), and all the headers are already present in include/openssl path.

Sunil-P:
> @richiksc - Yeah!! Thanks!! I almost forgot that.... Just for summary, i did
> 
> 1. SDKROOT=$(xcrun -sdk macosx11.1 --show-sdk-path)                 -    I had Xcode 12.3 hence 11,.1
> 2. MACOSX_DEPLOYMENT_TARGET=$(xcrun -sdk macosx11.1 --show-sdk-platform-version)
> 3. ./Configure shared enable-rc5 zlib darwin64-arm64-cc
> 4. make
> 
> and i could see the static and dynamic library of libcrypto(.a / .dylib) and libssl(.a / .dylib), and all the headers are already present in include/openssl path.

I did the same but for some reason my make command is picking up old sdk (10.15) even though 11.1 is installed and set.
My values are : 
SDKROOT = /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk
MACOSX_DEPLOYMENT_TARGET = 11.1

I am getting a bunch of errors like this 
`/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/sys/_types.h:76:9: error: unknown type name '__uint32_t'; did you mean '__uint128_t'?
`

What am I doing wrong ?

Sunil-P:
> > @richiksc - Yeah!! Thanks!! I almost forgot that.... Just for summary, i did
> > 
> > 1. SDKROOT=$(xcrun -sdk macosx11.1 --show-sdk-path)                 -    I had Xcode 12.3 hence 11,.1
> > 2. MACOSX_DEPLOYMENT_TARGET=$(xcrun -sdk macosx11.1 --show-sdk-platform-version)
> > 3. ./Configure shared enable-rc5 zlib darwin64-arm64-cc
> > 4. make
> > 
> > and i could see the static and dynamic library of libcrypto(.a / .dylib) and libssl(.a / .dylib), and all the headers are already present in include/openssl path.
> 
> I did the same but for some reason my make command is picking up old sdk (10.15) even though 11.1 is installed and set.
> My values are :
> SDKROOT = /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk
> MACOSX_DEPLOYMENT_TARGET = 11.1
> 
> I am getting a bunch of errors like this
> `/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/sys/_types.h:76:9: error: unknown type name '__uint32_t'; did you mean '__uint128_t'? `
> 
> What am I doing wrong ?

Make sure you run 
export SDKROOT=$(xcrun -sdk macosx11.1 --show-sdk-path)

instead of just setting it in terminal. rookie mistake.

Issue #12199, opened 2020-06-18
Broken on 32 bit big endian


kroeckx:
This code does not work on 32 bit big endian machines:

https://github.com/openssl/openssl/blob/9a7319b0b0b9ddaa440b102dc8e197d267700756/test/ossl_test_endian.h#L13-L20

I think it just needs to be changed to an int. But I wonder if we have other code like that that uses a long instead of an int. Maybe all that code should use that macro.

gcc also has built in macro's for that, I suggest we use that by default instead.

sebastianas:
> I think it just needs to be changed to an int. But I wonder if we have other code like that that uses a long instead of an int. Maybe all that code should use that macro.

look at test_param_uint64() in params_api_test.c. 

> gcc also has built in macro's for that, I suggest we use that by default instead.

There is __builtin_bswap16() … 128() if __HAVE_BUILTIN_BSWAP16__ … 128 is defined.
If it is not, the preferred way is to do something like
>static line uint16_t __do_the_swap16(uint16_t val)
>{
>        return 
>         (val & 0x00ffU) << 8) |
>         (val & 0xff00U) >> 8);
>}

or as a define or so. But not loosing the type, using an array and copy it byte by byte.
To figure out what endian you are on, there is
>    __BYTE_ORDER__

which points is defined as

> __ORDER_LITTLE_ENDIAN__
> __ORDER_PDP_ENDIAN__
> __ORDER_BIG_ENDIAN__

So you can either pass the value or do the swap in your LE_to_cpu() macro/inline function.

levitte:
I fail to understand why it matters if `one` is a long or an int...  as long as it's not a char, `little` should end up being zero on big endian machines, no?

paulidale:
We do have other code that uses the same endian test union, lots of it.  This union was copied from one such.  All of these files contain the same thing:

- crypto/modes/ctr128.c
- crypto/modes/gcm128.c
- crypto/modes/siv128.c
- crypto/modes/xts128.c
- crypto/chacha/chacha_enc.c
- crypto/evp/e_chacha20_poly1305.c
- crypto/evp/bio_ok.c
- crypto/sha/sha_local.h
- crypto/sha/sha256.c

This would mean much is broken on 32 bit big endian machines.  I wonder how many 32 bit big endian users we have given this list?


I wouldn't use `params_api_test.c` as a reference for anything endian.  It is verifying that the OSSL_PARAM code is doing things as per the design, it is very very specific.

[...]

PR #12371, opened 2020-07-05
Configurations: PowerPC is big endian

sebastianas:
Define B_ENDIAN on PowerPC because it is a big endian architecture. With
this change the BN* related tests pass.

Fixes: #12199

### DIFF
--- a/Configurations/10-main.conf
+++ b/Configurations/10-main.conf
@@ -706,6 +706,7 @@ my %targets = (
         inherit_from     => [ "linux-generic32" ],
         asm_arch         => 'ppc32',
         perlasm_scheme   => "linux32",
+        lib_cppflags     => add("-DB_ENDIAN"),
     },
     "linux-ppc64" => {
         inherit_from     => [ "linux-generic64" ],


kroeckx:
The test is broken. It should not pass on any big endian system. If this fixes the test, something else is broken.

It's also my understanding that we don't need the L_ENDIAN and B_ENDIAN defines, it should only select an optimization, it should work correctly without that define, which further points to other things being broken.

I would like to know what else is broken before merging this.

sebastianas:
On 2020-07-05 12:17:58 [-0700], Kurt Roeckx wrote:
> The test is broken. It should not pass on any big endian system. If this fixes the test, something else is broken.
> 
> It's also my understanding that we don't need the L_ENDIAN and B_ENDIAN defines, it should only select an optimization, it should work correctly without that define, which further points to other things being broken.
> 
> I would like to know what else is broken before merging this.

Sure. As I wrote I would like to get rid of the macros and use the
compiler provided built-ins/functions I suggested.
Feel free to remove the Fixes line (or I do it myself if you say so).

This just fixes the build I plan to upload alpha4 soonish.
Also, I added a B_ENDIAN to m68k because it was missing the define.

Sebastian

kroeckx:
I think hppa is also missing the define.

paulidale (approved):

paulidale:
It is worthwhile looking at 1.1.1 to see if this is relevant there too -- the endian detection code hasn't changed fundamentally.

levitte (approved):

levitte:
> It is worthwhile looking at 1.1.1 to see if this is relevant there too -- the endian detection code hasn't changed fundamentally.

I agree.  I'll add back the 1.1.1 branch label to clarify (this is a bug-fix, so I see no reason not to, and it's trivial to cherry-pick)

t8m (approved):

t8m:
Merged to master and 1.1.1 branches. Thank you for the PR.

PR #12375, opened 2020-07-06
x509_vfy.c: Improve key usage checks in internal_verify() of cert chains

DDvO:
This is the 'forward-port' I'm proposing as by-product of the backport done in #12357.
See the discussion thread there starting with https://github.com/openssl/openssl/pull/12357#pullrequestreview-441978307.

If a presumably self-signed cert is last in chain we verify its signature
only if X509_V_FLAG_CHECK_SS_SIGNATURE is set. Upon this request we do the
signature verification, but not in case it is a (non-conforming) self-issued
CA certificate with a key usage extension that does not include keyCertSign.

Make clear when we must verify the signature of a certificate
and when we must adhere to key usage restrictions of the 'issuing' cert.
Add some comments for making `internal_verify()` easier to understand.
Update the documentation of `X509_V_FLAG_CHECK_SS_SIGNATURE` accordingly.

Constify `X509_check_akid()` and prefer using `X509_get0_serialNumber()` over `X509_get_serialNumber()`.

### DIFF
--- a/crypto/x509/x509_vfy.c
+++ b/crypto/x509/x509_vfy.c
@@ -1746,6 +1746,7 @@ int x509_check_cert_time(X509_STORE_CTX *ctx, X509 *x, int depth)
     return 1;
 }
 
+/* verify the issuer signatures and cert times of ctx->chain */
 static int internal_verify(X509_STORE_CTX *ctx)
 {
     int n = sk_X509_num(ctx->chain) - 1;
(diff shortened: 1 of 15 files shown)


crypto/x509/x509_vfy.c
```diff
@@ -1784,31 +1785,54 @@ static int internal_verify(X509_STORE_CTX *ctx)
...
+         *
          * Skip signature check for self-signed certificates unless explicitly
          * asked for because it does not add any security and just wastes time.
-         * If the issuer's public key is not available or its key usage does
-         * not support issuing the subject cert, report the issuer certificate
-         * and its depth (rather than the depth of the subject).
          */
-        if (xs != xi || (ctx->param->flags & X509_V_FLAG_CHECK_SS_SIGNATURE)) {
+        if (xs != xi || ((ctx->param->flags & X509_V_FLAG_CHECK_SS_SIGNATURE)
+                         && (xi->ex_flags & EXFLAG_SS) != 0)) {
```
t8m, line 1798: I am not sure about the EXFLAG_SS check here. This will be set only for V3 certs with AKID / SUBJECT key id. I'd still prefer if we force the signature verification with X509_V_FLAG_CHECK_SS_SIGNATURE even for V1 certs. If I have just self signed end-entity V1 certificate and I set X509_V_FLAG_CHECK_SS_SIGNATURE, I'd expect it to be verified.

DDvO: Good thought to make sure that is works also for V1 certs, but I don't see a problem at this point:
For such certs there is no AKID and then for self-issued certs the EXFLAG_SS is always set.
This is directly visible by looking at the definition of `X509_check_akid()`.

t8m: Good, I did not investigate the code in depth. Sorry for the noise.

DDvO:
Merged - thanks again for the quick handling!

PR #12383, opened 2020-07-07
OCSP Response Processing

ashman-p:
Create OCSP_CERTID based on the digest algorithm pulled from the OCSP response.
Since the OCSP responder may use a hash other than SHA-1, allow a cert ID to be generated based on the hash used from the OCSP response.

##### Checklist


t8m:
This is a feature and thus not acceptable to 1.1.1 branch. Would you like to submit a PR for the master branch?

PR #12385, opened 2020-07-07
WIP: Fix aes-ctr based drbg's forking behavior

### LINKED ISSUE #12377: OpenSSL 3.0.0 RNG fork-safety regression? [security?]
Earlier versions of OpenSSL would take some pains to ensure that calling fork() was was safe in the presence of an RNG.  This appears to be gone with the OpenSSL 3.0.0 code.

Specifically: when I write a program that calls RAND_bytes() in the main process, then forks several times, and each child process calls RAND_bytes(), I get the same result from RAND_bytes() in every one of the child processes.  This is still the case when the child processes call RAND_poll(), which suggests to me that something is deeply wrong.

OpenSSL 1.1.1 behaves as expected in this case.

Found by Tor's unit tests. (I don't want to get preachy, but I'd strongly recommend that OpenSSL adds a regression test for this case: this is exactly the kind of a thing we worry about an RNG doing.)

Here is a demo program:
```
#include <openssl/rand.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <sys/types.h>

int should_poll = 1;

void
check_rand_status(void)
{
	if (RAND_status() != 1) {
		puts("RAND_status is unhappy!");
	} else {
		// puts("RAND_status says everything is fine...e");
	}
}

void
try_rand(void)
{
	unsigned u;
	check_rand_status();
	if (should_poll)
		RAND_poll();
	check_rand_status();
	int r = RAND_bytes((void*)&u, sizeof(u));
	if (r == 0) {
		printf("Process %d: error!\n", (int)getpid());
	} else {
		printf("Process %d: rng value is %u.\n",
		       (int)getpid(), u);
	}
	check_rand_status();
}

int main(int argc, char **argv)
{
	check_rand_status();
	try_rand();
	check_rand_status();
	for (int i = 0; i < 8; ++i) {
		pid_t child;
		if ((child = fork())) {
			// parent.
		} else {
			try_rand();
			exit(0);
		}
	}
	return 0;
}
```

Here is an example output from openssl-master:
```
Process 680496: rng value is 1796322885.
Process 680497: rng value is 345659543.
Process 680498: rng value is 345659543.
Process 680499: rng value is 345659543.
Process 680500: rng value is 345659543.
Process 680501: rng value is 345659543.
Process 680502: rng value is 345659543.
Process 680503: rng value is 345659543.
Process 680504: rng value is 345659543.
```

Here is an example output from openssl 1.1.1g:
```
Process 680544: rng value is 2039804308.
Process 680545: rng value is 2038236718.
Process 680546: rng value is 1374486369.
Process 680547: rng value is 2441531771.
Process 680548: rng value is 892829712.
Process 680549: rng value is 1349569669.
Process 680550: rng value is 3048351340.
Process 680551: rng value is 2647008208.
Process 680552: rng value is 1731649072.
```

All tests run on Fedora 32.

p-steuer:
Fixes #12377

### DIFF
--- a/providers/implementations/rands/drbg_ctr.c
+++ b/providers/implementations/rands/drbg_ctr.c
@@ -296,8 +296,7 @@ __owur static int ctr_update(PROV_DRBG *drbg,
         ctr_XOR(ctr, in2, in2len);
     }
 
-    if (!EVP_CipherInit_ex(ctr->ctx_ecb, NULL, NULL, ctr->K, NULL, -1)
-        || !EVP_CipherInit_ex(ctr->ctx_ctr, NULL, NULL, ctr->K, NULL, -1))
+    if (!EVP_CipherInit_ex(ctr->ctx_ecb, NULL, NULL, ctr->K, NULL, -1))
         return 0;
     return 1;
 }
@@ -404,6 +403,10 @@ static int drbg_ctr_generate(PROV_DRBG *drbg,
     }
 
     memset(out, 0, outlen);
+
+    if (!EVP_CipherInit_ex(ctr->ctx_ctr, ctr->cipher_ctr,
+                           NULL, ctr->K, NULL, -1))
+        return 0;
 
     do {
         if (!EVP_CipherInit_ex(ctr->ctx_ctr,


p-steuer:
Merging this one would only hide a broader fork-safety issue, see https://github.com/openssl/openssl/issues/12377#issuecomment-655416222 . Closing.

PR #12389, opened 2020-07-07
DOC: Modify one example in EVP_PKEY_fromdata(3)

### LINKED ISSUE #12388: EVP_PKEY_fromdata/OSSL_PARAM_BN requires data in both big and little endian
The EVP_PKEY_fromdata() manpage has this in it:
```
#ifndef B_ENDIAN
# error "We haven't prepared little endian arrays"
#endif
```

I think the API is wrong if the application needs to know the endianness and nees to have an array in both big and little endian.

(It should also say big, not little.)

levitte:
The example to create an EC key from user data didn't show what one
could expect and application to do, especially with regard to how it's
done with raw EC functions.  We therefore refactor it to make proper
use of a BIGNUM where expected, and also use OSSL_PARAM_BLD(3) for
easier handling of the OSSL_PARAM array.

Fixes #12388


doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -120,24 +120,19 @@ TODO Write a set of cookbook documents and link to them.
...
+ #include <openssl/param_build.h>
 
  /*
-  * These arrays represent large numbers, big endian organization.
-  * In a real application, these would probably be bignums that get
-  * converted to the native integer organization with BN_bn2nativepad().
-  * We're not doing that here, since this is not an example of BIGNUM
-  * functionality, but an example of EVP_PKEY_fromdata().
+  * Fix data to represent the private and public key.  The private
+  * is passed as an integer, while the public key is an octet string.
```
slontis, line 127: I think describing it as a integer here for the user is a bit confusing..
It looks like a BN..

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -148,24 +143,42 @@ TODO Write a set of cookbook documents and link to them.
...
-     OSSL_PARAM_utf8_string("curve-name", "prime256v1"),
-     OSSL_PARAM_BN("priv", priv, sizeof(priv)),
-     OSSL_PARAM_BN("pub", pub, sizeof(pub)),
-     OSSL_PARAM_END
- };
 
  int main()
  {
-     EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_from_name(NULL, "EC", NULL);
+     EVP_PKEY_CTX *ctx;
```
slontis, line 149: There a quite a lot of issues in the example..
Here is a working example that I compiled and tested.
(I use OSSL_PKEY_PARAM_GROUP_NAME as the string changed).
```
int main(void)
{
    EVP_PKEY_CTX *ctx = NULL;
    EVP_PKEY *pkey = NULL;
    BIGNUM *priv = NULL;
    OSSL_PARAM_BLD *param_bld = NULL;
    OSSL_PARAM *params = NULL;
    int ret = 0;
    const char *curve_name = "prime256v1";
    /* UNCOMPRESSED FORMAT */
    static const unsigned char pub_data[] = {
       POINT_CONVERSION_UNCOMPRESSED,
       0x1b, 0x93, 0x67, 0x55, 0x1c, 0x55, 0x9f, 0x63,
       0xd1, 0x22, 0xa4, 0xd8, 0xd1, 0x0a, 0x60, 0x6d,
       0x02, 0xa5, 0x77, 0x57, 0xc8, 0xa3, 0x47, 0x73,
       0x3a, 0x6a, 0x08, 0x28, 0x39, 0xbd, 0xc9, 0xd2,
       0x80, 0xec, 0xe9, 0xa7, 0x08, 0x29, 0x71, 0x2f,
       0xc9, 0x56, 0x82, 0xee, 0x9a, 0x85, 0x0f, 0x6d,
       0x7f, 0x59, 0x5f, 0x8c, 0xd1, 0x96, 0x0b, 0xdf,
       0x29, 0x3e, 0x49, 0x07, 0x88, 0x3f, 0x9a, 0x29
    };
    static const unsigned char priv_data[] = {
        0x33, 0xd0, 0x43, 0x83, 0xa9, 0x89, 0x56, 0x03,
        0xd2, 0xd7, 0xfe, 0x6b, 0x01, 0x6f, 0xe4, 0x59,
        0xcc, 0x0d, 0x9a, 0x24, 0x6c, 0x86, 0x1b, 0x2e,
        0xdc, 0x4b, 0x4d, 0x35, 0x43, 0xe1, 0x1b, 0xad
    };

     priv = BN_bin2bn(priv_data, sizeof(priv_data), NULL);

     param_bld = OSSL_PARAM_BLD_new();
     if (priv != NULL && param_bld != NULL) {
         if (OSSL_PARAM_BLD_push_utf8_string(param_bld,
                                             OSSL_PKEY_PARAM_GROUP_NAME,
                                             curve_name, 0)
             && OSSL_PARAM_BLD_push_BN(param_bld, OSSL_PKEY_PARAM_PRIV_KEY, priv)
             && OSSL_PARAM_BLD_push_octet_string(param_bld,
                                                 OSSL_PKEY_PARAM_PUB_KEY,
                                                 pub_data, sizeof(pub_data))) {
                 params = OSSL_PARAM_BLD_to_param(param_bld);
         }
     }

     ctx = EVP_PKEY_CTX_new_from_name(NULL, "EC", NULL);
     if (ctx == NULL
         || params == NULL
         || EVP_PKEY_key_fromdata_init(ctx) <= 0
         || EVP_PKEY_fromdata(ctx, &pkey, params) <= 0) {
         ret = 0;
     } else {
         ret = 1;
         /* Do what you want with |pkey| */
     }

     BN_free(priv);
     OSSL_PARAM_BLD_free_params(params);
     OSSL_PARAM_BLD_free(param_bld);
     EVP_PKEY_free(pkey);
     EVP_PKEY_CTX_free(ctx);

    return ret;
}
```

levitte: Side note: for a returned number from `main()`, you've got it mixed up...  `main()` is supposed to return an exit code.

slontis: That is because my test was not done as a main().
It doesnt need to be a main() it could just be a function e.g: int do_key(void)

doc/man3/EVP_PKEY_fromdata.pod
levitte: For all intents and purposes, a BIGNUM is an integer.  I would explain it with ℤ if there weren't for compilers not quite liking UTF-8 in comments...

levitte: I think that the easiest is to remove all but the first sentence of that comment

slontis: I know what you are saying and how it is implemented, but just think of your poor end user.. The thing we actually input in a BIGNUM represented as a sequence of bytes.
THAT WILL BE CONFUSING TO A NEW USER.

doc/man3/EVP_PKEY_fromdata.pod
levitte: I'm really not very keen on fleshing out example code too much, as the actual thing we want to give an example of *drowns* in code.  That being said, adding a few checks didn't make it much worse, so thank you.

doc/man3/EVP_PKEY_fromdata.pod
slontis: I typed the above before seeing your reply.

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -156,14 +156,13 @@ TODO Write a set of cookbook documents and link to them.
...
 
      param_bld = OSSL_PARAM_BLD_new();
-     if (param_bld != NULL) {
-         OSSL_PARAM_BLD_push_utf8_string(param_bld, "curve-name",
-                                         "prime256v1", 0);
-         OSSL_PARAM_BLD_push_BN(param_bld, "priv", priv);
-         OSSL_PARAM_BLD_push_octet_string(param_bld, "pub",
-                                          pub_data, sizeof(pub_data));
+     if (priv != NULL && param_bld != NULL
+         && OSSL_PARAM_BLD_push_utf8_string(param_bld, "curve-name",
```
slontis, line 160: Ok so you missed one very important change..
Which is the 'curve-name' got changed to 'group'. Which is why it is arguably better to use the #defines instead.

slontis: Also - my actual data for the key data works - as opposed to your crappy values that are not valid :)

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -120,52 +120,67 @@ TODO Write a set of cookbook documents and link to them.
...
+                                             pub_data, sizeof(pub_data)))
+         params = OSSL_PARAM_BLD_to_param(param_bld);
+
+     ctx = EVP_PKEY_CTX_new_from_name(NULL, "EC", NULL);
      if (ctx == NULL
+         || params != NULL
          || !EVP_PKEY_key_fromdata_init(ctx)
-         || !EVP_PKEY_fromdata(ctx, &pkey, params))
-         exit(1);
+         || !EVP_PKEY_fromdata(ctx, &pkey, params)) {
```
slontis, line 171: These also can return negative..

doc/man3/EVP_PKEY_fromdata.pod
levitte: Ah, thanks.  I had actually found that by testing the code myself, but apparently forgot to update that string.

doc/man3/EVP_PKEY_fromdata.pod
levitte: Argh

levitte: Fixed, in this and the previous example

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -120,52 +120,67 @@ TODO Write a set of cookbook documents and link to them.
...
  #include <openssl/evp.h>
+ #include <openssl/param_build.h>
 
  /*
-  * These arrays represent large numbers, big endian organization.
-  * In a real application, these would probably be bignums that get
-  * converted to the native integer organization with BN_bn2nativepad().
-  * We're not doing that here, since this is not an example of BIGNUM
-  * functionality, but an example of EVP_PKEY_fromdata().
+  * Fix data to represent the private and public key.
```
paulidale, line 126: Do you mean _Fixed_ here?

doc/man3/EVP_PKEY_fromdata.pod
```diff
@@ -120,52 +120,67 @@ TODO Write a set of cookbook documents and link to them.
...
+     BIGNUM *priv;
+     OSSL_PARAM_BLD *param_bld;
+     OSSL_PARAM *params = NULL;
+     int exitcode = 0;
+
+     priv = BN_bin2bn(priv_data, sizeof(priv_data), NULL);
+
+     param_bld = OSSL_PARAM_BLD_new();
+     if (priv != NULL && param_bld != NULL
+         && OSSL_PARAM_BLD_push_utf8_string(param_bld, "group",
```
romen, line 160: Once we stop using the `OSSL_PARAM`s directly and start using the `OSSL_PARAM_BLD` framework does it still make sense to use `"group"`, `"priv"`, `"pub"` directly instead of the appropriate name definitions?

romen: Is there a sane way of passing these 3 parameters (which are just a string and 2 octet buffers) using `OSSL_PARAM` directly?

It seems from the initial "simplicity" and portability of the `OSSL_PARAM` interface we have diverged quite a lot!

levitte: If you read [OSSL_PARAM(3)](https://github.com/openssl/openssl/blob/master/doc/man3/OSSL_PARAM.pod), you will find a few raw examples.  They should work in C99 and newer at least.

Otherwise, it's perfectly possible, although lengthier, do be even more raw (the same as you would with PKCS#11 templates):

``` C
const OSSL_PARAM params[4];

params[0].key = "curve-name";
params[0].data_type = OSSL_PARAM_UTF8_STRING;
params[0].data = (void *) "prime256v1";
params[0].data_size = sizeof("prime256v1");
params[0].return_size = 0;
params[1].key = "priv";
params[1].data_type = OSSL_PARAM_UNSIGNED_INTEGER;
params[1].data = priv;
params[1].data_size = sizeof(priv);
params[1].return_size = 0;
params[2].key = "pub";
params[2].data_type = OSSL_PARAM_OCTET_STRING;
params[2].data = pub;
params[2].data_size = sizeof(pub);
params[2].return_size = 0;
params[3].key = NULL;
params[3].data_type = 0;
params[3].data = NULL;
params[3].data_size = 0;
params[3].return_size = 0;
```

`OSSL_PARAM` *is*, after all, an open structure.  This is by design.

levitte: > Once we stop using the `OSSL_PARAM`s directly and start using the `OSSL_PARAM_BLD` framework does it still make sense to use `"group"`, `"priv"`, `"pub"` directly instead of the appropriate name definitions?
ren
I'm pushing for this to avoid adding a sense that OSSL_blah_PARAM_something macros are mandatory.  This is to avoid registration hell like we've had around `crypto/objects/objects.txt` and NIDs.

romen: @levitte thanks for the clarification!

slontis: The downside to this is that when someone decides to change a name like "named_curve" to "group" then all of the pod file examples get out of sync. It has happened in quite a few places.
I would suggest this is the entire reason for the defines, and that you are shooting yourself in the foot using the direct names.

doc/man3/EVP_PKEY_fromdata.pod
levitte: Yes.  Corrected

levitte:
Rebased

paulidale (approved):

levitte:
Merged

d9ea62c2c29340818d254ca45111749959236e50 DOC: Modify one example in EVP_PKEY_fromdata(3)
6353507e9d4afe666ade7b8fdf0f0e673f57b36c DOC: Fix check of EVP_PKEY_fromdata{,_init} in examples

PR #12390, opened 2020-07-07
BN: Check endianness in run-time, in BN_native2bn() and BN_bn2nativepad()

### LINKED ISSUE #12387: BN conversion to native requires B_ENDIAN define for big endian machines
In commit fa4d419c25c07b49789df96b32c4a1a85a984fa1 BN_native2bn and BN_bn2nativepad got introduced. It requires that B_ENDIAN is defines for big endian machines. We never had a requirement before that L_ENDIAN or B_ENDIAN is defined, and there are a lot of config targets that have neither defined. It's now assumes that if B_ENDIAN is not defined, that it's little endian.

This is most likely one of the issues seen in #12199.

levitte:
The code relied on B_ENDIAN being defined on all big-endian platform,
which turned out to not always be the case.

Fixes #12387

In doing this, test/ossl_test_endian.h is also moved to include/internal/endian.h
and thereby makes the macros in there our standard way to check endianness
in run-time.

### DIFF
--- a/crypto/bn/bn_lib.c
+++ b/crypto/bn/bn_lib.c
@@ -10,6 +10,7 @@
 #include <assert.h>
 #include <limits.h>
 #include "internal/cryptlib.h"
+#include "internal/endian.h"
 #include "bn_local.h"
 #include <openssl/opensslconf.h>
 #include "internal/constant_time.h"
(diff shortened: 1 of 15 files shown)


kroeckx:
I started on a PR that moves the test/ossl_test_endian.h to include/crypto/, and then use that.

levitte:
> I started on a PR that moves the test/ossl_test_endian.h to include/crypto/, and then use that.

Oooh, nice idea!  I had forgotten about that header file...  so uhm, I assume that was a suggestion for this PR

levitte:
This got refactored quite heavily, following @kroeckx' idea.  Needs re-review

levitte:
Merged

e23d850ff3281220f33ed78d9ca4fcadfa279565 Add and use internal header that implements endianness check
310a0edbd003dd7c580ae3cf78f1782b2c3d9ded BN: Check endianness in run-time, in BN_native2bn() and BN_bn2nativepad()

PR #12399, opened 2020-07-09
ocsp.h: Fix backward compatibility decl for OCSP_parse_url() by including http.h

### LINKED ISSUE #12386: poor backward compatibility of OCSP_parse_url()
(as of 3a19f1a9dd16f5855a7e2a42cd3e7f136f849bd3)

We have in `include/openssl/ocsp.h`

```c
#  define OCSP_parse_url OSSL_HTTP_parse_url /* for backward compatibility */
```

But `OSSL_HTTP_parse_url` requires including `<openssl/http.h>`, which is new in 3.0.0. So code can't really use `OCSP_parse_url` in a backward-compatible way without additional `#ifdef`s.

A simple fix might be to include `openssl/http.h` from `openssl/ocsp.h`.

DDvO:
Fixes #12386

### DIFF
--- a/include/openssl/ocsp.h
+++ b/include/openssl/ocsp.h
@@ -17,6 +17,7 @@
 # endif
 
 # include <openssl/opensslconf.h>
+# include <openssl/http.h> /* for OSSL_HTTP_parse_url */
 
 /*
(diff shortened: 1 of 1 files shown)


DDvO:
Merged

PR #11195, opened 2020-02-27
test/drbgtest.c: Fix error check test

ciz:
The condition in test_error_checks() was inverted, so the test succeeded
as long as error_check() failed. Incidently, error_check() contained
several bugs that assured it always failed, thus giving overall drbg
test success.

- [x] tests are added or updated

PR #12417, opened 2020-07-10
test/drbgtest.c: Fix error check test

ciz:
The condition in test_error_checks() was inverted, so the test succeeded
as long as error_check() failed. Incidently, error_check() contained
several bugs that assured it always failed, thus giving overall drbg
test success.

Remove the broken explicit zero check.
RAND_DRBG_uninstantiate() cleanses the data via drbg_ctr_uninstantiate(),
but right after that it resets drbg->data.ctr using RAND_DRBG_set(),
so TEST_mem_eq(zero, sizeof(drbg->data)) always failed.

(backport from https://github.com/openssl/openssl/pull/11195)

Signed-off-by: Vitezslav Cizek <vcizek@suse.com>

### DIFF
--- a/test/drbgtest.c
+++ b/test/drbgtest.c
@@ -286,7 +286,6 @@ static int instantiate(RAND_DRBG *drbg, DRBG_SELFTEST_DATA *td,
  */
 static int error_check(DRBG_SELFTEST_DATA *td)
 {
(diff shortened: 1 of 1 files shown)


mspncp:
Merged to 1.1.1. in 72a9868fe53e01bfe912a56b56e4474ff6e0a063, thanks!

PR #12418, opened 2020-07-10
Providerized libssl fallout: cleanup init

kaduk:
Since libssl is entirely using fetched cipher/digest implementations
from providers, we don't need to register the libcrypto cipher/digest
implementations in ossl_init_ssl_base().


levitte:
Someone forgot to close this :wink:

PR #12427, opened 2020-07-12
Fix memory leaks in tests

szmate1618:
CLA: trivial

This pull request aims to fix memory leaks in some tests. Leaky tests are annoying, because you can never be 100%
sure if it's the test or the actual library code which leaks memory, until you completely fix the problem.
I found them with the help of a proprietary memory leak detection tool, but in retrospect I think most of them are quite visible.
For example, I'm no OpenSSL expert, but if I think about it, it's very obvious that if we create a `ctx` and a `ctx2` context we will have to cleanup both of them.
In case you cannot accept this PR without a publicly available software assisted method of validating it, can you please point me to some free memory leak detection tools? I tried [https://drmemory.org/](url), and it seems to find a few memory issues (like use of unintialized memory), but none of them are leaks.

### DIFF
--- a/crypto/bn/bntest.c
+++ b/crypto/bn/bntest.c
@@ -345,6 +345,9 @@ int main(int argc, char *argv[])
     (void)BIO_flush(out);
 
     BN_CTX_free(ctx);
+    ERR_remove_thread_state(NULL);
+    CRYPTO_cleanup_all_ex_data();
+    CRYPTO_mem_leaks(out);
     BIO_free(out);
 
     EXIT(0);
--- a/crypto/bn/exptest.c
+++ b/crypto/bn/exptest.c
@@ -293,7 +293,9 @@ int main(int argc, char *argv[])
     BN_free(b);
     BN_free(m);
     BN_CTX_free(ctx);
+    ERR_free_strings();
     ERR_remove_thread_state(NULL);
+    CRYPTO_cleanup_all_ex_data();
     CRYPTO_mem_leaks(out);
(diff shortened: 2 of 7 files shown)


szmate1618:
I think it's trivial in the sense that I had zero degree of freedom and no 'creative control'.
For every leak I had to call 1 specific function, no more, no less (yes, in some files I added 3 new function calls, but that was for 3 separate leaks), and I had to call it in a specific place (between 2 other specific cleanup function), because the safest cleanup order is the reversed allocation order.
But if you disagree, I can, of course, sign a CLA.

beldmit:
@mattcaswell, could you please take a look - at least compression methods should be cleaned in atexit handlers...

slontis:
@beldmit did you mean to put the master tag? It says 1_0_2-stable at the top.

beldmit:
> @beldmit did you mean to put the master tag? It says 1_0_2-stable at the top.

Hmmm... I don't see the corresponding files in the 1_0_2 branch. But anyway it's my fault.

@szmate1618, which branch do you mean? 1.0.2 is out of support.

slontis:
@mattcaswell FYI..

mattcaswell:
This appears to address 1.0.2 issues which won't be fixed because 1.0.2 is out of (public) support. It appears to me that these issues won't happen in later versions because the auto-deinit functionality should handle most of it - with the exception of the hmactest change. Looking at master that issue has already been fixed there.

slontis:
Can this be closed?

PR #12429, opened 2020-07-13
Fixed EVP_MAC_final argument count in example

### LINKED ISSUE #12424: EVP_MAC.pod example uses wrong final signature
The [EVP_MAC.pod](https://github.com/openssl/openssl/blob/master/doc/man3/EVP_MAC.pod) documentation file uses the wrong ``EVP_MAC_final`` signature with only 3 parameters in the example:
``if (!EVP_MAC_final(ctx, buf, &final_l))``

The signature should be:
```
                   unsigned char *out, size_t *outl, size_t outsize);
```

ozppupbg:
EVP_MAC_final had only three arguments / the buffer/tag size was missing.
Fixes #12424

Note, that I didn't try to compile the example to look for other problems.

### DIFF
--- a/doc/man3/EVP_MAC.pod
+++ b/doc/man3/EVP_MAC.pod
@@ -339,7 +339,7 @@ EVP_MAC_do_all_provided() returns nothing at all.
               goto err;
       }
 
-      if (!EVP_MAC_final(ctx, buf, &final_l))
+      if (!EVP_MAC_final(ctx, buf, &final_l, sizeof(buf)))
           goto err;
 
       printf("Result: ");
--- a/doc/man3/EVP_MAC.pod
+++ b/doc/man3/EVP_MAC.pod
@@ -307,7 +307,7 @@ EVP_MAC_do_all_provided() returns nothing at all.
       EVP_MAC_CTX *ctx = NULL;
 
       unsigned char buf[4096];
-      ssize_t read_l;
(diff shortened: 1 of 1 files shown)


beldmit:
Could you please either submit the CLA or add the line "CLA: trivial" to your PR?

ozppupbg:
I have already sent my CLA

t8m:
As these are unrelated errors in the example I would not prevent merging this change. Of course if you fix those other errors as well, it would be nice.

ozppupbg:
I'll give it a try, but I have to set up some stuff  first

doc/man3/EVP_MAC.pod
```diff
@@ -317,12 +317,12 @@ EVP_MAC_do_all_provided() returns nothing at all.
 
       if (cipher != NULL)
           params[params_n++] =
-              OSSL_PARAM_construct_utf8_string("cipher", cipher, 0;
+              OSSL_PARAM_construct_utf8_string("cipher", (char*)cipher, 0);
```
slontis, line 320: Just a small NIT: Can you add an extra space in the 3 casts e.g. (char *)

slontis: @ozppupbg ?

paulidale:
Merged to master after fixing the missing space in the casts.

PR #12430, opened 2020-07-13
Fix for the bug in ssl_version_supported()

OlegPekar:
Bug in ssl_version_supported() - the chosen SSL_METHOD returned from the function is always a client side method, ever for the server side.

This is a typical sequence for TLS server side code to reproduce the bug:
1. Create new SSL_CTX_new( TLS_server_method() );
2. Call SSL_accept() that calls SSL_set_accept_state() that sets s->handshake_func = s->method->ssl_accept; i.e. initialises handshake_func of SSL* to the correct ssl_accept function of TLS_server_method()
3. Afterwards in the flow ssl_choose_server_version() is called that calls ssl_version_supported() that changes s->method to the method of the best suitable version but always to the client method, even for the server side
4. Therefore say we are using min and max TLS version TLS 1.2 - for the server side SSL* we have get s->method == tlsv1_2_client_method() nevertheless s->server == 1
This is not harmful meanwhile but could cause a bug in the further code development.

##### Checklist
- [x] tests are added

### DIFF
--- a/ssl/statem/statem_lib.c
+++ b/ssl/statem/statem_lib.c
@@ -1609,14 +1609,16 @@ int ssl_version_supported(const SSL *s, int version, const SSL_METHOD **meth)
     for (vent = table;
          vent->version != 0 && version_cmp(s, version, vent->version) <= 0;
          ++vent) {
-        if (vent->cmeth != NULL
+        const SSL_METHOD *(*ret_meth) (void);
+        ret_meth = s->server ? vent->smeth : vent->cmeth;
+        if ( ret_meth != NULL
                 && version_cmp(s, version, vent->version) == 0
-                && ssl_method_error(s, vent->cmeth()) == 0
+                && ssl_method_error(s, ret_meth()) == 0
                 && (!s->server
                     || version != TLS1_3_VERSION
                     || is_tls13_capable(s))) {
             if (meth != NULL)
-                *meth = vent->cmeth();
+                *meth = ret_meth();
             return 1;
         }
     }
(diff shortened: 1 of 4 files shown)


test/method_test.c
```diff
@@ -0,0 +1,82 @@
...
+	{ TLS1_VERSION, TLS1_3_VERSION, NULL, NULL }};
+
+static int test_method(int test)
+{
+    SSL_CTX *cctx = NULL, *sctx = NULL;
+    SSL *clientssl = NULL, *serverssl = NULL;
+    int testresult = 0;
+
+    if (!TEST_true(create_ssl_ctx_pair(NULL, TLS_server_method(), TLS_client_method(), subtests[test].min_version, subtests[test].max_version, &sctx, &cctx, cert, pvk))
+     || !TEST_true(create_ssl_objects(sctx, cctx, &serverssl, &clientssl, NULL, NULL)))
```
beldmit, line 42: Could you please split these lines and similar too long lines below?

ssl/statem/statem_lib.c
```diff
@@ -1609,14 +1609,16 @@ int ssl_version_supported(const SSL *s, int version, const SSL_METHOD **meth)
     for (vent = table;
          vent->version != 0 && version_cmp(s, version, vent->version) <= 0;
          ++vent) {
-        if (vent->cmeth != NULL
+        const SSL_METHOD *(*ret_meth) (void);
```
kaduk, line 1612: It may be simpler to just make `ret_meth` a `SSL_METHOD *` directly, and only call the function once, on the next line.

test/method_test.c
OlegPekar: Done!

ssl/statem/statem_lib.c
OlegPekar: The table entry contains function pointers:
```
typedef struct {
    int version;
    const SSL_METHOD *(*cmeth) (void);
    const SSL_METHOD *(*smeth) (void);
} version_info;
```

So we need first to check that the function pointer is not null before calling the function that returns a method. Just like it was before the fix.

mattcaswell:
Travis errors are relevant.

Also, please could you submit a CLA?

Issue #12432, opened 2020-07-13
1.1.1 build is broken


romen:
Backport of #12096 to `1.1.1` broke the build as the following functions are missing:

~~~C
const BIGNUM *EC_GROUP_get0_field(const EC_GROUP *group);
int EC_GROUP_get_field_type(const EC_GROUP *group);
~~~

This was also reported on [openssl-users][openssl-users msg].

[openssl-users msg]: https://www.mail-archive.com/openssl-users@openssl.org/msg88436.html

romen:
Fixed by:

- a5b8c19cda `[1.1.1][test] Avoid missing EC_GROUP wrappers`

PR #12433, opened 2020-07-13
[1.1.1][test] Avoid missing EC_GROUP wrappers

romen:
Backport of https://github.com/openssl/openssl/pull/12096 to 1.1.1 broke
the build as the following functions are missing:

    const BIGNUM *EC_GROUP_get0_field(const EC_GROUP *group);
    int EC_GROUP_get_field_type(const EC_GROUP *group);

Turns out that for the purposes of the test code, we don't really need
to differentiate between prime and binary fields, and we can directly
use the existing `EC_GROUP_get_degree()` in the same fashion as was
being done for binary fields also for prime fields.

Fixes https://github.com/openssl/openssl/issues/12432

##### Checklist

### DIFF
--- a/test/ectest.c
+++ b/test/ectest.c
@@ -2123,9 +2123,7 @@ static int custom_generator_test(int id)
         goto err;
 
     /* expected byte length of encoded points */
-    bsize = (EC_GROUP_get_field_type(group) == NID_X9_62_prime_field) ?
-             BN_num_bytes(EC_GROUP_get0_field(group)) :
-             (EC_GROUP_get_degree(group) + 7) / 8;
+    bsize = (EC_GROUP_get_degree(group) + 7) / 8;
     bsize = 2 * bsize + 1;
 
     if (!TEST_ptr(k = BN_CTX_get(ctx))


beldmit:
My fault :(

romen:
> My fault :(

Actually not, it was my fault, I endorsed backporting to 1.1.1 as cherry-picking cleanly, but clearly my dev worktree must have been in some weird state to not trigger build failures!

test/ectest.c
```diff
@@ -2123,9 +2123,7 @@ static int custom_generator_test(int id)
         goto err;
 
     /* expected byte length of encoded points */
-    bsize = (EC_GROUP_get_field_type(group) == NID_X9_62_prime_field) ?
-             BN_num_bytes(EC_GROUP_get0_field(group)) :
-             (EC_GROUP_get_degree(group) + 7) / 8;
+    bsize = (EC_GROUP_get_degree(group) + 7) / 8;
```
romen, line 2126: @bbbrumley Am I missing why (other than for clarity) you did not use `EC_GROUP_get_degree()` in both cases?

bbbrumley: Not really. It started out with no `EC_GROUP_get_degree` or branches at all but then failed for some old binary curves. So then I went digging for how to make that work and didn't back up to reconsider. (I would not have expected `EC_GROUP_get_degree` to do anything sane for GFp curves, but looks like I'm wrong.)

romen:
Merged to `1.1.1` as:

- a5b8c19cda `[1.1.1][test] Avoid missing EC_GROUP wrappers`


Thanks!

PR #12434, opened 2020-07-13
Prefix crlNumber output with 0x.

jsha:
When the crlNumber field contains only 0-9 digits, the output is
ambiguous as to what base it's in, which can be confusing. Adding this
prefix makes it explicit that it's in hex.

### DIFF
--- a/apps/crl.c
+++ b/apps/crl.c
@@ -288,6 +288,7 @@ int crl_main(int argc, char **argv)
                 crlnum = X509_CRL_get_ext_d2i(x, NID_crl_number, NULL, NULL);
                 BIO_printf(bio_out, "crlNumber=");
                 if (crlnum) {
+                    BIO_puts(bio_out, "0x");
                     i2a_ASN1_INTEGER(bio_out, crlnum);
                     ASN1_INTEGER_free(crlnum);
                 } else


apps/crl.c
```diff
@@ -286,7 +286,7 @@ int crl_main(int argc, char **argv)
             if (crlnumber == i) {
                 ASN1_INTEGER *crlnum;
                 crlnum = X509_CRL_get_ext_d2i(x, NID_crl_number, NULL, NULL);
-                BIO_printf(bio_out, "crlNumber=");
+                BIO_printf(bio_out, "crlNumber=0x");
```
slontis, line 289: You would need to do a seperate BIO_printf() inside the if otherwise we will end up with..
crlNumber=0x<None>

jsha:
Hi! I'm having trouble getting my printer to work in order to print out the CLA agreement. Is there any possibility to waive the CLA requirement? I can't imagine that I actually hold any copyright interest in this single line of code that doesn't do much special.

kaduk:
I am not privy to the details of how it is done, but I know that in several cases electronic signatures (no paper required) have been used for CLA submissions.

richsalz:
@jsha: amend your commit to put "CLA: trivial" in the text of the commit, I'm sure this is trivial.

jsha:
Thanks for the tip, @richsalz! Done.

paulidale:
I'm fine calling this trivial.

beldmit (approved):
LGTM. I also agree it's trivial.

paulidale:
Merged to master, thanks for this.

PR #12435, opened 2020-07-13
doc: Fix documentation of EVP_EncryptUpdate().

paulidale:
The documentation was off by one for the length this function could return.
The wrap functions could exceed the limit by one.

### DIFF
--- a/doc/man3/EVP_EncryptInit.pod
+++ b/doc/man3/EVP_EncryptInit.pod
@@ -210,10 +210,15 @@ specified.
 EVP_EncryptUpdate() encrypts B<inl> bytes from the buffer B<in> and
 writes the encrypted version to B<out>. This function can be called
 multiple times to encrypt successive blocks of data. The amount
-of data written depends on the block alignment of the encrypted data:
-as a result the amount of data written may be anything from zero bytes
-to (inl + cipher_block_size) bytes so B<out> should contain sufficient
-room. The actual number of bytes written is placed in B<outl>. It also
+of data written depends on the block alignment of the encrypted data.
+For most ciphers and modes, the amount of data written can be anything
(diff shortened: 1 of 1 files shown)


kroeckx:
Do you think it's useful to have the documentation say it's different for wrap ciphers?

t8m:
> Do you think it's useful to have the documentation say it's different for wrap ciphers?

Yeah, unless we can come up with some plausible case for other ciphers where it would happen as well?

Because otherwise we are basically making most existing uses of EVP_EncryptUpdate non-conforming to the documentation. Which is kind-of API break. Inevitable for wrap ciphers, yes, but not for the others.

kroeckx:
I guess the other exception is the stream ciphers.

Maybe it's also useful to document if you pass it a multiple of
the block size, you get exactly that amount back.

paulidale:
> I guess the other exception is the stream ciphers. Maybe it's also useful to document if you pass it a multiple of the block size, you get exactly that amount back.

Is this true for the wrap ciphers??  The `memmove(out + 8, in, inlen)` in `CRYPTO_128_wrap()` seems to indicate otherwise.

t8m:
I think Kurt meant regular block cipher modes.

So I believe we should document all the possible cases if we really want to make the documentation truly helpful and the API contract reasonable. It would be pretty bad if we required for all EVP_EncryptUpdate() calls to reserve inlen rounded up to blocksize + blocksize. Even if the underlying cipher mode is stream one and the extra blocksize is needed only for the wrap modes.

paulidale:
I've added statements for _most ciphers_, _wrapping_ and _stream ciphers_.

Ideally, we should document the behaviour for each cipher mode individually.

paulidale:
Merged to 1.1.1 and master.  Thanks for the feedback and suggestions.

PR #12450, opened 2020-07-14
capabilities: make capability selection case insensitive.

paulidale:
Everything else to do with algorithm selection and properties is case insensitive.


levitte:
Merged

ecca5b6e2ea5f364e4281193fd1526fbaf3f8248 capabilities: make capability selection case insensitive.

paulidale:
Thanks for merging this.

PR #12452, opened 2020-07-15
util/find-doc-nits: read full declarations as one line in name_synopsis()

levitte:
name_synopsis was reading physical SYNOPSIS lines.  This changes it to
consider a declaration at a time, so we treat a C declaration that's
been broken up in several lines as one.

This makes it mandatory to end all C declarations in the SYNOPSIS with
a semicolon.  Those can be detected in two ways:

1.  Parsing an individual .pod file outputs this error:

    doc/man3/SOMETHING.pod:1: Can't parse rest of synopsis:

     int SOMETHING_status(SOMETHING *s)
     int SOMETHING_start(SOMETHING *s)

    (declarations not ending with a semicolon (;)?)

2.  Errors like this:

    doc/man3/SOMETHING.pod:1: SOMETHING_status missing from SYNOPSIS
    doc/man3/SOMETHING.pod:1: SOMETHING_start missing from SYNOPSIS

### DIFF
--- a/doc/man3/OpenSSL_version.pod
+++ b/doc/man3/OpenSSL_version.pod
@@ -17,8 +17,11 @@ OPENSSL_VERSION_NUMBER, OpenSSL_version_num, OPENSSL_info
  #define OPENSSL_VERSION_MAJOR  x
  #define OPENSSL_VERSION_MINOR  y
  #define OPENSSL_VERSION_PATCH  z
+
+ /* The definitions here are typical release values */
  #define OPENSSL_VERSION_PRE_RELEASE ""
  #define OPENSSL_VERSION_BUILD_METADATA ""
+
  #define OPENSSL_VERSION_TEXT "OpenSSL x.y.z xx XXX xxxx"
 
  #include <openssl/crypto.h>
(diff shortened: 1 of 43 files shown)


levitte:
Ping @slontis, you were willing to help out with the .pod file fixes.  It should be easy, just add semicolons on declarations that lack them

slontis:
@levitte in OpenSSL_version.pod there were multiple comments and blanklines that caused issues with the change to the script. If you dont like the changes in that file revert them and allow these extra fields to be ignored.

doc/man3/OpenSSL_version.pod
```diff
@@ -17,11 +17,8 @@ OPENSSL_VERSION_NUMBER, OpenSSL_version_num, OPENSSL_info
  #define OPENSSL_VERSION_MAJOR  x
  #define OPENSSL_VERSION_MINOR  y
  #define OPENSSL_VERSION_PATCH  z
-
- /* The definitions here are typical release values */
```
levitte, line 21: Oh!  Yeahok, that's something I didn't anticipate in my change.  I absolutely think that comments should be allowed and remain.

levitte:
> @levitte in OpenSSL_version.pod there were multiple comments and blanklines that caused issues with the change to the script. If you dont like the changes in that file revert them and allow these extra fields to be ignored.

Noted, and will do.  I'm actually surprised that the empty lines caused a problem...  they shouldn't, but I'll have a look.

levitte:
Ping?

levitte:
> LGTM - But I am unsure if my approval counts on this one.

What happens is that you approve my parts and I approve yours (which I do), so it's all good in the end.

levitte:
Merged

d3cb5904f3ca62f23c95248f951477d322159bbb util/find-doc-nits: read full declarations as one line in name_synopsis()
93e32043cbf75d0802bca9782e61a241acb1ec2d util/find-doc-nits: relax some SYNOPSIS checks
f64f17c3e011a00febed2bf6bc17ca7e030fc69e Added missing ';' after methods in the synopsis section of pod files

Issue #10086, opened 2019-10-03
05-test_rand.t failing for debug builds


kimaluco17:
Whenever I build with --debug or debug-VC-WIN32, one of the tests (looks like drbg_cavs_test) in 05-test_rand.t fails. This does not happen when building release version. I've attached a log file with the results of the build and the tests.

I'm using this configuration:
`perl Configure VC-WIN32 no-asm --prefix=%prefix_base%%prefix_ext% --debug`

[test_debug.log](https://github.com/openssl/openssl/files/3686594/test_debug.log)

Is there something that I'm missing to make it pass?

mattcaswell:
Please try running:

````
nmake TESTS=test_rand V=1 test
````
And then post the output

kimaluco17:
Thanks for the quick response, here's the output.

[test.log](https://github.com/openssl/openssl/files/3687136/test.log)

mattcaswell:
These two lines seem somewhat surprising:

````
OPENSSL_Uplink(5A962350,08): no OPENSSL_Applink
F:\development\OpenSSL\openssl-1.1.1d\test\drbg_cavs_test.exe => 1
````

@levitte - any thoughts?

paulidale:
@slontis ???

mspncp:
FWIW: I was not able to reproduce the issue with OpenSSL_1_1_1-stable (currently at a80de34b27) using Visual Studio 2019, neither with the x86 native tools command prompt and `perl Configure VC-WIN32 no-asm`, nor with the x64 native tools command prompt and `perl Configure VC-WIN64A no-asm`.

kimaluco17:
I forgot to mention that I'm using 1.1.1d.

As a sanity check, I will try building a fresh debug build again to see if I get the same result. But I still don't understand why the release configuration would pass.

mspncp:
> I forgot to mention that I'm using 1.1.1d.

I can retry with the tagged commit, but I doubt it will make a difference.

mspncp:
`nmake TESTS=test_rand V=1 test` succeeds also with OpenSSL_1_1_1d (commit 894da2fb7ed5d314ee5c2fc9fd2d9b8b74111596) with Visual Studio 2019, x86 native tools command prompt.

kimaluco17:
Still getting the same result, I am running it from a VS 2017 x86 developer command prompt.

I've also tried building with the latest version of NASM, no difference there.

mspncp:
[...]

PR #12457, opened 2020-07-16
[1.1.1] Fix d2i_ECPKParameters_fp and i2d_ECPKParameters_fp macroes

### LINKED ISSUE #12443: d2i_ECPKParameters_fp throws lots of warnings
I think some of [these macros](https://github.com/openssl/openssl/blob/ce3080e931d77fda3bb4f2d923fcc6cec967d1a3/include/openssl/ec.h#L836) are stale. Below is master but I get the same warnings in 1.1.1, although the code seems to runs fine.

MWE:
```C
#include <openssl/ec.h>
#include <assert.h>

int main(int argc, char **argv) {
    EC_GROUP *group = NULL;
    FILE *fp = NULL;

    assert((fp = fopen("params.der", "rb")) != NULL);
    assert((group = d2i_ECPKParameters_fp(fp, NULL)) != NULL);
    fclose(fp);
    EC_GROUP_free(group);

    return 0;
}
```

```
bbrumley@arizona:/tmp$ rm foo; cc -o foo foo.c -lcrypto -L$HOME/.local/ssl/lib -I$HOME/.local/ssl/include
In file included from foo.c:2:0:
foo.c: In function ‘main’:
/home/bbrumley/.local/ssl/include/openssl/ec.h:841:35: warning: passing argument 2 of ‘ASN1_d2i_fp’ from incompatible pointer type [-Wincompatible-pointer-types]
     (EC_GROUP *)ASN1_d2i_fp(NULL, (char *(*)())d2i_ECPKParameters, (fp), \
                                   ^
foo.c:9:21: note: in expansion of macro ‘d2i_ECPKParameters_fp’
     assert((group = d2i_ECPKParameters_fp(fp, NULL)) != NULL);
                     ^~~~~~~~~~~~~~~~~~~~~
In file included from /home/bbrumley/.local/ssl/include/openssl/ec.h:23:0,
                 from foo.c:1:
/home/bbrumley/.local/ssl/include/openssl/asn1.h:689:7: note: expected ‘void * (*)(void **, const unsigned char **, long int)’ but argument is of type ‘char * (*)()’
 void *ASN1_d2i_fp(void *(*xnew) (void), d2i_of_void *d2i, FILE *in, void **x);
       ^~~~~~~~~~~
In file included from foo.c:2:0:
/home/bbrumley/.local/ssl/include/openssl/ec.h:842:29: warning: passing argument 4 of ‘ASN1_d2i_fp’ from incompatible pointer type [-Wincompatible-pointer-types]
                             (unsigned char **)(x))
                             ^
foo.c:9:21: note: in expansion of macro ‘d2i_ECPKParameters_fp’
     assert((group = d2i_ECPKParameters_fp(fp, NULL)) != NULL);
                     ^~~~~~~~~~~~~~~~~~~~~
In file included from /home/bbrumley/.local/ssl/include/openssl/ec.h:23:0,
                 from foo.c:1:
/home/bbrumley/.local/ssl/include/openssl/asn1.h:689:7: note: expected ‘void **’ but argument is of type ‘unsigned char **’
 void *ASN1_d2i_fp(void *(*xnew) (void), d2i_of_void *d2i, FILE *in, void **x);
       ^~~~~~~~~~~
In file included from foo.c:2:0:
/home/bbrumley/.local/ssl/include/openssl/ec.h:841:35: warning: passing argument 2 of ‘ASN1_d2i_fp’ from incompatible pointer type [-Wincompatible-pointer-types]
     (EC_GROUP *)ASN1_d2i_fp(NULL, (char *(*)())d2i_ECPKParameters, (fp), \
                                   ^
foo.c:9:21: note: in expansion of macro ‘d2i_ECPKParameters_fp’
     assert((group = d2i_ECPKParameters_fp(fp, NULL)) != NULL);
                     ^~~~~~~~~~~~~~~~~~~~~
In file included from /home/bbrumley/.local/ssl/include/openssl/ec.h:23:0,
                 from foo.c:1:
/home/bbrumley/.local/ssl/include/openssl/asn1.h:689:7: note: expected ‘void * (*)(void **, const unsigned char **, long int)’ but argument is of type ‘char * (*)()’
 void *ASN1_d2i_fp(void *(*xnew) (void), d2i_of_void *d2i, FILE *in, void **x);
       ^~~~~~~~~~~
In file included from foo.c:2:0:
/home/bbrumley/.local/ssl/include/openssl/ec.h:842:29: warning: passing argument 4 of ‘ASN1_d2i_fp’ from incompatible pointer type [-Wincompatible-pointer-types]
                             (unsigned char **)(x))
                             ^
foo.c:9:21: note: in expansion of macro ‘d2i_ECPKParameters_fp’
     assert((group = d2i_ECPKParameters_fp(fp, NULL)) != NULL);
                     ^~~~~~~~~~~~~~~~~~~~~
In file included from /home/bbrumley/.local/ssl/include/openssl/ec.h:23:0,
                 from foo.c:1:
/home/bbrumley/.local/ssl/include/openssl/asn1.h:689:7: note: expected ‘void **’ but argument is of type ‘unsigned char **’
 void *ASN1_d2i_fp(void *(*xnew) (void), d2i_of_void *d2i, FILE *in, void **x);
       ^~~~~~~~~~~
```

romen:
These functions are part of the public API but we don't have tests covering their usage.
They are actually implemented as macroes and the absence of tests has caused them to fall out-of-sync with the latest changes to ASN1 related functions and cause compilation warnings.

This PR fixes the public headers to reflect these changes.

Fixes #12443

### `1.1.1` first

Unusually I am proposing to fix first this issue as a `1.1.1` issue, and then forward-port the changes to `master` (see #16355). The reason is that master has inherited this bug from `1.1.1` and has compounded it with the current inability of handling serialization with the `OPENSSL_EC_EXPLICIT_CURVE` ASN1 flag.

This means that anyway we would have to review the changes separately as the cherry-picks are not trivial, and it makes sense to address this issue in `1.1.1` first and then apply incremental changes to adapt to the new architecture in `master`.

### Additional related commits included in this PR

- `Add tests for i2d_TYPE_fp and d2i_TYPE_fp`

  This commit add tesst to trigger the issue and make sure these macroes have at least a caller in our test suite to avoid this to happen again in the future.
  
  At the moment the test only covers ECPKParameters as a type, but this should probably be extended similarly to DH, DSA and (maybe) RSA equivalents.

- `[ec] Do not default to OPENSSL_EC_NAMED_CURVE for curves without OID`

  Some curves don't have an associated OID: for those we should not default to `OPENSSL_EC_NAMED_CURVE` encoding of parameters and instead set the ASN1 flag to `OPENSSL_EC_EXPLICIT_CURVE`.
  
  This is a follow-up to https://github.com/openssl/openssl/pull/12312

### Checklist

### DIFF
--- a/crypto/ec/ec_curve.c
+++ b/crypto/ec/ec_curve.c
@@ -12,6 +12,7 @@
 #include "ec_local.h"
 #include <openssl/err.h>
 #include <openssl/obj_mac.h>
+#include <openssl/objects.h>
 #include <openssl/opensslconf.h>
 #include "internal/nelem.h"
 
@@ -3097,6 +3098,32 @@ static EC_GROUP *ec_group_new_from_data(const ec_list_element curve)
             goto err;
         }
     }
+
+    if (EC_GROUP_get_asn1_flag(group) == OPENSSL_EC_NAMED_CURVE) {
+        /*
+         * Some curves don't have an associated OID: for those we should not
+         * default to `OPENSSL_EC_NAMED_CURVE` encoding of parameters and
+         * instead set the ASN1 flag to `OPENSSL_EC_EXPLICIT_CURVE`.
+         *
+         * Note that `OPENSSL_EC_NAMED_CURVE` is set as the default ASN1 flag on
+         * `EC_GROUP_new()`, when we don't have enough elements to determine if
+         * an OID for the curve name actually exists.
+         * We could implement this check on `EC_GROUP_set_curve_name()` but
+         * overloading the simple setter with this lookup could have a negative
+         * performance impact and unexpected consequences.
+         */
+        ASN1_OBJECT *asn1obj = OBJ_nid2obj(curve.nid);
+
+        if (asn1obj == NULL) {
+            ECerr(EC_F_EC_GROUP_NEW_FROM_DATA, ERR_R_OBJ_LIB);
+            goto err;
+        }
+        if (OBJ_length(asn1obj) == 0)
+            EC_GROUP_set_asn1_flag(group, OPENSSL_EC_EXPLICIT_CURVE);
+
+        ASN1_OBJECT_free(asn1obj);
+    }
+
     ok = 1;
  err:
     if (!ok) {
--- a/test/ec_internal_test.c
+++ b/test/ec_internal_test.c
@@ -283,6 +283,47 @@ static int decoded_flag_test(void)
     return testresult;
 }
 
+static
+int ecpkparams_i2d2i_test(int n)
+{
+    EC_GROUP *g1 = NULL, *g2 = NULL;
+    FILE *fp = NULL;
+    int nid = curves[n].nid;
+    int testresult = 0;
+
+    /* create group */
+    if (!TEST_ptr(g1 = EC_GROUP_new_by_curve_name(nid)))
+        goto end;
+
+    /* encode params to file */
+    if (!TEST_ptr(fp = fopen("params.der", "wb"))
+            || !TEST_true(i2d_ECPKParameters_fp(fp, g1)))
+        goto end;
+
+    /* flush and close file */
+    if (!TEST_int_eq(fclose(fp), 0)) {
+        fp = NULL;
+        goto end;
+    }
+    fp = NULL;
+
+    /* decode params from file */
+    if (!TEST_ptr(fp = fopen("params.der", "rb"))
+            || !TEST_ptr(g2 = d2i_ECPKParameters_fp(fp, NULL)))
+        goto end;
+
+    testresult = 1; /* PASS */
+
+end:
+    if (fp != NULL)
+        fclose(fp);
+
+    EC_GROUP_free(g1);
+    EC_GROUP_free(g2);
+
+    return testresult;
+}
+
 int setup_tests(void)
 {
     crv_len = EC_get_builtin_curves(NULL, 0);
@@ -297,6 +338,8 @@ int setup_tests(void)
 #endif
     ADD_ALL_TESTS(field_tests_default, crv_len);
     ADD_TEST(decoded_flag_test);
+    ADD_ALL_TESTS(ecpkparams_i2d2i_test, crv_len);
+
     return 1;
 }
 
--- a/crypto/ec/ec_asn1.c
+++ b/crypto/ec/ec_asn1.c
@@ -548,7 +548,7 @@ ECPKPARAMETERS *EC_GROUP_get_ecpkparameters(const EC_GROUP *group,
             ECPARAMETERS_free(ret->value.parameters);
     }
 
-    if (EC_GROUP_get_asn1_flag(group)) {
+    if (EC_GROUP_get_asn1_flag(group) == OPENSSL_EC_NAMED_CURVE) {
         /*
          * use the asn1 OID to describe the elliptic curve parameters
          */
--- a/include/openssl/ec.h
+++ b/include/openssl/ec.h
@@ -793,12 +793,15 @@ int EC_GROUP_get_pentanomial_basis(const EC_GROUP *, unsigned int *k1,
 EC_GROUP *d2i_ECPKParameters(EC_GROUP **, const unsigned char **in, long len);
 int i2d_ECPKParameters(const EC_GROUP *, unsigned char **out);
 
-# define d2i_ECPKParameters_bio(bp,x) ASN1_d2i_bio_of(EC_GROUP,NULL,d2i_ECPKParameters,bp,x)
-# define i2d_ECPKParameters_bio(bp,x) ASN1_i2d_bio_of_const(EC_GROUP,i2d_ECPKParameters,bp,x)
-# define d2i_ECPKParameters_fp(fp,x) (EC_GROUP *)ASN1_d2i_fp(NULL, \
-                (char *(*)())d2i_ECPKParameters,(fp),(unsigned char **)(x))
-# define i2d_ECPKParameters_fp(fp,x) ASN1_i2d_fp(i2d_ECPKParameters,(fp), \
-                (unsigned char *)(x))
+# define d2i_ECPKParameters_bio(bp,x) \
+    ASN1_d2i_bio_of(EC_GROUP, NULL, d2i_ECPKParameters, bp, x)
+# define i2d_ECPKParameters_bio(bp,x) \
+    ASN1_i2d_bio_of_const(EC_GROUP, i2d_ECPKParameters, bp, x)
+# define d2i_ECPKParameters_fp(fp,x) \
+    (EC_GROUP *)ASN1_d2i_fp(NULL, (d2i_of_void *)d2i_ECPKParameters, (fp), \
+                            (void **)(x))
+# define i2d_ECPKParameters_fp(fp,x) \
+    ASN1_i2d_fp((i2d_of_void *)i2d_ECPKParameters, (fp), (void *)(x))
 
 int ECPKParameters_print(BIO *bp, const EC_GROUP *x, int off);
 # ifndef OPENSSL_NO_STDIO
--- a/test/build.info
+++ b/test/build.info
@@ -515,7 +515,9 @@ INCLUDE_MAIN___test_libtestutil_OLB = /INCLUDE=MAIN
     INCLUDE[sm4_internal_test]=.. ../include
     DEPEND[sm4_internal_test]=../libcrypto.a libtestutil.a
 
-    SOURCE[ec_internal_test]=ec_internal_test.c
+    SOURCE[ec_internal_test]=ec_internal_test.c \
+                             {- rebase_files("../apps",
+                                  split(/\s+/, $target{apps_init_src})) -}
     INCLUDE[ec_internal_test]=../include ../crypto/ec
     DEPEND[ec_internal_test]=../libcrypto.a libtestutil.a
 


crypto/ec/ec_curve.c
```diff
@@ -3097,6 +3098,32 @@ static EC_GROUP *ec_group_new_from_data(const ec_list_element curve)
...
+         * instead set the ASN1 flag to `OPENSSL_EC_EXPLICIT_CURVE`.
+         *
+         * Note that `OPENSSL_EC_NAMED_CURVE` is set as the default ASN1 flag on
+         * `EC_GROUP_new()`, when we don't have enough elements to determine if
+         * an OID for the curve name actually exists.
+         * We could implement this check on `EC_GROUP_set_curve_name()` but
+         * overloading the simple setter with this lookup could have a negative
+         * performance impact and unexpected consequences.
+         */
+        ASN1_OBJECT *asn1obj = OBJ_nid2obj(curve.nid);
```
romen, line 3115: The careful reviewer would notice that this check is superfluous (as hinted in the comment), given that `EC_GROUP_new()` will set the asn1 flag to `OPENSSL_EC_NAMED_CURVE`: the reason why I have a fixup commit styling it this way intentionally is that I want to have a local scope for declaring `ASN1_OBJECT *asn1obj` and to handle its life-cycle.

In `1.1.1` this is superfluous --but still looks tidy to me-- but in the perspective of forward-porting this to 3.0.0 it makes more sense as likely this whole section will be disabled when `FIPS_MODE` is defined, and having this local scope prevents ugly `#ifdef` spaghettization (and as an Italian :it: I am sensitive about spaghetti :spaghetti: :stuck_out_tongue_closed_eyes:).

romen:
Ping @bbbrumley as he raised the original issue. (Thanks!)

paulidale:
Approved after @bbbrumley has commented :)

bbbrumley:
Yes LGTM. Tests are now in place and the warnings are gone when I test this branch.

slontis:
It looks like appveyor has an issue..

test/ec_internal_test.c
```diff
@@ -183,6 +183,47 @@ static int field_tests_default(int n)
...
+    if (!TEST_ptr(g1 = EC_GROUP_new_by_curve_name(nid)))
+        goto end;
+
+    /* encode params to file */
+    if (!TEST_ptr(fp = fopen("params.der", "wb"))
+            || !TEST_true(i2d_ECPKParameters_fp(fp, g1)))
+        goto end;
+
+    /* flush and close file */
+    if (!TEST_int_eq(fclose(fp), 0)) {
```
slontis, line 204: I assume windoze is complaining about one of the fopen/fclose calls.. maybe this check here?

romen: @slontis do you have a suggestion on how to address it?

slontis: Reading the logs indicates it an applink related problem..
For a semi-related issue see https://github.com/openssl/openssl/issues/10086.
Maybe try using BIO instead and see if it disappears.

slontis: @romen ?

romen: Sorry I lost track of this.

So BIO is a no go, because the whole point of this PR is to add tests to cover the fp versions that are currently not tested.

So in windows it is not possible to use fopen and friends at all?
Should I just disable the new tests in windows? (what kind of defines could I use for that?)

slontis: If you looks at the tricks that the bio fp does to get the fp to work you could do it.. otherwise disable it.

bbbrumley: it's a regression test so IMO it is perfectly fine to just disable it on Windows. If someone uses it on Windows and it breaks again, they can fix it.

levitte:
Dudes, *do not forget to re-label when a PR has been approved!*

romen:
Uhm,  this is not ready to merge, because Appveyor shows failures.

Previous similar issues opened about windows problems using the _fp API have had the problem of not being reproducible.

romen:
I could disable the tests for windows, but the whole point of this PR is to make sure we are testing these API and we either document them as not working in windows or we should fix the problem.

test/ec_internal_test.c
levitte: The applink/uplink stuff ensures that libcrypto understands FILE the way the application that links with it does, not the way the C RTL libcrypto was linked with does. This is crucial when libcrypto is linked with the debugging C RTL and the application is not.

This wouldn't be necessary if BIO was a more complete IO library.

So you must build every program that passes a FILE to be wrapped in a BIO with `ms/app link.c` on Windows. The beginning of `apps/build.info` shows how to set up a practical variable for it.

We may have been lucky with some programs...

romen: @levitte I can see something related to this in the apps/build.info in master, but not in 1.1.1.

How is this problem solved in 1.1.1?

Should this PR fix the Windows build for all the tests or only for this one?

romen: Ping @levitte

romen:
I still need help in figuring out the windows tests

t8m:
I think the test could be simply skipped on Windows in 1.1.1.

19. 6. 2021 6:38, 6:38, Nicola Tuveri ***@***.***> napsal/a:

romen:
> I could disable the tests for windows, but the whole point of this PR is to make sure we are testing these API and we either document them as not working in windows or we should fix the problem.

@t8m

t8m:
Is this still relevant?

romen:
I rebased on latest `OpenSSL_1_1_1-stable`.

The issue is still relevant: you can try an interactive rebase and stop after each commit, building and then running `make TESTS=test_internal_ec VF=1 test`.

- After 4a144cd052, adding the tests for the macroes that were not tested, it shows that currently in `1.1.1` programs using `d2i_ECPKParameters_fp()` or `i2d_ECPKParameters_fp()` do not compile at all
- After fe952b93e0, adjusting the signatures, they compile, but get runtime errors for unnamed curves
- After 791b4fd05a, adjusting the logic for curves without OID, everything passes

What was left to do was to adjust the testing of `d2i_ECPKParameters_fp()` and `i2d_ECPKParameters_fp()` so that it can be tested in Windows, or document that that part of the API cannot work in Windows (or only under certain conditions, which we should be able to reproduce in our testing).

It stalled because I need help to figure what are the conditions to get those functions to work in Windows and how to test it properly.

romen:
Uhm, it looks like since we moved to GitHub CI we are not testing PRs against the LTS branch on Windows (and a lot more platforms)... Was this intended?

t8m:
> Was this intended?

No, simply nobody created the CI config for 1.1.1 for Windows builds. Feel free to add some.

romen:
> > Was this intended?
> 
> No, simply nobody created the CI config for 1.1.1 for Windows builds yet. Feel free to add some.

I am working on it in #16252

romen:
Rebased on top of latest `OpenSSL_1_1_1-stable` which now includes more tests, including testing on Windows that previously failed and cause this PR to stall.

romen:
So, the Windows Applink problems in our testing scripts are still present: https://github.com/openssl/openssl/pull/12457/checks?check_run_id=3366331006#step:8:421

I don't really know how to fix this:

- are the `*_fp` versions of our functions supposed to work in Windows at all?
  - if not, why aren't them masked in the public headers with some define?
  - if yes, why can't we get our tests to pass?
  - if we cannot get our testing rig to do the right thing for `*_fp` functions, how do I conditionally disable only this test in windows?

romen:
Is a program like the MWE in https://github.com/openssl/openssl/issues/12443#issue-656485274 supposed to work in Windows compiled against a shared OpenSSL dll?

romen:
> > Was this intended?
> 
> No, simply nobody created the CI config for 1.1.1 for Windows builds yet. Feel free to add some.

@t8m confirmed, it is still relevant, also for `master` (see #16355 where I started rebasing on `master`)

romen:
Preliminary results from the ongoing CI runs on #16355 seem to suggest that the Windows Applink problem might be limited to the testing rig, and not to the main source code.

@levitte , if that is the case, how difficult would it be to backport the black magic you did in master to sove these Applink issues into 1.1.1?
If that is not an option, can you suggest the proper combination of `#ifdef` or perl-fu to avoid compiling the new subtest in the problematic Windows environments?

bernd-edlinger:
I think you need to link the ec_internal_test with ms/applink.c,
see https://github.com/bernd-edlinger/openssl/commit/96a463cede0070aa5c86629d683a214657a9ba9e
for how that can be done.

romen:
Thanks a lot @bernd-edlinger , I cherry-picked your commit and it is now under testing!

romen:
@bernd-edlinger found a working solution! All tests are now passing!

bernd-edlinger:
@romen I've just spotted a typo in the commit message title:
"Fix d2i_ECPKParameters_fp and i2d_ECPKParameters_fp **macroes**"
the same typo is also in the commit message:
"Add tests for i2d_TYPE_fp and d2i_TYPE_fp
[...]
They are actually implemented as **macroes** and the absence of tests has
caused them to fall out-of-sync with the latest changes to ASN1 related
functions and cause compilation warnings."

romen:
Thanks again @bernd-edlinger , I will replace the occurrences of "macroes" with "macros"!

romen:
The last two commits do not change any code, and are just there as a remainder about fixing the typos in the commit messages when merging

romen:
#16355 has now been merged.
I just force pushed with commits squashed and amended for both the typos and to refer to the commits now on `master`.

@t8m can you please review this PR as well?

t8m (approved):

romen:
Merged to `1.1.1` with:

- 3d97638062 `Add tests for i2d_TYPE_fp and d2i_TYPE_fp`
- 45487dba0f `Fix d2i_ECPKParameters_fp and i2d_ECPKParameters_fp macros`
- f397efb0b9 `[ec] Do not default to OPENSSL_EC_NAMED_CURVE for curves without OID`
- a9972440d2 `Use applink to fix windows tests`


Thanks everyone for all the comments and the help!

PR #12458, opened 2020-07-16
mac: always pass a non-NULL output size pointer to providers.

### LINKED ISSUE #12425: EVP_MAC_final segfault on NULL outl
Hello,

while playing with the EVP_MAC API I noticed, that ``EVP_MAC_final`` sometimes crashed with a NULL reference if ``outl`` was NULL.

While looking into this I noticed, that everything is fine and as documented in ``mac_lib.c``:
https://github.com/openssl/openssl/blob/a01cae99ac384cb6a74b46ccdc90736fe0754958/crypto/evp/mac_lib.c#L120-L132

But the provider implementations work very different:

**Blake2, CMAC and poly1305 don't set it at all:**
https://github.com/openssl/openssl/blob/a01cae99ac384cb6a74b46ccdc90736fe0754958/providers/implementations/macs/blake2_mac_impl.c#L98-L105

https://github.com/openssl/openssl/blob/a01cae99ac384cb6a74b46ccdc90736fe0754958/providers/implementations/macs/cmac_prov.c#L116-L122

https://github.com/openssl/openssl/blob/a01cae99ac384cb6a74b46ccdc90736fe0754958/providers/implementations/macs/poly1305_prov.c#L91-L98

**GMAC and Siphash set it without a NULL check:**
https://github.com/openssl/openssl/blob/a01cae99ac384cb6a74b46ccdc90736fe0754958/providers/implementations/macs/gmac_prov.c#L111-L128

https://github.com/openssl/openssl/blob/a01cae99ac384cb6a74b46ccdc90736fe0754958/providers/implementations/macs/siphash_prov.c#L98-L109

**HMAC sets it with a NULL check:**
https://github.com/openssl/openssl/blob/a01cae99ac384cb6a74b46ccdc90736fe0754958/providers/implementations/macs/hmac_prov.c#L125-L136

**KMAC only sets it if the operation was ok:**
https://github.com/openssl/openssl/blob/a01cae99ac384cb6a74b46ccdc90736fe0754958/providers/implementations/macs/kmac_prov.c#L286-L304

So my first question would be, what is the expected correct behaviour? Who is resposible for setting this value?

If it is ``EVP_MAC_final`` is would suggest removing the parameter from the provider implementations altogether.

paulidale:
The backend code varies for the different MACs and sometimes sets the output
length, sometimes checks the return pointer and sometimes neither/either.

Fixes #12425

### DIFF
--- a/crypto/evp/mac_lib.c
+++ b/crypto/evp/mac_lib.c
@@ -120,15 +120,14 @@ int EVP_MAC_update(EVP_MAC_CTX *ctx, const unsigned char *data, size_t datalen)
 int EVP_MAC_final(EVP_MAC_CTX *ctx,
                   unsigned char *out, size_t *outl, size_t outsize)
 {
-    int l = EVP_MAC_size(ctx);
+    size_t l = EVP_MAC_size(ctx);
+    int res = 1;
 
-    if (l < 0)
-        return 0;
+    if (out != NULL)
+        res = ctx->meth->final(ctx->data, out, &l, outsize);
     if (outl != NULL)
         *outl = l;
-    if (out == NULL)
-        return 1;
-    return ctx->meth->final(ctx->data, out, outl, outsize);
+    return res;
 }
 
 /*


paulidale:
Travis failures are not relevant.

crypto/evp/mac_lib.c
```diff
@@ -120,15 +120,14 @@ int EVP_MAC_update(EVP_MAC_CTX *ctx, const unsigned char *data, size_t datalen)
 int EVP_MAC_final(EVP_MAC_CTX *ctx,
                   unsigned char *out, size_t *outl, size_t outsize)
 {
-    int l = EVP_MAC_size(ctx);
+    size_t l = EVP_MAC_size(ctx);
+    int res = 1;
 
-    if (l < 0)
```
slontis, line 125: It looks like EVP_MAC_size() cant be negative now - so this removal should be ok..

slontis:
Should this change update the commit to run the external tests?

crypto/evp/mac_lib.c
paulidale: It returns a _size_t_ so could never be negative.  It was being assigned into an _int_ and the negative check was for overflow.  This cannot happen now since the assignment is now to a _size_t_.

paulidale:
> Should this change update the commit to run the external tests?

To what purpose?

slontis:
> To what purpose?

I was under the assumption that EVP_MAC_final is called by a few things (including apps) that could be affected by the change..

paulidale:
Merged to master.

PR #12459, opened 2020-07-16
Mark an argument of an inline function as unused

petere:
This allows users of this header file to compile their own code with the gcc option `-Wunused-parameter`.

CLA: trivial

### DIFF
--- a/include/openssl/err.h
+++ b/include/openssl/err.h
@@ -220,7 +220,7 @@ static ossl_inline int ERR_GET_LIB(unsigned long errcode)
     return (errcode >> ERR_LIB_OFFSET) & ERR_LIB_MASK;
 }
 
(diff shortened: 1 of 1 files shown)


slontis:
ping

mattcaswell:
For the record I agree that this is trivial

mattcaswell:
@slontis - do you also agree? Is there any reason not to backport this to 1.1.1?

slontis:
agree on both

slontis:
In 1_1_1 it is a macro so it does not apply..
# define ERR_GET_FUNC(l)         (int)(((l) >> 12L) & 0xFFFL)

@mattcaswell or @t8m - do you agree?

t8m:
Ah, I did not look. So yes, it clearly does not apply to 1.1.1.

t8m:
Merged to master. Thank you for the PR.

PR #12466, opened 2020-07-16
Remove util/openssl-update-copyright

### LINKED ISSUE #12462: Drop util/openssl-update-copyright
We should drop the `util/openssl-update-copyright` script from the main repo as it is overridden by [release-tools/do-copyright-year @ openssl/tools](https://github.com/openssl/tools/blob/master/release-tools/do-copyright-year) .

Having the legacy script in the main repo codebase is just a disaster waiting to happen!

levitte:
It was useful at the time for a one-time run.  However, since it does
its work based on file modification time stamps, and those are
notoriously untrustable in a git checkout, it ends up being harmful.

There is a replacement in OpenSSL's tools repository, which relies on
git history.

Fixes #12462


levitte:
Note that this should happen in 1.1.1 as well.

levitte:
Merged

1bb78e72b9785c2cae40570e3aa8c9eb72d735e7 Remove util/openssl-update-copyright

PR #12469, opened 2020-07-16
Fix compiler warning about unused variable in err.h

nmathewson:
When using the err.h header with GCC 10 and -Wunused, the compiler
warns that ERR_GET_FUNC() no longer uses its argument.  This patch
adds a (void) cast to keep the compiler from complaining.


slontis:
You could use ossl_unused macro instead.
See https://stackoverflow.com/questions/3599160/how-to-suppress-unused-parameter-warnings-in-c

slontis:
Sorry but I am going to have to close this one also..
#12459 got in first.

PR #12468, opened 2020-07-16
Fix typo for SSL_get_peer_certificate()

levitte:

PR #12470, opened 2020-07-16
Fix a spelling error in ssl.h

nmathewson:
"certifiate" should be "certificate".

This change fixes linking when the deprecated SSL_get_per_certificate() is in use.


misery:
https://github.com/openssl/openssl/pull/12468 :-)

slontis:
I will close this as a duplicate.. (Only because it has a higher number than #12468.
Thank you for trying to fix it though.

PR #12472, opened 2020-07-17
Avoid errors with a priori inapplicable protocol bounds

### LINKED ISSUE #12394: MinProtocol can only be used for TLS or DTLS, not both.
In openssl 1.1.1, only one value can be set to MinProtocol (https://www.openssl.org/docs/man1.1.1/man3/SSL_CONF_cmd.html)

If MinProtocol is set to "TLSv1.2", it is not possible for an application to use "DTLSv1.2".
To restrict the used protocol versions to the currently support ones >=1.2, you have to use the deprecated "Protocol" command:

```
# MinProtocol = TLSv1.2
Protocol = -SSLv3, -TLSv1, -TLSv1.1, TLSv1.2, TLSv1.3, -DTLSv1, DTLSv1.2
```

It should be possible, to set a minimal version for TLS and DTLS.

This issue (and possible solutions) have already been discussed on the openssl-users mailinglist: https://www.mail-archive.com/openssl-users@openssl.org/msg88405.html

vdukhovni:
The 'MinProtocol' and 'MaxProtocol' configuration commands now silently
ignore TLS protocol version bounds when configurign DTLS-based contexts,
and conversely, silently ignore DTLS protocol version bounds when
configuring TLS-based contexts.  The commands can be repeated to set
bounds of both types.  The same applies with the corresponding
"min_protocol" and "max_protocol" command-line switches, in case some
application uses both TLS and DTLS.

SSL_CTX instances that are created for a fixed protocol version (e.g.
TLSv1_server_method()) also silently ignore version bounds.  Previously
attempts to apply bounds to these protocol versions would result in an
error.  Now only the "version-flexible" SSL_CTX instances are subject to
limits in configuration files in command-line options.

##### Checklist
- [ x] documentation is added or updated

This may resolve #12394

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,22 @@ OpenSSL 3.0
 
 ### Changes between 1.1.1 and 3.0 [xx XXX xxxx]
 
+ * The 'MinProtocol' and 'MaxProtocol' configuration commands now silently
+   ignore TLS protocol version bounds when configurign DTLS-based contexts, and
+   conversely, silently ignore DTLS protocol version bounds when configuring
+   TLS-based contexts.  The commands can be repeated to set bounds of both
+   types.  The same applies with the corresponding "min_protocol" and
+   "max_protocol" command-line switches, in case some application uses both TLS
+   and DTLS.
+
+   SSL_CTX instances that are created for a fixed protocol version (e.g.
+   TLSv1_server_method()) also silently ignore version bounds.  Previously
+   attempts to apply bounds to these protocol versions would result in an
+   error.  Now only the "version-flexible" SSL_CTX instances are subject to
+   limits in configuration files in command-line options.
+
+   *Viktor Dukhovni*
+
  * Deprecated the `ENGINE` API.  Engines should be replaced with providers
    going forward.
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,7 +24,7 @@ OpenSSL 3.0
 ### Changes between 1.1.1 and 3.0 [xx XXX xxxx]
 
  * The 'MinProtocol' and 'MaxProtocol' configuration commands now silently
-   ignore TLS protocol version bounds when configurign DTLS-based contexts, and
+   ignore TLS protocol version bounds when configuring DTLS-based contexts, and
    conversely, silently ignore DTLS protocol version bounds when configuring
    TLS-based contexts.  The commands can be repeated to set bounds of both
    types.  The same applies with the corresponding "min_protocol" and
--- a/doc/man3/SSL_CONF_cmd.pod
+++ b/doc/man3/SSL_CONF_cmd.pod
@@ -180,7 +180,7 @@ See L<openssl-ciphers(1)> for more information.
 
 Sets the minimum and maximum supported protocol.
 Currently supported protocol values are B<SSLv3>, B<TLSv1>, B<TLSv1.1>,
-B<TLSv1.2>, B<TLSv1.3> for TLS and B<DTLSv1>, B<DTLSv1.2> for DTLS, and B<None>
+B<TLSv1.2>, B<TLSv1.3> for TLS; B<DTLSv1>, B<DTLSv1.2> for DTLS, and B<None>
 for no limit.
 If either the lower or upper bound is not specified then only the other bound
 applies, if specified.
@@ -398,7 +398,7 @@ The SSL and TLS bounds apply only to TLS-based contexts, while the DTLS bounds
 apply only to DTLS-based contexts.
 The command can be repeated with one instance setting a TLS bound, and the
 other setting a DTLS bound.
-The value B<None> applies to both types of contexts and disables the limit.
+The value B<None> applies to both types of contexts and disables the limits.
 
 =item B<MaxProtocol>
 
@@ -410,7 +410,7 @@ The SSL and TLS bounds apply only to TLS-based contexts, while the DTLS bounds
 apply only to DTLS-based contexts.
 The command can be repeated with one instance setting a TLS bound, and the
 other setting a DTLS bound.
-The value B<None> applies to both types of contexts and disables the limit.
+The value B<None> applies to both types of contexts and disables the limits.
 
 =item B<Protocol>
 
--- a/doc/man3/SSL_CONF_cmd.pod
+++ b/doc/man3/SSL_CONF_cmd.pod
@@ -178,12 +178,17 @@ See L<openssl-ciphers(1)> for more information.
 
 =item B<-min_protocol> I<minprot>, B<-max_protocol> I<maxprot>
 
-Sets the minimum and maximum supported protocol.  Currently supported
-protocol values are B<SSLv3>, B<TLSv1>, B<TLSv1.1>, B<TLSv1.2>, B<TLSv1.3>
-for TLS and B<DTLSv1>, B<DTLSv1.2> for DTLS, and B<None> for no limit.
-If either bound is not specified then only the other bound applies,
-if specified.  To restrict the supported protocol versions use these
-commands rather than the deprecated alternative commands below.
+Sets the minimum and maximum supported protocol.
+Currently supported protocol values are B<SSLv3>, B<TLSv1>, B<TLSv1.1>,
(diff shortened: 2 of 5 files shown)


CHANGES.md
```diff
@@ -23,6 +23,22 @@ OpenSSL 3.0
 
 ### Changes between 1.1.1 and 3.0 [xx XXX xxxx]
 
+ * The 'MinProtocol' and 'MaxProtocol' configuration commands now silently
+   ignore TLS protocol version bounds when configurign DTLS-based contexts, and
```
paulidale, line 27: _configuring_ typo.

doc/man3/SSL_CONF_cmd.pod
```diff
@@ -178,12 +178,17 @@ See L<openssl-ciphers(1)> for more information.
...
 
-Sets the minimum and maximum supported protocol.  Currently supported
-protocol values are B<SSLv3>, B<TLSv1>, B<TLSv1.1>, B<TLSv1.2>, B<TLSv1.3>
-for TLS and B<DTLSv1>, B<DTLSv1.2> for DTLS, and B<None> for no limit.
-If either bound is not specified then only the other bound applies,
-if specified.  To restrict the supported protocol versions use these
-commands rather than the deprecated alternative commands below.
+Sets the minimum and maximum supported protocol.
+Currently supported protocol values are B<SSLv3>, B<TLSv1>, B<TLSv1.1>,
+B<TLSv1.2>, B<TLSv1.3> for TLS and B<DTLSv1>, B<DTLSv1.2> for DTLS, and B<None>
```
paulidale, line 183: Just a suggestion: `B<TLSv1.2>, B<TLSv1.3> for TLS; B<DTLSv1>, B<DTLSv1.2> for DTLS, and B<None>` which avoids a list of the form A and B and C, changing it into A; B and C.

doc/man3/SSL_CONF_cmd.pod
```diff
@@ -178,12 +178,17 @@ See L<openssl-ciphers(1)> for more information.
...
+for no limit.
+If either the lower or upper bound is not specified then only the other bound
```
paulidale, line 185: What about this instead?

`If only one of the lower or upper bounds is specified, then the other bound does not apply.`

doc/man3/SSL_CONF_cmd.pod
```diff
@@ -389,15 +394,23 @@ This sets the minimum supported SSL, TLS or DTLS version.
 
 Currently supported protocol values are B<SSLv3>, B<TLSv1>, B<TLSv1.1>,
 B<TLSv1.2>, B<TLSv1.3>, B<DTLSv1> and B<DTLSv1.2>.
-The value B<None> will disable the limit.
+The SSL and TLS bounds apply only to TLS-based contexts, while the DTLS bounds
+apply only to DTLS-based contexts.
+The command can be repeated with one instance setting a TLS bound, and the
+other setting a DTLS bound.
+The value B<None> applies to both types of contexts and disables the limit.
```
paulidale, line 401: `s/limit/limits/`

doc/man3/SSL_CONF_cmd.pod
```diff
@@ -389,15 +394,23 @@ This sets the minimum supported SSL, TLS or DTLS version.
...
 This sets the maximum supported SSL, TLS or DTLS version.
 
 Currently supported protocol values are B<SSLv3>, B<TLSv1>, B<TLSv1.1>,
 B<TLSv1.2>, B<TLSv1.3>, B<DTLSv1> and B<DTLSv1.2>.
-The value B<None> will disable the limit.
+The SSL and TLS bounds apply only to TLS-based contexts, while the DTLS bounds
+apply only to DTLS-based contexts.
+The command can be repeated with one instance setting a TLS bound, and the
+other setting a DTLS bound.
+The value B<None> applies to both types of contexts and disables the limit.
```
paulidale, line 413: Again _limits_ should be plural.

ssl/statem/statem_lib.c
```diff
@@ -1679,11 +1679,22 @@ int ssl_check_version_downgrade(SSL *s)
...
+
     if (version == 0) {
         *bound = version;
         return 1;
     }
 
+    valid_tls = version >= SSL3_VERSION && version <= TLS_MAX_VERSION_INTERNAL;
+    valid_dtls =
+        DTLS_VERSION_LE(version, DTLS_MAX_VERSION_INTERNAL) &&
+        DTLS_VERSION_GE(version, DTLS1_BAD_VER);
```
paulidale, line 1693: A naïve question: why allow equality vs `DTLS1_BAD_VER`?

vdukhovni: > A naïve question: why allow equality vs `DTLS1_BAD_VER`?

Because it isn't actually a "bad version", rather it is a non-standard version used in some IIRC Cisco products, and still supported by OpenSSL, though we don't have a string limit that matches this version, at the API level, users can specify the constant. @mattcaswell may be able to confirm/deny.

doc/man3/SSL_CONF_cmd.pod
vdukhovni: I think this is one is fine as-is...

doc/man3/SSL_CONF_cmd.pod
vdukhovni: Sure.

CHANGES.md
vdukhovni: Thanks.

doc/man3/SSL_CONF_cmd.pod
vdukhovni: Sure.

ssl/statem/statem_lib.c
paulidale: Okay, makes sense.

paulidale:
I'm okay with this going to 1.1.1 as well, but won't insist on it.

vdukhovni:
@paulidale Thanks for the review. I didn't see you comment one way or the other on the topic of whether the protocol version bounds should raise errors when applications try to use a fixed version-inflexible `method` outside the specified min/max range.  For example, should there be any friction in using:
```C
SSL_CTX_new(TLSv1_client_method())
```
When the config file says `MinProtocol TLSv1.2`.  My take at this point is, no.  If an application is hard-coded for a particular protocol then breaking it is not terribly productive.  We're not raising the floor in that case to prevent downgrades, we're disabling support, and that can be better done at compile time (removing support for the method, ...).

Some might disagree.

Finally, some part or all of this should be backported to 1.1.1, after it is merged into master.  I see this as a bugfix.  My 2015 code was a bit too cautious on silently ignoring what could be construed as error conditions, but in retrospect, these sort of "conflicts" are expected, and we should be sensible about avoid breakage, while applying the policy when it makes sense.

paulidale:
I didn't see it as a question and am okay with the modified behaviour.  Ideally, I think it should produce a warning but we don't have that capability.

vdukhovni:
> I didn't see it as a question and am okay with the modified behaviour. Ideally, I think it should produce a warning but we don't have that capability.

It wasn't actually a question, but I just want to make sure that the change does not go unnoticed, and reviewers consider it carefully on the merits.

t8m:
IMO even warning would not be productive unless explicitly asked for.  In case you apply system-wide defaults you usually want to be able to override them with individual application settings without giving a warning. Although I could see some usefulness if there was a configuration entry that said - I really mean this configuration and do not want it to be overridden by applications at all.

vdukhovni:
One of the strict-warnings, no-deprecated builds failed, but it seems to be a pervasive issue with builds of the master branch, not related to this PR: https://travis-ci.com/github/openssl/openssl/jobs/361990397
Would be nice to fix or drop that build.

mattcaswell:
Looks like this is ready to go, so I applied the "ready-to-merge"label.

ssl/statem/statem_lib.c
mattcaswell: I can "confirm" @vdukhovni's description of `DTLS1_BAD_VER` above

vdukhovni:
No further comments, so squashed, rebased and merged. Will now open a backport PR for 1.1.1

PR #12474, opened 2020-07-17
Specific the engine pointer when EVP_CipherInit_ex

gujinqiangggs:
CLA: trivial

I found that when I wanted to use an engine by the option-engine XXX , it didn't work. Checking the code, I guess it missed the engine pointer when calling EVP_CipherInit_ex.

##### Checklist


gujinqiangggs:
CLA: trivial

InfoHunter:
> CLA: trivial

This needs to be included in the commit message, instead of Github comment.

